PR #2831, opened 2021-06-22
BUG: .drop(fields). The argument can now be either a list of strings …

### LINKED ISSUE #2829: Weird behavior of .drop(fields)
```
import string

my_project = 'pj-becfr-eagle-ci-dev'
my_dataset = 'dm_express_assortment_V1'

import ibis
import ibis_bigquery
from google.cloud import bigquery
import pandas as pd

my_table = 'issue_3_table'

pdf = pd.DataFrame({'a': [1], 'b': [2], 'c': [3]})

# Load client
client = bigquery.Client(project=my_project)

# Load data to BQ
client.delete_table(my_dataset + "." + my_table, not_found_ok=True)
job = client.load_table_from_dataframe(pdf, my_dataset + "." + my_table)

conn = ibis_bigquery.connect(
    project_id=my_project,
    dataset_id=my_dataset)

t = conn.table(my_table)

e = t.drop("ab")

e.schema()  # a and b are removed

e = t.mutate(ab=t.c)

e.drop("ab")  # a and b are removed


e = t.mutate(ab=t.c)

e.drop(["ab"])  # ab is removed

e = t.mutate(bz=t.c)

e.drop("bz")  => KeyError: "Fields not in table: frozenset({'z'})"

e.drop(["bz"])  # bz is removed

e.drop(e.a) => ValueError: The truth value of an Ibis expression is not defined

e.drop([e.a]) => KeyError: "Fields not in table: frozenset({ref_0\nBigQueryTable[table]\n  name: ...
```

I think there is a mismatch between a string and a list of strings

The .drop() method takes a list/sequence/... of fields as argument.

From the examples above, each field can be a string (at least).

So t.drop(["ab", "bc"]) attempts to remove the attributes: "ab" and "bc" and that's fine.

It seems that if I call t.drop("ab"), ibis considers that the 2 attributes are "a" and "b". 
But, that's not correct because the elements of a string are characters and not strings, unless we 
suppose that field names can also be characters...

Anyway, that's quite  confusing.

jcmincke:
Closes #2829


ibis/expr/api.py
```diff
@@ -4419,6 +4419,11 @@ def _table_view(self):
 
 
 def _table_drop(self, fields):
+
+    if isinstance(fields, str):
+        #  We want to  drop just one attribute.
+        return _table_drop(self, [fields])
```
datapythonista, line 4425: I'd prefer to use `fields = [fields]`, instead of recursively calling the function (which IMHO makes things more difficult to understand, and adds an overheat for no reason.

Also, maybe just personal preference, but seems like it'd make sense to keep the `if not fields:` first, and just check this `if` if fields is present.

docs/source/release/index.rst
```diff
@@ -12,6 +12,8 @@ Release Notes
    These release notes are for versions of ibis **1.0 and later**. Release
    notes for pre-1.0 versions of ibis can be found at :doc:`release-pre-1.0`
 
+* :bug:`2829` Fix .drop(fields). The argument can now be either a list of strings or a
+string. In the latter case, the given string is transformed into one element list.
```
datapythonista, line 16: If you check below, looks like for multi-line comments, we indent at the beginning of the test, not the bullet point (in short, can you indent this line two more spaces please).

ibis/tests/sql/test_compiler.py
```diff
@@ -2557,3 +2557,22 @@ def test_table_drop_with_filter():
...
+    )
+
+    e1 = t.projection(["a", "c"])
+    e2 = t.drop(["b"])
+    e3 = t.drop("b")
+
+    assert e1.schema() == e2.schema()
+    assert e1.schema() == e3.schema()
+
+    assert not(e1.schema() == t.schema())
```
datapythonista, line 2574:
```suggestion
    assert e1.schema() != t.schema()
```

ibis/tests/sql/test_compiler.py
```diff
@@ -2557,3 +2557,22 @@ def test_table_drop_with_filter():
...
+
+    assert not("b" in e1.columns)
```
datapythonista, line 2576:
```suggestion
    assert "b" not in e1.columns
```

datapythonista (commented):
Thanks for the PR @jcmincke, looks good. Added few minor comments, but other than that, nice improvement!

docs/source/release/index.rst
```diff
@@ -12,6 +12,8 @@ Release Notes
    These release notes are for versions of ibis **1.0 and later**. Release
    notes for pre-1.0 versions of ibis can be found at :doc:`release-pre-1.0`
 
+* :bug:`2829` Fix .drop(fields). The argument can now be either a list of strings or a string.
+  In the latter case, the given string is transformed into one element list.
```
datapythonista, line 16: These notes are for Ibis users. I think they'll care more on how this change affects them, `.drop()` now accepts single columns, than how we implemented this internally (transforming the string to on element list). Do you mind rephrasing a bit please? Sorry I didn't see this in the previous review. Thanks!

datapythonista (approved):
Thanks @jcmincke. Will wait to have a second pair of eyes here, but lgtm.

ibis/expr/api.py
```diff
@@ -4419,10 +4419,15 @@ def _table_view(self):
 
 
 def _table_drop(self, fields):
+
     if not fields:
         # no-op if nothing to be dropped
         return self
 
+    if isinstance(fields, str):
```
jreback, line 4427: hmm there isn't any doc-string, can you add one, also  type fields.

ibis/tests/sql/test_compiler.py
```diff
@@ -2557,3 +2557,22 @@ def test_table_drop_with_filter():
     ON t0.`b` = t1.`b`
 WHERE t0.`a` < 1.0"""
     assert result == expected
+
+
+def test_table_drop_consistency():
+    t = ibis.table(
```
jreback, line 2563: can you add the issue number as a comment

ibis/tests/sql/test_compiler.py
```diff
@@ -2557,3 +2557,22 @@ def test_table_drop_with_filter():
...
+        [('a', 'int64'), ('b', 'string'), ('c', 'timestamp')], name='t'
+    )
+
+    e1 = t.projection(["a", "c"])
```
jreback, line 2567: can you call this expected

ibis/tests/sql/test_compiler.py
```diff
@@ -2557,3 +2557,22 @@ def test_table_drop_with_filter():
...
+    e2 = t.drop(["b"])
```
jreback, line 2568: can you call this result1 and result2 for the one below

jreback (changes requested):

ibis/expr/api.py
```diff
@@ -4418,11 +4418,32 @@ def _table_view(self):
...
 
 
-def _table_drop(self, fields):
+def _table_drop(self: ir.TableExpr,
+                fields: Union[str, List[str]]
+                ) -> ir.TableExpr:
+    """
+    Remove one or more fields from a table.
+
+    _table_drop(table, "a-string") is equivalent
```
jreback, line 4427: you can remove these 2 lines, they are extraneous

jreback (changes requested):
small comment, ping on green

ibis/expr/api.py
```diff
@@ -4418,11 +4418,32 @@ def _table_view(self):
     return new_view.to_expr()
 
 
-def _table_drop(self, fields):
+def _table_drop(self: ir.TableExpr,
```
datapythonista, line 4421:
```suggestion
def _table_drop(self,
```
The convention with mypy is to not annotate `self`.

ibis/expr/api.py
jcmincke: I don't understand. You asked me: "hmm there isn't any doc-string, can you add one, also type fields."

Where should the doc-string be added?

ibis/expr/api.py
jcmincke: Do you also type the return value.

datapythonista: Yes. The idea is that mypy already knows which is the type of self, since it's the Python interpreter who provides it. But for the return, is the user (author of the function) who decides what is being returned, so mypy needs to be told.

ibis/expr/api.py
datapythonista: Docstring is fine where it is. Jeff's comment is that you can remove those two lines about the two calls being equivalent.

jreback:
pls also merge master

datapythonista:
@jcmincke do you have time to finish this? Just removing an annotation and a paragraph of the docstring is missing to merge this. Also, there is a small conflict in the release notes, if you can merge master and fix please. Thanks!

datapythonista:
Thanks for the update @jcmincke, looks like you'll have to run black to get the CI green.

datapythonista:
@jreback your comments were addressed, if you want to have a last look before merging...

jreback (approved):

datapythonista:
Thanks for the contribution @jcmincke

PR #2828, opened 2021-06-17
Roadmap proposal

datapythonista:
I think it'd be good to have a roadmap defined on what we would like to see in Ibis after 2.0 is released. So, users have an idea on what to expect, as well as contributors, who can also use the roadmap to know what makes sense to work on, and what doesn't.

Looks like there was already a roadmap, but by mistake wasn't published on the web. And most things have already been taken care of. So I remove everything except standardizing the UDF functionality, which still makes a lot of sense.

Would be good to have feedback on the points I added, as well as other things that could be added.

PR #2833, opened 2021-06-22
Load backends in a lazy way

### LINKED ISSUE #2827: Circular import when loading external backends
Looks like we've got a problem I haven't anticipated (and not sure why I didn't face this before while updating the OmniSci backend). We're currently loading all the backend found on entrypoints on `import ibis`. But backends also need to `import ibis`, so, there is a circular import when loading external backends.

The obvious solution would be to implement `__getattr__` for the `ibis` module. So, on `ibis.sqlite` the backend would be loaded lazily. The problem with that is shown in the next code and there is #2671 for it:
```python
import ibis

ibis.options.impala.temp_db = 'foo'

client = ibis.impala.connect(...)
```
We need to load the imapala backend on `import ibis`, if we do in `ibis.impala.connect(...)` the impala options are not defined, and `ibis.options.impala.temp_db = ...` fails. Loading the backend on options could be an option but seems too hacky to me.

So, in my opinion, we should take care of #2671 before the release (that would involve breaking changes, not just raising a `FutureWarning`). And I'd also implement #2532 which completely solves the problem, and backends would only be loaded after they are needed by an `ibis.connect(...)` call.

@jreback does this make sense to you? Any other idea?

datapythonista:
Closes #2827 

@jreback, there is a problem here, not sure how to move forward with.

Currently, pyarrow data types are registered into Ibis in the parquet backend. Since now backends are loaded in a lazy way, the next code fails unless the parquet backend is loaded first:
```python
import pyarrow

dt.dtype(pyarrow.int8())
```
There is a test failing because of this. I can easily solve it by calling `ibis.parquet` before the test. But not sure if that's what we want. Do we really want to register arrow types when the parquet backend is loaded? Or when?


jreback:
> Do we really want to register arrow types when the parquet backend is loaded? Or when?

right this is a fundamental issue. in theory we would want to actually load types w/o the entire backend, but this is not easily possible. So really no choice but to load when the backend is loaded. 

where is this a problem exactly?

datapythonista:
> where is this a problem exactly?

The test of the pyarrow dtypes is an Ibis core test, and the registration is in the parquet backend. So, the problem is when testing backends without the parquet backend being loaded. Those are the tests: `ibis/tests/expr/test_datatypes_arrow.py::test_convert_arrow`

For now I'm just loading the pandas and parquet backends at Ibis start. I think we need changes to the architecture of Ibis to solve properly. I opened #2834 for now, but I think there is more than that.

jreback:
> For now I'm just loading the pandas and parquet backends at Ibis start. I think we need changes to the architecture of Ibis to solve properly. I opened #2834 for now, but I think there is more than that.

yep sounds like a good idea to investigate

ibis/config.py
```diff
@@ -200,7 +202,17 @@ def __getattr__(self, key):
         try:
             v = self.d[key]
         except KeyError as e:
-            raise AttributeError(*e.args)
+            # See if the options doesn't exist because it's from a backend
+            # not yet loaded. Try to load, and try the option again.
+            try:
```
jreback, line 207: do we go down this path in tests?

jreback (changes requested):

jreback:
prob rebase to fix dask

ibis/config.py
datapythonista: Good point. Looks like we don't have any tests for options. Our options code is a copy of the pandas one, which does have tests. I don't think copying pandas tests, or implementing new ones are great ideas. If we want to do things properly, we should create a new package for the options code, and use it in both projects.

Let me know your thoughts, but in any case, I don't think this is for this PR.

datapythonista:
@jreback this is ready now. Main changes since you last reviewed:
- We register backends options on `ibis.__getattr__`, so in tests I had to replace `ibis.backends.dask.Backend()` by `ibis.dask`, so options are registered.
- Looks like the Spark backend got the `FloorDivide` operation registered from the PySpark backend. Now that PySpark is not necessarily loaded, had to implement it there.

Thanks!

ibis/__init__.py
```diff
@@ -51,15 +53,29 @@
...
-    except ImportError:
-        pass
-    else:
-        backend = backend_module.Backend()
-        setattr(ibis, entry_point.name, backend)
-        with ibis.config.config_prefix(entry_point.name):
-            backend.register_options()
+
+def __getattr__(name: str):
+    entry_points = list(
```
jreback, line 58: can you add a doc-string and return type

ibis/__init__.py
```diff
@@ -51,15 +53,29 @@
...
+        pkg_resources.iter_entry_points(group='ibis.backends', name=name)
+    )
+    if len(entry_points) == 0:
+        raise AttributeError(
+            f"module 'ibis' has no attribute '{name}'. "
```
jreback, line 63: are these 2 tested? why are we warning on multiple entry points? (and not raising)

ibis/config.py
jreback: i think you can remove *all* of this because you *already* load the options in the backend loading itself. I woudn't load the backend here at all.

jreback (changes requested):

jreback:
do you need to update any docs?

ibis/__init__.py
```diff
@@ -1,12 +1,14 @@
 """Initialize Ibis module."""
+import warnings
+
 import pkg_resources
 
+# Converting an Ibis schema to a pandas DataFrame requires registering
```
icexelloss, line 6: Sorry, do you mean `converting an Ibis schema to a pandas dtype` here? I am not sure what do you mean by converting ibis schema to a pandas DataFrame..

icexelloss: Also IMO, I think it's OK to treat pandas backend special since pandas can be a hard dependency and is kind of baked in the ibis core/expr layer.

ibis/__init__.py
```diff
@@ -51,15 +53,29 @@
...
+    )
+    if len(entry_points) == 0:
+        raise AttributeError(
+            f"module 'ibis' has no attribute '{name}'. "
+            f"If you are trying to access the '{name}' backend, "
+            f"try installing it first with `pip install ibis-{name}`"
+        )
+    elif len(entry_points) > 1:
+        warnings.warn(
+            f"More than one entrypoint found for backend '{name}', "
```
icexelloss, line 69: Sorry not super familiar with this. What exactly mean to have more than one entrypoint for backend?

icexelloss: (And why do we want to warn on this?)

ibis/backends/spark/compiler.py
```diff
@@ -50,6 +50,12 @@ class SparkExprTranslator(ExprTranslator):
 rewrites = SparkExprTranslator.rewrites
 
 
+@rewrites(ops.FloorDivide)
```
icexelloss, line 53: I like this.

But..is this related to this PR? XD

ibis/config.py
```diff
@@ -200,7 +202,17 @@ def __getattr__(self, key):
         try:
             v = self.d[key]
         except KeyError as e:
-            raise AttributeError(*e.args)
+            # See if the options doesn't exist because it's from a backend
```
icexelloss, line 205: In what way it breaks if we didn't do this here? It seems to me to be a reasonable behavior to error out if user tries to access an option defined by a backend without importing that backend first...

ibis/expr/schema.py
```diff
@@ -9,6 +9,28 @@
...
+in_dtype : Union[np.dtype, pandas_dtype]
+    The dtype of `column`, used for dispatching
+out_dtype : ibis.expr.datatypes.DataType
+    The requested ibis type of the output
+column : pd.Series
+    The column to convert
+
+Returns
+-------
+result : pd.Series
```
icexelloss, line 29: A bit confused.. Does return a dtype or a pd.Series?

ibis/expr/schema.py
```diff
@@ -133,6 +155,39 @@ def name_at_position(self, i):
             )
         return self.names[i]
 
+    def apply_to(self, df):
+        """Applies the Ibis schema to a pandas DataFrame
```
icexelloss, line 159: Sorry what does this method do? Does it cast the types of df according to this ibis schema?

icexelloss:
@datapythonista Looks good high level! I left some comments and would appreciate it if you take a look.

ibis/config.py
datapythonista: Unfortunately this is needed. See this code:
```python
ibis.options.impala. temp_table = 'foo'
ibis.impala.connect(...)
```
Without this code, the first line will fail, since there is no `impala.temp_table` option registered. The loading of the backend you're referring to is loaded in the second line.

ibis/__init__.py
datapythonista: Imagine that two different developers create packages for bigquery. And both of them add an entrypoint with the name `bigquery`. When a user with both packages installed call `ibis.bigquery`, both will be found, but the first to be found is used, and the second will be ignored. @jreback suggests to raise an exception. I thought it'd make sense to move forward, knowing the user (with this warning) which of the packages will be used.

I'm ok raising an exception. And for your feedback, I think it makes sense to rephrase the message, to make it less technical, and more user friendly. Thanks for it.

ibis/backends/spark/compiler.py
datapythonista: Somehow related. This is missing in master and tests fail because the lack of registering `FloorDivide` in Spark. In master tests don't really fail, because the PySpark backend is also loaded, and looks like the PySpark backend is also registering the Spark operation. I need to change this here (or in a separate PR merged before this one), because here we load one backend at a time, and without loading PySpark, the Spark backend tests fail.

I'm already working on removing the spark backend, so this addition makes the tests pass here, but this code won't live very long, so no big deal I think, even an empty function would be ok IMO.

ibis/config.py
datapythonista: This change makes this code work:
```python
import ibis
ibis.options.impala.temp_table = 'foo'
ibis.impala.connect(...)
```
If we break here, and make users write the next code, then wouldn't be needed (some other changes would be needed to this PR for it to work):
```python
import ibis
import ibis.backends.impala

ibis.options.impala.temp_table = 'foo'
ibis.impala.connect(...)
```

ibis/expr/schema.py
datapythonista: I just moved this function of file without change, probably better to fix things in a follow up. I can have a look since this is just the docs, but in general better to not make changes that don't show in a commit diff.

ibis/expr/schema.py
datapythonista: As said in the previous comment, I just moved this method, since IMHO it was in the wrong file. And I finally ended up always loading the pandas backend, but if we eventually don't load it at `import ibis`, this change would be needed, since this code is needed even if the backend is not loaded.

But answering your question, my understanding is that when we convert the cursor returned by the backend to pandas, types are inferred by pandas. And this method takes the original schema of the ibis expression (or let's say the SQL table), and casts columns to have the same types.

ibis/__init__.py
datapythonista: Thanks, yes, this is not clear, will rephrase. This refers to the `Schema.apply_to` I explained below.

Personally I don't think pandas should be a hard dependency, and the pandas backend shouldn't be preloaded. I don't even think the pandas backend should be shipped with Ibis in the long term. You can see my proposed roadmap at #2828, so you can have an idea on my thoughts for the project. And any feedback there is very welcome.

datapythonista:
I think I addressed all comments, and commented in the ones were I think no change should be make. Please let me know if I missed anything, or if you have further comments.

ibis/config.py
jreback: this is easy to replace by
```
import ibis.impala
ibis.options.impala.....
ibis.impala.connect()
```
right?

ibis/__init__.py
jreback: but please do not change at this time. pandas is not a big deal as a dep at all.

ibis/__init__.py
jreback: let's raise an exception. this is not tested at all in any event.

ibis/config.py
jreback: yes i think thi sis better

ibis/config.py
datapythonista: It would be:
```python
import ibis
ibis.impala
ibis.options.impala.temp_table = 'foo'
ibis.impala.connect()
```
But, yes, that's correct, if we are happy that the next code fails when setting the option, this is not needed:
```python
import ibis
ibis.options.impala.temp_table = 'foo'
ibis.impala.connect()
```

ibis/__init__.py
datapythonista: I changed it, now it's raising an exception, and there is a test for it.

ibis/__init__.py
datapythonista: Mhh, the change is not to make pandas not be a dependency (even if it would help if we do that). The change is trivial, just changing the function to the obvious file where it should be. And also makes it a regular method, instead of injecting it as a method in a hacky way. I think it's a small, but useful clean up, with no drawbacks. What is the problem with keeping it?

ibis/backends/spark/compiler.py
icexelloss: kk thanks for the explanation. SGTM

ibis/config.py
icexelloss: kk SGTM

ibis/expr/schema.py
icexelloss: kk SGTM

ibis/expr/schema.py
icexelloss: kk SGTM

icexelloss:
Thanks @datapythonista. +1 from me.

ibis/config.py
datapythonista: Should we simply use something like:
```python
import ibis.backends.impala

ibis.backends.impala.set_option('temp_table', 'foo')
ibis.backends.impala.connect(...)
```
or the syntax I proposed in the past
```python
import ibis

ibis.backends.impala.connect(..., temp_table='foo')
# or
ibis.backends.impala.connect(..., options={'temp_table': 'foo'})
```


I think `ibis.<backend-name>` is already cumbersome, and caused several problems. And IMHO this is becoming even more cumbersome, to have to import an unrelated module to make the above code work is far from what I'd expect to have to write as a user.

jreback: i prefer

```
import ibis.backends.impala 
ibis.backends.impala.set_option('temp_table', 'foo') 
ibis.backends.impala.connect(...)```
```

connect has nothing to do with setting the options of the backend itself (and may actually have multiple invokations). typing these together is not great.

all that said this can be a followup.

ibis/config.py
jreback: see comments above

jreback: i don't really love all of these nested try/excepts

jreback (approved):

jreback:
ok happy to merge this, but i do thing before release should fix up the option setting stuff.

datapythonista:
> ok happy to merge this, but i do thing before release should fix up the option setting stuff.

Thanks. I guess we may want to keep backward compatibility for Ibis 2.0, with deprecation warnings where needed. So, I guess it may make sense to keep this code until after the release. But next week I'll be opening issues to discuss more in detail what we exactly want to have, for the release and long term.

PR #2836, opened 2021-06-28
TST: Run tests for backends defined as entrypoints

datapythonista:
The backend tests currently run for all backends in `ibis/backends/`. With this PR it will run with backends defined as entrypoints. So, besides the core backends i nthat directory, tests can be run from third-party backends.

### DIFF
--- a/ibis/backends/tests/conftest.py
+++ b/ibis/backends/tests/conftest.py
@@ -1,6 +1,6 @@
 import importlib
 import os
-from pathlib import Path
+import pkg_resources
 from typing import List
 
 import pandas as pd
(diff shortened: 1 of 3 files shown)


datapythonista:
@jreback this is a blocker to continue adapting the backends to Ibis changes. Whenever you've got time would be good if you can review it, it's a very quick one.

ibis/backends/tests/conftest.py
```diff
@@ -1,9 +1,9 @@
 import importlib
 import os
-from pathlib import Path
 from typing import List
 
 import pandas as pd
+import pkg_resources
```
jreback, line 6: so we have setup tools in the requirements?

jreback (changes requested):

ibis/backends/tests/conftest.py
datapythonista: Good point. Just checked and we don't. We're already using `pkg_resources`, so I guess we're getting setuptools from a dependency.

I added setuptools to our dependencies (to `requirements.txt` which is read by `setup.py`).

jreback (approved):

jreback:
thanks @datapythonista

Issue #2823, opened 2021-06-11
chore: enable passing mypy checks


datapythonista:
We've got a mypy check in the CI, but we force it to be green, since mypy has errors.

I see that at the moment there are some issues with duplicate names of files, and mypy is not even checking or reporting errors other than that.

It'd be good to spend some time soon in fixing mypy errors, and failing the CI for mypy errors, otherwise we're adding type annotations for nothing.

cpcloud:
Over the next year or so we should make decision about whether to spend more time adding type annotations and enabling their verification in CI.

There are a number of challenges in either approach, but we should settle the issue this year.

cpcloud:
I think we're going to go about this incrementally as opposed to a mega-PR that addresses all the code in ibis and runs mypy in CI.

Feel free to submit PRs that add type annotations.

Issue #2804, opened 2021-06-02
API to interact with databases


datapythonista:
The current API to interact with databases is a bit cumbersome in my opinion. I think we should simplify it.

To set a database at connection, list databases, and use the default database, things are simple, no comments here:
To list the available databases, set a 

```python
con = ibis.my_backend.connect(..., database='default_db')

con.list_databases()

my_table = con.table('my_table')
```

But then, things become not very clear IMHO. If I want to get a table from a different database, there are different ways that a user could expect they work, based on what we have:

```python
con.table('other_db_table', database='other')

con.database('other').table('other_db_table')

con.set_database('other')
con.table('other_db_table')

con.current_database = 'other'
con.table('other_db_table')
```

I don't think we're making a favor to the users by providing so many options, but the opposite. To me what makes sense is to have a single way to change the default database `con.current_database = 'other'` (whatever name with think it's the best, `database`, `current_database`, `default_database`, `db`... no strong preference). And to have a single way to create a table, `con.table(name, database=default_database)`. I think everything else should be removed from the public API.

Would be good to know from users who use multiple databases, if this makes sense.

saulpw:
Revisiting this suggestion, a few points:

1. Is it reasonable to just create a new connection to interact with another database?  Is there any reason why someone would need/want to interact with two databases over the same connection?
2. We're increasingly heading towards URLs to specify connection parameters, including the database.  In light of this, maybe we can deprecate some of the other methods and kwargs?  Do people use these programmatically, or is it reasonable to focus exclusively on URLs?
3. If we're going to standardize on some of these methods, how about we keep the `set_database()` call, but disallow manually setting the `.current_database` attribute (convert it to a readonly property)?

Thoughts and opinions welcome!

Issue #2793, opened 2021-05-24
Deprecate aliases


datapythonista:
There are a significant amount of things in Ibis that can be done in more than one way. This is particularly true for the Impala backend. For example:

```python
expr.compile() or ibis.client.compile(expr)

ibis.client.drop_table('table_name') or con.table('table_name').drop()
```

I'm planning to remove the aliases, and just leave the `expr.compile()` or `table.drop()` syntax.

Please let me know if anyone has got comments.

Issue #2678, opened 2021-03-10
Backend public API


datapythonista:
At the moment there is no clear public API for the backends. Backends so far implement a `connect` method that returns a `Client` class. That's not required, since what is exposed at the moment is the backend module itself, so it's possible to implement a backend like:
```python
ibis.snowflake.start(server='whatever')
```
Which can return something totally inconsistent with the rest of `Client` classes of the backends. While so far backend developers tried to be consistent, and there are inconsistencies, but small, and probably created by mistake, I think we can do better.

In a first effort, I created a class named `Backend` to be exposed instead of the models, and all backends subclass an ABC, so some methods are guaranteed to exist, like the `connect` method. While IMO this is a good first step, and helped standardize things, like the `compile` and `verify` methods, which are now part of the ABC, ensuring consistency, and avoiding a decent amount of duplicate code.

I've been checking, and what it would make sense to me is to merge this `Backend` class (which used to be the functions in the backend module) with the `Client` classes. They became separate for historical reasons on the API being exposed as the backend, but I think the new resulting class should be what we define as the public backend API.

I'm unsure of what's the best name for the new class, I think good candidates are:
- Backend
- Client
- Connection

Given code like this:
```python
connection = ibis.connect('sqlite://db.sqlite3')
```
Feels like the most natural name for the class is connection, instead of the current `Client` or `Backend`. But all seem reasonable.

And I think it should implement the functionality to do things like:
- `connect`
- `list_databases`
- `database`
- `list_tables`
- `table`
- `compile`

And probably other optional things like `insert`, `drop_table`...

If that sounds good with everyone, and there are no strong preferences about the name, I'll be implementing this, and call the new class `Connection`.

tswast:
Would "connect" return such a class? Or would it be a `classmethod` factory method?

datapythonista:
Open to discussion, but personally I think `connect` should be a function, and return a class (a `Connection` class, which would be the union of the current `Backend` and `Client`). Happy to hear other opinions.

PR #2837, opened 2021-06-28
WIP: Initial version of base Connection classes

datapythonista:
xref #2678 

Starting the definition of the backend base classes. Planning to do it iteratively, otherwise this will end up being a huge and complex PR, while the changes should be simple.

The idea of these new classes is to be the main base class for backends, replacing `Backend` and `Client`. All the methods defined in them already exist in those two classes, just reorganizing in a better way.

I think we should do some clean up of the API before the release. But for now just keeping what's in Ibis, and when the new base classes are finished and used by the backends we can discuss. Examples of things I'd change:
- Merge `execute` and `raw_sql`, and have a single method to execute, which accepts an expression or a string (or an AST if useful). Same for the schema methods, which we have one for schema from expression, and schema from AST.
- Simplify API to interact with databases (see #2804)
- Make some methods private. Some are public because they are public for backend authors who need to implement them, but shouldn't be exposed to final users
- Delete duplicate stuff, and leave a single way to execute things (see #2793)
- Just an idea, but maybe we could group DML methods under a `DML` class or something, instead in the top level, which is already a huge namespace

The new classes aren't used yet, so there may be errors, that will be fixed in follow ups, since this is not yet being tested.

We also need to consider that we are not running mypy (see #2823). I'm adding types here, which should be mostly correct, but nothing is being validated. And enforcing types is a decent effort, not sure if something to do before or after the release

@jreback let me know if any thoughts about the above


jreback:
why are we changing the name from Backend? that is a pretty informative name.

datapythonista:
> why are we changing the name from Backend? that is a pretty informative name.

I'm fine with `Backend` too, I explain why I think `Connection` it's better in the issue, #2678, at the end of the description.

ibis/backends/base/connection.py
```diff
@@ -0,0 +1,187 @@
...
+            Spark based backends.
+        """
+        pass
+
+    @abc.abstractmethod
+    def connect(connection_string: str, **options) -> BaseConnection:
+        """
+        Connect to the underlying database and return a client object.
+        """
+        pass
```
icexelloss, line 49: I don't think we need `pass` for abstractmethod?

ibis/backends/base/connection.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def register_options(self) -> None:
+        """
+        If the backend has custom options, register them here.
+        They will be prefixed with the name of the backend.
+        """
+        pass
+
+    def compile(self, expr: ir.Expr, params=None) -> str:
```
icexelloss, line 58: I think only SQL backend would return a str for this. So many should change this to `Any` in base class.

ibis/backends/base/connection.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        Compile the expression.
+        """
+        return self.compiler.to_sql(expr, params=params)
+
+    def verify(self, expr: ir.Expr, params=None):
```
icexelloss, line 64:
```
def verify(self, expr: ir.Expr, params=None) -> bool:
```
?

ibis/backends/base/connection.py
```diff
@@ -0,0 +1,187 @@
...
+        in SQL backends, a NullLiteral operation could be translated simply
+        with the string "NULL".
+
+        Examples
+        --------
+        >>> @ibis.sqlite.add_operation(ibis.expr.operations.NullLiteral)
+        ... def _null_literal(translator, expression):
+        ...     return 'NULL'
+        """
+        if not hasattr(self.client, 'compiler'):
```
icexelloss, line 92: Maybe better to leave an empty implementation here and implement in SQL backends?

ibis/backends/base/connection.py
```diff
@@ -0,0 +1,187 @@
...
+                operation, translation_function
+            )
+
+        return decorator
+
+    @property
+    def current_database(self) -> str:
+        """Return the current database."""
+        # TODO: If we can assume this, we should define the `con` attribute
+        return self.con.database
```
icexelloss, line 108: What is `self.con`?

ibis/backends/base/connection.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def database(self, name: Optional[str] = None):
```
icexelloss, line 110: type annotate return type?

ibis/backends/base/sql/connection.py
```diff
@@ -0,0 +1,115 @@
...
+        Returns
+        -------
+        Backend cursor
+        """
+        cursor = self.con.execute(query)
+        if cursor:
+            return cursor
+        cursor.release()
+
+    def execute(self, expr, params=None, limit='default', **kwargs):
```
icexelloss, line 43: Perhaps this method can be lifted to base class?

icexelloss: (Since all backend supports execution)

icexelloss:
Thanks @datapythonista !

I like the idea of union "Backend" and "Client" since I think they are pretty much the same thing. But I am with @jreback that I think `Backend` is better. `Connection` to be seems to be a pretty simple/"dumb" class just from how the word is used in, for example, "HTTP Connection". I feel `Backend` is much more complex as it could do pretty confiscated work like translate to SQL or to a PySpark DataFrame. It's kind of a "complier".

Orthogonally, since this looks like breaking change if you are going to remove the existing Client and Backend classes in favor of this one. So I am curious about your thinking around when is a stable time to refactor implementation such as ibis-bigquery to work with newest ibis. Internally we can no longer pull the ibis default (we cherry pick from ibis default now to keep big-query working) because we need both ibis and ibis-bigquery at the moment.

icexelloss (changes requested):
Some comments

datapythonista:
Thanks for all the comments @icexelloss, very useful. Will address comments and reply more in detail later. But regarding backward compatibility: This doesn't break compatibility with users (so far), or the previous changes. The interface with third-party backends has been broken for a while with the refactoring, but I'm working on porting the OmniSciDB backend to the latest Ibis at the moment, and we can quickly do the same for bigquery after that. Sorry it's been annoying, but I don't think it should be a big concern, since it's few hours of work and fully under our control to keep the master of the backends we control in sync. Let me know if I'm wrong or missing something, but if you're using Ibis master in the company (I wasn't aware), we can plan better those things. It'd also be useful for me if you can tell me which backends are you using. I guess PySpark and BigQuery (not sure if Impala), but if it's few we can prioritize those.

Regarding compile/execute, I want to have a look, I don't think pandas should use them more than for compatibility. Not sure if requiring them in the base class is the best we can do.

jreback:
> > why are we changing the name from Backend? that is a pretty informative name.
> 
> I'm fine with `Backend` too, I explain why I think `Connection` it's better in the issue, #2678, at the end of the description. tl;dr: this class is being returned by `ibis.backend.connect()`, so `Connection` is more intuitive than `Backend` IMHO.

i c, so if we had to write a description (and we should)
- backend is the package of where the code lives
- `Connection` is kind of like Client/Backend now, e.g. its a connection to a database (or equiv for non sql ones).
?

datapythonista:
> i c, so if we had to write a description (and we should)
> 
> * backend is the package of where the code lives
> * `Connection` is kind of like Client/Backend now, e.g. its a connection to a database (or equiv for non sql ones).
>   ?

That's what I would do, yes. I think for end users that terminology should be more intuitive. If their code is something like:
```python
db = ibis.sqlite.connect('geography.db')
con = ibis.sqlite.connect('geography.db')
connection = ibis.sqlite.connect('geography.db')
```
I think the class being returned being a `Connection` makes more sense. Also, that's consistent with [SQLAlchemy Connection](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection) and [DBAPI2 Connection[(https://www.python.org/dev/peps/pep-0249/#connection-objects) which are equivalent classes in my opinion.

And agree with the docs, I'll review all backend documentation after the release, add a page on how to implement backends, and also see if in the tutorial or elsewhere it makes sense to add more info on how the backend system work, how to install backends...

ibis/backends/base/connection.py
datapythonista: This already exists in Ibis. I'm +1 on removing it, not much value added, but in this PR I'm aiming at not making breaking changes, and deprecate whatever is needed (lots of things IMHO) in dedicated follow ups.

ibis/backends/base/connection.py
datapythonista: Good catch. I moved it to the SQL Connection class, I don't think we need an empty method here, I think this is specific of SQL backends.

ibis/backends/base/connection.py
datapythonista: It's assumed to be the actual connection with the database, a DBAPI2 `Connection` instance. This is hacky, I added the TODO to take care of this later, but this is how it is right now in master, just copied the method.

ibis/backends/base/connection.py
icexelloss: kk

ibis/backends/base/connection.py
icexelloss: kk

icexelloss (approved):
+1

datapythonista:
@icexelloss Thanks for the reviews. There are still pending your comments on whether `compile` and `execute` should be in the base connection or only for SQL backends. I thinkI agree for `execute`, since we can have lazy non-SQL backends like Dask. But I want to have a look at the current behavior. Feels like having `compile` in pandas is more a hack for consistency of methods than a good decision, but I need to research before having an informed opinion.

icexelloss:
Looks good to me to merge.

* A few more thoughts on `Connection` vs `Backend`
1.  `Connection` does make more sense if this is connecting to an actual database. It might make less sense if the backend is `pandas` `spark` or `dask` since those backends doesn't really have a concept of a "connection". (In pandas backend, there isn't connection at all. In Spark, a similar concept is called `SparkSession`). So admittedly, there isn't really an universally good name for the class. And that's why I personally still feel `Backend` is perhaps a better name since it's more generic than `Connection` .

2.  It might feel a bit more natural to say "execute expression with a backend" than "execute expression with a connection". 
But I am OK with calling it `Connection` 

(internally we have a thin wrapped class around this object that we customize slightly and this object is not exposed directly to the users, therefore not a big issue for us what is this called)

* On our internal sync with ibis default (To provide some context)
This is a little annoying but not too bad for us now XD.

We try to pull ibis default changes into our internal environment instead of waiting for ibis releases, since our internal project is in very active development and relies pretty heavily on ibis changes. So currently we are in a state of not able to pull default changes because the compat issue with ibis-bigquery default. This is not a huge issue yet because we have been cherry-picking changes from ibis default and skip the refactor related changes, but I imagine this could break when we start to touch the same piece of code.

> but I don't think it should be a big concern, since it's few hours of work and fully under our control to keep the master of the backends we control in sync

It is currently a bit hard to update big query right now since the default ibis is not released so we need to wait on a new ibis in order to fix bigquery. But agree it's not a big concern.

icexelloss:
> There are still pending your comments on whether compile and execute should be in the base connection or only for SQL backends.

Yeah not a big concern I think and can be addressed follow ups. I agree the `compile` for pandas backend does nothing and can probably raise error (use should never need to call this with pandas backend). Anyway, my thoughts are

`compile` : return a backend specific object, for sql backend, this is `str`. For PySpark backend, this is `pyspark.sql.DataFrame` and for Dask backend, this is `dd.DataFrame`

`execute`: execute the expression and return as a `pd.DataFrame` (bring to memory)

jreback (changes requested):
this should just be called Backend. I think naming this as Connection is just causing more churn and confusion.

datapythonista:
Superseded by #2883

PR #2838, opened 2021-06-28
Loading TestConf correctly for third-party backends

datapythonista:
Follow up of #2836

I didn't realise that the `TestConf` class was being loaded from the path in the Ibis repo. So that part had to be changed to load relative to the module in #2836. Changing it here, so Ibis backend tests can run in third-party backends.

### DIFF
--- a/ibis/backends/tests/conftest.py
+++ b/ibis/backends/tests/conftest.py
@@ -33,16 +33,15 @@ def _backend_name_to_class(backend_str: str):
     """
     Convert a backend string to the test configuration class for the backend.
     """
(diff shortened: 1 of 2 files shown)


jreback:
> test_backends_are_cached 

is failing (but i think on master)

jreback (approved):

datapythonista:
> is failing (but i think on master)

My bad, looks like depending on the order of the tests the backends are loaded when that test is executed. The failing assert was just to make sure that the test asserts were running, so replaced it by deleting the backend attribute if exists.

Issue #2795, opened 2021-05-25
chore(dev-deps): locally reproducible builds


kszucs:
Is there an easy way to locally reproduce the CI builds?

If I recall correctly ibis used to have a docker-compose configuration, but I can't find it anymore.

datapythonista:
We don't have the docker-compose configuration anymore.  If you want to start a docker, you can get the parameters from the CI, and caller docker with them. For example, for postgres you've got the name of the image and the parameters to use in here https://github.com/ibis-project/ibis/blob/master/.github/workflows/main.yml#L62

datapythonista:
The previous system, running Ibis itself in a docker image was making things very complex. Having a docker-compose file for the backends that need one seems reasonable. The only thing I don't like is that tings will be duplicated with the CI (which is way simpler if we use GitHub actions services for docker images, instead of our own).

I'm not sure if it's your case, but I'd assume most developers of Ibis care about one, or a small subset of backends. Does it make sense that you've got your own docker-compose with the backends that you care, instead of having to maintain a docker-compose file with all the backends in the repo? IMHO all you want is a shortcut for a long docker call, so keeping this out of the repo makes sense to me. Does it make sense to you?

kszucs:
> The previous system, running Ibis itself in a docker image was making things very complex. Having a docker-compose file for the backends that need one seems reasonable. The only thing I don't like is that tings will be duplicated with the CI (which is way simpler if we use GitHub actions services for docker images, instead of our own).

The service definitions in GHA are almost identical to a docker-compose configuration while the latter would allow to spin up the same services both locally and GHA without code duplication.
It would be nice to be able to spin up the backends I'm interested in locally with a single command, including the testing data (this could be reused on CI to reduce maintenance cost). Something like:

```
ci/start_backends.sh mysql clickhouse  # perhaps this could be implemented in setup_env.sh
PYTEST_BACKENDS="mysql clickhouse" ci/run_tests.sh
```

> I'm not sure if it's your case, but I'd assume most developers of Ibis care about one, or a small subset of backends. 

I have plans for implementing new backends which may involve some refactoring as well eventually affecting multiple backends. 

> Does it make sense that you've got your own docker-compose with the backends that you care, instead of having to maintain a docker-compose file with all the backends in the repo? IMHO all you want is a shortcut for a long docker call, so keeping this out of the repo makes sense to me. Does it make sense to you?

Sure, I can keep it locally though I think it would make sense to have it upstream.

[...]

PR #2839, opened 2021-06-29
clickhouse: support `group_concat` operation

Lunaticus7:
Added `group_concat` operation to Clickhouse.

### DIFF
--- a/docs/source/release/index.rst
+++ b/docs/source/release/index.rst
@@ -18,6 +18,7 @@ Release Notes
 * :feature:`2873` Support binary operation with Timedelta in Pyspark backend
 * :support:`2865` Deprecated `ibis.<backend>.verify()` in favor of capturing exception in `ibis.<backend>.compile()`
 * :bug:`2845` Fix projection on differences and intersections for SQL backends
+* :feature:`2839`: Add `group_concat` operation for Clickhouse backend
 * :bug:`2827` Backends are loaded in a lazy way, so third-party backends can import Ibis without circular imports
 * :bug:`2830` Disable aggregation optimization due to N squared performance
 * :bug:`2821` Fix `.cast()` to array outputting list instead of np.array in Pandas backend
--- a/ibis/backends/clickhouse/registry.py
+++ b/ibis/backends/clickhouse/registry.py
@@ -1,6 +1,7 @@
 from datetime import date, datetime
 from io import StringIO
 
+import ibis
 import ibis.common.exceptions as com
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
@@ -512,6 +513,15 @@ def _string_like(translator, expr):
     )
 
 
+def _group_concat(translator, expr):
+    arg, sep, where = expr.op().args
+    if where is not None:
+        arg = where.ifelse(arg, ibis.NA)
+    return 'arrayStringConcat(groupArray({}), {})'.format(
+        *map(translator.translate, (arg, sep))
+    )
+
+
 # TODO: clickhouse uses different string functions
 #       for ascii and utf-8 encodings,
 
@@ -567,7 +577,7 @@ operation_registry = {
     ops.Min: _agg('min'),
     ops.StandardDev: _agg_variance_like('stddev'),
     ops.Variance: _agg_variance_like('var'),
-    # ops.GroupConcat: fixed_arity('group_concat', 2),
+    ops.GroupConcat: _group_concat,
     ops.Count: _agg('count'),
     ops.CountDistinct: _agg('uniq'),
     ops.Arbitrary: _arbitrary,
--- a/ibis/backends/clickhouse/tests/test_functions.py
+++ b/ibis/backends/clickhouse/tests/test_functions.py
@@ -532,3 +532,32 @@ def test_count_distinct_with_filter(alltypes):
     expected = alltypes.string_col.execute()
     expected = expected[expected.astype('int64') > 1].nunique()
     assert result == expected
+
+
+@pytest.mark.parametrize(
+    ('sep', 'where_case', 'expected'),
+    [
+        (',', None, "arrayStringConcat(groupArray(`string_col`), ',')"),
+        ('-', None, "arrayStringConcat(groupArray(`string_col`), '-')"),
+        pytest.param(
+            ',',
+            0,
+            (
+                "arrayStringConcat(groupArray("
+                "CASE WHEN `bool_col` = 0 THEN "
+                "`string_col` ELSE Null END), ',')"
+            ),
+            marks=pytest.mark.xfail(
+                reason=(
+                    '`where` param needs `Nullable` column '
+                    'but the all in testing data is not.'
+                    'See also issue #2891'
+                )
+            ),
+        ),
+    ],
+)
+def test_group_concat(alltypes, sep, where_case, expected, translate):
+    where = None if where_case is None else alltypes.bool_col == where_case
+    expr = alltypes.string_col.group_concat(sep, where)
+    assert translate(expr) == expected


jreback (changes requested):
@Lunaticus7 thanks can you add a release note

kszucs:
LGTM, though the linter fails.

jreback:
@Lunaticus7 can you run black: see https://github.com/ibis-project/ibis/pull/2839/checks?check_run_id=2942181027

Lunaticus7:
Thanks for review, release note has been added. 
@jreback @kszucs

docs/source/release/index.rst
```diff
@@ -11,7 +11,7 @@ Release Notes
 
    These release notes are for versions of ibis **1.0 and later**. Release
    notes for pre-1.0 versions of ibis can be found at :doc:`release-pre-1.0`
-
+* :feature:`2839`: Add `group_concat` operation for Clickhouse backend
```
datapythonista, line 14: Can you please leave a blank line between the previous paragraph and the list. I don't think this will render correctly in the html.

ibis/backends/clickhouse/registry.py
```diff
@@ -512,6 +513,17 @@ def _string_like(translator, expr):
     )
 
 
+def _group_concat(translator, expr):
+    arg, sep, where = expr.op().args
+    if where is not None:
+        arg = where.ifelse(arg, ibis.NA)
+        arg = translator.translate(arg)
```
datapythonista, line 520: Is this needed? You're already applying `translator.translate` to `arg` in the `return`.

Do you mind adding another test that hits the code in the `if` please.

ibis/backends/clickhouse/tests/test_functions.py
```diff
@@ -532,3 +532,11 @@ def test_count_distinct_with_filter(alltypes):
...
     expected = expected[expected.astype('int64') > 1].nunique()
     assert result == expected
+
+
+def test_group_concat(con, alltypes, translate):
+    t = alltypes
+    expr = t.string_col.group_concat(',')
+    expected = """arrayStringConcat(groupArray(`string_col`), ',')"""
+    assert translate(expr) == expected
+    assert len(con.execute(expr))
```
datapythonista, line 542: Can we be more specific on what this needs to return? Also, I guess the `len` is unnecessary, if `len(...) != 0` I guess this will evaluate to true without the `len`. But would be better to test the value, or whatever possible, more than just saying that this should return something not empty, if possible.

datapythonista (changes requested):
Thanks @Lunaticus7, nice work. I added couple of comments, if you don't mind having a look.

ibis/backends/clickhouse/registry.py
Lunaticus7: you are right，it will cause a bug 😣

ibis/backends/clickhouse/tests/test_functions.py
Lunaticus7: I have add test for `where` param in `group_concat` .
emmmm，i'm new in ibis and haven't found the right way to load test data and run ci locally，so i have no ideal to specify the return value of the group_concat expr 😇.

ibis/backends/clickhouse/tests/test_functions.py
```diff
@@ -532,3 +532,15 @@ def test_count_distinct_with_filter(alltypes):
...
+
+def test_group_concat(con, alltypes, translate):
+    t = alltypes
+    expr = t.string_col.group_concat(',')
+    expected = """arrayStringConcat(groupArray(`string_col`), ',')"""
+    assert translate(expr) == expected
+
+    expr = t.string_col.group_concat(sep='-', where=t.bool_col == 0)
+    expected = """arrayStringConcat(groupArray(\
+CASE WHEN `bool_col` = 0 THEN `string_col` ELSE Null END), '-')"""
```
datapythonista, line 545: No big deal, but I think this approach is more readable (maybe just to me):

```suggestion
    expected = ("arrayStringConcat(groupArray("
                "CASE WHEN `bool_col` = 0 THEN `string_col` ELSE Null END), '-')")
```

Also, it'd probably be nice to have this as a separate test, so when it fails, we know which of the two asserts is failing faster.

datapythonista (approved):
Thanks @Lunaticus7, looks good, just couple of suggestions, in case they make sense to you.

datapythonista:
@Lunaticus7 tests are failing, not sure if you've seen it.

Lunaticus7:
> @Lunaticus7 tests are failing, not sure if you've seen it.

Sorry for the delay.

I'm fighting with running CI locally 🤣(https://github.com/ibis-project/ibis/issues/2795#issuecomment-880681579)  and a little busy recently, maybe i could done this in the next week.

datapythonista:
I personally don't ever run the whole CI locally. I have the environment set up (the conda env), and then run the tests for specific backends. For example:
```python
PYTEST_BACKENDS="pandas sqlite" python -m pytest ibis/tests ibis/backends/tests
```

For clickhouse you can run the same, but you need to run the docker image, and call the scripts to download and load the test data to it.

datapythonista:
@Lunaticus7 do you have time to fix the conflicts and the broken tests? This shouldn't be far from ready.

Lunaticus7:
> @Lunaticus7 do you have time to fix the conflicts and the broken tests? This shouldn't be far from ready.

Thanks for your tips of test, it works 🎉!

About the broken test, i found all the column in testing data is `non-nullable`, so when set predicate as  `where` param in `groupArray`, it will cause an error:

> IbisTypeError: Cannot compute precedence for string[non-nullable] and null types

So i remove the test case about `where` temporary.
Maybe it's worthy to add some `nullable` data in testing data?

datapythonista:
There are conflicts with master, can you merge master into your branch and fix them please.

For the tests, if one can't pass right now, it's better to xfail it than to delete it. I'm not sure what's the best for the nulls. Can you check what other backends are doing, and try to do the same.

Lunaticus7:
@datapythonista 
I have xfailed the where case and rebased to master.

And there are some details about the null error

```python
# for clickhouse backends and ibis-testing-data

import ibis

connection = ibis.clickhouse.connect(
    host='127.0.0.1',
    database='ibis_testing',
    port=9000
)
table_alltypes = connection.table('functional_alltypes')

expr = table_alltypes.string_col.group_concat(',', table_alltypes.bool_col == 1)

expr.compile()
# will cause:
# > IbisTypeError: Cannot compute precedence for string[non-nullable] and null types

# the expected sql of `expr`
sql = """
SELECT arrayStringConcat(groupArray(CASE WHEN `bool_col` = 1 THEN `string_col` ELSE Null END), '-') AS `tmp`
FROM ibis_testing.`functional_alltypes`
"""

connection.raw_sql(sql)
# can get right result
```

I also tested the same case on mysql backend，and got the same result.

It seems like a bug of ibis on computing precedence of dtypes, when target column(e.g. `string_col`) is `string[non-nullable]`.
Should i open an issue about it?

datapythonista:
Sorry for the delay @Lunaticus7. Since the bug is general and unrelated to this change, I think we can move forward here, and open an issue for the general error. Does it make sense?

Lunaticus7:
@datapythonista Issue of the null error has moved to https://github.com/ibis-project/ibis/issues/2891
I have rebased my branch onto the newes master, is there anything else i should do 🤓?

Lunaticus7:
Lint fixed, I forgot pyproject.toml when black locally 🤕.

ibis/backends/clickhouse/tests/test_functions.py
```diff
@@ -532,3 +532,31 @@ def test_count_distinct_with_filter(alltypes):
     expected = alltypes.string_col.execute()
     expected = expected[expected.astype('int64') > 1].nunique()
     assert result == expected
+
+
+@pytest.mark.parametrize(
+    # FIXME: The `where` param needs `Nullable` column
+    #   but the all in testing data is not.
```
datapythonista, line 539: Is this related to the xfail? You should be able to add a message to the xfail itself, and if for whatever reason that's not an option, don't make it a `FIXME`, which in general is for tasks that need to be done before it's merged.

In the xfail message, besides saying why the test is failing, you can add the issue number you created for it.

ibis/backends/clickhouse/tests/test_functions.py
```diff
@@ -532,3 +532,31 @@ def test_count_distinct_with_filter(alltypes):
...
+            ),
+            marks=pytest.mark.xfail,
+        ),
+    ],
+)
+def test_group_concat(alltypes, sep, where_case, expected, translate):
+    if where_case is None:
+        where = None
+    else:
+        where = alltypes.bool_col == where_case
```
datapythonista, line 560:
```suggestion
    where = None if where_case is None else alltypes.bool_col == where_case
```

datapythonista (commented):
Thanks @Lunaticus7 just couple of small things, but look great.

Lunaticus7:
> Thanks @Lunaticus7 just couple of small things, but look great.

Thanks for your patience, it's all done 🥳.

datapythonista (approved):
Thanks @Lunaticus7

jreback (approved):

jreback:
thanks @Lunaticus7

Issue #2711, opened 2021-03-26
ibis.case() unintentionally forces computation in the dask backend


gerrymanoim:
When the dask backend hits `np.select` as part of https://github.com/ibis-project/ibis/blob/d2702fd12fe3e03477b00cd716aa4cafbc3c6bf3/ibis/backends/pandas/execution/generic.py#L1073 (or other case handlers there), `np.select` seems to be silently forcing execution of the underlying dask table. This (I believe) also affects `execute_simple_case_series` in dask backend. 

Test hitting this behavior: https://github.com/ibis-project/ibis/blob/b28536d12871e1611e5644aa2d47fa7dc5e21bcb/ibis/backends/tests/test_generic.py#L160

Demonstration:

```python
(Pdb++) whens
[Dask Series Structure:
npartitions=2
0       bool
3650     ...
7299     ...
Name: int_col, dtype: bool
Dask Name: eq, 6 tasks, Dask Series Structure:
npartitions=2
0       bool
3650     ...
7299     ...
Name: int_col, dtype: bool
Dask Name: eq, 6 tasks]
(Pdb++) thens
[20, 10]
(Pdb++) otherwise
0
(Pdb++) np.select(whens, thens, otherwise)
array([10, 20,  0, ...,  0,  0,  0])
```

gerrymanoim:
I think `np` is hitting https://github.com/dask/dask/blob/d6943a666264d44c6199bd46dca708fb4b94c70c/dask/dataframe/core.py#L378 which is forcing the compute.

Issue #2553, opened 2020-12-01
test: dask Backend specific xfailed tests


gerrymanoim:
This tracks tests xfailed **only** in the dask backend (i.e. not the pandas backend). This provides a good tracker of compatibility differences. In the codebase sections of not working code will be marked `TODO - <reason keyword> - #2553`. 

# Backend Specific Tests

| Done? | test file | test name | reason | keyword |GH Issue |
| ------- | --------- | --------- | ------ | ------- | ------- |
| :x: | test_datatypes | test_series_to_ibis_literal | `ir.as_value_expr` does not handle `dd.Series` | as_value_expr | |
| :x: | test_arrays | test_array_collect_rolling_partitioned | Windowing not implemented | windowing | |
| :x: | test_join | test_join_with_non_trivial_key | `execute_materialized_join` is broken | execute_materialized_join | |
| :x: | test_join | test_join_with_non_trivial_key_project_table | `execute_materialized_join` is broken | execute_materialized_join | |
| :x: | test_operations | test_project_scope_does_not_override | Windowing not implemented | windowing | |
| :heavy_check_mark: | test_operations | test_aggregation_group_by | Aggregations not implemented | aggregations | |
| :heavy_check_mark: | test_operations | test_boolean_aggregation | Aggregations not implemented | aggregations | |
| :x: | test_operations | test_sort_by | Sorting partially implemented | sorting | |
| :x: | test_operations | test_complex_sort_by | Sorting partially implemented | sorting | |
| :heavy_check_mark: | test_operations | test_weighted_average | Grouping not implemented | grouping | |
| :heavy_check_mark: | test_operations | test_groupby_with_unnamed_arithmetic | Grouping not implemented | grouping | |
| :heavy_check_mark: | test_operations | test_cast_on_group_by | Grouping not implemented | grouping | |
| :heavy_check_mark: | test_operations | test_left_binary_op_gb | Aggregations not implemented | aggregations | |
| :heavy_check_mark: | test_structs | test_struct_field_series_group_by_value | Grouping not implemented | grouping | |
| :x: | test_timecontext | test_context_adjustment_window | Windowing not implemented | windowing | |
| :x: | test_timecontext | test_setting_timecontext_in_scope | Windowing not implemented | windowing | |
| :x: | test_timecontext | test_context_adjustment_multi_window | Windowing not implemented | windowing | |
| :x: | test_timecontext | test_context_adjustment_window_groupby_id | Windowing not implemented | windowing | |
[...]

PR #2843, opened 2021-07-01
ENH: Support no-op windows in the dask backend

gerrymanoim:
Previously (in https://github.com/ibis-project/ibis/pull/2786) I tried to handle windowing + grouping/ordering in the same PR  and it got very messy. This is a simpler PR that "unwraps" a windowing operation and executes the logic inside. Only supports "no-op" windows (those with no lead/lag/group by/order by parameters).  But this can occur in situations like `test_analytic_udf_mutate`. The `add_partitioned_sorted_column` will be used in follow up PRs to support grouping and ordering windows (and I can probably get rid of all usages of `utils.safe_concat` now).

This code also fixes https://github.com/ibis-project/ibis/issues/2711 by using `da.select`. This was only introduced in dask 2021.6.1, so I've ported the code into a shim. https://github.com/ibis-project/ibis/issues/2847 is a follow up issue to remove this. 

Other follow-ups:

- Grouping + ordering windows
- consolidate some of the code around "wrapping a dd.core.Scalar" into a collection that dask can work with downstream. 
- Look into getting rid of `utils.safe_concat` now that selection is indexed

Post Merge:

- [ ] Update https://github.com/ibis-project/ibis/issues/2553

### DIFF
--- a/ibis/backends/dask/__init__.py
+++ b/ibis/backends/dask/__init__.py
@@ -62,3 +62,5 @@ class Backend(BaseBackend):
             'See ibis.dask.trace for details.',
             validator=ibis.config.is_bool,
         )
+        # forces the pandas backend to register options
+        getattr(ibis, "pandas")
--- a/ibis/backends/dask/execution/generic.py
+++ b/ibis/backends/dask/execution/generic.py
@@ -43,6 +43,7 @@ from ibis.backends.pandas.execution.generic import (
     execute_node_nullif_series_scalar,
     execute_node_self_reference_dataframe,
     execute_null_if_zero_series,
+    execute_searched_case,
     execute_series_clip,
     execute_series_isnull,
     execute_series_notnnull,
@@ -55,6 +56,7 @@ from ..core import execute
 from ..dispatch import execute_node
 from .util import (
     TypeRegistrationDict,
+    dask_array_select,
     make_selected_obj,
     register_types_to_dispatcher,
 )
@@ -399,6 +401,30 @@ def wrap_case_result(raw: np.ndarray, expr: ir.ValueExpr):
     if isinstance(expr, ir.ScalarExpr) and result.size.compute() == 1:
         return result.head().item()
     return result
+
+
+@execute_node.register(ops.SearchedCase, list, list, object)
+def execute_searched_case_dask(op, whens, thens, otherwise, **kwargs):
+    if not isinstance(whens[0], dd.Series):
+        # if we are not dealing with dask specific objects, fallback to the
+        # pandas logic. For example, in the case of ibis literals.
+        # See `test_functions/test_ifelse_returning_bool` or
+        # `test_operations/test_searched_case_scalar` for code that hits this.
+        return execute_searched_case(op, whens, thens, otherwise, **kwargs)
+    if otherwise is None:
+        otherwise = np.nan
+    idx = whens[0].index
+    whens = [w.to_dask_array() for w in whens]
+    if isinstance(thens[0], dd.Series):
+        # some computed column
+        thens = [t.to_dask_array() for t in thens]
+    else:
+        # scalar
+        thens = [da.from_array(np.array([t])) for t in thens]
+    # TODO this only exists as of 2021.6.1
+    raw = dask_array_select(whens, thens, otherwise)
+    out = dd.from_dask_array(raw, index=idx,)
+    return out
 
 
 @execute_node.register(ops.SimpleCase, dd.Series, list, list, object)
--- a/ibis/backends/dask/execution/selection.py
+++ b/ibis/backends/dask/execution/selection.py
@@ -25,7 +25,11 @@ from ibis.expr.typing import TimeContext
 from ..core import execute
 from ..dispatch import execute_node
 from ..execution import constants
-from ..execution.util import coerce_to_output, compute_sorted_frame
+from ..execution.util import (
+    add_partitioned_sorted_column,
+    coerce_to_output,
+    compute_sorted_frame,
+)
 
 
 @compute_projection.register(ir.ScalarExpr, ops.Selection, dd.DataFrame)
@@ -137,6 +141,9 @@ def execute_selection_dataframe(
 
     # Build up the individual dask structures from column expressions
     if selections:
+        # Create a unique row identifier and set it as the index. This is used
+        # in dd.concat to merge the pieces back together.
+        data = add_partitioned_sorted_column(data)
         data_pieces = []
         for selection in selections:
             dask_object = compute_projection(
@@ -149,7 +156,8 @@ def execute_selection_dataframe(
             )
             data_pieces.append(dask_object)
 
-        result = dd.concat(data_pieces, axis=1, ignore_unknown_divisions=True)
+        result = dd.concat(data_pieces, axis=1)
+        result.reset_index(drop=True)
 
     if predicates:
         predicates = _compute_predicates(
--- a/ibis/backends/dask/execution/util.py
+++ b/ibis/backends/dask/execution/util.py
@@ -1,7 +1,9 @@
 from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
 
+import dask.array as da
 import dask.dataframe as dd
 import dask.delayed
+import numpy as np
 import pandas as pd
 from dask.dataframe.groupby import SeriesGroupBy
 
@@ -345,3 +347,153 @@ def safe_scalar_type(output_meta):
         output_meta = pd.Timestamp(1, tz=output_meta.tz, unit=output_meta.unit)
 
     return output_meta
+
+
+def add_partitioned_sorted_column(
+    df: Union[dd.DataFrame, dd.Series],
+) -> dd.DataFrame:
+    """Add a column that is already partitioned and sorted
+    This columns acts as if we had a global index across the distributed data.
+    Important properties:
+    - Each row has a unique id (i.e. a value in this column)
+    - IDs within each partition are already sorted
+    - Any id in partition N_{t} is less than any id in partition N_{t+1}
+    We do this by designating a sufficiently large space of integers per
+    partition via a base and adding the existing index to that base. See
+    `helper` below.
+    Though the space per partition is bounded, real world usage should not
+    hit these bounds. We also do not explicity deal with overflow in the
+    bounds.
+
+    Parameters
+    ----------
+    df : dd.DataFrame
+        Dataframe to add the column to
+
+    Returns
+    -------
+    dd.DataFrame
+        New dask dataframe with sorted partitioned index
+
+    Examples
+    --------
+
+    >>> ddf = dd.from_pandas(pd.DataFrame({'a': [1, 2,3, 4]}), npartitions=2)
+    >>> ddf
+    Dask DataFrame Structure:
+                    a
+    npartitions=2
+    0              int64
+    2                ...
+    3                ...
+    Dask Name: from_pandas, 2 task
+    >>> ddf.compute()
+       a
+    0  1
+    1  2
+    2  3
+    3  4
+    >>> ddf = add_partitioned_sorted_column(ddf)
+    >>> ddf
+    Dask DataFrame Structure:
+                    a
+    npartitions=2
+    0              int64
+    4294967296       ...
+    8589934592       ...
+    Dask Name: set_index, 8 tasks
+    Name: result, dtype: int64
+    >>> ddf.compute()
+                a
+    _ibis_index
+    0            1
+    1            2
+    4294967296   3
+    4294967297   4
+    """
+    if isinstance(df, dd.Series):
+        df = df.to_frame()
+
+    col_name = "_ibis_index"
+
+    if col_name in df.columns:
+        raise ValueError(f"Column {col_name} is already present in DataFrame")
+
+    def helper(
+        df: Union[pd.Series, pd.DataFrame],
+        partition_info: Dict[str, Any],  # automatically injected by dask
+        col_name: str,
+    ):
+        """Assigns a column with a unique id for each row"""
+        if len(df) > (2 ** 31):
+            raise ValueError(
+                f"Too many items in partition {partition_info} to add"
+                "partitioned sorted column without overflowing."
+            )
+        base = partition_info["number"] << 32
+        return df.assign(**{col_name: [base + idx for idx in df.index]})
+
+    original_meta = df._meta.dtypes.to_dict()
+    new_meta = {**original_meta, **{col_name: "int64"}}
+    df = df.reset_index(drop=True)
+    df = df.map_partitions(helper, col_name=col_name, meta=new_meta)
+    # Divisions include the minimum value of every partition's index and the
+    # maximum value of the last partition's index
+    divisions = tuple(x << 32 for x in range(df.npartitions + 1))
+
+    df = df.set_index(col_name, sorted=True, divisions=divisions)
+
+    return df
+
+
+def dask_array_select(condlist, choicelist, default=0):
+    # shim taken from dask.array 2021.6.1
+    # https://github.com/ibis-project/ibis/issues/2847
+    try:
+        from dask.array import select
+
+        return select(condlist, choicelist, default)
+    except ImportError:
+
+        def _select(*args, **kwargs):
+            split_at = len(args) // 2
+            condlist = args[:split_at]
+            choicelist = args[split_at:]
+            return np.select(condlist, choicelist, **kwargs)
+
+        if len(condlist) != len(choicelist):
+            raise ValueError(
+                "list of cases must be same length as list of conditions"
+            )
+
+        if len(condlist) == 0:
+            raise ValueError(
+                "select with an empty condition list is not possible"
+            )
+
+        choicelist = [da.asarray(choice) for choice in choicelist]
+
+        try:
+            intermediate_dtype = da.result_type(*choicelist)
+        except TypeError as e:
+            msg = "Choicelist elements do not have a common dtype."
+            raise TypeError(msg) from e
+
+        blockwise_shape = tuple(range(choicelist[0].ndim))
+
+        condargs = [
+            arg for elem in condlist for arg in (elem, blockwise_shape)
+        ]
+        choiceargs = [
+            arg for elem in choicelist for arg in (elem, blockwise_shape)
+        ]
+
+        return da.blockwise(
+            _select,
+            blockwise_shape,
+            *condargs,
+            *choiceargs,
+            dtype=intermediate_dtype,
+            name="select",
+            default=default,
+        )
--- a/ibis/backends/dask/execution/window.py
+++ b/ibis/backends/dask/execution/window.py
@@ -1,0 +1,86 @@
+"""Code for computing window functions in the dask backend."""
+
+from typing import Any, Optional, Union
+
+import dask.dataframe as dd
+
+import ibis.expr.operations as ops
+import ibis.expr.window as win
+from ibis.expr.scope import Scope
+from ibis.expr.typing import TimeContext
+
+from ..core import execute_with_scope
+from ..dispatch import execute_node
+from .util import (
+    _pandas_dtype_from_dd_scalar,
+    _wrap_dd_scalar,
+    add_partitioned_sorted_column,
+    make_meta_series,
+)
+
+
+def _post_process_empty(
+    result: Any,
+    parent: Union[dd.Series, dd.DataFrame],
+    timecontext: Optional[TimeContext],
+) -> dd.Series:
+    """Post process non grouped, non ordered windows.
+
+    dd.Series/dd.DataFrame objects are passed through, otherwise we conform
+    the output to the parent input (i.e. so the shape an partitioning matches).
+
+    dd.core.Scalar needs special handling so downstream functions can work
+    with it.
+    """
+    if isinstance(result, (dd.Series, dd.DataFrame)):
+        return result
+    elif isinstance(result, dd.core.Scalar):
+        # TODO this should be refactored with similar logic in util.py
+        # both solve the generalish problem we have of wrapping a
+        # dd.core.Scalar into something dask can work with downstream
+        # TODO computation
+        lens = parent.index.map_partitions(len).compute().values
+        out_dtype = _pandas_dtype_from_dd_scalar(result)
+        meta = make_meta_series(dtype=out_dtype)
+        delayeds = [_wrap_dd_scalar(result, None, out_len) for out_len in lens]
+        series = dd.from_delayed(delayeds, meta=meta)
+        series = add_partitioned_sorted_column(series)
+        return series[0]
+    else:
+        # Project any non delayed object to the shape of "parent"
+        return parent.apply(lambda row: result, meta=(None, 'object'))
+
+
+@execute_node.register(ops.WindowOp, dd.Series, win.Window)
+def execute_window_op(
+    op,
+    data,
+    window,
+    scope: Scope = None,
+    timecontext: Optional[TimeContext] = None,
+    aggcontext=None,
+    clients=None,
+    **kwargs,
+):
+    # Currently this acts as an "unwrapper" for trivial windows (i.e. those
+    # with no ordering/grouping/preceding/following functionality).
+    if not all(
+        [
+            window.preceding is None,
+            window.following is None,
+            window._group_by == [],
+            window._order_by == [],
+        ]
+    ):
+        raise NotImplementedError(
+            "Window operations are unsuported in the dask backend"
+        )
+    result = execute_with_scope(
+        expr=op.expr,
+        scope=scope,
+        timecontext=timecontext,
+        aggcontext=aggcontext,
+        clients=clients,
+        **kwargs,
+    )
+    return _post_process_empty(result, data, timecontext)
--- a/ibis/backends/dask/tests/test_udf.py
+++ b/ibis/backends/dask/tests/test_udf.py
@@ -259,7 +259,7 @@ def test_udaf_analytic_groupby(con, t, df):
     def f(s):
         return s.sub(s.mean()).div(s.std())
 
-    expected = df.groupby('key').c.transform(f)
+    expected = df.groupby('key').c.transform(f).compute()
     tm.assert_series_equal(result, expected)
 
 
@@ -491,14 +491,11 @@ def test_array_return_type_reduction(con, t, df, qs):
     assert list(result) == expected.tolist()
 
 
-@pytest.mark.xfail(
-    raises=NotImplementedError, reason='TODO - windowing - #2553'
-)
 def test_array_return_type_reduction_window(con, t, df, qs):
     """Tests reduction UDF returning an array, used over a window."""
     expr = quantiles(t.b, quantiles=qs).over(ibis.window())
     result = expr.execute()
-    expected_raw = df.b.quantile(qs).tolist()
+    expected_raw = df.b.quantile(qs).compute().tolist()
     expected = pd.Series([expected_raw] * len(df))
     tm.assert_series_equal(result, expected)
 
--- a/ibis/backends/dask/udf.py
+++ b/ibis/backends/dask/udf.py
@@ -1,21 +1,35 @@
 import itertools
+from typing import List, Tuple
 
 import dask.dataframe as dd
 import dask.dataframe.groupby as ddgb
 import dask.delayed
+import numpy as np
 import pandas
 
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
+import ibis.expr.types as ir
 from ibis.backends.base import Client
 from ibis.backends.pandas.udf import nullable  # noqa
 
 from .dispatch import execute_node, pre_execute
 from .execution.util import (
     assert_identical_grouping_keys,
+    make_meta_series,
     make_selected_obj,
     safe_scalar_type,
 )
+
+
+def make_struct_op_meta(op: ir.Expr) -> List[Tuple[str, np.dtype]]:
+    """Unpacks a dt.Struct into a DataFrame meta"""
+    return list(
+        zip(
+            op._output_type.names,
+            [x.to_dask() for x in op._output_type.types],
+        )
+    )
 
 
 @pre_execute.register(ops.ElementWiseVectorizedUDF)
@@ -60,12 +74,7 @@ def pre_execute_elementwise_udf(op, *clients, scope=None, **kwargs):
         # file.
         # See ibis.udf.vectorized.UserDefinedFunction
         if isinstance(op._output_type, dt.Struct):
-            meta = list(
-                zip(
-                    op._output_type.names,
-                    [x.to_dask() for x in op._output_type.types],
-                )
-            )
+            meta = make_struct_op_meta(op)
 
             df = dd.map_partitions(op.func, *args, meta=meta)
             return df
@@ -116,10 +125,30 @@ def pre_execute_analytic_and_reduction_udf(op, *clients, scope=None, **kwargs):
         # Depending on the type of operation, lazy_result is a Delayed that
         # could become a dd.Series or a dd.core.Scalar
         if isinstance(op, ops.AnalyticVectorizedUDF):
-            meta = pandas.Series(
-                [], name=args[0].name, dtype=op._output_type.to_dask()
-            )
+            if isinstance(op._output_type, dt.Struct):
+                meta = make_struct_op_meta(op)
+            else:
+                meta = make_meta_series(
+                    dtype=op._output_type.to_dask(), name=args[0].name,
+                )
             result = dd.from_delayed(lazy_result, meta=meta)
+
+            if args[0].known_divisions:
+                if not len({a.divisions for a in args}) == 1:
+                    raise ValueError(
+                        "Mixed divisions passed to AnalyticVectorized UDF"
+                    )
+                # result is going to be a single partitioned thing, but we
+                # need it to be able to dd.concat it with other data
+                # downstream. We know that this udf operation did not change
+                # the index. Thus, we know the divisions, allowing dd.concat
+                # to align this piece with the other pieces.
+                original_divisions = args[0].divisions
+                result.divisions = (
+                    original_divisions[0],
+                    original_divisions[-1],
+                )
+                result = result.repartition(divisions=original_divisions)
         else:
             # lazy_result is a dd.core.Scalar from an ungrouped reduction
             if isinstance(op._output_type, (dt.Array, dt.Struct)):
--- a/ibis/backends/dask/execution/__init__.py
+++ b/ibis/backends/dask/execution/__init__.py
@@ -11,3 +11,4 @@ from .selection import *  # noqa: F401,F403
 from .strings import *  # noqa: F401,F403
 from .structs import *  # noqa: F401,F403
 from .temporal import *  # noqa: F401,F403
+from .window import *  # noqa: F401,F403
--- a/ibis/backends/tests/test_vectorized_udf.py
+++ b/ibis/backends/tests/test_vectorized_udf.py
@@ -265,8 +265,7 @@ def test_analytic_udf(backend, alltypes, df):
     backend.assert_series_equal(result, expected, check_names=False)
 
 
-@pytest.mark.only_on_backends(['pandas', 'pyspark'])
-# TODO - windowing - #2553
+@pytest.mark.only_on_backends(['pandas', 'pyspark', 'dask'])
 @pytest.mark.xfail_unsupported
 @pytest.mark.parametrize('udf', calc_zscore_udfs)
 def test_analytic_udf_mutate(backend, alltypes, df, udf):


gerrymanoim:
Some progress here with a new temporary column approach.  

I've now broken things that use `ibis.case()` though, so need to look into that. There's also a place we force computation that I need to think about. However, the ungrouped/unsorted test case runs correctly.

gerrymanoim:
Now fixed `ibis.case` (also removing a case where we forced computation). 

This will fail with versions of `dask<2021.6.1` (when dask implemented `da.select`) so I _think_ I might just steal the implementation at https://github.com/dask/dask/commit/868cba2405676442a7f4fe07ad82b038797ecdc4#diff-be1ec6725c3829aa2f919897e0d4cfc1e00b8db1e70e90c82531c5a12fd727afR1876 for a compat shim, and remove it later once this is the minimum version.

jreback:
> Now fixed `ibis.case` (also removing a case where we forced computation).
> 
> This will fail with versions of `dask<2021.6.1` (when dask implemented `da.select`) so I _think_ I might just steal the implementation at [dask/dask@868cba2#diff-be1ec6725c3829aa2f919897e0d4cfc1e00b8db1e70e90c82531c5a12fd727afR1876](https://github.com/dask/dask/commit/868cba2405676442a7f4fe07ad82b038797ecdc4#diff-be1ec6725c3829aa2f919897e0d4cfc1e00b8db1e70e90c82531c5a12fd727afR1876) for a compat shim, and remove it later once this is the minimum version.

sgtm. pls open an issue to revert this work-around

ibis/backends/dask/__init__.py
```diff
@@ -62,3 +62,5 @@ def register_options(self):
             'See ibis.dask.trace for details.',
             validator=ibis.config.is_bool,
         )
+        # forces the pandas backend to register options
+        getattr(ibis, "pandas")
```
jreback, line 66: you can just `import ibis.backends.pandas` right? (in the dask implementation itself)

jreback: but agree we should handle the options registration in a better way

jreback (changes requested):

ibis/backends/dask/__init__.py
gerrymanoim: There's a tracking issue here: https://github.com/ibis-project/ibis/issues/2844. This was mostly to get my dask tests working locally. Importing would accomplish the same thing, but generally should be fixed at the ibis level.

gerrymanoim:
Test failure in pyspark is some pyarrow compatibility, unrelated to these changes.

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,106 @@ def safe_scalar_type(output_meta):
...
+    - Each row has a unique id (i.e. a value in this column)
+    - IDs within each partition are already sorted
+    - Any id in partition N_{t} is less than any id in partition N_{t+1}
+    We do this by designating a sufficiently large space of integers per
+    partition via a base and adding the existing index to that base. See
+    `helper` below.
+    Though the space per partition is bounded, real world usage should not
+    hit these bounds. We also do not explicity deal with overflow in the
+    bounds.
+    Parameters
```
jreback, line 370: blank line before Parmameters

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,106 @@ def safe_scalar_type(output_meta):
...
+    ----------
+    df : dd.DataFrame
+        Dataframe to add a column to
+    set_index : bool, optional
+        Whether to also set that column as the index, by default False
+    col_name : str, optional
```
jreback, line 376: don't you always make a unique name? (e.g. uuid0?

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,106 @@ def safe_scalar_type(output_meta):
...
+    partition via a base and adding the existing index to that base. See
+    `helper` below.
+    Though the space per partition is bounded, real world usage should not
+    hit these bounds. We also do not explicity deal with overflow in the
+    bounds.
+    Parameters
+    ----------
+    df : dd.DataFrame
+        Dataframe to add a column to
+    set_index : bool, optional
```
jreback, line 374: shouldn't we do this explicity? (e.g. in the calling expression)

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,106 @@ def safe_scalar_type(output_meta):
...
+        The column name, by default "_ibis_index"
+    Returns
+    -------
+    Tuple[dd.DataFrame, List]
+        Tuple of the new dataframe with column and divisions of that column
+    """
+    if isinstance(df, dd.Series):
+        df = df.to_frame()
+
+    def helper(df, partition_info, col_name):
```
jreback, line 386: can you give a comment on what this does (type if you can)

ibis/backends/dask/execution/util.py
jreback: hmm, you need the division, can you ask the df for this?

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,106 @@ def safe_scalar_type(output_meta):
...
+    return (df, divisions)
+
+
+def dask_array_select(condlist, choicelist, default=0):
+    try:
+        from dask.array import select
+
+        return select(condlist, choicelist, default)
+    except ImportError:
+        # shim taken from dask.array 2021.6.1
```
jreback, line 410: move this up right below the function def (the comments)

ibis/backends/dask/execution/window.py
```diff
@@ -0,0 +1,74 @@
...
+
+from ..core import execute_with_scope
+from ..dispatch import execute_node
+from .util import (
+    _pandas_dtype_from_dd_scalar,
+    _wrap_dd_scalar,
+    add_partitioned_sorted_column,
+    make_meta_series,
+)
+
```
jreback, line 18: can you add a doc-string, pretty clear by the name but can't hurt

ibis/backends/dask/execution/window.py
```diff
@@ -0,0 +1,74 @@
...
+
+def _post_process_empty(
+    result: Any,
+    parent: Union[dd.Series, dd.DataFrame],
+    timecontext: Optional[TimeContext],
+) -> dd.Series:
+    if isinstance(result, (dd.Series, dd.DataFrame)):
```
jreback, line 25: doc-string :->

ibis/backends/dask/execution/window.py
```diff
@@ -0,0 +1,74 @@
...
+    op,
+    data,
+    window,
+    scope: Scope = None,
+    timecontext: Optional[TimeContext] = None,
+    aggcontext=None,
+    clients=None,
+    **kwargs,
+):
+    if not all(
```
jreback, line 55: comment that this is bypassing trivially windows

ibis/backends/dask/tests/test_udf.py
```diff
@@ -259,7 +259,7 @@ def test_udaf_analytic_groupby(con, t, df):
     def f(s):
         return s.sub(s.mean()).div(s.std())
 
-    expected = df.groupby('key').c.transform(f)
+    expected = df.groupby('key').c.transform(f).compute()
```
jreback, line 262: what did this do before?

jreback (changes requested):

ibis/backends/dask/tests/test_udf.py
gerrymanoim: Err this was/is still being skipped because the underlying functionality is `NotImplemented`. But the `compute` will need to be there anyway eventually (I think I forgot to fix this when we changed `.execute()` to return a computed object.

ibis/backends/dask/execution/util.py
gerrymanoim: I don't think we want a unique name - it is useful to be able to refer to the column downstream (this will be more important in windows grouped analytic udf code where I'll need to re-index on `_ibis_index` after performing an operation on groups. 

I can raise a `ValueError` if a user calls this more than once (i.e. raise if `col_name in df.columns`).

ibis/backends/dask/execution/util.py
gerrymanoim: This came out of my previous PR where I used this for grouped udfs (https://github.com/ibis-project/ibis/pull/2786/files#diff-cdacb0d0d05c5f6c1c50b7e78592154aacafc240a769fdf83f51c0945ecae4c0R201). There were cases where it was useful to create this column, but not assign it an index, but also keep around divisions for later (you can only get divisions off the df if you perform the `set_index` space). 

However, now we just `set_index` as part of the start of `selection`. This might simplify all the code here? I can refactor this a bit as part of doing the grouped windows (next on my list).

ibis/backends/dask/execution/generic.py
```diff
@@ -43,6 +43,7 @@
     execute_node_nullif_series_scalar,
     execute_node_self_reference_dataframe,
     execute_null_if_zero_series,
+    execute_searched_case,
```
icexelloss, line 46: What is this for?

ibis/backends/dask/execution/generic.py
```diff
@@ -401,6 +403,27 @@ def wrap_case_result(raw: np.ndarray, expr: ir.ValueExpr):
     return result
 
 
+@execute_node.register(ops.SearchedCase, list, list, object)
+def execute_searched_case_dask(op, whens, thens, otherwise, **kwargs):
+    if not isinstance(whens[0], dd.Series):
```
icexelloss, line 408: When would you have the case that this is not a dd.Series?

ibis/backends/dask/execution/generic.py
```diff
@@ -401,6 +403,27 @@ def wrap_case_result(raw: np.ndarray, expr: ir.ValueExpr):
     return result
 
 
+@execute_node.register(ops.SearchedCase, list, list, object)
+def execute_searched_case_dask(op, whens, thens, otherwise, **kwargs):
```
icexelloss, line 407: This looks fine to me but is this related to the no-op windows change? Not super clear to me.

ibis/backends/dask/execution/selection.py
```diff
@@ -137,6 +141,7 @@ def execute_selection_dataframe(
 
     # Build up the individual dask structures from column expressions
     if selections:
+        data, _ = add_partitioned_sorted_column(data, set_index=True)
```
icexelloss, line 144: What is the `_` here and why do we not need it?

icexelloss: Can you explain how this works in the comment? i.e., what uses partition column and what's the effect of that.

ibis/backends/dask/execution/util.py
```diff
@@ -112,6 +114,7 @@ def coerce_to_output(
         ir.DestructScalar,
         ir.StructScalar,
     )
+
```
icexelloss, line 117: Nit: whitespace

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,116 @@ def safe_scalar_type(output_meta):
...
+    set_index: bool = False,
+    col_name: str = "_ibis_index",
+) -> Tuple[dd.DataFrame, List]:
+    """Add a column that is already partitioned and sorted
+    This columns acts as if we had a global index across the distributed data.
+    Important properties:
+    - Each row has a unique id (i.e. a value in this column)
+    - IDs within each partition are already sorted
+    - Any id in partition N_{t} is less than any id in partition N_{t+1}
+    We do this by designating a sufficiently large space of integers per
```
icexelloss, line 364: What is the `sufficiently large space` here?

icexelloss: I kind of like the Spark way of describing this. pretty clear to me what's the max size of records in a partition and what is the max number of partitions and what is the size the the id column.

https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.sql.functions.monotonically_increasing_id.html#pyspark.sql.functions.monotonically_increasing_id

ibis/backends/dask/execution/util.py
icexelloss: My understanding is that the index column is "temporarily". If so, what is the scope of the index column? E.g., when will this be dropped?

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,116 @@ def safe_scalar_type(output_meta):
...
+    bounds.
+
+    Parameters
+    ----------
+    df : dd.DataFrame
+        Dataframe to add a column to
+    set_index : bool, optional
+        Whether to also set that column as the index, by default False
+    col_name : str, optional
+        The column name, by default "_ibis_index"
```
icexelloss, line 378: Maybe I am not understanding this correctly, but how do you communicate the column name from the producer of the index column and the consumer of the index column? (e.g., if the producer creates a index column "a", how would the downstream consumer know that "a" is an index column?)

icexelloss: Maybe `_row_index` is a better name than `_ibis_index`

icexelloss: Or just `_index`

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,116 @@ def safe_scalar_type(output_meta):
...
+        df: Union[dd.Series, dd.DataFrame],
+        partition_info: Dict[str, Any],
+        col_name: str,
+    ):
+        """Assigns a column with a unique id for each row"""
+        base = partition_info["number"] << 32
+        return df.assign(**{col_name: [base + idx for idx in df.index]})
+
+    original_meta = df._meta.dtypes.to_dict()
+    new_meta = {**original_meta, **{col_name: "int"}}
```
icexelloss, line 400: int is int64? (maybe use "long" or "int64") In some systems "int" meant int32

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,116 @@ def safe_scalar_type(output_meta):
...
+    def helper(
+        df: Union[dd.Series, dd.DataFrame],
+        partition_info: Dict[str, Any],
+        col_name: str,
+    ):
+        """Assigns a column with a unique id for each row"""
+        base = partition_info["number"] << 32
+        return df.assign(**{col_name: [base + idx for idx in df.index]})
+
+    original_meta = df._meta.dtypes.to_dict()
```
icexelloss, line 399: This looks slightly confusing to me. Is that anything inside `_meta` other than `dtypes`? Looks like we are dropping everything in the old meta except for the dtypes?

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,116 @@ def safe_scalar_type(output_meta):
...
+    new_meta = {**original_meta, **{col_name: "int"}}
+    df = df.reset_index(drop=True)
+    df = df.map_partitions(helper, col_name=col_name, meta=new_meta)
```
icexelloss, line 402: Do we not need to pass in `partition_info` here? If not when is that passed it?

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +348,116 @@ def safe_scalar_type(output_meta):
...
+    # Divisions include the minimum value of every partition's index and the
+    # maximum value of the last partition's index
+    divisions = tuple(x << 32 for x in range(df.npartitions + 1))
+
+    if set_index:
```
icexelloss, line 407: When do we want to set index vs not?

ibis/backends/dask/execution/util.py
gerrymanoim: We always drop the index as part of `execute_and_reset` (same as the pandas backend) (https://github.com/ibis-project/ibis/blob/b9556084c0b99b2b5dedcb6efb94472ff4da19c8/ibis/backends/dask/core.py#L455) I can drop this earlier as part of the select statement, but that feels unnecessary.

ibis/backends/dask/execution/window.py
```diff
@@ -0,0 +1,86 @@
...
+
+    dd.Series/dd.DataFrame objects are passed through, otherwise we conform
+    the output to the parent input (i.e. so the shape an partitioning matches).
+
+    dd.core.Scalar needs special handling so downstream functions can work
+    with it.
+    """
+    if isinstance(result, (dd.Series, dd.DataFrame)):
+        return result
+    elif isinstance(result, dd.core.Scalar):
```
icexelloss, line 37: Can you explain why do we need to add the index column for any Scalar results?

icexelloss: A quick google search of "Dask core Scalar" doesn't tell me much. Can you please explain this concept?

ibis/backends/dask/execution/generic.py
gerrymanoim: Please see the comments at the top of the PR. 

> This code also fixes #2711 by using da.select. This was only introduced in dask 2021.6.1, so I've ported the code into a shim. #2847 is a follow up issue to remove this.

Basically, working on windows I ran into a situation where it was easier to solve this in the same PR instead of making a work around on a work around.

ibis/backends/dask/execution/window.py
icexelloss: e.g, is it just wrapping a scalar or can this also be a result of an aggregation?

ibis/backends/dask/execution/window.py
```diff
@@ -0,0 +1,86 @@
...
+        lens = parent.index.map_partitions(len).compute().values
+        out_dtype = _pandas_dtype_from_dd_scalar(result)
+        meta = make_meta_series(dtype=out_dtype)
+        delayeds = [_wrap_dd_scalar(result, None, out_len) for out_len in lens]
+        series = dd.from_delayed(delayeds, meta=meta,)
+        series, _ = add_partitioned_sorted_column(series, set_index=True)
+        return series[0]
+
+    else:
+        return parent.apply(lambda row: result, meta=(None, 'object'))
```
icexelloss, line 51: What is this branch here?

icexelloss (changes requested):
Reviewed one around. Left some comments.

ibis/backends/dask/execution/selection.py
```diff
@@ -149,7 +154,7 @@ def execute_selection_dataframe(
             )
             data_pieces.append(dask_object)
 
-        result = dd.concat(data_pieces, axis=1, ignore_unknown_divisions=True)
+        result = dd.concat(data_pieces, axis=1)
```
icexelloss, line 157: I think I am slightly confused because I originally thought the index column will be dropped after concat.

ibis/backends/dask/execution/generic.py
gerrymanoim: https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/tests/execution/test_operations.py#L931

ibis/backends/dask/execution/generic.py
gerrymanoim: https://github.com/ibis-project/ibis/pull/2843#discussion_r668123946

ibis/backends/dask/execution/selection.py
gerrymanoim: I think the docstring for the function covers both of these questions.

ibis/backends/dask/execution/util.py
gerrymanoim: This is all formatted with black

ibis/backends/dask/execution/util.py
gerrymanoim: I can add specific text here (or a user can do the `x << 32` math here, but for all practical purposes it shouldn't matter and you should never need to know/rely on the specifics of "sufficiently large".

ibis/backends/dask/execution/util.py
gerrymanoim: > Maybe I am not understanding this correctly, but how do you communicate the column name from the producer of the index column and the consumer of the index column? 

Convention inside the ibis dask backend. This could have been hardcoded, but it felt easy enough to make it configurable.

> e.g., if the producer creates a index column "a", how would the downstream consumer know that "a" is an index column?

Theoretically we could have made it a global or hardcoded it, but nothing currently relies on this name explicitly. I think I will need to refer to this column explicitly when I do the grouped window udfs, but haven't gotten there yet. 

I have no real view on the name, but given that it is only used internally in ibis inside the dask backend, I'm happy to change it to anything you and Jeff agree on.

ibis/backends/dask/execution/selection.py
gerrymanoim: It is dropped automatically as part of `execute_and_reset` as part of normal ibis evaluation. I can drop it here as well but that feels redundant.

ibis/backends/dask/execution/util.py
gerrymanoim: https://docs.dask.org/en/latest/dataframe-design.html#metadata. Specifically

> A dict of {name: dtype} or an iterable of (name, dtype) specifies a DataFrame.

ibis/backends/dask/execution/util.py
gerrymanoim: Why would you need to pass partition info? `map_partitions` applies a function to each partition, it does not change the partition.

ibis/backends/dask/execution/util.py
gerrymanoim: There are times you might not want to set divisions (i.e. in followup grouped mapper code). See comments here: https://github.com/ibis-project/ibis/pull/2843#discussion_r667019954. 

I think I will end up simplifying this after doing the grouped window code, since we're now always setting this as an index as part of selection. But I'll simplify the interface if necessary there.

ibis/backends/dask/execution/window.py
gerrymanoim: https://github.com/dask/dask/blob/main/dask/dataframe/core.py#L115. It is a core dask dataframe type. 

Some things may result in a scalar `dd.col_name.mean()`.

ibis/backends/dask/execution/window.py
gerrymanoim: If you have a scalar that isn't delayed we can just insert it into every row of the parent.

ibis/backends/dask/execution/util.py
gerrymanoim: Pandas automatically treats this as `int64`. 

```python
>> pd.Series([1], dtype="int").dtype
dtype('int64')
```

ibis/backends/dask/execution/generic.py
icexelloss: Can you add a comment when this is dd.Series vs when it is not? Also for this example:

```
x.case().when(2, x - 1).when(3, x + 1).when(4, x + 2).end()
```

if x is a column, then are we still doing the right thing calling the panda execute?

ibis/backends/dask/execution/generic.py
```diff
@@ -401,6 +403,27 @@ def wrap_case_result(raw: np.ndarray, expr: ir.ValueExpr):
...
+    if not isinstance(whens[0], dd.Series):
+        # let pandas deal with it
+        return execute_searched_case(op, whens, thens, otherwise, **kwargs)
+    if otherwise is None:
+        otherwise = np.nan
+    idx = whens[0].index
+    whens = [w.to_dask_array() for w in whens]
```
icexelloss, line 414: when do we need to convert dd.Series to dask array here?

ibis/backends/dask/execution/selection.py
icexelloss: As a reader this code is not obvious to me. Actually I only understood this after reading through the whole review for a second time. So it would be helpful for future reader explaining what is going on here.

ibis/backends/dask/execution/selection.py
icexelloss: `execute_and_reset` is invoked after each node or after executing the entire tree?

ibis/backends/dask/execution/util.py
icexelloss: I am not sure if black will catch extra line here. Usually I'd avoid touch functions that is unrelated to the core change of the PR.

ibis/backends/dask/execution/util.py
icexelloss: I also feel similar to Jeff that this API looks a bit weird to me that the division of returned separately, and not clear to me when you would want to add an index column but use the division sometime later (and how would you pass the division information to some other node)

icexelloss: Is there a case now you actually want not to set the index and just return the division? (or future cases?) If so, it would be helpful to comment on those cases to help readers like myself.

ibis/backends/dask/execution/util.py
icexelloss: I think there is a major difference here, when dropped inside `execute_and_reset`, it is done only after the whole tree is executed, so the scope of the index is larger and we need to think more about implications of this column through out the entire tree. Whereas if we drop the column after selection, then the scope is much smaller and easier to reason about.

ibis/backends/dask/execution/util.py
icexelloss: > I think I will need to refer to this column explicitly when I do the grouped window udfs, but haven't gotten there yet.
It's a little hard for me to understand what you plan to do with different names with grouped window udfs so commenting sth like

"the names are technically configurable but not used yet. It might be used in the future so leaving it configurable for now."

So a bit more clear to reviewer of this PR

ibis/backends/dask/execution/util.py
icexelloss: Right, pandas does that but for the reader of ibis, there are many things inside ibis that doesn't all treat "int" as int64, so wouldn't hurt to make that clear :)

ibis/backends/dask/execution/util.py
icexelloss:
```
    def helper(
        df: Union[dd.Series, dd.DataFrame],
        partition_info: Dict[str, Any],
        col_name: str,
    ):
```
Oh the `partition_info` is passed by Dask automatically? It seems just appear magically so would be good to leave a comment here so the reader like me don't search for the variable `partition_info` and got confused XD

ibis/backends/dask/execution/util.py
icexelloss: Ah OK. Sorry it's a bit hard for me to get all context of what you plan to do in follow-up when reviewing this code.

ibis/backends/dask/execution/window.py
icexelloss: I see, it's Scalar value as a result of a distributed computation.

icexelloss: Hmm..reading this more, why do we want to add the index for a scalar value? Seems unnecessarily but maybe I am missing sth..

ibis/backends/dask/execution/window.py
icexelloss: I see, can you comment here? (Better if you can validate/assert)

ibis/backends/dask/execution/window.py
gerrymanoim: We project a scalar value to the full index, add the same partitioning scheme as the "parent" and return a `dd.Series` that can be concatted with all the other `selections`.

ibis/backends/dask/execution/util.py
gerrymanoim: Oh yes - sorry that's automatically injected by dask. I'll add a column.

icexelloss (changes requested):
Reviewed another around, Looks pretty good. 

The main point I have right now is that I think we should limit the scope of the index column and drop it right after concat. 

The reason that I think this is better is because passing around an "motorically increasing index" but not maintaining it in every op is dangerous. E.g., if user did some operations that breaks the index (e.g., sort), there is currently no way that we can guarantee the " motorically increasing index invariance" across the full life time of an execution.

Also commented https://github.com/ibis-project/ibis/pull/2843/files#r668344576 but the same thing. (Sorry for the duplicate comments)

ibis/backends/dask/execution/util.py
icexelloss: To elaborate a bit more, the reason that I think this is better is because passing around an "motorically increasing index" but not maintaining it in every op is dangerous. E.g., if user did some operations that breaks the index (e.g., sort), there is currently no way that we can guarantee the " motorically increasing index invariance" across the full life time of an execution.

ibis/backends/dask/execution/util.py
icexelloss: Ah no worries. Thanks! (I will probably get used to this reviewing more Dask code, but still helpful I think)

ibis/backends/dask/execution/util.py
icexelloss: This is fine leaving unchanged.

My point is mainly that we never know how user would break our code, e.g., if I managed to put 2^33 rows inside one partition this would break :(

ibis/backends/dask/execution/window.py
gerrymanoim: I'll add a comment. 

Validate which part?

icexelloss:
Another main point, while I am fine with combine `case when` change with the index change since it's done already. In the future I'd suggest avoid combining two unrelated changes together because it makes it harder for reviewer to separate what change is related to what.

ibis/backends/dask/execution/util.py
gerrymanoim: I've now changed `add_partitioned_sorted_column`:

1. Always `set_index`
2. Removed the ability to pass in the name of the column
3. Do not pass back divisions
4. Drop this index after selection is done. 

This simplifies the current state of this. I'll re-add if I need these features later.

ibis/backends/dask/execution/selection.py
gerrymanoim: Apologies - added clarifying comments that hopefully make this easier to follow.

ibis/backends/dask/execution/util.py
gerrymanoim: Apologies for the confusion - I've just dropped this configuration for now.

ibis/backends/dask/execution/window.py
icexelloss: Sorry, I was thinking to validate the object falling into the else branch is just a python scalar but I realized there probably isn't a very easy way to do that. So I am fine with just a comment XD

gerrymanoim:
> Another main point, while I am fine with combine case when change with the index change since it's done already. In the future I'd suggest avoid combining two unrelated changes together because it makes it harder for reviewer to separate what change is related to what.

I get that. This was an interesting case where debugging things that broke once I added windows also forced me to fix this case. For the future is it easier to:

1. Combine it in one PR?
2. xfail tests in this PR and simultaneously open a follow up PR?

ibis/backends/dask/execution/util.py
gerrymanoim: > On the side, perhaps we should validate inside each partition that we don't overflow.

I think the only way to do that would be to actually compute the lengths in each partition?

ibis/backends/dask/execution/window.py
gerrymanoim: > else branch is just a python scalar

Importantly this might not be a python scalar, it could be a numpy.ndarry. But we know at this point in the branching that we can use `apply` since the result isn't a dask object.

ibis/backends/dask/execution/generic.py
gerrymanoim: Sorry - ignore my original comment. 

```python
expr = ibis.case().when(True, 1).when(False, 2).end()
```
That is the code that will go through the pandas branch (I've added a pointer to this in the code). 

For your case, that actually gets handled by a different function (`execute_simple_case_series`). For both pandas and dask the following will return `nan`s:

```
    x = t.float64_with_zeros
    expr = x.case().when(2, x - 1).when(3, x + 1).when(4, x + 2).end()
```

gerrymanoim:
I've rebased against master to get rid of the spark ci issues, but we seem to have some mamba issues. 

I think I've addressed all the comments here, but let me know if I've missed something.

ibis/backends/dask/execution/util.py
icexelloss:
```
    def helper(
        df: Union[dd.Series, dd.DataFrame],
        partition_info: Dict[str, Any],  # automatically injected by dask
        col_name: str,
    ):
```

I think inside each of these we already have the data for each partition so we can check if that overflows?

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +347,111 @@ def safe_scalar_type(output_meta):
...
+    if isinstance(df, dd.Series):
+        df = df.to_frame()
+
+    col_name = "_ibis_index"
+
+    if col_name in df.columns:
+        raise ValueError(f"Column {col_name} is already present in DataFrame")
+
+    def helper(
+        df: Union[dd.Series, dd.DataFrame],
```
icexelloss, line 386: Is the df annotations correct here?

ibis/backends/dask/execution/window.py
icexelloss: I see. Fine to leave as well. It seems that the objects are can be anything that is not distributed Dask object

ibis/backends/dask/execution/selection.py
```diff
@@ -137,6 +141,9 @@ def execute_selection_dataframe(
 
     # Build up the individual dask structures from column expressions
     if selections:
+        # Create a unique row identifier and set it as the index. This is used
+        # in dd.concat to merge the pieces back together.
+        data = add_partitioned_sorted_column(data)
         data_pieces = []
         for selection in selections:
             dask_object = compute_projection(
```
icexelloss, line 149: I think I missed some points here in previous reviews:

1. Could `compute_project` changes the index? i.e, does the index on `dask_object` is always the same?
2. Where is the original columns in `data`? it seems it is not part of `data_pieces` so not clear to me how did we keep the original columns

ibis/backends/dask/udf.py
```diff
@@ -116,10 +125,25 @@ def lazy_agg(*series: pandas.Series):
...
+            if isinstance(op._output_type, dt.Struct):
+                meta = make_struct_op_meta(op)
+            else:
+                meta = make_meta_series(
+                    dtype=op._output_type.to_dask(), name=args[0].name,
+                )
             result = dd.from_delayed(lazy_result, meta=meta)
+
+            if args[0].known_divisions:
+                # result is going to be a single partitioned thing,
```
icexelloss, line 137: Does  AnalyticVectorizedUDF with groupby work? I don't see the `original_divisions` data being used there..

icexelloss (commented):
Reviewed again. Looks pretty good. Left some comments / clarifying questions.

icexelloss:
@gerrymanoim Sorry I might have missed some code in previous reviews so adding a couple of comments to get a better understanding of this.

ibis/backends/dask/udf.py
```diff
@@ -116,10 +125,25 @@ def lazy_agg(*series: pandas.Series):
...
+                # but we need it to match incoming data so we can dd.concat
+                # downstream. However we know that this udf did not change
+                # the index.
+                original_divisions = args[0].divisions
+                result.divisions = (
```
icexelloss, line 142: I am not sure I understand this correctly. Looks like the result doesn't match input partitions, how does dd.concat work in that case?

ibis/backends/dask/udf.py
```diff
@@ -116,10 +125,25 @@ def lazy_agg(*series: pandas.Series):
...
+                    dtype=op._output_type.to_dask(), name=args[0].name,
+                )
             result = dd.from_delayed(lazy_result, meta=meta)
+
+            if args[0].known_divisions:
+                # result is going to be a single partitioned thing,
+                # but we need it to match incoming data so we can dd.concat
+                # downstream. However we know that this udf did not change
+                # the index.
+                original_divisions = args[0].divisions
```
icexelloss, line 141: Probably good idea to validate all divisions are the same for all args, other wise I don't know what the behavior would be.

icexelloss: (More of a defensive check)

ibis/backends/dask/execution/selection.py
gerrymanoim: > 1. Could compute_project changes the index? i.e, does the index on dask_object is always the same?

Currently there is no operation that changes the index. Grouped window will, but we will take care of alignment in the udf. 

> 2. Where is the original columns in data? it seems it is not part of data_pieces so not clear to me how did we keep the original columns

I don't really understand the question here. Can you elaborate?

ibis/backends/dask/execution/util.py
gerrymanoim: Ah fair - I can throw an error if the len here is larger than the allowable number. This will only trigger at compute time.

ibis/backends/dask/udf.py
gerrymanoim: This is not the grouped handler. Regular analyticvectorizedUDFs with groups work. Making them work with windows will require some rewriting (follow up PR).

ibis/backends/dask/udf.py
gerrymanoim: It ends up aligning them and joining https://github.com/dask/dask/blob/main/dask/dataframe/multi.py#L973. 

The case you can't have in `dd.concat` is a mix of known and unknown divisions.

ibis/backends/dask/udf.py
icexelloss: > Regular analyticvectorizedUDFs with groups work

How does this work before this PR? I'd imagine you would need to align on the new index column here so I feel missed sth.

ibis/backends/dask/execution/selection.py
icexelloss: > Currently there is no operation that changes the index. Grouped window will, but we will take care of alignment in the udf.

Got it. If that is an invariant of the code here (compute_projection does not change index), I think it's probably a good idea to clearly document that in the code here. (Better if we can assert in compute_projection that this invariant is True, can be follow up).

ibis/backends/dask/execution/util.py
icexelloss: Computation time is fine - better than nothing XD

ibis/backends/dask/execution/window.py
icexelloss: KK. I think it might be worthy while add some comments here like

"
Eagerly compute the aggregation result, broadcast to match the parent shape and partitioning, finally adding index column so it can be concated back
"

"

ibis/backends/dask/udf.py
icexelloss: kk might be worthwhile adding some comments but it might not be obvious to reader that is not familiar with Dask behavior of concat when partitioning doesn't exactly align.

ibis/backends/dask/execution/selection.py
icexelloss: > I don't really understand the question here. Can you elaborate?

nvm. I figured it out...

I think what's confusing is that selection[0] (the parent table) is not a Projection but is being passed to the function called `compute_projection`, which confused me.  Might be good to add some comments here for future readers.

gerrymanoim: `selection` is an expression, that gets passed, along with `data` (the parent table), to `compute_projection` which is a dispatcher that is registered against function performing the action: https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/execution/selection.py#L31

ibis/backends/dask/udf.py
gerrymanoim: > Regular analyticvectorizedUDFs with groups work

Whoops - sorry in this codebase they're all xskipped because all the tests involve windowing. I think I was thinking of tests in another place. 

Currently reduction udfs work because we do a `safe_concat` that joins all the pieces manually. I think I can rip that out in the future. 

https://github.com/ibis-project/ibis/pull/2786/files#diff-cdacb0d0d05c5f6c1c50b7e78592154aacafc240a769fdf83f51c0945ecae4c0R181 (my previous window PoC did spend a lot of effort aligning pieces of grouped analytic functions. I think this will be a lot easier now.

icexelloss: Kk I see. Thanks for clarifying.

ibis/backends/dask/execution/selection.py
icexelloss: Thanks for adding this!

ibis/backends/dask/execution/generic.py
icexelloss: kk

ibis/backends/dask/execution/selection.py
icexelloss: I see. Consider my point (2) addressed

icexelloss (approved):
LGTM. +1

gerrymanoim:
Thanks for the detailed review, sorry it got reworked in a few pieces. I think the follow-up should be a bit cleaner now.

ibis/backends/dask/execution/selection.py
jreback: doesn't dask support `ignore_index=True` ?

ibis/backends/dask/execution/util.py
```diff
@@ -345,3 +347,116 @@ def safe_scalar_type(output_meta):
...
+    `helper` below.
+    Though the space per partition is bounded, real world usage should not
+    hit these bounds. We also do not explicity deal with overflow in the
+    bounds.
+
+    Parameters
+    ----------
+    df : dd.DataFrame
+        Dataframe to add a column to
+    Returns
```
jreback, line 372: blank line before each section.

*maybe* useful to add an Examples

ibis/backends/dask/udf.py
```diff
@@ -116,10 +125,30 @@ def lazy_agg(*series: pandas.Series):
...
+            else:
+                meta = make_meta_series(
+                    dtype=op._output_type.to_dask(), name=args[0].name,
+                )
             result = dd.from_delayed(lazy_result, meta=meta)
+
+            if args[0].known_divisions:
+                if not len({a.divisions for a in args}) == 1:
+                    raise ValueError(
+                        "Mixed divisions passed to AnalyticVectorized UDF"
```
jreback, line 139: hit in tests?

jreback (changes requested):
a couple of quetsions

ibis/backends/dask/udf.py
gerrymanoim: No - I'm honestly not really sure how you'd hit this or cause this to occur. Generally any args here are coming from the same dataframe, so you'd need to somehow repartition before hitting this function. 

I can write a very contrived test that would just import this directly?

jreback (approved):

jreback:
thanks @gerrymanoim very nice

PR #2848, opened 2021-07-08
CLN: Simplify SQLAlchemy client

datapythonista:
In preparation to move `Client` into `Backend`, some clean up of the SQLAlchemy backends:
- Moving repeated methods like `list_tables` or `table` to the main class, since those are common across SQLAlchemy backends
- Simplifying handling of `sqlalchemy.inspect()`. The function is immediate (and rarely called I think), and the code is overcomplicated to possibly save some micoseconds
- Remove `Client.client` property, which doesn't seem very useful


datapythonista:
Superseded by #2859, 2861, #2862

Issue #2845, opened 2021-07-08
Union except and intersect


saschahofmann:
A lot (all?) backends support some kind of except/intersect UNION 

E.g postgres https://www.postgresqltutorial.com/postgresql-union/
or bigquery https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#intersect

Does it make sense to implement these in the main repo? Do you have a feature freeze until the 2.0 release?

I could have a look at this myself if you agree this should go into ibis directly!

datapythonista:
Maybe not in the docs, but UNION is implemented in Ibis already.

saschahofmann:
Ahhh I see you implemented intersect and Except separately.! Yeah just discovered it in the [code](https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/compiler/query_builder.py#L493)?

Sorry! I think this might be broken in Bigquery then but will test and write there separately

datapythonista:
Bigquery needs to be adapted to the last changes in Ibis, but other than that, that part should be working.

Should we close this issue, or is there anything else that is missing from what you requested here?

saschahofmann:
Yes sorry closing this. Needs to be addressed in BigQuery https://github.com/ibis-project/ibis-bigquery/issues/87

saschahofmann:
Hey sorry,
I was just playing around with `difference` in bigquery and postgres. I can do things like this

```python
table.difference(table2)
```
But if I try to create a projection like this
```python
table.difference(table2).projection(['column1', 'column2'])
```

I get a `NotImplementedError: <class 'ibis.expr.operations.Difference'>`

Originating from here https://github.com/ibis-project/ibis/blob/1.4.0/ibis/backends/base_sqlalchemy/compiler.py#L401

It works for Union any ideas how I can work around this e.g. with what if I want to run a filter on a difference query, @datapythonista ?

saschahofmann:
*Note* this is for ibis 1.4.0. since main looks very different this could be fixed there alread?

datapythonista:
Not so familiar with this part, from the code looks like it should be implemented, not sure what is missing. I don't think there has been any change in that part recently, other than moving code from one place to another. Happy to take PRs for this, before or after the release.

saschahofmann:
Ok I'll have a look

saschahofmann:
Will probably create a PR for this tomorrow just for completeness. I am monkey-patching 1.4.0 like this in our project right now

```python
[...]

PR #2849, opened 2021-07-09
Fix: add collect methods for Difference and Intersection

saschahofmann:
Adds the collect functions for Difference and Intersection

ref #2845


saschahofmann:
@datapythonista I think the failures are not related to this PR.

Ready to merge IMO

jreback (changes requested):
what tests fail before this change and now work?

saschahofmann:
This fixes untested behaviour https://github.com/ibis-project/ibis/issues/2845#issuecomment-876627886

TLDR:
When you try to "select"=project from a difference or intersection it returns an error because the methods added here aren't defined.

jreback:
ok can u add a test to that effect

saschahofmann:
Sure any suggestions where I should put that?

jreback:
see how collect_Union is tested and mimic that, we test pretty generically for sql backends

saschahofmann:
Ok I added a test to _test_compiler.py_. I think that should do it?

ibis/tests/sql/test_compiler.py
```diff
@@ -2298,6 +2298,41 @@ def test_table_difference(self):
...
+        expected = """SELECT `key`
+FROM (
+  SELECT `string_col` AS `key`, CAST(`float_col` AS double) AS `value`
+  FROM functional_alltypes
+  WHERE `int_col` > 0
+  INTERSECT
+  SELECT `string_col` AS `key`, `double_col` AS `value`
+  FROM functional_alltypes
+  WHERE `int_col` <= 0
+) t0"""
```
datapythonista, line 2318: What do you think about creating a constant with this, with the `INTERSECT` / `EXCEPT` as variable, and in the tests just use `QUERY.format('INTERSECT')`? Besides avoiding the duplication, I think it should make tests more readable.

saschahofmann: done. Good idea much nicer now

ibis/tests/sql/test_compiler.py
```diff
@@ -2221,6 +2221,27 @@ def test_endswith(self):
...
+SELECT_QUERY = """SELECT `key`
+FROM (
+  SELECT `string_col` AS `key`, CAST(`float_col` AS double) AS `value`
+  FROM functional_alltypes
+  WHERE `int_col` > 0
+  {}
+  SELECT `string_col` AS `key`, `double_col` AS `value`
+  FROM functional_alltypes
+  WHERE `int_col` <= 0
+) t0"""
```
datapythonista, line 2242: Still a significant amount of duplication, wouldn't make sense to use inside the test something like this or similar instead?
```
SELECT `key` FROM ({QUERY.format('INTERSECT')}) t0
```

saschahofmann: It's either that or something like this

```python
QUERY = """\
SELECT `string_col` AS `key`, CAST(`float_col` AS double) AS `value`
FROM functional_alltypes
WHERE `int_col` > 0
{}
SELECT `string_col` AS `key`, `double_col` AS `value`
FROM functional_alltypes
WHERE `int_col` <= 0"""

SELECT_QUERY = f"SELECT `key` FROM ({QUERY}) t0"
```

I would just need to some add new the two spaces in front of each new line `QUERY`. Maybe your suggestions is nicer?

datapythonista: Ah, true. i think it's still worth, you can use `textwrap.indent`. But up to you, no big deal.

saschahofmann: Haha should have known there is a package for that. Done!

docs/source/release/index.rst
```diff
@@ -11,7 +11,8 @@ Release Notes
 
    These release notes are for versions of ibis **1.0 and later**. Release
    notes for pre-1.0 versions of ibis can be found at :doc:`release-pre-1.0`
-
+   
+* :bug:`2845` Fix projection on differences and intersections
```
jreback, line 15: can you say for the SQL backends.

jreback (changes requested):
looks good, small comment

docs/source/release/index.rst
saschahofmann: sure thing done

jreback (approved):

jreback:
thanks @saschahofmann

Issue #2245, opened 2020-06-11
ENH: Dask backend


zbrookle:
Are there any plan to create a dask backend? I imagine the implementation would be near identical to the pandas one that already exists

jreback:
would surely take this, and agree its pretty darn straightforward to do.

goodwanghan:
Hey, can you also add the support for Ray (modin?) and Rapids AI? I think they are all similar

acivitillo:
One difference between pandas and dask is that in dask you can build a computational graph programmatically. We use this pattern to create an extract graph to run ETL jobs from various backends (Redshift, Salesforce, etc.). We define the columns and other attributes in a .json and then pass that to an Extract() to generate and submit the graph to a cluster. Since we are planning to support various enterprise backends, we use Arrow to store data in memory before sending it to Parquet over S3.

cpcloud:
This exists.

PR #2858, opened 2021-07-14
ENH: Grouped unbounded windows in the dask backend 

gerrymanoim:
Part two of windows in the dask backend, built on top op noop windows in https://github.com/ibis-project/ibis/pull/2843. 

## Notes:

- `execute_grouped_window_op` directly follows the pandas logic in https://github.com/ibis-project/ibis/blob/master/ibis/backends/pandas/execution/window.py#L239. It felt better to section off all the grouping specific logic while windows are only partially supported by the dask backend, though we can refactor this later. 
- I've split up the handlers for grouped analytic and grouped reduction function application, the logic was getting a bit messy to put into one function like pandas does. 
- Using the udf directly, as in https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/tests/test_udf.py#L253 meant I needed to handle some of the "cleanup" actions inside the UDF. I dropped the names check as I couldn't think of a good way to get that working without "cheating" (i.e. making the name the single key). 
- As a follow up to this, I'd like to start doing some larger refactors in the dask and dask/pandas backends:
  - [ ] Cleaning up the usage `coerce_to_output`. I think a lot of the code paths were emergent and can now be cleaned up (for example struct output handling). 
  - [ ] Combining a lot of the pandas + dask udf testing
  - [ ] Combining a lot of the pandas + dask window testing.

### DIFF
--- a/ibis/backends/dask/execution/window.py
+++ b/ibis/backends/dask/execution/window.py
@@ -1,5 +1,6 @@
 """Code for computing window functions in the dask backend."""
 
+import operator
 from typing import Any, Optional, Union
 
 import dask.dataframe as dd
@@ -9,7 +10,7 @@ import ibis.expr.window as win
 from ibis.expr.scope import Scope
 from ibis.expr.typing import TimeContext
 
-from ..core import execute_with_scope
+from ..core import execute, execute_with_scope
 from ..dispatch import execute_node
 from .util import (
     _pandas_dtype_from_dd_scalar,
@@ -68,13 +69,27 @@ def execute_window_op(
         [
             window.preceding is None,
             window.following is None,
-            window._group_by == [],
             window._order_by == [],
         ]
     ):
         raise NotImplementedError(
             "Window operations are unsuported in the dask backend"
         )
+
+    if window._group_by:
+        # there's lots of complicated logic that only applies to grouped
+        # windows
+        return execute_grouped_window_op(
+            op,
+            data,
+            window,
+            scope,
+            timecontext,
+            aggcontext,
+            clients,
+            **kwargs,
+        )
+
     result = execute_with_scope(
         expr=op.expr,
         scope=scope,
@@ -84,3 +99,55 @@ def execute_window_op(
         **kwargs,
     )
     return _post_process_empty(result, data, timecontext)
+
+
+def execute_grouped_window_op(
+    op, data, window, scope, timecontext, aggcontext, clients, **kwargs,
+):
+    # extract the parent
+    (root,) = op.root_tables()
+    root_expr = root.to_expr()
+
+    root_data = execute(
+        root_expr,
+        scope=scope,
+        timecontext=timecontext,
+        clients=clients,
+        aggcontext=aggcontext,
+        **kwargs,
+    )
+
+    group_by = window._group_by
+    grouping_keys = [
+        key_op.name for key_op in map(operator.methodcaller('op'), group_by)
+    ]
+
+    grouped_root_data = root_data.groupby(grouping_keys)
+    scope = scope.merge_scopes(
+        [
+            Scope({t: grouped_root_data}, timecontext)
+            for t in op.expr.op().root_tables()
+        ],
+        overwrite=True,
+    )
+
+    result = execute_with_scope(
+        expr=op.expr,
+        scope=scope,
+        timecontext=timecontext,
+        aggcontext=aggcontext,
+        clients=clients,
+        **kwargs,
+    )
+    # If the grouped operation we performed is not an analytic UDF we have to
+    # realign the output to the input.
+    if not isinstance(op.expr._arg, ops.AnalyticVectorizedUDF):
+        result = dd.merge(
+            root_data[result.index.name].to_frame(),
+            result.to_frame(),
+            left_on=result.index.name,
+            right_index=True,
+        )[result.name]
+        result.divisions = root_data.divisions
+
+    return result
--- a/ibis/backends/dask/tests/test_udf.py
+++ b/ibis/backends/dask/tests/test_udf.py
@@ -246,9 +246,6 @@ def test_udaf_analytic(con, t, df):
     tm.assert_series_equal(result, expected)
 
 
-@pytest.mark.xfail(
-    raises=NotImplementedError, reason='TODO - windowing - #2553'
-)
 def test_udaf_analytic_groupby(con, t, df):
     expr = zscore(t.c).over(ibis.window(group_by=t.key))
 
@@ -260,7 +257,10 @@ def test_udaf_analytic_groupby(con, t, df):
         return s.sub(s.mean()).div(s.std())
 
     expected = df.groupby('key').c.transform(f).compute()
-    tm.assert_series_equal(result, expected)
+    # We don't check names here because the udf is used "directly".
+    # We could potentially special case this and set the name directly
+    # if the udf is only being run on one column.
+    tm.assert_series_equal(result, expected, check_names=False)
 
 
 def test_udaf_groupby(t2, df2):
--- a/ibis/backends/dask/udf.py
+++ b/ibis/backends/dask/udf.py
@@ -165,9 +165,10 @@ def pre_execute_analytic_and_reduction_udf(op, *clients, scope=None, **kwargs):
         return result
 
     @execute_node.register(
-        type(op), *(itertools.repeat(ddgb.SeriesGroupBy, nargs))
+        ops.ReductionVectorizedUDF,
+        *(itertools.repeat(ddgb.SeriesGroupBy, nargs)),
     )
-    def execute_udaf_node_groupby(op, *args, aggcontext, **kwargs):
+    def execute_reduction_node_groupby(op, *args, aggcontext, **kwargs):
         # To apply a udf func to a list of grouped series we:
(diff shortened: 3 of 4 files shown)


ibis/backends/dask/execution/window.py
```diff
@@ -84,3 +99,67 @@ def execute_window_op(
...
+
+
+def execute_grouped_window_op(
+    op, data, window, scope, timecontext, aggcontext, clients, **kwargs,
+):
+    # extract the parent
+    (root,) = op.root_tables()
+    root_expr = root.to_expr()
+
+    data = execute(
```
icexelloss, line 111: Why re-executing data?

ibis/backends/dask/execution/window.py
```diff
@@ -84,3 +99,67 @@ def execute_window_op(
...
+        clients=clients,
+        aggcontext=aggcontext,
+        **kwargs,
+    )
+
+    group_by = window._group_by
+    grouping_keys = [
+        key_op.name
+        if isinstance(key_op, ops.TableColumn)
+        else execute(
```
icexelloss, line 124: What is the type of grouping_keys? Looks like it sometimes is list of string and sometimes is list of dd.Series?

ibis/backends/dask/udf.py
```diff
@@ -204,4 +206,70 @@ def apply_wrapper(df, apply_func, col_names):
...
+                func,
+                col_names,
+                meta=pandas.Series(
+                    meta_value, index=meta_index, dtype=out_type
+                ),
+            )
+            # If you use the UDF directly (as in `test_udaf_analytic_groupby`)
+            # we need to do some renaming/cleanup to get the result to
+            # conform to what the pandas output would look like. Nomrally this
+            # is handled in `util.coerce_to_output`.
```
icexelloss, line 264: This is for follow up? (To move this to `coerce_to_output`) ?

ibis/backends/dask/execution/window.py
```diff
@@ -84,3 +99,67 @@ def execute_window_op(
...
+        scope=scope,
+        timecontext=timecontext,
+        aggcontext=aggcontext,
+        clients=clients,
+        **kwargs,
+    )
+    # If the grouped operation we performed is not an analytic UDF we have to
+    # realign the output to the input.
+    if not isinstance(op.expr._arg, ops.AnalyticVectorizedUDF):
+        result = dd.merge(
```
icexelloss, line 157: This feels a bit suboptimal because when you have many udfs inside (people often compute many window summarization for different columns at once), you would ended up doing many merging. 

Not an issue for this PR but I suggest leaving a note here to look into better ways of doing this in the future.

ibis/backends/dask/udf.py
```diff
@@ -165,9 +165,10 @@ def lazy_agg(*series: pandas.Series):
         return result
 
     @execute_node.register(
-        type(op), *(itertools.repeat(ddgb.SeriesGroupBy, nargs))
+        ops.ReductionVectorizedUDF,
```
icexelloss, line 168: I don't see tests changes around reductions. Are these changes covered by tests?

ibis/backends/dask/udf.py
```diff
@@ -165,9 +165,10 @@ def lazy_agg(*series: pandas.Series):
...
+        *(itertools.repeat(ddgb.SeriesGroupBy, nargs)),
     )
-    def execute_udaf_node_groupby(op, *args, aggcontext, **kwargs):
+    def execute_udrf_node_groupby(op, *args, aggcontext, **kwargs):
```
icexelloss, line 171: I think udaf is still the convention to call this (User defined aggregation functions), which is a common term in sql land.

ibis/backends/dask/execution/window.py
gerrymanoim: `data` actually becomes the parent df, because we're executing on the root_expr (basically the logic in https://github.com/ibis-project/ibis/blob/master/ibis/backends/pandas/execution/window.py#L275-L285). 

Before execute:
```
>>> data
Dask Series Structure:
npartitions=2
0             float64
4294967296        ...
8589934592        ...
Name: double_col, dtype: float64
```
after execute
```
>>> data
Dask DataFrame Structure:
               index Unnamed: 0     id bool_col tinyint_col smallint_col int_col bigint_col float_col double_col date_string_col string_col   timestamp_col   year  month
npartitions=2                                                                                                                                                            
0              int64      int64  int64     bool       int64        int64   int64      int64   float64    float64          object     object  datetime64[ns]  int64  int64
4294967296       ...        ...    ...      ...         ...          ...     ...        ...       ...        ...             ...        ...             ...    ...    ...
8589934592       ...        ...    ...      ...         ...          ...     ...        ...       ...        ...             ...        ...             ...    ...    ...
```

ibis/backends/dask/execution/window.py
gerrymanoim: They're strings. 

But I actually can't find a scenario in tests that hits the else branch here. I stole this pretty directly from https://github.com/ibis-project/ibis/blob/f6910dc666fb6c950e2b386d6b9b22554adcd0c0/ibis/backends/pandas/execution/window.py#L300, so maybe this logic is outdated?

ibis/backends/dask/udf.py
gerrymanoim: Yeah - I think either more or less functionality needs to be in coerce to output. I'd like to do a followup to clean up the usage post this commit.

ibis/backends/dask/udf.py
gerrymanoim: None of the reduction codepaths changed, I just split the grouped analytic one to a new handler.

ibis/backends/dask/udf.py
gerrymanoim: What would be a better term? I need to split out the handler for `ops.ReductionVectorizedUDF` and `ops.AnalyticVectorizedUDF`. I can switch it to `execute_reduction_node_groupby` and `execute_analytic_node_groupby`?

ibis/backends/dask/execution/window.py
gerrymanoim: Created https://github.com/ibis-project/ibis/issues/2880 as a followup

ibis/backends/dask/tests/test_udf.py
```diff
@@ -260,7 +257,7 @@ def f(s):
         return s.sub(s.mean()).div(s.std())
 
     expected = df.groupby('key').c.transform(f).compute()
-    tm.assert_series_equal(result, expected)
+    tm.assert_series_equal(result, expected, check_names=False)
```
jreback, line 260: these don't preserve?

jreback (changes requested):
small comment. does this need a release note?

gerrymanoim:
> small comment. does this need a release note?

All of the dask backend probably needs a release note. I don't know if "partial window support" is worth a release note as lots of these windows are generated under the hood.

ibis/backends/dask/tests/test_udf.py
gerrymanoim: Not when you're using the udf on the dask backend directly like this. 

I could cheat and say "if we're running the udf on a single column then set the output name to that column"

jreback:
> > small comment. does this need a release note?
> 
> All of the dask backend probably needs a release note. I don't know if "partial window support" is worth a release note as lots of these windows are generated under the hood. But maybe we're "feature complete" enough to call this "beta" and add a release note for the backend as a whole?

yep i think we have an issue about this, should prob do this soonish (and can just list the PRs / issues that are related)

ibis/backends/dask/tests/test_udf.py
jreback: kk (just maybe a comment about this as its slightly unusual to not match like this)

gerrymanoim:
We could close https://github.com/ibis-project/ibis/issues/2245 and cross link a bunch of things into there, calling this MVP complete.

ibis/backends/dask/udf.py
icexelloss: `execute_reduction_node_groupby` and `execute_analytic_node_groupby` sounds good to me

ibis/backends/dask/udf.py
icexelloss: sounds good

ibis/backends/dask/execution/window.py
icexelloss: kk thanks

ibis/backends/dask/execution/window.py
icexelloss: I think 

```
else execute(key, ...)
```
will not give you back a string. (Probably give it back a dd.Series)

icexelloss: If we never expect hit the else branch let's just remove it.

ibis/backends/dask/execution/window.py
icexelloss: Can we not override `data`? It seems a bit confusing just to reuse that variable. If you need to execute the root table let's give it another variable name

icexelloss (approved):
LGTM. +1

jreback (approved):

jreback:
thanks @gerrymanoim very nice!

PR #2861, opened 2021-07-14
CLN: Removing inspector optimization

datapythonista:
I don't think it adds much value to have this optimization in the code. `sqlalchemy.inspect(con).info_cache.clear()` is immediate, and we can save the calls to it only for code like `con.list_tables()`, which I don't see any use case for being called more than once (and not in a loop). So, I think this is making the code more complex, for no reason.


jreback (approved):
its fine, i suppose it might affect performance if its doing a lot of reflection.

PR #2862, opened 2021-07-14
CLN: Unify database methods in SQLAlchemy backends

datapythonista:
Moving duplicated methods from SQLAlchemy backends to base alchemy backend.

For `list_databases` and `list_schemas`, they do the same thing (in different ways) except for postgres. In all other backends `list_databases` is used, so `list_schemas` is deprecated everywhere except in postgres.


jreback:
can you rebase. note I am not sure if these are *really* tested that well.

datapythonista:
> can you rebase. note I am not sure if these are _really_ tested that well.

`list_databases` seems tested, I didn't check the rest. I'll try to spend some time reviewing the tests of the public API once it's well defined.

ibis/backends/base/sql/alchemy/client.py
```diff
@@ -229,6 +230,43 @@ def truncate_table(
...
+
+    def set_database(self, name):
+        raise NotImplementedError(
+            f'Cannot set database with {self.__class__.__name__} client. '
+            f'To use a different database, use `client.database("{name}")`'
+        )
+
+    def list_schemas(self):
+        warnings.warn(
+            '`list_schemas` is deprecated, use `list_databases` instead',
```
jreback, line 261: can you add a test for this. Also we should create an issue with deprecations as a reminder to remove them.

ibis/backends/tests/test_client.py
```diff
@@ -295,3 +295,27 @@ def test_insert_overwrite_from_expr(
...
     assert len(temporary.execute()) == 3
     tm.assert_frame_equal(temporary.execute(), from_table.execute())
+
+
+@pytest.mark.only_on_backends(
+    SQLALCHEMY_BACKENDS, reason="run only if backend is SQLAlchemy based",
+)
+def test_list_databases(con):
+    # Every backend has its own databases
+    TEST_DATABASES = {
```
jreback, line 305: could be a fixture at some point

ibis/backends/tests/test_client.py
```diff
@@ -295,3 +295,27 @@ def test_insert_overwrite_from_expr(
...
+
+
+@pytest.mark.only_on_backends(
+    set(SQLALCHEMY_BACKENDS) - {'postgres'},
+    reason="run only if backend is SQLAlchemy based, except postgres which "
+    "which has schemas different than databases",
+)
+def test_list_schemas(con):
+    with pytest.warns(FutureWarning):
+        schemas = con.list_schemas()
```
jreback, line 320: why is this the case?

datapythonista: Schemas as a PostgreSQL thing, and not common in many other database servers. They don't exist in SQLite or MySQL. And `list_schemas` is not present in SQL backends, only in SQLAlchemy backends.

What I want is to not have `list_schemas`  as an Ibis thing (it is now partly, as a SQLAlchemy method), and instead, leave it only as something the PostgreSQL backend implements. The warning is to not delete it directly, but to warn on Ibis 2.0, and remove it later. Does it sound reasonable?

ibis/backends/tests/test_client.py
datapythonista: I don't see this being used, now or in the future, outside of this test. That's what I thought a constant here was a better option. But happy to make it a fixture if you think it's better, even if I think it'll make things less readable to me.

jreback:
looks fine can you rebase

jreback (approved):

PR #2864, opened 2021-07-17
CLN: Create backend class hierarchy and remove duplicate code

datapythonista:
In preparation to merge the `Client` classes into the `Backend` classes, I create some base `Backend` classes to avoid code duplication, and some other minor changes that will help move the methods. To be specific:
- Created base classes for `Backend` classes that reuse code (file backends, pandas based backends, alchemy based backends)
- Removed the `kind` attribute, which wasn't used after removing the dialects, and won't be needed anymore after this
- Standardize backend classes to have a `client_class` attribute (this and the next point is needed, but temporary while merging `Backend` and `Client`)
- Standardize client classes to have a `backend` attribute
- Remove duplicate code in `Backend` (most of the duplicate code is in `Client` and will be removed later when merging the classes)


datapythonista:
@jreback whenever you've got the time to review my PRs, would be good if you can start by this one. This one is blocking me to continue with the work for the release. Thanks!

ibis/backends/dask/__init__.py
```diff
@@ -1,66 +1,17 @@
...
-        ----------
-        dictionary : dict
-
-        Returns
-        -------
-        DaskClient
-        """
-        return DaskClient(backend=self, dictionary=dictionary)
-
-    def from_dataframe(
```
jreback, line 32: where does this go?

jreback (changes requested):
you are removing dask apis

ibis/backends/dask/__init__.py
datapythonista: This is the same exact code as the same method in pandas. I created a base pandas backend class in the pandas backend, and both the pandas and the dask backend subclass it. So we can get rid of a bit of the code that was copied from pandas into dask and reuse it.

datapythonista:
@jreback any further comment here? This is blocking me from continuing to work on the pending stuff for the release. Thanks!

jreback (approved):

jreback:
lgtm @datapythonista thanks

PR #2868, opened 2021-07-18
TST: Moving data_directory fixture to a better location

datapythonista:
Having a `conftest.py` outside the package is strange and unconventional. And when running tests from third-party backends, the fixture is missing (not packaged) and tests fail.


jreback:
actually it's not strange at all
this is the original way pytest recommended

but no objection to moving it (as this is also acceptable)

datapythonista:
> actually it's not strange at all
> this is the original way pytest recommended
> 
> but no objection to moving it (as this is also acceptable)

Interesting, having seen any project using it. Well, the main reason for the PR is that it needs to be packaged, so it's availabe when running ibis tests in third-party backends. So, I need to move the fixture anyway.

jreback (approved):

PR #2869, opened 2021-07-18
CLN: Removing legacy Spark backend

### LINKED ISSUE #2775: DOC: Add information on why we have two Spark backends
We currently have a spark and a pyspark backend. The pyspark backends depends on the spark one, but both seem to be using pyspark.

It's not clear to me (and I guess it's not to users) why we have the two backends, and when one should be used or the other. I think we should provide this information in https://ibis-project.org/docs/backends/spark.html

@icexelloss @timothydijamco do you think you can take care of this?

datapythonista:
Closes #2775


jreback:
cc @icexelloss

ibis/backends/pyspark/client.py
```diff
@@ -66,3 +300,458 @@ def execute(
...
+        name : string
+          Database name
+        path : string, default None
+          Path where to store the database data; otherwise uses Spark
+          default
+        """
+        statement = CreateDatabase(name, path=path, can_exist=force)
+        return self.raw_sql(statement.compile())
+
+    def drop_database(self, name, force=False):
```
icexelloss, line 479: Are these method moved from the original spark file?

ibis/backends/pyspark/compiler.py
```diff
@@ -12,22 +12,33 @@
...
     spark_dtype,
 )
-from ibis.expr.timecontext import adjust_context
-from ibis.util import guid
-
 from .timecontext import combine_time_context, filter_by_time_context
 
 
-class PySparkTable(ops.DatabaseTable):
+class SparkUDFNode(ops.ValueOp):
```
icexelloss, line 31: I am not sure these are needed anymore. Does anything break if we remove this?

ibis/backends/pyspark/compiler.py
```diff
@@ -12,22 +12,33 @@
...
+    def output_type(self):
+        return rlz.shape_like(self.args, dtype=self.return_type)
+
+
+class SparkUDAFNode(ops.Reduction):
```
icexelloss, line 36: Same as SparkUDFNode

ibis/backends/pyspark/udf.py
```diff
@@ -84,7 +84,7 @@ def compiles_udf_node(t, expr):
                 UDFNode.__name__, ', '.join(map(t.translate, expr.op().args))
             )
 
-        SparkExprTranslator._registry[UDFNode] = compiles_udf_node
+        PySparkExprTranslator._registry[UDFNode] = compiles_udf_node
```
icexelloss, line 87: I don't think we need this anymore

icexelloss (changes requested):
Overall looks good. I think we can remove SparkUDF node otherwise LGTM.

ibis/backends/pyspark/compiler.py
datapythonista: Seems like those are used in `SparkUDF.__call__`. I don't understand the code of the spark udfs well, but I was wondering if only `SparkPandasUDF` should be left now, and `SparkUDF` was only for the spark backend, and not the pyspark one. Udf tests are failing, so there is something that needs to be done for sure.

If I understand correctly, the next are the public udf functions:
```python
>>> from ibis.spark.udf import udf
>>> @udf.elementwise(input_type=[dt.string], output_type=dt.int64)

>>> from ibis.spark.udf import udf
>>> @udf.elementwise.pandas([dt.string], dt.int64)

>>> from ibis.spark.udf import udf
>>> @udf.reduction(input_type=[dt.string], output_type=dt.int64)
```

Is the first one still supported after removing the spark backend, or can I remove it, and everything that depends on it? Including the classes that you mention.

icexelloss: The first one is not supported. (ibis.spark.udf namespace can be totally removed).

The correct namespace for UDF is under `ibis.udf.vectorized`.

icexelloss: (Sorry for the late reply, I don't have a very good notification setup..)

ibis/backends/pyspark/client.py
datapythonista: Yes, I moved everything that it was inherited from the base class, before removing it.

datapythonista:
Thanks for the feedback @icexelloss, I removed the ibis.spark.udf as suggested. For the ddl, I kept the functions from the spark backend, but since the ddl was only tested for the spark backend and not the pyspark, I had to xfail 5 tests that don't work in pyspark.

I guess we can make them pass in a follow up, or remove the functions being tested if they don't make sense, or they can't work for whatever reason. Does it make sense?

Let me know if there is anything else in this PR, I think should be ready.

ibis/backends/pyspark/tests/test_ddl.py
```diff
@@ -99,25 +107,29 @@ def test_create_empty_table(con, temp_table):
...
+    result_schema = client.get_schema(table_name)
     assert_equal(result_schema, schema)
 
-    assert con.table(table_name).execute().empty
+    assert client.table(table_name).execute().empty
 
 
-def test_insert_table(con, alltypes, temp_table, test_data_db):
+@pytest.mark.xfail(
+    reason='Moved from the legacy spark backend, '
```
jreback, line 119: can you create an issue for this

jreback (changes requested):

icexelloss (approved):
LGTM. Thanks!

jreback (approved):

jreback:
lgtm if you can create that issue (and then ok to merge), though if you want to put the issue number in the fail ok too

datapythonista:
> lgtm if you can create that issue (and then ok to merge), though if you want to put the issue number in the fail ok too

Good point, created #2879 and added the number to the xfail messages. Thanks both for the reviews.

PR #2870, opened 2021-07-19
Make TestConf.api work for third-party backends

datapythonista:
`TestConf.api` is assuming all backends can be accessed via `ibis.backends.<backend-name>.Backend()`, which is not true for third-party backends.

### DIFF
--- a/ibis/backends/tests/base.py
+++ b/ibis/backends/tests/base.py
@@ -144,7 +144,7 @@ class BackendTest(abc.ABC):
 
     @property
     def api(self):
-        return getattr(ibis.backends, self.name()).Backend()
+        return getattr(ibis, self.name())
 
     def make_context(
         self, params: Optional[Mapping[ir.ValueExpr, Any]] = None


ibis/backends/tests/base.py
```diff
@@ -144,7 +144,7 @@ def geo(self) -> Optional[ir.TableExpr]:
 
     @property
     def api(self):
-        return getattr(ibis.backends, self.name()).Backend()
```
jreback, line 147: this is for backwards compabtiblity?

datapythonista: No, they are equivalent, since `ibis.sqlite` resolves via `__getattr__` and entrypoints to `ibis.backends.sqlite.Backend()`. The problem is that `ibis.omniscidb` resolves to the OmniSciDB backend (if installed), but this line resolves to `ibis.backends.omniscidb.Backend()`, which doesn't exist, since the backend is in another package (not in the `ibis/backends` directory  in the Ibis repo).

jreback (approved):

PR #2873, opened 2021-07-21
Support binary operation with Timedelta in Pyspark backend

emilyreff7:
This PR adds support in the Pyspark backend for doing binary operations (e.g. add/subtract) on timestamp columns using Pandas Timedeltas, for example:

```
dt['timestamp_col'] + pd.Timedelta(days=5)
```

### DIFF
--- a/ibis/backends/pyspark/compiler.py
+++ b/ibis/backends/pyspark/compiler.py
@@ -1320,16 +1320,6 @@ _time_unit_mapping = {
     's': 'second',
 }
 
-# Ibis value to Pandas Timedelta value
-# The below mapping contains the union of the units represented
-# by Pandas Timedelta + units supported by PySpark.
-_time_unit_mapping_pd_timedelta = {
-    'D': 'days',
-    'h': 'hours',
-    'm': 'minutes',
-    's': 'seconds',
-}
-
 
 @compiles(ops.Date)
 def compile_date(t, expr, scope, timecontext, **kwargs):
@@ -1535,24 +1525,26 @@ def _get_interval_col(
     dtype = op.dtype
     if not isinstance(dtype, dtypes.Interval):
         raise com.UnsupportedArgumentError(
-            '{} expression cannot be converted to interval column. '
-            'Must be Interval dtype.'.format(dtype)
+            f'{dtype} expression cannot be converted to interval column. '
+            'Must be Interval dtype.'
         )
     if allowed_units and dtype.unit not in allowed_units:
         raise com.UnsupportedArgumentError(
-            'Interval unit "{}" is not allowed. Allowed units are: '
-            '{}'.format(dtype.unit, allowed_units)
+            f'Interval unit "{dtype.unit}" is not allowed. Allowed units are: '
+            f'{allowed_units}'
         )
 
     if isinstance(op.value, pd.Timedelta):
-        pandas_unit = _time_unit_mapping_pd_timedelta[dtype.unit]
-        value = getattr(op.value.components, pandas_unit)
+        td_nanos = op.value.value
+        if td_nanos % 1000 != 0:
+            raise com.UnsupportedArgumentError(
+                'Interval with nanoseconds is not supported. The '
+                'smallest unit supported by Spark is microseconds.'
+            )
+        td_micros = td_nanos // 1000
+        return F.expr(f'INTERVAL {td_micros} MICROSECOND')
     else:
-        value = op.value
-
-    return F.expr(
-        'INTERVAL {} {}'.format(value, _time_unit_mapping[dtype.unit])
-    )
+        return F.expr(f'INTERVAL {op.value} {_time_unit_mapping[dtype.unit]}')
 
 
 def _compile_datetime_binop(
--- a/ibis/backends/pyspark/compiler.py
+++ b/ibis/backends/pyspark/compiler.py
@@ -4,6 +4,7 @@ import functools
 import operator
 
 import numpy as np
+import pandas as pd
 import pyspark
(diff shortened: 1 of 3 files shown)


ibis/backends/pyspark/compiler.py
```diff
@@ -1532,8 +1543,15 @@ def _get_interval_col(
             'Interval unit "{}" is not allowed. Allowed units are: '
             '{}'.format(dtype.unit, allowed_units)
         )
+
+    if isinstance(op.value, pd.Timedelta):
+        pandas_unit = _time_unit_mapping_pd_timedelta[dtype.unit]
```
icexelloss, line 1548: I am not sure we need this. Does pyspark not work with pd.Timedelta directly?

emilyreff7: If we just pass in op.value as the existing code does, then Pyspark blows up with a ParseException, e.g. if we run 

```
t.timestamp_col + pd.Timedelta('10s')
```
the error is

```
E   pyspark.sql.utils.ParseException: 
E   mismatched input ':' expecting {'ADD', 'AFTER', 'ALL', 'ALTER'...
<truncated>
E   
E   == SQL ==
E   INTERVAL 0 days 00:00:10 second
E   ------------------^^^

```
so at least if we want to use this `INTERVAL` syntax, Pyspark doesn't do well with the colons in the Pandas Timedelta representation.

ibis/backends/pyspark/compiler.py
```diff
@@ -1532,8 +1543,15 @@ def _get_interval_col(
...
+        value = getattr(op.value.components, pandas_unit)
+    else:
+        value = op.value
```
icexelloss, line 1551: What hits the else here? (What other than pd.Timedelta supported)?

ibis/backends/pyspark/compiler.py
```diff
@@ -1532,8 +1543,15 @@ def _get_interval_col(
             'Interval unit "{}" is not allowed. Allowed units are: '
             '{}'.format(dtype.unit, allowed_units)
         )
+
+    if isinstance(op.value, pd.Timedelta):
+        pandas_unit = _time_unit_mapping_pd_timedelta[dtype.unit]
+        value = getattr(op.value.components, pandas_unit)
```
icexelloss, line 1549: Is this safe? Can there be more than one components? e.g., hour and minute?

icexelloss: I think it is probably better to do sth like this

```
td = op.value # pd.Timedelta
td_nanos = td.value
# Spark supports micro as the smallest unit, so if we have any nanos component, raise Error

td_micro = td_nanos / 1000

spark_inteval = F.expr(f'INTERVAL {td_mirco} MICROSECOND')
```

icexelloss: This way you don't even need to keep the mapping

ibis/backends/pyspark/compiler.py
icexelloss: kk

icexelloss (changes requested):
Request for small change to simply the logic

ibis/backends/tests/test_temporal.py
```diff
@@ -306,6 +307,29 @@ def test_temporal_binop(backend, con, alltypes, df, expr_fn, expected_fn):
     backend.assert_series_equal(result, expected)
 
 
+@pytest.mark.parametrize(
+    ('timedelta', 'temporal_fn'),
+    itertools.product(
+        ['5W', '3d', '2h', '3m', '10s'],
```
icexelloss, line 313: Please also test really large timedelta like 100 years to make sure we don't overflow

icexelloss: Please add also test for unsupported shifts (e.g. 1 nanos)

ibis/backends/pyspark/compiler.py
emilyreff7: All of these existing tests for example hit the else: https://github.com/ibis-project/ibis/blob/master/ibis/backends/tests/test_temporal.py#L299
`ibis.interval`, `ibis.date`, etc.

ibis/backends/pyspark/compiler.py
emilyreff7: Makes sense, definitely simpler than maintaining a mapping - and Pandas errors out immediately anyway if you pass a value to Timedelta that would cause an overflow, e.g. 

```
>>> pd.Timedelta('1000Y')
  File "pandas/_libs/tslibs/timedeltas.pyx", line 1217, in pandas._libs.tslibs.timedeltas.Timedelta.__new__
  File "pandas/_libs/tslibs/timedeltas.pyx", line 462, in pandas._libs.tslibs.timedeltas.parse_timedelta_string
  File "pandas/_libs/tslibs/timedeltas.pyx", line 514, in pandas._libs.tslibs.timedeltas.timedelta_from_spec
  File "pandas/_libs/tslibs/timedeltas.pyx", line 312, in pandas._libs.tslibs.timedeltas.cast_from_unit
OverflowError: Python int too large to convert to C long
```

ibis/backends/tests/test_temporal.py
emilyreff7: Added test for both

icexelloss (commented):
LGTM. +1

icexelloss (approved):

ibis/backends/pyspark/compiler.py
icexelloss: Great!

jreback:
thanks @emilyreff7

Issue #2852, opened 2021-07-13
refactor: fix mypy errors in the Dask backend


datapythonista:
xref #2823

Running `mypy ibis/backends/dask/` reports 166 errors, related to missing or inconsistent type annotations. We should fix them, and make that command finish in a clean way.

Here there are some sample fixes: #2857

cpcloud:
Rolling this up under #2823. Great job on the progress here @sofide, most of the errors are fixed!

We're down to 22 errors!

PR #2875, opened 2021-07-23
CI: Fix most mypy errors in the Dask backend

sofide:
Partially solves issue #2852 

There are some remaining errors, but I think they might need a deeper code refactor.

### DIFF
--- a/ibis/backends/dask/client.py
+++ b/ibis/backends/dask/client.py
@@ -244,12 +244,9 @@ class DaskClient(Client):
         schema: sch.Schema = None,
     ):
         """Create a table."""
-        if obj is None and schema is None:
-            raise com.IbisError('Must pass expr or schema')
-
         if obj is not None:
             df = obj
-        else:
+        elif schema is not None:
             dtypes = ibis_schema_to_dask(schema)
             df = schema.apply_to(
                 dd.from_pandas(
@@ -257,6 +254,8 @@ class DaskClient(Client):
                     npartitions=1,
                 )
             )
+        else:
+            raise com.IbisError('Must pass expr or schema')
 
         self.dictionary[table_name] = df
 
--- a/ibis/backends/dask/execution/selection.py
+++ b/ibis/backends/dask/execution/selection.py
@@ -37,7 +37,7 @@ def compute_projection_scalar_expr(
     expr,
     parent,
     data,
-    scope: Scope = None,
+    scope: Scope,
     timecontext: Optional[TimeContext] = None,
     **kwargs,
 ):
--- a/ibis/backends/dask/execution/util.py
+++ b/ibis/backends/dask/execution/util.py
@@ -160,9 +160,7 @@ def _pandas_dtype_from_dd_scalar(x: dd.core.Scalar):
 
 
 def _coerce_to_dataframe(
-    data: Any,
-    column_names: List[str],
-    types: Optional[List[dt.DataType]] = None,
+    data: Any, column_names: List[str], types: List[dt.DataType],
 ) -> dd.DataFrame:
     """
     Clone of ibis.util.coerce_to_dataframe that deals well with dask types
--- a/ibis/backends/dask/execution/util.py
+++ b/ibis/backends/dask/execution/util.py
@@ -20,7 +20,7 @@ from ibis.expr.typing import TimeContext
 
 from ..core import execute
 
-DispatchRule = Tuple[Tuple[Type, ...], Callable]
+DispatchRule = Tuple[Tuple[Union[Type, Tuple], ...], Callable]
 
 TypeRegistrationDict = Dict[Type[ops.Node], List[DispatchRule]]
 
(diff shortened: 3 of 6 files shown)


ibis/backends/dask/execution/selection.py
```diff
@@ -37,7 +37,7 @@ def compute_projection_scalar_expr(
     expr,
     parent,
     data,
-    scope: Scope = None,
+    scope: Scope,
```
datapythonista, line 40: Any reason to remove the default value, instead of setting the type to `Optional[Scope]`, which seems like a most conservative approach?

sofide: @datapythonista  The problem wasn't the missing type `Optional`: when a default is `None` the type `Optional` is implicitly added, and mypy doesn't have any problem with it (although explicit is always better than implicit :P )

The problem is the line 52:
```python
    scope = scope.merge_scopes(
```
If `scope` is `None` it will raise an exception.

datapythonista: Ah, great, thanka for the info. @icexelloss can you have a quick look at this PR please?

ibis/backends/dask/execution/util.py
```diff
@@ -158,9 +160,7 @@ def _pandas_dtype_from_dd_scalar(x: dd.core.Scalar):
 
 
 def _coerce_to_dataframe(
-    data: Any,
-    column_names: List[str],
-    types: Optional[List[dt.DataType]] = None,
```
sofide, line 163: @datapythonista The problem here is that if `types` is `None` and `data` is a `Series`, this part of the code will raise an error:
```python
    elif isinstance(data, dd.Series):
        dtypes = map(ibis_dtype_to_pandas, types)
```
Would you prefer to manage that case and raise a specific error?

ibis/backends/dask/client.py
```diff
@@ -244,19 +244,18 @@ def create_table(
         schema: sch.Schema = None,
     ):
         """Create a table."""
-        if obj is None and schema is None:
-            raise com.IbisError('Must pass expr or schema')
-
         if obj is not None:
             df = obj
-        else:
+        elif schema is not None:
```
jreback, line 249: why this change? this is a logic change

jreback (changes requested):
cc @gerrymanoim if you'd have a look

ibis/backends/dask/client.py
datapythonista: Not sure if there was any mypy problem in the original code, but the logic is the same (not immediately obvious, took me a bit to see it), and I think the new implementation is clearer.

sofide: Without this change, mypy raise this error:
```
ibis/backends/dask/client.py:254: error: Item "None" of "Optional[Schema]" has no attribute "apply_to"
```
The logic is the same, but in the original code mypy cannot figure out that schema will never be `None` in that line.

gerrymanoim: Shouldn't this look the same in pandas then? https://github.com/ibis-project/ibis/blob/master/ibis/backends/pandas/client.py#L409

I'm fine with the change, but I'd expect that the `PandasClient` and `DaskClient` look extremely similar code-wise.

ibis/backends/dask/execution/util.py
gerrymanoim: I'm fine removing the optional. This is only used internally in `coerce_to_output` which will always pass types.

ibis/backends/dask/client.py
datapythonista: I created issues to address one backend at a time. But we'll surely have to make the same change in the pandas backend when we fix the mypy errors on it, since the mypy problem will be the same.

datapythonista (approved):
Thanks @sofide, for the contribution, and for all the additional information. All good then. @jreback if you want to check the comments...

ibis/backends/dask/client.py
sofide: @arielrossanigo have taken this issue in the pandas backend, maybe we can coordinate to make the same change in both backends

arielrossanigo: Hi people, I've checked and currently in the `PandasClient` there is no mypy problem because of the lack of annotations. 

I think @sofide can reflect the change in the `PandasClient` and then I rebase my branch when this PR is merged, What do you think?

datapythonista: Don't worry about this, this PR should be ready. Will wait a bit more to see if @jreback can have a last look, but all is fine.

And regarding pandas, not so concerned that this function has the same code. I don't think we should have copied pasted anyway, so if we want to do things properly what we should do is use the pandas function for dask, not make sure we make the same changes in both functions. But let's focus on the mypy problems.

jreback (approved):

jreback:
thanks @sofide

PR #2883, opened 2021-07-29
CLN: Start definition of Backend API, and implement version

datapythonista:
xref #2678

First step towards defining the Backend API properly with abstract methods (Backend will replace Client).

To make PRs small, I define a first set of methods that Backend/Client must have, and so far I implement version.

To be able to do this step by step, I move the implementation from Client to Backend, and I created the property in the base Client, calling the Backend implementation. Eventually, `connect` will return `Backend` instead of `Client`, and Client will be removed.

I'm making a **backward-incompatible** change here, returning version as a string, instead of a setuptools object. I think this will match user expactations, since libraries rarely parse it (i.e. `any_library.__version__`) and it will also make our life easier with type annotations (setuptools have different version objects that can be returned).


datapythonista:
@jreback if you don't mind prioritizing this one, this is blocking me from finishing the stuff for the release

docs/source/api.rst
```diff
@@ -39,6 +39,25 @@ These methods are available directly in the ``ibis`` module namespace.
    trailing_range_window
    random
 
+.. _api.basebackend:
```
jreback, line 42: why does this need to be listed as its internal? (yes its public facing for library extenders but not for users).

ibis/backends/base/__init__.py
```diff
@@ -40,6 +41,57 @@ def connect(connection_string, **options):
         Connect to the underlying database and return a client object.
         """
 
+    # @abc.abstractmethod
```
jreback, line 44: is the intent to uncomment?

jreback (changes requested):
looks generally ok

docs/source/api.rst
datapythonista: Backends will be forced to implement them, and the subclass is public. My idea is to document the base class, instead of repeating the documentation for each backend, that will be the same. Also, if we do it this way, backends don't need to write the docs of the methods, which should ve all the same.

ibis/backends/base/__init__.py
datapythonista: Yes. I can't make them abstract in this PR, since the methods are not implemented in the subclasses yet (to not make this PR huge). Immediately after this is merged, I'll open a PR for each of these methods, with the same approach I did for `version` here. I'll uncomment in those, so didn't bother to add a comment, since it won't stay long commented.

datapythonista:
@jreback whenever you've got time, would be great if you can check my comments. I think this should be ready, and it's blocking me from continuing with clarifying the API for the backends. Thanks!

jreback (approved):

jreback:
thanks @datapythonista prob should add a release note about the version string change next time.

datapythonista:
> thanks @datapythonista prob should add a release note about the version string change next time.

Good point, added it in #2895, thanks for merging anyway.

ibis/backends/sqlite/__init__.py
```diff
@@ -41,4 +41,5 @@ def connect(self, path=None, create=False):
         create : boolean, default False
             If file does not exist, create it
         """
-        return SQLiteClient(backend=self, path=path, create=create)
+        self.client = SQLiteClient(backend=self, path=path, create=create)
```
cpcloud, line 44: I realize this PR is old, but is it possible to clarify why a reference cycle is necessary here? Why do both the client and backend need references to each other?

datapythonista: This is temporary, and was implemented to let transition from Client to Backend step by step. So, we had manageable PRs instead of a huge one.

The reason of the references is that any method moved from Client to Backend using self, had to use self.client instead. And methods not yet moved that called moved methods, had to use self.backend instead of the original self.

Will get rid of it soon, as I'm workingon merging the remaining methods in Client into Backend.

Does it make sense?

cpcloud: Makes sense. Thanks for clarifying.

PR #2884, opened 2021-07-29
BUG: Loading pandas options for file backends

### LINKED ISSUE #2844: BUG: File pseudo-backends failing for missing pandas option
The next code is failing in master since #2833:

```python
>>> import ibis
>>> con = ibis.csv.connect('/home/mgarcia/src/ibis/ci/ibis-testing-data/')
>>> expr = con.table('functional_alltypes')['double_col'] * 2
>>> print(expr.execute())
OptionError: "No such keys(s): 'pandas.enable_trace'"
```

The problem is when the `csv` backend  (or other file backends) are loaded, but the pandas backend is not. This is because `ibis.pandas` loads the pandas options, which looks like they are needed by the file pseudo-backends.

The CI is not failing, I guess because we test pandas and the file backends are tested together, and pandas is loaded when the file backends are tested.

datapythonista:
Closes #2844

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,8 @@ jobs:
     name: Tests pandas / files
     runs-on: ${{ matrix.os }}
     env:
-      BACKENDS: "pandas csv parquet hdf5"
+      # FIXME removing pandas just to test the CI, restore before merging
+      BACKENDS: "csv parquet hdf5"
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


jreback (approved):
lgtm. any tests hit this? (if you have something pls do as a followon is fine)

datapythonista:
We've got tests, the problem was that in the CI we load together the file backends and the pandas one, that's why it wasn't failing. Bit I test both locally and in the first commit of this PR that it works without loading the pandas backend.

PR #2888, opened 2021-07-30
Add GitHub Actions step with only pytest errors

datapythonista:
It takes a lot of scrolling to find the traceback of tests errors. This adds a new step in GitHub actions with only the useful information, so it's immediate to find.


datapythonista:
This is how the new step added here looks like: https://github.com/ibis-project/ibis/pull/2888/checks?check_run_id=3204005989#step:5:6

jreback (approved):

PR #2897, opened 2021-08-06
Deprecate Client.database method and move to BaseBackend

### LINKED ISSUE #2878: CLN: Deprecate `Database`
Our current `Database` class (i.e. `ibis.sqlite.database('foo')`) seems to be just a set of shortcuts of things that can be done directly with the `Client` class. Feels like it just adds noise and complexity, for some syntactic sugar that in my opinion is not worth. I'd simply deprecate it, so users have a single way to do things like `con.list_tables()` / `con.database('foo').list_tables()`.

datapythonista:
xref #2678
Closes #2878

`Client.database` has the same implementation in all backends, but it was repeated in each of them. Moved to `BaseBackend`, removed from the backends, and add a `FutureWarning` for its removal.


jreback:
do we have a depreacation removal issue?

jreback (approved):

PR #2899, opened 2021-08-06
CLN: Standardize and move list_tables method to backend

datapythonista:
xref #2678

Unifying the signatures of `list_tables`, reuse code as much as possible, and move from `Client` to `Backend`.


jreback:
can you rebase

ibis/backends/base/client.py
```diff
@@ -136,22 +139,5 @@ def table(self, name: str) -> ir.TableExpr:
...
-        list
-            A list with all tables available for the current database.
-        """
-        return self.client.list_tables(
-            like=self._qualify_like(like), database=self.name
-        )
-
-    def _qualify_like(self, like: str | None) -> str | None:
-        return like
+    def list_tables(self, like=None):
```
jreback, line 142: why does this have a different signature?

ibis/backends/base/sql/alchemy/client.py
```diff
@@ -262,35 +262,8 @@ def list_schemas(self):
...
-
-        """
-        inspector = self.inspector
-        # inspector returns a mutable version of its names, so make a copy.
-        names = inspector.get_table_names(schema=schema).copy()
-        names.extend(inspector.get_view_names(schema=schema))
-        if like is not None:
-            names = [x for x in names if like in x]
-        return sorted(names)
+    def list_tables(self, like=None, database=None, schema=None):
```
jreback, line 265: this signature is completely different that the others

jreback (changes requested):
these methods are having completely different signatures. is there a reason?

ibis/backends/base/client.py
datapythonista: This is `Database.list_tables()`. I didn't add the `database` argument here, since it feels a bit strange to do `con.database('foo').list_tables(database='bar')`. I can add it if you want to have the same signature, but in any case, `Database` is set for deprecation, so no big deal.

ibis/backends/base/sql/alchemy/client.py
datapythonista: This is to keep code backward-compatible, since in SQLAlchemy backends use `schema='foo'` instead of `database='foo'`. I see that to pass the test I removed the deprecation code of the `schema` argument and I haven't restored it. If you don't think it's a big deal, I'll just remove `schema`, much simpler.

datapythonista:
@jreback this should be ready now. I removed differences in the signature of SQLAlchemy backend, which didn't make sense (kept them before for backward-compatibility). And cleaned some of the related schema stuff (postgres supports databases and schemas, and most of it was removed from the base SQLAlchemy code, but few things were left).

There are two signatures that are still different:
- File backends (would be nice to remove the path parameters, which is cumbersome, as discussed with Phillip, but better in a follow up, after further discussion)
- `Database.list_tables`, where passing again a `database` parameter is redundant. And it's being removed, so probably not worth anyway. Can add if you have a strong opinion about it, but I'd leave it as is.

Let me know if this looks good now. Thanks!

ibis/backends/base/file/__init__.py
```diff
@@ -134,6 +134,25 @@ def connect(self, path):
     def version(self) -> str:
         return pd.__version__
 
+    def list_tables(self, path=None, like=None):
+        """
```
jreback, line 138: if you override the doc-string then you need to make a full doc-string. also pls type all args

jreback (changes requested):

ibis/backends/base/client.py
datapythonista: This operation is in a `Database` object.

What we have in this PR now is:
```python
>>> import ibis
>>> con = ibis.sqlite.connect(...)
>>> con.list_tables()  # `database` is the current database
>>> con.list_tables(database='bar')  # I can specify a different database when listing tables from the connection/client
>>> con.database('bar').list_tables()  # another way to list tables from a different database, equivalent to the previous line
>>> con.database('bar').list_tables(database='foobar')  # DOES NOT WORK. This could work if we make this signature the same as the rest.
```

What does need to be resolved? Do you want to make the last line work? I'm happy to add a `database` argument here if you want, but personally I don't think offering this to the user adds value.

ibis/backends/base/file/__init__.py
datapythonista: Good point, updated.

ibis/backends/base/client.py
jreback: i actually don't care about the last line, instead about a consistent signature. I don't think the database arg adds *any* value. so either remove it everywhere or have it everywhere.

datapythonista: Ok, I'll add it here for now, and will create an issue to discuss if there are better ways to support databases different than the default. Thanks!

jreback (approved):
lgtm. if there are actually any user facing changes can you list them in a release note @datapythonista

Issue #2854, opened 2021-07-13
refactor: fix mypy errors in the pandas backend


datapythonista:
xref #2823

Running `mypy ibis/backends/pandas/` reports 35 errors, related to missing or inconsistent type annotations. We should fix them, and make that command finish in a clean way.

Here there are some sample fixes: #2857

cpcloud:
There's only 1 error left and it's not actually related to ibis per se, which is awesome. Thanks @arielrossanigo.

PR #2900, opened 2021-08-09
CI: Fix mypy errors in the pandas backend

arielrossanigo:
Hi! This PR solves all errors reported in #2854.

I want to remark that I've ignored two mypy errors regarding monkeypatching of `DataType` and `Schema` in order to add `to_pandas` functionality. These are the lines for reference: https://github.com/ibis-project/ibis/blob/master/ibis/backends/pandas/client.py#L336

IMHO, this is not the best nor clearest way of adding `to_pandas`, for instance, a lot of extra mypy errors will appear if all convert functions above the reference line were completely annotated.

Maybe that functionality could be moved as methods of the classes or `ibis_dtype_to_pandas` and `ibis_schema_to_pandas` free functions could be used instead (I've started with this last option but later I realize `to_pandas` is being used in other places in the code.


environment.yml
```diff
@@ -45,6 +45,9 @@ dependencies:
   - pip
   - pip:
     - pysuerga
+    - types-python-dateutil
+    - types-pytz
+    - types-setuptools
```
datapythonista, line 50: These two are already at the end of the file, from conda-forge (not pip). I think we added them after you started this PR. Can you remove, and also check if we can add `types-python-dateutil` from conda-forge for consistency please.

ibis/backends/pandas/aggcontext.py
```diff
@@ -266,29 +266,37 @@ def agg(self, grouped_data, function, *args, **kwargs):
 
 def wrap_for_apply(
     function: Callable,
-    args: Optional[Tuple[Any]],
-    kwargs: Optional[Dict[Any, Any]],
+    args: Optional[Tuple[Any, ...]] = None,
+    kwargs: Optional[Dict[str, Any]] = None,
```
datapythonista, line 270: I'm sure there is a reason for this, but why do we need to set a default value for the annotations? If there wasn't a default value, I assume all the calls provide those values. It's not obvious how this helps.

ibis/backends/pandas/client.py
```diff
@@ -120,15 +121,14 @@ def from_pandas_categorical(value):
...
-    (np.generic,)
-    + tuple(
-        frozenset(
-            np.signedinteger.__subclasses__()
-            + np.unsignedinteger.__subclasses__()  # np.int64, np.uint64, etc.
-        )
-    )  # we need this because in Python 2 int is a parent of np.integer
-)
+_numpy_scalar_types: Tuple[Type, ...] = (np.generic,)
+# we need this because in Python 2 int is a parent of np.integer
```
datapythonista, line 125: We do not support Python 2 anymore, maybe this is not needed anymore.

datapythonista (commented):
Thanks! Great job. I added some comments. Will have another look later.

ibis/backends/pandas/aggcontext.py
arielrossanigo: The default value was not added for annotations but because there is a use of this function were those parameters were not provided. I was tempted of removing the Optional annotation but later find that call. Here you can see the call https://github.com/ibis-project/ibis/blob/18e967cac961285b05d8df560f40148bac1a2571/ibis/backends/pandas/aggcontext.py#L590-L593 

If you're ok, I can remove the Optional part, the default values and change the call to provide an empty tuple and dict instead.

environment.yml
arielrossanigo: Indeed, they were added after I started working in the PR. I've made the changes you suggest.

ibis/backends/pandas/client.py
arielrossanigo: Done! I've keep only `np.generic` and it is enough to catch all numpy dtypes.

jreback (approved):

jreback:
thanks @arielrossanigo

PR #2901, opened 2021-08-10
wip: Fix geo equals

### LINKED ISSUE #2197: mutate geo column causes TypeError: geo_equals() got an unexpected keyword argument 'cache'
Using `mutate` with the same GEO column more than once causes an error.  The error happens when printing or using `compile()`.  If I only reference the same column once, the result is correct.

I'm using this with omniscidb.

The `geo_equals` function does not have a cache arg: https://github.com/ibis-project/ibis/blob/cc0ebb6242e75670d282476d6ea10bf5b8583fc4/ibis/expr/api.py#L1845

```
tm = t.mutate([
    t['Location'].y().name('Location_Latitude'),
    t['Location'].y().name('Latitude')])
tm.compile()
```

```
/opt/conda/lib/python3.7/site-packages/ibis/expr/operations.py in all_equal(left, right, cache)
    171 
    172     if hasattr(left, 'equals'):
--> 173         return left.equals(right, cache=cache)
    174     return left == right
    175 

TypeError: geo_equals() got an unexpected keyword argument 'cache'
```

xmnlab:
Fix #2197

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -13,6 +13,7 @@ dependencies:
   - toolz
   - cached_property # for 3.7 compat, functools.cached_property is for >=3.8
   - setuptools
+  - python 3.8.*
 
   # Ibis soft dependencies
   - sqlalchemy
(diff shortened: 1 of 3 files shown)


datapythonista:
I think this is superseded by #2901. Let me know if there is something here not fixes in that PR @xmnlab and I'll reopen.

PR #2902, opened 2021-08-11
TST: Running and fixing geo tests

### LINKED ISSUE #2673: TST: Are geospatial tests running?
I think some tests are never running in the CI, because of missing dependencies:
```
SKIPPED [1] ibis/tests/expr/test_geospatial.py:11: could not import 'shapely': No module named 'shapely'
SKIPPED [1] ibis/backends/tests/test_geospatial.py:9: could not import 'geopandas': No module named 'geopandas'
SKIPPED [1] ibis/backends/postgres/tests/test_postgis.py:5: could not import 'geopandas': No module named 'geopandas'
```
I think we should first remove the pytest skips based on dependencies, and after that get the dependencies installed, so the tests run.

datapythonista:
Closes #2673 

Make sure geo tests are running, fixing broken ones, and moving geo tests from `ibis/tests` to `ibis/backends/tests` since they are backend dependent and not core tests.


datapythonista:
@xmnlab I still need to update the tests, but I think this is the best approach. Let me know if you've got any comment.

xmnlab:
@datapythonista the changes make sense to me! thanks for working on that!

jreback:
can you rebase

ibis/backends/postgres/tests/test_postgis.py
```diff
@@ -71,16 +68,20 @@ def test_select_multipolygon_geodata(geotable):
 
 
 def test_geo_area(geotable, gdf):
+    import geopandas
```
jreback, line 71: shouldn't you have a skip if this is not available? (as its not a strict requirement of postgres)

jreback (changes requested):

ibis/backends/postgres/tests/test_postgis.py
datapythonista: Those skips are what caused the tests to be silently skipped. Not a big fan of those, they've been causing to tests not run without anyone noticing several times both in pandas and in ibis. So, this approach doesn't skip silently, and uses markers to control what is being run. If dependencies are not present tests will fail. If we don't want those tests to run, because we don't want to test this, we can just call pytest with `-m "not geo"`. I think this is very simple, and we won't have these tests being skipped from the CI again for months without noticing.

I'd like to eventually have all these skips removed.

jreback: ok cool.

jreback (approved):

Issue #2863, opened 2021-07-17
refactor: remove deprecations


datapythonista:
Before releasing, we need to remove the deprecations from previous releases.

**Introduced in Ibis 2.0**
- [ ] `list_schema` on SQLAlchemy backends (#2862)
- [x] `expr.verify()` and `ibis.<backend>.verify()` (#2866)
- [ ] `Client.database` and `Database` class (#2897)
- [ ] `Client.exists_table` (#2905)

jreback:
no these can't be removed for 2.0; need another version for that

datapythonista:
Yep, clarified the description, 2.0 is when they are being introduced. I think we can remove them in 3.0, but leaving it open for now.

cpcloud:
We have the `@deprecated` decorator now, so it should be straightforward to figure when a thing was deprecated and remove it.

However, we still need to come up with a policy. Let's open up a GitHub Discussion and talk about it.

PR #2905, opened 2021-08-16
CLN: Deprecate and standardize exists_table

datapythonista:
Not very useful, and backends reimplement it for no reason.


jreback:
can you rebase

jreback:
can you rebase

jreback:
@datapythonista can you rebase here

ibis/backends/impala/tests/test_client.py
```diff
@@ -266,11 +266,15 @@ def test_close_drops_temp_tables(con, test_data_dir):
 
     table = con.parquet_file(hdfs_path)
 
-    name = table.op().name
-    assert con.exists_table(name) is True
+    qualified_name = table.op().name
```
cpcloud, line 269: After reading through this test, and the code path it's trying to exercise I no longer think this test is useful nor is there any real way to make this test reliable.

The main question here is: what should happen to any object that is referencing a temporary object from an *open* database connection, when that connection is closed?

I don't think this behavior is well defined, nor should we try to define it. However, since we're depending on `__del__` to clean up `ImpalaTemporaryTable`s (roughly speaking) when they are garbage collected, which should _definitely_ not be dropping tables when a connection is closed.

I will submit a PR to address that.

In this PR I think you can remove this test.

jreback (changes requested):
can you assert the FutureWarning in a test (I think this is how we handled other deprecated items)

also add a release note

ibis/backends/impala/tests/test_client.py
cpcloud: Ok, so this is actually a different problem (though the lifetime problem still exists).

The issue is that `con.exists_table(name)` has different semantics than what you're inlining here.

`exists_table(like=name)` will return true if the list of tables matching `name` is non-empty.

What you've written here is to assert that whatever tables are listed contains `name` as an exact match.

The important thing to note is that this is only true when `con` is connected to the same database that `name` is contained in. Otherwise, the test will fail.

The solution is to write the assertion as `assert con.list_tables(like=name)`.

cpcloud (changes requested):

datapythonista:
I saw the `like` issue. I think it's a bug, but I didn't overthink it, as `exists_table` is being deprecated anyway. I don't think `con.exists_able('cat')` should be true, if a table `catalog` exists, but no table `cat`. This is indirectly being fixed with this clean up.

But I don't think the problem with the test is that, I did try what you say, and also printed the tables in the database to see if it was a prefix problem, or something related, but the table just doesn't exist.

datapythonista:
This is green now. If it's still find to delete the tricky test, this should be ready.

cpcloud:
> I saw the `like` issue. I think it's a bug, but I didn't overthink it, as `exists_table` is being deprecated anyway. I don't think `con.exists_able('cat')` should be true, if a table `catalog` exists, but no table `cat`. This is indirectly being fixed with this clean up.

Can you clarify this a bit? Ibis doesn't currently have this behavior. Here's me executing a few `exists_table` calls in the test:

```
(Pdb) list
265
266         table = con.parquet_file(hdfs_path)
267
268         name = table.op().name
269         breakpoint()
270  ->     assert con.exists_table(name) is True
271         con.close()
272
273         assert not con.exists_table(name)
274
275
(Pdb) con.list_tables()
['alltypes', 'functional_alltypes', 'tpch_customer', 'tpch_lineitem', 'tpch_nation', 'tpch_orders', 'tpch_part', 'tpch_partsupp', 'tpch_region', 'tpch_region_avro', 'tpch_supplier']
(Pdb) con.exists_table('all')
False
(Pdb) con.exists_table('func')
False
(Pdb) con.exists_table('tpch_')
False
```

> But I don't think the problem with the test is that, I did try what you say, and also printed the tables in the database to see if it was a prefix problem, or something related, but the table just doesn't exist.

The table exists, it's just in a different database.

What's happening is the prefix is being split off by a regular expression and that prefix is used as the database in `SHOW TABLES IN __ibis_tmp LIKE '__ibis_tmp_<SOME UUID>'` query.

The `LIKE` doesn't contain any wildcards, so it'll perform an exact match..

We should not delete the test until we understand why it fails.

ibis/backends/base/file/__init__.py
```diff
@@ -134,7 +136,9 @@ def connect(self, path):
     def version(self) -> str:
         return pd.__version__
 
-    def list_tables(self, path: Path = None, like: str = None):
+    def list_tables(
+        self, path: Path = None, like: str = None, database: str = None
```
cpcloud, line 140: We should eventually require `Optional` here or use the new annotations syntax of `T | None`, but that can be done in a separate PR.

ibis/backends/base/__init__.py
```diff
@@ -153,6 +153,20 @@ def list_tables(self, like: str = None, database: str = None) -> list[str]:
...
+
+        Deprecated in Ibis 2.0. Use `name in client.list_tables()` instead.
+        """
+        warnings.warn(
+            '`client.exists_table(name)` is deprecated, and will be '
+            'removed in a future version of Ibis. Use '
+            '`name in client.list_tables()` instead.',
+            FutureWarning,
+        )
+        return name in self.client.list_tables(database=database)
```
cpcloud, line 168: Should this be implemented the way it was before as `len(self.client.list_tables(like=name, database=database)) > 0`?

cpcloud: `name` won't match tables in other databases, so I don't think you can use `in` here. Since this isn't doing something trivial, should we reconsider deprecating it?

datapythonista: I don't fully understand what's Impala doing. I don't think other backends are affected.

We've got a syntax for checking other databases, `name in con.list_tables(database='foo')`. And I think it's better than this magic failing the test. Just the explicit syntax doesn't seem to work in Impala (or there was an error in my code when I tried that, but I don't think so). I'm using the previous syntax of checking the list of tables with the `like` param for now, so we can get tests passing.

ibis/backends/base/file/__init__.py
datapythonista: Afaik mypy is smart enough to consider anything having a default value of `None` as optional, I don't think it's needed.

cpcloud: Unfortunately, mypy's ability to infer types is unrelated to the requirement for `Optional` or an equivalent syntax.

`Path` is simply not equivalent to `Optional[Path]`. If an API accepts `Path`, it should not accept `None`, because `Path` doesn't sufficiently model a option type.

ibis/backends/base/__init__.py
cpcloud: > I don't fully understand what's Impala doing. I don't think other backends are affected.

Did you read [my comment](https://github.com/ibis-project/ibis/pull/2905#issuecomment-925191316)?

cpcloud: > And I think it's better than this magic failing the test.

Are you referring to the regular expression splitting?

> Just the explicit syntax doesn't seem to work in Impala

Are you sure about that? Here's an example that seems to work just fine:

```
(Pdb) list
265
266         table = con.parquet_file(hdfs_path)
267
268         name = table.op().name
269         breakpoint()
270  ->     assert con.exists_table(name) is True
271         con.close()
272
273         assert not con.exists_table(name)
274
275
(Pdb) con.list_tables('__ibis_tmp_81c73096e0924eafba359d2a1de85f85', database='__ibis_tmp')
['__ibis_tmp_81c73096e0924eafba359d2a1de85f85']
```

datapythonista: Mhh, I see, you're right. I didn't fully understand the problem, now it's clear. Then, what's different in Impala is that other backends don't support fully qualified table names, as Impala handles in `list_tables`. Not sure why the test wasn't working when I splittted the table myself in the test, I guess I missed something.

In any case, I think this PR is fine now. Or is there anything else you would change?

cpcloud:
I think this is good to go! @datapythonista Thanks for putting up with my reviews!

cpcloud (approved):

jreback (approved):
lgtm. can you rebase & also i think we have a link somewhere for the deprecated apis? pls add to that

datapythonista:
Good point, added to #2863 and rebased

cpcloud:
thanks @datapythonista!

PR #2908, opened 2021-08-20
Support summarization of empty data in Pandas backend

emilyreff7:
### Proposed Change
Handle empty input data in summarization.

Currently, when a summarization gets passed empty data in the Pandas backend, we error out with a confusing 
```
UnboundLocalError: local variable 'grouped_col_name' referenced before assignment
```

This PR adds checks in a couple of places for empty data, such that we gracefully return an empty DataFrame instead.

### Tests
Added a new test for this in test_vectorized_udf.py.

### DIFF
--- a/ibis/expr/schema.py
+++ b/ibis/expr/schema.py
@@ -370,7 +370,7 @@ def coerce_to_dataframe(
         result = data
     elif isinstance(data, pd.Series):
         if not len(data):
-            result = pd.concat([data], axis=1)
+            result = data.to_frame()
         else:
             num_cols = len(data.iloc[0])
             series = [data.apply(lambda t: t[i]) for i in range(num_cols)]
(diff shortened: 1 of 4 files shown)


ibis/expr/schema.py
```diff
@@ -369,9 +369,12 @@ def coerce_to_dataframe(
     if isinstance(data, pd.DataFrame):
         result = data
     elif isinstance(data, pd.Series):
-        num_cols = len(data.iloc[0])
-        series = [data.apply(lambda t: t[i]) for i in range(num_cols)]
-        result = pd.concat(series, axis=1)
+        if not len(data):
+            result = pd.concat([data], axis=1)
```
datapythonista, line 373: Why not `result = data.to_frame()`?

datapythonista (approved):
lgtm

ibis/expr/schema.py
emilyreff7: Good point, updated to use `to_frame` instead.

icexelloss (approved):
LGTM

datapythonista:
Thanks @emilyreff7

PR #2910, opened 2021-08-21
CLN: Standardize list_databases and exists_database and move to Backend

datapythonista:
xref #2678

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import abc
+import re
 import warnings
 from typing import Any, Callable, List, Type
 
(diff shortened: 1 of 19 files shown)


datapythonista:
@jreback if you are still connected and want to have a quick look at this. This is very similar to all other PRs I'm working on. What it does is:


- Move `list_databases` from `Client` to `Backend` (when everything is moved to Backend, I'll delete `Client` and make `connect` return a `Backend` class
- Added to the backends that didn't implement it, for consistency
- Make the signature consistent. It has in some cases a `lke` parameter to filter databases. Added where it was missing, and ensure that `like` always filters in the same way.
- Added docs and tests for `list_databases`

Also here, standardized `exists_database` was in couple of backends, I add a deprecation warning since doesn't seem very useful (can just check `db in list_databases()` to see if it exists. I also move it to `Backend`.

In the other PRs I do the same as in `list_databases`, but for `list_tables`, `set_database`, `current_database`, and will be adding the rest of funcionality. Any feedback here will be useful for the rest, since it's mostly the same all the time.

jreback (approved):
looks fine

PR #2914, opened 2021-08-23
CLN: Remove overcomplicatted logging

datapythonista:
Some backends implement `Client.log`, which doesn't seem useful at all. I guess the original reason was to be able to overwrite the method, so the logging could be replaced. I don't think `Client` is subclassed for any of the backends that provide the `log` method. And also, the provided implementation of the logging uses `options.verbose_log`, which can be set to change the logging in an easier way than overwritting the method.

Removing those `log` methods here, which are inconsistent among backends, don't provided any feature, and polute the public `Client` API.


jreback:
sure can you add a release note as this is technically public.

cpcloud:
@datapythonista seems like this is ready for merge.

jreback (approved):

PR #2916, opened 2021-08-26
FEAT: ClickHouse: support joining on different columns.

Lunaticus7:
It seems that for some historical reasons ClickHouse backend banned joining on different columns, and that has been supported by ClickHouse now. So i removed the vailidation and added some codes for that (refer to ibis SQL backend).

### DIFF
--- a/ibis/backends/clickhouse/compiler.py
+++ b/ibis/backends/clickhouse/compiler.py
@@ -84,14 +84,35 @@ class ClickhouseTableSetFormatter(TableSetFormatter):
             self.join_types, self.join_tables[1:], self.join_predicates
         ):
             buf.write('\n')
-            buf.write(util.indent('{0} {1}'.format(jtype, table), self.indent))
+            buf.write(util.indent(f'{jtype} {table}', self.indent))
 
-            if len(preds):
+            fmt_preds = []
+            npreds = len(preds)
+            using_clause_flag = True
+            for pred in preds:
+                op = pred.op()
+                if (
+                    not isinstance(op, ops.Equals)
+                    or op.args[0].get_name() != op.args[1].get_name()
+                ):
+                    using_clause_flag = False
+                new_pred = self._translate(pred)
+                if npreds > 1:
+                    new_pred = f'({new_pred})'
+                fmt_preds.append(new_pred)
+
+            if npreds > 0:
                 buf.write('\n')
-                fmt_preds = map(self._format_predicate, preds)
-                fmt_preds = util.indent(
-                    'USING ' + ', '.join(fmt_preds), self.indent * 2
-                )
+                if using_clause_flag:
+                    fmt_preds = map(self._format_predicate, preds)
+                    fmt_preds = util.indent(
+                        'USING ' + ', '.join(fmt_preds), self.indent * 2
+                    )
+                else:
+                    conj = f' AND\n{" " * 3}'
+                    fmt_preds = util.indent(
+                        'ON ' + conj.join(fmt_preds), self.indent * 2
+                    )
                 buf.write(fmt_preds)
 
         return buf.getvalue()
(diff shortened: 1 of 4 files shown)


ibis/backends/clickhouse/tests/test_select.py
```diff
@@ -265,43 +267,43 @@ def test_non_equijoin(alltypes):
...
-    ],
+    ('join_type_and_clause', 'join_keys'),
+    product(
+        [
+            ('any_inner_join', 'ANY INNER JOIN'),
+            ('inner_join', 'ALL INNER JOIN'),
+            ('any_left_join', 'ANY LEFT JOIN'),
+            ('left_join', 'ALL LEFT JOIN'),
+        ],
+        [('playerID',), ('playerID', 'awardID'),],  # noqa: E231
```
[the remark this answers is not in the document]
cpcloud: They serve different purposes. `black` is for formatting, `flake8` will lint a number of things. The flake8 configuration is probably not maximally black compatible. There are also scenarios where black and flake8 will conflict, even if configured for maximum compatibility. Chalk it up to the Python community having 10 different tools and conventions for formatting code.

ibis/backends/clickhouse/tests/test_select.py
```diff
@@ -265,43 +267,43 @@ def test_non_equijoin(alltypes):
...
+        f'  {join_clause} {db.name}.`awards_players` t1\n'
+        f'{join_keys_str}'
     )
 
     assert ibis.clickhouse.compile(expr) == expected
-    assert len(con.execute(expr))
+    try:
+        con.execute(expr)
+    except clickhouse_driver.errors.ServerException:
+        pytest.fail('ClickHouse raise a `ServerException` error.')
```
cpcloud, line 306: Consider letting this fail, or if it must not fail then at least format the exception into the `fail` strings. It's nice to know that a `ServerException` was raised, but it's even better to know what that exception is.

ibis/backends/clickhouse/compiler.py
```diff
@@ -84,14 +84,35 @@ def get_result(self):
...
+            fmt_preds = []
+            npreds = len(preds)
+            using_clause_flag = True
+            for pred in preds:
+                op = pred.op()
+                if (
+                    not isinstance(op, ops.Equals)
+                    or op.args[0].get_name() != op.args[1].get_name()
+                ):
+                    using_clause_flag = False
```
cpcloud, line 98: Can we avoid using the `USING` keyword here? It's adding a ton of complexity to this code and if it isn't necessary then we shouldn't use it.

ibis/backends/clickhouse/registry.py
```diff
@@ -471,11 +471,10 @@ def _table_column(translator, expr):
...
 
-    # TODO(kszucs): table aliasing is partially supported
-    # if ctx.need_aliases():
-    #     alias = ctx.get_ref(table)
-    #     if alias is not None:
-    #         quoted_name = '{0}.{1}'.format(alias, quoted_name)
+    if ctx.need_aliases():
+        alias = ctx.get_ref(table)
+        if alias is not None:
+            quoted_name = '{0}.{1}'.format(alias, quoted_name)
```
cpcloud, line 477: Use an f-string here since you're already touching this code

cpcloud (changes requested):
Looks mostly good, except for the `USING` comment.

ibis/backends/clickhouse/compiler.py
Lunaticus7: When stop using the `USING` keyword, there is no need to override this method, so i remove it and some othere duplicated codes.

ibis/backends/clickhouse/tests/test_select.py
Lunaticus7: I have remove the try caluse.

ibis/backends/clickhouse/tests/test_select.py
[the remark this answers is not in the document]
Lunaticus7: thanks for explaining.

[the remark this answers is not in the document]
cpcloud: For sure!

ibis/backends/clickhouse/compiler.py
cpcloud: Awesome. Loving the code deletion.

cpcloud (approved):
Net negative lines of code. Ship it.

PR #2903, opened 2021-08-11
CLN: Remove unused and inconsistent backend markers

datapythonista:
Markers for backends don't seem to add any value, since they are in separate directories, and we call them explicitly. I don't think there is any use case where the marker is useful, surely not excluding a backend from tests, since tests for a backend need to be added by path (e.g. `pytest ibis/tests ibis/backends/pandas`).

Also, in case we wanted to use them, they are not very reliable, some backends don't use them, some use them in only some of the tests...

Better to remove them to simplify things and avoid noise.

PR #2918, opened 2021-08-30
feat(datafusion): initial implementation for Arrow Datafusion backend 

kszucs:
The main purpose if this PR is to share the work on the datafusion backend. Basic things are working, we need to add more operations (depending on the datafusion support) and additional backend specific tests.

The CI hasn't been configured yet, since it requires two PRs to be merged: 
- https://github.com/apache/arrow-rs/pull/691 
- https://github.com/apache/arrow-datafusion/pull/873

The parametrized testing suite has the following state:
```
 2 failed, 71 passed, 230 skipped, 261 xfailed, 1 warning
```

cc @cpcloud


ibis/backends/datafusion/tests/test_client.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+import pyarrow as pa
+
+from .. import Backend
+from ..client import DataFusionTable  # noqa: E402
```
cpcloud, line 6: I realize this is probably copied from another file, but relative imports in tests are a bad idea generally. Tests should import the library as a user would, and a user isn't going to be using imports relative to ibis when they are using ibis.

ibis/expr/types.py
```diff
@@ -421,6 +421,21 @@ class ScalarExpr(ValueExpr):
...
         return str(self.type())
 
+    def to_projection(self):
+        """
+        Promote this column expression to a table projection
+        """
+        roots = self._root_tables()
+        if len(roots) > 1:
+            raise com.RelationError(
+                'Cannot convert array expression '
```
cpcloud, line 431: Should this be `scalar expression` since this method is defined on `ScalarExpr`?

ibis/backends/datafusion/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+        config : DataFusionContext or dict
+
+        Returns
+        -------
+        DataFusionClient
+        """
+        return self.client_class(backend=self, config=config)
+
+    def current_database(self):
+        raise NotImplementedError("")
```
cpcloud, line 42: Should we go ahead and make `BaseBackend` an abstract base class so you don't have to have pointless boilerplate here?

ibis/backends/datafusion/client.py
```diff
@@ -0,0 +1,204 @@
...
+            self.context = config
+        else:
+            self.context = df.ExecutionContext()
+            for name, path in config.items():
+                if path.endswith('.csv'):
+                    self.register_csv(name, path)
+                elif path.endswith('.parquet'):
+                    self.register_parquet(name, path)
+                else:
+                    raise ValueError('Wrong format')
```
cpcloud, line 101: Maybe something a little more informative like `"The DataFusion backend only supports CSV files with the extension .csv and Parquet files with the .parquet extension"`

ibis/backends/datafusion/client.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def register_csv(self, name, path, schema=None):
+        self.context.register_csv(name, path, schema=schema)
+
+    def register_parquet(self, name, path, schema=None):
+        self.context.register_parquet(name, path, schema=schema)
+
+    # def database(self, name='public'):
```
cpcloud, line 109: Maybe just delete this code?

ibis/backends/datafusion/client.py
```diff
@@ -0,0 +1,204 @@
...
+    def list_tables(self, like=None):
+        """List the available tables."""
+        tables = list(self.context.tables())
+        if like is not None:
+            pattern = re.compile(like)
+            return list(filter(lambda t: pattern.findall(t), tables))
+        return tables
+
+    def list_databases(self, like=None):
+        raise com.NotImplementedError()
```
cpcloud, line 131: Why is `NotImplementedError` coming from `com`?

ibis/backends/datafusion/client.py
```diff
@@ -0,0 +1,204 @@
...
+        return bool(self.list_tables(like=name))
+
+    def execute(
+        self,
+        expr: ir.Expr,
+        params: Mapping[ir.Expr, object] = None,
+        limit: str = 'default',
+        **kwargs,
+    ):
+        def _collect(frame):
```
cpcloud, line 155: Maybe make this a module-level function since it's not using any outer scope.

ibis/backends/datafusion/client.py
```diff
@@ -0,0 +1,204 @@
...
+            )
+
+    def compile(
+        self, expr: ir.Expr, params: Mapping[ir.Expr, object] = None, **kwargs
+    ):
+        """Compile `expr`.
+
+        Notes
+        -----
+        For the dask backend returns a dask graph that you can run ``.compute``
```
cpcloud, line 199: Is the note about dask here relevant?

ibis/backends/datafusion/tests/conftest.py
```diff
@@ -0,0 +1,66 @@
...
+                ('double_col', 'float64'),
+                ('date_string_col', 'string'),
+                ('string_col', 'string'),
+                ('timestamp_col', 'string'),
+                ('year', 'int64'),
+                ('month', 'int64'),
+            ])
+        )
+        client.register_csv(
+            name='batting',
```
cpcloud, line 51: I believe there are also parquet equivalents here, maybe we should use parquet for at least one of these datasets?

ibis/backends/tests/conftest.py
```diff
@@ -143,18 +143,21 @@ def pytest_pyfunc_call(pyfuncitem):
...
-            backend, BackendTest
-        ), "backend has type {!r}".format(type(backend).__name__)
-        pytest.xfail(reason='{}: {}'.format(type(backend).__name__, e))
+        markers = list(pyfuncitem.iter_markers(name="xfail_unsupported"))
+
+        if len(markers) == 0:
+            # nothing has marked the failure as an expected one
+            raise e
+        elif len(markers) == 1:
+            pytest.xfail(reason='{}: {}'.format(type(backend).__name__, e))
```
cpcloud, line 153: Might as well take advantage of f-strings here.

ibis/backends/tests/conftest.py
```diff
@@ -143,18 +143,21 @@ def pytest_pyfunc_call(pyfuncitem):
...
+        if len(markers) != 1:
+            (marker,) = markers
+            if not isinstance(backend, BackendTest):
+                pytest.fail("Backend has type {!r}".format(type(backend).__name__))
```
cpcloud, line 157: f-strings here too.

ibis/backends/tests/conftest.py
```diff
@@ -143,18 +143,21 @@ def pytest_pyfunc_call(pyfuncitem):
...
+        else:
+            msg = "More than one xfail_unsupported marker found on test {}"
```
cpcloud, line 159: and ... more f-strings :)

ibis/backends/tests/test_numeric.py
```diff
@@ -325,6 +325,7 @@ def test_mod(backend, alltypes, df):
     backend.assert_series_equal(result, expected, check_dtype=False)
 
 
+@pytest.mark.xfail_unsupported
```
cpcloud, line 328: Were these code paths just never hit?

ibis/expr/types.py
```diff
@@ -611,6 +626,8 @@ def get_column(self, name):
         ref = ops.TableColumn(name, self)
         return ref.to_expr()
 
+    # should be renamed to column_names and columns should return
+    # with the TableColumn expressions
```
cpcloud, line 630: Maybe? `columns` seems like an API that would be pretty egregious to break, since most people will probably do the following if `.columns` returned a sequence of `TableColumn`:

1. Run `.columns`
2. Realize that's not what they want
3. Run `[c.name for c in t.columns]`

setup.cfg
```diff
@@ -38,6 +38,10 @@ xfail_strict = true
 addopts = --strict-markers
 markers =
     backend
+    clickhouse
+    csv
+    dask
+    datafusion
```
cpcloud, line 44: Shouldn't these be causing CI to fail right now if they're actually necessary?

cpcloud (changes requested):

ibis/backends/datafusion/__init__.py
datapythonista: There was some inconsistent behavior with this until #2911. The options I considered were returning `None` by default, and let backends supporting different databases overwrite it. And requiring this, expecting all backends to support this, even if only one database is supported (this can return `base` or `main`). SQLAlchemy is using the latter, and I think it makes things simpler (including tests and type annotations). This way we can garantee code like `assert client.current_database in client.list_databases()` to work for all backends, which I think it's nice for users and keeps things simpler than having to handle `None` and special cases.

Happy to discuss further (in a new issue), but unless we change that, you can return `main` or whatever here, and `['main']` in `list_databases()`.

ibis/backends/datafusion/client.py
datapythonista: For other backends, this already warns with a `FutureWarning`, and the plan is to deprecate it, and make things simpler.

ibis/backends/datafusion/client.py
```diff
@@ -0,0 +1,206 @@
...
+
+class DataFusionTable(ops.DatabaseTable):
+    pass
+
+
+class DataFusionDatabase(ops.DatabaseTable):
+    pass
+
+
+class DataFusionClient(Client):
```
datapythonista, line 89: Several of the methods defined here have already been moved to `Backend` since you started working on this. And the plan is to move `Client` completely and move everything there.

ibis/backends/datafusion/tests/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+    bool_is_int = True
+
+    @staticmethod
+    def connect(data_directory: Path):
+        # can be various types:
+        #   pyarrow.RecordBatch
+        #   parquet file path
+        #   csv file path
+        client = ibis.datafusion.connect({})
+        client.register_csv(
```
datapythonista, line 24: I wonder if we should deprecate the csv, hdf5 and parquet "backends" and have a standard system to register files that can be used in the pandas backend, as you are doing here.

ibis/backends/datafusion/tests/test_client.py
```diff
@@ -0,0 +1,39 @@
+import pyarrow as pa
+import pytest
+
+from .. import Backend
+
+pytestmark = pytest.mark.datafusion
+
+
+@pytest.fixture
+def client(data_directory):
```
datapythonista, line 10: I've seen this in other backends, but isn't it enough `TestConf.connect`?

setup.cfg
datapythonista: Those have been restored from a git merge, they were removed in #2903

datapythonista (commented):
Any reason to not create this backend in a separate repo (e.g. `github.com/ibis-project/ibis-datafusion`). I think it'll keep things simpler, and will allow to release ibis-datafusion more often than the ibis core if needed. It'll also help review this, since the changes required to Ibis itself, and the new backend will be in different PRs.

ibis/backends/datafusion/client.py
cpcloud: What is the rationale for deprecating this method?

datapythonista: I think the `Database` class doesn't add much value. Its methods are duplicated in `Client`, and just makes things more complex.

For example, I think we can just have:
```python
con = ibis.sqlite.connect(...)
con.drop_database('foo')
```

and this doesn't seem necessary to me:
```python
con = ibis.sqlite.connect(...)
db = con.database('foo')
db.drop()
```

Does it sound reasonable?

cpcloud: Sure, in the drop case that makes sense, but there are other use cases like cross-database joins where I think an object is warranted.

Generally agree that there's too much duplication of APIs though, so it's fine IMO to pare things down.

In that case we shouldn't be afraid to introduce breaking changes and leverage semantic versioning instead of deprecating (and removing the API at some arbitrarily later date).

datapythonista: I think cross-database joins should be possible with `con.table('foo', database='bar')`, right? If I'm missing something, happy to reconsider, but for the use cases I'm aware of, the whole `Database` class is just duplication of `Client`.

If there is some special case where is needed, I think making `.database()` return an instance of `Client` would make more sense.

Regarding deprecation, my plan is to show the `FutureWarning` in Ibis 2.* and have the feature removed in Ibis 3.*. Happy to discuss and hear other options if you have better ideas.

cpcloud: Why not remove it when bumping to version 2? That's a major version bump, and that's where a breaking API change is allowed and should occur.

datapythonista: The idea (and my understanding on how semver is supposed to be implemented) is next:

If an Ibis 1.* user has this code:
```python
import ibis

con = ibis.sqlite.connect(...)
db = con.database()
```

This is fine in Ibis 1.* (works and no warning is issued). If we do what you say, and the user uprades Ibis and runs this code in Ibis 2.*, the code won't work.

If instead, we add a `FutureWarning`, when the user upgrades to Ibis 2.0, the code will still work, but will start raising the warning. When the user wants to upgrade to 3.0, they should know that the code will still work, unless there is any `FutureWarning` in the logs.

Do you think we should do this for minor instead of major releases?

kszucs:
Steps required before proceeding with this PR:
- need to wait for arrow-rs 6.0.0 release (~2/3 weeks)
- merge the arrow-datafusion PR
- finalize and merge this PR

ibis/backends/datafusion/__init__.py
kszucs: Updated to use the new backend abstract class which is way nicer to use!

ibis/backends/datafusion/client.py
kszucs: Removed client.py

ibis/backends/datafusion/tests/conftest.py
kszucs: Definitely!

ibis/backends/tests/test_api.py
```diff
@@ -6,6 +6,7 @@ def test_backend_name(backend):
     assert backend.api.name == backend.name()
 
 
+@pytest.mark.skip_backends(['datafusion'])
```
kszucs, line 9: `datafusion.__version__` doesn't exist

ibis/backends/tests/test_generic.py
```diff
@@ -166,9 +166,9 @@ def test_notin(backend, alltypes, sorted_df, column, elements):
         (lambda t: ~t['bool_col'], lambda df: ~df['bool_col']),
     ],
 )
-@pytest.mark.skip_backends(['dask'])  # TODO - sorting - #2553
+@pytest.mark.skip_backends(['dask', 'datafusion'])  # TODO - sorting - #2553
```
kszucs, line 169:
```python
# bool_col is an integer here, convert it to be actually boolean
bool_col = (df.column("bool_col") == df.literal(1)).alias("bool_col")

t.select(bool_col).filter(df.column("bool_col")).collect()
```

raises

```
Exception: Error during planning: Ambiguous reference to field named 'bool_col'
```

ibis/backends/tests/test_generic.py
```diff
@@ -166,9 +166,9 @@ def test_notin(backend, alltypes, sorted_df, column, elements):
...
 @pytest.mark.xfail_unsupported
-def test_filter(backend, alltypes, sorted_df, predicate_fn, expected_fn):
+def test_filter(con, backend, alltypes, sorted_df, predicate_fn, expected_fn):
```
kszucs, line 171: Used for pdb, remove `con` argument.

ibis/backends/tests/test_temporal.py
```diff
@@ -358,7 +358,7 @@ def test_temporal_binop_invalid_interval_unit(con, alltypes):
     ],
 )
 @pytest.mark.xfail_unsupported
-@pytest.mark.skip_backends(['spark', 'sqlite'])
+@pytest.mark.skip_backends(['spark', 'sqlite', 'datafusion'])
```
kszucs, line 361: Same as https://github.com/ibis-project/ibis/pull/2918/files#r749318005

pyproject.toml
```diff
@@ -38,6 +38,8 @@ dask = { version = "^2021.2.0", optional = true, extras = [
   "array",
   "dataframe"
 ] }
+# TODO(kszucs): pin datafusion's version
+# datafusion = { verion = ">=0.4", optional = true }
```
kszucs, line 42: The datafusion release should happen soon.

ibis/backends/tests/test_generic.py
houqp: @kszucs i think this is a bug in datafusion, `.filter(df.column("bool_col"))` should ignore the projection alias and only use the actual `bool_col` table column to evaluate the predicate. This is be behavior I see in postgres. For example `SELECT id+1 as id2 FROM test where id2 > 2` results in `error: column "id2" does not exist`.

Do you mind filing a datafusion issue so we can dig into it?

kszucs: Thanks @houqp! Going to create an issue.

kszucs:
@cpcloud could you please give it a review? I'd like to finalize it and merge as an experimental backend, then create follow-up issues to add support for missing operations (windows, joins etc.).

.github/workflows/ibis-backends.yml
```diff
@@ -425,3 +425,90 @@ jobs:
...
+      - name: install python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pypoetry
+          key: datafusion-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('poetry.lock') }}
```
cpcloud, line 468: You need to cache with the specific dependency coming out of the `install python` step to ensure the cache is invalidated for minor Python version upgrades. We do this in `ibis-main.yml`. Check there for examples.

.github/workflows/ibis-backends.yml
```diff
@@ -425,3 +425,90 @@ jobs:
...
+
+      - name: create virtualenv
+        run: |
+          pip install virtualenv
+          virtualenv venv
```
cpcloud, line 473: Is this virtualenv stuff necessary?

ibis/backends/datafusion/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+# consolidate with the logic from the parquet backend
+
+
+class Backend(BaseBackend):
+    name = 'datafusion'
+    builder = None
+
+    @property
+    def version(self):
+        dist = pkg_resources.get_distribution("datafusion")
```
cpcloud, line 27: Use `importlib.metadata`/`importlib_metadata` for this. `pkg_resources` is slow and unnecessary at this point.

cpcloud: probably can use `Backend.name` to load this as well.

ibis/backends/datafusion/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+                # to a selection which we can directly convert to a datafusion
+                # plan
+                expr = expr.name('tmp').to_projection()
+                frame = self.compile(expr, params, **kwargs)
+            else:
+                # doesn't have any tables associated so create a plan from a
+                # dummy datafusion table
+                compiled = self.compile(expr, params, **kwargs)
+                frame = self.context.empty_table().select(compiled)
+            return _collect(frame).iloc[0, 0]
```
cpcloud, line 124: Does this return a numpy scalar or a python value? I think we should return python values as much as possible, since they are much more common than numpy scalars in external APIs.

ibis/backends/datafusion/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+            )
+
+    def compile(
+        self, expr: ir.Expr, params: Mapping[ir.Expr, object] = None, **kwargs
+    ):
+        """Compile `expr`.
+
+        Notes
+        -----
+        For the dask backend returns a dask graph that you can run ``.compute``
```
cpcloud, line 137: Is this correct?

ibis/backends/datafusion/datatypes.py
```diff
@@ -0,0 +1,87 @@
...
+    dt.UInt16: pa.uint16(),
+    dt.UInt32: pa.uint32(),
+    dt.UInt64: pa.uint64(),
+    dt.Float16: pa.float16(),
+    dt.Float32: pa.float32(),
+    dt.Float64: pa.float64(),
+    dt.String: pa.string(),
+    dt.Binary: pa.binary(),
+    dt.Boolean: pa.bool_(),
+    dt.Timestamp: pa.timestamp('ns'),
```
cpcloud, line 63: I don't think we have a unit defined for `dt.Timestamp`, intentionally so. I guess some choice is necessary even though it's irrelevant for expressing queries :disappointed:

ibis/backends/datafusion/datatypes.py
```diff
@@ -0,0 +1,87 @@
...
+def to_pyarrow_type(dtype):
+    return _to_pyarrow_types[dtype.__class__]
+
+
+@to_pyarrow_type.register(dt.Array)
+def from_ibis_array(dtype):
+    return pa.list_(to_pyarrow_type(dtype.value_type))
+
+
+@to_pyarrow_type.register(dt.Set)
```
cpcloud, line 77: I honestly do not know what the purpose of the `Set` operation is. We should remove it.

ibis/backends/datafusion/tests/conftest.py
cpcloud: This seems like a good idea. There are also a growing number existing and potential backends that support local file consumption: datafusion, duckdb, pyarrow, clickhouse, etc.

ibis/backends/datafusion/tests/test_udf.py
```diff
@@ -0,0 +1,38 @@
...
+def client(data_directory):
+    return ibis.datafusion.connect(
+        {"functional_alltypes": data_directory / 'functional_alltypes.csv'}
+    )
+
+
+@elementwise(input_type=['string'], output_type='int64')
+def my_string_length(arr, **kwargs):
+    # arr is a pyarrow.StringArray
+    return pc.cast(pc.multiply(pc.utf8_length(arr), 2), target_type='int64')
```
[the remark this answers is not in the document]
cpcloud: This is ... really awesome.

ibis/backends/tests/conftest.py
cpcloud: fstrings?

ibis/backends/tests/test_generic.py
cpcloud: What happened with this?

ibis/backends/tests/test_numeric.py
cpcloud: bump

ibis/expr/types.py
cpcloud: bump

ibis/expr/types.py
cpcloud: not sure if this comment should remain, i'm not sure what it conveys except that we should improve ibis :)

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dask = { version = "^2021.2.0", optional = true, extras = [
   "array",
   "dataframe"
 ] }
+datafusion = { version = ">=0.4", optional = true }
```
cpcloud, line 41: Please give this an upper bound, especially since this is pre 1.0. Unbounded upper bounds make it really really easy to start failing CI for hard-to-debug -reasons.

ibis/backends/datafusion/tests/test_udf.py
```diff
@@ -0,0 +1,69 @@
...
+    expr = my_add(t.smallint_col, t.int_col)
+    result = expr.execute()
+
+    df = t[['smallint_col', 'int_col']].execute()
+    expected = df.smallint_col + df.int_col
+
+    tm.assert_series_equal(result, expected.rename('tmp'))
+
+
+# def test_udaf(client):
```
cpcloud, line 59: Kill this code.

ibis/backends/datafusion/compiler.py
```diff
@@ -0,0 +1,410 @@
...
+                if column.has_name():
+                    field = field.alias(column.get_name())
+                selections.append(field)
+        elif isinstance(expr, ir.ValueExpr):
+            field = translate(expr)
+            if expr.has_name():
+                field = field.alias(expr.get_name())
+            selections.append(field)
+        else:
+            raise com.IbisError("TODO")
```
cpcloud, line 96: Needs a better error message.

ibis/backends/datafusion/compiler.py
```diff
@@ -0,0 +1,410 @@
...
+
+    plan = plan.select(*selections)
+
+    if op.predicates:
+        predicates = map(translate, op.predicates)
+        predicate = functools.reduce(operator.add, predicates)
```
cpcloud, line 102: This should be `and`, not `add` :sweat_smile:

cpcloud: Probably needs a test too, if this isn't failing.

ibis/backends/datafusion/compiler.py
```diff
@@ -0,0 +1,410 @@
...
+@translate.register(ops.ValueList)
+def value_list(op, expr):
+    return list(map(translate, op.values))
+
+
+@translate.register(ops.Contains)
+def contains(op, expr):
+    value = translate(op.value)
+    options = _prepare_contains_options(op.options)
+    return df.functions.in_list(value, options, False)
```
cpcloud, line 390: does the last argument `False` have a meaningful keyword argument? It's hard to see what the meaning of "a third boolean argument" is without reading the documentation. A keyword argument should help here.

ibis/backends/datafusion/compiler.py
```diff
@@ -0,0 +1,410 @@
...
+
+
+@translate.register(ops.NotContains)
+def not_contains(op, expr):
+    value = translate(op.value)
+    options = _prepare_contains_options(op.options)
+    return df.functions.in_list(value, options, True)
```
cpcloud, line 397: Same comment as above: kwarg the `True` :)

ibis/backends/datafusion/compiler.py
```diff
@@ -0,0 +1,410 @@
...
+
+@translate.register(ops.Substring)
+def substring(op, expr):
+    arg = translate(op.arg)
+    start = translate(op.start + 1)
+    length = translate(op.length)
+    return df.functions.substr(arg, start, length)
+
+
+# @translate.register(ops.RegexSearch)
```
cpcloud, line 250: Delete this code.

ibis/backends/datafusion/compiler.py
```diff
@@ -0,0 +1,410 @@
...
+    return df.functions.ceil(arg).cast(pa.int64())
+
+
+@translate.register(ops.Floor)
+def floor(op, expr):
+    arg = translate(op.arg)
+    return df.functions.floor(arg).cast(pa.int64())
+
+
+# TODO(kszucs): datafusion implementation of signum considers zero as positive
```
cpcloud, line 151: Delete this code.

cpcloud (changes requested):

cpcloud:
@kszucs When you have a chance can you rebase and resolve the conflicts?

setup.py
```diff
@@ -87,9 +90,10 @@
         "clickhouse-sqlalchemy>=0.1.4,<0.2",
         "sqlalchemy>=1.3,<1.4",
     ],
-    "dask": ["dask[array,dataframe]>=2021.2.0,<2022.0.0", "pyarrow>=1,<7"],
+    "dask": ["dask[dataframe,array]>=2021.2.0,<2022.0.0", "pyarrow>=1,<7"],
+    "datafusion": ["datafusion>=0.4"],
     "geospatial": [
-        "geoalchemy2>=0.6,<0.11",
+        "geoalchemy2>=0.6,<0.10",
```
cpcloud, line 96: This doesn't look right. We increased the upper bound recently to allow `0.10`.

ibis/backends/datafusion/compiler.py
cpcloud: bump

ibis/backends/datafusion/compiler.py
cpcloud: bump

ibis/backends/datafusion/compiler.py
cpcloud: bump

ibis/backends/datafusion/compiler.py
cpcloud: bump

ibis/backends/datafusion/compiler.py
cpcloud: bump

ibis/backends/datafusion/compiler.py
cpcloud: bump

ibis/backends/datafusion/compiler.py
```diff
@@ -0,0 +1,410 @@
...
+    return df.functions.in_list(value, options, True)
+
+
+@translate.register(ops.ElementWiseVectorizedUDF)
+def elementwise_udf(op, expr):
+    udf = df.udf(
+        op.func,
+        input_types=list(map(to_pyarrow_type, op.input_type)),
+        return_type=to_pyarrow_type(op.return_type),
+        volatility="volatile",
```
cpcloud, line 406: Guessing this is similar to SQLite's `SQLITE_DETERMINISTIC`, where `"volatile"` here means "possible impure"?

ibis/expr/types.py
cpcloud: bump

ibis/backends/datafusion/client.py
kszucs: Updated to conform with the new backend class.

pyproject.toml
```diff
@@ -100,6 +102,7 @@ all = [
 clickhouse = ["sqlalchemy", "clickhouse-sqlalchemy", "clickhouse-driver"]
 csv = []
 dask = ["dask", "pyarrow"]
+datafusion = ["datafusion"]
```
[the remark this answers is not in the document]
kszucs: Yes, do we need to define "pyarrow" as well? Datafusion should bring it as a dependency.

.github/workflows/ibis-backends.yml
kszucs: `maturin` requires a virtualenv to install the datafusion python bindings from source

.github/workflows/ibis-backends.yml
cpcloud: Following up here, you need to assign the `setup-python` step an `id` (like `install_python`) and then use `${{ steps.install_python.outputs.python-version }}` instead of `${{ matrix.python-version }}` here, otherwise the cache won't be invalidated when a new patch version of Python shows up (e.g., 3.9.8 -> 3.9.9) and this will break in a hard-to-debug way.

ibis/backends/datafusion/__init__.py
cpcloud: bump ... I don't think dask is in play here right?

.github/workflows/ibis-backends.yml
```diff
@@ -447,3 +447,90 @@ jobs:
...
+
+  test_datafusion:
+    name: DataFusion ${{ matrix.datafusion-version }} ubuntu-latest python-${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        datafusion-version:
+          - master
+          - latest
```
cpcloud, line 459: Do we need to test both or would `latest` suffice?

ibis/backends/datafusion/compiler.py
kszucs: Updated.

ibis/backends/datafusion/compiler.py
kszucs: Updated.

ibis/backends/datafusion/__init__.py
kszucs: Updated.

.github/workflows/ibis-backends.yml
kszucs: I'd prefer to test using both since we'll need to develop additional features to the datafusion python bindings to improve the support for ibis operators.

cpcloud:
It looks like you didn't set `PYTHONHASHSEED=42` when modifying `setup.py` (https://github.com/ibis-project/ibis/runs/4666725786?check_suite_focus=true). This is documented here https://ibis-project.org/contribute.html

.github/workflows/ibis-backends.yml
cpcloud: SGTM.

pyproject.toml
cpcloud: bump, this needs an upper bound

pyproject.toml
```diff
@@ -39,7 +39,8 @@ dask = { version = "^2021.2.0", optional = true, extras = [
   "array",
   "dataframe"
 ] }
-geoalchemy2 = { version = ">=0.6,<0.11", optional = true }
+datafusion = { version = ">=0.4", optional = true }
+geoalchemy2 = { version = ">=0.6,<0.10", optional = true }
```
cpcloud, line 43: Why the upper bound decrease?

pyproject.toml
```diff
@@ -39,7 +39,8 @@ dask = { version = "^2021.2.0", optional = true, extras = [
   "array",
   "dataframe"
 ] }
-geoalchemy2 = { version = ">=0.6,<0.11", optional = true }
+datafusion = { version = ">=0.4", optional = true }
```
cpcloud, line 42: You need to add `maturin` as a development dependency so that the nix development environment works. This is precisely why the nix setup exists, to catch underspecified dependencies :)

You can do this with `poetry add maturin --dev`

pyproject.toml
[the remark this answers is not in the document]
cpcloud: Cool, then it's fine as is.

ibis/backends/datafusion/compiler.py
kszucs: Nice catch! Updated and added a test case.

ibis/backends/datafusion/tests/test_udf.py
kszucs: Removed.

ibis/backends/datafusion/datatypes.py
kszucs: Agreed.

ibis/backends/datafusion/__init__.py
kszucs: Updated.

ibis/backends/datafusion/tests/test_client.py
kszucs: It is, updated.

ibis/backends/tests/test_numeric.py
kszucs: Removed

pyproject.toml
kszucs: Rebase mistake.

kszucs: Updated.

setup.py
kszucs: Updated.

ibis/backends/datafusion/__init__.py
kszucs: Updated to use pyarrow scalars and then convert them to python values.

pyproject.toml
kszucs: Updated.

cpcloud:
@kszucs All passing except for `commitlint`!

jreback (changes requested):
is this mentioned in the docs?
ok as a follow up (certainly can be marked experimental)

cpcloud (approved):

cpcloud:
@kszucs If you rebase and squash we can merge this for the 2.1 release, which we should cut next week, after #3218 is merged.

ibis/backends/datafusion/compiler.py
kszucs: Yes.

ibis/expr/types.py
kszucs: Updated.

ibis/backends/datafusion/compiler.py
kszucs: Updated.

cpcloud:
@kszucs rebase one more time and i'll merge this 🙏🏻

cpcloud (approved):

cpcloud:
@jreback Can you approve?

ibis/backends/datafusion/compiler.py
```diff
@@ -0,0 +1,396 @@
...
+
+
+@translate.register(ops.NotContains)
+def not_contains(op, expr):
+    value = translate(op.value)
+    options = _prepare_contains_options(op.options)
+    return df.functions.in_list(value, options, negated=True)
+
+
+@translate.register(ops.ElementWiseVectorizedUDF)
```
jreback, line 386: in a future version can you put the ops in the files consistent with other backends e.g. udfs

ibis/backends/datafusion/datatypes.py
```diff
@@ -0,0 +1,90 @@
...
+
+import ibis.common.exceptions as com
+import ibis.expr.datatypes as dt
+import ibis.expr.schema as sch
+
+# TODO(kszucs): the following conversions are really rudimentary
+# we should have a pyarrow backend which would be responsible
+# for conversions between ibis types to pyarrow types
+
+# TODO(kszucs): support nested and parametric types
```
jreback, line 13: i think we have an ibis issue about this already? if not can you create and reference it here. we have these definition already in spark / file based backends for example.

jreback: it might be worth doing this as a pre-cursor or short term followup here.

ibis/backends/datafusion/tests/test_udf.py
```diff
@@ -0,0 +1,43 @@
+import pandas.testing as tm
+import pyarrow.compute as pc
```
jreback, line 2: does datafusion have a min required pyarrow version? is this indicated somewhere? are you testing on this version? (or master or both)?

jreback (changes requested):
a few questions

ibis/backends/datafusion/tests/test_udf.py
cpcloud: https://github.com/datafusion-contrib/datafusion-python/blob/main/pyproject.toml#L47

This is almost certainly an untested and invalid claim (that `pyarrow==1` will work with any recent version of datafusion), and it's not ibis's problem to do CI for the datafusion project. This isn't a blocker to merge this PR.

ibis/backends/datafusion/datatypes.py
cpcloud: A generic "support foo" isn't a good issue. @kszucs if you're going to make an issue please make it for a specific thing to support like `feat: support read of complex types` or similar.

This isn't a blocker for this PR.

jreback: no its 'put all of the pyarrow to ibis type conversion code in a single place'

ibis/backends/datafusion/tests/test_udf.py
jreback: sure but we need a min version of pyarrow here (for datafusion), which IS a blocker (well, certainly can merge and do a followup is ok too)

cpcloud: I think I'm not communicating effectively here.

There is no knowable minimum version right now. The best can do is say the minimum version is the version of pyarrow as of this PR.

I propose that we continue on as is in the interest of merging this PR unless there's a specific alternative to consider.

ibis/backends/datafusion/datatypes.py
cpcloud: That's a reasonable issue. We likely already have it.

Again, this isn't a blocker for this PR.

cpcloud:
I'm going to merge this today, as there are no remaining actionable review items.

jreback:
no objection (but follow ups as indicated)

cpcloud:
@jreback Thanks, can you approve?

PR #2925, opened 2021-09-02
BUG: Fix race condition in temporary table deletion

cpcloud:
These changes fix a race condition when running tests in parallel. Without this
multiple clients try to remove `temp_to_table`/`temp_from_table` independently
and collide with each other.


jreback (approved):

jreback:
cool

PR #2930, opened 2021-09-03
TST: Remove low-utility test

cpcloud:
This test has low utility, as it cannot be validated in CI.


jreback:
it would be nice to have metrics on import time to avoid regressions, but nbd.

jreback (approved):

cpcloud:
I agree, but this test was definitely not that :)

PR #2933, opened 2021-09-05
fix: fix a few broken assumptions in the dask and pandas backends

cpcloud:
Fix bug in pandas and dask backends, where options can not be initialized, if tests are run in an arbitrary order.

- ci: run tests across multiple processes
- fix: ensure that the pandas test client registers options
- fix: ensure that dask trace option is initialized
- fix: ensure that pandas trace option is initialized

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -1,7 +1,6 @@
 #!/bin/bash -e
 # Run the Ibis tests. Two environment variables are considered:
 # - PYTEST_BACKENDS: Space-separated list of backends to run
-# - PYTEST_EXPRESSION: Marker expression, for example "not udf"
 
 TESTS_DIRS="ibis/tests ibis/backends/tests"
 for BACKEND in $PYTEST_BACKENDS; do
@@ -11,13 +10,11 @@ for BACKEND in $PYTEST_BACKENDS; do
 done
 
 echo "TESTS_DIRS: $TESTS_DIRS"
-echo "PYTEST_EXPRESSION: $PYTEST_EXPRESSION"
 
 set -o pipefail
 
 pytest $TESTS_DIRS \
     -q \
-    -m "${PYTEST_EXPRESSION}" \
     -ra \
     --junitxml=junit.xml \
     --cov=ibis \
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -21,4 +21,4 @@ pytest $TESTS_DIRS \
     -ra \
     --junitxml=junit.xml \
     --cov=ibis \
-    --cov-report=xml:coverage.xml | tee pytest.log
+    --cov-report=xml:coverage.xml "$@" | tee pytest.log
--- a/ibis/backends/pandas/trace.py
+++ b/ibis/backends/pandas/trace.py
@@ -117,13 +117,21 @@ def _log_trace(func, start=None):
 
 
 def trace(func):
-    """ Return a function decorator that wraped the decorated function with
+    """Return a function decorator that wraped the decorated function with
     tracing.
     """
     _trace_funcs.add(func.__name__)
 
     @functools.wraps(func)
     def traced_func(*args, **kwargs):
+        # Unfortunately, this function can be called before the `ibis.pandas`
+        # attribute has ever been accessed, which means the trace configuration
+        # option might never get registered and will raise an error. Accessing
+        # the pandas attribute here forces the option initialization
+        import ibis
+
+        ibis.pandas
+
         trace_enabled = get_option(_TRACE_CONFIG)
 
         if not trace_enabled:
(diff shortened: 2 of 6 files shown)


ci/run_tests.sh
```diff
@@ -21,4 +21,4 @@ pytest $TESTS_DIRS \
     -ra \
     --junitxml=junit.xml \
     --cov=ibis \
-    --cov-report=xml:coverage.xml | tee pytest.log
+    --cov-report=xml:coverage.xml "$@" | tee pytest.log
```
datapythonista, line 24: Would it make sense to add `-n auto` for all backends?

If we just want to have this for some backends, I'd implement this in the same way as `PYTEST_EXPRESSION`. With an environment variable for the number of processes, or using the `-m ...` in the same way you implemented this with `$@` (just adding the `-m` in the call in the CI). Having two different ways to do the same makes this more difficult to understand.

datapythonista (commented):
I would split the two separate things in this PR. Increasing the number of processes in one, and the pandas/dask stuff in another.

Besides a preference for atomic commits, I think running tests in parallel is something that can easily start making tests flaky, and would be good to be able to revert that if needed, without affecting the rest of this PR.

cpcloud:
> I would split the two separate things in this PR. Increasing the number of processes in one, and the pandas/dask stuff in another.
> 
> Besides a preference for atomic commits, I think running tests in parallel is something that can easily start making tests flaky, and would be good to be able to revert that if needed, without affecting the rest of this PR.

There's a reason I did this, this bug will never be seen without _some_ kind of test randomization. I don't think it makes sense to split this PR, I'm not parallelizing tests for no reason.

Let's try to avoid enforcement of atomic commits before getting the full context. I'll try to add some more information in the PR description next time.

ci/run_tests.sh
cpcloud: Can you clarify how adding another environment variable is an improvement above using the standard pytest options?

cpcloud:
> I think running tests in parallel is something that can easily start making tests flaky, and would be good to be able to revert that if needed, without affecting the rest of this PR.

It might be better to run things in a random order on every test run then (`pytest-randomly` is a plugin that does this) to catch ordering bugs like this one.

ci/run_tests.sh
datapythonista: I don't have a preference to use an environment variable or `$@`. My point is that what we're doing with `-n` and `-m` is the same, and after this PR, we're using different approaches. Id you prefer to use `$@` as you're doing, I'm fine. But I'd then add `-m ...` to the github yaml, and stop setting the `PYTEST_EXPRESSION` variable. Just for consistency and simplicity.

datapythonista:
> There's a reason I did this, this bug will never be seen without _some_ kind of test randomization. I don't think it makes sense to split this PR, I'm not parallelizing tests for no reason.

Yep, more context would be helpful. I didn't know there was a bug until this comment (still unsure what's the bug). :)

ibis/backends/pandas/trace.py
```diff
@@ -117,13 +117,21 @@ def _log_trace(func, start=None):
...
     tracing.
     """
     _trace_funcs.add(func.__name__)
 
     @functools.wraps(func)
     def traced_func(*args, **kwargs):
+        # Unfortunately, this function can be called before the `ibis.pandas`
+        # attribute has ever been accessed, which means the trace configuration
+        # option might never get registered and will raise an error. Accessing
+        # the pandas attribute here forces the option initialization
```
cpcloud, line 130: @datapythonista :point_up: the bug :)

cpcloud: Another thing to note is that this bug doesn't admit an obvious unit test, AFAICT. I was only able to trigger it when parallelizing tests with a medium-ish number of processes (32). Something like `pytest-randomly` might help.

datapythonista (approved):
Thanks for the additional context. Makes sense now. I added a sentence to the description to clarify what's the goal here. Feel free to edit if it's not accurate.

Lgtm. I still think that with these changes, we should eventually replace in the CI yaml:
```
PYTEST_BACKENDS=$BACKENDS PYTEST_EXPRESSION="not udf" ./ci/run_tests.sh
```
by
```
PYTEST_BACKENDS=$BACKENDS ./ci/run_tests.sh -m "not udf"
```

Would keep things simpler. But not a big deal, and can surely be done in a follow up.

cpcloud:
I do think it would make sense to add `auto` to the ones I'm currently passing `--numprocesses` to. I add those changes, and then roll in the `-m` changes as well.

ci/run_tests.sh
cpcloud: Went ahead and inlined the marker expression.

PR #2934, opened 2021-09-05
refactor: move some config to pyproject.toml

cpcloud:
This PR moves some of the existing configuration in `setup.cfg` (`isort`,
`pydocstyle`, and `pytest`) to `pyproject.toml`. `mypy` just released a version
with `pyproject.toml` support so it's a bit early to move IMO.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ markers = [
 [tool.black]
 line_length = 79
 skip_string_normalization = true
+exclude = "(ibis/_version|versioneer)\\.py"
 
 [tool.isort]
 ensure_newline_before_comments = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,40 @@
+[tool.pytest.ini_options]
+xfail_strict = true
+addopts = [
+  "--ignore=site-packages",
+  "--ignore=dist-packages",
+  "--strict-markers"
+]
+norecursedirs = ["site-packages", "dist-packages"]
+markers = [
+  "backend",
+  "geo",
+  "hdfs",
+  "kudu",
+  "min_spark_version",
+  "only_on_backends",
+  "skip_backends",
+  "skip_missing_feature",
+  "superuser",
+  "udf",
+  "xfail_backends",
+  "xfail_unsupported",
+  "xpass_backends",
+]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,5 +1,41 @@
...
+line_length = 79
+multi_line_output = 3
+include_trailing_comma = true
+profile = "black"
+skip = ["versioneer.py"]
+
+[tool.pydocstyle]
+inherit = false
+convention = "numpy"
+match_dir = "(ibis|omniscidb)"
```
kszucs, line 41: Why `omniscidb`?

setup.cfg
```diff
@@ -11,18 +11,6 @@ ignore = D202,D203,W503,E203
 max-line-length = 79
 exclude = versioneer.py
 
-[pydocstyle]
-inherit = false
-convention = numpy
-
-[isort]
-known_third_party = cached_property,click,clickhouse_driver,dateutil,google,graphviz,impala,kudu,mock,multipledispatch,numpy,pandas,pkg_resources,plumbum,psycopg2,pyarrow,pydata_google_auth,pygit2,pymapd,pymysql,pyspark,pytest,pytz,regex,requests,setuptools,sphinx_rtd_theme,sqlalchemy,thrift,toolz
```
kszucs, line 19: Were not these required previously?

kszucs (approved):
Just minor questions, but it looks good to me.

setup.cfg
cpcloud: Nope. I'm not sure why they were there. I think these are from a very long time ago.

pyproject.toml
cpcloud: This was taken from another config file. It's because these are the only two trees that are able to pass `pydocstyle` checks AFAICT.

jreback: omnisci is in another repo though

jreback (approved):
looks fine, comment but prob nbd

datapythonista:
Thanks @cpcloud. Yes, running pydocstyle for omniscidb can be removed, I forgot to update that when I moved the backend to another repo. Feel free to open a follow up, or I'll do it eventually.

PR #2924, opened 2021-09-01
refactor: use nix to manage development dependencies

cpcloud:
This is a PR to change the way local development is done in Ibis (while preserving as much backwards compatibility with conda and non-conda/non-nix workflows as possible!), accomplishing the following things:

The main change here is the addition of a development environment that is self-contained and derived from a single source of truth (`pyproject.toml`), using `nix`.

See `docs/web/contribute.md` for more details.

Note that **both conda and setuptools-based development workflows are still supported as first class citizens**.

DONE:
1. ~A move to [`poetry`](https://python-poetry.org) for dependency management.~ done in #2937 
   It is still possible and supported to use setup.py and setuptools if you like, this functionality is checked in CI.
1. ~More extensive CI testing, with minimal additional CI time used. It looks
   like roughly on average 3 minutes of additional CI time occur, with more of
   the library being tested on Windows in particular~ done in #2937 
1. ~Automatic generation of a conda recipe for every release, uploaded to GitHub
   releases (this is the remaining 1% that isn't automated but this can
   be automated by pushing a feedstock pull request directly from CI)~ I plan to do this in conda-forge directly, using their integrated bot automation
1. ~Automated Dependabot updates for github-actions and poetry dependencies~ this is done using Renovate in #3053
1. ~Generating a setup.py file from the project's `pyproject.toml` file~ ~This is now just checked that it doesn't need to be regenerated, since we want to ensure this is tested in CI where relevant.~ done in #3054, necessary for Renovate dependency update PRs
1. ~Automatic generation of a conda environment file, uploaded to GitHub releases, useful for developers using conda~ We have automatically generated lockfiles for linux, macos, and windows, for all three versions of Python currently supported (3.7, 3.8, 3.9). These live under `conda-lock` and can be used to set up a development environment very quickly since there's no solve step required. Done #3077, #3080, #3083, #3087

Follow ups:
1. `workflow_dispatch` (click-to-run) GitHub Actions workflow that cuts a release to PyPI and GitHub Releases.
1. Automatic license updates once a year or on-demand. This is optional, but it's another thing we can automate at some point.
1. Automatic nix dependency updates via PR submission every six hours or on-demand. This has to be done in a follow up.

Happy to address any concerns here, let's automate everything!

PR #2937, opened 2021-09-06
refactor(ci): move to poetry

cpcloud:
This PR is an extraction from https://github.com/ibis-project/ibis/pull/2924
of just the poetry/ci bits.

The piece of code here with the most changes is the main CI workflow
(`.github/workflows/main.yml`)

This file is the main thing to focus on for review here.


datapythonista (commented):
I don't think we want to add `poetry.lock` to the repo. This will freeze versions, we won't be aware if installing Ibis with the last dependencies is broken. And will also force us to regenerate the file. Or does the `linguist-generated` means some magic I'm unaware of?

cpcloud:
`linguist-generated` is the way that you inform github not to display generated files in a PR.

cpcloud:
>  This will freeze versions

Yes, that is their entire reason for existence :)

> we won't be aware if installing Ibis with the last dependencies is broken.

we can check this in CI with another job

> And will also force us to regenerate the file.

Yes, that's a good thing. It prevents two people from having different sets of dependencies.

Additionally, we can use dependabot to handle the boring process of upgrading existing dependencies on a regular basis.

Lock files are a really great way of preventing a large swath of "work on my machine" problems.

cpcloud:
On the other hand, `poetry.lock` isn't necessary at the moment, so if it'll make the code easier to review I'll remove it.

datapythonista:
I fail to see your point. How I see things, there are two different kind of projects:
- Something like a website, that is going to be deployed in a server, and where freezing dependencies and making sure eveybody (developers, production servers...) uses the exact same environment is a good thing
- A library used by lots of diverse users, in different operating systems, and different versions of the dependencies. Where freezing dependencies is good for the project developers, but bad for the users, who can just `conda install ibis-project` and get a broken environment, maintainers where not aware of

I guess I'm missing something, but `poetry.lock` seems to be great for the former case, but I fail to see how it's helpful for the latter. Do you want to have all devs test on an arbitrary freezed environment, and have jobs in the CI testing the latest and minimum dependencies? Or there is something else? How easy will be to run with the latest dependencies locally?

cpcloud:
> Where freezing dependencies is good for the project developers, but bad for the users, who can just conda install ibis-project and get a broken environment, maintainers where not aware of

I think this is misunderstanding how lock files work. Dependencies are constrained by `pyproject.toml`, which doesn't force users into anything except the version specifications in that file. Right now, those are fairly liberally constrained, for exactly the kind of scenario you are talking about: diverse sets of users.

With that in mind, how does checking in `poetry.lock` actually prevent **users** from doing anything? Users are never exposed to lockfiles.

cpcloud:
How easy will be to run with the latest dependencies locally?

As easy as `poetry update`. If you want to check in any changes made with those new dependencies, then you'd have to run it through CI.

docs/web/config.yml
```diff
@@ -51,7 +51,7 @@ layout:
       url: https://twitter.com/IbisData
     - font_awesome: github
       url: https://github.com/ibis-project/ibis/
-  footer_note: "© Copyright 2020, Ibis developers"
+  footer_note: "© Copyright 2021, Ibis developers"
```
jreback, line 54: prob should do 2017-2021 (or whatever start date)

ibis/__init__.py
```diff
@@ -49,8 +47,17 @@
         'Number of rows to be retrieved for an unlimited table expression',
     )
 
-__version__ = get_versions()['version']
-del get_versions
+try:
```
jreback, line 50: add a comment that this is >= PY38 so can remove in the future

pyproject.toml
```diff
@@ -1,3 +1,143 @@
...
+  "array"
+] }
+geoalchemy2 = { version = "^0.9", optional = true }
+geopandas = { version = "^0.9", optional = true }
+graphviz = { version = "^0.16", optional = true }
+hdfs = { version = "^2", optional = true }
+importlib-metadata = { version = "^4", python = "<3.8" }
+impyla = { version = "0.18a1", optional = true, extras = ["kerberos"] }
+psycopg2 = { version = "^2", optional = true }
+pyarrow = { version = "^5", optional = true }
```
jreback, line 45: is this bumping the min pyarrow version?

setup.py
```diff
@@ -1,118 +1,144 @@
...
-Ibis is a productivity-centric Python big data framework.
-
-See http://ibis-project.org
-"""
-
-VERSION = sys.version_info.major, sys.version_info.minor
-
-impala_requires = ['hdfs>=2.0.16', 'sqlalchemy>=1.3', 'requests']
-impala_requires.append('impyla[kerberos]>=0.15.0')
+packages = [
```
jreback, line 4: is there a reason to have this fixed vs the dynamic find?

setup.py
```diff
@@ -1,118 +1,144 @@
...
+        "clickhouse-driver>=0.1,<0.2",
+        "clickhouse-sqlalchemy>=0.1,<0.2",
+        "dask[dataframe,array]>2021.6.0",
+        "geoalchemy2>=0.9,<0.10",
+        "geopandas>=0.9,<0.10",
+        "graphviz>=0.16,<0.17",
+        "hdfs>=2,<3",
+        "impyla[kerberos]==0.18a1",
+        "psycopg2>=2,<3",
+        "pyarrow>=5,<6",
```
jreback, line 72: this is pretty strict here, why not a much earlier version of pyarrow?

setup.py
```diff
@@ -1,118 +1,144 @@
...
...is-project/ibis) |\n\n\nIbis is a toolbox to bridge the gap between local Python environments, remote\nstorage, execution systems like Hadoop components (HDFS, Impala, Hive, Spark)\nand SQL databases. Its goal is to simplify analytical workflows and make you\nmore productive.\n\nInstall Ibis from PyPI with:\n\n```sh\npip install ibis-framework\n```\n\nor from conda-forge with\n\n```sh\nconda install ibis-framework -c conda-forge\n```\n\nIbis currently provides tools for interacting with the following systems:\n\n- [Apache Impala](https://impala.apache.org/)\n- [Apache Kudu](https://kudu.apache.org/)\n- [Hadoop Distributed File System (HDFS)](https://hadoop.apache.org/)\n- [PostgreSQL](https://www.postgresql.org/)\n- [MySQL](https://www.mysql.com/)\n- [SQLite](https://www.sqlite.org/)\n- [Pandas](https://pandas.pydata.org/) [DataFrames](http://pandas.pydata.org/pandas-docs/stable/dsintro.html#dataframe)\n- [Clickhouse](https://clickhouse.yandex)\n- [BigQuery](https://cloud.google.com/bigquery)\n- [OmniSciDB](https://www.omnisci.com)\n- [PySpark](https://spark.apache.org)\n- [Dask](https://dask.org/) (Experimental)\n\nLearn more about using the library at http://ibis-project.org.\n",
```
jreback, line 129: shouldn't this be in a separate README? (the long_description)

jreback (changes requested):

setup.py
cpcloud: It's automatically generated, so in a sense it is still dynamic :)

pyproject.toml
cpcloud: It is. I can lower it if desired.

setup.py
cpcloud: It is, it's automatically generated.

setup.py
cpcloud: Mostly because pyarrow tends to break stuff liberally at every release (which is fine, it's major version bump). I can certainly try decreasing the lower bound.

cpcloud:
@jreback `setup.py` is automatically generated, so it's not that interesting to review.

setup.py
cpcloud: What's a good lower bound? 3, 4?

docs/web/config.yml
cpcloud: Looks like the first commit was 2014, I'll update.

ibis/__init__.py
cpcloud: Done.

docs/web/config.yml
cpcloud: Done.

jreback:
my main concern here is to have a lower bound of deps that *are* tested in the CI that are the oldest supported version. we don't have a reason to bump pyarrow for example from 1.* i think (which is where we had it). ibis likely just works on the oldest versions so we should allow as liberal as possible. now in a future version of course we should bump, but i don't see why we would do that now.

setup.py
jreback: i would say >=1 (yes these are not that different), but the release cadence is quite aggressive in pyarrow and this would *force* updating. for newer versions of ibis, can certainly bump.

cpcloud:
@jreback I've added some minimum version testing jobs, let's see what shakes out.

.github/workflows/main.yml
```diff
@@ -1,79 +1,233 @@
...
+        pyarrow:
+          - "0.12.1"
+          - "5"
+        exclude:
+          - python-version: "3.7"
+            pyspark: "3"
+            pyarrow: "0.12.1"
+          - python-version: "3.9"
+            pyspark: "3"
+            pyarrow: "0.12.1"
```
datapythonista, line 223: I personally find this a step backewards to what we've got now. If these versions can be obtained from the pyproject.toml or poetry files, my preference would be to keep what we've got now (files in `ci/deps/...`).

datapythonista (commented):
I also don't understand the new division of builds/backends. Maybe I just got used to what we've got. But those `only_backends`, `linux_only`... builds are not very clear to me. Do you mind explaining a bit why these new builds are better than what we've got now please?

setup.py
kszucs: We should try with minimum version `1.0` and bump as needed.

cpcloud:
> I also don't understand the new division of builds/backends. Maybe I just got used to what we've got. But those only_backends, linux_only... builds are not very clear to me. Do you mind explaining a bit why these new builds are better than what we've got now please?

There are two very different things in question here, let's discuss them individually:

1. The way the workflow jobs are named. Happy to rename to something that is more clear for you. Keep in mind that I did actually try to follow a similar structural pattern to what was already there. The way they are currently named is:
    - `tests_no_backends`: tests that don't require a backend, basically whatever's in`ibis/tests`.
    - `tests_simple_backends`: tests that require a "simple" backend; csv, dask, hdf5, pandas, parquet, and sqlite
    - `tests_linux_only_backends`: tests that require a backend that can only be **tested** on a linux runner due to use of services (mysql, postgresql) and pyspark. These backends likely work on Windows too, but GitHub does not offer services on Windows runners.
    - `tests_impala_clickhouse`: tests against the impala and clickhouse backends

There are a handful of other jobs whose role is hopefully clear from their name, e.g., `benchmarks`.

2. Why is this an improvement upon the existing workflows? A few reasons:
    - There's much less ad-hoc shell scripting. I think putting scripts in CI hurts the ability to understand what is happening in CI. Maintainers should basically never have to open up a shell script to understand a particular job. Encapsulation should be done with reuse of GitHub actions elements, not with ad hoc shell scripts. Any shell scripting is done inline now, because it's very short. Since no one is running these scripts outside of CI there's no real reason IMO to move any of the logic to a script.
    - There's much more (re)use of existing github actions. In the current workflow environments are setup in a totally ad hoc way. With this workflow, environments are created using `conda-incubator/setup-miniconda`, and don't require a bunch of shell scripting to create an environment. Another example, `docs` aren't using SSH anymore, they are using a GitHub action that pushes to the docs repo using a personal access token (which any maintainer can create)
    - Fewer environment variables. Minimizing the use of ad-hoc global state is basically always a good thing IMO.
    - There are more concurrent jobs with just a tiny increase in CI minutes. Dask for example is now being tested on windows.
    - Jobs are doing one thing and one thing only. There's no clear reason why linting and benchmarking (for example) are being run together. These things are unrelated and should run in separate tasks. In the new workflow linting is outsourced to `pre-commit.ci`, and benchmarking is its own job.

datapythonista:
Thanks for all the information, and for all the work on this. While I agree with your points, and there are clearly some advantages to this approach, there is a trade-off, and we're losing several things:

- The reason to have the (very simple) shell scripts, is that to change a paramter of the pytest, we could change it in just one place, and have it updated in all jobs. Also making sure pytest parameters are consistent among jobs
- That's also one of the reasons we don't use GitHub actions to setup conda. Instead of having to repeat all parameters at each job, we just have these two simple scripts to call, and we keep the github config very simple. You're adding 700+ lines of code to the config, to remove less than 100 in the two scripts with minimal complexity
- There is also the addition in the poetry python script in your implementation, which makes things even more complex
- Having dask tested on windows, or each job doing a single thing, is something we can easily implement with the current approach. The reason we don't do it is because we optimize for the time that the CI takes, and the number of jobs, which if too big, makes it more difficult to find failing jobs

We had lots of problems with the CI in the past, I spent monts to make it simple, robust and fast. And this PR to me seems to be adding a lot of complexity, and slower builds, for no gain. I think this PR is making several independent changes (I may be wrong, but I don't think removing the shell scripts in favor of github actions, or the new job division is related to poetry). Is any of this needed for the automation of packaging?

cpcloud:
> The reason to have the (very simple) shell scripts, is that to change a paramter of the pytest, we could change it in just one place, and have it updated in all jobs. Also making sure pytest parameters are consistent among jobs

We can keep `run_tests.sh` if you want, that's not the script that I think is problematic. Again, scripts are useful if something is changing all the time. Are pytest commands in CI changing so often that a script is warranted.

> Instead of having to repeat all parameters at each job, we just have these two simple scripts to call, and we keep the github config very simple.

Can you spell out the logic of `ci/setup_env.sh` here? It might help others to validate the claim that that it's a simple script.

Generally this script seems to be doing about a huge number of unnecessarily manual operation things, despite it being only 77 lines. This is abuse of Bash in the name of concision.

To my eye this is what's happening:

1. A bunch of totally unnecessary path munging on linux and windows (useless in general AFAICT, this PR demonstrates it is entirely unnecessary)
2. Update the base conda environment (unnecessary when using the setup-miniconda action)
3. install mamba (unnecessary when using the setup-miniconda action)
4. dynamically construct an environment.yml based on a bunch of files in some location, arbitrarily deciding that the minimum supported version of Python implies the minimum supported version of all the other dependencies (why?). This appears to be utterly untested in some cases and is in fact broken in the case of pyspark=2.4.3 ([see this PR's CI](https://github.com/ibis-project/ibis/runs/3534479466?check_suite_focus=true) which catches it), see [this part of the pyspark docs](https://spark.apache.org/docs/latest/api/python/user_guide/arrow_pandas.html#compatibility-setting-for-pyarrow-0-15-0-and-spark-2-3-x-2-4-x) for details. [Searching the ibis repo](https://github.com/ibis-project/ibis/search?q=ARROW_PRE_0_15_IPC_FORMAT) I see nothing about this. Here we have complexity in the form of file generation hiding a missed testing scenario
5. showing the environment file (unnecessary when using the setup-miniconda action)
6. updating the base environment again? (unnecessary when using the setup-miniconda action)
7. installing ibis
8. loading data for backends depends on the value of two global variables. (totally unnecessary by just inlining the main loop in CI)
9. showing the packages that were installed (unnecessary when using the setup-miniconda action)

> You're adding 700+ lines of code to the config, to remove less than 100 in the two scripts with minimal complexity

By my calculation I'm adding 611 (line count of `main.yml` in this PR) + 135 (poetry2recipe.py) + 56 (poetry2setup.py) minus 336 (existing `main.yml`) + 77 (setup_env.sh) + 21 (run_tests.sh) + 36 (the line count of the deps files) , for a grand total of **332** net lines of code if my arithmetic is correct.

How are you calculating 700+ **additional** lines?

> There is also the addition in the poetry python script in your implementation, which makes things even more complex

Which one are you talking about? There are two.

> Is any of this needed for the automation of packaging?

Ultimately yes.

I want to move to a model where we reuse github actions, even if that means repeating some of the configuration. That configuration hardly ever changes. Why should we continue to roll our own scripts that are clearly generating broken CI jobs?

datapythonista:
If we had a single job, I'd surely prefer your approach to setting up the conda environment. You don't like `ci/setup_env.sh`, but it solves the problem of having to repeat this code for each job (we would have 5 copies of it with the current CI):
```
      - name: download env file
        uses: actions/download-artifact@v2
        with:
          name: conda-env-file
          path: tmp

      - uses: conda-incubator/setup-miniconda@v2
        with:
          mamba-version: "*"
          miniforge-variant: Mambaforge
          miniforge-version: latest
          channel-priority: strict
          activate-environment: ibis
          python-version: ${{ matrix.python-version }}
          environment-file: tmp/environment.yaml
          condarc-file: ci/condarc

      - name: install boost
        run: conda install -n ibis boost

      - name: install ibis
        run: pip install .

      - name: download backend data
        run: python ci/datamgr.py download

      - name: set backends
        id: set_backends
        run: echo '::set-output name=backends::["impala", "clickhouse"]'

      - name: install backend data
        run: |
          set -euo pipefail
          python ci/impalamgr.py load --data
          python ci/datamgr.py clickhouse
```
Instead, we have that script implemented once, and in each job we simply have:
```
    - name: set up environment
      run: ./ci/setup_env.sh "${{ matrix.python_version }}" "$BACKENDS"
```

I'm open to your approach, even if I don't think it adds value, and it creates a lot of duplication. But I think this discussion will be much more efficient, if you propose each of the changes here in a separate PR. There are many unrelated things that are being mixed, and it's difficult to know what each of them involve with this huge PR.

Would you mind opening separate PRs for:
- Change the jobs to the new ones you propose
- Moving the code of `ci/setup_env.sh` to actions in the github yaml
- Moving the call in `ci/run_tests.sh` to the yaml
- Changes to the website
- Generation of the dependencies with poetry
- ...

I think discussions will be much more focussed, and reviewing much more efficient if we don't mix things.

cpcloud:
I will open separate PRs.

Just to set expectations:
> * Change the jobs to the new ones you propose

ok

> * Moving the code of ci/setup_env.sh to actions in the github yaml

ok

> * Moving the call in ci/run_tests.sh to the yaml

sounds good

> * Changes to the website

sure

> * Generation of the dependencies with poetry
> * Replacing versioneer by importlib.metadata

These need to be done together because poetry doesn't know anything about versioneer, and importlib metadata replaces the `__version__` attribute creation

cpcloud:
Small modification: the first two steps (yaml refactoring and setup_env.sh don't make sense to split because of the minimum deps issue)

.github/workflows/main.yml
```diff
@@ -1,79 +1,256 @@
...
-  Tests_dask:
-    name: Tests dask
+      - name: check whether setup.py needs to be regenerated
+        env:
+          # this environment variable is set is to ensure reproducible setup.py
+          # generation
+          #
+          # because various data structures in poetry are unordered, an can
+          # result in things like package extras appearing in a different order
+          PYTHONHASHSEED: 0
```
kszucs, line 31: Can we incorporate this into the script?

ci/pytest_errors.sh
```diff
@@ -1,15 +1,19 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
+
+set -euo pipefail
+
 # parse the pytest log and print the pytest errors in a concise way
 
 if [[ -f "pytest.log" ]]; then
```
kszucs, line 7: Could we use a pytest plugin instead of a wrapper script so it would be available for all contributors regardless of the platform?

dev/poetry2recipe.py
```diff
@@ -0,0 +1,135 @@
...
+from pathlib import Path
+from typing import Optional, Sequence
+
+import tomli
+from poetry.core.factory import Factory
+from poetry.core.masonry.builders.sdist import SdistBuilder
+from ruamel.yaml import YAML
+
+
+def main(
```
kszucs, line 13: @cpcloud could you consolidate the two poetry scripts into a single one? 

Eventually I think we should have a python library containing all the required functionality for development. Including the poetry conversion scripts, datamgr, impalamgr and possibly other utilities in the future.

.github/workflows/main.yml
```diff
@@ -1,79 +1,256 @@
...
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: download env file
+        uses: actions/download-artifact@v2
+        with:
+          name: conda-env-file
+          path: tmp
+
+      - uses: conda-incubator/setup-miniconda@v2
```
kszucs, line 94: Since we have a concise poetry setup now, I'd just remove conda from the CI setup altogether. We can install all of the dependencies here from PyPI.

.github/workflows/main.yml
```diff
@@ -1,79 +1,256 @@
...
```
kszucs, line 152: Same here, we don't have any dependencies requiring conda to be installed.

.github/workflows/main.yml
kszucs: I'd also remove the `gen_conda_env_file` job making the setup much simpler.

.github/workflows/main.yml
```diff
@@ -1,79 +1,256 @@
...
+        run: ./ci/pytest_errors.sh
+        if: failure()
+
+  test_simple_backends:
+    runs-on: ${{ matrix.os }}
+    needs:
+      - gen_conda_env_file
+    defaults:
+      run:
+        shell: bash -l {0}
```
kszucs, line 129: Shouldn't be required once once conda gets eliminated from the job.

.github/workflows/main.yml
```diff
@@ -1,79 +1,256 @@
...
+
+      - name: install ibis
+        run: pip install .
+
+      - name: download backend data
+        run: python ci/datamgr.py download
+
+      - name: set backends
+        id: set_backends
+        run: echo '::set-output name=backends::["csv", "dask", "hdf5", "pandas", "parquet", "sqlite"]'
```
kszucs, line 174: Could we just hardcode the backends and call the appropiate `datamgr.py` commands? 

```bash
python ci/datamgr.py download
python ci/datamgr.py parquet
python ci/datamgr.py sqlite
# the rest of the click commands do nothing according to the source code
```

.github/workflows/main.yml
```diff
@@ -1,79 +1,256 @@
...
+          for backend in ${{ join(fromJSON(steps.set_backends.outputs.backends), ' ') }}; do
+            python ci/datamgr.py "$backend"
+          done
+
+      - name: run tests
+        env:
+          PYTEST_BACKENDS: ${{ join(fromJSON(steps.set_backends.outputs.backends), ' ') }}
+        run: |
+          pytest \
+            ibis/backends/{${{ join(fromJSON(steps.set_backends.outputs.backends), ',') }}} \
```
kszucs, line 189: Why not `pytest ibis/backends/`? 

I assume `PYTEST_BACKENDS` should automatically handle to skip the not requested backends - otherwise we should tune that.

.github/workflows/main.yml
```diff
@@ -86,24 +263,80 @@ jobs:
...
+          miniforge-version: latest
+          channel-priority: strict
+          activate-environment: ibis
+          python-version: ${{ matrix.python-version }}
+          environment-file: tmp/environment.yaml
+          condarc-file: ci/condarc
+
+      - name: install pyspark
+        run: |
+          conda install -n ibis \
```
kszucs, line 288: Preferably we should use https://github.com/marketplace/actions/setup-apache-spark

.github/workflows/main.yml
```diff
@@ -86,24 +263,80 @@ jobs:
...
+            -ra \
+            --junitxml=junit.xml \
+            --cov=ibis \
+            --cov-report=xml:coverage.xml | tee pytest.log
+
+      - name: pytest errors
+        run: ./ci/pytest_errors.sh
+        if: failure()
+
+  test_impala_clickhouse:
```
kszucs, line 332: I'd prefer having separate jobs for impala and clickhouse, so we can directly see which backend is failing from github checks.

kszucs: Actually I'd prefer this approach for the simple backends as well by parametrizing the job matrix per backend. GHA provides enough parallelization so we don't need to worry about redundant install steps, which should be pretty click using poetry.

.github/workflows/main.yml
kszucs: Setting up miniconda and installing pyarrow in a separate step [takes 7 minutes](https://github.com/ibis-project/ibis/runs/3535730775) even with mamba.

kszucs: Poetry install without a lockfile took 420s on my local machine while poetry install with a lock file took less than 15 seconds to install.

kszucs: Experimented further with poetry. Seems like `poetry update` is really quick after having a lockfile.

My proposal is either:
- to have a poetry lockfile in the repository, keep it updated using dependabot and always call poetry update before install: `poetry update && poetry install` OR
- install ibis using pip: `pip install -e ibis` 

This way the CI jobs will use the latest dependencies.

.github/workflows/main.yml
```diff
@@ -67,6 +67,8 @@ jobs:
 
   test_no_backends:
     runs-on: ${{ matrix.os }}
+    needs:
+      - check_setup_py
```
kszucs, line 71: I suggest to remove this dependency and create another workflow file which would be only triggered if either the setup.py generator script of pyproject.toml changes. An example from arrow: https://github.com/apache/arrow/blob/master/.github/workflows/go.yml#L20-L33

.github/workflows/main.yml
cpcloud: Unfortunately no, because this variable appears to have no effect if set before startup.

ci/pytest_errors.sh
cpcloud: The pytest result log plugin may help here but it requires a bit of munging with jq or something similar. Would it help to state the problem that the log munging is solving?

dev/poetry2recipe.py
cpcloud: Maybe but these are pretty different tools, so I'm not sure putting them into one place makes sense. I think making poetry2recipe a separate package outside of ibis would be useful.

.github/workflows/main.yml
cpcloud: Yep, I've almost got this working and installation is much faster

.github/workflows/main.yml
cpcloud: Yep

.github/workflows/main.yml
cpcloud: Yep

.github/workflows/main.yml
cpcloud: Roger that. Will fix.

.github/workflows/main.yml
cpcloud: I will open up a new issue, I don't think it currently works that way unfortunately.

.github/workflows/main.yml
cpcloud: That actually isn't necessary, just a modern java runtime is needed. The pyspark jobs are now passing

pyproject.toml
cpcloud: This now set to 1.0

setup.py
cpcloud: This is set to 1.0 now.

.github/workflows/main.yml
cpcloud: done

.github/workflows/main.yml
cpcloud: done

.github/workflows/main.yml
cpcloud: Not necessary after matrix-ing the jobs.

.github/workflows/main.yml
cpcloud: Done.

.github/workflows/main.yml
cpcloud: Done.

.github/workflows/main.yml
cpcloud: Done.

cpcloud:
As @kszucs noted the job run times are much for poetry are much faster than with miniconda (even with mamba and mambaforge), due to a few reasons, but one major contributor is the removal of the solving step.

This speed up requires a lock file checked in to the repo. As I note above, the lockfile doesn't affect users.

ci/pytest_errors.sh
cpcloud: This is now capture using a github action that will report errors in the GitHub actions UI.

dev/poetry2recipe.py
cpcloud: After talking with @mariusvniekerk I think we can remove `poetry2recipe` altogether after the next release of ibis, and depend on conda-forge machinery to discover dependencies.

poetry.lock
```diff
@@ -0,0 +1,3225 @@
+[[package]]
+name = "alabaster"
+version = "0.7.12"
```
jreback, line 3: so this is autogenerated and is considered for the latest version of testing right?

jreback (changes requested):
can you explain how the min version builds get the list of deps?

cpcloud:
> can you explain how the min version builds get the list of deps?

Yeah, for sure.

Taking dask as an example:
The minimum version is specified as part of the build matrix: https://github.com/ibis-project/ibis/pull/2937/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R65. The existing version is the replaced in the lock file with this line https://github.com/ibis-project/ibis/pull/2937/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R94.

Pyspark and postgres deps are done similarly.

cpcloud:
@jreback 

These are actually testing the minimum versions against multiple Python versions as well, so that we can be sure that a newer Python version doesn't invalidate a minimum version.

cpcloud:
I suspect we should wait on this until after 2.0, so that the release procedure doesn't need to change (it shouldn't change too much with this PR, but there's no reason to risk it)

poetry.lock
cpcloud: Yes, but see [my comment below](https://github.com/ibis-project/ibis/pull/2937#issuecomment-921156067).

jreback:
thanks @cpcloud 

yeah i think better to wait on this change for after the release

poetry.lock
cpcloud: You can think of the lock file as a sort dumb cache for poetry's dependency solver. It's a record of what dependencies were produced at the time of the solve.

jreback:
thanks @cpcloud 

followups if you can verify that have sufficient documentation around how to use this (e.g. where / how to update the min versions of things).

PR #2938, opened 2021-09-07
ENH: Add __getstate__ and __setstate__ for Nodes

patcao:
From icexelloss:
We are trying to build "consistent hash" for ibis Expression (hash that doesn't change cross Python interpreter restart for ibis Expression) and the current implementation depends on first pickling the ibis Expression then use something like hashlib.sha256() to compute the hash from the pickled bytes.

During this, we found that because ibis Node class that "_cached_expr" and "_hash" attributes, the pickled bytes can be different for the same table depending on whether those two fields are created, which is not a very desirable behavior. This PR aims to fix the serialization/deserialization method so that they produce consistent results for the equivalent Node object.

### DIFF
--- a/docs/source/release/index.rst
+++ b/docs/source/release/index.rst
@@ -12,6 +12,7 @@ Release Notes
    These release notes are for versions of ibis **1.0 and later**. Release
    notes for pre-1.0 versions of ibis can be found at :doc:`release-pre-1.0`
 
+* :feature:`2938` Serialization-deserialization of Node via pickle is now byte compatible between different processes
 * :support:`2914` Removed `log` method of clients, in favor of `verbose_log` option
 * :feature:`2916` Support joining on different columns in ClickHouse backend
 * :feature:`2908` Support summarization of empty data in Pandas backend
--- a/ibis/expr/operations.py
+++ b/ibis/expr/operations.py
@@ -3,7 +3,7 @@ import functools
 import itertools
 import operator
 from contextlib import suppress
-from typing import List
+from typing import Any, Dict, List
 
 import numpy as np
 import toolz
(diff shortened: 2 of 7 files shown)


datapythonista:
@patcao can you link an issue number if you're working on one. Or provide a description to know why you are proposing this change please. Thank you

ibis/tests/expr/test_table.py
```diff
@@ -1298,6 +1298,12 @@ def test_pickle_table_expr():
     assert t1.equals(t0)
 
 
+def test_pickle_table_node(table):
```
icexelloss, line 1301: Can you add a few more test with more operations and UDFs?

ibis/expr/operations.py
```diff
@@ -59,6 +59,18 @@ def _pp(x):
 
         return '{}({})'.format(opname, ', '.join(pprint_args))
 
+    def __getstate__(self):
```
icexelloss, line 62: Can you add docstring for these two method describing what you are doing and why?

ibis/expr/operations.py
```diff
@@ -59,6 +59,18 @@ def _pp(x):
...
+        excluded_slots = {'_expr_cached', '_hash'}
```
icexelloss, line 63: Can you add comment here why it's OK to exclude these fields for serialization

ibis/expr/operations.py
```diff
@@ -59,6 +59,18 @@ def _pp(x):
...
+        return {
```
icexelloss, line 64: I am not certain how it works when slot is another ibis TableExpr or ColumnExpr, can you explain it works?

icexelloss:
@datapythonista Chime in a little bit here about the rationale here. We are trying to build "consistent hash" for ibis Expression (hash that doesn't change cross Python interpreter restart for ibis Expression) and the current implementation depends on first pickling the ibis Expression then use sth like hashlib.sha256() to compute the hash from the pickled bytes. 

During this, we found that because ibis Node class that "_cached_expr" and "_hash" attributes, the pickled bytes can be different for the same table depending on whether those two fields are created, which is not a very desirable behavior. This PR aims to fix the ser/de method so that they produce consistent results for the equivalent  Node object.

jreback (changes requested):
also pls add a release note

ibis/expr/operations.py
```diff
@@ -59,6 +59,37 @@ def _pp(x):
...
+            A dictionary storing the objects attributes.
+        """
+        excluded_slots = {'_expr_cached', '_hash'}
+        return {
+            slot: getattr(self, slot)
+            for slot in self.__slots__
+            if slot not in excluded_slots
+        }
+
+    def __setstate__(self, state: Dict[str, Any]):
```
jreback, line 83: return -> None

jreback (changes requested):
minor comment. @icexelloss has a couple of comments. pls add a release note as well.

icexelloss (approved):
LGTM

jreback:
@patcao can you add a release note. ping on green.

docs/source/release/index.rst
```diff
@@ -12,6 +12,7 @@ Release Notes
    These release notes are for versions of ibis **1.0 and later**. Release
    notes for pre-1.0 versions of ibis can be found at :doc:`release-pre-1.0`
 
+* :feature:`2938` Add `__getstate__` and `__setstate__` methods to Node
```
jreback, line 15: can you just say, that serialization-deserialization via pickle is now byte compatible between different processes.

jreback (approved):

jreback:
thanks @patcao

PR #2941, opened 2021-09-07
fix: forward the spark entry point to pyspark

cpcloud:
This fixes a bug that isn't caught in the current CI where the spark entrypoint
points to a non-existent `spark` attribute.


jreback:
yeah this is fine. IIRC we actually should be entirely killing the spark entry point cc @icexelloss ?

jreback (approved):

PR #2942, opened 2021-09-07
fix: allow an undefined PYTEST_BACKENDS variable to indicate no backends should be tested

cpcloud:
This PR slighty changes the meaning of PYTEST_BACKENDS.

It's a really terrible developer experience to suddenly be
testing all backends because of an **undefined** environment variable.

The new behavior is:

- If the variable is undefined or empty, return an empty list, so no backends are tested
- If the variable is defined, existing behavior is retained

### DIFF
--- a/ibis/backends/tests/conftest.py
+++ b/ibis/backends/tests/conftest.py
@@ -49,12 +49,20 @@ def _get_backends_to_test():
     """
     Get a list of `TestConf` classes of the backends to test.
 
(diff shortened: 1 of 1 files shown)


datapythonista (approved):
Looks reasonable, thanks

cpcloud:
Thanks for adding the issue labels, I keep forgetting to do that :sweat_smile:

PR #2950, opened 2021-09-08
ci: refactor ci part 1

### LINKED ISSUE #2946: CI: Pyspark dependencies are not thoroughly tested
Because we only test minimum and maximum versions, we have a lurking bug in the pyspark backend. See https://spark.apache.org/docs/latest/api/python/user_guide/arrow_pandas.html#compatibility-setting-for-pyarrow-0-15-0-and-spark-2-3-x-2-4-x for details.

### LINKED ISSUE #2948: CI: Windows jobs assume the shell is not a login shell
We have some development dependencies is this monstrous `environment.yml` that introduces a very subtle bug on Windows where pipes fail.

Here's an example where this fails: https://github.com/cpcloud/ibis/runs/3547021196?check_suite_focus=true

After removing the (totally underspecified) release and documentation dependencies, the tests are able to run on Windows again.

https://github.com/cpcloud/ibis/runs/3547642657?check_suite_focus=true

We should not be introducing dependencies for development that cause these sorts of problems.

cpcloud:
This PR refactors the CI to use more GitHub actions, and incrementally reduce
the SLOCs of ad-hoc shell scripts with more to come on that in follow up PRs.

Closes #2946.
Closes #2948.


jreback:
can you rebase (i merged the other PR)

jreback (approved):
lgtm

jreback:
worth a note about https://github.com/ibis-project/ibis/issues/2946 in the backend docs?

jreback:
dont a lot of @kszucs comments on https://github.com/ibis-project/ibis/pull/2937 apply here? e.g. request to separate the backends to different actions so they show up separately?

cpcloud:
@jreback Yeah, but I was going to address those in that PR, after this one lands to keep the patch smaller

jreback:
> @jreback Yeah, but I was going to address those in that PR, after this one lands to keep the patch smaller

cool looking

ci/merge_and_update_env.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
datapythonista, line 1: I fail to understand why you don't like setup_env.sh which saves lots of repetition in the github yaml, and this script is fine. It's nice that this scrip is more atomic and does a more specific thing, but seems like a small detail, considering the value that IMO add not having to duplicate the conda config for every job.

In any cas, I don't understand the criteria, but not a problem. I'm fine with this if you've got a strong opinion of this being better.

jreback:
thanks @cpcloud 

i think breakign out the jobs to individual reports is useful

environment.yml
```diff
@@ -1,8 +1,5 @@
 # This file should have all the dependencies for development excluding the specific to the backends.
-name: ibis-dev
-channels:
-  - nodefaults
-  - conda-forge
```
datapythonista, line 5: I guess with mambaforge and strict channel priority this is not needed in the CI. But I think it can still be helpful when devs create the environment locally with `conda env create`.

datapythonista, line 5: I guess with mambaforge and strict channel priority this is not needed in the CI. But I think it can still be helpful when devs create the environment locally with `conda env create`.

environment.yml
```diff
@@ -1,8 +1,5 @@
...
+name: ibis
```
datapythonista, line 2: I don't care that much about the name of the environment. `ibis-dev` was based on the pandas environemnt, which I assumed was named to avoid conflicts with `pandas` named environments created by devs. But I do think it's a poor open source development practice to make arbitrary changes based on personal preferences, and I would avoid them. It would consume a lot of everybody's energy if we keep changing things based on personal preferences.

cpcloud: Happy to revert this back, I don't really care what it's called.

cpcloud: It seemed more obvious that an environment would carry the name of the library but maybe that's just me.

PR #2953, opened 2021-09-09
ENH: Remove Expr._root_tables()

kszucs:
Root tables are operation specific, so call it explicitly rather than hiding it behind a proxy method.


jreback (approved):
lgtm. no user facing effects, so no note.

PR #2956, opened 2021-09-10
fix: Change map to geo_equals

### LINKED ISSUE #2197: mutate geo column causes TypeError: geo_equals() got an unexpected keyword argument 'cache'

xmnlab:
This PR fix the problem with `geo_equals`.

Currently, `equals` operation is mapped to `geo_equals` and it is given problems as reported at #2197

Fix #2197


ibis/backends/tests/test_geospatial.py
```diff
@@ -686,3 +686,45 @@ def test_geo_ops_smoke(backend, fn_expr):
     """Smoke tests for geo spatial operations."""
     geo_table = backend.geo
     assert fn_expr(geo_table).compile() != ''
+
+
+@pytest.mark.only_on_backends(["postgres"])
+@pytest.mark.xfail_unsupported
+def test_geo_equals(backend, geo):
+    """
+    Test geo_equals/equals.
```
jreback, line 695: reference the issue in a comment. the Notes section is not needed here

jreback (changes requested):
tests are failing; also pls add a release note

jreback (approved):

xmnlab:
@jreback thanks for the review.

I added the release note and also changed 2 tests that were using `.equals`. to `geo_equals`.
In ibis, as I understood, `equals` usually returns a boolean, not an ibis expression.

I also updated the documentation about `.equals`

ibis/backends/postgres/tests/test_postgis.py
```diff
@@ -149,9 +149,16 @@ def test_geo_disjoint(geotable):
...
-    expr = geotable.geo_point.equals(geotable.geo_point)
+    # note: == operation just works for comparison between same datatypes
+    expr = geotable.geo_point == geotable.geo_point
     assert expr.execute().all()
 
+    expr = geotable.geo_point.geo_equals(geotable.geo_point)
+    assert expr.execute().all()
+
+    expr = geotable.geo_point.geo_equals(geotable.geo_linestring)
+    assert not expr.execute().all()
```
datapythonista, line 160: Do you want all them to be different? Then `any` should be better I think.

datapythonista (approved):
lgtm

ibis/backends/postgres/tests/test_postgis.py
xmnlab: that makes sense. I will change it now. thanks @datapythonista

jreback:
thanks @xmnlab

PR #2957, opened 2021-09-10
ci: replace bespoke test munging with junit reporter

cpcloud:
This PR replaces our current bespoke test output munger with a GitHub action that generates a
report from pytest's junitxml test output.


jreback:
where does this show output?

cpcloud:
@jreback In a test report job:

Example: https://github.com/cpcloud/ibis/runs/3566756856?check_suite_focus=true

jreback (approved):

datapythonista:
If I'm not missing anything, I think you misunderstood what the delete script was doing.

The `pytest_errors.sh` script and CI step was just to not have to navigate the huge pytest logs in order to find failed tests tracebacks. Pytest shows all the progress of all the run tests, lots of warnings, and a long list of skipped tests at the end. The traceback of a failed test is in the middle of all this, and with Github UI takes a relevant time of scrolling to find the error traceback.

Would be nice if pytest would let for example have the tracebacks at the very end, that would make things much faster to find. But it doesn't afaik. So, the removed script was a bit of a hack, but creating an extra step in the CI, with just the most relevant output of the failures. The content that at least I care 95% of the times there is a failure in the CI.

If I'm not wrong, this has been replaced here by an XML artifact that needs to be downloaded in a zip. I'm fine having this added, and I'd be very happy if there is a simpler way to quickly access the test failures traceback, than with the hack I implemented. But feels like we just lost that, for an unrelated xml report.

cpcloud:
The XML output contains failures too and in theory the action should make a comment on the PR with a link to annotations where you can see the traceback for each individual failure.

So, it's definitely related, and designed to solve the exact same problem, albeit with actions and the GitHub UI.

datapythonista:
Ah, amazing, that's the part I missed. But seems like it's not working as expected. I checked with #2905, which I rebased after this was merged, and I couldn't find the failures anywhere. Do you mind having a look or giving me a lead or where to look for the problem please.

cpcloud:
Yep, but I'm on my phone right now. Might be later tonight or tomorrow before I can get to it.

cpcloud:
@datapythonista it's not showing up because the PR to add the test report didn't land until a few hours after your test run. The next run should have the report, assuming you don't change anything.

datapythonista:
Not sure what you mean by "I don't change anything". I merged master and made a small change to the PR, and after failing the job again, there is no comment. I'm not sure if it's not working, or if I'm expected to do something else.

cpcloud:
When you posted the question, the last failure on your linked PR was 5 hours before I'd merged the this one. Can you link to a failure that occurred after this PR was merged?

datapythonista:
> When you posted the question, the last failure on your linked PR was 5 hours before I'd merged the this one. Can you link to a failure that occurred after this PR was merged?

https://github.com/ibis-project/ibis/actions/runs/1223527303

cpcloud:
Hm, I see the test report succeeded, but yeah it does look like the comment isn't going anywhere. I can look into it over the next couple of days.

datapythonista:
Sure, no rush. It's a nice to have, but the tracebacks can be found by scrolling in the logs, so no rush at all.

PR #2958, opened 2021-09-10
ci: handle matrix for test output

cpcloud:
This PR is a follow up to #2957 to handle the test output from every job in the build matrix


jreback (approved):
cool

PR #2959, opened 2021-09-10
DEPR: Deprecate set_database

### LINKED ISSUE #2913: Implement .set_database() for all backends or remove it from the API
In #2909 the method `.set_database()` has been standardized, but this does not include making it always available in all backends. There are several backends, such as SQLAlchemy backends, File backends or clickhouse, that don't implement it, and don't let changing the backend.

The reason in most cases seems to be that the underlying connection object doesn't support changing the database. But in those cases, we can keep a `_current_database` attribute ourselves, and use that database on queries.

Having `.set_database()` always implemented should make things easier for users.

Another alternative is to simply remove it from the few databasets that implement it. Users can connect to a different database, and have different variables with connections to different databases.

datapythonista:
Closes #2913

Only Impala and pyspark support `set_database`. It doesn't add much value, since a new client can be created for the different database. But it adds extra complexity, having to handle the not implement errors, and managing a state for the two backends implemented. Also, it creates wrong expectations in the users, since it's in the documentation, but almost no backend supports it.

CC: @cpcloud @jreback

### DIFF
--- a/ibis/backends/impala/__init__.py
+++ b/ibis/backends/impala/__init__.py
@@ -1,4 +1,6 @@
 """Impala backend"""
+import warnings
+
 import ibis.config
 from ibis.backends.base.sql import BaseSQLBackend
 from ibis.backends.base.sql.ddl import fully_qualified_re
@@ -137,6 +139,12 @@ class Backend(BaseSQLBackend):
         )
 
     def set_database(self, name):
+        warnings.warn(
+            '`set_database` is deprecated and will be removed in a future '
+            'version of Ibis. Create a new connection to the desired database '
+            'instead',
+            FutureWarning,
+        )
         self.client.con.set_database(name)
 
     @property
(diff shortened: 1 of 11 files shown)


ibis/backends/impala/__init__.py
```diff
@@ -137,6 +139,12 @@ def register_options(self):
         )
 
     def set_database(self, name):
+        warnings.warn(
```
cpcloud, line 142: Random thought to be done later: should we consider using https://deprecation.readthedocs.io/en/latest to automate the annoyances of deprecation?

cpcloud (approved):
LGTM.

ibis/backends/impala/__init__.py
datapythonista: Didn't know about the library, seems like a great idea.

jreback:
thanks @datapythonista

PR #2963, opened 2021-09-10
ci: only produce a test report if the workflow succeeded or failed

cpcloud:
This PR ensures the test report is only produced when a workflow suceeded or failed, and not for example in the case where the workflow was cancelled.


jreback (approved):

cpcloud:
@datapythonista It seems like PRs that were opened before the test reporter was merged are not receiving report comments. I'll open up a PR with a bogus failing test to see if that might be the case.

PR #2965, opened 2021-09-11
Raise NotImplementedError when databases are not supported

### LINKED ISSUE #2912: CLN: Better database API
The behavior for the database API is not ideal in my opinion. For SQL databases, we can have something like:
```python
>>> con.list_databases()
['db1', 'db2', 'db3']
>>> con.current_database
'db2'
```
In #2909, #2910 and #2911 I'm making the API more consistent, like making sure that `current_database` always exist, or `list_databases` has always the same API. But the behavior is not consistent with the above. For example, the pandas backend is returning an empty list for the list of databases, and `None` for the current database. Or the file backends returning the list of directories in the path as databases, and `None` as the current database.

What I think would make things simpler, is that backends that can't have more than one database, use `main` as the current_database, and `list_databases()` returns `['main']`. This seems consistent with what SQLAlchemy does with SQLite (which doesn't support databases). And would make code simpler, always having database names, and having the current database name in the list of databases.

I'd also implement this for the file backends. What I would do is to just support one "database", the current directory (except for hdf5, which does support databases). Or if anyone thinks it's useful, we could find all directories with files of the backend format, and list those directories as the databases (can be in a follow up once it's needed).

Let me know if there are any objections and I'll implement it this way.

### LINKED ISSUE #2962: Raise NotImplementedError if current_database and list_tables do not apply
Following the discussion on #2912, for backends that doesn't support databases, there are 3 options:

Option 1 (original)
```python
>>> con.current_database
None
>>> con.list_databases()
[]
```

Option 2 (new implemented version, `main` is an arbitrary name)
```python
>>> con.current_database
'main'
>>> con.list_databases()
['main']
```

Option 3
```python
>>> con.current_database
...
NotImplementedError: Backend foo does not support multiple databases
>>> con.list_databases()
...
NotImplementedError: Backend foo does not support multiple databases
```

I think option 1 makes sense in an interactive, but can introduce problems, if for example reusing code from a backend with databases with code reliying on `con.current_database in con.list_databases()`. Or having to handle special cases like `if con.current_database is None:`.

Option 2 solves this problem, but relying on a made up database name may not be ideal.

I would settle on option 3, where code won't fail silently. Maybe we can support it with a boolean attribute `con.supports_databases`, if we think users want to check that, in a nicer way than a try/except.

@cpcloud @jreback how does this sound?

datapythonista:
Closes #2962 
Closes #2912

### DIFF
--- a/ibis/backends/tests/conftest.py
+++ b/ibis/backends/tests/conftest.py
@@ -286,12 +286,11 @@ def temp_view(con) -> str:
 def current_data_db(con, backend) -> str:
     """Return current database name."""
     try:
-        current_database = con.current_database
+        return con.current_database
     except NotImplementedError:
         pytest.skip(
-            f'{backend.name()} backend doesn\'t have current_database method.'
+            f"{backend.name()} backend doesn't have current_database method."
         )
-    return current_database
 
 
 @pytest.fixture
--- a/ibis/backends/tests/conftest.py
+++ b/ibis/backends/tests/conftest.py
@@ -285,11 +285,13 @@ def temp_view(con) -> str:
 @pytest.fixture(scope='session')
 def current_data_db(con, backend) -> str:
     """Return current database name."""
-    if not hasattr(con, 'current_database'):
+    try:
+        current_database = con.current_database
+    except NotImplementedError:
         pytest.skip(
             f'{backend.name()} backend doesn\'t have current_database method.'
         )
-    return con.current_database
+    return current_database
 
 
 @pytest.fixture
--- a/docs/source/release/index.rst
+++ b/docs/source/release/index.rst
@@ -12,6 +12,7 @@ Release Notes
    These release notes are for versions of ibis **1.0 and later**. Release
    notes for pre-1.0 versions of ibis can be found at :doc:`release-pre-1.0`
 
+* :support:`2962` Methods `current_database` and `list_databases` raise an exception for backends that do not support databases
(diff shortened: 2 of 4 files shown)


ibis/backends/tests/conftest.py
```diff
@@ -285,11 +285,13 @@ def temp_view(con) -> str:
 @pytest.fixture(scope='session')
 def current_data_db(con, backend) -> str:
     """Return current database name."""
-    if not hasattr(con, 'current_database'):
+    try:
+        current_database = con.current_database
+    except NotImplementedError:
         pytest.skip(
             f'{backend.name()} backend doesn\'t have current_database method.'
```
cpcloud, line 292: How about a double quote to avoid the backslash?

ibis/backends/tests/conftest.py
```diff
@@ -285,11 +285,13 @@ def temp_view(con) -> str:
 @pytest.fixture(scope='session')
 def current_data_db(con, backend) -> str:
     """Return current database name."""
-    if not hasattr(con, 'current_database'):
+    try:
+        current_database = con.current_database
```
cpcloud, line 289: You can `return con.current_database` instead of creating a variable and returning later.

cpcloud (approved):

datapythonista:
Thanks for the feedback, good points. Made the changes.

PR #2966, opened 2021-09-12
ci: try using the specific commit for publishing

cpcloud:
This PR attempts to fix the test reporter to post comments to the right place.


datapythonista (approved):
lgtm. Seems like there is a failure on the CI

cpcloud:
Yup, I think there's a concurrency bug in the dask backend. I'm going to try to reproduce, and submit a PR to disable parallel testing if I can't, and also make an issue.

PR #2968, opened 2021-09-12
chore: disable dask parallel testing

cpcloud:
Dask tests seem unable to be run in parallel.

This disables parallel in the simpler backend testing until it can be reproduced and fixed.

I do think it's a bug that we can't execute tests in parallel without failure
but I don't think this should hold up the release.


jreback (approved):

jreback:
sure

PR #2970, opened 2021-09-12
test: see if the test report works

cpcloud:


cpcloud:
I _think_ this is working now. We need to see a few PRs with failing tests, by different contributors, to confirm.

PR #2974, opened 2021-09-13
chore: update pre-commit

cpcloud:
This PR updates `pre-commit` to use the latest isort and black.

Generally not a fan of these kinds of changes, but we have to update these
dependencies at some point.

Going forward, we should enable dependabot after we move to poetry, which will
automate this process and should result in smaller sets of changes.

Supercedes #2972.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -155,8 +155,7 @@ class BaseBackend(abc.ABC):
 
     # @abc.abstractmethod
     def table(self, name: str, database: str = None) -> ir.TableExpr:
-        """
-        """
+        """ """
         warnings.warn(
             '`database` argument of `.table()` is deprecated and '
             'will be removed in a future version of Ibis. Change '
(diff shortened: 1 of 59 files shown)


cpcloud:
How do people feel about batching a run of [`flynt`](https://github.com/ikamensh/flynt#flynt---string-formatting-converter) with this PR to avoid two PRs with a lot of formatting changes?

jreback:
> How do people feel about batching a run of [`flynt`](https://github.com/ikamensh/flynt#flynt---string-formatting-converter) with this PR to avoid two PRs with a lot of formatting changes?

no objection (there is also the pyupgrade set of actions which do this)

jreback (approved):

jreback:
actually this PR is fine.

ibis/backends/base/__init__.py
```diff
@@ -155,8 +155,7 @@ def list_tables(self, like: str = None, database: str = None) -> List[str]:
 
     # @abc.abstractmethod
     def table(self, name: str, database: str = None) -> ir.TableExpr:
-        """
-        """
+        """ """
```
jreback, line 158: should prob just kill these (can be another PR)

jreback:
minor comment for followup

ibis/backends/base/__init__.py
cpcloud: Yup.

PR #2976, opened 2021-09-13
chore: run pyupgrade on the repo

cpcloud:
This PR runs pyupgrade on the repo.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import abc
 import re
 import warnings
-from typing import Any, Callable, List, Type
+from typing import Any, Callable
 
 import ibis.expr.operations as ops
 import ibis.expr.schema as sch
@@ -24,7 +24,7 @@ class BaseBackend(abc.ABC):
     """
 
     database_class = Database
-    table_class: Type[ops.DatabaseTable] = ops.DatabaseTable
+    table_class: type[ops.DatabaseTable] = ops.DatabaseTable
 
     @property
     @abc.abstractmethod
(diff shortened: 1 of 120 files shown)


jreback:
wow this is a lot, but needed

ibis/backends/base/__init__.py
```diff
@@ -83,7 +83,7 @@ def current_database(self) -> str | None:
         """
 
     @abc.abstractmethod
-    def list_databases(self, like: str = None) -> List[str]:
+    def list_databases(self, like: str = None) -> list[str]:
```
jreback, line 86: i think this typing sytntax is fine, but requires `from __future__ import annotations` in every module that supports it in order for this to not break 3.7. odd that the CI is passing.

jreback (changes requested):

ibis/backends/base/__init__.py
cpcloud: `from __future__ import annotations` is already in this module. If you expand the file in the GitHub UI you can see it on line 1.

cpcloud:
@jreback `from __future__ import annotations` is already imported where needed (or added by `pyupgrade`). As you note, the CI wouldn't pass if it wasn't.

jreback:
> @jreback `from __future__ import annotations` is already imported where needed (or added by `pyupgrade`). As you note, the CI wouldn't pass if it wasn't.

ahh must have been added previously, ok!

jreback:
not a terrible idea to have pyupgrade as a periodic action, like once a month or something.

cpcloud:
We can run it as a pre commit check as well. Either way regular run is a good idea.

PR #2980, opened 2021-09-14
ci: rename workflow

cpcloud:
This is a small PR to rename the main CI workflow to `CI` instead of the much longer `Continuous Integration`
that tends to push useful information about various checks in PRs to the right and obscure it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,5 @@
-name: Continuous Integration
+# vim: filetype=yaml
+name: CI
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


jreback:
https://github.com/ibis-project/ibis/pull/2980/checks?check_run_id=3601278926 ?

cpcloud:
Hm, those failures are for the type parser, I think this might be related to the fact that I cancelled one of the type parser runs.

cpcloud:
@jreback I think this is good to go.

jreback (approved):

PR #2983, opened 2021-09-17
fix: fix literal of Nullable and order of limit & offset in Clickhouse.

Lunaticus7:
Fix two bugs of Clickhouse backend.

1. The order of limit and offset in compiled SQL is reversed 🤣.
  https://clickhouse.tech/docs/en/sql-reference/statements/select/limit/#limit-clause

2. When literal `None` to some Nullable column's type will raise error.

``` python
expr = table.mutate(ibis.literal(None, table['nullable_string_column'].type()).name('new_column'))
expr.compile()
# AttributeError: 'NoneType' object has no attribute 'replace'
```

### DIFF
--- a/ibis/backends/clickhouse/registry.py
+++ b/ibis/backends/clickhouse/registry.py
@@ -304,6 +304,8 @@ def _interval_from_integer(translator, expr):
 
 def _literal(translator, expr):
     value = expr.op().value
+    if value is None and expr._dtype.nullable:
+        return _null_literal(translator, expr)
     if isinstance(expr, ir.BooleanValue):
         return '1' if value else '0'
     elif isinstance(expr, ir.StringValue):
--- a/ibis/backends/clickhouse/tests/test_functions.py
+++ b/ibis/backends/clickhouse/tests/test_functions.py
@@ -500,6 +500,19 @@ def test_null_column(alltypes, translate):
     tm.assert_series_equal(result, expected)
 
 
+def test_literal_none_to_nullable_colum(alltypes):
+    t = alltypes
+    nrows = t.count().execute()
+    expr = t.mutate(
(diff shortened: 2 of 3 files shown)


Lunaticus7:
CI failed, emmmm, there is an import error ?

https://github.com/ibis-project/ibis/pull/2983/checks?check_run_id=3632398102#step:11:105

cpcloud:
@Lunaticus7 Unfortunately, we're now picking up a later version of impyla that uses a dependency that isn't declared in its `setup.py`. I'll patch the conda-forge package to include this dependency, hopefully I don't also need to package `puresasl` as well.

I've created an issue on the `impyla` issue tracker (https://github.com/cloudera/impyla/issues/471).

cpcloud:
I've put up a PR to add `pure-sasl` to the `impyla` conda package, which should at least allow CI to pass https://github.com/conda-forge/impyla-feedstock/pull/28

ibis/backends/clickhouse/registry.py
```diff
@@ -304,6 +304,8 @@ def _interval_from_integer(translator, expr):
 
 def _literal(translator, expr):
     value = expr.op().value
+    if value is None and expr._dtype.nullable:
```
cpcloud, line 307: Is the case of `value is None and not expr._dtype.nullable` handled? This could be done in a separate PR if you'd like.

Also, I think you should use the public `type()` API instead of digging around in `_` attributes :)

cpcloud (changes requested):
Some small nits.

Other than the CI (which doesn't really matter for this PR anyway, but we should still wait for it to pass) this LGTM.

Lunaticus7:
Closed, split this PR to https://github.com/ibis-project/ibis/pull/2984 https://github.com/ibis-project/ibis/pull/2985

cc @cpcloud

PR #2995, opened 2021-09-20
Call aggregation function directly for pandas Summarize

emilyreff7:
### Proposed Change
In the Pandas backend Summarize agg function, since we are already looping through the grouped data, we can simply get the aggregation results by calling the function directly rather than calling Pandas `agg`, which should save us some overhead.

### Tests
This is an internal refactor, all existing tests should pass.


jreback (approved):

cpcloud (approved):
Sweet. Net -1 line of code!

jreback:
FYI this actually does have a nice perf benefit, e.g we have seen a 10x improvement in a certain class of operations

cpcloud:
Awesome!

Issue #3000, opened 2021-09-22
refactor: remove ir.ExprList, ops.ExpressionList and ibis.expr_list()


kszucs:
My guess is that these constructs are a band aid for accepting multiple expressions instead of a single one - though I need to verify this. 

There is another operation called `ops.ValueList` which does almost the same but creates a `ir.ListExpr` instead.

According to a [github search](https://github.com/search?q=ibis+expr_list&type=code) the publicly exposed `api.expr_list()` function is not widely used.

datapythonista:
Checked that in the past, and I didn't find a use case for those, so happy to see them removed.

cpcloud:
@kszucs Is this done?

cpcloud:
Done in #3112

PR #3001, opened 2021-09-22
refactor(ir): Don't bypass argument coercion and validation for user defined functions

kszucs:
While adding UDF support for datafusion backend I recognized some issues with the UDF operations.

### DIFF
--- a/ibis/backends/dask/udf.py
+++ b/ibis/backends/dask/udf.py
@@ -79,9 +79,7 @@ def pre_execute_elementwise_udf(op, *clients, scope=None, **kwargs):
             return df
         else:
             name = args[0].name if len(args) == 1 else None
-            meta = pandas.Series(
-                [], name=name, dtype=op.return_type.to_dask()
-            )
+            meta = pandas.Series([], name=name, dtype=op.return_type.to_dask())
             df = dd.map_partitions(op.func, *args, meta=meta)
 
             return df
--- a/ibis/backends/dask/udf.py
+++ b/ibis/backends/dask/udf.py
@@ -26,8 +26,8 @@ def make_struct_op_meta(op: ir.Expr) -> List[Tuple[str, np.dtype]]:
     """Unpacks a dt.Struct into a DataFrame meta"""
     return list(
         zip(
-            op._output_type.names,
-            [x.to_dask() for x in op._output_type.types],
+            op.return_type.names,
+            [x.to_dask() for x in op.return_type.types],
         )
     )
 
@@ -72,7 +72,7 @@ def pre_execute_elementwise_udf(op, *clients, scope=None, **kwargs):
         # kwargs here. This is true for all udf execution in this
         # file.
         # See ibis.udf.vectorized.UserDefinedFunction
-        if isinstance(op._output_type, dt.Struct):
+        if isinstance(op.return_type, dt.Struct):
             meta = make_struct_op_meta(op)
 
             df = dd.map_partitions(op.func, *args, meta=meta)
@@ -80,7 +80,7 @@ def pre_execute_elementwise_udf(op, *clients, scope=None, **kwargs):
         else:
             name = args[0].name if len(args) == 1 else None
             meta = pandas.Series(
-                [], name=name, dtype=op._output_type.to_dask()
+                [], name=name, dtype=op.return_type.to_dask()
             )
             df = dd.map_partitions(op.func, *args, meta=meta)
 
@@ -124,11 +124,11 @@ def pre_execute_analytic_and_reduction_udf(op, *clients, scope=None, **kwargs):
         # Depending on the type of operation, lazy_result is a Delayed that
         # could become a dd.Series or a dd.core.Scalar
         if isinstance(op, ops.AnalyticVectorizedUDF):
-            if isinstance(op._output_type, dt.Struct):
+            if isinstance(op.return_type, dt.Struct):
                 meta = make_struct_op_meta(op)
             else:
                 meta = make_meta_series(
-                    dtype=op._output_type.to_dask(),
+                    dtype=op.return_type.to_dask(),
                     name=args[0].name,
                 )
             result = dd.from_delayed(lazy_result, meta=meta)
@@ -151,13 +151,13 @@ def pre_execute_analytic_and_reduction_udf(op, *clients, scope=None, **kwargs):
                 result = result.repartition(divisions=original_divisions)
         else:
             # lazy_result is a dd.core.Scalar from an ungrouped reduction
-            if isinstance(op._output_type, (dt.Array, dt.Struct)):
+            if isinstance(op.return_type, (dt.Array, dt.Struct)):
                 # we're outputing a dt.Struct that will need to be destructured
                 # or an array of an unknown size.
                 # we compute so we can work with items inside downstream.
                 result = lazy_result.compute()
             else:
-                output_meta = safe_scalar_type(op._output_type.to_dask())
+                output_meta = safe_scalar_type(op.return_type.to_dask())
                 result = dd.from_delayed(
                     lazy_result, meta=output_meta, verify_meta=False
                 )
@@ -181,7 +181,7 @@ def pre_execute_analytic_and_reduction_udf(op, *clients, scope=None, **kwargs):
         func = op.func
         groupings = args[0].index
         parent_df = args[0].obj
-        out_type = op._output_type.to_dask()
+        out_type = op.return_type.to_dask()
 
         grouped_df = parent_df.groupby(groupings)
         col_names = [col._meta._selected_obj.name for col in args]
@@ -223,7 +223,7 @@ def pre_execute_analytic_and_reduction_udf(op, *clients, scope=None, **kwargs):
         func = op.func
         groupings = args[0].index
         parent_df = args[0].obj
-        out_type = op._output_type.to_dask()
+        out_type = op.return_type.to_dask()
 
         grouped_df = parent_df.groupby(groupings)
         col_names = [col._meta._selected_obj.name for col in args]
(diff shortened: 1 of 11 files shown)


ibis/expr/operations.py
```diff
@@ -3623,23 +3624,17 @@ class GeoAsText(GeoSpatialUnOp):
 class ElementWiseVectorizedUDF(ValueOp):
     """Node for element wise UDF."""
 
-    func = Arg(callable)
-    func_args = Arg(tuple)
-    input_type = Arg(rlz.shape_like('func_args'))
-    _output_type = Arg(rlz.noop)
-
-    def __init__(self, func, args, input_type, output_type):
```
kszucs, line 3631: By redefining the constructor none of the input arguments were validated.

ibis/expr/operations.py
```diff
@@ -3623,23 +3624,17 @@ class GeoAsText(GeoSpatialUnOp):
...
+    input_type = Arg(rlz.list_of(rlz.datatype, coerce_to=list))
+    return_type = Arg(rlz.datatype)
 
     @property
     def inputs(self):
         return self.func_args
 
     def output_type(self):
-        return self._output_type.column_type()
+        return rlz.shape_like(self.func_args, dtype=self.return_type)
```
kszucs, line 3637: `my_add_function(1.0, 2.0)` returns with a scalar type rather than a column type - presuming that the `func_arg` are correctly converted to ibis literals.

ibis/expr/rules.py
```diff
@@ -147,7 +147,7 @@ def member_of(obj, arg):
 
 
 @validator
-def list_of(inner, arg, min_length=0):
+def list_of(inner, arg, min_length=0, coerce_to=ir.sequence):
```
kszucs, line 150: This validator actually returns with an `ir.ValueList` so the naming could be better, like `list_of_exprs`. We may revisit this later.

ibis/udf/vectorized.py
```diff
@@ -85,9 +85,9 @@ def func(*args):
 
         op = self.func_type(
             func=func,
-            args=args,
+            func_args=args,
             input_type=self.input_type,
-            output_type=self.output_type,
+            return_type=self.output_type,
```
kszucs, line 90: I'd prefer moving the files from `ibis/udf` directory to `ibis/expr/udf` and define the API utilities there as well, similarly to [SearchedCaseBuilder](https://github.com/ibis-project/ibis/blob/master/ibis/expr/operations.py#L1575).

cpcloud: That seems reasonable to me.

ibis/expr/rules.py
cpcloud: I suppose one could implement `value_list` etc in terms of `list_of`, now that you've added a `coerce_to` argument. OTOH removing some of the `*List*` APIs would probably be better than having a few more specific validators.

cpcloud (approved):
LGTM.

cpcloud:
@kszucs Any chance you can add a test to avoid a regression?

jreback:
cc @icexelloss @gerrymanoim

jreback:
this doesn't have any user facing changes right?

ibis/backends/dask/udf.py
```diff
@@ -223,7 +221,7 @@ def execute_analytic_node_groupby(op, *args, aggcontext, **kwargs):
         func = op.func
         groupings = args[0].index
         parent_df = args[0].obj
-        out_type = op._output_type.to_dask()
+        out_type = op.return_type.to_dask()
```
icexelloss, line 224: for naming consistency - I think it's better to call this `output_type`  since it's more common in ibis codebase

ibis/udf/vectorized.py
icexelloss: I originally put it under `ibis` because this is meant to be an API namespace and I think most of stuff under ibis/expr are not API modules?

ibis/expr/operations.py
```diff
@@ -3673,23 +3662,17 @@ def root_tables(self):
...
-    func_args = Arg(tuple)
-    input_type = Arg(rlz.shape_like('func_args'))
-    _output_type = Arg(rlz.noop)
-
-    def __init__(self, func, args, input_type, output_type):
-        self.func = func
-        self.func_args = args
-        self.input_type = input_type
-        self._output_type = output_type
+    func = Arg((types.FunctionType, types.LambdaType))
```
icexelloss, line 3665: Nice!

ibis/expr/operations.py
icexelloss: what is `my_add_function`?

gerrymanoim:
LGTM

ibis/expr/operations.py
icexelloss: I am ok with this change but also seems a bit unnecessary to handle scalar types for this class.

The original thinking was that the vectorized udf to not handle scalars at all for simplicity (hence the same "vectorized"). If a scalar is passed it, it should probably be converted to a column literal. i.e., "1.0" meaning a vector of all 1.0s

icexelloss:
Thanks @kszucs for adding validations for vectorized UDFs. Do you mind adding some tests for this?

icexelloss:
> While adding UDF support for datafusion backend I recognized some issues with the UDF operations.

I am curious - datafusion is a rusty implementation? how does it use the vectorized UDF classes?

ibis/expr/operations.py
kszucs: See https://github.com/ibis-project/ibis/blob/master/ibis/expr/signature.py#L180-L183

ibis/expr/operations.py
kszucs: Sorry, missed the context. 

I ported a test case from the [pandas backend](https://github.com/ibis-project/ibis/blob/master/ibis/backends/pandas/tests/test_udf.py#L115-L118) to the [datafusion backend](https://github.com/ibis-project/ibis/pull/2918/files#diff-009e15b368654dd7b798e7acef1ac005a55a91a8f815515ae3fadf31f37e90cdR49-R52) and got shape errors. 

Indeed, vectorized should imply that we handle array-like inputs, though the pandas backend tests it for scalar inputs. 
I'm fine with keeping array-like output for vectorized UDFs, but than we may need non-vectorized user defined functions as well?

kszucs:
> this doesn't have any user facing changes right?

It may, depending on how UDFs were exposed previously. I'd hold off merging this PR until the next release, unless someone confirms that it's not a breaking change.

ibis/backends/dask/udf.py
kszucs: `output_type` is a method of the operation classes, I think that was the main reason to use an underscored alternative.

jreback:
> > this doesn't have any user facing changes right?
> 
> It may, depending on how UDFs were exposed previously. I'd hold off merging this PR until the next release, unless someone confirms that it's not a breaking change.

ok perfect, pls add a release note as well,

ibis/expr/rules.py
kszucs: I'm not really satisfied with this keyword argument, so I'm open to alternatives. Perhaps https://github.com/ibis-project/ibis/issues/3000 will clean this up a little bit.

jreback:
created a 2.x milestone for a few PRs @kszucs @cpcloud

kszucs:
> > While adding UDF support for datafusion backend I recognized some issues with the UDF operations.
> 
> I am curious - datafusion is a Rust implementation? how does it use the vectorized UDF classes?

I have an open PR with support for ["vectorized elementwise" UDFs](https://github.com/ibis-project/ibis/pull/2918/files#diff-009e15b368654dd7b798e7acef1ac005a55a91a8f815515ae3fadf31f37e90cdR21). 
Datafusion is a query engine implemented in rust on top of the rust arrow implementation. Thanks to the [Arrow C data interface](https://arrow.apache.org/docs/format/CDataInterface.html) we can exchange data between different arrow implementations (without copying), so the datafusion python bindings can accept and return pyarrow objects.

kszucs:
> @kszucs Any chance you can add a test to avoid a regression?

Definitely, though I got a bit confused about the actual responsibility of a vectorized UDF based on https://github.com/ibis-project/ibis/pull/3001#discussion_r714301274

icexelloss:
@cpcloud Sorry for the delay I will take a look later Today

kszucs:
Depends on https://github.com/ibis-project/ibis/pull/3105

ibis/expr/operations/vectorized.py
```diff
@@ -7,79 +9,39 @@
...
-
-    def __init__(self, func, args, input_type, output_type):
-        self.func = func
-        self.func_args = args
-        self.input_type = input_type
-        self._output_type = output_type
+class VectorizedUDF(ValueOp):
+    func = Arg(rlz.instance_of((FunctionType, LambdaType)))
+    func_args = Arg(rlz.list_of(rlz.any))
+    input_type = Arg(rlz.list_of(rlz.datatype))
```
kszucs, line 15: Shall we rename `input_type` to `input_types`?

ibis/udf/vectorized.py
kszucs: Deferring it to a follow-up.

ibis/expr/operations.py
cpcloud: Let's keep it unchanged for now. It seems like a use case that isn't necessary or desirable to support at the moment.

kszucs: I can't really keep it unchanged since we do validate the inputs now.

It would make sense to support both column and array scalar inputs:

```py
class VectorizedUDF(ValueOp):
    func = Arg(rlz.instance_of((FunctionType, LambdaType)))
    func_args = Arg(
        rlz.list_of(rlz.one_of([rlz.column(rlz.any), rlz.scalar(rlz.array)]))
    )
    input_type = Arg(rlz.list_of(rlz.datatype))
    return_type = Arg(rlz.datatype)
```

Though I'm not sure how could we handle the mixture of those input types without explicit transpose of arrays to columns.

cpcloud (approved):

PR #3009, opened 2021-09-23
refactor(ir): Remove ir.ExprList, ops.ExpressionList and ibis.expr_list() [WIP]

kszucs:
Resolves #3000, this is an API breaking change.

Seems like `ops.ExpressionList` was mainly used to group the value expressions of summary into a standalone expression in order to execute it by itself or through `table.aggregate(..)`

### DIFF
--- a/ibis/backends/clickhouse/tests/test_client.py
+++ b/ibis/backends/clickhouse/tests/test_client.py
@@ -135,18 +135,18 @@ def test_table_info(alltypes):
     assert buf.getvalue() is not None
 
 
-def test_execute_exprs_no_table_ref(con):
-    cases = [(L(1) + L(2), 3)]
+# def test_execute_exprs_no_table_ref(con):
+#     cases = [(L(1) + L(2), 3)]
 
-    for expr, expected in cases:
-        result = con.execute(expr)
-        assert result == expected
+#     for expr, expected in cases:
+#         result = con.execute(expr)
+#         assert result == expected
 
-    # ExprList
-    exlist = ibis.api.expr_list(
-        [L(1).name('a'), ibis.now().name('b'), L(2).log().name('c')]
-    )
-    con.execute(exlist)
+#     # ExprList
+#     exlist = ibis.api.expr_list(
+#         [L(1).name('a'), ibis.now().name('b'), L(2).log().name('c')]
+#     )
+#     con.execute(exlist)
 
 
 @pytest.mark.skip(reason="FIXME: it is raising KeyError: 'Unnamed: 0'")
--- a/ibis/backends/clickhouse/tests/test_client.py
+++ b/ibis/backends/clickhouse/tests/test_client.py
@@ -4,10 +4,8 @@ import pandas as pd
 import pandas.testing as tm
 import pytest
 
-import ibis
 import ibis.config as config
 import ibis.expr.types as ir
-from ibis import literal as L
 
 
 def test_get_table_ref(db):
--- a/ibis/backends/impala/tests/test_exprs.py
+++ b/ibis/backends/impala/tests/test_exprs.py
@@ -1010,16 +1010,16 @@ def test_table_info(alltypes):
     assert buf.getvalue() is not None
 
 
-@pytest.mark.parametrize(('expr', 'expected'), [(L(1) + L(2), 3)])
-def test_execute_exprs_no_table_ref(con, expr, expected):
-    result = con.execute(expr)
-    assert result == expected
+# @pytest.mark.parametrize(('expr', 'expected'), [(L(1) + L(2), 3)])
+# def test_execute_exprs_no_table_ref(con, expr, expected):
+#     result = con.execute(expr)
+#     assert result == expected
 
-    # ExprList
-    exlist = ibis.api.expr_list(
-        [L(1).name('a'), ibis.now().name('b'), L(2).log().name('c')]
-    )
-    con.execute(exlist)
+#     # ExprList
+#     exlist = ibis.api.expr_list(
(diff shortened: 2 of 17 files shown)


ibis/tests/expr/test_table.py
```diff
@@ -523,13 +523,14 @@ def test_groupby_alias(table):
...
-    metric = table.g.group_concat().name('bar')
-    result = table.aggregate([metric, summ])
-    expected = table.aggregate([metric] + summ.exprs())
-    assert_equal(result, expected)
+# TODO(kszucs): we should perhaps change the API to
+# table.aggregate(metrics=table.f.summary()) ?
+# def test_summary_expand_list(table):
+#     summ = table.f.summary()
+#     metric = table.g.group_concat().name('bar')
+#     result = table.aggregate([metric, summ])
```
cpcloud, line 531: It seems like this can be done by the `aggregate` method by something like this:

```python
exprs = []
for expr in inputs:
    op = expr.op()
    if isinstance(op, ops.Summary):
        exprs.append(op.exprs)
    else:
        exprs.append(expr)
```

ibis/tests/sql/test_compiler.py
```diff
@@ -220,18 +220,18 @@ def test_scalar_exprs_no_table_refs(self):
...
-                ibis.literal(1).name('a'),
-                ibis.now().name('b'),
-                ibis.literal(2).log().name('c'),
-            ]
-        )
-        result = Compiler.to_sql(exlist)
-        expected = """\
-SELECT 1 AS `a`, now() AS `b`, ln(2) AS `c`"""
-        assert result == expected
+    #     def test_expr_list_no_table_refs(self):
```
cpcloud, line 223: I think it's fine to remove this code instead of commenting it out since we're explicitly removing the API in this PR.

ibis/tests/expr/test_value_exprs.py
```diff
@@ -891,40 +891,6 @@ def expr():
     return ibis.expr_list(exprs)
 
 
-def test_names(expr):
```
cpcloud, line 894: Should the `expr` fixture also be removed?

ibis/backends/impala/tests/test_exprs.py
```diff
@@ -1010,39 +1010,38 @@ def test_table_info(alltypes):
...
+#     expr = table.double_col.summary()
+#     repr(expr)
+
+#     result = expr.execute()
+#     assert isinstance(result, pd.DataFrame)
+
+#     expr = table.group_by('string_col').aggregate(
+#         [
+#             table.double_col.summary(prefix='double_'),
+#             table.float_col.summary(prefix='float_'),
```
cpcloud, line 1040: Per my comment below I think this can be handled in the `aggregate` method.

ibis/backends/impala/tests/test_exprs.py
```diff
@@ -1010,39 +1010,38 @@ def test_table_info(alltypes):
...
-        [
-            table.double_col.summary().prefix('double_'),
-            table.float_col.summary().prefix('float_'),
-            table.string_col.summary().suffix('_string'),
-        ]
-    )
-    result = expr.execute()
-    assert isinstance(result, pd.DataFrame)
+# @pytest.mark.parametrize(('expr', 'expected'), [(L(1) + L(2), 3)])
+# def test_execute_exprs_no_table_ref(con, expr, expected):
```
cpcloud, line 1014: I think this test is fine, just replace the use of `expr_list` with a regular `list`.

ibis/backends/clickhouse/tests/test_client.py
```diff
@@ -135,18 +133,18 @@ def test_table_info(alltypes):
     assert buf.getvalue() is not None
 
 
-def test_execute_exprs_no_table_ref(con):
-    cases = [(L(1) + L(2), 3)]
+# def test_execute_exprs_no_table_ref(con):
```
cpcloud, line 136: Same as below I think we can use a `list` here and keep the test as is otherwise.

cpcloud (changes requested):
I'm glad this was able to happen, the various expression lists are a really confusing API IMO and a rather annoying maintenance burden.

A few tweaks and this should be good to go.

ibis/backends/clickhouse/tests/test_client.py
kszucs: My plan was the same, but requires more work on the SQL backends since expr would be a list [here](https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/compiler/query_builder.py#L561-L582).

ibis/tests/expr/test_table.py
kszucs: That could be the solution perhaps!

We don't yet have a `Summary` operation, we expose the summaries as a [list of expressions](https://github.com/ibis-project/ibis/blob/master/ibis/expr/api.py#L1232-L1239). 

Though that would mean that we shouldn't enforce an explicit [output type for all of the operations](https://github.com/ibis-project/ibis/blob/master/ibis/expr/operations.py#L156-L161).

ibis/backends/impala/tests/test_exprs.py
kszucs: Right, we can handle this case and flatten the input metrics, but that mean we couldn't keep the following feature:

```py
table.column.summary().execute()
```

kszucs:
@cpcloud could we defer this to 3.0?

It's still unclear to me what kind of API do we want to provide for:
- summary metrics 
- list of unbound expressions

kszucs:
Closing in favor of https://github.com/ibis-project/ibis/pull/3112

PR #3010, opened 2021-09-23
CLN: Merge alchemy table methods into base class

datapythonista:


cpcloud:
@datapythonista Are these breaking API changes? If not, can we release 2.0 before merging this?

datapythonista:
I thought this would be a quick clean up, but seems there is something subtle breaking tests, I'll leave it for a later time.

PR #3011, opened 2021-09-23
CLN: Move Clickhouse Client class to Backend

datapythonista:
xref #2678

Replacing the `Client` class of the clickhouse backend by its `Backend` class.


cpcloud:
@datapythonista same question as #3010: Are these breaking API changes? If not, can we release 2.0 before merging this?

datapythonista:
> @datapythonista same question as #3010: Are these breaking API changes? If not, can we release 2.0 before merging this?

I personally think this (for all backends) is the main thing we should have implemented before the release. This shouldn't have any impact on Ibis users, so no API changes in this sense.

But with Ibis 2, we're implementing a "official" way to implement third-party backends. And so far we moved bigquery and omnisci out, besides the already existing (mssql, and not sure if any other relevant one). I'd also move clickhouse and impala out before the release, but that's a separate topic.

With an ecosystem of backends now officially supported which will leave in separate repos, and be released independently from Ibis, there is a stronger requirement for compatibility. If we release Ibis as is, we update all backends to support the current backend API, and then we merge this PR, all third party backends will be broken. That's why the focus of a lot of work for the Ibis 2 release was into specifying the backend API as clearly as possible, with a base class, standardized signatures...

As this is has been much more complex than initially expected, and is taking forever, I fully agree that we should try to move faster, and release asap. But having a single class for `Backend` and `Client`, even if it'll delay things a bit more, I think it'll add a lot of value in being able to offer a decent level of backward compatibility in future changes.

Does this make sense?

cpcloud:
@datapythonista Makes sense. Do we have issues for moving each of the backends? I'm happy to move the postgres backend out and own its maintenance and release.

datapythonista:
> @datapythonista Makes sense. Do we have issues for moving each of the backends? I'm happy to move the postgres backend out and own its maintenance and release.

We've got the general #2678 to update backends to a better API. For moving backends out of the main repo there are no issues. To me personally, it makes sense to start with Impala and Clickhouse, since then we can easily port them, or implement new backends in parallel, that use SQLAlchemy. And when that happens, we can get rid of the original Ibis SQL compiler, and use only SQLAlchemy. Or if it takes too long, we can move the compiler to a third-party for these two backends, and leave Ibis with just the SQLAlchemy compiler.

Then, I'd probably move pyspark and dask out. And I'd leave the SQLAlchemy to the end, since I think it's API can be improve. For example, using a class instead of a dictionary for the operation registry, the `.table()` clean up you've seen, or trying to make backend implementations more like config files, than code. I think those fixes should make backend development and maintenance easier, so more backends can be easily supported. And I think it's probably easier to make all these changes with the SQLALchemy backends in this same repo.

For Ibis 2, I'd just move out Impala and Clickhouse, which should be quite fast. But no big deal if we move after the release.

In #2828 I tried to explain this vision. Feedback surely welcome.

cpcloud:
Sounds good. The only wrinkle I see is with moving Impala to a SQLAlchemy based backend. I don't know if there's enough interest to warrant the effort, to be totally frank.

I suspect we want to move the impala backend out along with the string-based SQLCompiler, to its own corner of GitHub.

Anyway, I'll open up issues for the specific backends.

datapythonista:
> I suspect we want to move the impala backend out along with the string-based SQLCompiler, to its own corner of GitHub.

In my mind, if we improve the base SQLALchemy backend, it shouldn't be that time consuming to implement backends. That probably doesn't apply to Impala, which has more things than other backends. So I agree we may just port clickhouse (and bigquery and omnisci which are already planned), and then make the SQL compiler part of the Impala backend. But we can see in detail after the release.

jreback:
i don't think we need to move anything befor 2.0 - this is just going to delay things even more

just fix the compatibility issue

datapythonista:
Superseded by #3014. Once the code is ready to move backends with that PR, I'll change clickhouse (and the other backends) to use only `Backend` instead of `Backend+Client`.

PR #3013, opened 2021-09-24
refactor: remove __del__ from pandas_interop

cpcloud:
This PR removes the `__del__` method from `DataFrameWriter` in favor of a context manager.

### DIFF
--- a/ibis/backends/impala/client.py
+++ b/ibis/backends/impala/client.py
@@ -1,3 +1,4 @@
+import contextlib
 import io
 import operator
 import re
@@ -43,6 +44,7 @@ from . import ddl, udf
 from .compat import HS2Error, ImpylaError, impyla
 from .compiler import ImpalaCompiler
 from .hdfs import HDFS, WebHDFS
+from .pandas_interop import DataFrameWriter
 
 
 class ImpalaDatabase(Database):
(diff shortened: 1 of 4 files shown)


datapythonista (approved):
Looks good.

Not ideal having nested blocks so big, but feels like it's worth in this case.

Only thing is that I would maybe rename `_setup_insert` to something more descriptive of that it returns. `_insert_values_as_expr`, `_parse_insert_values`... But no big deal.

cpcloud:
> Looks good.
> 
> Not ideal having nested blocks so big, but feels like it's worth in this case.
> 
> Only thing is that I would maybe rename `_setup_insert` to something more descriptive of that it returns. `_insert_values_as_expr`, `_parse_insert_values`... But no big deal.

How about I annotate the return values? Then we don't have to bake type information into the method name :)

datapythonista:
> How about I annotate the return type of the method? Then we don't have to bake type information into the method name :)

It's a minor comment, feel free to ignore, or add the type information as you say. But I found useful having a more descriptive name whenreading where the method is used, than when reading the signature. Adding types won't help in that case.

So, to be specific, I find this a bit mysterious on what's happening here, when I read this code:
```python
with self._client._setup_insert(obj) as expr:
```
And I think something like next (any more specific name would work) could help:
```python
with self._client._dataframe_to_temp_expr(obj) as expr:
```
But as said, no big deal.

jreback:
thanks @cpcloud

PR #3014, opened 2021-09-24
Prepare backends to merge Client methods into Backend

datapythonista:
This is a bit hacky, but it's the only way I found to be able to move methods from Client to Backend one backend at a time.

With this PR `connect` can return both `Client` (as now) or `Backend` (as we are moving to).

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -33,10 +33,36 @@ class BaseBackend(abc.ABC):
         Name of the backend, for example 'sqlite'.
         """
 
+    # XXX For backends where `Backend` and `Client` coexist, `client`,
+    # `client_class` and `backend` will be overwritten by the associated
+    # instances. For backends where `Client` was fully removed, these
+    # properties will be used instead, using the `Backend` class, containing
+    # all `Client` methods
+    # TODO remove once all backends have been ported to the new API
     @property
-    @abc.abstractmethod
+    def client(self):
+        return self._set_client if hasattr(self, '_set_client') else self
+
+    @client.setter
+    def client(self, value):
+        self._set_client = value
+
+    @property
     def client_class(self):
-        """Class of the client, like `PandasClient`."""
+        return (
+            self._set_client_class
+            if hasattr(self, '_set_client_class')
+            else self.__class__
+        )
+
+    @client_class.setter
+    def client_class(self, value):
+        self._set_client_class = value
+
+    def backend(self):
+        return self
+
+    # End of TODO
 
     @abc.abstractmethod
     def connect(connection_string, **options):
@@ -260,3 +286,17 @@ class BaseBackend(abc.ABC):
             )
 
         return decorator
+
+
+class BaseBackendForClient(BaseBackend):
+    # This class is implemented so `Client` can implement from it, without
+    # having to overwrite abstract methods.
+    # We want clients to subclass backend, so then they can be merged to
+    # backend without requiring further changes to Ibis, only to the backend
+    # being moved.
+    # TODO this class should be deleted when all subclasses of `Client` have
+    # been deleted
+    name = None
+
+    def connect(connection_string, **options):
+        pass
--- a/ibis/backends/clickhouse/client.py
+++ b/ibis/backends/clickhouse/client.py
@@ -10,6 +10,7 @@ import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
 import ibis.expr.schema as sch
 import ibis.expr.types as ir
+from ibis.backends.base import BaseBackendForClient
 from ibis.backends.base.sql import SQLClient
 from ibis.config import options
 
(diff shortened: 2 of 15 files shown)


datapythonista:
@jreback @cpcloud can you have a look here? I need this to continue the work on merging `Client` to `Backend`.

ibis/backends/base/__init__.py
```diff
@@ -33,10 +33,36 @@ def name(self) -> str:
...
         """
 
+    # XXX For backends where `Backend` and `Client` coexist, `client`,
+    # `client_class` and `backend` will be overwritten by the associated
+    # instances. For backends where `Client` was fully removed, these
+    # properties will be used instead, using the `Backend` class, containing
+    # all `Client` methods
+    # TODO remove once all backends have been ported to the new API
+    @property
+    def client(self):
```
jreback, line 43: umm this is like 3 layers of indirection. why not just `_client`

ibis/backends/base/__init__.py
```diff
@@ -33,10 +33,36 @@ def name(self) -> str:
...
+
+    @client.setter
+    def client(self, value):
+        self._set_client = value
+
     @property
-    @abc.abstractmethod
     def client_class(self):
-        """Class of the client, like `PandasClient`."""
+        return (
```
jreback, line 52: same question  this is a lot of idirection

jreback (changes requested):
i don't think you are actually testing wether this works, do we have a backend that doesn't have a client now?

datapythonista:
The goal here is to have the backends working as they are now, and when `Client` is removed and all methods have been moved to `Backend`. The levels of indirection are just going to exist while the backends are moved, and after that, the properties you're referring to can be deleted, and there will be zero levels of indirection.

The challenge here is that we've got like 12 or 14 backends, and it's very tricky to make this change with a huge PR moving all them at the same time. Allowing both the old and new backend structures is quite challenging, I tried several approaches now, and none of them is great. There are many tests that fail with side effects, and hacks that are required. And this PR doesn't even solve all that, since there are the intermediate base `Clients` (alchemy, SQL, and file). I don't think we'll be able to move one backend at a time, I think minimum we'll have to move all pandas backends together, all alchemy together, and all SQL together (and alchemy backends depend at the moment on the base SQL backend, so not even sure if SQL and alchemy will have to be move together).

So, the options I see to move this forward are:
1. Move all backends in a massive PR
2. Move backends one at a time, breaking the tests for the not yet moved, and not having the CI green until all them have been moved
3. Use a hacky approach like this PR, and some follow up PR for the base clients
4. Let me know if you have a better idea

jreback:
I am not sure why we cannot just leave the code in place, release, then come back later and clean up after the backends are moved.

datapythonista:
> I am not sure why we cannot just leave the code in place, release, then come back later and clean up after the backends are moved.

When we move a backend and remove `Client`, we start returning a `Backend` instance when `connect()` is called. Since there is a `BaseBackend` class that is reused by both the legacy backends, and the new backends, it needs to support both. New backends don't have a `self.client` reference to `Client`, since it doesn't exist, so the new backend will fail if those are not updated. And if they are updated, the old backends will fail. That's why this PR implements a hybrid `self.client`, that is able to return whatever class makes sense, depending if the backend is new or old.

And there are similar things like that addressed here, but not all. This is a first step in the right direction, but it's still quite tricky to get both legacy and new backends working at the same time. The option to start moving backends breaking the old ones in the CI while waiting to be moved would make things way easier.

jreback:
@datapythonista i think you are missing my point.

Why cannot we simply make a *new* interface and just leave the existing entirely alone. Then once everything is moved you delete it. This hybrid approach is *very* challenging as you have noted.

we want to have backends in a good state either IN ibis proper or in separate repos, but this is going to take some time. We *need* flexiblity and cannot force movement at a super quick pace. 

What i suggest is that we basically leave Client around for a while, maybe make a LegacyBackend to handle the compatiblity things and then we can move backends one by one form Legacy to Backend.

datapythonista:
I see. Backend and Client are already coupled, so it wouldn't be a new interface replacing Client, but replacing both. Then, if we decide to do this, and we don't port all backends to the new interface, and we support the legacy one in Ibis 2.0, then there is no reason to not release now with what we've got.

Another solution could be continue as we did until now, moving methods from `Client` to `Backend`, but virtually keeping them in `Client`, do that one backend at a time, and then make `connect` return `Backend` instead of `Client` all backends at a time. Still quite a complex approach, with some big PRs, but more manageable than other approaches probably.

Moving backends out of this repo, besides something useful for other reasons, would help with this, probably worth considering too.

I think first question is if we want to release without further changes, which is not ideal, but probably the only way to release in the very short term. Depending on the answer to that, we can discuss what approach makes more sense. To summarize options:
1. Move everything with a huge PR
2. Move all methods from Client to Backend first as we've been doing, and then fully replace Client with all backends at a time
3. Try to implement a hybrid approach that supports both backends at the same time
4. Move one backend at a time, breaking the backends not yet moved in the CI
5. Move a significant amount of backends to other repos, so option 1 or 2 are more feasible

datapythonista:
Will move forward with the approach described in #3017, which moves us significantly into the right direction, and should be mostly trivial to implement.

PR #3015, opened 2021-09-24
CLN: Removing repeated methods from dask backend

datapythonista:
We'll be moving the methods from Client to Backend shortly, probably worth deleting all the duplicate ones from dask, and just moving the needed ones.

### DIFF
--- a/ibis/backends/dask/client.py
+++ b/ibis/backends/dask/client.py
@@ -1,6 +1,6 @@
 """The dask client implementation."""
 from functools import partial
-from typing import Dict, Mapping
+from typing import Mapping
 
 import dask.dataframe as dd
 import dateutil.parser
@@ -15,10 +15,11 @@ import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 import ibis.expr.schema as sch
 import ibis.expr.types as ir
-from ibis.backends.base import Client, Database
+from ibis.backends.base import Database
 from ibis.backends.pandas.client import (
     PANDAS_DATE_TYPES,
     PANDAS_STRING_TYPES,
+    PandasClient,
     _inferable_pandas_dtypes,
     convert_timezone,
     ibis_dtype_to_pandas,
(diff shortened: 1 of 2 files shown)


ibis/backends/dask/client.py
```diff
@@ -156,18 +157,7 @@ class DaskDatabase(Database):
...
-        self.database_class = backend.database_class
-        self.table_class = backend.table_class
-        self.dictionary = dictionary
-
-    def table(self, name: str, schema: sch.Schema = None) -> DaskTable:
-        df = self.dictionary[name]
-        schema = sch.infer(df, schema=schema)
-        return self.table_class(name, schema, self).to_expr()
-
+class DaskClient(PandasClient):
```
cpcloud, line 160: I'm not sure it makes sense conceptually to answer "yes" to the question "Is a dask client a pandas client?"

I don't think deduplicating code is a strong enough rationale for coupling these two components.

Now, `DaskClient`s would inherit any change to `PandasClient` APIs that are now duplicated.

I could probably get behind a more specific parent class of both `PandasClient` and `DaskClient`, like `DataFrameClient`, where the duplication was reduced but pandas and dask aren't coupled.

cpcloud (changes requested):

ibis/backends/dask/client.py
datapythonista: Agree, that's how it is implemented for `Backend`. The name is `BasePandasBackend`, since it's not a generic dataframe backend, but a pandas-API base backend.

Whenever we move the methods to `Backend`, they'll have the structure you mention. I'm ok not merging this, and consolidating the methods when we move to `Backend`, but I thought having this early should simplify the other PR.

cpcloud: I see. That makes sense then. Can you add a `TODO` with a some of what you've written here? Maybe something like:

```
TODO: these are coupled here to remove duplication, and are uncoupled in the Backend refactor
```

jreback:
cc @gerrymanoim I suppose this passes all tests, but am a bit leary of doing this.

datapythonista:
I don't think we should be copying code like this, but if this is controversial, happy to leave it for another time after the release.

PR #3016, opened 2021-09-24
Copy base SQL/Alchemy client methods into base Backends

datapythonista:
In order to move one backend at a time from `Client` to `Backend` we need the methods of `Client` in both classes. So backends with the old and new API can coexist. The idea here is to duplicate the methods, and after the last backend using the base class is moved, the whole `client.py` can be moved.


datapythonista:
Will move forward with the approach described in #3017, which moves us significantly into the right direction, and should be mostly trivial to implement.

Issue #3017, opened 2021-09-25
refactor: move methods implemented in Client to corresponding Backend methods


datapythonista:
xref #2678

Follow up of discussion in #3014

Seems like it's quite challenging at this point to make `Backend.connect()` methods return an instance of `Backend. See [this comment](https://github.com/ibis-project/ibis/pull/3014#issuecomment-927156539) for a list of the possible options we've got.

Regardless of implementing this particular change (making `connect` return `Backend` instead of `Client`), we've been working on moving methods from `Client` classes to `Backend` classes, and then creating a proxy method in `Client` that calls the method in the new location. This was done while standardizing the API.

A first step towards the new backend API structure would be to finish moving all the remaining methods. This will need to eventually be done. And even if we don't fully remove `Client`, or if the API is not fully specified or standard, it'll greatly simplify the code of backends having all methods in the same classes, and not arbitrarily split between `Client` and `Backend`.

To move forward with this approach, the next PRs will be open:
- Delegate all objects not found in `Client` to the corresponding `Backend` instance
- Move methods from `Client` classes into `Backend`, one backend at a time

cpcloud:
Partially addressed by #3127.

cpcloud:
@datapythonista Does #3127 fully address this issue?

PR #3018, opened 2021-09-26
Move methods of SQL backends from Client to Backend

datapythonista:
xref #3017

The diff is big, but the changes here are quite simple:
- Moving all methods of SQL backends from `Client` to `Backend`
- Creating proxy methods in `Client`, so the `Backend` methods are called
- Replacing `self.client` by `self` where needed (since attributes of `Backend` can now be called from `Backend` directly)
- Updates to tests. Mainly `MockConnection` and how tests are ignored when `create_table` doesn't exist.


ibis/backends/base/client.py
```diff
@@ -29,6 +29,68 @@ def exists_database(self, name):
     def exists_table(self, name, database=None):
         return self.backend.exists_table(name, database)
 
+    # Non-standard methods
```
jreback, line 32: make the comment that these are for compatiblity, ref an issue here (e.g. to remove later in time)

jreback: ideally if you can have the same signature; the doc-string not so important

ibis/backends/base/client.py
```diff
@@ -29,6 +29,68 @@ def exists_database(self, name):
...
+    def compile(self, *args, **kwargs):
+        return self.backend.compile(*args, **kwargs)
+
+    def create_table(self, *args, **kwargs):
+        if not hasattr(self.backend, 'create_table'):
+            raise NotImplementedError
+        return self.backend.create_table(*args, **kwargs)
+
+    def drop_table(self, *args, **kwargs):
+        if not hasattr(self.backend, 'drop_table'):
```
jreback, line 52: why is this needed, shouldn't this notimplemented exist on the BaseBackend?

ibis/backends/base/client.py
jreback: event better you could put these in a separate mixin class e.g. ClientDispatchToBackend or ClientCompat or something and just mix here. (then its also easy for a Backend to opt *out* of compatiblity once they are good.

jreback (changes requested):
cconceptually +1 on this way. suggestions in line

datapythonista:
Thanks for the feedback @jreback. Good points, addressed your comments.  I don't think the mixin class is a good idea at this point. The methods there aren't specific to a backend (yet). Most of them should be standard probably, it's just that we didn't have the time to decide what should be in the common backend API, standardize their signatures and create the abstract methods. I think it'll make things easier for the future to keep all methods as they are, and when the API is fully defined, consider if we want to move the methods in the standard API to mixin classes. I created the issue #3019 and linked it in the comment to have this information there. Does it make sense?

datapythonista:
@jreback @cpcloud this is finally green. Sorry for the big diff, but I think with this we're almost ready for the release. We should open a similar PR for the pandas based backends, and no backend in this repo should implement much in the client classes, but the constructor. We can have a look to see if we can easily move the stuff from the constructor to `connect`, and then backends should not have anything in Client, and we can remove the Client code after the release, without requiring changes to the backends.

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -43,30 +54,369 @@
...
-        databases = self.client.inspector.get_schema_names()
+        databases = self.inspector.get_schema_names()
         return self._filter_with_like(databases, like)
+
+    @property
+    def inspector(self):
+        self._inspector.info_cache.clear()
+        return self._inspector
+
+    @staticmethod
```
jreback, line 110: we typically don't use staticmethods as this is an anti-pattern.

ibis/backends/base/client.py
```diff
@@ -29,6 +29,127 @@ def exists_database(self, name):
     def exists_table(self, name, database=None):
         return self.backend.exists_table(name, database)
 
+    # Methods below haven't been standardized among backends yet, #3019
```
jreback, line 32: what is wrong with a Mixin?

jreback (changes requested):

ibis/backends/base/sql/__init__.py
```diff
@@ -9,3 +18,200 @@ class BaseSQLBackend(BaseBackend):
...
+        Parameters
+        ----------
+        query : string
+          DML or DDL statement
+
+        Returns
+        -------
+        Backend cursor
+        """
+        # TODO results is unused, it can be removed
```
cpcloud, line 82: I don't think this comment is valid anymore.

ibis/backends/base/sql/__init__.py
```diff
@@ -9,3 +18,200 @@ class BaseSQLBackend(BaseBackend):
...
+        # (requires updating Impala tests)
+        # TODO `self.con` is assumed to be defined in subclasses, but there
+        # is nothing that enforces it. We should find a way to make sure
+        # `self.con` is always a DBAPI2 connection, or raise an error
```
cpcloud, line 86: What is the precise procedure for determining whether an object is a DBAPI2 connection?

ibis/backends/base/sql/__init__.py
```diff
@@ -9,3 +18,200 @@ class BaseSQLBackend(BaseBackend):
...
+        Returns
+        -------
+        output : input type dependent
+          Table expressions: pandas.DataFrame
+          Array expressions: pandas.Series
+          Scalar expressions: Python scalar value
+        """
+        # TODO Reconsider having `kwargs` here. It's needed to support
+        # `external_tables` in clickhouse, but better to deprecate that
+        # feature than all this magic.
```
cpcloud, line 117: Can you make an issue for this so that it can be discussed?

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -43,30 +54,369 @@
...
+
 class BaseAlchemyBackend(BaseSQLBackend):
     """
     Base backend class for backends that compile to SQL with SQLAlchemy.
     """
 
     database_class = AlchemyDatabase
     table_class = AlchemyTable
+    compiler = AlchemyCompiler
+    has_attachment = False
```
cpcloud, line 87: What does this mean? Is this referring to the SQLite `ATTACH` statement?

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: I think this a carryover from the previous location. `staticmethod`s are useful in cases where an otherwise free function is specific to an operation in a class, but doesn't need the instance (`self`) or the type (`cls`) of the class to do its job. I personally think it's fine to leave as a `staticmethod`.

ibis/backends/base/sql/alchemy/client.py
```diff
@@ -1,370 +1,15 @@
...
-        else:
-            raise ValueError(
-                "No operation is being performed. Either the obj parameter "
-                "is not a pandas DataFrame or is not a ibis TableExpr."
-                f"The given obj is of type {type(obj).__name__} ."
-            )
+        self.backend.con = con
+        self.backend.meta = sa.MetaData(bind=con)
+        self.backend._inspector = sa.inspect(con)
+        self.backend._schemas: Dict[str, sch.Schema] = {}
```
cpcloud, line 15: We definitely need to remember to go back and make sure we don't have any unnecessary reference cycles once things are moved.

ibis/backends/clickhouse/__init__.py
```diff
@@ -114,3 +128,102 @@ def list_tables(self, like=None, database=None):
...
+
+        df = pd.DataFrame.from_dict(OrderedDict(zip(schema.names, data)))
+        return schema.apply_to(df)
+
+    def close(self):
+        """Close Clickhouse connection and drop any temporary objects"""
+        self.con.disconnect()
+
+    def _fully_qualified_name(self, name, database):
+        if bool(fully_qualified_re.search(name)):
```
cpcloud, line 182: This `bool` is unnecessary.

ibis/backends/clickhouse/__init__.py
```diff
@@ -114,3 +128,102 @@ def list_tables(self, like=None, database=None):
...
+    def set_options(self, options):
+        self.con.set_options(options)
+
+    def reset_options(self):
+        # Must nuke all cursors
+        raise NotImplementedError
+
+    def _ensure_temp_db_exists(self):
+        name = (options.clickhouse.temp_db,)
+        if name not in self.list_databases():
```
cpcloud, line 218: Is `list_databases` returning a `list[tuple[str]]`? That seems like it violates the exist contract. Is this just a carryover from the move?

ibis/backends/impala/tests/test_patched.py
```diff
@@ -8,6 +8,7 @@ def patch_execute(con):
 
 
 def test_invalidate_metadata(con, test_data_db):
+    con = con.backend
```
cpcloud, line 11: Can you instead return this from the fixture? Not a huge fan of rebinding variables of with different types of values.

ibis/backends/mysql/__init__.py
```diff
@@ -86,3 +91,47 @@ def connect(
...
+    @contextlib.contextmanager
+    def begin(self):
+        with super().begin() as bind:
+            previous_timezone = bind.execute(
+                'SELECT @@session.time_zone'
+            ).scalar()
+            try:
+                bind.execute("SET @@session.time_zone = 'UTC'")
+            except Exception as e:
+                warnings.warn(f"Couldn't set mysql timezone: {str(e)}")
```
cpcloud, line 104: I think this warning is pretty useless, but we can evaluate that again later.

ibis/backends/sqlite/client.py
```diff
@@ -1,90 +1,22 @@
...
-
     def __init__(self, backend, path=None, create=False):
-        super().__init__(sa.create_engine("sqlite://"))
         self.backend = backend
-        self.database_class = backend.database_class
-        self.table_class = backend.table_class
-        self.name = path
-        self.database_name = "base"
+        super().__init__(sa.create_engine("sqlite://"))
+        # self.backend.name = path
```
cpcloud, line 14: Can this be deleted?

ibis/expr/rules.py
```diff
@@ -291,9 +291,12 @@ def interval(arg, units=None):
 
 @validator
 def client(arg):
-    from ibis.backends.base import Client
+    from ibis.backends.base import BaseBackend, Client
 
-    return instance_of(Client, arg)
+    try:
+        return instance_of(Client, arg)
```
cpcloud, line 297: I think you can probably write this as:

```python
instance_of((Client, BaseBackend), arg)
```

ibis/backends/base/client.py
```diff
@@ -29,6 +29,127 @@ def exists_database(self, name):
...
+    def sql(self, *args, **kwargs):
+        return self.backend.sql(*args, **kwargs)
+
+    def set_database(self, *args, **kwargs):
+        return self.backend.set_database(*args, **kwargs)
+
+    def close(self, *args, **kwargs):
+        return self.backend.close(*args, **kwargs)
+
+    # impala
```
cpcloud, line 100: Do these methods need to be here? Many of them are highly specific to impala.

cpcloud (commented):
Looks mostly good, just a few comments.

ibis/backends/impala/tests/test_patched.py
datapythonista: Good point, implemented this way first to see if tests were passing, and never updated. Thanks.

ibis/backends/mysql/__init__.py
datapythonista: Also moved this from Client without changes, were this was implemented.

ibis/backends/sqlite/client.py
datapythonista: Looks like, thanks for the heads up.

datapythonista (commented):
Thanks for the feedback both. If that's fine, I'll leave all the feedback in code that was just moved here to a follow up, and just address the comments on the code implemented or modified here. The diff here is huge, and for any future problem and clearer git history, I think it's better to be able to assume that the big chunks of code here, were moved without changes. And having a separate small PR for the changes, will let revert and identify possible problems in a much easier way.

jreback:
ok if this is a straight moved then ok with merging

datapythonista:
> ok if this is a straight moved then ok with merging

Yes, all comments not addressed are from code just moved from one class to another. I'll be taking care of those comments in follow up PRs, since it's good feedback that otherwise will get lost.

I addressed every comment in a separate commit, in case it's easier to review, so you don't need to go through the whole diff again. Let me know if there is anything left.

jreback:
ok lgtm. prob should have a nice section of the release notes to show the migration path, but that should be a dedicated issue / PR.

jreback (approved):

jreback:
@cpcloud over to you.

cpcloud (approved):
LGTM.

jreback:
thanks @datapythonista

ibis/backends/base/sql/alchemy/__init__.py
datapythonista: I had a look, this is only used when inserting data in SQLite (`insert` and `load_data`). When `attach` in SQLite has been called, this add the parameter `DataFrame.to_sql(schema=current_database)`.

PR #3024, opened 2021-09-27
fix(ir): remove accidental print statement

kszucs:
At least I think it serves no purpose, I think it's a debugging leftover.


jreback (approved):
yep looks right

datapythonista:
There is a flake8 issue, since `e` is captured but not used now.

datapythonista (approved):
lgtm

PR #3030, opened 2021-09-29
Move methods from Client to Backend for pandas based backends

datapythonista:
xref #3017

Same as #3018, but for pandas backends (pandas, dask, csv, parquet and hdf5).

### DIFF
--- a/ibis/backends/base/file/__init__.py
+++ b/ibis/backends/base/file/__init__.py
@@ -1,3 +1,4 @@
+import abc
 import warnings
 from pathlib import Path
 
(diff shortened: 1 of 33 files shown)


ibis/backends/csv/__init__.py
```diff
@@ -56,12 +64,8 @@ def table(self, name, path=None, schema=None, **kwargs):
 
         return table
 
-    @property
-    def version(self):
-        return parse_version(pd.__version__)
 
-
-@pre_execute.register(ops.Selection, CSVClient)
+@pre_execute.register(ops.Selection, (Backend, CSVClient))
```
cpcloud, line 68: I guess this will go away after everything becomes a backend?

datapythonista: Yes, when client classes are deleted and fully replaced by backend classes (with some luck in the next PR after this one), all those will register only backend classes, not both.

datapythonista:
@jreback @cpcloud this should be ready now. As in the previous PR, this moves the methods from client classes to backend classes without changes more than the required (like `self.client` to `self` to find the new location of the methods, or to update the classes in the dispatchers).

In this case, I'm already removing the client classes here. I'll follow up removing the client classes from the SQL backends, which should be easy now, and from my side we're good to release.

ibis/backends/pandas/__init__.py
```diff
@@ -24,8 +26,13 @@ def connect(self, dictionary):
         -------
         Client
         """
-        self.client = self.client_class(backend=self, dictionary=dictionary)
-        return self.client
+        # register dispatchers
+        from . import execution  # noqa F401
```
jreback, line 30: can these imports be done at the top?

ibis/backends/pandas/__init__.py
```diff
@@ -70,16 +77,60 @@ def list_databases(self, like=None):
...
 class Backend(BasePandasBackend):
     name = 'pandas'
     database_class = PandasDatabase
     table_class = PandasTable
-    client_class = PandasClient
 
-    def execute(self, *args, **kwargs):
-        return execute(*args, **kwargs)
+    def execute(self, query, params=None, limit='default', **kwargs):
+        from .core import execute_and_reset
```
jreback, line 107: same

jreback (changes requested):

ibis/backends/base/file/__init__.py
```diff
@@ -175,6 +115,22 @@ def _list_databases_dirs(self, path=None):
                     tables.append(d.name)
         return tables
 
+    def _list_tables_files(self, path=None):
+        # tables are files in a dir
```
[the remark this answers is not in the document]
datapythonista: Yes, happy to work on typing stuff after the release. But I think we should first the broken mypy errors, since we're not validating types at the moment, and adding types that are not checked doesn't add that much value in my opinion.

ibis/backends/pandas/__init__.py
datapythonista: Unfortunately, the `ibis.backends.pandas.execution` module imports `ibis.backends.pandas` to be able to register the pandas `Backend` class in some dispatchers.  This created a lot of problems in the dispatching, and I'd be happy to try to improve that code later, since it's quite difficult to maintain. But for now this is the best solution I've found.

Same applies to the code in the other comment below.

ibis/backends/base/file/__init__.py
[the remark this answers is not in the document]
jreback: sure that's why i said followup

jreback:
ok seems fine to me, @cpcloud any issues here?

@datapythonista after this anything else for release (aside from maybe cleaning release notes)

jreback (approved):

datapythonista:
> @datapythonista after this anything else for release (aside from maybe cleaning release notes)

From my side just removing the `Client` classes from SQL backends. With some luck there are no problems with that, and I can get it ready today too, so we can release as soon as you want.

I wouldn't mind removing some backends from this repo, I think that would help with the release too, if we're going to release a package for each backend, which I think it's the plan. But we can leave it for after the release too.

jreback:
> I wouldn't mind removing some backends from this repo, I think that would help with the release too, if we're going to release a package for each backend, which I think it's the plan. But we can leave it for after the release too.

let's leave for after

datapythonista:
@cpcloud mind having a look at this please?

cpcloud:
I am on vacation this week, so I will defer to other reviewers until I am back on the 12th.

datapythonista:
> I am on vacation this week, so I will defer to other reviewers until I am back on the 12th.

Thanks for letting me know. I'll merge and we can follow up if you have any feedback later on.

Issue #2919, opened 2021-08-30
refactor: simplify relational operators


cpcloud:
Many logically distinct operations like filter, project, and aggregate are fused together into a single operation.

While this was useful initially for generating clean SQL, it has a number of downsides regarding maintenance and complexity of the library:

1. The current design has led to a very tight coupling between the logical representation of an expression tree, and its optimization.
2. It becomes difficult to maintain the behavior of these operators in isolation, because one must handle every aspect of a non-join relation when dealing with them in pretty much every context. There's no way to just think about a predicate.

To the end of simplifying the design and maintenance of ibis, I propose two related internal changes.

Proposal:

Separate `Selection` into distinct operators for each operation:

1. Project
2. Filter
3. OrderBy

Whittle down `Aggregation` into a class with *only* `metrics` and `by` as instance members, and use `where` to implement the `having` API, and `Filter` and `OrderBy` to take the place of `predicates` and `sort_keys` respectively.

jreback:
would for sure be +1 here

maybe could even do this w/o breaking the world (eg back compatible)

cpcloud:
Yeah, I think this can be done in a backwards compatible way. The generated SQL for certain backends will almost certainly change though.

icexelloss:
+1 to this. We internally depend on it currently being projection (isinstance checks) but not a huge deal to update to using the new classes.

lostmygithubaccount:
@cpcloud is this issue still valid?

cpcloud:
@lostmygithubaccount Yep, happening as part of #7580!

cpcloud:
Done in `the-epic-split`!

PR #3031, opened 2021-09-30
refactor: remove predicate fusion and pushdown

cpcloud:
This PR removes predicate pushdown and fusion in `ibis/expr/analysis.py`
in an effort to greatly simplify expression construction and increase
maintainability.

This is the first in a series of backwards-incompatible changes that
will culminate in the simplication of the core relational operators (#2919).

There are a number of benefits to this:

1. Fewer lines of code to maintain
1. Fewer lines of highly complex code to maintain
1. Much easier to debug broken queries
1. New contributors can more easily understand the library

We can revisit adding fusion back later, after designing a more robust and
modular API for optimizing at expression construction time.

There's one major breaking change:

1. Filter expressions that refer to a non-immediate child table are no longer
   valid and will raise an error (projections are almost all still valid,
   removing projection fusion will come in a follow-up).

   An example of this is:

```python
t = ibis.table([("a", "string"), ("b", "int64")])
t[t.a == "foo"][t.b == 1]
```

   previously this would fuse `t.a == "foo"` and `t.b == 1` into a conjunction
   and return `t[(t.a == "foo") & (t.b == 1)]`.

   The behavior is still available through other means, like the `&` operator, passing
   a list of predicates, and if you must chain, you need to use a lambda.

Non-breaking changes:

1. SQL queries contain more nesting. Queries that compose a lot of filters will
   contain more subqueries.

   I don't view this as problematic, as most database engines having fairly
   comprehensive subquery unnesting capabilities.

### DIFF
--- a/ibis/backends/base/sql/compiler/select_builder.py
+++ b/ibis/backends/base/sql/compiler/select_builder.py
@@ -44,9 +44,14 @@ class _AnyToExistsTransform:
             if isinstance(arg, ir.TableExpr):
                 self._visit_table(arg)
             elif isinstance(arg, ir.BooleanColumn):
-                for sub_expr in L.flatten_predicate(arg):
-                    self.predicates.append(sub_expr)
-                    self._visit(sub_expr)
+                op = arg.op()
+                if isinstance(op, ops.And):
+                    for pred in (op.left, op.right):
+                        self._visit(pred)
+                        self.predicates.append(pred)
+                else:
+                    self.predicates.append(arg)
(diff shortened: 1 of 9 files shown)


cpcloud:
This is still a WIP (hence the draft).

There will probably be more temporaries created.

I may ultimately close this PR in favor of a different approach that would allow us to use rewrite rules to keep most optimizations.

cpcloud:
Closing this out for now.

PR #3035, opened 2021-10-04
CLN: Removing pending Client classes

datapythonista:
Removing all the pending `Client` classes. In some cases I move the content of `client.py` to `__init__.py`, since it's small and not worth the extra file. But other than that, just removing the classes, updating the `connect` methods to not use them, and updating any coded that breaks because of it.


datapythonista:
@jreback this is ready. Let me know if there is anything, otherwise I'll start working on the release when this is merged.

jreback (approved):
ok this is fine

jreback:
ok 2.0 is a breaking change so this doesn't make it any worse. As soon as release, then need to fix the backends who are not in the repo and likley will completely break. e.g. bigquery (and i am sure lots of others), but this ship has sailed.

@datapythonista pls fix up the release notes and can release 2.0

datapythonista:
I've got OmniSci almost working, will finish updating it before releasing, to make sure nothing in Ibis is breaking it. Same for bigquery. @tswast what's the state of it? I think now that the refactoring is over should be easy to get it working. Do you need help?

What does need to be fixed in the release notes?

jreback:
> What does need to be fixed in the release notes?


oh just the release date and maybe give a quick edit, & expl of the Client  -> Backend change.

tswast:
> I've got OmniSci almost working, will finish updating it before releasing, to make sure nothing in Ibis is breaking it. Same for bigquery. @tswast what's the state of it? I think now that the refactoring is over should be easy to get it working. Do you need help?

The BigQuery client is still pretty busted with v2. Some help would be great. I was thinking of maybe making the `main` branch a `v1` branch since keeping the backend compatible across v1 and v2 seems like it's not worth the effort anymore.

datapythonista:
Cool. Yes, when working in https://github.com/ibis-project/ibis-bigquery/pull/93 I saw the compatibility stuff to support Ibis 1 and 2 in the some codebase, and I don't think it's feasible. I removed some of the imports.

PR #3037, opened 2021-10-06
ENH: Pass scope to context-adjustment functions

timothydijamco:
This PR passes `scope` to the functions that compute context adjustment: `compute_context_adjustment` and `adjust_context`.

### Justification
The existing `compute_context_adjustment` and `adjust_context` functions don't make any use of `scope`. This change is useful for making context adjustment more extensible downstream of Ibis.

The `scope` variable is mainly used to cache ops that have already been calculated, but since it can store anything, it's also useful for storing general information that is accessible/updatable during the execution process. Before this PR, `adjust_context` has a limited view—the only information it has access to is the op-in-question and the current timecontext. Giving context adjustment functions access to `scope` makes them more extensible/flexible (e.g. to implement context-adjustment logic that needs any sort of information outside of the op and current timecontext).

### Testing
New test in `test_timecontext.py` that creates a custom `adjust_context` function and asserts that `scope` is passed in.

### DIFF
--- a/ibis/backends/dask/core.py
+++ b/ibis/backends/dask/core.py
@@ -257,6 +257,7 @@ def execute_until_in_scope(
             num_args=len(computable_args),
             timecontext=timecontext,
             clients=clients,
+            scope=scope,
         )
     else:
         arg_timecontexts = [None] * len(computable_args)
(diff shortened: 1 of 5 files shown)


jreback (approved):
thanks @timothydijamco very nice

PR #3044, opened 2021-10-07
bug: fix dask dispatch

### LINKED ISSUE #3042: bug: isolated dask backend tests fail due to removed imports
For some reason lines 6 and 8 here: https://github.com/ibis-project/ibis/commit/a1262410310bb4d638a73e1cdfbe93c2b4089905#diff-96d84d9b6e9e84a2be7a046dc9853df1ca5fc6e894307339b02cd61e666c0149L6-L8

were removed.

This causes dasks tests to fail when they are run in isolation from other tests that (transitively) import from the pandas backend.

This is both a ci bug and a bug in the code, since we're not testing backends independently. Perhaps unsurprisingly I discovered the bug in #2937, which fixes the CI part of this problem.

cpcloud:
This PR adds a necessary import to use the dask backend whose import was broken
by #3030.

We should cut a 2.0.1 release fairly soon so that dask users aren't upgrading
to a broken release.

Fixes #3042.

### DIFF
--- a/ibis/backends/dask/__init__.py
+++ b/ibis/backends/dask/__init__.py
@@ -6,6 +6,9 @@ import pandas as pd
 import toolz
 from dask.base import DaskMethodsMixin
 
+# import the pandas execution module to register dispatched implementations of
+# execute_node that the dask backend will later override
+import ibis.backends.pandas.execution  # noqa: F401
 import ibis.common.exceptions as com
 import ibis.config
 import ibis.expr.schema as sch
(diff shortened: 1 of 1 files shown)


cpcloud:
I think we should just cut a new minor release in a couple weeks after shaking out some more bugs.

jreback:
> I think we should just cut a new ~minor~ patch release in a couple weeks after shaking out some more bugs.

oh a 2.1? e.g. no breaking changes?

cpcloud:
@jreback Not 2.1, but 2.0.1.

PR #3077, opened 2021-10-19
ci: generate conda lock files in CI

cpcloud:
This PR generates conda lock files in CI for each major platform (linux, osx,
and windows) for conda development environments.


cpcloud:
@jreback nope, this is just to support development with conda without having to do any manual labor to support it

jreback:
> We can. How does adding an install and run of the non-backend tests sound? This is similar to what we do with `setup.py` generation.

i guess it might be sufficient to have a job which builds with this just to make sure that these are correctly generated, but not blocking rn

PR #3097, opened 2021-10-29
Support fillna and dropna on TableExpr

emilyreff7:
### Proposed APIs

```
TableExpr.dropna(subset=None, how='any')
"""Drop  rows with null values from the table"""
   subset: Optional list of column names to consider
   how: {'any', 'all'} - whether to drop rows if any or all values are null

TableExpr.fillna(replacements)
"""Fill null values in the table"""
   replacements: scalar value (to fill all columns) or mapping of column names to replacement value
```

This PR adds support for the Pandas, Dask, and Pyspark backends only.
There is a follow-up issue to add support for SQL backends: #3110.

### Tests
Added new tests for both methods in test_generic.py

### DIFF
--- a/ibis/backends/pyspark/compiler.py
+++ b/ibis/backends/pyspark/compiler.py
@@ -1772,6 +1772,30 @@ def compile_not_null(t, expr, scope, timecontext, **kwargs):
     return ~F.isnull(col) & ~F.isnan(col)
 
 
+@compiles(ops.DropNa)
+def compile_dropna_table(t, expr, scope, timecontext, **kwargs):
+    op = expr.op()
+    table = t.translate(op.table, scope, timecontext)
+    how = op.how.op().value
+    if op.subset:
+        subset = [col.op().value for col in op.subset]
+        return table.dropna(how=how, subset=subset)
+    else:
+        return table.dropna(how=how)
+
+
+@compiles(ops.FillNa)
+def compile_fillna_table(t, expr, scope, timecontext, **kwargs):
+    op = expr.op()
+    table = t.translate(op.table, scope, timecontext)
+    value = op.value.op().value
+    if op.subset:
+        subset = [col.op().value for col in op.subset]
+        return table.fillna(value=value, subset=subset)
+    else:
+        return table.fillna(value)
+
+
 # ------------------------- User defined function ------------------------
 
 
--- a/ibis/backends/pyspark/compiler.py
+++ b/ibis/backends/pyspark/compiler.py
@@ -1777,11 +1777,8 @@ def compile_dropna_table(t, expr, scope, timecontext, **kwargs):
     op = expr.op()
     table = t.translate(op.table, scope, timecontext)
     how = op.how.op().value
-    if op.subset:
-        subset = [col.op().value for col in op.subset]
-        return table.dropna(how=how, subset=subset)
-    else:
-        return table.dropna(how=how)
+    subset = [col.op().value for col in op.subset] if op.subset else None
+    return table.dropna(how=how, subset=subset)
 
 
 @compiles(ops.FillNa)
@@ -1789,11 +1786,8 @@ def compile_fillna_table(t, expr, scope, timecontext, **kwargs):
     op = expr.op()
     table = t.translate(op.table, scope, timecontext)
     value = op.value.op().value
-    if op.subset:
-        subset = [col.op().value for col in op.subset]
-        return table.fillna(value=value, subset=subset)
-    else:
-        return table.fillna(value)
+    subset = [col.op().value for col in op.subset] if op.subset else None
+    return table.fillna(value=value, subset=subset)
 
 
 # ------------------------- User defined function ------------------------
--- a/ibis/backends/pyspark/compiler.py
+++ b/ibis/backends/pyspark/compiler.py
@@ -1776,18 +1776,20 @@ def compile_not_null(t, expr, scope, timecontext, **kwargs):
 def compile_dropna_table(t, expr, scope, timecontext, **kwargs):
     op = expr.op()
     table = t.translate(op.table, scope, timecontext)
-    how = op.how.op().value
-    subset = [col.op().value for col in op.subset] if op.subset else None
-    return table.dropna(how=how, subset=subset)
+    subset = [col.get_name() for col in op.subset] if op.subset else None
+    return table.dropna(how=op.how, subset=subset)
 
 
 @compiles(ops.FillNa)
 def compile_fillna_table(t, expr, scope, timecontext, **kwargs):
     op = expr.op()
     table = t.translate(op.table, scope, timecontext)
-    value = op.value.op().value
-    subset = [col.op().value for col in op.subset] if op.subset else None
-    return table.fillna(value=value, subset=subset)
+    replacements = (
+        op.replacements.op().value
+        if hasattr(op.replacements, 'op')
+        else op.replacements
+    )
+    return table.fillna(replacements)
 
 
 # ------------------------- User defined function ------------------------
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -294,7 +294,7 @@ def test_fillna_table(backend, alltypes, value, subset):
     values = {c: value for c in subset} if subset else value
     expected = table_pandas.fillna(values).reset_index(drop=True)
 
-    backend.assert_frame_equal(result, expected)
+    backend.assert_frame_equal(result, expected, check_dtype=False)
 
 
 @pytest.mark.parametrize(
@@ -321,4 +321,4 @@ def test_dropna_table(backend, alltypes, how, subset):
         drop=True
     )
 
-    backend.assert_frame_equal(result, expected)
+    backend.assert_frame_equal(result, expected, check_dtype=False)
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -5,6 +5,7 @@ import pandas as pd
 import pytest
 
 import ibis
+import ibis.common.exceptions as com
 from ibis import literal as L
 
 
@@ -258,41 +259,42 @@ def test_select_filter_mutate(backend, alltypes, df):
 
 def test_fillna_invalid(alltypes):
     with pytest.raises(
-        ValueError, match=r"Columns \['invalid_col'\] do not exist.*"
+        com.IbisTypeError, match=r"value \['invalid_col'\] is not a field in.*"
     ):
-        alltypes.fillna(0.0, subset=['invalid_col'])
+        alltypes.fillna({'invalid_col': 0.0})
 
 
 def test_dropna_invalid(alltypes):
     with pytest.raises(
-        ValueError, match=r"Columns \['invalid_col'\] do not exist.*"
+        com.IbisTypeError, match=r"value 'invalid_col' is not a field in.*"
     ):
         alltypes.dropna(subset=['invalid_col'])
 
-    with pytest.raises(ValueError, match=r".*either 'any' or 'all'.*"):
+    with pytest.raises(ValueError, match=r".*is not in.*"):
         alltypes.dropna(how='invalid')
 
 
 @pytest.mark.parametrize(
-    ('value', 'subset'),
+    'replacements',
     [
-        (0.0, None),
-        (0, []),
-        (0.0, ['na_col']),
-        (1, None),
-        (0.0, ['none_col']),
+        0.0,
+        0,
+        1,
+        ({'na_col': 0.0}),
+        ({'na_col': 1}),
+        ({'none_col': 0.0}),
+        ({'none_col': 1}),
     ],
 )
 @pytest.mark.only_on_backends(['pandas', 'dask', 'pyspark'])
-def test_fillna_table(backend, alltypes, value, subset):
+def test_fillna_table(backend, alltypes, replacements):
     table = alltypes.mutate(na_col=np.nan)
     table = table.mutate(none_col=None)
     table = table.mutate(none_col=table['none_col'].cast('float64'))
     table_pandas = table.execute()
 
-    result = table.fillna(value, subset).execute().reset_index(drop=True)
-    values = {c: value for c in subset} if subset else value
-    expected = table_pandas.fillna(values).reset_index(drop=True)
+    result = table.fillna(replacements).execute().reset_index(drop=True)
+    expected = table_pandas.fillna(replacements).reset_index(drop=True)
 
     backend.assert_frame_equal(result, expected, check_dtype=False)
 
@@ -315,7 +317,7 @@ def test_dropna_table(backend, alltypes, how, subset):
     table = table.mutate(none_col=table['none_col'].cast('float64'))
     table_pandas = table.execute()
 
-    result = table.dropna(how, subset).execute().reset_index(drop=True)
+    result = table.dropna(subset, how).execute().reset_index(drop=True)
     subset = subset if subset else table_pandas.columns
     expected = table_pandas.dropna(how=how, subset=subset).reset_index(
         drop=True
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -297,6 +297,10 @@ def test_fillna_table(backend, alltypes, replacements):
     result = table.fillna(replacements).execute().reset_index(drop=True)
     expected = table_pandas.fillna(replacements).reset_index(drop=True)
 
+    # check_dtype is False here because there are dtype diffs between
+    # Pyspark and Pandas on Java 8 - filling the 'none_col' with an int
+    # results in float in Pyspark, and int in Pandas. This diff does
+    # not exist in Java 11.
     backend.assert_frame_equal(result, expected, check_dtype=False)
 
 
@@ -324,4 +328,8 @@ def test_dropna_table(backend, alltypes, how, subset):
         drop=True
     )
 
+    # check_dtype is False here because there are dtype diffs between
+    # Pyspark and Pandas on Java 8 - the 'bool_col' of an empty DataFrame
+    # is type object in Pyspark, and type bool in Pandas. This diff does
+    # not exist in Java 11.
     backend.assert_frame_equal(result, expected, check_dtype=False)
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -6,6 +6,7 @@ import pytest
 
 import ibis
 import ibis.common.exceptions as com
+import ibis.util as util
 from ibis import literal as L
 
 
@@ -307,7 +308,7 @@ def test_fillna_table(backend, alltypes, replacements):
         ('any', ['int_col', 'na_col']),
         ('all', None),
         ('all', ['int_col', 'na_col']),
-        ('all', ['none_col']),
+        ('all', 'none_col'),
     ],
 )
 @pytest.mark.only_on_backends(['pandas', 'dask', 'pyspark'])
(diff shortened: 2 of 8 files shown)


ibis/expr/operations/relations.py
```diff
@@ -785,3 +785,29 @@ class NotExistsSubquery(Node):
...
         return ir.ExistsExpr
+
+
+@public
+class FillNa(TableNode, sch.HasSchema):
+    """Fill null values in the table."""
+
+    table = Arg(rlz.table)
+    value = Arg(rlz.any)
+    subset = Arg(rlz.list_of(rlz.string), default=[])
```
cpcloud, line 796: This should probably be

```python
Arg(rlz.list_of(rlz.column_from("table")))
```

IIUC. It looks like you want to restrict `subset` to be a (possibly empty) list of column names from `table`. The way you have it now will allow any value that can be converted to a `StringValue`.

ibis/expr/operations/relations.py
```diff
@@ -785,3 +785,29 @@ class NotExistsSubquery(Node):
...
+        return self.table.schema()
+
+
+@public
+class DropNa(TableNode, sch.HasSchema):
+    """Drop null values in the table."""
+
+    table = Arg(rlz.table)
+    how = Arg(rlz.string)
+    subset = Arg(rlz.list_of(rlz.string), default=[])
```
cpcloud, line 809: Similar to the above this should probably be further restricted to just column names (or column expressions with a `TableColumn` op) by using `rlz.column_from("table")`.

`rlz.column_from` is a recent addition to the rule sets.

ibis/expr/operations/relations.py
```diff
@@ -785,3 +785,29 @@ class NotExistsSubquery(Node):
...
     def output_type(self):
         return ir.ExistsExpr
+
+
+@public
+class FillNa(TableNode, sch.HasSchema):
+    """Fill null values in the table."""
+
+    table = Arg(rlz.table)
+    value = Arg(rlz.any)
```
cpcloud, line 795: What happens with a `value` that doesn't match the type of one or more elements of `subset`?

cpcloud: Perhaps this should accept a dictionary mapping column names to value to use for replacement?

ibis/expr/operations/relations.py
cpcloud: If you end up accepting a dictionary here, you won't be able to use this rule. Instead you'll have to do something like:

```python
Arg(rlz.instance_of(collections.abc.Mapping))
```

and we can add a more thorough rule later. I'd suggest doing validation in the API functions for now, so that later when we add a proper rule, we are backwards compatible.

ibis/util.py
```diff
@@ -84,6 +84,8 @@ def promote_list(val: Union[V, List[V]]) -> List[V]:
     -------
     list
     """
+    if val is None:
+        val = []
```
cpcloud, line 88: I don't think we should automatically turn `None` into an empty list, since it gives `None` special meaning that it is equivalent to an empty sequence. On a practical note, IIRC there are a few places in the codebase where code assumes `None` becomes `[None]`.

ibis/expr/api.py
```diff
@@ -4032,6 +4032,53 @@ def _table_count(self):
...
+    table : TableExpr
+      New table expression
+    """
+    if how not in {'any', 'all'}:
+        raise ValueError("Please specify how as either 'any' or 'all'.")
+    subset = util.promote_list(subset)
+    invalid = set(subset) - set(self.schema().names)
+    if invalid:
+        raise ValueError(f'Columns {list(invalid)} do not exist in the table.')
+    return ops.DropNa(self, how, subset).to_expr()
```
cpcloud, line 4058: We should implement this for at least one SQL backend in a follow up. It can likely be done as a rewrite along the lines of:

```python
from operator import and_, or_
from functools import reduce

t[reduce(or_ if how == 'any' else and_, subset)]
```

cpcloud (changes requested):
Looks mostly good, just a few points around validation of arguments and the `fillna` API.

ibis/expr/operations/relations.py
cpcloud: Also, `how` can be further restricted by using `rlz.isin{('any', 'all'})`.

ibis/expr/api.py
```diff
@@ -4032,6 +4032,53 @@ def _table_count(self):
     return ops.Count(self, None).to_expr().name('count')
 
 
+def _table_dropna(self, how='any', subset=None):
+    """
+    Remove rows with null values from the table.
+
+    Parameters
+    ----------
+    how : determine whether a row is removed if there is at least one null
```
cpcloud, line 4041: Can you move the description of the parameter below its name, like this:

```
Parameters
----------
how
  determine whether a row is removed if there is at least one null
```

only types should come after the colon.

ibis/expr/api.py
```diff
@@ -4032,6 +4032,53 @@ def _table_count(self):
...
+
+
+def _table_fillna(self, value, subset=None):
+    """
+    Fill null values in the table.
+
+    Parameters
+    ----------
+    value : value with which to fill the nulls
+    subset : Optional list of columns to fill. Defaults to all columns.
```
cpcloud, line 4068: Similar to above can you move the parameter description below its name?

cpcloud:
If you rebase you should pick up changes to put your PR in a passing state.

ibis/backends/pyspark/compiler.py
```diff
@@ -1772,6 +1772,24 @@ def compile_not_null(t, expr, scope, timecontext, **kwargs):
     return ~F.isnull(col) & ~F.isnan(col)
 
 
+@compiles(ops.DropNa)
+def compile_dropna_table(t, expr, scope, timecontext, **kwargs):
+    op = expr.op()
+    table = t.translate(op.table, scope, timecontext)
+    how = op.how.op().value
+    subset = [col.op().value for col in op.subset] if op.subset else None
+    return table.dropna(how=how, subset=subset)
```
icexelloss, line 1781: We need to be careful here - Spark treats Na and None differently so this might be a behavior difference between Spark/Pandas.

@emilyreff7 Do we have test to validate Na and None values?

ibis/expr/operations/relations.py
icexelloss: > rlz.column_from is a recent addition to the rule sets.
Sweet

ibis/expr/operations/relations.py
icexelloss: I think this is convenient,  but also a bit weird because now "subset" is leaked into "value" (they are redudant) so not sure

```
empty_values = {
    dt.int32: 0,
    dt.float64: 0.0,
    dt.string: ""
}

to_fill = {
    table[col].name: empty_values[table[col].type()] for col in table.columns
}
dt.fillna(value=to_fill)
```

icexelloss: Does seem like pandas allows this? Seems reasonable to me then
```
df.fillna({'Name':'.', 'City':'.'}, inplace=True)
```

cpcloud: Right, we don't need to support the full pandas API, but if there's a stricter subset (it sounds like there is) then we can definitely support that.

Definitely don't think we should map value type to value, and instead remove the `subset` argument and accept a mapping like @icexelloss is showing here. I'll throw out calling the argument `replacements` but no strong feeling there.

icexelloss (changes requested):
High level looks good. Left some comments

ibis/expr/operations/relations.py
emilyreff7: So in this case the proposed usages would be:

Case 1: I want to fill all empty values with 0
```
table.fillna(0)
```

Case 2: I want to fill only a subset of columns with 0:
```
table.fillna({'A': 0, 'B': 0})
```

Case 3: I want to use different replacement values for different columns
```
table.fillna({'A': 1, 'B': 0})
```
I agree this is more concise than accepting both `value` and `subset`. And that `replacements` sounds better than `values`.

emilyreff7: > What happens with a `value` that doesn't match the type of one or more elements of `subset`?

I believe Pyspark will ignore in this case, whereas Pandas will actually fill (e.g. df.fillna({'float_col': 'string value'}) will do the replacement and promote the type of 'float_col' from float to object). So we may want to be more opinionated about the types matching like Pyspark is.

cpcloud: I suppose there are a couple recipes that could be useful for folks coming from pandas as well:

```
# replace all na values in all columns with the same value
t.fillna({c: "." for c in t.columns})

# replace all na values in all columns of a given type with the same value
t.fillna({c: 0.0 for c, t in t.schema().items() if t == dt.float64})
```

ibis/expr/api.py
```diff
@@ -4032,6 +4032,53 @@ def _table_count(self):
     return ops.Count(self, None).to_expr().name('count')
 
 
+def _table_dropna(self, how='any', subset=None):
```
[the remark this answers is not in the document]
jreback: yep, subset should be first. can you type it (e.g. Union[str, List[str]])

ibis/expr/api.py
jreback: +1 on Examples

jreback (changes requested):

ibis/expr/operations/relations.py
emilyreff7: So for the first example, I think we are better off accepting either a scalar or a dict in the API, which would prevent users from having to type a dict comprehension in the case where they want to fill all columns with the same value.

For the second, I wonder if we want to be strict about something like 
```
t.fillna({'float_col': '.'})
```
giving the same behavior between Pandas/Dask and Pyspark. If so, we'd take care of that second example internally, i.e. in Pandas/Dask execution rules we'd compare the dtypes and only consider the columns with matching type.

emilyreff7: Actually I think Pyspark may only be specifically ignoring (str, numeric) mismatches, while it will cast ints and bools to floats and fillna with those values. It might be messy trying to converge this behavior internally

ibis/expr/operations/relations.py
emilyreff7: Updated to use the `Arg(rlz.instance_of(collections.abc.Mapping))` validator + included manual validation in the API function that the columns exist in the table.

ibis/expr/operations/relations.py
emilyreff7: Updated both validators

ibis/util.py
emilyreff7: That makes sense - I was curious if there were any usages that expected `[None]` but at least no tests seemed to fail as a result...

ibis/expr/api.py
emilyreff7: Fixed the parameters + added examples

ibis/expr/api.py
emilyreff7: Fixed

ibis/backends/pyspark/compiler.py
emilyreff7: My tests do include both na_col and none_col examples

ibis/expr/api.py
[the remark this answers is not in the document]
emilyreff7: Fixed

ibis/expr/api.py
```diff
@@ -4032,6 +4032,53 @@ def _table_count(self):
...
+    if how not in {'any', 'all'}:
+        raise ValueError("Please specify how as either 'any' or 'all'.")
+    subset = util.promote_list(subset)
+    invalid = set(subset) - set(self.schema().names)
+    if invalid:
+        raise ValueError(f'Columns {list(invalid)} do not exist in the table.')
+    return ops.DropNa(self, how, subset).to_expr()
+
+
+def _table_fillna(self, value, subset=None):
```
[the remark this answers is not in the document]
emilyreff7: A subset of it at least

ibis/backends/pyspark/compiler.py
```diff
@@ -1772,6 +1772,26 @@ def compile_not_null(t, expr, scope, timecontext, **kwargs):
     return ~F.isnull(col) & ~F.isnan(col)
 
 
+@compiles(ops.DropNa)
+def compile_dropna_table(t, expr, scope, timecontext, **kwargs):
+    op = expr.op()
+    table = t.translate(op.table, scope, timecontext)
+    subset = [col.get_name() for col in op.subset] if op.subset else None
+    return table.dropna(how=op.how, subset=subset)
```
icexelloss, line 1780: This documentation here says "omitting rows with null values", but it works with "Na" values too?

https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.sql.DataFrame.dropna.html

icexelloss: Well it seems that `pyspark.sql.DataFrame.dropna` does handle both `Null` and `Na`. 

`pyspark.sql.functions.isnan` only handles `Na`

`pyspark.sql.functions.isnull` only handles `Null`

Kind of confusing behavior but I think what you are doing here is correct

icexelloss (approved):
LGTM. +1

ibis/expr/operations/relations.py
```diff
@@ -785,3 +786,36 @@ class NotExistsSubquery(Node):
...
+@public
+class FillNa(TableNode, sch.HasSchema):
+    """Fill null values in the table."""
+
+    table = Arg(rlz.table)
+    replacements = Arg(
+        rlz.one_of(
+            (
+                rlz.numeric,
+                rlz.string,
```
cpcloud, line 800: Couple of questions :) I think the non-mapping behavior is pretty fraught.

* What is the behavior of `rlz.numeric`/`rlz.string` where one or more columns do not match the type of the value provided?
* Similarly, what if the replacement value cannot be safely promoted to one or more column types?
* Does the operation fail in those cases? If so, does it fail at ibis-compile-time or is it backend specific?
* Is the behavior when `replacements` is a column expression equivalent to `coalesce`?

ibis/backends/tests/test_generic.py
```diff
@@ -254,3 +255,72 @@ def test_select_filter_mutate(backend, alltypes, df):
...
+    table = table.mutate(none_col=table['none_col'].cast('float64'))
+    table_pandas = table.execute()
+
+    result = table.dropna(subset, how).execute().reset_index(drop=True)
+    subset = subset if subset else table_pandas.columns
+    expected = table_pandas.dropna(how=how, subset=subset).reset_index(
+        drop=True
+    )
+
+    backend.assert_frame_equal(result, expected, check_dtype=False)
```
jreback, line 326: why do you have `check_type=False`? this should match

ibis/backends/tests/test_generic.py
```diff
@@ -254,3 +255,72 @@ def test_select_filter_mutate(backend, alltypes, df):
...
+
+
+def test_fillna_invalid(alltypes):
+    with pytest.raises(
+        com.IbisTypeError, match=r"value \['invalid_col'\] is not a field in.*"
+    ):
+        alltypes.fillna({'invalid_col': 0.0})
+
+
+def test_dropna_invalid(alltypes):
```
jreback, line 267: can you add a test for subset is not a list

ibis/expr/api.py
```diff
@@ -4032,6 +4032,72 @@ def _table_count(self):
...
+    -------
+    table : TableExpr
+      New table expression
+    """
+    if isinstance(replacements, collections.abc.Mapping):
+        columns = replacements.keys()
+        table_columns = self.schema().names
+        invalid = set(columns) - set(table_columns)
+        if invalid:
+            raise com.IbisTypeError(
```
jreback, line 4095: test hits this?

jreback (changes requested):

ibis/expr/api.py
emilyreff7: Yep I have a test for this

ibis/backends/tests/test_generic.py
emilyreff7: So when the CI ran for this, the pyspark tests with Java 8 had diffs in the column type when doing fillna(int_value) on a null column (int in pandas vs. float in Pyspark), and also differences in the column type of a bool_col of an empty DataFrame (bool vs. object): https://github.com/ibis-project/ibis/runs/4051801741?check_suite_focus=true
The pyspark tests on Java 11 passed fine though.

ibis/expr/operations/relations.py
emilyreff7: The behavior when the user passes a value whose type does not match one or more of the columns is backend-specific - for example, `df.fillna('string_value')` - Pandas will successfully fill float columns with 'string_value', while Pyspark will ignore. Or, `df.fillna(False)` - Pandas will fill float columns with False, while Pyspark will fill with 0.0. I'm not aware of a case where `type(value) != df[column].dtype` will actually fail; the individual backend will either box the replacement value into the column type, promote it, or ignore it.

At the same time, `t.fillna(val)` is just syntactic sugar for `t.fillna({c: val for c in t.columns})`, so unless we commit to doing some internal ibis validation at compile time, I am not sure why the syntactic sugar version of the API is problematic compared to `t.fillna({c: val for c in t.columns})`.

Also for the final point, I don't think we need to support replacements as a column expression if we already have coalesce.

cpcloud: > At the same time, t.fillna(val) is just syntactic sugar for t.fillna({c: val for c in t.columns}), so unless we commit to doing some internal ibis validation at compile time, I am not sure why the syntactic sugar version of the API is problematic compared to t.fillna({c: val for c in t.columns}).

Good point. I think I am just bikeshedding on whether we should do some kind of type checking here, but this isn't blocking the PR.

cpcloud (approved):
Great work. Thanks for doing this.

ibis/expr/operations/relations.py
emilyreff7: Yeah it's definitely something to consider, if the goal is to achieve as close to 100% consistent behavior across all backends then we'd be constrained by the most restrictive backend behavior which seems to be Pyspark at this point.

ibis/expr/api.py
emilyreff7: Created an issue to track https://github.com/ibis-project/ibis/issues/3110

ibis/backends/tests/test_generic.py
jreback: kk i have to loook at the doc again, but we should highlite this in the Notes section (e.g. potential lack of type stability) on various backends (also add a comment here)

ibis/expr/api.py
```diff
@@ -4032,6 +4032,72 @@ def _table_count(self):
...
+      Optional, columns names to consider. Defaults to all columns.
+    how : string
+      Determine whether a row is removed if there is at least one null
+      value in the row ('any'), or if all row values are null ('all').
+      Options are 'any' or 'all'. Default is 'any'.
+
+    Examples
+    --------
+    >>> import ibis
+    >>> t = ibis.table([('a', 'int64'), ('b', 'string')])
```
jreback, line 4051: i prefer these comments to be on a separate line *before*, but whatever is consistent other doc strings

ibis/expr/api.py
```diff
@@ -4032,6 +4032,72 @@ def _table_count(self):
...
+    >>> t = t.fillna(0.0)  # Replace nulls in all columns with 0.0
+    >>> t.fillna({c: 0.0 for c, t in t.schema().items() if t == dt.float64})  # Replace all na values in all columns of a given type with the same value  # noqa: E501
+
+    Returns
+    -------
+    table : TableExpr
```
jreback, line 4087: add comment about type stability in Notes

jreback (changes requested):
also pls add a release note

ibis/expr/api.py
emilyreff7: This was done for consistency with the other examples in this file

ibis/expr/api.py
emilyreff7: Added

jreback:
thanks @emilyreff7 very nice!

PR #3099, opened 2021-10-31
chore(poetry.lock): regenerate

cpcloud:
Looks like we're hitting https://github.com/python-poetry/poetry/issues/4210 in CI.

The mentioned workaround is to purge the cache (or disable the new installer).

This PR should cause a new cache to be created.


cpcloud:
Merging on green in the interest of getting CI back in working order

PR #3104, opened 2021-11-02
FIX: Pandas and Dask backends fail when adjusted context is not a superset of original context

timothydijamco:
## Problem

Before this PR, any `adjust_context` function (i.e., a function that defines logic for adjusting the timecontext for a particular Node) that results in **an adjusted context is not a superset of the original context** would cause an error in the Pandas and Dask backends.

--

Example:
- Begin adjusted from `2020-02-01` to `2020-01-29`
- End adjusted from `2020-02-07` to `2020-02-03`

The adjusted context is _not_ a superset of the original context.

--

This is not causing any user-facing issues currently, because all of our existing `adjust_context` functions produce an adjusted timecontext that is a superset of the original timecontext, which doesn't trigger this error. However, this bug is not ideal because context adjustment functions should not be restricted to only being able to _expand_ the original timecontext.

### Root cause

In the Pandas backend's and Dask backend's core execution code, there is a problem in the logic that fetches materialized data (e.g. materialized left and right tables when executing `AsOfJoinNode`) from `scope` to feed into `execute_node`. The bug is that it was querying `scope` for the materialized data **using the old, non-adjusted timecontexts as part of the key, instead of the new, adjusted timecontexts**. Then, it would fail to fetch the materialized data, and try to execute the node with `None`s instead.

## Fix

This PR patches the Pandas and Dask backend's core execution code by making it query `scope` using adjusted timecontexts.

## Testing

New test that creates a contrived `adjust_context` function for a mock `AsOfJoinNode`, which creates a completely-shifted adjusted context. Created this test for both backends:
- `ibis/backends/pandas/tests/execution/test_timecontext.py`
- `ibis/backends/dask/tests/execution/test_timecontext.py`

### DIFF
--- a/ibis/backends/dask/tests/execution/test_timecontext.py
+++ b/ibis/backends/dask/tests/execution/test_timecontext.py
@@ -8,6 +8,7 @@ from pandas import Timedelta, Timestamp
 import ibis
 import ibis.common.exceptions as com
 import ibis.expr.operations as ops
+from ibis.backends.pandas.execution import execute
 from ibis.expr.scope import Scope
 from ibis.expr.timecontext import (
     TimeContextRelation,
@@ -269,6 +270,8 @@ def test_adjust_context_complete_shift(
     This results in an adjusted context that is NOT a subset of the
     original context. This is unlike an `adjust_context` function
     that only expands the context.
+
+    See #3104
     """
 
     # Create a contrived `adjust_context` function for
@@ -281,8 +284,6 @@ def test_adjust_context_complete_shift(
         scope: Optional[Scope] = None,
     ) -> TimeContext:
         """Shifts both the begin and end in the same direction."""
-        from ibis.backends.pandas.execution import execute
-
         begin, end = timecontext
         timedelta = execute(op.tolerance)
         return (begin - timedelta, end - timedelta)
--- a/ibis/backends/dask/tests/execution/test_timecontext.py
+++ b/ibis/backends/dask/tests/execution/test_timecontext.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 import dask.dataframe as dd
 import pytest
 from dask.dataframe.utils import tm
@@ -5,7 +7,18 @@ from pandas import Timedelta, Timestamp
 
 import ibis
 import ibis.common.exceptions as com
-from ibis.expr.timecontext import TimeContextRelation, compare_timecontext
+import ibis.expr.operations as ops
+from ibis.expr.scope import Scope
+from ibis.expr.timecontext import (
+    TimeContextRelation,
+    adjust_context,
+    compare_timecontext,
+)
+from ibis.expr.types import TimeContext
+
+
+class CustomAsOfJoin(ops.AsOfJoin):
+    pass
 
 
 def test_execute_with_timecontext(time_table):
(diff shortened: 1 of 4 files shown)


ibis/backends/dask/tests/execution/test_timecontext.py
```diff
@@ -218,3 +231,88 @@ def test_context_adjustment_window_groupby_id(time_table, time_df3):
     # result should adjust time context accordingly
     result = expr.execute(timecontext=context)
     tm.assert_series_equal(result, expected)
+
+
+def test_adjust_context_scope(time_keyed_left, time_keyed_right):
```
timothydijamco, line 236: To avoid confusion, this new Dask backend test is not very related to this PR. This should have been mirrored from the Pandas tests in #3037, and I am just fixing this now, since this PR already needs to bring over the `CustomAsOfJoin` text fixtures from the Pandas tests anyways.

ibis/backends/dask/tests/execution/test_timecontext.py
```diff
@@ -218,3 +231,88 @@ def test_context_adjustment_window_groupby_id(time_table, time_df3):
...
+        predicates='time',
+        by='key',
+        tolerance=ibis.interval(days=4),
+    ).to_expr()
+    expr = expr[time_keyed_left, time_keyed_right.other_value]
+    context = (Timestamp('20170105'), Timestamp('20170111'))
+    expr.execute(timecontext=context)
+
+
+def test_adjust_context_complete_shift(
```
timothydijamco, line 261: This is the new Dask test that is specifically for this bug fix.

timothydijamco:
(cc @icexelloss @jreback)

ibis/backends/pandas/tests/execution/test_timecontext.py
```diff
@@ -289,13 +289,73 @@ def adjust_context_custom_asof_join(
...
         by='key',
-        tolerance=4 * ibis.interval(days=1),
+        tolerance=ibis.interval(days=4),
     ).to_expr()
     expr = expr[time_keyed_left, time_keyed_right.other_value]
     context = (pd.Timestamp('20170105'), pd.Timestamp('20170111'))
     expr.execute(timecontext=context)
 
 
+def test_adjust_context_complete_shift(
```
cpcloud, line 299: I guess at some point we should probably try to deduplicate this testing code where possible.

cpcloud (approved):
Great find, and excellent job on the fix.

ibis/backends/dask/tests/execution/test_timecontext.py
jreback: can you put a refernce to this PR in a comment

ibis/backends/dask/tests/execution/test_timecontext.py
```diff
@@ -218,3 +231,88 @@ def test_context_adjustment_window_groupby_id(time_table, time_df3):
...
+    # CustomAsOfJoin to mock this.
+
+    @adjust_context.register(CustomAsOfJoin)
+    def adjust_context_custom_asof_join(
+        op: ops.AsOfJoin,
+        timecontext: TimeContext,
+        scope: Optional[Scope] = None,
+    ) -> TimeContext:
+        """Shifts both the begin and end in the same direction."""
+        from ibis.backends.pandas.execution import execute
```
jreback, line 284: can you move this import to the top? or does this cause issues?

ibis/backends/pandas/tests/execution/test_timecontext.py
```diff
@@ -289,13 +289,73 @@ def adjust_context_custom_asof_join(
...
+    # CustomAsOfJoin to mock this.
+
+    @adjust_context.register(CustomAsOfJoin)
+    def adjust_context_custom_asof_join(
+        op: ops.AsOfJoin,
+        timecontext: TimeContext,
+        scope: Optional[Scope] = None,
+    ) -> TimeContext:
+        """Shifts both the begin and end in the same direction."""
+        from ibis.backends.pandas.execution import execute
```
jreback, line 322: same comment as above

jreback (commented):
minor comments, ping on green.

ibis/backends/pandas/tests/execution/test_timecontext.py
timothydijamco: agreed! For some tests it will be a little tricky due to small differences like using `dd` instead of `pd`, and calling `.compute()` in the Dask tests. It would be nice to go through and minimize this duplication eventually though

ibis/backends/dask/tests/execution/test_timecontext.py
timothydijamco: ah looks like this is fine

ibis/backends/dask/tests/execution/test_timecontext.py
icexelloss: I am still confused. Why are most of the tests in this PR under dask tests? How are we testing pandas backend?

icexelloss: Oh nvm saw ot

ibis/backends/pandas/tests/execution/test_timecontext.py
icexelloss: +1 to deduplicated test. We should test Spark as well.

icexelloss: @timothydijamco I don't understand the argument - why can this not be a shared tests?

icexelloss (commented):
Reviewed one around. 

High level the changes look good but have some questions about the tests

jreback (approved):

ibis/backends/pandas/tests/execution/test_timecontext.py
timothydijamco: oh actually you're right, this can be shared. The usage of Dask-specific things is only to compute the expected result, but for a shared test we can compute the expected result using Pandas for all backends (which is what we normally do in our shared tests I believe)

Let me refactor

timothydijamco: Unfortunately none of our source datasets that can be used in our shared tests (i.e. the data defined in https://github.com/ibis-project/testing-data) have a `time` column, which is required when testing context adjustment code (the source table itself must have a `time` column—it cannot be added in a test to bypass this)

Since there is no shared dataset with a `time` column, each backend (Pandas, Dask, and PySpark) is currently defining test datasets using their respective backend DataFrame APIs:
- [Pandas](https://github.com/ibis-project/ibis/blob/master/ibis/backends/pandas/tests/execution/conftest.py#L131-L187)
- [Dask](https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/tests/execution/conftest.py#L126-L187)
- [PySpark](https://github.com/ibis-project/ibis/blob/master/ibis/backends/pyspark/tests/conftest.py#L235-L249)

If we add test data with a `time` column to https://github.com/ibis-project/testing-data, then we would be able to consolidate many of the separate tests that across Pandas/PySpark/Dask that use `time` column into shared tests

Would everyone be OK with this being a separate issue / PR to narrow the scope of this PR?

jreback: ok with doing this as a separate issue / followup.

icexelloss: Yep I am OK with that

icexelloss (approved):
LGTM. +1

jreback:
kk pls create an issue for followup on the testing.

ibis/backends/pandas/tests/execution/test_timecontext.py
timothydijamco: ok, I created a follow-up: https://github.com/ibis-project/ibis/issues/3109

PR #3105, opened 2021-11-03
refactor(signature): Inherit slots instead of redefining them

kszucs:


ibis/tests/expr/test_signature.py
```diff
@@ -198,32 +217,77 @@ class NoHooves(Farm):
...
+
+
 def test_slots_are_inherited_and_overridable():
     class Op(Annotable):
         __slots__ = ('_cache',)  # first definition
         arg = Argument(lambda x: x)
 
     class StringOp(Op):
-        arg = Argument(str)  # inherit
+        arg = Argument(str)  # new overridden slot
```
kszucs, line 236: We do redefine the `arg` slot here, although we could use the parent slot too, but then we need to aggregate the parent slots to check whether the slot already exists or not.

ibis/expr/signature.py
```diff
@@ -112,37 +106,30 @@ def Argument(validator, default=EMPTY):
 
 class AnnotableMeta(type):
     def __new__(metacls, clsname, bases, dct):
-        slots, params = [], {}
-
+        params = {}
         for parent in bases:
-            # inherit parent slots
-            if hasattr(parent, '__slots__'):
-                slots += parent.__slots__
```
kszucs, line 120: This should spare some memory.

ibis/expr/operations/core.py
```diff
@@ -95,37 +93,6 @@ def _pp(x):
...
-
-        Returns
-        -------
-        Dict[str, Any]
-            A dictionary storing the objects attributes.
-        """
-        excluded_slots = {'_expr_cached', '_hash'}
-        return {
-            slot: getattr(self, slot)
-            for slot in self.__slots__
```
kszucs, line 115: `__slots__` doesn't store the inherited keys anymore, so we need to use `self.__signature__.keys()`. Since this is a useful feature for `Annotable` I moved this implementation there.

ibis/expr/signature.py
```diff
@@ -112,37 +106,30 @@ def Argument(validator, default=EMPTY):
...
-        # if slots or signature are defined no inheritance happens
-        slots = attribs.get('__slots__', tuple(slots))
-        slots += tuple(params.keys())
-
         # mandatory fields without default values must preceed the optional
         # ones in the function signature, the partial ordering will be kept
         params = sorted(
             params.values(), key=lambda p: p.default is EMPTY, reverse=True
         )
-        signature = attribs.get('__signature__', inspect.Signature(params))
```
kszucs, line 142: Haven not used this feature anywhere.

ibis/expr/operations/core.py
```diff
@@ -95,37 +93,6 @@ def _pp(x):
...
         return '{}({})'.format(opname, ', '.join(pprint_args))
 
-    def __getstate__(self) -> Dict[str, AnyT]:
-        """The attributes _expr_cached and _hash are
-        used as caches; they can be excluded from
-        serialization without affecting correctness.
-
-        Excluding _expr_cached and _hash from serialization
-        will allow the serialized bytes to be the same for
-        equivalent Node objets.
```
cpcloud, line 105: Do these comments still apply?

ibis/expr/signature.py
cpcloud: This is fine to remove. We could potentially use it in a world where we use `matchpy` to implement algebraic optimizations, but that's a ways out.

ibis/tests/expr/test_signature.py
cpcloud: I don't know if this is reasonably possible. What would happen in the case of `C(A, B)` where `A(Annotable)` and `B(Annotable)` and `A` and `B` have a slot with the same name?

cpcloud (approved):
LGTM.

ibis/tests/expr/test_signature.py
kszucs: That's explicitly disallowed by python, see https://github.com/ibis-project/ibis/pull/3105/files#diff-21051dad17424e3006baa46b3c0555827349335eb54ea9d6eb290cda9992c13fR276-R280

ibis/expr/operations/core.py
kszucs: We only serialize the fields part of the signature, nothing else.

cpcloud: I think this is probably a regression then, because those fields are explicitly being serialized.

cpcloud (changes requested):
Should get a review from @jreback @icexelloss on the pickling changes.

ibis/expr/operations/core.py
kszucs: The behavior hasn't changed, `_expr_cached` and `_hash` are not part of the signature https://github.com/ibis-project/ibis/pull/3105/files#diff-09f7d95a14c73bb445f8c811277d0b3be65436979662a505a2275e6059ae5237R160

cpcloud (approved):

PR #3111, opened 2021-11-03
feat(ci): More readable workflow job titles 

kszucs:


cpcloud:
Seems fine, though the main reason maybe not to rename is now maybe it's more difficult to find the job in the YAML corresponding to the pretty name in the GHA UI? There's probably a way to do that though, and these job names do look much better.

.github/workflows/ibis-main.yml
```diff
@@ -0,0 +1,206 @@
...
+  test_no_backends:
+    name: Test ${{ matrix.os }} python-${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - windows-latest
+        python-version:
```
cpcloud, line 22: This should include 3.8, right? There is a job running windows-py38, but it's actually showing up in the `Backends` workflow instead of this one.

cpcloud (changes requested):
Small bit about python 3.8 in the `Ibis` workflow, otherwise LGTM.

.github/workflows/ibis-main.yml
kszucs: And possibly 3.10 too. Though I'm not sure whether we need the extensive OS parametrization too.

kszucs: We already have 48 builds which seems to be a bit overwhelming.

cpcloud: :shrug: The number of builds that is overwhelming will vary with people. The builds finish relatively quickly given the extent of testing being done. I get what you're saying but I guess I'm more inclined to add tooling (hopefully in the form of a GitHub action, e.g., the test report) to make large build matrices manageable as opposed to cutting out tests to do so.

cpcloud: FWIW, Cutting out 3.8 is probably fine, unless there are concerns about 3.8 specific bugs that would show up in our builds.

cpcloud: My comment here was also about the fact that there is a 3.8 build for this set of parameters, but it's showing up in *Backends* in the UI and not in the *Ibis* workflow.

kszucs: That build seems to belong to the HDF5 backend, fixing.

.github/workflows/test-report.yml
```diff
@@ -1,7 +1,7 @@
 name: Test Report
 on:
   workflow_run:
-    workflows: ['CI']
+    workflows: ['Ibis', 'Backends']
```
kszucs, line 4: @cpcloud is it fine?

cpcloud: Yup, LGTM

.github/workflows/ibis-backends.yml
```diff
@@ -9,6 +9,10 @@ on:
     branches:
       - master
 
+concurrency:
+  group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
```
cpcloud, line 13: Are the semantics here to cancel any in progress builds when a new build shows up that:

1. Is in the ibis-project/ibis (or `$GITHUB_USER/ibis`) repo
2. Has the same head ref or SHA (the latter seems unlikely given SHA256)
3. Is the same workflow

Is my understanding correct?

kszucs: Yes. See additional explanation about the `${{ github.head_ref || github.sha }}` part here https://github.com/apache/arrow/pull/11520#issuecomment-952978639

cpcloud (approved):

PR #3112, opened 2021-11-04
refactor(ir): Remove ir.ExprList, ops.ExpressionList and ibis.expr_list()

kszucs:


ibis/expr/api.py
```diff
@@ -1265,35 +1264,30 @@ def _numeric_summary(arg, exact_nunique=False, prefix=None):
     -------
     summary : (count, # nulls, min, max, sum, mean, nunique)
     """
+    if exact_nunique:
+        unique_metric = arg.nunique().name('nunique')
+    else:
+        unique_metric = arg.approx_nunique().name('approx_nunique')
+
+    # TODO(kszucs): handle prefix
```
kszucs, line 1272: Seems like there are no tests for the `prefix` argument.

cpcloud:
@kszucs Can you add a release note?

ibis/expr/api.py
```diff
@@ -1234,23 +1233,29 @@ def _generic_summary(arg, exact_nunique=False, prefix=None):
...
         unique_metric = arg.nunique().name('uniques')
     else:
         unique_metric = arg.approx_nunique().name('uniques')
 
-    metrics.append(unique_metric)
-    return _wrap_summary_metrics(metrics, prefix)
+    metrics = [arg.count(), arg.isnull().sum().name('nulls'), unique_metric]
+
+    if prefix:
+        metrics = [m.name(f"{prefix}{m.get_name()}") for m in metrics]
```
cpcloud, line 1251: You could probably do this in one line with:

```python
[m.name(f"{prefix or ''}{m.get_name()}{suffix or ''}") for m in metrics]
```

ibis/expr/api.py
```diff
@@ -1260,40 +1265,34 @@ def _numeric_summary(arg, exact_nunique=False, prefix=None):
...
-def _wrap_summary_metrics(metrics, prefix):
-    result = expr_list(metrics)
-    if prefix is not None:
-        result = result.prefix(prefix)
-    return result
-
+    if prefix:
+        metrics = [m.name(f"{prefix}{m.get_name()}") for m in metrics]
+    if suffix:
+        metrics = [m.name(f"{m.get_name()}{suffix}") for m in metrics]
```
cpcloud, line 1293: Similarly here you can write this in one line.

cpcloud (approved):
LGTM

ibis/expr/api.py
kszucs: Silly me :) Updating.

kszucs:
@cpcloud updated

cpcloud:
This is a breaking change, but we believe it will have minimal effect due to `ibis.expr_list` being a fairly esoteric API.

PR #3117, opened 2021-11-08
perf: fast path for simple column selection

### LINKED ISSUE #2742: Add fast path for simple selections in dask backend
For simple selections we don't need to bother with the `dd.concat` here: https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/execution/selection.py#L154 and should probably select on the data directly

gerrymanoim:
This change generates much less intermediate code/operations/concat calls in the simple case where we are doing a simple selection or selection + rename. It still falls back to the projection logic in the case where we need to perform some computation (though lots of this can probably be cleaned up/refactored in the future).

Resolves #2742.

### DIFF
--- a/ibis/backends/dask/execution/selection.py
+++ b/ibis/backends/dask/execution/selection.py
@@ -4,7 +4,7 @@
 
 import functools
 import operator
-from typing import Optional
+from typing import List, Optional
 
 import dask.dataframe as dd
 import pandas
@@ -47,7 +47,6 @@ def compute_projection_scalar_expr(
     parent_table_op = parent.table.op()
 
     data_columns = frozenset(data.columns)
-
     scope = scope.merge_scopes(
         Scope(
             {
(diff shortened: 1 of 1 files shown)


ibis/backends/dask/execution/selection.py
```diff
@@ -140,23 +140,28 @@ def execute_selection_dataframe(
...
-                data,
-                scope=scope,
-                timecontext=timecontext,
-                **kwargs,
-            )
-            data_pieces.append(dask_object)
+        if all(isinstance(s.op(), ops.TableColumn) for s in selections) and {
+            s.op().name for s in selections
+        }.issubset(set(result.columns)):
+            result = result[[s.op().name for s in selections]]
```
jreback, line 146: add a comment that this is for straight projection

cpcloud:
@gerrymanoim You still working on this? If not, can we close it out?

gerrymanoim:
Still WIP

cpcloud:
@gerrymanoim Glad to see this out of draft mode! Ready for review?

gerrymanoim:
Yep!

ibis/backends/dask/execution/selection.py
```diff
@@ -129,6 +128,42 @@ def compute_projection_column_expr(
...
+) -> dd.DataFrame:
+    """Build up a df by doing direct selections, renaming if necessary."""
+    cols = [
+        (s.op().name, getattr(s, "_name", s.op().name)) for s in selections
+    ]
+    renamed_cols = {
+        col: renamed_col for col, renamed_col in cols if col != renamed_col
+    }
+
+    result = data[[col[0] for col in cols]]
```
cpcloud, line 142:
```suggestion
    result = data[[col for col, _ in cols]]
```

ibis/backends/dask/execution/selection.py
```diff
@@ -138,25 +173,22 @@ def execute_selection_dataframe(
...
-        # Create a unique row identifier and set it as the index. This is used
-        # in dd.concat to merge the pieces back together.
-        data = add_partitioned_sorted_column(data)
-        data_pieces = []
-        for selection in selections:
-            dask_object = compute_projection(
-                selection,
+        # if we are just performing select operations and all columns are in
+        # the table we can do a direct selection
+        if all(isinstance(s.op(), (ops.TableColumn)) for s in selections) and {
```
cpcloud, line 179:
```suggestion
        if all(isinstance(s.op(), ops.TableColumn) for s in selections) and {
```

ibis/backends/dask/execution/selection.py
```diff
@@ -138,25 +173,22 @@ def execute_selection_dataframe(
...
+            s.op().name for s in selections
+        }.issubset(set(result.columns)):
```
cpcloud, line 181: The second condition (checking that the op names are a subset of the `result.columns` in theory should be guaranteed by the flow of execution (child relations are evaluated and passed in). Is it necessary for another reason?

cpcloud (changes requested):

ibis/backends/dask/execution/selection.py
gerrymanoim: I _think_ this could be removed if I was able to work in the join selection logic https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/execution/selection.py#L82-L101 where the names have some postfixing. I can do that as a followup?

cpcloud: Makes sense. No need to create an issue though, it's fine to just submit the PR right after this one.

gerrymanoim:
Let me see if I can put something public facing together. We definitely see both a performance speedup (much easier to select than do this whole concat branch) and much less stress on the dask scheduler on large workflows (this generates a much simpler graph/fewer computations that the dask scheduler has to find a home for).

cpcloud:
Sounds good, that'd be great. Certainly not a blocker. Merging, thanks again for the PR!

cpcloud (approved):

PR #3123, opened 2021-11-16
refactor(ir): remove all occurrences of Arg(*)

kszucs:
Follow-up for https://github.com/ibis-project/ibis/pull/3047

Kept the `Argument()` function in `signature.py` should we remove it altogether?


jreback (changes requested):
this looks an api change. do you want to emit a warning if using Arg right?

kszucs:
> this looks an api change. do you want to emit a warning if using Arg right?

If we want to keep `Arg()` around then yes.

jreback:
> > this looks an api change. do you want to emit a warning if using Arg right?
> 
> If we want to keep `Arg()` around then yes.

well we don't want to keep it around, but breaking the world is not great. can you add a test of a FutureWarning if its used?

cpcloud:
@kszucs Let's keep the class around for now. Can we attach a deprecation warning by subclassing?

Something like this:

```python
class Arg(Argument):
        super().__init__(*args, **kwargs)
        warnings.warn("Deprecated in $SOME_VERSION")
```

cpcloud:
I guess that won't work if folks are aliasing, but that should continue to work I think, right?

cpcloud:
I don't think `Arg` was ever public API, since it's an import alias, though Python inability to control visibility makes everything a public API :disappointed:

kszucs:
Well, arg is not a class since https://github.com/ibis-project/ibis/pull/3047/files#diff-09f7d95a14c73bb445f8c811277d0b3be65436979662a505a2275e6059ae5237R74 so we cannot maintain "full" backward compatibility. I can raise a warning of course.

What's our deprication policy? Can we break API between consecutive major releases or do we want to warn for another major before actually breaking the API?

cpcloud:
IMO a major release implies at least one user-facing breaking API change, otherwise the version numbers are meaningless. Minor releases are effectively new features, patch releases are bug fix only.

kszucs:
More precisely my question is: can we break the API for 3.0 rather than raising a future warning?

cpcloud:
There are some other breaking API changes like #3138 that I think we could get in the next release. Alternatively, we can wait to merge those until after the next release.

kszucs:
Do you refer to 2.1 or 3.0 as the next release?

cpcloud:
3.0

kszucs:
If we are allowed to break API between major releases then why not to for 3.0? We could cherry-pick backward compatible changes for minor releases with additional future warnings for the breakages of the upcoming major release. 

I don't have a strong opinion though, just the deprecation policy is still not entirely clear to me.

jreback:
this is a break the world thing 
if backward compat is relatively easy then prefer to warn if at all possible for a release 

so ideally have full compat and show a future warning in 2.x then remove in 3

cpcloud:
If we're going to remove in 3 anyways, then let's release 3.0 next. There's not much to be gained by releasing 2.x AFAICT.

kszucs:
@jreback @cpcloud deprecated the `Argument` function.

ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -161,20 +161,23 @@ def _not_contains(t, expr):
 def reduction(sa_func):
     def formatter(t, expr):
         op = expr.op()
-        *args, where = op.args
-
-        return _reduction_format(t, sa_func, where, *args)
+        if op.where is not None:
+            arg = t.translate(op.where.ifelse(op.arg, ibis.NA))
+        else:
+            arg = t.translate(op.arg)
```
cpcloud, line 167: I guess you could also eliminate the duplicated call to `t.translate`, but probably not worth it :)

ibis/backends/base/sql/compiler/select_builder.py
```diff
@@ -559,9 +559,7 @@ def _visit_filter(self, expr):
                 return expr._factory(type(op)(left, right))
             else:
                 return expr
-        elif isinstance(
-            op, (ops.Any, ops.BooleanValueOp, ops.TableColumn, ops.Literal)
```
cpcloud, line 563: Was `ops.BooleanValueOp` just never hit?

ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -422,7 +425,7 @@ def _sort_key(t, expr):
     ops.Min: reduction(sa.func.min),
     ops.Max: reduction(sa.func.max),
     ops.CountDistinct: _count_distinct,
-    ops.GroupConcat: reduction(sa.func.group_concat),
+    ops.GroupConcat: _group_concat,
```
cpcloud, line 428: I guess this can't be a reduction because of the `sep` argument?

ibis/expr/signature.py
```diff
@@ -81,6 +82,11 @@ def Argument(validator, default=EMPTY):
         If callable is passed, it will be executed just before the inner,
         and itsreturn value will be treaded as default.
     """
+    warnings.warn(
+        "ibis.expr.signature.Argument will be removed in ibis 3.0",
+        FutureWarning,
```
cpcloud, line 87: I don't think this is necessary if we bump straight to 3.0

kszucs: Will we bump straight to 3.0?

ibis/backends/base/sql/alchemy/registry.py
kszucs: Correct.

ibis/backends/base/sql/compiler/select_builder.py
kszucs: Correct.

ibis/backends/base/sql/alchemy/registry.py
kszucs: Then I'm going to leave it as is.

ibis/expr/signature.py
cpcloud: I think we should

kszucs: Ok, then I'm removing the deprecation warning.

ibis/expr/operations/reductions.py
```diff
@@ -14,25 +13,27 @@ class Reduction(ValueOp):
...
     def output_type(self):
         return functools.partial(ir.IntegerScalar, dtype=dt.int64)
 
 
 @public
-class Arbitrary(Reduction):
-    arg = Arg(rlz.column(rlz.any))
-    how = Arg(rlz.isin({'first', 'last', 'heavy'}), default=None)
-    where = Arg(rlz.boolean, default=None)
+class Arbitrary(Filterable, Reduction):
```
kszucs, line 29: @cpcloud Note that introducing the `Filterable` parent changes the previous argument order:

```python
Arbitrary(arg, how, where) -> Arbitrary(arg, where, how)
```

I can restore the previous behavior if desired, but this enables better code reuse.

kszucs: See https://github.com/ibis-project/ibis/pull/3123/commits/c5d23caf516161edc371d6d386abe7cfaf152669

ibis/tests/expr/test_signature.py
```diff
@@ -4,71 +4,50 @@
...
-            raise TypeError(float)
+        if not isinstance(arg, self.typ):
+            raise TypeError(self.typ)
         return arg
 
 
+IsAny = InstanceOf(object)
+IsBoolean = InstanceOf(bool)
+IsFloat = InstanceOf(float)
+IsInteger = InstanceOf(int)
```
cpcloud, line 31: Let's calls this `IsInt` to match the `int`, same with `IsBool`/`bool` and `IsStr`/`str`

jreback:
will look soon

cpcloud (changes requested):

cpcloud:
@kszucs Can you rebase on `master` and also adjust the final commit message to conform to `cz schema`:

```
<type>(<scope>): <subject>
<BLANK LINE>
<body>
<BLANK LINE>
(BREAKING CHANGE: )<footer>
```

So, in this case it'd be something like:

```
refactor(ir): remove all occurrences of Arg(*)

Occurrences of Arg were removed as well as the Argument import alias

BREAKING CHANGE: Arg and Argument are no longer available in the API
```

kszucs:
@cpcloud updated

cpcloud:
@kszucs Can you squash to the final commit? Will merge after that.

jreback:
umm before u start breaking changes would be good to do a point release

jreback (changes requested):
actually this shows a warning right?

kszucs:
> actually this shows a warning right?

No, removed the warning https://github.com/ibis-project/ibis/pull/3123#discussion_r766111019

cpcloud (approved):

cpcloud:
@jreback Can you approve?

jreback:
would be nicer if we can deprecate & accept the current usage for a little

cpcloud:
@kszucs Mind keeping the alias around until 4.x? If it's a huge pain to make the `import` warn then I wouldn't worry about it.

kszucs:
@cpcloud updated

ibis/expr/signature.py
```diff
@@ -81,17 +102,22 @@ def Argument(validator, default=EMPTY):
         If callable is passed, it will be executed just before the inner,
         and itsreturn value will be treaded as default.
     """
+    warnings.warn(
+        "ibis.expr.signature.Argument will be removed in ibis 3.0",
```
cpcloud, line 106: Should be changed to 4.0.0 right? The idea is to keep it around for one more major release and then remove it.

Also, there's a new `ibis.util.deprecated` decorator to use here.

cpcloud (changes requested):
Small nit about using `ibis.util.deprecated` and the version to remove being 4.0.0, other than that it's good to go.

ibis/tests/expr/test_signature.py
```diff
@@ -47,6 +56,12 @@ class MagicString(StringOp):
     baz = Argument(int)
 
 
+def test_argument_is_deprecated():
+    msg = r"\"Argument\" is deprecated since v3\.0; use Validator\."
```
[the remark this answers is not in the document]
kszucs: from a recently added deprecation utility https://github.com/ibis-project/ibis/pull/3155, we can improve it in a follow-up

cc @saulpw

cpcloud (approved):

cpcloud:
@jreback Can you approve?

jreback (approved):

PR #3124, opened 2021-11-18
fix(datatypes): infer bytes literal as binary #2915

### LINKED ISSUE #2915: BUG: Binary literals' type is not inferred
Is there any reason not to infer `bytes` literals as `binary`?

```
>>> import ibis
>>> ibis.literal(b"abc")
  File "/home/cloud/src/ibis/ibis/expr/types.py", line 1288, in literal
    raise TypeError(
TypeError: The datatype of value b'abc' cannot be inferred, try passing it explicitly with the `type` keyword.
```

saulpw:
Closes #2915

### DIFF
--- a/ibis/expr/datatypes.py
+++ b/ibis/expr/datatypes.py
@@ -1232,6 +1232,11 @@ def infer_string(value: str) -> String:
     return string
 
 
+@infer.register(bytes)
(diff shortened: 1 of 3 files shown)


cpcloud (approved):
LGTM. Merging. Thanks!

PR #3127, opened 2021-11-20
fix(sqlite): defer db connection until needed

### LINKED ISSUE #64: test: detect table reference pathological issues during SQL translation
<a href="http://github.mtv.cloudera.com/wesm"></img></a> **Issue by [wesm](http://github.mtv.cloudera.com/wesm)**
_Sunday Jan 18, 2015 at 21:49 GMT_
_Originally opened as http://github.mtv.cloudera.com/wesm/ibis/issues/64_

---

Tables used in building an expression could come from different databases altogether; while pathological it would be good to preemptively address this and raise a helpful error message

### LINKED ISSUE #1768: ADD REFRESH capability to Postgres Client
When using the Postgres client, if a table is added to the database/schema after I create a connection, then when trying to refer to that new table I get a `NoSuchTableError`. If I browse the database via a well-established IDE (pgAdmin or DBeaver), then I see that the table is clearly present. 

There should be a way to refresh (manually or automatically) ibis's cached understanding of what schema and tables exist. 

I am using the following version of ibis on Windows:
```
version: 1.0.0
revisionid: 33201ac9cdbdafde
```

saulpw:
- to .execute() the same table (or a copy) from multiple threads

Closes #64
Closes #1768

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -151,10 +151,10 @@ class BaseBackend(abc.ABC):
         )
 
     def __hash__(self):
-        return hash(self.datasource)
+        return hash(self.db_identity)
 
     def __eq__(self, other):
-        return self.datasource == other.datasource
+        return self.db_identity == other.db_identity
 
     @property
     @abc.abstractmethod
@@ -164,10 +164,10 @@ class BaseBackend(abc.ABC):
         """
 
     @cached_property
-    def datasource(self) -> str:
+    def db_identity(self) -> str:
         """
         Identity of the database.  Multiple connections to the same
-        database will have the same datasource.  Default implementation
+        database will have the same db_identity.  Default implementation
         assumes connection parameters uniquely specify the database.
         """
         parts = [self.table_class.__name__]
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import abc
 import re
 import warnings
-from typing import Any, Callable, List, Dict, Tuple
+from typing import Any, Callable, Dict, List, Tuple
 
 import ibis
 import ibis.expr.operations as ops
@@ -135,15 +135,16 @@ class BaseBackend(abc.ABC):
 
     database_class = Database
     table_class: type[ops.DatabaseTable] = ops.DatabaseTable
-    _con_args: Tuple[Any]
-    _con_kwargs: Dict[str, Any]
+    _con_args: tuple[Any]
+    _con_kwargs: dict[str, Any]
 
     def __getstate__(self):
         return dict(
             database_class=self.database_class,
             table_class=self.table_class,
             _con_args=self._con_args,
-            _con_kwargs=self._con_kwargs)
+            _con_kwargs=self._con_kwargs,
+        )
 
     def __hash__(self):
         return hash(self.datasource)
@@ -165,9 +166,10 @@ class BaseBackend(abc.ABC):
         database will have the same datasource.  Default implementation
         assumes connect
         """
-        return '_'.join(str(x) for x in (self.table_class,
-                                         self._con_args,
-                                         self._con_kwargs))
+        return '_'.join(
+            str(x)
+            for x in (self.table_class, self._con_args, self._con_kwargs)
+        )
 
     def connect(self, *args, **kwargs):
         """
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -168,10 +168,10 @@ class BaseBackend(abc.ABC):
         database will have the same datasource.  Default implementation
         assumes connection parameters uniquely specify the database.
         """
-        return '_'.join(
-            str(x)
-            for x in (self.table_class, self._con_args, self._con_kwargs)
-        )
+        parts = [self.table_class.__name__]
+        parts.extend(self._con_args)
+        parts.extend(f'{k}={v}' for k, v in self._con_kwargs.items())
+        return '_'.join(map(str, parts))
 
     def connect(self, *args, **kwargs):
         """
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import abc
 import re
 import warnings
-from typing import Any, Callable
+from typing import Any, Callable, List, Dict, Tuple
 
 import ibis
 import ibis.expr.operations as ops
@@ -135,8 +135,15 @@ class BaseBackend(abc.ABC):
 
     database_class = Database
     table_class: type[ops.DatabaseTable] = ops.DatabaseTable
-    con_str = None
-    con_options = None
+    _con_args: Tuple[Any]
+    _con_kwargs: Dict[str, Any]
+
+    def __getstate__(self):
+        return dict(
+            database_class=self.database_class,
+            table_class=self.table_class,
+            _con_args=self._con_args,
+            _con_kwargs=self._con_kwargs)
 
     @property
     @abc.abstractmethod
@@ -145,21 +152,27 @@ class BaseBackend(abc.ABC):
         Name of the backend, for example 'sqlite'.
         """
 
-    def connect(self, *args, **options):
+    def connect(self, *args, **kwargs):
         """
-        Return client object, ready to connect (via do_connect())
-            to given *connection_string* with given *options*.
+        Return new client object with saved args/kwargs, having called
+        .reconnect() on it.
         """
         new_backend = self.__class__()
-        new_backend.con_args = args
-        new_backend.con_options = options
-        new_backend.do_connect(*args, **options)
+        new_backend._con_args = args
+        new_backend._con_kwargs = kwargs
+        new_backend.reconnect()
         return new_backend
 
-    @abc.abstractmethod
-    def do_connect(self, connection_string, **options):
+    def reconnect(self):
         """
-        Connect to underlying database and set internal state.
+        Reconnect to the target database already configured with connect().
+        """
+        self.do_connect(*self._con_args, **self._con_kwargs)
+
+    @abc.abstractmethod
+    def do_connect(self, *args, **kwargs):
+        """
+        Connect to database specified by args and kwargs.
         """
 
     def database(self, name: str = None) -> Database:
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -137,8 +137,8 @@ class BaseBackend(abc.ABC):
     table_class: type[ops.DatabaseTable] = ops.DatabaseTable
 
     def __init__(self):
-        self._con_args: Tuple[Any] = tuple()
-        self._con_kwargs: Dict[str, Any] = dict()
+        self._con_args: tuple[Any] = tuple()
+        self._con_kwargs: dict[str, Any] = dict()
 
     def __getstate__(self):
         return dict(
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -135,8 +135,10 @@ class BaseBackend(abc.ABC):
 
     database_class = Database
     table_class: type[ops.DatabaseTable] = ops.DatabaseTable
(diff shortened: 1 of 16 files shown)


jreback (changes requested):
umm at the very minimum this needs tests

can u show an example why this is needed?

saulpw:
> umm at the very minimum this needs tests
> 
> can u show an example why this is needed?

Yes, sorry, this isn't a complete PR, I wanted to submit it early so we could discuss how to clean it up and how to test it.

This is needed because the sqlite connection is created in the thread that gets a reference to the table, and if another thread tries to use that table, sqlite/sqlalchemy gets confused during execute and raises an Exception with a message like `no such table: base.dept`.  (`"base"` is the default database_name).  Doing both actions in the same thread works fine, however.

So this defers the connection creation until it's needed, and also allows the objects to be copied so they can be used in other threads later (new connections will be created for them).  This works adequately for my use case.

I'll talk with @cpcloud on Monday about what a proper test for this looks like.

ibis/backends/base/__init__.py
```diff
@@ -135,6 +135,8 @@ class BaseBackend(abc.ABC):
 
     database_class = Database
     table_class: type[ops.DatabaseTable] = ops.DatabaseTable
+    con_str = None
```
jreback, line 138: none of this can be here

it should be in the sql base

ibis/backends/base/file/__init__.py
```diff
@@ -56,7 +56,7 @@ class BaseFileBackend(BaseBackend):
 
     database_class = FileDatabase
 
-    def connect(self, path):
```
jreback, line 59: you cannot rename any of these methods as they r public

jreback (changes requested):

ibis/backends/base/__init__.py
saulpw: Agreed, I'll move them to a better place.

saulpw:
Ibis needs a connection to the data source in order to create expressions which contain reference the connection to their data source. I need to be able to copy (or better, pickle) these expressions and call `.execute()` on those copies from different threads.

So this PR proposes a small API change.  Currently, `connect()` does two things: it creates an object, and it connects the object to the source.  With this PR, `connect()` still creates the object, and the actual source-connecting is moved to `do_connect()`.  This way, Expr derived from these Backends can be `copy()`ed and each copy can make its own connection on demand.

The `connect()` constructor turns out to be boilerplate and can be given in BaseBackend so backends don't need to override it anymore.  But every backend will need to override `do_connect()`, which just needs to do the connecting (not the actual object creation).

Here's an example of a current connect():

```
    def connect(self, ...):
        new_backend = self.__class__()
        new_backend._context = foodb.connect(...)
        return new_backend
```

and here's the replacement `do_connect()`:
```
    def do_connect(self, ...):
        self._context = foodb.connect(...)
```

This is a mechanical transformation and even results in a bit less code.  The signature for `do_connect` should be identical; the args and kwargs are saved and passed through verbatim.

All backends will eventually need a `__getstate__` that does not include unreusable connection context. Most Backends can rely on the parent `BaseBackend.__getstate__` and possibly include some conveniences that connections can share, like:

```
    def __getstate__(self):
        r = super().__getstate__()
        r.update(dict(_session_token=self._session_token))
        return r
```

I think maybe there should be an API method for invalidating the connection.  Or we could make `do_connect` a forced invalidation; but maybe it should be allowed to be a no-op in case it wants to always reuse the existing connection?

If we can come up with better names than `connect` and `do_connect`, we can use those instead.

ibis/backends/base/file/__init__.py
saulpw: You're right, I'm proposing an API change.  (See the toplevel comment I just posted.)

cpcloud: @jreback The top level API here isn't changing. Every backend still has the same `connect` API by way of threading args and options through `do_connect`.

jreback: sure it is. this is leaking the `db_connect` into the base class where its not needed. you can move the `base` changes to the sql layer i think and achieve what you want.

cpcloud: Right, do_connect would be required for new backends or those that want to upgrade. I don't think I see the problem, we'd release a new major version.

ibis/backends/base/__init__.py
saulpw: Moved to `__init__` constructors.

ibis/tests/expr/test_table.py
```diff
@@ -1404,3 +1405,23 @@ def test_mutate_chain():
...
+    Expr from multiple connections to same DB should be compatible.
+    """
+    con1 = MockBackend()
+    con2 = MockBackend()
+
+    con1.table('alltypes').union(con2.table('alltypes')).execute()
+
+
+@pytest.mark.xfail(
+    raises=RelationError, reason='Expr from different DB should fail #64'
```
saulpw, line 1421: This should also resolve #64

cpcloud: Cool, can you add `Closes #64` to the PR description?

cpcloud (commented):
Almost there, small comment about constructor arguments.

ibis/backends/base/__init__.py
```diff
@@ -136,17 +136,64 @@ class BaseBackend(abc.ABC):
...
+        parts.extend(self._con_args)
+        parts.extend(f'{k}={v}' for k, v in self._con_kwargs.items())
+        return '_'.join(map(str, parts))
+
+    def connect(self, *args, **kwargs):
+        """
+        Return new client object with saved args/kwargs, having called
+        .reconnect() on it.
+        """
+        new_backend = self.__class__()
```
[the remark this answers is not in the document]
saulpw: So do you mean, having a default BaseBackend() constructor that takes the same `*args, **kwargs` parameters as its `connect()`/`do_connect()` methods would take?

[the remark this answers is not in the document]
saulpw: I pushed a change, let me know if this is what you meant.

cpcloud (approved):
LGTM. Over to you @jreback

jreback:
will look soon

ibis/backends/base/__init__.py
```diff
@@ -136,17 +136,62 @@ class BaseBackend(abc.ABC):
...
+        Identity of the database.  Multiple connections to the same
+        database will have the same datasource.  Default implementation
+        assumes connection parameters uniquely specify the database.
+        """
+        parts = [self.table_class.__name__]
+        parts.extend(self._con_args)
+        parts.extend(f'{k}={v}' for k, v in self._con_kwargs.items())
+        return '_'.join(map(str, parts))
+
+    def connect(self, *args, **kwargs):
```
jreback, line 176: can you type the return

ibis/backends/base/__init__.py
```diff
@@ -136,17 +136,62 @@ class BaseBackend(abc.ABC):
...
+        return self.datasource == other.datasource
+
     @property
     @abc.abstractmethod
     def name(self) -> str:
         """
         Name of the backend, for example 'sqlite'.
         """
 
+    @property
```
jreback, line 164: this can be a cached_property right?

ibis/backends/base/__init__.py
```diff
@@ -136,17 +136,62 @@ class BaseBackend(abc.ABC):
...
+    def datasource(self) -> str:
```
jreback, line 165: can we call this db_identity? (unless you have a definition for datasource here)

ibis/backends/base/__init__.py
```diff
@@ -136,17 +136,62 @@ class BaseBackend(abc.ABC):
...
+        return new_backend
+
+    def reconnect(self):
+        """
+        Reconnect to the target database already configured with connect().
+        """
+        self.do_connect(*self._con_args, **self._con_kwargs)
+
     @abc.abstractmethod
-    def connect(connection_string, **options):
```
jreback, line 147: can we preserve this for backwards compatibilty

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -101,13 +101,11 @@ def _build_alchemy_url(
...
         )
 
-    def connect(self, con: sqlalchemy.engine.Engine):
-        new_backend = self.__class__()
-        new_backend.con = con
-        new_backend.meta = sqlalchemy.MetaData(bind=con)
-        new_backend._inspector = sqlalchemy.inspect(con)
-        new_backend._schemas: Dict[str, sch.Schema] = {}
-        return new_backend
+    def do_connect(self, con):
```
jreback, line 104: can you type con (and the return)

jreback (changes requested):
i don't understand why we have to break the world by renaming connect -> do_connect?

ibis/backends/base/__init__.py
saulpw: Sure, it's not very expensive though.

ibis/backends/base/__init__.py
saulpw: @pcloud @gforsyth what do you think?  Is `db_identity` a better name than `datasource`?

ibis/backends/base/__init__.py
saulpw: If you look a few lines above this, it's still there, so you can continue to use `connect()` like you always have.

ibis/backends/base/__init__.py
jreback: ok if its not then nbd

ibis/backends/base/__init__.py
jreback: ahh fi that's the case, then just document connect & db_connect a bit

saulpw: Okay, I'm not clear on what's missing, or where this documentation would go.  Backends need to override the abstractmethod `do_connect`, but that's virtually identical to what they had to do previously with `connect` (as shown in [this comment in the PR](https://github.com/ibis-project/ibis/pull/3127#issuecomment-977467918)).  What else do you need?

ibis/backends/base/__init__.py
gforsyth: `db_identity` seems less semantically overloaded, and does track with the first line of the docstring

jreback (changes requested):
also can you add a whatsnew note and describe the change

cpcloud:
> also can you add a whatsnew note and describe the change

On a separate note, can you take a look at #2924? This moves us much closer to automate release notes based on commit messages so we don't have to keep asking people to write release notes.

ibis/backends/base/__init__.py
saulpw: Okay, changed and pushed.

ibis/backends/base/sql/alchemy/__init__.py
saulpw: Done.

ibis/backends/base/__init__.py
saulpw: Done.

cpcloud:
This looks good to me. @jreback ?

cpcloud (approved):

jreback (approved):

jreback:
@cpcloud this appears to have broken ibis-bigquery, but they don't test against master. They are a pretty good proxy for other backends.

I am not sure what to do here as there are a lot of breaking changes in the pipelines.

cpcloud:
`ibis-bigquery` has an unrealistic `requirements.txt` with unbounded versions for every dependency. This is bound to break at some point and certainly there's no guarantee that `master` is compatible with dependents.

cpcloud:
Ah looks like `setup.py` has constraints.

cpcloud:
I believe that if we remove the `abstractmethod` decorator we can avoid the breakage, and reintroduce it in 3.0.

jreback:
we shouldnt be constantly breaking downstream like this.

cpcloud:
Yep. I don't think we at "constantly" yet. The goal is to avoid breakage between major releases. A major will always include at least one breaking change.

PR #3128, opened 2021-11-20
docs(index): move header above toctree

saulpw:


saulpw:
Noticed this while reading the docs and was confused for a second.  The header should separate the "needing contributors" list from the toctree.

cpcloud (approved):
LGTM

Issue #3129, opened 2021-11-22
refactor: normalize underlying types of literal values


cpcloud:
Currently many of the literal types we support accept multiple values. This is desirable, since not everyone will come to ibis with the same kind of literal values.

However, we don't normalize the underlying value to a single type in some cases.

For example, with the `uuid` type we keep whatever value the user passed in:

```
In [27]: import uuid

In [28]: import ibis

In [29]: value = uuid.uuid4()

In [30]: x = ibis.literal(value, type='uuid')

In [31]: x
Out[31]:
Literal[uuid]
  e9cd2672-e276-4bef-acec-68423d3c6ff2

In [32]: y = ibis.literal(str(value), type='uuid')

In [33]: y
Out[33]:
Literal[uuid]
  e9cd2672-e276-4bef-acec-68423d3c6ff2

In [34]: x.equals(y)
Out[34]: False
```

We should address this by adding N to 1 tests: N different user-input types all compare equal to the normalized-type input.

PR #3133, opened 2021-11-22
refactor: normalize underlying types of literals

gforsyth:
This is a first cut towards resolving #3129

It adds a `_normalize` method to the `Expr` class that defaults to a
no-op, this then gets overridden in the type classes (for those types
where it is possible to have ambiuous underlying types).

For instance, there is no override for normalizing integer inputs if the
input argument is a float, since that will already be rejected by the
current type-checking.  Instead, this is normalizing values to the
closest Pythonic type available for those types that are sufficiently
ambiguous.

In this first cut, the following normalizations are made:

* `str` are cast to `uuid.UUID` if `type="uuid"`
* `int` are cast to `float` if `type=("float" | "float16" | "float32" | "float64")`
* `UUID` are cast to `str` if `type="uuid"`

The current main point of ugliness (to me) is that there are
`FloatConstant` values like `E()` that we cannot call `float()` on.
That's fine, but if I import `ops.generic` to do an `isinstance`
check it causes a circular import and everything goes boom.

There are almost certainly other types which can be normalized, this was
a first pass and we can add more.


ibis/expr/types.py
```diff
@@ -63,6 +64,13 @@ def _repr(self, memo: 'Optional[FormatMemo]' = None):
 
         return ExprFormatter(self, memo=memo).get_result()
 
+    @staticmethod
+    def _normalize(arg):
```
cpcloud, line 68: What do you think about a dispatcher for literal values that gets called in the `ops.Literal` constructor?

Something like

```python

_normalize = Dispatcher("_normalize")

@_normalize.register(dt.Float64, (int, float))
def _normalize(type, value) -> float:
    return float(value)
```

Then this would be called wherever the `value` is passed into `ops.Literal`.

gforsyth: oooh, yeah, I like that a lot

ibis/expr/operations/generic.py
```diff
@@ -257,6 +288,9 @@ class Literal(ValueOp):
     )
     dtype = Arg(rlz.datatype)
 
+    def _validate(self):
+        self.value = _normalize(self.dtype, self.value)
```
gforsyth, line 292: Currently hooking in here since `_validate` already exists, but happy to add something purpose-built

cpcloud: I think you can call `_normalize(value)` here instead: https://github.com/ibis-project/ibis/blob/master/ibis/expr/types.py#L1242

ibis/expr/operations/generic.py
```diff
@@ -221,6 +222,36 @@ class Least(CoalesceLike):
...
+
+@_normalize.register(dt.String, uuid.UUID)
+def _uuid_to_str(type: dt.String, value: uuid.UUID) -> str:
+    return str(value)
+
+
+@_normalize.register(dt.UUID, uuid.UUID)
+def _uuid_to_uuid(type: dt.UUID, value: uuid.UUID) -> uuid.UUID:
+    """Need this to override _uuid_to_str since dt.UUID is a child of
+    dt.String"""
```
cpcloud, line 251: Yikes. I don't think that's correct, I'll make an issue.

ibis/tests/expr/test_literal.py
```diff
@@ -60,3 +64,31 @@ def test_pickle_literal_interval():
...
+        pytest.param(0.0, "float", float, id="float_zero_as_float"),
+        pytest.param(42, "float", float, id="int_as_float"),
+        pytest.param(42.0, "float", float, id="float_as_float"),
+        pytest.param(42.0, None, float, id="float_implicit_type_as_float"),
+    ],
+)
+def test_normalized_underlying_value(userinput, literal_type, expected_type):
+    a = ibis.literal(userinput, type=literal_type)
+
+    assert isinstance(a._arg.value, expected_type)
```
cpcloud, line 94: Typically `.op()` is used (`.op()` is the public API for accessing the underlying operation) mostly to avoid potential changes to the name of `._arg`. In practice this hasn't really changed much though so it's not a big deal if you want to leave it as is.

ibis/expr/operations/generic.py
gforsyth: I can fix that here or open a separate PR

ibis/tests/expr/test_literal.py
gforsyth: I'm all for using public APIs

ibis/expr/operations/generic.py
kszucs: How about adding the functionality to `datatypes.py` instead?

ibis/expr/datatypes.py
```diff
@@ -801,7 +801,7 @@ class MultiPolygon(GeoSpatial):
     __slots__ = ()
 
 
-class UUID(String):
+class UUID(DataType):
```
cpcloud, line 804: This should probably be in a separate PR since it's technically a breaking change. All of the `datatypes.py` classes are public at the moment.

ibis/expr/operations/generic.py
gforsyth: Moved `_normalize` into `types.py` and defined the `Dispatcher` in `datatypes.py`

ibis/expr/datatypes.py
gforsyth: Removed the UUID change and will push up a fast-follow PR for this change

cpcloud (approved):
LGTM, blocked on #3137

cpcloud:
Merging on green.

PR #3138, opened 2021-11-29
refactor: dt.UUID inherits from DataType, not String

### LINKED ISSUE #3136: bug: UUID type is a subclass of String
UUIDs are not strings, they are really byte blobs with various representations, one of which is a string.

We should remove the String superclass.

I don't think UUID should have any parent class other than DataType.

_Originally posted by @cpcloud in https://github.com/ibis-project/ibis/pull/3133#r755252233_

gforsyth:
Fixes #3136 

You can still cast from UUID to String, but UUID no longer inherits from string (this also removes the need for an extra dispatch rule in the `_normalize` function)

### DIFF
--- a/ibis/expr/datatypes.py
+++ b/ibis/expr/datatypes.py
@@ -802,7 +802,7 @@ class MultiPolygon(GeoSpatial):
     __slots__ = ()
 
 
-class UUID(String):
+class UUID(DataType):
     """A universally unique identifier (UUID) is a 128-bit number used to
     identify information in computer systems.
     """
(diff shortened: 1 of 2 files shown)


cpcloud:
@gforsyth Can you rebase and format your commit like this?

```
refactor: dt.UUID inherits from DataType, not String

Short description

BREAKING CHANGE: Description of what breaks
```

gforsyth:
Rebased and reformatted -- I'll keep an eye on CI and ping here when it's green

gforsyth:
Ready for another look @cpcloud !

cpcloud (approved):

cpcloud:
Thanks @gforsyth !

Issue #3152, opened 2021-12-09
Allow column_from to take `ColumnExpr` 


gerrymanoim:
Right now `rlz.column_from` https://github.com/ibis-project/ibis/blob/master/ibis/expr/rules.py#L426 only allows you to pass `str` and `int`. Given that it is returning the column itself, I'd like to change this to allow it to take a `ColumnExpr` and pass it through unchanged. 

cc @jreback

cpcloud:
`rlz.column_from` should not accept arbitrary `ColumnExpr`s, it's designed to be a rule that pulls out a column from a child table by name or schema index. Passing `ColumnExpr`s through without validation would allow invalid columns through.

It'd be fine to allow them through if the result of accessing them by name or index equals the input column expression, though that seems redundant.

@gerrymanoim Can you describe the use case a bit more?

cpcloud:
@gerrymanoim If you just want to allow any column expression, you can use `rlz.column(rlz.any)` for that.

gerrymanoim:
We have certain logic that takes ibis nodes and rewrites them, creating a new operation. Because we can't pass a `ColumnExpr`, we can't take the arguments from something like a `DropNa` and pass them in again to a new `DropNa`. This we now need to special case, just to extract the column names from the list of column exprs, so we could pass them back in again. 

I'd be fine checking if the root table is the same or if the getattr produces the same expr.

cpcloud:
> I'd be fine checking if the root table is the same or if the getattr produces the same expr.

Cool, the latter sounds good to me. Want to PR it? It should be a pretty small change and test.

PR #3153, opened 2021-12-09
feat: allow column_of to take a column expression

gerrymanoim:
Closes https://github.com/ibis-project/ibis/issues/3152

### DIFF
--- a/ibis/expr/rules.py
+++ b/ibis/expr/rules.py
@@ -424,11 +424,40 @@ def table(arg, *, schema=None, **kwargs):
 
 @validator
 def column_from(name, column, *, this):
-    if not isinstance(column, (str, int)):
-        raise com.IbisTypeError(
-            f"value must be an int or str, got {type(column).__name__}"
-        )
-    return getattr(this, name)[column]
+    """A column from a named table.
+
+    This validator accepts columns passed as string, integer, or column
+    expression. In the case of a column expression, this validator
+    checks if the column in the table is equal to the column being
+    passed.
+    """
+    if not hasattr(this, name):
+        raise com.IbisTypeError(f"Could not get table {name} from {this}")
+    table = getattr(this, name)
+
+    if isinstance(column, (str, int)):
+        return table[column]
+    elif isinstance(column, ir.AnyColumn):
+        if not column.has_name():
+            raise com.IbisTypeError(f"Passed column {column} has no name")
+
+        maybe_column = column.get_name()
+        try:
+            if column.equals(table[maybe_column]):
+                return column
+            else:
+                raise com.IbisTypeError(
+                    f"Passed column is not a column in {table}"
+                )
+        except com.IbisError:
+            raise com.IbisTypeError(
+                f"Cannot get column {maybe_column} from {table}"
+            )
+
+    raise com.IbisTypeError(
+        "value must be an int or str or AnyColumn, got "
+        f"{type(column).__name__}"
+    )
 
 
 @validator
--- a/ibis/tests/expr/test_rules.py
+++ b/ibis/tests/expr/test_rules.py
@@ -14,6 +14,10 @@ table = ibis.table(
     [('int_col', 'int64'), ('string_col', 'string'), ('double_col', 'double')]
 )
 
+similar_table = ibis.table(
+    [('int_col', 'int64'), ('string_col', 'string'), ('double_col', 'double')]
+)
+
 
 @pytest.mark.parametrize(
     ('value', 'expected'),
@@ -288,6 +292,43 @@ def test_valid_column_or_scalar(validator, value, expected):
 def test_invalid_column_or_scalar(validator, value, expected):
     with pytest.raises(expected):
         validator(value)
+
+
+@pytest.mark.parametrize(
+    ('check_table', 'value', 'expected'),
+    [
+        (table, "int_col", table.int_col),
+        (table, table.int_col, table.int_col),
+    ],
+)
+def test_valid_column_from(check_table, value, expected):
+    class Test:
+        table = check_table
+
+    validator = rlz.column_from("table")
+    assert validator(value, this=Test()).equals(expected)
+
+
+@pytest.mark.parametrize(
+    ('check_table', 'validator', 'value'),
+    [
+        (table, rlz.column_from("not_table"), "int_col"),
+        (table, rlz.column_from("table"), "col_not_in_table"),
+        (
+            table,
+            rlz.column_from("table"),
+            similar_table.int_col,
+        ),
+    ],
+)
+def test_invalid_column_from(check_table, validator, value):
+    class Test:
+        table = check_table
+
+    test = Test()
+
+    with pytest.raises(IbisTypeError):
+        validator(value, this=test)
 
 
 @pytest.mark.parametrize(


gerrymanoim:
Ah this was not the correct way to do this. Will update when I can get to my personal computer.

ibis/expr/rules.py
```diff
@@ -425,9 +425,12 @@ def table(arg, *, schema=None, **kwargs):
 @validator
 def column_from(name, column, *, this):
     if not isinstance(column, (str, int)):
-        raise com.IbisTypeError(
-            f"value must be an int or str, got {type(column).__name__}"
-        )
+        if isinstance(column, ir.AnyColumn):
+            column = column.op().name
```
cpcloud, line 429: You can't just look at the name, because a column operation's name (or index) may match `column` while originating from a different table.

cpcloud: You can do something like:

```python
if isinstance(column, ir.ColumnExpr):
    # return column if getattr(this, name)[column_name] equals column, otherwise fail
else:
    # existing logic
```

cpcloud (changes requested):

ibis/expr/rules.py
```diff
@@ -424,11 +424,17 @@ def table(arg, *, schema=None, **kwargs):
...
 def column_from(name, column, *, this):
-    if not isinstance(column, (str, int)):
-        raise com.IbisTypeError(
-            f"value must be an int or str, got {type(column).__name__}"
-        )
-    return getattr(this, name)[column]
+    if isinstance(column, ir.AnyColumn) and hasattr(column.op(), "name"):
+        maybe_column = column.op().name
+        if column.equals(getattr(this, name)[maybe_column]):
+            return column
```
cpcloud, line 430: Probably should give a more specific error message in the `else:` clause like `passed column is not a column in {getattr(this, name)}`

ibis/expr/rules.py
```diff
@@ -424,11 +424,17 @@ def table(arg, *, schema=None, **kwargs):
...
+    elif isinstance(column, (str, int)):
+        return getattr(this, name)[column]
```
cpcloud, line 432: Can you pull up `getattr(this, name)` into a variable (it's a table)? No reason to repeat it.

ibis/expr/rules.py
```diff
@@ -424,11 +424,17 @@ def table(arg, *, schema=None, **kwargs):
 
 @validator
 def column_from(name, column, *, this):
-    if not isinstance(column, (str, int)):
-        raise com.IbisTypeError(
-            f"value must be an int or str, got {type(column).__name__}"
-        )
-    return getattr(this, name)[column]
+    if isinstance(column, ir.AnyColumn) and hasattr(column.op(), "name"):
```
cpcloud, line 427: It might also be useful to short circuit here and raise an error if the column doesn't have a name:

```python
if isinstance(column, ir.ColumnExpr):
    col_name = column.get_name()  # this will raise an error if the column isn't named
    if col_name != column:
        # raise a useful error message
    # now check for equality
```

cpcloud (changes requested):

ibis/expr/rules.py
gerrymanoim: This should still be a `com.IbisTypeError`?

ibis/expr/rules.py
gerrymanoim: This should still be a `com.IbisTypeError`?

cpcloud: Yep.

cpcloud:
@gerrymanoim Is this still WIP? Can you move it out of WIP if not?

gerrymanoim:
Yes - still WIP, I need to address your suggestions. Will try to do that this week.

ibis/expr/rules.py
```diff
@@ -424,11 +424,32 @@ def table(arg, *, schema=None, **kwargs):
...
+                return column
+            else:
+                raise com.IbisTypeError(
+                    f"Passed column is not a column in {table}"
+                )
+        except com.IbisError:
+            raise com.IbisTypeError(
+                f"Cannot get column {maybe_column} from {table}"
+            )
+    elif isinstance(column, (str, int)):
```
cpcloud, line 446: Can you move this up and do an early `return`?

ibis/expr/rules.py
```diff
@@ -424,11 +424,32 @@ def table(arg, *, schema=None, **kwargs):
...
+        maybe_column = column.get_name()
+        try:
+            if column.equals(table[maybe_column]):
+                return column
+            else:
+                raise com.IbisTypeError(
+                    f"Passed column is not a column in {table}"
+                )
+        except com.IbisError:
+            raise com.IbisTypeError(
```
cpcloud, line 443: Can we just let the `IbisError` propagate?

ibis/expr/rules.py
```diff
@@ -424,11 +424,32 @@ def table(arg, *, schema=None, **kwargs):
...
-        )
-    return getattr(this, name)[column]
+    if not hasattr(this, name):
+        raise com.IbisTypeError(f"Could not get table {name} from {this}")
+    table = getattr(this, name)
+    if isinstance(column, ir.AnyColumn):
+        if not column.has_name():
+            raise com.IbisTypeError(f"Passed column {column} has no name")
+
+        maybe_column = column.get_name()
```
cpcloud, line 434: Can you also let the exception propagate here? Then you don't need to `has_name`, you can simply get the name.

ibis/expr/rules.py
gerrymanoim: This causes large test failures.

gerrymanoim: https://github.com/ibis-project/ibis/runs/4562427238?check_suite_focus=true as an example.

ibis/expr/rules.py
gerrymanoim: See above.

cpcloud: Cool

ibis/expr/rules.py
cpcloud: Got it.

gerrymanoim: We could change the `suppress` call here to accept `IbisError` as well if we wanted to not let it propagate: https://github.com/ibis-project/ibis/blob/7da815cf725fa2001d1608922ca261ac2a74e788/ibis/expr/rules.py#L98. 

I think that would be cleaner, but I don't know of the original intent of only using `IbisTypeError` here.

ibis/expr/rules.py
```diff
@@ -424,11 +424,33 @@ def table(arg, *, schema=None, **kwargs):
...
+        return table[column]
+    elif isinstance(column, ir.AnyColumn):
+        if not column.has_name():
+            raise com.IbisTypeError(f"Passed column {column} has no name")
+
+        maybe_column = column.get_name()
+        try:
+            if column.equals(table[maybe_column]):
+                return column
+            else:
```
jreback, line 441: actually this else looks superflous, e.g. either the columns are equal (in which case it returns) or they are not or an error is raised.

so can you rm L441-444, pass on L446 ? and then you have the raise in L450

ibis/tests/expr/test_rules.py
```diff
@@ -290,6 +290,21 @@ def test_invalid_column_or_scalar(validator, value, expected):
...
+        (table, table.int_col, table.int_col),
+    ],
+)
+def test_valid_column_from(check_table, value, expected):
+    class Test:
+        table = check_table
+
+    validator = rlz.column_from("table")
+    assert validator(value, this=Test()).equals(expected)
+
```
jreback, line 306: any check for the raising?, what about the cases where the column doesn't have a name?

jreback (changes requested):

ibis/expr/rules.py
gerrymanoim: The raised error would be not helpful (`"value must be an int or str or AnyColumn, got "`)

ibis/tests/expr/test_rules.py
gerrymanoim: With a column that doesn't have a name you'd raise `IbisTypeError`. I can add invalid tests.

ibis/expr/rules.py
```diff
@@ -424,11 +424,33 @@ def table(arg, *, schema=None, **kwargs):
 
 @validator
 def column_from(name, column, *, this):
-    if not isinstance(column, (str, int)):
-        raise com.IbisTypeError(
-            f"value must be an int or str, got {type(column).__name__}"
-        )
-    return getattr(this, name)[column]
+    if not hasattr(this, name):
```
jreback, line 427: can you add some comments here about the goal here (for the function)

jreback (approved):
looks fine by me (request for some comments)

ibis/tests/expr/test_rules.py
```diff
@@ -290,6 +294,42 @@ def test_invalid_column_or_scalar(validator, value, expected):
...
+            IbisTypeError,
+        ),
+    ],
+)
+def test_invalid_column_from(check_table, validator, value, expected):
+    class Test:
+        table = check_table
+
+    with pytest.raises(expected):
+        validator(value, this=Test())
```
cpcloud, line 330: Small nit: can you construct `Test` outside of the `with` block?

ibis/tests/expr/test_rules.py
```diff
@@ -290,6 +294,42 @@ def test_invalid_column_or_scalar(validator, value, expected):
...
+            similar_table.int_col,
+            IbisTypeError,
+        ),
+    ],
+)
+def test_invalid_column_from(check_table, validator, value, expected):
+    class Test:
+        table = check_table
+
+    with pytest.raises(expected):
```
cpcloud, line 329: You can remove the `expected` parameter and `IbisTypeError` instead, since the exception type isn't varying.

I realize you probably copied the pattern from above, but no reason to keep that around (someone can fix that in another PR).

cpcloud (changes requested):
Small changes, then this is good to go.

cpcloud:
@gerrymanoim Can you rebase and `fixup` to a single commit with the title of the PR `feat: feat: allow column_of to take a column expression`?

gerrymanoim:
Squashed to a single one.

cpcloud:
@gerrymanoim Thanks!

It looks like your commit has `feat: feat:` as a prefix. You only need one `feat:` 🤣!

cpcloud (changes requested):
LGTM after removing one `feat:`!

gerrymanoim:
Whoops - copy/paste issue, sorry about that!

cpcloud (approved):
Merging! Thanks!

PR #3155, opened 2021-12-10
refactor(deprecated): add @util.deprecated and util.warn_deprecated

saulpw:


ibis/util.py
```diff
@@ -355,3 +356,39 @@ def flatten_iterable(iterable):
...
+
+    import traceback
+
+    warnlines.append('Deprecated call traceback (most recent last):')
+    for f in traceback.extract_stack(limit=6)[:-ignoreframes]:
+        warnlines.append(f'{f.filename} at line {f.lineno} in {f.name}')
+    warnings.warn('\n'.join(warnlines), DeprecationWarning)
+
+
+def deprecated(ver='', instead=''):
```
cpcloud, line 382: Are these defaults meaningful? If I decorate without passing them, do I still get a useful error message?

I suspect that we should make these keyword-only arguments (and rename `ver` to `version`, setting it to `None` by default as well) + making `instead` a required keyword-only argument so that it's clear at the call site what the arguments mean. Requiring an `instead` also forces developers to document what to do when the warning is shown.

ibis/util.py
```diff
@@ -355,3 +356,39 @@ def flatten_iterable(iterable):
...
+    msg = f'"{name}" is deprecated'
+    if ver:
+        msg += f' since v{ver}'
+
+    if instead:
+        msg += f'; use {instead}.'
+
+    warnlines.append(msg)
+
+    import traceback
```
cpcloud, line 374: Can you import this at the top of the module?

ibis/util.py
```diff
@@ -355,3 +356,39 @@ def flatten_iterable(iterable):
...
+
+    warnlines.append('Deprecated call traceback (most recent last):')
+    for f in traceback.extract_stack(limit=6)[:-ignoreframes]:
+        warnlines.append(f'{f.filename} at line {f.lineno} in {f.name}')
+    warnings.warn('\n'.join(warnlines), DeprecationWarning)
```
cpcloud, line 379: `DeprecationWarning` isn't shown by default (https://docs.python.org/3/library/exceptions.html#DeprecationWarning), unless you're in the `__main__` module, which isn't the primary way ibis is used. We have to use something else instead, and I think `FutureWarning` is probably the okay-est type for that :)

ibis/util.py
```diff
@@ -355,3 +356,39 @@ def flatten_iterable(iterable):
...
+
+    if instead:
+        msg += f'; use {instead}.'
+
+    warnlines.append(msg)
+
+    import traceback
+
+    warnlines.append('Deprecated call traceback (most recent last):')
+    for f in traceback.extract_stack(limit=6)[:-ignoreframes]:
```
cpcloud, line 377: This hardcoded 6 feels a bit arbitrary, any particular reason why 6?

ibis/util.py
```diff
@@ -355,3 +356,39 @@ def flatten_iterable(iterable):
             yield from flatten_iterable(item)
         else:
             yield item
+
+
+def warn_deprecated(name, ver='', instead='', ignoreframes=1):
```
cpcloud, line 361: I think `ver` (`version` ideally) and `instead` should be required keyword arguments similar to the decorator.

cpcloud (changes requested):

ibis/util.py
saulpw: `warn_deprecated` does compose a meaningful message if those arguments are omitted.  I can make `instead` be required and keyword-only.  I prefer a default of empty string for `version` instead of `None` (for typing and clarity reasons, since `''` isn't a reasonable version anyway) but am happy to make that change too.

ibis/util.py
saulpw: Probably better would be `ignoredframes+N`, where 1 <= N <= 5.  A full stacktrace is unwieldy, but it can be helpful to get more context than just 1 frame would provide. I would choose N=3 as a simple heuristic.  N=1 is also acceptable though if you want to keep logging noise to a minimum.

ibis/util.py
saulpw: One issue with always requiring the keywords is that the maximum `instead` that can fit on a single 79-character `@deprecated` line is 32 chars (without keywords can be +16 chars).

ibis/util.py
saulpw: I'm thinking of removing the stacktrace entirely and just passing `ignoreframes+1` to `warnings.warn`.

cpcloud:
@saulpw Can you rebase and fixup/squash? Can merge shortly thereafter.

ibis/backends/base/__init__.py
```diff
@@ -304,42 +292,26 @@ def list_tables(self, like: str = None, database: str = None) -> list[str]:
...
         """
-        warnings.warn(
-            '`client.exists_table(name)` is deprecated, and will be '
-            'removed in a future version of Ibis. Use '
-            '`name in client.list_tables()` instead.',
-            FutureWarning,
-        )
         return len(self.list_tables(like=name, database=database)) > 0
 
     # @abc.abstractmethod
```
cpcloud, line 302: Maybe kill this line.

cpcloud:
@saulpw Looks like a failing lint check still: https://github.com/ibis-project/ibis/runs/4782091015?check_suite_focus=true#step:5:1259

cpcloud (approved):

cpcloud:
Thanks @saulpw!

PR #3158, opened 2021-12-12
ci: move to app generated tokens

cpcloud:
This PR moves us off of specific PATs from my GitHub account
to a tiny GitHub app used to generate tokens, that any admin of `ibis-project`
can manage.

Additionally, tokens used are emphemeral for every run.

It's still a good idea to rotate the private key used to sign access token
requests on a quarterly (or more frequent basis). Currently GitHub doesn't offer a way to
automate the expiration of private keys, so we'll need to rotate those ourselves.

I've included a job that runs every 3 months to create an issue telling us to rotate the key.


.github/workflows/create-rotate-key-issue.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: tibdex/github-app-token@v1
+        id: generate_token
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: add condalock comment
```
kszucs, line 22: Step name looks wrong

kszucs (approved):
Just a nit otherwise it LGTM

.rgignore
```diff
@@ -0,0 +1 @@
+!.github
```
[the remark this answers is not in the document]
cpcloud: it is, just wanted to avoid a one line change in a single PR

.github/workflows/create-rotate-key-issue.yml
cpcloud: fixed

PR #3167, opened 2021-12-15
refactor: remove pkg_resources usage in favor of importlib.metadata

cpcloud:
This PR removes our use of `entry_points` (and therefore `pkg_resources`),
which doesn't seem to be necessary.

`pkg_resources` adds a huge amount of import time (100ms+) to `import ibis`
relative to what were using it for.

`importlib.metadata` (and `importlib_metadata` for Python 3.7) suffices here and is fast.

### DIFF
--- a/ibis/__init__.py
+++ b/ibis/__init__.py
@@ -90,7 +90,8 @@ def __getattr__(name: str) -> BaseBackend:
 
     if len(entry_points) > 1:
         raise RuntimeError(
-            f"{len(entry_points)} packages found for backend '{name}'. "
+            f"{len(entry_points)} packages found for backend '{name}': "
+            f"{entry_points}\n"
             "There should be only one, please uninstall the unused packages "
             "and just leave the one that needs to be used."
         )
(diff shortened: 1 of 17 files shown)


saulpw (commented):
Sweet, nice find @cpcloud!

ibis/__init__.py
```diff
@@ -76,23 +72,33 @@ def __getattr__(name: str) -> BaseBackend:
...
-    )
-    if len(entry_points) == 0:
+    entry_points = [
+        entry_point
+        for entry_point in importlib_metadata.entry_points().get(
+            "ibis.backends",
+            (),
+        )
+        if entry_point.name == name
+    ]
```
gforsyth, line 82: This seems to be double-counting all the backends for a reason I don't understand. It would be a bit gross but if we list(set( it then it should work

cpcloud: I'm using the `importlib.metadata.distribution` API now to avoid picking up multiple names. I think this shows up as both `ibis-framework` _and_ `ibis` when installed with `setuptools`.

kszucs (approved):
LGTM

PR #3170, opened 2021-12-16
refactor: remove dead pandas code for handling windows

### LINKED ISSUE #2438: Clean up potentially unreachable code in Window.agg (Pandas backend)
There is logic in `Window.agg` (in the Pandas backend) that I believe is unreachable:
https://github.com/ibis-project/ibis/blob/master/ibis/pandas/aggcontext.py#L459-L483

If `group_by` and `order_by` are both `None`, then a `Window` aggcontext **should not exist in the first place** because of this upstream logic that creates a `Summarize` aggcontext instead of a `Window` aggcontext:
https://github.com/ibis-project/ibis/blob/4fc17ab801b8d1b0b02a25551b4753247d837dfd/ibis/pandas/execution/window.py#L132-L133

It is definitely not reached by the tests, so possibly contains bugs.

We need to:
1. Determine whether it really is unreachable or not
2. If reachable, add tests for it and possibly fix broken logic
3. If unreachable, remove

cpcloud:
Closes #2438

### DIFF
--- a/ibis/backends/pandas/aggcontext.py
+++ b/ibis/backends/pandas/aggcontext.py
@@ -574,121 +574,94 @@ class Window(AggregationContext):
         group_by = self.group_by
         order_by = self.order_by
 
-        # if we don't have a grouping key, just call into pandas
-        if not group_by and not order_by:
-            # the result of calling .rolling(...) in pandas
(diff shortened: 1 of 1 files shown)


jreback (approved):
lgtm!

PR #3181, opened 2021-12-22
ci: clean up issue assignment job

cpcloud:
This PR removes `assign.yml` which currently doesn't seem to serve a purpose
other than to eat up CI time.

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -1,14 +1,14 @@
-name: Assign
+name: Assign Issue
 on:
   issue_comment:
-    types: created
-
+    types:
+      - created
 jobs:
-  one:
+  assign_issue:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


cpcloud:
This jobs does have an effect, but only when someone writes `<slash> take`. No one appears to be doing that, so there's no reason to keep the job around.

jreback:
hmm this is how contributors take issues.....

cpcloud:
I haven't seen anyone use it for a while. Maybe we just need to publicize it better?

jreback:
> I haven't seen anyone use it for a while. Maybe we just need to publicize it better?

yeah i would add to the docs

cpcloud:
> > I haven't seen anyone use it for a while. Maybe we just need to publicize it better?
> 
> yeah i would add to the docs

Turns out it was already there, I updated a bit.

jreback (approved):

PR #3185, opened 2021-12-22
feat: support of arrays and tuples for clickhouse

hikitani:
Added support arrays for clickhouse.

Implemented next operations:
* ArrayIndex
* ArrayConcat
* ArrayRepeat
* ArraySlice

### DIFF
--- a/ibis/backends/clickhouse/registry.py
+++ b/ibis/backends/clickhouse/registry.py
@@ -74,6 +74,58 @@ def _fixed_arity(func_name, arity):
     return formatter
 
 
+def _array_index_op(translator, expr):
+    op = expr.op()
+
+    arr = op.args[0]
+    idx = op.args[1]
+
+    arr_ = translator.translate(arr)
+    idx_ = _parenthesize(translator, idx)
+
+    correct_idx = f'if({idx_} >= 0, {idx_} + 1, {idx_})'
+
+    return f'arrayElement({arr_}, {correct_idx})'
+
+
+def _array_repeat_op(translator, expr):
+    op = expr.op()
+    arr, times = op.args
+
+    arr_ = _parenthesize(translator, arr)
+    times_ = _parenthesize(translator, times)
+
+    select = 'arrayFlatten(groupArray(arr))'
+    from_ = f'(select {arr_} as arr from system.numbers limit {times_})'
+    return f'(select {select} from {from_})'
+
+
+def _array_slice_op(translator, expr):
+    op = expr.op()
+    arg, start, stop = op.args
+
+    start_ = _parenthesize(translator, start)
+    arg_ = translator.translate(arg)
+
+    start_correct_ = f'if({start_} < 0, {start_}, {start_} + 1)'
+
+    if stop is not None:
+        stop_ = _parenthesize(translator, stop)
+
+        cast_arg_ = f'if({arg_} = [], CAST({arg_} AS Array(UInt8)), {arg_})'
+        neg_start_ = f'(arrayCount({cast_arg_}) + {start_})'
+        diff_fmt = f'greatest(-0, {stop_} - {{}})'.format
+
+        length_ = (
+            f'if({stop_} < 0, {stop_}, '
+            + f'if({start_} < 0, {diff_fmt(neg_start_)}, {diff_fmt(start_)}))'
+        )
+
+        return f'arraySlice({arg_}, {start_correct_}, {length_})'
+
+    return f'arraySlice({arg_}, {start_correct_})'
+
+
 def _agg(func):
     def formatter(translator, expr):
         return _aggregate(translator, func, *expr.op().args)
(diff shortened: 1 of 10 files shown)


ibis/backends/clickhouse/registry.py
```diff
@@ -74,6 +74,64 @@ def formatter(translator, expr):
...
+        stop_ = _parenthesize(translator, stop)
+
+        diff = f'({stop_} - {start_})'
+        length_ = f'''
+            if(
+                {stop_} < 0,
+                {stop_},
+                if({diff} < 0, 0, {diff})
+            )
+        '''
```
cpcloud, line 128:
```suggestion
        length_ = f"if({stop_} < 0, {stop_}, greatest(0, {diff}))"
```

cpcloud (changes requested):
Thanks! Looking good so far, just a few comments.

hikitani:
Negative slice indices are not supported

ibis/backends/clickhouse/registry.py
```diff
@@ -74,6 +74,63 @@ def formatter(translator, expr):
...
+    times_ = _parenthesize(translator, times)
+
+    return f'''
+        (select arrayFlatten(groupArray(arr)) from
+        (
+            select {arr_} as arr
+            from system.numbers
+            limit {times_}
+        ))
+    '''
```
cpcloud, line 105:
```suggestion
    select = 'arrayFlatten(groupArray(arr))'
    from_ = f'(select {arr_} as arr from system.numbers limit {times_})'
    return f'(select {select} from {from_})'
```

ibis/backends/clickhouse/registry.py
```diff
@@ -74,6 +74,63 @@ def formatter(translator, expr):
...
+
+
+def _array_slice_op(translator, expr):
+    op = expr.op()
+    arg, start, stop = op.args
+
+    start_ = _parenthesize(translator, start)
+    arg_ = translator.translate(arg)
+
+    start_correct_ = f'if({start_} < 0, {start_}, {start_} + 1)'
```
cpcloud, line 115: What are the semantics when `start_` is negative?

ibis/backends/clickhouse/registry.py
```diff
@@ -74,6 +74,63 @@ def formatter(translator, expr):
...
+    if stop is not None:
+        stop_ = _parenthesize(translator, stop)
+
+        cast_arg_ = f'if({arg_} = [], CAST({arg_} AS Array(UInt8)), {arg_})'
+        neg_start_ = f'(arrayCount({cast_arg_}) + {start_})'
+        diff_fmt = f'greatest(-0, {stop_} - {{}})'.format
+
+        length_ = (
+            f'if({stop_} < 0, {stop_}, '
+            + f'if({start_} < 0, {diff_fmt(neg_start_)}, {diff_fmt(start_)}))'
```
cpcloud, line 126:
```suggestion
            f'if({start_} < 0, {diff_fmt(neg_start_)}, {diff_fmt(start_)}))'
```

ibis/backends/clickhouse/registry.py
```diff
@@ -74,6 +74,63 @@ def formatter(translator, expr):
...
+    arg_ = translator.translate(arg)
+
+    start_correct_ = f'if({start_} < 0, {start_}, {start_} + 1)'
+
+    if stop is not None:
+        stop_ = _parenthesize(translator, stop)
+
+        cast_arg_ = f'if({arg_} = [], CAST({arg_} AS Array(UInt8)), {arg_})'
+        neg_start_ = f'(arrayCount({cast_arg_}) + {start_})'
+        diff_fmt = f'greatest(-0, {stop_} - {{}})'.format
```
cpcloud, line 122:
```suggestion
        diff_fmt = f'greatest(0, {stop_} - {{}})'.format
```

cpcloud (changes requested):

ibis/backends/clickhouse/registry.py
hikitani: When `start_` is negative, `arraySlice` works as a python slice with only the "start" argument, for example, if you define `arr[-2:]` in python, then for CH it will be `arraySlice(arr, -2)`.

cpcloud: Got it, thanks.

cpcloud:
Looking at https://github.com/ibis-project/ibis/runs/4656549237 I see

image

I think you need to remove the `xfail` on the array slicing tests since you've removed `ValueError` in the API function.

cpcloud:
@hikitani Can you rebase and fixup to a single commit with the title

```
feat: support of arrays and tuples for clickhouse
```

cpcloud:
@hikitani Thanks for all your work on this.

This is a breaking change due to the removal of the exception in the top level API.

We're going to release 2.1.0 next week, without this change, but we'll merge this immediately after that release and your PR will be in the subsequent release (3.0.0).

The timeline for that isn't completely locked down, but we'd like to cut another release by the end of January/beginning of February with a good chunk of breaking changes including this one.

cpcloud (approved):

cpcloud:
Thanks @hikitani! Looking forward to this in 3.0.0!

PR #3188, opened 2021-12-22
feat(api): add order_by alias for sort_by

saulpw:
@gforsyth and I experienced minor friction as we expected `order_by` to exist in the expr api, for parity with SQL `ORDER BY` . An alias seems reasonable.

This PR also includes a minor internal rename of the `filter()` function to avoid conflict with Python builtin `filter()`.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -4658,6 +4658,7 @@ _table_methods = {
     'anti_join': _regular_join_method('anti_join', 'anti'),
     'asof_join': asof_join,
     'sort_by': _table_sort_by,
+    'order_by': _table_sort_by,
     'to_array': _table_to_array,
     'union': _table_union,
     'intersect': _table_intersect,
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -4126,7 +4126,7 @@ def _regular_join_method(name, how, doc=None):
     return f
 
 
-def filter(table, predicates):
+def _filter(table, predicates):
     """
     Select rows from table based on boolean expressions
 
@@ -4641,7 +4641,7 @@ _table_methods = {
     'limit': _table_limit,
     'head': _head,
     'set_column': _table_set_column,
-    'filter': filter,
+    'filter': _filter,
     'materialize': _table_materialize,
     'mutate': mutate,
     'projection': projection,


ibis/expr/api.py
```diff
@@ -4126,7 +4126,7 @@ def f(self, other, predicates=()):
     return f
 
 
-def filter(table, predicates):
+def _filter(table, predicates):
```
jreback, line 4129: i don't think this is needed here

ibis/expr/api.py
```diff
@@ -4658,6 +4658,7 @@ def _rowid(self):
     'anti_join': _regular_join_method('anti_join', 'anti'),
     'asof_join': asof_join,
     'sort_by': _table_sort_by,
+    'order_by': _table_sort_by,
```
jreback, line 4661: -0.5 on this. duplicating names causes even more friction. no objection to adding some working to the doc-string though.

jreback (changes requested):

ibis/expr/api.py
saulpw: Maybe it would be preferable to rename `sort_by` to `order_by`, and deprecate `sort_by`?  Equivalent in the near-term and still cleans up the API in the longer term.

ibis/expr/api.py
saulpw: I can make this a separate PR if you want.  It's not "needed" but it's not a good idea in general to override a Python builtin--even confined to file scope, it prevent the builtin from being usable.

jreback: its not about a separate PR, what does this fix? sure its a builtin, but so what

ibis/expr/api.py
jreback: > Maybe it would be preferable to rename sort_by to order_by, and deprecate sort_by? Equivalent in the near-term and still cleans up the API in the longer term.

-1 on that. this is a long established api.

ibis/expr/api.py
cpcloud: It doesn't fix anything, but makes the use of filter inside the module less surprising, which is reasonable. I think this change is fine.

ibis/expr/api.py
cpcloud: I don't think there's a huge amount to be gained by duplicating (opens the Pandora's box of questions about why there are two) or by renaming (breaks a lot of existing code).

I would however like to see a dead simple mapping of SQL constructs to ibis apis in a quickstart. This would decrease friction without disruption. I'm all for disruption but I don't think this particular breaking change gives us much bang for our buck.

saulpw: SQL predates the Ibis API by 30+ years.  `order_by` is intuitive to data folks and obeys the [Principle of Least Surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).  We're trying to make API easier for larger numbers of people to use, and adding an alias like this is one small thing we can do.

cpcloud: That's fair. One option is to remove references to sort_by and deprecate it in 4.0. We've got a good chunk of breaking API changes slated for 3.0. How about a deprecation in 3.0, full removal in 4.0? I agree that order_by is what folks coming from SQL will immediately reach for.

kszucs: Just noting, that there are users coming from pandas as well where the term `sort` is used.

jayceslesar: Definitely quite a debate...whatever solution is selected to move forward, updating docs can never hurt, been scouring docs the past few weeks so will flag stuff when I see it....on another note would be nice to have examples of direct comparisons of SQL --> ibis in the docs with functions like `.diff` as well as `merge_asof`,  Maybe biased coming from a mostly time series perspective...

saulpw:
IBEP-2 will address the general direction and strategy for the Ibis API.  Closing this and will resubmit depending on how that turns out.

PR #3201, opened 2021-12-27
ci: stop running backend tests in default.nix

cpcloud:
We run the backends tests elsewher in CI, so we do not need to run them in the nix setup. Non backend tests are still run as a sanity check for the environment.


cpcloud:
@jreback This should shave some time off the `nix build`s in CI. Please review/approve at your earliest convenience.

jreback (approved):

PR #3217, opened 2021-12-30
ci: add codeql security analysis workflow

cpcloud:


kszucs:
Is it intentionally running on all dependencies under the `.poetry` directory?

cpcloud:
> Is it intentionally running on all of the dependencies under the `.poetry` directory?

Yeah, the point is to check for vulnerabilities in the full stack.

Since we have a large number of dependencies this takes quite a while to run, so I'll keep the cron schedule, but remove the `pull_request`/`push` triggers.

kszucs (approved):

PR #3222, opened 2021-12-31
ci(perf): improve caching in ci

cpcloud:
This PR uses two github actions to consolidate the ci caching of poetry dependencies as well as allow updates to the cache if anythin chagnes during the run

### DIFF
--- a/.github/workflows/check-setup-py.yml
+++ b/.github/workflows/check-setup-py.yml
@@ -78,9 +78,6 @@ jobs:
           requirement_files: requirements.txt
           custom_cache_key_element: check-setuptools-install-${{ steps.install_python.outputs.python-version }}
 
-      - name: upgrade pip
-        run: pip install -U pip
-
       - uses: syphar/restore-pip-download-cache@v1
         if: ${{ !fromJSON(steps.cache-virtualenv.outputs.cache-hit) }}
         with:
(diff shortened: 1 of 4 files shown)


jreback:
does this make much of a difference in total CI time per build?

cpcloud:
@jreback

> total CI time per build?

What does that mean? Pending-to-green duration? Pending-to-green per *workflow*?

jreback:
> @jreback
> 
> > total CI time per build?
> 
> What does that mean? Pending-to-green duration? Pending-to-green per _workflow_?

i mean does caching help like 10% on a build on average or much more?

cpcloud:
Caching in general reduces the build time by about a factor of two (so this is the duration of a build that updates the lockfile, adds a dependency etc).

I just ran a cached build, so we'll see in a bit how much this helps.

It should reduce the amount of time spent in the builds that weren't previously cached (dask extra builds, pyspark). I'd expect somewhere between 10-30% reduction so a few minutes per build, on average.

jreback:
> Caching in general reduces the build time by about a factor of two (so this is the duration of a build that updates the lockfile, adds a dependency etc).
> 
> I just ran a cached build, so we'll see in a bit how much this helps.
> 
> It should reduce the amount of time spent in the builds that weren't previously cached (dask extra builds, pyspark). I'd expect somewhere between 10-30% reduction so a few minutes per build, on average.

ok seems reasonable.

cpcloud:
End to end for all jobs takes about 16 minutes.

image

jreback (approved):

PR #3227, opened 2022-01-02
ci: upload workflow run data to GCS daily

cpcloud:
This PR adds a workflow to upload the previous day's workflow 
run information to a GCS bucket

### DIFF
--- a/.github/workflows/ci-data.yml
+++ b/.github/workflows/ci-data.yml
@@ -5,10 +5,6 @@ on:
     - cron: "0 1 * * *"
   workflow_dispatch:
 
-concurrency:
-  group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
-  cancel-in-progress: true
-
 jobs:
   gcs_insert:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-data.yml
+++ b/.github/workflows/ci-data.yml
@@ -1,0 +1,61 @@
+name: GCS Insert
+
+on:
+  schedule:
+    - cron: "0 1 * * *"
+  workflow_dispatch:
+
+concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-data.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: generate data
+        id: generate_data
+        run: |
+          set -euo pipefail
+
+          yesterday="${{ steps.set_date.outputs.yesterday }}"
+
+          [ -n "$yesterday" ] || exit 1
+
+          gh api --paginate "repos/ibis-project/ibis/actions/runs?per_page=100&created=${yesterday}" --jq '.workflow_runs[]' > workflows.json
```
kszucs, line 28: Does it use the `GITHUB_TOKEN` provided by github actions? In that case the rate limit is just 1000, so sleeping one second won't help much.

.github/workflows/ci-data.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - run: gcloud info
+
+      - name: copy to gcs
+        run: |
+          set -euo pipefail
+
+          yesterday="${{ steps.set_date.outputs.yesterday }}"
+
+          gsutil cp -Z workflows.json "gs://ibis-workflow-data/${yesterday}/workflows.json"
```
jreback, line 52: not a big deal but do we need daily? wouldn't it be better to have this run when the deps of this change?

cpcloud: Dependencies are unrelated here. I want to store all actions runs data.

.github/workflows/ci-data.yml
cpcloud: Fixed.

cpcloud:
@jreback can you approve?

jreback (approved):

PR #3243, opened 2022-01-06
ci: disable python 3.7 for datafusion

cpcloud:
Python 3.7 support is broken is datafusion. See https://github.com/datafusion-contrib/datafusion-python8.


jreback (approved):
lgtm

prob should drop 3.7 in next major release as well

PR #3249, opened 2022-01-07
fix(format): remove some newlines from formatted expr repr

saulpw:
This tightens up the format string returned by `repr(expr)`.  Some tests had to be adjusted as they depended on the literal output.

### DIFF
--- a/ibis/expr/format.py
+++ b/ibis/expr/format.py
@@ -113,9 +113,7 @@ class ExprFormatter:
         elif isinstance(what, ops.TableColumn):
             text = self._format_column(self.expr)
         elif isinstance(what, ops.Literal):
-            text = 'Literal[{}]\n  {}'.format(
-                self._get_type_display(), str(what.value)
-            )
+            text = f'Literal[{self._get_type_display()}] {what.value}'
         elif isinstance(what, ops.ScalarParameter):
             text = f'ScalarParameter[{self._get_type_display()}]'
         elif isinstance(what, ops.Node):
@@ -138,7 +136,7 @@ class ExprFormatter:
             # A hack to suppress printing out of a ref that is the result of
             # the top level expression
             refs = [
-                x + '\n' + y
+                f'{x}  {y}'
                 for x, y, op in alias_to_text
                 if not op.equals(what)
             ]
@@ -184,14 +182,14 @@ class ExprFormatter:
     def _format_table(self, expr):
         table = expr.op()
         # format the schema
-        rows = [f'name: {table.name}\nschema:']
+        rows = [f'{table.name}:']
         rows.extend(
             map('  {} : {}'.format, table.schema.names, table.schema.types)
         )
         opname = type(table).__name__
         type_display = self._get_type_display(expr)
         opline = f'{opname}[{type_display}]'
-        return '{}\n{}'.format(opline, self._indent('\n'.join(rows)))
+        return '{} {}'.format(opline, self._indent('\n'.join(rows)))
 
     def _format_column(self, expr):
         # HACK: if column is pulled from a Filter of another table, this parent
@@ -203,10 +201,9 @@ class ExprFormatter:
             self.memo.observe(parent, formatter=self._format_node)
 
         table_formatted = self.memo.get_alias(parent)
-        table_formatted = self._indent(table_formatted)
 
         type_display = self._get_type_display(self.expr)
-        return "Column[{}] '{}' from table\n{}".format(
+        return "Column[{}] '{!r}' from {}".format(
             type_display, col.name, table_formatted
         )
 
--- a/ibis/tests/expr/test_format.py
+++ b/ibis/tests/expr/test_format.py
@@ -8,12 +8,12 @@ def test_format_custom_expr():
         def _type_display(self):
             return 'my-custom'
 
-    op = ibis.literal(5).op()
+    op = ibis.literal(1234567890).op()
     expr = CustomExpr(op)
 
     result = repr(expr)
-    expected = 'Literal[my-custom]\n  5'
-    assert result == expected
+    assert 'my-custom' in result
+    assert '1234567890' in result
 
 
 def test_format_table_column(table):
@@ -197,23 +197,11 @@ def test_memoize_filtered_tables_in_join():
 
 
 def test_argument_repr_shows_name():
-    t = ibis.table([('a', 'int64')], name='t')
-    expr = t.a.nullif(2)
+    t = ibis.table([('fakecolname1', 'int64')], name='fakename2')
+    expr = t.fakecolname1.nullif(2)
     result = repr(expr)
-    expected = """\
-ref_0
-UnboundTable[table]
-  name: t
-  schema:
-    a : int64
-
-NullIf[int64*]
-  a = Column[int64*] 'a' from table
-    ref_0
-  null_if_expr:
-    Literal[int8]
-      2"""
-    assert result == expected
+    assert 'fakecolname1' in result
+    assert 'fakename2' in result
 
 
 def test_scalar_parameter_formatting():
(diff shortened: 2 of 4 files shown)


jreback:
this is a breaking change (not objecting to it but needs to be in next major release)

cpcloud:
@jreback Can you clarify what's breaking about it?

The repr is not in any way guaranteed to be stable across versions.

ibis/expr/format.py
```diff
@@ -138,7 +138,7 @@ def get_result(self):
             # A hack to suppress printing out of a ref that is the result of
             # the top level expression
             refs = [
-                x + '\n' + y
+                x + ' ' + y
```
cpcloud, line 141:
```suggestion
                f"{x} {y}"
```

ibis/expr/format.py
```diff
@@ -203,10 +203,10 @@ def _format_column(self, expr):
             self.memo.observe(parent, formatter=self._format_node)
 
         table_formatted = self.memo.get_alias(parent)
-        table_formatted = self._indent(table_formatted)
+        table_formatted = table_formatted
 
         type_display = self._get_type_display(self.expr)
-        return "Column[{}] '{}' from table\n{}".format(
+        return "Column[{}] '{}' from {}".format(
```
cpcloud, line 209:
```suggestion
        return "Column[{}] {!r} from {}".format(
```

ibis/expr/format.py
```diff
@@ -203,10 +203,10 @@ def _format_column(self, expr):
             self.memo.observe(parent, formatter=self._format_node)
 
         table_formatted = self.memo.get_alias(parent)
-        table_formatted = self._indent(table_formatted)
+        table_formatted = table_formatted
```
cpcloud, line 206:
```suggestion
```

No need to reassign a variable to itself :)

ibis/expr/format.py
```diff
@@ -113,7 +113,7 @@ def get_result(self):
         elif isinstance(what, ops.TableColumn):
             text = self._format_column(self.expr)
         elif isinstance(what, ops.Literal):
-            text = 'Literal[{}]\n  {}'.format(
+            text = 'Literal[{}] {}'.format(
                 self._get_type_display(), str(what.value)
             )
```
cpcloud, line 118:
```suggestion
            text = 'Literal[{self._get_type_display()}] {what.value}'
```

Fine to keep as is if this puts you over the maximum line length

cpcloud (changes requested):

ibis/expr/format.py
saulpw: Will the `repr` of the column name ever be different from its `str`?

saulpw:
BTW the tests that are failing are checking that the `repr` exactly matches explicit output, as a kind of proxy for "did this expression get encoded properly".   If we want to test validity of expressions, I think it should be at a bit of a higher level that doesn't depend on an unstable result like `repr`.  Open to ideas for how to do this, but unless something jumps out, for now I think we should just remove these failing repr tests.

cpcloud:
How about just keeping the `repr(expr)` and removing the `==` assertion?

ibis/expr/format.py
cpcloud: I suppose that's possible, but very unlikely.

jreback:
> @jreback Can you clarify what's breaking about it?
> 
> The repr is not in any way guaranteed to be stable across versions.

my issue here is this a user visible change and should clearly be announced, and absolutely we should test these. these *should* be guaranteed.

cpcloud:
> > @jreback Can you clarify what's breaking about it?
> > The repr is not in any way guaranteed to be stable across versions.
> 
> my issue here is this a user visible change and should clearly be announced

What about this PR indicates this wouldn't be announced?

> and absolutely we should test these.

No one is arguing that we shouldn't test `repr` at all, but testing the exact `repr` is not viable going forward. At best we can do `thing in repr(expr)`.

> these _should_ be guaranteed.

Says who?

[Python itself breaks repr between minor releases](https://www.python.org/dev/peps/pep-0606/#python-3-7) despite [PEP 663 claiming otherwise](https://www.python.org/dev/peps/pep-0663/#backwards-compatibility).

There's plenty of reason and precedent to not guarantee the repr between major releases. We're not a formatting library.

jreback:
just because some people don't do it doesn't mean that should pick and choose what is an api break

my point is this should be in a major release and have a note saying that it changed 

i agree testing in repr is prob enough (not quibbling about very minor formatting things)

cpcloud:
@saulpw Looks like there's still a failing format check, you need to run `black .` 😬

cpcloud:
Sorry, wrong PR!

cpcloud (approved):

cpcloud:
Thanks @saulpw!

ibis/tests/expr/test_format.py
```diff
@@ -197,23 +197,11 @@ def test_memoize_filtered_tables_in_join():
 
 
 def test_argument_repr_shows_name():
-    t = ibis.table([('a', 'int64')], name='t')
-    expr = t.a.nullif(2)
+    t = ibis.table([('fakecolname1', 'int64')], name='fakename2')
+    expr = t.fakecolname1.nullif(2)
     result = repr(expr)
```
kszucs, line 202: @cpcloud @saulpw there might be side effects of this PR:
- strange indentation of the table name
- 2 single colons around the column names
- [perhaps] a missing newline from table references

After:

```
ref_0  UnboundTable[table]   t:
    col : int64

Selection[table]
  table:
    Table: ref_0
  selections:
    fakealias1 = Column[int64*] ''col'' from ref_0
    fakealias2 = Column[int64*] ''col'' from ref_0
```

Before:

```
ref_0
UnboundTable[table]
  name: t
  schema:
    col : int64

Selection[table]
  table:
    Table: ref_0
  selections:
    fakealias1 = Column[int64*] 'col' from table
      ref_0
    fakealias2 = Column[int64*] 'col' from table
      ref_0
```

Could we keep explicit output tests at least in `test_format.py`?

saulpw: Hi @kszucs...I feel pretty strongly that we shouldn't test functionality that's not part of the contract.  I made a concession for checking for fragments in `test_format.py` because it does seem useful to make sure the repr has particular meaningful information in it, though I wish it wasn't dependent on the format of that information even in those cases.

I agree that the 2 single quotes around the column name is a negative side effect and I'll fix that (not quite sure how it happened; this should have been almost entirely whitespace).  The removal of newline from table references was intentional, to make the whole repr more compact.  I'm afraid I don't see the "strange indentation" you're talking about though; do you just mean that the table name is on the same line as the table type, now?  That was also intentional, and for the same goal of compactness.

cpcloud: I think string indentation may be `col : int64` being indented 4 spaces instead of 2 which is what the rest of the tree uses.

I think instead of `UnboundTable[table]` which has a useless `table` in there, we should just put the table name, like `UnboundTable[name={name}, ref=r{ref_num}]`. This gets rid of the strange `t` and makes the references front and center without adding a newline.

cpcloud: Here's an example:

```
r0
UnboundTable[name=t, ref=r0]
  col: int64

Selection[table]
  table:
    ref: r0
  selections:
    fakealias1 = Column[int64*] <- r0.col
    fakealias2 = Column[int64*] <- r0.col
```

cpcloud: We can get of the first `r0` too, I just haven't been able to figure out exactly where that's being placed into the final output.

cpcloud: Alright, found it:

```
UnboundTable[name=t, ref=r0]
  col: int64

Selection[table]
  table:
    ref: r0
  selections:
    fakealias1 = Column[int64*] <- r0.col
    fakealias2 = Column[int64*] <- r0.col
```

saulpw: Happy to do (or have someone else do) any/all of this. I have several more things I think should be compactified in the repr actually (like ideally, any arg with only one value should have the value on the same line as the arg name).  But I futzed with it and couldn't get it all working like I wanted in a short amount of time and just decided that 30% less vertical whitespace was a solid step, hence this PR.

cpcloud: Cool, I'll throw up what I've got so far

saulpw: Also, and this is completely just personal preference, I'm not that much of a fan of `<-` syntax.  I've always had trouble with languages that use it for assignment (it just doesn't look "right" to me), and it's ungreppable (not sure how relevant that is in this case though).

PR #3250, opened 2022-01-07
ci: shave some time off the data download in ci

cpcloud:


cpcloud:
This appears to make things *slower* :(

PR #3261, opened 2022-01-09
fix: correlated subquery with join generates incorrect SQL

### LINKED ISSUE #3163: bug: correlated subquery with join generates incorrect SQL
This Ibis expression (using the TPC-H database; a simplified version of TPC-H query 2) appears to be valid, but does not generate correct SQL:

```
    part = con.table("part")
    supplier = con.table("supplier")
    partsupp = con.table("partsupp")

    q = part.join(partsupp, part.P_PARTKEY == partsupp.PS_PARTKEY)
    q = q[
        part.P_PARTKEY,
        partsupp.PS_SUPPLYCOST,
    ]
    subq = partsupp.join(supplier, supplier.S_SUPPKEY == partsupp.PS_SUPPKEY)
    subq = subq.materialize()
    subq = subq[subq.PS_PARTKEY == q.P_PARTKEY]

    q = q[q.PS_SUPPLYCOST == subq.PS_SUPPLYCOST.min()]
```

### Expected Result

```sql
SELECT t0."P_PARTKEY",
       t0."PS_SUPPLYCOST"
   FROM base.part AS t1
   JOIN base.partsupp AS t2 ON t1."P_PARTKEY" = t2."PS_PARTKEY") AS t0
WHERE t0."PS_SUPPLYCOST" =
    (SELECT min(t3."PS_SUPPLYCOST") AS MIN
        FROM base.partsupp AS t2
        JOIN base.supplier AS t6 ON t6."S_SUPPKEY" = t2."PS_SUPPKEY") AS t3
     WHERE t3."PS_PARTKEY" = "P_PARTKEY"
```

### Actual Result

The generated SQL is:

```sql
SELECT t0."P_PARTKEY",
       t0."PS_SUPPLYCOST"
FROM
  (SELECT t1."P_PARTKEY" AS "P_PARTKEY",
          t2."PS_SUPPLYCOST" AS "PS_SUPPLYCOST"
   FROM base.part AS t1
   JOIN base.partsupp AS t2 ON t1."P_PARTKEY" = t2."PS_PARTKEY") AS t0
WHERE t0."PS_SUPPLYCOST" =
    (SELECT min(t3."PS_SUPPLYCOST") AS MIN
     FROM
       (SELECT [...]
        FROM base.partsupp AS t2
        JOIN base.supplier AS t6 ON t6."S_SUPPKEY" = t2."PS_SUPPKEY") AS t3,
       (SELECT t1."P_PARTKEY" AS "P_PARTKEY",
               t2."PS_SUPPLYCOST" AS "PS_SUPPLYCOST"
        FROM base.part AS t1
        JOIN base.partsupp AS t2 ON t1."P_PARTKEY" = t2."PS_PARTKEY")
     WHERE t3."PS_PARTKEY" = "P_PARTKEY")
```

The subquery above does an extra join with a copy of the outer query, whereas it should just reference the existing value in the outer query.

cpcloud:
This PR addresses #3163.

It turns out that we weren't examining parent contexts when looking up tables
in the SQLAlchemy backend. Without this, a table subexpression would get
recompiled instead of looked up. We have to look into the outer scope since the
referenced expression may not be available in the current context, in this case
precisely because of the correlated subquery.

Closes #3163.

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -77,7 +77,7 @@ def _varargs_call(sa_func, t, expr):
 
 
 def get_sqla_table(ctx, table):
-    if ctx.has_ref(table):
+    if ctx.has_ref(table, parent_contexts=True):
         ctx_level = ctx
         sa_table = ctx_level.get_ref(table)
         while sa_table is None and ctx_level.parent is not ctx_level:
--- a/ibis/tests/sql/test_sqlalchemy.py
+++ b/ibis/tests/sql/test_sqlalchemy.py
@@ -622,6 +622,87 @@ class TestSQLAlchemySelect(unittest.TestCase, ExprTestCases):
 
         self._compare_sqla(expr, ex)
 
+    def test_where_correlated_subquery_with_join(self):
+        # GH3163
+        # ibis code
+        part = ibis.table([("p_partkey", "int64")], name="part")
+        partsupp = ibis.table(
+            [
+                ("ps_partkey", "int64"),
+                ("ps_supplycost", "float64"),
+                ("ps_suppkey", "int64"),
+            ],
+            name="partsupp",
+        )
+        supplier = ibis.table([("s_suppkey", "int64")], name="supplier")
+
+        q = part.join(partsupp, part.p_partkey == partsupp.ps_partkey)
+        q = q[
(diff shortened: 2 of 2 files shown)


gforsyth (approved):
Nice job tracking that down!  What a hilariously frustrating tinydiff.

PR #3267, opened 2022-01-10
ci: fix pyspark caching, give cache keys a name when deps are empty

cpcloud:


.github/workflows/ibis-backends.yml
```diff
@@ -111,7 +111,7 @@ jobs:
         uses: actions/upload-artifact@v2
         if: success() || failure()
         with:
-          name: ${{ matrix.backend.name }}-${{ join(matrix.backend.deps, '-') }}-${{ matrix.os }}-${{ matrix.python-version }}
+          name: ${{ matrix.backend.name }}-${{ join(matrix.backend.deps || fromJSON('["no-deps"]'), '-') }}-${{ matrix.os }}-${{ matrix.python-version }}
```
kszucs, line 114: Why is `fromJSON` needed?

cpcloud: Because I can't be bothered to remember whether `null` propagates in GitHub Actions minilanguage 🤣

kszucs (approved):

PR #3273, opened 2022-01-10
skip backend CI if only docs are touched

gforsyth:
This includes a useless commit that I will remove if this works.

Attempting to gate the backend CI such that it only runs if files not in `docs/**` have been touched (without preventing the docs from building and other actions from firing)


cpcloud:
I wonder if this will work with check suite constraints. Right now we require all PRs to pass the backend suite, but I think we'll need to turn that off for this to work.

I'd be okay with that.

cpcloud:
We may have to make a few other procedural adjustments but that should be fine.

cpcloud:
Ah, it looks like the per-job `path` might be the ticket!

gforsyth:
I'll reopen this after some refactoring so I'm not burning CI minutes

PR #3282, opened 2022-01-11
chore(deps): remove maturin from development dependencies

cpcloud:
This PR removes maturin from ibis's development dependencies since is not necessary for development of ibis, only for the development of datafusion-python itself.


kszucs (approved):
LGTM, but the poetry lockfile conflicts.

PR #3288, opened 2022-01-12
ci: test latest postgres backend deps with python 3.7

cpcloud:
This PR brings back an additional job which is running the postgres backend tests with 3.7 agains the latest set of dependencies for that backend

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -122,7 +122,7 @@ jobs:
           path: junit.xml
 
   test_postgres:
-    name: PostgreSQL deps-${{ (matrix.deps && '') || 'un' }}bounded python-${{ matrix.python-version }}
+    name: PostgreSQL deps-${{ (matrix.deps && 'bounded') || 'unbounded' }} python-${{ matrix.python-version }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ibis-backends.yml
```diff
@@ -130,6 +130,7 @@ jobs:
         python-version:
           - "3.7"
           - "3.9"
+        deps: null
```
kszucs, line 133: So we should have three variations here, one for 3.9 and two for 3.9?

cpcloud: I guess you mean by the second 3.9 3.7 :)

3.7 runs with the lower bounds and unrestricted bounds, and 3.9 with just the unrestricted

kszucs:
Could you please assign different titles for the two 3.7 jobs?

cpcloud:
> Could you please assign different titles for the two 3.7 jobs?

Roger, on it.

kszucs:
@cpcloud We may want to remove the "ubuntu-latest" part as well to make them more readable.

cpcloud:
> @cpcloud We may want to remove the "ubuntu-latest" part as well to make them more readable. Either in this PR or in a follow-up.

I can remove them here.

jreback (approved):

PR #3289, opened 2022-01-12
docs: use makefile on CI to fix sphinx dates

### LINKED ISSUE #3280: docs: use Makefile to build docs so that SOURCE_DATE_EPOCH is being used
We're not using the `Makefile` to build our Sphinx docs, so we don't pick up the fix in #3264 just yet.

We're not using the `Makefile` likely because we want to control the output directory, but that can be set with `make -C docs BUILDDIR="web/docs" html` and then adjusting `docs/mkdocs.yml` accordingly:

```yaml
nav:
  - Documentation: /docs/html
```

gforsyth:
Resolves #3280


docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -W -T
```
cpcloud, line 5: Can you also set `-j auto` so we can get some sweet doc build parallelism?

gforsyth: Oh Yeah!

cpcloud (approved):

PR #3299, opened 2022-01-12
fix: allow major version changes in CalVer dependencies

gerrymanoim:
Major version bumps have no real meaning under [calver](https://calver.org/). We should just be specifying a min.


cpcloud (approved):
SGTM. Doesn't make sense to apply semver rules to calver deps.

cpcloud:
Thanks @gerrymanoim!

PR #3304, opened 2022-01-13
chore: drop Python 3.7 support

cpcloud:
This PR drops Python 3.7 support following the NumPy cadence outlined here: https://numpy.org/neps/nep-0029-deprecation_policy.html


pyproject.toml
```diff
@@ -59,7 +59,7 @@ click = "^8.0.1"
 commitizen = "^2.20.3"
 flake8 = "^4.0.0"
 ipykernel = "^6"
-ipython = "^7.27.0"
+ipython = ">=7.27.0,<9"
```
kszucs, line 62: `<9` is intentional here?

cpcloud: Yep, ipython 8 can be used only with python 3.8

cpcloud:
Looks like dropping 3.7 will also require dropping support for PySpark 2.x, and bumping the psycopg2 lower bound to 2.8.4

jreback (approved):
looks fine, a failing nix check :-<

PR #3309, opened 2022-01-13
bug: fix postgres udfs and reenable ci tests

### LINKED ISSUE #3308: bug: 'PostgreSQLCompiler' has no attribute 'add_operation'
As turned out during my local refactor, the postgres udf tests are not running, there should be a `test_udf.py` [here](https://github.com/ibis-project/ibis/runs/4808027717?check_suite_focus=true#step:13:43). 

There is a valid error, since the `.add_operation()` method has been moved to the `ExpressionTranslator` base class:

```py
        def _translate_udf(t, expr):
            func_obj = sa.func
            if schema is not None:
                func_obj = getattr(func_obj, schema)
            func_obj = getattr(func_obj, name)

            sa_args = [t.translate(arg) for arg in expr.op().args]

            return func_obj(*sa_args)

>       PostgreSQLCompiler.add_operation(udf_node, _translate_udf)
E       AttributeError: type object 'PostgreSQLCompiler' has no attribute 'add_operation'

ibis/backends/postgres/udf.py:124: AttributeError
```

cpcloud:
Closes #3308

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -194,7 +194,7 @@ jobs:
       - name: run tests
         env:
           PYTEST_BACKENDS: postgres
-        run: ./ci/run_tests.sh -m "not udf"
+        run: ./ci/run_tests.sh
 
       - name: publish test report
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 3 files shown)


kszucs (approved):
LGTM, let's wait for the CI results.

PR #3314, opened 2022-01-14
ci(docs): add dummy jobs that return in place of skipped required jobs

gforsyth:
These intentionally collide with the jobs in `ibis-main.yml` and
`ibis-backends.yml` and run on the inverse of the path skipping logic
present in those two files.

This should allow CI to finish and all jobs to return, even if they're
skipped.

This is an ugly hack, but is the supported one for now:
https://github.com/github/docs/commit/4364076e0fb56c2579ae90cd048939eaa2c18954#diff-3052509b0e9a9ca90d0b911cc12c333506e652dbac7c7d5bd41d462be5c07dfdR48

Rendered docs for skipping procedure: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks


cpcloud (approved):
image

PR #3315, opened 2022-01-14
feat(api): Implement a merge_asof

### LINKED ISSUE #3295: bug: merge asof raises an error when the predicate column(s) are overlapping
Not sure how this is happening but I keep getting `ibis.common.exceptions.RelationError: Joined tables have overlapping names:`.

Both dataframes have 3 columns, where one column in each is `timestamp`. Unsure how that is happening, and unsure how [this test](https://github.com/ibis-project/ibis/pull/1162/files#diff-dd3a3bc8a96180b07abe227b7f3dd9edf5c5f75ec3d4303fa96755d0bbca07e0R732) even passes given I have the exact same scenario.

jayceslesar:
Previously this was not implemented at the SQL engine level per #3295

This is a rough attempt at exposing that functionality.

Will close #3295


jayceslesar:
Hah, every little change I make seems to break something.  Generally want to implement `merge_asof` like:
```py
def merge_asof(left: TableExpr, right: TableExpr, on: str, tolerance: float = 0, tolerance_direction: str = 'upper'):
    """Perform a merge asof join on two tables.

    Args:
        left (TableExpr): left table
        right (TableExpr): right table
        on (str): column to join on
        tolerance (float): essentially an offset
        tolerance_direction (str, optional): direction of tolerance. Defaults to 'upper'.

    Returns:
        [type]: [description]
    """
    following = (1, None)
    win = ibis.range_window(following=following, order_by=on)
    class_ = right.mutate(
        timeend=right[on].min().over(win)
    )

    if tolerance_direction == 'upper':
        clause = left[on] >= class_[on] - tolerance
    elif tolerance_direction == 'lower':
        clause = left[on] <= class_[on] + tolerance
    else:
        clause =  (left[on] >= class_[on] - tolerance) | (left[on] <= class_[on] + tolerance)

    res = left.left_join(
        class_,
        [
            clause,
            (left[on] < class_.timeend) | class_.timeend.isnull(),
        ],
    )

    return res
```

cpcloud:
@jayceslesar Thanks for doing this! Looks like it's going to require more lifting than expected.

In the meantime, I've taken the great work you did in #3295 and started a wiki, with the `asof_join` as the inaugural recipe. Check it out [here](https://github.com/ibis-project/ibis/wiki/%60asof_join%60-in-SQL).

PR #3316, opened 2022-01-14
refactor: unify dask and pandas selection logic, fix dask bug

gerrymanoim:
Follow up for https://github.com/ibis-project/ibis/pull/3117. 

This started as a bug fix for my previous PR (that I've now added a test case for). But in the process I realized that the dask and pandas backends should share this logic. So this PR:

- fixes a bug in `build_df_from_selection` 
- enhances `build_df_from_selection` to work in the case of joins
- moves `build_df_from_selection` to the pandas backend
- Modifies selection in the pandas and dask backends to use common functions where appropriate

The performance improvements are not as impactful in the pandas backend, but I think these backends should share code where possible.


gerrymanoim:
Looks like errors are pyspark failing to launch?

cpcloud:
@gerrymanoim Yeah, I'm not sure what's going on with either of those yet. I'm trying to reproduce the Spark failures locally.

cpcloud (approved):
This looks reasonable to me! Thanks @gerrymanoim

Issue #2815, opened 2021-06-08
bug: fix verify_meta=False usage in dask udfs


gerrymanoim:
Fix/Remove usage of `verify_meta=False` in the dask udf code for scalar outputs as that's blocking us from updating dask.

cpcloud:
@gerrymanoim Is this still a problem?

gerrymanoim:
We still have this piece of code: https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/udf.py#L160. Clearly it hasn't stopped us from updating dask, but I don't know if that's still required. Let me try removing it.

cpcloud:
@gerrymanoim Did you get a chance to look into this?

gerrymanoim:
I put this fix up https://github.com/ibis-project/ibis/pull/3316, but this implementation required basically doing extracting all the pieces of `dd.from_delayed` besides the meta verification. Given that, I'm going to just close this issue and change my PR to simply remove the shim we don't need anymore.

cpcloud:
Awesome, thanks!

PR #3318, opened 2022-01-14
refactor: remove shim which is no longer needed

gerrymanoim:
Closes https://github.com/ibis-project/ibis/issues/2815. 

Also remove a shim we no longer need.


cpcloud (approved):
LGTM, thanks!

PR #3320, opened 2022-01-15
ci: run nix builds only when relevant files have changed

cpcloud:
This PR changes our nix builds to only when relevant files such as any nix files as well as dependency related files have changed.


kszucs (approved):
I'm not a nix expert but on principle it looks good to me.

PR #3328, opened 2022-01-18
refactor(sqlalchemy): remove unnecessary _AutoCloseCursor #2947

### LINKED ISSUE #2947: refactor: remove __del__ methods
We seem to be depending on destructor-like behavior of `__del__` which is a mistake. `__del__` is not guaranteed to be called, basically ever: https://docs.python.org/3/reference/datamodel.html#object.__del__

We should find alternatives to these methods (often a context manager works, but not always) and then remove them from the library.

saulpw:
Closes #2947


cpcloud (approved):

cpcloud:
Thanks @saulpw!

PR #3329, opened 2022-01-18
ci: split up min version jobs

cpcloud:
Fixes CI so that caching does not happening on min version jobs


cpcloud:
Merging to get CI green again.

PR #3332, opened 2022-01-19
test: add test for DatabaseTable equality semantics

### LINKED ISSUE #2973: test: check the semantics of DatabaseTable equality
Right now, `ops.DatabaseTable`s will compare equal if all `args` compare equal, which is problematic since `source` is a database connection.

A database connection (a `Client` in ibis) has no clear equality semantics, nor should it IMO.

Our test suite is actually depending on pointing to the same client object, and will fail if run in a different order than the default pytest ordering because the same database will point to different clients. This is a bug: https://github.com/ibis-project/ibis/pull/2971/checks?check_run_id=3587181882.

cpcloud:
Closes #2973


cpcloud:
@jreback Mind reviewing this?

jreback (approved):
lgtm.

Issue #3334, opened 2022-01-19
feat: rethink expr repr


emilyreff7:
### Context
By default, `Expr`'s repr does a tree traversal to give a [representation of the full expression tree](https://github.com/ibis-project/ibis/blob/a3583c3ffd711a220d9269d9f13f98ec68b99f8f/ibis/expr/format.py#L68). While this repr can contain useful debugging information, it can also have adverse performance implications that are difficult to catch until after the fact. Performance regressions can occur for example when the repr is included in error messages, particularly when the tree is large and/or when the repr is requested in a tight loop (e.g. within an error message that is hit on validation of each table column, which itself may be called within an arbitrary-sized loop depending on user code).

### Proposal
To avoid future performance regressions, it may be beneficial to rethink the default repr for expression trees. Perhaps a more reasonable approach would define a simpler, constant-time implementation for `__repr__`, and also expose a new `Expr` method (e.g. `explain`, `inspect`, `info`, etc.) that would display the entire expression tree. If we go this route, we should decide what exactly to include in the repr (e.g. the type of the expr, any schema information? column names but up to a set limit?)

### Relevant PRs
https://github.com/ibis-project/ibis/pull/2917
https://github.com/ibis-project/ibis/pull/3335

saulpw:
I think this is a good idea, @emilyreff7.  How about `.get_repr()` with a kwarg to indicate how much information to display (by default the entire tree), and then `__repr__()` would call something like `.get_repr(depth=None)`?

cpcloud:
Closing this out, we've got a much improved repr in both UX and speed.

PR #3335, opened 2022-01-19
Remove full tree repr from rule validator error message

emilyreff7:
### Proposed Change
Avoid calling into the TableExpr's repr within these rule validator error messages, and instead use `type(table)`. Using the O(N) repr causes performance issues when called in a tight loop.

See this issue for a follow-up idea: https://github.com/ibis-project/ibis/issues/3334

### DIFF
--- a/ibis/expr/rules.py
+++ b/ibis/expr/rules.py
@@ -447,11 +447,11 @@ def column_from(name, column, *, this):
                 return column
             else:
                 raise com.IbisTypeError(
-                    f"Passed column is not a column in {table}"
+                    f"Passed column is not a column in {type(table)}"
                 )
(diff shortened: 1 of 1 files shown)


cpcloud:
@emilyreff7 Looks good.

Is this a performance enhancement? If so, would you mind formatting your commit like this:

```
perf(repr): remove full tree repr from rule validator error message
```

If not then this works:

```
chore(repr): remove full tree repr from rule validator error message
```

jreback (approved):
lgtm

cpcloud (approved):

cpcloud:
Thanks @emilyreff7!

Issue #1223, opened 2017-11-13
Getting started with Ibis on Redshift


CerebralMastication:
Hey team, kudos for hard work on Ibis. I've been trying to start using Ibis against a Redshift DB and the Postgres connectivity seems to work just dandy. However Ibis is not behaving as I would expect, which might be an issue with my expectations rather than Ibis. 

let me give you an example of my snag:

I set up a connection to redshift using ibis.sql.postgres.api.connect() and all looks good. 

Then I create some Ibis table objects:

```
ds = con.table('dimOfInterest')
fev = con.table('factOfInterest')
```
this works as expected and I can execute ds.info() for example and see info. So I then do a join:

`df = ds.inner_join(fev, ds.key == fev.key)
`
all's good... so I do a filter:

`df_filtered = df.filter( [df.snapshotname == 'MyName'] )
`
and that fails with `AttributeError: 'TableExpr' object has no attribute 'snapshotname'`

So I get the general idea that even though I have done a join to create df, somehow Ibis does not know what fields are in df. Yet I don't yet want to fetch all of df by doing a `df.execute()`. 

Q1: What am I missing here? Is there a way to combine multiple filter steps and accumulate them and then at the end have Ibis roll those all into one SQL statement and execute it? I think that is my expectation primarily because that's how dplyr works over in R land. Is my expectation wrong?

Interestingly, when I run df.execute() I get the following error: `RelationError: Joined tables have overlapping names: ['key', 'modelkey']`. Which is pointing out the obvious to me. Key is my joining field, so it better be in both tables! And modelkey is, indeed, redundant in naming, but I'm surprised that this is a problem since in Pandas I don't recall redundant names borking a join. 

Q2: How does one avoid redundant naming? Am I going to have to explicitly list which fields to keep? That begins to feel a lot like just writing SQL and is the sort of thing I had hoped to avoid in Ibis-land. 

I have read the documentation about [Ibis for SQL Programmers](http://docs.ibis-project.org/sql.html) and it illustrates multiple times the use of `ibis.impala.compile()` which no longer seems to be the location of this function. I looked around and I can't seem to find where that now lives. 

Q3: What's the modern incarnation of `ibis.impala.compile()`? I'm using Ibis 0.12.0.

Thank you all for your hard work and for handholding a newbie a bit!

cpcloud:
@CerebralMastication Thanks for raising an issue!

Q1/Q2:
---

[...]

PR #3339, opened 2022-01-20
refactor(api): remove the materialize method from TableExprs

### LINKED ISSUE #1045: bug: selection of renamed columns is optimized incorrectly
Hello.
In doing a double join to connect three tables, I've identified behavior of Ibis that I don't understand. I'm hoping that someone can shed light on why it's occurring, or how the error message relates to what is occurring. If I write the script one way, the double join works. If I change a few lines, the second join throws a "More than one base table not implemented" error. I'm using the postgres connection.

A test script is below. In it, I create some small test tables that mimic the data I'm working with. For example, some column types need to be recast, and some column names appear in more than one table. I rename some of these columns to avoid conflicts.  

In the script, if the "if True" is changed to False, the alternative script is run, which throws the error. Otherwise, the script works fine.  Any suggestions to improve this code snippet are also welcome. 

```
con = ibis.postgres.connect(host="localhost", database="...", user='...', password="...")
con.raw_sql("drop table if exists test1")
con.raw_sql("drop table if exists test2")
con.raw_sql("drop table if exists test3")

con.raw_sql("create table test1 (id1 integer not null, val1 real)")
for i in range(10):
  con.raw_sql("insert into test1 values (1, 10), (2, 20), (3, 20), (2, 21)")

con.raw_sql("create table test2 (id2a integer not null, id2b bigint, val2 double precision)")
for i in range(10):
  con.raw_sql("insert into test2 values (1, 4, 15), (2, 5, 25)")

con.raw_sql("create table test3 (id3 text not null, val2 double precision, dt timestamp without time zone)")
for i in range(10):
  con.raw_sql("insert into test3 values ('4', 15, '2016-07-17 19:55:03'), (2, 25, '2016-07-17 19:55:03')")

t1 = con.table('test1')
t2 = con.table('test2')
t3 = con.table('test3')

c = t3.id3.cast('int64').name('id3')
ls = t3.columns
ls.remove('id3')
t3 = t3[ls].mutate(c)

if True:
  """this alternative works fine""" 
  t3 = t3[[t3[i].name('t3_'+ i) for i in t3.columns]]
  print("\n first t3 info:")
  print(t3.info())
  
  cond1 = t3.t3_id3 == t2.id2b

else:
  """this alternative causes an error on the second join"""
  c = t3.id3.name('t3_val2')
  ls = t3.columns
  ls.remove('val2')
  t3 = t3[ls].mutate(c) 
  print("\n second t3 info:")
  print(t3.info()) 
  
  cond1 = t3.id3 == t2.id2b

t4 = t3.join(t2, cond1).materialize()  # first join

t1 = t1[[t1[i].name('t1_'+ i) for i in t1.columns]]

cond2 = t1.t1_id1 == t4.id2a
t5 = t1.join(t4, cond2).materialize()  # second join

print(t5.info())

```
The t3 table seems to be mostly the same with either alternative, other than some renamed columns.
The error raised on print(t5.info()) is:
  File ".../lib/python3.5/site-packages/ibis/expr/analysis.py", line 1056, in find_source_table
  raise NotImplementedError('More than one base table not implemented')
  NotImplementedError: More than one base table not implemented

### LINKED ISSUE #1387: bug: Selecting unambiguous columns with strings raises an exception
```
In [18]: t = ibis.table([('a', 'int64'), ('b', 'string')])
                                                                          
In [19]: s = ibis.table([('b', 'int64'), ('c', 'string')])   
                                   
In [20]: joined = t.left_join(s, [t.b == s.c]) 
                                                          
In [21]: joined[t, 'c']
```

### LINKED ISSUE #2427: bug: filters not applied to joined tables
Given, two tables, if they are joined, and then a filter is applied, the filter is actually being ignored:

```python
import ibis

connection = ibis.sqlite.connect('data/geography.db')
countries = connection.table('countries')
gdp = connection.table('gdp')

expr = (countries.inner_join(gdp,
                             predicates=countries['iso_alpha3'] == gdp['country_code'])
                 .filter(gdp['year'] == 2017))
expr
```
Intuitively, the result should only contain the rows with `year == 2017`, but the filter is not applied.

The database in this example can be found at: https://github.com/datapythonista/ibis/tree/tutorial/docs/source/tutorial/data

The SQL compiled from the expression, can be obtained with:
```python
>>> str(expr.compile())
SELECT t0.iso_alpha2, t0.iso_alpha3, t0.iso_numeric, t0.fips, t0.name, t0.capital, t0.area_km2, t0.population, t0.continent, t0.country_code, t0.year, t0.value 
FROM (SELECT t1.iso_alpha2 AS iso_alpha2, t1.iso_alpha3 AS iso_alpha3, t1.iso_numeric AS iso_numeric, t1.fips AS fips, t1.name AS name, t1.capital AS capital, t1.area_km2 AS area_km2, t1.population AS population, t1.continent AS continent, t2.country_code AS country_code, t2.year AS year, t2.value AS value 
FROM base.countries AS t1 JOIN base.gdp AS t2 ON t1.iso_alpha3 = t2.country_code) AS t0, base.gdp 
WHERE base.gdp.year = ?
```

### LINKED ISSUE #2997: feat: support suffixes in join API
Ibis' `join` API does not currently accept suffixes as a parameter, leaving backends to either use some default value or raise an error at execution time. It would be nice to directly support suffixes in the API itself.

### LINKED ISSUE #3090: bug: error in mutating joined tables using Pandas backend
Hey again, thanks for the super quick help on the previous issue, #3082 . I think I may have found an additional bug as a follow up to that fix in that mutating the newly joined table doesn't seem to work. Here's an example code:

```
import ibis
import pandas as pd

df = pd.DataFrame(
    {"p_Order_Priority": ["C", "H", "L", "M"], "p_count": [9, 9, 15, 11], "p_density": [0.204545, 0.204545, 0.340909, 0.250000]})
df_2 = pd.DataFrame(
    {"q_Order_Priority": ["C", "H", "L", "M"], "q_count": [13, 21, 12, 10], "q_density": [0.232143, 0.375000, 0.214286, 0.178571]})

conn = ibis.pandas.connect({"df": df, "df_2": df_2})

ibis_table_1 = conn.table("df")
ibis_table_2 = conn.table("df_2")

joined = ibis_table_1.outer_join(
    ibis_table_2, predicates=ibis_table_1["p_Order_Priority"] == ibis_table_2["q_Order_Priority"])
joined = joined.materialize()

joined = joined.mutate(
    bins=joined["p_Order_Priority"].isnull().ifelse(
        joined["q_Order_Priority"], joined["p_Order_Priority"]),
    p_count=joined["p_count"].fillna(0),
    q_count=joined["q_count"].fillna(0),
    p_density=joined.p_density.fillna(1e-10),
    q_density=joined.q_density.fillna(1e-10),
    features="Order_Priority"
)

print(joined.execute())
```

The above results in the following error:
```
  File "C:\Users\tanaimn\AppData\Roaming\Python\Python38\site-packages\pandas\core\indexes\base.py", line 3361, in get_loc    
    return self._engine.get_loc(casted_key)
  File "pandas\_libs\index.pyx", line 76, in pandas._libs.index.IndexEngine.get_loc
  File "pandas\_libs\index.pyx", line 108, in pandas._libs.index.IndexEngine.get_loc
  File "pandas\_libs\hashtable_class_helper.pxi", line 5198, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas\_libs\hashtable_class_helper.pxi", line 5206, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'p_count'

The above exception was the direct cause of the following exception:

  File "bug_test.py", line 28, in <module>
    print(joined.execute())
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\expr\types.py", line 273, in execute
    return self._find_backend().execute(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\__init__.py", line 121, in execute      
    return execute_and_reset(query, params=params, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\core.py", line 487, in execute_and_reset
    result = execute(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\multipledispatch\dispatcher.py", line 278, in __call__       
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\trace.py", line 138, in traced_func     
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\core.py", line 432, in main_execute
    return execute_with_scope(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\core.py", line 218, in execute_with_scope
    result = execute_until_in_scope(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\trace.py", line 138, in traced_func
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\core.py", line 359, in execute_until_in_scope
    result = execute_node(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\multipledispatch\dispatcher.py", line 278, in __call__
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\trace.py", line 138, in traced_func
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\execution\selection.py", line 328, in execute_selection_dataframe        
    pandas_object = compute_projection(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\multipledispatch\dispatcher.py", line 278, in __call__
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\execution\selection.py", line 144, in compute_projection_column_expr     
    execute(expr, scope=scope, timecontext=timecontext, **kwargs),
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\multipledispatch\dispatcher.py", line 278, in __call__
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\trace.py", line 138, in traced_func
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\core.py", line 432, in main_execute
    return execute_with_scope(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\core.py", line 218, in execute_with_scope
    result = execute_until_in_scope(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\trace.py", line 138, in traced_func
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\core.py", line 327, in execute_until_in_scope
    scopes = [
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\core.py", line 328, in <listcomp>
    execute_until_in_scope(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\trace.py", line 138, in traced_func
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\core.py", line 359, in execute_until_in_scope
    result = execute_node(
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\multipledispatch\dispatcher.py", line 278, in __call__
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\trace.py", line 138, in traced_func
    return func(*args, **kwargs)
  File "C:\Users\tanaimn\.conda\envs\ibis_test\lib\site-packages\ibis\backends\pandas\execution\generic.py", line 389, in execute_table_column_df_or_df_groupby    return data[op.name]
  File "C:\Users\tanaimn\AppData\Roaming\Python\Python38\site-packages\pandas\core\frame.py", line 3455, in __getitem__
    indexer = self.columns.get_loc(key)
  File "C:\Users\tanaimn\AppData\Roaming\Python\Python38\site-packages\pandas\core\indexes\base.py", line 3363, in get_loc
    raise KeyError(key) from err
KeyError: 'p_count'
```

One workaround I found which solved this problem was to add an empty mutate after materializing the joined table, eg.

```
joined = joined.materialize().mutate()
```

but I do imagine that this wouldn't be the intended functionality. Thanks again for any help!

### LINKED ISSUE #3295: bug: merge asof raises an error when the predicate column(s) are overlapping

cpcloud:
This PR removes the `.materialize()` method from `TableExpr`.

This method causes a lot of confusion, and created a situation where doing simple
things like

```python
t.join(s, t.foo == s.foo).select(["unambiguous_column"])
```

raised an exception.

It turns out that `materialize` isn't necessary, but a few changes were necessary:

1. Add a join suffixes API (#2997) to resolve overlapping column names. This is a new API.
1. Disable predicate pushdown in the case when there's a predicate on a
   `Selection`. This was necessary to get the pandas and dask backends to work
   without materialize. This was working before because the onus to project columns
   out was placed on the user.

This is a breaking change for any code that uses `materialize`.

~However, existing code that does *not* use `materialize` should continue to work as is.~

There are some breaking changes introduced here in the case of overlapping column names.

If there are any overlapping column names, a suffix will be attached to both the left and right tables. So, in the case of `s.asof_join(t, "time")` the resulting schema will have both a `time_x` and a `time_y` column.

See my explanation here https://github.com/ibis-project/ibis/issues/1223#issuecomment-345476096 for why this is the case.

~#1045~ the initial problem is addressed, but another bug was exposed
Closes #1387
Closes #2427
Closes #2997
Closes #3090
Closes #3295


wesm:
cc @tswast

wesm:
It's been a while since I worked on this part of the codebase, but IIRC one of the main purposes of of `materialize` was to put off the resolution of overlapping column names. Is it necessary to disambiguate the column names even for keys that are included in equijoin predicates? For a `$table0.$COL = $table1.$COL` predicate, wouldn't it be more intuitive for only `$COL` to appear in the result rather than `$COL_x` and `$COL_y`?

cpcloud:
@wesm Yeah, that came up #1223.

There's only a single case where it's possible to do this and that's inner joins where the join comparison operation is equality.

I walk through an example in that issue laying out why disambiguation is necessary to prevent information loss.

Given the number of issues and confusion that have come up because of materialize it seems like a reasonable tradeoff for users to have to handle the disambiguation.

It could be a useful follow up to investigate how much work it is to handle the single case where it is possible.

cpcloud:
The main challenge is coming up with a reliable and maintainable function that takes a join predicate as input and returns a boolean indicating whether the involved keys are potentially ambiguous.

I think this requires full traversal of the expression, since any expression and subexpressions that aren't of the form `t1.$COL_tn1 = t2.$COL_t2 AND ... AND tn1.$COL_tn1 = tn.COL_tn` would return `False`.

ibis/backends/pandas/execution/join.py
```diff
@@ -32,27 +31,14 @@ def execute_cross_join(op, left, right, **kwargs):
...
-    # inner/outer doesn't matter because every row matches every other row
-    result = pd.merge(
-        new_left,
-        new_right,
-        how='inner',
-        on=key,
+    return pd.merge(
+        left,
+        right,
+        how='cross',
```
jreback, line 37: hmm this requires IIRC pandas 1.3.x or great (for the cross)

jreback (changes requested):

ibis/backends/pandas/execution/join.py
cpcloud: Thoughts on bumping the minimum pandas version?

jreback: oh actually this is in 1.2. yeah ok with bumping.

cpcloud: We're already on 1.2.5, so this is a no-op! Woohoo!

tswast:
I've definitely been confused by `materialize()` before, so +1 on this change. Tested this PR locally against the `ibis-bigquery` package and didn't see any failures that would have been caused by this. That said, I'm not sure if we have test cases for the overlapping equijoin predicates Wes identified.

There do appear to be some whitespace changes since the last time I ran the test suite, but that's not such a big deal:

```
E           SELECT count(`foo`) AS `count`
E           FROM (
E             SELECT `string_col`, sum(`float_col`) AS `foo`
E             FROM (
E               SELECT `float_col`, `timestamp_col`, `int_col`, `string_col`
E               FROM `swast-scratch.ibis_gbq_testing.functional_alltypes`
E         -     WHERE `timestamp_col` < @my_param
E             ) t1
E         +   WHERE `timestamp_col` < @my_param
E             GROUP BY 1
E           ) t0
```

tswast:
Actually, that failure isn't just whitespace, it moved to a different subquery.

cpcloud:
Interesting. I'll need to add a test case.

Regarding overlapping predicates, we have a good number of tests for that style of usage, it's one of the more common things for pandas users to try.

tswast:
Sure thing. Here's the full test output, including the ibis expression:

```
(ibis-bigquery-dev) ➜  ibis-bigquery git:(d0e43ef) pytest tests/system/test_client.py::test_subquery_scalar_params -vv
============================= test session starts ==============================
platform darwin -- Python 3.8.10, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- /usr/local/Caskroom/miniconda/base/envs/ibis-bigquery-dev/bin/python
cachedir: .pytest_cache
rootdir: /Users/swast/src/github.com/ibis-project/ibis-bigquery
plugins: cov-2.12.1, mock-3.6.1
collected 1 item

tests/system/test_client.py::test_subquery_scalar_params FAILED          [100%]

=================================== FAILURES ===================================
_________________________ test_subquery_scalar_params __________________________

alltypes = BigQueryTable[r0, name=swast-scratch.ibis_gbq_testing.functional_alltypes]
  index           int64
  Unnamed_0       i..._string_col string
  string_col      string
  timestamp_col   timestamp
  year            int64
  month           int64
project_id = 'swast-scratch', dataset_id = 'ibis_gbq_testing'

    def test_subquery_scalar_params(alltypes, project_id, dataset_id):
        t = alltypes
        param = ibis.param("timestamp").name("my_param")
        expr = (
            t[["float_col", "timestamp_col", "int_col", "string_col"]][
                lambda t: t.timestamp_col < param
            ]
            .groupby("string_col")
            .aggregate(foo=lambda t: t.float_col.sum())
            .foo.count()
        )
        result = expr.compile(params={param: "20140101"})
        expected = """\
    SELECT count(`foo`) AS `count`
    FROM (
      SELECT `string_col`, sum(`float_col`) AS `foo`
      FROM (
        SELECT `float_col`, `timestamp_col`, `int_col`, `string_col`
        FROM `{}.{}.functional_alltypes`
        WHERE `timestamp_col` < @my_param
      ) t1
      GROUP BY 1
    ) t0""".format(
            project_id, dataset_id
        )
>       assert result == expected
E       AssertionError: assert ('SELECT count(`foo`) AS `count`\n'\n 'FROM (\n'\n '  SELECT `string_col`, sum(`float_col`) AS `foo`\n'\n '  FROM (\n'\n '    SELECT `float_col`, `timestamp_col`, `int_col`, `string_col`\n'\n '    FROM `swast-scratch.ibis_gbq_testing.functional_alltypes`\n'\n '  ) t1\n'\n '  WHERE `timestamp_col` < @my_param\n'\n '  GROUP BY 1\n'\n ') t0') == ('SELECT count(`foo`) AS `count`\n'\n 'FROM (\n'\n '  SELECT `string_col`, sum(`float_col`) AS `foo`\n'\n '  FROM (\n'\n '    SELECT `float_col`, `timestamp_col`, `int_col`, `string_col`\n'\n '    FROM `swast-scratch.ibis_gbq_testing.functional_alltypes`\n'\n '    WHERE `timestamp_col` < @my_param\n'\n '  ) t1\n'\n '  GROUP BY 1\n'\n ') t0')
E           SELECT count(`foo`) AS `count`
E           FROM (
E             SELECT `string_col`, sum(`float_col`) AS `foo`
E             FROM (
E               SELECT `float_col`, `timestamp_col`, `int_col`, `string_col`
E               FROM `swast-scratch.ibis_gbq_testing.functional_alltypes`
E         -     WHERE `timestamp_col` < @my_param
E             ) t1
E         +   WHERE `timestamp_col` < @my_param
E             GROUP BY 1
E           ) t0

tests/system/test_client.py:230: AssertionError
=========================== short test summary info ============================
FAILED tests/system/test_client.py::test_subquery_scalar_params - AssertionEr...
============================== 1 failed in 42.37s =============================
```

cpcloud:
Is it really strange that this happens, given that nearly all the code I touched was related to joins. Having a look now.

cpcloud:
Well, it's this line 😞 

https://github.com/ibis-project/ibis/pull/3339/files#diff-9bfeef14b16e62a849fc8c72b04bfab4fc8cee65fa236f1ad54bd440fded7bcfR769

cpcloud:
@tswast The query is still correct though isn't it?

cpcloud:
Unfortunately, to remove materialize I had to remove some of ibis's overly aggressive predicate pushdown.

The generated code produced is probably not a good thing to explicitly test, as we will need to make changes from time to time to the to compilers that result in different generated code to fix bugs or add features.

ibis/backends/pandas/execution/join.py
cpcloud: @jreback Is this addressed? Can you remove the change request when you get a moment?

tswast:
> @tswast The query is still correct though isn't it?

Yes. Appears to be. I'm able to execute it, which I wasn't sure I would be able to do because of the subquery location.

> Unfortunately, to remove materialize I had to remove some of ibis's overly aggressive predicate pushdown.

Makes sense. BigQuery has its own query optimizer, so I'm not all too worried about that.

> The generated code produced is probably not a good thing to explicitly test, as we will need to make changes from time to time to the to compilers that result in different generated code to fix bugs or add features.

We can update the tests to actually execute the SQL. It might increase test time by quite a bit, though.

cpcloud:
@tswast On the other hand, this was a nice catch, so I can add this as a test to the generic SQL test suite and we can take on the burden of finding this first at least!

cpcloud:
Added a test for the subquery case. Thanks @tswast!

ibis/expr/api.py
```diff
@@ -3844,37 +3844,93 @@ def _time_sub(left, right):
 }
 
 
-def join(left, right, predicates=(), how='inner'):
-    """Perform a relational join between two tables. Does not resolve resulting
-    table schema.
+def _dedup_join_columns(expr, *, left, right, rename_left, rename_right):
```
jreback, line 3847: should be be located elsewhere and imported here?

jreback (approved):
lgtm, minor comment

ibis/expr/api.py
cpcloud: Sounds good, I'll move to `operations/relations.py`.

cpcloud: Done.

cpcloud:
Thinking about this some more, I think it's probably best to keep a simpler `suffixes` API here. It's familiar elsewhere (pandas, dplyr) and more general renaming can be done with `TableExpr.relabel`

Issue #3358, opened 2022-01-24
bug: ibis.expr.signature.Parameter is not pickleable 


gerrymanoim:
Minimal repro:

```python
import pickle
from ibis.expr.signature import Parameter

test = Parameter("test")
pickle.loads(pickle.dumps(test)
```
Raises `TypeError: __init__() takes 2 positional arguments but 3 were given`.

PR #3359, opened 2022-01-24
fix: ibis.expr.signature.Parameter is not pickleable

gerrymanoim:
Fixes https://github.com/ibis-project/ibis/issues/3358.


cpcloud:
Probably worth adding a small test to catch regressions. Thanks for the PR!

gerrymanoim:
> Probably worth adding a small test to catch regressions. Thanks for the PR!

Yeah agreed. Let me add that to this PR.

cpcloud (approved):
LGTM, thanks!

PR #3367, opened 2022-01-28
fix(api): add ibis.map and .struct

### LINKED ISSUE #3118: feat: allow dicts in ibis.literal for structs
Currently a struct must be an `OrderedDict`, but since 3.6 there's no reason we can't allow at least `dict`s any possibly any mapping:

```
>>> import ibis
>>> ibis.literal({"a": "b", "c": "d"}, type="struct<a: string, c: string>")
TypeError: Value {'a': 'b', 'c': 'd'} cannot be safely coerced to struct<a: string, c: string>
```

saulpw:
Closes #3118.

### DIFF
--- a/ibis/expr/datatypes.py
+++ b/ibis/expr/datatypes.py
@@ -535,7 +535,7 @@ class Struct(DataType):
 
     @property
     def pairs(self) -> Mapping:
-        return collections.OrderedDict(zip(self.names, self.types))
+        return dict(zip(self.names, self.types))
 
     def __getitem__(self, key: str) -> DataType:
         return self.pairs[key]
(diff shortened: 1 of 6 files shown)


ibis/expr/datatypes.py
```diff
@@ -1414,6 +1414,24 @@ def can_cast_string_to_temporal(
...
+    )
+
+
+@castable.register(Struct, Struct)
+def can_cast_struct(source, target, **kwargs):
+    source_pairs = source.pairs
+    target_pairs = target.pairs
+    for k in {*source.names, *target.names}:
+        if k in target_pairs:
+            if not castable(source_pairs[k], target_pairs[k]):
```
cpcloud, line 1430:
```suggestion
    for name in {*source.names, *target.names}:
        if name in target_pairs:
            if not castable(source_pairs[name], target_pairs[name]
```

ibis/expr/types.py
```diff
@@ -1247,7 +1248,7 @@ def literal(value, type=None):
         return ops.Literal(value, dtype=dtype).to_expr()
 
 
-def array(values):
+def array(values, type=None):
```
cpcloud, line 1251: Can you add a test case for `array(..., type="some_type_string")`?

cpcloud (changes requested):
Small bit about `k` -> `name`. Thanks, looks great!

ibis/expr/types.py
saulpw: done!

cpcloud (approved):
Sweet. This looks great.

PR #3368, opened 2022-01-28
refactor(benchmarks): move to pytest-benchmark

cpcloud:
This PR moves our benchmarking suite to `pytest-benchmark`.


jreback (commented):
looks fine. no real reason to switch off of asv

cpcloud:
> looks fine. no real reason to switch off of asv

I think there is. It's not particularly well maintained and it doesn't integrate with existing tools.

It's also doing too many things, like managing environments.

It also doesn't offer anything in the way of statistical benchmarking. It tells me nothing about the distribution of results.

Lastly I find it difficult and annoying to use. Why do I have to tell it about the details of the machine I'm on (for example)?

pytest-benchmark does all of these things and doesn't have a bunch of extra features we don't need and it works the same way as any other pytest plugin.

jreback:
ok it's getting a facelift actually rn

and we don't really have many benchmarks so it's easy to switch 

so not objecting strongly

PR #3369, opened 2022-01-28
fix: implement create_table correctly for pandas and dask backends

### LINKED ISSUE #3020: test: pandas and dask backends tests are not inferring the correct schema
The next tests were xfailed because pandas and dask don't implement `drop_table`, even if it seems somehow unrelated:

- `ibis/backends/tests/test_client.py::test_create_table_from_schema`
- `ibis/backends/tests/test_client.py::test_nullable_input_output`

When executed, the tests fail because nullable fields in the schema are not nullable in the created table.

We moved to xfail the tests only when `create_table` is not implemented, ignoring `drop_table`, which is not used in the test. Manually xfailing the tests for pandas and dask in #3018, but we probably want to revisit this, and see if it's a bug.

cpcloud:
This PR addresses incorrect implementations of create_table in the pandas and dask backends.

The backends now store an additional dictionary mapping inferred or user-specified (when `create_table` is called)
ibis Schema objects, and uses those to construct `TableExpr`s from `client.table`.

Closes #3020.

### DIFF
--- a/ibis/backends/pandas/__init__.py
+++ b/ibis/backends/pandas/__init__.py
@@ -1,5 +1,8 @@
+from __future__ import annotations
+
+from typing import Any, MutableMapping
+
 import pandas as pd
-import toolz
 
 import ibis.common.exceptions as com
 import ibis.config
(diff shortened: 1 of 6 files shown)


ibis/backends/pandas/__init__.py
```diff
@@ -90,11 +96,58 @@ def load_data(self, table_name, obj, **kwargs):
...
+            )
+        return schema
 
     def compile(self, expr, *args, **kwargs):
         return expr
 
+    def create_table(self, table_name, obj=None, schema=None):
+        """Create a table."""
+        if obj is None and schema is None:
+            raise com.IbisError('Must pass expr or schema')
```
jreback, line 114: test that hits this?

jreback: why isn't obj required?

jreback: oh i c you can just pass schema ok

jreback (approved):
small comment about testing

ibis/backends/pandas/__init__.py
cpcloud: This was preexisting, I just moved the code into the base class.

PR #3371, opened 2022-01-28
test(api): remove unnecessary api test

saulpw:
There's no need to lock-down the Ibis API, so this test only introduces unnecessary overhead to changing the API.


cpcloud (approved):

cpcloud:
Thanks 😊

PR #3375, opened 2022-01-29
refactor: using ibis.expr.api.schema inside ibis.expr.api.table

### LINKED ISSUE #3040: refactor: use ibis.schema() to coerce the schema argument of ibis.table()
The schema coercion is slightly different between

https://github.com/ibis-project/ibis/blob/430b79213979a8ef80f681f94b25690b3bfd194a/ibis/expr/api.py#L289-L293

and

https://github.com/ibis-project/ibis/blob/430b79213979a8ef80f681f94b25690b3bfd194a/ibis/expr/api.py#L265-L269

We should unify the behavior.

See [TODO note](https://github.com/ibis-project/ibis/pull/3025/files#diff-8f6e03b936fae42459cef36d6bb2691392fb1ed9bd03c9846152ac6ebb6da892R281) for reference.

EdAbati:
fixes #3040

I had to create an alias for the `schema` function since the `table` function has a `schema` argument, please let me know if there is a cleaner approach or if I missed something.
Thank you very much!

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -7,7 +7,7 @@ import datetime
 import functools
 import numbers
 import operator
-from typing import Any, Iterable
+from typing import Any, Iterable, Mapping
 
 import dateutil.parser
 import pandas as pd
@@ -258,7 +258,9 @@ def sequence(values):
 
 
 def schema(
-    pairs: Iterable[tuple[str, dt.DataType]] | None = None,
+    pairs: Iterable[tuple[str, dt.DataType]]
+    | Mapping[str, dt.DataType]
+    | None = None,
     names: Iterable[str] | None = None,
     types: Iterable[str | dt.DataType] | None = None,
 ) -> sch.Schema:
@@ -267,7 +269,8 @@ def schema(
     Parameters
     ----------
     pairs
-        List of name, type pairs. Mutually exclusive with `names` and `types`.
+        List or dictionary of name, type pairs. Mutually exclusive with `names`
+        and `types`.
     names
         Field names. Mutually exclusive with `pairs`.
     types
@@ -283,9 +286,12 @@ def schema(
     ...              types=['string', 'int64', 'boolean'])
     """
     if pairs is not None:
-        return Schema.from_tuples(pairs)
+        return Schema.from_dict(dict(pairs))
     else:
         return Schema(names, types)
+
+
+_schema = schema
 
 
 def table(schema, name=None):
@@ -306,10 +312,7 @@ def table(schema, name=None):
     table : TableExpr
     """
     if not isinstance(schema, Schema):
-        if isinstance(schema, dict):
-            schema = Schema.from_dict(schema)
-        else:
-            schema = Schema.from_tuples(schema)
+        schema = _schema(pairs=schema)
 
     node = ops.UnboundTable(schema, name=name)
     return node.to_expr()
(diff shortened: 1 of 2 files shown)


ibis/expr/api.py
```diff
@@ -283,11 +286,16 @@ def schema(
     ...              types=['string', 'int64', 'boolean'])
     """
     if pairs is not None:
+        if isinstance(pairs, dict):
+            return Schema.from_dict(pairs)
         return Schema.from_tuples(pairs)
```
cpcloud, line 291:
```suggestion
        return Schema.from_dict(dict(pairs))
```

ibis/expr/api.py
```diff
@@ -258,7 +258,9 @@ def sequence(values):
 
 
 def schema(
-    pairs: Iterable[tuple[str, dt.DataType]] | None = None,
+    pairs: Iterable[tuple[str, dt.DataType]]
+    | dict[str, dt.DataType]
```
cpcloud, line 262:
```suggestion
    | Mapping[str, dt.DataType]
```

cpcloud (changes requested):
Thanks @EdAbati! Couple of small suggestions then this is good to go!

EdAbati:
Hi @cpcloud , I removed the `Schema.from_tuples` as you suggested but some tests failed. It seems that `Schema.from_dict` didn't support empty schemas.

I fixed it in the last commit.

ibis/expr/api.py
```diff
@@ -283,11 +286,14 @@ def schema(
     ...              types=['string', 'int64', 'boolean'])
     """
     if pairs is not None:
-        return Schema.from_tuples(pairs)
+        return Schema.from_dict(dict(pairs))
     else:
         return Schema(names, types)
 
 
+_schema_func = schema
```
cpcloud, line 294:
```suggestion
_schema = schema
```

ibis/expr/api.py
```diff
@@ -306,10 +312,7 @@ def table(schema, name=None):
     table : TableExpr
     """
     if not isinstance(schema, Schema):
-        if isinstance(schema, dict):
-            schema = Schema.from_dict(schema)
-        else:
-            schema = Schema.from_tuples(schema)
+        schema = _schema_func(pairs=schema)
```
cpcloud, line 315:
```suggestion
        schema = _schema(pairs=schema)
```

cpcloud (changes requested):
Small change from `_schema_func -> _schema` then this is good to go! Thanks for digging into this!

EdAbati:
Sorry I had to revert the last commit and force push. I used `cz` and named the commit `refactor: renamed _schema_func to _schema`. After I pushed, I had only the commit in my fork but it wasn't showing in this PR.
I am not super familiar with `cz`, did I make a mistake? Should it just be used for the message of the 1st commit of a branch?

cpcloud:
@EdAbati No worries!

`cz` doesn't need be used for every commit. However, in the end you'll need to squash to 1 commit and it needs to be used for that one.

With that in mind, would you mind squashing and fixing up the final commit? Will land after that!

EdAbati:
Hi @cpcloud , after a few trials I now squashed everything into 1 commit. Let me know if I missed something. Thanks

cpcloud (approved):
Awesome! Thank @EdAbati!

PR #3376, opened 2022-01-29
refactor(config): replace config with pydantic

cpcloud:
This PR gets rid of our old config code that is not maintained and for which better solutions exist nowadays. I chose pydantic over everett and dynaconf for its simplicity, because it is a lightweight dependency and because it has the ability to validate on assignment, which was critical for preserving backwards compatibility.


jreback (approved):
looks fine to me. @gerrymanoim should look as well

cpcloud:
@gerrymanoim Would you mind reviewing this today?

gerrymanoim:
I know dask and pandas both use set_option inside their trace.py when enabling those functions, which is deprecated in this PR. 

I think the dask backend needs a similar change as the pandas backend to register enable_trace, but I'll confirm when I'm next at the computer.

cpcloud:
> I know dask and pandas both use set_option inside their trace.py when enabling those functions, which is deprecated in this PR.
> 
> I think the dask backend needs a similar change as the pandas backend to register enable_trace, but I'll confirm when I'm next at the computer.

The config is shared from `BasePandasBackend`.

gerrymanoim:
LGTM. We should change https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/trace.py#L74 and https://github.com/ibis-project/ibis/blob/master/ibis/backends/pandas/trace.py#L88 but can do as a followup.

gerrymanoim (approved):

PR #3387, opened 2022-02-01
refactor: move coercion functions from schema.py to udf

### LINKED ISSUE #3033: refactor: move the coercion functions from schema.py to udf.py
[These](https://github.com/ibis-project/ibis/blob/master/ibis/expr/schema.py#L244-L391) are data manipulating routines related to UDFs.

EdAbati:
fixes #3033

I have not found `udf.py` so for now I moved the functions to `udf/vectorized.py`, where most of them are used. Please let me know if you think I should move them somewhere else. Thanks:)


cpcloud:
@kszucs Can you review this?

kszucs:
Looks good to me, though we need to mark this change as a breaking one.

cpcloud:
@kszucs Are these functions really part of ibis's public API? I realize that "everything" in Python is public, but these functions seem very much like they shouldn't guarantee any stability since they appear to be utility functions.

jreback:
these shouldn't be part of the public api. could consider making things under `udf` be _ modules

cpcloud:
I'm going to add the `BREAKING CHANGE` bit to the commit message and merge.

@EdAbati Would you mind doing a follow up to make these functions private by prefixing their name with an underscore?

EdAbati:
Hi @cpcloud , yes it sounds good!

cpcloud:
Awesome, merging this on green! Thanks!

cpcloud (approved):

PR #3394, opened 2022-02-03
docs: add better comparison section and fix broken links

cpcloud:
- docs: add better comparison section
- chore: fix links


cpcloud:
Here's the rendered docs:

image

image

PR #3395, opened 2022-02-03
refactor: split ibis/expr/types.py into separate modules

cpcloud:


jreback (approved):
looks like this keeps the same namespacing from a user api POV so +1

cpcloud:
@jreback Mind officially approving?

cpcloud:
Ah I see you did! Thanks

PR #3396, opened 2022-02-03
fix(util): quote_identifier backslash escape quotechar

### LINKED ISSUE #488: bug: backticks in quoted identifiers are not escaped

saulpw:
closes #488


saulpw:
No one seems to need this after all.

PR #3401, opened 2022-02-04
chore: fix setup.py gen

cpcloud:
Looks like this slipped through from #3395


cpcloud:
Merging on green to get back to green `master`.

PR #3414, opened 2022-02-07
refactor: remove the csv backend

cpcloud:
The CSV backend is not providing much value over other backends that support CSVs.

BREAKING CHANGE: The CSV backend is removed. Use Datafusion for CSV execution.


cpcloud:
@jreback Thoughts?

jreback (approved):

PR #3417, opened 2022-02-07
fix: disconnect from clickhouse when computing version

cpcloud:
Tests starting failing for `clickhouse_driver==0.2.3` (showing up after running `poetry update`). This PR fixes the underlying cause.


cpcloud:
Merging on green.

PR #3425, opened 2022-02-09
Pw22 date from ymd

### LINKED ISSUE #386: feat: expression for creating timestamp from component fields
This may need to be implemented in Impala or as an add-on UDF

```
as_timestamp(int year, int month, int day)
as_timestamp(int year, int month, int day, int hours, int minutes, int seconds, int milliseconds, int microseconds)
```

saulpw:
Fixes #386.

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -506,6 +506,11 @@ sqlalchemy_operation_registry: Dict[Any, Any] = {
     # other
     ops.SortKey: _sort_key,
     ops.Date: unary(lambda arg: sa.cast(arg, sa.DATE)),
+    ops.DateFromYMD: fixed_arity(sa.func.date, 3),
+    ops.TimeFromHMS: fixed_arity(sa.func.time, 3),
+    ops.TimestampFromYMDHMS: lambda t, expr: sa.func.make_timestamp(
+        *map(t.translate, expr.op().args[:6])  # ignore timezone
+    ),
 }
 
 
--- a/ibis/backends/tests/test_temporal.py
+++ b/ibis/backends/tests/test_temporal.py
@@ -622,3 +622,75 @@ def test_now_from_projection(backend, alltypes):
     now = pd.Timestamp('now')
     year_expected = pd.Series([now.year] * n, name='ts')
     tm.assert_series_equal(ts.dt.year, year_expected)
+
+
+@pytest.mark.notimpl(["pandas", "datafusion", "mysql", "dask", "pyspark"])
+@pytest.mark.notyet(["clickhouse", "impala"])
+def test_date_literal(con):
+    expr = ibis.date(2022, 2, 4)
+    result = con.execute(expr)
+    assert result.strftime('%Y-%m-%d') == '2022-02-04'
+
+
+@pytest.mark.notimpl(["pandas", "datafusion", "mysql", "dask", "pyspark"])
+@pytest.mark.notyet(["clickhouse", "impala"])
+def test_timestamp_literal(con):
+    expr = ibis.timestamp(2022, 2, 4, 16, 20, 0)
+    result = con.execute(expr)
+    if not isinstance(result, str):
+        result = result.strftime('%Y-%m-%d %H:%M:%S%Z')
+    assert result == '2022-02-04 16:20:00'
+
+
+@pytest.mark.notimpl(["pandas", "datafusion", "mysql", "dask", "pyspark"])
+@pytest.mark.notyet(["clickhouse", "impala"])
+def test_time_literal(con):
+    expr = ibis.time(16, 20, 0)
+    result = con.execute(expr)
+    if not isinstance(result, str):
+        result = result.strftime('%H:%M:%S')
+    assert result == '16:20:00'
+
+
+@pytest.mark.notimpl(["pandas", "datafusion", "mysql", "dask", "pyspark"])
+@pytest.mark.notyet(["clickhouse", "impala"])
+def test_date_column_from_ymd(con, alltypes, df):
+    c = alltypes.timestamp_col
+    expr = ibis.date(c.year(), c.month(), c.day())
+    tbl = alltypes[
+        expr.name('timestamp_col'),
+    ]
+    result = con.execute(tbl)
+
+    golden = df.timestamp_col.dt.date.astype('datetime64[ns]')
+    tm.assert_series_equal(golden, result.timestamp_col)
+
+
+@pytest.mark.notimpl(["datafusion", "impala"])
+def test_date_scalar_from_iso(con):
+    expr = ibis.literal('2022-02-24')
+    expr2 = ibis.date(expr)
+
+    result = con.execute(expr2)
+    assert result.strftime('%Y-%m-%d') == '2022-02-24'
+
+
+@pytest.mark.notimpl(["datafusion", "impala", "pyspark"])
+def test_date_column_from_iso(con, alltypes, df):
+    expr = (
+        alltypes.year.cast('string')
+        + '-'
+        + alltypes.month.cast('string').lpad(2, '0')
+        + '-13'
+    )
+    expr = ibis.date(expr)
+
+    result = con.execute(expr)
+    golden = (
+        df.year.astype(str)
+        + '-'
+        + df.month.astype(str).str.rjust(2, '0')
+        + '-13'
+    )
+    actual = result.dt.strftime('%Y-%m-%d')
+    tm.assert_series_equal(golden.rename('tmp'), actual.rename('tmp'))
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -3,10 +3,11 @@
 from __future__ import annotations
 
 import datetime
-import numbers
+import functools
 from typing import Iterable, Mapping, Sequence, TypeVar
 
 import dateutil.parser
+import numpy as np
 import pandas as pd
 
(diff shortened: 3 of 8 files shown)


ibis/expr/operations/temporal.py
```diff
@@ -213,6 +213,45 @@ class Date(UnaryOp):
...
+    month = rlz.integer
+    day = rlz.integer
+    hours = rlz.integer
+    minutes = rlz.integer
+    seconds = rlz.integer
+    timezone = rlz.optional(rlz.string)
+    output_type = rlz.shape_like('args', dt.timestamp)
+
+    @property
+    def arg(self):
```
cpcloud, line 244: I don't think this is necessary. You should be able to access everything as `.args[:-1]`, or preferably as each individual component so the usage of the operations arguments isn't sensitive to argument declaration order.

ibis/expr/api.py
```diff
@@ -389,23 +392,53 @@ def timestamp(
...
+@timestamp.register(np.int8)
+@timestamp.register(np.int16)
+@timestamp.register(np.int32)
+@timestamp.register(np.int64)
+@timestamp.register(np.uint8)
+@timestamp.register(np.uint16)
+@timestamp.register(np.uint32)
+@timestamp.register(np.uint64)
+@timestamp.register(np.float32)
+@timestamp.register(np.float64)
```
cpcloud, line 407:
```suggestion
@timestamp.register(np.integer)
@timestamp.register(np.floating)
```

ibis/backends/tests/test_temporal.py
```diff
@@ -622,3 +622,75 @@ def test_now_from_projection(backend, alltypes):
     now = pd.Timestamp('now')
     year_expected = pd.Series([now.year] * n, name='ts')
     tm.assert_series_equal(ts.dt.year, year_expected)
+
+
+@pytest.mark.notimpl(["pandas", "datafusion", "mysql", "dask", "pyspark"])
+@pytest.mark.notyet(["clickhouse", "impala"])
+def test_date_literal_ex(con):
```
cpcloud, line 629:
```suggestion
def test_date_literal(con):
```

ibis/backends/tests/test_temporal.py
```diff
@@ -622,3 +622,75 @@ def test_now_from_projection(backend, alltypes):
...
+    expr = ibis.date(2022, 2, 4)
+    result = con.execute(expr)
+    assert result.strftime('%Y-%m-%d') == '2022-02-04'
+
+
+@pytest.mark.notimpl(["pandas", "datafusion", "mysql", "dask", "pyspark"])
+@pytest.mark.notyet(["clickhouse", "impala"])
+def test_timestamp_literal_ex(con):
```
cpcloud, line 637:
```suggestion
def test_timestamp_literal(con):
```

ibis/backends/tests/test_temporal.py
```diff
@@ -622,3 +622,75 @@ def test_now_from_projection(backend, alltypes):
...
+    expr = ibis.timestamp(2022, 2, 4, 16, 20, 0)
+    result = con.execute(expr)
+    if not isinstance(result, str):
+        result = result.strftime('%Y-%m-%d %H:%M:%S%Z')
+    assert result == '2022-02-04 16:20:00'
+
+
+@pytest.mark.notimpl(["pandas", "datafusion", "mysql", "dask", "pyspark"])
+@pytest.mark.notyet(["clickhouse", "impala"])
+def test_time_literal_ex(con):
```
cpcloud, line 647:
```suggestion
def test_time_literal(con):
```

ibis/tests/expr/test_temporal.py
```diff
@@ -705,3 +705,9 @@ def test_time_truncate(table, operand, unit):
     expr = operand(table).truncate(unit)
     assert isinstance(expr, ir.TimeValue)
     assert isinstance(expr.op(), ops.TimeTruncate)
+
+
+def test_date_time_literals():
+    ibis.date(2022, 2, 4)
+    ibis.time(16, 20, 00)
+    ibis.timestamp(2022, 2, 4, 16, 20, 00)
```
cpcloud, line 713:
```suggestion
    assert ibis.timestamp(2022, 2, 4, 16, 20, 00).type() == dt.timestamp
```

I think the important thing here is to test that the data type of the resulting expression is inferred correctly.

You may need to add `import ibis.expr.datatypes as dt`.

You may also need to run `black` and `isort` again after these changes.

ibis/tests/expr/test_temporal.py
```diff
@@ -705,3 +705,9 @@ def test_time_truncate(table, operand, unit):
     expr = operand(table).truncate(unit)
     assert isinstance(expr, ir.TimeValue)
     assert isinstance(expr.op(), ops.TimeTruncate)
+
+
+def test_date_time_literals():
+    ibis.date(2022, 2, 4)
+    ibis.time(16, 20, 00)
```
cpcloud, line 712:
```suggestion
    assert ibis.time(16, 20, 00).type() == dt.time
```

ibis/tests/expr/test_temporal.py
```diff
@@ -705,3 +705,9 @@ def test_time_truncate(table, operand, unit):
     expr = operand(table).truncate(unit)
     assert isinstance(expr, ir.TimeValue)
     assert isinstance(expr.op(), ops.TimeTruncate)
+
+
+def test_date_time_literals():
+    ibis.date(2022, 2, 4)
```
cpcloud, line 711:
```suggestion
    assert ibis.date(2022, 2, 4).type() == dt.date
```

ibis/backends/duckdb/registry.py
```diff
@@ -160,6 +160,7 @@ def _regex_extract(t, expr):
         ops.TableColumn: _table_column,
         ops.TimestampDiff: fixed_arity('age', 2),
         ops.TimestampFromUNIX: _timestamp_from_unix,
+        ops.DateFromYMD: fixed_arity('make_date', 3),
```
cpcloud, line 163:
```suggestion
```

You should be able to remove this since DuckDB will pick up the rules from the postgres backend.

ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -506,6 +506,9 @@ def _string_join(t, expr):
     # other
     ops.SortKey: _sort_key,
     ops.Date: unary(lambda arg: sa.cast(arg, sa.DATE)),
+    ops.DateFromYMD: fixed_arity(sa.func.date, 3),
+    ops.TimeFromHMS: varargs(sa.func.time),
+    ops.TimestampFromYMDHMS: varargs(sa.func.make_timestamp),
```
cpcloud, line 511: Are these functions actually variadic?

cpcloud (commented):
Looking good! Few comments.

ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -489,6 +489,7 @@ def _sort_key(t, expr):
     ops.FloorDivide: _floor_divide,
     # other
     ops.SortKey: _sort_key,
+    ops.DateFromYMD: sa.func.date,
```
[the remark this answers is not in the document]
saulpw: Hm, weird, this is already the case, I'm not sure why it's not showing up in the diff..

ibis/expr/operations/temporal.py
saulpw: `varargs(...)` requires an `.arg`.  Changing those to `fixed_arity(..., 6)` doesn't work because of the extra timezone field (unused currently but seemed important to include in the API, and then to capture in the op).  I got everything to work this way with the `arg` property; is this actually incorrect?

ibis/expr/api.py
saulpw: I looked for this and couldn't find it!  Thanks.

ibis/expr/operations/temporal.py
saulpw: How do you feel about changing `fixed_arity` to trim the number of args instead of raising an error?  Or making a similar helper?  I could do it inline but that seems too painful.

ibis/backends/base/sql/alchemy/registry.py
cpcloud: We should probably be more explicit in our developer guide here, but generally I think we want to avoid implicit behavior like that and even though it's more annoying to write out the explicit translation rule, it's also less error prone and has no implicit behavior.

That said, this PR has been open for a while and I'm okay with this bit as is.

ibis/expr/operations/temporal.py
cpcloud: I'd like to avoid giving `fixed_arity` any convenience functionality. It's only there to make the simplest cases convenient, and everything else has to be explicit.

cpcloud: You can also write a translation rule and remove the `.arg` property and also avoid having to change `fixed_arity` or add a similar helper:

```python
def rule(t, expr):
	return sa.func.make_timestamp(*map(t.translate, expr.op().args[:-1]))
```

ibis/backends/base/sql/alchemy/registry.py
cpcloud: Sorry, I just realized that 

> I'm okay with this bit as is

conflicts with what I wrote above. Would you mind implementing the rule directly? That's should be the final thing and then I'll merge it!

cpcloud (approved):
Awesome!

PR #3428, opened 2022-02-09
test: remove dead fixtures

cpcloud:
- chore: remove unused fixture
- ci: add dead fixtures test
- chore(dev-deps): add pytest-deadfixtures
- test: remove dead fixtures

PR #3429, opened 2022-02-09
test: dynamic fixtures act on backend-specific tests

gforsyth:
`conftest.py` files get run for child directories, so we can re-use the
backend xfails in backend specific tests if they are defined in a parent
directory.

So I've moved the contents of `backends/tests/conftest.py` into
`backends/conftest.py` to accomplish this.

I've also added in checks to make sure we don't inadvertantly run a test
with multiple backends simultaneously (this can happen if a
backend-specific con is included alongside a general backend, for instance).

Also require that `backends_never` usage provide a `reason`


gforsyth:
After #3428 is in I can rebase this and then remove some more backend-specific fixture bits that we don't need now that `csv`, `parquet` and `hdf5` backends are gone.

cpcloud:
It may make sense to land this before #3428, because it looks like I'll have to package up `pytest-deadfixtures` for `conda-forge`.

cpcloud (approved):

cpcloud:
Sweet PR, thanks!

PR #3432, opened 2022-02-10
perf: use assign instead of concat in projections when possible

gerrymanoim:
Right now we build up projections by creating pieces and then concating them all together (following the pandas backend). This is a fairly inefficient operation in dask and some local testing shows that dask will struggle to optimize operations between the pieces you're concating (TODO elaborate). 

This change assigns into the existing frame, reducing operations and speeding things up.

### DIFF
--- a/ibis/backends/dask/execution/selection.py
+++ b/ibis/backends/dask/execution/selection.py
@@ -29,6 +29,7 @@ from ..execution.util import (
     add_partitioned_sorted_column,
     coerce_to_output,
     compute_sorted_frame,
+    is_row_order_preserving,
 )
 
 
@@ -130,7 +131,21 @@ def build_df_from_projection(
     Build up a df from individual pieces by dispatching to `compute_projection`
     for each expression.
     """
+    # Fast path for when we're assigning columns into the same table.
+    if (selection_exprs[0] is op.table) and all(
+        is_row_order_preserving(selection_exprs[1:])
+    ):
+        for expr in selection_exprs[1:]:
+            projection = compute_projection(expr, op, data, **kwargs)
+            if isinstance(projection, dd.Series):
+                data = data.assign(**{projection.name: projection})
+            else:
+                data = data.assign(
+                    **{c: projection[c] for c in projection.columns}
+                )
+        return data
 
+    # Slow path when we cannot do direct assigns
     # Create a unique row identifier and set it as the index. This is
     # used in dd.concat to merge the pieces back together.
     data = add_partitioned_sorted_column(data)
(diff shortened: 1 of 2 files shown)


ibis/backends/dask/execution/util.py
```diff
@@ -436,3 +437,20 @@ def helper(
     df = df.set_index(col_name, sorted=True, divisions=divisions)
 
     return df
+
+
+def is_order_preserving(expr: ir.Expr) -> bool:
+    """Detects if the operation preserves ordering."""
+
+    def _is_order_preserving(op):
```
cpcloud, line 445: You might be able to use `ibis.expr.lineage.traverse` here to handle the recursion and iteration.

gerrymanoim: I actually borrowed this from https://github.com/ibis-project/ibis/blob/master/ibis/expr/analysis.py#L1192. I'll look into seeing if `traverse` is cleaner.

gerrymanoim:
Will rebase/squash when this is ready for code review/I'm actually at my home computer.

ibis/backends/dask/execution/selection.py
```diff
@@ -130,7 +131,20 @@ def build_df_from_projection(
     Build up a df from individual pieces by dispatching to `compute_projection`
     for each expression.
     """
+    # Fast path for when we're assigning columns into the same table.
+    if (selection_exprs[0] is op.table) and all(
```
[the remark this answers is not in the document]
gerrymanoim: I believe so. `selection_exprs` is `selections = op.selections` and the validator for `op.Selection` guarantees this is a `list`. The `if` check in `execute_selection_dataframe` guarantees that there's at least one element.

cpcloud:
@gerrymanoim In the final squash can you give this a `perf:` prefix instead of `feat`? This looks like a performance improvement rather than a new feature.

ibis/backends/dask/execution/selection.py
```diff
@@ -130,7 +131,21 @@ def build_df_from_projection(
...
+        is_row_order_preserving(selection_exprs[1:])
+    ):
+        for expr in selection_exprs[1:]:
+            projection = compute_projection(expr, op, data, **kwargs)
+            if isinstance(projection, dd.Series):
+                data = data.assign(**{projection.name: projection})
+            else:
+                data = data.assign(
+                    **{c: projection[c] for c in projection.columns}
```
jreback, line 144: it might be better if we have a *lot* of columns to use the concat path, though would require a heurestic / magic number e.g. say > 100 or something

jreback (approved):
small comment but lgtm

ibis/backends/dask/execution/selection.py
cpcloud: Seems like a good time to kick the tires on the new config stuff!

cpcloud: This can be a follow up if it ever becomes an issue!

PR #3434, opened 2022-02-11
test: bring back backend testing markers

cpcloud:
This PR brings back the very-convenient use of pytest markers for backends. Additionally, when new test files are added to backends they are automatically picked up by a pytest collection hook, making it easy for contributors to add new tests without forgetting to add pytest.mark.${backend}


.github/workflows/ibis-backends.yml
```diff
@@ -393,18 +337,25 @@ jobs:
       - run: python -m pip install --upgrade pip poetry
 
       - name: install ibis
+        if: ${{ matrix.backend.name != 'postgres' }}
```
kszucs, line 340: Seems like there are extras defined for postgres as well https://github.com/ibis-project/ibis/blob/master/pyproject.toml#L109

.github/workflows/check-setup-py.yml
```diff
@@ -87,8 +87,8 @@ jobs:
       - name: install ibis in development mode
         run: pip install -e .
 
-      - name: run simple tests
-        run: pytest ibis/tests --numprocesses auto
+      - name: run tests
+        run: pytest -m 'not backend' --numprocesses auto
```
kszucs, line 91: I wonder, could we mark `not backend` tests implicitly with its own marker, so we could run `pytest -m ibis` or `pytest -m core` instead of `pytest -m 'not backend'`?

ibis/backends/clickhouse/tests/test_aggregations.py
```diff
@@ -7,6 +7,8 @@
 
 from ibis import literal as L
 
+pytest.importorskip("clickhouse_driver")
```
kszucs, line 10: Does pytest support importorskip in `clickhoust/conftest.py`?

ibis/backends/conftest.py
```diff
@@ -38,72 +40,111 @@ def data_directory() -> Path:
...
     """
-    return [
+    return frozenset(
         entry_point.name
         for entry_point in importlib_metadata.entry_points()["ibis.backends"]
-    ]
+    )
 
 
 def _backend_name_to_class(backend_str: str):
```
kszucs, line 64: Can we rewrite this to `_load_backend[_or_skip](name, data_directory)`?

ibis/backends/conftest.py
```diff
@@ -38,72 +40,111 @@ def data_directory() -> Path:
...
-    return f'__ibis_test_{suffix}_{util.guid()}'
+def _random_identifier(suffix: str) -> str:
+    return f"__ibis_test_{suffix}_{util.guid()}"
 
 
-def _get_all_backends() -> List[str]:
-    """
-    Return the list of known backend names.
+@lru_cache(maxsize=None)
+def _get_all_backends() -> frozenset[str]:
```
kszucs, line 48: Rename it to `_get_backend_names`?

ibis/backends/conftest.py
```diff
@@ -38,72 +40,111 @@ def data_directory() -> Path:
...
+    expr = _pytest.mark.expression.Expression.compile(mark_expr)
+    # we check the "backend" marker as well since if that's passed
+    # any file matching a backed should be skipped
+    keep = expr.evaluate(lambda s: s in (backend, "backend"))
+    return not keep
+
+
+def pytest_collection_modifyitems(session, config, items):
+    # add the backend marker to any tests are inside "ibis/backends"
+    all_backends = _get_all_backends()
```
kszucs, line 120: `_get_all_backends` is ambiguous for me, I'm not sure whether it is a backend name, module or conf class.

.github/workflows/ibis-backends.yml
```diff
@@ -393,18 +337,25 @@ jobs:
...
         run: poetry install --extras ${{ matrix.backend.name }}
 
+      - name: install ibis
+        if: ${{ matrix.backend.name == 'postgres' }}
+        run: poetry install --extras ${{ matrix.backend.name }} --extras geospatial
```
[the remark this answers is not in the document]
cpcloud: Maybe, though we don't have any other backends that support geospatial things. It seems like waiting until that happens to refactor this is fine.

ibis/backends/conftest.py
cpcloud: Changed to `_get_backend_conf`. I can do the `data_directory` move in a follow up.

ibis/backends/conftest.py
cpcloud: Done.

ibis/backends/conftest.py
cpcloud: Yep, changed to `_get_backend_names`.

.github/workflows/ibis-backends.yml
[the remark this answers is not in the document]
kszucs: Actually there is the https://github.com/omnisci/ibis-omniscidb backend supporting geospatial types, but maintained elsewhere.

kszucs (approved):
Much better testing experience! Thanks Phillip!

PR #3437, opened 2022-02-12
refactor(types api): more cohesive primitive types

saulpw:
- do not allow constructing value/column of "concrete" `Any` type
- add aliases for common pythonic types:
   - `str` (string)
   - `int` (int64)
   - `float` (float64; was float32)
   - `bytes` (binary)
   - `bool` (boolean)
- docprecate extraneous imprecise types:
   - `double` (use `float` or `float64`)
   - `halffloat` (use `float16`)
   - `boolean` (use `bool`)

BREAKING CHANGE: `float` may need to be changed to `float32` in user code for total compatibility.


ibis/backends/base/sql/alchemy/datatypes.py
```diff
@@ -38,8 +38,9 @@ def table_from_schema(name, meta, schema, database: Optional[str] = None):
     dt.String: sa.Text,
     dt.Decimal: sa.NUMERIC,
     # Mantissa-based
-    dt.Float: sa.REAL,
-    dt.Double: sa.FLOAT,
+    dt.Float16: sa.REAL,
+    dt.Float32: sa.REAL,
+    dt.Float64: sa.REAL,
```
cpcloud, line 43: `dt.Float64` maps to `sa.FLOAT`. `sa.REAL` is 32-bit floating point. Yeah, it's confusing, and taken from SQL 😑

ibis/backends/mysql/__init__.py
```diff
@@ -155,12 +155,12 @@ def table(self, name, database=None, schema=None):
...
 @dt.dtype.register((mysql.DOUBLE, mysql.REAL))
 def mysql_double(satype, nullable=True):
-    return dt.Double(nullable=nullable)
+    return dt.Float64(nullable=nullable)
 
 
 @dt.dtype.register(mysql.FLOAT)
 def mysql_float(satype, nullable=True):
-    return dt.Float(nullable=nullable)
+    return dt.Float64(nullable=nullable)
```
cpcloud, line 163: This should be `Float32`

cpcloud (changes requested):
Couple of small things around various annoying SQL things. Looks like some lint errors.

Otherwise looks great!

saulpw:
I think these lint errors are actually necessary, unless we want to specify all imported symbols from `.core` explicity (which seems like a waste).  Can we just nolint these errors in this file?

ibis/expr/datatypes/__init__.py
```diff
@@ -0,0 +1,13 @@
+# flake8: noqa
```
cpcloud, line 1: Can you mark the individual lines instead, with their specific ignore codes? `flake8: noqa` disables any and all linting for the file.

ibis/expr/rules.py
```diff
@@ -237,7 +237,10 @@ def value(dtype, arg, **kwargs):
         # coerce python literal to ibis literal
         arg = ir.literal(arg)
 
-    if not isinstance(arg, ir.AnyValue):
+    if not isinstance(arg, ir.AnyValue) or type(arg) in (
```
kszucs, line 240: Could you please restore the previous check so we can keep the previous behavior? Eventually we should use abstract base classes to prevent creating untyped value expressions.

cpcloud: I think there's value in preventing this kind of thing now, versus waiting until a refactor that may never happen. The current behavior of allowing `AnyValue`s is incorrect, until we have a valid use case for it.

saulpw: @kszucs Do you have any code that might rely on this behavior?  From what I understand, this is now basically incorrect and without this check the error message is just more incomprehensible.

kszucs: Fair point on the error message. 

My problem is that the actual type is determined by the underlying `datatype` and the `ir` types are merely shims to provide the python API. The value rules actually don't care about `ir` type, they check the `expr.type()` property which is a datatype, so `ir.AnyScalar(dt.int64)` would be valid according to the rules.

If we're not paranoid about the backward compatibility then it's fine to introduce this, though it requires a test case then.

saulpw: Thanks @kszucs, I think this next release will include several backward-incompatible changes so we don't need to be paranoid.  I'll add a test case to make sure the code is exercised.

ibis/backends/pandas/tests/test_datatypes.py
```diff
@@ -146,7 +144,7 @@ def test_pandas_dtype(pandas_dtype, ibis_dtype):
         (np.array([-5, 0, 12], dtype='int16'), 'int16'),
         (np.array([-12, 3, 25000], dtype='int32'), 'int32'),
         (np.array([102, 67228734, -0], dtype='int64'), 'int64'),
-        (np.array([45e-3, -0.4, 99.0], dtype='float32'), 'float'),
+        (np.array([45e-3, -0.4, 99.0], dtype='float64'), 'float64'),
```
cpcloud, line 147: This should remain `float32`, so that the test is testing the same thing.

ibis/backends/clickhouse/tests/test_functions.py
```diff
@@ -21,9 +21,9 @@
     [
         ('int8', 'CAST(`double_col` AS Int8)'),
         ('int16', 'CAST(`double_col` AS Int16)'),
-        ('float', 'CAST(`double_col` AS Float32)'),
+        ('float', 'CAST(`double_col` AS Float64)'),
```
cpcloud, line 24: Can we keep a test case for original behavior too?

ibis/backends/dask/tests/test_schema.py
```diff
@@ -146,8 +146,8 @@ def test_infer_exhaustive_dataframe(npartitions):
         ('bool_col', dt.boolean),
         ('bool_obj_col', dt.boolean),
         ('date_string_col', dt.string),
-        ('double_col', dt.double),
-        ('float_col', dt.float),
+        ('double_col', dt.float64),
+        ('float_col', dt.float64),
```
cpcloud, line 150: This should be `float32` as `float_col` is specifically for testing float32.

cpcloud (changes requested):
Very close! We just need to keep the float32 behavior tests as we had them, after that this is good to go!

cpcloud:
Since this is so close I'm just going to make the few very small remaining changes.

cpcloud (approved):

PR #3431, opened 2022-02-10
feat: duckdb backend

cpcloud:
This is a starting point for the duckdb backend. cc @gforsyth

PR #3448, opened 2022-02-14
feat: add duckdb backend

gforsyth:
Still very much a work-in-progress, building on @cpcloud 's start in #3431 

A lot of tests to xfail for now as `notimpl`, but a surprising amount of stuff just works.

~There's one major issue I'm still digging in to, which involves missing downcasts on certain ops (like `floor` and `ceil`)~ solved by using `schema.apply_to`

Added in a `registry.py` to override a few `postgres` behaviors that aren't compatible with duckdb -- i think the remaining work towards a pretty ok first-cut is to fix up the remaining string tests.


ci/datamgr.py
```diff
@@ -708,6 +708,47 @@ def all(ctx):
...
+        file_okay=False,
+        dir_okay=True,
+        writable=True,
+        readable=True,
+        path_type=Path,
+    ),
+)
+def duckdb(schema, tables, data_directory, database, **params):
+    try:
+        import duckdb  # noqa: F401
```
cpcloud, line 736: I think we can just leave this as `import duckdb`.

ibis/backends/duckdb/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+            import importlib_metadata
+        return importlib_metadata.version("duckdb")
+
+    def do_connect(
+        self, path: str | None = None, read_only: bool = False
+    ) -> None:
+        """Create an Ibis client connected to a DuckDB database."""
+
+        if path is None:
+            path = ":memory:"
```
cpcloud, line 71: I think we can just set the default value of `path` to `":memory:"` and type the `path` parameter as `path: str`.

ibis/backends/duckdb/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+        self.con = duckdb.connect(path, read_only=read_only)
+
+    def list_tables(self, like: str | None = None, **kwargs) -> list[str]:
+        """List the available tables."""
+        _tables = self.con.execute("PRAGMA show_tables;").fetchall()
```
cpcloud, line 76: micro-est of nits: I don't think the final `;` is necessary.

ibis/backends/duckdb/registry.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(self, string, pattern, index):
+        super().__init__(string, pattern, index)
+        self.string = string
+        self.pattern = pattern
+        self.index = index
+
+
+@compiles(_regex_extract, 'duckdb')
+def _compile_regex_extract(element, compiler, **kw):
+    result = '(SELECT * FROM REGEXP_MATCHES({}, {}))[{}]'.format(
```
cpcloud, line 52: Interesting. Is it necessary for this to be a scalar subquery? I would expect `REGEXP_MATCHES(...)[...]` to be sufficient.

ibis/backends/duckdb/registry.py
```diff
@@ -0,0 +1,118 @@
...
+                sa.func.regexp_matches(string, pattern),
+                sa.func._regex_extract(string, pattern, index + 1),
+            )
+        ],
+        else_='',
+    )
+    return result
+
+
+# def _find_in_set(t, expr):
```
cpcloud, line 74: This operation is a hack for systems that don't have proper list/array types support. It's probably fine to leave it out in v1 of this backend, or implement it as the duckdb equivalent of `','.split(haystack).index(needle)`.

pyproject.toml
```diff
@@ -49,6 +49,7 @@ pyspark = { version = ">=3,<4", optional = true }
 requests = { version = ">=2,<3", optional = true }
 Shapely = { version = ">=1.6,<1.8.1", optional = true }
 sqlalchemy = { version = ">=1.3,<1.5", optional = true }
+duckdb = { version = "^0.3.2", optional = true }
```
cpcloud, line 52:
```suggestion
duckdb = { version = ">=0.3.2,<1", optional = true }
```

This is so renovate (which isn't particularly smart about how it increases upper bounds) won't start adding `v1 || v2 || ... || vN` style version constraints.

ibis/expr/datatypes/__init__.py
```diff
@@ -1074,18 +1074,24 @@ def parser():
...
         | spaceless_string("uint8").result(uint8)
         | spaceless_string("uint16").result(uint16)
         | spaceless_string("uint32").result(uint32)
         | spaceless_string("uint64").result(uint64)
+        | spaceless_string("ubigint").result(uint64)
         | spaceless_string("halffloat").result(halffloat)
         | spaceless_string("double").result(double)
         | spaceless_string("float16").result(float16)
         | spaceless_string("float32").result(float32)
+        | spaceless_string("real").result(float32)
```
cpcloud, line 1094: I think I have an alternative type parser laying around in an older version my branch on this, I'll push it up.

ibis/backends/tests/test_aggregation.py
```diff
@@ -312,7 +323,7 @@ def test_approx_median(backend, alltypes, df):
         )
     ],
 )
-@mark.notimpl(["datafusion"])
+@mark.notimpl(["datafusion", "duckdb"])
```
cpcloud, line 326: This function should be expressible as `string_agg` (https://duckdb.org/docs/sql/aggregates).

ibis/backends/tests/test_array.py
```diff
@@ -36,7 +38,7 @@ def test_array_scalar(backend, con, alltypes, df):
 
 
 # Issues #2370
-@pytest.mark.notimpl(["impala", "datafusion"])
+@pytest.mark.notimpl(["impala", "datafusion", "duckdb"])
```
cpcloud, line 41: I think most of these should be expressible, I'll see if I can push up the implementations to get them passing.

ibis/backends/tests/test_generic.py
```diff
@@ -34,7 +34,15 @@ def test_fillna_nullif(backend, con, expr, expected):
 
 
 @pytest.mark.notimpl(
-    ["clickhouse", "datafusion", "impala", "mysql", "postgres", "sqlite"]
+    [
+        "clickhouse",
+        "datafusion",
+        "duckdb",
```
cpcloud, line 40: This just isn't possible in duckdb right now since you can't interact with non-finite values yet. This is slated to change in an upcoming release.

Should it be a `notyet`?

ibis/backends/tests/test_generic.py
```diff
@@ -52,7 +60,7 @@ def test_isna(backend, alltypes):
 
 
 @pytest.mark.notimpl(
-    ["clickhouse", "datafusion", "impala", "mysql", "postgres"]
+    ["clickhouse", "datafusion", "duckdb", "" "impala", "mysql", "postgres"]
```
cpcloud, line 63:
```suggestion
    ["clickhouse", "datafusion", "duckdb", "impala", "mysql", "postgres"]
```

ibis/backends/tests/test_join.py
```diff
@@ -20,6 +20,7 @@
                 [
                     "clickhouse",
                     "datafusion",
+                    "duckdb",
```
cpcloud, line 23: Most of the backends are broken for this test. I'll have a PR for this today.

ibis/backends/tests/test_numeric.py
```diff
@@ -507,7 +510,15 @@ def test_random(con):
...
 @pytest.mark.notimpl(
-    ["sqlite", "clickhouse", "datafusion", "impala", "mysql", "postgres"]
+    [
+        "clickhouse",
+        "duckdb",
+        "datafusion",
+        "impala",
+        "mysql",
+        "postgres",
+        "sqlite",
```
cpcloud, line 520: I'm not sure why all these are missing a clip implementation, but separately we should implement that.

ibis/backends/tests/test_param.py
```diff
@@ -55,7 +55,16 @@ def test_timestamp_accepts_date_literals(backend, alltypes):
 
 
 @pytest.mark.notimpl(
-    ["dask", "datafusion", "impala", "mysql", "pandas", "pyspark", "sqlite"]
+    [
+        "dask",
+        "datafusion",
+        "duckdb",
```
cpcloud, line 61: This is some that I think should be supported in v1 of the backend.

ibis/backends/tests/test_union.py
```diff
@@ -23,7 +23,7 @@ def test_union(backend, alltypes, df, distinct):
         param(
             True,
             marks=pytest.mark.notimpl(
-                ["postgres", "pyspark"],
+                ["duckdb", "postgres", "pyspark"],
```
cpcloud, line 26: Hm, this seems like it should be supported out of the box.

cpcloud (changes requested):

ibis/backends/duckdb/registry.py
gforsyth: Ah, I don't think that's working currently -- I was fighting with some of the regexp stuff and trying things out.

ibis/backends/tests/test_union.py
gforsyth: I think the only bit that isn't supported here is guaranteed ordering with a distinct

ibis/expr/datatypes/__init__.py
gforsyth: ah, cool.  Ripping this out into a separate parser was on my todo list, but if it's already done, that's :man_cook: :kiss:

ibis/backends/duckdb/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+
+from .compiler import DuckDBSQLCompiler
+
+
+class Backend(BaseAlchemyBackend):
+    name = "duckdb"
+    compiler = DuckDBSQLCompiler
+
+    def current_database(self):
+        raise NotImplementedError
```
cpcloud, line 22: I suppose this could be `"main"`, since duckdb doesn't support `ATTACH` yet.

ibis/backends/duckdb/registry.py
```diff
@@ -0,0 +1,126 @@
...
+    {
+        ops.Round: _round,
+        ops.Log2: unary(sa.func.log2),
+        ops.Modulus: _mod,
+        ops.Log: _log,
+        ops.Translate: fixed_arity('replace', 3),
+        ops.DayOfWeekName: unary(sa.func.dayname),
+        ops.TimestampFromUNIX: _timestamp_from_unix,
+        ops.TimestampDiff: fixed_arity('age', 2),
+        #  ops.TimestampDiff: _timestamp_diff,
```
cpcloud, line 119: This can be deleted I think.

cpcloud (approved):
LGTM, great work @gforsyth!

cpcloud:
@gforsyth Can squash down to as small as makes sense? Then I'll merge!

cpcloud:
Merging this before any more CI gremlins try to gum up the works.

PR #3449, opened 2022-02-14
fix: change TimestampType to Timestamp

gerrymanoim:
In datafusion backend. 

I believe this is the correct type as I can't find `TimestampType` in `ibis.expr.datatypes`.


cpcloud:
@gerrymanoim can you change the commit type to `fix`. I'll merge it after that.

cpcloud (approved):
LGTM, thanks for the hotfix!

PR #3454, opened 2022-02-15
refactor: make coerce functions in ibis.udf.vectorized private

EdAbati:
As mentioned by @cpcloud in https://github.com/ibis-project/ibis/pull/3387#issuecomment-1031342083, I made the `coerce_` functions private.
I am a bit unsure about the `_coerce_to_dataframe`, does it need to be private even if it is imported in another module?

Please let me know if I should change some other things. Thanks :)


cpcloud (approved):

cpcloud:
Thanks @EdAbati! Setting to automerge on green.

PR #3458, opened 2022-02-16
docs: give the landing page a facelift

cpcloud:
This PR reworks the landing page to add more visual callouts and examples
of what ibis and why it's useful.


cpcloud:
Here's the rendered page. Note that the terminal at the top is an `asciinema` recording and isn't static.

image

PR #3463, opened 2022-02-17
refactor(hdfs): replace HDFS with `fsspec`

cpcloud:
Ibis's HDFS-related classes are replaced with `fsspec` APIs.

BREAKING CHANGE: Use `fsspec` instead of HDFS from ibis


kszucs:
@cpcloud needs a rebase

cpcloud:
@kszucs Done

cpcloud:
> FYI you might want to map "hdfs" to "arrow_hdfs" when passing the protocol name to `fsspec.filesystem(..)`. fsspec is still using the old deprecated `pyarrow.filesystem` by default, and "arrow_hdfs" will use a new implementation based on `pyarrow.fs`

I don't think so, because hdfs currently maps to webhdfs, and arrow_hdfs appears to be missing a number of operations like chmod.

jorisvandenbossche:
If it is missing functionality, it would be useful to bring that up at https://github.com/fsspec/filesystem_spec then, as I am planning to remove the deprecated implementation, and so the current fsspec default hdfs filesystem will stop working. 
(although for `chmod`, I am not sure that is currently possible when wrapping `pyarrow.fs`, so that might also be an Arrow JIRA)

Now, if you are only using webhdfs (which doesn't use arrow, neither old or new filesystems), then you can ignore my comment :)

cpcloud:
Yeah, I tried using `arrow_hdfs`, and while it did seem faster it requires libhdfs and a JDK, which seems really overkill for ibis's use cases at the moment.

PR #3470, opened 2022-02-18
ci(impala): speed up data loading by ~2-3x

cpcloud:


cpcloud:
The 3x was observed locally. 2x appears to be the case in CI.

ci/datamgr.py
```diff
@@ -645,38 +617,84 @@ def impala(data_dir):
...
             user=env.hdfs_user,
         ),
-        pool_size=CPU_COUNT,
+        pool_size=URLLIB_DEFAULT_POOL_SIZE,
     )
 
-    # validate our environment before performing possibly expensive operations
-    impala_raise_if_cannot_write_to_hdfs(con, env)
+    fs = fsspec.filesystem("file")
+    data_files = fs.find(data_dir)
```
kszucs, line 624: Could you filter the data files here so the set comprehension get simpler below?

ci/datamgr.py
```diff
@@ -645,38 +617,84 @@ def impala(data_dir):
...
 
-    # build and upload the UDFs
-    impala_build_udfs()
-    impala_upload_udfs(con, env)
+    hdfs = con.hdfs
+    # make the database and build and upload UDFs and data files
+    for future in concurrent.futures.as_completed(
+        {
+            executor.submit(impala_create_test_database, con, env),
+            *impala_build_and_upload_udfs(hdfs, env, fs=fs, executor=executor),
```
kszucs, line 633: Could you return the `(callback, args)` from `impala_build_and_upload_udfs` and call `executor.submit()` on those tasks instead?

ci/datamgr.py
```diff
@@ -645,38 +617,84 @@ def impala(data_dir):
...
+    executor = ctx.obj["executor"]
 
-    # build and upload the UDFs
-    impala_build_udfs()
-    impala_upload_udfs(con, env)
+    hdfs = con.hdfs
+    # make the database and build and upload UDFs and data files
+    for future in concurrent.futures.as_completed(
+        {
+            executor.submit(impala_create_test_database, con, env),
```
kszucs, line 632: It would be more readable if you'd construct the tasks before the loop with a couple of supplemental comments.

kszucs:
Besides a couple of code organization thoughts it looks good to me.

ci/datamgr.py
```diff
@@ -214,47 +195,34 @@ def impala_create_tables(con, env, *, executor=None):
...
+    hdfs.mkdir(bitcode_dir, create_parents=True)
 
-def load_impala_data(con, data_dir, env):
-    impala_upload_ibis_test_data_to_hdfs(con, data_dir, env)
-    impala_create_test_database(con, env)
-    with concurrent.futures.ThreadPoolExecutor(
-        max_workers=CPU_COUNT
-    ) as executor:
-        table_futures = (
-            executor.submit(table_future.result().compute_stats)
```
kszucs, line 251: Was this the main culprit?

.github/workflows/ibis-backends.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         run: just download-data
 
       - name: install backend data
-        run: just load-data ${{ matrix.backend.name }}
+        run: poetry run python ci/datamgr.py load ${{ matrix.backend.name }}
```
kszucs, line 79: What's our preference here?

ci/datamgr.py
cpcloud: No, the main culprit was the fact that recursive copy of a directory from the client to HDFS was happening in series, rather than creating a task per file and allowing concurrent IO.

cpcloud: `fsspec` supports async file system operations, but the `hdfs` filesystem doesn't implement it.

.github/workflows/ibis-backends.yml
cpcloud: `poetry run` is required in CI, but not for local dev (but even that depends!). I don't know if it's straightforward to make one command work for both. I can look into it, but the short answer for now is: prefer explicit `poetry run` in CI (`ci-check` notwithstanding) and use `just` for local dev.

kszucs: Thanks! Asked because we have a mixture in the ymls.

ci/datamgr.py
```diff
@@ -645,38 +617,95 @@ def impala(data_dir):
...
-    # load the data files
-    load_impala_data(con, str(data_dir), env)
+    executor = ctx.obj["executor"]
 
-    # build and upload the UDFs
-    impala_build_udfs()
-    impala_upload_udfs(con, env)
+    hdfs = con.hdfs
+    tasks = {
+        # make the database
```
kszucs, line 642: Thanks, much easier to follow!

kszucs (approved):

PR #3483, opened 2022-02-22
Removing easy install which is depreciated 

marlenezw:
Hi! 

I went through the contributing guide using miniconda and tried to install the ibis dependencies with easy install. It turns out it was [depreciated](https://setuptools.pypa.io/en/latest/deprecated/easy_install.html). Running `pip install -e` or `pip3 install -e` works fine though!


marlenezw:
Closing this PR made a mistake with CZ and I think its causing the linter to fail. I'll make a new one!

PR #3487, opened 2022-02-22
docs(workflow.md): adding to the cz install tip

marlenezw:
Adding to the cz install tip in case users don't know the package name, don't have it  and want to install it.

### DIFF
--- a/docs/contribute/workflow.md
+++ b/docs/contribute/workflow.md
@@ -77,7 +77,7 @@ Setting `PYTEST_BACKENDS` unless you want to run the backend test suite
 !!! tip
 
     `cz` is already installed in your environment if you followed the [setup
-    instructions](environment.md)
+    instructions](environment.md) If not install it using `pip install commitizen`
 
 Next, you'll want to commit your changes.
 


marlenezw:
Sorry for the very small commits 😅  just trying to get used to the workflow! I'll add all the docs suggestions I have in the next PR I make!

cpcloud:
> Sorry for the very small commits sweat_smile just trying to get used to the workflow! I'll add all the docs suggestions I have in the next PR I make!

No worries! A smaller change can typically be merged much faster. Much appreciated.

docs/contribute/workflow.md
```diff
@@ -77,7 +77,7 @@ Setting `PYTEST_BACKENDS` unless you want to run the backend test suite
 !!! tip
 
     `cz` is already installed in your environment if you followed the [setup
-    instructions](environment.md)
+    instructions](environment.md) If not install it using `pip install commitizen`
```
cpcloud, line 80: There are probably a lot of cases where the alternative installation method is `pip install thing` if someone isn't following the development environment setup.

I don't think we should document this workflow because it's unlikely to be supported. We can't support every possible setup under the sun.

We document the existing installation methods (hopefully sufficiently clearly!) to encourage people to use those. Anything else is basically "you're on your own" 😬.

The existing conda and nix environments are well-supported, fast to setup, and are extremely likely to give someone a working environment capable of testing everything, including the impala backend.

I know there are others who don't want to use either conda or nix, which is fine, but I don't think we should try to document another setup, given that the current two systems provide all dependencies necessary for development.

marlenezw: Sounds good, I created my developer environment using conda and when I tried to run `cz commit` it didn't work for some reason, which is why I used pip to install it. The cz command wasn't recognized, but I'm wondering if the version of ibis I have is older and just needs updating. I might try rerun everything in case. Thanks for the help Phillip 😄

cpcloud: Ah, hm, I am seeing `commitizen` in our `conda-lock/*` files:

```
❯ grep -R commitizen conda-lock | head -1
conda-lock/win-64-3.8.lock:https://conda.anaconda.org/conda-forge/noarch/commitizen-2.21.0-pyhd8ed1ab_0.tar.bz2#7e360da7ba3a89a11becd87ae2f72502
```

Did you use one of those to create an environment?

PR #3488, opened 2022-02-22
feat: add optional rich repr to interactive mode

cpcloud:
This PR adds an optional `rich` dependency and implements a fancy display
of tables for interactive mode.


ibis/expr/types/core.py
```diff
@@ -48,7 +48,18 @@ def __repr__(self) -> str:
             )
             return output
         else:
-            return repr(result)
+            try:
+                from rich.console import Console
+            except ImportError:
```
jreback, line 53: i would encourage this to be a separate method (maybe called from here)

repr can be called in many places and if doing a lot of work can be a perf issue in graph construction 

eg repr should itself be very simple, but maybe options can control more advanaced things

jreback (changes requested):

ibis/expr/types/core.py
```diff
@@ -48,7 +48,18 @@ def __repr__(self) -> str:
             )
             return output
         else:
-            return repr(result)
+            try:
+                from rich.console import Console
```
jreback, line 52: also i would generally put these imports at module level (and protect them) so calling this w/o the import won't fail repeatedly by trying to import (which can be expensive as well)

cpcloud: A failed import is even cheaper

jreback: sure but again this can be a hidden performance issue in a large graph (imagine it fails every time then it's extremely expensive)

jreback: no maybe we have a special case but the issue we have seen is basically this

a try/except in some operations will format an exception as part of the except operation (but maybe just is flow control)

since repr is invoked there (and could be very expensive) graph construction in some edge cases can be very large cost

it doesn't happen often but it can be very hard to do

so i think strategically repr should be very cheap and other display modes should be intentional

cpcloud: True, but is that different from the existing code? I'm not saying that what you're describing doesn't happen, only that this code has no effect on what you're describing.

ibis/expr/types/core.py
cpcloud: The imports have been moved to toplevel.

PR #3492, opened 2022-02-22
test: cleanup unused fixtures and remove all usage of xfail_unsupported

gforsyth:
Sorry in advance for how gross this PR is, but the changes are spread out by nature.  I've tried to group changes to each file into individual commits to hopefully make it easier to browse through.

This removes a few fixtures that were only used previously to exclude the file-based backends  (`rw_con` and `rw_backend`) -- now that those backends are removed, the fixtures aren't needed anymore.

This also removes all usage of `xfail_unsupported` and replaces it with explicit xfails (mostly `notimpl`) for each test that was failing after `xfail_unsupported` was removed.  The idea here is to be explicit about which tests are failing, and also to avoid accidentally swallowing regressions, since tests that start to fail will be dynamically set to an `xfail` and the regression can slip through (I uncovered a few of these in the process of putting this together).

A few sets of tests, notably `test_geospatial.py` and `test_json.py` and `test_nettypes.py` now live inside the `postgres` backend tests, since they (now) are only run on the `postgres` backend (since `omniscidb` is no longer included in this repo).

Additionally, any test that was marked as `only_on_XXX` and which failed all other backends was moved into a backend-specific test.

I've tried to remove all skips that I found, as these tend to swallow both regressions and fixes, I'm sure a few have slipped through.


gforsyth:
Ok, I've fixed up the `sqlite` test issues -- the sqlite tests need to be run on a single process or else sqlite gets very angry, so I'm dynamically adding an `xdist_group` marker to all `sqlite` tests and then using `--dist loadgroup` to distribute those tests to the same process when the tests are run in parallel.

If you run `just test sqlite` and keep an eye on your process monitor, you'll see `pytest` spin up processes equal to the number of cores you have, but only one of them should pin to 100%.

gforsyth:
Since this PR is already enormous, I'll wait on this next idea until it's merged, but it might be worth grouping `pyspark` tests together in the same way as the `sqlite` tests -- then we won't end up with `N_CORES` copies of the JVM running and we can avoid the "test everything but pyspark" shenanigans

cpcloud (approved):
Great work. We can now very easily see what backends do not implement specific operations. It kind of reads almost like "here's the set of possible gaps someone could fill in". Amazing.

cpcloud:
@gforsyth Can you squash to as few commits as makes sense? Then let's get this in right after that.

gforsyth:
squashed down into:
conda-lock
removing old backend stuff
killing xfail_unsupported
hilarious sqlite testing stuff (for easy/easier future reference)

PR #3501, opened 2022-02-25
feat: implement most remaining joins

cpcloud:
This PR implements most remaining joins, to allow nearly all of them to
successfully pass the join backend test. The test itself has also been
simplified.

Things done:

- clickhouse joins
- pyspark coalesce
- semi/anti join for pyspark
- semi/anti join for pandas
- enforcing column names, not just types, in `Schema.apply_to`; this is the handle the
  impala case where column names' case may not match the actual schema's
  columns' case.

Things still not working:

- clickhouse right/outer joins; the results don't seem to match up to any other
  backend because values selected from the right table on right join are zero
  or null when there's no match, similar behavior occurs with outer joins

Things not yet implemented in ibis:

- datafusion; joins aren't implemented
- dask; semi/anti joins aren't implemented

Things that aren't yet implemented upstream and may never be:

- right joins in sqlite
- full outer joins in mysql and sqlite

Things that seem broken:

- The clickhouse and impala compilers generate invalid SQL for the semi/anti join query in `test_filtering_join`. This needs to be fixed in a subsequent PR.

### DIFF
--- a/ci/datamgr.py
+++ b/ci/datamgr.py
@@ -167,6 +167,11 @@ PARQUET_SCHEMAS = {
         ]
     ),
 }
+PARQUET_SCHEMAS.update(
+    (table, schema)
+    for table, schema in TEST_TABLES.items()
+    if table != "functional_alltypes"
+)
 
 
 AVRO_SCHEMAS = {
@@ -182,6 +187,30 @@ AVRO_SCHEMAS = {
 }
 
 ALL_SCHEMAS = collections.ChainMap(PARQUET_SCHEMAS, AVRO_SCHEMAS)
+
+
+@click.group()
+@click.option("-v", "--verbose", count=True)
+def cli(verbose):
+    codes = {0: logging.ERROR, 1: logging.INFO, 2: logging.DEBUG}
+    logger.setLevel(codes.get(verbose, logging.DEBUG))
+
+
+@cli.command()
+@click.pass_context
+def create(ctx):
+    import pyarrow.parquet as pq
+
+    executor = ctx.obj["executor"]
+
+    for name, table in read_tables(
+        (name for name in TEST_TABLES if name != "functional_alltypes"),
+        DATA_DIR,
+    ):
+        logger.info(f"Creating {name}.parquet from CSV")
+        dirname = DATA_DIR / "parquet" / name
+        dirname.mkdir(parents=True, exist_ok=True)
+        executor.submit(pq.write_table, table, dirname / f"{name}.parquet")
 
 
 def impala_create_tables(con, env, *, executor=None):
@@ -283,13 +312,6 @@ def read_tables(
             data_directory / f'{name}.csv',
             convert_options=convert_options,
         )
-
-
-@click.group()
-@click.option("-v", "--verbose", count=True)
-def cli(verbose):
-    codes = {0: logging.ERROR, 1: logging.INFO, 2: logging.DEBUG}
-    logger.setLevel(codes.get(verbose, logging.DEBUG))
 
 
 @cli.group()
@@ -755,6 +777,11 @@ if __name__ == '__main__':
      - etc.
     """
     with concurrent.futures.ThreadPoolExecutor(
-        max_workers=URLLIB_DEFAULT_POOL_SIZE
+        max_workers=int(
+            os.environ.get(
+                "IBIS_DATA_MAX_WORKERS",
+                URLLIB_DEFAULT_POOL_SIZE,
+            )
+        )
     ) as executor:
         cli(auto_envvar_prefix='IBIS_TEST', obj=dict(executor=executor))
--- a/ci/datamgr.py
+++ b/ci/datamgr.py
@@ -198,7 +198,7 @@ def cli(verbose):
 
 @cli.command()
 @click.pass_context
-def create(ctx):
+def generate_parquet(ctx):
     import pyarrow.parquet as pq
 
     executor = ctx.obj["executor"]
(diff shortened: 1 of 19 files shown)


ci/datamgr.py
```diff
@@ -184,6 +189,30 @@ def impala_create_test_database(con, env):
...
 
+@click.group()
+@click.option("-v", "--verbose", count=True)
+def cli(verbose):
+    codes = {0: logging.ERROR, 1: logging.INFO, 2: logging.DEBUG}
+    logger.setLevel(codes.get(verbose, logging.DEBUG))
+
+
+@cli.command()
+@click.pass_context
```
kszucs, line 200: Is the usage `python ci/datamgr.py create` here? If yes, can we use a more descriptive command?

ibis/backends/tests/test_join.py
```diff
@@ -1,81 +1,153 @@
...
             marks=pytest.mark.notimpl(
-                [
-                    "clickhouse",
-                    "datafusion",
-                    "mysql",
-                    "postgres",
-                    "sqlite",
-                    "impala",
-                ]
+                ["clickhouse", "datafusion", "mysql", "sqlite"]
```
kszucs, line 85: Haven't you [added](https://github.com/ibis-project/ibis/pull/3501/files#diff-ecfec13cc30bc29a47126b496dc07567347a41c67922abb7ad0cc7e996eb7f72R62) support for clickhouse?

ibis/backends/tests/test_temporal.py
```diff
@@ -383,20 +383,53 @@ def test_temporal_binop_pandas_timedelta(
...
     [
-        lambda t: t.timestamp_col > pd.Timestamp('20100301'),
-        lambda t: t.timestamp_col >= pd.Timestamp('20100301'),
-        lambda t: t.timestamp_col < pd.Timestamp('20100301'),
-        lambda t: t.timestamp_col <= pd.Timestamp('20100301'),
-        lambda t: t.timestamp_col == pd.Timestamp('20100301'),
-        lambda t: t.timestamp_col != pd.Timestamp('20100301'),
+        param(
+            lambda col: (
+                col > pd.Timestamp('20100301', tz="UTC").to_pydatetime()
```
kszucs, line 388: Maybe factor out the datetime value to a global constant?

ibis/backends/tests/test_temporal.py
```diff
@@ -475,13 +508,13 @@ def test_to_timestamp(backend, con, unit):
...
-        ('2017-01-07', 5, 'Saturday'),
+        param('2017-01-01', 6, 'Sunday', id="sunday"),
+        param('2017-01-02', 0, 'Monday', id="monday"),
+        param('2017-01-03', 1, 'Tuesday', id="tuesday"),
+        param('2017-01-04', 2, 'Wednesday', id="wednesday"),
+        param('2017-01-05', 3, 'Thursday', id="thursday"),
+        param('2017-01-06', 4, 'Friday', id="friday"),
+        param('2017-01-07', 5, 'Saturday', id="saturday"),
     ],
 )
```
kszucs, line 519: You can use the `ids` keyword for parametrize alternatively:

```suggestion
    ids=lambda tpl: tpl[2].lower()
)
```

ibis/backends/tests/test_join.py
```diff
@@ -1,81 +1,153 @@
...
-                    "dask",
-                    "datafusion",
-                    "impala",
-                    "mysql",
-                    "pandas",
-                    "postgres",
-                    "pyspark",
-                    "sqlite",
-                ]
+                # clickhouse and impala are broken, the rest are not
```
kszucs, line 143: Please create follow-up tickets.

ibis/expr/schema.py
```diff
@@ -167,10 +167,11 @@ def apply_to(self, df):
         -----
         Mutates `df`
         """
+        assert len(self) == len(df.columns)
 
-        for column, dtype in self.items():
+        for (column, dtype), orig_col in zip(self.items(), df.columns):
             pandas_dtype = dtype.to_pandas()
-            col = df[column]
+            col = df.pop(orig_col)
```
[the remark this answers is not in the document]
cpcloud: No, but I'm potentially renaming the columns so I'm removing them first.

ci/datamgr.py
cpcloud: Done.

ibis/backends/tests/test_join.py
cpcloud: Done.

ibis/backends/tests/test_temporal.py
cpcloud: This is called for every argument unfortunately so I don't think there's much deduplication to be had here.

ibis/backends/tests/test_temporal.py
cpcloud: Done.

kszucs:
@cpcloud DuckDB build are failing.

ibis/backends/tests/test_temporal.py
```diff
@@ -401,20 +402,24 @@ def test_temporal_binop_pandas_timedelta(
...
-        lambda t: t.timestamp_col != pd.Timestamp('20100301'),
+        operator.gt,
+        operator.ge,
+        operator.lt,
+        operator.le,
+        operator.eq,
+        operator.ne,
     ],
 )
-@pytest.mark.notimpl(["datafusion", "sqlite"])
```
cpcloud, line 412: Why is datafusion passing?

kszucs: Could actually be implemented and no need to mark. I need to try it out locally to tell more.

PR #3510, opened 2022-02-28
refactor: reuse more sqlalchemy base class code

cpcloud:
This PR deletes a good chunk of code from some of the sqlalchemy-based backends,
allowing them to use the base class code where possible.

Notable changes:

- Remove of the `has_attachment` hack in favor of a private `_current_schema` property, specific to a backend.
- Removal of the `create` parameter to the sqlite backend. There is no substitute here, this is a breaking change.
- Allowance of the string `":memory:"` for the `path` argument to the sqlite connection API.
- Implementation of the `force` argument to `create_table` on the base sqlalchemy backend class.

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -92,10 +92,9 @@ class BaseAlchemyBackend(BaseSQLBackend):
         return '.'.join(map(str, self.con.dialect.server_version_info))
 
     def list_tables(self, like=None, database=None):
-        inspector = sa.inspect(self.con)
-        tables = inspector.get_table_names(
+        tables = self._inspector.get_table_names(
             schema=database
-        ) + inspector.get_view_names(schema=database)
+        ) + self._inspector.get_view_names(schema=database)
         return self._filter_with_like(tables, like)
 
     def list_databases(self, like=None):
@@ -333,7 +332,7 @@ class BaseAlchemyBackend(BaseSQLBackend):
         return self.database_name
 
     @util.deprecated(version='2.0', instead='`list_databases`')
-    def list_schemas(self):
+    def list_schemas(self, like: str | None = None) -> list[str]:
         return self.list_databases()
 
     def _log(self, sql):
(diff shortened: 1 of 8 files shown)


ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -191,27 +197,25 @@ def create_table(
         )
 
         with self.begin() as bind:
-            t.create(bind=bind)
+            t.create(bind=bind, checkfirst=force)
```
kszucs, line 200: `not force`?

cpcloud: I'm not following.

`checkfirst=True` means `CREATE TABLE IF NOT EXISTS` which is consistent with "force the creation of this table" (force is perhaps a misnomer here, since there's no destructive behavior)

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -78,18 +77,22 @@ def _build_alchemy_url(
...
+        self.meta = sa.MetaData(bind=self.con)
         self._schemas: dict[str, sch.Schema] = {}
 
     @property
     def version(self):
         return '.'.join(map(str, self.con.dialect.server_version_info))
 
     def list_tables(self, like=None, database=None):
-        inspector = sqlalchemy.inspect(self.con)
+        inspector = sa.inspect(self.con)
```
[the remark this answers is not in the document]
cpcloud: Looks like it. This was just a move of the code, I didn't look into deduplicating thi.s

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -335,26 +333,53 @@ def current_database(self) -> str:
...
+        schema = self._schemas.get(table.name)
+        node = self.table_class(table, self, schema)
         return self.table_expr_class(node)
 
+    def table(
+        self,
+        name: str,
+        **kwargs: Any,
+    ) -> ir.TableExpr:
+        """Create a table expression from a table in the SQLite database.
```
kszucs, line 366: I assume SQLite shouldn't be mentioned here.

cpcloud: Yes, thank you :)

ibis/backends/postgres/__init__.py
```diff
@@ -134,32 +134,37 @@ def begin(self):
...
+        TableExpr
+            A table expression
         """
         if database is not None and database != self.current_database:
-            return self.database(name=database).table(name=name, schema=schema)
-        else:
-            alch_table = self._get_sqla_table(name, schema=schema)
-            node = self.table_class(alch_table, self, self._schemas.get(name))
-            return self.table_expr_class(node)
+            return self.database(name=database).table(
```
kszucs, line 162: Could the `database` argument be moved to the base class?

ibis/backends/base/sql/alchemy/__init__.py
kszucs: Force indicates "no matter what" for me.

cpcloud: How is that different from what I described?

kszucs: `force` indicates destructive behavior for me: `DROP TABLE IF EXISTS...` with `CREATE TABLE...` in a transaction, though I assume we don't want that.

cpcloud: `force` is a misnomer, but I think it's out of scope to change the name of this parameter in this PR. However, `force` is consistent with "don't fail if the thing already exists", which is what the current semantics are.

ibis/backends/base/sql/alchemy/__init__.py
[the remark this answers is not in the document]
cpcloud: Done.

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: Done.

kszucs (approved):
LGTM

PR #3519, opened 2022-03-01
docs: clarify connection parameters

### LINKED ISSUE #3514: Inconsistency in documentation
In [Impala backend](https://ibis-project.org/backends/Impala/), in Connection API>method>do_connect:

- The method listed is called `do_connect`
- However, the examples use `connect`
- I ahve always used `connect`

Question is: Is it then `connect` or `do_connect` ? The incorrect one should be ajusted in the documentation. If both of them are available and doing the same thing, then why is this redundant stuff in here ? Moreover, it should be stated somewhere.

cpcloud:
This PR clarifies connection parameters and fills out related docs. Closes #3514.


ibis/backends/clickhouse/__init__.py
```diff
@@ -62,7 +62,7 @@ def do_connect(
         password
             Password to authenticate with
         client_name
-            This will appear in clickhouse server logs
+            Name of client that wil appear in clickhouse server logs
```
cpcloud, line 65:
```suggestion
            Name of client that will appear in clickhouse server logs
```

saulpw (approved):
undo `@classmethod` on `connect`, otherwise looks good.

PR #3521, opened 2022-03-01
chore: small duckdb follow ups

cpcloud:
This PR adds backend docs for duckdb and removes the table method implementation in favor of the base class implementation.


gforsyth (approved):
looks good!

PR #3535, opened 2022-03-03
docs: add goatcounter analytics

cpcloud:
This adds analytics for ibis-project.org using [goatcounter](https://www.goatcounter.com/).

goatcounter is privacy preserving, and strives to be GDPR compliant.


gforsyth (approved):
Goooooooooooat!

Issue #1711, opened 2019-03-07
refactor: remove .distinct method from ColumnExpr


toryhaavik:
Given an expression like:
`expr[expr.column.distinct().name('thing'), ibis.literal(1).name('dummy')]`
you'd expect to have a table with N rows per distinct values in `expr.column`, each with a value of `1` in the `dummy` column.

This is similar to `SELECT DISTINCT column, 1 AS 'dummy' FROM table` in SQL.

With the pandas backend, you get the same number of rows as `expr` would generate, with `nan`s filled in the column after the distinct values.

cpcloud:
So, I implemented a "fix" for this which involved adding a an analysis pass and spending some time thinking about the `DistinctColumn` operation.

I'm now really against keeping this operation around as leads to expressions that are very confusing to read and whose semantics are difficult to reason about, negating any convenience provided by having this method. Moreover, no backends appear to implements this operation correctly

Here's an expression that would make me raise an eyebrow if I came across it:

```python
t[2 * t.a.distinct() + t.b.distinct().log().distinct() + t.c.log().distinct()]
```

I'd like to propose that we deprecate the `distinct()` method on columns. Any correct behavior that was previously possible can be done with a combination of `TableExpr.distinct()` and projections.

@toryhaavik Any thoughts on this? Making this change would move this issue into major release territory, if that's helpful to know.

cpcloud:
Closed by #3545.

PR #3545, opened 2022-03-05
refactor(api): remove distinct ColumnExpr API

cpcloud:
Remove `ColumnExpr.distinct()` because it's a confusing API.

This is a breaking change.

Here are a few examples of why this method is easily misused, confusing and error prone:

```
>>> import ibis
>>> t = ibis.table(schema=[("a", "int64"), ("b", "string"), ("c", "string")], name="t")
>>> t.a.distinct().log().distinct()  # what does this even compile to?
>>> t.b.length().distinct() + t.c.distinct().length()  # again, this is very ambiguous if defined at all
```

The `TableExpr.distinct()` method still exists, so it's still possible to
compute distict rows, but it can no longer be done with `t.foo.distinct()`.

Note that this doesn't prevent the following operation from being constructed, but
it will fail to compile due to the expression having multiple leaf tables outside of
a join or correlated subquery:

```
>>> t[["a"]].distinct().a + t["a"]
UnboundTable[r0, name=t]
  a int64
  b string
  c string

Selection[r1]
  table:
    ref: r0
  selections:
    a: int64 = r0.a

Distinct[r2]
  table:
    ref: r1

int64 = Add
  left:
    a: int64 = r2.a
  right:
    a: int64 = r0.a
```

### DIFF
--- a/ibis/backends/base/sql/compiler/select_builder.py
+++ b/ibis/backends/base/sql/compiler/select_builder.py
@@ -317,20 +317,8 @@ class SelectBuilder:
                 # Something more complicated.
                 base_table = L.find_source_table(expr)
 
-                if isinstance(op, ops.DistinctColumn):
-                    expr = op.arg
-                    try:
-                        name = op.arg.get_name()
-                    except Exception:
-                        name = 'tmp'
-
-                    table_expr = base_table.projection(
-                        [expr.name(name)]
-                    ).distinct()
-                    result_handler = _get_column(name)
-                else:
-                    table_expr = base_table.projection([expr.name('tmp')])
-                    result_handler = _get_column('tmp')
+                table_expr = base_table.projection([expr.name('tmp')])
+                result_handler = _get_column('tmp')
 
             return table_expr, result_handler
         else:
--- a/ibis/backends/clickhouse/tests/test_functions.py
+++ b/ibis/backends/clickhouse/tests/test_functions.py
@@ -1,5 +1,4 @@
 import math
-import operator
 from datetime import date, datetime
 from operator import methodcaller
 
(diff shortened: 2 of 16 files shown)


jreback:
is >  t.a.distinct() still supported? otherwise this is very odd

cpcloud:
> is >  t.a.distinct() still supported? otherwise this is very odd (agree its a bit odd to chain them but couldn't we just detect this an raise at compile time)?

Can you rephrase? Hard to tell what you're asking.

jreback:
is the `.distict()` api still available on a column? this seems like a big change if not

cpcloud:
> is the `.distict()` api still available on a column? this seems like a big change if not

Yes, it still exists.

It's a breaking change and is marked as such.

Detection isn't worth the effort IMO. It requires traversing the tree and encoding a set of ad-hoc rules that prevent certain kinds of operations. We'd be play whack-a-mole ad infinitum prevent all the weird things that can be done.

Moreover there's basically only a single thing that's valid and that's projecting a single `DISTINCT` column. Any operations on `t.foo.distinct()` are ambiguous. For example, does `t.foo.distinct() + 1` mean `(t.foo + 1).distinct()` or something else?

It's also not only about chaining. This is a general problem with having a `distinct` method. How does the operation compose? If the operation is too difficult to make compose in an unambiguous way then IMO it's a candidate for removal.

For example, what does

```
col_expr1.distinct() + col_expr2.distinct()
```

mean? `col_expr1` and `col_expr2` can be any arbitrary column expression.

jreback:
> I'm arguing that there's only a single valid, unambiguous operation, and that it can be done without a distinct method on ColumnExpr.

this is saying 2 things, you are allowing `t.distinct()` e.g. a `TableExpr` operation, ok sure but *not* allowing `t.foo.distinct()` e.g. a `ColumnExpr` operation?

cpcloud:
> > I'm arguing that there's only a single valid, unambiguous operation, and that it can be done without a distinct method on ColumnExpr.
> 
> this is saying 2 things, you are allowing `t.distinct()` e.g. a `TableExpr` operation, ok sure but _not_ allowing `t.foo.distinct()` e.g. a `ColumnExpr` operation?

That is correct. Composing `table_expr.distinct()` is much less error prone.

cpcloud:
@kszucs Thoughts?

kszucs:
@cpcloud needs a rebase

kszucs:
@jreback does it look good to you?

jreback:
> Moreover there's basically only a single thing that's valid and that's projecting a single DISTINCT column.

This is a very valid usecase. I agree this *could* be implented by using `table_expr.distinct()` and then projecting. if that is solution then i guess its ok.

cpcloud:
> > Moreover there's basically only a single thing that's valid and that's projecting a single DISTINCT column.
> 
> This is a very valid usecase. I agree this _could_ be implented by using `table_expr.distinct()` and then projecting. if that is solution then i guess its ok.

Yeah, it is a valid usecase. The question is not about whether computing the distinct values in a column is a valid use case, it's about whether to allow a virtually unlimited number of incorrect expression constructions to make a single use case more convenient.

saulpw:
I think this change is a good one.  xref #1711.

PR #3546, opened 2022-03-05
feat(sqlite): implement date_truncate

cpcloud:
This PR implements truncate for date types in the SQLite backend, and generalizes the implementation of date for the SQLAlchemy based backends.

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -496,6 +496,7 @@ sqlalchemy_operation_registry: Dict[Any, Any] = {
     ops.FloorDivide: _floor_divide,
     # other
     ops.SortKey: _sort_key,
+    ops.Date: unary(lambda arg: sa.cast(arg, sa.DATE)),
 }
 
 
--- a/ibis/backends/postgres/registry.py
+++ b/ibis/backends/postgres/registry.py
@@ -119,15 +119,10 @@ def _timestamp_truncate(t, expr):
 
 
 def _interval_from_integer(t, expr):
-    arg, unit = expr.op().args
-    sa_arg = t.translate(arg)
+    op = expr.op()
+    sa_arg = t.translate(op.arg)
     interval = sa.text(f"INTERVAL '1 {expr.type().resolution}'")
     return sa_arg * interval
-
-
-def _timestamp_add(t, expr):
-    sa_args = list(map(t.translate, expr.op().args))
-    return sa_args[0] + sa_args[1]
 
 
 def _is_nan(t, expr):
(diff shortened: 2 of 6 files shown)


ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -496,6 +496,7 @@ def _sort_key(t, expr):
     ops.FloorDivide: _floor_divide,
     # other
     ops.SortKey: _sort_key,
+    ops.Date: unary(lambda arg: sa.cast(arg, sa.DATE)),
```
kszucs, line 499: Shouldn't we cast only if `arg` is not a `Date` already?
I'm not sure how `sa` handles it internally.

kszucs: Or ibis should prevent that scenario entirely?

ibis/backends/postgres/registry.py
```diff
@@ -119,17 +119,12 @@ def _timestamp_truncate(t, expr):
 
 
 def _interval_from_integer(t, expr):
-    arg, unit = expr.op().args
+    arg, _ = expr.op().args
```
kszucs, line 122: Since we support node inheritance the  order of `args` isn't that explicit anymore, so I'd prefer using attributed explicitly. Besides that accessing properties should be quicker too.

ibis/backends/sqlite/registry.py
```diff
@@ -19,32 +21,57 @@
 operation_registry.update(sqlalchemy_window_functions_registry)
 
 
+sqlite_cast = Dispatcher("sqlite_cast")
+
+
+@sqlite_cast.register(AlchemyExprTranslator, ir.IntegerValue, dt.Timestamp)
+def _(t, arg, _):
```
kszucs, line 28: Can we have a short function name here, like `int_to_timestamp`? Can make it easier to understand the actual overload.

ibis/backends/sqlite/registry.py
```diff
@@ -19,32 +21,57 @@
...
-            return sa.func.date(sa.func.datetime(sa_arg, 'unixepoch'))
-        elif isinstance(arg, ir.StringValue):
-            return sa.func.date(sa_arg)
-        raise com.UnsupportedOperationError(type(arg))
-
-    if isinstance(arg, ir.CategoryValue) and target_type == 'int32':
-        return sa_arg
-    else:
-        return sa.cast(sa_arg, sa_type)
+    return sqlite_cast(t, arg, target_type)
```
kszucs, line 74:
```suggestion
    return sqlite_cast(t, op.arg, op.target_type)
```

ibis/backends/postgres/registry.py
cpcloud: Roger, good catch.

ibis/backends/base/sql/alchemy/registry.py
cpcloud: In this scenario, it's likely the database itself will do no additional work to cast if the input argument isn't a `DATE`.

In the long term I don't think ibis should do that. It's not clear to me that that's a safe operation in every case, and that it's worth any potential trouble.

kszucs:
@cpcloud sqlite is failing

kszucs (approved):

PR #3554, opened 2022-03-07
test: split up extract tests and add epoch_seconds casting type

cpcloud:
This PR splits up the different special cases of timestamp extraction, allowing confirmation that duckdb implements most of the existing functionality.


ibis/backends/postgres/registry.py
```diff
@@ -710,7 +710,7 @@ def _day_of_week_name(t, expr):
         ops.StandardDev: variance_reduction('stddev'),
         ops.RandomScalar: _random,
         # now is in the timezone of the server, but we want UTC
-        ops.TimestampNow: lambda *args: sa.func.timezone('UTC', sa.func.now()),
+        ops.TimestampNow: lambda *_: sa.func.timezone('UTC', sa.func.now()),
```
kszucs, line 713: Perhaps it deserves a named function `_now`?

cpcloud:
@kszucs 

> I assume the implementations are following.

Nope, they're already there. Tests were failing due to dtype checking.

ibis/backends/postgres/registry.py
cpcloud: Seems simple enough to leave it as a `lambda` for now.

gforsyth (approved):
looks good to me

PR #3559, opened 2022-03-07
docs(backends): add operation support matrix

cpcloud:
This PR adds an operation support matrix table the backend docs, showing for each backend whether an operation is implemented by the backend:

image


ibis/backends/datafusion/__init__.py
```diff
@@ -195,3 +196,14 @@ def compile(
         **kwargs: Any,
     ):
         return translate(expr)
+
+    @classmethod
+    def has_operation(self, operation: type[ops.ValueOp]) -> bool:
```
cpcloud, line 201: Should be called `cls`

gen_matrix.py
```diff
@@ -0,0 +1,68 @@
...
+    print("---", file=f)
+    print(file=f)
+    print("# Operation Support Matrix", file=f)
+    print(file=f)
+    print(
+        '!!! note "Backends are shown in descending order of the number of supported operations"',  # noqa: E501
+        file=f,
+    )
+    print(file=f)
+    print('<div class="support-matrix" markdown>', file=f)
```
kszucs, line 46: I wonder, could we [feed an existing plugin](https://github.com/timvink/mkdocs-table-reader-plugin) with a csv instead?

cpcloud: Great idea, will look into it!

cpcloud: This worked out pretty nicely. I'm now using the plugin to read the file.

ibis/backends/base/sql/__init__.py
```diff
@@ -252,3 +252,9 @@ def explain(
         cur.release()
 
         return '\n'.join(['Query:', util.indent(query, 2), '', *result])
+
+    @classmethod
+    def has_operation(cls, operation: type[ops.ValueOp]) -> bool:
+        translator = cls.compiler.translator_class
+        op_classes = translator._registry.keys() | translator._rewrites.keys()
```
kszucs, line 259: We should have a special case for `ScalarParameter`

kszucs: Seems like `Pi` is not implemented for any of the backends, shall we remove it?

kszucs: `SummaryFilter` is implemented for the SQL backends but there are no tests, I assume it's unused.

kszucs:
I rendered the table locally and it looks pretty blurry for me:

image

Shouldn't be a blocker, but definitely deserves a follow-up ticket.

ibis/backends/base/sql/__init__.py
cpcloud: Mind creating those issues?

cpcloud:
@kszucs Mind approving?

PR #3565, opened 2022-03-08
Add clip_lower and clip_upper

rtpsw:


rtpsw:
Following offline discussions, I will hold on pushing this upstream for the time being.

PR #3566, opened 2022-03-08
refactor(ir): remove unused lineage `roots` and `find_nodes` functions

kszucs:
BREAKING CHANGE: removed `ibis.expr.lineage.{roots,find_nodes}` functions


cpcloud (approved):

cpcloud:
Nice work! Merging.

cpcloud:
@icexelloss We can revert, or if you'd be okay copying this out then we can avoid reverting.

The main reason was that this code isn't being used anywhere in the codebase.

gerrymanoim:
Let me see if I can rewrite the equivalents via `traverse`.

icexelloss:
@cpcloud Gotcha. Don't think we necessarily need to revert because it's a small thing (worst case we just copy the code)

But IMO `lineage` is a semi-public/util API that people might just need to inspect their ibis trees.

PR #3570, opened 2022-03-09
refactor(strings): move string API to types module

cpcloud:
This PR moves string operations to their respective types/ module.


ibis/expr/types/strings.py
```diff
@@ -1,18 +1,836 @@
...
+        ----------
+        substr
+            Substring for which to check
+
+        Returns
+        -------
+        BooleanValue
+            Boolean indicating the presence of `substr` in the expression
+        """
+        return self.find(substr) >= 0
```
kszucs, line 182: Eventually if there are backends with specific support for this we may want to introduce a `StringContains` operator and implement this for backends not supporting it.

kszucs (approved):
LGTM

ibis/expr/types/strings.py
cpcloud: Agreed. It looks like DuckDB has `instr` and `position(needle in haystack)` for this.

kszucs: Can you create a follow-up ticket?

PR #3571, opened 2022-03-09
ci: add kind explicit to conda-lock invocation

mariusvniekerk:
This also pins to conda-lock pre 1.0 to debug why the extras weren't working correctly


cpcloud (approved):
Thanks!

PR #3573, opened 2022-03-09
refactor(expressions): move generic methods to ValueExpr class

cpcloud:
This PR moves generic operations to their respective types/ module.


kszucs:
@cpcloud could you move the methods to `AnyValue`, `AnyScalar` and `AnyColumn` instead? I think it would be clearer to define them there. Eventually the `ValueExpr` classes will be removed.

kszucs:
@cpcloud needs a rebase

ibis/expr/types/generic.py
```diff
@@ -119,7 +140,479 @@ def _repr_html_(self) -> str | None:
...
+        return ops.GroupConcat(self, sep=sep, where=where).to_expr()
+
+    def __hash__(self) -> int:
+        return super().__hash__()
+
+    def __eq__(self, other: AnyValue) -> ir.BooleanValue:
+        import ibis.expr.operations as ops
+        import ibis.expr.rules as rlz
+
+        return _binop(ops.Equals, self, rlz.any(other))
```
kszucs, line 585: Why is `rlz.any()` required here?

cpcloud: It may not be, I think was copying the existing code.

cpcloud: This is necessary for `_maybe_cast_args`. Changing that is out of scope for this PR.

kszucs (approved):

PR #3581, opened 2022-03-10
refactor: remove geospatial from unsupported backends

cpcloud:
This PR removes the spurious support for geospatial operations in duckdb


gforsyth (approved):
Looks good -- was it intentionally to leave `mariadb` available in the `nix.shell`?  No harm in having it there for local testing.

cpcloud:
> Looks good -- was it intentional to leave `mariadb` available in the `nix.shell`? No harm in having it there for local testing.

Yep. Ideally we have enough of the local tools at the ready to connect to any of the backends.

PR #3582, opened 2022-03-10
test(marks): add broken mark, remove unneeded markers

gforsyth:
This adds a new marker named `broken` for marking broken tests (:wink:).
This should be used sparingly and is designed for scenarios where a user
has added a new test and has exposed previously broken functionality
that is unrelated to their feature -- in that case, we don't want to
impose the necessity of fixing it on them.

I've also cleaned up a few of the older markers that were used in only a
few spots and replaced them with `notyet`, `notimpl`, and `never`.

The markers removed are:
`ddl_only` - unused
`skip_missing_feature` - unused
`xfail_backends` - few replacements made
`only_on_backends` - few replacements made

I've left in `skip_backends` because there are a few tests (which I
would like to fix) that crash the test suite if they aren't skipped on
particular backends (mostly memory explosions with pandas and dask).


cpcloud (approved):
Excellent! Glad to see nothing `broken` was found during this refactor 😅

PR #3589, opened 2022-03-11
feat(api): Add `ibis.asc` expression

### LINKED ISSUE #1454: feat: expose APIs for constructing SortExprs using asc/desc functions and methods
This will enable sort expr specifications without having to pass an `ascending` argument or have a function like `to_sort_expr` which makes assumptions about the meaning of its second argument.

Moreover, it becomes a lot less cumbersome to specify multiple sort exprs in a function like `array_agg` that, in some backends supports an `ORDER BY` clause.

hussainsultan:
Fixes #1454


cpcloud (approved):
Sweet! Thank you.

PR #3594, opened 2022-03-12
feat(repr): rethink expr repr

cpcloud:
# Rethink the Expr repr

This PR is a total rewrite and rethink of the expression `__repr__`.

There's been a number of PRs and issues about the repr performance and usability.

The main points of feedback are:

1. The repr is just too darn slow.
1. The repr is too nested for deep join and set operation trees
1. The repr contains too much noise for value expressions

This PR attempts to address all of those concerns.

The output format is based on the ideas in [this
gist](https://gist.github.com/gforsyth/1d41488a8680d2dd43bff8a07f1f7931).

The implementation is based on `functools.singledispatch`, where each op type
has its own dispatch rule.

## General approach

1. Convert the ibis expression to a DAG of the underlying `ops.TableNode`s
   using an adjacency list representation.
1. Topologically sort the graph
1. Loop through the DAG, using the loop index as the current node's alias
1. Each TableNode subclass prints its fields
1. The root node is printed last and is special cased since it can be a
   `TableExpr`, `ColumnExpr`, or `ScalarExpr`.

## Results

1. The repr is ~3.5~ ~4.5~ ~7.4~ ~8~ ~9~ **10x** faster to repr TPC-H query 02 versus master.
1. The `str-large` benchmark shows just over a ~2~ **3x** speedup.
1. The repr output no longer has enormous nesting of relational operators. All
   operators are flat, and shown in topological order.
1. Despite there being more code, the code is more maintainable as each node
   and value op is handled by a single rule rather than a giant mututally
   recursive loop handling everything ad-hoc.

### `ibis.options.repr.depth`

This option controls the number of relational operators shown in the repr.

```
In [22]: ibis.options.repr.depth = 2

In [23]: t.group_by("cut").price.mean().sort_by("cut")
Out[23]:
r0 := AlchemyTable[diamonds]
  carat   float32
  cut     string
  color   string
  clarity string
  depth   float32
  table   float32
  price   int64
  x       float32
  y       float32
  z       float32

⋮

Selection[r1]
  sort_keys:
     asc|r1.cut
```

Here we set the `depth` to 2, so we only see the first and last node out of the
3 in the expression.

### `ibis.options.repr.show_types`

This option is a `bool`, which when set to `True` will show inferred types in
`ValueExpr`s in a projection/aggregation or when a `ValueExpr` is the root
node.

```
In [19]: con = ibis.duckdb.connect("ci/ibis-testing-data/ibis_testing.ddb")

In [20]: t = con.table("diamonds")

[...]

PR #3600, opened 2022-03-13
perf(expr/operations): speed up args and argnames attribute access

cpcloud:
This PR speeds up attributes access.

Local benchmarking shows about a **6x** speed up when accessing attributes.

An individual access isn't that expensive, but in a tight loop, the overhead of `@property` plus tuple construction adds up to quite a lot.

This PR is a prerequisite for #3594. We should be able to get a sub-millisecond TPC-H query 2 repr in #3594 after this PR.

### DIFF
--- a/ibis/expr/signature.py
+++ b/ibis/expr/signature.py
@@ -55,7 +55,7 @@ class Parameter(inspect.Parameter):
         name,
         kind=inspect.Parameter.POSITIONAL_OR_KEYWORD,
         *,
-        validator=EMPTY
+        validator=EMPTY,
     ):
         super().__init__(
             name,
@@ -159,16 +159,17 @@ class AnnotableMeta(type):
             params.values(), key=lambda p: p.default is EMPTY, reverse=True
         )
 
-        attribs['__slots__'] = tuple(slots)
-        attribs['__signature__'] = inspect.Signature(params)
+        attribs["__slots__"] = tuple(slots)
+        attribs["__signature__"] = inspect.Signature(params)
+        attribs["argnames"] = tuple(attribs["__signature__"].parameters.keys())
 
         return super().__new__(metacls, clsname, bases, attribs)
 
 
 class Annotable(metaclass=AnnotableMeta):
-    """
-    Base class for dataclass-like objects with custom validation rules.
-    """
+    """Base class for objects with custom validation rules."""
+
+    __slots__ = ("_args",)
 
     def __init__(self, *args, **kwargs):
         bound = self.__signature__.bind(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


ibis/expr/signature.py
```diff
@@ -187,25 +189,24 @@ def __eq__(self, other) -> bool:
...
     def __getstate__(self) -> Dict[str, Any]:
-        return {
-            key: getattr(self, key)
-            for key in self.__signature__.parameters.keys()
-        }
+        return {key: getattr(self, key) for key in self.argnames}
+
+    @property
+    def args(self):
+        if not (result := self._args):
```
jreback, line 196: isn't this 3.10 only syntax? or is there compat providing this

cpcloud: 3.8: https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions

cpcloud:
@jreback Thoughts? Would like to merge this today if possible.

jreback (commented):
will look soon

jreback (approved):

kszucs:
@cpcloud I'm strongly against this change. Usage of `op.args` should be discouraged in favour of explicit attribute access which should be the quickest and less error prone. This change also introduces an inconsistent state handling since the operation nodes are actually mutable, we doesn't prohibit that yet. Caching `args` and `argnames` presumes an immutable object.
If we insist on quicker `.args` and `.argnames` access the same should be achievable by using `@cached_property` decorator.

cpcloud:
> @cpcloud I'm strongly against this change. Usage of `op.args` should be discouraged in favour of explicit attribute access which should be the quickest and less error prone. This change also introduces an inconsistent state handling since the operation nodes are actually mutable, we doesn't prohibit that yet. Caching `args` and `argnames` presumes an immutable object. If we insist on quicker `.args` and `.argnames` access the same should be achievable by using `@cached_property` decorator.

> Usage of `op.args` should be discouraged in favour of explicit attribute access which should be the quickest and less error prone.

Indeed, but this convention does not apply to traversal use cases.

> This change also introduces an inconsistent state handling since the operation nodes are actually mutable, we doesn't prohibit that yet.

That's fair, I can do a follow up to prevent mutation on `ops.Node`.

> If we insist on quicker `.args` and `.argnames` access the same should be achievable by using `@cached_property` decorator.

I do insist, as the traversal speedups are a big win (3-6x depending various factors) 😅 

`@cached_property` cannot be used unless we want to jettison `__slots__` on `Node`s. We could add ``__dict__`` to `__slots__`, but then that defeats any space savings from using `__slots__`. Should we just get rid of `__slots__`?

jreback:
@kszucs does make some good points here about using cached_ instead

does using slots really help? (size trade off)

cpcloud:
> @kszucs does make some good points here about using cached_ instead
> 
> does using slots really help? (size trade off)

I can look into it, but it'll be a week or two.

PR #3609, opened 2022-03-14
refactor: make Annotable truly immutable

cpcloud:
This PR makes Annotable subclasses immutable to the extent possible, as a follow up to #3600.

PR #3619, opened 2022-03-17
refactor(ir): make Annotable immutable

kszucs:
Alternative implementation of https://github.com/ibis-project/ibis/pull/3609

Additions:
 - enforced immutable node arguments making the hashing easier
 - execute the rules before calling `__init__()` (using `AnnotableMeta.__call__()`)
 - implicitly called `__post_init__` setting the `args` tuple and `_hash`

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -354,7 +354,12 @@ class BaseAlchemyBackend(BaseSQLBackend):
 
     def _sqla_table_to_expr(self, table: sa.Table) -> ir.TableExpr:
         schema = self._schemas.get(table.name)
-        node = self.table_class(table, self, schema)
+        node = self.table_class(
+            source=self,
+            sqla_table=table,
+            name=table.name,
+            schema=schema,
+        )
         return self.table_expr_class(node)
 
     def table(
--- a/ibis/expr/datatypes/__init__.py
+++ b/ibis/expr/datatypes/__init__.py
@@ -12,6 +12,7 @@ import re
 import typing
 import uuid as _uuid
 from typing import (
+    AbstractSet,
     Iterable,
     Iterator,
     Literal,
@@ -21,6 +22,7 @@ from typing import (
     TypeVar,
 )
 
+import numpy as np
 import pandas as pd
 import parsy as p
 import toolz
@@ -30,13 +32,12 @@ import ibis.common.exceptions as com
 import ibis.expr.types as ir
 from ibis import util
 
-IS_SHAPELY_AVAILABLE = False
 try:
     import shapely.geometry
 
     IS_SHAPELY_AVAILABLE = True
 except ImportError:
-    pass
+    IS_SHAPELY_AVAILABLE = False
 
 
 class DataType:
@@ -148,10 +149,6 @@ class DataType:
     def column_type(self):
         """Return a column expression with this data type."""
         return functools.partial(self.column, dtype=self)
-
-    def _literal_value_hash_key(self, value) -> tuple[DataType, typing.Any]:
-        """Return a hash for `value`."""
-        return self, value
 
 
 class Any(DataType):
@@ -648,21 +645,6 @@ class Struct(DataType):
             self.name, list(self.pairs.items()), self.nullable
         )
 
-    def _literal_value_hash_key(self, value):
-        return self, _tuplize(value.items())
-
-
-def _tuplize(values):
-    """Recursively convert `values` to a tuple of tuples."""
-
-    def tuplize_iter(values):
-        yield from (
-            tuple(tuplize_iter(value)) if util.is_iterable(value) else value
-            for value in values
-        )
-
-    return tuple(tuplize_iter(values))
-
 
 class Array(Variadic):
     """Array values."""
@@ -680,9 +662,6 @@ class Array(Variadic):
         super().__init__(nullable=nullable)
         self.value_type = dtype(value_type)
         self._pretty += f"<{self.value_type}>"
-
-    def _literal_value_hash_key(self, value):
-        return self, _tuplize(value)
 
 
 class Set(Variadic):
@@ -743,9 +722,6 @@ class Map(Variadic):
         self.key_type = dtype(key_type)
         self.value_type = dtype(value_type)
         self._pretty += f"<{self.key_type}, {self.value_type}>"
-
-    def _literal_value_hash_key(self, value):
-        return self, _tuplize(value.items())
 
 
 class JSON(String):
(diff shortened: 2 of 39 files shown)


kszucs:
@cpcloud still need to fix the remaining issues, just wanted to put up the PR for comparison.

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 import contextlib
 import getpass
+from multiprocessing.sharedctypes import Value
```
cpcloud, line 5: Where is this used?

ibis/expr/operations/relations.py
```diff
@@ -322,7 +309,7 @@ def root_tables(self):
 class SelfReference(TableNode, sch.HasSchema):
     table = rlz.table
 
-    @cached_property
+    @property
```
cpcloud, line 312: Why remove the caching?

ibis/expr/operations/relations.py
```diff
@@ -640,26 +634,31 @@ class Aggregation(TableNode, sch.HasSchema):
...
 
         # Validate predicates
-        validator = FilterValidator([self.table])
-        validator.validate_all(self.predicates)
-
+        validator = FilterValidator([table])
+        validator.validate_all(predicates)
+
+        if not by:
+            sort_keys = tuple()
```
cpcloud, line 652: I'm not following the logic here. `sort_keys` is an empty tuple if `by` is `False`-y?

ibis/expr/operations/relations.py
```diff
@@ -723,11 +722,12 @@ class Distinct(TableNode, sch.HasSchema):
...
 
-    def _validate(self):
+    def __init__(self, table):
         # check whether schema has overlapping columns or not
-        assert self.schema
+        assert table.schema()
+        super().__init__(table=table)
 
-    @cached_property
+    @property
```
cpcloud, line 730: I see, this is because of `__slots__`

ibis/expr/signature.py
```diff
@@ -165,48 +166,63 @@ def __new__(metacls, clsname, bases, dct):
...
 
     def __eq__(self, other) -> bool:
         if self is other:
             return True
+        # TODO(kszucs): measure using the precomputed hash
+        # (
+        #     type(self) == type(other) and
+        #     self._hash == other._hash and
+        #     self.args == other.args
+        # )
```
cpcloud, line 207: I would avoid doing anything with `__eq__` or `equals` just yet, see #3621.

ibis/expr/signature.py
```diff
@@ -165,48 +166,63 @@ def __new__(metacls, clsname, bases, dct):
...
-        ----------
-        state
-            A dictionary storing the objects attributes.
-        """
-        self._args = None
-        for key, value in state.items():
-            setattr(self, key, value)
+    def __setattr__(self, name: str, _: Any) -> None:
+        raise TypeError(
+            f"{self.__class__.__name__} attribute {name!r} is read-only"
```
cpcloud, line 215: Ideally the message indicates that the entire object is immutable, not just a specific property.

ibis/expr/types/core.py
```diff
@@ -51,7 +51,7 @@ def __repr__(self) -> str:
             return repr(result)
 
     def __hash__(self) -> int:
-        return hash(self._key)
+        return hash(self._arg)
```
cpcloud, line 54: The means that sticking and `Expr` and its `Node` into a hash table will trigger equality for or the other and fail for `Expr` and likely for `Node` as well.

ibis/backends/base/sql/alchemy/__init__.py
kszucs: Nowhere, it must have been acted by my IDE.

ibis/expr/datatypes/__init__.py
```diff
@@ -1723,3 +1700,47 @@ def _uuid_to_str(typ: String, value: _uuid.UUID) -> str:
 @_normalize.register(Decimal, int)
 def _int_to_decimal(typ: Decimal, value: int) -> decimal.Decimal:
     return decimal.Decimal(value).scaleb(-typ.scale)
+
+
+@_normalize.register(Array, (tuple, list))
```
kszucs, line 1705: Eventually we could make `normalize` public.

ibis/expr/operations/core.py
```diff
@@ -153,9 +152,12 @@ def is_ancestor(self, other):
...
 
     def to_expr(self):
-        if not hasattr(self, '_expr_cached'):
-            self._expr_cached = self._make_expr()
-        return self._expr_cached
+        try:
+            result = self._expr_cached
+        except AttributeError:
+            result = self._make_expr()
+            object.__setattr__(self, "_expr_cached", result)
```
kszucs, line 159: Shall we remove `_expr_cached` in this PR?

ibis/expr/operations/relations.py
kszucs: Cached property can add more overhead if `self.table.schema()` is cheap. I think we should use cached properties only where schema is actually calculated.

kszucs: Or even better, materialize the expensive properties in `__init__()` since the nodes are immutable now.

ibis/expr/operations/relations.py
kszucs: This was the previous implementation:

```python
if not self.by:
       self.sort_keys.clear()
```

I'm not exactly sure why we need it either.

ibis/expr/signature.py
kszucs: Ok, removing the comment.

ibis/expr/types/core.py
kszucs: Originally I hashed the `self.__class__` as well to avoid that. Shall I put that back?

ibis/util.py
```diff
@@ -30,6 +32,35 @@
 V = TypeVar("V")
 
 
+class frozendict(Mapping, Hashable):
+
+    __slots__ = ("_dict", "_hash")
+
+    def __init__(self, *args, **kwargs):
+        # create a read-only view of the underlying mutable mapping
+        # self._dict = types.MappingProxyType(dict(*args, **kwargs))
```
kszucs, line 41: Remove the `types.MappingProxyType` since it prevent the object from being pickled.

ibis/expr/operations/relations.py
cpcloud: Ok, we can leave it for now

ibis/expr/types/core.py
cpcloud: Yeah, let's keep `__class__`

ibis/expr/operations/core.py
cpcloud: Let's remove it in a follow up.

kszucs (commented):
Thanks @cpcloud for finishing it! 

+1 from me.

cpcloud:
@kszucs Is there any way we can avoid changing the `execute_node` interface to pass in tuples as arguments? If not, no worries.

kszucs:
Can't think of any clean solution.

jreback (approved):
seems fine

PR #3613, opened 2022-03-15
refactor(datatypes): make DataType and friends immutable and remove __slots__usage

cpcloud:
This PR moves `DataType` and its subclasses away from `__slots__` and replaces the current
property caching with `cached_property`s.

PR #3621, opened 2022-03-17
perf(ir): speed up expr and op equality comparisons by caching

cpcloud:
This PR refactors the caching mechanism for expr and op equality comparison. Previously we were caching in a very ad hoc manner. After this PR, subclasses of util.CachedEq need only implement the `_raw_equals` interface to gain the benefits of caching. Additionally, some comparisons have been specifically implemented to hit the short circuit path as soon as possible.

### DIFF
--- a/ibis/expr/datatypes/__init__.py
+++ b/ibis/expr/datatypes/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import abc
 import ast
 import builtins
 import collections
@@ -13,10 +14,12 @@ import typing
 import uuid as _uuid
 from typing import (
     AbstractSet,
+    Hashable,
     Iterable,
     Iterator,
     Literal,
     Mapping,
+    MutableMapping,
     NamedTuple,
     Sequence,
     TypeVar,
@@ -40,7 +43,7 @@ except ImportError:
     IS_SHAPELY_AVAILABLE = False
 
 
-class DataType:
+class DataType(util.CachedEqMixin):
     """Base class for all data types."""
 
     nullable: bool
@@ -83,11 +86,17 @@ class DataType:
         }
         return self.__class__(nullable=nullable, **slots)
 
-    def __eq__(self, other) -> bool:
-        return self.equals(other)
-
-    def __ne__(self, other) -> bool:
-        return not (self == other)
+    def equals(
+        self,
+        other: typing.Any,
+        cache: MutableMapping[Hashable, bool] | None = None,
+    ) -> bool:
+        if not isinstance(other, DataType):
+            raise TypeError(
+                'Comparing datatypes to other types is not allowed. Convert '
+                f'{other!r} to the equivalent DataType instance.'
+            )
+        return super().equals(other, cache=cache)
 
     def _make_hash(self) -> int:
         custom_parts = tuple(getattr(self, slot) for slot in self._slot_list)
@@ -113,26 +122,29 @@ class DataType:
         """Return the name of the data type."""
         return self.__class__.__name__
 
-    def equals(
+    def __component_eq__(
         self,
         other: DataType,
-        cache: Mapping[typing.Any, bool] | None = None,
+        cache: MutableMapping[Hashable, bool],
     ) -> bool:
-        """Return whether this data type equals `other`."""
-        if isinstance(other, str):
-            raise TypeError(
-                'Comparing datatypes to strings is not allowed. Convert '
-                f'{other!r} to the equivalent DataType instance.'
-            )
-        return (
-            isinstance(other, self.__class__)
-            and self.nullable == other.nullable
-            and self._slot_list == other._slot_list
-            and all(
-                getattr(self, slot) == getattr(other, slot)
-                for slot in self._slot_list
-            )
+        return self.nullable == other.nullable and self.__fields_eq__(
+            other, cache=cache
         )
+
+    @abc.abstractmethod
+    def __fields_eq__(
+        self,
+        other: DataType,
+        cache: MutableMapping[Hashable, bool],
+    ) -> bool:
+        """Return whether the fields of two datatypes are equal."""
+
+    def _type_check(self, other: typing.Any) -> None:
+        if not isinstance(other, DataType):
+            raise TypeError(
+                "invalid equality comparison between "
+                f"DataType and {type(other)}"
+            )
 
     def castable(self, target, **kwargs):
         """Return whether this data type is castable to `target`."""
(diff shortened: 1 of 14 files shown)


ibis/expr/datatypes/__init__.py
```diff
@@ -82,11 +84,17 @@ def _factory(self, nullable: bool = True) -> DataType:
...
-        return self.equals(other)
-
-    def __ne__(self, other) -> bool:
-        return not (self == other)
+    def equals(
+        self,
+        other: typing.Any,
+        cache: MutableMapping[Hashable, bool] | None = None,
+    ) -> bool:
+        if not isinstance(other, DataType):
```
kszucs, line 92: Can we move this check to `CachedEq`?

ibis/util.py
```diff
@@ -407,3 +407,56 @@ def wrapper(*args, **kwargs):
...
     return decorator
+
+
+class Eq:
+    """A mixin for abstracting equality checks."""
+
+    __slots__ = ()
+
+    def __eq__(self, other: Any) -> bool:
+        return self.equals(other, cache={})
```
kszucs, line 418: How about 

```py
try:
    return self.equals(other, cache={})
except:
    raise NotImplemented
```

?

ibis/util.py
```diff
@@ -407,3 +407,56 @@ def wrapper(*args, **kwargs):
...
+
+    def __ne__(self, other: Any) -> bool:
+        return not (self == other)
+
+    def equals(
+        self,
+        other: Any,
+        cache: MutableMapping[Hashable, bool] | None = None,
+    ) -> bool:
+        return self is other or (
```
kszucs, line 428: Along with my previous suggestion:

```python
if self is other:
    return True
if type(self) != type(other):
    raise TypeError(...)
return self._raw_equals(other, cache=cache)
```

ibis/util.py
```diff
@@ -407,3 +407,56 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+class Eq:
```
kszucs, line 412: Can we call it either `EqMixin` or some `*able`, like `Comparable`?

ibis/util.py
```diff
@@ -407,3 +407,56 @@ def wrapper(*args, **kwargs):
...
+        cache: MutableMapping[Hashable, bool] | None = None,
+    ) -> bool:
+        return self is other or (
+            type(self) == type(other) and self._raw_equals(other, cache=cache)
+        )
+
+    def _raw_equals(
+        self,
+        other: Any,
+        cache: MutableMapping[Hashable, bool],
```
kszucs, line 435: Turn it into an `@abstractmethod`?

ibis/util.py
```diff
@@ -407,3 +407,56 @@ def wrapper(*args, **kwargs):
...
+    ) -> bool:
+        """Compare the individual fields of a subclass."""
+        return True
+
+
+class CachedEq(Eq):
```
kszucs, line 441: How about merging these classes? `cache` arg in `Eq._raw_equals`'s signature seems odd.

ibis/util.py
```diff
@@ -407,3 +407,56 @@ def wrapper(*args, **kwargs):
...
+    def equals(
+        self,
+        other: Any,
+        cache: MutableMapping[Hashable, bool] | None = None,
+    ) -> bool:
+        return self is other or (
+            type(self) == type(other) and self._raw_equals(other, cache=cache)
+        )
+
+    def _raw_equals(
```
kszucs, line 432: Perhaps `__equals__` or that would be too confusing?

cpcloud: How about `__component_eq__`?

ibis/util.py
cpcloud: They are separate to allow `Annotable` subclasses that aren't hashable. I'm not sure whether this occurs in practice but there are a few tests that use them.

kszucs: That shouldn't be possible asof the immutable enforcement PRs land.

cpcloud: Makes sense, let's land that first then.

cpcloud:
@kszucs Thoughts on this one?

ibis/util.py
kszucs: Actually __field_eq__ used at the datatypes looks like a better name. It should make the port of datatypes to be annotable easier too.

ibis/util.py
kszucs: Possibly requires a rebase but on top of that PR

cpcloud:
@jreback Actually if you could review this one, we should land it before #3613.

ibis/expr/datatypes/__init__.py
```diff
@@ -82,11 +85,17 @@ def _factory(self, nullable: bool = True) -> DataType:
...
+            raise TypeError(
+                'Comparing datatypes to other types is not allowed. Convert '
```
jreback, line 95: shouldn't this return False? and not raise

ibis/expr/datatypes/__init__.py
```diff
@@ -702,6 +790,16 @@ def __init__(
         self.value_type = dtype(value_type)
         self._pretty += f"<{self.value_type}>"
 
+    def _literal_value_hash_key(self, value):
+        return self, _tuplize(value)
+
+    def __fields_eq__(
+        self,
+        other: Set,
```
jreback, line 798: can u put some of these in a super class?

ibis/expr/operations/relations.py
```diff
@@ -743,6 +935,19 @@ class ExistsSubquery(Node):
...
+    def __component_eq__(
+        self,
+        other: ExistsSubquery,
+        cache: MutableMapping[Hashable, bool],
+    ) -> bool:
+        return self.foreign_table.equals(
+            other.foreign_table,
+            cache=cache,
+        ) and all(
+            left_pred.equals(right_pred, cache=cache)
```
jreback, line 947: maybe with a helper function to avoid writing all of the left right equals repeatedly

jreback (changes requested):

ibis/expr/datatypes/__init__.py
cpcloud: No, the point is to ensure that any subclass of `DataType` defines it and avoid any implicit behavior.

ibis/expr/datatypes/__init__.py
cpcloud: No, comparing a datatype with something other than a datatype is an error.

ibis/expr/datatypes/__init__.py
jreback: got it makes sense

ibis/expr/operations/relations.py
cpcloud: Done. Added `util.seq_eq` for comparing sequences of `CachedEqMixin` subclasses.

cpcloud:
@jreback This should be good to go on green.

jreback (approved):

PR #3626, opened 2022-03-21
refactor(ir): global equality cache / annotable based datatype and schema

kszucs:
- implement DataType(s) using Annotable
- implement Schema using Annotable
- use a global equality cache with:
a. weakly referenced objects
b. unordered cache key


kszucs:
Need to check the benchmarks.

PR #3630, opened 2022-03-22
fix(duckdb): add casts to literals to fix binding errors

### LINKED ISSUE #3629: bug(duckdb): duckdb backend should add in CAST for some bind parameters
DuckDB casts bind parameters `?` to strings which leads to binder errors with some queries

If we have a small tpch dataset:

```python
import duckdb
con = duckdb.connect("tpch.ddb")
con.execute("CALL dbgen(sf=0.1)")

import ibis
con = ibis.duckdb.connect("tpch.ddb")

t = con.table('orders')
expr = t.aggregate(high_line_count=(t.o_orderpriority.case().when('1-URGENT', 1).else_(0).end().sum()

expr.execute()
```

raises

```
RuntimeError: Binder Error: No function matches the given name and argument types 'sum(VARCHAR)'. You might need to add explicit type casts.
	Candidate functions:
	sum(DECIMAL) -> DECIMAL
	sum(SMALLINT) -> HUGEINT
	sum(INTEGER) -> HUGEINT
	sum(BIGINT) -> HUGEINT
	sum(HUGEINT) -> HUGEINT
	sum(DOUBLE) -> DOUBLE

LINE 1: SELECT sum(CASE WHEN (t0.o_orderpriority = ?) ...
```

because our generated SQL doesn't have explicit casts:

```
print(expr.compile())
SELECT sum(CASE WHEN (t0.o_orderpriority = ?) THEN ? ELSE ? END) AS high_line_count 
FROM orders AS t0
```

we want to generate

```
SELECT sum(CASE WHEN (t0.o_orderpriority = ?) THEN cast(? as INTEGER) ELSE cast(? as INTEGER) END) AS high_line_count FROM orders as t0
```

gforsyth:
Fixes #3629

### DIFF
--- a/ibis/backends/duckdb/registry.py
+++ b/ibis/backends/duckdb/registry.py
@@ -99,7 +99,7 @@ def _literal(_, expr):
             f"`{type(value).__name__}` isn't yet supported with the duckdb "
             "backend"
         )
-    return sa.literal(value)
+    return sa.cast(sa.literal(value), sqla_type)
 
 
 def _array_column(t, expr):
--- a/ibis/backends/tests/test_aggregation.py
+++ b/ibis/backends/tests/test_aggregation.py
@@ -461,3 +461,14 @@ def test_aggregate_mixed(backend, alltypes, df):
     )
 
     backend.assert_frame_equal(result, expected, check_like=True)
+
+
+@pytest.mark.notimpl(["datafusion", "pyspark"])
+def test_binds_are_cast(alltypes):
+    expr = alltypes.aggregate(
+        high_line_count=(
+            alltypes.string_col.case().when('1-URGENT', 1).else_(0).end().sum()
+        )
+    )
+
+    expr.execute()
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -111,7 +111,7 @@ def test_coalesce(backend, con, expr, expected):
         # False
         assert result == decimal.Decimal(str(expected))
     else:
-        assert result == expected
+        assert result == pytest.approx(expected)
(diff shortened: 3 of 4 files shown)


gforsyth:
I added a TPCH fixture because that's often where I'm hitting these issues, but if it's too heavy I can use some of the existing fixtures for this test

cpcloud:
Looks like there's a floating point test that fails, probably `pytest.approx` would help here.

ibis/backends/duckdb/tests/conftest.py
```diff
@@ -14,3 +16,15 @@ def __init__(self, data_directory: Path) -> None:
     def connect(data_directory: Path):
         path = data_directory / "ibis_testing.ddb"
         return ibis.duckdb.connect(str(path))  # type: ignore
+
+
+@pytest.fixture(scope="session")
+def tpch(data_directory):
+    import duckdb
```
cpcloud, line 23: In theory `duckdb` can be imported at the top level without causing annoying failures when it's not installed.

ibis/backends/duckdb/tests/test_bind.py
```diff
@@ -0,0 +1,9 @@
+def test_binds_are_cast(tpch):
```
cpcloud, line 1: Probably should just stick this in `test_aggregation.py` in the backends test suite.

gforsyth:
Ok, stripped out the tpch stuff and stuck it in `test_aggregation.py` using `alltypes`.  Everything but `datafusion` and `pyspark` works out of the box!

cpcloud (approved):
Thanks!

PR #3636, opened 2022-03-23
fix: added wheel to pyproject.toml for venv users

p-a-a-a-trick:
fix: added wheel to pyproject.toml for venv users


gforsyth:
Hey @p-a-a-a-trick -- we have a lint bot for making sure commit message adhere to the conventional changelog format standard.

If you re-title your commit `fix: add wheel to pyproject.toml for venv users` it should get CI in the green.

PR #3637, opened 2022-03-23
fix: regression in sort_by with resolved_keys

gforsyth:
Fix for minor regression introduced in #3619


gforsyth:
This test should maybe go in `test_client.py` instead?  Not sure.

cpcloud:
> This test should maybe go in `test_client.py` instead? Not sure.

Seems fine as you've written it.

cpcloud (approved):
Thanks!

PR #3642, opened 2022-03-24
feat(sql): implement methods for mixing sql with expressions

cpcloud:
# Description

This PR implements methods for mixing SQL strings inline with ibis expressions
as well as the reverse, using ibis table expressions in SQL strings:

Here's an example of what this looks like:

```python
alltypes = con.table("functional_alltypes")
t = (
    alltypes.sql(
        """
        SELECT
            string_col as s,
            double_col + 1.0 AS new_col
        FROM functional_alltypes
        """
    )
    .group_by("s")  # group by a column from SQL
    .aggregate(fancy_af=lambda t: t.new_col.mean())
    .alias("awesome_t")  # create a name for the aggregate
    .sql("SELECT fancy_af AS yas FROM awesome_t ORDER BY fancy_af")
)
```

# Implementation

## Limitations

This functionality currently only works for the following backends:

1. PostgreSQL
1. DuckDB
1. PySpark

In principle any backend that:

1. Is SQL-based
2. Supports views
3. Implements `_get_schema_using_query`

Should be able to implement this functionality.

## Design

There are two new `PhysicalTable` nodes:

1. `View` for named tables, created by the `alias(name)` method.
1. `SQLStringView` for names tables whose computation is described by a SQL string.

## Implementation

During compilation a temporary view is created using `CREATE OR REPLACE VIEW ...`.

Views with the same name as an existing view are replaced.

### `View`

For `View` the view's name is the user-provided `name` argument to the `alias`
method.

Its definition is the compiled ibis expression.

### `SQLStringView`

For `SQLStringView` a temporary name (current `_ibis_view_$N`, where `N` is an incrementing integer) is generated.
Its definition is the SQL string provided by the user.

### Cleanup

~Any views created are dropped using an `atexit` handler.~ All cleanup is handled by the database. Any future implementations for databases that do not support temporary view creation will need to handle cleanup.

### DIFF
--- a/ibis/backends/base/sql/alchemy/query_builder.py
+++ b/ibis/backends/base/sql/alchemy/query_builder.py
@@ -1,9 +1,12 @@
+from __future__ import annotations
+
 import functools
 
 import sqlalchemy as sa
 import sqlalchemy.sql as sql
 
 import ibis.expr.operations as ops
+import ibis.expr.schema as sch
 import ibis.expr.types as ir
 from ibis.backends.base.sql.compiler import (
     Compiler,
@@ -16,6 +19,10 @@ from ibis.backends.base.sql.compiler import (
 from .database import AlchemyTable
 from .datatypes import to_sqla_type
 from .translator import AlchemyContext, AlchemyExprTranslator
+
+
+def _schema_to_sqlalchemy_columns(schema: sch.Schema) -> list[sa.Column]:
+    return [sa.column(n, to_sqla_type(t)) for n, t in schema.items()]
 
 
 class _AlchemyTableSetFormatter(TableSetFormatter):
@@ -85,8 +92,19 @@ class _AlchemyTableSetFormatter(TableSetFormatter):
             schema = ref_op.schema
             result = sa.table(
                 ref_op.name,
-                *(sa.column(n, to_sqla_type(t)) for n, t in schema.items()),
+                *_schema_to_sqlalchemy_columns(schema),
             )
+        elif isinstance(ref_op, ops.SQLStringView):
+            columns = _schema_to_sqlalchemy_columns(ref_op.schema)
+            result = sa.text(ref_op.query).columns(*columns).cte(ref_op.name)
+        elif isinstance(ref_op, ops.View):
+            definition = ref_op.child.compile()
+            result = sa.table(
+                ref_op.name,
+                *_schema_to_sqlalchemy_columns(ref_op.schema),
+            )
+            backend = ref_op.child._find_backend()
+            backend._create_temp_view(view=result, definition=definition)
         else:
             # A subquery
             if ctx.is_extracted(ref_expr):
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -14,6 +14,7 @@ import ibis.expr.schema as sch
 from ibis.backends.base.sql.alchemy import BaseAlchemyBackend
 
 from .compiler import DuckDBSQLCompiler
+from .datatypes import parse_type
 
 
 class Backend(BaseAlchemyBackend):
@@ -69,4 +70,16 @@ class Backend(BaseAlchemyBackend):
         """Return an ibis Schema from a SQL string."""
         with self.con.connect() as con:
             rel = con.connection.c.query(query)
-        return sch.infer(rel)
+        return sch.Schema.from_dict(
+            {
+                name: parse_type(type)
+                for name, type in zip(rel.columns, rel.types)
+            }
+        )
+
+    def _get_temp_view_definition(
+        self,
+        name: str,
+        definition: sa.sql.compiler.Compiled,
+    ) -> str:
+        return f"CREATE OR REPLACE TEMPORARY VIEW {name} AS {definition}"
--- a/ibis/backends/postgres/__init__.py
+++ b/ibis/backends/postgres/__init__.py
@@ -7,13 +7,12 @@ from typing import Literal
 
 import sqlalchemy as sa
 
-import ibis.backends.duckdb.datatypes as ddb
-import ibis.expr.datatypes as dt
 import ibis.expr.schema as sch
 from ibis import util
 from ibis.backends.base.sql.alchemy import BaseAlchemyBackend
 
 from .compiler import PostgreSQLCompiler
+from .datatypes import _get_type
 from .udf import udf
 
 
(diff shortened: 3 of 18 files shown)


ibis/backends/base/sql/alchemy/query_builder.py
```diff
@@ -87,6 +88,33 @@ def _format_table(self, expr):
...
+                    for n, t in ref_expr.schema().items()
+                ),
+            )
+
+            create_view = CreateView(view, definition, or_replace=True)
+            with ref_op.backend.con.connect() as con:
+                con.execute(create_view)
+
+            # drop the view just before interpreter shutdown
+            atexit.register(
```
ssanderson, line 114: :wave: Apologies for the drive-by comment here. I saw a link to this on twitter and was curious about the implementation.

One thing to maybe be nervous about w/r/t cleaning up using `atexit` is that Python can't guarantee that `atexit` handlers actually run. In particular, they won't run in the event of a power failure or if the process receives a `SIGKILL`.

I'm not sure about the various other backends that Ibis supports, but postgres at least has the concept of a temporary view, which is automatically destroyed when the postgres session ends. Using that might be safer for postgres to ensure that created views don't outlive the process.

cpcloud: @ssanderson Yeah, I actually used `CREATE TEMPORARY VIEW` first, but didn't generalize it so that a backend could customize temporary view creation.

I've now made it so that each backend has to implement a `_create_temporary_view` method, which turned out nice since I can remove the `sqlalchemy-views` dependency since only postgres and duckdb from the sqlalchemy backends support it right now.

cpcloud: In the future though, I don't know if there's much better we can do than `atexit` for databases that don't have temporary view creation.

I suppose we could clean up using the `after_execute` method, but then there's still the case of `.sql` and `.alias` expressions that were created but never executed to deal with.

I don't know if there's a robust way to solve this without the help of the database.

ssanderson: > I don't know if there's a robust way to solve this without the help of the database.

Makes sense. You're fundamentally creating an entity that lives outside of the process, so if you want to be robust to sudden death of the process, the thing that handles that death needs to live outside the process.

> I don't know if there's a robust way to solve this without the help of the database.

The other possible option seems like it would be to wrap the custom sql in a CTE and attach it to the query in a way that makes the scoping work out with the desired Ibis semantics? The nice thing about that would be that you wouldn't have to deal with the implications of a semantically readonly operation having side-effects in the database (I could imagine this causing other headaches w/r/t permissions, or name collisions between multiple users). Without knowing more about Ibis' architecture, I'm not sure how hard or easy that would be though.

cpcloud: > Makes sense. You're fundamentally creating an entity that lives outside of the process, so if you want to be robust to sudden death of the process, the thing that handles that death needs to live outside the process.

Indeed.

> The other possible option seems like it would be to wrap the custom sql in a CTE and attach it to the query in a way that makes the scoping work out with the desired Ibis semantics? The nice thing about that would be that you wouldn't have to deal with the implications of a semantically readonly operation having side-effects in the database (I could imagine this causing other headaches w/r/t permissions, or name collisions between multiple users). Without knowing more about Ibis' architecture, I'm not sure how hard or easy that would be though.

I went down this path initially 😄, and I agree it would be the ideal solution. It was pretty challenging and I couldn't come up with something that I was happy with.

cpcloud: Alright, I was able to cut out one temporary view creation out by using CTEs for the `.sql` method. We still need the the temp view for `.alias` for now, since any `.sql` expressions that use the result of `.alias` won't have the entire query available (it does, but the architecture doesn't support glomming these together without a pile of hacks).

With that, I think I am done being nerd sniped for now 😂.

cpcloud:
@tswast Out of curiosity is there a cheap way to get the schema for a BigQuery query using the BigQuery Python API?

ibis/backends/postgres/__init__.py
```diff
@@ -179,3 +180,88 @@ def udf(
             name=name,
             language=language,
         )
+
+    def _get_schema_using_query(self, query: str) -> sch.Schema:
```
kszucs, line 184: Can you move these utilities to `postgres/datatypes.py`?

ibis/backends/duckdb/__init__.py
```diff
@@ -61,3 +69,13 @@ def fetch_from_cursor(
     ):
         df = cursor.cursor.fetch_df()
         return schema.apply_to(df)
+
+    def _get_schema_using_query(self, query: str) -> sch.Schema:
```
kszucs, line 73: Seems like it will need a rebase after #3656

ibis/expr/types/relations.py
```diff
@@ -1177,6 +1181,30 @@ def prevent_rewrite(self, client=None) -> TableExpr:
     def materialize(self) -> TableExpr:
         return self
 
+    def alias(self, alias: str) -> ir.TableExpr:
```
kszucs, line 1184: I find it a little confusing, could't we call it `view`?

ibis/expr/types/relations.py
```diff
@@ -1177,6 +1181,30 @@ def prevent_rewrite(self, client=None) -> TableExpr:
...
+        expr = ops.View(
+            child=self,
+            name=alias,
+            backend=self._find_backend(),
+        ).to_expr()
+
+        # NB: calling compile is necessary so that any temporary views are
+        # created so that we can infer the schema without executing the entire
+        # query
+        expr.compile()
```
kszucs, line 1196: I don't follow, can you point me where views get created?

cpcloud: https://github.com/ibis-project/ibis/pull/3642/files#diff-9f928791a077be8718dc80822ee8d11db44bf4f1fac268e6897232ed46255d99R105

ibis/expr/types/relations.py
cpcloud: `view` is already a method that creates a referentially distinct alias to a relation, and is unrelated to a temporary name.

ibis/expr/operations/relations.py
```diff
@@ -981,6 +981,33 @@ def __component_eq__(
         )
 
 
+@public
+class View(PhysicalTable):
+    """A view created from an expression."""
+
+    child = rlz.table
+    name = rlz.instance_of(str)
+    backend = rlz.client
```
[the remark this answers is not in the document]
cpcloud: To avoid repeated calls to `_find_backend`. I can probably remove that until we know it's actually a perf issue.

ibis/backends/postgres/__init__.py
cpcloud: Not really, this is a method on the backend.

cpcloud: Or are you talking about the datatype mapping stuff?

ibis/expr/operations/relations.py
```diff
@@ -981,6 +981,33 @@ def __component_eq__(
...
+
+    @cached_property
```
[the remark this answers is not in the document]
cpcloud: True, will fix.

ibis/backends/duckdb/__init__.py

ibis/expr/operations/relations.py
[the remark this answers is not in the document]
cpcloud: Done.

ibis/expr/operations/relations.py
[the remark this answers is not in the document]
cpcloud: Done.

ibis/backends/postgres/__init__.py
cpcloud: Moved the mapping code to `datatypes.py`.

kszucs (approved):
I'm not in favor of the side effect, but with a follow-up ticket to refactor it to CTEs it's LGTM.

tobymao:
@cpcloud I wonder if my library https://github.com/tobymao/sqlglot would be useful here. It would allow users to write SQL in various input dialects, like hive, duckdb, or presto, which could then be transpiled into the actual execution engine.

cpcloud:
@tobymao Ha! `sqlglot` just showed up on my radar yesterday. We should definitely look into adding an API that allows mixing SQL dialects for `.sql`.

PR #3643, opened 2022-03-24
refactor(api): add no-op materialize

cpcloud:
This PR adds a no-op materialize along with a deprecation warning.

This will help us break less code out in the wild, since in many cases
materialize was a pass through.


saulpw (approved):
Could be `instead='nothing; .materialize() now has no effect'` instead of changing all `deprecated` usage.  But this is also fine, I had to work around one case myself grammatically too.

cpcloud:
> Could be `instead='nothing; .materialize() now has no effect'` instead of changing all `deprecated` usage. But this is also fine, I had to work around one case myself grammatically too.

There are 3 or 4 cases of `use {foo}` not making sense, so it seemed worthwhile to do the simpler thing and just put the string there.

PR #3645, opened 2022-03-24
fix(repr): allow ops.TableNode in fmt_value

gforsyth:
Fixes a minor regression in the new repr

### DIFF
--- a/ibis/expr/format.py
+++ b/ibis/expr/format.py
@@ -569,6 +569,18 @@ def _fmt_value_physical_table(op: ops.PhysicalTable, **_: Any) -> str:
 
 
 @fmt_value.register
+def _fmt_value_table_node(
+    op: ops.TableNode, *, aliases: Aliases, **_: Any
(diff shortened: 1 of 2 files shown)


ibis/expr/format.py
```diff
@@ -568,6 +568,18 @@ def _fmt_value_physical_table(op: ops.PhysicalTable, **_: Any) -> str:
...
+@fmt_value.register
+def _fmt_value_table_node(
+    op: ops.TableNode, *, aliases: Aliases, **_: Any
+) -> str:
+    """Format a table as value.
+
+    This function is called when a table is used in a value expression. An
+    example is `table.count()`.
+    """
+    return f"r{aliases[op.table.op()].value}"
```
cpcloud, line 580: You can just use the alias itself, without `r` or `.value`

cpcloud (approved):
Nice find, thanks for the fix!

PR #3655, opened 2022-03-27
feat(datatypes): allow unbounded precision and scale in decimal type

cpcloud:
This PR removes the restriction on the decimal type's precision and scale
fields to remove hacks the support with systems with unbounded decimals.

PR #3656, opened 2022-03-27
feat(postgres/duckdb): implement _get_schema_using_query

cpcloud:
This PR implements the `_get_schema_using_query` method for the PostgreSQL and DuckDB backends.

It sits on top of #3655, since postgres supports unbounded decimals.


ibis/expr/datatypes/duckdb.py
```diff
@@ -52,6 +52,9 @@ def parse_type(text: str, default_decimal_parameters=(18, 3)) -> DataType:
     lparen = spaceless_string("(")
     rparen = spaceless_string(")")
 
+    lbracket = spaceless_string("[")
```
kszucs, line 55: Couldn't we maintain this file under the backend as `ibis/backends/duckdb/datatypes.py`?

I'd like to move the `datatypes/pyarrow.py` to `ibis/backends/pyarrow/datatypes.py` as well.

This way we can make backend implementors to plug-in functionality to ibis.

ibis/tests/expr/test_duckdb_datatype_parser.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+    ),
+)
+
+
+@pytest.mark.parametrize(
+    ("column", "type"),
+    [
+        param(colname, type, id=type.lower())
+        for colname, type in [
```
kszucs, line 66: Could you move this list to a constant instead?

ibis/expr/datatypes/duckdb.py
cpcloud: The thing is that none of these files in datatypes are dependent on backend dependencies and I think that if backends want to reuse this functionality they shouldn't have to depend on the entire other backend. In this PR duckdb and postgres are sharing the the `duckdb` type parser, and any future `pyarrow` backend would probably share this datatype infrastructure with `datafusion`.

I don't think these should live under the backends directory.

kszucs: This is my proposal: https://github.com/ibis-project/ibis/pull/3674

Couldn't we make `ibis.backends.<backend>.datatypes` modules always importable? (in case of arrow we will always need pyarrow to be installed)

kszucs: The mandatory duckdb import https://github.com/ibis-project/ibis/blob/b9809036a39f02e44dea688b7ed06d6374502c7d/ibis/backends/duckdb/__init__.py#L7 isn't actually needed.

kszucs (approved):

PR #3663, opened 2022-03-28
refactor(ir): make schema annotable

kszucs:


cpcloud:
I think all the failures are caused by the change to use `tuple` for `names` and `types`. Probably just need to slog through and track it down.

Alternatively, you could change `names` to `_names` and `types` to `_types` and add `@property`s for both that return `tuple(self._names)`/`tuple(self._types)`.

kszucs:
Since we have a quick CI setup I developed the bad habit to check just one or two backends locally before pushing :) 
Going to update the rest of the backends.

cpcloud (changes requested):
I think we need to keep `schema.names` and `schema.types` as `list`s to preserve backwards compatibility.

Otherwise, this PR is good to go.

PR #3664, opened 2022-03-28
refactor(expr): preserve column ordering in aggregations

gforsyth:
BREAKING CHANGE: Columns that were added or used in an aggregation would
be alphabetically sorted in compiled SQL outputs.  This was a vestige
from when Python dicts didn't preserve insertion order.  Now columns
will appear in the order in which they were passed to `aggregate`

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -356,10 +356,7 @@ class TableExpr(Expr):
         """
         metrics = [] if metrics is None else util.promote_list(metrics)
         metrics.extend(
-            self._ensure_expr(expr).name(name)
-            for name, expr in sorted(
-                kwargs.items(), key=operator.itemgetter(0)
-            )
+            self._ensure_expr(expr).name(name) for name, expr in kwargs.items()
         )
 
         op = self.op().aggregate(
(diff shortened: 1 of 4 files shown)


cpcloud (approved):
LGTM, thanks!

cpcloud:
In case anyone lands here looking for the previous behavior, here's an example:

### `.mutate`

```python
t = ibis.table(dict(a="int64", b="int64"))
new = t.mutate(d=1, c=2)  # columns == ["a", "b", "d", "c"]
old = new[sorted(new.columns)]
```

### `.aggregate`

```python
t = ibis.table(dict(a="int64", b="int64"))
new = t.aggregate(d=t.a.sum(), c=t.b.mean())  # columns == ["a", "b", "d", "c"]
old = new[sorted(new.columns)]
```

ibis/expr/types/relations.py
```diff
@@ -356,10 +356,7 @@ def aggregate(
         """
         metrics = [] if metrics is None else util.promote_list(metrics)
         metrics.extend(
-            self._ensure_expr(expr).name(name)
-            for name, expr in sorted(
-                kwargs.items(), key=operator.itemgetter(0)
-            )
+            self._ensure_expr(expr).name(name) for name, expr in kwargs.items()
```
cpcloud, line 359: This also happens in `mutate`, would you mind fixing it there as well?

gforsyth:
Removed the sorting behavior from `mutate`, as well.  I also removed the test for whether mutate sorted keys

cpcloud (approved):

PR #3665, opened 2022-03-28
chore(dev): speed up black a bit

cpcloud:
This small PR speeds up black a bit by depending on its built-in exclude rules. I did the same for isort, but it did not yield any speed up


gforsyth:
I'm confused about what part of this is switching all double-quotes to single-quotes in the generated `setup.py`.  Not really a huge deal, just weird.

cpcloud:
It was because we were running black using its API, which wasn't properly following our current string normalization setting, whereas this PR uses our settings (which was the intent).

gforsyth (approved):

PR #3666, opened 2022-03-28
update docstrings to match new expr repr

gforsyth:
I haven't gone through everything, but the largest chunk of repr docstrings are in `ibis/expr/` and I've updated all of those to reflect the new expr repr.  I also ran across a repr that needed `ops.TimeAdd` so I've added that to the mapping of binop formatting options

### DIFF
--- a/ibis/expr/types/core.py
+++ b/ibis/expr/types/core.py
@@ -154,25 +154,12 @@ class Expr:
         >>> f = lambda a: (a + 1).name('a')
         >>> g = lambda a: (a * 2).name('a')
         >>> result1 = t.a.pipe(f).pipe(g)
-        >>> result1  # doctest: +NORMALIZE_WHITESPACE
-        ref_0
-        UnboundTable[table]
-          name: t
-          schema:
-            a : int64
-            b : string
-        a = Multiply[int64*]
-          left:
-            a = Add[int64*]
-              left:
-                a = Column[int64*] 'a' from table
-                  ref_0
-              right:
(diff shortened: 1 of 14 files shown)


ibis/expr/types/core.py
```diff
@@ -353,11 +340,7 @@ def _binop(
     >>> import ibis.expr.operations as ops
     >>> expr = _binop(ops.TimeAdd, ibis.time("01:00"), ibis.interval(hours=1))
     >>> expr
-    time = TimeAdd
-      left:
-        value: time = datetime.time(1, 0)
-      right:
-        value: interval<int8>(unit='h') = 1
+    datetime.time(1, 0) + 1
```
cpcloud, line 343: We should definitely repr the `1` here more intelligently. I'll make an issue.

cpcloud (approved):
Excellent.

gforsyth:
I'm going to move the `NORMALIZE_WHITESPACE` and friends into the `pyproject.toml` and the add in a CI check just to check the doctests in `ibis/expr` which covers almost all of the important ones

ibis/expr/types/core.py
cpcloud: Done. #3668

gforsyth:
I opted to leave out the CI check -- I know I've felt pretty annoyed at them in the past.  But the configuration is all there if we decide to add them in, and they can be run locally without issue (for occasional checkups and cleanups)

PR #3667, opened 2022-03-28
perf(ir): add back `_str` property

cpcloud:
This PR adds back the `_str` attribute that was removed in 088169a which caused a repr performance regression.


cpcloud:
We can leave this as is for now. The additional memory overhead and additional complexity required to deal with the immutability is probably not worth the performance.

PR #3671, opened 2022-03-29
fix: regression in Selection.sort_by with resolved_keys

gforsyth:
Sibling PR to #3637 -- I ran across this doing some unrelated debugging. The test is a bit wacky but was the first way I found to reliable reach the code path in question


ibis/backends/tests/test_generic.py
```diff
@@ -397,3 +397,10 @@ def test_dropna_table(backend, alltypes, how, subset):
     # is type object in Pyspark, and type bool in Pandas. This diff does
     # not exist in Java 11.
     backend.assert_frame_equal(result, expected, check_dtype=False)
+
+
+@pytest.mark.notimpl(["dask"])
+def test_select_sort_sort(alltypes):
+    query = alltypes[alltypes.year, alltypes.bool_col]
+    query = query.sort_by(query.year).sort_by(query.bool_col)
+    query.execute()
```
cpcloud, line 406: Is the `execute` required?

gforsyth: Nope, I'll remove it

PR #3672, opened 2022-03-29
fix(clickhouse): use a context manager for execution

cpcloud:
Previously we ignored "ignored exception" warnings when running `pytest`.

It turns out that these were coming from ibis's slight misuse of the clickhouse-driver
Client, where we weren't closing the connection when an error occurred during execution of
an `xfail` test.

Using the `Client` instance as a context manager handles this for us, and the test is that
no new failures occur from removing the warning ignorance.


kszucs:
@cpcloud strange but some of the pyspark and impala tests have started to fail

kszucs (approved):

PR #3676, opened 2022-03-29
feat(ci): automatically convert imports to relative ones

kszucs:
Put up another PR with absolute imports for comparison: #3679

### DIFF
--- a/ibis/common/geospatial.py
+++ b/ibis/common/geospatial.py
@@ -2,7 +2,8 @@ from typing import Iterable, List, TypeVar
 
 import ibis.expr.datatypes as dt
 import ibis.expr.types as ir
-from ibis.common import exceptions as ex
+
+from . import exceptions as ex
 
 NumberType = TypeVar('NumberType', int, float)
 # Geometry primitives (2D)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,16 @@ repos:
         types:
           - python
         types_or: []
+      - entry: absolufy-imports --never
+        name: relatify
+        exclude: ^.*tests.*$
+        id: relatify
+        language: python
+        files: ""
+        stages:
+          - commit
+        types:
+          - python
       - entry: isort --check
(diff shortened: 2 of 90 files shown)


ibis/common/geospatial.py
```diff
@@ -2,7 +2,8 @@
 
 import ibis.expr.datatypes as dt
```
kszucs, line 3: `absolufy` doesn't seem to handle plain imports, but I guess that's OK.

cpcloud:
@kszucs Can we close this one or #3679?

kszucs:
> @kszucs Can we close this one or #3679?

I put up both so we can decide which one to prefer. Once we have a clear preference I'll close the other.

ibis/common/geospatial.py
MarcoGorelli: Hey, thanks for using `absolufy-imports` - I don't think plain imports can be written as relative, can they?

kszucs: Hi @MarcoGorelli! Thanks for chiming in!

It should be possible to rewrite them an relative, in this case `from ..expr import datatypes as dt`

MarcoGorelli: true, but that would change it from a direct import to a from import, not sure if people would want that...worth considering anyway, thanks!

gforsyth:
> I put up both so we can decide which one to prefer. Once we have a clear preference I'll close the other.

I don't have a strong opinion one way or the other re: absolute vs. relative, but I'm definitely on board with standardizing on one or the other.

Given that `absolufy` currently doesn't convert plain imports to relative `from` imports, it seems like we'd get more consistency from standardizing on absolute imports.  So if it just comes down to "pick one", I think I'd lean towards absolutes.

cpcloud:
I would agree with @gforsyth. Let's stick with absolute.

cpcloud:
@kszucs Should we close this out and go with #3679?

kszucs:
Closing in favor of #3679

PR #3276, opened 2022-01-10
refactor(ir): simplify expressions by not storing dtype and name

kszucs:
The point here is to not store any data in the expression classes but only the user facing API. This will enable us to keep a simpler operation hierarchy (without intermediare expressions) wrapped with a single top level user facing expression. 

A pretty small objective ended up as an invasive chage which highlights how coupled these module are. 

Though I managed to fix the majority of the issues, there are still multiple errors left.

PR #3681, opened 2022-03-29
refactor(common): more convenient way to define cached properties for Annotable subclasses

kszucs:
Should be useful for scenarios like `output_dtype`, `output_shape` and `output_type` defined in https://github.com/ibis-project/ibis/pull/3276

The following class calculates and stores `double_a`'s output value in a slot like it were a cached property.

```py
    class ValueOp(Annotable):
        a = IsInt

        @promoter
        def double_a(self):
            return 2 * self.a
```

TODO: I'd like to mimic the use case of `output_dtype` in additional unit tests, meanwhile preliminary reviews are welcome.

### DIFF
--- a/ibis/common/grounds.py
+++ b/ibis/common/grounds.py
@@ -8,7 +8,7 @@ from weakref import WeakValueDictionary
 from ibis.util import frozendict
 
 from .caching import WeakCache
-from .validators import Optional, Validator
+from .validators import ImmutableProperty, Optional, Validator
 
 EMPTY = inspect.Parameter.empty  # marker for missing argument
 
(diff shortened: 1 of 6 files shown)


ibis/expr/operations/relations.py
```diff
@@ -127,6 +127,7 @@ def _clean_join_predicates(left, right, predicates):
         elif isinstance(pred, str):
             pred = left[pred] == right[pred]
         elif not isinstance(pred, ir.Expr):
+            print(type(pred))
```
cpcloud, line 130:
```suggestion
```

ibis/common/grounds.py
```diff
@@ -9,7 +9,7 @@
 
 from .. import util
 from .caching import WeakCache
-from .validators import Optional, Validator
+from .validators import Optional, Promoter, Validator
```
cpcloud, line 12: Can we call this `SlottedProperty` or `ImmutableProperty` (and change the corresponding decorator? It doesn't seem clear what the relationship between promotion and caching immutable frequently accessed properties is.

kszucs:
> This is pretty sweet, just a little unsure of the `Promoter` name.

Me too, brought it from the output type rules. I agree that we should rename it, just can't come up with a good name. Suggestions?

kszucs:
How about calling it `@attribute`?

cpcloud:
@kszucs `@slotted_property` or `@immutable_property` are other options.

cpcloud:
I guess another name might be `@frozen_attr` or `@immutable_attr`.

kszucs:
`@ immutable_property` SGTM, renaming it.

cpcloud (approved):
Sweet, LGTM!

PR #3683, opened 2022-03-30
refactor(common): use a specialized curry alternative for validation

kszucs:
With a slightly less convenient rules API we should achieve some speedup for argument validation, still need to write benchmarks though.

### DIFF
--- a/ibis/common/validators.py
+++ b/ibis/common/validators.py
@@ -10,7 +10,7 @@ EMPTY = inspect.Parameter.empty  # marker for missing argument
 
 
 class Validator:
-    __slots__ = tuple()
+    __slots__ = ()
 
     def validate(self, arg, **kwargs):
         raise NotImplementedError()
@@ -38,7 +38,7 @@ class Optional(Validator):
         self.default = default
 
     def __repr__(self):
-        return f"Optional({self.validator!r}, {self.default!r})"
+        return f"{self.__class__.__name__}({self.validator!r}, {self.default!r})"
 
     def validate(self, arg, **kwargs):
         if arg is None:
@@ -59,7 +59,7 @@ class CurriedValidator(Validator):
     def __init__(self, func, args=None, kwargs=None):
         # TODO(kszucs): validate that func is callable
         self.func = func
-        self.args = args or tuple()
+        self.args = args or ()
         self.kwargs = kwargs or {}
 
     def __call__(self, *args, **kwargs):
@@ -69,7 +69,7 @@ class CurriedValidator(Validator):
 
     def __repr__(self):
         return (
-            f"CurriedValidator({self.func!r}, {self.args!r}, {self.kwargs!r})"
+            f"{self.__class__.__name__}({self.func!r}, {self.args!r}, {self.kwargs!r})"
         )
 
     def validate(self, arg, **kwargs):
--- a/ibis/common/validators.py
+++ b/ibis/common/validators.py
@@ -2,9 +2,6 @@ from __future__ import annotations
 
 import inspect
 from contextlib import suppress
-from typing import Callable
-
-import toolz
 
 from ..common.exceptions import IbisTypeError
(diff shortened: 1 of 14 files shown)


ibis/common/validators.py
```diff
@@ -2,20 +2,18 @@
...
 
 EMPTY = inspect.Parameter.empty  # marker for missing argument
 
 
-class Validator(Callable):
-    """
-    Abstract base class for defining argument validators.
-    """
+class Validator:
+    __slots__ = tuple()
```
cpcloud, line 13:
```suggestion
    __slots__ = ()
```

ibis/common/validators.py
```diff
@@ -30,13 +28,19 @@ class Optional(Validator):
...
+        'default',
+    )
 
     def __init__(self, validator, default=None):
         self.validator = validator
         self.default = default
 
-    def __call__(self, arg, **kwargs):
+    def __repr__(self):
+        return f"Optional({self.validator!r}, {self.default!r})"
```
cpcloud, line 41:
```suggestion
        return f"{self.__class__.__name__}({self.validator!r}, {self.default!r})"
```

ibis/common/validators.py
```diff
@@ -45,23 +49,35 @@ def __call__(self, arg, **kwargs):
...
+        new_kwargs = {**self.kwargs, **kwargs}
+        return self.__class__(self.func, new_args, new_kwargs)
 
     def __repr__(self):
-        return '{}({}{})'.format(
-            self.func.__name__,
-            repr(self.args)[1:-1],
-            ', '.join(f'{k}={v!r}' for k, v in self.keywords.items()),
+        return (
+            f"CurriedValidator({self.func!r}, {self.args!r}, {self.kwargs!r})"
```
cpcloud, line 72:
```suggestion
            f"{self.__class__.__name__}({self.func!r}, {self.args!r}, {self.kwargs!r})"
```

ibis/common/validators.py
```diff
@@ -45,23 +49,35 @@ def __call__(self, arg, **kwargs):
...
-    Enable convenient validator definition by decorating plain functions.
-    """
+class CurriedValidator(Validator):
+
+    __slots__ = ("func", "args", "kwargs")
+
+    def __init__(self, func, args=None, kwargs=None):
+        # TODO(kszucs): validate that func is callable
+        self.func = func
+        self.args = args or tuple()
```
cpcloud, line 62:
```suggestion
        self.args = args or ()
```

ibis/common/validators.py
```diff
@@ -30,13 +28,19 @@ class Optional(Validator):
...
 
     def __init__(self, validator, default=None):
         self.validator = validator
         self.default = default
 
-    def __call__(self, arg, **kwargs):
+    def __repr__(self):
+        return f"Optional({self.validator!r}, {self.default!r})"
+
+    def validate(self, arg, **kwargs):
```
cpcloud, line 43: Why can't we preserve `__call__`?

kszucs: It's an experimental PR to see the performance gain if any. 

Preserving `__call__` would require a decent amount of signature tracking, see the `toolz.curry` implementation: https://github.com/pytoolz/toolz/blob/f3ac8211f510edff77c1479061d99bb2dc8ab878/toolz/functoolz.py#L170-L377

cpcloud:
@kszucs needs a rebase, then we can merge it.

kszucs:
If there is no significant benchmark improvement then we should keep the earlier curry based implementation since that's more convenient.

cpcloud:
> If there is no significant benchmark improvement then we should keep the earlier curry based implementation since that's more convenient.

Got it. It might be useful to see whether this PR fixes rules docstrings, which are currently not working for anything curried.

cpcloud:
I futzed with this a bit to see if it helps with the `__doc__` issue, and it doesn't, so we can close it out.

PR #3692, opened 2022-04-01
fix(dask): workaround dask/dask#8857

cpcloud:
This PR works around a bug introduced in merge_asof in https://github.com/dask/dask/pull/8857, where when the `on` argument is falsey the `right_on` columns are duplicated in the right dataframe. The subsequent call to `reset_index` in the `merge_asof` implementation fails.


gforsyth (approved):
looks good

gforsyth:
xref dask/dask#8874

PR #3700, opened 2022-04-04
docs: add .sql and .alias method docstrings

cpcloud:


ibis/expr/types/relations.py
```diff
@@ -1129,6 +1129,42 @@ def materialize(self) -> TableExpr:
         return self
 
     def alias(self, alias: str) -> ir.TableExpr:
+        """Create a table expression with a specific name `alias`.
```
kszucs, line 1132: Add a note/warning about the side effect?

cpcloud: Yep, I'll make it clear that this side is in no way part of the public API.

cpcloud: Done.

kszucs (approved):

PR #3703, opened 2022-04-04
fix: filter SettingWithCopyWarning to silence user-facing warning

### LINKED ISSUE #3702: SettingWithCopyWarning on table select using pandas backend
```
import ibis
import pandas as pd

conn = ibis.pandas.connect({'foo': pd.read_csv('foo.csv', dtype=str)})

foo_table = conn.table('foo')

foo_table.select('col')
```

Yields a SettingWithCopyWarning warning:

```
SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  result[new_col] = result.loc[:, from_col]
/home/patrick/dev/repos/ibis/ibis/backends/pandas/execution/selection.py:336: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  result[new_col] = result.loc[:, from_col]
```

This warning is also blasted when selecting one column in a list:

```
foo_table.select(['col'])
```

This warning, however, is not blasted when selecting by multiple columns in a list:

```
foo_table.select(['col', 'col1'])
```

gforsyth:
The warning is valid in many contexts, but it's spurious here and it's
quite loud.

Despite my memory of `warnings.filterwarnings` serving as a contextmanager, it
doesn't appear to do so, so I've created a simple one that silences the pandas
warning in question, then restores it on exit.

Fixes #3702

### DIFF
--- a/ibis/backends/pandas/execution/selection.py
+++ b/ibis/backends/pandas/execution/selection.py
@@ -333,7 +333,8 @@ def build_df_from_selection(
             renamed_cols[from_col] = to_cols[0]
         else:
             for new_col in to_cols:
-                result[new_col] = result.loc[:, from_col]
+                if from_col != new_col:
+                    result[new_col] = result[from_col]
 
     if renamed_cols:
         result = result.rename(columns=renamed_cols)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -188,7 +188,6 @@ filterwarnings = [
   'ignore:Using \.astype to convert from timezone-(naive|aware) dtype:FutureWarning',
   "ignore:This pattern is interpreted as a regular expression, and has match groups:UserWarning",
   "ignore:The default dtype for empty Series will be 'object':FutureWarning",
-  'ignore:\nA value is trying to be set on a copy of a slice from a DataFrame:',
   "ignore:the `interpolation=` argument to percentile was renamed to `method=`:DeprecationWarning",
   "ignore:Explicitly passing `name=None`:FutureWarning",
   # the warning suggestion to simply remove the argument in the call isn't correct


cpcloud:
Would https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings work? Something like

```python
with warnings.catch_warnings():
    warnings.filterwarnings(
        action="ignore", category=pd.core.common.SettingWithCopyWarning
    )
    ...
```

gforsyth:
Lol, definitely forgot how to use the catch warnings context manager. I'll
make the change

jreback:
pls don't filter the warning itself

you need to copy the frame after filtering

ibis/backends/pandas/execution/selection.py
```diff
@@ -333,7 +352,8 @@ def build_df_from_selection(
             renamed_cols[from_col] = to_cols[0]
         else:
             for new_col in to_cols:
-                result[new_col] = result.loc[:, from_col]
```
jreback, line 336: you need to copy the frame before here

jreback (changes requested):

ibis/backends/pandas/execution/selection.py
cpcloud: What is the material difference for users of ibis other than slower execution? Is there a concrete example showing the problem that involves this code path?

jreback: you should never suppress these - they may change and the assignment is a bit janky - don't use .loc here as it's not idiomatic

jreback: also fixing it properly will not involve a perf hit if u just filter

cpcloud: > you should never suppress these - they may change and the assignment is a bit janky - don't use .loc here as it's not idiomatic

Perhaps you can suggest a change that is compatible with both dask and pandas then. This code path is shared needs to work with both without a deeper refactoring.

jreback:
```
result[to_col] = result[from_col]
```

cpcloud: This is fixed.

cpcloud (approved):

cpcloud:
@jreback I would like to merge this today. Please review.

jreback:
looks fine

cpcloud:
@jreback Can you approve?

jreback (approved):

PR #3710, opened 2022-04-05
feat(repr): add interval info to interval repr

### LINKED ISSUE #3668: bug: repr for interval literals doesn't show unit information
`interval` literal values don't show any information about their units, so all values show up looking the same:

```
In [4]: import ibis

In [5]: ibis.interval(1, unit="s")
Out[5]: 1

In [6]: ibis.interval(1, unit="h")
Out[6]: 1
```

gforsyth:
trying this out.  there are a lot of ways to make this more complicated
and some of them might be worth it, but this seems to provide a little
more information with a tiny footprint.

closes #3668

in re: other options:
If we don't want to special-case the Interval repr, we could redefine Interval as a subclass of Literal so we can catch it explicitly in the single dispatch.

Other possible changes / improvements:
More verbose unit descriptions (`seconds` instead of `s`)

### DIFF
--- a/ibis/expr/format.py
+++ b/ibis/expr/format.py
@@ -534,6 +534,8 @@ def _fmt_value_negate(op: ops.Negate, *, aliases: Aliases) -> str:
 
 @fmt_value.register
 def _fmt_value_literal(op: ops.Literal, **_: Any) -> str:
+    if isinstance(op.dtype, dt.Interval):
+        return f"{op.value} {op.dtype.unit}"
     return repr(op.value)
 
 


ibis/expr/format.py
```diff
@@ -534,6 +534,8 @@ def _fmt_value_negate(op: ops.Negate, *, aliases: Aliases) -> str:
 
 @fmt_value.register
 def _fmt_value_literal(op: ops.Literal, **_: Any) -> str:
+    if isinstance(op.dtype, dt.Interval):
+        return f"{repr(op.value)} {op.dtype.unit}"
```
cpcloud, line 538:
```suggestion
        return f"{op.value} {op.dtype.unit}"
```

I don't think you need the additional `repr` since `op.value`'ll be embedded in the string.

cpcloud (approved):
LGTM.

ibis/expr/format.py
gforsyth: good call, removed.

PR #3722, opened 2022-04-06
chore(dev): make it easier to run benchmarks

cpcloud:
This PR makes it much easier to run benchmarks locally


gforsyth (approved):
Can confirm, this made it much easier to run the benchmarks locally

PR #3728, opened 2022-04-07
feat(ver): dynamic version identifiers

gforsyth:
This adds a dynamic version identifier to Ibis installs -- this should
have no discernable impact on users installing from `conda` or
`pypi` (or from released versions generally).

For users who install from source, it generates a version of the form:

```
<most_recent_release>-post.<numcommits_since_release>+<short_sha_of_HEAD>(.dirty)?
```

The `dirty` suffix is added if there are unstaged changes when the
install is performed.
This will help us help users and other devs track which exact version is
installed when troubleshooting / comparing performance.

The `version` identifier in `pyproject.toml` is not affected by this
process and should display the last released version (currently 2.1.1).

The `__version__` string in `ibis/__init__.py` is now also hard-coded to `"2.1.1"`

During a release, `poetry version` will bump the version in the
`pyproject.toml` file.  I've also added a plugin for the semantic
release bot to automatically replace the version string in the
`__init__.py` file.

There's not a huge rush on merging this -- either before or after the 3.0.0 release should work fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ Icon?
 # documentation files
 docs/source/generated
 docs/source/generated-notebooks
+*.org
 
 # Ibis testing data
 ci/ibis-testing-data*
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -20,6 +20,27 @@
       }
     ],
     [
+      "@google/semantic-release-replace-plugin",
+      {
+        "replacements": [
+          {
+            "file": ["ibis/__init__.py"],
+            "from": "__version__ = \".*\"",
+            "to": "__version__ = \"${nextRelease.version}\"",
+            "results": [
+              {
+                "file": "ibis/__init__.py",
+                "hasChanged": true,
+                "numMatches": 1,
+                "numReplacements": 1
+              }
+            ],
+            "countMatches": true
+          }
+        ]
+      }
+    ],
+    [
       "@semantic-release/github",
       {
         "successComment": false,
@@ -29,7 +50,12 @@
     [
       "@semantic-release/git",
       {
-        "assets": ["pyproject.toml", "docs/release_notes.md", "setup.py"],
+        "assets": [
+          "pyproject.toml",
+          "docs/release_notes.md",
+          "setup.py",
+          "ibis/__init__.py"
+        ],
         "message": "chore(release): ${nextRelease.version}"
       }
     ]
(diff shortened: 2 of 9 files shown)


.releaserc.json
```diff
@@ -32,6 +32,27 @@
         "assets": ["pyproject.toml", "docs/release_notes.md", "setup.py"],
```
cpcloud, line 32:
```suggestion
        "assets": ["pyproject.toml", "docs/release_notes.md", "setup.py", "ibis/__init__.py"],
```

This change is necessary for semantic-release to commit the changes from the replace plugin.

ibis/__init__.py
```diff
@@ -21,10 +21,7 @@
 __all__ = ['api', 'ir', 'util', 'IbisError', 'options']
 __all__ += api.__all__
 
-try:
-    __version__ = importlib_metadata.version(__name__)
-except Exception:
-    __version__ = importlib_metadata.version("ibis-framework")
```
cpcloud, line 27: Yay, I hated this code.

poetry-overrides.nix
```diff
@@ -72,4 +72,10 @@ self: super:
       substituteAllInPlace ./nbconvert/exporters/templateexporter.py
     '';
   });
+
+  poetry-dynamic-versioning = super.poetry-dynamic-versioning.overridePythonAttrs (attrs: {
+    nativeBuildInputs = (attrs.nativeBuildInputs or [ ]) ++ [
+      self.poetry
```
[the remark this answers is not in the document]
gforsyth: probably?  I'll try it out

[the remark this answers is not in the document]
gforsyth: yep!

.gitignore
```diff
@@ -48,6 +48,7 @@ Icon?
 # documentation files
 docs/source/generated
 docs/source/generated-notebooks
+*.org
```
cpcloud, line 51: I see what you did there.

gforsyth: I suppose I could've just added an override to have `org-mode` treat the `.egg` suffix like a org file?

cpcloud: 🤷🏻 I personally do not mind this here.

cpcloud (changes requested):
Ah, I forgot: you need to add `-p "@google/semantic-release-replace-plugin"` to `ci/release/run.sh` and also do the same in `ci/release/dry_run.sh` and tack it onto the `--prepare` list in `ci/release/dry_run.sh`.

Should be good to go after that!

cpcloud (approved):
LGTM, might as well merge this for 3.0.0 and work out any kinks.

PR #3737, opened 2022-04-08
fix(sql): allow joins with joins on the right

### LINKED ISSUE #3729: bug: joins with join on the right hand side are not implemented
Paring down the example from #1045, this test fails:

```python
def test_1045(con):
    con.raw_sql("drop table if exists test1")
    con.raw_sql("drop table if exists test2")
    con.raw_sql("drop table if exists test3")

    con.raw_sql("create table test1 (id1 integer not null, val1 real)")
    for i in range(10):
        con.raw_sql(
            "insert into test1 values (1, 10), (2, 20), (3, 20), (2, 21)"
        )

    con.raw_sql(
        "create table test2 (id2a integer not null, id2b bigint, val2 double precision)"
    )
    for _ in range(10):
        con.raw_sql("insert into test2 values (1, 4, 15), (2, 5, 25)")

    con.raw_sql(
        "create table test3 (id3 text not null, val2 double precision, dt timestamp without time zone)"
    )
    for i in range(10):
        con.raw_sql(
            "insert into test3 values ('4', 15, '2016-07-17 19:55:03'), (2, 25, '2016-07-17 19:55:03')"
        )

    t1 = con.table('test1')
    t2 = con.table('test2')
    t3 = con.table('test3')

    t3 = t3[[c for c in t3.columns if c != "id3"]].mutate(
        id3=t3.id3.cast('int64')
    )

    t3 = t3[[c for c in t3.columns if c != "val2"]].mutate(t3_val2=t3.id3)
    from ibis.util import psql

    t4 = t3.join(t2, t2.id2b == t3.id3)

    t1 = t1[[t1[c].name(f"t1_{c}") for c in t1.columns]]

    t5 = t1.join(t4, t1.t1_id1 == t4.id2a)
```

cpcloud:
This PR allows joins whose right operator is itself a join.

The thing that made this initially tricky was that walking the join
tree isn't enough, because join predicates will refer to table in the right
join before it's available. The problem is one of precedence: the right hand
join must logically be run before its parents, but doesn't guarantee that
in a query like

```python
t3 = t1.join(t2, t1.a == t2.a)
t4.join(t3, t4.b = t3.b)
```

that `t3` will be defined before `t4`.

The solution is to force subquery creation using the `TableExpr.view` method on
right side of the join if the right side is itself a join and then replace the
predicate children with this new subquery.

Closes #3729.


gforsyth (approved):
Nice!

PR #3742, opened 2022-04-09
fix(expr): fix formatting of table info using `tabulate`

cpcloud:
This PR fixes the rendering of `table.info()` using the lightweight `tabulate` library.


kszucs:
Shall we have a test for it?

cpcloud:
At the very least we should invoke it somewhere if we're not already.

PR #3755, opened 2022-04-11
fix(sql): retain column names for named ColumnExprs

### LINKED ISSUE #3754: Base SQL backend discards column names
For a ColumnExpr that isn't a TableColumn, the base SQL backend unconditionally uses the name 'tmp' for the resulting query, rather than retaining the name 

```
>>> import ibis.backends.base.sql
>>> simple_table = ibis.table(
...         [
...             ('int_col', 'int32'),
...         ],
...         'simple_table',
...     )
>>> expr = (simple_table.int_col + 3).name('foo')
>>> ibis.backends.base.sql.Compiler().to_sql(expr)
'SELECT `int_col` + 3 AS `tmp`\nFROM simple_table'
```

It seems like this should result in `'SELECT `int_col` + 3 AS `foo`\nFROM simple_table'`

mikegraham:
If the ColumnExpr is named, keep that name in the SQL query (and eventually in the resulting pandas series) rather than unconditionally using the name `tmp`

fixes #3754

### DIFF
--- a/ibis/backends/base/sql/compiler/select_builder.py
+++ b/ibis/backends/base/sql/compiler/select_builder.py
@@ -317,8 +317,11 @@ class SelectBuilder:
                 # Something more complicated.
                 base_table = L.find_source_table(expr)
 
-                table_expr = base_table.projection([expr.name('tmp')])
-                result_handler = _get_column('tmp')
+                if not expr.has_name():
+                    expr = expr.name('tmp')
+
+                table_expr = base_table.projection([expr])
+                result_handler = _get_column(expr.get_name())
 
             return table_expr, result_handler
         else:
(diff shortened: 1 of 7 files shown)


mikegraham:
I hadn't run the backend tests before creating this PR, as you can see.

If the general approach is something you're interested in, I can fix the tests.

cpcloud:
@mikegraham Thanks for the PR, this seems like a reasonable idea. If you fix up the tests we can go ahead and merge.

cpcloud:
@mikegraham We'd like to release 3.0.0 on April 25th, and it'd be swell if we could get this in before then. Otherwise, we'll definitely get this in the next point release.

cpcloud:
@mikegraham I ended up fixing this up so that we can get it in for 3.0.0.

cpcloud (approved):

mikegraham:
Thanks a million!

PR #3774, opened 2022-04-17
chore(dev): make it easier to render the changelog

cpcloud:
This PR adds two tools for looking at a generated changelog that more closely matches the final output as opposed to looking at the output of ci/release/dry_run.sh which uses a fake repo.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -9,6 +9,7 @@ let
     curl
     duckdb
     git
+    glow
     jq
     just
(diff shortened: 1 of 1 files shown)


cpcloud:
The changelog can be rendered inside a `nix-shell` using:

```
changelog | glow - -w 230
```

Adjust the `-w` (width) parameter as needed.

PR #3798, opened 2022-04-21
experiment(ir): remove intermediate expressions

kszucs:


jreback (changes requested):
can we keep the original values around for a while - this is a major breaking change

kszucs:
@jreback this is just a prototype where we can experiment with pattern matching on operation trees using matchpy

ibis/expr/operations/generic.py
```diff
@@ -273,7 +274,7 @@ def root_tables(self):
 
 
 @public
-class NullLiteral(Literal):
+class NullLiteral(Literal, Singleton):
```
kszucs, line 277: Ohh how handy it is!

kszucs:
In order to allow proper pattern matching on ibis operations we would need to accept three kinds of arguments:
1. Expressions arriving from the standard usage of the API
2. `matchpy.Wildcard` and perhaps other `matchpy.Symbol` objects
3. ibis operations so we can substitute operations without converting to expression before construction 

These three ideally should be achieved by rewriting the current rules to return with ibis operations instead of expressions.

kszucs:
Closing in favor of https://github.com/ibis-project/ibis/pull/4135

PR #3800, opened 2022-04-21
feat(api): unnest

### LINKED ISSUE #1146: feat: support for UNNEST (array flatten operation)
Each of the following systems has a slightly different way of expressing flattening of arrays:

Given a table `t`:

``` sql
CREATE TABLE t (
  elements ARRAY<STRING>
)
```

Hive:

``` sql
SELECT element
FROM t
LATERAL VIEW EXPLODE(elements) s AS element;
```

Presto and PostgreSQL:

``` sql
SELECT s.element
FROM t
CROSS JOIN UNNEST(elements) AS s (element)
```

Impala:

``` sql
SELECT elements.item
FROM t, t.elements
```

BigQuery:

```sql
SELECT element
FROM t.elements

-- or

SELECT element
FROM UNNEST(t.elements)
```

Clickhouse:

```sql
SELECT elements
FROM t ARRAY JOIN elements
```

This issue is to track crafting an API for this operation in Ibis.

cpcloud:
This PR adds an unnest API. Closes #1146.

### DIFF
--- a/ibis/backends/clickhouse/registry.py
+++ b/ibis/backends/clickhouse/registry.py
@@ -739,6 +739,7 @@ operation_registry = {
     ops.ArrayConcat: _fixed_arity('arrayConcat', 2),
     ops.ArrayRepeat: _array_repeat_op,
     ops.ArraySlice: _array_slice_op,
+    ops.Unnest: _unary("arrayJoin"),
 }
 
 
--- a/ibis/backends/clickhouse/registry.py
+++ b/ibis/backends/clickhouse/registry.py
@@ -658,6 +658,7 @@ operation_registry = {
     ops.Sum: _agg('sum'),
     ops.Max: _agg('max'),
     ops.Min: _agg('min'),
+    ops.ArrayCollect: _agg('groupArray'),
     ops.StandardDev: _agg_variance_like('stddev'),
     ops.Variance: _agg_variance_like('var'),
     ops.GroupConcat: _group_concat,
--- a/ibis/backends/postgres/registry.py
+++ b/ibis/backends/postgres/registry.py
@@ -721,5 +721,6 @@ operation_registry.update(
         ops.ArrayRepeat: _array_repeat,
         ops.IdenticalTo: _identical_to,
         ops.HLLCardinality: _hll_cardinality,
+        ops.Unnest: unary(sa.func.unnest),
     }
 )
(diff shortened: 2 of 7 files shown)


tswast:
I'm wondering if this enabled the "table from literals" feature request I just filed? https://github.com/ibis-project/ibis/issues/3832

If so, maybe add a test for that, too?

cpcloud:
> I'm wondering if this enabled the "table from literals" feature request I just filed? #3832
> 
> If so, maybe add a test for that, too?

I don't think so, because this implementation treats `unnest` as a column expression, not a table.

This is actually something I wanted to bring up with you regarding BigQuery. It appears to be the only backend that Ibis supports that has unnest support but doesn't implement a `SELECT UNNEST(...)` version.

DuckDB, postgres, clickhouse and pyspark all implement some version of `SELECT UNNEST` which makes the ibis implementation incredibly simple. I initially implemented the `TableExpr`-based version of `UNNEST` and it was very complex. Are there any plans to add support for `SELECT UNNEST`-style syntax in BigQuery?

tswast:
> some version of SELECT UNNEST which makes the ibis implementation incredibly simple.

Usability-wise, I've found the BigQuery version of UNNEST a lot to wrap my head around. 

> Are there any plans to add support for SELECT UNNEST-style syntax in BigQuery?

I'm not aware of any plans, but I've filed internal Google issue 231230963 to request this support. I think it would be a big usability improvement over the current requirement to do a CROSS JOIN to acheive the equivalent in BigQuery.

cpcloud:
I'm going to merge this in the interest of getting the API in front of users.

jreback:
how is this different than the structure and destructure api?

cpcloud:
> how is this different than the structure and destructure api?

Take a look at #1146, that goes into great detail about unnest.

jreback:
doesn't answer the question

i could be wrong but this is just a specialized destructure

jreback:
cc @icexelloss

cpcloud:
The issue provides sufficient information to answer the question.

jreback:
it does not address nor answer this question in any way (nor even mention it at all)

icexelloss:
If I understand correctly. "Unnest" is "expand an array col to multiple of rows", while `destructure` is "expand an struct column to multiple columns"

e.g. 

Before unnest
|    col        | 
| ---------- | 
| [1, 2]        | 

After unnest
|    col        | 
| ---------- | 
|      1         |
|      2         |


Before Destruct
|    col              | 
| -------------- | 
| {a: 1, b: 2}      | 

After Destruct
|    a                 |   b  |
| -------------- | ---- |
|  1                   |   2  |

PR #3804, opened 2022-04-22
feat(api): underscore convenience api

cpcloud:
This PR is an experiment to play with a way of avoiding `lambda`s using `_`.

Here's an example:

```python
t = ibis.table(dict(a="int64"), name="t")

expr = (
    t.mutate(b=_.a - 2)
    .select([(_.a.sum() + _.b.count()).name("b")])
    .group_by(key=_.b.cast("string"))
    .aggregate(total=_.b.mean())
)
```

The way this works is that a `Deferred` object graph is built, and resolved
element by element as needed, for example, in a projection or aggregation.
Every `Deferred` instance has a `resolve` attribute that accepts an ibis
expression and returns another ibis expression effectively replaying the graph.
The `_` object's `resolve` attribute is the identity function, which resolves to the
table expression from which the rest of the graph is evaluated over.

### DIFF
--- a/ibis/expr/deferred.py
+++ b/ibis/expr/deferred.py
@@ -1,0 +1,138 @@
+from __future__ import annotations
+
+import operator
+from typing import TYPE_CHECKING, Any, Callable
+
+import toolz
+
+if TYPE_CHECKING:
+    import ibis.expr.types as ir
+
+
+class Deferred:
+    """A deferred expression."""
+
+    __slots__ = ("resolve",)
+
+    def __init__(self, resolve: Callable = toolz.identity) -> None:
+        assert callable(
+            resolve
+        ), f"resolve argument is not callable, got {type(resolve)}"
+        self.resolve = resolve
+
+    def _defer(self, func: Callable, *args: Any, **kwargs: Any) -> Deferred:
+        """Wrap `func` in a `Deferred` instance."""
+
+        def resolve(expr, func=func, self=self, args=args, kwargs=kwargs):
+            resolved_expr = self.resolve(expr)
+            resolved_args = [_resolve(arg, expr=expr) for arg in args]
+            resolved_kwargs = {
+                name: _resolve(arg, expr=expr) for name, arg in kwargs.items()
+            }
+            return func(resolved_expr, *resolved_args, **resolved_kwargs)
+
+        return self.__class__(resolve)
+
+    def __getattr__(self, name: str) -> Deferred:
+        return self._defer(getattr, name)
+
+    def __getitem__(self, key: Any) -> Deferred:
+        return self._defer(operator.itemgetter(key))
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Deferred:
+        return self._defer(
+            lambda expr, *args, **kwargs: expr(*args, **kwargs),
+            *args,
+            **kwargs,
+        )
+
+    def __add__(self, other: Any) -> Deferred:
+        return self._defer(operator.add, other)
+
+    def __radd__(self, other: Any) -> Deferred:
+        return self._defer(toolz.flip(operator.add), other)
+
+    def __sub__(self, other: Any) -> Deferred:
+        return self._defer(operator.sub, other)
+
+    def __rsub__(self, other: Any) -> Deferred:
+        return self._defer(toolz.flip(operator.sub), other)
+
+    def __mul__(self, other: Any) -> Deferred:
+        return self._defer(operator.mul, other)
+
+    def __rmul__(self, other: Any) -> Deferred:
+        return self._defer(toolz.flip(operator.mul), other)
+
+    def __truediv__(self, other: Any) -> Deferred:
+        return self._defer(operator.truediv, other)
+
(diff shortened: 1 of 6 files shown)


kszucs:
Pretty nice and concise implementation! Going to have a more thorough look tomorrow.

ibis/expr/deferred.py
```diff
@@ -0,0 +1,121 @@
...
+                self.resolve(expr),
+                *(_resolve(arg, expr=expr) for arg in args),
+                **{
+                    name: _resolve(arg, expr=expr)
+                    for name, arg in kwargs.items()
+                },
+            )
+        )
+
+    def __getattr__(self, name: str) -> Deferred:
```
kszucs, line 34: Also `__getitem__`?

ibis/expr/deferred.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def __neg__(self) -> Deferred:
+        return self._wrap(operator.neg)
+
+
+T = TypeVar("T")
+
+
+@functools.singledispatch
+def _resolve(arg: T, **_: Any) -> T:
```
kszucs, line 115: Is this for making it pluggable? If not could we avoid using singledispatch?

ibis/expr/deferred.py
```diff
@@ -0,0 +1,121 @@
...
+import operator
+from typing import TYPE_CHECKING, Any, Callable, TypeVar
+
+import toolz
+
+if TYPE_CHECKING:
+    import ibis.expr.types as ir
+
+
+class Deferred:
```
kszucs, line 13: We can use `Annotable` here as well.

ibis/expr/types/relations.py
```diff
@@ -170,14 +171,15 @@ def _resolve(self, exprs):
 
     # TODO(kszucs): should be removed
     def _ensure_expr(self, expr):
+        if isinstance(expr, Deferred):
+            return expr.resolve(self)
         if isinstance(expr, str):
             return self[expr]
-        elif isinstance(expr, (int, np.integer)):
```
kszucs, line 175: I tend to prefer more explicit branches using `elif` and `else`.

ibis/expr/deferred.py
```diff
@@ -0,0 +1,121 @@
...
+
+    __slots__ = ("resolve",)
+
+    def __init__(self, resolve: Callable = toolz.identity) -> None:
+        self.resolve = resolve
+
+    def _wrap(self, func: Callable, *args: Any, **kwargs: Any) -> Deferred:
+        """Wrap `func` in a `Deferred` instance."""
+        return self.__class__(
+            lambda expr, func=func, self=self, args=args, kwargs=kwargs: func(
```
kszucs, line 24: Wouldn't it be better to define a local function for better readability?

kszucs: Something like: 

```python
def resolve(expr):
    expr = self.result(expr)
    args = (_resolve(arg, expr=expr) for arg in args)
    kwargs = {name: _resolve(arg, expr=expr) for name, arg in kwargs.items()}
    return func(expr, *args, **kwargs)
```

ibis/expr/deferred.py
cpcloud: I'm not sure we get much by doing that here, and it looks to require use of `rlz` which introduces a circular import.

kszucs: Indeed, perhaps validate that `resolve` is callable then?

ibis/expr/types/relations.py
cpcloud: Done.

ibis/expr/deferred.py
cpcloud: Roger, done.

ibis/expr/deferred.py
cpcloud: Removed.

ibis/expr/deferred.py
cpcloud: Added.

ibis/expr/deferred.py
cpcloud: `resolve` should always be callable, and isn't user facing, so I'll add an `assert`.

cpcloud: Done.

kszucs (approved):

cpcloud:
Since we're releasing today, let's wait to merge this until after that.

cpcloud:
cc @machow, I took inspiration from https://github.com/machow/siuba here 😄

machow:
Hey--glad it was handy inspiration! I was tinkering with ibis, and was using lambdas a lot--so this format would go a long way (for a least the way R people are used to :). cc @pwwang, who also is probably interested in libraries doing this kind of thing..

kszucs:
@cpcloud can we go ahead and merge this?

cpcloud:
Yep. I was waiting to see if any more bugs from the 3.0 release shook out before merging any feature PRs, but I think it's probably fine to start merging these. We'll likely release a 3.1 in a month or two anyway so that folks can pick up non-breaking changes along the way to 4.0.

PR #3809, opened 2022-04-23
feat(clickhouse): add proper type serialization

cpcloud:
This PR adds type parsing and serialization for ClickHouse, as well as fixes existing issues with nullable clickhouse types.

### DIFF
--- a/ibis/backends/clickhouse/datatypes.py
+++ b/ibis/backends/clickhouse/datatypes.py
@@ -1,0 +1,220 @@
+from __future__ import annotations
+
+import functools
+from typing import TYPE_CHECKING
+
+import parsy as p
+
+if TYPE_CHECKING:
+    from ibis.expr.datatypes import DataType
+
+import ibis.expr.datatypes as dt
+
+
+def parse(text: str) -> DataType:
+    @p.generate
+    def datetime():
+        yield dt.spaceless_string("datetime64", "datetime")
+        timezone = yield parened_string.optional()
+        return dt.Timestamp(timezone=timezone, nullable=False)
+
+    primitive = (
+        datetime
+        | dt.spaceless_string("null", "nothing").result(dt.null)
+        | dt.spaceless_string("bigint", "int64").result(
+            dt.Int64(nullable=False)
+        )
+        | dt.spaceless_string("double", "float64").result(
+            dt.Float64(nullable=False)
+        )
+        | dt.spaceless_string("float32", "float").result(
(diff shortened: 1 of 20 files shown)


ibis/expr/datatypes/core.py
```diff
@@ -1692,3 +1726,19 @@ def _geom_to_wkt(
...
+        import ibis.expr.datatypes as dt
+
+        ty = getattr(dt, name)
+        if not issubclass(ty, dt.DataType):
+            raise AttributeError(name)
+
+        return functools.partial(ty, nullable=False)
+
+
+public(n=NonNullables())
```
kszucs, line 1744: I may miss something but I rather dislike this syntax sugar due to the following reasons:
- why is `dt.n.Int64()` is better then `dt.Int64(nullable=False)`?
- why to expose it publicly?
- confusing shorthand, I'd rather think that `dt.n` means `nullable` and `dt.nn` means `not nullable`

ibis/backends/clickhouse/datatypes.py
```diff
@@ -0,0 +1,213 @@
...
+
+import parsy as p
+
+if TYPE_CHECKING:
+    from ibis.expr.datatypes import DataType
+
+import ibis.expr.datatypes as dt
+
+
+def parse_type(text: str) -> DataType:
```
kszucs, line 14: Perhaps simply `parse` and `serialize`?

Have you thought about a "standard" datatypes API for the backends which we could align to?

ibis/expr/datatypes/core.py
kszucs: The PR is a nice improvement otherwise!

cpcloud: Yeah, this was just an experiment in making it more convenient. I'll revert the `n` syntax and keep it verbose for now.

ibis/backends/clickhouse/datatypes.py
cpcloud: Not formally, I think I'm still trying to characterize the set of behaviors that data types need to have.

ibis/expr/datatypes/core.py
cpcloud: Done.

ibis/backends/clickhouse/datatypes.py
cpcloud: Changed to `parse` and `serialize`, deprecated existing `parse_type` (slated for removal in 4).

PR #3823, opened 2022-04-26
docs: Adding updates to breaking changes section and materialize() 

marlenezw:
Changing the breaking changes section to 'Other Changes' and clarifying depreciation of materialize instead of removal of it.


cpcloud (approved):
Made a couple of spelling changes (`depreciation` to `deprecation`), otherwise looks good!

PR #3825, opened 2022-04-27
feat(sql): add support for cross-dialect `.sql` API

cpcloud:
This PR adds support for passing SQL strings of different dialects to the .sql method.


cpcloud:
The complexity of this PR is too high relative to the gain. Until someone asks for it, people can use `sqlglot` outside of ibis and pass in the query to `.sql`.

PR #3830, opened 2022-04-28
fix(build): replace version before exec plugin runs

cpcloud:
This PR moves the version replacement step to execute before exec plugin so that wheels are published with an updated `__init__.py`.


gforsyth (approved):
yep, that makes sense.

PR #3831, opened 2022-04-28
fix(docs): fix tempdir location for docs build

cpcloud:
This PR fixes the possible TEMPDIR mismatch when building docs.


gforsyth (approved):
looks good.  clearly makes sense that his change leads to a 40% drop in coverage, right?

cpcloud:
> looks good. clearly makes sense that his change leads to a 40% drop in coverage, right?

:roflcopter: The coverage is updated incrementally as the various jobs finish, so the situation looks dire initially but sorts itself out eventually.

PR #3836, opened 2022-04-29
fix: use bounded precision for decimal aggregations

### LINKED ISSUE #3833: bug: loss of decimal precision in aggregate in 3.x
With ibis 2.1.1:

```
In [1]: import ibis.expr.datatypes as dt

In [2]: import ibis

In [3]: t = ibis.table(
   ...:     [
   ...:         ("l_quantity", dt.Decimal(15, 2)),
   ...:     ],
   ...:     name="t",
   ...: )

In [4]: lquantsum = t.aggregate(sum_qty=lambda t: t.l_quantity.sum())

In [5]: lquantsum._get_type("sum_qty")
Out[5]: Decimal(precision=38, scale=2, nullable=True)
```

On 3.x.x
```
In [1]: import ibis.expr.datatypes as dt

In [2]: import ibis

In [3]: t = ibis.table(
   ...:     [
   ...:         ("l_quantity", dt.Decimal(15, 2)),
   ...:     ],
   ...:     name="t",
   ...: )

In [4]: lquantsum = t.aggregate(sum_qty=lambda t: t.l_quantity.sum())

In [5]: lquantsum._get_type("sum_qty")
Out[5]: Decimal(precision=None, scale=None, nullable=True)
```

This is causing some understandable issues when serializing to substrait

gforsyth:
This tweaks the new behavior around decimal precision and scale.  If a
decimal has unbounded precision/scale, specified as `(None, None)`, then
the precision/scale of an aggregation of that column will also have
unbounded precision.

For a decimal with some bounded precision/scale, e.g. `(18, 2)`, the
precision/scale of an aggregation of that column will have a precision
of `(38, 2)` which is the previous default maximum for bounded precision
and matches up with _most_ backends.

Fixes #3833

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -440,7 +440,10 @@ class Decimal(DataType):
     @property
     def largest(self):
         """Return the largest type of decimal."""
-        return self.__class__(precision=None, scale=None)
+        return self.__class__(
+            precision=38 if self.precision is not None else None,
+            scale=2 if self.scale is not None else None,
+        )
 
     @property
     def _pretty_piece(self) -> str:
--- a/ibis/tests/expr/test_decimal.py
+++ b/ibis/tests/expr/test_decimal.py
@@ -2,6 +2,7 @@ import operator
 
 import pytest
 
+import ibis
 import ibis.expr.api as api
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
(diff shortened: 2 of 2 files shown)


cpcloud (approved):
LGTM, should this be a `fix:` instead of `refactor:`?

gforsyth:
> LGTM, should this be a `fix:` instead of `refactor:`?

yeah, probably -- updated the commit

saulpw:
So this means that `precision` and `scale` are minimums; if they are given as greater than these 'largest', should we use the max between them? (So platforms which support higher values can use them?)

PR #3855, opened 2022-05-03
refactor(decimal): allow user to specify precision > 38

gforsyth:
The previous fix for decimals imposes an effective limit on decimal
precision -- it can either be 38 or unbounded.  In the case that the
user does want to specify a bounded precision/scale that are larger than
our defaults, they can now do that.


gforsyth:
xref: #3855

gforsyth:
xref: #3855

gforsyth:
xref #3836 (oops)

PR #3916, opened 2022-05-16
refactor(ir): use more consistent naming for internal types and operations

kszucs:
This is a backward compatible change, added two test cases in `ibis/tests/expr/test_operations.py` to ensure that.


timothydijamco:
Thanks for the backwards compatibility with this—just to confirm, should the old names (that are aliased to the new ones), e.g. `ibis.expr.types.TableExpr` be thought of as deprecated / expected to be removed at some point?

PR #3917, opened 2022-05-16
refactor: move AnyToExistsTransform to expression construction time

cpcloud:
This PR moves the transformation of `.any()`/`.notany()` filters with outer
table references to happen at expression construction time, rather than the
current phase which is inside the compiler.

The changes are motivated by the following requirements:

1. Allow representation of `EXISTS` queries using Substrait
2. Simplify the codebase by moving towards removing rewrites during
   compilation.

This PR fixes the modeling of `EXISTS` which is currently incorrectly modeled
as an analytic expression.

`EXISTS` expressions are nothing more than fancy predicates so they are modeled
as `BooleanColumn`s here.

### DIFF
--- a/ibis/expr/operations/logical.py
+++ b/ibis/expr/operations/logical.py
@@ -157,3 +157,39 @@ class Where(Value):
 
     output_dtype = rlz.dtype_like("true_expr")
     output_shape = rlz.shape_like("bool_expr")
+
+
+class _AnyBase(Value):
+    # Depending on the kind of input boolean array, the result might either be
+    # array-like (an existence-type predicate) or scalar (a reduction)
+    arg = rlz.column(rlz.boolean)
+
+    output_dtype = dt.boolean
+
+    @property
+    def _reduction(self):
+        roots = self.arg.op().root_tables()
+        return len(roots) < 2
+
+    @immutable_property
+    def output_shape(self):
+        if self._reduction:
+            return rlz.Shape.SCALAR
+        else:
+            return rlz.Shape.COLUMNAR
+
+    @abc.abstractmethod
+    def negate(self):  # pragma: no cover
+        ...
+
+
+@public
+class Any(_AnyBase):
+    def negate(self):
+        return NotAny(self.arg)
+
+
+@public
+class NotAny(_AnyBase):
+    def negate(self):
+        return Any(self.arg)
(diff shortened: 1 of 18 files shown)


ibis/expr/operations/logical.py
```diff
@@ -157,3 +162,39 @@ class Where(Value):
...
+            return rlz.Shape.COLUMNAR
+
+    @abc.abstractmethod
+    def negate(self):
+        ...
+
+
+@public
+class Any(_AnyBase):
+    def negate(self):
```
kszucs, line 193: Could we remove `negate()` methods? Or mark them as deprecated?

ibis/expr/operations/logical.py
```diff
@@ -157,3 +162,39 @@ class Where(Value):
...
+
+class _AnyBase(Value):
+    # Depending on the kind of input boolean array, the result might either be
+    # array-like (an existence-type predicate) or scalar (a reduction)
+    arg = rlz.column(rlz.boolean)
+
+    output_dtype = dt.boolean
+
+    @property
+    def _reduction(self):
```
kszucs, line 175: I think we should define two seperate operations instead, one is a reduction the other is not.

kszucs: We could construct them from the expression's corresponding method based on `arg`'s type.

ibis/expr/types/relations.py
```diff
@@ -1281,3 +1289,81 @@ def find_base_table(expr):
 
 
 public(TableExpr=Table)
+
+
+class _AnyToExistsTransform:
```
kszucs, line 1294: Could we move this to `analysis.py` instead?

ibis/expr/operations/logical.py
cpcloud: Agree, but not as part of this PR. This PR is only about moving the transformation out of the compiler to expression construction.

ibis/expr/types/relations.py
kszucs: To a module which we wan't to refactor badly :)

ibis/expr/operations/logical.py
cpcloud: Maybe, but in a follow-up.

ibis/expr/operations/logical.py
kszucs: Sure, could you create an issue?

cpcloud: Done: https://github.com/ibis-project/ibis/issues/3925.

PR #3921, opened 2022-05-17
refactor(ir): make NullLiteral a singleton

kszucs:


kszucs:
Just ported your change from #3798 :)

PR #3924, opened 2022-05-17
refactor(ir): remove Node.compatible_with() method

kszucs:


ibis/expr/analysis.py
```diff
@@ -935,9 +935,24 @@ def validate(self, expr):
     def _among_roots(self, node):
         return self.roots_shared(node) > 0
 
+    def compatible_with(self, a, b):
+        # self and other are equivalent except for predicates, selections, or
+        # sort keys any of which is allowed to be empty. If both are not empty
+        # then they must be equal
+        if a.equals(b):
+            return True
```
cpcloud, line 943: It looks like this branch is never hit. Should we remove this and the previous line?

ibis/expr/operations/relations.py
```diff
@@ -439,42 +439,13 @@ def schema(self):
...
         return bool(self.selections)
 
+    @util.deprecated(instead="instantiate Selection directly", version="4.0.0")
     def substitute_table(self, table_expr):
         return Selection(table_expr, self.selections)
 
     def root_tables(self):
         return [self]
 
-    def can_add_filters(self, wrapped_expr, predicates):
```
kszucs, line 448: Should I restore these as well?

cpcloud: Yeah I would keep everything that isn't prefixed with an underscore and mark it deprecated. We can delete them in the 4.0.0 release.

kszucs:
@cpcloud updated

PR #3935, opened 2022-05-18
chore: use full hex sha in git-blame-ignore-rev file

kszucs:


cpcloud:
Probably doesn't need to be `fix` since it's not a thing that affects users.

cpcloud:
Could be `chore(dev): ...`

kszucs:
Updated.

cpcloud (approved):

PR #3936, opened 2022-05-18
refactor(ir): use lin.traverse() to implement is_reduction

kszucs:


ibis/expr/analysis.py
```diff
@@ -1181,17 +1181,13 @@ def is_reduction(expr):
...
-
-        return False
+    def predicate(expr):
+        if getattr(expr.op(), '_reduction', False):
+            return lin.halt, expr
+        else:
+            return lin.proceed, None
 
-    return has_reduction(expr.op() if isinstance(expr, ir.Expr) else expr)
+    return bool(list(lin.traverse(predicate, expr)))
```
cpcloud, line 1190: Do we need to materialize the entire list just to determine whether any of ops are reductions? What about returning a boolean instead of `expr` or `None` and then summing the results?

cpcloud (commented):
Small comment on the implementation.

ibis/expr/analysis.py
kszucs: Updated.

ibis/expr/analysis.py
```diff
@@ -1154,7 +1122,8 @@ def _is_analytic(op):
 
 
 def is_reduction(expr):
```
kszucs, line 1124: Memoization would spare quite some operation here (though would need benchmarks).

cpcloud: Yeah, probably best to leave it as is for now until we get to a better state with `analysis.py`.

cpcloud (approved):

PR #3937, opened 2022-05-18
refactor: move filter rewriting to expression construction

cpcloud:
- refactor: rewrite filters at expression construction time
- chore(pandas/dask): implement TableArrayView
- chore: don't rewrite filters with Window operations

### DIFF
--- a/ibis/expr/analysis.py
+++ b/ibis/expr/analysis.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import functools
 import operator
+from typing import AbstractSet, MutableSequence
 
 import toolz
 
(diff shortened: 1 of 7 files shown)


ibis/expr/analysis.py
```diff
@@ -1265,3 +1265,67 @@ def _visit_table(self, expr):
...
+@_rewrite_filter.register(ops.Literal)
+@_rewrite_filter.register(ops.ExistsSubquery)
+@_rewrite_filter.register(ops.NotExistsSubquery)
+@_rewrite_filter.register(ops.Window)
+def _rewrite_filter_subqueries(_, expr, **kwargs):
+    return expr
+
+
+@_rewrite_filter.register(ops.Binary)
+def _rewrite_filter_binary(op, expr, **kwargs):
```
kszucs, line 1297: Can you add comments describing what each overload does?

ibis/expr/types/relations.py
```diff
@@ -738,7 +738,13 @@ def filter(
         child = self
         for predicate, right in top_ks:
             child = child.semi_join(right, predicate)[child]
-        return an.apply_filter(child, resolved_predicates)
+        return an.apply_filter(
```
kszucs, line 741: What exactly happens here? :)

cpcloud: Ultimately a `Selection` is constructed. Is there something specific you're wanting to know about?

ibis/expr/analysis.py
cpcloud: Done, added docstrings.

ibis/expr/types/relations.py
cpcloud: The point of this rewrite is to turn expressions like `a > a.mean()` into an expression that is meaningful to SQL.

ibis/expr/analysis.py
```diff
@@ -1207,3 +1207,55 @@ def _visit_table(self, expr):
                 and base_table.op() not in self.query_roots
             ):
                 self.foreign_table = expr
+
+
+@functools.singledispatch
```
kszucs, line 1212: What do you think about creating a new package just for rewrite rules (one file per purpose)?

Each file should have a well defined behaviour on the input expression.

cpcloud: Yeah, I think that makes sense. Let's do it in a follow up.

kszucs (approved):

PR #3947, opened 2022-05-20
refactor(ir): deprecate ibis.prevent_rewrite

kszucs:
Doesn't seem to be used anywhere else than in an ibis-bigquery test case: https://github.com/ibis-project/ibis-bigquery/blob/9dc580800d607b809433bb2a3f2da2ba43b2f679/tests/system/test_client.py#L643

cc @tswast

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -1098,6 +1098,7 @@ class Table(Expr):
             suffixes=suffixes,
         )
 
+    @util.deprecated(version="4.0", instead="")
     def prevent_rewrite(self, client=None) -> Table:
         """Prevent optimization from happening below this expression.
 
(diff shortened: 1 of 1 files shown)


ibis/expr/types/relations.py
```diff
@@ -1098,6 +1098,7 @@ def cross_join(
             suffixes=suffixes,
         )
 
+    @util.deprecated(version="4.0", instead="")
     def prevent_rewrite(self, client=None) -> Table:
```
cpcloud, line 1102:
```suggestion
    def prevent_rewrite(self, client=None) -> Table:  # pragma: no cover
```

cpcloud (approved):
LGTM. For future reference a rewrite can usually be prevented by using an empty projection, like `t.projection([])`.

PR #3948, opened 2022-05-20
refactor(ir): deprecate unused ibis.expr.rules.promoter

kszucs:


ibis/expr/rules.py
```diff
@@ -225,9 +225,10 @@ def client(arg, **kwargs):
 
 
 # ---------------------------------------------------------------------
-# Ouput type promoter functions
+# Ouput type functions
 
 
+@util.deprecated(version="4.0", instead="")
 def promoter(fn):
```
cpcloud, line 232:
```suggestion
def promoter(fn):  # pragma: no cover
```

cpcloud (approved):

PR #3949, opened 2022-05-20
refactor(ir): remove ibis.expr.rules.is_computable_input rule

kszucs:


icexelloss:
I don't think so, we use
https://github.com/ibis-project/ibis/blob/master/ibis/backends/pandas/core.py#L150

But not the validator

cpcloud:
@kszucs Can we deprecate and remove in 4?

cpcloud (approved):
Thanks for doing this!

ibis/expr/rules.py
```diff
@@ -436,6 +436,7 @@ def python_literal(value, arg, **kwargs):
 
 
 @validator
+@util.deprecated(version="4.0", instead="")
 def is_computable_input(value, **kwargs):
```
cpcloud, line 440:
```suggestion
def is_computable_input(value, **kwargs):  # pragma: no cover
```

ibis/expr/operations/generic.py
```diff
@@ -262,8 +262,6 @@ class Literal(Value):
                     decimal.Decimal,
                 )
             ),
-            # this seems buggy
-            rlz.is_computable_input,
```
cpcloud, line 266: Actually, can we keep this line, and fully remove in 4.0.0?

cpcloud (changes requested):

ibis/expr/operations/generic.py
kszucs: That would raise deprecation warnings I think resulting pytest errors.

cpcloud: We can selectively ignore warnings to prevent that.

cpcloud (approved):

PR #3952, opened 2022-05-20
refactor(ir): slightly simplify ScalarAggregate's implementation

kszucs:
@cpcloud the new `_rewrite_filter` essentially does the same as `reduction_to_aggregation` which dispatches the logic to the `ScalarAggregate` class doing an almost identical `traverse and reconstruct expression` replacement as `_rewrite_filter`.

Could we generalize (even before using matchpy) this substitution logic to shortly represent the following?

```py
is_scalar_reduction =>
    has_multiple_bases => 
        cross_join_tables().aggregate()
    else => 
        table.aggregate()
```

Actually there is the substitutor class already, could we use that here?


cpcloud (approved):
Great work and nice find!

PR #3983, opened 2022-05-24
refactor(ir): split `Any`/`NotAny` to reduction and `Exists`/`NotExists` subquery operators

### LINKED ISSUE #3925: refactor(operations): split Any into reduction and non-reduction operation
`Any` is currently coupled into a `bool_or`-style reduction and an `EXISTS`-style column operation. We should split these into two different operations internally while preserving the user-facing API.

cpcloud:
This PR splits `Any` and `NotAny` into their reduction versions with the same names,
and the previously-mashed-together `ExistSubquery` while preserving the `.any()` API and semantics.

This is a backwards compatible expression API change, but the value of `.op()`
is different.

It's slightly more optimistic than that though, because we're only moving the
time at which `ExistSubquery` is constructed to earlier in the expression
construction pipeline. Previously it was happening right after a user called
`.compile()` whereas now the transformation happens any time a user calls `.any()`.

Closes #3925.

### DIFF
--- a/ibis/expr/analysis.py
+++ b/ibis/expr/analysis.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import functools
 import operator
-from typing import AbstractSet, MutableSequence
 
 import toolz
 
@@ -575,7 +574,7 @@ def windowize_function(expr, w=None):
                 walked = x
 
         op = walked.op()
-        if isinstance(op, ops.Analytic) or getattr(op, '_reduction', False):
+        if isinstance(op, (ops.Analytic, ops.Reduction)):
             if w is None:
                 w = window()
             return walked.over(w)
@@ -984,9 +983,7 @@ def is_ancestor(parent, child):
 
 def is_analytic(expr):
     def predicate(expr):
-        if isinstance(
-            expr.op(), (ops.Reduction, ops.Analytic, ops.logical._AnyBase)
-        ):
+        if isinstance(expr.op(), (ops.Reduction, ops.Analytic)):
             return lin.halt, True
         else:
             return lin.proceed, None
@@ -1024,7 +1021,7 @@ def is_reduction(expr):
     """
 
     def predicate(expr):
-        if getattr(expr.op(), '_reduction', False):
+        if isinstance(expr.op(), ops.Reduction):
             return lin.halt, True
         elif isinstance(expr.op(), ops.TableNode):
             # don't go below any table nodes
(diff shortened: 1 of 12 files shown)


ibis/expr/analysis.py
```diff
@@ -1039,75 +1037,71 @@ def is_scalar_reduction(expr):
...
+            self.tables.add(table)
 
-    def _visit_table(self, expr: ir.Expr) -> None:
-        if (
-            isinstance(expr, ir.Table)
-            and (base_table := self._find_blocking_table(expr)) is not None
-            and base_table.op() not in self.query_roots
-        ):
-            self.foreign_table = expr
+    def _find_blocking_table(self, node: ops.TableNode) -> ops.TableNode:
```
kszucs, line 1100: It seems to return with a `bool`.

Could you replace it with a new analysis function instead:

```python
def has_blocking_table(expr):
    def predicate(expr):
        op = expr.op()
        if op.blocks():
            return lin.halt, True
        else:
            return lin.proceed, None
        
    return any(lin.traverse(predicate, expr))
```

?

cpcloud: For example, `util.to_op_graph` :)

ibis/expr/operations/logical.py
```diff
@@ -164,37 +168,130 @@ class Where(Value):
...
+    @abc.abstractmethod
+    def _negated_class(self) -> type:  # pragma: no cover
+        ...
+
+    def negate(self):
+        return self._negated_class(*self.args)
+
+
+@public
+class Any(Reduction, Negatable):
```
kszucs, line 182: Move these classes to `reduction.py` and keep aliases in this file using `public()`?

ibis/expr/analysis.py
cpcloud: Yep, this can be replaced with a simple DAG iteration.

cpcloud: Done.

ibis/expr/analysis.py
```diff
@@ -1039,75 +1037,65 @@ def is_scalar_reduction(expr):
...
+        assert isinstance(table, ir.Table)
 
-    def _visit_table(self, expr: ir.Expr) -> None:
-        if (
-            isinstance(expr, ir.Table)
-            and (base_table := self._find_blocking_table(expr)) is not None
-            and base_table.op() not in self.query_roots
-        ):
-            self.foreign_table = expr
+        if any(op.blocks() for op in util.to_op_dag(table)):
```
kszucs, line 1101: `to_op_dag()` is equivalent with:

```python
def has_blocking_table(expr):
    def predicate(expr):
        return lin.proceed, expr.op()
           
    return lin.traverse(predicate, expr)
```

At some point we should unify these two utility functions.

cpcloud: Indeed. Let's do that in a follow-up.

ibis/expr/operations/logical.py
```diff
@@ -164,37 +168,130 @@ class Where(Value):
     output_shape = rlz.shape_like("bool_expr")
 
 
-class _AnyBase(Value):
-    # Depending on the kind of input boolean array, the result might either be
-    # array-like (an existence-type predicate) or scalar (a reduction)
+class Negatable:
```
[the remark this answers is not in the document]
cpcloud: I don't think it'll be fewer lines, since then I have to implement `negate` for all the child classes.

ibis/expr/operations/logical.py
cpcloud: Done.

ibis/expr/analysis.py
```diff
@@ -1039,75 +1037,69 @@ def is_scalar_reduction(expr):
...
         )
-        self.foreign_table: ir.Table | None = None
-        self.predicates: MutableSequence[ir.BooleanValue] = []
+    return op.to_expr()
 
-    def get_result(self) -> ir.BooleanColumn:
-        self._visit(self.expr)
 
-        op = self.expr.op()
+class _AnyToExistsTransform:
```
kszucs, line 1065: How about removing this class entirely?

https://github.com/ibis-project/ibis/commit/38738d488cc0c66b4a4444a42ac567f33e7658ef

ibis/expr/operations/generic.py
```diff
@@ -477,3 +478,13 @@ def root_tables(self):
     def output_dtype(self):
         exprs = [*self.results, self.default]
         return rlz.highest_precedence_dtype(exprs)
+
+
+class _Negatable(metaclass=abc.ABCMeta):
+    @property
+    @abc.abstractmethod
+    def _negated_class(self) -> type:  # pragma: no cover
```
kszucs, line 486: I'd prefer implementing the `negate()` abstractmethod directly in childs to prevent confusion of `_negated_class`  classprop with the argument and return type annotation.

cpcloud: Cool, done.

ibis/expr/analysis.py
cpcloud: Rolled this in. Amazing. 🙇🏻

ibis/expr/operations/logical.py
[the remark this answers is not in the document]
cpcloud: Resolved elsewhere.

ibis/expr/analysis.py
```diff
@@ -1039,75 +1036,41 @@ def is_scalar_reduction(expr):
...
-                result
-                for arg in node.flat_args()
-                if (
-                    isinstance(arg, ir.Expr)
-                    and (result := self._find_blocking_table(arg)) is not None
-                )
-            ),
-            None,
+    ntables = len(tables)
+    if (ntables := len(tables)) == 1:
```
kszucs, line 1065: Use `analysis.has_multiple_bases()`?

ibis/expr/operations/relations.py
```diff
@@ -748,24 +747,6 @@ def blocks(self):
         return True
 
 
-@public
-class ExistsSubquery(Value):
```
kszucs, line 752: Could we keep these here to keep API compat?

ibis/expr/analysis.py
cpcloud: > Use `analysis.has_multiple_bases()`?

Doesn't that mean I would traverse the expr twice? I've already got the table count, so there's no point in doing that extra computation.

kszucs (approved):

kszucs (approved):

PR #3988, opened 2022-05-25
test: use pytest fixtures to initialize and populate the backends once per run

### LINKED ISSUE #2952: test: use pytest fixtures to initialize and populate the backends
I propose to port the click commands from `datamgr.py` and `impalamgr` to pytest fixtures (with `scope="session"` to execute once per pytest invocation).

The python testing suite is dependent on the files under the `ci/` directory. Examples:

`PYTEST_BACKEND=pandas pytest ibis/backends` or `pytest --enable-pandas ibis/backends` should:
1. Check that the testing data exists under the target directory
2. If not, download it otherwise skip
3. Execute the tests with pandas marker enabled

`PYTEST_BACKEND=clickhouse pytest ibis/backends` or `pytest --enable-clickhouse ibis/backends` should:
1. Check that the testing data exists under the target directory
2. If not, download it otherwise skip
3. Check that the required tables exist in clickhouse
4. If not, populate the tables otherwise skip
3. Execute the tests with clickhouse marker enabled

We can add pytest flags to `--force-download`/ `--skip-download` and `--force-populate`/`--skip-populate` the requested backends. 

Even though the first run would take more time since it would implicitly download the data (the second run would be instant), new contributors could use the a single `pytest` command to execute the testing suite. We can add additional logging / progress bar to inform the developer about the data[base] initialization.

anjakefala:
This PR adds a `load_data` to the `TestConf` for all of the current backends in `ibis/backends/{backend}/tests/conftest.py`. PR also removes references to `ci/datamgr.py load` from the `docs/contribute/05_backend_tests.md`.

If the database already exists, the database will be dropped, and recreated.
    
Expected that every backend in the future will have a `load_data` function in its `TestConf`.

This PR uses [filelock](https://py-filelock.readthedocs.io/en/latest/) to ensure that the databases are only instantiated once per test run. It handles the case for if `pytest-xdist` is used to create multiple `pytest` threads. With `pytest-xdist` session fixtures are invoked once per thread.

I attempted to follow the instructions [here](https://github.com/ibis-project/ibis/blob/master/docs/contribute/05_maintainers_guide.md#adding-or-changing-dependencies) to add `filelock` as a dependency, but feedback is welcome if there is something specific I should to ensure it is:
* explicitly marked as a testing-dependency only
* added to the conda.lock
    
 As of this code, `ci/datamgr.py load {backend}` will still work.
    
Closes #2952

### DIFF
--- a/docs/contribute/04_backend_tests.md
+++ b/docs/contribute/04_backend_tests.md
@@ -72,23 +72,16 @@ success
 
 Congrats, you now have a PostgreSQL server running and are ready to run tests!
 
-#### Load Data
+#### Download Data
 
-The backend needs to be populated with test data:
+The backend needs to be populated with test data.
+The data will be loaded automatically, when the test is run, but it needs to be downloaded first.
 
-1.  Download the data
+To download the data
 
-    ```sh
-    python ci/datamgr.py download
-    ```
-
-2.  In the original terminal, run
-
-    ```sh
-    python ci/datamgr.py load postgres
-    ```
-
-    You should see a bit of logging, and the command should complete shortly thereafter.
+```sh
+python ci/datamgr.py download
+```
 
 #### Run the test suite
 
--- a/ibis/backends/clickhouse/tests/conftest.py
+++ b/ibis/backends/clickhouse/tests/conftest.py
@@ -6,6 +6,7 @@ import pytest
 
 import ibis
 import ibis.expr.types as ir
+from ibis.backends.conftest import TEST_TABLES, read_tables
 from ibis.backends.tests.base import (
     BackendTest,
     RoundHalfToEven,
@@ -26,6 +27,49 @@ class TestConf(UnorderedComparator, BackendTest, RoundHalfToEven):
     supported_to_timestamp_units = {'s'}
     supports_floating_modulus = False
     bool_is_int = True
+
+    @staticmethod
+    def load_data(data_dir: Path, script_dir: Path, **kwargs) -> None:
+        """Load testdata into a clikhouse backend.
+
+        Parameters
+        ----------
+        data_dir : Path
+            Location of testdata
+        script_dir : Path
+            Location of scripts defining schemas
+        """
+        import clickhouse_driver
+
+        host = kwargs.get("host", CLICKHOUSE_HOST)
+        port = kwargs.get("port", CLICKHOUSE_PORT)
+        user = kwargs.get("user", CLICKHOUSE_USER)
+        password = kwargs.get("password", CLICKHOUSE_PASS)
+        database = kwargs.get("database", IBIS_TEST_CLICKHOUSE_DB)
+
+        tables = list(TEST_TABLES.keys())
+        client = clickhouse_driver.Client(
+            host=host,
+            port=port,
+            user=user,
+            password=password,
+        )
+
+        client.execute(f"DROP DATABASE IF EXISTS {database}")
+        client.execute(f"CREATE DATABASE {database}")
+        client.execute(f"USE {database}")
+
+        with open(script_dir / 'schema' / 'clickhouse.sql') as schema:
+            for stmt in filter(None, map(str.strip, schema.read().split(";"))):
+                client.execute(stmt)
+
+        for table, df in read_tables(tables, data_dir):
+            query = f"INSERT INTO {table} VALUES"
+            client.insert_dataframe(
+                query,
+                df.to_pandas(),
+                settings={"use_numpy": True},
+            )
 
     @staticmethod
     def connect(data_directory: Path):
--- a/ibis/backends/conftest.py
+++ b/ibis/backends/conftest.py
@@ -5,10 +5,15 @@ import os
 import platform
 from functools import lru_cache
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING, Any, Iterable, Iterator, TextIO
 
 import _pytest
 import pandas as pd
+import sqlalchemy as sa
+
+if TYPE_CHECKING:
+    import pyarrow as pa
+
 import pytest
 
 import ibis
@@ -20,6 +25,91 @@ except ImportError:
     import importlib_metadata
 
 
+TEST_TABLES = {
+    "functional_alltypes": ibis.schema(
+        {
+            "index": "int64",
+            "Unnamed: 0": "int64",
+            "id": "int32",
+            "bool_col": "boolean",
+            "tinyint_col": "int8",
+            "smallint_col": "int16",
+            "int_col": "int32",
+            "bigint_col": "int64",
+            "float_col": "float32",
+            "double_col": "float64",
+            "date_string_col": "string",
+            "string_col": "string",
+            "timestamp_col": "timestamp",
+            "year": "int32",
+            "month": "int32",
+        }
+    ),
+    "diamonds": ibis.schema(
+        {
+            "carat": "float64",
+            "cut": "string",
+            "color": "string",
+            "clarity": "string",
+            "depth": "float64",
+            "table": "float64",
+            "price": "int64",
+            "x": "float64",
+            "y": "float64",
+            "z": "float64",
+        }
+    ),
+    "batting": ibis.schema(
+        {
+            "playerID": "string",
+            "yearID": "int64",
+            "stint": "int64",
+            "teamID": "string",
+            "lgID": "string",
+            "G": "int64",
+            "AB": "int64",
+            "R": "int64",
+            "H": "int64",
+            "X2B": "int64",
+            "X3B": "int64",
+            "HR": "int64",
+            "RBI": "int64",
+            "SB": "int64",
+            "CS": "int64",
+            "BB": "int64",
+            "SO": "int64",
+            "IBB": "int64",
+            "HBP": "int64",
+            "SH": "int64",
+            "SF": "int64",
+            "GIDP": "int64",
+        }
+    ),
+    "awards_players": ibis.schema(
+        {
+            "playerID": "string",
+            "awardID": "string",
+            "yearID": "int64",
+            "lgID": "string",
+            "tie": "string",
+            "notes": "string",
+        }
+    ),
+}
+
+
+@pytest.fixture(scope='session')
+def script_directory() -> Path:
+    """Return the test script directory.
+
+    Returns
+    -------
+    Path
+        Test script directory
+    """
+    return Path(__file__).absolute().parents[2] / "ci"
+
+
 @pytest.fixture(scope='session')
 def data_directory() -> Path:
     """Return the test data directory.
@@ -29,7 +119,7 @@ def data_directory() -> Path:
     Path
         Test data directory
     """
-    root = Path(__file__).absolute().parent.parent.parent
+    root = Path(__file__).absolute().parents[2]
 
     return Path(
         os.environ.get(
@@ -37,6 +127,98 @@ def data_directory() -> Path:
             root / "ci" / "ibis-testing-data",
         )
     )
+
+
+def recreate_database(
+    url: sa.engine.url.URL,
+    database: str,
+    **kwargs: Any,
+) -> None:
+    """Drop the {database} at {url}, if it exists.
+
+    Create a new, blank database with the same name.
+
+    Parameters
+    ----------
+    url : url.sa.engine.url.URL
+        Connection url to the database
+    database : str
+        Name of the database to be dropped.
+    """
+    engine = sa.create_engine(url, **kwargs)
+
+    if url.database is not None:
+        with engine.connect() as conn:
+            conn.execute(f'DROP DATABASE IF EXISTS {database}')
+            conn.execute(f'CREATE DATABASE {database}')
+
+
+def init_database(
+    url: sa.engine.url.URL,
+    database: str,
+    schema: TextIO | None = None,
+    recreate: bool = True,
+    **kwargs: Any,
+) -> sa.engine.Engine:
+    """Initialise {database} at {url} with {schema}.
+
+    If {recreate}, drop the {database} at {url}, if it exists.
+
+    Parameters
+    ----------
+    url : url.sa.engine.url.URL
+        Connection url to the database
+    database : str
+        Name of the database to be dropped
+    schema : TextIO
+        File object containing schema to use
+    recreate : bool
+        If true, drop the database if it exists
+
+    Returns
+    -------
+    sa.engine.Engine for the database created
+    """
+    if recreate:
+        recreate_database(url, database, **kwargs)
+
+    try:
+        url.database = database
+    except AttributeError:
+        url = url.set(database=database)
+
+    engine = sa.create_engine(url, **kwargs)
+
+    if schema:
+        with engine.connect() as conn:
+            for stmt in filter(None, map(str.strip, schema.read().split(';'))):
+                conn.execute(stmt)
+
+    return engine
+
+
+def read_tables(
+    names: Iterable[str],
+    data_dir: Path,
+) -> Iterator[tuple[str, pa.Table]]:
+    """For each csv {names} in {data_dir} return a pyarrow.Table"""
+
+    import pyarrow.csv as pac
+
+    import ibis.backends.pyarrow.datatypes as pa_dt
+
+    for name in names:
+        schema = TEST_TABLES[name]
+        convert_options = pac.ConvertOptions(
+            column_types={
+                name: pa_dt.to_pyarrow_type(type)
+                for name, type in schema.items()
+            }
+        )
+        yield name, pac.read_csv(
+            data_dir / f'{name}.csv',
+            convert_options=convert_options,
+        )
 
 
 def _random_identifier(suffix: str) -> str:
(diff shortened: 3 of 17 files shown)


ibis/backends/conftest.py
```diff
@@ -258,10 +443,464 @@ def pytest_runtest_call(item):
...
+    "duckdb": load_duckdb,
+}
+
+
 @pytest.fixture(params=_get_backends_to_test(), scope='session')
-def backend(request, data_directory):
-    """Return an instance of BackendTest."""
+def backend(request, data_directory, script_directory):
+    """Return an instance of BackendTest, loaded with data."""
+    # TODO: Will request.param ever be a list of backends?
```
anjakefala, line 896: TODO should be removed before merge: add an assert or handle the case for list of backends.

docs/contribute/04_backend_tests.md
```diff
@@ -74,22 +74,15 @@ Congrats, you now have a PostgreSQL server running and are ready to run tests!
 
 #### Load Data
 
-The backend needs to be populated with test data:
+The backend needs to be populated with test data.
+The data will be loaded automatically, when the test is run, but it needs to be downloaded first.
 
 1.  Download the data
```
saulpw, line 80: This is now a 1-item list so it can be de-listed.

.github/workflows/ibis-backends.yml
```diff
@@ -96,9 +96,6 @@ jobs:
       - name: download backend data
         run: just download-data
 
-      - name: install backend data
-        run: poetry run python ci/datamgr.py load ${{ matrix.backend.name }}
-
       - name: run non-pyspark tests
         if: ${{ matrix.backend.name != 'pyspark' }}
         run: just ci-check -m ${{ matrix.backend.name }} --numprocesses auto --dist=loadgroup
```
anjakefala, line 101: What is `just ci-check`?

cpcloud: If you run `just` (it defaults to running `just --list`) inside the ibis repo you can see descriptions of each command:

```
❯ just
just --list
Available recipes:
    bench *args               # run the benchmark suite
    check *args               # run all non-backend tests; additional arguments are forwarded to pytest
    checklinks *args          # check for invalid links in a locally built version of the docs
    ci-check *args            # run pytest for ci; additional arguments are forwarded to pytest
    clean                     # clean untracked files
    default                   # list justfile recipes
    down                      # stop and remove containers; clean up networks and volumes
    download-data owner="ibis-project" repo="testing-data" rev="master" # download testing data
    fmt                       # format code
    lint                      # lint code
    list-backends             # show all backends
    load-data *backends="all" # load data into running backends; requires a running backend
    test +backends            # run the test suite for one or more backends
    typecheck                 # type check code using mypy
    up *backends              # start backends using docker compose and load data; no arguments starts all backends
```

cpcloud:
@anjakefala Thanks for this, it's great work.

The reason there are CI failures is because some of the backends have tests that run against backend data, but aren't "backend" tests in that they do not run across all the backends. The non-backend-backend tests and the proper backend tests share the same data. Without loading that data at some point, the tests will fail to find the things they need to run the test.

In an ideal world we only have proper backend tests for things that ultimately call `execute`, but the backend test suite didn't exist before the need to check that ibis expressions returned the right results when executing 😅.

As for how to proceed, I think if possible we can leave the `python ci/datamgr.py` line in CI and work towards moving the backend-specific backend tests to the larger suite. This will take time, since we'll have to go test by test and remove duplicated tests as well as `notimpl`/`notyet`/`never`-mark the ones that don't work on the other backends. It's kind of a grind 😞.

That said, I'm definitely willing to take on a sizable chunk of that work, which is mostly in the postgres, sqlite and clickhouse backends.

cpcloud:
Is it possible to use the code in `ibis/backends/conftest.py` in `ci/datamgr.py` to prevent duplication?

cpcloud:
Regarding `git clone`, we're already doing that with `just download-data` so I think we can probably call that done :)

anjakefala:
> As for how to proceed, I think if possible we can leave the python ci/datamgr.py line in CI and work towards moving the backend-specific backend tests to the larger suite. This will take time, since we'll have to go test by test and remove duplicated tests as well as notimpl/notyet/never-mark the ones that don't work on the other backends. It's kind of a grind :disappointed:.

> That said, I'm definitely willing to take on a sizable chunk of that work, which is mostly in the postgres, sqlite and clickhouse backends.

Ohhhh, that explains it! I am always pretty happy to help with grindwork. It does not feel like a grind to me.

:+1: if you want me to open an issue for this?

cpcloud:
Opening an issue sounds good.

The places to start looking are `ibis/backends/{sqlite,postgres,clickhouse}/tests/test_functions.py`. Much of the duplicated backend testing lies there.

I suspect that after we're done deduplicating that CI times will improve by a decent amount.

cpcloud:
I will put up a PR that dedups a test as an example 😄

ibis/backends/conftest.py
```diff
@@ -258,10 +445,556 @@ def pytest_runtest_call(item):
...
+    data_dir : Path
+        Location of testdata
+    script_dir : Path
+        Location of scripts defining schemas
+    """
+    PG_USER = (
+        os.environ.get(
+            'IBIS_TEST_POSTGRES_USER', os.environ.get('PGUSER', 'postgres')
+        )
+        if not kwargs.get("username", None)
```
anjakefala, line 462: These `kwarg` checks were added to support users using non-default values for `ci/datamgr.py`. It does add a lot of boiler plate. 

Do we want to continue supporting those `kwargs`? Boiler plate could also be decreased by me assuming that the `kwargs` for all of the values will be set, if there are `kwargs` being set. Then, I could just do an `if kwargs` check, and set variables appropriately.

ibis/backends/conftest.py
```diff
@@ -20,6 +30,91 @@
...
+@pytest.fixture(scope='session')
+def script_directory() -> Path:
+    """Return the test script directory.
+
+    Returns
+    -------
+    Path
+        Test script directory
+    """
+    return Path(__file__).absolute().parent.parent.parent / "ci"
```
kszucs, line 115: Consider to use [.parents](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parents)?

kszucs:
The `ibis/backends/conftest.py` file is already rather heavy. Would it be possible to reorganize the loaders as staticmethods for the respective `TestConf` classes? Something like 

```
class ClickhouseTestConf(...):
    @staticmethod
    def load_data(data_dir: Path, script_dir: Path, **kwargs):
        ...
```

That could help with the credential deduplication as well as encapsulating backend specific logic in a single file.

anjakefala:
@kszucs I agree with this! I am trying to understand why the build is failing before refactoring things. Then, I will refactor with your suggested changes.

```
pytest --durations=25 -ra --junitxml=junit.xml --cov=ibis --cov-report=xml:coverage.xml -m postgres --numprocesses auto --dist=loadgroup
```

Errors. `pytest -s -m postgres` passes. This is why there are failures in the CI, while the build was passing for me locally.

The flavour of errors:

```
>       cursor.execute(statement, parameters)                                                                                                                  
E       sqlalchemy.exc.IntegrityError: (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "pg_class_relname_nsp_index"           
E       DETAIL:  Key (relname, relnamespace)=(test_sequence, 2200) already exists.                                                                             
E                                                                                                                                                              
E       [SQL: CREATE SEQUENCE IF NOT EXISTS test_sequence]                                                                                                     
E       (Background on this error at: https://sqlalche.me/e/14/gkpj)    
```

anjakefala:
@cpcloud Yes, exactly. The longer `pytest` fails locally, but not the second one. The longer one is the one being run by the CI tests. 

My current guess from the error messages is that the tests are trying to create the same tables multiple times.

anjakefala:
I pushed an experiment that uses [FileLock](https://py-filelock.readthedocs.io/en/latest/index.html) to handle multiple concurrent pytest threads. Ideally, it ensures that the database is re-loaded once per run, with other threads waiting until the loading completes. FileLock supports both Posix and Windows.

If this does look good, things that will need to be done:
- [ ] FileLock will needed to be added to the testing dependencies for poetry and conda. 
- [ ] I should ensure that the teardown works if the `pytest` is run multiple times, and crashes mid-way.

anjakefala:
I added this to my PR description:

This PR uses [filelock](https://py-filelock.readthedocs.io/en/latest/) to ensure that the databases are only instantiated once per test session. It handles the case for if `pytest-xdist` is used to create multiple `pytest` threads. I attempted to follow the instructions [here](https://github.com/ibis-project/ibis/blob/master/docs/contribute/05_maintainers_guide.md#adding-or-changing-dependencies) to add `filelock` as a dependency, but feedback is welcome if there is something specific I should to ensure it is:
* explicitly marked as a testing-dependency only
* added to the conda.lock

anjakefala:
I tried @kszucs suggestion experimentally, and I did get it to work for both `pytest` and `ci/datamgr.py`. It feels like the approach with the least duplication, so I am going to be implementing it. I will then re-open for review.

ibis/backends/clickhouse/tests/conftest.py
```diff
@@ -27,6 +28,72 @@ class TestConf(UnorderedComparator, BackendTest, RoundHalfToEven):
...
+        Parameters
+        ----------
+        data_dir : Path
+            Location of testdata
+        script_dir : Path
+            Location of scripts defining schemas
+        """
+        import clickhouse_driver
+
+        host = (
```
kszucs, line 44: `kwargs.get("host", CLICKHOUSE_HOST)` should do the same, or you could explicitly define these as default keyword arguments.

ibis/backends/conftest.py
```diff
@@ -259,9 +441,29 @@ def pytest_runtest_call(item):
...
+    if worker_id == "master":  # single-threaded pytest
+        cls.load_data(data_directory, script_directory)
+        return cls(data_directory)
+
+    # multi-threaded pytest
+
+    # get the temp directory shared by all workers
+    root_tmp_dir = tmp_path_factory.getbasetemp().parent
+
+    fn = root_tmp_dir / "lockfile"
```
kszucs, line 462:
```suggestion
    fn = root_tmp_dir / "lockfile.lock"
```

ibis/backends/datafusion/tests/conftest.py
```diff
@@ -16,6 +16,10 @@ class TestConf(BackendTest, RoundAwayFromZero):
     # returned_timestamp_unit = 'ns'
     bool_is_int = True
 
+    @staticmethod
+    def load_data(data_dir, script_dir, **kwargs):
```
kszucs, line 20: I suggest to add this default implementation to the base `BackendTest` class.

ibis/backends/duckdb/tests/conftest.py
```diff
@@ -2,13 +2,54 @@
...
+from ibis.backends.conftest import TEST_TABLES
 from ibis.backends.tests.base import BackendTest, RoundAwayFromZero
 
 
 class TestConf(BackendTest, RoundAwayFromZero):
     def __init__(self, data_directory: Path) -> None:
         self.connection = self.connect(data_directory)
 
+    @staticmethod
+    def load_data(data_dir, script_dir, **kwargs):
```
kszucs, line 14:
```suggestion
    def load_data(data_dir, script_dir, database="ibis_testing", **kwargs):
```

ibis/backends/duckdb/tests/conftest.py
```diff
@@ -2,13 +2,54 @@
...
+            if not kwargs.get("database", None)
+            else kwargs["database"]
+        )
+
+        schema = (
+            open(script_dir / 'schema' / 'duckdb.sql')
+            if not kwargs.get("schema", None)
+            else kwargs["schema"]
+        )
+        tables = (
```
kszucs, line 36: If we don't define table schemas anywhere else then baked in `TEST_TABLES` would be enough here as well.

ibis/backends/mysql/tests/conftest.py
```diff
@@ -48,6 +50,67 @@ def __init__(self, data_directory: Path) -> None:
...
+            if not kwargs.get("database", None)
+            else kwargs["database"]
+        )
+
+        schema = (
+            open(script_dir / 'schema' / 'mysql.sql')
+            if not kwargs.get("schema", None)
+            else kwargs["schema"]
+        )
+        tables = (
```
kszucs, line 87: tables = list(TEST_TABLES.keys())

ibis/backends/mysql/tests/conftest.py
```diff
@@ -48,6 +50,67 @@ def __init__(self, data_directory: Path) -> None:
...
+                f"mysql+pymysql://{user}:{password}@{host}:{port}?local_infile=1",  # noqa: E501
+            ),
+            database=database,
+            schema=schema,
+            isolation_level="AUTOCOMMIT",
+        )
+        with engine.connect() as con:
+            for table in tables:
+                con.execute(
+                    f"""\
```
kszucs, line 104: I like to use `textwrap.dedent()`  to preserve the tabulation of the python source code.

ibis/backends/postgres/tests/conftest.py
```diff
@@ -45,6 +47,70 @@ class TestConf(BackendTest, RoundHalfToEven):
...
+
+        Parameters
+        ----------
+        data_dir : Path
+            Location of testdata
+        script_dir : Path
+            Location of scripts defining schemas
+        """
+        user = (
+            PG_USER if not kwargs.get("username", None) else kwargs["username"]
```
kszucs, line 62: Use default keyword arguments instead?

kszucs (changes requested):
This is really nice, just a couple of nits.

ibis/backends/conftest.py
anjakefala: I think what I have is correct!

There is `lockfile` and `lockfile.lock`. If `lockfile.lock` does not exist and `lockfile` does not, create `lockfile.lock`, load data, and then create `lockfile`.
If `lockfile.lock` exists, wait until it does not exist, and then continue.
If `lockfile.lock` does not exist and `lockfile` does, then continue.

kszucs: Ahh right, so `lockfile` itself indicates whether the backend has been initialized or not. 

Is this `lockfile` shared for all backends? Do we need separate lockfiles perhaps?

ibis/backends/conftest.py
```diff
@@ -259,9 +441,29 @@ def pytest_runtest_call(item):
...
+    with FileLock(str(fn) + ".lock"):
+        if not fn.is_file():
+            cls.load_data(data_directory, script_directory)
```
kszucs, line 465: Is the lockfile unique across pytest invocations meaning that we load data each time we execute pytest?

anjakefala: Oh, good point! The lockfile is unique, and so as long as it exists, we will not load data each time we execute.

Would the preferred behaviour be to:
* keep that behaviour
* clean up the lockfile in teardown

ibis/backends/conftest.py
anjakefala: > Is this lockfile shared for all backends? Do we need separate lockfiles perhaps?

This is a really good point! I think we do.

ibis/backends/clickhouse/tests/conftest.py
```diff
@@ -27,6 +28,50 @@ class TestConf(UnorderedComparator, BackendTest, RoundHalfToEven):
...
+            port=port,
+            user=user,
+            password=password,
+        )
+
+        client.execute(f"DROP DATABASE IF EXISTS {database}")
+        client.execute(f"CREATE DATABASE {database}")
+        client.execute(f"USE {database}")
+
+        for stmt in filter(None, map(str.strip, schema.read().split(";"))):
```
cpcloud, line 64:
```suggestion
        with open(script_dir / 'schema' / 'clickhouse.sql') as schema:
	          tables = list(TEST_TABLES.keys())
	  
	          client = clickhouse_driver.Client(
	              host=host,
	              port=port,
	              user=user,
	              password=password,
	          )
	  
	          client.execute(f"DROP DATABASE IF EXISTS {database}")
	          client.execute(f"CREATE DATABASE {database}")
	          client.execute(f"USE {database}")
	  
	          for stmt in filter(None, map(str.strip, schema.read().split(";"))):
```

cpcloud (changes requested):
Few changes around file handles needed, but after that this is good to go!

kszucs (approved):
Just tested it out locally and everything is working as expected, so it looks good to me.

Thanks @anjakefala!

@cpcloud do you want to have a final look?

ibis/backends/conftest.py
```diff
@@ -39,6 +129,98 @@ def data_directory() -> Path:
...
+        Connection url to the database
+    database : str
+        Name of the database to be dropped.
+    """
+    engine = sa.create_engine(url, **kwargs)
+
+    if url.database is not None:
+        with engine.connect() as conn:
+            conn.execute(f'DROP DATABASE IF EXISTS {database}')
+            conn.execute(f'CREATE DATABASE {database}')
```
cpcloud, line 153: Looks like this isn't covered, but we can explore (and hopefully delete!) in a follow up.

ibis/backends/duckdb/tests/conftest.py
cpcloud: This is fine for now. I will address in a follow up.

ibis/backends/mysql/tests/conftest.py
```diff
@@ -48,6 +50,67 @@ def __init__(self, data_directory: Path) -> None:
             # mysql supports window operations after version 8
             self.__class__.supports_window_operations = True
 
+    @staticmethod
+    def load_data(data_dir: Path, script_dir: Path, **kwargs) -> None:
```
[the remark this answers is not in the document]
cpcloud: Similarly, I will address later.

cpcloud (approved):
One of my favorite PRs :)

Excellent work.

PR #4000, opened 2022-05-27
feat: implement quantile in more backends

cpcloud:
This PR implements the `Quantile` and `MultiQuantile` (where supported by the
backend) for the ClickHouse, DuckDB and PostgreSQL backends.

### DIFF
--- a/ibis/expr/analysis.py
+++ b/ibis/expr/analysis.py
@@ -201,10 +201,7 @@ def substitute(fn, expr):
     """
     Substitute expressions with other expressions.
     """
-    try:
-        node = expr.op()
-    except AttributeError:
-        return expr
+    node = expr.op()
 
     result = fn(node)
     if result is lin.halt:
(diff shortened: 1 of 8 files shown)


cpcloud:
I'm going to remove the bits that make this operation filterable into a separate PR since that's a breaking change.

ibis/expr/analysis.py
```diff
@@ -201,7 +201,10 @@ def substitute(fn, expr):
     """
     Substitute expressions with other expressions.
     """
-    node = expr.op()
+    try:
```
kszucs, line 204: Is this required?

kszucs (approved):
Just a nit otherwise LGTM

ibis/expr/analysis.py
cpcloud: Only if we merge this before #4015. I'll remove it so this PR fails until we merge that one.

PR #4015, opened 2022-05-30
refactor(analysis): simplify Substitutor

kszucs:


ibis/expr/analysis.py
```diff
@@ -44,60 +44,41 @@ def sub_for(expr, substitutions):
...
+    for arg in node.args:
+        if isinstance(arg, tuple):
+            arg = tuple(_substitute(item, mapping) for item in arg)
+        elif isinstance(arg, ir.Expr):
+            arg = _substitute(arg, mapping)
+        new_args.append(arg)
+
+    try:
+        new_node = type(node)(*new_args)
+    except IbisTypeError:
```
kszucs, line 78: This should really go, but that breaks several test cases.

cpcloud (commented):
This seems like a great use case for the graph utilities in util to avoid recursion. I may try to tinker with that after we merge this one.

kszucs:
Well, we need to keep this (actually wrong) implementation because certain rewrites are tied together with conditions in  `_apply_filter` and `_rewrite_filter`  analysis functions, but this is a tiny step further.

ibis/expr/analysis.py
```diff
@@ -38,61 +38,16 @@ def sub_for(expr, substitutions):
...
-        expr : ibis.expr.types.Expr
-        mapping : Mapping[ibis.expr.operations.Node, ibis.expr.types.Expr]
-
-        Returns
-        -------
-        ibis.expr.types.Expr
-        """
-        node = expr.op()
+
+    def fn(node, mapping={k.op(): v for k, v in substitutions}):
```
kszucs, line 42: It's a bit strange, how about just using fn as a closure and set mapping before the function definition?

ibis/expr/analysis.py
```diff
@@ -217,8 +176,12 @@ def substitute(fn, expr):
             arg = substitute(fn, arg)
         new_args.append(arg)
 
-    new_node = type(node)(*new_args)
-    return new_node.to_expr()
+    try:
+        new_node = node.__class__(*new_args)
```
kszucs, line 180: This should never fail, swallowing the exception here can cause weird behaviours.

ibis/expr/analysis.py
```diff
@@ -198,16 +153,20 @@ def finder(expr):
 
 
 def substitute(fn, expr):
-    """
-    Substitute expressions with other expressions.
-    """
-    node = expr.op()
+    """Substitute expressions with other expressions."""
+    try:
+        node = expr.op()
```
kszucs, line 158: `expr` should always be an expression type, we should investigate otherwise

cpcloud:
```python
            arg = tuple(substitute(fn, expr) for expr in arg)
```
doesn't check `arg`'s type which will lead to a recursive call potentially being not-an-`Expr`. I'll fix.

cpcloud: Done.

ibis/expr/analysis.py
cpcloud: Done.

ibis/expr/analysis.py
```diff
@@ -198,27 +155,37 @@ def finder(expr):
...
-        return result.to_expr()
+        assert isinstance(result, ir.Expr), type(result)
+        return result
 
     new_args = []
     for arg in node.args:
         if isinstance(arg, tuple):
-            arg = tuple(substitute(fn, expr) for expr in arg)
+            arg = tuple(
+                substitute(fn, expr) if isinstance(arg, ir.Expr) else expr
```
kszucs, line 172: If `arg` is a tuple then it won't apply the substitution on those values.

These checks must be moved to the top of the substitute() function.

ibis/expr/analysis.py
cpcloud: I think we have to keep this for now.

ibis/expr/analysis.py
cpcloud: Yes, but unfortunately there's code in pushdown and projection optimization that depends on the exception to indicate that substitution wasn't possible.

kszucs (commented):
+1

kszucs:
@cpcloud I cannot merge this PR

cpcloud (approved):

cpcloud:
@kszucs I approved. If codecov is still nagging us then I'll merge it manually.

PR #4018, opened 2022-05-30
refactor(ir): deprecate rlz.named_literal

kszucs:


cpcloud:
Did this ever have a use case that we tested?

kszucs:
I'm not sure. Perhaps other refactors made it obsolete.

PR #4019, opened 2022-05-30
refactor(ir): remove ExprValidator and FilterValidator from analysis

kszucs:


kszucs:
There is a single failing test case I need to investiage:

```
FAILED ibis/tests/sql/test_select_sql.py::test_fuse_projections - ibis.common.exceptions.RelationError: Selection expressions don't fully originate from dependencies of the table expression.
```

ibis/tests/sql/test_select_sql.py
```diff
@@ -373,6 +373,7 @@ def test_no_aliases_needed():
     assert not context.need_aliases()
 
 
+@pytest.mark.skip
```
kszucs, line 376: Need to fix this before merging

ibis/expr/analysis.py
```diff
@@ -804,75 +736,29 @@ def predicate(expr):
...
 
 
-def fully_originate_from(exprs, parents):
-    def finder(expr):
-        op = expr.op()
+def _find_root_table(expr):
+    op = expr.op()
+
+    if isinstance(op, ops.Selection):
+        return lin.proceed, op.table
```
kszucs, line 743: I'm not sure I can fully reason about this condition, but my guess is that it omits the topmost selection.

ibis/tests/sql/test_select_sql.py
```diff
@@ -1471,3 +1471,20 @@ def test_endswith(endswith):
...
+        lambda x: x.color.lower().rlike('.*ge.*'),
+    ]
+
+    expr = table
+    for pred in predicates:
+        filtered = expr.filter(pred(expr))
+        projected = filtered.projection([expr])
+        expr = projected
+
+    Compiler.to_sql(expr)
```
cpcloud, line 1490: Should we assert something here?

cpcloud (changes requested):

ibis/expr/operations/relations.py
```diff
@@ -646,15 +647,20 @@ class Aggregation(TableNode, sch.HasSchema):
...
+                "Selection expressions don't fully originate from "
+                "dependencies of the table expression."
+            )
 
-        # Validate predicates
-        validator = FilterValidator([table])
-        validator.validate_all(predicates)
+        for predicate in predicates:
+            if not shares_some_roots(predicate, table):
+                raise com.RelationError(
```
cpcloud, line 661: Should we add tests to cover this case?

ibis/expr/operations/relations.py
```diff
@@ -646,15 +647,20 @@ class Aggregation(TableNode, sch.HasSchema):
...
+            raise com.RelationError(
+                "Selection expressions don't fully originate from "
+                "dependencies of the table expression."
+            )
 
-        # Validate predicates
-        validator = FilterValidator([table])
-        validator.validate_all(predicates)
+        for predicate in predicates:
+            if not shares_some_roots(predicate, table):
```
cpcloud, line 660: I don't think it's possible to hit this code path.

To hit it would require a `Selection` to also fail this check which isn't possible because it's checked with `fully_originates_from` or `predicates` aren't involved at every construction site of `Aggregation`.

This can be turned into an `assert`.

ibis/expr/operations/relations.py
```diff
@@ -646,15 +647,20 @@ class Aggregation(TableNode, sch.HasSchema):
...
 
     def __init__(self, table, metrics, by, having, predicates, sort_keys):
-        from ibis.expr.analysis import FilterValidator
+        from ibis.expr.analysis import shares_all_roots, shares_some_roots
 
         # All non-scalar refs originate from the input table
-        all_exprs = metrics + by + having + sort_keys
-        table._assert_valid(all_exprs)
+        if not shares_all_roots(metrics + by + having + sort_keys, table):
+            raise com.RelationError(
```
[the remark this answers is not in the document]
cpcloud: I'll push up a test that covers this.

PR #4025, opened 2022-05-31
feat(api): add toplevel ibis.connect()

saulpw:
Add a toplevel API that can take either a URL or a path string, and connect to it, without having to specify the backend explicitly.  Auto-detect based on URL scheme or path extension.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -686,3 +686,24 @@ def _(
     con = getattr(ibis, backend).connect(**kwargs)
     con.register(f"{extension}://{filename}")
     return con
+
+
+@_connect.register(
+    r"(?P<filename>.+\.(?P<extension>parquet|csv))",
+    priority=8,
+)
+def _(
+    _: str,
+    *,
+    filename: str,
+    extension: str,
+    **kwargs: Any,
+) -> BaseBackend:
+    """Connect to `duckdb` and register a parquet or csv file.
+
+    Examples
+    --------
+    >>> con = ibis.connect("relative/path/to/data.csv")
+    >>> con = ibis.connect("relative/path/to/more/data.parquet")
+    """
+    return _connect(f"duckdb://{filename}", **kwargs)
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
 import abc
+import functools
 import re
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, ClassVar, Iterable, Mapping
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


ibis/backends/base/__init__.py
```diff
@@ -593,3 +595,28 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+        if '' in schemeparts:
+            schemeparts.remove('')
+
+        if schemeparts:
+            return getattr(ibis, schemeparts[0]).connect(r.path)
+
+    ext = path.suffix[1:]
+
+    if ext in {'parquet', 'csv'}:
+        return ibis.datafusion.connect({path.name: path})
```
cpcloud, line 620: We probably want to use DuckDB for both of these since Datafusion operation support is pretty sparse right now. Something like this:

```python
con = ibis.duckdb.connect()  # in memory connection
con.register_csv(...)  # this API doesn't exist yet
# or 
con.register_parquet(...)  # also doesn't exist yet
```

ibis/backends/base/__init__.py
```diff
@@ -593,3 +595,28 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
         raise NotImplementedError(
             f"{cls.name} backend has not implemented `has_operation` API"
         )
+
+
+def connect(url_or_path: str) -> BaseBackend:
+    'Connect to given *url_or_path*. Auto-detect backend from url scheme or path extension.'
+    path = Path(url_or_path)
```
cpcloud, line 602: Can we reuse [`sqlalchemy.engine.make_url`](https://docs.sqlalchemy.org/en/14/core/engines.html?highlight=make_url#sqlalchemy.engine.make_url) here?

The [`sqlalchemy.engine.URL`](https://docs.sqlalchemy.org/en/14/core/engines.html?highlight=make_url#sqlalchemy.engine.URL) object nicely captures most or all of the things we care about and `make_url` handles the parsing includes the driver with `+`.

saulpw: Sure thing, I will check it out!

cpcloud: > It seems like we still have to split the driver out ourselves

I think if you reduce the scope to not include the extension then it becomes:

```
In [18]: import ibis, sqlalchemy as sa

In [19]: url = sa.engine.url.make_url('duckdb:///ci/ibis-testing-data/ibis_testing.ddb')

In [20]: backend = getattr(ibis, url.drivername)

In [21]: con = backend.connect(url.database)
```

cpcloud: What's the purpose of the `+`? Apologies if we discussed this already, I can't remember.

cpcloud: I'm thinking we just start with `backend://` URLs for now, instead of supporting `backend://`, `ibis+backend://`, files with extensions matching the backend name.

ibis/expr/operations/relations.py
```diff
@@ -379,31 +379,6 @@ class Selection(TableNode, sch.HasSchema):
         default=(),
     )
 
-    def __init__(self, table, selections, predicates, sort_keys):
```
cpcloud, line 382: We can't delete this code just yet. This may have been included from #4051.

kszucs (commented):
Could we register backend specific overloads from the backends' modules?  
That would provide better encapsulation.

kszucs (approved):

PR #4037, opened 2022-06-02
refactor: rename CMSMedian to ApproxMedian, and HLLCardinality to ApproxCountDistinct

### LINKED ISSUE #3961: refactor: deprecate CMSMedian and HLLCardinality names
These operations' names are too specific. It's kind of a crap shoot whether a backend implements approximate medians using [count-min sketch](https://en.wikipedia.org/wiki/Count%E2%80%93min_sketch) or something else such as [t-digest](https://arxiv.org/abs/1902.04023).

We should rename `CMSMedian` to `ApproxMedian`.

Similarly, `HLLCardinality` implies a guarantee that the algorithm used to estimate cardinality is [Hyper Log-Log](https://en.wikipedia.org/wiki/HyperLogLog). In many cases this is true, but what we're after is not "approximate count distinct using a specific method" but "approximate count distinct".

We should rename `HLLCardinality` to `ApproxCountDistinct`.

In both cases, we'll deprecate and fully remove in 4.0.0.

anjakefala:
Deprecated, with a plan to be fully removed in 4.0.0.

Closes #3961

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -542,6 +542,9 @@ sqlalchemy_operation_registry: Dict[Any, Any] = {
     ops.BitXor: reduction(sa.func.bit_xor),
     ops.CountDistinct: reduction(lambda arg: sa.func.count(arg.distinct())),
     ops.HLLCardinality: reduction(lambda arg: sa.func.count(arg.distinct())),
+    ops.ApproxCountDistinct: reduction(
+        lambda arg: sa.func.count(arg.distinct())
+    ),
     ops.GroupConcat: _group_concat,
     ops.Between: fixed_arity(sa.between, 3),
     ops.IsNull: _is_null,
--- a/ibis/backends/base/sql/registry/window.py
+++ b/ibis/backends/base/sql/registry/window.py
@@ -254,9 +254,9 @@ def window(translator, expr):
     )
 
     _unsupported_reductions = (
-        ops.CMSMedian,
+        ops.ApproxMedian,
         ops.GroupConcat,
-        ops.HLLCardinality,
+        ops.ApproxCountDistinct,
     )
 
     if isinstance(window_op, _unsupported_reductions):
--- a/ibis/expr/operations/reductions.py
+++ b/ibis/expr/operations/reductions.py
@@ -8,6 +8,7 @@ from ibis.expr import rules as rlz
 from ibis.expr import types as ir
 from ibis.expr.operations.core import Value, distinct_roots
 from ibis.expr.operations.generic import _Negatable
+from ibis.util import deprecated
 
 
 @public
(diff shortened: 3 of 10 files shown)


ibis/expr/operations/reductions.py
```diff
@@ -222,6 +232,13 @@ class CMSMedian(Filterable, Reduction):
     output_dtype = rlz.dtype_like('arg')
 
 
+@public
+@deprecated(version="4.0", instead="use ApproxMedian")
+class CMSMedian(Filterable, Reduction):
```
saulpw, line 237: Instead of duplicating this code, how about just inheriting from the real one?

```
class CMSMedian(ApproxMedian):
     pass
```

ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -541,7 +541,9 @@ def translate(t, expr):
     ops.BitOr: reduction(sa.func.bit_or),
     ops.BitXor: reduction(sa.func.bit_xor),
     ops.CountDistinct: reduction(lambda arg: sa.func.count(arg.distinct())),
-    ops.HLLCardinality: reduction(lambda arg: sa.func.count(arg.distinct())),
```
cpcloud, line 544: Let's leave the duplicated rules in until we delete `HLLCardinality`/`CMSMedian` in case anyone is constructing these operations externally. We'll delete the duplicated rules when we delete the classes.

cpcloud (changes requested):

ibis/backends/base/sql/alchemy/registry.py
anjakefala: I did find this: https://ibis-project.org/docs/3.0.2/user_guide/design/

anjakefala: I am going to attempt to do this, but could you teach me what is meant by rules in this context? The registry lists in the `registry.py`s? Are the `@execute_node.register()` also creating rules?

cpcloud: Sure!

I'm using the word "rule" here informally to mean a mapping from an `ibis.expr.operations.Node` subclass to a function that produces something in the backend's target representation (SQL string, SQLAlchemy expression, Dask DAG, etc.).

`@execute_node.register` **as used in the pandas backend** doesn't quite fit into that because there's no target representation. In that case the rule is mapping an operation to code that will execute the operation directly and return a DataFrame/Series/scalar.

Does that help?

ibis/backends/base/sql/registry/window.py
```diff
@@ -254,9 +254,9 @@ def window(translator, expr):
     )
 
     _unsupported_reductions = (
-        ops.CMSMedian,
+        ops.ApproxMedian,
```
anjakefala, line 257: We do not need to have them both here, because this list is only used by an `isinstance` which handles class inheritance. It will check if a class is derived from a base class in the tuple.

cpcloud (approved):
This looks great! Thanks for taking it on.

PR #4041, opened 2022-06-03
perf: make multiple joins fast again

### LINKED ISSUE #4036: perf: "large compile" test failing with some recent commit
I noticed while working on https://github.com/ibis-project/ibis/pull/4006 that after some recent sync with master that the "large compile" test in the BigQuery backend started failing.

https://github.com/ibis-project/ibis-bigquery/blob/d3acf50de9d8e9015819efe300b4b5f9615c027a/tests/unit/test_compiler.py#L425

I was hoping this was just a misconfiguration with my local machine, but I see now that it is failing / infinitely looping on the GitHub Actions setup as well: https://github.com/ibis-project/ibis-bigquery/runs/6711618312?check_suite_focus=true

cpcloud:
This PR addresses a performance regression when constructing chains of joins.

The proximate issue was the reconstruction of equivalent projections of
a traversed `Selection` using `op.__class__(op.table, op.selections)`.
I've addressed this by adding a cached private `_projection` property that is
tied to unique instances of `Selection`.

There's some more investigation of the repeated computation to be done, but that
will be done in a follow-up.

Fixes #4036.

### DIFF
--- a/ibis/expr/analysis.py
+++ b/ibis/expr/analysis.py
@@ -666,10 +666,7 @@ def _find_root_table(expr):
     if isinstance(op, ops.Selection):
         # remove predicates and sort_keys, so that child tables are considered
         # equivalent even if their predicates and sort_keys are not
-        return lin.proceed, op.__class__(
-            table=op.table,
-            selections=op.selections,
-        )
+        return lin.proceed, op._projection
     elif op.blocks():
         return lin.halt, op
     else:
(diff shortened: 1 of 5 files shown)


kszucs:
We could use a memoization decorator for the analysis functions using a `weakref.WeakKeyDictionary()` instead of the usual `functools.lru_cache()`

kszucs:
We could use a memoization decorator for the analysis functions using a `weakref.WeakKeyDictionary()` instead of the usual `functools.lru_cache()`

kszucs (approved):

PR #4051, opened 2022-06-07
fix(expr): remove checks for expression within same tree

saulpw:
I'd like to remove these checks.  For the VisiData plugin, I do re-use expressions from other trees, and allowing that (by removing these) seems to work fine.  Do we need this requirement or can we loosen it? @kszucs

### DIFF
--- a/ibis/expr/operations/relations.py
+++ b/ibis/expr/operations/relations.py
@@ -379,31 +379,6 @@ class Selection(TableNode, sch.HasSchema):
         default=(),
     )
 
-    def __init__(self, table, selections, predicates, sort_keys):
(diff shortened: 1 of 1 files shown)


kszucs:
> I'd like to remove these checks. For the VisiData plugin, I do re-use expressions from other trees, and allowing that (by removing these) seems to work fine. Do we need this requirement or can we loosen it? @kszucs

Me too, but I'm afraid that we need to keep them around for now. Currently we're trying to remove as much noise as we can from the analysis code so we can refactor these checks using pattern matching. That would eventually allow us simpler expression rewrites, which I assume should be similar to what you do in the VisiData plugin. 

@saulpw Can you point me what kind of expression replacement do you do?

PR #4053, opened 2022-06-07
perf: add memoize decorator with weakly referenced cache

kszucs:
Currently it is limited to positional only arguments which all must be weakly referenceable.

### DIFF
--- a/ibis/common/caching.py
+++ b/ibis/common/caching.py
@@ -66,7 +66,7 @@ class WeakCache(MutableMapping):
 
 def memoize(fn):
     """
-    Cache function calls and keep results until all of the arguments are no longer alive.
+    Cache function call results until all of the arguments are no longer alive.
     """
 
     memo = WeakCache()


kszucs:
@cpcloud this may help performance a bit

ibis/common/caching.py
```diff
@@ -47,3 +62,23 @@ def __delitem__(self, key):
 
     def __repr__(self):
         return f"{self.__class__.__name__}({self._data})"
+
+
+def memoize(fn):
+    """
+    Cache function calls and keep results until all of the arguments are alive.
```
cpcloud, line 69:
```suggestion
    Cache function calls and keep results until all of the arguments are no longer alive.
```

cpcloud:
@kszucs Do we have any existing benchmarks that we would expect to see improvement from after this PR?

kszucs:
Need to check but I still need to cover this with tests.

cpcloud:
@kszucs I like this idea, but it's not clear we have a strong motivation to do it yet. What do you think about closing it out?

cpcloud:
I'm going to convert this to draft until more can be done with it.

PR #4063, opened 2022-06-08
refactor(ir): remove node.root_tables() and unify parent table handling

kszucs:


kszucs:
@cpcloud thanks for the updates! 

If it looks good to you we can merge it.

ibis/expr/operations/core.py
```diff
@@ -67,10 +66,6 @@ def equals(self, other):
     def inputs(self):
         return self.args
 
-    @property
-    def exprs(self):
```
cpcloud, line 71: This is a public API, so I don't think we can remove it yet. I'll update the PR.

kszucs: Or we can mark this PR as 4.0 without preserving backward compat APIs.

jreback: pls just deprecate removing apis is causing breakages

cpcloud: That's the idea!

pyproject.toml
```diff
@@ -47,7 +47,7 @@ dask = { version = ">=2021.10.0", optional = true, extras = [
   "array",
   "dataframe"
 ] }
-datafusion = { version = ">=0.4,<0.7", optional = true }
+datafusion = { version = ">=0.6,<0.7", optional = true }
```
cpcloud, line 50: Since the lockfile didn't change, that means we're testing with 0.6 so I don't think this will help the issue. Can you just xfail only on versions older than 0.6?

kszucs:
@cpcloud I'm not sure what causes the benchmark problem for datafusion. Do we consider that as a blocker?

cpcloud:
@kszucs 

image

jreback:
@cpcloud we are still behind master. e cause of the last set of updates - so we cannot even look at this yet

so what you need to do but as we have discussed on other threads the amount of breaking changes is very problematic

cpcloud (approved):
LGTM.

To be clear, this is *NOT* going into `master`, it's going into the `4.x.x` branch.

PR #4072, opened 2022-06-10
feat: add register method to duckdb backend to load parquet and csv files

gforsyth:
First cut of adding a `register` method to allow loading in `parquet` and `csv` files using the `duckdb` backend.

Needs tests and I should probably add an option to pass through CSV options


ibis/backends/duckdb/__init__.py
```diff
@@ -67,6 +67,51 @@ def do_connect(
...
+
+        # Use prefix for file_type.  If omitted, infer from file extension
+        file_type = prefix.strip(":") if prefix != "file:" else suffix
+        table_name = table_name or file_name.stem.replace("-", "_")
+        if file_type == "parquet":
+            view = f"""
+            CREATE VIEW {table_name} as SELECT * from
+            read_parquet('{file_name}')
+            """
+        elif file_type.startswith("csv"):
```
cpcloud, line 103: What do you think about ad(a|o)pting this code https://github.com/blaze/odo/blob/master/odo/regex.py so we can avoid all the branching?

We can probably make use of this in `ibis.connect` as well.

gforsyth: `blaze` really is the phoenix.

gforsyth:
hmm, does `pep8` speaks not recognize the walrus operator?

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -0,0 +1,67 @@
...
+import ibis
+from ibis.backends.conftest import read_tables
+
+
+@contextlib.contextmanager
+def pushd(new_dir):
+    previous_dir = os.getcwd()
+    os.chdir(new_dir)
+    yield
+    os.chdir(previous_dir)
```
cpcloud, line 16:
```suggestion
    try:
        yield
    finally:
        os.chdir(previous_dir)
```

So that you `chdir` back to the original directory even if an exception occurs in the yieldee.

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -0,0 +1,67 @@
...
+        ),
+        ("functional_alltypes.parquet", "funk_all", "funk_all"),
+        ("parquet://functional_alltypes.parq", "funk_all", "funk_all"),
+        ("parquet://functional_alltypes", None, "functional_alltypes"),
+    ],
+)
+def test_register_parquet(
+    tmp_path, data_directory, fname, in_table_name, out_table_name
+):
+    import pyarrow.parquet as pq
```
cpcloud, line 53:
```suggestion
    pq = pytest.importorskip("pyarrow.parquet")
```

in case someone is developing with duckdb installed but not pyarrow-built-with-parquet.

ibis/backends/duckdb/__init__.py
```diff
@@ -26,6 +27,50 @@ class _ColumnMetadata(NamedTuple):
     type: dt.DataType
 
 
+_register = RegexDispatcher("register")
```
cpcloud, line 30:
```suggestion
_register = RegexDispatcher("_register")
```
The tiniest of nits.

ibis/backends/duckdb/__init__.py
```diff
@@ -67,6 +112,24 @@ def do_connect(
         )
         self._meta = sa.MetaData(bind=self.con)
 
+    def register(
+        self,
+        file_name: str | Path,
```
cpcloud, line 117:
```suggestion
        file_name: str | Path,
```

What do you think about `path` here instead?

ibis/backends/duckdb/__init__.py
```diff
@@ -67,6 +112,24 @@ def do_connect(
...
+        table_name: str | None = None,
+    ) -> None:
+        """Register an external file (csv or parquet) as a table in the current
```
cpcloud, line 120: Probably best to leave off the mention of specific file types in the top level docstring to allow for later expansion without having to update this list. I imagine it will be ever-growing, if slowly.

ibis/backends/duckdb/__init__.py
```diff
@@ -67,6 +112,24 @@ def do_connect(
...
+        connection database
+
+        Parameters
+        ----------
+        file_name
+            Name of the parquet or CSV file
+        table_name
+            Name for the created table.  Defaults to filename if not given
```
cpcloud, line 128: Might be worth documenting here the behavior of `table_name` with respect to dashes if `table_name` is `None`.

ibis/backends/base/regex.py
```diff
@@ -0,0 +1,98 @@
...
+    def add(self, regex, func, priority=10):
+        self.funcs[normalize(regex)] = func
+        self.priorities[func] = priority
+
+    def register(self, regex, priority=10):
+        """Register a new handler in this regex dispatcher.
+
+        Parameters
+        ----------
+        regex : str or Pattern
```
cpcloud, line 64: In general I think we should avoid the colon typing syntax in docstrings and just add the equivalent type annotations. `mkdocstrings` does a better job with type annotations than with the numpydoc type strings.

ibis/backends/base/regex.py
```diff
@@ -0,0 +1,98 @@
...
+            The pattern to match against.
+        priority : int, optional
```
cpcloud, line 66: Similarly, I think adding this `optional` string isn't very useful. Using `?`/`??` in IPython or the builtin `help` function will show the default value. `mkdocstrings` also doesn't do anything with this information and uses the fact that there's a default value to indicate that it's optional.

ibis/backends/base/regex.py
```diff
@@ -0,0 +1,98 @@
...
+            The priority for this pattern. This is used to resolve ambigious
+            matches. The highest priority match wins.
+
+        Returns
+        -------
+        decorator : callable
```
cpcloud, line 72: While this pattern of `variable : type` exists in the codebase in a number of places, I think it's a mistake and doesn't really make any sense. The name of the variable (in the callee's scope) that's returned is meaningless to the caller. It may or may not even be a variable, and none of this information should be communicated in user-facing documentation.

Using the `variable : type` pattern will also confuse `mkdocstrings`. You can just put the type here (`Callable`).

cpcloud (changes requested):
Looking real good! Couple of questions/thoughts.

cpcloud:
> hmm, does `pep8` speaks not recognize the walrus operator?

Unfortunately is does not see https://github.com/OrkoHunter/pep8speaks/issues/173.

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -0,0 +1,67 @@
...
+
+    fname = Path(fname)
+    _, table = next(read_tables([fname.stem], data_directory))
+
+    pq.write_table(table, tmp_path / fname.name)
+
+    con = ibis.duckdb.connect()
+    with pushd(tmp_path):
+        con.register("parquet://" + str(fname.name), table_name=in_table_name)
```
cpcloud, line 62:
```suggestion
        con.register(f"parquet://{fname.name}", table_name=in_table_name)
```

Micronit.

ibis/backends/duckdb/__init__.py
Mause, line 30: While this RegexDispatcher is very cool, why not just use `urlparse` as get the scheme?

ibis/backends/base/regex.py
```diff
@@ -0,0 +1,98 @@
...
+    Returns
+    -------
+    p : Pattern
+        The compiled regex.
+    """
+    r = getattr(r, "pattern", r)
+    return re.compile('^' + r.lstrip('^').rstrip('$') + '$')
+
+
+class RegexDispatcher:
```
kszucs, line 22: Nice utility! 

Could we place this to somewhere else, like `ibis.common.dispatch`?

kszucs:
@gforsyth @cpcloud we have multiple backends supporting csv/parquet files, e.g. pandas or datafusion [but with a different API](https://github.com/ibis-project/ibis/blob/master/ibis/backends/datafusion/__init__.py#L125-L158). 

Could we introduce an interface like a mixin to standardize the API for registering files with different formats?

ibis/backends/duckdb/__init__.py
cpcloud: Two reasons:

1. We need to handle file paths as well, and the code to do that is more readable if we don't mix URI handling and path handling in the same block of code.
2. We're going to reuse this in another API `ibis.connect` that also needs to handle both URIs and paths, and we'd end up either duplicating the logic or extracting it out into a similar amount of code.

ibis/backends/base/regex.py
gforsyth: I'll move it there.  And yeah, it's slick!  @cpcloud pointed me to it.  It's from `odo`: https://github.com/blaze/odo/blob/master/odo/regex.py

kszucs: Then we need to reference it in the LICENSE file :)

cpcloud:
Ideally we can standardize the interface. I think we should to do that in a follow up though since this API adds a lot of value.

@gforsyth How do you feel about taking a crack at this for 4.0?

gforsyth:
> @gforsyth How do you feel about taking a crack at this for 4.0?

I'm happy to give it a shot as a followup

cpcloud (approved):
@gforsyth I changed the `feat` commit to be `feat(duckdb)` so that it shows up alongside the other duckdb notes:

image

PR #4081, opened 2022-06-13
refactor: only display internal duckdb tables if requested

gforsyth:


gforsyth:
Pushed a fix upstream in https://github.com/Mause/duckdb_engine/pull/259 so will close this for now in favor of that

PR #4087, opened 2022-06-14
refactor(test): remove datamgr.py

cpcloud:
This PR removes `ci/datamgr.py` in favor loading backend test data 
whenever a backend test is run.

Notes:

1. Lockfiles are deleted at the end of every **global** test session. That is,
   the end-of-session hook is invoked once. I verified this empirically.
1. The data are reloaded whenever `pytest -m $BACKEND` is invoked to exercise
   the data loading code and to ensure that it's rerun if there are any changes
   to that code. It adds a negligible amount of overhead to the non-Impala
   backend tests. The Impala test suite is about 10% slower after this. Running
   the Impala tests locally I didn't notice a difference in test UX from having
   to load the data whenever `pytest` is invoked.


ibis/backends/clickhouse/tests/conftest.py
```diff
@@ -107,14 +107,13 @@ def least(f: Callable[..., ir.Value], *args: ir.Value) -> ir.Value:
...
 @pytest.fixture(scope='module')
-def con():
-    return ibis.clickhouse.connect(
-        host=CLICKHOUSE_HOST,
-        port=CLICKHOUSE_PORT,
-        user=CLICKHOUSE_USER,
-        password=CLICKHOUSE_PASS,
-        database=IBIS_TEST_CLICKHOUSE_DB,
-    )
+def con(tmp_path_factory, data_directory, script_directory):
```
anjakefala, line 110: Cool! So by factoring it out, you could integrate it into the `connect`.

ibis/backends/conftest.py
```diff
@@ -461,6 +464,31 @@ def lock_load_data(
     return cls(data_directory)
 
 
+def pytest_sessionfinish(session, exitstatus):
```
anjakefala: For reference to anyone curious. I have never heard of this: https://docs.pytest.org/en/7.1.x/_modules/_pytest/hookspec.html#pytest_sessionfinish

kszucs (approved):

PR #4088, opened 2022-06-14
chore: update duckdb-engine to 0.1.9

gforsyth:
Includes upstream fix to exclude the duckdb internal metadata views, so
they no longer show up when running `con.list_tables()`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -429,7 +429,7 @@ numpy = ">=1.14"
 
 [[package]]
 name = "duckdb-engine"
-version = "0.1.8"
+version = "0.1.9"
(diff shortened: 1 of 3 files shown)


cpcloud:
@gil I think you need to run

```
poetry export --dev --without-hashes --no-ansi --extras all > requirements.txt
```

I'll add this to the contributing docs.

gforsyth:
As per a side-discussion, we don't really need to bump this here -- it's a nice-to-have but users can install `duckdb-engine==0.1.9` but it doesn't change the core functionality if they're still on 0.1.8

PR #4089, opened 2022-06-14
chore(nix): add shell shortcut to update reqs/lock files/setup.py

gforsyth:
For those of us who can't remember the assorted incantations.


cpcloud (approved):

gforsyth:
hmm, should we kick over that in-progress job?

PR #4097, opened 2022-06-16
refactor: allow selection subclasses to add new node args

timothydijamco:
Nodes that override `__init__` should take in `**kwargs` and pass `**kwargs` to the superclass's `__init__`. Doing this allows subclasses of that node to define new node arguments, and to allow values for those node arguments that are passed to the node's constructor to be correctly passed up the chain of parent classes, all the way to `Annotable`. (`Annotable` has the logic for actually assigning those values to the class's attributes representing those node arguments.)

The `Selection` node was not doing this, and this PR fixes this.

Examples of classes that already do this:
* [`Join`](https://github.com/ibis-project/ibis/blob/98510c8d4073996e7e6c3b25a758bc15661caf17/ibis/expr/operations/relations.py#L170-L176)
* [`FillNa`](https://github.com/ibis-project/ibis/blob/98510c8d4073996e7e6c3b25a758bc15661caf17/ibis/expr/operations/relations.py#L774-L783)
* [`SimpleCase`](https://github.com/ibis-project/ibis/blob/98510c8d4073996e7e6c3b25a758bc15661caf17/ibis/expr/operations/generic.py#L447-L449)


cpcloud:
@timothydijamco Should this be a `fix:`/`feat:` commit? Is this a bug to make an existing use case work or a new feature that would enable as-of-yet-to-exist use cases?

cpcloud:
It's likely we're going to be splitting up `Selection` into its components (`Project`, `Filter`, `Sort`) in the 4.0 release FYI.

If you can add a test that reflects your use case somewhat, that'd be helpful during the refactoring.

cpcloud (approved):

PR #4102, opened 2022-06-17
test: remove pytest_sessionfinish hook

cpcloud:
This hook isn't necessary because a new temporary directory is created for each pytest invocation.

### DIFF
--- a/justfile
+++ b/justfile
@@ -24,7 +24,7 @@ check *args:
 
 # run pytest for ci; additional arguments are forwarded to pytest
 ci-check *args:
-    poetry run pytest --durations=25 -ra --junitxml=junit.xml --cov=ibis --cov-report=xml:coverage.xml {{ args }}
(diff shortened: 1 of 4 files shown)


justfile
```diff
@@ -24,7 +24,7 @@ check *args:
 
 # run pytest for ci; additional arguments are forwarded to pytest
 ci-check *args:
-    poetry run pytest --durations=25 -ra --junitxml=junit.xml --cov=ibis --cov-report=xml:coverage.xml {{ args }}
```
anjakefala, line 27: What is the default output behaviour without the `-ra` flag? Do we still get stacktraces of fails?

cpcloud: Yep. `-ra` shows all the xfails and skips, which IMO aren't that useful in CI

anjakefala: Ah! I agree with this.

cpcloud:
The DuckDB failures are unrelated. I'm going to merge this so the impala failures are gone in #4103.

PR #4111, opened 2022-06-19
fix(impala): replace connection pool logic with sqlalchemy pooling

cpcloud:
This PR addresses a long standing issue, that I finally was annoyed enough with, where the impala connection pooling implementation depends on garbage collection behavior. This caused the number of checked out cursors to blow past the limit rather than reusing an existing connection that was free. The logic has been replaced entirely with SQLAlchemy's built-in QueuePool which has the desired behavior.


kszucs (approved):
LGTM

kszucs (approved):

PR #4116, opened 2022-06-20
ci: add automation for rebasing `*.x.x` branches

cpcloud:


.github/workflows/ibis-rebase-nightly.yml
```diff
@@ -0,0 +1,82 @@
...
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: output branches list
+        id: set-matrix
+        run: |
+          set -euo pipefail
+
+          branches=$(git ls-remote --heads https://github.com/ibis-project/ibis.git \
+              | sed -e 's#\t# #g' -e 's#refs/heads/##g' \
```
kszucs, line 20: In the future we should do the rebasing only for newer than actual release versions, so we can keep branches for earlier releases as well. We can handle this later though.

kszucs (approved):
Going to merge this and trigger a workflow dispatch event.

PR #4125, opened 2022-06-20
docs(tutorial): correct spelling of Antarctica

anjakefala:
Referenced:https://en.wikipedia.org/wiki/Antarctica


cpcloud (approved):
Sweet, thanks!

PR #3577, opened 2022-03-09
fix(mutate): do not lift table column that results from mutate

gforsyth:
Fixes #3529 
Fixes #3515 

I traced the failure mode in #3529 to `ibis.expr.analysis._lift_TableColumn` -- there wasn't a readily apparent way to fix the failure there (or even how to special case it), but removing the entire code path definitely fixes things.  That's not always the best way to fix things, but I think a lot of the expr rewriting stuff is probably causing a lot of issues on the edges.

In any case, I tried a bit to quantify what, if any, time savings are offered by having the `_lift_TableColumn` path available and the results are below.  Usual caveat around how all benchmarks are terrible.

TLDR: For the runs below, removing the expr rewrite has no performance impacts and allows joining on mutated columns, but there may be cases not covered here that would suffer.

**All times are in seconds**

## Running the whole backend test suite 

Use `pytest-repeat` to run each test 3 times to iron out any jitter (maybe)

`pytest -m "<backend>" ibis/backends/tests/ --count=3`

| Backend | with lifting | without lifting | pytest-repeat count |
|---------|--------------|-----------------|---------------------|
| duckdb  | 68.99        | 69.64           | 3                   |
| sqlite  | 83.13        | 84.95           | 3                   |

## Running only tests that hit `_lift_TableColumn`

tests that hit the code in question are:

`test_generic.py::test_select_filter_mutate`
`test_generic.py::test_mutate_rename`
`test_join.py::test_mutating_join[inner | left | right | outer]`
`test_join.py::test_filtering_join[anti | semi ]`

Only join tests (x5)

    pytest ibis/backends/tests/ -m "duckdb" -k "test_filtering_join or test_mutating_join" -v --count=5

| Backend | with lifting | without lifting | pytest-repeat count |
|---------|--------------|-----------------|---------------------|
| duckdb  | 4.15         | 4.15            | 5                   |
| sqlite  | 9.59         | 9.51            | 5                   |
| mysql   | 34.86        | 34.60           | 5                   |

Only generic tests (x5)

``` bash
pytest ibis/backends/tests/test_generic.py -k "test_select_filter_mutate or test_mutate_rename" -v --count=5 -m "duckdb"
```

| Backend | with lifting | without lifting | pytest-repeat count |
|---------|--------------|-----------------|---------------------|
| duckdb  | 3.60         | 3.59            | 15                  |
| sqlite  | 2.90         | 2.88            | 15                  |
| mysql   | 3.62         | 3.58            | 15                  |

[...]

PR #3357, opened 2022-01-24
fix: mutate + filter no longer compiles to unintentionnal cross join

cpcloud:
Closes #2744.

PR #4135, opened 2022-06-22
refactor: remove intermediate expressions

kszucs:
Both the core and sqlite tests pass now, result which is a pretty good baseline to work with.


icexelloss:
stumbled upon this..

I wonder if the main idea to use `Node` as arguments for `Node` (instead of using `Expr`)? This looks like a pretty aggressive change and I wonder what's the main driving reason for this?

cpcloud:
@icexelloss Thanks for chiming in!

There are a few reasons why we're undertaking this change.

We want to tackle big ticket items all of which are made unreasonably difficult by the current indirection between `Expr`s and `Node`s.

Here's an example that's been top of mind for us recently.

### Replacing the functionality in [`analysis.py`](https://github.com/ibis-project/ibis/blob/master/ibis/expr/analysis.py) with a rule based rewrite system

I prototyped the beginnings of such as system [here](https://github.com/cpcloud/ibis/blob/nix-env-optimize/ibis/expr/optimize.py) based on pattern matching using [`matchpy`](https://github.com/HPAC/matchpy).

Bugs in `analysis.py` have the worst combination of effects: they tend to propagate to many parts of ibis unpredictably and are difficult to debug. The code is hard to reason about because of the way its current pattern matching is implemented. The rules are encoded ad-hoc as complex conditionals and the actions taken when the conditionals are `True` are difficult to follow.

Here's a subset of the issues caused by bugs in analysis.py:

   * #1045
   * #1387
   * #2427
   * #2830
   * #3341
   * #3357
   * #3577
   * #4005

To implement this pattern matching based system, we need to canonicalize the internal representation to `Node`, and avoid jumping back and forth internally between `Expr` and `Node`.

---

Here a few outcomes that we're targeting for the end state of these refactors:

1. `analysis.py` is completely subsumed by the new system
2. End-user APIs require no changes (APIs based on use of internals like `Node` subclasses will unfortunately break)
3. Generated SQL code is nearly as good as it would be if written by hand. For example, no unnecessary projections, fused and pushed-down predicates, column/projection pruning. In some cases this will help backends better optimize plans resulting in both more readable target code *and* faster execution.
4. Conveniences like #2703 can be encoded by maintainers with relative ease as long as a pattern can be written to describe the transformation.

icexelloss:
Thanks @cpcloud. 

> To implement this pattern matching based system, we need to canonicalize the internal representation to Node, and avoid jumping back and forth internally between Expr and Node

Other than this, are there other reason that the `Node/Expr` structure are bad (other than the annoyance of having to jump back and forth in various places). I feel like there is a reason that `Node/Expr` are structured this way in the first place and concerned about that fitting with `matchpy` shadows the benefit that `Node/Expr` are structured this way in the first place.

cpcloud:
> I feel like there is a reason that Node/Expr are structured this way in the first place and concerned about that fitting with matchpy shadows the reason/benefit that Node/Expr are structured this way in the first place.

The reason the separation exists is to keep the API and internal IR separate, which remains true in this PR. Exprs aren't going away, they're going away internally. After this PR we're only dealing with ops inside of ibis (there are some cases where that isn't true in the current PR, and we'll need to address that before merging this).

> I also wonder how bad it is to implement the matchpy based system without this change since I am not familiar with matchpy myself. Or is it impossible without this?

It's not related to matchpy per se. We're considering alternatives like [`kanren`](https://github.com/pythological/kanren/blob/master/doc/graphs.md). Ultimately we need a single set of types for some pattern matching library to operate on to keep the integration low on hacks.

> we spent quite a bit of time on dealing with ibis refactors this year

Sorry to hear that. We're happy to help with that. For example, if there was a breaking change to a public API in a non-major release, we'd like to know about it. I appreciate the PRs and issues y'all have opened too, keep that coming :)

> if so, whether there is a less aggressive way to approach this.

Hard to say, but no other alternatives are proposed and there's a finite amount of time we can spend here.

I think the requirements for an alternative are a path towards the following:

1. ability to remove or fully replace analysis.py
2. no expression API breakage
3. easily-extensible rewrite system for relational algebra rewrites
4. predicate pushdown with less code and ideally no API regressions
5. column pruning with less code and ideally no API regressions

We're not planning to release this until 4.0.0, and there'll be a 3.1 release before that (in the next couple weeks) so there's some time to try things out and give feedback before we merge this.

jreback:
@cpcloud this is a really aggressive change and will likey blow up the world. Can you explain the rationale

cpcloud:
> @cpcloud this is a really aggressive change and will likey blow up the world. Can you explain the rationale

See all the text above. Happy to answer any specific questions or concerns.

jreback:
i read the text above and it doesn't answer the question of why here

in any event please don't merge these u til we have a discussion

cpcloud:
It'll be a bit before we merge it.

Are there any specific questions I can answer?

jreback:
well see @icexelloss yhis seems like a huge amount of churn for a very small gain

so am -1 at this point

kszucs:
> I feel like there is a reason that `Node/Expr` are structured this way in the first place and concerned about that fitting with `matchpy` shadows the reason/benefit that `Node/Expr` are structured this way in the first place.

The main reason to separate them is to provide a typed user facing API and syntax sugar. We need to bound different API methods for differently typed value operations. 

Historically this difference hasn't been that clear since Nodes/Operations were used as factories to instantiate Expression objects. At that time Expression classes held `op`, `name` and `datatype` properties, so they acted as an extension on top of operation instances. The `datatype` information was directly available from the underlying operation and name was the only data attribute held by Expressions which wasn't available directly from the operation. Since https://github.com/ibis-project/ibis/pull/3276 we managed move the name property to the operations by introducing an `[Value]Alias` node, meaning that expressions don't store anything but the underlying operation. This means that Expressions hold just as much information as the operation they wrap which makes them a redundant abstraction - in regard of internal representation. 

The idea here is to unify the ibis internal representation containing only `ops.Node` instances and wrap them with a single `Expr` for user consumption (by simply calling `node.to_expr()`). 

> I also wonder how bad it is to implement the `matchpy` based system without this change since I am not familiar with `matchpy` myself. Or is it impossible without this?

Matchpy requires an object hierarchy of the same type, so I din't think it's possible without this change.  I have a matchpy [prototype implementation](https://github.com/ibis-project/ibis/blob/2be27c8f81d94e5dc51a1608fb86a4153a7ab739/ibis/expr/operations/core.py#L51-L82), but I wasn't able to hook it up while having intermediate expressions. 

> Other than this, are there other reason that the `Node/Expr` structure are bad (I understand the annoyance of having to jump back and forth in various places, which is annoying but not a huge deal IMHO). 

I disagree, the duality of the type hierarchy introduces a serious confusion resulting leaky abstractions and various workarounds specifically applicable for some specific use cases (as an example see @cpcloud's collection of bugs above).

I think it's crucial for ibis's long term success to maintain a simple to understand and strict internal representation with generic utilities (for traversing, matching and replacing nodes in the hierarchy). Otherwise the technical dept we have in ibis is going to block future developments. 

> I am supportive of improving the project health long term since our critical infrastructure depend on it, but honestly we spent quite a bit of time on dealing with ibis refactors this year and this one seem to be even much more breaking so just wonder (1) if this is the right thing to do and (2) if so, whether there is a less aggressive way to approach this.

The reason of all those refactors was to reach a point where we can actually enable this refactor. Lately we merged several PRs aiming to simplify the analysis code in order to understand what happens in those routines at the first place, so this is a "less aggressive" way. I cannot really think of a better approach, but I'm totally open to suggestions. 

> in any event please don't merge these u til we have a discussion

Sure, I don't have any intention to merge this change until we have a clear consensus on how to (or not to) proceed.

kszucs:
closing in favor of #4219

PR #4148, opened 2022-06-25
ci: update dependencies less often

cpcloud:
This PR reduces the dependency update cycle to once a week for conda lock and every 3 days for nix updates. The goal here is to reduce the number of commits from bots while still staying up to date. The nix updates could be less often, but I wanted to avoid dealing with having to split huge commit lists in the update action for now.


kszucs (approved):

kszucs:
The impala tests look flaky, but should be unrelated.

PR #4155, opened 2022-06-28
feat(mysql): implement StringToTimestamp

anjakefala:
* Added test_string_to_timestamp
* Proposed removal of "duplicate" test_string_to_timestamp in the pyspark-specific tests.

Timezones are currently not taken into account. Additionally, we might be able to factor this out so it can be used with other `sqlalchemy` backends. 

There was an existing `test_to_timestamp`, but it turned out to be specifically written for "integer to timestamp".

Potentially out of scope:
- [ ] Timezone behaviour (Should `CONVERT_TZ` be its own operation?)
- [ ] Incorporation of locale settings.

### DIFF
--- a/ibis/backends/mysql/registry.py
+++ b/ibis/backends/mysql/registry.py
@@ -112,6 +112,18 @@ def _timestamp_diff(t, expr):
     return sa.func.timestampdiff(sa.text('SECOND'), sa_right, sa_left)
 
 
+def _string_to_timestamp(t, expr):
+    op = expr.op()
+    sa_arg = t.translate(op.arg)
+    sa_format_str = t.translate(op.format_str)
+    if (op.timezone is not None) and op.timezone.op().value != "UTC":
+        raise com.UnsupportedArgumentError(
+            'MySQL backend only supports timezone UTC for converting'
+            'string to timestamp.'
+        )
+    return sa.func.str_to_date(sa_arg, sa_format_str)
+
+
 def _literal(_, expr):
     if isinstance(expr, ir.IntervalScalar):
         if expr.type().unit in {'ms', 'ns'}:
@@ -190,6 +202,7 @@ operation_registry.update(
         ops.TimestampAdd: fixed_arity(operator.add, 2),
         ops.TimestampSub: fixed_arity(operator.sub, 2),
         ops.TimestampDiff: _timestamp_diff,
+        ops.StringToTimestamp: _string_to_timestamp,
         ops.DateTruncate: _truncate,
         ops.TimestampTruncate: _truncate,
         ops.IntervalFromInteger: _interval_from_integer,
--- a/ibis/backends/tests/test_temporal.py
+++ b/ibis/backends/tests/test_temporal.py
@@ -10,6 +10,7 @@ import pytest
 from pytest import param
 
 import ibis
+import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
 from ibis.backends.pandas.execution.temporal import day_name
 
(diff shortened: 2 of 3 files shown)


ibis/backends/tests/test_temporal.py
```diff
@@ -578,6 +578,41 @@ def test_to_timestamp(backend, con, unit):
     assert result == expected
 
 
+@pytest.mark.parametrize(
+    'fn',
+    [
+        param(lambda t: t.date_string_col.to_timestamp('%m/%d/%y')),
+        param(
+            lambda t: t.date_string_col.to_timestamp(
+                '%m/%d/%y', timezone='UTC'
```
rok, line 587: Try "Pacific/Marquesas"? :)

ibis/backends/tests/test_temporal.py
```diff
@@ -578,6 +578,41 @@ def test_to_timestamp(backend, con, unit):
...
+        'datafusion',
+    ]
+)
+def test_string_to_timestamp(backend, con, fn):
+    table = con.table('functional_alltypes')
+    result = table.mutate(date=fn(table)).execute()
+
+    df = table.execute()
+    df['date'] = pd.to_datetime(
+        df['date_string_col'], utc=True, format="%m/%d/%y"
```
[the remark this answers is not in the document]
anjakefala: Marvellous.

ibis/backends/tests/test_temporal.py
anjakefala: Yes! To test this properly, we will need non-UTC timezones.

ibis/backends/tests/test_temporal.py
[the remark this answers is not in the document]
anjakefala: It is absolutely hideous. I love it.

ibis/backends/tests/test_temporal.py
rok: "Asia/Tehran" around DST jumps can be fun.

ibis/backends/tests/test_temporal.py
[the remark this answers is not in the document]
rok: It's also functional and efficient! :)
I [should take](https://github.com/apache/arrow/blob/master/python/pyarrow/tests/test_compute.py#L1763-L1765) my own advice.

ibis/backends/mysql/registry.py
```diff
@@ -110,6 +110,17 @@ def _timestamp_diff(t, expr):
     return sa.func.timestampdiff(sa.text('SECOND'), sa_right, sa_left)
 
 
+def _string_to_timestamp(t, expr):
+    arg, format_str, timezone = expr.op().args
+
+    sa_arg = t.translate(arg)
+    sa_format_str = t.translate(format_str)
+    # if timezone is not None:
+    #    sa_timezone = t.translate(timezone)
```
cpcloud, line 119:
```suggestion
def _string_to_timestamp(t, expr):
    op = expr.op()

    sa_arg = t.translate(op.arg)
    sa_format_str = t.translate(op.format_str)
    # if (timezone := op.timezone) is not None:
    #    sa_timezone = t.translate(timezone)
```
For new translation rules we're moving away from using `.args` on operations because we want rules to not be sensitive to the order in which arguments are listed in the operation class's body.

cpcloud (changes requested):

cpcloud:
Looks like pyspark is failing the new `test_string_to_timestamp` test (https://github.com/ibis-project/ibis/runs/7242858773?check_suite_focus=true#step:16:110)

ibis/backends/tests/test_temporal.py
```diff
@@ -578,6 +579,74 @@ def test_to_timestamp(backend, con, unit):
...
+        param(
+            '%m/%d/%y',
+            "UTC",
+            marks=pytest.mark.never(
+                ["pyspark"], reason="datetime formatting style not supported"
+            ),
+        ),
+        # pyspark
+        param(
+            'MM/dd/yy',
```
rok, line 596: Maybe:
```suggestion
            '%m/%d/%y',
```

cpcloud (approved):
LGTM. Nice work on removing the pyspark test. 🎉

PR #4168, opened 2022-07-01
chore(impala): workaround retry failure by adding a `retries` pass through argument

cpcloud:
Hopefully this alleviates some of the impala test flakiness until we can merge the 4.x.x branch into master


cpcloud:
I think we'll just live with this until we can merge `4.x.x` into `master` and pick up the connection pool changes.

PR #4169, opened 2022-07-01
bring back null parse

### LINKED ISSUE #3879: Ibis datatype parser doesn't parse "null" string
I noticed that the following no longer parses correctly:

```
schema = [('a', 'null'), ('b', 'null')]
ibis.schema(schema)
```
Because we're missing a null definition here: https://github.com/ibis-project/ibis/blob/master/ibis/expr/datatypes/core.py#L1060. 

Is that intentional?

cpcloud:
This PR fixes a regression where parsing the string `"null"` stopped working. Closes #3879.


gforsyth (approved):
This looks good to me -- and I think we can ignore the Impala failures for now

PR #4193, opened 2022-07-05
ci: ensure tpc-h test deps installed

jcrist:
Since these tests are installed from a separate repo, we'll want to install the requirements from that repo to ensure the tests still run properly. Currently things are broken since `ibis-substrait` is missing.


jcrist:
TPC-H queries ran successfully, looks to have fixed the problem.

Issue #3171, opened 2021-12-16
bug: trace.enable() no longer works in pandas/dask backends


gerrymanoim:
Something with option registration doesn't seem quite right, I haven't dug into why yet:

```python
In [2]: from ibis.backends.pandas import trace as pandas_trace

In [3]: pandas_trace.enable()
---------------------------------------------------------------------------
OptionError                               Traceback (most recent call last)
<ipython-input-3-ede9e3aa5fd1> in <module>
----> 1 pandas_trace.enable()

../ibis/backends/pandas/trace.py in enable()
     86 def enable():
     87     """Enable tracing."""
---> 88     set_option(_TRACE_CONFIG, True)
     89     logging.getLogger('ibis.backends.pandas.trace').setLevel(logging.DEBUG)
     90 

../ibis/config.py in __call__(self, *args, **kwds)
    246     def __call__(self, *args, **kwds):
    247         """Call the the function defined when the object was initialized."""
--> 248         return self.__func__(*args, **kwds)
    249 
    250     @property

.../ibis/config.py in _set_option(*args, **kwargs)
     92 
     93     for k, v in zip(args[::2], args[1::2]):
---> 94         key = _get_single_key(k, silent)
     95 
     96         o = _get_registered_option(key)

../ibis/config.py in _get_single_key(pat, silent)
     59         if not silent:
     60             _warn_if_deprecated(pat)
---> 61         raise OptionError('No such keys(s): %r' % pat)
     62     if len(keys) > 1:
     63         raise OptionError('Pattern matched multiple keys')

OptionError: "No such keys(s): 'pandas.enable_trace'"

In [4]: from ibis.backends.dask import trace as dask_trace

In [5]: dask_trace.enable()
---------------------------------------------------------------------------
OptionError                               Traceback (most recent call last)
<ipython-input-5-9948765eb5ab> in <module>
----> 1 dask_trace.enable()

.../ibis/backends/dask/trace.py in enable()
     72 def enable():
     73     """Enable tracing."""
---> 74     set_option(_TRACE_CONFIG, True)
     75     logging.getLogger('ibis.dask.trace').setLevel(logging.DEBUG)
     76 

.../ibis/config.py in __call__(self, *args, **kwds)
    246     def __call__(self, *args, **kwds):
    247         """Call the the function defined when the object was initialized."""
--> 248         return self.__func__(*args, **kwds)
    249 
    250     @property

.../ibis/config.py in _set_option(*args, **kwargs)
     92 
[...]

PR #4196, opened 2022-07-06
fix(pandas): force backend options registration on trace.enable() calls

gerrymanoim:
Closes https://github.com/ibis-project/ibis/issues/3171. 

We already had similarish logic in the `trace` functions themselves.


cpcloud (approved):

cpcloud:
Sweet! Thanks @gerrymanoim.

PR #4203, opened 2022-07-07
fix(duckdb): support large timestamps by way of pyarrow.Table

cpcloud:
This PR fixes support for timestamps that are too big for nanosecond
resolution.

To do this, I had to call `fetch_arrow_table` (before calling `to_pandas` on
_that_) because arrow handles coarser timestamps correctly and pandas doesn't.

A mildly annoying side effect of doing that is that `array<T>`s are now `numpy`
arrays in the resulting `DataFrame` which is why I had to add another `convert`
overload to convert `array<T>` elements back to lists to avoid breaking the
existing API.

pyarrow seems to be the odd one out, since the other array-supporting backends
all return their `array<T>` elements as `list`s.


jcrist (commented):
Small nit, otherwise looks good to me.

cpcloud:
Finally I can reproduce these bizarre never-before-failing failures. It seems to be only reproducible when running with something other than `--numprocesses auto` (for example `--numprocesses 2`) at the moment. I have no idea why the number of cores being "low" would have such a magical effect.

cpcloud:
This requires a fix in pyarrow, so I'm going to hold off on it until there's a release of pyarrow that contains the necessary fix.

cpcloud:
Sadly, this seems to break our TPC-H queries. I'll look into it today.

ibis/expr/schema.py
```diff
@@ -389,7 +389,7 @@ def apply_to(self, df: pd.DataFrame) -> pd.DataFrame:
                 # assume not equal
                 not_equal = True
 
-            if not_equal or isinstance(dtype, (dt.String, dt.Struct)):
+            if not_equal or not isinstance(dtype, dt.Primitive):
```
kszucs, line 392: Seems much cleaner

kszucs (approved):

kszucs:
@cpcloud can you squash it?

cpcloud:
Squashing!

ibis/expr/schema.py
cpcloud: Thanks!

PR #4205, opened 2022-07-07
feat(connect): add glob and brace expansion

### LINKED ISSUE #4204: feat(duckdb): support the ability to load multiple files in a single command
Request: allow registering multiple tables using glob when connecting to the duckdb backend using `ibis.connect('duckdb://...`

`conn = ibis.connect('duckdb://farmers.csv')` will only give me one table, `farmers`.

To connect to multiple CSVs using one DuckDB connection via the `ibis.connect` function, I instead use `:memory:` and do the following:

```python
conn = ibis.connect('duckdb://:memory:')

files = glob.glob('farm_data/*.csv')
for file in files:
    fname = 'csv://' + file
    tname = file.replace('farm_data/', '').replace('.csv', '')
    conn.register(fname, tname)
```
Ideally we'd get something like this instead:

```python
conn = ibis.connect('duckdb://farm_data/*.csv')
```

cpcloud:
This PR adds glob and brace expansion support for `ibis.connect`. Closes #4204.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -3,9 +3,12 @@ from __future__ import annotations
 import abc
 import functools
 import glob
+import itertools
 import re
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, ClassVar, Iterable, Mapping
+
+import bracex
 
 if TYPE_CHECKING:
     import pandas as pd
@@ -691,8 +694,10 @@ def _(_: str, *, filename: str, extension: str, **kwargs: Any) -> BaseBackend:
     >>> con = ibis.connect("duckdb://relative/path/to/more/data.parquet")
     """
     con = ibis.duckdb.connect(**kwargs)
-    for path in glob.glob(filename):
-        con.register(f"{extension}://{path}")
+    for path in itertools.chain.from_iterable(
+        map(glob.glob, bracex.iexpand(filename))
+    ):
+        con.register(f"file://{path}")
     return con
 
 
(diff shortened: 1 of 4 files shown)


cpcloud:
Yeah, that could make sense. One option might be to allow a list of files to `connect` to mean separate tables, and a glob to indicate what you're describing here: a union of all the tables.

Ideally we can have an API that supports the collection of independent tables case and the union case.

cpcloud:
That API might look like this:

```
ibis.connect(glob.glob("*.parquet"))  # independent tables
ibis.connect("*.parquet")  # tables all have the same schema
```

jcrist:
What about supporting the following:

```python
# a single string works, no globbing supported here
# table name is taken from file name
ibis.connect("farms.csv")

# a list of strings works, again no globbing supported here
# multiple tables created, each with their name taken from file name
ibis.connect(["farms.csv", "farmers.csv"])

# a dict of table name to str works. Globbing _is_ supported here, but not required
# the other 2 can be thought of as a shorthand for this one, which exposes the full
# set of options.
ibis.connect({"farms": "farms.csv", "farmers": "farmers.csv", "crop_history": "history/*.csv"})
```

cpcloud:
I guess a tricky bit there is specify the backend. Making that specification more uniform (through URIs) across backends was part of the idea of `ibis.connect`.

It's not clear how the URI scheme would fit in here, for example would the user be required to pass `duckdb://` at the beginning of every value in the `dict`? A `backend` keyword argument could be a solution there, but then if we're going to make people write `duckdb`, they should just use `ibis.duckdb.connect(...)` and we should try to make lists/dicts/globs work with that.

jcrist:
Yeah, I honestly find the top-level connect weird for files. Database URIs makes sense to me, but files feel a bit odd.

PR #4206, opened 2022-07-07
fix(dask) aggregation with multi-key groupby fails on dask backend

patcao:
# Problem

Dask backend fails When executing .aggregate() with a reduction that gives back multiple columns on a multi-key groupby

```
@reduction(
    input_type=[dt.double],
    output_type=dt.Struct(['mean', 'std'], [dt.double, dt.double]),
)	
def mean_and_std(v):
    return v.mean(), v.std()

alltypes.groupby(['bigint_col', 'int_col']).aggregate(
    mean_and_std(alltypes['double_col']).destructure()
)
```

Fails with:
```
>           raise KeyError(f"{not_found} not in index")
E           KeyError: "['bigint_col', 'int_col'] not in index"
```

The bug is in _coerce_to_dataframe, [util.py#L217](https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/execution/util.py#L217), where we're not keeping the proper index when constructing the meta series. make_meta_series, [util.py#L43](https://github.com/ibis-project/ibis/blob/d1ea2693df8c77d86ac302749f0fa17b2c5f2401/ibis/backends/dask/execution/util.py#L43)

make_meta_series assumes only single column indices and doesn't correctly handle Multiindex.

The incorrect _meta causes an issue in the execution of aggregation when we try to get the groupby keys back [aggregations.py#L105](https://github.com/ibis-project/ibis/blob/master/ibis/backends/dask/execution/aggregations.py#L105)

# Fix

The fix is to have the make_meta_series function properly construct an empty MultiIndex for the meta series.


cpcloud:
@patcao Thanks for the fix!

It looks like this is not only about multikey group by operations, but multikey group by operations with a vectorized udf reduction, is that correct?

Any idea why this code path is only hit for UDF reductions and not builtins like mean/std etc?

cpcloud:
> > Any idea why this code path is only hit for UDF reductions and not builtins like mean/std etc?
> 
> I believe those get handled by dask natively.

That makes sense. Thanks again, merging.

cpcloud (approved):

PR #4216, opened 2022-07-10
refactor: remove deprecated functions and classes

kszucs:
BREAKING CHANGE: functions, methods and classes marked as deprecated are removed now


cpcloud (approved):

cpcloud:
@jreback Thoughts on this? All of these have been deprecated since before 4.x or have had the warning for a while.

**Note that this is targeting the `4.x.x4` branch, NOT `master`**.

jreback:
these are probably ok but we r still catching up on prior charges and would like to test this out. so would appreciate waiting a bit

@icexelloss 
@timothydijamco 
@gerrymanoim

cpcloud:
Let's merge this into `4.x.x`. Changes can be tested against commits from the `4.x.x` branch and are not in `master` so nothing is permanent.

jreback:
the downside is that then this is very difficult to update / find issues / revert

cpcloud:
> the downside is that then this is very difficult to update / find issues / revert

Can you clarify that a bit? What does having this in a PR help with that having it as a single commit on a branch makes more challenging?

cpcloud:
> i don't think having it as a PR helps - so agree there
> 
> but the huge changes in this branch make this almost like a fork

So far, nothing has been merged to `4.x.x` that is a library-wide breaking change. The `4.x.x` branch is ahead by a single commit that is only breaking because a method from the impala backend that was unused in the codebase and untested was removed (https://github.com/ibis-project/ibis/compare/master...4.x.x).
 
> i think updating is going to be quite difficult and painful

Let's avoid conflating individual PRs with the process of using a branch to collect changes.

Let's talk about the specific changes on the individual PRs.

In this PR, we're removing a bunch of code that has been deprecated for at least 2 releases (in some cases more than 2 releases) and has either no effect or has an alternative implementation. I think it's safe to merge this code.

> i would really rethink the strategy of bundling massive breaking changes at all

This is not something that we should resolve in a PR discussion. A GitHub discussion would be the right place for this.

Issue #3987, opened 2022-05-25
feat: support stable semantics hash for ibis expr 


icexelloss:
This is something we did with our internal use of ibis and broken recently with the newer ibis version, so I'd like to discuss the possibility of adding this support in ibis itself.

## Semantics Hash
The idea of semantics hash is that we want to compute an hash to present the result of an ibis expr so they can be cached / communicated between different program that uses ibis expr.

Example:
(1) Program A pass an ibis expr to Program B as input
(2) Program B computes the semantics hash on the ibis expr
(3) Program B checks if it already knows the result of that ibis expr using the semantics hash, and potentially uses the cached result for the computation.

## Internal Implementation
Internally, we have effectively implemented a version of semantics hash on ibis node. Our implementation is basically use cloudpickle to serialize the object and then use `hashlib.sha256` to hash the bytes. This has worked recently well for us, but because this is maintained outside of ibis, this is easily broken (recently broke by the change to make `_safe_name` from a property to a cached property because that causes the same object produces a different pickling results depending on whether `_safe_name` is called)

## Proposed change in Ibis
I think it makes sense to add this feature in ibis for easier maintenance for us as well as benefiting other project that uses ibis. The main change I'd like to propose in ibis is to support stable picking for ibis expr. By stable picking I mean that ibis expr that equals should result in the same bytes when being pickled (assuming the same cloud pickle version) , and then, supporting semantics hashing is relatively easy by just hashing the bytes.

At the minimal, I think I'd like at least to support stable picking for ibis expr, meaning that cached properties and attrs (such like _cached_expr) doesn't affect the result of the serialized bytes and we can build semantics hash internally based on that. Considering that ibis expr/ops are already immutable/hashable, adding stable pickling or semantics hash would seem a nice additional.

We are happy to take this work since we have pretty much this implemented already internally. 

Thoughts? @cpcloud 

cc @emilyreff

saulpw:
A stable hash within a single Ibis version makes a lot of sense.  Does it really have to be stable across Ibis versions though?  That seems like it would be unnecessarily restrictive for many changes (and would not be guaranteed stable across major versions anyway).

icexelloss:
It doesn't need to be stable across ibis version (at least from our use cases). The hash is usually used for duration of a couple of days to weeks. And even the hash does change it is not a big deal, we can always recompute.

cpcloud:
What changes would need to be made to ibis to support this?

icexelloss:
[...]

PR #4217, opened 2022-07-10
refactor(ir): remove DestructValue expressions

kszucs:
Defer `DestructValue` expression's functionality to`ops.StructField`

BREAKING CHANGE: removed ir.DestructValue, ir.DestructScalar and ir.DestructColumn


jreback (changes requested):
oh understand the need to refactor 

but breaking changes like this are pretty painful

kszucs:
The `DestructValue` expression is a pandas execution level optimization leaked to the expression API. In addition to that this expression type is the only one without a corresponding operation, hence the workaround code in `analysis.py`.

Due to the lack of a corresponding operation node #4135 requires this patch. The drawback caused by this PR is that the (rather specific) pandas execution optimization goes away (the reuse of UDF results returning a struct type). 

This optimization should be implemented using a different approach. Either by refactoring the pandas backend to execute the nodes in a topological order and track the dependencies between nodes (experimented with it [here](https://github.com/kszucs/ibis/blob/behemoth-pandas/ibis/common/graph.py#L89)) or pass a mutable cache object along the `execute_node` calls since the `Scope` objects cannot propagate upward in the execution tree.

Additionally I'd like to merge this PR to the `4.x.x` branch which is basically the preparation for the intermediate expression removal. Since we don't have a clear consensus there yet, you can also think of this PR and the (`4.x.x` branch in general) as part of that effort. I'm trying to encapsulate all the independent changes into single commits, so we can easily cherry-pick the ones we agree on.

jreback:
problem is we r still catching up on earlier breaking changes and cant even test anything like this

have found that our internal tests do catch a lot more than ibis tests

conceptually not averse to these but we need some time

kszucs:
> problem is we r still catching up on earlier breaking changes and cant even test anything like this
> 
> have found that our internal tests do catch a lot more than ibis tests

would it be possible to improve upstream's coverage in order to prevent downstream test failures?

> 
> conceptually not averse to these but we need some time

jreback:
cc @icexelloss

cpcloud:
@icexelloss Thanks for chiming in. Is there any way we could get some information about the things that are breaking?

If there are no `BREAKING CHANGE` commits between your current version the next commit you track and things are breaking when using public ibis APIs then we'd really like to know about it so we don't accidentally break something and release it!

icexelloss:
@cpcloud Thanks for chiming in..

I think we are in the tricky situation that lot of our code is not just build against the ibis public API (And to be fair there is no way to build that just using the ibis public API) and we use the ibis expr/node level to build our code (e.g., adding new expr/node, creating sub classes of existing node etc, implement functions to traverse, rewrite expression trees). Ironically breaking public APIs are actually easier for us to fix and expr/node level changes are harder.

The way that we know something might be breaking is really pull the ibis changes and test our internals (for example, the change to make node immutable is merged, we pulled that in and fixes various subclasses that has unhashable args, i.e., list and dict), but that's is not easy even for us to know what things might be breaking until we actually upgrade it.

That being said, I don't have a good idea how best to deal with the situation since we are building heavily on the expr/node layer which is not stable this year.

One thing I can think of is we can try to push some of the stuff we have so that at least those are more protected from future refactors, e.g.,

* https://github.com/ibis-project/ibis/issues/3987
* Tree rewrite libraries (we call this rule libraries which takes a ibis tree, perform transformation and output an ibis tree)
* Some of the custom node, i.e, TableNode backed by pandas DataFrame, and Spark DataFrame (we call these external tables)

cpcloud:
@icexelloss It sounds like you are tracking `master` or working off of a commit that isn't part of a release.

That has the potential to cause frequent breakage that no approach other than "things will never break" can protect against. The current policy is to follow semantic versioning, which means breaking changes between major releases and every other release doesn't break anything. Tracking master may or may not work in that scenario, depending on what the next release will be.

Considering that the time between `2.x` and `3.x` was about 6 months, what about tracking released versions instead and selectively applying patches on top of the release internally as needed?

Then the work to upgrade can be batched to happen after a major release, which will generally be much less frequent than non major releases.

kszucs:
> With regard to this PR, in principle this change seems fine to me but I'd much prefer not to merge this to master without solving the pandas execution backend issue w.r.t duplicate execution. Since we use pandas backend in production internally, merging this to master without solving that would mean a pretty big performance regression and we cannot sync with ibis master any more.

I understand the importance of this problem, so let's try to figure out a proper way to re-implement it. I can see three approaches here, though its challenging due to the top->down nature of the execution:
1. Introduce a new mutable cache variable and pass that to the `execute_node` functions and handle the reuse of struct UDFs manually, though this require some lifetime handling. 
2. A more generic version of the previous step where a first traversal on the tree would create a result store tracking which nodes require which nodes' results, so we could immediately free up results not required anymore. This would further improve (I assume significantly) the execution speed since it would reuse all the intermediate results, not just the struct types. This could actually replace the `Scope` object while keeping the top to bottom execution flow, but not sure how to handle the timecontext in this case. 
3. Rewrite the execution model to follow topological order, this way the nodes' dependencies get calculated once (similarly to the previous approach) but in a bottom to up execution flow. I have an incomplete prototype for it, a couple of pointers:
      - [execution in topological order](https://github.com/kszucs/ibis/blob/behemoth-pandas/ibis/common/graph.py#L89)
      - [node rewrite prior execution](https://github.com/kszucs/ibis/blob/behemoth-pandas/ibis/backends/pandas/operations.py)
      - [dispatch of actual execution](https://github.com/kszucs/ibis/blob/behemoth-pandas/ibis/backends/pandas/execution.py)

I stopped working on my prototype since it requires more advanced rewrites for aggregations where it would be preferable to use a generic tree rewrite tool (like matchpy or a logical programming approach like kanren). Since you mentioned it as one of the features you have, I'd be especially curious for your implementation.*

Any other ideas how could we re-add support for it?

> I think we are in the tricky situation that lot of our code is not just build against the ibis public API (And to be fair there is no way to build our system just using the ibis public API) and we use the ibis expr/node level to build our code (e.g., adding new expr/node, creating sub classes of existing node etc, implement functions to traverse, rewrite expression trees). Ironically breaking public APIs are actually easier for us to fix and expr/node level changes are harder.

I think it's related to the technical dept laying around the internal parts of ibis, ideally this shouldn't be that hard.

> The way that we know something might be breaking is really pull the ibis changes and test our internals (for example, the change to make node immutable is merged, we pulled that in and fixes various subclasses that has unhashable args, i.e., list and dict), but that's is not easy even for us to know what things might be breaking until we actually upgrade it.
> 
> That being said, I don't have a good idea how best to deal with the situation since we are building heavily on the expr/node layer which is not stable.

I've rewritten most of the backends in https://github.com/ibis-project/ibis/pull/4219 to the new intermediate representation without intermediate expressions just operation nodes, and it was a pretty straightforward. Actually it was a much more pleasant  experience than I expected. @cpcloud has rewritten a couple of backends too, I think he can confirm that porting code to the new representation is way easier than it looks like at first glance. 

> One thing I can think of is we can try to push some of the stuff we have so that at least those are more protected from future refactors, e.g.,

Actually that would be great!

> * [feat: support stable semantics hash for ibis expr  #3987](https://github.com/ibis-project/ibis/issues/3987)

Didn't see that issue but seems reasonable depending on the "stableness" requirements. Like having another method like `__hash__` but with `md5` instead of the builtin `hash()`

> * Tree rewrite libraries (we call this rule libraries which takes a ibis tree, perform transformation and output an ibis tree)

* As mentioned above, a generic rule based tree rewrite solution is something we want to have, so this is something I'm really interested in. 

> * Some of the custom node, i.e, TableNode backed by pandas DataFrame, and Spark DataFrame (we call these external tables)

Sounds like another useful feature which seemingly implies multi-backend execution.

kszucs:
@icexelloss @cpcloud I forwarded a global cache mapping to the execute nodes to preserve the cache behaviour for struct udfs. Added it only to the elementwise udfs since the others are not covered with tests.

I think we could go ahead and merge it in order to unblock the upcoming developments, and address possible problems later.

ibis/backends/dask/udf.py
```diff
@@ -63,29 +63,43 @@ def execute_udf_node_groupby(op, *args, **kwargs):
...
+        cache[op] = df
+
+        return df
 
     @execute_node.register(
         ops.ElementWiseVectorizedUDF, *(itertools.repeat(object, nargs))
     )
-    def execute_udf_node_non_dask(op, *args, **kwargs):
-        return op.func(*args)
+    def execute_udf_node_non_dask(op, *args, cache=None, **kwargs):
```
cpcloud, line 93: This function never appears to be hit.

ibis/backends/pandas/execution/util.py
```diff
@@ -118,30 +117,23 @@ def coerce_to_output(
...
-        raise ValueError(f"Cannot coerce_to_output. Result: {result}")
+
+    # scalar result
+    if index is None:
+        # Wrap `result` into a single-element Series.
+        return pd.Series([result], name=result_name)
+
+    # Broadcast `result` to a multi-element Series according to the given
+    # `index`.
+    return pd.Series(
```
cpcloud, line 137: It looks like this line is never reached.

ibis/udf/vectorized.py
```diff
@@ -31,7 +33,14 @@ def _coerce_to_tuple(
     (2) An np.ndarray
     (3) A Series
     """
-    return tuple(data)
+    if util.is_iterable(data):
+        return dict(zip(output_type.names, data))
+    elif isinstance(data, collections.abc.Mapping):
+        return data
```
cpcloud, line 39: Looks like the second and third branches are never hit.

cpcloud (changes requested):
This is looking better each time I review it!

* Can you remove the uncovered code?
* Can you make the `BREAKING CHANGE` note indicate what to do instead? I think we settled on `unpack`, IIRC.

ibis/backends/pyspark/compiler.py
```diff
@@ -222,6 +210,13 @@ def compile_column(t, expr, scope, timecontext, **kwargs):
     return table[op.name]
 
 
+@compiles(ops.StructField)
+def compile_struct_field(t, expr, scope, timecontext, **kwargs):
```
cpcloud, line 214: Probably want to slap `@treat_nan_as_null` on here, from #4050.

cpcloud:
@jreback @icexelloss Please review.

The API is now preserved, with the original test verifying that the UDF is executed once **passing**.

We'd like to merge this PR this week.

cpcloud (approved):
Excellent work. Thank you.

jreback:
will do so

ibis/backends/pandas/udf.py
```diff
@@ -92,13 +92,20 @@ def execute_udf_node_groupby(op, *args, **kwargs):
...
         # We have rewritten op.func to be a closure enclosing
         # the kwargs, and therefore, we do not need to pass
         # kwargs here. This is true for all udf execution in this
         # file.
         # See ibis.udf.vectorized.UserDefinedFunction
-        return op.func(*args)
+
+        # prevent executing UDFs multiple times on different execution branches
+        try:
+            result = cache[op]
```
icexelloss, line 104: I think you also want to pass the timecontext in here, otherwise the results might be wrong (same UDF subtree appears in both the left subtree and right subtree with different time context) and ideally have a test for this case (asof join the same table with UDF, right subtree time context would be expanded so you will have the same op but with different time context)

ibis/backends/tests/test_vectorized_udf.py
```diff
@@ -497,27 +494,33 @@ def test_elementwise_udf_overwrite_destruct_and_assign(
...
+        assert not path.exists()
+        path.touch()
+        return v + 1, v + 2
+
+    struct = add_one_struct_exact_once(udf_alltypes['index'])
+
+    if method == "destructure":
+        expr = udf_alltypes.mutate(struct.destructure())
+    elif method == "unpack":
+        expr = udf_alltypes.mutate(struct=struct).unpack("struct")
```
icexelloss, line 517: Does `unpack` does the same thing as `destructure` but a slight different API?

cpcloud: Yes. We kept `destructure` to hopefully avoid some breakage.

ibis/udf/vectorized.py
```diff
@@ -20,7 +20,7 @@
 )
 
 
-def _coerce_to_tuple(
+def _coerce_to_dict(
```
icexelloss, line 23: Can you remind me what this change is for?

ibis/udf/vectorized.py
```diff
@@ -198,7 +198,7 @@ def _get_coercion_function(self):
                 return _coerce_to_dataframe
             else:
                 # Case 2: Struct output, reduction UDF -> coerce to tuple
```
icexelloss, line 200: Update this comment

ibis/udf/vectorized.py
cpcloud: The comment below:

```
# Case 2: Struct output, reduction UDF -> coerce to tuple
```

Though the comment is now partially incorrect (should be coerce to `dict`), the first part indicates that the output is a struct. `dict` is the correct return type for a scalar struct value.

Previously the names of the fields in the tuple were handled elsewhere (roughly speaking, wherever `Destruct*` pseudo-expression types needed to provide names, essentially in projections or things that desugar to projections).

ibis/backends/dask/udf.py
```diff
@@ -63,29 +63,29 @@ def execute_udf_node_groupby(op, *args, **kwargs):
...
     )
-    def execute_udf_node(op, *args, **kwargs):
+    def execute_udf_node(op, *args, cache=None, **kwargs):
         # We have rewritten op.func to be a closure enclosing
         # the kwargs, and therefore, we do not need to pass
         # kwargs here. This is true for all udf execution in this
         # file.
         # See ibis.udf.vectorized.UserDefinedFunction
+        try:
+            return cache[op]
```
icexelloss, line 73: Can you add some comments here explaining how the cache prevent struct UDF re-execution - Does the struct UDF first gets executed in here then the StructField ops just read from the cache?

icexelloss: Also, does it make sense to cache UDF that is not Struct output? Can those be reused as well? 

Thinking about memory foot print too - in a UDF heavy expression, every UDF execution get cached, which means the total memory footprint could be quite higher than before? (effectively all UDF results are stored twice in memory, one in the cache and one in the DataFrame)?

icexelloss (commented):
I took a look and left some comment/questions.

ibis/backends/tests/test_vectorized_udf.py
icexelloss: kk, `destructure` isn't directly exposed to our users so it shouldn't be too hard for us to update that API surface.

ibis/udf/vectorized.py
icexelloss: kk

ibis/backends/dask/udf.py
cpcloud: It's probably true that the pandas backend will use extra memory given the copying that it does. There's not much we can do there without a rewrite of that backend which isn't in scope at the moment (unless someone wants to volunteer for that!)

Hard to say what happens in the dask case without measuring a representative workload.

I think we should address this later, if it even shows up.

I think @kszucs has done a great job of addressing the original concern of avoiding re-execution of UDFs.

ibis/backends/pandas/udf.py
kszucs: @icexelloss is it possible that https://github.com/ibis-project/ibis/blob/master/ibis/backends/tests/test_timecontext.py#L107 already covers it?

ibis/backends/dask/udf.py
kszucs: That should be the current behaviour already since we don't check for struct datatypes anywhere.

icexelloss: >It's probably true that the pandas backend will use extra memory given the copying that it does. There's not much we can do there without a rewrite of that backend which isn't in scope at the moment (unless someone wants to volunteer for that!)

Yeah I won't object to move forward with this. But should probably document these limitations.

icexelloss: > Possibly. That's a good follow-up issue.

What I meant is for non Struct output these cache probably won't be reused so it's kind of like a waste of memory but I haven't really thought this through.

ibis/backends/pandas/udf.py
icexelloss: I think that test name might be confusing. It is a window operations but I don't think the context are adjusted because it's unbounded window (basically a groupby demean operation). I am not sure what's a good example to demonstrate the issue but I think since `Scope` uses op + timecontext as the lookup key, we should probably use the same key since the semantics of the cache here is the same as `Scope` object if I understand correctly. I am a concerned about subtle correctness bugs if we don't pass timecontext here.

ibis/backends/dask/udf.py
kszucs: We should extend the UDF operations to mark them idempotent, [see how spark handles deterministic VS non-deterministic UDFs](https://spark.apache.org/docs/3.1.3/api/python/reference/api/pyspark.sql.functions.udf.html). DataFusion uses the [volatility enum](https://github.com/apache/arrow-datafusion/blob/41b4e491663029f653e491b110d0b5e74d08a0b6/datafusion/expr/src/signature.rs#L25-L37) for this purpose.

ibis/backends/pandas/udf.py
kszucs: @icexelloss updated, now using a compound `(op, timecontext)` key. 

Could you please add a test case in a follow-up since you are more experienced with the timecontext handling?

ibis/backends/dask/udf.py
kszucs: Created a follow-up issue to track this https://github.com/ibis-project/ibis/issues/4355

icexelloss: Sounds good to me

ibis/backends/pandas/udf.py
icexelloss: Sounds good. Not sure if I can tackle this soon but happy to leave as follow ups.

cpcloud (approved):
This LGTM. Merging!

PR #4221, opened 2022-07-11
fix(duckdb): handle multiple dotted extensions; quote names; consolidate implementations

cpcloud:
- test(duckdb): add `file://` prefix test
- test(duckdb): add gz test
- test(duckdb): test register with name that requires quoting
- fix(duckdb): handle multiple dotted extensions; quote names; consolidate implementations

This PR fixes a couple issues with the current duckdb `register` implementation.

1. `\w+` was used to parse paths, when in actuality paths can be arbitrary bytes.
1. Similarly for extensions, `\w+` prevents `.csv.gz` from being parsed as an extension
1. Table names that require quotes such as `fancy stones` are now quoted

I also did a bit of consolidation of the implementations where it made sense, for example `csv` and `csv.gz`.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Iterator, NamedTuple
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
This looks good! Especially nice to get rid of the gross manual extension handling

PR #4222, opened 2022-07-11
feat(ux): use `rich` to format `Table.info()` output

cpcloud:
**Please read the entire description before reviewing**.

This PR introduces `rich` for outputting the content of `Table.info()`.

Ultimately I want to use `rich` to format the `Node` tree including datatypes
and schemas, so that deeply nested objects are approachable.

This PR is just a taste of using it in a non-invasive place.

**Note: this DOES NOT change the `Expr` repr**.

The `Expr` repr will continue to be the same: optimized for speed and concision.

Here's a screenshot of what it looks like after this PR:

image

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -2,6 +2,7 @@ import decimal
 import importlib
 import io
 import operator
+from contextlib import redirect_stdout
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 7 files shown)


cpcloud:
Here's an example:

image

cpcloud:
This coverage report makes zero sense to me. Somehow pyspark and impala coverage dipped by > 1% even though there's no related code touched here.

PR #4225, opened 2022-07-12
fix(duckdb): ensure that paths with non-extension `.` chars are parsed correctly

cpcloud:
This PR fixes an issue where the non-specific `file://`-or-no-prefix dispatcher rule split on the first `os.extsep` which meant
that paths like `/tmp/python3.10` would return an extension of `10`. Here we change it to use `Path.suffixes` instead.


kszucs:
Can you provide a test case?

PR #4230, opened 2022-07-13
feat(api): add `StructValue.lift()` and `Table.unpack(*args)` API for projecting struct fields

cpcloud:
This PR adds a `Table.unpack()` API to support projecting struct fields into
their child tables as well as a `StructValue.lift` method that lifts struct fields into a table.

Tests and docstrings are included and showcase examples.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -867,6 +867,54 @@ class Table(Expr):
                 )
         return ops.FillNa(self, replacements).to_expr()
 
+    def unpack(self, *columns: str) -> Table:
+        """Project the struct fields of each of `columns` into `self`.
+
+        Existing fields are retained in the projection.
+
+        Parameters
+        ----------
+        columns
+            String column names to project into `self`.
+
+        Returns
+        -------
+        Table
+            The child table with struct fields of each of `columns` projected.
+
+        Examples
+        --------
+        >>> schema = dict(a="struct<b: float, c: string>", d="string")
+        >>> t = ibis.table(schema, name="t")
+        >>> t
+        UnboundTable: t
+          a struct<b: float64, c: string>
+          d string
+        >>> t.unpack("a")
+        r0 := UnboundTable: t
+          a struct<b: float64, c: string>
+          d string
+
+        Selection[r0]
+          selections:
+            b: StructField(r0.a, field='b')
+            c: StructField(r0.a, field='c')
+            d: r0.d
+
+        See Also
+        --------
+        ibis.expr.types.structs.StructValue.lift
+        """
+        columns_to_unpack = frozenset(columns)
+        result_columns = []
+        for column in self.columns:
+            if column in columns_to_unpack:
+                expr = self[column]
+                result_columns.extend(expr[field] for field in expr.names)
+            else:
+                result_columns.append(column)
+        return self[result_columns]
+
     def info(self, buf: IO[str] | None = None) -> None:
         """Show column names, types and null counts.
 
--- a/ibis/expr/types/structs.py
+++ b/ibis/expr/types/structs.py
@@ -112,6 +112,43 @@ class StructValue(Value):
         """Return a mapping from field name to field type of the struct."""
         return util.frozendict(self.type().pairs)
 
+    def lift(self) -> ir.Table:
+        """Project the fields of `self` into a table.
+
+        This method is useful when analyzing data that has deeply nested
+        structs or arrays of structs. `lift` can be chained to avoid repeating
+        column names and table references.
+
+        See also [`Table.unpack`][ibis.expr.types.relations.Table.unpack].
+
+        Returns
+        -------
+        Table
+            A projection with this struct expression's fields.
+
+        Examples
+        --------
+        >>> schema = dict(a="struct<b: float, c: string>", d="string")
+        >>> t = ibis.table(schema, name="t")
+        >>> t
+        UnboundTable: t
+          a struct<b: float64, c: string>
+          d string
+        >>> t.a.lift()
+        r0 := UnboundTable: t
+          a struct<b: float64, c: string>
+          d string
+
+        Selection[r0]
+          selections:
+            b: StructField(r0.a, field='b')
+            c: StructField(r0.a, field='c')
+        """
+        import ibis.expr.analysis as an
+
+        table = an.find_first_base_table(self)
+        return table[[self[name] for name in self.names]]
+
     def destructure(self) -> DestructValue:
         """Destructure `self` into a `DestructValue`.
 
--- a/ibis/tests/expr/test_struct.py
+++ b/ibis/tests/expr/test_struct.py
@@ -5,7 +5,20 @@ import pytest
 import ibis
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
+from ibis import _
 from ibis.tests.util import assert_pickle_roundtrip
+
+
+@pytest.fixture
+def t():
+    return ibis.table(
+        dict(a="struct<b: float, c: string>", d="string"), name="t"
+    )
+
+
+@pytest.fixture
+def s():
+    return ibis.table(dict(a="struct<f: float, g: string>"), name="s")
 
 
 def test_struct_operations():
@@ -42,3 +55,34 @@ def test_struct_pickle():
     )
 
     assert_pickle_roundtrip(struct_scalar_expr)
+
+
+def test_lift(t):
+    assert t.a.lift().equals(t[_.a.b, _.a.c])
+
+
+def test_unpack_from_table(t):
+    assert t.unpack("a").equals(t[_.a.b, _.a.c, _.d])
+
+
+def test_lift_join(t, s):
+    join = t.join(s, t.d == s.a.g)
+    result = join.a_y.lift()
+    expected = join[_.a_y.f, _.a_y.g]
+    assert result.equals(expected)
+
+
+def test_unpack_join_from_table(t, s):
+    join = t.join(s, t.d == s.a.g)
+    result = join.unpack("a_y")
+    expected = join[_.a_x, _.d, _.a_y.f, _.a_y.g]
+    assert result.equals(expected)
+
+
+def test_nested_lift():
+    t = ibis.table(
+        {"a": "struct<b:struct<x: int, y: int>, c: string>", "d": "string"},
+        name="t",
+    )
+    expr = t.a.b.lift()
+    assert expr.schema() == ibis.schema({"x": "int", "y": "int"})


saulpw:
Is `splat` the well-known name for this?

jreback:
wouldn't it be better to support conversion a dict then splatting is standard

cpcloud:
In playing with this I think there are four APIs here:

### `Table`-based

* `unpack(*columns)`: project fields of each struct, **discarding** the base table's existing fields
* `splat(*columns)`: project fields of each struct **keeping** the base table's existing fields

### `Column`-based
* `StructValue.unpack()`: project the struct, **discarding** the base table's existing fields
* `StructValue.splat()`: project the struct **keeping** the base table's existing fields

I'm not sold on `splat`, perhaps `unpack_into()` is better?

cpcloud:
I think I'm going to go with only the `splat` semantics for now to avoid API sprawl for what is effectively the same use case.

ibis/expr/types/structs.py
```diff
@@ -112,6 +112,56 @@ def fields(self) -> Mapping[str, dt.DataType]:
         """Return a mapping from field name to field type of the struct."""
         return util.frozendict(self.type().pairs)
 
+    def unpack(self) -> ir.Table:
+        """Project the fields of `self` into its child table.
```
jcrist, line 116: Terminology question: I would have used called this it's parent table - the fields from inside the struct are projected out into the table containing the struct (the "parent" table). Is that an incorrect view of this term in ibis?

gforsyth (approved):
This is sweet.  And I think `unpack` is a better choice.

ibis/expr/types/relations.py
```diff
@@ -864,6 +864,45 @@ def fillna(
                 )
         return ops.FillNa(self, replacements).to_expr()
 
+    def unpack(self, *columns: str) -> Table:
+        """Project the struct fields `columns` the table.
```
jcrist, line 868: I think this is missing a word, but I'm not sure what/where.

ibis/expr/types/relations.py
```diff
@@ -864,6 +864,45 @@ def fillna(
...
+
+        Examples
+        --------
+        >>> schema = dict(a="struct<b: float, c: string>", d="string")
+        >>> t = ibis.table(schema, name="t")
+        >>> t
+        UnboundTable: t
+          a struct<b: float64, c: string>
+          d string
+        >>> t.unpack("a")
```
jcrist, line 885: If given a nested struct, could this also be make to take `"a.b"` to unpack the fields found at `t.a.b` into `t`?

jcrist (commented):
First, :+1: for `unpack` over `splat`. I think this is a much better name for this operation than `splat`.

As mentioned elsewhere, I find the `StructValue.unpack()` version a bit weird. It's an operation on a column-like-thing, but it returns a table that includes information not present in the `StructValue` itself, but from the column's parent table. I can't think of other operations that work this way. It feels magical, like information not visibly present in the method call is making its way into the result. To the user `t.a` is a column named `a`, but a method on it can return a table containing a column named `d` from the parent table context?

Presumably this would work too:

```
schema = {"a": "struct<b:struct<x: int, y: int>, c: string>", "d": "string"}

t = ibis.table(schema, name="t")

t.a.unpack()  # new table has columns [b, c, d]

t.a.b.unpack()  # new table has columns [x, y, d]?
```

I don't find the arguments that this lets operations chain better compelling. Use a local variable if necessary.

I do like the `Table.unpack` version though. I think having both versions may be confusing for users, since it gives two methods with the same name, but with different semantics.

So in summary, I'm +1 on `Table.unpack`, -1 on `StructValue.unpack`.

ibis/expr/types/structs.py
cpcloud: The root of the expression tree is the output of calling `unpack` in this case. An input to that output is a child of it:

```mermaid
graph BT;
    Table-->Unpack;
```

Similar to arithmetic:

```mermaid
graph BT;
	1-->+;
	2-->+;
```

cpcloud:
> First, +1 for `unpack` over `splat`. I think this is a much better name for this operation than `splat`.
> 
> As mentioned elsewhere, I find the `StructValue.unpack()` version a bit weird. It's an operation on a column-like-thing, but it returns a table that includes information not present in the `StructValue` itself, but from the column's parent table. I can't think of other operations that work this way. It feels magical, like information not visibly present in the method call is making its way into the result. To the user `t.a` is a column named `a`, but a method on it can return a table containing a column named `d` from the parent table context?
> 

> I can't think of other operations that work this way

We have a number of important operations that look below their current context, such as `topk` and our specialized `.any()` that desugars into a correlated subquery depending on how the boolean column it outputs is subsequently used.

Projections and filters do this as well during fusion and pushdown, both of which happen on every construction.

For example:

```
In [1]: import ibis

In [2]: schema = {"a": "struct<b:struct<x: int, y: int>, c: string>", "d": "string"}

In [3]: t = ibis.table(schema, name="t")

In [4]: t[['d']][t.d == "a"]  # seems a bit odd to be able to use `t` here given that you're operating on `t.d`, but we allow this anyway.
Out[4]:
r0 := UnboundTable: t
  a struct<b: struct<x: int64, y: int64>, c: string>
  d string

Selection[r0]
  selections:
    d: r0.d
  predicates:
    r0.d == 'a'
```

> Presumably this would work too:
> 
> ```
> schema = {"a": "struct<b:struct<x: int, y: int>, c: string>", "d": "string"}
> 
> t = ibis.table(schema, name="t")
> 
> t.a.unpack()  # new table has columns [b, c, d]
> 
> t.a.b.unpack()  # new table has columns [x, y, d]?
> ```

Indeed this should work, but it doesn't, thanks for finding it!

`t.a.b.unpack()` would have columns `[x, y, c, d]`. No columns are removed except those of the unpacked struct.

> 
> I don't find the arguments that this lets operations chain better compelling. Use a local variable if necessary.

The UX of having to create a local variable every time you want to unnest another level when you have many levels of nesting is terrible.

I encourage you to poke around some deeply nested data using local variables.

This also conflicts with one of the goals of ibis which is to be composable. Local variables are unfortunately not very composable.

> 
> I do like the `Table.unpack` version though. I think having both versions may be confusing for users, since it gives two methods with the same name, but with different semantics.
> 
> So in summary, I'm +1 on `Table.unpack`, -1 on `StructValue.unpack`.

What differs in their semantics?

Take a look at `test_unpack` and `test_unpack_from_table`, they're making the same assertion with their respective APIs.

ibis/expr/types/structs.py
jcrist: Sure, but that's at the expression level, which I *think* an ibis user reading this docstring shouldn't care about? Rather what they care about is the effect on the data/tables of the operation. Since `unpack()` unpacks from a struct column into the containing table, that's what I think the docstring here should be describing. Not sure how best to phrase that though.

ibis/tests/expr/test_struct.py
```diff
@@ -42,3 +55,34 @@ def test_struct_pickle():
...
+    assert result.equals(expected)
+
+
+def test_nested_unpack():
+    t = ibis.table(
+        {"a": "struct<b:struct<x: int, y: int>, c: string>", "d": "string"},
+        name="t",
+    )
+    expr = t.a.b.unpack()
+    assert expr.columns == list("xycd")
```
jcrist, line 88: Wait, I find this change _super_ confusing - I would've expected `list("xyd")`.

I read `obj.unpack()` as taking all the immediate fields of `obj` and unpacking them into the parent table. Since `c` isn't in `obj` (but rather a parent struct of `obj`), I find it very confusing that it's included in the unpacked columns. What if the column names conflict (say `c` is named `x` instead)?

jcrist:
> What differs in their semantics?
>
> Take a look at test_unpack and test_unpack_from_table, they're making the same assertion with their respective APIs.

It's not so much that they do unrelated things, its 'that the meaning of the `unpack` method is different depending on the object you're calling the method on. If on a table it's "unpack the child columns with these names into this table", while on a `StructValue` it's "unpack this column into it's containing tables". The meaning of the verb `unpack` is different in each context.

ibis/tests/expr/test_struct.py
cpcloud: Hm, yeah that's a good point. I'm not sure I like this change either. It *is* very convenient but incredibly non-obvious and hard to predict.

I think having to explicitly `.unpack()` is better. I'll revisit.

ibis/expr/types/structs.py
cpcloud: > Sure, but that's at the expression level, which I think an ibis user reading this docstring shouldn't care about?

I'm not sure I follow how that would change the terminology used. "Expression level" has no bearing on the tree structure of the object exposed to the user.

ibis/expr/types/relations.py
cpcloud: Is this any less confusing than `t.a.b.unpack()`?

ibis/tests/expr/test_struct.py
cpcloud: Tinkering with this some more I think there are only two things that `t.a.b.unpack()` (t.unpack("a.b") would be no different) could mean:

1. Recursively unpack until the child is no longer a `StructField` operation.
2. Fail, because we want people to be explicit about the operation.

In any case, unpacked struct field names that overlap with existing child table field names are an error because duplicate column names are not allowed. This is true whether or not we take care of unpacking the entire dotted path for the user.

cpcloud:
> > What differs in their semantics?
> > Take a look at test_unpack and test_unpack_from_table, they're making the same assertion with their respective APIs.
> 
> It's not so much that they do unrelated things, its 'that the meaning of the `unpack` method is different depending on the object you're calling the method on. If on a table it's "unpack the child columns with these names into this table", while on a `StructValue` it's "unpack this column into it's containing tables". The meaning of the verb `unpack` is different in each context.

I think we may have different philosophical viewpoints about what the meaning of an operation is.

I believe the meaning of an operation is a distinct idea from how the operation is expressed similar to the contrast between implementation and interface. I can express many operations in many different ways but the end result of those operations is the same. For example, I can call `t.x.mean()` or I can call `t.x.sum() / t.x.count()`. In either case, I'm computing the average.

In this case I view the meaning of this operation as "lift the struct fields into the table that contains the struct". Where that is expressed and how it is expressed does not change this meaning.

I wonder if the notion of meaning you have in mind is related not to the consequence of the operation but to notation. From that perspective I can see that `t.x.mean()` and `t.x.sum() / t.x.count()` would have a different meaning since the way the desired output is expressed in each case is different.

ibis/tests/expr/test_struct.py
cpcloud: Ok, I think was confused by why you were expecting `xyd` here, but now I see understand. It seems like there is a third option for behavior which is inlining the struct without unpacking the children.

cpcloud: Alright, I've implemented this in a way that I think reflects a reasonable compromise between convenience and semantic clarity. `t.a.b.unpack()` works, but doesn't recursively unpack, it pulls up the nested struct into top level fields.

ibis/expr/types/relations.py
cpcloud: I think `into the table` is what I meant to write here.

ibis/tests/expr/test_struct.py
jcrist: Yeah, that's what I was asking for (inline, but not recursively). Sorry that wasn't clear above.

jcrist:
> I think we may have different philosophical viewpoints about what the meaning of an operation is.

I don't disagree with any of that.

I think we're talking past each other here. Let me try to restate the problems I see here:

**1. Methods that return results based on implicit parameters**

The `StructValue.unpack` method takes no arguments, and is present on a `StructValue` object. I'd expect the return value of this unpack call to rely only on the `StructValue` itself. However, it instead relies on the table the `StructValue` is part of. This is an *implicit* parameter to this method - something that changes the output that isn't visible in the call structure.

Perhaps the following example will make this clearer:

```python
In [1]: import ibis

In [2]: ibis.options.interactive = True

In [3]: con = ibis.duckdb.connect()

In [4]: t1 = con.register("test.parquet")

In [5]: t1
Out[5]: 
   a                 b
0  1  {'x': 1, 'y': 2}
1  2  {'x': 3, 'y': 4}
2  3  {'x': 5, 'y': 6}

In [6]: t2 = t1.mutate(c=t1.a + 1)

In [7]: b1 = t1.b

In [8]: b2 = t2.b

In [9]: b1
Out[9]: 
0    {'x': 1, 'y': 2}
1    {'x': 3, 'y': 4}
2    {'x': 5, 'y': 6}
Name: b, dtype: object

In [10]: b2  # executes to exactly the same value as b1...
Out[10]: 
0    {'x': 1, 'y': 2}
1    {'x': 3, 'y': 4}
2    {'x': 5, 'y': 6}
Name: b, dtype: object

In [11]: b1.unpack()
Out[11]: 
   a  x  y
0  1  1  2
1  2  3  4
2  3  5  6

In [12]: b2.unpack()  # but results of unpack method call differ!
Out[12]: 
   a  x  y  c
0  1  1  2  2
1  2  3  4  3
2  3  5  6  4
```

In this example, both `b1` and `b2` are `StructColumn` objects that both `execute` to the same result. However, calling the `unpack()` method on these seemingly identical objects returns different results, since the result depends *implicitly* on the containing table (since the table isn't visibly part of the method call).

In contrast, `Table.unpack` doesn't hide this information flow - everything the result depends on is visible in the `unpack` method call. The user can be certain that if `t1.execute() == t2.execute()`, then `t1.unpack("b") == t2.unpack("b")`.

**2. The subject the verb `unpack` applies to differs between `Table.unpack` and `StructColumn.unpack`**

This is what I was getting at in my previous comment. `Table.unpack` means "unpack the child columns with these names into this table" - the verb "unpack" applies to the specified columns, and they're being unpacked into the table (`self`). In contrast, `StructColumn.unpack` means "unpack this column into its containing table" - the verb "unpack" applies to the column itself (`self`).

As a kind of absurd example, imagine if Python ints also had an `append` method for appending themselves to a list:

```python
>>> a = []
>>> x = 1
>>> a.append(x)  # append x to list a, normal `list.append` method
>>> x.append(a) # append x to list a, new `int.append` method
```

The new `int.append` method is well defined and definitely implementable, but the `list` and `element` are swapped from `list.append` - the meaning of the `append` verb with respect to its arguments is no longer the same. In this case, an `int.append_to` method might be a better name.

Does that make sense?

ibis/expr/types/relations.py
cpcloud: Let's avoid adding the `.` for now and wait to see how this gets used.

cpcloud:
@jcrist Any more comments here?

ibis/expr/types/relations.py
```diff
@@ -867,6 +867,54 @@ def fillna(
...
+        """
+        columns_to_unpack = frozenset(columns)
+        result_columns = []
+        for column in self.columns:
+            if column in columns_to_unpack:
+                expr = self[column]
+                result_columns.extend(expr[field] for field in expr.names)
+            else:
+                result_columns.append(column)
+        return self[result_columns]
```
jcrist, line 916: This currently will silently no-op if a column named passed in doesn't exist in the table. That should probably error instead.

Also, does the order of columns ever matter in ibis? If so, we should probably iterate over the passed in `columns` in order.

ibis/expr/types/structs.py
```diff
@@ -112,6 +112,51 @@ def fields(self) -> Mapping[str, dt.DataType]:
         """Return a mapping from field name to field type of the struct."""
         return util.frozendict(self.type().pairs)
 
+    def lift(self) -> ir.Table:
+        """Project the fields of `self` into a table.
+
+        Existing fields from the child table are not retained in the
+        projection.
```
jcrist, line 119: I still think the phrase "child table" is confusing to users (or at least to myself) when thinking about tables and columns, rather than expressions. Feels like an implementation detail. I also wouldn't expect the fields from the containing table to be present, so mentioning it here feels unnecessary, I might just drop this whole sentence.

jcrist (commented):
Overall this looks pretty good to me! Is anything needed to add these methods to the api docs?

cpcloud:
> Overall this looks pretty good to me! Is anything needed to add these methods to the api docs?

Thanks! Nothing is needed to get these into the docs. They'll be picked up by `mkdocstrings`.

ibis/expr/types/structs.py
cpcloud: Fair enough. I'll remove it.

ibis/expr/types/relations.py
cpcloud: > This currently will silently no-op if a column named passed in doesn't exist in the table. That should probably error instead.

Makes sense.

> Also, does the order of columns ever matter in ibis? If so, we should probably iterate over the passed in columns in order.

Column order matters in ibis.

In this case I'm preserving the relative order of the field in the output. What alternative did you have in mind?

cpcloud: Agreed. The way I've implemented it preserves **an** order, but of course not the only one. We could for example stick the unpacked fields at the beginning or the end of the projection but that seemed more arbitrary than preserving the unpacked fields existing starting offset.

ibis/expr/types/structs.py
cpcloud: Done.

PR #4233, opened 2022-07-14
feat(connect): support globs and brace expansion using `bracex` in `ibis.connect`

cpcloud:
This PR is a slight rework of #4205.

The original discussion was around the ambiguity of a glob pattern in `ibis.connect`.

My thinking is that since `ibis.connect` always returns a `BaseBackend`
instance, the only reasonable thing to do with a glob in `ibis.connect` is to
register individual tables one by one, expanding the glob pattern directly in
Python.

`ibis.connect` is not meant to cover the case of globs of files with the same
schema and is provided as a convenience.

On the other hand I think it makes sense for `con.register(glob_pattern)` to
handle the union case: its API is explicitly per-table.
The main wrinkle with the glob pattern there is that we have to at least expand
the glob to 2 elements to force the user to pass a `table_name` argument.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -2,9 +2,13 @@ from __future__ import annotations
 
 import abc
 import functools
+import glob
+import itertools
 import re
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, ClassVar, Iterable, Mapping
+
+import bracex
 
 if TYPE_CHECKING:
     import pandas as pd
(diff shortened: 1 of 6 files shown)


ibis/backends/base/__init__.py
```diff
@@ -2,10 +2,14 @@
 
 import abc
 import functools
+import glob
+import itertools
 import re
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, ClassVar, Iterable, Mapping
 
+import bracex
```
jreback, line 11: is this a light dep?

cpcloud: The current release of `bracex` has no dependencies https://github.com/facelessuser/bracex.

cpcloud:
> > `ibis.connect` is not meant to cover the case of globs of files with the same schema and is provided as a convenience.
> > On the other hand I think it makes sense for `con.register(glob_pattern)` to handle the union case: its API is explicitly per-table. The main wrinkle with the glob pattern there is that we have to at least expand the glob to 2 elements to force the user to pass a `table_name` argument.
> 
> Sorry, I'm having a bit of trouble parsing what `ibis.connect` vs `ibis.duckdb.connect` supports now w/ respect to globbing. Can you provide a few examples of what works/what's forbidden?

Totally.

After this PR:

1. `ibis.connect` globs always connect to the database then register the tables
   that match the glob. There's no `UNION ALL`-ing behavior here: one file, one
   table.
1. `ibis.duckdb.connect` is the same as it was before: it accepts a path to a
   DuckDB database and a `read_only` flag.
1. When `con.register` (where `con` is a DuckDB backend instance) is called
   with a glob pattern, the files matching that glob are assumed to have the
   same schema, and will result in `UNION ALL` behavior.

Does that make sense? Is it confusing?

* `ibis.connect` is really about convenience above all else: it combines `con =
  ibis.duckdb.connect(...)` and `con.register` to make mapping files to tables
  in the database easy.
* `ibis.duckdb.connect` is not related to globbing and is only for connecting
  to the database
* `con.register` is one level of abstraction down: you can register individual
  tables or glob patterns to assemble a `UNION ALL` of individual files into a
  single table.

cpcloud:
After considering your thoughts on this @jcrist I agree with you this still has too many points of interaction that are confusing.

I'm going to close it out until it comes up from a user and/or someone wants to spend some more time making this less confusing!

PR #4239, opened 2022-07-15
chore(deps): drop `cached-property` dependency

jcrist:
This is now available as `functools.cached_property` in Python 3.8+. An
external dependency should be unnecessary afaict.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -9,8 +9,6 @@ from typing import TYPE_CHECKING, Any, Callable, ClassVar, Iterable, Mapping
 if TYPE_CHECKING:
     import pandas as pd
 
-from cached_property import cached_property
-
 import ibis
 import ibis.common.exceptions as exc
 import ibis.config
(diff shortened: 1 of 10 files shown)


cpcloud (approved):
codecov's `-1` line report is ... something.

Sweet PR.

Looks like the lockfile is still stale.

jcrist:
> Looks like the lockfile is still stale.

Ah, needed  to also update the `requirements.txt` file. That wasn't noted in the [dev docs](https://ibis-project.org/docs/3.0.2/contribute/05_maintainers_guide/#automatic-dependency-updates). Since the command is a bit long, and they need to remain in sync, could all that be rolled into the same script in `/dev`? At worst, I'll add that to the things to fix in the dev docs.

cpcloud:
There's a script in the nix setup that we could easily dump into a shell script to work in any platform that has bash.

cpcloud:
https://github.com/ibis-project/ibis/blob/master/shell.nix#L52-L54

cpcloud:
Courtesy of @gforsyth for the same reason :)

PR #4240, opened 2022-07-15
ci: delay codecov reporting

jcrist:
Tries to delay codecov reporting until most CI runs have finished,
should cut down on PR noise.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,12 +1,21 @@
+codecov:
+  notify:
+    # a high number to try to delay codecov reporting until most of the test
+    # runs have finished. Should match with comment.after_n_builds below.
+    after_n_builds: 32
+
 comment:
+  after_n_builds: 32
   layout: "reach, diff, files"
   behavior: default
   require_changes: true # if true: only post the comment if coverage changes
   require_base: false # [yes :: must have a base report to post]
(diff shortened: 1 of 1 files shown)


gforsyth:
I'm very much in favor of quieting `codecov` down.  Should note that for docs-only builds, I think we only have around 16 CI jobs because we skip all the backend tests -- does `codecov` do the smart thing and realize it's the only job that hasn't run?  Or does it just hang forever?

gforsyth:
> Do we need coverage reporting for docs only builds though?

Definitely not, so we could just skip it entirely

gforsyth:
> Definitely not, so we could just skip it entirely

In fact, if I remember, they routinely fail on docs-only PRs because it thinks that coverage has "decreased" be a precipitous amount (since most tests aren't run)

cpcloud:
I would expect the

```
ignore:
   - "docs/**"
```

bit to take care of this. I guess it doesn't though.

jcrist:
Hmmm, the comment posting already makes me wonder if this worked. I can't remember if when I set this up for dask if it worked on the initial PR, or only on follow-up PRs.

cpcloud (approved):

PR #2224, opened 2020-05-31
ENH: Add properly cume_dist and percent_rank ops

xmnlab:
This PR aims to resolve the confusion between cume_dist and percent_rank.

This gist shows the difference between both functions: https://gist.github.com/xmnlab/d676ff1b0ff474c634d62010ebca8b07 

depends on #2228  #2232 

resolve #949 
resolve #1975

PR #4241, opened 2022-07-15
feat(api): add cume_dist operation

### LINKED ISSUE #3590: feat(api): add cume_dist function
We have `percent_rank` (doesn't include the upper bound) but not `cume_dist` (includes the upper bound).

Let's add it.

xref: #1975

gforsyth:
I didn't get this working with PySpark but it seems like the rest are
doing ok.

Resolves #3590

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -361,6 +361,7 @@ def _window(t, expr):
         ops.MinRank,
         ops.NTile,
         ops.PercentRank,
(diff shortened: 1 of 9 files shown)


gforsyth:
lol, well, I just found https://github.com/ibis-project/ibis/pull/2224 which matches with what I've done pretty well.

On that PR there's a note that 
```
        'Pyspark cume_dist() function indexes from 0 '
        'instead of 1, and does not match expected '
        'output of ibis expressions.'
```

PR #4244, opened 2022-07-16
chore(dev-deps): remove explicit `platformdirs` dependency

cpcloud:
We no longer need to artificially constrain this dependency.


cpcloud:
I'm going to close this out in favor of a PR that removes all the dev-deps constraints.

PR #4246, opened 2022-07-16
feat(api): add `ops.StructColumn` operation

cpcloud:
This PR implements the ability to construct a struct column using `ibis.struct`
from a mix of literal values and column expressions and enables the mix and
match of expression type for `ibis.array`.


ibis/backends/duckdb/registry.py
```diff
@@ -177,6 +178,25 @@ def _string_agg(t, expr):
...
+                "{} := {}".format,
+                op.names,
+                map(
+                    methodcaller(
+                        "compile", compile_kwargs=dict(literal_binds=True)
+                    ),
+                    map(t.translate, op.values),
+                ),
+            ),
+        )
```
jcrist, line 196: Small nit: I find a single generator expression to be much more readable than nested map calls. Untested (and unlinted), but I believe the following is equivalent:

```suggestion
        *(
            sa.text(f"{name} := {t.translate(value).compile(literal_binds=True)}")
            for name, value in zip(op.names, op.values)
        )
```

cpcloud: Yeah, that's better. Thanks!

cpcloud:
I'll retry codecov one more time.

PR #4256, opened 2022-07-19
feat(api): Backend tab completion

jcrist:
- Adds tab completion for backends, so e.g. `ibis.sq<tab>` will autocomplete to `ibis.sqlite`.
- Removes need for `importlib_metadata` backport dependency

### DIFF
--- a/ibis/__init__.py
+++ b/ibis/__init__.py
@@ -1,4 +1,7 @@
 """Initialize Ibis module."""
+from __future__ import annotations
+
+import importlib.metadata as _importlib_metadata
 
 # Converting an Ibis schema to a pandas DataFrame requires registering
 # some type conversions that are currently registered in the pandas backend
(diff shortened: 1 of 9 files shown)


ibis/__init__.py
```diff
@@ -12,18 +15,29 @@
...
 from ibis.expr.api import *  # noqa: F401,F403
 
-try:
-    import importlib.metadata as importlib_metadata
-except ImportError:
-    # TODO: remove this when Python 3.9 support is dropped
-    import importlib_metadata
-
-__all__ = ['api', 'ir', 'util', 'IbisError', 'options']
+__all__ = ['api', 'ir', 'util', 'BaseBackend', 'IbisError', 'options']
```
gforsyth, line 18: I'm not clear on why we're exposing `BaseBackend` here

jcrist: :shrug:. It was the only thing exposed in the top level namespace that wasn't also part of `__all__`. Figured we might as well be consistent here.

cpcloud (approved):
Sweet. This is great.

PR #4259, opened 2022-07-20
feat(duckdb): unsigned integer support

### LINKED ISSUE #4224: bug(sqlalchemy): unsigned integer produces error at expression compilation
Unsigned integer types produce error at compilation of expression:
```
import ibis
tbl = ibis.table(name='A', schema=[('a', 'uint64')])
compiled = ibis.duckdb.compile(tbl['a'])
```
Produces `KeyError`, while similar snippet with column `'a'` of `'int64'` type works good.

jcrist:
- Adds unsigned integer support for the sqlalchemy backends, only implemented for duckdb so far. Things work, but I'm 100% sure there's a better way.
- Modifies the integer numeric promotion rules so that (some) unsigned integer arithmetic works

At this point for duckdb, you can:
- create tables with unsigned integers in the schema
- load tables with unsigned integer columns
- compute reductions on unsigned integer columns
- do (some) arithmetic between unsigned integer columns

Still needs more work and tests.

Fixes #4224.

### DIFF
--- a/ibis/backends/base/sql/alchemy/datatypes.py
+++ b/ibis/backends/base/sql/alchemy/datatypes.py
@@ -9,6 +9,7 @@ from sqlalchemy.dialects.mysql.base import MySQLDialect
 from sqlalchemy.dialects.postgresql.base import PGDialect
 from sqlalchemy.dialects.sqlite.base import SQLiteDialect
 from sqlalchemy.engine.interfaces import Dialect
+from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.types import UserDefinedType
 
 import ibis.expr.datatypes as dt
@@ -31,6 +32,41 @@ class StructType(UserDefinedType):
     def get_col_spec(self, **_):
         pairs = ", ".join(f"{k} {v}" for k, v in self.pairs)
         return f"STRUCT({pairs})"
+
+
+class UInt64(sa.types.Integer):
+    pass
+
+
+class UInt32(sa.types.Integer):
+    pass
+
+
+class UInt16(sa.types.Integer):
+    pass
+
+
+class UInt8(sa.types.Integer):
+    pass
+
+
+@compiles(UInt64, "postgresql")
+@compiles(UInt32, "postgresql")
+@compiles(UInt16, "postgresql")
+@compiles(UInt8, "postgresql")
+@compiles(UInt64, "mysql")
+@compiles(UInt32, "mysql")
+@compiles(UInt16, "mysql")
+@compiles(UInt8, "mysql")
+@compiles(UInt64, "sqlite")
+@compiles(UInt32, "sqlite")
+@compiles(UInt16, "sqlite")
+@compiles(UInt8, "sqlite")
+def compile_uint(element, compiler, **kw):
+    dialect_name = compiler.dialect.name
+    raise TypeError(
+        f"unsigned integers are not supported in the {dialect_name} backend"
+    )
 
 
 def table_from_schema(name, meta, schema, database: str | None = None):
@@ -62,6 +98,10 @@ ibis_type_to_sqla = {
     dt.Int16: sa.SmallInteger,
     dt.Int32: sa.Integer,
     dt.Int64: sa.BigInteger,
+    dt.UInt8: UInt8,
+    dt.UInt16: UInt16,
+    dt.UInt32: UInt32,
+    dt.UInt64: UInt64,
     dt.JSON: sa.JSON,
 }
 
--- a/ibis/backends/duckdb/compiler.py
+++ b/ibis/backends/duckdb/compiler.py
@@ -1,3 +1,7 @@
+from sqlalchemy.ext.compiler import compiles
+
+import ibis.backends.base.sql.alchemy.datatypes as sat
+import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 from ibis.backends.base.sql.alchemy import (
     AlchemyCompiler,
@@ -14,6 +18,28 @@ class DuckDBSQLExprTranslator(AlchemyExprTranslator):
     # the updated `operation_registry` from postgres
     _type_map = AlchemyExprTranslator._type_map.copy()
     _has_reduction_filter_syntax = True
+
+
+@compiles(sat.UInt64, "duckdb")
+@compiles(sat.UInt32, "duckdb")
+@compiles(sat.UInt16, "duckdb")
+@compiles(sat.UInt8, "duckdb")
+def compile_uint(element, compiler, **kw):
+    return element.__class__.__name__.upper()
+
+
+try:
+    import duckdb_engine
+except ImportError:
+    pass
+else:
+
+    @dt.dtype.register(duckdb_engine.Dialect, sat.UInt64)
+    @dt.dtype.register(duckdb_engine.Dialect, sat.UInt32)
+    @dt.dtype.register(duckdb_engine.Dialect, sat.UInt16)
+    @dt.dtype.register(duckdb_engine.Dialect, sat.UInt8)
+    def dtype_uint(_, satype, nullable=True):
+        return getattr(dt, satype.__class__.__name__)(nullable=nullable)
 
 
 rewrites = DuckDBSQLExprTranslator.rewrites
--- a/ibis/expr/rules.py
+++ b/ibis/expr/rules.py
@@ -275,22 +275,22 @@ def shape_like(name):
 # TODO(kszucs): might just use bounds instead of actual literal values
 # that could simplify interval binop output_type methods
 # TODO(kszucs): pre-generate mapping?
-def _promote_numeric_binop(exprs, op):
-    bounds, dtypes = [], []
-    for arg in exprs:
-        dtypes.append(arg.type())
-        if hasattr(arg.op(), 'value'):
-            # arg.op() is a literal
-            bounds.append([arg.op().value])
-        else:
-            bounds.append(arg.type().bounds)
+def _promote_integral_binop(exprs, op):
+    dtypes = []
+    bounds = []
+    for expr in exprs:
+        try:
+            bounds.append([expr.op().value])
+        except AttributeError:
+            dtypes.append(expr.type())
+            bounds.append(expr.type().bounds)
 
+    all_unsigned = dtypes and util.all_of(dtypes, dt.UnsignedInteger)
     # In some cases, the bounding type might be int8, even though neither
     # of the types are that small. We want to ensure the containing type is
     # _at least_ as large as the smallest type in the expression.
-    values = starmap(op, product(*bounds))
-    dtypes += [dt.infer(value) for value in values]
-
+    values = list(starmap(op, product(*bounds)))
+    dtypes.extend(dt.infer(v, prefer_unsigned=all_unsigned) for v in values)
     return dt.highest_precedence(dtypes)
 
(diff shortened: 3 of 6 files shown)


ibis/backends/base/sql/alchemy/datatypes.py
```diff
@@ -33,6 +33,22 @@ def get_col_spec(self, **_):
         return f"STRUCT({pairs})"
 
 
+class UInt64(sa.types.Integer):
```
jcrist, line 36: This opts for creating a single unsigned integer type used across all backends (with specialization handled by the `compiles` decorator), rather than one specific to the duckdb backend. I don't know if this is a good idea, but it seems to work.

Note that mysql also supports unsigned integers, and the mysql dialect has its own unsigned integer sqlalchemy types that we may want to use for mysql instead of our own custom ones here. See e.g. https://docs.sqlalchemy.org/en/14/dialects/mysql.html#sqlalchemy.dialects.mysql.BIGINT.__init__.

ibis/backends/duckdb/compiler.py
```diff
@@ -16,6 +21,46 @@ class DuckDBSQLExprTranslator(AlchemyExprTranslator):
...
+
+
+@compiles(dt.UInt16, "duckdb")
+def compile_uint16(element, compiler, **kw):
+    return "USMALLINT"
+
+
+@compiles(dt.UInt8, "duckdb")
+def compile_uint8(element, compiler, **kw):
+    return "UTINYINT"
```
jcrist, line 41: The `@compiles` decorator lets us specialize how our custom unsigned sqlalchemy types compile for a specific dialect.

ibis/backends/duckdb/compiler.py
```diff
@@ -16,6 +21,46 @@ class DuckDBSQLExprTranslator(AlchemyExprTranslator):
...
+
+
+@ibis_dt.dtype.register(duckdb_engine.Dialect, dt.UInt16)
+def dtype_uint16(_, satype, nullable=True):
+    return ibis_dt.UInt16(nullable=nullable)
+
+
+@ibis_dt.dtype.register(duckdb_engine.Dialect, dt.UInt8)
+def dtype_uint8(_, satype, nullable=True):
+    return ibis_dt.UInt8(nullable=nullable)
```
jcrist, line 61: This implements the reverse mapping, turning the sqlalchemy dtypes back into an ibis schema.

If we keep a generic set of unsigned sqlalchemy types as defined here currently, the `dtype.register` calls here could actually be moved into `ibis.backends.base.sql.alchemy.datatypes` and generic for all dialects, since the mapping here would be consistent for all supporting backends. No need to specialize on duckdb here.

ibis/expr/rules.py
```diff
@@ -275,22 +275,22 @@ def output_shape(self):
...
-def _promote_numeric_binop(exprs, op):
-    bounds, dtypes = [], []
-    for arg in exprs:
-        dtypes.append(arg.type())
-        if hasattr(arg.op(), 'value'):
-            # arg.op() is a literal
-            bounds.append([arg.op().value])
-        else:
-            bounds.append(arg.type().bounds)
+def _promote_integral_binop(exprs, op):
```
jcrist, line 278: To support unsigned addition, I had to modify the promotion algorithm here a bit. I don't believe there should be any change in behavior for signed integer/literal addition, but it now works for (some) cases of unsigned integers.

Note that this promotion rule assumes that the backing compute engine promotes on overflow rather than wraps or saturates. This means that `uint32 + 1` always promotes to `uint64`, even if all values in the column would still fit in a `uint32`. It also means that `uint64 - 1` currently errors, since there's no ibis type that can fit `[-1, uint64.max - 1]`. `uint32 - 1` also currently errors, but could be made to work (I guess it should be an `int64`). I'm not really sure what the expected behavior should be here, suggestions welcome.

ibis/backends/duckdb/compiler.py
```diff
@@ -1,3 +1,8 @@
+import duckdb_engine
+from sqlalchemy.ext.compiler import compiles
+
+import ibis.backends.base.sql.alchemy.datatypes as dt
+import ibis.expr.datatypes as ibis_dt
```
cpcloud, line 5:
```suggestion
import ibis.backends.base.sql.alchemy.datatypes as sat
import ibis.expr.datatypes as dt
```

I believe this is the pattern we use when both libraries' type systems are in the mix.

ibis/backends/duckdb/compiler.py
```diff
@@ -16,6 +21,46 @@ class DuckDBSQLExprTranslator(AlchemyExprTranslator):
     _has_reduction_filter_syntax = True
 
 
+@compiles(dt.UInt64, "duckdb")
```
cpcloud, line 24:
```suggestion
@compiles(sat.UInt64, "duckdb")
```

ibis/backends/duckdb/compiler.py
```diff
@@ -16,6 +21,46 @@ class DuckDBSQLExprTranslator(AlchemyExprTranslator):
...
+def compile_uint64(element, compiler, **kw):
+    return "UBIGINT"
+
+
+@compiles(dt.UInt32, "duckdb")
```
cpcloud, line 29:
```suggestion
@compiles(sat.UInt32, "duckdb")
```

ibis/backends/duckdb/compiler.py
```diff
@@ -16,6 +21,46 @@ class DuckDBSQLExprTranslator(AlchemyExprTranslator):
...
+def compile_uint32(element, compiler, **kw):
+    return "UINTEGER"
+
+
+@compiles(dt.UInt16, "duckdb")
```
cpcloud, line 34:
```suggestion
@compiles(sat.UInt16, "duckdb")
```

ibis/backends/duckdb/compiler.py
```diff
@@ -16,6 +21,46 @@ class DuckDBSQLExprTranslator(AlchemyExprTranslator):
...
+def compile_uint16(element, compiler, **kw):
+    return "USMALLINT"
+
+
+@compiles(dt.UInt8, "duckdb")
```
cpcloud, line 39:
```suggestion
@compiles(sat.UInt8, "duckdb")
```

ibis/backends/duckdb/compiler.py
cpcloud: Nice. The `@compiles` decorator is pretty neat.

ibis/backends/duckdb/compiler.py
cpcloud, line 61:
```suggestion
@dt.dtype.register(duckdb_engine.Dialect, sat.UInt64)
def dtype_uint64(_, satype, nullable=True):
    return dt.UInt64(nullable=nullable)


@dt.dtype.register(duckdb_engine.Dialect, sat.UInt32)
def dtype_uint32(_, satype, nullable=True):
    return dt.UInt32(nullable=nullable)


@dt.dtype.register(duckdb_engine.Dialect, sat.UInt16)
def dtype_uint16(_, satype, nullable=True):
    return dt.UInt16(nullable=nullable)


@dt.dtype.register(duckdb_engine.Dialect, sat.UInt8)
def dtype_uint8(_, satype, nullable=True):
    return dt.UInt8(nullable=nullable)
```

ibis/backends/base/sql/alchemy/datatypes.py
gforsyth: It would mean going from the general case here to a more duckdb-specialized one, but I wonder if we should upstream this bit into duckdb_engine?

ibis/expr/rules.py
cpcloud: I think the result type of `+/- 1` depends on what the inferred type of `1` is. If we infer `1` as any kind of signed type then I think we have to promote to the largest signed integer type `int64`. The only case where that might be different is in the case of two literal operands and I think that would complicate the promotion logic quite a bit since we'd have to look not only at the value but also at the operation type.

Looking at julia and numpy, both promote to `int64` in the case of `uint32(0) +/- 1`:

numpy:

```
In [3]: import numpy as np

In [4]: type(np.uint32(0) - 1)
Out[4]: numpy.int64

In [5]: type(np.uint32(0) + 1)
Out[5]: numpy.int64
```

Julia:
```
julia> typeof(UInt32(0) - 1)
Int64

julia> typeof(UInt32(0) + 1)
Int64
```

In the case where `1` is an unsigned integer then both numpy and Julia wrap.

In the database world things are a bit dicey:

DuckDB:

```
D select typeof(cast(0 as ubigint) - 1) as x;
┌────────┐
│   x    │
├────────┤
│ DOUBLE │
└────────┘
```

It looks like MySQL supports `UNSIGNED` in DDL, but I'm not sure if there's a syntax for using these types at runtime, for example, in a `CAST` expression. The naive approach doesn't seem to work:

```
MariaDB [ibis_testing]> select cast(1 as int unsigned);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'unsigned)' at line 1
```

All this is to say I think the behavior should be:

* `unsigned ± signed -> int64`
* `unsignedX ± unsignedY -> max(unsignedX, unsignedY)` (this may or may not error depending on the system, DuckDB errors if the result is out of range for type, pandas/numpy wrap)

cpcloud (changes requested):
Couple of changes in imports needed.

Also added some thoughts on the promotion behavior.

ibis/backends/duckdb/compiler.py
cpcloud: That seems like a good idea.

ibis/backends/base/sql/alchemy/datatypes.py
cpcloud: It would be useful in duckdb_engine regardless of what we decided to do here I think, but I'm not sure we should necessarily take that on in any form other than to make [an upstream issue in duckdb_engine](https://github.com/Mause/duckdb_engine/issues/329)

ibis/expr/rules.py
jcrist: > Looking at julia and numpy, both promote to int64 in the case of uint32(0) +/- 1:

NumPy's promotion rules differ depending on if the item is an array or a scalar. In the case of unsigned-array + python-int, the dtype stays the same and addition wraps.

```python
In [8]: (np.uint32(1) + 1).dtype
Out[8]: dtype('int64')

In [9]: (np.array([1], dtype=np.uint32) + 1).dtype
Out[9]: dtype('uint32')
```

Since duckdb tries to do exact math when possible (promoting to a larger type on overflow), trying to emulate that intuitively in ibis would be nice.

So I guess I agree with your behavior described, except that I think in the case of adding with a positive python integer, the signedness shouldn't change, so `uint64 + 1` is still a `uint64`, but `uint64 - 1` would be an `int64`, as would `uint64 + ibis.literal(1, "int64")` (since that's an explicitly typed integer).

cpcloud: That sounds reasonable to me.

cpcloud:
@jcrist I fixed the buglet around inference of integral binops, so in theory more of the test suite should pass now!

cpcloud (approved):

jcrist:
Thanks for finishing this up Phillip, this looks good to me :).

cpcloud (approved):

PR #4267, opened 2022-07-22
feat: add ADBC backend

lidavidm:
This adds the skeleton of a very minimal ADBC backend. Most tests
do not yet pass. The main thing is that ADBC is really a database
API abstraction akin to PEP 249, JDBC, or ODBC, so this backend
eventually needs to be able to let the user specify the SQL
dialect to compile to (or: to use Substrait instead).

Also, I haven't yet updated the dependencies, CI, etc. This is 
mostly to prove that it (generally) is possible.

ADBC doesn't yet have packages/releases, so getting this up and
working takes some work. Basically:

1. Add cython to shell.nix
1. Get the ADBC C SQLite driver built (I did this outside of Nix): https://github.com/apache/arrow-adbc/blob/main/CONTRIBUTING.md#cc
1. Inside the Nix dev environment, get the Python package built: https://github.com/apache/arrow-adbc/blob/main/CONTRIBUTING.md#python (this only worked if I made a virtualenv inside the Nix shell)
1. `export LD_LIBRARY_PATH=path/to/folder/with/libadbc_driver_sqlite.so`
1. `export PYTHONPATH=$PYTHONPATH:path/to/folder/with/adbc_driver_manager/package`
1. `python -m pytest -m adbc`

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -35,8 +35,8 @@ jobs:
           - "3.8"
           - "3.10"
         backend:
-          - name: adbc
-            title: ADBC
+          - name: adbc_sqlite
+            title: ADBC (SQLite driver)
           - name: dask
             title: Dask
           - name: duckdb
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -39,7 +39,7 @@ jobs:
             title: ADBC
             sys-deps:
               - cmake
-              - ninja
+              - ninja-build
           - name: dask
             title: Dask
           - name: duckdb
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -25,6 +25,9 @@ jobs:
   test_backends:
     name: ${{ matrix.backend.title }} ${{ matrix.os }} python-${{ matrix.python-version }}
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        working-directory: ibis
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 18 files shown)


lidavidm:
On Phillip's suggestion, now the ADBC package [implements](https://github.com/apache/arrow-adbc/pull/56) DBAPI, and this backend just defers to SQLAlchemy (except for fetching results, which gets overridden). I also added a bit of pytest magic here (thanks to Phillip again for explaining how it works).

Current status:

```
91 failed, 260 passed, 2 skipped, 8681 deselected, 192 xfailed in 50.52s
```

Most of the remaining failures are because of arithmetic UDFs, which need manual skips. The driver has no way of registering Python UDFs; perhaps the driver could optionally register all the ones we need here, though. Part of the pytest magic is to auto-skip anything already skipped for SQLite.

Next up, I'd like to try reintegrating the Substrait backend

lidavidm:
What I ended up doing was futzing around so that the driver can also be distributed as a package: https://github.com/apache/arrow-adbc/pull/57

Once that gets updated, then instead of having to build the driver and such here, we can just depend on the (prerelease) packages.

kszucs:
ADBC should be able to be used by various backends so it's not a backend on its own. 

I think ibis will need another abstraction layer which is responsible for the interfacing between various databases.

lidavidm:
I guess then we need to decide what specific backend(s) to target here? (Since this won't be as featureful as the existing SQLite driver, without being able to register UDFs.) Possibly Postgres and Acero?

lidavidm:
In that case though, instead of trying to add all the pytest marks to get everything to pass, I'll focus on looking more into Substrait integration

cpcloud:
@lidavidm Should we keep this PR open?

lidavidm:
Sorry, I've been letting this sit. I'll close it for now, and I'll follow up with a new PR once I get (nightly) wheel releases for the ADBC packages set up; that'll make development easier.

PR #4271, opened 2022-07-23
fix(impala): A delimited table explicitly declare stored as textfile

### LINKED ISSUE #4260: bug(impala) A delimited table should be explicitly stored as textfile
https://github.com/ibis-project/ibis/blob/88ffe3367cb6a34936e578f6a9b68dc30d559507/ibis/backends/impala/ddl.py#L67
when the cluster's default format is set as parquet, this will cause an exception. It should be explicitly stored as textfile. such as

```python
        if self.lineterminator is not None:
            yield f"LINES TERMINATED BY '{self.lineterminator}'"

        yield 'STORED AS TEXTFILE'
        yield f"LOCATION '{self.path}'"
```

yi-you:
fixes #4260


cpcloud (approved):

cpcloud:
Thanks!

PR #4272, opened 2022-07-23
fix(pandas): use localized UTC time for `now` operation

cpcloud:
This PR unifies the result value of `now` between pandas and dask and the SQL backends to be a localized UTC value


kszucs:
@cpcloud I think we should switch to the `4.x.x.` as the base branch.

kszucs (approved):

kszucs:
@cpcloud impala doesn't have a translation rule for strftime

kszucs:
I removed the two commits from the `4.x.x` because we need to fix the impala failures first.

Issue #4268, opened 2022-07-22
Pandas Anti-Join does not behave as expected


cwisecarver:
Hi ibis-project maintainers!

This is such a great project and I've been trying it out for a new project I'm prototyping. In testing it I've been using pandas and duckdb. I tried to use an anti-join to filter a set of events created by "bad actors" from a much larger set of events. My expectation of an anti-join was that rows where all three columns in the events table matched all three columns in the bad actors table would be removed from the events table. That was the behavior using the duckdb backend. The pandas backend appears to have removed all rows from the events table where **any** of the columns of the bad actors set matched **any** of the columns of the events table. Below is some example code that exhibits this behavior.

```python
import ibis
import pandas as pd

df = pd.DataFrame.from_records([
    {
        "domain": "parsely.com",
        "id_one": "AAAAAA",
        "id_two": "BBBBBB",
    },
    {
        "domain": "parsely.com",
        "id_one": "CCCCCC",
        "id_two": "BBBBBB",
    },
    {
        "domain": "parse.ly",
        "id_one": "AAAAAA",
        "id_two": "BBBBBB",
    },
    {
        "domain": "parse.ly",
        "id_one": "CCCCCC",
        "id_two": "BBBBBB",
    },
    {
        "domain": "blog.parsely.com",
        "id_one": "DDDDDD",
        "id_two": "EEEEEE",
    },
])
bad_actors_df = pd.DataFrame.from_records([
    {
        "domain": "parsely.com",
        "id_one": "AAAAAA",
        "id_two": "BBBBBB",
    },
    {
        "domain": "parse.ly",
        "id_one": "CCCCCC",
        "id_two": "BBBBBB",
    }
])


pandas_conn = ibis.pandas.connect({"events": df, "bad_actors": bad_actors_df})
events_table = pandas_conn.table("events")
bad_actors_table = pandas_conn.table("bad_actors")

# exclude bad actors events from table
pandas_expression = events_table.anti_join(bad_actors_table, ["domain", "id_two", "id_one"])
pandas_compiled_expression = ibis.pandas.compile(pandas_expression)
pandas_result = pandas_compiled_expression.execute()
print(pandas_result)


duck_conn = ibis.duckdb.connect(path=":memory:", read_only=False)
duck_conn.con.execute("CREATE TABLE events AS SELECT * FROM df")
events_duck_table = duck_conn.table("events")
[...]

PR #4278, opened 2022-07-25
fix(pandas): Fix anti-joins

cwisecarver:
This PR fixes support for anti-joins in pandas. Currently a pandas anti-join will filter out a row if *any* of the columns match versus all of the columns matching. This changes the behavior to only remove rows where all columns match. This matches the behavior of the duckdb backend.

Happy to work with maintainers to get this merged. I tried to install Nix on my M1 mac to run the unit tests but lychee wouldn't compile on macOS even under rosetta2.

Potentially fixes: https://github.com/ibis-project/ibis/issues/4268


ibis/backends/tests/test_join.py
```diff
@@ -17,7 +17,9 @@ def _pandas_semi_join(left, right, on, **_):
 def _pandas_anti_join(left, right, on, **_):
     assert len(on) == 1, str(on)
     inner = pd.merge(left, right, how="outer", indicator=True, on=on)
-    return inner["_merge"].apply(lambda x: True if x == "both" else False).astype(bool)
+    return inner["_merge"].apply(
+        lambda x: True if x == "both" else False
```
cpcloud, line 21: Would `inner["_merge"] == "both"` work here?

cwisecarver: That worked in my local test case that is included in the linked issue but failed CI with a different set of errors. Seems like progress at least.

cwisecarver: I think that change causes the length of the list to change (which caused the different errors) whereas mine just replaced everything with something. I'll try and find a more pandas-way to get the same result. I'm waiting for nix to finish installing on a linux box so I can hopefully run the tests locally.

cpcloud: FWIW you can use nix, conda or pip to develop with ibis.

cpcloud:
The CI errors look unrelated to whether you use `apply`, and are related to a new warning from pandas about `sort_values` becoming keyword-only in the future.

cwisecarver:
closing in favor of https://github.com/ibis-project/ibis/pull/4302

PR #4279, opened 2022-07-25
Pin Mariadb to 10.8.2

thornycrackers:
When trying to run `docker-compose up` the `mysql` container would never come up. The logs mentioned "Can't initialize timers" and the first result that came up was https://github.com/MariaDB/mariadb-docker/issues/434 which has this comment https://github.com/MariaDB/mariadb-docker/issues/434#issuecomment-1135889824 which mentions pinning to `10.8.2`. After pinning to that version everything seemed to work fine.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -89,7 +89,7 @@ services:
         - mysqladmin
         - ping
       timeout: 5s
-    image: mariadb:10.8
+    image: mariadb:10.8.2
     ports:
       - 3306:3306
     networks:


gforsyth:
Hey @thornycrackers -- thanks for the PR!  I can't reproduce the issue with `10.8.3` on my end.  I don't see a problem with pinning to 10.8.2 until whatever is going on is fixed, though.

What version of `docker` are you using?

cpcloud:
@thornycrackers Thanks for the PR!

Can you adjust the commit message to `ci: pin mariadb to 10.8.2`? After that this'll be good to merge!

thornycrackers:
Updated the commit message! The version of docker I'm using is:
```
Docker version 20.10.9, build v20.10.9
```
I know it's a bit old but my NixOS flake input is set to `nixpkgs/nixos-21.11`

cpcloud:
It might be related to your nixos version, but it's fine for us to pin to 10.8.2 for now. At what point should we try to bump it again?

cpcloud (approved):

thornycrackers:
Ehhhh, I don't know what your goals for the "dev" experience are. From my experience it's pretty easy to update docker so it's not unreasonable to say it's a requirement. Maybe after the next release just unpin it and say that old docker versions aren't supported for dev.

cpcloud:
To be honest we haven't had a lot of people show up with different docker versions. I think many contributors don't run the backend tests because it can seem like a big lift.

We've spent some time over the last year trying to reduce the number of steps required to run the backend test suite. In particular @anjakefala recently landed a wonderful PR that removed the requirement to load data manually before running tests.

We don't have any explicit goals for the dev experience beyond:

1. wide support for environment tooling
2. a small number of commands needed to run tests, right now it's `docker compose up --wait && pytest -m some_backend`

We don't have a strong requirement on which mariadb version to run against, so your suggestion sounds fine to me.

PR #4285, opened 2022-07-26
docs: add how-to section to docs

jcrist:
Also moves topk example to the how-to section.


docs/how_to/topk.md
```diff
@@ -0,0 +1,35 @@
+# Top-K Filtering
```
anjakefala, line 1: Should the title for this also include the simplified phrasing (e.g. "How to compute the top 5 records"?) From "Top-K Filtering" on its own, I was not sure what the article would cover.

jcrist: Good point! This was just copy-pasted from before.

I'm slightly against starting each of these with "How to ...", since it be redundant with the subsection, and be repetitive in the table of contents.

What about "Compute the Top K records"?

cpcloud:
@jcrist Ping! If you want to change the title that'd be swell and then we can merge this!

cpcloud (approved):

PR #4288, opened 2022-07-27
feat(api/backend): flesh out and fix Intersection and Difference operations

### LINKED ISSUE #4284: tbl1.intersect(tbl2) fails in MySQL
I'm trying to do a simple intersect of two tables using the MySQL backend in Ibis 3.0.2 (with Python 3.9), but I'm getting an error.

```
conn.raw_sql('CREATE TABLE t(a int, b varchar(30))')
conn.raw_sql('CREATE TABLE s(a int, b varchar(30))')
conn.raw_sql('INSERT INTO t VALUES(1, "x"), (2, "y"), (3, "z")')
conn.raw_sql('INSERT INTO s VALUES(1, "x"), (2, "y"), (4, "q")')

t = conn.table('t')
s = conn.table('s')

expr = t.intersect(s)

expr.compile()
```

Here is the traceback. The issue is that each of the table expressions is still an `sqlalchemy.sql.selectable.Select` object rather than a string. Either they should be converted to strings to make the `join` work, or the wrong compile is happening which would create an SQLAlchemy object to do the intersect.

```
/tmp/ipykernel_2235005/1443667768.py in <module>
----> 1 expr.compile()

~/.pyenv/versions/3.9.4/lib/python3.9/site-packages/ibis/expr/types/core.py in compile(self, limit, timecontext, params)
    291             Mapping of scalar parameter expressions to value
    292         """
--> 293         return self._find_backend().compile(
    294             self, limit=limit, timecontext=timecontext, params=params
    295         )

~/.pyenv/versions/3.9.4/lib/python3.9/site-packages/ibis/backends/base/sql/__init__.py in compile(self, expr, limit, params, timecontext)
    221             backend.
    222         """
--> 223         return self.compiler.to_ast_ensure_limit(
    224             expr, limit, params=params
    225         ).compile()

~/.pyenv/versions/3.9.4/lib/python3.9/site-packages/ibis/backends/base/sql/compiler/base.py in compile(self)
     40     def compile(self):
     41         compiled_setup_queries = [q.compile() for q in self.setup_queries]
---> 42         compiled_queries = [q.compile() for q in self.queries]
     43         compiled_teardown_queries = [
     44             q.compile() for q in self.teardown_queries

~/.pyenv/versions/3.9.4/lib/python3.9/site-packages/ibis/backends/base/sql/compiler/base.py in <listcomp>(.0)
     40     def compile(self):
     41         compiled_setup_queries = [q.compile() for q in self.setup_queries]
---> 42         compiled_queries = [q.compile() for q in self.queries]
     43         compiled_teardown_queries = [
     44             q.compile() for q in self.teardown_queries

~/.pyenv/versions/3.9.4/lib/python3.9/site-packages/ibis/backends/base/sql/compiler/base.py in compile(self)
    106             )
    107         )
--> 108         return '\n'.join(buf)

TypeError: sequence item 0: expected str instance, Select found
```

cpcloud:
This PR implements `Intersection` and `Difference` for the remaining SQL backends.

It also adds a `distinct` flag similar to the `.union()` API. Notably the
default value is set to `True` to preserve existing behavior. This is different
from `union`'s' default value of `distinct` which is `False`.

Fixes #4284.


gforsyth (approved):
This looks good to me!  

The pyspark `subtract:exceptAll` vs `except_:except_all` seems designed to confuse, but I'm sure it arose organically.

cpcloud:
> This looks good to me!
> 
> The pyspark `subtract:exceptAll` vs `except_:except_all` seems designed to confuse, but I'm sure it arose organically.

I was definitely searching the docs for a bit before I finally broke down and Googled.

PR #4292, opened 2022-07-28
refactor(raw_sql): Remove unused `results` kwarg

gforsyth:
This is _technically_ a breaking change but I am not marking it as
breaking because I don't imagine anyone is actually using the `results`
kwarg, and because it does nothing.


jcrist (approved):
:+1:

cpcloud (approved):
Nice.

PR #4293, opened 2022-07-29
feat(api): add `FileTable`

lidavidm:
This is intended to model a "table" that is actually a collection
of files (local or remote), which is more common in things that
look like "query engines" (e.g. Substrait, Acero, Pandas).

Substrait integration is the specific purpose here, but the
definition here is much simplified compared to Substrait's. In
particular: globs are not modeled, and all files are assumed
to be of the same type.


lidavidm:
Would it be useful to implement this in a backend (e.g. DuckDB)? Effectively it would differ by not defining a view, instead inlining the table definition into the final query (so not really a difference to the user).

ibis/expr/operations/relations.py
```diff
@@ -84,6 +85,57 @@ def change_name(self, new_name):
...
+    """
+    A table sourced from a files on a local/remote filesystem.
+
+    More common in systems shaped like 'query engines' (e.g. Pandas,
+    Acero, DuckDB).
+    """
+
+    items = rlz.tuple_of(uri)
+    schema = rlz.instance_of(sch.Schema)
+    file_format = rlz.instance_of(FileFormat)
```
kszucs, line 136: Do we plan to extend the formats with additional descriptors? Otherwise shall we use an enum instead?

ibis/expr/operations/relations.py
```diff
@@ -84,6 +85,57 @@ def change_name(self, new_name):
         return type(self)(new_name, self.args[1], self.source)
 
 
+@rlz.validator
+def uri(arg, **kwargs):
```
kszucs, line 89: This should go to `ibis.expr.rules.py`

ibis/expr/operations/relations.py
lidavidm: Hmm, Parquet might want to be able to support some options (dictionary-encoding?). If/when we add CSV (maybe I should add that now?), we'd also want options for things like delimiters and headers, at the very least.

ibis/tests/expr/test_operations.py
```diff
@@ -114,6 +114,37 @@ def output_type(self):
...
+    [
+        (ops.ArrowFileFormat, 'arrow'),
+        (ops.OrcFileFormat, 'orc'),
+        (ops.ParquetFileFormat, 'parquet'),
+    ],
+)
+def test_file_table(format_class, extension):
+    node = ops.FileTable(
+        items=[
+            f's3://foo/2009/01/data.{extension}',
```
kszucs, line 128: We should expose this to the ibis api, e.g.:

```python
ibis.table(list_of_paths, schema)
```

though we may want to choose a different function name, not sure about the ergonomics.

cc @cpcloud

kszucs:
> Would it be useful to implement this in a backend (e.g. DuckDB)? Effectively it would differ by not defining a view, instead inlining the table definition into the final query (so not really a difference to the user).

We should definitely try to integrate it with at least one of the file based backends to see how well would it fit, one good candidate is duckdb but datafusion and pandas should support these operations as well.

ibis/tests/expr/test_operations.py
cpcloud: I think `ibis.table` should remain the sole entry point for creating `UnboundTable`-based expressions. I'm not sure `FileTable`s should have their own distinct API beyond possibly getting them back from a call to `con.table(...)`.

Thoughts?

kszucs: I think there is an advantage of describing datasets on the operations level making them backend agnostic. How could we define and run queries agains a set of s3 files with different engines otherwise?

cpcloud: Can you write down a small expression-creation-to-execution example? It will help drive the conversation.

jcrist: Since this is for a collection of files for a specific format, I'd assume this operation would be what's returned from an e.g. `read_parquet` call (if we were to add that). I think I agree with Phillip that `ibis.table` would feel a bit weird to handle this use case.

```python
table = con.read_parquet(list_of_paths_or_maybe_a_glob, **extra_options_here_perhaps)
```

cpcloud: We do have some precedent for `con.parquet_file` in the impala backend and a few others like it `delimited_file`.

These APIs were never fully baked outside of the impala backend. Perhaps we should consider fully baking them, even if they turn out to look a bit different, for example different arguments or even making them plural (`parquet_files`, `text_files`, etc).

lidavidm: I added a quick Pandas backend so you can do `ibis.pandas.parquet_files([...]).execute()`

kszucs: > Can you write down a small expression-creation-to-execution example? It will help drive the conversation.

Will do once I have the time.

jayceslesar: what if the schema is not consistent between files?

lidavidm: @jayceslesar currently it would raise an error at runtime. I would expect the backends might deal with it in different ways: they could try to unify schemas or simply error.

ibis/backends/pandas/__init__.py
```diff
@@ -148,6 +150,53 @@ def create_table(self, table_name, obj=None, schema=None):
...
+            # describe S3 paths, etc.
+            path = Path(raw_path).expanduser().resolve()
+            if not path.is_file():
+                raise ValueError(
+                    f"'{raw_path}' does not exist or is not a file!"
+                )
+
+            if not schema:
+                # It'd be nice to be able to assume PyArrow here to avoid
+                # reading the entire file
```
cpcloud, line 185: If you add `pyarrow` as a dependency of the `pandas` extra in `pyproject.toml` then you can make this assumption.

ibis/expr/operations/relations.py
```diff
@@ -84,6 +84,48 @@ def change_name(self, new_name):
         return type(self)(new_name, self.args[1], self.source)
 
 
+@public
+class FileFormat(Node):
```
cpcloud, line 88: I'm not sure we want these to be `Node`s vs `Annotable`s. @kszucs Thoughts?

cpcloud (commented):
Lookin' good. Couple of comments.

ibis/backends/pandas/client.py
```diff
@@ -299,3 +300,7 @@ class PandasTable(ops.DatabaseTable):
 
 class PandasDatabase(Database):
     pass
+
+
+class PandasFileTable(ops.FileTable):
```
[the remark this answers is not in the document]
lidavidm: I was facing some issue along the lines of not being able to infer the backend for a node

ibis/expr/operations/relations.py
kszucs: I'm really in favor of them being Node instances. This would generalize the table definitions for the file based backends.

```python

dataset = ibis.dataset("s3://bucket/<paritioning-info>/<glob-pattern>.parquet")  # or we could also overload ibis.table()
```

The underlying operation would be a `FileTable` with `PaquetFormat` and perhaps additional partitioning directives like `HivePartitioning` (this is similar how pyarrow datasets are defined). Then we could feed these tables to pyspark, datafusion, pyarrow, duckdb, clickhouse and possibly to other databases as well. 

Currently we need to define the source tables and the schemas for each individual backend using custom implementations in `ibis.<backend>.connect()`. IMO those should be moved to be expression level constructs.

kszucs: So instead of:

```python
client = ibis.pandas.connect(dictionary={'table': pd.read_csv(...)})
t = client['table']
```

it would be:

```python
t = ibis.dataset(pd.read_csv(...), schema=...)
# t.op() => ops.PandasTable extending ops.InMemoryTable
t = ibis.dataset(data_directory / 'table.csv', schema=...)
# t.op() => ops.FileTable 

expr = t['field'].sum()

ibis.pandas.execute(expr)
```

kszucs: I misunderstood your question, the format classes could indeed be Annotable subclasses rather than Node subclasses since these are not Operations.

ibis/backends/pandas/client.py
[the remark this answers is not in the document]
lidavidm: I'll just move `source` to `FileTable` and remove the extra subclass

ibis/expr/operations/relations.py
lidavidm: Thanks, I'll update this

PR #4304, opened 2022-08-01
fix(pandas): fix `contains` with a filter inside of `group_concat`

### LINKED ISSUE #4300: bug(pandas): pandas backend does not support aggregation with list-based where condition
Pandas backend throws a `NotImplementedError` when evaluating an ibis expression that contains a `groupby()` with an `aggregate()` that depends on a list-based `where` condition using `isin()`.

MWE

```
import ibis
import pandas as pd

test_df = pd.DataFrame({
    'id': ['1', '1', '2', '3', '4', '4'],
    'value': ['a', 'a', 'b', 'c', 'd', 'a']
})

conn = ibis.pandas.connect({'sample_df': test_df})
table = conn.table('sample_df')

expr1 = table.groupby("id").aggregate([
    table.id.count(table.value == "a")
])

expr2 = table.groupby("id").aggregate([
    table.id.count(table.value.isin(["a", "b"]))
])

#this works
print(expr1.execute())

#this fails
print(expr2.execute())
```

The specific error yielded is `NotImplementedError: Could not find signature for execute_node: <Contains, SeriesGroupBy, list>`

cpcloud:
Fixes #4300.

### DIFF
--- a/ibis/backends/pandas/execution/strings.py
+++ b/ibis/backends/pandas/execution/strings.py
@@ -237,6 +237,8 @@ def execute_group_concat_series_gb_mask(
     op, data, sep, mask, aggcontext=None, **kwargs
 ):
     def method(series, sep=sep):
+        if series.empty:
+            return pd.NA
         return sep.join(series.values.astype(str))
 
     return aggcontext.agg(
--- a/ibis/backends/pandas/execution/generic.py
+++ b/ibis/backends/pandas/execution/generic.py
@@ -983,6 +983,17 @@ def execute_node_contains_series_sequence(op, data, elements, **kwargs):
 
 
 @execute_node.register(
+    ops.Contains,
+    SeriesGroupBy,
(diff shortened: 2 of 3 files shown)


ibis/backends/pandas/execution/strings.py
```diff
@@ -236,15 +236,16 @@ def execute_group_concat_series_gb(
 def execute_group_concat_series_gb_mask(
     op, data, sep, mask, aggcontext=None, **kwargs
 ):
-    def method(series, sep=sep):
-        return sep.join(series.values.astype(str))
+    def method(series, mask, sep=sep):
+        if mask.any():
```
cpcloud, line 240: I think we can simply check if the series is empty here instead of incurring the overhead of looking at every `mask` element.

cpcloud: Done

PR #4306, opened 2022-08-02
fix(duckdb): ensure that in-memory connections remain in their creating thread

### LINKED ISSUE #4282: bug(duckdb): newly created tables in an in-memory connection are not accessible
```
In [11]: con = ibis.duckdb.connect()

In [12]: con.raw_sql("CREATE TABLE t (x int)")
Out[12]: <sqlalchemy.engine.cursor.LegacyCursorResult at 0x7fe3bfa82620>                         

In [13]: con.list_tables()
Out[13]: []
```

cpcloud:
Backwards compatible fix for #4282.

This is fixed upstream in https://github.com/Mause/duckdb_engine/issues/338,
but we can address it in ibis without requiring users to upgrade `duckdb_engine`.

This PR implements the same fix in ibis.

Fixes #4282.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -111,12 +111,13 @@ class Backend(BaseAlchemyBackend):
         read_only
             Whether the database is read-only
         """
-        if path != ":memory:":
+        if not (in_memory := path == ":memory:"):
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
looks good!

PR #4318, opened 2022-08-05
docs: revert docs deps and document Schema public API

cpcloud:
- revert: chore(dev-deps): remove constraints from transitive dev dependencies
- chore(dev-deps): bound `docstring_parser`
- docs(schema): fill out Schema public API
- docs: unnest API docs
- docs: add `asc` function to top level API docs
- docs: fix example docstrings

This PR fixes the API documentation rendering which seems to be broken if
`docstring_parser>=0.14.1`. To avoid doing a ton of work tracking down the
exact breakage, I've pinned that dependency for now.

I've also added docstrings to the public API of `Schema`, and cleaned up a bit
here and there.

Finally, looking at the API documentation, I thought there was too much
nesting. Currently there's a `module > class > methods > actual thing I care
about` hierarchy. I've removed the module top-level, and everything now starts
at the class level.

I'd like to avoid reviewing docstring content on pre-existing docstrings.

Review of the new `Schema` API docs is welcomed!


ibis/expr/schema.py
```diff
@@ -166,20 +302,72 @@ def name_at_position(self, i):
...
-
-        Notes
-        -----
-        Mutates `df`
+        DataFrame
+            Type-converted DataFrame
+
+        Examples
+        --------
+        Import then necessary modules
```
gforsyth, line 322:
```suggestion
        Import the necessary modules
```

gforsyth (changes requested):
small nit: typo

ibis/expr/schema.py
cpcloud: Ok, now I actually fixed it :)

PR #4323, opened 2022-08-08
feat(api): add `ibis.memtable` API to support in-memory table expressions

### LINKED ISSUE #3832: feat: create a table of literal values
I'd like to be able to create a table from literal values. This can be handy for small utility tables that I might like to join to. It's also handy for test scripts to avoid having to create physical tables when not needed.

One possible way to implement this would be with the `UNNEST` operator (https://github.com/ibis-project/ibis/issues/1146), though there may be other ways of accomplishing this.

**Expected Ibis usage**

With `UNNEST` support, I'd expect it to be used in combination with an [array literal](https://ibis-project.org/docs/3.0.1/api/expressions/collections/#ibis.expr.types.arrays.array) or [struct literals](https://ibis-project.org/docs/3.0.1/api/expressions/collections/#ibis.expr.types.structs.struct).

```python
table = ibis.unnest(
  ibis.array(
    [
      ibis.struct(dict(col1='hello', col2=1), type='struct<col1: string, col2: integer>'),
      ibis.struct(dict(col1='world', col2=2), type='struct<col1: string, col2: integer>'),
      ibis.struct(dict(col1='!', col2=3), type='struct<col1: string, col2: integer>'),
    ]
  )
)
```

**Equivalent BigQuery SQL**

Using the [UNNEST operator](https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays#flattening_arrays) ([reference](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#unnest_operator)) on an [array literal](https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays#constructing_arrays) of [struct literals](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type) results in an anonymous table-valued expression.

```sql
SELECT col1, col2
FROM UNNEST(
  [
    STRUCT<col1 STRING, col2 INT64>("hello", 1),
    STRUCT<col1 STRING, col2 INT64>("world", 2),
    STRUCT<col1 STRING, col2 INT64>("!", 3)
  ]
)
-- Optionally, include a table name via the alias feature
AS my_temp_table
```

Row | col1 | col2
-- | -- | --
1 | hello | 1
2 | world | 2
3 | ! | 3

**Alternative Ibis expression**

Alternatively, maybe [ibis.table](https://ibis-project.org/docs/3.0.1/api/expressions/top_level/#ibis.expr.api.table) could take an optional `values` parameter? In the BigQuery dialect, I'd still expect this to compile to an UNNEST operator.

```python
table = ibis.table(
  values=[
      ibis.struct(dict(col1='hello', col2=1), type='struct<col1: string, col2: integer>'),
      ibis.struct(dict(col1='world', col2=2), type='struct<col1: string, col2: integer>'),
      ibis.struct(dict(col1='!', col2=3), type='struct<col1: string, col2: integer>'),
  ],
)
```

**See also**

* [feat: support for UNNEST (array flatten operation) #1146](https://github.com/ibis-project/ibis/issues/1146)
* [Creating a table from literal values in Ibis (StackOverflow)](https://stackoverflow.com/questions/71886813/creating-a-table-from-literal-values-in-ibis)

cpcloud:
This PR adds support for an API to work with in-memory tables: `ibis.memtable`.

`ibis.memtable` returns a `Table` expression.

The motivation is to allow use of simple in memory datasets in expressions.

`ibis.memtable` defers to pandas to for type inference which makes it suitable
for handling the case of inserting into/creating tables from various kinds of
Python objects.

#### API Stuff

1. Whatever the `pd.DataFrame` constructor supports can be passed to `memtable`.
1. `ibis.memtable` has both a `columns` argument and a `schema` argument. If
   `schema` is `None`, the function performs type inference on the entire table.
1. I'm not entirely sure where the row count performance falls over.
1. When both `columns` and `schema` are `None` and the row type is not a
   mapping the default naming scheme is `col{i}` where i is the zero-indexed
   ith column number. This template can be changed using an option
   `rows_default_column_name_template`.

#### Backend Specific Stuff

1. ClickHouse doesn't seem to support the notion of `VALUES`. It might be
   posible to implement it by `arrayJoin`-ing an array of `Tuple` types. I left
   that as a follow up. Even better would be to support `ibis.memtable`
   constructs using `external_tables`. This will probably have to wait until
   4.0.
1. MySQL and SQLite don't support `SELECT`-ing from a `VALUES` statement
   directly, because SQLAlchemy compiles these incorrectly to a form that these
   two dialects of SQL don't support. Expressions based on `ibis.rows` won't
   work those backends. OTOH, `con.create_table` works and is tested for both.
1. PySpark translation is implemented.
1. Pandas and Dask are not implemented, these are in theory straightforward.

cc @tswast

Closes #3832.

### DIFF
--- a/ibis/backends/pandas/client.py
+++ b/ibis/backends/pandas/client.py
@@ -9,8 +9,11 @@ from pandas.api.types import CategoricalDtype, DatetimeTZDtype
 
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
+import ibis.expr.rules as rlz
 import ibis.expr.schema as sch
+from ibis import util
 from ibis.backends.base import Database
+from ibis.common.grounds import Immutable
 
 infer_pandas_dtype = pd.api.types.infer_dtype
 
@@ -296,6 +299,28 @@ def convert_array_to_series(in_dtype, out_dtype, column):
 
 dt.DataType.to_pandas = ibis_dtype_to_pandas  # type: ignore
 sch.Schema.to_pandas = ibis_schema_to_pandas  # type: ignore
+
+
+class DataFrameProxy(Immutable):
+    __slots__ = ('_df', '_hash')
+
+    def __init__(self, df):
+        object.__setattr__(self, "_df", df)
+        object.__setattr__(self, "_hash", hash((type(df), id(df))))
+
+    def __getattr__(self, name):
+        return getattr(self._df, name)
+
+    def __hash__(self):
+        return self._hash
+
+    def __repr__(self):
+        df_repr = util.indent(repr(self._df), spaces=2)
+        return f"{self.__class__.__name__}:\n{df_repr}"
+
+
+class PandasInMemoryTable(ops.InMemoryTable):
+    data = rlz.instance_of(DataFrameProxy)
 
 
 class PandasTable(ops.DatabaseTable):
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -558,3 +558,78 @@ def test_invalid_connect():
 def test_deprecated_path_argument(backend, tmp_path):
     with pytest.warns(UserWarning, match="The `path` argument is deprecated"):
         getattr(ibis, backend.name()).connect(path=str(tmp_path / "test.db"))
+
+
+@pytest.mark.parametrize(
+    ("expr", "expected"),
+    [
+        param(
+            ibis.memtable([(1, 2.0, "3")], columns=list("abc")),
+            pd.DataFrame([(1, 2.0, "3")], columns=list("abc")),
+            id="simple",
+        ),
+        param(
+            ibis.memtable([(1, 2.0, "3")]),
+            pd.DataFrame([(1, 2.0, "3")], columns=["col0", "col1", "col2"]),
+            id="simple_auto_named",
+        ),
+        param(
+            ibis.memtable(
+                [(1, 2.0, "3")],
+                schema=ibis.schema(dict(a="int8", b="float32", c="string")),
+            ),
+            pd.DataFrame([(1, 2.0, "3")], columns=list("abc")).astype(
+                {"a": "int8", "b": "float32"}
+            ),
+            id="simple_schema",
+        ),
+        param(
+            ibis.memtable(
+                pd.DataFrame({"a": [1], "b": [2.0], "c": ["3"]}).astype(
+                    {"a": "int8", "b": "float32"}
+                )
+            ),
+            pd.DataFrame([(1, 2.0, "3")], columns=list("abc")).astype(
+                {"a": "int8", "b": "float32"}
+            ),
+            id="dataframe",
+        ),
+    ],
+)
+@pytest.mark.notyet(
+    ["clickhouse"],
+    reason="ClickHouse doesn't support a VALUES construct",
+)
+@pytest.mark.notyet(
+    ["mysql", "sqlite"],
+    reason="SQLAlchemy generates incorrect code for `VALUES` projections.",
+    raises=(sa.exc.ProgrammingError, sa.exc.OperationalError),
+)
+@pytest.mark.notimpl(["dask", "datafusion", "pandas"])
+def test_in_memory_table(backend, con, expr, expected):
+    result = con.execute(expr)
+    backend.assert_frame_equal(result, expected)
+
+
+@pytest.mark.parametrize(
+    "t",
+    [
+        param(
+            ibis.memtable([("a", 1.0)], columns=["a", "b"]),
+            id="python",
+        ),
+        param(
+            ibis.memtable(pd.DataFrame([("a", 1.0)], columns=["a", "b"])),
+            id="pandas",
+        ),
+    ],
+)
+@pytest.mark.notimpl(["clickhouse", "dask", "datafusion", "pandas"])
+def test_create_from_in_memory_table(con, t):
+    tmp_name = guid()
+    con.create_table(tmp_name, t)
+    try:
+        assert tmp_name in con.list_tables()
+    finally:
+        con.drop_table(tmp_name)
+        assert tmp_name not in con.list_tables()
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -188,6 +188,7 @@ __all__ = (
     'least',
     'literal',
     'map',
+    'memtable',
     'NA',
     'negate',
     'now',
@@ -335,10 +336,10 @@ def schema(
 
 
 def table(
-    schema: SupportsSchema,
+    schema: SupportsSchema | None = None,
     name: str | None = None,
 ) -> ir.Table:
-    """Create an unbound table for building expressions without data.
+    """Create a table literal or an abstract table without data.
 
     Parameters
     ----------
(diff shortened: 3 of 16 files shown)


ibis/expr/api.py
```diff
@@ -330,13 +335,45 @@ def table(
         A schema for the table
     name
         Name for the table. One is generated if this value is `None`.
+    data
+        Table literal data.
+    columns
+        A sequence of [`str`][str] column names.
```
jcrist, line 341: Perhaps expand the docstring for these parameters a bit, or provide an example of usage? From the docstring alone I can't tell how this is intended to be used.

jcrist (commented):
Just a few small fixups/questions, otherwise this looks good to me.

ibis/expr/api.py
cpcloud: I'll add an example.

ibis/config.py
```diff
@@ -37,6 +37,27 @@ class SQL(BaseModel):
...
+class Rows(BaseModel):
+    """ibis.Rows-related options."""
+
+    column_name_template: str = Field(
+        default="col{:d}",
+        description="Format string template to use for unnamed columns in `ibis.rows`.",  # noqa: E501
+    )
+    display: RowsDisplay = Field(
+        default=RowsDisplay(), description=RowsDisplay.__doc__
+    )
```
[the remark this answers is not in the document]
cpcloud: I'm just going to kill these options altogether and document the `col{x}` behavior in the docstring.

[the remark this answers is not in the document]
cpcloud: Done.

ibis/backends/tests/test_client.py
```diff
@@ -378,3 +380,57 @@ def test_unsigned_integer_type(alchemy_con, coltype):
...
+    raises=(sa.exc.ProgrammingError, sa.exc.OperationalError),
+)
+@pytest.mark.notimpl(["dask", "datafusion", "pandas"])
+def test_rows(backend, con, expr, expected):
+    result = con.execute(expr)
+    backend.assert_frame_equal(result, expected)
+
+
+@pytest.mark.notimpl(["clickhouse", "dask", "datafusion", "pandas"])
+def test_create_from_rows(con):
```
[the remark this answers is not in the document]
cpcloud: Done.

ibis/backends/tests/test_client.py
```diff
@@ -378,3 +380,57 @@ def test_unsigned_integer_type(alchemy_con, coltype):
...
+    ["clickhouse"],
+    reason="ClickHouse doesn't support a VALUES construct",
+)
+@pytest.mark.notyet(
+    ["mysql", "sqlite"],
+    reason="SQLAlchemy generates incorrect code for `VALUES` projections.",
+    raises=(sa.exc.ProgrammingError, sa.exc.OperationalError),
+)
+@pytest.mark.notimpl(["dask", "datafusion", "pandas"])
+def test_rows(backend, con, expr, expected):
```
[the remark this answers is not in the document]
cpcloud: Done.

ibis/expr/format.py
```diff
@@ -430,6 +432,25 @@ def _fmt_table_op_limit(op: ops.Limit, *, aliases: Aliases, **_: Any) -> str:
     return f"{op.__class__.__name__}[{', '.join(params)}]"
 
 
+@fmt_table_op.register
+def _fmt_table_op_rows(
```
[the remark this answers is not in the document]
cpcloud: Done.

ibis/tests/expr/test_table.py
```diff
@@ -1526,3 +1526,15 @@ def test_exprs_to_select():
     with pytest.warns(FutureWarning, match="Passing `exprs`"):
         result = t.select(exprs=exprs)
     assert result.equals(t.select(len=t.a.length()))
+
+
+def test_rows_ambiguous():
```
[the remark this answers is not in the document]
cpcloud: Done.

ibis/expr/api.py
cpcloud: Done.

ibis/expr/operations/relations.py
```diff
@@ -96,6 +96,12 @@ def blocks(self):
         return True
 
 
+@public
+class TableLiteral(TableNode, sch.HasSchema):
```
kszucs, line 100: How is table literal different from a literal with datatype `dt.Array(dt.Struct(...))`?

ibis/expr/rules.py
```diff
@@ -450,6 +452,62 @@ def python_literal(value, arg, **kwargs):
     return arg
 
 
+@functools.singledispatch
+def _make_hashable(value) -> Hashable:
```
kszucs, line 456: I feel like this should be done by `dt._normalize`

ibis/expr/operations/relations.py
cpcloud: In principle nothing, but in practice there are issues dealing with nullable, the APIs are different, and many backends supports `VALUES` and do not support arrays or structs.

cpcloud: Semantically there's also the issue that a relation and an array of structs are distinct entities.

ibis/expr/api.py
```diff
@@ -318,25 +318,96 @@ def schema(
...
+
+    Examples
+    --------
+    Create a table with no data backing it
+
+    >>> t = ibis.table(schema=dict(a="int", b="string"))
+    >>> t
+    UnboundTable: unbound_table_0
+      a int64
+      b string
```
jcrist, line 356: Since some users are already confused by unbound tables, perhaps we shouldn't highlight this calling pattern in an example? Up to you.

ibis/expr/operations/relations.py
kszucs: I'd rather like to see a bigger picture here providing a pluggable `ibis.table()` API, something like:

```python
@public
class TableNode(Node):
    def sort_by(self, sort_exprs):
        return Selection(self, [], sort_keys=sort_exprs)

    @property
    @abstractmethod
    def schema(self) -> sch.Schema:
        """Return a schema."""

    def to_expr(self):
        import ibis.expr.types as ir

        return ir.Table(self)


@public
class PhysicalTable(TableNode, Named):
    pass


@public
class UnboundTable(PhysicalTable):
    schema = rlz.instance_of(sch.Schema)
    name = rlz.optional(rlz.instance_of(str), default=genname)


@public
class DatabaseTable(PhysicalTable):
    name = rlz.instance_of(str)
    schema = rlz.instance_of(sch.Schema)
    source = rlz.client

    def change_name(self, new_name):
        return type(self)(new_name, self.args[1], self.source)


class InMemoryTable(PhysicalTable):
    pass


# implemented in the pandas backend
class PandasTable(InMemoryTable):
    data = rlz.hashable_wrapper_of(pd.DataFrame)

    @property
    def name(self):
        return self.data.name


# supported by duckdb, clickhouse, datafusion, acero
class S3Dataset(PhysicalTable):
    name = rlz.instance_of(str)
    pattern = rlz.instance_of(str)
    format = rlz.isin({"csv", "parquet", "json"})  # or whatever
```

Then a backend could register its own datatype as in `InMemoryTable`

```python
@ibis.table.register(pd.DataFrame):
def table_from_pandas(df, schema):
    # oversimplified impl for demonstration purposes
    return PandasTable(df, schema=schema)
```

ibis/expr/rules.py
kszucs: I'd prefer to offload the data schema inference to more performant libraries, in this case directly to pandas since it is a mandatory dependency for now.

```python
@ibis.table.register(typing.Sequence)
def table_from_sequence(data, schema):
    # something like
    return pd.DataFrame(data, dtypes=ibis_schema_to_pandas(schema)).to_lists()  # something similar
```

ibis/expr/operations/relations.py
kszucs: Another relevant operation is FileTable https://github.com/ibis-project/ibis/pull/4293

cpcloud: I don't think `FileTable` is related to in memory tables. They are independent types of table scans.

kszucs: I mean they are related on the abstraction level, so if we are about to support in memory tables then we should consider how can we add support for additional table types.

ibis/expr/rules.py
cpcloud: This is done now!

ibis/expr/api.py
cpcloud: Got rid of these.

ibis/expr/rules.py
```diff
@@ -487,6 +493,62 @@ def python_literal(value, arg, **kwargs):
...
+    return tuple(sequence)
+
+
+@_to_tuple.register(collections.abc.Mapping)
+def _to_tuple_mapping(mapping):
+    return tuple(mapping.values())
+
+
+@validator
+def table_data(data, **kwargs):
```
kszucs, line 548: Are these stil required?

ibis/expr/rules.py
```diff
@@ -487,6 +493,62 @@ def python_literal(value, arg, **kwargs):
     return arg
 
 
+@functools.singledispatch
+def _make_hashable(value) -> Hashable:
```
kszucs, line 497: Do we still need `_make_hashable`?

cpcloud: Nope, I thought I got rid of it. Doing it now.

ibis/expr/rules.py
cpcloud: Nope, deleting.

ibis/backends/pandas/client.py
```diff
@@ -298,6 +299,28 @@ def convert_array_to_series(in_dtype, out_dtype, column):
 sch.Schema.to_pandas = ibis_schema_to_pandas  # type: ignore
 
 
+# ImmutableProxy?
+class DataFrameProxy:
```
kszucs, line 303: It should subclass `Immutable` in the `4.x.x` branch.

kszucs: Ahh, we already have immutable here. Could you please subclass it and remove the comment above?

cpcloud: Yep

kszucs (approved):
Just a nit otherwise it looks good to me!

ibis/backends/pandas/client.py
cpcloud: Done.

PR #4325, opened 2022-08-08
fix(backends): only recommend installing ibis-foo when foo is a known backend

anjakefala:
- list for KNOWN_BACKENDS was compiled by referencing the backends in
  here that were not already part of ibis-framework:
  https://ibis-project.org/docs/3.1.0/backends/

### DIFF
--- a/ibis/__init__.py
+++ b/ibis/__init__.py
@@ -20,6 +20,8 @@ __all__ += api.__all__
 
 __version__ = "3.1.0"
 
+_KNOWN_BACKENDS = ['bigquery', 'heavyai']
+
 
 def _get_backend_entrypoints() -> list[_importlib_metadata.EntryPoint]:
     """Get the list of installed `ibis.backend` entrypoints"""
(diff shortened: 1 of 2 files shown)


ibis/__init__.py
```diff
@@ -20,6 +20,8 @@
 
 __version__ = "3.1.0"
 
+KNOWN_BACKENDS = ['bigquery', 'heavyai']
```
gforsyth, line 23: There's `mssql` and `omniscidb` (former name for `heavyai`) although I'm not sure if those backends are up-to-date with ibis 3.x

cpcloud: They likely aren't but I think that's okay.

cpcloud: Is it worth hiding this as `_KNOWN_BACKENDS`?

gforsyth: ok, it looks like `ibis-omniscidb` hasn't been updates, so skip that one.  checking on `ibis-mssql`

jcrist: I'd say so, we don't want to export this as public API.

cpcloud (approved):
LGTM. Just a question about whether we want to make the attribute private through `_KNOWN_BACKENDS`. No strong opinion either way.

ibis/__init__.py
gforsyth: yeah, and `ibis-mssql` hasn't been updated in 18 months, ok, ignore this!

ibis/tests/test_api.py
```diff
@@ -26,7 +26,7 @@ def test_backends_tab_completion():
 
 
 def test_missing_backend():
-    msg = "If you are trying to access the 'foo' backend"
+    msg = "module 'ibis' has no attribute 'foo'."
```
cpcloud, line 29: Funny enough this passes because `match` accepts a regular expression and `.` of course matches any single character which includes both the literal period as well as the space that's actually there.

This is fine. We can address in a follow-up if needed.

cpcloud:
Thanks @anjakefala!

PR #4330, opened 2022-08-09
fix(decimal): add decimal type inference

### LINKED ISSUE #4319: bug: decimal type evolution during queries
During my [work](https://github.com/iree-org/iree-llvm-sandbox/pull/551) on representing decimals in relational queries, I have come across behavior where ibis deviates from the SQL standard (as described [here](https://www.ibm.com/docs/en/i/7.1?topic=operators-decimal-arithmetic-in-sql)).

```python
import ibis

table = ibis.table(schema=[("a", "decimal(4, 2)"), ("b", "decimal(5, 3)")],
                   name="t")

res = table.projection((table.a * table.b).name('c'))

print(res.c.type())
```

This code executed with ibis v3.0.0 shows that the type of c is `decimal(prec=5, scale=3)`, whereas I would argue that according to the SQL standard it should be `decimal(prec=9, scale=5)`. Is this a WIP or a deliberate design decision?

webmiche:
closes #4319 

This PR addresses the type evolution for decimal types as specified [here](https://www.ibm.com/docs/en/i/7.1?topic=operators-decimal-arithmetic-in-sql). As I am relatively new to the ibis development environment, I was not sure where/how to add a testcase for this, so I just used the following code:

```python
import ibis

table = ibis.table(
    schema=[("a", "decimal(4, 2)"), ("b", "decimal(5, 3)")], name="t"
)

res = table.projection((table.a + table.b).name('c'))

print(res.c.type())

res = table.projection((table.a * table.b).name('c'))

print(res.c.type())
```

Currently there are two main shortcomings of this PR that I see, both are hard for me to fix as I do not know ibis in-depth.

1. It only handles cases where there are two args given. I cannot see a case where this is not a given as it only works on sub, mul and add, but feel free to point out if I am missing something.
2. The maximum precision and maximum scale are currently not handled at all. This is because both seem to be specifiable in some way, something I don't know where it should/could happen in the ibis infrastructure and how to implement it properly. I am very open for suggestions.

### DIFF
--- a/ibis/expr/rules.py
+++ b/ibis/expr/rules.py
@@ -1,5 +1,6 @@
 import enum
 import functools
+import operator
 from itertools import product, starmap
 
 import ibis.common.exceptions as com
@@ -294,12 +295,48 @@ def _promote_integral_binop(exprs, op):
     return dt.highest_precedence(dtypes)
 
 
+def _promote_decimal_dtype(args, op):
+
+    if len(args) != 2:
+        return highest_precedence_dtype(args)
+
+    # TODO: Add support for setting the maximum precision and maximum scale
+    lhs_prec = args[0].type().precision
+    lhs_scale = args[0].type().scale
+    rhs_prec = args[1].type().precision
+    rhs_scale = args[1].type().scale
+    max_prec = 31 if lhs_prec <= 31 and rhs_prec <= 31 else 63
+    max_scale = 31
+
+    if op is operator.mul:
+        return dt.Decimal(
+            min(max_prec, lhs_prec + rhs_prec),
+            min(max_scale, lhs_scale + rhs_scale),
+        )
+    if op is operator.add or op is operator.sub:
+        return dt.Decimal(
+            min(
+                max_prec,
+                max(
+                    lhs_prec - lhs_scale,
+                    rhs_prec - rhs_scale,
+                )
+                + max(lhs_scale, rhs_scale)
+                + 1,
+            ),
+            max(lhs_scale, rhs_scale),
+        )
+    return highest_precedence_dtype(args)
+
+
 def numeric_like(name, op):
     @immutable_property
     def output_dtype(self):
         args = getattr(self, name)
         if util.all_of(args, ir.IntegerValue):
             result = _promote_integral_binop(args, op)
+        elif util.all_of(args, ir.DecimalValue):
+            result = _promote_decimal_dtype(args, op)
         else:
             result = highest_precedence_dtype(args)
 
--- a/ibis/tests/expr/test_decimal.py
+++ b/ibis/tests/expr/test_decimal.py
@@ -31,6 +31,37 @@ def test_decimal_sum_type(lineitem):
     assert result.type() == dt.Decimal(38, 2)
 
 
+def test_promote_decimal_type_mul(lineitem):
+    col_1 = lineitem.l_extendedprice
+    col_2 = lineitem.l_discount
+    result = col_1 * col_2
+    assert result.type().precision == 24
+    assert result.type().scale == 4
+
+
+def test_promote_decimal_type_add(lineitem):
+    col_1 = lineitem.l_extendedprice
+    col_2 = lineitem.l_discount
+    result = col_1 + col_2
+    assert result.type().precision == 13
+    assert result.type().scale == 2
+
+
+def test_promote_decimal_type_mod(lineitem):
+    col_1 = lineitem.l_extendedprice
+    col_2 = lineitem.l_discount
+    result = col_1 % col_2
+    assert result.type().precision == 12
+    assert result.type().scale == 2
+
+
+def test_promote_decimal_type_max():
+    t = ibis.table([("a", "decimal(31, 3)"), ("b", "decimal(31, 3)")], "t")
+    result = t.a * t.b
+    assert result.type().precision == 31
+    assert result.type().scale == 6
+
+
 @pytest.mark.parametrize(
     "precision, scale, expected",
     [


ibis/expr/rules.py
```diff
@@ -294,12 +295,35 @@ def _promote_integral_binop(exprs, op):
     return dt.highest_precedence(dtypes)
 
 
+def _infer_decimal_dtype(args, op):
+    # TODO: Add support for the maximum precision and maximum scale
+    # TODO: Currently assumes that there are only two args. Can there be more?
```
cpcloud, line 300: I suppose in the case of `ibis.coalesce(*only_decimal_args)` there are a variable number of decimal arguments, but the rules for computing the precision and scale of such an expression are going to be different than those here.

We don't have any arithmetic operations that are modeled as variadic, but it's an idea we've been considering in an effort to make pattern matching and simple expression optimizations easier to encode. If we did do that then it'd be in 4.0.

ibis/expr/rules.py
```diff
@@ -294,12 +295,35 @@ def _promote_integral_binop(exprs, op):
...
+    if op is operator.mul:
+        return dt.Decimal(
+            args[0].type().precision + args[1].type().precision,
+            args[0].type().scale + args[1].type().scale,
```
cpcloud, line 304: Small style nit: can you lift some of this code into a few variables?

ibis/expr/rules.py
```diff
@@ -294,12 +295,35 @@ def _promote_integral_binop(exprs, op):
...
+
+
 def numeric_like(name, op):
     @immutable_property
     def output_dtype(self):
         args = getattr(self, name)
         if util.all_of(args, ir.IntegerValue):
             result = _promote_integral_binop(args, op)
+        elif util.all_of(args, ir.DecimalValue):
+            result = _infer_decimal_dtype(args, op)
```
cpcloud, line 326: Can you add a test that hits this code path?

This can go in `ibis/tests/expr/test_decimal.py`.

cpcloud (changes requested):
Thanks for the PR!

If you can add a test or two that hit the new code path, that'd be swell!

ibis/expr/rules.py
webmiche: Okay, I see. I changed the code s.t. it checks that there are exactly 2 args.

Assuming that the semantics of a variadic add is something like "sum all", we could probably just use the same functions to compute the precision and scale in a pairwise manner, i.e. take the first two, compute their precision/scale, add the result to the list, take the next two... until the list has exactly one element.

cpcloud: Yeah, that makes sense to me.

ibis/expr/rules.py
```diff
@@ -294,12 +295,41 @@ def _promote_integral_binop(exprs, op):
...
+    lhs_prec = args[0].type().precision
+    lhs_scale = args[0].type().scale
+    rhs_prec = args[1].type().precision
+    rhs_scale = args[1].type().scale
+    if op is operator.mul:
+        return dt.Decimal(
+            lhs_prec + rhs_prec,
+            lhs_scale + rhs_scale,
+        )
+    if op is operator.add or op is operator.sub:
```
cpcloud, line 312: Should I make a follow up issue to address division? It looks like that'll require max precision/scale as well.

webmiche: It seems that the output type for division is computed differently within ibis. For `/` it's *always* float 64, for `//` it's always int64. I was not sure why this works this way, so I left division out for the moment.

webmiche: I think we could implement the thing with max_prec and max_scale within this PR as well, if you prefer. I am just not certain where to put it within ibis.

cpcloud: Ok, that makes sense. I'll make a follow-up issue. Longer term we should probably split out the arithmetic operations into a few subclasses based type category to simplify the type derivation logic. We could then address revision.

cpcloud: #4334 is the follow up issue.

cpcloud:
@webmiche Regarding maximum precision/scale, I think the best place for that is https://github.com/ibis-project/ibis/blob/master/ibis/expr/datatypes/core.py#L443 (the `largest` property).

The existing logic is designed to support systems that allow arbitrary precision (I believe that may just be postgres at the moment).

What happens is that if a decimal type has no specified precision or scale then that is left alone, otherwise the defaults from earlier versions of ibis were left intact.

I said this elsewhere, but we're not tied to these particular choices (except that we'd ideally preserve support for arbitrary precision) since decimals haven't gotten any love for some time now.

ibis/expr/rules.py
```diff
@@ -294,12 +295,41 @@ def _promote_integral_binop(exprs, op):
     return dt.highest_precedence(dtypes)
 
 
+def _infer_decimal_dtype(args, op):
```
kszucs, line 298: Please rename it to `_promote_decimal_dtype`

webmiche:
> @webmiche Regarding maximum precision/scale, I think the best place for that is https://github.com/ibis-project/ibis/blob/master/ibis/expr/datatypes/core.py#L443 (the `largest` property).
> 
> The existing logic is designed to support systems that allow arbitrary precision (I believe that may just be postgres at the moment).
> 
> What happens is that if a decimal type has no specified precision or scale then that is left alone, otherwise the defaults from earlier versions of ibis were left intact.
> 
> I said this elsewhere, but we're not tied to these particular choices (except that we'd ideally preserve support for arbitrary precision) since decimals haven't gotten any love for some time now.

I have now implemented a version, where the max_prec and max_scale cannot be set explicitely, but they are computed within the function. I feel that having them part of each individual datatype class is not quite what IBM suggests in their docu. I understand it as that max_precision and max_scale are fields that you can set in a global sense, so all decimals must adhere to those maxima.

What is also confusing is that IBM only describes the maxima for the case where you have to operands (i.e. in terms of p and p') and not when you only have one single operand.

cpcloud (approved):
Thanks for tackling this!

webmiche:
I feel that we can't actually get the necessary test coverage just yet as the two lines that are not covered are a problem for later, i.e. when the operations can be variadic, which AFAIK is not yet the case. I feel it still makes sense to keep these lines in, so I suggest to merge with this one CI job failing (obviously only when you guys have expected the code fully :grin:).

ingomueller-net:
I looked into max precision a bit. As always with SQL, things are quite chaotic (even *within different version of the same system*):

* DB2 on i has a maximum precision of 63 digits: https://www.ibm.com/docs/en/i/7.5?topic=types-numbers
* DB2 on ZOS has a maximum precision of 31 digits: https://www.ibm.com/docs/en/db2-for-zos/13?topic=numbers-decimal-decimal-numeric
* DB2 further allows to configure modes for decimal arithmetic: https://www.ibm.com/docs/en/db2/11.5?topic=parameters-dec-arithmetic-decimal-arithmetic-mode
* (At least) SQL Server 2016 and later have a maximum precision of 38 digits: https://docs.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql?view=sql-server-2016 (but it says that "earlier versions" only supported up to 28 digits).

Looks quite involved to solve comprehensively...

webmiche:
I am not sure how ibis exactly interacts with the backends, but it seems to be the case that the exact type of a decimal operation does not matter, in that this part seems to have been broken, but all the backends still worked fine.

I think it might make sense to keep this part of type inference relatively simple as it is a bit of a corner case IMO and doesn't seem ot matter much. Or, implement all the different decimal flavors for each backend and choose the right one on the fly. Or, if you can interact with the chosen backend in this way, ask the backend what the type should be...

cpcloud:
We can ignore the failing coverage check in this case.

> Looks quite involved to solve comprehensively...

Indeed. This PR is a huge improvement over what was there before so I think we'll take this PR as is.

> Or, implement all the different decimal flavors for each backend and choose the right one on the fly.

We currently don't have a canonical way of altering expression construction based on backend knowledge. The goal is to be able to construct valid expression cheaply and abstractly (without reference to a backend), which as y'all are pointing out is not truly possible with decimals. I don't think it's worth the effort to what effectively amounts to reimplementing type derivation logic for each backend that ibis supports.

Thinking a bit more broadly, I wonder if we shouldn't get rid of precision and scale entirely and just have a generic `decimal` type. That's something we can tackle later, after this PR is merged.

webmiche:
Judging by the documentation of IBM and that I feel that even though we have a decimal(12, 5) or something, it probably still is encoded into a 32bit integer, I feel that having the exact precision might be more of an optimization opportunity than really a semantic difference (discover that something cannot overflow, so we can save a check). I feel that scale however is more important to have in ibis, as it defines how your data really looks. @ingomueller-net probably has better intuition/more context on that though.

So I feel like precision could probably be dropped, but scale really feels important to me.

cpcloud:
@webmiche Thanks for the thoughts. I agree that keeping scale around will likely lead to a better ux than dropping it entirely.

cpcloud (approved):
Thank you for the PR! Let's continue the collaboration!

PR #4337, opened 2022-08-10
feat(api): promote `psql` to a public API

cpcloud:
This PR promotes `ibis.util.psql` to a public `ibis.psql` API.

It also fixes the fact that we are writing out the generic sqlglot dialect
instead of writing the output in the desired input dialect.


kszucs:
I'm not sure that we should place SQL prettifier utilities under the core `ibis.expr` package. The internals should be about the ibis internal representation nothing else (like the scope and timecontext files should be moved to the pandas backend and the random.py should be in the api.py). 

Could we place it under `ibis.common` perhaps?

ibis/expr/api.py
```diff
@@ -192,6 +193,7 @@
     'param',
     'pi',
     'prevent_rewrite',
+    'psql',
```
kszucs, line 196: How about calling is `pretty_sql`? 

`psql` is a bit cryptic for me, first I associated to postgres.

cpcloud: I want it to be short to type :) Ideally <= 5 characters and no underscores.

cpcloud: I'd settle for `pretty` I guess.

jcrist: Not everything in ibis is sql though, I think sql has to be in the name. How about `tosql`? Or `viewsql`? Idk, don't want to bikeshed here.

cpcloud: How about `prsql`?

kszucs: Or just `sql()`? 

`from ibis.common.pretty import sql`

But we can always alias an import.

cpcloud:
We can certainly move this under `ibis.common`

ibis/expr/api.py
cpcloud: I'm going with `showsql`. `sql` is too generic.

ibis/common/pretty.py
```diff
@@ -0,0 +1,152 @@
...
+import sqlalchemy.dialects.postgresql
+from multipledispatch import Dispatcher
+from sqlalchemy.engine import Dialect
+from sqlalchemy.sql import ClauseElement
+
+import ibis
+import ibis.common.exceptions as com
+import ibis.expr.types as ir
+
+_get_compiled_string = Dispatcher("_get_compiled_string")
```
kszucs, line 17: Could we move this implementation to the base backends?

ibis/common/pretty.py
```diff
@@ -0,0 +1,152 @@
...
+@_get_compiled_string.register(
+    ClauseElement,
+    (sa.dialects.mysql.dialect, sa.dialects.postgresql.dialect),
+)
+def _get_compiled_string_mysql_pg(expr: ClauseElement, _: Dialect) -> str:
+    return str(expr.compile(compile_kwargs=dict(literal_binds=True)))
+
+
+@functools.singledispatch
+def _get_backend_name(expr: ir.Expr | ClauseElement) -> str:
```
kszucs, line 64: The backends should already define their name via `backend.name` class field.

cpcloud: Sure, but this attempts to find the name for both ibis expression **and** sqlalchemy expressions.

ibis/common/pretty.py
cpcloud: It's kind of specific to pretty printing SQL. How about leaving it here until it becomes clear that it would be more broadly useful?

kszucs (approved):
LGTM, thanks @cpcloud!

kszucs:
@cpcloud needs a rebase

PR #4341, opened 2022-08-11
feat(log): make BaseSQLBackend._log print by default

saulpw:


ibis/backends/base/sql/__init__.py
```diff
@@ -200,6 +200,7 @@ def _log(self, sql: str) -> None:
         This method can be implemented by subclasses. The logging happens
         when `ibis.options.verbose` is `True`.
         """
+        print(sql)
```
cpcloud, line 203: You should be able to achieve this by setting `ibis.options.verbose = True`. Does that not work?

cpcloud, line 203:
```suggestion
        util.log(sql)
```

`util.log` will account for the option!

cpcloud (approved):
Sweet, thanks!

PR #4363, opened 2022-08-18
feat(clickhouse): add enum, ipaddr, json, lowcardinality to type parser

saulpw:
Using vdsql to connect to Clickhouse's own [playground](https://clickhouse.com/docs/en/getting-started/playground/) (`vdsql clickhouse://explorer@play.clickhouse.com:9440/\?secure=1`) exposed several types that weren't being handled by the Clickhouse backend type parser.  This PR adds them.

### DIFF
--- a/ibis/backends/clickhouse/datatypes.py
+++ b/ibis/backends/clickhouse/datatypes.py
@@ -151,6 +151,32 @@ def parse(text: str) -> DataType:
             nullable=False,
         )
 
+    @p.generate
+    def enum_value():
+        yield dt.SPACES
+        key = yield dt.RAW_STRING
+        yield dt.spaceless_string('=')
+        value = yield p.digit.at_least(1).concat()
+        return (key, int(value))
+
+    @p.generate
+    def lowcardinality():
+        yield dt.spaceless_string('LowCardinality')
+        yield dt.LPAREN
+        r = yield ty
+        yield dt.RPAREN
+        return r
+
+    @p.generate
+    def enum():
+        yield dt.spaceless_string('enum')
+        enumsz = yield p.digit.at_least(1).concat()
+        enumsz = int(enumsz)
+        yield dt.LPAREN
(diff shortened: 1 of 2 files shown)


ibis/backends/clickhouse/datatypes.py
```diff
@@ -151,6 +152,44 @@ def struct():
...
+        if enumsz == 8:
+            return dt.Int8(nullable=False)
+        else:
+            return dt.Int64(nullable=False)
+
+    @p.generate
+    def other():
+        r = yield p.regex(r'.*')
+        warnings.warn(f'Unknown Clickhouse type "{r}"')
+        return dt.String(nullable=True)
```
cpcloud, line 188: I think this has limited utility and will make whatever columns are inferred here unusable in any subsequent queries.

Ibis may happily plod along with a string type, but the code it generates is extremely unlikely to work once you query the database.

Types in ibis determine the API, so whatever expression is a `String`-but-not-really will have APIs like `lpad` and `substr` which likely do not make any sense for the type that couldn't be inferred.

ClickHouse does no implicit casting, and since the user is presented with APIs that aren't valid there's not much reason to "keep going" only to fail later.

cpcloud (changes requested):
Can you add some test cases to `ibis/backends/clickhouse/tests/test_types.py`?

I also think we should avoid defaulting to inferring a `String` type for unknown types, see my comment next to the code for the rationale.

ibis/backends/clickhouse/datatypes.py
saulpw: Yeah, I came across this with the Enum type too, which doesn't work for even simple comparisons (i8 != string).  I'll remove and let it fail as before.

cpcloud (approved):
LGTM! I can fix up the lint errors and merge on green.

cpcloud:
This is fine. I'm fixing the failing job in #4367.

PR #4364, opened 2022-08-18
feat(api): add `argmin` and `argmax` APIs

cpcloud:
This PR adds support for `argmin` and `argmax` methods on column expressions. This was motivated by trying to reproduce [Scrooge McDuck](https://pdet-blog.github.io/2022/08/16/scrooge.html)

### DIFF
--- a/ibis/backends/base/sql/alchemy/translator.py
+++ b/ibis/backends/base/sql/alchemy/translator.py
@@ -56,30 +56,35 @@ class AlchemyExprTranslator(ExprTranslator):
     def get_sqla_type(self, data_type):
         return to_sqla_type(data_type, type_map=self._type_map)
 
-    def _reduction(self, sa_func, expr):
-        op = expr.op()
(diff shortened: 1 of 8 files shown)


ibis/backends/base/sql/alchemy/translator.py
```diff
@@ -69,17 +69,24 @@ def _reduction(self, sa_func, expr):
         ):
             arg = arg.cast(dt.Int32(nullable=type.nullable))
 
+        argtuple = tuple(
+            arg
+            for argname in op.argnames
+            if argname != "where"
+            if isinstance(arg := getattr(op, argname), ir.Expr)
```
kszucs, line 76: `arg for name, arg in zip(op.argnames, op.args) if isinstance(arg, ir.Expr) and name != "where"`

kszucs:
@cpcloud could you please squash it?

PR #4368, opened 2022-08-19
feat(mysql): register BLOB types

saulpw:
Missing BLOB types found while accessing `mysql://rfamro@mysql-rfam-public.ebi.ac.uk:4497/Rfam`.


cpcloud (commented):
Can you add a small test for this?

cpcloud:
Actually, this looks good. I'll add a test in a new PR in a bit.

cpcloud (approved):

PR #4370, opened 2022-08-20
refactor(common): move `Parameter` and `Signature` to validators.py

kszucs:
also make the Signature object validatable

BREAKING CHANGE: use ibis.common.validators.{Patameter, Signature} instead

### DIFF
--- a/ibis/common/validators.py
+++ b/ibis/common/validators.py
@@ -61,6 +61,61 @@ class Curried(toolz.curry, Validator):
         )
 
 
+class Parameter(inspect.Parameter):
+    """
+    Augmented Parameter class to additionally hold a validator object.
+    """
+
+    __slots__ = ('_validator',)
+
+    def __init__(
+        self,
+        name,
+        kind=inspect.Parameter.POSITIONAL_OR_KEYWORD,
+        *,
+        validator=EMPTY,
+    ):
+        super().__init__(
+            name,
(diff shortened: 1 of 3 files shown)


kszucs:
Perhaps we should move it to `ibis.common.signature` instead?

ibis/common/validators.py
```diff
@@ -61,6 +61,61 @@ def __repr__(self):
...
+            return arg
+        else:
+            return self._validator(arg, this=this)
+
+
+class Signature(inspect.Signature):
+    """
+    Validatable signature.
+
+    Primarly used in the implementation of ibis.ground.Annotable.
```
cpcloud, line 100:
```suggestion
    Primarly used in the implementation of ibis.common.grounds.Annotable.
```

ibis/common/validators.py
```diff
@@ -61,6 +61,61 @@ def __repr__(self):
...
+    """
+
+    def validate(self, *args, **kwargs):
+        bound = self.bind(*args, **kwargs)
+        bound.apply_defaults()
+
+        # bound the signature to the passed arguments and apply the validators
```
cpcloud, line 107:
```suggestion
        # bind the signature to the passed arguments and apply the validators
```

cpcloud (approved):
LGTM, couple of small fixes if you want

PR #4374, opened 2022-08-20
refactor(ir): move `random()` to api.py

kszucs:


cpcloud:
If you stick `from ibis.expr.api import random` into `ibis/expr/random.py` then this can target `master` I think.

kszucs:
> If you stick `from ibis.expr.api import random` into `ibis/expr/random.py` then this can target `master` I think.

I'd prefer to remove the file entirely.

cpcloud:
SGTM, slated for 4.0 then!

cpcloud (approved):

PR #4375, opened 2022-08-20
ci: only run benchmarks against push to `master` branch

cpcloud:
This PR should reduce the amount of time to merge a PR by reducing the duration
of CI runs, specifically benchmarks.

I've moved the full benchmark suite to run only after a push to master, while
still running a single iteration of each benchmark to make sure they aren't
broken. I also reduced the join size in the multiple joins benchmark because a
single iteration of it takes about 60 seconds.


kszucs (approved):
Nice!

PR #4376, opened 2022-08-20
fix(pandas): use localized UTC time for now operation

kszucs:
Reopened https://github.com/ibis-project/ibis/pull/4272


kszucs:
@cpcloud need to fix the impala tests before merging

cpcloud:
@kszucs Merging!

PR #4377, opened 2022-08-20
fix(api): enable calling API functions with deferred objects

### LINKED ISSUE #4365: bug(api): make `date` work with deferred `_` api
Currently you can't use `date(_.foo)` to make a deferred expression in `ibis`. Right now you have to use a `lambda`:

```python
t.mutate(year=lambda _: ibis.date(_.date_column).year())
```

We should make `date(_)` work so you can instead do:

```python
t.mutate(year=ibis.date(_.date_column).year())
```

cpcloud:
Enable calling API functions with `_` (`ibis.expr.deferred.Deferred` instances).

Fixes #4365.

### DIFF
--- a/ibis/backends/clickhouse/tests/test_select.py
+++ b/ibis/backends/clickhouse/tests/test_select.py
@@ -221,7 +221,7 @@ def test_isnull_case_expr_rewrite_failure(db, alltypes):
 
     result = ibis.clickhouse.compile(reduction)
     expected = """\
-SELECT sum(CASE WHEN isNull(`string_col`) THEN 1 ELSE 0 END) AS `sum`
+SELECT sum(if(isNull(`string_col`), 1, 0)) AS `sum`
 FROM {0}.`functional_alltypes`"""
     assert result == expected.format(db.name)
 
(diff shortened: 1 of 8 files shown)


kszucs:
@cpcloud shouldn't we enable this from the rules, so the value methods will gain support for deferred objects as well?

cpcloud:
@kszucs The first approach I took was to try to integrate it into the `value` rule, I couldn't figure out a way to do it. Rules are "eager" in their validation, will construct operations with deferred instances. I think we'd ultimately have to add support for operations to hold deferred objects and that just doesn't seem worth the churn given the limited number of APIs that are functions.

cpcloud:
@kszucs can you give this another review?

kszucs (approved):
LGTM

kszucs:
> @kszucs The first approach I took was to try to integrate it into the `value` rule, I couldn't figure out a way to do it. Rules are "eager" in their validation, will construct operations with deferred instances. I think we'd ultimately have to add support for operations to hold deferred objects and that just doesn't seem worth the churn given the limited number of APIs that are functions.

Created a follow-up about trying to integrate it into the rules system https://github.com/ibis-project/ibis/issues/4382

PR #4383, opened 2022-08-22
refactor(ir): resolve circular imports so operations can be globally imported for types

kszucs:
depends on #4135


cpcloud (approved):
LGTM. Wonderful to delete just under 600 LOC!

Issue #4385, opened 2022-08-22
docs: Create and Populate How-To Guides


p-a-a-a-trick:
# Create+Populate How-To Guides

There is a lack of how to guides on the website. Let's make a list of them and start populating a useful section for code schematics.

I hope to keep this issue up as a rolling list of How to Guides, so if you stumble on this and your need isn't included in the docs or this list, then comment your request here!

## v3.2

- [x] How to compute the top k records
- [x] How to register tables in a DuckDB backend [#4386](https://github.com/ibis-project/ibis/pull/4386)
- [x] How to ffill and bfill using window functions ([source](https://github.com/ibis-project/ibis/wiki/ffill-and-bfill-using-window-functions)) [#4402](https://github.com/ibis-project/ibis/pull/4402)

## v4.0
- [x] How to join a table expression to a pandas dataframe ([merged](https://github.com/ibis-project/ibis/pull/4323)) ([docs PR](https://github.com/ibis-project/ibis/pull/4679))
- [x] (WIP) How to chain ibis expressions (`_` API) (#4914)
- [x] How to [sessionize](https://knowledge.dataiku.com/latest/courses/advanced-code/python/sessionization.html) a log of events based on the elapsed time between successive transactions for a give user identifier (#4843).
- [ ] How to convert column structures into columns (`unpack`)

## v4.1+
- [ ] How to create a table with a user-defined schema and efficiently load a large CSV file into it (handling common annoying things like the choice of the separator and custom date / time formats).
- [ ] How to efficiently sample random rows from a large table (or expr) to quickly compute approximate descriptive statistics such as quantiles.
- [ ] How to perform an `asof_join` ([source](https://github.com/ibis-project/ibis/wiki/%60asof_join%60-in-SQL))
- [ ] How to interpolate dates (and/or other sequences)
- [ ] [maybe?] How to use Ibis to generate and save Substrait/SQL
- [ ] How to create a backend (incl. a section on SQLAlchemy backends specifically, but should focus on min requirements to get one up and running w/o SQLAlch)
- [ ] How to chain joins (unknown if this should even be a how-to guide, but if [#4309](https://github.com/ibis-project/ibis/issues/4309) is a fact of life then we should add this)

Let me know if you have any ideas and I'll add them to the list.

cpcloud:
I the things in the wiki are useful bits of information to put in the how-tos.

ogrisel:
Here are some more ideas:

[...]

PR #4386, opened 2022-08-22
docs: add how-to for ibis+duckdb register

gforsyth:
Adding a how-to for using `register` with the duckdb backend

xref #4385

### DIFF
--- a/docs/how_to/duckdb_register.md
+++ b/docs/how_to/duckdb_register.md
@@ -1,0 +1,112 @@
+# Use `register` to load external data files with the DuckDB backend
+
+<!-- prettier-ignore-start -->
+Here we use the [`register`][ibis.backends.duckdb.Backend.register] method to load external data files and join them.
+<!-- prettier-ignore-end -->
+
+We're going to download one month of [NYC Taxi
+data](https://www1.nyc.gov/site/tlc/about/tlc-trip-record-data.page) in
+`parquet` format and also download the "Taxi Zone Lookup Table" which is a `csv`
+
+https://d37ci6vzurychx.cloudfront.net/trip-data/green_tripdata_2022-01.parquet
+https://d37ci6vzurychx.cloudfront.net/misc/taxi+_zone_lookup.csv
+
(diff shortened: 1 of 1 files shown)


docs/how_to/duckdb_register.md
```diff
@@ -0,0 +1,111 @@
...
+
+```python
+>>> con.register("taxi+_zone_lookup.csv")
+AlchemyTable: taxi+_zone_lookup
+  LocationID   int32
+  Borough      string
+  Zone         string
+  service_zone string
+
+>>> con.register("green_tripdata_2022-01.parquet")
```
p-a-a-a-trick, line 33: Do you think showing the use of the optional `table_name` (2nd) argument will show people it's a good way to better control the name of their registered tables? That way if they're iterating through files or want to target specific names for easy reuse they can rather than strictly referencing the table name every time if it's not easily predictable.

e.g. in this case, the hyphen converts to an underscore, which causes the table name to diverge from the file name a bit.

gforsyth: seems reasonable

docs/how_to/duckdb_register.md
```diff
@@ -0,0 +1,112 @@
...
+0  2022-01-01 00:14:21            42          42  Manhattan  Central Harlem North    Boro Zone
+1  2022-01-01 00:20:55           116         116  Manhattan      Hamilton Heights    Boro Zone
+2  2022-01-01 00:57:02            41          41  Manhattan        Central Harlem    Boro Zone
+3  2022-01-01 00:07:42           181         181   Brooklyn            Park Slope    Boro Zone
+4  2022-01-01 00:07:50            33          33   Brooklyn      Brooklyn Heights    Boro Zone
+```
+
+That's it!
+
+Ibis+duckdb currently support registering `parquet`, `csv`, and `csv.gz`.
```
cpcloud, line 104:
```suggestion
Ibis+duckdb currently supports registering `parquet`, `csv`, and `csv.gz`.
```

cpcloud (approved):

PR #4387, opened 2022-08-22
fix(pyspark): make sure `ibis.connect` works with pyspark

cpcloud:
Fixes a bug so that ibis.connect can be used with pyspark.

### DIFF
--- a/ibis/backends/pyspark/__init__.py
+++ b/ibis/backends/pyspark/__init__.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import itertools
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Mapping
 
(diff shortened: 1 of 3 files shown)


ibis/backends/pyspark/__init__.py
```diff
@@ -101,12 +100,15 @@ class Options(ibis.config.BaseModel):
...
         if database := url.database:
-            params.append("spark.sql.warehouse.dir")
-            params.append(str(Path(database).absolute()))
+            params.append(
+                ("spark.sql.warehouse.dir", str(Path(database).absolute()))
+            )
 
-        builder = SparkSession.builder.config(*params)
+        builder = SparkSession.builder
+        for key, value in params:
```
kszucs, line 110: How about directly calling `builder.config()` instead of having an intermediate params container?

kszucs (approved):
Just a nit, otherwise it looks good to me.

ibis/backends/pyspark/__init__.py
cpcloud: You can't do that, `.config` takes a single key value pair. It's meant to be chained.

cpcloud: Ah, but I can use `SparkConf.setAll`!

PR #4390, opened 2022-08-23
fix(backend): ensure that limit chains respect inner limits

cpcloud:
Our current behavior with chained limits is questionable. IMO it qualifies as a
bug even though it appears to have been intentional (see the deleted
comment):

```python
t.limit(5).limit(10)
```

is currently equivalent to

```python
t.limit(10)
```

and

```python
t.limit(10).limit(5)
```

is currently equivalent to

```python
t.limit(5)
```

Both of these behaviors are inconsistent with every other relational operation
we implement in that these chains-of-limits are sensitive to the fact that
there's a specific relational operation underneath. Instead, limits should be
"dumb" and just do whatever the user asks for.

Yes, it may be ultimately pointless to chain a limit in some cases, but we
should strive hard to not second guess the user.

This PR removes the special casing of chained limits.

### DIFF
--- a/ibis/backends/base/sql/alchemy/query_builder.py
+++ b/ibis/backends/base/sql/alchemy/query_builder.py
@@ -330,11 +330,9 @@ class AlchemySelect(Select):
         if self.limit is None:
             return fragment
 
-        n, offset = self.limit['n'], self.limit['offset']
-        fragment = fragment.limit(n)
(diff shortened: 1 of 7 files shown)


cpcloud:
Ah, dask is failing because `.loc` has inclusive upper bounds!

jcrist (approved):
Nice bug find!

PR #4391, opened 2022-08-23
feat: variadic union, intersect, & difference functions

jcrist:
Also makes the corresponding methods variadic.

This makes using `union` to concatenate multiple tables much more ergonomic, since you don't have to manually chain calls to `union` for every pair. For example:

```python
tables = [some_subquery(table, arg) for arg in args]
combined = ibis.union(*tables)
```

I also see value in the top level `intersect` for similar reasons, but it seems less important to me. I only included `difference` for completeness. Unlike `union` and `intersect`, `difference` operations aren't commutative - the first table has different meaning than all subsequent tables. I could see an argument against making `difference` variadic or a top-level function, no strong thoughts here.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -76,6 +76,25 @@ def _regular_join_method(
     return f
 
 
+def _set_op_handle_deprecated_right_kwarg(
+    tables: tuple[Table], **kwargs: Any
+) -> tuple[Table]:
+    """Handle the deprecated `right` argument in union/intersect/difference
+
+    In 4.0 this can be deleted along with the kwargs argument in these methods.
+    """
+    if (right := kwargs.pop("right", None)) is not None:
+        warnings.warn(
+            "The `right` keyword argument has been deprecated and will "
+            "removed in 4.0. Pass tables in as positional arguments",
+            FutureWarning,
+        )
+        tables = (right, *tables)
+    if kwargs:
+        raise TypeError(f"Unexpected keyword arguments: {sorted(kwargs)!r}")
+    return tables
+
+
 @public
 class Table(Expr):
     def __contains__(self, name):
@@ -305,7 +324,9 @@ class Table(Expr):
 
         return ops.SelfReference(self).to_expr()
 
-    def difference(self, *tables: Table, distinct: bool = True) -> Table:
+    def difference(
+        self, *tables: Table, distinct: bool = True, **kwargs
+    ) -> Table:
         """Compute the set difference of multiple table expressions.
 
         The input tables must have identical schemas.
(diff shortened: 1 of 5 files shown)


ibis/expr/types/relations.py
```diff
@@ -305,26 +305,29 @@ def view(self) -> Table:
...
+    def difference(self, *tables: Table, distinct: bool = True) -> Table:
+        """Compute the set difference of multiple table expressions.
 
         The input tables must have identical schemas.
 
         Parameters
         ----------
-        right
-            Table expression
+        *tables
```
jreback, line 315: this is a breaking change?

jreback (changes requested):

ibis/expr/types/relations.py
```diff
@@ -465,49 +468,55 @@ def sort_by(
 
     def union(
         self,
-        right: Table,
+        *tables: Table,
```
cpcloud, line 471: How about `right, *tables` for all of these to ensure that at least one table is present? This would also ensure backwards compatibility.

ibis/expr/types/relations.py
cpcloud: See my comment below, I think we can address this pretty easily.

cpcloud:
While intersect and difference are less common in practice than union, they shouldn't be second class citizens in ibis relative to union.

Second, I don't think if we expose a variadic API for a binary operator that it somehow implies the underlying operation is commutative so +1 on having variadic difference.

ibis/expr/types/relations.py
cpcloud: > a bit of a weird signature to the user

I would argue it's also weird that t.union(*args) may return `t`.

> how do I refer to the tables being joined

`self`, `right`, and `*tables` seems fine to me

> Do we care that a positional argument has been renamed?

I personally don't, but if it's straightforward to keep it backwards compatible then it seems like we should do it.

jcrist: > `self`, `right`, and `*tables` seems fine to me

That still seems awkward to write out when really a top-level function here would want `*tables` alone with a requirement of 2 or more tables. `right` and `*tables` (and self for `union`/`intersect`) are identical in purpose, but calling the args different things makes the user question they have different meanings.

How would you document `union` with that signature?

> I would argue it's also weird that `t.union(*args)` may return `t`.

Good catch, we shouldn't allow that. Technically it'd be fine if `distinct=False`, but `ibis.union(t, distinct=True)` should drop duplicates from `t`, and calling out to a different way of handling that within a `union` call would be weird.

ibis/expr/types/relations.py
jcrist: Ok, we now accept the `right` kwarg still but raise a `FutureWarning`. This should be resolved.

cpcloud (approved):
Awesome. This is *much* more convenient.

PR #4393, opened 2022-08-23
 Ibis design draft: Tabular user-defined tables (UDTs)

rtpsw:
This is a design draft not intended to be committed. The code is part of an Ibis/Substrait/PyArrow prototype.

### DIFF
--- a/ibis/expr/analysis.py
+++ b/ibis/expr/analysis.py
@@ -826,7 +826,7 @@ class Projector:
         )
         self.clean_exprs = list(map(windowize_function, self.resolved_exprs))
 
-    def get_result(self):
+    def get_result(self, **kwargs):
         roots = self.parent_roots
         first_root = roots[0]
 
@@ -835,7 +835,7 @@ class Projector:
             if fused_op is not None:
                 return fused_op
 
-        return ops.Selection(self.parent, self.clean_exprs)
+        return ops.Selection(self.parent, self.clean_exprs, **kwargs)
 
     def try_fusion(self, root):
         assert self.parent.op() == root
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -178,6 +178,7 @@ __all__ = (
     'join',
     'least',
     'literal',
+    'local_table',
     'map',
     'NA',
     'negate',
@@ -321,6 +322,29 @@ def table(schema: sch.Schema, name: str | None = None) -> ir.TableExpr:
         schema = _schema(pairs=schema)
 
     node = ops.UnboundTable(schema, name=name)
+    return node.to_expr()
+
+
+def local_table(schema: sch.Schema, name: str | None = None) -> ir.TableExpr:
+    """Create a local table for build expressions without data.
+
+
+    Parameters
+    ----------
+    schema
+        A schema for the table
+    name
+        Name of local file or files of the table
+
+    Returns
+    -------
+    TableExpr
+        A local table expression
+    """
+    if not isinstance(schema, Schema):
+        schema = _schema(pairs=schema)
+
+    node = ops.LocalTable(schema, name=name)
     return node.to_expr()
 
 
--- a/ibis/expr/operations/numeric.py
+++ b/ibis/expr/operations/numeric.py
@@ -178,6 +178,24 @@ class Clip(ValueOp):
 
 
 @public
+class ClipLower(ValueOp):
+    arg = rlz.strict_numeric
+    lower = rlz.optional(rlz.strict_numeric)
+
+    output_dtype = rlz.dtype_like("arg")
+    output_shape = rlz.shape_like("args")
+
+
+@public
+class ClipUpper(ValueOp):
+    arg = rlz.strict_numeric
+    upper = rlz.optional(rlz.strict_numeric)
+
+    output_dtype = rlz.dtype_like("arg")
+    output_shape = rlz.shape_like("args")
+
+
+@public
 class BaseConvert(ValueOp):
     arg = rlz.one_of([rlz.integer, rlz.string])
     from_base = rlz.integer
--- a/ibis/expr/operations/relations.py
+++ b/ibis/expr/operations/relations.py
@@ -79,6 +79,67 @@ class UnboundTable(PhysicalTable):
 
 
 @public
+class LocalTable(PhysicalTable):
+    schema = rlz.instance_of(sch.Schema)
+    name = rlz.instance_of(str)
+
+
+@public
+class LocalWrite(TableNode, sch.HasSchema):
+    table = rlz.table
+    name = rlz.instance_of(str)
+
+    @cached_property
+    def schema(self):
+        return self.table.schema()
+
+
+@public
+class AsOfMerge(TableNode, sch.HasSchema):
+    tolerance = rlz.instance_of(int)
+    table0 = rlz.optional(rlz.table)
+    key_column0 = rlz.optional(rlz.column_from("table0"))
+    time_column0 = rlz.optional(rlz.column_from("table0"))
+    table1 = rlz.optional(rlz.table)
+    key_column1 = rlz.optional(rlz.column_from("table1"))
+    time_column1 = rlz.optional(rlz.column_from("table1"))
+    table2 = rlz.optional(rlz.table)
+    key_column2 = rlz.optional(rlz.column_from("table2"))
+    time_column2 = rlz.optional(rlz.column_from("table2"))
+    table3 = rlz.optional(rlz.table)
+    key_column3 = rlz.optional(rlz.column_from("table3"))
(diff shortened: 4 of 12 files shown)


rtpsw:
cc @icexelloss, @rok, @cpcloud, @westonpace

ibis/expr/operations/numeric.py
```diff
@@ -177,6 +177,24 @@ class Clip(ValueOp):
...
+class ClipLower(ValueOp):
+    arg = rlz.strict_numeric
+    lower = rlz.optional(rlz.strict_numeric)
+
+    output_dtype = rlz.dtype_like("arg")
+    output_shape = rlz.shape_like("args")
+
+
+@public
+class ClipUpper(ValueOp):
```
cpcloud, line 190: I don't think this should be in ibis core. You can add these nodes in your own codebase and compile them however you'd like.

ibis/expr/types/relations.py
```diff
@@ -666,6 +666,68 @@ def select(
...
+
+        Returns
+        -------
+        TableExpr
+            A table expression with the same schema
+        """
+        node = ops.LocalWrite(table, name=name)
+        return node.to_expr()
+
+    def as_of_merge(
```
cpcloud, line 687: Is there a reason to have `as_of_merge` and `asof_join`? Why doesn't the existing one work for you?

ibis/expr/types/numeric.py
```diff
@@ -112,7 +112,57 @@ def clip(
...
-        return ops.Clip(self, lower, upper).to_expr()
+        if lower is None:
+            op = ops.ClipUpper(self, upper)
+        elif upper is None:
+            op = ops.ClipLower(self, lower)
+        else:
+            op = ops.Clip(self, lower, upper)
+        return op.to_expr()
+
+    def clip_lower(
```
cpcloud, line 123: Is this intentionally a public API? `clip_lower`'s functionality can be captured by inspecting the upper and lower bounds and producing the expression you want to produce.

You can override ibis-substrait compilation of `ops.Clip` and define your own translation rules for `ClipLower` and `ClipUpper` and then this doesn't need to live upstream.

ibis/expr/schema.py
```diff
@@ -143,6 +143,16 @@ def from_dict(cls, dictionary):
         names, types = zip(*dictionary.items()) if dictionary else ([], [])
         return Schema(names, types)
 
+    @classmethod
+    def merge(cls, schemas):
```
cpcloud, line 147: Does this need to be a method on `Schema` as opposed to a utility function in your own codebase?

ibis/expr/operations/relations.py
```diff
@@ -78,6 +78,67 @@ def resolve_name(self):
...
+    table = rlz.table
+    name = rlz.instance_of(str)
+
+    @cached_property
+    def schema(self):
+        return self.table.schema()
+
+
+@public
+class AsOfMerge(TableNode, sch.HasSchema):
```
cpcloud, line 98: It's hard to tell why this has a bunch of hardcoded fields with their index as part of their name.

Why doesn't the existing `AsOfJoin` work for you?

ibis/expr/operations/relations.py
```diff
@@ -78,6 +78,67 @@ def resolve_name(self):
...
 
 
+@public
+class LocalTable(PhysicalTable):
+    schema = rlz.instance_of(sch.Schema)
+    name = rlz.instance_of(str)
+
+
+@public
+class LocalWrite(TableNode, sch.HasSchema):
```
cpcloud, line 88: I'm not sure if this should be modeled as a `TableNode`.

The thing that determines whether a node should be a table node is whether the operation can **produce** rows.

From your perspective, does this or should this operation produce rows?

ibis/expr/operations/relations.py
```diff
@@ -78,6 +78,67 @@ def resolve_name(self):
         return self.name
 
 
+@public
+class LocalTable(PhysicalTable):
```
cpcloud, line 82: This looks very similar to `ops.UnboundTable`. Does that not work for your use case?

ibis/expr/api.py
```diff
@@ -324,6 +325,29 @@ def table(schema: sch.Schema, name: str | None = None) -> ir.TableExpr:
     return node.to_expr()
 
 
+def local_table(schema: sch.Schema, name: str | None = None) -> ir.TableExpr:
+    """Create a local table for build expressions without data.
```
cpcloud, line 329: This is exactly what `ibis.table` does. Can you use that instead?

ibis/expr/analysis.py
```diff
@@ -826,7 +826,7 @@ def __init__(self, parent, proj_exprs):
         )
         self.clean_exprs = list(map(windowize_function, self.resolved_exprs))
 
-    def get_result(self):
+    def get_result(self, **kwargs):
```
cpcloud, line 829: Can you elaborate a bit on why this is necessary?

cpcloud (changes requested):

ibis/expr/operations/numeric.py
rtpsw: Not related to UDT; please skip.

ibis/expr/types/relations.py
rtpsw: Not related to UDT; please skip.

ibis/expr/types/numeric.py
rtpsw: Not related to UDT; please skip.

ibis/expr/schema.py
rtpsw: It can be, of course. OTOH, I think it is useful and generic enough to be considered for inclusion in core. In any case, it's not a question we need to resolve for UDT design purposes.

ibis/expr/operations/relations.py
rtpsw: Not related to UDT; please skip.

ibis/expr/operations/relations.py
rtpsw: Yes, it produces rows by passing-through. Not related to UDT; please skip.

ibis/expr/operations/relations.py
rtpsw: `LocalTable` is used for reading from a local resource. Not related to UDT; please skip.

ibis/expr/api.py
rtpsw: Not related to UDT; please skip.

ibis/expr/analysis.py
rtpsw: It might be a leftover. Not related to UDT; please skip.

rtpsw:
For the UDT design, please see the following:
- `TabularUserDefinedFunction` in [ibis/expr/operations/tabular.py](https://github.com/ibis-project/ibis/pull/4393/files#diff-b76ba5021d3b4ff2037da7b35f01850a17de49ddf901533967e915e80ac4ca95)
- [ibis/udf/table.py](https://github.com/ibis-project/ibis/pull/4393/files#diff-2eb11788febb740047c234b4c461763501697c0fe3ef887a067fe2b5c33ed26e)
- [ibis/expr/operations/udf.py](https://github.com/ibis-project/ibis/pull/4393/files#diff-246b9fe50fc80d395f987589bcb81cb5e20cf0781d1de2251a8b0690b78c5aa5)
- `VectorizedUDF` in [ibis/expr/operations/vectorized.py](https://github.com/ibis-project/ibis/pull/4393/files#diff-5392264b1de368f248ba38203e523e0092dc869a66dd3e348923f7658793e209)

ibis/expr/operations/tabular.py
```diff
@@ -0,0 +1,20 @@
...
+class TabularUserDefinedFunction(PhysicalTable, UDFMixin):
+    input_type = container_of(rlz.datatype, type=tuple, max_length=0)
+
+    @property
+    def schema(self):
+        if isinstance(self.return_type, dt.Struct):
+            return sch.Schema(self.return_type.names, self.return_type.types)
+        raise ValueError(f"Unknwon schema for return type {self.return_type}")
+
+    def output_dtype(self):
```
cpcloud, line 19: I don't think this is the right design. A table isn't a `dtype`. In fact, I don't think you need the `output_dtype` field at all.

cpcloud (changes requested):
Please show an example of usage, in a test preferably.

cpcloud:
@rtpsw Is this PR still viable? Happy to review it more, but it looks like maybe it's a bit stale now.

rtpsw:
> @rtpsw Is this PR still viable? Happy to review it more, but it looks like maybe it's a bit stale now.

@icexelloss, given current priorities, do we want to push this in the near future?

cpcloud:
Closing this out due to inactivity.

PR #4395, opened 2022-08-23
ci: remove pep8speaks

jcrist:
`pep8speaks` doesn't understand some modern python constructs like `:=`, causing it to unnecessarily flag some PRs. Since we don't let it actually fail PRs or fix issues, it's just a noisy bot. Lets remove it and rely on the precommit linters instead.

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,9 +1,0 @@
-# File : .pep8speaks.yml
-
-scanner:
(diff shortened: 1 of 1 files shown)


jcrist:
cc @cpcloud - in case there's an external change that may need to be made to fully disable the bot besides just deleting the config file.

cpcloud (approved):
LGTM! Thanks.

PR #4415, opened 2022-08-25
refactor(ir): remove ops.AggregateSelection in favor of an.simplify_aggregation

kszucs:


cpcloud (approved):
Sweeeet!

cpcloud:
Hm, the postgres failure looks like a possible race condition. Let's see if it reproduces.

PR #4417, opened 2022-08-25
refactor(ir): simplify SortKey construction using rules

kszucs:


ibis/expr/format.py
```diff
@@ -563,7 +563,9 @@ def _fmt_value_binary_op(op: ops.Binary, *, aliases: Aliases) -> str:
 
 @fmt_value.register
 def _fmt_value_negate(op: ops.Negate, *, aliases: Aliases) -> str:
-    op_name = "Not" if isinstance(op.arg.type(), dt.Boolean) else "Negate"
+    op_name = (
+        "Not" if isinstance(op.arg.output_dtype, dt.Boolean) else "Negate"
```
cpcloud, line 567: Can you use `op.output_dtype` here instead?

cpcloud (approved):
NICE. Goodbye sort key monters.

ibis/expr/format.py
kszucs: Updating

PR #4421, opened 2022-08-25
refactor(ir): remove ops.Node.inputs property because it is an implementation detail of the pandas backend

kszucs:


jreback:
prefer to merge to 4.x

PR #4422, opened 2022-08-25
fix: support filters on InMemoryTable exprs

jcrist:
Allows calling `filter` on `memtable` expressions. Wasn't sure how best to test this, may be overkill here. The bug was just a missing `blocks` implementation.


cpcloud:
Excellent, thank you. I'll take a look at the impala failures.

cpcloud (approved):
Thanks for the fix!

PR #4424, opened 2022-08-26
refactor(ir): generalize handling and traversal of node sequences

kszucs:
depends on #4421

### DIFF
--- a/ibis/backends/base/sql/alchemy/query_builder.py
+++ b/ibis/backends/base/sql/alchemy/query_builder.py
@@ -261,10 +261,7 @@ class AlchemySelect(Select):
         if not num_froms:
             return result.select_from(table_set)
 
-        # otherwise we expect a single FROM in `result`
-        # assert num_froms == 1, f"num_froms == {num_froms:d}"
-
-        # and then we need to replace every occurrence of `result`'s `FROM`
+        # we need to replace every occurrence of `result`'s `FROM`
         # with `table_set` to handle correlated EXISTs coming from
         # semi/anti-join
         #
@@ -273,7 +270,10 @@ class AlchemySelect(Select):
         #
         # sqlalchemy suggests using the functionality in sa.sql.visitors, but
         # that would effectively require reimplementing ClauseAdapter
-        return sa.sql.util.ClauseAdapter(table_set).traverse(result)
+        replaced = sa.sql.util.ClauseAdapter(table_set).traverse(result)
+        num_froms = len(replaced.get_final_froms())
+        assert num_froms == 1, f"num_froms == {num_froms:d}"
+        return replaced
 
     def _add_groupby(self, fragment):
         # GROUP BY and HAVING
--- a/ibis/backends/base/sql/alchemy/query_builder.py
+++ b/ibis/backends/base/sql/alchemy/query_builder.py
@@ -262,7 +262,7 @@ class AlchemySelect(Select):
             return result.select_from(table_set)
 
         # otherwise we expect a single FROM in `result`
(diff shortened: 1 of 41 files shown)


ibis/expr/rules.py
```diff
@@ -90,22 +97,17 @@ def just(arg):
     return lambda **_: arg
 
 
+# TODO(kszucs): perhaps rename to nodes_of
 @rule
-def value_list_of(inner, arg, **kwargs):
-    # TODO(kszucs): would be nice to remove ops.ValueList
+def list_of(inner, arg, **kwargs):
```
kszucs, line 102: This rule must be used to define a sequence of nodes which automatically traversed by other utilities. We already have a `tuple_of` rule representing a python tuple of arbitrary types so this naming might be a bit misleading, the reason I chose this was that to align with the operation name `ops.List`.

@cpcloud opinions?

ibis/backends/base/sql/alchemy/query_builder.py
```diff
@@ -262,7 +262,7 @@ def _add_select(self, table_set):
             return result.select_from(table_set)
 
         # otherwise we expect a single FROM in `result`
-        assert num_froms == 1, f"num_froms == {num_froms:d}"
+        # assert num_froms == 1, f"num_froms == {num_froms:d}"
```
kszucs, line 265: If I keep this assertion then `ibis/backends/tests/test_aggregation.py::test_filter` is failing for certain SQL backends.

ibis/expr/operations/core.py
```diff
@@ -1,21 +1,29 @@
...
 from ibis.expr.rules import Shape
-from ibis.expr.schema import Schema
-from ibis.util import UnnamedMarker, is_iterable
+from ibis.util import UnnamedMarker
 
 
 @public
 class Node(Annotable, Comparable):
+    @property
+    def args(self):
```
kszucs, line 18: I'd be up to removing of `args` and `argnames` in a follow-up.

cpcloud: Let's leave this as is for now.

ibis/expr/rules.py
cpcloud: Can we rename `ops.List` to `ops.NodeList` and `list_of` to `node_list_of`?

ibis/backends/base/sql/alchemy/query_builder.py
cpcloud: Well, if everything else is passing then it's probably fine to remove!

cpcloud: Can you delete this line?

ibis/expr/rules.py
cpcloud: Can you rename? I'll merge after that and the comment below on removing the assertion.

ibis/backends/base/sql/alchemy/query_builder.py
cpcloud: Actually, this is broken.

cpcloud: Ah, I think we need to move the assertion. One sec, I'll fix it up.

kszucs: Thanks!

ibis/expr/rules.py
kszucs: Sure

cpcloud (commented):
Nice work!

cpcloud (approved):

PR #4425, opened 2022-08-26
refactor(ir): make all value operations 'Named' for more consistent naming semantics

kszucs:
depends on #4424


ibis/expr/operations/core.py
```diff
@@ -37,17 +37,36 @@ def equals(self, other):
     def to_expr(self):
         ...
 
-    # TODO(kszucs): introduce a HasName schema, or NamedValue with a .name
-    # abstractproperty
-    def resolve_name(self):
-        raise ExpressionError(f'Expression is not named: {type(self)}')
 
-    def has_resolved_name(self):
```
kszucs, line 45: We can maintain a backward compatible API for another major release if that would be preferable, though we break internals heavily with the removal of intermediate expressions.

ibis/expr/types/numeric.py
```diff
@@ -461,6 +461,8 @@ def mean(
         NumericScalar
             The mean of the input expression
         """
+        # TODO(kszucs): remove the alias from the reduction method in favor
+        # of default name generated by ops.Value operations
         return ops.Mean(self, where=where).to_expr().name("mean")
```
kszucs, line 466: I tend prefer the default naming scheme here, so `t.float_col.mean()` gets the name `Mean(float_col)` instead of the current `mean`, opinions?

cpcloud: Let's remove the current set of default aliases in a follow-up.

cpcloud (approved):

ibis/expr/operations/core.py
cpcloud: This seems fine. I don't think there's much use of `has_resolved_name()` in the wild.

PR #4427, opened 2022-08-26
refactor(ir): split out table count into `CountStar` operation

cpcloud:


ibis/expr/types/relations.py
```diff
@@ -743,15 +743,20 @@ def filter(
...
         return an.apply_filter(table.op(), predicates).to_expr()
 
-    def count(self) -> ir.IntegerScalar:
+    def count(self, where: ir.BooleanValue | None = None) -> ir.IntegerScalar:
         """Compute the number of rows in the table.
 
+        Parameters
+        ----------
+        where
+            Filter expression
```
jcrist, line 752: This seems a bit brief for a new user to understand what `where` is for, could you expand this a bit (or add an example)?

Also ibis docstrings question: for optional parameters, do we usually say "Optional" somewhere in the docstring/say what the default behavior is? Or is that inferred from the function signature?

jcrist (commented):
I'm curious, why use this over `table.filter(...).count()`? Just a different way to spell it? Are there other reductions that accept a filter like this?

ibis/expr/types/relations.py
cpcloud: > I'm curious, why use this over `table.filter(...).count()`? Just a different way to spell it? Are there other reductions that accept a filter like this?

All of the other reductions accept a filter :)

cpcloud: I can add a more thorough description.

We usually don't say optional unless the behavior is more complex than "has no effect".

cpcloud:
Actually, aren't you using those filters in your datasette blog post?

jcrist:
No, I only use `.filter(...)` for filtering.

kszucs:
@cpcloud looks good to me on principle though a couple of builds are failing

kszucs:
The semantics for filtering reductions are different than applying reductions on a filtered table. As an example see the if aggregation combinator of clickhouse https://clickhouse.com/docs/en/sql-reference/aggregate-functions/combinators/#-if

ibis/expr/types/relations.py
cpcloud: Fixed!

PR #4430, opened 2022-08-26
fix: fix pandas & dask where implementations

jcrist:
This fixes a few issues with the pandas and dask `where` implementations:

- Neither implementation supported scalars for the `true` case, only the `false` case. This was most likely an oversight made difficult to see due to multiple functions being used to handle each case. We now use a single function for all cases, making the logic easier to understand.
- Neither implementation supported `None` or `str` values for `true` or `false`
- The dask implementation was _highly_ inefficient, resulting in unnecessary computation. This is now handled through dask-native mechanisms, resulting in major speedups.

### DIFF
--- a/ibis/backends/dask/execution/indexing.py
+++ b/ibis/backends/dask/execution/indexing.py
@@ -1,64 +1,40 @@
 """Execution rules for ops.Where operations"""
 
 import dask.dataframe as dd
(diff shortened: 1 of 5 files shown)


cpcloud (approved):
LGTM, thank you!

PR #4435, opened 2022-08-28
feat(clickhouse): enable support for working window functions

cpcloud:
This PR enables support for window functions that behave
out of the box like the rest of the SQL world does.

Issues:

1. A few window functions appear broken entirely when an `ORDER BY` statement
   is added. 
1. ~Some "working" functions cause the resulting query to return zero rows when the window function expression
   is part of another expression, e.g., `rank() over () - 1`.~ Fixed upstream already, waiting on release.
1. Some window functions just don't exist upstream, `percent_rank` and `cume_dist`

Otherwise, the basic aggregates work as expected.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.4"
 services:
   clickhouse:
-    image: clickhouse/clickhouse-server:22-alpine
+    image: clickhouse/clickhouse-server:22.6.6.16-alpine
     ports:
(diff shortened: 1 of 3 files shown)


cpcloud:
Running against clickhouse `HEAD`, the arithmetic-in-window-functions problem looks fixed: 
image

kszucs (approved):
Nice!

PR #4439, opened 2022-08-29
fix(duckdb): use specialized `array_length` implementation

### LINKED ISSUE #4438: fix(duckdb): `array_length` implementation generates incorrect code
```
In [16]: print(ibis.duckdb.compile(ibis.array([1]).length().name("n")).compile(compile_kwargs=dict(literal_binds=True)))
SELECT CASE WHEN (CAST(list_value(1) AS ARRAY) IS NULL) THEN NULL ELSE coalesce(array_length(CAST(list_value(1) AS ARRAY), 1), 0) END AS n
```

The `0` as the second argument to `array_length` is generated by the postgres implementation because it supports computing the length in a particular dimension of an array.

DuckDB doesn't support this, so our implementation should just call `sa.func.array_length(translated_arg)`.

The `ops.ArrayLength` translation rule should be specialized for duckdb.

gforsyth:
and avoid the extraneous arguments that are required for postgres arrays
that aren't required/supported in duckdb.

Fixes #4438

### DIFF
--- a/ibis/backends/duckdb/tests/test_compiler.py
+++ b/ibis/backends/duckdb/tests/test_compiler.py
@@ -1,0 +1,24 @@
+import pytest
+
+import ibis
+from ibis.backends.duckdb.compiler import DuckDBSQLCompiler
+
+compiler = DuckDBSQLCompiler()
+
+QUERY = """SELECT array_length(CAST(list_value{} AS ARRAY)) AS n"""
+
+
+@pytest.mark.parametrize(
+    "array",
+    [
+        range(10),
+        range(5),
+        [-1, 0, 1],
+    ],
+)
+def test_compiler_array_length(array):
+    t = ibis.array(array).length().name("n")
+    result = str(
+        compiler.to_sql(t).compile(compile_kwargs={"literal_binds": True})
+    )
+    assert result == QUERY.format(tuple(array))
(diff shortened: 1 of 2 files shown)


ibis/backends/duckdb/tests/test_compiler.py
```diff
@@ -0,0 +1,24 @@
...
+
+@pytest.mark.parametrize(
+    "array",
+    [
+        range(10),
+        range(5),
+        [-1, 0, 1],
+    ],
+)
+def test_compiler_array_length(array):
```
cpcloud, line 19: I thought we had a backend test for array_length, but maybe not. Can we add one instead of doing this?

gforsyth: we do, but it was passing for `duckdb` despite the wacky translation.

gforsyth: I'm happy to remove this wacky SQL string comparison thing, but I don't think the existing `array_length` tests are going to prevent a regression (this may not be a huge concern)

cpcloud:
Ok, I think I was just overzealous and assumed that duckdb didn't work with this and our tests were strange.

Sorry for roping you into this, I think there's nothing to here.

```
D select array_length([1, 2], 1);
┌────────────────────────────────────────┐
│ array_length(main.list_value(1, 2), 1) │
├────────────────────────────────────────┤
│ 2                                      │
└────────────────────────────────────────┘
```

PR #4453, opened 2022-09-01
feat: add elementwise bit operations

### LINKED ISSUE #866: feat: elementwise bit operations
As far as I can tell, there's no way to use the Impala [bitand](http://www.cloudera.com/documentation/enterprise/5-5-x/topics/impala_bit_functions.html) function or `&` operator on integer columns. That would be a valuable feature.

cpcloud:
I wanted to leave this issue for onboarding, but it of course turned out to be quite a slog. Closes #866.


kszucs (approved):
LGTM

Issue #4443, opened 2022-08-30
feat: methods to execute and return pyarrow Table or iterator of RecordBatches


jcrist:
Currently users call `expr.execute()` to execute an `ibis` expression and get back a `pandas.DataFrame`.

For efficiency and ease of interoperation with other tools, it would be useful to add some alternative result representations:

- A `pyarrow.Table`
- An iterator of `pyarrow.RecordBatch` objects

The second would allow for streaming back potentially large results without requiring the results fit in memory. The first _could_ be implemented using the second using something like:

```python
def to_pyarrow(self):
    return pa.Table.from_batches(self.to_pyarrow_batches())
```

however, some backends may have a more efficient method of returning a `Table` than that (e.g. duckdb's `fetch_record_batch`).

Since some additional options may show up on these methods (perhaps a batch size for the record batches version), I think it would be best to implement these as new methods on `Table` rather than overloading the existing `execute` method. No strong thoughts on what the methods should be called, for now I'll suggest:

- `to_pyarrow(self) -> pa.Table`
- `to_pyarrow_batches(self) -> Iterator[pa.RecordBatch]`

gforsyth:
For the `sqlalchemy` backends that don't have special methods there's going to be at least a minor annoyance here in that the values returned from a `cursor` look like `tuple`s and `dict`s but in fact are `sqlalchemy.engine.row.LegacyRow` and `sqlalchemy.engine.row.RowMapping`.  

It seems like `pyarrow` really doesn't like these, e.g.

```python
In [168]: batch[:5]
Out[168]: [(1, 2173), (1, 943), (1, 892), (1, 30), (1, 337)]

In [169]: pa_schema = pa.struct([("l_orderkey", pa.int32()), ("l_partkey", pa.int32())])

In [170]: pa.array(batch[:5], type=pa_schema)
---------------------------------------------------------------------------
ArrowTypeError                            Traceback (most recent call last)
Input In [170], in <cell line: 1>()
----> 1 pa.array(batch[:5], type=pa_schema)

File /nix/store/z9qn3g22d8nx1x4mgzq3497iy8ji5h8x-python3-3.10.6-env/lib/python3.10/site-packages/pyarrow/array.pxi:317, in pyarrow.lib.array()

[...]

PR #4454, opened 2022-09-02
feat: add to_pyarrow and to_pyarrow_batches

gforsyth:
Adds `to_pyarrow` and `to_pyarrow_batches` to the alchemy backends, datafusion, and pandas.  More to come.

Some open questions / issues:
Where should the schema inference stuff live?  The type mapping is already defined in the `pyarrow` backend but it feels weird to import that backend into other backends.

`chunk_size`?  `chunksize`?  `batch_size`?

In this context, are chunks / batches properly defined in terms of rows or is that a conflation?

Datafusion has some slightly wonky behavior when it comes to a consistent schema across recordbatches
DuckDB does not seem to respect their own `chunksize` argument.

xref #4443

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -231,6 +231,7 @@ class ResultHandler:
             # ColumnExpr has no schema method, define single-column schema
             return sch.schema([(expr.get_name(), expr.type())])
 
+    @util.experimental
     def to_pyarrow(
         self,
         expr: ir.Expr,
@@ -239,6 +240,26 @@ class ResultHandler:
         limit: int | str | None = None,
         **kwargs: Any,
     ) -> pa.Table:
+        """Execute expression and return results in as a pyarrow table.
+
+        This method is eager and will execute the associated expression
+        immediately.
+
+        Parameters
+        ----------
+        expr
+            Ibis expression to export to pyarrow
+        params
+            Mapping of scalar parameter expressions to value.
+        limit
+            An integer to effect a specific row limit. A value of `None` means
+            "no limit". The default is in `ibis/config.py`.
+
+        Returns
+        -------
+        Table
+            A pyarrow table holding the results of the executed expression.
+        """
         pa = self._import_pyarrow()
         try:
             # Can't construct an array from record batches
@@ -269,6 +290,7 @@ class ResultHandler:
         else:
             raise ValueError
 
+    @util.experimental
     def to_pyarrow_batches(
         self,
         expr: ir.Expr,
@@ -278,6 +300,29 @@ class ResultHandler:
         chunk_size: int = 1_000_000,
         **kwargs: Any,
     ) -> pa.RecordBatchReader:
+        """Execute expression and return results in an iterator of pyarrow
+        record batches.
+
+        This method is eager and will execute the associated expression
+        immediately.
+
+        Parameters
+        ----------
+        expr
+            Ibis expression to export to pyarrow
+        limit
+            An integer to effect a specific row limit. A value of `None` means
+            "no limit". The default is in `ibis/config.py`.
+        params
+            Mapping of scalar parameter expressions to value.
+        chunk_size
+            Number of rows in each returned record batch.
+
+        Returns
+        -------
+        record_batches
+            An iterator of pyarrow record batches.
+        """
         raise NotImplementedError
 
 
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -22,6 +22,7 @@ from typing import (
 
 if TYPE_CHECKING:
     import pandas as pd
+    import pyarrow as pa
 
 import ibis
 import ibis.common.exceptions as exc
@@ -206,7 +207,65 @@ class TablesAccessor(collections.abc.Mapping):
         return self._backend.list_tables()
 
 
-class BaseBackend(abc.ABC):
+# should have a better name
+class ResultHandler:
+    @staticmethod
+    def _import_pyarrow():
+        try:
+            import pyarrow
+        except ImportError:
+            raise ModuleNotFoundError(
+                "Exporting to arrow formats requires `pyarrow` but it is not installed"  # noqa: ignore
+            )
+        else:
+            return pyarrow
+
+    def to_pyarrow(
+        self,
+        expr: ir.Expr,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        **kwargs: Any,
+    ) -> pa.Table:
+        pa = self._import_pyarrow()
+        try:
+            # Can't construct an array from record batches
+            # so construct at one column table (if applicable)
+            # then return the column _from_ the table
+            table = pa.Table.from_batches(
+                self.to_pyarrow_batches(
+                    expr, params=params, limit=limit, **kwargs
+                )
+            )
+            if isinstance(expr, ir.Table):
+                return table
+            elif isinstance(expr, ir.Column):
+                # Column will be a ChunkedArray, `combine_chunks` will
+                # flatten it
+                return table.columns[0].combine_chunks()
+            elif isinstance(expr, ir.Scalar):
+                return table.columns[0].combine_chunks()[0]
+            else:
+                raise ValueError
+        except ValueError:
+            # The pyarrow batches iterator is empty so pass in an empty
+            # iterator and the pyarrow schema
+            return pa.Table.from_batches([], schema=expr.schema().to_pyarrow())
+
+    def to_pyarrow_batches(
+        self,
+        expr: ir.Expr,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        chunk_size: int = 1_000_000,
+        **kwargs: Any,
+    ) -> pa.RecordBatchReader:
+        raise NotImplementedError
+
+
+class BaseBackend(abc.ABC, ResultHandler):
     """Base backend class.
 
     All Ibis backends must subclass this class and implement all the
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -23,6 +23,7 @@ from typing import (
 if TYPE_CHECKING:
     import pandas as pd
     import pyarrow as pa
+    import ibis.expr.schema as sch
 
 import ibis
 import ibis.common.exceptions as exc
@@ -220,6 +221,16 @@ class ResultHandler:
         else:
             return pyarrow
 
+    @staticmethod
+    def _table_or_column_schema(expr: ir.Expr) -> sch.Schema:
+        from ibis.backends.pyarrow.datatypes import sch
+
+        if isinstance(expr, ir.Table):
+            return expr.schema()
+        else:
+            # ColumnExpr has no schema method, define single-column schema
+            return sch.schema([(expr.get_name(), expr.type())])
+
     def to_pyarrow(
         self,
         expr: ir.Expr,
@@ -238,20 +249,25 @@ class ResultHandler:
                     expr, params=params, limit=limit, **kwargs
                 )
             )
-            if isinstance(expr, ir.Table):
-                return table
-            elif isinstance(expr, ir.Column):
-                # Column will be a ChunkedArray, `combine_chunks` will
-                # flatten it
-                return table.columns[0].combine_chunks()
-            elif isinstance(expr, ir.Scalar):
-                return table.columns[0][0]
-            else:
-                raise ValueError
         except ValueError:
             # The pyarrow batches iterator is empty so pass in an empty
-            # iterator and the pyarrow schema
-            return pa.Table.from_batches([], schema=expr.schema().to_pyarrow())
+            # iterator and a pyarrow schema
+            schema = self._table_or_column_schema(expr)
+            table = pa.Table.from_batches([], schema=schema.to_pyarrow())
+
+        if isinstance(expr, ir.Table):
+            return table
+        elif isinstance(expr, ir.Column):
+            # Column will be a ChunkedArray, `combine_chunks` will
+            # flatten it
+            if len(table.columns[0]):
+                return table.columns[0].combine_chunks()
+            else:
+                return pa.array(table.columns[0])
+        elif isinstance(expr, ir.Scalar):
+            return table.columns[0][0]
+        else:
+            raise ValueError
 
     def to_pyarrow_batches(
         self,
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -245,7 +245,7 @@ class ResultHandler:
                 # flatten it
                 return table.columns[0].combine_chunks()
             elif isinstance(expr, ir.Scalar):
-                return table.columns[0].combine_chunks()[0]
+                return table.columns[0][0]
             else:
                 raise ValueError
         except ValueError:
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -171,11 +171,7 @@ class BaseSQLBackend(BaseBackend):
 
         from ibis.backends.pyarrow.datatypes import ibis_to_pyarrow_struct
 
-        if hasattr(expr, "schema"):
-            schema = expr.schema()
-        else:
-            # ColumnExpr has no schema method, define single-column schema
-            schema = sch.schema([(expr.get_name(), expr.type())])
+        schema = self._table_or_column_schema(expr)
 
         def _batches():
             for batch in self._cursor_batches(
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -4,7 +4,7 @@ import abc
 import contextlib
 import os
 from functools import lru_cache
-from typing import Any, Mapping
+from typing import TYPE_CHECKING, Any, Iterable, Mapping
 
 import sqlalchemy as sa
 
@@ -16,6 +16,9 @@ import ibis.util as util
 from ibis.backends.base import BaseBackend
 from ibis.backends.base.sql.compiler import Compiler
 from ibis.expr.typing import TimeContext
+
+if TYPE_CHECKING:
+    import pyarrow as pa
 
 __all__ = [
     'BaseSQLBackend',
@@ -127,8 +130,6 @@ class BaseSQLBackend(BaseBackend):
         Any
             Backend cursor
         """
-        # TODO results is unused, it can be removed
-        # (requires updating Impala tests)
         # TODO `self.con` is assumed to be defined in subclasses, but there
         # is nothing that enforces it. We should find a way to make sure
         # `self.con` is always a DBAPI2 connection, or raise an error
@@ -140,6 +141,55 @@ class BaseSQLBackend(BaseBackend):
     @contextlib.contextmanager
     def _safe_raw_sql(self, *args, **kwargs):
         yield self.raw_sql(*args, **kwargs)
+
+    def _cursor_batches(
+        self,
+        expr: ir.Expr,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        chunk_size: int = 1_000_000,
+    ) -> Iterable[list]:
+        query_ast = self.compiler.to_ast_ensure_limit(
+            expr, limit, params=params
+        )
+        sql = query_ast.compile()
+
+        with self._safe_raw_sql(sql) as cursor:
+            while batch := cursor.fetchmany(chunk_size):
+                yield batch
+
+    def to_pyarrow_batches(
+        self,
+        expr: ir.Expr,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        chunk_size: int = 1_000_000,
+        **kwargs: Any,
+    ) -> pa.RecordBatchReader:
+        pa = self._import_pyarrow()
+
+        from ibis.backends.pyarrow.datatypes import ibis_to_pyarrow_struct
+
+        if hasattr(expr, "schema"):
+            schema = expr.schema()
+        else:
+            # ColumnExpr has no schema method, define single-column schema
+            schema = sch.schema([(expr.get_name(), expr.type())])
+
+        def _batches():
+            for batch in self._cursor_batches(
+                expr, params=params, limit=limit, chunk_size=chunk_size
+            ):
+                struct_array = pa.array(
+                    map(tuple, batch),
+                    type=ibis_to_pyarrow_struct(schema),
+                )
+                yield pa.RecordBatch.from_struct_array(struct_array)
+
+        return pa.RecordBatchReader.from_batches(
+            schema.to_pyarrow(), _batches()
+        )
 
     def execute(
         self,
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -158,6 +158,7 @@ class BaseSQLBackend(BaseBackend):
             while batch := cursor.fetchmany(chunk_size):
                 yield batch
 
+    @util.experimental
     def to_pyarrow_batches(
         self,
         expr: ir.Expr,
@@ -167,6 +168,29 @@ class BaseSQLBackend(BaseBackend):
         chunk_size: int = 1_000_000,
         **kwargs: Any,
     ) -> pa.RecordBatchReader:
+        """Execute expression and return results in an iterator of pyarrow
+        record batches.
+
+        This method is eager and will execute the associated expression
+        immediately.
+
+        Parameters
+        ----------
+        expr
+            Ibis expression to export to pyarrow
+        limit
+            An integer to effect a specific row limit. A value of `None` means
+            "no limit". The default is in `ibis/config.py`.
+        params
+            Mapping of scalar parameter expressions to value.
+        chunk_size
+            Number of rows in each returned record batch.
+
+        Returns
+        -------
+        record_batches
+            An iterator of pyarrow record batches.
+        """
         pa = self._import_pyarrow()
 
         from ibis.backends.pyarrow.datatypes import ibis_to_pyarrow_struct
--- a/ibis/backends/datafusion/__init__.py
+++ b/ibis/backends/datafusion/__init__.py
@@ -20,6 +20,8 @@ try:
 except ImportError:
     from datafusion import SessionContext
 
+import datafusion
+
 
 def _to_pyarrow_table(frame):
     batches = frame.collect()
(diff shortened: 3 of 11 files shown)


ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -144,6 +146,37 @@ def _to_geodataframe(df, schema):
...
+        with self._safe_raw_sql(sql) as cursor:
+            while batch := cursor.fetchmany(chunk_size):
+                struct_array = pa.array(
+                    map(tuple, batch),
+                    type=_schema_to_pyarrow_struct(expr.schema()),
+                )
+                # TODO should we return a RecordBatchReader from this iterator?
+                yield pa.RecordBatch.from_struct_array(struct_array)
+
+    def to_pyarrow(
```
kszucs, line 170: For clarity we could call this `to_pyarrow_table`

ibis/backends/pyarrow/datatypes.py
```diff
@@ -129,3 +131,14 @@ def infer_pyarrow_schema(schema: pa.Schema) -> sch.Schema:
     return sch.schema(
         [(f.name, dt.dtype(f.type, nullable=f.nullable)) for f in schema]
     )
+
```
kszucs, line 134: I think we should keep the pyarrow dependent logic under the pyarrow subpackage.

For example we register the `DataType.to_pandas` and `Schema.to_pandas` methods from the pandas backend, see https://github.com/ibis-project/ibis/blob/master/ibis/backends/pandas/client.py#L300-L301

ibis/backends/tests/test_export.py
```diff
@@ -0,0 +1,110 @@
...
+    ["datafusion"],
+    reason="""
+        fields' nullability from frame.schema() is not always consistent with
+        the first record batch's schema
+""",
+)
+def test_to_pyarrow_table_schema(backend, awards_players):
+    table = awards_players.to_pyarrow()
+    assert isinstance(table, pa.Table)
+    assert table.schema == pa.schema(
```
kszucs, line 73: This could be `table.schema == awards_players.schema().to_pyarrow()`

ibis/expr/types/core.py
```diff
@@ -237,6 +238,59 @@ def _find_backend(self) -> BaseBackend:
 
         return backends[0]
 
+    def to_pyarrow_batches(
```
kszucs, line 241: Could we generalize this API to make the output formats pluggable? I also like the verb `execute` because it indicates that we are actually executing the ibis expression.

I have something like the following in mind:

```python
expr.execute(format=pd.DataFrame, **options)
expr.execute(pa.Table, **options)
expr.execute(pa.RecordBatch, **options)
expr.execute(polars.DataFrame, **options)
```

We could have string aliases for different formats too.

kszucs: Perhaps we could have an intermediate abstraction, like a `ResultFormatter` or `Executor` which can be overloaded:

```py
@executor.register(PandasBackend, pa.Table)
def pandas_to_pyarrow_table(backend, typ):
    ....


@executor.register(PandasBackend, pa.RecordBatch)
def pandas_to_pyarrow_recordbatch(backend, typ):
    ....


@executor.register(PandasBackend, list)
def pandas_to_plain_python_list(backend, typ):
    ....

@executor.register(PyarrowBackend, pd.DataFrame)
def pyarrow_to_pandas_dataframe(backend, typ):
    ...
```

ibis/backends/base/sql/alchemy/__init__.py
gforsyth: I'm torn on this -- I like things to be explicit AND I don't want to make it too verbose

ibis/backends/pyarrow/datatypes.py
gforsyth: Yeah, this looks nicer, doing this now

ibis/backends/base/sql/alchemy/__init__.py
kszucs: If we consider `pa.Table` as the default pyarrow dataframe-like object, then I'm fine with leaving it as is.

jcrist: Both `cudf` and `polars` call this `to_arrow`, so there's precedent for a simple name. I have a slight preference for `to_pyarrow` over `to_arrow`, but could go either way here. I don't think we need to specify `_table` though.

ibis/expr/types/core.py
kszucs: I may not mentioned my primary concern previously. Let's say we have something like the following in the future:

```python
ds = ibis.dataset("s3://bucket/...", schema=...)
expr = ds.filter(...).to_arrow()
```

I would like to execute it using datafusion not pyarrow, how could I define it?
Or I'd like to try out clickhouse and compare the performance of the two?

As far as I can see we need to define both the backend and the format somehow so the user can easily switch engines under the hood.

kszucs: We usually allow to `ibis.<backend>.execute(expr)`. I think that `.to_arrow()` or `.to_csv()` shouldn't trigger an actual execution until `.execute()` is called. 

```python
expr = table.filter(...).to_csv('local.csv')
ibis.pandas.execute(expr)
ibis.pyarrow.execute(expr)
```

I'd also like to avoid hard dependencies like pyarrow or pandas (which is currently). If we hardcode these methods to the internal types then we need to handle multiple optional imports and it can grow in the future.

kszucs: Perhaps we could define specific operation nodes for different formats. 3rdparty backends could define their own operations and the corresponding execution logic too.

kszucs:
```py
class Format(Node):
    pass

class PandasFormat(Format):
    options...

class ArrowFormat(Format):
    options...

# it would be still an ibis expression
arrow_expr = t.filter(...).to_arrow(**options)
pandas_expr = t.filter(...).to_pandas(**options)

# this would execute the expression using the datafusion backend
ibis.datafusion.execute(arrow_expr) -> pa.Table
ibis.datafusion.execute(pandas_expr) -> pd.DataFrame
```

jcrist: For the record, I'm not saying that `pyarrow` should be a required dependency, these `to_*` methods can raise a nice exception if their dependencies aren't installed.

jreback: I agree preserving unbound table behavior is paramount

@cpcloud thoughts here

gforsyth: Guarding the imports to avoid the required dependency is pretty straightforward (I'm just working on clearing out a strange issue in the clickhouse tests).

We don't want to hurt the ergonomics of someone using a bound-table, but we also don't want to kneecap unbound tables.

When we have something like `Sinks` in a more concrete form, then we should have no issue executing various file and stream exports on unbound tables via `execute` or with a method on the connection.

kszucs: I'm fine with doing this incrementally, though if we reserve the `Expr.to_pyarrow()` method for actual execution in the upcoming release then we cannot really change it to produce an expression later on.

jcrist: I don't love the idea of an expression that indicates an operation that's really handled in `ibis` itself rather than the backend. A `ToPyarrow` expression couldn't be composed with anything else - it's just a delayed task to eventually execute with `.execute()`. Further, it changes the output of `.execute()`, which currently expects to return a scalar or a pandas thing - this wouldn't play well with our existing `__repr__` code for interactive mode, and makes it harder to tell what `.execute()` will actually return.

Since `to_pyarrow` and `to_pyarrow_batches` are methods on the backend, what's wrong with using `con.to_pyarrow(expr)` to work with unbound tables?

Alternatively (I like this less), what if `to_*` methods execute immediately, but support an optional `execute=False` kwarg to return a delayed node that can later be passed to `con.execute(...)`?

kszucs: > Since to_pyarrow and to_pyarrow_batches are methods on the backend, what's wrong with using con.to_pyarrow(expr) to work with unbound tables?

That's fine, but it is also exposed on `Expr`.

jcrist: > That's fine, but it is also exposed on Expr.

I don't follow? So is `.execute()`, which also doesn't work for unbound tables. All our docs point users to using bound tables - a method version for these conversions is more ergonomic and is what I would expect as a user. For expert users making use of unbound tables, the method on the `Backend` is still available. The existence of one doesn't prevent the user of the other.

cpcloud: Unbound tables are an important use case. When using ibis as a library, you often are just merrily constructing expressions without a backend. Eventually you apply that to the backend. To belabor the point: this scenario is most common when ibis is being used as a library not as a tool for interactive data analysis.

We should take unbound tables into account when designing APIs (I am definitely guilty not doing this all the time!) and not treat them as uncommon or not the norm.

That's the library case. On the other hand when you're in interactive mode, you want to type less and have things "just run" or "just work" or "Just Do What I Think This Should".

There are few things to disentangle here, but I think before we do any more work on this we should come up with a few scenarios and design the API from that.

I think the dependencies issue can be resolved :)

The most important thing in my mind is to get those scenarios (let's say 2 or 3).

I'll start a github discussion so we can map those out.

ibis/backends/base/sql/__init__.py
```diff
@@ -132,6 +132,32 @@ def raw_sql(self, query: str) -> Any:
...
+    ):
+        raise NotImplementedError
+
+    def to_pyarrow(
+        self,
+        expr: ir.Expr,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+    ):
+        raise NotImplementedError
```
kszucs, line 159: A default implementation could create the table from the result of `to_pyarrow_batches()`

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -144,6 +147,44 @@ def _to_geodataframe(df, schema):
...
+
+        with self._safe_raw_sql(sql) as cursor:
+            while batch := cursor.fetchmany(chunk_size):
+                struct_array = pa.array(
+                    map(tuple, batch), type=expr.schema().to_pyarrow_struct()
+                )
+                # TODO should we return a RecordBatchReader from this iterator?
+                yield pa.RecordBatch.from_struct_array(struct_array)
+
+    def to_pyarrow(
```
kszucs, line 175: Having a default implementation would make it obsolete.

ibis/backends/base/sql/__init__.py
```diff
@@ -132,6 +132,32 @@ def raw_sql(self, query: str) -> Any:
...
+    @staticmethod
+    def _check_pyarrow_install():
+        try:
+            import pyarrow as pa  # noqa: ignore
+        except ImportError:
+            raise ModuleNotFoundError(
+                "Exporting to Arrow formats requires `pyarrow` but it is not installed"  # noqa: ignore
+            )
+
+    def to_pyarrow_batches(
```
kszucs, line 144: This could go to the `BaseBackend` class.

ibis/backends/base/sql/alchemy/__init__.py
kszucs: How about factoring out these methods to a utility class where we can have sane defaults? Something like:

```py
# should have a better name
class ResultHandler:

    @staticmethod
    def _import_pandas():
        try:
            import pandas
        except ImportError:
            raise ModuleNotFoundError(
                "Exporting to pandas formats requires `pandas` but it is not installed"  # noqa: ignore
            )
        else:
            return pandas

    @staticmethod
    def _import_pyarrow():
        try:
            import pyarrow
        except ImportError:
            raise ModuleNotFoundError(
                "Exporting to arrow formats requires `pyarrow` but it is not installed"  # noqa: ignore
            )
        else:
            return pyarrow

    def to_pylist(self, ...):
        return toolz.concat(self.to_pylist_batches(expr, ...))

    def to_pylist_batches(self, expr, ...):
        ...

    def to_pandas(self, expr, ...):
        pd = self._import_pandas()
        return pd.DataFrame(self.to_pylist(expr, ...))

    def to_pyarrow(self, expr, ...):
        pa = self._import_pyarrow()
        return pa.Table.from_batches(self.to_pyarrow_batches(expr, ...))

    def to_pyarrow_batches(self, expr, ...):
        pa = self._import_pyarrow()
        for batch in self.to_pylist_batches(expr, ...):
            struct_array = pa.array(
                map(tuple, batch), type=expr.schema().to_pyarrow_struct()
            )
            yield pa.RecordBatch.from_struct_array(struct_array)
        

class BaseBackend(abc.ABC, ResultHandler)
    ...

    def execute(self, expr, ...):
        return self.to_pandas(self, expr, ...)
```

ibis/backends/datafusion/__init__.py
```diff
@@ -183,7 +181,46 @@ def execute(
...
+        expr: ir.Expr,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        chunk_size: int = 1_000_000,
+        **kwargs: Any,
+    ) -> Iterable[pa.RecordBatch]:
+        frame = self._get_frame(expr, params, limit, **kwargs)
+        yield from frame.collect()
+
+    def to_pyarrow(
```
kszucs, line 201: This shouldn't be necessary by having a default implementation for `to_pyarrow()`

ibis/backends/datafusion/__init__.py
```diff
@@ -183,7 +181,46 @@ def execute(
...
+        self,
+        expr: ir.Expr,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        **kwargs: Any,
+    ) -> pa.Table:
+        frame = self._get_frame(expr, params, limit, **kwargs)
+        return _to_pyarrow_table(frame)
```
kszucs, line 209: Possibly `_to_pyarrow_table()` could be removed.

ibis/backends/pyarrow/datatypes.py
```diff
@@ -129,3 +131,21 @@ def infer_pyarrow_schema(schema: pa.Schema) -> sch.Schema:
...
+    return pa.struct(_schema_to_pyarrow_schema_fields(schema))
+
+
+def ibis_to_pyarrow_schema(schema: sch.Schema) -> pa.Schema:
+    return pa.schema(_schema_to_pyarrow_schema_fields(schema))
+
+
+dt.DataType.to_pyarrow = to_pyarrow_type  # type: ignore
+sch.Schema.to_pyarrow = ibis_to_pyarrow_schema  # type: ignore
+sch.Schema.to_pyarrow_struct = ibis_to_pyarrow_struct  # type: ignore
```
kszucs, line 151: It's a bit weird, could we call `ibis_to_pyarrow_struct` directly instead of binding it to `Schema`?

ibis/expr/types/core.py
kszucs: Sounds good, until that we can merge this marked as an experimental API.

jreback: umm how about we have this discussion first before merging

ibis/backends/base/sql/__init__.py
gforsyth: Ahh, yeah, that's what I'm doing in the `AlchemyBackend` but I can move that bit one step up the class hieararchy

ibis/backends/pyarrow/datatypes.py
gforsyth: No problem

ibis/backends/duckdb/__init__.py
```diff
@@ -170,6 +171,39 @@ def register(
...
+        chunk_size: int = 1_000_000,
+    ) -> Iterator[pa.RecordBatch]:
+        self._check_pyarrow_install()
+        query_ast = self.compiler.to_ast_ensure_limit(
+            expr, limit, params=params
+        )
+        sql = query_ast.compile()
+
+        with self._safe_raw_sql(sql) as cursor:
+            yield from cursor.cursor.fetch_record_batch(chunk_size=chunk_size)
```
Mause, line 188: Just so I understand this code correctly, it's calling down into the `ConnectionWrapper` instance in `duckdb-engine`,  through `__getattr__` on `ConnectionWrapper`?

gforsyth: Sorry for the delay -- yes, that's what it's doing.

ibis/backends/base/sql/alchemy/__init__.py
gforsyth: Tried this out as you suggested and I think I like it.  Not yet re-routing calls to `execute`

ibis/backends/base/__init__.py
```diff
@@ -206,7 +210,82 @@ def _ipython_key_completions_(self) -> list[str]:
...
+
+    def to_pyarrow_batches(
+        self,
+        expr: ir.Expr,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        chunk_size: int = 1_000_000,
+        **kwargs: Any,
+    ) -> Iterable[pa.RecordBatch]:
```
wjones127, line 277: IMO it would be preferable to return a `RecordBatchReader`, which allows introspecting the schema without needing get any of the batches out of the iterable.

gforsyth: Cool, I can definitely swap that in.

Question: do you know if I pass in an iterator of batches to `RecordBatchReader.from_batches` are they consumed eagerly or will it defer until the `read_next_batch` call?

wjones127: It will defer.

ibis/backends/duckdb/__init__.py
```diff
@@ -184,6 +186,44 @@ def register(
...
+        pa = self._import_pyarrow()
+        query_ast = self.compiler.to_ast_ensure_limit(
+            expr, limit, params=params
+        )
+        sql = query_ast.compile()
+
+        cursor = self.raw_sql(sql)
+
+        _reader = cursor.cursor.fetch_record_batch(chunk_size=chunk_size)
+        # Horrible hack to make sure cursor isn't garbage collected before batches
```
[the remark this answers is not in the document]
gforsyth: Was just about to!  And I was going to ping you, because I can't tell if DuckDB is doing the right thing and duckdb_engine should copy, or vice-versa, or if there is no right thing at all

ibis/backends/tests/test_export.py
```diff
@@ -0,0 +1,132 @@
...
+                    "pyspark",
+                ]
+            ),
+        ],
+    ),
+]
+
+
+@pytest.mark.parametrize("limit", limit_no_limit)
+def test_to_pyarrow_batches(limit, backend, awards_players):
```
kszucs, line 65: @gforsyth could you please add tests for expressions with different shape, like `table.column.to_pyarrow()` and `table.column.sum().to_pyarrow()`?

jcrist: Hmmm, `to_pyarrow`/`to_pyarrow_batches`/`to_pylist`/`to_pylist_batches` all feel weird to me to be on a `Scalar` expr. Should we instead add those only on `Column` and `Table` expressions?

gforsyth: I moved `to_pyarrow` and `to_pyarrow_batches` to `Column` and `Table` -- also added tests for both table and column exports.

It's still possible to call `to_pyarrow` from the backend directly on a scalar -- I wonder if we should be adding in (basic) dimension information, like a `ndims` attribute to expressions to make this easier to dispatch on

cpcloud: I think it's strange to not define `to_pyarrow` for scalars. It makes downstream integration and use of the API difficult because now such consumers have to build additional code to handle the scalar case uniformly. There are pyarrow scalar objects that would make sense here. Can't we handle that for users?

gforsyth: Ok, i can see a two ways of handling this (there are others, but they seem less self-consistent):

1.  Everything is a table
Table -> pa.Table
Column -> pa.Table with only one column
Scalar -> pa.Table with only one column and one row
^^^ this is probably wacky

2. Things go to their corresponding shape
Table -> pa.Table
Column -> pa.Array
Scalar -> pa.Scalar

This is more consistent, but downstream users will probably still need additional code to handle the output

cpcloud: I am also aware this PR has been open for a while so please do not consider my comments blocking.

jcrist: 1 makes the most sense to me for Tables/columns, but for scalars (if we handle those) I think the scalar should be returned directly. Idk though.

Phillip's argument about API consistency makes sense to me for `to_pyarrow`, especially since there are `pyarrow` types for scalars. `to_pylist` for scalars still feels a bit off (if the scalar type is an `Array`/`Struct`/`Map` type are we converting to a list, or is a list just the outer collection for table/column types?). The `to_*_batches` methods also feel off to me to define on scalars, Phillip are you fine dropping those from scalar types?

kszucs: I'd favor the second case for consistency, the user can easily convert the pyarrow scalar to a python scalar using `pa_scalar.as_py()`

jcrist: @kszucs do you think a `Column` should map to a `pa.Table` with a single column, or a `pa.Array` type?

kszucs: IMO the latter, currently we return with a pandas series for a column expression

gforsyth:
Adds `to_pyarrow` and `to_pyarrow_batches` to the `BaseBackend`.

`to_pyarrow` returns pyarrow objects consistent with the dimension of
the output:
  - a table -> pa.Table
  - a column -> pa.Array
  - a scalar -> pa.Scalar

`to_pyarrow_batches` returns a `RecordBatchReader` that returns batches of
pyarrow tables.  It does not have the same dimension handling because
that is not available in RecordBatchReaders.

`to_pyarrow_batches` is implemented for `AlchemyBackend`, `datafusion`,
and `duckdb`.

The `pandas` backend has `to_pyarrow` implemented using
`pandas.DataFrame.to_pyarrow()`.

Backends that do not require pyarrow already will only require it when
using `to_pyarrow*` methods.

There are warnings on these methods to indicate that they are
experimental and that they may break in the future irrespective of
semantic versioning.

The DuckDB `to_pyarrow_batches` makes use of a proxy object to escape
garbage collection so that the underlying record batches are still
available even after the `cursor` used to generate them _would have_
been garbage collected (but isn't because it is embedded in the proxy
object)

`to_pyarrow_batches` is implemented for scalars even though it's ambiguous, because `to_pyarrow` makes use of the `RecordBatchReader` from `to_pyarrow_batches` to output tables (and arrays, and scalars).  If it is called on a Scalar, it returns a RecordBatchReader that has a single batch that resolves to a pyarrow table with one row and one column.

ibis/backends/base/__init__.py
```diff
@@ -206,7 +207,65 @@ def _ipython_key_completions_(self) -> list[str]:
...
+        try:
+            import pyarrow
+        except ImportError:
+            raise ModuleNotFoundError(
+                "Exporting to arrow formats requires `pyarrow` but it is not installed"  # noqa: ignore
+            )
+        else:
+            return pyarrow
+
+    def to_pyarrow(
```
kszucs, line 223: I wonder, could we move this to `ibis.backends.pyarrow` somehow (fine in a follow-up too)?

Also, how could another backend, say polars register its own result types?

kszucs:
Thanks for the updates and the summary, going to have a more thorough look tomorrow.

ibis/backends/base/__init__.py
gforsyth: We could move it, but I was planning (in a follow-up) to add a `to_pandas` method to the mixin.  However, it might be useful (to your point about e.g. polars registering a result type) to have a means for backends to register their own `to_*` methods.  Then the `to_pyarrow` would be defined in the `pyarrow` backend and `to_pandas` in the `pandas` backend.

Seems doable for the methods called via `backend.to_*` -- I'm not sure yet how we might also then add the relevant methods to expression objects

kszucs: Doing it in a follow up sounds good to me.

ibis/backends/datafusion/__init__.py
```diff
@@ -184,8 +182,43 @@ def execute(
...
+        params: Mapping[ir.Expr, object] = None,
+        limit: int | str | None = "default",
+        **kwargs: Any,
+    ):
+        output = self.to_pyarrow(expr, params=params, limit=limit, **kwargs)
+        if isinstance(expr, ir.Table):
+            return output.to_pandas()
+        elif isinstance(expr, ir.Column):
+            series = output.to_pandas()
+            series.name = "tmp"
```
kszucs, line 218:
```suggestion
            series.name = expr.get_name()
```

ibis/backends/base/__init__.py
```diff
@@ -206,7 +207,65 @@ def _ipython_key_completions_(self) -> list[str]:
...
+                # flatten it
+                return table.columns[0].combine_chunks()
+            elif isinstance(expr, ir.Scalar):
+                return table.columns[0].combine_chunks()[0]
+            else:
+                raise ValueError
+        except ValueError:
+            # The pyarrow batches iterator is empty so pass in an empty
+            # iterator and the pyarrow schema
+            return pa.Table.from_batches([], schema=expr.schema().to_pyarrow())
```
kszucs, line 254: Can this only occur for tables?

ibis/backends/duckdb/__init__.py
```diff
@@ -184,6 +185,57 @@ def register(
...
+        table = cursor.cursor.fetch_arrow_table()
+
+        if isinstance(expr, ir.Table):
+            return table
+        elif isinstance(expr, ir.Column):
+            # Column will be a ChunkedArray, `combine_chunks` will
+            # flatten it
+            return table.columns[0].combine_chunks()
+        elif isinstance(expr, ir.Scalar):
+            return table.columns[0].combine_chunks()[0]
```
kszucs, line 235: I think we should be able to directly index into the chunked array, could you try whether
`table.columns[0][0]` works?

ibis/backends/base/sql/__init__.py
```diff
@@ -141,6 +142,55 @@ def raw_sql(self, query: str) -> Any:
...
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        chunk_size: int = 1_000_000,
+        **kwargs: Any,
+    ) -> pa.RecordBatchReader:
+        pa = self._import_pyarrow()
+
+        from ibis.backends.pyarrow.datatypes import ibis_to_pyarrow_struct
+
+        if hasattr(expr, "schema"):
```
kszucs, line 174: I think `to_pyarrow_batches()` should raise if the expression is not table like. I understand that this is used by the other methods, so it's fine for now. We can probably flesh it out when adding support for backends to register their own result types.

ibis/backends/datafusion/__init__.py
```diff
@@ -184,8 +182,43 @@ def execute(
...
                 frame = self._context.empty_table().select(compiled)
-            table = _to_pyarrow_table(frame)
-            return table[0][0].as_py()
+            return frame
+        else:
+            raise com.IbisError(
+                f"Cannot execute expression of type: {type(expr)}"
+            )
+
+    def to_pyarrow_batches(
```
kszucs, line 191: Could you please mark these method as experimental? Perhaps an `@experimental` decorator would we useful similar to `ibis.util.deprecated`.

I'm not convinced yet that the eagerly evaluated API is the preferable. I'm planning to put up a prototype to use specific operations for different outputs so we can compare them, similar to https://github.com/kszucs/ibis/pull/5/files

ibis/backends/base/__init__.py
gforsyth: Nice catch -- adding tests and handling for that now

ibis/backends/duckdb/__init__.py
gforsyth: It does, thanks

cpcloud:
@kszucs If you're happy with this can you merge it?

kszucs (approved):
Thanks @gforsyth! Merging.

PR #4455, opened 2022-09-02
feat(clickhouse): use http API

cpcloud:


cpcloud:
Ok, I've just rebased this PR and want to recap why things are currently
implemented the way they are:

### We have to use JSONEachRow as the output format

Unfortunately we can't use Arrow/Parquet/ORC or anything else that ClickHouse
outputs for IPC that pyarrow can handle. ClickHouse doesn't support
converting its own Enum type to Arrow or Parquet.

The clickhouse playground is **chock full** of enum columns (as an example of
an incredibly useful collection of tables), so "just not supporting them" isn't
really a viable option.

### Even if we could use ArrowStream/Arrow/Parquet we can't really

The main blocker here is that [the setting to turn ClickHouse strings into
Arrow](https://clickhouse.com/docs/en/operations/settings/settings#output_format_arrow_string_as_string)/Parquet
strings just doesn't work at all if the server is running in read-only mode.
This makes it really hard to justify using the HTTP interface despite it being
better in almost every other way that matters to us than clickhouse-driver,
because we now have to handle strings coming back as binary, which is going to
introduce a bunch of whack-a-mole style bugs or UX problems that we currently
do not have.

This did prompt me to make HTTP 500s actually report the server error, so that's a win I guess.

### There's still the pesky problem of columns with strings in their names not working 

See this issue for more detail: https://github.com/ClickHouse/ClickHouse/issues/41112.

PR #4464, opened 2022-09-06
perf(sqlalchemy): register DataFrames cheaply where possible

cpcloud:
This PR registers in-memory tables directly to a backend where native support for it
exists in the backend.

### Backend Support

* DuckDB, new in this PR
* PySpark, already worked
* ClickHouse, works but could be more efficient if we set `use_numpy=True` in the driver settings.

### DIFF
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -186,9 +186,7 @@ class BaseSQLBackend(BaseBackend):
 
         # register all in memory tables if the backend supports cheap access
         # to them
-        if self.compiler.cheap_in_memory_tables:
-            for memtable in lin.traverse(_find_memtables, expr):
-                self._register_in_memory_table(memtable)
+        self._register_in_memory_tables(expr)
 
         with self._safe_raw_sql(sql, **kwargs) as cursor:
             result = self.fetch_from_cursor(cursor, schema)
@@ -200,6 +198,11 @@ class BaseSQLBackend(BaseBackend):
 
     def _register_in_memory_table(self, table_op):
         raise NotImplementedError
+
+    def _register_in_memory_tables(self, expr):
+        if self.compiler.cheap_in_memory_tables:
+            for memtable in lin.traverse(_find_memtables, expr):
+                self._register_in_memory_table(memtable)
 
     @abc.abstractmethod
     def fetch_from_cursor(self, cursor, schema):
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -7,6 +7,7 @@ from typing import Any, Mapping
 
 import sqlalchemy as sa
 
+import ibis.expr.lineage as lin
 import ibis.expr.operations as ops
 import ibis.expr.schema as sch
 import ibis.expr.types as ir
@@ -18,6 +19,11 @@ from ibis.expr.typing import TimeContext
 __all__ = [
     'BaseSQLBackend',
 ]
+
+
+def _find_memtables(expr):
+    op = expr.op()
+    return lin.proceed, op if isinstance(op, ops.InMemoryTable) else None
 
 
 class BaseSQLBackend(BaseBackend):
@@ -178,6 +184,12 @@ class BaseSQLBackend(BaseBackend):
 
         schema = self.ast_schema(query_ast, **kwargs)
 
+        # register all in memory tables if the backend supports cheap access
+        # to them
+        if self.compiler.cheap_in_memory_tables:
+            for memtable in lin.traverse(_find_memtables, expr):
+                self._register_in_memory_table(memtable)
+
         with self._safe_raw_sql(sql, **kwargs) as cursor:
             result = self.fetch_from_cursor(cursor, schema)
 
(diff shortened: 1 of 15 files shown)


ibis/backends/base/sql/__init__.py
```diff
@@ -20,6 +21,11 @@
 ]
 
 
+def _find_memtables(expr):
+    is_in_memory_table = isinstance(op := expr.op(), ops.InMemoryTable)
+    return lin.proceed, op if is_in_memory_table else None
```
jcrist, line 26: Nit, 100% feel free to ignore.

I find the usage of the walrus operator in locations other than `if`/`while` loops kinda hard to read. I'm not used to seeing variables defined inside a call to a function, my brain doesn't expect to look for them there.

```suggestion
    op = expr.op()
    return lin.proceed, op if isinstance(op, ops.InMemoryTable) else None
```

ibis/backends/pandas/client.py
```diff
@@ -301,23 +301,23 @@ def convert_array_to_series(in_dtype, out_dtype, column):
...
-        return getattr(self._df, name)
-
     def __hash__(self):
         return self._hash
 
     def __repr__(self):
         df_repr = util.indent(repr(self._df), spaces=2)
         return f"{self.__class__.__name__}:\n{df_repr}"
 
+    def to_frame(self):
```
jcrist, line 318: I'm not familiar with `DataFrameProxy` objects - are these user facing? If so:

- Perhaps this should be `to_pandas(self)`, mirroring conventions used in other projects (e.g. `pyarrow.Table.to_pandas`)
- Perhaps this should have a docstring

ibis/backends/base/sql/alchemy/query_builder.py
```diff
@@ -109,8 +108,14 @@ def _format_table(self, expr):
             backend._create_temp_view(view=result, definition=definition)
         elif isinstance(ref_op, ops.InMemoryTable):
             columns = _schema_to_sqlalchemy_columns(ref_op.schema)
-            rows = list(ref_op.data.itertuples(index=False))
-            result = sa.values(*columns).data(rows)
+
+            if self.context.compiler.cheap_in_memory_tables:
+                result = sa.table(ref_op.name, *columns)
```
jcrist, line 113: Is `ref_op.name` guaranteed to be unique? Where do these names come from?

ibis/backends/base/sql/__init__.py
```diff
@@ -186,6 +198,9 @@ def execute(
 
         return result
 
+    def _register_in_memory_table(self, table):
+        pass
```
jcrist, line 202: Can you add a brief docstring here for future readers?

Also, this should probably raise `NotImplementedError`, since calling this with a no-op will lead to bugs downstream

```suggestion
        raise NotImplementedError
```

jcrist (commented):
Only some minor comments/questions, otherwise this looks good to me!

ibis/backends/base/sql/alchemy/query_builder.py
cpcloud: They are either generated or user provided. `_ibis_memtable_{a_nonzero_number}` is the template for the generated names.

ibis/backends/pandas/client.py
cpcloud: `DataFrameProxy` is not user facing. I personally prefer `to_frame()` since this is never dealing with Series. I'd like to be able to read the code and know that I'm getting a DataFrame, rather than just any ol' pandas object.

ibis/backends/base/sql/__init__.py
cpcloud: That's fair. I don't feel strongly about this stylistic choice here.

ibis/backends/base/sql/alchemy/query_builder.py
cpcloud: The names aren't guaranteed to be unique if the user passes `name`, but that's true of any other API that we provide that accepts an argument for naming a table.

If no name is passed, then there's potentially overlapping names if a table in the database is called `_ibis_memtable_{number}`. However, since such expressions would hash differently, they'll be unique and aliasing will take care of making sure things are unique except in 2 cases:

1. joining the two tables that have the same name
2. set operations on two tables with the same name

I specifically chose a prefix that I think will be extremely unlikely to collide with anyone's table names. An alternative would be to guarantee uniqueness by using using a UUID, but then the table name is hard to read IMO.

jcrist: No, what you have here is fine - I was just curious how this worked. Thanks!

ibis/backends/duckdb/__init__.py
```diff
@@ -191,6 +191,10 @@ def _get_schema_using_query(self, query: str) -> sch.Schema:
         """Return an ibis Schema from a DuckDB SQL string."""
         return sch.Schema.from_tuples(self._metadata(query))
 
+    def _register_in_memory_table(self, table_op):
+        df = table_op.data.to_frame()
+        self.con.execute("register", (table_op.name, df))
```
Mause, line 196: It might be worth having a conversation about how to best expose functionality like this from duckdb through the sqlalchemy layer? Engine signals perhaps?

PR #4465, opened 2022-09-06
refactor(ir): split datatypes package into multiple submodules

kszucs:


ibis/common/parsing.py
```diff
@@ -0,0 +1,40 @@
...
+
+def spaceless_string(*strings: str):
+    return spaceless(
+        parsy.alt(
+            *(parsy.string(string, transform=str.lower) for string in strings)
+        )
+    )
+
+
+RAW_NUMBER = parsy.digit.at_least(1).concat()
```
cpcloud, line 23: This should really be `parsy.decimal_digit`, since `digit` includes all sorts of things like `²` that Python cannot turn into an `int`:

```
In [6]: import parsy

In [7]: parsy.digit.parse('²')
Out[7]: '²'

In [8]: parsy.decimal_digit.parse('²')
ParseError: expected '[0123456789]' at 0:0

In [9]: parsy.decimal_digit.parse('2')
Out[9]: '2'
```

kszucs: Updated.

ibis/expr/datatypes/core.py
```diff
@@ -128,13 +72,17 @@ def equals(self, other):
             )
         return super().__cached_equals__(other)
 
-    def castable(self, target, **kwargs):
-        """Return whether this data type is castable to `target`."""
-        return castable(self, target, **kwargs)
+    def cast(self, other):
```
kszucs, line 75: Shall we keep the `cast` and `castable` methods?

ibis/expr/datatypes/core.py
```diff
@@ -539,6 +487,8 @@ def __repr__(self):
         return f"{self.name}(cardinality={cardinality})"
 
     def to_integer_type(self):
```
kszucs, line 489: This method doesn't seem to be used anywhere, can we delete it?

ibis/tests/expr/test_value_exprs.py
```diff
@@ -1213,17 +1211,16 @@ def test_map_get_with_compatible_value_bigger():
 
 def test_map_get_with_incompatible_value_different_kind():
     value = ibis.literal({'A': 1000, 'B': 2000})
-    with pytest.raises(IbisTypeError):
-        value.get('C', 3.0)
+    assert value.get('C', 3.0).type() == dt.float64
```
kszucs, line 1214: I changed the semantics here to use the highest precedence one.

cpcloud: We should try this on at least two backends to see if in practice this behavior works.

ibis/expr/datatypes/core.py
kszucs: I defer it to a follow-up issue since I'm unable to run the impala tests locally.

kszucs: https://github.com/ibis-project/ibis/issues/4477

ibis/tests/expr/test_value_exprs.py
kszucs: Addressed it in a follow-up PR https://github.com/ibis-project/ibis/pull/4476

cpcloud (approved):
Since maps don't really have great support across backends and maps haven't seen a lot of love I don't think their change should block this PR.

kszucs:
> Since maps don't really have great support across backends and maps haven't seen a lot of love I don't think their change should block this PR.

There is a follow up https://github.com/ibis-project/ibis/pull/4476 which aims to clean the semantics by properly testing the map operations both for pandas and clickhouse.

PR #4468, opened 2022-09-08
fix(sqlalchemy): implement `SQLQueryResult` translation

cpcloud:
This PR makes SQLQueryResult aka con.sql work for SQLAlchemy backends


jcrist (approved):
LGTM

PR #4479, opened 2022-09-09
ci: load data into bigquery native tables

cpcloud:
This PR uploads our CI data into a bigquery native table instead of relying on the external JSON (which will still be available). This should make querying the data faster.


cpcloud:
not really much to do here except wait for the next `GCS Insert` job to run, so I'm going to merge this.

PR #4482, opened 2022-09-09
feat(api): add `and_` and `or_` helpers

jcrist:
This adds some helper methods for composing multiple predicates together using boolean logic. I think these methods are useful when programmatically combining expressions, but I'm not sold on the names/variadic nature.

- Should these be called `and_`/`or_` or `all`/`any`?
- Should these be variadic or take an iterable?

I went with `and_`/`or_` and variadic for now, but happy to change if others feel strongly.

I do think it's important that these helpers work with 0 or more elements to make it easier to programmatically combine predicates - this lets the caller not worry about the 0 or 1 element case. For no predicates I went with the behavior of python's `all`/`any` methods:

- `and_()` returns `literal(True)`
- `or_()` returns `literal(False)`

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import datetime
 import functools
 import itertools
+import operator
 from typing import Iterable, Mapping, Sequence
 from typing import Tuple as _Tuple
 from typing import TypeVar
@@ -115,7 +116,9 @@ from ibis.expr.window import (
 
 __all__ = (
     'aggregate',
+    'and_',
     'array',
+    'asc',
     'case',
     'coalesce',
     'connect',
@@ -124,7 +127,6 @@ __all__ = (
     'date',
     'desc',
     'difference',
-    'asc',
     'e',
     'Expr',
     'geo_area',
@@ -194,6 +196,7 @@ __all__ = (
     'negate',
     'now',
     'null',
+    'or_',
     'param',
     'pi',
     'prevent_rewrite',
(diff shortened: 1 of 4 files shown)


ibis/expr/api.py
```diff
@@ -543,6 +546,44 @@ def asc(expr: ir.Column | str) -> ir.SortExpr | ops.DeferredSortKey:
...
+
+    Returns
+    -------
+    BooleanValue
+        A new predicate that evaluates to True if any composing predicates are
+        True. If no predicates were provided, returns False.
+    """
+    if not predicates:
+        return literal(False)
+    return functools.reduce(operator.or_, predicates)
```
cpcloud, line 584: Would `functools.reduce(operator.or_, predicates, literal(False))` work? Similarly for `and_`

kszucs:
> Should these be called and_/or_ or all/any?

My preference is `all`/`any`

cpcloud:
-0.5 on `all`/`any` because they overlap with the column reduction names in both Python and elsewhere in our API.

`or_` and `and_` are significantly differently named that I think there's a chance people may read the docstring instead of assuming a particular behavior. I think there's a high likelihood that people will assume `any` and `all` behave like `builtins.any/`builtins.all`.

jcrist:
> I think there's a high likelihood that people will assume `any` and `all` behave like `builtins.any/`builtins.all`.

If we switch to `any`/`all`, I'd make them take an iterable rather than being variadic, so they'd feel more like the builtins `any`/`all`.

Mild agreement about them overlapping with the column reductions, but for me it's more like -0.05. Python users are already used to `builtins.all` and `pandas.Series.all` doing different things.

ibis/expr/api.py
jcrist: If you pass an initial value to `reduce`, it's always used, not just in the case of no elements. This means that these would be equivalent:

```
ibis.or_(a, b, c)

ibis.literal(false) | a | b | c
```

I didn't want `literal(False)`/`literal(True)` hanging around in every generated expression.

cpcloud:
> Python users are already used to builtins.all and pandas.Series.all doing different things.

Hm, I think I'm missing something here. Don't these basically have the same behavior (the pseudo-ternary logic in pandas notwithstanding)?

cpcloud (approved):

PR #4489, opened 2022-09-12
feat(ux): add duckdb as the default backend

cpcloud:
This PR adds DuckDB as the default backend to use for things that don't have a backend.

Right now, those things are:

1. `ops.Literal`s
2. `ops.InMemoryTable`s

This is useful for doing analysis on in-memory things without having to think too hard about
how to do it. The workflow is one call to `ibis.memtable` on the object and you're off to
the races.

### DIFF
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -1,4 +1,5 @@
 import platform
+import re
 
 import pandas as pd
 import pandas.testing as tm
@@ -7,6 +8,7 @@ import sqlalchemy as sa
 from pytest import mark, param
 
 import ibis
+import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
 from ibis.util import guid
 
@@ -661,3 +663,51 @@ def test_create_from_in_memory_table(con, t):
     finally:
         con.drop_table(tmp_name)
         assert tmp_name not in con.list_tables()
+
+
+def test_default_backend_no_duckdb(backend):
+    # backend is used to ensure that this test runs in CI in the setting
+    # where only the dependencies for a a given backend are installed
+
+    # if duckdb is available then this test won't fail and so we skip it
+    try:
+        import duckdb  # noqa: F401
+
+        pytest.skip(
+            "duckdb is installed; it will be used as the default backend"
+        )
+    except ImportError:
+        pass
+
+    df = pd.DataFrame({'a': [1, 2, 3]})
+    t = ibis.memtable(df)
+    expr = t.a.sum()
+
+    # run this twice to ensure that we hit the optimizations in
+    # `_default_backend`
+    for _ in range(2):
+        with pytest.raises(
+            com.IbisError,
+            match="Expression depends on no backends",
+        ):
+            expr.execute()
+
+
+@pytest.mark.duckdb
+def test_default_backend():
+    pytest.importorskip("duckdb")
+
+    df = pd.DataFrame({'a': [1, 2, 3]})
+    t = ibis.memtable(df)
+    expr = t.a.sum()
+    # run this twice to ensure that we hit the optimizations in
+    # `_default_backend`
+    for _ in range(2):
+        assert expr.execute() == df.a.sum()
+
+    sql = ibis.to_sql(expr)
+    rx = """\
+SELECT
+  SUM\\((\\w+)\\.a\\) AS sum
+FROM \\w+ AS \\1"""
+    assert re.match(rx, sql) is not None
--- a/ibis/expr/types/core.py
+++ b/ibis/expr/types/core.py
@@ -7,13 +7,13 @@ from typing import TYPE_CHECKING, Any, Hashable, Mapping
 
 from public import public
 
-from ibis import config
 from ibis.common.exceptions import (
     ExpressionError,
     IbisError,
     IbisTypeError,
     TranslationError,
 )
+from ibis.config import _default_backend, options
 from ibis.expr.typing import TimeContext
 from ibis.util import UnnamedMarker, deprecated
 
@@ -33,7 +33,7 @@ class Expr:
         self._arg = arg
 
     def __repr__(self) -> str:
-        if not config.options.interactive:
+        if not options.interactive:
             return self._repr()
 
         try:
@@ -103,7 +103,7 @@ class Expr:
         return type(self), self._safe_name, self.op()
 
     def _repr_png_(self) -> bytes | None:
-        if config.options.interactive or not config.options.graphviz_repr:
+        if options.interactive or not options.graphviz_repr:
             return None
         try:
             import ibis.expr.visualize as viz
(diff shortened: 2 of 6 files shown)


cpcloud:
I'll change the `default_backend_func` to be `ibis.config._default_backend_func`

ibis/backends/tests/test_client.py
```diff
@@ -661,3 +663,21 @@ def test_create_from_in_memory_table(con, t):
...
+
+
+@pytest.mark.duckdb
+def test_default_backend():
+    pytest.importorskip("duckdb")
+
+    df = pd.DataFrame({'a': [1, 2, 3]})
+    t = ibis.memtable(df)
+    expr = t.a.sum()
+    assert expr.execute() == df.a.sum()
```
kszucs, line 675: What happens when the user tries to execute an expression based on an unbound table?

cpcloud: The execution will fail.

cpcloud: The failure mode isn't great:

```
DBAPIError: (duckdb.Error) Invalid Error: AttributeError: 'function' object has no attribute '_export_to_c'
[SQL: SELECT t0.a
FROM t AS t0
 LIMIT ?]
[parameters: (10000,)]
(Background on this error at: https://sqlalche.me/e/14/dbapi)
```

ibis/expr/types/core.py
```diff
@@ -219,13 +223,34 @@ def _find_backends(self) -> list[BaseBackend]:
...
+        BaseBackend
+            A backend that is attached to the expression
+        """
+        backends, has_unbound = self._find_backends()
 
         if not backends:
-            default = config.options.default_backend
+            if has_unbound:
+                raise IbisError(
+                    "Expression contains unbound tables and therefore cannot "
```
kszucs, line 249: We could be a bit more verbose here and tell the user to `ibis.<backend>.execute()` or set a default backend.

kszucs: Also, could you please cover it with a test case so we ensure that this error reporting remains persistent after the refactors?

cpcloud: Yep!

ibis/backends/tests/test_client.py
cpcloud: This is no longer relevant due to refactoring the code to handle the case where an expression isn't executable.

ibis/expr/types/core.py
cpcloud: Done!

kszucs (approved):

jreback (changes requested):
does this not require an additional import?

cpcloud:
> does this not require an additional import?

Can you clarify where you think an additional import is required?

Also, can you avoid using the "Request Changes" functionality to ask questions and reserve it only for requesting specific changes?

jreback:
this makes duckdb a required import now

i think that is a major breaking change

kszucs:
Could we fall back to pandas if duckdb is not available? Pandas should be a mandatory dependency, at least for now.

jreback:
falling back to pandas seems reasonable though their is a failed import cost constantly if that happens which may not be trivial

jreback:
not averse to in the future having duckdb as a default 

but that needs to be a separate and deliberate decision

cpcloud:
> this makes duckdb a required import now

> i think that is a major breaking change

DuckDB is not a required import and so this isn't a breaking change.

Did you read the code? Did you notice that there are no changes to `pyproject.toml`? Can you demonstrate a scenario where this is a required import? Most of the CI jobs do **not** install duckdb, and yet somehow they continue to pass!

> but that needs to be a separate and deliberate decision

The PR **is** the deliberate decision.

If you can demonstrate a specific verifiable problem then we can continue the discussion. Otherwise, this PR is ready to merge.

jreback:
> The PR is the deliberate decision.

where exactly is this discussion?

this is such a huge change that it needs discussion

and would be -1 at this time

jreback:
further i think you need to prove that this isn't a major performance regression in thr lack of import 

oh and since when do we have a default backend?

cpcloud:
> > The PR is the deliberate decision.
> 
> where exactly is this discussion?

There's no requirement to have discussion before putting up a PR, that would be ridiculously prohibitive.

If anyone has actionable objections they are welcome to describe them and comment on the PR.

> 
> this is such a huge change that it needs discussion
> 
> and would be -1 at this time

You still haven't demonstrated what the problem is. Again, **there is no requirement to have duckdb installed**. Can you describe a scenario where the import is required?

jreback:
you are skipping the test if it's not installed

cpcloud:
I've added tests that cover both cases, and also made it cheap to fail: the import failure overhead is incurred once per process.

After everything's green, I'm going to merge the PR.

cpcloud:
@jreback

In general, one person's -1 is not enough to tank a PR, especially if it's already been approved. If you want to start contributing again, then your input will carry more weight.

I know that you've contributed a lot in the past, but continued interaction in some other form besides blocking PRs is what carries weight.

This applies to anyone who wants to contribute.

Adding these tests and performance enhancements is a courtesy and I will not be doing things like this going forward.

jreback:
@cpcloud that is not a great attitude at all

you are aware we heavily rely on ibis and are in collaboration with you and others

we deeply care about the breaking changes and is it important to telegraph these and not simply override eveyone

-1 but a core maintainer should trigger discussion otherwise this is not open source

cpcloud:
> you are aware we heavily rely on ibis and are in collaboration with you and others

I am very aware of that.

We have spent a lot of time and effort accommodating your requests to avoid going to 4.0.0 immediately after 3.1.0 to avoid a large swatch of breaking changes. We have done so by splitting our development into two separate branches, causing a lot of additional unnecessary rebase work to be done.

We did that because we value users' input and decided it was in the best interest of the project to give you more time to catch up.

I think there's been plenty of good will from our side.

All I'm asking for is a larger quantity of and higher quality maintenance work from you, in the form of actions other than blocking reviews with questions that no reasonable person can understand. We aren't going to try and guess what your concerns are.

> we deeply care about the breaking changes and is it important to telegraph these and not simply override eveyone

We also care about breaking changes otherwise we'd have released 4.0 a long time ago, but this isn't a breaking change, so there's not much to telegraph beyond a release note in the Features section of the release notes for 3.2.

> -1 but a core maintainer should trigger discussion otherwise this is not open source

Discussion is highly encouraged. Blocking PRs with a single nondescript question is not.

cpcloud:
Alright, this has been approved and contains no breaking changes that we are aware of, with the rationale explicitly stated above, so I'm going to merge it.

We're happy to address fixes for this if they are needed in a subsequent release and we'll create patch releases for that from a branch.

After this is green, I'll cut the 3.2.0 release 🎉

jreback:
still beating a dead horse with my objection that this is semantically changing execution 

we now can execute inbound tables implicitly - this breaks a tenant of ibis that execution must specify a backend

this is actually a slippery slope

who's to say that we should just always default to x engine and try to execute

i think this is bad user semantics and is too implicit with a very narrow good case (eg that is what u want)

the shouting is not helping anyone here and is distracting

PR #4493, opened 2022-09-12
refactor(ir): remove pydantic dependency and make grounds more composable

kszucs:


ibis/backends/base/sql/compiler/query_builder.py
```diff
@@ -232,6 +232,9 @@ def _translate(self, expr, named=False, permit_subquery=False):
         )
         return translator.get_result()
 
+    def __hash__(self):
```
cpcloud, line 235: Why does this need to be hashable?

ibis/common/grounds.py
```diff
@@ -117,78 +106,74 @@ def __new__(metacls, clsname, bases, dct):
...
             object.__setattr__(self, name, prop(self))
 
-        # any supplemental custom code provided by descendant classes
-        self.__post_init__()
-
-    def __post_init__(self):
-        pass
+    def __setattr__(self, name, value):
+        param = self.__signature__.parameters[name]
+        if not (param.default is None and value is None):
```
cpcloud, line 133:
```suggestion
        if param.default is not None or value is not None:
```

I personally find it difficult to read `not (complicated boolean expression)` versus its De Morgan equivalent without the initial `not`.

ibis/common/grounds.py
```diff
@@ -245,3 +231,31 @@ def __cached_equals__(self, other):
...
+
+    def __equals__(self, other):
+        return self.__args__ == other.__args__
+
+    @property
+    def args(self):
+        return self.__args__
+
+    @property
+    def argnames(self):
```
cpcloud, line 260: Can we have a property that zips `argnames` and `args` together? This is a common enough pattern that we should make it convenient IMO.

ibis/expr/datatypes/core.py
```diff
@@ -57,13 +53,14 @@ def name(self) -> str:
         """Return the name of the data type."""
         return self.__class__.__name__
 
+    # TODO(kszucs): remove it in favor of copy()
```
cpcloud, line 56: Seems fine to keep this around, it's not doing any harm or causing extra work.

ibis/common/grounds.py
kszucs: I added an `as_dict()` method to `Annotable`: https://github.com/ibis-project/ibis/pull/4493/files#diff-1ba248744f86eb25b4e863648c19353294f34c47b637d1fe88d66eccfc1e1119R152

Perhaps it would be better just to return with the items view: `zip(self.__argnames__, self.__args__)`

ibis/common/grounds.py
kszucs: Going to update.

ibis/backends/base/sql/compiler/query_builder.py
kszucs: Because the object must be hashable now to be a subclass of comparable.

ibis/common/grounds.py
cpcloud: Yeah, it's easy to just call `dict` IMO

ibis/backends/base/sql/compiler/query_builder.py
cpcloud: Got it, thanks for explaining.

ibis/common/grounds.py
cpcloud: I'm okay with `as_dict` for now.

ibis/backends/base/sql/compiler/query_builder.py
kszucs: Actually I added defailt implementation for `Comparable.__hash__()` now to defer it to `super().__hash__()`

cpcloud (approved):
Sweet, thanks! LGTM.

kszucs:
I forgot to add a couple of test cases, let me add them.

kszucs:
@cpcloud could you please review it again?

PR #4495, opened 2022-09-13
fix(duckdb): add missing types

### LINKED ISSUE #4481: bug/feat: errors querying remote parquet with Ibis via DuckDB backend
Hello,

I've been experimenting with Ibis as a way to query Parquet files in cloud storage via the DuckDB backend, but have been running into errors like `Python Object module not suitable for replacement scans` and `ParseError: expected one of 'EOF', '[' at 0:9`. Code sample to reproduce below, the cloud files are public access so anyone can query them.

I also tried to do something similar via the DataFusion backend but was unable to figure out how to connect to a cloud storage filesystem via Ibis. (Does it only work with local parquet files?)

Is this a bug? Should I be able to query DuckDB views over Parquet files with Ibis? If not, what is the recommended way to query parquet files in cloud storage with Ibis? Could someone please write a code sample?

```python
import adlfs
import duckdb
import ibis
from pyarrow import dataset

# Example 1:
fs = adlfs.AzureBlobFileSystem(account_name="azureopendatastorage", sas_token="", container_name="mlsamples")
dt = dataset.dataset("mlsamples/diabetes", filesystem=fs)
conn = duckdb.connect("diabetes.db")
conn.register("diabetes", dt)
conn.view("diabetes").aggregate("SEX, count(*) AS ct").df()
# works

ibis_conn = ibis.duckdb.connect("diabetes.db")
ibis_conn.raw_sql("create view diabetes as select * from dt")
diabetes = ibis_conn.table("diabetes")
diabetes.group_by(diabetes.SEX).count().execute()
# RuntimeError: Python Object module not suitable for replacement scans

# Example 2:
fs = adlfs.AzureBlobFileSystem(account_name="azureopendatastorage", sas_token="", container_name="holidaydatacontainer")
dt = dataset.dataset("holidaydatacontainer/Processed", filesystem=fs)
conn = duckdb.connect("holidays.db")
conn.register("holidays", dt)
conn.view("holidays").aggregate("CountryOrRegion, count(*) AS ct").df()
# works

ibis_conn = ibis.duckdb.connect("holidays.db")
ibis_conn.raw_sql("create view holidays as select * from dt")
holidays = ibis_conn.table("holidays")
# ParseError: expected one of 'EOF', '[' at 0:9

### LINKED ISSUE #4491: bug(duckdb): handle `TIMESTAMP_NS` parsing

### PR DESCRIPTION
**Pull request by cpcloud**
This PR adds missing DuckDB types. Closes #4481. Closes #4491. We can address handling any unknown type in a separate PR.

### CODE REVIEW

File: ibis/backends/duckdb/datatypes.py [hunk 1]
```diff
@@ -63,9 +64,14 @@ def parse(text: str, default_decimal_parameters=(18, 3)) -> DataType:
         | spaceless_string("double", "float8").result(float64)
         | spaceless_string("real", "float4", "float").result(float32)
         | spaceless_string("smallint", "int2", "short").result(int16)
-        | spaceless_string("timestamp", "datetime").result(
-            Timestamp(timezone="UTC")
-        )
+        | spaceless_string(
+            "timestamp_tz",
+            "timestamp_sec",
+            "timestamp_ms",
```
Mause, line 70: Not preserving the scale in the conversion at all?

cpcloud: We don't have a way to do that yet in ibis. Currently, it doesn't have any meaningful affect on the APIs we present to users, so it shouldn't be a huge deal. We are considering adding an optional resolution/scale property to our `Timestamp` type for better compatibility with backends that support it but that'll be a separate PR.

cpcloud: The plan is to have something similar to the way our `Decimal` type works: backends that care about scale can use/create such information in the type, otherwise the scale value is `None` meaning no scale or unknown scale.

PR #4498, opened 2022-09-13
feat(api): add `ibis.to_sql` to easily get a formatted SQL string

cpcloud:
This PR adds a top-level `.sql` API for getting the formatted SQL string from an ibis expression.

### DIFF
--- a/docs/api/expressions/top_level.md
+++ b/docs/api/expressions/top_level.md
@@ -28,7 +28,7 @@ These methods and objects are available directly in the `ibis` module.
 ::: ibis.or_
 ::: ibis.param
 ::: ibis.show_sql
-::: ibis.sql
+::: ibis.to_sql
 ::: ibis.random
 ::: ibis.range_window
 ::: ibis.row_number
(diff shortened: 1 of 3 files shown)


kszucs:
@cpcloud can you add a small test case?

ibis/common/pretty.py
```diff
@@ -49,6 +49,24 @@ def show_sql(
       t0.a * 2 AS c
     FROM t AS t0
     """
+    print(sql(expr, dialect=dialect), file=file)
+
+
+def sql(expr: ir.Expr, dialect: str | None = None) -> str:
```
jcrist, line 55: I'm a bit worried about this being conflated with `Backend.sql` - could we call it `to_sql` instead?

cpcloud: Sure.

docs/api/expressions/top_level.md
```diff
@@ -28,6 +28,7 @@ These methods and objects are available directly in the `ibis` module.
 ::: ibis.or_
 ::: ibis.param
 ::: ibis.show_sql
+::: ibis.sql
```
gforsyth, line 31:
```suggestion
::: ibis.to_sql
```

saulpw:
I'm not sure about `.sql()` as a name for this, it feels too generic and I can see myself getting confused as to why `.raw_sql()` means "use this sql" while `.sql()` means "return the sql()".

cpcloud:
@saulpw I changed it `to_sql`.

One thing to note is that `.to_sql` is not a method backends. Does that help clear anything up?

Issue #4505, opened 2022-09-15
feat: support amending/overriding reflected types in SQLite backend


jcrist:
I ran into a few problems trying to load data from [PyPI-Data](https://github.com/sethmlarson/pypi-data) that I'm not sure how best to solve:

- The database schemas use `STRING` instead of `TEXT` for string types (this has since been fixed, but [doesn't seem to be part of the latest release](https://github.com/sethmlarson/pypi-data/issues/12)). It would be nice if we could just force sqlalchemy to treat these columns as `TEXT` types rather than falling back to `NUMERIC`. The SQLAlchemy docs illustrate one way of [overriding reflected types](https://docs.sqlalchemy.org/en/14/core/reflection.html#overriding-reflected-columns), but like most things in SQLAlchemy there are probably other ways of handling this.
- The timestamp columns are using RFC3339 format (`YYYY-mm-ddTHH:MM:SS`) which isn't handled by SQLAlchemy's standard datetime parsing routine (it doesn't like the `T` delimiter). The way around this in SQLAlchemy is to pass [additional options to the DATETIME type constructor](https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#sqlalchemy.dialects.sqlite.DATETIME), but ibis currently AFAICT doesn't expose this functionality.

I'm honestly not sure what a good API here would be.

Brainstorming a bit, perhaps additional arguments to the `ibis.sqlite.connect` providing:

- A type mapping from SQLITE type (e.g. `TIMESTAMP`) to ibis type (`typemap={...}`?). This would let us map `{"STRING": ibis.expr.datatypes.String()}`
- A datetime format string in strftime syntax to use when parsing `TIMESTAMP` columns to datetime objects (`timestamp_format=...`?). This assumes the database uses the same datetime format for all tables, but hopefully that's the case.

A few things to like about this approach:

- It doesn't expose that SQLAlchemy is used to the user anywhere. Everything is a mapping between SQLite concepts and ibis itself. This would give us the freedom to drop SQLAlchemy if we wanted without forcing an API change.
- It doesn't require listing the full table schemas for all poorly typed columns - the user only needs to pass in overrides for the specific SQLite type names that require it.

For my use case the connect call might look like:

```python
con = ibis.sqlite.connect("pypi.db", typemap={"STRING": ibis.expr.datatypes.String()}, timestamp_format="%Y-%m-%dT%H:%M:%S")
```

There's probably a better way of handling this though.

[...]

PR #4507, opened 2022-09-15
feat(sqlite): support wider range of datetime formats

jcrist:
This lets ibis load `TIMESTAMP` columns containing a wider range of the string formats accepted by SQLite. This changes nothing at the SQL execution level, just when pulling results back into the Python process.

Formats 1-7 in https://sqlite.org/lang_datefunc.html#time_values are accepted, with optional timezone suffixes.

Fixes the timestamp handling part of #4505.

### DIFF
--- a/ibis/backends/sqlite/tests/test_types.py
+++ b/ibis/backends/sqlite/tests/test_types.py
@@ -1,0 +1,61 @@
+import sqlite3
+
+import pandas as pd
+import pytest
+
+import ibis
+import ibis.expr.datatypes as dt
+from ibis.backends.sqlite import to_datetime
+
+# Test with formats 1-7 (with T, Z, and offset modifiers) from:
+# https://sqlite.org/lang_datefunc.html#time_values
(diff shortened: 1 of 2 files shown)


ibis/backends/sqlite/tests/test_types.py
```diff
@@ -0,0 +1,76 @@
...
+import sqlite3
+
+import pandas as pd
+import pytest
+
+import ibis
+import ibis.expr.datatypes as dt
+
+
+def to_datetime(value):
```
cpcloud, line 11: It seems fine to use the function from `ibis.backends.sqlite` instead of duplicating it here.

cpcloud (approved):
Small nit about duplicating the `to_datetime` function, otherwise this LGTM.

ibis/backends/sqlite/tests/test_types.py
jcrist: Done.

PR #4508, opened 2022-09-15
feat(sqlite): add `type_map` arg to override type reflection

### LINKED ISSUE #4505: feat: support amending/overriding reflected types in SQLite backend

jcrist:
This adds a new `type_map` kwarg to `ibis.sqlite.connect` for overriding type reflection. Since SQLite doesn't have string types, people use all sorts of names for their types, and not all of them are handled properly by ibis/sqlalchemy.

For example, a `STRING` SQLite type will be inferred by sqlalchemy as a `NUMERIC` type (based on the [type affinity rules](https://www.sqlite.org/datatype3.html#type_affinity)) - and `ibis` will then treat this incorrectly as a `Decimal` type.

To work around this, users can pass in a `type_map` option. This takes a mapping between SQLite types (as strings) and the corresponding ibis types they map to.

For example:

```python
con = ibis.sqlite.connect("some.db", type_map={"STRING": "string"})
```

~Builds on #4507~, fixes #4505.

### DIFF
--- a/ibis/backends/sqlite/__init__.py
+++ b/ibis/backends/sqlite/__init__.py
@@ -26,11 +26,13 @@ from sqlalchemy.dialects.sqlite import DATETIME, TIMESTAMP
 
 if TYPE_CHECKING:
     import ibis.expr.types as ir
+    import ibis.expr.datatypes as dt
 
 from ibis.backends.base import Database
-from ibis.backends.base.sql.alchemy import BaseAlchemyBackend
+from ibis.backends.base.sql.alchemy import BaseAlchemyBackend, to_sqla_type
 from ibis.backends.sqlite import udf
 from ibis.backends.sqlite.compiler import SQLiteCompiler
+from ibis.expr.schema import datatype
 
 
 def to_datetime(value: str | None) -> datetime.datetime | None:
@@ -86,6 +88,7 @@ class Backend(BaseAlchemyBackend):
         self,
         database: str | Path | None = None,
         path: str | Path | None = None,
+        type_map: dict[str, str | dt.DataType] | None = None,
     ) -> None:
         """Create an Ibis client connected to a SQLite database.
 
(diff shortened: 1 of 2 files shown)


jcrist:
@cpcloud this should be ready for review.

cpcloud (approved):

PR #4513, opened 2022-09-16
feat(api): make `Table.drop` variadic

jcrist:
This was a minor rough edge found while using `ibis`. I kept calling `drop` incorrectly since I expected it to be variadic like `select` (viewing it as the inverse of select for column subsetting).

This PR isn't a breaking change, the old calling style is deprecated in favor of a variadic call.

```python
In [1]: import ibis

In [2]: t = ibis.table(dict.fromkeys("abcd", "int"))

In [3]: t.select("a", "b").columns
Out[3]: ['a', 'b']

In [4]: t.drop("c", "d").columns
Out[4]: ['a', 'b']
```

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -5,6 +5,7 @@ import functools
 import itertools
 import operator
 import sys
+import warnings
 from functools import cached_property
 from typing import IO, TYPE_CHECKING, Any, Iterable, Literal, Mapping, Sequence
 
@@ -685,7 +686,7 @@ class Table(Expr):
 
         return self.select(exprs)
 
-    def drop(self, fields: str | Sequence[str]) -> Table:
+    def drop(self, *fields: str) -> Table:
         """Remove fields from a table.
 
         Parameters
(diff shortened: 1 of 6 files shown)


jcrist:
This took about as long to write up as an issue would take to create. Happy to discuss if we even want this change here.

cpcloud (approved):
LGTM. Makes sense to have `drop` match `select`.

ibis/expr/types/relations.py
```diff
@@ -696,13 +697,19 @@ def drop(self, fields: str | Sequence[str]) -> Table:
...
         if not fields:
             # no-op if nothing to be dropped
             return self
 
-        fields = util.promote_list(fields)
+        if len(fields) == 1 and not isinstance(fields[0], str):
+            fields = util.promote_list(fields[0])
+            warnings.warn(
+                "Passing a sequence of fields to `drop` is deprecated, "
+                "use `drop(*fields)` instead",
```
kszucs, line 710: Shall we mark a version here?

jcrist: Happy to if we know when the deprecation cycle should complete. I'm not sure what our deprecation policies are.

kszucs: I think removing it in 5.0 would be fine.

jcrist: Done.

PR #4514, opened 2022-09-16
fix: make `col.day_of_week` not an expr

jcrist:
Previously `table.col.day_of_week` was an `Expr`. This didn't really make sense: it couldn't ever be directly computed, and only served as a namespace to store more specific methods (e.g. `col.day_of_week.full_name()`. However, having it as an expression made it easy to accidentally pass on to other expressions, leading to cryptic errors later.

This makes it not an `Expr`, removing those issues. This shouldn't be a breaking change, since any treatment of this node as an expression would have errored (just more cryptically) before.

### DIFF
--- a/ibis/expr/operations/temporal.py
+++ b/ibis/expr/operations/temporal.py
@@ -7,7 +7,7 @@ import ibis.expr.datatypes as dt
 import ibis.expr.rules as rlz
 from ibis import util
 from ibis.common.validators import immutable_property
-from ibis.expr.operations.core import Binary, Node, Unary, Value
+from ibis.expr.operations.core import Binary, Unary, Value
 from ibis.expr.operations.generic import Cast
 from ibis.expr.operations.logical import Between
 
@@ -205,16 +205,6 @@ class DayOfWeekIndex(Unary):
 class DayOfWeekName(Unary):
     arg = rlz.one_of([rlz.date, rlz.timestamp])
     output_dtype = dt.string
-
-
-@public
-class DayOfWeekNode(Node):
-    arg = rlz.one_of([rlz.date, rlz.timestamp])
-
-    def to_expr(self):
-        import ibis.expr.types as ir
-
-        return ir.DayOfWeek(self)
 
 
 @public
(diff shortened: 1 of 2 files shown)


jcrist:
Another spurious test failure due to rate limiting: https://github.com/ibis-project/ibis/actions/runs/3069710492/jobs/4958672734

ibis/expr/types/temporal.py
```diff
@@ -578,7 +578,12 @@ class IntervalColumn(Column, IntervalValue):
 
 
 @public
-class DayOfWeek(Expr):
+class DayOfWeek:
+    """A namespace of methods for extracting day of week information."""
+
+    def __init__(self, expr):
+        self._expr = expr
```
kszucs, line 585: Could we generalize this? I think this kind of objects are going to be more frequest as the API evolves. There is similar class for the case API at https://github.com/ibis-project/ibis/blob/master/ibis/expr/builders.py

jcrist: What benefits would generalizing this bring? What is there to generalize?

kszucs: Nothing in this particular example. Just thinking loudly on how can we extend the API's functionality in the future.

ibis/expr/types/temporal.py
```diff
@@ -578,7 +578,12 @@ class IntervalColumn(Column, IntervalValue):
 
 
 @public
```
kszucs, line 580: Does this need to be public?

ibis/expr/operations/temporal.py
```diff
@@ -207,16 +207,6 @@ class DayOfWeekName(Unary):
     output_dtype = dt.string
 
 
-@public
-class DayOfWeekNode(Node):
-    arg = rlz.one_of([rlz.date, rlz.timestamp])
-
-    def to_expr(self):
```
kszucs, line 214: I wonder whether we have similar "unused" nodes left, `def to_expr()` should indicate a custom behaviour.

ibis/expr/types/temporal.py
jcrist: Probably not, but it was before.

cpcloud (approved):
LGTM.

PR #4521, opened 2022-09-16
docs: Correct small typo in aggregates-joins

ksuarez1423:
In the joining data section, the prose claims that the GDP country code corresponds to iso_alpha2, when it actually corresponds to iso_alpha3, which is reflected in the code. This PR fixes that.


cpcloud (approved):
Thanks!

PR #4525, opened 2022-09-17
feat(api): add JSON getitem support

cpcloud:
This PR adds the ability to extract object values or array elements of a JSON-typed column.

Notable things:

- ClickHouse support for this is experimental, and doesn't seem to work in ClickHouse itself and is also broken/unsupported in `clickhouse_driver`.
- PySpark works, but since we're not handling result conversion at all really the test fails.
- The SQLAlchemy backends basically all "just work", the main differences being how the `JSON` type is return in Python. We handle this here by trying to parse the JSON and if that fails assuming the value is already deserialized.

### DIFF
--- a/ibis/backends/clickhouse/tests/conftest.py
+++ b/ibis/backends/clickhouse/tests/conftest.py
@@ -27,6 +27,7 @@ class TestConf(UnorderedComparator, BackendTest, RoundHalfToEven):
     supported_to_timestamp_units = {'s'}
     supports_floating_modulus = False
     bool_is_int = True
+    supports_json = False
 
     @staticmethod
     def _load_data(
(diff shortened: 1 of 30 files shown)


ibis/expr/types/json.py
```diff
@@ -1,34 +1,36 @@
...
-from ibis.expr.types.binary import BinaryColumn, BinaryScalar, BinaryValue
-from ibis.expr.types.strings import StringColumn, StringScalar, StringValue
+from ibis.expr.types import Column, Scalar, Value
 
 
 @public
-class JSONValue(StringValue):
-    pass  # noqa: E701,E302
+class JSONValue(Value):
+    def __getitem__(self, key):
```
kszucs, line 8: Could you please add a follow-up to provide `.get(key, default)` API similar to maps?

ibis/backends/clickhouse/tests/conftest.py
```diff
@@ -27,6 +27,7 @@ class TestConf(UnorderedComparator, BackendTest, RoundHalfToEven):
     supported_to_timestamp_units = {'s'}
     supports_floating_modulus = False
     bool_is_int = True
+    supports_json = False
```
kszucs, line 30: Please add a follow-up issue to investigate it further.

kszucs (approved):

PR #4529, opened 2022-09-18
refactor(ir): remove `ops.DeferredSortKey`

cpcloud:
This commit removes `ops.DeferredSortKey` in favor using `Deferred`
objects. A sort direction is now rendered in every case, to be explicit
about sorting in the case that a backend has a configurable default sort
order.

BREAKING CHANGE: `ops.DeferredSortKey` is removed. Use `ops.SortKey` directly instead.


ibis/expr/deferred.py
```diff
@@ -20,6 +20,11 @@ def __init__(self, resolve: Callable = toolz.identity) -> None:
         ), f"resolve argument is not callable, got {type(resolve)}"
         self.resolve = resolve
 
+    def __hash__(self) -> int:
```
kszucs, line 23: I guess this is the default implementation by python, need to verify though.

cpcloud: It's only the default if you leave `__eq__` and `__hash__` **both** not implemented:


```
In [1]: class Foo:
   ...:     def __eq__(self, other):
   ...:         return True
   ...:

In [2]: hash(Foo())
Cell In [2], line 1
----> 1 hash(Foo())

TypeError: unhashable type: 'Foo'
```

ibis/expr/api.py
```diff
@@ -542,13 +544,15 @@ def asc(expr: ir.Column | str) -> ir.SortExpr | ops.DeferredSortKey:
...
     Returns
     -------
-    ir.SortExpr | ops.DeferredSortKey
-        A sort expression or deferred sort key
+    ir.ValueExpr
+        An expression
     """
-    if isinstance(expr, Expr):
-        return ops.SortKey(expr, ascending=True).to_expr()
+    if isinstance(expr, str):
```
kszucs, line 550: Nice!

ibis/expr/deferred.py
cpcloud: We override `__eq__` to have the expression behavior, so we have to override `__hash__` to get any kind of hashing at al.

kszucs (approved):
Much cleaner, thanks!

kszucs:
@cpcloud could you please highlight the breaking change in the commit message and and `ir` as scope?

PR #4530, opened 2022-09-18
perf: fix join performance by avoiding Projection construction

cpcloud:
This avoids the large number of repeated traversals on construction.


kszucs:
Do we have a benchmark to catch this regression?

cpcloud:
Yes, it's been altering us, but we decided to ignore it until after behemoth landed.

kszucs (approved):

PR #4531, opened 2022-09-18
fix(visualize): dedup nodes and edges and add `verbose` argument for debugging

### LINKED ISSUE #4528: visualize.py `to_graph(...)` adds same node multiple times to graphviz graph AND doesn't display all nodes
When running the "example" from:
https://github.com/ibis-project/ibis/blob/9edbbc0cc22fc8d2db8c5e5614c6c5a0c40c33a0/ibis/expr/visualize.py#L124-L138

...and setting a breakpoint at the end of the `to_graph(...)`
https://github.com/ibis-project/ibis/blob/9edbbc0cc22fc8d2db8c5e5614c6c5a0c40c33a0/ibis/expr/visualize.py#L75-L106

1) You will see that the graphviz `DOT code` (aka `print(g.source)`) generated by the `graphviz` lib via `ibis/ibis/expr/visualize.py`s' `to_graph(...)` contains duplicated nodes (highlighted with red in below debugging screenshot).

image

These duplicated nodes do NOT impact the visual representation because they have the same "node id" and **graphviz basically takes into consideration only the LAST declaration of a node**, `BUT` they impact a bit the performance:

```
the same `node` should be present (added/"seen") only once in the graphviz graph.
```

_The same `edge` (from, to) can be present more than once if we want to see it duplicated._

One solution would be to check if the node has already been added to the `g` graph BEFORE adding it, BUT...

2. ....Also there seem to be different nodes with the same id (with green in the screenshot above), which can IMPACT the visual representation, ex:
```
1st appearance:
-7147190702382226628 [label=<<I>having</I>: <B>NodeList</B> ⟶ ∅>]
...
2nd appearance with:
-7147190702382226628 [label=<<I>sort_keys</I>: <B>NodeList</B> ⟶ ∅>]
```
Reusing the same node ID for a different subsequent node _- (at least) before creating an edge with that ID -_ will cause the 1st node not to be rendered -> ex. indeed searching for "having" in the graphviz svg of the above-mentioned example will yield NO RESULTS (`"having" is missing`).

I'll try to investigate why is `hash(...)` returning the same thing for different instances and get back to you...

cpcloud:
Closes #4528.

### DIFF
--- a/ibis/expr/visualize.py
+++ b/ibis/expr/visualize.py
@@ -1,3 +1,4 @@
+import sys
 import tempfile
 from html import escape
 
@@ -6,7 +7,6 @@ import graphviz as gv
 import ibis
 import ibis.common.exceptions as com
 import ibis.expr.operations as ops
-import ibis.expr.types as ir
 
 
 def get_type(node):
@@ -41,7 +41,7 @@ def get_type(node):
 
     return (
         ''.join(
-            '<BR ALIGN="LEFT" />  <I>{}</I>: {}'.format(
+            '<BR ALIGN="LEFT" /> <I>{}</I>: {}'.format(
                 escape(name), escape(str(type))
             )
             for name, type in zip(schema.names, schema.types)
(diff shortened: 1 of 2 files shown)


cpcloud:
Here's an example of the new graph.

out

cpcloud:
Just so we can avoid lots of back and forth in review:

There are many, many ways the graphviz representation can be improved.

I won't be implementing any of those improvements in this PR. This PR is scoped to fixing the duplicated node id issue and reusing the `ibis.util.to_op_dag` functionality instead of doing yet another traversal.

ibis/expr/visualize.py
```diff
@@ -72,41 +78,44 @@ def get_label(node, argname=None):
...
+        for u in us:
+            if isinstance(u, ops.NodeList) and not u:
+                continue
+
+            uhash = str(hash(u))
+            if u not in seen:
+                g.node(uhash, label=get_label(u))
+                seen.add(u)
+            if (edge := (u, v)) not in edges:
+                g.edge(uhash, vhash)
```
ams1, line 111: maybe show arg names as edge lables
```suggestion
                arg_name = None
                if not isinstance(v, ops.NodeList):
                    arg_name = v.argnames[v.args.index(u)]
                g.edge(uhash, vhash, label=arg_name)
```

cpcloud: Ah interesting, I was thinking about how to show that but couldn't come up with anything reasonable. Let me see what this looks like!

ams1:
For me seeing the `args` helps.

Below img:
1) shows the argnames as edge lables 

2) shows nodes having the return on the next line
```py
def get_label(node):
    typename = get_type(node)  # Already an escaped string
    name = type(node).__name__
    nodename = (
        node.name
        if isinstance(
            node, (ops.Literal, ops.TableColumn, ops.Alias, ops.PhysicalTable)
        )
        else None
    )
    if nodename is not None:
        if isinstance(node, ops.TableNode):
            label_fmt = '<<I>{}</I>: <B>{}</B>{}>'
        else:
            label_fmt = (
                '<<TABLE CELLPADDING="0" CELLSPACING="0" BORDER="0">'
                '<TR><TD><I>{}</I>:<B>{}</B></TD></TR>'
                '<TR><TD>{}</TD></TR>'
                '</TABLE>>'
            )
        label = label_fmt.format(escape(nodename), escape(name), typename)
    else:
        if isinstance(node, ops.TableNode):
            label_fmt = '<<B>{}</B>{}>'
        else:
            label_fmt = (
                '<<TABLE CELLPADDING="0" CELLSPACING="0" BORDER="0">'
                '<TR><TD><B>{}</B></TD></TR>'
                '<TR><TD>{}</TD></TR>'
                '</TABLE>>'
            )
        label = label_fmt.format(escape(name), typename)
    return label
```

image

cpcloud:
@ams1 Sweet, the latest commit adds edge labels!

jreback:
would make edge labels optional if u can

cpcloud:
@jreback Thanks for the input. I've made edge labels optional and the default value for the argument to enable them is `False`, which preserves the existing label-less edges :)

cpcloud:
With edge labels:

tmpfps5gw39

Without edge labels:

tmphdus24xf

jreback:
cool ty

kszucs (approved):
LGTM

Issue #4335, opened 2022-08-09
feat(api): generate Python code from an ibis expression


jcrist:
Given an `ibis` expression, it'd be useful to be able to generate Python code that results in the same expression.

Example user session:

```python
import ibis


con = ibis.sqlite.connect("geography.db")

t = con.tables["countries"]

asian_countries = t['name', 'continent', 'population'].filter(t.continent == 'AS')
result = asian_countries.sort_by((asian_countries.population, False)).limit(10)


code = ibis.util.to_python_code(result)
print(code)
# output
"""
countries = con.tables["countries"]

temp1 = countries['name', 'continent', 'population'].filter(countries["continent"] == 'AS')
result = temp1.sort_by((temp1.population, False)).limit(10)
"""
```

Notes:
- Given an expression, we can know the table names involved, but not the connection variable name. A good default behavior here might be to default to `con`, and make the connection name configurable via `connection_varname="foo"`.
- We can generate readable variable names for initial tables based on the table name
- We will sometimes need to generate intermediate expressions. Here I go with `temp%d` with an incrementing suffix
- We need to provide a variable name for the output expression. Here I use `result` - this could also be configurable, idk

I don't expect this to be used in a 100% automated fashion, rather I expect this to be a useful tool with a human-in-the-loop. I also don't expect the generated code to match 1:1 with how a human would write it. I view this mostly as a starting point that a human can then edit further.

Possible use cases:

- Given #4324, we could then print out ibis code for generating those expressions.
- Sometimes in an interactive session you work non-linearly, defining and redefining temporary variables until you get the expected result. It may be tricky to backtrack back to how you came up with the correct expression, but since you have the full expression tree ibis should be able to output some example code that would regenerate it.

saulpw:
I think this is a great idea!  I would like it if the output were at least nominally executable immediately; like if called with `exec()` it defined a well-known function that `eval()` could call to return the expression.

PR #4535, opened 2022-09-19
feat(decompile): support rendering ibis expression as python code

kszucs:
depends on #4493 

Closes https://github.com/ibis-project/ibis/issues/4335.

### DIFF
--- a/ibis/expr/decompile.py
+++ b/ibis/expr/decompile.py
@@ -1,0 +1,422 @@
+import collections
+import functools
+import io
+import itertools
+
+import ibis
+import ibis.expr.datatypes as dt
+import ibis.expr.operations as ops
+import ibis.expr.schema as sch
+import ibis.expr.types as ir
+from ibis.common.graph import Graph
+from ibis.util import experimental
+
+_method_overrides = {
+    ops.CountDistinct: "nunique",
+    ops.CountStar: "count",
+    ops.EndsWith: "endswith",
+    ops.ExtractDay: "day",
+    ops.ExtractDayOfYear: "day_of_year",
+    ops.ExtractEpochSeconds: "epoch_seconds",
+    ops.ExtractHour: "hour",
+    ops.ExtractMillisecond: "millisecond",
+    ops.ExtractMinute: "minute",
+    ops.ExtractMinute: "minute",
+    ops.ExtractMonth: "month",
+    ops.ExtractQuarter: "quarter",
+    ops.ExtractSecond: "second",
+    ops.ExtractWeekOfYear: "week_of_year",
+    ops.ExtractYear: "year",
+    ops.Intersection: "intersect",
+    ops.IsNull: "isnull",
+    ops.LeftAntiJoin: "anti_join",
+    ops.LeftSemiJoin: "semi_join",
+    ops.Lowercase: "lower",
+    ops.RegexSearch: "re_search",
+    ops.SelfReference: "view",
+    ops.StartsWith: "startswith",
+    ops.StringContains: "contains",
+    ops.StringSQLILike: "ilike",
+    ops.StringSQLLike: "like",
+    ops.TimestampNow: "now",
+}
+
+
+def _to_snake_case(camel_case):
+    """Convert a camelCase string to snake_case."""
+    result = list(camel_case[:1].lower())
+    for char in camel_case[1:]:
+        if char.isupper():
+            result.append("_")
+        result.append(char.lower())
+    return "".join(result)
+
+
+def _get_method_name(op):
+    typ = op.__class__
+    try:
+        return _method_overrides[typ]
+    except KeyError:
+        return _to_snake_case(typ.__name__)
+
+
+def _maybe_add_parens(op, string):
+    if isinstance(op, ops.Binary):
+        return f"({string})"
+    elif isinstance(string, CallStatement):
+        return string.args
+    else:
+        return string
+
+
+# helpers classes to let parent nodes decide about omitting certain parts of the statement
+class ListStatement(list):
+    def __str__(self):
+        values = ", ".join(self)
+        return f"[{values}]"
+
+
+class CallStatement:
+    def __init__(self, func, args):
+        self.func = func
+        self.args = args
+
+    def __str__(self):
+        return f"{self.func}({self.args})"
+
+
+@functools.singledispatch
+def translate(op, *args, **kwargs):
+    """Translate an ibis operation into a Python expression."""
+    raise NotImplementedError(op)
+
+
+@translate.register(dt.DataType)
+def datatype(dtype, **kwargs):
+    return f"\"{dtype}\""
+
+
+@translate.register(sch.Schema)
+def schema(schema, **kwargs):
+    fields = dict(zip(schema.names, map(str, schema.types)))
+    return CallStatement("ibis.schema", fields)
+
+
+# TODO(kszucs): we do rewrites on construction, so we need to handle specific
+# cases like when reduction_to_aggregation is called
+
+
+@translate.register(ops.Value)
+@translate.register(ops.TableNode)
+def value(op, *args, **kwargs):
+    method = _get_method_name(op)
+    kwargs = [(k, v) for k, v in kwargs.items() if v is not None]
+
+    if args:
+        this, *args = args
+    else:
+        (_, this), *kwargs = kwargs
+
+    # if there is a single keyword argument prefer to pass that as positional
+    if not args and len(kwargs) == 1:
+        args = [kwargs[0][1]]
+        kwargs = []
+
+    args = ", ".join(map(str, args))
+    kwargs = ", ".join(f"{k}={v}" for k, v in kwargs)
+    parameters = ", ".join(filter(None, [args, kwargs]))
+
+    return f"{this}.{method}({parameters})"
+
+
+@translate.register(ops.ScalarParameter)
+def scalar_parameter(op, dtype, counter):
+    return f"ibis.param({dtype})"
+
+
+@translate.register(ops.UnboundTable)
+@translate.register(ops.DatabaseTable)
+def table(op, schema, name, **kwargs):
+    if isinstance(schema, CallStatement):
+        schema = schema.args
+    return f"ibis.table(name={name!r}, schema={schema})"
+
+
+def _try_unwrap(stmt):
+    if isinstance(stmt, list):
+        return stmt[0] if len(stmt) == 1 else stmt
+    else:
+        return stmt
+
+
+@translate.register(ops.Selection)
+def selection(op, table, selections, predicates, sort_keys):
+    out = f"{table}"
+    if selections:
+        out = f"{out}.select({_try_unwrap(selections)})"
+    if predicates:
+        out = f"{out}.filter({_try_unwrap(predicates)})"
+    if sort_keys:
+        out = f"{out}.order_by({_try_unwrap(sort_keys)})"
+    return out
+
+
+@translate.register(ops.Aggregation)
+def aggregation(op, table, by, metrics, predicates, having, sort_keys):
+    out = f"{table}"
+    if predicates:
+        out = f"{out}.filter({_try_unwrap(predicates)})"
+    if by:
+        out = f"{out}.group_by({_try_unwrap(by)})"
+    if having:
+        out = f"{out}.having({_try_unwrap(having)})"
+    if metrics:
+        out = f"{out}.aggregate({_try_unwrap(metrics)})"
+    if sort_keys:
+        out = f"{out}.order_by({_try_unwrap(sort_keys)})"
+    return out
+
+
+@translate.register(ops.Join)
+def join(op, left, right, predicates):
+    method = _get_method_name(op)
+    # TODO(kszucs): Join.predicates is a tuple of predicates, we need to clean
+    # up the implementation to use NodeList
+    predicates = ListStatement(predicates)
+    return f"{left}.{method}({right}, {_try_unwrap(predicates)})"
+
+
+@translate.register(ops.SetOp)
+def union(op, left, right, distinct):
+    method = _get_method_name(op)
+    if distinct:
+        return f"{left}.{method}({right}, distinct=True)"
+    else:
+        return f"{left}.{method}({right})"
+
+
+@translate.register(ops.Limit)
+def limit(op, table, n, offset):
+    if offset:
+        return f"{table}.limit({n}, {offset})"
+    else:
+        return f"{table}.limit({n})"
+
+
+@translate.register(ops.TableColumn)
+def table_column(op, table, name):
+    return f"{table}.{name}"
+
+
+@translate.register(ops.SortKey)
+def sort_key(op, expr, ascending):
+    if ascending:
+        return f"{expr}.asc()"
+    else:
+        return f"{expr}.desc()"
+
+
+@translate.register(ops.Reduction)
+def reduction(op, arg, where, **kwargs):
+    method = _get_method_name(op)
+    return f"{arg}.{method}()"
+
+
+@translate.register(ops.Alias)
+def alias(op, arg, name):
+    arg = _maybe_add_parens(op.arg, arg)
+    return f"{arg}.name({name!r})"
+
+
+@translate.register(ops.Constant)
+def constant(op, **kwargs):
+    method = _get_method_name(op)
+    return f"ibis.{method}()"
+
+
+@translate.register(ops.Literal)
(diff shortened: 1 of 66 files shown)


ibis/common/graph.py
```diff
@@ -0,0 +1,151 @@
...
+        queue.extend(children)
+
+    return graph
+
+
+def dfs(node: Traversable, types=Traversable):
+    assert issubclass(types, Traversable)
+    assert isinstance(node, types)
+
+    stack = deque([node])
```
cpcloud, line 45: No reason to use a `deque` here.

```suggestion
    stack = [node]
```

ibis/common/graph.py
```diff
@@ -0,0 +1,151 @@
...
+"""
+from __future__ import annotations
+
+from abc import abstractmethod
+from collections import deque
+from typing import Any, Callable, Iterable, Iterator
+
+import toolz
+
+# TODO(kszucs): add recursive implementation for bfs and dfs
```
cpcloud, line 12: Do we need to?

ibis/common/graph.py
kszucs: deque is supposed to be slightly more performant in this scenario due to frequent subsequend adding and popping

ibis/common/validators.py
```diff
@@ -8,18 +8,26 @@
...
+from ibis.util import DotDict, flatten_iterable, is_function, is_iterable
 
 EMPTY = inspect.Parameter.empty  # marker for missing argument
+VAR_POSITIONAL = inspect.Parameter.VAR_POSITIONAL
+POSITIONAL_OR_KEYWORD = inspect.Parameter.POSITIONAL_OR_KEYWORD
 
 
-class Validator(Callable):
+# TODO(kszucs): rename Validator to Annotation
+class Annotation(Callable):
```
kszucs, line 19: I'd prefer to rename this module to `annotations.py`

ibis/common/validators.py
```diff
@@ -48,12 +59,41 @@ def __call__(self, arg, **kwargs):
...
+    Marker class for validating variadic arguments.
+    """
+
+    __slots__ = ('validator',)
+
+    def __init__(self, validator):
+        self.validator = validator
+
+    def __repr__(self):
+        return f"Variadic({self.validator!r})"
```
cpcloud, line 81: Might as well use `self.__class__` here.

```suggestion
        return f"{self.__class__}({self.validator!r})"
```

ibis/common/validators.py
cpcloud: Sounds good to me!

ibis/common/tests/test_graph.py
```diff
@@ -0,0 +1,64 @@
...
+    assert g == {D: [], E: [], B: [D, E], C: [], A: [B, C]}
+
+    i = invert(g)
+    assert i == {D: [B], E: [B], B: [A], C: [A], A: []}
+
+    j = invert(i)
+    assert j == g
+
+
+def test_toposort():
```
cpcloud, line 52: What happened to the test cases for edge cases like an empty graph?

ibis/common/grounds.py
```diff
@@ -244,3 +260,14 @@ def args(self):
     @property
     def argnames(self):
         return self.__argnames__
+
+    def map(self, fn):
```
cpcloud, line 264: Looks like this method could use a test or two.

ibis/common/grounds.py
```diff
@@ -219,15 +232,18 @@ def __cached_equals__(self, other):
         return result
 
 
-class Concrete(Immutable, Comparable, Annotable):
+# TODO(kszucs): consider to move it somewhere else, like ibis/expr/core.py
+class Concrete(Immutable, Comparable, Traversable, Annotable):
```
cpcloud, line 236: It doesn't seem desirable to have every ibis object be traversible. I would expect `Traversable` to exclude data types and schemas and effectively be a filter for `ops.Node`. Why would data types and schemas ever need to be traversable in the same way that `Node`s are?

ibis/common/grounds.py
```diff
@@ -201,7 +214,7 @@ def __cached_equals__(self, other):
             return True
 
         # type comparison should be cheap
-        if type(self) != type(other):
+        if type(self) is not type(other):
```
cpcloud, line 217: Is this change meaningful in some way? When is `!=` not equivalent to `is not` in this context? Dynamically generated types?

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,145 @@
...
+def literal(op, value, dtype):
+    return repr(value)
+
+
+@translate.register(ops.NodeList)
+def node_list(op, *values):
+    return ", ".join(values)
+
+
+@translate.register(ops.Equals)
```
cpcloud, line 87: Instead of defining each of the non-syntactic APIs, e.g., `StringColumn.split()` can we have a generic `Value` implementation that looks at a new class property and if that's `None` (the default) snake cases the operation class name to be the method call?

Otherwise it's a super annoying game of whack-a-mole for every operation.

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@translate.register(dt.DataType)
+def datatype(dtype, *args, **kwargs):
+    return str(dtype)
+
+
+@translate.register(sch.Schema)
+def schema(schema, *, names, types):
+    return repr(list(zip(names, map(str, types))))
```
cpcloud, line 25: Can we use the `dict` literal representation here? The list of tuples repr is super noisy vs the dict.

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,145 @@
...
+
+@translate.register(ops.TableColumn)
+def table_column(op, table, name):
+    return f"{table}.{name}"
+
+
+@translate.register(ops.SortKey)
+def sort_key(op, expr, ascending):
+    if ascending:
+        return f"ibis.asc({expr})"
```
cpcloud, line 61:
```suggestion
        return f"{expr}.asc()"
```

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,145 @@
...
+    else:
+        return f"ibis.desc({expr})"
```
cpcloud, line 63:
```suggestion
        return f"{expr}.desc()"
```

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,145 @@
...
+def node_list(op, *values):
+    return ", ".join(values)
+
+
+@translate.register(ops.Equals)
+def equals(op, left, right):
+    return f"{left} == {right}"
+
+
+@translate.register(ops.Divide)
```
cpcloud, line 92: Eventually the arithmetic operations will have to be parenthesized. I know you haven't implemented `Add` or `Subtract` yet, but you won't be able to just render the operands recursively.

Consider this case:

```
In [11]: (ibis.literal(2) - ibis.literal(3)) + 4
Out[11]: Add(Subtract(2, 3), 4): 2 - 3 + 4

In [12]: ibis.literal(2) - (ibis.literal(3) + 4)
Out[12]: Subtract(2, Add(3, 4)): 2 - 3 + 4
```

We're technically rendering this incorrectly in the repr right now, and you'll have to deal with the parenthesization problem here too.

cpcloud (changes requested):
Looking good! A few more things to tackle before we can merge.

Can we move this out of the `backends` directory? Backends do things like create tables and execute expressions and this does neither of those things.

How about moving it to the top level, and adding either a top level `ibis.decompile()` method or `expr.decompile()` method?

ibis/common/validators.py
kszucs: Renamed in #4552

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,145 @@
...
+class Varnames:
+    def __init__(self):
+        self.t = itertools.count()
+        self.v = itertools.count()
+
+    def generate(self, node):
+        if isinstance(node, ops.TableNode):
+            return f"t{next(self.t)}"
+        elif isinstance(node, sch.Schema):
+            return f"s{next(self.s)}"
```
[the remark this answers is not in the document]
kszucs: Forgot to initialize it, fixing.

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@functools.singledispatch
+def translate(op, *args, **kwargs):
+    raise NotImplementedError(op)
+
+
+@translate.register(dt.DataType)
+def datatype(dtype, *args, **kwargs):
+    return str(dtype)
```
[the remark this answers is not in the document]
kszucs: IIRC `str()` should place `!` for non-nullable datatypes

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@translate.register(sch.Schema)
+def schema(schema, *, names, types):
+    return repr(list(zip(names, map(str, types))))
+
+
+@translate.register(ops.UnboundTable)
+def table(op, schema, name):
+    return f"ibis.table(name={name!r}, schema={schema})"
```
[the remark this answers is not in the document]
kszucs: Done.

ibis/common/grounds.py
```diff
@@ -150,6 +161,9 @@ def __eq__(self, other):
     def __getstate__(self):
         return {name: getattr(self, name) for name in self.__argnames__}
 
+    def __dismantle__(self, prefer_keywords=False):
```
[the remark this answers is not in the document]
kszucs: Removed it

ibis/common/validators.py
kszucs: Not relevant of this PR's scope anymore.

ibis/common/grounds.py
kszucs: Moved the comment to the other PR

ibis/common/grounds.py
kszucs: It's just more expressive syntactically.

ibis/backends/decompile/compiler.py
kszucs: Updated.

ibis/backends/decompile/compiler.py
kszucs: Updated.

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+@translate.register(ops.Alias)
+def alias(op, arg, name):
+    return f"{arg}.name({name!r})"
+
+
+@translate.register(ops.Literal)
+def literal(op, value, dtype):
+    return f"ibis.literal({value!r}, \"{dtype}\")"
```
mik-laj, line 85:
```suggestion
    return f"ibis.literal({value!r}, {dtype!r})"
```

ibis/backends/decompile/compiler.py
```diff
@@ -0,0 +1,145 @@
...
+        elif n_dependents == 1:
+            return result
+        else:
+            out.write(f"expr = {result}\n")
+            return result
+
+    node.map(fn)
+    # graph.map(fn, node)
+
+    rendered = f"import ibis\n{out.getvalue()}"
```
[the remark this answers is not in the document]
kszucs: Added an option.

ibis/backends/decompile/compiler.py
kszucs: Updated.

ibis/backends/decompile/compiler.py
kszucs: Updated.

ibis/backends/decompile/compiler.py
kszucs: @cpcloud I may misunderstand, do you mean we should introduce a new class property?

kszucs:
I'd recommend marking this as an experimental feature since a couple of expressions don't roundtrip properly, most notably the reductions rewritten to aggregations. 

On the other hand I managed to reach a rather high coverage using the already removed sqlgolden fixture to exercise `decompile()` on multiple expressions (see the temporarily committed, generated python files). @cpcloud could you recommend a method to properly test the decompiler without manually maintaining a bunch of tests? 

I'm not planning to invest much more time to this PR other than finalizing it. 

cc @saulpw @jcrist

ibis/backends/decompile/compiler.py
kszucs: I'm handling parenthesization from parent infix operators now, seems to work fine.

cpcloud:
@kszucs Can you use `pytest-snapshot` for the tests here? It's hard to see if you are because the large number of files changed makes the browser slow 😬

kszucs:
@cpcloud updated, could you take another look?

ibis/expr/decompile.py
```diff
@@ -0,0 +1,409 @@
...
+
+    node.map(fn)
+
+    result = out.getvalue()
+    if render_import:
+        result = f"import ibis\n\n\n{result}"
+    if format:
+        import black
+
+        result = black.format_str(result, mode=black.FileMode())
```
cpcloud, line 407: Will this pickup configuration using black's default `pyproject.toml` search procedure?

ibis/tests/sql/conftest.py
```diff
@@ -4,6 +4,60 @@
...
+# def sqlgolden(func):  # decorator
+#     @functools.wraps(func)
+#     def _decorated(*args, **kwargs):
+#         sql_golden_check(func.__name__, func(*args, **kwargs))
+#         ibis_golden_check(func.__name__, func(*args, **kwargs))
+
+#     return _decorated
+
+
+# def sqlgoldens(func):  # decorator
```
cpcloud, line 53: I think all this commented out code can be deleted.

ibis/tests/sql/snapshots/test_compiler/test_column_expr_default_name/decompiled.py
```diff
@@ -0,0 +1,4 @@
+import ibis
+
+
+result = ibis.table(name="int_col_table", schema={"int_col": "int32"}).int_col + 4
```
cpcloud, line 4: Should this be generating an intermediate table variable?

ibis/tests/sql/snapshots/test_select_sql/test_case_in_projection/decompiled.py
```diff
@@ -0,0 +1,40 @@
+import ibis
+
+
+l = ibis.literal("foo")
```
cpcloud, line 4: Not a huge fan of variable names that start with `l`, but it's generated code so 🤷🏻

cpcloud (approved):
Brilliant start!

🚢 it.

ibis/backends/decompile/compiler.py
cpcloud: This is resolved!

ibis/tests/sql/snapshots/test_compiler/test_column_expr_default_name/decompiled.py
kszucs: Currently when an expression gets used once it doesn't generate an intermediate variable, though we can add exceptions based on the types.

ibis/tests/sql/conftest.py
kszucs: Removed.

ibis/tests/sql/snapshots/test_compiler/test_column_expr_default_name/decompiled.py
cpcloud: Ah, ok. I think this is fine as is.

kszucs: Updated.

ibis/expr/decompile.py
kszucs: Great question, I have not checked it.

ibis/tests/sql/snapshots/test_select_sql/test_case_in_projection/decompiled.py
kszucs: Renamed it to `lit` for now

kszucs:
@cpcloud addressed the review comments and also did some refactoring

cpcloud:
Sweet, merging!

PR #4541, opened 2022-09-20
refactor(tests): move sql output to golden dir

saulpw:
@cpcloud lamented that it was difficult to update all the tests that check SQL output inline.  I decided to refactor it to use new `@sqlgolden` and `@sqlgoldens` decorators, which take the returned expression and compare the generated SQL to the corresponding SQL that exists on disk (and in the repo).  If the files don't exist, it writes them out instead; this makes it very easy to generate new golden outputs, as you just have to remove the contents of the `golden` directory and run the tests again.

I'm open to discussing the names of things and where the goldens go!  Also the fixtures that were converted directly to tests should probably be moved into `test_select_sql.py`.


saulpw:
Failing test is because a commit msg contains `ASC`:  "subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]"

cpcloud (approved):
This LGTM. Any improvements can be done in a follow up.

This really helps unblock #2919. Thank you for doing this!

PR #4547, opened 2022-09-21
feat(api): implement `__array__`

cpcloud:
This PR adds an implementation of `__array__` for `Column` and `Table` expressions.

There's hardly any code here, but this gives the ability to plot ibis expressions **out of the box**.

Here's an example:

image

which yields this plot:

image

### DIFF
--- a/ibis/backends/datafusion/compiler.py
+++ b/ibis/backends/datafusion/compiler.py
@@ -107,6 +107,13 @@ def selection(op):
     return plan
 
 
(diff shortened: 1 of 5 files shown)


cpcloud:
cc @jcmkk3, I think this is a good first step on the way to being able to interop with seaborn and possibly other libraries.

kszucs (approved):

gforsyth (approved):
:shipit:

PR #4551, opened 2022-09-21
fix: auto-register csv and parquet with duckdb using `ibis.connect`

### LINKED ISSUE #4548: bug: `ibis.connect` not registering csv files correctly
As noted in #4542 , the pattern `con = ibis.connect('./local.csv')` doesn't do the expected thing, which would be to throw it into `duckdb` and then return a connection, but it is nominally supported in the `connect` dispatcher.

gforsyth:
Need to add some tests and this is less elegant, but works.

Fix #4548

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -760,4 +760,6 @@ def _(filename: str, **kwargs: Any) -> BaseBackend:
     >>> con = ibis.connect("relative/path/to/data.csv")
     >>> con = ibis.connect("relative/path/to/more/data.parquet")
     """
-    return _connect(f"duckdb:///{filename}", **kwargs)
+    con = ibis.duckdb.connect()
+    con.register(filename)
+    return con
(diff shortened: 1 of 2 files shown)


ibis/backends/base/__init__.py
```diff
@@ -760,4 +760,6 @@ def _(filename: str, **kwargs: Any) -> BaseBackend:
     >>> con = ibis.connect("relative/path/to/data.csv")
     >>> con = ibis.connect("relative/path/to/more/data.parquet")
     """
-    return _connect(f"duckdb:///{filename}", **kwargs)
+    con = ibis.duckdb.connect()
+    con.register(filename)
+    return con
```
cpcloud, line 765: This seems like a fine solution :)

cpcloud (commented):
LGTM.

cpcloud (approved):

gforsyth:
Hey @cpcloud -- had to rebase away some conflicts -- can you kick over that mysql job that fell on its face?

PR #4553, opened 2022-09-21
feat(duckdb): make it more convenient to access files

cpcloud:
- feat(duckdb): overwrite views when calling `register`
- chore: return `options.default_backend` from `_default_backend`
- feat(api): return the table from `ibis.connect('foo.csv')`

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -751,15 +751,15 @@ def _(
     return con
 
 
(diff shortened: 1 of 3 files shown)


gforsyth:
My `connect` tests for parquet and CSV are checking `list_tables()` for registration success -- I like returning the `Table` better -- probably an `isinstance` check is enough for those tests now.

gforsyth (approved):
nice!

PR #4557, opened 2022-09-21
fix: ensure numpy ops dont accidentally cast ibis types

jcrist:
This fixes a small bug introduced in #4547. That PR added an `__array__` method, which lets users pass ibis columns and tables to functions expecting array-like objects, and have those implicitly work.

However, defining only `__array__` means that `numpy` will happily convert ibis types to numpy types when used in larger arithmetic expressions. We _could_ make this work by properly defining `__array_ufunc__`, but that's a bit tricky. Instead we define:

- `__array_priority__` to something higher than numpy (0) & dask priorities (11)
- `__array_ufunc__ = None` to disable the array-ufunc protocol.

Doing this means that:

- numpy scalar + ibis type works as before, no accidental computation
- numpy array + ibis type errors with a `TypeError`, rather than implicitly computing

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -513,6 +513,11 @@ class Scalar(Value):
 
 @public
 class Column(Value):
+    # Higher than numpy & dask objects
+    __array_priority__ = 20
+
+    __array_ufunc__ = None
+
     def __array__(self):
         return self.execute().__array__()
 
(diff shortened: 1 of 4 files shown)


kszucs (approved):
Nice catch!

cpcloud (approved):
Sweet, thanks!

PR #4577, opened 2022-09-27
style: format and check docstrings using `docformatter`

cpcloud:
This PR adds a precommit check for docstring formatting and runs docformatter on the codebase.


ibis/udf/vectorized.py
```diff
@@ -41,9 +39,7 @@ def _coerce_to_np_array(
 ) -> np.ndarray:
     """Coerce the following shapes to an np.ndarray.
 
-    (1) A list
-    (2) An np.ndarray
-    (3) A Series
+    (1) A list (2) An np.ndarray (3) A Series
```
cpcloud, line 42: False positive here.

PR #4579, opened 2022-09-27
feat(duckdb): enable registration of parquet files over http or s3

gforsyth:
I've tested this locally and it works.  

There are some (hopefully minor) issues.  
> One, DuckDB doesn't use the aws sdk -- so auth (if required) must be set via some SQL commands.  I've added in a very brittle shim here to support the common environment variables, but this won't work for `credentials` files or via roles on an EC2 instance, etc.  Once DuckDB is using the aws sdk to handle auth, that can all be ripped out.

Non-issue using `pyarrow.dataset`

> Two -- not sure how we'd like to test this.  I can go to town with some mocks, but I'm not entirely sure of the utility.  Open to suggestions.

Not going to test it, really, since it's mostly testing string parsing

> Three -- passing in a glob works, because we pass it through to `duckdb`, but we currently can't resolve what the glob expands to since that happens between duckdb and S3.  If you do `con.register("s3://some_bucket/*.parquet")` then `duckdb` will read all the parquet files in there and give the table the name of `*` (because we tell it to).  
The workaround is to explicitly pass in a `table_name=` to the `register` call.

Because we're using `pyarrow.dataset`, we now get the list of files and can compute a common prefix -- need to add documentation that globs are now not explicitly supported, but are also unnecessary, just pass in the bucket prefix

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -20,6 +20,7 @@ from ibis.backends.base.sql.alchemy.datatypes import to_sqla_type
 if TYPE_CHECKING:
     import duckdb
     import ibis.expr.types as ir
+    import pyarrow as pa
 
 import ibis.expr.schema as sch
 from ibis.backends.base.sql.alchemy import BaseAlchemyBackend
@@ -36,6 +37,10 @@ _gen_table_names = (f"registered_table{i:d}" for i in itertools.count())
 def _name_from_path(path: Path) -> str:
     base, *_ = path.name.partition(os.extsep)
     return base.replace("-", "_")
+
+
+def _name_from_dataset(dataset: pa.dataset.FileSystemDataset) -> str:
+    return _name_from_path(Path(os.path.commonprefix(dataset.files)))
 
 
 def _quote(name: str):
@@ -69,6 +74,17 @@ def _file(_, path, table_name=None):
     num_sep_chars = len(os.extsep)
     extension = "".join(Path(path).suffixes)[num_sep_chars:]
     return _generate_view_code(f"{extension}://{path}", table_name=table_name)
+
+
+@_generate_view_code.register(r"s3://.+", priority=10)
+def _s3(full_path, table_name=None):
+    # TODO: gate this import once the ResultHandler mixin is merged #4454
+    import pyarrow.dataset as ds
+
+    dataset = ds.dataset(full_path)
+    table_name = table_name or _name_from_dataset(dataset)
+    quoted_table_name = _quote(table_name)
+    return quoted_table_name, dataset
 
 
 @_generate_view_code.register(r".+", priority=1)
@@ -172,7 +188,20 @@ class Backend(BaseAlchemyBackend):
         ir.Table
             The just-registered table
         """
-        if isinstance(source, (str, Path)):
+        if isinstance(source, str) and source.startswith("s3://"):
+            table_name, dataset = _generate_view_code(
+                source, table_name=table_name
+            )
+            # We don't create a view since DuckDB special cases Arrow Datasets
+            # so if we also create a view we end up with both a "lazy table"
+            # and a view with the same name
+            with self._safe_raw_sql("SELECT 1") as cursor:
+                # DuckDB normally auto-detects Arrow Datasets that are defined
+                # in local variables but the `dataset` variable won't be local
+                # by the time we execute against this so we register it
+                # explicitly.
+                cursor.cursor.c.register(table_name, dataset)
+        elif isinstance(source, (str, Path)):
             sql, table_name = _generate_view_code(
                 source, table_name=table_name
             )


gforsyth:
Other options:
We use `botocore` to properly handle the various auth paths 

We use fsspec.fuse to mount the bucket locally and then point duckdb at it (this wouldn't support windows)

gerrymanoim:
Could you bypass the auth completely and provide a `from_arrow_table` mechanism that registers passed arrow tables (i.e. https://duckdb.org/docs/guides/python/sql_on_arrow). Arrow datasets can provide their own `filesystem` calls https://arrow.apache.org/docs/python/dataset.html#reading-from-cloud-storage.

gforsyth:
> Could you bypass the auth completely and provide a `from_arrow_table` mechanism that registers passed arrow tables (i.e. [duckdb.org/docs/guides/python/sql_on_arrow](https://duckdb.org/docs/guides/python/sql_on_arrow)). Arrow datasets can provide their own `filesystem` calls [arrow.apache.org/docs/python/dataset.html#reading-from-cloud-storage](https://arrow.apache.org/docs/python/dataset.html#reading-from-cloud-storage).

This is a great idea @gerrymanoim !  I have it working with `register` as above but no more wacky auth handling.

ibis/backends/duckdb/__init__.py
```diff
@@ -172,7 +192,17 @@ def register(
         ir.Table
             The just-registered table
         """
-        if isinstance(source, (str, Path)):
+        if isinstance(source, str) and source.startswith("s3://"):
```
gforsyth, line 195: this extra conditional is a bit annoying, but it's necessary since we need a cursor to register the arrow dataset with (and we also need the dataset itself, so if we're keeping the dispatcher "pure" we need to add in a fork here

ibis/backends/duckdb/__init__.py
```diff
@@ -172,7 +192,17 @@ def register(
...
+            sql, (table_name, dataset) = _generate_view_code(
+                source, table_name=table_name
+            )
+            with self._safe_raw_sql(sql) as cursor:
+                # DuckDB normally auto-detects Arrow Datasets that are defined
+                # in local variables but the `dataset` variable won't be local
+                # by the time we execute against this so we register it
+                # explicitly.
+                cursor.cursor.c.register(table_name, dataset)
```
cpcloud, line 204: Should this be `cursor.cursor.register`? I vaguely recall that there's some reason related to connection pooling to prefer using the proxy, but if this works then it's probably fine for now.

cpcloud:
Perhaps a screenshot showing successful usage 😅?

gforsyth:
```python
In [1]: import ibis

In [3]: # public dataset

In [4]: con.register("s3://voltrondata-labs-datasets/nyc-taxi/")
Out[4]: 
AlchemyTable: year=20
  vendor_name           string
  pickup_datetime       timestamp('UTC')
  dropoff_datetime      timestamp('UTC')
  passenger_count       int64
  trip_distance         float64
  pickup_longitude      float64
  pickup_latitude       float64
  rate_code             string
  store_and_fwd         string
  dropoff_longitude     float64
  dropoff_latitude      float64
  payment_type          string
  fare_amount           float64
  extra                 float64
  mta_tax               float64
  tip_amount            float64
  tolls_amount          float64
  total_amount          float64
  improvement_surcharge float64
  congestion_surcharge  float64
  pickup_location_id    int64
  dropoff_location_id   int64

In [5]: taxi = con.table("year=20")

In [6]: taxi[["extra", "mta_tax"]].head().execute()
Out[6]: 
   extra  mta_tax
0    0.5      NaN
1    0.5      NaN
2    0.0      NaN
3    0.5      NaN
4    0.0      NaN

In [7]: # private dataset (of public information)

In [8]: con.register("s3://<redacted>/10/parquet/lineitem/")
Out[8]: 
AlchemyTable: lineitem
  l_orderkey      int64
  l_partkey       int64
  l_suppkey       int64
  l_linenumber    int64
  l_quantity      float32
  l_extendedprice float32
  l_discount      float32
  l_tax           float32
  l_returnflag    string
  l_linestatus    string
  l_shipdate      date
  l_commitdate    date
  l_receiptdate   date
  l_shipinstruct  string
  l_shipmode      string
  l_comment       string

In [9]: lineitem = con.tables.lineitem

In [10]: lineitem[["l_orderkey", "l_partkey"]].head().execute()
Out[10]: 
   l_orderkey  l_partkey
0           1    1551894
1           1     673091
2           1     636998
3           1      21315
4           1     240267

In [11]: import os

In [12]: assert "AWS_SECRET_ACCESS_KEY" in os.environ

In [13]: assert "AWS_ACCESS_KEY_ID" in os.environ
```

gerrymanoim:
> > Could you bypass the auth completely and provide a `from_arrow_table` mechanism that registers passed arrow tables (i.e. [duckdb.org/docs/guides/python/sql_on_arrow](https://duckdb.org/docs/guides/python/sql_on_arrow)). Arrow datasets can provide their own `filesystem` calls [arrow.apache.org/docs/python/dataset.html#reading-from-cloud-storage](https://arrow.apache.org/docs/python/dataset.html#reading-from-cloud-storage).
> 
> This is a great idea @gerrymanoim ! I have it working with `register` as above but no more wacky auth handling.

I'd also recommend you add a way for users to pass arrow datasets directly as those can be created with arbitrary filesystems (say other clouds or other s3 API compatible storage systems)  but I leave that up to @cpcloud 's discretion.

matthewmturner:
@gforsyth this is very cool!

matthewmturner:
@gforsyth is the expectation that this implementation would work on ec2 or in a lambda function?

gforsyth:
> @gforsyth is the expectation that this implementation would work on ec2 or in a lambda function?

I haven't tested that, but at least with respect to auth that comes from bound instance roles, etc, then yes, it should work.

matthewmturner:
@gforsyth thanks!

cpcloud (approved):
LGTM!

PR #4586, opened 2022-09-29
fix: pandas/dask now cast int<->timestamp as seconds since epoch

### LINKED ISSUE #4062: bug: pandas and dask interpret cast of integer to timestamps as nanos since epoch
Every other backend interprets these as seconds since epoch. This is a breaking change.

jcrist:
The Pandas and Dask backends now interpret casting ints to/from timestamps as seconds since the unix epoch, matching other backends.

Fixes #4062.

### DIFF
--- a/ibis/backends/dask/execution/generic.py
+++ b/ibis/backends/dask/execution/generic.py
@@ -255,6 +255,29 @@ def execute_cast_series_group_by(op, data, type, **kwargs):
     return result.groupby(data.index)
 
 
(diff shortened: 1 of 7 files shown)


kszucs:
@jcrist there is a pyspark build failure which seems relevant

jcrist:
This should be ready for review now.

cpcloud (approved):
LGTM! Thanks for fixing.

Issue #4403, opened 2022-08-24
refactor(ir): remove `ir.TopK` 


kszucs:
We really need to get rid of `ir.TopK` if that still exists (I realize you're using the `ops` variant here).

_Originally posted by @cpcloud in https://github.com/ibis-project/ibis/pull/4378#discussion_r953774606_

PR #4587, opened 2022-09-29
fix: improve rich repr support

### LINKED ISSUE #4537: When verbose = True and interactive = True are used in Jupyter, SQL log messages are printed twice
When using Jupyter in a browser, it will render outputs in both text and HTML every time regardless of which is being displayed. Both of these calls are generating the SQL log output when verbose = True. These log calls should be disabled in all of the `_repr_*_` methods like `_repr_html_` that call `execute()`.

I don't know if there is a better way to handle it, but the easy fix is to change `_repr_html_` methods to something like the following:
```
    def _repr_html_(self) -> str | None:
            return None

        verbose = ibis.options.verbose
        ibis.options.verbose = False
        out = self.execute().to_frame()._repr_html_()
        ibis.options.verbose = verbose
        return out
```

I found two occurrences of these methods in `expr/types/relations.py` and `expr/types/generic.py`.

jcrist:
This PR originally was started to fix #4537, but it spiraled out into fixing a bunch of other warts with the new repr.

In summary:

- We now only compute once when repr-ing in a Jupyter notebook or terminal, as all requests go through a single `_repr_mimebundle_` rather than both `_repr_html_` and `__repr__`. Fixes #4537.
- We now use the same repr for both Jupyter notebooks and terminal sessions, providing a nice uniform experience.
- We no longer accidentally include terminal control characters in `__repr__`. This means that you won't see colored output in the classic `python` terminal with just `repr`, but you will see it in IPython or Jupyter notebooks, as those hit `_repr_mimebundle_`. IMO this is the correct behavior, `repr(obj)` should never return non-printable control characters.
- The colors and stylings are adjusted a bit to make things work well against both light and dark backgrounds
- Several formatting cleanups to make some columnar types easier to read. For example, floats now align their decimal place and don't render all 16-17 digits of precision (as done before).
- Long string data never wraps to a new line. Only nested (struct/map/array) types can result in multiple output lines per row.
- The default config options for nested types are adjusted to reduce the content length. This IMO provides a more useful repr experience, as it reduces vertical space.
- Columns with fixed-width data (numbers, short strings, etc...) never have their data truncated, and instead always show the full results. Only variable width data (long strings, nested types, ...) may result in truncated data in a row.
- `rich` has a bug in its column width calculations, resulting in columns not compressing properly for wide frames, instead overflowing and breaking the table alignment (things look really bad). Until that's fixed upstream, we now manually adjust column widths within ibis itself to provide a good view experience. Even if this is fixed upstream we might want to keep handling this internally, as it's a bit finicky.

A limited number of tests for formatting behavior have been added (mostly testing the finicky bits). I'm happy to add more tests, but would be reluctant to remove any of these.

**IPython**

image

**Standard Python**

image

**Jupyter Notebook**

image


ibis/common/rich_reprs.py
```diff
@@ -0,0 +1,333 @@
...
+
+import ibis
+import ibis.expr.datatypes as dt
+
+console = Console()
+# A console with all color/markup disabled, used for `__repr__`
+simple_console = Console(force_terminal=False)
+
+
+@singledispatch
```
kszucs, line 19: I like this approach!

Could you please move this file to something like `expr/types/pretty.py`? `ibis.common` should not be aware of the types defined under `ibis.expr`.

jcrist: Done.

cpcloud:
It looks like there's no color on this `topk` computation for some reason:

image

Strangely enough, another derived computation `.value_counts()` seems to show color just fine:
image

cpcloud:
Don't consider that a blocker, you've done enough great work here already!

jcrist:
`topk` doesn't return a `Table`, so it's falling back to `__repr__` (no `_repr_mimebundle_` on non `Table`/`Column` types).  Rather than adding a repr to that type, we should just remove it as described in #4403.

ibis/expr/types/generic.py
```diff
@@ -500,17 +500,15 @@ def to_projection(self) -> ir.Table:
 @public
 class Scalar(Value):
     def __rich_console__(self, console, options):
-        return console.render(
-            _pretty_value(_format_value(self.execute()), self.type()),
-            options=options,
-        )
+        from rich.text import Text
```
cpcloud, line 503: Are we importing here to avoid the cost?

ibis/expr/types/pretty.py
```diff
@@ -0,0 +1,333 @@
...
+    remaining = console_width - 1
+    if columns_truncated:
+        remaining -= 4
+    for name, _, min_width, max_width in col_info:
+        remaining -= min_width + 3
+        col_widths[name] = min_width
+        if min_width != max_width:
+            flex_cols.append((name, max_width))
+
+    while True:
```
cpcloud, line 273: Can you give a short comment about what this is doing at a high level?

cpcloud (approved):
Honestly, this is great.

None of my comments are blockers.

cpcloud:
@jcrist if you rebase the `simulate_release` failure should be fixed

ibis/backends/tests/test_client.py
```diff
@@ -756,18 +756,48 @@ def test_dunder_array_column(alltypes, df):
...
-    ibis.options.interactive = True
+    old = ibis.options.interactive
+    ibis.options.interactive = interactive
     try:
-        assert 'table' in t._repr_html_()
-        assert 'table' in t.int_col._repr_html_()
-        assert t.int_col.sum()._repr_html_() is None
+        s = repr(expr)
+        # no control characters
+        assert all(c.isprintable() or c in "\n\r\t" for c in s)
```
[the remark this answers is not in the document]
cpcloud: Ignore this comment.

cpcloud:
Merging. We can address any comments in a follow up.

PR #4598, opened 2022-09-30
ci: fix nix builds, relock files, remove `setup.py`

cpcloud:
This PR fixes failing `master` nix builds and also remove `setup.py` and related code. [As of `poetry-core@1.0.8`](https://github.com/python-poetry/poetry/issues/761#issuecomment-1072797923), editable installs (`pip install -e .`) are supported.


kszucs (approved):
Thanks for fixing these!

PR #4600, opened 2022-09-30
depr: deprecate `Table.sort_by` in favor of `Table.order_by`

### LINKED ISSUE #4518: api: should `GroupedTable.order_by` / `Table.sort_by` have the same method name?
That these methods have different names is a bit unexpected to me, perhaps we should settle on one and deprecate the other?

jcrist:
Fixes #4518.

### DIFF
--- a/ibis/tests/sql/test_select_sql.py
+++ b/ibis/tests/sql/test_select_sql.py
@@ -16,8 +16,8 @@ def test_select_sql(alltypes, star1, star2):
         agg_string_columns=lambda t=star1: t.aggregate(
             [t['f'].sum().name('total')], ['foo_id', 'bar_id']
         ),
-        single_column=lambda t=star1: t.sort_by("f"),
-        mixed_columns_ascending=lambda t=star1: t.sort_by(["c", ("f", 0)]),
+        single_column=lambda t=star1: t.order_by("f"),
+        mixed_columns_ascending=lambda t=star1: t.order_by(["c", ("f", 0)]),
         limit_simple=lambda t=star1: t.limit(10),
         limit_with_offset=lambda t=star1: t.limit(10, offset=5),
         filter_then_limit=lambda t=star1: t[t.f > 0].limit(10),
@@ -56,7 +56,7 @@ def test_simple_joins(star1, star2):
 
 
 @sqlgolden
-def multiple_joins(con, star1, star2, star3):
+def test_multiple_joins(con, star1, star2, star3):
     t1 = star1
     t2 = star2
     t3 = star3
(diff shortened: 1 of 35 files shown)


ibis/tests/sql/test_select_sql.py
```diff
@@ -56,7 +56,7 @@ def test_simple_joins(star1, star2):
 
 
 @sqlgolden
-def multiple_joins(con, star1, star2, star3):
```
jcrist, line 59: While updating tests, I found a few test functions that didn't start with `test_*` and thus weren't being run. Fixed here.

kszucs:
My preference would be to keep the alias.

jcrist:
That's definitely an opinion thing, I personally dislike method aliases (there's discussion of this in the original issue [here](https://github.com/ibis-project/ibis/issues/4518#issuecomment-1249700859)). When working interactively, if I see multiple methods that sound like they do the same thing (say when tab-completing), my first reaction is "why are these both here, what's the difference between them?". I was actually planning on deprecating `groupby` in favor of `group_by` next :).

I'd rather have one consistent way of spelling an API and clear docs than guessing all the different ways a user might want to spell something and supporting them all. One of ibis's draws is a cleaner API abstraction for dataframes, I'd rather not muddy that up with a bunch of method aliases.

jcrist:
One compromise idea - what about keeping a mapping of common misspellings -> the recommended method name? We could use this to improve the attribute error raised in `__getattr__` to say something like `"Table object has no attribute 'groupby', did you mean 'group_by'?"`. That way we're not cluttering up the namespace with method aliases, but still can guide the user towards the proper API.

gforsyth:
> what about keeping a mapping of common misspellings -> the recommended method name?

I like this idea -- I'm much less annoyed by variations in spelling / underscores / etc if something tells me what to try, as opposed to "just" erroring out.

Regarding whether we want to keep the alias around -- we could leave the alias in but override `__dir__` to keep the clutter out of tab completions

jcrist:
I'm happy to improve `AttributeError` messages for `Table` objects, though I'd rather do that in a follow up PR. I'd also want to handle deprecating the `groupby` alias there, and do some cleanups around `__dir__`, all of which seem ancillary to this PR.

> Regarding whether we want to keep the alias around -- we could leave the alias in but override __dir__ to keep the clutter out of tab completions

I'm not enthused about secret aliases to mask user typos. I'd rather error nicely telling the user what they should do to fix their code rather than just silently fixing it.

kszucs:
> One compromise idea - what about keeping a mapping of common misspellings -> the recommended method name? We could use this to improve the attribute error raised in `__getattr__` to say something like `"Table object has no attribute 'groupby', did you mean 'group_by'?"`. 

I like it! Then I'm fine with deprecating the aliases.

kszucs (approved):

kszucs:
cc @cpcloud for final review

cpcloud (commented):
LGTM!

cpcloud (approved):

PR #4602, opened 2022-09-30
fix: visualize(label_edges=True) works for NodeList ops

### LINKED ISSUE #4583: bug: `.visualize(label_edges=True)` case ops.NodeList we get ValueError tuple.index(x): x not in tuple
Hi,

There's still one small thing that needs fixing in the great `.visualize(label_edges=True)` feature 😄 :

When running...
```py
import ibis
t = ibis.table((("a", "int32"), ("b", "string")))
expr = t[(t["a"] == 1) & (t["b"] == "x")]

expr.visualize(label_edges=True)
```
...I get:
```
Exception has occurred: ValueError - tuple.index(x): x not in tuple
```
at the following line:
https://github.com/ibis-project/ibis/blob/2c9cfea15fc4d5f61e9099c3b270ea61498b5e45/ibis/expr/visualize.py#L117

This is happening when `v` is an `ops.NodeList` and thus its `v.args` is a tuple `of tuples` and NOT a tuple of nodes.

Given that on the next line we have special logic for `ops.NodeList`, maybe one quick fix could be to use the `.values`/`.args[0]` in such cases:

image

Thank you.

jcrist:
Fixes #4583.

### DIFF
--- a/ibis/expr/visualize.py
+++ b/ibis/expr/visualize.py
@@ -114,10 +114,11 @@ def to_graph(expr, node_attr=None, edge_attr=None, label_edges: bool = False):
                 if not label_edges:
                     label = None
                 else:
-                    index = v.args.index(u)
                     if isinstance(v, ops.NodeList):
+                        index = v.values.index(u)
                         arg_name = f"values[{index}]"
                     else:
+                        index = v.args.index(u)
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
Awesome, thank you.

PR #4611, opened 2022-10-03
fix(common): support copying variadic annotable instances

kszucs:


ibis/common/annotations.py
```diff
@@ -193,6 +193,22 @@ class Signature(inspect.Signature):
...
+        """Merge multiple signatures.
+
+        In addition to concatenating the parameters, it also reorders the
+        parameters so that optional arguments come after mandatory arguments.
+
+        Parameters
+        ----------
+        *signatures : Signature
+            Signature instances to merge.
+        **annotations : dict
```
cpcloud, line 205:
```suggestion
        signatures
            Signature instances to merge.
        annotations
```

The `*`/`**` are not valid parameter names, and the types should be annotations, which will get picked up by `mkdocstrings` when rendering docstrings.

ibis/common/grounds.py
```diff
@@ -131,24 +133,39 @@ def __repr__(self) -> str:
...
+        ----------
+        overrides : dict
+
+        Returns
+        -------
+        copy : Annotable
+        """
+        this = copy(self)
+        for name, value in overrides.items():
+            if field := self.__attributes__.get(name):
```
cpcloud, line 164: Should this check for `not None` instead of falsey?

```suggestion
            if (field := self.__attributes__.get(name)) is not None:
```

ibis/common/grounds.py
```diff
@@ -211,20 +228,30 @@ def __cached_equals__(self, other):
         return result
 
 
-# TODO(kszucs): consider to move it somewhere else, like ibis/expr/core.py
-class Concrete(Immutable, Comparable, Annotable):
+class Concrete(Immutable, Comparable, Annotable, Traversable):
```
cpcloud, line 231: Hm, why move `Traversable` here? I thought we decided that traversing schema/datatype fields was of low utility.

ibis/expr/schema.py
```diff
@@ -90,12 +90,10 @@ def __getitem__(self, name: str) -> dt.DataType:
 
     def equals(self, other: Schema) -> bool:
         """Return whether `other` is equal to `self`.
-
         Parameters
         ----------
         other
             Schema to compare `self` to.
-
```
cpcloud, line 98: I believe these newlines are necessary. In any case, I find it hard to read this all scrunched together.

ibis/common/grounds.py
kszucs: I think it could be useful to e.g. replace certain datatypes in a schema:

```py
schema.replace({dt.bool: dt.tinyint})
```

cpcloud: How much of this change requires all code that uses it to now filter on `ops.Node` everywhere?

cpcloud (approved):
Sweet!

ibis/common/grounds.py
kszucs: Going to add that feature.

PR #4615, opened 2022-10-05
test(snowflake): fix `test_export` and skip tests if envar isn't defined

cpcloud:
Fixes failing snowflake tests.

### DIFF
--- a/ibis/backends/tests/test_export.py
+++ b/ibis/backends/tests/test_export.py
@@ -43,7 +43,6 @@ limit = [
                     "impala",
                     "pandas",
                     "pyspark",
-                    "snowflake",
                 ]
             ),
         ],
@@ -61,7 +60,6 @@ no_limit = [
                     "dask",
                     "impala",
                     "pyspark",
-                    "snowflake",
                 ]
             ),
         ],
--- a/ibis/backends/snowflake/tests/conftest.py
+++ b/ibis/backends/snowflake/tests/conftest.py
@@ -6,7 +6,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 import pytest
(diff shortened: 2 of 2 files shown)


ibis/backends/tests/test_export.py
```diff
@@ -43,7 +43,6 @@ def no_pyarrow(backend):
                     "impala",
                     "pandas",
                     "pyspark",
-                    "snowflake",
```
gforsyth, line 46: obviously having CI pass is the more important issue here, but if we skip the snowflake tests, does it cause failures if it's marked in `notimpl`?  

My minor nit with removing it from here is that it looks like this _is_ implemented for `snowflake`

cpcloud: > obviously having CI pass is the more important issue here, but if we skip the snowflake tests, does it cause failures if it's marked in notimpl?

In this case the tests are passing. The test will only run if `SNOWFLAKE_URL` is defined, otherwise it's skipped.

The existing semantics around `notimpl` are no different when `SNOWFLAKE_URL` is defined.

cpcloud: The way I think about it after this PR is "whether something is notimpl/notyet/broken is unknowable if the test is skipped".

An alternative would be to require use of `not snowflake` if you don't have the environment variable defined.

Another alternative might be to skip discard snowflake backend tests from discovery if the environment variable isn't defined.

cpcloud: @gforsyth Anything else you want me to do hereE?

gforsyth: no, I think we're good -- thanks for the clarification

Issue #4582, opened 2022-09-28
feat: implement table valued functions


kesmit13:
I've been looking in the documentation and source for a way to do user-defined table valued functions, but I haven't seen any evidence that they are possible. I see in `ibis/expr/rules.py` that there is a `TABULAR` enum element defined and commented out. I was trying to come up with a work-around by injecting various things into the `output_shape`, `output_type`, and `output_dtype`, but I haven't come across any way to cheat one in. I was just wondering if this was something on the roadmap?

cpcloud:
Thanks for opening this!

This is something that other folks have expressed interest in but we haven't put any concerted effort into making it happen.

Can you give a few examples of functions (other than `UNNEST` or sampling!) that you're interested in expressing? It sounds like you also want a way to bake in an existing one, is that right? Are you also interesting in defining the data in the table in the UDF as well, e.g., generating a random set of rows (with a fixed schema) using a generator, or just mapping in already-defined UDTFs into ibis?

My first thought for the case where the function already exists in the database is that once we refactor the SQL compilers to use dispatching, where all operations--not just `Value` subclasses--are implemented using dispatching (Table operations are kind of hacked in right now and compiled differently than others) then this will be supported out of the box and be done like this:

1. define a new subclass of `TableNode` (e.g., `MyTableFunction`)
2. define an API to construct the node and expr,
   e.g., `def my_table_function(...) -> TableExpr`
3. define a compilation rule for a specific backend
   e.g.,

   ```python
   @ibis.bigquery.add_operation(MyTableFunction)
   def compile_my_table_function(t, op):
       return f"MY_TABLE_FUNCTION(...)"  # sql specific to the backend and function
   ```

kesmit13:
The major use-case we have is supporting customers UDTFs. At the moment, we are trying attempting to implement correlation matrices for heat maps and descriptive statistics (like `pandas.describe`). Generating data would also be very useful, but we would probably just implement the data generation in a UDTF and call it.

NickCrews:
Echoing that I just ran into a need to do this. The use case is similar to a coordinate reprojection, eg `(x, y) -> (x', y')` (except of course I can't figure out how to do it in ibis, it only seems feasible to do numerically with scipy)

cpcloud:
@NickCrews Can you give a bit more detail about what you're doing and if you had to express it Python or SQL how you'd write (or ideally write it?)

NickCrews:
[...]

PR #4622, opened 2022-10-07
refactor: rewrite clickhouse compiler

cpcloud:
This PR is a ground-up refactor of the ClickHouse compiler.

The purpose of this refactor is to simplify the architecture of the SQL
compilers such that a number of future use cases are unlocked and to make the maintenance of
the SQL compilers actually possible.

Right now, the architecture is far too complex. There's recursion across a
large number of files, classes, and base classes and in general a large number
of unnecessary classes.

Aliasing, for example, is something that doesn't need its own class or much
state beyond a top level dictionary.

In this PR I generate the aliases for each `TableNode` one by one (except the
root -- the final expression), before doing a bottom-up translation of each
one.

The vision after this PR is to port the SQL compilers one by one (though,
probably not the impala backend) to use the structure I'm laying out here,
including the SQLAlchemy backends. The way this will be accomplished is through
use of the [`sqlglot`](https://github.com/tobymao/sqlglot) library.

Eventually _all_ of the SQL compilers will share the patterns implemented here.

This PR also paves the way for users being able to map their existing UDTFs
(user-defined table functions) into ibis (xref
https://github.com/ibis-project/ibis/issues/4582).

TODO:

- [x] Investigate/fix joins
- [x] sqlglot window function bug (https://github.com/tobymao/sqlglot/issues/558)

Depends on #4963.


ibis/expr/types/generic.py
```diff
@@ -506,6 +507,30 @@ def __rich_console__(self, console, options):
             return console.render(Text(self._repr()), options=options)
         return console.render(repr(self.execute()), options=options)
 
+    def as_table(self) -> ir.Table:
```
kszucs, line 510: +1

kszucs: We should use it in both the datafusion and the polars backends.

ibis/expr/operations/relations.py
```diff
@@ -373,6 +373,15 @@ def _projection(self):
         return Projection(self.table, self.selections)
 
 
+@public
+class DummyTable(TableNode):
+    ops = rlz.nodes_of(rlz.scalar(rlz.any), min_length=1)
```
kszucs, line 378: How about the following:

```py
values = rlz.nodes_of(rlz.any, min_length=1)
```

I assume we could use this for unbound column expressions as well.

ibis/expr/operations/generic.py
```diff
@@ -184,32 +184,31 @@ class Least(CoalesceLike):
...
-            np.generic,
-            np.ndarray,
-            pd.Timedelta,
-            pd.Timestamp,
-            str,
-            tuple,
-            type(None),
-            uuid.UUID,
-        )
+    __valid_input_types__ = (
```
kszucs, line 187: Just for better readability?

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -0,0 +1,102 @@
...
+    # TODO(cpcloud): can values be translated bottom up as well?
+    nodes = list(dag.toposort())
+
+    # generate aliases for tables that don't have names already
+    # don't include the root ([:-1]) because it's not necessary to alias
+    aliases = {node: f"t{i:d}" for i, node in enumerate(nodes[:-1]) if dag[node]}
+
+    cache = {}
+
+    for node in nodes:
```
kszucs, line 83: I wonder, could we use `node.map()` here? Seemingly they serve similar purposes.

kszucs:
I took a preliminary look and this implementation is  **waaaay cleaner** than the previous!

ibis/backends/clickhouse/__init__.py
```diff
@@ -136,10 +254,94 @@ def list_tables(self, like=None, database=None):
...
+                try:
+                    sql = sql.limit(limit)
+                except AttributeError:
+                    sql = sg.select("*").from_(sql).limit(limit)
+            else:
+                curlimit = int(limexpr.args["expression"].args["this"])
+                newlimit = min(limit, curlimit)
+                sql.args["limit"] = limexpr.replace(
+                    sg.exp.Limit(
+                        expression=sg.exp.Literal(this=newlimit, is_string=False)
```
tobymao, line 274: sg.exp.Limit.number(newlimit)

ibis/expr/operations/relations.py
cpcloud: Nah, this is only for modeling `SELECT` statements that don't require a `FROM`, like `SELECT 1 + 1` or `SELECT 'a'`, etc.

ibis/backends/clickhouse/__init__.py
```diff
@@ -29,12 +37,77 @@
...
+        )
+
+
+def _get_result_handler(expr):
+    if isinstance(expr, ir.Scalar):
+        return lambda result: result.iloc[0, 0]
+    elif isinstance(expr, ir.Column):
+        return lambda result: result.iloc[:, 0]
+    else:
+        return toolz.identity
```
kszucs, line 106: I can imagine having an `op.output_shape.is_tabular()` here, but in order to do that we would need to turn TableNodes into Value subclasses which could make sense eventually.

ibis/backends/clickhouse/compiler/core.py
kszucs: This really reminds me of `Node.map()`'s implementation, could you please try to use that here?

ibis/backends/clickhouse/compiler/values.py
```diff
@@ -0,0 +1,1399 @@
...
+    raise NotImplementedError(type(op))
+
+
+for _t in ops.Literal.__valid_input_types__:
+    translate_val.register(_t)(translate_val.register(_t)(lambda val, **_: val))
+
+del _t
+
+
+@translate_val.register(dt.DataType)
```
kszucs, line 36: `DataType` is not a `value` :)

kszucs:
At first glance it's much nicer and cleanet, I really like it! Going to have a more thorough look.

ibis/backends/clickhouse/compiler/__init__.py
```diff
@@ -0,0 +1,11 @@
+from public import public
+
+from ibis.backends.clickhouse.compiler.core import translate
+from ibis.backends.clickhouse.compiler.relations import translate_rel
+from ibis.backends.clickhouse.compiler.values import translate_val
+
+public(
+    translate=translate,
+    translate_rel=translate_rel,
```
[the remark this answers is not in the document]
cpcloud: It's also easier to understand, since there isn't a giant `translate` dispatched many many times.

ibis/backends/clickhouse/__init__.py
```diff
@@ -54,6 +127,52 @@ def __init__(self, *args, external_tables=None, **kwargs):
...
+            if (value := getattr(url, name, None))
+        }
+        if username := url.username:
+            kwargs["user"] = username
+
+        kwargs.update(url.query)
+        self._convert_kwargs(kwargs)
+        return self.connect(**kwargs)
+
+    def _convert_kwargs(self, kwargs):
```
kszucs, line 170: Is it used anywhere else than `_from_url`?

ibis/common/grounds.py
```diff
@@ -270,15 +270,15 @@ def map(self, fn, filter=None):
         results = {}
         for node in Graph.from_bfs(self, filter=filter).toposort():
             kwargs = recursive_get(node.__getstate__(), results)
-            results[node] = fn(node, **kwargs)
+            results[node] = fn(node, **kwargs, cache=results)
```
kszucs, line 273: Rather pass it as positional so we don't interfere with kwargs?

ibis/util.py
```diff
@@ -509,3 +509,10 @@ def backend_entry_points() -> list[_importlib_metadata.EntryPoint]:
     else:
         eps = _importlib_metadata.entry_points(group="ibis.backends")
     return sorted(eps)
+
+
+def find_memtables(op):
```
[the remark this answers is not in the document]
kszucs: Could we move this to analysis.py?

ibis/backends/clickhouse/tests/test_literals.py
```diff
@@ -54,18 +54,30 @@ def test_subsecond_timestamp_literals(con, translate, expr, expected):
...
-        ('simple', "'simple'"),
-        ('I can\'t', "'I can\\'t'"),
-        ('An "escape"', "'An \"escape\"'"),
+        param("simple", "'simple'", "simple", id="simple"),
+        param(
+            'I can\'t',
+            "'I can\\\\'t'",
+            "",
+            id="nested_quote",
+            marks=pytest.mark.skip(reason="figure out quoting"),
```
kszucs, line 65: Is it need to be fixed?

ibis/expr/types/relations.py
```diff
@@ -92,6 +92,9 @@ class Table(Expr, JupyterMixin):
     def __array__(self, dtype=None):
         return self.execute().__array__(dtype)
 
+    def as_table(self):
```
kszucs, line 95: Shall we add `as_table` to the base Expr as an abstractmethod?

cpcloud: Hm, I thought I did. I'll do that if not.

ibis/backends/clickhouse/tests/snapshots/test_select/test_simple_joins/join_keys0-inner_join/out.sql
```diff
@@ -0,0 +1,5 @@
+SELECT
+  batting.*
```
kszucs, line 2: `t1`?

ibis/backends/clickhouse/__init__.py
```diff
@@ -140,10 +259,86 @@ def list_tables(self, like=None, database=None):
...
+        limit: str | None = 'default',
+        external_tables: Mapping[str, pd.DataFrame] | None = None,
+        **kwargs: Any,
+    ) -> Any:
+        """Execute an expression."""
+        table_expr = expr.as_table()
+        sql = self.compile(table_expr, limit=limit, **kwargs)
+        self._log(sql)
+
+        for memtable in graph.traverse(util.find_memtables, expr.op()):
```
kszucs: `finf_memtables` looks like an analysis function, could we move it there?

ibis/backends/clickhouse/__init__.py
```diff
@@ -140,10 +259,86 @@ def list_tables(self, like=None, database=None):
         databases = list(data[0])
         return self._filter_with_like(databases, like)
 
+    def _apply_limit(self, sql, limit):
```
kszucs, line 263: Could we rewrite this to wrap the input ibis expression with an `ops.Limit` instead of working with sqlglot expressions after compilation?

cpcloud: Yup, good catch, thanks.

ibis/backends/clickhouse/__init__.py
cpcloud: Moved!

ibis/expr/types/relations.py
cpcloud: Done!

ibis/expr/operations/relations.py
```diff
@@ -384,6 +384,15 @@ def _projection(self):
         return Projection(self.table, self.selections)
 
 
+@public
+class DummyTable(TableNode):
+    ops = rlz.tuple_of(rlz.scalar(rlz.any), min_length=1)
```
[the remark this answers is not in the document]
cpcloud: Done!

ibis/backends/clickhouse/tests/test_literals.py
cpcloud: Should be fixed now.

ibis/expr/operations/relations.py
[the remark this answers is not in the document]
cpcloud: Done!

ibis/backends/clickhouse/tests/snapshots/test_select/test_simple_joins/join_keys0-inner_join/out.sql
cpcloud: Fixed.

ibis/backends/clickhouse/tests/snapshots/test_select/test_join_self_reference/out.sql
```diff
@@ -0,0 +1,5 @@
+SELECT
+  functional_alltypes.*
```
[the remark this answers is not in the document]
cpcloud: Fixed.

ibis/expr/analysis.py
```diff
@@ -867,3 +875,11 @@ def _rewrite_filter_value_list(op, **kwargs):
         return op
 
     return op.__class__(*visited)
+
+
+def find_memtables(op):
```
kszucs, line 880: The analysis functions encapsulate the traversal method itself. to remain consistent could we have a `find_memtables()` with returns with an iterator of in memory tables?

ibis/backends/clickhouse/tests/snapshots/test_select/test_simple_joins/join_keys0-inner_join/out.sql
kszucs: Did you rename the test case? Asking because the snapshot hasn't changed.

kszucs: I assume yes. How can we remove the outdated snapshots?

ibis/backends/clickhouse/__init__.py
cpcloud: No

ibis/backends/clickhouse/compiler/__init__.py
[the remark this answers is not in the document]
kszucs: I prefer a single function, but doesn't have a strong opinion here.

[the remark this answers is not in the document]
kszucs: I see now, translate_rel is called by `node.map()` with pre-translated keywords.

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -0,0 +1,91 @@
...
+    -------
+    sqlglot.expressions.Expression
+        A sqlglot expression
+    """
+    params = {param.op(): value for param, value in params.items()}
+
+    alias_index = 0
+    aliases = {}
+
+    def fn(node, *_, params=params, **kwargs):
```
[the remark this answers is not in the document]
cpcloud: I suppose I could! Thanks!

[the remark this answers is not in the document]
kszucs: Due to kwargs?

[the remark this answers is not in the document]
cpcloud: Done.

ibis/expr/analysis.py
cpcloud: Done.

ibis/backends/clickhouse/__init__.py
cpcloud: It's also not a new API in this PR, so I'm inclined to leave it as is.

ibis/backends/clickhouse/compiler/values.py
```diff
@@ -0,0 +1,1368 @@
...
+
+    if (default := op.default) is not None:
+        buf.append(f"ELSE {translate_val(default, **kw)}")
+
+    buf.append("END")
+    return " ".join(map(_sql, buf))
+
+
+@translate_val.register(ops.TableArrayView)
+def _table_array_view(op, *, cache, **kw):
```
kszucs, line 517: Why does it require the results object (cache) here?

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -0,0 +1,91 @@
...
+
+        if alias := aliases.get(node):
+            try:
+                return raw_rel.subquery(alias)
+            except AttributeError:
+                return sg.alias(raw_rel, alias)
+        else:
+            return raw_rel
+
+    results = op.map(fn, filter=ops.TableNode)
```
kszucs, line 89: So it traverses only the table nodes here, hmmm.

kszucs: Why not traverse all nodes and also execute translate on value nodes in a bottom-up fashion?

cpcloud: Because we allow passing in relations as value expressions e.g., `t.join(s).select(t)`. We can tackle this refactor if we "desugar" those into projections on expression construction.

ibis/backends/clickhouse/tests/snapshots/test_select/test_simple_joins/join_keys1-inner_join/out.sql
```diff
@@ -0,0 +1,5 @@
+SELECT
+  batting.*
```
kszucs, line 2: Shouldn't this be `t1`?

ibis/backends/clickhouse/__init__.py
```diff
@@ -31,12 +38,77 @@
     _default_compression = False
 
 
-class Backend(BaseSQLBackend):
+class ClickhouseTable(ir.Table):
+    """References a physical table in Clickhouse."""
+
+    @property
+    def _qualified_name(self):
```
kszucs, line 45: Seems to be identical with `.name`

ibis/backends/clickhouse/tests/snapshots/test_select/test_simple_joins/join_keys1-inner_join/out.sql
cpcloud: It might be an old snapshot, lemme delete and try again.

ibis/backends/clickhouse/__init__.py
```diff
@@ -31,12 +38,77 @@
...
+        assert set(schema.names) >= set(obj.columns)
+
+        columns = ", ".join(map(quote_identifier, obj.columns))
+        query = f"INSERT INTO {self._qualified_name} ({columns}) VALUES"
+
+        # convert data columns with datetime64 pandas dtype to native date
+        # because clickhouse-driver 0.0.10 does arithmetic operations on it
+        obj = obj.copy()
+        for col in obj.select_dtypes(include=["datetime64"]):
+            if isinstance(schema[col], dt.Date):
```
kszucs, line 88:
```suggestion
            if schema[col].is_date():
```

ibis/backends/clickhouse/__init__.py
```diff
@@ -31,12 +38,77 @@
...
+                obj[col] = obj[col].dt.date
```
kszucs, line 89: According to codecov we don't have tests for this.

cpcloud: I think it can be deleted

ibis/backends/clickhouse/__init__.py
cpcloud: Removed

ibis/backends/clickhouse/__init__.py
cpcloud: Deleted this code.

ibis/backends/clickhouse/__init__.py
cpcloud: Deleted.

ibis/backends/clickhouse/__init__.py
```diff
@@ -31,12 +38,77 @@
...
+        """
+        return self._client.describe_formatted(self._qualified_name)
+
+    describe_formatted = metadata
+
+    @property
+    def name(self):
+        return self.op().name
+
+    def insert(self, obj, **kwargs):
```
kszucs, line 66: Shall we move this implementation to `_client` (the clickhouse backend) similarly to `desribte_formatted()` and `invalidate_metadata()`

cpcloud: We can delete the `invalidate_metadata` impl here because it's not actually implemented.

cpcloud: similarly with `describe_formatted`

cpcloud: Removed the dead methods.

Let's keep `insert` here for now.

ibis/backends/clickhouse/__init__.py
```diff
@@ -184,26 +302,136 @@ def _client_execute(self, query, external_tables=None):
...
+        from ibis.backends.pyarrow.datatypes import ibis_to_pyarrow_struct
+
+        schema = self._table_or_column_schema(expr)
+
+        def _batches():
+            for batch in self._cursor_batches(
+                expr, params=params, limit=limit, chunk_size=chunk_size
+            ):
+                struct_array = pa.array(
+                    map(tuple, batch),
```
kszucs, line 321: Is this conversion required? What's the previous type?

cpcloud: This is copypasted from the `BaseSQLBackend` implementation. We'll have to do another cleaning once our compilers are all based on `sqlglot`.

ibis/expr/analysis.py
kszucs: I mean something like 

```py
def find_memtables(node)
    def finder(node):
        return g.proceed, op if isinstance(op, ops.InMemoryTable) else None

    return tuple(g.traverse(finder, filter=ops.Node))
```

ibis/backends/base/sql/__init__.py
```diff
@@ -270,7 +267,7 @@ def _register_in_memory_table(self, table_op):
 
     def _register_in_memory_tables(self, expr):
         if self.compiler.cheap_in_memory_tables:
-            for memtable in graph.traverse(_find_memtables, expr.op()):
+            for memtable in graph.traverse(an.find_memtables, expr.op()):
```
kszucs, line 270: ideally this should be:
```suggestion
            for memtable in an.find_memtables(expr.op()):
```

ibis/expr/analysis.py
cpcloud: hm, I swear I implemented this.

ibis/backends/clickhouse/__init__.py
```diff
@@ -184,26 +275,136 @@ def _client_execute(self, query, external_tables=None):
...
+        limit: str | None = 'default',
+        external_tables: Mapping[str, pd.DataFrame] | None = None,
+        **kwargs: Any,
+    ) -> Any:
+        """Execute an expression."""
+        table_expr = expr.as_table()
+        sql = self.compile(table_expr, limit=limit, **kwargs)
+        self._log(sql)
+
+        for memtable in graph.traverse(an.find_memtables, expr.op()):
```
kszucs, line 355:
```suggestion
        for memtable in an.find_memtables(expr.op()):
```

ibis/backends/clickhouse/__init__.py
```diff
@@ -184,26 +275,136 @@ def _client_execute(self, query, external_tables=None):
...
+            self._register_in_memory_table(memtable)
+
+        result = self.fetch_from_cursor(
+            self.raw_sql(sql, external_tables=external_tables),
+            table_expr.schema(),
+        )
+        result_handler = _get_result_handler(expr)
```
kszucs, line 362: `_get_result_handler` is not used anywhere else, so 
```suggestion
        if isinstance(expr, ir.Scalar):
            return result.iloc[0, 0]
        elif isinstance(expr, ir.Column):
            return result.iloc[:, 0]
        else:
            return result
```

ibis/expr/analysis.py
cpcloud: Done ... again 😂

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -0,0 +1,199 @@
...
+def _selection(op: ops.Selection, *, table, needs_alias=False, **kw):
+    # needs_alias should never be true here in explicitly, but it may get
+    # passed via a (recursive) call to translate_val
+    assert not needs_alias, "needs_alias is True"
+    if needs_alias := isinstance(op.table, ops.Join) and not isinstance(
+        op.table, (ops.LeftSemiJoin, ops.LeftAntiJoin)
+    ):
+        args = table.this.args
+        from_ = args["from"]
+        (join,) = args["joins"]
```
kszucs, line 43: Can there be more joins?

cpcloud: Yes, but not from the code we generate.

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -0,0 +1,199 @@
...
+    return sel
+
+
+@translate_rel.register(ops.Aggregation)
+def _aggregation(op: ops.Aggregation, *, table, **kw):
+    tr_val = partial(translate_val, **kw)
+
+    by = tuple(map(tr_val, op.by))
+    metrics = tuple(map(tr_val, op.metrics))
+    selections = (by + metrics) or "*"
```
kszucs, line 73: This is really elegant!

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -0,0 +1,199 @@
...
+        from_ if from_ is not None else table, dialect="clickhouse"
+    )
+
+    if join is not None:
+        sel = sel.join(join)
+
+    if predicates := op.predicates:
+        if join is not None:
+            sel = sg.select("*").from_(sel.subquery(kw["aliases"][op.table]))
+        res = functools.reduce(lambda x, y: f"{x} AND {y}", map(tr_val, predicates))
```
kszucs, line 58: In the `_join()` translator you are using `functools.reduce(lambda left, right: left.and_(right), preds)`, is there a difference?

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -0,0 +1,199 @@
...
+    if predicates:
+        preds = [sg.condition(translate_val(pred, **kw)) for pred in predicates]
+        on = functools.reduce(lambda left, right: left.and_(right), preds)
+    else:
+        on = None
+    join_type = _JOIN_TYPES[type(op)]
+    try:
+        return left.join(right, join_type=join_type, on=on, dialect="clickhouse")
+    except AttributeError:
+        select_args = (f"{left.alias_or_name}.*",) + (f"{right.alias_or_name}.*",) * (
```
kszucs, line 116: Can you add a comment explaining this branch?

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -0,0 +1,199 @@
...
+            not isinstance(op, (ops.LeftSemiJoin, ops.LeftAntiJoin))
+        )
+        return (
+            sg.select(*select_args, dialect="clickhouse")
+            .from_(left, dialect="clickhouse")
+            .join(right, join_type=join_type, on=on, dialect="clickhouse")
```
kszucs, line 122: need to pass clickhouse dialect everywhere?

ibis/backends/clickhouse/compiler/values.py
```diff
@@ -0,0 +1,1368 @@
...
+
+
+@functools.singledispatch
+def translate_val(op, **_):
+    """Translate a value expression into sqlglot."""
+    raise NotImplementedError(type(op))
+
+
+for _t in ops.Literal.__valid_input_types__:
+    translate_val.register(_t)(lambda val, **_: val)
```
kszucs, line 30: Why are these overloads needed?

ibis/backends/clickhouse/__init__.py
kszucs: Please add a TODO note or a follow-up issue so we don't forget it.

ibis/backends/clickhouse/compiler/relations.py
cpcloud: Yeah, unfortunately I haven't figured out a way to avoid this yet. After chatting with Toby, I tried doing this by only returning sqlglot nodes everywhere, but that didn't work. We should do this at some point so I'll add a `TODO`.

ibis/backends/clickhouse/compiler/values.py
cpcloud: They aren't! Good catch, I had some overly aggressive `translate_val` calls left over.

ibis/backends/clickhouse/compiler/relations.py
cpcloud: Not really. Made them consistent.

ibis/backends/clickhouse/compiler/relations.py
cpcloud: Done!

ibis/common/grounds.py
cpcloud: Sounds good.

cpcloud: Done!

ibis/backends/clickhouse/compiler/relations.py
cpcloud: Done.

kszucs:
Could you please squash the PR before merging?

cpcloud:
@kszucs Squashing now!

cpcloud:
@kszucs I'm doing one last test run to prune snapshots, please don't merge yet!

cpcloud:
Bombs away!

PR #4623, opened 2022-10-07
feat: add way to specify sqlglot dialect on backend

jcrist:
We use `sqlglot` for pretty-printing SQL. Not all backends are directly supported by sqlglot, but most have a dialect that sqlglot supports that's close enough.

Instead of hardcoding a mapping in the `pretty.py` module, we now add a private optional attribute on the backend, and check there first.

### DIFF
--- a/ibis/backends/clickhouse/__init__.py
+++ b/ibis/backends/clickhouse/__init__.py
@@ -29,6 +29,8 @@ except ImportError:
 
 class Backend(BaseSQLBackend):
     name = 'clickhouse'
+    # for now map clickhouse to mysql so that _something_ works
+    _sqlglot_dialect = "mysql"
     table_expr_class = ClickhouseTable
     compiler = ClickhouseCompiler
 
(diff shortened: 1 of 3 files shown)


jcrist:
Ok, this should be ready for review. I was over thinking it before, both `ibis.show_sql(some_impala_expr)` and `ibis.show_sql(some_unbound_expr, dialect="impala")` now work and are prettified.

ibis/backends/clickhouse/__init__.py
```diff
@@ -29,6 +29,8 @@
 
 class Backend(BaseSQLBackend):
     name = 'clickhouse'
+    # for now map clickhouse to mysql so that _something_ works
+    _sqlglot_dialect = "mysql"
```
cpcloud, line 33: clickhouse exists as a dialect in sqlglot now, so `"clickhouse"` should work

cpcloud: Good question, I don't think so, but feel free to do that here. Bumping the minimum version to the latest version `9.0.5` is fine.

cpcloud (approved):

PR #4625, opened 2022-10-08
build: upgrade builds to poetry 1.2

cpcloud:
This PR upgrades our build systems to use poetry 1.2

Improvements:

- Python startup time in the presence of poetry_dynamic_versioning is now sane
- Support for dependency groups, (though unused at the moment) for speeding up
  nix environment creation (e.g., envs for only tests, only dev, only docs)
- Dependency groups make CI faster across the board by allowing selective installation of test dependencies, which allows us to avoid installing documentation dependencies

Less than Ideal Things:

- Pinning pandas <1.5. Note this is unrelated to the poetry bump, but related to an
  implicit version bump caused by the `pyproject.toml` constraint.
  Pandas 1.5 adds two new warnings that require a bunch of changes in ibis to address.
  Even after addressing them--which I did in this PR--there's a bunch of dask
  related breakage that I don't have the patience to debug. The pandas
  backend will work with 1.5, but the dask backend is broken. Maybe we should
  keep the pandas<2 pin and declare the dask tests `@pytest.mark.broken`?


dev/update-lock-files.sh
```diff
@@ -9,5 +9,5 @@ TOP="${1:-$(dirname "$(dirname "$(readlink -f "$0")")")}"
 
 pushd "${TOP}" > /dev/null || exit 1
 poetry lock --no-update
-poetry export --dev --without-hashes --no-ansi --extras all > "${TOP}/requirements.txt"
+poetry export --with dev --without-hashes --no-ansi > "${TOP}/requirements.txt"
```
cpcloud, line 12: This removes extras from requirements.txt. I think this is desirable, since there's a bunch of implicit system dependencies that show up if you include all the extras, whereas all the non-optional dependencies have wheels for every platform or are platform independent.

poetry-overrides.nix
```diff
@@ -6,25 +6,23 @@ let
...
-  });
   versionBetween = version: lower: upper:
     lib.versionAtLeast version lower && lib.versionOlder version upper;
   customizeCython = (lib.versionAtLeast pandasVersion "1.4.4") && (lib.versionOlder cythonVersion "0.29.32");
+  addBuildSystems = drv: buildSystems: drv.overridePythonAttrs (attrs: {
+    nativeBuildInputs = attrs.nativeBuildInputs or [ ] ++ buildSystems;
+  });
 in
 {
   cython = super.cython.overridePythonAttrs (
```
cpcloud, line 17: This override probably isn't necessary anymore.

ibis/backends/polars/compiler.py
```diff
@@ -244,7 +244,9 @@ def fillna(op):
             value = op.replacements.value
 
         if value is not None:
-            column = column.fill_nan(value).fill_null(value)
+            if isinstance(dtype, dt.Floating):
```
saulpw, line 247: Did you mean to include this in this PR, or should this be a separate change?

cpcloud: I meant to include it, because it's a change necessary for the latest polars version.

gforsyth (approved):
Woo!  This looks good to me.  The poetry groups are nice!

PR #4626, opened 2022-10-08
ci: remove impala and pyspark special case testing

cpcloud:
The latest impala driver appears to be much more robust in the face of multiple processes running. This PR cleans up some CI configuration to avoid special casing the test runs in CI.

After this PR it should be possible to run `pytest -n auto --dist=loadgroup` verbatim and get parallel testing for things that support it and serial testing for things that don't.


cpcloud:
Doesn't seem to be true locally. Leaving this as is.

PR #4633, opened 2022-10-11
feat(insert): add support for list+dict to sqlalchemy backends

saulpw:


cpcloud (approved):
LGTM! The patch coverage seems ... wrong. I checked this out and that line is definitely being hit.

PR #4637, opened 2022-10-12
fix: fix expression repr for table -> value operations

### LINKED ISSUE #4636: Problem with formatting union expressions when using `value_counts`
I'm working on a subclass of the MySQL backend and using unions. When attempting to do a `value_counts` on a union, I get an attribute error. Here is a simple test using our backend (this DataFrame upload might not work in the actual MySQL, but should be fairly simple to replicate).
```
df = pd.DataFrame([[0, 1], [2, 3]], columns=['a', 'b'])
tbl = conn.create_table('test_union', df, force=True)
u = ibis.union(tbl, tbl)
u.a.value_counts()
```

Here is the tail end of the exception.
```
...
~/.pyenv/versions/3.9.4/lib/python3.9/site-packages/ibis/expr/format.py in _fmt_value_expr(expr, aliases)
    555     Forwards the call on to the specific operation dispatch rule.
    556     """
--> 557     return fmt_value(expr.op(), aliases=aliases)
    558 
    559 

~/.pyenv/versions/3.9.4/lib/python3.9/functools.py in wrapper(*args, **kw)
    875                             '1 positional argument')
    876 
--> 877         return dispatch(args[0].__class__)(*args, **kw)
    878 
    879     funcname = getattr(func, '__name__', 'singledispatch function')

~/.pyenv/versions/3.9.4/lib/python3.9/site-packages/ibis/expr/format.py in _fmt_value_table_node(op, aliases, **_)
    669     if not hasattr(op, 'table'):
    670         import pdb; pdb.set_trace()
--> 671     return f"{aliases[op.table.op()]}"
    672 
    673 

AttributeError: 'Union' object has no attribute 'table'
```

jcrist:
Fixes #4636.

### DIFF
--- a/ibis/expr/format.py
+++ b/ibis/expr/format.py
@@ -643,7 +643,7 @@ def _fmt_value_table_node(op: ops.TableNode, *, aliases: Aliases, **_: Any) -> s
     This function is called when a table is used in a value expression.
     An example is `table.count()`.
     """
-    return f"{aliases[op.table]}"
+    return f"{aliases[op]}"
 
 
 @fmt_value.register
--- a/ibis/tests/expr/test_format.py
+++ b/ibis/tests/expr/test_format.py
@@ -275,6 +275,21 @@ def test_tables_have_format_value_rules(cls):
     assert cls in ibis.expr.format.fmt_value.registry
 
 
+@pytest.mark.parametrize(
+    "f",
+    [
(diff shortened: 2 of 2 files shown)


gforsyth (approved):
looks good to me!

cpcloud (approved):

PR #4648, opened 2022-10-14
feat(duckdb): support registering external postgres tables with duckdb

jcrist:
This adds support for registering external postgres tables with duckdb, using the [postgres_scanner](https://duckdb.org/2022/09/30/postgres-scanner) extension. Example:

```python
con = ibis.connect("duckdb://")
# Register the table `mytable` in an external postgres database with duckdb
t = con.register("postgres://localhost:5432", "mytable")
# use `t` just like any other duckdb table
```

Filters are pushed down into the backing postgres engine when possible, limiting the amount of data loaded into duckdb to perform queries against the external tables. It's pretty slick that this works at all :).

A few warts:

- The latest release of the `postgres_scanner` extension is buggy. These bugs are fixed on master, but not yet released. I've written the code here to work as if they were released, but to test locally you need to build and load the extension yourself.
- This API works _ok_ with `register`, but there's some weirdness:
    - It forces the `postgres` table name and mirroring `duckdb` table name to be the same name. This may be undesired.
    - It doesn't expose the `schema` parameter in `postgres_scan_pushdown`. Right now this is hardcoded as `'public'`.

This also still needs tests.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -119,6 +119,18 @@ def _s3(full_path, table_name=None):
     return table_name, dataset, []
 
 
+@_generate_view_code.register(r"postgres(ql)?://.+", priority=10)
+def _postgres(uri, table_name=None):
+    if table_name is None:
+        raise ValueError("`table_name` is required when registering a postgres table")
+    quoted_table_name = _quote(table_name)
+    sql = (
+        f"CREATE OR REPLACE VIEW {quoted_table_name} AS "
+        f"SELECT * FROM postgres_scan_pushdown('{uri}', 'public', '{table_name}')"
+    )
+    return sql, table_name, ["postgres_scanner"]
+
+
 @_generate_view_code.register(r".+", priority=1)
 def _default(path, **kwargs):
     raise ValueError(
--- a/ibis/backends/duckdb/tests/test_register.py
+++ b/ibis/backends/duckdb/tests/test_register.py
@@ -229,3 +229,18 @@ def test_temp_directory(tmp_path):
 def test_s3_parquet(path):
     with pytest.raises(OSError):
         _generate_view_code(path)
+
+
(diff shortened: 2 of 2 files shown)


jcrist:
Here's a working demo joining a table of PyPI maintainers stored in postgres with a table of package dependencies stored as a local parquet file (both tables extracted from https://github.com/sethmlarson/pypi-data).

```python
import ibis
from ibis import _

# Connect to DuckDB
con = ibis.connect("duckdb://")

# Register data sources
# - deps is a local parquet file
# - maintainers is a remote postgres table
deps = con.register("deps.parquet")
maintainers = con.register("postgres://postgres:postgres@localhost:5432", "maintainers")

# Find all packages that depend directly on things I maintain
my_dependents = (
    deps
    .join(
        maintainers.filter(_.name == "jcrist"),
        [("dep_name", "package_name")]
    )
    .select(package="dep_name", dependent="package_name_x")
    .distinct()
)

# Execute, returning a pyarrow table
result = my_dependents.to_pyarrow()
```

Execution only happens in the final `to_pyarrow` call. The filter on maintainer name is pushed down into the backing postgres database, only loading the necessary subset into duckdb.

cpcloud:
This feature looks great.

Anything else to do besides a bit of testing?

Regarding the table names being the same, I think we can leave that as not implemented until someone requests it.

Regarding the schema being hardcoded to `public`, we may want to at least allow a keyword argument for that.

jcrist:
Sure, I can add some tests.

> Regarding the schema being hardcoded to public, we may want to at least allow a keyword argument for that.

I'm tempted to punt on this for now if that's ok.

jcrist:
Ok, this has been updated with master, and now includes some tests. I can confirm this works with the latest duckdb release on postgres tables that both do and don't require quoting.

cpcloud (approved):
LGTM!

PR #4650, opened 2022-10-14
fix: change default limit to None

### LINKED ISSUE #2017: feat: warn when limit reached in expr.execute()
When executing an expression a limit is silently applied to what is returned (unless you pass `limit=None`). A user could accidentally continue with local analysis on an arbitrary subset without realizing it's a subset. I'd like an optional parameter/setting that causes it to write a warning to `stderr` or raise an exception if the limit was exceeded. 

Since the limit is currently implemented by adding a limit expression to the original query, it may be difficult to know whether the limit would have been exceeded. In that case, maybe it could just check to see if the number of returned records equals the limit and issue the warning then. It might not definitively mean that the limit would've been exceeded, but would still be useful to get a warning that the limit was potentially exceeded.

jcrist:
Previously Table/Column expressions had a default limit of 10,000 rows applied when `execute` was called. The existence of this limit was easy for users to miss (it's not obvious that it's applied, and the effects won't be seen for small data), leading to results sometimes being silently truncated to the first 10,000 rows.

We now change the default limit to `None`, meaning no limit. This seems like a safer default value, as it won't result in silent truncation. With the new support for streaming results back (e.g. `to_pyarrow_batches`) we also expect users will want to pull back larger result sets, the presence of a limit seems to harm more than help.

**This doesn't remove the presence of a default limit, just changes the default value**. Downstream systems relying on the presence of this limit to stop users from shooting themselves in the foot can still make use of it by manually configuring a new default value:

```python
# Restore the old default limit
ibis.options.sql.default_limit = 10_000
```

Fixes #2017

### DIFF
--- a/ibis/backends/duckdb/registry.py
+++ b/ibis/backends/duckdb/registry.py
@@ -144,6 +144,15 @@ def _regex_extract(t, op):
     return result
 
 
+def _json_get_item(t, op):
+    left, path = map(t.translate, op.args)
+    # Workaround for https://github.com/duckdb/duckdb/issues/5063
+    # In some situations duckdb silently does the wrong thing if
+    # the path is parametrized.
+    sa_path = sa.text(str(path.compile(compile_kwargs=dict(literal_binds=True))))
+    return left.op("->")(sa_path)
+
+
 def _strftime(t, op):
     format_str = op.format_str
     if not isinstance(format_str_op := format_str, ops.Literal):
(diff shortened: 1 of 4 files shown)


ibis/backends/duckdb/registry.py
```diff
@@ -144,6 +144,15 @@ def _regex_extract(t, op):
     return result
 
 
+def _json_get_item(t, op):
```
jcrist, line 147: This change turned up a bug in duckdb's JSON extract operator implementation (https://github.com/duckdb/duckdb/issues/5063), which only shows up for queries where the `path` provided to `->` is parametrized AND there are no other parameters in the query. Previously this was masked since an implicit limit was always present, which would provide a 2nd parameter, avoiding the bug.

cpcloud: Ah, fun. Nice find.

cpcloud:
I'll merge this after #4726, since that should address the `simulate_release` failure with no changes to this PR.

cpcloud (approved):
LGTM. Nice to see this finally addressed! Thank you!

PR #4667, opened 2022-10-18
refactor: don't sort new columns in mutate

### LINKED ISSUE #4656: refactor: remove `sorted` calls from mutation
I thought we had removed `sorted`, but I guess not. We should probably do that for 4.0.0.

_Originally posted by @cpcloud in https://github.com/ibis-project/ibis/pull/4649#discussion_r996311096_

Let's remove `sorted` from mutation calls since `dict`s are now ordered.

jcrist:
Fixes #4656.


cpcloud (approved):
LGTM! Thanks.

PR #4680, opened 2022-10-19
docs: improve the docstring of the generic connect method

### LINKED ISSUE #4675: docs(api): backend specific connection parameters in the docstring of the connect method
At the moment the docstring of the main public factory method for a given backend is not helpful because it does not document what are the backend-specific connection parameters:

```python
>>> help(ibis.duckdb.connect)
Help on method connect in module ibis.backends.base:
 
connect(*args, **kwargs) -> 'BaseBackend' method of ibis.backends.duckdb.Backend instance
    Connect to the database.
    
    Parameters
    ----------
    args
        Connection parameters
    kwargs
        Additional connection parameters
    
    Notes
    -----
    This returns a new backend instance with saved `args` and `kwargs`,
    calling `reconnect` is called before returning.
    
    Returns
    -------
    BaseBackend
        An instance of the backend
```

However the docstring of the concrete `do_connect` method is more helpful:

```python
>>> help(ibis.duckdb.do_connect)
Help on method do_connect in module ibis.backends.duckdb:
 
do_connect(database: 'str | Path' = ':memory:', path: 'str | Path' = None, read_only: 'bool' = False, **config: 'Any') -> 'None' method of ibis.backends.duckdb.Backend instance
    Create an Ibis client connected to a DuckDB database.
    
    Parameters
    ----------
    database
        Path to a duckdb database.
    read_only
        Whether the database is read-only.
    config
        DuckDB configuration parameters. See the [DuckDB configuration
        documentation](https://duckdb.org/docs/sql/configuration) for
        possible configuration values.
    
    Examples
    --------
    >>> import ibis
    >>> ibis.duckdb.connect("database.ddb", threads=4, memory_limit="1GB")
```

and the example actually documents the usage of the `connect` method instead of the `do_connect` method.

I think the `ibis.duckdb.connect.__doc__` attribute should be updated one way or another to either include the content of `ibis.duckdb.do_connect.__doc__` or at least refer to the `do_connect` method of the backend for more details.

ogrisel:
Fixes #4675.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -387,8 +387,9 @@ class BaseBackend(abc.ABC, ResultHandler):
 
         Notes
         -----
-        This returns a new backend instance with saved `args` and `kwargs`,
-        calling `reconnect` is called before returning the backend instance.
+        This creates a new backend instance with saved `args` and `kwargs`,
+        then calls `reconnect` and finally returns the newly created and
+        connected backend instance.
 
         Returns
         -------
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -379,14 +379,16 @@ class BaseBackend(abc.ABC, ResultHandler):
         Parameters
         ----------
         args
-            Connection parameters
+            Mandatory connection parameters, see the docstring of `do_connect`
+            for details.
         kwargs
-            Additional connection parameters
+            Extra connection parameters, see the docstring of `do_connect` for
+            details.
 
         Notes
         -----
         This returns a new backend instance with saved `args` and `kwargs`,
-        calling `reconnect` is called before returning.
+        calling `reconnect` is called before returning the backend instance.
 
         Returns
         -------


ogrisel:
BTW, I don't see the point of the `reconnect` method. It seems like an unnecessary indirection to me.

ogrisel:
In #4675 I hinted that it might be possible to dynamically define the docstring of the connect method but I am not sure how to do that in a robust way so I went to just refer the reader to the other docstring by themselves.

Maybe there is a smarter and even more user friendly way to dynamically update the docstring of the `connect` method but I think this method is already a net UX improvement and does not risk introducing any regression.

saulpw:
Hey @ogrisel, I added `reconnect()` for `vdsql`, which creates a connection in one thread, and then needs to call `.execute()` on an equivalent connection in another thread (some backends do not allow using a connection across multiple-threads).  This allows me to make a copy of the connection object and then call `.reconnect()` in order to make it valid in the new thread.

saulpw (approved):

ogrisel:
Re-reading the diff of the PR I noticed that the last sentence was not really making sense. I pushed a new commit.

ogrisel:
The code coverage decrease is not caused by this PR.

cpcloud (approved):
LGTM, thanks!

PR #4681, opened 2022-10-19
chore(nix): override duckdb to respect NIX_BUILD_CORES

gforsyth:
DuckDB makes use of `multiprocessing.cpu_count()` and it locks up my machine whenever it has to build by very efficiently consuming all my cores.  This hack-patch makes it respect `NIX_BUILD_CORES` so that doesn't happen.


cpcloud (approved):
LGTM

Issue #4677, opened 2022-10-19
bug: RecursionError raised when attempting to create a duckdb table from a pandas-backed exp


ogrisel:
Reading the documentation on [Creating and Inserting Data](https://ibis-project.org/docs/3.2.0/tutorial/05-IO-Create-Insert-External-Data/), I was under the impression that it would be possible to create a new table from an arbitrary ibis expression, even if it did not stem from the same connection/backend.

Hence I wanted to use pandas to create an in-memory table and then store that into a duckdb database:

```python
import ibis
import pandas as pd

df = pd.DataFrame(
    {

        "g": ["a", "a", "a", "a", "a"],
        "x": [0, 1, 2, 3, 4],
        "y": [3, 2, 0, 1, 1],
    }
)
t_pandas = ibis.pandas.connect({"t": df}).table("t")
duckdb_conn = ibis.duckdb.connect()
duckdb_conn.create_table("t", t_pandas)
```

However this results in `RecursionError: maximum recursion depth exceeded`:

<details>

```python
Traceback (most recent call last):
  File "/Users/ogrisel/tmp/ibis_pandas_window.py", line 16, in <module>
    duckdb_conn.create_table("t", t_pandas)
  File "/Users/ogrisel/code/ibis/ibis/backends/base/sql/alchemy/__init__.py", line 215, in create_table
    method = self._get_insert_method(expr)
  File "/Users/ogrisel/code/ibis/ibis/backends/base/sql/alchemy/__init__.py", line 219, in _get_insert_method
    compiled = self.compile(expr)
  File "/Users/ogrisel/code/ibis/ibis/backends/base/sql/__init__.py", line 344, in compile
    return self.compiler.to_ast_ensure_limit(expr, limit, params=params).compile()
  File "/Users/ogrisel/code/ibis/ibis/backends/base/sql/compiler/base.py", line 39, in compile
    compiled_queries = [q.compile() for q in self.queries]
  File "/Users/ogrisel/code/ibis/ibis/backends/base/sql/compiler/base.py", line 39, in <listcomp>
    compiled_queries = [q.compile() for q in self.queries]
  File "/Users/ogrisel/code/ibis/ibis/backends/base/sql/alchemy/query_builder.py", line 176, in compile
    frag = self._compile_table_set()
  File "/Users/ogrisel/code/ibis/ibis/backends/base/sql/alchemy/query_builder.py", line 203, in _compile_table_set
    result = helper.get_result()
[...]

PR #4679, opened 2022-10-19
docs: added memtable howto guide

p-a-a-a-trick:

PR #4682, opened 2022-10-19
docs: document how to create a table from a pandas dataframe using ibis.memtable

ogrisel:
Addresses the concern raised in https://github.com/ibis-project/ibis/issues/4677#issuecomment-1283920881.

Note: I decided to update the "Setup" cells to only use Python instead of curl because the `TEMPDIR` does not seem to match on my macos system. Other tutorials might benefit from a similar treatment.


ogrisel:
This is related to how `memtable` is presented in #4679 but I believe it's also good to speak about pandas dataframe interop directly in the "Create and Insert External Data" tutorial.

cpcloud (approved):

cpcloud:
Sweet! Thank you.

docs/tutorial/05-IO-Create-Insert-External-Data.ipynb
```diff
@@ -16,11 +16,25 @@
   },
```
saulpw, line 16: <pre class="ql-syntax" spellcheck="false">Line #2.    if students_db_file.exists():
</pre><p>Can we use <code>tempfile.TemporaryDirectory</code>? then we shouldn't have do this exists check and unlink (here and above).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/4682/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

Issue #4683, opened 2022-10-19
Inconsistent support of pandas.DataFrame in `create_table` and `insert`


ogrisel:
Assuming the following setup:

```python
>>> import ibis
>>> import pandas as pd

>>> df = pd.DataFrame({"x": range(3)})
>>> conn = ibis.duckdb.connect()
```

I find it non-intuitive that the following fails:

```python
>>> conn.create_table("t", df)
Traceback (most recent call last):
  Input In [10] in <cell line: 6>
    ibis.duckdb.connect().create_table("t", df)
  File ~/code/ibis/ibis/backends/base/sql/alchemy/__init__.py:200 in create_table
    schema = expr.schema()
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/pandas/core/generic.py:5907 in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'DataFrame' object has no attribute 'schema'
```

UPDATE: the above has been fixed in #4684.

while the following works:

```python
>>> conn.create_table("t", ibis.memtable(df))
>>> t = conn.table("t")
>>> t.count().execute()
3
```

and furthermore `insert` accepts inserting from a pandas dataframe without complaining:

```python
>>> conn.insert("t", df)
>>> t.count().execute()
6
```

but it does complain when trying to insert from a memtable of the very same dataframe:

```python
>>> conn.insert("t", ibis.memtable(df))
Traceback (most recent call last):
  Input In [16] in <cell line: 1>
    conn.insert("t", ibis.memtable(df))
  File ~/code/ibis/ibis/backends/base/sql/alchemy/__init__.py:493 in insert
    from_table_expr.compile(),
  File ~/code/ibis/ibis/expr/types/core.py:302 in compile
    return self._find_backend().compile(
  File ~/code/ibis/ibis/expr/types/core.py:239 in _find_backend
    raise IbisError(
IbisError: Expression depends on no backends, and found no default
```

I think it would be convenient to have `conn.create_table` and `conn.insert` to always accept both `ibis.memtable` and `pandas.DataFrame` objects.

ogrisel:
If this is fixed one way or another, one my want to update #4682 accordingly.

gforsyth:
Thanks for reporting @ogrisel!  Yes the behavior should be consistent across these two methods.

`create_table` failing when passed a DataFrame is a bug -- the docstring of `create_table`, at least, indicates that we accept pandas dataframes as input

PR #4684, opened 2022-10-19
fix(sqlalchemy): allow passing pd.DataFrame to create

gforsyth:
The docstring of `create` indicates this should work but it relies on
there being a `schema` method, which isn't true of dataframes.

If it's a dataframe, we wrap it in a `memtable` and then continue.

xref #4683 -- although this doesn't handle the other inconsistency raised there


jcrist (approved):
:rocket:

cpcloud (approved):
LGTM, thanks!

PR #4692, opened 2022-10-20
feat: support both `postgresql://` and `postgres://` in `ibis.connect`

jcrist:
This PR does a few things:

- Enables support for using `ibis.connect` to connect to postgres URLs starting with either `postgres://` or `postgresql://`, since postgres itself accepts both forms.
- Simplifies the implementation of `ibis.connect`. The previous tiered dispatch system was hard to read IMO and hid several bugs ("ravioli code").
- Expands the test suite for `ibis.connect` to better cover the functionality here.

I recognize the removal of the dispatch system may not be desired, since it forces external backends to only work with a standardized URL scheme (which honestly may be fine). I'm fine with reverting that change, but would like us to then be clear about which parts of `ibis.connect`'s dispatch system are intended to be a public extension interface rather than just an implementation detail.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -31,7 +31,6 @@ import ibis.config
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
 import ibis.util as util
-from ibis.common.dispatch import RegexDispatcher
 
 __all__ = ('BaseBackend', 'Database', 'connect')
 
@@ -734,9 +733,6 @@ class BaseBackend(abc.ABC, ResultHandler):
         )
 
 
-_connect = RegexDispatcher("_connect")
-
-
 @functools.lru_cache(maxsize=None)
 def _get_backend_names() -> frozenset[str]:
     """Return the set of known backend names.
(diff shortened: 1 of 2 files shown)


ibis/backends/base/__init__.py
```diff
@@ -753,125 +749,83 @@ def _get_backend_names() -> frozenset[str]:
...
+
+    # Merge explicit kwargs with query string, explicit kwargs
+    # taking precedence
+    kwargs = dict(urllib.parse.parse_qsl(parsed.query), **kwargs)
+
+    if scheme == "file":
+        path = parsed.netloc + parsed.path
+        if path.endswith(".duckdb"):
+            return ibis.duckdb.connect(path, **kwargs)
+        elif path.endswith(".sqlite"):
```
saulpw, line 793: Can we add `or .endswith(".db")` and I'll cancel my PR #4695?

saulpw (approved):
I was *just* complaining about this!  Thanks for doing this, it looks much better to me.

cpcloud:
Can backends register a url scheme independent of ibis with this change?

ibis/backends/base/__init__.py
jcrist: That's easy to do, sure.

jcrist: Resolved.

> Can we add or `.endswith(".db")`

I only recently learned about this, but `.endswith`/`.startswith` can take tuples of suffixes/prefixes, so you don't have to manually do the `or`. `path.endswith((".sqlite", ".db"))` does it for you.

cpcloud (approved):
LGTM! Thanks.

PR #4713, opened 2022-10-25
fix(ux): improve error messages for rlz.comparable failures

### LINKED ISSUE #4710: feat: better tracebacks for type comparison errors
I have an expression something like this, but with more comparisons and more operations, generally (so harder to visually parse)
```python
expr = basics.join(ratings, basics.tconst == ratings.tconst).filter(
                    [_.titleType == "movie", _.numVotes > 10, _.averageRating > 3.5]
                    )
```

When I wrote that, I didn't know that `numVotes` was actually a string column, so I get this error:

```python
         64 if not rlz.comparable(left, right):
    ---> 65     raise TypeError(
         66         f'Arguments with datatype {left.output_dtype} and '
         67         f'{right.output_dtype} are not comparable'
         68     )
         69 super().__init__(left=left, right=right)
    
    TypeError: Arguments with datatype string and int8 are not comparable
```

which is definitely valid, but even if I scroll all the way to the top of the traceback, all I'm going to find is that the offending line is the _single line_:

```python
----> 1 expr = basics.join(ratings, basics.tconst == ratings.tconst).filter(
      2     [_.titleType == "movie", _.numVotes > 10, _.averageRating > 3.5]
      3 )
```

So I'm thinking we can add in the names of these expressions to the error message.
I think we can probably iterate a bit on how the information is presented, but I think even without the exact logical operation that led to this error, giving the user a pointer to which expression names are involved should save a bunch of time when working interactively.

```python
         64 if not rlz.comparable(left, right):
    ---> 65     raise TypeError(
         66         f'Arguments "{left.name}" ({left.output_dtype}) and '
         67         f'"{right.name}" ({right.output_dtype}) are not comparable'
         68     )
         69 super().__init__(left=left, right=right)
    
    TypeError: Arguments "numVotes" (string) and "10" (int8) are not comparable
```

Are there cases where the expression would be unnamed?  Is it weird to have the literal `"10"` presented as the argument name?

gforsyth:
Provides a little more information to the user about which expressions are causing the rule failure in logical comparisons and the Simple and Searched case builders.

Resolves #4710

### DIFF
--- a/ibis/expr/builders.py
+++ b/ibis/expr/builders.py
@@ -66,7 +66,10 @@ class SimpleCaseBuilder(TypedCaseBuilder):
         result_expr = rlz.any(result_expr)
 
         if not rlz.comparable(self.base, case_expr):
-            raise TypeError('Base expression and passed case are not ' 'comparable')
+            raise TypeError(
+                f'Base expression {rlz._arg_type_error_format(self.base)} and '
+                f'case {rlz._arg_type_error_format(case_expr)} are not comparable'
+            )
 
         cases = list(self.cases)
         cases.append(case_expr)
--- a/ibis/expr/operations/logical.py
+++ b/ibis/expr/operations/logical.py
@@ -63,8 +63,8 @@ class Comparison(Binary):
         """
         if not rlz.comparable(left, right):
             raise TypeError(
-                f'Arguments with datatype {left.output_dtype} and '
-                f'{right.output_dtype} are not comparable'
+                f'Arguments {rlz._arg_type_error_format(left)} and '
+                f'{rlz._arg_type_error_format(right)} are not comparable'
             )
         super().__init__(left=left, right=right)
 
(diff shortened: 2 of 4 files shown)


ibis/expr/builders.py
```diff
@@ -66,7 +66,10 @@ def when(self, case_expr, result_expr):
         result_expr = rlz.any(result_expr)
 
         if not rlz.comparable(self.base, case_expr):
-            raise TypeError('Base expression and passed case are not ' 'comparable')
+            raise TypeError(
+                f'Base expression {rlz._arg_type_error_format(self.base)} and '
+                f'case {rlz._arg_type_error_format(case_expr)} are not comparable'
```
cpcloud, line 71: The smallest of tests here might be a good idea so that we can ensure this doesn't regress.

gforsyth: Agreed.  Pushed up.

cpcloud: Thanks!

cpcloud (approved):
LGTM, thanks!

PR #4726, opened 2022-10-26
ci: relax lockfile freshness check

cpcloud:
This PR (finally) relaxes the poetry lockfile freshness check.

[poetry 1.2.0 introduces `poetry check --lock`](https://python-poetry.org/blog/announcing-poetry-1.2.0/#lock)
which allows us to relax the lock file check to a semantic of "consistent"
rather than our current semantic of "freshest".

This should alleviate developers having to clear their cache to update the lockfile.


gforsyth (approved):
**SMASHES APPROVE BUTTON**

cpcloud:
Merging on green.

PR #4733, opened 2022-10-27
fix(enum): add sqlalchemy types for Enum/Category

saulpw:
Need to figure out an appropriate test that will work for all backends that support an enum type.

### DIFF
--- a/ci/schema/duckdb.sql
+++ b/ci/schema/duckdb.sql
@@ -45,11 +45,14 @@ CREATE OR REPLACE TABLE awards_players (
     notes TEXT
 );
 
+CREATE TYPE enum_t AS ENUM ('alice', 'bob', 'carol');
+
 CREATE OR REPLACE TABLE functional_alltypes (
     "index" BIGINT,
     "Unnamed: 0" BIGINT,
     id INTEGER,
     bool_col BOOLEAN,
+    enum_col enum_t,
     tinyint_col SMALLINT,
     smallint_col SMALLINT,
     int_col INTEGER,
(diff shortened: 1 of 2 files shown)


ci/schema/duckdb.sql
```diff
@@ -45,11 +45,14 @@ CREATE OR REPLACE TABLE awards_players (
...
 );
 
+CREATE TYPE enum_t AS ENUM ('alice', 'bob', 'carol');
+
 CREATE OR REPLACE TABLE functional_alltypes (
     "index" BIGINT,
     "Unnamed: 0" BIGINT,
     id INTEGER,
     bool_col BOOLEAN,
+    enum_col enum_t,
```
cpcloud, line 55: You can't add a column to this table without modifying the data.

I would suggest something like changing the `string_col` column to be an enum of `0 to 9 inclusive, which reflects the data in `ci/ibis-testing-data/functional_alltypes.csv`.

ibis/backends/base/sql/alchemy/datatypes.py
```diff
@@ -82,6 +82,8 @@ def table_from_schema(name, meta, schema, database: str | None = None):
 # TODO(cleanup)
 ibis_type_to_sqla = {
     dt.Null: sa.types.NullType,
+    dt.Enum: sa.Enum,
+    dt.Category: sa.Enum,
```
cpcloud, line 86: Eventually I want to deprecate `dt.Category` in favor of `dt.Enum` only, so ideally we don't add net new functionality that uses `Category`.

ci/schema/duckdb.sql
```diff
@@ -45,11 +45,14 @@ CREATE OR REPLACE TABLE awards_players (
     notes TEXT
 );
 
+CREATE TYPE enum_t AS ENUM ('alice', 'bob', 'carol');
```
cpcloud, line 48:
```suggestion
DROP TYPE IF EXISTS enum_t CASCADE;
CREATE TYPE enum_t AS ENUM ('alice', 'bob', 'carol');
```

This needs to be recreated on every test run, so the type needs to be dropped if exists.

cpcloud:
Closing this out for now!

PR #4736, opened 2022-10-27
docs: fix snowflake doc link in readme.md

lostmygithubaccount:
the link to the doc points to `/latest/`, which 404s. change to `/dev/` to fix


saulpw:
Thanks for catching this, @lostmygithubaccount!  I'm not sure if this is how we want to solve this issue; ideally `/latest/` would redirect to `/dev/` if the page isn't available in the latest release.  (Otherwise we'll have to remember to update this in the next release anyway).

@cpcloud What do you think, should we redirect latest to dev always?  Or is it possible to conditionally redirect?

cpcloud:
Let's redirect to dev until we decide to promote the snowflake backend to non-experimental.

cpcloud (approved):

cpcloud:
Thanks @lostmygithubaccount, merging on green!

PR #4748, opened 2022-10-31
docs: update readme

### LINKED ISSUE #4665: docs: update README to include clear project value proposition
I'm having trouble figuring out the value proposition of this project after reading the README & looking at the project website.

Here's the current description:

> Ibis is a Python library to help you write expressive analytics at any scale, small to large. Its goal is to simplify analytical workflows and make you more productive.

Looks like you can create an Ibis table per the [pandas Documentation page](https://ibis-project.org/docs/3.2.0/backends/Pandas/).  But why should I create an Ibis table?  What value does this provide me?

p-a-a-a-trick:
Resolves #4665.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 ## What is Ibis?
 
-Ibis is a Python project that provides a lightweight, universal interface for data wrangling. It helps Python users explore and transform data of any size, stored anywhere.
+Ibis is a Python library that provides a lightweight, universal interface for data wrangling. It helps Python users explore and transform data of any size, stored anywhere.
 
 Ibis has three primary pieces:
 1. **A dataframe API for Python**.
--- a/README.md
+++ b/README.md
@@ -11,76 +11,41 @@
 
 Ibis is a Python library that provides a lightweight, universal interface for data wrangling. It helps Python users explore and transform data of any size, stored anywhere.
 
-Ibis has three primary feature sets.
-
-### A dataframe API for Python
-
-Python users can write Ibis code to manipulate tabular data.
-
-### Utilities for connecting to 10+ query engines
-
-Wherever data is stored, data scientists can use Ibis as their API of choice to
-communicate with any of those query engines.
-
-### Deferred execution
-
-Execution of code is pushed to the query engine. Users execute code at the
-speed of their backend, not their local computer.
+Ibis has three primary pieces:
+1. **A dataframe API for Python**.
+This means that Python users can write Ibis code to manipulate tabular data.
+2. **Utilities for connecting to 10+ query engines.**
+This means that wherever data is stored, data scientists can use Ibis as their API of choice to communicate with any of those query engines. 
+3. **Deferred execution**.
+Ibis uses deferred execution, meaning that execution of code is pushed to the query engine.
+This means users can execute at the speed of their backend, not their local computer.
 
 ## Why Use Ibis?
 
 Ibis aims to be a future-proof solution to interacting with data using Python and can accomplish this goal through its main features:
 
-### Familiar API
-
-Ibis’s API design borrows from popular APIs like pandas and dplyr that most users already know and like to use.
-
-### Consistent syntax
-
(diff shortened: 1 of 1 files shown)


p-a-a-a-trick:
@marlenezw

cpcloud:
@p-a-a-a-trick @marlenezw I pushed up a few commits, let me know whether you like them. The primary change is to use level-3 headings instead of a bulleted list.

README.md
```diff
@@ -1,31 +1,56 @@
...
-```sh
-conda install ibis-framework -c conda-forge
-```
+## Why Use Ibis?
+
+Ibis aims to be a future-proof solution to interacting with data using Python and can accomplish this goal through its main features:
+
+- **Familiar API**: Ibis’s API design borrows from popular APIs like pandas and dplyr that most users already know and like to use.
+- **Consistent syntax**: After learning how to use Ibis, users can use the same API and concepts no matter where their data lives.
+  **Learn once, write anywhere**.
```
datametrician, line 30: I guess we're going with this. I'm growing more fond of this but I still think "learn" is a bit passive. I'd like to push us to think about something more aggressive without alienating users. Maybe:
*A Universal Python API
*Standardized Python Query Language
*Python made Portable
Again, feel free to over rule me, but learn in my head highlights the lack of guarantees, when in reality there are clusters of functionality that are very portable, and those clusters needs to be articulated better.

cpcloud: How about `Portable Python Analytics`?

p-a-a-a-trick: Yeah I'm not a big fan of "learn" if I'm being honest; if we went with "Learn once, write anywhere" as a bit then the first word someone runs into wrt Ibis will be "learn", which sounds like a homework assignment.

Learning isn't a bad thing obviously, and it's something a new user will be doing anyways, but it, to me, feels like an unwelcoming hurdle to Ibis vs demonstrating that Ibis is something you can jump into and immediately get long-term value out of _while_ learning a universal interface to tabular data.

For this particular bullet, does this make sense?

<hr>

- **Consistent syntax**: After learning how to use Ibis, users can use the same API and concepts no matter where their data lives. **Ibis aims to be universal python API for tabular data, big or small.**.

<hr>

Or do we think there's a nuance that this is eliding?

cpcloud: I think you can probably just keep the last sentence.

README.md
```diff
@@ -1,31 +1,55 @@
...
 
-```sh
-conda install ibis-framework -c conda-forge
-```
+## Why Use Ibis?
+
+Ibis aims to be a future-proof solution to interacting with data using Python and can accomplish this goal through its main features:
+
+- **Familiar API**: Ibis’s API design borrows from popular APIs like pandas and dplyr that most users already know and like to use.
+- **Consistent syntax**:  Ibis aims to be universal python API for tabular data, big or small.
```
p-a-a-a-trick, line 29: @datametrician , @cpcloud , @marlenezw thoughts?

cpcloud:
Frankly, this PR is ready to merge as is IMO.

1. There's nothing incorrect here.
2. It's a huge improvement over the existing README.md

Perfect is the enemy of the good. Ship it.

README.md
saulpw: @datametrician said:
> Again, feel free to over rule me, but learn in my head highlights the lack of guarantees, when in reality there are clusters of functionality that are very portable, and those clusters could be articulated better.

We're not necessarily making a long-term decision, just incremental progress to push the ball forward.  It's too easy to get caught up in all these details when reviewing a PR; the ideal is to take this PR and then whenever you/we/whomever has some better words, they can submit a new PR and we'll take that one too.  Even if it's tomorrow :)

cpcloud (approved):

marlenezw:
Sorry for the lack of comments from me I was traveling back home yesterday! Happy this was merged and will probably add a few more improvements as another PR soon!

README.md
datametrician: Noted @saulpw and understood.

I love Portable Python Analytics @cpcloud!

Issue #4444, opened 2022-08-30
feat: file outputs


jcrist:
Ibis currently lacks the ability to write to output files (`csv`, `parquet`, ...). One way this could be handled is by streaming the results back (#4443) and handling all IO in python using the relevant `pyarrow`/`fsspec` apis.

However, some backends (duckdb, dask, pyspark, ...) may have a more efficient way of writing certain formats natively. It'd be good to support using the backend-native mechanisms when possible.

jcrist:
One way to architect this would be to define a generic "Sink" abstraction, and make use of it to build different IO routines:

```python
class Sink:
    def __init__(self, **kwargs: Any) -> None:
        """Sink options (output directory, ...) would be passed in to the
        constructor"""
        pass

    def setup(self) -> None:
        """Called at the start of processing to setup the sink"""
        pass

    def handle_batch(self, batch: pyarrow.RecordBatch) -> None:
        """Called during results handling on every RecordBatch in the table."""
        pass

    def finish(self, succeeded: bool) -> Any:
        """Called at the end of processing to cleanup the sink.

        If an error occurred at any point during processing, `succeeded` will
        be false. The result of `finish` is returned to the caller.
        """
        pass
```

For example, a simple parquet sink might look like:

```python
class ParquetSink(Sink):
    def __init__(self, directory: str) -> None:
        self.directory = directory
        self.count = 0

    def setup(self) -> None:
        os.makedirs(self.directory, exist_ok=True)

    def handle_batch(self, batch: pyarrow.RecordBatch) -> None:
        table = pa.Table.from_batches([batch])
        path = os.path.join(self.directory, f"part-{self.count}.parquet")
        pa.parquet.write_table(table, path)
        self.count += 1
```

The `ParquetSink` class itself wouldn't be primarily user-facing, but would rather be wrapped in a nicer user-facing method:

```python
class Table:
    ...

    def to_sinks(self, *sinks: Sink) -> Any:
        """Write to multiple sinks"""
        # todo

    def to_parquet(self, directory: str) -> None:
        self.to_sinks(ParquetSink(directory))
```

The `to_sinks` method would handle the execution and dispatch to the various sinks. In its simplest form its a loop around `to_pyarrow_batches()` (from #4443), calling `handle_batch` on each sink.

However, we also want to be able to possibly dispatch the sink operation to be handled directly in a backend. One way to do that would be to expose a new `multipledispatch` function:

```python
@dispatch(Backend, Sink)
[...]

PR #4759, opened 2022-11-02
feat(api): add `read` top level function for interactive analysis with CSV and Parquet files

cpcloud:
This PR adds a `read` top level API function for reading the following kinds of things:

1. Local CSV, TSV, TXT and Parquet files. CSV/TSV/TXT files can be gzip compressed
1. Remote versions of any of those
1. Globs of 1., mixing compressed/uncompressed is allowed.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -34,9 +34,9 @@ _dialect = sa.dialects.postgresql.dialect()
 _gen_table_names = (f"registered_table{i:d}" for i in itertools.count())
 
 
-def _name_from_path(path: str) -> str:
+def _name_from_path(path: str | Path) -> str:
     # https://github.com/duckdb/duckdb/issues/5203
-    return path.replace(".", "_")
+    return str(path).replace(".", "_")
 
 
 def _name_from_dataset(dataset: pa.dataset.FileSystemDataset) -> str:
@@ -117,8 +117,7 @@ def _s3(full_path, table_name=None):
 
     dataset = ds.dataset(full_path)
     table_name = table_name or _name_from_dataset(dataset)
-    quoted_table_name = _quote(table_name)
-    return quoted_table_name, dataset, ()
+    return table_name, dataset, []
 
 
 @_generate_view_code.register(r".+", priority=1)
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -34,9 +34,9 @@ _dialect = sa.dialects.postgresql.dialect()
 _gen_table_names = (f"registered_table{i:d}" for i in itertools.count())
 
 
-def _name_from_path(path: Path) -> str:
-    base, *_ = path.name.partition(os.extsep)
-    return base.replace("-", "_")
+def _name_from_path(path: str) -> str:
+    # https://github.com/duckdb/duckdb/issues/5203
+    return path.replace(".", "_")
 
 
 def _name_from_dataset(dataset: pa.dataset.FileSystemDataset) -> str:
@@ -47,49 +47,67 @@ def _quote(name: str):
     return _dialect.identifier_preparer.quote(name)
 
 
-@_generate_view_code.register(r"parquet://(?P<path>.+)", priority=10)
-def _parquet(_, path, table_name=None, **kwargs):
-    path = Path(path).absolute()
-    table_name = table_name or _name_from_path(path)
-    quoted_table_name = _quote(table_name)
-    args = [f"'{str(path)}'"]
-    if kwargs:
-        args.extend([f"{k}={v}" for k, v in kwargs.items()])
+def _get_scheme(scheme):
+    if scheme is None or scheme == "file://":
+        return ""
+    return scheme
+
+
+def _format_kwargs(kwargs):
     return (
-        f"CREATE OR REPLACE VIEW {quoted_table_name} as SELECT * "
-        f"from read_parquet({', '.join(args)})",
-        table_name,
+        f"{k}='{v}'" if isinstance(v, str) else f"{k}={v!r}" for k, v in kwargs.items()
     )
 
 
-@_generate_view_code.register(r"csv(?:\.gz)?://(?P<path>.+)", priority=10)
-def _csv(_, path, table_name=None, **kwargs):
-    path = Path(path).absolute()
-    table_name = table_name or _name_from_path(path)
+@_generate_view_code.register(r"parquet://(?P<path>.+)", priority=13)
+def _parquet(_, path, table_name=None, scheme=None, **kwargs):
+    scheme = _get_scheme(scheme)
+    if not scheme:
+        path = os.path.abspath(path)
+    if not table_name:
+        table_name = _name_from_path(path)
     quoted_table_name = _quote(table_name)
-    # AUTO_DETECT and COLUMNS collide, so we set AUTO_DETECT=True
+    args = [f"'{scheme}{path}'", *_format_kwargs(kwargs)]
+    code = f"""\
+CREATE OR REPLACE VIEW {quoted_table_name} AS
+SELECT * FROM read_parquet({', '.join(args)})"""
+    return code, table_name, ["parquet"] + ["httpfs"] if scheme else []
+
+
+@_generate_view_code.register(r"(c|t)sv://(?P<path>.+)", priority=13)
+def _csv(_, path, table_name=None, scheme=None, **kwargs):
+    scheme = _get_scheme(scheme)
+    if not scheme:
+        path = os.path.abspath(path)
+    if not table_name:
+        table_name = _name_from_path(path)
+    quoted_table_name = _quote(table_name)
+    # auto_detect and columns collide, so we set auto_detect=True
     # unless COLUMNS has been specified
-    args = [f"'{str(path)}'"]
-    args.extend(
-        [
-            f"AUTO_DETECT="
-            f"{kwargs.pop('AUTO_DETECT', False if 'COLUMNS' in kwargs else True)}"
-        ]
-    )
-    if kwargs:
-        args.extend([f"{k}={v}" for k, v in kwargs.items()])
-    return (
-        f"CREATE OR REPLACE VIEW {quoted_table_name} as SELECT * "
-        f"from read_csv({', '.join(args)})",
-        table_name,
-    )
+    args = [
+        f"'{scheme}{path}'",
+        f"auto_detect={kwargs.pop('auto_detect', 'columns' not in kwargs)}",
+        *_format_kwargs(kwargs),
+    ]
+    code = f"""\
+CREATE OR REPLACE VIEW {quoted_table_name} AS
+SELECT * FROM read_csv({', '.join(args)})"""
+    return code, table_name, ["httpfs"] if scheme else []
 
 
-@_generate_view_code.register(r"(?:file://)?(?P<path>.+)", priority=9)
-def _file(_, path, table_name=None, **kwargs):
-    num_sep_chars = len(os.extsep)
-    extension = "".join(Path(path).suffixes)[num_sep_chars:]
-    return _generate_view_code(f"{extension}://{path}", table_name=table_name, **kwargs)
+@_generate_view_code.register(
+    r"(?P<scheme>(?:file|https?)://)?(?P<path>.+?\.((?:c|t)sv|txt)(?:\.gz)?)",
+    priority=12,
+)
+def _csv_file_or_url(_, path, table_name=None, **kwargs):
+    return _csv(f"csv://{path}", path=path, table_name=table_name, **kwargs)
+
+
+@_generate_view_code.register(
+    r"(?P<scheme>(?:file|https?)://)?(?P<path>.+?\.parquet)", priority=12
+)
+def _parquet_file_or_url(_, path, table_name=None, **kwargs):
+    return _parquet(f"parquet://{path}", path=path, table_name=table_name, **kwargs)
 
 
 @_generate_view_code.register(r"s3://.+", priority=10)
@@ -100,17 +118,16 @@ def _s3(full_path, table_name=None):
     dataset = ds.dataset(full_path)
     table_name = table_name or _name_from_dataset(dataset)
     quoted_table_name = _quote(table_name)
-    return quoted_table_name, dataset
+    return quoted_table_name, dataset, ()
 
 
 @_generate_view_code.register(r".+", priority=1)
-def _default(_, **kwargs):
+def _default(path, **kwargs):
     raise ValueError(
-        """
-Unrecognized filetype or extension.
-Valid prefixes are parquet://, csv://, s3://, or file://
+        f"""Unrecognized file type or extension: {path}.
 
-Supported filetypes are parquet, csv, and csv.gz
+Valid prefixes are parquet://, csv://, tsv://, s3://, or file://
+Supported file extensions are parquet, csv, tsv, txt, csv.gz, tsv.gz, and txt.gz
     """
     )
 
@@ -180,6 +197,15 @@ class Backend(BaseAlchemyBackend):
             )
         )
         self._meta = sa.MetaData(bind=self.con)
+        self._extensions = set()
+
+    def _load_extensions(self, extensions):
+        for extension in extensions:
+            if extension not in self._extensions:
+                with self.con.connect() as con:
+                    con.execute(f"INSTALL '{extension}'")
+                    con.execute(f"LOAD '{extension}'")
+                self._extensions.add(extension)
 
     def register(
         self,
@@ -210,7 +236,10 @@ class Backend(BaseAlchemyBackend):
             The just-registered table
         """
         if isinstance(source, str) and source.startswith("s3://"):
-            table_name, dataset = _generate_view_code(source, table_name=table_name)
+            table_name, dataset, extensions_required = _generate_view_code(
+                source, table_name=table_name
+            )
+            self._load_extensions(extensions_required)
             # We don't create a view since DuckDB special cases Arrow Datasets
             # so if we also create a view we end up with both a "lazy table"
             # and a view with the same name
@@ -221,9 +250,10 @@ class Backend(BaseAlchemyBackend):
                 # explicitly.
                 cursor.cursor.c.register(table_name, dataset)
         elif isinstance(source, (str, Path)):
-            sql, table_name = _generate_view_code(
+            sql, table_name, extensions_required = _generate_view_code(
                 str(source), table_name=table_name, **kwargs
             )
+            self._load_extensions(extensions_required)
             self.con.execute(sql)
         else:
             if table_name is None:
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -6,7 +6,8 @@ import datetime
 import functools
 import itertools
 import operator
-from typing import Iterable, Literal, Mapping, Sequence
+from pathlib import Path
+from typing import Any, Iterable, Literal, Mapping, Sequence
 from typing import Tuple as _Tuple
 from typing import TypeVar
 from typing import Union as _Union
@@ -198,6 +199,7 @@ __all__ = (
     'pi',
     'random',
     'range_window',
+    'read',
     'row_number',
     'rows_with_max_lookback',
     'schema',
@@ -869,6 +871,36 @@ def row_number() -> ir.IntegerColumn:
         A column expression enumerating rows
     """
     return ops.RowNumber().to_expr()
+
+
+def read(path: str | Path, **kwargs: Any) -> ir.Table:
+    """Lazily load a data source located at `path`.
+
+    Parameters
+    ----------
+    path
+        A filesystem path or URL. Supports CSV, TSV, and Parquet files.
+    kwargs
+        DuckDB-specific keyword arguments for the file type.
+
+        * CSV/TSV: https://duckdb.org/docs/data/csv#parameters.
+        * Parquet: https://duckdb.org/docs/data/parquet
+
+    Returns
+    -------
+    ir.Table
+        Table expression representing a file
+
+    Examples
+    --------
+    >>> batting = ibis.read("ci/ibis-testing-data/batting.csv")
+    >>> diamonds = ibis.read("ci/ibis-testing-data/parquet/diamonds/diamonds.parquet")
+    >>> ft = ibis.read("parquet://ci/ibis-testing-data/parquet/functional_alltypes/*")
+    """
+    from ibis.config import _default_backend
+
+    con = _default_backend()
+    return con.register(str(path), **kwargs)
 
 
 e = ops.E().to_expr()
--- a/docs/api/expressions/top_level.md
+++ b/docs/api/expressions/top_level.md
@@ -31,6 +31,7 @@ These methods and objects are available directly in the `ibis` module.
 ::: ibis.to_sql
 ::: ibis.random
 ::: ibis.range_window
+::: ibis.read
 ::: ibis.row_number
 ::: ibis.schema
 ::: ibis.struct
--- a/ibis/backends/duckdb/tests/test_register.py
+++ b/ibis/backends/duckdb/tests/test_register.py
@@ -2,6 +2,7 @@ import contextlib
 import csv
 import gzip
 import os
+import re
 from pathlib import Path
 
 import pytest
@@ -32,9 +33,9 @@ def gzip_csv(data_directory, tmp_path):
 
 
 @pytest.mark.parametrize(
-    "fname, in_table_name, out_table_name",
+    ("fname", "in_table_name", "out_table_name"),
     [
-        param("diamonds.csv", None, "diamonds", id="default"),
+        param("diamonds.csv", None, None, id="default"),
         param("csv://diamonds.csv", "Diamonds", "Diamonds", id="csv_name"),
         param(
             "file://diamonds.csv",
@@ -55,14 +56,17 @@ def test_register_csv(
     data_directory, fname, in_table_name, out_table_name, ext, gzip_csv
 ):
     con = ibis.duckdb.connect()
-    if ext is not None:
+    if ext:
         fname = gzip_csv
     with pushd(data_directory):
-        con.register(fname, table_name=in_table_name)
+        table = con.register(fname, table_name=in_table_name)
 
-    assert out_table_name in con.list_tables()
+    if out_table_name is not None:
+        out_table_name += (os.extsep * bool(ext) + (ext or "")) * (
+            in_table_name is None
+        )
+        assert out_table_name in con.list_tables()
 
-    table = con.table(out_table_name)
     assert table.count().execute()
 
 
@@ -73,18 +77,17 @@ def test_register_with_dotted_name(data_directory, tmp_path):
     f.parent.mkdir()
     data = data_directory.joinpath("diamonds.csv").read_bytes()
     f.write_bytes(data)
-    con.register(str(f.absolute()))
-    table = con.table("diamonds")
+    table = con.register(str(f.absolute()))
     assert table.count().execute()
 
 
 @pytest.mark.parametrize(
-    "fname, in_table_name, out_table_name",
+    ("fname", "in_table_name", "out_table_name"),
     [
-        (
+        pytest.param(
             "parquet://functional_alltypes.parquet",
             None,
-            "functional_alltypes",
+            "functional_alltypes_parquet",
         ),
         ("functional_alltypes.parquet", "funk_all", "funk_all"),
         ("parquet://functional_alltypes.parq", "funk_all", "funk_all"),
@@ -103,11 +106,10 @@ def test_register_parquet(
 
     con = ibis.duckdb.connect()
     with pushd(tmp_path):
-        con.register(f"parquet://{fname.name}", table_name=in_table_name)
+        table = con.register(f"parquet://{fname.name}", table_name=in_table_name)
 
-    assert out_table_name in con.list_tables()
+    assert any(out_table_name in t for t in con.list_tables())
 
-    table = con.table(out_table_name)
     assert table.count().execute()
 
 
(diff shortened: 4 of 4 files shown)


jcrist (commented):
This seems like a nice win here.

It also removes a wart with the current top-level `connect` where `ibis.connect` can return either a connection _or_ a table dependent on the value of the object passed in (not the type). This makes type annotating `ibis.connect` impossible to do in a nice way.

Perhaps we should deprecate loading parquet/csv files in `ibis.connect` in favor of these functions?

jcmkk3:
I'm really happy to see these functions. I would use these commonly for how I work with Ibis, plus it makes it so much easier for folks to try out Ibis for the first time.

I can't tell from the duckdb documentation, but can these functions read from a URL? I often use [tidytuesday](https://github.com/rfordatascience/tidytuesday) datasets when I want to try something out and it is nice in pandas or dplyr to be able to read directly from the URL.

cpcloud:
@jcmkk3 Really appreciate the rapid feedback!

I've added support for URLs in the latest commit 🎉:

```
In [1]: import ibis; ibis.options.interactive = True; from ibis import _

In [2]: t = ibis.read_csv('https://raw.githubusercontent.com/ArthurCheib/analytical-politics-project/main/dat
   ...: a/tidy-data-fh-un.csv')

In [3]: t
Out[3]:
┏━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━┓
┃ country     ┃ year  ┃ CL    ┃ PR    ┃ Status ┃ Region_Code ┃ Region_Name ┃ is_ldc ┃
┡━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━┩
│ string      │ int32 │ int32 │ int32 │ string │ int32       │ string      │ int32  │
├─────────────┼───────┼───────┼───────┼────────┼─────────────┼─────────────┼────────┤
│ Afghanistan │  1995 │     7 │     7 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  1996 │     7 │     7 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  1997 │     7 │     7 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  1998 │     7 │     7 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  1999 │     7 │     7 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  2000 │     7 │     7 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  2001 │     7 │     7 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  2002 │     6 │     6 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  2003 │     6 │     6 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  2004 │     6 │     5 │ NF     │         142 │ Asia        │      1 │
│ Afghanistan │  2005 │     5 │     5 │ PF     │         142 │ Asia        │      1 │
│ …           │     … │     … │     … │ …      │           … │ …           │      … │
└─────────────┴───────┴───────┴───────┴────────┴─────────────┴─────────────┴────────┘


In [4]: t.Region_Name.value_counts()
Out[4]:
┏━━━━━━━━━━━━━┳━━━━━━━┓
┃ Region_Name ┃ count ┃
┡━━━━━━━━━━━━━╇━━━━━━━┩
│ string      │ int64 │
├─────────────┼───────┤
│ Asia        │  1218 │
│ Europe      │  1099 │
│ Africa      │  1388 │
│ Americas    │   910 │
│ Oceania     │   364 │
└─────────────┴───────┘
```

ogrisel:
Nice feature, especially with the https URL support.

Maybe you should make it explicit in the docstring that this will read the full data in a in-memory (duckdb) table?

If the file is large, it might be recommended instead to create an empty in-memory duckdb connection and subsequently register the parquet (or CSV) file as a foreign table:

https://ibis-project.org/docs/3.1.0/backends/DuckDB/#ibis.backends.duckdb.Backend.register--returns

cpcloud:
> Nice feature, especially with the https URL support.
> 
> Maybe you should make it explicit in the docstring that this will read the full data in a in-memory (duckdb) table?

Ah, that's the thing though: it doesn't. It creates a view referencing the file.

> 
> If the file is large, it might be recommended instead to create an empty in-memory duckdb connection and subsequently register the parquet (or CSV) file as a foreign table:
> 
> [ibis-project.org/docs/3.1.0/backends/DuckDB/#ibis.backends.duckdb.Backend.register--returns](https://ibis-project.org/docs/3.1.0/backends/DuckDB/#ibis.backends.duckdb.Backend.register--returns)

This is exactly what it does :)

ogrisel:
> This is exactly what it does :)

Nice! I think it might be worth making that explicit.

When the file is a remote https file, where does it go and when is it garbage collected?

cpcloud:
> > This is exactly what it does :)
> 
> Nice! I think it might be worth making that explicit because the name `read_parquet` made me think it was read in memory ahead of query time.
> 
> When the file is a remote https file, where does it go and when is it garbage collected?
> 
> I mean, I assume it does not do HTTP range requests each time your do a query, or does it?

It does do range requests every time, because it's a view so it's not executed until you ask for it.

The tradeoffs here optimize for the case of fast construction of expressions, which in turn optimize for more rapid feedback. The cost is in repeated computation.

Hopefully, that's still useful if not optimal for some use cases :)

I'd be happy to call these `ibis.csv`/`ibis.parquet` or something else to make it less misleading about its in-memory status.

I did a bit of rudimentary data engineering in [this gist](https://gist.github.com/cpcloud/4676e93f505c9fb9071f6be7d1fc14c9) to determine the viability of the API and how it feels and so far I'm pretty happy with it.

jcmkk3:
> I'd be happy to call these `ibis.csv`/`ibis.parquet` or something else to make it less misleading about its in-memory status.

I'm not sure if it is really the exact same behavior but polars uses [`scan_csv`](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.scan_csv.html) for a similar function.

Arrow has an [`open_csv`](https://arrow.apache.org/docs/python/generated/pyarrow.csv.open_csv.html#pyarrow.csv.open_csv) function, but again, I'm not sure if it is really analogous.

Both libraries also have an eager `read_csv` function.

Could `connect_csv` and `connect_parquet` be an option? They are a bit verbose, but would at least show up in autocomplete with the normal `connect`.

ogrisel:
Maybe it's ok to keep the `read_*` names if the docstrings are explicit enough.

cpcloud:
I've changed the name to `ibis.file`, which also allows users to avoid having to care about which function to call based on the file type they're accessing.

jcrist:
I personally think "read" is the proper verb to use here, and is probably the word users coming from pandas/dask/pyspark will look for when browsing the API. Dask is also lazy and doesn't load any data, but we call things `read_csv`/`read_parquet`/.... anyway and no one has complained so far.

> I've changed the name to `ibis.file`, which also allows users to avoid having to care about which function to call based on the file type they're accessing.

I find the idea behind a single read-this-file command good, but:

- I worry about any file-type specific keywords making their way into this method, leading to hard to document/confusing kwargs. Having separate `read_csv`/`read_parquet` methods help avoid this issue IMO. We could keep a single API method that dispatched to these as well for users that want to auto-detect file types.
- I don't love "file", since it's a noun and not a verb. What's being done with the file? I'd be happier with `ibis.read`.
- Directories of csvs/parquet files are common, and something that many tools can read (and users may expect). Limiting a method from `some_string -> Table` to only handle "file" things feels limiting, and passing in a directory/glob string to a method called `file` would be confusing IMO.
- We're likely to get writer methods (#4444) at some point. Having parity in naming between the source and sink methods would be nice.

So in short:
- I'm happy with having a single method for now, but I'd prefer it was called `ibis.read`
- We might want to keep around `read_parquet`/`read_csv` for users wanting to avoid the auto-dispatch, and for a place to document filetype specific kwargs if those ever appear (the `ibis.read` method could then just say it "forwards kwargs to the file specific methods).

cpcloud:
`read` seems fine too, at this point I don't really care about the name anymore as long as it's one of the few thrown out here. The functionality is really what matters :)

jreback (commented):
could consider

ibis.read.parquet and similar

to avoid api lock-in

ibis/expr/api.py
```diff
@@ -871,6 +873,33 @@ def row_number() -> ir.IntegerColumn:
...
+    ir.Table
+        Table expression representing a file
+    """
+    from ibis.config import _default_backend
+
+    con = _default_backend()
+    return con.register(str(path), **kwargs)
+
+
+read_csv = read_parquet = read
```
jcrist, line 900: I don't love this, since the file type is still auto-determined from the file name (or scheme). If I have a file with a nonstandard extension I can't call `ibis.read_csv("test.data")` and have it work.

I'd prefer either:

- Dropping these for now
- Making them true top-level functions that don't autodetect based on the path/scheme.

jcrist (commented):
Overall this looks pretty good to me, would love to have this functionality in ibis. Just one small comment, otherwise :+1:.

ibis/expr/api.py
cpcloud: I'll drop them for now.

cpcloud:
Confirmed that this works with globs!

### CSV Globs

```
❯ ls /data/nyc_taxi/yellow-subset
Permissions Size User  Date Modified Name
.rw-r--r--  352M cloud  5 Nov 08:58  yellow_tripdata_2021-10.csv
.rw-r--r--   66M cloud  5 Nov 08:58  yellow_tripdata_2021-11.csv.gz
.rw-r--r--   62M cloud  5 Nov 08:58  yellow_tripdata_2021-12.csv.gz
…/ibis on  read-file is 📦 v3.2.0 via 🐍 v3.10.8 via ❄️  impure (ibis310)
❯ ipython

In [1]: import ibis; ibis.options.interactive = True; from ibis import _

In [2]: t = ibis.read('csv:///data/nyc_taxi/yellow-subset/*')

In [3]: t
Out[3]:
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━┓
┃ VendorID ┃ tpep_pickup_datetime ┃ tpep_dropoff_datetime ┃ passenger_count ┃ trip_distance ┃ RatecodeID ┃ store_and_fwd_flag ┃ PULocationID ┃ DOLocationID ┃ payment_type ┃ … ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━┩
│ int32    │ timestamp            │ timestamp             │ float64         │ float64       │ float64    │ string             │ int32        │ int32        │ int32        │ … │
├──────────┼──────────────────────┼───────────────────────┼─────────────────┼───────────────┼────────────┼────────────────────┼──────────────┼──────────────┼──────────────┼───┤
│        1 │ 2021-11-01 00:34:15  │ 2021-11-01 00:57:21   │             1.0 │          7.00 │        1.0 │ Y                  │          137 │          264 │            2 │ … │
│        1 │ 2021-11-01 00:00:29  │ 2021-11-01 00:14:13   │             1.0 │          3.20 │        1.0 │ N                  │          233 │          232 │            1 │ … │
│        1 │ 2021-11-01 00:27:51  │ 2021-11-01 00:34:21   │             2.0 │          1.80 │        1.0 │ N                  │           79 │          170 │            1 │ … │
│        1 │ 2021-11-01 00:01:56  │ 2021-11-01 00:25:18   │             0.0 │          4.70 │        1.0 │ N                  │          230 │           42 │            2 │ … │
│        1 │ 2021-11-01 00:10:44  │ 2021-11-01 00:29:06   │             1.0 │          3.50 │        1.0 │ N                  │          144 │           50 │            1 │ … │
│        1 │ 2021-11-01 00:36:21  │ 2021-11-01 00:54:46   │             2.0 │          3.40 │        1.0 │ N                  │          230 │          263 │            2 │ … │
│        1 │ 2021-11-01 00:12:05  │ 2021-11-01 00:31:06   │             0.0 │         14.30 │        1.0 │ N                  │          138 │          257 │            1 │ … │
│        1 │ 2021-11-01 00:49:43  │ 2021-11-01 00:59:38   │             1.0 │          1.40 │        1.0 │ N                  │          249 │          107 │            2 │ … │
│        2 │ 2021-11-01 00:10:05  │ 2021-11-01 00:37:59   │             1.0 │          6.29 │        1.0 │ N                  │          249 │           37 │            2 │ … │
│        2 │ 2021-11-01 00:26:13  │ 2021-11-01 00:37:37   │             1.0 │          2.15 │        1.0 │ N                  │           50 │           90 │            1 │ … │
│        2 │ 2021-11-01 00:41:44  │ 2021-11-01 00:46:21   │             1.0 │          0.73 │        1.0 │ N                  │           90 │          234 │            1 │ … │
│        … │ …                    │ …                     │               … │             … │          … │ …                  │            … │            … │            … │ … │
└──────────┴──────────────────────┴───────────────────────┴─────────────────┴───────────────┴────────────┴────────────────────┴──────────────┴──────────────┴──────────────┴───┘


In [4]: t.passenger_count.value_counts()
Out[4]:
┏━━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ passenger_count ┃ count   ┃
┡━━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│ float64         │ int64   │
├─────────────────┼─────────┤
│             1.0 │ 7065734 │
│             2.0 │ 1545144 │
│             0.0 │  239513 │
│             3.0 │  407262 │
│             4.0 │  171978 │
│             5.0 │  220194 │
│             6.0 │  134720 │
│             7.0 │      29 │
│             9.0 │      14 │
│             8.0 │      11 │
│             nan │  366223 │
│               … │       … │
└─────────────────┴─────────┘
```

### Parquet Globs

```
In [5]: t = ibis.read('/data/ibis-testing/*.parquet')

In [6]: t
Out[6]:
┏━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━┓
┃ id    ┃ bool_col ┃ tinyint_col ┃ smallint_col ┃ int_col ┃ bigint_col ┃ float_col ┃ double_col ┃ date_string_col      ┃ string_col           ┃ timestamp_col           ┃ year  ┃ month ┃
┡━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━┩
│ int32 │ boolean  │ int32       │ int32        │ int32   │ int64      │ float32   │ float64    │ binary               │ binary               │ timestamp               │ int32 │ int32 │
├───────┼──────────┼─────────────┼──────────────┼─────────┼────────────┼───────────┼────────────┼──────────────────────┼──────────────────────┼─────────────────────────┼───────┼───────┤
│  5160 │ True     │           0 │            0 │       0 │          0 │       0.0 │        0.0 │ b'06/01/10'          │ b'0'                 │ 2010-06-01 00:00:00.000 │  2010 │     6 │
│  5161 │ False    │           1 │            1 │       1 │         10 │       1.1 │       10.1 │ b'06/01/10'          │ b'1'                 │ 2010-06-01 00:01:00.000 │  2010 │     6 │
│  5162 │ True     │           2 │            2 │       2 │         20 │       2.2 │       20.2 │ b'06/01/10'          │ b'2'                 │ 2010-06-01 00:02:00.100 │  2010 │     6 │
│  5163 │ False    │           3 │            3 │       3 │         30 │       3.3 │       30.3 │ b'06/01/10'          │ b'3'                 │ 2010-06-01 00:03:00.300 │  2010 │     6 │
│  5164 │ True     │           4 │            4 │       4 │         40 │       4.4 │       40.4 │ b'06/01/10'          │ b'4'                 │ 2010-06-01 00:04:00.600 │  2010 │     6 │
│  5165 │ False    │           5 │            5 │       5 │         50 │       5.5 │       50.5 │ b'06/01/10'          │ b'5'                 │ 2010-06-01 00:05:00.100 │  2010 │     6 │
│  5166 │ True     │           6 │            6 │       6 │         60 │       6.6 │       60.6 │ b'06/01/10'          │ b'6'                 │ 2010-06-01 00:06:00.150 │  2010 │     6 │
│  5167 │ False    │           7 │            7 │       7 │         70 │       7.7 │       70.7 │ b'06/01/10'          │ b'7'                 │ 2010-06-01 00:07:00.210 │  2010 │     6 │
│  5168 │ True     │           8 │            8 │       8 │         80 │       8.8 │       80.8 │ b'06/01/10'          │ b'8'                 │ 2010-06-01 00:08:00.280 │  2010 │     6 │
│  5169 │ False    │           9 │            9 │       9 │         90 │       9.9 │       90.9 │ b'06/01/10'          │ b'9'                 │ 2010-06-01 00:09:00.360 │  2010 │     6 │
│  5170 │ True     │           0 │            0 │       0 │          0 │       0.0 │        0.0 │ b'06/02/10'          │ b'0'                 │ 2010-06-02 00:10:00.450 │  2010 │     6 │
│     … │ …        │           … │            … │       … │          … │         … │          … │ …                    │ …                    │ …                       │     … │     … │
└───────┴──────────┴─────────────┴──────────────┴─────────┴────────────┴───────────┴────────────┴──────────────────────┴──────────────────────┴─────────────────────────┴───────┴───────┘
```

ogrisel:
This is great, however the error message could be improved to make explicit that a specific file format is not support, for instance with an excel file:

```python
>>> ibis.read("df.xslx")
  Input In [26] in <cell line: 1>
    ibis.read("df.xslx")
  File ~/code/ibis/ibis/expr/api.py:897 in read
    return con.register(str(path), **kwargs)
  File ~/code/ibis/ibis/backends/duckdb/__init__.py:264 in register
    sql, table_name, extensions_required = _generate_view_code(
  File ~/code/ibis/ibis/common/dispatch.py:103 in __call__
    return func(s, *args, **kwargs, **match.groupdict())
  File ~/code/ibis/ibis/backends/duckdb/__init__.py:105 in _file_or_url
    return _generate_view_code(f"{extension}://{path}", table_name=table_name, **kwargs)
  File ~/code/ibis/ibis/common/dispatch.py:103 in __call__
    return func(s, *args, **kwargs, **match.groupdict())
TypeError: ibis.backends.duckdb._file_or_url() got multiple values for keyword argument 'scheme'
```

or with mixed supported file formats matching in the same glob:

```python
>>> ls
df.csv      df.parquet
>>> ibis.read("df.*")
  Input In [32] in <cell line: 1>
    ibis.read("df.*")
  File ~/code/ibis/ibis/expr/api.py:897 in read
    return con.register(str(path), **kwargs)
  File ~/code/ibis/ibis/backends/duckdb/__init__.py:264 in register
    sql, table_name, extensions_required = _generate_view_code(
  File ~/code/ibis/ibis/common/dispatch.py:103 in __call__
    return func(s, *args, **kwargs, **match.groupdict())
  File ~/code/ibis/ibis/backends/duckdb/__init__.py:105 in _file_or_url
    return _generate_view_code(f"{extension}://{path}", table_name=table_name, **kwargs)
  File ~/code/ibis/ibis/common/dispatch.py:103 in __call__
    return func(s, *args, **kwargs, **match.groupdict())
TypeError: ibis.backends.duckdb._file_or_url() got multiple values for keyword argument 'scheme'
```

cpcloud:
@ogrisel Ah, thanks for the feedback, great points. I'll see what I can do about making the error message less cryptic!

cpcloud:
Alright, this now should have reasonable error messages for things that aren't supported.

ibis/backends/duckdb/__init__.py
```diff
@@ -99,18 +117,16 @@ def _s3(full_path, table_name=None):
...
 
 
 @_generate_view_code.register(r".+", priority=1)
-def _default(_, **kwargs):
+def _default(path, **kwargs):
     raise ValueError(
-        """
-Unrecognized filetype or extension.
-Valid prefixes are parquet://, csv://, s3://, or file://
+        """Unrecognized file type or extension: {path}.
```
ogrisel, line 126: The `path` variable is not interpolated at the moment.


```suggestion
        f"""Unrecognized file type or extension: {path}.
```

ogrisel (commented):
There are still some serious problems with the input validation logic:

```python
>>> import ibis
>>> ls
df.csv      df.parquet  df.xlsx
>>> ibis.read("csv://./*")
[1]    85705 segmentation fault  ipython
```

and:

```python
>>> import ibis
>>> ibis.read("parquet://./*")
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1900 in _execute_context
    self.dialect.do_execute(
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/default.py:736 in do_execute
    cursor.execute(statement, parameters)
  File ~/code/duckdb_engine/duckdb_engine/__init__.py:120 in execute
    self.c.execute(statement, parameters)
InvalidInputException: Invalid Input Error: No magic bytes found at end of file '/private/tmp/tmpdata/df.xlsx'

The above exception was the direct cause of the following exception:
  Input In [3] in <cell line: 1>
    ibis.read("parquet://./*")
  File ~/code/ibis/ibis/expr/api.py:897 in read
    return con.register(str(path), **kwargs)
  File ~/code/ibis/ibis/backends/duckdb/__init__.py:256 in register
    self.con.execute(sql)
  File <string>:2 in execute
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/util/deprecations.py:402 in warned
    return fn(*args, **kwargs)
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:3257 in execute
    return connection.execute(statement, *multiparams, **params)
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1365 in execute
    return self._exec_driver_sql(
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1669 in _exec_driver_sql
    ret = self._execute_context(
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1943 in _execute_context
    self._handle_dbapi_exception(
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:2124 in _handle_dbapi_exception
    util.raise_(
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/util/compat.py:208 in raise_
    raise exception
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1900 in _execute_context
    self.dialect.do_execute(
  File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/default.py:736 in do_execute
    cursor.execute(statement, parameters)
  File ~/code/duckdb_engine/duckdb_engine/__init__.py:120 in execute
    self.c.execute(statement, parameters)
ProgrammingError: (duckdb.InvalidInputException) Invalid Input Error: No magic bytes found at end of file '/private/tmp/tmpdata/df.xlsx'
[SQL: CREATE OR REPLACE VIEW "/private/tmp/tmpdata/*" AS
SELECT * FROM read_parquet('/private/tmp/tmpdata/*')]
(Background on this error at: https://sqlalche.me/e/14/f405)
```

the segfault is probably revealing a lower level bug in the duckdb Python API which should probably be reported upstream with minimal reproducer.

cpcloud:
> There are still some serious problems with the input validation logic:
> 
> ```python
> >>> import ibis
> >>> ls
> df.csv      df.parquet  df.xlsx
> >>> ibis.read("csv://./*")
> [1]    85705 segmentation fault  ipython
> ```
> 
> and:
> 
> ```python
> >>> import ibis
> >>> ibis.read("parquet://./*")
> Traceback (most recent call last):
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1900 in _execute_context
>     self.dialect.do_execute(
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/default.py:736 in do_execute
>     cursor.execute(statement, parameters)
>   File ~/code/duckdb_engine/duckdb_engine/__init__.py:120 in execute
>     self.c.execute(statement, parameters)
> InvalidInputException: Invalid Input Error: No magic bytes found at end of file '/private/tmp/tmpdata/df.xlsx'
> 
> The above exception was the direct cause of the following exception:
> Traceback (most recent call last):
>   Input In [3] in <cell line: 1>
>     ibis.read("parquet://./*")
>   File ~/code/ibis/ibis/expr/api.py:897 in read
>     return con.register(str(path), **kwargs)
>   File ~/code/ibis/ibis/backends/duckdb/__init__.py:256 in register
>     self.con.execute(sql)
>   File <string>:2 in execute
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/util/deprecations.py:402 in warned
>     return fn(*args, **kwargs)
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:3257 in execute
>     return connection.execute(statement, *multiparams, **params)
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1365 in execute
>     return self._exec_driver_sql(
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1669 in _exec_driver_sql
>     ret = self._execute_context(
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1943 in _execute_context
>     self._handle_dbapi_exception(
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:2124 in _handle_dbapi_exception
>     util.raise_(
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/util/compat.py:208 in raise_
>     raise exception
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py:1900 in _execute_context
>     self.dialect.do_execute(
>   File ~/mambaforge/envs/dev/lib/python3.10/site-packages/sqlalchemy/engine/default.py:736 in do_execute
>     cursor.execute(statement, parameters)
>   File ~/code/duckdb_engine/duckdb_engine/__init__.py:120 in execute
>     self.c.execute(statement, parameters)
> ProgrammingError: (duckdb.InvalidInputException) Invalid Input Error: No magic bytes found at end of file '/private/tmp/tmpdata/df.xlsx'
> [SQL: CREATE OR REPLACE VIEW "/private/tmp/tmpdata/*" AS
> SELECT * FROM read_parquet('/private/tmp/tmpdata/*')]
> (Background on this error at: https://sqlalche.me/e/14/f405)
> ```
> 
> the segfault is probably revealing a lower level bug in the duckdb Python API which should probably be reported upstream with minimal reproducer.

Oh boy 😬

cpcloud:
@ogrisel Can you show the contents of those `df.*` files? I can't reproduce the segfault on my machine using a random collection of mixed files.

ogrisel:
This is the same pandas dataframe exported 3 times:

```python
>>> pd.read_excel("df.xlsx")
    Unnamed: 0         a
0            0 -1.937078
1            1  0.528488
2            2  0.605134
3            3 -0.874927
4            4 -0.446203
..         ...       ...
95          95  0.207459
96          96 -2.141455
97          97  0.241468
98          98  0.411752
99          99 -0.940849

[100 rows x 2 columns]
```

I think they were generated with something like the following:

```python
>>> import pandas as pd
>>> import numpy as np
>>> df = pd.DataFrame({"a": np.random.randn(100)})
>>> df.to_excel("df.xlsx")
>>> df.to_parquet("df.parquet")
>>> df.to_csv("df.csv")
```

ogrisel:
Not that the segfault can also happen with loading the parquet file with the CSV reader:

```python
In [4]: ibis.read("csv://./*.parquet").execute()
[1]    87118 segmentation fault  ipython
```

cpcloud:
> Not that the segfault can also happen with loading the parquet file with the CSV reader:
> 
> ```python
> In [4]: ibis.read("csv://./*.parquet").execute()
> [1]    87118 segmentation fault  ipython
> ```

I think it _may_ be something specific to the way the file is being written, because I can't reproduce this when using files generated by duckdb:

```
In [4]: import ibis; ibis.options.interactive = True; from ibis import _

In [5]: t = ibis.read("csv:///data/ml-parquet/*.parquet")
ProgrammingError: (duckdb.InvalidInputException) Invalid Input Error: Error on line 1025: expected 5 values per row, but got 1. (DELIMITER='    ' (auto detected), QUOTE='"' (auto detected), ESCAPE='' (auto detected), HEADER=0 (auto detected), SAMPLE_SIZE=10240, IGNORE_ERRORS=0, ALL_VARCHAR=0)
[SQL: CREATE OR REPLACE VIEW "/data/ml-parquet/*_parquet" AS
SELECT * FROM read_csv('/data/ml-parquet/*.parquet', auto_detect=True)]
(Background on this error at: https://sqlalche.me/e/14/f405)
```

cpcloud:
This is reproducible outside of ibis entirely, so I don't think it should block this PR.

cpcloud:
On duckdb latest `master` I cannot reproduce the segfault.

ogrisel:
The code LGTM then. I think this should be used in examples and tutorials where appropriate. Maybe the example databases could be published as `.parquet` files (in additiona to sqlite db files). But this can be done in follow-up PRs.

gforsyth (approved):
This looks good to me, ~pending a rebase.~ scratch that, I think this is good to go in now.

gforsyth:
Thanks @cpcloud !

cpcloud:
Heck yeah!

PR #4762, opened 2022-11-03
feat: add `mode` aggregation

jcrist:
This adds a scalar `mode` aggregation. Here the `mode` is defined as the most common value in a column, excluding `NULL` values. If there are multiple values with the same count, it's backend defined which value is returned.

It is currently implemented for `duckdb`, `postgres`, `snowflake`, `pandas`, `dask`, and `polars`.


cpcloud (approved):
Sweet. This PR is an excellent example of "going wide" with an operations: taking an operation and implementing it for as many backends as possible.

PR #4765, opened 2022-11-03
feat(api): add `agg` alias for `aggregate`

cpcloud:
Add an `agg` alias for `aggregate`. I am really tired of typing out the entire word.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -400,6 +400,8 @@ class Table(Expr, JupyterMixin):
 
         return agg.to_expr()
 
+    agg = aggregate
+
     def distinct(self) -> Table:
         """Compute the set of unique rows in the table."""
         return ops.Distinct(self).to_expr()


jcrist (approved):
:+1:

jcmkk3:
Looks like you already merged this and I understand that `agg` has precedence in the pydata ecosystem, but I think that my favorite verb for this operation is `rollup` as used by [arquero](https://uwdata.github.io/arquero/api/verbs). I can no longer find the reference, but I'm sure that I remember [Jeffrey Heer](https://homes.cs.washington.edu/~jheer/) saying that the verbs came out of academic user studies done for data wrangler (which became [Trifacta](https://www.trifacta.com/)).

Beyond `rollup` evoking a better visual image of how the operation changes the data shape, it also happens to have a really nice visual alignment with some of the other common verbs.
```python
t.rollup
t.select
t.mutate
t.filter
```

It is a slippery slope boxing yourself into trying to keep names the same length, of course.

Another shortened verb that has some precedence is `sel` for `select` as used in [xarray](https://docs.xarray.dev/en/stable/user-guide/indexing.html#quick-overview).

I mostly just geek out about this stuff. Not proposing that you change your mind on any of these names.

ogrisel:
Personally I prefer explicit full English words in APIs.

Also, I tend to dislike aliases because then it breaks the community in term of programming style and the extra alternatives are confusing for newcomers.

cpcloud:
> Personally I prefer explicit full English words in APIs.

I sympathize with this for sure: when writing pipelines being explicit made things easier to understand. There's probably an analogy (and a similar tension with iteration speed and convenience) with CI and production shell script arguments here.

> 
> I also like the symmetry between `aggregate` and `mutate`.

There's always the possibility of `mut` I guess :)

> 
> Also, I tend to dislike aliases because then it breaks the community in term of programming style and the extra alternatives are confusing for newcomers.

Yeah, I understand this too. The way we've currently been thinking about this is to pick a primary API and remove all references to the secondary API in documentation and tests but not actually remove the secondary API to avoid breaking existing code that uses it. Optionally we can add a deprecation warning if we actually do intend to remove it at some point.

ogrisel:
> There's always the possibility of mut I guess :)

Please no!

PR #4770, opened 2022-11-04
feat: support `Table.fillna` for SQL backends

jcrist:
This also fixes an incompatibility bug when passing in incompatibly typed scalars to `Table.fillna` (e.g. `table.fillna(0)` in a table containing string columns). Previously the output was backend defined - pandas, for example, would happily return a column mixing strings and integers.

We now allow casting of the fill value, but only between different numeric types. All other fill values must be of the proper column type, otherwise a nice error message is raised explaining the issue.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -22,6 +22,7 @@ from rich.jupyter import JupyterMixin
 
 import ibis
 import ibis.common.exceptions as com
+import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 from ibis import util
 from ibis.expr.deferred import Deferred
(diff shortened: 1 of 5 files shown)


jcrist:
TPC-H failure is due to the code-cov report failing to push :shrug:. This should be ready for review.

ibis/expr/types/relations.py
```diff
@@ -908,31 +909,49 @@ def fillna(
...
+        >>> res = t.fillna({c: 0.0 for c, t in t.schema().items() if t == dt.float64})
 
         Returns
         -------
         Table
             Table expression
-        """  # noqa: E501
+        """
+
+        def compatible(val_type, col_type):
```
cpcloud, line 921: Should we use `rlz.castable` instead of defining this?

cpcloud (commented):
Tests look good, but I think we may be able to use `rlz.castable` to reduce the amount of special case code here.

ibis/expr/types/relations.py
cpcloud: Actually, `dt.castable` is what you want here.

jcrist: Thanks, fixed.

cpcloud (approved):
LGTM, 🚢.

PR #4771, opened 2022-11-04
fix: exclude geospatial ops from pandas/dask/polars `has_operation`

### LINKED ISSUE #4699: docs: operation support matrix incorrectly states pandas/dask support geospatial operations
The [operation support matrix](https://ibis-project.org/docs/3.2.0/backends/support_matrix/) currently incorrectly states that pandas/dask support geospatial operations.

This is because:

- The `gen_matrix` [script](https://github.com/ibis-project/ibis/blob/master/gen_matrix.py) uses `backend.has_operation` to check if an operation is supported for a backend.
- The `pandas`/`dask` backends implement this method by checking if the operation type (or a parent class of that type) are registered by the `execute_node`/`pre_execute` dispatch functions.
- The current implementation of the pandas/dask backends register a single handler for base classes like `ops.Binary` rather than individual handlers for subclasses like `ops.Add`
- The geospatial operations also subclass from these base classes, and are thus mistakenly identified as supported

The "proper" fix for this is to probably rework the pandas `execute_node` dispatch registry to avoid registering handlers for base classes like `Binary`, instead registering handlers for each individual binary op.

A more pragmatic fix would be to add a hardcoded exception to either `has_operation` or the `gen_matrix` script to mark the geospatial operations as unsupported.

jcrist:
Fixes #4699.

### DIFF
--- a/ibis/backends/dask/tests/test_client.py
+++ b/ibis/backends/dask/tests/test_client.py
@@ -1,16 +1,14 @@
 import re
 
+import dask.dataframe as dd
 import numpy as np
 import pandas as pd
 import pytest
+from dask.dataframe.utils import tm
 from pytest import param
 
 import ibis
-
-dd = pytest.importorskip("dask.dataframe")
-from dask.dataframe.utils import tm  # noqa: E402
-
-from ibis.backends.dask.client import DaskTable  # noqa: E402
+from ibis.backends.dask.client import DaskTable
 
 
 def make_dask_data_frame(npartitions):
--- a/ibis/backends/pandas/__init__.py
+++ b/ibis/backends/pandas/__init__.py
@@ -180,6 +180,11 @@ class BasePandasBackend(BaseBackend):
 
     @classmethod
     def has_operation(cls, operation: type[ops.Value]) -> bool:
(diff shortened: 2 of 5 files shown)


ibis/backends/dask/tests/test_client.py
```diff
@@ -136,3 +135,9 @@ def test_invalid_connection_parameter_types(npartitions):
     )
     with pytest.raises(TypeError, match=expeced_msg):
         ibis.dask.from_dataframe("file.csv")
+
+
+def test_has_operation(client):
```
cpcloud, line 140: Any reason not to make this a backend test? Seems simple enough to `never` the ones that do support it.

jcrist: Bug felt specific enough to these three backends. Happy to move the test if you think that'd be better.

cpcloud: Let's put this into a single backend test.

cpcloud:
Friendly ping @jcrist!

ibis/backends/dask/tests/test_client.py
jcrist: Done

cpcloud (approved):
Thanks!

PR #4777, opened 2022-11-05
ci: remove `datafusion` from conda-lock file gen for windows

cpcloud:
This PR removes datafusion from conda lockfile generation. We can reenable it if the package ever starts working again.


cpcloud:
Merging, will kick off the job after that to see if it works.

PR #4779, opened 2022-11-05
feat(api): add `distinct` arg to `ArrayCollect`

cpcloud:
This PR adds a `distinct` boolean argument to `ArrayCollect` and also makes it a subclass of `Filterable` allowing the `where` argument.

### DIFF
--- a/ibis/backends/pandas/execution/arrays.py
+++ b/ibis/backends/pandas/execution/arrays.py
@@ -110,6 +110,21 @@ def execute_array_repeat_scalar(op, data, n, **kwargs):
     return np.tile(data, max(n, 0))
 
 
-@execute_node.register(ops.ArrayCollect, (pd.Series, SeriesGroupBy))
-def execute_array_collect(op, data, aggcontext=None, **kwargs):
-    return aggcontext.agg(data, np.array)
+@execute_node.register(ops.ArrayCollect, pd.Series, bool, (type(None), pd.Series))
+def execute_array_collect(op, data, distinct, where, aggcontext=None, **kwargs):
+    return aggcontext.agg(
+        data.loc[where] if where is not None else data,
+        pd.unique if distinct else np.array,
+    )
+
+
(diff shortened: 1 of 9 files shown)


cpcloud:
I'm not 100% sure whether I like the `distinct=True` API, since it looks like `DISTINCT` is a property of the column, but I can't think of another way to support distinct inputs to aggregates without reintroducing `col.distinct()`.

ibis/backends/pandas/execution/arrays.py
```diff
@@ -110,6 +110,21 @@ def execute_array_repeat_scalar(op, data, n, **kwargs):
...
 
 
-@execute_node.register(ops.ArrayCollect, (pd.Series, SeriesGroupBy))
-def execute_array_collect(op, data, aggcontext=None, **kwargs):
-    return aggcontext.agg(data, np.array)
+@execute_node.register(ops.ArrayCollect, pd.Series, bool, (type(None), pd.Series))
+def execute_array_collect(op, data, distinct, where, aggcontext=None, **kwargs):
+    return aggcontext.agg(
+        data.loc[where] if where is not None else data,
+        np.unique if distinct else np.array,
```
jreback, line 117: use pd.unique as np.unique sorts

cpcloud: Technically the order is undefined, but happy to use `pd.unique`.

jreback: sure - pd.unique preserves whatever it is

cpcloud: Fixed! Appreciate the review!

cpcloud:
I'll split this into two PRs so that we can discuss the `distinct` API separate from the `Filterable` implementation, which I believe is not controversial in any way.

cpcloud:
Closing this out for now, don't really feel like tackling this yet and there's no pressing need.

PR #4786, opened 2022-11-06
perf(ux): remove pandas import overhead from `import ibis`

cpcloud:
This PR moves pandas imports around to speed up `import ibis` by about 2x. In my local tests I see import times on `master` at about 500 ms, and after this PR I see import times at around 250 ms.

### DIFF
--- a/ibis/__init__.py
+++ b/ibis/__init__.py
@@ -1,16 +1,12 @@
 """Initialize Ibis module."""
 from __future__ import annotations
 
-# Converting an Ibis schema to a pandas DataFrame requires registering
-# some type conversions that are currently registered in the pandas backend
-import ibis.backends.pandas
-import ibis.config
-import ibis.expr.types as ir
 from ibis import util
 from ibis.backends.base import BaseBackend
 from ibis.common.exceptions import IbisError
 from ibis.config import options
 from ibis.expr import api
+from ibis.expr import types as ir
 from ibis.expr.api import *  # noqa: F401,F403
 
 __all__ = ['api', 'ir', 'util', 'BaseBackend', 'IbisError', 'options']
@@ -23,6 +19,7 @@ _KNOWN_BACKENDS = ['bigquery', 'heavyai']
 
 def __dir__() -> list[str]:
     """Adds tab completion for ibis backends to the top-level module."""
+
     out = set(__all__)
     out.update(ep.name for ep in util.backend_entry_points())
     return sorted(out)
@@ -67,6 +64,8 @@ def __getattr__(name: str) -> BaseBackend:
     # The first time a backend is loaded, we register its options, and we set
     # it as an attribute of `ibis`, so `__getattr__` is not called again for it
     backend.register_options()
+
+    import ibis
 
     setattr(ibis, name, backend)
     return backend
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -15,10 +15,11 @@ import ibis.expr.types as ir
 import ibis.util as util
 from ibis.backends.base import BaseBackend
 from ibis.backends.base.sql.compiler import Compiler
-from ibis.expr.typing import TimeContext
 
 if TYPE_CHECKING:
     import pyarrow as pa
+
+    from ibis.expr.typing import TimeContext
 
 __all__ = [
     'BaseSQLBackend',
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -3,9 +3,8 @@ from __future__ import annotations
 import contextlib
 import getpass
 from operator import methodcaller
-from typing import Any, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
-import pandas as pd
 import sqlalchemy as sa
 
 import ibis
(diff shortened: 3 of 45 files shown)


ibis/__init__.py
```diff
@@ -43,6 +39,11 @@ def __getattr__(name: str) -> BaseBackend:
     the `ibis.backends` entrypoints. If successful, the `ibis.sqlite`
     attribute is "cached", so this function is only called the first time.
     """
+    # Converting an Ibis schema to a pandas DataFrame requires registering some
+    # type conversions that are currently registered in the pandas backend
+    import ibis.backends.pandas
+    from ibis import util
```
ogrisel, line 45: Based on the change in the `__all__` variable, I think that `ibis.backends.pandas` wasn't public but `ir` was.

cpcloud: Yep, `ir` is not something we want to be public long term so I removed it here.

cpcloud: I think I can break that out into a separate PR so that it'll get listed as a breaking change.

ibis/__init__.py
```diff
@@ -23,6 +17,8 @@
 
 def __dir__() -> list[str]:
     """Adds tab completion for ibis backends to the top-level module."""
+    from ibis import util
```
jcrist, line 20: What's in `ibis.util` that needs to be delayed loading? AFAICT pandas is only imported there if type checking?

jcrist (commented):
I only gave this a cursory glance, but overall this makes sense to me.

What's the intended level of pandas-less support here?

- Remove pandas as a mandatory dependency (not yet, but at some point?)
- No need to import pandas for users that never hit it?
- Just delay the pandas import until later (this seems less useful to me)?

We might want to add a test for this to ensure import behavior doesn't change. In other projects I've done this by executing a python script in a subprocess that does something and then checks the contents of `sys.modules` for any offending modules.

If the intent is just no pandas on import, then something like:

```python
script = """
import ibis
import sys
assert "pandas" not in sys.modules
"""

# execute `script` in subprocess as part of a test
```

may be sufficient.

If we also want to ensure that pandas isn't mandatory for a connect -> query -> `to_pyarrow` call then we might want to hit that workflow as well as a generic backend test, `notimpl`-ing it for any pandas specific backends.

cpcloud:
> I only gave this a cursory glance, but overall this makes sense to me.
> 
> What's the intended level of pandas-less support here?
> 
> * Remove pandas as a mandatory dependency (not yet, but at some point?)
> * No need to import pandas for users that never hit it?
> * Just delay the pandas import until later (this seems less useful to me)?
> 
> We might want to add a test for this to ensure import behavior doesn't change. In other projects I've done this by executing a python script in a subprocess that does something and then checks the contents of `sys.modules` for any offending modules.
> 
> If the intent is just no pandas on import, then something like:
> 
> ```python
> script = """
> import ibis
> import sys
> assert "pandas" not in sys.modules
> """
> 
> # execute `script` in subprocess as part of a test
> ```
> 
> may be sufficient.
> 
> If we also want to ensure that pandas isn't mandatory for a connect -> query -> `to_pyarrow` call then we might want to hit that workflow as well as a generic backend test, `notimpl`-ing it for any pandas specific backends.

Ultimately making pandas an optional dependency is ideal. This PR is the first step on the way to that.

ogrisel:
> Ultimately making pandas an optional dependency is ideal.

What type of data would be returned by `execute` in this case? A `pyarrow.Table` or something else? Or something configurable as a module level setting?

cpcloud:
> > Ultimately making pandas an optional dependency is ideal.
> 
> I am curious: what type of data would be returned by `execute` in this case? A `pyarrow.Table` or something else? Or something configurable as a module level setting?
> 
> Sorry for side-tracking the original PR review.

No worries, happy to discuss. Does it makes sense to move this to a proper github discussion so that we can really dig into the question?

ogrisel:
> Does it makes sense to move this to a proper github discussion so that we can really dig into the question?

Done in https://github.com/ibis-project/ibis/discussions/4819

ibis/__init__.py
cpcloud: Nothing, this can be reverted.

cpcloud: Done.

ibis/__init__.py
```diff
@@ -43,6 +40,10 @@ def __getattr__(name: str) -> BaseBackend:
     the `ibis.backends` entrypoints. If successful, the `ibis.sqlite`
     attribute is "cached", so this function is only called the first time.
     """
+    # Converting an Ibis schema to a pandas DataFrame requires registering some
+    # type conversions that are currently registered in the pandas backend
+    import ibis.backends.pandas
```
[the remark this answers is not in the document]
cpcloud: Oh, great, then I can delete this line!

jcrist:
@cpcloud can you also add a test to prevent accidentally importing pandas again later (as requested in https://github.com/ibis-project/ibis/pull/4786#pullrequestreview-1170751977)?

cpcloud:
> @cpcloud can you also add a test to prevent accidentally importing pandas again later (as requested in [#4786 (review)](https://github.com/ibis-project/ibis/pull/4786#pullrequestreview-1170751977))?

Yes!

cpcloud:
@jcrist Done, test added.

ibis/__init__.py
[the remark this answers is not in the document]
cpcloud: Deleted.

jcrist (approved):
Assuming tests pass, this LGTM.

PR #4790, opened 2022-11-07
build(docs): fix docs build cycle

### LINKED ISSUE #4761: docs: infinite build when using `mkdocs serve`
It appears that when using `mkdocs serve` the docs are repeatedly rebuilt to no end.

I suspect there's a file that we're generating (maybe the operation matrix?) that is being considered new and triggering a rebuild.

cpcloud:
This PR fixes a bug when using `mkdocs serve` that causes the build to keep building even if nothing has apparently changed. Closes #4761.

### DIFF
--- a/gen_matrix.py
+++ b/gen_matrix.py
@@ -69,7 +69,15 @@ def main():
         "backends",
         "support_matrix.csv",
     )
-    table.to_csv(dst, index_label="Backends")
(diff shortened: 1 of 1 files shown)


cpcloud:
Merging, the only thing remaining is the backend build requirement.

PR #4793, opened 2022-11-07
fix: standardize list_tables signature everywhere

### LINKED ISSUE #2877: refactor: standardize list_tables signature
All backends seem to provide `list_tables`, but with a variety of signatures.

I propose to standardize all them to the same, so things are simpler and clearer, and code can be reused from one backend to another easier. My preferred signature would be:

```python
def list_tables(self, like: str = None) -> List[str]:
```
We can add a `FutureWarning` for Ibis 2.0, if other parameters are used. I'll add `**kwargs`, and if not empty, a warning will be shown, and in 3.0 we can remove 3.0. I'll also deprecate `list_tables` in `Database`, in favor of the `Client` one.

Let me know if anyone disagrees, or any comment.

The list of current signatures is next:
```
ibis/backends/clickhouse/client.py:    def list_tables(self, like=None, database=None):
ibis/backends/mysql/client.py:    def list_tables(self, like=None, database=None, schema=None):
ibis/backends/csv/__init__.py:    def list_tables(self, path=None):
ibis/backends/parquet/__init__.py:    def list_tables(self, path=None):
ibis/backends/dask/client.py:    def list_tables(self, like: str = None) -> List[str]:
ibis/backends/hdf5/__init__.py:    def list_tables(self, path=None):
ibis/backends/spark/client.py:    def list_tables(self, like=None, database=None):
ibis/backends/postgres/client.py:    def list_tables(self, like=None, database=None, schema=None):
ibis/backends/impala/kudu_support.py:    def list_tables(self, filter=''):
ibis/backends/impala/client.py:    def list_tables(self, like=None, database=None):
ibis/backends/sqlite/client.py:    def list_tables(self, like=None, database=None, schema=None):
ibis/backends/base/file/__init__.py:    def list_tables(self, path=None):
ibis/backends/base/file/__init__.py:    def list_tables(self, path=None):
ibis/backends/base/sql/alchemy/client.py:    def list_tables(
ibis/backends/base/sql/alchemy/database.py:    def list_tables(self, like=None):
ibis/backends/base/sql/alchemy/database.py:    def list_tables(self, like=None, schema=None):
ibis/backends/base/client.py:    def list_tables(self, like: str = None) -> list:
ibis/backends/pandas/client.py:    def list_tables(self, like=None):
```

gforsyth:
Fixes #2877 -- most of the differing signatures were removed in earlier refactors -- this is the only one left (except for a mock fixture in the test suite)

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -147,15 +147,17 @@ class Database:
         qualified_name = self._qualify(name)
         return self.client.table(qualified_name, self.name)
 
-    def list_tables(self, like=None):
+    def list_tables(self, like=None, database=None):
         """List the tables in the database.
 
         Parameters
         ----------
         like
             A pattern to use for listing tables.
+        database
+            The database to perform the list against
         """
-        return self.client.list_tables(like, database=self.name)
+        return self.client.list_tables(like, database=database or self.name)
 
 
 class TablesAccessor(collections.abc.Mapping):


jcrist (approved):
🚀

PR #4797, opened 2022-11-08
feat(bigquery): move the bigquery backend back into the main ibis repo

cpcloud:
This PR moves bigquery back into the main ibis repo.

~Still working through the failing tests, though many are fixed.~ Tests are passing.

TODOs:

- [x] get all tests passing locally
- [x] see if we can automatically handle the autonaming we're doing that bigquery doesn't accept
- [x] setup ci similar to ibis-bigquery if possible, though maybe we just run these tests on push events only similar to snowflake

Possible follow ups:

- [x] delete <4 legacy code
- [x] delete <4 legacy tests
- [x] move SQL tests to write-then-compare so that they are easy to modify


.github/workflows/ibis-backends-cloud.yml
```diff
@@ -58,10 +61,17 @@ jobs:
...
 
+      - uses: google-github-actions/auth@v1
+        if: ${{ matrix.backend.name == 'bigquery' }}
+        with:
+          credentials_json: ${{ secrets.GCP_CREDENTIALS }}
+
+      - name: setup snowflake credentials
+        if: ${{ matrix.backend.name == 'snowflake' }}
+        run: echo '${{ secrets.SNOWFLAKE_URL }}' >> "$GITHUB_ENV"
+
```
mik-laj, line 72:
```suggestion
        run: echo "SNOWFLAKE_URL=${SNOWFLAKE_URL}" >> "$GITHUB_ENV"
        env:
            SNOWFLAKE_URL: ${{ secrets.SNOWFLAKE_URL }}
```
We should use with 'env' option to avoid script injection attacks or injection problems. See: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections

To set the environment variable, you should also pass the environment variable name. https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

ibis/tests/sql/conftest.py
```diff
@@ -11,15 +11,19 @@
 from ibis.tests.expr.mocks import MockBackend
 
 
-def sql_golden_check(funcname: str, path: Path, expr: ir.Expr | str) -> None:
```
mik-laj: What do you think about using the `pytest-snapshot `package to write these tests? 
Thanks to this, we will be able to update the tests with one command.
```
pytest --snapshot-update.
```
We also won't have to have a separate file for each assertion, but we will have more assertions in one snapshot file.

cpcloud: I put up #4836 to do this for our existing sqlgolden checks. I'll merge that PR when it's green.

jreback:
@cpcloud sure 

is this a general change in policy though? or a specific one off for BQ?

eg what about a lot of the other google variants or mssql for example

ibis/backends/base/sql/registry/window.py
```diff
@@ -54,36 +55,22 @@ def _replace_interval_with_scalar(expr: Union[ir.Expr, dt.Interval, float]):
...
-
-    if not isinstance(expr, (dt.Interval, ir.IntervalValue)):
-        # Literal expressions have op method but native types do not.
-        if isinstance(expr_op, ops.Literal):
-            return expr_op.value
-        else:
-            return expr
-    elif isinstance(expr, dt.Interval):
+    if isinstance(op, ops.Literal):
+        unit = getattr(op.output_dtype, "unit", "us")
```
kszucs, line 59: During the refactors my general experience was that `getattr` this way made it harder to locate type problems by swallowing attribute errors. 
Based on the function's name `_replace_interval_with_scalar` I'd expect to always have an interval dtype here, are other inputs allowed?

ibis/backends/base/sql/registry/window.py
```diff
@@ -104,20 +91,20 @@ def cumulative_to_window(translator, op, window):
 
 def time_range_to_range_window(_, window):
     # Check that ORDER BY column is a single time column:
-    order_by_vars = [x.op().args[0] for x in window._order_by]
+    order_by_vars = [x.args[0] for x in window._order_by]
```
kszucs, line 94: Do we expect `ops.SortKey` instances here? If yes could we use `x.expr` instead?

I know that the window._order_by member is not properly validated yet.

ibis/backends/base/sql/registry/window.py
```diff
@@ -251,15 +215,18 @@ def window(translator, op):
...
     # Time ranges need to be converted to microseconds.
     # FIXME(kszucs): avoid the expression roundtrip
     if window.how == 'range':
-        order_by_types = [type(x.op().args[0]) for x in window._order_by]
-        time_range_types = (ir.TimeColumn, ir.DateColumn, ir.TimestampColumn)
-        if any(col_type in time_range_types for col_type in order_by_types):
+        time_range_types = (dt.Time, dt.Date, dt.Timestamp)
+        if any(
+            isinstance(c.output_dtype, time_range_types)
+            and c.output_shape == rlz.Shape.COLUMNAR
```
kszucs, line 227: You should be able to use `c.output_shape.is_columnar()` here.

kszucs: Perhaps a similar `c.output_dtype.is_temporal()` convenience method would be useful too.

ibis/backends/bigquery/backports.py
```diff
@@ -0,0 +1,132 @@
+"""Backports for changes in Ibis."""
```
kszucs, line 1: Since the backend will be tied with the corresponding ibis core revision now, do we still need this backports file?

ibis/backends/base/sql/registry/window.py
cpcloud: `ops.SortKey`s are allowed, to support `DESC`

.github/workflows/ibis-backends-cloud.yml
cpcloud: Fixed, thanks!

ibis/tests/sql/conftest.py
cpcloud: #4836 is merged!

ibis/backends/base/sql/registry/window.py
cpcloud: Other inputs are allowed, e.g. in `2 * my_interval` the recursion will eventually hit the `2` which will have an `int8` dtype.

ibis/backends/bigquery/__init__.py
```diff
@@ -0,0 +1,470 @@
...
+            names.append(col.name)
+            typename = bigquery_field_to_ibis_dtype(col)
+            adapted_types.append(typename)
+        return names, adapted_types
+
+    def _execute(self, stmt, results=True, query_parameters=None):
+        job_config = bq.job.QueryJobConfig()
+        job_config.query_parameters = query_parameters or []
+        job_config.use_legacy_sql = False  # False by default in >=0.28
+        query = self.client.query(
```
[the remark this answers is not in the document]
kszucs: Great to know, thanks Gil!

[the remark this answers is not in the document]
cpcloud: These warnings are correct, because we don't run bigquery tests until a merge to `master`.

ibis/backends/base/sql/registry/window.py
cpcloud: This was done in #4841.

ibis/backends/bigquery/client.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+_LEGACY_TO_STANDARD = {
+    "INTEGER": "INT64",
+    "FLOAT": "FLOAT64",
+    "BOOLEAN": "BOOL",
+}
+
+
+@dt.dtype.register(bq.schema.SchemaField)
```
kszucs, line 42: Please create a follow-up issue to factor the datatype conversions out to `ibis.backends.bigquery.datatypes`

ibis/backends/bigquery/compiler.py
```diff
@@ -0,0 +1,123 @@
...
+    union_class = BigQueryUnion
+    intersect_class = BigQueryIntersection
+    difference_class = BigQueryDifference
+
+    @staticmethod
+    def _generate_setup_queries(expr, context):
+        """Generate DDL for temporary resources."""
+        # NB: compatibility with 3.x and 4.x
+        try:
+            nodes = lin.traverse(find_bigquery_udf, expr, filter=(ops.Node, ir.Expr))
```
kszucs, line 116: `nodes = lin.traverse(find_bigquery_udf, expr, filter=ops.Node)` should be sufficient

ibis/backends/bigquery/client.py
cpcloud: Would you mind if I did that in a follow-up?

ibis/backends/bigquery/compiler.py
cpcloud: Yeah, this was leftover from an attempted update to support 4.0.0 in the external backend. I'll fix this up.

kszucs (approved):
Just nits, other than those it's good to go from my side

ibis/backends/bigquery/client.py
kszucs: Not at all, just have an issue to track this

cpcloud:
@tswast Would you mind having a look here?

There are a lot of files, but most of the new files are generated SQL tests to make it easier to update expected code when tests break but don't change semantics, so you can likely ignore those.

I was able to either preserve each test as it was in `ibis-bigquery` or generalize it and move to a backend test, so that should give us some confidence about functionality going forward.

ibis/backends/bigquery/compiler.py
cpcloud: Fixed!

cpcloud:
Ok, this is no longer blocked on https://github.com/googleapis/python-bigquery/pull/1413, I set `python = "<3.11"` for the `google-cloud-bigquery` dependency, which we can remove once that PRs commit is in a release.

ibis/backends/bigquery/client.py
cpcloud: Done: https://github.com/ibis-project/ibis/issues/4881

ibis/backends/bigquery/tests/unit/udf/test_core.py
```diff
@@ -0,0 +1,552 @@
...
+def compile(f):
+    return PythonToJavaScriptTranslator(f).compile()
+
+
+def test_function_def():
+    def f(a, b):
+        return a + b
+
+    expected = """\
+function f(a, b) {
```
tswast, line 25: Are these tests ones that we'd want to parametrize with `snapshot` as well?

tswast (commented):
BQ changes LGTM. I like the "snapshot" structure in the tests.

ibis/backends/bigquery/tests/unit/udf/test_core.py
cpcloud: Yeah, I think so!

cpcloud: Done.

ibis/backends/bigquery/__init__.py
```diff
@@ -350,7 +350,20 @@ def exists_table(self, name: str, database: str = None) -> bool:
 
     def fetch_from_cursor(self, cursor, schema):
         query = cursor.query
-        df = query.to_arrow().to_pandas(timestamp_as_object=True)
+        query_result = query.result()
```
cpcloud, line 353: @tswast Can you take a look at this block of code and say whether this is expected behavior?

The use case is reading from in `bigquery-public-data.hacker_news.comments`, but having `ibis-gbq` be the billing project.

Without this workaround, the storage API creates a read session in the data project (`bigquery-public-data`), which causes queries to fail when using the pyarrow functionality.

tswast: I wouldn't expect this to be necessary. If the query succeeded, then I assume the billing project is being set correctly in the client constructor https://github.com/ibis-project/ibis/blob/01bd4026e463984ab9fd17ea6d57ee6e42d50b8b/ibis/backends/bigquery/__init__.py#L169 and in the query method https://github.com/ibis-project/ibis/blob/01bd4026e463984ab9fd17ea6d57ee6e42d50b8b/ibis/backends/bigquery/__init__.py#L236

I've filed https://github.com/googleapis/python-bigquery/issues/1422 to investigate this further, but I think it's fine to keep this workaround if there really is a bug.

tswast: Update: I think there really is a bug. The project from "client" is used instead of the project from the QueryJob.

cpcloud: Ah, okay. Thanks Tim. I'll keep this comment unresolved so the link is easier to find.

cpcloud:
Ok, I'm going to merge this in and fix any issues with the CI. Thanks all for the help reviewing, great to see this back in the main repo!

mik-laj (approved):

PR #4798, opened 2022-11-08
feat(sqlite): support `con.sql` with explicit schema specified

jcrist:
This adds support for `con.sql` to SQLite. It does this by exposing an optional `schema` kwarg to `.sql` for all backends, allowing the query schema to be specified explicitly. This is optional for all backends except SQLite where it is required - if not specified a nice error message is raised.

Note that SQLite still doesn't support `TableExpr.sql`, since that requires temporary view creation.

This also rearranges the `.sql` tests a bit to better test the supported backends. In particular, `clickhouse` and `impala` do support `con.sql` but not `TableExpr.sql`. But since all backend tests were using the latter these backends weren't being tested.

### DIFF
--- a/ibis/backends/snowflake/__init__.py
+++ b/ibis/backends/snowflake/__init__.py
@@ -106,7 +106,7 @@ class Backend(BaseAlchemyBackend):
 
     def _get_schema_using_query(self, query):
         with self.begin() as bind:
-            result = bind.execute(query)
+            result = bind.execute(f"SELECT * FROM ({query}) t0 LIMIT 0")
             info_rows = bind.execute(f"DESCRIBE RESULT {result.cursor.sfqid!r}")
 
         schema = {}
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -90,23 +90,26 @@ class BaseSQLBackend(BaseBackend):
         # XXX
         return name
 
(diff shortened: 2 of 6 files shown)


ibis/backends/snowflake/__init__.py
```diff
@@ -106,7 +106,7 @@ def fetch_from_cursor(self, cursor, schema: sch.Schema) -> pd.DataFrame:
 
     def _get_schema_using_query(self, query):
         with self.begin() as bind:
-            result = bind.execute(query)
+            result = bind.execute(f"SELECT * FROM ({query}) t0 LIMIT 0")
```
cpcloud, line 109: Why do we have to inline these `LIMIT 0` queries? Seems like having this in one place if we can is better than repeating it.

jcrist: `_get_schema_using_query` is called from both `con.sql` and `Table.sql` - the latter wasn't adding the `LIMIT 0` automatically, while the former was. While looking into this I noticed that:

- Some backends were already manually adding the `LIMIT 0` in their implementation
- Some backends had no need for a `LIMIT 0`

It felt cleaner to me to have `_get_schema_using_query` receive the raw query and handle it however the backend needs rather than relying on the caller to add a 0 limit automatically.

cpcloud: Ah, ok. Yeah, that's cleaner.

PR #4799, opened 2022-11-08
ci: drop coverage from tpch tests

jcrist:
The tpch tests have been failing *very* frequently lately during the codecov upload step. Since these tests are extra and shouldn't be testing any specific functionality, I don't feel like we really need coverage here - dropping coverage entirely for tpch rather than debugging the cause of the failures.


cpcloud (approved):
Yeah, SGTM.

PR #4804, opened 2022-11-08
fix: a few `startswith`/`endswith` fixes

jcrist:
- Improve the tests for `startswith`/`endswith`; previously the test case resulted in a fully `False` series, letting several bugs slip through.
- Fix `sqlite` and `mysql` implementations of `startswith`/`endswith` to be case sensitive
- Fix bug in `polars` implementation of `startswith` to call the proper method

### DIFF
--- a/ibis/backends/mysql/registry.py
+++ b/ibis/backends/mysql/registry.py
@@ -29,6 +29,19 @@ def _capitalize(t, op):
     )
 
 
+def _startswith(t, op):
+    arg = t.translate(op.arg)
+    start = t.translate(op.start)
+    # LIKE in mysql is case insensitive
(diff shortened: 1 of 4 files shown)


ibis/backends/mysql/registry.py
```diff
@@ -29,6 +29,19 @@ def _capitalize(t, op):
     )
 
 
+def _startswith(t, op):
+    arg = t.translate(op.arg)
+    start = t.translate(op.start)
+    # LIKE in mysql is case insensitive
+    return arg.op("LIKE BINARY")(sa.func.concat(start, "%"))
```
cpcloud, line 36: TIL about `LIKE BINARY`!

cpcloud (approved):
LGTM, thanks!

PR #4815, opened 2022-11-09
fix(snowflake): fix snowflake list_databases

frsann:
The current implementation relies on the base sqlalchemy backend implementation which returns schemas, instead of databases. This PR fixes this by implementing a custom function for the Snowflake backend, which utilizes the `information_schema.databases` view ([docs](https://docs.snowflake.com/en/sql-reference/info-schema/databases.html))


cpcloud (approved):
LGTM. Ideally we can add a test, but if you've tested locally then that's good enough.

PR #4824, opened 2022-11-11
feat(snowflake): impl day of week name functions

cpcloud:
The translation rules for this function exists, and the tests are now
`XPASS`-ing.


cpcloud:
Hm, this is still failing for me locally. I'm going to close this out until I can figure out what's happening.

PR #4827, opened 2022-11-11
docs: add issue templates

cpcloud:


mik-laj:
What do you think about using [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) instead of plain text?

cpcloud:
Ah, nice. Yeah, that'd be better.

cpcloud:
Merging, so that I can see what it looks like.

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,47 @@
...
+      required: true
+  - type: textarea
+    id: backends
+    attributes:
+      label: Backend(s)
+      description: What backend(s) are you using, if any?
+      placeholder: "Ex: DuckDB"
+    validations:
+      required: false
+  - type: checkboxes
```
mik-laj, line 40: Screenshot 2022-11-11 at 16 51 06
I like to add a question to the forms for a new feature whether the person is interested in contributing. If so, it means that this ticket may need to be prioritized so as not to discourage the person from contributing.

cpcloud: Great idea, I'll push up a PR for that :)

PR #4828, opened 2022-11-11
feat(snowflake): implement `rowid` scalar

mik-laj:
Hello,

The follwoing code:
```python
t = con.table('functional_alltypes')
result = t[t.rowid()].execute()
```
io translated to the following SQL:
```sql
SELECT seq8() + 1 AS number 
FROM functional_alltypes AS t0
```


cpcloud (approved):
Awesome, thank you! Merging.

PR #4836, opened 2022-11-14
test: use `pytest-snapshot` for testing generated SQL

cpcloud:
This PR uses pytest-snapshot (instead of our existing sqlgolden decorator) to test generated SQL.

The main reason to use a plugin is to integrate with the rest of pytest's features, namely parameterization. The plugin also localizes snapshot directories to the directory containing the test file. With `sqlgolden` parameterized tests are clobbered and any backends that ever want to test their own compilation will also be clobbered if they happen to have tests that have the same name. None of these issues exist with `pytest-snapshot`.


cpcloud:
In the interest of getting started on the rebase in #4797 as a result of this PR I'm going to merge it.

PR #4839, opened 2022-11-14
fix: support dtype in `__array__` methods

jcrist:
Previously the `__array__` methods were missing this parameter, causing `np.asarray(table, dtype="some_dtype")` to fail.

Also changes up the `__array__` test a bit so it works on all backends.


cpcloud (approved):
Thanks!

PR #4841, opened 2022-11-15
feat(ir): add DataType convenience methods

### LINKED ISSUE #4409: feat(ir): give datatype some conveniences like `is_boolean`, `is_integer`, `is_int64`
Let's make a follow-up issue to give datatype some Rust `enum`-like conveniences like `is_boolean`, `is_integer`, `is_int64` etc.

_Originally posted by @cpcloud in https://github.com/ibis-project/ibis/pull/4378#discussion_r953733703_

cpcloud:
Closes #4409.

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -78,195 +78,147 @@ class DataType(Concrete):
 
         return castable(self, other, **kwargs)
 
-    @property
     def is_array(self) -> bool:
         return isinstance(self, Array)
 
-    @property
     def is_binary(self) -> bool:
         return isinstance(self, Binary)
 
-    @property
     def is_boolean(self) -> bool:
         return isinstance(self, Boolean)
 
-    @property
     def is_category(self) -> bool:
         return isinstance(self, Category)
 
-    @property
     def is_date(self) -> bool:
         return isinstance(self, Date)
 
-    @property
     def is_decimal(self) -> bool:
         return isinstance(self, Decimal)
 
-    @property
     def is_enum(self) -> bool:
         return isinstance(self, Enum)
 
-    @property
     def is_float16(self) -> bool:
         return isinstance(self, Float16)
 
-    @property
     def is_float32(self) -> bool:
         return isinstance(self, Float32)
 
-    @property
     def is_float64(self) -> bool:
         return isinstance(self, Float64)
 
-    @property
     def is_floating(self) -> bool:
         return isinstance(self, Floating)
 
-    @property
     def is_geography(self) -> bool:
         return isinstance(self, Geography)
 
-    @property
     def is_geometry(self) -> bool:
         return isinstance(self, Geometry)
 
-    @property
     def is_geospatial(self) -> bool:
         return isinstance(self, GeoSpatial)
 
-    @property
     def is_inet(self) -> bool:
         return isinstance(self, INET)
 
-    @property
     def is_int16(self) -> bool:
         return isinstance(self, Int16)
 
-    @property
     def is_int32(self) -> bool:
         return isinstance(self, Int32)
 
-    @property
     def is_int64(self) -> bool:
         return isinstance(self, Int64)
 
-    @property
     def is_int8(self) -> bool:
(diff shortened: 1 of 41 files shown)


ibis/expr/datatypes/core.py
```diff
@@ -78,6 +78,198 @@ def castable(self, other):
 
         return castable(self, other)
 
+    @property
```
kszucs, line 81: I may prefer methods over properties, though I'm not sure. Do you have a strong preference?

ibis/expr/datatypes/core.py
```diff
@@ -78,6 +78,198 @@ def castable(self, other):
...
+    def is_uint64(self) -> bool:
+        return isinstance(self, UInt64)
+
+    @property
+    def is_uint8(self) -> bool:
+        return isinstance(self, UInt8)
+
+    @property
+    def is_unsigned_integer(self) -> bool:
```
kszucs, line 262: missing `is_signed_integer`

ibis/tests/expr/test_datatypes.py
```diff
@@ -545,3 +545,97 @@ def test_traversal_replace():
     result = dtype.replace(subs)
     expected = dt.Struct.from_tuples([('a', 'bool'), ('b', dt.Array('timestamp'))])
     assert result == expected
+
+
+def get_leaf_classes(op):
```
kszucs, line 550: Would it make sense to use this utility for actually defining the methods?

ibis/expr/datatypes/core.py
kszucs: Either way we should update rlz.Shape enum to follow the same convention: https://github.com/ibis-project/ibis/blob/master/ibis/expr/rules.py#L32-L36

ibis/expr/datatypes/core.py
kszucs: never mind, it's upper at line 222

ibis/expr/datatypes/core.py
jcrist: I also lean slightly towards methods here, probably due to familiarity with the pyarrow `is_*` functions.

ibis/expr/datatypes/core.py
```diff
@@ -78,6 +78,198 @@ def castable(self, other):
...
+    def is_float16(self) -> bool:
+        return isinstance(self, Float16)
+
+    @property
+    def is_float32(self) -> bool:
+        return isinstance(self, Float32)
+
+    @property
+    def is_float64(self) -> bool:
+        return isinstance(self, Float64)
```
jcrist, line 119: Do you have a use case in mind for distinct methods for every datatype? The main use case I have for something like this is detecting classes of datatypes (e.g. `is_numeric()`, `is_floating()`, `is_temporal()`) rather than anything specific like `is_float64()`. Having a method for everything is fine, just curious about the use case.

cpcloud: I generated these methods and didn't want to curate them to arbitrarily exclude things.

It also seems a bit arbitrary to include leaf type versions of these for arrays (for example) but then exclude other leaf types.

Strings in theory are a leaf type but right now the JSON type subclasses it. However, if String were a leaf type I think we'd want an `is_string`.

I'll replace the cases of instance checking in the codebase where I can to see which of these are most used. This may inform how valuable (or not) having some of these more specific methods is.

ibis/expr/datatypes/core.py
cpcloud: Slight preference for properties, but I also don't feel too strong about it.

ibis/tests/expr/test_datatypes.py
cpcloud: Maybe, but let's see how it is to use the manually defined ones first.

ibis/expr/datatypes/core.py
cpcloud: Changed to methods.

kszucs (approved):

jcrist (approved):

PR #4845, opened 2022-11-15
[WIP] feat: support `con.create_table(name, expr)` where `expr` is from another backend

### LINKED ISSUE #4800: Create/Insert data into one backend from another backend
I'd like to be able to take a query from one backend and use it to create a table (or insert into an existing table) in another backend.

Example user session:

```python
import datetime

import ibis
from ibis import _

# Get the last day's worth of data from bigquery
bigquery = ibis.bigquery.connect(...)
query = bigquery.tables.some_table.filter(_.timestamp >= (ibis.now() - datetime.timedelta(days=1))

# Save it to a local sqlite database
sqlite = ibis.sqlite.connect("local.db")
sqlite.create_table("snapshot", query)
```

A few use cases:

- Downloading a subset of data from a large remote database (say bigquery) into a smaller local database (say duckdb or sqlite) for further analysis.
- ETL workloads where the data source and sink are in different databases (for example, load the last day's worth of data from bigquery, transform it, then insert it into an existing table in postgres).

I'm not sure if there are backend pairs where the data transfer wouldn't have to go through the ibis client process directly (meaning writing from backend1 to backend2 might have some optimized path), but if there are then it might be useful to design in a way that we could support that at some point.

For now a default implementation *might* be using the `to_pyarrow_batches` method on the source table and doing incremental writes to the sink table.

jcrist:
Fixes #4800.

Still needs tests and `insert` support.

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import contextlib
 import getpass
-from operator import methodcaller
 from typing import Any, Literal
 
 import pandas as pd
@@ -186,7 +185,7 @@ class BaseAlchemyBackend(BaseSQLBackend):
 
         if database is not None:
             raise NotImplementedError(
-                'Creating tables from a different database is not yet ' 'implemented'
+                'Creating tables from a different database is not yet implemented'
             )
 
         if expr is None and schema is None:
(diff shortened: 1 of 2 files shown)


ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -206,31 +205,46 @@ def create_table(
...
+        self._register_in_memory_tables(expr)
+
+        backends, _ = expr._find_backends()
+        if len(backends) == 1:
+            same_backend = backends[0] == self
+        elif len(backends) > 1:
+            raise ValueError('Multiple backends found')
+        else:
+            # Treat unbound exprs as based on this backend, matching `execute`
+            same_backend = True
```
cpcloud, line 224: Is this logic something that should be extracted into the parent SQL backend class?

cpcloud (commented):
Wondering whether this is worth generalizing to the non-SQLAlchemy-but-still-SQL backends.

cpcloud:
@jcrist Is this still a WIP or is it ready for a round of review?

jcrist:
Enough time has passed (and internals changed) that this would be better handled by just redoing the work, rather than updating this PR. Closing for now as stale.

PR #4848, opened 2022-11-16
refactor(ir): simplify analysis.substitute_unbound()

kszucs:


ibis/expr/analysis.py
```diff
@@ -222,26 +222,9 @@ def substitute_unbound(node):
...
-            # occur
-            table = node.table
-
-            if isinstance(table, ops.Selection):
-                for val in table.selections:
-                    if isinstance(val, ops.PhysicalTable) and node.name in val.schema:
-                        return ops.TableColumn(val, node.name)
-        elif isinstance(node, ops.Join):
-            return node.__class__(
-                substitute_unbound(node.left),
```
kszucs, line 239: The substitution happens in depth-first order so these nodes should have database tables already replaced.

cpcloud (approved):
Nice find!

ibis/expr/analysis.py
gforsyth: For whatever reason, this doesn't descend into join predicates and also leaves some `TableColumn`s bound to the previous parent table.

kszucs: Do we have tests for that case?

kszucs: Probably because the join predicates don't have proper rules set: `Join.predicates = rlz.optional(lambda x, this: x, default=())`, so it's a tuple instead of a nodelist.

gforsyth: Ahh, yeah, it was definitely tuple, hence the `map` call in my original version

kszucs: Sadly we still have missing rule definitions scattered across the codebase.

cpcloud:
@kszucs Should we merge the original while you address the missing rules?

cpcloud:
Nevermind, I'm tired 😪

kszucs:
Updated to use the more recent traversal methods.

gforsyth (approved):
Nice, this looks good to me.

cpcloud (approved):
Ship it!

PR #4849, opened 2022-11-16
fix: define `to_pandas`/`to_pyarrow` on DataType/Schema classes directly

jcrist:
Previously these were being monkeypatched onto the classes only if the submodules were imported. This leads to these methods not always existing (the pandas ones were always added, the pyarrow ones were only added if the user ever imported `ibis.backends.pyarrow.datatypes` themselves). Better to define them directly on the classes and avoid this complication.


gforsyth (approved):
:shipit:

PR #4853, opened 2022-11-17
feat(ir): prototype for parsing SQL into an ibis expression

kszucs:
depends on #4535 

Early prototype for using sqlglot as parser, but it can serve as a base for further work. 

Example converting the following SQL string to ibis code:

Input:

```sql
SELECT *, first_name as first FROM employee 
LEFT JOIN call ON employee.id = call.employee_id 
WHERE id < 5 
ORDER BY id DESC
```

```py
catalog = {
    "employee": {"first_name": "string", "last_name": "string", "id": "int64"},
    "call": {
        "start_time": "timestamp",
        "end_time": "timestamp",
        "employee_id": "int64",
        "call_outcome_id": "int64",
    },
    "call_outcome": {"outcome_text": "string", "id": "int64"},
}
```

Usage:

```py
expr = ibis.parse_sql(sql, catalog)

print(ibis.decompile(expr, format=True))
```

Output:

```py
employee = ibis.table(
    name="employee",
    schema={"first_name": "string", "last_name": "string", "id": "int64"},
)
call = ibis.table(
    name="call",
    schema={
        "start_time": "timestamp",
        "end_time": "timestamp",
        "employee_id": "int64",
        "call_outcome_id": "int64",
    },
)
proj = employee.left_join(call, employee.id == call.employee_id).filter(employee.id < 5)

result = proj.select(
    [
        proj.first_name,
        proj.last_name,
        proj.id,
        call.start_time,
        call.end_time,
        call.employee_id,
        call.call_outcome_id,
        proj.first_name.name("first"),
    ]
).order_by(proj.id.desc())
```


ibis/expr/sequel.py
```diff
@@ -0,0 +1,225 @@
+from __future__ import annotations
```
cpcloud, line 1: Any issue with calling this `sql.py`? Seems a bit clever to call it `sequel.py` 😉

cpcloud:
I know it's a bit buried in the output, but `poetry.lock` and `pyproject.toml` don't match: https://github.com/ibis-project/ibis/actions/runs/3496250209/jobs/5853948961#step:6:383

tobymao:
incredible

cpcloud:
@kszucs Looks like there are some snapshots that aren't being ignored.

cpcloud (approved):

kszucs:
Note, that this is far from being complete and bug free.

cpcloud:
It looks like you've updated the lockfile to include `numpy 1.24` which breaks any library using `np.bool` since it's apparently been removed in that version.

Gotta revert back to the upstream lock file, and then run `poetry lock --no-update` which will perform appropriate additions and deletions of dependencies but will not upgrade anything that doesn't need to be upgraded.

cpcloud:
I'll push those changes here in a bit.

kszucs:
> I'll push those changes here in a bit.

Oops, I pushed it too.

kszucs:
@cpcloud shall we merge it so we can iterate with it more easily?

I marked this feature as experimental.

cpcloud:
@kszucs SGTM, let's merge it!

PR #4861, opened 2022-11-18
fix(deps): bound shapely to a version that doesn't segfault

### LINKED ISSUE #4856: bug: segfault when shapely and duckdb are both installed
### What happened?

I expect `import ibis.backends.duckdb` not to segfault.

It's likely that https://github.com/shapely/shapely/issues/1616 is the cause.

`master`

DuckDB

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #4856.

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -61,6 +61,8 @@ jobs:
             has_geo: true
             services:
               - mysql
+            sys-deps:
+              - libgeos-dev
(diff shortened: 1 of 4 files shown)


kszucs (approved):

kszucs:
I'm getting a segfault on macOS using `shapely==1.7.1` while `shapely==1.8.5.post1` works, we will need to update the pin because it seems to be architecture dependent.

PR #4866, opened 2022-11-18
fix: allow inserting memtable with alchemy backends

### LINKED ISSUE #4683: Inconsistent support of pandas.DataFrame in `create_table` and `insert`

gforsyth:
Now if a user passes a `memtable` to `insert`` we'll pull out the
underlying dataframe and use that to update the table.

Resolves #4683

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -473,6 +473,12 @@ class BaseAlchemyBackend(BaseSQLBackend):
                 'yet implemented'
             )
 
+        # If we've been passed a `memtable`, pull out the underlying dataframe
+        if isinstance(obj, ir.Table) and isinstance(
+            in_mem_table := obj.op(), ops.InMemoryTable
+        ):
+            obj = in_mem_table.data.to_frame()
+
         if isinstance(obj, pd.DataFrame):
             obj.to_sql(
                 table_name,
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -349,6 +349,21 @@ def test_insert_overwrite_from_list(
     assert len(alchemy_con.table(employee_data_1_temp_table).execute()) == 3
 
 
+def test_insert_from_memtable(alchemy_con):
+    df = pd.DataFrame({"x": range(3)})
+    table_name = "memtable_test"
+    alchemy_con.insert(table_name, ibis.memtable(df))
+    alchemy_con.insert(table_name, ibis.memtable(df))
+
+    try:
+        table = alchemy_con.tables[table_name]
+        assert len(table.execute()) == 6
+        assert alchemy_con.tables[table_name].schema() == ibis.schema({"x": "int64"})
+    finally:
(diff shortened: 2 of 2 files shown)


ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -473,6 +474,10 @@ def insert(
                 'yet implemented'
             )
 
+        # If we've been passed a `memtable`, pull out the underlying dataframe
+        if isinstance(in_mem_table := obj.op(), PandasInMemoryTable):
+            obj = in_mem_table.data._df
```
cpcloud, line 479:
```suggestion
        if isinstance(in_mem_table := obj.op(), InMemoryTable):
            obj = in_mem_table.data.to_frame()
```

This will handle any subclass of `InMemoryTable`

cpcloud (changes requested):

ibis/backends/base/sql/alchemy/__init__.py
gforsyth: ahh, cool.  Some other stuff was also borked, but should be fixed up now.

cpcloud (approved):

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: Sweet!

ogrisel:
Thanks for the fix!

PR #4874, opened 2022-11-20
bug: `re_extract` behavior differs between backends

### LINKED ISSUE #4593: bug: `re_extract` behavior differs between backends
The behavior of `re_extract` differs between backends, with regards to what the `index` parameter means.

This method has signature `re_extract(pattern, index)`, where `pattern` is a regex pattern with optional groups, and `index` is an index of the group to extract (returning `NULL` if no match or no group matches that index).

- `duckdb`, `postgres`, `clickhouse`, ...: `re_extract(pattern, 0)` returns the part matching the first group if there's a match, and `NULL` otherwise.
- `sqlite`, `pandas`, `dask`, `pyspark`: `re_extract(pattern, 0)` returns the _whole string_ if there's a match, and `NULL` otherwise. You need to pass in `1` not `0` to extract the first group.

To put it another way, given a column with a value `"row_one"`, `column.re_extract("row_([a-z])", 0)` returns `"one"` for backends in the first group, and `"row_one"` for backends in the second.

Given the docstring, I _think_ the first group has the intended behavior.

cpcloud:
This PR makes `StringValue.re_extract` conform to the semantics of `re.match` in the Python standard library.

Fixes #4593.

### DIFF
--- a/ibis/backends/clickhouse/registry.py
+++ b/ibis/backends/clickhouse/registry.py
@@ -220,14 +220,28 @@ def _string_find(translator, op):
 
 
 def _regex_extract(translator, op):
-    arg_ = translator.translate(op.arg)
-    pattern_ = translator.translate(op.pattern)
-    index = op.index
+    arg = translator.translate(op.arg)
+    pattern = translator.translate(op.pattern)
+    index = "Null" if op.index is None else translator.translate(op.index)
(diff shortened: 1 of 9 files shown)


kszucs (approved):
LGTM

PR #4877, opened 2022-11-20
refactor(ir): replace enum functionality with string

### LINKED ISSUE #4477: refactor(ir): replace enum functionality with string
It is being used only in the impala codebase.

cpcloud:
This PR aliases Enum to String. See commit messages for details. Closes #4477.

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -753,17 +753,6 @@ class Set(Variadic):
 
 
 @public
-class Enum(DataType):
-    """Enumeration values."""
-
-    rep_type = datatype
-    value_type = datatype
-
-    scalar = ir.EnumScalar
-    column = ir.EnumColumn
-
-
-@public
 class Map(Variadic):
     """Associative array values."""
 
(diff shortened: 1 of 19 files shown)


kszucs:
@cpcloud the impala backend has failing tests

cpcloud:
Turns out that `dt.Category` is providing some value in the form of tracking the cardinality of expressions.

For example, bucketing an expression using a list of bins of say, length 4, will be tracked in the cardinality property of the `Category` dtype which allows validation of the labels to the `CategoryValue.label` API.

I'm going to leave Category as-is for now and just do the enum aliasing in this PR.

ibis/expr/datatypes/core.py
```diff
@@ -995,4 +986,5 @@ class INET(String):
     macaddr=macaddr,
     inet=inet,
     decimal=decimal,
+    Enum=Enum,
```
kszucs, line 989: `Enum=String`? Shouldn't we deprecate it?

kszucs:
> I'm going to leave Category as-is for now and just do the enum aliasing in this PR.

Need to update the Pr's title then

cpcloud:
@kszucs Done, but looks like there's still some failing impala tests. Looking 👀

kszucs (approved):

PR #4878, opened 2022-11-20
refactor(ir): remove `TopK` abstraction

### LINKED ISSUE #4403: refactor(ir): remove `ir.TopK`

cpcloud:
This PR removes classes related to TopK as an abstract expression.
`t[t.x.topk(...)]` becomes `t.x.topk_filter(...)`.

I'm in no way committed to `t.x.topk_filter(...)` as the replacement for `t[t.x.topk(...)]`.

Some possible alternatives:

1. ~`t.x.topk(k, by, filter=True|False)`, IMO `filter` is doing too much and all the table extraction make this very not-column-ish~
1. ~`t.filter_topk(expr, k, by)`, tab completable alongside `t.filter` and thus discoverable~
1. ~`t.topk(expr, k, by, filter=True|False)`, IMO confusing to have filter have such a dramatic effect~
1. ~`t.topk(expr, k, by)` is always a filter, and the `Column` method is always
   the agg version. This is somewhat confusing because of the name equivalence.~
1. ✅  Remove it altogether and add a how to. **EDIT:** Since the equivalent semi join expression is so short we went with this.

Closes #4403.


jcrist:
> Remove it altogether and add a how to.

I'm leaning towards this option.

jreback:
@cpcloud this is quite a common operation; why are you proposing removing?

cpcloud:
@jreback The `t.x.topk(...)` operation isn't going anywhere.

What's potentially on the table for removal is

```t[t.x.topk(...)]```

Keeping the filter shorthand around doesn't seem worth the additional cruft, since it's two lines of code, which seems suitable for a how-to docs page:

```
topk = t.x.topk(10, by=b)
filtered = t.semi_join(topk, "x")
```

jreback:
ah ok then agree documenting is prob sufficient 

thanks

kszucs (approved):

PR #4879, opened 2022-11-20
feat(backends/mssql): Add backend support for Microsoft Sql Server

oneextrafact:


ibis/backends/mssql/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def do_connect(
+        self,
+        host: str = "localhost",
+        user: str | None = None,
+        password: str | None = None,
+        port: int = 1433,
+        database: str | None = None,
+        url: str | None = None,
+        driver: Literal["pyodbc"] = "pyodbc",
```
cpcloud, line 23: This should be `pymssql`.

ibis/backends/mssql/tests/conftest.py
```diff
@@ -0,0 +1,121 @@
...
+                ),
+                database=database,
+                schema=schema,
+                isolation_level="AUTOCOMMIT",
+            )
+            with engine.begin() as con:
+                for table in TEST_TABLES:
+                    csv_path = data_dir / f"{table}.csv"
+                    lines = [
+                        f"LOAD DATA LOCAL INFILE {str(csv_path)!r}",
```
cpcloud, line 89: Haven't gotten this working just yet.

oneextrafact: I keep getting "Adaptive Server is unavailable or does not exist", but I can use the same url to make an engine in sqlalchemy outside of the nix-shell environment...

cpcloud: I was able to get this working by setting up a volume mount to `/data` from `ci/ibis-testing-data` and using `BULK INSERT`.

cpcloud: The volume mount is set up in `docker-compose.yml`, so this should be usable in CI and developers

cpcloud:
This is now passing CI, and is ready for review.

cpcloud:
@oneextrafact It looks like we're all green now! Let me know if you're still having trouble running the tests.

saulpw (approved):

PR #4884, opened 2022-11-22
refactor: make shapely import lazy

jcrist:
Previously if `shapely` was installed, `ibis` would import it at import time. Most users don't use the geospatial functionality of ibis, so making this import eager adds unnecessary import time. Further, recent `shapely` builds segfault when mixed with duckdb (#4856).

To accomplish this we:

- Add a `lazy_singledispatch` dispatcher. This works much like `functools.singledispatch`, but supports registering implementations lazily based on a toplevel module name (e.g. `"pandas"`). When an object with a type from that module is encountered, the lazy loader method is called, then the dispatch proceeds as normal. This lets us load the shapely methods only when needed. The use of this lazy loader required moving some definitions around, and making chanages to `dt.infer` and `dt.normalize`. No functionality changes should have occurred.
- Makes `instance_of` support string qualnames for types rather than only types. This lets us lazily load the shapely types for `rlz.instance_of` rather than having to eagerly import it.

While working on this I noticed some dispatch methods were defined in the pandas backend, which isn't always imported. This would lead to different behavior depending on imports used. Some of this was fixed while refactoring `dt.infer`/`dt.normalize`, but there are some further definitions that may need to be moved out to `ibis.expr.datatypes`. I think this makes sense - anything that affects all backends shouldn't be relegated to live in a single backend IMO. I've only done the minimal work here needed to make `shapely` imports lazy though, the rest will be handled in a followup PR.


ibis/common/validators.py
```diff
@@ -58,6 +58,57 @@ def __repr__(self):
 validator = Curried
 
 
+class instance_of(Validator):
```
kszucs, line 61: Could we keep the original validator by combining it with a lazy one?

Something like the following should work:

```py
rlz.one_of([
    rlz.instance_of(...), 
    rlz.lazy_instance_of("shapely.<...>")
])
```

jcrist: Sure, that's a simpler idea. Will do.

ibis/common/tests/test_dispatch.py
```diff
@@ -0,0 +1,122 @@
...
+    assert foo(1.0, 2) == -3.0
+    assert foo(1, 2, c=3) == 6
+
+
+def test_lazy_singledispatch_lazy_loader():
+    @lazy_singledispatch
+    def foo(a):
+        return a
+
+    @foo.lazy_loader("decimal")
```
kszucs, line 61: Just wondering, could we directly register the fully qualified symbol as a string instead of adding a lazy loader?

```py
@foo.lazy_loader("decimal.Decimal")
def inc_decimal(a):
    return a + 1
```

jcrist: We _could_ do that, but in practice we'd have to register all types from a lazily loaded module all at the same time to ensure subclasses were handled properly by the dispatch table (effectively recreating the pattern we have now, just hiding it from the user).

If you think a single `foo.register` method that takes in a string type representation is more ergonomic, then I'm happy to change it, it'll just require some more code to make work.

kszucs: Yes, please do :)

ibis/common/dispatch.py
```diff
@@ -106,3 +108,81 @@ def __call__(self, s: str, *args: Any, **kwargs: Any) -> Any:
...
+                    return impl
+        # Can never get here, since a base `object` implementation is
+        # always registered
+        assert False, "should never get here"  # pragma: no cover
+
+    @functools.wraps(func)
+    def call(arg, *args, **kwargs):
+        return dispatch(type(arg))(arg, *args, **kwargs)
+
+    call.dispatch = dispatch
```
kszucs, line 184: Would it make sense to factor this out to a class?

jcrist: This is implemented in the same style as `functools.singledispatch`. A class can work, but brings with it different introspection (and serialization) issues that just using a function and some closures avoids. Would rather keep things this way if possible.

ibis/common/validators.py
jcrist: done.

ibis/common/tests/test_dispatch.py
jcrist: done.

ibis/common/dispatch.py
```diff
@@ -106,3 +111,72 @@ def __call__(self, s: str, *args: Any, **kwargs: Any) -> Any:
...
+            elif isinstance(cls, str):
+                module = cls.split(".", 1)[0]
+                lazy_lookup[module][cls] = func
+            else:
+                lookup[cls] = func
+            return func
+
+        return inner if func is None else inner(func)
+
+    def dispatch(cls):
```
cpcloud, line 142: This all seems like it should be in its own library, but I guess we can always do that later!

ibis/common/tests/test_dispatch.py
```diff
@@ -0,0 +1,118 @@
+import collections
+import decimal
+
+from ibis.common.dispatch import lazy_singledispatch
+
+
+def test_lazy_singledispatch():
+    @lazy_singledispatch
```
cpcloud, line 8: Is there any reason not to use `lazy_singledispatch` everywhere, other than that the eager version isn't maintained by us?

cpcloud (approved):
LGTM! @kszucs Any more thoughts here?

ibis/common/tests/test_dispatch.py
kszucs: Thanks!

ibis/common/dispatch.py
kszucs: Makes sense.

ibis/common/tests/test_dispatch.py
kszucs: This implementation actually looks cleaner than the one in functools. One major difference is that `functools.singledispatch()` supports type annotation, though that's something we could live without. I'd be curious of a `first dispatch` benchmark between the two.

ibis/common/validators.py
```diff
@@ -89,11 +90,36 @@ def instance_of(klasses, arg, **kwargs):
...
+    """A version of `instance_of` that accepts class qualnames instead of "
+    "already imported classes.
+
+    Useful for delaying imports.
+    """
+
+    def __init__(self, classes):
+        classes = (classes,) if isinstance(classes, str) else tuple(classes)
+        self._classes = classes
+        self._check = lazy_singledispatch(lambda x: False)
```
kszucs, line 108: Would it be possible to factor out the lazy loading part from `lazy_singledispatch` without the dispatching functionality? 

```py
@validator
def lazy_instance_of(types, obj):
    concrete_types = lazy_load_types(types, module=get_module_name(obj))
    if not isinstance(obj, concrete_types):
        raise ...
    return obj
```

where `lazy_load_types` (with a better name) would correspond to the second step in the lazy singledispatch impl with a global lookup cache:

```py
            # 2. Check lazy implementations
             module = cls2.__module__.split(".", 1)[0]
             if lazy := lazy_lookup.get(module):
                 # Import all lazy implementations first before registering
                 # (which should never fail), to ensure an error anywhere
                 # doesn't result in a half-registered state.
                 new = {import_object(name): func for name, func in lazy.items()}
                 lookup.update(new)
                 # drop lazy implementations, idempotent for thread safety
                 lazy_lookup.pop(module, None)
```

No need to do this in this PR of course.

ibis/expr/datatypes/core.py
```diff
@@ -950,6 +951,37 @@ class INET(String):
...
+    np.dtype("str"): string,
+    np.dtype("datetime64"): timestamp,
+    np.dtype("datetime64[ns]"): timestamp,
+    np.dtype("timedelta64"): interval,
+    np.dtype("timedelta64[ns]"): Interval("ns"),
+}
+
+
+@dtype.register(np.dtype)
+def _(value):
```
kszucs, line 978: Do we have tests for these new overloads?

ibis/expr/datatypes/value.py
```diff
@@ -280,39 +171,151 @@ def __repr__(self):
...
+    np_dtype = value.dtype
+    if np_dtype.type == np.object_:
+        return dt.Array(_infer_object_array_dtype(value))
+    elif np_dtype.type == np.str_:
+        return dt.Array(dt.string)
+    return dt.Array(dt.dtype(np_dtype))
+
+
+@infer.register("pandas.Series")
+def _(value):
```
kszucs, line 225: Can you name the function similarly to the ones above?

ibis/expr/datatypes/value.py
```diff
@@ -280,39 +171,151 @@ def __repr__(self):
...
+def infer_shapely_linestring(value) -> dt.LineString:
+    return dt.linestring
+
+
+@infer.register("shapely.geometry.Polygon")
+def infer_shapely_polygon(value) -> dt.Polygon:
+    return dt.polygon
+
+
+@infer.register("shapely.geometry.MultiLineString")
```
kszucs, line 257: The usage here is really nice!

kszucs (approved):
Just a few nits, otherwise it looks good to me!

ibis/common/validators.py
jcrist: If I understand what you're asking for correctly, that would only delay the imports until the first time this validator was hit, regardless of type used.

As written, this implementation only imports the lazily loaded types when an object from that lazy module is provided. If the user never imports shapely, then ibis won't either.

kszucs:
Have you measured the impact on import times?

jcrist:
> Have you measured the impact on import times?

On my machine I see `import ibis` as ~20-30 ms faster.

That was only a part of the goal though. The goals of this PR were:

- Avoid importing shapely unless unnecessary, reducing import times
- Better avoid the shapely/duckdb segfault by making it so shapely doesn't have to be imported first when using ibis + duckdb.
- Add in infrastructure so other imports can be delayed, or existing delayed imports can be cleaner (may help with avoiding importing pandas in more situations, which will have a more noticeable impact).

ibis/common/validators.py
kszucs: No, I'd preserve the behaviour, so `lazy_instance_of("np.ndarray", [1, 2, 3])` would not load numpy either, hence the `lazy_load_types(types, module=get_module_name(obj))` module arg in my pseudocode above.

PR #4891, opened 2022-11-23
refactor(ir): remove ops.NodeList and ir.List in favor of builtin sequences

kszucs:


ibis/common/graph.py
```diff
@@ -101,6 +137,10 @@ def toposort(node: Traversable) -> Graph:
     return Graph(node).toposort()
 
 
+# def map(fn, node: Traversable) -> Iterator[Any]:
+#     return traverse(lambda x: (proceed, fn(x)), node)
```
cpcloud, line 141: Should this be deleted?

cpcloud (approved):
This seems like a solid refactor. Thanks!

ibis/common/graph.py
kszucs: Yes, removing it.

PR #4912, opened 2022-11-28
perf(expr): traverse nodes only once during compilation

cpcloud:
This PR addresses the performance issue in #4911. The issue is repeated traversal of the same node in an expression during the correlated ref check. For example, if two columns share the same child table that table will be traversed twice. Let's wait until #4911 is merged before merging this one.


kszucs (approved):
LGTM

Issue #4704, opened 2022-10-22
ux: problems with ibis' `_` convenience API for deferred attribute resolution


ogrisel:
#3804 introduced a nice way to be able chain several operations that use new columns not originally part of in input expression:

```python
import pandas as pd
import ibis
from ibis import _

table = ibis.memtable(pd.DataFrame({"x": list("12213")}))
(
   table.mutate(a=2 * _.x, b=_.x.cast("float64"))
   .group_by(_.a)
   .aggregate(count=_.b.count(), mean_b=_.b.mean())
   .order_by(_.mean_b)
).execute()
```

This makes it very easy to complex logic in functions that add or rename columns and the use the `.pipe` method to chain them in particular.

The choice of `_` as the name of the importable deferred attribute resolver is nice because it leads to quite concise and readable code. However I also find it problematic from an UX point of view for the following reasons:

- `a)` it's not easily googleable: someone not familiar with this idiom will have a hard time googling or stackoverflowing for it.
- `b)` it can conflict with the `_` variable of Jupyter notebooks (that store the value of the last executed cell) and can therefore lead to very confusing error messages for the unsuspecting users that do exploratory data analysis with Ibis in a Jupyter notebook with multiple cells.
- `c)` even in regular, non-interactive Python code it can conflict with the common idiom of assigning ignored function call results to a dummy `_` variable to express that we do not need a variable for an ancillary value. E.g.:

```python
a, _ = function_returning_a_pair_of_values()  # ignore the second value
```

```python
for _ in range(n_trials):
    if attempt():
        break
else:
    raise Exception(f"{n_trials} consecutive failed attempts")
```

## Possible solutions

- Solution for `a)` only could be to change the default name of the singleton while allowing to keep on using the `_` idiom later in the code:

```python
from ibis import deferred_resolver as _

# continue as previously
```

At least people reading this code for the first time will understand that `_` is some kind of deferred attribute resolver just by reading the code: no complex googling or random documentation scanning.

I don't have good solutions for  `b)` and `c)`. Here are ideas:

### Suggest (by convention) to name the resolver `c` instead of `_`

I picked up `c` for "column" because most of the time the attribute lookup matches a column look-up. However a variable named `c` might still frequently conflict with user code. Adapting the original example would be lead to code that looks like

```python
import pandas as pd
import ibis
from ibis import deferred_resolver as c


table = ibis.memtable(pd.DataFrame({"x": list("12213")}))
(
[...]

PR #4914, opened 2022-11-28
docs: created how to guide for deferred expressions

p-a-a-a-trick:


docs/how_to/chain-expressions.md
```diff
@@ -0,0 +1,66 @@
...
+To get started, we need to import `_` from ibis:
+
+```python
+import ibis
+from ibis import _
+
+import pandas as pd
+```
+
+Let's set up two tables using [ibis.MemTables](memtable-join.md) that were introduced in 3.2:
```
cpcloud, line 18:
```suggestion
Let's create two in-memory tables using [ibis.MemTables](memtable-join.md), an API introduced in 3.2:
```

docs/how_to/chain-expressions.md
```diff
@@ -0,0 +1,66 @@
...
+
+```python
+t1 = ibis.memtable(pd.DataFrame({'x': range(5), 'y': list('ab')*2 + list('e')}))
+t2 = ibis.memtable(pd.DataFrame({'x': range(10), 'z': list(reversed(list('ab')*2 + list('e')))*2}))
+```
+
+## Creating ColumnExpressions
+
+We can use `_` to create ColumnExpressions without explicitly referencing TableExpression objects:
```
cpcloud, line 27:
```suggestion
We can use `_` to create new column expressions without explicit reference to the previous table expression:
```

docs/how_to/chain-expressions.md
```diff
@@ -0,0 +1,66 @@
...
+
+```python
+# We can pass a deferred expression into a function:
+def modf(t): return t.x % 3
```
cpcloud, line 31:
```suggestion
def modf(t):
    return t.x % 3
    
```

cpcloud (commented):
Couple of comments, but otherwise looks great!

docs/how_to/chain-expressions.md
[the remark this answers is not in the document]
p-a-a-a-trick: Fixed; thanks @cpcloud

docs/how_to/chain-expressions.md
```diff
@@ -0,0 +1,68 @@
+# How to Chain Ibis Expressions using the Deferred Expression API (Also known as the Underscore (`_`) API)
```
cpcloud, line 1: Let's make the title as concise as possible. How about this?

```suggestion
# Chaining Expressions with `_`
```

p-a-a-a-trick: I think I wanted to avoid just using `_` in the title for SEO purposes per #4704: people probably won't know to call it _ when wanting to chain expressions and I don't know how the _ character would affect searches, but that is basically pure speculation on my part.

What do you think? I do like the more concise title, mostly because the sidebar looks really cluttered right now, but also because I'm not a fan of really long titles.  I'm just worried about how people will _find_ this info.

p-a-a-a-trick: I guess the key is "chain Ibis expressions", and the "with _" will tell them "you can do it using this thing".  I also call out the actual name of the object in the text, so I'm going to go ahead and accept this one.

Thanks @cpcloud !

cpcloud: Yeah, it could also be `Chaining Expressions with Underscore`.

docs/how_to/chain-expressions.md
```diff
@@ -0,0 +1,68 @@
+# Chaining Expressions with Underscore
```
cpcloud, line 1: I ended up changing this to `Underscore` for readability.

cpcloud (approved):
Ship it!

PR #4915, opened 2022-11-28
refactor(ir): remove unnecessary complexity introduced by variadic annotation

kszucs:
Since the removal of `ops.NodeList` where we implicitly traverse tuples, the variadic annotation mostly provides syntax sugar. In addition to that `rlz.variadic` hasn't been released yet, so it's better to prefer a simpler approach to reconstruct nodes using purely keyword arguments rather than tracking both `*args` and `**kwargs`.

depends on #4891


cpcloud (commented):
LGTM, great work!

cpcloud (approved):

PR #4916, opened 2022-11-28
fix(api): allow boolean scalars in predicate APIs

cpcloud:
This PR fixes a bug where we currently do not allow bool literals in `Table.__getitem__` or  `Table.filter` APIs.


kszucs (approved):
Nice catch!

PR #4922, opened 2022-11-28
refactor(datafusion): align register API across backends

gforsyth:
BREAKING CHANGE: `register_csv` and `register_parquet` are removed.
Pass filename to `register` method instead.

This also adds in a general backend test for the `register` API so we
can make sure things are consistent, and fixes a small bug with the
dispatch in polars (and datafusion which I copied) that raises a
recursion error if you pass in garbage.


gforsyth:
I removed the argument and wording around passing `schema` for loading CSV files because datafusion doesn't accept `ibis.Schema` as an argument and I can't find any Python documentation online to point at in the docstring.

ibis/backends/tests/test_client.py
```diff
@@ -845,3 +845,67 @@ def test_has_operation_no_geo(con):
...
+        "pyspark",
+        "snowflake",
+        "sqlite",
+    ]
+)
+def test_register_garbage(con):
+    with pytest.raises(ValueError) as exc_info:
+        con.register("garbage_notafile")
+
+    assert "Unrecognized file type or extension" in str(exc_info.value)
```
cpcloud, line 911:
```suggestion
    with pytest.raises(ValueError, match="Unrecognized file type or extension"):
        con.register("garbage_notafile")
```

cpcloud (approved):
LGTM, small and optional nit. Feel free to merge!

ibis/backends/tests/test_client.py
gforsyth: oh, that's much nicer

PR #4935, opened 2022-11-30
test: initial support for property-based testing using hypothesis

kszucs:
Still need to configure a CI build for it.


ibis/tests/expr/test_schema.py
```diff
@@ -134,6 +136,14 @@ def test_whole_schema():
     assert customers.schema() == expected
 
 
+def test_schema_names_and_types_legth_must_match():
```
cpcloud, line 139:
```suggestion
def test_schema_names_and_types_length_must_match():
```

ibis/tests/strategies.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+)
+
+
+# Strategies for generating schema
+
+
+@st.composite
+def schema(draw, item_strategy=primitive_dtypes, max_size=20):
+    num_fields = draw(st.integers(min_value=0, max_value=max_size))
```
cpcloud, line 165: Should we still keep an empty schema as a valid construction?

kszucs: If an empty table is valid then an empty schema should be valid too.

cpcloud: How about raising an exception on empty schema construction and disallowing them altogether?

kszucs: I wouldn't disallow empty schema construction, it is also allowed by pyarrow.

kszucs:
Additional work to do, preferably in follow-up PRs:
 - configure a CI job running hypothesis builds in a reproducible fashion
 - utilize the hypothesis strategies for more test cases

@cpcloud can we merge this and handle the rest in follow-ups?

cpcloud:
@kszucs Yeah, that's fine. Can you put an upper bound on the data generation for the test that's failing? The others that passed are likely just barely doing so.

kszucs:
Limited to generate memtables with 5 columns at most.

cpcloud (approved):

PR #4943, opened 2022-12-01
refactor(impala): move impala SQL tests to snapshots

cpcloud:
This PR moves the remaining impala SQL string tests to snapshot-based tests.


jcrist (approved):
:shipit:

PR #4944, opened 2022-12-01
refactor(api): move `histogram` method to `Table` expressions

cpcloud:
This PR moves the histogram method to table expressions in an effort to further simplify the compiler and avoid having expressions or their construction tied to the internal details of the compiler.

### DIFF
--- a/ibis/expr/types/numeric.py
+++ b/ibis/expr/types/numeric.py
@@ -526,38 +526,6 @@ class NumericColumn(Column, NumericValue):
             include_over=include_over,
         ).to_expr()
 
-    def histogram(
-        self,
-        nbins: int | None = None,
-        binwidth: float | None = None,
-        base: float | None = None,
-        closed: Literal["left", "right"] = "left",
-        aux_hash: str | None = None,
-    ) -> ir.CategoryColumn:
-        """Compute a histogram with fixed width bins.
-
-        Parameters
-        ----------
-        nbins
-            If supplied, will be used to compute the binwidth
-        binwidth
-            If not supplied, computed from the data (actual max and min values)
-        base
-            Histogram base
-        closed
-            Which side of each interval is closed
-        aux_hash
-            Auxiliary hash value to add to bucket names
-
-        Returns
-        -------
-        CategoryColumn
-            Coded value expression
-        """
-        return ops.Histogram(
-            self, nbins, binwidth, base, closed=closed, aux_hash=aux_hash
-        ).to_expr()
-
     def summary(
         self,
         exact_nunique: bool = False,
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -1334,6 +1334,79 @@ class Table(Expr, JupyterMixin):
         )
         return op.to_expr()
 
+    def histogram(
+        self,
+        column: str,
+        nbins: int | None = None,
+        binwidth: float | None = None,
+        base: float | None = None,
+        suffix: str | None = None,
+        eps: float = 1e-13,
+    ):
+        """Compute a histogram with fixed width bins.
+
+        Parameters
+        ----------
+        column
+            Column to bucket
+        nbins
+            If supplied, will be used to compute the binwidth
+        binwidth
+            If not supplied, computed from the data (actual max and min values)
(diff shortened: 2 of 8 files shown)


ibis/expr/types/numeric.py
```diff
@@ -526,38 +526,6 @@ def bucket(
...
-            Which side of each interval is closed
-        aux_hash
-            Auxiliary hash value to add to bucket names
-
-        Returns
-        -------
-        CategoryColumn
-            Coded value expression
-        """
-        return ops.Histogram(
```
jcrist, line 557: AFAICT  you can delete the `Histogram` op as well, since this was the only usage.

ibis/expr/types/numeric.py
```diff
@@ -526,38 +526,6 @@ def bucket(
             include_over=include_over,
         ).to_expr()
 
-    def histogram(
```
jcrist, line 529: This is a breaking change since it deletes a method, not a refactor.

- Are we fine with it being a breaking change? (I think it's fine, it's a rare method and this is much cleaner)
- Can you update the message so it shows up as a breaking change in the release notes?

ibis/expr/types/relations.py
```diff
@@ -1334,6 +1334,66 @@ def sql(self, query: str) -> ir.Table:
...
+        aux_hash
+            Auxiliary hash value to add to bucket names
+        eps
+            Allowed floating point epsilon for histogram base
+
+        Returns
+        -------
+        Table
+            New table expression with a bucketed column expression
+        """
```
jcrist, line 1367: Would you mind expanding the docstring here a bit? Questions I have reading this:

- I can't picture the output schema of this operation. Maybe an example would help clarify?
- What does "Histogram base" mean?

I recognize that this was just copy-pasted from the column method, so if you don't want to tackle this now that's fine.

ibis/expr/types/relations.py
```diff
@@ -1334,6 +1334,66 @@ def sql(self, query: str) -> ir.Table:
...
+        ----------
+        column
+            Column to bucket
+        nbins
+            If supplied, will be used to compute the binwidth
+        binwidth
+            If not supplied, computed from the data (actual max and min values)
+        base
+            Histogram base
+        aux_hash
```
jcrist, line 1358: Since this is a breaking change moving this method, perhaps we should also take the chance to rename this keyword? This is just a column name suffix, it doesn't actually have to be a hash. Perhaps `suffix`, `col_suffix`, `bin_suffix` (I like `suffix` best, since it's similar to `suffixes` in `join`)?

ibis/expr/types/numeric.py
cpcloud: Is this not showing up somewhere? I see it in the commit body.

ibis/expr/types/relations.py
cpcloud: Yeah, `suffix` sounds good to me

ibis/expr/types/numeric.py
jcrist: Oh, I thought the breaking change thing required something in the commit title too, not just a breaking change in the body. NVM, this is fine.

ibis/expr/types/numeric.py
cpcloud: Yep thanks!

ibis/expr/types/relations.py
cpcloud: OTOH, do we even need this level of optionality here?

ibis/expr/types/numeric.py
cpcloud: Done!

ibis/expr/types/relations.py
cpcloud: Done, thanks for calling this out!

ibis/expr/types/numeric.py
kszucs: Shouldn't we have `table.histogram()` implementation here instead? It would remain backward compatible.

cpcloud:
Closing this PR, I think there's a backwards compatible way we can do this.

PR #4957, opened 2022-12-05
docs(pandas-udf-examples): UDF Examples for Pandas have the incorrect import path.

abelsonlive:
Documentation for UDFs in Python instruct users to import the `udf` class via the following statement:

```python
from ibis.backends.pandas import udf
```

It should actually be:

```python
from ibis.backends.pandas.udf import udf
```


cpcloud (approved):
Awesome! Thank you.

PR #4963, opened 2022-12-06
refactor: reimplement `histogram` API in terms of window functions

cpcloud:
This PR rewrites the current histogram API to use window functions instead of a cross join, in an effort to further simplify the compiler and avoid having expressions or their construction tied to the internal details of the compiler. Depends on #4962.


kszucs:
Mering, thanks @cpcloud!

PR #4968, opened 2022-12-06
fix(clickhouse): use clickhouse's dbapi

jcrist:
This fixes two issues with the clickhouse backend:

- `raw_sql` returns a nonstandard object (a tuple of results and other info), rather than a `dbapi` cursor object.
- because of this, the generic implementation of `to_pyarrow_batches` doesn't work.

However, clickhouse's `dbapi` starts a new connection for every `cursor`, which adds some overhead. To remedy this, we:

- Continue to use a single `Client` for all ibis-internal calls
- Implement `raw_sql` using the `dbapi`, which creates a new connection for every cursor
- Implement `_cursor_batches` manually calling `raw_sql` instead of `_safe_raw_sql` since `_safe_raw_sql` continues to use the old `client.execute` code path (which shares a connection).

All of this is gross, but functional. If people have suggestions for a cleaner approach I'm happy to change things, but this PR is strictly more functional than what we had before so I'm also fine merging as is.

On top of #4967.

### DIFF
--- a/ibis/backends/clickhouse/__init__.py
+++ b/ibis/backends/clickhouse/__init__.py
@@ -1,14 +1,16 @@
 from __future__ import annotations
 
+import contextlib
 import json
-from typing import TYPE_CHECKING, Any, Literal, Mapping
+from typing import TYPE_CHECKING, Any, Iterable, Literal, Mapping
 
+import clickhouse_driver
 import toolz
-from clickhouse_driver.client import Client as _DriverClient
 
 import ibis
 import ibis.config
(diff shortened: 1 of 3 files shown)


cpcloud (approved):
This seems like an acceptable amount of code to get the value that the pyarrow conversions bring!

cpcloud:
Looks like a rebase is needed + a small return type docstring in `raw_sql`: https://github.com/ibis-project/ibis/actions/runs/3634206514/jobs/6132065145#step:5:674

PR #4975, opened 2022-12-07
ci: remove `python-dev`

cpcloud:
This PR removes the no longer extant `python-dev` package from our CI setup. I am not sure if we still need it, buf if we do, apaprently it is merged into the vanilla `python` package.


cpcloud:
Merging, so that upstream can start passing again.

PR #4980, opened 2022-12-08
feat(trino): add trino backend

cpcloud:
This PR adds an extremely bare bones trino backend. ~`testconf.py` is definitely a stub and unlikely to work as it is.~

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -49,38 +49,60 @@ jobs:
         backend:
           - name: dask
             title: Dask
+            extras:
+              - dask
           - name: duckdb
             title: DuckDB
+            extras:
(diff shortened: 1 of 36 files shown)


gforsyth:
Hilariously, we can't ingest a CSV into `trino` without a hive metastore, so we might steal this in order to create the most complicated local filesystem we've ever seen: https://github.com/starburstdata/dbt-trino/blob/master/docker-compose-trino.yml

gforsyth (approved):
Not bad for 24 hours of work.  I think this is good to go as a first cut.  Obviously we'll round out the ops and add a page in the docs when things are a touch more polished.

cpcloud:
I'm happy with this. I'l squash and then merge!

PR #4984, opened 2022-12-09
test: rename files that may conflict with case-insensitive filesystems

### LINKED ISSUE #4983: Distinguish `M-TIMESTAMP` and `m-TIMESTAMP`
Is it possible to name https://github.com/ibis-project/ibis/blob/master/ibis/backends/bigquery/tests/unit/snapshots/test_compiler/test_temporal_truncate/M-TIMESTAMP/out.sql or https://github.com/ibis-project/ibis/blob/master/ibis/backends/bigquery/tests/unit/snapshots/test_compiler/test_temporal_truncate/m-TIMESTAMP/out.sql differently? On a case-insensitive system (like a Mac), only one file can exist, making it impossible to pull code without conflicts.

I'm aware I can [create a case-sensitive virtual drive](https://stackoverflow.com/a/26185365/1093967), but I wonder if it's OK to solve this in a friendlier way? Right now, I'm pulling on an Ubuntu Codespace when I need to sync latest.

cpcloud:
Fixes an issue where code cannot be checked out due to case insensitive filesystems.

Closes #4983.


cpcloud:
I'm going to remove `pytest-snapshot` in favor of `syrupy` which has 1) a single file per test function (even if the test is parameterized), and 2) more convenient syntax.

cpcloud:
Picked this up again due to syrupy not supporting pytest-xdist.

Merging.

PR #4988, opened 2022-12-09
feat: add extract functionality to trino backend

cpcloud:
This PR adds generic extract rules so that they can be shared between the sqlalchemy backends where possible. This fixes failing benchmark tests on master.


gforsyth (approved):
Nice!  :goberserk:

PR #5003, opened 2022-12-11
refactor: split up custom nix code; remove unused derivations

cpcloud:
This PR splits up some of our code in `flake.nix` to a few distinct files under `nix/`.

The upshot here is that we will spend much less time downloading/building dependencies that are never used for certain tasks, for example, building pyarrow before running `pre-commit`.

The new shells `.#release`, `.#docs`, etc only contain the dependencies needed to perform the specific task.

This PR is in draft mode until https://github.com/nix-community/poetry2nix/pull/867 is merged.


cpcloud:
Merging on green!

Issue #4906, opened 2022-11-27
feat: Read Multiple Files DuckDB


jayceslesar:
### Is your feature request related to a problem?

I can take a look at this but seems like the _generate_view_code Regex Dispatcher might need to be robustified to be able to take a list of regexes, assert that they are all the same and provide the same functionality. Alternatively we could assume that the elements in the list all follow the same general pattern and will point to the same place but I don't really like that.

This is definitely a nice to have as I wouldn't think there to be a speed difference here? But would be really interesting if there was speed difference (multiple connections vs one vs maybe even threading the connection to each file downstream)

### Describe the solution you'd like

We can see [here](https://duckdb.org/docs/data/parquet) that DuckDB can take in a list of exact paths, which seems much cleaner than 

```py
import ibis

con = ibis.duckdb.connect()

# partitions is a list of my files
t = con.register(partitions[0])
for path in partitions[1:]:
    t = t.union(con.register(path))
```

Would be nice to just:

```py
import ibis

con = ibis.duckdb.connect()

# partitions is a list of my files
t = con.register(partitions)
```

### What version of ibis are you running?

Latest Master

### What backend(s) are you using, if any?

DuckDB

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
Thanks @jayceslesar, yeah this would be really nice to have.

cpcloud:
This was addressed in 4.1:

```
…/ibis on  deferred-docs is 📦 v4.1.0 via 🐍 v3.10.9 via ❄️  impure (ibis-3.10.9-env)
❯ ipython

In [1]: import ibis

In [2]: t = ibis.read_parquet('ci/ibis-testing-data/parquet/functional_alltypes/*.parq')

In [3]: t.head().execute()
Out[3]:
     id  bool_col  tinyint_col  smallint_col  int_col  bigint_col  float_col  double_col date_string_col string_col           timestamp_col  year  month
0  5770      True            0             0        0           0        0.0         0.0     b'08/01/10'       b'0' 2010-08-01 00:00:00.000  2010      8
1  5771     False            1             1        1          10        1.1        10.1     b'08/01/10'       b'1' 2010-08-01 00:01:00.000  2010      8
2  5772      True            2             2        2          20        2.2        20.2     b'08/01/10'       b'2' 2010-08-01 00:02:00.100  2010      8
[...]

PR #5005, opened 2022-12-12
refactor: rework registration / file loading

gforsyth:
Introduces explicit `read_` methods for various types of file /
in-memory object loading.

Removes the `RegexDispatcher` from the DuckDB, polars, and datafusion backends

Should be backwards compatible with existing register functionality,
plus users can now pass in iterables of CSV or parquet files (duckdb only), and also
make use of the explicit `read_` methods if they choose (say if filenames
are non-standard)

Somehow, Datafusion can't handle a `count` so I've added a gross
conditional to the tests for the moment.


jayceslesar:
This will close #4906

This is definitely the best way to go about it, regex dispatcher was an issue when I was looking at this

ibis/backends/duckdb/__init__.py
```diff
@@ -261,30 +172,176 @@ def register(
...
+
+            quoted_table_name = _quote(table_name)
+            sql = f"""CREATE OR REPLACE VIEW {quoted_table_name} AS
+            SELECT * FROM read_parquet({dataset})"""
+
+            self.con.execute(sql)
+
+        return self._read(table_name)
+
+    def read_pandas_or_arrow(self, dataframe, table_name=None):
```
cpcloud, line 308: Can we either:

1. split this into `read_pandas` and `read_arrow`
2. call this `read_in_memory`/`read_memory`/`read_mem`

ibis/backends/duckdb/__init__.py
```diff
@@ -25,123 +26,33 @@
...
 
-_generate_view_code = RegexDispatcher("_register")
 _dialect = sa.dialects.postgresql.dialect()
 
-_gen_table_names = (f"registered_table{i:d}" for i in itertools.count())
+# counters for in-memory, parquet, and csv reads
+# used if no table name is specified
+pdN = itertools.count(0)
+paN = itertools.count(0)
+csvN = itertools.count(0)
```
cpcloud, line 36: Can we suffix these with `_n` instead of `N`?

ibis/backends/datafusion/__init__.py
```diff
@@ -24,40 +24,10 @@
...
-            f"""Unrecognized file type or extension: {raw}
-
-        Valid prefixes are parquet://, csv://, or file://
-        Supported file extensions are parquet and csv"""
-        )
-    return _register_file(f"{extension}://{path}", table_name=table_name, **kwargs)
+# counters for in-memory, parquet, and csv reads
+# used if no table name is specified
+paN = itertools.count(0)
+csvN = itertools.count(0)
```
cpcloud, line 30: Can we suffix these with `_n` instead of `N`?

ibis/backends/duckdb/__init__.py
```diff
@@ -261,30 +172,176 @@ def register(
...
                 # by the time we execute against this so we register it
                 # explicitly.
-                cursor.cursor.c.register(table_name, dataset)
-        elif isinstance(source, (str, Path)):
-            sql, table_name, extensions_required = _generate_view_code(
-                str(source), table_name=table_name, **kwargs
-            )
-            self._load_extensions(extensions_required)
-            self.con.execute(sql)
+                con.connection.c.register(table_name, dataset)
```
cpcloud, line 291: You *might* be able to get rid of the `.c` entirely. I believe SQLAlchemy creates a somewhat-faithful proxy for the underlying dbapi connection.

ibis/backends/duckdb/__init__.py
```diff
@@ -261,30 +172,176 @@ def register(
...
+            )
+        self._load_extensions(["postgres_scanner"])
+        quoted_table_name = _quote(table_name)
+        sql = (
+            f"CREATE OR REPLACE VIEW {quoted_table_name} AS "
+            f"SELECT * FROM postgres_scan_pushdown('{uri}', 'public', '{table_name}')"
+        )
+        self.con.execute(sql)
+
+        return self._read(table_name)
```
cpcloud, line 342: I guess we're not testing any of this code. Seems like we could spin the `postgres` service up in CI for the duckdb backend to hit this code. Thoughts?

ibis/backends/tests/test_register.py
```diff
@@ -0,0 +1,321 @@
...
+        "snowflake",
+        "sqlite",
+        "trino",
+    ]
+)
+def test_register_iterator_parquet(
+    tmp_path,
+    data_directory,
+):
+    pq = pytest.importorskip("pyarrow.parquet")
```
cpcloud, line 209: Hm, any idea why this code isn't hit? Since this line isn't ever hit, that means it's not the import failing, but something else that is causing this test to never execute.

cpcloud (commented):
Looking good! A few questions about coverage/testing.

ibis/backends/duckdb/__init__.py
gforsyth: Yeah, the old test was just that the appropriate `CREATE VIEW` code was called -- I can add another(?) postgres service for this

ibis/backends/tests/test_register.py
gforsyth: I forgot to add a `backend` or `con` fixture -- nice catch!

ibis/backends/duckdb/__init__.py
gforsyth: yep, that worked!

ibis/backends/duckdb/__init__.py
cpcloud: I don't think you need another service, the existing one is fine, you can add a `services:` array to the `ibis-backends.yml` workflow. There should be some examples in the area.

ibis/backends/datafusion/__init__.py
```diff
@@ -151,13 +121,86 @@ def register(
...
+
+        Returns
+        -------
+        ir.Table
+            The just-registered table
+        """
+        path = normalize_filename(path)
+        table_name = table_name or f"ibis_read_csv_{next(csv_n)}"
+        # Our other backends support overwriting views / tables when reregistering
+        self._context.deregister_table(table_name)
```
gforsyth, line 174: I feel like this would be _way_ too aggressive for a non-file-based backend but I think this is ok for Datafusion?

ibis/backends/duckdb/__init__.py
gforsyth: Added a test for this, but after a few rounds of CI shenanigans, it's marked with a `skipif(True)`

ibis/backends/datafusion/__init__.py
cpcloud: @kszucs Thoughts on this?

cpcloud: I think this is fine.

cpcloud (approved):
:shipit:

PR #5008, opened 2022-12-12
fix: `ibis.connect` always returns a backend

jcrist:
Previously `ibis.connect` would return a `Table` object when calling `connect` on a parquet/csv file. This now returns a backend containing a single table created from that file. This fixes an inconsistency with the current interface, where the return type depended on the value of the connection string, making it hard to type check. When possible users may use `ibis.read` instead to read files into ibis tables.

I went with a clean break here rather than a deprecation because:

- This feature hasn't been out that long anyway, I doubt a breaking change now will affect that many users
- 4.0 has other breaking changes

Happy to move to a deprecation if needed.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -763,10 +763,6 @@ def connect(resource: Path | str, **kwargs: Any) -> BaseBackend:
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
Yeah, I think the consistency here is a good thing.

PR #5021, opened 2022-12-17
build(perf): use wheels with nix

cpcloud:
This is ~an experiment~ to see if we can setup the nix environments with wheels only, including pyarrow, which allows us to avoid having to build anything from source as long as there is an appropriate binary wheel. There are a few PRs to poetry2nix that need to be merged for this to work which is why it is in a draft state.

Conclusion: it works!


cpcloud:
Merging on green! The upstream work in `poetry2nix` has been been merged. We'll be wheeling hopefully from now on, which should help us avoid wasting time on expensive builds.

gforsyth:
:rocket:

PR #5024, opened 2022-12-18
Avoid roundtripping to expression for IFF

krzysztof-kwitt:


cpcloud (approved):
Awesome, thanks for the PR!

PR #5026, opened 2022-12-18
Simplify ops.ArrayColumn in Postgres backend

krzysztof-kwitt:
Hello,

We have an auxiliary function that will generate the full code of this function, so we can get rid of specialized implementation.

In addition, we gain support for backends that do not support reduction filter syntax.
https://github.com/ibis-project/ibis/blob/40ef35ee974f79d6132cc44c089fc430b17283a5/ibis/backends/base/sql/alchemy/translator.py#L79-L87

Best regards


cpcloud (approved):
Sweet! Thanks for the PR. I can help fix the commit message so your change shows up in the release notes!

PR #1521, opened 2018-07-06
ENH/CLN: Fix alias generation to be clearer

cpcloud:
Closes #1460

Makes it clear what kind of aliases we're generating from unbound
tables.

PR #5030, opened 2022-12-18
chore(deps): bump duckdb lower bound

cpcloud:
This PR bumps the minimum duckdb version to 0.3.3 so we can remove a dependency and some old compatibility code.


cpcloud:
Merging this after #1521.

PR #5034, opened 2022-12-19
ci: reduce pyspark flakiness by running tests serially

cpcloud:
It seems like running pyspark tests in parallel in CI is flaky, so this PR reverts back to serial tests for pyspark.


jcrist (approved):
Thanks for the fix. I noticed the recent CI pyspark flakiness, but didn't know that anything in CI had changed. :+1: from me.

PR #5036, opened 2022-12-19
ci: re-enable release notes in `simulate_release`

cpcloud:
When this was ported to use `nix develop` we had to remove `--pure`,
since we removed `nix-shell` usage here, and the `GITHUB_ACTIONS`
variable used by `semantic-release` crept in. We `unset` it so that it
has no effect.


cpcloud:
This has the desired effect (release notes showing up in the output of `simulate_release`) and the trino CI failure is unrelated.

PR #5038, opened 2022-12-19
ci: increase GH API rate limit for nix updates and don't `fail-fast`

cpcloud:
This PR gives the nix updates workflow a higher GH API rate limit, which typically is not necessary, but a recent failure indicate an API limit likely due to a bunch of other concurrent jobs and workflows also hitting the GH API at the same time. Since it is easy to increase the limit we might as well do it.


cpcloud:
Merging, since this can't really be tested until it is.

PR #5052, opened 2022-12-20
chore(deps): constrain pyarrow via `snowflake-sqlalchemy` by adding the `pandas` extra

cpcloud:
This PR prevents an incompatible version of pyarrow being installed when using the snowflake backend. As a bonus, we no longer need to hack it pyarrow@8 in the Snowflake CI setup.


cpcloud:
There's a PR to update `snowflake`'s python API to use arrow 10 (https://github.com/snowflakedb/snowflake-connector-python/pull/1336).

cpcloud:
We'll pick this up automatically when they do the bump and cut a new release.

cpcloud:
Without this constraint, segfaults abound if you just `pip install pyarrow` without a version constraint.

pyproject.toml
```diff
@@ -65,9 +65,14 @@ pyspark = { version = ">=3,<4", optional = true }
 regex = { version = ">=2021.7.6", optional = true }
 requests = { version = ">=2,<3", optional = true }
 Shapely = { version = ">=1.6,<1.8|>=1.9,<2", optional = true }
+snowflake-connector-python = { version = ">=2.7.10,<3", optional = true, extras = [
+  "pandas"
+] }
+snowflake-sqlalchemy = { version = ">=1.4.1,<2", optional = true, extras = [
+  "pandas"
+] }
```
gforsyth, line 73: Can you add a comment about why we have to do this crazy business?

cpcloud: Which part? Adding the transitive dependency, or something else?

cpcloud: Done!

gforsyth (approved):

PR #5056, opened 2022-12-20
fix: support `arbitrary` with no arguments

jcrist:
Previously the default value for `how` was `None`, which was appears to have been intended to mean `"first"`. The test suite always specified `how` explicitly though, which meant most backends failed to supported the default `how=None`. This PR changes the default to `"first"`, and updates the test suite to include a no-args version of `arbitrary`.


cpcloud (approved):
LGTM! Thanks!

PR #5057, opened 2022-12-20
chore(dev): replace current flake8-based and other linting tools with `ruff`

cpcloud:
This PR replaces all of our flake8 checks and tools with [`ruff`](https://github.com/charliermarsh/ruff), which is Python linter implemented in Rust designed to be fast, lightweight and combine the most commonly used flake8 plugins into a single tool. While I do not care too much about the performance for ibis developement (though I do for my LSP configuration!) I do care about tool consolidation, which is the main purpose of this PR.


ibis/expr/api.py
```diff
@@ -26,77 +26,13 @@
 from ibis.expr.deferred import Deferred
 from ibis.expr.schema import Schema
 from ibis.expr.sql import parse_sql, show_sql, to_sql
-from ibis.expr.types import (  # noqa: F401
-    ArrayColumn,
```
kszucs, line 30: This looks like a breaking change. Are you sure that we should remove these?

.pre-commit-config.yaml
```diff
@@ -25,26 +25,16 @@ repos:
...
-      - id: flake8
-        exclude: (^ibis/tests/sql/snapshots/|_py310\.py$)
+      - id: ruff
+        # makes sure to exclude a file even if it's passed in explicitly
+        args: ["--force-exclude"]
   - repo: https://github.com/pycqa/docformatter
     rev: v1.5.1
     hooks:
       - id: docformatter
-  - repo: https://github.com/MarcoGorelli/absolufy-imports
```
kszucs, line 37: Is ruff handling it?

kszucs:
@cpcloud can you add a more descriptive name to the PR? :)

ibis/expr/api.py
cpcloud: It is, but these types are stable under `ibis.expr.types as ir`. I'd rather have just one place to access them than a hodgepodge list of things that doesn't actually include everything and has to be manually maintained.

.pre-commit-config.yaml
cpcloud: It's handling checking for new relative imports using the [`"TID"`](https://github.com/charliermarsh/ruff#flake8-tidy-imports-tid) code.

cpcloud:
> @cpcloud can you add a more descriptive name to the PR? :)

Done!

ibis/expr/api.py
kszucs: Can you separate this into another commit including the breaking change note?

kszucs (approved):
After including the breaking change note it looks good to me!

ibis/expr/api.py
cpcloud: Will do!

cpcloud:
I'll squash into a single commit.

ibis/expr/api.py
cpcloud: Done.

jcrist:
I have `flake8` configured with my editor in a way that I don't really want to change right now. If anyone else is in this situation, you can:

- readd the deleted `setup.cfg` to your local dev repo
- append `setup.cfg` to `.git/info/exclude` to avoid ever committing that file to the repo, without having to modify your local or global `.gitignore`.

PR #5058, opened 2022-12-20
chore: gate pyarrow imports in DuckDB backend

gforsyth:
I don't think pyarrow is a strict dependency of DuckDB and we have the machinery to make it friendlier.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -9,8 +9,6 @@ import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Iterator, Mapping, MutableMapping
 
-import pyarrow as pa
-import pyarrow.types as pat
 import sqlalchemy as sa
 import toolz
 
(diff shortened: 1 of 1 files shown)


jcrist:
This makes sense to me. I think we'll also want to remove `pyarrow` from the duckdb extra here (https://github.com/ibis-project/ibis/blob/45156f5ba03d8ff69e8a002364a44b15d4302bc7/pyproject.toml#L131).

cpcloud:
I don't think we should remove the extra, otherwise when you run `pip install ibis-framework[duckdb]` you get an install that doesn't have enough for all the public duckdb backend APIs to work.

cpcloud:
Is the idea to fail at runtime when someone tries to run `to_pyarrow()`?

cpcloud:
Hm, maybe it's not a big deal to remove it. How are we going to test the `to_pyarrow()` functionality in CI for the duckdb backend?

cpcloud:
Ah, we have `pyarrow` as a test dependency for other reasons, so this should "just work".

gforsyth:
I forgot about the custom record batch reader and I'm not sure this is worth it anymore.  I was mostly trying to fix an annoying dependency issue with ibis-substrait, but I can handle it over there instead.

PR #5061, opened 2022-12-21
chore(dev): remove codecov PR comments

cpcloud:
The numbers in the comment are highly variable in their accuracy given
the coverage that runs on the cloud backends, which makes the PR
comments pretty useless.


kszucs (approved):
Thanks!

PR #5067, opened 2022-12-22
core(dev-deps): replace `docformatter` with `ruff`'s `pydocstyle` support

cpcloud:
This PR removes `docformatter` from our development dependencies in favor of ruff's pydocstyle support. The fixups found by ruff are included here as well. Most docstrings didn't need to change.


pyproject.toml
```diff
@@ -318,10 +318,34 @@ select = [
...
+  "D102", # public method
+  "D103", # public function
+  "D105", # magic methods
+  "D107", # init
+  "D202", # blank lines after function docstring
+  "D203", # blank line before class docstring
+  "D205", # blank line between summary line and description
+  "D213", # Multi-line docstring summary should start at the second line
+  "D402", # First line should not be the function's signature
+  "D417", # Missing argument descriptions
```
kszucs, line 340: I assume we can incrementally remove these skipped rules.

kszucs (approved):
Nice!

pyproject.toml
cpcloud: Perhaps, but there a ton of warnings, so it'll quite some time :)

cpcloud:
Lemme squash

PR #5077, opened 2022-12-24
fix: don't fuse filters & selections that contain window functions

jcrist:
Since window functions act on multiple rows, fusing a `filter` operation with a prior selection that contained a window function can result in incorrect results.

This came up when doing today's [Hanukkah of Data](https://hanukkah.bluebird.sh/5783/) puzzle.


cpcloud (approved):
Nice find! Thank you.

PR #5082, opened 2022-12-25
fix: fix bug when disabling `show_types` in interactive repr

jcrist:
Previously this option was accidentally ignored. This fixes the bug and adds a test.


cpcloud (approved):
Sweet!

PR #5084, opened 2022-12-26
feat(trino): add struct operations

cpcloud:
This PR adds support for struct field access and constructing structs to the Trino backend.

### DIFF
--- a/ibis/backends/trino/datatypes.py
+++ b/ibis/backends/trino/datatypes.py
@@ -2,9 +2,11 @@ from __future__ import annotations
 
 import parsy as p
 import sqlalchemy as sa
-import trino
+from sqlalchemy.ext.compiler import compiles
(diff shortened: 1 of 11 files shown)


ibis/backends/trino/datatypes.py
```diff
@@ -133,3 +135,27 @@ def sa_trino_double(_, satype, nullable=True):
...
+
+
+@dt.dtype.register(TrinoDialect, JSON)
+def sa_jsonb(_, satype, nullable=True):
+    return dt.JSON(nullable=nullable)
+
+
+@compiles(sa.TEXT, "trino")
+def compiles_text(element, compiler, **kw):
+    return "VARCHAR"
```
cpcloud, line 153: This is definitely hit in local testing, as the struct tests fail without it.

PR #5087, opened 2022-12-27
fix(mssql): use double-dollar sign to prevent from interpolating a value

krzysztof-kwitt:
Hello,

We have a password set as an environment variable in the container, so we need to use a double-dollar sign. Otherwise, the password is read from the host.

Best regards,
Kamil Breguła


cpcloud:
Looks like a small typo in the commit message, should be `fix(mssql):` instead of `fix(mssqll):`. This might've been caused by a GitHub hiccup, I was just seeing strange things when reviewing your other PR.

krzysztof-kwitt:
@cpcloud Fixed.

cpcloud (approved):
Thank you!

PR #5091, opened 2022-12-28
feat(trino): add array slicing support

cpcloud:
This PR adds support for array slicing to the Trino backend, and fixes a bug in negative slicing in the postgres and duckdb backends.

### DIFF
--- a/ibis/backends/postgres/registry.py
+++ b/ibis/backends/postgres/registry.py
@@ -313,13 +313,6 @@ def _regex_extract(t, op):
     return sa.case([(does_match, matches[index])], else_=None)
 
 
-def _cardinality(array):
-    return sa.case(
-        [(array.is_(None), None)],
-        else_=sa.func.coalesce(sa.func.array_length(array, 1), 0),
-    )
-
-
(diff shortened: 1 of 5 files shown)


ibis/backends/postgres/registry.py
```diff
@@ -404,36 +397,40 @@ def _mod(t, op):
...
-        else_=idx,
-    )
+    return sa.case([(idx < 0, sa.func.cardinality(array) + idx)], else_=idx)
 
 
-def _array_slice(t, op):
-    arg, start, stop = op.args
-    sa_arg = t.translate(arg)
-    sa_start = t.translate(start)
+def _array_slice(*, index_converter, array_length):
```
kszucs, line 403: Why not to use `_neg_idx_to_pos` here directly?

kszucs (approved):
Just a nit, other than that it looks good to me.

ibis/backends/postgres/registry.py
cpcloud: Just so that the code can be shared with the duckdb implementation.

PR #5094, opened 2022-12-28
feat/fix(mssql): fix and implement some operations

krzysztof-kwitt:
Hello,

In PR: https://github.com/ibis-project/ibis/pull/5089/files I implemented `ops.TimeFromHMS` and  `ops.TimestampFromUNIX:` but when I looked at it again in the morning I saw that it wasn't working.

I also added `ops.TimestampTruncate` and `ops.DateTruncate`.

Now I want to add dot_sql support. For now I found that the metadata about the query can be fetched like below:
```python
result = bind.execute(
    f"DECLARE @query nvarchar(max) = 'SELECT 1 as tmp'; "
    f"EXEC sp_describe_first_result_set @query, null, 0; "
)
```
Best regards,

### DIFF
--- a/ibis/backends/mssql/registry.py
+++ b/ibis/backends/mssql/registry.py
@@ -165,7 +165,7 @@ operation_registry.update(
             6,
         ),
         ops.TimeFromHMS: fixed_arity(
-            lambda h, m, s: sa.func.timefromparts(h, m, s, 0), 3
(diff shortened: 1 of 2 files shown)


cpcloud:
Sweet!

Looks like you need to run `black` and also lowercase the first character after the `:` in the commit message.

After that this PR is good to merge.

krzysztof-kwitt:
@cpcloud Done

cpcloud (approved):

PR #5095, opened 2022-12-29
refactor: clean up timezone support in ops.TimestampFromYMDHMS

krzysztof-kwitt:
ops.TimestampFromYMDHMS doesn't have timezone field, so these checks are redundant.
https://github.com/ibis-project/ibis/blob/26ca1e4f3160de594c8bb3047a3588e607385ad1/ibis/expr/operations/temporal.py#L239-L249


cpcloud (approved):
Nice, thanks!

PR #5096, opened 2022-12-29
feat: implement `parse_url` for more backends

cpcloud:
This PR implements `parse_url` for snowflake and trino, and adds support for a few more values of `extract` for the clickhouse backend.

### DIFF
--- a/ibis/backends/clickhouse/compiler/values.py
+++ b/ibis/backends/clickhouse/compiler/values.py
@@ -292,28 +292,6 @@ def _regex_extract(op, **kw):
     return f"if({does_match}, {then}, NULL)"
 
 
-@translate_val.register(ops.ParseURL)
-def _parse_url(op, **kw):
-    arg = translate_val(op.arg, **kw)
-    extract = op.extract
-    if extract == "HOST":
-        return f"domain({arg})"
-    elif extract == "PROTOCOL":
-        return f"protocol({arg})"
-    elif extract == "PATH":
-        return f"path({arg})"
-    elif extract == "QUERY":
-        if (key := op.key) is not None:
-            key = translate_val(key, **kw)
-            return f"extractURLParameter({arg}, {key})"
-        else:
-            return f"queryString({arg})"
-    else:
-        raise com.UnsupportedOperationError(
-            f"ParseURL with extract {extract} is not supported"
-        )
-
-
 @translate_val.register(ops.FindInSet)
 def _index_of(op, **kw):
     values = map(partial(translate_val, **kw), op.values)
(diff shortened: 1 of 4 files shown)


cpcloud:
Passing on the snowflake backend as intended:

```
❯ pytest ibis/backends/tests/test_string.py::test_parse_url -m snowflake
======================================================================================== test session starts =========================================================================================
platform linux -- Python 3.10.9, pytest-7.2.0, pluggy-1.0.0
Using --randomly-seed=563107298
rootdir: /home/cloud/src/ibis, configfile: pyproject.toml
plugins: mock-3.10.0, repeat-0.9.1, benchmark-4.0.0, xdist-3.1.0, hypothesis-6.61.0, snapshot-0.9.0, clarity-1.0.1, profiling-1.7.0, randomly-3.12.0, cov-4.0.0
collected 120 items / 112 deselected / 8 selected

ibis/backends/tests/test_string.py ..x.x...                                                                                                                                                    [100%]

=========================================================================== 6 passed, 112 deselected, 2 xfailed in 16.35s ============================================================================
```

krzysztof-kwitt:
I wonder why we have one operation defined for data extraction from the URL, and 14 operations for data extraction?
https://github.com/ibis-project/ibis/blob/0f037cc7ca957c00fb15d55823a356cd30b038d2/ibis/expr/operations/strings.py#L225-L243
https://github.com/ibis-project/ibis/blob/0f037cc7ca957c00fb15d55823a356cd30b038d2/ibis/expr/operations/temporal.py#L125-L195
I also wonder if the name of the ParseURL operation and expression method describes its purpose well.
Here is a description of this function:
https://github.com/ibis-project/ibis/blob/b2260c5717dfe9d4d1038d46efb93d92f501551c/ibis/expr/types/strings.py#L638-L674
and here it is described that this function does two things: parses the URL and extract its components. We named this functions based on the first one, but I think it should be named after the second part, e.g. ExtractUrlComponent, ExtractUrlParts or something similar. 
What do you think?

cpcloud:
@krzysztof-kwitt `parse_url` is closely based on the Impala SQL API (which does both the parsing and extraction), and the various extraction operations are very loosely based on the `extract(property from y)` SQL function.

I agree that a rework of this particular API is probably in order, up to and including renaming it and/or splitting it into `ExtractFoo`-style operations, that accept `rlz.string`s as input. Something like:

```python
# current api

t.my_url  # a string column
t.my_url.parse_url('protocol')

# possible new api
t.my_url.protocol()
t.my_url.host()
t.my_url.path()
t.my_url.fragment()  # I prefer this to ref, but that's just me
t.my_url.authority()
t.my_url.file()
t.my_url.userinfo()
t.my_url.query(param="foo")  # extracts the value of the query parameter e.g., 1 in url?foo=1
```

Feel free to take a crack at this, and maybe we can get it in for 4.0.0!

Thanks for all of your great work!

PR #5097, opened 2022-12-29
bug: fix execution return type and casting in many backends

### LINKED ISSUE #2064: bug: different treatment of int columns in Pandas vs. Pyspark backend
If I create an ibis TableExpr that I mutate with an `IntegerColumn` `(dype=int8)`, the dtype of the resulting materialized pandas DataFrame is different between the Pandas backend and the Pyspark backend. For example:

`import pandas as pd`
`df = pd.DataFrame({'value': [1, 2, 3]})`
`client = ibis.pandas.connect({'table': df})`
`table = client.table('table')`
`expr = table.mutate(v=7)`
`expr.schema()  ## prints ibis.Schema({value  int64, v  int8})`
`result = expr.execute()`
`result.dtypes ## prints value  int64, v  int64`

That is, the `dtype=int8` of the assigned `IntegerColumn` is ultimately converted to `int64` in the resulting DataFrame.

Executing the same on the Pyspark backend, and then calling `.toPandas()` on the result, gives `dtype=int32` for column `v`.

It seems there should be a consistent dtype returned for column `v` between the two execution backends.

cpcloud:
This PR closes a long-standing issue seen in #2064.

While fixing that issue, there were a number of other casting issues related to pandas (and thus dask) and pyspark that are also fixed here.

Happily, a bunch of tests that were previous mark as `notyet` or `notimpl` started passing by fixing these issues.

There are still a couple failures that I need to address, but otherwise this PR is ready for review.

Closes #2064.


ibis/backends/pandas/client.py
```diff
@@ -135,6 +136,13 @@ def convert_boolean_to_series(in_dtype, out_dtype, column):
     return column
 
 
+@sch.convert.register(DatetimeTZDtype, dt.Date, pd.Series)
```
kszucs, line 139: Since we support more output types now I think we should factor out `sch.convert()` from `ibis.expr.schema` directly to the pandas backend.

cpcloud: Should we do that in a follow-up? I'm not 100% sure how we should do it though, but happy to make an issue if that's helpful!

kszucs: Sure, follow-up sgtm.

kszucs:
@cpcloud could you please rebase and squash?

kszucs (approved):
LGTM

PR #5115, opened 2023-01-01
feat(mssql): detect schema from SQL

krzysztof-kwitt:


ibis/backends/mssql/datatypes.py
```diff
@@ -0,0 +1,77 @@
...
+    'IMAGE': dt.Binary,
+    'VARBINARY': dt.Binary,
+    # Other data types
+    'UNIQUEIDENTIFIER': dt.UUID,
+    'GEOMETRY': dt.Geometry,
+    'GEOGRAPHY': dt.Geography,
+    # This timestamp datatype is also known as "rowversion", and the original name is really unfortunate.
+    # See:
+    # https://learn.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver16
+    'TIMESTAMP': dt.Timestamp,
```
cpcloud, line 76: It seems like this should map to `int64` instead, given the description:

> The rowversion data type is just an incrementing number and does not preserve a date or a time.

cpcloud (changes requested):
Small question about `rowversion`/`timestamp`.

cpcloud:
Other than the comment above, this PR looks good to me!

ibis/backends/mssql/datatypes.py
krzysztof-kwitt: Good catch. I looked at this docs again and i found the following snippet:

> A nonnullable rowversion column is semantically equivalent to a binary(8) column. A nullable rowversion column is semantically equivalent to a varbinary(8) column.

So I think, we should have `dt.Binary` here.

cpcloud: Sounds good to me!

cpcloud:
@krzysztof-kwitt Can you squash the `fixup!` commits? Then I'll merge!

cpcloud (approved):

mik-laj:
@cpcloud Done

cpcloud (approved):

PR #5137, opened 2023-01-04
refactor: make `Any`/`All`/`Quantile` subclass from `Filterable`

cpcloud:
This PR makes the remaining not filterable reductions filterable. I still need to add tests that hit the filtering branches of these aggs.

### DIFF
--- a/ibis/backends/tests/test_aggregation.py
+++ b/ibis/backends/tests/test_aggregation.py
@@ -501,6 +501,31 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                 )
             ],
         ),
+        param(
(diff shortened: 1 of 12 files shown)


ibis/backends/tests/test_aggregation.py
```diff
@@ -501,6 +501,31 @@ def mean_and_std(v):
...
+                        "duckdb",
+                        "impala",
+                        "mssql",
+                        "mysql",
+                        "polars",
+                        "postgres",
+                        "pyspark",
+                        "snowflake",
+                        "sqlite",
+                        "trino",
```
cpcloud, line 524: I'll follow up to make a good chunk of these pass.

kszucs (approved):
Nice!

PR #5139, opened 2023-01-04
ci: add more extras to lockfiles

cpcloud:
This PR updates the conda-lock file generation to include some of the newer extras.


.github/workflows/conda-lock.yml
```diff
@@ -71,15 +88,21 @@ jobs:
             --file "${python_version_file}" \
             --platform win-64 \
             --filename-template "${template}" \
+            -e bigquery \
```
gforsyth, line 91: if you want to add `duckdb` here, it's available on conda-forge now, so this should work.

gforsyth (approved):
looks good -- one comment, not blocking

.github/workflows/conda-lock.yml
cpcloud: Huzzah!

cpcloud:
I'm going to run this locally to make sure it works, moving to draft.

.github/workflows/conda-lock.yml
cpcloud: Done!

cpcloud:
Hm, encountering some issues with shapely, moving this back to draft.

cpcloud:
Ok, fixed the shapely issue, conda-lock appears not to do package name normalization and requires an exact match.

cpcloud:
Definitely meant to not include the shapely bump, reverting that, then merging on green.

PR #5145, opened 2023-01-05
bump docs deps

cpcloud:
This PR bumps our version of `mkdocstrings` to the latest version. Among other improvements, this bump makes it possible to build the docs in any of our supported environments, not just the nix environment. I have done a spot check of the API documentation page using `mkdocs serve --strict` and things are in order. We are picking up the nifty new interlinked `Returns` types as well. A few fixes to our docstrings found by the bump are fixed as well.


kszucs (approved):
Tried it out locally, works nicely!

PR #5147, opened 2023-01-06
fix(ir): force an alias if projecting columns

### LINKED ISSUE #5753: bug(clickhouse): backend doesn't generate necessary aliases for some subqueries
### What happened?

Here's a failing test case from #5147:

```python
@pytest.mark.notimpl(["datafusion"])
def test_value_counts_on_expr(backend, alltypes, df):
    expr = alltypes.timestamp_col.date().value_counts()
    result = expr.execute().sort_values(["Date(timestamp_col)"]).reset_index(drop=True)
    expected = (
        df.timestamp_col.dt.date.astype("datetime64[ns]").value_counts().reset_index()
    )
    expected.columns = result.columns
    expected = expected.sort_values(["Date(timestamp_col)"]).reset_index(drop=True)
    backend.assert_frame_equal(result, expected)
```

`master`

ClickHouse

### Relevant log output

_No response_

### Code of Conduct

### LINKED ISSUE #5969: bug(datafusion): Invalid aggregate expression 'COUNT(Int64(1)) AS tmp'
### What happened?

Count of records in the table to be returned. 

Minimal repro (I'm using one month of NYC taxi data stored as parquet)

```
import ibis

trips_df = ibis.datafusion.connect({'trips': '../data/yellow_tripdata_2009-08.parquet'}).table('trips')
print(f'Original record count: {trips_df.count().execute():,}')
```

This throws `Exception: Internal error: Invalid aggregate expression 'COUNT(Int64(1)) AS tmp'. This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker`

5.1.0

Datafusion

### Relevant log output

```sh
Exception                                 Traceback (most recent call last)
Cell In[12], line 1
----> 1 print(f'Original record count: {trips_df.count().execute():,}')

File ~/miniconda3/envs/demo_env/lib/python3.10/site-packages/ibis/expr/types/core.py:303, in Expr.execute(self, limit, timecontext, params, **kwargs)
    276 def execute(
    277     self,
    278     limit: int | str | None = 'default',
   (...)
    281     **kwargs: Any,
    282 ):
    283     """Execute an expression against its backend if one exists.
    284 
    285     Parameters
   (...)
    301         Keyword arguments
    302     """
--> 303     return self._find_backend(use_default=True).execute(
    304         self, limit=limit, timecontext=timecontext, params=params, **kwargs
    305     )

File ~/miniconda3/envs/demo_env/lib/python3.10/site-packages/ibis/backends/datafusion/__init__.py:248, in Backend.execute(self, expr, params, limit, **kwargs)
    241 def execute(
    242     self,
    243     expr: ir.Expr,
   (...)
    246     **kwargs: Any,
    247 ):
--> 248     output = self.to_pyarrow(expr, params=params, limit=limit, **kwargs)
    249     if isinstance(expr, ir.Table):
    250         return output.to_pandas()

File ~/miniconda3/envs/demo_env/lib/python3.10/site-packages/ibis/backends/base/__init__.py:267, in _FileIOHandler.to_pyarrow(self, expr, params, limit, **kwargs)
    261 pa = self._import_pyarrow()
    262 try:
    263     # Can't construct an array from record batches
    264     # so construct at one column table (if applicable)
    265     # then return the column _from_ the table
    266     table = pa.Table.from_batches(
--> 267         self.to_pyarrow_batches(expr, params=params, limit=limit, **kwargs)
    268     )
    269 except ValueError:
    270     # The pyarrow batches iterator is empty so pass in an empty
    271     # iterator and a pyarrow schema
    272     schema = expr.as_table().schema()

File ~/miniconda3/envs/demo_env/lib/python3.10/site-packages/ibis/backends/datafusion/__init__.py:239, in Backend.to_pyarrow_batches(self, expr, params, limit, chunk_size, **kwargs)
    237 pa = self._import_pyarrow()
    238 frame = self._get_frame(expr, params, limit, **kwargs)
--> 239 return pa.ipc.RecordBatchReader.from_batches(frame.schema(), frame.collect())

Exception: Internal error: Invalid aggregate expression 'COUNT(Int64(1)) AS tmp'. This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker
```

### Code of Conduct

cpcloud:
This PR fixes a bug where the new ClickHouse compiler was not naming all columns in a projection or aggregation.

Closes #5753.
Closes #5969.


gforsyth (commented):
Looks good on the Clickhouse end -- not sure what Polars is upset about, though

cpcloud:
Looking into it. BigQuery also fails, and I'll push up a somewhat gross fix for that soon.

cpcloud:
Ok, I'm reworking this to include the transformation at the `Node` level, since both ClickHouse and Polars have the same problem.

cpcloud:
The polars breakage goes deeper than I'd like to address in this PR, so I'll do a follow up.

ibis/backends/polars/compiler.py
```diff
@@ -619,8 +619,10 @@ def distinct(op):
 def count_star(op):
     if (where := op.where) is not None:
         condition = translate(where)
-        return condition.filter(condition).count()
-    return pl.count()
+        # TODO: clean up the casts and use schema.apply_to in the backend's
+        # execute method
+        return condition.filter(condition).count().cast(pl.Int64)
```
cpcloud, line 624: I'll have another PR for this, but it's going to disable support for some kinds of joins because polars seems to only return one of the key columns.

jcrist (approved):
:shipit:

ibis/expr/operations/relations.py
```diff
@@ -329,6 +329,14 @@ def schema(self):
         return sch.Schema(names, types)
 
 
+def _add_alias(op: ops.Value | ops.TableNode):
```
kszucs, line 332: Why is it required? All value operations must be named already.

kszucs: Even if we want to force aliasing here why not for TableColumns then?

kszucs: We would also need to restrict the selections fields to only allow aliases instead of generic values.

cpcloud: This is required because subselects generate expressions without aliases in some backends. This means any subsequent references to the corresponding column in such subselects will produce invalid code, because there's no predictable name generated to reference.

cpcloud: We don't need to do this for `TableColumn`s because those have predictable names.

cpcloud: > We would also need to restrict the selections fields to only allow aliases instead of generic values.

Why? Simply making this the case, as I am doing here should be sufficient.

cpcloud: We also cannot just blindly alias everything, since that would start to generate invalid code like

```
SELECT *
FROM t
WHERE x AS thing = 1
```

cpcloud:
Closing in favor of a better solution. I made #5753 to track this.

cc @kszucs

cpcloud:
Reopening this since a better solution hasn't yet come along 🙂

ibis/backends/polars/compiler.py
cpcloud: This comment is now outdated, resolving.

kszucs (approved):

PR #5151, opened 2023-01-06
DOCS: chore: remove warning in extending tutorial

NickCrews:
Before, I would get

```
/var/folders/nz/wfwcn6md1796t48tx9dhf0v40000gn/T/ipykernel_98244/3752312616.py:26: FutureWarning: `Node.op` is deprecated as of v4.0; remove intermediate .op() calls
  (arg,) = expr.op().args
```

After this change, I don't.

### DIFF
--- a/docs/user_guide/extending/elementwise.ipynb
+++ b/docs/user_guide/extending/elementwise.ipynb
@@ -138,7 +138,7 @@
     "@ibis.sqlite.add_operation(JulianDay)\n",
     "def _julianday(translator, expr):\n",
     "    # pull out the arguments to the expression\n",
-    "    (arg,) = expr.op().args\n",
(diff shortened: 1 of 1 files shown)


jcrist (commented):
Thanks @NickCrews! The fix looks good to me - the only issue is the formatting of your commit message. We use structured commit messages to generate the changelog, tests will fail if you don't follow the same format (see https://ibis-project.org/docs/dev/community/contribute/02_workflow/#writing-the-commit).

Can you amend your commit message to something like:

```
chore: remove warning in extending tutorial 
```

NickCrews:
It was easier for me to just make a new PR at https://github.com/ibis-project/ibis/pull/5152 than to actually pull locally and `git commit --amend.` See that PR.

PR #5156, opened 2023-01-07
test(backend): add tests for ops.DecimalPrecision, ops.DecimalScale

krzysztof-kwitt:
Hello,

I am not convinced that this is implemented in any backend, but I want to check it on CI.

### DIFF
--- a/ibis/backends/tests/test_numeric.py
+++ b/ibis/backends/tests/test_numeric.py
@@ -174,7 +174,7 @@ def test_isnan_isinf(
         param(L(11) % 3, 11 % 3, id='mod'),
         param(
             L(12.3).cast(dt.Decimal(30, 8)).precision(),
-            30,
+            8,
             id='scale',
             marks=[
                 # Only impala support .scale()
@@ -200,7 +200,7 @@ def test_isnan_isinf(
         ),
         param(
             L(12.3).cast(dt.Decimal(30, 8)).precision(),
-            8,
+            30,
             id='precision',
             marks=[
                 # Only impala support .precision()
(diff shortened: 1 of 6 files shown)


ibis/backends/tests/test_numeric.py
```diff
@@ -172,6 +172,58 @@ def test_isnan_isinf(
             marks=pytest.mark.notimpl(["datafusion", "impala"]),
         ),
         param(L(11) % 3, 11 % 3, id='mod'),
+        param(
+            L(12.3).cast(dt.Decimal(30, 8)).precision(),
+            8,
+            id='scale',
+            marks=[
+                # Only impala support .scale()
```
cpcloud, line 180: Should we remove these APIs instead of adding tests for them? An API that only a single backend implements is a strong candidate for removal IMO.

cpcloud: If someone really needs this, then a custom elementwise operation would be the way to do it. In other cases than impala, this information can be accessed cheaply and easily using the expression's `.type()` method followed by a `.precision`/`.scale` access:

```python
t = ...
expr = t.foo # decimal column
expr.type().precision
expr.type().scale
```

This has the benefit of never having to run a query to access the precision/scale information.

If that's truly needed as a column of constants, then a user can project those.

krzysztof-kwitt: As the first step, I wanted to add a test, so that later in the follow PR I would think about how to implement it in the other backends, but I also think that it will not be too popular API. I will prepare a new PR that will delete these operations.

krzysztof-kwitt: Here is PR: https://github.com/ibis-project/ibis/pull/5161

kszucs:
Closing in favor of https://github.com/ibis-project/ibis/pull/5161

PR #5158, opened 2023-01-07
test(perf): remove unnecessary and very slow tests

cpcloud:
This PR removes tests that are either duplicated (`test_computable_input`) or unnecessary (the `multipledispatch` ambiguity tests) because we have strict warnings enabled (`AmbiguityWarning` will cause tests to fail).


ibis/backends/dask/tests/test_core.py
```diff
@@ -108,58 +99,6 @@ def tmp_left_join_exe(op, lhs, **kwargs):
     assert count[0] == 1
 
 
-def test_is_computable_input():
```
cpcloud, line 111: There's an exact copy of this test in the pandas backend, so this test can safely be deleted entirely.

ibis/backends/dask/tests/test_core.py
```diff
@@ -15,19 +11,14 @@
...
+from ibis.backends.dask.core import execute  # noqa: E402
 from ibis.backends.dask.dispatch import (  # noqa: E402
     execute_node,
     post_execute,
     pre_execute,
 )
 
 
-@pytest.mark.parametrize('func', [execute_node, pre_execute, post_execute])
-def test_no_execute_ambiguities(func):
```
cpcloud, line 27: This will fail if there's an ambiguity due to our strict warnings setting for pytest.

ibis/backends/pandas/tests/test_core.py
```diff
@@ -168,11 +162,6 @@ def is_computable_input_my_object(_):
     result = execute(four.op())
     assert result == 4.0
 
-    del execute_node[ops.Add, int, MyObject]
-
-    execute_node.reorder()
-    execute_node._cache.clear()
```
cpcloud, line 174: This seems to have no effect on anything, and is slow so I deleted it.

ibis/tests/expr/test_analysis.py
```diff
@@ -306,19 +304,3 @@ def test_agg_selection_does_not_share_roots():
 
     with pytest.raises(com.RelationError, match="Selection expressions"):
         gb.aggregate(n=n)
-
-
-@pytest.mark.parametrize("num_joins", [1, 10])
-def test_large_compile(num_joins):
```
cpcloud, line 312: We have a benchmark for this already, so I deleted it.

PR #5161, opened 2023-01-07
refactor(ir): remove DecimalValue.precision(), DecimalValue.scale() method

krzysztof-kwitt:
Hello,

This API is only implemented by one backend, and there is an alternative method to obtain this information.
See: https://github.com/ibis-project/ibis/pull/5156#discussion_r1064000868

Best regards,


kszucs (approved):
Thanks @krzysztof-kwitt!

PR #5165, opened 2023-01-08
feat(api): return object from `to_sql` to support notebook syntax highlighting

cpcloud:
Add a custom `SQLString` object that formats with syntax highlighting in the notebook.


kszucs:
Hmm, the duckdb tests are failing.

kszucs:
> Hmm, the duckdb tests are failing.

A test case with a regex match on the output string was failing, fixed it.

PR #5167, opened 2023-01-08
chore(dev-deps): bump ruff and fixup/ignore new failures

cpcloud:
Bump ruff to the latest version and fixup/ignore new failures.


ibis/expr/datatypes/cast.py
```diff
@@ -165,11 +165,10 @@ def can_cast_map(source, target, **kwargs):
 def can_cast_struct(source, target, **kwargs):
     source_pairs = source.pairs
     target_pairs = target.pairs
-    for name in {*source.names, *target.names}:
-        if name in target_pairs:
-            if not castable(source_pairs[name], target_pairs[name]):
-                return False
-    return True
+    return all(
```
kszucs, line 168: Comprehensions like this is less readable for me than the previous for loop.

kszucs: The whole function body could be reduced to 

```py
return all(castable(source[field], target[field]) for field in target.names)
```

kszucs:
Going to squash this before merging.

kszucs (approved):

PR #5176, opened 2023-01-09
docs: add ibis ci data analysis blog

cpcloud:
- docs: add bigquery ci data analysis notebook


ibis/backends/bigquery/registry.py
```diff
@@ -688,13 +688,10 @@ def _nth_value(t, op):
     ops.RandomScalar: fixed_arity("RAND", 0),
     ops.NthValue: _nth_value,
     ops.JSONGetItem: lambda t, op: f"{t.translate(op.arg)}[{t.translate(op.index)}]",
+    ops.TimestampDiff: lambda t, op: f"TIMESTAMP_DIFF({t.translate(op.left)}, {t.translate(op.right)}, SECOND)",
```
krzysztof-kwitt, line 691:
```suggestion
    ops.TimestampDiff: binary_infix.binary_infix_op('-'),
```
Thanks to this, the column type will be set to [INTERVAL type](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#interval_type), but i am not sure if this is the best solution.

cpcloud: Oh, let me give that a try, that's better for ibis, since that's what we actually infer the type as.

jcrist (approved):
This LGTM. Nice example of using ibis on a real-world problem, as well as integrating it with the larger python ecosystem (`plotnine`). Nice work!

ibis/backends/bigquery/registry.py
cpcloud: Turns out to be a big pain to do this, because the execution return type is `DateOffset`, which is pretty awful to use.

docs/blog/rendered/ci-analysis.ipynb
```diff
@@ -0,0 +1,1591 @@
...
...F8gDJ8/jECw7/MNRaVQApHsDwOAxSzD6TDD6TDD5bDA5bQgNcWKVKcFGWkpKCnMMKTtfOmtUrh9IRTkpAzI95/r8Bp1yaS+JJQmte3vrOO9997Dl770JcyaNQs7d+4c6uYIBAKBQCAQCAQCwVkD5UadQCTlVcATSd/XrfabZgzWG9IixnZJiqYk4juurXZmSOPRQdzQzuvW9JJKk4YCUL1uKO52qD43M+6EgqBBH6jZBoyby4wuwSDUze8wp5gSZOm8nOkg51+vpf1S3/gZiN8dCTfhbSaQVn0DUg6LkiSEAJ0toI1VILIFZNgYyHan9l5350bM70jUDOFpsnqoC6hdI3f06VNicaOaLsqPainLdAb4SIo+IumMajwll8nEHB12p+acSU4GvMwRoI+a8HREotkCwMQFkHJLoPo8oHs+Ak4cZq/njWBG6dKPmSGWG7ElGWTWSuawmDAf6GiC+tbTfYtQmXYBSFYhyNSlMddC2xpAy3ezPrM7mWPX5mROMbuTOR57um5u1NQb4HlKQC3VoU7OeVrESDQhj8ZKytlGVS0aQqtRFw4Bnc2RaBRvtPaXrk4dWXAZyKjpUP/wfaPxUpKAJddCmnKe4TrVg1tA17wMLbZKkoExMyF96etA9SFQTwfrz7wRLA3tzg9A2xuB1GxI085njkRd/9DqQ0BnM/uOlDQgq4hdv6r0OQJtoKA8WsdsZU6QU0R59UfMGaKHp7kcDHg9RV3tLmneF0HyRkDdvY45JCw2zWivRa7wNLmnKZqKqmp0EwdPP9paH0mj6zduNgj62f3wdkQdc35PHyPX+kBv90eSmRMvMx/y9Y+y62k4DtrVyqJ/03JATxxisstrM3LHarcadf3dFEJVhT3vkTqKCPmZs8LvYc5in5vVp/R7opHbPPJv5nJg5wfxTzxxIci080ECPqjHdgAHt0T6hgCyDOkr32P1HBO1K7JpAgCr66nTX2rlPtB3nu312qRbfszuRVcr1G3/Yk7iSDQ1MVuAtBxWT7KtIeKcdIKMmgYSGVMppSx9o2yOpjKM9H28fqc8pSF3xmg16nyaA5BqDm5d/dSAV6sLiIAPZMYySFPP7/X64n6/u5XNc1LSoL74zUi0drd+uemHoMcPgH76VpKR92DX3t0xpY+M1l6TQeZcBGnJ1X1re9APNNdEHG8RHaIfs6yRiMKOJmN2ASWSH
```
gforsyth, line 1559:
```suggestion
    "If you have a short queue time, but long job runs then you'll be bottlenecked on individual jobs, and if you have more jobs than queue slots then you'll be blocked on queueing time.\n",
```

gforsyth (commented):
Very cool write-up -- another confounding factor (that would be messy to detangle) is the increasing number of jobs we run overall, which is almost certainly adding to the queueing time.

Small wording nits is all I would change -- one other possible change is to add `;` to the end of the plotting cells to get rid of the `<ggplot: ...>` text outputs but they're pretty inoffensive.

ibis/backends/bigquery/registry.py
cpcloud: In a follow up PR, I'll introduce a `.delta` method that allows specifying units and returning `int64`.s

cpcloud: I'll rework the notebook to use the epoch seconds from the timestamps.

PR #5178, opened 2023-01-09
feat(common): make frozendict truly immutable 

kszucs:
- now frozendict instances are truly immutable using the builtin dictproxy which only provides a read-only view to the underlying mapping
- support validator construction from `UnionType` (`type1 | type2 | ...`) which is different from `typing.Union`
- add a `Coercible` abstract base class for marking types which can be constructed from differently typed object, this makes type annotation based validations more flexible


cpcloud (approved):
Sweet!

PR #5180, opened 2023-01-09
ci: there can be only one docs pushing build

cpcloud:


.github/workflows/ibis-docs-release.yml
```diff
@@ -6,7 +6,7 @@ on:
       - published
 jobs:
   docs:
-    concurrency: docs-${{ github.repository }}-${{ github.head_ref || github.sha }}
```
cpcloud, line 9: My understanding of what went wrong in the docs push, and why it collided was:

`concurrency` for the release commit for the `docs_push` job was

```
docs-ibis-781c1e4b98f29c7d6f10ed6013b88c17c996bf2b
```

while `concurrency` for the `docs` job in `ibis-docs-release.yml` was:

```
docs-ibis-4.0.0
```

coming from `github.head_ref`.

These point to the same commit, but are not equal values of `concurrency` so they were allowed to execute in parallel.

This PR changes it so that the concurrency is just `docs-ibis`, making it so that any job that will push to the docs repo has to wait for any currently running jobs to finish.

PR #5184, opened 2023-01-10
docs(release): update breaking changes in the release notes for 4.0.0

kszucs:


docs/release_notes.md
```diff
@@ -6,62 +6,40 @@ Release Notes
...
-This also adds in a general backend test for the `register` API so we
-can make sure things are consistent, and fixes a small bug with the
-dispatch in polars (and datafusion which I copied) that raises a
-recursion error if you pass in garbage.
-
-Update ibis/backends/tests/test_client.py
-
-Co-authored-by: Phillip Cloud <417981+cpcloud@users.noreply.github.com>
-* **ir:** NodeList and ir.List are removed. Use tuples instead.
+* functions, methods and classes marked as deprecated are removed now
```
cpcloud, line 9: Does all of this need to be fixed? Can you just fix the datafusion text and leave the rest alone?

cpcloud: The idea is to have the breaking change message match the release notes exactly, and improve over time. Rewriting history should be a last resort, in the case the docs are unreadable without the hand-editing so we have to do a bit of rewriting.

cpcloud: It looks fine just fixing that bit:

image

cpcloud: Nevermind, this is fine as is.

cpcloud (approved):

PR #5185, opened 2023-01-10
fix(deps): make sure pyarrow is not an implicit dependency

cpcloud:
This PR removes pyarrow as test dependency which was hiding an implicit postgres backend dependent.


ibis/backends/tests/test_register.py
```diff
@@ -341,6 +360,7 @@ def test_csv_reregister_schema(con, tmp_path):
     ]
 )
 def test_register_garbage(con):
+    sa = pytest.importorskip("sa")
```
gforsyth, line 363:
```suggestion
    sa = pytest.importorskip("sqlalchemy")
```

gforsyth (changes requested):
Looks good, just the one change, I think.

ibis/backends/tests/test_register.py
cpcloud: Thank you.

gforsyth:
I think the mssql error is it being flaky on login, but the impala failure looks like an issue with the reconfigured test setup

ibis/backends/impala/tests/conftest.py
```diff
@@ -152,12 +152,7 @@ def connect(
         )
 
     def _get_original_column_names(self, tablename: str) -> list[str]:
-        import pyarrow.parquet as pq
-
-        pq_file = pq.ParquetFile(
-            self.data_directory / "parquet" / tablename / f"{tablename}.parquet"
-        )
-        return pq_file.schema.names
+        return list(TEST_TABLES[tablename].keys())
```
gforsyth, line 155:
```suggestion
        return list(TEST_TABLES[tablename].names)
```
?

PR #5186, opened 2023-01-10
feat(clickhouse): properly support native boolean types

cpcloud:
This PR add support for native boolean types in the clickhouse backend

### DIFF
--- a/ibis/backends/clickhouse/datatypes.py
+++ b/ibis/backends/clickhouse/datatypes.py
@@ -4,6 +4,7 @@ import functools
 
 import parsy
 
+import ibis
 import ibis.expr.datatypes as dt
 from ibis.common.parsing import (
     COMMA,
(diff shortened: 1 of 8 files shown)


ibis/backends/clickhouse/datatypes.py
```diff
@@ -19,6 +20,10 @@
 )
 
 
+def _bool_type():
+    return getattr(getattr(ibis.options, "clickhouse", None), "bool_type", "Boolean")
```
kszucs, line 24: Isn't that always available in the configuration?

cpcloud: `ibis.options.clickhouse` doesn't get registered until the module-level `__getattr__` in `ibis/__init__.py` is called with `'clickhouse'`. It's possible to parse a clickhouse type without ever having accessed a backend instance.

Will it ever be the case that a user runs code in `clickhouse/datatypes.py` without having accessed `ibis.clickhouse`? Unlikely. However, I do not want to have to type `ibis.clickhouse` to test this code locally.

PR #5187, opened 2023-01-10
test(mssql): simplify test data setup

cpcloud:
This PR removes the use of concurrent.futures for data loading in the mssql backend, in an attempt simplify debugging the flaky schema-getting test. I also moved some things that should be in transactions but are not to be so.


gforsyth (approved):
looks good!

PR #5182, opened 2023-01-10
fix(snowflake): Remove invalid operations from registry

krzysztof-kwitt:
Hello,

Deleting these operations does not affect the result of the tests, so it looks like they are invalid. When we exclude them from the register, the operation matrix will show a more realistic coverage and we will also know where a contribution is needed.

PR #5188, opened 2023-01-10
Fix snowflake tests after deleting unsupported operations

krzysztof-kwitt:
Follow-up: https://github.com/ibis-project/ibis/pull/5182

### DIFF
--- a/ibis/backends/snowflake/registry.py
+++ b/ibis/backends/snowflake/registry.py
@@ -187,13 +187,6 @@ _invalid_operations = {
     ops.NotExistsSubquery,
     # ibis.expr.operations.maps
     ops.MapKeys,
-    # ibis.expr.operations.numeric
-    ops.BitwiseAnd,
-    ops.BitwiseLeftShift,
-    ops.BitwiseNot,
-    ops.BitwiseOr,
-    ops.BitwiseRightShift,
-    ops.BitwiseXor,
     # ibis.expr.operations.reductions
     ops.All,
     ops.Any,
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -699,9 +699,6 @@ def test_select_filter_select(backend, alltypes, df):
     backend.assert_series_equal(result, expected)
 
 
-@pytest.mark.notimpl(
-    ["snowflake"], reason="snowflake doesn't implement timestamp subtraction"
-)
 def test_interactive(alltypes):
     expr = alltypes.mutate(
         str_col=_.string_col.replace("1", "").nullif("2"),
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -1,7 +1,7 @@
 import decimal
 import io
 from contextlib import redirect_stdout
-from operator import and_, invert, lshift, neg, or_, rshift, xor
+from operator import invert, neg
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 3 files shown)


ibis/backends/snowflake/registry.py
```diff
@@ -187,13 +187,6 @@ def _extract_url_query(t, op):
     ops.NotExistsSubquery,
     # ibis.expr.operations.maps
     ops.MapKeys,
-    # ibis.expr.operations.numeric
```
krzysztof-kwitt, line 190: The tests for numeric operations were expected to be in the `test_numeric.py` file, so that's where the problem came from. I'm moving it now.

ibis/backends/tests/test_generic.py
```diff
@@ -699,111 +699,6 @@ def test_select_filter_select(backend, alltypes, df):
...
-
-@pytest.mark.notimpl(["bigquery", "dask", "datafusion", "pandas", "snowflake"])
-def test_bitwise_not_col(backend, alltypes, df):
-    expr = (~alltypes.int_col).name("tmp")
-    result = expr.execute()
-    expected = ~df.int_col
-    backend.assert_series_equal(result, expected.rename("tmp"))
-
-
-@pytest.mark.notimpl(
```
krzysztof-kwitt, line 804: Now the result of the operation is as below. I'm not sure if this is expected behavior or if we should throw an exception.
```
Translation to backend failed
Error message: No translation rule for <class 'ibis.expr.operations.temporal.TimestampDiff'>
Expression repr follows:
r0 := AlchemyTable: functional_alltypes
  index           int64
  Unnamed: 0      int64
  id              int64
  bool_col        boolean
  tinyint_col     int64
  smallint_col    int64
  int_col         int64
  bigint_col      int64
  float_col       float64
  double_col      float64
  date_string_col string
  string_col      string
  timestamp_col   timestamp
  year            int64
  month           int64

Selection[r0]
  selections:
    r0
    str_col:             NullIf(StringReplace(r0.string_col, pattern='1', replacement=''), null_if_expr='2')
    date_col:            Date(r0.timestamp_col)
    delta_col:           TimestampNow() - r0.timestamp_col
```

cpcloud: Any particular reason you removed this marker?

ibis/backends/snowflake/registry.py
cpcloud: Can you elaborate a bit on what you mean? This list of ops shouldn't be related to where we happen to be testing them.

cpcloud: I think I understand. Were you only running a subset of the tests?

ibis/backends/tests/test_generic.py
cpcloud: Oh, I see now. This behavior is expected, but not ideal. Ideally we should implement some form of `TimestampDiff` for snowflake, but that can be done in a follow up.

krzysztof-kwitt: Now this tests does not raise exception, so when we have market with `strict` set to True set, this test fails.

> [XPASS(strict)] snowflake doesn't implement timestamp subtraction

https://github.com/ibis-project/ibis/actions/runs/3885821095/jobs/6630175834

ibis/backends/snowflake/registry.py
krzysztof-kwitt: I run tests multiple times. Each time I run it, I remove a different operation, then run a subset of tests depending on the name of the module in which the operation is.

My heuristic is similar to the code below.
```python
ops = [
    ...
    'ibis.expr.operations.analytic.CumeDist',
    ...
]
....
for op in ops
  if '.numeric.' in op:
      test_files = list(map(str, Path("ibis/backends/tests/").glob("test_numeric.py")))
  elif '.strings.' in op:
      test_files = list(map(str, Path("ibis/backends/tests/").glob("test_string.py")))
  elif '.arrays.' in op:
      test_files = list(map(str, Path("ibis/backends/tests/").glob("test_array.py")))
  elif '.reductions.' in op:
      test_files = list(map(str, Path("ibis/backends/tests/").glob("test_agg*.py")))
  else:
      test_files = list(map(str, Path("ibis/backends/tests/").glob("test_*.py")))
```
We don't have any problem in the Ibis core, but it was a mistake on my side.

cpcloud (approved):
LGTM, thanks!

PR #5189, opened 2023-01-10
fix(backend): read the default backend setting in `_default_backend`

cpcloud:
- test(backend): add default backend setting and reading test
- fix(backend): read the default backend setting in `_default_backend`

### DIFF
--- a/ibis/config.py
+++ b/ibis/config.py
@@ -167,29 +167,19 @@ class Options(Config):
     pyspark: Optional[Config] = None
 
 
-_HAS_DUCKDB = True
-_DUCKDB_CON = None
-
-
 def _default_backend() -> Any:
-    global _HAS_DUCKDB, _DUCKDB_CON
-
-    if not _HAS_DUCKDB:
-        return None
-
-    if _DUCKDB_CON is not None:
-        return _DUCKDB_CON
+    if (backend := options.default_backend) is not None:
+        return backend
 
     try:
         import duckdb as _  # noqa: F401
     except ImportError:
-        _HAS_DUCKDB = False
         return None
 
     import ibis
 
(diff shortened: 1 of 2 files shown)


ibis/config.py
```diff
@@ -172,6 +172,9 @@ class Options(Config):
 
 
 def _default_backend() -> Any:
+    if (backend := options.default_backend) is not None:
+        return backend
```
jcrist, line 176: Should we maybe move to storing the duckdb con on `options.default_backend` rather than a hidden global variable?

jcrist: I guess as written the user can reset `options.default_backend = None` later on and the same duckdb connection will be reused, rather than opening a new one.

cpcloud: The idea was to reuse the existing in-memory duckdb connection if it's ever created.

jcrist: right, but it could be reused if it's set on `options.default_backend` instead of a hidden global.

The only difference is that with the current implementation the following works:

```python
con1 = _default_backend()  # a duckdb backend

# Manually set the default backend to something else
con2 = ibis.bigquery.connect(...)
options.default_backend = con2
assert _default_backend() is con2

# Reset it back to None
options.default_backend = None

con3 = _default_backend()
assert con1 is con3  # original duckdb backend is reused
```

If this is intended behavior, then what you have here seems fine. No strong thoughts.

jcrist (approved):

ibis/config.py
cpcloud: I see what you're saying. It wasn't intentional when I wrote this originally. I think it probably makes more sense for `None` to reset the default backend rather than have a global floating around.

PR #5190, opened 2023-01-10
fix(trino/duckdb): remove invalid operations from registry

krzysztof-kwitt:


ibis/backends/duckdb/registry.py
```diff
@@ -255,3 +255,29 @@ def _struct_column(t, op):
...
+    ops.CumulativeOp,
+    ops.NTile,
+    # ibis.expr.operations.generic
+    ops.TableArrayView,
+    ops.TypeOf,
+    # ibis.expr.operations.logical
+    ops.ExistsSubquery,
+    ops.NotExistsSubquery,
+    # ibis.expr.operations.reductions
+    ops.CountDistinct,
```
cpcloud, line 273: I don't think you can remove a bunch of these because duckdb reuses the postgres rules, most of which are valid for duckdb.

cpcloud (changes requested):

ibis/backends/duckdb/registry.py
krzysztof-kwitt: I run all the backend tests locally and they succeed, so now we should either add tests to cover these operations or delete them because it's dead code. however, I see that the TPC-H job failed, so we should probably add new tests. I'll deal with it tomorrow.

krzysztof-kwitt: Screenshot 2023-01-10 at 21 49 58

krzysztof-kwitt: CI passed.

cpcloud: Interesting, yeah then we are lacking coverage! Thanks for finding these, much appreciated.

cpcloud:
@krzysztof-kwitt LGTM! Can you roll up the commits into the two `fix` commits? Then I'll merge!

krzysztof-kwitt:
@cpcloud Done

cpcloud (approved):

PR #5195, opened 2023-01-11
feat(impala): implement ops.Clip

krzysztof-kwitt:


cpcloud:
@krzysztof-kwitt This seems okay codewise, but there are some test failures that seem to indicate something is off.

cpcloud:
@krzysztof-kwitt Anything you need help with here? Let's see if we can wrap this up soon!

krzysztof-kwitt:
I don't understand why the tests fail, and in addition, running tests for Impala locally is very problematic on Mac OS, because UDF functions are compiled during the tests, so for now I gave it up hoping that someday I'll have more time and then I'll be able to fight with the environment and this change. If you have the capacity you can take over this PR.

cpcloud:
This looks like an ordering issue. I'll fix it up and merge!

cpcloud (approved):

Issue #5191, opened 2023-01-10
bug: Scalar.ifelse(Column, Column) results in Scalar, not Column


NickCrews:
### What happened?

```python
import pandas as pd
import ibis
ibis.options.interactive = True
ibis.options.verbose = True

df = pd.DataFrame({"x": range(3)})
inp = ibis.memtable(df)
ibis.literal(True).ifelse(inp.x, -inp.x)
```
Results in the scalar 0, what I'm guessing is the first entry in inp.x:

```
SELECT CASE WHEN CAST(:param_1 AS BOOLEAN) THEN t0.x ELSE -t0.x END AS "Where(True, x, Negate(x))" 
FROM _ibis_memtable1 AS t0
0
```

I would expect the Scalar `True` to get broadcast to the cardinality of the two columns, so I get `[0, 1, 2]`

### What version of ibis are you using?

4.0.0

### What backend(s) are you using, if any?

DuckDB

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kszucs:
Thanks for the report, https://github.com/ibis-project/ibis/pull/5197 should fix it.

NickCrews:
Thank you for the super quick fix! Been really liking Ibis so far. Occasionally I find a rough edge, but you seem to be working on this so actively that I'm confident that I can find a workaround. Thank you!

PR #5197, opened 2023-01-11
fix(ir): ops.Where output shape should consider all arguments

kszucs:
fixes https://github.com/ibis-project/ibis/issues/5191


cpcloud (approved):
Thanks for the fix!

PR #5203, opened 2023-01-11
fix(set ops): raise if no tables passed to set operations

gforsyth:
Raises if no tables are passed to set operations that require tables to execute against (`union`, `intersect`, `difference`)

### DIFF
--- a/ibis/backends/tests/test_set_ops.py
+++ b/ibis/backends/tests/test_set_ops.py
@@ -3,6 +3,7 @@ import pytest
 from pytest import param
 
 import ibis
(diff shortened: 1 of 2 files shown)


cpcloud:
Can we just require a second argument for all of these instead of having to return an incorrect result for intersect?

cpcloud:
Also I don't think you can apply set logic in the non distinct case so I think we should probably require two arguments.

gforsyth:
> Can we just require a second argument for all of these instead of having to return an incorrect result for intersect?

yep, done

jcrist (approved):

PR #5213, opened 2023-01-12
fix(sqlalchemy): fix correlated subqueries on projections

cpcloud:
While adding test coverage for ExistsSubquery and NotExistsSubquery I discovered that basic execution of correlated subqueries does not work. This PR fixes that. A couple of tests are still failing, hence the draft.

### DIFF
--- a/ibis/backends/base/sql/alchemy/query_builder.py
+++ b/ibis/backends/base/sql/alchemy/query_builder.py
@@ -42,7 +42,7 @@ class _AlchemyTableSetFormatter(TableSetFormatter):
         for jtype, table, preds in zip(
             self.join_types, self.join_tables[1:], self.join_predicates
         ):
-            if len(preds):
+            if preds:
                 sqla_preds = [self._translate(pred) for pred in preds]
                 onclause = functools.reduce(sql.and_, sqla_preds)
             else:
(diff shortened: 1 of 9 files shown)


ibis/backends/base/sql/alchemy/query_builder.py
```diff
@@ -227,32 +241,7 @@ def _add_select(self, table_set):
...
-        #
-        # previously this was `replace_selectable`, but that's deprecated so we
-        # inline its implementation here
-        #
-        # sqlalchemy suggests using the functionality in sa.sql.visitors, but
-        # that would effectively require reimplementing ClauseAdapter
-        replaced = sa.sql.util.ClauseAdapter(table_set).traverse(result)
-        num_froms = len(replaced.get_final_froms())
-        assert num_froms == 1, f"num_froms == {num_froms:d}"
-        return replaced
```
gforsyth, line 255: I am happy to see this go!

ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -81,24 +81,24 @@ def get_sqla_table(ctx, table):
...
     except KeyError:
-        return sa.column(colname)
+        # XXX: cols is a sqlalchemy column collection which contains column
+        # names that are secretly prefixed by their containing table
+        #
+        # sqlalchemy doesn't let you select by the *un*prefixed column name
+        # despite the uniqueness of `colname`
+        #
+        # however, in ibis we have already deduplicated column names so we can
+        # refer to the name by position
```
gforsyth, line 99:
```suggestion
        # cols is a sqlalchemy column collection which contains column
        # names that are secretly prefixed by their containing table
        #
        # sqlalchemy doesn't let you select by the *un*prefixed column name
        # despite the uniqueness of `colname`
        #
        # however, in ibis we have already deduplicated column names so we can
        # refer to the name by position
```

The micro-ist of nits, but to me `# XXX` reads like something that needs fixing before merge, but that is not a blocker at all.

gforsyth (approved):
This is a nice set of changes.

cpcloud:
Turns out this was even more broken in the impala backend. I've got a fix that I'll push up if the test suite passes locally.

ibis/backends/base/sql/alchemy/query_builder.py
cpcloud: Good riddance!

ibis/backends/base/sql/alchemy/registry.py
cpcloud: Fixed!

cpcloud:
Ok, so to fix this I had to completely eliminate our alias eliding, which makes the SQL slightly worse. I don't think that matters here since it is now correct and verbose whereas before it was incorrect and concise.

cpcloud:
Actually, to make this less annoying to get in, I'm going to mark the stuff I just fixed as broken and do an immediate follow up.

PR #5219, opened 2023-01-13
fix(impala): remove broken alias elision

cpcloud:
This PR is a follow-up to #5213. Here I fixed a bug with string backends where aliases were elided much too aggressively. Unfortunately this means the PR has a ton of snapshot files regenerated.


gforsyth (approved):
this is great.  I'll let you merge in case you want a few more eyes on this, but I think this is good to go.

PR #5221, opened 2023-01-13
docs: embed streamlit operation matrix app to docs

cpcloud:
This PR embeds the streamlit backend operation matrix app in our docs.


gforsyth:
I tried to build this locally to check out how it looks and I'm getting a permission error -- should this work on a local docs build?

gforsyth:
Apparently I forgot what URLs were for a second.

Anyway, I can't access https://ibis-project-ibis-docsbackend-info-app-k0skja.streamlit.app in browser directly, so I think we need to set some permissions somewhere.

cpcloud:
Yep, lemme fix that up.

gforsyth (approved):
:shipit:

jcrist:
This doesn't seem to render properly on firefox (I'm on version 108.0.1 if that matters). I checked chrome and that looks fine, but on firefox the app window is super tiny.

image

cpcloud:
Hm, lemme see if I can reproduce

.streamlit/config.toml
```diff
@@ -0,0 +1,5 @@
+[browser]
+gatherUsageStats = false
+
+[theme]
+base = "dark"
```
krzysztof-kwitt, line 5: It seems to me that this file must be in the directory with the application i.e. `docs/backends/app/.streaamlitt/config.toml`

cpcloud: I read the docs as saying it must be in the directory the application is run from, which I assumed to be the repo:

image

PR #5225, opened 2023-01-13
docs: clean up gen_matrix.py after adding streamlit app

krzysztof-kwitt:
Hello,

Now this script can be much simpler.


cpcloud (approved):
Ah, nice!

PR #5234, opened 2023-01-14
depr(api): deprecate table `get_column`/`s` in favor of `__getitem__`

cpcloud:
Deprecate `get_column`-style API in favor of `col`/`cols`. Open to other naming suggestions!

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -211,7 +211,9 @@ class Table(Expr, JupyterMixin):
             return expr
 
     @util.deprecated(
-        version="5.0", instead="use a list comprehension and attribute/getitem syntax"
+        as_of="4.1",
+        removed_in="5.0",
+        instead="use a list comprehension and attribute/getitem syntax",
     )
     def get_columns(self, iterable: Iterable[str]) -> list[Column]:
         """Get multiple columns from the table.
@@ -234,7 +236,7 @@ class Table(Expr, JupyterMixin):
         """
         return list(map(self.get_column, iterable))
 
-    @util.deprecated(version="5.0", instead="use t.<name> or t[name]")
+    @util.deprecated(as_of="4.1", removed_in="5.0", instead="use t.<name> or t[name]")
     def get_column(self, name: str) -> Column:
         """Get a reference to a single column from the table.
 
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -115,7 +115,7 @@ class Table(Expr, JupyterMixin):
         from ibis.expr.types.logical import BooleanValue
 
         if isinstance(what, (str, int)):
-            return self.get_column(what)
+            return ops.TableColumn(self, what).to_expr()
 
         if isinstance(what, slice):
             step = what.step
@@ -154,7 +154,7 @@ class Table(Expr, JupyterMixin):
 
     def __getattr__(self, key):
         with contextlib.suppress(com.IbisTypeError):
-            return self.get_column(key)
+            return ops.TableColumn(self, key).to_expr()
 
         # Handle deprecated `groupby` and `sort_by` methods
         if key == "groupby":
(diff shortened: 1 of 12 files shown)


kszucs:
I don't have a strong opinion on this, the change itself looks good of course.

What APIs do other libraries provide for accessing columns?

cpcloud:
pandas: `__getitem__`/`__getattr__`
spark: `col("x")` (top level function)
sqlalchemy: `.c[...]`/`.c.<attr>` `.c` is an alias for `.columns`
polars: `col("x")` (top level function)
datafusion: `column` (top level function)

It's kind of all over the place.

jcrist:
Is there a strong reason to provide a method here rather than `__getitem__`/`__getattr__` (or `.select(...)` for `cols`)? I confess I didn't even know we had a `get_column`/`get_columns` method.

cpcloud:
@jcrist Not particularly. My thought was that `expr.col` is slightly less annoying to read and easier to discover if you want to pass it around versus `expr.__getitem__` or `expr.__getattr__`. No strong opinions on it though.

jcrist:
Yeah, I personally prefer:

```python
col = table.foo
col = table["foo"]
```

to

```python
col = table.col("foo")
```

Less typing, and doesn't add a new set of methods to `Table`  which prevents columns with those names from being accessed as attributes (although a column named `col` may not be a good idea anyway). The `col` method seems redundant for `__getitem__`/`__getattr__`, which are what pandas users will be used to using.

I'm not sure I understand the use case for `get_columns`/`cols` though. When might you want to extract a collection of `Column` objects rather than working with them on the parent table directly? Does that happen often enough that you need a special `get_columns`/`cols` method for it?

Anyway, not a blocker from me. My (slight) preference would be to deprecate `get_column`/`get_columns` in favor of using the existing `__getitem__`/`__getattr__` apis (and maybe a manual list comprehension for when `get_columns` is needed :shrug:).

cpcloud:
That seems fine to me, I'd rather get rid of them too.

kszucs:
> Anyway, not a blocker from me. My (slight) preference would be to deprecate `get_column`/`get_columns` in favor of using the existing `__getitem__`/`__getattr__` apis (and maybe a manual list comprehension for when `get_columns` is needed 🤷).

Agree.

ibis/expr/types/relations.py
```diff
@@ -210,33 +210,39 @@ def _ensure_expr(self, expr):
...
-        ...        ('a', 'int64'),
-        ...        ('b', 'string'),
-        ...        ('c', 'timestamp'),
-        ...        ('d', 'float'),
-        ...    ],
-        ...    name='t'
-        ... )
-        >>> a, b, c = table.get_columns(['a', 'b', 'c'])
+        >>> table = ibis.table(dict(a='int64', b='string', c='timestamp', d='float'))
+        >>> a, b, c = table.cols(['a', 'b', 'c'])
```
kszucs, line 228:
```suggestion
        >>> a, b, c = table.get_columns(['a', 'b', 'c'])
```

ibis/expr/types/relations.py
```diff
@@ -210,33 +210,39 @@ def _ensure_expr(self, expr):
         else:
             return expr
 
+    @util.deprecated(
+        version="5.0", instead="use a list comprehension and attribute/getitem syntax"
```
kszucs, line 214: Shouldn't this be `4.1`? I'm usually a bit confused the version argument here.

Perhaps we should have two arguments, one highlighting the version we deprecated the API in `asof` and one for marking the version it's going to be removed at?

`util.deprecated(asof="4.1", removed_in="5.0", instead=...)`

cpcloud: Good idea! I'll add those arguments.

cpcloud: For clarity, I've been treating `version` like your `removed_in`.

kszucs: We can use `removed_in` to raise if we forget to remove the API before the release.

cpcloud: How so?

kszucs (approved):
Nice!

PR #5240, opened 2023-01-15
Implement _get_schema_using_query in SQLite backend

salva:
As SQLite does not provide this functionality directly the code here creates a temporal view for the query and then uses `table_info` pragma for retrieving the schema.

A convoluted and changing named is used for the view in order to avoid collisions. The lifetime of SQLAlchemy SQLite backend connections is not documented so this seems to be the safest thing to do.

The code has been highly inspired by DuckDB backend code.

### DIFF
--- a/ibis/backends/sqlite/__init__.py
+++ b/ibis/backends/sqlite/__init__.py
@@ -22,6 +22,7 @@ from typing import TYPE_CHECKING
 
 import sqlalchemy as sa
 from sqlalchemy.dialects.sqlite import DATETIME, TIMESTAMP
+import toolz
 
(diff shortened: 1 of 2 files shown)


cpcloud:
@salva Thanks for the PR!

I took a look at it more closely and there's some refactoring that needs to be done to make this work.

I'll push up my changes to your branch, and we can hopefully merge this very soon!

cpcloud:
Hm, it looks like I can't push to your fork! I will open a new PR that preserves your commits.

cpcloud:
Closing in favor of #5242.

PR #5248, opened 2023-01-17
chore: remove reference to ibis-bigquery package

krzysztof-kwitt:
Hello,

I checked and we still had some references from ibis-bigquery which is no longer maintained.

Now we still have one reference, but it's only in tests, so it's okay.
```
git ls-files -z | xargs -0 grep 'ibis.bigquery' | grep -v "ibis\.bigquery\.\(connect\|compile\)"
docs/release_notes.md:* Move BigQuery backend to a `separate repository <https://github.com/ibis-project/ibis-bigquery>`_. The backend will be released separately, use `pip install ibis-bigquery` or `conda install ibis-bigquery` to install it, and then use as before. ([#2665](https://github.com/ibis-project/ibis/issues/2665))
ibis/backends/bigquery/tests/conftest.py:    # https://github.com/ibis-project/ibis-bigquery/issues/30
```

Best regards

### DIFF
--- a/ibis/__init__.py
+++ b/ibis/__init__.py
@@ -14,7 +14,7 @@ __all__ += api.__all__
 
 __version__ = "4.0.0"
 
-_KNOWN_BACKENDS = ['bigquery', 'heavyai']
+_KNOWN_BACKENDS = ['heavyai']
 
 
 def __dir__() -> list[str]:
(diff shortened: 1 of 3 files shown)


cpcloud (approved):
Thanks!

PR #5260, opened 2023-01-18
feat: add new `ibis.get_backend` function

### LINKED ISSUE #5064: feat(ux): expose `_find_backend` as a public method on TableExpr

Before `ibis.read` existed, there wasn't a way to get a table expression without either making a connection to a backend, or creating an unbound table.

Now though, it's possible to get a TableExpr from `read` without knowing which backend you are running on.  For now, this is almost always `duckdb`, but that may change.

I think we could provide a public version of `_find_backend` -- more akin to the `name` attribute on connections, so that new users can check what backend their table expression is running against.

master

_No response_

### Code of Conduct

jcrist:
This PR adds a new `ibis.get_backend` function for finding the current execution backend for an expression, or the default backend if no expression is provided.

```python
# returns the default backend
ibis.get_backend() 

# returns the backend for a given bound expression. An error is raised if unbound tables are
# present, or if multiple backends are found
ibis.get_backend(some_expr)
```

Also fixes a small bug where `to_pyarrow`/`to_pyarrow_batches` wouldn't fallback to using the default backend if no bound tables were found (while `.execute` would).

Fixes #5064.

### DIFF
--- a/docs/api/expressions/top_level.md
+++ b/docs/api/expressions/top_level.md
@@ -15,6 +15,7 @@ These methods and objects are available directly in the `ibis` module.
 ::: ibis.date
 ::: ibis.desc
 ::: ibis.difference
+::: ibis.get_backend
 ::: ibis.greatest
 ::: ibis.ifelse
 ::: ibis.intersect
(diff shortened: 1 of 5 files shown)


cpcloud (approved):
LGTM, nice!

PR #5261, opened 2023-01-19
fix(ir): error for `order_by` on nonexistent column

### LINKED ISSUE #5214: bug: order_by() with non-existent column does not error
### What happened?

I would expect order_by(DNE) to fail. But instead it silently gives me back the original unsorted table:

```python
import ibis
import pandas as pd

df = pd.DataFrame({"x": [3,1,2]})
t = ibis.memtable(df)
t.order_by("bogus")
┏━━━━━━━┓
┃ x     ┃
┡━━━━━━━┩
│ int64 │
├───────┤
│     3 │
│     1 │
│     2 │
└───────┘
```

4.0.0

duckdb

### Relevant log output

_No response_

### Code of Conduct

jcrist:
Fixes #5214

### DIFF
--- a/ibis/expr/rules.py
+++ b/ibis/expr/rules.py
@@ -134,11 +134,16 @@ def sort_key_from(table_ref, key, **kwargs):
     else:
         key, order = key, True
 
-    key = one_of(
-        (function_of(table_ref), column_from(table_ref), any),
-        key,
-        **kwargs,
-    )
+    if isinstance(key, (str, int)):
+        # Actual str/int keys must refer to columns in the table, we don't want
+        # to fallback to converting them to expressions with ibis.literal
+        key = column_from(table_ref, key, **kwargs)
+    else:
+        key = one_of(
+            (function_of(table_ref), column_from(table_ref), any),
+            key,
+            **kwargs,
+        )
 
     if isinstance(order, str):
         order = order.lower()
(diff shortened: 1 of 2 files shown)


ibis/expr/rules.py
```diff
@@ -134,11 +134,16 @@ def sort_key_from(table_ref, key, **kwargs):
...
-        key,
-        **kwargs,
-    )
+    if isinstance(key, (str, int)):
+        # Actual str/int keys must refer to columns in the table, we don't want
+        # to fallback to converting them to expressions with ibis.literal
+        key = column_from(table_ref, key, **kwargs)
+    else:
+        key = one_of(
+            (function_of(table_ref), column_from(table_ref), any),
```
kszucs, line 143: How about removing `any`?

cpcloud: It's still legitimate to order by a constant, despite likely limited utility. IMO we shouldn't make that judgment call (yet).

kszucs: What is the point of ordering by a constant (at least in ibis land)? Ordering by a function like `RAND()` would make sense, but we can add a rule for that.

cpcloud: Well, in one case, it's a workaround for snowflake's hard requirement for an `ORDER BY` in some window functions (even where techincally it isn't required).

I don't think we should make this choice on behalf of users right now. Maybe at some point we will, but it's also not causing any issues other than this one.

kszucs (approved):

ibis/expr/rules.py
kszucs: I'm going to merge it, though it would be nice to better formalize sort key requirements in the future. It also applies to grouping keys and join keys which we are handling non uniformly.

PR #5268, opened 2023-01-20
feat: make Backend.tables.__repr__ look like list

NickCrews:
In a notebook I want to see the names of the tables in the database.
Before I would get
'<ibis.backends.base.TablesAccessor object at 0x118321c90>'

Now I get

```
[
    "df",
    "df0",
    "df1",
    "df10",
    "df11",
    "df12"
]
```

I considered just doing `return repr(list(self))`,
but I wanted the tables to be in sorted order,
and I wanted them to be indented so they'd be
easier to scan with your eyes.

The test is a bit lax, but I wanted it to be future-proof.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -204,6 +204,12 @@ class TablesAccessor(collections.abc.Mapping):
         )
         return list(o)
 
+    def __repr__(self) -> str:
+        tables = self._backend.list_tables()
+        rows = ["Tables", "------"]
+        rows.extend(f"- {name}" for name in sorted(tables))
+        return "\n".join(rows)
+
     def _ipython_key_completions_(self) -> list[str]:
         return self._backend.list_tables()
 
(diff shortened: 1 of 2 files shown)


cpcloud:
@NickCrews Thanks for the PR!

One thing that may make this hard for some users to grok is that the resulting object looks like a `list` and they will expect it to behave like one.

What do you think about Attaching `self.__class__.__name__` on a line above the `list`-ified output?

NickCrews:
Yeah that sounds good, I can update in the morning. Do you have a
preference on force pushing new commits vs adding on fixup commits?

NickCrews:
Now it looks like

```
TablesAccessor([
    "df0",
    "df1",
    "df10",
    "df11"
])
```

jcrist:
Thanks @NickCrews! After discussing this internally, we decided a markdown-list-like repr might be easier to read and make more sense. Since we're hoping to get this release out shortly, I hope you don't mind that I pushed these changes up to your branch! Thanks for the fix, a good repr for `con.tables` is definitely a nice UX improvement.

NickCrews:
Yeah that looks fine, I don't really care :)

jcrist (approved):

PR #5271, opened 2023-01-20
chore(dev): stop and start containers with a single command

cpcloud:
This PR adds a convenience command `just reup [container]` that makes it more concise to restart one or more services.


kszucs (approved):
Nice!

PR #5273, opened 2023-01-20
feat(postgres): add `Map(string, string)` support via the built-in `HSTORE` extension

cpcloud:
This PR adds support for a subset of map types supported by postgres via the [`HSTORE`](https://www.postgresql.org/docs/current/hstore.html) type.


kszucs (approved):
Nice!

PR #5274, opened 2023-01-20
chore(ux): let the transaction error bubble up to avoid nested useless tracebacks

cpcloud:
This PR removes `finally:` block that tries to reset the session time zone. This is mostly useless since the entire block is happening inside of a transaction and will be rolleed back if an exception is thrown.


cpcloud:
Well, at least we know it's working as intended!

cpcloud:
I'm going to merge this so that we can get better tracebacks and I can debug the flaky temp view creation.

PR #5275, opened 2023-01-20
feat(polars): allow ibis.polars.connect() to function without any passed arguments

ksuarez1423:
Adds functionality to Polars backend so `ibis.polars.connect()` yields a Polars connection with nothing in it.

### DIFF
--- a/ibis/backends/polars/__init__.py
+++ b/ibis/backends/polars/__init__.py
@@ -34,7 +34,9 @@ class Backend(BaseBackend):
         super().__init__(*args, **kwargs)
         self._tables = dict()
 
-    def do_connect(self, tables: MutableMapping[str, pl.LazyFrame]) -> None:
+    def do_connect(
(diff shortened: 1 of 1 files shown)


ibis/backends/polars/__init__.py
```diff
@@ -34,14 +34,16 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._tables = dict()
 
-    def do_connect(self, tables: MutableMapping[str, pl.LazyFrame]) -> None:
+    def do_connect(self, tables: MutableMapping[str, pl.LazyFrame] = None) -> None:
```
jcrist, line 37:
```suggestion
    def do_connect(self, tables: MutableMapping[str, pl.LazyFrame] | None = None) -> None:
```

jcrist (commented):
One small fix, otherwise this lgtm.

jcrist (approved):

PR #5276, opened 2023-01-20
test(dot-sql): make `dot_sql` tests less flaky

cpcloud:
This PR seems to fix the current flakiness in `test_dot_sql.py`.

While I'm not 100% sure it does, I was able to reproduce the [current flakiness](https://github.com/ibis-project/ibis/actions/runs/3968820831/jobs/6802496984)
by running the tests locally like so:

```
pytest -n 4 -m postgres -x --dist=loadgroup
```

When I run the tests without `--dist=loadgroup` I cannot reproduce the issue. This isn't viable generally because 
some backends cannot run queries across processes (which is the concurrency model for `pytest-xdist`).

This change moves the `xdist_group` marker to decorate each test instead of being a module level marker, which doesn't seem to work.

Things I tried:

1. A larger number of processes seems cause fewer failures, much to my surprise.
2. Running the `dot_sql` tests in isolation (either alone or running all the tests from `test_dot_sql.py` doesn't fail

I am going to keep running the tests locally to see if I can reproduce it and will report back either way.

### DIFF
--- a/ibis/backends/postgres/__init__.py
+++ b/ibis/backends/postgres/__init__.py
@@ -176,7 +176,7 @@ class Backend(BaseAlchemyBackend):
     def _metadata(self, query: str) -> Iterable[tuple[str, dt.DataType]]:
         raw_name = util.guid()
         name = self._quote(raw_name)
-        type_info_sql = f"""\
+        type_info_sql = """\
 SELECT
   attname,
   format_type(atttypid, atttypmod) AS type
@@ -194,8 +194,6 @@ ORDER BY attnum"""
             con.execute(sa.text(f"DROP VIEW IF EXISTS {name}"))
 
     def _get_temp_view_definition(
-        self,
-        name: str,
-        definition: sa.sql.compiler.Compiled,
+        self, name: str, definition: sa.sql.compiler.Compiled
     ) -> str:
         yield f"CREATE OR REPLACE VIEW {name} AS {definition}"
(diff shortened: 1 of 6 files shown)


cpcloud:
My initial hypothesis turned out to be wrong.

This problem was due to my incorrect understanding of how the lifetime of
temporary views interacts with SQLAlchemy's connection pooling.

Until we're able to refactor this to remove views entirely, I've changed
the postgres backend to use standard views, that are persisted in the database
and visible to all connections.

ibis/backends/postgres/__init__.py
```diff
@@ -197,4 +198,4 @@ def _get_temp_view_definition(
         name: str,
         definition: sa.sql.compiler.Compiled,
     ) -> str:
-        yield f"CREATE OR REPLACE TEMPORARY VIEW {name} AS {definition}"
+        yield f"CREATE OR REPLACE VIEW {name} AS {definition}"
```
jcrist, line 201: Are these ever dropped (and if so, where?)

cpcloud: Yep, they are dropped in an atexit handler in the base class.

cpcloud: https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/alchemy/__init__.py#L588 is where this method is defined.

It's called just below that here: https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/alchemy/__init__.py#L622

`_create_temp_view` is called here: https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/alchemy/query_builder.py#L114

cpcloud:
Merging, so that we can get back to a more reliably passing `master` branch.

PR #5284, opened 2023-01-21
refactor(snowflake): remove unnecessary `parse_json` call in `ops.StructField` impl

cpcloud:
This PR removes an unnecessary `parse_json` call for the Snowflake backend implementation of StructField.


kszucs (approved):
LGTM

PR #5288, opened 2023-01-21
refactor(datatype): use a dictionary to store `StructType` fields rather than `names` and `types` tuples

kszucs:
BREAKING CHANGE: Struct(names, types) is not supported anymore, pass a dictionary to Struct constructor instead

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -14,10 +14,10 @@ from ibis.common.exceptions import IbisTypeError
 from ibis.common.grounds import Concrete, Singleton
 from ibis.common.validators import (
     all_of,
+    frozendict_of,
     instance_of,
     isin,
     map_to,
-    tuple_of,
     validator,
 )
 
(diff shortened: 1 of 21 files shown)


ibis/expr/datatypes/core.py
```diff
@@ -673,47 +665,22 @@ def from_tuples(
         Struct
             Struct data type instance
         """
-        names, types = zip(*pairs)
-        return cls(names, types, nullable=nullable)
-
-    @classmethod
-    def from_dict(
```
kszucs, line 680: Shall we deprecate these as of `4.1`?

ibis/expr/datatypes/value.py
```diff
@@ -32,12 +32,15 @@ def infer(value: Any) -> dt.DataType:
     raise InputTypeError(value)
 
 
+# TODO(kszucs): support NamedTuples and dataclasses instead of OrderedDict
```
kszucs, line 35: Create a follow-up issue for this.

ibis/expr/datatypes/core.py
```diff
@@ -642,19 +642,11 @@ def to_integer_type(self):
 class Struct(DataType):
```
kszucs, line 642: We could also implement `Mapping` abc here, though I'm not sure whether that's desired.

ibis/expr/datatypes/core.py
cpcloud: Sounds good to me!

ibis/expr/datatypes/core.py
cpcloud: It would be nice to have `keys()` and `values()` instead of `names` and `types` or whatever it is that we implement now.

cpcloud (approved):
LGTM!

PR #5289, opened 2023-01-21
chore: rewrite streamlit app to use Ibis

krzysztof-kwitt:
Hello,

To show how Ibis can be used with Streamlit, I rewrote our application to use Ibis.

Best regards

### DIFF
--- a/docs/backends/app/backend_info_app.py
+++ b/docs/backends/app/backend_info_app.py
@@ -1,98 +1,131 @@
-import contextlib
-import csv
-from enum import Enum
+import datetime
 from pathlib import Path
+from typing import List, Optional
 
 import pandas as pd
-import requests
+import sqlglot
 import streamlit as st
+
+import ibis
+from ibis import _
+from ibis.backends import duckdb
+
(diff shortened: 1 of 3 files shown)


docs/backends/app/backend_info_app.py
```diff
@@ -133,9 +180,19 @@ def get_selected_operation_categories(all_ops_categories):
...
 
     table = pd.concat([coverage, df.replace({True: "✔", False: "🚫"})])
-    st.table(table)
+    st.dataframe(table)
 else:
     st.write("No data")
 
+with st.expander("SQL queries"):
+    for sql_query in sql_queries:
+        pretty_sql_query = sqlglot.transpile(
```
cpcloud, line 189: You should be able to `ibis.to_sql` here. Does that not work?

cpcloud: Oh, I see. Nevermind!

cpcloud:
Sweet, looks great 👍 I tried it out locally.

Do we need to update the requirements to include ibis-framework as a dependency?

cpcloud:
Added the `ibis-framework[duckdb]` dependency. Let's see how it goes on the deploy.

PR #5294, opened 2023-01-22
refactor(schema): use a dictionary to store `Schema` fields rather than names and types tuples

kszucs:
BREAKING CHANGE: Schema(names, types) is not supported anymore, pass a dictionary to Schema constructor instead or use ibis.schema(names, types)

depends on #5288 

there are weird SQL compilation failures, which require investigation before mering


ibis/tests/sql/snapshots/test_compiler/test_table_drop_with_filter/out.sql
```diff
@@ -1,18 +1,18 @@
...
-      ) t4
-      WHERE t4.`C` = '2018-01-01T00:00:00'
-    ) t3
-  ) t1
-    INNER JOIN s t2
-      ON t1.`b` = t2.`b`
-) t0
-WHERE t0.`a` < 1.0
\ No newline at end of file
+WITH t0 AS (
```
kszucs, line 1: This actually looks better.

ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -100,7 +100,7 @@ def get_col(sa_table, op: ops.TableColumn) -> sa.sql.ColumnClause:
     #
     # however, in ibis we have already deduplicated column names so we can
     # refer to the name by position
-    colindex = op.table.schema._name_locs[colname]
+    colindex = op.table.schema.names.index(colname)
```
cpcloud, line 103: The point of using `_name_locs` here was to avoid the `O(n)` look up cost. Can we keep `_name_locs` around?

ibis/tests/sql/snapshots/test_compiler/test_table_drop_with_filter/out.sql
cpcloud: Why does this change as a result of changing the internal schema representation?

kszucs:
Superceded by #5323

PR #5297, opened 2023-01-23
ci: add `removed_in` release verification script

cpcloud:
This PR adds a script to the [`verifyRelease`](https://semantic-release.gitbook.io/semantic-release/#release-steps) step of our release process, that checks whether all `removed_in` decorated functions have a value whose version is greater than that of the next release.


cpcloud:
It's a bit buried in the output, but you can [see the script running successfully](https://github.com/ibis-project/ibis/actions/runs/3985682174/jobs/6833360180#step:6:570).

kszucs:
Do we run the tests after version bumping? If yes it could be easier to just raise from the decorator if we already passed the removed_in version.

ci/release/verify_release.sh
```diff
@@ -0,0 +1,12 @@
...
+set -euo pipefail
+
+next_version="${1}"
+
+# check that the release removed_in arguments are in a later version than the one being released
+quote="(?:\"|')"
+verpat="\d(?:\.\d){0,2}"
+nix run 'nixpkgs#ripgrep' -- "removed_in=${quote}${verpat}${quote}" "$PWD" --type py --json |
+  nix run 'nixpkgs#jq' -- 'select(.type == "match").data' -rcM |
+  python ci/release/verify_release.py "$next_version" -
```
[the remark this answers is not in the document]
cpcloud: No.

`verify.sh` happens before the release version is computed. This phase is called `verifyConditions`. This is a phase to run verification steps that don't require knowledge of what the next version is.


`verify_release.sh` runs *after* the release version is computed, but *before* any tagging, artifact creation or file modification. This phase is called `verifyRelease`.

cpcloud:
> Do we run the tests after version bumping? If yes it could be easier to just raise from the decorator if we already passed the removed_in version.
> 
> Or you intentionally would like to keep it soft-failing?

I don't think I would say this is soft-failing.

The release dry-run will fail in CI long before any release is ever made if any `removed_in`s are incompatible with the next release.

ci/release/verify_release.sh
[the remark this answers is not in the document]
cpcloud: Ok, I can rename `verify.sh` to `verify_conditions.sh`. I would like to keep the scripts named to match the phase names. The order in which phases are run is not something I think makes sense to bake into the script naming.

cpcloud:
I'm going to take a slightly different approach here that should be more robust and require fewer command line tools.

cpcloud:
The postgres tests occasionally fail in this way (one or more of test processes trying to clobber the `ibis_testing` database while it's in use).

I'm not sure what the issue is, but we can probably avoid dropping the database.

ci/release/verify_release.py
```diff
@@ -0,0 +1,90 @@
...
+    visitor.visit(node)
+    return [
+        (path, removed_in)
+        for removed_in in visitor.removed_ins
+        if removed_in.version <= next_ver
+    ]
+
+
+def main(next_ver: v.Version, root: str) -> int:
+    pyfiles = root.rglob("*.py")
```
gforsyth, line 58:
```suggestion
def main(next_ver: v.Version, root: Path) -> int:
    pyfiles = root.rglob("*.py")
```
if you're calling `rglob` this can only be a `pathlib.Path`, I think

pyproject.toml
```diff
@@ -396,10 +396,12 @@ fix = true
 # don't delete print statements
 # don't remove unused imports
 # don't remove unused noqa
-unfixable = ["T201", "F401", "RUF100"]
```
gforsyth, line 399: Nothing to change here right now, but if we have too many more ignored errors here, we'll probably want to have one item per line with inline comments

gforsyth (changes requested):
small nit, but overall this looks good and it's nice to have the check

pyproject.toml
cpcloud: Hm, let me remove the comments. I turned off `fix = true` for now because it's a bit too aggressive while developing. It's turned on for `pre-commit` though which I think is probably the appropriate time to auto fix things.

cpcloud: Oh, whoops I misread this because the comment only shows up on the change line and no the addition line.

PR #5298, opened 2023-01-23
depr(datatype): schedule `Struct.from_dict()`, `Struct.pairs` and `Struct(names, types)` constructor for removal

kszucs:
These are going to be removed in `5.0` via #5288 

I ended up applying most of the changes of the previous PR in a backward compatible fashion.

### DIFF
--- a/ibis/backends/base/sql/alchemy/datatypes.py
+++ b/ibis/backends/base/sql/alchemy/datatypes.py
@@ -175,7 +175,7 @@ def _pg_array(dialect, itype):
 @to_sqla_type.register(Dialect, dt.Struct)
 def _struct(dialect, itype):
     return StructType(
-        [(name, to_sqla_type(dialect, type)) for name, type in itype.pairs.items()]
+        [(name, to_sqla_type(dialect, type)) for name, type in itype.fields.items()]
     )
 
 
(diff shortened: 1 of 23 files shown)


ibis/backends/base/sql/alchemy/datatypes.py
```diff
@@ -175,7 +175,7 @@ def _pg_array(dialect, itype):
 @to_sqla_type.register(Dialect, dt.Struct)
 def _struct(dialect, itype):
     return StructType(
-        [(name, to_sqla_type(dialect, type)) for name, type in itype.pairs.items()]
+        [(name, to_sqla_type(dialect, type)) for name, type in itype.fields.items()]
```
cpcloud, line 178: Can we get rid of `fields` entirely and rely on the `Mapping` interface? Doing that can definitely be a follow-up PR.

cpcloud (approved):
This LGTM. One small question, not a blocker.

ibis/backends/base/sql/alchemy/datatypes.py
kszucs: I wasn't entirely sure whether we want that, but we already provide most of the mapping API for schema, so I guess we should do the same with StructType. Going to do that in a follow-up.

PR #5300, opened 2023-01-23
refactor(duckdb): use sqlalchemy-views to reduce string hacking

cpcloud:
This PR cleans up some of our SQL string creation in the DuckDB backend using [`sqlalchemy-views`](https://pypi.org/project/sqlalchemy-views/), a nice package for integrating view creation into SQLAlchemy.


pyproject.toml
```diff
@@ -169,7 +170,13 @@ bigquery = [
 clickhouse = ["clickhouse-driver", "clickhouse-cityhash", "lz4"]
 dask = ["dask", "pyarrow", "regex"]
 datafusion = ["datafusion"]
-duckdb = ["duckdb", "duckdb-engine", "pyarrow", "sqlalchemy"]
+duckdb = [
+  "duckdb",
+  "duckdb-engine",
+  "pyarrow",
+  "sqlalchemy",
+  "sqlalchemy-views"
```
krzysztof-kwitt, line 178: FYI: The `sqlalchemy-views` library does not support `CREATE TEMP VIEW`, because we may have a problem to use this for snowflake backend.

cpcloud: True, but we're only using that in the duckdb backend to implement the `Table.sql` method.

cpcloud: It doesn't look like [`CREATE TEMP VIEW`](https://docs.snowflake.com/en/sql-reference/sql/create-view.html) exists for snowflake. Am I reading that correctly?

krzysztof-kwitt: This is supported but covered by documentation.
Screenshot 2023-01-23 at 14 03 46
.

I found out about this feature while checking the Snowpark code.
https://github.com/snowflakedb/snowpark-python/blob/c806a03e36fe683511b6e90baf758e42ba4fb9a6/src/snowflake/snowpark/dataframe.py#L2883-L2889

For now, we only use this in duckdb, but it's worth making the backends share as much code as possible.

krzysztof-kwitt: PostgresSQL support TEMP modifier.
https://www.postgresql.org/docs/current/sql-createview.html

cpcloud: @krzysztof-kwitt Did you have a use case in mind for `TEMP VIEW`? Want to make sure I'm not dismissing anything!

cpcloud: Ah, I see. Code sharing, makes sense. I think in this case I'm using `sqlalchemy-views` to create permanent views and clean up all the string interpolation we're currently doing. We  may be able to share some code in the future, but for now I think we don't need to worry too much about lack of support for `TEMP` just yet IMO.

krzysztof-kwitt: I don't have any, but I have mixed feelings about the use of `atexit` because the developer doesn't have much control over it.  In the case of Snowpark, the objects exist as long as the snowpark session exists, and the snowpark session can be closed explicitly or when the script ends.

cpcloud: Yeah, I don't like the `atexit` handler either 😄. I think there's a way we can do the whole `.sql` method using common table expressions and a notion of setup queries (which exists but isn't heavily used except in the BigQuery backend) and avoid pooping any state around the user's database.

gforsyth (approved):
this is nice!  Much better than trying to stare really hard at a bunch of nested single- and double-quotes

Mause:
Is the expectation here that users explicitly install the sqlalchemy_views package @cpcloud?

cpcloud:
Hey no worries! Thanks for checking it out.

cpcloud:
Your question did make me remember to add `sqlalchemy-views` to the `all` extra though! PR-ing that now

PR #5303, opened 2023-01-23
chore(dev): don't try to drop databases that are automatically created

cpcloud:
This PR tries to address an issue observed in [this comment](https://github.com/ibis-project/ibis/pull/5297#issuecomment-1400400732) where the ibis_testing database is trying to be dropped while another process is using it. The approach here is to simply not try to drop databases that are created by the container on startup.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -125,7 +125,7 @@ services:
       retries: 3
       test:
         - CMD-SHELL
-        - /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "DROP DATABASE IF EXISTS [ibis_testing]; CREATE DATABASE [ibis_testing]"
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -125,7 +125,7 @@ services:
       retries: 3
       test:
         - CMD-SHELL
-        - /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "DROP DATABASE IF EXISTS [ibis_testing]; CREATE DATABASE [ibis_testing]"
+        - /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "IF DB_ID('ibis_testing') IS NULL BEGIN CREATE DATABASE [ibis_testing] END"
```
cpcloud, line 128: This command runs periodically, which would lead to strange connection issues. Here, it's still periodic but once the `ibis_testing` database is created it's no longer recreated when the healthcheck runs.

gforsyth (commented):
This fix makes sense to me -- is there anything to do besides merge and then see if the issue stops popping up?

cpcloud:
Yep, let's merge and see how it goes!

PR #5307, opened 2023-01-24
feat(api): add selectors for easier selection of columns

cpcloud:
Convenient columnn selectors, shamelessly cribbed from https://dplyr.tidyverse.org/reference/select.html.

### DIFF
--- a/ibis/expr/selectors.py
+++ b/ibis/expr/selectors.py
@@ -1,0 +1,120 @@
+"""Construct column selectors."""
+
+from __future__ import annotations
+
+import re
+from typing import Callable, Iterable, Sequence
+
+import ibis.expr.datatypes as dt
+import ibis.expr.types as ir
+from ibis import util
+
+
+class Selector:
+    __slots__ = ("predicate",)
+
+    def __init__(self, predicate: Callable[[ir.Column], bool]) -> None:
+        """Construct a `Selector` with `predicate`."""
+        self.predicate = predicate
+
+    def expand(self, table: ir.Table) -> Sequence[ir.Column]:
+        """Evaluate `self.predicate` on every column of `table`."""
+        return [col for column in table.columns if self.predicate(col := table[column])]
+
+    def __and__(self, other: Selector) -> Selector:
+        """Compute the conjunction of two `Selectors`."""
+        return self.__class__(lambda col: self.predicate(col) and other.predicate(col))
+
+    def __or__(self, other: Selector) -> Selector:
+        """Compute the disjunction of two `Selectors`."""
+        return self.__class__(lambda col: self.predicate(col) or other.predicate(col))
+
+    def __invert__(self) -> Selector:
+        """Compute the logical negation of two `Selectors`."""
+        return self.__class__(lambda col: not self.predicate(col))
+
+
+def where(predicate: Callable[[ir.Value], bool]) -> Selector:
+    """Return columns that satisfy `predicate`.
+
+    Examples
+    --------
+    >>> t = ibis.table(dict(a="float32"), name="t")
+    >>> t.select(s.where(lambda col: col.get_name() == "a"))
+    r0 := UnboundTable: t
+      a float32
+
+    Selection[r0]
+      selections:
+        a: r0.a
+    """
+    return Selector(predicate)
+
+
+def numeric() -> Selector:
+    """Return numeric columns.
+
+    Examples
+    --------
+    >>> import ibis.selectors as s
+    >>> t = ibis.table(dict(a="int", b="string", c="array<string>"), name="t")
+    >>> t
+    r0 := UnboundTable: t
+      a int64
+      b string
+      c array<string>
+    >>> t.select(s.numeric())  # `a` has integer type, so it's numeric
+    r0 := UnboundTable: t
+      a int64
+      b string
+      c array<string>
+
+    Selection[r0]
+      selections:
+        a: r0.a
+    """
+    return Selector(lambda col: col.type().is_numeric())
+
+
+def of_type(dtype: dt.DataType | str | type[dt.DataType]) -> Selector:
+    """Select columns of type `dtype`."""
+    if isinstance(dtype, type):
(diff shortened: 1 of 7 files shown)


kszucs:
Nice! Are you still working on this?

cpcloud:
@kszucs Just finished up the first iteration of the docs. It's ready for review now.

jreback:
pretty neat @cpcloud

ibis/expr/selectors.py
```diff
@@ -0,0 +1,286 @@
...
+from typing import Callable, Iterable, Sequence
+
+from public import public
+
+import ibis.expr.datatypes as dt
+import ibis.expr.types as ir
+from ibis import util
+
+
+class Selector:
```
kszucs, line 56: Could we subclass deferred here? I mean a deferred base to not provide all of the overloads. 

Mostly just asking, shouldn't block this PR.

kszucs (approved):
Nice API addition!

ibis/expr/selectors.py
cpcloud: Maybe, but I didn't think much about what the semantics of the other overloads would be. It did cross my mind that maybe there's some shared code between this and `Deferred` but I didn't want to do a bunch of refactoring here.

krzysztof-kwitt:
I wonder if this API will be easy to use in real world., so I'm looking at the Ibis code that used column groups. Is it possible to rewrite this piece of code to use this new API? 
https://github.com/ibis-project/ibis/blob/318179add952a3b2fc80a1b6bf82191dc84b5157/docs/backends/app/backend_info_app.py#L155-L158

cpcloud:
> I wonder if this API will be easy to use in real world., so I'm looking at the Ibis code that used column groups. Is it possible to rewrite this piece of code to use this new API?
> 
> https://github.com/ibis-project/ibis/blob/318179add952a3b2fc80a1b6bf82191dc84b5157/docs/backends/app/backend_info_app.py#L155-L158

I don't think so, because that bit of code isn't filtering the columns by name or property in any way, it's looking at every column in the table.

mik-laj:
Can we use this API to calculate the number of boolean columns and count the number of True values?

cpcloud:
You could do something like:

```python
sum(map(t.__getitem__, t.select(s.of_type("bool")).columns))
```

cpcloud:
I wonder if it makes sense to define `TableExpr.__iter__` to iterate over the column objects, which which make the computation much shorter:

```python
sum(t.select(s.of_type("bool")))
```

cpcloud:
I know some systems are starting to allow concise syntax for doing whole-table operations. Maybe we can revisit adding aggregations and some kind of `apply expression to every column` method, so you can do things like:

```python
t.select(s.of_type("bool")).cast("int").sum()  # single row table, cast applies to every column
```

cpcloud:
@mik-laj You might find https://dplyr.tidyverse.org/articles/colwise.html interesting. Those APIs are perhaps a good place to start experimenting with APIs for summarizing things across a subset of the columns in a table.

I can imagine something like:

```python
# group by a, b and count all string columns excluding a and b
t.group_by([_.a, _.b]).summarize(s.across(s.of_type("string") & !s.contains(("a", "b")), [_.count()]))
```

jcmkk3:
> @mik-laj You might find https://dplyr.tidyverse.org/articles/colwise.html interesting. Those examples are perhaps a good place to start experimenting with APIs for summarizing things across a subset of the columns in a table.

It might also be worth taking a look at the [polars API](https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.col.html) for this. Basically, the `pl.col` selector allows selecting multiple columns in a few different ways, then when you perform methods on that selection it will apply to all columns that were selected. dplyr has to approach the problem a bit differently since the data transformations are functions instead of methods.

```python
# group by a, b and count all string columns excluding a and b
t.groupby(["a", "b"]).agg(pl.col(pl.Utf8).exclude(["a", "b"]).count())
```

cpcloud:
Neat, didn't know `pl.col` did all of that!

cpcloud:
Can you comment more on why the method/function distinction matters for anything other than than the API design (as opposed to being an implementation detail)?

I would expect that the fundamentals of the problem are unrelated to how a thing is called, e.g., a function vs a method on an object.

jcmkk3:
> Can you comment more on why the method/function distinction matters for anything other than than the API design (as opposed to being an implementation detail)?

I think that you're generally right in that it is a design constraint and also differences in style between the languages. R *can* do object oriented programming, but it is not the norm (especially in the tidyverse ecosystem).

The dplyr package doesn't own the compute functions like Ibis does so it makes a lot of sense for them to implement the `across` helper as a higher order function.

I personally think that something like the polars approach is more natural in python (minus the `Utf8` type that should probably be `Str`). It could still be a function like `across` that originally selects multiple columns, but the transformations/computations would be applied through method chaining just like working with a single column.

ksuarez1423:
While `across` seems to fulfill the need quite nicely, I agree that the Polars approach feels more natural in Python. When I went to try the selector with an aggregate, my instinct was to call `t.select(s.numeric()).min()`, because of its similarity to `t["col"].min()`. From my perspective, `pl.col()` matches that experience well enough for it to feel natural.

Table-wide operations also make sense from a user perspective, because it means if I have a fully numeric table, I could do `t.sum()` or even `t.all().sum()` if that would make code easier. Then, that makes `t.select(s.numeric()).sum()` a very easy counterpart -- I'd be able to fairly easily think, "oh, `t.select()` gets me a subset that's a table, so all the table-wide things work!"

cpcloud:
`pl.col` has a bit too much going on for my taste but I do agree that method chaining to build up the selectors is the way to go.

jcmkk3:
> `pl.col` has a bit too much going on for my taste but I do agree that method chaining to build up the selectors is the way to go.

I agree with this as well. I think that starting with a specific function for performing an operation on multiple columns would stand out more when scanning the code.

PR #5309, opened 2023-01-24
feat(duckdb): add `read_json` function for consuming newline-delimited JSON files

cpcloud:
This PR adds experimental support for reading newline delimited JSON files using duckdb.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -26,11 +26,12 @@ from ibis.backends.base.sql.alchemy import BaseAlchemyBackend
 from ibis.backends.duckdb.compiler import DuckDBSQLCompiler
 from ibis.backends.duckdb.datatypes import parse
 
-# counters for in-memory, parquet, and csv reads
+# counters for in-memory, parquet, csv, and json reads
 # used if no table name is specified
 pd_n = itertools.count(0)
 pa_n = itertools.count(0)
 csv_n = itertools.count(0)
+json_n = itertools.count(0)
 
 
 def normalize_filenames(source_list):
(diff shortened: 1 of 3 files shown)


ibis/expr/api.py
```diff
@@ -845,6 +846,31 @@ def read_csv(sources: str | Path, **kwargs: Any) -> ir.Table:
     return con.read_csv(sources, **kwargs)
 
 
+def read_json(sources: str | Path, **kwargs: Any) -> ir.Table:
```
gforsyth, line 849: do we want to mark this with `@experimental`?

cpcloud: Yes. Yes I do.

cpcloud: Done.

ibis/backends/duckdb/__init__.py
```diff
@@ -208,6 +209,55 @@ def _register_failure(self):
...
+        """Read newline-delimited JSON into an ibis table.
+
+        Parameters
+        ----------
+        source_list
+            File or list of files
+        table_name
+            Optional table name
+        kwargs
+            Additional keyword arguments passed to DuckDB's `read_json_objects` function
```
gforsyth, line 227: Does `read_json_objects` take any keyword arguments?  I can't find any.

cpcloud: It doesn't :)

cpcloud: Should I take this out or leave it for some future where there may or may not be kwargs?

gforsyth: I suppose it isn't doing any harm where it is, and if they add kwargs in the future it _may_ "just work"

ibis/backends/duckdb/__init__.py
```diff
@@ -208,6 +209,55 @@ def _register_failure(self):
...
+
+        objects = (
+            sa.func.read_json_objects(
+                sa.func.list_value(*source_list), _format_kwargs(kwargs)
+            )
+            .table_valued("raw")
+            .render_derived()
+        )
+        # read a single row out to get the schema, assumes the first row is representative
+        json_structure_query = sa.select(sa.func.json_structure(objects.c.raw)).limit(1)
```
jcrist, line 247: I tried this with the following JSON file and got an error from ibis

```
{"x": 1, "y": 2, "z": null}
{"x": 3, "y": 4, "z": 5}
```

Results in

```
>>> ibis.read_json("test.json")
...
ParseError: expected one of 'bigint', 'binary', 'blob', 'bool', 'boolean', 'bpchar', 'bytea', 'char', 'date', 'datetime', 'decimal', 'double', 'float', 'float4', 'float8', 'int', 'int1', 'int2', 'int4', 'int8', 'integer', 'interval', 'json', 'list', 'logical', 'long', 'map', 'numeric', 'real', 'short', 'signed', 'smallint', 'string', 'struct', 'text', 'time', 'timestamp with time zone', 'timestamp', 'timestamp_ms', 'timestamp_ns', 'timestamp_sec', 'timestamp_tz', 'tinyint', 'ubigint', 'uinteger', 'usmallint', 'utinyint', 'uuid', 'varbinary', 'varchar' at 0:0
```

The issue seems to be parsing `NULL` as a datatype doesn't work. However, fixing that results in silently dropping of column data, since duckdb will happily cast non-null data to `NULL`:

```
D select json_transform(json, '{"x": "UBIGINT", "y": "UBIGINT", "z": "NULL"}') FROM read_json_objects("test.json");
┌─────────────────────────────────────────────────────────────────────────┐
│ json_transform("json", '{"x": "UBIGINT", "y": "UBIGINT", "z": "NULL"}') │
│                 struct(x ubigint, y ubigint, z integer)                 │
├─────────────────────────────────────────────────────────────────────────┤
│ {'x': 1, 'y': 2, 'z': NULL}                                             │
│ {'x': 3, 'y': 4, 'z': NULL}                                             │
└─────────────────────────────────────────────────────────────────────────┘
```

- I wonder if we should select the first non-null value from every column and use that to infer the datatype?
- Alternatively, could/should we support an optional `schema` kwarg that takes an ibis schema for the expected table? This may be nice too, since it may allow for only extracting relevant fields from the input data.

jcrist (commented):
Overall this looks great, and the issue I mention below shouldn't stop this from getting in. We can always make improvements later, and it's marked as experimental as is :).

ibis/backends/duckdb/__init__.py
cpcloud: I think we probably want to do the latter, since the former has weird and annoying edge case behavior (all nulls except the last value).

jcrist: I have some test data I'd like to try this on, mind if I take a stab at improving the resiliency here in a follow up PR?

cpcloud: Heck yeah!

jcrist (approved):

gforsyth (approved):

PR #5315, opened 2023-01-24
feat: support `connect()` with no arguments for dask and pandas

jcrist:
A minor UX improvement, following up from #5275.


cpcloud (approved):
Thanks!

PR #5330, opened 2023-01-25
feat(datatype/schema): support datatype and schema declaration using type annotated classes

kszucs:
The point here is to support more declarative schema and table definitions, example:

```py
import datetime

import ibis


class FooSchema:
     a: int
     b: str
     c: float
     d: tuple[str]
     e: list[int]
     f: dict[str, int]
     g: datetime.date


class BarSchema:
     a: dt.int64
     b: dt.string
     c: dt.float64
     d: dt.Array(dt.string)
     e: dt.Array(dt.int64)
     f: dt.Map(dt.string, dt.string)
     g: dt.Date


class BazSchema:
     a: dt.Int64
     b: dt.String
     c: dt.Float64
     d: dt.Array[dt.String]
     e: dt.Array[dt.Int64]
     f: dt.Map[dt.String, dt.String]
     g: dt.Date


foo = ibis.table(name="foo", schema=FooSchema)
# perhaps a `table()` overload could be useful here to support `ibis.table(FooSchema)`
```

Nested types and annotations are also supported, tested for plain classes, dataclasses and namedtuples.

`NamedTuple` and various `dataclasses` are gaining tractions with more recent python versions so we should enable tighter integration without too much boilerplate.

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -1,17 +1,23 @@
 from __future__ import annotations
 
+import datetime as pydatetime
+import decimal as pydecimal
 import numbers
+import uuid as pyuuid
 from abc import abstractmethod
-from collections.abc import Iterator, Mapping
+from collections.abc import Iterator, Mapping, Sequence
+from collections.abc import Set as PySet
+from numbers import Integral, Real
 from typing import Any, Iterable, NamedTuple
 
 import numpy as np
+import toolz
 from multipledispatch import Dispatcher
 from public import public
+from typing_extensions import get_args, get_origin, get_type_hints
 
 import ibis.expr.types as ir
 from ibis.common.annotations import attribute, optional
-from ibis.common.exceptions import IbisTypeError
 from ibis.common.grounds import Concrete, Singleton
 from ibis.common.validators import (
     all_of,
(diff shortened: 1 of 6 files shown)


ibis/expr/datatypes/core.py
```diff
@@ -177,6 +224,9 @@ def is_null(self) -> bool:
     def is_numeric(self) -> bool:
         return isinstance(self, Numeric)
 
+    def is_parametric(self) -> bool:
```
cpcloud, line 227: Is this method used anywhere, or do you have a plan to use it somewhere?

ibis/expr/datatypes/core.py
```diff
@@ -239,10 +289,21 @@ class Primitive(DataType, Singleton):
     """Values with known size."""
 
 
+# TODO(kszucs): consider to remove since we don't actually use this information
 @public
 class Variadic(DataType):
     """Values with unknown size."""
 
+    pass
```
cpcloud, line 297:
```suggestion
```

cpcloud (commented):
LGTM, minor comment and a question about the purpose of `is_parametric`

ibis/expr/datatypes/core.py
kszucs: Removed.

ibis/expr/datatypes/core.py
kszucs: Indeed, we shouldn't expose it. Removed.

kszucs: Also made the Parametric class private to the module.

krzysztof-kwitt:
Should we add any kind of user-facing documentation? For nowe, we have only PR description.

kszucs:
Yes, quickly created a follow-up issue so we don't forget it https://github.com/ibis-project/ibis/issues/5410

PR #5333, opened 2023-01-25
fix: only include valid python identifiers in table/struct tab completion

jcrist:
Previously getattr tab completion would include things that weren't valid python identifiers (e.g. column names including spaces/python keywords).

Also adds getitem completion for structs, previously these only supported getattr completion.


cpcloud (approved):
Sweet, thanks!

PR #5341, opened 2023-01-26
refactor(sqlalchemy): split data type dispatches into backend specific modules

cpcloud:
This PR moves the respective backend data type dispatch to individual `backends/<backend>/datatypes.py` modules.

### DIFF
--- a/ibis/backends/base/sql/alchemy/datatypes.py
+++ b/ibis/backends/base/sql/alchemy/datatypes.py
@@ -3,16 +3,11 @@ from __future__ import annotations
 from typing import Iterable
 
 import sqlalchemy as sa
+import sqlalchemy.types as sat
 from multipledispatch import Dispatcher
-from sqlalchemy.dialects import mssql, mysql, postgresql, sqlite
-from sqlalchemy.dialects.mssql.base import MSDialect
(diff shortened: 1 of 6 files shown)


ibis/backends/base/sql/alchemy/datatypes.py
```diff
@@ -61,34 +56,46 @@ def compiles_map(element, compiler, **kw):
...
+@compiles(UInt16, "snowflake")
+@compiles(UInt8, "snowflake")
 @compiles(UInt64, "sqlite")
 @compiles(UInt32, "sqlite")
 @compiles(UInt16, "sqlite")
 @compiles(UInt8, "sqlite")
+@compiles(UInt64, "trino")
+@compiles(UInt32, "trino")
+@compiles(UInt16, "trino")
+@compiles(UInt8, "trino")
```
cpcloud, line 98: I kept these here to avoid repeating this code for each of the backends that doesn't support unsigned integer types.

There are no hidden dependencies here, the dialects are never imported if this function is never called.

kszucs (approved):
LGTM, much nicer now!

kszucs:
@cpcloud can you squash the smaller commits?

PR #5346, opened 2023-01-27
fix(parsy): remove unexpected keyword argument to `optional`

gforsyth:
The `default={}` is throwing errors with `parsy==1.3.0` and removing it
doesn't seem to cause problems with parsy 2.0


gforsyth:
wow, that broke everything.
shall we just bump up the minimum version of parsy?

cpcloud:
> shall we just bump up the minimum version of parsy?

Definitely

gforsyth:
> Definitely

Hmm, do we consider this a breaking change?

cpcloud:
Yeah, but since we're targeting 5.0 I think that's fine!

cpcloud (approved):

PR #5348, opened 2023-01-27
doc(backend-info): add version selector

krzysztof-kwitt:


docs/backends/app/backend_info_app.py
```diff
@@ -21,10 +22,67 @@
...
+    next_url = url
+    resources = []
+    headers = {
+        'X-GitHub-Api-Version': '2022-11-28',
+        'Accept': 'application/vnd.github+json',
+    }
+    # On Streamlit Cloud, token should be stored as secret.
+    # See:
+    # https://docs.streamlit.io/streamlit-community-cloud/get-started/deploy-an-app/connect-to-data-sources/secrets-management
+    github_token = os.environ.get('GITHUB_TOKEN')
```
cpcloud, line 42: What permissions does this token need?

docs/backends/app/backend_info_app.py
```diff
@@ -21,10 +22,67 @@
...
+@st.experimental_memo(ttl=ONE_HOUR_IN_SECONDS)
+def ibis_versions_with_raw_data_list():
+    releases = github_paginated_request(
+        'https://api.github.com/repos/ibis-project/ibis/releases'
+    )
+    versions = (release['name'] for release in releases)
+    versions = (
+        version
+        for version in versions
+        if version.startswith("4.") and version != '4.0.0'
```
cpcloud, line 63: What about using `packaging.version.parse`?

```python
from packaging.version import parse as vparse

vparse(version) >= vparse("4.1")
```

docs/backends/app/backend_info_app.py
```diff
@@ -21,10 +22,67 @@
...
+    return resources
+
+
+@st.experimental_memo(ttl=ONE_HOUR_IN_SECONDS)
+def ibis_versions_with_raw_data_list():
+    releases = github_paginated_request(
+        'https://api.github.com/repos/ibis-project/ibis/releases'
+    )
+    versions = (release['name'] for release in releases)
+    versions = (
```
cpcloud, line 60: Might as well make this into a list comprehension since you're immediately turning this into a list in the next line.

cpcloud:
@krzysztof-kwitt Can you address the comments here? This is pretty close to merge!

krzysztof-kwitt:
I don't know if we want to merge it because it conflicts with https://github.com/ibis-project/ibis/pull/5403 It is still valid?

docs/backends/app/backend_info_app.py
krzysztof-kwitt: I don't think any are needed because we're connecting to public repositories only.

cpcloud:
@krzysztof-kwitt Ok, let's close this out for now. Thanks!

PR #5352, opened 2023-01-28
Various literal improvements

krzysztof-kwitt:


ibis/backends/bigquery/registry.py
```diff
@@ -216,6 +216,20 @@ def _log(translator, op):
...
     dtype = op.output_dtype
+    if dtype.is_decimal():
+        value = op.value
+        if value.is_nan():
+            return "CAST('NaN' AS FLOAT64)"
+        if value.is_infinite():
+            if not value.is_signed():
+                return "CAST('inf' AS FLOAT64)"
+            else:
+                return "CAST('-inf' AS FLOAT64)"
```
cpcloud, line 227:
```suggestion
            prefix = "-" * value.is_signed()
            return f"CAST('{prefix}inf' AS FLOAT64)"
```

ibis/backends/duckdb/datatypes.py
```diff
@@ -139,3 +144,13 @@ def struct():
 )
 def parse_type(*args, **kwargs):
     return parse(*args, **kwargs)
+
+
+@to_sqla_type.register(DuckDBDialect, dt.UUID)
+def sa_duckdb_uuid(_, itype):
+    return postgresql.UUID
```
cpcloud, line 151: How about just returning the toplevel `sa.UUID` instead?

```suggestion
    return sa.UUID()
```

ibis/backends/duckdb/datatypes.py
```diff
@@ -139,3 +144,13 @@ def struct():
...
+
+
+@to_sqla_type.register(DuckDBDialect, (dt.MACADDR, dt.INET))
+def sa_duckdb_macaddr(_, itype):
+    return sa.Text
```
cpcloud, line 156:
```suggestion
    return sa.TEXT()
```

ibis/backends/snowflake/registry.py
```diff
@@ -53,6 +53,8 @@ def _literal(t, op):
         return sa.func.object_construct_keep_null(
             *itertools.chain.from_iterable(value.items())
         )
+    elif dtype.is_uuid():
+        return sa.literal(str(value))
```
cpcloud, line 57: Does this work? If so, we should apply it elsewhere in this PR where it makes sense.

```suggestion
        return sa.literal(str(value), type_=sa.UUID())
```

ibis/backends/tests/test_sql.py
```diff
@@ -38,12 +38,12 @@ def test_table(con):
...
-            ["mysql", "sqlite", "mssql"],
+            ["mysql", "mssql"],
             raises=sa.exc.CompileError,
             reason="arrays not supported in the backend",
         ),
         mark.notyet(
-            ["impala"],
+            ["impala", "sqlite"],
             raises=NotImplementedError,
             reason="Impala hasn't implemented array literals",
```
cpcloud, line 48:
```suggestion
            reason="backend hasn't implemented array literals",
```

cpcloud (commented):
LGTM, a few minor comments!

ibis/backends/snowflake/registry.py
krzysztof-kwitt: Here we want to emulate a UUID type that is not supported by the database using the `STRING` type, so here it should be `type_=sa.Text()`, but SQLAlchemy is smart enough to know that.
https://docs.snowflake.com/en/sql-reference/intro-summary-data-types.html

ibis/backends/duckdb/datatypes.py
krzysztof-kwitt: It is not supported in SQLAlchemy < 2.0.
https://github.com/sqlalchemy/sqlalchemy/blob/b3bb55aaea9ee94e50f8db594d65c2bfcdb314ef/lib/sqlalchemy/__init__.py#L89-L131

cpcloud (approved):
LGTM! It looks like there's a merge conflict to resolve.

cpcloud:
@krzysztof-kwitt Can you fix the merge conflict and squash the `fixup!`s? Will merge on green after that!

PR #5358, opened 2023-01-28
feat(sqlalchemy): generalize unnest to work on backends that don't currently support it

cpcloud:
This PR (finally) generalizes the unnest implementation to handle the case where unnest is not supported in a `SELECT` statement, but must come after a cross join in a `FROM` clause.

### DIFF
--- a/ibis/expr/analysis.py
+++ b/ibis/expr/analysis.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import functools
 import operator
 from collections import Counter
-from typing import Iterator, Mapping
+from typing import Iterable, Iterator, Mapping
 
 import toolz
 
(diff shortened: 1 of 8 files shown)


ibis/expr/analysis.py
```diff
@@ -867,3 +868,15 @@ def finder(node):
...
+
+def find_toplevel_unnest_children(nodes: Iterable[ops.Node]) -> Iterator[ops.Table]:
+    def finder(node):
+        return (
+            isinstance(node, ops.Value),
+            find_first_base_table(node) if isinstance(node, ops.Unnest) else None,
+        )
+
+    return itertools.chain.from_iterable(
+        g.traverse(finder, node, filter=ops.Node) for node in nodes
```
kszucs, line 881: Does `g.traverse(finder, nodes, filter=ops.Node)` work?

kszucs (approved):
LGTM on principle.

ibis/expr/analysis.py
cpcloud: Yep!

PR #5359, opened 2023-01-28
feat(duckdb): implement create_view and drop_view method

krzysztof-kwitt:


cpcloud (approved):

krzysztof-kwitt:
I'm wondering if we shouldn't implement these methods in the base class. The same syntax will work with most SQLAlchemy backends.

PR #5364, opened 2023-01-29
ci: skip test that segfaults on Windows

cpcloud:
This PR skips a sporadically segfaulting test on windows. Observed in https://github.com/ibis-project/ibis/actions/runs/4036343394/jobs/6938876498. This segfault does not seem to be reproducible.


cpcloud:
Well, it looks like it's reproducible on Linux too: https://github.com/ibis-project/ibis/actions/runs/4036380326/jobs/6938934300.

cpcloud:
I am going to fix this in a different way

PR #5365, opened 2023-01-29
fix(duckdb): workaround an ownership bug at the interaction of duckdb, pandas and pyarrow

cpcloud:
This PR works around an apparent lifetime issue where reads of pandas.DataFrame with pyarrow-dtype columns through duckdb exit with a segmentation fault.

Here's an example CI run that shows the issue: https://github.com/ibis-project/ibis/actions/runs/4036380326/jobs/6938934300.

I was able to reproduce the issue by repeating the failing test case a large-ish number of times:

```shell
pytest ibis/backends/duckdb/tests/test_register.py::test_memtable_with_nullable_pyarrow_string --count=10000
```

I cannot reproduce it with this fix.

### DIFF
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -248,10 +248,10 @@ class BaseSQLBackend(BaseBackend):
 
         return result
 
-    def _register_in_memory_table(self, table_op):
-        raise NotImplementedError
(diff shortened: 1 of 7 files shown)


cpcloud:
This is causing lots of failures right now, so I'm going to merge this. If anyone objects to anything here later, I am happy to revert and revisit alternative solutions.

PR #5373, opened 2023-01-30
fix(clickhouse): fix decimal literal

krzysztof-kwitt:


ibis/backends/tests/test_generic.py
```diff
@@ -1190,9 +1190,8 @@ def test_exists(batting, awards_players, method_name):
             marks=[
                 pytest.mark.broken(
```
krzysztof-kwitt, line 1191: Probably, we should split this test into two: one for literal and one for typeof.

ibis/backends/clickhouse/compiler/values.py
```diff
@@ -412,6 +412,22 @@ def _literal(op, **kw):
     elif dtype.is_string():
         quoted = value.replace("'", "''").replace("\\", "\\\\")
         return f"'{quoted}'"
+    elif dtype.is_decimal():
+        precision = dtype.precision
+        if precision is None or not 1 <= precision <= 76:
+            raise NotImplementedError(
+                f'Unsupported precision. Supported values: [1 : 76]. Current value: {precision !r}'
```
cpcloud, line 419:
```suggestion
                f'Unsupported precision. Supported values: [1 : 76]. Current value: {precision!r}'
```

cpcloud (approved):

PR #5376, opened 2023-01-30
fix(snowflake): handle the case where pyarrow isn't installed

cpcloud:
This PR fixes an issue where we are trying to use snowflake pyarrow integration without pyarrow installed at all.


cpcloud:
We can really only test this against `master`, and lints passed here so I'm going to merge.

PR #5377, opened 2023-01-30
docs(api): document `_` and its pitfalls

### LINKED ISSUE #4704: ux: problems with ibis' `_` convenience API for deferred attribute resolution

cpcloud:
Closes #4704.


ibis/expr/api.py
```diff
@@ -1031,4 +1032,40 @@ def get_backend(expr: Expr | None = None) -> BaseBackend:
...
+    key: r0.key - 1
+>>> t.mutate(other_value=_.value * 2)
+r0 := UnboundTable: t
+  key   int64
+  value float64
+<BLANKLINE>
+Selection[r0]
+  selections:
+    r0
+    other_value: r0.value * 2
```
jcrist, line 1070: Since our repr sometimes changes (and IMO doesn't provide the user with a ton of actionable info), I've taken to removing the expr reprs from docstrings. You might assign the results of these expressions to a variable instead:

```python
expr2 = t.mutate(other_value=_.value * 2)
```

ibis/expr/api.py
```diff
@@ -1031,4 +1032,40 @@ def get_backend(expr: Expr | None = None) -> BaseBackend:
...
+    See https://github.com/ibis-project/ibis/issues/4704 for details.
+
+    Use `from ibis import deferred as <NAME>` to assign a different name to
+    the deferred object builder.
+
+Examples
+--------
+>>> from ibis import _
+>>> t = ibis.table(dict(key="int", value="float"), name="t")
+>>> t.group_by(key=_.key - 1).agg(total=_.value.sum())
```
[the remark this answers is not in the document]
jcrist: On the other hand, recommending they use `_` everywhere would avoid the "why does `t` work here but not later on" issue. No strong thoughts.

jcrist (approved):
Minor comments, feel free to ignore. Nice to have this documented!

ibis/expr/api.py
cpcloud: Fair, I can remove it. I always find the assignment leaves me wanting for some kind of resolution though.

cpcloud: I'll show the schema repr, that hasn't changed in forever.

cpcloud: Done.

ibis/expr/api.py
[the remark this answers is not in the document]
cpcloud: Well, I just accidentally clobbered @gforsyth's commit. I'm going to leave it as is, because @jcrist makes a good point and I think we should avoid giving the impression that `t` should work everywhere, `deferred`-first IOW.

ogrisel:
Thanks for the fix!

PR #5379, opened 2023-01-30
fix(pandas): don't try to dispatch on arrow dtype if not available

cpcloud:
This PR fixes an issue where a dispatch of `dt.dtype` was attempted on the `ArrowDtype` object, which is not available until pandas 1.5.0.

### DIFF
--- a/ibis/backends/pandas/client.py
+++ b/ibis/backends/pandas/client.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import warnings
 from typing import TYPE_CHECKING
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


ibis/backends/pandas/client.py
```diff
@@ -65,11 +66,20 @@ def from_pandas_extension_dtype(t):
     return getattr(dt, t.__class__.__name__.replace("Dtype", "").lower())
 
 
-@dt.dtype.register(pd.core.arrays.arrow.dtype.ArrowDtype)
-def from_pandas_arrow_extension_dtype(t):
-    import ibis.backends.pyarrow.datatypes as _  # noqa: F401
+try:
+    _arrow_dtype_class = pd.core.arrays.arrow.dtype.ArrowDtype
```
mroeschke, line 70: This is available at the top level as well `pd.ArrowDtype`

cpcloud: Great, thank you!

kszucs (approved):

PR #5380, opened 2023-01-30
fix(duckdb): ensure that `temp_directory` exists

cpcloud:
This PR fixes an issue where a non-existent directory can be used for out of core processing. The solution is to create the directory and its parents.


ibis/backends/duckdb/__init__.py
```diff
@@ -134,6 +134,7 @@ def do_connect(
                 )
 
         if temp_directory is not None:
+            Path(temp_directory).mkdir(parents=True, exist_ok=True)
```
jcrist, line 137: Does duckdb clean up these temporary directories? If not, should we (with `atexit` or something)?

cpcloud: Ah, it looks like duckdb does clean these up, yay!

kszucs (approved):

PR #5389, opened 2023-01-31
feat: change Rich repr of dtypes from blue to dim

NickCrews:
On dark theme screens it was very hard to read the dtype text. The `dim`
style is what is used by the null symbol
"∅", and that looks great in both
white and dark mode
on my machine.


NickCrews:
Before:
image
image
After:
image
image

jcrist (approved):
Thanks @NickCrews! I'm ok with this change. cc @cpcloud for a quick 2nd opinion, since he picked that color originally.

cpcloud:
Ship it!

cpcloud (approved):

PR #5397, opened 2023-02-01
feat(datatype): enable inference of `Decimal` type

deepyaman:


ibis/expr/datatypes/value.py
```diff
@@ -150,6 +150,12 @@ def infer_enum(_: enum.Enum) -> dt.String:
     return dt.string
 
 
+@infer.register(decimal.Decimal)
+def infer_decimal(value: decimal.Decimal) -> dt.Decimal:
+    """Infer the [`Decimal`][ibis.expr.datatypes.Decimal] type of `value`."""
```
jcrist, line 155: No need for a docstring here, lets remove it to be consistent with the other dispatches.

```suggestion
```

jcrist (commented):
Implementation LGTM! Can you add a test case for this? It should go right after this line [here](https://github.com/ibis-project/ibis/blob/cefc20921e95956a19a43fb6050ce8d3804bf28f/ibis/tests/expr/test_datatypes.py#L771).

ibis/expr/datatypes/value.py
deepyaman: @jcrist Thanks for the speedy review! I've added the test. I'm also happy to do this, but wanted to point out that a lot of the dispatches have this docstring (from where I copied it), while others don't. I can standardize it across the dispatches, if you feel strongly one way or another, or I can just delete the docstring from this one.

There's also other minor inconsistencies, like some of the docstrings saying `value` even though the parameter name for them is `values`.

jcrist: Ah, you're right. No need to drop the docstring here, and don't worry about standardizing things. Lets leave it as is.

deepyaman:
I haven't spent much time looking at it, but I'm not sure why I've broken type coercion such that it's not able to collect the tests where this is happening in the parametrization--could it be there was a reason `decimal.Decimal` didn't have a dispatch?

FWIW the reason I'd initially added this was to address the fact that `dt.infer(decimal.Decimal())` didn't work for me.

cpcloud:
Hi @deepyaman, thanks for the PR!

This is currently broken here because:

1. We construct ibis literals in fixtures to construct decimal test cases
2. `ibis.literal` will hit the `castable` function to ensure that its input is castable to the specified or inferred type
3. The `castabl` implementation for `decimal.Decimal` doesn't consider a source type with unspecified precision and scale as castable to target type with a specified precision and scale.

The main thing to resolve here is the last point: is unspecified precision and scale castable to a specified precision and scale?

Currently the meaning of precision and scale both `None` is "backend's default".

I am inclined to loosen the current restriction and let all `decimal -> decimal` casts through and let the backend complain if it doesn't like what you're asking it to do 😄.

Thoughts?

jcrist:
> I am inclined to loosen the current restriction and let all decimal -> decimal casts through and let the backend complain if it doesn't like what you're asking it to do

That makes sense to me. The other option is to have `infer` return a decimal type with `precision` and `scale` inferred from the literal value.

deepyaman:
> > I am inclined to loosen the current restriction and let all decimal -> decimal casts through and let the backend complain if it doesn't like what you're asking it to do
> 
> That makes sense to me. The other option is to have `infer` return a decimal type with `precision` and `scale` inferred from the literal value.

Upon looking into it a bit, it doesn't seem that Python decimals have a concept of scale, and even precision is only used during arithmetic operations and doesn't matter otherwise.

> I am inclined to loosen the current restriction and let all `decimal -> decimal` casts through and let the backend complain if it doesn't like what you're asking it to do 😄.

Happy to do this, if want to go this route. I suppose it's possible to preserve the current behavior as an option--maybe `upcast` defaulting to `False`, as it is for float casting?--but I don't have the context on when this would be used.

```python
@castable.register(dt.Decimal, dt.Decimal)
def can_cast_decimals(source: dt.Decimal, target: dt.Decimal, upcast: bool = False, **kwargs) -> bool:
    if upcast:
        target_prec = target.precision
        source_prec = source.precision
        target_sc = target.scale
        source_sc = source.scale
        return (
            target_prec is None or (source_prec is not None and target_prec >= source_prec)
        ) and (target_sc is None or (source_sc is not None and target_sc >= source_sc))

    return True
```

jcrist:
>  it doesn't seem that Python decimals have a concept of scale, and even precision is only used during arithmetic operations and doesn't matter otherwise.

Yeah. I was saying more we'd infer the scale/precision from the current decimal value. So `Decimal("1.2345")` would have a precision of 5 and a scale of 2.

cpcloud:
Let's relax the casting rules. 

The new casting rules would be:

1. If either side's precision and scale **both** None
2. Fallback to the current logic (sans overlap with point 1.)

deepyaman:
> Let's relax the casting rules.
> 
> The new casting rules would be:
> 
> 1. If either side's precision and scale are **both** None
> 2. Fallback to the current logic (sans overlap with point 1.)

@cpcloud Just to clarify, for rule 1, do you mean, "If either side's precision or scale are `None`, return `True`"?

cpcloud:
> > Let's relax the casting rules.
> > The new casting rules would be:
> > 
> > 1. If either side's precision and scale are **both** None
> > 2. Fallback to the current logic (sans overlap with point 1.)
> 
> @cpcloud Just to clarify, for rule 1, do you mean, "If either side's precision or scale are `None`, return `True`"?

Yep, fixing!

cpcloud (approved):
Sweet, thanks!

PR #5398, opened 2023-02-01
docs: improve/update pandas walkthrough

NickCrews:
- Remove deprecated usage of .drop()
- Add ifelse() example
- Add example on how column expressions get default names unless
  you .name() them
- add note on lazy/eagerness
- add note on lack of indexes


cpcloud (approved):
Awesome, thank you!

PR #5401, opened 2023-02-01
feat(api): implement `to_pandas()` API for ecosystem compatibility

cpcloud:
This PR adds a `to_pandas()`, which the `plotnine` library uses to turn non-DataFrame objects into pandas objects, providing lightweight compatibility with an ecosystem project.

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -8,11 +8,12 @@ import ibis
 import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
-from ibis import util
 from ibis.common.grounds import Singleton
 from ibis.expr.types.core import Expr, _binop, _FixedTextJupyterMixin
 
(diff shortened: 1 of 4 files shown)


ibis/expr/types/generic.py
```diff
@@ -509,6 +510,18 @@ def as_table(self) -> ir.Table:
         table = roots[0].to_expr()
         return table.select(self)
 
+    def to_pandas(self, *args, **kwargs) -> pd.Series:
```
jcrist, line 513: For `to_pyarrow`/`to_pyarrow_batches` we moved to keyword-only arguments. Since this is a new API, I think that makes sense to do here too. Only change would be to drop the `*args` from here and the docstring below.

jcrist (commented):
Just a small nit, otherwise this LGTM

ibis/expr/types/generic.py
cpcloud: Fixed.

PR #5405, opened 2023-02-01
docs(tutorial): remove impala from sql tutorial

gforsyth:
Instead of mentioning the Impala SQL compiler, use `ibis.show_sql` to show sql examples.


jcrist (approved):
Nice find!

PR #5407, opened 2023-02-01
fix(imports): remove top-level import of sqlalchemy from base backend

gforsyth:
I noticed that the `bigquery` extra didn't require `sqlalchemy` but I
hit an import error for `sqlalchemy` via `base/sql/__init__.py`.

Most of our sql backends make use of sqlalchemy, but not all, and so it
seems odd to require it.  Moving it inside the single function call that
makes use of it.


cpcloud (approved):
Good catch. We should probably move the backends that don't depend on sqlalchemy, but support `ibis.connect` to use `urllib`.

PR #5412, opened 2023-02-02
feat(dask/pandas): raise OperationNotDefinedError exc for not defined operations

krzysztof-kwitt:


krzysztof-kwitt:
@cpcloud it is ready for review.

ibis/backends/pandas/dispatch.py
```diff
@@ -21,8 +21,12 @@
 
 
 @execute_node.register(ops.Node)
-def raise_unknown_op(node, **kwargs):
-    raise NotImplementedError(
+@execute_node.register(ops.Node, object)
+@execute_node.register(ops.Node, object, object)
+@execute_node.register(ops.Node, object, object, object)
+@execute_node.register(ops.Node, object, object, object, object)
```
cpcloud, line 27: You should be able to write all 5 of these dispatches as `@execute_node.register(ops.Node, [object])`. This will use `multipledispatch`'s variadic dispatch support.

cpcloud (approved):
One small nit, otherwise LGTM!

ibis/backends/pandas/dispatch.py
krzysztof-kwitt: Fixed.

PR #5415, opened 2023-02-02
docs: remove duplicate "or"

cpcloud:
Follow up from #5411


cpcloud:
The tpch skip helper is added in #5416.

Issue #5420, opened 2023-02-02
bug: ibis.read_parquet fails AttributeError: 'NoneType' object has no attribute 'read_parquet' if the DuckDB backend is not installed


ogrisel:
### What happened?

This is mostly a UX problem, I am not sure what fixes are best. Let's start with the reproducer to highlight how a naive user (e.g. myself :) could get confused by the current state of the online doc and public API:

- Install `ibis-framework` in a fresh Python environment with the default instructions at the top of https://ibis-project.org/docs/4.1.0/install/

```
pip install ibis-framework
```

or alternatively, install ibis with a backend that would be expected to naturally support parquet file reading (but not duckdb):

```
pip install ibis-framework[dask]
```

- Then naively call `ibis.read_parquet`:

```python
>>> import ibis
>>> data_url = "https://storage.googleapis.com/ibis-tutorial-data/wowah_data/wowah_data_raw.parquet"
>>> ibis.read_parquet(data_url)
Traceback (most recent call last):
  Cell In[7], line 2
    ibis.read_parquet(data_url)
  File ~/mambaforge/envs/tmp/lib/python3.11/site-packages/ibis/expr/api.py:902 in read_parquet
    return con.read_parquet(sources, **kwargs)
AttributeError: 'NoneType' object has no attribute 'read_parquet'
```

At least the error message should be improved to state that no default backend has been configured and point to the documentation on how to do so.

I googled and found:

- https://ibis-project.org/docs/4.1.0/user_guide/configuration/ but that does not mention the backend config
- https://ibis-project.org/docs/4.1.0/api/config/?h=default_backend that suggests that using a `str` (likely the backend name) should be a valid type for that config option:

But then:

```python
>>> ibis.options.default_backend = 'dask'
>>> ibis.read_parquet(data_url)
Traceback (most recent call last):
  Cell In[10], line 1
    ibis.read_parquet(data_url)
  File ~/mambaforge/envs/tmp/lib/python3.11/site-packages/ibis/expr/api.py:902 in read_parquet
    return con.read_parquet(sources, **kwargs)
AttributeError: 'str' object has no attribute 'read_parquet'
```

But then even if configuring as follows does not work:

```python
>>> ibis.options.default_backend = ibis.dask
>>> ibis.read_parquet(data_url)
Traceback (most recent call last):
  Cell In[16], line 1
    ibis.read_parquet(data_url)
[...]

PR #5423, opened 2023-02-02
docs(io): doc/ux improvements for read_parquet and friends

gforsyth:
xref #5420

Default to `duckdb` in the install doc, fix the API doc of
`default_backend`, and note that the `read_*` functions passthrough to
the default backend.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -10,10 +10,12 @@ hide:
 === "pip"
 
     ```sh
-    pip install ibis-framework # (1)
+    pip install 'ibis-framework[duckdb]' # (1) (2)
     ```
 
-    1. Note that the `ibis-framework` package is *not* the same as the `ibis` package in PyPI.  These two libraries cannot coexist in the same Python environment, as they are both imported with the `ibis` module name.
+    1. We suggest starting with the DuckDB backend.  It's performant and fully featured.  If you would like to use a different backend, all of the available options are listed below.
+
+    2. Note that the `ibis-framework` package is *not* the same as the `ibis` package in PyPI.  These two libraries cannot coexist in the same Python environment, as they are both imported with the `ibis` module name.
 
 {% for mgr in ["conda", "mamba"] %}
 === "{{ mgr }}"
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -826,12 +826,16 @@ def row_number() -> ir.IntegerColumn:
 def read_csv(sources: str | Path | Sequence[str | Path], **kwargs: Any) -> ir.Table:
     """Lazily load a CSV or set of CSVs.
 
+    This function delegates to the `read_csv` method on the current default
+    backend (DuckDB or `ibis.config.default_backend`).
+
     Parameters
     ----------
     sources
         A filesystem path or URL or list of same.  Supports CSV and TSV files.
     kwargs
-        DuckDB-specific keyword arguments for the file type.
+        Backend-specific keyword arguments for the file type. For the DuckDB
+        backend used by default, please refer to:
 
         * CSV/TSV: https://duckdb.org/docs/data/csv#parameters.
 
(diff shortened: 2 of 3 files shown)


docs/install.md
```diff
@@ -10,10 +10,12 @@ hide:
...
+    pip install 'ibis-framework[duckdb]' # (1) (2)
     ```
 
-    1. Note that the `ibis-framework` package is *not* the same as the `ibis` package in PyPI.  These two libraries cannot coexist in the same Python environment, as they are both imported with the `ibis` module name.
+    1. We suggest starting with the DuckDB backend.  It's performant and fully featured.  If you would like to use a different backend, all of the available options are listed below.
+
+    2. Note that the `ibis-framework` package is *not* the same as the `ibis` package in PyPI.  These two libraries cannot coexist in the same Python environment, as they are both imported with the `ibis` module name.
 
 {% for mgr in ["conda", "mamba"] %}
 === "{{ mgr }}"
```
ogrisel, line 21: Thanks for the quick PR!

Is it intentional that the conda-forge `ibis-framework` comes by default with the duckdb engine while the pypi does not? I had not realized this discrepancy existed when opening the issue.

ibis/expr/api.py
```diff
@@ -826,6 +826,9 @@ def row_number() -> ir.IntegerColumn:
 def read_csv(sources: str | Path | Sequence[str | Path], **kwargs: Any) -> ir.Table:
     """Lazily load a CSV or set of CSVs.
 
+    This function delegates to the `read_csv` method on the current default
+    backend (DuckDB or `ibis.config.default_backend`)
```
ogrisel, line 830:
```suggestion
    backend (DuckDB or `ibis.config.default_backend`).
```

ibis/expr/api.py
```diff
@@ -854,6 +857,9 @@ def read_csv(sources: str | Path | Sequence[str | Path], **kwargs: Any) -> ir.Ta
 def read_json(sources: str | Path | Sequence[str | Path], **kwargs: Any) -> ir.Table:
     """Lazily load newline-delimited JSON data.
 
+    This function delegates to the `read_json` method on the current default
+    backend (DuckDB or `ibis.config.default_backend`)
```
ogrisel, line 861:
```suggestion
    backend (DuckDB or `ibis.config.default_backend`).
```

ibis/expr/api.py
```diff
@@ -879,6 +885,9 @@ def read_json(sources: str | Path | Sequence[str | Path], **kwargs: Any) -> ir.T
 def read_parquet(sources: str | Path | Sequence[str | Path], **kwargs: Any) -> ir.Table:
     """Lazily load a parquet file or set of parquet files.
 
+    This function delegates to the `read_parquet` method on the current default
+    backend (DuckDB or `ibis.config.default_backend`)
```
ogrisel, line 889:
```suggestion
    backend (DuckDB or `ibis.config.default_backend`).
```

ogrisel: The `kwargs` parameter below (and in the other functions) should accordingly be updated to replace:

```
        DuckDB-specific keyword arguments for the file type.
```

by:

```
        Backend-specific keyword arguments for the file type. For the DuckDB
        backend used by default, please refer to:
```

docs/install.md
cpcloud: Not really, except that conda-forge doesn't have a notion of optional dependencies so we decided to make the default install a bit more convenient.

If we did this for pypi installs then duckdb (and a few other things, like sqlalchemy) would become required dependencies.

ibis/expr/api.py
gforsyth: done!  and thanks for the review!

ibis/expr/api.py
```diff
@@ -854,12 +858,16 @@ def read_csv(sources: str | Path | Sequence[str | Path], **kwargs: Any) -> ir.Ta
...
+    backend (DuckDB or `ibis.config.default_backend`).
+
     Parameters
     ----------
     sources
         A filesystem path or URL or list of same.
     kwargs
-        DuckDB-specific keyword arguments for the file type.
+        Backend-specific keyword arguments for the file type. For the DuckDB
+        backend used by default, there are no valid keywork arguments
```
ogrisel, line 870:
```suggestion
        backend used by default, there are no valid keywork arguments.
```

ogrisel (approved):
LGTM, thanks for the great improvements!

PR #5425, opened 2023-02-02
fix(io): add `read_parquet` and `read_csv` to base backend mixin

gforsyth:
xref #5420

This adds `read_parquet` and `read_csv` as methods to the `FileIOHandler` mixin (renamed from `ResultHandler`).

Now every backend will have these two `read_` methods and will raise a `NotImplementedError`.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -214,8 +214,7 @@ class TablesAccessor(collections.abc.Mapping):
         return self._backend.list_tables()
 
 
-# should have a better name
-class ResultHandler:
+class _FileIOHandler:
     @staticmethod
     def _import_pyarrow():
         try:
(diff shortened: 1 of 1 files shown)


ibis/backends/base/__init__.py
```diff
@@ -215,7 +215,7 @@ def _ipython_key_completions_(self) -> list[str]:
 
 
 # should have a better name
-class ResultHandler:
+class FileIOHandler:
```
ogrisel, line 218: Isn't this class name part of the public API (for backend developers)? Doesn't this renaming a deprecation cycle with a temporary backward compat alias?

ibis/backends/base/__init__.py
```diff
@@ -322,8 +322,52 @@ def to_pyarrow_batches(
...
+            a sequentially generated name.
+        **kwargs
+            Additional keyword arguments passed to the backend loading function.
+
+        Returns
+        -------
+        ir.Table
+            The just-registered table
+        """
+        raise NotImplementedError
```
ogrisel, line 345: It can be informative to report the name of the specific backend that raised this expression in case this is not the one the user would have expected to make it easier for them to updated their install / runtime config / code accordingly.


```suggestion
        raise NotImplementedError(
            f"{self.__class__} does not support direct registration of parquet data."
        )
```

ibis/backends/base/__init__.py
```diff
@@ -322,8 +322,52 @@ def to_pyarrow_batches(
...
```
ogrisel, line 367:
```suggestion
        raise NotImplementedError(
            f"{self.__class__} does not support direct registration of CSV data."
        )
```

ogrisel (commented):
Thanks for the fix. Here is some feedback.

ogrisel:
Also, what about `read_json`? Is it too experimental to make it part of the default backend API?

ibis/backends/base/__init__.py
```diff
@@ -215,7 +215,7 @@ def _ipython_key_completions_(self) -> list[str]:
 
 
 # should have a better name
```
cpcloud, line 217: The name seems reasonable now.

cpcloud:
> Also, what about `read_json`? Is it too experimental to make it part of the default backend API?

Yeah, it's pretty rough right now and only works if your data are well-structured.

ibis/backends/base/__init__.py
cpcloud: I don't think so, this was introduced in `4.0` and it is extremely unlikely that someone has built a backend in the last month that is using this class.

@gforsyth Can you make this private with a leading underscore?

gforsyth: yeah, will do

gforsyth:
Fixed up.
My thinking was that experimental `ibis.read_` methods shouldn't be included yet, but if that inconsistency introduces extra confusion I can change course.

cpcloud (approved):
Ship it!

PR #5426, opened 2023-02-02
fix(backend): provide useful error if default backend is unavailable

gforsyth:
If a user installs only `ibis-framework` and then tries to run `ibis.read_parquet` or similar, they get an unhelpful error message. This tries to guide them in a helpful direction.

### DIFF
--- a/ibis/backends/conftest.py
+++ b/ibis/backends/conftest.py
@@ -4,6 +4,7 @@ import contextlib
 import importlib
 import importlib.metadata
 import platform
+import sys
 from functools import lru_cache
 from pathlib import Path
 from typing import Any, TextIO
@@ -730,3 +731,13 @@ def test_employee_data_2():
     )
 
     return df2
+
+
+@pytest.fixture
+def no_duckdb(backend, monkeypatch):
+    monkeypatch.setitem(sys.modules, "duckdb", None)
+
+
+@pytest.fixture
+def no_pyarrow(backend, monkeypatch):
+    monkeypatch.setitem(sys.modules, "pyarrow", None)
--- a/ibis/config.py
+++ b/ibis/config.py
@@ -4,6 +4,7 @@ from typing import Any, Callable, Optional
 from public import public
 from typing_extensions import Annotated
 
+import ibis.common.exceptions as com
 from ibis.common.grounds import Annotable
 from ibis.common.validators import min_
 
@@ -174,7 +175,26 @@ def _default_backend() -> Any:
     try:
         import duckdb as _  # noqa: F401
     except ImportError:
-        return None
+        raise com.IbisError(
+            """\
+You have used a function that relies on the default backend, but the default
+backend (DuckDB) is not installed.
+
+You may specify an alternate backend to use, e.g.
+
+ibis.set_backend("polars")
+
+or to install the DuckDB backend, run:
+
+    pip install 'ibis-framework[duckdb]'
+
+or
+
+    conda install -c conda-forge ibis-framework
+
+For more information on available backends, visit https://ibis-project.org/install
+"""
+        )
 
     import ibis
 
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -770,8 +770,19 @@ def test_default_backend_no_duckdb(backend):
     # run this twice to ensure that we hit the optimizations in
     # `_default_backend`
     for _ in range(2):
-        with pytest.raises(com.IbisError, match="Expression depends on no backends"):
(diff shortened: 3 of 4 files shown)


ibis/backends/conftest.py
```diff
@@ -730,3 +731,38 @@ def test_employee_data_2():
...
+        self.pkgnames = []
+
+    def find_spec(self, fullname, path, target=None):
+        if fullname in self.pkgnames:
+            raise ImportError(fullname)
+
+
+@pytest.fixture
+@pytest.mark.usefixtures("backend")
+def no_duckdb():
```
jcrist, line 747: This can be done simpler using `monkeypatch`, and the fact that if `sys.modules["foo"]` returns `None` then an `ImportError` is raised. Something like:

```python
@pytest.fixture
def no_duckdb(monkeypatch):
    monkeypatch.setitem(sys.modules, "duckdb", None)
```

ibis/config.py
```diff
@@ -174,7 +174,26 @@ def _default_backend() -> Any:
     try:
         import duckdb as _  # noqa: F401
     except ImportError:
-        return None
+        raise ImportError(
+            """
```
jcrist, line 178: So the error message doesn't start with a newline

```suggestion
            """\
```

jcrist (commented):
A minor nit and a suggestion, otherwise this LGTM

ibis/backends/conftest.py
gforsyth: that's much nicer, thanks!

ogrisel: Very nice indeed, I did no know about that either.

ibis/backends/conftest.py
```diff
@@ -730,3 +731,21 @@ def test_employee_data_2():
...
     return df2
+
+
+@pytest.fixture
+@pytest.mark.usefixtures("backend")
+def no_duckdb(monkeypatch):
+    _duckdb = sys.modules.get("duckdb", None)
+    monkeypatch.setitem(sys.modules, "duckdb", None)
+    yield
+    monkeypatch.setitem(sys.modules, "duckdb", _duckdb)
```
ogrisel, line 742: The fact that codecov complains that this line is not covered is fishy. Is a try / finally protection required here to avoid introducing tests with a side effect? OR does the `monkeypatch` fixture offers this protection by default?

I seems that since not other duckdb-dependant tests fail, the latter might hold, but still it's confusing.

ogrisel (approved):
Thanks for the PR. Indeed raising in `_default_backend` is much better than warning or raising in `get_backend` as originally suggested in #5420.

ibis/backends/conftest.py
cpcloud: The automatic `finally` behavior is part of `pytest` fixtures.

ogrisel: But they why would codecov complain?

cpcloud: Unclear, maybe codecov is broken 🤷🏻

ogrisel: Honestly I doubt it but it's probably not that important as long as it does not have a side effect break other unrelated tests.

cpcloud: Actually, this test is never run when passing any pytest markers:

```
```


I think what's happening is that `usefixtures` doesn't actually pass `backend` as an argument to the fixture (not sure why, we probably are misunderstanding what `usefixtures` does) so this test never gets picked up by our backend test collection code.

Passing `backend` explicitly to `no_duckdb` fixes the issue.

Thanks for keeping us honest @ogrisel!

cpcloud: Ah, yeah we're just using it wrong here, it has no effect when applied to a fixture: https://docs.pytest.org/en/latest/reference/reference.html#pytest-mark-usefixtures

cpcloud: Pushed up a commit to fix.

cpcloud (approved):

ibis/backends/conftest.py
jcrist: You also don't need to yield and cleanup after, monkeypatch does all that for you. (Apologies for the brief comment, I'm on my phone).

gforsyth: I added the `yield` initially because every backend test suite was failing, glad we sorted out the actual issue.  Time to read more docs.

gforsyth:
I'm going to squash and remove the extraneous `yield` and `setitem`

cpcloud:
Merge as you like!

PR #5431, opened 2023-02-03
ci: remove autorebase job

cpcloud:
This is doing nothing except eating up API calls towards the rate limit quota.


cpcloud:
None of the CI jobs depend on this in any way, so it's pointless to run them.

PR #5434, opened 2023-02-03
ci: renable nix builds

cpcloud:
Enable nix builds in CI again. Previously we disabled them because they built things from source, but now that they are using wheels they are well within reasonable CI times.


cpcloud:
Merging on green to get upstream passing again.

PR #5014, opened 2022-12-15
refactor(ir/api): introduce window frame operation and revamp the window API

kszucs:
resolves #3305
mainly resolves #1834 we just need to expose the same arguments to `.over()`

possible follow-ups: 
 - add support for `GROUPS` frame
 - remove max lookback support
 - deprecate legacy API
 - support for deferred objects (a more generic approach would be nice)
 - update the documentation

PR #5437, opened 2023-02-03
feat(ux): expose `rows` and `range` keyword arguments in `value.over()`

kszucs:
depends on #5014

### DIFF
--- a/ibis/expr/types/groupby.py
+++ b/ibis/expr/types/groupby.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import types
 from typing import Iterable, Sequence
 
+import ibis
 import ibis.expr.analysis as an
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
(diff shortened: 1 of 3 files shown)


cpcloud:
@kszucs Can you rebase?

ibis/tests/expr/test_window_functions.py
```diff
@@ -38,3 +38,27 @@ def test_mutate_with_analytic_functions(alltypes):
...
+    expr1 = t.f.cumsum().over(rows=(0, 1), group_by=t.g, order_by=[t.f, t.h])
+    expected1 = t.f.cumsum().over(w1)
+    assert expr1.equals(expected1)
+
+    expr2 = t.f.cumsum().over(range=(-1, 1), group_by=[t.g, t.a], order_by=[t.f])
+    expected2 = t.f.cumsum().over(w2)
+    assert expr2.equals(expected2)
+
+    expr3 = t.f.cumsum().over(
+        between=(ibis.interval(days=1), ibis.interval(days=5)), order_by=[t.f]
```
kszucs, line 61: I'm not sure that we should expose the between keyword argument, it enables inferring the window frame kind if it isn't passed explicitly.

I tend to think that we should enforce either `over(rows=` or `over(range=)` but not `over(between=)`.

ibis/expr/types/groupby.py
```diff
@@ -222,19 +223,43 @@ def _get_window(self):
                 order_by=self._window.order_by + self._order_by,
             )
 
-    def over(self, window) -> GroupedTable:
+    def over(
```
kszucs, line 226: Actually this method is not used anywhere, at least removing it doesn't break a single test.

cpcloud: Yeah. I didn't realize we had support for applying a window over a group by. This makes sense to me though. It avoids having to duplicate a window if the rows/range specification are more complex than can be done with `.group_by(...).mutate(...)`

cpcloud (approved):

PR #5446, opened 2023-02-05
feat(clickhouse): clickhouse more ops

krzysztof-kwitt:


ibis/backends/clickhouse/compiler/values.py
```diff
@@ -653,6 +653,11 @@ def _string_ilike(op, **kw):
     return f"lower({arg}) LIKE lower({pattern})"
 
 
+@translate_val.register(ops.Capitalize)
+def _string_capitalize(op, **kw):
+    return f"CONCAT(UPPER(SUBSTR({translate_val(op.arg, **kw)}, 1, 1)), SUBSTR({translate_val(op.arg, **kw)}, 2))"
```
cpcloud, line 658: Can you pull out the `op.arg` translate call into a variable so we don't translate it twice?

cpcloud: Actually, I can do this so we can just merge it!

cpcloud (approved):

cpcloud:
Thank you!

PR #5449, opened 2023-02-05
Normalize exceptions in backends

krzysztof-kwitt:
To detect some hidden issues, I added stricter exceptions validation.


cpcloud:
@krzysztof-kwitt What is the status of this PR? It looks like there are a lot of conflicts. Happy to help however!

krzysztof-kwitt:
I changed my approach and now I am making changes file by file, not all at once. Potentially, we can extract changes  that are not in the tests because they look sensible.

cpcloud (approved):
Changes look good to me!

I'll add a breaking change note to the commit since changing the exception types potentially breaks downstream users and then merge!

Thanks for this!

PR #5451, opened 2023-02-06
refactor(datatype): clean up parsing rules

cpcloud:
This PR cleans up the parsing rules to remove `parsy.generate` usage.


jcrist (approved):
I confess I only gave this PR a cursory review, but all tests are passing so :shipit:.

PR #5455, opened 2023-02-06
refactor(datatype): use lazy dispatch when inferring pandas Timedelta objects

kszucs:


jcrist (approved):
LGTM provided tests pass

PR #5457, opened 2023-02-06
feat(io): add to_parquet, to_csv to backends

gforsyth:
`to_csv` and `to_parquet` seem self-evidently necessary.  ~PyArrow makes adding `to_orc` very easy, so I thought why not?~  because of windows, that's why not.

Also adding in a slightly silly `to_pyarrow_batches` method for the Pandas backend -- for CSV and parquet, I'm writing out the results of an expression as batches so we don't blow up memory.  I can either override `to_csv` and `to_parquet` in the pandas backend, or add in the slightly roundabout recordbatch producer.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -369,6 +369,86 @@ class _FileIOHandler:
             f"{self.name} does not support direct registration of CSV data."
         )
 
+    @util.experimental
+    def to_parquet(
+        self,
+        expr: ir.Table,
+        path: str | Path,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        **kwargs: Any,
+    ) -> None:
+        """Write the results of executing the given expression to a parquet file.
+
+        This method is eager and will execute the associated expression
+        immediately.
+
+        Parameters
+        ----------
+        expr
+            The ibis expression to execute and persist to parquet.
+        path
+            The data source. A string or Path to the parquet file.
+        params
+            Mapping of scalar parameter expressions to value.
+        limit
+            An integer to effect a specific row limit. A value of `None` means
+            "no limit". The default is in `ibis/config.py`.
+        **kwargs
+            Additional keyword arguments passed to pyarrow.parquet.ParquetWriter
+
+        https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html
+        """
+        self._import_pyarrow()
+        import pyarrow.parquet as pq
+
+        batch_reader = expr.to_pyarrow_batches(params=params, limit=limit)
+
+        with pq.ParquetWriter(path, batch_reader.schema) as writer:
+            for batch in batch_reader:
+                writer.write_batch(batch)
+
+    @util.experimental
+    def to_csv(
+        self,
+        expr: ir.Table,
+        path: str | Path,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        **kwargs: Any,
+    ) -> None:
+        """Write the results of executing the given expression to a CSV file.
+
+        This method is eager and will execute the associated expression
+        immediately.
+
+        Parameters
+        ----------
+        expr
+            The ibis expression to execute and persist to CSV.
+        path
+            The data source. A string or Path to the CSV file.
+        params
+            Mapping of scalar parameter expressions to value.
+        limit
+            An integer to effect a specific row limit. A value of `None` means
+            "no limit". The default is in `ibis/config.py`.
+        **kwargs
+            Additional keyword arguments passed to pyarrow.csv.CSVWriter
+
+        https://arrow.apache.org/docs/python/generated/pyarrow.csv.CSVWriter.html
+        """
+        self._import_pyarrow()
+        import pyarrow.csv as pcsv
+
+        batch_reader = expr.to_pyarrow_batches(params=params, limit=limit)
+
+        with pcsv.CSVWriter(path, batch_reader.schema) as writer:
+            for batch in batch_reader:
+                writer.write_batch(batch)
+
 
 class BaseBackend(abc.ABC, _FileIOHandler):
     """Base backend class.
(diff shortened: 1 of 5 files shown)


ibis/backends/base/__init__.py
```diff
@@ -369,6 +369,100 @@ def read_csv(
...
+            Mapping of scalar parameter expressions to value.
+        limit
+            An integer to effect a specific row limit. A value of `None` means
+            "no limit". The default is in `ibis/config.py`.
+        **kwargs
+            Additional keyword arguments passed to pyarrow.parquet.ParquetWriter
+
+        https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html
+        """
+        _ = self._import_pyarrow()
```
kszucs, line 403: Is it for better error reporting? 
Is the underscore required?

ibis/backends/base/__init__.py
```diff
@@ -369,6 +369,100 @@ def read_csv(
...
+        import pyarrow.parquet as pq
```
kszucs, line 404: Just a nit, but we could have a `pq = self._import_parquet()` similarly to `_import_pyarrow()` if we want even better error reporting.

ibis/backends/base/__init__.py
```diff
@@ -369,6 +369,100 @@ def read_csv(
...
+
+        batch_reader = expr.to_pyarrow_batches(params=params, limit=limit)
+
+        writer = pq.ParquetWriter(str(path), batch_reader.schema)
```
kszucs, line 408: IIRC parquet writer should already support path like objects.

ibis/backends/pandas/__init__.py
```diff
@@ -212,6 +212,21 @@ def to_pyarrow(
...
+        expr: ir.Expr,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        chunk_size: int = 1000000,
+        **kwargs: Any,
+    ) -> pa.ipc.RecordBatchReader:
+        pa = self._import_pyarrow()
+        pa_table = self.to_pyarrow(expr, params=params, limit=limit)
+        return pa.RecordBatchReader.from_batches(
```
kszucs, line 226: How about creating record batches directly from the dataframe using `pa.RecordBatch.from_pandas` and explicitly raising here if the result is not table-like? Do we support chunked arrays here as well?

ibis/backends/tests/test_export.py
```diff
@@ -156,3 +150,37 @@ def test_no_pyarrow_message(awards_players, monkeypatch):
...
     with pytest.raises(ModuleNotFoundError, match="requires `pyarrow` but"):
         awards_players.to_pyarrow()
+
+
+@pytest.mark.parametrize("limit", limit_no_limit)
+def test_table_to_parquet(tmp_path, backend, limit, awards_players):
+    outparquet = tmp_path / "out.parquet"
+    awards_players.to_parquet(outparquet, limit=limit)
+
+    import pandas as pd
```
kszucs, line 160: I think you can safely import pandas at the top, since we use it across the whole test suite.

ibis/backends/pandas/__init__.py
gforsyth: I think this requires all the same acrobatics as the `to_pyarrow` handler, since you can't create a `RecordBatch` from a series, so we'd need to promote single-column result sets to single-column tables (at which point we might as well just use the `to_pyarrow` to feed the batches)

ibis/backends/base/__init__.py
gforsyth: What's the likelihood of a user having `pyarrow` installed without `parquet` / `csv` support built?  I know it's possible to build `pyarrow` without that, but are there any released binaries / wheels / whatever that don't have it built in?

ibis/backends/base/__init__.py
gforsyth: Yes and No.  Removed the underscore.

ibis/backends/base/__init__.py
```diff
@@ -369,6 +369,100 @@ def read_csv(
...
+            "no limit". The default is in `ibis/config.py`.
+        **kwargs
+            Additional keyword arguments passed to pyarrow.csv.CSVWriter
+
+        https://arrow.apache.org/docs/python/generated/pyarrow.csv.CSVWriter.html
+        """
+        self._import_pyarrow()
+        import pyarrow.csv as pcsv
+
+        pcsv.write_csv(expr.to_pyarrow(params=params, limit=limit), str(path), **kwargs)
```
cpcloud, line 452: It looks like we're writing the CSV twice, once using `to_pyarrow` and the other using `to_pyarrow_batches`. Am I reading this code correctly?

ibis/backends/base/__init__.py
```diff
@@ -369,6 +369,100 @@ def read_csv(
...
+            Additional keyword arguments passed to pyarrow.parquet.ParquetWriter
+
+        https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html
+        """
+        self._import_pyarrow()
+        import pyarrow.parquet as pq
+
+        batch_reader = expr.to_pyarrow_batches(params=params, limit=limit)
+
+        writer = pq.ParquetWriter(path, batch_reader.schema)
```
cpcloud, line 408: Can we use the `*Writer`s as context managers to avoid the `try:`/`finally:`?

ibis/backends/base/__init__.py
gforsyth: lol, whoops -- don't know how that reappeared

ibis/backends/base/__init__.py
gforsyth: We can use them as context managers, but the record batch reader will still raise a `StopIteration` when it's exhausted.

ibis/backends/base/__init__.py
```diff
@@ -369,6 +369,88 @@ def read_csv(
...
+
+        with pq.ParquetWriter(path, batch_reader.schema) as writer:
```
cpcloud, line 410: It looks like there are two writers now, I think we only need one.

gforsyth: ugh, fixed

cpcloud (approved):
LGTM, great work!

gforsyth:
Yes, xref https://github.com/ibis-project/ibis/issues/4444

PR #5462, opened 2023-02-07
fix(backends): fix capitalize implementations across all backends

cpcloud:
This PR fixes an issue where an extra character was sneaking in to some capitalize implementations.


cpcloud:
Ok, even though the postgres implementation is inconsistent with Python's definition of `str.capitalize()` I'll leave it for now and put up another (breaking change) PR to make the behavior consistent.

gforsyth (approved):
:shipit:

Issue #3739, opened 2022-04-08
docs(perf): figure out how to stop re-rendering all jupyter notebooks during mkdocs serve


p-a-a-a-trick:
mkdocs renders all jupyter notebooks during `mkdocs serve`, slowing documentation update/checking to a snail's pace.  It's tedious.  Is there some way to speed it up, have them render concurrently, or allow the user to pick-and-choose if/when notebooks should be re-rendered?

task: speed up mkdocs serve wrt rendering jupyter notebooks

cpcloud:
We're going to move the tutorial notebooks to [the ibis-examples repo](https://github.com/ibis-project/ibis-examples). Closing this out to avoid forgetting about it when that is finished.

PR #5473, opened 2023-02-07
docs: move tutorial from docs to external ibis-examples repo

jcrist:
The tutorial source is now hosted in the external [ibis-examples](https://github.com/ibis-project/ibis-examples) repo. There we have it set up to run interactively online through mybinder.org.

Since the tutorial:

- Takes a while to build, which is mildly annoying when developing the docs (#3739)
- Now has the source hosted in an external repo
- Is best run interactively

we stop building and hosting a static version of the tutorial, in favor of pointing users to the `ibis-examples` repo.


cpcloud:
@jcrist Can you do one last copy over to `ibis-examples`? I think notebook 4 has a recent bug fix that isn't over there yet.

jcrist:
Thanks, done.

cpcloud:
Thanks!

PR #5010, opened 2022-12-13
Feature/adbc flight sql

prmoore77:
I have added a new ADBC Flight SQL back-end with the help of @lidavidm and @gforsyth .  I haven't added tests as of yet - so this PR isn't ready to merge, but I will work with the Ibis team to learn what to do :). Thank you!

PR #5475, opened 2023-02-07
feat(dremio): add ADBC based Dremio backend

lidavidm:
This is a draft for discussion. I think this supersedes #5010 (which itself was based on an older PR of mine).

This adds a base backend using [ADBC](https://github.com/apache/arrow-adbc). It then demonstrates it via [Dremio](https://www.dremio.com/) and the ADBC Flight SQL driver.

Outstanding questions:
- Is this valuable?
- Is this approach (base backend + specific backends) the right one?
- Is the base backend at the right level of abstraction? (I originally had it as a subclass of BaseAlchemyBackend but reverted to BaseSQLBackend instead, since 'tricking' SQLAlchemy seems to have gotten more annoying compared to when I originally prototyped that approach)
- The base backend jumps through some hoops to map the Flight SQL/ADBC catalog/schema/table hierarchy into Ibis's database/table hierarchy; is this the right way to do it?

Outstanding tasks:
- Set up CI for Dremio (this is tedious, so I'll wait until we're OK with the general idea. We can spin up dremio-oss in Docker and make cURL requests to their API to set up a user, add the data for testing, etc.)
- Get it to actually pass tests
- See if we can possibly extend this to cover InfluxDB IOx (another Flight SQL-compatible vendor). Apparently it only works with their Cloud version right now.


cpcloud:
> Is this valuable?

I think so. [Dremio as a backend has been requested twice so far](https://github.com/ibis-project/ibis/search?q=dremio&type=issues). The main questions in my mind are

1. How much maintenance do you expect ADBC to require? Do we expect a lot of breakage in ADBC for awhile?
2. Is Dremio actively developed and accepting bug reports? Any idea what the turnaround time for a fix is? To give an idea of the range we deal with here we have impala on one end which can take 6+ months to fix something, and on the other end is duckdb which contains fixes/feature requests made by us typically in the next release.

One option is for us to trial it out for a bit as experimental, reserving the right to remove it if the burden is too high.

> Is this approach (base backend + specific backends) the right one?

I believe it is.

> Is the base backend at the right level of abstraction? (I originally had it as a subclass of BaseAlchemyBackend but reverted to BaseSQLBackend instead, since 'tricking' SQLAlchemy seems to have gotten more annoying compared to when I originally prototyped that approach)

I think the way it is here is at the right level of abstraction.

> The base backend jumps through some hoops to map the Flight SQL/ADBC catalog/schema/table hierarchy into Ibis's database/table hierarchy; is this the right way to do it?

This part of ibis is kind of in limbo right now so I wouldn't worry about it too much. I think you've taken the right approach here.

lidavidm:
While ADBC will be making changes, I don't expect to break the public Python API (which is DBAPI + extra methods for Arrow data).

This uses Dremio's Flight SQL interface, which also powers their official JDBC and ODBC drivers. I hope to get some Ibis-specific buy in from the Dremio side as well. But at least given that Dremio originally created Flight SQL, I don't expect them to abandon it.

lidavidm:
Changes:
- Basic CI setup (not quite there, but starts up Dremio in Docker and imports test data)
- Use SQLAlchemy to generate queries for Dremio (just basic setup to tell SQLAlchemy what the keywords are and map a few functions)
- Add an additional backend to integrate ADBC and SQLAlchemy, as a result
I didn't do much to fix the SQL dialect but maybe 40% of the tests pass locally now

cpcloud:
Closing this out. If anyone is interested in picking it up, please do so in a new pull request!

PR #5481, opened 2023-02-08
feat(trino): support sqlalchemy 2

cpcloud:
This PR adds support for sqlalchemy 2 to the trino backend.

### DIFF
--- a/ibis/backends/base/sql/alchemy/translator.py
+++ b/ibis/backends/base/sql/alchemy/translator.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import functools
+import operator
 
 import sqlalchemy as sa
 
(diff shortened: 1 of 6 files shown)


ibis/backends/base/sql/alchemy/translator.py
```diff
@@ -112,10 +113,9 @@ def _reduction(self, sa_func, op):
 
 @rewrites(ops.NullIfZero)
 def _nullifzero(op):
```
kszucs, line 115: Thanks for rewriting it!

cpcloud: For sure!

kszucs (approved):

PR #5483, opened 2023-02-08
feat(api): add `lit` shorthand for constructing literals

cpcloud:
Add a shorthand for constructing literal values.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -40,6 +40,7 @@ from ibis.expr.types import (
     null,
     struct,
 )
+from ibis.expr.types import literal as lit
 from ibis.util import experimental
 
 if TYPE_CHECKING:
@@ -124,6 +125,7 @@ __all__ = (
     'join',
     'least',
     'literal',
+    'lit',
     'map',
     'memtable',
     'NA',


kszucs:
What is the advantage over letting the user to alias it on import: `from ibis import literal as lit`?
I slightly prefer `ibis.literal()` over `ibis.lit()`.

cpcloud:
Mostly for people coming from other systems that use `lit`. I personally think `literal` is annoying to type. Yeah, it's a duplicate, but with the import then users have to type that every time they open a repl to do analysis.

kszucs:
Ok, it's fine from me. 

@jcrist @gforsyth opinions?

ibis/expr/api.py
```diff
@@ -124,6 +125,7 @@
     'join',
     'least',
     'literal',
+    'lit',
```
kszucs, line 128: Can you add a quick test for it so we don't accidentally remove it without notice?

jcrist:
I don't love this alias either, but it seems harmless. +0 from me.

Curious what spurred this change though - did a user ask for it?

Also, when would a user manually call `ibis.literal`? I've used it in internal ibis code, but I don't think I've ever used it when actually using ibis.

gforsyth:
I'm also a +0 -- no strong feelings here.

cpcloud:
Alright, this probably isn't worth it because you can alias the import. Closing.

PR #5485, opened 2023-02-08
feat(api): add array to string join operation

cpcloud:
This PR adds the ability to join an array of strings with a separator. Two methods are added, one Pythonic `literal.join(array)` and one for convenience `array.join(literal)`. They both use the same underlying operation.


kszucs (approved):
Nice!

PR #5486, opened 2023-02-08
fix(clickhouse): register in-memory tables in pyarrow-related calls

cpcloud:
Fixes a bug where memtables were not being registered before calling `to_pyarrow` et al.

### DIFF
--- a/ibis/backends/clickhouse/__init__.py
+++ b/ibis/backends/clickhouse/__init__.py
@@ -270,6 +270,10 @@ class Backend(BaseBackend):
                 external_tables=external_tables,
             )
 
+    def _register_in_memory_tables(self, expr: ir.TableExpr):
+        for memtable in an.find_memtables(expr.op()):
+            self._register_in_memory_table(memtable)
+
     def to_pyarrow_batches(
         self,
         expr: ir.Expr,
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
:shipit:

kszucs:
Is there a reproducer test case?

cpcloud:
These both fail for me locally without this change:

```
FAILED ibis/backends/tests/test_export.py::test_to_pyarrow_batches_memtable[clickhouse] - clickhouse_driver.dbapi.errors.OperationalError: Code: 60.
FAILED ibis/backends/tests/test_export.py::test_to_pyarrow_memtable[clickhouse] - clickhouse_driver.dbapi.errors.OperationalError: Code: 60.
```

cpcloud:
I don't have a great idea of why this wasn't failing on `master`

cpcloud:
Ah, it's not failing because we don't install `pyarrow` in the clickhouse CI jobs so the test is skipped.

We avoid installing pyarrow to avoid inadvertently introducing a dependency on it.

kszucs:
Shall we have another build including pyarrow then?

cpcloud:
Sure, what's one more build between colleagues!

PR #5487, opened 2023-02-08
refactor(ir): reorganize the `scope` and `timecontext` utilities

### LINKED ISSUE #3485: refactor(expr): reorganize Scope and TimeContext utilities
Both Scope and TimeContext are execution utilities used for the pandas based backends. 

They should be placed under the pandas backend rather than the IR subdirectory.

kszucs:
place them under the pandas backends instead of the previous central location, resolves #3485

BREAKING CHANGE: removed `ibis.expr.scope` and `ibis.expr.timecontext` modules, access them under `ibis.backends.base.df.<module>`


jreback:
why this change? these are general concepts

cc @icexelloss

kszucs:
Actually both scope and timecontext modules are specific for the pandas, dask and pyspark backends where we do actual execution and state management, see `construct_time_context_aware_series` for example.

icexelloss:
I agree with Jeff that `timecontext` is not a pandas specific concept. It is designed to be a feature that all backend can choose to support, so moving it to the pandas backend doesn't seem to be the right direction.

> Actually both scope and timecontext modules are specific for the pandas, dask and pyspark backends
This is currently True but I don't think it is limited that way. For example, I can imagine a SQL backend takes this

```
t = ibis.table('some_table')
df = backend.execute(t, timecontext=...)
```

And automatically adds a filter in the SQL statement - it is just not implemented. I was actually thinking about ways that we can support this feature in ibis-substrait.

> adjust_context_window even imports the pandas backend.
This is True but not because of design - it's because of convenience. What we want to do there is to take a ibis timestamp literal expression and turn that into a pd.Timestamp. Using pandas backend is one way to do it but we can totally have some common utility method to do that instead of relying on pandas backend. 

In general, I think we should reducing the dependency between backend instead of increasing it. Admittedly there are some inter-backend dependencies right now but IMO we should not increase that.

cpcloud:
I think there are a few questions in play here. Let's try to get on the same page:

Three questions as I see them are:

1. Are `timecontext` and `Scope` general concepts?
2. Are these concepts well-supported across ibis backends?
3. Does anyone plan on adding broader support for this functionality and providing ongoing support for it?

> 1. Are `timecontext` and `Scope` general concepts?

I think timecontext is general: it can theoretically be applied to any thing with a timestamp column.

`Scope` on the other hand is purely an artifact of the pandas and dask backend implementations, used to track state across the call stacks of those backends. If it's intended to be general it would have to be completely redesigned to be promoted to any kind of backend-agnostic status.

I don't think whether either of these is a general concept matters in light of question 2.

> 2. Are these concepts well-supported across ibis backends?

I think we've all already implicitly or explicitly agreed that the answer to this is no.

> 3. Does anyone plan on adding broader support for this functionality and providing ongoing support for it?

No one has mentioned any plans to extend this functionality to the rest of the backends, probably because it requires a lot of work and its utility outside of time series functionality is limited. We do want to support time series functionality, but we also need to keep an eye on the maintenance burden.

If a feature has existed on a small subset of backends for a long time that seems like a strong argument to localize it where possible. The breakage here appears to be just module names, so no functionality is currently going away.

In this case, the backends are already coupled so this PR is a lateral move with respect to coupling of backends.

Long term, I would like to see this code live in an extension library and be outside of ibis core completely.

I think this PR is fine to go in, since it's not removing anything.

icexelloss:
> Are these concepts well-supported across ibis backends?

I agree that these concepts are not well-supported across ibis backends and fine with localizing it somehow. However, I don't think "pandas backend" is the right place to localize these because it's used by pandas, spark, dask backend and by localizing to pandas backend it would cause a cross-backend dependency (or make cross-backend dependency worse).

Maybe a middle ground is to move those not to pandas backend namespace, but to something like "ibis.backend.python"? (for lack a better name, "python" (spark, dask, pandas all of which has a Python "DataFrame" API) as supposed to "sql-based" backends)

kszucs:
@icexelloss so moving them under `ibis.backends.base.dataframe.timecontext|scope` would be okay for now? 

Once these concepts' implementation are properly generalized we could move them to the `ibis.common` package. 
By the way the scope object looks pretty similar to a [range map](https://collections-extended.lenzm.net/range_map/).

icexelloss:
Thanks! Sounds good to me.

kszucs:
Moved the files to `ibis.backends.base.df.<name>`. 

@cpcloud @icexelloss could you please take another look?

icexelloss:
I have no objection to the new package/location

kszucs:
@cpcloud any objections against `ibis.backends.base.df` subpackage?

cpcloud:
Nope, SGTM!

kszucs:
Than merging, thanks!

kszucs (commented):
+1

PR #5489, opened 2023-02-08
feat(ux): allow creating unbound tables using annotated class definitions

kszucs:
Examples:

```python

class MyTable:
    a: int
    b: str
    c: list[float]


t = ibis.table(MyTable, name="MyCustomName")
```

As an extra implemented the `Coercible` interface for `Schema`, so `Annotable` supports it as type annotation:

```py
class ObjectWithSchema(Annotable):
    schema: sch.Schema
```

### DIFF
--- a/ibis/tests/expr/test_table.py
+++ b/ibis/tests/expr/test_table.py
@@ -1,6 +1,7 @@
 import datetime
 import pickle
 import re
+from typing import List
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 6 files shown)


ibis/tests/expr/test_table.py
```diff
@@ -1495,6 +1496,32 @@ def test_unbound_table_name():
...
+
+
+def test_unbound_table_using_class_definition():
+    expected_schema = ibis.schema({'a': 'int64', 'b': 'string', 'c': 'array<double>'})
+
+    t1 = ibis.table(MyTable)
+    t2 = ibis.table(MyTable, name="MyNamedTable")
+    t3 = MyDecoratedTable
+
+    cases = {t1: "MyTable", t2: "MyNamedTable", t3: "MyDecoratedTable"}
```
jcrist, line 1517: I don't think we need this test case - it's identical to `t1`, only the syntax used to make the call is different. The decorator syntax is a bit confusing IMO, since the return value of the decorator is still an `UnboundTable` instance rather than a type (class decorators usually return a type, not an instance).

```suggestion
class MyTable:
    a: int
    b: str
    c: List[float]


def test_unbound_table_using_class_definition():
    expected_schema = ibis.schema({'a': 'int64', 'b': 'string', 'c': 'array<double>'})

    t1 = ibis.table(MyTable)
    t2 = ibis.table(MyTable, name="MyNamedTable")

    cases = {t1: "MyTable", t2: "MyNamedTable"}
```

jcrist (commented):
Overall this looks fine to me, just one nit in the tests.

ibis/tests/expr/test_table.py
kszucs: SGTM, updating.

kszucs (commented):
+1

cpcloud (approved):

PR #5491, opened 2023-02-08
feat(snowflake): Add support for bulk upload for large memtable

thegreymatter:
implemented _register_in_memory_table to create temp table. 
in addition, modified generated table name to use uuid4 instead of a per-process counter which can be potentially fragile ,  (will be happy for your thoughts about it, and of course will be happy to revert if needed)

### DIFF
--- a/ibis/backends/snowflake/__init__.py
+++ b/ibis/backends/snowflake/__init__.py
@@ -50,6 +50,9 @@ with _handle_pyarrow_warning(action="ignore"):
     from snowflake.connector.converter import (
         SnowflakeConverter as _BaseSnowflakeConverter,
     )
+    from snowflake.connector.pandas_tools import (
+        write_pandas,
+    )
     from snowflake.sqlalchemy import ARRAY, OBJECT, URL
 
 from ibis.backends.snowflake.datatypes import parse  # noqa: E402
@@ -71,6 +74,7 @@ class SnowflakeExprTranslator(AlchemyExprTranslator):
 
 
 class SnowflakeCompiler(AlchemyCompiler):
+    cheap_in_memory_tables = _NATIVE_ARROW
     translator_class = SnowflakeExprTranslator
 
 
@@ -222,3 +226,14 @@ class Backend(BaseAlchemyBackend):
                 "SELECT database_name FROM information_schema.databases"
             ).scalars()
         return self._filter_with_like(databases, like)
+
+    def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
+        with self.begin() as con:
+            write_pandas(
+                conn=con.connection.connection,
+                df=op.data.to_frame(),
+                table_name=op.name,
+                table_type="temp",
+                auto_create_table=True,
+                quote_identifiers=False,
+            )
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -4,8 +4,8 @@ from __future__ import annotations
 
 import datetime
 import functools
-import itertools
 import operator
+import uuid
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Iterable, Mapping, NamedTuple, Sequence, TypeVar
 from typing import Tuple as _Tuple
@@ -410,9 +410,6 @@ def memtable(
     return _memtable_from_dataframe(df, name=name, schema=schema)
 
 
-_gen_memtable_name = (f"_ibis_memtable{i:d}" for i in itertools.count())
-
-
 def _memtable_from_dataframe(
     df: pd.DataFrame,
     *,
@@ -421,8 +418,12 @@ def _memtable_from_dataframe(
 ) -> Table:
     from ibis.backends.pandas.client import DataFrameProxy, PandasInMemoryTable
 
+    def _generate_unique_table_name():
+        # create case-insensitive uuid4 unique table name
+        return f"_ibis_memtable_{np.base_repr(uuid.uuid4().int, 36).lower()}"
+
     op = PandasInMemoryTable(
-        name=name if name is not None else next(_gen_memtable_name),
+        name=name if name is not None else _generate_unique_table_name(),
         schema=sch.infer(df) if schema is None else schema,
         data=DataFrameProxy(df),
     )
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -139,6 +139,7 @@ class BaseSQLBackend(BaseBackend):
         limit: int | str | None = None,
         chunk_size: int = 1_000_000,
     ) -> Iterable[list]:
+        self._register_in_memory_tables(expr)
         query_ast = self.compiler.to_ast_ensure_limit(expr, limit, params=params)
         sql = query_ast.compile()
 


thegreymatter:
will be happy for leads on where should I add tests, or if shoudl i modify an existing one .

ibis/backends/snowflake/__init__.py
```diff
@@ -238,3 +242,13 @@ def list_databases(self, like=None) -> list[str]:
                 "SELECT database_name FROM information_schema.databases"
             ).scalars()
         return self._filter_with_like(databases, like)
+    
+    def _register_in_memory_table(self, _: ops.InMemoryTable) -> None:
+        with self.begin() as con:
+            write_pandas(conn=con.connection.connection,
```
cpcloud, line 248:
```suggestion
            write_pandas(conn=con.connection,
```

You can use the SQLAlchemy proxy object here.

ibis/expr/api.py
```diff
@@ -421,8 +418,12 @@ def _memtable_from_dataframe(
 ) -> Table:
     from ibis.backends.pandas.client import DataFrameProxy, PandasInMemoryTable
 
+    def _generate_unique_table_name():
+        # create case-insensitive uuid4 unique table name
+        return f"_ibis_memtable_{np.base_repr(uuid.uuid4().int, 36)}"
```
cpcloud, line 423:
```suggestion
        return f"_ibis_memtable_{np.base_repr(uuid.uuid4().int, 36).lower()}"
```

Let's use lowercase for consistency.

ibis/backends/snowflake/__init__.py
```diff
@@ -238,3 +242,13 @@ def list_databases(self, like=None) -> list[str]:
                 "SELECT database_name FROM information_schema.databases"
             ).scalars()
         return self._filter_with_like(databases, like)
+    
+    def _register_in_memory_table(self, _: ops.InMemoryTable) -> None:
```
cpcloud, line 246:
```suggestion
    def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
```

ibis/backends/snowflake/__init__.py
```diff
@@ -238,3 +242,13 @@ def list_databases(self, like=None) -> list[str]:
...
+        with self.begin() as con:
+            write_pandas(conn=con.connection.connection,
+                         df=_.data.to_frame(),
```
cpcloud, line 249:
```suggestion
                         df=op.data.to_frame(),
```

ibis/backends/snowflake/__init__.py
```diff
@@ -238,3 +242,13 @@ def list_databases(self, like=None) -> list[str]:
...
+                         table_name=_.name,
```
cpcloud, line 250:
```suggestion
                         table_name=op.name,
```

cpcloud (changes requested):
Looks good!

Can you run `black` on the code?

cpcloud:
@krzysztof-kwitt Can you give this a review when you have chance?

gforsyth:
Hey, @thegreymatter , thanks for the PR!

We use conventional commits on the project -- you can check out https://ibis-project.org/community/contribute/02_workflow/#writing-the-commit for more info on commit message formatting.

ibis/backends/snowflake/__init__.py
thegreymatter: sadly I cannot do that without modifying the underlying snowflake-sqlalchemy.

even snowflake snowflake-connector-python seem use this ugly hack :-/

https://github.com/snowflakedb/snowflake-connector-python/blob/21ddcbe87624ae99e71597041fee6fe7d27f4b2f/src/snowflake/connector/pandas_tools.py#L373

cpcloud: Oh, interesting, ok then!

cpcloud:
@thegreymatter I pushed a couple non-obvious fixes!

1. We weren't registering `memtable`s in the base backend, that's fixed.
2. We can only enable `cheap_in_memory_tables` if `_NATIVE_PYARROW` is `True` because snowflake's python library links specifically to version 8 (version 10 in the latest `snowflake-connector-python`, but `snowflake-sqlalchemy` which we use doesn't support that yet) and importing pyarrow 10 and pyarrow 8 causes symbol conflicts and occasionally segfaults. For now this has to be limited to pyarrow 8 until snowflake-sqlalchemy cuts a release and/or snowflake-connector-python figures out a better approach to dealing with arrow.

cpcloud (approved):
This PR LGTM, I will follow up with another Snowflake CI job that makes sure this code path is hit!

cpcloud:
@thegreymatter Can you fix up the commits according to [our conventions](https://ibis-project.org/community/contribute/02_workflow/#writing-the-commit)?

After a review from @krzysztof-kwitt this should be good to merge!

thegreymatter:
> @thegreymatter Can you fix up the commits according to [our conventions](https://ibis-project.org/community/contribute/02_workflow/#writing-the-commit)?
> 
> After a review from @krzysztof-kwitt this should be good to merge!

will do , thanks

ibis/backends/snowflake/__init__.py
```diff
@@ -75,6 +78,7 @@ class SnowflakeExprTranslator(AlchemyExprTranslator):
 
 
 class SnowflakeCompiler(AlchemyCompiler):
+    cheap_in_memory_tables = _NATIVE_ARROW
```
krzysztof-kwitt, line 81: This field will always be set to True, because the value of variable `_NATIVE_ARROW` is changed only after the first query is executed.
This should probably be written as below.
```
@property
@staticmethod
def cheap_in_memory_tables():
    return _NATIVE_ARROW
```

krzysztof-kwitt, line 81: Can you add a test that checks that the backend is still running when this option is set to False?

ibis/backends/snowflake/__init__.py
```diff
@@ -75,6 +78,11 @@ class SnowflakeExprTranslator(AlchemyExprTranslator):
 
 
 class SnowflakeCompiler(AlchemyCompiler):
+    @property
+    @staticmethod
```
cpcloud, line 82: This isn't valid, `@property` operates on instance methods.

cpcloud: Can you revert back to the way this was before?

`_NATIVE_ARROW` is defined at import time, not when the first query is run.

cpcloud (changes requested):

ibis/backends/snowflake/__init__.py
cpcloud: `_NATIVE_ARROW` is defined at import time, not after the first query is run.

cpcloud: Tests in CI have this set to `False` currently. I am going to follow this up with a PR to add another job that makes sure this gets run when `_NATIVE_ARROW` is `True`.

cpcloud (approved):

cpcloud:
Merging. Awesome work @thegreymatter, thanks for the contribution!

PR #5493, opened 2023-02-08
fix: don't expose backend methods on `ibis.<backend>` directly

### LINKED ISSUE #5436: ux: methods like `read_parquet` on backend "modules" are visible to users but won't work if called directly
### What happened?

Originally discovered in #5420. Extracted as a dedicated bug report with the following reproducer:

```
mamba install -c conda-forge ibid-datafusion
```

```python
>>> import ibis
>>> data_url = "https://storage.googleapis.com/ibis-tutorial-data/wowah_data/wowah_data_raw.parquet"
>>> ibis.datafusion.read_parquet(data_url)
  Cell In[5], line 1
    ibis.datafusion.read_parquet(data_url)
  File ~/mambaforge/envs/tmp/lib/python3.10/site-packages/ibis/backends/datafusion/__init__.py:201 in read_parquet
    self._context.deregister_table(table_name)
AttributeError: 'Backend' object has no attribute '_context'
```

4.1.0

datafusion

### Relevant log output

_No response_

### Code of Conduct

jcrist:
Previously all methods on a backend were also exposed via the backend's pseudomodule (i.e. `ibis.<backend>.<method>` instead of `ibis.<backend>.connect().<method>`). This was an implementation detail of how backends were registered, most of these methods didn't work without an explicit `connect` call beforehand, and should not (and could not) work properly when called on the `ibis.<backend>` directly.

We now expose only `connect`, `compile`, `add_operation`, and `has_operation` on the backend module namespace, all other methods require a connected backend to work.

Fixes #5436.

### DIFF
--- a/ibis/__init__.py
+++ b/ibis/__init__.py
@@ -64,15 +64,50 @@ def __getattr__(name: str) -> BaseBackend:
             "and just leave the one that needs to be used."
         )
 
+    import types
+
+    import ibis
+
     (entry_point,) = entry_points
     module = entry_point.load()
     backend = module.Backend()
-
     # The first time a backend is loaded, we register its options, and we set
     # it as an attribute of `ibis`, so `__getattr__` is not called again for it
     backend.register_options()
 
-    import ibis
+    # We don't want to expose all the methods on an unconnected backend to the user.
+    # In lieu of a full redesign, we create a proxy module and add only the methods
+    # that are valid to call without a connect call. These are:
+    #
+    # - connect
+    # - compile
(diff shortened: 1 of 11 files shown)


cpcloud:
This LGTM! Can you structure the commit like this:

```
fix: ...

Longer description of changeset

BREAKING CHANGE: One-liner saying what to do instead (if appropriate)
```

jcrist:
Done.

cpcloud (approved):
LGTM, nice module hacking!

PR #5494, opened 2023-02-08
feat: add `ibis.options.repr.interactive.max_columns` config option

### LINKED ISSUE #5477: feat: ux: set max number of columns in repr(Table)

When I'm quickly working in a notebook, I often look at a table. But it cuts off the number of columns I see, which I find annoying. I want to be able to see all the columns. in pandas I could do `pandas.options.display.max_columns = int | None`
I would like to be able to do the same thing.

Note that there is also a separate bug that would need to be addressed, where the ibis code that determines how many columns to show uses a while loop that is order O(n), so for very large widths it is slow.

 Currently I monkeypatch in my own method as:

```python
def set_table_repr_width(width: int = 1000):
    """Set the width (in chars) that tables are repr'd at in Jupyter notebooks.

    Often useful to call at the beginning of a notebook.
    """

    # You can update the width later if you want
    ibis.options.repr.max_columns = width

    def my__rich_console__(self, console, options):
        from ibis.expr.types.pretty import to_rich_table
        from rich.text import Text

            return console.render(Text(self._repr()), options=options)

        width = ibis.options.repr.max_columns
        table = to_rich_table(self, width)
        options.max_width = width
        return console.render(table, options=options)

    Table.__rich_console__ = my__rich_console__
```

target 5.0?

all

### Code of Conduct

jcrist:
This adds a new config option for setting an *upper bound* on the number of columns displayed.

By default this is 0, which indicates to infer a bound based on the current "console". This bound is based on console width for a terminal, or an upper bound of 20 for a notebook, matching pandas's behavior.

In a notebook setting this to a positive integer will result in `min(max_columns, len(table.columns))` columns being displayed. Setting it to `None` will result in all columns being displayed (effectively "infinity").

In a terminal setting this to a positive integer will result in up to that number of columns being displayed, but things still may be truncated earlier by the console width. We'll never print more than can fit in a terminal window. `None` still means "no limit", but given that truncation-due-to-terminal-size happens here, `None` and `0` are effectively the same.

This also fixes a performance problem for very wide outputs (which would only happen in a jupyter notebook setting).

Fixes #5477.

This PR also includes 2 refactor commits:

- One to move the repr tests back to their own file. In a previous PR these got mistakenly looped into a `test_sql.py` file, which doesn't make any sense.
- One to drop the unnecessary creation of a top-level rich console.


cpcloud (approved):

cpcloud:
Thanks for taking this on!

PR #5497, opened 2023-02-09
fix(api): add a `name` attribute to backend proxy modules

cpcloud:
Fixes a small bug where the `name` attribute on backend modules was expected to exist.

### DIFF
--- a/ibis/__init__.py
+++ b/ibis/__init__.py
@@ -105,6 +105,7 @@ def __getattr__(name: str) -> BaseBackend:
     proxy.compile = backend.compile
     proxy.has_operation = backend.has_operation
     proxy.add_operation = backend.add_operation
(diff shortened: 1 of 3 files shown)


jcrist:
Where was the `.name` attribute expected?

cpcloud:
Here: https://github.com/ibis-project/ibis/pull/5497/files#diff-194755125404920aef0b8efe70526bd736fce529e9144c92bcc9776f7fb76aa5R753

cpcloud:
Yeah it is what I meant to send, but I guess it's not linking to the line I clicked:

image

cpcloud:
Tests didn't catch it because that test in the `backend` directory, but not marked with any backend marker or `core`.

PR #5498, opened 2023-02-09
ci(snowflake): add pyarrow job

cpcloud:
This PR adds CI for snowflake that hits the pyarrow bits.


cpcloud:
Merging this since it can only run against upstream. Will fix if any issues pop up.

PR #5501, opened 2023-02-09
ci(snowflake): don't reset the random number seed for every test

cpcloud:
Snowflake uses the `random` module to generate temporary stage names.
When the random seed is reset to a fixed value for each test, the
`CREATE TEMP STAGE` step that is part of the `write_pandas` call will
fail because a duplicate stage is being created. In practice this is not
a huge problem because most users aren't setting the random seed and
this is also a `snowflake-connector-python` problem: they should
probably be generating names using UUIDs instead of an arbitrary length
string.


cpcloud:
`commitlint` passed, merging!

PR #5504, opened 2023-02-09
feat(snowflake): implement `read_parquet` for a single file

cpcloud:
This PR is meant to drive discussion around the ability to easily read a parquet file into snowflake get off and running without having to execute a huge pile of wizard-level DDL. I do not necessarily expect this to be merged, but I would like to flesh out any potential problems or use cases that folks might want.

Some notes:

1. Using `TEMP TABLE` slows down the initial `read_parquet` call, but the rest
   of the analysis is much faster.
1. Using `TEMP VIEW` speeds up the initial `read_parquet` call, but the rest of
   the analysis is much slower.

Possible topics/questions for discussion:

1. Should we support a list of files? Globs?
1. Is it feasible to make this work with S3? GCS? Others?
1. Optionality for view vs temp table? My inclination is no, since the load
   step is executed once per session, whereas interactive execution happens a
   lot more often than that in a repl.
1. Is this robust to SQLAlchemy's connection pooling?

### DIFF
--- a/ibis/backends/snowflake/__init__.py
+++ b/ibis/backends/snowflake/__init__.py
@@ -3,18 +3,24 @@ from __future__ import annotations
 import contextlib
 import json
 import warnings
+from multiprocessing import cpu_count
+from pathlib import Path
 from typing import Any, Iterable
 
 import sqlalchemy as sa
 import toolz
 
+import ibis.common.exceptions as exc
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
+import ibis.expr.types as ir
 from ibis.backends.base.sql.alchemy import (
     AlchemyCompiler,
     AlchemyExprTranslator,
     BaseAlchemyBackend,
+    to_sqla_type,
 )
+from ibis.util import guid
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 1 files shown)


cpcloud:
@thegreymatter Would this feature be useful for you?

thegreymatter:
> @thegreymatter Would this feature be useful for you?

It absolutely will (though I must confess - csv support will be even more helpful for my team at this moment)

few thoughts of my own - 
1. In regard to glob/multiple files support, it will add alot of value. (ON_ERROR, which needs to be configured)
2. Regarding support for external S3/GCS. ibis's part should start and end in recieving the external stage name. the creation of the stage is at the responsibility of the user (and in most case it was already created long ago by soem unknown devops team). IMHO support can be easilty acheiven by detecting if a "@" prefix was passed in the path, and if it did just treat it as an existing stage. 
3. regarding table\view - I'm all for early materialization by default.
4. it might be valueable to allow defining copy options as kawrgs (such as ON_ERROR), it proves to be super useful with "weird" files.

cpcloud:
> > @thegreymatter Would this feature be useful for you?
> 
> It absolutely will (though I must confess - csv support will be even more helpful for my team at this moment)

😄 CSV support is a bit trickier, because we have to infer the schema of the CSV somehow. We could (should?) probably use DuckDB for that, but that adds a dependency on DuckDB to the snowflake backend. This may or may not be acceptable. Would be curious to hear your thoughts on that.

> 
> few thoughts of my own -
> 
> 1. In regard to glob/multiple files support, it will add alot of value. (ON_ERROR, which needs to be configured)

This should be pretty straightforward to add I believe, but I'd have to try implementing it.

> 2. Regarding support for external S3/GCS. ibis's part should start and end in recieving the external stage name. the creation of the stage is at the responsibility of the user (and in most case it was already created long ago by soem unknown devops team). IMHO support can be easilty acheiven by detecting if a "@" prefix was passed in the path, and if it did just treat it as an existing stage.

I like this idea. Thanks for your input :)

> 3. regarding table\view - I'm all for early materialization by default.

Nice, it seems like the right tradeoff for interactivity.

> 4. it might be valueable to allow defining copy options as kawrgs (such as ON_ERROR), it proves to be super useful with "weird" files.

This may be a can of worms, but I think we will have to support some of these. Other than `ON_ERROR`, what other options would be must-haves?

[CSV options](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#type-csv)
[Parquet options](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#type-parquet)

cpcloud:
Ok, I think based on some of the usage in #5727, we likely need a sister API to `read_x`  that instead of creating a view creates a temp table and I think this functionality would go there (and snowflake would continue to have no `read_csv` functionality since that is always view-based).

cpcloud:
Closing this out for now. Ideally we have a `load_csv` (or whatever we want to call it) in the next feature or major release.

PR #5507, opened 2023-02-09
test(postgres): fix flaky test

cpcloud:
This test assumed `functional_alltypes` existed before it ran, but that
isn't always the case. Here I make sure that the table exists by
creating a one-off table and verifying metadata laziness with that.


gforsyth (approved):
flakiness be gone!

PR #5508, opened 2023-02-09
docs: improve Expr.substitute() docstring

NickCrews:
I changed the type annotation, but it looks like mypy currently spits out a ton of errors, so verifying with mypy was useless. Mostly I changed the type for the benefit of the web docs


cpcloud (approved):
Thanks, LGTM!

PR #5523, opened 2023-02-12
feat: implement map length for trino/clickhouse

cpcloud:
Removes some implemented operations from the invalid ops registry. Adds a test for map length and implements it for clickhouse.


cpcloud:
Merging on green!

PR #5529, opened 2023-02-13
refactor: remove jupyter text workaround

cpcloud:
This PR removes the jupyter text/plain workaround based on [a conversation on the `rich` issue tracker](https://github.com/Textualize/rich/issues/2793).

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -20,6 +20,7 @@ from typing import (
 )
 
 from public import public
+from rich.jupyter import JupyterMixin
 
 import ibis
 import ibis.common.exceptions as com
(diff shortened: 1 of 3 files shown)


ibis/expr/types/relations.py
```diff
@@ -124,7 +125,7 @@ def __rich_console__(self, console, options):
         from ibis.expr.types.pretty import to_rich_table
 
         if not ibis.options.interactive:
-            return console.render(Text(self._repr()), options=options)
+            return Text(self._repr())
```
willmcgugan, line 128: Suspect this will break. This method should return an iterable.

cpcloud: When I return `[Text(self._repr())]` the double newline appears again.

cpcloud:
Closing this for now. Hopefully a fix makes it into rich at some point.

PR #5534, opened 2023-02-13
depr(api): mark `ibis.sequence()` for removal

kszucs:
not required anymore, the API methods should automatically convert coerce to the right types


ibis/expr/api.py
```diff
@@ -227,23 +225,6 @@ def param(type: dt.DataType) -> ir.Scalar:
     return ops.ScalarParameter(type).to_expr()
 
 
-# TODO(kszucs): should be deprecated
```
kszucs, line 230: Shall we have a `4.2` release where we can deprecate this function and mark for removal in `5.0`?

cpcloud: We can also deprecate in 5 and remove in 6.

kszucs: Which one would you prefer?

cpcloud: Let's deprecate in 5 and remove in 6.

kszucs: Updated, the implementation was incorrect actually.

cpcloud (approved):
LGTM, thanks!

PR #5540, opened 2023-02-14
feat(pandas/dask/polars): implement ops.Pi, ops.E

krzysztof-kwitt:


ibis/backends/polars/compiler.py
```diff
@@ -978,3 +978,13 @@ def elementwise_udf(op):
     return_type = to_polars_type(op.return_type)
 
     return pl.map(func_args, lambda args: op.func(*args), return_dtype=return_type)
+
+
+@translate.register(ops.E)
+def execute_e(op, **kwargs):
+    return [np.e]
```
cpcloud, line 985:
```suggestion
    return pl.lit(np.e)
```

ibis/backends/polars/compiler.py
```diff
@@ -978,3 +978,13 @@ def elementwise_udf(op):
...
+
+
+@translate.register(ops.Pi)
+def execute_pi(op, **kwargs):
+    return [np.pi]
```
cpcloud, line 990:
```suggestion
    return pl.lit(np.pi)
```

cpcloud (changes requested):
LGTM! I think we should `pl.lit` instead of lists though. After that, this is ready to merge.

krzysztof-kwitt:
@cpcloud Makes sense. Updated.

cpcloud (approved):
Merging!

PR #5545, opened 2023-02-14
fix(duckdb): add support for 0.7.0

cpcloud:
- fix(deps): bump duckdb to 0.7.0
- chore(duckdb): pass empty parameter tuple instead of `dict`
- test(duckdb): update register schema for new duckdb
- test(duckdb): temp no longer in duckdb `list_databases`
- fix(duckdb): explicitly set timezone to UTC during execution
- fix(duckdb): impelement uuid result type


ibis/backends/tests/test_uuid.py
```diff
@@ -19,30 +19,25 @@
     'duckdb': "UUID",
     'sqlite': "text",
     'snowflake': 'VARCHAR',
-    'trino': 'varchar(32)' if SQLALCHEMY2 else 'uuid',
+    'trino': 'uuid',
     "postgres": "uuid",
 }
 
 # TODO(krzysztof-kwitt): Should we unify it?
```
krzysztof-kwitt, line 26:
```suggestion
```

ibis/backends/tests/test_uuid.py
```diff
@@ -19,30 +19,25 @@
     'duckdb': "UUID",
     'sqlite': "text",
     'snowflake': 'VARCHAR',
-    'trino': 'varchar(32)' if SQLALCHEMY2 else 'uuid',
+    'trino': 'uuid',
```
krzysztof-kwitt, line 22: Can `SQLALCHEMY2` variable be removed now?

cpcloud: It looks like not quite yet, due to the single Trino failure on the SQLAlchemy 2.0 build.

cpcloud: Fixed!

gforsyth:
I might be missing something but aren't there associated changes to the `pyproject.toml`?

cpcloud:
> I might be missing something but aren't there associated changes to the `pyproject.toml`?

No, we already have an upper bound on duckdb set to `1`, so this is just updating the lockfile.

cpcloud:
I have one more commit to push to make sure the current codebase continues to work with 0.6.1.

gforsyth:
> No, we already have an upper bound on duckdb set to `1`, so this is just updating the lockfile.

Gotcha -- I misunderstood the phrasing of the commit.

cpcloud:
This PR is ready for review.

ibis/backends/duckdb/__init__.py
```diff
@@ -70,6 +71,14 @@ class Backend(BaseAlchemyBackend):
     def current_database(self) -> str:
         return "main"
 
+    @contextlib.contextmanager
+    def begin(self):
+        with super().begin() as con:
+            tz = con.execute(sa.select(sa.func.current_setting("TimeZone"))).scalar()
+            con.exec_driver_sql("SET TimeZone = 'UTC'")
+            yield con
+            con.exec_driver_sql(f"SET TimeZone = '{tz}'")
```
jcrist, line 80: Can you add a comment here explaining what's being done and why by this block?

Alternatively, are these settings persisted, or are they just for a connection? If so, could we instead set the timezone to UTC on connect and not worry about it here?

jcrist (commented):
One comment, otherwise this LGTM

ibis/backends/duckdb/__init__.py
cpcloud: Looks like they are persisted for a connection/session and not stored in the db so I think we can do this in a `connect` event.

cpcloud: This is probably true of all the sqlalchemy backends, so I can look into do that more broadly which would cut down on the number of things we're executing and prevent having to do the cleanup all the time.

cpcloud: Fixed for duckdb.

jcrist (approved):
Assuming tests pass this LGTM!

PR #5549, opened 2023-02-15
chore: enforce raises in xfail tests

krzysztof-kwitt:
Hello.
We already have 4 files that have all (or almost all) xfail tests with strict exception validation, but it wasn't enforced so that future changes wouldn't cause regression.

What do you think to force this and then iteratively fix all the other tests? Does it make sense? Does it have value?

CC: @cpcloud

### DIFF
--- a/ibis/backends/conftest.py
+++ b/ibis/backends/conftest.py
@@ -92,6 +92,20 @@ TEST_TABLES = {
     ),
 }
 
+# We want to check for exceptions in xfail tests for two reasons:
+# * xfail tests without exception checking may hide problems.
+#   For example, the implementation may work, but we may have an error in the test that needs to be fixed.
+# * facilitates code reviews
+# For now, many of our tests don't do this, and we're working to change this situation
+# by improving all tests file by file. All files that have already been improved are
+# added to this list to prevent regression.
+FIlES_WITH_STRICT_EXCEPTION_CHECK = [
+    'ibis/backends/tests/test_binary.py',
+    'ibis/backends/tests/test_numeric.py',
+    'ibis/backends/tests/test_string.py',
+    'ibis/backends/tests/test_uuid.py',
+]
+
 
 @pytest.fixture(scope='session')
 def script_directory() -> Path:
@@ -380,6 +394,11 @@ def pytest_runtest_call(item):
     # This xfails so that you know when it starts to pass
     for marker in item.iter_markers(name="notimpl"):
         if backend in marker.args[0]:
+            if (
+                item.location[0] in FIlES_WITH_STRICT_EXCEPTION_CHECK
+                and "raises" not in marker.kwargs.keys()
+            ):
+                raise ValueError("notimpl requires a raises")
             reason = marker.kwargs.get("reason")
             item.add_marker(
                 pytest.mark.xfail(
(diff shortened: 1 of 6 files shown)


cpcloud:
@krzysztof-kwitt Thanks for bringing this up. I think it makes sense and adds value. It's already become much more clear why things are `xfail`-ing in our test suite thanks to your recent work.

cpcloud:
@kszucs Thoughts on this?

cpcloud:
@kszucs Ping!

krzysztof-kwitt:
@kszucs thanks to this, we can detect if an operation is not implemented or if an operation is incorrectly implemented, which improves the quality of our [operation support matrix](https://ibis-project.org/backends/support_matrix/).

Here example:
https://github.com/ibis-project/ibis/blob/cc86d4a4982a5068cd52762bfad8b2758067ab16/ibis/backends/tests/test_temporal.py#L1198-L1213

kszucs:
@krzysztof-kwitt sorry for the late response!

I'm definitely in favour of your proposal! I slightly dislike the test file listing in conftest, but I understand its advantage, at least until all of the tests get ported to the new approach. 

Also I'd like to reduce the possible types of raised exceptions, one way is to have more specific pytest markers which expect a concrete exception type, like:
- `com.OperationNotDefinedError` indicates that the `ibis` backend implementations hasn't defined a translation rule for that operation yet, though it's not clear whether is it possible in the first place and whether we plan to add support for it or not
- `sa.exc.ProgrammingError` indicates that the backend's sqlalchemy plugin is unable to compile the expression, meaning that ibis backend itself support the translation of the operation (the operation is defined in contrary of the previous error), but it's again not clear whether our implementation is wrong or the sqlalchemy plugin has bugs 
- `com.UnsupportedArgumentError` indicates that the backend's ibis implementation doesn't support an argument of that operation, but it's unclear whether it could support it at all or we just haven't implemented it yet
- there are more exception types...

It would be nice to encode the possible cases and have separate, case specific pytest markers. Since we currently handle too many exceptions, the backends should catch those and reduce the possible exception types to 5 or so. 

My proposal could be a follow-up of your work. What do you think?

krzysztof-kwitt:
If we already have tests that also cover exceptions, it will be easier for us to standardize and introduce further improvements.  

Druid tests passed now,  so It is ready for review.

krzysztof-kwitt:
If we already have tests that also cover exceptions, it will be easier for us to standardize and introduce further improvements.  

Druid tests passed now,  so It is ready for review.

Cloud tests also passed.
```
$ git diff HEAD..upstream/master --name-only | grep -v conf | xargs pytest -m 'bigquery or snowflake' --numprocesses auto --dist=loadgroup
========================================================================================== test session starts ===========================================================================================
platform darwin -- Python 3.10.7, pytest-7.2.1, pluggy-1.0.0
Using --randomly-seed=93531508
rootdir: /ibis/ibis, configfile: pyproject.toml
plugins: snapshot-0.9.0, xdist-3.1.0, mock-3.10.0, clarity-1.0.1, randomly-3.12.0, profiling-1.7.0, cov-4.0.0, repeat-0.9.1, benchmark-4.0.0, hypothesis-6.65.2
gw0 [456] / gw1 [456] / gw2 [456] / gw3 [456]
.........................x..........x.........x.............................x.................xx.x...................xx...x..x............x...x....x...x..................................x..x.... [ 42%]
...................x..x....x....x.........................................x.......................................................x........................................................x..x... [ 85%]
..x.x...x..x..xx..x.x....x.......x................x...x..xxx........                                                                                                                               [100%]
============================================================================== 416 passed, 40 xfailed in 203.36s (0:03:23) ===============================================================================
```

cpcloud (approved):
LGTM, nice work! Thanks for doing this.

PR #5550, opened 2023-02-15
feat(duckdb): use `read_json_auto` when reading json

cpcloud:
This PR changes the duckdb read_json implementation to use the new-in-duckdb-0.7.0 function `read_json_auto`. Depends on #5545.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -12,8 +12,14 @@ from typing import TYPE_CHECKING, Any, Iterable, Iterator, Mapping, MutableMappi
 import sqlalchemy as sa
 import toolz
 
+import ibis.common.exceptions as exc
 import ibis.expr.datatypes as dt
+import ibis.expr.schema as sch
+import ibis.expr.types as ir
 from ibis import util
+from ibis.backends.base.sql.alchemy import BaseAlchemyBackend
+from ibis.backends.duckdb.compiler import DuckDBSQLCompiler
+from ibis.backends.duckdb.datatypes import parse
 
 if TYPE_CHECKING:
     import duckdb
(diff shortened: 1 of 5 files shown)


ibis/backends/duckdb/__init__.py
```diff
@@ -232,14 +233,16 @@ def read_json(
     ) -> ir.Table:
         """Read newline-delimited JSON into an ibis table.
 
+        !!! note "This feature requires duckdb>=0.7.0"
```
jcrist, line 236: Is there a way we could check that this is true on the connection, and raise a better error message if the `duckdb` version is too low? If this is hard to do don't worry about it.

ibis/expr/api.py
```diff
@@ -880,7 +880,27 @@ def read_json(sources: str | Path | Sequence[str | Path], **kwargs: Any) -> ir.T
...
+    >>> import json
+    >>> from pathlib import Path
+    >>> import ibis
+    >>> ibis.options.interactive = True
+    >>> rows = [
+    ...    {"a": 1, "b": "d"},
+    ...    {"a": 2, "b": None},
+    ...    {"a": None, "b": "f"},
+    ... ]
+    >>> Path("lines.json").write_text("\n".join(map(json.dumps, rows)))
```
jcrist, line 892: Simplifying the example to reduce lines before the user sees Ibis code:

```suggestion
    >>> import ibis
    >>> ibis.options.interactive = True
    >>> jsonlines = """
    ... {"a": 1, "b": "d"}
    ... {"a": 2, "b": null}
    ... {"a": null, "b": "f"}
    ... """
    >>> open("lines.json", "w").write(jsonlines)
```

ibis/backends/duckdb/__init__.py
cpcloud: Yep, lemme give it a shot.

cpcloud: Fixed. Let me know if the error message is sufficient.

jcrist (approved):

PR #5551, opened 2023-02-15
refactor(sqlalchemy): set session parameters once per connection

cpcloud:
This PR sets ibis-required parameters once per connection instead of once per transaction. This should cut down the number of database calls and simplifies the code a bit.


jcrist (approved):
:shipit:

PR #5552, opened 2023-02-15
fix(sql): ensure that expressions are traversed in dependency order

### LINKED ISSUE #5215: bug: CTEs for SQL backends generate invalid forward references
### What happened?

# Summary
When constructing a complex transformation graph for use with a SQL backend, the non-recursive CTEs extracted from the graph can contain forward references to CTEs defined later in the query.  Per the ANSI SQL standard, forward references for non-recursive CTEs are not standards-compliant, so the generated SQL may fail to execute on standards-compliant backends like BigQuery.

# Reproduction
This Ibis code is as simple as I could make it while still exhibiting the issue, derived from a valid real-world use case.  Depending on the graph construction, the forward-referenced CTE can appear in different places.
```python
import ibis


def _select_lowest(idf_table, key_columns):
    idf_keys = (
        idf_table.group_by(key_columns)
        .order_by(['value'])
        .mutate(__row_num__DROP=ibis.row_number())
    )
    idf_keys = idf_keys.filter(idf_keys['__row_num__DROP'] == 0)
    idf_return = idf_keys.join(
        idf_table,
        key_columns + ['value'],
        how='inner',
        suffixes=('', '__DROP'),
    )
    idf_return = idf_return[[c for c in idf_return.columns if not c.endswith('__DROP')]]
    return idf_return


def _table_schema():
    import ibis.expr.datatypes as dtypes
    return ibis.schema([
        ('key1', dtypes.int),
        ('key2', dtypes.int),
        ('value', dtypes.float),
    ])


def min_repro_bq():
    from ibis.backends import bigquery
    backend = bigquery.Backend()
    # conn = backend.connect(project_id='bigquery-public-data', dataset_id='bls')
    # mr0 = conn.table('wm')
    mr0 = ibis.table(_table_schema(), 'test_tbl')

    mr1 = mr0[mr0.key2 >= 50]
    mr2 = mr1['key1', 'key2', 'value']
    mr3 = mr2.distinct()

    mr4 = mr3['key1', 'key2']
    mr5 = mr1.left_join(mr4, mr1.key1 == mr4.key1, suffixes=('', '_y'))
    mr6 = _select_lowest(mr5, ['key1', 'key2'])
    mr7 = mr6['key1', 'key2', 'value']

    mr8 = mr3.union(mr7)

    final = mr8
    sql = backend.compile(final)
    print(final)
    print('\n\n\n\n\n')
    print(sql)
```

This generates the following SQL when run with Ibis 4.0.0.  Note the forward reference from `t0` to `t3`.
```sql
WITH t0 AS (
  SELECT `key1`, `key2`
  FROM t3
),
t1 AS (
  SELECT *
  FROM test_tbl
  WHERE `key2` >= 50
),
t2 AS (
  SELECT `key1`, `key2`, `value`
  FROM t1
),
t3 AS (
  SELECT DISTINCT `key1`, `key2`, `value`
  FROM t1
)
SELECT *
FROM t3
UNION ALL
SELECT `key1`, `key2`, `value`
FROM (
  SELECT `key1`, `key2`, `value`, `key1_y`, `key2_y`
  FROM (
    SELECT t6.`key1`, t6.`key2`, t6.`value`, t6.`key1_y`, t6.`key2_y`,
           t6.`__row_num__DROP`, t7.`key1_y` AS `key1_y__DROP`,
           t7.`key2_y` AS `key2_y__DROP`
    FROM (
      SELECT t8.*
      FROM (
        SELECT t1.`key1`, t1.`key2`, t1.`value`, t0.`key1` AS `key1_y`,
               t0.`key2` AS `key2_y`,
               (row_number() OVER (PARTITION BY `key1`, `key2` ORDER BY `value`) - 1) AS `tmp`
        FROM t1
          LEFT OUTER JOIN t0
            ON t1.`key1` = t0.`key1`
      ) t8
      WHERE t8.`__row_num__DROP` = 0
    ) t6
      INNER JOIN (
        SELECT t1.`key1`, t1.`key2`, t1.`value`, t0.`key1` AS `key1_y`,
               t0.`key2` AS `key2_y`
        FROM t1
          LEFT OUTER JOIN t0
            ON t1.`key1` = t0.`key1`
      ) t7
        ON (t6.`key1` = t7.`key1`) AND
           (t6.`key2` = t7.`key2`) AND
           (t6.`value` = t7.`value`)
  ) t5
) t4
```

Subquery extraction and alias assignment [happen in separate passes](https://github.com/ibis-project/ibis/blob/8d53d724275ebe4b2a0bb0bd7e2c2dfc399e049b/ibis/backends/base/sql/compiler/base.py#L96-L98), which I suspect is contributing to the lookahead issue.

Confirmed in in 3.0.0, 3.2.0, 4.0.0 (trunk)

BigQuery, but the subquery extraction and aliasing logic lives in the base SQL backend

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
This PR fixes our core traversal algorithm, which only works on trees and not
general DAGs. I take a "two-stacks" approach. This approach is simpler than some alternatives
at the cost of using more memory (a `deque`).

Fixes #5215.

I still need to fix the sqlalchemy tests.

### DIFF
--- a/ibis/backends/tests/snapshots/test_sql/test_cte_refs_in_topo_order/clickhouse/out.sql
+++ b/ibis/backends/tests/snapshots/test_sql/test_cte_refs_in_topo_order/clickhouse/out.sql
@@ -1,0 +1,49 @@
+SELECT
+  t4.key
+FROM (
+  SELECT
+    t1.key
+  FROM (
+    SELECT
+      *
+    FROM leaf AS t0
+    WHERE
+      1
+  ) AS t1
+  INNER JOIN (
+    SELECT
+      t1.key
+    FROM (
+      SELECT
+        *
+      FROM leaf AS t0
+      WHERE
+        1
+    ) AS t1
+  ) AS t2
+    ON t1.key = t2.key
+) AS t4
+INNER JOIN (
+  SELECT
+    t1.key
+  FROM (
+    SELECT
+      *
+    FROM leaf AS t0
+    WHERE
+      1
+  ) AS t1
+  INNER JOIN (
+    SELECT
+      t1.key
+    FROM (
+      SELECT
+        *
+      FROM leaf AS t0
+      WHERE
+        1
+    ) AS t1
+  ) AS t2
+    ON t1.key = t2.key
+) AS t5
+  ON t4.key = t5.key
--- a/ibis/backends/base/sql/alchemy/query_builder.py
+++ b/ibis/backends/base/sql/alchemy/query_builder.py
@@ -144,13 +144,10 @@ class _AlchemyTableSetFormatter(TableSetFormatter):
                 # Was put elsewhere, e.g. WITH block, we just need to grab
                 # its alias
                 alias = ctx.get_ref(op)
+                if not hasattr(alias, "name"):
+                    alias = ctx.get_ref(ref_op).alias(alias)
 
-                # hack
-                if isinstance(op, ops.SelfReference):
-                    table = ctx.get_ref(ref_op)
-                    self_ref = alias if hasattr(alias, "name") else table.alias(alias)
-                    ctx.set_ref(op, self_ref)
-                    return self_ref
+                ctx.top_context.set_ref(op, alias)
                 return alias
 
             alias = ctx.get_ref(op)
--- a/ibis/backends/base/sql/alchemy/translator.py
+++ b/ibis/backends/base/sql/alchemy/translator.py
@@ -35,6 +35,9 @@ class AlchemyContext(QueryContext):
             params=self.params,
         )
 
+    def _compile_subquery(self, op):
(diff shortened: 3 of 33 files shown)


ibis/backends/tests/snapshots/test_sql/test_cte_refs_in_topo_order/clickhouse/out.sql
```diff
@@ -0,0 +1,49 @@
+SELECT
```
cpcloud, line 1: This isn't optimized, and we're not yet extracting subqueries in the clickhouse backend yet after the sqlglot refactor.

cpcloud:
This is uncovering some other annoying bugs, I'll put up a new PR when I've fixed them.

PR #5554, opened 2023-02-15
refactor(ir): update `ArrayMap` to use the new `callable_with` validation rule

kszucs:
depends on https://github.com/ibis-project/ibis/pull/5515

### DIFF
--- a/ibis/expr/operations/arrays.py
+++ b/ibis/expr/operations/arrays.py
@@ -1,8 +1,5 @@
 from __future__ import annotations
 
-import inspect
-from typing import Callable
-
 from public import public
 
 import ibis.common.exceptions as com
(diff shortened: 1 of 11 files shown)


ibis/expr/operations/arrays.py
```diff
@@ -94,30 +91,28 @@ class ArrayRepeat(Value):
...
-        ]
-        expr = self.func(*args)
-        return expr.op()
-
-    @property
-    def signature(self):
-        return list(inspect.signature(self.func).parameters.keys())
+        arg = Argument(
+            name=self.parameter,
+            shape=rlz.Shape.COLUMNAR,
```
cpcloud, line 105: This isn't correct. The output shape depends on the input shape. Consider the case of a literal array.

cpcloud (commented):
Looking pretty nice!

ibis/expr/operations/arrays.py
kszucs: Updating.

kszucs:
@cpcloud ready for another look

cpcloud (approved):

PR #5560, opened 2023-02-16
test(backend): add test for ops.Between

krzysztof-kwitt:
I noticed that we don't have any test for this operation.


cpcloud (approved):
LGTM, thanks!

cpcloud:
Looks like there's no implement for DataFusion, which isn't surprising.

With MSSQL it's the usual annoying "we don't allow boolean expressions in a `SELECT`" limitation.

Feel free to `notimpl`/`notyet` them.

PR #5562, opened 2023-02-16
fix(duckdb): add `null` to duckdb datatype parser

gforsyth:


jcrist (approved):
Did this come up from trying something?

gforsyth:
> Did this come up from trying something?

Debugging another issue, but I ended up with a dataframe with a column of `None`s and ran into this.

PR #5567, opened 2023-02-17
feat(common): add support for variadic positional and variadic keyword annotations

kszucs:
The main motivation for this change was to not limit `callable_with` support for functions with either variadic positional or variadic keyword arguments. `callable_with` converts the passed callable into an `@annotated` function which hasn't been supporting variadic arguments. 

To be precise, the annotation system used to support variadic arguments but we removed it due to implementation complications around node traversal (e.g. `NodeList`) and confusion when reconstructing annotable nodes. This time I managed to come up with a simpler and more complete implementation supporting variadic arguments for both function annotations via `@annotated`, variadic arguments for `Annotable` using `rlz.varargs()` and `rlz.varkwds()`.

### DIFF
--- a/ibis/common/annotations.py
+++ b/ibis/common/annotations.py
@@ -103,7 +103,7 @@ class Argument(Annotation):
         return cls(validator, kind=VAR_POSITIONAL)
 
     @classmethod
-    def varkwds(cls, validator=None):
+    def varkwargs(cls, validator=None):
         validator = None if validator is None else frozendict_of(any_, validator)
         return cls(validator, kind=VAR_KEYWORD)
 
@@ -240,7 +240,7 @@ class Signature(inspect.Signature):
             if param.kind is VAR_POSITIONAL:
                 annot = Argument.varargs(validator)
             elif param.kind is VAR_KEYWORD:
-                annot = Argument.varkwds(validator)
+                annot = Argument.varkwargs(validator)
             elif param.default is EMPTY:
                 annot = Argument.required(validator, kind=param.kind)
             else:
@@ -350,13 +350,13 @@ class Signature(inspect.Signature):
 
 
 # aliases for convenience
-attribute = Attribute
 argument = Argument
-required = Argument.required
-optional = Argument.optional
-varargs = Argument.varargs
-varkwds = Argument.varkwds
+attribute = Attribute
 default = Argument.default
+optional = Argument.optional
+required = Argument.required
+varargs = Argument.varargs
+varkwargs = Argument.varkwargs
 
 
 # TODO(kszucs): try to cache validator objects
--- a/ibis/common/annotations.py
+++ b/ibis/common/annotations.py
@@ -4,7 +4,7 @@ import functools
 import inspect
 from typing import Any
 
-from ibis.common.validators import Validator, any_, option
+from ibis.common.validators import Validator, any_, frozendict_of, option, tuple_of
 from ibis.util import DotDict
 
 EMPTY = inspect.Parameter.empty  # marker for missing argument
@@ -70,8 +70,15 @@ class Attribute(Annotation):
 class Argument(Annotation):
     """Base class for all fields which should be passed as arguments."""
 
+    __slots__ = ('_kind',)
+
+    def __init__(self, validator=None, default=EMPTY, kind=POSITIONAL_OR_KEYWORD):
+        self._kind = kind
+        self._default = default
+        self._validator = validator
+
     @classmethod
-    def mandatory(cls, validator=None):
+    def required(cls, validator=None):
         """Annotation to mark a mandatory argument."""
         return cls(validator)
 
@@ -89,6 +96,17 @@ class Argument(Annotation):
             validator = option(validator, default=default)
         return cls(validator, default=None)
 
+    @classmethod
+    def varargs(cls, validator=None):
+        """Annotation to mark a variable length positional argument."""
+        validator = None if validator is None else tuple_of(validator)
+        return cls(validator, kind=VAR_POSITIONAL)
+
+    @classmethod
+    def varkwds(cls, validator=None):
+        validator = None if validator is None else frozendict_of(any_, validator)
+        return cls(validator, kind=VAR_KEYWORD)
+
 
 class Parameter(inspect.Parameter):
     """Augmented Parameter class to additionally hold a validator object."""
@@ -102,7 +120,7 @@ class Parameter(inspect.Parameter):
             )
         super().__init__(
             name,
-            kind=POSITIONAL_OR_KEYWORD,
+            kind=annotation._kind,
             default=annotation._default,
             annotation=annotation._validator,
         )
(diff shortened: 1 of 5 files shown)


ibis/common/annotations.py
```diff
@@ -384,9 +435,10 @@ def annotated(_1=None, _2=None, _3=None, **kwargs):
 
     @functools.wraps(func)
     def wrapped(*args, **kwargs):
-        kwargs = sig.validate(*args, **kwargs)
-        result = sig.validate_return(func(**kwargs))
-        return result
+        values = sig.validate(*args, **kwargs)
```
kszucs, line 438: Add comments to describe the steps:

1. Argument validation
2. Call reconstruction
3. Actual function execution
4. Return value validation

ibis/common/grounds.py
```diff
@@ -96,6 +96,12 @@ def __create__(cls, *args, **kwargs) -> Annotable:
         kwargs = cls.__signature__.validate(*args, **kwargs)
         return super().__create__(**kwargs)
 
+    @classmethod
+    def __recreate__(cls, kwargs) -> Annotable:
```
kszucs, line 100: I couldn't come up with a better name, it's primarily used in `__reduce__`.

It is basically a variant of `__create__` without the flexible signature handling. For a signature like `Signature(a, b, c, *args, **kwargs)` the following two calls create the same object:

```py
from ibis.common.validators import instance_of
from ibis.common.annotations import varargs, varkwds
from ibis.common.grounds import Annotable

class MyClass(Annotable):
    a = instance_of(int)
    b = instance_of(int)
    args = varargs(instance_of(int))
    kwargs = varkwds(instance_of(int))

MyClass.__create__(1, 2, 3, 4, e=5, f=6)
MyClass.__recreate__(a=1, b=1, args=(3, 4), kwargs={'e': 5, 'f': 6})
```

I was thinking of calling it `def new(cls, kwargs)` to better align with `copy()` which also only supports direct field/attribute passing using kwargs.

ibis/common/annotations.py
```diff
@@ -278,7 +318,16 @@ def validate(self, *args, **kwargs):
             param = self.parameters[name]
             # TODO(kszucs): provide more error context on failure
             this[name] = param.validate(value, this=this)
+        return this
 
+    def validate_nobind(self, **kwargs):
```
kszucs, line 323: Direct parameter/argument validation without calling `Signature.bind()`

ibis/common/grounds.py
kszucs:
```py
MyClass(1, 2, 3, 4, e=5, f=6)
MyClass.new(a=1, b=1, args=(3, 4), kwargs={'e': 5, 'f': 6})
```

The latter should be more performant since it bypasses https://github.com/python/cpython/blob/3.11/Lib/inspect.py#L3075

ibis/common/annotations.py
```diff
@@ -89,6 +96,17 @@ def optional(cls, validator=None, default=None):
             validator = option(validator, default=default)
         return cls(validator, default=None)
 
+    @classmethod
+    def varargs(cls, validator=None):
```
kszucs, line 100: `variadic` could be a slightly better naming choice

ibis/common/annotations.py
```diff
@@ -89,6 +96,17 @@ def optional(cls, validator=None, default=None):
...
+        """Annotation to mark a variable length positional argument."""
+        validator = None if validator is None else tuple_of(validator)
+        return cls(validator, kind=VAR_POSITIONAL)
+
+    @classmethod
+    def varkwds(cls, validator=None):
```
kszucs, line 106: consider spelling it out as `varkeywords`

ibis/common/annotations.py
```diff
@@ -150,22 +168,34 @@ def merge(cls, *signatures, **annotations):
...
-                    inherited_kwargs.append(param)
+                    old_kwargs.append(param)
             else:
                 if param.default is EMPTY:
                     new_args.append(param)
                 else:
                     new_kwargs.append(param)
 
-        return cls(inherited_args + new_args + new_kwargs + inherited_kwargs)
+        if len(var_args) > 1:
```
kszucs, line 191: These could be moved to the specific branches above.

ibis/common/annotations.py
```diff
@@ -199,25 +229,24 @@ def from_callable(cls, fn, validators=None, return_validator=None):
 
         parameters = []
         for param in sig.parameters.values():
-            if param.kind in {
-                VAR_POSITIONAL,
-                VAR_KEYWORD,
-                POSITIONAL_ONLY,
-                KEYWORD_ONLY,
-            }:
+            if param.kind in {POSITIONAL_ONLY, KEYWORD_ONLY}:
```
kszucs, line 232: Should be possible to support both parameter kinds with slight modifications in `unbind()`

ibis/common/annotations.py
kszucs: Done.

ibis/common/annotations.py
kszucs: Added support for both of them.

ibis/common/annotations.py
kszucs: Done.

ibis/common/annotations.py
cpcloud: How about `varkwargs` for consistency with the usual spelling of `*args, **kwargs`?

ibis/common/annotations.py
cpcloud: I think `varargs` is fine, if only for consistency with `*args, **kwargs`.

ibis/common/grounds.py
cpcloud: How about `__strict_create__`?

cpcloud: Or even `__create_impl__` might be okay.

ibis/common/tests/test_annotations.py
```diff
@@ -123,31 +123,62 @@ def test(a: int, b: int, c: int = 1):
...
 
-def test_signature_from_callable_unsupported_argument_kinds():
-    def test(a: int, b: int, *args):
-        pass
 
-    with pytest.raises(TypeError, match="unsupported parameter kind VAR_POSITIONAL"):
-        Signature.from_callable(test)
+def test_signature_from_callable_with_varargs():
+    def test(a: int, b: int, *args: int):
+        return a + b + sum(args)
```
cpcloud, line 133: Can you put `...` to indicate that the body of the function doesn't matter, and that what's being tested is the signature? Same for the below tests.

cpcloud (commented):
Looking good. A couple of comments.

ibis/common/tests/test_annotations.py
kszucs: Sure, updating.

ibis/common/annotations.py
kszucs: Sounds good, updating.

ibis/common/annotations.py
kszucs: Ok, keeping as is.

ibis/common/grounds.py
kszucs: This is the API we need to use in the tree substitute/replace functions, so perhaps this should be public. 

Another approach could be the following:

```py
class Annotable:
    ...

    @classmethod
    def from_dict(cls, dct):
        # bypass signature binding by requiring keyword arguments only
        kwargs = cls.__signature__.validate_nobind(**kwargs)

    def to_dict(self):
        return {k: getattr(self, k) for k in self.__attributes__}
```

kszucs: Though `from_dict` and `to_dict` wouldn't be symmetric in the example above, since the latter would return with the generated state/attributes as well while the former would accept only the argument kind of attributes.

cpcloud: We can leave it private for now IMO

cpcloud (approved):

PR #5575, opened 2023-02-20
refactor(analysis): slightly simplify `find_subqueries()`

kszucs:
depends on #5571

### DIFF
--- a/ibis/backends/base/sql/compiler/base.py
+++ b/ibis/backends/base/sql/compiler/base.py
@@ -68,8 +68,8 @@ class SetOp(DML):
         return map(self.keyword, self.distincts)
 
     def _extract_subqueries(self):
-        self.subqueries = _extract_common_table_expressions(
-            [self.table_set, *self.filters]
+        self.subqueries = an.find_subqueries(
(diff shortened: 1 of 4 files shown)


ibis/backends/base/sql/compiler/base.py
```diff
@@ -116,7 +116,5 @@ def compile(self):
 def _extract_common_table_expressions(nodes):
     # filter out None values
     nodes = list(filter(None, nodes))
-    counts = an.find_subqueries(nodes)
-    duplicates = [op for op, count in counts.items() if count > 1]
-    duplicates.reverse()
-    return duplicates
+    subqueries = an.find_subqueries(nodes)
+    return tuple(reversed(subqueries))
```
cpcloud, line 120: Not a blocker, but it seems a bit wasteful to return a `tuple` from `find_subqueries`, reverse it, and then turn that into a tuple, when we can return a `list` and reverse it in place.

cpcloud (approved):

cpcloud (approved):
Sweet, :shipit:

PR #5578, opened 2023-02-20
fix(sqlalchemy): ensure that isin contains full column expression

### LINKED ISSUE #5553: bug: .isin(Expr) returns wrong results
### What happened?

Not sure exactly what conditions cause it to not work, but here is one example:

```python
import pandas as pd
import ibis

df = pd.DataFrame(
    {
        "x": [1, 2, 3, 4, 5],
    }
)
t = ibis.memtable(df)
good = t[t.x > 2].x
t.x.isin(good)
```
results in
```
┏━━━━━━━━━━━━━━━━┓
┃ Contains(x, x) ┃
┡━━━━━━━━━━━━━━━━┩
│ boolean        │
├────────────────┤
│ True           │
│ True           │
│ True           │
│ True           │
│ True           │
└────────────────┘
```
when the first two should be False

master

duckdb

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
This PR fixes an issue where we failed to compile the entire right-hand-side
expression of a `Contains` operation when the operation was a non-trivial
query.

Fixes #5553.

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -189,14 +189,15 @@ def _contains(func):
     def translate(t, op):
         left = t.translate(op.value)
 
-        if isinstance(op.options, tuple):
+        options = op.options
+        if isinstance(options, tuple):
             right = [t.translate(x) for x in op.options]
-        elif op.options.output_shape.is_columnar():
-            right = t.translate(op.options)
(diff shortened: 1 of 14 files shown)


kszucs (approved):

kszucs:
Going to rebase on top of https://github.com/ibis-project/ibis/pull/5571 before merging to ensure that the generated queries don't differ.

PR #5582, opened 2023-02-20
fix(snowflake): add snapshot for `test_many_subqueries()` test case

kszucs:
was missing from https://github.com/ibis-project/ibis/pull/5571


cpcloud:
I added this in #5578 :)

kszucs:
Ohh, cool.

PR #5585, opened 2023-02-21
fix(ux): make top-level set ops work in interactive mode

cpcloud:
Make top-level set operations work.


ibis/backends/tests/test_set_ops.py
```diff
@@ -139,6 +140,52 @@ def test_difference(backend, alltypes, df, distinct):
...
+@pytest.mark.parametrize("distinct", [True, False])
+@pytest.mark.notimpl(["dask", "pandas"], raises=com.UnboundExpressionError)
+@pytest.mark.notyet(
+    ["clickhouse"],
+    raises=sg.errors.ParseError,
+    reason="sqlglot fails to parse UNION ALL/DISTINCT",
+)
+def test_top_level_union(con, distinct, monkeypatch):
+    monkeypatch.setattr(ibis.options, "default_backend", con)
+    monkeypatch.setattr(ibis.options, "interactive", True)
```
kszucs, line 157: Why is this new testing approach?

cpcloud: What do you mean?

ibis/backends/tests/test_set_ops.py
```diff
@@ -139,6 +140,52 @@ def test_difference(backend, alltypes, df, distinct):
...
+
+    t1 = ibis.memtable({"a": [1, 2, 3]}, name="t1")
+    t2 = ibis.memtable({"a": [3, 2, 1]}, name="t2")
+    expr = t1.union(t2, distinct=distinct)
+    repr(expr)
```
kszucs, line 162: Do you plan to add assertion here?

cpcloud: No, it's enough that it doesn't raise.

ibis/backends/tests/test_set_ops.py
cpcloud: I'm setting the default backend so that I can transparently test using `memtable`. I'm setting interactive mode since that is require to reproduce the failure.

kszucs: I can't see any issue referenced, based on the select builder changes it looks like a new feature, that's why I find the testing weird.

cpcloud: I think I'm still not following :)

How is whether the thing is a new feature versus a fix related to how it's being tested?

kszucs: I'd expect to test like we usually do, execute the expression and compare it with the expected pandas result.

kszucs (approved):
LGTM

ibis/backends/tests/test_set_ops.py
cpcloud: Fixed.

ibis/backends/tests/test_set_ops.py
cpcloud: Fixed.

Issue #5546, opened 2023-02-14
feat: Allow set operations for schemas


NickCrews:
I've wanted this several times before. One common place would be renaming columns before a join that would result in colliding column names.

Another is to find a common schema before union()ing two tables

This would really set in stone the idea that columns are unordered. IDK what other repercussions this would have.

```python
right_but_not_left = right.schema() - left.schema() 
union = left.schema() | right.schema()
intersection = left.schema() & right.schema()
```

_Originally posted by @NickCrews in https://github.com/ibis-project/ibis/issues/5537#issuecomment-1428873736_

cpcloud:
I personally am not opposed to formalizing the unordered nature of schemas, but there will probably be a bunch of follow up fixes and changes where there's an implicit assumption of ordering inside of ibis.

It's hard to tell exactly if/where such code exists right now.

jreback:
umm this would be a major major change to allow unordered schemas 

this is a staple of all major dataframe systems

would be highly susoect if this is event possible

cpcloud:
> umm this would be a major major change to allow unordered schemas

It's not clear what you mean by "allow unordered schemas". Can you elaborate?

Ibis already treats schemas as equal if their field names and types are equal. This has been true for many releases.

> 
> this is a staple of all major dataframe systems
> 
> would be highly susoect if this is event possible

If what is possible?

Right now there are no explicit allowances for schemas that aren't in the same order except for allowing the union of two table expressions that have equal schemas with fields in a different order which is a recent addition, and had no effects other than to allow that use case to work.

NickCrews:
I think we could also bastardize the set operations and keep them ordered:
(this doesn't run but you get the idea)
- a - b == [c for c in a.schema() if c not in b.schema()]
- a & b == [c for c in a.schema() if c in b.schema()]
- a | b == deduped(list(a.schema()) + list(b.schema()))

this would work for my use cases above.

jcrist:
I agree with that adding set operations on a schema doesn't actually require any changes in our schema ordering assumptions.

However, it's not clear to me what the expected behavior of set operations on schemas should be? Do we consider name only, or name & type? For example:

```python
In [1]: import ibis

In [2]: s1 = ibis.schema({"x": "int", "y": "string"})

In [3]: s2 = ibis.schema({"x": "string", "z": "string"})

In [4]: s1 & s2  # what happens here?
```

[...]

PR #5588, opened 2023-02-21
feat(common): add abstract mapping collection with support for set operations

kszucs:
TODOs:
- [x] need to add tests for schema
- [x] need to add tests for StructType

Resolves https://github.com/ibis-project/ibis/issues/5546

### DIFF
--- a/ibis/common/tests/test_collections.py
+++ b/ibis/common/tests/test_collections.py
@@ -1,0 +1,130 @@
+from collections.abc import ItemsView, Iterator, KeysView, ValuesView
+
+import pytest
+
+from ibis.common.collections import MapSet
+
+
+class MySchema(MapSet):
+    def __init__(self, dct=None, **kwargs):
+        self._fields = dict(dct or kwargs)
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}({self._fields})'
+
+    def __getitem__(self, key):
+        return self._fields[key]
+
+    def __iter__(self):
+        return iter(self._fields)
+
+    def __len__(self):
+        return len(self._fields)
+
+    def identical(self, other):
+        return type(self) == type(other) and tuple(self.items()) == tuple(other.items())
+
+
+def test_myschema_identical():
+    ms1 = MySchema(a=1, b=2)
+    ms2 = MySchema(a=1, b=2)
+    ms3 = MySchema(b=2, a=1)
+    ms4 = MySchema(a=1, b=2, c=3)
+    ms5 = {}
+
+    assert ms1.identical(ms2)
+    assert not ms1.identical(ms3)
+    assert not ms1.identical(ms4)
+    assert not ms1.identical(ms5)
+
+
+def test_mapset_mapping_api():
+    ms = MySchema(a=1, b=2)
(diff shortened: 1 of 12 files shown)


ibis/common/tests/test_collections.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # __xor__
+    with pytest.raises(ValueError):
+        # duplicate keys with different values
+        a ^ a_
+
+    assert a ^ b == MySchema(c=3)
+    assert b ^ a == MySchema(c=3)
+    assert a ^ f == MySchema(a=1, b=2, d=4, e=5)
+    assert f ^ a == MySchema(a=1, b=2, d=4, e=5)
```
cpcloud, line 119: Why wouldn't `f ^ a` return `MySchema(d=4, e=5, a=1, b=2)`? I would expect the order to be preserved in binary operations.

kszucs: Need better assertions since `__eq__` doesn't check for order.

ibis/expr/tests/test_rules.py
```diff
@@ -332,7 +332,7 @@ def test_table_with_schema(table):
 )
 def test_table_with_schema_invalid(table):
     validator = rlz.table(schema=[('group', dt.double), ('value', dt.timestamp)])
-    with pytest.raises(IbisTypeError):
+    with pytest.raises(ValueError):
```
kszucs, line 335: We need to clean up the exceptions in the rules.

ibis/common/tests/test_collections.py
kszucs: Fixed.

ibis/common/tests/test_collections.py
```diff
@@ -0,0 +1,130 @@
...
+    assert b >= a
+    assert a >= a
+    assert not a >= b
+    assert not a > b
+    assert not a_ > a
+    with pytest.raises(ValueError, match="Conflicting values"):
+        a_ >= a  # noqa: B015
+
+    # __and__
+    with pytest.raises(ValueError, match="Conflicting values"):
```
kszucs, line 101: The semantics are not straightforward in all cases, but I chose to always raise on conflicting values for consistency.

cc @NickCrews 

The

kszucs:
@cpcloud shall I squash?

cpcloud:
@kszucs SGTM!

@NickCrews Would you mind giving this a go against some of your use cases and reporting back?

ibis/expr/schema.py
```diff
@@ -182,11 +180,11 @@ def merge(self, other: Schema) -> Schema:
...
         ibis.Schema {
-          a  int64
-          b  string
-          c  float64
-          d  int16
-        }
+        a  int64
+        b  string
+        c  float64
+        d  int16
```
cpcloud, line 186: Can you reindent this back to the old one?

kszucs: Updated.

kszucs:
@cpcloud can we go ahead and merge it? I have a follow-up #5622

NickCrews:
Sorry this is so slow, but this looks great!

PR #5589, opened 2023-02-21
feat: `ibis.examples`

### LINKED ISSUE #5404: docs: add `ibis.example()` for an example dataset

Many of the [docstrings have examples](https://ibis-project.org/docs/4.1.0/api/expressions/tables/#ibis.expr.types.groupby.GroupedTable.mutate) that only use expressions that are unbound to data. It is not obvious to me what that groupby().mutate() funciton does from that. It would be nice if there was concrete data that we could see.

This would also be nice on the walkthroughs, eg in https://ibis-project.org/docs/4.1.0/ibis-for-pandas-users/ we download the iris dataset from github in the middle of the walkthrough, it would be nice if this worked offline.

[Vaex has `vaex.example()`](https://vaex.readthedocs.io/en/latest/index.html#Getting-started) that just gives you a dataset packaged with it, ready to go.

[sklearn does somethign similar](https://scikit-learn.org/stable/datasets/toy_dataset.html)

I think I might actually prefer sklearn's approach because it is better namespaced, and more extendable in the future. For example depending on the example you might want different datasets. Some datasets might be only ~100 rows and could be committed to git, so they are always available, while other larger ones could be fetched remotely.

I can write a PR if you give some guidelines.

target 5.0?

all

### Code of Conduct

cpcloud:
This PR adds an `ibis.examples` namespace and corresponding API. Keeping draft mode for now so people can experiment with it before I go further.

Closes #5404.


cpcloud:
If we don't want to keep any files in the repo (it's about 4MB now that they're all compressed) we can just keep the `fetch` API for now.

gforsyth:
Having easy-to-access examples is a great, great feature and we should ship this for 5.0.

Oooh, and we should use some of these in our unit tests since they'll have less predictable data and I'll bet we smoke out a bunch of bugs (not in this PR).

My thoughts on shipping vs. not-shipping the data in-repo:
- 4 mb isn't that big, so either way, it's fine
- It probably _isn't_ ok for us to bundle 4mb of compressed CSVs in an ibis wheel or tarball for our next release, so there will need to be a way to fetch these for end-users (which you've already added).
- Currently, though, there's no way for users to _know_ which datasets they can fetch

Given the above, I think we want a mini (and very dumb) catalog of available datasets, so users can do an `ibis.examples.fetch("penguins", "Nile")` to grab a few named datasets, and then also a convenience method to "download all example data".

And if we're going to have the convenience method, maybe we have it download into the `data/` directory and then `gitignore` it and that's easy access for developers.

In terms of catalog, we could use the description files as a listing of what's available (or we can smush them down into a CSV/json/whatever if we want a single file).

For now, maybe `fetch` always overwrites if the data are present, and later we can get fancy and try using something like https://github.com/fatiando/pooch, which @cpcloud just introduced me to.

cpcloud:
Alright, I think this is ready for review.

I ended up:

1. Using pooch. I don't want to deal with all the things pooch is doing for us now like caching and directory configuration.
2. Writing a couple scripts to generate the data from R -> rsync the data to the bucket -> generate the decriptions JSON -> regen the registry using pooch
3. Removing the `load` method, now there's only `fetch`. `fetch` will download the data if it doesn't exist and otherwise use the pooch cache.

I'm pretty happy with the results.

We now have `ibis.examples` that:

1. Does minimal computation until you ask for a table
2. Is fast to tab complete

Here's some example usage:

```

In [2]: t = ex.diamonds.fetch()

In [3]: t.group_by("cut").depth.mean()
Out[3]:
┏━━━━━━━━━━━┳━━━━━━━━━━━━━┓
┃ cut       ┃ mean(depth) ┃
┡━━━━━━━━━━━╇━━━━━━━━━━━━━┩
│ string    │ float64     │
├───────────┼─────────────┤
│ Ideal     │   61.709401 │
│ Premium   │   61.264673 │
│ Good      │   62.365879 │
│ Very Good │   61.818275 │
│ Fair      │   64.041677 │
└───────────┴─────────────┘

In [4]: t = ex.agriculture.fetch()
Downloading file 'agriculture.csv.gz' from 'https://storage.googleapis.com/ibis-examples/data/agriculture.csv.gz' to '/home/cloud/.cache/ibis-framework/4.1.0'.

In [5]: t.head()
Out[5]:
┏━━━━━━━━━┳━━━━━━━━━┓
┃ x       ┃ y       ┃
┡━━━━━━━━━╇━━━━━━━━━┩
│ float64 │ float64 │
├─────────┼─────────┤
│    16.8 │     2.7 │
│    21.3 │     5.7 │
│    18.7 │     3.5 │
│     5.9 │    22.2 │
│    11.4 │    10.9 │
└─────────┴─────────┘

In [6]: t = ex.agriculture.fetch()

In [7]: t.head()
Out[7]:
┏━━━━━━━━━┳━━━━━━━━━┓
┃ x       ┃ y       ┃
┡━━━━━━━━━╇━━━━━━━━━┩
│ float64 │ float64 │
├─────────┼─────────┤
│    16.8 │     2.7 │
│    21.3 │     5.7 │
│    18.7 │     3.5 │
│     5.9 │    22.2 │
│    11.4 │    10.9 │
└─────────┴─────────┘
```

gforsyth:
Hmm, weirdly, the failing windows jobs still have the actions in the old (breaking) order:

cpcloud:
Ah, because I only adjusted the backend CI suite! Nice catch.

PR #5591, opened 2023-02-21
fix(sqlalchemy): reflect most recent schema when view is replaced

### LINKED ISSUE #5563: bug: schema of replaced view not refreshed on registration
### What happened?

```python
[ins] In [1]: import ibis

[ins] In [2]: import pandas as pd

[ins] In [3]: df = pd.DataFrame([["a", 1], ["b", 2]], columns=["let", "num"])

[ins] In [4]: con = ibis.duckdb.connect()

[ins] In [5]: con.register(df, table_name="t")
Out[5]: 
AlchemyTable: t
  let string
  num int64

[ins] In [6]: df = pd.DataFrame([["a", "a"], ["b", "b"]], columns=["let", "num"])

[ins] In [7]: con.register(df, table_name="t")
Out[7]: 
AlchemyTable: t
  let string
  num int64

[ins] In [8]: con.tables.t.execute()
Out[8]: 
  let num
0   a   a
1   b   b
```

When we re-register something with the same name, the view in DuckDB is updated, but not the Ibis schema.  For something like the `int->str` example above, the schema is wrong but the result is correct.
If you do `int->float` however, ibis helpfully casts back results as integers (:grimacing:)

Or, if you have extra columns...

```python
[ins] In [9]: df = pd.DataFrame([["a", "a", 2], ["b", "b", 3]], columns=["let", "let2", "num"])

[ins] In [10]: con.register(df, table_name="t")
Out[10]: 
AlchemyTable: t
  let string
  num int64
```

I'm nearly certain this can be fixed with `reflect_table`, i just didn't have time to work up the unit tests right now and didn't want to forget about this.

master

DuckDB

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
This may be too coarse of a method for this.

This is a fix for creating a table or view with the same name as an existing table or view.  This is most likely to occur when using the various `parquet` and `csv` registration features, because we do a lot of `CREATE OR REPLACE VIEW`.

Before this, the ibis table would reflect the schema of the initial registration, but would not update if you try to re-use the same view name for a different file (with a different schema).

Other fixes I tried:
`sa.table(..., extend_existing=True)`
This works if the subsequently registered schema is a superset of the existing schema, but if it's a subset, then you end up with a union of the two registration runs.

Calling `reflect_table`
This is actually already happening, but it refers to the contents of `self.meta` -- this oddly seems to update the _types_ of any columns if that is merited, but doesn't drop columns that are no longer present in the most recent view.

Manually popping out columns
It's called an ImmutableColumnCollection for a reason, I guess.

Fixes #5563

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -387,6 +387,10 @@ class BaseAlchemyBackend(BaseSQLBackend):
     def _get_sqla_table(
         self, name: str, schema: str | None = None, autoload: bool = True, **kwargs: Any
     ) -> sa.Table:
+        # If the underlying table (or more likely, view) has changed, remove it
+        # to ensure a correct reflection
+        if autoload and self.inspector.has_table(name):
+            self.meta.remove(sa.Table(name, self.meta))
         with warnings.catch_warnings():
             warnings.filterwarnings(
                 "ignore", message="Did not recognize type", category=sa.exc.SAWarning
--- a/ibis/backends/duckdb/tests/test_register.py
+++ b/ibis/backends/duckdb/tests/test_register.py
@@ -124,6 +124,21 @@ def test_read_in_memory():
     assert "df_pandas" in con.list_tables()
 
 
+def test_re_read_in_memory_overwrite():
+    con = ibis.duckdb.connect()
+
+    df_pandas_1 = pd.DataFrame({"a": ["a"], "b": [1], "d": ["hi"]})
+    df_pandas_2 = pd.DataFrame({"a": [1], "c": [1.4]})
+
+    table = con.read_in_memory(df_pandas_1, table_name="df")
+    assert len(table.columns) == 3
+    assert table.schema() == ibis.schema([("a", "str"), ("b", "int"), ("d", "str")])
+
+    table = con.read_in_memory(df_pandas_2, table_name="df")
+    assert len(table.columns) == 2
+    assert table.schema() == ibis.schema([("a", "int"), ("c", "float")])
+
+
 def test_memtable_with_nullable_dtypes():
     data = pd.DataFrame(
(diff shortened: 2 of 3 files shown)


ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -387,6 +387,10 @@ def _log(self, sql):
     def _get_sqla_table(
         self, name: str, schema: str | None = None, autoload: bool = True, **kwargs: Any
     ) -> sa.Table:
+        # If the underlying table (or more likely, view) has changed, remove it
+        # to ensure a correct reflection
+        if autoload and self.inspector.has_table(name):
+            self.meta.remove(sa.Table(name, self.meta))
```
cpcloud, line 393: Is the coarseness issue that it may be expensive to repeatedly call `con.table("t")`?

gforsyth: That was my thinking, that calling tables now involves this extra reflection.

We _could_ probably do a set comparison of the existing columns in `self.meta` and the columns in `self._metadata` but I don't know if that will be more expensive

cpcloud: Eh, I don't think it's worth it yet.

Some very light profiling against the snowflake backend (one of the slowest backends for metadata queries) shows about a 20% performance hit:

```
# `master`
In [10]: %timeit con.table("diamonds")
1.41 s ± 223 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# this PR
In [4]: %timeit con.table("diamonds")
1.75 s ± 229 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Since this is a correctness issue I think we should fix it even if there's a performance hit (for now).

cpcloud (approved):

PR #5596, opened 2023-02-22
feat(pandas): add unnest support

### LINKED ISSUE #5478: feat(pandas): unnest support

The pandas backend does not currently support unnest.

Would [pd.Series.explode](https://pandas.pydata.org/docs/reference/api/pandas.Series.explode.html#pandas.Series.explode) work here? It looks like the problem with such a solution is the cardinality change which means the original data index cannot be assigned here (assertion fails):
https://github.com/ibis-project/ibis/blob/0e8b06643bfe81fd8694c5578324f263e49b9890/ibis/backends/pandas/execution/selection.py#L261-L267
If the check fails, perhaps you could just skip the assignment of the original `data.index`? This seems to work for a very simple test case... Thanks for this project!

4.1.0

pandas

### Code of Conduct

dakoop:
~This will not pass the basic unnest test for reasons discussed in #5478~ (this now passes!), but I wanted to share the initial progress. For simpler cases, it worked for me. Also not sure if the changes to `build_df_from_projection` will cause problems.

Closes #5478.

### DIFF
--- a/ibis/backends/pandas/client.py
+++ b/ibis/backends/pandas/client.py
@@ -195,7 +195,7 @@ def convert_struct_to_dict(_, out_dtype, column):
 
 @sch.convert.register(np.dtype, dt.Array, pd.Series)
 def convert_array_to_series(in_dtype, out_dtype, column):
-    return column.map(lambda x: x if x is None else list(x))
+    return column.map(lambda x: list(x) if util.is_iterable(x) else x)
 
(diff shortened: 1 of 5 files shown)


cpcloud:
After a small bit of surgery, this passes the full backend test suite 🎉

cpcloud (approved):

PR #5598, opened 2023-02-22
feat(ux): `across`/`if_any`/`if_all` selectors for whole-table operations

### LINKED ISSUE #5422: feat: table-wide operations
It has come up a number of times now, both internally at Voltron Data and in
other contexts that the ability to apply an operation across a set of columns
in one call--without complex list comphrehensions--would be a useful API to
provide.

This issue is to discuss the problems this would solve and the solution space and various tradeoffs.

I think there are three primary use cases that subsume the others.

## Use case 1: Apply an expression to one or more columns in table

This is a fairly common thing to want to do.

#### Examples

1. Round every column
2. Append a string to every row in every string column
3. Apply a function based on a complex selector (e.g., numerics whose name begins with at least two `a` characters)

Replacing existing columns is a special case of this.

---

## Use case 2: Aggregate over all columns of a given type

Let's look at normalization:

```
t.min()
```

Here, `t` is a table, and `min`/`max` are methods that return a single
row table containing the `min`/`max` of all the table's columns.

### Pandas

Pandas has methods on `DataFrame` to express this computation succinctly:

```
df.min()
```

### R

Using the tidyverse:

```
r$> iris %>% as_tibble %>% summarize(across(where(is.numeric), min))
# A tibble: 1 × 4
  Sepal.Length Sepal.Width Petal.Length Petal.Width
         <dbl>       <dbl>        <dbl>       <dbl>
1          4.3           2            1         0.1
```

This is certainly more verbose, and we could make this more convenient in ibis for sure, but it composes
with column selection much much better than pandas, and also composes with grouping and windowing operations.

## Use case 3: Combining cases one and two

```
(t - t.mean()) / t.std()
```

Here, the semantics of this operation are:

1. compute the mean for each column and return a single row table
1. do the same for the standard deviation
1. for every row in `t` substract the columns from the mean
1. then divide column column in the row by the standard deviation

This expressed with more or less difficulty depending on the language used

### Pandas

Pandas makes this computation very concise:

```
(df - df.mean()) / df.std()
```

This is made easy by a very implementation detail of pandas, and that is that all data are in-memory.

### R

Using tidyverse libraries:

```
r$> iris %>% as_tibble %>% mutate(across(where(is.numeric), ~(.x - mean(.x, na.rm=T)) / sd(.x)))
# A tibble: 150 × 5
   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
          <dbl>       <dbl>        <dbl>       <dbl> <fct>
 1       -0.898      1.02          -1.34       -1.31 setosa
 2       -1.14      -0.132         -1.34       -1.31 setosa
 3       -1.38       0.327         -1.39       -1.31 setosa
 4       -1.50       0.0979        -1.28       -1.31 setosa
 5       -1.02       1.25          -1.34       -1.31 setosa
 6       -0.535      1.93          -1.17       -1.05 setosa
 7       -1.50       0.786         -1.34       -1.18 setosa
 8       -1.02       0.786         -1.28       -1.31 setosa
 9       -1.74      -0.361         -1.34       -1.31 setosa
10       -1.14       0.0979        -1.28       -1.44 setosa
# … with 140 more rows
# ℹ Use `print(n = ...)` to see more rows
```

What is really compelling about this interface is the ability to throw in a `group_by`, and the `mutate` expression is unchanged:

```
r$> iris %>% as_tibble %>% group_by(Species) %>% mutate(across(where(is.numeric), ~(.x - mean(.x, na.rm=T)) / sd(.x)))
# A tibble: 150 × 5
# Groups:   Species [3]
   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
          <dbl>       <dbl>        <dbl>       <dbl> <fct>
 1       0.267       0.190        -0.357      -0.436 setosa
 2      -0.301      -1.13         -0.357      -0.436 setosa
 3      -0.868      -0.601        -0.933      -0.436 setosa
 4      -1.15       -0.865         0.219      -0.436 setosa
 5      -0.0170      0.454        -0.357      -0.436 setosa
 6       1.12        1.25          1.37        1.46  setosa
 7      -1.15       -0.0739       -0.357       0.512 setosa
 8      -0.0170     -0.0739        0.219      -0.436 setosa
 9      -1.72       -1.39         -0.357      -0.436 setosa
10      -0.301      -0.865         0.219      -1.39  setosa
# … with 140 more rows
# ℹ Use `print(n = ...)` to see more rows
```

### SQL

I don't think there's a dialect of SQL that lets you spell things like "all columns", "only columns with property X", etc.

ClickHouse and DuckDB have `COLUMNS(*)` but they are pretty all or none, you either do something with all columns or columns
that match each other exactly, but AFAICT there's no way to say "combine these columns with these other distinct columns".

#### Try to make something like `(t - t.mean()) / t.std()` work

While tantalzing, this approach is IMO a bad idea:

1. It has a lot of easy-to-misunderstand semantics baked in. What does `table - other table` mean? What if one of them is a group by?
1. The codebase would gain more messy analysis code to detect specific cases that don't have well-defined semantics like the group by case
1. It doesn't compose in an obvious way: how can I as a user discover that I can use `-` with tables and that `t1 - t2` has meaning (and discover its meaning as well!).
1. It's not clear how far we'd want to take this.
1. It's not clear what SQL this would generate in non trivial cases (group by)

#### Expand selector functionality and make them work with `.mutate` and `.aggregation`

This is better approach than trying to force fit the pandas API into ibis. It has the benefit of hindsight and prior art
which we can look to for inspiration.

- The semantics are clear for `mutate`: for scalar functions call the function
  on every row, for aggregate functions use a window operator, partitioning if
  there's group by
- The semantics are very similar for `agg` (call `summari(s|z)e` in dplyr) as well and follow similar logic as `mutate` sans scalar and window functions.
- It composes well.
- It decouples the column selection, grouping and windowing from the
  computation. I.e., group_by + mutate indicates a partitioned window
  operation, and mean etc remain the same.
- We could potentially add some convenience methods on tables for extremely simple cases like `t.mean()` etc.

I realize that a lot of this is aesthetic preferences and depends on what perspective I am coming from, so I would like to discuss!

cpcloud:
This PR adds `dplyr`-style [`across`, `if_any` and `if_all` selectors](https://dplyr.tidyverse.org/reference/across.html).

The goal here is to add the ability to execute table-wide operations, where "table-wide" means some subset of the existing table's columns.

The model is very different from pandas. You can't simply subtract two tables like you can with pandas, because such an operation is not well defined
for two arbitrary relations.

However, it's well within the realm of ibis to make doing operations like normalization and scaling much much more concise.

Check out the blog entry for more details.

Closes #5422.

### DIFF
--- a/docs/blog/selectors.md
+++ b/docs/blog/selectors.md
@@ -1,0 +1,448 @@
+# Maximizing Productivity with Selectors
+
+Before Ibis 5.0 it's been challenging to concisely express whole-table
+operations with ibis. Happily this is no longer the case in ibis 5.0.
+
+Let's jump right in!
+
+We'll look at selectors examples using the [`palmerpenguins` data
+set](https://allisonhorst.github.io/palmerpenguins/) with the [DuckDB
+backend](https://ibis-project.org/backends/DuckDB/).
+
+## Setup
+
+```
+In [8]: from ibis.interactive import *
+
+In [11]: t = ex.penguins.fetch()
+
+In [12]: t
+Out[12]:
+┏━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━┓
+┃ species ┃ island    ┃ bill_length_mm ┃ bill_depth_mm ┃ flipper_length_mm ┃ body_mass_g ┃ sex    ┃ year  ┃
+┡━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━┩
+│ string  │ string    │ float64        │ float64       │ int64             │ int64       │ string │ int64 │
+├─────────┼───────────┼────────────────┼───────────────┼───────────────────┼─────────────┼────────┼───────┤
+│ Adelie  │ Torgersen │           39.1 │          18.7 │               181 │        3750 │ male   │  2007 │
+│ Adelie  │ Torgersen │           39.5 │          17.4 │               186 │        3800 │ female │  2007 │
+│ Adelie  │ Torgersen │           40.3 │          18.0 │               195 │        3250 │ female │  2007 │
+│ Adelie  │ Torgersen │            nan │           nan │                 ∅ │           ∅ │ ∅      │  2007 │
+│ Adelie  │ Torgersen │           36.7 │          19.3 │               193 │        3450 │ female │  2007 │
+│ Adelie  │ Torgersen │           39.3 │          20.6 │               190 │        3650 │ male   │  2007 │
+│ Adelie  │ Torgersen │           38.9 │          17.8 │               181 │        3625 │ female │  2007 │
+│ Adelie  │ Torgersen │           39.2 │          19.6 │               195 │        4675 │ male   │  2007 │
+│ Adelie  │ Torgersen │           34.1 │          18.1 │               193 │        3475 │ ∅      │  2007 │
+│ Adelie  │ Torgersen │           42.0 │          20.2 │               190 │        4250 │ ∅      │  2007 │
+│ …       │ …         │              … │             … │                 … │           … │ …      │     … │
+└─────────┴───────────┴────────────────┴───────────────┴───────────────────┴─────────────┴────────┴───────┘
+```
+
+## Examples
+
+### Normalization
+
+Let's say you want to compute the
+[z-score](https://en.wikipedia.org/wiki/Standard_score) of every numeric column
+and replace the existing data with that normalized value. Here's how you'd do
+that with selectors:
+
+```
+In [13]: t.mutate(s.across(s.numeric(), (_ - _.mean()) / _.std()))
+Out[13]:
+┏━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━┓
+┃ species ┃ island    ┃ bill_length_mm ┃ bill_depth_mm ┃ flipper_length_mm ┃ body_mass_g ┃ sex    ┃ year      ┃
+┡━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━┩
+│ string  │ string    │ float64        │ float64       │ float64           │ float64     │ string │ float64   │
+├─────────┼───────────┼────────────────┼───────────────┼───────────────────┼─────────────┼────────┼───────────┤
+│ Adelie  │ Torgersen │      -0.883205 │      0.784300 │         -1.416272 │   -0.563317 │ male   │ -1.257484 │
+│ Adelie  │ Torgersen │      -0.809939 │      0.126003 │         -1.060696 │   -0.500969 │ female │ -1.257484 │
+│ Adelie  │ Torgersen │      -0.663408 │      0.429833 │         -0.420660 │   -1.186793 │ female │ -1.257484 │
+│ Adelie  │ Torgersen │            nan │           nan │               nan │         nan │ ∅      │ -1.257484 │
+│ Adelie  │ Torgersen │      -1.322799 │      1.088129 │         -0.562890 │   -0.937403 │ female │ -1.257484 │
+│ Adelie  │ Torgersen │      -0.846572 │      1.746426 │         -0.776236 │   -0.688012 │ male   │ -1.257484 │
+│ Adelie  │ Torgersen │      -0.919837 │      0.328556 │         -1.416272 │   -0.719186 │ female │ -1.257484 │
+│ Adelie  │ Torgersen │      -0.864888 │      1.240044 │         -0.420660 │    0.590115 │ male   │ -1.257484 │
+│ Adelie  │ Torgersen │      -1.799025 │      0.480471 │         -0.562890 │   -0.906229 │ ∅      │ -1.257484 │
+│ Adelie  │ Torgersen │      -0.352029 │      1.543873 │         -0.776236 │    0.060160 │ ∅      │ -1.257484 │
+│ …       │ …         │              … │             … │                 … │           … │ …      │         … │
+└─────────┴───────────┴────────────────┴───────────────┴───────────────────┴─────────────┴────────┴───────────┘
+```
+
+### What's Up With the `year` Column?
+
+Whoops, looks like we included `year` in our normalization because it's an
+`int64` column (and therefore numeric) but normalizing the year doesn't make
+sense.
+
+We can exclude `year` from the normalization using another selector:
+
+```
+In [14]: t.mutate(s.across(s.numeric() & ~s.c("year"), (_ - _.mean()) / _.std()))
+Out[14]:
+┏━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━┓
+┃ species ┃ island    ┃ bill_length_mm ┃ bill_depth_mm ┃ flipper_length_mm ┃ body_mass_g ┃ sex    ┃ year  ┃
+┡━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━┩
+│ string  │ string    │ float64        │ float64       │ float64           │ float64     │ string │ int64 │
+├─────────┼───────────┼────────────────┼───────────────┼───────────────────┼─────────────┼────────┼───────┤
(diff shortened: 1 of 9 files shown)


ibis/expr/selectors.py
```diff
@@ -284,3 +307,137 @@ def matches(regex: str | re.Pattern) -> Selector:
...
+        return expanded
+
+
+@public
+def across(selector, func) -> Across:
+    return Across(selector, func)
+
+
+class IfAnyAll(Selector):
+    __slots__ = "selector", "predicate", "summarizer"
```
kszucs, line 370: You could use `Annotable` or `Concrete` here to reduce boilerplate.

docs/blog/selectors.md
```diff
@@ -0,0 +1,386 @@
...
+│ Gentoo  │ Biscoe │      -0.326062 │     -1.510481 │         -1.108253 │   -1.043442 │ female │  2007 │
+│ Gentoo  │ Biscoe │      -0.682990 │     -0.389427 │         -0.954050 │   -0.547525 │ female │  2007 │
+│ Gentoo  │ Biscoe │      -0.261167 │      0.323970 │          0.279571 │    0.245943 │ male   │  2007 │
+│ Gentoo  │ Biscoe │      -1.364397 │     -1.612395 │         -1.262455 │   -1.340993 │ female │  2007 │
+│ Gentoo  │ Biscoe │      -0.228719 │      0.425884 │         -0.337240 │    0.146759 │ male   │  2007 │
+│ …       │ …      │              … │             … │                 … │           … │ …      │     … │
+└─────────┴────────┴────────────────┴───────────────┴───────────────────┴─────────────┴────────┴───────┘
+```
+
+Since this "desugars" into a plain selection as if you had called `select` or
```
gforsyth, line 133:
```suggestion
Since ibis translates this into a plain selection as if you had called `select` or
```

docs/blog/selectors.md
```diff
@@ -0,0 +1,386 @@
...
+└─────────┴───────────┴────────────────┴───────────────┴───────────────────┴─────────────┴────────┴───────┘
+```
+
+## Examples
+
+### Normalization
+
+Let's say you want to compute the z-score of every numeric column and replace
+the existing data with that normalized value. Here's how you'd do that with
+selectors:
```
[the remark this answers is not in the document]
gforsyth: This suggestion, though, might be more appropriate for a tutorial and less for a blog post, feel free to ignore!

docs/blog/selectors.md
```diff
@@ -0,0 +1,386 @@
...
+└─────────┴───────────┴────────────────┴───────────────┴───────────────────┴─────────────┴────────┴───────┘
+```
+
+### Multiple Computations per Column
+
+What if I want to compute multiple things? Heck yeah!
+
+```
+In [26]: t.group_by("sex").mutate(
+    ...:     s.across(s.numeric() & ~s.c("year"), dict(cent=_ - _.mean(), zscore=(_ - _.mean()) / _.std()))
```
gforsyth, line 223:
```suggestion
    ...:     s.across(s.numeric() & ~s.c("year"), dict(center=_ - _.mean(), zscore=(_ - _.mean()) / _.std()))
```

docs/blog/selectors.md
```diff
@@ -0,0 +1,386 @@
...
+│ Adelie  │ Dream  │           43.2 │          18.5 │               192 │        4100 │ male   │  2008 │                3.087500 │              -0.387500 │                       -3.0 │ … │
+│ Adelie  │ Dream  │           39.5 │          16.7 │               178 │        3250 │ female │  2007 │                1.644444 │              -1.144444 │                       -7.0 │ … │
+│ Adelie  │ Dream  │           39.5 │          17.8 │               188 │        3300 │ female │  2007 │                1.644444 │              -0.044444 │                        3.0 │ … │
+│ …       │ …      │              … │             … │                 … │           … │ …      │     … │                       … │                      … │                          … │ … │
+└─────────┴────────┴────────────────┴───────────────┴───────────────────┴─────────────┴────────┴───────┴─────────────────────────┴────────────────────────┴────────────────────────────┴───┘
+```
+
+### Filtering Selectors
+
+You can also express complex filters more concise.
```
gforsyth, line 291:
```suggestion
You can also express complex filters more concisely.
```

docs/blog/selectors.md
```diff
@@ -0,0 +1,386 @@
...
+
+Let's say we only want to keep rows where all the bill size related columns' absolute values are greater than 2.
```
gforsyth, line 293: bill size z_score related columns?

gforsyth (commented):
Features and blog post look _sweet_.  Some small comments on the blog post, but this is awesome.

docs/blog/selectors.md
gforsyth: Hmm, I guess below you call this `centered` -- either is good, but nice to be consistent.

docs/blog/selectors.md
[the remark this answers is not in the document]
cpcloud: Is a link to wikipedia a solid middle ground?

docs/blog/selectors.md
cpcloud: I did this this cut down the width of the table a bit. I'll change the other to `cent`.

docs/blog/selectors.md
cpcloud: SGTM!

docs/blog/selectors.md
cpcloud: Went with `centered` and just used another selector to subset some columns :)

ibis/expr/selectors.py
cpcloud: Done.

jreback:
how does the example in the blog get columns that end with _centered and _zscore (3rd or 4th one down)

when you are actually passing dict(centered=...)?

ibis/expr/selectors.py
```diff
@@ -284,3 +298,157 @@ def matches(regex: str | re.Pattern) -> Selector:
...
+    def step(self):
+        return self.slice.step
+
+    @attribute.default
+    def __precomputed_hash__(self) -> int:
+        return hash((self.__class__, self.start, self.stop, self.step))
+
+
+class RangeSelector(Selector):
+    key = rlz.one_of((rlz.instance_of((str, int)), rlz.coerced_to(HashableSlice)))
```
kszucs, line 411: Type annotations here?

ibis/expr/types/groupby.py
```diff
@@ -62,9 +66,18 @@ def __init__(
         self, table, by, having=None, order_by=None, window=None, **expressions
     ):
         self.table = table
-        self.by = [_get_group_by_key(table, v) for v in util.promote_list(by)] + [
-            _get_group_by_key(table, v).name(k) for k, v in expressions.items()
-        ]
+        self.by = list(
```
kszucs, line 69: Can we add a reusable ` group_key_from()` rule?

kszucs: Then converting GroupedTable to annotable should be easier.

ibis/expr/types/relations.py
```diff
@@ -384,14 +399,25 @@ def aggregate(
         """
         import ibis.expr.analysis as an
 
-        metrics = util.promote_list(metrics)
-        metrics.extend(
-            self._ensure_expr(expr).name(name) for name, expr in kwargs.items()
+        metrics = itertools.chain(
```
kszucs, line 402: Can we add an `aggregation_from` or `metric_from` rule? Seems like that could be reused a couple of times. 

Fine in a follow-up too.

ibis/expr/types/groupby.py
kszucs: Fine in a follow-up too.

ibis/expr/types/groupby.py
```diff
@@ -43,15 +45,17 @@
 
 def _get_group_by_key(table, value):
     if isinstance(value, str):
-        return table[value]
+        return [table[value]]
```
kszucs, line 48: How about yielding instead?

ibis/expr/selectors.py
cpcloud: How do I use `coerced_to` in a type annotation?

cpcloud: Looks like maybe `Annotated` is the ticket.

ibis/expr/selectors.py
```diff
@@ -284,3 +298,157 @@ def matches(regex: str | re.Pattern) -> Selector:
...
+def if_any(selector: Selector, predicate: Deferred | Callable) -> IfAnyAll:
+    return IfAnyAll(selector=selector, predicate=predicate, summarizer=operator.or_)
+
+
+@public
+def if_all(selector: Selector, predicate: Deferred | Callable) -> IfAnyAll:
+    return IfAnyAll(selector=selector, predicate=predicate, summarizer=operator.and_)
+
+
+class HashableSlice(Concrete):
```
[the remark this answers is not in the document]
cpcloud: I can remove the base class, but moving it to `util.py` creates a huge circular import headache if I want to also use `Coercible`. I'd rather just leave it here for now.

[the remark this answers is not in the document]
cpcloud: I don't think removing Concrete makes sense. There's a `slice` field that needs to be initialized and if I only subclass `Coercible` there's no constructor. I'll leave it as is for now.

ibis/expr/selectors.py
cpcloud: Fixed.

ibis/expr/selectors.py
```diff
@@ -284,3 +298,157 @@ def matches(regex: str | re.Pattern) -> Selector:
...
+    slice: slice
+
+    @classmethod
+    def __coerce__(cls, slice):
```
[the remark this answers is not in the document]
cpcloud: Fixed.

cpcloud:
@jreback 

> how does the example in the blog get columns that end with _centered and _zscore (3rd or 4th one down)
> 
> when you are actually passing dict(centered=...)?

The `dict` keys are appended to the selected columns in the `Across.expand` implementation.

ibis/expr/types/relations.py
cpcloud: Let's do it in a follow-up.

ibis/expr/types/groupby.py
cpcloud: Let's do it in a follow-up.

kszucs (approved):
LGTM

Issue #5532, opened 2023-02-13
feat(io): add option to write a table expression as a partitioned parquet folder


ogrisel:
### Is your feature request related to a problem?

At the moment (since #4444) there is a generic implementation of the `to_parquet` method in `ibis/ibis/backends/base/__init__.py`'s `_FileIOHandler` that uses pyarrow's `ParquetWriter`:

```python
        import pyarrow.parquet as pq

        batch_reader = expr.to_pyarrow_batches(params=params)

        with pq.ParquetWriter(path, batch_reader.schema) as writer:
            for batch in batch_reader:
                writer.write_batch(batch)
```

This works for all backends that to support `to_pyarrow_batches` which is nice. However, this abstraction makes it impossible to leverage backend-specific features such as partitioned writing. For instance, since duckdb 0.7.0 it is possible to use:

```sql
COPY orders TO 'orders' (FORMAT PARQUET, PARTITION_BY (year, month));
```

See the [duckdb 0.7.0 announcement](https://duckdb.org/2023/02/13/announcing-duckdb-070.html#data-ingestionexport-improvements) for more details.

which is very nice to generate parquet files trees that can be very efficiently queried upon by any DB engine that supports predicate push downs down to the parquet folder structure, especially when those files are meant to be archived in a large cloud object store ("data lake").

### Describe the solution you'd like

Either find a way to generate partitioned parquet writes while still relying on `to_pyarrow_batches` of the generic `_FileIOHandler` but I am not sure how this would be possible with the `ParquetWriter` API.

Alternatively, find a way to use [pq.write_to_dataset](https://arrow.apache.org/docs/python/parquet.html#writing-to-partitioned-datasets) but I am not sure if this is possible to use this API incrementally with a generator that yields arrow tables, chunk by chunk.

Finally, one could specialize the `to_parquet` method for datasets that already support writing to partitioned parquet datasets such as DuckDB. But then I am not sure about the impact of non-default data set handling, e.g. writing to an `s3://` folder. The potentital discrepancies between pyarrow filesystem support and duckdb's might cause new UX problems.

### What version of ibis are you running?

ibis master

### What backend(s) are you using, if any?

DuckDB

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gforsyth:
Thanks for opening this @ogrisel.

I agree that the option to write out to partitioned parquet folders is both useful and desirable.

And I think that you've hit on the available approaches, and their downsides.

[...]

PR #5603, opened 2023-02-23
feat(dataset): add to_dataset for fine-grained file outputs

gforsyth:
Opening this as a draft because I think there are several unanswered questions.

This is a potential solution for #5532.

As mentioned there, `pyarrow.dataset.write_dataset` will take a `RecordBatchReader` and write it out to a directory, optionally partitioning the files on column value(s), and supports both `directory` and `hive` naming conventions for those directories.

I was replacing `to_parquet` with this approach, when I realized that `write_dataset` also supports `csv` (and `feather`).
Do we also want to rewrite `to_csv` to allow for partitioned outputs?  Maybe?  But I think for CSV the behavior of always writing to a directory and not a single file is much less expected than for parquet.  I think that isn't a good idea.

But to force `format="parquet"` in a call to `write_dataset` seems unnecessarily constraining.

So, one proposal (contained herein), is to offer a `to_dataset` option alongside `to_csv` and `to_parquet`.  The file-specific methods write out a single file (for now) and `to_dataset` is there for users who need more control over the outputs.

On the plus side
- this keeps the "simple" behavior available, but opens up the "advanced" behavior.
- if we dispatch to backends to handle the writing and they have specialized but less-fully-featured options, we can use both, so `to_parquet` would call out to e.g. `clickhouse` specialized parquet writer but `to_dataset` is available if you need hive naming
- we avoid an explosion of keyword arguments to handle partitioning in `to_parquet` and `to_csv` (maybe)

On the down side:
- I would wager 95+% of users will want `format="parquet"` and I'm making them type that every time (we could default to `parquet`)
- there's more potential for splintering of behavior across backends between what `to_dataset(..., format="parquet")` does and what `to_parquet(...)` does (and `csv`, mutatis mutandis)
- it's probably not immediately clear that `to_dataset` is an option for writing out files

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -370,6 +370,61 @@ class _FileIOHandler:
         )
 
     @util.experimental
+    def to_dataset(
+        self,
+        expr: ir.Table,
+        path: str | Path,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        format: str | None = None,
+        **kwargs: Any,
+    ):
+        """Write results to a directory, optionally partitioning or splitting.
+
+        This method is eager and will execute the associated expression
+        immediately.
+
+        Parameters
+        ----------
+        expr
+            The ibis expression to execute and write to file(s).
+        path
+            The data source. A string or Path to the output directory.
+        params
+            Mapping of scalar parameter expressions to value.
+        format
+            Output format. Supported values are "parquet", "csv", "feather".
+        **kwargs
+            Additional keyword arguments passed to pyarrow.dataset.write_dataset
+
+        https://arrow.apache.org/docs/python/generated/pyarrow.dataset.write_dataset.html
+
+        Example
+        -------
+        >>> con = ibis.duckdb.connect("penguins.duckdb")
+        >>> penguins = con.tables.penguins
+        >>> con.to_dataset(penguins,
+                           "/tmp/penguins",
+                           partitioning=["species", "island"],
+                           )
+
+        >>> con.to_dataset(penguins,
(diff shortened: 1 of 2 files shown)


NickCrews:
What about adding a `partition` kwarg to `to_parquet()` and `to_csv()`? Default is to write a single file, but available to partition.

> it's probably not immediately clear that to_dataset is an option for writing out files

I agree.

I also don't like adding `format="parquet"` as a kwarg to a function. Choosing the format is a first-class choice. That's why we have `to_csv()` and `to_parquet()`, not `to_file(format=...)`

gforsyth:
> What about adding a `partition` kwarg to `to_parquet()` and `to_csv()`? Default is to write a single file, but available to partition.

We could do that -- a possible issue with that approach is that the option to partition comes with _other_ keyword arguments that we would (maybe?) want to provide.  
Because if you want to partition, you also need to specify what columns to partition on, and then there are several methods of organizing/naming those files.

We could add a `partition_on` keyword argument, that if it's `not None` we partition out with some reasonable defaults, but allow passing through kwargs to `write_dataset` to specify other behavior.

> I also don't like adding `format="parquet"` as a kwarg to a function. Choosing the format is a first-class choice. That's why we have `to_csv()` and `to_parquet()`, not `to_file(format=...)`

I agree, if we stick with `to_dataset` I would probably default it to `parquet`, since that's likely to be the majority use-case.

gforsyth:
I'm closing this out -- we'll probably change the name, but in any case, getting consistent _read_ behavior for partitioned parquet files is non-trivial and we don't want to write out files that we can't read back in.

PR #5604, opened 2023-02-23
refactor(sqlalchemy): less eager check for whether to reload table metadata

gforsyth:
I _think_ this method of checking for a need to refresh table metadata is less expensive for backends with higher network latency, but I haven't timed it yet.

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -389,8 +389,13 @@ class BaseAlchemyBackend(BaseSQLBackend):
     ) -> sa.Table:
         # If the underlying table (or more likely, view) has changed, remove it
         # to ensure a correct reflection
-        if autoload and self.inspector.has_table(name):
-            self.meta.remove(sa.Table(name, self.meta))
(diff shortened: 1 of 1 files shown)


ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -389,8 +389,13 @@ def _get_sqla_table(
     ) -> sa.Table:
         # If the underlying table (or more likely, view) has changed, remove it
         # to ensure a correct reflection
-        if autoload and self.inspector.has_table(name):
-            self.meta.remove(sa.Table(name, self.meta))
+        if (sa_table := self.meta.tables.get(name, None)) is not None and autoload:
+            # check if self._metadata and self.meta agree
+            # (ibis vs. sqlalchemy bookkeeping)
+            sa_cols = {col.name for col in sa_table.columns}
+            ibis_cols = {col for col, _ in self._metadata(name)}
```
cpcloud, line 396: This call won't be particularly cheap since it's making a database query. I bet that it'll be awash with respect to `self.inspector.has_table` since that probably also makes at least one database query.

On the other hand, I'll be happy if this is in fact faster.

gforsyth:
Yep, this is slower.

PR #5609, opened 2023-02-24
chore(examples): add Lahman baseball data examples

cpcloud:
This PR adds the Lahman baseball datasets that ship with the eponymous package to our example registry. Ideally we can start using these for test data ingest soon.


gforsyth (approved):
:shipit:

PR #5621, opened 2023-02-27
feat(snowflake): add support for alias in snowflake

thegreymatter:
add support for alias() in snowflake.

### DIFF
--- a/ibis/backends/snowflake/__init__.py
+++ b/ibis/backends/snowflake/__init__.py
@@ -236,3 +236,8 @@ class Backend(BaseAlchemyBackend):
                 auto_create_table=True,
                 quote_identifiers=False,
             )
+
+    def _get_temp_view_definition(
+        self, name: str, definition: sa.sql.compiler.Compiled
+    ) -> str:
+        yield f"CREATE OR REPLACE TEMPORARY VIEW {name} AS {definition}"
--- a/ibis/backends/tests/test_dot_sql.py
+++ b/ibis/backends/tests/test_dot_sql.py
@@ -191,7 +191,6 @@ def test_table_dot_sql_does_not_clobber_existing_tables(con):
 
 @table_dot_sql_notimpl
 @dot_sql_notimpl
-@dot_sql_notyet
 @dot_sql_never
 @pytest.mark.notimpl(["trino"])
 def test_dot_sql_alias_with_params(backend, alltypes, df):


cpcloud (approved):
Fixed a couple xfailing snowflake tests that now start to pass after this PR!

LGTM!

cpcloud:
@thegreymatter Just out of curiosity is there any functionality missing from the snowflake backend that we could add to avoid you needing to reach for `.sql`/`.alias`?

These APIs are meant as escape hatches for when ibis is missing functionality and as entry points into ibis when you're saddled with a large amount of existing SQL, and I'm wondering if you're hitting the former case.

thegreymatter:
> @thegreymatter Just out of curiosity is there any functionality missing from the snowflake backend that we could add to avoid you needing to reach for `.sql`/`.alias`?
> 
> These APIs are meant as escape hatches for when ibis is missing functionality and as entry points into ibis when you're saddled with a large amount of existing SQL, and I'm wondering if you're hitting the former case.

@cpcloud there are few reasons for us to use this capability. 
+ making complex queries simpler to debug and analyze in case of error or weird behavior (basically a cheap CTE alternative) 
+ ability to bring in already existing queries and code while we do the transition, i.e. sort of backward compatibility. 
+ few capabilities we use SQL  - PIVOT, SAMPLE, TABLESAMPLE, MERGE (i.e. upserts), APPROX_PERCENTILE (and  its family of functions)

cpcloud:
@thegreymatter Awesome, thanks for your reply. This is all very useful information for us.

Couple of thoughts:

> @cpcloud there are few reasons for us to use this capability.
> 
> * making complex queries simpler to debug and analyze in case of error or weird behavior (basically a cheap CTE alternative)

Nice. We would like to improve the error reporting in ibis overall. It sounds like there's maybe an area of "successfully compiled but failed-to-run SQL queries" that we can focus on as part of that effort.

> * ability to bring in already existing queries and code while we do the transition, i.e. sort of backward compatibility.

Sweet, this is definitely one of major use cases for `.sql` :)

> * few capabilities we use SQL  - PIVOT, SAMPLE, TABLESAMPLE, MERGE (i.e. upserts), APPROX_PERCENTILE (and  its family of functions)

Some of these are 100% candidates for inclusion into core ibis:

- `PIVOT`, definitely
- `SAMPLE`, definitely
- `MERGE`, not sure, but maybe. We'd have to look at how to roll this into the expression API if that makes sense.
- `APPROX_PERCENTILE`, we do have `quantile` but no explicit `approx_quantile`, definitely a candidate for a new API. The state accumulation/estimation APIs may also be eligible but we'd have to look into it more.

Issue #5393, opened 2023-01-31
feat: support creating temporary tables


jcrist:
We should support creating temporary tables. I _think_ the cleanest way to do this would be to add a new `temp` (or `temporary`?) boolean kwarg to `create_table`, which defaults to `False`.

For backends that don't support temporary tables I think we should error nicely rather than trying to emulate things with a non-temporary table and an `atexit` handler. Since these APIs modify the backing database and potentially move lots of data around, I think relying on `atexit` to clean things up would be too risky (`atexit` handlers aren't guaranteed to run, and often don't in some common jupyter notebook workflows).

jcrist:
One potential use case for this is computing an intermediate result and stashing it in an efficient way native to the backend. This would be similar to `.persist()` in dask. This was also discussed in #2339.

A concrete use case: currently our `read_csv`/`read_parquet`/`read_json` calls all create _views_ of the data, meaning any computations will have to redo the data parsing. Adding a new `.persist()` (or `.checkpoint()`, there's a few common names here) method that called `con.create_table(f"temp_{uuid}", expr, temp=True)` and returned the temporary table would be a more intuitive way of caching the parsed table inside the backing backend.

```python
t = ibis.read_csv("some_large_file.csv").persist()
```

Currently the user has to do:

```python
t = ibis.read_csv()
con = ibis.get_backend()
con.create_table("temp", t)
t = con.tables["temp"]
# do some work...
con.drop_table("temp")  # since it's not a temporary table
```

lgonzalezsa:
Was exploring the usage of ibis mainly because of the deferred execution feature, to speed up my analysis. One of the main workflows I do have is the usage of temp tables in MSSQL engines, facing issues and just found this. Hope this can be implemented, the workaround using tables works but is not ideal.

ogrisel:
I was just googling this myself after [reading this section](https://www.pola.rs/posts/the-expressions-api-in-polars-is-amazing/#cherry-on-top-clever-caching) of a polars tutorial that speaks about how natural it is with polars to cache intermediate results when interactively crafting a data processing pipeline.

I think it would be nice to have such streamlined caching API with various kinds of possible caching strategies:

- backend-specific, temporary tables as you suggest (I assume tied to the Python GC);
- backend-agnostic in-memory results (e.g. a-la `ibis.memtable`);
[...]

PR #5623, opened 2023-02-27
feat(sql): create temp table

thegreymatter:
Initial commit for allowing to create temp table which is a first step towards ` .persist()` https://github.com/ibis-project/ibis/issues/5393 

My current dilemmas -
* Still looking for a way to prove a table is temporary in tests without performing introspection on schemas. my first thought is opening two connections one after the other. not sure how it works with transactions.
* sqlite limits temp tables to temp databases only, I wonder what is the best approach to handle this limitation, one approach can be to transfer the responsibility for that to the end user, or to block this capability all together with a nice error. 

Will be happy for your thoughts.

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -174,6 +174,7 @@ class BaseAlchemyBackend(BaseSQLBackend):
         schema: sch.Schema | None = None,
         database: str | None = None,
         force: bool = False,
+        temp: bool = False,
     ) -> None:
         """Create a table.
 
@@ -189,6 +190,8 @@ class BaseAlchemyBackend(BaseSQLBackend):
             A database
         force
             Check whether a table exists before creating it
+        temp
+            Should the table be temporary for the session.
         """
         import pandas as pd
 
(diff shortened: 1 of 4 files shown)


ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -173,6 +173,7 @@ def create_table(
         schema: sch.Schema | None = None,
         database: str | None = None,
         force: bool = False,
+        temporary: bool = False,
```
cpcloud, line 176: Can we call this `temp`?

cpcloud:
> Initial commit for allowing to create temp table which is a first step towards ` .persist()` #5393
> 
> My current dilemmas -
> 
> * Still looking for a way to prove a table is temporary in tests without performing introspection on schemas. my first thought is opening two connections one after the other. not sure how it works with transactions.

Creating a second connection is probably the only way to reliably do this right now. Even for sqlalchemy, most dialects do not implement the `get_temp_table_names` Dialect method which we could otherwise use.

> * sqlite limits temp tables to temp databases only, I wonder what is the best approach to handle this limitation, one approach can be to transfer the responsibility for that to the end user, or to block this capability all together with a nice error.

Hm, interesting. I guess we should raise an error for now.

ibis/backends/tests/test_client.py
```diff
@@ -135,6 +135,45 @@ def test_create_table_from_schema(con, new_schema, temp_table):
...
+
+    # verify table exist in the current session
+    backend_mapping = backend_type_mapping.get(con.name, dict())
+    for column_name, column_type in table.schema().items():
+        assert (
+            backend_mapping.get(new_schema[column_name], new_schema[column_name])
+            == column_type
+        )
+
+    con.reconnect()
```
cpcloud, line 172: Nice!

cpcloud (approved):
This PR LGTM. I'm going to rebase it and let it do one more CI run then merge it!

PR #5624, opened 2023-02-27
feat(common): hold typehint in the annotation objects

kszucs:
The argument annotation hasn't been holding the passed typehint, just the validator created from it. While experimenting with `egg-smol` I had to create a parallel type system to feed the library with but I didn't want to repeat every definitions. 
In order to do that I needed the actual types, not the validators. In addition to that discovered that we haven't been properly supporting generic type coercions.

### DIFF
--- a/ibis/common/annotations.py
+++ b/ibis/common/annotations.py
@@ -5,6 +5,7 @@ import inspect
 from typing import Any
 
 from ibis.common.collections import DotDict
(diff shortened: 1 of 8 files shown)


cpcloud:
@kszucs Can you rebase to pick up #5627?

cpcloud (approved):
Awesome, glad to see this!

PR #5640, opened 2023-03-01
docs: add examples to table expression API

cpcloud:
This PR fills out most of the table APIs examples sections with actual running examples.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -222,7 +222,7 @@ def param(type: dt.DataType) -> ir.Scalar:
       predicates:
         r0.timestamp_col >= $(date)
         r0.timestamp_col <= $(date)
-    sum: Sum(r1.value)
+    Sum(value): Sum(r1.value)
     """
     return ops.ScalarParameter(type).to_expr()
 
@@ -269,7 +269,7 @@ def schema(
     >>> sc = schema(names=['foo', 'bar', 'baz'],
     ...             types=['string', 'int64', 'boolean'])
     >>> sc = schema(dict(foo="string"))
-    >>> sc = schema(Schema(['foo'], ['string']))  # no-op
+    >>> sc = schema(Schema(dict(foo="string")))  # no-op
 
     Returns
     -------
@@ -304,9 +304,12 @@ def table(
     --------
     Create a table with no data backing it
 
-    >>> t = ibis.table(schema=dict(a="int", b="string"))
+    >>> import ibis
+    >>> ibis.options.interactive
+    False
+    >>> t = ibis.table(schema=dict(a="int", b="string"), name="t")
     >>> t
-    UnboundTable: unbound_table_0
+    UnboundTable: t
       a int64
       b string
     """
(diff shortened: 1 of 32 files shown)


ibis/expr/api.py
```diff
@@ -453,18 +456,20 @@ def desc(expr: ir.Column | str) -> ir.Value:
...
-      metrics:
-        count: Count(t)
-      by:
-        g: r0.g
-    Selection[r1]
-      sort_keys:
-        desc|r1.count
+    >>> ibis.options.interactive = True
+    >>> t = ibis.examples.penguins.fetch()
+    >>> t[["species", "year"]].order_by(_.year.desc()).head()
```
gforsyth, line 461: How is this completing if `_` hasn't been imported from `ibis`?

cpcloud: Ah, because it's an interactive session, so `_` is `t` :)

That's not intentional, so I'll fix it up.

Nice catch.

cpcloud: Fixed.

ibis/expr/types/relations.py
```diff
@@ -263,62 +283,123 @@ def _ensure_expr(self, expr):
...
+        ┃ fruit  ┃ price   ┃
+        ┡━━━━━━━━╇━━━━━━━━━┩
+        │ string │ float64 │
+        ├────────┼─────────┤
+        │ apple  │    0.50 │
+        │ apple  │    0.50 │
+        │ banana │    0.25 │
+        │ orange │    0.33 │
+        └────────┴─────────┘
+        >>> t.group_by("fruit").agg(total_cost=_.price.sum(), avg_cost=_.price.mean())
```
gforsyth, line 380: Another errant `_`

ibis/expr/types/relations.py
```diff
@@ -396,6 +477,32 @@ def aggregate(
...
+        ┃ fruit  ┃ price   ┃
+        ┡━━━━━━━━╇━━━━━━━━━┩
+        │ string │ float64 │
+        ├────────┼─────────┤
+        │ apple  │    0.50 │
+        │ apple  │    0.50 │
+        │ banana │    0.25 │
+        │ orange │    0.33 │
+        └────────┴─────────┘
+        >>> t.aggregate(by=["fruit"], total_cost=_.price.sum(), avg_cost=_.price.mean(), having=_.price.sum() < 0.5)
```
gforsyth, line 497: Needs `_` import

gforsyth (changes requested):
This looks good!  Two more `_` to fix, but other than that I think this can go in.

ibis/expr/types/relations.py
cpcloud: This all turns out to be because once you import something it stays in scope for the entire pytest session 😬

gforsyth: :grimacing: that is... not great

pyproject.toml
```diff
@@ -281,22 +297,10 @@ filterwarnings = [
   "ignore:Deprecated call to `pkg_resources\\.declare_namespace\\('google.*'\\):DeprecationWarning",
   # pyspark on python 3.11
   "ignore:typing\\.io is deprecated:DeprecationWarning",
+  # warnings from google's use of the cgi module
```
gforsyth, line 300: ewww

gforsyth (approved):
this is great!  I think all the doctest examples are properly standalone.

PR #5646, opened 2023-03-01
feat: Add 'cache' API to table

thegreymatter:
adding `persist()` API to table  expression to solve https://github.com/ibis-project/ibis/issues/5393 

Currently it works for pandas,dask,polars and sqlalchemy backends which support temp tables.
I believe I can make it work for spark with a little bit of tweaking.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -879,6 +879,37 @@ class BaseBackend(abc.ABC, _FileIOHandler):
             f"{cls.name} backend has not implemented `has_operation` API"
         )
 
+    def _cache(self, expr):
+        """Cache the provided expression. All subsequent operations on the returned expression will be performed on the cached data.
+
+        Parameters
+        ----------
+        expr
+            Table expression to cache
+
+        Returns
+        -------
+        Expr
+            Cached table
+
+        """
+        raise NotImplementedError(
+            f"{self.name} backend has not implemented `cache` API"
+        )
+
+    def _release_cache(self, expr):
+        """Releases the provided cached expression.
+
+        Parameters
+        ----------
+        expr
+            Cached expression to release
+
+        """
+        raise NotImplementedError(
+            f"{self.name} backend has not implemented `release_cache` API"
+        )
+
 
 @functools.lru_cache(maxsize=None)
 def _get_backend_names() -> frozenset[str]:
--- a/ibis/backends/dask/__init__.py
+++ b/ibis/backends/dask/__init__.py
@@ -13,6 +13,7 @@ import ibis.backends.pandas.execution
 import ibis.config
 import ibis.expr.schema as sch
 import ibis.expr.types as ir
+from ibis import util
 from ibis.backends.dask.client import DaskDatabase, DaskTable, ibis_schema_to_dask
 from ibis.backends.dask.core import execute_and_reset
 from ibis.backends.pandas import BasePandasBackend
@@ -135,3 +136,15 @@ class Backend(BasePandasBackend):
     ):
         """Create a table."""
         super().create_table(table_name, obj=obj, schema=schema)
+
+    def _cache(self, expr):
+        persisted_table_name = util.generate_unique_table_name("cache")
+        df = self.compile(expr).persist()
+        self.load_data(persisted_table_name, df)
+        return self.table(persisted_table_name)
+
+    def _release_cache(self, expr):
+        if isinstance(expr._arg, DaskTable):
+            del self.dictionary[expr._arg.name]
+        else:
+            raise NotImplementedError(f"{expr.arg} is not releasable.")
--- a/ibis/backends/polars/__init__.py
+++ b/ibis/backends/polars/__init__.py
@@ -13,8 +13,10 @@ import ibis.expr.analysis as an
 import ibis.expr.operations as ops
 import ibis.expr.schema as sch
 import ibis.expr.types as ir
+from ibis import util
 from ibis.backends.base import BaseBackend
 from ibis.backends.polars.compiler import translate
+from ibis.expr.operations import DatabaseTable
 from ibis.util import normalize_filename
 
 if TYPE_CHECKING:
@@ -350,3 +352,15 @@ class Backend(BaseBackend):
         self._import_pyarrow()
         table = self._to_pyarrow_table(expr, params=params, limit=limit, **kwargs)
         return table.to_reader(chunk_size)
+
+    def _cache(self, expr):
+        persisted_table_name = util.generate_unique_table_name("cache")
+        lf = self.compile(expr).cache()
+        self.load_data(persisted_table_name, lf)
+        return self.table(persisted_table_name, expr.schema)
+
+    def _release_cache(self, expr):
+        if isinstance(expr._arg, DatabaseTable):
+            del self._tables[expr._arg.name]
+        else:
+            raise NotImplementedError(f"{expr._arg} is not releasable.")
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -27,7 +27,6 @@ if TYPE_CHECKING:
     from ibis.expr.selectors import IfAnyAll, Selector
     from ibis.expr.types.groupby import GroupedTable
 
-
 _ALIASES = (f"_ibis_view_{n:d}" for n in itertools.count())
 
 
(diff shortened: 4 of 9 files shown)


krzysztof-kwitt:
FYI: Snowpark named this function as [`cache_result`](https://docs.snowflake.com/developer-guide/snowpark/reference/python/api/snowflake.snowpark.DataFrame.cache_result.html). I think this better describes how this function will be used and also shows the volatility of this data. The name "persist" does not describe whether this data will be stored permanently or temporarily.
Besides, Snowpark allows you to use the result of a function as a context manager.
```
with df2.cache_result() as df4:
...     df4.collect()
```

cpcloud:
Naming aside, I wonder if it's valuable to implement this feature in such a way that the following things are true:

1. Execution happens only once per session per unique expression per backend.
2. The expression is lazy, that is, no computation is triggered when the API is called.

If so, I would suggest that we merge this PR once it's been reviewed and then work on follow ups to get to the desired end state.

Here's an example of what I'm thinking:

```
t = ...  # something expensive
s = t.persist()  # returns ~immediately, stores enough information to later compute the result
df1 = s.execute()  # run the `t` query -> cache it -> returns `select * from cached_thing`
df2 = s.execute()  # returns `select * from cached_thing`
df3 = s.agg(...)  # operation occurs on the cached data, never reruns the `t` query
```

jcrist:
> The expression is lazy, that is, no computation is triggered when the API is called.

What kind of workloads would this enable that running and caching eagerly & explicitly would not? I'm not against it, but I worry that delayed execution would make it harder to debug what's happening. An explicit execution point is easy from a user perspective to reason about - something lazy *may* be harder to reason about. A few "what ifs":
- What if the cached expression errors during execution, would it be easy to track down what expression evaluation caused the error?
- What if the user calls `.persist()` on the same expression twice, does that share the same temp table?

cpcloud:
> What kind of workloads would this enable that running and caching eagerly & explicitly would not?

I think the possible workloads are unrelated to the choice of deferred or not, which is a UX concern.

> What if the cached expression errors during execution, would it be easy to track down what expression evaluation caused the error?

What about deferring the execution of the persist would make this problem worse than it is now? A "flat" SQL query would be executed, or at worst one statement after the other.

> What if the user calls .persist() on the same expression twice, does that share the same temp table?

I think it would. I'm not aware of a reason why this wouldn't be the right thing to do.

jcrist:
> I think the possible workloads are unrelated to the choice of deferred or not, which is a UX concern.

Sure. I guess I'm just asking why this change would be worth the trouble. I'm not as deep of an ibis expert as you, but making this lazy feels to me to be more work than making it eager. I'm wondering what benefits making it lazy would bring to the user, that's all.

cpcloud:
I think the eagerness is a departure from nearly everything else outside of DDL.

It seems like the eager version forces a user out of a method chain if they don't want to execute the child expression of a `persist` call multiple times, and I don't like that an implementation choice _forces_ a specific style.

If someone wants to chain they should be able to, if they don't want to chain, that's fine too.

Perhaps I'm being a bit ideological 😅

cpcloud:
The value-add for persist in my mind is "transparent caching". To me that means I can insert a persist call wherever I want to avoid repeating computation, without having to think about whether I am inadvertently executing things twice.

ogrisel:
Maybe a concrete example would help feel the impacts of this design more concretely.

```python
preprocessed = (
    raw_data_source
    .pipe(costly_aggregate)  # output would be small to store on disk
    .persist()  # lazy
    .pipe(check_expansion)  # fast to compute, large output
)

preprocessed.pipe(additional_steps_for_a).to_parquet("preprocessed_a.parquet")
preprocessed.pipe(additional_steps_for_b).to_parquet("preprocessed_b.parquet")
```

If `persist()` is lazy, adding or removing it does not change the fact that only the last two lines trigger the actual computation, which is nice I think.

kszucs:
### Lazyness

I think ibis should prefer deferred expression whenever it's possible, so API-wise I'm in favour of a lazy persist operation. 
On the other hand we need to be careful since we would introduce state management including new questions, like how to invalidate the cache if the underlying data source has changes? Preferably we would handle that transparently for better user experience.

### Required IR changes
- The expression preceding the `persist()` call should have a stable-ish hash, so the backends can identify the same _base_ expression even between different python invocations (we actually have an issue for this https://github.com/ibis-project/ibis/issues/3987)
- We need to have one (or more in the future) `Persist` table operations which contains the persistable expression as well as possible persist options

### Imagining a possible pandas implementation

1. The execution model would locate the base tables of the persistable expression
2. Query their last change timestamp (and maybe other metadata highlighting whether it is changed or not) 
3. Generate the output file name based on the persistable expression's stable hash and the previously queried metadata about the data sources
4. If the file exists, use that as the datasource, and drop everything preceding the persist call 

These steps would need to run before actually executing anything, so that the expression can be rewritten to the simpler form using the persisted datasets.

Note, that this sounds fairly complicated thus I'd wouldn't pursue adding this feature in the pandas backend. 

### A possible clickhouse implementation

1. Locate the `Persist` operations before execution
2. Compile the persistable expressions to SQL
3. Generate a name based on the stable hash of those expressions
4. `CREATE` a materialized view for each persistable operation (views are dynamic so clickhouse would handle updates automatically)
5. Compile the input expression using the materialized views instead of the persisted expressions
6. Actually execute the SQL string.

One downside is that we can pollute the database with plenty of materialized views, imagine someone repeatedly editing the persistable expression in editable mode e.g. in a jupyter notebook. Clickhouse would quickly have plenty of unreadable materialized views in the database. Not sure how to prevent this though. 

### Possibly `PySpark` implementation

If I recall correctly pyspark has a builtin feature for this exact use case, so this should be the simplest.

### Conclusion

For me a lazy persist operation sounds doable, and would have great impacts for the execution times, especially in certain backends, *BUT* it's certainly more complicated to implement.

cpcloud:
>  think ibis should prefer deferred expression whenever it's possible, so API-wise I'm in favour of a lazy persist operation.

I am too, but I think we can still accept this PR as incremental progress towards some end state. We should see whether the lack of laziness is a footgun/pain point for users before taking on additional work to make the thing fully deferred.

> how to invalidate the cache if the underlying data source has changes?

Why do we have to solve this problem because we introduce this API? Is it even feasible without full support from the database?

thegreymatter:
took the liberty to rename the api to `cache()` , i've opted out of the cache_result options as the result part kind of redundent, and it allows us to be in sync with polars + spark.

also adding the ability to drop cache explicitly using context manager. it opens the ability to implement cache for SQL backends with no obvious TEMP table mechanism. 

+ adding my 2c regarding  eager\lazy. while I do agree lazy is probably much more efficient, after playing with this API interactively one of the nice thing I liked with the eager option is that it gave some kind of "predictability" regarding why my query is slow and why not. It really helped my data science team to understand at what point they need to cache things to get a consistent experience.

ibis/expr/types/relations.py
```diff
@@ -2370,6 +2369,79 @@ def to_pandas(self, **kwargs) -> pd.DataFrame:
         """
         return self.execute(**kwargs)
 
+    def cache(self) -> Table:
```
cpcloud, line 2372: +1 on the name `cache`. Not a huge fan of the redundancy of `cache_result`.

ibis/expr/types/relations.py
```diff
@@ -2370,6 +2369,79 @@ def to_pandas(self, **kwargs) -> pd.DataFrame:
...
+        """Cache the provided expression.
+
+         All subsequent operations on the returned expression will be performed on the cached data.
+         You may use the ``with`` statement to explicitly clean up the cached expression when it's not needed.
+
+        !!! note "cache_result will eagerly evaluate the expression"
```
cpcloud, line 2378:
```suggestion
        !!! note "This method eagerly evaluates the expression prior to caching"
```

ibis/expr/types/relations.py
```diff
@@ -2370,6 +2369,79 @@ def to_pandas(self, **kwargs) -> pd.DataFrame:
...
+
+        Returns
+        -------
+        Table
+            Cached table.
+
+
```
cpcloud, line 2385:
```suggestion
```

ibis/expr/types/relations.py
```diff
@@ -2370,6 +2369,79 @@ def to_pandas(self, **kwargs) -> pd.DataFrame:
...
+        │ Adelie  │ Torgersen │           39.3 │          20.6 │               190 │ … │
+        │ Adelie  │ Torgersen │           38.9 │          17.8 │               181 │ … │
+        │ Adelie  │ Torgersen │           39.2 │          19.6 │               195 │ … │
+        │ Adelie  │ Torgersen │           34.1 │          18.1 │               193 │ … │
+        │ Adelie  │ Torgersen │           42.0 │          20.2 │               190 │ … │
+        │ …       │ …         │              … │             … │                 … │ … │
+        └─────────┴───────────┴────────────────┴───────────────┴───────────────────┴───┘
+
+        Explicit cache cleanup
+        >>> import ibis
```
cpcloud, line 2412:
```suggestion
        Explicit cache cleanup

        >>> import ibis
```

ibis/util.py
```diff
@@ -555,3 +557,8 @@ def _absolufy_paths(name):
 
     source = _absolufy_paths(source)
     return source
+
+
+def generate_unique_table_name(namespace: str) -> str:
+    """Creates case-insensitive uuid4 unique table name."""
+    return f"_ibis_{namespace}_{np.base_repr(uuid.uuid4().int, 36)}".lower()
```
cpcloud, line 564: I'll adjust `ibis.util.guid()` to use the base 36 representation for more compact output. Just a note-to-self for me here.

ibis/backends/base/__init__.py
```diff
@@ -879,6 +879,37 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+            f"{self.name} backend has not implemented `cache` API"
+        )
+
+    def _release_cache(self, expr):
+        """Releases the provided cached expression.
+
+        Parameters
+        ----------
+        expr
+            Cacehd Expression to release
```
gforsyth, line 906:
```suggestion
            Cached expression to release
```

ibis/backends/base/__init__.py
```diff
@@ -879,6 +879,37 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
             f"{cls.name} backend has not implemented `has_operation` API"
         )
 
+    def _cache(self, expr):
+        """Cache the provided expression. All subsequent operations on the returned expression will be performed on the cached data .
```
gforsyth, line 883:
```suggestion
        """Cache the provided expression. All subsequent operations on the returned expression will be performed on the cached data.
```

ibis/backends/base/__init__.py
```diff
@@ -879,6 +879,37 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+
+        Parameters
+        ----------
+        expr
+            Table Expression to cache
```
gforsyth, line 888:
```suggestion
            Table expression to cache
```

ibis/backends/base/__init__.py
```diff
@@ -879,6 +879,37 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+
+        Returns
+        -------
+        Expr
+            Cached table.
```
gforsyth, line 893:
```suggestion
        Expr
            Cached table
```

Small nit, but I think docstring convention is to avoid periods at the end of argument descriptions

ibis/backends/dask/__init__.py
```diff
@@ -135,3 +136,15 @@ def create_table(
...
+        persisted_table_name = util.generate_unique_table_name("cache")
+        df = self.compile(expr).persist()
+        self.load_data(persisted_table_name, df)
+        return self.table(persisted_table_name)
+
+    def _release_cache(self, expr):
+        if isinstance(expr._arg, DaskTable):
+            del self.dictionary[expr._arg.name]
+        else:
+            raise NotImplementedError(f"{expr.arg} is not releaseable.")
```
gforsyth, line 150:
```suggestion
            raise NotImplementedError(f"{expr.arg} is not releasable.")
```

ibis/backends/polars/__init__.py
```diff
@@ -350,3 +352,15 @@ def to_pyarrow_batches(
...
+        persisted_table_name = util.generate_unique_table_name("cache")
+        lf = self.compile(expr).cache()
+        self.load_data(persisted_table_name, lf)
+        return self.table(persisted_table_name, expr.schema)
+
+    def _release_cache(self, expr):
+        if isinstance(expr._arg, DatabaseTable):
+            del self._tables[expr._arg.name]
+        else:
+            raise NotImplementedError(f"{expr._arg} is not releaseable.")
```
gforsyth, line 366:
```suggestion
            raise NotImplementedError(f"{expr._arg} is not releasable.")
```

ibis/expr/types/relations.py
```diff
@@ -2370,6 +2369,79 @@ def to_pandas(self, **kwargs) -> pd.DataFrame:
...
+
+         All subsequent operations on the returned expression will be performed on the cached data.
+         You may use the ``with`` statement to explicitly clean up the cached expression when it's not needed.
+
+        !!! note "This method eagerly evaluates the expression prior to caching"
+
+        Returns
+        -------
+        Table
+            Cached table.
```
gforsyth, line 2383:
```suggestion
            Cached table
```

gforsyth (commented):
Looking good @thegreymatter !  This is all docstring nits.

ibis/expr/types/relations.py
```diff
@@ -2370,6 +2369,79 @@ def to_pandas(self, **kwargs) -> pd.DataFrame:
...
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.examples.penguins.fetch(table_name="penguins")
+        >>> cached_penguins = t.cache()
+        >>> cached_penguins
```
gforsyth, line 2391: I wonder if turning off interactive mode would offer a more illustrative example, since the repr of the cached table would have the word "cache" in the table name.

As this is written right now, it's more of an example of how to fetch an example dataset.

thegreymatter: @gforsyth, I can do show_sql() instead and show how the select is from the cached table. 
but it is very specific to a DB based backends.

what I think is better is -  `t.mutate(calculation="very slow calculation ;-)").cache()` to show how potentially complex calculation can be cached.

WDYT?

gforsyth: The `mutate` example sounds good to me!

PR #5652, opened 2023-03-02
tests(pyspark): xfail pyspark on python311

cpcloud:
This PR adds xfails and CI back for Python 3.11 on PySpark so that when pyspark 3.4.0 is released we will know. This also lets devs run tests locally without having to attach `not pyspark` to every backend test suite invocation that might pyspark tests.


gforsyth (approved):
looks good!

PR #5653, opened 2023-03-02
refactor(ux): return expression from `Table.info`

cpcloud:
This PR changes the `Table.info` method to be just a regular old expression like everything else.


cpcloud:
I'm not sure if we want this change, but it seems more useful to bake `info` into the rest of ibis than to have it print to a buffer.

The downside is that druid loses the ability to express this because it has very poor support for unions.

cpcloud:
> I like the lazy evaluation approach better.

Nice!

Can we merge it?

PR #5656, opened 2023-03-02
test(pyspark): allow doctests to pass

cpcloud:
This PR fixes an issue where pyspark doctests were marked as xfail but do not actually fail.


cpcloud:
Since spark isn't running on the macos nix jobs, I'm merging this!

PR #5660, opened 2023-03-03
chore(dev-deps): run ruff as a local hook to avoid dual updating chores

cpcloud:
This PR prevents us from having to update ruff in multiple places. We already have a development dependency on it so there is no need to have pre-commit install it in the environment, possibly with a different version that we have to keep in sync in two places.


cpcloud:
This [passed the `lint` stage](https://github.com/ibis-project/ibis/actions/runs/4323988506/jobs/7548243355), merging.

PR #5663, opened 2023-03-03
feat(table): `pivot_longer` API

cpcloud:
This is a PR to discuss a [`pivot_longer`](https://tidyr.tidyverse.org/reference/pivot_longer.html)-style API for table expressions. Ideally this can go into 5.0, but the main purpose is to generate discussion around working code. Depends on #5662.

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -12,6 +12,7 @@ from pytest import param
 import ibis
 import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
+import ibis.expr.selectors as s
 from ibis import _
 from ibis import literal as L
 
@@ -937,3 +938,37 @@ def test_many_subqueries(con, snapshot):
     t3 = query(t2, group_cols=["street"])
 
     snapshot.assert_match(str(ibis.to_sql(t3, dialect=con.name)), "out.sql")
+
+
+@pytest.mark.notimpl(["dask", "pandas"], raises=com.OperationNotDefinedError)
+@pytest.mark.notyet(["polars"], reason="polars doesn't expand > 1 explode")
+@pytest.mark.notimpl(["druid"], raises=AssertionError)
+@pytest.mark.notyet(
+    ["pyspark"], reason="pyspark doesn't allow more than one explode in a select"
+)
+@pytest.mark.notyet(
+    ["bigquery"],
+    reason="backend doesn't implement unnest",
+    raises=com.OperationNotDefinedError,
+)
+@pytest.mark.notyet(
+    ["datafusion", "impala", "mssql", "mysql", "sqlite"],
+    reason="backend doesn't support arrays and we don't implement pivot_longer with unions yet",
+    raises=com.OperationNotDefinedError,
+)
+def test_pivot_longer(con):
+    diamonds = con.tables.diamonds
+    res = diamonds.pivot_longer(s.c("x", "y", "z"), names_to="pos", values_to="xyz")
+    assert res.schema().names == (
+        "carat",
+        "cut",
+        "color",
+        "clarity",
+        "depth",
+        "table",
+        "price",
+        "pos",
+        "xyz",
+    )
+    df = res.limit(5).execute()
+    assert not df.empty
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -9,6 +9,7 @@ import warnings
 from keyword import iskeyword
 from typing import TYPE_CHECKING, Callable, Iterable, Literal, Mapping, Sequence
 
+import toolz
 from public import public
 
 import ibis
@@ -23,6 +24,8 @@ from ibis.expr.types.core import Expr, _FixedTextJupyterMixin
 if TYPE_CHECKING:
     import pandas as pd
 
+    import ibis.expr.schema as sch
+    import ibis.expr.selectors as s
     import ibis.expr.types as ir
     from ibis.common.typing import SupportsSchema
     from ibis.expr.selectors import IfAnyAll, Selector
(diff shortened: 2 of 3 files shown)


thegreymatter:
I love the portable implementation approach, but it raises a questions - 
since many backends support native impl (which might, or might not be more optimised, but probably more readable), is there a way to allow backend specific impl, while keeping the portable impl as a fallback?

I think it is possible if implementing the BaseBackend version with the base expression.

cpcloud:
> I love the portable implementation approach, but it raises a questions - since many backends support native impl (which might, or might not be more optimised, but probably more readable), is there a way to allow backend specific impl, while keeping the portable impl as a fallback?
> 

Totally!

The current way I'm implementing it also very slow for the DuckDB backend in particular. Implementing it by [using `UNNEST`](https://github.com/duckdb/duckdb/pull/6387) makes the generic implementation much much faster.

The main problem with the generic approach using unnest is that many fewer backends support arrays and unnest than support union alls.

> I think it is possible if implementing the BaseBackend version with the base expression.

Yeah, I think for now that's probably a reasonable approach until we have a way to dispatch per-backend at the expression level and not just the operation level.

ibis/expr/types/relations.py
```diff
@@ -2319,6 +2321,366 @@ def to_pandas(self, **kwargs) -> pd.DataFrame:
...
+        │ Afghanistan │ AF     │ AFG    │  1980 │ sp        │      1 │    4 │     ∅ │
+        │ Afghanistan │ AF     │ AFG    │  1980 │ sp        │      1 │    5 │     ∅ │
+        │ Afghanistan │ AF     │ AFG    │  1980 │ sp        │      1 │    6 │     ∅ │
+        │ Afghanistan │ AF     │ AFG    │  1980 │ sp        │      2 │    0 │     ∅ │
+        │ Afghanistan │ AF     │ AFG    │  1980 │ sp        │      2 │    1 │     ∅ │
+        │ Afghanistan │ AF     │ AFG    │  1980 │ sp        │      2 │    2 │     ∅ │
+        │ …           │ …      │ …      │     … │ …         │      … │    … │     … │
+        └─────────────┴────────┴────────┴───────┴───────────┴────────┴──────┴───────┘
+
+        `names_transform` can to be:
```
jcrist, line 2557: typo

```suggestion
        `names_transform` can be:
```

cpcloud: Done.

ibis/expr/types/relations.py
```diff
@@ -2319,6 +2321,366 @@ def to_pandas(self, **kwargs) -> pd.DataFrame:
...
+
+            1. A mapping of one or more names in `names_to` to callable
+            2. A callable that will be applied to every names
+            3. An iterable of callables that will be applied to the values from
+               `names_to` in order. Any `None` values are interpreted as "no
+               transformation".
```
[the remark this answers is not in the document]
cpcloud: Let's keep it as single callable or mapping of name -> callable for now

ibis/expr/types/relations.py
```diff
@@ -2319,6 +2321,366 @@ def to_pandas(self, **kwargs) -> pd.DataFrame:
...
+        │ Afghanistan │ AF     │ AFG    │  1980 │ sp        │      2 │    0 │     ∅ │
+        │ Afghanistan │ AF     │ AFG    │  1980 │ sp        │      2 │    1 │     ∅ │
+        │ Afghanistan │ AF     │ AFG    │  1980 │ sp        │      2 │    2 │     ∅ │
+        │ …           │ …      │ …      │     … │ …         │      … │    … │     … │
+        └─────────────┴────────┴────────┴───────┴───────────┴────────┴──────┴───────┘
+
+        `names_transform` can to be:
+
+            1. A mapping of one or more names in `names_to` to callable
+            2. A callable that will be applied to every names
```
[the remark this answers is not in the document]
cpcloud: Move the description up and deleted the example.

ibis/backends/tests/test_generic.py
```diff
@@ -937,3 +938,37 @@ def query(t, group_cols):
...
+    raises=com.OperationNotDefinedError,
+)
+@pytest.mark.notyet(
+    ["datafusion", "impala", "mssql", "mysql", "sqlite"],
+    reason="backend doesn't support arrays and we don't implement pivot_longer with unions yet",
+    raises=com.OperationNotDefinedError,
+)
+def test_pivot_longer(con):
+    diamonds = con.tables.diamonds
+    res = diamonds.pivot_longer(s.c("x", "y", "z"), names_to="pos", values_to="xyz")
```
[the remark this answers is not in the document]
cpcloud: Will do!

ibis/expr/types/relations.py
```diff
@@ -2481,6 +2484,314 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            transform a column name to a value.
+        values_to
+            Name of the pivoted value column.
+        values_transform
+            Apply a function to the value column. This can be a lambda or
+            deferred expression.
+
+        Returns
+        -------
+            Pivoted table
```
jcrist, line 2522: Should this be indented like this? I'm used to the numpydoc style, not sure what our mkdocs api docs formats nicely.

ibis/expr/types/relations.py
```diff
@@ -2481,6 +2484,314 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        ┃ country ┃ indicator   ┃ year   ┃ value   ┃
+        ┡━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━┩
+        │ string  │ string      │ string │ float64 │
+        ├─────────┼─────────────┼────────┼─────────┤
+        │ ABW     │ SP.URB.TOTL │ 2000   │ 42444.0 │
+        │ ABW     │ SP.URB.TOTL │ 2001   │ 43048.0 │
+        │ ABW     │ SP.URB.TOTL │ 2002   │ 43670.0 │
+        │ ABW     │ SP.URB.TOTL │ 2003   │ 44246.0 │
+        │ ABW     │ SP.URB.TOTL │ 2004   │ 44669.0 │
+        └─────────┴─────────────┴────────┴─────────┘
```
jcrist, line 2593: Would you mind adding a sentence or two describing what's done in the two examples above. This is both to provide more context into what `pivot_longer` is doing, and to also add a break between the two examples above, which kinda blend together without a sentence to break them apart.

jcrist (approved):
One question, and one small nit, otherwise this looks good to me!

ibis/expr/types/relations.py
cpcloud: Ha! This is a rebase error on my part. Thanks for catching.

ibis/expr/types/relations.py
cpcloud: Done, filled out the rest of the examples with more prose.

ibis/expr/types/relations.py
cpcloud: Fixed!

PR #5668, opened 2023-03-04
feat(examples): enable progressbar and faster hashing

cpcloud:
This PR enables the progress bar for pooch downloads, which is a nice UX improvement for end users. It also enables the xxhash extra which is a non-cryptographic and faster hashing algorithm used to hash pooch data when generating the registry.


kszucs (approved):

syvlabs:
@cpcloud @kszucs is there any way to make the xxHash dependency optional? It is currently the only blocker to running this library in [Pyodide](https://pyodide.org/en/stable/) (i.e. running Ibis completely inside the browser, through wasm). Running Ibis in-browser would greatly expand the potential of the library 🙂

cpcloud:
Yes, we can remove that as an optional dependency. It's not clearly adding a lot of value.

cpcloud:
@syvlabs Out of curiosity is `pyarrow` available with Pyodide? We are taking on pyarrow as a required dependency in the next major release (6.0.0) of ibis to simplify a bunch of data conversion internals as well as make output types much more consistent across backends.

syvlabs:
@cpcloud not yet -- see https://pyodide.org/en/latest/usage/packages-in-pyodide.html

syvlabs:
In that case maybe it doesnt make sense yet to run ibis in-browser -- the pyarrow integration seems like a good workaround though!

cpcloud:
We are happy to help in whatever way we can though!

Please keep us updated, and we'll keep an eye on that package list 😃

PR #5672, opened 2023-03-06
ci: remove unnecessary Python version in custom venv cache key

cpcloud:
This PR removes an unnecessary cache key element from the key used to refer to a virtualenv cache in CI. Previously we did this because the action did not store the patch version of Python, resulting in a broken environment.. The action is now encoding this information so we no longer need to do it ourselves.


cpcloud:
Looks like the pip cache is still ignorant of the python version. I'll revert that part.

PR #5675, opened 2023-03-06
fix(bigquery): regenerate sql for predicate pushdown fix

cpcloud:
Quick follow up to the recent predicate pushdown fix to regenerate some test SQL for the BigQuery backend.


cpcloud:
Merging when lints pass.

PR #5690, opened 2023-03-07
fix(ir): use static connection pooling to prevent dropping temporary state

cpcloud:
This PR changes all SQLAlchemy engines to use the [`StaticPool`](https://docs.sqlalchemy.org/en/20/core/pooling.html#sqlalchemy.pool.StaticPool) connection pool.

`StaticPool` uses one connection for all operations.

The reason this is desirable--and necessary if we ever want to have predictably
accessible `TEMP` tables--is that we don't want a new connection created from the connection pool
to appear to discard the state associated with a user's session. It doesn't actually do that, but the new connection lacks any session-specific state from the previous one.

It's not clear to me what the downsides are here except perhaps that a use case
where ibis is part of a server application that fields a large number of
concurrent requests can be bottlenecked by any long running queries.

### DIFF
--- a/ibis/backends/druid/__init__.py
+++ b/ibis/backends/druid/__init__.py
@@ -43,7 +43,7 @@ class Backend(BaseAlchemyBackend):
 
         self.database_name = "default"  # not sure what should go here
 
-        engine = sa.create_engine(url)
+        engine = sa.create_engine(url, poolclass=sa.pool.StaticPool)
 
         super().do_connect(engine)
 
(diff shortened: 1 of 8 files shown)


jcrist (approved):
Change makes sense to me. Are changes needed at all for the not-sqlalchemy-but-still-sql backends as well?

cpcloud:
> Change makes sense to me. Are changes needed at all for the not-sqlalchemy-but-still-sql backends as well?

Good catch. Impala needs to be adjusted since it's using SQLAlchemy's pool implementation. Neither ClickHouse nor BigQuery use connection pooling so those can be left as is.

I will also run the cloud backend tests and post here to cover snowflake.

cpcloud:
Ok, since changing the impala backend to use `StaticPool` seems to break nearly everything about it, I'm going to leave it as is.

PR #5694, opened 2023-03-07
fix(snowflake): bring back `cache` and always quote column names in expressions

cpcloud:
This PR reverts the revert of the `cache()` PR and adds a second commit that fixes the snowflake backend implementation.

Issue #5691, opened 2023-03-07
feat: consider context manager API for temp tables


NickCrews:
### Is your feature request related to a problem?

During my processing I sometimes create intermediate tables as scratch space, to speed up subsequent queries on them. But I don't want these tables to persist after I'm done with this processing.

Another use case is I have a lookup table, eg from state name to state code. I want that in the database so I can JOIN on it to do some translation. But I don't want that lookup table in there permanently, cluttering up the DB.

Using the new temp keyword for creating tables works for dropping tables at the end of connection, but I want to drop the tables sooner than that.

### Describe the solution you'd like

I haven't totally thought this through, but perhaps a context manager API would work here?

```python
lookup = conn.read_csv("state_codes.csv")
with conn.create_table("lookup", lookup) as lu:
    return data.left_join(lu, "state_name")
```

Perhaps the Table class would implement this as

```python
def __enter__(self):
    return self

def __exit__(self, exc_type, exc_value, exc_tb):
    try:
        name = self.get_name()
    except AttributeError:
        # this isn't a named table in the DB
        return
    self._find_backend().drop_table(name)
    # or .drop_view() if this is a view
```

I'm not sure if we want to reserve the context manager API for some other purpose, or if this use case warrants taking it up.

### What version of ibis are you running?

target 5.0?

### What backend(s) are you using, if any?

I think this would be implemented in the Table class and be agnostic to the backend.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

NickCrews:
Oh, neat, I saw the start of [that PR](https://github.com/ibis-project/ibis/pull/5646/files) but didn't see the final result.

It almost works I think, but this bit of code for the backend:

```python
    def _release_cache(self, expr):
        if isinstance(expr._arg, AlchemyTable):
            self.drop_table(expr._arg.name)
        else:
            raise NotImplementedError(f"{expr.arg} is not releaseable.")
```

might be too strict? I'm not exactly sure, I would need to test, but I think it should just no-op instead of raising if the expression isn't releaseable

And I think it needs to work with views, unless https://github.com/ibis-project/ibis/issues/5469 is implemented.

cc @thegreymatter, can we reframe your PR so __exit__ works for any Table, not just ones created with cache() ?

cpcloud:
My hunch is that a context manager that can drop any table is production nightmare waiting to happen.

[...]

PR #5699, opened 2023-03-08
feat(api): make `cache` evaluate only once per session per expression

cpcloud:
This PR makes calls to the `cache` method idempotent for the same expression on
the same backend.

This allows method chaining to incur the cost of caching only once, in keeping
with the rest of the ibis API.

The implementation uses `bidict`, with the cached table source expression as
the key and the temp table (or its equivalent) as the value. The reason a
bidict was desirable was to avoid the need to maintain two dictionaries:

* One for the mapping from input expression to temp table (for lookups of already-cached expressions)
* One for the mapping from cached expression to input expression (for cache invalidation and temp table cleanup)o

I also took the liberty of isolating the context manager behavior of table
expression to a new `CachedTable(Table)` subclass, to avoid accidental dropping
of arbitrary base tables as well as to avoid having to check whether a table is
allowed to be dropped: `CachedTable` is guaranteed by construction to contain a
temp table from the result of caching an expression.

Depends on #5694.


cpcloud:
> @cpcloud this is an excellent optimization. just to make sure I understand the change correctly -
> 
> in this scenario, the cache will only happen once because the expression points to the same cache.
> 
> ```
> b = a.cache().mutate(...)
> c = a.cache().mutate(...)
> ```
> 
> A use case where this behavior might be confusing is when the data changes between the first and second call (i.e. 50,000 new lines got inserted to the a table). and now my only way to refresh the cache would be to actively release the older cache.

Yeah, that's what you'd have to do for the time being.

We could try to use materialized views for some of the backends to solve this problem, but they're not supported across most of our backends.

I think the answer for now at least is:

1. If you know updates might be coming in, using the context manager
2. Otherwise, the contract is that the cached table persists for the session lifetime.

cpcloud:
@thegreymatter I added a `release` method to the `CachedTable` class here, so that if you need to manually invalidate the cache it's easy to do so.

NickCrews:
Could CachedTable be made more generic so it could be used for #5691? 

eg rename it to DroppableTable (that's a bad name, but something more generic) and then you could call .droppable() on a Table and get a DroppableTable as a result that you could still use in a context manager. 

Also, I must be missing something, but couldn't we implement this using the `create_table(temp=True)` API? Is the point that in the future, .cached() could be lazy, so we want to name it something else, whereas create_table() is always eager?

cpcloud:
> Could CachedTable be made more generic so it could be used for #5691?
> 
> eg rename it to DroppableTable (that's a bad name, but something more generic) and then you could call .droppable() on a Table and get a DroppableTable as a result that you could still use in a context manager.

I'm still not convinced that a context manager that drops arbitrary tables enables a use case that can't be done with `cache()`. 😄 

Do you have an example that might help me understand why you want this?

Even so, I'd rather tackle this use case in both a separate PR and likely with a separate implementation from `CachedTable`. If there's some generic code they share then we can consider refactoring it then.

> 
> Also, I must be missing something, but couldn't we implement this using the `create_table(temp=True)` API? Is the point that in the future, .cached() could be lazy, so we want to name it something else, whereas create_table() is always eager?

We're using `create_table` to implement this API. Is there something else you had in mind? `create_table` is tied to a backend, so it doesn't compose with expressions whereas `cache` does.

NickCrews:
> I'm still not convinced that a context manager that drops arbitrary tables enables a use case that can't be done with cache(). 😄

You may be right. Let me actually use the new API and see if I am finding a limitation. I haven't used it so I'm not sure I understand it.

> We're using create_table to implement this API

Err, you're right, sorry I'm not totally grokking this so I'm not phrasing things well.

I think I mean, what if we made it so basically we mirrored the .create_table() and .create_view() APIs of a backend for a Table, where the Table just dispatches to its backend:

```python
# in TableExpr
def create_table(
    self,
    name: str,
    temp: bool = True,
    force: bool = True,
) -> <instance of same type of self, not sure the type hint>:
        current_backend = self._find_backend(use_default=True)
        return current_backend.create_table(self, temp=temp, force=force)
```

The the API is very familiar to users of Backend.create_table(), it's just some sugar. The change here is that we are being more explicit in how this caching is taking place, and we are forcing the user to specify a name (although I guess we could make that arg optional and generate a name...).

ibis/backends/base/__init__.py
```diff
@@ -456,6 +459,12 @@ class BaseBackend(abc.ABC, _FileIOHandler):
     def __init__(self, *args, **kwargs):
         self._con_args: tuple[Any] = args
         self._con_kwargs: dict[str, Any] = kwargs
+        # expression cache
+        self._cache: MutableBidirectionalMapping[
```
kszucs, line 463: Perhaps call it `_result_cache` or `_query_cache`

ibis/backends/base/__init__.py
```diff
@@ -903,22 +914,40 @@ def _cache(self, expr):
...
-            f"{self.name} backend has not implemented `cache` API"
-        )
-
-    def _release_cache(self, expr):
+        op = expr.op()
+        if (result := self._cache.get(op)) is None:
+            name = util.generate_unique_table_name("cache")
+            self._load_into_cache(name, expr)
+            self._cache[op] = result = self.table(name, expr.schema()).op()
+        self._refs[op] += 1
```
kszucs, line 922: Could we factor this out into a refcounted cache container under `ibis.common.caching` (e.g. using callbacks)?

I'd prefer to see this as a standalone utility because:
1. It's cleaner to test the refcounting, unique-key generation, and cache operations
2. It would be useful for other purposes, e.g. as a result cache for the pandas backend
3. Harder to make mistakes if we refactor the backend class

cpcloud: Roger. I've started this in a branch called [`ref-counted-cache`](https://github.com/cpcloud/ibis/tree/ref-counted-cache) that I will PR after this one lands!

ibis/backends/base/__init__.py
```diff
@@ -903,22 +914,40 @@ def _cache(self, expr):
...
-        )
+        op = expr.op()
+        # we need to remove the expression representing the temp table as well
+        # as the expression that was used to create the temp table
+        #
+        # bidict automatically handles this for us; without it we'd have to
+        # do to the bookkeeping ourselves with two dicts
+        if (key := self._cache.inverse.get(op)) is None:
+            raise exc.IbisError(
+                "This expression has already be released. Did you call "
```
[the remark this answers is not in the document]
cpcloud: Fixed.

ibis/backends/base/__init__.py
cpcloud: Called it `_query_cache`.

PR #5701, opened 2023-03-08
docs: add druid to the backends index docs page

cpcloud:
This PR adds the druid backend to the backends index page, and also sorts the backends in the index by name.


cpcloud:
Not sure why those builds are waiting, they should fall under the "hack" required to skip required builds for things that don't actually require it.

Merging.

Issue #5693, opened 2023-03-07
feat: `.create_table()` and `.create_view()` should return the new table to allow chaining


NickCrews:
### Is your feature request related to a problem?

Currently I have to do
```python
conn.create_table("persisted", temp)
persisted = conn.table("persisted")
# continue working with the persisted table
```

When I would like to just do

```python
persisted = conn.create_table("persisted", temp)
# continue working with the persisted table
```

### Describe the solution you'd like

Make them return the new table.

I was trying to think if this promotes bad habits, and we should keep this returning None like how `list.sort()` returns None so you KNOW the sort happened in-place, but I don't think that applies here.

### What version of ibis are you running?

Target 5.0

### What backend(s) are you using, if any?

All

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

NickCrews:
On it. PR sometime today.

PR #5702, opened 2023-03-08
feat: return Table from `create_table()` and `create_view()`

NickCrews:
Closes https://github.com/ibis-project/ibis/issues/5693

I didn't add any tests sepcific to this,
because I modified some existing tests so that
they relied on this new behavior, so this new
behavior is getting tested as a side benefit.
The tests are also clearer now because of this.
I did NOT adjust some tests that I thought would
more clear to have two separate steps of
creating the table/view and accessing it later.

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -174,23 +174,32 @@ class BaseAlchemyBackend(BaseSQLBackend):
         database: str | None = None,
         force: bool = False,
         temp: bool = False,
-    ) -> None:
+    ) -> ir.Table:
         """Create a table.
 
         Parameters
         ----------
         name
-            Table name to create
+            Name of the new table.
         expr
-            DataFrame or table expression to use as the data source
+            An Ibis table expression or pandas table that will be used to
+            extract the schema and the data of the new table. If not provided,
+            `schema` must be given.
         schema
-            An ibis schema
+            The schema for the new table. Only one of `schema` or `obj` can be
+            provided.
         database
-            A database
+            Name of the database where the table will be created, if not the
+            default.
         force
             Check whether a table exists before creating it
         temp
             Should the table be temporary for the session.
+
+        Returns
+        -------
+        Table
+            The table that was created.
         """
         import pandas as pd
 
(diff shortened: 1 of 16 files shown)


jcrist (approved):
Thanks @NickCrews! Overall this looks good to me, but I'd like a second set of :eyes: on it before merging (cc @cpcloud).

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -174,23 +174,32 @@ def create_table(
...
         """Create a table.
 
         Parameters
         ----------
         name
-            Table name to create
+            Name of the new table.
         expr
-            DataFrame or table expression to use as the data source
+            An Ibis table expression or pandas table that will be used to
```
NickCrews, line 185: copied from docstring of super()

ibis/backends/bigquery/__init__.py
```diff
@@ -460,6 +460,7 @@ def create_table(
             sql_select = self.compile(table)
             table_ref = f"`{project_id}`.`{dataset}`.`{name}`"
             self.raw_sql(f'CREATE TABLE {table_ref} AS ({sql_select})')
+        return self.table(name, database=database)
```
NickCrews, line 463: codecov says this isn't covered, but this is unrelated to this change. Open an issue to add a test for create_table() on bigquery?

ibis/backends/dask/__init__.py
```diff
@@ -128,15 +128,6 @@ def _convert_schema(schema: sch.Schema):
     def _convert_object(cls, obj: dd.DataFrame) -> dd.DataFrame:
         return obj
 
-    def create_table(
-        self,
-        table_name: str,
-        obj: dd.DataFrame | None = None,
```
NickCrews, line 134: Actually, I guess overriding this method made it so we could have different type annotations from super(). Should this not get removed so we keep the custom annotations?

ibis/backends/bigquery/__init__.py
cpcloud: It's reporting no coverage because we don't run the bigquery tests until after merge, because we don't want PRs to have access to creds.

cpcloud: No need for an issue right now.

PR #5704, opened 2023-03-08
chore: remove unused _create_table_or_temp_view_from_csv()

NickCrews:
A grep for this shows it only appears in these
two places, and it's a private method so
it shouldn't be used by anyone.


NickCrews:
I ran into this in #5702 when I accidentally modified this test, so then I started investigating.

jcrist (approved):
:+1: nice find

cpcloud (approved):

NickCrews:
That failing test looks unrelated?

cpcloud:
It does. Lemme kick it and see if it's reproducible.

PR #5711, opened 2023-03-09
fix(cache): do not pass invalid arguments to table method

cpcloud:
Fixes an invalid call to self.table when looking up a cached expression.


kszucs:
Tests were not catching this?

cpcloud:
I believe they only trigger a warning which we I think we are ignoring. I'll add a test once I'm in front of a keyboard.

kszucs (approved):

PR #5716, opened 2023-03-09
fix: support negative string indices

### LINKED ISSUE #5037: bug: Negative string index gives incorrect results for some SQL backends
### What happened?

Depending on the backend used, a negative string index can give incorrect results. Things are usually incorrect in one of two ways:
- Off-by-one (sqlite, duckdb, probably others)
- Empty string (postgres, probably others)

```python
In [1]: import ibis, pandas as pd

In [2]: df = pd.DataFrame({"x": ["abcd", "xy"]})

In [3]: sqlite = ibis.sqlite.connect()

In [4]: sqlite.create_table("test", df)

In [5]: sqlite.tables.test.x[-2].execute()  # wrong, off-by-one
Out[5]: 
0    d
1    y
Name: Substring(x, -2, 1), dtype: object

In [6]: postgres = ibis.connect("postgres://postgres:postgres@localhost:5432")

In [7]: postgres.create_table("test", df)

In [8]: postgres.tables.test.x[-2].execute()
Out[8]: 
0    
1    
Name: Substring(x, -2, 1), dtype: object

In [9]: ibis.show_sql(postgres.tables.test.x[-2])
SELECT
  SUBSTR(t0.x, -2 + 1, 1) AS "Substring(x, -2, 1)"
FROM test AS t0
```

This has to do with how `Substring` is implemented - for most SQL backends we translate `string_col[index]` as:

```
SUBSTR(column, index + 1, 1)
```

which is correct for nonnegative indices, but incorrect for negative ones. Duckdb and SQLite support negative indices for `SUBSTR`, while postgres silently ignores negative indices.

A few options:
- We could outlaw negative indices. Negative slices currently error. This wouldn't work if the `index` is an expression though.
- We could fix this for literal values only. This is easy to do for all backends I've surveyed, but wouldn't work if `index` is an expression.
- We could implement this at the SQL level where the proper `SUBSTR`/`RIGHT` calls are made dynamically based on the value of `index` (fully supporting expressions for `index`).

dev

_No response_

### Relevant log output

_No response_

### Code of Conduct

jcrist:
This fixes `substr` to support negative string indices properly (e.g. `t.string_col[-2]`). Fixes #5037.

The current implementation includes a branch for handling the case where the index is a literal (meaning we can avoid using a case statement to branch at runtime), and a fallback for using a case statement to branch at runtime. I've tested that this logic is valid (and included a kinda weird test for it here). However, we currently have expr-level checks that basically mandate that string indexes are literal ints only (can't do `t.string_col[t.int_col]`), so this extra complication is unlikely to be hit by users as is. I haven't taken steps to relax these checks here, but I suppose that could also be done.

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -6,7 +6,7 @@ import operator
 from typing import Any
 
 import sqlalchemy as sa
-from sqlalchemy.sql.functions import FunctionElement
+from sqlalchemy.sql.functions import FunctionElement, GenericFunction
 
 import ibis.common.exceptions as com
 import ibis.expr.analysis as an
@@ -14,6 +14,13 @@ import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
 from ibis.backends.base.sql.alchemy.database import AlchemyTable
+
+
+class substr(GenericFunction):
+    """A generic substr function, so dialects can customize compilation."""
+
+    type = sa.types.String()
+    inherit_cache = True
 
 
 def variance_reduction(func_name):
@@ -418,12 +425,21 @@ def _zero_if_null(t, op):
 
 
 def _substring(t, op):
-    args = t.translate(op.arg), t.translate(op.start) + 1
-
-    if (length := op.length) is not None:
-        args += (t.translate(length),)
-
-    return sa.func.substr(*args)
+    sa_arg = t.translate(op.arg)
+    sa_start = t.translate(op.start) + 1
+    # Start is an expression, need a runtime branch
+    sa_arg_length = t.translate(ops.StringLength(op.arg))
+    if op.length is None:
+        return sa.case(
+            ((sa_start >= 1), sa.func.substr(sa_arg, sa_start)),
+            else_=sa.func.substr(sa_arg, sa_start + sa_arg_length),
+        )
+    else:
+        sa_length = t.translate(op.length)
+        return sa.case(
+            ((sa_start >= 1), sa.func.substr(sa_arg, sa_start, sa_length)),
+            else_=sa.func.substr(sa_arg, sa_start + sa_arg_length, sa_length),
+        )
 
 
 def _gen_string_find(func):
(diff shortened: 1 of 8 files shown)


ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -418,12 +418,32 @@ def _zero_if_null(t, op):
...
-    if (length := op.length) is not None:
-        args += (t.translate(length),)
-
-    return sa.func.substr(*args)
+    sa_arg = t.translate(op.arg)
+    sa_start = t.translate(op.start) + 1
+    if isinstance(op.start, ops.Literal):
+        # Start is a literal, no need for a runtime branch
+        if op.start.value < 0:
+            sa_arg_length = t.translate(op.arg.to_expr().length().op())
```
cpcloud, line 426:
```suggestion
            sa_arg_length = t.translate(ops.StringLength(op.arg))
```

ibis/backends/base/sql/alchemy/registry.py
```diff
@@ -418,12 +418,32 @@ def _zero_if_null(t, op):
...
+            sa_start += sa_arg_length
+        if op.length is None:
+            return sa.func.substr(sa_arg, sa_start)
+        else:
+            sa_length = t.translate(op.length)
+            return sa.func.substr(sa_arg, sa_start, sa_length)
+    else:
+        # Start is an expression, need a runtime branch
+        sa_arg_length = t.translate(op.arg.to_expr().length().op())
```
cpcloud, line 435:
```suggestion
        sa_arg_length = t.translate(ops.StringLength(op.arg))
```

cpcloud (dismissed):
Any chance we can leave the literal optimization out for now?

jcrist:
Ok, I believe this should be good-to-go. The new test fails still on a few backends, but I've marked those failures for now:

- `datafusion` & `pyspark`: didn't feel like looking into this right now
- `druid`: gives an inscrutable error about arguments not being a literal - I'm interpreting this as `substr` on dynamic indices not being supported (but found no docs to confirm this).

I haven't run the tests against cloud backends yet (I currently lack credentials to do so), so there might be failures on `bigquery`/`snowflake`, not sure.

cpcloud (approved):

PR #5722, opened 2023-03-10
feat(api): upcast pandas DataFrames to memtables in `rlz.table` rule

cpcloud:
This PR allows using a DataFrame wherever an operation accepts a table via `ibis.expr.rules.table`.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -687,8 +687,15 @@ class Backend(BaseAlchemyBackend):
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         # in theory we could use pandas dataframes, but when using dataframes
         # with pyarrow datatypes later reads of this data segfault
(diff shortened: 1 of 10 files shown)


kszucs (approved):
Nice!

cpcloud:
There's a segfault in duckdb that appears to be related to arrow/duckdb interaction. It's unrelated to this PR in that the PR isn't causing it, but the problem is reproducible enough that we should avoid merging this PR until we can get some idea about what's happening.

cpcloud:
I'm pretty sure I've worked around the duckdb issue, but there's no 100% solution without a fix upstream.

Filed https://github.com/duckdb/duckdb/issues/6682 to address it.

cpcloud:
I think this feature is valuable enough, and has a low enough blast radius that we can ship this in 5.0.

I will xfail the duckdb test and add some documentation about `null` typed columns.

kszucs (approved):
Nice!

PR #5736, opened 2023-03-13
feat(api): make create_table uniform

### LINKED ISSUE #2339: feat: standardize APIs for creating tables from expressions
A common pattern for me is to build up a SQL expression, execute it against the database backend and save the results as a database table.  I think it's not currently possible to do this entirely in Ibis?  What I generally do is compile the expression to SQL and then execute that using either the BigQuery web interface or the Python client, both of which have options to save the query as a table.  It would be really nice if I could do all of this without leaving Ibis.

Some other things I often do:
- List datasets (Ibis does this)
- List tables (Ibis does this)
- Create and delete tables and datasets 
- Execute queries and save the results directly as files on GCS or local disk
- Copy tables to local disk or Google Cloud Storage, e.g. as CSV, Avro, etc
- Upload CSV files or dataframes to database tables 

I realize this stuff is potentially out of scope, but just thought I'd ask.  So far I end up writing a wrapper library around Ibis to do all of this but it strikes me that it could fit really nicely in the main Ibis project.   I think some of the Blaze project was focused on use cases like this and I think maybe there's a gap here.

### LINKED ISSUE #4316: feat(clickhouse): add implementations of `create_table`, `create_view`, and `create_database`
All of these are accomplishable through `raw_sql` method but it would be nice for these to be fully implemented. In addition, it would also be useful to have the inverse methods implemented: `drop_table`, `drop_view`, and `drop_database`.

# Proposal

From my experience with Clickhouse, and some experience with implementing these for my own use cases, I've outlined a brief propsal for implementing these methods. Feedback would be greatly appreciated.

## Create Table

This one would be the most complicated to implement. There would be two major use cases:

1. Standard table creation (from `Schema` object)
2. Create table from `Expr` object

Implementation for the first use case would also be used in the second.

For the first use case, a `Schema` object would be passed in. You would be able to grab the name and type for each field of the table from that object.

For the second use case, an `Expr` object would be passed in. You would be able to grab the name and type of each field from the `Expr` schema attribute using the same code as the previous use case.

One thing to note for creating a table from an expression: in my experience it is better to call the `CREATE TABLE...` DDL statement and then separately call  the `INSERT ...` statement instead of `CREATE TABLE ... AS ...`.

You would also need to check to see if the table already exists. If it exists, there are a few ways to replace it. You could create a temp table and once that is complete drop the target table and rename the temp table. This is how `dbt-clickhouse` handles it but I've found this can present challenges with the `ReplicatedMergeTree` table engine. Another method is to drop the target table and then run the DDL and insertion statements. The downside to this is that there is a longer period of time where not table with the identifier will be available for users to query.

Backing up a bit, there is also the question of whether a full replace is desirable to the user. Should the user have an option to abandon the process if the table already exists?

There are also a number of additional options that would need to be available to the user when creating a table:

| Option | Description | Argument Type |
|---|---|---|
| Database | The database to create the table in. If not provided, then the session database passed in to `connect()` should be used | `str` |
| Table name | Since not all `Expr` child objects have a name attribute, this would need to be provided by the user | `str` |
| Engine | Clickhouse [table engine](https://clickhouse.com/docs/en/engines/table-engines/) | `str` |
| On Cluster | Specify a cluster | `str` |
| Partition By | Values to partition the table by. Should be a tuple of strings. Could potentially use column objects but that might be more difficult for defining partition keys wrapped in a method (i.e. `toYYYYMM(some_timestamp)`) | `tuple[str]` |
| Order By | Sorting keys | `tuple[str]` |
| Primary Key | Indexes | `tuple[str]` |
| Settings | Key-value pairs of additional table settings | `dict[str, Any]` |

## Create View

This one would be much easier to implement and Clickhouse provides the standard `CREATE OR REPLACE VIEW ...` option. This would just be created from an `Expr` object.

Only the following options would need to be provided to the user:

| Option | Description | Argument Type |
|---|---|---|
| Database | The database to create the table in. If not provided, then the session database passed in to `connect()` should be used | `str` |
| View name | Since not all `Expr` child objects have a name attribute, this would need to be provided by the user | `str` |
| On Cluster | Specify a cluster | `str` |

## Create Database

This one is likely the lowest priority out of the three. It would also likely be the easiest to implement. The user would need to be presented the following options:

| Option | Description | Argument Type |
|---|---|---|
| Database Name | Database name | `str` |
| On Cluster | Specify a cluster | `str` |
| Engine | Clickhouse [database engine](https://clickhouse.com/docs/en/sql-reference/statements/create/database/#engine) | `str` |

### `IF NOT EXISTS` Behavior

It would likely be beneficial to also allow some way to only create the object if it doesn't already exist. 

[TODO]

## Notes

I'll keep working on this proposal and flesh out some ideas in a project I already have going. I'm pretty sure I've got the codegen down for the create table piece but I'll need to do some more digging to see how it is already handled in Ibis so I'm following the current approaches.

I'm happy to be the one to implement this over a weekend or two but I'd like some feedback before I dive into the code.

### LINKED ISSUE #5390: Deprecate `Backend.load_data`
`con.load_data` is identical to `con.create_table` with `force=True`. We should:

- Deprecate `load_data` in favor of `create_table`
- Rename `force` to `overwrite` in `create_table` to better match the `overwrite` kwarg in `insert`.

### LINKED ISSUE #5392: Review DDL APIs for uniformity across backends
There are a few existing DDL apis that we think are fine as is, but may not be uniformly supported across all backends. It would be good to review that all backends _uniformly_ support the following operations (where it makes sense), and that generic backend tests exist for the following operations:

- [ ] `drop_table(name, database=None, force=False)`
- [ ] `drop_view(name, database=None, force=False)`
- [ ] `truncate_table(name, database=None)`

### LINKED ISSUE #5394: feat: support `create_table`/`create_view` for pandas/dask/polars
For API uniformity we should support `create_table` and `create_view` for these non-SQL backends. Potential implementations:

- `create_table(name, schema)` - create an empty "dataframe" with the provided schema, and store it in the backend's table mapping
- `create_table(name, obj)` - execute the passed in expression, convert it to the backing "dataframe" type, and store it in the backend's table mapping. For `dask` it would likely make sense to use `.persist()` instead of `.compute()` here so the data stays on the cluster.
- `create_view(name, expr)` - just store the backing expression directly in a view mapping on the backend?

cpcloud:
This PR address recent issues created around DDL.

Notes

- I didn't deprecate `ImpalaTable.load_data`. I'm not entirely what should be done with it.
- For clickhouse's `create_table`, since clickhouse itself doesn't have a default
  table engine, I preserved that: you have to pass `engine="Something"` to its
  create_table. It breaks the uniformity, but I'm not sure how else to handle
  it.

TODO

- Implement `create_view` for pandas/dask/polars
- Implement `create_view`/`drop_table`/`truncate_table` for clickhouse
- Review/implement `truncate_table` (I don't think these makes sense for the in-memory backends, so those'll raise an exception)

Closes #5394
Closes #5392
Closes #5390
Closes #4316
Closes #2339

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING, Any, Iterable, Literal, Mapping
 import sqlalchemy as sa
 
 import ibis
-import ibis.common.exceptions as exc
+import ibis.common.exceptions as com
 import ibis.expr.operations as ops
 import ibis.expr.schema as sch
 import ibis.expr.types as ir
@@ -92,7 +92,6 @@ class BaseAlchemyBackend(BaseSQLBackend):
     def do_connect(self, con: sa.engine.Engine) -> None:
         self.con = con
         self._inspector = sa.inspect(self.con)
-        self.meta = sa.MetaData()
         self._schemas: dict[str, sch.Schema] = {}
         self._temp_views: set[str] = set()
 
@@ -170,11 +169,12 @@ class BaseAlchemyBackend(BaseSQLBackend):
     def create_table(
         self,
         name: str,
-        expr: pd.DataFrame | ir.Table | None = None,
+        obj: pd.DataFrame | ir.Table | None = None,
+        *,
         schema: sch.Schema | None = None,
         database: str | None = None,
-        force: bool = False,
         temp: bool = False,
+        overwrite: bool = False,
     ) -> ir.Table:
         """Create a table.
 
@@ -182,7 +182,7 @@ class BaseAlchemyBackend(BaseSQLBackend):
         ----------
         name
             Name of the new table.
-        expr
+        obj
             An Ibis table expression or pandas table that will be used to
             extract the schema and the data of the new table. If not provided,
             `schema` must be given.
@@ -192,20 +192,23 @@ class BaseAlchemyBackend(BaseSQLBackend):
         database
             Name of the database where the table will be created, if not the
             default.
-        force
-            Check whether a table exists before creating it
         temp
             Should the table be temporary for the session.
+        overwrite
+            Clobber existing data
 
         Returns
         -------
         Table
             The table that was created.
         """
+        if obj is None and schema is None:
+            raise com.IbisError("The schema or obj parameter is required")
+
         import pandas as pd
 
-        if isinstance(expr, pd.DataFrame):
-            expr = ibis.memtable(expr)
+        if isinstance(obj, pd.DataFrame):
+            obj = ibis.memtable(obj)
 
         if database == self.current_database:
             # avoid fully qualified name
@@ -213,20 +216,17 @@ class BaseAlchemyBackend(BaseSQLBackend):
 
         if database is not None:
             raise NotImplementedError(
-                'Creating tables from a different database is not yet implemented'
+                "Creating tables from a different database is not yet implemented"
             )
 
-        if expr is None and schema is None:
-            raise ValueError('You must pass either an expression or a schema')
-
-        if expr is not None and schema is not None:
-            if not expr.schema().equals(ibis.schema(schema)):
-                raise TypeError(
+        if obj is not None and schema is not None:
+            if not obj.schema().equals(ibis.schema(schema)):
+                raise com.IbisTypeError(
                     'Expression schema is not equal to passed schema. '
                     'Try passing the expression without the schema'
                 )
         if schema is None:
-            schema = expr.schema()
+            schema = obj.schema()
 
         self._schemas[self._fully_qualified_name(name, database)] = schema
         table = self._table_from_schema(
@@ -236,15 +236,17 @@ class BaseAlchemyBackend(BaseSQLBackend):
             temp=temp,
         )
 
-        if has_expr := expr is not None:
+        if has_expr := obj is not None:
             # this has to happen outside the `begin` block, so that in-memory
             # tables are visible inside the transaction created by it
-            self._register_in_memory_tables(expr)
+            self._register_in_memory_tables(obj)
 
         with self.begin() as bind:
-            table.create(bind=bind, checkfirst=force)
+            if overwrite:
+                table.drop(bind=bind, checkfirst=True)
+            table.create(bind=bind)
             if has_expr:
-                method = self._get_insert_method(expr)
+                method = self._get_insert_method(obj)
                 bind.execute(method(table.insert()))
         return self.table(name, database=database)
 
@@ -260,8 +262,12 @@ class BaseAlchemyBackend(BaseSQLBackend):
             and isinstance(expr.op(), ops.InMemoryTable)
         ):
             (from_,) = compiled.get_final_froms()
-            (rows,) = from_._data
-            return methodcaller("values", rows)
+            try:
+                (rows,) = from_._data
+            except AttributeError:
+                return methodcaller("from_select", list(expr.columns), from_)
+            else:
+                return methodcaller("values", rows)
 
         return methodcaller("from_select", list(expr.columns), compiled)
 
@@ -278,31 +284,28 @@ class BaseAlchemyBackend(BaseSQLBackend):
         ]
 
     def _table_from_schema(
-        self,
-        name: str,
-        schema: sch.Schema,
-        database: str | None = None,
-        temp: bool = False,
+        self, name: str, schema: sch.Schema, temp: bool = False, **_: Any
     ) -> sa.Table:
         prefixes = []
         if temp:
             prefixes.append('TEMPORARY')
         columns = self._columns_from_schema(name, schema)
         return sa.Table(
-            name, self.meta, *columns, prefixes=prefixes, quote=self.quote_table_names
+            name,
+            sa.MetaData(),
+            *columns,
+            prefixes=prefixes,
+            quote=self.quote_table_names,
         )
 
     def drop_table(
-        self,
-        table_name: str,
-        database: str | None = None,
-        force: bool = False,
+        self, name: str, *, database: str | None = None, force: bool = False
     ) -> None:
         """Drop a table.
 
         Parameters
         ----------
-        table_name
+        name
             Table to drop
         database
             Database to drop table from
@@ -314,26 +317,23 @@ class BaseAlchemyBackend(BaseSQLBackend):
             database = None
 
         if database is not None:
-            raise NotImplementedError(
-                'Dropping tables from a different database is not yet implemented'
+            raise com.IbisInputError(
+                "Dropping tables from a different database is not yet implemented"
             )
 
-        t = self._get_sqla_table(table_name, schema=database, autoload=False)
+        t = self._get_sqla_table(name, schema=database, autoload=False)
         with self.begin() as bind:
             t.drop(bind=bind, checkfirst=force)
 
-        assert not self.inspector.has_table(
-            table_name
-        ), f"Something went wrong during DROP of table {table_name!r}"
-
-        self.meta.remove(t)
-
-        qualified_name = self._fully_qualified_name(table_name, database)
+        qualified_name = self._fully_qualified_name(name, database)
 
         with contextlib.suppress(KeyError):
             # schemas won't be cached if created with raw_sql
             del self._schemas[qualified_name]
 
+    @util.deprecated(
+        as_of="5.0", removed_in="6.0", instead="Use create_table(overwrite=True)"
+    )
     def load_data(
         self,
         table_name: str,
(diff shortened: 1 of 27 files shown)


gforsyth:
Thoughts on forcing keyword arguments for a bunch of these DDL functions?
I was thinking everything after `name` and `obj` should be keyword-only (so `schema`, `database`, `overwrite` and `temp`)

cpcloud:
SGTM

gforsyth (approved):
this looks good to me -- I'll leave it for a bit in case anyone else wants to take a reviewing pass.

cpcloud:
There are some snowflake things that are failing locally for me. #5741 should help with that

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -287,19 +302,16 @@ def _table_from_schema(
         if temp:
             prefixes.append('TEMPORARY')
         columns = self._columns_from_schema(name, schema)
-        return sa.Table(name, self.meta, *columns, prefixes=prefixes)
+        return sa.Table(name, sa.MetaData(), *columns, prefixes=prefixes)
```
jcrist, line 305: Why this change?

ibis/backends/bigquery/__init__.py
```diff
@@ -441,11 +442,22 @@ def create_table(
...
     ) -> ir.Table:
         if obj is None and schema is None:
-            raise ValueError("The schema or obj parameter is required")
+            raise com.IbisError("The schema or obj parameter is required")
+        if temp is True:
+            raise com.IbisInputError(
+                "BigQuery backend does not yet support temporary tables"
+            )
+        if overwrite is not False:
+            raise com.IbisInputError(
```
jcrist, line 458: Should these be `NotImplementedError` instead?

ibis/backends/pandas/__init__.py
```diff
@@ -120,11 +124,30 @@ def compile(self, expr, *args, **kwargs):
...
+                "Passing `temp=True` to the Pandas backend create_table method has no "
+                "effect: all tables are in memory and temporary. Avoid passing the "
+                "`temp` argument."
+            )
+        if database:
+            raise com.IbisInputError(
+                "Passing `database` to the Pandas backend create_table method has no "
+                "effect: Pandas cannot set a database. Avoid passing the `database` "
+                "argument."
+            )
```
jcrist, line 148: To support writing backend-agnostic code, perhaps these should be warnings (or just silently ignored)? That's easy to loosen the restrictions later though, so a decision on this now doesn't need to be a blocker.

jcrist (approved):
Overall looks good to me. A few small questions/comment, but none should be blockers.

ibis/backends/bigquery/__init__.py
gforsyth: probably -- I think i did this to keep our exceptions within the ibis coms, but it's just confusing.  We probably need to revamp all the ibis exceptions anyway.  I'll change these.

ibis/backends/pandas/__init__.py
gforsyth: it's a bit sticky for me.
if someone passes an invalid argument (and they only use a single backend) I'd expect an error, since users ignore warnings (in my experience).
But we don't want to make it harder to write backend-agnostic code, so I guess warnings might be the way to go.

gforsyth: made them warnings.

jcrist: Gah, sorry for the back-and-forth on this here. I guess there's a lot of backends that don't truly support these options (dask, polars, pyspark?, ...). We should either warn uniformly, or just revert back to erroring here for now. I don't feel strongly either way, but it's easier to relax restrictions later than to add them back.

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: I'm not sure this will be necessary after #5741. I'll confirm and report back here once that's merged.

ibis/backends/pandas/__init__.py
gforsyth: we'll need to distinguish eventually between `WillNeverBeImplemented` because, e.g. `temp` doesn't mean anything in pandas-land vs `NotImplemented` because `ibis` doesn't yet support a thing the backend can do.

My inclination is to `warn` if an option is (effectively) meaningless and `error` if it has meaning but isn't yet supported.

So I think that would translate to warning for pandas, dask, polars and maybe pyspark?

jcrist: I think that split makes sense to me.

cpcloud: For PySpark, I think it's probably an error in both database case and temp view case. I think whether a view is permanent depends on whether there's a hive metastore in the mix or not.

I guess I'm +1 on erroring here so that people don't get the wrong idea about backend compatibility with DDL.

warnings indicate a meaningful difference in behavior IMO should be errors

gforsyth: Ok -- how about `com.IbisError` for those backends where `temp` doesn't make sense and `NotImplementedError` otherwise?

cpcloud: SGTM.

jcrist: That makes sense I think. Nice distinction between "will never be supported" and "may be eventually supported".

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: This is necessary to avoid this function looking at existing tables in metadata. This function is just for constructing an unbound `sa.Table` without knowledge of what's already in the database.

ibis/backends/pandas/__init__.py
cpcloud: Great conversation, thanks all!

cpcloud:
Cloud backends:

```
cloud in 🌐 albatross in …/ibis on  create-table-uniformity is 📦 v4.1.0 via 🐍 v3.10.10 via ❄️  impure (ibis-3.10.10-env)
❯ pytest -m 'bigquery or snowflake' -q -n auto --dist loadgroup
bringing up nodes...
xxxxxxx.xx.xx.x..xx...........xx...x.......xx..x.x...............x..x..x....x.x........x......x.x.....x.x..x.. [  5%]
..x................x.x..x.x.x.x..x..........xx.........x......x..x.x...x......x.x.x.x.xx.x...x.xxxx........... [ 10%]
x..x...xx.x.x.x.xx.....x.x..xx.........x.x..xx....xx.......x.....x.....x..x..........x.......x...............x [ 16%]
..x......................................x.................................................................... [ 21%]
.x.xx.............x.....x......x..xx.....xx.....x..........x..x..x........xx..........xx.x..x........x........ [ 27%]
...x...............................................x..............x....................................xxxxxxx [ 32%]
x.xxx.x.xxx.x.xxx.xxxx.xxxxxxxx.xx.xxxxxx..x.xxxx..xx.....x...x........xx......x...x.......x...x......x.x..... [ 38%]
........x...x......x.....x.xx.x.......x.....x.x.x.....x...x.x..x..x........x......x...x..x......x...x..x.s.x.. [ 43%]
...x....s......x................................x.....................x........x.............x........s.....x. [ 48%]
.x........x.........x.....x.x....xx.......x...x......x..xx...x..xxx.x..x..........xx.x.......x..xx..x.x..x.xx. [ 54%]
.x..x..xx.x.x....xx.....xx....x.....x.x............x...................x...x.....x..........x.....x....x...... [ 59%]
..............x...x........x..................x.xx......................................................x.xx.. [ 65%]
....x.x.x...........xx..xxx..x.x..x.x....xx......x........x..........x.....x...............x.x................ [ 70%]
....x....................................x..........................x...xx.xxxxxxxxxx..xx.x..........xx....... [ 76%]
................x...........x..............x.....................x.x.....x...........x.....xx..x.............. [ 81%]
...x......xx....x......x...x........x...s...x........x.x.x............x.....x..xxx.s......x................... [ 86%]
.........x.................................................................................................... [ 92%]
................................x....x...s...........x..................x..................................... [ 97%]
...........................x...............                                                                    [100%]
1688 passed, 6 skipped, 329 xfailed in 273.12s (0:04:33)
```

cpcloud:
I am authoring a longer commit message, hold off on merging please!

cpcloud:
Ok, merging on green!

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: I ended up removing the `meta` instance attribute. It was causing annoying problems with invalidating identifiers from snowflake. I added a bit of prose in the commit message about it.

PR #5738, opened 2023-03-14
chore(dev-deps): set the default nix environment python version to 3.10

cpcloud:
Primarily for the ability to test PySpark integration, which is tied to Python 3.10 until PySpark 3.4.


cpcloud:
Fine to just merge this on approval!

PR #4157, opened 2022-06-29
[WIP] feat: add dask windows

gerrymanoim:
Previously dask supported "noop" windows. This PR adds some limited window functionality. Draft to see how far I can reasonably get while keeping this a reasonable size. 

Notes:
- dask only supports single value sorts, so any window with grouping and sorting is not supported
- I'm trying to keep the structure quite close the pandas backend, which will make some eventual consolidation and refactoring much easier to do. However, the pandas windowing code definitely needs some broad refactoring so things look a bit messy.

PR #5747, opened 2023-03-15
feat: add dask windowing

patcao:
Previously dask supported "noop" windows. This PR adds basic dask window functionality.
Following @gerrymanoim's initial attempt (https://github.com/ibis-project/ibis/pull/4157), I kept the structure quite close the pandas backend, which will make some eventual consolidation and refactoring much easier to do. I also wanted to keep the changes of this PR out of the Pandas backend, so there is duplicated code from Pandas backend.

**Follow Ups:**
Support for groupby + orderby windows.
Support cumulative windows
Cleanup and refactors code that can be shared between Pandas and Dask backends


ibis/backends/dask/aggcontext.py
```diff
@@ -0,0 +1,235 @@
...
+    __slots__ = ()
+
+    def agg(self, grouped_data, function, *args, **kwargs):
+        if isinstance(function, str):
+            return getattr(grouped_data, function)(*args, **kwargs)
+
+        if not callable(function):
+            raise TypeError(f'Object {function} is not callable or a string')
+
+        if isinstance(grouped_data, SeriesGroupBy) and len(grouped_data):
```
gerrymanoim, line 29: Codecov seems to believe that none of these branches get hit with tests. Can you confirm whether that's the case?

ibis/backends/dask/aggcontext.py
```diff
@@ -0,0 +1,235 @@
...
+
+        method = operator.methodcaller('apply', sliced_agg, raw=False)
+
+    result = method(windowed)
+    # No MultiIndex support in dask
+    result.index = frame.index
+    return result
+
+
+class Window(AggregationContext):
```
gerrymanoim, line 99: Most of this is a direct-ish port of what's in the pandas backend?

ibis/backends/dask/execution/selection.py
```diff
@@ -137,10 +136,16 @@ def build_df_from_projection(
...
-    data_pieces = [compute_projection(node, op, data, **kwargs) for node in selections]
+    partitioned_data = add_partitioned_sorted_column(data)
+
+    data_pieces = [
+        compute_projection(node, op, partitioned_data, **kwargs) for node in selections
+    ]
 
-    return dd.concat(data_pieces, axis=1).reset_index(drop=True)
+    result = dd.concat(data_pieces, axis=1)
+    result.index = result.index.rename(None)
```
gerrymanoim, line 146: Why do we need to drop the index name?

ibis/backends/dask/execution/util.py
```diff
@@ -146,8 +143,15 @@ def coerce_to_output(
             divisions=(0, out_len - 1),
         )
         return series
+    elif isinstance(result, dd.DataFrame):
```
gerrymanoim, line 146: I'm confused by the logic here - shouldn't any `dd.DataFrame` case be handled above? https://github.com/ibis-project/ibis/pull/5747/files#diff-7560a298b0242d66bc062059c8a1189235eea9088c8174ce22b7154caf1d629fR124

ibis/backends/dask/execution/util.py
```diff
@@ -146,8 +143,15 @@ def coerce_to_output(
...
+        return result
 
-    return dd.from_pandas(pd_util.coerce_to_output(result, node, index), npartitions=1)
+    try:
```
gerrymanoim, line 149: I'm unsure how we would hit this case? And cov says that tests aren't hitting this path.

ibis/backends/dask/execution/util.py
```diff
@@ -199,48 +203,57 @@ def safe_concat(dfs: list[dd.Series | dd.DataFrame]) -> dd.DataFrame:
...
-    use the dask `execute` method.
-
-    This function borrows the logic in the pandas backend. `by` can be a
-    string or an expression. If `by.get_name()` raises an exception, we must
-    `execute` the expression and sort by the new derived column.
-    """
-    name = ibis.util.guid()
-    if key.name in data:
-        return name, data[key.name]
+def compute_sort_key(key, data, timecontext, scope=None, **kwargs):
```
gerrymanoim, line 206: Typehints please

ibis/backends/dask/execution/util.py
```diff
@@ -199,48 +203,57 @@ def safe_concat(dfs: list[dd.Series | dd.DataFrame]) -> dd.DataFrame:
...
 
-def compute_sorted_frame(
-    df: dd.DataFrame,
-    order_by: ir.Value,
-    timecontext: TimeContext | None = None,
-    **kwargs,
-) -> dd.DataFrame:
-    sort_col_name, temporary_column = compute_sort_key(
-        order_by, df, timecontext, **kwargs
+def compute_sorted_frame(df, order_by, group_by=(), timecontext=None, **kwargs):
```
gerrymanoim, line 223: typehint please

ibis/backends/dask/execution/window.py
```diff
@@ -2,28 +2,163 @@
...
+    order_by,
+    **kwargs,
+) -> NoReturn:
+    raise NotImplementedError(
+        f"get_aggcontext is not implemented for {type(window).__name__}"
+    )
+
+
+@get_aggcontext.register(ops.WindowFrame)
+def get_aggcontext_window(
```
gerrymanoim, line 98: TODO consolidate with pandas

ibis/backends/dask/execution/window.py
```diff
@@ -49,11 +184,43 @@ def _post_process_empty(
         return series[0]
     else:
         # Project any non delayed object to the shape of "parent"
-        return parent.apply(lambda row: result, meta=(None, 'object'))
+        if isinstance(parent, dd.DataFrame):
```
gerrymanoim, line 187: In what case do we need to select the first column out first?

ibis/backends/dask/execution/window.py
```diff
@@ -49,11 +184,43 @@ def _post_process_empty(
...
+    series,
+    parent: dd.DataFrame,
+    order_by: list[str],
+    group_by: list[str],
+    timecontext: TimeContext | None,
+    op,
+    **kwargs,
+) -> dd.Series:
+    assert not order_by and group_by
+    # FIXME This is likely not needed anymore.
```
gerrymanoim, line 218: Should we just not have these?

ibis/backends/dask/execution/window.py
```diff
@@ -62,86 +229,236 @@ def execute_window_frame(
...
+            #     # Dask will raise an exception about not supporting multiple Series in group by key
+            #     # even if it is passed a length 1 list of Series.
+            #     # For this case we just make group_by_cols a single Series.
+            #     group_by_cols = grouping_keys[0]
+            # else:
+            #     group_by_cols = grouping_keys
+            # source = sorted_df.groupby(group_by_cols, sort=True, group_keys=False)
+        else:
+            if len(grouping_keys) == 1 and isinstance(grouping_keys[0], dd.Series):
+                # Dask will raise an exception about not supporting multiple Series in group by key
```
gerrymanoim, line 296: Do you think this is worth opening a bug report for?

ibis/backends/dask/tests/conftest.py
```diff
@@ -14,7 +14,10 @@
 
 dd = pytest.importorskip("dask.dataframe")
 
-NPARTITIONS = 2
+# FIXME Force npartitions=1 because of a dask issue with non deterministic groupby results.
+# The issue relates to the shuffle method on a local cluster, using npartitions=1 in tests avoids it.
+# https://github.com/dask/dask/issues/10034
+NPARTITIONS = 1
```
gerrymanoim, line 20: We need `NPARTITIONS` to be greater than one so as to not mask cross partitioning bugs. 

Can you use the alternative fix suggested for the jira for any tests where we are getting bugs from non determinism before?

ibis/backends/dask/tests/execution/test_window.py
```diff
@@ -0,0 +1,718 @@
...
+    expected = pandas_method(
+        batting_df[['G', 'yearID']]
+        .sort_values('yearID', kind=sort_kind)
+        .G.rolling(len(batting_df), min_periods=1)
+    ).reset_index(drop=True)
+    expected = expected.compute()
+    tm.assert_series_equal(result, expected.rename(f"Cumulative{op.capitalize()}(G)"))
+
+
+# @pytest.mark.xfail(
```
gerrymanoim, line 298: Delete

ibis/backends/dask/execution/selection.py
patcao: The point of calling `reset_index(drop=True)` here is to drop `_ibis_index` and give it an arbitrary new index. The dataframe here is already nicely partitioned on `_ibis_index`.
It would be a waste to drop it, so instead, we just drop the name and use that as the dataframe index.

ibis/backends/tests/test_window.py
```diff
@@ -287,13 +287,16 @@ def test_grouped_bounded_expanding_window(
                         "snowflake",
                         "trino",
                     ]
-                )
+                ),
+                pytest.mark.broken(
+                    ["dask"], reason="FIXME Dask rolling object not subscriptable"
```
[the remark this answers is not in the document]
patcao: No there isn't, Dask complains about non-unique index being used during this test. However, from inspecting the dataframe the index IS unique.
I haven't been able to reproduce this issue in order to create a Dask issue yet.

ibis/backends/dask/aggcontext.py
patcao: Yeah, it's pretty much just a Dask port from the Pandas backend version.
The difference is mostly just needing to make sure the resulting dataframes are partitioned for Dask operations downstream and where the Dask api is slightly different than Pandas api.

ibis/backends/dask/execution/window.py
patcao: Yeah, I mostly left this for when I do cleanup work in here later.
How we handle `_get_post_process_function` method needs to be refactored.

ibis/backends/dask/aggcontext.py
patcao: Yeah, this branch isn't getting hit. Dask aggregations hit slightly different code paths than Pandas. So the tests aren't exercising it in the same way.

ibis/backends/dask/execution/window.py
patcao: Yeah, this can be simply implemented with just passing `axis=1` to the apply.
We just want to broadcast the result value across the whole parent index.

ibis/backends/dask/execution/util.py
```diff
@@ -355,6 +396,52 @@ def helper(
     return df
 
 
+def add_sequential_partitioned_sorted_column(
```
patcao, line 399: Using this to partition dataframes over the previous implementation.
This is definitely less efficient and less safe from overflow, but having an index that's globally sequential is important for working with downstream operations.

ibis/backends/dask/execution/util.py
patcao: yeah, we weren't hitting that branch. removed now.

ibis/backends/dask/aggcontext.py
gerrymanoim: Do we need this code then? Or we will in the future?

ibis/backends/dask/execution/util.py
gerrymanoim: What is `result` in this case? I'm not sure what non pandas/dask types we'd get here.

ibis/backends/dask/execution/util.py
```diff
@@ -218,29 +223,65 @@ def compute_sort_key(
     name = ibis.util.guid()
     if key.name in data:
         return name, data[key.name]
+    if isinstance(key, str):
+        return key, None
+    elif key.name in data:
```
gerrymanoim, line 228: Isn't this already handled in line 224? https://github.com/ibis-project/ibis/pull/5747/files#diff-7560a298b0242d66bc062059c8a1189235eea9088c8174ce22b7154caf1d629fR224

gerrymanoim: I'm not sure how we would ever hit this.

ibis/backends/dask/execution/util.py
```diff
@@ -218,29 +223,65 @@ def compute_sort_key(
...
+        return key.name, None
     else:
         if scope is None:
             scope = Scope()
         scope = scope.merge_scopes(
             Scope({t: data}, timecontext) for t in an.find_immediate_parent_tables(key)
         )
         new_column = execute(key, scope=scope, **kwargs)
+        name = ibis.util.guid()
```
gerrymanoim, line 237: `name`. should already be a guid from line 223? https://github.com/ibis-project/ibis/pull/5747/files#diff-7560a298b0242d66bc062059c8a1189235eea9088c8174ce22b7154caf1d629fR223

ibis/backends/dask/execution/util.py
```diff
@@ -355,6 +396,52 @@ def helper(
...
+
+    if col_name in df.columns:
+        raise ValueError(f"Column {col_name} is already present in DataFrame")
+
+    df = df.reset_index(drop=True)
+    df = df.assign(**{col_name: 1})
+    df = df.assign(**{col_name: df[col_name].cumsum() - 1})
+    df = df.set_index(
+        col_name,
+        sorted=True,
```
gerrymanoim, line 440: If you don't pass a value for `divisions` I believe this triggers dask to compute the whole dataframe up to this point, which we very much want to avoid. I think we'll have to think of something different here.

ibis/backends/dask/execution/util.py
gerrymanoim: Maybe you explain this later, but can you elaborate on 

> index that's globally sequential is important for working with downstream operations.

In the prior implementation you know that the index is partitioned and what those values are.

ibis/backends/dask/execution/util.py
```diff
@@ -371,3 +458,36 @@ def _is_row_order_preserving(node: ops.Node):
...
+        # Special case where there's a column with name 'index'
+        unique_col_name = uuid.uuid4().hex
+
+        df = rename_index(df, unique_col_name)
+        df = df.reset_index().set_index(unique_col_name)
+        df = rename_index(df, None)
+    elif df.index.name is not None:
+        df = df.reset_index().set_index(df.index.name)
+    else:
+        df = df.reset_index().set_index('index')
```
gerrymanoim, line 486: Every `set_index` here will force a computation which we want to avoid. You need to pass `divisons` and `sorted` if you want set index to be an actual lazy operation.

gerrymanoim (changes requested):

ibis/backends/dask/execution/util.py
patcao: An example expression where this would come up is
`alltypes.int_col - alltypes.int_col.min().over(ibis.window())`

The problem shows up when we try to execute the subtraction binary op. 
`alltypes.int_col` is a dataframe with indices `[0, len(alltypes))`
`alltypes.int_col.min().over(ibis.window())` gives back a dataframe with `_ibis_index` that we set.

ibis/backends/dask/execution/util.py
patcao: For reduction ops, `result` can just be of type `int`.

ibis/backends/dask/execution/util.py
patcao: Yeah, i think this got introduced from resolving some merge conflicts poorly.

ibis/backends/dask/execution/util.py
patcao: Deleted this whole helper method.

ibis/backends/dask/execution/util.py
patcao: Yup, you're right here. It's duplicated and not needed

ibis/backends/dask/aggcontext.py
patcao: I've left it around because this is exactly what Pandas is doing right now.  
When we improve Dask test coverage it should hit this branch.

ibis/backends/dask/execution/util.py
```diff
@@ -355,6 +395,49 @@ def helper(
     return df
 
 
+def add_sequential_partitioned_sorted_column(
+    df: dd.DataFrame | dd.Series, index_name: str = '_ibis_index'
+) -> dd.DataFrame:
+    """Add a column that is already partitioned and sorted.
+
+    This column acts as if we had a global index across the distributed data.
+    This index needs to be consecutive in the range of [0, len(df)) allows
```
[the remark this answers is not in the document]
patcao: The cumsum does make the index consecutive. So that's an even stricter condition than sequential

ibis/backends/dask/execution/generic.py
```diff
@@ -339,7 +342,11 @@ def execute_cast_series_date(op, data, type, **kwargs):
 @execute_node.register(ops.Limit, dd.DataFrame, integer_types, integer_types)
 def execute_limit_frame(op, data, nrows, offset, **kwargs):
     # NOTE: Dask Dataframes do not support iloc row based indexing
-    return data.loc[offset : (offset + nrows) - 1]
```
patcao, line 342: This didn't behave correctly with multi-partition dataframes.
If the index of `data` wasn't consecutive this also wouldn't select the correct number of rows.

ibis/backends/dask/aggcontext.py
```diff
@@ -0,0 +1,234 @@
...
+from ibis.backends.pandas.aggcontext import (
+    AggregationContext,
+    compute_window_spec,
+    construct_time_context_aware_series,
+    get_time_col,
+    window_agg_udf,
+    wrap_for_agg,
+)
+
+# TODO Consolidate this logic with the pandas aggcontext.
```
jreback, line 18: do we have an issue to track?

ibis/backends/dask/aggcontext.py
```diff
@@ -0,0 +1,234 @@
...
+
+class Summarize(AggregationContext):
+    __slots__ = ()
+
+    def agg(self, grouped_data, function, *args, **kwargs):
+        if isinstance(function, str):
+            return getattr(grouped_data, function)(*args, **kwargs)
+
+        if not callable(function):
+            raise TypeError(f'Object {function} is not callable or a string')
```
jreback, line 30: tests hit this?

ibis/backends/dask/aggcontext.py
```diff
@@ -0,0 +1,234 @@
...
+            name = f"{name}_{ibis.util.guid()}"
+            frame = frame.assign(**{name: obj})
+
+        # set the index to our order_by keys and append it to the existing
+        # index
+        # TODO: see if we can do this in the caller, when the context
+        # is constructed rather than pulling out the data
+        columns = group_by + order_by + [name]
+        # Create a new frame to avoid mutating the original one
+        indexed_by_ordering = frame[columns].copy()
```
jreback, line 145: could use .assign here instead?

ibis/backends/dask/execution/window.py
```diff
@@ -2,28 +2,161 @@
...
+)
+from ibis.backends.pandas.execution.window import _post_process_group_by_order_by
+
+
+def _check_valid_window_frame(frame):
+    # TODO consolidate this with pandas
+    if frame.how == "range" and any(
+        not col.output_dtype.is_temporal() for col in frame.order_by
+    ):
+        raise NotImplementedError(
```
jreback, line 47: these caught in tests?

jreback (commented):
just a few comments

ibis/backends/dask/aggcontext.py
patcao: yeah this gets hit by `test_aggregation.test_reduction_ops`

ibis/backends/dask/execution/window.py
patcao: Yeah, the relevant tests are xfailed

ibis/backends/dask/aggcontext.py
patcao: https://github.com/ibis-project/ibis/issues/5911

patcao:
@cpcloud Do you mind reviewing this change? I don't seem to be able to request a review from you.

jreback:
@cpcloud if you have comments here
@patcao need to rebase as well

jreback:
i think the failures are unrelated.

cc @ibis-project/ibis-owners if can take a look

ibis/backends/dask/execution/window.py
```diff
@@ -62,86 +238,244 @@ def execute_window_frame(
...
-    )
+    if frame.group_by:
+        if frame.order_by:
+            raise NotImplementedError('Grouped and order windows not supported yet')
+            # TODO finish implementeing grouped/order windows.
+            # The implementaiton below is mostly working. Need to figure out a few Dask quirks.
+            # sorted_df, grouping_keys, ordering_keys = compute_sorted_frame(
+            #     df = root_data,
+            #     order_by=frame.order_by,
+            #     group_by=frame.group_by,
```
cpcloud, line 290: Please delete this commented code.

ibis/backends/dask/execution/window.py
```diff
@@ -62,86 +238,244 @@ def execute_window_frame(
...
+@execute_node.register(
+    (ops.Lead, ops.Lag),
+    (dd.Series, ddgb.SeriesGroupBy),
+    timedelta_types,
+    date_types + timestamp_types + (str, type(None)),
+)
+def execute_series_lead_lag_timedelta(
+    op, data, offset, default, aggcontext=None, **kwargs
+):
+    """An implementation of shifting a column relative to another one that is in units of time rather than rows."""
```
cpcloud, line 421:
```suggestion
    """Shift a column relative to another one in units of time rather than rows."""
```

ibis/backends/dask/udf.py
```diff
@@ -20,6 +21,24 @@
 )
 
 
+def create_gens_from_args_groupby(*args: tuple[ddgb.SeriesGroupBy, ...]):
```
cpcloud, line 24: Doesn't this function already exist? Why make an exact copy?

ibis/backends/tests/test_window.py
```diff
@@ -309,6 +309,13 @@ def test_grouped_bounded_expanding_window(
             lambda t, win: t.double_col.mean().over(win),
             lambda df: (df.double_col.expanding().mean()),
             id='mean',
+            # marks=[
+            #     pytest.mark.broken(
+            #         ["dask"],
+            #         reason="FIXME Dask complains about non-unique index. Unable to reproduce outside of this test.",
+            #     ),
+            # ],
```
cpcloud, line 317: Delete this code.

ibis/backends/dask/tests/execution/test_window.py
```diff
@@ -0,0 +1,679 @@
...
+
+@default
+@row_offset
+def test_lead(t, df, row_offset, default, row_window):
+    expr = t.dup_strings.lead(row_offset, default=default).over(row_window)
+    result = expr.execute()
+    expected = df.dup_strings.shift(execute((-row_offset).op())).compute()
+    if default is not ibis.NA:
+        expected = expected.fillna(execute(default.op()))
+    expected.name = f"Lead(dup_strings, {row_offset}, {default})"
```
cpcloud, line 54: All these name checks are unnecessary, use `check_names=False`

cpcloud (changes requested):

ibis/backends/dask/tests/execution/test_window.py
```diff
@@ -0,0 +1,679 @@
...
+
+@pytest.fixture
+def range_window():
+    return ibis.window(following=0, order_by='plain_datetimes_naive')
+
+
+@default
+@row_offset
+def test_lead(t, df, row_offset, default, row_window):
+    expr = t.dup_strings.lead(row_offset, default=default).over(row_window)
```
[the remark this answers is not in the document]
patcao: I agree that Dask test suite should be able to share the same code as the Pandas test suite. But I think this will be a large change, as we should refactor _ALL_ of the Dask test suite.

We can create test base class with the Pandas tests and have the Dask suite inherit from that. 
Just inheriting from a base class won't be enough for the Dask test suite though, as there's some subtle differences between Dask and Pandas. e.g. Dask dataframes are lazily evaluated, so in the test impls, we must call `.compute()`. 

We can deal with all of this, but we'd need to change the way that Pandas test suite is written. I think this would be best addressed in a separate PR where we can refactor _ALL_ of the Dask test suite at once.

@jreback We can create a follow up issue and schedule the work
@cpcloud Let me know if you're okay with this.

ibis/backends/dask/aggcontext.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class Cumulative(Window):
+    __slots__ = ()
+
+    def __init__(self, window, *args, **kwargs):
+        super().__init__('rolling', *args, window=window, min_periods=1, **kwargs)
+
+
+class Moving(Window):
```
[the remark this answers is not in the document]
patcao: They all have equivalent classes in the pandas aggcontext. I tried to inherit and share as much code as possible from the pandas impl.

patcao:
@cpcloud please give this another review

gforsyth:
> We can deal with all of this, but we'd need to change the way that Pandas test suite is written. I think this would be best addressed in a separate PR where we can refactor ALL of the Dask test suite at once.

I understand that this is a big undertaking, but in the interim, before the unknown date in the future when this happens, we're on the hook for maintaining that many more lines of duplicated tests.

cpcloud:
I think I am okay with the current situation with the following additions:

1. Someone from the TS side is added as a codeowner for anything dask related.
2. Until the test suite can be refactored, if there's a non-trivial failing test then any maintainer can add `@pytest.mark.broken` to that test without it blocking a PR from merging.

jreback:
> I think I am okay with the current situation with the following additions:
> 
> 1. Someone from the TS side is added as a codeowner for anything dask related.
> 2. Until the test suite can be refactored, if there's a non-trivial failing test then any maintainer can add `@pytest.mark.broken` to that test without it blocking a PR from merging.

@cpcloud sounds fine to me. I am already on the repo, we can certainly add @patcao here as well.

gforsyth (changes requested):
small nit

jreback:
@patcao looks like a conflict if you can rebase

jreback (approved):
lgtm

jreback:
@cpcloud ok here?

jreback:
going to merge shortly

jreback:
thanks @patcao

PR #5756, opened 2023-03-15
fix(bigquery): regenerate negative string index sql snapshots

cpcloud:
Fix the generated SQL for bigquery negative string indexing


cpcloud:
Commit lint passed, merging.

PR #5768, opened 2023-03-17
feat(table): implement `pivot_wider` API

cpcloud:
This PR implements `pivot_wider`, sometimes referred to as just "pivot" in other systems such as SQL.

The fundamental operation here is turning column values into column names.

This cannot be done without computation, to retrieve the new column names by
computing the distinct values of the to-pivot columns.

To that end, to get this information ibis *does* incur the overhead of computing these values.

While this behavior does deviate from other APIs in that it is not fully lazy,
I believe that the ability to express this computation concisely outweighs the
API deviance. There's _some_ precedent from `.cache()` as well which computes immediately.

However, if a user decides that they cannot live with this overhead then
another argument, `names` exists to allow someone to specify the specific
values that they would like to use as new columns and that will be used in
pivoting.

Another thing that is worth noting is that the default aggregate is `arbitrary`.

This may or may not be desirable, we can certainly change this to enforce passing an
aggregate function explicitly if folks think that's better.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -4,6 +4,7 @@ import collections
 import contextlib
 import functools
 import itertools
+import operator
 import re
 import warnings
 from keyword import iskeyword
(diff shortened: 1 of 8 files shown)


ibis/expr/types/relations.py
```diff
@@ -2957,6 +2958,415 @@ def pivot_longer(
...
+            id_cols = s._to_selector(id_cols)
+
+        if values_fn is None:
+            values_fn = _.arbitrary
+
+        if names is None:
+            # no names provided, compute them from the data
+            #
+            # cache them to avoid recomputing the same distinct pivot values more than once
+            names = self.select(names_from).distinct().cache().execute()
```
cpcloud, line 3334: `.cache()` can certainly be left out of this PR if desired.

ibis/expr/types/relations.py
```diff
@@ -2957,6 +2958,415 @@ def pivot_longer(
...
+        self,
+        *,
+        id_cols: s.Selector | None = None,
+        names_from: str | Iterable[str | s.Selector] = "name",
+        names: Iterable[str] | None = None,
+        names_prefix: str = "",
+        names_sep: str = "_",
+        names_sort: bool = False,
+        values_from: str | Iterable[str | s.Selector] = "value",
+        values_fill: int | float | str | ir.Scalar | None = None,
```
gforsyth, line 2972: I think on `pivot_longer` we opted to drop this argument because it could just be followed by a `fillna` -- but I can see the utility here where there can be many columns which require filling.

ibis/expr/types/relations.py
```diff
@@ -2957,6 +2958,415 @@ def pivot_longer(
...
+    @util.experimental
+    def pivot_wider(
+        self,
+        *,
+        id_cols: s.Selector | None = None,
+        names_from: str | Iterable[str | s.Selector] = "name",
+        names: Iterable[str] | None = None,
+        names_prefix: str = "",
+        names_sep: str = "_",
+        names_sort: bool = False,
```
gforsyth, line 2970: Do we need the `names_` prefixes?  I'm leaning towards not having them.

ibis/expr/types/relations.py
gforsyth, line 2972: maybe this is just `fill`?

gforsyth (commented):
all the examples are really great in this.

some small points (but all about naming, sorry)

ibis/expr/types/relations.py
cpcloud: Selector support in `fillna` might obviate this a bit, but I still things it's common enough that having to spell the selectors for the values columns out twice is kind of a nuisance.

ibis/expr/types/relations.py
cpcloud: I actually prefer them so that I know the thing the argument is applied to. IMO it's not obvious that `prefix`, `sep`, and `sort` would apply to column names and not the resulting values.

ibis/expr/types/relations.py
cpcloud: Simlilar to [prefixes for `names`-related things](https://github.com/ibis-project/ibis/pull/5768#discussion_r1142295077) I'd like to keep the `values_` prefix.

ibis/expr/types/relations.py
```diff
@@ -2957,6 +2958,415 @@ def pivot_longer(
...
+        values_from: str | Iterable[str | s.Selector] = "value",
+        values_fill: int | float | str | ir.Scalar | None = None,
+        values_fn: str
```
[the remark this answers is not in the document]
gforsyth: I think `agg` is good -- better that it describes the thing instead of just `fn`

cpcloud:
Thanks, yeah `names` is maybe not the best, but I don't have any better alternatives.

I will also remove the `.cache()` call too, I don't think it's necessary yet.

ibis/expr/types/relations.py
```diff
@@ -2957,6 +2958,427 @@ def pivot_longer(
...
+    @util.experimental
+    def pivot_wider(
+        self,
+        *,
+        id_cols: s.Selector | None = None,
+        names_from: str | Iterable[str | s.Selector] = "name",
+        names: Iterable[str] | None = None,
+        names_prefix: str = "",
+        names_sep: str = "_",
+        names_sort: bool = False,
```
gforsyth, line 2970:
```suggestion
        names_from: str | Iterable[str | s.Selector] = "name",
        names_prefix: str = "",
        names_sep: str = "_",
        names_sort: bool = False,
        use_names: Iterable[str] | None = None,
```
??

gforsyth: definitely not a blocker

cpcloud: Hm, `use_names` sounds like a bool to me. Let's try with `names`, though I am +1 on moving the `names` argument to this position.

cpcloud:
Thanks for the review!

PR #5769, opened 2023-03-17
feat(api): expand `distinct` API to support dropping duplicates based on a column subset

### LINKED ISSUE #5509: feat: Table.drop_duplicates()

I use this function in pandas all the time, and I had to roll my own for Ibis, and I don't think it's trivial, my implementation might have some bugs in it. I think it would get a lot of use from people.

```python

def drop_duplicates(
    t: Table,
    subset: str | list[str] | None = None,
    keep: Literal["first"] | Literal["last"] | Literal["none"] = "first",
) -> Table:
    """Return a Table with duplicate rows removed.

    Analogous to pandas.DataFrame.drop_duplicates().

    Parameters
    ----------
    t : Table
        The table to drop duplicates from.
    subset : str | list[str] | None
        Only consider certain columns for identifying duplicates,
        by default use all of the columns.
    keep : {"first", "last", "none"}, default "first"
        Determines which duplicates (if any) to keep.
        - first : Drop duplicates except for the first occurrence.
        - last : Drop duplicates except for the last occurrence.
        - none : Drop all duplicates
    """
    if subset is None:
        subset = t.columns
    if isinstance(subset, str):
        subset = [subset]
    if keep == "none":
        having = _.count() == 1
        how = "first"
    elif keep in {"first", "last"}:
        having = None
        how = keep
    else:
        raise ValueError(f"Invalid value for keep: {keep}")
    aggs = {col: t[col].arbitrary(how=how) for col in t.columns if col not in subset}
    gb = t.group_by(subset)
    if having is not None:
        gb = gb.having(having)
    return gb.agg(**aggs)
```

Note that I changed the signature to use "none" instead of `False` for the one value of `keep`, perhaps you want to stay with teh exact pandas API but I prefer "none" to be more clear, and more consistent so all options are strings.

target 5.0?

all

### Code of Conduct

cpcloud:
This PR expands our existing `distinct` method to support row deduplication based on a subset of columns.

Closes #5509.

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -995,3 +995,153 @@ def test_pivot_longer(backend):
     )
     df = res.limit(5).execute()
     assert not df.empty
+
+
+@pytest.mark.parametrize(
+    "on",
+    [
+        param(
+            ["cut"],
+            marks=[
+                pytest.mark.notimpl(
+                    ["mssql", "mysql"], raises=com.OperationNotDefinedError
+                ),
+            ],
+            id="one",
+        ),
+        param(
+            ["clarity", "cut"],
+            marks=[
+                pytest.mark.notimpl(
+                    ["mssql", "mysql"], raises=com.OperationNotDefinedError
+                ),
+            ],
+            id="many",
+        ),
+    ],
+)
+@pytest.mark.parametrize(
+    "keep",
+    [
+        param(
+            "first",
+            marks=pytest.mark.notimpl(
+                ["trino"],
+                raises=AssertionError,
+                reason="trino is more arbitrary than other backends",
+                strict=False,
+            ),
+        ),
+        param(
+            "last",
+            marks=[
+                pytest.mark.notimpl(
+                    ["bigquery", "snowflake"],
+                    raises=com.UnsupportedOperationError,
+                    reason="backend doesn't support last argument to arbitrary",
+                ),
+                pytest.mark.notimpl(
+                    ["trino"],
+                    raises=AssertionError,
+                    reason="trino is more arbitrary than other backends",
+                    strict=False,
+                ),
+            ],
+        ),
+    ],
+)
+@pytest.mark.notimpl(
+    ["druid", "impala"],
+    raises=(NotImplementedError, sa.exc.ProgrammingError, com.OperationNotDefinedError),
+    reason="arbitrary not implemented in the backend",
+)
+@pytest.mark.notimpl(
+    ["dask", "datafusion", "polars"],
+    raises=com.OperationNotDefinedError,
+    reason="backend doesn't implement window functions",
+)
+@pytest.mark.notimpl(
+    ["pandas"],
+    raises=com.OperationNotDefinedError,
+    reason="backend doesn't implement ops.WindowFunction",
+)
+def test_distinct_on_keep(backend, on, keep):
+    from ibis import _
+
+    t = backend.diamonds.mutate(one=ibis.literal(1)).mutate(
+        idx=ibis.row_number().over(order_by=_.one, rows=(None, 0))
+    )
+
+    requires_cache = backend.name() in ("mysql", "impala")
+
+    if requires_cache:
+        t = t.cache()
+    expr = t.distinct(on=on, keep=keep).order_by(ibis.asc("idx"))
+    result = expr.execute()
+    df = t.execute()
+    expected = (
+        df.drop_duplicates(subset=on, keep=keep or False)
+        .sort_values(by=["idx"])
+        .reset_index(drop=True)
+    )
+    backend.assert_frame_equal(result, expected)
+
+
+@pytest.mark.parametrize(
+    "on",
+    [
+        param(
+            ["cut"],
+            marks=[
+                pytest.mark.notimpl(
+                    ["mssql", "mysql"], raises=com.OperationNotDefinedError
+                ),
+            ],
+            id="one",
+        ),
+        param(
+            ["clarity", "cut"],
+            marks=[
+                pytest.mark.notimpl(
+                    ["mssql", "mysql"], raises=com.OperationNotDefinedError
+                ),
+            ],
+            id="many",
+        ),
+    ],
+)
+@pytest.mark.notimpl(
+    ["druid", "impala"],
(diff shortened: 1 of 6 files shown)


cpcloud:
cc @NickCrews if you're interested in trying this out for your use case!

ibis/expr/types/relations.py
```diff
@@ -917,40 +917,173 @@ def aggregate(
...
+          ...
+        ibis.common.exceptions.IbisError: Invalid value for keep: 'second' ...
         """
-        return ops.Distinct(self).to_expr()
+
+        import ibis.expr.selectors as s
+
+        if on is None:
+            # dedup everything
+            if keep != "first":
```
NickCrews, line 1056: wouldn't keep="last" also makes sense? It would just give you the same result as "first", but it still makes sense.

keep=None also would make sense, but to support that we would need to update the Distinct op.

ibis/backends/tests/test_generic.py
```diff
@@ -995,3 +995,93 @@ def test_pivot_longer(backend):
     )
     df = res.limit(5).execute()
     assert not df.empty
+
```
NickCrews, line 998: Should this test be in test_table.py, next to the other test function? It is a function on a table, not on a generic expression?

ibis/backends/tests/test_generic.py
```diff
@@ -995,3 +995,93 @@ def test_pivot_longer(backend):
...
+
+@pytest.mark.parametrize(
+    "on",
```
NickCrews, line 1001: See my comment below with merging this with the existing https://github.com/ibis-project/ibis/blob/9bcdf77ddafce75f0e5d8714d01dde81ed0b90f2/ibis/tests/sql/test_sqlalchemy.py#L358

That existing function implicitly tests the cases of on=None and keep=None, but perhaps you want to merge/update the tests so you explicitly are testing those cases as well?

ibis/expr/types/relations.py
cpcloud: `keep=None` can be supported by dropping the `DISTINCT` operator entirely and just using group by agg.

A `GROUP BY` with no aggregates is a valid operation that is effectively a `DISTINCT` so that might be the way to go here.

cpcloud: I probably won't implement that in this PR though.

ibis/expr/types/relations.py
```diff
@@ -917,40 +917,173 @@ def aggregate(
...
+    ) -> Table:
+        """Return a Table with duplicate rows removed.
 
-        Returns
-        -------
-        Table
-            Unique rows of `self`
+        Similar to `pandas.DataFrame.drop_duplicates()`.
+
+        !!! note "Some backends do not support `keep='last'`"
```
NickCrews, line 930: According to the tests, the backends won't work because they dont support window functions. That would imply that for these backends, it wouldn't work for any code path that goes down the group_by().agg() path, which would happen whenever on!=None. Am I understanding that correctly? So this note should be more like "Some backends do not support `on != None`"

ibis/tests/expr/test_table.py
```diff
@@ -1812,9 +1812,14 @@ def test_pivot_longer_no_match():
...
-    import ibis
-
     t = ibis.table(dict(value="int", lagged_value="int"), name="t")
 
     with pytest.raises(com.IbisTypeError, match="Deferred input is not allowed"):
         ibis.greatest(t.value, ibis._.lagged_value)
+
+
+@pytest.mark.parametrize("keep", ["last", None])
+def test_invalid_distinct(keep):
```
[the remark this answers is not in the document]
NickCrews: I would like all 3 tests (the two you add here, and the 1 existing test) to all be closer to each other, or combined, so it's easier to see which cases are getting tested.

NickCrews (changes requested):

ibis/expr/types/relations.py
cpcloud: > According to the tests, the backends won't work because they dont support window functions.

The window function use in the test is unrelated to the functionality of distinct.

cpcloud: It's used to ensure that the ordering of the output is comparable. This documentation line is correct and as intended.

ibis/backends/tests/test_generic.py
cpcloud: I'd rather do that in a follow-up so that there's less code to review in this PR.

ibis/tests/expr/test_table.py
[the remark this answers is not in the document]
NickCrews: Can all the tests at least all be located next to each other?

ibis/expr/types/relations.py
NickCrews: ahhh, I see, I thought the groupby caused a window function under the hood. Now I see it in the test.

ibis/tests/expr/test_table.py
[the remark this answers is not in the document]
cpcloud: I don't think it makes sense to put these all next to each other, they are testing different things in different ways and at different levels of ibis:

- `test_generic.py::test_distinct` is a part of our backend integration test suite that actually executes against all backends and verifies correctness
- `test_sqlalchemy.py::test_distinct` is a unit test to verify sqlalchemy output
- `test_table.py::test_invalid_distinct` is an expression-level test to verify failure mode behavior

I'm going to leave them as is for now.

ibis/backends/tests/test_generic.py
cpcloud: These are different kinds of tests, testing different things. I'm going to leave them as is.

ibis/backends/tests/test_generic.py
```diff
@@ -995,3 +995,93 @@ def test_pivot_longer(backend):
...
+    ["dask", "datafusion", "polars"],
+    raises=com.OperationNotDefinedError,
+    reason="backend doesn't implement window functions",
+)
+@pytest.mark.notimpl(
+    ["pandas"],
+    raises=com.OperationNotDefinedError,
+    reason="backend doesn't implement ops.WindowFunction",
+)
+def test_distinct(backend, on, keep):
```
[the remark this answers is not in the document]
cpcloud: Happy to rename!

jcrist (approved):
This API LGTM! Love the thorough examples in the docstring too.

PR #5786, opened 2023-03-21
test(backends): strict validate exceptions in test_window

krzysztof-kwitt:


cpcloud:
@krzysztof-kwitt Are you still working on this? No rush, just wanted to check in and make sure we should keep it open!

cpcloud:
I worked through the failures. This should be good to go.

PR #5787, opened 2023-03-21
refactor(duckdb): use native duckdb parquet reader unless auth required

### LINKED ISSUE #5748: does duckdb support for s3:// use fsspec or duckdb's native mechanism?

Queries generated by ibis against remote S3 resources using duckdb are taking about twice as long to run for me as when I call `duckdb.sql` with a SQL string directly (using duckdb's `httpfs` extension, not the fsspec support).  I was just wondering if the cause for this might be that ibis is passing the I/O through an additional python layer? (I notice considerably lower bandwidth load when the query is running in ibis too, data transfer rates of around 200 Mb/s instead of around 500 Mb/s).  

If ibis is currently using fsspec to handle the queries, would you consider using duckdb's native httpfs extension instead? I think it can provide considerably better performance in some cases.

4.1.0

duckdb

### Code of Conduct

gforsyth:
This does two things:

We now support passing in a list of paths / S3 URIs to read parquet files using pyarrow.dataset.

We use the native DuckDB parquet reader unless we hit an auth error, at which point we fall back to using pyarrow.dataset.

Resolves #5748 

Some local tests (not adding mocks for s3 stuff):

### private s3 read

```python
[ins] In [1]: import ibis

[ins] In [2]: con = ibis.duckdb.connect()

[ins] In [3]: con.read_parquet("s3://<snip>/10/parquet/customer/")  # private bucket
Out[3]: 
AlchemyTable: ibis_read_parquet_0
  c_custkey    int64
  c_name       string
  c_address    string
  c_nationkey  int64
  c_phone      string
  c_acctbal    float32
  c_mktsegment string
  c_comment    string
```

### direct duckdb read

```python
[ins] In [1]: import ibis
c
[ins] In [2]: con = ibis.duckdb.connect()

[ins] In [3]: files = [
         ...:     f"s3://gbif-open-data-us-east-1/occurrence/2023-02-01/occurrence.parquet/00000{i}"
         ...:     for i in range(3)
         ...: ]

[ins] In [4]: %time con.con.execute(f"CREATE VIEW testread AS SELECT * FROM read_parquet({files})")
CPU times: user 98 ms, sys: 16.5 ms, total: 115 ms
Wall time: 292 ms
Out[4]: <sqlalchemy.engine.cursor.LegacyCursorResult at 0x7f7efb05c700>
```

### ibis using duckdb read by default

```python
[ins] In [1]: import ibis

[ins] In [2]: con = ibis.duckdb.connect()

[ins] In [3]: files = [
         ...:     f"s3://gbif-open-data-us-east-1/occurrence/2023-02-01/occurrence.parquet/00000{i}"
         ...:     for i in range(3)
         ...: ]

[ins] In [4]: %time con.read_parquet(files)
CPU times: user 250 ms, sys: 37.2 ms, total: 288 ms
Wall time: 468 ms
Out[4]: 
AlchemyTable: ibis_read_parquet_0
  gbifid                           string
  datasetkey                       string
  occurrenceid                     string
  kingdom                          string
  phylum                           string
<snip>
```

sub 200 ms difference, which I will happily attribute to sql generation overhead.


gforsyth:
if I add only the fix to allow multiple s3 URIs to pyarrow.dataset, this is the timing on `master`:

```python
[ins] In [1]: import ibis

[ins] In [2]: con = ibis.duckdb.connect()

[ins] In [3]: files = [
         ...:     f"s3://gbif-open-data-us-east-1/occurrence/2023-02-01/occurrence.parquet/00000{i}"
         ...:     for i in range(3)
         ...: ]

[ins] In [4]: %time con.read_parquet(files)
CPU times: user 400 ms, sys: 21.8 ms, total: 421 ms
Wall time: 12.4 s
Out[4]: 
AlchemyTable: ibis_read_parquet_0
  gbifid                           string
  datasetkey                       string
  occurrenceid                     string
  kingdom                          string
  phylum                           string
<snip>
```

`pyarrow.dataset` doesn't seem to currently support reading from a list of s3 URIs, so we have to union a bunch of datasets together, which means this incurs (on my machine) about a ~4s penalty per-file

gforsyth:
> `pyarrow.dataset` doesn't seem to currently support reading from a list of s3 URIs, so we have to union a bunch of datasets together, which means this incurs (on my machine) about a ~4s penalty per-file

On a recent pyarrow nightly, this drops below 1 second per-file, about a 5-6x perf improvement

jcrist (approved):
This generally looks good to me!

Is there any meaningful way we could add tests here? I don't love mocks, but it also would be good to check that the error message raised here doesn't change (since we're filtering on both error type and message).

ibis/backends/duckdb/__init__.py
```diff
@@ -340,47 +340,67 @@ def read_parquet(
...
-        else:
-            if any(
-                source.startswith(("http://", "https://")) for source in source_list
+        # Default to using the native duckdb parquet reader
+        # If that fails because of auth issues, fall back to ingesting via
+        # pyarrow dataset
+        try:
+            self._read_parquet_duckdb_native(source_list, table_name, **kwargs)
+        except sa.exc.OperationalError as e:
+            if ("403 (Forbidden)" in repr(e)) or (
```
cpcloud, line 353: Is there nothing more specific coming from duckdb that we might be able to catch?

If not, is there any reason to match the exception at all?

gforsyth:
> Is there any meaningful way we could add tests here? I don't love mocks, but it also would be good to check that the error message raised here doesn't change (since we're filtering on both error type and message).

I couldn't come up with anything that wasn't so contrived as to be useless -- DuckDB (via duckdb_engine, I think) throws a 403 because something is forbidden and that's the cue to fall back to `pyarrow` -- short of spinning up a small webserver (hmm) I don't know how to test this

ibis/backends/duckdb/__init__.py
jcrist: I think it's useful to distinguish errors due to permission/url support issues (where falling back to the more flexible but slower option is needed) from errors due to:

- A file not existing
- A file being corrupted
- Some other IO failure (e.g. 500 errors or something)

Especially for network requests where there might be a bit of a lag before an error happens, falling back on any error seems risky. If the errors like these aren't `OperationalErrors` though then we might be fine to drop the message filtering.

cpcloud: I see, this is fine for now, though I wonder if we should aggressively remove this once pyarrow 12 is out and just tell people they have to upgrade pyarrow if they want better performance.

jcrist:
> short of spinning up a small webserver (hmm) I don't know how to test this

What about something like:

- Use `monkeypatch` and `MagicMock` to wrap both the native and fallback implementations (just to be able to detect if they're called during the test)
- Spin up a [HTTPServer](https://docs.python.org/3/library/http.server.html#http.server.HTTPServer) that always returns a 403 for all requests
- Check that both methods were called

This would check that if a server 403s the fallback is hit, which should detect if the error message ever changes.

ibis/backends/duckdb/__init__.py
cpcloud: It's [valid to raise an exception subclass](https://peps.python.org/pep-0249/#exceptions), so we can ask them to isolate those errors upstream.

cpcloud: That's not blocking this PR though.

gforsyth:
Ok -- `pytest-httpserver` is handy and handles throwing a 403, and then we can check that the error for the bad URI is coming from `pyarrow` instead, which confirms the fallback is working.

I also simplified the error check -- both the error types have an original error of  `duckdb.IOException` which I'm fine using as our fallback signal.  Then if we get more detailed errors upstream we can improve it.

pyproject.toml
```diff
@@ -114,6 +114,7 @@ pytest-snapshot = ">=0.9.0,<1"
 pytest-xdist = ">=2.3.0,<4"
 hypothesis = ">=6.58.0,<7"
 sqlalchemy = ">=1.4,<3"
+pytest-httpserver = "^1.0.6"
```
cpcloud, line 117: Nice!

cpcloud (approved):
🚢 it!

PR #5789, opened 2023-03-21
fix(duckdb): disable the progress bar by default

cpcloud:
This PR fixes an issue where the progress bar enabled by default in duckdb causes jupyter lab kernels to crash. I am not sure why this happens, but disabling the progress bar by default serves as a viable temporary workaround.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -135,8 +135,10 @@ class Backend(BaseAlchemyBackend):
         )
 
         @sa.event.listens_for(engine, "connect")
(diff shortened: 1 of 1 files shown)


NickCrews:
OMG @cpcloud you're a lifesaver! This was blocking me in so many places. Thank you! Adding `conn.raw_sql("SET enable_progress_bar = false").connection.close()` at the beginning of every connection solves it for me.

I filed a bug with duckdb at https://github.com/duckdb/duckdb/issues/6815, IDK if that is the responsible lib or if it's jupyterlab.

PR #5804, opened 2023-03-23
refactor(duckdb): simplify `to_pyarrow_batches` implementation

jcrist:
For `to_pyarrow_batches` we need to keep the backing cursor alive for the lifetime of the `RecordBatchReader`. Previously we did this by implementing our own `IbisRecordBatchReader` class that wrapped a real `RecordBatchReader` along with a `cursor` object. However, this implementation was flawed for any consumers trying to access the raw C++ level reader, since the wrapping reader wasn't initialized properly (subclassing the cython class, but not initializing it properly results in invalid instances). We alternatively could choose to implement the same `RecordBatchReader` interface at the python level but _not_ subclass from `pyarrow.ipc.RecordBatchReader`, but this wouldn't work for any consumers doing explicit `isinstance` checks or trying to access the raw C++ readers.

Instead, we now use `weakref.finalize` to keep a reference to the cursor alive for the lifetime of the reader. This means that the cursor won't be drop when `reader.close()` is called, but will be dropped when the reader is GC'd.


ibis/backends/duckdb/__init__.py
```diff
@@ -574,7 +572,12 @@ def to_pyarrow_batches(
         cursor = con.execute(sql)
 
         reader = cursor.cursor.fetch_record_batch(chunk_size=chunk_size)
-        return IbisRecordBatchReader(reader, cursor)
+        # Use a weakref finalizer to keep the cursor alive until the record
+        # batch reader is garbage collected. It would be nicer if we could make
+        # the cursor cleanup happen when the reader is closed, but that's not
+        # currently possible with pyarrow.
+        weakref.finalize(reader, lambda: cursor.close())
```
kszucs, line 579: Much nicer!

Do we have a test to check the cursor's lifetime?

gforsyth (approved):
:shipit:

cpcloud:
This is great, thank you @jcrist!

Issue #5813, opened 2023-03-24
feat(ux): unlimited columns in interactive repr in jupyter


NickCrews:
### Is your feature request related to a problem?

Originally from https://app.reviewnb.com/ibis-project/ibis/pull/5727/discussion/#comment-1133062663

```python
import ibis

ibis.options.interactive = True

cars = ibis.examples.mtcars.fetch() # 11 columns
big = cars.join(t, "mpg")  #21 columns
big
```

In a notebook, where I can scroll sideways, I would like to be able to see all columns by default. Currently, it looks like it's capped at 20.

<img width="210" alt="image" src="https://user-images.githubusercontent.com/10820686/227621492-edc81bfe-db7a-4a42-ad3b-df30a235eeb0.png">

### Describe the solution you'd like

all columns when there is a scrollbar. I'm not sure how to detect if there's a scrollbar. In a vanilla terminal, it would be nice if we continued to cap the width to the terminal width.

### What version of ibis are you running?

master

### What backend(s) are you using, if any?

NA

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #5814, opened 2023-03-24
feat(repr): Set default options.repr.interactive.max_columns to None

NickCrews:
Fixes https://github.com/ibis-project/ibis/issues/5813

### DIFF
--- a/ibis/config.py
+++ b/ibis/config.py
@@ -92,7 +92,7 @@ class Interactive(Config):
     """
 
     max_rows: int = 10
-    max_columns: Optional[int] = 0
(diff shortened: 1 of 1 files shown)


jcrist:
This isn't quite right - `None` means no limit, even for terminals. Previously we had an inferred cap of 20, matching pandas's behavior in notebooks. I've pushed up an alternative fix in #5816 removing this (apologies for not commenting here, but it was easier to just push up the proper fix myself). Closing in favor of #5816.

NickCrews:
That looks great, thank you!

PR #5819, opened 2023-03-24
test(snowflake): fix arbitrary comparisons

cpcloud:
This PR fixes a flaky snowflake test for the new distinct API.

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -1025,30 +1025,14 @@ def test_pivot_longer(backend):
 @pytest.mark.parametrize(
     "keep",
     [
-        param(
-            "first",
-            marks=pytest.mark.notimpl(
-                ["trino"],
-                raises=AssertionError,
-                reason="trino is more arbitrary than other backends",
-                strict=False,
-            ),
-        ),
+        "first",
         param(
             "last",
-            marks=[
-                pytest.mark.notimpl(
-                    ["bigquery", "snowflake"],
-                    raises=com.UnsupportedOperationError,
-                    reason="backend doesn't support last argument to arbitrary",
-                ),
-                pytest.mark.notimpl(
-                    ["trino"],
-                    raises=AssertionError,
-                    reason="trino is more arbitrary than other backends",
-                    strict=False,
-                ),
-            ],
+            marks=pytest.mark.notimpl(
+                ["bigquery", "snowflake", "trino"],
+                raises=com.UnsupportedOperationError,
+                reason="backend doesn't support how='last'",
+            ),
         ),
     ],
 )
@@ -1086,7 +1070,7 @@ def test_distinct_on_keep(backend, on, keep):
         .sort_values(by=["idx"])
         .reset_index(drop=True)
     )
-    backend.assert_frame_equal(result, expected)
+    assert len(result) == len(expected)
 
 
 @pytest.mark.parametrize(
@@ -1151,4 +1135,4 @@ def test_distinct_on_keep_is_none(backend, on):
         .sort_values(by=["idx"])
         .reset_index(drop=True)
     )
-    backend.assert_frame_equal(result, expected)
+    assert len(result) == len(expected)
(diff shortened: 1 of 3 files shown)


ibis/backends/tests/test_generic.py
```diff
@@ -1028,9 +1028,9 @@ def test_pivot_longer(backend):
         param(
             "first",
             marks=pytest.mark.notimpl(
-                ["trino"],
+                ["trino", "snowflake"],
                 raises=AssertionError,
-                reason="trino is more arbitrary than other backends",
+                reason="backend is more arbitrary than other backends",
```
jcrist, line 1033: Does this mean that `pivot_longer` works and is correct, but the row output/ordering isn't consistent so the test sometimes fails? If so, perhaps we should change the test to not error, since right now it looks like `pivot_longer` isn't implemented for these backends.

jcrist: Not a blocker of course, if this fixes the flaky test feel free to merge.

cpcloud: > Does this mean that `distinct` works and is correct, but the row output/ordering isn't consistent so the test sometimes fails? If so, perhaps we should change the test to not error, since right now it looks like `distinct` isn't implemented for these backends.

Yeah, unfortunately arbitrary really is arbitrary, so this is hard to compare against pandas, because snowflake's any_value keeps nulls, so we use min in our implementation to workaround that.

jcrist: In these cases, could we just check that the guarantees of `distinct` are true (the rows are distinct on the proper columns) and not compare against a single "correct" solution?

cpcloud: I'll move it to just a length check for now.

cpcloud:
Cloud backends look good

```
…/ibis on  fix-flaky-snowflake-tests is 📦 v5.0.0 via 🐍 v3.10.10 via ❄️  impure (ibis-3.10.10-env)
❯ pytest -m 'bigquery or snowflake' -n auto --dist=loadgroup --snapshot-update --allow-snapshot-deletion -q
bringing up nodes...
xxxxxxxxxxxxxxxxxx.x.x.....x............................................x..........x....x.x...x...........x........x..............................x............................x......x......x.xx.....xxx.xxx.....x.....x. [ 10%]
...................x...................x....x...x.x...x...x...................x........x..xx.......x....xxxx.......x.....xx.........x.x..x..xx...xx...x......xx...x...s......................................x..x......... [ 21%]
x.......x.x.................x.....................x...xx..x.x..x............x...x..x..x..x...x............x....x...x........................x..x............x.....x..............x........x................s.............. [ 31%]
...x..s..x.............................x.x........x..........xx...x.x...x.x..x...x...x..x..x.x..x.xxx..s..x..x..x.x.x.xx.x...........xx..x..x.xx..x......x..x...xxxx....................x....x....x.x..x..s.........x..x.x [ 42%]
..xxxxxxxxxxxxxxxxxxxxxx...x...............x..........................................x..x....x.............................x.......................x..........................................x..........x..x.xx....x..x. [ 53%]
xx....x..x...xx....x.xx.x.x..x....x.xx..........xx.......x...x..........x..xx..x.....x.x..x..............xx......x..x.xx......x.........x..x...........x..x.xx...x......xx...............x....x................x.........x [ 63%]
.x....x.x.x.......x......x...x...........x.......x....x..x.....x.....x..xx....x.x..x..............xx...x..............................x.........x......x..x..........................x....................x..........xx... [ 74%]
.xxxx.....xx...x.xxxxxx.x.xxxxxxx.x.xx.x...xxx.x..x.x.x..xx.....x.x.x.x.x.x.x.........x.x.x....xx....xxx.xx.x..........x.xx..xx.....x.x...xxxx.xx.x..x....x....x.x.x......x........x...xx................x...xx........... [ 85%]
..........x............................xxx..x...x.........................................................................................x..........................................................................x.... [ 95%]
...........x......x..................................x......x....s...s......x.........                                                                                                                                     [100%]
1690 passed, 7 skipped, 351 xfailed in 208.02s (0:03:28)
```

Merging!

PR #5824, opened 2023-03-25
feat(common): add pattern matchers

kszucs:
Converted the previous validator system into a pattern matching system.

The previously used validator system had the following problems:
- Used curried functions which was error prone to missing arguments and was hard to debug.
- Used exceptions for control flow which raising exceptions from the innermost function call giving cryptic error messages.
- While it was similarly composable it was hard to see whether the validator was fully constructed or not.
- We wasn't able traverse the nested validators due to the curried functions.

In addition to those the new approach has the following advantages:
- The pattern matching system is fully composable.
- Not we support syntax sugar for combining patterns using & and |.
- We can capture values at mutiple levels using either `>>` or `@` syntax.
- Support structural pattern matching for sequences, mappings and objects.
- We will be able to cache the pattern matches more easily.

The following TODOs are left:
- refactor the validators to provide a backward compatibility API but using the new patterns under the hood
- switch to use the new patterns everywhere in ibis core

but going to handle them in #5899


kszucs:
Closing in favor #5899

PR #5825, opened 2023-03-25
feat(pyspark): add ArrayFilter operation

tokoko:
This PR adds `ops.ArrayFilter` support to pyspark backend. 

Both Ibis and PySpark use lambdas as a way to provide a predicate, so there needs to be a function-to-function translation. Function argument resolution looks a bit hacky, but I don't think it could have been done simpler.

### DIFF
--- a/ibis/backends/pyspark/compiler.py
+++ b/ibis/backends/pyspark/compiler.py
@@ -1655,6 +1655,20 @@ def compile_array_collect(t, op, **kwargs):
     return F.collect_list(src_column)
 
 
+@compiles(ops.Argument)
+def compile_argument(t, op, arg_columns, **kwargs):
+    return arg_columns[op.name]
+
+
+@compiles(ops.ArrayFilter)
+def compile_array_filter(t, op, **kwargs):
+    src_column = t.translate(op.arg, **kwargs)
+    return F.filter(
+        src_column,
+        lambda x: t.translate(op.result, arg_columns={op.parameter: x}, **kwargs),
+    )
+
+
 # --------------------------- Null Operations -----------------------------
 
 
--- a/ibis/backends/pyspark/tests/test_array.py
+++ b/ibis/backends/pyspark/tests/test_array.py
@@ -166,3 +166,17 @@ def test_array_collect(client):
         .rename(columns={'array_int': 'collected'})
     )
     tm.assert_frame_equal(result, expected)
+
+
+def test_array_filter(client):
+    table = client.table('array_table')
+    expr = table.select(
+        table.array_int.filter(lambda item: item != 3).name('array_int')
+    )
+    result = expr.execute()
+    df = table.compile().toPandas()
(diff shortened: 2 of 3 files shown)


ibis/backends/pyspark/compiler.py
```diff
@@ -1655,6 +1655,30 @@ def compile_array_collect(t, op, **kwargs):
...
+        )
+
+
+@compiles(ops.ArrayFilter)
+def compile_array_filter(t, op, **kwargs):
+    src_column = t.translate(op.arg, **kwargs)
+    return F.filter(
+        src_column,
+        lambda x: t.translate(
+            op.result, **{**kwargs, **{PARAM_ARG_COLUMNS: {op.parameter: x}}}
```
cpcloud, line 1677: Then this becomes:

```suggestion
            op.result, arg_columns={op.parameter: x}, **kwargs
```

ibis/backends/pyspark/compiler.py
```diff
@@ -1655,6 +1655,30 @@ def compile_array_collect(t, op, **kwargs):
...
+
+
+@compiles(ops.Argument)
+def compile_argument(t, op, **kwargs):
+    if PARAM_ARG_COLUMNS in kwargs:
+        return kwargs[PARAM_ARG_COLUMNS][op.name]
+    else:
+        raise ValueError(
+            f'{PARAM_ARG_COLUMNS} needs to be passed for ops.Argument to be translated'
+        )
```
cpcloud, line 1668: Any reason we can't make this an explicit keyword argument?

```suggestion


@compiles(ops.Argument)
def compile_argument(t, op, arg_columns=None, **kwargs):
    if arg_columns is not None:
        return arg_columns[op.name]
    else:
        raise ValueError(
            f'`arg_columns` needs to be passed for ops.Argument to be translated'
        )
```

cpcloud (changes requested):
Thanks for doing this! A couple of suggestions on the implementation!

ibis/backends/pyspark/compiler.py
tokoko: No reason. I instinctively thought it wouldn't be passed down the chain this way, but of course it will. In this case I should probably also make arg_columns a required parameter and drop the validation as well.

cpcloud: No worries!

If you're interested, after this PR you should be able to use the same approach to implement `ops.ArrayMap`!

tokoko: Made the changes, thanks. Sure, I will add `ops.ArrayMap` as well.

cpcloud (approved):
LGTM! 🚢 it!

cpcloud:
@tokoko Can you squash to a single commit? Will merge on green after that!

cpcloud:
Actually, I can squash it for you, don't worry about it!

cpcloud:
Thanks for the PR, keep 'em comming!

PR #5826, opened 2023-03-25
feat(pyspark): add partial support for interval types

tokoko:
This PR adds partial support for Ibis `Interval` types by mapping pyspark `DayTimeIntervalType` to appropriate Ibis types. Only four `DayTimeIntervalType` interval types are supported as part of this PR:

- INTERVAL DAY
- INTERVAL HOUR
- INTERVAL MINUTE
- INTERVAL SECOND

types with several units at once (e.g. `INTERVAL DAY TO HOUR`) still lead to an error, as well as types with a different time unit.

P.S. Spark also has a `YearMonthIntervalType`, but it's not yet exposed in pyspark.

### DIFF
--- a/ibis/backends/pyspark/datatypes.py
+++ b/ibis/backends/pyspark/datatypes.py
@@ -79,6 +79,28 @@ def _spark_struct(spark_dtype_obj, nullable=True):
     return dt.Struct(fields, nullable=nullable)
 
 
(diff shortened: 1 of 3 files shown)


cpcloud (approved):
Added a check against `execute` to the test you added.

This PR LGTM!

PR #5832, opened 2023-03-26
fix(duckdb): allow configuration options from plugins

cpcloud:
This PR allows users to configure duckdb options on `connect` when the configuration value lives in an extension, as long as the extension is installed and not already loaded.


cpcloud:
This incurs a nearly 2x overhead to `ibis.duckdb.connect()`, so I'm going to close this PR. It doesn't seem worth the hit.

PR #5834, opened 2023-03-26
feat(pyspark): add DateDiff operation

tokoko:
This PR adds support for `ops.DateDiff` in pyspark backend. Wraps the function with the same name in pyspark, but returns `Interval('D')` type instead of an integer.


cpcloud (approved):
LGTM, thank you!

PR #5837, opened 2023-03-27
test(sqlite): run tests in parallel using an in-memory database

cpcloud:
This PR moves our sqlite test suite to be based entirely on an in-memory database which lets us run the test suite in parallel for this backend. After this PR, all SQLAlchemy backend test suites can be safely run in parallel!


ibis/backends/sqlite/tests/test_functions.py
```diff
@@ -716,25 +716,6 @@ def test_scalar_parameter(alltypes):
     tm.assert_series_equal(result, expected)
 
 
-def test_compile_twice(dbpath):
```
cpcloud, line 719: I removed this test because it doesn't appear to be testing anything useful.

cpcloud:
Thanks! Squashing and merging!

PR #5843, opened 2023-03-27
refactor(datatypes): use typehints instead of rules

kszucs:
Depends on #5847

- switched to use type annotations in `datatypes/core.py` and `schema.py` instead of rules
- introduced enum classes to handle temporal units, these enums can be used in annotations to automatically coerce variants of different units

### DIFF
--- a/ibis/common/enums.py
+++ b/ibis/common/enums.py
@@ -1,0 +1,108 @@
+from abc import ABCMeta
+from enum import Enum, EnumMeta
+
+from public import public
+
+from ibis.common.validators import Coercible
+
+
+class ABCEnumMeta(EnumMeta, ABCMeta):
+    pass
+
+
+class Unit(Coercible, Enum, metaclass=ABCEnumMeta):
+    @classmethod
+    def __coerce__(cls, value):
+        # first look for aliases
+        value = cls.aliases().get(value, value)
+
+        # then look for the enum value (unit value)
+        try:
+            return cls(value)
+        except ValueError:
+            pass
+
+        # then look for the enum name (unit name)
+        if value.endswith("s"):
+            value = value[:-1]
+        try:
+            return cls[value.upper()]
+        except KeyError:
+            raise ValueError(f"Unable to coerce {value} to {cls.__name__}")
+
+    @classmethod
+    def aliases(cls):
+        return {}
+
+    @property
+    def singular(self) -> str:
+        return self.name.lower()
+
+    @property
+    def plural(self) -> str:
+        return self.singular + "s"
+
+    @property
+    def short(self) -> str:
+        return self.value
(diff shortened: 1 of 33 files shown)


ibis/expr/operations/temporal.py
```diff
@@ -99,7 +97,7 @@ class DateTruncate(Value):
 @public
 class TimeTruncate(Value):
     arg = rlz.time
-    unit = rlz.map_to(_time_units)
+    unit = rlz.coerced_to(TimeUnit)
```
kszucs, line 100: Where we use `unit = rlz.coerced_to(TimeUnit)` we can switch to type annotations `unit: TimeUnit`

ibis/common/enums.py
```diff
@@ -0,0 +1,108 @@
+from abc import ABCMeta
+from enum import Enum, EnumMeta
+
+from public import public
+
+from ibis.common.validators import Coercible
+
```
kszucs, line 7: Do you have a bettern name for the module than `enums`?

ibis/expr/datatypes/core.py
```diff
@@ -411,10 +404,11 @@ class Time(Temporal, Primitive):
 class Timestamp(Temporal, Parametric):
     """Timestamp values."""
 
-    timezone = optional(instance_of(str))
+    timezone: Optional[str] = None
     """The timezone of values of this type."""
 
-    scale = optional(isin(range(10)))
+    # Literal[*range(10)] is only supported from 3.11
+    scale: Optional[Literal[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]] = None
```
kszucs, line 411: We could use `Annotated[int, IsIn(range(10))]` here as well.

pyproject.toml
```diff
@@ -414,6 +414,7 @@ ignore = [
   "SIM118", # remove .keys() calls from dictionaries
   "SIM300", # yoda conditions
   "UP037", # remove quotes from type annotation
+  "UP007", # Optional[str] -> str | None
```
kszucs, line 417: Our typehint evaluation doesn't support the new syntax for `python<3.10`. To support that we would need to rewrite the AST before evaluating, but I don't think that worth the effort. 

Nonetheless if we would like to support it we can use the [solution from typeguard](https://github.com/agronholm/typeguard/blob/master/src/typeguard/_union_transformer.py).

ibis/common/enums.py
cpcloud: They seem fine to me!

cpcloud (approved):
LGTM!

Did you get a chance to test against cloud backends?

kszucs:
Both snowflake and bigquery pass.

PR #5847, opened 2023-03-28
refactor(datatypes): remove direct `ir` dependency from `datatypes`

kszucs:
Previously we had a dependency circle:
- datatypes used ir
- operations used datatypes
- ir used both operations and datatypes

depends on https://github.com/ibis-project/ibis/pull/5848


kszucs (commented):
+1

PR #5848, opened 2023-03-28
refactor(api): remove deprecated functions marked for removal in 6.0

kszucs:
BREAKING CHANGE: removed the following deprecated methods:
- `Backend.load_data()` from all backends
- `Backend.exists_database()` from all backends
- `Backend.exists_table()` from all backends
- `ibis.sequence()`
- `Schema.merge()`
- `Column.summary()`
- `NumericColumn.summary()`
- `GroupedArray.summary()`


cpcloud:
There are a few other things floating around that are deprecated but don't use the `deprecated` decorator. Can you do a sweep through the codebase and see if anything else is lurking?

kszucs:
Those do not have an explicit `removed_in` marker, so we may need to discuss them separately. How about removing them in follow-up PRs?

cpcloud:
That works!

kszucs:
I will go ahead and merge it, so that I can roll-up the remaining PRs.

kszucs (commented):
+1

PR #5854, opened 2023-03-28
chore(deps): remove unnecessary pyarrow extra

cpcloud:
This PR removes the unnecessary pyarrow extra

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -104,6 +104,7 @@ jobs:
             extras:
               - postgres
               - geospatial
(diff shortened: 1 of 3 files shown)


cpcloud:
It's not in a release though, so there's no reason to release with it.

kszucs:
> It's not in a release though, so there's no reason to release with it only to immediately delete it.

Has it been added since 5.0?

cpcloud:
> > It's not in a release though, so there's no reason to release with it only to immediately delete it.
> 
> Has it been added since 5.0?

Yep.

PR #5858, opened 2023-03-28
feat(api): support passing a format string to `Table.relabel`

jcrist:
This adds support for passing a format string to `Table.relabel`, to allow relabeling all columns with a particular format. This is most useful for adding a common prefix or suffix, possibly before forwarding to a `join` call.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -1724,6 +1724,7 @@ class Table(Expr, _FixedTextJupyterMixin):
         self,
         substitutions: Mapping[str, str]
         | Callable[[str], str | None]
+        | str
         | Literal["snake_case"],
     ) -> Table:
         """Rename columns in the table.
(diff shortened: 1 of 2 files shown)


ibis/expr/types/relations.py
```diff
@@ -1818,6 +1831,18 @@ def rename(c):
...
                 return c.lower()
 
+        elif isinstance(substitutions, str):
+
+            def rename(name):
+                return substitutions.format(name=name)
+
+            # Detect the case of missing or extra format string parameters
+            try:
+                invalid = rename("b026fa95") == rename("530505b1")
```
cpcloud, line 1841: Are we trying to prevent a misunderstanding of format string syntax?

jcrist: It's to throw a better error in the case of typos/user error. Otherwise you'll get a column name conflict error, which masks the real error that you just gave a bad format string. A few possible failure modes:

```python
# Typo of `snake_case`
t.relabel("snake_cse")

# Typo of format string
t.relabel("prefix_{name")

# Wrong format parameter
t.relabel("prefix_{column}")

# Using %s formatting instead
t.relabel("prefix_%s")
```

cpcloud: can we generate two value and stick them in globals to avoid the magic strings?

jcrist: Would a better comment fix things? It's not clear to me how adding indirection through globals would improve things.

Could instead do something that looks a bit clearer like

```
rename("_unlikely_str1_") == rename("_unlikely_str2_")
```

cpcloud: Random looking strings obfuscate intent. Comments help, but why comment when you can name a variable in a way that expresses the intent?

jcrist: I've pushed up a change. Is that sufficient?

cpcloud (approved):
LGTM!

PR #5812, opened 2023-03-24
docs: add a getting started guide

gforsyth:
A draft that needs more work, but putting it up so we can iterate on it.

My thought was that we should have a sample DB to download and then connect to, as opposed to using `examples` (just for this, the examples are great)

PR #5859, opened 2023-03-28
docs: style and review fixes in getting started guide

gforsyth:
Followup to #5812, addressing some comments from @jcrist


gforsyth:
I opted to leave in the selectors for the moment, but I'm also waffling

PR #5865, opened 2023-03-29
ci: constrain mamba because later versions are broken

cpcloud:
This PR constrains mamba to a version that works for me to generate lock files. The current version running in CI seems to be broken.


cpcloud:
Merging, so this can be tested in CI.

PR #5867, opened 2023-03-29
fix(ux): use guaranteed length-1 characters for `NULL` values

### LINKED ISSUE #5860: bug(ux): choose a different character for `NULL` values in interactive repr
Currently we use display null values as `∅` in the interactive repr. This works, some supposedly-monospace fonts display this character with a double width, leading to the repr columns no longer being aligned.

See this screenshot of our new getting started guide, for example:

image

While this could be fixed by changing the font, this has happened enough in enough locations that I think we should just pick a less fancy character.

A few options:

- No value. This is pretty common in table reprs (e.g. duckdb's table repr does this). Since we display the empty string as `~`, we shouldn't have issues distinguishing `""` from `NULL`.
- `*`
- `-`
- `~` (we currently use this for the empty string)
- `NA`

Thoughts?

cpcloud:
Closes #5860.

Here's the new output:

image


cpcloud:
I'll wait to address the failures so that people can chime in with opinions before I do the work of updating doctests and examples.

jcrist (commented):
I was going to comment that I didn't like the right-justified `NULL`, but I see you've already reverted that :). Overall this looks good to me!

PR #5868, opened 2023-03-29
test(duckdb): make remaining tests pass through extension load on init

cpcloud:
Hopefully this should fix the last of the extension loading issues


gforsyth (approved):
Looks good to me!  :crossed_fingers: that this sorts the rest out.

PR #5873, opened 2023-03-30
ci: avoid building macos nix devShell

cpcloud:
Avoid build the nix macos devShell so that CI finishes faster. Right now the job is building duckdb, and it is not clear that anyone is using the nix setup for macos. The devShell step is intended to populate the environment cache, so we do not need it for macos, but we should keep the previous test step so that we have some coverage of macos.


cpcloud:
Syntax is valid and job runs. Merging to verify that the devshell build doesn't run on macos against `master`.

PR #5875, opened 2023-03-30
feat: rethink join column deduplication

### LINKED ISSUE #5537: Rethink `join` column deduplication
Based on comments in #5514 (and some internal discussions), I think we might want to change how we handle duplicate column names in joins.

## The problem

Given tables `t1` and `t2` like:

```python
import ibis

t1 = ibis.table({"id": "int", "a": "string", "b": "string"})
t2 = ibis.table({"id": "int", "a": "string", "c": "string"})
t3 = ibis.table({"id": "int", "d"})
```

joining these tables on `id` will result in a table with duplicate column names (`a`, and `id` if not an inner join). Ibis requires tables have a valid schema with unique column names after every operation. As such, we can't have two `a` columns in the output, so we need to deduplicate these names. The current strategy is:

- Duplicate column names have suffixes appended to both the left & right versions (defaulting to `"_x"` and `"_y"`)

- *Except* for duplicate columns that are the join keys in an inner join only. These columns are known to be exactly identical (due to the inner join); only one copy is left in the output.

```python
In [4]: t1.join(t2, "id").columns  # inner join, only one copy of `id` kept
Out[4]: ['id', 'a_x', 'b', 'a_y', 'c']

In [5]: t1.left_join(t2, "id").columns  # left join, both `id` and `a` are deduplicated
Out[5]: ['id_x', 'a_x', 'b', 'id_y', 'a_y', 'c']
```

While simple and correct, this behavior can be unergonomic/annoying for a few reasons:

- Chaining multiple non-inner joins where all keys have the same name (e.g. `t1.left_join(t2, "id").left_join(t3, "id")`). In this case the result of the first join will have `id` deduplicated into `id_x` and `id_y`. This causes the second join to error since there's no `id` column anymore.

- The magical `_x` and `_y` suffixes only show up for conflicting column names. When joining large tables it can be hard to know if a conflict will happen, and if so what columns will conflict. Is `name` ok? Or do I need to use `name_x`?

- The suffixes often result in weird names. Every row in a table is generally an entity of some form, it often reads clearer to me to use a prefix instead. For example, in a table of customers, it's clearer to have `customer_name`/`customer_id` rather than `name_customer`/`id_customer`. Both are better than the default of `name_x`/`id_x`.

- Our current mixed support for passing in expressions that don't share a common parent (e.g. `t1.join(t2, "id").join(t3, "id").select(t1.a)`) can lead to invalid or weird queries. In the past we had talked about fully outlawing these cases and erroring nicely. I think this would be good to do, and would eliminate some of the issues here.

Often when using our current API I find myself doing one of two things:

- Passing in `suffixes=("", "_some_name")` to only append a suffix to duplicate columns in the right table, leaving the left columns as they are.

- Renaming all columns in the right table with a prefix before joining using `Table.relabel`. Something like `right.relabel(lambda c: f"prefix_{c}")`.

There are some rough edges here, and I think they can be improved.

jcrist:
This deprecates the existing `suffixes` kwarg in our join methods in favor of `lname`/`rname` kwargs.

```python
# rename any duplicate columns in the right table with a `_right` suffix
t.join(t2, "id", rname="{name}_right")

# rename any duplicate columns in the left table with a `left_` prefix
t.join(t2, "id", lname="left_{name}")
```

It also deprecates the default deduplication behavior; in the future any joins that would require column renaming for deduplication will require the user to explicitly pass in `lname`/`rname` (otherwise they'll get a nice error telling them to do so).

I'm not sold on `lname`/`rname` instead of `lsuffix`/`rsuffix` (as pandas does). I went with the more powerful/flexible option for now, but if this is confusing/verbose we can change it. I'm also not sold on not having a default behavior, as many joins will require the user to pass in a kwarg to resolve duplicate column names. We might instead do as polars does, and only rename duplicate columns in the right table (which is what users commonly want).

Just pushing this up for conversation for now. I've fixed a bunch of tests, but wouldn't be surprised if things fail in lots of places still.

Fixes #5537.

### DIFF
--- a/ibis/backends/dask/tests/execution/test_join.py
+++ b/ibis/backends/dask/tests/execution/test_join.py
@@ -38,10 +38,12 @@ join_type = pytest.mark.parametrize(
     ],
 )
 
+kws = {"lname": "{name}_x", "rname": "{name}_y"}
+
 
 @join_type
 def test_join(how, left, right, df1, df2):
-    expr = left.join(right, left.key == right.key, how=how)[
+    expr = left.join(right, left.key == right.key, how=how, **kws)[
         left, right.other_value, right.key3
     ]
     result = expr.compile()
@@ -53,7 +55,7 @@ def test_join(how, left, right, df1, df2):
 
 
 def test_cross_join(left, right, df1, df2):
-    expr = left.cross_join(right)[left, right.other_value, right.key3]
+    expr = left.cross_join(right, **kws)[left, right.other_value, right.key3]
     result = expr.compile()
     expected = dd.merge(
         df1.assign(dummy=1), df2.assign(dummy=1), how='inner', on='dummy'
@@ -67,7 +69,7 @@ def test_cross_join(left, right, df1, df2):
 
 @join_type
 def test_join_project_left_table(how, left, right, df1, df2):
-    expr = left.join(right, left.key == right.key, how=how)[left, right.key3]
+    expr = left.join(right, left.key == right.key, how=how, **kws)[left, right.key3]
     result = expr.compile()
     expected = dd.merge(df1, df2, how=how, on='key')[list(left.columns) + ['key3']]
     tm.assert_frame_equal(
@@ -77,7 +79,7 @@ def test_join_project_left_table(how, left, right, df1, df2):
 
 
 def test_cross_join_project_left_table(left, right, df1, df2):
-    expr = left.cross_join(right)[left, right.key3]
+    expr = left.cross_join(right, **kws)[left, right.key3]
     result = expr.compile()
     expected = dd.merge(
         df1.assign(dummy=1), df2.assign(dummy=1), how='inner', on='dummy'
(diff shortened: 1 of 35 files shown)


jcrist:
After thinking about this more (and talking with other Ibis maintainers), I think we need to have a default renaming scheme. Forcing users to pass in `rname` or `lname` explicitly whenever a duplicate column issue comes up is too onerous for common use cases. Lots of existing code works without the user ever realizing there's duplicate column issues, because the duplicate columns are immediately dropped:

```python
In [1]: import ibis

In [2]: t = ibis.table({"a": "int", "b": "int", "c": "int"})

In [3]: t2 = ibis.table({"a": "int", "b": "int", "d": "int"})

In [4]: expr = t.join(t2, "a")["a", "d"]  # `b` is duplicated, but since we immediately subselect, the issue never comes up
```

Removing the default renaming scheme would force users to pass in `rname` or `lname` above, even though their code never hits any of the renamed columns.

As such, I'm closing this in favor of #5883.

PR #5880, opened 2023-03-31
fix(duckdb): workaround upstreame bug when scanning numpy.str_ objects

### LINKED ISSUE #5879: bug: Numpy-generated random choices appear as NULLs
### What happened?

A Pandas data frame created by using `numpy.random.choice()`, when registered as a table in DuckDB backend, appears as all NULLs.

```python
import ibis
import numpy as np
import pandas as pd
import random

data = pd.DataFrame(
    {
        # works OK
        # "np_rand": np.random.choice(["foo", "bar"], p=[0.7, 0.3], size=10),
        # "core_rand": random.choices(["foo", "bar"], weights=[0.7, 0.3], k=10),

        "np_rand": [np.random.choice(["foo", "bar"], p=[0.7, 0.3]) for _ in range(5)],  # will be None in Ibis
        "core_rand": [random.choices(["foo", "bar"], weights=[0.7, 0.3])[0] for _ in range(5)],  # OK in Ibis
    }
)

if __name__ == "__main__":
    print("----- DF data -----")
    print(data)

    print()
    print("Numpy-generated column ends up as NULLs.")
    print("----- Ibis table data -----")
    conn = ibis.duckdb.connect()
    conn.register(data, table_name="data")
    t = conn.table("data")
    print(t.execute())

```

5.0.0

DuckDB

### Relevant log output

```sh
----- DF data -----
  np_rand core_rand
0     foo       foo
1     foo       foo
2     bar       bar
3     foo       foo
4     foo       foo

Numpy-generated column ends up as NULLs.
----- Ibis table data -----
  np_rand core_rand
0    None       foo
1    None       foo
2    None       bar
3    None       foo
4    None       foo
```

### Code of Conduct

cpcloud:
Fixes #5879.


ibis/backends/duckdb/__init__.py
```diff
@@ -437,9 +437,29 @@ def read_in_memory(
...
+
+            if not isinstance(t, pd.DataFrame):
+                return t
+
+            from ibis.backends.pandas.client import sch
+
+            schema = sch.infer(t)
+            return t.assign(
+                **{
+                    name: col.map(str, na_action="ignore")
```
jcrist, line 454: I think we technically only need to do this for columns that are inferred as `str` type if the input pandas type is an `object` column. The other new pandas str dtypes won't result in this bug, and avoiding the conversation in those cases would be slightly faster. Not sure if the extra complications are worth it though.

I also think we'll want to check for `None` here, rather than blindly calling `str`.

```suggestion
                    name: col.map(lambda x: x if x is None else str(x), na_action="ignore")
```

ibis/backends/duckdb/__init__.py
```diff
@@ -437,9 +437,29 @@ def read_in_memory(
         ir.Table
             The just-registered table
         """
+
+        def _clean_up_string_columns(
```
jcrist, line 441: Can you add a comment somewhere about what this is working around, and noting when it can be removed?

ibis/backends/duckdb/__init__.py
cpcloud: `na_action="ignore"` allows us to elide the `None`

ibis/backends/duckdb/__init__.py
cpcloud: Yup

cpcloud: Added a docstring.

jcrist (approved):
:shipit:

PR #5883, opened 2023-03-31
feat(api): remove `suffixes` arg to `join` in favor of `lname`/`rname`

### LINKED ISSUE #5537: Rethink `join` column deduplication

jcrist:
This PR removes the existing `suffixes` arg to all `join` methods in favor of individual `lname`/`rname` arguments. The default renaming scheme for handling duplicate column names has also changed to rename only duplicates in the right table by adding a `_right` suffix.

This is a breaking change - unfortunately there's no easy way to deprecate towards this behavior. To get the exact same behavior as before, you can pass in `lname="{name}_x", rname="{name}_y"` to join calls. This PR is targeted towards 6.0.

We choose this behavior over the existing one for a few reasons:

- Separating the arguments makes it easier to only rename columns in the left or right table.
- Using pattern strings rather than suffixes gives users more flexibility in handling duplicate column names. For example, they can use a prefix via `rname="right_{name}"`.
- Defaulting to only renaming duplicates in the right table lets joins compose better (e.g. `t1.left_join(t2, "id").left_join(t3, "id")`). Previously chained left joins may result in issues where the join keys are renamed in the first join (`id_x` and `id_y` in this example), and thus missing in the 2nd join. Renaming only the right duplicate columns lets joins like this work as expected.

Fixes #5537.


thegreymatter:
i wonder if there is something that can be done in 5.x to reduce the pain of this breaking change in 6.0.
for example -  add the option to add lname,rname and and raise deprecation warning when using suffixes.

kszucs:
Didn't take a thorough look yet just stumbled upon the removal of `docs/ibis-for-pandas-users.ipynb` notebook, it it intentional?

ibis/expr/types/relations.py
```diff
@@ -70,7 +69,9 @@ def f(  # noqa: D417
         | Sequence[
             str | tuple[str | ir.Column, str | ir.Column] | ir.BooleanValue
         ] = (),
-        suffixes: tuple[str, str] = ("_x", "_y"),
+        *,
+        lname: str = "",
+        rname: str = "{name}_right",
```
kszucs, line 74: I like it, it certainly provides more flexibility.

kszucs: > To get the exact same behavior as before, you can pass in lname="{name}_x", rname="{name}_y" to join calls.

Could we unpack a `suffixes` keyword argument if it is passed to the method?

```py
if 'suffixes' in kwargs:
    # raise deprecation warning
    lsuffix, rsuffix = suffixes
    lname, rname = "{name}" + lsuffix, "{name}" + rsuffix
```

If the workaround is that simple then I'd favor having a deprecation notice for one major release.

kszucs (commented):
The implementation looks good to me, but still need to walk through the test changes.

ibis/expr/types/relations.py
jcrist: Since the default behavior has changed, deprecating this way would work for users that were explicitly passing in `suffixes`, but would still be a breaking change for the majority of users that were relying on the default behavior.

Since we can't deprecate the change in default behavior, I'd rather this be a breaking change for all cases.

jcrist:
>  add the option to add lname,rname and and raise deprecation warning when using suffixes.

We could do this to ease the transition only for users who are already explicitly passing in `suffixes=`, but not for the majority of users who are relying on the default behavior. Two options:

- We could error nicely if `suffixes` is passed in. This would be something better than the default `TypeError` python raises when you pass in an invalid argument. I'm +0 on doing this. It's easy to do, but I'm not convinced it's necessary.
- We could warn if `suffixes` is passed in, and convert the suffixes to `lname`/`rname`. I'm -0.5 on doing this. It provides a smoother transition for users who were passing in `suffixes` explicitly. But since the default behavior has changed, this is still a breaking change for anyone not passing in suffixes explicitly. I personally feel it's cleaner to just break uniformly for all cases, rather than trying to smooth over just this one case.

That said, if anyone feels strongly on this I'm happy to do either of the above.

kszucs:
> > Didn't take a thorough look yet just stumbled upon the removal of docs/ibis-for-pandas-users.ipynb notebook, is it intentional?
> 
> It's not removed, I just stripped the output cells since we rerun this notebook when generating the docs. No need to store the output cells in git.

The UI fooled me 

image

kszucs (approved):
It looks good to me, thanks @jcrist!

@cpcloud would you like to take a look?

kszucs:
Could you please squash the chore commits following the feature commit?

ibis/expr/types/relations.py
```diff
@@ -2551,37 +2567,36 @@ def cross_join(
...
-        >>> pprint(expr.columns)
+        ┏━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━┓
+        ┃ species ┃ island    ┃ bill_length_mm ┃ bill_depth_mm ┃ flipper_length_mm ┃ … ┃
+        ┡━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━┩
+        │ string  │ string    │ float64        │ float64       │ int64             │ … │
+        ├─────────┼───────────┼────────────────┼───────────────┼───────────────────┼───┤
+        │ Adelie  │ Torgersen │           39.1 │          18.7 │               181 │ … │
+        │ Adelie  │ Torgersen │           39.5 │          17.4 │               186 │ … │
+        │ Adelie  │ Torgersen │           40.3 │          18.0 │               195 │ … │
+        │ Adelie  │ Torgersen │            nan │           nan │                 ∅ │ … │
```
gforsyth, line 2578:
```suggestion
        │ Adelie  │ Torgersen │            nan │           nan │              NULL │ … │
```

gforsyth (approved):
Micro-nit, but this looks good to me.

ibis/expr/types/relations.py
jcrist: The changes to the null repr haven't been merged to the 6.x.x branch yet; can't make this change until that happens.

jcrist:
> Could you please squash the chore commits following the feature commit?

Done

PR #5885, opened 2023-03-31
feat(sql): move to temporary views for `con.read_x` APIs to avoid view proliferation

### LINKED ISSUE #5881: feat(sql): move to temporary views for `con.read_x` APIs to avoid view proliferation

Not related to a problem.

Now that we're using a static connection pool, we should be able to use temporary views again to avoid pooping a bunch of random-looking views that persist in a database over time.

`master`

DuckDB

### Code of Conduct

cpcloud:
Fixes #5881.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -238,6 +238,12 @@ class Backend(BaseAlchemyBackend):
             f"please call one of {msg} directly"
         )
 
+    def _compile_temp_view(self, table_name, source):
+        raw_source = source.compile(
+            dialect=self.con.dialect, compile_kwargs=dict(literal_binds=True)
+        )
+        return f'CREATE OR REPLACE TEMPORARY VIEW "{table_name}" AS {raw_source}'
+
     @util.experimental
     def read_json(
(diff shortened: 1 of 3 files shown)


ibis/backends/duckdb/__init__.py
```diff
@@ -45,6 +45,15 @@ def normalize_filenames(source_list):
     return list(map(util.normalize_filename, source_list))
 
 
+def _compile_temp_view(table_name, source):
+    from duckdb_engine import Dialect as DuckDBDialect
+
+    raw_source = source.compile(
+        dialect=DuckDBDialect(), compile_kwargs=dict(literal_binds=True)
```
jcrist, line 52: If this was a method on the backend instead of a standalone function, you could reuse the dialect at `self.con.dialect`. Not sure how much that matters, just a note.

jcrist (commented):
Small note, otherwise :+1:

ibis/backends/duckdb/__init__.py
cpcloud: Oh, right. I think that makes more sense.

PR #5892, opened 2023-04-01
docs: add warning about comparisons to `None`

### LINKED ISSUE #5766: docs: raise warning or error on incorrect NULL check in 'having' clause
## Problem description

SQL and Python developers can be frustrated if they, by force of habit, use a `is None` or a similar expression in the `having` parameter of the `aggregate` method. In that case, they will get neither error nor warning, and the result will be most likely not what they intended.

Given this example code below, the `HAVING` clause in the generated SQL will be:
```sql
HAVING
  CAST(FALSE AS BOOLEAN)
```

The query will work and will return the result (an empty data frame) which was not what was intended.
If this clause is part of an `OR` expression, then the issue will be even harder to detect, as this clause will have no effect.
In addition, using a similar pattern in the `filter` method will raise an error. So this might be also a case of inconsistent behavior.

## Example code

```python
"""Demonstrate no warning and wrong results for `having=t["foo"].sum() is None` """

import ibis
from ibis import _

import pandas as pd

# -------- data --------
produce = ibis.memtable(
    pd.DataFrame({
        "category": ["fruit", "fruit", "berry", "veg"],
        "item": ["apple", "orange", "raspberry", "potato"],
        "price": [2.5, 3.0, 7.0, None],
    })
)

if __name__ == '__main__':
    expr = (
        produce
        # .filter(_["price"] is None)  # error is raised
        .filter(_["price"].isnull())  # works as intended
        .aggregate(
            _["price"].sum(),
            by="category",
            having=_["price"].sum() is None  # no error or warning, wrong results
            # having=_["price"].sum() == ibis.null  # no error or warning, wrong results
            # having=_["price"].sum().isnull()  # works as expected
        )
    )

    print(ibis.to_sql(expr))
    print(expr.execute())

```

An error or at least a warning is to be raised if the described pattern is used.is 

5.0.0

Pandas, but also observed the same in BigQuery.

### Code of Conduct

cpcloud:
Closes #5766.

### DIFF
--- a/ibis/expr/types/groupby.py
+++ b/ibis/expr/types/groupby.py
@@ -108,6 +108,8 @@ class GroupedTable:
     def having(self, expr: ir.BooleanScalar) -> GroupedTable:
         """Add a post-aggregation result filter `expr`.
 
+        !!! warning "Expressions like `x is None` return `bool` and **will not** generate a SQL comparison to `NULL`"
+
         Parameters
         ----------
         expr
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -853,6 +853,8 @@ class Table(Expr, _FixedTextJupyterMixin):
         having
             Post-aggregation filters. The shape requirements are the same
             `metrics`, but the output type for `having` is `boolean`.
+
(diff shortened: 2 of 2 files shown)


cpcloud:
Failing test is unrelated.

PR #5894, opened 2023-04-01
feat(api): scalar UDFs

cpcloud:
This is a PR to get some early feedback on the end-user API and implementation
of support scalar (vectorized and not) Python UDFs.

### TODOs

- [ ] More tests. Error handling for invalid calls for one, invalid types, and probably others.
- [x] ~Map out what the vectorized API looks like, see if existing code can be salvaged.~ Vectorized API is an attribute on the `@udf` decorator; not much code was salvaged.
- [x] ~Explore ways to get backend-and-udf-specific args/kwargs to the UDF without
   losing backend agnosticity at definition time.~ The decorator accepts keyword arguments that can be used for configuration.
- [x] ~Try implementing existing UDF support in terms of this API once we're happy with it.~ Postgres is re-implemented in the new API, and follow ups are listed below.

### Thing I don't like

1. There's a lot of indirection in `udfs.py`. This is basically a first draft, and can probably be simplified.

Feedback on any part of this is welcome.

### Example

Here's what it looks like to define and run a scalar UDF on the Snowflake backend:

```
In [19]: from ibis.interactive import *

In [20]: import os

In [21]: con = ibis.connect(os.environ["SNOWFLAKE_URL"])

In [22]: @udf
    ...: def num_vowels(s: str) -> int:
    ...:     return sum(s.count(v) for v in "aeiou")

In [23]: batting = con.tables.BATTING

In [24]: nvs = batting.group_by(num_vowels=lambda t: num_vowels(t.playerID)).agg(n=_.count())

In [25]: nvs
Out[25]:
┏━━━━━━━━━━━━┳━━━━━━━┓
┃ num_vowels ┃ n     ┃
┡━━━━━━━━━━━━╇━━━━━━━┩
│ int64      │ int64 │
├────────────┼───────┤
│          1 │  4986 │
│          5 │    75 │
│          3 │ 55085 │
│          2 │ 34656 │
│          4 │  6477 │
│          0 │    53 │
└────────────┴───────┘
```

### Move existing UDF APIs to the one defined here

- pandas
- dask
- pyspark
- bigquery (javascript and sql (maybe?))

### Add APIs for other UDF-supporting backends
- duckdb (only supported in 0.8.0)
- datafusion
- polars
- sqlite (?)


ibis/backends/snowflake/__init__.py
```diff
@@ -228,6 +231,53 @@ def normalize_name(name):
...
+                func = getattr(sa.func, type(op).__name__)
+                return func(*map(t.translate, op.args))
+
+            sql_blocks.append(self._compile_scalar_udf(udf_node))
+
+        with self.begin() as con:
+            con.exec_driver_sql("\n\n".join(sql_blocks))
+
+    def execute(self, expr, *args, **kwargs):
+        self._register_udfs(expr)
```
cpcloud, line 278: This should probably be moved to the base class.

ibis/backends/snowflake/__init__.py
```diff
@@ -242,6 +292,7 @@ def to_pyarrow(
         import pyarrow as pa
 
         self._register_in_memory_tables(expr)
+        self._register_udfs(expr)
```
cpcloud, line 295: Current unused, but ultimately necessary.

ibis/expr/rules.py
```diff
@@ -359,12 +359,17 @@ def output_dtype(self):
 
 
 @public
-def shape_like(name):
+def shape_like(*names):
```
cpcloud, line 362: Should probably extend `dtype_like` similarly.

pyproject.toml
```diff
@@ -83,8 +83,12 @@ shapely = { version = ">=1.6,<1.8|>=1.9,<3", optional = true }
 # appears to cause poetry's solver to get stuck
 #
 # also, we don't support arbitrarily old versions of this library
-snowflake-connector-python = { version = ">=2.7.10,<4", optional = true }
-snowflake-sqlalchemy = { version = ">=1.4.1,<2", optional = true }
+snowflake-connector-python = { version = ">=2.7.10,<4", optional = true, extras = [
+  "pandas",
+] }
+snowflake-sqlalchemy = { version = ">=1.4.1,<2", optional = true, extras = [
+  "pandas",
```
cpcloud, line 90: Temporary to silence the incredibly annoying snowflake warnings.

ibis/backends/snowflake/__init__.py
```diff
@@ -228,6 +231,53 @@ def normalize_name(name):
...
+        )
+
+        res = f"""\
+CREATE OR REPLACE TEMPORARY FUNCTION {name}({signature})
+RETURNS {snowflake_type}
+LANGUAGE python
+RUNTIME_VERSION = '3.8'
+HANDLER = '{name}'
+AS $$
+{source}
```
cpcloud, line 256: For snowflake's batch API we'll need to produce slightly different code, that looks like this:

```
...
HANDLER = 'wrapper'
...
AS $$

import _snowflake
import pandas as pd

{source}

@_snowflake.vectorized(input=pd.DataFrame)
def wrapper(df):
    return {name}(*(col for _, col in df.items()))
$$
```

This is so that a UDF's signature is compatible with both scalar and batched.

ibis/backends/snowflake/__init__.py
```diff
@@ -228,6 +231,53 @@ def normalize_name(name):
...
+        signature = ", ".join(
+            f"{name} {sa.types.to_instance(to_sqla_type(dialect, arg.output_dtype)).compile(dialect=dialect)}"
+            for name, arg in zip(udf_node.argnames, udf_node.args)
+        )
+        snowflake_type = sa.types.to_instance(
+            to_sqla_type(dialect, udf_node.output_dtype)
+        ).compile(dialect=dialect)
+        source = textwrap.dedent(inspect.getsource(udf_node.__func__)).strip()
+        source = "\n".join(
+            line for line in source.splitlines() if not line.startswith("@")
```
cpcloud, line 246: Decorators show up in the source, but we don't want the `@udf` one to show up because it's coming from ibis.

We should probably _just_ ignore that one and not all decorators.

ibis/expr/rules.py
```diff
@@ -359,12 +359,17 @@ def output_dtype(self):
...
     @attribute.default
     def output_shape(self):
-        args = getattr(self, name)
-        args = args if util.is_iterable(args) else [args]
-        return highest_precedence_shape(args)
+        input_args = []
+        for name in names:
+            args = getattr(self, name)
+            if util.is_iterable(args):
```
kszucs, line 368: Could you please implement it directly in the op since it is required for a single operation only.

ibis/expr/operations/udfs.py
```diff
@@ -0,0 +1,109 @@
...
+from public import public
+
+import ibis.expr.datatypes as dt
+import ibis.expr.operations as ops
+import ibis.expr.rules as rlz
+from ibis.common.annotations import attribute
+
+
+@public
+class ScalarUDFNode(ops.Value):
```
kszucs, line 18: Can you remove the `Node` postfix?

kszucs: What does `ScalarUDF` mean?

We have another nomenclature `ElementWise` in the `ibis/expr/operations/vectorized.py` file. It would be nice to unify the operations over there and here.

ibis/expr/operations/udfs.py
```diff
@@ -0,0 +1,109 @@
...
+    @abc.abstractmethod
+    def make_vectorized_wrapper(self, fn, *args, **kwargs):
+        ...
+
+    @abc.abstractmethod
+    def make_wrapper(self, fn, *args, **kwargs):
+        ...
+
+
+class ScalarUDF(UDF):
```
kszucs, line 53: This seems to be analogous with `UserDefinedFunction` defined in `ibis/expr/udf/vectorized.py`, can we unify them?

cpcloud: It's similar, but I think this whole thing needs to be rebuilt. It's pretty scattered across a bunch the backends, as opposed to having an interface that each backend can implement separately.

ibis/expr/rules.py
cpcloud: Done.

ibis/backends/snowflake/__init__.py
cpcloud: Done.

ibis/backends/postgres/__init__.py
```diff
@@ -184,53 +194,102 @@ def list_databases(self, like=None):
...
-        out_type,
-        schema=None,
-        replace=False,
-        name=None,
-        language="plpythonu",
-    ):
-        """Decorator that defines a PL/Python UDF in-database.
+    def create_function(self, name: str, *, schema: str | None = None) -> Callable:
+        query = f"""
+SELECT
```
cpcloud, line 199: This query is a slightly modified version of what comes out of running

```
\df some_schema.*
```

in `psql`.

To see the query you need to start `psql` with `-E`

ibis/backends/postgres/__init__.py
```diff
@@ -184,53 +194,102 @@ def list_databases(self, like=None):
...
+        fake_func.__annotations__ = {"return": return_type, **dict(signature)}
+        op = ops.udf.scalar(fake_func, schema=schema, __is_server_defined__=True)
+        return op
+
+    def _get_udf_source(self, udf_node: ops.SingleScalarUDF):
+        config = udf_node.__config__["kwargs"]
+        immutable = config.get("immutable", False) * "IMMUTABLE"
+        strict = config.get("strict", False) * "STRICT"
+        leakproof = config.get("leakproof", False) * "LEAKPROOF"
+        parallel = config.get("parallel", "UNSAFE").upper()
```
cpcloud, line 256: I may kill all of these options, they're kind of noisy and I am not sure what their effect on performance is. In theory enabling them improves performance but adding these at the start might be overkill.

ibis/common/exceptions.py
```diff
@@ -111,7 +112,16 @@ def __init__(self, name: str) -> None:
         self.name = name
 
     def __str__(self) -> str:
-        return f"No function found with name {self.name}"
+        return f"No user-defined function found with name {self.name}"
+
+
+class AmbiguousUDFError(DDLError):
```
cpcloud, line 118: Probably should be a subclass of `IbisUDFError`

ibis/expr/datatypes/core.py
```diff
@@ -59,6 +59,8 @@ def dtype_from_object(value, **kwargs) -> DataType:
                 raise TypeError(
                     f"Cannot construct an ibis datatype from python type `{value!r}`"
                 )
+        elif value is Any:
+            return json
```
cpcloud, line 63: Investigate if this is still necessary and document why

ibis/expr/datatypes/cast.py
```diff
@@ -186,6 +186,8 @@ def can_cast_array_or_set(
 
 
 @castable.register(dt.JSON, dt.JSON)
+@castable.register(dt.Floating, dt.JSON)
+@castable.register(dt.Integer, dt.JSON)
```
cpcloud, line 190: Investigate if this is still necessary and document why

ibis/backends/base/__init__.py
```diff
@@ -259,6 +259,8 @@ def to_pyarrow(
             A pyarrow table holding the results of the executed expression.
         """
         pa = self._import_pyarrow()
+        self._register_udfs(expr)
+        self._register_in_memory_tables(expr)
```
jcrist, line 263: We now have `_run_pre_execute_hooks` (currently in the base SQL backend, but not the base backend, but that'd be an easy fix), where backends can add hooks like this. We might move `_register_udfs` to be called from there in the base implementation. https://github.com/ibis-project/ibis/blob/ea1529d14aec891974076c4b093ac6c7c6bc9854/ibis/backends/base/sql/__init__.py#L263

ibis/backends/base/__init__.py
```diff
@@ -713,6 +715,10 @@ def register_options(cls) -> None:
             except ValueError as e:
                 raise exc.BackendConfigurationNotRegistered(backend_name) from e
 
+    @abc.abstractmethod
+    def _register_udfs(self, expr: ir.Expr):
```
jcrist, line 719: An `abstractmethod` requires a subclass to implement it, and raises an error if it's unimplemented. We might instead raise an appropriate `NotImplementedError` (or equivalent) in the base implementation, rather than forcing all subclasses (even those not supporting UDFs) to implement this.

ibis/backends/base/sql/__init__.py
```diff
@@ -198,6 +196,12 @@ def to_pyarrow_batches(
 
         return pa.ipc.RecordBatchReader.from_batches(schema.to_pyarrow(), batches)
 
+    def _compile_udfs(self, expr: ir.Expr):
+        pass
+
+    def _register_udfs(self, expr: ir.Expr):
+        pass
```
jcrist, line 203: Should these raise a `NotImplementedError` (or fallback to the above) rather than a silent `pass`? Won't this silently pass through for backends that don't support UDFs?

Also could use a docstring on both of these.

ibis/backends/postgres/__init__.py
```diff
@@ -184,53 +194,102 @@ def list_databases(self, like=None):
...
-        pyfunc,
-        in_types,
-        out_type,
-        schema=None,
-        replace=False,
-        name=None,
-        language="plpythonu",
-    ):
-        """Decorator that defines a PL/Python UDF in-database.
+    def create_function(self, name: str, *, schema: str | None = None) -> Callable:
```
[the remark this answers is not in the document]
jcrist: Agreed, I think `.function` is a better method name.

ibis/backends/postgres/__init__.py
jcrist: Keeping things simpler for now makes sense. Can always add them back later if/when someone asks for it.

ibis/common/exceptions.py
```diff
@@ -127,3 +137,27 @@ def _mark_as_unsupported(self):
...
+
+
+class MissingParameterAnnotationError(IbisUDFError):
+    def __str__(self):
+        fn, param = self.args
+        signature = f"{fn.__qualname__}{inspect.signature(fn)}"
+        return f"parameter `{param}` in function `{signature}` is missing a type annotation"
+
+
+class InvalidDecoratorError(IbisUDFError):
```
jcrist, line 160: Do we need to split these exceptions out into 3 different types. There's both a maintenance and user-understanding cost ("which exception should I be catching here?")  to having deep exception hierarchies.

If I was to do this I'd just use the base `IbisUDFError` everywhere (or maybe make a single subclass `IbisUDFDefinitionError` if there are non-signature-related errors that also need to be raised). I'd also perform the str formatting at exception instantiation time rather than in the `__str__` so only a str is attached to the instance. This is definitely part personal preference, but I prefer my exceptions to be simple data containers. Some exception tooling (for e.g. wrapping/chaining/serializing exceptions) also reasonably expects all exceptions to take a single str as their first argument - messing with this expectation can break things in hard to debug ways.

ibis/expr/operations/udf.py
```diff
@@ -0,0 +1,134 @@
...
+    if fn is None:
+        return lambda fn: functools.update_wrapper(wrapper(fn, *args, **kwargs), fn)
+    return functools.update_wrapper(wrapper(fn, *args, **kwargs), fn)
+
+
+S1 = TypeVar("S1", bound=ScalarUDF)
+S2 = TypeVar("S2", bound=ScalarUDF)
+
+
+class UDF(abc.ABC):
```
jcrist, line 55: This class could use some docstrings (at least one-line descriptions).

jcrist: Since we only have the `Scalar` subclass, could we avoid creating an ABC for now? I suspect whenever we add a second UDF type we'll have to amend this abstraction in some way anyway. Not a big deal though, up to you.

ibis/expr/operations/udf.py
```diff
@@ -0,0 +1,134 @@
...
+            fields[name] = arg
+
+        fields["output_dtype"] = dt.dtype(return_annotation)
+
+        fields["__func__"] = fn
+        fields["__config__"] = FrozenDict(args=args, kwargs=FrozenDict(**kwargs))
+        fields["__namespace__"] = kwargs.get("schema")
+        fields["__is_server_defined__"] = kwargs.get("__is_server_defined__", False)
+
+        return type(fn.__name__, (base,), fields)
```
jcrist, line 102: IIUC this is defining a new subclass for every UDF. I'm a bit worried about this approach for the following reasons:

- Dynamically defined types won't serialize (easily), preventing the operation tree from being pickled using `pickle`. This may not be important now, but if it was in the future then this approach would be a blocker.
- It's a bit complex. It took me a few readings of this implementation to understand all the plumbing here.

Would it be possible to avoid creating dynamic subclasses, and instead plumb through the function info as attributes on *instances* of the various operations, instead of storing them as attributes on dynamically generated classes? This would mean that all scalar vectorized UDF calls would result in instances of `VectorizedScalarUDF`, rather than instances of custom subclasses for each `UDF` definition.

jcrist (commented):
Overall this looks pretty good to me!

ibis/expr/operations/udf.py
cpcloud: The primary issue with this is that there isn't way to hook into the validation system without creating fields on a class.

All of the machinery to validate inputs is created based on a class definition and can't really be moved to runtime without weird hacks AFAICT.

cpcloud: So, while I can plumb almost everything through as instance attributes, the signature is unique to an operation class.

cpcloud: The other thing to note is that validation happens during operation construction, so constructing an instance that doesn't validate anything (because the signature would have to live elsewhere) would just bypass all validation.

ibis/common/exceptions.py
cpcloud: This particular way of constructing exceptions is very deliberate.

It's based on my experience working with systems that are a mix of Rust and Python.

Based this experience, I've come to the conclusion that the typical way that we author (effectively no useful information beyond a string) and raise (raising incredibly generic exception types that again do nothing to help aid debugging `ValueError`, `TypeError`) exception types in Python results exceptions that are effectively impossible to debug without a debugger. It _is_ possible to better understand and debug an exception **without** the crutch of a debugger.

The way that is accomplished (at least to start) is to have more specific error types, that are capable of constructing error messages from a set of simple inputs (that aren't trivially a formatted string with the error).

[This blog post](https://kazlauskas.me/entries/errors) by a former colleague of mine captures the idea well, and shows how the same problem exists in Rust (and other languages besides Python!) and how to approach improvements. Notice that the specificity also helps with maintenance: you know what kind of thing went wrong at the call site by reading the exception type, rather than having to plumb through some format string.

This kind of approach is not without precedent in Python itself: the various OSError subclasses are an example of this approach.

> "which exception should I be catching here?"

I think the user-understanding part of this is a bit of a red herring. The exceptions we're talking about here are not something any end user of ibis should catch, they are problems to address (bugs in ibis, incorrect input from the user etc). Now, a library author might want to catch these, but IMO the situation is an improvement: you can now handle exceptions with more specificity. The tradeoff is, yeah, you have to know what those more specific exceptions are.

> Some exception tooling (for e.g. wrapping/chaining/serializing exceptions) also reasonably expects all exceptions to take a single str as their first argument.

I guess we'll just have to disagree here :) I don't think this is a reasonable assumption. There's nothing special about an exception that dictates this other than convention IIUC.

Now, all of this isn't without tradeoffs and there's almost certainly some level of specificity beyond which there's no additional utility.

I can give this another think through to see if there's a way to reduce the hierarchy while bringing in some good ideas from other communities.

ibis/expr/operations/udf.py
cpcloud: Yep, that makes sense.

ibis/backends/base/__init__.py
cpcloud: Refactored this and moved the methods to the base class.

ibis/backends/base/__init__.py
cpcloud: Remove the abc and made methods raise `NotImplementedError`.

ibis/backends/base/sql/__init__.py
cpcloud: Added docstrings, and now raising `NotImplementedError`.

ibis/expr/operations/udf.py
cpcloud: Removed the `ABC`

ibis/backends/postgres/__init__.py
cpcloud: Removed.

ibis/backends/postgres/__init__.py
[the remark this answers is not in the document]
cpcloud: Changed to `.function`.

ibis/expr/datatypes/core.py
cpcloud: This isn't necessary, snowflake users will have to annotation with `json` explicitly for now.

ibis/backends/base/sql/__init__.py
```diff
@@ -206,6 +204,13 @@ def to_pyarrow_batches(
 
         return pa.ipc.RecordBatchReader.from_batches(schema.to_pyarrow(), batches)
 
+    def _compile_udfs(self, expr: ir.Expr) -> Iterator[str]:
+        """Return an iterator of DDL strings, once for each UDFs contained within `expr`."""
+        if self.supports_python_udfs:
+            raise NotImplementedError(self.name)
+        return
+        yield
```
cpcloud, line 212: This is expected to return a generator, and for backends that don't support UDFs it's a no-op. This appears to be one way to do it.

I suppose `return iter(())` could also work 🤷🏻

cpcloud: I think we should make this method an abstract method on a mixin instead of these odd configuration settings and runtime checks.

I'll do that in a follow-up though.

cpcloud:
@jcrist This is ready for review.

I'll update the description with follow-ups and open issues where it makes sense.

ibis/expr/operations/udf.py
```diff
@@ -0,0 +1,157 @@
...
+    def output_shape(self):
+        return rlz.highest_precedence_shape(
+            itertools.chain.from_iterable(
+                util.promote_list(getattr(self, name)) for name in self.argnames
+            )
+        )
+
+
+@public
+class SingleScalarUDF(ScalarUDF):
```
kszucs, line 35: `ElementwiseScalarUDF`?

ibis/expr/operations/udf.py
```diff
@@ -0,0 +1,157 @@
...
+    if fn is None:
+        return lambda fn: functools.update_wrapper(wrapper(fn, *args, **kwargs), fn)
+    return functools.update_wrapper(wrapper(fn, *args, **kwargs), fn)
+
+
+S1 = TypeVar("S1", bound=ScalarUDF)
+S2 = TypeVar("S2", bound=ScalarUDF)
+
+
+class UDF:
```
kszucs, line 55: Could this code live in the `builders.py`?

ibis/expr/operations/udf.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def _make_vectorized_wrapper(
+        self, fn: Callable, *args: Any, **kwargs: Any
+    ) -> Callable:
+        """Generate an ibis operation node type wrapper for a vectorized scalar function."""
+
+    def _make_wrapper(self, fn: Callable, *args: Any, **kwargs: Any) -> Callable:
+        """Generate an ibis operation node type wrapper for a simple scalar function."""
+
+    def make_node(self, fn: Callable, base: type[S1], *args, **kwargs) -> type[S2]:
```
kszucs, line 102: The previous udf implementation looks like the following:

```py
class VectorizedUDF(Value):
    func = rlz.instance_of((FunctionType, LambdaType))
    func_args = rlz.tuple_of(rlz.column(rlz.any))
    input_type = rlz.tuple_of(rlz.datatype)
    return_type = rlz.datatype

    @property
    def output_dtype(self):
        return self.return_type
```

What's the advantage of the following node factory compared to that? 
Does it do an extra layer of validation?

ibis/backends/base/sql/__init__.py
jcrist: Is there a good reason to *require* this method return a generator (rather than leaving that backend implementation dependent)? If not, I'd change this to `Iterable[str]` and return [] by default. No change should be needed downstream consumers.

jcrist: Alternatively, do we expect the output of `list(_compile_udfs(...))` to be large enough that using a generator for memory reasons makes sense? Otherwise it may be nicer to always return a `list[str]`. This would remove the need for calling `consume` below.

ibis/backends/base/sql/__init__.py
```diff
@@ -344,6 +345,7 @@ def compile(
             The output of compilation. The type of this value depends on the
             backend.
         """
+        util.consume(self._compile_udfs(expr))
```
jcrist, line 348: Why are we compiling the UDFs here (and ignoring the results)? Just so we error here if the UDF can't be compiled?

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -698,6 +705,31 @@ def insert(
                 f"The given obj is of type {type(obj).__name__} ."
             )
 
+    def _compile_udfs(self, expr) -> Iterator[str]:
+        for udf_node in an.find_scalar_udfs(expr.op()):
+
+            @self.add_operation(type(udf_node))
```
jcrist, line 711: Could we avoid doing `add_operation` for each of these types, and instead rely on a single handler on the `ScalarUDF` base class registered statically in the backend that handles the operation, possibly exactly as you have below? Having a side effect in `_compile_udfs` feels wrong to me.

ibis/interactive.py
```diff
@@ -2,7 +2,8 @@
 import ibis.examples as ex
 from ibis import deferred as _
 from ibis import selectors as s
+from ibis.expr.operations import udf
```
jcrist, line 5: Do we also want to expose this as `ibis.udf` so users can access this as `ibis.udf.scalar`?

ibis/backends/base/sql/__init__.py
cpcloud: What if you want to call `.alias` on an expression that uses a UDF? How would an engine be able to infer the types of the query without knowing about UDFs?

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: Yeah, that should work!

ibis/expr/operations/udf.py
```diff
@@ -0,0 +1,157 @@
...
+from ibis.common.collections import FrozenDict
+
+if TYPE_CHECKING:
+    import ibis.expr.types as ir
+
+
+@public
+class ScalarUDF(ops.Value):
+    @attribute.default
+    def output_shape(self):
```
kszucs, line 26: Is it equivalent with `rlz.shape_like("args")`?

ibis/interactive.py
cpcloud: Definitely, I will do that in a follow up.

ibis/backends/base/sql/__init__.py
cpcloud: Ah right, this should be unnecessary by registering the generic `ScalarUDF` dispatcher as you describe below. That's the reason why I was doing it here!

cpcloud: This will have to stay in light of rule lookup not taking into account inheritanc.e

ibis/expr/analysis.py
```diff
@@ -770,6 +770,13 @@ def find_subqueries(node: ops.Node, min_dependents=1) -> tuple[ops.Node, ...]:
     )
 
 
+def find_scalar_udfs(node: ops.Node):
```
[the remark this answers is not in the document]
cpcloud: Yep, thanks!

ibis/backends/base/sql/__init__.py
cpcloud: It seems wasteful to return a `list` when all I'm doing is iterating over it. The implementation is not meaningfully simpler with a list IMO except in the UDFs-not-supported case.

That said, definitely willing to change the types and simplify the udfs-not-supported case.

ibis/expr/operations/udf.py
cpcloud: I think so!

ibis/expr/operations/udf.py
cpcloud: > Does it do an extra layer of validation?

It's not an additional layer, but baked into how we're defining each node.

The inputs are not an arbitrary collection of expressions, they are built using rules based on the user's type annotations. This means the UDFs inputs are validated as much (or as little) as the user wants.

ibis/expr/operations/udf.py
```diff
@@ -0,0 +1,149 @@
...
+    @util.experimental
+    def __call__(
+        self, fn: Callable | None = None, *args: Any, **kwargs: Any
+    ) -> Callable:
+        """Construct a scalar user-defined function.
+
+        Examples
+        --------
+        >>> from ibis.interactive import *
+        >>> @udf.scalar
```
NickCrews, line 59: I would prefer this to be explicitly elementwise, I don't like the bare @udf.scalar

@udf.scalar.elementwise
and
@udf.scalar.vectorized

of course I could see someone wanting .vectorized to either use pandas or pyarrow, so following this pattern would allow for extension, eg
- @udf.scalar.elementwise
- @udf.scalar.pandas
- @udf.scalar.arrow

NickCrews:
EDIT: after looking through the code, I think some of this is addressed

As to UX, can we make it so that you can simply do `batting.group_by(num_vowels(_.playerID)).agg(n=_.count())`?
- The name of the resulting expression ("num_vowels") is determined from the function name
- By applying the @udf decorator, now our function is lazily computed, because the raw python function would barf if it were handed the ibis.Deferred

I would also love it so I could turn the python function into a UDF later, eg
```python

def num_vowels(s: str) -> int:
    return sum(s.count(v) for v in "aeiou")

batting.group_by(num_vowels=lambda t: udf(num_vowels(t.playerID))).agg(n=_.count())
```
I assume this already works as is, but I just wanted to make it explicit as a use case.

ibis/expr/operations/udf.py
cpcloud: Not opposed to `@udf.scalar.python` and the others. I would like to avoid too much `.`-ing since it can get confusing to know in what order things should be spelled when there are a lot of dots necessary.

ibis/expr/operations/udf.py
cpcloud: Maybe, but I think I can do that in a follow-up too.

cpcloud:
Thanks for the feedback @NickCrews!

> The name of the resulting expression ("num_vowels") is determined from the function name

This is already the case:

```

```

If you're asking to simply make the name be `num_vowels`, I don't think we should do that. You then can't have two calls to `num_vowels` without renaming.

> I would also love it so I could turn the python function into a UDF later

You should be able to leverage the way decorators work for this:

```
def num_vowels(s: str) -> int:
    return sum(s.count(v) for v in "aeiou")

batting.group_by(num_vowels=lambda t: udf(num_vowels)(t.playerID)).agg(n=_.count())
```

> Or, maybe even better would be to be able to decorate the function, and then afterwards you can still pass in the normal types as well:

I don't think this is feasible. How would ibis know whether to execute the underlying python function directly or to promote inputs to an expression and return an expression?

ibis/backends/base/sql/__init__.py
cpcloud: Types fixed, udfs-not-supported case simplified.

cpcloud:
Merging!

ibis/expr/operations/udf.py
cpcloud: I was able to implement these:

```
@udf.scalar.python
@udf.scalar.pandas
@udf.scalar.pyarrow
```

PR #5905, opened 2023-04-04
fix(sql): generate consistent `pivot_longer` semantics in the presence of multiple `unnest`s

### LINKED ISSUE #5855: bug: `pivot_longer` returns the wrong result in the snowflake backend
### What happened?

`pivot_longer` expects `t.select(unnest(), unnest())` to have pairwise unnest semantics, but the snowflake backend is compiling each unnest to a cross-join-unnest. This results in duplicate rows.

Snowflake doesn't really have a straightforward way to do this, but since its collections aren't strongly typed, an array of length-2 arrays could be built from the unnest arguments and then the outer array can be unnested.

`master`

Snowflake

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #5855.

This PR fixes an issue caused by the difference in behavior between multiple
unnest calls in a projection.

In the duckdb and postgres backends

```
t.select(t.x.unnest(), t.y.unnest())
```

is analogous to Python's zip, and will perform a pairwise unnest.

In the snowflake and trino backends, we desugar that into a series of cross joins
which generates the Cartesian product of the unnest calls.

This PR doesn't take a position on that, but fixes `pivot_longer` to have the
desired pairwise behavior regardless of backend.

We **do** need to address the difference in behavior I mention, but that should
be done in 6.0.

I created https://github.com/ibis-project/ibis/issues/5913 to discuss options.

### DIFF
--- a/ibis/backends/base/sql/alchemy/datatypes.py
+++ b/ibis/backends/base/sql/alchemy/datatypes.py
@@ -13,6 +13,7 @@ from sqlalchemy.ext.compiler import compiles
 import ibis.expr.datatypes as dt
 import ibis.expr.schema as sch
 from ibis.backends.base.sql.alchemy.geospatial import geospatial_supported
+from ibis.common.collections import FrozenDict
 
 if geospatial_supported:
     import geoalchemy2 as ga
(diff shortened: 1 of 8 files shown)


cpcloud:
Ok, after fighting the battle of `unnest` and `row` types, I think this may actually work!

cpcloud:
Moved this back to 5.1, I think this is fine to go in before 6.0

ibis/backends/base/sql/alchemy/datatypes.py
```diff
@@ -29,10 +30,12 @@ def compiles_array(element, compiler, **kw):
...
 
 class StructType(sat.UserDefinedType):
+    cache_ok = True
+
     def __init__(self, fields: Mapping[str, sat.TypeEngine]) -> None:
-        self.fields = {
-            name: sat.to_instance(type) for name, type in dict(fields).items()
-        }
+        self.fields = FrozenDict(
+            **{name: sat.to_instance(type) for name, type in dict(fields).items()}
```
kszucs, line 37: No need to unpack it.

cpcloud: Fixed.

kszucs (approved):
LGTM

PR #5914, opened 2023-04-05
fix(snowflake): make sure ephemeral tables following backend quoting rules

cpcloud:
This PR fixes an issue where we were not threading the quoting parameters through to the various places where ephemeral tables are being created. This manifests mostly in the snowflake backend due to broken quoting behavior in `snowflake-sqlalchemy`.

### DIFF
--- a/ibis/backends/base/sql/alchemy/query_builder.py
+++ b/ibis/backends/base/sql/alchemy/query_builder.py
@@ -96,8 +96,11 @@ class _AlchemyTableSetFormatter(TableSetFormatter):
             result = ref_op.sqla_table
         elif isinstance(ref_op, ops.UnboundTable):
             # use SQLAlchemy's TableClause for unbound tables
-            result = sa.table(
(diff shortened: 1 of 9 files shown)


ibis/backends/base/sql/alchemy/query_builder.py
```diff
@@ -96,8 +96,11 @@ def _format_table(self, op):
             result = ref_op.sqla_table
         elif isinstance(ref_op, ops.UnboundTable):
             # use SQLAlchemy's TableClause for unbound tables
-            result = sa.table(
-                ref_op.name, *translator._schema_to_sqlalchemy_columns(ref_op.schema)
+            result = sa.Table(
```
cpcloud, line 99: Unfortunately the lighter-weight `sa.table` doesn't accept a `quote` argument so we can't continue to use it here.

kszucs (approved):

PR #5915, opened 2023-04-05
ci(pyspark): be more specific with tested versions

cpcloud:
Otherwise we can end up with solve that incorrectly passes tests


cpcloud:
Hm, maybe pyspark _does_ support some versions of numpy 1.24. Investigating now.

cpcloud:
Ah, the problem is that some tests don't hit the failing code path 😬.

PR #5916, opened 2023-04-05
ci(pyspark): fix numpy 1.24 ci

cpcloud:
The issue is that pyspark does not always hit the code path that causes it to break with numpy 1.24, so here I just skip pyspark tests if numpy >= 1.24 is installed.

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -439,7 +439,7 @@ jobs:
         run: docker compose logs
 
   test_pyspark:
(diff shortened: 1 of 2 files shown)


cpcloud:
Actually, if we can't xfail easily with pyspark and numpy 1.24, I'm going to remove that from the build matrix. It's just wasting time at that point.

cpcloud:
Besides, PySpark 3.3.3 [will support numpy 1.24](https://issues.apache.org/jira/browse/SPARK-40376)

PR #5919, opened 2023-04-05
fix(api): disambiguate attribute errors from a missing `resolve` method versus from an invalid attribute access _during_ resolution

### LINKED ISSUE #5918: bug: filtering with `_` erroneously "works" with non-existent columns
### What happened?

I think this is probably related to some existing discussions on the semantics of the `_` API, but I have an example that I think is particularly tricky. The following should throw an error:

```python
from ibis import _, examples, show_sql

t = examples.aids.fetch()
# the column "bogus" does not exist
result = t.filter([_.year != _.bogus])
show_sql(result)
```
shows that no filtering is happening:
```sql
SELECT
  t0.year,
  t0.quarter,
  t0.delay,
  t0.dud,
  t0.time,
  t0.y
FROM ibis_read_csv_29 AS t0
```

If I reference two columns that actually exist, then the filter behaves as expected.

5.0.0

I think this would be relevant for all backends

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #5918.


NickCrews:
Looks great, thanks!

gforsyth (approved):
:shipit:

cpcloud:
For posterity, the reason this was able to make it though the query was due to our default filtering out of any object that isn't an `ops.Node` during `ibis.common.graph.traverse` calls.

cpcloud:
That particular issue should be fixed at the operation node level, ideally using a rule. This would prevent invalid inputs from going into the `ops.Selection` constructor.

PR #5935, opened 2023-04-06
fix(duckdb): drop use of experimental parallel csv reader

jcrist:
I just spent the last 30 minutes debugging why a query would succeed in the duckdb CLI, but fail in ibis - ended up tracking the issue down to the use of the parallel csv reader.

For now I'm removing the use of this option by default. If needed, users can always manually reenable it. We also heard from the duckdb team that this option will be going away next release anyway.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -120,8 +120,6 @@ class Backend(BaseAlchemyBackend):
             Path(temp_directory).mkdir(parents=True, exist_ok=True)
             config["temp_directory"] = str(temp_directory)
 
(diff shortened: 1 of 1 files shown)


cpcloud (approved):

jcrist:
For reference, the issue I found had to do with usage of the `skip` option to skip some header rows in a file. I've [filed an upstream issue](https://github.com/duckdb/duckdb/issues/6994) to track (see the linked issue for a failing example). Since the duckdb team has already been working on improving the parallel reader there's a decent chance this is already fixed upstream, but I haven't had a chance to rebuild duckdb to check.

PR #5952, opened 2023-04-09
feat(api): add a bunch of array functionality

cpcloud:
Add a bunch of array operations common to backends that support them.


ibis/backends/clickhouse/__init__.py
```diff
@@ -260,10 +259,21 @@ def _normalize_external_tables(self, external_tables=None):
...
-            external_tables_list.append(
-                dict(name=name, data=df.to_dict("records"), structure=structure)
+            structure = list(
+                zip(
+                    schema.names,
+                    map(
+                        serialize,
+                        (
+                            # unwrap nested structures
+                            typ.copy(nullable=not typ.is_nested())
```
kszucs, line 269: Can you add a bit more context why is it needed here?

ibis/expr/operations/arrays.py
```diff
@@ -135,3 +135,54 @@ def output_dtype(self):
...
+class ArrayContains(Value):
+    arg = rlz.array
+    other = rlz.any
+
+    output_dtype = dt.boolean
+    output_shape = rlz.shape_like("args")
+
+
+@public
+class ArrayPosition(Value):
```
kszucs, line 150: Shall we rename ArrayIndex to ArreyGet and use ArrayIndex here instead? That would better align with the python semantics and the method naming we use on the expr level

ibis/expr/types/arrays.py
```diff
@@ -471,6 +471,294 @@ def filter(self, predicate: Callable[[ir.Value], ir.BooleanValue]) -> ir.ArrayVa
...
+        ├──────────────────────┤
+        │ [2, 3]               │
+        │ []                   │
+        │ [42, 42]             │
+        │ NULL                 │
+        └──────────────────────┘
+        """
+        return ops.ArraySort(self).to_expr()
+
+    def union(self, other: ir.ArrayValue) -> ir.ArrayValue:
```
kszucs, line 711: I wonder what should we do with the set datatype

cpcloud: Maybe we should remove it?

ibis/backends/clickhouse/__init__.py
cpcloud: Yep. Will do.

cpcloud: Done.

ibis/expr/types/arrays.py
```diff
@@ -471,6 +471,294 @@ def filter(self, predicate: Callable[[ir.Value], ir.BooleanValue]) -> ir.ArrayVa
...
+        └────────────────────────┘
+        >>> t.arr.index(800)
+        ┏━━━━━━━━━━━━━━━━━━━━━━━━━┓
+        ┃ ArrayPosition(arr, 800) ┃
+        ┡━━━━━━━━━━━━━━━━━━━━━━━━━┩
+        │ int64                   │
+        ├─────────────────────────┤
+        │                      -1 │
+        │                      -1 │
+        │                      -1 │
```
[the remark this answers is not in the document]
cpcloud: Yes. Take a look at the test for this operation.

kszucs (approved):

PR #5957, opened 2023-04-10
docs(selectors): make doctests more complete and actually run them

cpcloud:
Make our selectors docstring examples actually run with doctests and make them a bit more complete.

### DIFF
--- a/ibis/conftest.py
+++ b/ibis/conftest.py
@@ -1,3 +1,4 @@
+import builtins
 import os
 
 import pytest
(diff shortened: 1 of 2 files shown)


ibis/conftest.py
```diff
@@ -6,11 +7,16 @@
...
     monkeypatch.setitem(os.environ, "NO_COLOR", "1")
     # reset interactive mode to False for doctests that don't execute
     # expressions
     ibis.options.interactive = False
+    # workaround the fact that doctests include everything in the tested module
+    # for selectors we have an `all` function
+    #
+    # the clash doesn't really pop up in practice, but we can rename it to
+    # `all_` in 6.0 if desired
+    doctest_namespace["all"] = builtins.all
```
jcrist, line 22: I think the naming is fine, but that this messes with doctests at all is gross.

jcrist (approved):

ibis/conftest.py
cpcloud: Yeah, 🤷🏻 doctests are fun!

PR #5958, opened 2023-04-10
docs(comparison): add a "why ibis" page

gforsyth:


docs/why_ibis.md
```diff
@@ -0,0 +1,65 @@
...
+  - footer
+---
+
+# Why try Ibis?
+
+Ibis is a dataframe interface to execution engines with support for [15+
+backends](./backends/index.md). Ibis doesn't replace your existing execution
+engine, it _extends_ it with powerful abstractions and intuitive syntax.
+
+You are probably already working with one or more `$BIG_DATA_QUERY_ENGINE` and
```
cpcloud, line 12:
```suggestion
You are probably already working with one or more `$BIG_DATA_QUERY_ENGINE`s and
```

docs/why_ibis.md
```diff
@@ -0,0 +1,65 @@
...
+whether you're an individual or an enterprise, changing deployments can be a
+massive headache.
+
+For an individual, you're already doing everything yourself. Changing up your
+workflow and moving all your data around is a pain.
+
+If you work at a large enterprise, you probably don't have the option to change
+what execution resources are available to you.
```
jcrist: I don't know if we want to presume what the work environment is like for potential users. If a user's experience doesn't match this, they may not think ibis is for them. I think we can go hard on "ibis wraps your existing tools to make them more useful" front without presuming they're doing big data or working at a large inflexible corporation.

docs/why_ibis.md
```diff
@@ -0,0 +1,65 @@
...
+
+Ibis works with what you already have, so why not check out our [getting started
+guide](./getting_started.md)?
+
+# How does Ibis compare to...
+
+Let us stop you there. Ibis is an interface that empowers you to craft complex
+and powerful queries that execute on your engine of choice.
+
+The answer to, "how does Ibis compare to `X`?" is "Ibis helps you use `X`."
```
jcrist, line 30: This is excellent phrasing :100:.

cpcloud (approved):
LGTM!

jcrist (commented):
Nice work!

Two other sections we might want to add:

- "Big Data engines like <insert some list of common ones here>". This would be another heading to get the user's attention, but reiterate the point you already made that Ibis just helps you use "X".
- "Pandas". People are going to ask about it. IMO we should touch on the composability of ibis queries, and that ibis can work with pandas (again reiterating that Ibis helps you use tools you already use).

cpcloud:
I like the idea of the pandas comparison. One thing that may be worth noting is that ibis can run SQL queries directly on DataFrames via DuckDB, but I don't know if that's too in the weeds for this doc.

docs/why_ibis.md
gforsyth: Yep, my biases are showing and I don't know that this adds much.

gforsyth:
Ok, I removed the assumptions about a user's existing environment, added a quick comparison to BIG DATA, and added a comparison to `pandas` -- I tried to strike the right balance on that last one.

cpcloud (approved):

jcrist (approved):

docs/why_ibis.md
```diff
@@ -0,0 +1,80 @@
...
+- Ibis is written in Python and features some pretty killer tab-completion
+- Ibis lets you use SQL when you want to (for our SQL-based backends)
+
+If your SQL-fu is strong, we might not convince you to leave it all behind, but
+check out our [Ibis for SQL Programmers guide](./ibis-for-sql-programmers.ipynb)
+and see if it whets your appetite.
+
+## `pandas`
+
+`pandas` is the 800 lb panda in the room. Ibis, like every dataframe API in the
```
kszucs, line 47: How about using a standard, e.g. SI here :)

PR #5959, opened 2023-04-10
fix(duckdb): branch to avoid unnecessary dataframe construction

gforsyth:


gforsyth:
urgh, so do we want to also branch on whether `to_frame` is going to hand us arrow-backed dtypes?

jcrist:
Could we instead catch the duckdb NotImplementedError on the register call, and fallback to conversion?

gforsyth:
> Could we instead catch the duckdb NotImplementedError on the register call, and fallback to conversion?

that sounds better

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -213,9 +211,3 @@ def test_s3_403_fallback(con, httpserver, monkeypatch):
     # pyarrow, which indicates the fallback worked
     with pytest.raises(pa.lib.ArrowInvalid):
         con.read_parquet(httpserver.url_for("/myfile"))
-
-
-def test_register_numpy_str(con):
```
cpcloud, line 218: Can we keep this test and xfail it?

ibis/backends/duckdb/__init__.py
```diff
@@ -745,8 +719,23 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
 
         name = op.name
-        table = op.data.to_pyarrow(schema)
-        self.read_in_memory(table, table_name=name)
+        # If op.data is backed by a Pandas dataframe, to_pyarrow
+        # can take several seconds for larger dataframes
+        if isinstance(op.data, DataFrameProxy):
+            table = op.data.to_frame()
+        else:
+            table = op.data.to_pyarrow(schema)
```
cpcloud, line 727: I think we should follow this up with a `to_native`, since we have other backends that can work with both (not for compute, avoiding the copy going to one or the other is the point):

- duckdb
- polars
- bigquery
- clickhouse (not yet implemented)

ibis/backends/duckdb/tests/test_register.py
gforsyth: yeah -- I kept getting segfaults but it's related to a separate test, I'll investigate that a bit more

cpcloud:
I added some code to convert any non-object string types to object string types. The non-pyarrow strings are very slow to register, and the pyarrow ones are segfaulting.

cpcloud (approved):

cpcloud:
Just to wrap this up with some `repr` performance measurement:

### setup

```
In [4]: from ibis.interactive import *

In [5]: import pandas as pd, numpy as np, pyarrow as pa

In [6]: n = 100_000_000

In [7]: df_object_strings = pd.DataFrame(
   ...:     {"a": np.random.randn(int(n)), "b": ["a", "b"] * int(n // 2)}
   ...: )

In [8]: df_nullable_strings = df_object_strings.astype({"b": "string"})

In [9]: df_pyarrow_strings = df_object_strings.astype({"b": "string[pyarrow]"})

In [10]: t_object_strings = ibis.memtable(df_object_strings)

In [11]: t_nullable_strings = ibis.memtable(df_nullable_strings)

In [12]: t_pyarrow_strings = ibis.memtable(df_pyarrow_strings)
```

### object string repr

```
In [13]: %timeit repr(t_object_strings)
10.3 ms ± 606 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### nullable string repr

```
In [14]: %timeit repr(t_nullable_strings)
10.1 ms ± 234 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### pyarrow string repr

```
In [15]: %timeit repr(t_pyarrow_strings)
10 ms ± 185 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

PR #5960, opened 2023-04-10
docs: move conda before nix in dev setup instructions

jcrist:
Most python devs are more familiar with conda than with nix. Since not everyone may notice the tabs in our setup instructions, we move `conda` to be the first tab in the list.


cpcloud (approved):
Seems legit!

PR #5974, opened 2023-04-12
test(ir): fix maximum console width

kszucs:
`ibis/backends/tests/test_client.py::test_interactive_repr_max_columns[duckdb-False] ` has been consistently failing for me locally on macOS with alacritty terminal. Rich used the inferred console width.

The following patch resolved the issue for me.

PR #5975, opened 2023-04-12
feat(api): add a `console_width` configuration option to explicitly set the width of the interactive repr

kszucs:
Depends on #5974 

I wasn't able to properly run the doctests because all the tables were pretty printed in the full width of my console. 
In order to have reproducible builds I added a configuration option to hardcode the console width when we execute the doctests.

### DIFF
--- a/ibis/config.py
+++ b/ibis/config.py
@@ -97,6 +97,7 @@ class Interactive(Config):
     max_string: int = 80
     max_depth: int = 1
     show_types: bool = True
+    console_width: Optional[int] = None
 
 
 class Repr(Config):
--- a/ibis/conftest.py
+++ b/ibis/conftest.py
@@ -13,7 +13,8 @@ def add_ibis(monkeypatch, doctest_namespace):
     monkeypatch.setitem(os.environ, "NO_COLOR", "1")
     # reset interactive mode to False for doctests that don't execute
     # expressions
-    ibis.options.interactive = False
+    monkeypatch.setattr(ibis.options, "interactive", False)
+    monkeypatch.setattr(ibis.options.repr.interactive, "console_width", 80)
     # workaround the fact that doctests include everything in the tested module
     # for selectors we have an `all` function
     #
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -890,7 +890,7 @@ def test_interactive_repr_max_columns(alltypes, is_jupyter, monkeypatch):
     cols = {f"c_{i}": ibis._.id + i for i in range(50)}
     expr = alltypes.mutate(**cols).select(*cols)
 
-    console = rich.console.Console(force_jupyter=is_jupyter, width=80)
+    console = rich.console.Console(force_jupyter=is_jupyter)
     options = console.options.copy()
 
     # max_columns = 0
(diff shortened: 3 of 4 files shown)


ibis/conftest.py
```diff
@@ -14,6 +14,7 @@ def add_ibis(monkeypatch, doctest_namespace):
     # reset interactive mode to False for doctests that don't execute
     # expressions
     ibis.options.interactive = False
+    ibis.options.repr.interactive.console_width = 80
```
kszucs, line 17: This seems to be setting the option for all the tests, so this os probably not the right place to set it. Although if I don't set it the doctests fail for me.

cpcloud (approved):

ibis/config.py
```diff
@@ -97,6 +97,7 @@ class Interactive(Config):
     max_string: int = 80
     max_depth: int = 1
     show_types: bool = True
+    console_width: Optional[int] = None
```
jcrist, line 100: I don't think exposing this option to the user makes much sense. I'd rather figure out what we need to do to make the doctests reproducible without changing anything user facing.

jcrist (commented):
That this is needed is a bit odd - if I make my terminal very very large the doctests still all pass for me.

What's the output of the following for you:

```
python -c "import rich;print(rich.get_console().width)"
```

ibis/config.py
cpcloud: I agree.

cpcloud:
I am able to run the doctests on linux with silly terminal widths like:

- 15
- 911

gforsyth:
@kszucs -- does this occur in terminal emulators besides `alacritty` on your machine?

kszucs:
> python -c "import rich;print(rich.get_console().width)"

248 in my usual terminal window. I just checked three different terminal emulators and the console width detection works as expected in all three.

ibis/config.py
kszucs: We can use another global variable to force a terminal width when running the doctests, though I think exposing it could be useful too, perhaps with a different name. 

This option is similar to `max_columns` but basically forces the table repr to fit in a certain number of characters.

kszucs:
Can you try it out locally? The doctests pass for me. 

I propose to keep the global configuration option but with a better name indicating `max textual width of tables measured in characters`.

cpcloud:
This is related to the order in which terminal width is captured. See [this comment](https://github.com/pytest-dev/pytest/issues/4030#issuecomment-425673898) and the discussion for more details.

ibis/config.py
cpcloud: Actually, nevermind, that would set it for all users and force it to be static.

jcrist: I pushed up an alternative fix at #5999 that I believe should resolve this issue without the need for a new option.

ibis/conftest.py
```diff
@@ -13,7 +13,8 @@ def add_ibis(monkeypatch, doctest_namespace):
     monkeypatch.setitem(os.environ, "NO_COLOR", "1")
     # reset interactive mode to False for doctests that don't execute
     # expressions
-    ibis.options.interactive = False
+    monkeypatch.setattr(ibis.options, "interactive", False)
```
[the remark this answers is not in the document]
kszucs: That wasn't entirely clear based on the comment.

kszucs:
Closing in favor of #5999

PR #5977, opened 2023-04-12
refactor(deps): remove clickhouse-cityhash and lz4 as explicitly optional dependencies

cpcloud:
This PR removes some optional dependencies that were only being imported in the clickhouse backend to set the default value of a dependency.

With its removal, conda lockfiles can be generated for clickhouse for all platforms.

I consider this a breaking change, as any new environment will now have a different default setting for compression.


cpcloud:
I'm going to target `6.x.x` on this for now.

cpcloud:
I'll retarget `master` now, since `6.x.x` was merged in.

jcrist (approved):

PR #5979, opened 2023-04-12
ci(deps): remove snowflake deps when testing pyspark with pandas2

cpcloud:
This PR removes snowflake deps in CI when testing against pandas2 with the pyspark backend.

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -439,7 +439,7 @@ jobs:
         run: docker compose logs
 
   test_pyspark:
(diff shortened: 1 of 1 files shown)


.github/workflows/ibis-backends.yml
```diff
@@ -448,16 +448,24 @@ jobs:
...
         python-version:
           - "3.10"
-        pandas-version:
-          - "1.5.*"
-          - "2.*.*"
+        pandas:
+          - version: "1.5.*"
+          - version: "2.*.*"
+            conflicts:
+              - snowflake-sqlalchemy
```
kszucs, line 455: Nice parametrization!

kszucs (approved):

PR #5987, opened 2023-04-12
docs: improve docstrings for map value and struct value

wjones127:
Changes included:

 * Fixed the docstrings for `ibis.struct` and `ibis.map` to say they create literals *or* columns, instead of just literals.
 * Added more specific type hint to `ibis.map`.
 * Moved docstrings to use memtable for struct and map, so they are similar to the array doc strings. There are some that I couldn't, such as `Map.values()` and `Map.__add__()` since those operations aren't supported by DuckDB.

### DIFF
--- a/ibis/backends/duckdb/registry.py
+++ b/ibis/backends/duckdb/registry.py
@@ -267,12 +267,20 @@ def _map_keys(t, op):
     )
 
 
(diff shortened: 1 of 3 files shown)


cpcloud (approved):
Thank you!

PR #5996, opened 2023-04-13
chore(deps): bump lower bound of ipython in dev dependencies

gforsyth:
We had a contributor run into issues with a very old version of
`prompt_toolkit` (3.0.4, ~3 years old) in the conda-lock files cause
trouble.

This bumps the lower bound of `ipython` (which is the package that pulls
in `prompt_toolkit`) so that we get a less-old version.

I don't think this counts as a breaking change in that it only impacts
dev dependencies.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,7 +96,7 @@ trino = { version = ">=0.319,<1", optional = true, extras = ["sqlalchemy"] }
 [tool.poetry.group.dev.dependencies]
 black = ">=22.1.0,<24"
 google-cloud-storage = ">=2.7.0,<3"
-ipython = ">=7.27.0,<9"
+ipython = ">=8.7.0,<9"
 poetry-dynamic-versioning = ">=0.18.0,<1"
 pre-commit = ">=3.1,<4"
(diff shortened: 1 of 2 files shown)


cpcloud:
Definitely not a breaking change. Only things under `tools.poetry.dependencies` are subject to that. Everything else is developer tooling.

pyproject.toml
```diff
@@ -96,7 +96,7 @@ trino = { version = ">=0.319,<1", optional = true, extras = ["sqlalchemy"] }
 [tool.poetry.group.dev.dependencies]
 black = ">=22.1.0,<24"
 google-cloud-storage = ">=2.7.0,<3"
-ipython = ">=7.27.0,<9"
+ipython = ">8.6.0,<9"
```
cpcloud, line 99: I guess 8.6.0 is not okay? Maybe we should have the `>=` version for consistency with the rest of the specs and for clarity on which specific lower bound works?

cpcloud (approved):
Small comment/question, not a blocker.

Thanks for doing this!

PR #6005, opened 2023-04-14
fix(impala): expose `hdfs_connect` function as `ibis.impala.hdfs_connect`

### LINKED ISSUE #5976: bug: Impala Documentation is out of date
### What happened?

I was following https://ibis-project.org/backends/Impala/ and most of the things described there are not working anymore. The docstrings seem to be out of date.
For example in the first connection example:
```
>>> hdfs = ibis.impala.hdfs_connect(host=hdfs_host, port=hdfs_port)
```
returns an error: `AttributeError: module 'ibis' has no attribute 'impala'`

5.1.0

Impala

### Relevant log output

_No response_

### Code of Conduct

jcrist:
Fixes #5976

### DIFF
--- a/ibis/__init__.py
+++ b/ibis/__init__.py
@@ -111,5 +111,8 @@ def __getattr__(name: str) -> BaseBackend:
     proxy._to_sql = backend._to_sql
     if hasattr(backend, "_sqlglot_dialect"):
         proxy._sqlglot_dialect = backend._sqlglot_dialect
+    # Add any additional methods that should be exposed at the top level
+    for name in getattr(backend, "_top_level_methods", ()):
+        setattr(proxy, name, getattr(backend, name))
 
     return proxy
--- a/ibis/backends/impala/__init__.py
+++ b/ibis/backends/impala/__init__.py
@@ -182,6 +182,7 @@ class Backend(BaseSQLBackend):
     name = 'impala'
     # not 100% accurate, but very close
     _sqlglot_dialect = "hive"
+    _top_level_methods = ("hdfs_connect",)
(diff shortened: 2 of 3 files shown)


cpcloud (approved):
This feels just a bit too special case-y for impala.

We're hard coding the attributes in the module getattr, and instead we can turn this into a base class attribute that the caller can use, that doesn't need to be aware of some other specific thing that's only defined for a single object.

That said, I or someone else can refactor this as needed later.

cpcloud:
> I agree that the impala client creation setup here doesn't fit with our standard model. I'm not sure I understand what you're saying by
> 
> > We're hard coding the attributes in the module getattr, and instead we can turn this into a base class attribute that the caller can use, that doesn't need to be aware of some other specific thing that's only defined for a single object.
> 
> What changes are you suggesting here?

Something like this is what I had in mind (but I don't think it's worth the trouble right now):

```python
class BaseBackend:
    __core_api__ = ["compile", ...]
    __extra_api__ = []

    @property
    def __api__(self):
        return self.__core_api__ + self.__extra_api__


class ImpalaBackend(BaseBackend):
    __extra_api__ = ["hdfs_connect"]


for api in backend.__api__:
    setattr(proxy, api, getattr(backend, api))
```

jcrist:
I think I'd rather remove the magic proxy objects in the long run. It's too much magic. I went with a list of extra attribute names on the backend to avoid hardcoding `hdfs_connect` itself in the module `__getattr__`, which feels like a pragmatic but still general fix to me.

cpcloud:
Yep, that's why I approved!

PR #6006, opened 2023-04-14
perf(snowflake): send everything through pyarrow

cpcloud:
## What does this PR Implement?

This PR removes the pyarrow optionality in the snowflake backend in an effort to

1. Remove the divergence between slow-path-no-pyarrow and fast-path-pyarrow. This simplifies the backend code.
1. Improve performance (numbers below) meaningfully, when pulling back medium to large results.
1. ~Make type handling more consistent (though, it's not clear if that's actually happening in this PR because of the added complexity around post-execute conversion!)~ This is much much simpler after mapping every complex type to `convert_JSON` (see the diff for context).

Some implications:

We're back in the situation of `pyarrow` being limited to version 10 in the lock file.

This doesn't limit end user versions of pyarrow except for snowflake backend
end users. Previously, you could have a snowflake-incompatible version of
pyarrow installed and the snowflake backend would work, but it'd be much slower.

We can artificially install a newer version of pyarrow in our other builds,
but that will slow down CI quite a bit. Perhaps we can add a few special
jobs of the faster-to-test backends that explicitly install `pyarrow>10`

Alternatively, we can risk users having a broken install by leaving the `"pandas"` extras
off, but that seems like a bad idea to me.

### Setup

tpch scale factor 1, lineitem table, ~6MM rows, ran each twice to ensure the warehouse was up and warm

### Baseline, **WITHOUT** PyArrow

```
In [1]: import os

In [2]: from ibis.interactive import *

In [3]: con = ibis.connect(os.environ["SNOWFLAKE_URL"])
li
In [4]: lineitem = con.table("LINEITEM", schema="snowflake_sample_data.tpch_sf1")

In [5]: %time df = lineitem.to_pandas()
CPU times: user 1min 35s, sys: 7.92 s, total: 1min 43s
Wall time: 1min 42s
```

### Candidate, **WITH** PyArrow

```
In [8]: %time df = lineitem.to_pandas()
CPU times: user 9.87 s, sys: 4.17 s, total: 14 s
Wall time: 20.5 s
```

The speedup is pretty not bad!


ibis/backends/pandas/client.py
```diff
@@ -192,37 +195,131 @@ def convert_any_to_timestamp(_, out_dtype, column):
...
+@sch.convert.register(object, dt.Boolean, bool)
+@sch.convert.register(object, dt.String, str)
+@sch.convert.register(object, dt.Integer, (int, np.integer))
+@sch.convert.register(object, dt.Floating, float)
+@sch.convert.register(object, dt.Date, datetime.date)
+@sch.convert.register(object, dt.Timestamp, datetime.datetime)
+@sch.convert.register(object, dt.Time, datetime.time)
+@sch.convert.register(object, dt.Interval, datetime.timedelta)
+@sch.convert.register(object, dt.JSON, (list, dict, int, float))
+def scalar_no_op(_, out_dtype, v):
```
jcrist, line 208: It'd be nice if all these `sch.convert.register` calls were somewhere outside the pandas backend, since they'll be needed for any `expr.to_pandas()`/`expr.execute()` call (not just in the pandas backend). This doesn't have be done here though.

ibis/backends/pandas/client.py
```diff
@@ -192,37 +195,131 @@ def convert_any_to_timestamp(_, out_dtype, column):
...
-    return column.map(convert_element)
+@sch.convert.register(object, dt.Integer, float)
+def scalar_int_convert(_, out_dtype, v):
+    if math.isnan(v):
+        return None
+    return int(v)
+
+
+@sch.convert.register(np.dtype, (dt.JSON, dt.Array, dt.Map, dt.Struct), str)
+def convert_json_to_scalar(in_, out, value: str):
```
jcrist, line 257: Do we really want to enable this conversion for all backends? IIUC this is to needed since snowflake returns all nested types as JSON strings - could we handle this in a separate external mechanism instead of in `Schema.apply_to`?

cpcloud: This is done per backend now!

cpcloud:
Snowflake backend tests look good:

```
cloud in 🌐 falcon in …/ibis on  snowflake-only-pyarrow is 📦 v5.1.0 via 🐍 v3.10.11 via ❄️  impure (ibis-3.10.11-env)
❯ pytest -m snowflake -n auto --dist loadgroup --snapshot-update
================================================ test session starts =================================================
platform linux -- Python 3.10.11, pytest-7.3.1, pluggy-1.0.0
Using --randomly-seed=3943998521
rootdir: /home/cloud/ibis
configfile: pyproject.toml
plugins: randomly-3.12.0, hypothesis-6.76.0, profiling-1.7.0, benchmark-4.0.0, clarity-1.0.1, xdist-3.3.1, cov-4.1.0, mock-3.10.0, httpserver-1.0.8, repeat-0.9.1, snapshot-0.9.0
32 workers [959 items]
...xs.x........s......................s........xx...x..............xx.........x.....x...x..xx.x..x...xx.x.x... [ 11%]
x...x..x...x..x.xx...x..x..xx.....x...x...x.........x..x.......x.x.xxxxx.....xx..............x.....x..x......x [ 22%]
.................x.x...........................................x...........x.xx...x.x...x.x.x.x............... [ 34%]
.x........x...x...x..x..x...x..........x..........x...........x.x........xx......x.x.....s...xxxxxxxx.xx...xxx [ 45%]
......xx..x......................x.....xx....x...x........x.xxxx..xxx.....x...x.....x.xx........x............. [ 57%]
..........x.......x......x..x...x............................................................x.........x...... [ 68%]
....x.......................x.........x..................x.....x.............x......x......................... [ 80%]
.........x.....x......................x..x.......x.x.......x.....x.x....x..xx.............x....x..........x... [ 91%]
..xxx....xs...s.x.x...xx.....x....xx..x...x.x..x..x..x.x..x.xx.xxx.xxxxxxxxxxx.                                [100%]
============================== 780 passed, 6 skipped, 173 xfailed in 169.61s (0:02:49) ===============================
```

ibis/formats/pandas.py
```diff
@@ -100,90 +96,131 @@ def schema_from_dask_dataframe(df, schema=None):
...
+        try:
+            return s.astype(pandas_type)
+        except Exception:  # noqa: BLE001
+            return s
 
-    if s.dtype == pandas_type and dtype.is_primitive():
-        return s
+    @staticmethod
+    def convert_Boolean(s, dtype, pandas_type):
+        import pandas.api.types as pdt
```
kszucs, line 135: It can be removed.

kszucs (approved):

PR #6020, opened 2023-04-15
fix(sql): ensure that overwrite is valid when clobbering an existing table

### LINKED ISSUE #6014: bug: create_table() from same table results in empty table
### What happened?

Not sure exactly what criteria need to happen in order to hit this, but this does it:

```python
import ibis

ex = ibis.examples.aids.fetch()
conn = ex._find_backend()
t = conn.create_table("x", ex)
t2 = t.filter(t.delay > 10)
assert t2.count().execute() > 0
conn.create_table("y", t2)  # This works fine, because we are storing in a new table
assert conn.table("y").count().execute() > 0  # The new table looks as expected
conn.create_table("x", t2, overwrite=True)  # this doesn't throw an error
assert conn.table("x").count().execute() == 0  # but the table is empty
assert t2.count().execute() == 0  # AND the original table is empty because it referenced "x"
```

5.1.0

duckdb

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
This PR fixes an issue where we are currently deleting an existing table when
asked to overwrite that table, before executing an expression that may depend
on that table.

The solution is a bit complex, but is robust and takes advantage of `CREATE OR REPLACE` syntax
in backends that support it.

The way the backends that don't support `CREATE OR REPLACE` work is that they:

1. create a temporary table with the input expression's data
1. recreate the target table
1. insert the temp table's data into the target table
1. drop the temp table

Fixes #6014.

### DIFF
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -1,3 +1,6 @@
+from __future__ import annotations
+
+import contextlib
 import os
 import platform
 import re
@@ -18,7 +21,8 @@ import ibis
 import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
-from ibis.util import guid
+from ibis.backends.base import BaseBackend
+from ibis.util import gen_name, guid
 
 
 @pytest.fixture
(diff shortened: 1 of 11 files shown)


ibis/backends/tests/test_client.py
```diff
@@ -1201,3 +1205,41 @@ def test_persist_expression_release(con, alltypes):
...
+
+@mark.notimpl(
+    ["bigquery", "datafusion", "polars"],
+    raises=NotImplementedError,
+    reason="overwriting not implemented in ibis for this backend",
+)
+@mark.broken(
+    ["druid"], raises=sa.exc.ProgrammingError, reason="generated SQL fails to parse"
+)
+@mark.notimpl(["clickhouse"], raises=TypeError, reason="requires the `engine` argument")
```
cpcloud, line 1227: This test should pass now.

kszucs (approved):

PR #6028, opened 2023-04-17
fix(snowflake): always use pyarrow for memtables

cpcloud:
Fixes the failing Snowflake backend tests against `master`.

Originally, I implemented just the quoting fixes, but this turned out to expose
some annoyances with quoting in `write_pandas` that were a pain to workaround.

`write_pandas` is overkill for our needs, so I went ahead implemented similar
functionality directly in the Snowflake backend's `_register_in_memory_table`
method. The result is relatively easy to understand compared to what is going
on in `write_pandas`.

It also turned out that we can use this code path regardless of whether a
version of PyArrow is installed that is compatible with bringing back results
from Snowflake in pandas/arrow format. This is because we're only using pyarrow
to write parquet

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Any, Iterable, Mapping
 
 import sqlalchemy as sa
 from sqlalchemy.ext.compiler import compiles
+from sqlalchemy.sql import quoted_name
 from sqlalchemy.sql.expression import ClauseElement, Executable
 
 import ibis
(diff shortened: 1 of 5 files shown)


ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -91,7 +94,8 @@ def _create_table_as(element, compiler, **kw):
     if element.temp:
         stmt += "TEMPORARY "
 
-    return stmt + f"TABLE {element.name} AS {compiler.process(element.query, **kw)}"
+    name = compiler.preparer.quote(quoted_name(element.name, quote=element.quote))
```
gforsyth, line 97: Is it necessary to call `quote` on `quoted_name`?
If it is, maybe a brief comment on what's going on.

gforsyth (commented):
Looks good to me -- just the one question.

gforsyth (approved):
:shipit:

PR #6034, opened 2023-04-19
feat(sqlite): add argmin and argmax functions

mesejo:
### Implementation

Since version [3.7.11]( http://www.sqlite.org/releaselog/3_7_11.html), SQLite supports queries of the form:

```sqlite
SELECT max(x), y FROM table
```
that:

> returns the value of y on the same row that contains the maximum x value.

The same applies to the min aggregate function. 

This solution combines the max (min) aggregate function with the JSON functions: [json_array](https://www.sqlite.org/json1.html#jarray) and [json_extract](https://www.sqlite.org/json1.html#jex) to first build an array: `[y, max(x)]` from the aggregate and then extract `y`.

### DIFF
--- a/ibis/backends/sqlite/registry.py
+++ b/ibis/backends/sqlite/registry.py
@@ -27,7 +27,6 @@ from ibis.common.enums import DateUnit, IntervalUnit
 operation_registry = sqlalchemy_operation_registry.copy()
 operation_registry.update(sqlalchemy_window_functions_registry)
 
-
 sqlite_cast = Dispatcher("sqlite_cast")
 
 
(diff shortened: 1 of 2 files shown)


cpcloud:
I can help with the Windows failure.

mesejo:
> I can help with the Windows failure.

Thanks 🙏

cpcloud:
Ended up skipping on Windows CI. It seems like the documentation for JSON functions isn't entirely correct, or there's some mismatch/my own misunderstanding of what the Python reported version of SQLite means.

For example, Linux reports the version of SQLite to be 3.37, but the SQLite JSON docs say that JSON functions weren't available until 3.38.

Windows claims that the version is at least 3.38, and yet the JSON functions aren't available.

cpcloud (approved):
Thanks, LGTM!

PR #6036, opened 2023-04-19
chore(deps): pyspark and dask upgrades

cpcloud:
This PR upgrades dependencies to allow PySpark to work with recent numpy and pandas versions and fixes some failing dask tests.


cpcloud:
This seems pretty fraught, I'm going to leave it alone for now.

Issue #6038, opened 2023-04-19
Bug: Unable to create connection with DuckDB data source in Windows environment.


keithbritt:
### What happened?

I've tried this also with a separate data source that was a csv imported and saved as a ddb file, but I double checked with the getting started dataset and get the same exact result.

Here's the code I'm running:

import ibis
import urllib.request
urllib.request.urlretrieve(
        "https://storage.googleapis.com/ibis-tutorial-data/palmer_penguins.ddb",
        "palmer_penguins.ddb",
    )
ibis.options.interactive = True
ibis.options.verbose     = True

con = ibis.duckdb.connect("palmer_penguins.ddb")
con.list_tables()

birds = con.table("penguins")

And here's the command line I run along with the error message I received (it doesn't just warn, it fails and terminates). Notice it's a Postgres error even though I'm using DuckDB.

py Hello_World.py

...Python\Python310\site-packages\ibis\backends\postgres\registry.py:119: UserWarning: locale specific date formats (%c, %x, %X) are not yet implemented for Windows
  warnings.warn(

### What version of ibis are you using?

5.0

### What backend(s) are you using, if any?

DuckDB

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
@keithbritt Thanks for the issue!

Can you show the full error message? You're showing only the warning, but then you say that it's not only warning but a failure and termination.

Can you show exactly what happens?

keithbritt:
That's all the information I'm getting. Here's a screen shot if that's helpful.

<img width="1082" alt="image" src="https://user-images.githubusercontent.com/7809612/233086794-b7207473-8fa0-47a9-b403-5f09482f1aa8.png">

cpcloud:
I don't think that's a failure. Your program appears to be terminating normally.

cpcloud:
This code: ```
import ibis
import urllib.request
urllib.request.urlretrieve(
"[storage.googleapis.com/ibis-tutorial-data/palmer_penguins.ddb](https://storage.googleapis.com/ibis-tutorial-data/palmer_penguins.ddb)",
"palmer_penguins.ddb",
)
ibis.options.interactive = True
ibis.options.verbose = True

con = ibis.duckdb.connect("palmer_penguins.ddb")
con.list_tables()

birds = con.table("penguins")
```

will not show any output.

You need to take an action that prints to stdout.

For example, you could add

```
print(birds)
```

as the final line of code.

[...]

PR #6039, opened 2023-04-19
chore(windows): move warning to error on locale-specific format string usage

cpcloud:
This PR moves a not-so-useful warning to a more informative error message. Previously this would be a `KeyError`.

xref: #6038

### DIFF
--- a/ibis/backends/postgres/registry.py
+++ b/ibis/backends/postgres/registry.py
@@ -118,10 +118,9 @@ try:
         }
     )
 except AttributeError:
-    warnings.warn(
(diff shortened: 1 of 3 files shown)


ibis/backends/postgres/registry.py
```diff
@@ -170,6 +170,10 @@ def _reduce_tokens(tokens, arg):
 
     # TODO: how much of a hack is this?
     for token in tokens:
+        if token in _locale_specific_formats and not HAS_LANGINFO:
+            raise com.UnsupportedOperationError(
```
cpcloud, line 174: This isn't covered because we don't run postgres on windows.

PR #6041, opened 2023-04-19
feat(common): add pattern matchers

kszucs:
Converted the previous validator system into a pattern matching system.

The previously used validator system had the following problems:
- Used curried functions which was error prone to missing arguments and was hard to debug.
- Used exceptions for control flow which raising exceptions from the innermost function call giving cryptic error messages.
- While it was similarly composable it was hard to see whether the validator was fully constructed or not.
- We wasn't able traverse the nested validators due to the curried functions.

In addition to those the new approach has the following advantages:
- The pattern matching system is fully composable.
- Not we support syntax sugar for combining patterns using & and |.
- We can capture values at mutiple levels using either `>>` or `@` syntax.
- Support structural pattern matching for sequences, mappings and objects.

### DIFF
--- a/ibis/common/annotations.py
+++ b/ibis/common/annotations.py
@@ -5,7 +5,7 @@ import inspect
 from typing import Any
 
 from ibis.common.collections import DotDict
-from ibis.common.typing import evaluate_typehint
+from ibis.common.typing import get_type_hints
 from ibis.common.validators import Validator, any_, frozendict_of, option, tuple_of
 
 EMPTY = inspect.Parameter.empty  # marker for missing argument
@@ -247,6 +247,7 @@ class Signature(inspect.Signature):
         Signature
         """
         sig = super().from_callable(fn)
+        typehints = get_type_hints(fn)
 
         if validators is None:
             validators = {}
@@ -263,12 +264,11 @@ class Signature(inspect.Signature):
             name = param.name
             kind = param.kind
             default = param.default
-            typehint = param.annotation
+            typehint = typehints.get(name)
 
             if name in validators:
                 validator = validators[name]
-            elif typehint is not EMPTY:
-                typehint = evaluate_typehint(typehint, fn.__module__)
+            elif typehint is not None:
                 validator = Validator.from_typehint(typehint)
             else:
                 validator = None
@@ -288,8 +288,7 @@ class Signature(inspect.Signature):
 
         if return_validator is not None:
             return_annotation = return_validator
-        elif sig.return_annotation is not EMPTY:
-            typehint = evaluate_typehint(sig.return_annotation, fn.__module__)
+        elif (typehint := typehints.get("return")) is not None:
             return_annotation = Validator.from_typehint(typehint)
         else:
             return_annotation = EMPTY
--- a/ibis/common/caching.py
+++ b/ibis/common/caching.py
@@ -1,13 +1,32 @@
 from __future__ import annotations
 
+import functools
 import weakref
 from collections import Counter, defaultdict
+from collections.abc import Iterator
 from typing import Any, Callable, MutableMapping
 
-import toolz
 from bidict import bidict
 
+from ibis.common.collections import FrozenDict
 from ibis.common.exceptions import IbisError
+
+
+def memoize(func: Callable) -> Callable:
+    """Memoize a function."""
+    cache = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        key = (args, FrozenDict(kwargs))
+        try:
+            return cache[key]
+        except KeyError:
+            result = func(*args, **kwargs)
+            cache[key] = result
+            return result
+
+    return wrapper
 
 
 class WeakCache(MutableMapping):
@@ -19,13 +38,13 @@ class WeakCache(MutableMapping):
     def __setattr__(self, name, value):
         raise TypeError(f"can't set {name}")
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self._data)
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator[Any]:
         return iter(self._data)
 
-    def __setitem__(self, key, value):
+    def __setitem__(self, key, value) -> None:
         # construct an alternative representation of the key using the id()
         # of the key's components, this prevents infinite recursions
         identifiers = tuple(id(item) for item in key)
@@ -78,7 +97,7 @@ class RefCountedCache:
         self.lookup = lookup
         self.finalize = finalize
         self.names = defaultdict(generate_name)
-        self.key = key or toolz.identity
+        self.key = key or (lambda x: x)
 
     def get(self, key, default=None):
         try:
--- a/ibis/common/patterns.py
+++ b/ibis/common/patterns.py
@@ -1,0 +1,1196 @@
+from __future__ import annotations
+
+import inspect
+import math
+import numbers
+from abc import ABC, abstractmethod
+from collections.abc import Callable, Hashable, Mapping, Sequence
+from enum import Enum
+from inspect import Parameter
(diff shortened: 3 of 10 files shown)


ibis/common/annotations.py
```diff
@@ -288,9 +288,8 @@ def from_callable(cls, fn, validators=None, return_validator=None):
 
         if return_validator is not None:
             return_annotation = return_validator
-        elif sig.return_annotation is not EMPTY:
-            typehint = evaluate_typehint(sig.return_annotation, fn.__module__)
-            return_annotation = Validator.from_typehint(typehint)
+        elif "return" in typehints:
+            return_annotation = Validator.from_typehint(typehints["return"])
```
cpcloud, line 292:
```suggestion
        elif (typehint := typehints.get("return")) is not None:
            return_annotation = Validator.from_typehint(typehint)
```

Maybe this doesn't matter in the grand scheme of things, but you can avoid the extra lookup here for about the same amount of code 🤷🏻

ibis/common/caching.py
```diff
@@ -1,15 +1,34 @@
...
 from typing import Any, Callable, MutableMapping
 
-import toolz
 from bidict import bidict
 
+from ibis.common.collections import FrozenDict
 from ibis.common.exceptions import IbisError
 
 
+def memoize(func: Callable) -> Callable:
```
cpcloud, line 15: Isn't this exactly what `toolz.memoize` does?

ibis/common/typing.py
```diff
@@ -2,58 +2,182 @@
...
+    return result
+
+
+@memoize
+def get_bound_typevars(obj: Any) -> dict[str, Any]:
+    """Get type variables bound to concrete types for a generic class.
+
+    Parameters
+    ----------
+    obj : Any
```
cpcloud, line 114:
```suggestion
    obj
```

ibis/common/typing.py
```diff
@@ -2,58 +2,182 @@
...
+    """Get type hints for a callable or class.
+
+    Extension of typing.get_type_hints that supports getting type hints for
+    class properties.
+
+    Parameters
+    ----------
+    obj : Any
+        Callable or class to get type hints for.
+    include_extras : bool, default True
```
cpcloud, line 39:
```suggestion
    include_extras
```

Likewise the default is visible in the function signature, no need to indicate it here.

ibis/common/typing.py
```diff
@@ -2,58 +2,182 @@
...
+        Whether to include extra type hints such as Annotated.
+    include_properties : bool, default False
```
cpcloud, line 41:
```suggestion
    include_properties
```

ibis/common/typing.py
```diff
@@ -2,58 +2,182 @@
...
+    return hints
+
+
+@memoize
+def get_type_params(obj: Any) -> dict[str, Any]:
+    """Get type parameters for a generic class.
+
+    Parameters
+    ----------
+    obj : Any
```
cpcloud, line 71:
```suggestion
    obj
```

ibis/common/typing.py
```diff
@@ -2,58 +2,182 @@
...
+        Type annotations to evaluate.
+    module_name : str
+        The name of the module that the annotations are defined in, hence
+        providing global scope.
+    localns : Optional[Namespace]
+        The local namespace to use for evaluation.
+
+    Returns
+    -------
+    typehints
```
cpcloud, line 169:
```suggestion
    dict[str, Any]
```

ibis/common/patterns.py
```diff
@@ -0,0 +1,1199 @@
...
+        # TODO(kszucs): cache the result of this function
+        origin, args = get_origin(annot), get_args(annot)
+
+        if origin is None:
+            if annot is Ellipsis:
+                return Any()
+            elif annot is None:
+                return Is(None)
+            elif annot is AnyType:
+                return Any()
```
cpcloud, line 108:
```suggestion
            if annot is Ellipsis or annot is AnyType:
                return Any()
            elif annot is None:
                return Is(None)
```

ibis/common/patterns.py
```diff
@@ -0,0 +1,1199 @@
...
+
+    def __eq__(self, other) -> bool:
+        if self is other:
+            return True
+        if type(self) is not type(other):
+            return NotImplemented
+        for name in self.__slots__:
+            if getattr(self, name) != getattr(other, name):
+                return False
+        return True
```
cpcloud, line 250:
```suggestion
        return all(getattr(self, name) == getattr(other, name) for name in self.__slots__)
```

ibis/common/patterns.py
```diff
@@ -0,0 +1,1199 @@
...
+def match(pat: Pattern, value: AnyType, context: dict[str, AnyType] = None):
+    """Match a value against a pattern.
+
+    Parameters
+    ----------
+    pat :
+        The pattern to match against.
+    value :
+        The value to match.
+    context :
```
cpcloud, line 1179:
```suggestion
    pat
        The pattern to match against.
    value
        The value to match.
    context
```

cpcloud (approved):
Wow, what an improvement!

This code looks great.

ibis/common/caching.py
kszucs: A custom implementation of `@functools.cache` using our `frozendict`

ibis/common/typing.py
kszucs: Updating.

ibis/common/patterns.py
kszucs: Updated.

ibis/common/patterns.py
kszucs: Updated.

ibis/common/patterns.py
kszucs: Updated.

ibis/common/patterns.py
```diff
@@ -0,0 +1,1199 @@
...
+
+
+class Between(Matcher):
+    """Match a value between two bounds.
+
+    Parameters
+    ----------
+    lower :
+        The lower bound.
+    upper :
```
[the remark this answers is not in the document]
kszucs: Updated.

ibis/common/caching.py
cpcloud: Ah, right! Didn't catch that at first, but I see it now!

ibis/common/patterns.py
```diff
@@ -0,0 +1,1196 @@
...
+    def validate(self, value, context):
+        ...
+
+
+class MatchError(Exception):
+    ...
+
+
+class NoMatch:
+    """Sentinel value for when a pattern doesn't match."""
```
jcrist, line 74: Minor minor nit, but could you use an enum here instead? This pattern plays better with type checkers, and makes it clearer that `NoMatch` is a singleton *value* rather than a singleton type (i.e. consumers should check for `is NoMatch` rather than `isinstance(x, NoMatch)`). Not a big deal though.

```python
class NoMatchType(enum.Enum):
    NoMatch = "NoMatch"

NoMatch = NoMatchType.NoMatch
```

This way if we ever wanted to type annotate the `match` methods below, their signature could be:

```python
def match(self, value: Value) -> Value | NoMatchType:
    ...
```

See https://peps.python.org/pep-0484/#support-for-singleton-types-in-unions for reference.

jcrist (approved):
One nit (feel free to ignore), but overall this LGTM.

There are a few uncovered lines in `patterns.py`, mostly around `NoMatch` handling. Should we add tests to cover those?

ibis/common/patterns.py
kszucs: Sounds good to me, but I'm deferring it to a follow-up if you don't mind.

PR #6042, opened 2023-04-19
feat(duckdb): set `header=True` by default

cpcloud:
This PR changes the default value of `header` to be `True` for the DuckDB
backend's `read_csv` API.


gforsyth (approved):
looks good to me!

PR #6044, opened 2023-04-19
fix(bigquery): add srid=4326 to the geography dtype mapping

### LINKED ISSUE #6043: bug: bigquery geography dtype misses the srid (spatial reference identifier) configuration
### What happened?

BigQuery Geography column maps to `ibis_dtypes.GeoSpatial(geotype="geography")` with a None `srid` configuration. However, BigQuery uses points on the WGS84 reference ellipsoid according to [doc](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types?hl=en#geography_type ).

0.14.0

BigQuery

### Relevant log output

_No response_

### Code of Conduct

chelsea-lin:
This change can fix #6043.

Fixes #6043.

### DIFF
--- a/ibis/backends/bigquery/client.py
+++ b/ibis/backends/bigquery/client.py
@@ -29,6 +29,7 @@ _DTYPE_TO_IBIS_TYPE = {
     "TIMESTAMP": dt.timestamp,
     "BYTES": dt.binary,
     "NUMERIC": dt.Decimal(38, 9),
+    "GEOGRAPHY": dt.GeoSpatial(geotype="geography", srid=4326),
 }
 
 
--- a/ibis/backends/bigquery/datatypes.py
+++ b/ibis/backends/bigquery/datatypes.py
@@ -82,6 +82,16 @@ def trans_json(t):
     return "JSON"
 
 
+@ibis_type_to_bigquery_type.register(dt.GeoSpatial)
+def trans_geography(t):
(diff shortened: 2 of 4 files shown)


jcrist (commented):
Thanks @chelsea-lin - the fix makes sense to me!

Would you mind adding a test for this? It should be just another case added here: https://github.com/ibis-project/ibis/blob/e4b515daa6f96eaa6826c8623e5a885d05a4b3c4/ibis/backends/bigquery/tests/unit/test_datatypes.py#L53

chelsea-lin:
@jcrist thanks for the reviewing.
Added GeoGraphy test to `test_datatypes.py`, which is testing `ibis_dtype`->`bigquery type`. Hence, I have to register `dt.GeoSpatial` to `GEOGRAPHY` in datatypes.py. 
My initial change is mapping `bigquery type`->`ibis_dtype`. I am thinking which tests can cover this line change?

chelsea-lin:
@jcrist, please take another look. The tests should cover both directions of ibis and bigquery dtypes mappings. The test coverage missing at `datatypes.py` doesn't seem to be right for me, because the `test_datatypes.py` should cover that already. Let me know if anything I can follow up. Thanks!

cpcloud (approved):
LGTM, thanks!

The coverage isn't quite accurate because we don't run BigQuery tests against PRs for security reasons.

Issue #6007, opened 2023-04-14
docs: align 'Ibis for pandas Users' with recommended way of working


ianmcook:
### Describe the solution you'd like

The "Ibis for pandas Users" guide directs readers to use Ibis's pandas backend. Is this aligned with the way that the Ibis developers recommend using Ibis? Should it instead (or additionally) emphasize to readers that they can take a pandas DataFrame, use it to create an Ibis table backed by the default DuckDB backend, and then run Ibis on that?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ianmcook:
There is some risk in showing users how to create a DuckDB-backed Ibis table directly from a pandas DataFrame, because this might encourage less efficient ways of working, like: Parquet file -> pandas DataFrame -> Ibis table backed by DuckDB, when it would be more efficient to go directly Parquet file -> Ibis table backed by DuckDB. So the guide should explicitly discourage that. But there are plenty of other cases in which the user already has pandas DataFrame in memory and wants to use Ibis to work with it using the most efficient local backend.

gforsyth:
> Is this aligned with the way that the Ibis developers recommend using Ibis?

No, it is not.

I think we should demonstrate that if you already have something in pandas, you can start from there and continue with Ibis.  But that should be balanced against using the DuckDB backend directly.

ianmcook:
Digging into this deeper: there are a couple of options available for a user who has a pandas DataFrame `df` in memory and wants to use Ibis to query it with the embedded DuckDB backend:

Option 1:
```py
t = ibis.memtable(df)
```

Option 2:
```py
conn = ibis.duckdb.connect(':memory:')
conn.create_table('name', df)
t = conn.table('name')
```

In Option 1, I believe no data is copied when `t` is created, and DuckDB scans the pandas DataFrame when the query is executed. In Option 2, I believe the data is copied from the pandas DataFrame into a transient DuckDB in-memory table when `t` is created. Please correct me if any of that is wrong.

ianmcook:
Also: I believe this is shorthand for Option 2:
```py
t = ibis.memtable(df).cache()
```

jcrist:
Thanks for opening this Ian. In general I think this doc should cover two things:

- How to load data into ibis for traditional pandas tasks. This would cover loading from local files (csv & parquet), and from existing pandas dataframes
[...]

Issue #6045, opened 2023-04-19
feat: Better error message or support for joins involving multiple backends


amoeba:
### Is your feature request related to a problem?

I'm new to Ibis so I, a person new to Ibis, wanted to play with joins. I decided to bring pandas dataframes into Ibis to try that out. I first made two data frames that would be good candidates for a join:

```python
left_df = pd.DataFrame(
    [
        ['a', 2],
        ['b', 2]
    ], 
    columns=['key', 'two',],
)
right_df = pd.DataFrame(
    [
        ['a', 3],
        ['b', 3]
    ], 
    columns=['key', 'three',],
)
```
 
I then found the code to bring one into Ibis at the top of [Ibis for pandas users](https://ibis-project.org/ibis-for-pandas-users/) and wrote:

```python
t_left = ibis.pandas.connect({'left': left_df}).table('left')
```

Then to bring in my second dataframe I, a new person to Ibis, duplicated the above and added to the above:

```python
t_right = ibis.pandas.connect({'right': right_df}).table('right')
```

I then tried a join and got error:

```python
t_left.left_join(t_right, t_left.key == t_right.key)
```

> IbisError: Multiple backends found for this expression

### Describe the solution you'd like

My first thought when I saw the above error was that I needed to set or explicitly offer a backend because Ibis wasn't able to decide on which one to use to execute my expression. But after scrolling down in [Ibis for pandas users](https://ibis-project.org/ibis-for-pandas-users/), I see I need to load both dataframes into _the same backend_.

i.e., 

```python
pd_ibis = ibis.pandas.connect({'t_left': df_left, 't_right': df_right})
```

I guess I have two questions:

- What does that error message mean? Could it be improved? It sounds like I needed to give Ibis more info but I think it might have meant "you can't do this".
- Are cross-backend operations a thing? Could they be?

### What version of ibis are you running?

5.1.0

### What backend(s) are you using, if any?

Pandas, I think?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gforsyth:
> * What does that error message mean? Could it be improved? It sounds like I needed to give Ibis more info but I think it might have meant "you can't do this".

It means "you can't do this" -- and it can certainly be improved.  One bump in the road here is that the notion of connecting to a backend makes a lot more sense when you're dealing with something that isn't an in-memory dataframe.

[...]

PR #6048, opened 2023-04-20
docs(introguides): use DuckDB for intro pandas notebook, remove iris

gforsyth:
This swaps in the DuckDB backend for the Pandas backend in the "Ibis for pandas Users" notebook.  I also swapped out the Iris data for the palmer penguins data.

Very brief mention of `memtable` vs. making a backend connection.

Not a complete solution for #6045 but should hopefully help new users coming from Pandas not run into multiple backend error messages.


docs/ibis-for-pandas-users.ipynb
```diff
@@ -46,7 +46,7 @@
    "id": "afd136cd-18b5-48a2-8359-c47cde181c7b",
```
jcrist, line 46: <p>I don't have a suggestion for better wording here, but this isn't <em>strictly</em> correct. <code>memtable</code> is a tool for wrapping a dataframe and making it part of the expression tree so it can be used with any backend.   In this context what you're saying is true (and maybe for this tutorial the user doesn't need the full explanation), but I worry that a sentence like this will make users think <code>memtable</code> is only for working with the duckdb backend.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/6048/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

jcrist:
Overall this looks good to me, and is a nice improvement over the current pandas-backend centric doc. This also would fix #6007.

docs/ibis-for-pandas-users.ipynb
cpcloud: <p>How about:</p><p> </p><blockquote>You can do this more concisely with <code>memtable</code>,which also works with other backends.  </blockquote>

cpcloud, line 46: <p>Can you remove this bit? <code>offset</code> requires a caveat because of ordering. This is valid in pandas, but not in DuckDB or any other SQL backend. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/6048/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='59'/>

cpcloud, line 46: <pre class="ql-syntax" spellcheck="false">Line #1.    unioned = ibis.union(t_1, t_2)
</pre><p>This should be <code>union_all</code> to match <code>pd.concat</code>   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/6048/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='123'/>

jcrist: Yeah, I think that makes sense. IMO the point of this doc should be:

- Show how to load pandas dataframes into ibis
- Show how to load local files (csv, parquet) directly into ibis, mimicking normal pandas workloads
- Compare/contrast how ibis spells common query operations with their pandas equivalents

gforsyth: <p>I don't think we have a <code>union_all</code></p>

jcrist: `ibis.union` is a `UNION ALL`. `ibis.union(*args, distinct=True)` is `UNION`.

cpcloud: Ah, right! Apologies, this is good to go then!

Issue #5922, opened 2023-04-05
docs: ibis for R users


jcrist:
We currently have:

- [Ibis for Pandas Users](https://ibis-project.org/ibis-for-pandas-users/)
- [Ibis for SQL Programmers](https://ibis-project.org/ibis-for-sql-programmers/)

It'd be nice to have an onboarding notebook like these for R users as well (probably coming from a tidyverse/dplyr perspective).

ianmcook:
Limiting this only to "Ibis for dplyr Users" (at least initially) would make it a lot more straightforward. Later we could come back and address the other popular data manipulation APIs in R (base R, data.table).

jcrist:
This was resolved in #6050. Closing.

PR #6050, opened 2023-04-20
docs: Create Ibis for dplyr users document

amoeba:
Addresses https://github.com/ibis-project/ibis/issues/5922.

@fmichonneau and I paired up on https://github.com/ibis-project/ibis/issues/5922 during the recent Ibis doc-a-thon. To keep the work in scope for the doc-a-thon, we decided to start small and base the content on https://dplyr.tidyverse.org/articles/dplyr.html. @fmichonneau brought that content in and developed Ibis equivalents for each bit. I then extended that by adding a section at the top highlighting major differences and also added in joins and pivoting to make it more comprehensive.

@fmichonneau could you take a look, especially at the content I added?

After that, I think it'd be good to get @ianmcook to review and also a core Ibis dev.

Specific bits of feedback I'm interested in are:

- Is the size/scope of this enough? The other Ibis for X guides are more comprehensive.
- Does the section at the top about major differences work? Is it accurate and complete?

### DIFF
--- a/docs/ibis-for-dplyr-users.ipynb
+++ b/docs/ibis-for-dplyr-users.ipynb
@@ -1,0 +1,288 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Ibis for dplyr Users"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[R](https://www.r-project.org/) users familiar with [dplyr](https://dplyr.tidyverse.org/) are likely to find some parts of Ibis familiar.\n",
+    "In fact, some Ibis verbs have been named to match their corresponding dplyr verbs.\n",
+    "\n",
+    "However, due to constraints of the Python programming language and the design and goals of Ibis itself, analysts familiar with dplyr may notice some big between the two right away:\n",
+    "\n",
+    "TODO: Fill this list in more.\n",
+    "\n",
+    "- No pipe syntax but you do have chaining\n",
+    "- mutate can't do internal references, you have to chain mutates\n",
+    "- ibis has more similar to dplyr+dbplyr (TODO: expand on this)\n",
+    "- `_` as a helper\n",
+    "- NULL ordering. ibis sorts NULLs differently than dplyr/R. TODO:\n",
+    "    - Figure out why this is and explain it well\n",
+    "- There are multiple ways to reference columns (by string or by \"reference\")\n",
+    "- Wrapping complex expressions in parens to make the evaluate correctly"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Comparison\n",
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Loading Ibis"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import ibis\n",
+    "import ibis.examples as ex\n",
+    "import ibis.selectors as s"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from ibis import _"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ibis.options.interactive = True"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Loading example data"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars = ex.starwars.fetch()"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### head()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.head()"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### filter()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.filter([_.skin_color == \"light\", _.eye_color == \"brown\"])"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### arrange()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.order_by(_.height)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.order_by(_.height.desc())"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### slice()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.limit(5, offset = 4)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.select([\"hair_color\", \"skin_color\", \"eye_color\"])"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### select()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.select(s.endswith(\"color\"))"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### rename()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.relabel({\"homeworld\": \"home_world\"})"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### mutate()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.mutate(height_m = _.height / 100).select(\"height_m\", \"height\", ~s.contains(\"height\"))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "(starwars\n",
+    "    .mutate(\n",
+    "        height_m = _.height / 100\n",
+    "    )\n",
+    "    .mutate(        \n",
+    "        BMI = _.mass / (_.height_m**2)\n",
+    "    )\n",
+    "    .select(\"BMI\", ~s.matches(\"BMI\"))\n",
+    ")"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### summarize() / summarise()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.aggregate(height = _.height.mean())"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "ibis-dev",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.11.3"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 2
+}
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -22,6 +22,7 @@
         * [Configuration](api/config.md)
     * [Ibis for SQL Programmers](ibis-for-sql-programmers.ipynb)
     * [Ibis for pandas Users](ibis-for-pandas-users.ipynb)
+    * [Ibis for dplyr Users](ibis-for-dplyr-users.ipynb)
     * [Backend Operations Matrix](backends/support_matrix.md)
     * [Why Ibis?](why_ibis.md)
 * [Releases](release_notes.md)


docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,777 @@
+{
```
cpcloud, line 1: <p>Most databases allow you to choose null ordering either as a global configuration option, in the expression itself (e.g., <code>NULLS LAST</code>) or both. So far nobody has asked for this functionality so we haven't implemented it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/6050/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

cpcloud, line 1: <p>You can elide the <code>[]</code> here too, similar to dplyr. That's the preferred syntax,  <code>[]</code> was kept for backwards compatibility. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/6050/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='39'/>

cpcloud, line 1: <p>Interesting, didn't realize this.</p><p> </p><p>We could potentially support this but it might get pretty hairy. What does dplyr do when there's ambiguity, e.g., <code>mutate(x=x + 1, y=x + 1)</code>? Which <code>x</code> gets used there?   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/6050/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='49'/>

cpcloud, line 1: <p>Should one of the <code>summarize</code>s have an <code>s</code>?    </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/6050/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='51'/>

cpcloud, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.        band_members.name == band_instruments.name
</pre><p>This could be just <code>"name"</code> if you want. Not sure if you're going for explicitness here. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/6050/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='56'/>

cpcloud, line 1: <p>What is the guessing algorithm in dplyr?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ibis-project/ibis/pull/6050/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='57'/>

amoeba: <p>Oh neat. This honestly isn't something I've ever had to deal with in database-land. <code>NULLS LAST</code> is good to know about. </p>

amoeba: <p>I didn't actually know. But:</p><p> </p><pre class="ql-syntax" spellcheck="false">&gt; data.frame(x=1:3) |&gt;&nbsp;
&nbsp; mutate(x = x + 1, y = x + 1)
&nbsp; x y
1 2 3
2 3 4
3 4 5
</pre>

amoeba: <p>Yes! Good eye, thanks.</p>

amoeba: <p>Thanks. That's a good improvement.</p>

amoeba: <p>Nice. I don't think we were going for explicitness here.</p>

amoeba: <p>I just looked and I think I'll at least update the language here. When the join predicate is missing, dplyr does a natural join so it's not "guessing" which is what I thought.</p>

amoeba:
Thanks for having a go-through @cpcloud. I resolved all but two of your comments.

docs/ibis-for-dplyr-users.ipynb
cpcloud: Cool, I wonder if the order of the usage matters but we can leave that for later!

cpcloud: We could potentially implement the same natural join semantics which I think is "join on columns with the same name", but not sure if it's worth the trouble right now.

gforsyth (changes requested):
Thanks for putting this together @amoeba and @fmichonneau!

I flagged a few typos and some small language nits, but nothing major.

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,774 @@
...
+    "In fact, some Ibis features were even inspired by similar features in the [Tidyverse](https://www.tidyverse.org/).\n",
+    "\n",
+    "However, due to differences between Python and R and the design and goals of Ibis itself, you may notice some big differences right away:\n",
+    "\n",
+    "- **No pipe:** The handy [magrittr pipe](https://magrittr.tidyverse.org/) (`%>%`) or R's newer native pipe (`|>`) don't exist in Python so you instead have to chain sequences of operations together with a period (`.`).\n",
+    "- **No unquoted column names:** Non-standard evaluation is common in R but not present in Python. To reference a column in Ibis, you can pass a string, property on a table (e.g., `tbl.some_column`), or you can make use of [selectors](https://ibis-project.org/api/selectors/).\n",
+    "- **Ibis is lazy by default:** Similar to [dbplyr](https://dbplyr.tidyverse.org/) and its `collect()` method, Ibis does not evaluate our queries until we call `.execute()`. For the purposes of this document, we set `ibis.options.interactive = True` which calls `execute` automatically for us and prints a nicely-formatted table."
```
ianmcook, line 22: When `ibis.options.interactive = True`, Ibis also adds a `LIMIT 11` to the query. That way the backend engine is only processing as much data is necessary to display 10 rows of the result in the nicely-formatted table (plus an ellipsis if there are more than 10 rows in the result... that's why it's 11 not 10). IIRC this is the same thing dbplyr does when you print a `tbl_sql`.

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,774 @@
...
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[R](https://www.r-project.org/) users familiar with [dplyr](https://dplyr.tidyverse.org/) and other packages in the [Tidyverse](https://www.tidyverse.org/) are likely find Ibis familiar.\n",
+    "In fact, some Ibis features were even inspired by similar features in the [Tidyverse](https://www.tidyverse.org/).\n",
+    "\n",
+    "However, due to differences between Python and R and the design and goals of Ibis itself, you may notice some big differences right away:\n",
+    "\n",
+    "- **No pipe:** The handy [magrittr pipe](https://magrittr.tidyverse.org/) (`%>%`) or R's newer native pipe (`|>`) don't exist in Python so you instead have to chain sequences of operations together with a period (`.`).\n",
```
ianmcook, line 20: Is it maybe worth explaining here that `.` with Python objects is like `$` with S3 objects R? (It is the operator used to access attributes and methods of an object.)

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,774 @@
...
+    "In Ibis:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.filter(_.skin_color == \"light\")"
```
ianmcook, line 221: This is the first place here where Ibis's underscore is used. I think it would be good to briefly explain why it's used. Maybe just something like:
> As mentioned above, Python lacks non-standard evaluation, so we cannot use bare words to refer to column names like we can with dplyr. In Ibis, the columns of a table are represented as attributes of the table object, so you can access them using the syntax `tbl.some_column`. But notice here that we use an underscore (`_`) instead of the table name. The `_` in Ibis represents the table returned by the previous step in the chained sequence of operations.

ianmcook: D'oh, I see that this is already explained below :)

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,774 @@
...
+   "source": [
+    "starwars.filter([_.skin_color == \"light\", _.eye_color == \"brown\"])"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In previous code, we used the `_` helper we imported earlier. The `_` is shorthand for the table we referenced at the beginning of the expression (in this case, `starwars`). We could have also written the more verbose form,\n",
```
ianmcook, line 252: I don't think this is strictly correct. I think it is more accurate to say:
> The `_` is shorthand for the table returned by the previous step in the chained sequence of operations.

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,774 @@
...
+   "outputs": [],
+   "source": [
+    "starwars.order_by(_.height)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You might notice that while dplyr puts missing values at the end, Ibis places them at the top.\n",
```
ianmcook, line 325: > `Ibis places them at the top`

Does this vary depending on what backend you use?

ianmcook:
One gotcha with Ibis that might trip up users coming from dplyr is when a column name collides with the name of a method or attribute of the Ibis table object. For example if you have a column named `count`. In this case you need to use `_["count"]` instead of `tbl.count` or `_.count`. It might be worth mentioning that somewhere.

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,774 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To start using dplyr in R we would run:\n",
+    "\n",
+    "```r\n",
+    "library(dplyr)\n",
+    "```\n",
+    "\n",
+    "When loading Ibis, it's useful to first load Ibis:"
```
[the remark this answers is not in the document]
amoeba: 😬. thanks :)

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,774 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "starwars.select(\"hair_color\", \"skin_color\", \"eye_color\")"
+   ]
```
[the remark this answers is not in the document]
amoeba: Thanks, changed.

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,774 @@
...
+    "        .select(\"name\", \"height_m\")\n",
+    ")"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "A big difference between dplyr's `mutate` and Ibis' is that, in Ibis, you have to chain separate `mutate` calls together you reference newly-created columns in the same `mutate` whereas in dplyr, you can put them all in the same call. This makes Ibis' `mutate` more similar to `transform` in base R.\n",
```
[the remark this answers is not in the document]
amoeba: Yep! Thanks.

docs/ibis-for-dplyr-users.ipynb
amoeba: Ah, the 11 vs 10 thing is a nice trick. Accepted basically what you wrote. Thanks.

docs/ibis-for-dplyr-users.ipynb
amoeba: Sure, I think it's safe to assume a person who has only used R/tidyverse will find `.` is new syntax.

docs/ibis-for-dplyr-users.ipynb
amoeba: Good catch, thanks!

docs/ibis-for-dplyr-users.ipynb
amoeba: @gforsyth do you happen to know? I'm guessing yes based on what @cpcloud said in review,

> Most databases allow you to choose null ordering either as a global configuration option, in the expression itself (e.g., NULLS LAST) or both. So far nobody has asked for this functionality so we haven't implemented it.

gforsyth: Yeah, I'm pretty sure this is backend-dependent.  We haven't yet exposed a way to configure it (and I don't know that there's a consensus on whether `FIRST` or `LAST` is a "better" default)

amoeba:
Thanks for the review @gforsyth, I took all of your suggestions as-is and left one question for you about NULL ordering.

Thanks also for the review @ianmcook! I took all of your suggestions and left one conversation unresolved. Your point about name collisions on columns is a good one. It didn't feel high-level enough to cover at the top so I put it as a note in the `select` section.

docs/ibis-for-dplyr-users.ipynb
amoeba: Thanks @gforsyth, I clarify the text here.

amoeba:
Thanks all. I've rebased the branch off main and I think all feedback has been incorporated without issue. This is ready for a final review if anyone wants.

docs/ibis-for-dplyr-users.ipynb
ianmcook: It would be good to clarify that even when `ibis.options.interactive = True`, calling `execute()` returns a pandas DataFrame with _all_ the result rows, not just the first 10.

Maybe add an example showing how you can store the result as a Python variable which is a pandas DataFrame?

amoeba: I think that makes sense. I added some text and an example in 0609126f44e347461919e3a216c874abd86fd879.

amoeba:
I addressed one last bit of feedback from @ianmcook in 0609126f44e347461919e3a216c874abd86fd879 and this is again ready for review/merge.

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,835 @@
...
+   "source": [
+    "starwars"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In addition to printing a nicely-formatted table and automatically calling `execute()`, setting `ibis.options.interactive` to `True` also causes our query to be limited to 10 rows. To get Ibis to give us all rows, we can directly call `execute()` and even save the result as a Pandas DataFrame:"
```
ianmcook, line 146: pandas should be all lowercase everywhere
```suggestion
    "In addition to printing a nicely-formatted table and automatically calling `execute()`, setting `ibis.options.interactive` to `True` also causes our query to be limited to 10 rows. To get Ibis to give us all rows, we can directly call `execute()` and even save the result as a pandas DataFrame:"
```

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,835 @@
...
+   "source": [
+    "starwars_df = starwars.execute()"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Which then gives us all of the data as a Pandas DataFrame:"
```
ianmcook, line 163:
```suggestion
    "Which then gives us all of the data as a pandas DataFrame:"
```

docs/ibis-for-dplyr-users.ipynb
```diff
@@ -0,0 +1,835 @@
...
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Next Steps\n",
+    "\n",
+    "Now that you've gotten an introudction to the common differences between dplyr and Ibis, head over to [Getting Started with ibis](https://ibis-project.org/getting_started/) for a full introduction. If you're familiar with SQL, check out [Ibis for SQL Programmers](https://ibis-project.org/ibis-for-sql-programmers/). If you're familiar with pandas, take a look at [Ibis for Pandas Users](https://ibis-project.org/ibis-for-pandas-users/.)"
```
ianmcook, line 810:
```suggestion
    "Now that you've gotten an introudction to the common differences between dplyr and Ibis, head over to [Getting Started with ibis](https://ibis-project.org/getting_started/) for a full introduction. If you're familiar with SQL, check out [Ibis for SQL Programmers](https://ibis-project.org/ibis-for-sql-programmers/). If you're familiar with pandas, take a look at [Ibis for pandas Users](https://ibis-project.org/ibis-for-pandas-users/.)"
```

amoeba: Good eye on these, thanks.

ianmcook (approved):

gforsyth (approved):
Thanks @amoeba and @fmichonneau !

This looks great and we should :shipit: 

Can you squash the commits down and format the message using [conventional commit](https://ibis-project.org/community/contribute/02_workflow/#writing-the-commit) style?  

Maybe something like `docs(intro): create Ibis for dplyr users document`

amoeba:
Squashed and amended, thanks @gforsyth. See https://github.com/ibis-project/ibis/pull/6050/commits/0a96ee45003ed00c66dd752d50f84bae5944f9e9.

cpcloud:
Fixed up the pre-commit issue! Merging!

PR #6053, opened 2023-04-20
refactor(ops.Hash): remove `how` from backend-specific hash operation

gforsyth:
BREAKING CHANGE: The `hash` method on table columns on longer accepts the `how` argument.  The hashing functions available are highly backend-dependent and the intention of the hash operation is to provide a fast, consistent (on the same backend, only) integer value. If you have been passing in a value for `how`, you can remove it and you will get the same results as before, as there were no backends with multiple hash functions working.

I was lazy and made the duckdb test a snapshot.  

And I think `sipHash64` is the fastest cross-platform hashing thing available (only because city hash is, I think, not available on conda for windows)

Also I can't run the bigquery tests right now since the auth got messed up -- I think I've got it refactored correctly but can't verify that at the moment.


ibis/expr/types/generic.py
```diff
@@ -63,20 +63,17 @@ def type(self) -> dt.DataType:
         """Return the [DataType] of this expression."""
         return self.op().output_dtype
 
-    def hash(self, how: str = "fnv") -> ir.IntegerValue:
+    def hash(self) -> ir.IntegerValue:
```
jcrist, line 66: Do we have a generic backend test that check that the output type is integral, and consistent across two executions? Something like:

```python
def test_hash(t):
    h1 = t.x.hash().execute()
    h2 = t.x.hash().execute()
    assert h1 == h2
    assert h1.dtype == "i8"  # or should this be "u8"? Do we care about resolution or only it's an integer?
```

jcrist (commented):
Overall :+1: from me.

ibis/backends/impala/compiler.py
```diff
@@ -17,7 +22,7 @@ def _get_join_type(self, op):
 
 
 class ImpalaExprTranslator(ExprTranslator):
-    _registry = {**operation_registry, **binary_infix_ops}
+    _registry = {**operation_registry, **binary_infix_ops, **{ops.Hash: hash}}
```
cpcloud, line 25: Could probably even do `**{ops.Hash: unary("fnv_hash")}` so you can kill the `hash` function entirely.

cpcloud (approved):

ibis/expr/types/generic.py
gforsyth: :+1: added and also added the op across a few more backends.

cpcloud (approved):

gforsyth:
forgot to `xfail` snowflake on this, but I think it's probably easy to add support

gforsyth:
well, "unit test" but

```
[ins] In [8]: t = con.tables.FUNCTIONAL_ALLTYPES

[ins] In [9]: t.string_col.hash().execute(limit=10)
Out[9]: 
0     158476100978179184
1   -6433072210751770779
2    8569864530471244163
3    -366024230374334232
4    3019900512728790636
5    2438387698619647920
6   -1717717285064877015
7    1433831815616377146
8    -643495880244649435
9    3677501896321825106
Name: Hash(string_col), dtype: int64

[ins] In [10]: ibis.get_backend(t)
Out[10]: <ibis.backends.snowflake.Backend at 0x7fb450dff640>
```

PR #6055, opened 2023-04-21
chore(deps): clean up the nix setup a bit

cpcloud:
This PR cleans up the nix setup a bit, making a couple tests sandbox aware and bumping the version of clang we use.


kszucs (approved):
LGTM

PR #6057, opened 2023-04-21
test(bigquery): return the client instead of a bogus SQL result

cpcloud:
This PR fixes the failing BigQuery CI.


cpcloud:
Passing the basic linters, merging!

PR #6061, opened 2023-04-21
refactor(common): use an enum as a sentinel value instead of NoMatch class

### LINKED ISSUE #6049: refactor(common): use an enum as a sentinel value instead of NoMatch class
This pattern plays better with type checkers, and makes it clearer that `NoMatch` is a singleton *value* rather than a singleton type (i.e. consumers should check for `is NoMatch` rather than `isinstance(x, NoMatch)`). 

```python
class NoMatchType(enum.Enum):
    NoMatch = "NoMatch"

NoMatch = NoMatchType.NoMatch
```

This way if we ever wanted to type annotate the `match` methods below, their signature could be:

```python
def match(self, value: Value) -> Value | NoMatchType:
    ...
```

See https://peps.python.org/pep-0484/#support-for-singleton-types-in-unions for reference.

_Originally posted by @jcrist in https://github.com/ibis-project/ibis/pull/6041#discussion_r1172829530_

mesejo:
fixes #6049

### DIFF
--- a/ibis/common/patterns.py
+++ b/ibis/common/patterns.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import enum
 import inspect
 import math
 import numbers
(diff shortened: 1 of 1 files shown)


cpcloud (approved):
LGTM! Thanks!

PR #6072, opened 2023-04-24
test(deps): remove aiohttp in favor of plain requests + threads

cpcloud:
This PR replaces our use of aiohttp (which was only used in the druid backend test setup) with an equivalent setup using plain requests and with threads. Local testing revealed that the asyncio was really only buying us IO concurrency data load query and that any intra-data-load-query concurrency was not noticable. Using threads gives us concurrency at the right level with less complexity.


kszucs (approved):
LGTM

PR #6073, opened 2023-04-24
feat(common): add `Topmost` and `Innermost` pattern matchers

kszucs:
These matchers enable matching node hierarchies in either top-down or bottom-up fashion, e.g.:

```py
Topmost(InstanceOf(ops.Reduction)) # to locate the reduction node if any
Innermost(InstanceOf(ops.TableNode))  # to locate the root table
```

### DIFF
--- a/ibis/common/patterns.py
+++ b/ibis/common/patterns.py
@@ -920,8 +920,7 @@ class Object(Matcher):
     __slots__ = ("type", "field_patterns")
 
     def __init__(self, type, *args, **kwargs):
-        match_args = getattr(type, "__match_args__", tuple())
-        kwargs.update(dict(zip(args, match_args)))
+        kwargs.update(dict(zip(type.__match_args__, args)))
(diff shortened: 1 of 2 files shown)


jcrist (approved):
The code here looks good to me here in isolation (tests, docs, implementation, ...). I don't think I understand how these abstractions will be used well enough to comment on the overall design.

cpcloud:
I think the idea here is to replace some of the analysis functions with these matchers.

Taking the last function in `ibis/expr/analysis.py` (`find_toplevel_aggs`):

```
def find_toplevel_aggs(nodes: Iterable[ops.Node]) -> Iterator[ops.Table]:
    def finder(node):
        return (
            isinstance(node, ops.Value),
            node if isinstance(node, ops.Reduction) else None,
        )

    return g.traverse(finder, nodes, filter=ops.Node)
```

I think that becomes

```
# apply this multiple times
Topmost(InstanceOf[ops.Reduction])
```

PR #6074, opened 2023-04-24
fix(pandas): solve problem with first and last window function

### LINKED ISSUE #4918: bug: `first` returns different values across backends
### What happened?

I would expect `first` (or `last`) of a table column to return a singular value and not a pd.Series of the same value. 

``` python
import os
import sqlite3

import ibis
import pandas as pd


def create_sqlite(df, name):
    filename = "tmp.sqlite"
    os.remove(filename)
    con = sqlite3.Connection(filename)
    df.to_sql(name, con, index=False)
    return ibis.sqlite.connect(filename).table(name)


df = pd.DataFrame(
    {"string": ["Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun"]},
)
sqlite_table = create_sqlite(df, "table")
pandas_table = ibis.pandas.connect({"df": df}).table("df")

pandas_table["string"], sqlite_table["string"]
pandas_table["string"].first().execute()
sqlite_table["string"].first().execute()
```

image

3.2.0

Pandas and Sqlite

### Relevant log output

_No response_

### Code of Conduct

mesejo:
### Implementation 
In the case of the first function, I chose to implement it with np.repeat as opposed to np.zeros for two reasons:

1. To use the same function as for the last.
2. Using repeat avoids the need for specifying a dtype.

fixes #4918

### DIFF
--- a/ibis/backends/pandas/execution/window.py
+++ b/ibis/backends/pandas/execution/window.py
@@ -500,7 +500,7 @@ def execute_series_lead_lag_timedelta(
 
 @execute_node.register(ops.FirstValue, pd.Series)
 def execute_series_first_value(op, data, **kwargs):
-    return data.values[0]
+    return data.iloc[np.repeat(0, len(data))]
 
 
 def _getter(x: pd.Series | np.ndarray, idx: int):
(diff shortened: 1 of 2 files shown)


cpcloud:
The current failures are unrelated to this PR.

PR #5899, opened 2023-04-02
refactor(ir): glue patterns and rules together 

kszucs:
Depends on https://github.com/ibis-project/ibis/pull/5824

Replace the previous rules-based validation system with the new pattern
matching system. This enables to use type annotations for operation
definitions as well as better error handling. Also lays the groundwork
to enable static type checking in the future.

BREAKING CHANGE: the `ibis.common.validators` module has been removed
and all validation rules from `ibis.expr.rules`, either use typehints
or patterns from `ibis.common.patterns`

PR #6075, opened 2023-04-24
refactor(ir): remove `rlz.column_from`, `rlz.base_table_of` and `rlz.function_of` rules

kszucs:
Simplify relational operations by enforcing input arguments already bound to the parent relation using `bind_expr()`

This change is pulled out from https://github.com/ibis-project/ibis/pull/5899


cpcloud (approved):
🚢 it!

PR #6082, opened 2023-04-25
refactor(api): deprecate tuple syntax for order by keys

kszucs:


cpcloud:
Can you make this a `depr(api)` commit so it shows up under the `Deprecations` header in the changelog?

kszucs:
@cpcloud the tpch build error should be unrelated

PR #6087, opened 2023-04-25
test(bigquery): rename client fixture to con

cpcloud:
Fixes the failing tests upstream


cpcloud:
Since we can only test upstream, I'm going to merge this in to `master`.

PR #6094, opened 2023-04-26
feat(api): exact median

cpcloud:
This PR adds a `median` API to numeric columns

### DIFF
--- a/ibis/backends/tests/test_aggregation.py
+++ b/ibis/backends/tests/test_aggregation.py
@@ -108,8 +108,8 @@ aggregate_test_params = [
                 "datafusion",
                 "impala",
                 "mysql",
-                "pyspark",
                 "mssql",
+                "pyspark",
                 "trino",
                 "druid",
             ],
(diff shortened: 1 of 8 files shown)


ibis/backends/tests/test_aggregation.py
```diff
@@ -94,20 +94,22 @@ def mean_udf(s):
...
+                    "datafusion",
+                    "impala",
+                    "mysql",
+                    "mssql",
+                    "trino",
+                    "druid",
+                ],
+                raises=com.OperationNotDefinedError,
+            ),
+            pytest.mark.notimpl(["pyspark"], raises=AttributeError),
```
gforsyth, line 111: You've got the compile dispatch for `ops.Mode` in the `pyspark` backend above, but that is saying that the tests aren't hitting it and you've got the `notimpl` here.  That seems odd.  Should we remove the `compile` rule altogether?

gforsyth (approved):
Looks good, just the one question around the `pyspark` mode stuff

ibis/backends/tests/test_aggregation.py
cpcloud: Hm, yeah. It's implemented in pyspark 3.4, but we're not testing that version yet because it breaks a ton of stuff.

cpcloud: Median too, so I'll remove that.

cpcloud: Fixed.

gforsyth (approved):

PR #6095, opened 2023-04-26
fix(duckdb): ensure connection lifetime is greater than or equal to record batch reader lifetime

cpcloud:
This PR fixes an issue where a RecordBatchReader produced by DuckDB
via ibis can outlive the ibis backend, and therefore its connection, causing a variety of fun problems.

I saw all of these errors while debugging this:

- double free
- corrupt doubly linked list
- sigsegv
- sigill (illegal instruction)
- sigabrt

Ultimately it was looking at a core dump from GDB showing the failing call was
a decref on an apparently already-freed DataFrame (from a `PandasScan` node)
that convinced me there was still a lifetime issue.

The approach I took, which I believe should be sufficient to guarantee that the
connection lives at least as long as the RecordBatchReader, is to use a
generator inside of a context manager created by `sa.Engine.begin()`. That way,
the generator's termination is commensurate with `__exit__` from the context
manager and a decref of the backend's connection object ensureing that the
generator holds a reference to the connection as long as it needs to.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -3,14 +3,15 @@
 from __future__ import annotations
 
 import ast
+import contextlib
 import os
 import warnings
-import weakref
 from functools import partial
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Iterable, Iterator, Mapping, MutableMapping
 
 import duckdb
+import pyarrow as pa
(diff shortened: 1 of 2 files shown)


cpcloud:
Thanks to @prmoore77 for finding this bug!

kszucs (approved):
LGTM

Issue #5982, opened 2023-04-12
docs: printing a memtable with a map freezes in interactive mode, document allowed input


wjones127:
### What happened?

```
import ibis
ibis.options.interactive = True
t = ibis.memtable({"a": [ibis.map({"a": 1, "b": 2})]})
t
```

Freezes up my Python process and doesn't print anything.

### What version of ibis are you using?

On current master branch.

### What backend(s) are you using, if any?

the default one, so DuckDB?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
I'm not sure how aggressive we want to be about this. Recursively scanning every element of every column might be prohibitively expensive, but maybe there's a less costly approach?

wjones127:
It does work without the `ibis.map` wrapper if there is only one valid element. With more than one I get:

```python
>>> t = ibis.memtable({"a": [{"a": 1, "b": 2}, {"a": 2}, None]})
>>> t
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/willjones/Documents/ibis/ibis/expr/types/core.py", line 56, in __repr__
    simple_console.print(self)
  File "/Users/willjones/mambaforge/envs/ibis-dev/lib/python3.10/site-packages/rich/console.py", line 1698, in print
    extend(render(renderable, render_options))
  File "/Users/willjones/mambaforge/envs/ibis-dev/lib/python3.10/site-packages/rich/console.py", line 1310, in render
    render_iterable = renderable.__rich_console__(self, _options)  # type: ignore[union-attr]
  File "/Users/willjones/Documents/ibis/ibis/expr/types/relations.py", line 266, in __rich_console__
    table = to_rich_table(self, width)
  File "/Users/willjones/Documents/ibis/ibis/expr/types/pretty.py", line 229, in to_rich_table
    result = table.limit(nrows + 1).execute()
  File "/Users/willjones/Documents/ibis/ibis/expr/types/core.py", line 303, in execute
    return self._find_backend(use_default=True).execute(
  File "/Users/willjones/Documents/ibis/ibis/backends/base/sql/__init__.py", line 254, in execute
    result = self.fetch_from_cursor(cursor, schema)
  File "/Users/willjones/Documents/ibis/ibis/backends/duckdb/__init__.py", line 690, in fetch_from_cursor
    return schema.apply_to(df)
[...]

PR #6097, opened 2023-04-27
docs: clarify map refers to Python Mapping container

anjakefala:
This is a proposed docstring change, and I am very much open to this PR being rejected.

While working on this: https://github.com/ibis-project/ibis/issues/5982 it took me a bit of time to understand the behaviour of `ibis.map` because there exists an [expr.types.ArrayValue.map](https://ibis-project.org/api/expressions/collections/?h=map#ibis.expr.types.arrays.ArrayValue.map) that has different behaviour.

The [docstring for ibis.map](https://ibis-project.org/api/expressions/top_level/?h=mapscalar#ibis.map) confused me more. Where was the lambda?

Speaking with a colleague, I learned about `mapping` being used as a concept to describe an object where there is a mapping between keys and values: https://docs.python.org/3/glossary.html#term-mapping. I have only been familiar with the functional definition, and not the object-category definition. I think because this colleague is familiar with `C++ std::map<>` they grokked it faster.

I am proposing we just add a small link in the docstring to the object-category definition. It would have helped me realise my confusion a lot sooner. I am not sure if this is the proper way to add a link, or if it would be better to add a definition instead of a link. I am also open to it being rare that somebody else would be confused by this, and the PR not being merged in.

### DIFF
--- a/ibis/expr/types/maps.py
+++ b/ibis/expr/types/maps.py
@@ -396,7 +396,7 @@ def map(
     keys: Iterable[Any] | Mapping[Any, Any] | ArrayColumn,
     values: Iterable[Any] | ArrayColumn | None = None,
 ) -> MapValue:
-    """Create a map expression.
+    """Create a [map container object](https://docs.python.org/3/glossary.html#term-mapping).
 
     If the `keys` and `values` are Python literals, then the output will be a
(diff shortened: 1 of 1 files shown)


cpcloud (approved):
LGTM!

Thanks for clarifying this.

Issue #6101, opened 2023-04-27
feat: Do not use Positional Arguments in ClickHouse GROUP BY


JackFielding:
### Is your feature request related to a problem?

It would be preferable if the generated ClickHouse SQL did not use positional arguments for `GROUP BY`, e.g. `GROUP BY 1, 2` and used the columns names instead, e.g. `GROUP BY col_1, col_2`.

Using positional arguments rather than named arguments has the following inconveniences:
* The generated SQL queries are invalid for ClickHouse versions prior to [21.10](https://clickhouse.com/docs/en/whats-new/changelog/2021#clickhouse-release-v2110-2021-10-16).
* For ClickHouse versions from 21.10 but before [22.7](https://clickhouse.com/docs/en/whats-new/changelog/2022#upgrade-notes-1) the user must explicitly specify the [enable_positional_arguments](https://clickhouse.com/docs/en/operations/settings/settings#enable-positional-arguments) setting otherwise the queries are invalid. e.g. pass `settings={"enable_positional_arguments": True}` to the backend.

Using named rather than positional arguments would be valid in all ClickHouse versions.

In the discussion: https://github.com/ibis-project/ibis/discussions/6099 I had a go at hacking the code and found what appeared to be a solution.

### Describe the solution you'd like

As per: https://github.com/ibis-project/ibis/discussions/6099. Using the code:

```python
import ibis

t = ibis.table(dict(a="int", b="str", c="str"), name="t")
expr = t.group_by(["b", "c"]).agg(t["a"].sum().name("hi"))
ibis.show_sql(expr, dialect="clickhouse")
```

instead of generating the query:

```sql
SELECT
  t0.b,
  t0.c,
  SUM(t0.a) AS hi
FROM t AS t0
GROUP BY
  1,
  2
```

it would be preferable if ibis generated the query:

```sql
SELECT
  t0.b,
  t0.c,
  SUM(t0.a) AS hi
FROM t AS t0
GROUP BY
  t0.b,
  t0.c
```

### What version of ibis are you running?

5.0.1

### What backend(s) are you using, if any?

ClickHouse

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #6106, opened 2023-04-28
fix(clickhouse): use named rather than positional group by

JackFielding:
Closes: https://github.com/ibis-project/ibis/issues/6101

### DIFF
--- a/ibis/backends/clickhouse/compiler/relations.py
+++ b/ibis/backends/clickhouse/compiler/relations.py
@@ -79,8 +79,8 @@ def _aggregation(op: ops.Aggregation, *, table, **kw):
     selections = (by + metrics) or "*"
     sel = sg.select(*selections).from_(table)
 
-    if by:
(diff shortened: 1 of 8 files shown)


JackFielding:
I'll stop work on this for now because it's clear that in the tests using positional arguments was intentional. It would probably benefit from a look over from someone more familiar with `ibis`.

gforsyth (changes requested):
Hey @JackFielding -- thanks for putting this in!  This looks like a reasonable change to me -- the failing tests are SQL snapshot tests and this change is (understandably) changing that output to something semantically equivalent but not explicitly equal.

You can update the snapshot files by running `pytest --snapshot-update`

cpcloud (approved):
Fixed up the snapshots, LGTM!

cpcloud:
Thanks for the PR @JackFielding!

JackFielding:
Thank you both for your help!

PR #6111, opened 2023-04-28
fix(bigquery): truncate when casting float to int

sudonhim:
BREAKING CHANGE: when using the bigquery backend, floats will be truncated instead of rounded to the nearest integer

### DIFF
--- a/ibis/backends/bigquery/registry.py
+++ b/ibis/backends/bigquery/registry.py
@@ -56,6 +56,12 @@ def bigquery_cast_interval_to_integer(compiled_arg, from_, to):
     return f"EXTRACT({from_.resolution.upper()} from {compiled_arg})"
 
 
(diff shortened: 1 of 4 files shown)


cpcloud:
Does this PR supersede #6110? If so, can you close #6110? Thanks!

sudonhim:
@cpcloud Done! GitHub newbie here, I didn't realize draft PRs were public :)

cpcloud:
@sudonhim No worries, happy to help!

Are you able to test this against BigQuery?

We have testing set up to run against `master`, but we aren't currently able to test it locally (which will be possible soon, but unclear exactly when).

cpcloud:
Impala failures are unrelated.

> I'm assuming that matching Python (like Pandas does) is correct here, hence this change.

Yep!

@tswast Mind having a look at this?

tswast (approved):
Thanks so much!

PR #6117, opened 2023-04-29
docs: change plural variable name cols to col

### LINKED ISSUE #6115: bug: pivot_longer() doesn't support iterable of strings
### What happened?

The docstring claims this should work, but it only works with a single str or a full fledged selector. 

PS I love the names pivot_wider and pivot_longer. So much better than the melt and explode. Thanks for adding this function, this would have been a doozy for me to implement myself.

main

NA

### Relevant log output

_No response_

### Code of Conduct

mesejo:
fixes #6115


cpcloud (approved):
Thanks!

PR #6121, opened 2023-05-01
test(impala): fix udf/uda error testing

cpcloud:
Fixes failing upstream impala tests.


cpcloud:
Merging on green to get CI passing again.

PR #6123, opened 2023-05-01
feat(pandas): add approx_median

mesejo:
### Background
In the PR [6094](https://github.com/ibis-project/ibis/pull/6094), the exact median operation has been introduced to numeric columns. This allows various backends to support both approx_median and median operations. It would be suitable to support approx_median for pandas, similar to PostgreSQL. Furthermore, some backends implement approx_median via the median, such as polars.

### Implementation
The approx_median has been implemented for pandas through the exact median.

### Notes
If this PR is accepted the same could be done for the Dask backend.

Apologies if this contribution makes no sense 😅

### DIFF
--- a/ibis/backends/pandas/execution/generic.py
+++ b/ibis/backends/pandas/execution/generic.py
@@ -916,6 +916,23 @@ def execute_mode_series_groupby(_, data, mask, aggcontext=None, **kwargs):
     return aggcontext.agg(data, mode)
 
 
+@execute_node.register(ops.ApproxMedian, pd.Series, (pd.Series, type(None)))
+def execute_approx_median_series(_, data, mask, aggcontext=None, **kwargs):
+    return aggcontext.agg(
+        data[mask] if mask is not None else data, lambda x: x.median()
+    )
+
+
+@execute_node.register(ops.ApproxMedian, SeriesGroupBy, (SeriesGroupBy, type(None)))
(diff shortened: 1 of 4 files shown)


cpcloud:
@mesejo Thanks for the PR!

I wonder if it makes sense to implement `ApproxMedian` for just Dask instead, since `ApproxMedian` doesn't really make sense for the pandas backend.

Curious what your thoughts are.

mesejo:
Hey @cpcloud,  thanks for sharing your feedback.

I completely understand your point about implementing `ApproxMedian` for pandas. Since the code ultimately uses the exact median, it would be more appropriate for someone to use the `Median` instead when working with pandas.

However, I do see the value in having the `ApproxMedian` for pandas given that one of the strengths of Ibis is the ability to switch between backends easily. In a scenario where the production environment uses BigQuery (or any other backend that doesn't support the median function), while the test/development environment uses Pandas, having `ApproxMedian` as an alias of `Median` would allow me to test the code with Pandas without altering it.

Of course, there is the effort associated with maintaining `ApproxMedian` for pandas, but that would probably be low. 

An alternative would be to truly implement an `ApproxMedian` for pandas using [t-digest](https://github.com/CamDavidsonPilon/tdigest), the same algorithm DuckDB uses (see [here](https://duckdb.org/docs/sql/aggregates.html#approximate-aggregates)). 

What do you think?

PR #6124, opened 2023-05-01
feat(duckdb): warn when querying an already consumed RecordBatchReader

### LINKED ISSUE #6062: feat(duckdb): should we error/warn when executing against an already exhausted RecordBatchReader?
Currently `con.read_in_memory` can register a table from a pyarrow `RecordBatchReader`. However, record batch readers are stateful objects (like reading a file/socket/...), only allowing a single pass over the data. Subsequent computations using the same table will result in either an error or (more likely) a result with 0 rows since the batch stream was already consumed.

```python
In [1]: import ibis

In [2]: import pyarrow as pa, pandas as pd

In [3]: df = pd.DataFrame({"x": range(10), "y": range(10, 20)})

In [4]: pyarrow_table = pa.Table.from_pandas(df)

In [5]: con = ibis.duckdb.connect()

In [6]: reader = pyarrow_table.to_reader()  # a RecordBatchReader

In [7]: t = con.read_in_memory(reader)

In [8]: t.schema()  # Schema is inferred properly
Out[8]: 
ibis.Schema {
  x  int64
  y  int64
}

In [9]: t.count().execute()  # First execution works fine
Out[9]: 10

In [10]: t.count().execute()  # Second execution is possibly unexpected
Out[10]: 0
```

If you have a firm understanding of how `pyarrow.RecordBatchReader`, `duckdb`, and `ibis` are fitting together here, then the above behavior makes sense. However, I've now seen 2 separate users tripped up by this behavior.

Is there a way that we could make subsequent executions against `t` warn that the `RecordBatchReader` has already been exhausted? Or should that be an error instead? I think this would be a nice way of informing the user of the issue without preventing streaming over a reader in cases where that's useful.

jcrist:
Fixes #6062.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -125,6 +125,7 @@ class Backend(BaseAlchemyBackend):
             # the progress bar causes kernel crashes in jupyterlab ¯\_(ツ)_/¯
             dbapi_connection.execute("SET enable_progress_bar = false")
 
+        self._record_batch_readers_consumed = {}
         super().do_connect(engine)
 
     def _load_extensions(self, extensions):
@@ -395,13 +396,15 @@ class Backend(BaseAlchemyBackend):
             con.connection.register(table_name, dataset)
 
     def read_in_memory(
-        self, dataframe: pd.DataFrame | pa.Table, table_name: str | None = None
+        self,
+        source: pd.DataFrame | pa.Table | pa.RecordBatchReader,
+        table_name: str | None = None,
     ) -> ir.Table:
-        """Register a Pandas DataFrame or pyarrow Table as a table in the current database.
+        """Register a Pandas DataFrame or pyarrow object as a table in the current database.
 
         Parameters
         ----------
-        dataframe
+        source
             The data source.
         table_name
             An optional name to use for the created table. This defaults to
@@ -414,7 +417,12 @@ class Backend(BaseAlchemyBackend):
         """
         table_name = table_name or util.gen_name("read_in_memory")
         with self.begin() as con:
-            con.connection.register(table_name, dataframe)
+            con.connection.register(table_name, source)
+
+        if isinstance(source, pa.RecordBatchReader):
+            # Ensure the reader isn't marked as started, in case the name is
+            # being overwritten.
+            self._record_batch_readers_consumed[table_name] = False
 
         return self.table(table_name)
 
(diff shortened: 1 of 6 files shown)


ibis/backends/duckdb/__init__.py
```diff
@@ -412,9 +415,16 @@ def read_in_memory(
...
+
         table_name = table_name or util.gen_name("read_in_memory")
         with self.begin() as con:
-            con.connection.register(table_name, dataframe)
+            con.connection.register(table_name, source)
+
+        if isinstance(source, pa.RecordBatchReader):
+            # Ensure the reader isn't marked as started, in case the name is
+            # being overwritten.
+            self._record_batch_reader_states[table_name] = False
```
jcrist, line 427: Note that items in this dict are never removed (but they are overwritten). This means that a key-value (`str -> bool`) pair is kept for every unique `RecordBatchReader` name ever queried with a connection. I think this is fine, but wanted to point out the implications. I tried doing some weakref stuff to automatically clear, but duckdb drops the root source `RecordBatchReader` once consumed, so the record that it was already consumed goes away before it's ever referenced to generate the warning.

ibis/backends/duckdb/__init__.py
```diff
@@ -534,6 +544,25 @@ def attach_sqlite(
                 sa.text(f"CALL sqlite_attach('{str(path)}', overwrite={overwrite})")
             )
 
+    def _register_in_memory_tables(self, expr: ir.Expr) -> None:
+        from ibis.expr.analysis import find_physical_tables
+
+        # Warn for any tables depending on RecordBatchReaders that have already
+        # started being consumed.
+        for t in find_physical_tables(expr.op()):
```
jcrist, line 552: This hook needs to be run before every query, same as `_register_in_memory_tables`. I piggy-backed on the same method here for now, but perhaps we should refactor to a general `_pre_execute(expr: Expr) -> None` hook, the default implementation of which just calls `_register_in_memory_tables`, just so we don't add stuff that isn't memtable related here.

ibis/expr/analysis.py
```diff
@@ -118,6 +118,18 @@ def reduction_to_aggregation(node):
     return agg
 
 
+def find_physical_tables(node):
```
jcrist, line 121: I suspect there's already a method for doing this, but I couldn't find one.

ibis/backends/duckdb/__init__.py
```diff
@@ -534,6 +544,25 @@ def attach_sqlite(
...
+            started = self._record_batch_reader_states.get(t.name)
+            if started is True:
+                warnings.warn(
+                    f"Table {t.name!r} is backed by a `pyarrow.RecordBatchReader` "
+                    "that has already been partially consumed. This may lead to "
+                    "incorrect results. Either recreate the table from a new "
+                    "`pyarrow.RecordBatchReader`, or use `con.create_table` to "
```
cpcloud, line 559: Perhaps we should recommend `con.create_table` or `.cache()` here as the "obvious" thing to do to persist the thing (one permanent, one temporary but not ephemeral).

ibis/backends/duckdb/__init__.py
cpcloud: Yeah, +1 on refactoring this to something slightly more generic.

ibis/backends/duckdb/__init__.py
```diff
@@ -125,6 +125,7 @@ def configure_connection(dbapi_connection, connection_record):
             # the progress bar causes kernel crashes in jupyterlab ¯\_(ツ)_/¯
             dbapi_connection.execute("SET enable_progress_bar = false")
 
+        self._record_batch_reader_states = {}
```
cpcloud, line 128: Can we call this something that indicates what the values represent? Maybe `_readers_consumed`?

ibis/backends/duckdb/__init__.py
```diff
@@ -534,6 +544,25 @@ def attach_sqlite(
...
+
+        # Warn for any tables depending on RecordBatchReaders that have already
+        # started being consumed.
+        for t in find_physical_tables(expr.op()):
+            started = self._record_batch_reader_states.get(t.name)
+            if started is True:
+                warnings.warn(
+                    f"Table {t.name!r} is backed by a `pyarrow.RecordBatchReader` "
+                    "that has already been partially consumed. This may lead to "
+                    "incorrect results. Either recreate the table from a new "
```
cpcloud, line 558:
```suggestion
                    "unpredictable results. Either recreate the table from a new "
```

Or perhaps `surprising`? I feel like `incorrect` is maybe overselling the severity a bit. Not super opinionated on this though.

cpcloud (approved):
Mostly just nitpicks, otherwise LGTM!

cpcloud (approved):

cpcloud:
Thanks!

PR #6137, opened 2023-05-03
fix(impala): clean up internal cursor usage

### LINKED ISSUE #6128: bug: Impala returns a cursor for operations that are expected to not return a cursor
### What happened?

Impala returns a cursor for operations that are expected to not return a cursor. This is a problem because ibis.backends.base.sql.raw_sql will release empty cursors and a lot of code relies on this behaviour, i.e. a lot of code doesn't capture the returned cursor and release it. For example, When you create a table this code (https://github.com/ibis-project/ibis/blob/master/ibis/backends/impala/__init__.py#L626) calls raw_sql and ignored the returned cursor. This violates the warning here: https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/__init__.py#L127

The cursor then goes out of scope and when garbage collection runs it will be closed and removed from the connection-pool - but in the meantime the connection-pool has to create new connections or can run out of connections if a lot of operations (that don't usually return cursors) are performed (e.g. during test setup/teardown when we are creating/dropping tables/databases). I.e. https://github.com/ibis-project/ibis/blob/2.1.1/ibis/backends/impala/client.py#L122 ( this is replaced in main by sqlalchemy QueuePool)

A hacky work-around is something like this - but it far from satisfactory:

```
class ImpalaCursor(ibis_impala_client.ImpalaCursor):
    def __bool__(self):
        # In ibis.backends.base.sql.raw_sql the code automatically releases any cursor that evaluates to False.
        # The (implied) intention is that operations that return no rows, like CREATE and DROP, should return
        # an empty cursor. Impala does return a cursor for these types of operations with one "row" that says
        # some like "The table has been created" (or similar). We can't examine the value of the rows fetched
        # as this would affect callers who wish to fetch the result - so instead we look for queries that
        # just return summary information.
        # REFRESH does indeed seem to return no cursor, hence the check for None
        # Note - will break for a query like
        #  SELECT CAST(COUNT(*) AS STRING) AS summary FROM __ibis_tmp.ibis_tmp_pandas_8bcdd4082ff0477ba4a11fb0a0cd12f4
        if (
            self._cursor is None
            or self._cursor.description is None
            or len(self._cursor.description) == 1
            and self._cursor.description[0]
            == (
                "summary",
                "STRING",
                None,
                None,
                None,
                None,
                None,
            )
        ):
            return False
        return True
```

It would be nice if there was some way to turn off these messages from Impala, but I couldn't find a way to do this. It may be related to the fact that sometimes these operations are long running and thus can return an asynchronous completion handle? (or whatever it is called)

2.1.1

Impala

### Relevant log output

```sh
N/A - I can provide logging but and a traceback but it will be for 2.1.1 code, so the failure mode on master will be different. The warning on https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/__init__.py#L127 is still being violated though by code like https://github.com/ibis-project/ibis/blob/2.1.1/ibis/backends/impala/client.py#L122
```

### Code of Conduct

cpcloud:
This PR tries to address #6128 by more aggressively cleaning up our use of raw_sql internally. Closes #6128.

### DIFF
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -131,11 +131,7 @@ class BaseSQLBackend(BaseBackend):
         query
             DDL or DML statement
         """
-        cursor = self.con.execute(query)
-        if cursor:
-            return cursor
-        cursor.release()
-        return None
+        return self.con.execute(query)
 
     @contextlib.contextmanager
     def _safe_raw_sql(self, *args, **kwargs):
--- a/ibis/backends/impala/__init__.py
+++ b/ibis/backends/impala/__init__.py
@@ -8,6 +8,7 @@ import operator
 import os
 import re
 import weakref
+from functools import cached_property
 from pathlib import Path
 from posixpath import join as pjoin
 from typing import TYPE_CHECKING, Any, Literal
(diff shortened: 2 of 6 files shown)


cpcloud:
Merging to help the flakiness in the impala test suite.

PR #6138, opened 2023-05-03
feat(api): allow passing multiple keyword arguments to `ibis.interval`

### LINKED ISSUE #5983: feat: support 2+ arguments to `ibis.interval()`

In Ibis, if you want to add 1 year + 1 month + 1 day to a date, you need to do:
```py
ibis.date('2023-04-12') + ibis.interval(years=1) + ibis.interval(months=1) + ibis.interval(days=1)
```
It would be nice if you could instead do:
```py
ibis.date('2023-04-12') + ibis.interval(years=1, months=1, days=1)
```
But this is not implemented and returns:
```
ValueError: Exactly one argument is required
```

Support passing multiple arguments to `ibis.interval()` and add the units together the construct the returned interval.

5.0.0

_No response_

### Code of Conduct

jcrist:
Also does some general UX cleanups and test coverage for this utility.

Fixes #5983.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -717,6 +717,7 @@ def _time_from_deferred(value: Deferred) -> Deferred:
 def interval(
     value: int | datetime.timedelta | None = None,
     unit: str = 's',
+    *,
     years: int | None = None,
     quarters: int | None = None,
     months: int | None = None,
@@ -734,7 +735,7 @@ def interval(
     Parameters
     ----------
     value
-        Interval value. If passed, must be combined with `unit`.
+        Interval value.
     unit
         Unit of `value`
     years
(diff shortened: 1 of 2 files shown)


ibis/expr/api.py
```diff
@@ -717,6 +717,7 @@ def _time_from_deferred(value: Deferred) -> Deferred:
 def interval(
     value: int | datetime.timedelta | None = None,
     unit: str = 's',
+    *,
```
cpcloud, line 720: It might be good to note this as a breaking change. Up to you.

cpcloud (approved):
LGTM.

ibis/expr/api.py
```diff
@@ -765,37 +766,46 @@ def interval(
...
+
+    # If no components, default to 0 s
+    if not components:
+        components.append((0, "s"))
+
+    intervals = [
+        literal(v, type=dt.Interval(u, value_type=literal(v).type()))
+        for v, u in components
+    ]
+    return functools.reduce(operator.add, intervals)
```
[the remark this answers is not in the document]
jcrist: `sum` always adds in the `start` value (`0`), which we don't want. `functools.reduce` only includes values from the iterable, and won't call the reducer in the case of a single value (what we want).

ibis/expr/api.py
jcrist: Calling `interval(1, 's', 2)` wouldn't have worked before either. Changing the signature just makes the lack of this working more visibly explicit.

PR #6144, opened 2023-05-03
fix(bigquery): bigquery timestamp and datetime dtypes

### LINKED ISSUE #6143: bug: bigquery TIMESTAMP type can support timezone
### What happened?

BigQuery has [TIMESTAMP](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp_type) and [DATETIME](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#datetime_type) types. Here `DATETIME` is the none timezone version, while the `TIMESTAMP` is `UTC` timezone. However, the latest version of ibis takes `TIMESTAMP` as none timezone and misses `DATETIME` type.

0.14.0

BigQuery

### Relevant log output

_No response_

### Code of Conduct

chelsea-lin:
Fixes #6143 

BREAKING CHANGE:
Before this change, ibis `timestamp` is mapping to Bigquery `TIMESTAMP` type and no timezone supports. However, it's not correct, BigQuery `TIMESTAMP` type should have `UTC` timezone, while `DATETIME` type is the no timezone version. Hence, this change is breaking the ibis `timestamp` mapping to BigQuery:
- If ibis `timestamp` has the `UTC` timezone, will map to BigQuery `TIMESTAMP` type 
- If ibis `timestamp` has no timezone, will map to BigQuery `DATETIME` type

### DIFF
--- a/ibis/backends/bigquery/datatypes.py
+++ b/ibis/backends/bigquery/datatypes.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from multipledispatch import Dispatcher
 
+import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
 
 ibis_type_to_bigquery_type = Dispatcher("ibis_type_to_bigquery_type")
@@ -54,9 +55,14 @@ def trans_date(t):
 
 @ibis_type_to_bigquery_type.register(dt.Timestamp)
 def trans_timestamp(t):
-    if t.timezone is not None:
-        raise TypeError("BigQuery does not support timestamps with timezones")
-    return "TIMESTAMP"
+    if t.timezone is None:
+        return "DATETIME"
+    elif t.timezone == 'UTC':
+        return "TIMESTAMP"
+    else:
+        raise com.UnsupportedOperationError(
+            "BigQuery does not support timestamps with timezones other than 'UTC'"
+        )
 
 
 @ibis_type_to_bigquery_type.register(dt.DataType)
@@ -86,7 +92,7 @@ def trans_json(t):
 def trans_geography(t):
     if (t.geotype, t.srid) == ("geography", 4326):
         return "GEOGRAPHY"
-    raise TypeError(
+    raise com.UnsupportedOperationError(
         "BigQuery geography uses points on WGS84 reference ellipsoid."
         f"Current geotype: {t.geotype}, Current srid: {t.srid}"
     )
--- a/ibis/backends/bigquery/registry.py
+++ b/ibis/backends/bigquery/registry.py
@@ -376,9 +376,10 @@ def _geo_simplify(translator, op):
 
 
 STRFTIME_FORMAT_FUNCTIONS = {
-    dt.Date: "DATE",
-    dt.Time: "TIME",
-    dt.Timestamp: "TIMESTAMP",
+    dt.date: "DATE",
+    dt.time: "TIME",
+    dt.Timestamp(timezone=None): "DATETIME",
+    dt.Timestamp(timezone="UTC"): "TIMESTAMP",
 }
 
 
@@ -404,19 +405,26 @@ def compiles_strftime(translator, op):
     arg = op.arg
     format_str = op.format_str
     arg_type = arg.output_dtype
-    strftime_format_func_name = STRFTIME_FORMAT_FUNCTIONS[type(arg_type)]
+    strftime_format_func_name = STRFTIME_FORMAT_FUNCTIONS[arg_type]
     fmt_string = translator.translate(format_str)
     arg_formatted = translator.translate(arg)
-    if isinstance(arg_type, dt.Timestamp):
+    if isinstance(arg_type, dt.Timestamp) and arg_type.timezone is None:
+        return "FORMAT_{}({}, {})".format(
+            strftime_format_func_name,
+            fmt_string,
+            arg_formatted,
+        )
+    elif isinstance(arg_type, dt.Timestamp):
         return "FORMAT_{}({}, {}, {!r})".format(
             strftime_format_func_name,
             fmt_string,
             arg_formatted,
-            arg_type.timezone if arg_type.timezone is not None else "UTC",
+            arg_type.timezone,
         )
-    return "FORMAT_{}({}, {})".format(
-        strftime_format_func_name, fmt_string, arg_formatted
-    )
+    else:
+        return "FORMAT_{}({}, {})".format(
+            strftime_format_func_name, fmt_string, arg_formatted
+        )
 
 
 def compiles_string_to_timestamp(translator, op):
--- a/ibis/backends/tests/test_param.py
+++ b/ibis/backends/tests/test_param.py
@@ -12,6 +12,11 @@ import ibis
 import ibis.expr.datatypes as dt
 from ibis import _
 
+try:
+    from google.api_core.exceptions import BadRequest as GoogleBadRequest
+except ImportError:
+    GoogleBadRequest = None
+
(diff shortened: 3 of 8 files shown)


ibis/backends/bigquery/datatypes.py
```diff
@@ -54,9 +54,14 @@ def trans_date(t):
...
 def trans_timestamp(t):
-    if t.timezone is not None:
-        raise TypeError("BigQuery does not support timestamps with timezones")
-    return "TIMESTAMP"
+    if t.timezone is None:
+        return "DATETIME"
+    elif t.timezone == 'UTC':
+        return "TIMESTAMP"
+    else:
+        raise TypeError(
```
cpcloud, line 62:
```suggestion
        raise com.UnsupportedOperationError(
            "BigQuery does not support timestamps with timezones other than 'UTC'"
        )
```

ibis/backends/tests/test_param.py
```diff
@@ -116,21 +116,21 @@ def test_scalar_param_map(con):
             "timestamp",
             "timestamp_col",
             id="string_timestamp",
-            marks=pytest.mark.notimpl(["druid"]),
+            marks=pytest.mark.notimpl(["druid", "bigquery"]),
```
cpcloud, line 119: Are these regressions since they were passing before? Why are they now failing?

ibis/backends/bigquery/tests/unit/snapshots/test_compiler/test_union/False/out.sql
```diff
@@ -1,7 +1,7 @@
 SELECT t0.`index`, t0.`Unnamed_0`, t0.`id`, t0.`bool_col`, t0.`tinyint_col`,
        t0.`smallint_col`, t0.`int_col`, t0.`bigint_col`, t0.`float_col`,
        t0.`double_col`, t0.`date_string_col`, t0.`string_col`,
-       t0.`timestamp_col`, t0.`year`, t0.`month`
+       t0.`datetime_col`, t0.`timestamp_col`, t0.`year`, t0.`month`
```
cpcloud, line 4: It looks like you may have modified this table upstream in BigQuery.

I'd like to keep this schema fixed if possible. If you need to get codegen, it might be best to use a one-off table that doesn't require a table to exist in BigQuery.

You should be able to use unbound tables for this:

```python
t = ibis.table({"a": "timestamp", "b": "timestamp('UTC')"}, name="t")
# test codegen using `t`
```

ibis/backends/tests/test_temporal.py
```diff
@@ -253,6 +253,11 @@ def test_timestamp_extract_milliseconds(backend, alltypes, df):
     raises=AttributeError,
     reason="'StringColumn' object has no attribute 'epoch_seconds'",
 )
+@pytest.mark.broken(
+    ["bigquery"],
+    raises=GoogleBadRequest,
+    reason="UNIX_SECONDS does not support for DATETIME arguments",
```
cpcloud, line 259:
```suggestion
    reason="UNIX_SECONDS does not support DATETIME arguments",
```

ibis/backends/tests/test_temporal.py
```diff
@@ -1432,6 +1427,11 @@ def test_interval_add_cast_column(backend, alltypes, df):
     raises=AttributeError,
     reason="'StringColumn' object has no attribute 'strftime'",
 )
+@pytest.mark.broken(
+    ["bigquery"],
+    raises=GoogleBadRequest,
+    reason="FORMAT_TIMESTAMP does not support for DATETIME arguments",
```
cpcloud, line 1433:
```suggestion
    reason="FORMAT_TIMESTAMP does not support DATETIME arguments",
```

Is there a `FORMAT_DATETIME` function we could call instead?

cpcloud (changes requested):
@chelsea-lin Thanks for the PR!

Couple of questions and a few changes requested.

Overall looks good.

ibis/backends/tests/test_param.py
chelsea-lin: The time without timezone would be converted into `DATETIME` instead of `TIMESTAMP` after this change. Hence, the new behavior is breaking the tests. I added the error reasons for more details.

ibis/backends/tests/test_temporal.py
chelsea-lin: Done.

ibis/backends/bigquery/datatypes.py
chelsea-lin: Done!

ibis/backends/tests/test_param.py
```diff
@@ -116,21 +121,42 @@ def test_scalar_param_map(con):
...
         param(
             datetime.datetime(2009, 1, 20, 1, 2, 3),
             "timestamp",
             "timestamp_col",
             id="datetime_timestamp",
-            marks=pytest.mark.notimpl(["druid"]),
+            marks=[
+                pytest.mark.notimpl(["druid"]),
+                pytest.mark.broken(
+                    ["bigquery"],
```
cpcloud, line 155: We can make these pass, but we'll need to adjust the column type of `timestamp_col` in `functional_alltypes`. That can be done in a follow up.

cpcloud (changes requested):
@chelsea-lin This is close!

Can you add a `BREAKING CHANGE` section to the commit describing a bit about what has changed? A sentence of two is fine!

After that passes the commit linter then this is good to go!

cpcloud:
Also @chelsea-lin can you run the BigQuery tests with

```
pytest -m bigquery -n auto --dist loadgroup --snapshot-update --allow-snapshot-deletion
```
?

I'm not currently able to run them. They are running in CI, but that only runs after a PR is merged.

ibis/backends/bigquery/registry.py
```diff
@@ -382,19 +383,26 @@ def compiles_strftime(translator, op):
     arg = op.arg
     format_str = op.format_str
     arg_type = arg.output_dtype
-    strftime_format_func_name = STRFTIME_FORMAT_FUNCTIONS[type(arg_type)]
+    strftime_format_func_name = STRFTIME_FORMAT_FUNCTIONS[arg_type]
```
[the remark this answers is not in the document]
chelsea-lin: Thanks to catch that. I was wondering if it will break the `dt.date` and `dt.time` but tests look good. Fixed as your suggested.

ibis/backends/tests/test_param.py
chelsea-lin: Yeah, I checked `functional_alltypes` is defined at `ibis/backends/conftest.py` for all backends. But I didn't find a way to specify `timestamp_col` as `timestamp` with timezone, and I was worried about breaking other backend so that I didn't make it. Feel free to let me know if I can help for follow up here.

chelsea-lin:
> Also @chelsea-lin can you run the BigQuery tests with
> 
> ```
> pytest -m bigquery -n auto --dist loadgroup --snapshot-update --allow-snapshot-deletion
> ```
> 
> ?
> 
> I'm not currently able to run them. They are running in CI, but that only runs after a PR is merged.

The test runs well in my side:
```
$ pytest -m bigquery -n auto --dist loadgroup --snapshot-update --allow-snapshot-deletion
...
976 passed, 1 skipped, 202 xfailed in 260.05s (0:04:20) 
```

cpcloud (approved):
LGTM! Thanks!

cpcloud:
Failing test is unrelated! Merging!

PR #6146, opened 2023-05-04
test(backend): add xfail test for memtable with null strings

NickCrews:
The nulls are parsed into `"<NA>"`:

```
AssertionError: assert ['a', '<NA>', 'c'] == ['a', None, 'c']
```

run with

```
pytest ibis/backends/tests/test_generic.py -m duckdb -k test_memtable_null_str
```

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -923,6 +923,17 @@ def test_memtable_bool_column(backend, con, monkeypatch):
     backend.assert_series_equal(t.a.execute(), pd.Series([True, False, True], name="a"))
 
 
+@pytest.mark.parametrize("dtype", [None, "string"])
+def test_memtable_null_str(backend, con, monkeypatch, dtype):
+    """Ensure that null strings are translated to null, not string literal '<NA>'"""
+    monkeypatch.setattr(ibis.options, "default_backend", con)
+
+    strings = ["a", None, "c"]
+    df = pd.DataFrame({"a": strings}, dtype=dtype)
+    t = ibis.memtable(df)
+    assert t.a.execute().tolist() == strings
+
+
 @pytest.mark.notimpl(["datafusion"], raises=com.OperationNotDefinedError)
 @pytest.mark.broken(
     ["druid"],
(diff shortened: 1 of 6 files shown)


ibis/backends/tests/test_generic.py
```diff
@@ -916,6 +916,46 @@ def test_memtable_bool_column(backend, con, monkeypatch):
...
+)
+@pytest.mark.broken(
+    ["druid"], reason="returns empty string for None", raises=AssertionError
+)
+def test_memtable_null_str_default_dtype(con):
+    """Ensure that null strings are translated to some kind of null value."""
+    t = ibis.memtable(pd.DataFrame({"a": ["a", None, "c"]}))
+    result = con.execute(t.a)
+    assert result.iat[0] == "a"
+    assert pd.isna(result.iat[1])
```
NickCrews, line 930: why the general `pd.isna`? What is stopping us from doing `assert result.iat[1] is None`? Adding a comment explaining would be nice

cpcloud: Some backends (it may just be impala, but I don't really want to fix that right now) do not return `None` and return `np.nan`.

ibis/expr/datatypes/value.py
```diff
@@ -267,9 +267,14 @@ def infer_shapely_multipolygon(value) -> dt.MultiPolygon:
 
 def normalize(typ, value):
     """Ensure that the Python type underlying a literal resolves to a single type."""
-    if value is None:
+    import pandas as pd
+
+    if value is None or isinstance(value, type(pd.NA)):
```
NickCrews, line 272: could this use the same `pd.isna()` check that is used in the test?

ibis/tests/strategies.py
```diff
@@ -187,3 +188,6 @@ def memtable(draw, schema=schema(primitive_dtypes)):  # noqa: B008
 
     df = draw(dataframe)
     return ibis.memtable(df)
+
+
+null_values = st.sampled_from([None, pd.NA])
```
NickCrews, line 193: Should this include `np.nan`?

cpcloud: I don't think so, since that implies we're endorsing `np.nan` as NULL value. While it is used that way in pandas, it's not like that in any other system that isn't based on pandas.

I hope that doesn't sound too ideological. I'm trying to avoid making the mistake of supporting `np.nan` as NULL everywhere and setting expectations according.

ibis/expr/datatypes/value.py
cpcloud: I don't think so, because I'm intentionally not including `NaN` in this check.

cpcloud:
I don't think this is viable at the moment, closing.

PR #6149, opened 2023-05-04
refactor(pandas): solve and remove stale TODOs

mesejo:
`Expr` are hashable, introduced in the commit c6fb0c040 
`JSON` is no longer a child from `String`, introduced in the commit 34f389825
pandas now return `float64` columns


cpcloud (approved):
LGTM! Thank you.

PR #6151, opened 2023-05-04
refactor(datatypes): use `pyarrow.infer_type` instead of constructing an throwaway array

cpcloud:


cpcloud:
Unfortunately we need to traverse the entire array, and `pa.infer_type` appears to only look at the first element.

PR #6156, opened 2023-05-05
feat(snowflake): create an ibis backend from a snowpark session

### LINKED ISSUE #5877: feat(snowflake): run ibis code inside of stored procedure

Writing stored procedures for Snowflake using Python [depends on the Snowpark API](https://docs.snowflake.com/en/sql-reference/stored-procedures-python), centering around a [session object](https://docs.snowflake.com/ko/developer-guide/snowpark/reference/python/api/snowflake.snowpark.Session.html) rather than a database connection.

To use Ibis for data wrangling within a Snowflake stored procedure, a new backend is required. It would presumably use `Session.sql` as the interface between SQLAlchemy-based expressions and Snowpark API (which has its own query builder).

5.0.0

Snowflake

### Code of Conduct

### LINKED ISSUE #6153: feat(snowflake): connect to snowflake using existing snowpark session/client

We use Ibis for a bunch of things. We also use Snowflake. The great news is that Ibis and Snowflake integration are really solid. However, we've been using more and more code inside Snowflake itself (in the Snowpark python env). Unfortunately, right now it is difficult to make use of Ibis inside Snowflake. We think we've gotten something working (hope to blog soon about it) but one problem we have is that Snowpark's [current_session()](https://docs.snowflake.com/ko/developer-guide/snowpark/reference/python/api/snowflake.snowpark.functions.current_session.html) can't be passed to Ibis as a connection for the Snowflake backend. 

Some people may ask why not just use Snowpark dataframe operations when inside Snowflake. It's a fair question. Unfortunately, having analysts need to learn multiple different data abstraction libraries is more work. Additionally, we're heavily invested in Substrait and Ibis's Substrait backend is super awesome.

We wish there was a way to just tell Ibis "use the snowpark session" to run your Snowflake operations.  Something akin to:
```
ibis.connect("snowflake://", snowpark_connection=current_session())
```

We're pretty sure that below the scenes Snowpark is using the python Snowflake connector that Ibis also uses so it seems like it is more glue than an entirely backend.

We'd run any version of Ibis to be able to do this.

Snowflake and Substrait

### Code of Conduct

cpcloud:
This PR implements the ability to connect to snowflake using an existing snowpark session.

Closes #6153.

### DIFF
--- a/ibis/backends/snowflake/__init__.py
+++ b/ibis/backends/snowflake/__init__.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, Any, Iterable, Mapping
 import pyarrow as pa
 import sqlalchemy as sa
 from snowflake.connector.constants import FIELD_ID_TO_NAME
+from snowflake.connector.converter import SnowflakeConverter
 from snowflake.sqlalchemy import ARRAY, OBJECT, URL
 
 import ibis.expr.datatypes as dt
@@ -87,6 +88,15 @@ return longest.map((_, i) => {
 })""",
     },
 }
+
+
+class IbisSnowflakeConverter(SnowflakeConverter):
+    # when using snowpark, con._paramstyle is set to qmark and is inconsistent
+    # with snowflake-sqlalchemy which uses pyformat this allows strings that
+    # are used in metadata queries to be used in qmark style queries which are
+    # supported by snowflake-sqlalchemy (just not the default)
+    def _quoted_name_to_snowflake_bindings(self, snowflake_type, value):
+        return self._str_to_snowflake_bindings(snowflake_type, value)
 
 
 class Backend(BaseAlchemyBackend):
@@ -192,7 +202,7 @@ $$ {defn["source"]} $$"""
             if authenticator is not None:
                 connect_args.setdefault("authenticator", authenticator)
 
-            # snowflake-connector-python does not handle `None` for password, but
+            # snowflake-connector-python does not handle `None` for passyword, but
             # accepts the empty string
             url = URL(
                 account=account,
@@ -207,7 +217,11 @@ $$ {defn["source"]} $$"""
         else:
             # we're connecting from a snowpark session most likely
             engine = sa.create_engine(
-                "snowflake://", creator=creator, poolclass=sa.pool.StaticPool
+                "snowflake://",
+                creator=creator,
+                connect_args=connect_args,
+                poolclass=sa.pool.StaticPool,
+                paramstyle="qmark",
             )
 
         @sa.event.listens_for(engine, "connect")
@@ -291,14 +305,12 @@ $$ {defn["source"]} $$"""
         └───────────┴───────┘
         """
         con = session._conn._conn
-        # workaround a misconfigured implementation of the DBAPI cursor
-        # when con._paramstyle is not set to "pyformat"
-        #
-        # when using snowpark, con._paramstyle is set to qmark
-        con._paramstyle = "pyformat"
-        return cls(
+        con.converter = IbisSnowflakeConverter()
+        backend = cls(
             database=f"{con.database}/{con.schema}", __creator__=lambda con=con: con
-        ).reconnect()
+        )
+        backend.reconnect()
+        return backend
 
     def to_pyarrow(
         self,
--- a/ibis/backends/snowflake/__init__.py
+++ b/ibis/backends/snowflake/__init__.py
@@ -94,6 +94,8 @@ class Backend(BaseAlchemyBackend):
     compiler = SnowflakeCompiler
     supports_create_or_replace = True
 
+    _top_level_methods = ("from_snowpark",)
+
     @functools.cached_property
     def _pandas_converter(self):
         from ibis.backends.snowflake.converter import SnowflakePandasConverter
@@ -136,9 +138,10 @@ $$ {defn["source"]} $$"""
 
     def do_connect(
         self,
-        user: str,
-        account: str,
+        *,
         database: str,
+        user: str | None = None,
+        account: str | None = None,
         password: str | None = None,
         authenticator: str | None = None,
         connect_args: Mapping[str, Any] | None = None,
@@ -175,6 +178,9 @@ $$ {defn["source"]} $$"""
             See https://docs.snowflake.com/en/developer-guide/python-connector/sqlalchemy#additional-connection-parameters
             for more details
         """
+        if connect_args is None:
+            connect_args = {}
+
         dbparams = dict(zip(("database", "schema"), database.split("/", 1)))
         if dbparams.get("schema") is None:
             raise ValueError(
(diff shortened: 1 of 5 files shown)


cpcloud:
Ran the test suite successfully using Python 3.8: ```
…/ibis-from-snowpark-branched on  ibis-from-snowpark-branched is 📦 v5.1.0 via 🐍 v3.8.16 via 🅒 snowpark
❯ pytest -m snowflake -n auto --dist loadgroup --snapshot-update --allow-snapshot-deletion -q
bringing up nodes...
...........................x...x.................x..x.xx.....x..x...................x................................xx.xx.xs..............xs.s..x......x......xs..x......x...................x..........x......xx....x..................................x.....x....xx...........x....................x.......................................x..........x..........................x.....x.......... [ 42%]
.....................................................x............................................x...x............xxx......x.........x.xxxx.x.....x....xx.xx......x...x..x...xx..xx....x.x..xx.x..x.x.x....x......x.x.............x.xx....xx............x................x............s...............................x..xx....xx.xx..x.......x....xxx.x.x....x..x.......x..x.x..x...x.x..xxxxxx.x.. [ 85%]
x...xx.....x..xx.......x.....x.x.x.....xx....x....x.x...x........x.x.xxx.x........x....s......x.x.sx......xx...x..x...x....x.............                                                                                                                                                                                                                                                             [100%]
769 passed, 9 skipped, 139 xfailed in 93.11s (0:01:33)
```

**Comment by cpcloud** [comment]
> I suppose one option is to commit to supporting 3.8 for one more major release (6.x), and if snowflake hasn't released 3.9+ support by 7.x we can rip out the snowflake backend and make it a third-party package and pin it to ibis 6.x

This sounds reasonable to me.

File: ibis/backends/snowflake/__init__.py [hunk 1]
```diff
@@ -221,7 +234,65 @@ def normalize_name(name):
...
+            con._paramstyle = "pyformat"
+            return con
+
+        return ibis.connect(
+            "snowflake://",
+            creator=creator,
+            user=con.user,
+            password=None,
+            database=f"{con.database}/{con.schema}",
+            account=con.host.replace(".snowflakecomputing.com", ""),
```
jacques-n, line 294: Rather than do this, I think you can call [get_current_account()](https://github.com/snowflakedb/snowpark-python/blob/b50e4fcf4e2875eb1d392253fdaf0b8c117ee215/src/snowflake/snowpark/session.py#L1705).

cpcloud: Unfortunately that doesn't include the cloud provider, just the `xx12345` part.

jacques-n: Should it include the extra stuff in this case? I suppose it depends on what ibis is doing with it. The other part is part of the url but I don't think of it as an account property. 

¯\\\_(ツ)_/¯

cpcloud: The only thing we definitely need is the `database=` argument, so maybe I'll just drop the rest.

cpcloud: Fixed, removed the unnecessary stuff.

IndexSeek:
The other day, we discussed possibly getting this working in Snowflake. I did some digging and tried to consider the dependencies and what would be required to get Ibis running within Snowflake. Anything not in the Anaconda channel would need to be staged and follow the guidelines at [Making Dependencies Available to your Code | Snowflake Documentation](https://docs.snowflake.com/en/developer-guide/upload-dependencies). 

I was reviewing the dependencies for 3.8, and here's what I came up with:

These packages are in the Anaconda channel.
- mdurl
- packaging
- toolz
- sqlglot
- idna
- pytz
- numpy
- multipledispatch
- requests
- certifi
- tqdm
- rich
- xxhash
- urllib3
- pandas
- platformdirs
- typing_extensions
- six

These packages are not in the Anaconda channel.
- pygments
- tzdata
- markdown_it
- parsy
- charset_normalizer
- public
- ibis
- tlz
- pooch
- bidict
- dateutil

jacques-n:
Sorry I didn't post more of my findings earlier on this post. I went all the way down the rabbit hole. There are good instructions from snowflake on how to package additional pure python libraries available. It worked pretty well for most missing libraries because they were all pure python. Ultimately we hit two problems that we got stuck on. 

1) getting snowflake-sqlalchemy to load is a nightmare since it shares the Snowflake namespace with other packages that are installed. (and it isn't) 
2) it turns out that the connection in snowpark when running inside Snowflake is incomplete. It wraps a stored procedure internally which means certain operations aren't possible. For example, you can't use bind variables. (I can find actual example line numbers if you'd like to see the code) 

Given these challenges, we ultimately gave up on trying to make ibis work with Snowpark in Snowflake.

cpcloud:
@IndexSeek It seems like getting `snowflake-sqlalchemy` into the conda channel should be relatively straightforward, no?

@jacques-n Regarding the second point, can you share some failing code?

jacques-n:
This is an example of parts of the client bailing in certain circumstances: 
https://github.com/snowflakedb/snowpark-python/blob/182069e50e52028236532ec4eeab3fb86a758954/src/snowflake/snowpark/session.py#LL1205C16-L1205C28

cpcloud:
@jacques-n Got it. Can you elaborate on how that limitation prevents you from using ibis inside of snowpark?

Let's say the packaging issues were addressed. Would you still be unable to use ibis inside of snowpark because of the second limitation?

jacques-n:
>Can you elaborate on how that limitation prevents you from using ibis inside of snowpark?

I'm not sure that it would. My main observation is that any application that assumes the Snowflake connection used by Snowpark inside Snowflake will fail. Things like bind, async, etc are not implemented. I'd guess that will break ibis features but I have not done an analysis to confirm.

cpcloud:
I've added a CI job that will run our test suite using the snowpark connection.

jacques-n:
> I've added a CI job that will run our entire test suite using the snowpark connection.

Inside Snowflake? All the limitations only exist if Snowpark/Snowflake connector use the execute via stored procedure path in the code. If you're using Snowpark outside of Snowflake, I don't think it is possible to use that path.

cpcloud:
No, just to see if the connection can be used to run our test suite outside of snowflake.

That at least shows that there's nothing in ibis preventing the original use case.

cpcloud:
I was able to get the entire backend test suite passing with this PR.

I think this is ready for review and merge.

ibis/backends/snowflake/__init__.py
```diff
@@ -239,7 +251,54 @@ def normalize_name(name):
...
+        │ allisdo01 │    27 │
+        │ ansonca01 │    16 │
+        └───────────┴───────┘
+        """
+        con = session._conn._conn
+        # workaround a misconfigured implementation of the DBAPI cursor
+        # when con._paramstyle is not set to "pyformat"
+        #
+        # when using snowpark, con._paramstyle is set to qmark
+        con._paramstyle = "pyformat"
```
jcrist, line 298: Does mutating the internals of the snow park session connection break anything if the user tries to reuse the snow park session for snow park stuff later (not sure if that workflow is possible/likely)?

ibis/backends/snowflake/__init__.py
```diff
@@ -239,7 +251,54 @@ def normalize_name(name):
                 return name
 
         self.con.dialect.normalize_name = normalize_name
-        return res
+        self.database_name = self._default_database = dbparams["database"]
+        self._default_schema = dbparams["schema"]
+        return eng
+
+    @classmethod
+    def from_snowpark(cls, session) -> Backend:
```
jcrist, line 259: If possible, might be nice to add an annotation for `session` here.

jcrist (approved):
Nice work!

ibis/backends/snowflake/__init__.py
cpcloud: Good question, I'm not 100% sure. I'll see if I can add a test that connects to ibis and triggers use of parameters with snowpark.

cpcloud: This does in fact break any further use of bind parameters in snowflake.

I can look into whether there's a way around this.

cpcloud: Ok, addressed this.

jcrist (approved):
:shipit:

cpcloud:
Closing this out. It's a good idea, but there's too many rough edges in the UX and implementation.

PR #6166, opened 2023-05-07
feat(oracle): support temporary tables

cpcloud:
This PR adds support for temporary tables to the oracle backend.

The implementation uses Oracle's [`GLOBAL TEMPORARY`] tables because they have
fewer restrictions than and work with many older versions of oracle than
private temporary tables.

For example, private temporary tables cannot hold `CLOB` or `BLOB` columns.

Unfortunately that means they come with some caveats that require some annoying hacks.

* Global temporary tables cannot be `DROP`ed with first `TRUNCATE`ing them. Oracle documentts this as

> A session becomes unbound to a temporary table with a TRUNCATE statement or at session termination...

* Global temporary tables' **data** are removed when a session ends, but the
  table's definition (its name and schema) are still visible to all users. I'm
  using `atexit` to handle this, but it's not ideal. I looked into using some
  of the SQLAlchemy pool events (`close`, `reset`, `invalidate`, and
  `soft_invalidate`) but none of them seemed to fire in concert with a session
  ending.

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -105,6 +105,7 @@ class BaseAlchemyBackend(BaseSQLBackend):
     table_class = AlchemyTable
     compiler = AlchemyCompiler
     supports_temporary_tables = True
+    _temporary_prefix = "TEMPORARY"
 
     def _build_alchemy_url(self, url, host, port, user, password, database, driver):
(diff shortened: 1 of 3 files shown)


kszucs:
> Global temporary tables cannot be DROPed with first TRUNCATEing them. Oracle documentts this as

How about dropping it immediately without truncating it first?

PR #6169, opened 2023-05-08
feat(deps): make `pooch` an optional dependency

cpcloud:
Some users running in production application environments may not want or need to load ibis examples on import. This PR moves the loading to the `ibis.examples` module `__getattr__` and defers importing pooch to that time as well. `pooch` is now an optional dependency that is installed by running `pip install ibis-framework[examples]`. Installing `pooch` separately is also valid.


jcrist (commented):
The implementation here seems fine, but I'm not sure if I agree with the motivation. This does mean that all the documentation examples only run if the user installed `ibis-framework[examples]`, which feels like a bit of a rough edge for new users.

Was there a user request for this somewhere? `pooch` is not a large dependency, it's negligible in size compared to other other required dependencies.

cpcloud:
@jcrist That's fair.

@jacques-n Hit this when trying to use ibis in the snowflake snowpark environment.

Maybe it would suffice to do keep the lazy loading in this PR, but not change dependencies?

jcrist:
To let someone manually install without a supposedly required dependency? I'd support that change, since it would only affect users opting into doing something weird.

jcrist (approved):

PR #6172, opened 2023-05-08
fix(common): fix bug in re_extract

### LINKED ISSUE #6167: bug: re_extract not working as expected for big query
### What happened?

For duckdb backend, the code works as expected:
```
import ibis
ibis.set_backend("duckdb")
t = ibis.memtable({"s": ["a|b|c", "b|a|c", "b|b|b|c|a"]})
expr = t.s.re_extract(get_regex("|", 1), 0).execute()
0    b|c
1    a|c
2    c|a
Name: RegexExtract(s, '([a-zA-Z0-9]+(?:\\|[a-zA-Z0-9]+){0,1})$', 0), dtype: object
```

But when I run in BQ, I get a different result:
```
0 | a|b|c|  
1 | b|a|c|  
2 | b|b|b|c|a
```

The compiled SQL statements are very different:
Duckdb:
```
CASE
    WHEN REGEXP_MATCHES(t0.s, '([a-zA-Z0-9]+(?:\\|[a-zA-Z0-9]+){0,1})$')
    THEN REGEXP_EXTRACT(t0.s, '([a-zA-Z0-9]+(?:\\|[a-zA-Z0-9]+){0,1})$')
    ELSE ''
  END
```
BQ:
```
CASE
    WHEN REGEXP_CONTAINS(t0.`s`, '([a-zA-Z0-9]+(?:\|[a-zA-Z0-9]+){0,1})$')
    THEN CASE
      WHEN COALESCE(0, 0) = 0
      THEN t0.`s`
      ELSE REGEXP_EXTRACT_ALL(t0.`s`, '([a-zA-Z0-9]+(?:\|[a-zA-Z0-9]+){0,1})$')[SAFE_ORDINAL(0)]
    END
    ELSE NULL
  END
```

5.1.0

BigQuery

### Relevant log output

_No response_

### Code of Conduct

mesejo:
### Background
The current definition of `re_extract` needs to be clarified. The documentation
claims that it works as `re.match`, but it works differently:

> when index is zero and there's a match, return the entire string, otherwise
> return the content of the index-th match group.

A Python's Match Object returns the entire **match**, not the entire **string**, from the [documentation](https://docs.python.org/3/library/re.html#re.Match.group) (**emphasis mine**):

> Without arguments, group1 defaults to zero (the whole match is returned). If a groupN argument is zero, the corresponding return value is the entire **matching string**;

 
For example:

```python
>>> import re
>>> pattern = re.compile("a(bc)d")
>>> match = pattern.match("abcd foo bar")
>>> match.group(0)
Out[5]: 'abcd'
>>> match.group(1)
Out[6]: 'bc'
```

### Implementation
This PR updates the documentation and the implementation of some of the backends
to actually work as a Match Object, note that some backends were already
returning the whole match, not the whole string: pandas, sqlite, duckdb, postgres.

1. pandas: use `re.search` instead of `re.match` to match anywhere on the string (`re.match` only works at the beginning)
2. clickhouse: use the [regexpExtract](https://clickhouse.com/docs/en/sql-reference/functions/string-search-functions#regexpextract) method that offers the expected functionality
3. bigquery: the index parameter makes no sense because it returns either the whole match or the first capturing group that matches.

fixes #6167

### DIFF
--- a/ibis/backends/bigquery/registry.py
+++ b/ibis/backends/bigquery/registry.py
@@ -163,8 +163,16 @@ def _regex_extract(translator, op):
     regex = _translate_pattern(translator, op.pattern)
     index = translator.translate(op.index)
     matches = f"REGEXP_CONTAINS({arg}, {regex})"
-    extract = f"REGEXP_EXTRACT_ALL({arg}, {regex})[SAFE_ORDINAL({index})]"
-    return f"IF({matches}, IF(COALESCE({index}, 0) = 0, {arg}, {extract}), NULL)"
+
+    extract = f'''
+    IF(
+    {index} = 0,
+    REGEXP_EXTRACT({arg}, {regex}),
+    REGEXP_REPLACE({arg}, CONCAT({regex}, ".*"), CONCAT("\\", CAST({index} AS STRING)))
+    )
+    '''
+
+    return f"IF({matches}, {extract}, NULL)"
 
 
 def _regex_replace(translator, op):
--- a/ibis/expr/types/strings.py
+++ b/ibis/expr/types/strings.py
@@ -1030,13 +1030,13 @@ class StringValue(Value):
         Parameters
         ----------
         pattern
-            Reguar expression pattern string
+            Regular expression pattern string
         index
             The index of the match group to return.
 
             The behavior of this function follows the behavior of Python's
-            [`re.match`](https://docs.python.org/3/library/re.html#match-objects):
-            when `index` is zero and there's a match, return the entire string,
+            [`match objects`](https://docs.python.org/3/library/re.html#match-objects):
+            when `index` is zero and there's a match, return the entire match,
             otherwise return the content of the `index`-th match group.
 
         Returns
--- a/ibis/backends/clickhouse/compiler/values.py
+++ b/ibis/backends/clickhouse/compiler/values.py
@@ -303,26 +303,18 @@ def _regex_search(op, **kw):
 @translate_val.register(ops.RegexExtract)
 def _regex_extract(op, **kw):
     arg = translate_val(op.arg, **kw)
-    pattern = translate_val(op.pattern, **kw)
-    index = "Null" if op.index is None else translate_val(op.index, **kw)
-
-    # arg can be Nullable, which is not allowed in extractAll, so cast to non
-    # nullable type
     arg = f"CAST({arg} AS String)"
 
-    # extract all matches in pattern
-    extracted = f"CAST(extractAll({arg}, {pattern}) AS Array(Nullable(String)))"
+    wrapped_op_pattern = op.pattern.copy(value="(" + op.pattern.value + ")")
+    pattern = translate_val(wrapped_op_pattern, **kw)
 
-    # if there's a match
-    #   if the index IS zero or null
-    #     return the full string
-    #   else
-    #     return the Nth match group
-    # else
-    #   return null
-    does_match = f"multiMatchAny({arg}, [{pattern}])"
-    idx = f"CAST(nullIf({index}, 0) AS Nullable(Int64))"
-    then = f"if({idx} IS NULL, {arg}, {extracted}[{idx}])"
+    then = f"extractGroups({arg}, {pattern})[1]"
+    if op.index is not None:
+        index = translate_val(op.index, **kw)
+        then = f"extractGroups({arg}, {pattern})[{index} + 1]"
+
+    does_match = f"notEmpty({then})"
+
     return f"if({does_match}, {then}, NULL)"
 
 
--- a/ibis/backends/clickhouse/tests/snapshots/test_functions/test_column_regexp_extract/out.sql
+++ b/ibis/backends/clickhouse/tests/snapshots/test_functions/test_column_regexp_extract/out.sql
@@ -1,1 +1,1 @@
-if(multiMatchAny(CAST(string_col AS String), ['[\\d]+']), if(CAST(nullIf(3, 0) AS Nullable(Int64)) IS NULL, CAST(string_col AS String), CAST(extractAll(CAST(string_col AS String), '[\\d]+') AS Array(Nullable(String)))[CAST(nullIf(3, 0) AS Nullable(Int64))]), NULL)
+if(notEmpty(extractGroups(CAST(string_col AS String), '([\\d]+)')[3 + 1]), extractGroups(CAST(string_col AS String), '([\\d]+)')[3 + 1], NULL)
--- a/ibis/backends/duckdb/registry.py
+++ b/ibis/backends/duckdb/registry.py
@@ -153,20 +153,13 @@ def _neg_idx_to_pos(array, idx):
 
 
 def _regex_extract(string, pattern, index):
-    result = sa.case(
-        (
-            sa.func.regexp_matches(string, pattern),
-            sa.func.regexp_extract(
-                string,
-                pattern,
-                # DuckDB requires the index to be a constant so we compile
-                # the value and inline it using sa.text
-                sa.text(str(index.compile(compile_kwargs=dict(literal_binds=True)))),
-            ),
-        ),
-        else_="",
+    return sa.func.regexp_extract(
+        string,
+        pattern,
+        # DuckDB requires the index to be a constant, so we compile
+        # the value and inline it by using sa.text
+        sa.text(str(index.compile(compile_kwargs=dict(literal_binds=True)))),
     )
-    return result
 
 
 def _json_get_item(left, path):
--- a/ibis/backends/pandas/execution/strings.py
+++ b/ibis/backends/pandas/execution/strings.py
@@ -277,15 +277,14 @@ def execute_series_ends_with(op, data, pattern, **kwargs):
 @execute_node.register(ops.RegexExtract, pd.Series, str, integer_types)
 def execute_series_regex_extract(op, data, pattern, index, **kwargs):
     pattern = re.compile(pattern)
-
-    def extract(x, pattern=pattern, index=index):
-        match = pattern.match(x)
-        if match is not None:
-            return match.group(index) or np.nan
-        return np.nan
-
-    extracted = data.apply(extract)
-    return extracted
+    return pd.Series(
+        [
+            None if (match is None or index > match.lastindex) else match[index]
+            for match in map(pattern.search, data)
+        ],
+        dtype=data.dtype,
+        name=data.name,
+    )
 
 
 @execute_node.register(ops.RegexExtract, SeriesGroupBy, str, integer_types)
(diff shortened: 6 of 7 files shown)


ibis/backends/bigquery/registry.py
```diff
@@ -161,10 +162,14 @@ def _regex_search(translator, op):
 def _regex_extract(translator, op):
     arg = translator.translate(op.arg)
     regex = _translate_pattern(translator, op.pattern)
-    index = translator.translate(op.index)
-    matches = f"REGEXP_CONTAINS({arg}, {regex})"
-    extract = f"REGEXP_EXTRACT_ALL({arg}, {regex})[SAFE_ORDINAL({index})]"
-    return f"IF({matches}, IF(COALESCE({index}, 0) = 0, {arg}, {extract}), NULL)"
+
+    if op.index is not None:
+        warnings.warn(
```
cpcloud, line 167: Can you raise an exception here instead?

ibis/expr/types/strings.py
```diff
@@ -1030,13 +1030,13 @@ def re_extract(
...
         pattern
-            Reguar expression pattern string
+            Regular expression pattern string
         index
             The index of the match group to return.
 
             The behavior of this function follows the behavior of Python's
-            [`re.match`](https://docs.python.org/3/library/re.html#match-objects):
-            when `index` is zero and there's a match, return the entire string,
+            [`Match Objects`](https://docs.python.org/3/library/re.html#match-objects):
```
cpcloud, line 1038:
```suggestion
            [`match objects`](https://docs.python.org/3/library/re.html#match-objects):
```

ibis/backends/clickhouse/tests/test_functions.py
```diff
@@ -386,16 +386,14 @@ def test_regexp(con, expr, expected):
         # (L('abcd').re_extract('(ab)(cd)', 1), 'cd'),
         # valid group number but no match => None
         param(L('abcd').re_extract(r'(\\d)', 0), None, id="valid_group_no_match"),
-        # match but not a valid group number => NULL
-        param(L('abcd').re_extract('abcd', 3), None, id="invalid_group_match"),
```
cpcloud, line 390: Hm, why did you remove this test?

cpcloud (changes requested):

ibis/backends/bigquery/registry.py
mesejo: Would ValueError be appropriated?

ibis/backends/clickhouse/tests/test_functions.py
mesejo: The `regexpExtract` of clickhouse raises an exception when an invalid group index is passed; I wasn't sure if I should xfail this test or create a new one asserting that an exception is raised. Which one should I use?

ibis/backends/bigquery/registry.py
cpcloud: I think `com.OperationNotDefinedError` is appropriate here.

ibis/backends/clickhouse/tests/test_functions.py
mesejo: How can I test BigQuery? The tests are not triggered and I don't see it in the [list](https://ibis-project.org/community/contribute/02_workflow/#setting-up-non-trivial-backends) of non-trivial backends.

cpcloud: Unfortunately we can only run the BigQuery tests against `master` right now.

cpcloud: This would be a breaking change for the ClickHouse backend. Is there a way we can preserve the behavior? Otherwise, we have make sure all the other backends also raise an exception when a match is out of bounds.

ibis/backends/bigquery/registry.py
mesejo: The index parameter is marked as required in the [docs](https://ibis-project.org/api/expressions/strings/?h=re_e#ibis.expr.types.strings.StringValue.re_extract). Should I raise an exception and update the docs?

cpcloud: It looks like bigquery allows one capture group, so it seems like `0` or `1` would be the allow values for `op.index`. It's probably best to let the database handle this one since this behavior is kind of an outlier.

mesejo: So I think the best option is to warn the user that the index parameter has no effect for BigQuery

cpcloud: I see. Ok, I think then the warning is probably the best option.

Thanks for working through this!

cpcloud: @tswast Can you take a look at this? Thanks!

cpcloud (approved):

tswast (changes requested):
Is there a way we can fallback to `REGEXP_EXTRACT_ALL` when index is non-zero?

tswast (changes requested):
I asked internally, and another BQ engineer suggested using `REGEXP_REPLACE`, which _does_ support multiple capture groups. I'll play around with some SQL to see if I can workaround this index issue.

tswast:
Other than the "at most 1 capturing group" restriction when index = 0, this appears to do the right thing:

```sql
SELECT 
IF(
  re_index = 0,
  REGEXP_EXTRACT(my_string, pattern),  
  REGEXP_REPLACE(my_string, CONCAT(pattern, ".*"), CONCAT("\\", CAST(re_index AS STRING)))
)
FROM UNNEST([
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("123.456.789.nomatch", "([0-9]+)\\.([0-9]+)\\.([0-9]+)", 1),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("123.456.789.nomatch", "([0-9]+)\\.([0-9]+)\\.([0-9]+)", 2),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("123.456.789.nomatch", "([0-9]+)\\.([0-9]+)\\.([0-9]+)", 3),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("123.456.789.nomatch", "[0-9]+\\.[0-9]+\\.[0-9]+", 0),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("nomatch.123.456.789.nomatch", "[0-9]+\\.[0-9]+\\.[0-9]+", 0),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("a|b|c", "([a-zA-Z0-9]+(?:\\|[a-zA-Z0-9]+){0,1})$", 0),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("b|a|c", "([a-zA-Z0-9]+(?:\\|[a-zA-Z0-9]+){0,1})$", 0),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("b|b|b|c|a", "([a-zA-Z0-9]+(?:\\|[a-zA-Z0-9]+){0,1})$", 0)
])
```

Row | f0_ |  
-- | -- | --
1 | 123 |  
2 | 456 |  
3 | 789 |  
4 | 123.456.789 |  
5 | 123.456.789 |  
6 | b\|c |  
7 | a\|c |  
8 | c\|a

Unless, we think `"nomatch.123.456.789.nomatch"` shouldn't capture anything, in which case we should also prepend `^` to the pattern to match Python's `match` method.

ibis/backends/bigquery/registry.py
tswast: See my latest comment on the main PR for an alternative that does support `index != 0`.

tswast: Ugh. I need to correct it for when there aren't matches. The `REGEXP_CONTAINS` logic thankfully doesn't care if there are > 1 matching groups.

```
SELECT 
IF(REGEXP_CONTAINS(my_string, pattern),
  IF(
  re_index = 0,
  REGEXP_EXTRACT(my_string, pattern),  
  REGEXP_REPLACE(my_string, CONCAT(pattern, ".*"), CONCAT("\\", CAST(re_index AS STRING)))
  ),
  NULL
)
FROM UNNEST([
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("123.456.789.nomatch", "([0-9]+)\\.([0-9]+)\\.([0-9]+)", 1),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("123.456.789.nomatch", "([0-9]+)\\.([0-9]+)\\.([0-9]+)", 2),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("123.456.789.nomatch", "([0-9]+)\\.([0-9]+)\\.([0-9]+)", 3),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("nomatch.nomatch.nomatch", "([0-9]+)\\.([0-9]+)\\.([0-9]+)", 3),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("123.456.789.nomatch", "[0-9]+\\.[0-9]+\\.[0-9]+", 0),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("nomatch.123.456.789.nomatch", "[0-9]+\\.[0-9]+\\.[0-9]+", 0),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("a|b|c", "([a-zA-Z0-9]+(?:\\|[a-zA-Z0-9]+){0,1})$", 0),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("b|a|c", "([a-zA-Z0-9]+(?:\\|[a-zA-Z0-9]+){0,1})$", 0),
  STRUCT<my_string STRING, pattern STRING, re_index INT64>("b|b|b|c|a", "([a-zA-Z0-9]+(?:\\|[a-zA-Z0-9]+){0,1})$", 0)
])
```

tswast (approved):

gforsyth:
Thanks for putting this in @mesejo !

I ran the bigquery tests locally and it looks like something is tripping up the `re_extract` tests:

```
FAILED ibis/backends/tests/test_string.py::test_string[bigquery-re_extract_posix] - google.api_core.exceptions.BadRequest: 400 Syntax error: Unclosed string literal at [6:79]
FAILED ibis/backends/tests/test_string.py::test_string[bigquery-re_extract_group] - google.api_core.exceptions.BadRequest: 400 Syntax error: Unclosed string literal at [6:104]
FAILED ibis/backends/tests/test_string.py::test_string[bigquery-re_extract_whole_group] - google.api_core.exceptions.BadRequest: 400 Syntax error: Unclosed string literal at [6:85]
FAILED ibis/backends/tests/test_string.py::test_string[bigquery-re_extract] - google.api_core.exceptions.BadRequest: 400 Syntax error: Unclosed string literal at [6:101]
```

cpcloud:
I have a fix for this that I can push up later today.

cpcloud:
@mesejo I'm going to close this PR out and open a new one with your original commits and my changes.

I can't easily push to the `master` branch on your fork :)

PR #6179, opened 2023-05-09
ci: expand sqlalchemy upper bound

cpcloud:
Removes the sqlalchemy2 upper bound due to a bug in duckdb-engine is that is [fixed](https://github.com/Mause/duckdb_engine/pull/642) in 0.7.1


cpcloud:
All green, merging!

PR #6185, opened 2023-05-10
fix(pyspark): enable joining on columns with different names as well as complex predicates

cpcloud:
This PR allows use of more complex join predicates in the pyspark backend.


cpcloud:
CI is more reliable now, though I don't know why coverage is suggesting none of the changed lines are covered. Running locally, those lines are all required to pass tests.

kszucs (approved):

PR #6188, opened 2023-05-10
feat(streamlit): add support for streamlit connection interface

cpcloud:
This PR adds a tiny implementation of the new experimental streamlit connection API, supporting streamlit apps that use this interface.

### DIFF
--- a/ibis/streamlit/__init__.py
+++ b/ibis/streamlit/__init__.py
@@ -1,0 +1,74 @@
+from __future__ import annotations
+
+from typing import Any
+
+from streamlit.connections import ExperimentalBaseConnection
+from streamlit.runtime.caching import cache_data
+
+import ibis
+from ibis.backends.base import BaseBackend
+
+__all__ = ["IbisConnection"]
+
+
+class IbisConnection(ExperimentalBaseConnection[BaseBackend]):
+    def _connect(self, **kwargs: Any) -> BaseBackend:
+        """Connect to the backend and return a client object.
+
+        This method is invoked when `st.experimental_connection` is called and
+        pulls information from streamlit secrets. `_connect` is part of the
+        streamlit connection API to be implemented by developers of specific
+        connection types.
+
+        Here's an example not-so-secret configuration:
+
+        ```toml
+        [connections.ch]
+        url = "clickhouse://play:clickhouse@play.clickhouse.com:9440/?secure=1"
+        ```
+
+        Alternatively, you can specify individual arguments under a connection
+        whose name matches the backend type. For example:
+
+        ```toml
+        [connections.clickhouse]
+        user = "play"
+        password = "clickhouse"
+        host = "play.clickhouse.com"
+        port = 9440
+        secure = 1
+        ```
+
+        This file can be placed at `~/.streamlit/secrets.toml`.
+
+        You can then connect to the backend using:
+
+        ```python
+        import streamlit as st
+
+        from ibis.streamlit import IbisConnection
+
+        con = st.experimental_connection("ch", type=IbisConnection)
+
+        # Now you can use `con` as if it were an ibis backend
+        con.list_tables()
+        ```
+        """
+        # add secrets config to kwargs
+        kwargs.update(self._secrets.to_dict())
+
+        # pop out the connection string
+        url = kwargs.pop("url", None)
+
+        # if there's no url parameter, use the connection name to get connection arguments
+        if url is None:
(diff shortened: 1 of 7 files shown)


cpcloud:
Here's an exmaple of tiny app that connects to the clickhouse playground using ibis:

### `~/.streamlit/secrets.toml`

```toml
[connections.ch]
url = "clickhouse://play:clickhouse@play.clickhouse.com:9440/?secure=1"
```

### `ibis_app.py`

```python
import streamlit as st
from ibis.streamlit import IbisConnection

con = st.experimental_connection("ch", type=IbisConnection)

st.title("Prepare to be Amazed!")

if selected := st.sidebar.selectbox("Table", options=con.list_tables()):
    try:
        t = con.table(selected)
        df = t.limit(50).execute()
    except Exception as e:
        st.error(e)
    else:
        st.header(f"The `{selected}` table!")
        st.dataframe(df)

        st.header(f"Information about `{selected}`!")
        st.dataframe(t.info().execute())
```

ibis/streamlit/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+import ibis
+from ibis.backends.base import BaseBackend
+
+__all__ = ["IbisConnection"]
+
+
+class IbisConnection(ExperimentalBaseConnection[BaseBackend]):
+    def _connect(self, **kwargs) -> BaseBackend:
+        kwargs.update(self._secrets.to_dict())
+        kwargs.pop("type", None)
```
jcrist, line 17: What is the `type` option we're ignoring here?

ibis/streamlit/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+from streamlit.runtime.caching import cache_data
+
+import ibis
+from ibis.backends.base import BaseBackend
+
+__all__ = ["IbisConnection"]
+
+
+class IbisConnection(ExperimentalBaseConnection[BaseBackend]):
+    def _connect(self, **kwargs) -> BaseBackend:
```
jcrist, line 15: A tiny docstring showing an example streamlit config (and what options are supported) would be useful in understanding this bit of code IMO.

jcrist (approved):
Overall this LGTM! Nice how short this was to get working.

ibis/streamlit/__init__.py
cpcloud: It's the `type` argument here:

```python
con = st.experimental_connection("ch", type=IbisConnection)
```

We don't want to pass this to `ibis.connect` since `ibis.connect` doesn't know what to do with `type`.

pyproject.toml
```diff
@@ -86,6 +86,7 @@ snowflake-connector-python = { version = ">=2.7.10,<4", optional = true }
 snowflake-sqlalchemy = { version = ">=1.4.1,<2", optional = true }
 sqlalchemy = { version = ">=1.4,<3", optional = true }
 sqlalchemy-views = { version = ">=0.3.1,<1", optional = true }
+streamlit = { version = ">=1.22,<2", optional = true, python = ">=3.8,!=3.9.7,<4.0" }
```
cpcloud, line 89: Maybe we should put this in dev deps only?

I'm not sure what benefit there is to including it as a proper optional dependency since in the case that someone is using our `ibis.streamlit` connection implementation `streamlit` must be installed.

ibis/streamlit/__init__.py
cpcloud: Done!

pyproject.toml
cpcloud: Split it out to an dev/test-only dependency.

cpcloud:
Ok, I ended up removing the soft dependency on streamlit since it's not required for anything except testing.

I don't currently have a plan to test this other than to have an example streamlit app that we throw in the docs.

ibis/streamlit/__init__.py
jcrist: Should we file this as an upstream bug? Since AFAICT `experimental_connection` creates a connection by calling `type._connect(**kwargs)`, I'd expect any implementation to want to ignore the `type` kwarg.

pyproject.toml
jcrist: Makes sense to me. Nice to keep streamlit and deps out of our already large dev environment too.

ibis/streamlit/__init__.py
cpcloud: It's possible that was fixed since the time I wrote this a month or two ago. Let me confirm and then I'll file a bug if it's still an issue.

PR #6203, opened 2023-05-12
fix(core): interval resolution should upcast to smallest unit

### LINKED ISSUE #6139: bug: interval addition/subtraction should upcast to smallest unit
### What happened?

Currently adding/subtracting intervals results in an interval with a unit of the `left` operand. This can result in accidental information loss if the smaller unit in the operation is the right operand:

```python
In [1]: import ibis

In [2]: one_second = ibis.interval(seconds=1)

In [3]: two_minutes = ibis.interval(minutes=2)

In [4]: (one_second + two_minutes).type().unit
Out[4]: <IntervalUnit.SECOND: 's'>

In [5]: (two_minutes + one_second).type().unit  # I'd expect seconds here too
Out[5]: <IntervalUnit.MINUTE: 'm'>

In [6]: (one_second + two_minutes).execute()
Out[6]: Timedelta('0 days 00:02:01')

In [7]: (two_minutes + one_second).execute()  # the second is dropped on rounding to minutes
Out[7]: Timedelta('0 days 00:02:00')
```

The offending code is [here](https://github.com/ibis-project/ibis/blob/5a013ffd3032e596a16a6a71476c2754f9bc9c8b/ibis/expr/operations/temporal.py#L297). With some pointers I can probably fix this. I know we have casting utilities for handling integer upcasting, but I'm not sure if the same exists for intervals.

dev

_No response_

### Relevant log output

_No response_

### Code of Conduct

mesejo:
Inspired by pandas [resolution_string](https://github.com/pandas-dev/pandas/blob/e19ceb76a4c5903469fea93040c174633c84b9b9/pandas/_libs/tslibs/timedeltas.pyx#L1383)

fixes #6139

### DIFF
--- a/ibis/expr/rules.py
+++ b/ibis/expr/rules.py
@@ -12,6 +12,7 @@ import ibis.expr.schema as sch
 import ibis.expr.types as ir
 from ibis import util
 from ibis.common.annotations import attribute, optional
+from ibis.common.enums import IntervalUnit
 from ibis.common.validators import (
     bool_,
     callable_with,  # noqa: F401
@@ -429,6 +430,14 @@ def numeric_like(name, op):
         return result
 
     return output_dtype
+
+
+def _promote_interval_resolution(units: list[IntervalUnit]) -> IntervalUnit:
+    # Find the smallest unit present in units
+    for unit in reversed(IntervalUnit):
+        if unit in units:
+            return unit
+    raise AssertionError('unreachable')
 
 
 # TODO(kszucs): it could be as simple as rlz.instance_of(ops.TableNode)
--- a/ibis/backends/tests/test_temporal.py
+++ b/ibis/backends/tests/test_temporal.py
@@ -831,6 +831,38 @@ timestamp_value = pd.Timestamp('2018-01-01 18:18:18')
             ],
         ),
         param(
+            lambda t, _: t.timestamp_col
+            + (ibis.interval(days=4) + ibis.interval(hours=2)),
+            lambda t, _: t.timestamp_col
+            + (pd.Timedelta(days=4) + pd.Timedelta(hours=2)),
+            id='timestamp-add-interval-binop-different-units',
+            marks=[
+                pytest.mark.notimpl(
+                    [
+                        "clickhouse",
+                        "sqlite",
+                        "postgres",
+                        "polars",
+                        "mysql",
+                        "impala",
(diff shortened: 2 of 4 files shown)


ibis/expr/rules.py
```diff
@@ -432,6 +433,31 @@ def output_dtype(self):
...
+    elif IntervalUnit.DAY in units:
+        return IntervalUnit.DAY
+    elif IntervalUnit.WEEK in units:
+        return IntervalUnit.WEEK
+    elif IntervalUnit.MONTH in units:
+        return IntervalUnit.MONTH
+    elif IntervalUnit.QUARTER in units:
+        return IntervalUnit.QUARTER
+    else:
+        return IntervalUnit.YEAR
```
jcrist, line 458:
```suggestion
    # Find the smallest unit present in units
    for unit in reversed(IntervalUnit):
        if unit in units:
            return unit
    assert False, "unreachable"
```

jcrist, line 458: The following might be a bit cleaner:

```suggestion
    # Find the smallest unit present in units
    for unit in reversed(IntervalUnit):
        if unit in units:
            return unit
    assert False, "unreachable"
```

jcrist (commented):
Overall this looks good to me! Just one small fixup.

kszucs:
Could you please add a `backends` test case to `ibis/backends/tests/test_temporal.py`, since some of the backends may not support operations with operands using different units?

ibis/expr/rules.py
mesejo: Thanks for the suggestion!

mesejo:
> Could you please add a backends test case to ibis/backends/tests/test_temporal.py, since some of the backends may not support operations with operands using different units?

@kszucs Thanks for the feedback. I added a test case [here](https://github.com/ibis-project/ibis/pull/6203/files#diff-561fbc13e95852435279b84b20e04d88a399427943267a0cd50d59d00e839636R833) Do you think it is enough?

jcrist:
Yep, looks good to me! Can you squash your commits together, and ensure the commit message follows our [commit message requirements](https://ibis-project.org/community/contribute/02_workflow/#submit-a-pr) (this is what caused the failing test)? After that this should be good to merge.

kszucs (approved):
LGTM, thanks @mesejo!

PR #6205, opened 2023-05-12
refactor(datatypes): remove `Set()` in favor of `Array()` datatype

kszucs:
BREAKING CHANGE: `dt.Set` is now an alias for `dt.Array`


jcrist (approved):
Hooray for deleting code

PR #6211, opened 2023-05-13
feat(api): add array zip method and implementations

cpcloud:
This PR implements a `.zip` method for array expressions and adds implementations for:

* clickhouse
* pyspark
* trino


kszucs (approved):
Just a naming request, other than that it looks good to me!

kszucs (approved):

PR #6218, opened 2023-05-15
refactor(polars): align datatype conversion functions with the new convention

kszucs:
The new convention for datatype conversions is to have two simple functions:
- `dtype_from_<target>(typ, nullable=True)`
- `dtype_to_<target>(dtype)`


jcrist (approved):
LGTM. CI failure is unrelated, and should be fixed by #6220.

PR #6219, opened 2023-05-15
refactor(pyspark): align datatype conversion functions with the new convention

kszucs:
The new convention for datatype conversions is to have to simple functions:
- `dtype_from_<target>(typ, nullable=True)`
- `dtype_to_<target>(dtype)`

depends on #6184


jcrist:
Conceptually this LGTM, but looks like there's a few test failures.

kszucs:
We should merge the other related PRs first then I'll rebase this one.

PR #6228, opened 2023-05-16
refactor(api): remove automatic count name from relations

### LINKED ISSUE #6221: bug: Remove name of "count" aggregation when where is specified
### What happened?

A `count` aggregation is always given an alias `.name("count")`. This is the case even when the `where` parameter is specified. If the `where` parameter is specified it no longer makes sense to name the resulting column `"count"` and it causes issues with duplicate aliases (see example below). 

The following code snippet generates an invalid query because both:
* `t.count(where=~t["aBool"])` and
* `t.count(where=t["aBool"])`
are given an alias of `"count"` from [here](https://github.com/ibis-project/ibis/blob/9079bc4fec31721d5cca7f1091aee26624d1e692/ibis/expr/types/relations.py#L2071).

```python
import ibis

t = ibis.table(
    dict(aString="string", aBool="bool"), name="hello"
)

expr = t.group_by(t["aString"]).agg([
    t.count(where=~t["aBool"]).zeroifnull().name("A"),
    t.count(where=t["aBool"]).zeroifnull().name("B"),
])

ibis.show_sql(expr, dialect="clickhouse")
```
prints
```sql
SELECT
  t0.aString,
  COALESCE(COUNTIF(NOT t0.aBool) AS count, 0) AS A,
  COALESCE(COUNTIF(t0.aBool) AS count, 0) AS B
FROM hello AS t0
GROUP BY
  1
```

In [here](https://github.com/ibis-project/ibis/blob/9079bc4fec31721d5cca7f1091aee26624d1e692/ibis/expr/types/relations.py#L2071), we could change the code from:

```python
return ops.CountStar(self, where).to_expr().name("count")
```
to
```python
expr = ops.CountStar(self, where).to_expr()
if not where:
    expr = expr.name("count")
return expr
```

(Another alternative, and one that avoids a hardcoded string entirely, would be to delete `.name("count")` but this may break things elsewhere).

5.0.1

ClickHouse

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #6221.


jcrist (approved):
:shipit:

gforsyth (approved):
:shipit:

kszucs (approved):

kszucs:
Apparently we need to fix a couple test cases.

cpcloud:
Fixing these now.

cpcloud:
This is slightly more complicated because of impala. Will fixup today.

PR #6230, opened 2023-05-17
build: add a tool to decouple both external in internal module dependencies

kszucs:


decouple.py
```diff
@@ -0,0 +1,61 @@
...
+
+CURRENT_DIR = pathlib.Path(__file__).parent.absolute()
+
+
+def generate_dependency_graph(*args):
+    argstring = ' '.join(args)
+    command = f'pydeps --nodot --show-deps {argstring}'
+    print(f"Running: {command}")  # noqa: T201
+    result = subprocess.run(
+        command, shell=True, check=True, capture_output=True, text=True
```
cpcloud, line 20: Is `shell=True` necessary? It should only be needed if you want to expand globs or propagate environment variables, otherwise it's best avoided.

See [Python security considerations for `shell=True`](https://docs.python.org/3/library/subprocess.html#security-considerations) and [this related SO answer](https://stackoverflow.com/a/3172488) for the rationale.

decouple.py
```diff
@@ -0,0 +1,61 @@
...
+import sys
+
+import yaml
+
+CURRENT_DIR = pathlib.Path(__file__).parent.absolute()
+
+
+def generate_dependency_graph(*args):
+    argstring = ' '.join(args)
+    command = f'pydeps --nodot --show-deps {argstring}'
```
cpcloud, line 17:
```suggestion
    command = shlex.join(["pydeps", "--nodot", "--show-deps", *args])
```

cpcloud (changes requested):

decouple.py
kszucs: Not at all, removing.

decouple.py
kszucs: Since we don't use shell=True now new can just pass the args.

justfile
```diff
@@ -121,3 +121,8 @@ view-changelog flags="":
     npx -y -p conventional-changelog-cli \
         -- conventional-changelog --config ./.conventionalcommits.js \
         | ([ "{{ flags }}" = "--pretty" ] && glow -p - || cat -)
+
+# run the decouple script to check for prohibited inter-module dependencies
+decouple *args:
+    #!/usr/bin/env bash
```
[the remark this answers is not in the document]
kszucs: Removed.

decouple.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
```
jcrist, line 1: Could we convert this to a `pytest` in our current test directory (skipping if `pydeps` isn't installed)? I'd prefer that since it's no longer one-more-tool to run to lint the repo (it's just a normal test), and it wouldn't clutter up the top level of the repo.

Alternatively could we move the script to the `ci/` directory, and maybe rename it to something that better indicates it's a linter script (`ci/check_decoupled.py`). The current `decouple.py` name makes it sound like the script itself decouples something.

jcrist (commented):
I conceptually like this, just a few comments.

decouple.py
kszucs: > Could we convert this to a pytest in our current test directory (skipping if pydeps isn't installed)?

`pydeps` is actually quite slow on ibis, so I'd rather keep it as a separate script for now.

decouple.yaml
```diff
@@ -0,0 +1,3 @@
+ibis.expr.*:
```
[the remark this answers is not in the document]
kszucs: Yes, moving it into the script.

decouple.py
cpcloud: Any opposition to changing the name to `check_decoupled.py`/`just check_decoupled`?

cpcloud: Ok, that seems like a fine compromise!

ci/check_disallowed_imports.py
```diff
@@ -0,0 +1,54 @@
...
+
+    for module, module_data in dependency_graph.items():
+        imports = module_data.get("imports", [])
+
+        for pattern, disallow_rules in disallowed_imports.items():
+            if fnmatch.fnmatch(module, pattern):
+                for disallow_rule in disallow_rules:
+                    for imported in imports:
+                        if fnmatch.fnmatch(imported, disallow_rule):
+                            prohibited_deps[module].add(imported)
```
cpcloud, line 31:
```suggestion
				prohibited_deps[module] |= {
					imported
					for disallow_rule in disallow_rules
					for imported in imports
					if fnmatch.fnmatch(imported, disallow_rule)
				}
```

The nesting is getting a bit deep here 😅

cpcloud (approved):
Small style nit, 100% optional!

cpcloud (approved):

PR #6231, opened 2023-05-17
feat(polars): add correlation operation

mesejo:


mesejo:
The tests for the `corr` function are broken. These tests use the same pandas' function call as `expected_fn` to test with both `sample` and `pop`. 
```python 
        param(
            lambda t, where: t.G.corr(t.RBI, where=where, how="pop"),
            lambda t, where: t.G[where].corr(t.RBI[where]),
            id='corr_pop',
            marks=[...],
        ),
        param(
            lambda t, where: t.G.corr(t.RBI, where=where, how="sample"),
            lambda t, where: t.G[where].corr(t.RBI[where]),
            id='corr_samp',
            marks=[...],
        )
```
The [`corr`](https://pandas.pydata.org/docs/reference/api/pandas.Series.corr.html) function in pandas doesn't have an argument for sample or population. 

What do you think should be done here?

cpcloud:
@mesejo Thanks for the PR! I think we can merge this as is, and follow up with fixes to address the lack of sample correlation in pandas. We'll need to figure out a way to compute that in pandas to make the comparison representative of what we're trying to do.

mesejo:
> @mesejo Thanks for the PR! I think we can merge this as is, and follow up with fixes to address the lack of sample correlation in pandas. 

Glad I could help! 

> We'll need to figure out a way to compute that in pandas to make the comparison representative of what we're trying to do.

I think we can drop the parameter `how` from `corr`, for two reasons:

1. All backends provide one or the other, not both, with the exception being polars. (but I have to double-check this)
2. All backends seem to compute the same function but call it differently. Ultimately, all are being tested (approximately) against the same `expected_fn`. It seems that what is `how=sample` in one is `how=pop` in another.

I'll do a bit of research and put my findings into a discussion.

PR #6244, opened 2023-05-19
refactor(ir): simplify `Expr._find_backends()` implementation by using the `ibis.common.graph` utilities

kszucs:


ibis/expr/types/core.py
```diff
@@ -217,23 +217,16 @@ def _find_backends(self) -> tuple[list[BaseBackend], bool]:
...
-        all_backends = []
-        any_unbound = False
-        for backends, has_unbound in g.traverse(finder, self.op()):
-            all_backends += backends
-            any_unbound |= has_unbound
-
-        return list(toolz.unique(all_backends)), any_unbound
+        backends = set()
+        has_unbound = False
+        node_types = (ops.DatabaseTable, ops.SQLQueryResult, ops.UnboundTable)
```
kszucs, line 222: Perhaps it would be even better to define a base node which specifies the `.source` property, like

```python
class PhyisicalOp(ops.Node):
    source: Any
```

Though I'm not insisting to this for now.

cpcloud: Yeah, perhaps `UnboundTable` is no longer a type even, and it's just a `PhysicalTable` (or `PhysicalOp`) with `source = None`.

PR #6245, opened 2023-05-19
refactor(api): enforce at least one argument for `Table` set operations

kszucs:
Previously it was possible to call: `table.difference|union|intersect()` without any argument which doesn't make sense.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -1274,6 +1274,21 @@ def trailing_range_window(preceding, order_by, group_by=None):
     )
 
 
+@functools.wraps(ir.Table.union)
+def union(table: ir.Table, *rest: ir.Table, distinct: bool = False):
+    return table.union(*rest, distinct=distinct) if rest else table
+
+
+@functools.wraps(ir.Table.intersect)
+def intersect(table: ir.Table, *rest: ir.Table, distinct: bool = True):
+    return table.intersect(*rest, distinct=distinct) if rest else table
+
+
+@functools.wraps(ir.Table.difference)
+def difference(table: ir.Table, *rest: ir.Table, distinct: bool = True):
+    return table.difference(*rest, distinct=distinct) if rest else table
+
+
 e = ops.E().to_expr()
 pi = ops.Pi().to_expr()
 
@@ -1341,10 +1356,6 @@ aggregate = ir.Table.aggregate
 cross_join = ir.Table.cross_join
 join = ir.Table.join
 asof_join = ir.Table.asof_join
-
-union = ir.Table.union
-intersect = ir.Table.intersect
-difference = ir.Table.difference
 
 _ = deferred = Deferred()
 """Deferred expression object.
--- a/ibis/backends/tests/test_set_ops.py
+++ b/ibis/backends/tests/test_set_ops.py
@@ -141,10 +141,16 @@ def test_difference(backend, alltypes, df, distinct):
 
 
 @pytest.mark.parametrize("method", ["intersect", "difference", "union"])
-@pytest.mark.parametrize("source", [ibis, ir.Table], ids=["top_level", "method"])
-def test_empty_set_op(alltypes, method, source):
-    result = getattr(source, method)(alltypes)
+def test_table_set_operations_api(alltypes, method):
+    # top level variadic
+    result = getattr(ibis, method)(alltypes)
     assert result.equals(alltypes)
+
+    # table level methods require at least one argument
+    with pytest.raises(
+        TypeError, match="missing 1 required positional argument: 'table'"
+    ):
+        getattr(ir.Table, method)(alltypes)
 
 
 @pytest.mark.parametrize(
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -724,15 +724,17 @@ class Table(Expr, _FixedTextJupyterMixin):
         """
         return ops.SelfReference(self).to_expr()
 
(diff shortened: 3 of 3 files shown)


kszucs:
@cpcloud updated the commit message and added a test.

cpcloud:
True, but set intersection is where things get weird.

Just to nail down some terminology: we're considering nothing the same thing as the empty set for the purposes of this discussion (this equivalence may ultimately be incorrect though).

Other thoughts:

1. If we're considering nothing equal to the empty set then intersection with "nothing-the-same-as-the-empty-set" should return the empty set, not `self` as it is now.
2. This makes the one argument form of intersection somewhat useless without making it incorrect by returning `self`.
3. Ideally we shouldn't special case any of these operations for the other, so if we're allow the one argument form for union and difference then the one argument form for intersection needs to match the set-theoretic definition of intersection with the empty set, which is the empty set. We don't currently have a convenient way to spell "An empty table with a given schema" (`ops.DummyTable` would fit here, but it's not implemented for most backends IIRC).

It seems better to limit functionality and preserve correctness then to keep the convenience and also the incorrectness.

jcrist:
I added this functionality originally, and would be sad to see it go. The motivation was mainly for unions, where I had a query that would have one or more tables that would need to be unioned together. Having `ibis.union(*tables)` work for any number of tables (including only 1 table) was much cleaner than having to special case my code to handle the single table case alone. `intersect` and `difference` were added for completeness. I think having these variadic methods still work in the single table case (at least for `union`) is genuinely useful functionality, and would be reluctant to remove it. I made use of this functionality earlier this week even.

That said, I can see why `Table.union()`/`Table.difference()`/`Table.intersect()` (with no args) look weird. If we moved the top level functions to not just be unbound versions of the methods, then we could continue to support this functionality for `ibis.union` etc. while still being able to require a second argument for the method versions.

```
# This works
ibis.union(t)
# This fails, requires a second table argument to the method version
t.union()
```

I disagree that we need to allow a single table version for `intersect` (but I also think the current behavior is what I'd expect as a user, even if that seems inconsistent to you). If you think we should drop it for `intersect` for "correctness" reasons, I don't see why we'd need to drop it from `union`.

cpcloud:
For intersection I hope we can agree that the issue isn't based on my personal preference, but that we should be consistent with the math behind the operations even if that violates someone's expectation of what the output should be.

cpcloud:
I am okay with special casing intersection for the two argument version.

gforsyth:
> For intersection I hope we can agree that the correctness issue isn't based on my personal preference, and that we should be consistent with the math behind the operations even if that violates a user's expectation of what the output should be.

Python itself distinguishes between `None` and the empty set for `intersection` -- that to me seems like a question of "what would we like to show users?"

```python
[ins] In [1]: {1, 2, 3}.intersection()
Out[1]: {1, 2, 3}

[ins] In [2]: {1, 2, 3}.intersection(set())
Out[2]: set()
```

If we ARE treating `None` as equivalent to the empty set, then returning `self` is incorrect.

cpcloud:
Ok, I think I have changed my mind :)

I think I was stuck on the empty set being the identity, but the particular value of the identity is different for different operations. It wasn't until I realized that is of course true for `+` (0) and `*` (1).

For union the identity is the empty set, but for intersection it's the universal set so I think the current behavior is both correct and it sounds like desired.

cpcloud:
@kszucs I think that a single input to these operations makes sense given the above discussion. Is there any reason to change the existing behavior?

kszucs:
I don't think the problem has anything to do with the identity element or with set theory in practice. It's **purely a user experience issue** where I'd expect the methods to require at least one argument.

Regarding the python analogy, we don't necessarily align everywhere with the python semantics, like we raise errors when we try to compare values with not implicitly castable types, python would just return `False`.  

Even for the python method I'd explicitly expect to pass at least one argument, or an explicit identity element at least.
```py
 {1, 2, 3}.intersection()
```

My preference is to enforce at least one argument for the methods, and I'm not opposed to the top level functions. Corresponding to @jcrist's suggestion:
> * It's fine to require an argument to the method versions. I agree that `t.union()` looks weird and could be forbidden.
> * It'd be nice to have keep variadic top-level function versions of these apis that accept only a single table, but if you're against having `ibis.intersect(t)` I don't see why we also have to remove `ibis.union(t)`. A bit of nonuniformity here seems fine for the convenience - I make use of `ibis.union` in several places where not supporting a single table case would be annoying to work around.

cpcloud:
SGTM, let's keep the variadic top-level functions, and require at least one additional argument (beyond `self`) for the methods.

kszucs:
@cpcloud added the variadic top-level functions.

ibis/expr/api.py
```diff
@@ -1274,6 +1274,21 @@ def trailing_range_window(preceding, order_by, group_by=None):
     )
 
 
+@functools.wraps(ir.Table.union)
+def union(table: ir.Table, *rest: ir.Table, distinct: bool = True):
```
cpcloud, line 1278: `distinct` should default to `False` to match the method.

I believe the defaults differ because `union` with `distinct=True` is not as common as `distinct=False` and also a performance footgun.

cpcloud (changes requested):

ibis/expr/api.py
kszucs: Right, updating.

cpcloud:
@kszucs I will fix the up failing polars tests in a separate PR, incoming shortly.

cpcloud (approved):
LGTM!

PR #6269, opened 2023-05-22
fix(bigquery): fix table naming from `count` rename removal refactor

cpcloud:
Fixes upstream failing tests on the bigquery backend.


jcrist (approved):
:+1:

PR #6287, opened 2023-05-23
test(graphviz): make graphviz CLI optional for tests

jcrist:
Previously these tests would fail if `graphviz` or the corresponding CLI weren't installed.


gforsyth (approved):
:shipit:

PR #6294, opened 2023-05-24
refactor(dtypes): turn dt.dtype() into lazily dispatched factory function

kszucs:
- refactor(ir): glue patterns and rules together
- chore: address review comments
- chore: remove `ops.Backend` base class used only for type checking `source`
- chore: move commented out annotable with bound typevars test case to test_grounds
- fix(expr): restore expression comparison behavior
- chore: resolve post-rebase issues
- chore: fix post-rebase issues
- chore: allow creating parametric types with default parameters from type annotations
- refactor(dtypes): turn dt.dtype() into lazily dispatched factory function


ibis/expr/datatypes/core.py
```diff
@@ -9,68 +9,49 @@
...
-    elif issubclass(origin_type, Sequence):
-        (value_type,) = map(dtype, get_args(value))
-        return Array(value_type)
-    elif issubclass(origin_type, Mapping):
-        key_type, value_type = map(dtype, get_args(value))
-        return Map(key_type, value_type)
+
+@lazy_singledispatch
+def dtype(value, nullable=True) -> DataType:
+    """Construct an ibis datatype from a python value."""
```
cpcloud, line 26:
```suggestion
	"""Construct an ibis datatype from a python type."""
```
since these "values" are actually types or type descriptors.

ibis/expr/datatypes/core.py
```diff
@@ -9,68 +9,49 @@
 from numbers import Integral, Real
 from typing import Any, Iterable, Literal, NamedTuple, Optional
 
-import numpy as np
-import pandas as pd
 import toolz
-from multipledispatch import Dispatcher
 from public import public
-from typing_extensions import get_args, get_origin, get_type_hints
+from typing_extensions import Self, get_args, get_origin, get_type_hints
```
cpcloud, line 14: Isn't the `Self` type only available since 3.11? Or is that not an issue because you're pulling it from `typing_extensions`?

ibis/expr/datatypes/core.py
```diff
@@ -9,68 +9,49 @@
...
+    return DataType.from_pandas(value, nullable)
+
+
+@dtype.register("pyarrow.lib.DataType")
+def from_pyarrow(value, nullable=True):
+    return DataType.from_pyarrow(value, nullable)
+
+
+# lock the dispatcher to prevent new types from being registered
+del dtype.register
```
cpcloud, line 54: This seems like it might be overly restrictive, but I also think it's probably for the best for now until there's a request for custom external data types.

That said, we'll need to extend this shortly after this PR lands to support the various pandas nullable dtypes.

ibis/expr/schema.py
```diff
@@ -153,6 +153,34 @@ def from_tuples(
...
+    def from_numpy(self, numpy_schema):
+        """Return the equivalent ibis schema."""
+        from ibis.formats.numpy import schema_from_numpy
+
+        return schema_from_numpy(numpy_schema)
+
+    @classmethod
+    def from_pandas(self, pandas_schema):
+        """Return the equivalent ibis schema."""
+        from ibis.formats.pandas import schema_from_pandas
```
cpcloud, line 166: Does this need some tests?

ibis/expr/schema.py
```diff
@@ -153,6 +153,34 @@ def from_tuples(
...
+        from ibis.formats.pyarrow import schema_from_pyarrow
+
+        return schema_from_pyarrow(pyarrow_schema)
+
+    @classmethod
+    def from_dask(self, dask_schema):
+        """Return the equivalent ibis schema."""
+        from ibis.formats.dask import schema_from_dask
+
+        return schema_from_dask(dask_schema)
```
cpcloud, line 182: Looks like maybe this module is missing some test coverage.

cpcloud (changes requested):

ibis/expr/schema.py
kszucs: I'm removing the whole `schema_from_dask` in a follow-up :)

ibis/expr/schema.py
kszucs: `schema_from_pandas` is covered just the method isn't

ibis/expr/datatypes/core.py
kszucs: Correct.

ibis/expr/schema.py
kszucs: Covered it with a test.

ibis/expr/schema.py
kszucs: Covered the method in a test.

ibis/expr/datatypes/core.py
kszucs: Updated.

cpcloud (approved):

PR #6297, opened 2023-05-24
docs: update copyright year and add a period (minor docs nits)

lostmygithubaccount:
- update copyright notice to 2023
- add period for consistency

This PR makes very minor edits on the homepage to update the copyright year and add a period at the end of a sentence in the bullet points for consistency

I'm not sure if there's a way to automatically use the current year in the `mkdocs.yml`?

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -164,4 +164,4 @@ extra:
       icon: :material-cancel:{ .cancel }
       description: Unlikely to ever be supported or no upstream support.
 
-copyright: "Copyright &copy; 2014-2022"
+copyright: "Copyright &copy; Ibis Maintainers"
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -164,4 +164,4 @@ extra:
       icon: :material-cancel:{ .cancel }
       description: Unlikely to ever be supported or no upstream support.
 
-copyright: "Copyright &copy; 2014-2022"
+copyright: "Copyright &copy; 2014-2023"
```
jcrist, line 167: Copyright years don't need to be updated every year, they're valid from the year the content was first created. https://hynek.me/til/copyright-years/. I think we should just remove the years entirely.

```suggestion
copyright: "Copyright &copy; Ibis Maintainers"
```

jcrist:
Thanks @lostmygithubaccount, just one last fix.

We generate the changelog from commit messages, so we're pretty picky about the commit message structure. See https://ibis-project.org/community/contribute/02_workflow/#submit-a-pr.

Can you squash all your commits together, and update the commit message title to match the required format? Should be something like:

```
docs: update copyright and some minor edits
```

cpcloud:
There's a way to use custom HTML (with Jinja if necessary) for the footer, but that seems a bit overkill for now.

PR #6301, opened 2023-05-24
fix(ir): raise Exception for group by with no keys

### LINKED ISSUE #6237: bug: .distinct(on="does_not_exist") does not crash
### What happened?

```python
ibis.examples.topo.fetch().distinct(on="does_not_exist")
```

I would expect this to error.

5.1.0, have not tested on main

duckdb, but probably irrelevant?

### Relevant log output

_No response_

### Code of Conduct

mesejo:
fixes #6237

### DIFF
--- a/ibis/expr/types/groupby.py
+++ b/ibis/expr/types/groupby.py
@@ -28,6 +28,7 @@ from ibis import util
 from ibis.expr.deferred import Deferred
 from ibis.selectors import Selector
 from ibis.expr.types.relations import bind_expr
+import ibis.common.exceptions as com
 
 _function_types = tuple(
     filter(
@@ -79,6 +80,10 @@ class GroupedTable:
                 ),
             )
         )
+
+        if not self.by:
+            raise com.IbisInputError("The grouping keys list is empty")
+
         self._order_by = order_by or []
         self._having = having or []
         self._window = window
--- a/ibis/tests/expr/test_table.py
+++ b/ibis/tests/expr/test_table.py
@@ -1447,6 +1447,13 @@ def test_group_by_key_function():
     assert expr.columns == ['new_key', 'foo']
 
 
+def test_group_by_no_keys():
+    t = ibis.table([('a', 'timestamp'), ('b', 'string'), ('c', 'double')])
+
+    with pytest.raises(com.IbisInputError):
+        t.group_by(s.startswith("x")).aggregate(foo=t.c.mean())
+
+
 def test_unbound_table_name():
     t = ibis.table([('a', 'timestamp')])
     name = t.op().name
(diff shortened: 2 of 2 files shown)


cpcloud:
Thanks for the PR @mesejo!

A `GROUP BY` with no keys is a valid expression in I think all of our SQL backends so I think we want to preserve that.

Would you mind addressing #6237 in a different way? How about fixing the issue directly in the `distinct` method? This would also allow us to craft a more specific error message that may be a bit more informative for end-users.

cpcloud:
@mesejo Actually, I think I am mistaken!

I was thinking about something else and it's actually that `GROUP BY` is valid without any aggregates!

I think what you've done here is correct, so I'll review it as is.

ibis/expr/types/groupby.py
```diff
@@ -79,6 +79,10 @@ def __init__(
                 ),
             )
         )
+
+        if not self.by:
+            raise ValueError("The grouping keys list is empty!")
```
cpcloud, line 84: `ValueError` seems like it lacks useful type information. What about raising an `com.IbisInputError`, which while not super informative itself, it conveys that there was an error caused by some input from the user.

cpcloud, line 84:
```suggestion
            raise ValueError("The grouping keys list is empty")
```

No need to shout :)

ibis/tests/expr/test_table.py
```diff
@@ -1812,3 +1819,9 @@ def test_invalid_keep_distinct():
     t = ibis.table(dict(a="int", b="string"), name="t")
     with pytest.raises(com.IbisError, match="Invalid value for `keep`:"):
         t.distinct(on="a", keep="invalid")
+
+
+def test_invalid_distinct_empty_key():
+    t = ibis.table(dict(a="int", b="string"), name="t")
+    with pytest.raises(ValueError):
+        t.distinct(on="c", keep="first")
```
cpcloud, line 1827: I think we should still raise a more specific error message in `distinct(on="does_not_exist")`. Erroring in the group by that is the implementation of `distinct(on...)` leaks implementation details and may leave users wondering why a group by is involved.

cpcloud: This can be done in a follow-up for sure!

cpcloud (changes requested):

ibis/expr/types/groupby.py
mesejo: Agree, `com.IbisInputError` seems like a better option.

ibis/tests/expr/test_table.py
mesejo: I'll try to do it in a follow-up since I've to figure out how to do it.

cpcloud:
The dask failure appears to be a flaky test related to ordering and the impala failure looks related to github actions.

cpcloud:
Thanks!

PR #6310, opened 2023-05-25
fix: add issue write permission to assign.yml

lostmygithubaccount:
workflow is failing for `/take` in comments, e.g.: https://github.com/ibis-project/ibis/actions/runs/5084258325

this simply adds the write issue permission, hopefully fixing

not 100% sure this is the problem


cpcloud (approved):
LGTM! Let's try it out!

Issue #5923, opened 2023-04-05
docs: improve backend-specific docs


jcrist:
Currently we have a docs page for [each of our backends](https://ibis-project.org/backends/). These pages aren't terribly useful as written currently.

The different backend pages are mostly identical, filling out a template with a few specific details. They're also mostly API documentation, where many of the documented methods are common across all backends.

I'd expect that the typical Ibis user will want to know the following:

1. What backends are available (we have a general high-level summary page for that)
2. What methods/attributes are available on _all_ backends. What can they rely on to always exist?
3. How do they install their chosen backend?
4. How do they connect to their chosen backend. We should document both `ibis.<backend>.connect` and the URL format for `ibis.connect` for these cases.
5. Any backend-specific behavior for their chosen backend.
    - Any methods that exist only on that backend
    - Any limitations for this backend (e.g. "doesn't support `create_table`")

IMO 1 and 2 would be best handled by a general backend page, with the remaining bits relegated to backend specific pages. mkdocstrings has an option to document only certain methods in api docs, we might use that to split out common backend methods from backend specific things? If anyone has a better idea of how to arrange this info though, please speak up!

cpcloud:
From the original list of things:

> What backends are available (we have a general high-level summary page for that)

This can be seen in multiple places, but for my money the easiest path is clicking on the `Backend` nav bar item:

![image](https://github.com/ibis-project/ibis/assets/417981/22566c63-2556-4694-885d-addaae1c8137)

> What methods/attributes are available on all backends. What can they rely on to always exist?

We don't have anything for this yet

> How do they install their chosen backend?

Listed in multiple places including on each backend page

> How do they connect to their chosen backend. We should document both ibis.<backend>.connect and the URL format for ibis.connect for these cases.

Listed on each backend page

> Any backend-specific behavior for their chosen backend.
>   * Any methods that exist only on that backend

We don't differentiate here but perhaps this is achievable by looking at all public methods that are not overrides of a parent class method. I am going to take a crack at this.

>   * Any limitations for this backend (e.g. "doesn't support create_table")

We're slowly moving towards a model where these methods simply don't exist if the backend doesn't support them.

PR #6318, opened 2023-05-26
Add connection information to bigquery and snowflake backend docs

gforsyth:
xref #5923 

A first cut of improving some of the backend docs -- I started with Snowflake and Bigquery as those have been (to me) the hardest to figure out in re: connecting.

There's a lot more to be done for all the backends, but I think this is an improvement.


gforsyth:
Oh, and it's _highly_ unclear to me which of the various account identifiers would be considered sensitive information, so all of those have been changed in the screenshots.

cpcloud (approved):
Wonderful!

The images are just great!

PR #6324, opened 2023-05-27
test(clickhouse): avoid running tests that depend on `ibis_testing` existing

cpcloud:
This PR fixes an issue with a ClickHouse backend test that assumes the `ibis_testing` database exists when in fact it might not. Test failure observed in the wild here: https://github.com/ibis-project/ibis/actions/runs/5093340617/jobs/9155829965?pr=6298. There may be similar issues with other backends. We can fix those as they arise.


cpcloud:
I was able to remove some of the other backends' dependence on `ibis_testing` in `ibis/backends/tests/test_client.py`.

PR #6340, opened 2023-05-30
docs(file support): add badge and docstrings for read_* methods

gforsyth:
xref #5923 

Adding non-templated backend doc pages for Datafusion, DuckDB, Polars, and PySpark.

What do y'all think of having a few badges at the top of each backend page?  I added the ones here for the backends that have some notion of local file support.  It could also get crowded very quickly.

### DIFF
--- a/docs/backends/DuckDB.md
+++ b/docs/backends/DuckDB.md
@@ -63,6 +63,21 @@ con = ibis.duckdb.connect("mydb.duckdb")  # (1)
       heading_level: 4
 <!-- prettier-ignore-end -->
 
+### `ibis.connect` URL format
+
+In addition to `ibis.duckdb.connect`, you can also connect to DuckDB by
+passing a properly formatted DuckDB connection URL to `ibis.connect`
+
+```python
+con = ibis.connect("duckdb:///path/to/local/file")
+```
+
+```python
+con = ibis.connect("duckdb://") # (1)
+```
+
+1. ephemeral, in-memory database
+
 ## File Support
 
 <!-- prettier-ignore-start -->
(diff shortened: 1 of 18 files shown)


docs/backends/DuckDB.md
```diff
@@ -2,17 +2,85 @@
...
-backend_param_style: a path to a DuckDB database
-backend_connection_example: ibis.duckdb.connect("path/to/my.duckdb")
-version_added: "3.0"
-intro: |
-  !!! danger "`duckdb` >= 0.5.0 requires `duckdb-engine` >= 0.6.2"
+---
+
+![filebadge](https://img.shields.io/badge/Supports-CSV%20|%20Parquet%20|%20JSON%20|%20Arrow%20|%20Pandas%20|%20SQLite%20|%20Postgres-blue?style=flat-square)
+
+??? danger "`duckdb` >= 0.5.0 requires `duckdb-engine` >= 0.6.2"
```
gforsyth, line 9: This is a collapsing admonition, which I think is nice for a longer message.

cpcloud:
I like the badges, they are a quick first approximation to "what can I do with this backend?".

gforsyth:
Representative view.

Need to finish adding these to the rest of the backends.

gforsyth:
Ok, I think I've got this PR where I want it.  

I'm going to mess around a little bit with the sizing of method names (or rather the header level of those method names) as I find the current sizing a little hard to parse visually.  I'll put up another PR with some examples.

Oh, and I enabled a "copy button" for all the code blocks.

The CSS sheet enables things like a code block like

```
>>> print('hi')
```

where if you select it, the `>>>` are unselectable.  This behavior, sadly, doesn't extend to the copy button.

PR #6361, opened 2023-06-02
fix(docs): add documentation for across

fmichonneau:
I noticed that across wasn't documented. I made an attempt at adding it. Please suggest improvements!


cpcloud (approved):
Fixed up the failures and added some type annotations!

PR #6363, opened 2023-06-02
feat: add easy type hints to GroupBy

NickCrews:
I just wrote this in the GitHub web UI so no idea if this will pass mypy

### DIFF
--- a/ibis/expr/types/groupby.py
+++ b/ibis/expr/types/groupby.py
@@ -105,7 +105,7 @@ class GroupedTable:
         else:
             return GroupedArray(col, self)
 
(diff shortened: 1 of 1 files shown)


NickCrews:
pre-commit.ci autofix

NickCrews:
Wow, that's the first time I've use the pre-commit bot, that is awesome!

NickCrews:
Oh, does Ibis not actually do any type checking? I don't see any mypy config anywhere...

cpcloud (approved):
We don't run any mypy checks, but this is still helpful for autocompletion!

PR #6372, opened 2023-06-05
ci: fix flaky snowflake ci jobs

### LINKED ISSUE #6371: ci: snowflake tests are a little flaky
### What happened?

https://github.com/ibis-project/ibis/actions/runs/5176144293/jobs/9324589816 suggests one or more tables are modified outside of the test run between setup and when the failing tests execute. This is entirely possible given that we run two concurrent snowflake jobs against the same schema.

We should address this flakiness.

One possible way to address this is to avoid sharing a database schema between the without-pyarrow snowflake job and the with-pyarrow snowflake job. This would require a bit of plumbing of some new environment variables (`SNOWFLAKE_SCHEMA` perhaps).

`master`

Snowflake

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #6371.

### DIFF
--- a/.github/workflows/ibis-backends-cloud.yml
+++ b/.github/workflows/ibis-backends-cloud.yml
@@ -42,7 +42,7 @@ jobs:
             randomly-seed: false
 
           - name: snowflake
-            key: snowflake-pyarrow
+            key: snowflake_pyarrow
             title: Snowflake + PyArrow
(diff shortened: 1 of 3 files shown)


cpcloud:
Merging, so that I can work out any kinks.

PR #6373, opened 2023-06-05
perf(imports): speed up imports by at least 2x

cpcloud:
This PR speeds up importing the individual backends by at least 2x and in the
clickhouse case 3x+.

This shows up for users as additional unnecessary overhead when connecting to
databases due entry point loading on module attribute access.

I noticed that the time to call `ibis.duckdb.connect()`, was much much slower
than calling `duckdb.connect()`.

I expected ibis to be a bit slower, but not **27x** slower:

```
In [1]: %time import ibis; con = ibis.duckdb.connect()
CPU times: user 1.94 s, sys: 6.35 s, total: 8.29 s
Wall time: 801 ms

In [1]: %time import ibis; con = ibis.duckdb.connect()
CPU times: user 296 ms, sys: 40.1 ms, total: 336 ms
Wall time: 333 ms
```

It turned out that good chunk of time is spent importing `geoalchemy2` and
`geopandas`, just to check whether they are installed:

```
In [1]: %%time
   ...: import geopandas
   ...: import geoalchemy2
   ...:
   ...:
CPU times: user 611 ms, sys: 367 ms, total: 978 ms
Wall time: 351 ms
```

This overhead is incurred for backends that don't have geospatial support (by definition)

It turns out we are slower than necessary to import across most of our backends, mostly because
of pandas, pyarrow and numpy imports followed by some
less-expensive-to-import-but-still-non-negligible modules like sqlalchemy.

Any of the sqlalchemy backends that reuse the postgres registry were slow
because of the geospatial dependency imports.

What I did here is generally to move imports closer to where they are used/needed.

This involved a few things

1. Remove the geospatial dependency checking from the import process. We'll now
 	 always register geospatial ops and errors from missing dependencies will occur
 	 when trying to convert outputs to GeoDataFrames, or perform other Python-side
 	 geospatial operations or type conversions.
2. Move imports that are more expensive into methods and function calls.

Additionally, `import google.cloud.bigquery` alone appears to take about 700
milliseconds, so I moved that import closer to call sites as needed:

```
In [1]: %time import google.cloud.bigquery
CPU times: user 1.84 s, sys: 6.39 s, total: 8.23 s
Wall time: 705 ms
```

Finally, `clickhouse_connect` is also very slow, so I inlined those imports too:

```
In [1]: %time import clickhouse_connect
CPU times: user 1.6 s, sys: 6.42 s, total: 8.01 s
Wall time: 491 ms
```

I've also added a benchmark so that we can keep track of import times.


jcrist (commented):
+1 on removing the geospatial imports, that's a nice find.

I'm less convinced on the larger changes here - there's a readability and maintenance cost to making certain imports local (especially common ones like `sqlalchemy`). Unless we're testing that a certain file or backend can be used without that import being imported, I could see these imports accidentally creeping back in.

IMO, local imports should be used only for optional dependencies, or modules that are known to be expensive to avoid importing them unless that functionality is likely to be used. So delaying the geospatial imports makes sense to me (expensive, unlikely to be used), but delaying the import of bigquery from the bigquery backend (where it will always be needed) makes less sense.

cpcloud:
> I'm less convinced on the larger changes here - there's a readability and maintenance cost to making certain imports local (especially common ones like sqlalchemy). Unless we're testing that a certain file or backend can be used without that import being imported, I could see these imports accidentally creeping back in.

> IMO, local imports should be used only for optional dependencies, or modules that are known to be expensive to avoid importing them unless that functionality is likely to be used. So delaying the geospatial imports makes sense to me (expensive, unlikely to be used), but delaying the import of bigquery from the bigquery backend (where it will always be needed) makes less sense.

I hear what you're saying.

All of these changes here were driven by iteration using tuna and looking at what was "reasonable" to move around and what would actually decrease import times a noticeable amount. Sometimes, something was only noticeable once the sum of the import times from a bunch of apparently "cheap to import" modules were moved to the call site.

I'd encourage you to try removing some of the commits here and see what changes. It's pretty easy to look at this code and say it's hard to read etc because a lot of moving these imports are not obviously impactful. If you remove enough of the smaller moves then the import will creep back up.

> but delaying the import of bigquery from the bigquery backend (where it will always be needed) makes less sense

I don't think that's true. You're right that it's always required if you're connecting to the BigQuery service and executing queries etc, but if you're using unbound tables and the compiler you may never need some of the imports (e.g., the storage client).

I believe this holds true for all the SQL backends.

jcrist:
>  I don't think that's true. You're right that it's always required if you're connecting to the BigQuery service and executing queries etc, but if you're using unbound tables and the compiler you may never need some of the imports (e.g., the storage client).

I understand not needing the storage client, but unless you're only using `ibis.bigquery.compile`, AFAICT you'll always need `google.bigquery` imported. My issue is that making this import local makes the code a bit harder to read, and without a test ensuring `ibis.bigquery.compile` can work without importing `google.cloud.bigquery`, there's a decent chance there's a regression in this dependency management. I'm not convinced this is worth it. Same goes for `import duckdb` in the `duckdb` backend.

> I'd encourage you to try removing some of the commits here and see what changes.

The other two I pointed out I don't think have a perf effect. `rich` is a very light import, looking at that profile it seems pretty negligible compared to other imports. But making that import local seems fine if you disagree, since it's pretty manageable (though see my comment above about a behavior change).

The `sqlalchemy` one doesn't have an effect (see my comment above). For any `sqlalchemy` backend, `sqlalchemy` is already imported in the base class module. Making those imports local just muddies the code without reducing import times.

gforsyth:
I'm very much of two minds on this.  

Readability counts, and Ibis is very intertwined, so some of the inlining can make it that much harder for new contributors to get up to speed.

AND

I have done horrible things in the name of improving startup times, because startup / import times do matter.

Like all parts of a user-facing library, I think it comes down to a balance of what the user experiences, and also how the maintainers are able to fix, improve, and extend the code.

We have a few categories of things here, and I think they probably each deserve their own PR / discussion, because I don't think this is a one-size-fits-all situation:

1. Geospatial-y imports -- very expensive, used by a subset of users on a subset of backends.  We should almost certainly inline these.
2. Bigquery-ish imports -- very expensive, used by almost all users of one _specific_ backend.
3. SQLAlchemical imports -- moderately expensive, used by the majority of Ibis users (supposition)
4. Pandonic imports -- very expensive, used by all users (for the moment)
5. PyArrow imports -- moderately expensive, used by most users (at least indirectly)


These each occupy a separate semantic import space.  

We can say that most users are not using the geospatial stuff, so we can (and should!) accept the loss of readability for the sake of improving performance for all other users.  

The obverse of this is bigquery imports -- we would inline those so that users who are using bigquery with unbound tables don't pay the import price.  We don't have metrics on this, but I would guess that the more common pattern is to execute those queries via Ibis (and not just using unbound tables and the compiler). In that case, we probably don't want to sacrifice readability for that subset of users.

Then we have sqlalchemy -- I think we should definitely strive to not import sqlalchemy for the backends that don't use it, but a BUNCH of backends use it, and use it everywhere, so inlining it over and over again is painful, and I think will eventually lead to a regression.

Pandas is expensive enough that it might be worth the readability-hit.

I'm unsure of how to handle pyarrow.

One thing we might try for some of the expensive but oft-used or just "very expensive" imports is https://github.com/xonsh/lazyasd -- it also comes with a readability cost, but will let us keep expensive imports at the top of a given file and avoid all the inlined imports.

I do think that we should split up this PR into components, though, to make it easier to discuss each set of changes, and also easier to measure each group of changes independently.  I know that makes it more work @cpcloud and I'm happy to take some of that so we aren't just burning all of your time.

cpcloud:
Closing this out for now. We can selectively cannibalize it later as desired.

PR #6380, opened 2023-06-06
feat(polars): support SQL dialect translation

cpcloud:
Add support for SQL dialect translation and fix upstream test failures.


cpcloud:
Merging to get upstream passing.

PR #6381, opened 2023-06-06
ci(snowflake): check `matrix.backend.key` before `matrix.backend.name`

cpcloud:
`name` is never null, so checking it first leads to testing against the same schema, the prevention of which is the whole point of using name/key to specify the schema.


cpcloud:
No sense in running all the CI for every non-snowflake backend.

PR #6386, opened 2023-06-06
feat(formats): support timestamps in delta output; default to micros for pyarrow conversion

cpcloud:
This PR brings back timestamp scale round tripping, which allows delta timestamp writing, and gives a couple new xpasses from datafusion. Huzzah!

### DIFF
--- a/ibis/formats/tests/test_pyarrow.py
+++ b/ibis/formats/tests/test_pyarrow.py
@@ -34,6 +34,7 @@ roundtripable_types = st.deferred(
         | past.duration_types
         | past.string_type
         | past.binary_type
+        | past.timestamp_types
         | st.builds(pa.list_, roundtripable_types)
         | past.struct_types(roundtripable_types)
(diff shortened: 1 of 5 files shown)


ibis/formats/tests/test_pyarrow.py
```diff
@@ -84,6 +81,10 @@ def test_non_roundtripable_types(arrow_type, ibis_type, restored_type):
     assert_dtype_roundtrip(arrow_type, ibis_type, restored_type)
 
 
+def test_timestamp_with_no_scale():
+    assert dt.Timestamp.from_unit("us").to_pyarrow() == pa.timestamp("us")
```
kszucs, line 85: I guess this should be 'dt.Timestamp(scale=None)'

cpcloud: Yup!

kszucs (approved):

PR #6391, opened 2023-06-07
refactor(util): use base64 to compress uuid table names

kszucs:
The objective is to remove the hard numpy dependency from `ibis.util`.

### DIFF
--- a/ibis/util.py
+++ b/ibis/util.py
@@ -1,6 +1,7 @@
 """Ibis utility functions."""
 from __future__ import annotations
 
+import base64
 import collections
 import functools
 import importlib.metadata
@@ -25,7 +26,6 @@ from typing import (
 )
 from uuid import uuid4
 
-import numpy as np
 import toolz
 
 if TYPE_CHECKING:
@@ -540,4 +540,5 @@ def normalize_filename(source: str | Path) -> str:
 
 def gen_name(namespace: str) -> str:
     """Create a case-insensitive uuid4 unique table name."""
-    return f"_ibis_{namespace}_{np.base_repr(uuid.uuid4().int, 36)}".lower()
+    uid = base64.b32encode(uuid.uuid4().bytes).decode().rstrip("=").lower()
+    return f"_ibis_{namespace}_{uid}"


ibis/util.py
```diff
@@ -540,4 +540,6 @@ def _absolufy_paths(name):
 
 def gen_name(namespace: str) -> str:
     """Create a case-insensitive uuid4 unique table name."""
-    return f"_ibis_{namespace}_{np.base_repr(uuid.uuid4().int, 36)}".lower()
```
cpcloud, line 543: Why not inline the implementation of `np.base_repr`? It's really short and doesn't result in information loss (removing the `-_=` characters).

cpcloud: Base 36 is ideal because it only uses alphanumeric characters, thus preventing any hacks like having to remove characters that many databases don't accept.

cpcloud: As we discussed base 32 also works as the only non alphanumeric character is `=` which is only used for padding, and thus safe to remove for our use of it here.

jcrist: Instead of worrying about uuids and encodings, you could use `random.choices` to choose from the base36 charset. A uuid4 is just a standard str representation of `os.urandom(16)`, nothing more. The following is functionally equivalent to what we had before:

```python
import random
charset = "abcdefghijklmnopqrstuvwxyz0123456789"
return f"_ibis_{namespace}_{''.join(random.choices(charset, k=24))}"
```

cpcloud: Here's an example:

```
>>> import uuid
>>> import random
>>> random.seed(1234)
>>> uuid.uuid4()
UUID('cd7bf8fe-f88b-4776-8237-efff7c33b20c')
>>> random.seed(1234)
>>> uuid.uuid4()
UUID('b995096e-aafe-4612-b808-21d8d41a0d06')
>>> random.seed(234); "".join(random.choices('abcdefhijklmopqrstuvwxyz0123456789',k=24))
'm557ts688ms5kpzppkkaodzv'
>>> random.seed(234); "".join(random.choices('abcdefhijklmopqrstuvwxyz0123456789',k=24))
'm557ts688ms5kpzppkkaodzv'
```

cpcloud: I forgot the letter `n` in the `random.choices` version, but you get the idea :)

cpcloud: The only reason I am acutely aware of this is because we hit this **exact** problem when testing snowflake. https://github.com/snowflakedb/snowflake-connector-python/issues/1478

cpcloud: True. And we can continue to use uuids instead of worrying about that 🙂

PR #6395, opened 2023-06-07
fix(snowflake): fix `tmpdir` construction for python <3.10

cpcloud:
Fixes a bug where we are using Python 3.10-only APIs (the `ignore_cleanup_errors` in the `tempfile.TemporaryDirectory` constructor)


cpcloud:
Found a bug, pushing a fix.

cpcloud:
Merging!

PR #6397, opened 2023-06-07
refactor(formats): encapsulate the various conversion logics to TypeMapper, SchemaMapper and DataMapper subclasses

kszucs:
- feat(formats): add a format specification using an abstract base class
- refactor(formats): encapsulate the various conversion logics to TypeMapper, SchemaMapper and DataMapper subclasses

### DIFF
--- a/ibis/formats/__init__.py
+++ b/ibis/formats/__init__.py
@@ -1,0 +1,182 @@
+from abc import ABC, abstractmethod
+from typing import Generic, TypeVar
+
+from ibis.expr.datatypes import DataType
+from ibis.expr.schema import Schema
+
+C = TypeVar('C')
+T = TypeVar('T')
+S = TypeVar('S')
+
+
+class TypeMapper(ABC, Generic[T]):
+    # `T` is the format-specific type object, e.g. pyarrow.DataType or
+    # sqlalchemy.types.TypeEngine
+
+    @classmethod
+    @abstractmethod
+    def from_ibis(cls, dtype: DataType) -> T:
+        """Convert an Ibis DataType to a format-specific type object.
+
+        Parameters
+        ----------
+        dtype
+            The Ibis DataType to convert.
+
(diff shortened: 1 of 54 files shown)


ibis/formats/__init__.py
```diff
@@ -0,0 +1,175 @@
+from abc import ABC, abstractmethod
+from typing import Generic, TypeVar
+
+from ibis.expr.datatypes import DataType
+from ibis.expr.schema import Schema
```
cpcloud, line 5: Perhaps these should be under a `TYPE_CHECKING` guard to avoid the cost of imports only used for type-checking/annotation.

ibis/formats/__init__.py
```diff
@@ -0,0 +1,175 @@
...
+
+T = TypeVar('T')
```
cpcloud, line 7: I think each of these type variables should have some documentation that gives one or more examples of a concrete type of `T`. For example, (I think) a concrete type for `T` would be `sqlalchemy.sql.type_api.TypeEngine` (the super class of all sqlalchemy types).

ibis/formats/__init__.py
```diff
@@ -0,0 +1,175 @@
...
+S = TypeVar('S')
+VS = TypeVar('VS')
+VC = TypeVar('VC')
+VT = TypeVar('VT')
```
cpcloud, line 11:
```suggestion
ScalarT = TypeVar('VS')
ColumnT = TypeVar('VC')
TableT = TypeVar('VT')
```

cpcloud (commented):
+1 on the design.

Really nice to see you tame the complexity of all the conversions and mappings with a concrete interface with a relatively small API surface.

This should made adding a new backend needing any custom conversions or mappings a lot more straightforward.

cpcloud (approved):
Sweet, let's merge it!

PR #6408, opened 2023-06-09
feat(flink): introduce Apache Flink backend

chloeh13q:
This PR looks to add support for Apache Flink (streaming) in Ibis.

### Motivation
As data grows, developers want to do increasingly sophisticated data science, as well as reduce the end-to-end time to value. There has been an explosion of tools, libraries, and frameworks to democratize the performance and functionality needed to create increasingly complex data systems that address these new data science demands. That said, the combinatorics of tools has forced developers to spend an increasing amount of their time rewriting code or learning new APIs. Building and extending existing standards would ameliorate much of this pain by allowing developers to write code once and use it across a myriad of tools in the data tool chain. Ibis already works with numerous backends for batch analytics, and we want to extend Ibis into real-time systems, starting with Apache Flink.

### Current State
Ibis is currently a batch-oriented library. All of the current supported backends derive from a batch paradigm (aside from Spark, which [does offer support for stream processing, albeit using micro-batches underneath the hood](https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html#overview)).

Unlike batch systems, which operate on bounded data, streaming systems are [designed with unbounded data in mind](https://www.oreilly.com/radar/the-world-beyond-batch-streaming-101/). In order to deal with an infinite data stream, streaming data systems operate with unique concepts such as “event time”, “processing time”, “watermark”, etc.

As streaming systems gain more use cases, there have been efforts to close the gap between batch and streaming. [Flink SQL, for example, was born as a part of such effort and, through allowing users to write streaming engines in a SQL-like manner](https://www.ververica.com/blog/apache-flink-sql-past-present-and-future), have been vastly successful in that regard. The success of Flink SQL both validates the potential of stream and batch unification and inspires the community to push for better standards, a vision that Ibis is at a unique and valuable position to help build.

[Adding support for a streaming engine like Flink, however, is non-trivial](https://gist.github.com/gforsyth/365ff3c9ab44743906613d027d083c1c).

### This PR
We would like to make the work to support Apache Flink & streaming in Ibis incremental. In the first stage of the implementation, we will focus on a "string-generating backend", where a top-level function (similar to `ibis.backends.clickhouse.compiler.core.translate`) can be used to handle just the expr -> SQL compilation. Later on we will use this as part of the backend, when we're ready to add support for it.

This PR introduces the aforementioned function and sets up the testing infrastructure via `pytest-snapshot` without introducing new APIs for the time being.

### DIFF
--- a/ibis/backends/flink/registry.py
+++ b/ibis/backends/flink/registry.py
@@ -1,0 +1,31 @@
+import ibis.expr.operations as ops
+from ibis.backends.base.sql.compiler import ExprTranslator
+from ibis.backends.base.sql.registry import helpers
+from ibis.backends.base.sql.registry import (
+    operation_registry as base_operation_registry,
+)
+from ibis.common.temporal import TimestampUnit
+
+operation_registry = base_operation_registry.copy()
+
+
+def _count_star(translator: ExprTranslator, op: ops.Node) -> str:
+    return "count(*)"
+
+
+def _timestamp_from_unix(translator, op):
+    arg, unit = op.args
+
+    if unit == TimestampUnit.MILLISECOND:
+        return f"TO_TIMESTAMP_LTZ({helpers.quote_identifier(arg.name, force=True)}, 3)"
+    elif unit == TimestampUnit.SECOND:
+        return f"TO_TIMESTAMP_LTZ({helpers.quote_identifier(arg.name, force=True)}, 0)"
+    raise ValueError(f"{unit!r} unit is not supported!")
+
+
+operation_registry.update(
+    {
+        ops.CountStar: _count_star,
+        ops.TimestampFromUNIX: _timestamp_from_unix,
+    }
+)
--- a/ibis/backends/flink/tests/test_translator.py
+++ b/ibis/backends/flink/tests/test_translator.py
@@ -1,0 +1,52 @@
+import pytest
+from pytest import param
+
+import ibis
+from ibis.backends.flink.compiler.core import translate
+
+
+@pytest.fixture
+def schema():
+    return [
+        ('a', 'int8'),
+        ('b', 'int16'),
+        ('c', 'int32'),
+        ('d', 'int64'),
+        ('e', 'float32'),
+        ('f', 'float64'),
+        ('g', 'string'),
+        ('h', 'boolean'),
+        ('i', 'timestamp'),
+        ('j', 'date'),
+        ('k', 'time'),
+    ]
+
+
+@pytest.fixture
+def table(schema):
+    return ibis.table(schema, name='table')
+
+
+def test_translate_sum(snapshot, table):
+    expr = table.a.sum()
+    result = translate(expr.as_table().op())
+    snapshot.assert_match(str(result), "out.sql")
+
+
+def test_translate_count_star(snapshot, table):
+    expr = table.group_by(table.i).size()
+    result = translate(expr.as_table().op())
+    snapshot.assert_match(str(result), "out.sql")
+
+
+@pytest.mark.parametrize(
+    "unit",
+    [
+        param("ms", id="timestamp_ms"),
+        param("s", id="timestamp_s"),
+    ],
+)
+def test_translate_timestamp_from_unix(snapshot, table, unit):
+    expr = table.d.to_timestamp(unit=unit)
+    result = translate(expr.as_table().op())
+    snapshot.assert_match(result, "out.sql")
--- a/poetry.lock
+++ b/poetry.lock
@@ -5360,6 +5360,7 @@ decompiler = ["black"]
 deltalake = ["deltalake"]
 druid = ["pydruid", "sqlalchemy"]
 duckdb = ["duckdb", "duckdb-engine", "packaging", "sqlalchemy", "sqlalchemy-views"]
+flink = []
 geospatial = ["GeoAlchemy2", "geopandas", "shapely"]
 impala = ["fsspec", "impyla", "requests", "sqlalchemy"]
 mssql = ["sqlalchemy", "pymssql", "sqlalchemy-views"]
@@ -5377,4 +5378,4 @@ visualization = ["graphviz"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.8"
-content-hash = "d61d58e049c2b8fb56429e651945ae82f1097de873b78748dd36f7ce935e3a92"
+content-hash = "31ede702597c1952f600129a8af54e43994948e10c59c49923594a8a9308de30"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -195,6 +195,7 @@ duckdb = [
   "sqlalchemy",
   "sqlalchemy-views",
 ]
+flink = []
 geospatial = ["geoalchemy2", "geopandas", "shapely"]
 impala = ["fsspec", "impyla", "requests", "sqlalchemy"]
 mssql = ["sqlalchemy", "pymssql", "sqlalchemy-views"]
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -775,6 +775,79 @@ jobs:
         if: matrix.backend.services != null && failure()
         run: docker compose logs
 
+  test_flink:
+    name: Flink ${{ matrix.os }} python-${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    needs:
+      - gen_lockfile_backends
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - windows-latest
+        python-version:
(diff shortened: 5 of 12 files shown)


poetry.lock
deepyaman, line 1: @chloeh13q Should we revert this file? I don't think it needs to change, given there are no new requirements introduced.

README.md
```diff
@@ -54,6 +54,7 @@ Ibis acts as a universal frontend to the following systems:
 
 - [Apache Arrow DataFusion](https://ibis-project.org/backends/Datafusion/) (experimental)
 - [Apache Druid](https://ibis-project.org/backends/Druid/) (experimental)
+- [Apache Flink](https://ibis-project.org/backends/Flink) (experimental)
```
deepyaman, line 57:
```suggestion
- [Apache Flink](https://ibis-project.org/backends/Flink/) (experimental)
```
Super nit :P

jcrist: Since this isn't a full fledged backend yet, I think we should leave it out of this list for now. Otherwise I expect we'll get users confused about why they can't do `ibis.flink.connect(...)` yet.

docs/backends/Flink.md
```diff
@@ -0,0 +1,6 @@
+---
+backend_name: Flink
+backend_url: https://flink.apache.org/
+backend_module: flink
+is_experimental: true
+---
```
jcrist, line 6: Since this isn't a full fledged backend yet, I think we should leave it out of the backends docs for now.

docs/backends/index.md
```diff
@@ -10,6 +10,7 @@ The first category of backend translate Ibis table expressions into query string
 The compiler turns each table expression into a query string and passes that query
 to the database through a driver API for execution.
 
+- [Apache Flink](Flink.md)
```
jcrist, line 13: Same here. I'd drop this for now.

ibis/backends/flink/registry.py
```diff
@@ -0,0 +1,72 @@
...
+from ibis.backends.base.sql.registry import (
+    operation_registry as base_operation_registry,
+)
+from ibis.backends.flink.utils import format_interval_literal_segments
+
+operation_registry = base_operation_registry.copy()
+
+
+def _count_star(translator: ExprTranslator, op: ops.Node) -> str:
+    return "count(*)"
```
jcrist, line 13: Since you're inheriting the base registry, `CountStar` should already be handled - I don't believe you'll need to define a handler for this op here.

ibis/backends/flink/registry.py
```diff
@@ -0,0 +1,72 @@
...
+    elif dtype.is_date():
+        typeclass = 'date'
+    elif dtype.is_numeric():
+        typeclass = 'number'
+    elif dtype.is_timestamp():
+        typeclass = 'timestamp'
+    elif dtype.is_interval():
+        typeclass = 'interval'
+    elif dtype.is_set():
+        typeclass = 'set'
```
jcrist, line 43: The `Set` type has been removed in the dev branch - `dtype.is_set()` no longer exists.

```suggestion
```

ibis/backends/flink/registry.py
```diff
@@ -0,0 +1,72 @@
...
+    if unit == "ms":
+        return f"TO_TIMESTAMP_LTZ({helpers.quote_identifier(arg.name, force=True)}, 3)"
+    elif unit == "s":
+        return f"TO_TIMESTAMP_LTZ({helpers.quote_identifier(arg.name, force=True)}, 0)"
+    raise ValueError(f"{unit!r} unit is not supported!")
+
+
+def _string_concat(translator, op):
+    args = ", ".join(map(translator.translate, op.arg))
+    return f"CONCAT({args})"
```
jcrist, line 62: The base registry should already handle `StringConcat` this way,  I don't believe defining an override here is necessary.

ibis/backends/flink/tests/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+from ibis.backends.tests.base import BackendTest, RoundAwayFromZero, UnorderedComparator
+
+FLINK_CATALOG = "default_catalog"
+FLINK_DATABASE = "default_database"
+
+
+class TestConf(UnorderedComparator, BackendTest, RoundAwayFromZero):
+    @staticmethod
+    def connect(data_directory: Path):
+        return ibis.flink.connect(catalog=FLINK_CATALOG, database=FLINK_DATABASE)
```
jcrist, line 13: Since we're not implementing execution tests yet, I don't think you should define this subclass at all. This whole file may be deleted. Deleting this should also let you revert the changes in `ibis/backends/conftest.py`.

poetry.lock
jcrist: Yeah, this whole file can be reverted.

pyproject.toml
```diff
@@ -223,6 +224,7 @@ dask = "ibis.backends.dask"
 datafusion = "ibis.backends.datafusion"
 druid = "ibis.backends.druid"
 duckdb = "ibis.backends.duckdb"
+flink = "ibis.backends.flink"
```
jcrist, line 227: Since flink isn't a full-fledged backend yet, we should revert this entrypoint registry. Doing this should also let you revert the changes in `ibis/__init__.py` and `gen_matrix.py`.

ibis/__init__.py
```diff
@@ -48,6 +48,13 @@ def __getattr__(name: str) -> BaseBackend:
     the `ibis.backends` entrypoints. If successful, the `ibis.sqlite`
     attribute is "cached", so this function is only called the first time.
     """
+    # Prevent flink backend from being accessed as `ibis.flink`, since this
+    # will raise `AttributeError: module 'ibis.backends.flink' has no
+    # attribute 'Backend'`. TODO(chloeh13q): we will remove this when we
+    # have an implementation for Flink backend.
+    if name == "flink":
+        raise AttributeError("module 'ibis' has no attribute 'flink'")
+
```
jcrist, line 57: This can be reverted (see below).

ibis/backends/conftest.py
```diff
@@ -273,6 +266,17 @@ def pytest_ignore_collect(path, config):
...
+
+    # get the backend path part
+    #
+    # path is a py.path.local object hence the conversion to Path first
+    backend = _get_backend_from_parts(Path(path).parts)
+    if backend is None or backend not in _get_backend_names():
+        # TODO(chloeh13q): temporarily ignore all base tests if the marker selects for
+        # flink. In the future, we should re-enable base tests or appropriately mark tests
+        # using the `notimpl` marker
+        return mark_expr == "flink"
```
jcrist, line 278: This can be reverted (see below).

ibis/backends/flink/tests/test_translator.py
```diff
@@ -0,0 +1,9 @@
+import ibis
+from ibis.backends.flink.compiler.core import translate
+
+
+def test_translate_toplevel(snapshot):
+    t = ibis.table([('foo', 'double')], name='t0')
+    expr = t.foo.sum()
+    result = translate(expr.as_table().op())
+    snapshot.assert_match(str(result), "out.sql")
```
jcrist, line 9: It'd be good to add some more tests in `test_translator.py` to cover all the operations you defined in `registry.py`. Right now most of that file has no coverage.

jcrist (commented):
Thanks for the PR! This is off to a good start - I've left some comments below.

poetry.lock
chloeh13q: if i don't commit this file, it will fail the `Check Generated Files / check_setuptools_install` workflow
Screenshot 2023-06-12 at 2 27 16 PM

jcrist: Hmmm, I'm a bit surprised that this change would affect the lockfile at all (since no dependencies change). But you're right, you will need to update the lockfile then. Note that when doing so you'll need to make sure you're using the version of poetry pinned in our test suite (currently pinned at `poetry<1.4`).

ibis/backends/flink/registry.py
chloeh13q: i think it compiles into something like `SELECT t0.`i`, count(1) AS `count`` instead of `count(*)`

jcrist:
Hi Chloe, overall this looks good. I pushed up a commit fixing the lockfile changes - what version of `poetry` did you use when updating the lockfile?

Last thing - can you squash your commits? We generate our changelog from commit messages - specifically anything with `feat` or `fix` will feature highly in the changelog (see e.g. https://ibis-project.org/release_notes/#features). Commits with `chore` as the type are ignored. IMO this PR should ideally be a single commit with a message like `feat(flink): add initial flink SQL compiler`.

chloeh13q:
@jcrist I'm using poetry 1.2.2. Also squashed the commits. Lmk if there's anything else you'd like me to address!

jcrist (approved):
LGTM, thansk! :rocket:

PR #6418, opened 2023-06-10
fix(deps): prevent vulnerable snowflake-connector-python versions

maresb:
There is a vulnerability in snowflake-connector-python in versions below v3.0.2: <https://github.com/advisories/GHSA-5w5m-pfw9-c8fp>

I'm not so sure what are the considerations going into this lower pin, so if this isn't viable, feel free to just close. I mainly wanted to bring attention to this vulnerability.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -5363,4 +5363,4 @@ visualization = ["graphviz"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.8"
-content-hash = "de298252f963b9ba073c0639b35e3f5805a79e432691a0e7e9f827aad3bc16d4"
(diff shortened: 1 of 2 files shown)


cpcloud:
@maresb Welcome 👋🏻!

Thanks for the PR, and for the heads up.

> I'm not so sure what are the considerations going into this lower pin

The only considerations were setting as low a bound as possible that also includes the feature set that we need in ibis.

I'll update the commit message to include the breaking change since this will require all existing users to upgrade their version of `snowflake-connector-python`.

maresb:
Cool, thanks!

cpcloud (approved):

PR #6424, opened 2023-06-12
perf(imports): speed up checking for geospatial support

cpcloud:
This PR speeds up one of the sources of slow backend module import times without needing to rewrite all imports to the call site or change a lot of existing code.

### DIFF
--- a/ibis/backends/base/sql/alchemy/geospatial.py
+++ b/ibis/backends/base/sql/alchemy/geospatial.py
@@ -1,10 +1,8 @@
 from __future__ import annotations
 
-try:
-    import geoalchemy2
-    import geoalchemy2.shape  # noqa: F401
-    import geopandas  # noqa: F401
-except ImportError:
-    geospatial_supported = False
-else:
-    geospatial_supported = True
+from importlib.util import find_spec as _find_spec
+
+geospatial_supported = (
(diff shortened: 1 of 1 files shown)


ibis/backends/base/sql/alchemy/geospatial.py
```diff
@@ -1,10 +1,8 @@
...
-    import geoalchemy2.shape  # noqa: F401
-    import geopandas  # noqa: F401
-except ImportError:
-    geospatial_supported = False
-else:
-    geospatial_supported = True
+from importlib.util import find_spec as _find_spec
+
+geospatial_supported = (
+    _find_spec("geoalchemy2") is not None and _find_spec("geopandas") is not None
```
kszucs, line 6: Perhaps a utility function supporting multiple arguments?

`if imports_available("geoalchemy2", "geopandas")`?

kszucs (approved):

ibis/backends/base/sql/alchemy/geospatial.py
```diff
@@ -1,10 +1,8 @@
 from __future__ import annotations
 
-try:
-    import geoalchemy2
-    import geoalchemy2.shape  # noqa: F401
```
cpcloud, line 5: I was able to remove this because it's always available in the versions of `geoalchemy2` + `shapely` that we support.

PR #6429, opened 2023-06-12
ci: publish pre-release dists to pypi

cpcloud:
Turns on publishing pre-releases to pypi proper.


cpcloud:
Merging, no need to run the backend suite twice!

PR #6437, opened 2023-06-13
chore(dev): remove use of mapfile because macOS ships with a very old version of bash

### LINKED ISSUE #6427: bug: `mapfile` doesn't exist on macOS's version of bash which causes `just doctest` to fail
### What happened?

I would expect to be able to run the tests successfully on macOS but the failure is ignored and causes files that should be ignored to not be ignored.

`master`

N/A

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Closes #6427.

### DIFF
--- a/justfile
+++ b/justfile
@@ -52,7 +52,7 @@ doctest *args:
     #!/usr/bin/env bash
 
     # TODO(cpcloud): why doesn't pytest --ignore-glob=test_*.py work?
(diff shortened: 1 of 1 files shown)


cpcloud:
@lostmygithubaccount Can you give this PR a try?

jcrist (approved):

PR #6439, opened 2023-06-13
docs(api): manually add inherited mixin methods to timey classes

### LINKED ISSUE #6428: bug: .epoch_seconds() doesn't show up in docs
### What happened?

title; there is a docstring but it's not discoverable in documentation (other than mention in release notes)

latest

n/a

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
This wasn't exactly fun and it isn't the MOST maintainable thing, but
until mkdocstrings has support for rendering inherited class members,
this will at least give users something to look at / refer to.

Fixes #6428

### DIFF
--- a/docs/api/expressions/timestamps.md
+++ b/docs/api/expressions/timestamps.md
@@ -6,15 +6,164 @@ All temporal operations are valid for both scalars and columns.
 ::: ibis.expr.types.temporal.TemporalValue
     options:
       show_source: true
+
+
 ::: ibis.expr.types.temporal.TimestampValue
     options:
       show_source: true
+      members: false
+<!-- prettier-ignore-end -->
+
+### Functions
+
+<!-- prettier-ignore-start -->
+::: ibis.expr.types.temporal.TimestampValue.add
+    options:
+      heading_level: 4
+::: ibis.expr.types.temporal._TimeComponentMixin.between
+    options:
+      heading_level: 4
+::: ibis.expr.types.temporal.TimestampValue.date
+    options:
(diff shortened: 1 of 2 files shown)


docs/api/expressions/timestamps.md
```diff
@@ -6,15 +6,164 @@ All temporal operations are valid for both scalars and columns.
...
 ::: ibis.expr.types.temporal.TimestampValue
     options:
       show_source: true
+      members: false
+<!-- prettier-ignore-end -->
+
+### Functions
+
+<!-- prettier-ignore-start -->
+::: ibis.expr.types.temporal.TimestampValue.add
```
cpcloud, line 20: I wonder if it's actually less work to duplicate the method definitions in the class. I think they're all one-liners and there's only two of them.

gforsyth: it ends up looking pretty funky, because `mkdocstrings` will pick up those two methods, place them  under a header called `Functions`, then the manually inserted ones will appear below, so you end up with

```
*** Functions
**** add
**** substract
*** Functions
**** between
...
...
```

I don't think we want to duplicate all of the mixin methods in each of these (cue conversation on whether code structure should be influenced by ability of documentation tool)

cpcloud: Ugh, yeah okay, let's leave it as you have it!

cpcloud:
@gforsyth https://github.com/mkdocstrings/griffe/pull/170 :)

PR #6440, opened 2023-06-13
docs: add read_delta and related to backends docs

lostmygithubaccount:
adds the `read_delta` docstrings to the relevant backend docs and add in the badges

minor consistency in DuckDB docstrings with `**kwargs` instead of `kwargs`

~~in the top-level API doc, move `read_delta` to after `read_parquet` for consistency~~

### DIFF
--- a/docs/backends/DuckDB.md
+++ b/docs/backends/DuckDB.md
@@ -2,8 +2,18 @@
 backend_name: DuckDB
 backend_url: https://duckdb.org/
 backend_module: duckdb
-exports: ["PyArrow", "Parquet", "CSV", "Pandas"]
-imports: ["CSV", "Parquet", "JSON", "PyArrow", "Pandas", "SQLite", "Postgres"]
+exports: ["PyArrow", "Parquet", "Delta Lake", "CSV", "Pandas"]
+imports:
+  [
+    "CSV",
+    "Parquet",
+    "Delta Lake",
+    "JSON",
+    "PyArrow",
+    "Pandas",
+    "SQLite",
+    "Postgres",
+  ]
 ---
 
 {% include 'backends/badges.md' %}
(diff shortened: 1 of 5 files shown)


docs/backends/DuckDB.md
```diff
@@ -89,6 +99,10 @@ con = ibis.connect("duckdb://") # (1)
     options:
       heading_level: 4
       show_docstring_returns: false
+::: ibis.backends.duckdb.Backend.read_delta
+    options:
+      heading_level: 4
+      show_docstring_returns: false
```
gforsyth, line 105: Probably should make this an alphabetical list, too.

gforsyth (changes requested):
More docs!

I think we should have the top-level API docs sorted alphabetically, and we should probably do the same for the `read_*` methods on the backend docs.  I was just adding things as I went when I wrote these so the inconsistency is mine.

It probably also makes sense to sort the ordering for the badges -- maybe we just add `sorted` to the template in `badges.md`?

docs/backends/DuckDB.md
lostmygithubaccount: seems like a larger lift? they don't seem to be in alphabetical order too

what's the `sorted` in `badges.md` suggestion? I'd prefer not to manually sort them 🙃

gforsyth: yeah, let's not manually sort the badges.  I was thinking this:

```diff
diff --git a/docs/backends/badges.md b/docs/backends/badges.md
index 8e30aeca6..7201770e7 100644
--- a/docs/backends/badges.md
+++ b/docs/backends/badges.md
@@ -1,3 +1,3 @@
-{% if imports %} ![filebadge](https://img.shields.io/badge/Reads-{{ "%20|%20".join(imports) }}-blue?style=flat-square) {% endif %}
+{% if imports %} ![filebadge](https://img.shields.io/badge/Reads-{{ "%20|%20".join(sorted(imports)) }}-blue?style=flat-square) {% endif %}
 
-{% if exports %} ![exportbadge](https://img.shields.io/badge/Exports-{{ "%20|%20".join(exports) }}-orange?style=flat-square) {% endif %}
+{% if exports %} ![exportbadge](https://img.shields.io/badge/Exports-{{ "%20|%20".join(sorted(exports)) }}-orange?style=flat-square) {% endif %}
```

lostmygithubaccount: @gforsyth updated w/ the suggestion!

gforsyth (approved):
:shipit:

PR #6446, opened 2023-06-14
fix(expr): fix graphviz rendering

chloeh13q:
The expression rendering is broken (as shown in the picture below):
broken_graphviz

This PR fixes the expression rendering:
fixed-graphviz

### DIFF
--- a/ibis/expr/visualize.py
+++ b/ibis/expr/visualize.py
@@ -15,7 +15,7 @@ from ibis.common.graph import Graph
 
 def get_type(node):
     with contextlib.suppress(AttributeError, NotImplementedError):
-        return str(node.output_dtype)
+        return escape(str(node.output_dtype))
 
     try:
         schema = node.schema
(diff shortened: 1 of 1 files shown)


ibis/expr/visualize.py
```diff
@@ -41,37 +41,52 @@ def get_type(node):
...
     )
 
 
 def get_label(node):
     typename = get_type(node)  # Already an escaped string
+    # If the typename is of the type array<[type]>, e.g. array<int8>,
+    # graphviz will try to render <int8> as an HTML tag and fail
+    if typename.startswith("array<") and typename.endswith(">"):
+        typename = typename.replace("<", "[")
+        typename = typename.replace(">", "]")
```
cpcloud, line 56: Would `urllib.parse.quote_plus(typename)` also work?

cpcloud: Nevermind, there's a different way to do this!

cpcloud: Went ahead and did it in the interest of landing the PR!

cpcloud (approved):

ibis/expr/visualize.py
chloeh13q: Thank you!

Issue #5941, opened 2023-04-07
feat(ddl): support PyArrow objects in `create_table`


ianmcook:
### Describe the solution you'd like

It would be really nice if `create_table` allowed users to pass PyArrow Table and RecordBatch objects as `obj`.

### What version of ibis are you running?

5.0.0

### What backend(s) are you using, if any?

Trino

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gforsyth:
Added in #6470 (we don't support using record batch readers because it leads to unpredictable behavior)

PR #6452, opened 2023-06-15
feat(memtable): allow creating memtable from RecordBatchReader

gforsyth:
xref #5941 

Adding support for other objects in `create_table` is (usually) about making them available as `memtable`able objects.  
This adds naive support for feeding a `RecordBatchReader` into `memtable` -- I say naive because it will load all the batches into memory at creation time.

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -949,6 +949,29 @@ def test_memtable_construct(backend, con, monkeypatch):
     backend.assert_frame_equal(t.execute(), pa_t.to_pandas())
 
 
+@pytest.mark.notimpl(["datafusion"], raises=com.OperationNotDefinedError)
+@pytest.mark.broken(
+    ["druid"],
+    raises=AssertionError,
+    reason="result contains empty strings instead of None",
+)
+def test_memtable_construct_rbr(backend, con, monkeypatch):
+    pa = pytest.importorskip("pyarrow")
+    monkeypatch.setattr(ibis.options, "default_backend", con)
+
+    pa_t = pa.Table.from_pydict(
+        {
+            "a": list("abc"),
+            "b": [1, 2, 3],
+            "c": [1.0, 2.0, 3.0],
+            "d": [None, "b", None],
+        }
+    )
+    pa_rbr = pa.RecordBatchReader.from_batches(pa_t.schema, pa_t.to_batches())
+    t = ibis.memtable(pa_rbr)
+    backend.assert_frame_equal(t.execute(), pa_pyarrow objects.invt.to_pandas())
+
+
 @pytest.mark.notimpl(
     ["dask", "datafusion", "pandas", "polars"],
     raises=NotImplementedError,
--- a/ibis/expr/operations/relations.py
+++ b/ibis/expr/operations/relations.py
@@ -126,6 +126,16 @@ class PyArrowTableProxy(TableProxy):
         return self._data
 
 
+class PyArrowRBRProxy(TableProxy):
+    __slots__ = ()
+
+    def to_frame(self):
+        return self._data.read_pandas()
+
+    def to_pyarrow(self, schema: Schema) -> pa.Table:
+        return self._data.read_all()
+
+
 class PandasDataFrameProxy(TableProxy):
     __slots__ = ()
 
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -424,6 +424,26 @@ def _memtable_from_pyarrow_table(
     ).to_expr()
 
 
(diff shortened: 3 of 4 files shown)


ibis/expr/operations/relations.py
```diff
@@ -126,6 +126,16 @@ def to_pyarrow(self, schema: Schema) -> pa.Table:
         return self._data
 
 
+class PyArrowRBRProxy(TableProxy):
```
cpcloud, line 129:
```suggestion
class PyArrowBatchesProxy(TableProxy):
```

Would that work?

cpcloud (changes requested):
LGTM, nothing blocking!

cpcloud (changes requested):
LGTM, nothing blocking!

ibis/backends/tests/test_generic.py
```diff
@@ -949,6 +949,29 @@ def test_memtable_construct(backend, con, monkeypatch):
...
+        {
+            "a": list("abc"),
+            "b": [1, 2, 3],
+            "c": [1.0, 2.0, 3.0],
+            "d": [None, "b", None],
+        }
+    )
+    pa_rbr = pa.RecordBatchReader.from_batches(pa_t.schema, pa_t.to_batches())
+    t = ibis.memtable(pa_rbr)
+    backend.assert_frame_equal(t.execute(), pa_pyarrow objects.invt.to_pandas())
```
cpcloud, line 972: Looks like a syntax error here.

ibis/expr/operations/relations.py
```diff
@@ -126,6 +126,16 @@ def to_pyarrow(self, schema: Schema) -> pa.Table:
...
+    __slots__ = ()
+
+    def to_frame(self):
+        return self._data.read_pandas()
+
+    def to_pyarrow(self, schema: Schema) -> pa.Table:
+        return self._data.read_all()
```
jcrist, line 136: The `to_pyarrow` call for a `memtable` registered for a `RecordBatchReader` will only work once, meaning that the `memtable` can only ever have `execute` executed against it once. This seems like a source of confusion for users.

jcrist (commented):
I'm a bit against this. Creating a `memtable` should be a cheap way to wrap an in-memory datasource and expose it as a table in ibis expressions. I don't think `RecordBatchReader`s can support this contract:

- They can only ever be computed on once. Subsequent computations will silently yield empty data, which is likely to be a source of subtle bugs and wrong answers.
- This implementation loads the data fully into memory, which isn't always cheap. Since some backends can work natively with a `RecordBatchReader` (e.g. duckdb), some users may expect `con.register(reader)` and `ibis.memtable(reader)` to have the same perf characteristics, but in this case they won't.

Is the goal here just to allow passing a `RecordBatchReader` to `create_table`? If so, I think we can support that in a more efficient way without adding `memtable` support.

ibis/expr/operations/relations.py
cpcloud: Hm, that's a good point. Perhaps we need some stronger conceptual and documented requirements for something to be `memtable`-able, such as "must be replayable/repeatable/able to be consumed more than once"

cpcloud:
I think @jcrist makes some good points so I'm removing my approval, though I am +1 on `create_table` supporting record batches.

gforsyth:
I view it as a convenience -- we aren't promising performance with `memtable`, so much as compatibility.

> * They can only ever be computed on once. Subsequent computations will silently yield empty data, which is likely to be a source of subtle bugs and wrong answers.

That is the major footgun, but if we load it into memory, it's a non-issue (on the Ibis side).

> * This implementation loads the data fully into memory, which isn't always cheap. Since some backends can work natively with a `RecordBatchReader` (e.g. duckdb), some users may expect `con.register(reader)` and `ibis.memtable(reader)` to have the same perf characteristics, but in this case they won't.

It's not always cheap.  We can make the same argument about calling `ibis.memtable` on a pandas dataframe vs `con.register` on a pandas dataframe.

ibis/backends/tests/test_generic.py
gforsyth: yeah, bad keystroke pasted in some junk.  I've committed a fix locally, but I'll save the CI run to see if we want to do this or not.

gforsyth:
> We can make the same argument about calling `ibis.memtable` on a pandas dataframe vs `con.register` on a pandas dataframe.

My brain just re-read this and yeah, that's not right.
In any case, it's not always cheap.  I don't know that that's a reason not to allow it.

gforsyth:
The record batch readers weren't getting exhausted on repeated execution, which ended up being because I was only testing it against DuckDB.  Given that level of footgun, definitely going to shelve this one.

PR #6468, opened 2023-06-16
ci: run nix macos jobs on macos-13 to try and avoid SIP

cpcloud:
Fixes an issue with GitHub changing their macos deployments to include SIP on macos and nix failing to work with that. xref https://github.com/cachix/install-nix-action/issues/183

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -37,7 +37,7 @@ jobs:
           - "3.10"
           - "3.11"
         include:
-          - os: macos-latest
+          - os: macos-13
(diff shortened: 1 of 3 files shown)


.github/workflows/nix.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           - "3.10"
           - "3.11"
         include:
-          - os: macos-latest
+          - os: macos-13
```
cpcloud, line 40: the list of these is hardcoded in actionlint unfortunately

PR #6469, opened 2023-06-16
fix(duckdb): ensure `to_timestamp` returns a UTC timestamp

jcrist:
Previously these were naive timestamps in the backend (timestamp, not timestamp with timezone), and the timezone was being applied in `.execute()` as part of the expected schema being enforced. This would cause issues if chaining operations (since the type in the backend was different) or if returning through `to_pyarrow` since the result wouldn't have a timezone attached.

I'm not sure of a good way to test this with our current test setup. A test using `to_pyarrow` would show the issue, but not all backends support `to_pyarrow` yet, kinda complicating the testing path. Either way, I can confirm this patch fixes the issue.

### DIFF
--- a/ibis/backends/duckdb/registry.py
+++ b/ibis/backends/duckdb/registry.py
@@ -178,12 +178,22 @@ def _json_get_item(left, path):
 
 
 def _strftime(t, op):
-    format_str = op.format_str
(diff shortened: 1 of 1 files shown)


lostmygithubaccount:
@cpcloud any chance this would be related to what I was seeing w/ delta tables?

PR #6470, opened 2023-06-16
feat(create_table): support pyarrow Table in table creation

gforsyth:
Some backends already supported this, but a bunch didn't only because of `isinstance` checks.

The remaining backends that don't have support for pyarrow tables in `create_table` are `dask`, `datafusion`, `druid`, and `impala`.  
I also ran into a big in the bigquery backend, where passing both `obj` and `schema` to `create_table` led to `obj` being ignored entirely.  

Also there's probably some import cost to this, although it should all be inlined and will only hit at the `create_table` call.

### DIFF
--- a/ibis/backends/bigquery/__init__.py
+++ b/ibis/backends/bigquery/__init__.py
@@ -440,7 +440,7 @@ class Backend(BaseSQLBackend):
     def create_table(
         self,
         name: str,
-        obj: pd.DataFrame | ir.Table | None = None,
+        obj: pd.DataFrame | pa.Table | ir.Table | None = None,
         *,
         schema: ibis.Schema | None = None,
         database: str | None = None,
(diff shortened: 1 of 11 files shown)


ibis/backends/bigquery/__init__.py
```diff
@@ -457,19 +457,29 @@ def create_table(
...
-            table_id = self._fully_qualified_name(name, database)
-            table = bq.Table(table_id, schema=BigQuerySchema.from_ibis(schema))
-            self.client.create_table(table)
-        else:
+
+        if isinstance(obj, ir.Table) and schema is not None:
+            if not schema.equals(obj.schema()):
+                raise com.IbisTypeError(
+                    """Provided schema and Ibis table schema are incompatible.
+                    Please align the two schemas, or provide only one of the two arguments."""
```
cpcloud, line 465: How does this render in the terminal given the preservation of whitespace?

cpcloud (approved):

ibis/backends/bigquery/__init__.py
gforsyth: Very badly!

gforsyth: pushed up a fix

PR #6472, opened 2023-06-16
feat(api): better repr and pickle support for deferred expressions

jcrist:
This is a major refactor of the `Deferred` implementation, with the following goals:

- Better repr. The previous repr was just the default python object repr, making it impossible to see what the underlying deferred expression represented. The new repr attempts to mimic the python syntax used to generate the deferred expression.
- Pickleability. The previous expressions weren't pickleable, which prevented using them in place of lambdas for certain workflows. They now are pickleable.

```python
In [1]: from ibis import _

In [2]: _.x.log(1)
Out[2]: _.x.log(1)

In [3]: _.x.log(1) + _.y
Out[3]: (_.x.log(1) + _.y)
```

### DIFF
--- a/ibis/expr/deferred.py
+++ b/ibis/expr/deferred.py
@@ -1,142 +1,275 @@
 from __future__ import annotations
 
 import operator
-from typing import TYPE_CHECKING, Any, Callable
+from typing import Any, Callable
 
-import toolz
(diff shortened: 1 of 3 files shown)


cpcloud (approved):
Wonderful! This is really great.

PR #6476, opened 2023-06-17
feat(flink): implement translation rules for literal expressions in flink compiler

chloeh13q:
Implement translation rules for literal expressions in Flink compiler.


ibis/backends/flink/compiler/core.py
```diff
@@ -46,15 +49,53 @@ class FlinkCompiler(Compiler):
...
+    elif dtype.is_string():
+        quoted = value.replace("'", "''").replace("\\", "\\\\")
+        return f"'{quoted}'"
+    elif dtype.is_date():
+        if isinstance(value, datetime.date):
+            value = value.strftime('%Y-%m-%d')
+        return repr(value)
+    elif dtype.is_numeric():
+        if math.isnan(value):
+            raise ValueError("NaN is not supported in Flink SQL")
```
deepyaman, line 77: Can it not be translated to null?

ibis/backends/flink/compiler/core.py
```diff
@@ -46,15 +49,53 @@ class FlinkCompiler(Compiler):
...
+        elif math.isinf(value):
+            raise ValueError("Infinity is not supported in Flink SQL")
+        return repr(value)
+    elif dtype.is_timestamp():
+        # TODO(chloeh13q): support timestamp with local timezone
+        if isinstance(value, datetime.datetime):
+            return 'TIMESTAMP ' + repr(value.strftime('%Y-%m-%d %H:%M:%S'))
```
deepyaman, line 84: Flink parses from format `"yyyy-MM-dd HH:mm:ss[.SSS]"`, so I think milliseconds should be included, if available.

ibis/backends/flink/compiler/core.py
```diff
@@ -46,15 +49,53 @@ class FlinkCompiler(Compiler):
...
+        raise NotImplementedError(f'No translation rule for timestamp {value}')
```
deepyaman, line 85: Looking at the Clickhouse backend, it seems to just assume `.is_timestamp()` implies you can access attributes like `.microsecond`, so I wonder if this is necessary? Haven't checked other backend implementations (yet?).

ibis/backends/flink/compiler/core.py
```diff
@@ -46,15 +49,53 @@ class FlinkCompiler(Compiler):
...
+    value = op.value
+    dtype = op.output_dtype
+
+    if dtype.is_boolean():
+        # TODO(chloeh13q): Flink supports a third boolean called "UNKNOWN"
+        return 'TRUE' if value else 'FALSE'
+    elif dtype.is_string():
+        quoted = value.replace("'", "''").replace("\\", "\\\\")
+        return f"'{quoted}'"
+    elif dtype.is_date():
```
deepyaman, line 71: Nit: Feels like it would be more logical to have this with the other datetime handling below.

ibis/backends/flink/compiler/core.py
deepyaman: Also, this seems to be the main place you've errored if not `isinstance` for the timestamp checks; feels inconsistent, but I may be missing something.

deepyaman (commented):
Took a quick initial pass; haven't dived into the interval stuff at all yet.

ibis/backends/flink/compiler/core.py
chloeh13q: I think it's pretty normal to handle this separately, as DATE type is its own type. For example, see the implementation in the base class: https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/registry/literal.py#L74

ibis/backends/flink/compiler/core.py
chloeh13q: I've discussed this with @zhenzhongxu - I'm not sure if it's entirely appropriate / accurate to translate NaN into NULL because they have different meanings (also see [Flink's documentation on NULL](https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/dev/table/types/#null)). We were not able to find any documentation on whether `NaN` exists in Flink, so I think it might be best to error out for now.

ibis/backends/flink/compiler/core.py
chloeh13q: > I wonder if this is necessary

Is what necessary?

chloeh13q: > this seems to be the main place you've errored if not `isinstance` for the timestamp checks

I took the `isinstance` check from the base SQL implementation & BigQuery, the difference being that the other two implementations do not error out if it's not a `datetime` instance. I just chose to add an error here.

ibis/backends/flink/utils.py
```diff
@@ -0,0 +1,232 @@
+import datetime
+import math
+from abc import ABC, abstractmethod
+from collections import defaultdict
+
+from ibis.common.temporal import IntervalUnit
+from ibis.util import convert_unit
+
+MIN_ALLOWED_PRECISION = {
```
zhenzhongxu, line 9: is precision an ibis or flink concept? if the later, we should add some comments for clarity (i.e., what does each precision mean).

ibis/backends/flink/utils.py
```diff
@@ -0,0 +1,232 @@
...
+def _format_precision(precision: int, unit: IntervalUnit) -> str:
+    if precision > MAX_ALLOWED_PRECISION[unit]:
+        raise ValueError(
+            f"{precision} is bigger than the allowed precision for {unit} ({MAX_ALLOWED_PRECISION[unit]})"
+        )
+
+    return '' if precision <= 2 else f'({precision})'
+
+
+class FlinkIntervalType(ABC):
```
zhenzhongxu, line 79: what's ibis's standard on docstring? we should still try to add docstring for public classes and methods

zhenzhongxu (commented):
Thanks for putting this together!

ibis/backends/flink/compiler/core.py
deepyaman: Oh, this was a total nit to just put the if block physically closer to the other datetime handling; 0 actual change. :)

ibis/backends/flink/tests/test_literals.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.mark.parametrize(
+    "value",
+    [
+        param(5, id="int"),
+        param(1.5, id="float"),
+        param(True, id="true"),
+        param(False, id="false"),
+    ],
+)
+def test_simple_literals(snapshot, value):
```
jcrist, line 21: While snapshot tests can work here, for short cases like this where the expected value isn't a long or multi-line string, I slightly prefer manually writing out the full expected value in the parametrized test itself. For translating literals I'd expect this to be the case here.

Snapshot tests are nice when testing a larger query structure, where the output is larger and small changes in formatting (e.g. whitespace changes) can result in test failures. Snapshot tests make it easier in those cases to compare the failures with the new results and check that they're semantically the same.

Something like (untested):

```python
@pytest.mark.parametrize(
    "value, sol",
    [
        param(5, "5", id="int"),
        param(1.5, "1.5", id="float"),
        param(True, "TRUE", id="true"),
        param(False, "FALSE", id="false"),
    ],
)
def test_simple_literals(value, sol):
    expr = ibis.literal(value)
    result = translate(expr.op())
    assert result == sol
```

I'd do the same for all these literal tests. Definitely a style preference though, not a big deal if you disagree.

chloeh13q: No I think that's valid. Thanks for the feedback. I've made the changes.

ibis/backends/flink/utils.py
```diff
@@ -0,0 +1,232 @@
...
+            f"{_format_value_with_precision(minutes, self.precisions[IntervalUnit.MINUTE])}:"
+            f"{_format_value_with_precision(seconds, self.precisions[IntervalUnit.SECOND])}."
+            f"{_format_value_with_precision(microseconds, self.precisions[IntervalUnit.MICROSECOND])}' "
+            f"DAY{_format_precision(self.precisions[IntervalUnit.DAY], IntervalUnit.DAY)} "
+            "TO SECOND"
+            f"{_format_precision(self.precisions[IntervalUnit.SECOND], IntervalUnit.SECOND)}"
+        )
+
+
+class IntervalType:
```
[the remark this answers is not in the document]
chloeh13q: Good point. Thanks for the suggestion!

ibis/backends/flink/utils.py
chloeh13q: Added!

jcrist (approved):

PR #6484, opened 2023-06-19
feat(polars): add support for limit with `offset`, binary literals, and `dropna(how='all')`

alexander-beedie:
Some low-hanging fruit for the `polars` backend ;)

* support limit + offset.
* support binary literals (bytes).
* support "how='all'" for dropna.


alexander-beedie:
Note: the failing test appears to be completely unrelated; `test_trig_functions_columns` errored against an `mssql` backend.

alexander-beedie:
> Excellent work, thank you!

I'll have a dig through and see what else is missing 👍

PR #6490, opened 2023-06-20
feat(polars): add support for `identical_to`

alexander-beedie:
Added polars support for `identical_to`, and tweaked some `@pytest.mark` designations relating to window functions; polars has support for these, but `ops.WindowFunction` hasn't been implemented yet (I might have a crack at that later), so I moved the label into the more appropriate message.

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -328,6 +328,13 @@ def fillna(op, **kw):
     return table.select(columns)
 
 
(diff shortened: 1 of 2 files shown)


alexander-beedie:
Odd; an unrelated `mssql` test failure (`test_bitwise_columns`).

cpcloud:
We're aware of the mssql failures, but it's not 100% clear what's causing them.

You can ignore them for the purposes of this PR!

PR #6491, opened 2023-06-20
chore: drop support for Python 3.8

gforsyth:
BREAKING CHANGE: The minimum supported Python version is now Python 3.9


cpcloud (approved):
Just for posterity, we noted this here: https://github.com/ibis-project/ibis/blob/5015677d78909473014a61725d371b4bf772cdff/docs/blog/Ibis-version-3.0.0-release.md?plain=1#L83

And not for this PR, but perhaps a follow up is to put the fact that we follow NEP 29 somewhere with more visibility.

gforsyth:
@cpcloud -- can we remove the `nix 3.8` job from the list of expected checks?

cpcloud:
Yup! Will do in a bit!

cpcloud:
@gforsyth Done!

PR #6493, opened 2023-06-20
feat(polars): allow seamless connection for `DataFrame` as well as `LazyFrame`

alexander-beedie:
The `_add_table` method in the polars Backend was already typed such that it appeared to be expecting to handle both types of frame, but didn't ensure eager → lazy registration - now it does ;)

### DIFF
--- a/ibis/backends/polars/__init__.py
+++ b/ibis/backends/polars/__init__.py
@@ -33,9 +33,9 @@ class Backend(BaseBackend):
         self._context = pl.SQLContext()
 
     def do_connect(
-        self, tables: MutableMapping[str, pl.LazyFrame] | None = None
+        self, tables: MutableMapping[str, pl.LazyFrame | pl.DataFrame] | None = None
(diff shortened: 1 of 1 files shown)


ibis/backends/polars/__init__.py
```diff
@@ -128,6 +128,8 @@ def _register_failure(self):
         )
 
     def _add_table(self, name: str, obj: pl.LazyFrame | pl.DataFrame) -> None:
+        if isinstance(obj, pl.DataFrame):
+            obj = obj.lazy()
```
cpcloud, line 132: Seems reasonable. Wondering if we even need to convert things to be lazy? Can we just leave them as is?

cpcloud: Oh, sweet! Well, in that case, I will approve this PR!

cpcloud (approved):
Awesome, thank you!

PR #6494, opened 2023-06-20
feat(polars): support table registration from multiple parquet files

alexander-beedie:
Added support for registering a single table from multiple parquet files (when passed as an iterable of individual paths - It already handles a directory or file path glob).


cpcloud (approved):
Great!

PR #6500, opened 2023-06-21
feat(duckdb): add support for native and pyarrow UDFs

cpcloud:
This PR adds experimental support for scalar UDFs implemented in Python with
support for vectorized PyArrow-backed inputs as well as support for native
Python inputs.

TODOs:

- [x] Figure out a better way to abstract over `con.create_function` calls
  versus needing to execute a DDL SQL string. ~Right now I'm just passing the
  open connection around and it's only being used by the duckdb backend.~


ibis/backends/duckdb/__init__.py
```diff
@@ -914,6 +914,29 @@ def _get_temp_view_definition(
...
+    def _compile_pyarrow_udf(self, udf_node: ops.ScalarUDF) -> str:
+        func = udf_node.__func__
+        name = func.__name__
+        input_types = [DuckDBType.to_string(arg.output_dtype) for arg in udf_node.args]
+        output_type = DuckDBType.to_string(udf_node.output_dtype)
+        return lambda con: con.connection.connection.create_function(
+            name, func, input_types, output_type, type=duckdb.functional.ARROW
+        )
+
+    def _register_udfs(self, expr: ir.Expr) -> None:
```
cpcloud, line 935: This method probably isn't necessary anymore.

ibis/backends/duckdb/tests/test_udf.py
```diff
@@ -0,0 +1,105 @@
+from __future__ import annotations
```
cpcloud, line 1: This file should likely be moved to the backend test suite given how similar it is between the backends that support Python UDFs.

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -752,7 +752,10 @@ def _compile_pyarrow_udf(self, udf_node: ops.ScalarUDF) -> str:
     def _register_udfs(self, expr: ir.Expr) -> None:
         with self.begin() as con:
             for sql in self._compile_udfs(expr):
-                con.exec_driver_sql(sql)
+                if callable(sql):
+                    sql(con)
```
gforsyth, line 756: Maybe we can name this something other than `sql`?  `sql(con)` looks very strange.

cpcloud: Changed to `register_udf`

ibis/backends/duckdb/tests/test_udf.py
cpcloud: Done!

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -751,8 +751,11 @@ def _compile_pyarrow_udf(self, udf_node: ops.ScalarUDF) -> str:
 
     def _register_udfs(self, expr: ir.Expr) -> None:
         with self.begin() as con:
-            for sql in self._compile_udfs(expr):
-                con.exec_driver_sql(sql)
+            for register_udf in self._compile_udfs(expr):
```
jcrist, line 754: `_compile_udfs` is called in two places currently:

- `_register_udfs`, where the strings/functions yielded by `_compile_udfs` are immediately executed
- `Backend.compile`, where it's only needed for the side effect of registering the dynamically generated UDF ops with the backend (so they're compilable).

If the latter use case was extracted out to a new generic function, then `_compile_udfs` and `_register_udfs` could be merged into the same method, which IMO would feel a bit cleaner than having `_compile_udfs` yield strings or functions to immediately execute on the connection.

ibis/backends/base/sql/__init__.py
```diff
@@ -204,7 +204,7 @@ def to_pyarrow_batches(
 
         return pa.ipc.RecordBatchReader.from_batches(schema.to_pyarrow(), batches)
 
-    def _compile_udfs(self, expr: ir.Expr) -> Iterable[str]:
+    def _compile_udfs(self, expr: ir.Expr) -> Iterable[str | Callable[..., None]]:
         """Return an iterator of DDL strings, once for each UDFs contained within `expr`."""
```
jcrist, line 208: If you don't do the refactor I suggested below, we'd at least want to update this docstring to be correct.

cpcloud: Did the refactor!

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: Done!

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: Ultimately went with a slightly different approach here, so the naming should be even clearer!

ibis/formats/__init__.py
```diff
@@ -45,6 +45,11 @@ def to_ibis(cls, typ: T, nullable: bool = True) -> DataType:
         Ibis DataType.
         """
 
+    @classmethod
+    def to_string(cls, dtype: DataType) -> str:
```
kszucs, line 49: We could have a from_string as well which is basically "parse the backend specific string", we have multiple "parse" implementations scatterend in the codebase actually.

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -749,10 +728,28 @@ def _compile_pyarrow_udf(self, udf_node: ops.ScalarUDF) -> str:
                 f"The {self.name} backend does not support PyArrow-based vectorized scalar UDFs"
             )
 
+    def _define_udf_translation_rules(self, expr):
```
kszucs, line 731: I find this implementation fairly complicated, we should refactor it soon enough.

kszucs (approved):

kszucs:
@cpcloud tests are failing

PR #6501, opened 2023-06-21
test(snowflake): xfail annotated map UDFs on snowflake

cpcloud:
Fixes the failing upstream snowflake tests.


gforsyth (approved):
We should remember to remove this after #6491 is merged.

PR #6502, opened 2023-06-21
docs(python-versions): document how we decide to drop support for Python versions

cpcloud:
Adds docs that reference NEP29 to loudly document how we deal with Python versions support.


gforsyth (approved):
:shipit:

Issue #6503, opened 2023-06-21
bug: mssql flakiness in CI


gforsyth:
### What happened?

Just recording some notes and observations here:

Running `mssql` in docker and running all tests with `--count 20`

```python
ibis/backends/tests/test_export.py::test_table_to_parquet[mssql-16-20] FAILED [ 12%]
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

tmp_path = PosixPath('/tmp/nix-shell.BiFH3I/pytest-of-gil/pytest-8/test_table_to_parquet_mssql_160')
backend = <ibis.backends.mssql.tests.conftest.TestConf object at 0x7f036b8f9d50>
awards_players = DatabaseTable: awards_players
  playerID string
  awardID  string
  yearID   int64
  lgID     string
  tie      string
  notes    string

    @pytest.mark.notimpl(["dask", "impala", "pyspark", "druid"])
    def test_table_to_parquet(tmp_path, backend, awards_players):
        outparquet = tmp_path / "out.parquet"
        awards_players.to_parquet(outparquet)
    
        df = pd.read_parquet(outparquet)
    
>       backend.assert_frame_equal(awards_players.execute(), df)

ibis/backends/tests/test_export.py:225: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'ibis.backends.mssql.tests.conftest.TestConf'>
left =        playerID                awardID  yearID lgID   tie notes
0      bondto01  Pitching Triple Crown    1877   NL  N...Glove    2015   AL  None    CF
6037  cespeyo01             Gold Glove    2015   AL  None    LF

[6038 rows x 6 columns]
right =        playerID                              awardID  yearID lgID   tie notes
0      bondto01                Pitching ...  ML  None  None
6077  perezsa02                     World Series MVP    2015   ML  None  None

[6078 rows x 6 columns]
args = (), kwargs = {'check_dtype': False}

    @classmethod
    def assert_frame_equal(
        cls, left: pd.DataFrame, right: pd.DataFrame, *args: Any, **kwargs: Any
    ) -> None:
        left = left.reset_index(drop=True)
        right = right.reset_index(drop=True)
[...]

PR #6509, opened 2023-06-22
test(mssql): add explicit close for record batch reader

gforsyth:
This has (I think) fixed the flaky `mssql` tests on my machine -- I don't think this is a robust solution to the problem, but I'm partially pushing it up to see if CI runs without issue.

xref #6503


gforsyth:
I've manually re-run the mssql 3.8 job, now running the mssql 3.11 job.  if these both pass, I think this is probably _some_ kind of fix.

cpcloud (approved):
LGTM!

PR #6513, opened 2023-06-22
fix(mssql): dont yield from inside a cursor

gforsyth:
Continuing in the efforts against this heisenbug.

Previous fix didn't do the trick.  xref #6503 #6509

We noticed two failure modes:
1. the LHS is wrong (expr.execute())
2. the RHS is wrong (wrote the wrong number of rows to a file)

both of these point, at their core, to some issue with the cursor, since the mssql backend relies on a SQLAlchemy cursor for feeding both `execute` and `to_pyarrow_batches` (which itself feeds `to_parquet` and `to_csv`).

We found this:
http://www.pymssql.org/en/stable/pymssql_examples.html#important-note-about-cursors

which lets us know that we _cannot_ have multiple cursors on the same connection.

Attempting to close the cursor after the `yield` loop doesn't work.

So instead, for now, this terrible workaround.
We fetch the entire contents, then re-batch it using `toolz`.

We should noodle on this some more, but this is one possible "fix"

### DIFF
--- a/ibis/backends/mssql/__init__.py
+++ b/ibis/backends/mssql/__init__.py
@@ -2,9 +2,10 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Any, Iterable, Literal, Mapping
 
 import sqlalchemy as sa
+import toolz
 
 from ibis.backends.base.sql.alchemy import BaseAlchemyBackend
 from ibis.backends.mssql.compiler import MsSqlCompiler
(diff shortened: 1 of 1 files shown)


cpcloud (approved):

cpcloud:
Having explored this from many angles, I'm not sure there's an obvious other solution so I'm going to merge this PR.

Here's what we know:

1. Being more judicious about closing cursors *and* RecordBatchReaders (which we should probably do anyway) does not address the failure.
2. Calling `gc.collect()` at the end of `_cursor_batches` does seem to mitigate the problem with large performance hit.
3. The problem is not reproducible unless a RecordBatchReader is in the mix.

PR #6518, opened 2023-06-23
feat(io): add `to_torch` API

cpcloud:
This PR adds a `to_torch` method that returns a dict of column name to torch tensor. This came up during experimentation with DuckDB pyarrow UDFs where I wanted to pass some training data to a model. I am opening this PR to discuss the API, and am not strongly opinionated about whether this should be merged or not. The API copies what DuckDB does and uses its implementation in the DuckDB backend.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -32,6 +32,7 @@ if TYPE_CHECKING:
 
     import pandas as pd
     import pyarrow as pa
+    import torch
 
 __all__ = ('BaseBackend', 'Database', 'connect')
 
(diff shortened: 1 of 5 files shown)


lostmygithubaccount:
would using torcharrow help at all? https://github.com/pytorch/torcharrow

cpcloud:
> would using torcharrow help at all? [pytorch/torcharrow](https://github.com/pytorch/torcharrow)

Possibly, though the dependency situation looks a bit rough at the moment, they only support nightly wheels and only for Pythons 3.7-3.9.

ibis/expr/types/core.py
```diff
@@ -493,6 +493,34 @@ def to_delta(
...
         self._find_backend(use_default=True).to_delta(self, path, **kwargs)
 
+    @experimental
+    def to_torch(
+        self,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        **kwargs: Any,
+    ) -> None:
```
gforsyth, line 503:
```suggestion
    ) -> dict[str, torch.Tensor]:
```

although if we don't want to have to install `torch` just for the type hint I get it

ibis/backends/duckdb/__init__.py
```diff
@@ -732,6 +731,37 @@ def to_pyarrow(
...
 
+    @util.experimental
+    def to_torch(
+        self,
+        expr: ir.Expr,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        **kwargs: Any,
+    ) -> dict[str, torch.array]:
```
gforsyth, line 742:
```suggestion
    ) -> dict[str, torch.Tensor]:
```

even if we wish it were true...

ibis/backends/base/__init__.py
```diff
@@ -334,6 +335,44 @@ def to_pyarrow_batches(
...
 
+    @util.experimental
+    def to_torch(
+        self,
+        expr: ir.Expr,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        limit: int | str | None = None,
+        **kwargs: Any,
+    ) -> dict[str, torch.array]:
```
gforsyth, line 346:
```suggestion
    ) -> dict[str, torch.Tensor]:
```

gforsyth (changes requested):
just some nits on the type hints.

As for whether `dict[str, torch.Tensor]` is the best output format, I think we can copy DuckDB for the moment and always change things if there's a strong sentiment from users.

ibis/expr/types/core.py
cpcloud: I think it should be fine to just leave it in. You won't get tab completion if you don't have it installed but it shouldn't break anything to use the annotation.

ibis/backends/duckdb/__init__.py
cpcloud: 😂

gforsyth (approved):
:shipit:

PR #6519, opened 2023-06-23
refactor: replace custom `_merge` using `pd.merge`

deepyaman:
As I was replicating logic in `test_join.py` for manual testing of a new backend, I couldn't figure out how the result would be different than vanilla `pd.merge`, because the replacement conditions should never trigger for the respective joins. Tests seem to pass after simplification.


cpcloud (approved):
Nice find! Thanks for the PR!

PR #6527, opened 2023-06-26
refactor(imports): move old UDF implementation to `ibis.legacy.udf` module

cpcloud:
This PR moves existing UDF infrastructure to a new `ibis.elgacy` module with no other API changes in order to allow `from ibis import udf` to avoid clashing with `import ibis.udf` which no longer works after this PR.


gforsyth (approved):
This looks good to me, just the conflict in `test_aggregation.py` -- I did a check through the docs to make sure we weren't missing anything, but all mentions of UDF are properly handled here.

PR #6529, opened 2023-06-26
feat(datafusion): add support for pyarrow UDFs

cpcloud:
This PR adds support for pyarrow UDFs to datafusion using the new UDF API.


gforsyth (approved):
:shipit:

ibis/backends/datafusion/compiler.py
```diff
@@ -467,6 +468,23 @@ def elementwise_udf(op):
     return udf(*args)
 
 
+@translate.register(ops.ScalarUDF)
+def scalar_udf(op):
+    if (input_type := op.__input_type__) != InputType.PYARROW:
```
cpcloud, line 473: At some point we may want encode the input types as separate classes to make translation code a bit cleaner.

PR #5336, opened 2023-01-26
feat(api): add builtin support for relabeling columns to snake case

jcrist:
This adds support for passing in the literal string `"snake_case"` to `Table.relabel`, for relabeling column names to use a `snake_case` convention. While this is doable using the existing APIs, it requires manually writing a helper utility to pass to `Table.relabel` to handle the relabeling. I've found this to be a common operation when working with weird/messy data.

The utility currently handles renaming:

- `"Spaced Out Strings" -> "spaced_out_strings"`
- `"camelCase" -> "camel_case"`
- `"PascalCase" -> "pascal_case"`
- `"kebab-case" -> "kebab_case"`

I think this should be resilient enough to handle most naming conventions. Anything messier the user can always fallback to manually working with `table.relabel`.

PR #6532, opened 2023-06-26
feat(api): relabel columns to ALL_CAPS

IndexSeek:
Using "snake_case" has been a useful short-hand as was implemented in https://github.com/ibis-project/ibis/pull/5336/. In addition to "snake_case" I wanted to add support for "screaming_snake_case". Some systems, such as Snowflake, Oracle SQL, and PL/SQL follow this sort of convention. 

I think this would be a worthwhile addition that might help when working with different backends or databases that treat double-quoted identifiers as case-insensitive. 

https://en.wikipedia.org/wiki/Snake_case

https://docs.snowflake.com/en/sql-reference/identifiers-syntax#migrating-from-databases-that-treat-double-quoted-identifiers-as-case-insensitive

### DIFF
--- a/docs/blog/rendered/model.py
+++ b/docs/blog/rendered/model.py
@@ -1,4 +1,4 @@
-"""Linear regiression model for predicting cab fares using PyTorch.
+"""Linear regression model for predicting cab fares using PyTorch.
 
 Adapted from https://gist.github.com/pdet/e8d38734232c08e6c15aba79b4eb8368#file-taxi_prediction_example-py.
 """
--- a/docs/blog/rendered/torch.ipynb
+++ b/docs/blog/rendered/torch.ipynb
@@ -50,7 +50,7 @@
     "- Remove negative trip distances and negative fare amounts\n",
     "- Cast inputs to `float32`, since that's what PyTorch wants\n",
     "\n",
-    "We use a function here ito ensure that we can run the same code on the test data set before inference."
+    "We use a function here to ensure that we can run the same code on the test data set before prediction."
    ]
   },
   {
@@ -239,7 +239,7 @@
    "id": "817838db-661e-449a-a5a9-0dceb332a18f",
    "metadata": {},
    "source": [
-    "## Define an Ibis UDF that Runs Inference\n",
+    "## Define an Ibis UDF that predicts fares\n",
     "\n",
     "Now we get to the meaty part: defining an Ibis UDF (user-defined function) that invokes our model on unseen data!"
    ]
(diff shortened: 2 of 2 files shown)


lostmygithubaccount:
should it be `"SCREAMING_SNAKE_CASE"`?

IndexSeek:
> should it be `"SCREAMING_SNAKE_CASE"`? 

I thought about this, as that makes sense, as it's usually how it's stylized. I'm happy to make that tweak if that's how it would be preferred to be referenced. 

It looks like it could also be referred to as the following:
- ALL_CAPS
- MACRO_CASE
- CONSTANT_CASE

But in each case, the styling is all caps, as you mention.

cpcloud:
@IndexSeek Nice, thanks for the PR.

Can we go with `ALL_CAPS`? `SCREAMING_SNAKE_CASE` is kind of a handful to type and ideally we can keep these shorthands ... short 😂.

IndexSeek:
> @IndexSeek Nice, thanks for the PR.
> 
> Can we go with `ALL_CAPS`? `SCREAMING_SNAKE_CASE` is kind of a handful to type and ideally we can keep these shorthands ... short 😂.

Definitely! I have made that change now and agree, short and simple. I thought the original "snake_case" short-hand was really handy; I'm hoping this version of it will also be convenient, especially for going between backends.

cpcloud:
@IndexSeek You may want to disable GitHub Actions on your fork. I believe that will cause a bit of CI congestion for you since you'll be running every job twice: once against the main ibis repo and once against your fork.

cpcloud:
@IndexSeek I think there may be a whitespace change picked up by the markdown autoformatter. I'll fix it up.

IndexSeek:
> @IndexSeek You may want to disable GitHub Actions on your fork. I believe that will cause a bit of CI congestion for you since you'll be running every job twice: once against the main ibis repo and once against your fork.

This is a good call, disabled them now, that'll help keep my email inbox a little cleaner as well. 

> @IndexSeek I think there may be a whitespace change picked up by the markdown autoformatter. I'll fix it up.
> Hm, looks like I can't push to the branch because it's `master` (I can only push to non-`master` branches).

I've made the whitespace fix now, thank you for looking at that and attempting to fix it. I should have just made a feature branch on my fork. Hopefully it is good now, checks are still queued as I'm commenting.

ibis/expr/types/relations.py
```diff
@@ -1656,7 +1656,8 @@ def relabel(
         substitutions: Mapping[str, str]
         | Callable[[str], str | None]
         | str
-        | Literal["snake_case"],
+        | Literal["snake_case"]
+        | Literal["ALL_CAPS"],
```
jcrist, line 1660:
```suggestion
        | Literal["snake_case", "ALL_CAPS"],
```

jcrist (commented):
One small type annotation nit, otherwise this looks good to me!

Could you please squash your commits together into a single commit (with a message something like "feat(api): add ALL_CAPS option to `Table.relabel`"). We generate our changelog from commit messages, so we're a bit picky here about this.

IndexSeek:
I've opened https://github.com/ibis-project/ibis/pull/6543 to squash those commits and try this again. I truly appreciate all of your patience and support. I understand wanting to keep things simple with the CHANGELOG. 

As I'm commenting on this, I realize I must combine the Literal typing as was suggested above.

PR #6534, opened 2023-06-27
ci(examples): address flakiness in `test_examples`

### LINKED ISSUE #6533: ci(examples): occasional test flakiness in `test_examples`
### What happened?

The test (`test_examples`) that downloads examples and runs a simple computation with them is flaky. See https://github.com/ibis-project/ibis/actions/runs/5358583495/jobs/9721149724?pr=6521 for a run that failed because of (I think) the penguins CSV file being in use by multiple processes at the same time which Windows doesn't allow.

`master`

DuckDB

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
This PR addresses recentish test flakiness with examples that are used in both doctests and the `test_examples` test. The solution here is to avoid testing the shared tables in both doctests and `test_examples`. This may seem fragile, but in practice the number of unique example tables used in doctests does not vary much at all. Fixes #6533.

### DIFF
--- a/ibis/backends/conftest.py
+++ b/ibis/backends/conftest.py
@@ -30,6 +30,7 @@ SANDBOXED = (
 LINUX = platform.system() == "Linux"
 MACOS = platform.system() == "Darwin"
 WINDOWS = platform.system() == "Windows"
+CI = os.environ.get("CI") is not None
 
 TEST_TABLES = {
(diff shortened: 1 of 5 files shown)


gforsyth (approved):
great -- this was an annoying failure

PR #6539, opened 2023-06-27
fix: support multipledispatch version 1

tswast:
multipledispatch version 1 was released with a few cleanups based on 0.6


cpcloud:
Wow, I didn't realize `multipledispatch` was still under development!

cpcloud:
It seems like you may have run `poetry update` which will update all dependencies (latest within the current bounds) and I think you may want to run the following instead to only update the `multipledispatch` dependency:

```
poetry update multipledispatch
just lock
```

Also make sure you're running poetry 1.3.x!

cpcloud (approved):
Fixed up the lockfile! Thanks for doing this!

PR #6541, opened 2023-06-27
fix(examples): move lockfile creation to data directory

cpcloud:
This PR makes sure that examples lock files are created next to the data instead of in `$PWD`.

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -378,6 +378,10 @@ jobs:
           IBIS_TEST_WEBHDFS_USER: hdfs
           IBIS_EXAMPLES_DATA: ${{ runner.temp }}/examples-${{ matrix.backend.name }}-${{ matrix.os }}-${{ steps.install_python.outputs.python-version }}
 
+      - name: check that no untracked files were produced
+        shell: bash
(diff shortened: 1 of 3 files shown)


jcrist (approved):

.github/workflows/ibis-backends.yml
```diff
@@ -760,6 +772,10 @@ jobs:
       - name: run tests
         run: just ci-check -m ${{ matrix.backend.name }} --numprocesses auto --dist=loadgroup
 
+      - name: check that no untracked files were produced
+        shell: bash
+        run: git checkout poetry.lock pyproject.toml && ! git status --porcelain | tee /dev/stderr | grep .
+
```
gforsyth, line 778: this seems like the conclusion to a frustrating story

gforsyth (approved):
:shipit:

.github/workflows/ibis-backends.yml
cpcloud: Yeah ... ideally we don't have to modify the pyproject.toml with any `poetry add` commands, but we're bound to pyarrow 10 in the lock file because of `$REASONS`.

cpcloud: I add the latest pyarrow so that we can test against it. Once snowflake-connector-python moves to nanoarrow at least the `git checkout` bits can go away.

PR #6543, opened 2023-06-27
feat(api): add ALL_CAPS option to Table.relabel

IndexSeek:
The intent of the change is to add another shorthand in addition to "snake_case" for relabeling. I wanted to add support for "SCREAMING_SNAKE_CASE". Some systems, such as Snowflake, Oracle SQL, and PL/SQL, follow this convention. This would be a worthwhile addition that might help with different backends or databases that treat double-quoted identifiers as case-insensitive.

https://github.com/ibis-project/ibis/pull/6532 was the original PR for this, but it was closed upon resyncing the fork and discarding changes. After the soft reset and force push of accidentally going back too far, I branched off the fork and redid the changes for ALL_CAPS as the table relabeling option on a separate branch.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -1656,7 +1656,7 @@ class Table(Expr, _FixedTextJupyterMixin):
         substitutions: Mapping[str, str]
         | Callable[[str], str | None]
         | str
-        | Literal["snake_case"],
+        | Literal["snake_case", "ALL_CAPS"],
     ) -> Table:
         """Rename columns in the table.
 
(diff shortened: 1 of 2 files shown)


ibis/expr/types/relations.py
```diff
@@ -1656,7 +1656,8 @@ def relabel(
         substitutions: Mapping[str, str]
         | Callable[[str], str | None]
         | str
-        | Literal["snake_case"],
+        | Literal["snake_case"]
+        | Literal["ALL_CAPS"],
```
jcrist, line 1660:
```suggestion
        | Literal["snake_case", "ALL_CAPS"],
```

jcrist (commented):
Just a small nit, otherwise LGTM!

ibis/expr/types/relations.py
IndexSeek: Thank you! Sorry, I forgot to resolve this in the midst of the squashing. I just amended the commit with this change so hopefully we're set now!

jcrist (approved):
Assuming tests pass, this LGTM! Thanks for your work here :).

IndexSeek:
You're welcome! It's a pleasure to help. I appreciate your cooperation and patience too!

Issue #6492, opened 2023-06-20
feat: use try_cast() in duckdb-engine


NickCrews:
### Is your feature request related to a problem?
EDIT: the PR has now landed in duckdb-engine

try_cast() is getting added to duckdb-engine in https://github.com/Mause/duckdb_engine/pull/696. Once that lands, I think we can redo our implementation of try_cast to just use that instead of re-inventing it ourselves.

### Describe the solution you'd like

NA

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

NickCrews:
We _could_ backport the support for people with older versions of duckdb-engine:

```python
if duckdb_engine.__version__ < "0.9.0":

    @compiles(try_cast, "duckdb")
    def compiles_try_cast(element, compiler, **kw):
        return "TRY_CAST({} AS {})".format(
            compiler.process(element.clauses.clauses[0], **kw),
            compiler.visit_typeclause(element),
        )
```

but I think I would rather just delete that code, and if people want the support then they should update to duckdb-engine>=0.9.0

Thoughts?

lostmygithubaccount:
given how recently the `try_cast` operation in Ibis landed, I think it's reasonable to only support the new DuckDB functionality and require users to upgrade

cpcloud:
Closing this out.

Since we have to support sqlalchemy 1.4.x and it would be strictly additional code to use the duckdb-engine implementation because we'd also have to add a version check.

NickCrews:
Sure. Can revisit once we can drop old versions. Should we add a TODO in the code that says this?

PR #6545, opened 2023-06-27
Use try_cast() from sqlalchemy and duckdb-engine

NickCrews:
Fixes https://github.com/ibis-project/ibis/issues/6492.

I'm not exactly sure how all these pieces fit together, feel free to take this over as needed.

I don't think this will work quite yet, because we need [sqlalchemy>=2.0.14](https://github.com/sqlalchemy/sqlalchemy/releases/tag/rel_2_0_14), which isn't in requirements.txt yet.

But once that is upgraded, then this should be good to merge?

requirements.txt already has the new version of duckdb-engine>=0.9.0 that we need.

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -22,10 +22,6 @@ class substr(GenericFunction):
     inherit_cache = True
 
 
-class try_cast(GenericFunction):
-    pass
-
-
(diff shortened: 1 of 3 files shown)


NickCrews:
Ahh, I just assumed that we were being a little slow in our update of requirements.txt. that makes sense that we need to continue support of sqlalchemy 1 though.

PR #6548, opened 2023-06-28
refactor(ci): roll flink into the existing backend ci matrix

cpcloud:
This PR removes duplicate CI code related to Flink.


cpcloud:
Doesn't seem like we can do this until we add a proper backend test configuration.

PR #6553, opened 2023-06-28
feat(cast): add table-level try_cast

gforsyth:
I ran into this the other day and it seems like a nice convenience to also have a table-level `try_cast` -- I'm happy to add it as a separate method (which is probably better for consistency with the column methods).

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -225,6 +225,43 @@ class Table(Expr, _FixedTextJupyterMixin):
             ...
         ibis.common.exceptions.IbisError: Cast schema has fields that are not in the table: ['foo']
         """
+        return self._cast(schema, cast_method="cast")
+
+    def try_cast(self, schema: SupportsSchema) -> Table:
+        """Cast the columns of a table.
+
+        If the cast fails for a row, the value is returned
+        as `NULL` or `NaN` depending on backend behavior.
+
+        Parameters
+        ----------
+        schema
+            Mapping, schema or iterable of pairs to use for casting
+
+        Returns
+        -------
+        Table
+            Casted table
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.memtable({"a": ["1", "2", "3"], "b": ["2.2", "3.3", "book"]})
+        >>> t.try_cast({"a": "int", "b": "float"})
+        ┏━━━━━━━┳━━━━━━━━━┓
+        ┃ a     ┃ b       ┃
+        ┡━━━━━━━╇━━━━━━━━━┩
+        │ int64 │ float64 │
+        ├───────┼─────────┤
+        │     1 │     2.2 │
+        │     2 │     3.3 │
+        │     3 │     nan │
+        └───────┴─────────┘
+        """
+        return self._cast(schema, cast_method="try_cast")
+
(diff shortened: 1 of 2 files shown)


jcrist:
>  I'm happy to add it as a separate method (which is probably better for consistency with the column methods).

Agreed, I think parity between method name for columns/tables here is a good thing to strive for here.

lostmygithubaccount:
I do think separate method would be better for consistency (or alternatively changing the existing implementation to use a strict flag)

gforsyth:
added a separate method

ibis/expr/types/relations.py
```diff
@@ -225,6 +225,26 @@ def cast(self, schema: SupportsSchema) -> Table:
             ...
         ibis.common.exceptions.IbisError: Cast schema has fields that are not in the table: ['foo']
         """
+        return self._cast(schema, strict=True)
+
+    def try_cast(self, schema: SupportsSchema, strict=True) -> Table:
+        """Cast the columns of a table.
+        If the cast fails for a row, the value is returned
```
jcrist, line 232: I think for proper formatting you'll want a newline here.

```suggestion

        If the cast fails for a row, the value is returned
```

ibis/expr/types/relations.py
```diff
@@ -225,6 +225,26 @@ def cast(self, schema: SupportsSchema) -> Table:
...
+
+        Parameters
+        ----------
+        schema
+            Mapping, schema or iterable of pairs to use for casting
+
+        Returns
+        -------
+        Table
+            Casted table
```
jcrist, line 243: It'd be nice to also have an example too if it's easy enough to add.

ibis/expr/types/relations.py
```diff
@@ -225,6 +225,26 @@ def cast(self, schema: SupportsSchema) -> Table:
...
+        """
+        return self._cast(schema, strict=False)
+
+    def _cast(self, schema: SupportsSchema, strict=True) -> Table:
```
jcrist, line 247: Could also do something like

```suggestion
    def _cast(self, schema: SupportsSchema, cast_method: str = "cast") -> Table:
```

to avoid boolean logic below. Either works though.

jcrist (commented):
Overall LGTM, one small nit on the docstring, the rest are ignorable suggestions.

ibis/expr/types/relations.py
gforsyth: that seems cleaner

ibis/expr/types/relations.py
gforsyth: done

ibis/expr/types/relations.py
gforsyth: done

ibis/expr/types/relations.py
```diff
@@ -225,6 +225,43 @@ def cast(self, schema: SupportsSchema) -> Table:
             ...
         ibis.common.exceptions.IbisError: Cast schema has fields that are not in the table: ['foo']
         """
+        return self._cast(schema, cast_method="cast")
+
+    def try_cast(self, schema: SupportsSchema, strict=True) -> Table:
```
cpcloud, line 230: Is `strict` used anywhere?

gforsyth: It was not.  thanks for flagging!

cpcloud (approved):
Nice!

PR #6555, opened 2023-06-29
fix(mysql): ensure that floats and double do not come back as Python Decimal objects

cpcloud:
This PR addresses the recent test flakiness seen in the MySQL backend.

The manifestation of the flakiness (dropped connection) was a bit of a red
herring.

The dropped connection was ultimately due to a connection that was
indeterminately closed, usually at some point later during interpreter GC but
not always (leading to the flakiness) because the batch producer for the
`RecordBatchReader` uses a connection in a context manager and it was left open
because of the **failure during the attempt by pyarrow to cast Python
`decimal.Decimal` objects to `float`s**.

The cast itself was the result of SQLAlchemy automatically returning
`decimal.Decimal` objects by default for columns whose type was
`mysql.DOUBLE`/`mysql.FLOAT`. The solution was to prevent that automatic
conversion during table column reflection.

Still, this PR **does not** address potential the footguns of using
`to_pyarrow_batches`, which may only be possible to address through better
documentation, but that is for another PR.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -274,9 +274,10 @@ class _FileIOHandler:
             # Can't construct an array from record batches
             # so construct at one column table (if applicable)
             # then return the column _from_ the table
-            table = pa.Table.from_batches(
-                self.to_pyarrow_batches(expr, params=params, limit=limit, **kwargs)
-            )
+            with self.to_pyarrow_batches(
+                expr, params=params, limit=limit, **kwargs
+            ) as reader:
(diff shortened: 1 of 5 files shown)


gforsyth:
Looks like `test_column_to_pyarrow_batches[pandas-nolimit]` is now an `XPASS`!

cpcloud:
The druid changes here are not necessary this PR, but I made them because I inadvertently removed the `awardID` expression from the `test_column_to_pyarrow_batches` test.

I'll move them to a separate PR.

PR #6559, opened 2023-06-30
ci: remove unused test report workflow

cpcloud:
It seems this is currently failing and also not causing a drop in developer QoL, so let's remove it.


gforsyth (approved):
:shipit:

PR #6561, opened 2023-06-30
docs: fix some redirects

lostmygithubaccount:
fixing some URLs in the README.md I noticed I had changed in the recent docs PR (these redirects were working, but may as well adjust them)

then adding redirects for all the `/api/` docs to `/reference/` as these broke


jcrist (approved):
:shipit:

PR #6563, opened 2023-06-30
feat: make output of `to_sql` a proper `str` subclass

jcrist:
This makes it easier to immediately pass the output to some other tool for execution.

### DIFF
--- a/ibis/expr/sql.py
+++ b/ibis/expr/sql.py
@@ -317,22 +317,14 @@ def show_sql(
     print(to_sql(expr, dialect=dialect), file=file)
 
 
-class SQLString:
+class SQLString(str):
     """Object to hold a formatted SQL string.
 
     Syntax highlights in Jupyter notebooks.
     """
 
-    __slots__ = ("sql",)
-
-    def __init__(self, sql: str) -> None:
-        self.sql = sql
-
     def __repr__(self) -> str:
-        return f"{self.__class__.__name__}(sql={self.sql!r})"
-
-    def __str__(self) -> str:
-        return self.sql
+        return f"{self.__class__.__name__}({str(self)!r})"
 
     def _repr_markdown_(self) -> str:
         return f"```sql\n{self!s}\n```"


cpcloud (commented):
Anything in particular that is made more difficult by the current implementation?

I'm not hugely opposed to this but not subclassing `str` was intentional to avoid having to deal with defining how these objects compose.

It's also convenient to think about `SQLString` as an arbitrary `str`. I guess I don't see as a huge issue to have to call `str` on this to get a `str` back as opposed to allowing arbitrary string operations.

jcrist:
As is you can't pass the output of `to_sql` to another tool or system as easily. A few possible use cases:

1. Using ibis to generate a SQL query, then executing against a DB api connection
2. Same, but something async like `asyncpg`.
3. Writing the SQL to a file. As is, you have to wrap it in a str.

```python
sql = ibis.to_sql(expr)

with open("query.sql", "w") as f:
    f.write(str(sql))  # the str here is annoying
```

I personally expected the output of `to_sql` to be a string, and was surprised to find it wasn't (it just looks like one for interactive use). I understand wanting to make it easier to disambiguate SQL strings from any str, but IMO the custom subclass does that already, and having it be a subclass of `str` removes a small user-facing rough edge.

lostmygithubaccount:
FWIW I ran into this writing the v6 blog. While not exactly a real user scenario, I would expect `to_sql()` to return a string for the reasons mentioned.

I was showing off the `.sql(sql_str, dialect=dialect)` functionality and using Ibis to get a SQL string to use. At first I tried with `ibis.show_sql()`, but that was NoneType. I found `ibis.to_sql()`, but trying to pass that into `con.sql()` directly gives `IbisTypeError: Given argument with type <class 'ibis.expr.sql.SQLString'> is not an instance of <class 'str'>`, so I realized I needed to cast to `str(ibis.to_sql(t))` myself. Not a huge deal, but it wasn't what I expected as a user.

PR #6585, opened 2023-07-05
fix(datafusion): always quote column names to prevent datafusion from normalizing case

cpcloud:
DataFusion normalizes unquoted identifiers to lowercase. This PR quotes all `TableColumn` references during compilation to avoid normalizing names that are case sensitive.


gforsyth (approved):
:shipit:

PR #6590, opened 2023-07-06
fix(mysql): fix UUID type reflection for sqlalchemy 2.0.18

jcrist:
The recent sqlalchemy 2.0.18 release added reflection support for UUID types for mysql, which broke a test for us.

This also turned up a bug in the type mapping refactor where the `super().to_ibis` call could never be called (which is intended to handle fallback cases like this). That fix should definitely go in.

The `mysql uuid <-> ibis str` type fix here I'm less sure about. Only mysql 2.0.18+ supports properly reflecting these types, and a bit of rudimentary testing seems to indicate that swapping to `mysql uuid <-> ibis uuid` may not fully work yet. I opted for the fix that keeps the same behavior for now.

### DIFF
--- a/ibis/backends/mysql/datatypes.py
+++ b/ibis/backends/mysql/datatypes.py
@@ -6,7 +6,7 @@ import sqlalchemy.types as sat
 from sqlalchemy.dialects import mysql
 
 import ibis.expr.datatypes as dt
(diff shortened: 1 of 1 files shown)


cpcloud (approved):
Let's keep the existing behavior for now. Thanks!

PR #6592, opened 2023-07-06
docs(social): add social media preview cards

gforsyth:
@marlenezw pointed out that links to the ibis docs don't render preview images on various social media platforms.

There's a mkdocs-material plugin to render those cards, so why not enable it?

Caveats:
1. `cairosvg` is currently broken in `poetry2nix` and will probably need an override, but I don't remember how to do that
2. The `social` plugin doesn't work with notebooks, so the preview cards aren't generated for any of the notebook pages (like the recent 6.0 release blog).

(I would be interested in figuring out a better notebook workflow in the docs because building them takes forever and the diffs are heinous).


cpcloud:
I can help with number 1.

Maybe @danielfrg can help with number 2?

cpcloud:
Upstream poetry2nix fixes are PR'd: https://github.com/nix-community/poetry2nix/pull/1204

cpcloud (approved):
LGTM!

PR #6596, opened 2023-07-06
fix(postgres): by default use domain socket

saulpw:
postgres:///dbname tries to connect via localhost, which may not be
  allowed.  With localhost as the default parameter, there is no way to
  force it to use the local unix domain socket.  This would make the
  default using the domain socket, and then connecting via localhost
  can be used via "postgres://localhost/dbname".

### DIFF
--- a/ibis/backends/postgres/__init__.py
+++ b/ibis/backends/postgres/__init__.py
@@ -34,7 +34,7 @@ class Backend(BaseAlchemyBackend, AlchemyCanCreateSchema):
 
     def do_connect(
         self,
-        host: str = 'localhost',
+        host: str | None = None,
         user: str | None = None,
         password: str | None = None,
         port: int = 5432,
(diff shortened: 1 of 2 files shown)


cpcloud:
I understand this change seems simple and fixes the issue of connecting to domain sockets seemingly "just work", but I think we can do this without breaking existing workflows that assume `localhost`.

Instead of changing the default for `do_connect`, `postgres:///dbname` should be passed through with a lot less keyword argument munging than we're doing now, which would allow the pass-through of no `host` (and therefore no ibis-defined default).

cpcloud:
Finally got back to poking around the bits here, and I think this PR is the way forward. I think we can probably avoid breaking the `ibis.postgres.connect` case here, while changing the existing behavior of `postgres:///dbname` to match what sqlalchemy would do which is the bug fix here.

I'll push up that change as well as a tiny test to verify the hostname after constructing a backend (but before connecting, to avoid having to spin up a another container).

cpcloud (approved):
Thank you @saulpw 

I added a test to make sure we don't break this in the future :)

This PR makes sense to me now after poking around SQLAlchemy a bit.

Ideally in ibis we should avoid second-guessing the user on their input URL, which we were doing before. With this change we're no longer doing so, by matching the default behavior of SQLAlchemy itself.

Thanks again for your patience!

PR #6597, opened 2023-07-07
feat(flink): implement translation rules and tests for over aggregation operations in Flink backend

chloeh13q:
This PR implements translation rules and tests for over aggregation operations in Apache Flink backend.

What Flink calls "over aggregation" follows similar syntax as windowing in basic SQL. There are, however, a few differences, for example:
1. Since tables do not have an inherent order, the `ORDER BY` clause is mandatory
2. For streaming queries, Flink currently only supports OVER windows that are defined with an ascending order
3. Flink only supports `CURRENT ROW` as the upper boundary

etc.

See [Flink's documentation on over aggregation](https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/dev/table/sql/queries/over-agg/) for additional details.

Here I attempted to add as many guardrails as possible to catch invalid expressions in the translation to avoid error in Flink runtime. 

There are a lot of subtleties in the "over aggregation" operator in Flink. As time attributes are a crucial hallmark of Flink & the streaming world, we're going to need more complex logic for expressing & representing time semantics. Over time, we want to build up better support for time-aware operators and expressions like `.over()`, `.window()`, `ibis.interval()` in Ibis.


ibis/backends/flink/utils.py
```diff
@@ -206,3 +210,78 @@ def format_as_string(self) -> str:
...
+    if len(nonzero_interval_segments) == 1:
+        unit = next(iter(nonzero_interval_segments))
+        value = nonzero_interval_segments[unit]
+        return f"'{value}' {unit.value}{format_precision(value, unit)}"
+
+    # YEAR TO MONTH, DAY TO SECOND
+    return interval.format_as_string()
+
+
+def literal(op: ops.Literal) -> str:
```
jcrist, line 249: Can you rename this to something like `translate_literal`? The bare `literal` is easy to confuse with `ibis.literal`.

jcrist (commented):
Apologies for the delayed review here. One small naming nit, otherwise looks good to me!

ibis/backends/flink/utils.py
chloeh13q: Done!

chloeh13q:
@jcrist fixed it!

jcrist (approved):

PR #6599, opened 2023-07-07
chore: allow pyupgrade to upgrade type annotations that were required by python 3.8

cpcloud:
This PR removes the remaining ruff pyupgrade constraints that we had due to supporting Python 3.8, which is no longer supported.


ibis/common/grounds.py
```diff
@@ -47,7 +48,14 @@ def __new__(metacls, clsname, bases, dct, **kwargs):
         # collection type annotations and convert them to validators
         module_name = dct.get('__module__')
         annotations = dct.get('__annotations__', {})
-        typehints = evaluate_annotations(annotations, module_name)
+
+        # pull in locals from the caller, purely for type annotation evaluation
+        #
+        # not great, but likely can't be avoided until
+        # https://peps.python.org/pep-0649/ is available
+        localns = sys._getframe(1).f_locals
```
kszucs, line 56: This wouldn't work for more nested function definitions.

I think we can restrict annotable to work only with globally defined classes for now. There are two test cases relying on local namespace in `ibis/common/tests/test_egraph.py` where we can pull out the classes to be module level definitions.

cpcloud: Removed, and updated the tests.

ibis/common/tests/test_egraph.py
```diff
@@ -146,70 +148,78 @@ def test_enode():
         node.args = (2, 3)
 
 
-def test_enode_roundtrip():
-    class MyNode(Concrete, Node):
-        a: int
-        b: int
-        c: str
+class MyENode(Concrete, Node):
```
kszucs, line 151: This is actually a `Node` not an `ENode` :)

kszucs: Updated.

kszucs (approved):

PR #6609, opened 2023-07-08
fix(snowflake): use `regexp_instr > 0` instead of `REGEXP` keyword

cpcloud:
Fixes regexp search for snowflake.


cpcloud:
Snowflake string tests look good:

```
cloud in 🌐 albatross in …/ibis on  fix-snowflake-regex-search is 📦 v6.0.0 via 🐍 v3.10.12 via ❄️  impure (ibis-3.10.12-env) took 3s
❯ pytest -m snowflake ibis/backends/tests/test_string.py -n auto -q
bringing up nodes...
..x.....x.................................x......................x.............                                [100%]
75 passed, 4 xfailed in 42.83s
```

PR #6615, opened 2023-07-10
feat(pyspark): to_csv and to_parquet

lostmygithubaccount:
implements `to_csv` and `to_parquet` for the pyspark backend

example: https://gist.github.com/lostmygithubaccount/2d2ab5f70e8afda643c600e98c3989a8

### DIFF
--- a/ibis/backends/pyspark/__init__.py
+++ b/ibis/backends/pyspark/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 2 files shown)


lostmygithubaccount:
seems weird that only those two tests are failing -- issue is I think pandas trying to read the CSV file, which is a directory. the actual CSV file will be in the directory with the name `part-GUID.csv`

not sure what the best option here is -- in the test case, perhaps read the CSV back in with the PySpark connection instead of using pandas?

jcrist:
Hmmm, I remember us talking about this before (sorry, this issue had slipped my mind when we talked about this last week).

Currently all our `to_csv`/`to_parquet` writers produce a single file (rather than a directory of files). There was an open question as to what the expected behavior was for backends where a single-file output is tricky/inefficient/impossible (backends like spark or dask).

Enumerating all the options I can think of:

### 1. `to_csv`/`to_parquet` always outputs a single file

We could fallback to pyarrow or error for backends where this is tricky. This is the current behavior.

### 2. `to_csv`/`to_parquet` always output a directory

Backends that only write a single file would convert to writing a directory with a single file. This would be a breaking change, but should work with any backend's native csv/parquet writer. This would make a common case of writing a small csv of results trickier though.

### 3. `to_csv`/`to_parquet` write the backend-native style

The output style (single file or directory) depends on the backend. This is easy to implement, but would mean different backends would result in different behaviors.

### 4. `to_csv`/`to_parquet` have an option to write a directory

These methods would write to a single file by default, but would have an option to instead write to a directory. :shrug: whether this is a `directory=True` option, or inferred somehow from the input path. This would mean that users using backends that can only efficiently write to directories may need to opt into this support, but would make things explicit about the output type.

```python
t.to_csv("test.csv")  # would error for the spark backend? Or fallback to pyarrow?
t.to_csv("test/", directory=True)  # output a directory can use pyspark native behavior
```

### 5. New methods for writing directories

`to_csv`/`to_parquet` keep their existing behavior, and we add new `to_csv_dir`/`to_parqet_dir` (or much better named) methods for writing directories. Same caveats/questions as 4.

Right now I'm leaning towards 4. An extra flag seems fine to me, and I like it better than adding another top-level-method per file format just to support partitioned writing. cc'ing @gforsyth for a 2nd opinion though, since he may remember what conclusions we came to last time this came up.

gforsyth:
Welcome back to my nightmares!

I agree with @jcrist that Option 4 is probably best.  

The extra wrinkle is that the various backends also have different behaviors for how they _read_ directory partitioned parquet files.  I have written up way too much about this here: https://gist.github.com/gforsyth/8dd4ca981b2beed6ef4db80f5e8afbfd

We punted on this the last time it came up because we want:
1) A user should be able to create the same (ish) parquet output no matter which backend 
2) Ibis should always be able to read in a parquet dataset that it had a hand in creating

One of the major bumps last time was that DuckDB didn't support recursive globs -- i think that may have changed in the interim, though.

lostmygithubaccount:
I don't have a super strong opinion but would expect 3: just do what the backend does. Having an extra parameter for option 4 seems fine to me too.

lostmygithubaccount:
I also find it weird that the pandas 2.X tests are passing -- need to double check later but looking at their docs it doesn't seem like reading from a directory should work.

jcrist:
> I also find it weird that the pandas 2.X tests are passing -- need to double check later but looking at their docs it doesn't seem like reading from a directory should work.

The pyspark tests are set to xfail all tests if pandas >= 2.0, since pyspark doesn't support pandas 2.

lostmygithubaccount:
closing as I don't intend to take this over the finish line -- some further work to implement one of the options above would be needed

if you need this feature, feel free to comment here or open a new issue. this PR does work. but also, you can just use Delta Lake tables!

PR #6634, opened 2023-07-14
fix(duckdb): avoid double escaping backslashes for bind parameters

cpcloud:
This PR fixes an issue with bind parameter escaping, where backslashes are doubly escaped which results in a minimum of two backslashes in any bind parameter that has them.


cpcloud:
There's more to the story here, this PR is only a fix for duckdb at the moment.

Double-escaping dialects:

1. postgres
2. duckdb (because it inherits from the postgres dialect)
3. mysql

Non-double-escaping dialects:

1. sqlite
2. the default dialect

cpcloud:
The examples are failing because I updated them (and the hash changed). PR to address that in #6635.

cpcloud:
I'll close this PR in favor of #6635 so that we don't have to merge broken code.

PR #6637, opened 2023-07-15
fix(oracle): add support for oracle LONG types

gforsyth:
The Oracle backend did not survive first contact with real data -- this is a hacky branch to try to resolve.

### DIFF
--- a/ibis/backends/base/sql/alchemy/datatypes.py
+++ b/ibis/backends/base/sql/alchemy/datatypes.py
@@ -304,6 +304,8 @@ class AlchemyType(TypeMapper):
         elif isinstance(typ, sa.DateTime):
             timezone = "UTC" if typ.timezone else None
             return dt.Timestamp(timezone, nullable=nullable)
+        elif isinstance(typ, sat.String):
+            return dt.String(nullable=nullable)
         elif geospatial_supported and isinstance(typ, ga.types._GISType):
             name = typ.geometry_type.upper()
             try:
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -210,6 +210,11 @@ def tmpcon(alchemy_con):
     return alchemy_con._from_url(alchemy_con.con.url)
 
 
+@mark.broken(
+    ["oracle"],
+    reason="oracle temp tables aren't cleaned up on reconnect -- they need to "
+    "be switched from using atexit to weakref.finalize",
+)
 @mark.notimpl(["trino", "druid"], reason="doesn't implement temporary tables")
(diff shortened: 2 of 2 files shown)


gforsyth:
We have a user at the SciPy sprints who was having errors with oracle table schema conversion to alchemy table -- this fix is a bit blunt but fixed an issue I was having loading a column of `LONG` type and also the new user who had table with `NVARCHAR2` types

ibis/backends/oracle/datatypes.py
```diff
@@ -32,6 +32,8 @@ def to_ibis(cls, typ, nullable=True):
                     scale=typ.scale or 0,
                     nullable=nullable,
                 )
+        elif isinstance(typ, sat.String):
+            return dt.String()
```
cpcloud, line 36: Is this not covered by the base class implementation?

gforsyth: I thought it would be, but apparently not?  I'll take a look at the base implementation to see what's going sideways.

gforsyth: Ahh, Jim noticed we aren't doing a subclass check, so `if dtype := _from_sqlalchemy_types.get(type(typ))` isn't catching things like the oracle `LONG` and `NVARCHAR2`.

I can move the `isinstance` check into the base alchemy file so we have it as a fallback for all the backends.

gforsyth:
Ok, I'm not sure what to do here.  

I think the fix is correct, in that we should be able to handle string types.

The test failures are very confusing.

It's failing because the temp tables in Oracle aren't cleaned up until we hit the `atexit` handler, which doesn't happen during a `reconnect`.
But it also isn't clear why the temp tables are showing up at all.  The `tmpcon.tables.keys()` is a `KeysView` which should be created from the `__iter__` of the class in question, so it should be the contents of `list_tables`.  That is what it `repr`s as, but `some_temp_table in tmpcon.tables.keys()` is `True`.

What I truly cannot understand is how changing the base alchemy backend to handle sqlalchemy string types is making this happen.  I can xfail the test, but I don't know why it's failing.

gforsyth:
Ok, tracked this down with help from @jcrist, @cpcloud, and @lostmygithubaccount.

The test was previously passing because we were failing to parse the oracle `CLOB` type, which meant that attempting to load the temp table would fail because we failed to parse the schema.  This failure was swallowed by our tmp table handler.

So, for now, adding in the improved string dtype parsing is working, which makes the temp table cleanup test fail, but that test was never truly passing.

jcrist (approved):
:+1:

cpcloud (approved):

PR #6639, opened 2023-07-15
docs: add conda/mamba install instructions for specific backends

jcrist:
This adds install instructions for installing specific backends with `conda` and `mamba`. Previously we only had backend install instructions for `pip`.

I tried putting all this under nested tabs, but I found it harder to read. Splitting `pip` and `conda`/`mamba` into separate sections felt cleaner IMO.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -7,26 +7,17 @@ hide:
 
 # Install Ibis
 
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
:shipit:

PR #6641, opened 2023-07-15
refactor(examples): replace pooch with lighter weight pins

cpcloud:
This PR replaces `pooch` with the lighter weight and less strict (more on that
below) `pins` library.

`pooch` is nice, but it's not clear how to use it in such a way where example
data can be updated without breaking the ability to read that data from older
versions of ibis, due to the strict hashing and checksumming.

`pins` doesn't perform checksums, which has a few implications:

1. **Older versions of ibis will be able to read updated data.**
1. There's added network overhead for fetching, because pins checks for new
   data each fetch. It's noticable if you're fetching the same thing over and
   over, but acceptable IMO.
1. There's no confirmation that what's downloaded matches anything in
   particular. I don't think this is a huge concern given that only the ibis
   core team and a few folks at Google have write access to the bucket.

I went with pins, since I think think the value we're getting from checksums is
less than that of old ibis versions being able to read updated example data.


gforsyth (approved):
I went to check to make sure that `pins` was on conda-forge and discovered that I added it a few weeks ago.  

This looks good to me.  I think the people most likely to repeatedly pull down example datasets are the core devs -- and I am a fan of not requiring an ibis point release just to unbreak an example dataset.

:shipit:

PR #6643, opened 2023-07-17
fix: forward arguments through `__dataframe__` protocol

jcrist:
The `__dataframe__` method takes a few keyword arguments (see [here](https://data-apis.org/dataframe-protocol/latest/API.html)). Rather than list them out, I went with a `*args, **kwargs` approach in case the protocol is updated later.


gforsyth (approved):
:shipit:

PR #6645, opened 2023-07-17
fix(duckdb): run pre-execute-hooks in duckdb before file export

### LINKED ISSUE #6091: bug: to_parquet() writes incorrect data without previous .cache()
### What happened?

OK, I don't have a nice repro for this because I'm not sure exactly what is causing it, and it's too complicated for me to share it all. If you could help me figure out where exactly to look then I gladly will.

I am loading some data, doing a bunch of processing, and then writing `.to_parquet()`. It writes fine, but several columns are always null, when they should have contents. If I call .cache() before writing, then they write with the correct contents.

Here is the output of `ibis.show_sql(table)` just before the `.to_parquet()` if that is helpful: [log.sql](https://github.com/ibis-project/ibis/files/11327266/log.txt)

When I do this, I get these warnings, if you think that could be related. Using `sqlglot==11.5.7`
```
WARNING:sqlglot:REGEXP_EXTRACT does not support arg(s) ['position']
WARNING:sqlglot:REGEXP_EXTRACT does not support arg(s) ['position']
WARNING:sqlglot:REGEXP_EXTRACT does not support arg(s) ['position']
WARNING:sqlglot:REGEXP_EXTRACT does not support arg(s) ['position']
WARNING:sqlglot:REGEXP_EXTRACT does not support arg(s) ['position']
WARNING:sqlglot:REGEXP_EXTRACT does not support arg(s) ['position']
WARNING:sqlglot:REGEXP_EXTRACT does not support arg(s) ['position']
```

Let me know how you think I can debug this further.

5.1.0

duckdb

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
We weren't running these hooks before, which meant that a `memtable`
that hadn't been executed yet (via interactive mode or otherwise) would
not be registered with DuckDB at export time.

Fixes #6091 

While I was adding a test, I noticed that the partitioned parquet roundtripping test was comparing the same table to itself.  Made a few tweaks there.

### DIFF
--- a/ibis/backends/tests/test_export.py
+++ b/ibis/backends/tests/test_export.py
@@ -266,14 +266,14 @@ def test_roundtrip_partitioned_parquet(tmp_path, con, backend, awards_players):
 
 
 @pytest.mark.notimpl(
-    ["dask", "druid", "impala", "pyspark"], reason="No support for exporting files"
+    ["dask", "impala", "pyspark"], reason="No support for exporting files"
 )
 @pytest.mark.notimpl(
     ["datafusion"],
     reason="No memtable support",
 )
 @pytest.mark.parametrize("ftype", ["csv", "parquet"])
-def test_memtable_to_file(tmp_path, con, ftype):
+def test_memtable_to_file(tmp_path, con, ftype, monkeypatch):
     """
     Tests against a regression spotted in #6091 where a `memtable` that is
     created and then immediately exported to `parquet` (or csv) will error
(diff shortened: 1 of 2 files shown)


jcrist (approved):
Nice find!

ibis/backends/tests/test_export.py
```diff
@@ -254,11 +254,36 @@ def test_roundtrip_partitioned_parquet(tmp_path, con, backend, awards_players):
...
+
+@pytest.mark.notimpl(
+    ["dask", "impala", "pyspark", "druid"], reason="No to_parquet support"
+)
+@pytest.mark.parametrize("ftype", ["csv", "parquet"])
+def test_memtable_to_file(tmp_path, con, ftype):
+    """
+    Tests against a regression spotted in #6091 where a `memtable` that is
+    created and then immediately exported to `parquet` (or csv) will error
+    because we weren't registering the in-memory table with DuckDB before trying
```
cpcloud, line 276:
```suggestion
    because we weren't registering the in-memory table before trying
```

PR #6649, opened 2023-07-18
fix(snowflake): fix broken unnest functionality

cpcloud:
Snowflake broke our unnest functionality in https://docs.snowflake.com/en/release-notes/bcr-bundles/2023_04/bcr-1057.

This was a rabbit hole.

Unfortunately there's no straightforward way that I found to tell SQLAlchemy
to render a cross join using `,` syntax without using implicit cross joins. However,
implicit cross joins aren't viable for a couple of reasons

1. SQLAlchemy forces the immediate child expression to render first in an
   implicit cross join, but for unnest (flatten in snowflake) the containing
   table must occur before the flatten (or table producing function;
   `split_to_table` in our case) expression. There doesn't seem to be a way to force SQLAlchemy to
   render child relations in a specific order.
1. Using implicit cross joins required me to add a bunch of warnings handlers.
   This is less important than the first point, but nonetheless undesirable.

The solution I came up with was to alter the way a lateral cross join is compiled. They are straightforward
to recognize, since for UNNEST the lateral operand is always on the right and it will always render with `ON true` as
the end of the string.

Here, I simply chop off the `ON true` if the right operand is a lateral object.


ibis/backends/snowflake/__init__.py
```diff
@@ -499,3 +501,14 @@ def _get_temp_view_definition(
...
+
+
+@compiles(sa.sql.Join, "snowflake")
+def compile_join(element, compiler, **kw):
+    result = compiler.visit_join(element, **kw)
+
+    # snowflake doesn't support lateral joins with ON clauses as of
+    # https://docs.snowflake.com/en/release-notes/bcr-bundles/2023_04/bcr-1057
+    if element.right._is_lateral:
+        return re.sub(r"^(.*)\s+ON true$", r"\1", result, flags=re.IGNORECASE)
```
gforsyth, line 513: is there ever a scenario in which (in code we generate) that there will be multiple whitespace characters between `ON` and `true`?

gforsyth (approved):
Looks good -- one comment and non-blocking.

ibis/backends/snowflake/__init__.py
cpcloud: Not that I am aware of. Actually, it shouldn't matter since I'm using `.*` which matches whitespace.

gforsyth (approved):

cpcloud:
Snowflake tests look good:


```
❯ pytest -m snowflake -n auto --dist=loadgroup -q
bringing up nodes...
.....x......x................x...................xx....x.......xx...s.........s....s.......................x.....x.x......x......x....x...x......................................x......x..... [ 19%]
...........x....x...................x.x.......x.x.x...x..x..x...x........x.............x..x..x....................x..x................x.x............xx...x...xxx.xxxx..xxxxxxx.xx.x....x...xx [ 38%]
..xx.x.............x....x...x.x..xx.xxx......x....xx...........xx..............xxx..........xxx......x........x.........x.....xx...x....................x.xx.x....x......sx...x........xxx.x.x [ 57%]
....sx.x.......x.xx..x..x................xx.......x...x.......x..x............x....x.x...........x....x...x......xxxx...x........x......x.x.....x.......x..s..x......x.xx.x...x..x....x.x..... [ 77%]
.x..........x..x.xxx......x.....xxxxx....xx...x....x..........................x......x...x...............x..x...........x................x........xx.....x..........x.....x............x...... [ 96%]
....x.......sx.s..................                                                                                                                                                             [100%]
807 passed, 8 skipped, 169 xfailed in 270.47s (0:04:30)
```

Merging!

PR #6569, opened 2023-07-01
refactor(conversion): replace adhoc result handling with dunder methods specific to the `Expr` subclass

cpcloud:
This PR removes our ad-hoc `result_handler`-based shape reconciliation (Table
-> DataFrame | PyArrow Table, etc) in favor of two new methods:

* `__pandas_result__`
* `__pyarrow_result__`

each implemented on `ir.Table`, `ir.Column` and `ir.Scalar`. This has the benefit of removing quite a bit
of duplicated handling of expression-shape-to-in-memory-result-shape conversion.

PR #6655, opened 2023-07-19
refactor(analysis): remove `ScalarAggregate`, `reduction_to_aggregation` and `has_multiple_bases`

kszucs:
Most of the test suites do work, haven't checked impala yet. 

Apparently we no longer need the cross join, possibly due to recent changes.

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -77,11 +77,7 @@ def varargs(sa_func):
 
 def get_sqla_table(ctx, table):
     if ctx.has_ref(table, parent_contexts=True):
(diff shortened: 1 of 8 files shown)


cpcloud:
This seems doable in 6.1, is it not?

kszucs:
I would be a bit more defensive in this case, so my preference is to ship it with 7.0.

cpcloud (approved):
LGTM, nice work!

cpcloud:
Note that I am removing the `result_handler` things in #6569.

kszucs:
> Note that I am removing the `result_handler` things in #6569.

Even better, that annoyed me too.

PR #6661, opened 2023-07-20
fix(exprs): ensure that `how="left_semi"` and `how="semi"` are equivalent

cpcloud:
This PR fixes an issue where we were not treating `how="semi"` and `how="left_semi"` the same WRT deduplicating columns.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -2442,7 +2442,7 @@ class Table(Expr, _FixedTextJupyterMixin):
 
         # semi/anti join only give access to the left table's fields, so
         # there's never overlap
-        if how in ("semi", "anti"):
+        if how in ("left_semi", "semi", "anti"):
             return expr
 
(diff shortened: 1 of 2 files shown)


ibis/expr/types/relations.py
```diff
@@ -2442,7 +2442,7 @@ def join(
 
         # semi/anti join only give access to the left table's fields, so
         # there's never overlap
-        if how in ("semi", "anti"):
+        if how in ("left_semi", "semi", "anti"):
```
jcrist, line 2445: May want to update the comment above this line.

jcrist (approved):
Nice find. One tiny nit, feel free to ignore.

ibis/expr/types/relations.py
cpcloud: `left_semi` and `semi` are the same thing and there's no `right_semi`, so the comment is correct if incomplete. I'll leave it as is for now.

PR #6662, opened 2023-07-20
chore(clickhouse): workaround broken sqlglot

cpcloud:
Works around broken sqlglot version that cannot compile XOR for clickhouse.

Upstream bug report: https://github.com/tobymao/sqlglot/issues/1936


cpcloud:
Merging to get CI passing again.

PR #6663, opened 2023-07-20
fix(api): allow scalar window order keys

kszucs:
previously the builder API disallowed passing scalar order keys like ibis.NA and ibis.random()


cpcloud:
cc @saulpw

cpcloud (approved):
LGTM, thanks!

kszucs:
It had build errors for impala and mssql. Fixing them in a follow-up.

PR #6664, opened 2023-07-20
test(backends): mark window order by scalar test broken for impala and mssql

kszucs:


ibis/backends/tests/test_window.py
```diff
@@ -571,7 +571,17 @@ def test_simple_ungrouped_unbound_following_window(
...
-@pytest.mark.broken(["pandas", "dask"], raises=NotImplementedError)
+@pytest.mark.broken(
+    ["pandas", "dask"],
+    raises=NotImplementedError,
+    reason="support scalar sorting keys are not yet implemented",
+)
+@pytest.mark.broken(
+    ["pyspark"], raises=AnalysisException, reason="pyspark tries to locate None column"
+)
+@pytest.mark.broken(
```
cpcloud, line 582: `broken` should be used for things that we can potentially work around or fix, not for backend-unsupported features. For those, use `pytest.mark.notyet` or `pytest.mark.never`

cpcloud: Unless you think that we _can_ work around this, even if it's just to error early.

cpcloud (approved):

Issue #6669, opened 2023-07-21
feat: Python-like slicing for tables


lostmygithubaccount:
### Is your feature request related to a problem?

I find myself "paging" through tables sometimes, e.g. after computing the `topk(100)` it only displays the top 10 by default and I want to go see the next 10, etc.

I can accomplish this today with:

```python
t.some_column.topk(100)[10:20]
```

and so on

### Describe the solution you'd like

it'd be slightly nicer if I could use negative numbers and empty colon slicing as in Python lists and other objects, e.g.:

```python
t.some_column.topk(100)[10:] # show the second 10
t.some_column.topk(100)[-10:] # show the last 10
```

currently you get an error like:

```python
[ins] In [28]: t.browser.topk(20)[10:]
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[28], line 1
----> 1 t.browser.topk(20)[10:]

File ~/repos/ia2/venv/lib/python3.11/site-packages/ibis/expr/types/relations.py:522, in Table.__getitem__(self, what)
    519 stop = what.stop
    521 if stop is None or stop < 0:
--> 522     raise ValueError('End index must be a positive number')
    524 if start < 0:
    525     raise ValueError('Start index must be a positive number')

ValueError: End index must be a positive number
```

I'm not sure if this is easy or feasible to implement 

### What version of ibis are you running?

master

### What backend(s) are you using, if any?

duckdb, but any

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
I don't see an obvious issue with allowing `None`.

In general, negative slicing requires knowledge of the number of rows in the result set or some kind of SQL syntax for it as well as a unique ordering of the rows.

In the special case of a top-k we know the row count statically and we also have a unique ordering key, which means we the information necessary to return a meaningful result for negative slicing.

Unfortunately we don't have a way to specialize APIs like this at the expression level right now.

cpcloud:
Just had a bit of insight and remembered that some of our backends support scalar subqueries in `LIMIT` and `OFFSET`, which means we can dynamically compute those values using `COUNT(*)`, thus enabling negative slicing in some cases.

The main caveat is that the performance of `COUNT(*)` will depend on the query being counted, so the performance for negative slicing may be significantly different from non-negative slicing.

PR #6673, opened 2023-07-22
feat(api): allow slice endpoint to be `None`

cpcloud:
Adds support for a slice endpoint of `None`. Partially addresses #6669.


cpcloud:
Handling in #6675.

PR #1735, opened 2019-03-18
ENH/CLN: Reformat the codebase using black

cpcloud:
Also adds a pre-commit hook for black.

PR #6684, opened 2023-07-25
style: don't skip string normalization using Black

deepyaman:
[When Black was first introduced to Ibis over four years ago](https://github.com/ibis-project/ibis/pull/1735), `skip-string-normalization` was set to `true`. Even though there's no justification for the setting in the original PR, it's likely there was a reason (e.g. since the path filters weren't in place then, perhaps it messed up some files; haven't really looked into it).

[The purpose of the option is to aid migration for codebases with another existing convention.](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings) However, it's really just allowing needless inconsistency in Ibis at this time (e.g. [switching from double to single quotes halfway through a statement](https://github.com/ibis-project/ibis/blob/master/ibis/backends/pyspark/tests/conftest.py#L138-L156)).

(This also fixes two instances of "Specifiying" -> "Specifying", caught by `codespell` if you modify the files.) 

(Also removes an unused import.)


kszucs:
Thanks @deepyaman!

I think we want to enable this option, however the `7.x.x` branch contains several refactors and merging this PR would cause   too many conflicts, so we should target it against the `7.x.x` branch. Could you please update the PR?

cpcloud (approved):

cpcloud:
Thanks, approved!

deepyaman:
> Thanks @deepyaman!
>
> I think we want to enable this option, however the 7.x.x branch contains several refactors and merging this PR would cause > too many conflicts, so we should target it against the 7.x.x branch.
> 
> Could you please update the PR? Rerunning the linter on the 7.x.x is probably easier than rebasing this PR.

Sounds good; done!

> @deepyaman Probably best to `git reset --hard upstream/7.x.x` -> rerun `black .` -> force push, to avoid the conflicts.

Ah, yeah, I just changed the base before I made the changes. :) And then did more-or-less exactly what you said.

kszucs:
Thanks for updating!

We also need to add the commit to `.git-blame-ignore-revs` in the root directory, so we can easily look up a particular line's history.

deepyaman:
> We also need to add the commit to `.git-blame-ignore-revs` in the root directory, so we can easily look up a particular line's history.

Very interesting; I didn't know about this. Done! (I just took the commit hash from this PR for now; I assume this is fine if not squashing it down.)

cpcloud:
> > We also need to add the commit to `.git-blame-ignore-revs` in the root directory, so we can easily look up a particular line's history.
> 
> Very interesting; I didn't know about this. Done! (I just took the commit hash from this PR for now; I assume this is fine if not squashing it down.)

I believe this can only be done afterward because GitHub is going to rebase your commits on top of `upstream/master`. If anything is merged in the meantime that'll change the hash.

Not a big deal though, we can handle putting the specific commit in there as a follow-up.

PR #6702, opened 2023-07-27
fix(backends): fix pyspark `has_operation` definition; test that `has_operation` doesn't return `True` spuriously

cpcloud:
This PR fixes a few issues with `has_operation`:

1. The PySpark backend implementation didn't have a `@classmethod` decorator, so it didn't follow the base class definition and every other backend's implementation.
1. The ClickHouse backend was looking up operands using the mapping that backs the singledispatch registry, so `Backend.has_operation(object)` returned `True`.

Tests are included.


cpcloud:
Merging on green.

PR #6705, opened 2023-07-27
docs(dot-sql): add examples of mixing ibis expressions and SQL strings

cpcloud:
Adds an example of mixing SQL strings with ibis expressions to the `.sql` method docstring

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -2650,18 +2650,15 @@ class Table(Expr, _FixedTextJupyterMixin):
         return expr
 
     def sql(self, query: str, dialect: str | None = None) -> ir.Table:
-        """Run a SQL query against a table expression.
-
-        See [`Table.alias`][ibis.expr.types.relations.Table.alias] for
-        details on using named table expressions in a SQL string.
+        '''Run a SQL query against a table expression.
 
         Parameters
         ----------
         query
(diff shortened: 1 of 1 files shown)


ibis/expr/types/relations.py
```diff
@@ -2671,20 +2668,67 @@ def sql(self, query: str, dialect: str | None = None) -> ir.Table:
...
+        ┏━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
+        ┃ island    ┃ avg_bill_length ┃
+        ┡━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
+        │ string    │ float64         │
+        ├───────────┼─────────────────┤
+        │ biscoe    │            45.3 │
+        │ dream     │            44.2 │
+        │ torgersen │            39.0 │
+        └───────────┴─────────────────┘
+        >>> expr.alias("b").sql("SELECT * FROM b WHERE avg_bill_length > 40")
```
jcrist, line 2718: Can you add a brief sentence above this line saying why the alias is necessary here? Something like "If you want to use `.sql` on an existing ibis expression, you first need to assign it an alias to use in the query. This can be done with the `alias` method".

jcrist (approved):
Overall LGTM

PR #6712, opened 2023-07-28
refactor(ir): construct `ArrayContains` instead of `Contains` for `value.isin(array_value)`

kszucs:
`ops.Contains` support three kind of haystack arguments, one is `Value[dt.Array]` where the expression is identical with `arr.contains(value)`

### DIFF
--- a/ibis/expr/operations/logical.py
+++ b/ibis/expr/operations/logical.py
@@ -133,11 +133,7 @@ class Between(Value):
 @public
 class Contains(Value):
     value: Value
-    options: Union[
-        VarTuple[Value],
-        Column[dt.Any],
-        Value[dt.Array],
-    ]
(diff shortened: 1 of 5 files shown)


ibis/expr/operations/logical.py
```diff
@@ -133,11 +133,7 @@ def __init__(self, arg, lower_bound, upper_bound):
 @public
 class Contains(Value):
     value: Value
-    options: Union[
-        VarTuple[Value],
-        Column[dt.Any],
-        Value[dt.Array],
-    ]
+    options: Union[VarTuple[Value], Column[dt.Any]]
```
kszucs, line 136: Going to further reduce the supported haytask kinds in a follow-up.

cpcloud (approved):
LGTM, nice PR!

PR #6714, opened 2023-07-28
refactor(ir): decompose `Contains` into `InValues` and `InColumn`

kszucs:
Depends on #6712 

Also removes the `ops.NotContains` operator in favor of `ops.Not(ops.Contains())`


ibis/backends/base/sql/registry/binary_infix.py
```diff
@@ -50,37 +48,41 @@ def xor(translator, op):
...
-            right = translator.translate(op.options)
-
-        # we explicitly do NOT parenthesize the right side because it doesn't
-        # make sense to do so for Sequence operations
-        return f"{left} {op_string} {right}"
-
-    return translate
+def in_values(translator, op):
+    if not op.options:
+        return "FALSE"
```
cpcloud, line 53: This seems like a really strange edge, but such is life 😮‍💨

kszucs: Well, we could avoid it on the expression level, but that can be done in a follow-up.

cpcloud (approved):
Sweet!

PR #6716, opened 2023-07-28
feat(api): support boolean literals in join API

### LINKED ISSUE #6696: feat: support `True`, `ibis.literal(True)` as predicates to join

I am trying to make a framework where users can specify a join condition, and I inject that into a `left.inner_join(right, predicates=condition)`.  If a user wants to do a cross join, in SQL they would be able to pass in `1=1`, so they should be able to do the equivalent of this by passing in a literal True.

support `True`, `ibis.literal(True)` as predicates. Maybe I'm not including some other values that make this more general. I guess `False` would also work, and should just give a table with the right schema, but 0 rows.

6.0.0

NA

### Code of Conduct

cpcloud:
This PR adds support for literal `True` and `False` join predicates, which are supported across most backends. Closes #6696.

### DIFF
--- a/ibis/expr/operations/relations.py
+++ b/ibis/expr/operations/relations.py
@@ -169,6 +169,8 @@ def _clean_join_predicates(left, right, predicates):
             pred = lk == rk
         elif isinstance(pred, str):
             pred = left.to_expr()[pred] == right.to_expr()[pred]
+        elif pred is True or pred is False:
+            pred = ops.Literal(pred, dtype="bool").to_expr()
         elif isinstance(pred, Value):
             pred = pred.to_expr()
         elif isinstance(pred, Deferred):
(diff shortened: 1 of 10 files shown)


ibis/expr/operations/relations.py
```diff
@@ -169,6 +169,8 @@ def _clean_join_predicates(left, right, predicates):
             pred = lk == rk
         elif isinstance(pred, str):
             pred = left.to_expr()[pred] == right.to_expr()[pred]
+        elif pred is True or pred is False:
```
kszucs, line 172:
```suggestion
        elif isinstance(pred, bool):
```

kszucs: It better aligns with the rest of the conditions and also support bool subclasses.

kszucs (approved):
LGTM, just a nit.

ibis/expr/operations/relations.py
cpcloud: `bool` cannot be subclassed.

```
In [1]: class MyBool(bool):
   ...:     pass
Cell In[1], line 1
----> 1 class MyBool(bool):
      2     pass

TypeError: type 'bool' is not an acceptable base type
```

PR #6722, opened 2023-07-29
feat(snowflake): allow disabling creation of object UDFs

### LINKED ISSUE #6654: feat: Make creating UDF's optional

When opening a Snowflake connection, there is some code that creates UDFs in snowflake. For reading tables/executing plain SQL on snowflake they do not seem to be required, so I would like to turn that behaviour off and save some seconds on establishing the initial connection.

I see this code here https://github.com/ibis-project/ibis/blob/master/ibis/backends/snowflake/__init__.py#L230. It would be great to have a connection parameter that allows me to skip this part (e.g. `udf_support=False`). 

6.0.0

Snowflake, BigQuery, ...

### Code of Conduct

cpcloud:
Allow disabling object UDFs when connecting to the snowflake backend. Closes #6654.

### DIFF
--- a/ibis/backends/snowflake/__init__.py
+++ b/ibis/backends/snowflake/__init__.py
@@ -163,6 +163,7 @@ $$ {defn["source"]} $$"""
         password: str | None = None,
         authenticator: str | None = None,
         connect_args: Mapping[str, Any] | None = None,
+        create_object_udfs: bool = True,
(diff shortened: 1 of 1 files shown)


cpcloud:
@kszucs Can you take a look when you get a chance?

kszucs (approved):

PR #6724, opened 2023-07-29
feat(api): add `concat` and `repeat` methods for easier array operation discovery

### LINKED ISSUE #6720: feat: document Array.__add__()

I was looking for Array.concat(), and couldn't find it. I had to go try `a + b`, and that worked, but that would be nice if it was discoverable.

Add __add__() to the docs. OR perhaps just add in an explicit `.concat()` method to Arrays so it is even easier. At least in the __add__method docstring include the keyword concat so that it shows up in search.

main

Na

### Code of Conduct

cpcloud:
Make array concatenation and repetition easier to discover.

Closes #6720.

### DIFF
--- a/ibis/expr/types/arrays.py
+++ b/ibis/expr/types/arrays.py
@@ -125,18 +125,20 @@ class ArrayValue(Value):
             op = ops.ArrayIndex(self, index)
         return op.to_expr()
 
-    def __add__(self, other: ArrayValue) -> ArrayValue:
-        """Concatenate this array with another.
+    def concat(self, other: ArrayValue, *args: ArrayValue) -> ArrayValue:
+        """Concatenate this array with one or more arrays.
 
         Parameters
         ----------
         other
-            Array to concat with `self`
+            Other array to concat with `self`
+        args
+            Other arrays to concat with `self`
 
         Returns
         -------
         ArrayValue
-            `self` concatenated with `other`
+            `self` concatenated with `other` and `args`
 
         Examples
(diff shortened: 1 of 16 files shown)


ibis/expr/types/arrays.py
```diff
@@ -153,71 +155,59 @@ def __add__(self, other: ArrayValue) -> ArrayValue:
...
         └──────────────────────┘
         """
-        return ops.ArrayConcat(other, self).to_expr()
+        return ops.ArrayConcat((self, other, *args)).to_expr()
+
+    def __add__(self, other: ArrayValue) -> ArrayValue:
+        return self.concat(other)
+
+    def __radd__(self, other: ArrayValue) -> ArrayValue:
+        return ops.ArrayConcat((other, self)).to_expr()
```
kszucs, line 208: `other.concat(self)`?

kszucs (approved):
Nice!

ibis/expr/types/arrays.py
cpcloud: `other` can be a different type than `ArrayValue`. I should probably remove or fix the type annotation there.

kszucs: Right, and `self` cannot be used because of the concat order.

kszucs: Regarding the annotations we should have type aliases at some point, e.g. `ArrayLike` which contains the types coercible to array values as well.

PR #6727, opened 2023-07-30
chore(error-reporting): add more thorough information about slicing a non-sliceable column

### LINKED ISSUE #6174: docs: document how slicing differs from numpy and pandas
xref discussion in https://github.com/ibis-project/ibis/issues/6090.

@NickCrews suggested we document this to help out users coming from numpy/pandas.

cpcloud:
Closes #6174.

### DIFF
--- a/ibis/expr/types/arrays.py
+++ b/ibis/expr/types/arrays.py
@@ -825,7 +825,8 @@ class ArrayScalar(Scalar, ArrayValue):
 
 @public
 class ArrayColumn(Column, ArrayValue):
-    pass
+    def __getitem__(self, index: int | ir.IntegerValue | slice) -> ir.Column:
+        return ArrayValue.__getitem__(self, index)
 
 
 @public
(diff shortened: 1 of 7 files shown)


ibis/expr/types/arrays.py
```diff
@@ -825,7 +825,8 @@ class ArrayScalar(Scalar, ArrayValue):
 
 @public
 class ArrayColumn(Column, ArrayValue):
-    pass
+    def __getitem__(self, index: int | ir.IntegerValue | slice) -> ir.Column:
```
kszucs, line 828: Why to override this method? Only to change the return annotation?

kszucs: I guess we would need to turn Column to a generic type then which we can implement later.

cpcloud: It's less invasive to not reorder inheritance. Fewer changes were needed. We can reorder the base classes later if we want to.

kszucs (approved):

PR #6733, opened 2023-08-01
feat(api): improve efficiency of `__dataframe__` protocol

jcrist:
This improves the efficiency and usability of the `__dataframe__` protocol in a few ways:

- It avoids executing the wrapped query for operations where that's not strictly required. This allows fetching things like column names and types, or subselecting columns to occur without executing the query first.
- Given a single `__dataframe__()` output (an `IbisDataFrame` instance), the wrapped query will execute at most once for all methods called. Once executed, the resulting `pyarrow.Table` is stored on the `IbisDataFrame` instance to be reused by subsequent access.
- If column subselection is done on an already executed `IbisDataFrame`, no computation results.

Given these improvements, an `ibis.Table` passed to `altair` (dev) now results in only one execution, rather than potentially several executions.

### DIFF
--- a/ibis/backends/tests/test_export.py
+++ b/ibis/backends/tests/test_export.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import pandas as pd
 import pandas.testing as tm
+import pyarrow as pa
 import pytest
 import sqlalchemy as sa
 from packaging.version import parse as vparse
@@ -10,8 +11,6 @@ from pytest import param
 import ibis
 import ibis.expr.datatypes as dt
 from ibis import util
-
-pa = pytest.importorskip("pyarrow")
 
 try:
     from pyspark.sql.utils import ParseException
@@ -417,17 +416,6 @@ def test_arrow_timestamp_with_time_zone(alltypes):
     with t.to_pyarrow_batches() as reader:
         (batch,) = reader
     assert batch.schema.types == expected
-
-
-@pytest.mark.notimpl(["dask", "druid"])
-@pytest.mark.notimpl(
-    ["impala"], raises=AttributeError, reason="missing `fetchmany` on the cursor"
-)
-def test_dataframe_protocol(alltypes):
-    pytest.importorskip("pyarrow", minversion="12")
-    output = alltypes.__dataframe__()
-    assert list(output.column_names()) == alltypes.columns
-    assert alltypes.count().to_pandas() == output.num_rows()
 
 
 @pytest.mark.notimpl(["dask", "druid"])
--- a/ibis/backends/tests/test_dataframe_interchange.py
+++ b/ibis/backends/tests/test_dataframe_interchange.py
@@ -1,0 +1,132 @@
+from __future__ import annotations
+
+import pyarrow as pa
+import pytest
+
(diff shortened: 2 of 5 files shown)


ibis/backends/tests/test_export.py
```diff
@@ -456,3 +446,130 @@ def test_empty_memtable(backend, con):
...
+class TestDataframeInterchange:
+    """Test the __dataframe__ interchange protocol"""
+
+    def test_dataframe_protocol_no_execute(self, con, alltypes, monkeypatch):
+        t = alltypes.select("int_col", "double_col", "string_col")
+        pa_df = t.to_pyarrow().__dataframe__()
+
+        # Patch `con.to_pyarrow` to allow tracking execution
+        to_pyarrow = unittest.mock.Mock(wraps=con.to_pyarrow)
+        monkeypatch.setattr(con, "to_pyarrow", to_pyarrow)
```
jcrist, line 465: I don't love patching and mocks, but for the case of counting method calls (to ensure no execution/at most one execution) I think they're a fine solution. I couldn't find another method within ibis that would suffice, but would be happy to switch if there is one.

ibis/backends/tests/test_export.py
```diff
@@ -456,3 +446,130 @@ def test_empty_memtable(backend, con):
...
     result = con.execute(table)
     backend.assert_frame_equal(result, expected)
+
+
+@pytest.mark.notimpl(["dask", "druid"])
+@pytest.mark.notimpl(
+    ["impala"], raises=AttributeError, reason="missing `fetchmany` on the cursor"
+)
+@pytest.mark.skipif(pa.__version__ < "12.", reason="Pyarrow >= 12 required")
+class TestDataframeInterchange:
```
jcrist, line 456: In other projects I've started to group similar tests into test classes. I like this method for keeping the test files organized, and have found it to be effective even for large and evolving test files. Otherwise I've found pytest test cases can end up scattered throughout the test file, making it harder to see what tests exist for a specific piece of functionality.

I do recognize this is a new pattern in ibis though - happy to move these to functions alone if needed.

cpcloud: I'd rather keep these as functions, but you can achieve the same level of organization by making a new test module, which is even easier to discover since you don't even have to open a file to know what's inside!

ibis/backends/tests/test_export.py
cpcloud: You should be able to use `mocker.spy` here, which avoids having to mess around with mocking directly. Here's an example where we currently use it to test the call count when registering memtables using the snowflake backend:

```
def test_repeated_memtable_registration(simple_con, mocker):
    data = {"key": list("abc"), "value": [[1], [2], [3]]}
    expected = pd.DataFrame(data)
    t = ibis.memtable(data)

    spy = mocker.spy(simple_con, "_register_in_memory_table")

    n = 2

    for _ in range(n):
        tm.assert_frame_equal(simple_con.execute(t), expected)

    # assert that we called _register_in_memory_table exactly n times
    assert spy.call_count == n
```

jcrist:
Thanks for the review, I believe this should be good-to-go now.

cpcloud (approved):
Excellent work.

Excited to see this unblock some sweet integrations!

PR #6735, opened 2023-08-01
feat(common): match and replace graph nodes

kszucs:
This PR adds a `Node.match()` and `Node.replace()` methods to do search and replace tasks on node/operation trees. 

While `Node.match()` can return with nodes rewritten by the pattern, `Node.replace()` actually rebuilds the whole expression so the first is suitable for matching tasks while the latter is to rewrite operation trees.

Examples:

```py
# replace scalar parameters with the provided values prior to execution
from ibis.common.patterns import Replace

rule = Replace(
    ops.ScalarParameter,
    lambda op, ctx: ops.Literal(value=params[op], dtype=op.dtype),
)
node = node.replace(rule)
```

```py
# compose two replace rules
from ibis.common.patterns import EqualTo

class Aliased: ...
class Name: ...

ketto = Aliased(one, "ketto")

first_rule = EqualTo(Name("one")) >> Name("zero")
second_rule = EqualTo(two) >> ketto

new_values = values.replace(first_rule | second_rule)
```

```py
# propagate windows down in the expression tree
from ibis.common.patterns import Variable, Object, Call

col = ...
expected_frame = ...

func = Variable("func")
frame = Variable("frame")

p = Object.namespace(ops)
c = Call.namespace(ops)

pattern = Node(
    p.Value & ~p.WindowFunction,
    each_arg=func @ p.Analytic >> c.WindowFunction(func, frame),
)
result = col.replace(pattern, context={frame: expected_frame})
```

Also:
- move `util.recursive_get()` to `ibis.common.graph`
- add docstrings to `ibis.common.graph`
- improve docstrings for `ibis.common.patterns`

### DIFF
--- a/ibis/common/graph.py
+++ b/ibis/common/graph.py
@@ -104,55 +104,6 @@ class Node(Hashable):
         """Support for rich reprerentation of the node."""
         return zip(self.__argnames__, self.__args__)
 
-    @experimental
-    def branches(self) -> Iterator[Sequence[Node]]:
-        """Yield all branches of the graph.
-
-        A branch is a path from the root to a leaf node. This method is primarily
-        used to implement the `path` method supporting `XPath`-like queries.
-
-        Yields
-        ------
-        A sequence of nodes representing a branch.
-        """
-        stack = [(self, [])]
-
-        while stack:
-            node, path = stack.pop()
-
-            if children := node.__children__():
-                for child in reversed(children):
-                    stack.append((child, path + [node]))
-            else:
-                yield path + [node]
-
-    @experimental
-    def path(
-        self, *pats: Any, context: Optional[dict] = None
-    ) -> Iterator[Sequence[Node]]:
-        """Return the first tree branch matching a given sequence pattern.
-
-        This method provides a way to query the graph using `XPath`-like expressions.
-        The following XPath expression "//Alias//Value[dtype==int64]" would roughly
-        translate to the following Python code:
-
-            node.path(..., Alias, ..., Object(Value, dtype=dt.Int64), ...)
-
-        Parameters
-        ----------
-        pats
-            Sequence which is coerced to a sequence pattern. See `ibis.common.patterns`
-            for more details.
-        context
-            Optional context to use for the pattern matching.
-        """
-        pat = pattern(list(pats))
-        for branch in self.branches():
-            result = pat.match(branch, context)
-            if result is not NoMatch:
-                return result
-        return NoMatch
-
     def map(self, fn: Callable, filter: Optional[type] = None) -> dict[Node, Any]:
         """Apply a function to all nodes in the graph.
 
--- a/ibis/common/graph.py
+++ b/ibis/common/graph.py
@@ -4,66 +4,19 @@ from __future__ import annotations
 from abc import abstractmethod
 from collections import deque
 from collections.abc import Hashable, Iterable, Iterator, Mapping
-from typing import Any, Callable, Dict, Sequence
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Sequence
 
-from ibis.util import recursive_get
+from ibis.common.patterns import NoMatch, pattern
+from ibis.util import experimental
 
-
-class Node(Hashable):
-    __slots__ = ()
-
-    @property
-    @abstractmethod
-    def __args__(self) -> Sequence:
-        ...
-
-    @property
-    @abstractmethod
-    def __argnames__(self) -> Sequence:
-        ...
-
-    def __children__(self, filter=None):
-        return tuple(_flatten_collections(self.__args__, filter or Node))
-
-    def __rich_repr__(self):
-        return zip(self.__argnames__, self.__args__)
-
-    def map(self, fn, filter=None):
-        results = {}
-        for node in Graph.from_bfs(self, filter=filter).toposort():
-            kwargs = dict(zip(node.__argnames__, node.__args__))
-            kwargs = recursive_get(kwargs, results)
-            results[node] = fn(node, results, **kwargs)
-
-        return results
-
-    def find(self, type, filter=None):
-        def fn(node, _, **kwargs):
-            if isinstance(node, type):
-                return node
-            return None
-
-        result = self.map(fn, filter=filter)
-
-        return {node for node in result.values() if node is not None}
-
-    def substitute(self, fn, filter=None):
-        return self.map(fn, filter=filter)[self]
-
-    def replace(self, subs, filter=None):
-        def fn(node, _, **kwargs):
-            try:
-                return subs[node]
-            except KeyError:
-                return node.__class__(**kwargs)
-
-        return self.substitute(fn, filter=filter)
+if TYPE_CHECKING:
+    from typing_extensions import Self
 
 
-def _flatten_collections(node, filter=Node):
+def _flatten_collections(node: Node, filter: type) -> Iterator[Node]:
     """Flatten collections of nodes into a single iterator.
 
-    We treat common collection types inherently Node (e.g. list, tuple, dict)
+    We treat common collection types inherently traversable (e.g. list, tuple, dict)
     but as undesired in a graph representation, so we traverse them implicitly.
 
     Parameters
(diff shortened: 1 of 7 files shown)


ibis/tests/expr/test_operations.py
```diff
@@ -156,18 +156,18 @@ def record(node, _, *args, **kwargs):
     ]
 
 
-def test_node_subtitution():
-    class Aliased(Base):
-        arg: ops.Node
-        name: str
+# def test_node_subtitution():
```
kszucs, line 159: Need to rewrite this test case to work with the new API.

kszucs: Rewritten.

NickCrews:
This is a bit of a long shot, but in a framework I'm trying to write, people can give me join predicates such as `(l.name == r.name) & ((l.dob == r.dob) | (l.address == r.address)` (a bunch of OR clauses joined by ANDs). I want to be able to rewrite these in [DNF](https://en.wikipedia.org/wiki/Disjunctive_normal_form) (a bunch of AND clauses joined by ORs ie `((l.name ==r.name) & (l.dob == r.dob)) | ((l.name ==r.name) & (l.address == r.address)) because that would give much better join performance, because I can convert it to a bunch of individual joins UNIONed together

```python
a = l.join(r, (l.name == r.name) & (l.dob == r.dob))
b = l.join(r, (l.name == r.name) & (l.address == r.address))
result = ibis.union(a, b)
```
This is better because if the join condition is just AND conditions then the engine can use a hash join, but if an OR slips in there, it has to use a nested loop join. See [here](https://github.com/moj-analytical-services/splink/discussions/1417#discussioncomment-6420575) for more details. Does that make sense?

Would this feature support this? Is this use case within the scope of Ibis? I understand how this is beyond what 99% of Ibis users would ever want.

kszucs:
Definitely makes sense!

We are working towards a toolset which should provide a framework for it. Since the problem is not trivial, we already had a couple of iterations with it, but I hope this time we will finally achieve the goal. 

Currently ibis has a more flexible pattern system under `ibis.common.patterns` and a less flexible but more robust graph rewriting solution using [EGraphs](https://egraphs-good.github.io) under `ibis.common.egraphs`. The former should be more suitable for *eager* rewrites where we are sure that the outcome is strictly better while the latter should be preferred for *optimization-like* rewrites.

ibis/common/tests/test_patterns.py
```diff
@@ -97,6 +104,30 @@ def __eq__(self, other):
...
+
+
+def test_never():
+    p = Never()
+    assert p.match(1, context={}) is NoMatch
+    assert p.match(2, context={}) is NoMatch
+
+
+def test_just():
+    p = Just(1)
```
cpcloud, line 125: How about `Const`?

ibis/common/tests/test_patterns.py
```diff
@@ -97,6 +104,30 @@ def __eq__(self, other):
         return self.__class__ == other.__class__ and self.min == other.min
 
 
+x = Variable("x")
+y = Variable("y")
+z = Variable("z")
+
+
+def test_always():
+    p = Always()
```
cpcloud, line 113: `Ident` or `Identity`?

ibis/common/tests/test_patterns.py
```diff
@@ -401,24 +432,50 @@ def test_mapping_of():
...
+def test_object_pattern():
     p = Object(Foo, 1, b=2)
     o = Foo(1, 2)
     r = match(p, o)
     assert r is o
     assert r == Foo(1, 2)
 
 
+def test_object_pattern_complex_type():
+    p = Object(Not(Foo), 1, 2)
```
cpcloud, line 465: This may be just my brain, but I would find the following easier to read:

```python
Not(Object(Foo, 1, 2))
```

Though now after staring at it for a minute I see that the pattern you've written has different semantics than what I wrote.

The one I wrote says: "match anything that isn't an instance of Foo constructed with arguments 1 and 2"

The one you wrote says: "match anything that was constructed with arguments 1 and 2, as long as it is **not** an instance of `Foo`"

Tricky!

ibis/common/tests/test_patterns.py
```diff
@@ -495,6 +554,60 @@ def test_matching():
...
 
+def test_replace_passes_matched_value_as_underscore():
+    class MyInt:
+        def __init__(self, value):
+            self.value = value
+
+        def __eq__(self, other):
+            return self.value == other.value
+
+    p = InstanceOf(int) >> Call(MyInt, value=_)
```
cpcloud, line 564: What is the difference between `Call` and `Object`?

ibis/common/tests/test_patterns.py
```diff
@@ -495,6 +554,60 @@ def test_matching():
...
+    assert g.b.d == 4
+
+    # nested example using reference to replace a value with a captured value
+    p = Object(
+        Foo,
+        1,
+        b=Replace(
+            Object(Bar, 2, d="d" @ Any()), lambda value, ctx: Foo(-1, b=ctx["d"])
+        ),
+    )
```
cpcloud, line 588: 🤯

ibis/common/tests/test_patterns.py
```diff
@@ -865,13 +975,16 @@ def f(x):
...
-class Term(Node):
+class Term(GraphNode):
     def __eq__(self, other):
         return type(self) is type(other) and self.__args__ == other.__args__
 
     def __hash__(self):
         return hash((self.__class__, self.__args__))
 
+    def __repr__(self):
+        return f"{self.__class__.__name__}({', '.join(map(repr, self.__args__))})"
```
cpcloud, line 986: Perhaps we should add a very simple test?

ibis/util.py
```diff
@@ -351,16 +351,6 @@ def consume(iterator: Iterator[T], n: int | None = None) -> None:
         next(itertools.islice(iterator, n, n), None)
 
 
-# TODO(kszucs): make it a more robust to better align with graph._flatten_collections()
```
cpcloud, line 354: Yay!

ibis/tests/expr/test_operations.py
```diff
@@ -162,9 +163,11 @@ class Aliased(Base):
         name: str
 
     ketto = Aliased(one, "ketto")
-    subs = {Name("one"): Name("zero"), two: ketto}
 
-    new_values = values.replace(subs)
+    first_rule = EqualTo(Name("one")) >> Name("zero")
+    second_rule = EqualTo(two) >> ketto
+
+    new_values = values.replace(first_rule | second_rule)
```
cpcloud, line 170: What are the semantics of the `|` operator here?

ibis/expr/operations/tests/test_matching.py
cpcloud, line 1: Should this file be empty?

ibis/common/graph.py
```diff
@@ -4,100 +4,302 @@
...
+        A sequence of nodes representing a branch.
+        """
+        stack = [(self, [])]
+
+        while stack:
+            node, path = stack.pop()
+
+            if children := node.__children__():
+                for child in reversed(children):
+                    stack.append((child, path + [node]))
```
cpcloud, line 125: This will not traverse the graph in topological order. Is that intentional?

ibis/common/patterns.py
```diff
@@ -275,6 +294,160 @@ def __rich_repr__(self):
             yield name, getattr(self, name)
 
 
+class Creator(Hashable):
```
cpcloud, line 297: This needs a name that is more descriptive, or at least a docstring describing what concept it's modeling.

ibis/common/patterns.py
```diff
@@ -275,6 +294,160 @@ def __rich_repr__(self):
...
+    __slots__ = ()
+
+    @abstractmethod
+    def __eq__(self, other):
+        ...
+
+    @abstractmethod
+    def make(self, context):
```
cpcloud, line 305: This looks like it is the primary interface for this class, so it should have a docstring.

ibis/common/patterns.py
```diff
@@ -275,6 +294,160 @@ def __rich_repr__(self):
...
+        return context[self]
+
+    def __getattr__(self, name):
+        return Call(operator.attrgetter(name), self)
+
+    def __getitem__(self, name):
+        return Call(operator.itemgetter(name), self)
+
+
+class Just(_Slotted, Creator):
```
cpcloud, line 345: The objects here need docstrings describing what concept they are modeling.

cpcloud (changes requested):
Looking good! A few comments.

ibis/common/tests/test_patterns.py
kszucs: Or `Exactly`? I'm not settled with the naming, we should make these more descriptive. Though in this particular case `Just` is implicitly defined in most of the cases.

ibis/common/tests/test_patterns.py
kszucs: Then how would you call `Never()`?

ibis/common/tests/test_patterns.py
kszucs: Call is `Creator` (awful naming I know) while `Object` is a pattern. The difference between the two is the former only receives the `context` with the captured values to *construct* a new object which is substituted by the `Replace` pattern. The latter does matching on objects considering both the objects' type and arguments.

ibis/common/tests/test_patterns.py
kszucs: Yeah, this is not the nicest way to spell it. There is already some syntax sugar which improves it, but I'm definitely open for suggestions!

ibis/tests/expr/test_operations.py
kszucs: The `|` operator constructs an `AnyOf(first_rule, second_rule)` pattern.

ibis/expr/operations/tests/test_matching.py
kszucs: It is probably a leftover, removing it.

ibis/common/graph.py
kszucs: In this case the topological order is not relevant because we don't rewrite the tree, it just yields all branches and `node.path()` matches on those branches, but doesn't reconstruct the tree afterwards.

I'm not settled with this implementation yet, perhaps we won't need the `XPath`-like matching at all, then we can remove it.

ibis/common/patterns.py
kszucs: Yeah, I'm not satisfied with it either. Going to add docstring, in the meantime I'm open for suggestions. 

There is a single place where creator objects can be used, the right hand side of the replace pattern: `Replace(pattern, creator)`. Semantically when a match occurs for the pattern, the context contains all the captured values and information relevant for the creator, and the creator construct a new object only given by the context.

ibis/common/patterns.py
kszucs: Going to add. If you have a better name suggestion for the `Creator` base class then please come up with a good verb for this method too.

ibis/common/patterns.py
kszucs: Adding.

ibis/common/patterns.py
kszucs: I was think `Replacer`, `Applier`, `Substitutor`, but the replacement is actually done by the `Replace` pattern and `Creator` objects just construct things given the context.

ibis/common/graph.py
cpcloud: I think it might be relevant if there's a parent pattern that would match had it been traversed before a child pattern.

Consider this traversal order:

```mermaid
flowchart TD
    A[1]
    A --> B[2]
    A --> C[4]
    B --> D[3]
    C --> D  
```

Are there any issues with matching on 3 before matching on 4, if both match?

kszucs: For the following node: `A(B(C, D), E)`

```py
branches() ->
    [A, E]
    [A, B, D]
    [A, B, C]
```

`path()` iterates over that list of lists and tries to sequence match them.

So no pattern matching is happening during the traversal.

kszucs: This is totally experimental, we may not need it eventually, or we may come up with a better way to do matching like this - I kinda have an alternative idea.

kszucs: After discussing it, the problem comes from the ambiguity of the branch order, because `.path()` returns with the first match. I'm removing the `XPath` like matching until we actually need it. It will be in a separate commit so we can easily restore it.

ibis/common/patterns.py
kszucs: I renamed it to Builder, still not satisfied but we can change it later.

ibis/common/tests/test_patterns.py
cpcloud: Reconsidering this, I think `Always` is fine.

ibis/common/tests/test_patterns.py
cpcloud: This is fine!

ibis/common/tests/test_patterns.py
cpcloud: I guess the question is whether the nouns here (`Call`, `Object`, etc.) are referring to the identities of the match candidate ("this is a `Call`") or whether the nouns are referring to properties of the match candidate ("this was `CalledWith`").

It seems like you've gone with the first option, which is fine!

cpcloud (approved):
LGTM!

kszucs:
Fixed the doctest failure.

NickCrews:
@kszucs Nice on this! Do I understand it correctly then that this API isn't mature yet, so I shouldn't rely on it yet? Is there any timeline planned, or that is more of a longterm goal? Could I make an issue or something to track progress? I'm not blocked by this, but without it I think I am going to have to expose a more restrictive API to my users so they can't write non-performant queries.

kszucs:
> Nice on this! Do I understand it correctly then that this API isn't mature yet, so I shouldn't yet try to code against it?
Probably not yet, this is rather an internal API and most likely we are going to iterate with it.

> Is there any timeline planned, or that is more of a longterm goal? 
No exact timeline, first we incrementally want to reduce the internal technical debt we have, then we can add more sophisticated rewrites.

> Could I make an issue or something to track progress? 
Of course. Actually some feedback about the current rewrite capabilities, desired features or just examples you with to work would help moving forward with the effort!

PR #6736, opened 2023-08-01
feat(trino): add support for table properties

### LINKED ISSUE #6488: feat(trino): enable passing table properties to trino's `create_table`

I am unable to pass [Trino's table properties](https://trino.io/docs/current/sql/create-table.html) to ibis's `con.create_table` method using the Trino backend.

I'd like `create_table` to have an API for passing table properties to the Trino backend.

N/A

Trino

### Code of Conduct

cpcloud:
This PR adds support for `create_table` with table properties to the Trino backend.

I don't like that for just a single test, I'm having to add three new containers
to our CI setup.

However, as far as I can tell Hive is only connector that has table properties.
If there are other connectors that are easier to setup that have any table
properties, I'm all ears.

There's some additional utility in having a functioning Hive connector around
which is that I plan to use it to test the cross-schema functionality as a
follow-up to this PR.

Ideally we can also get rid of our additional `trino-postgres` container but
after trying that locally it's going to be a bit more work and worth a separate PR.

Closes #6488.

### DIFF
--- a/docker/trino/catalog/hive.properties
+++ b/docker/trino/catalog/hive.properties
@@ -1,0 +1,15 @@
+connector.name=hive
+
+hive.allow-drop-table=true
+hive.ignore-absent-partitions=true
+hive.metastore.thrift.delete-files-on-drop=true
+hive.metastore.uri=thrift://hive-metastore:9083
+hive.metastore.username=admin
+hive.non-managed-table-writes-enabled=true
+hive.s3.aws-access-key=accesskey
+hive.s3.aws-secret-key=secretkey
+hive.s3.endpoint=http://hive-metastore-minio:9000
+hive.s3.path-style-access=true
+hive.s3select-pushdown.enabled=true
+hive.storage-format=PARQUET
+hive.timestamp-precision=MICROSECONDS
--- a/ibis/backends/trino/__init__.py
+++ b/ibis/backends/trino/__init__.py
@@ -2,21 +2,31 @@
 
 from __future__ import annotations
 
+import collections
 import contextlib
 import warnings
 from functools import cached_property
-from typing import Iterator
+from typing import TYPE_CHECKING, Any, Iterator, Mapping
 
+import pandas as pd
 import sqlalchemy as sa
 import toolz
 from trino.sqlalchemy.datatype import ROW as _ROW
 
+import ibis
+import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
+import ibis.expr.types as ir
 from ibis import util
 from ibis.backends.base.sql.alchemy import AlchemyCanCreateSchema, BaseAlchemyBackend
 from ibis.backends.base.sql.alchemy.datatypes import ArrayType
 from ibis.backends.trino.compiler import TrinoSQLCompiler
-from ibis.backends.trino.datatypes import ROW, parse
+from ibis.backends.trino.datatypes import ROW, TrinoType, parse
+
+if TYPE_CHECKING:
+    import pyarrow as pa
+
+    import ibis.expr.schema as sch
 
 
 class Backend(BaseAlchemyBackend, AlchemyCanCreateSchema):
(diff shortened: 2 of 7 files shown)


docker/trino/catalog/hive.properties
```diff
@@ -0,0 +1,15 @@
...
+hive.metastore.uri=thrift://hive-metastore:9083
+hive.metastore.username=admin
+hive.non-managed-table-writes-enabled=true
+hive.s3.aws-access-key=accesskey
+hive.s3.aws-secret-key=secretkey
+hive.s3.endpoint=http://hive-metastore-minio:9000
+hive.s3.path-style-access=true
+hive.s3select-pushdown.enabled=true
+hive.storage-format=PARQUET
+hive.timestamp-precision=MICROSECONDS
```
cpcloud, line 15: It turns out that you can only have a single timestamp precision for all columns in hive. I'm not sure what's up with that.

ibis/backends/trino/__init__.py
```diff
@@ -112,3 +121,91 @@ def _execute_view_creation(self, name, definition):
...
+                for name, typ in schema.items()
+            )
+            create_stmt += f" ({schema_str})"
+
+        if comment is not None:
+            create_stmt += f" COMMENT {comment!r}"
+
+        if properties:
+            pairs = ", ".join(f"{k}={v!r}" for k, v in properties.items())
+            create_stmt += f" WITH ({pairs})"
```
jcrist, line 171: Do we need any quoting of properties here?

Same question for column/table names above.

jcrist (approved):

ibis/backends/trino/__init__.py
cpcloud: For `WITH` properties, I don't believe so. For column and table names, yes.

ibis/backends/trino/__init__.py
```diff
@@ -112,3 +121,91 @@ def _execute_view_creation(self, name, definition):
...
+        name: str,
+        obj: pd.DataFrame | pa.Table | ir.Table | None = None,
+        *,
+        schema: sch.Schema | None = None,
+        database: str | None = None,
+        temp: bool = False,
+        overwrite: bool = False,
+        comment: str | None = None,
+        properties: Mapping[str, Any] | None = None,
+    ) -> ir.Table:
```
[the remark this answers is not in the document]
cpcloud: Yep.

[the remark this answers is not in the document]
cpcloud: Done.

ibis/backends/trino/__init__.py
cpcloud: Done. I overrode the `_fully_qualified_name` implementation and am quoting everything there.

jcrist: Since the default implementation of `_fully_qualified_name` doesn't do this, won't that potentially have repercussions in other methods? For example, the implementation of `BaseAlchemyBackend.drop_table` uses the fully qualified name in the `_schemas` mapping - moving to quoting that name might cause issues? I'd also worry that some backends handling the quoting in `_fully_qualified_name` (while others don't) might make it harder to know if a name is quoted or not.

cpcloud: Potentially, but we don't have a test case that breaks after changing to this implementation.

Do you have anything in mind?

cpcloud: `_fully_qualified_name` doesn't really have a contract around quoting as far as I can tell. Perhaps we should nail that down soon.

cpcloud: I'm going to have to address this to enable cross schema table access, and I will definitely be changing the implementation of `_fully_qualified_name` to quote everything.

The current base implementation doesn't even do it what it says on the box, it just returns `name`.

cpcloud: I'll propose that this particular thing seem like a thing that is too broken to quibble over.

jcrist: I'm not trying to "quibble over" this, I'm just pointing it out as a potential issue as part of the code review process. If the answer is that it's not an issue or that we don't worry about it now that's fine.

cpcloud: Lemme see if in the meantime I can use the current broken implementation until I get around to address its issues.

cpcloud: Ok, I just removed the use of `_fully_qualified_name` and noted that the `database` argument isn't yet supported.

cpcloud: @jcrist Ah! I think my comment came across wrong! I only meant that I think I have some deeper work around that method to do and that it does need some kind of specification and a bit of thinking aloud that **I** am 100% willing to discard the changes to `_fully_qualified_name`!

jcrist: No worries!

PR #6737, opened 2023-08-01
chore(deps): bring in the remaining wheels to the pymssql lockfile metadata

cpcloud:
This PR brings back wheel installation for pymssql. For some reason these files were missing from the lockfile.

The way I resolved this was by clearing all the poetry caches and then running `poetry lock --no-update`:

```sh
$ poetry cache clear PyPI --all
$ poetry cache clear _default_cache --all
$ poetry lock --no-update
```


jcrist (approved):

gforsyth (approved):
:shipit:

PR #6739, opened 2023-08-01
docs: add docstrings to `DataType.is_*` methods

jcrist:
Just adds a short docstring to each so they show up in the reference docs.


cpcloud (approved):
Thanks!

PR #6740, opened 2023-08-01
docs: document versioning policy

jcrist:
Adds a docs page about our versioning policy.


cpcloud (approved):
LGTM, thanks for doing this!

PR #6741, opened 2023-08-01
feat: add dtype to top level api

jcrist:
This:

- improves the docstring for the `dtype` function
- Adds `dtype` to the public top-level API

Since we already expose `ibis.schema`, having `ibis.dtype` makes sense. It can also be useful when building larger APIs based on ibis, where you might want to coerce something to a `DataType` alone, rather than as part of a schema.

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -41,6 +41,19 @@ def dtype(value: Any, nullable: bool = True) -> DataType:
     Int32(nullable=True)
     >>> ibis.dtype("array<float>")
     Array(value_type=Float64(nullable=True), nullable=True)
+
+    DataType objects may also be created from Python types:
+
+    >>> ibis.dtype(int)
+    Int64(nullable=True)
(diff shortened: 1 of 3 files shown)


ibis/expr/datatypes/core.py
```diff
@@ -22,8 +22,26 @@
...
+    nullable
+        Whether the type should be nullable. Defaults to True.
+
+    Examples
+    --------
+    >>> import ibis
+    >>> ibis.dtype("int32")
+    Int32(nullable=True)
+    >>> ibis.dtype("array<float>")
+    Array(value_type=Float64(nullable=True), nullable=True)
```
kszucs, line 43: Can you add examples about the other overloads?

```py
ibis.dtype(int)
```

```py
class MyStruct:
    a: int
    b: str

ibis.dtype(MyStruct)
```

```py
import pyarrow as pa

ibis.dtype(pa.int32(), nullable=True)
```

kszucs (approved):
LGTM, just a small documentation request.

PR #6743, opened 2023-08-01
refactor(common): remove `ibis.collections.DotDict`

kszucs:
The hybrid getitem/getattr access to the validation context is not
necessary since the move from the previous rule-based validation
to the current pattern matching system.


cpcloud (approved):
Nice!

PR #6744, opened 2023-08-01
refactor(ir): use @annotated decorator to coerce Selection.order_by and Aggregation.order_by arguments

kszucs:
Depends on https://github.com/ibis-project/ibis/pull/6743

Prefer not to manually validate the input arguments.

PR #6745, opened 2023-08-01
refactor(common): remove `Validator` mixin for better clarity

kszucs:
Depends on #6744 

Also removes a redundant function call by directly calling the pattern  match API from the annotation classes.
The previous validation logic called the following methods:

```py
1. Argument.validate()
2. Validator.validate()
3. Pattern.match()
```

which is not reduced to `Argument.validate() -> Pattern.match()`

### DIFF
--- a/ibis/common/annotations.py
+++ b/ibis/common/annotations.py
@@ -8,9 +8,10 @@ from ibis.common.patterns import (
     Any,
     FrozenDictOf,
     Function,
+    NoMatch,
     Option,
+    Pattern,
     TupleOf,
-    Validator,
 )
 from ibis.common.typing import get_type_hints
 
@@ -22,6 +23,10 @@ VAR_KEYWORD = inspect.Parameter.VAR_KEYWORD
 VAR_POSITIONAL = inspect.Parameter.VAR_POSITIONAL
 
 
+class ValidationError(Exception):
+    ...
+
+
 class Annotation:
     """Base class for all annotations.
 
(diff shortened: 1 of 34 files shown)


ibis/common/annotations.py
```diff
@@ -22,64 +23,73 @@
 VAR_POSITIONAL = inspect.Parameter.VAR_POSITIONAL
 
 
+class ValidationError(Exception):
```
kszucs, line 26: We may want to move this exception to `ibis.common.exceptions` but I'm deferring that to a follow-up.

ibis/common/patterns.py
```diff
@@ -293,7 +263,7 @@ def __hash__(self) -> int:
         return self.__precomputed_hash__
 
     def __setattr__(self, name, value) -> None:
-        raise AttributeError("Can't set attributes on immutable ENode instance")
+        raise AttributeError("Can't set attributes on immutable instance")
```
cpcloud, line 266: Micro-est of nits

```suggestion
        raise AttributeError("Can't set attributes on an immutable instance")
```

ibis/common/annotations.py
cpcloud: SGTM!

cpcloud (approved):
Nice refactor.

Glad to see some consolidation here and solid reuse of the new patterns system.

ibis/common/patterns.py
kszucs: Going to fix it in a follow-up.

PR #6752, opened 2023-08-02
fix(backends): make `current_database` implementation and API consistent across all backends

cpcloud:
Make all of the `current_database` implementations consistent with each other.


ibis/backends/bigquery/__init__.py
```diff
@@ -326,7 +326,7 @@ def raw_sql(self, query: str, results=False, params=None):
         return self._execute(query, results=results, query_parameters=query_parameters)
 
     @property
-    def current_database(self) -> str:
+    def current_database(self) -> str | None:
```
cpcloud, line 329: I'll deprecate this in a follow up.

gforsyth (approved):
Looks good!

PR #6760, opened 2023-08-03
refactor(common): factor out base classes to `ibis.common.bases` from `ibis.common.grounds`

kszucs:
this enables us to use the base classes like `Slotted`, `Immutable` and `Singleton`
in `ibis.common.patterns`


ibis/common/bases.py
```diff
@@ -0,0 +1,189 @@
...
+        try:
+            result = self.__cache__[key]
+        except KeyError:
+            result = self.__equals__(other)
+            self.__cache__[key] = result
+
+        return result
+
+
+class Slotted:
```
jcrist, line 154: Should this subclass from Base?

jcrist (approved):
Very much appreciate the new docstrings!

ibis/common/bases.py
jcrist: Reading further I guess this intentional? Can you add a line in the docstrings saying this isn't a subclass of `Base`?

kszucs: Yes, definitely. I've done too much rebasing and it got lost somewhere. Updating.

kszucs: Updated.

PR #6763, opened 2023-08-03
refactor(analysis): remove `find_memtables` function in favor of `node.find()`

kszucs:
The `ibis.expr.analysis.find_memtables(node)` function is semantically equivalent with `node.find(ops.InMemoryTable)`


jcrist (approved):
Nice

PR #6766, opened 2023-08-03
refactor: delete unused schemas cache

jcrist:
As is the `._schemas` cache would only be hit for tables created with `create_table`. Since we don't want to cache metadata queries in case the backing table changes, we drop the cache entirely here.


cpcloud:
Huh, maybe it is being used 😕

jcrist:
Some of the errors make sense (the mssql ones look like potential mismatches in the sql type translation to/from ibis types), but I'm very confused by the postgres timezone errors.

jcrist:
Closing this for now - the current code works fine, we can revisit removing this cache later.

PR #6768, opened 2023-08-03
refactor: add types to Case and Window Builders

NickCrews:
Unrelated to this, but I noticed that the exposed API of the CaseBuilders is probably a little more than we want. We probably just want people to have access to .when(), .else_(), and .end(), but since CaseBuilder inherits from Concrete, users get type hint suggestions for .copy(), and the other Expr internals. But that is separate from this issue

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
     import pandas as pd
     import pyarrow as pa
 
+    import ibis.expr.builders as bl
     import ibis.expr.types as ir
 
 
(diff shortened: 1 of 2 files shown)


ibis/expr/types/generic.py
```diff
@@ -17,6 +17,7 @@
     import pyarrow as pa
 
     import ibis.expr.types as ir
+    import ibis.expr.builders as bl
```
cpcloud, line 20: Is this additional import here necessary?

cpcloud (approved):
LGTM, just wondering about the additional import.

ibis/expr/types/generic.py
NickCrews: Oops, I amended the commit and the point of this import disappeared. In new version it is needed.

kszucs (approved):
Thanks @NickCrews!

Issue #6719, opened 2023-07-28
feat: Sequence comparisons eg Array.Intersection(), Array.jaccard(), String.jaccard(), String.Levenshtein(), etc


NickCrews:
### Is your feature request related to a problem?

I am writing a record linkage framework on top of ibis. Part of this is the need to compare two records to determine if they are the same. Examples of this:

- Take the jaccard similarity of tags, eg ["shoes", "blue"] and ["shoes", "green"]. I can currently do this with some custom functions that use Array.union() and Array.length(), but some backends such as duckdb support jaccard natively so that should be faster. duckdb also supports this with strings as well as arrays, so we could do that here as well. At the least, ibis has Array.union(), so I think for symmetry it would be nice if it had Array.intersection(). Then at the least jaccard() would be trivial for a user, since that is just intersection().length() / union().length() (but still probably a bit less performant than a builtin jaccard).
- Compute the string edit distance such as levenshtein, damerau-levenshtein, jaro-winkler. [Duckdb supports many of these](https://duckdb.org/docs/sql/functions/char.html#text-similarity-functions). I could see how supporting ALL of these could quickly get out of hand since there are so many different variations, but curious what you think.

I can look up other backends if the number of supporting ones is important for you to decide if its worth it.

### Describe the solution you'd like

Order of desire:

- [x] `Array.intersection()`: #6782
- [ ] ~`Array.jaccard()`~
- [x] `String.levenshtein()`: #6773
- [ ] `String.jaccard`

### What version of ibis are you running?

main

### What backend(s) are you using, if any?

duckdb, some other backends support this but not all.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jcrist:
Hmmm, I wonder if these would be a good use case for our new UDF support. This isn't finished yet, but there's a future where the following would work:

```python
con  = ibis.duckdb.connect()

# This would query the duckdb engine to automatically determine the signature of `jaccard`,
# so we can error nicely on misuse at expression build time rather than at run time.
# This works in some backends, but not in duckdb yet.
jaccard = con.function("jaccard")

t = con.table(...)
res = jaccard(t.x, t.y)

# We're also thinking about making a general way to define udf functions as expressions alone,
# which would make the expression-level versions backend agnostic. Something like (don't focus
# too much on the spelling).

@ibis.udf.builtin
def jaccard(x: str, y: str) -> float:
    ...

[...]

PR #6773, opened 2023-08-04
feat(api): add `levenshtein` edit distance API

cpcloud:
This PR adds a `levenshtein` method to strings for computing the edit distanceb
between two strings. Open to changing the name, as `levenshtein` is quite long.
The main reason I went with it is that something like `edit_dist` seems too
vague given there are multiple methods that "edit distance" can refer to.

xref #6719


kszucs (approved):

ibis/backends/tests/test_string.py
```diff
@@ -1023,3 +1023,28 @@ def test_multiple_subs(con):
...
+        "mssql",
+        "mysql",
+        "pandas",
+        "polars",
+        "sqlite",
+    ],
+    raises=com.OperationNotDefinedError,
+)
+@pytest.mark.parametrize(
+    "right", ["sitting", ibis.literal("sitting")], ids=["python", "ibis"]
```
NickCrews, line 1044: @cpcloud IDK if you want to be this careful, but there is a very similar Damerau-Levenshtein metric that counts transpositions as one edit, so maybe we want to verify that the backend is using the levenshtein, and not damerau-levenshtein

levenshten("iktton", "kitten") -> 3 
damerau_levenshten("iktton", "kitten") -> 2 (swap ik, then change e to o)

cpcloud: @NickCrews Yeah! I think we can always add the Damerau-Levenshtein API later if desired.

mesejo: I realize I'm super late for this discussion, and I agree that the name `edit_dist` is vague, but perhaps we could use it with an optional parameter `type`? I'm thinking of:

`edit_distance(type='levenshtein') `

The possible options for type would be backend-dependent. 

This has at least two points in favor:

- Will allow for the inclusion of backends that do not define Levenshtein, for example (pandas with the built-in Ratcliff-Obershelp distance from [difflib](https://docs.python.org/3/library/difflib.html)) 
- For backends that support more than one type (like [DuckDB](https://duckdb.org/docs/sql/functions/char.html#text-similarity-functions)) allow the user to select the type without creating additional operations.

PR #6782, opened 2023-08-04
feat(api): add `ArrayIntersect` operation and corresponding `ArrayValue.intersect` API

cpcloud:
This PR adds an `intersection` method to arrays.

### DIFF
--- a/ibis/expr/types/arrays.py
+++ b/ibis/expr/types/arrays.py
@@ -725,7 +725,48 @@ class ArrayValue(Value):
         """
         return ops.ArrayUnion(self, other).to_expr()
 
-    def zip(self, other: ir.Array, *others: ir.Array) -> ir.Array:
+    def intersect(self, other: ArrayValue) -> ArrayValue:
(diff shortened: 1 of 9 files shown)


ibis/expr/types/arrays.py
```diff
@@ -725,7 +725,48 @@ def union(self, other: ir.ArrayValue) -> ir.ArrayValue:
         """
         return ops.ArrayUnion(self, other).to_expr()
 
-    def zip(self, other: ir.Array, *others: ir.Array) -> ir.Array:
+    def intersection(self, other: ArrayValue) -> ArrayValue:
```
kszucs, line 728: I have a slight preference to call both the API method and the operation it intersect instead of intersection.

kszucs (approved):

ibis/expr/types/arrays.py
cpcloud: Ah, definitely! We should be consistent with `Table`, which def `intersect`

PR #6784, opened 2023-08-05
fix(snowflake): use ibis-defined `array_sort` until upstream lands

cpcloud:
Looks like `array_sort` is not actually release upstream. I define `array_sort` here as a JavaScript UDF in the meantime.


cpcloud:
Merging to get upstream passing again.

PR #6791, opened 2023-08-06
refactor(analysis): rewrite substitute_unbound to use the new pattern system

kszucs:


ibis/expr/analysis.py
```diff
@@ -167,15 +170,9 @@ def fn(node):
...
-
-    def fn(node, _, **kwargs):
-        if isinstance(node, ops.DatabaseTable):
-            return ops.UnboundTable(name=node.name, schema=node.schema)
-        else:
-            return node.__class__(**kwargs)
-
-    return node.map(fn)[node]
+    return node.replace(
+        p.DatabaseTable(name=x, schema=y) >> c.UnboundTable(name=x, schema=y)
```
cpcloud, line 174: Can we instantiate the pattern variables in this function's scope instead of globally, and also call them `name` and `schema`?

cpcloud (approved):
This looks pretty awesome.

ibis/expr/analysis.py
kszucs: We could just define the two variables in the function or we could use capture and variable object directly in the pattern too. Matter of preference mostly, but we can revisit it later.

PR #6792, opened 2023-08-06
fix(snowflake): fix timestamp scale inference

cpcloud:
This PR fixes an issue with some logic that was incorrectly assuming that if a precision and scale were
specified in a [`ResultMetadata`](https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-api#label-python-connector-resultmetadata-object) object, then the type was unambiguously a numeric type because that's what
the documentation says.

The documentation seems to be wrong in practice, because timestamp metadata comes along with a `precision`
equal to 0 and a non-zero `scale`.

Unfortunately, this information is completely missing from timestamp type modeling in `snowflake-sqlalchemy`.

This means that type reflection from sqlalchemy for timestamps doesn't match that of what's reported by
Snowflake itself, so until we move away from `snowflake-sqlalchemy` there'll be some inconsistency in the
inferred column types of timestamp columns.

This fact is reflected in the tests added in this PR.

### DIFF
--- a/ibis/backends/snowflake/tests/test_datatypes.py
+++ b/ibis/backends/snowflake/tests/test_datatypes.py
@@ -1,0 +1,158 @@
+from __future__ import annotations
+
+import pytest
+from pytest import param
+
+import ibis
+import ibis.expr.datatypes as dt
+from ibis.backends.snowflake.datatypes import parse
+from ibis.backends.snowflake.tests.conftest import _get_url
+from ibis.util import gen_name
+
+dtypes = [
+    ("FIXED", dt.int64),
+    ("REAL", dt.float64),
+    ("TEXT", dt.string),
+    ("DATE", dt.date),
+    ("TIMESTAMP", dt.Timestamp(scale=9)),
+    ("VARIANT", dt.json),
+    ("TIMESTAMP_LTZ", dt.Timestamp(scale=9)),
+    ("TIMESTAMP_TZ", dt.Timestamp(timezone="UTC", scale=9)),
+    ("TIMESTAMP_NTZ", dt.Timestamp(scale=9)),
+    ("OBJECT", dt.Map(dt.string, dt.json)),
+    ("ARRAY", dt.Array(dt.json)),
+    ("BINARY", dt.binary),
+    ("TIME", dt.time),
+    ("BOOLEAN", dt.boolean),
+]
+
+
+@pytest.mark.parametrize(
+    ("snowflake_type", "ibis_type"),
+    [
+        param(snowflake_type, ibis_type, id=snowflake_type)
+        for snowflake_type, ibis_type in dtypes
+    ],
+)
+def test_parse(snowflake_type, ibis_type):
+    assert parse(snowflake_type.upper()) == ibis_type
+
+
+@pytest.fixture(scope="module")
+def con():
+    return ibis.connect(_get_url())
+
+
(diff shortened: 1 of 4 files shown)


cpcloud:
Snowflake backend tests look good:

```
…/ibis on  fix-snowflake-precision-inference is 📦 v6.1.0 via 🐍 v3.10.12 via ❄️  impure (ibis-3.10.12-env)
❯ pytest -m snowflake -n auto --dist loadgroup -q
bringing up nodes...
.s..................x........x.x............xx..........x..x..........xx..x........x......x.................x.........................x...xx.....x..x..x.x....x............x....x...x.xx...... [ 15%]
x..........xxx.xxxxxxxxxxxxx.xx....x.....xx................x.............x...x..................x.........x................x............................................x.......xx...x........ [ 30%]
...................x................................x...x....................xx.s.x....x...x.s.......s........x..............................x...................x.........x.................. [ 46%]
.x.......x...x........x...........x....x.................x...x..........x.....x.........................x.xxx.........xxxx.x.x..x...........x.x.x..x......................x......x........x... [ 61%]
.x............x....................xx....x.....x.x....x............................................................................xx........x...xxxx..xx...........xx.....x....x..xs.s..sx... [ 77%]
.s.......x...x.......x.x..........xx...x......s........x.............x........x.......................x......xx...x.....xx..xxx............................................................... [ 92%]
..................xx....x......xx.s....x..x....xx.x..xxx.x.xx..xx.xxx...x...xxxx..xxxxxxxxxx                                                                                                   [100%]
1046 passed, 10 skipped, 176 xfailed in 282.41s (0:04:42)
```

cpcloud:
Once we move away from `snowflake-sqlalchemy` we can remove any workarounds and inconsistencies in this PR.

ibis/backends/snowflake/tests/test_datatypes.py
```diff
@@ -0,0 +1,158 @@
...
+]
+
+
+@pytest.mark.parametrize(
+    ("snowflake_type", "ibis_type"),
+    [
+        param(snowflake_type, ibis_type, id=snowflake_type)
+        for snowflake_type, ibis_type in dtypes
+    ],
+)
```
kszucs, line 36: You could use `pytest.mark.parametrize(..., ids=callable)` to generate the labels, see https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids

ibis/backends/snowflake/tests/test_datatypes.py
```diff
@@ -0,0 +1,158 @@
...
+    ("DECIMAL", dt.int64),
+    ("NUMERIC", dt.int64),
+    ("NUMBER(5)", dt.int64),
+    ("DECIMAL(5, 2)", dt.Decimal(5, 2)),
+    ("NUMERIC(21, 17)", dt.Decimal(21, 17)),
+    ("INT", dt.int64),
+    ("INTEGER", dt.int64),
+    ("BIGINT", dt.int64),
+    ("SMALLINT", dt.int64),
+    ("TINYINT", dt.int64),
```
kszucs, line 57: Shoudln't we parse integer types with the right bitwidth here?

cpcloud: No, because that's not snowflake's behavior.

kszucs (approved):

PR #6798, opened 2023-08-07
fix(struct-column): make `ops.StructColumn` dshape depend on its input

cpcloud:
Fixes an incorrect modeling of the `StructColumn` operation dshape: it should depend on the dshape of the inputs.

### DIFF
--- a/ibis/expr/operations/structs.py
+++ b/ibis/expr/operations/structs.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from public import public
 
-import ibis.expr.datashape as ds
 import ibis.expr.datatypes as dt
 import ibis.expr.rules as rlz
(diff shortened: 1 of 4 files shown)


kszucs (approved):

ibis/expr/operations/structs.py
```diff
@@ -33,7 +32,7 @@ class StructColumn(Value):
     names: VarTuple[str]
     values: VarTuple[Value]
 
-    shape = ds.columnar
+    shape = rlz.shape_like("values")
```
kszucs, line 35: Could you please add a simple operations test for it under `ibis/expr/operations/test_structs.py`?

cpcloud: Done.

PR #6804, opened 2023-08-08
fix(histogram): ensure that the bin width calculation matches numpy

NickCrews:
.histogram has a serious bug, where the bin width is calculated wrong. If you do the same calculation with np.histogram(), you get very different results. I added a fix for this.

But this still doesn't solve another bug:

`ibis.examples.penguins.fetch().bill_depth_mm.histogram(5).name("hist").value_counts().order_by("hist")`

gives 
```
┏━━━━━━━┳━━━━━━━━━━━━┓
┃ hist  ┃ hist_count ┃
┡━━━━━━━╇━━━━━━━━━━━━┩
│ int64 │ int64      │
├───────┼────────────┤
│     0 │         56 │
│     1 │         66 │
│     2 │         99 │
│     3 │         95 │
│     4 │         25 │
│     5 │          1 │
│  NULL │          2 │
└───────┴────────────┘
```

Note how the max value is assigned the bin `5`. That isn't right, the bins should only be labeled 0 to 4.

Seeing as how no one is noticing that this is broken, can we just totally update this API? Can we make it the same as numpy's? eg it returns bin edges, and supports many other nice arguments. Or just delete it or merge it with bucket()?

### DIFF
--- a/ibis/backends/tests/test_numeric.py
+++ b/ibis/backends/tests/test_numeric.py
@@ -1371,8 +1371,10 @@ def test_clip(backend, alltypes, df, ibis_func, pandas_func):
 )
 def test_histogram(con, alltypes):
     n = 10
-    results = con.execute(alltypes.int_col.histogram(n).name("tmp"))
-    assert len(results.value_counts()) == n
+    hist = con.execute(alltypes.int_col.histogram(n).name("hist"))
+    vc = hist.value_counts().sort_index()
+    vc_np, _bin_edges = np.histogram(alltypes.int_col.execute(), bins=n)
+    assert vc.tolist() == vc_np.tolist()
 
 
 @pytest.mark.parametrize("const", ["pi", "e"])
--- a/ibis/expr/types/numeric.py
+++ b/ibis/expr/types/numeric.py
@@ -1032,17 +1032,13 @@ class NumericColumn(Column, NumericValue):
             )
 
         if binwidth is None or base is None:
-            import ibis
-
             if nbins is None:
                 raise ValueError("`nbins` is required if `binwidth` is not provided")
 
-            empty_window = ibis.window()
-
             if base is None:
-                base = self.min().over(empty_window) - eps
+                base = self.min() - eps
 
-            binwidth = (self.max().over(empty_window) - base) / (nbins - 1)
+            binwidth = (self.max() - base) / nbins
 
         return ((self - base) / binwidth).floor()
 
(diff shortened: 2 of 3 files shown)


ibis/backends/tests/test_numeric.py
```diff
@@ -1371,8 +1371,10 @@ def test_clip(backend, alltypes, df, ibis_func, pandas_func):
 )
```
NickCrews, line 1371: unrelated: I think we can remove the @pytest.mark.broken("druid")?

ibis/expr/types/numeric.py
```diff
@@ -1042,7 +1042,7 @@ def histogram(
             if base is None:
                 base = self.min().over(empty_window) - eps
```
NickCrews, line 1043: Why do we need this window here? can't we do a simple min() and max()?

ibis/backends/tests/test_numeric.py
cpcloud: Does it pass for you? We can only remove it if the test failure says `XPASS` in it, meaning it's actually **not** failing.

cpcloud:
@NickCrews Nice find. Thanks for submitting a PR to fix the issue.

Is there something not sufficient about your fix? Your commit message `sorta fix histogram` suggests that there's something else wrong with the implementation. If so, let's see if we can address here and give a slightly more confidence-inspiring commit message 😂

ibis/expr/types/numeric.py
cpcloud: Made these changes. Great suggestion, cleaned up more code than I was expecting!

ibis/backends/tests/test_numeric.py
NickCrews: Something is broken about my local testing environment and I haven't taken the time to fix it so I haven't actually run these. Just trying to rely on CI

NickCrews:
So my one existing commit does fix the bin with issue, so yeah, we probably should edit the commit message to be more confidence. Inspiring.

I named it that because the other bug that I described in the PR still exists. But we could make that be in another PR

cpcloud:
@NickCrews How would you like to proceed here?

NickCrews:
I'm about to be out of office until next Wednesday so I won't be that helpful until then. I would say adjust the commit message and merge this just to get some progress out of the way, and then let's file an issue for the n plus one bug. But at least this PR gets us closer to being fixed.

cpcloud (approved):

PR #6807, opened 2023-08-08
fix(api): fail on trying to construct an iterable of a deferred object

cpcloud:
This PR fixes an issue where non-ibis APIs that depend on implicit
implementations of `__iter__` that use `__getitem__` would never terminate due
to `Deferred` instances implementing an always-succeeding version of
`__getitem__`.

The main downside is that because `Deferred` now implements `__iter__` it
appears to be iterable, so we have to exclude it from `is_iterable` instances
checks explicitly.

### DIFF
--- a/ibis/util.py
+++ b/ibis/util.py
@@ -217,7 +217,15 @@ def is_iterable(o: Any) -> bool:
     >>> is_iterable([])
     True
     """
-    return not isinstance(o, (str, bytes)) and isinstance(o, collections.abc.Iterable)
+    if isinstance(o, (str, bytes)):
+        return False
+
+    try:
+        iter(o)
+    except TypeError:
+        return False
+    else:
+        return True
 
 
 def convert_unit(value, unit, to, floor: bool = True):
--- a/ibis/expr/deferred.py
+++ b/ibis/expr/deferred.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import operator
-from typing import Any, Callable
+from typing import Any, Callable, NoReturn
 
 _BINARY_OPS: dict[str, Callable[[Any, Any], Any]] = {
     "+": operator.add,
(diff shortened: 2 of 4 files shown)


ibis/util.py
```diff
@@ -217,7 +217,11 @@ def is_iterable(o: Any) -> bool:
     >>> is_iterable([])
     True
     """
-    return not isinstance(o, (str, bytes)) and isinstance(o, collections.abc.Iterable)
+    from ibis.expr.deferred import Deferred
```
kszucs, line 220: Could we check for deferreds before calling `is_iterable()` in the call site instead? I assume it is checked in `bind_expr()`?

cpcloud: It's used in a bunch of places. It's a lot more fragile than to check it in the core method that's powering all the call sites.

cpcloud: What's the issue with having it here? The import?

kszucs: I have two alternative implementation suggestions:
1. investigate the call sites and first check for deferreds
2. have a `NotIterable` ABC defined here, and extend that from `Deferred`, we could also add `NotIterable.register(str|bytes)`

cpcloud: I don't think it indicates a design issue at the call site as much as a limitation of using duck typing to define an interface: you can't rely on method **existence** to accurately determine whether a thing implements a protocol.

cpcloud: In any case, it seems strictly worse to introduce a bunch of call-site checks that are likely to be forgotten.

kszucs: I agree, though we could also follow the `try first` philosophy instead of checking with `is_iterable()`, that way deferred would clearly raise.

kszucs: > In any case, it seems strictly worse to introduce a bunch of call-site checks that are likely to be forgotten.

How about option 2?

cpcloud: Seems fine! Lemme implement it.

kszucs (approved):

PR #6808, opened 2023-08-08
feat(api): add `shape` property for computing the number of rows and columns

cpcloud:
This PR adds a `shape` property.

I'm not convinced this is something we should do, and if there's enough -1s
I'd be happy to close it.

The main downside I can think of is introducing an API that looks like the API
from other libraries (numpy, pandas) that is to be fast in those libraries, but
can be arbitrarily slow in ibis due to the way a backend implements `count` as
well, the query being counted, the deferred nature of ibis and the streaming
nature of many of our backends.

I'm not 100% convinced this is a deal breaker as the convenience may outweigh
this behavior.

After all if someone needs to compute these values, they'll still need to call
`count` on the expression, so why not make that easier to write?


kszucs:
I'm not against exposing a shape information, but I'd prefer doing that lazily since all the expression methods do lazy computation, so I'm mostly against of eagerly computed expression attributes and methods of all kinds.

cpcloud:
The utility of a lazy `shape` is extremely limited IMO.

cpcloud:
Another option is to cache the output of `shape`, which amortizes the cost over repeated calls, perhaps making it viable as an eager API.

kszucs:
I still dislike that the property hides the fact that a computation is being triggered. Something like `expr.compute_shape()` would clearly highlight the action, but in that case I'd prefer having `expr.shape.execute()`.

jcrist:
-1. I agree with @kszucs here - a property that may execute an expensive computation on access feels like a potential footgun to me. One potential failure mode - some tab-completion systems will assume properties are cheap and attempt to execute `.shape` accidentally leading to user's terminal sessions locking up or lagging  (I've run into this with IPython in the past).

There's also a consistency argument here - right now all execution is explicitly marked (excluding interactive repr or `__array__`/`__dataframe__` protocol. Supporting `.shape` here would diverge from that, possibly defying user expectations. We already explicitly don't implement other methods that would require execution (`__len__` for example raises a nice error saying to use `.count()` instead).

> The utility of a lazy `shape` is extremely limited IMO.

I'd also agree with this. My vote would be to just not implement `.shape` at all.

cpcloud:
Alright, closing! Thanks for the review @kszucs @jcrist!

PR #6811, opened 2023-08-08
fix(api): allow `ArrayValue.__iter__` again

cpcloud:
Apparently iterating over array columns was allowed, and used in the BigQuery test suite.


kszucs:
Shall we allow this without knowing the array length? What happens if someone calls 'list(arrayval)'?

cpcloud:
> Shall we allow this without knowing the array length? What happens if someone calls 'list(arrayval)'?

It never terminates, which was the behavior before

cpcloud:
Hm, maybe this is incompatible with our current use of `promote_list` everywhere.

kszucs (approved):

kszucs (approved):

kszucs (approved):

PR #6818, opened 2023-08-09
feat(api): add `relocate` table expression API for moving columns around based on selectors

### LINKED ISSUE #6803: feat: `Table.relocate` method
often a user will want to change the order of columns, see linked discussion

---

We could consider adding a `Table.relocate` method, similar to [R's method of the same name](https://dplyr.tidyverse.org/reference/relocate.html).

_Originally posted by @cpcloud in https://github.com/ibis-project/ibis/discussions/6795#discussioncomment-6662292_

cpcloud:
This PR adds a new `relocate` method on `Table` expressions.

The implementation is a near exact clone of the [`dplyr` implementation](https://dplyr.tidyverse.org/reference/relocate.html).

The tests are also based on the [`dplyr` test suite](https://github.com/tidyverse/dplyr/blob/c963d4db08ab4b4e7f75e3b99b04ef5f6e903a25/tests/testthat/test-relocate.R) for this feature.

Closes #6803.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -3609,6 +3609,189 @@ class Table(Expr, _FixedTextJupyterMixin):
 
         return self.group_by(id_cols).aggregate(**aggs)
 
+    def relocate(
+        self,
+        *columns: str | s.Selector,
+        before: str | s.Selector | None = None,
+        after: str | s.Selector | None = None,
+        **kwargs: str,
+    ) -> Table:
+        """Relocate `columns` before or after other specified columns.
+
+        Parameters
+        ----------
+        columns
+            Columns to relocate. Selectors are accepted.
+        before
+            A column name or selector to insert the new columns before.
+        after
+            A column name or selector. Columns in `columns` are relocated after the last
+            column selected in `after`.
+        kwargs
+            Additional column names to relocate, renaming argument values to
+            keyword argument names.
+
+        Returns
+        -------
+        Table
+            A table with the columns relocated.
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> import ibis.selectors as s
+        >>> t = ibis.memtable(dict(a=[1], b=[1], c=[1], d=["a"], e=["a"], f=["a"]))
+        >>> t
+        ┏━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┓
+        ┃ a     ┃ b     ┃ c     ┃ d      ┃ e      ┃ f      ┃
+        ┡━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━┩
+        │ int64 │ int64 │ int64 │ string │ string │ string │
+        ├───────┼───────┼───────┼────────┼────────┼────────┤
+        │     1 │     1 │     1 │ a      │ a      │ a      │
+        └───────┴───────┴───────┴────────┴────────┴────────┘
+        >>> t.relocate("f")
+        ┏━━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━━┳━━━━━━━━┓
+        ┃ f      ┃ a     ┃ b     ┃ c     ┃ d      ┃ e      ┃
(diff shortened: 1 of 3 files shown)


ibis/expr/types/relations.py
```diff
@@ -3611,6 +3611,189 @@ def pivot_wider(
...
+        before: str | s.Selector | None = None,
+        after: str | s.Selector | None = None,
+        **kwargs: str | s.Selector,
+    ) -> Table:
+        """Relocate columns in a table.
+
+        Parameters
+        ----------
+        selectors
+            Selectors to relocate.
```
jcrist, line 3626: What do you think about naming this argument `columns` instead of selectors? My initial expectation on seeing an arg named `selectors` was that this probably doesn't support column names directly. We already have several places where selectors can be passed in where columns are supported (making that use case more expected). Calling it `columns` also makes describing the function easier ("relocate takes one or more columns (or selections of columns) and relocates them before or after other specified columns").

If you want to keep the name as is, adding `column name or selector` here as you've done below for `before`/`after` would help.

ibis/expr/types/relations.py
```diff
@@ -3611,6 +3611,189 @@ def pivot_wider(
...
+        before
+            A column name or selector to insert the new columns before.
+        after
+            A column name or selector to insert the new columns after.
```
jcrist, line 3630: Can you add a clarification here on what before/after a selector means? It's clear if the selector returns one column, but unclear what it means if it returns more than one. Perhaps something like

```suggestion
            A column name or selector. Columns in `columns` are relocated after the last
            column selected in `after`.
```

ibis/expr/types/relations.py
```diff
@@ -3611,6 +3611,189 @@ def pivot_wider(
 
         return self.group_by(id_cols).aggregate(**aggs)
 
+    def relocate(
+        self,
+        *selectors: str | s.Selector,
+        before: str | s.Selector | None = None,
+        after: str | s.Selector | None = None,
+        **kwargs: str | s.Selector,
```
jcrist, line 3619: How would a selector work here in kwargs? What's the meaning of:

```python
t.relocate(after="x", colname=s.of_type("float"))
```

ibis/expr/types/relations.py
cpcloud: `columns` makes sense to me

ibis/expr/types/relations.py
cpcloud: Yeah, that's definitely a mistake on my part. I'll remove the `s.Selector` annotation here.

cpcloud:
@jcrist This is ready for another round of review when you get a chance.

ibis/expr/types/relations.py
```diff
@@ -3609,6 +3609,189 @@ def pivot_wider(
...
+
+        for name, sel in itertools.chain(
+            zip(itertools.repeat(None), map(s._to_selector, columns)),
+            zip(kwargs.keys(), map(s._to_selector, kwargs.values())),
+        ):
+            for pos in sel.positions(self):
+                if pos in sels:
+                    # make sure the last duplicate column wins by reinserting
+                    # the position if it already exists
+                    del sels[pos]
```
gforsyth, line 3767: I could definitely be missing something, but isn't this redundant? `sels[pos]` is going to be overwritten on the next line.

jcrist: I *think* this is to reorder the selections. The dict ordering behavior in python is that updates don't reorder, but a del and reinsert will.

```python
In [3]: d = {"a": 1, "b": 2, "c": 3}

In [4]: list(d)
Out[4]: ['a', 'b', 'c']

In [5]: d["b"] = 4

In [6]: list(d)  # order is still the same
Out[6]: ['a', 'b', 'c']

In [7]: del d["b"]

In [8]: d["b"] = 4

In [9]: list(d)  # b is now last
Out[9]: ['a', 'c', 'b']
```

jcrist (approved):

ibis/expr/types/relations.py
gforsyth: ahh, gotcha.  thanks for the explainer!  I missed that it was using the key-ordering to map the column-ordering, but that makes sense.

NickCrews:
This looks lovely, thank you @cpcloud @jcrist @gforsyth !

PR #6820, opened 2023-08-09
feat(ux): promote lists of strings to `any_of` selectors

cpcloud:
This PR adds promotion of lists of column names to `s.any_of(s.c(col0), s.c(col1), ...)` selectors, which allows specifying them wherever you can specifyc `s.c(...)`.

### DIFF
--- a/ibis/selectors.py
+++ b/ibis/selectors.py
@@ -56,6 +56,7 @@ from typing import Callable, Iterable, Mapping, Optional, Sequence, Union
 
 from public import public
 
+import ibis.common.exceptions as exc
 import ibis.expr.datatypes as dt
 import ibis.expr.types as ir
 from ibis import util
@@ -366,7 +367,16 @@ def all_of(*predicates: str | Predicate) -> Predicate:
 def c(*names: str) -> Predicate:
     """Select specific column names."""
     names = frozenset(names)
-    return where(lambda col: col.get_name() in names)
+
+    def func(col: ir.Value) -> bool:
+        schema = col.op().table.schema
+        if extra_cols := (names - schema.keys()):
+            raise exc.IbisInputError(
+                f"Columns {extra_cols} are not present in {schema.names}"
+            )
+        return col.get_name() in names
+
+    return where(func)
 
 
 class Across(Selector):
--- a/ibis/selectors.py
+++ b/ibis/selectors.py
@@ -351,15 +351,15 @@ def matches(regex: str | re.Pattern) -> Selector:
 
 
 @public
-def any_of(*predicates: Predicate) -> Predicate:
+def any_of(*predicates: str | Predicate) -> Predicate:
     """Include columns satisfying any of `predicates`."""
-    return functools.reduce(operator.or_, predicates)
+    return functools.reduce(operator.or_, map(_to_selector, predicates))
 
 
 @public
-def all_of(*predicates: Predicate) -> Predicate:
+def all_of(*predicates: str | Predicate) -> Predicate:
     """Include columns satisfying all of `predicates`."""
-    return functools.reduce(operator.and_, predicates)
+    return functools.reduce(operator.and_, map(_to_selector, predicates))
 
 
 @public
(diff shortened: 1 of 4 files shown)


cpcloud:
> Can you comment with an example of when this would come up?

The doctest example in `pivot_wider`:

```
        >>> us_rent_income.pivot_wider(
        ...     names_from="variable",
        ...     names_sep=".",
        ...     values_from=("estimate", "moe"),
        ... )
```

`values_from` can be a list of selectors or a list of strings. They're collapsed into a single selector with `any_of`.

> Also, I'd expect selectors to treat ["col1", "col2"] the same as c("col1", "col2")

That's what happening here, or at least it's my intent. Where do you see something different?

>  I find the use of any_of here a bit surprising. Can you comment on why you went with this approach?

Because that was the easiest way to map to the semantics I want here: a list of selectors means "columns matching any of the input selectors", similar to how `s.c(...)` means "columns whose name matches any of `...`"

cpcloud:
Maybe it's the fact that `s.c` doesn't select things in the order of its inputs, but behaves like a set and preserves the order of the columns its applied to?

jcrist:
> That's what happening here, or at least it's my intent. Where do you see something different?

I was thinking about the behavior when a specified column is missing (where `any_of` would by design ignore a selector that doesn't match), but it looks like `selectors.c` has the same behavior - the specified name is silently dropped?

```python
# errors, since `islad` is missing (typo for `island`)
>>> t.select(["species", "islad"])

# silently only selects `species` (should we change this?)
>>> t.select(s.c("species", "islad"))

# also only selects `species`, but the `any_of` makes it clearer that unmatching selectors are ignored
>>> t.select(s.any_of(s.c("species"), s.c("islad")))
```

The user experience I'd expect with a specified list of column names is that if I typo one of them I'll get an error rather than it being ignored. This is what happens with lists of columns in other parts of ibis, I'd expect selectors to have the same behavior.

jcrist:
That wouldn't have the same behavior though, since `any_of(c("col1"), c("col2"))` would by nature still silently ignore `col2` not existing in the input table. I think to get the behavior I'm describing we'll need to:

- Coerce lists of strings to `c(*list_of_strs)`
- Coerce lists of selectors into `any_of(*list_of_selectors)`
- Modify `c` to error if a specified column isn't in the input table

Perhaps we're talking past each other here? Can you describe the intended behavior you'd like for a selectors with a list of columns like `["exists_in_table", "missing_from_table"]`? How would your proposed `any_of` solution solve this problem?

cpcloud:
Pushed up a second commit to illustrate what I said.

jcrist:
Ah, makes sense. I think I was fixated on `any_of` ignoring selectors that don't match anything in its input and missed that `s.c` could still just raise an error as part of its normal execution.

ibis/selectors.py
```diff
@@ -366,7 +367,16 @@ def all_of(*predicates: str | Predicate) -> Predicate:
...
     names = frozenset(names)
-    return where(lambda col: col.get_name() in names)
+
+    def func(col: ir.Value) -> bool:
+        schema = col.op().table.schema
+        if extra_cols := (names - schema.keys()):
+            raise exc.IbisInputError(
+                f"Columns {extra_cols} are not present in {schema.names}"
+            )
+        return col.get_name() in names
```
jcrist, line 377: This might be cleaner/more efficient to implement `c` directly as a subclass of `Selector`? This way you could define `expand` directly, which would give direct access to the input table rather than having to hop through `.op().table`. The column names existing check would also only happen once rather than `len(table.columns)` times.

cpcloud: That is a much better idea than what I am currently doing, thank you.

cpcloud: Hm, this approach turns out to break composition of selectors with `~`, `&` and `|` :(

cpcloud: @jcrist Not sure if you have any other ideas here, but I'm inclined to merge this as it is.

jcrist: That's unfortunate. I think this should be doable as I suggested above, but it would require refactoring how the composition operators work. Going with what you have seems reasonable for now.

PR #6825, opened 2023-08-10
fix(trino): fix `list_databases` to show all catalogs and `list_schemas` to allow the `database` argument

cpcloud:
This PR fixes the trino implementations of `list_databases` (only lists the current catalog instead of all available catalogs), and `list_schemas` (incorrect quoting of the schema in the base sqlalchemy implementation).


docker/trino/catalog/hive.properties
```diff
@@ -1,6 +1,7 @@
 connector.name=hive
 
 hive.allow-drop-table=true
+hive.allow-rename-table=true
```
cpcloud, line 4: Moving this change to a separate PR

cpcloud: #6826

kszucs (approved):

PR #6827, opened 2023-08-10
feat(trino): add support for cross-schema and cross-catalog table operations

cpcloud:
This PR adds support for cross-schema and cross-catalog table operations to the trino backend. I pulled out the logic to access metadata into a specialized SQLAlchemy backend that is shared with the Snowflake backend due to their near-identical logic for achieving the same result.

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -877,3 +877,83 @@ class BaseAlchemyBackend(BaseSQLBackend):
 
         with self.begin() as con:
             con.execute(view)
+
+
+class AlchemyCrossSchemaBackend(BaseAlchemyBackend):
+    """A SQLAlchemy backend that supports cross-schema queries.
+
+    This backend differs from the default SQLAlchemy backend in that it
+    overrides `_get_sqla_table` to potentially switch schemas during table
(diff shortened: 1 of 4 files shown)


jcrist (approved):
One small request for a docstring, otherwise LGTM

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -885,3 +885,75 @@ def drop_view(
 
         with self.begin() as con:
             con.execute(view)
+
+
+class AlchemyCrossSchemaBackend(BaseAlchemyBackend):
```
jcrist, line 890: Would you mind adding short (possibly just one line) docstrings on:

- the class, indicating how this differs from the `BaseAlchemyBackend`
- The asbtractmethod below, indicating what a subclass should do to satisfy the contract.

jcrist (approved):
Gah, forgot to actually submit the comment :/

ibis/backends/base/sql/alchemy/__init__.py
cpcloud: Done!

cpcloud:
I caused a regression in the snowflake backend in this PR, fix is on the way!

PR #6830, opened 2023-08-10
chore(benchmarks): require a benchmark run number when invoking `just benchcmp`

cpcloud:
Require a benchmark run number when running `just benchcmp` to avoid spurious comparisons to benchmark runs on the same commit.


kszucs (approved):
Thanks!

PR #6831, opened 2023-08-10
perf(snowflake): speed up metadata accesses when database and schema match the connection

cpcloud:
This PR adds `just profile` command that runs
[`pyinstrument`](https://github.com/joerick/pyinstrument).

I then used that command to find some unnecessary transactions (in
`current_database` and `current_schema`) as well as unnecessary additional
invocations of those properties to shave off ~25% on `_get_sqla_table` in the
snowflake backend, which is an often called function in all of our
SQLAlchemy-based backends.


kszucs (approved):
Niice!

cpcloud:
😬 I should've probably posted the snowflake tests results 😬

PR #6834, opened 2023-08-10
perf(duckdb): speed up performance with wide tables

### LINKED ISSUE #6832: perf(sqlalchemy): wide table experience suffers from performance problems
### What happened?

When playing with the World Health Organization data on world heathcare cost and expenditure, I noticed that the interactive experience suffers from performance problems.

The table in question has 4224 rows and 3220 columns.

I isolated this problem to ibis with this exploration.

First, I did the simple thing and used `ibis.read_csv` with the default backend, DuckDB.

```

In [2]: %time t = ibis.read_csv("/home/cloud/downloads/GHED_data.csv")
CPU times: user 4.25 s, sys: 2.31 s, total: 6.56 s
Wall time: 4.14 s
```

Then I looked at DuckDB's time for this using `EXPLAIN ANALYZE`:

```
D explain analyze select * from read_csv_auto('/home/cloud/downloads/GHED_data.csv');

│┌───────────────────────────┐│
│└───────────────────────────┘│
┌─────────────────────────────────────┐
│┌───────────────────────────────────┐│
││    Query Profiling Information    ││
│└───────────────────────────────────┘│
└─────────────────────────────────────┘
explain analyze select * from read_csv_auto('/home/cloud/downloads/GHED_data.csv');
┌─────────────────────────────────────┐
│┌───────────────────────────────────┐│
││         Total Time: 0.758s        ││
│└───────────────────────────────────┘│
└─────────────────────────────────────┘
```

So, ibis is already taking about 5.5x longer to construct a view, not great.

My hunch is that this is due to extracting type information for all 3k columns in the `self.table` at the end of `read_csv`, but this warrants further investigation.

Data: [GHED_data.csv](https://github.com/ibis-project/ibis/files/12313409/GHED_data.csv)

`master`

DuckDB

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
This PR implements some performance improvements to the duckdb backend in support
of making one of the more difficult-to-make-usable use cases workable.

That use case is CSV files with lots of columns.

The primary part of the duckdb backend that's faster here is table metadata handling.

This is made up of a few components, and process is roughly:

1. Run a `DESCRIBE SELECT * FROM table` query to get the names and types from duckdb
1. Parse the type strings into ibis types
1. Construct a `sqlalchemy.Table` object from the resulting list of column
   names and types

It turns out that **all three** of these steps were very slow in "lots of columns" case.

### `DESCRIBE` has slightly surprising behavior (to me)

`DESCRIBE SELECT * FROM table` seems to be **slower** than `DESCRIBE table` if
`table` is a view of CSV file. I'm not sure why, but this PR moves everything
to use plain `DESCRIBE table`

### `parsy` is much slower than `sqlglot`

`parsy` was extremely to parse that many type strings. I measured about 6
seconds to parse them all versus the new sqlglot-based implementation takes
less than a second to parse them all.

**parsy**:

```
In [1]: from ibis.backends.duckdb.datatypes import parse

In [2]: %time res = list(map(parse, ['VARCHAR'] * 3220))  # 3220 is the number of columns
CPU times: user 5.76 s, sys: 0 ns, total: 5.76 s
Wall time: 5.76 s
```

**sqlglot**:

```
In [3]: import sqlglot as sg

In [4]: def parse(s):
   ...:     return sg.parse_one(s, into=sg.exp.DataType, read="duckdb")
   ...:

In [5]: %time res = list(map(parse, ['VARCHAR'] * 3220))
CPU times: user 82.1 ms, sys: 15 µs, total: 82.1 ms
Wall time: 81.6 ms
```

These aren't equivalent, because we're constructing ibis datatype objects after parsing, but the times are effectively the same for the scalar types:

```
cloud in 🌐 falcon in …/ibis on  duckdb-metadata-speedup is 📦 v6.1.0 via 🐍 v3.10.12 via ❄️  impure (ibis-3.10.12-env) took 19s
❯ ipython

In [1]: from ibis.backends.duckdb.datatypes import parse

In [2]: %time res = list(map(parse, ['VARCHAR'] * 3220))
CPU times: user 82.1 ms, sys: 1.24 ms, total: 83.4 ms
Wall time: 81.4 ms
```

That's a nice 70x (!) win for type parsing.

### Getting metadata in arrow is much faster; constructing `sa.Table`s is very expensive

The first part of this heading is not suprprising: I changed the implementation
to bring back the metadata from `DESCRIBE` as an arrow table instead of a list
of tuples. This saved a bit of time, but not that much.

The real win was avoiding `sa.Table`, which appears to do a lot of work on construction.

I tried the lighter weight `sa.table`, which does a lot less work on construction.

Using `pyinstrument`, it was actually hard to tell how much time was saved, because the `sa.table`
construction doesn't even show up!

Maybe there's a way to continue to use `sa.Table`, but I didn't really feel
like going down that rabbit hole.

### Wide-table CSVs are now usable in ibis

After this PR the CSV from the #6832 is now usable in ibis in an interactive shell session.

Closes #6832.

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -700,13 +700,12 @@ class BaseAlchemyBackend(BaseSQLBackend):
 
             from_table_expr = obj
 
-            with self.begin() as bind:
-                if from_table_expr is not None:
-                    compiled = from_table_expr.compile()
-                    columns = [
-                        self.con.dialect.normalize_name(c)
-                        for c in from_table_expr.columns
-                    ]
+            if from_table_expr is not None:
+                compiled = from_table_expr.compile()
+                columns = [
+                    self.con.dialect.normalize_name(c) for c in from_table_expr.columns
+                ]
+                with self.begin() as bind:
                     bind.execute(to_table.insert().from_select(columns, compiled))
         elif isinstance(obj, (list, dict)):
             to_table = self._get_sqla_table(table_name, schema=database)
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -443,6 +443,57 @@ WHERE catalog_name = :database"""
             con.exec_driver_sql(view)
         return self.table(table_name)
 
+    def _get_sqla_table(
+        self,
+        name: str,
+        schema: str | None = None,
+        database: str | None = None,
+        **_: Any,
+    ) -> sa.Table:
+        current_db = self.current_database
+        current_schema = self.current_schema
+        if schema is None:
+            schema = current_schema
+        *db, schema = schema.split(".")
+        db = "".join(db) or database
+        ident = ".".join(
+            map(
+                self._quote,
+                filter(None, (db if db != current_db else None, schema)),
+            )
+        )
+
+        query = f"DESCRIBE SELECT * FROM {ident}.{self._quote(name)}"
+
+        with self.begin() as con:
+            # fetch metadata with pyarrow, it's much faster for tables with "lots"
+            # of columns
+            meta = con.exec_driver_sql(query).cursor.fetch_arrow_table()
+
+        names = meta["column_name"].to_pylist()
+        types = meta["column_type"].to_pylist()
+        nullables = pa.compute.equal(meta["null"], "YES").to_pylist()
+
+        ibis_schema = sch.Schema(
+            {
+                name: parse(typ).copy(nullable=nullable)
+                for name, typ, nullable in zip(names, types, nullables)
+            }
+        )
+        columns = self._columns_from_schema(name, ibis_schema)
+        return sa.table(name, *columns, schema=ident)
+
+    def drop_table(
+        self, name: str, database: str | None = None, force: bool = False
+    ) -> None:
+        name = self._quote(name)
+        # TODO: handle database quoting
+        if database is not None:
+            name = f"{database}.{name}"
+        drop_stmt = "DROP TABLE" + (" IF EXISTS" * force) + f" {name}"
+        with self.begin() as con:
+            con.exec_driver_sql(drop_stmt)
+
     def read_parquet(
         self,
         source_list: str | Iterable[str],
@@ -991,17 +1042,6 @@ WHERE catalog_name = :database"""
                 _register(name, table)
             except duckdb.NotImplementedException:
                 _register(name, data.to_pyarrow(schema))
-
-    def _get_sqla_table(
-        self, name: str, schema: str | None = None, **kwargs: Any
-    ) -> sa.Table:
-        with warnings.catch_warnings():
-            # We don't rely on index reflection, ignore this warning
-            warnings.filterwarnings(
-                "ignore",
-                message="duckdb-engine doesn't yet support reflection on indices",
-            )
-            return super()._get_sqla_table(name, schema, **kwargs)
 
     def _get_temp_view_definition(
         self, name: str, definition: sa.sql.compiler.Compiled
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -450,29 +450,57 @@ WHERE catalog_name = :database"""
         database: str | None = None,
         **_: Any,
     ) -> sa.Table:
-        current_db = self.current_database
-        current_schema = self.current_schema
         if schema is None:
-            schema = current_schema
+            schema = self.current_schema
         *db, schema = schema.split(".")
         db = "".join(db) or database
         ident = ".".join(
             map(
                 self._quote,
-                filter(None, (db if db != current_db else None, schema)),
+                filter(None, (db if db != self.current_database else None, schema)),
             )
         )
 
-        query = f"DESCRIBE SELECT * FROM {ident}.{self._quote(name)}"
+        s = sa.table(
+            "columns",
+            sa.column("table_catalog", sa.TEXT()),
+            sa.column("table_schema", sa.TEXT()),
+            sa.column("table_name", sa.TEXT()),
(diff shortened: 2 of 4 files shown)


cpcloud:
There's a strange bug (https://github.com/duckdb/duckdb/issues/8539) that prevents us from using the faster `describe t` version of things for now :(

ibis/backends/duckdb/tests/test_datatypes.py
```diff
@@ -73,12 +74,6 @@
                 ),
             ),
             ("INTEGER[][]", dt.Array(dt.Array(dt.int32))),
-            ("TIMESTAMP_TZ", dt.Timestamp("UTC")),
```
cpcloud, line 76: These types are no longer (or were never) valid.

ibis/backends/duckdb/datatypes.py
```diff
@@ -1,98 +1,94 @@
...
+    else:
+        raise exc.IbisTypeError(f"Unknown type: {typ}")
+
+
+# these are special cases that need to be handled before sqlglot because
+# sqlglot doesn't support them yet
+SQLGLOT_UNSUPPORTED_TYPES = {
+    "INT128": dt.Decimal(38, 0),
+    "HUGEINT": dt.Decimal(38, 0),
+    "NUMERIC": dt.Decimal(18, 3),
```
cpcloud, line 82: These were added upstream today.

ibis/backends/duckdb/datatypes.py
```diff
@@ -1,98 +1,94 @@
...
-        | spaceless_string("integer", "int4", "int", "signed").result(dt.int32)
-        | spaceless_string("ubigint").result(dt.uint64)
-        | spaceless_string("usmallint").result(dt.uint16)
-        | spaceless_string("uinteger").result(dt.uint32)
-        | spaceless_string("utinyint").result(dt.uint8)
-        | spaceless_string("uuid").result(dt.uuid)
-        | spaceless_string("varchar", "char", "bpchar", "text", "string").result(
-            dt.string
+
+SQLGLOT_TYPE_TO_IBIS_TYPE = {
```
cpcloud, line 12: I'm going to replace our parsy usage in the specifc backends with sqlglot, so this code will be refactored to be more generic then, but for now I'm happy with it as is.

cpcloud: Actually I think I'll revert the type parsing changes in this PR and put them in the parsing refactor PR (to-be-PR'd) and use `functools.cache` to speed things up here for the time being.

cpcloud:
Alright, I was able to workaround the upstream bug using the `information_schema.columns` view to get the view metadata much faster than `DESCRIBE SELECT * FROM t`, along with the correct ordinal position of the columns.

Here's the current profile:

**`master`**:

image

**This PR**:

image

So about a 2x speed up overall, with most the improvement coming from querying static type information instead of using `DESCRIBE SELECT *`

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -700,13 +700,12 @@ def insert(
...
 
-            with self.begin() as bind:
-                if from_table_expr is not None:
-                    compiled = from_table_expr.compile()
-                    columns = [
-                        self.con.dialect.normalize_name(c)
-                        for c in from_table_expr.columns
-                    ]
+            if from_table_expr is not None:
+                compiled = from_table_expr.compile()
```
cpcloud, line 704: With these changes `compile` now potentially calls `self.current_database` (only for `ops.DatabaseTable`s, so we know we have a connection; `ops.UnboundTable`s are handled elsewhere), which would introduce nested transactions and those aren't supported in DuckDB it seems. So, I moved the transaction such that it's no longer nested.

ibis/backends/duckdb/__init__.py
```diff
@@ -443,6 +443,85 @@ def read_csv(
...
+        name: str,
+        schema: str | None = None,
+        database: str | None = None,
+        **_: Any,
+    ) -> sa.Table:
+        if schema is None:
+            schema = self.current_schema
+        *db, schema = schema.split(".")
+        db = "".join(db) or database
+        ident = ".".join(
```
kszucs, line 457: I would find a simple `if/else` with f-string more readable here.

ibis/backends/duckdb/__init__.py
```diff
@@ -443,6 +443,85 @@ def read_csv(
...
+        if not meta:
+            raise sa.exc.NoSuchTableError(name)
+
+        names = meta["column_name"].to_pylist()
+        types = meta["data_type"].to_pylist()
+        nullables = meta["nullable"].to_pylist()
+
+        ibis_schema = sch.Schema(
+            {
+                name: parse(typ).copy(nullable=nullable)
```
kszucs, line 507: Parse should accept a nullable argument to prevent copying.

kszucs (approved):

ibis/backends/duckdb/__init__.py
cpcloud: I don't think the `if`/`else` will be simple 😂

ibis/backends/duckdb/__init__.py
cpcloud: I thought it was super cheap to copy in primitive type case. Is that not true?

ibis/backends/duckdb/__init__.py
kszucs: Isn't this equivalent with 

```py
if db == self.current_database:
    ident = self._quote(schema)
else:
    ident = f"{self._quote(db)}.{self._quote(schema)}"
```
?

ibis/backends/duckdb/__init__.py
kszucs: It is not supposed to be expensive, but validation must happen twice.

PR #6835, opened 2023-08-10
feat(flink): implement a minimal PyFlink `Backend`

deepyaman:
# How is this tested?

_NB: This requires `pip install apache-flink` on top of the Ibis requirements, which will overwrite some dependency versions._

As a quick update, as of https://github.com/ibis-project/ibis/pull/6835/commits/7685791c877557349072e17c28dc2e58a2649612, the backend "works" (with many caveats). For example:

```python
from pyflink.table import EnvironmentSettings, TableEnvironment

# using batch table environment to execute the queries
env_settings = EnvironmentSettings.in_batch_mode()
table_env = TableEnvironment.create(env_settings)

# create a backend from the Flink table environment
import ibis

backend = ibis.flink.connect(table_env)

# load data into the table environment
table_env.execute_sql(
    """
    CREATE TABLE awards_players (
      playerID STRING,
      awardID STRING,
      yearID BIGINT,
      lgID STRING,
      tie STRING,
      notes STRING
    ) WITH (
      'connector' = 'filesystem',
      'path' = '/workspaces/ibis/ci/ibis-testing-data/csv/awards_players.csv',  # replace with path to your example data
      'format' = 'csv',
      'csv.ignore-parse-errors' = 'true'
    )
    """
)

table_env.execute_sql(
    """
    CREATE TABLE batting (
      playerID STRING,
      yearID BIGINT,
      stint BIGINT,
      teamID STRING,
      lgID STRING,
      G BIGINT,
      AB BIGINT,
      R BIGINT,
      H BIGINT,
      X2B BIGINT,
      X3B BIGINT,
      HR BIGINT,
      RBI BIGINT,
      SB BIGINT,
      CS BIGINT,
      BB BIGINT,
      SO BIGINT,
      IBB BIGINT,
      HBP BIGINT,
      SH BIGINT,
      SF BIGINT,
      GIDP BIGINT
    ) WITH (
      'connector' = 'filesystem',
      'path' = '/workspaces/ibis/ci/ibis-testing-data/csv/batting.csv',  # replace with path to your example data
      'format' = 'csv',
      'csv.ignore-parse-errors' = 'true'
    )
    """
)

backend.list_tables()  # ['awards_players', 'batting']
backend.list_databases()  # ['default_database']

batting = backend.table("batting")
awards_players = backend.table("awards_players")

left = batting[batting.yearID == 2015]
right = awards_players[awards_players.lgID == "NL"].drop("yearID", "lgID")
predicate = ["playerID"]
expr = left.join(right, predicate, how="inner")

expr.to_pandas()
```

## Changelog

* `expr.to_pandas()` and `expr.execute()` work instead of the roundabout `backend.execute(expr)` (fixed in https://github.com/ibis-project/ibis/pull/6835/commits/e9b2e6058bd84169240ced878c14ed5ec6fb72e6)

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -44,6 +44,7 @@ jobs:
       matrix:
         python-version:
           - "3.9"
+          - "3.10" # For PyFlink, which does not support Python 3.11 yet
           - "3.11"
     steps:
       - name: checkout
@@ -786,7 +787,7 @@ jobs:
           - windows-latest
         python-version:
           - "3.9"
-          - "3.11"
+          - "3.10"
     steps:
       - name: checkout
         uses: actions/checkout@v3
@@ -826,6 +827,10 @@ jobs:
 
       - name: install ibis
         run: poetry install --without dev --without docs --extras flink
+
+      # TODO(deepyaman): Remove step once Ibis and Flink are compatible.
+      - name: install pyflink
+        run: poetry run pip install apache-flink
 
       - name: show installed deps
         run: poetry run pip list
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -1,0 +1,269 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import pyflink.version
+import sqlglot as sg
+from pyflink.table.types import create_arrow_schema
+
+import ibis.common.exceptions as exc
+import ibis.expr.operations as ops
+import ibis.expr.schema as sch
+from ibis.backends.base import BaseBackend, CanListDatabases
+from ibis.backends.base.sql.ddl import fully_qualified_re, is_fully_qualified
+from ibis.backends.flink.compiler.core import FlinkCompiler
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+    import pandas as pd
+    import pyarrow as pa
+    from pyflink.table import TableEnvironment
+
+    import ibis.expr.types as ir
+
+
+class Backend(BaseBackend, CanListDatabases):
+    name = "flink"
+    compiler = FlinkCompiler
+    supports_temporary_tables = True
+    supports_python_udfs = True
+
+    def do_connect(self, t_env: TableEnvironment) -> None:
+        """Create a Flink `Backend` for use with Ibis.
+
+        Parameters
+        ----------
+        t_env
+            A table environment
+
+        Examples
+        --------
+        >>> import ibis
+        >>> from pyflink.table import EnvironmentSettings, TableEnvironment
+        >>> t_env = TableEnvironment.create(EnvironmentSettings.in_streaming_mode())
+        >>> ibis.flink.connect(t_env)
+        <ibis.backends.flink.Backend at 0x...>
+        """
+        self._t_env = t_env
+
+    def list_databases(self, like: str | None = None) -> list[str]:
+        databases = self._t_env.list_databases()
+        return self._filter_with_like(databases, like)
+
+    @property
+    def current_database(self) -> str:
+        return self._t_env.get_current_database()
+
+    def list_tables(
+        self, like: str | None = None, database: str | None = None
+    ) -> list[str]:
+        tables = self._t_env._j_tenv.listTables(
+            self._t_env.get_current_catalog(), database or self.current_database
+        )
+        return self._filter_with_like(tables, like)
+
+    def _fully_qualified_name(self, name: str, database: str | None) -> str:
+        if is_fully_qualified(name):
+            return name
+
+        return sg.table(name, db=database or self.current_database).sql(dialect="hive")
+
+    def table(self, name: str, database: str | None = None) -> ir.Table:
+        """Return a table expression from a table or view in the database.
+
+        Parameters
+        ----------
+        name
+            Table name
+        database
+            Database in which the table resides
+
+        Returns
+        -------
+        Table
+            Table named `name` from `database`
+        """
+        if database is not None and not isinstance(database, str):
+            raise exc.IbisTypeError(
+                f"`database` must be a string; got {type(database)}"
+            )
+        schema = self.get_schema(name, database=database)
+        qualified_name = self._fully_qualified_name(name, database)
+        _, quoted, unquoted = fully_qualified_re.search(qualified_name).groups()
+        unqualified_name = quoted or unquoted
+        node = ops.DatabaseTable(unqualified_name, schema, self, namespace=database)
+        return node.to_expr()
+
+    def get_schema(self, table_name: str, database: str | None = None) -> sch.Schema:
+        """Return a Schema object for the indicated table and database.
+
+        Parameters
+        ----------
+        table_name
+            Table name
+        database
+            Database name
+
+        Returns
+        -------
+        sch.Schema
+            Ibis schema
+        """
+        qualified_name = self._fully_qualified_name(table_name, database)
+        table = self._t_env.from_path(qualified_name)
+        schema = table.get_schema()
+        return sch.Schema.from_pyarrow(
+            create_arrow_schema(schema.get_field_names(), schema.get_field_data_types())
+        )
+
+    @property
+    def version(self) -> str:
+        return pyflink.version.__version__
+
+    def compile(
+        self,
+        expr: ir.Expr,
+        params: Mapping[ir.Expr, Any] | None = None,
+        **kwargs: Any,
+    ) -> Any:
+        """Compile an expression."""
+        return super().compile(expr, params=params)  # Discard `limit` and other kwargs.
+
+    def _to_sql(self, expr: ir.Expr, **kwargs: Any) -> str:
+        return str(self.compile(expr, **kwargs))
+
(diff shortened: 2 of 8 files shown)


lostmygithubaccount:
very excited to try this out if we can get a minimal examples I could follow -- happy to setup a local Kafka cluster or whatever is needed

deepyaman:
> very excited to try this out if we can get a minimal examples I could follow -- happy to setup a local Kafka cluster or whatever is needed

Of course! I should have mentioned that this is still WIP (read: not working); I just pushed it up so I could share progress later today. :) I'll let you all know when have a minimal example working, even if things like CI aren't passing!

deepyaman:
As a quick update, as of https://github.com/ibis-project/ibis/pull/6835/commits/7685791c877557349072e17c28dc2e58a2649612, the backend "works" (with many caveats). For example:

```python
from pyflink.table import EnvironmentSettings, TableEnvironment

# using batch table environment to execute the queries
env_settings = EnvironmentSettings.in_batch_mode()
table_env = TableEnvironment.create(env_settings)

# create a backend from the Flink table environment
from ibis.backends import flink

backend = flink.Backend().connect(table_env)  # I need to get `ibis.flink.connect()` hooked up.

# load data into the table environment
table_env.execute_sql(
    """
    CREATE TABLE awards_players (
      playerID STRING,
      awardID STRING,
      yearID BIGINT,
      lgID STRING,
      tie STRING,
      notes STRING
    ) WITH (
      'connector' = 'filesystem',
      'path' = '/workspaces/ibis/ci/ibis-testing-data/csv/awards_players.csv',
      'format' = 'csv',
      'csv.ignore-parse-errors' = 'true'
    )
    """
)

table_env.execute_sql(
    """
    CREATE TABLE batting (
      playerID STRING,
      yearID BIGINT,
      stint BIGINT,
      teamID STRING,
      lgID STRING,
      G BIGINT,
      AB BIGINT,
      R BIGINT,
      H BIGINT,
      X2B BIGINT,
      X3B BIGINT,
      HR BIGINT,
      RBI BIGINT,
      SB BIGINT,
      CS BIGINT,
      BB BIGINT,
      SO BIGINT,
      IBB BIGINT,
      HBP BIGINT,
      SH BIGINT,
      SF BIGINT,
      GIDP BIGINT
    ) WITH (
      'connector' = 'filesystem',
      'path' = '/workspaces/ibis/ci/ibis-testing-data/csv/batting.csv',
      'format' = 'csv',
      'csv.ignore-parse-errors' = 'true'
    )
    """
)

backend.list_tables()  # ['awards_players', 'batting']
backend.list_databases()  # ['default_database']

batting = backend.table("batting")
awards_players = backend.table("awards_players")

left = batting[batting.yearID == 2015]
right = awards_players[awards_players.lgID == "NL"].drop("yearID", "lgID")
predicate = ["playerID"]
expr = left.join(right, predicate, how="inner")

backend.execute(expr)  # `expr.execute()`/`expr.to_pandas()` aren't working because it's getting an extra `limit` argument; think it should be an easy fix somewhere
```

As I said, definitely not complete, but you should be able to start playing with it if you'd like. I'm working on fixing some of the issues pointed out above, but happy that it's at least partially functional!

ibis/backends/flink/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+    import ibis.expr.types as ir
+
+
+class Backend(BaseBackend, CanListDatabases):
+    name = "flink"
+    compiler = FlinkCompiler
+    supports_temporary_tables = True
+    supports_python_udfs = True
+
+    def do_connect(self, t_env: TableEnvironment) -> None:
```
chloeh13q, line 30: I'm assuming that we're going to pass keyword arguments here or some kind of Environment class that we manage instead of expecting the user to construct a `pyflink.table.TableEnvironment` themselves and passing it in?

deepyaman: I actually think this is fine/what was suggested by the team. The `TableEnvironment` is an analog for the `SparkSession` that users are expected to pass in for the PySpark backend.

chloeh13q: Hmm - I wonder what the design considerations are here? I did take a look at other backends and it seems that the pyspark backend is the only one that lets the user pass an Environment object. Does it have to do with the distributed cluster?

ibis/backends/flink/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+        if database is not None and not isinstance(database, str):
+            raise exc.IbisTypeError(
+                f"`database` must be a string; got {type(database)}"
+            )
+        schema = self.get_schema(name, database=database)
+        qualified_name = self._fully_qualified_name(name, database)
+        _, quoted, unquoted = fully_qualified_re.search(qualified_name).groups()
+        unqualified_name = quoted or unquoted
+        node = ops.DatabaseTable(unqualified_name, schema, self, namespace=database)
+        return node.to_expr()
```
chloeh13q, line 94: Do we intend to return, for example, a `FlinkTable` object here?

deepyaman: We can, if/when we need to define custom behaviors. I'm not sure it's necessary for the minimal implementation.

This is most similar to [the `BaseSQLBackend` implementation](https://github.com/ibis-project/ibis/blob/7.x.x/ibis/backends/base/sql/__init__.py#L70-L92). If you look at the Clickhouse backend, it has [a relatively-simple `ClickhouseTable` it constructs](https://github.com/ibis-project/ibis/blob/7.x.x/ibis/backends/clickhouse/__init__.py#L428) from the `DatabaseTable`; some of the other backends have much more complex implementations. I think, if anything, this is definitely something that could be in a separate PR; at this point, I don't understand backend implementation enough to understand what to actually implement in a `FlinkTable`. :)

ibis/backends/flink/__init__.py
deepyaman: I believe it has to do with the way you're wrapping the underlying library (PySpark or PyFlink).

chloeh13q:
I know this is still very much WIP and just a very basic implementation of the backend, but wondering what is the consideration behind using specific methods from pyflink such as `create_temporary_view` & `drop_temporary_view`, rather than turning everything into SQL strings and simply executing the SQL string? 

There are a few reasons that I'm raising this question: 1) we have previously discussed keeping the interface flexible and minimizing the amount of code we'd need to change, should we decide to switch away from pyflink. This seems to go against that; 2) we have been working on a "string-generating backend" thus far, and it seems natural for us to leverage the existing work there rather than re-implement everything; 3) we want to go with the route that that maximizes the return for both parties (which is also something we have previously aligned on).

deepyaman:
> I know this is still very much WIP and just a very basic implementation of the backend, but wondering what is the consideration behind using specific methods from pyflink such as `create_temporary_view` & `drop_temporary_view`, rather than turning everything into SQL strings and simply executing the SQL string?
> 
> There are a few reasons that I'm raising this question: 1) we have previously discussed keeping the interface flexible and minimizing the amount of code we'd need to change, should we decide to switch away from pyflink. This seems to go against that; 2) we have been working on a "string-generating backend" thus far, and it seems natural for us to leverage the existing work there rather than re-implement everything; 3) we want to go with the route that that maximizes the return for both parties (which is also something we have previously aligned on).
> 
> But, perhaps, the current implementation in this PR is meant to be placeholders, which is fine, I just want to make sure we're on the same page.

I'm not super tied to using PyFlink methods from these, and we could refactor to the `CREATE TEMPORARY VIEW`/`DROP VIEW` SQL. I was anyway thinking that the PyFlink methods may not provide sufficient flexibility, and that we would need to use SQL to some extent (since Ibis exposes a broad combination of things). As long as the user-facing behavior is correct, I don't think there should be any issue with it.

TL;DR: Sure, let's just treat them as placeholders for the minimal version.

.github/workflows/ibis-backends.yml
```diff
@@ -827,6 +828,10 @@ jobs:
       - name: install ibis
         run: poetry install --without dev --without docs --extras flink
 
+      # TODO(deepyaman): Remove step once Ibis and Flink are compatible.
+      - name: install pyflink
+        run: pip install apache-flink
```
cpcloud, line 833: This looks like it's installing `apache-flink` for every backend. You should be able to install it only for the `flink` backend by using `if: ${{ matrix... }}` on the job step.

cpcloud: Same thing: foiled by the GitHub UI code folding.

.github/workflows/ibis-backends.yml
```diff
@@ -786,7 +787,7 @@ jobs:
           - windows-latest
         python-version:
           - "3.9"
-          - "3.11"
+          - "3.10"
```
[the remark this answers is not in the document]
deepyaman: In this PR, it's a separate `test_flink` job (close to what's already in `main`), so it is only for that backend. :)

In the other PR, I used `exclude` on 3.11 and only worried about 3.9, trying to get Flink added to the big matrix.

cpcloud (approved):

cpcloud:
@deepyaman Would you mind squashing down to the `feat(pyflink)` commit?

deepyaman:
> @deepyaman Would you mind squashing down to the `feat(flink)` commit?

@cpcloud Done!

cpcloud:
Automerge enabled! 🚀

PR #6846, opened 2023-08-11
feat(trino): add application source to the trino client by default

cpcloud:
- feat(trino): add source application to trino backend
- docs(trino): add connection docstring


cpcloud:
Merging!

PR #6851, opened 2023-08-12
feat(snowflake): implement `read_json`

cpcloud:
Add `read_json` support to the snowflake backend. Closes #6849.

PR #6853, opened 2023-08-12
fix(snowflake): ensure that laterals joins with newlines are also rewritten

cpcloud:
Fixes a bug in snowflake lateral join rewriting where we were not accounting for newlines. Discovered fiddling around with #6851.


cpcloud:
Map tests are passing for snowflake:

```
❯ pytest -m snowflake ibis/backends/tests/test_map.py -n auto --dist loadgroup -q
bringing up nodes...
...........................                                                                                                                                                                       [100%]
27 passed in 23.20s
```

Issue #6844, opened 2023-08-11
bug: Incomplete type hints


binste:
### What happened?

While evaluating Ibis at work, I found some unexpected mypy errors:

```python
import ibis
import ibis.selectors as s
from ibis import deferred as d_

table = ibis.table({"col1": "int"})

# error: Argument 1 to "order_by" of "Table" has incompatible type "Value"; expected "str | Column | tuple[str | Column, bool] | Sequence[str] | Sequence[Column] | Sequence[tuple[str | Column, bool]] | None"  [arg-type]
table.order_by(ibis.desc("col1"))

# error: Argument 1 to "mutate" of "Table" has incompatible type "Across"; expected "Sequence[Expr] | None"  [arg-type]
table.mutate(s.across(["col1"], d_.mean()))

# error: Argument "col1_again" to "mutate" of "Table" has incompatible type "Deferred"; expected "Value"  [arg-type]
table.mutate(col1_again=d_["col1"])
```

I'll keep adding to this issue in case I'd find more. Also, a big thank you for making Ibis a typed library! :)

### What version of ibis are you using?

Ibis 6.1
Mypy 1.4.0
Python 3.11

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
@binste Thanks for the issue!

Yeah, we're not testing mypy at the moment. The vast majority of ibis was implemented before type hints became widespread and it's a lot of work to add them.

I know @jcrist has thoughts here. I think there may be a way we can ensure that our type hints are as-expected without having to annotate the entire codebase.

cpcloud:
Hopefully this isn't a blocker for you, but let us know!

binste:
> The vast majority of ibis was implemented before type hints became widespread and it's a lot of work to add them.

I can relate to that. It's the same story for Vega-Altair where I'm currently [in the process of typing at least the public API](https://github.com/altair-viz/altair/issues/2951) and it's not easy...

> I know @jcrist has thoughts here. I think there may be a way we can ensure that our type hints are as-expected without having to annotate the entire codebase.

I think the issue title I chose is misleading. Mypy allows for gradually typing existing codebases and so it's fine if Ibis is not completely typed. Issues arise when the provided type hints are either wrong (e.g. specified types on function input arguments cannot be processed by function) or incomplete (e.g. a union of types does not cover all accepted types) where the later one seems to be the case for the 3 errors I mentioned above.

[...]

PR #6854, opened 2023-08-12
fix(type hints): remove notimplemented as type hints as not valid

binste:
`NotImplemented` is an instance and the actual type is `NotImplementedType`. The mypy errors which this fixes look like this:
```bash
...
ibis/expr/types/numeric.py:711: error: Variable "builtins.NotImplemented" is not valid as a type  [valid-type]
ibis/expr/types/numeric.py:711: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
...
```

As mentioned in #6844, I can create a few more of these smaller PRs or I can bundle more changes into a single PR. Whatever works best for you.

Edit: NotImplementedType was introduced in Python 3.10 so tests fail for 3.9. I'll look into it.

### DIFF
--- a/ibis/expr/types/core.py
+++ b/ibis/expr/types/core.py
@@ -544,9 +544,7 @@ class Expr(Immutable, Coercible):
         raise NotImplementedError(type(self))
 
 
-def _binop(
-    op_class: type[ops.Binary], left: ir.Value, right: ir.Value
-) -> ir.Value | NotImplemented:
+def _binop(op_class: type[ops.Binary], left: ir.Value, right: ir.Value) -> ir.Value:
     """Try to construct a binary operation.
 
     Parameters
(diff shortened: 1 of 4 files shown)


binste:
In Python < 3.10, `NotImplemented` cannot be typed at all as mypy does not accept `type(NotImplemented)` as a type hint. Apparently it is in general unclear if and how to type hint `NotImplemented`, see https://github.com/python/mypy/issues/363. 

Based on some testing, [this SO thread](https://stackoverflow.com/questions/56615826/what-is-the-type-of-notimplemented), and looking into other libraries such as [pandas-stubs](https://github.com/pandas-dev/pandas-stubs/blob/main/pandas-stubs/_libs/interval.pyi#L139), my understanding is that newer versions of mypy work well if `NotImplementedType` is not provided as a return type but instead only the type of the object that is returned if the operation was successful.

cpcloud (approved):
LGTM! Thank you :)

PR #6855, opened 2023-08-13
refactor(datatypes): use sqlglot for parsing backend-specific types

cpcloud:
This PR moves our backend type parsing system to use sqlglot for all backends that were previously using `parsy`. The ibis type parser continues to use `parsy` due to heavy customization for geospatial and timestamp types.


ibis/formats/parser.py
```diff
@@ -0,0 +1,213 @@
...
+    # TIMESTAMPLTZ = auto()
+    # UNKNOWN = auto()  # Sentinel value, useful for type annotation
+    # UINT128 = auto()
+    # UINT256 = auto()
+    # USERDEFINED = "USER-DEFINED"
+    # XML = auto()
+}
+
+
+class TypeParser(Singleton, Slotted):
```
kszucs, line 97: Prefer to use `Annotable` or rather `Concrete` (if the object is supposed to be immutable) over `Slotted` since those are way more robust. `Slotted` is mostly for reducing the boilerplate for the patterns which are dependencies for `Annotable`.

kszucs, line 97: Do we need to instantiate the same type parser with different parameters somewhere?

Of not, could we use a plain class with classvars and classmethods instead? Similarly to `TypeMapper`, `SchemaMapper`  and `DataMapper`.

ibis/backends/trino/datatypes.py
```diff
@@ -84,99 +72,30 @@ def _floating(element, compiler, **kw):
...
 
 
-_from_trino_types = {
-    DOUBLE: dt.Float64,
-    sat.REAL: dt.Float32,
-    JSON: dt.JSON,
-}
+def parse(text: str) -> dt.DataType:
+    """Parse a Trino type into an ibis data type."""
+    return TrinoTypeParser().parse(text)
```
kszucs, line 85: We wouldn't need the parse function if we could use `TrinoTypeParser.parse()` as a classmethod directly, see my comments at `TypeParser`'s definition.

ibis/formats/parser.py
```diff
@@ -0,0 +1,213 @@
...
+import sqlglot as sg
+from sqlglot.expressions import ColumnDef, DataType, DataTypeSize
+
+import ibis.common.exceptions as exc
+import ibis.expr.datatypes as dt
+from ibis.common.bases import Singleton, Slotted
+from ibis.common.collections import FrozenDict
+
+SQLGLOT_TYPE_TO_IBIS_TYPE = {
+    DataType.Type.BIGDECIMAL: dt.Decimal(76, 38),
```
kszucs, line 15: Can you alias `DataType.Type` before the mapping, e.g. `sgt = DataType.Type`?

ibis/formats/parser.py
```diff
@@ -0,0 +1,213 @@
...
+    default_fallback: Mapping[str, dt.DataType] = FrozenDict(
+        {
+            "INTERVAL YEAR TO MONTH": dt.Interval("M"),
+            "INTERVAL DAY TO SECOND": dt.Interval("ms"),
+        }
+    )
+    """Mapping of SQL string types to ibis types shared by all parser child classes."""
+
+    @property
+    def _fallback(self) -> Mapping[str, dt.DataType]:
```
kszucs, line 130: Is this used anywhere else then the following `parse()` method?

ibis/formats/parser.py
```diff
@@ -0,0 +1,213 @@
...
+        return FrozenDict({**self.default_fallback, **self.fallback})
+
+    def parse(self, text: str) -> dt.DataType:
+        """Parse a type string into an ibis data type."""
+        fallback = self._fallback
+        if dtype := fallback.get(text, fallback.get(text.upper())):
```
kszucs, line 136: It seems to be a bit counterintuitive to me, if there is a fallback then return with it before parsing?

cpcloud: Yes. It's a way to workaround missing support for a given type in sqlglot.

ibis/formats/parser.py
cpcloud: No. I guess I could inline it.

ibis/formats/parser.py
cpcloud: I'd rather leave it like this. While it's very repetitive, it's also one less level of indirection.

ibis/formats/parser.py
kszucs: I mean I'd expect to return with the fallback value after failing to parse the input.

ibis/formats/parser.py
cpcloud: > Do we need to instantiate the same type parser with different parameters somewhere?

Nope, I can look into this

ibis/formats/parser.py
cpcloud: This is used to support the various specific geometry types in postgres like `line` and `point`.

kszucs: So IIUC if there is an exact typestring match in either of the fallback mappings this will return with that exact ibis type, otherwise it will use sqlglot to do the actual parsing. 

Could we use a different naming here instead of `fallback`? That indicates a different behaviour to me.

kszucs (commented):
Really nice refactor! I mostly have code organization related questions.

ibis/formats/parser.py
kszucs: Fair enough.

ibis/backends/trino/datatypes.py
cpcloud: We need to keep the function, because its the uniform interface to parsing a backends types, and not all backends need to or can use `TypeParser`.

kszucs: Sure, this is similar to the `_to_sqla_types` and `_from_sqla_types` classmethods in the base translator classes.

ibis/formats/parser.py
cpcloud: Done!

ibis/formats/parser.py
cpcloud: Done

cpcloud: No longer applies; I moved everything to be simple classes with only classmethods and class members.

ibis/formats/parser.py
cpcloud: Changed to `short_circuit`

ibis/formats/parser.py
```diff
@@ -0,0 +1,214 @@
...
+    # TIMESTAMPLTZ = auto()
+    # UNKNOWN = auto()  # Sentinel value, useful for type annotation
+    # UINT128 = auto()
+    # UINT256 = auto()
+    # USERDEFINED = "USER-DEFINED"
+    # XML = auto()
+}
+
+
+class TypeParser:
```
kszucs, line 96: Perhaps we should have a base class something like

```py
class Uninstantiable(Base):
    def __create__(cls, *args, **kwargs):
        raise TypeError(f"{cls} is not supposed to be instantiated")
```

thoughts?

ibis/formats/parser.py
```diff
@@ -0,0 +1,214 @@
...
+
+    @property
+    @abc.abstractmethod
+    def dialect(self) -> str:
+        """The dialect this parser is for."""
+
+    default_nullable = True
+    """Default nullability when not specified."""
+
+    default_decimal_precision: int | None = None
```
kszucs, line 107: IIRC typehints for class variables should be `typing.ClassVar[int | None]`

kszucs:
Thanks for updating! Having a final look.

cpcloud:
Lemme add the breaking change for bumping the lower bound of sqlglot

ibis/formats/parser.py
```diff
@@ -0,0 +1,214 @@
...
+            "INTERVAL YEAR TO MONTH": dt.Interval("M"),
+            "INTERVAL DAY TO SECOND": dt.Interval("ms"),
+        }
+    )
+    """Mapping of SQL string types to ibis types shared by all parser child classes."""
+
+    @classmethod
+    def parse(cls, text: str) -> dt.DataType:
+        """Parse a type string into an ibis data type."""
+        short_circuit = FrozenDict({**cls.default_short_circuit, **cls.short_circuit})
```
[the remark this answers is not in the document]
cpcloud: Turns out to be only used in the Trino backend!

PR #6860, opened 2023-08-13
feat(clickhouse): support temporary tables and caching

cpcloud:
Add support for temp tables and `cache` for clickhouse

### DIFF
--- a/ibis/backends/clickhouse/__init__.py
+++ b/ibis/backends/clickhouse/__init__.py
@@ -468,9 +468,7 @@ class Backend(BaseBackend, CanCreateDatabase):
         self.con.close()
 
     def _fully_qualified_name(self, name: str, database: str | None) -> str:
-        return sg.table(name, db=database or self.current_database or None).sql(
-            dialect="clickhouse"
-        )
+        return sg.table(name, db=database).sql(dialect="clickhouse")
 
     def get_schema(self, table_name: str, database: str | None = None) -> sch.Schema:
         """Return a Schema object for the indicated table and database.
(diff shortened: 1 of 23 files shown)


ibis/backends/clickhouse/__init__.py
```diff
@@ -585,14 +583,17 @@ def create_table(
...
-            raise com.IbisError(
-                "ClickHouse temporary tables are not yet supported due to a bug in `clickhouse_driver`"
-            )
-
         tmp = "TEMPORARY " * temp
         replace = "OR REPLACE " * overwrite
-        table = self._fully_qualified_name(name, database)
+
+        if temp and overwrite:
+            raise com.IbisInputError("Cannot specify both temp and overwrite")
```
cpcloud, line 590: ClickHouse currently does not allow `CREATE OR REPLACE TEMPORARY`.

We could probably implement this ourselves using `ALTER`, but that can be done in a follow-up.

ibis/backends/clickhouse/__init__.py
```diff
@@ -660,3 +661,9 @@ def drop_view(
         if_exists = "IF EXISTS " * force
         with closing(self.raw_sql(f"DROP VIEW {if_exists}{name}")):
             pass
+
+    def _load_into_cache(self, name, expr):
+        self.create_table(name, expr, schema=expr.schema(), temp=True)
+
+    def _clean_up_cached_table(self, op):
+        self.drop_table(op.name)
```
cpcloud, line 669: Perhaps these two methods should be the base implementations.

kszucs (approved):

PR #6862, opened 2023-08-13
fix(type hints): various improvements to type hints

binste:
Related to the discussion in #6844. See the individual commits for the details but let me know if something is unclear. In 1b8d1f12fc6672718c79bcef6526ec2f40253ce2 I added type hints to all attributes which are defined in `__slots__` as mypy struggled to infer it for some classes and reported that the class does not have a certain attribute.

Running `mypy ibis --ignore-missing-imports` on master:
```
Found 126 errors in 37 files (checked 558 source files)
```

With this PR:
```
Found 114 errors in 31 files (checked 558 source files)
```

### DIFF
--- a/ibis/backends/dask/aggcontext.py
+++ b/ibis/backends/dask/aggcontext.py
@@ -76,6 +76,7 @@ def dask_window_agg_built_in(
 
 class Window(AggregationContext):
     __slots__ = ("construct_window",)
+    construct_window: operator.methodcaller
 
     def __init__(self, kind, *args, **kwargs):
         super().__init__(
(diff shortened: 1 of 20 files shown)


kszucs:
@binste could we partition the typehint changes per package? Like we could start typehinting `ibis.common` first since the rest of the library depends on it, and it would be easier to review and discuss by focusing to a single subpackage.

binste:
Sounds good! See https://github.com/ibis-project/ibis/pull/6867

binste:
Sure sounds good!

PR #6875, opened 2023-08-16
perf(duckdb): improve `to_pyarrow` performance

### LINKED ISSUE #6874: feat: DuckDB ASOF join in Ibis slower than native execution

I noticed that executing an ASOF join using the DuckDB backend is much slower than executing duckDB natively.

Ran benchmarks comparing Pandas, DuckDB, and DuckDB via Ibis from 10k to 100 million rows and 2 / 5 cols.
- DuckDB via Ibis seems to be quite a bit slower than executing the ASOF join via DuckDB natively

[Code for benchmarking](https://gist.github.com/tonysun9/dc395b08144f30dcfbf8a1f4827bcbf2)

[Results Spreadsheet](https://docs.google.com/spreadsheets/d/16q-Ld103Yi9iAxH4nA7Lz6SSzL1GgLEiCv1MjMbSZ54/edit?usp=sharing)

I would like to be able to execute an ASOF join in DuckDB quickly via Ibis. My understanding is that someone is actively implementing the merge ASOF join in DuckDB via Ibis.

Any insights into why executing `.sql()` takes additional time would also be appreciated.

6.0.0

DuckDB

### Code of Conduct

jcrist:
Given the feedback [here](https://github.com/duckdb/duckdb/discussions/8582#discussioncomment-6737173) this switches `to_pyarrow` to use `.sql(...).to_arrow_table()` instead of `.execute(...)`. This has a measurable impact on performance on some queries (up to 4x faster on my machine) as it allows duckdb to include the result conversion to pyarrow in the physical plan.

With this fix ibis & duckdb vs duckdb alone perform ~the same on the workload described in #6874.

Fixes #6874.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -862,11 +862,20 @@ WHERE catalog_name = :database"""
     ) -> pa.Table:
         self._run_pre_execute_hooks(expr)
         query_ast = self.compiler.to_ast_ensure_limit(expr, limit, params=params)
-        sql = query_ast.compile()
+
+        # We use `.sql` instead of `.execute` below for performance - in
+        # certain cases duckdb query -> arrow table can be significantly faster
+        # in this configuration. Currently `.sql` doesn't support parametrized
+        # queries, so we need to compile with literal_binds for now.
+        sql = str(
+            query_ast.compile().compile(
+                dialect=self.con.dialect, compile_kwargs={"literal_binds": True}
+            )
+        )
 
         with self.begin() as con:
-            cursor = con.execute(sql)
-            table = cursor.cursor.fetch_arrow_table()
+            cursor = con.connection.cursor()
+            table = cursor.sql(sql).to_arrow_table()
 
         return expr.__pyarrow_result__(table)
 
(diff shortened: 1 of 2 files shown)


tonysun9:
Thanks for putting this together! Appreciate the helpful answers and quick response.

cpcloud (approved):
LGTM, this is really great.

ibis/backends/duckdb/__init__.py
```diff
@@ -862,11 +862,20 @@ def to_pyarrow(
...
+        sql = str(
+            query_ast.compile().compile(
+                dialect=self.con.dialect, compile_kwargs={"literal_binds": True}
+            )
+        )
 
         with self.begin() as con:
-            cursor = con.execute(sql)
-            table = cursor.cursor.fetch_arrow_table()
+            cursor = con.connection.cursor()
```
cpcloud, line 877: I _think_ this should be used via a context manager.

cpcloud: Actually, you don't even need the `cursor`, you can just do `con.connection.sql(sql).to_arrow_table()`.

cpcloud:
Merging as is, and we can follow up with the other changes in another PR!

PR #6877, opened 2023-08-16
test(flink): add `TestConf` (without common tests)

deepyaman:
# Development notes

I've split the basic `TestConf` implementation (that can unblock @chloeh13q's work) into a separate PR from trying to resolve all of the common tests.


ibis/backends/flink/__init__.py
```diff
@@ -0,0 +1,280 @@
...
+        self,
+        expr: ir.Expr,
+        params: Mapping[ir.Expr, Any] | None = None,
+        **kwargs: Any,
+    ) -> Any:
+        """Compile an expression."""
+        return super().compile(expr, params=params)  # Discard `limit` and other kwargs.
+
+    def _to_sql(self, expr: ir.Expr, **kwargs: Any) -> str:
+        return str(self.compile(expr, **kwargs))
```
cpcloud, line 135: Does Flink follow a particular SQL dialect? If so, we should use that when generating SQL.

ibis/backends/flink/__init__.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def execute(self, expr: ir.Expr, **kwargs: Any) -> Any:
+        """Execute an expression."""
+        table_expr = expr.as_table()
+        sql = self.compile(table_expr, **kwargs)
+        df = self._t_env.sql_query(sql).to_pandas()
+
+        # TODO: remove the extra conversion
+        return expr.__pandas_result__(table_expr.__pandas_result__(df))
```
cpcloud, line 144: I would try removing the extra conversion here. If it can't be trivially removed then leave as is!

ibis/backends/flink/__init__.py
```diff
@@ -0,0 +1,280 @@
...
+
+        if obj is None and schema is None:
+            raise exc.IbisError("The schema or obj parameter is required")
+        if isinstance(obj, pa.Table):
+            obj = obj.to_pandas()
+        if isinstance(obj, pd.DataFrame):
+            qualified_name = self._fully_qualified_name(name, database)
+            table = self._t_env.from_pandas(obj)
+            # FIXME(deepyaman): Create a catalog table, not a temp view.
+            self._t_env.create_temporary_view(qualified_name, table)
```
cpcloud, line 193: Yeah, ideally this creates a table in whatever storage format is best suited for Flink (hopefully that is a thing that exists!).

pyproject.toml
```diff
@@ -196,7 +196,7 @@ duckdb = [
   "sqlalchemy",
   "sqlalchemy-views",
 ]
-flink = []
+flink = ["pyspark"]
```
cpcloud, line 199: Hm, why does the flink backend depend on pyspark?

cpcloud (changes requested):

.github/workflows/ibis-backends.yml
```diff
@@ -328,6 +346,25 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: delete header rows on ubuntu
+        if: matrix.os == 'ubuntu-latest' && matrix.backend.name == 'flink'
+        run: |
+          python -c "exec(\"TEST_FILES = ['functional_alltypes', 'diamonds', 'batting', 'awards_players']\nfor file in TEST_FILES:\n    with open(f'ci/ibis-testing-data/csv/{file}.csv', 'r+') as f:\n        f.readline()\n        data=f.read()\n        f.seek(0)\n        f.write(data)\n        f.truncate()\n\")"
```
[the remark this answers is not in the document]
cpcloud: Actually, why not use the equivalent parquet files? It looks like flink supports ingesting those. That way we don't need to maintain any hacks to remove headers from CSV files :)

ibis/backends/flink/__init__.py
deepyaman: It's SQL dialect is nonstandard/not supported by SQLGlot, if that's what you mean. It does also support a Hive dialect, but I don't think it's the more commonly used dialect, and think it may have some additional JAR dependencies based on trying it a couple days ago. I'm not confident on the answer about Hive dialect; @chloeh13q or @zhenzhongxu may know better.

.github/workflows/ibis-backends.yml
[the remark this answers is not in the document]
deepyaman: > Can this be done in the `TestConf` class? As is, this will be challenging to maintain.

Yeah, I think this is possible, especially given I ended up using Python to do this in order to support Windows in the end.

> Actually, why not use the equivalent parquet files? It looks like flink supports ingesting those. That way we don't need to maintain any hacks to remove headers from CSV files :)

I believe Parquet requires some additional JAR to be installed; see https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/connectors/datastream/formats/parquet/. I'm sure it's very doable, but I tried a while back when doing the manual Flink tests, and never managed to get it set up properly.

So, my preference would be to move this into `TestConf`, if that's OK.

ibis/backends/flink/__init__.py
```diff
@@ -0,0 +1,280 @@
...
+
+        Examples
+        --------
+        >>> import ibis
+        >>> from pyflink.table import EnvironmentSettings, TableEnvironment
+        >>> t_env = TableEnvironment.create(EnvironmentSettings.in_streaming_mode())
+        >>> ibis.flink.connect(t_env)
+        <ibis.backends.flink.Backend at 0x...>
+        """
+        self._t_env = t_env
```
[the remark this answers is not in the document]
deepyaman: How about `table_env`?

ibis/backends/flink/__init__.py
deepyaman: To be honest, I don't really understand what this is doing. :) I just saw it in most backends and added it. Would removing the extra conversion be as follows?

```suggestion
        return table_expr.__pandas_result__(df)
```

ibis/backends/flink/__init__.py
deepyaman: I think @chloeh13q is working on cleaning up this part in a separate PR, since she's more generally fleshing out the connectors.

pyproject.toml
deepyaman: I just added it as a placeholder initially, since I couldn't actually add `pyflink`. 🙈 Very initially, thought would start with executing the generated SQL in Spark until could use PyFlink, but have a workaround for now, so probably should revert this.

ibis/backends/flink/__init__.py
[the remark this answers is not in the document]
cpcloud: SGTM.

ibis/backends/flink/__init__.py
cpcloud: If it's additional dependencies, then ... yeah let's see if the current dialect is fine :)

ibis/backends/flink/__init__.py
cpcloud: It would actually be removing the inner `table_expr.__pandas_result__(df)` call in favor of `expr.__pandas_result__(df)`.

In short, this code is adapting an output DataFrame to the appropriately shaped object (scalar, Series, or DataFrame) and performing type conversion to the output type/schema of the ibis expression.

ibis/backends/flink/__init__.py
deepyaman: Confirmed with @zhenzhongxu that we don't plan to work with Hive dialect at this time.

deepyaman: @cpcloud To close the thread... upon looking into it a bit more, the Hive dialect is just intended for use when working with Hive-based systems (to require fewer rewrites), and not meant for general use. This also explains why it has a dependency on Hive JARs (since it's intended for talking with Hive filesystem). As such, we don't think it makes sense to use the Hive dialect in any way, and it's not just a more-standard alternative SQL that Flink can output.

ibis/backends/flink/__init__.py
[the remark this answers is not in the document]
deepyaman: Done.

.github/workflows/ibis-backends.yml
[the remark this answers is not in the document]
deepyaman: I actually found I could just create tables from pandas dataframes, and I loaded those from parquet (as in the pandas backend `TestConf`).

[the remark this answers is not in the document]
deepyaman: (I also couldn't think of a great way to do what I had originally in `TestConf`, since I didn't want to corrupt the local files, and wasn't sure if `TestConf` had a reliable post-run hook somewhere.)

ibis/backends/flink/__init__.py
deepyaman: Seems to work!

jcrist (approved):
Overall this LGTM! Agreed that it'll be nice to get this in to unblock other work.

PR #6879, opened 2023-08-17
feat(selectors): support column references in column selector

### LINKED ISSUE #6872: bug: `t.drop(t.a)` unexpectedly selects `BitwiseNot(a)` instead of selecting the other columns
### What happened?

Reproducer setup:

```python
import ibis

t = ibis.memtable({"a": [1], "b": [0.1]})
t.drop("a")
```

works as expected (selects column "b" and discards column "a"):

```
r0 := InMemoryTable
  data:
    PandasDataFrameProxy:
         a    b
      0  1  0.1

Selection[r0]
  selections:
    b: r0.b
```

however, the following does not do the same, and attempts to perform an `BitwiseNot` operation instead:

```python
t.drop(t.a)
```

```
r0 := InMemoryTable
  data:
    PandasDataFrameProxy:
         a    b
      0  1  0.1

Selection[r0]
  selections:
    BitwiseNot(a): BitwiseNot(r0.a)
```

For column types for which the negation operator is not valid, this causes the following crash:

```python
t.drop(t.b)
```
```python-traceback
Cell In[7], line 1
----> 1 t.drop(t.b)

File ~/code/ibis/ibis/expr/types/relations.py:1940, in Table.drop(self, *fields)
   1937     raise KeyError(f"Fields not in table: {sorted(missing_fields)}")
   1939 sels = (s.c(f) if isinstance(f, str) else f for f in fields)
-> 1940 return self.select(~s.any_of(*sels))

TypeError: bad operand type for unary ~: 'FloatingColumn'
```

6.1.0 (actually the dev source code from today).

DuckDB but it seems independent of the backend.

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
Resolves #6872

There are other ways to accomplish supporting column references in `drop`, but I think it makes sense to also support column references in the named column selector.

### DIFF
--- a/ibis/selectors.py
+++ b/ibis/selectors.py
@@ -365,9 +365,9 @@ def all_of(*predicates: str | Predicate) -> Predicate:
 
 
 @public
-def c(*names: str) -> Predicate:
+def c(*names: str | ir.Column) -> Predicate:
     """Select specific column names."""
-    names = frozenset(names)
+    names = frozenset(col if isinstance(col, str) else col.get_name() for col in names)
 
     def func(col: ir.Value) -> bool:
         schema = col.op().table.schema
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
LGTM, thanks!

PR #6882, opened 2023-08-17
test(benchmarks): add `to_pyarrow` benchmark for duckdb

cpcloud:
Adds a benchmark using the code from #6874.

### DIFF
--- a/ibis/tests/benchmarks/test_benchmarks.py
+++ b/ibis/tests/benchmarks/test_benchmarks.py
@@ -753,3 +753,59 @@ def test_parse_many_duckdb_types(benchmark):
 
     types = ["VARCHAR", "INTEGER", "DOUBLE", "BIGINT"] * 1000
     benchmark(parse_many, types)
+
+
+@pytest.fixture(scope="session")
+def sql() -> str:
+    return """
+    SELECT t1.id as t1_id, x, t2.id as t2_id, y
+    FROM t1 INNER JOIN t2
+      ON t1.id = t2.id
+    """
+
+
+@pytest.fixture(scope="session")
+def ddb(tmp_path_factory):
+    duckdb = pytest.importorskip("duckdb")
+
+    N = 20_000_000
+
+    con = duckdb.connect()
+
+    path = str(tmp_path_factory.mktemp("duckdb") / "data.ddb")
+    sql = (
+        lambda var, table, n=N: f"""
+        CREATE TABLE {table} AS
+        SELECT ROW_NUMBER() OVER () AS id, {var}
+        FROM (
+            SELECT {var}
+            FROM RANGE({n}) _ ({var})
+            ORDER BY RANDOM()
+        )
+        """
+    )
+
+    with duckdb.connect(path) as con:
+        con.execute(sql("x", table="t1"))
+        con.execute(sql("y", table="t2"))
+    return path
+
+
+def test_duckdb_to_pyarrow(benchmark, sql, ddb) -> None:
+    # yes, we're benchmarking duckdb here, not ibis
+    #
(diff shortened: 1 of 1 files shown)


ibis/tests/benchmarks/test_benchmarks.py
```diff
@@ -753,3 +753,80 @@ def parse_many(types):
...
+    with duckdb.connect(path) as con:
+        con.execute("CREATE TABLE transactions AS SELECT * FROM transactions_df")
+        con.execute("CREATE TABLE prices AS SELECT * FROM prices_df")
+    return path
+
+
+def test_asof_duckdb(benchmark, asof_sql, ddb_asof_join) -> None:
+    # yes, we're benchmarking duckdb here, not ibis
+    #
+    # we do this to get a baseline for comparison
```
jcrist, line 819: Is there a way where the benchmark result could be a ratio of `ibis-time / duckdb-time`? That way if the benchmark ratio number increases we'd see that as an immediate regression. I know airspeed velocity supports arbitrary benchmark metrics, not sure about pytest-benchmark.

ibis/tests/benchmarks/test_benchmarks.py
```diff
@@ -753,3 +753,80 @@ def parse_many(types):
...
+    # Sort transactions_df by time
+    return prices_df, transactions_df.sort_values(by="time")
+
+
+@pytest.fixture
+def asof_sql() -> str:
+    return """
+    SELECT t.time, t.shares, p.price
+    FROM transactions t
+    ASOF JOIN prices p
```
jcrist, line 795: Duckdb's ASOF join is known to be slower than pandas', I'd expect that at some point there will be a major perf improvement here just due to duckdb further optimizing this op.

Any complicated-enough query can reproduce the performance difference between `.sql()` and `.execute()` though, here's a much simpler one using their already pretty zippy standard joins. Perhaps it would be better to avoid asof joins in this case to keep the benchmark focused more on the specific issue?

```python
from contextlib import contextmanager
import timeit

import pandas as pd
import numpy as np
import duckdb

import ibis


@contextmanager
def timer(name):
    start = timeit.default_timer()
    yield
    end = timeit.default_timer()
    elapsed = end - start

    print(f"{name}: {elapsed:.2f} seconds")


N = 20_000_000
id1 = np.arange(N)
np.random.shuffle(id1)
id2 = np.arange(N)
np.random.shuffle(id2)

t1 = pd.DataFrame({"id": id1, "x": np.arange(N)})
t2 = pd.DataFrame({"id": id2, "y": np.arange(N)})

con = duckdb.connect()
con.execute("CREATE TABLE t1 AS SELECT * FROM t1")
con.execute("CREATE TABLE t2 AS SELECT * FROM t2")

sql = "SELECT * FROM t1 INNER JOIN t2 ON t1.id = t2.id"

with timer("Using .sql"):
    con.sql(sql).arrow()

with timer("Using .execute"):
    con.execute(sql).arrow()
```

jcrist (approved):
Fine by me as is. Two comments on ways this may be simplified/altered, but nothing pressing.

ibis/tests/benchmarks/test_benchmarks.py
cpcloud: I don't see anything in the documentation around this.

ibis/tests/benchmarks/test_benchmarks.py
cpcloud: Thanks, that's a good idea. Using `INNER JOIN` should be a bit more robust.

cpcloud: Done!

PR #6858, opened 2023-08-13
feat(snowflake): implement `read_parquet`

cpcloud:
Add `read_parquet` support to the snowflake backend.

PR #6884, opened 2023-08-17
refactor(snowflake): replace custom temp table ddl for memtables with `read_parquet`

cpcloud:
Delete a bunch of code for uploading memtables to snowflake that is redundant with `read_parquet`, depends on #6858.


kszucs (approved):
Much better!

PR #6867, opened 2023-08-14
fix(type hints): improve type hints in ibis.common

binste:
As suggested by @kszucs in https://github.com/ibis-project/ibis/pull/6862#issuecomment-1676799118, I copied all type hint changes from #6852 into this PR. I also fixed a few more mypy errors but this PR does not yet fix all in `ibis.common`. I think I'm not familiar enough with Ibis to suggest changes for the remaining ones but happy to brainstorm and review!

Some but not all outstanding points:

- [ ] Subclasses of `Annotable` create `Attributes without a default cannot follow attributes with one` errors. See https://github.com/ibis-project/ibis/issues/6844#issuecomment-1676403220
- [ ] `MapSet` does not take any input arguments but methods such as `__and__` pass arguments to class (`self.__class__(intersection)`) which gives `error: Too many arguments for "MapSet"  [call-arg]` errors. Should `MapSet` be an abstract base class with abstract methods and/or define `__init__`?

PR #6887, opened 2023-08-17
fix(type hints): improvements to type hints in ibis.expr

binste:
These are the changes proposed in #6862 for `ibis.expr`. There are many more errors in `ibis/expr`. As mentioned in https://github.com/ibis-project/ibis/pull/6867#issuecomment-1682735449 I'm currently getting a few hundred errors, depending on which part of ibis I check. For `ibis.expr` it's 1157 errors.

### DIFF
--- a/ibis/expr/analysis.py
+++ b/ibis/expr/analysis.py
@@ -30,7 +30,7 @@ y = Variable("y")
 # compilation later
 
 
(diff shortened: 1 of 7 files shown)


kszucs (approved):
LGTM, thanks @binste!

PR #6897, opened 2023-08-18
feat(chdb): simple prototype of an embedded clickhouse backend

kszucs:
A really basic but functional backend implementation prototype for `chDB` which is an embedded distribution of `clickhouse-local`. The [chDB project](https://github.com/chdb-io/chdb) seems to be in-progress, but could be a good material to start with.

I put up the PR mostly for discussion purposes.

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -39,6 +39,7 @@ _IBIS_TO_SQLGLOT_DIALECT = {
     "impala": "hive",
     "pyspark": "spark",
     "polars": "postgres",
+    "chdb": "clickhouse",
 }
 
 
--- a/ibis/backends/chdb/__init__.py
+++ b/ibis/backends/chdb/__init__.py
@@ -1,0 +1,119 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import chdb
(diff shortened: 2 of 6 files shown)


lostmygithubaccount:
I like it a lot conceptually. Can you add a small writeup somewhere on how someone could play around with this?

cpcloud:
I like this!

@lostmygithubaccount It looks like this is missing some basics that will prevent playing around with it (the `table` method of the backend is not implemented, and you can't do much without that). I suspect a bit of work is needed to get this into a playable state.

cpcloud:
Having more options for fast local analytics backends, especially if there's a backend that can actually go from dev to prod with no rewrites because it's the same backend just deployed differently, is a good thing IMO.

lostmygithubaccount:
> Having more options for fast local analytics backends, especially if there's a backend that can actually go from dev to prod with no rewrites because it's the same backend just deployed differently, is a good thing IMO.

agreed -- between DuckDB/MotherDuck, Polars/Polars Company, and Clickhouse here hopefully we have a great experience in this direction soon!

cpcloud:
Closing this out for now! Love the idea, but this would need to be a bit more fleshed out.

@seandavi Is a local clickhouse backend something you'd be interested in?

PR #6900, opened 2023-08-18
perf(common): improve `Concrete` construction performance

kszucs:
- added a benchmark for constructing `Concrete` objects
- entirely separated the two kinds of `Annotation`s, `Arguments` are now only available from `cls.__signature__.parameters` and `Attributes` are stored in `cls.__attributes__`
- inline the construction of `__args__` and `__precomputed_hash__` instead of slower default attribute initialization
- added a specialized `SequenceOf` pattern matcher which contains less checks and can be used in most of the cases, renamed the previous one as `GenericSequenceOf`
- construct `.__args__` and set the arg values using a single pass instead of the previous two iterations

These resulted in a rough 12% improvement in the particular benchmark. Since we heavily depend on this codepath this could even have a higher impact over the full test suite.

```
------------------------------------------------------------------------------- benchmark 'test_concrete_construction': 2 tests --------------------------------------------------------------------------------
Name (time in us)                                Min                 Max              Mean            StdDev            Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
test_concrete_construction (0145_6cb0dca)     8.2500 (1.11)     145.5000 (4.73)     8.6693 (1.12)     2.4106 (2.44)     8.5410 (1.12)     0.1260 (1.01)       38;310      115.3498 (0.89)       4991           1
test_concrete_construction (NOW)              7.4170 (1.0)       30.7920 (1.0)      7.7262 (1.0)      0.9884 (1.0)      7.6250 (1.0)      0.1250 (1.0)        45;352      129.4294 (1.0)        6437           1
```

### DIFF
--- a/ibis/common/grounds.py
+++ b/ibis/common/grounds.py
@@ -17,7 +17,6 @@ from ibis.common.annotations import (
     Argument,
     Attribute,
     Signature,
-    attribute,
 )
 from ibis.common.bases import (  # noqa: F401
     Base,
@@ -67,11 +66,10 @@ class AnnotableMeta(BaseMeta):
         namespace, arguments = {}, {}
         for name, attrib in dct.items():
             if isinstance(attrib, Pattern):
-                attrib = Argument.required(attrib)
-
-            if isinstance(attrib, Argument):
+                arguments[name] = Argument.required(attrib)
+                slots.append(name)
+            elif isinstance(attrib, Argument):
                 arguments[name] = attrib
-                attributes[name] = attrib
                 slots.append(name)
             elif isinstance(attrib, Attribute):
                 attributes[name] = attrib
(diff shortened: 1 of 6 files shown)


ibis/common/grounds.py
```diff
@@ -120,19 +125,20 @@ def __recreate__(cls, kwargs: Any) -> Self:
...
         for name, field in self.__attributes__.items():
-            if isinstance(field, Attribute):
-                if (value := field.initialize(self)) is not EMPTY:
-                    object.__setattr__(self, name, value)
+            if (default := field.initialize(self)) is not EMPTY:
+                object.__setattr__(self, name, default)
 
     def __setattr__(self, name, value) -> None:
-        if field := self.__attributes__.get(name):
+        # first try to look up the argument then the attribute
```
kszucs, line 138: This is mostly used by the `Config` global object since that is the only mutable `Annotable` subclass I'm aware of.

ibis/common/grounds.py
```diff
@@ -142,13 +148,17 @@ def __repr__(self) -> str:
...
     def __eq__(self, other) -> bool:
+        # compare types
         if type(self) is not type(other):
             return NotImplemented
-
-        return all(
-            getattr(self, name, None) == getattr(other, name, None)
-            for name in self.__attributes__
-        )
+        # compare arguments
```
kszucs, line 154: These changes are also less important, only relevant for the mutable `Annotable` objects (`Config` mostly).

cpcloud (approved):
LGTM!

Issue #6899, opened 2023-08-18
bug (maybe?): ibis.literal(<INTEGER>, type='decimal') errors out


AndrewRook:
### What happened?

I'm trying to replicate a SQL query in Ibis, and as part of that I needed to defer evaluation of a fraction to the DB (rather than having it happen in Python) due to some precision mismatches. Specifically, the logic I had is something like:
```sql
column_name ** (1 / 12)
```
It looked like a good way to defer the evaluation of the `(1 / 12)` would be to encase the numerator or denominator in an `ibis.literal`. But my first attempts failed with an unhelpful error:
```python
Python 3.10.12 | packaged by conda-forge | (main, Jun 23 2023, 22:41:52) [Clang 15.0.7 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import ibis
>>> print(ibis.__version__)
6.1.0
>>> int_literal = ibis.literal(12, type="int")  # works fine
>>> decimal_literal = ibis.literal(12, type="decimal")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<snip>/ibis/expr/types/generic.py", line 1669, in literal
    return rlz.literal(type, value).to_expr()
  File "<snip>/ibis/expr/rules.py", line 98, in __call__
    result = super().__call__(*args, **kwargs)
  File "<snip>/toolz/functoolz.py", line 304, in __call__
    return self._partial(*args, **kwargs)
  File "<snip>/ibis/expr/rules.py", line 164, in literal
    value = dt.normalize(dtype, value)
  File "<snip>/ibis/expr/datatypes/value.py", line 274, in normalize
    return out.scaleb(-dtype.scale)
TypeError: bad operand type for unary -: 'NoneType'
```

Eventually I realized that the problem was that the 12 was an integer; passing floats works fine:
```python
>>> decimal_literal = ibis.literal(12.5, type="decimal")   # works
>>> decimal_literal = ibis.literal(12., type="decimal")  # also works
```
I'm not sure this is really a bug, given that I can see a logic to saying "if you want a decimal, provide an actual decimal" 😄 . And the workaround (just add the decimal at the end) is pretty easy. But it did stump me for a good 20 minutes or so, and passing integers into places that may be designed for floats is a relatively common thing in my experience. If it's not something that's worth anyone's time to fix, maybe a more informative error message would be easier to implement and capture most of the value?

### What version of ibis are you using?

6.1.0

[...]

PR #6901, opened 2023-08-18
fix(datatypes): decimal normalization failed for integers

kszucs:
Should resolve https://github.com/ibis-project/ibis/issues/6899


kszucs:
Another option could be to just convert the input values to decimals but omit precision/scale normalization and let the backends handle it.

cpcloud:
The approach in this PR looks good to me, even looks like you fixed some bugs in scaling input decimals 🚀

It looks like there's a just a few tests that need to be updated.

cpcloud (approved):

PR #6904, opened 2023-08-19
refactor(common): improve error messages raised during validation

kszucs:
Currently we have really cryptic, hard to interpret error messages:

```py
class Example(Annotable):
    a: int
    b: int = 0
    c: str = "foo"
    d: Optional[float] = None
    e: tuple[int, ...] = (1, 2, 3)
    f: As[int] = 1

Example("1", "2", "3", "4", "5", [])
```

Previously we raised

```
'1' doesn't match InstanceOf(type=<class 'int'>)
```

After this change we have the additional information to construct the error message: 
- pattern used for validation
- mismatched value
- argument/field the validation failed on
- annotable class failed to construct

Currently this PR produces the following error message:

```py
Example('1', '2', '3', '4', '5', []) has failed due to the following errors:
   `a`: '1' is not an int
   `b`: '2' is not an int
   `d`: '4' is not either None or a float
   `e`: '5' is not a tuple of ints
   `f`: [] is not coercible to an int

 Expected signature: Example(a: int, b: int = 0, c: str = 'foo', d: Optional[float] = None, e: tuple = (1, 2, 3), f: CoercedTo[int] = 1)
```

The implementation extends the `ValidationError` with a list of `(name, value, pattern)` triplets and on display it calls `pattern.describe()` for each arguments to provide a more "natural language"-like error message. 

As an extra feature this PR turns `CoercedTo` pattern to a generic class to support explicit annotations for coercions:

```py
In [3]: from ibis.common.patterns import As # alias for CoercedTo
   ...: from ibis.common.annotations import annotated
   ...:
   ...: @annotated
   ...: def inc(a: As[int]) -> int:
   ...:     return a + 1
   ...:
   ...: inc("123")
Out[3]: 124

In [4]: inc("string")
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 inc("string")
...
ValueError: invalid literal for int() with base 10: 'string'
```

### DIFF
--- a/ibis/common/patterns.py
+++ b/ibis/common/patterns.py
@@ -12,7 +12,7 @@ from itertools import chain
 from typing import (
     Annotated,
     ForwardRef,
-    Generic,  # noqa: F401
+    Generic,
     Literal,
     Optional,
     TypeVar,
@@ -32,6 +32,7 @@ from ibis.common.typing import (
     Sentinel,
     UnionType,
     _ClassInfo,
+    format_typehint,
     get_bound_typevars,
     get_type_params,
 )
@@ -128,6 +129,8 @@ class Pattern(Hashable):
                 return LazyInstanceOf(annot)
             else:
                 raise TypeError(f"Cannot create validator from annotation {annot!r}")
+        elif origin is CoercedTo:
+            return CoercedTo(args[0])
         elif origin is Literal:
             # for literal types we check the value against the literal values
             return IsIn(args)
@@ -236,6 +239,9 @@ class Pattern(Hashable):
         Whether the value matches the pattern.
         """
         return self.match(value, context) is not NoMatch
+
+    def describe(self, plural=False):
+        return "matching {self!r}"
 
     @abstractmethod
     def __eq__(self, other: Pattern) -> bool:
@@ -504,20 +510,6 @@ class Call(Slotted, Builder):
 _ = Variable("_")
 
 
-class Always(Slotted, Singleton, Pattern):
-    """Pattern that matches everything."""
-
-    def match(self, value, context):
-        return value
-
-
-class Never(Slotted, Singleton, Pattern):
-    """Pattern that matches nothing."""
-
-    def match(self, value, context):
-        return NoMatch
-
-
 class Is(Slotted, Pattern):
     """Pattern that matches a value against a reference value.
 
@@ -545,6 +537,13 @@ class Any(Slotted, Singleton, Pattern):
 
 
 _any = Any()
+
+
+class Nothing(Slotted, Singleton, Pattern):
+    """Pattern that no values."""
+
+    def match(self, value, context):
+        return NoMatch
 
 
 class Capture(Slotted, Pattern):
@@ -617,6 +616,12 @@ class Check(Slotted, Pattern):
         assert callable(predicate)
         super().__init__(predicate=predicate)
 
+    def describe(self, plural=False):
+        if plural:
+            return f"values that satisfy {self.predicate.__name__}()"
+        else:
+            return f"a value that satisfies {self.predicate.__name__}()"
+
     def match(self, value, context):
         if self.predicate(value):
             return value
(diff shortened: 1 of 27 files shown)


cpcloud:
Can you show some examples comparing the old versus the new messages?

kszucs:
So far it collects the relevant validation data for when validating a signature: `(argument name, argument value, pattern used to validate`

cpcloud:
Is this ready for review?

kszucs:
Not yet.

kszucs:
@cpcloud @lostmygithubaccount I updated the description. 

Since I don't have much time lately, may I hand it off to you? The messages generated by `Pattern.fail_for()` methods must be improved and implemented for all patterns. Also so more thorough testing would be nice too. I can imagine adding `snapshot` testing to these messages.

cpcloud:
@kszucs

```
Failed to construct Example(a: int, b: int = 0, c: str = 'foo', d: Optional[float] = None, e: tuple[int, ...] = (1, 2, 3), f: CoercedTo[int] = 1) because of the following arguments:
 a: '1' is not an int
 b: '2' is not an int
 d: '4' is not None nor a float
 e: '5' is not a sequence of ints
 f: [] is not coercible to int
```

Honestly, that is **such** a wonderful improvement over the current situation, I'm wondering if we start there and see how it fares.

ibis/common/patterns.py
```diff
@@ -759,6 +727,9 @@ class Option(Slotted, Pattern):
     def __init__(self, pat, default=None):
         super().__init__(pattern=pattern(pat), default=default)
 
+    def describe(self):
+        return f"None nor {self.pattern.describe()}"
```
cpcloud, line 731: Should this be `None or` instead of `None nor`?

ibis/common/patterns.py
```diff
@@ -821,6 +798,15 @@ class InstanceOf(Slotted, Singleton, Pattern):
...
         super().__init__(type=typ)
 
+    def describe(self, plural=False):
+        typename = self.type.__name__
+        if plural:
+            return f"{self.type.__name__}s"
+        elif typename[0].lower() in "aeiou":
+            return f"an {self.type.__name__}"
+        else:
+            return f"a {self.type.__name__}"
```
cpcloud, line 808: Great attention to detail 🧐

ibis/common/patterns.py
cpcloud: Ah, I see it's because of the way it's used. English being pesky 🤔

ibis/common/patterns.py
kszucs: I'm not sure how well will it compose with nested patterns. I still need to implement describe() for at least the common ones AnyOf, AllOf, GenericInstanceOf and GenericCoercedTo

kszucs:
The error message currently looks like:

```py
Example('1', '2', '3', '4', '5', []) has failed due to the following errors:
   `a`: '1' is not an int
   `b`: '2' is not an int
   `d`: '4' is not either None or a float
   `e`: '5' is not a tuple of ints
   `f`: [] is not coercible to an int

 Expected signature: Example(a: int, b: int = 0, c: str = 'foo', d: Optional[float] = None, e: tuple = (1, 2, 3), f: CoercedTo[int] = 
```

ibis/common/patterns.py
```diff
@@ -688,39 +693,6 @@ def __getattr__(self, name: str) -> Pattern:
         return self.pattern(getattr(self.module, name))
 
 
-class Apply(Slotted, Pattern):
```
kszucs, line 691: Consider to keep it as a specialized pattern for `CoercedTo(type not implementing Coercible)`

cpcloud:
Personally find it a bit easier to understand the failure with the signature at the top. The explanations make more sense to me when the constraint (the signature) comes before the constraint violations(s).

kszucs:
Sounds good to me. We should have a couple of snapshot tests for validating operation nodes, because the current example is not that representative.

kszucs:
I tried reorganizing the error message to highlight the expected signature at the top, but it became pretty tedious. 
I think we could improve it, it should be fairly easy to try out variations because the error message is rendered as a template with a couple of placeholders (`call`, `sig`, `errors`).

kszucs:
@cpcloud @gforsyth ready for review, feel free to adjust the error message templates and pattern descriptions.

cpcloud (approved):

cpcloud:
Excellent work! This sets a solid foundation for making really nice error messages.

PR #6906, opened 2023-08-20
fix(api): ensure that normalization of boolean, ints and floats fail with readable error message

cpcloud:
Fixes an issue where error messages normalizing certains kinds of literals was not readable due to missing to f-string sigil.


cpcloud:
Merging!

PR #6918, opened 2023-08-22
feat(datafusion): add temporal functions

mesejo:
This PR adds the following functions:

- ExtractWeekOfYear
- ExtractMicrosecond
- ExtractEpochSeconds
- ~~Strftime~~

It also fixes a minor typo.

### DIFF
--- a/ibis/backends/datafusion/compiler.py
+++ b/ibis/backends/datafusion/compiler.py
@@ -986,7 +986,7 @@ def extract_down(op, **kw):
         input_types=[PyArrowType.from_ibis(op.arg.dtype)],
         return_type=PyArrowType.from_ibis(op.dtype),
         volatility="immutable",
-        name="extract_seconds_udf",
+        name="extract_down_udf",
     )
     arg = translate(op.arg, **kw)
     return extract_down_udf(arg)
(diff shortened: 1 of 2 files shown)


ibis/backends/datafusion/compiler.py
```diff
@@ -996,3 +996,68 @@ def down(array: pa.Array) -> pa.Array:
...
+        name="extract_microseconds_udf",
+    )
+    arg = translate(op.arg, **kw)
+    return extract_microseconds_udf(arg)
+
+
+@translate.register(ops.ExtractEpochSeconds)
+def extract_epoch_seconds(op, **kw):
+    def epoch_seconds(array: pa.Array) -> pa.Array:
+        return pc.cast(pc.divide(pc.cast(array, pa.int64()), 1000_000), pa.int32())
```
mesejo, line 1027: I could only come up with this ugly cast to extract the microseconds from a timestamp. If someone has a better idea, please let me know.

ibis/backends/tests/test_temporal.py
```diff
@@ -1337,6 +1337,13 @@ def test_interval_add_cast_column(backend, alltypes, df):
             ),
             "%Y%m%d",
             marks=[
+                pytest.mark.broken(
+                    [
+                        "datafusion",
+                    ],
+                    raises=com.UnsupportedArgumentError,
```
mesejo, line 1344: I also noticed that other backends raise generic exceptions such as AttributeError (or ValueError). Would it be interesting to make those exceptions more specific, like `com.UnsupportedArgumentError`?

ibis/backends/tests/test_temporal.py
```diff
@@ -1337,6 +1337,13 @@ def test_interval_add_cast_column(backend, alltypes, df):
             ),
             "%Y%m%d",
             marks=[
+                pytest.mark.broken(
```
gforsyth, line 1340: `broken` should be used for marking tests that are failing for reasons unrelated to a given feature / PR (so a contributor isn't blocked by a random failure due to an upstream change).

If this operation will never be support by `datafusion` it should be a `pytest.mark.never`

mesejo: I'm leaving `strftime` out of the PR since is failing in some environments.

PR #6919, opened 2023-08-22
fix: remove pandas license

lostmygithubaccount:


jcrist (approved):

kszucs:
Did you check that we don't have vendored code left in the codebase?

gforsyth:
I grepped around -- the initial vendored code was in `ibis.config` and I can't find any other references to code forked from Pandas.

PR #6920, opened 2023-08-23
test(flink): turn on the common backend test suite

deepyaman:
# Development notes

* `test_dot_sql.py`
  * **notimpl** Need to implement `Backend.sql()` for the Flink backend (or consider inheriting from `BaseSQLBackend`?); transpilation would further require a Flink SQLGlot dialect
* `test_column.py`
  * **notimpl** `ROWID` isn't defined for Flink
* `test_export.py`
  * **notimpl** PyArrow export needs to be implemented; Flink doesn't expose `Table.to_pyarrow()`; logic will need to be implemented like `Table.to_pandas()`, as in https://github.com/apache/flink/blob/release-1.17.1/flink-python/pyflink/table/table.py#L948 (or PyArrow export needs to be added upstream)
* `test_array.py` 
  * **notimpl** Flink does support arrays, but we need to load the `arraytypes` table in `conftest`, which we can most easily do once memtables are supported (from `ibis.backends.tests.data`)
* `test_dataframe_interchange.py`
  * **notimpl** depends on `Table.to_pyarrow()`, too
* `test_examples.py`
  * **notimpl**(?) I assume this needs some examples to be added in a bucket somewhere? Should this be `notimpl` or `notyet`? Seems like most of the `notyet` are for database backends, but not entirely sure on this one...
* `test_join.py`
  * **notyet** [Flink doesn't support semi joins, and it's not a priority to do so](https://issues.apache.org/jira/browse/FLINK-685)
  * **notyet**(?) Flink doesn't support anti joins
  * **notimpl** `test_join_with_pandas*` should work once `memtable` support is added
* `test_sql.py`
  * **notimpl** `UNNEST` not yet supported in backend
* `test_struct.py`
  * **notimpl** [`ROW` type is similar to the `STRUCT` type known from other non-standard-compliant frameworks](https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/dev/table/types/#row) (but not yet supported in backend)
* `test_string.py`
  * **notyet** `TRANSLATE` not yet supported in backend
  * **notyet** `FIND_IN_SET` not yet supported in backend
  * **broken** negative indexing is broken
  * **notimpl** `STRING_SPLIT` not yet supported in backend
  * **notimpl** `ARRAY_JOIN` not yet supported in backend
  * **notimpl** Levenshtein distance not supported in backend
* `test_numeric.py`
  * **notimpl** `IS_NAN` not yet supported in backend
  * **notimpl** `IS_INF` not yet supported in backend
  * **broken** [`SIGN` doesn't work on `TINYINT` (or `SMALLINT`)](https://apache-flink.slack.com/archives/C03G7LJTS2G/p1694066074658329)
  * **notyet** floating mod not yet supported in backend
  * **notyet** int division by int 0 isn't supported
  * **never** Flink isn't a SQLAlchemy backend
  * **never** Flink does not support 'MIN' or 'MAX' operation without specifying window (used in `histogram` implementation, but unbounded min/max doesn't make sense in a streaming context)
  * **notyet** [Flink doesn't implement bitwise operators yet](https://issues.apache.org/jira/browse/FLINK-6810)
* `test_generic.py`
  * **broken** [Flink runtime does not support untyped `NULL` values](https://nightlies.apache.org/flink/flink-docs-release-1.14/api/java/org/apache/flink/table/types/logical/NullType.html)
  * **broken** Sort on a non-time-attribute field is not supported.
  * **broken** NaN is not supported in Flink SQL
  * **notimpl** No SQLGlot dialect
  * **notimpl** Backend doesn't implement struct support yet
  * **notimpl** Backend doesn't implement deduplication yet (`test_distinct_keep*`), but [it's feasible](https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/dev/table/sql/queries/deduplication/) (just very different from the default SQL implementation)
  * **notimpl** [`ops.Hash` implementation requires some thinking--here might be some weird impacts of using sha256 + a 64-bit sha256 hash is going to give back a much larger number than a 64bit integer](https://claypotai.slack.com/archives/C047W8D9674/p1695863030847869?thread_ts=1695860163.571209&cid=C047W8D9674)
* `test_aggregation.py`
  * **notimpl** Backend doesn't implement struct support yet 
  * **notyet** [Flink doesn't implement bitwise operators yet](https://issues.apache.org/jira/browse/FLINK-6810)
  * **notimpl** Backend doesn't implement deduplication yet (`test_reduction_ops[flink-no_cond-*]`, ), but it should be feasible (except `heavy`) (see `test_generic.py` notes)
  * **notyet** Attempted implementation of `_count_distinct_star`, but ran into, "We now only support the count of one field."

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -158,15 +158,6 @@ jobs:
               - oracle
             services:
               - oracle
-          - name: flink
-            title: Flink
-            serial: true
-            extras:
-              - flink
-            additional_deps:
-              - apache-flink
-            even_more_deps:
-              - pandas~=1.5
         exclude:
           - os: windows-latest
             backend:
@@ -263,17 +254,6 @@ jobs:
                 - oracle
               services:
                 - oracle
-          - python-version: "3.11"
-            backend:
-              name: flink
-              title: Flink
-              serial: true
-              extras:
-                - flink
-              additional_deps:
-                - apache-flink
-              even_more_deps:
-                - pandas~=1.5
     steps:
       - name: update and install system dependencies
         if: matrix.os == 'ubuntu-latest' && matrix.backend.sys-deps != null
@@ -327,11 +307,6 @@ jobs:
         if: matrix.backend.additional_deps != null
         run: poetry run pip install ${{ join(matrix.backend.additional_deps, ' ') }}
 
-      # FIXME(deepyaman)
-      - name: install even more deps
-        if: matrix.backend.even_more_deps != null
-        run: poetry run pip install ${{ join(matrix.backend.even_more_deps, ' ') }}
-
       - name: show installed deps
         run: poetry run pip list
 
@@ -351,13 +326,7 @@ jobs:
           IBIS_EXAMPLES_DATA: ${{ runner.temp }}/examples-${{ matrix.backend.name }}-${{ matrix.os }}-${{ steps.install_python.outputs.python-version }}
 
       - name: "run serial tests: ${{ matrix.backend.name }}"
-        if: matrix.backend.serial && matrix.backend.name == 'flink'
-        run: just ci-check -m ${{ matrix.backend.name }} ibis/backends/flink/tests
-        env:
-          IBIS_EXAMPLES_DATA: ${{ runner.temp }}/examples-${{ matrix.backend.name }}-${{ matrix.os }}-${{ steps.install_python.outputs.python-version }}
-
-      - name: "run serial tests: ${{ matrix.backend.name }}"
-        if: matrix.backend.serial && matrix.backend.name != 'impala' && matrix.backend.name != 'flink'
+        if: matrix.backend.serial && matrix.backend.name != 'impala'
         run: just ci-check -m ${{ matrix.backend.name }}
         env:
           IBIS_EXAMPLES_DATA: ${{ runner.temp }}/examples-${{ matrix.backend.name }}-${{ matrix.os }}-${{ steps.install_python.outputs.python-version }}
--- a/ibis/backends/flink/registry.py
+++ b/ibis/backends/flink/registry.py
@@ -50,7 +50,8 @@ def _cast(translator: ExprTranslator, op: ops.generic.Cast) -> str:
             return f"TO_TIMESTAMP(CONVERT_TZ(CAST({arg_translated} AS STRING), 'UTC+0', '{to.timezone}'))"
         else:
             return f"TO_TIMESTAMP({arg_translated})"
-
+    elif to.is_date():
+        return f"CAST({arg_translated} AS date)"
     elif to.is_json():
         return arg_translated
 
--- a/ibis/backends/flink/tests/conftest.py
+++ b/ibis/backends/flink/tests/conftest.py
@@ -28,9 +28,13 @@ class TestConf(BackendTest, RoundAwayFromZero):
     def _load_data(self, **_: Any) -> None:
         import pandas as pd
 
+        from ibis.backends.tests.data import json_types
+
         for table_name in TEST_TABLES:
             path = self.data_dir / "parquet" / f"{table_name}.parquet"
             self.connection.create_table(table_name, pd.read_parquet(path))
+
+        self.connection.create_table("json_t", json_types)
 
 
 class TestConfForStreaming(TestConf):
--- a/ibis/backends/tests/test_aggregation.py
+++ b/ibis/backends/tests/test_aggregation.py
@@ -4,7 +4,7 @@ import numpy as np
 import pandas as pd
 import pytest
 import sqlalchemy as sa
-from pytest import mark, param
+from pytest import param
 
 import ibis
 import ibis.common.exceptions as com
@@ -23,7 +23,6 @@ try:
 except ImportError:
     GoogleBadRequest = None
 
-
 try:
     from polars.exceptions import ComputeError
 except ImportError:
@@ -35,6 +34,11 @@ try:
     )
 except ImportError:
     ClickhouseDatabaseError = None
+
+try:
+    from py4j.protocol import Py4JError
+except ImportError:
+    Py4JError = None
 
 
 @reduction(input_type=[dt.double], output_type=dt.double)
@@ -71,6 +75,7 @@ aggregate_test_params = [
                 reason="no udf support",
                 raises=com.OperationNotDefinedError,
             ),
+            pytest.mark.notimpl(["flink"], "WIP", raises=com.OperationNotDefinedError),
         ],
     ),
     param(lambda t: t.double_col.min(), lambda t: t.double_col.min(), id="min"),
@@ -80,21 +85,24 @@ aggregate_test_params = [
         lambda t: (t.int_col % 3).mode(),
         lambda t: (t.int_col % 3).mode().iloc[0],
         id="mode",
-        marks=pytest.mark.notyet(
-            [
-                "bigquery",
-                "clickhouse",
-                "datafusion",
-                "impala",
-                "mysql",
-                "mssql",
-                "pyspark",
-                "trino",
-                "druid",
-                "oracle",
-            ],
-            raises=com.OperationNotDefinedError,
-        ),
+        marks=[
+            pytest.mark.notyet(
+                [
+                    "bigquery",
+                    "clickhouse",
+                    "datafusion",
+                    "impala",
+                    "mysql",
+                    "mssql",
+                    "pyspark",
+                    "trino",
+                    "druid",
+                    "oracle",
+                ],
+                raises=com.OperationNotDefinedError,
+            ),
+            pytest.mark.notimpl(["flink"], "WIP", raises=com.OperationNotDefinedError),
+        ],
     ),
     param(
         lambda t: (t.double_col + 5).sum(),
@@ -129,6 +137,7 @@ argidx_grouped_marks = ["dask"] + argidx_not_grouped_marks
 def make_argidx_params(marks):
     marks = [
         pytest.mark.notyet(marks, raises=com.OperationNotDefinedError),
+        pytest.mark.notimpl(["flink"], "WIP", raises=com.OperationNotDefinedError),
     ]
     return [
         param(
@@ -191,7 +200,7 @@ def test_aggregate_grouped(backend, alltypes, df, result_fn, expected_fn):
     backend.assert_frame_equal(result2, expected, check_dtype=False)
 
 
-@mark.notimpl(
+@pytest.mark.notimpl(
     [
         "bigquery",
         "clickhouse",
@@ -207,10 +216,11 @@ def test_aggregate_grouped(backend, alltypes, df, result_fn, expected_fn):
         "trino",
         "druid",
         "oracle",
+        "flink",
     ],
     raises=com.OperationNotDefinedError,
 )
-@mark.notimpl(
+@pytest.mark.notimpl(
     ["pyspark"],
     raises=NotImplementedError,
     reason=(
@@ -383,6 +393,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                     raises=sa.exc.DatabaseError,
                     reason="ORA-02000: missing AS keyword",
                 ),
+                pytest.mark.notimpl(["flink"], "WIP", raises=Py4JError),
             ],
         ),
         param(
@@ -405,21 +416,26 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: (t.int_col % 3).mode(where=where),
             lambda t, where: (t.int_col % 3)[where].mode().iloc[0],
             id="mode",
-            marks=pytest.mark.notyet(
-                [
-                    "bigquery",
-                    "clickhouse",
-                    "datafusion",
-                    "impala",
-                    "mysql",
-                    "pyspark",
-                    "mssql",
-                    "trino",
-                    "druid",
-                    "oracle",
-                ],
-                raises=com.OperationNotDefinedError,
-            ),
+            marks=[
+                pytest.mark.notyet(
+                    [
+                        "bigquery",
+                        "clickhouse",
+                        "datafusion",
+                        "impala",
+                        "mysql",
+                        "pyspark",
+                        "mssql",
+                        "trino",
+                        "druid",
+                        "oracle",
+                    ],
+                    raises=com.OperationNotDefinedError,
+                ),
+                pytest.mark.notimpl(
+                    ["flink"], "WIP", raises=com.OperationNotDefinedError
+                ),
+            ],
         ),
         param(
             lambda t, where: t.double_col.argmin(t.int_col, where=where),
@@ -436,6 +452,9 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                         "oracle",
                     ],
                     raises=com.OperationNotDefinedError,
+                ),
+                pytest.mark.notimpl(
+                    ["flink"], "WIP", raises=com.OperationNotDefinedError
                 ),
             ],
         ),
@@ -455,6 +474,9 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                     ],
                     raises=com.OperationNotDefinedError,
                 ),
+                pytest.mark.notimpl(
+                    ["flink"], "WIP", raises=com.OperationNotDefinedError
+                ),
             ],
         ),
         param(
@@ -462,7 +484,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: t.double_col[where].std(ddof=1),
             id="std",
             marks=[
-                mark.notimpl(
+                pytest.mark.notimpl(
                     ["druid"],
                     raises=sa.exc.ProgrammingError,
                     reason="No match found for function signature stddev_samp(<NUMERIC>)",
@@ -474,7 +496,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: t.double_col[where].var(ddof=1),
             id="var",
             marks=[
-                mark.notimpl(
+                pytest.mark.notimpl(
                     ["druid"],
                     raises=sa.exc.ProgrammingError,
                     reason="No match found for function signature var_samp(<NUMERIC>)",
@@ -486,7 +508,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: t.double_col[where].std(ddof=0),
             id="std_pop",
             marks=[
-                mark.notimpl(
+                pytest.mark.notimpl(
                     ["druid"],
                     raises=sa.exc.ProgrammingError,
                     reason="No match found for function signature stddev_pop(<NUMERIC>)",
@@ -498,7 +520,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: t.double_col[where].var(ddof=0),
             id="var_pop",
             marks=[
-                mark.notimpl(
+                pytest.mark.notimpl(
                     ["druid"],
                     raises=sa.exc.ProgrammingError,
                     reason="No match found for function signature var_pop(<NUMERIC>)",
@@ -509,14 +531,26 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: t.string_col.approx_nunique(where=where),
             lambda t, where: t.string_col[where].nunique(),
             id="approx_nunique",
-            marks=pytest.mark.notimpl(["polars"], raises=com.OperationNotDefinedError),
+            marks=[
+                pytest.mark.notimpl(["polars"], raises=com.OperationNotDefinedError),
+                pytest.mark.notimpl(["flink"], "WIP", raises=Py4JError),
+            ],
         ),
         param(
             lambda t, where: t.double_col.arbitrary(where=where),
             lambda t, where: t.double_col[where].iloc[0],
             id="arbitrary_default",
             marks=pytest.mark.notimpl(
-                ["impala", "mysql", "polars", "datafusion", "mssql", "druid", "oracle"],
+                [
+                    "impala",
+                    "mysql",
+                    "polars",
+                    "datafusion",
+                    "mssql",
+                    "druid",
+                    "oracle",
+                    "flink",
+                ],
                 raises=com.OperationNotDefinedError,
             ),
         ),
@@ -525,7 +559,16 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: t.double_col[where].iloc[0],
             id="arbitrary_first",
             marks=pytest.mark.notimpl(
-                ["impala", "mysql", "polars", "datafusion", "mssql", "druid", "oracle"],
+                [
+                    "impala",
+                    "mysql",
+                    "polars",
+                    "datafusion",
+                    "mssql",
+                    "druid",
+                    "oracle",
+                    "flink",
+                ],
                 raises=com.OperationNotDefinedError,
             ),
         ),
@@ -543,6 +586,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                         "mssql",
                         "druid",
                         "oracle",
+                        "flink",
                     ],
                     raises=com.OperationNotDefinedError,
                 ),
@@ -571,6 +615,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                         "pandas",
                         "polars",
                         "sqlite",
+                        "flink",
                     ],
                     raises=com.OperationNotDefinedError,
                 ),
@@ -591,7 +636,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: t.double_col[where].iloc[0],
             id="first",
             marks=pytest.mark.notimpl(
-                ["dask", "druid", "impala", "mssql", "mysql", "oracle"],
+                ["dask", "druid", "impala", "mssql", "mysql", "oracle", "flink"],
                 raises=com.OperationNotDefinedError,
             ),
         ),
@@ -600,7 +645,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: t.double_col[where].iloc[-1],
             id="last",
             marks=pytest.mark.notimpl(
-                ["dask", "druid", "impala", "mssql", "mysql", "oracle"],
+                ["dask", "druid", "impala", "mssql", "mysql", "oracle", "flink"],
                 raises=com.OperationNotDefinedError,
             ),
         ),
@@ -615,7 +660,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                 ),
                 pytest.mark.notimpl(["druid"], strict=False, raises=AssertionError),
                 pytest.mark.notyet(
-                    ["impala", "pyspark"], raises=com.OperationNotDefinedError
+                    ["impala", "pyspark", "flink"], raises=com.OperationNotDefinedError
                 ),
                 pytest.mark.broken(
                     ["dask"],
@@ -638,7 +683,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                     raises=com.OperationNotDefinedError,
                 ),
                 pytest.mark.notyet(
-                    ["impala", "pyspark"], raises=com.OperationNotDefinedError
+                    ["impala", "pyspark", "flink"], raises=com.OperationNotDefinedError
                 ),
                 pytest.mark.broken(
                     ["dask"],
@@ -662,7 +707,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                     raises=com.OperationNotDefinedError,
                 ),
                 pytest.mark.notyet(
-                    ["impala", "pyspark"], raises=com.OperationNotDefinedError
+                    ["impala", "pyspark", "flink"], raises=com.OperationNotDefinedError
                 ),
                 pytest.mark.broken(
                     ["dask"],
@@ -686,7 +731,7 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
             lambda t, where: t.string_col[where].tolist(),
             id="collect",
             marks=[
-                mark.notimpl(
+                pytest.mark.notimpl(
                     [
                         "impala",
                         "mysql",
@@ -707,6 +752,9 @@ def test_aggregate_multikey_group_reduction_udf(backend, alltypes, df):
                         "<dask.dataframe.groupby.Aggregation object at 0x124569840> is not "
                         "callable or a string"
                     ),
+                ),
+                pytest.mark.notimpl(
+                    ["flink"], "WIP", raises=com.OperationNotDefinedError
                 ),
             ],
         ),
@@ -757,7 +805,7 @@ def test_reduction_ops(
             lambda t: t.string_col.isin(["1", "7"]),
             id="cond",
             marks=[
-                mark.notyet(
+                pytest.mark.notyet(
                     ["snowflake", "mysql"],
                     raises=com.UnsupportedOperationError,
                     reason="backend does not support filtered count distinct with more than one column",
@@ -766,16 +814,17 @@ def test_reduction_ops(
         ),
     ],
 )
-@mark.notyet(
-    ["bigquery", "druid", "mssql", "oracle", "sqlite"],
+@pytest.mark.notyet(
+    ["bigquery", "druid", "mssql", "oracle", "sqlite", "flink"],
     raises=(
         sa.exc.OperationalError,
         sa.exc.DatabaseError,
         com.UnsupportedOperationError,
+        com.OperationNotDefinedError,
     ),
     reason="backend doesn't support count distinct with multiple columns",
 )
-@mark.notyet(
+@pytest.mark.notyet(
     ["datafusion", "impala"],
     raises=com.OperationNotDefinedError,
     reason="no one has attempted implementation yet",
@@ -797,7 +846,7 @@ def test_count_distinct_star(alltypes, df, ibis_cond, pandas_cond):
             lambda t, where: t.double_col[where].quantile(0.5),
             id="quantile",
             marks=[
-                mark.notimpl(
+                pytest.mark.notimpl(
                     [
                         "bigquery",
                         "datafusion",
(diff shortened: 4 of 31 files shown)


.github/workflows/ibis-backends.yml
```diff
@@ -329,6 +349,25 @@ jobs:
...
+          TEST_FILES = ['functional_alltypes', 'diamonds', 'batting', 'awards_players']
+          for file in TEST_FILES:
+              with open(f'ci/ibis-testing-data/csv/{file}.csv', 'r+') as f:
+                  f.readline()
+                  data=f.read()
+                  f.seek(0)
+                  f.write(data)
+                  f.truncate()
+          "@
+
```
deepyaman, line 370: I'll remove this once CI finishes once; will cherry-pick it over from the other PR.

ibis/backends/tests/test_timecontext.py
```diff
@@ -88,12 +100,33 @@ def test_context_adjustment_window_udf(alltypes, context, window, monkeypatch):
...
+    monkeypatch.setattr(ibis.options.context_adjustment, "time_col", "timestamp_col")
+
+
+@pytest.mark.notimpl(["dask", "duckdb"])
+@pytest.mark.xfail_version(pandas=["pyarrow>=13", "pandas>=2.1"])
+@pytest.mark.broken(
+    ["flink"],
+    raises=Py4JJavaError,
+    reason="Cannot cast org.apache.flink.table.data.TimestampData to java.lang.Long",
+)
```
deepyaman, line 119:
```suggestion
```

ibis/backends/tests/test_timecontext.py
```diff
@@ -88,12 +100,33 @@ def test_context_adjustment_window_udf(alltypes, context, window, monkeypatch):
...
 def test_context_adjustment_filter_before_window(alltypes, context, monkeypatch):
     monkeypatch.setattr(ibis.options.context_adjustment, "time_col", "timestamp_col")
 
     window = ibis.trailing_window(ibis.interval(days=3), order_by=ORDER_BY_COL)
+    # Note(mehmet): Works.
+    # window = ibis.trailing_window(3, order_by="id")
```
deepyaman, line 125:
```suggestion
```

ibis/backends/tests/test_timecontext.py
```diff
@@ -88,12 +100,33 @@ def test_context_adjustment_window_udf(alltypes, context, window, monkeypatch):
...
 
     expr = alltypes[alltypes["bool_col"]]
+    # Note(mehmet): Does not work; gives the same error as the above line.
+    # expr = functional_alltypes_w_watermark[functional_alltypes_w_watermark["bool_col"]]
```
deepyaman, line 129:
```suggestion
```

ibis/backends/tests/test_uuid.py
```diff
@@ -29,6 +30,7 @@
     "pandas": TEST_UUID,
     "bigquery": TEST_UUID,
     "duckdb": TEST_UUID,
+    "flink": str(TEST_UUID),
```
deepyaman, line 33:
```suggestion
```

ibis/backends/tests/test_temporal.py
```diff
@@ -1861,18 +2147,19 @@ def test_timestamp_literal(con, backend):
     ["oracle"], raises=sa.exc.DatabaseError, reason="ORA-00904: MAKE TIMESTAMP invalid"
 )
 @pytest.mark.parametrize(
-    ("timezone", "expected"),
+    ("timezone", "expected", "expected_for_flink"),
```
deepyaman, line 2150: @mfatihaktas should we be allowing different expectations for Flink? Kind of kills the portability across backends

ibis/backends/tests/test_param.py
```diff
@@ -226,8 +243,9 @@ def test_scalar_param_date(backend, alltypes, value):
         "druid",
     ]
 )
+@pytest.mark.broken(["flink"], "WIP")
 def test_scalar_param_nested(con):
     param = ibis.param("struct<x: array<struct<y: array<double>>>>")
     value = OrderedDict([("x", [OrderedDict([("y", [1.0, 2.0, 3.0])])])])
-    result = con.execute(param, {param: value})
+    result = con.execute(param, params={param: value})
```
deepyaman, line 250:
```suggestion
    result = con.execute(param, {param: value})
```

ibis/backends/tests/test_param.py
```diff
@@ -80,11 +85,14 @@ def test_scalar_param_array(con):
     assert result == len(value)
 
 
-@pytest.mark.notimpl(["datafusion", "impala", "postgres", "pyspark", "druid", "oracle"])
+@pytest.mark.notimpl(
+    ["datafusion", "impala", "flink", "postgres", "pyspark", "druid", "oracle"]
+)
```
deepyaman, line 90:
```suggestion
@pytest.mark.notimpl(["datafusion", "impala", "flink", "postgres", "pyspark", "druid", "oracle"])
```

ibis/backends/tests/test_map.py
```diff
@@ -232,19 +317,43 @@ def test_map_get_with_null_on_null_type_with_null(con, null_value):
...
-    t = con.create_table(
-        temp_table, schema=ibis.schema(dict(xyz="map<string, string>"))
-    )
+    if con.name == "flink":
+        t = con.create_table(
+            temp_table,
+            schema=ibis.schema(dict(xyz="map<string, string>")),
+            tbl_properties={
+                "connector": None,
+            },
```
deepyaman, line 343: What's this?

deepyaman, line 343: @mfatihaktas what's this? haven't looked into it, but just curious to see a different interface required

ibis/backends/tests/test_map.py
```diff
@@ -170,20 +220,30 @@ def test_literal_map_get_broadcast(backend, alltypes, df):
...
             [1, 2],
             id="string",
-            marks=pytest.mark.notyet(
-                ["postgres"], reason="only support maps of string -> string"
-            ),
+            marks=[
+                pytest.mark.notyet(
+                    ["postgres"], reason="only support maps of string -> string"
+                ),
+            ],
```
deepyaman, line 227:
```suggestion
            marks=pytest.mark.notyet(
                ["postgres"], reason="only support maps of string -> string"
            ),
```

ibis/backends/tests/test_map.py
```diff
@@ -60,6 +60,11 @@ def test_column_map_merge(backend):
     tm.assert_series_equal(result, expected)
 
 
+@pytest.mark.notimpl(
+    ["flink"],
+    raises=exc.OperationNotDefinedError,
+    reason=("No translation rule for <class 'ibis.expr.operations.maps.MapKeys'>"),
```
deepyaman, line 66:
```suggestion
    reason="No translation rule for <class 'ibis.expr.operations.maps.MapKeys'>",
```

ibis/backends/tests/test_json.py
```diff
@@ -36,9 +38,15 @@
...
 )
-def test_json_getitem(json_t, expr_fn, expected):
+def test_json_getitem(json_t, expr_fn, expected, backend):
     expr = expr_fn(json_t)
     result = expr.execute()
+
+    # Flink returns to queries
+    if backend.name() == "flink":
+        expected = [e if (e is None or isinstance(e, list)) else [e] for e in expected]
+    expected = pd.Series(expected, name="res")
```
deepyaman, line 48: @mfatihaktas again, I'm a bit concerned around making an exception for Flink specifically

ibis/backends/tests/test_generic.py
```diff
@@ -1235,6 +1279,11 @@ def test_hash_consistent(backend, alltypes):
                     reason="raises TrinoUserError",
                 ),
                 pytest.mark.broken(["polars"], reason="casts to 1672531200000000000"),
+                pytest.mark.broken(
+                    ["flink"],
+                    raises=AssertionError,
+                    reason="assert 1672549200 == 1672531200",
+                ),
```
deepyaman, line 1286:
```suggestion
```

ibis/backends/tests/test_generic.py
```diff
@@ -272,8 +285,16 @@ def test_notin(backend, alltypes, sorted_df, column, elements):
...
+        param(
+            lambda t: t["bool_col"],
+            lambda df: df["bool_col"],
+            id="no_op",
+        ),
+        param(
+            lambda t: ~t["bool_col"],
+            lambda df: ~df["bool_col"],
+            id="negate",
+        ),
```
deepyaman, line 297:
```suggestion
        param(lambda t: t["bool_col"], lambda df: df["bool_col"], id="no_op"),
        param(lambda t: ~t["bool_col"], lambda df: ~df["bool_col"], id="negate"),
```

ibis/backends/tests/test_export.py
```diff
@@ -307,11 +323,7 @@ def test_table_to_csv(tmp_path, backend, awards_players):
                 ),
                 pytest.mark.notyet(["oracle"], raises=sa.exc.DatabaseError),
                 pytest.mark.notyet(["mssql", "mysql"], raises=sa.exc.OperationalError),
-                pytest.mark.notyet(
-                    ["pyspark"],
-                    raises=AnalysisException,
-                    reason="precision is out of range",
-                ),
+                pytest.mark.notyet(["flink"], raises=NotImplementedError),
```
deepyaman, line 326:
```suggestion
                pytest.mark.notyet(
                     ["pyspark"],
                     raises=AnalysisException,
                     reason="precision is out of range",
                 ),
                pytest.mark.notyet(["flink"], raises=NotImplementedError),
```

ibis/backends/tests/test_export.py
```diff
@@ -35,13 +35,20 @@
...
         ],
     ),
 ]
 
-no_limit = [param(None, id="nolimit")]
+no_limit = [
+    param(
+        None,
+        id="nolimit",
+        marks=[pytest.mark.notimpl(["flink"])],
```
deepyaman, line 49:
```suggestion
        marks=pytest.mark.notimpl(["flink"]),
```

ibis/backends/tests/test_client.py
```diff
@@ -1144,17 +1246,33 @@ def test_set_backend_url(url, monkeypatch):
...
     )
-    con.create_table(temp_table, schema=schema, overwrite=True)
+
+    if con.name == "flink":
+        con.create_table(
+            temp_table,
+            schema=schema,
+            tbl_properties={
+                "connector": None,
+            },
```
deepyaman, line 1265: @mfatihaktas same concern as below around divergent API just for Flink

ibis/backends/tests/test_client.py
```diff
@@ -277,18 +313,36 @@ def test_rename_table(con, temp_table, temp_table_orig):
...
     )
-    t = con.create_table(temp_table, schema=sch)
+
+    if con.name == "flink":
+        t = con.create_table(
+            temp_table,
+            schema=sch,
+            tbl_properties={
+                "connector": None,
+            },
```
deepyaman, line 335: Same

ibis/backends/tests/test_client.py
```diff
@@ -170,10 +200,15 @@ def test_query_schema(ddl_backend, expr_fn, expected):
...
 @pytest.mark.never(["dask", "pandas"], reason="dask and pandas do not support SQL")
+@pytest.mark.notimpl(
+    ["flink"], raises=AttributeError, reason="'Backend' object has no attribute 'sql'"
+)
 def test_sql(backend, con):
     # execute the expression using SQL query
     table = backend.format_table("functional_alltypes")
     limit = _LIMIT.get(backend.name(), "LIMIT 10")
+
+    # TODO (mehmet): Is there a way to get ir.Expr from SQL statement for Flink?
```
deepyaman, line 211:
```suggestion
```

ibis/backends/tests/test_client.py
```diff
@@ -128,6 +149,7 @@ def test_load_data_sqlalchemy(alchemy_backend, alchemy_con, alchemy_temp_table,
 
     obj = lamduh(df)
     alchemy_con.create_table(alchemy_temp_table, obj, schema=sch, overwrite=True)
+
```
deepyaman, line 152:
```suggestion
```

ibis/backends/tests/test_client.py
```diff
@@ -94,7 +104,18 @@ def test_create_table(backend, con, temp_table, lamduh, sch):
...
     obj = lamduh(df)
-    con.create_table(temp_table, obj, schema=sch)
+    if backend.name() == "flink":
+        con.create_table(
+            temp_table,
+            obj,
+            schema=sch,
+            tbl_properties={
+                "connector": None,
+            },
```
deepyaman, line 114: same

ibis/backends/tests/test_client.py
```diff
@@ -70,17 +75,22 @@ def _create_temp_table_with_schema(con, temp_table_name, schema, data=None):
...
     "sch",
     [
-        ibis.schema(
-            [
-                ("first_name", "string"),
-                ("last_name", "string"),
-                ("department_name", "string"),
-                ("salary", "float64"),
-            ]
+        param(
```
deepyaman, line 78: Is there a reason for this diff?

ibis/backends/conftest.py
```diff
@@ -582,6 +582,78 @@ def alltypes(backend):
...
+            "connector": "filesystem",
+            "path": f"ci/ibis-testing-data/csv/{table_name}.csv",
+            "format": "csv",
+            "csv.ignore-parse-errors": "true",
+        },
+        watermark=ibis.watermark(
+            time_col="timestamp_col", allowed_delay=ibis.interval(seconds=15)
+        ),
+    )
+
```
deepyaman, line 655:
```suggestion
```

ibis/backends/conftest.py
```diff
@@ -582,6 +582,78 @@ def alltypes(backend):
...
+
```
deepyaman, line 656:
```suggestion
```

ibis/backends/flink/tests/conftest.py
```diff
@@ -27,11 +27,18 @@ def connect(*, tmpdir, worker_id, **kw: Any):
...
     def _load_data(self, **_: Any) -> None:
         import pandas as pd
+        from ibis.backends.tests.data import json_types
+
+        from ibis.backends.tests.data import json_types
 
         for table_name in TEST_TABLES:
             path = self.data_dir / "parquet" / f"{table_name}.parquet"
+
+            # import pdb; pdb.set_trace()
```
deepyaman, line 37:
```suggestion
```

ibis/backends/flink/tests/conftest.py
```diff
@@ -27,11 +27,18 @@ def connect(*, tmpdir, worker_id, **kw: Any):
 
     def _load_data(self, **_: Any) -> None:
         import pandas as pd
+        from ibis.backends.tests.data import json_types
```
deepyaman, line 30:
```suggestion
```

ibis/backends/tests/test_json.py
```diff
@@ -48,6 +56,15 @@ def test_json_getitem(json_t, expr_fn, expected):
 @pytest.mark.notyet(
     ["pyspark", "trino"], reason="should work but doesn't deserialize JSON"
 )
+@pytest.mark.notimpl(["duckdb"], raises=OperationNotDefinedError)
```
deepyaman, line 59: fix

deepyaman, line 59:
```suggestion
```

ibis/backends/tests/test_json.py
```diff
@@ -72,6 +89,15 @@ def test_json_map(json_t):
     ["pyspark", "trino"], reason="should work but doesn't deserialize JSON"
 )
 @pytest.mark.notyet(["bigquery"], reason="doesn't allow null in arrays")
+@pytest.mark.notimpl(["duckdb"], raises=OperationNotDefinedError)
```
deepyaman, line 92:
```suggestion
```

ibis/backends/tests/test_timecontext.py
```diff
@@ -106,9 +124,13 @@ def test_context_adjustment_filter_before_window(alltypes, context, monkeypatch)
 
 
 @pytest.mark.notimpl(["duckdb", "pyspark"])
+@pytest.mark.notimpl(
+    ["flink"],
+    raises=com.OperationNotDefinedError,
+    reason="No translation rule for <class 'ibis.expr.operations.structs.StructField'>",
+)
 def test_context_adjustment_multi_col_udf_non_grouped(alltypes, context, monkeypatch):
     monkeypatch.setattr(ibis.options.context_adjustment, "time_col", "timestamp_col")
```
deepyaman, line 133:
```suggestion
)
def test_context_adjustment_multi_col_udf_non_grouped(alltypes, context, monkeypatch):
    monkeypatch.setattr(ibis.options.context_adjustment, "time_col", "timestamp_col")

```

ibis/backends/tests/test_temporal.py
mfatihaktas: Yes, I agree. I think the correct result should include the timezone literal (e.g., `GMT`, `PST`) per the test function name `test_timestamp_with_timezone_literal`. However, I have not been able to find a way to get the timezone literal from Flink. We can obviously append it with `CONCAT` (after mapping the given timezone to the correct literal) but I was not sure if that would be appropriate.

Do you have further thoughts on this?

ibis/backends/tests/test_map.py
mfatihaktas: This was "my workaround" for the following error raised in `create_table`:
```python
        import pandas as pd
        import pyarrow as pa

        import ibis.expr.types as ir

            raise exc.IbisError("The schema or obj parameter is required")

        if overwrite:
            self.drop_table(name=name, catalog=catalog, database=database, force=True)

        if isinstance(obj, pa.Table):
            obj = obj.to_pandas()
        if isinstance(obj, pd.DataFrame):
            qualified_name = self._fully_qualified_name(name, database, catalog)
            table = self._table_env.from_pandas(obj)
            # in-memory data is created as views in `pyflink`
            # TODO(chloeh13q): alternatively, we can do CREATE TABLE and then INSERT
            # INTO ... VALUES to keep things consistent
            self._table_env.create_temporary_view(qualified_name, table)
        if isinstance(obj, ir.Table):
            # TODO(chloeh13q): implement CREATE TABLE for expressions
            raise NotImplementedError
        if schema is not None:
            if not tbl_properties:
>               raise exc.IbisError(
                    "tbl_properties is required when creating table with schema"
                )
E               ibis.common.exceptions.IbisError: tbl_properties is required when creating table with schema
```

I did not know how I should modify `create_table()`, so chose to work around it this way :) Given that it works with
```
    tbl_properties={
                "connector": None,
            },
```
which is technically "empty", this assert
```
        if schema is not None:
            if not tbl_properties:
>               raise exc.IbisError(
                    "tbl_properties is required when creating table with schema"
                )
```
is unnecessary perhaps?

ibis/backends/tests/test_client.py
mfatihaktas: Referring to the [reply above](https://github.com/ibis-project/ibis/pull/6920#discussion_r1372456721).

ibis/backends/tests/test_client.py
mfatihaktas: Referring to the [reply above](https://github.com/ibis-project/ibis/pull/6920#discussion_r1372456721).

ibis/backends/tests/test_client.py
mfatihaktas: Referring to the [reply above](https://github.com/ibis-project/ibis/pull/6920#discussion_r1372456721).

ibis/backends/tests/test_json.py
mfatihaktas: I also agree that, ideally, they all should return the same results. The discrepancy in the results is caused by this operator in Flink:
```python
def _json_get_item(translator: ExprTranslator, op: ops.json.JSONGetItem) -> str:
    arg_translated = translator.translate(op.arg)
    if op.index.dtype.is_integer():
        query_path = f"$[{op.index.value}]"
    else:  # is string
        query_path = f"$.{op.index.value}"

    return (
        f"JSON_QUERY({arg_translated}, '{query_path}' WITH CONDITIONAL ARRAY WRAPPER)"
    )
```
The additional term `WITH CONDITIONAL ARRAY WRAPPER` here wraps around non-none scalar results with an array. This is the reason for `if backend.name() == "flink":` in the test.

I added the term `WITH CONDITIONAL ARRAY WRAPPER` because `json_query` in Flink does not seem to work as expected otherwise (or I am missing something). Flink returns `null` for scalar fields without the additional term. I reproduced this in [here](https://github.com/mfatihaktas/flink-playground/blob/main/tests/test_json_query.py). 

How to reproduce:

`JSON_QUERY`:
```python
def test_json_query(table_env):
    # Define the source table
    dataframe = pandas.DataFrame(
        {
            "js": [
                '{"a": [1,2,3,4], "b": 1}',
                '{"a":null,"b":2}',
                '{"a":"foo", "c":null}',
                "null",
                "[42,47,55]",
                "[]",
            ]
        }
    )
    table = table_env.from_pandas(dataframe)
    table_env.create_temporary_view("source_table", table)

    # Define the sink table
    table_env.execute_sql("""
        CREATE TABLE sink_table (
            res STRING
        ) WITH (
            'connector' = 'print'
        )
    """)

    # Define a transformation
    table_result = table_env.execute_sql("""
        INSERT INTO sink_table
        SELECT JSON_QUERY(source_table.`js`, '$.a') AS `res`
        FROM source_table
    """)
    table_result.print()
```

Result:
```
[1,2,3,4]
null
null
null
null
null
```

`JSON_QUERY WITH CONDITIONAL ARRAY WRAPPER`:
```python
def test_json_query_w_conditional_array_wrapper(table_env):
    dataframe = pandas.DataFrame(
        {
            "js": [
                '{"a": [1,2,3,4], "b": 1}',
                '{"a":null,"b":2}',
                '{"a":"foo", "c":null}',
                "null",
                "[42,47,55]",
                "[]",
            ]
        }
    )
    table = table_env.from_pandas(dataframe)
    table.print()
    table_env.create_temporary_view("source_table", table)

    # Define the sink table
    table_env.execute_sql("""
        CREATE TABLE sink_table (
            res STRING
        ) WITH (
            'connector' = 'print'
        )
    """)

    # Define a transformation
    table_result = table_env.execute_sql("""
        INSERT INTO sink_table
        SELECT JSON_QUERY(source_table.`js`, '$.a' WITH CONDITIONAL ARRAY WRAPPER) AS `res`
        FROM source_table
    """)
    table_result.print()
```

Result:
```
[1,2,3,4]
null
["foo"]
null
null
null
```

ibis/backends/tests/test_map.py
deepyaman: @chloeh13q thoughts? Doesn't seem like passing `"connector": None` and getting

```
CREATE TABLE...
WITH(
    'connector' = None
)
```

doesn't seem it should be right

ibis/backends/flink/registry.py
```diff
@@ -50,7 +50,8 @@ def _cast(translator: ExprTranslator, op: ops.generic.Cast) -> str:
             return f"TO_TIMESTAMP(CONVERT_TZ(CAST({arg_translated} AS STRING), 'UTC+0', '{to.timezone}'))"
         else:
             return f"TO_TIMESTAMP({arg_translated})"
-
+    elif to.is_date():
+        return f"CAST({arg_translated} AS date)"
```
deepyaman, line 54: @mfatihaktas I removed "date" from `sql_type_mapping` and added this here; does it sound good for now?

Probably should look to make `TRY_CAST` consistent with the added logic here in the future, but I think it's not a priority immediately.

mfatihaktas: I think it looks good. Thanks!

ibis/backends/tests/test_map.py
chloeh13q: I just tested this in Flink and I think it's technically possible to create tables without table properties, i.e.
```sql
CREATE TABLE MyTable (
  `user_id` BIGINT,
  `item_id` BIGINT,
  `behavior` STRING
);
```
This would assume a managed table (where the connection information is already filled in the session environment; as opposed to an external table) & would throw `org.apache.flink.table.api.ValidationException` if such managed table is not found. Therefore, I think we can remove this validation & the exception.

That being said, we can also mark this test as not supported / not implemented in Flink backend and create a separate test somewhere else testing this functionality (given the slight difference in expected behavior as compared to other backends). That's what we did with other tests around `create_table()`: https://github.com/ibis-project/ibis/blob/master/ibis/backends/flink/tests/test_ddl.py .

deepyaman: I would prefer to remove this validation, if that can allow for a more uniform API. There are several instances of this, and would like to avoid having to replicate a lot of tests where it doesn't seem necessary (i.e. it looks like just dropping the difference would be fine).

ibis/backends/tests/test_client.py
mfatihaktas: I think that appeared after I switched to using `param` for the parameters.

ibis/backends/tests/test_map.py
deepyaman: Will address this in a follow-up PR, as it will require modifying `create_table` (don't want that side effect to get conflated with this PR).

deepyaman:
Note: force push from [03a5f72 ](https://github.com/ibis-project/ibis/commit/03a5f7287c0247c03d58fd0c8d33f57040b48bd9)to [e995674 ](https://github.com/ibis-project/ibis/commit/e995674f45c84a47d2f36b287cc0855671770bb0) is [nearly equivalent](https://github.com/ibis-project/ibis/compare/03a5f7287c0247c03d58fd0c8d33f57040b48bd9..e995674f45c84a47d2f36b287cc0855671770bb0), and just meant to squash down to 3 commits for merge.

ibis/backends/tests/test_map.py
chloeh13q: sounds reasonable to me

ibis/backends/tests/test_aggregation.py
```diff
@@ -80,21 +85,24 @@ def mean_udf(s):
...
+                    "mysql",
+                    "mssql",
+                    "pyspark",
+                    "trino",
+                    "druid",
+                    "oracle",
+                ],
+                raises=com.OperationNotDefinedError,
+            ),
+            pytest.mark.broken(["flink"], "WIP", raises=com.OperationNotDefinedError),
```
gforsyth, line 104: We try to reserve using `broken` for things that are both truly broken and also unrelated to the PR in question (maybe some upstream thing in a different backend breaks while you are adding functionality for a separate backend).

All of these that are marked as WIP should be changed over to either `notimpl` or `notyet`, depending on whether it's planned work or waiting on upstream improvements.

ibis/backends/tests/test_client.py
```diff
@@ -94,7 +104,18 @@ def test_create_table(backend, con, temp_table, lamduh, sch):
...
+            temp_table,
+            obj,
+            schema=sch,
+            tbl_properties={
+                "connector": None,
+            },
+        )
+    else:
+        con.create_table(temp_table, obj, schema=sch)
+
```
gforsyth, line 118: I know unifying `create_table` is on the radar, but I think we can clean this up a bit more, how about defining the extra `kwargs` in a dictionary that is otherwise empty and then passing those in to a single `create_table` call?

```suggestion
    kwargs = {}
    if backend.name() == "flink":
        kwargs = dict(tbl_properties={"connector": None})
    con.create_table(temp_table, obj, schema=sch, **kwargs)
```

ibis/backends/tests/test_client.py
```diff
@@ -834,10 +885,24 @@ def test_self_join_memory_table(backend, con):
         param(
             ibis.memtable([("a", 1.0)], columns=["a", "b"]),
             id="python",
+            marks=[
+                pytest.mark.notimpl(
+                    ["flink"],
+                    raises=NotImplementedError,
+                    reason="NotImplementedError('flink') raised in repr()",
```
gforsyth, line 892: totally fine to have a `reason` in here, but `reason` isn't required for a `notimpl`, the implicit reason always being "we haven't implemented it"

ibis/backends/tests/test_generic.py
```diff
@@ -61,6 +66,7 @@
 
 
 @pytest.mark.broken(["impala", "bigquery"], "assert nan is None")
+@pytest.mark.broken(["flink"], "The runtime does not support untyped `NULL` values.")
```
gforsyth, line 69:
```suggestion
@pytest.mark.notyet(["flink"], "The runtime does not support untyped `NULL` values.")
```

ibis/backends/tests/test_json.py
gforsyth: This should be marked as `broken` or `notimpl`. There are a few tests where we expect different results based on the backend, but all the other JSON-supporting backends do the same thing here.

ibis/backends/tests/test_json.py
```diff
@@ -48,6 +56,14 @@ def test_json_getitem(json_t, expr_fn, expected):
...
     ["pyspark", "trino"], reason="should work but doesn't deserialize JSON"
 )
+@pytest.mark.notimpl(
+    ["flink"],
+    raises=OperationNotDefinedError,
+    reason=(
+        "No translation rule for <class 'ibis.expr.operations.json.ToJSONMap'"
+        "Flink does not support ToJSONMap"
+    ),
+)
```
gforsyth, line 66: if the upstream doesn't support it, this should be marked as a `notyet`

ibis/backends/tests/test_json.py
```diff
@@ -72,6 +88,14 @@ def test_json_map(json_t):
     ["pyspark", "trino"], reason="should work but doesn't deserialize JSON"
 )
 @pytest.mark.notyet(["bigquery"], reason="doesn't allow null in arrays")
+@pytest.mark.notimpl(
```
gforsyth, line 91: if the upstream doesn't support it, this should be marked as a `notyet`

ibis/backends/tests/test_temporal.py
```diff
@@ -317,6 +331,16 @@ def test_timestamp_extract_milliseconds(backend, alltypes, df):
...
...69600, 1288569660, 1288569720, 1288569780, 1288569840, 1288569900, 1288569960, 1288570020, 1288570080, 1288570140, 1288656600, 1288656660, 1288656720, 1288656780, 1288656840, 1288656900, 1288656960, 1288657020, 1288657080, 1288657140, 1288743600, 1288743660, 1288743720, 1288743780, 1288743840, 1288743901, 1288743961, 1288744021, 1288744081, 1288744141, 1288830601, 1288830661, 1288830721, 1288830781, 1288830841, 1288830901, 1288830961, 1288831021, 1288831081, 1288831141, 1288917601, 1288917661, 1288917721, 1288917781, 1288917841, 1288917901, 1288917961, 1288918022, 1288918082, 1288918142, 1289004602, 1289004662, 1289004722, 1289004782, 1289004842, 1289004902, 1289004962, 1289005022, 1289005082, 1289005142, 1289091602, 1289091662, 1289091722, 1289091782, 1289091842, 1289091902, 1289091962, 1289092022, 1289092082, 1289092143, 1289178603, 1289178663, 1289178723, 1289178783, 1289178843, 1289178903, 1289178963, 1289179023, 1289179083, 1289179143, 1289265603, 1289265663, 1289265723, 1289265783, 1289265843, 1289265903, 1289265963, 1289266023, 1289266083, 1289266143, 1289352604, 1289352664, 1289352724, 1289352784, 1289352844, 1289352904, 1289352964, 1289353024, 1289353084, 1289353144, ...]"
```
gforsyth, line 341: The messages in reason shouldn't contain this entire pytest diff, you can include the first line and that's enough.

ibis/backends/tests/test_temporal.py
gforsyth: If Flink can't produce the expected values, then it fails this test.

We can add separate tests just for Flink in the backend-specific locations

gforsyth (changes requested):

ibis/backends/tests/test_aggregation.py
deepyaman: @gforsyth All of the "WIP" are from me, just as a way to mark things that I didn't even look into/just marked as failing (as suggested in the sync call), in order to get a fully-green test suite. I can switch them all to `notimpl`, as you recommend.

The first thing we are doing is going through the "WIP" ones, to actually evaluate what needs to be done on them.

ibis/backends/tests/test_temporal.py
deepyaman: > We can obviously append it with CONCAT (after mapping the given timezone to the correct literal) but I was not sure if that would be appropriate.

@mfatihaktas I think this seems pretty reasonable? I can mark it `notimpl` for now, but with a TODO to do this.

ibis/backends/tests/test_json.py
deepyaman: @mfatihaktas I played around with this a bit on Friday, and while I can figure out how to check whether something is a JSON object, I wasn't able to find a way to actually construct a map from a string JSON object (even though it sounds like it should be doable...). I'm happy to go with `not yet`, as @gforsyth suggests, unless you know otherwise.

ibis/backends/tests/test_temporal.py
```diff
@@ -317,6 +331,9 @@ def test_timestamp_extract_milliseconds(backend, alltypes, df):
     pyspark=["pandas<2.1"],
     reason="test was adjusted to work with pandas 2.1 output; pyspark doesn't support pandas 2",
 )
+@pytest.mark.broken(
+    ["flink"], raises=AssertionError, reason="numpy array are different"
+)
```
deepyaman, line 336: @mfatihaktas I went ahead and removed the extra-long error message, as per Gil's comment. I'm curious to understand why this test can't be made to work, though? I saw the translation is not including the string argument to provide timestamp format; do you think that could be related?

ibis/backends/tests/test_temporal.py
deepyaman: Done; following up to understand why it's broken in https://github.com/ibis-project/ibis/pull/6920#discussion_r1375464423

ibis/backends/tests/test_client.py
deepyaman: Agreed; removed for all such instances. FYI @mfatihaktas

ibis/backends/tests/test_client.py
deepyaman: Did something similar!

ibis/backends/tests/test_aggregation.py
deepyaman: Done!

ibis/backends/tests/test_temporal.py
mfatihaktas: > I can mark it notimpl for now, but with a TODO to do this.

Sounds good to me. Thanks!

ibis/backends/tests/test_temporal.py
mfatihaktas: > I went ahead and removed the extra-long error message, as per Gil's comment.

Thanks.

> I saw the translation is not including the string argument to provide timestamp format; do you think that could be related?

Sorry, I could not understand what you mean by "string argument to provide timestamp format". I also cannot find which test function this comment was for; it is outdated and I could not locate the relevant function.

mfatihaktas: @deepyaman Thanks for the details. In my opinion, we can mark this with `TODO` and work on it during our second iteration of the tests.

gforsyth (approved):
This seems like a solid starting point to me.

gforsyth:
Hey @deepyaman -- I think this is good to go, can you fix up the conflicts and then we can merge?

deepyaman:
> Hey @deepyaman -- I think this is good to go, can you fix up the conflicts and then we can merge?

@gforsyth Done!

PR #6922, opened 2023-08-23
fix(sqlglot): ensure back compat for `DataTypeParam` import

cpcloud:
Adds backcompat imports for some recently renamed objects in sqlglot 17.15.0. The TPC-H job exposed this failure because it has no dependency pins for better or worse.


gforsyth:
> because it has no dependency pins for better or worse.

It's a feature! (?)

gforsyth (approved):

PR #6923, opened 2023-08-23
feat(flink): implement table-related ddl in Flink backend to support streaming connectors

chloeh13q:
Implements table-related ddl in Flink backend to support streaming connectors.

### DIFF
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -10,9 +10,15 @@ from pyflink.table.types import create_arrow_schema
 import ibis.common.exceptions as exc
 import ibis.expr.operations as ops
 import ibis.expr.schema as sch
-from ibis.backends.base import BaseBackend, CanListDatabases
+from ibis.backends.base import BaseBackend, CanCreateDatabase
 from ibis.backends.base.sql.ddl import fully_qualified_re, is_fully_qualified
 from ibis.backends.flink.compiler.core import FlinkCompiler
+from ibis.backends.flink.ddl import (
+    CreateDatabase,
+    CreateTableFromConnector,
+    DropDatabase,
+    DropTable,
+)
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -24,7 +30,7 @@ if TYPE_CHECKING:
     import ibis.expr.types as ir
 
 
-class Backend(BaseBackend, CanListDatabases):
+class Backend(BaseBackend, CanCreateDatabase):
     name = "flink"
     compiler = FlinkCompiler
     supports_temporary_tables = True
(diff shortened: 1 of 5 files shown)


ibis/backends/flink/__init__.py
```diff
@@ -148,11 +253,25 @@ def create_table(
...
+
+        VIEWS can be created from an existing Table object, usually the result
+        of a Table API or SQL query. TABLES describe external data, such as a
+        file, database table, or message queue. In other words, TABLES refer
+        explicitly to tables constructed directly from source/sink connectors.
+
+        In `pyflink`, `create_temporary_view()` used to be `register_table()`,
+        and `create_table()` used to be `register_table_source()` and
+        `register_table_sink()`.
+
```
deepyaman, line 274:
```suggestion
```
Good point from an implementation perspective, but I don't think it needs to be go in the docstring (especially because it's more a historical thing).

ibis/backends/flink/__init__.py
```diff
@@ -180,17 +306,40 @@ def create_table(
         import pandas as pd
         import pyarrow as pa
 
+        import ibis.expr.types as ir
+
         if obj is None and schema is None:
             raise exc.IbisError("The schema or obj parameter is required")
+
+        if overwrite is True:
```
deepyaman, line 314:
```suggestion
        if overwrite:
```

ibis/backends/flink/ddl.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def _storage(self) -> str:
+        return f"STORED AS {self.format}" if self.format else None
+
+    def _format_tbl_properties(self) -> str:
+        return f"WITH {_format_properties(self.tbl_properties)}"
+
+    @property
+    def _prefix(self) -> str:
+        if self.temp is True:
```
deepyaman, line 68:
```suggestion
        if self.temp:
```

ibis/backends/flink/ddl.py
```diff
@@ -0,0 +1,154 @@
...
+            # This is not documented in Flink's documentation
+            return "CREATE TEMPORARY TABLE"
+        else:
+            return "CREATE TABLE"
```
deepyaman, line 72:
```suggestion
        return f"CREATE {"TEMPORARY " if self.temp else ""} TABLE"
```
Nit: I like the condensed expression :)

ibis/backends/flink/ddl.py
```diff
@@ -0,0 +1,154 @@
...
+        self.table_name = table_name
+        self.database = database
+        self.catalog = catalog
+        self.temp = temp
+
+    def _object_name(self):
+        return self._get_scoped_name(self.table_name, self.database, self.catalog)
+
+    def compile(self):
+        temp = "TEMPORARY " * self.temp
```
deepyaman, line 107: Interesting, never seen this 😄 on my point about rendering "TEMPORARY " before, if this clever way is preferable, can just use that above too.

ibis/backends/flink/ddl.py
```diff
@@ -0,0 +1,154 @@
...
+            f"WITH {_format_properties(self.db_properties)}"
+            if self.db_properties
+            else ""
+        )
+
+    def compile(self):
+        create_decl = "CREATE DATABASE"
+        create_line = f"{create_decl} {self._if_exists()}{self._object_name()}"
+
+        return create_line + "\n" + self._format_db_properties()
```
deepyaman, line 145:
```suggestion
        return "\n".join(create_line, self._format_db_properties())
```
Nit: IIRC `.join` is usually more efficient than string concatenation + more Pythonic, but I could be remembering wrong. Not that efficiency really matters in the two-element case.

deepyaman (commented):
Left a bunch of minor comments, but looks fine to me on a pass through.

ibis/backends/flink/__init__.py
jcrist: IMO docstrings here may want to add details about how concepts map to backend-specific details when relevant to the user. This seems to much of a historical/implementation detail though to document.

ibis/backends/flink/ddl.py
jcrist: I actually find the current way more readable, but either works. Personally I'd go with:

```python
modifier = " TEMPORARY" if self.temp else ""
return f"CREATE{modifier} TABLE"
```

ibis/backends/flink/ddl.py
jcrist: While clever, I'd find writing the if-else more readable. Not a big deal though either way.

```suggestion
        temp = "TEMPORARY " if self.temp else ""
```

ibis/backends/flink/ddl.py
jcrist: `.join` is more efficient, but that likely doesn't matter much here. In cases when joining a known number of strings like this, I'd use an f-string instead:

```python
f"{create_line}\n{self._format_db_properties()}"
```

jcrist (commented):
Overall LGTM! A few small code nits that should be cleaned up, but otherwise this is good-to-go. Thanks y'all!

chloeh13q:
@jcrist Thanks! I addressed all the comments and all the tests have passed.

PR #6929, opened 2023-08-24
chore(dev-deps): check poetry version when invoking `just lock`

### LINKED ISSUE #6926: build: centralize poetry version pin

Calling `just lock` calls `poetry lock --no-update` which usually ends up in a heavily rewritten `poetry.lock` file mostly because my local poetry version doesn't correspond to the poetry version we use in the CI.

We have a couple of places in the codebase pinning poetry < 1.4, it is also advised by the developer docs, but honestly I wouldn't like to manually check whether my poetry version corresponding to the expected one. 

I'd like to just blindly call `just lock` and don't care about the poetry version at all because it would ensure that both my local setup and CI use the same poetry version. 

6.1

_No response_

### Code of Conduct

cpcloud:
Closes #6926.

### DIFF
--- a/justfile
+++ b/justfile
@@ -8,6 +8,12 @@ clean:
 
 # lock dependencies without updating existing versions
 lock:
+    #!/usr/bin/env bash
+    version="$(poetry --version)"
+    if ! grep -qP '\(version 1\.3\.\d+\)' <<< "${version}"; then
+        >&2 echo "poetry version must be 1.3.x, got ${version}"
+        exit 1
+    fi
     poetry lock --no-update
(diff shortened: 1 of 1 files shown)


kszucs (approved):

kszucs:
What happens when we want to bump the poetry version? Where do we need to update the version number?

cpcloud:
> What happens when we want to bump the poetry version? Where do we need to update the version number?

In multiple places. We'll have to grep around for this. AFAICT there's no _fast to set up_ way to strictly manage poetry versions other than to enforce its version at install time.

PR #6948, opened 2023-08-28
fix(oracle): clarify sid vs service_name handling and allow dsn

### LINKED ISSUE #6913: feat: Support for OracleDB thick mode

I consider using a tool that uses ibis as a dependency. I need to connect to Oracle DB. I am aware that only THICK python-oracledb clients are permitted to connect to my DB.

I see that [ibis uses the python-oracledb "thin client" to connect to Oracle databases](https://ibis-project.org/backends/oracle/#connecting-to-older-oracle-databases), which means that if it's true, I won't be able to use it.

Is my thinking correct?

I would like ibis to support oracledb thick mode: https://python-oracledb.readthedocs.io/en/latest/user_guide/initialization.html#enabling-python-oracledb-thick-mode

6.1.0

Oracle

### Code of Conduct

gforsyth:
Maybe resolves #6913 

just so I don't forget this _again_:
A `sid` is a unique name of an INSTANCE running an oracle process (a single, identifiable machine).
A `service name` is an ALIAS to one (or many) individual instances that can be hotswapped without the client knowing / caring -- so a `service_name` might map to one or more `sid`s but this would be opaque to the user.
A `dsn` is a 'data source name' and is a very lispy connection definition that the oracle python clients can parse for things like `sid`, `service_name`, `host`, `port`, etc.

In order to allow for maximal connection flexibility, this PR lets users pass in a DSN directly, which will override every other keyword argument.  If they don't pass in a DSN, we construct one from the provided keyword arguments (like `host`, `port`, `sid`, `service_name`).

I've tested this locally and it seems to work (although I don't understand how to specify my machine as a `sid`, so that particular bit of functionality is untested).

### DIFF
--- a/ibis/backends/oracle/__init__.py
+++ b/ibis/backends/oracle/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import atexit
 import contextlib
 import sys
+import warnings
 from typing import TYPE_CHECKING, Any
 
 import oracledb
@@ -24,6 +25,7 @@ sys.modules["cx_Oracle"] = oracledb
 
 import sqlalchemy as sa  # noqa: E402
 
+import ibis.common.exceptions as exc  # noqa: E402
 import ibis.expr.datatypes as dt  # noqa: E402
 import ibis.expr.operations as ops  # noqa: E402
 from ibis.backends.base.sql.alchemy import (  # noqa: E402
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
This seems reasonable to me. Shall we merge?

gforsyth:
Here we go!  @kangshung -- keep us posted if this helps out!

kangshung:
Thank you very much for adding this. It solved the issue with the db connection. Both `service_name` and `dsn` work fine.

PR #6950, opened 2023-08-28
fix(bigquery,impala): escape all ASCII escape sequences in string literals

### LINKED ISSUE #6949: bug(bigquery): memtable and string literals not escaping `\n` or `\` which results in invalid syntax
### What happened?

Code:

```python
import ibis

ibis_client = ibis.bigquery.connect()
table = ibis.memtable(
    {
        "col1": ["a\tb\nc", "d e f", "g'e\"h"],
    }
)
print(ibis_client.compile(table))
```

Output:

```
SELECT t0.*
FROM UNNEST(ARRAY<STRUCT<col1 STRING>>[STRUCT('a        b
c' AS col1), STRUCT('d e f' AS col1), STRUCT('g\'e"h' AS col1)]) t0
```

Note, the following SQL works as expected:

```
SELECT t0.*
FROM UNNEST(ARRAY<STRUCT<col1 STRING>>[STRUCT('a        b\nc' AS col1), STRUCT('d e f' AS col1), STRUCT('g\'e"h' AS col1)]) t0
```

Therefore, we should really be escaping `\n` in addition to `'`. Though, perhaps there are other characters that could break BigQuery syntax? See: https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#string_and_bytes_literals

Alternatively, using triple-quoted strings allows for newline characters in the string literal itself.

6.1.0

also tested on latest commit: 15f8d9575

BigQuery

### Relevant log output

```sh
BigQuery API: Syntax error: Unclosed string literal at [2:47]
```

### Code of Conduct

tswast:
Closes #6949


ibis/backends/base/sql/registry/literal.py
```diff
@@ -21,7 +21,13 @@ def _boolean_literal_format(translator, op):
 
 
 def _string_literal_format(translator, op):
```
tswast, line 23: Note: this is in the base sql class, not BigQuery-specific. I don't know if the other backends suffer the same limitation of no unescaped newlines in string literals, but it seems as though it wouldn't do any harm to have it.

ibis/backends/base/sql/registry/literal.py
```diff
@@ -21,7 +21,13 @@ def _boolean_literal_format(translator, op):
...
-    return "'{}'".format(op.value.replace("'", "\\'"))
+    return "'{}'".format(
+        op.value
+        # Escape \ first so we don't double escape other characters.
+        .replace("\\", "\\\\")
+        .replace("\n", "\\n")
+        .replace("'", "\\'")
```
cpcloud, line 29: Should we use `re.escape(op.value)` instead? That'll handle other whitespace characters like `\t` as well as `\n` and `\v` and any others that are easy to miss.

tswast: Looks like `re.escape` doesn't escape some things that we need escaped (`'`, `\n`) and does escape some things we don't want escaped (`~`, `#`, `(`, `)`, etc). I'll add a test that includes those characters.

tswast: Updated! Turns out this fix also fixes some problems encountered in Impala regex support.

tswast:
CC @jayceslesar 

I believe this may be related to the report you had here: https://github.com/ibis-project/ibis-bigquery/issues/110

It could be we also need to turn off the "raw" string now in the BQ regex functions. I'll run some tests locally to check.

ibis/backends/base/sql/registry/literal.py
```diff
@@ -21,7 +21,22 @@ def _boolean_literal_format(translator, op):
...
-    return "'{}'".format(op.value.replace("'", "\\'"))
+    return "'{}'".format(
+        op.value
+        # Escape \ first so we don't double escape other characters.
+        .replace("\\", "\\\\")
+        # Escape ' since we're using those for the string literal.
+        .replace("'", "\\'")
+        # ASCII escape sequences that are recognized in Python:
+        # https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
+        .replace("\a", "\\a")  # Bell
```
cpcloud, line 32: 🔔

ibis/backends/bigquery/tests/unit/snapshots/test_compiler/test_literal_string/escape_ascii_sequences/out.sql
```diff
@@ -0,0 +1 @@
+SELECT 'a\ab\bc\fd\ne\rf\tg\vh' AS `'a_x07b_x08c_x0cd_ne_rf_tg_x0bh'`
```
cpcloud, line 1: Wow, this is glorious!

cpcloud (approved):
Awesome, thanks for the fixes!

PR #6954, opened 2023-08-29
refactor(clickhouse): move `ClickhouseTable.insert` method to clickhouse backend and remove `ClickhouseTable` class

cpcloud:
Remove the unnecessary `ClickhouseTable` class in favor of using the backend `insert` method.


jcrist (approved):
👍👍

PR #6959, opened 2023-08-29
feat(api): `add Table.rename`, deprecate `Table.relabel`

### LINKED ISSUE #6801: feat: Pass in mapping to Table.relabel() as **kwargs

Often, I only am relabeling a single column using a simple boilerplate mapping. Needing to construct a dict with the string literals makes things less more verbose. I like the kwargs API of .select() and .mutate(), I wished .relabel() was consistent with that.

`t.relabel(**{"old": "new"})` should be able to be converted to `t.relabel(old="new")`

EDIT: this should actually be `t.mutate(new=_.old)`
Note that this is slightly different from `t.mutate(old=_.new)` (and thus I can't use that), because

1. column order is preserved with .relabel()
2. if "new" already exists, .relabel() errors, .mutate() does not

I think something like:

```python
# Use / to force `substitutions` to be positional-only
def relabel(self, substitutions = None, /, **kwargs: dict[str, str]) -> Table:
    if kwargs:
        if substitutions is not None:
            raise TypeError("Cannot specify both substitutions and kwargs")
        substitutions = kwargs
    # continue with old logic
```

main

NA

### Code of Conduct

jcrist:
As suggested in #6801, this:

- Adds a new `Table.rename` method that accepts column names as `new_name="old_name"` kwargs. This new method also swaps the order of key-value pairs when passing a mapping as an argument to `{"new_name": "old_name"}` for consistency.
- Deprecates `Tabel.relabel` in favor of `Table.rename`.

Fixes #6801.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -1729,23 +1729,45 @@ class Table(Expr, _FixedTextJupyterMixin):
         | str
         | Literal["snake_case", "ALL_CAPS"],
     ) -> Table:
+        """Deprecated in favor of `Table.rename`"""
+        if isinstance(substitutions, Mapping):
+            substitutions = {new: old for old, new in substitutions.items()}
+        return self.rename(substitutions)
+
+    def rename(
+        self,
+        method: str
+        | Callable[[str], str | None]
+        | Literal["snake_case", "ALL_CAPS"]
+        | Mapping[str, str]
+        | None = None,
+        /,
+        **substitutions: str,
+    ) -> Table:
         """Rename columns in the table.
 
         Parameters
         ----------
+        method
+            An optional method for renaming columns. May be one of:
+
+            - A format string to use to rename all columns, like
+              ``"prefix_{name}"``.
+            - A function from old name to new name. If the function returns
+              ``None`` the old name is used.
+            - The literal strings ``"snake_case"`` or ``"ALL_CAPS"`` to
+              rename all columns using a ``snake_case`` or ``"ALL_CAPS"``
+              naming convention respectively.
+            - A mapping from new name to old name. Existing columns not present
+              in the mapping will passthrough with their original name.
         substitutions
-            A mapping, function, or format string mapping old to new column
-            names. If a column isn't in the mapping (or if the callable returns
-            None) it is left with its original name. May also pass a format
-            string to rename all columns, like ``"prefix_{name}"``. Also
-            accepts the literal string ``"snake_case"`` or ``"ALL_CAPS"`` which
-            will relabel all columns to use a ``snake_case`` or ``"ALL_CAPS"``
-            naming convention.
+            Columns to be explicitly renamed, expressed as ``new_name=old_name``
+            keyword arguments.
 
         Returns
         -------
         Table
-            A relabeled table expression
+            A renamed table expression
 
         Examples
         --------
@@ -1773,9 +1795,10 @@ class Table(Expr, _FixedTextJupyterMixin):
         │ …         │             … │ …                                   │
         └───────────┴───────────────┴─────────────────────────────────────┘
 
-        Relabel column names using a mapping from old name to new name
+        Rename specific columns by passing keyword arguments like
+        ``new_name="old_name"``
 
-        >>> t.relabel({"studyName": "study_name"}).head(1)
+        >>> t.rename(study_name="studyName").head(1)
         ┏━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
         ┃ study_name ┃ Sample Number ┃ Species                             ┃
         ┡━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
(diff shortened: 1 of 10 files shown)


jcrist:
Gah, the impala backend has an existing `ImpalaTable.rename` method that shadows this one. I _think_ we can just remove that method completely, cc @cpcloud for confirmation.

cpcloud:
😞 Yeah, that's a DDL method that alters the table metadata.

Looks like it's on pyspark too.

Would you mind moving those `rename` methods to `rename_table` on the backend?

Or, if you don't feel like doing that here you can delete them and I can resurrect them as backend methods in a follow-up PR.

jcrist:
> Would you mind moving those rename methods to rename_table on the backend?

Sure, I'll take care of it

ibis/expr/types/relations.py
```diff
@@ -1722,30 +1722,58 @@ def select(
...
+        return self.rename(substitutions)
+
+    def rename(
+        self,
+        method: str
+        | Callable[[str], str | None]
+        | Literal["snake_case", "ALL_CAPS"]
+        | Mapping[str, str]
+        | None = None,
+        /,
```
NickCrews, line 1751: nice use of forcing positional

NickCrews:
What about if we kept the API smaller, and removed support for supplying `Mapping[str, str]`, instead relying on just **kwargs? If I have `m= {"new": "old"}`, it is pretty easy to just unpack it as `t.rename(**m)`

ibis/expr/types/relations.py
```diff
@@ -1722,30 +1722,58 @@ def select(
...
+    ) -> Table:
+        """Deprecated in favor of `Table.rename`"""
+        if isinstance(substitutions, Mapping):
+            substitutions = {new: old for old, new in substitutions.items()}
+        return self.rename(substitutions)
+
+    def rename(
+        self,
+        method: str
+        | Callable[[str], str | None]
```
NickCrews, line 1747: document what returning `None` from the lambda does. I think it results in no renaming? I would prefer if it were simpler: you always had to return a str, and if you wanted no renaming, then just return the original string.

ibis/expr/types/relations.py
```diff
@@ -1828,33 +1857,52 @@ def relabel(
...
 
-        if isinstance(substitutions, Mapping):
-            rename = substitutions.get
-        elif substitutions in {"snake_case", "ALL_CAPS"}:
+        if method is None:
+
+            def rename(c):
+                return None
+
+        elif isinstance(method, str) and method in {"snake_case", "ALL_CAPS"}:
```
NickCrews, line 1883: consider extracting this out into a separate function, for easier unit testing (this logic is complex) and for a slimmer `def rename()`

ibis/expr/types/relations.py
```diff
@@ -1722,30 +1722,58 @@ def select(
...
 
         Parameters
         ----------
+        method
+            An optional method for renaming columns. May be one of:
+
+            - A format string to use to rename all columns, like
+              ``"prefix_{name}"``.
+            - A function from old name to new name.
+            - The literal strings ``"snake_case"`` or ``"ALL_CAPS"`` to
```
NickCrews, line 1764: total nit: "The literal string" not "The literal strings"

NickCrews:
Thanks @jcrist , this looks great! I have a few suggestions, some of them I feel strongly about, some are nits/style.

ibis/expr/types/relations.py
jcrist: There are two literal strings here, having it plural seems more grammatically correct to me.

ibis/expr/types/relations.py
jcrist: The logic is fully tested in our existing test suite. Keeping it inline in the method that uses it [avoids indirection](https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection) making it easier for a reader to see what's going on.

ibis/expr/types/relations.py
jcrist: Sure, I'll update the docstring. This is existing behavior in `relabel`, I'd be reluctant to change it.

> if you wanted no renaming, then just return the original string.

The existence of support for `None` doesn't prevent you from doing this. If this is how you'd want to use this API, then that should work for you.

ibis/expr/types/relations.py
```diff
@@ -1795,9 +1835,9 @@ def relabel(
         │ PAL0708    │             1 │ Adelie Penguin (Pygoscelis adeliae) │
         └────────────┴───────────────┴─────────────────────────────────────┘
 
-        Relabel column names using a ALL_CAPS convention
+        Rename all columns using an ALL_CAPS convention
```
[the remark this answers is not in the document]
jcrist: I don't have a good dataset to show that with. This docstring was just ported from the old `relabel` docstring, leaving it the same for now.

ibis/expr/types/relations.py
NickCrews: ah, the existing behavior probably is the decider for this.

jcrist:
> I think we should keep the API smaller, and consistent with `.mutate()` and `.select()`, and remove support for supplying `Mapping[str, str]`, instead relying on just **kwargs. If I have `m= {"new": "old"}`, it is pretty easy to just unpack it as `t.rename(**m)`.

Both `mutate` and `select` support passing in `*args` instead, which fits a similar use case for a mapping here. Without a mapping, there's no (clean) way to rename columns to names that aren't valid python identifiers. You could construct a dict and splat it in, but that looks weird:

```python
t.rename(**{"not a valid python id": "old_column_name"})
```

I don't see the harm in supporting an explicit mapping for users that may want to make use of it.

ibis/expr/types/relations.py
NickCrews: Ah, I see that now. Thanks! I added a comment inline there, I think this could do with apostrophes, other whitespace chars besides spaces, and numbers.

NickCrews:
> I don't see the harm in supporting an explicit mapping for users that may want to make use of it.

If we add it here, then we should add it to .relabel() and .mutate() too, I think all 3 should be as consistent as possible.

> that looks weird

I think this is personal taste as to how nasty this looks. I agree I don't love it, but I don't find it too disgusting, and I think it definitely is worth it to keep the above consistency.

jcrist:
> If we add it here, then we should add it to .relabel() and .mutate() too, I think all 3 should be as consistent as possible.

Assuming you meant `.select` here. I don't think we need to be consistent across these operations. `select` and `mutate` are more similar, but `rename` is just for renaming columns.

The thing a user passes to `.rename` is a mapping from old names to new names - all options to `rename` are just various ways of spelling this.

The thing a user passes to `.select` is one or more column expressions to include in the output table - all options to `.select` are various spellings of this.

These are not semantically equivalent operations. You can't pass a column expression to the RHS of a keyword argument in `rename` (e.g. `t.rename(foo=_.foo + 1)` is invalid). While `.rename` is a wrapper around `.select`, I don't think they need to be 100% consistent in args here.

NickCrews:
> Assuming you meant .select here.

yup, thanks :)

> I don't think they need to be 100% consistent in args here.

I agree, but even though we can't have 100%, we should try to aim for 90% instead of 85%. But I can see the usability benefits of allowing Mapping possibly being worth it.

I think this might just be a matter of opinion, I'm fine being overruled. @cpcloud @lostmygithubaccount @kszucs you have opinions?

ibis/tests/expr/test_table.py
```diff
@@ -374,68 +374,107 @@ def test_filter_fusion_distinct_table_objects(con):
...
+        ("col_b", "ColB"),
+        ("col_c", "colC"),
+        ("col_d", "col-d"),
         ("col_e", "col_e"),
-        (" Column F ", "column_f"),
-        ("Column G-with-hyphens", "column_g_with_hyphens"),
-        ("Col H notCamelCase", "col_h_notcamelcase"),
+        ("column_f", " Column F "),
+        ("column_g_with_hyphens", "Column G-with-hyphens"),
+        ("col_h_notcamelcase", "Col H notCamelCase"),
```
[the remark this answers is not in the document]
jcrist: Going to leave these as is for now, this is unrelated to this PR.

ibis/tests/expr/test_table.py
```diff
@@ -374,68 +374,103 @@ def test_filter_fusion_distinct_table_objects(con):
...
+    sol = sch.schema({"x_1": "int32", "y_1": "string", "z": "double"})
     assert_equal(res, sol)
 
     # Using a function
     res = table.relabel(lambda x: None if x == "z" else f"{x}_1").schema()
     assert_equal(res, sol)
 
+    # Using a format string
+    res = table.relabel("_{name}_")
+    sol = table.relabel({"x": "_x_", "y": "_y_", "z": "_z_"})
```
cpcloud: Do these need to be underneath a `with pytest.warns`?

cpcloud (approved):
LGTM!

ibis/tests/expr/test_table.py
cpcloud: Ah, next commit :)

contang0:
I was just curious, what is the rationale behind swapping key-value order? I experienced an error when I switched from `relabel` to `rename`, and it wasn't immediately obvious to me that the new column names should now be keys.

lostmygithubaccount:
@contang0 it was discussed in the linked issue here: https://github.com/ibis-project/ibis/issues/6801

in short to match how other methods work

PR #6970, opened 2023-08-30
docs: draft posts to draft and very minor edits

lostmygithubaccount:


docs/posts/ibis-analytics/index.qmd
```diff
@@ -7,7 +7,7 @@ code-annotations: below
 categories:
     - blog
     - demo
-draft: false
+draft: true
```
cpcloud, line 10: What determines when a thing moves out of draft mode?

cpcloud (approved):

docs/posts/ibis-analytics/index.qmd
lostmygithubaccount: changing that line; it just won't show up in the posts until draft is false (or that line is removed, with default of false), so we can flip these over once i re-write

lostmygithubaccount: I want to re-write the ibis-analytics one a bit and the Quarto one also needs a rewrite. I'll try to get those in this week but don't want them showing publicly just yet

PR #6979, opened 2023-08-31
feat(flink): allow translation of decimal literals

deepyaman:


ibis/backends/flink/utils.py
```diff
@@ -282,6 +282,17 @@ def translate_literal(op: ops.Literal) -> str:
...
             raise ValueError("Infinity is not supported in Flink SQL")
+        elif dtype.is_decimal():
+            # When PyFlink infers schema from `decimal.Decimal` objects,
+            # it will be `DecimalType(38, 18)`.
+            # https://github.com/apache/flink/blob/release-1.17.1/flink-python/pyflink/table/types.py#L336-L337
+            precision = 38 if dtype.precision is None else dtype.precision
+            scale = 18 if dtype.scale is None else dtype.scale
+
+            if precision > 38:
+                raise ValueError("The precision can be up to 38 in Flink")
```
deepyaman, line 293: I don't see other backends/tests do this; however, if I don't raise an error early, the resulting error on the PyFlink side is something uninformative about writing a NULL to a non-nullable field. I assume this happens because, with invalid input, it makes the value NULL. Is this way to handle it OK?

ibis/backends/flink/utils.py
```diff
@@ -282,6 +282,17 @@ def translate_literal(op: ops.Literal) -> str:
             raise ValueError("NaN is not supported in Flink SQL")
         elif math.isinf(value):
             raise ValueError("Infinity is not supported in Flink SQL")
+        elif dtype.is_decimal():
+            # When PyFlink infers schema from `decimal.Decimal` objects,
+            # it will be `DecimalType(38, 18)`.
```
deepyaman, line 287: It defaults to (10, 0), but using this, since at the `ibis.literal` seems to be constructed from `decimal.Decimal`.

ibis/backends/flink/utils.py
```diff
@@ -282,6 +282,17 @@ def translate_literal(op: ops.Literal) -> str:
...
+            # https://github.com/apache/flink/blob/release-1.17.1/flink-python/pyflink/table/types.py#L336-L337
+            precision = 38 if dtype.precision is None else dtype.precision
+            scale = 18 if dtype.scale is None else dtype.scale
```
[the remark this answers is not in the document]
jcrist: A `TypeParser` is for parsing type info (as strings) when reflecting existing tables from the backend into ibis. When ibis connects to a postgres database and tries to load a table `mytable` it will first query the backend to get the schema of that table (usually represented as strings in a backend-specific syntax). The `TypeParser` is what's used to convert those type strings into ibis types.

Here you're handling translations of ibis literals into SQL, so the `TypeParser` abstraction isn't relevant.

jcrist (approved):

PR #6980, opened 2023-08-31
chore: backport a few fixes and one feature to `6.x.x`

cpcloud:
This PR backports a few fixes to the `6.x.x` branch from `master`, and sets up the ability to release from the `6.x.x` branch.

Backporting was requested by @tswast (for BigQuery DataFrames), and also by some folks from the trino side so they can measure ibis + trino usage.

After this PR is merged, I'll cut a 6.2 release using the `release.yml` workflow run from the 6.x.x branch. This PR includes a change to `.releaserc.js` that allow `semantic-release` to run from the 6.x.x branch, which should be enough to allow it to successfully cut a release.

PR #6988, opened 2023-08-31
build: bump the major version to ensure that prereleases are always sorted after existing versions

cpcloud:
This PR changes pre-release versions to always start with the next major version, so that any releases in between major releases such as backports (#6980) will not be sorted _after_ a pre-release.

The change is visible by running `poetry build` from this PR's commit:

```
❯ poetry build
Building ibis-framework (7.0.0.dev217)
  - Building sdist
  - Built ibis_framework-7.0.0.dev217.tar.gz
  - Building wheel
  - Built ibis_framework-7.0.0.dev217-py3-none-any.whl
```

Note the 7.0.0 prefix.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -532,7 +532,8 @@ enable = true
 dirty = true
 style = "pep440"
 pattern = '^(?P<base>\d+(\.\d+)*)'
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
:shipit:

PR #6995, opened 2023-08-31
ci: build with vendored version of pandoc and bump quarto version

cpcloud:
This PR bumps to the most recent quarto pre-release version and removes the nix use of pandoc in favor of the binary that quarto ships directly. This cuts down quite a lot of the size of the dependency closure.


cpcloud:
@lostmygithubaccount This should address the buttons showing up at the bottom of the landing page

image

PR #7007, opened 2023-09-01
chore: render do_connect arguments and file support backend APIs using quartodoc

cpcloud:
This PR fixes rendering of backend specific connection API docs and file support API docs


docs/backends/clickhouse.qmd
```diff
@@ -77,17 +77,20 @@ con = ibis.clickhouse.connect(
 ```
 
 ::: {.callout-note}
-ibis.clickhouse.connect` is a thin wrapper around [`ibis.backends.clickhouse.Backend.do_connect`][ibis.backends.clickhouse.Backend.do_connect].
+`ibis.clickhouse.connect` is a thin wrapper around [`ibis.backends.clickhouse.Backend.do_connect`](ibis.backends.clickhouse.Backend.do_connect).
```
cpcloud, line 80: Still need to figure out how to do the interlinks. Looks like a new version of quartodoc (0.6.0) was released that might better support this.

docs/backends/bigquery.qmd
```diff
@@ -77,17 +77,21 @@ con = ibis.bigquery.connect(
...
-      heading_level: 4
-<!-- prettier-ignore-end -->
+```{python}
+#| echo: false
+#| output: asis
+from quartodoc import get_object, MdRenderer
+
+renderer = MdRenderer(header_level=4)
+obj = get_object("ibis.backends.bigquery", "Backend")
+print(renderer.render(obj.get_member("do_connect")))
```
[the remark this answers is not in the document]
cpcloud: Ok, my desire to avoid copypasta got the better of me, there's now a `_utils.py` module that wraps up a few of the repetitive bits into a few functions.

docs/backends/clickhouse.qmd
cpcloud: Figured this out and it didn't require a version bump, but #7013 will be bumping for other reasons.

PR #7009, opened 2023-09-01
chore(polars): pin polars>=0.19 due to the recent deprecations

kszucs:
Multiple methods have been deprecated with the new release: https://github.com/pola-rs/polars/releases

This might not be the best but it is the quickest way to avoid deprecation warnings in out test suite.


cpcloud:
If they are simple deprecations you can also try the new way and fallback to the old way.

kszucs:
Updating, but that is a bit slower to implement :)

cpcloud:
@kszucs Let's just increase the lower bound for now. Polars has enough breakage between releases (no judgment!) that there's not a lot of reason to support a huge range of versions.

cpcloud (approved):

PR #7017, opened 2023-09-02
docs: fix self and external links

cpcloud:
This PR fixes the remaining interlinks between ibis documentation and itself and links to inventories like the core Python docs.

### DIFF
--- a/docs/posts/_metadata.yml
+++ b/docs/posts/_metadata.yml
@@ -2,7 +2,7 @@
 
 # freeze computational output
 # (see https://quarto.org/docs/projects/code-execution.html#freeze)
-freeze: true
+freeze: auto
 
 # Enable banner style title blocks
 title-block-banner: true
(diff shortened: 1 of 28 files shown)


docs/posts/_metadata.yml
```diff
@@ -11,6 +11,5 @@ title-block-banner: true
 comments:
   giscus:
     repo: ibis-project/ibis
-    theme: light
```
cpcloud, line 14: Removing the explicit `light` theme makes the theme automatic based on whether the site theme is light or dark 🚀

docs/posts/_metadata.yml
```diff
@@ -2,7 +2,7 @@
 
 # freeze computational output
 # (see https://quarto.org/docs/projects/code-execution.html#freeze)
-freeze: true
+freeze: auto
```
cpcloud, line 5: This causes recomputation if the source of a post changes.

PR #7067, opened 2023-09-06
docs: fix edit this page button

lostmygithubaccount:
we should consider switching to `main` as the default branch; the `repo-subdir` will be needed either way

docs: https://quarto.org/docs/websites/website-navigation.html#github-links


jcrist (approved):
:shipit:

PR #7068, opened 2023-09-06
docs: remove stray bracket

lostmygithubaccount:
this was the change I wanted to edit through the browser! removes a single stray bracket in a how-to guide


jcrist (approved):
Nice catch!

PR #7069, opened 2023-09-06
feat(duckdb): expose loading extensions

### LINKED ISSUE #7063: feat: user-facing method for installing duckdb extensions w/o SQL

If I want to use DuckDB with an extension such as httpfs, I have to drop into SQL to install and load it. Due to this, I might run into a situation where I set up a connection that attempts to configure behavior provided by a non-installed, non-loaded extension,

```python
c = ibis.duckdb.connect("database.ddb", s3_region="us-east-2")
```

This returns successfully. But if I run

```python
c.raw_sql("SELECT current_setting('s3_region')")
```

I get:

> (duckdb.CatalogException) Catalog Error: Setting with name "s3_region" is not in the catalog, but it exists in the httpfs extension.
> 
> To install and load the extension, run:
> INSTALL httpfs;
> LOAD httpfs;

This isn't hard to work around if you know about DuckDB and `raw_sql` but it might trip users up and could overall be improved.

It would be nice if I could specify extensions to load with the `ibis.duckdb.connect` call, similar to how I can specify configuration options as kwargs. It might also be nice to handle the above error message in a more useful way though that may be out of scope of this feature request.

6.2.0

DuckDB

### Code of Conduct

jcrist:
- Adds an `extensions` option to `ibis.duckdb.connect` for automatically installing/loading extensions by name on connect
- Adds a `con.load_extension` method for installing/loading an extension on an existing connection

~Still needs tests, will finish up tomorrow.~

Fixes #7063.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -6,7 +6,6 @@ import ast
 import contextlib
 import os
 import warnings
-from functools import partial
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
@@ -34,7 +33,7 @@ from ibis.expr.operations.udf import InputType
 from ibis.formats.pandas import PandasData
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable, Iterator, Mapping, MutableMapping
+    from collections.abc import Iterable, Iterator, Mapping, MutableMapping, Sequence
 
     import pandas as pd
     import torch
@@ -173,6 +172,7 @@ WHERE catalog_name = :database"""
         database: str | Path = ":memory:",
         read_only: bool = False,
         temp_directory: str | Path | None = None,
+        extensions: Sequence[str] | None = None,
         **config: Any,
     ) -> None:
         """Create an Ibis client connected to a DuckDB database.
@@ -186,6 +186,8 @@ WHERE catalog_name = :database"""
         temp_directory
             Directory to use for spilling to disk. Only set by default for
             in-memory connections.
+        extensions
+            A list of duckdb extensions to install/load upon connection.
         config
             DuckDB configuration parameters. See the [DuckDB configuration
             documentation](https://duckdb.org/docs/sql/configuration) for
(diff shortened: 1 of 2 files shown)


ibis/backends/duckdb/__init__.py
```diff
@@ -237,31 +241,36 @@ def configure_connection(dbapi_connection, connection_record):
 
         super().do_connect(engine)
 
+    @staticmethod
+    def _sa_load_extensions(dbapi_con, extensions):
+        query = """
+        SELECT extension_name, aliases
+        FROM duckdb_extensions()
+        WHERE installed AND loaded
+        """
```
jcrist, line 250: The previous query would ignore extensions that weren't known to duckdb (`duckdb_extensions()` returns known or installed extensions), meaning you couldn't use it to install any custom extensions. I couldn't figure out a way to do the filtering I wanted fully in SQL. The data we're querying is super small though, so doing it manually in python should be fine.

IndexSeek: There may be a way to incorporate an `allow_unsigned_extensions` argument that would allow you to load your own extensions or third-party extensions, overriding the check or something. That might could get messy.

IndexSeek: I fiddled around with the query, and might need to think through it, but the small dataset makes me think it is okay. 

```python
con.sql("""SELECT extension_name 
           FROM duckdb_extensions() 
           UNION SELECT UNNEST(aliases) 
           FROM duckdb_extensions();"""
```

I was looking to see if the goal was to unnest the aliases (with the `installed.update`) we might could use the `UNNEST` function in SQL to reduce a little bit of the Python. 

It might need some tweaking with the installed and loaded predicates if those need to be used as part of an inner/outer query.

jcrist: Thanks for the suggestion (`UNNEST` is a new one for me), with that prompting I was able to get the set of installed extension names fully in SQL. Probably doesn't matter either way for performance, but it does simplify the code a bit.

cpcloud (approved):

PR #7071, opened 2023-09-06
docs: basic starburst galaxy tutorial

lostmygithubaccount:
have some other parts in drafts, but want to get this minimal PR in. shows the basics of connecting and using Starburst Galaxy with Ibis

I'm not sure on the location of this -- I think it might make sense to have a bunch of data platform-specific tutorials? but need to avoid sprawl too

we could move this to the backends documentation like BigQuery, or consider turning that into a tutorial

feel free to make any edits


lostmygithubaccount:
failing because `python-dotenv` is not installed -- would also fail due to not having the environment variables -- trying to freeze execution

cpcloud (approved):
LGTM! Thanks!

PR #7075, opened 2023-09-07
docs: add exending how-to guides

### LINKED ISSUE #7074: bug: docs: extending guide isn't in nav
### What happened?

The page at https://ibis-project.org/how-to/extending/elementwise.html isn't publicly findable from any menu. I only could find it because I knew what to search for from my experience using the old version of the site. (nice job on the new site BTW!!)

NA

Na

### Relevant log output

_No response_

### Code of Conduct

lostmygithubaccount:
these look fine to me, just missed in the `_quarto.yml`

closes #7074

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -120,6 +120,7 @@ website:
         - auto: "how-to/input-output"
         - auto: "how-to/analytics"
         - auto: "how-to/visualization"
+        - auto: "how-to/extending"
     - id: contribute
       title: "Contribute"
       style: "docked"
--- a/docs/how-to/extending/elementwise.qmd
+++ b/docs/how-to/extending/elementwise.qmd
@@ -4,7 +4,7 @@ This notebook will show you how to add a new elementwise operation to an existin
 
 We are going to add `julianday`, a function supported by the SQLite database, to the SQLite Ibis backend.
 
-The Julian day of a date, is the number of days since January 1st, 4713 BC. For more information check the [Julian day](https://en.wikipedia.org/wiki/Julian_day) wikipedia page.
+The Julian day of a date, is the number of days since January 1st, 4713 BC. For more information check the [Julian day](https://en.wikipedia.org/wiki/Julian_day) Wikipedia page.
 
 ## Step 1: Define the Operation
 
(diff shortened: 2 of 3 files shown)


NickCrews:
nice!

Another important detail: the current version of the tutorial doesn't work in the released version of Ibis, 6.1.0. The tutorial uses the new internal Expr API, but 6.1.0 didn't include this. So I have to do something like this:

```python
import ibis
from ibis.expr.operations import Value
import ibis.expr.rules as rlz
from ibis.expr.types import ArrayValue, NumericValue
import sqlalchemy as sa

try:
    # For newer versions of ibis
    import ibis.expr.datashape as ds

    class ArraySum(Value):
        arg: Value[dt.Array, ds.Any]

        dtype = dt.float32
        shape = rlz.shape_like("arg")

except ImportError:
    # For older versions of ibis

    class ArraySum(Value):
        arg = rlz.array

        output_dtype = dt.int64
        output_shape = rlz.shape_like("arg")


def array_sum(array_value: ArrayValue) -> NumericValue:
    return ArraySum(array_value).to_expr()


@ibis.duckdb.add_operation(ArraySum)
def _array_sum(translator, expr):
    # pull out the arguments to the expression
    (arg,) = expr.args

    # compile the argument
    compiled_arg = translator.translate(arg)

    # return a SQLAlchemy expression that calls into the SQLite list_sum function
    return sa.func.list_sum(compiled_arg)


ArrayValue.sum = array_sum
```

jcrist:
Since we [no longer consider operations to be part of our public API](https://github.com/ibis-project/ibis/pull/6740/files#diff-c1cc304e56dd03cbb5cf65c32d83c9b7ab069800b2ce780249ac87d320e76789) (or at least don't consider any changes between versions to be user-facing breaking changes), I wonder if we should drop this how-to completely?

*(side-note - I think the versioning.md page got dropped in the quarto refactor? Perhaps we should add it back (under reference??))*

NickCrews:
> versioning.md

I would personally put this under concepts, as I would categorize it as [understanding-oriented](https://www.writethedocs.org/videos/eu/2017/the-four-kinds-of-documentation-and-why-you-need-to-understand-what-they-are-daniele-procida/)

> part of the public API

Ah, I was assuming this page being missing from the docs was accidental. Extending Ibis like this is pretty essential for me to workaround missing features eg ArrayValue.sum(). I would appreciate it if this guide remained, and I'm fine with it explicitly saying the API is unstable. I can write compat code myself and I think that is OK to burden users with that responsibility.

cpcloud (approved):

PR #7079, opened 2023-09-07
docs: update link to sqlalchemy tutorial

fmichonneau:
The previous link goes to a page that says the resource has moved, so I updated the link to that resource.


cpcloud (approved):
Thank you!

PR #7083, opened 2023-09-07
refactor(api): remove string as a parent type from expression API

cpcloud:
Removes String* as a parent class of other types. This is technically a breaking change, ~though we do not have any coverage of these that I could see.~ We have some coverage of the data type changes by way of literal normalization.

### DIFF
--- a/ibis/backends/base/sql/registry/literal.py
+++ b/ibis/backends/base/sql/registry/literal.py
@@ -96,7 +96,7 @@ def literal(translator, op):
 
     if dtype.is_boolean():
         typeclass = "boolean"
-    elif dtype.is_string():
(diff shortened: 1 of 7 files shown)


kszucs:
What's the rationale behind this change? I agree that the current datatype hierarchy isn't entirely correct. How about something similar to arrow's extension dtypes instead?

cpcloud:
These aren't strings, and so they shouldn't be modeled as such. I don't think there's an obvious logical base class here, and we shouldn't force one.

cpcloud:
The same argument exists for UUID. All of these types have string representations, but I don't think it makes sense to equate that with be a string value.

jcrist (approved):
:shipit:

Issue #7060, opened 2023-09-06
bug: Ibis doesn't work with pySpark 3.2


contang0:
### What happened?

This probably doesn't strictly qualify as a bug, but Ibis doesn't seem to work with pySpark version 3.2, and I don't see a reason why. I checked Ibis dependencies and I see that Ibis assumes version >3, but nothing more specific.

My problem is that at work our Cloudera cluster runs Spark version 3.2, and pySpark version must match it. When I run `pip install ibis-framework[pyspark]` Ibis installs pySpark version 3.4.

```
import ibis
from pyspark.sql import SparkSession
session = SparkSession.builder.getOrCreate()
```

This throws Error 1 (below).

If I manually downgrade pySpark to 3.2.0 I get Error 2, i.e. a Spark session can now be created, but Ibis complains about some missing attribute.

### What version of ibis are you using?

6.2.0

### What backend(s) are you using, if any?

pyspark

### Relevant log output

Error 1:

```python
Py4JError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 session = SparkSession.builder.getOrCreate()

File ~/.local/lib/python3.10/site-packages/pyspark/sql/session.py:480, in SparkSession.Builder.getOrCreate(self)
    477     sc = SparkContext.getOrCreate(sparkConf)
    478     # Do not update `SparkConf` for existing `SparkContext`, as it's shared
    479     # by all sessions.
--> 480     session = SparkSession(sc, options=self._options)
    481 else:
    482     getattr(
    483         getattr(session._jvm, "SparkSession$"), "MODULE$"
    484     ).applyModifiableSettings(session._jsparkSession, self._options)

File ~/.local/lib/python3.10/site-packages/pyspark/sql/session.py:525, in SparkSession.__init__(self, sparkContext, jsparkSession, options)
    521         getattr(getattr(self._jvm, "SparkSession$"), "MODULE$").applyModifiableSettings(
    522             jsparkSession, options
    523         )
    524     else:
--> 525         jsparkSession = self._jvm.SparkSession(self._jsc.sc(), options)
    526 else:
    527     getattr(getattr(self._jvm, "SparkSession$"), "MODULE$").applyModifiableSettings(
    528         jsparkSession, options
    529     )

File ~/.local/lib/python3.10/site-packages/py4j/java_gateway.py:1587, in JavaClass.__call__(self, *args)
   1581 command = proto.CONSTRUCTOR_COMMAND_NAME +\
[...]

PR #7085, opened 2023-09-07
fix(pyspark): gate other usage of DayTimeIntervalType for PySpark 3.2

gforsyth:
Finish fixing #7060 

I missed the additional usage of `DayTimeIntervalType` in the `pyspark.datatypes` module but that's also gated now

```python
[ins] In [1]: import pyspark

[ins] In [2]: pyspark.__version__
Out[2]: '3.2.0'

[ins] In [3]: from ibis.backends.pyspark.datatypes import type_to_sql_string

[ins] In [4]: 
```

We should probably be testing on PySpark 3.2 if we want to maintain it as a supported version.

### DIFF
--- a/ibis/backends/pyspark/datatypes.py
+++ b/ibis/backends/pyspark/datatypes.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import pyspark
 import pyspark.sql.types as pt
+from packaging.version import parse as vparse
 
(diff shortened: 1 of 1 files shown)


jcrist (approved):
:shipit:

Agreed that we might want to add tests to cover this to avoid future regressions.

PR #7086, opened 2023-09-07
ci: do not quote joined arguments to `pip install`

deepyaman:
If you do, you can get things like:

```
Run poetry run pip install "apache-flink grpcio-status"
ERROR: Invalid requirement: 'apache-flink grpcio-status'
```


jcrist (approved):
Makes sense to me, nice find

cpcloud (approved):

PR #7087, opened 2023-09-07
docs: remove streamlit app on front page

lostmygithubaccount:
meant to do this pre-merge -- Streamlit doesn't really fit into the category of visualization of the rest; we have a how-to elsewhere; this app is sluggish and from my personal accounts (and goes down frequently) so we shouldn't have it on the front page

edit: oh and the post it links to is hidden and needs a full rewrite


jcrist (approved):
Makes sense to me

PR #7100, opened 2023-09-08
fix(api): support passing `functools.partial` to array `.map`/`.filter` methods

### LINKED ISSUE #7095: bug: can't use functools.partial with Array.map()
### What happened?

small example:

```python
import ibis
from functools import partial

a = ibis.array([1,2,3])

def add_n(x, n):
    return x + n

add_2 = partial(add_n, n=2)
a.map(add_2)
```

causes `IbisTypeError: Callable has mandatory keyword-only arguments which cannot be specified`

What I'm actually trying to do is have a general function, but then specialize it for different values and have a different column for each one. eg:

```python
t2 = t.group_by("species").agg(
    years=_.year.collect()
)
m = {f"year_plus_{i}": _.years.map(partial(add_n, n=i)) for i in range(3)}
t2.mutate(**m)
```

I tried to do this:
```python
t2 = t.group_by("species").agg(
    years=_.year.collect()
)
m = {}
for i in range(3):
    def add_i(x):
        return x + i
    m[f"year_plus_{i}"] = _.years.map(add_i)
t2.mutate(**m)
┏━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ species   ┃ years                  ┃ year_plus_0            ┃ year_plus_1            ┃ year_plus_2            ┃
┡━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━┩
│ string    │ array<int64>           │ array<int64>           │ array<int64>           │ array<int64>           │
├───────────┼────────────────────────┼────────────────────────┼────────────────────────┼────────────────────────┤
│ Adelie    │ [2007, 2007, ... +150] │ [2009, 2009, ... +150] │ [2009, 2009, ... +150] │ [2009, 2009, ... +150] │
│ Gentoo    │ [2007, 2007, ... +122] │ [2009, 2009, ... +122] │ [2009, 2009, ... +122] │ [2009, 2009, ... +122] │
│ Chinstrap │ [2007, 2007, ... +66]  │ [2009, 2009, ... +66]  │ [2009, 2009, ... +66]  │ [2009, 2009, ... +66]  │
└───────────┴────────────────────────┴────────────────────────┴────────────────────────┴────────────────────────┘
```
but notice that the `i` is only using the last seen value of 2, which is to be expected with the semantics of how python doesn't bind that value at definition time. I can get it to work with a class:

```python
t2 = t.group_by("species").agg(
    years=_.year.collect()
)

class AddN:
    def __init__(self, n):
        self.n = n

    def __call__(self, x):
        return x + self.n

m = {}
for i in range(3):
    m[f"year_plus_{i}"] = _.years.map(AddN(i))
t2.mutate(**m)
```

but functools.partial is the canonical way to do this and I think it should be supported. IDK if there are other pathways that we should consider here.

6.1.0

NA

### Relevant log output

_No response_

### Code of Conduct

jcrist:
Fixes #7095.

### DIFF
--- a/ibis/backends/tests/test_array.py
+++ b/ibis/backends/tests/test_array.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+import functools
 import os
 
 import numpy as np
@@ -527,9 +528,14 @@ def test_array_slice(backend, start, stop):
 )
 def test_array_map(backend, con, input, output):
     t = ibis.memtable(input, schema=ibis.schema(dict(a="!array<int8>")))
+    expected = pd.DataFrame(output)
+
     expr = t.select(a=t.a.map(lambda x: x + 1))
     result = con.execute(expr)
-    expected = pd.DataFrame(output)
+    backend.assert_frame_equal(result, expected)
+
+    expr = t.select(a=t.a.map(functools.partial(lambda x, y: x + y, y=1)))
+    result = con.execute(expr)
     backend.assert_frame_equal(result, expected)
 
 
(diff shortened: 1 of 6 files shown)


NickCrews:
Looks great! Thanks so much!

cpcloud (approved):
Sweet!

kszucs:
Thanks for fixing it!

PR #7107, opened 2023-09-09
chore(deps): remove `pins` nix override and tell direnv to watch some additional files

cpcloud:


cpcloud:
Nix CI jobs are passing. Merging!

Issue #6549, opened 2023-06-28
feat: squash merges on PRs


lostmygithubaccount:
### Is your feature request related to a problem?

After some internal discussion, opening this issue for tracking. There were recently 2 issues with PRs due to the requirement to squash commits into a single commit, which can be a more complicated git operation (edit: adding a third more recent example, certainly others): 

1. https://github.com/ibis-project/ibis/pull/6543
2. https://github.com/ibis-project/ibis/pull/6460#issuecomment-1594987395
3. https://github.com/ibis-project/ibis/pull/7804

It would be easier to accept PRs if this requirement were removed. We should enable PR squash merge on GitHub. This likely involves some work to ensure the changelog and everything else works as expected.

### Describe the solution you'd like

As a user, I can simply push commits to a PR and have it merged without a rebase or other more advanced git operations.

### What version of ibis are you running?

n/a

### What backend(s) are you using, if any?

n/a

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gforsyth:
> It would be easier to accept PRs if this requirement were removed. We should enable PR squash merge on GitHub. This likely involves some work to ensure the changelog and everything else works as expected.

On some of the substrait repos (like substrait-python), we have a github action that checks that the PR title conforms to the conventional commits style -- since the PR title is the default commit message for a squashed PR, this (mostly) takes care of ensuring that the squash merge messages conform:

https://github.com/substrait-io/substrait-python/blob/main/.github/workflows/pr_title.yml

gforsyth:
I think for the core devs, we still mostly squash and rebase our own PRs, but I agree this can be helpful, especially for new contributors.

gforsyth:
(bonus, if the PR title lint fails and you edit the PR title, it doesn't require re-running the entire CI suite)

cpcloud:
Seems legit, let's do it!

gforsyth:
Ooooh, question: should we disable `commitlint`?
Or allow failures?

cpcloud:
I think we can disable it in PRs, and continue to run it on a merge to `master`.

cpcloud:
Which appears to be basically the opposite of what we're doing now!

cpcloud:
It'll be annoying to specify the commit range though because there's no obvious reference point once the commit is merged. Perhaps just disable it for now.

cpcloud:
Also there's not much value in checking after a merge to `master` because we disabled force pushing :)

PR #7116, opened 2023-09-09
fix(impala): allow arbitrary kwargs when connecting

contang0:
Hello,

At the moment it is not possible to pass arbitrary keyword arguments to Impala backend, which makes it impossible to connect if, for example, Impala is configured to use HTTP transport. It should be possible to supply any arguments required by Impyla. I tested this on Impala (with S3) and both querying and table creation worked fine.

I also added one more example of how a connection can be created, which I think is more realistic to what users encounter in practice, especially with CDP. As hdfs_client is optional and not relevant to S3 setups I purposely leave it out from this example.

### DIFF
--- a/.github/workflows/ibis-backends-cloud.yml
+++ b/.github/workflows/ibis-backends-cloud.yml
@@ -5,9 +5,12 @@ on:
     # Skip the backend suite if all changes are in the docs directory
     paths-ignore:
       - "docs/**"
+      - "**/*.md"
+      - "**/*.qmd"
+      - "codecov.yml"
+      - ".envrc"
     branches:
       - master
-      - quarto
 
 permissions:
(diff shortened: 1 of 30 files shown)


contang0:
sorry I've no idea how to fix the commitlint errors. will close this and open another PR.

lostmygithubaccount:
@contang0 we appreciate the contributions! you have to rebase into a single commit that follows a certain format (see https://ibis-project.org/contribute/02_workflow#writing-the-commit), happy to provide guidance on that if needed

opening a new PR with a single commit should also fix it. we have an open issue to remove this requirement and make contributions easier! https://github.com/ibis-project/ibis/issues/6549

contang0:
Thank you. I was having some issues with rebase. I opened a new PR now.

PR #7117, opened 2023-09-09
fix(impala): allow arbitrary connection params

contang0:
Hello,

At the moment it is not possible to pass arbitrary keyword arguments to Impala backend, which makes it impossible to connect if, for example, Impala is configured to use HTTP transport. It should be possible to supply any arguments required by Impyla. I tested this on Impala (with S3) and both querying and table creation worked fine.

I also added one more example of how a connection can be created, which I think is more realistic to what users encounter in practice, especially with CDP. As hdfs_client is optional and not relevant to S3 setups I purposely leave it out from this example.

p.s. I closed [!7116](https://github.com/ibis-project/ibis/pull/7116) because I was having trouble rebasing my commits, hopefully commitlint will not complain anymore.


cpcloud (approved):
LGTM, thanks!

PR #7119, opened 2023-09-10
feat(api): expose the ability to refer to existing functions not defined in ibis APIs

cpcloud:
This PR exposes our already-existing opaque UDFs (renamed to `builtin` here), and adds tests for the API. ~Draft mode until I add tests for more backends.~


ibis/expr/operations/udf.py
```diff
@@ -149,15 +222,73 @@ def pyarrow(fn: Callable | None = None, *args: Any, **kwargs: Any) -> Callable:
...
+        """Construct a scalar user-defined function that is built-in to the backend.
+
+        Parameters
+        ----------
+        fn
+            The The function to wrap.
+        args
+            Configuration arguments for the UDF.
+        target
+            The name of the UDF in the backend if different from the function name.
```
jcrist, line 253: Thoughts on renaming this to `name`? `target` doesn't feel intuitive to me.

docs/how-to/extending/builtin.qmd
```diff
@@ -0,0 +1,134 @@
...
+```
+
+The [`...`](https://docs.python.org/3/library/constants.html#Ellipsis) is
+a visual indicator that the function definition is unknown to Ibis.
+
+::: {.callout-note collapse="true"}
+## Ibis does not do anything with the function body.
+
+Ibis will not inspect the function body or otherwise inspect it. Any code you
+write in the function body **will be ignored**.
```
jcrist, line 40: Repeat of `inspect` here feels like a typo. Not 100% sure what was intended here, but guessing something like

```suggestion
Ibis will not execute the function body or otherwise inspect it. Any code you
write in the function body **will be ignored**.
```

docs/how-to/extending/builtin.qmd
```diff
@@ -0,0 +1,134 @@
...
+
+And let's take a look at the SQL
+
+```{python}
+ibis.to_sql(expr, dialect="snowflake")
+```
+
+## `TODO`
+
+- Allow function overloading
```
jcrist, line 134: Is this something you want to get in before merging this PR? Or a TODO for later?

jcrist (commented):
Excited to see this feature. Just a quick surface-level review for now.

docs/how-to/extending/builtin.qmd
cpcloud: It's a TODO for later.

ibis/expr/operations/udf.py
cpcloud: I used `target` because I am trying to surface the idea that this value will be the name the function compiles to. `name` seems more intuitive, but IMO too vague to convey what I am going for.

cpcloud: How about `compiles_to`?

docs/how-to/extending/builtin.qmd
jcrist: I'd personally leave it out of the docs then (IMO TODOs don't belong in docs).

cpcloud: Yep, SGMT.

ibis/expr/operations/udf.py
cpcloud: Ok, renamed to `compiles_to`.

docs/how-to/extending/builtin.qmd
cpcloud: Removed!

ibis/expr/operations/udf.py
cpcloud: Ok, I ended up going with your suggestion @jcrist. It's a lot less vague-looking than I thought it would be1

ibis/backends/clickhouse/tests/test_client.py
```diff
@@ -244,3 +244,21 @@ def test_truncate_table(con, engine, temp_table):
     assert len(t.execute()) == 1
     con.truncate_table(temp_table)
     assert t.execute().empty
+
+
+@udf.scalar.builtin(name="arrayJaccardIndex")
+def array_jaccard_index(a: dt.Array[dt.int64], b: dt.Array[dt.int64]) -> float:
```
kszucs, line 250: Datatype types should be the preferred spelling: `dt.Array[dt.Int64]`

docs/how-to/extending/builtin.qmd
```diff
@@ -0,0 +1,130 @@
...
+
+Ibis doesn't directly expose many of the DuckDB [text similarity
+functions](https://duckdb.org/docs/sql/functions/char.html#text-similarity-functions).
+Let's expose the `mismatches` API.
+
+```{python}
+from ibis import udf
+
+@udf.scalar.builtin
+def mismatches(left: str, right: str) -> int:
```
kszucs, line 29: This looks pretty nice!

kszucs: Do we need a note about annotations being optional?

kszucs (approved):
LGTM! This will be pretty useful!

NickCrews:
This is super useful! Nice walkthrough writeup, that really is going to mean this gets used more. Thanks!

PR #7121, opened 2023-09-11
docs(IO): add missing word, add line breaks

gforsyth:
Removing some usage of the word `simple` or `simply` where it doesn't
add anything to the prose.

My favorite (now removed) example:

```
... simply pass `auth_mechanism='GSSAPI'` or
`auth_mechanism='LDAP'` (and set `kerberos_service_name` if necessary along
with `user` and `password` if necessary) to the
`ibis.impala_connect(...)` method when instantiating an
`ImpalaConnection`.
```


cpcloud (approved):

cpcloud:
We could change some of these to not freeze their outputs. Some of the blog posts are not too annoying to render locally.

PR #7122, opened 2023-09-11
feat(udf): support inputs without type annotations

cpcloud:
Adds support for eliding inputs types to UDFs.

Depends on #7119.


ibis/expr/operations/udf.py
```diff
@@ -298,10 +298,11 @@ def _make_node(
         func_name = name or fn.__name__
 
         for arg_name, param in inspect.signature(fn).parameters.items():
-            if (raw_dtype := annotations.get(arg_name)) is None:
-                raise exc.MissingParameterAnnotationError(fn, arg_name)
-
-            arg = rlz.ValueOf(dt.dtype(raw_dtype))
+            if (raw_dtype := annotations.get(arg_name)) is not None:
```
kszucs, line 301: `pattern = rlz.ValueOf(annotations.get(arg_name))` should be sufficient.

cpcloud: Ah, good call!

cpcloud: Fixed!

kszucs: Did you push it, I still see the previous diff.

kszucs (approved):
After rebasing and pushing the missing change it is good to go.

ibis/expr/operations/udf.py
cpcloud: Oh, lemme push this up.

cpcloud: Ah, it's pushed now!

PR #7127, opened 2023-09-12
feat: introduce watermarks in ibis api

chloeh13q:
Introduce watermarks in Ibis API and implements `CREATE TABLE` DDL with optional watermark strategy definition in Flink backend.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -22,6 +22,7 @@ from ibis.expr.decompile import decompile
 from ibis.expr.deferred import Deferred
 from ibis.expr.schema import Schema
 from ibis.expr.sql import parse_sql, show_sql, to_sql
+from ibis.expr.streaming import Watermark
 from ibis.expr.types import (
     DateValue,
     Expr,
@@ -157,6 +158,7 @@ __all__ = (
     "trailing_range_window",
     "trailing_window",
     "union",
+    "watermark",
     "where",
     "window",
     "preceding",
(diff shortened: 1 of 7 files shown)


ibis/expr/streaming.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
+
+import ibis.expr.types as ir  # noqa: TCH001
+from ibis.common.grounds import Concrete
+
+
+class Watermark(Concrete):
+    time_col: str  # TODO(chloeh13q): allow column references
```
jcrist, line 8: Since these will always be used as part of a `create_table` call, I don't see how the user would ever have access to a direct `Column` here. Supporting only a string name seems fine to me. I'd drop this TODO entirely.

ibis/expr/api.py
```diff
@@ -1544,6 +1546,24 @@ def difference(table: ir.Table, *rest: ir.Table, distinct: bool = True):
...
 
+def watermark(time_col: str, allowed_delay: ir.IntervalScalar) -> Watermark:
+    """Return a watermark object.
+
+    Parameters
+    ----------
+    time_col:
+        The timestamp column that will be used to generate watermarks in event time processing.
+    allowed_delay:
+        Length of time that events are allowed to be late.
```
jcrist, line 1557: Docstring formatting nit: no need for `:` here

```suggestion
    time_col
        The timestamp column that will be used to generate watermarks in event time processing.
    allowed_delay
        Length of time that events are allowed to be late.
```

jcrist (commented):
Overall this LGTM! Just a few small nits.

chloeh13q:
@jcrist Thanks! Addressed the comments

jcrist (approved):
:100:

PR #7130, opened 2023-09-12
fix(duckdb): remove hack to workaround bug that was fixed upstream

### LINKED ISSUE #7103: bug(duckdb): percent signs aren't escaped correctly in `.case()`
### What happened?

EDIT: I originally filed this as a repr bug, but it actually is a bug with .case()

```python
cased = ibis.case().when(t.sex == "male", r"%").else_("").end()
added = r"%" + _.sex
t.select("sex", added=added, cased=cased)
```

results in (note the double %%, should be a single %)
```
┏━━━━━━━━┳━━━━━━━━━┳━━━━━━━━┓
┃ sex    ┃ added   ┃ cased  ┃
┡━━━━━━━━╇━━━━━━━━━╇━━━━━━━━┩
│ string │ string  │ string │
├────────┼─────────┼────────┤
│ male   │ %male   │ %%     │
│ female │ %female │ ~      │
│ female │ %female │ ~      │
│ NULL   │ %       │ ~      │
│ female │ %female │ ~      │
│ male   │ %male   │ %%     │
│ female │ %female │ ~      │
│ male   │ %male   │ %%     │
│ NULL   │ %       │ ~      │
│ NULL   │ %       │ ~      │
│ …      │ …       │ …      │
└────────┴─────────┴────────┘
```

The problem is that this is getting compiled to:

```sql
SELECT
  t0.sex,
  CONCAT(COALESCE(CAST('%' AS TEXT), ''), COALESCE(CAST(t0.sex AS TEXT), '')) AS added,
  CASE WHEN (
    t0.sex = 'male'
  ) THEN '%%' ELSE '' END AS cased
FROM _ibis_examples_penguins_3bvfvrlqvvfxtohz4nsbvfdxca AS t0
```

Using .ifelse() works fine.

6.1.0

duckdb

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Address doubly escaped percent signs in the duckdb backend, by bumping the lower bound of duckdb. Fixes #7103.

Apparently druid was broken here too, fixed that as well.

### DIFF
--- a/ibis/backends/druid/__init__.py
+++ b/ibis/backends/druid/__init__.py
@@ -62,6 +62,9 @@ class Backend(BaseAlchemyBackend):
         # workaround a broken pydruid `has_table` implementation
         engine.dialect.has_table = self._has_table
 
+        # don't double percent signs
+        engine.dialect.identifier_preparer._double_percents = False
+
     @staticmethod
     def _new_sa_metadata():
         meta = sa.MetaData()
--- a/ibis/backends/druid/datatypes.py
+++ b/ibis/backends/druid/datatypes.py
@@ -54,6 +54,11 @@ def _smallint(element, compiler, **kw):
     return "SMALLINT"
 
 
+@compiles(DruidString, "druid")
+def _string(element, compiler, **kw):
+    return "VARCHAR"
+
+
 class DruidType(AlchemyType):
     dialect = "hive"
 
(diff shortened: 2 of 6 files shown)


jcrist (approved):

NickCrews:
Looks good, thanks @cpcloud ! 

In the 7.0 release notes I would call this a bug fix, as if I have ibis<7.0 and duckdb>0.7.0 then I experience the double percent sign bug. Just to warn anyone in that place (which is probably most people)

I would almost say relax the duckdb bounds to be >0.7.0, but I think I agree it's not worth it for the two conditionals (one for the percent sign, one for maps) that would be required.

PR #7134, opened 2023-09-12
feat(udf): add support for builtin aggregate UDFs

cpcloud:
Adds support for builtin user-defined aggregate functions. This came up as a use case during a blog post I am authoring on the expanded array functionality for BigQuery that is landing in ibis 7.0.0. The blog post no longer uses this feature, but it's a useful one regardless.

Depends on #7122.


ibis/expr/operations/udf.py
```diff
@@ -51,21 +57,130 @@ def _wrap(
...
+S = TypeVar("S", bound=ops.Value)
+B = TypeVar("B", bound=ops.Value)
+
+
+class _UDF(abc.ABC):
+    __slots__ = ()
+
+    @property
+    @abc.abstractmethod
+    def base(self) -> type[B]:
```
jcrist, line 69: Perhaps make `.base` a private property (so it's not available as `ibis.udf.scalar.base`).

ibis/expr/operations/udf.py
```diff
@@ -51,21 +57,130 @@ def _wrap(
...
+    @classmethod
+    def builtin(
+        cls,
+        fn: Callable | None = None,
+        *args: Any,
+        name: str | None = None,
+        schema: str | None = None,
+        **kwargs: Any,
+    ) -> Callable:
+        """Construct a scalar user-defined function that is built-in to the backend.
```
jcrist, line 82: This docstring will show up for both `ibis.udf.scalar.builtin` and `ibis.udf.agg.builtin`, but says "scalar" above (and shows a scalar example below). Perhaps this belongs on the subclasses rather than the base class? The implementation is short enough that repeating the code feels fine to me.

docs/how-to/extending/builtin.qmd
```diff
@@ -169,3 +172,60 @@ con.execute(cardinality("foo"))
 
 Here, Snowflake is informing us that the `ARRAY_SIZE` function does not accept
 strings as input.
+
+
+## Aggregate functions
+
+Let's the use the BigQuery backend to demonstrate how to access a BigQuery-only
```
jcrist, line 179:
```suggestion
Let's use the BigQuery backend to demonstrate how to access a BigQuery-only
```

docs/how-to/extending/builtin.qmd
```diff
@@ -169,3 +172,60 @@ con.execute(cardinality("foo"))
 
 Here, Snowflake is informing us that the `ARRAY_SIZE` function does not accept
 strings as input.
+
+
+## Aggregate functions
```
jcrist, line 177: I think a few sentences of description would be useful to set the scene here before an example. Copy-pasting and slight edits from the scalar docs above:

```
Aggregate functions that aren't exposed in ibis directly can be accessed using the
`@ibis.udf.agg.builtin` decorator.

::: {.callout-tip}
### [Ibis APIs](../../reference/index.qmd) may already exist for your function.

Builtin aggregate UDFs are designed to be an escape hatch when Ibis doesn't have
a defined API for a built-in database function.

See [the reference documentation](../../reference/index.qmd) for existing APIs.
:::
```

docs/how-to/extending/builtin.qmd
```diff
@@ -169,3 +172,60 @@ con.execute(cardinality("foo"))
...
+
+```{python}
+con = ibis.connect("bigquery://ibis-gbq")  # <1>
+
+con.set_database("bigquery-public-data.imdb")
+```
+
+1. Replace `ibis-gbq` with your GCP project name
+
+Now define the builtin aggregate function:
```
jcrist, line 194: A sentence or two about how the user should annotate the function signature would be useful. Perhaps describe what `array_concat_agg` does, and how the signature types below correspond to it?

Questions a user might ask:

- Why annotate only the return type and not the argument?
- Is the return type the aggregate? Or a column of the aggregate?
- What is `dt.Array[str]` (I know it's an array of strings, but users haven't had to interact with ibis datatypes in signatures anywhere else, so this is new to them).

I guess the general feedback here is "some more description of what's happening" would be nice.

jcrist: Alternatively, we could use a simpler backend and example which may reduce the complexity budget here. Perhaps wrapping duckdb's kurtosis function instead (https://duckdb.org/docs/sql/aggregates.html)?

This would also avoid the need for the biquery-specific callouts in the example around the connection string/use of `head`.

jcrist (commented):
Overall this seems pretty good.

One thing I thought about while reviewing this is that maybe grouping functionality by UDF implementation (builtin, pyarrow, ...) rather than kind (scalar, agg, ...) would make more sense? So it'd be `udf.builtin.scalar` instead of `udf.scalar.builtin`?

Most/all backends should be able to support `udf.builtin` for exposing builtin functionality, but the other UDF types would be rarer to support. If a user wants to expose a builtin function directing them to `udf.builtin.*` rather than `udf.*.builtin` might make things clearer since that functionality is all grouped together?

No need to deal with this now/here if we want to make this change, but it came up while reviewing it so I thought I'd mention it.

docs/how-to/extending/builtin.qmd
cpcloud: Good points, I will adjust this to use DuckDB.

ibis/expr/operations/udf.py
cpcloud: Done, it's now `_base`

ibis/expr/operations/udf.py
cpcloud: Repeated the docstring and made it `agg`-worthy

docs/how-to/extending/builtin.qmd
cpcloud: No longer relevant because I switched this section to use DuckDB

docs/how-to/extending/builtin.qmd
cpcloud: Put the above text in verbatim, thanks!

docs/how-to/extending/builtin.qmd
cpcloud: Adjusted to use DuckDB and kurtosis, thanks for the feedback!

cpcloud:
> One thing I thought about while reviewing this is that maybe grouping functionality by UDF implementation (builtin, pyarrow, ...) rather than kind (scalar, agg, ...) would make more sense? So it'd be udf.builtin.scalar instead of udf.scalar.builtin?

Yeah ...

I went back and forth on this a bit when building the API.

I don't have a super strong opinion, but my thinking is based on how people talk about the separate kinds of UDFs in other systems.

The first point of departure in my experience has always been the shape of the output: scalar, aggregate, or tabular, and so that was the first level of namespacing I chose.

I **think** but am not entirely sure, that this makes implementation a bit cleaner because there's more overlap in behavior in the scalar category than the builtin category. That's not really based on thorough analysis of anything though, just my intuition.

I think when I am reading UDF signatures the first thing I want to know is where I can use it, and `scalar`/`agg`/`tabular` gives more information in that dimension than whether the thing is builtin or not.

jcrist:
> I think when I am reading UDF signatures the first thing I want to know is where I can use it, and scalar/agg/tabular is a stronger determinant of where I can call it than whether the thing is builtin or not.

That's a good point - I think for reading existing udfs scalar/agg/tabular is more critical, but for writing whether it's builtin/pyarrow/pandas feels more critical. I also don't have strong thoughts on this, what we have now seems fine, we can always reorg later if it a different layout becomes clearly better.

docs/how-to/extending/builtin.qmd
```diff
@@ -169,3 +172,98 @@ con.execute(cardinality("foo"))
...
+   .relocate("kurt", after="tconst")
+   .filter(
+      [
+         _.original_title.lower().contains("godfather"),
+         _.title_type == "movie",
+         _.genres.contains("Crime") & _.genres.contains("Drama"),
+      ]
+   )
+)
+expr
```
jcrist, line 268: Nice example

jcrist (approved):
Excellent, excited to use this new functionality

PR #7140, opened 2023-09-12
docs: add interactivity to altair example

NickCrews:
We weren't showing off all the features of altair and giving it justice.


lostmygithubaccount:
could you update the example on the home page (`docs/index.qmd`) as well?

cpcloud (approved):

PR #7141, opened 2023-09-12
docs: re-organize reference sections

NickCrews:
Just some nits:

- Only have expressions in the Expressions section
- Only 3 main sections: Expressions, Other, and Configuration
- Sort expressions by usefulness/prevalence: table, generic, numeric/string
- Sort the "Other API" section by my opinion of usefulness/prevalence, but not as sure here.
- Rename Complex to Collections
- Improve descriptions for some pages

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -194,7 +194,7 @@ quartodoc:
           path: top_level
           summary:
             name: Top-level APIs
-            desc: These methods and objects are available directly on the `ibis` module.
+            desc: Methods and objects available directly on the `ibis` module.
           contents:
             - name: and_
               dynamic: true
@@ -294,7 +294,7 @@ quartodoc:
           package: ibis.expr.types.relations
           summary:
             name: Table expressions
-            desc: Table expressions form the basis for most Ibis expressions.
+            desc: Tables are one of the core data structures in Ibis.
           contents:
             - Table
             - GroupedTable
@@ -304,7 +304,7 @@ quartodoc:
           package: ibis.expr.types.generic
           summary:
             name: Generic expressions
-            desc: These expressions are available on scalars and columns of any element type.
+            desc: Scalars and columns of any element type.
           contents:
             - Value
             - Column
@@ -314,8 +314,8 @@ quartodoc:
           path: expression-numeric
           package: ibis.expr.types.numeric
           summary:
-            name: Numeric and boolean expressions
-            desc: These APIs are available on numeric and boolean expressions.
+            name: Numeric and Boolean expressions
+            desc: Integer, floating point, decimal, and boolean expressions.
           contents:
             - NumericValue
             - NumericColumn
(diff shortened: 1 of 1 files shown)


cpcloud:
@NickCrews Thanks for the PR!

Would it be possible to split some of these changes into separate PRs?

> * Only have expressions in the Expressions section

What APIs are in that section that aren't expressions?

> * Only 3 main sections: Expressions, Other, and Configuration

I'm not a huge fan of this categorization. I think sections should have names so that it's clear what the pages in that section are about. Otherwise a user is forced to use the search bar.

> * Sort expressions by usefulness/prevalence: table, generic, numeric/string

Yeah, this makes sense to me.

> * Sort the "Other API" section by my opinion of usefulness/prevalence, but not as sure here.

Personally I am not in favor of a catch-all section :)

> * Rename Complex to Collections

+1

> * Improve descriptions for some pages

This is great, thanks!

NickCrews:
> Would it be possible to split some of these changes into separate PRs?

Will do. I'll do the easy ones first :)

> What APIs are in that section that aren't expressions?

I think I see what you mean. Many of them are related to expressions because ALL of ibis is related to expressions :). But there are a couple that really are not:

- ibis.read_*()
- ibis.connect()
- g/set_backend()
- row_number()

How about this: Currently it is organized by location (ie if you find it under `ibis.` or under `StringValue.`). What if we grouped things more by usecase?

- `ibis.case()`, `ibis.coalesce()`, etc under Generic
- `ibis.union()` et al under Table
- `ibis.or_`, `ibis.and_` under Boolean Values
- everything that doesn't fit, like `ibis.row_number()` or `ibis.connect()` is pulled out the expression section into a Top Level API section
- (maybe?) the window-related stuff is pulled into its own section? Not underneath expressions?

I think this better mirrors a user's search path. They are trying to get something done when they visit the reference section, eg "get the first non-null value", or "if any of N booleans are true, I want true". If it were me, I would start looking under Generic Values for the first question, and Boolean values for the second. People don't look for an answer by thinking "oh, this is probably going to be under the `ibis` top-level namespace". They search by datatype/theme/category, not physical namespace.

Or often I am looking for the API for `ibis.read_csv()` or some other global function. The expression-related stuff, like `ibis.or_()` is just clutter as I browse the list.

I can see the argument for keeping Selectors underneath the expressions API. Maybe this should just get put under the Table section? That is the only time when you would use them.

> I'm not a huge fan of this categorization. I think sections should have names so that it's clear what the pages in that section are about. Otherwise a user is forced to use the search bar.
> Personally I am not in favor of a catch-all section :)

OK, I can ditch this catchall "Other"

cpcloud:
> How about this: Currently it is organized by location (ie if you find it under ibis. or under StringValue.). What if we grouped things more by usecase?

💯 I like this idea.

NickCrews:
How about we just merge those easy tweaks and then I'll do a new PR, so there is a cleaner diff that is more obvious what is happening? I can also add more commits and do everything at once if you think that's better.

cpcloud (approved):
LGTM, thanks!

PR #7148, opened 2023-09-13
perf(ops): allow larger table sets in expressions

### LINKED ISSUE #7124: bug: maximum recursion depth with join operation
### What happened?

We're seeing a `RecursionError: maximum recursion depth exceeded while calling a Python object` when running a JOIN: `source_difference = source.join(differences, join_keys, how="outer")`
Both 'source' and 'differences' are pandas.Table()s with many columns (~120).

We don't hit this error with smaller, less wide tables. I've provided a abridged version of the stack trace below - it does look like there is a cyclical portion of the code when testing if left and right tables have a common parent expr [here](https://github.com/ibis-project/ibis/blob/5.1.0/ibis/expr/operations/relations.py#L178).

Trying to understand if this is a Python limitation due to how wide the table is, or an Ibis bug. Appreciate the help! 

5.1.0

Pandas

### Relevant log output

```sh
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 210, in __cached_equals__
    result = self.__cache__[key]
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/caching.py", line 46, in __getitem__
    value, _ = self._data[identifiers]
KeyError: (139856786236240, 139856788868000)

During handling of the above exception, another exception occurred:
...
During handling of the above exception, another exception occurred:

  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 210, in __cached_equals__
    result = self.__cache__[key]
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/caching.py", line 46, in __getitem__
    value, _ = self._data[identifiers]
KeyError: (139856787188608, 139856789577136)

During handling of the above exception, another exception occurred:

  File "/x/home/user/new_fl/dvt4/bin/data-validation", line 11, in <module>
    load_entry_point('google-pso-data-validator==4.1.0', 'console_scripts', 'data-validation')()
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/data_validation/__main__.py", line 581, in main
    run_validation_configs(args)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/data_validation/__main__.py", line 551, in run_validation_configs
    config_runner(args)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/data_validation/__main__.py", line 304, in config_runner
    run_validations(args, config_managers)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/data_validation/__main__.py", line 478, in run_validations
    run_validation(
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/data_validation/__main__.py", line 461, in run_validation
    validator.execute()
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/data_validation/data_validation.py", line 96, in execute
    result_df = self._execute_validation(
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/data_validation/data_validation.py", line 314, in _execute_validation
    result_df = combiner.generate_report(
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/data_validation/combiner.py", line 83, in generate_report
    joined = _join_pivots(source_pivot, target_pivot, differences_pivot, join_on_fields)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/data_validation/combiner.py", line 317, in _join_pivots
    source_difference = source.join(differences, join_keys, how="outer")[
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/expr/types/relations.py", line 2497, in join
    expr = klass(left, right, predicates).to_expr()
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 25, in __call__
    return cls.__create__(*args, **kwargs)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 99, in __create__
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 33, in __create__
    return type.__call__(cls, *args, **kwargs)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/expr/operations/relations.py", line 178, in __init__
    if left.equals(right):
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/expr/operations/core.py", line 24, in equals
    return self.__cached_equals__(other)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 212, in __cached_equals__
    result = self.__equals__(other)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 239, in __equals__
    return self.__args__ == other.__args__
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 187, in __eq__
    return self.__cached_equals__(other)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 212, in __cached_equals__
    result = self.__equals__(other)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 239, in __equals__
    return self.__args__ == other.__args__
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 187, in __eq__
    return self.__cached_equals__(other)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 212, in __cached_equals__
    result = self.__equals__(other)
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 239, in __equals__
    return self.__args__ == other.__args__
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 187, in __eq__
    return self.__cached_equals__(other)
File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/grounds.py", line 210, in __cached_equals__
    result = self.__cache__[key]
  File "/x/home/user/new_fl/dvt4/lib/python3.8/site-packages/ibis/common/caching.py", line 45, in __getitem__
    identifiers = tuple(id(item) for item in key)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

### Code of Conduct

cpcloud:
This PR is ~a draft of~ a fix for #7124.

There are at least four issues, two of which are addressed in this PR.

1. For giant set operations, like a nested union of 400 tables with another
   giant union, the schema was computed dynamically, resulting in a traversal
   of that entire 400 deep set on every access. I've replaced all the schema `@property`
   with `@attribute` where possible (e.g., not joins because `__init__`
   can affect the result and `@attribute` is computed on construction).
   This effectively removes the traversal at the cost of some additional memory
   used to store the `Schema` object on each relation.
2. The second problem is the recursion in `__cached_equals__` and is somewhat
   related to the first.

   Here, the stack builds rapidly due to `__eq__` being called for hash
   collision resolution in the global equality cache. When `schema` was
   accessed for comparison, it was done so on every child table expression,
   resulting in a traversal of N things, then N - 1, then N - 2 and so on until
   bottoming out. By storing attributes instead of computing them we avoid that
   adding to the call stack, but there's still the problem of the
   `__cached_equals__` call stack itself.

   This was a bit trickier to arrive at, but I was thinking about other ways that
   we've replaced some recursion with iteration in the code base and
   topological sort came to mind.

   I've reimplemented `equals` to call `__cached_equals__` in a "bottom up"
   fashion after topologically sorting the nodes of the left and right side
   operations.

   This populates the cache from the starting from the leaves and should avoid
   the recursion entirely: the (N + 1)th's child expressions comparison results
   are already cached, so only the top level fields are compared.
3. The third problem is backend specific and that is that the pandas backend is
   also *highly* recursive and will hit the recursion limit during evaluation.
   There's not much to do there I think except encourage folks to move off it
   ASAP.
4. The fourth problem is unrelated to property access or equality, but perhaps
   is another place where moving some recursive code to iteration is warranted
   and that is set operation compilation.

   Set operations are compiled recursively at the moment.

   This is almost certainly something that will improve with the move to
   sqlglot, so I don't think we should address it until after that is well
   underway.

This is in draft mode to get feedback on the idea and approach.

**EDIT**: The initial approach was about 1000x slower than current `master`.

@kszucs came up with a better idea (see discussion below) and that fixes
the issue and keeps the performance roughly the same if not slightly better.

Closes #7124.

### DIFF
--- a/ibis/expr/operations/relations.py
+++ b/ibis/expr/operations/relations.py
@@ -269,7 +269,7 @@ class Join(Relation):
 
     @property
     def schema(self):
-        # TODO(kszucs): use `return self.lefts.chema | self.right.schema` instead which
+        # TODO(kszucs): use `return self.left.schema | self.right.schema` instead which
         # eliminates unnecessary projection over the join, but currently breaks the
         # pandas backend
         left, right = self.left.schema, self.right.schema
@@ -310,14 +310,14 @@ class AnyLeftJoin(Join):
 
 @public
 class LeftSemiJoin(Join):
-    @property
+    @attribute
     def schema(self):
         return self.left.schema
 
 
 @public
 class LeftAntiJoin(Join):
-    @property
+    @attribute
     def schema(self):
         return self.left.schema
 
@@ -345,7 +345,9 @@ class SetOp(Relation):
     distinct: bool = False
 
     def __init__(self, left, right, **kwargs):
-        if left.schema != right.schema:
+        # convert to dictionary first, to get key-unordered comparison
+        # semantics
+        if dict(left.schema) != dict(right.schema):
             raise com.RelationError("Table schemas must be equal for set operations")
         elif left.schema.names != right.schema.names:
             # rewrite so that both sides have the columns in the same order making it
@@ -354,7 +356,7 @@ class SetOp(Relation):
             right = Selection(right, cols)
         super().__init__(left=left, right=right, **kwargs)
 
-    @property
+    @attribute
     def schema(self):
         return self.left.schema
 
@@ -380,7 +382,7 @@ class Limit(Relation):
     n: UnionType[int, Scalar[dt.Integer], None] = None
     offset: UnionType[int, Scalar[dt.Integer]] = 0
 
-    @property
+    @attribute
     def schema(self):
         return self.table.schema
 
@@ -395,7 +397,7 @@ class SelfReference(Relation):
             return f"{name}_ref"
         return util.gen_name("self_ref")
 
-    @property
+    @attribute
     def schema(self):
         return self.table.schema
 
@@ -489,7 +491,7 @@ class DummyTable(Relation):
     # TODO(kszucs): verify that it has at least one element: Length(at_least=1)
     values: VarTuple[Value[dt.Any, ds.Scalar]]
 
-    @property
+    @attribute
     def schema(self):
         return Schema({op.name: op.dtype for op in self.values})
 
@@ -572,7 +574,7 @@ class Distinct(Relation):
 
     table: Relation
 
-    @property
+    @attribute
     def schema(self):
         return self.table.schema
 
@@ -587,7 +589,7 @@ class FillNa(Relation):
     table: Relation
     replacements: UnionType[Value[dt.Numeric | dt.String], FrozenDict[str, Any]]
 
-    @property
+    @attribute
     def schema(self):
         return self.table.schema
 
@@ -600,7 +602,7 @@ class DropNa(Relation):
     how: Literal["any", "all"]
     subset: Optional[VarTuple[Column[dt.Any]]] = None
 
-    @property
+    @attribute
     def schema(self):
         return self.table.schema
 
@@ -612,7 +614,7 @@ class View(PhysicalTable):
     child: Relation
     name: str
 
-    @property
+    @attribute
     def schema(self):
         return self.child.schema
 
--- a/ibis/tests/benchmarks/test_benchmarks.py
+++ b/ibis/tests/benchmarks/test_benchmarks.py
@@ -15,6 +15,7 @@ from packaging.version import parse as vparse
 
 import ibis
 import ibis.expr.datatypes as dt
+import ibis.expr.operations as ops
 import ibis.expr.types as ir
 from ibis.backends.base import _get_backend_names
 from ibis.backends.pandas.udf import udf
@@ -809,3 +810,93 @@ def test_ibis_duckdb_to_pyarrow(benchmark, sql, ddb) -> None:
 
     expr = con.sql(sql)
     benchmark(expr.to_pyarrow)
+
+
+@pytest.fixture
+def diffs():
+    return ibis.table(
+        {
+            "id": "int64",
+            "validation_name": "string",
+            "difference": "float64",
+            "pct_difference": "float64",
+            "pct_threshold": "float64",
+            "validation_status": "string",
+        },
+        name="diffs",
+    )
+
+
+@pytest.fixture
+def srcs():
+    return ibis.table(
+        {
+            "id": "int64",
+            "validation_name": "string",
+            "validation_type": "string",
+            "aggregation_type": "string",
+            "table_name": "string",
+            "column_name": "string",
+            "primary_keys": "string",
+            "num_random_rows": "string",
+            "agg_value": "float64",
+        },
+        name="srcs",
+    )
+
+
+@pytest.fixture
+def nrels():
+    return 300
+
+
+def make_big_union(t, nrels):
+    return ibis.union(*[t] * nrels)
+
+
+@pytest.fixture
+def src(srcs, nrels):
+    return make_big_union(srcs, nrels)
+
+
+@pytest.fixture
+def diff(diffs, nrels):
+    return make_big_union(diffs, nrels)
+
+
+def test_big_eq_expr(benchmark, src, diff):
+    benchmark(ops.core.Node.equals, src.op(), diff.op())
+
+
+def test_big_join_expr(benchmark, src, diff):
+    benchmark(ir.Table.join, src, diff, ["validation_name"], how="outer")
+
+
+def test_big_join_execute(benchmark, nrels):
+    pytest.importorskip("duckdb")
+    pytest.importorskip("duckdb_engine")
+
+    con = ibis.duckdb.connect()
+
+    # cache to avoid a request-per-union operand
+    src = make_big_union(
+        con.read_csv(
+            "https://github.com/ibis-project/ibis/files/12580336/source_pivot.csv"
+        )
+        .rename(id="column0")
+        .cache(),
+        nrels,
+    )
+
+    diff = make_big_union(
+        con.read_csv(
+            "https://github.com/ibis-project/ibis/files/12580340/differences_pivot.csv"
+        )
+        .rename(id="column0")
+        .cache(),
+        nrels,
+    )
+
+    expr = src.join(diff, ["validation_name"], how="outer")
+    t = benchmark.pedantic(expr.to_pyarrow, rounds=1, iterations=1, warmup_rounds=1)
+    assert len(t)
--- a/ibis/common/grounds.py
+++ b/ibis/common/grounds.py
@@ -217,7 +217,7 @@ class Concrete(Immutable, Comparable, Annotable):
         return self.__precomputed_hash__
 
     def __equals__(self, other) -> bool:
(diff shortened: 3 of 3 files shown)


ibis/tests/benchmarks/test_benchmarks.py
```diff
@@ -809,3 +809,29 @@ def test_ibis_duckdb_to_pyarrow(benchmark, sql, ddb) -> None:
 
     expr = con.sql(sql)
     benchmark(expr.to_pyarrow)
+
+
+def test_large_join(benchmark):
```
cpcloud, line 814: This will be run when pushed to `master`, we don't run the benchmark suite in the `ibis-main.yml` workflow test suite to make sure it's fast.

cpcloud:
This PR makes `Node.equals` 1000x slower (~1 microsecond in `master`, ~1 millisecond here) due to ABC instance checks during traversal of the expression graph 😞

cpcloud:
There are some things we can do to get it down to about 500x slower. Not sure if there's other approaches that might result in less of a performance hit to `equals`.

ibis/expr/operations/core.py
```diff
@@ -23,7 +24,12 @@ def equals(self, other) -> bool:
             raise TypeError(
                 f"invalid equality comparison between Node and {type(other)}"
             )
-        return self.__cached_equals__(other)
+
+        lefts = Graph.from_bfs(self).toposort()
```
kszucs, line 28: This is pretty aggressive to me. Topological sort is not a cheap algorithm, and we shouldn't optimize ibis for unreasonable usage. Besides that  this is inconsistent with `self.op() == other.op()` which would trigger the original code path.

ibis/expr/operations/relations.py
```diff
@@ -395,7 +395,7 @@ def name(self) -> str:
             return f"{name}_ref"
         return util.gen_name("self_ref")
 
-    @property
+    @attribute
```
kszucs, line 398: This shouldn't contribute to the call stack, the calculated properties are not compared in `Concrete` subclasses just the `__args__`.

ibis/expr/operations/relations.py
```diff
@@ -310,14 +310,14 @@ class AnyLeftJoin(Join):
 
 @public
 class LeftSemiJoin(Join):
-    @property
+    @attribute
```
kszucs, line 313: For simple properties we shouldn't use attribute, and it is not contributing to the call stack, see my explanation below.

ibis/expr/operations/relations.py
```diff
@@ -529,7 +529,7 @@ def __init__(self, table, metrics, by, having, predicates, sort_keys):
             sort_keys=sort_keys,
         )
 
-    @attribute
+    @property
     def schema(self):
```
kszucs, line 533: Why the simply `schema` properties are turned into attributes while a complicated one like this turned into a property?

ibis/expr/operations/relations.py
```diff
@@ -572,7 +572,7 @@ class Distinct(Relation):
 
     table: Relation
 
-    @property
+    @attribute
```
kszucs, line 575: Same as above, I'd keep the original property.

ibis/expr/operations/core.py
cpcloud: > This is pretty aggressive to me. Topological sort is not a cheap algorithm, and we shouldn't optimize ibis for unreasonable usage

Agree, though it's hard to say what a ... reasonable definition of unreasonable is 😄 

> Besides that this is inconsistent with self.op() == other.op() which would trigger the original code path

How are verifying this? It doesn't seem to be true.

kszucs:
We can remove one third of the function calls by removing `Comparable.__cached_equals__` and moving that implementation directly to `__eq__`.

ibis/expr/operations/relations.py
cpcloud: Think about a deeply nested semi join. What would the `@property` behavior be?

ibis/expr/operations/relations.py
cpcloud: This was accidental. I agree, it should be an `attribute`.

ibis/expr/operations/relations.py
cpcloud: They don't directly contribute to the call stack, but the traversal (at least for binary relational operations) O(N ** 2) at the moment.

kszucs: Attributes are not traversed either.

ibis/expr/operations/relations.py
kszucs: 1000 is the usual recursion limit, this would be triggered after 1000 nested joins.

kszucs: I'm not totally against it since it has a negligible effect, but I'm not sure what expectations we have against our IR representation.

ibis/expr/operations/core.py
kszucs: `self.op() == other.op()` calls `Node.__eq__` which does the equality check recursively.

kszucs:
I have no objections against:
- using `@attribute`s more extensively 
- reducing the call stack by removing `Concrete.__cached_equals__`

on the other hand I'm objecting against doing expensive traversals for `.equals()` calls just to support a single and possibly rare edge case.

ibis/expr/operations/relations.py
cpcloud: **Something** is accessing `schema` during `left.equals(right)` so I'm not sure what you mean by "attributes aren't traversed"

ibis/expr/operations/relations.py
cpcloud: We're hitting the limit somewhere between 600 and 800 relations right now I believe.

cpcloud:
Let me see how far removing the `__cached_equals__` call gets us

kszucs:
Additionally we can have a shortcut checking whether the hash of the two comparable object are equal or not in 'Concrete.__equals__'. The we have a precomputed hash which should be the same for equal objects.

ibis/expr/operations/relations.py
cpcloud: Brought back.

kszucs (approved):

PR #5558, opened 2023-02-16
feat: implement ops.RandomScalar in all backends

krzysztof-kwitt:

PR #7149, opened 2023-09-14
fix(flink): translate `ops.RandomScalar` to `rand`

deepyaman:
Nit: Is there a reason why the default translation of `ops.RandomScalar` for SQL backends is `rand(utc_to_unix_micros(utc_timestamp()))` (originally from https://github.com/ibis-project/ibis/pull/5558/files), instead of that being a specific implementation for Impala? The way I understand it, Impala has the unique behavior of always returning the same values for `rand()`, so you need to essentially randomize the seed, but this shouldn't be necessary for most backends?

### DIFF
--- a/ibis/backends/flink/registry.py
+++ b/ibis/backends/flink/registry.py
@@ -195,6 +195,7 @@ operation_registry.update(
     {
         # Unary operations
         ops.NullIfZero: _nullifzero,
+        ops.RandomScalar: lambda *_: "rand()",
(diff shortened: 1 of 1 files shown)


cpcloud:
I think it's historical precedent that `rand` has a very Impala-specific implementation.

For a long time Impala was the only string generating backend, so the other string generating backends (BigQuery, and ClickHouse at the time) implemented their own which wasn't too burdensome.

cpcloud (approved):

PR #7151, opened 2023-09-14
feat(examples): name examples tables according to example name

cpcloud:
For illustrative purposes, especially when using `.sql` methods, having
examples tables named the same as the example reduces setup code.

### DIFF
--- a/ibis/examples/__init__.py
+++ b/ibis/examples/__init__.py
@@ -38,7 +38,7 @@ class Example(Concrete):
         name = self.name
 
         if table_name is None:
-            table_name = ibis.util.gen_name(f"examples_{name}")
+            table_name = name
 
         board = _get_board()
 
--- a/ibis/examples/tests/test_examples.py
+++ b/ibis/examples/tests/test_examples.py
@@ -65,8 +65,11 @@ def test_examples(example, example_con):
 
     assert example in repr(ex)
 
-    df = ex.fetch(backend=example_con).limit(1).execute()
-    assert len(df) == 1
+    t = ex.fetch(backend=example_con)
+    assert t.op().name == example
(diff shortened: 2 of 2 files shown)


ibis/examples/__init__.py
```diff
@@ -38,7 +38,7 @@ def fetch(
         name = self.name
 
         if table_name is None:
-            table_name = ibis.util.gen_name(f"examples_{name}")
```
jcrist, line 41: Do we have any SQL backends where creating a temporary table with the same name as a non-temporary table will clobber the non-temporary table? If so, dropping the UUID suffix here could result in data loss if the user has any tables that match the example name.

jcrist: We might also decide this isn't a problem we care about (what user would load example data into a production database?). On further thought even if the above is true for some backends, IMO the risk here is low and the benefits of consistent naming feel worth it.

jcrist (approved):
LGTM

ibis/examples/__init__.py
cpcloud: AFAIK all of our backends that support temporary tables either fail with an error about already existing tables, or implement shadowing, where the temporary table takes precedence over the non-temporary when resolving table names.

cpcloud: DuckDB (for example) implements shadowing:

```
D create table t as select 1 as a;
D create temp table t as select 1 as b;
D table t;
┌───────┐
│   b   │
│ int32 │
├───────┤
│     1 │
└───────┘
D drop table t;
D table t;
┌───────┐
│   a   │
│ int32 │
├───────┤
│     1 │
└───────┘
```

PR #7152, opened 2023-09-14
feat(polars): implement new UDF API

cpcloud:
Adds support for the new scalar UDF API to the polars backend.

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -1187,7 +1187,8 @@ _UDF_INVOKERS = {
 
 @translate.register(ops.ScalarUDF)
 def execute_scalar_udf(op, **kw):
-    if (input_type := op.__input_type__) in (InputType.PYARROW, InputType.PYTHON):
+    input_type = op.__input_type__
+    if input_type in _UDF_INVOKERS:
         dtype = op.dtype
         return pl.map_batches(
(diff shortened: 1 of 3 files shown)


ibis/backends/polars/compiler.py
```diff
@@ -1167,10 +1167,37 @@ def execute_count_distinct_star(op, **kw):
...
+    InputType.PYARROW: lambda func, dtype, args: pl.from_arrow(
+        func(*(arg.to_arrow() for arg in args)).cast(dtype.to_pyarrow()),
+    ),
+}
+
+
 @translate.register(ops.ScalarUDF)
 def execute_scalar_udf(op, **kw):
-    if op.__input_type__ == InputType.BUILTIN:
+    if (input_type := op.__input_type__) in (InputType.PYARROW, InputType.PYTHON):
```
jcrist, line 1190: Two nits:
- Can use `_UDF_INVOKERS` instead of a tuple for membership checking here
- Since `input_type` is used in every branch below, it feels cleaner to me to define it at the top-level rather than using `:=`. IMO `:=` is clearest when the new variable is only used in the block below the statement (e.g. within the corresponding `if`/`for` block).

```suggestion
    input_type = op.__input_type__
    if input_type in _UDF_INVOKERS:
```

jcrist (approved):
One small fixup, otherwise this LGTM.

PR #7158, opened 2023-09-14
docs: add prelude lines to all code samples

NickCrews:
Inspired by a problem I encountered in #7159 

I'm not sure if this is a good idea, but I just wanted to throw it out there in case
it interests you.

If you want to actually use this, we need to go through all the docstrings and
remove these lines from them, otherwise
they are included twice!

Could also just import ibis, and
not set interactive.

### DIFF
--- a/docs/_renderer.py
+++ b/docs/_renderer.py
@@ -21,6 +21,12 @@ class Renderer(qd.MdRenderer):
         expect_failure = "quartodoc: +EXPECTED_FAILURE"
         quartodoc_skip_doctest = "quartodoc: +SKIP"
 
+        # Avoid needing this boilerplate in every example and docstring
+        prelude_lines = [
+            ">>> import ibis",
+            ">>> ibis.options.interactive = True",
+        ]
+
         chunker = lambda line: line.startswith((prompt, continuation))
         should_skip = (
             lambda line: quartodoc_skip_doctest in line or skip_doctest in line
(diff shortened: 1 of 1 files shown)


docs/_renderer.py
```diff
@@ -21,6 +21,12 @@ def render(self, el: qd.ast.ExampleCode) -> str:
         expect_failure = "quartodoc: +EXPECTED_FAILURE"
         quartodoc_skip_doctest = "quartodoc: +SKIP"
 
+        # Avoid needing this boilerplate in every example and docstring
+        prelude_lines = [
+            ">>> import ibis",
+            ">>> ibis.options.interactive = True",
```
cpcloud, line 27: Some of our examples aren't interactive, and it might be weird to see

```
>>> import ibis
>>> ibis.options.interactive = False
```

in those examples.

Another awkward thing is having `import ibis` at the beginning of **every** code example.

For example

```python
"""Docstring for function.

Examples
--------
Do some stuff

>>> import ibis

Continue doing stuff based on previous stuff

>>> import ibis
>>> ibis.options.interactive = True  # confusing why this is here again
"""
```

NickCrews: Yes, that seems unworkable.

What if we put this prelude stuff in a separate code block, with a directive to quarto to hide the whole thing.

cpcloud: I'm not sure what we get out of that. While it _is_ boilerplate, I'm not sure it's worth the hidden code that's being executed on the developer's behalf.

NickCrews: Roger. I'm fine closing this out then!

cpcloud: Thanks! Keep the PRs coming!

PR #7159, opened 2023-09-14
Docs: Reorganize reference v2

NickCrews:
This is a followup per #7141

- Move every datatype-specific API out of the top-level section. eg ibis.array() got moved to collections section
- Now only backend-specific stuff got left in the top-level section. Therefore, I renamed it to "Connections API".

In addition to these high-level changes, also a few tweaks:
- soft deprecate ibis.where
- fix ibis.ifelse docstring (plenty other functions like this are still missing docstrings, but at least the existing ones should be correct)
- import ibis in a doctest to prevent a rendering failure. This was little weird. This wasn't failing before only due to the ordering of this function in the larger API tab. All of the members in a a quartodoc section get squashed into a single .qmd file. Then quarto goes through and evaluates all the docstrings. So if the very first docstring in each API page does `import ibis`, then we are all good. But due to the reordering I did, this no longer happened here, so I had to add this import. At a higher-level, this isn't great, because it means different docstrings share state, and that depends on their ordering in a page. yuck. But maybe not worth it to fix. Possibly we could del all existing variables with some prelude lines, similar to what is done in https://github.com/ibis-project/ibis/pull/7158, but IDK this might all be overkill.

Things that I think should still happen in a followup PR:
- I want it to be more obvious how top level stuff is accessed from ibis. e.g. looking at `now()` in the temporal section, I wish the function signature was `ibis.now()`.
- Make things appear in the correct place where possible, eg `ibis.random()` instead of `ibis.expr.api.random()`.

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -140,7 +140,7 @@ website:
       contents:
         - section: Expression API
           contents:
-            - reference/top_level.qmd
+            - reference/connection.qmd
             - reference/expression-tables.qmd
             - reference/selectors.qmd
             - reference/expression-generic.qmd
@@ -191,10 +191,10 @@ quartodoc:
       contents:
         - kind: page
           package: ibis
-          path: top_level
+          path: connection
           summary:
-            name: Top-level APIs
-            desc: Methods and objects available directly on the `ibis` module.
+            name: Connection API
+            desc: Create and manage backend connections.
           contents:
             - name: connect
               package: ibis.backends.base
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -140,7 +140,6 @@ website:
       contents:
         - section: Expression API
           contents:
-            - reference/connection.qmd
             - reference/expression-tables.qmd
             - reference/selectors.qmd
             - reference/expression-generic.qmd
@@ -158,6 +157,10 @@ website:
         - section: UDFs
           contents:
             - reference/scalar-udfs.qmd
+
+        - section: Connection APIs
+          contents:
+            - reference/connection.qmd
 
         - section: Configuration
           contents:
@@ -189,20 +192,6 @@ quartodoc:
       desc: "APIs for manipulating table, column and scalar expressions"
       package: ibis.expr.types
       contents:
-        - kind: page
-          package: ibis
-          path: connection
-          summary:
-            name: Connection API
-            desc: Create and manage backend connections.
-          contents:
-            - name: connect
-              package: ibis.backends.base
-            - name: get_backend
-              dynamic: true
-            - name: set_backend
-              dynamic: true
-
         - kind: page
           path: expression-tables
           package: ibis.expr.types.relations
@@ -305,27 +294,35 @@ quartodoc:
 
         - kind: page
           path: expression-numeric
-          package: ibis.expr.types.numeric
           summary:
             name: Numeric and Boolean expressions
             desc: Integer, floating point, decimal, and boolean expressions.
           contents:
-            - NumericValue
-            - NumericColumn
-            - IntegerValue
-            - IntegerColumn
-            - FloatingValue
-            - DecimalValue
+            - name: NumericValue
+              package: ibis.expr.types.numeric
+            - name: NumericColumn
+              package: ibis.expr.types.numeric
+            - name: IntegerValue
+              package: ibis.expr.types.numeric
+            - name: IntegerColumn
+              package: ibis.expr.types.numeric
+            - name: FloatingValue
+              package: ibis.expr.types.numeric
+            - name: DecimalValue
+              package: ibis.expr.types.numeric
             - name: BooleanValue
               package: ibis.expr.types.logical
             - name: and_
               dynamic: true
+              package: ibis
             - name: or_
               dynamic: true
+              package: ibis
             - name: negate
               package: ibis.expr.api
             - name: random
               dynamic: true
+              package: ibis
 
         - kind: page
           path: expression-strings
@@ -478,6 +475,23 @@ quartodoc:
               package: ibis
               dynamic: true
             - Schema
+
+    - title: Connection APIs
+      package: ibis.expr.operations.udf
+      contents:
+        - kind: page
+          package: ibis
+          path: connection
+          summary:
+            name: Top-level connection APIs
+            desc: Create and manage backend connections.
+          contents:
+            - name: connect
+              package: ibis.backends.base
+            - name: get_backend
+              dynamic: true
+            - name: set_backend
+              dynamic: true
 
     - title: UDFs
       desc: "User-defined function APIs"
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -196,21 +196,15 @@ quartodoc:
             name: Top-level APIs
             desc: Methods and objects available directly on the `ibis` module.
           contents:
-            - name: array
-              dynamic: true
             - name: connect
               package: ibis.backends.base
             - name: dtype
               dynamic: true
             - name: get_backend
               dynamic: true
-            - name: map
-              dynamic: true
             - name: schema
               dynamic: true
             - name: set_backend
-              dynamic: true
-            - name: struct
               dynamic: true
 
         - kind: page
@@ -376,16 +370,23 @@ quartodoc:
 
         - kind: page
           path: expression-collections
+          package: ibis
           summary:
             name: Collection expressions
             desc: Arrays, maps and structs.
           contents:
             - name: ArrayValue
               package: ibis.expr.types.arrays
-            - name: StructValue
-              package: ibis.expr.types.structs
             - name: MapValue
               package: ibis.expr.types.maps
+            - name: StructValue
+              package: ibis.expr.types.structs
+            - name: array
+              dynamic: true
+            - name: map
+              dynamic: true
+            - name: struct
+              dynamic: true
 
         - kind: page
           path: expression-geospatial
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -200,37 +200,17 @@ quartodoc:
               dynamic: true
             - name: connect
               package: ibis.backends.base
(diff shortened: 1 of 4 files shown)


NickCrews:
I just fixed those broken links, but this is ready for review! It is a little hefty, so I appreciate the effort!

docs/_quarto.yml
```diff
@@ -138,9 +130,16 @@ website:
       style: "docked"
       collapse-level: 2
       contents:
+        - section: Backend API
+          contents:
+            - support_matrix.qmd
+            - reference/connection.qmd
+            - section: Specific Backend APIs
+              contents:
+                - auto: backends/*.qmd
```
cpcloud, line 139: I like our existing backend layout.

Can remove this section and the separate connection section?

How about a how-to that describes how to use `connect`, `set_backend` and `get_backend` instead, that shows their API.

Alternatively, we can have a page that has the base backend API, along with these top level functions.

Either of these can be done in a follow-up.

cpcloud (changes requested):
If you can remove the changes to the backend layout this LGTM!

NickCrews:
I am getting

```
Warning: The computation used to determine source line information from the engine timed out.
Line number information will be unavailable and/or inaccurate.
```
is this to be expected or did I break something?

lostmygithubaccount:
> I am getting
> 
> ```
> Warning: The computation used to determine source line information from the engine timed out.
> Line number information will be unavailable and/or inaccurate.
> ```
> 
> is this to be expected or did I break something?

I've never seen this before working w/ Quarto 

I will give a drive-by comment while I'm here that I don't agree w/ hiding the backend docs within the API docs. I understand that it makes sense, as they are API docs to a large extent, but it's useful for users IMO to easily see the list of available backends at a top-level and we can do more platform-specific content in those pages going forward

NickCrews:
Ah I think my latest screenshot was confusing. The backend stuff is NOT underneath the rest of the API stuff, it's in its own top level tab as it was before.

docs/_quarto.yml
```diff
@@ -309,6 +267,37 @@ quartodoc:
...
+            - name: case
+              package: ibis
+              dynamic: true
+            - name: show_sql
+              package: ibis
+              dynamic: true
+            - name: to_sql
+              package: ibis
+              dynamic: true
+            - name: get_backend
```
cpcloud, line 298: It looks like this documented in two places: once in `expression-generic` and once in `connection`.

Is that intentional?

docs/_quarto.yml
```diff
@@ -325,6 +314,23 @@ quartodoc:
...
+            - name: ifelse
+              package: ibis.expr.api
+            - name: and_
+              package: ibis
+              dynamic: true
+            - name: or_
+              package: ibis
+              dynamic: true
+            - name: least
+              package: ibis.expr.api
```
cpcloud, line 326: `least` and `greatest` should probably be under generic expressions, after all `"a" < "b"` is a well-defined operation for most (all?) backends.

cpcloud: An aside for another PR, maybe we want to aliases these to `max` and `min` for Python users' familiarity. No strong thoughts, just popped into my head.

docs/_quarto.yml
```diff
@@ -325,6 +314,23 @@ quartodoc:
             - DecimalValue
             - name: BooleanValue
               package: ibis.expr.types.logical
+            - name: ifelse
```
cpcloud, line 317: I think I would put `ifelse` in generic.

It's closer to the notion of "branching", part of which is of course boolean logic related.

Perhaps a good rule of thumb is:

- If the operation returns or operates on a large variety of types it should go in generic
- Otherwise it should go in the specific docs

I am aware that `ifelse` is defined on bool expressions :)

cpcloud (approved):
This LGTM!

@lostmygithubaccount Can you give a review?

docs/_quarto.yml
NickCrews: Sure! I would vote for putting it in both sections, instead of moving it. What do you think?

docs/_quarto.yml
NickCrews: Yes. I think this is a functionality that is non-obvious to people, they aren't going to be looking for it. SQL, pandas, Polars, etc have no notion of this. So, I wanted to make it discoverable.

ifelse, on the other hand, people expect to exist and so are going to be looking for it, so that doesn't need to be advertised as much.

I considered the danger of these two places getting out of sync, but I think that is a low chance/impossible, since this is mostly a pointer to the docstring that actually has the content

lostmygithubaccount:
where is this coming from? should be "Top-level backend API"

image

docs/_quarto.yml
NickCrews: Oops, yes these should be under generic.

I'm +0 on adding min/max aliases. In that other issue we just removed .where so there is only .ifelse, but this feels warranted. Including the keywords min and max in the docstring would at least make it so I would find it

lostmygithubaccount (approved):
LGTM -- can ignore that nit but if it's easy to fix that'd be nice for consistency

NickCrews:
I will push those two fixes!

NickCrews:
- fixed the title (I messed up that screenshot, it was a from a stale build of quartodoc, now it is correct)
- moved least, greatest, ifelse to ALSO be in generic. I kept them in the old place too for discoverability.

docs/_quarto.yml
```diff
@@ -111,6 +111,7 @@ website:
       collapse-level: 2
       contents:
         - support_matrix.qmd
+        - reference/connection.qmd
```
cpcloud, line 114: Can you move this to be only in the API docs? I feel like we should keep the backend tab to only contain the support matrix and the backends and nothing else. Otherwise it looks cluttered to me.

docs/_quarto.yml
cpcloud: I think I'd rather have it in single place so that there aren't multiple search results that could be confusing. I see this with sphinx sometimes and I really dislike it.

docs/_quarto.yml
cpcloud: Generally not in favor of duplicating API documentation like this. I think there are better ways to advertise the API without duplicating its docs, like a blog post or a how-to.

docs/_quarto.yml
NickCrews: Sure. Under what section and what should the page be named?

machow:
RE 

> I want it to be more obvious how top level stuff is accessed from ibis. e.g. looking at now() in the temporal section, I wish the function signature was ibis.now().

Just a heads up, in quartodoc you can set `signature_name: full` to show the full path to an object in its signature (e.g. `ibis.now()`). The tricky thing is that you probably don't want the long signatures in methods, so you may need to set it just for those top level functions. (It looks like signature_name is undocumented :/  ).

Here are the docs on setting options: https://machow.github.io/quartodoc/get-started/basic-content.html#setting-default-options

docs/_quarto.yml
```diff
@@ -187,118 +187,76 @@ quartodoc:
...
-            - name: param
-              dynamic: true
-            - name: show_sql
-              dynamic: true
-            - name: to_sql
-              dynamic: true
-            - name: random
-              dynamic: true
-            - name: range_window
+            - name: set_backend
```
cpcloud, line 202: Actually can you add these in a follow up PR? It's hard to see the structural changes here without building the docs.

docs/_quarto.yml
cpcloud: Can you add it this in a follow up PR? In that PR I think you remove this section and let the API docs show up as they usually do under the `Reference` tab.

docs/_quarto.yml
NickCrews: OK, now it is just in generic in my latest commit stack.

docs/_quarto.yml
NickCrews: Not quite sure what you mean by "these"? In my latest push I didn't move anything to underneath the Backends tab. Now just the "top-level" section only contains connection stuff and is renamed "Connection API". Hopefully this is clear enough.

I wish quarto had a preview feature for PRs like this, like readthedocs does... :(

NickCrews: Oh, do you mean just remove commit b3429e9597a0ab73aaddf3a97e4db80345c5f768 from this PR and follow up with that later?

cpcloud: Sorry! By "these" I mean the `set_backend`/`get_backend`/`connect` methods.

NickCrews:
OK, I removed https://github.com/ibis-project/ibis/commit/b3429e9597a0ab73aaddf3a97e4db80345c5f768 from this PR because it is a bit unrelated and clutters up the main structure here.

In a followup PR I want to
- include https://github.com/ibis-project/ibis/commit/b3429e9597a0ab73aaddf3a97e4db80345c5f768
- change to `signature_name: full` as @machow describes

docs/_quarto.yml
NickCrews: I assumed so, so I removed it from the commit stack.

NickCrews: Ah, now I see your clarification, thanks! I think now your concern is rendered useless, as now these methods aren't getting moved. Everything else is just getting moved out, and this section is getting renamed.

docs/_quarto.yml
```diff
@@ -479,6 +476,23 @@ quartodoc:
               dynamic: true
             - Schema
 
+    - title: Connection APIs
+      package: ibis.expr.operations.udf
```
NickCrews, line 480: this top-level package can get removed.

cpcloud: Thanks

cpcloud:
I think this is good to go in now!

I moved Connections API to its own section, and fixed a few failing references.

cpcloud (approved):

NickCrews:
Money, thank you so much for finishing this up and getting it merged!

PR #7162, opened 2023-09-15
feat: just chat to open Zulip in terminal

lostmygithubaccount:
opening for discussion -- I think just chat would be a very convenient command when working in the repo to instantly open Zulip chat in the terminal

should probably adjust the justfile to pass through args (note some of the themes don't seem to work out of the box on Mac) and add to dev-requirements or something (note 3.11 is not supported on released versions, I'm installing from git+https://)

### DIFF
--- a/justfile
+++ b/justfile
@@ -156,3 +156,7 @@ docs-build-all:
     just docs-apigen --verbose
     just docs-render
     just checklinks docs/_output --offline --no-progress
+
+# open chat
+chat *args:
+    zulip-term {{ args }}
--- a/.gitignore
+++ b/.gitignore
@@ -128,3 +128,6 @@ prof/
 # generated examples data
 ibis/examples/data
 ibis/examples/descriptions
+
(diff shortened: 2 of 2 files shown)


justfile
```diff
@@ -156,3 +156,7 @@ docs-build-all:
     just docs-apigen --verbose
     just docs-render
     just checklinks docs/_output --offline --no-progress
+
+# open chat
+chat:
+    zulip-term --color-depth 24bit -c ./.zuliprc
```
cpcloud, line 162:
```suggestion
chat *args:
    zulip-term {{ args }}
```

Not a huge fan of overriding the default behavior of the program, especially since the config file can contain plaintext secrets and will be placed in the current directory if it doesn't exist 😬 .

You can also set all the other settings including color depth in the config file, which IMO obviates the need for passing any arguments here.

cpcloud (approved):
LGTM

PR #7166, opened 2023-09-15
fix(trino): differentiate between a single column struct and a column when chaining unnest expressions with field accesses

cpcloud:
Fixes a bug in the Trino backend when accessing fields on an `array<struct<...>>`.


jcrist (approved):
Presuming the lack of coverage indicators are incorrect (IIUC those lines _should_ be covered), this all LGTM.

PR #7173, opened 2023-09-18
depr(api): deprecate top-level `ibis.geo_*` functions in favor of their corresponding methods

jcrist:
For historical reasons all geo methods were re-exported as top-level functions. This API duplication isn't necessary, and isn't done for other types (i.e. there is no `ibis.sin` to correspond to `NumericValue.sin`). We deprecate these APIs here in favor of their corresponding methods.


jcrist:
```python
In [5]: t = ibis.table({"x": "polygon"})

In [6]: ibis.geo_area(t.x)
<ipython-input-6-ea546f2e3bf5>:1: FutureWarning: `geo_area` is deprecated as of v7.0; use the `GeoSpatialValue.area` method instead
  ibis.geo_area(t.x)
Out[6]: 
r0 := UnboundTable: unbound_table_1
  x polygon

GeoArea(x): GeoArea(r0.x)
```

ibis/backends/postgres/tests/test_geospatial.py
```diff
@@ -192,7 +192,7 @@ def test_geo_spatial_unops(geotable, expr_fn, expected):
             marks=pytest.mark.notimpl(["postgres"]),
         ),
         param(
-            lambda t: t.geo_polygon.contains(ibis.geo_point(30, 10)),
+            lambda t: t.geo_polygon.contains(ibis.literal(30).point(10)),
```
cpcloud, line 195: We many want to add these constructor-like functions as proper APIs at some point, but that can be done later.

cpcloud (approved):

PR #7174, opened 2023-09-18
ci(flink): don't generate now-unused 3.10 lockfile

deepyaman:
Should have been dropped in #6877


cpcloud (approved):
Nice! Save those CI minutes!

PR #7181, opened 2023-09-19
docs: fix zulip link for new members

lostmygithubaccount:
updates to use a join-as-member link that never expires, my bad on missing this earlier


cpcloud (approved):
Thanks!

PR #7189, opened 2023-09-21
fix(clip): preserve nulls when clipping

### LINKED ISSUE #7168: bug: .clip() doesn't preserve NULLs
### What happened?

I would expect nulls to stay nulls. But they get turned into the upper value (if upper bound is provided) or the lower bound (if no upper bound is provided)

```python
import ibis

t = ibis.memtable({"values": [0, 1.0, 2.0, 3.0, 4.0, None]})
t.mutate(clipped1=t.values.clip(lower=2.5, upper=3), clipped2=t.values.clip(lower=2.5))
┏━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┓
┃ values  ┃ clipped1 ┃ clipped2 ┃
┡━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━┩
│ float64 │ float64  │ float64  │
├─────────┼──────────┼──────────┤
│     0.0 │      2.5 │      2.5 │
│     1.0 │      2.5 │      2.5 │
│     2.0 │      2.5 │      2.5 │
│     3.0 │      3.0 │      3.0 │
│     4.0 │      3.0 │      4.0 │
│     nan │      3.0 │      2.5 │
└─────────┴──────────┴──────────┘
```

main

duckdb, maybe this is irrelevant. Not sure how different backends handle NULLs in GREATEST() and LEAST()

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Ensure that NULLs are propagated through `clip`

Fixes #7168.

### DIFF
--- a/ibis/backends/pyspark/compiler.py
+++ b/ibis/backends/pyspark/compiler.py
@@ -775,11 +775,11 @@ def compile_clip(t, op, **kwargs):
 
     def column_min(value, limit):
         """Return values greater than or equal to `limit`."""
-        return F.when(value < limit, limit).otherwise(value)
+        return F.when((value < limit) & ~F.isnull(value), limit).otherwise(value)
 
     def column_max(value, limit):
         """Return values less than or equal to `limit`."""
-        return F.when(value > limit, limit).otherwise(value)
+        return F.when((value > limit) & ~F.isnull(value), limit).otherwise(value)
 
     def clip(column, lower_value, upper_value):
         return column_max(column_min(column, F.lit(lower_value)), F.lit(upper_value))
--- a/ibis/backends/tests/test_numeric.py
+++ b/ibis/backends/tests/test_numeric.py
@@ -1364,10 +1364,12 @@ def test_random(con):
 @pytest.mark.parametrize(
     ("ibis_func", "pandas_func"),
     [
-        (lambda x: x.clip(lower=0), lambda x: x.clip(lower=0)),
-        (lambda x: x.clip(lower=0.0), lambda x: x.clip(lower=0.0)),
-        (lambda x: x.clip(upper=0), lambda x: x.clip(upper=0)),
-        pytest.param(
+        param(lambda x: x.clip(lower=0), lambda x: x.clip(lower=0), id="lower-int"),
+        param(
+            lambda x: x.clip(lower=0.0), lambda x: x.clip(lower=0.0), id="lower-float"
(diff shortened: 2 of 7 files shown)


ibis/expr/types/numeric.py
```diff
@@ -198,7 +200,13 @@ def clip(
         --------
         >>> import ibis
         >>> ibis.options.interactive = True
-        >>> t = ibis.memtable({"values": range(8)})
+        >>> t = ibis.memtable(
```
NickCrews, line 203: nit: can you just add some NULLs to the existing column of data? I think that shows all aspects of clip, and is less complicated and less lines

ibis/backends/tests/test_numeric.py
```diff
@@ -1364,25 +1364,40 @@ def test_random(con):
...
-        (
+        param(
             lambda x: x.clip(lower=0, upper=1.0),
             lambda x: x.clip(lower=0, upper=1.0),
+            id="lower-upper-float",
+        ),
+        param(
+            lambda x: x.nullif(1).clip(lower=0),
+            lambda x: x.where(x != 1).clip(lower=0),
+            id="nullif-lower",
```
NickCrews, line 1395: nit: change id to "null-preserved-lower" or simply "null-lower", to show the intention of the test. "nullif" is more a description of how you accomplished that.

ibis/backends/pyspark/compiler.py
```diff
@@ -775,11 +775,11 @@ def compile_clip(t, op, **kwargs):
 
     def column_min(value, limit):
         """Return values greater than or equal to `limit`."""
-        return F.when(value < limit, limit).otherwise(value)
+        return F.when((value < limit) & ~F.isnull(value), limit).otherwise(value)
```
NickCrews, line 778: Is this codecov warning accurate? It looks to me like this should get hit every time `compile_clip()` is called. Is `compile_clip()` not getting called at all?

NickCrews (approved):
few nits, but looks good, thanks!

ibis/backends/pyspark/compiler.py
cpcloud: It's probably not accurate.

ibis/backends/tests/test_numeric.py
cpcloud: Done.

PR #7192, opened 2023-09-21
fix(flink): rewrite `ops.Date` as a cast operation + fix(flink): fix translation of `ops.TimestampDiff`

deepyaman:


ibis/backends/flink/registry.py
```diff
@@ -63,18 +72,20 @@ def _filter(translator: ExprTranslator, op: ops.Node) -> str:
...
+def _timestamp_from_unix(translator: ExprTranslator, op: ops.Node) -> str:
+    arg, unit = op.args
 
-def _nullifzero(translator: ExprTranslator, op: ops.Literal) -> str:
-    casted = translate_literal(ops.Literal("0", dtype=op.dtype))
-    return f"NULLIF({translator.translate(op.arg)}, {casted})"
+    if unit == TimestampUnit.MILLISECOND:
+        return f"TO_TIMESTAMP_LTZ({helpers.quote_identifier(arg.name, force=True)}, 3)"
+    elif unit == TimestampUnit.SECOND:
+        return f"TO_TIMESTAMP_LTZ({helpers.quote_identifier(arg.name, force=True)}, 0)"
```
jcrist, line 87: Nit - these lines look identical except for the 3/0 as the 2nd arg. Could you refactor to reduce the code duplication?

jcrist (commented):
One tiny nit, otherwise LGTM

ibis/backends/flink/registry.py
deepyaman: Done! I think it probably copies something similar from, say, `mssql` backend.

deepyaman: (But it didn't feel clear enough value to reduce the duplication on that backend, from what I saw.)

jcrist: Thanks! And yeah, no need to fixup existing code now; just trying to do small cleanups as new code is added.

jcrist (approved):

PR #7198, opened 2023-09-22
fix(sql): unbind expressions when generating sql to avoid backend-specific API calls

cpcloud:
Allow generating sql from any backend.


jcrist (approved):
Fix LGTM!

A bit unfortunate that we have n-backend new SQL snapshot files for SQL that should be ~the same across all backends. We're always generating duckdb sql, so the only differences are the input table names and the auto-generated column names (due to differences between the `functional_alltypes` implementation in each backend). Not a big deal though, files are cheap.

jcrist:
Gah, gave an approval review on a PR that's failing tests due to dependencies. That's what I get for reviewing the code before waiting for CI to pass :/.

cpcloud:
This is proving to be more annoying than I thought.

There are some issues around propagating databases/catalogs/schemas between unbound expressions. We may need to address that eventually, but I don't think the changes in this PR warrant addressing that yet.

PR #7199, opened 2023-09-22
feat(flink): implement insert dml

chloeh13q:
Implement INSERT DML in Flink backend.

### DIFF
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -8,6 +8,7 @@ import sqlglot as sg
 import ibis.common.exceptions as exc
 import ibis.expr.operations as ops
 import ibis.expr.schema as sch
+import ibis.expr.types as ir
 from ibis.backends.base import BaseBackend, CanCreateDatabase
 from ibis.backends.base.sql.ddl import fully_qualified_re, is_fully_qualified
 from ibis.backends.flink.compiler.core import FlinkCompiler
@@ -16,6 +17,7 @@ from ibis.backends.flink.ddl import (
     CreateTableFromConnector,
     DropDatabase,
     DropTable,
+    InsertSelect,
 )
 
 if TYPE_CHECKING:
@@ -24,8 +26,8 @@ if TYPE_CHECKING:
     import pandas as pd
     import pyarrow as pa
     from pyflink.table import TableEnvironment
+    from pyflink.table.table_result import TableResult
 
-    import ibis.expr.types as ir
     from ibis.expr.streaming import Watermark
 
 
@@ -41,7 +43,7 @@ class Backend(BaseBackend, CanCreateDatabase):
         Parameters
         ----------
         table_env
-            A table environment
+            A table environment.
 
         Examples
         --------
@@ -53,8 +55,8 @@ class Backend(BaseBackend, CanCreateDatabase):
         """
         self._table_env = table_env
 
-    def _exec_sql(self, query: str) -> None:
-        self._table_env.execute_sql(query)
+    def _exec_sql(self, query: str) -> TableResult:
+        return self._table_env.execute_sql(query)
 
     def list_databases(self, like: str | None = None) -> list[str]:
         databases = self._table_env.list_databases()
@@ -169,11 +171,11 @@ class Backend(BaseBackend, CanCreateDatabase):
         Parameters
         ----------
         name
-            Table name
+            Table name.
         database
-            Database in which the table resides
+            Database in which the table resides.
         catalog
-            Catalog in which the table resides
+            Catalog in which the table resides.
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


ibis/backends/flink/__init__.py
```diff
@@ -3,13 +3,15 @@
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any
 
+import pandas as pd
```
gforsyth, line 6: I think elsewhere (like in `create_table`) we're gating this import inside of the method that needs to the `isinstance` check.  This should probably get moved down inside of `insert`

ibis/backends/flink/__init__.py
```diff
@@ -442,3 +444,67 @@ def _get_operations(cls):
...
         return operation in cls._get_operations()
+
+    def insert(
+        self,
+        table_name: str,
+        obj: pd.DataFrame | ir.Table | list | dict,
+        database: str | None = None,
+        catalog: str | None = None,
+        overwrite: bool = False,
+    ) -> TableResult:
```
gforsyth, line 455: Ibis itself is already inconsistent about whether `insert` calls return anything.  The other sqlalchemy backends return `None` -- Impala and PySpark both return some `result` object. 
Is there useful information encoded in the `TableResult`?  If not, I'd say we can just execute and not return it.

ibis/backends/flink/__init__.py
```diff
@@ -442,3 +444,67 @@ def _get_operations(cls):
...
+        """Insert data into a table.
+
+        Parameters
+        ----------
+        table_name
+            The name of the table to which data needs will be inserted
```
gforsyth, line 461:
```suggestion
            The name of the table to insert data into.
```

ibis/backends/flink/__init__.py
```diff
@@ -442,3 +444,67 @@ def _get_operations(cls):
...
+        obj
+            The source data or expression to insert
```
gforsyth, line 463:
```suggestion
            The source data or expression to insert.
```

ibis/backends/flink/__init__.py
```diff
@@ -442,3 +444,67 @@ def _get_operations(cls):
...
+        database
+            Name of the attached database that the table is located in.
+        catalog
+            Name of the attached catalog that the table is located in.
+        overwrite
+            If `True` then replace existing contents of table
```
gforsyth, line 469:
```suggestion
            If `True` then replace existing contents of table.
```

ibis/backends/flink/__init__.py
```diff
@@ -442,3 +444,67 @@ def _get_operations(cls):
     @classmethod
     def has_operation(cls, operation: type[ops.Value]) -> bool:
         return operation in cls._get_operations()
+
+    def insert(
+        self,
+        table_name: str,
+        obj: pd.DataFrame | ir.Table | list | dict,
```
gforsyth, line 451: `create_table` also accepts `pyarrow.Table` -- do we also want to accept that here?

gforsyth (commented):
Looks good overall -- a few questions and a few small nits.

ibis/backends/flink/__init__.py
chloeh13q: When `TableResult` is returned, we can call `wait()` on the `TableResult` to wait for the job to finish. I think it's at least useful for debugging purposes for now, but I don't feel strongly about returning it, although `_exec_sql` also just returns the `TableResult`.

ibis/backends/flink/__init__.py
chloeh13q: That's a good point! We definitely can - I think we can just convert pyarrow tables to pandas tables and insert them the same way.

FWIW though, the `insert()` method in alchemy and pyspark backends don't accept pyarrow tables.

chloeh13q: Added `pyarrow.Table`!

ibis/backends/flink/__init__.py
gforsyth: That seems like a good reason to return it!

ibis/backends/flink/tests/test_ddl.py
```diff
@@ -94,15 +113,56 @@ def test_create_table(
...
+    sink_schema = sch.Schema({"name": dt.string, "age": dt.int64, "gpa": dt.float64})
+    with tempfile.TemporaryDirectory() as tempdir:
+        con.create_table(
+            "tempdir_sink",
+            schema=sink_schema,
+            tbl_properties=tempdir_sink_configs(tempdir),
+        )
+        con.insert(
+            "tempdir_sink",
+            [("fred flintstone", 35, 1.28), ("barney rubble", 32, 2.32)],
```
gforsyth, line 140: I think this should be parametrized to also cover inserting DataFrames and pyarrow tables.

gforsyth (commented):
two small nits, then I think this is good to go in!

ibis/backends/flink/__init__.py
```diff
@@ -3,11 +3,13 @@
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any
 
+import pyarrow as pa
```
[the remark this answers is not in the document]
chloeh13q: Good catch - sorry accidentally moved it outside

gforsyth (approved):
:shipit:

PR #7208, opened 2023-09-24
perf: introduce quicker abstract base classes

kszucs:
Almost all of the ibis objects have been extending `ABC` which hooks into the instance and subclass checks making the very common `isinstance()` calls more expensive. It had a significant impact based on occasional code profilings. 

I was thinking on how could we avoid the performance impact on instance checks while keep supporting the handy abstract methods and properties. Turns out that we can easily mitigate `ABC`'s overhead by setting the `__abstractmethods__` attribute for the created types and the python internals automatically raises the "Can't instantiate abstract class Foo with abstract methods bar, foo" error. This way we provide our own, less flexible abstract base classes using the python builtin isinstance checks.

On direct isinstance checks against base classes the improvement is significant:

```console
-------------------------------------------------------------------------------- benchmark 'test_concrete_isinstance': 2 tests ---------------------------------------------------------------------------------
Name (time in us)                               Min                Max               Mean            StdDev             Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
test_concrete_isinstance (0292_b0b62cc)     21.5420 (2.21)     57.0000 (1.21)     22.2807 (2.24)     0.8539 (1.29)     22.5000 (2.28)     0.9160 (11.04)    1153;309       44.8818 (0.45)      20051           1
test_concrete_isinstance (NOW)               9.7500 (1.0)      47.2920 (1.0)       9.9492 (1.0)      0.6595 (1.0)       9.8750 (1.0)      0.0830 (1.0)     1050;8082      100.5107 (1.0)       69566           1
```

On average benchmarks show 10-30% performance gain: https://gist.github.com/kszucs/814e4336df010fa356012ef04b729c11

### DIFF
--- a/ibis/common/bases.py
+++ b/ibis/common/bases.py
@@ -1,12 +1,11 @@
 from __future__ import annotations
 
-from abc import ABCMeta, abstractmethod
-from collections.abc import Hashable
+import collections.abc
+from abc import abstractmethod
 from typing import TYPE_CHECKING, Any
 from weakref import WeakValueDictionary
 
 from ibis.common.caching import WeakCache
-from ibis.common.collections import FrozenDict
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -14,7 +13,7 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
 
-class BaseMeta(ABCMeta):
+class BaseMeta(type):
     """Base metaclass for many of the ibis core classes.
 
     This metaclass enforces the subclasses to define a `__slots__` attribute and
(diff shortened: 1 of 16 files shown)


ibis/common/bases.py
```diff
@@ -86,7 +121,7 @@ class Singleton(Base):
 
     @classmethod
     def __create__(cls, *args, **kwargs):
-        key = (cls, args, FrozenDict(kwargs))
+        key = (cls, args, tuple(kwargs.items()))
```
kszucs, line 124: `ibis.common.collections` is now depending on `ibis.common.bases` so `FrozenDict` would cause a circular import.

PR #7209, opened 2023-09-24
refactor(clickhouse): move `translate_val` to toposorted compiler loop

cpcloud:
This PR moves `translate_val` to function the same way that `translate_rel`
does: inputs are compiled and then passed into `translate_val` instead of
translation rules having to compile their own inputs.

A number interesting things come out of this:

1. Recursion is entirely removed from the compiler. This makes the logic it
   **much** easier to follow IMO.
1. Most rules are greatly simplified: no need to call `translate_val`
   recursively, inputs are compiled and then passed to functions.
1. A few features "just work" after this refactor because the translation rule
   for the op is simpler (window functions)

Additionally, I was able to remove most occurrences of `dialect="clickhouse"`
and convert the clickhouse translate rules to use sqlglot functions or or other
constructs (`SQLStringView` and `Join`s need the dialect parameter).

Not counting the newly generated SQL, the net lines of code are reduced as well:

```
❯ git diff --shortstat upstream/master HEAD~
 24 files changed, 975 insertions(+), 1258 deletions(-)
```


ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -17,79 +16,93 @@ def translate_rel(op: ops.TableNode, **_):
...
+def _aggregation(
+    op: ops.Aggregation, *, table, metrics, by, having, predicates, sort_keys, **_
+):
+    selections = (by + metrics) or STAR
     sel = sg.select(*selections).from_(table)
 
-    if group_keys := op.by:
-        sel = sel.group_by(*map(tr_val_no_alias, group_keys), dialect="clickhouse")
+    if by:
+        sel = sel.group_by(*range(1, len(by) + 1))
```
cpcloud, line 96: I think this should be reverted back to the original. The group by indexes are a newish feature and some of our users need the expression compiled in.

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -160,71 +161,81 @@ def _query(op: ops.SQLQueryResult, *, aliases, **_):
...
+            limit = sg.select(n).from_(table).subquery()
+        else:
+            limit = n
         result = result.limit(limit)
 
-    if not isinstance(offset := op.offset, int):
-        offset = f"(SELECT {translate_val(offset, **kw)} FROM {table})"
+    if not isinstance(offset, int):
+        return result.offset(
+            sg.select(offset).from_(table).subquery().sql("clickhouse")
```
cpcloud, line 192: I think the need for the dialect here is a bug in sqlglot

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -160,71 +161,81 @@ def _query(op: ops.SQLQueryResult, *, aliases, **_):
...
     try:
-        return table.where(predicate, dialect="clickhouse")
+        return table.where(predicate)
     except AttributeError:
-        return sg.select("*").from_(table).where(predicate, dialect="clickhouse")
+        return sg.select(STAR).from_(table).where(predicate)
+
+
+@translate_rel.register
+def _view(op: ops.View, *, child, name: str, **_):
```
cpcloud, line 231: I'll implement these in a follow up.

ibis/backends/clickhouse/compiler/relations.py
cpcloud: Fixed.

ibis/backends/clickhouse/compiler/relations.py
cpcloud: Created an issue https://github.com/tobymao/sqlglot/issues/2312

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -68,31 +91,52 @@ def translate(op: ops.TableNode, params: Mapping[ir.Value, Any]) -> sg.exp.Expre
...
+        >> (lambda op, _: ops.Literal(value=params[op], dtype=op.dtype))
+    )
+
+    # rewrite cumulative functions to window functions, so that we don't have
+    # to think about handling them in the compiler, we need only compile window
+    # functions
+    op = op.replace(
+        p.WindowFunction(p.Cumulative, ...)
+        >> (lambda op, _: cumulative_to_window(op.func, op.frame))
+    )
```
cpcloud, line 137: The pattern system is pretty darn sweet.

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -17,79 +16,94 @@ def translate_rel(op: ops.TableNode, **_):
...
-    tr_val = partial(translate_val, **kw)
-    tr_val_no_alias = partial(translate_val, render_aliases=False, **kw)
-
-    by = tuple(map(tr_val, op.by))
-    metrics = tuple(map(tr_val, op.metrics))
-    selections = (by + metrics) or "*"
+def _aggregation(
+    op: ops.Aggregation, *, table, metrics, by, having, predicates, sort_keys, **_
+):
+    selections = (by + metrics) or STAR
```
cpcloud, line 91: `STAR` should be `(STAR,)`

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -17,79 +16,94 @@ def translate_rel(op: ops.TableNode, **_):
...
+    # if we're only reprojecting table, return the table
+    if (
+        len(selections) == 1
+        and from_ is None
+        and join is None
+        and not predicates
+        and not sort_keys
+        and all(
+            isinstance(node, sg.exp.Column) and node.this == STAR for node in selections
+        )
```
cpcloud, line 68: Not a huge fan of this, but without it, queries like `table.*` reference the wrong alias.

Will investigate further.

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -109,20 +123,11 @@ def _aggregation(op: ops.Aggregation, *, table, **kw):
...
-                for predicate in predicates
-            ),
-        )
-    else:
-        on = None
+def _join(op: ops.Join, *, left, right, predicates, **_):
+    on = sg.and_(*predicates) if predicates else None
     join_type = _JOIN_TYPES[type(op)]
     try:
+        # dialect must be passed to allow clickhouse's ANY/ANY/ASOF joins
```
cpcloud, line 130: Remove the additional `ANY`

ibis/backends/clickhouse/compiler/relations.py
cpcloud: This has since been fixed upstream.

ibis/backends/clickhouse/compiler/relations.py
cpcloud: Only a single test fails when this code is deleted: `ibis/backends/clickhouse/tests/test_select.py::test_filter_predicates`

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -68,31 +91,52 @@ def translate(op: ops.TableNode, params: Mapping[ir.Value, Any]) -> sg.exp.Expre
...
+
+    # apply translate rules in topological order
+    results = op.map(fn, filter=(ops.TableNode, ops.Value))
```
kszucs, line 140: Restricting the traversal is required or just performance optimization?

ibis/backends/clickhouse/compiler/relations.py
cpcloud: The problem is that we need to replace the `.*` table with the one we're about to select from, in every case **except** selection from a join, because joins don't have a single alias to select from.

I found a slightly better solution.

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -45,14 +45,37 @@
...
+    func: ops.Cumulative, frame: ops.WindowFrame
+) -> ops.WindowFunction:
+    klass = getattr(ops, func.__class__.__name__.replace("Cumulative", ""))
+    new_op = klass(*func.args)
+    new_frame = frame.copy(start=None, end=0)
+    new_expr = an.windowize_function(new_op.to_expr(), frame=new_frame)
+    return new_expr.op()
+
+
+def _translate_node(node, *args, **kwargs):
```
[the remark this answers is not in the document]
cpcloud: I can do that in a follow up.

ibis/backends/clickhouse/compiler/core.py
cpcloud: I haven't tried it without. I'd rather keep it as is, if only for readability. It's easier to see that the code should only work with these node types than if there were no filter, which would suggest it handles generic `Node` instances.

ibis/backends/clickhouse/compiler/core.py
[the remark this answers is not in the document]
cpcloud: Actually I think this adds unnecessary implicit behavior:

1. I now have to deal with import cycles
2. I now have to ensure that both `values` and `relations` `translate` is imported so that all the rules are registered.

I don't really see what we gain by making these into a single rule.

ibis/backends/clickhouse/compiler/core.py
cpcloud: Let's avoid building that until we have some more use cases as input

cpcloud: Single-pass replacement worked 🎉

ibis/backends/clickhouse/compiler/relations.py
```diff
@@ -17,79 +16,94 @@ def translate_rel(op: ops.TableNode, **_):
     raise com.OperationNotDefinedError(f"No translation rule for {type(op)}")
 
 
-@translate_rel.register(ops.DummyTable)
-def _dummy(op: ops.DummyTable, **kw):
-    return sg.select(
-        *map(partial(translate_val, **kw), op.values), dialect="clickhouse"
-    )
+@translate_rel.register
+def _dummy(op: ops.DummyTable, *, values, **_):
```
kszucs, line 20: Can you remove the keyword-only marker `*` from the functions? I usually find it more distracting than useful.

ibis/backends/clickhouse/compiler/core.py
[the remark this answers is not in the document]
kszucs: Put all the translate implementations into a single file?

ibis/backends/clickhouse/compiler/core.py
kszucs: Note that this will work until the rules have distinct matchers because the first matching pattern gets applied.

ibis/backends/clickhouse/compiler/values.py
```diff
@@ -493,386 +409,290 @@ def _literal(op, **kw):
...
-        value = helpers.parenthesize(value)
-
-    # this will fail to execute if there's a correlation, but it's too
-    # annoying to detect so we let it through to enable the
-    # uncorrelated use case (pandas-style `.isin`)
-    subquery = translate(op.options.to_expr().as_table().op(), {})
-    return f"{value} IN ({_sql(subquery)})"
+def _in_column(op, *, value, options, cache, aliases, **_):
+    base = an.find_first_base_table(op.options)
+    # TODO: cache is only used in this one place in the entire compiler; is it
```
kszucs, line 686: I would love to see it removed. Can you do `aliases.get(base)` instead?

kszucs: If cache would be removed then it could be removed from `node.map()` entirely.

cpcloud: No.

Without this, the subquery doesn't get compiled because subquery is used here directly.

cpcloud: In general we currently cannot know whether the subquery on the right hand side is already present somewhere in the query. Just because it's already been compiled doesn't mean it's been `SELECT`-ed from.

ibis/backends/clickhouse/compiler/core.py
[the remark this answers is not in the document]
cpcloud: I like that they are separate now, it makes it easier to things IMO.

ibis/backends/clickhouse/compiler/values.py
cpcloud: Was able to remove this using a rewrite rule from `InColumn(left, right)` -> `InColumn(left, table_array_view(right))` in the compiler.

ibis/expr/analysis.py
```diff
@@ -793,3 +793,13 @@ def finder(node):
         )
 
     return g.traverse(finder, nodes, filter=ops.Node)
+
+
+def cumulative_to_window(
+    func: ops.Cumulative, frame: ops.WindowFrame
+) -> ops.WindowFunction:
+    klass = getattr(ops, func.__class__.__name__.replace("Cumulative", ""))
```
kszucs, line 801: It looks a bit fragile to me but fine for now.

ibis/backends/tests/test_window.py
```diff
@@ -952,9 +949,6 @@ def agg(df):
 
 
 @pytest.mark.notimpl(["datafusion", "polars"], raises=com.OperationNotDefinedError)
-@pytest.mark.broken(
-    ["clickhouse"], reason="clickhouse returns incorrect results", raises=AssertionError
```
kszucs, line 956: Nice!

ibis/backends/clickhouse/tests/test_functions.py
```diff
@@ -459,12 +460,38 @@ def test_count_distinct_with_filter(alltypes):
...
+    @udf.scalar.builtin(name="plus")
+    def my_add(a: int, b: int) -> int:
+        ...
+
+    n = 5
+    expr = (
+        alltypes[alltypes.int_col == 1]
+        .limit(n)
+        .int_col.collect()
+        .map(lambda x: my_add(x, 1))
```
kszucs, line 484: Pretty nice!

ibis/expr/analysis.py
cpcloud: We haven't changed the name of cumulative operations ever ... maybe?

ibis/backends/clickhouse/tests/conftest.py
```diff
@@ -120,20 +120,3 @@ def alltypes(con):
 @pytest.fixture(scope="session")
 def df(alltypes):
     return alltypes.execute()
-
-
-@pytest.fixture
-def translate():
```
kszucs, line 126: Nice to see this gone!

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -68,31 +82,80 @@ def translate(op: ops.TableNode, params: Mapping[ir.Value, Any]) -> sg.exp.Expre
...
+    # `translate_val` rule
+    params = {param.op(): value for param, value in params.items()}
+    replace_literals = p.ScalarParameter >> (
+        lambda op, _: ops.Literal(value=params[op], dtype=op.dtype)
+    )
+
+    # rewrite cumulative functions to window functions, so that we don't have
+    # to think about handling them in the compiler, we need only compile window
+    # functions
+    replace_cumulative_ops = p.WindowFunction(
```
kszucs, line 124: Like it!

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -68,31 +82,80 @@ def translate(op: ops.TableNode, params: Mapping[ir.Value, Any]) -> sg.exp.Expre
...
+    ) >> a.cumulative_to_window(x, y)
+
+    # replace the right side of InColumn into a scalar subquery for sql
+    # backends
+    replace_in_column_with_table_array_view = p.InColumn >> (
+        lambda op, _: op.__class__(
+            op.value,
+            ops.TableArrayView(
+                ops.Selection(
+                    table=an.find_first_base_table(op.options), selections=(op.options,)
```
kszucs, line 135: We should switch to `Topmost` pattern to locate the base table, but in a follow-up.

kszucs: Something like `BaseTableOf = Topmost(ops.Relation)`

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -68,31 +82,80 @@ def translate(op: ops.TableNode, params: Mapping[ir.Value, Any]) -> sg.exp.Expre
...
+    # `False`
+    replace_empty_in_values_with_false = p.InValues(x, ()) >> c.Literal(
+        False, dtype="bool"
+    )
+
+    replace_notexists_subquery_with_not_exists = p.NotExistsSubquery(x) >> c.Not(
+        c.ExistsSubquery(x)
+    )
+
+    op = op.replace(
```
kszucs, line 151: Starting to have a ruleset here, great to see it in action!

kszucs (approved):
Great improvement, thanks @cpcloud!

PR #7218, opened 2023-09-26
feat(patterns): support building sequences in replacement patterns

kszucs:
Support sequences in the right hand side of replacements:
```py
from ibis.common.patterns import Replace, Variable

a = Variable("a")
r = Replace(searcher, (1, 2, a))
```

Added a new decorator based syntax for defining replacement rules:

```py
from ibis.common.patterns import replace

@replace(searcher)
def substitute(value, context):
    # build replacement value
```

Miscellaneous changes:
 - renamed `Builder.make()` method to `Builder.build()`
 - use `Base.__create__()` instead of `__new__()`

### DIFF
--- a/ibis/common/patterns.py
+++ b/ibis/common/patterns.py
@@ -328,7 +328,7 @@ class Builder(Hashable):
         ...
 
     @abstractmethod
-    def make(self, context: dict):
+    def build(self, context: dict):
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
LGTM! Thanks!

PR #7222, opened 2023-09-26
feat(patterns): support building sequences in replacement patterns       

kszucs:
Reopened version of #7218 with a fix for infinite recursion in case of string inputs.

Support sequences in the right hand side of replacements:

```py
from ibis.common.patterns import Replace, Variable

a = Variable("a")
r = Replace(searcher, (1, 2, a))
```

Added a new decorator based syntax for defining replacement rules:

```py
from ibis.common.patterns import replace

@replace(searcher)
def substitute(value, context):
    # build replacement value
```

Miscellaneous changes:
 - renamed `Builder.make()` method to `Builder.build()`
 - use `Base.__create__()` instead of `__new__()`

### DIFF
--- a/ibis/common/patterns.py
+++ b/ibis/common/patterns.py
@@ -322,7 +322,7 @@ class Builder(Hashable):
         ...
 
     @abstractmethod
-    def make(self, context: dict):
+    def build(self, context: dict):
         """Construct a new object from the context.
 
         Parameters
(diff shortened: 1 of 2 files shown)


ibis/common/patterns.py
```diff
@@ -1753,6 +1782,39 @@ def FrozenDictOf(key_pattern, value_pattern):
...
+    if isinstance(obj, Builder):
+        # already a builder, no need to convert
+        return obj
+    elif callable(obj):
+        # the callable builds the substitution
+        return Factory(obj)
+    elif isinstance(obj, Mapping):
+        # allow nesting builder patterns in dicts
+        return Call(type(obj), **obj)
+    elif is_iterable(obj):
```
cpcloud, line 1810: This is handling the empty tuple/list/etc case?

cpcloud (approved):

ibis/common/patterns.py
kszucs: Yes, added tests to cover them.

PR #7232, opened 2023-09-27
feat(api): improve repr for deferred expressions containing Column/Scalar values

jcrist:
This makes `Column`/`Scalar` values found in a deferred expression repr as `<column[type]>`/`<scalar[type]>`. Previously they'd repr using their standard repr, which would lead to weird displays (especially in interactive mode).

Example:

```python
In [1]: import ibis

In [2]: from ibis import _

In [3]: t = ibis.table({"x": "int"})

In [4]: ibis.date(_.h, _.m, t.x.sum())
Out[4]: date(_.h, _.m, <scalar[int64]>)
```

### DIFF
--- a/ibis/expr/deferred.py
+++ b/ibis/expr/deferred.py
@@ -32,6 +32,17 @@ _UNARY_OPS: dict[str, Callable[[Any], Any]] = {
 }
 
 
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
🚢

PR #7235, opened 2023-09-27
feat(api): support deferred arguments to udfs

jcrist:
```python
In [1]: import ibis

In [2]: from ibis import _

In [3]: @ibis.udf.scalar.builtin
   ...: def myfunc(x: int) -> int:
   ...:     pass
   ...: 

In [4]: myfunc(_.x)
Out[4]: myfunc(_.x)
```


cpcloud (approved):
NICE

PR #7236, opened 2023-09-27
depr: deprecate `ibis.where` in favor of `ibis.ifelse`

jcrist:
- Deprecates `ibis.where` in favor of `ibis.ifelse`
- Renames `ops.Where` to `ops.IfElse` for consistency. The main way this affects users is in the auto-generated column names; I thought it looked weird to have a column named `Where(...)` when the user called `ifelse`. If needed _might_ want to add an alias so `ops.Where == ops.IfElse`, but I've opted a clean break for now.


cpcloud (approved):
Looks like a couple of small conflicts. Other than that, LGTM!

PR #7238, opened 2023-09-28
feat(api): support deferred and literal values in `ibis.ifelse`

jcrist:
- Adds support for passing literal and deferred values to `ibis.ifelse`
- Cleans up the docstring and type signature of `ifelse` a bit to be more accurate
- Adds tests to cover the new behavior


jcrist:
```python
In [1]: import ibis

In [2]: from ibis import _

In [3]: ibis.ifelse(False, 1, 2).execute()  # can now pass literals to `ifelse`
Out[3]: 2

In [4]: ibis.ifelse(_.cond, _.x, _.y)  # can now pass deferreds to ifelse
Out[4]: ifelse(_.cond, _.x, _.y)
```

PR #7240, opened 2023-09-28
feat(api): support deferred values in `ibis.coalesce`/`ibis.greatest`/`ibis.least`

jcrist:


jcrist:
```python
In [1]: import ibis

In [2]: from ibis import _

In [3]: t = ibis.table({"x": "int", "y": "int"})

In [4]: expr = ibis.least(_.x, 2)

In [5]: expr
Out[5]: least(_.x, 2)

In [6]: expr.resolve(t)
Out[6]: 
r0 := UnboundTable: unbound_table_0
  x int64
  y int64

Least(): Least([r0.x, 2])
```

ibis/expr/api.py
```diff
@@ -1831,9 +1831,84 @@ def where(cond, true_expr, false_expr) -> ir.Value:
...
+        Minimum of the passed arguments
+
+    Examples
+    --------
+    >>> import ibis
+    >>> ibis.options.interactive = True
+    >>> ibis.least(None, 4, 5)
+    4
+    """
+    if not args:
```
cpcloud, line 1907: We should be able to do this with `Annotated` and a `Length(at_least=1)` pattern on the op. Not a blocker.

cpcloud (approved):

ibis/expr/api.py
cpcloud: Went ahead and did this after I resolved the conflicts :)

PR #7242, opened 2023-09-28
test(dev-tools): allow skipping flink with pytest markers

cpcloud:
Allows use of `pytest -m "not pyflink"` to avoid collection failures.

### DIFF
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -3,9 +3,7 @@ from __future__ import annotations
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any
 
-import pyflink.version
 import sqlglot as sg
-from pyflink.table.types import create_arrow_schema
 
 import ibis.common.exceptions as exc
 import ibis.expr.operations as ops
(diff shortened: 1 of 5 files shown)


deepyaman (approved):
Basically, don't keep any top-level `pyflink` imports, and skip tests in case a dependency is missing? LGTM

cpcloud:
Yeah!

I think we can potentially bring some of these back once flink dependencies are up-to-date.

cpcloud:
> We don't have to avoid top-level imports of backend-specific libs in other backends, why is that needed here? I thought this was something that the `deps` specifiers ([`master`/ibis/backends/flink/tests/conftest.py#L14](https://github.com/ibis-project/ibis/blob/master/ibis/backends/flink/tests/conftest.py?rgh-link-date=2023-09-28T14%3A01%3A35Z#L14)) were supposed to help avoid?

`deps` are only checked in `load_data`, by that time the backend has probably been imported by `conftest`. It's possible that `deps` isn't useful and we should remove it.

This problem is only triggered in a specific case:

- a logically negated marker: `pytest -m 'not thing'`
- no test directory specified

Note that saying `-m 'not pyflink'` doesn't skip collection of `flink` stuff.

This comes up when I want to run all the backends except a few.

My particular issue is that I don't have pyflink deps installed because they are not yet specified in `pyproject.toml`, and I'd rather not install them only to skip their tests (I realize this is effectively what's happening with `pyproject.toml`, but there I don't have to manage anything out of band).

We could do this for all the backends if that makes this PR more palatable.

gforsyth:
We could add a `skip=` to the `pyproject.toml` and that would skip these tests for everyone -- and the folks working on the flink backend would have to remove that from the config

gforsyth (approved):

PR #7244, opened 2023-09-28
refactor: a few `ibis.expr.api` cleanups

jcrist:
- Remove some dead code
- Deprecate `ibis.negate` in favor of `-expr`/`expr.negate`. This was the last method that doesn't seem potentially useful as a top-level function.
- Reorg the file so all exported constants are at the top. Before they were intermingled with function definitions, making it easy to miss some dead code.
- Add a few missing type annotations to `ibis.union`/`ibis.intersect`/`ibis.difference`


cpcloud (approved):
LGTM

ibis/expr/api.py
```diff
@@ -171,6 +171,13 @@
 dtype = dt.dtype
 infer_dtype = dt.infer
 infer_schema = sch.infer
+aggregate = ir.Table.aggregate
+cross_join = ir.Table.cross_join
+join = ir.Table.join
+asof_join = ir.Table.asof_join
```
cpcloud, line 177: I'm a little dubious on these, but they're also maybe the way some people like to write joins.

jcrist: Yeah, I could go either way here. Since they're table-level operations I felt fine leaving them for now.

PR #7245, opened 2023-09-28
refactor(analysis): vastly simplify windowize_function

kszucs:
I have hacked a stop condition to the graph abstraction which I need to clean up first.

I also need to figure out both a proper API and implementation for controlling the traversals (currently `stop` limits the traversal and builds up a subgraph). 

depends on #7330

### DIFF
--- a/ibis/common/graph.py
+++ b/ibis/common/graph.py
@@ -145,9 +145,8 @@ class Node(Hashable):
             the results as the second and the results of the children as keyword
             arguments.
         filter
-            Pattern-like object to filter out nodes from the traversal. Essentially
-            the traversal will only visit nodes that match the given pattern and
-            stop otherwise.
+            Pattern-like object to filter out nodes from the traversal. The traversal
+            will only visit nodes that match the given pattern and stop otherwise.
 
         Returns
         -------
@@ -171,9 +170,8 @@ class Node(Hashable):
         type
             Type or tuple of types to find.
         filter
-            Pattern-like object to filter out nodes from the traversal. Essentially
-            the traversal will only visit nodes that match the given pattern and
-            stop otherwise.
+            Pattern-like object to filter out nodes from the traversal. The traversal
+            will only visit nodes that match the given pattern and stop otherwise.
 
         Returns
         -------
@@ -197,9 +195,8 @@ class Node(Hashable):
             Pattern to match. `ibis.common.pattern()` function is used to coerce the
             input value into a pattern. See the pattern module for more details.
         filter
-            Pattern-like object to filter out nodes from the traversal. Essentially
-            the traversal will only visit nodes that match the given pattern and
-            stop otherwise.
+            Pattern-like object to filter out nodes from the traversal. The traversal
+            will only visit nodes that match the given pattern and stop otherwise.
         context
             Optional context to use for the pattern matching.
 
@@ -288,9 +285,8 @@ class Graph(dict[Node, Sequence[Node]]):
         root
             Root node of the graph.
         filter
-            Pattern-like object to filter out nodes from the traversal. Essentially
-            the traversal will only visit nodes that match the given pattern and
-            stop otherwise.
+            Pattern-like object to filter out nodes from the traversal. The traversal
+            will only visit nodes that match the given pattern and stop otherwise.
 
         Returns
         -------
@@ -338,9 +334,8 @@ class Graph(dict[Node, Sequence[Node]]):
         root
             Root node of the graph.
         filter
-            Pattern-like object to filter out nodes from the traversal. Essentially
-            the traversal will only visit nodes that match the given pattern and
-            stop otherwise.
+            Pattern-like object to filter out nodes from the traversal. The traversal
+            will only visit nodes that match the given pattern and stop otherwise.
 
         Returns
         -------
@@ -441,9 +436,8 @@ def bfs(node: Node, filter: Optional[Any] = None) -> Graph:
     node
         Root node of the graph.
     filter
-        Pattern-like object to filter out nodes from the traversal. Essentially
-        the traversal will only visit nodes that match the given pattern and
-        stop otherwise.
+        Pattern-like object to filter out nodes from the traversal. The traversal
+        will only visit nodes that match the given pattern and stop otherwise.
 
     Returns
     -------
@@ -460,9 +454,8 @@ def dfs(node: Node, filter: Optional[Any] = None) -> Graph:
     node
         Root node of the graph.
     filter
-        Pattern-like object to filter out nodes from the traversal. Essentially
-        the traversal will only visit nodes that match the given pattern and
-        stop otherwise.
+        Pattern-like object to filter out nodes from the traversal. The traversal
+        will only visit nodes that match the given pattern and stop otherwise.
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


ibis/common/graph.py
```diff
@@ -327,13 +341,24 @@ def from_dfs(cls, root: Node, filter=Node) -> Self:
         if not isinstance(root, Node):
             raise TypeError("node must be an instance of ibis.common.graph.Node")
 
-        stack = deque([root])
+        stack = deque()
+        if stop is None:
```
kszucs, line 345: This is just ugly.

cpcloud:
@kszucs This is pretty close, just needs a couple arguments' documentation!

ibis/expr/analysis.py
```diff
@@ -336,48 +336,29 @@ def propagate_down_window(func: ops.Value, frame: ops.WindowFrame):
...
+    from ibis.common.patterns import Variable, _
+
+    func = Variable("func")
+    frame = Variable("frame")
+
+    node = expr.op()
+    res = node.replace(
+        (
+            p.WindowFunction(func, frame)
+            >> c.WindowFunction(
```
kszucs, line 349: This branch is only used from `GroupedTable` so it will better to remove this function entirely and use the specific replacement rules in `Projector` and `GroupedTable`.

kszucs: Well, we need both rewrites there as well.

ibis/common/graph.py
```diff
@@ -201,12 +193,14 @@ def match(
...
         ----------
-        pat : Any
+        pat
             Pattern to match. `ibis.common.pattern()` function is used to coerce the
             input value into a pattern. See the pattern module for more details.
-        filter : Optional[type], default None
-            Type to filter out for the traversal, Node is filtered out by default.
-        context : Optional[dict], default None
+        filter
+            Pattern-like object to filter out nodes from the traversal. Essentially
```
cpcloud, line 200: I think we can remove "Essentially"

ibis/common/graph.py
```diff
@@ -118,70 +128,52 @@ def __args__(self) -> tuple[Any, ...]:
...
 
         Parameters
         ----------
-        type : type | tuple[type]
+        type
             Type or tuple of types to find.
-        filter : Optional[type], default None
-            Type to filter out for the traversal, Node is filtered out by default.
+        filter
+            Pattern-like object to filter out nodes from the traversal. Essentially
```
cpcloud, line 174: I think we can remove "Essentially"

ibis/common/graph.py
```diff
@@ -284,17 +278,19 @@ def __init__(self, mapping=(), /, **kwargs):
...
 
         Parameters
         ----------
-        root : Node
+        root
             Root node of the graph.
-        filter : Optional[type], default None
-            Type to filter out for the traversal, Node is filtered out by default.
+        filter
+            Pattern-like object to filter out nodes from the traversal. Essentially
```
cpcloud, line 291: I think we can remove "Essentially"

ibis/common/graph.py
```diff
@@ -399,34 +433,42 @@ def toposort(self) -> Self:
...
+def dfs(node: Node, filter: Optional[Any] = None) -> Graph:
     """Construct a graph from a root node using a depth-first search.
 
     Parameters
     ----------
-    node : Node
+    node
         Root node of the graph.
+    filter
+        Pattern-like object to filter out nodes from the traversal. Essentially
```
cpcloud, line 463: Ok, maybe just remove all the "Essentially"s :)

ibis/expr/types/groupby.py
```diff
@@ -250,33 +246,24 @@ def _selectables(self, *exprs, **kwexprs):
...
-            return ops.RowsWindowFrame(
-                table=self.table,
-                group_by=self.by,
-                order_by=bind_expr(self.table, self._order_by),
-            )
-        else:
-            return self._window.copy(
-                groupy_by=bind_expr(self.table, self._window.group_by + self.by),
-                order_by=bind_expr(self.table, self._window.order_by + self._order_by),
-            )
```
cpcloud, line 278: Nice, this was pretty gross.

ibis/expr/analysis.py
```diff
@@ -338,48 +338,24 @@ def propagate_down_window(func: ops.Value, frame: ops.WindowFrame):
...
-        for arg in op.args:
-            if isinstance(arg, ops.Value):
-                arg = _windowize(arg, frame)
-            elif isinstance(arg, tuple):
-                arg = tuple(_windowize(x, frame) for x in arg)
+def windowize_function(expr, default_frame):
+    func = var("func")
+    frame = var("frame")
 
-            windowed_args.append(arg)
```
cpcloud, line 378: Huzzah! 🎉 

This is waaaaayyyy better.

ibis/common/tests/test_graph_benchmarks.py
```diff
@@ -1,19 +1,49 @@
...
+        "2",
+        c=(3, 4),
+        d=frozendict(e=5, f=6),
+        e=generate_node(0),
+        f=(generate_node(depth - 1), generate_node(0)),
+    )
+
+
+def test_generate_node():
+    for depth in [0, 1, 2, 10, 100]:
```
cpcloud, line 36: Not a huge fan of for loops in tests. It makes any specific failures difficult to debug IMO. Can you make this `parametrize`d?

cpcloud: Alternatively, `0`, `1`, `10` seems fine.

cpcloud (approved):
Few comments, nothing blocking, though I would like those "Essentially"s gone :)

ibis/common/graph.py
kszucs: Rebase issue.

ibis/common/graph.py
```diff
@@ -303,28 +299,48 @@ def from_bfs(cls, root: Node, filter=Node) -> Self:
...
 
         Parameters
         ----------
-        root : Node
+        root
             Root node of the graph.
-        filter : Optional[type], default None
-            Type to filter out for the traversal, Node is filtered out by default.
+        filter
+            Pattern-like object to filter out nodes from the traversal. Essentially
```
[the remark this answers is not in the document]
kszucs: Removed.

ibis/common/graph.py
kszucs: Removed.

ibis/common/graph.py
kszucs: Removed.

ibis/common/graph.py
kszucs: Removed.

ibis/expr/types/groupby.py
kszucs: The whole GroupedTable object must be refactor, subject to a follow-up.

ibis/expr/analysis.py
kszucs: I'm still not satisfied with it. The issue is that we push things through Projector several times and it makes the logic hardly trackable. For example the window merging only happens from the GroupedTable projections, but I should either call `wrap_analytic` rule before passing to Projector with the grouped frame or keep it as is and pass the desired frame as an argument.

Anyway, once we clean up the projector we may be able to further simplify this.

PR #7262, opened 2023-09-29
docs(reference): improve signatures

NickCrews:
Followup to https://github.com/ibis-project/ibis/pull/7159

Make the signatures appear show the full path,
eg `random()` is now `ibis.random()`.

Also fix some of the paths where possible,
eg now it is `ibis.connect()` instead of `backends.base.connect()`

@machow, note how quartodoc was silently ignoring the previously-incorrectly-configured `signature_name` option. It would have been nice if the build had failed here, instead of us thinking that we
had configured something.

See the difference in `ibis.connect()`:
image

and note how in the sidebar you just see "random", not "ibis.random"
image

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -200,7 +200,7 @@ quartodoc:
             - name: Table
               package: ibis.expr.types.relations
             - name: GroupedTable
-              package: ibis.expr.types.relations
+              package: ibis.expr.types.groupby
             - name: read_csv
               dynamic: true
             - name: read_delta
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -185,7 +185,7 @@ quartodoc:
   render_interlinks: true
   renderer: _renderer.py
   options:
-    signature_name: full
+    signature_name: short
   sections:
     - title: Expression API
       desc: "APIs for manipulating table, column and scalar expressions"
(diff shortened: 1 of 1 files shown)


cpcloud:
Not really a huge fan of showing the full module path for objects:

image

For top-level functions I think we **should** show the object, but for everything else the module where the object resides is both 1) most irrelevant for end-users, and 2) subject to change at any time and not part of the public API in any way.

cpcloud:
I'm looking into ways to render the longer name selectively

cpcloud:
Alright, I was able to specify `signature_name: full` for all the top level functions.

Unfortunately I **can't** specify `signature_name: full` at the top of the config and then specify `signature_name: short` for the objects I want to have short signatures, because that option doesn't cascade to auto-discovered methods.

NickCrews:
This is better, thank you!

cpcloud (approved):

machow:
Sorry, just getting to this. If I'm understanding the original docs did this:

* Set default member_options, so that methods on the automatically discovered classes (e.g. `- TimestampValue` in contents) showed short signatures.
* Everything else used the default behavior of showing whatever is specified as `name: `.

I wonder if this strategy would work?

* set `package: null` (instead of ibis) in the top-level quartodoc config
* set member_options with `signature_name: short` again
* replace entries that used `package: ibis` to just include ibis in the `name: `

So something like...

```yaml
# 
quartodoc:
  package: ibis
  sections:
    ...
            - name: read_delta
              dynamic: true
```

becomes

```yaml
quartodoc:
  package: null
  sections:
    ...
            - name: ibis.read_delta
              dynamic: true
```

It seems like getting the config right here is tricky, because one set of things (e.g. GroupedTable) are displayed in a way that is less centered on how you might import them (i.e. if you see a GroupedTable, here are the methods you might be able to use), while the others need to make clear where they can be imported from (e.g. you can't `import ibis.GroupedTable`, but you probably want people to realize right away they can import `ibis.read_json`)

PR #7263, opened 2023-09-29
depr(api): replace `nullifzero` with `ifnull(0)` and `zeroifnull` with `fillna(0)`

NickCrews:
WIP!

I didn't bother clearing out all the implementation, but just
curious if this will pass the tests.


cpcloud:
I can help with this. Looks like the failures are not too complex to address.

NickCrews:
Thanks @cpcloud this looks great! Nothing makes me happier than deleting code!

ibis/expr/types/numeric.py
```diff
@@ -480,13 +481,15 @@ def sqrt(self) -> NumericValue:
         """
         return ops.Sqrt(self).to_expr()
 
+    @util.deprecated(instead="use nullif(0)", as_of="7.0", removed_in="8.0")
     def nullifzero(self) -> NumericValue:
         """Return `NULL` if an expression is zero."""
-        return ops.NullIfZero(self).to_expr()
+        return self.nullif(ops.Literal(0, dtype=self.type()).to_expr())
```
NickCrews, line 487: I can see why we need to be careful with casting the 0 with fillna(0) vs fillna(0.0) so you end up with the right type.in the result, but here we are just testing for equality so zeros of any type will still be equal, right? Eg ibis.literal(0).nullif(0.0) will still work? As expected?

If so, I would say remove this extra cast.

cpcloud: It needs to stay to propagate the type information in case a backend ever needs to insert an actual cast.

And also the use of ops is to avoid having to import Ibis.

cpcloud: There won't be a cast inserted unless the backend does it specifically. The literal construction is merely propagating type information.

NickCrews: Sorry if I'm missing something here. Or maybe I wasn't clear enough, I didn't just mean, "remove the cast", I meant the entire implementation should just be simply `return self.nullif(0)`. If we can't, that implies the migration path that we suggest to users is flawed. Maybe can you give an example of why `return self.nullif(0)` wouldn't work? Thanks!

cpcloud: I don't think I follow why this implementation detail would change the behavior of the result and change any equivalence between what was previously happening. It may not be necessary, but if it's implemented this way, I don't have to worry about whether all databases' notion of equality match python's

NickCrews: Ah, it wouldn't change in relation to previous behavior. I'm saying that if we are suggesting users to switch from `zeroifnull()` to `fillna(0)`, then we want those two to be equivalent. If we switch the implementation here to just be a wrapper around `fillna(0)`, and we don't change the .zeroifnull() tests at all (except for adding the pytest.warns...), then we are actively testing for that equivalence. If we make the implementation here slightly different, then the behavior between `.fillna(0)` and `zeroifnull()` could be slightly different, our tests won't catch it, users will switch over to `.fillna(0)`, and their results will change. We WANT our tests to fail and to know if databases' have different notions of equality, because then that means that the migration path from `zeroifnull()` to `.fillna(0)` is trickier than we thought, and we need to add in something extra.

NickCrews: In fact, per that argument, the implementation for `nullifzero()` should also be simply `nullif(0)`, to verify that clean migration for that method.

cpcloud: Ok, I see what you're saying! Makes sense. It's possible that `fillna` of `ifnull` may need to be modified to make the recommendation work :)

ibis/tests/expr/test_sql_builtins.py
```diff
@@ -83,10 +83,11 @@ def test_group_concat(functional_alltypes):
 
 
 def test_zeroifnull(functional_alltypes):
-    dresult = functional_alltypes.double_col.zeroifnull()
-    iresult = functional_alltypes.int_col.zeroifnull()
+    with pytest.warns(FutureWarning):
```
cpcloud, line 86: This should be two `warns` context managers.

cpcloud: Fixed.

cpcloud (approved):

cpcloud:
Clouds are passing:

```
cloud in 🌐 falcon in …/ibis on  deprecate-nullifzero-zeroifnull is 📦 v6.1.0 via 🐍 v3.10.12 via ❄️  impure (ibis-3.10.12-env) took 4s
❯ pytest -m 'bigquery or snowflake' -n 16 --dist loadgroup --snapshot-update -q
bringing up nodes...
.....xxxxxx....x....s.x.........xx..xxx...xxxx..xxx..xxxxx...xx.x..xxxxx..x.........x.xx.....x....xx.xx..xx.......x....x........x.....x.x.x.xx........................x....................x.. [  6%]
......xx..............x.....x......xx..x........xxxx........x.............x.s.x.x.x.x..x..x.....x.x...x.....x.......x..x.xx.x.....x............x..x.x..x.........xx..........x.x..x.x..x...... [ 13%]
..........x.............x.....x.................x.....................................................................x..........x..x..............x............................x.....x....... [ 20%]
.....x.......x.x.........x............s.s..x.x..x.s.s...x...............x....x.x..x....xxx..xxxx.x.xx...x.xx..xxx.x.x...x.xx..xx..xx..xx.....x.................x..xx..x..x.x.x..x..x...x..x... [ 26%]
........x.......x.....x..xx.....x...x...x..x.x.....xx....................x...........................x.x.....x..x........x..xx....x..x..x....xx...x.x...xx..x.x..x..x.xxx......x.......xx...x. [ 33%]
.....x...........xx.....x..x..........x..x.....x....x........x.............x............x...x.....x....x..x........x........x....x.................x......x....x..x..x......xx...........x..xx [ 40%]
.................x.......x.s..s...xsxx.xxsxxxxxxxxxx.xxxxxxxxxxxx.xx..x.s...xx.x.....xss.xx.xx.x...sx.s..xx..s..x........x.........x........x..xx....x..xxxx.sx........x..x.x..............s.. [ 47%]
.............s.s............s.................................xx...x.x...........xx..x.x.......x..ss.x.........x.xx..x..x..xx....x...xx.x....s..............x.....x.x......................... [ 53%]
...................................................s...s.s..s..xsx.xs.s......x........s....................................................................................................... [ 60%]
............s...s.s........xxx..s...x..........x...xx......x.........x....x....x...x.......x....x...x.x...x.x.........x.x...x.....x............x......xxx.x.............x.....x....x...xxx.xx. [ 67%]
xxxxxxxxxxxxx.xxxxxx.xx.xxxxx....x.x..x................xxx..x.........x...x......x..............................x.......x...x..x.x...x............x...x..x...x..x.......xx...x.......x........ [ 74%]
..x..x.....x.x..x.x........x..............x.......x.....x.....xx.......x......x.x..xx......x...........x...x.........x................x.....................................x................. [ 80%]
..............x................................................................................x...................x.......x..........x.................................x..................... [ 87%]
..............x.................x....x.x.x.....x.................................x...s..........................x...x......................................................................... [ 94%]
........................................................................................................x..............x..............x........................s...                            [100%]
2347 passed, 38 skipped, 438 xfailed in 443.28s (0:07:23)
```

jcrist:
Commit message is wrong I think? Isn't this replacing `nullifzero` with `nullif(0)` not `ifnull(0)`? Not a big deal, but we may want to manually tweak the changelog after release since this deprecation is user facing.

cpcloud:
That just landed, so I will force push a fix.

cpcloud:
Oh wait, nevermind :)

NickCrews:
Brilliant, thanks for tag teaming this @cpcloud !

PR #7265, opened 2023-09-29
feat(flink): add temporal operators

mfatihaktas:
# Summary

Part of the effort to make the common unit tests work for `flink` backend.

This PR adds new operators in `ibis/backends/flink/registry.py` to make the tests in `ibis/ibis/backends/tests/test_temporal.py` work with `pytest -m flink ...`.

### DIFF
--- a/ibis/backends/flink/registry.py
+++ b/ibis/backends/flink/registry.py
@@ -39,6 +39,31 @@ def _extract_field(sql_attr: str) -> str:
     return extract_field_formatter
 
 
+def _cast(translator: ExprTranslator, op: ops.generic.Cast) -> str:
+    if op.to.is_timestamp() and op.to.timezone:
+        arg_translated = translator.translate(op.arg)
+        return f"TO_TIMESTAMP(CONVERT_TZ(CAST({arg_translated} AS STRING), 'UTC+0', '{op.to.timezone}'))"
+
+    from ibis.backends.base.sql.registry.main import cast
+
+    return cast(translator=translator, op=op)
+
+
+def _left_op_right(translator: ExprTranslator, op_node: ops.Node, op_sign: str) -> str:
+    """Utility to be used in operators that perform '{op.left} {op_sign} {op.right}'."""
+    return f"{translator.translate(op_node.left)} {op_sign} {translator.translate(op_node.right)}"
+
+
+def _interval_add(translator: ExprTranslator, op: ops.temporal.IntervalSubtract) -> str:
+    return _left_op_right(translator=translator, op_node=op, op_sign="+")
+
+
+def _interval_subtract(
+    translator: ExprTranslator, op: ops.temporal.IntervalSubtract
+) -> str:
+    return _left_op_right(translator=translator, op_node=op, op_sign="-")
+
+
 def _literal(translator: ExprTranslator, op: ops.Literal) -> str:
     from ibis.backends.flink.utils import translate_literal
 
@@ -76,10 +101,21 @@ def _filter(translator: ExprTranslator, op: ops.Node) -> str:
     return f"CASE WHEN {bool_expr} THEN {true_expr} ELSE {false_null_expr} END"
 
 
-def _timestamp_diff(translator: ExprTranslator, op: ops.Node) -> str:
-    left = translator.translate(op.left)
-    right = translator.translate(op.right)
-    return f"timestampdiff(second, {left}, {right})"
+def _timestamp_add(translator: ExprTranslator, op: ops.temporal.TimestampAdd) -> str:
+    return _left_op_right(translator=translator, op_node=op, op_sign="+")
+
+
+def _timestamp_diff(translator: ExprTranslator, op: ops.temporal.TimestampDiff) -> str:
+    return _left_op_right(translator=translator, op_node=op, op_sign="-")
+
+
+def _timestamp_sub(translator: ExprTranslator, op: ops.temporal.TimestampSub) -> str:
+    table_column = op.left
+    interval = op.right
+
+    table_column_translated = translator.translate(table_column)
+    interval_translated = translator.translate(interval)
+    return f"{table_column_translated} - {interval_translated}"
 
 
 def _timestamp_from_unix(translator: ExprTranslator, op: ops.Node) -> str:
@@ -94,6 +130,17 @@ def _timestamp_from_unix(translator: ExprTranslator, op: ops.Node) -> str:
         raise ValueError(f"{unit!r} unit is not supported!")
 
     return f"TO_TIMESTAMP_LTZ({numeric}, {precision})"
+
+
+def _timestamp_from_ymdhms(
+    translator: ExprTranslator, op: ops.temporal.TimestampFromYMDHMS
+) -> str:
+    year, month, day, hours, minutes, seconds = (
+        f"CAST({translator.translate(e)} AS STRING)"
+        for e in [op.year, op.month, op.day, op.hours, op.minutes, op.seconds]
+    )
+    concat_string = f"CONCAT({year}, '-', {month}, '-', {day}, ' ', {hours}, ':', {minutes}, ':', {seconds})"
+    return f"CAST({concat_string} AS TIMESTAMP)"
 
 
 def _format_window_start(translator: ExprTranslator, boundary):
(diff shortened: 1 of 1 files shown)


ibis/backends/flink/registry.py
```diff
@@ -34,6 +34,35 @@ def extract_field_formatter(translator: ExprTranslator, op: ops.Node) -> str:
...
+    return f"{interval_left_translated} + {interval_right_translated}"
+
+
+def _interval_subtract(
+    translator: ExprTranslator, op: ops.temporal.IntervalSubtract
+) -> str:
+    interval_left, interval_right = op.left, op.right
+    interval_left_translated = translator.translate(interval_left)
+    interval_right_translated = translator.translate(interval_right)
+    return f"{interval_left_translated} - {interval_right_translated}"
```
deepyaman, line 63: There's a lot of duplication for `_interval_add`, `_interval_subtract`, `_timestamp_add`, etc.; they could all basically be one function-creating function, that takes the operator as an argument (because you're just doing `return f"{translator.translate(op.left)} {op} {translator.translate(op.right)}"` for all of them.

ibis/backends/flink/registry.py
```diff
@@ -34,6 +34,35 @@ def extract_field_formatter(translator: ExprTranslator, op: ops.Node) -> str:
     return extract_field_formatter
 
 
+def _cast(translator: ExprTranslator, op: ops.generic.Cast) -> str:
+    from ibis.expr.datatypes.core import Timestamp
+
+    arg, to = op.arg, op.to
+    if isinstance(to, Timestamp) and to.timezone:
+        arg_translated = translator.translate(arg)
+        return f"CONVERT_TZ(CAST({arg_translated} AS STRING), 'UTC+0', '{to.timezone}')"
```
deepyaman, line 43: Doesn't this return a string in the end? Should it return a timestamp?

ibis/backends/flink/registry.py
```diff
@@ -207,6 +267,58 @@ def _floor_divide(translator: ExprTranslator, op: ops.Node) -> str:
...
+
+    table_column_translated = translator.translate(table_column)
+    index_translated = translator.translate(index)
+
+    return f"{table_column_translated} '[' {index_translated} ']'"
+
+
+def _day_of_week_index(translator: ExprTranslator, op: ops.Node) -> str:
+    arg = translator.translate(op.args[0])
+    return f"MOD(DAYOFWEEK({arg}) + 5, 7)"
```
deepyaman, line 282: Nice.

Nit: Nothing to change, but I just wonder why it wouldn't make sense to use this as the default SQL translation for this operation. Adding 5 just seems like it has to always work, compared to -2 that can result in complications with negative numbers.

ibis/backends/flink/registry.py
```diff
@@ -207,6 +267,58 @@ def _floor_divide(translator: ExprTranslator, op: ops.Node) -> str:
...
+def _date_sub(translator: ExprTranslator, op: ops.temporal.DateSub) -> str:
+    date, interval = op.left, op.right
+    date_translated = translator.translate(date)
+    interval_translated = translator.translate(interval)
+    return f"{date_translated} - {interval_translated}"
+
+
+def extract_epoch_seconds(translator: ExprTranslator, op: ops.Node) -> str:
+    arg = translator.translate(op.arg)
+    return f"UNIX_TIMESTAMP(CAST({arg} AS STRING))"
```
deepyaman, line 311: Nice; this translation makes sense (think there was something non-working before had removed it in between).

ibis/backends/flink/registry.py
```diff
@@ -34,6 +34,35 @@ def extract_field_formatter(translator: ExprTranslator, op: ops.Node) -> str:
     return extract_field_formatter
 
 
+def _cast(translator: ExprTranslator, op: ops.generic.Cast) -> str:
```
deepyaman, line 37:
```suggestion
def _cast(translator: ExprTranslator, op: ops.Cast) -> str:
```
Nit: I feel like, in other parts of the codebase, I usually just see the non-namespaced version. But may be something better to check with one of the maintainers, what the best practice here is. (I'd also be curious, since there are inconsistencies.)

ibis/backends/flink/registry.py
mfatihaktas: Thanks, great catch! Wrapped it around with `TO_TIMESTAMP()`.

ibis/backends/flink/registry.py
mfatihaktas: > Adding 5 just seems like it has to always work, compared to -2 that can result in complications with negative numbers.

I agree. Let's enumerate all possibilities.

> DAYOFWEEK(date) :  Returns the day of a week (an integer between 1 and 7) from SQL date. Equivalent to EXTRACT(DOW FROM date). E.g., DAYOFWEEK(DATE ‘1994-09-27’) returns 3.

**Note**: Sunday is taken as the first day of the week.

> MOD(numeric1, numeric2): Returns the remainder (modulus) of numeric1 divided by numeric2. The result is negative only if numeric1 is negative.

* When `Day = Sunday`, `MOD(DAYOFWEEK({arg}) + 5, 7) = MOD(1 + 5, 7) = 6`.
* When `Day = Monday`, `MOD(DAYOFWEEK({arg}) + 5, 7) = MOD(2 + 5, 7) = 0`.
...
* When `Day = Friday`, `MOD(DAYOFWEEK({arg}) + 5, 7) = MOD(6 + 5, 7) = 4`.
* When `Day = Saturday`, `MOD(DAYOFWEEK({arg}) + 5, 7) = MOD(6 + 6, 7) = 5`.

So, adding `5` works as expected.

mfatihaktas:
> @mfatihaktas Is this PR ready for review/merge?

@cpcloud Yes.

ibis/backends/flink/registry.py
```diff
@@ -34,6 +34,34 @@ def extract_field_formatter(translator: ExprTranslator, op: ops.Node) -> str:
...
+    from ibis.expr.datatypes.core import Timestamp
+
+    arg, to = op.arg, op.to
+    if isinstance(to, Timestamp) and to.timezone:
```
cpcloud, line 41:
```suggestion
    arg, to = op.arg, op.to
    if to.is_timestamp() and to.timezone:
```

ibis/backends/flink/registry.py
```diff
@@ -219,6 +268,52 @@ def _floor_divide(translator: ExprTranslator, op: ops.Node) -> str:
...
 
 
+def _array_index(translator: ExprTranslator, op: ops.arrays.ArrayIndex):
+    table_column = op.arg
+    index = op.index
+
+    table_column_translated = translator.translate(table_column)
+    index_translated = translator.translate(index)
+
+    return f"{table_column_translated} '[' {index_translated} ']'"
```
cpcloud, line 278: Can you point to the place where this syntax is defined? It looks a little strange to my eye, like an array column would render as `a['0']` or `a['expr']` in the case of a more complex `expr`. Is that correct?

ibis/backends/flink/registry.py
```diff
@@ -219,6 +268,52 @@ def _floor_divide(translator: ExprTranslator, op: ops.Node) -> str:
...
+
+
+def _day_of_week_index(translator: ExprTranslator, op: ops.Node) -> str:
+    arg = translator.translate(op.args[0])
```
cpcloud, line 282: Prefer `op.arg` (and in general accessing the argument by name) over accessing `op.args[index]`

ibis/backends/flink/registry.py
```diff
@@ -91,6 +130,16 @@ def _timestamp_from_unix(translator: ExprTranslator, op: ops.Node) -> str:
     return f"TO_TIMESTAMP_LTZ({numeric}, {precision})"
 
 
+def _timestamp_from_ymdhms(
+    translator: ExprTranslator, op: ops.temporal.TimestampFromYMDHMS
+) -> str:
+    year, month, day = op.year.value, op.month.value, op.day.value
+    hours, minutes, seconds = op.hours.value, op.minutes.value, op.seconds.value
```
cpcloud, line 137: These two lines assume that all of the inputs are literals. Is that a requirement? If not, then you should call `translate` on each one and use the database's string concat functionality to combine the parts.

cpcloud (changes requested):
Looking good! A few questions and comments.

ibis/backends/flink/registry.py
mfatihaktas: > Can you point to the place where this syntax is defined?

It is defined [here](https://nightlies.apache.org/flink/flink-docs-release-1.13/docs/dev/table/functions/systemfunctions/#collection-functions) as `array '[' INT ']'`.

> It looks a little strange to my eye

I agree :)

> like an array column would render as a['0'] or a['expr'] in the case of a more complex expr.

It would look like `array'['0']'`. It is only defined for an integer index.

ibis/backends/flink/registry.py
mfatihaktas: Thanks, changed to `op.arg`.

ibis/backends/flink/registry.py
mfatihaktas: Thanks, fixed the function according to your suggestion, please take a look again.

ibis/backends/flink/registry.py
mfatihaktas: Thanks, applied your suggestion.

ibis/backends/flink/registry.py
cpcloud: Ah, got it.

I read the docs for the operation and I think you need to add ` + 1` to the index, because arrays are zero-indexed in ibis.

ibis/backends/flink/registry.py
```diff
@@ -96,6 +132,29 @@ def _timestamp_from_unix(translator: ExprTranslator, op: ops.Node) -> str:
...
+        )
+        + ", ' ', "
+        + ", ':', ".join(
+            [
+                f"CAST({translator.translate(e)} AS STRING)"
+                for e in [op.hours, op.minutes, op.seconds]
+            ]
+        )
+        + ")"
+    )
```
cpcloud, line 154: All the punctuation here is a bit challenging to read :)

How about this?

```suggestion
        year, month, day = [
            f"CAST({translator.translate(e)} AS STRING)"
            for e in [op.year, op.month, op.day]
        ]
        hours, minutes, seconds = [
            f"CAST({translator.translate(e)} AS STRING)"
            for e in [op.hours, op.minutes, op.seconds]
        ]
        date_string = f"CONCAT({year}, '-', {month}, '-', {day}, ' ', {hours}, ':', {minutes}, ':', {seconds})"
```

cpcloud (changes requested):
Two remaining issues, then this is good to go!

ibis/backends/flink/registry.py
cpcloud: I'm pretty sure these embedded quotes are incorrect. I understand that the Flink documentation uses this notation.

However, this notation is often how parser-generator grammars for this operation are written, and I suspect that `'['` was copy-pasted over directly from the parser and that

```
x'['0']'
```

is definitely not a valid Flink SQL expression.

@mfatihaktas @deepyaman @chloeh13q Can someone validate this?

cpcloud: https://docs.confluent.io/cloud/current/flink/reference/functions/collection-functions.html#flink-sql-array-collection is an example of a place where an example uses `array_name[index]`.

mfatihaktas: Thanks, you are right, this caused an error and I fixed it now. The problem earlier was that the test function(s) that I thought were validating this operator were actually not covering the operator.

ibis/backends/flink/registry.py
mfatihaktas: Thanks! I applied your suggestion.

cpcloud (approved):

PR #7271, opened 2023-10-01
refactor(clickhouse): use more sqlglot constructs

cpcloud:
Add a few more sqlglot conveniences for codegen and default to the MergeTree engine for file reading.

### DIFF
--- a/ibis/backends/base/sqlglot/__init__.py
+++ b/ibis/backends/base/sqlglot/__init__.py
@@ -52,11 +52,30 @@ class FuncGen:
     def concat(self, *args):
         return sg.exp.Concat.from_arg_list(list(map(_to_sqlglot, args)))
 
+    def map(self, keys, values):
+        return sg.exp.Map(keys=keys, values=values)
+
+
+class ColGen:
+    __slots__ = ()
+
+    def __getattr__(self, name: str) -> sg.exp.Column:
+        return sg.column(name)
+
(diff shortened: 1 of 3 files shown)


ibis/backends/base/sqlglot/__init__.py
```diff
@@ -52,11 +52,30 @@ def exists(self, query):
...
 def lit(val):
     return sg.exp.Literal(this=str(val), is_string=isinstance(val, str))
 
 
+def interval(value, *, unit):
+    return sg.exp.Interval(this=_to_sqlglot(value), unit=sg.exp.var(unit))
+
+
+f = FuncGen()
+c = ColGen()
```
jcrist, line 78: Single letter lowercase names intended to be imported by other modules feels weird to me. How about calling these `FUNCS` and `COLS` instead? Or `F` and `C` at least?

cpcloud: `F` and `C` sounds good to me. Definitely want to keep them easy to type.

cpcloud: Fixed!

jcrist (approved):

jcrist:
Looks like you have a small conflict, otherwise LGTM!

cpcloud:
I'm going to revert the `sign` change in case anyone is using ibis with an older clickhouse version.

PR #7273, opened 2023-10-02
feat(api): add `.delta` method for computing difference in units between two temporal values

### LINKED ISSUE #7195: feat(api): add API for computing the difference between two temporal objects
We current have an implementation of the `-` operator for `timestamp`, `date`
and `time`.

These operators are hard to use, because their return type is fixed: you can
only ever extract the delta in whatever unit we've hardcoded for the user
(units of seconds).

We don't have an API to compute an integer of the number of any other units
between two temporal expressions.

Let's create one!

Proposed API:

```python
class TemporalValue:
    def delta(self, other: TemporalValue, unit: str | IntervalUnit) -> IntegerValue:
        """Return the delta between this value and the other, in the given unit."""
```

cpcloud:
This PR adds a new API to temporal types, `delta`, that allows users to compute
the difference between two temporal types (times, dates, and timestamps).

This implemented for all the backends for which it was relatively
straightforward to get matching behavior out of.

Backends differ in many annoying and meaningful ways in how they implement this
operation:

1. Naming: `datediff` (MS SQL), `date_diff` (DuckDB)
1. Whether the inputs work on all temporal types or there's a separate function
   for each type: `date_diff` (DuckDB, MS SQL, etc) vs `time_diff`,
   `date_diff`, timestamp_diff (BigQuery)
1. **The meaning of the order of operands**. (ClickHouse follows subtraction
   order, DuckDB subtracts the first argument from the second argument)
1. Whether the unit is a string or magic identifier `'day'` (ClickHouse) vs
   `DAY` (MySQL)
1. **Whether the computation truncates the inputs and then computes the delta,
   or truncates afterward**. (Trino and MySQL truncate after, others before)

Tests hit each of these cases.

Backends that support one or more of these operations:

1. BigQuery
1. ClickHouse
1. DuckDB
1. MS SQL
1. MySQL
1. Snowflake
1. Trino

Notably, postgres and oracle do not have functions for computing the delta between two temporal types in integer units.

- Postgres only has functions for computing the delta that return intervals
- Oracle has [this blog post](https://blogs.oracle.com/sql/post/difference-between-datetimes-in-days-hours-minutes-seconds-in-oracle-database) showing how to use advanced database features to implement support for this operation.

Closes #7195


ibis/backends/clickhouse/compiler/values.py
```diff
@@ -1063,3 +1063,9 @@ def _scalar_udf(op, **kw) -> str:
 @translate_val.register(ops.AggUDF)
 def _agg_udf(op, *, where, **kw) -> str:
     return agg[op.__full_name__](*kw.values(), where=where)
+
+
+@translate_val.register(ops.DateDelta)
+@translate_val.register(ops.TimestampDelta)
```
cpcloud, line 1069: ClickHouse doesn't have a time type. There's probably a way to simulate the `TimeDelta` operation, but that can be done in a follow up.

ibis/backends/mysql/registry.py
```diff
@@ -242,6 +255,8 @@ def compiles_mysql_trim(element, compiler, **kw):
         ops.Strip: unary(lambda arg: _mysql_trim(arg, "both")),
         ops.LStrip: unary(lambda arg: _mysql_trim(arg, "leading")),
         ops.RStrip: unary(lambda arg: _mysql_trim(arg, "trailing")),
+        ops.TimeDelta: _temporal_delta,
+        ops.DateDelta: _temporal_delta,
```
cpcloud, line 259: MySQL has a `timestampdiff` function, but it truncates after compute the difference and there's no `date_trunc` function to manually truncate prior to diffing.

This could be done with date formatting, but that can be done in a follow up.

cpcloud:
Clouds are passing

```
…/ibis on  delta-api is 📦 v6.1.0 via 🐍 v3.10.12 via ❄️  impure (ibis-3.10.12-env)
❯ pytest -m 'bigquery or snowflake' --snapshot-update --allow-snapshot-deletion -n auto -q
bringing up nodes...
....x.....xx.x......x.x......xx..x...............x.......x.....x.x..x.x.x...............x.x...xx....x..x.....x....x.......x..x......x...x.......x...x.x..xx..x..x.....x.....x..x....x......... [  6%]
.xxx..x.x.x..x...x..x..x.....xx...............x.......x......x.............xx......x..x.......x.......xx..xx.........x................x....................x....x...............xx..........xx [ 13%]
.x.x....x.................x..........x...x.....................x.xx.x.x.............x.....x...x..xx....xx........x...x.......x.x...x.........x.s.............x..............................x. [ 20%]
s.......x..........x........................x...............x...x....x......xxxx......x........x......x....x.......xx.....x.x.x....x..x............x...x......x......x.................ssss... [ 26%]
x......x...xxx.....x.......x.x.xx.x....x.x.x........x.x....x..........x....................x.....xx.....x..x...............x.......xx..x.........xx.....xx..xx.x.x.x....x.x...x.xxx.x.x....... [ 33%]
.......x..x....x.xxxxx....x..xxx.x...x..xxx...sssss...xx.x.xx.x.xxxx.xxxx.x.xxx..........x..x.x.......xxsxx...xxxs......x...x...x..x...x...............x.....x.x.................xxx....x..... [ 40%]
.......x..............x...x...x....................x.x..x.x.................x..............x.....x.....................ss..........s....x..x..x....x..x.xx....................x......xx....... [ 47%]
....x...x......x.....x...x..x........ss.x.......x.....sx.........xx..x..x...x.x....x.................x.x.........x.......ss..x.......x..x..........x.x.x.....x......x...xx..x.xx.......xxx..x. [ 53%]
.........xxxxxx.....x.xx.xxxxs.xxxxxxxx...x.xx.xx.x.xxx.......x.x......x.x..........s.xx.xxxx....xx.......xx...........x...x.....x..x.............x.x..x....x.......s......................... [ 60%]
.....x..x.x............x.......x............x.x.......................................x........x.x................x.........ssss.............xx..x...........xxxxxxxx..x...................... [ 67%]
..............xx.....x.....xx....xx.x.x..x.x.x.....x...x...ss..x..x....x...x.....x....x........x.....xx.......x...x...............x......x.x.xx............s.x....x................x......x..x [ 73%]
.......x...x...x......x.........xs............................x.............................x.....................x...........................................................x...s..x........ [ 80%]
...x...xx..xss....x......x...........................x......xxx.xx....x.............x...................x.......x.............x................x.xx..x........................................ [ 87%]
.................................................................................................................................................x.............x.............................. [ 94%]
...................................x...................x......................x...............x...x..s.xxsx..........xx.................x........xx...............s...x.x                      [100%]
2353 passed, 38 skipped, 438 xfailed in 445.06s (0:07:25)
```

cpcloud:
Just to show that the clouds are running the relevant tests successfully:


```
ibis/backends/tests/test_temporal.py::test_delta[snowflake-timestamp] PASSED [ 16%]
ibis/backends/tests/test_temporal.py::test_delta[snowflake-date] PASSED  [ 33%]
ibis/backends/tests/test_temporal.py::test_delta[snowflake-time] PASSED  [ 50%]
ibis/backends/tests/test_temporal.py::test_delta[bigquery-timestamp] PASSED [ 66%]
ibis/backends/tests/test_temporal.py::test_delta[bigquery-date] PASSED   [ 83%]
ibis/backends/tests/test_temporal.py::test_delta[bigquery-time] PASSED   [100%]
```

gforsyth:
> 3\. DuckDB subtracts the first argument from the second argument

WUT

cpcloud:
This whole operation replaces `TRIM` as my go-to for things that look similar but are wildly different.

The temporal delta is a better example because there are multiple correct ways to implement it :)

cpcloud:
@gforsyth I don't see your comment!

ibis/expr/types/temporal.py
```diff
@@ -253,6 +253,67 @@ def __rsub__(self, other: ops.Value[dt.Interval | dt.Time, ds.Any]):
...
+        >>> end.delta(start, "hour")
+        22
+        >>> data = '''tpep_pickup_datetime,tpep_dropoff_datetime
+        ... 2016-02-01T00:23:56,2016-02-01T00:42:28
+        ... 2016-02-01T00:12:14,2016-02-01T00:21:41
+        ... 2016-02-01T00:43:24,2016-02-01T00:46:14
+        ... 2016-02-01T00:55:11,2016-02-01T01:24:34
+        ... 2016-02-01T00:11:13,2016-02-01T00:16:59'''
+        >>> with open("/tmp/triptimes.csv", "w") as f:
+        ...     _ = f.write(data)
```
gforsyth, line 294: Apparently I never hit enter...

Anyway, my question:

I wonder if we should endeavor to avoid using underscores in docstrings that aren't specifically `Deferred`?

cpcloud: Yeah, I think that's a good idea. The problem I was having is that since `f.write` returns a value the doctest expects some output.

Is there a better way to handle ignoring this output?

cpcloud: I suppose I could simply assign it to a different variable that I never use.

cpcloud: Oh, it looks like I can simply use `doctest: +SKIP`, sweet.

cpcloud: I will clean these up across the codebase in a follow-up. Thanks for the callout!

PR #7277, opened 2023-10-02
refactor(ir): unify `ibis.common.pattern` builders and `ibis.expr.deferred`

kszucs:
The main objective here is to avoid logic duplication for deferred and the right hand side of replacement rules/patterns. We started to gradually add similar syntax support for Builders defined in ibis.common.patterns.

The pattern builders are almost identical to the deferred objects except that they must support multiple placeholders (variables), they must be hashable and they must implement proper equality checks which are not supported by deferred expressions.

Basically I moved ibis.expr.deferred to ibis.common.deferred but rewritten the internal abstraction to follow the builder logic existing in ibis.common.patterns. We could say this PR add deferred like functionality to the patterns and tidies the implementation at the same time.

Supplemental but related changes:
- disallow type coercions in `pattern()` factory function by default, this
  prevents unintentional value coercions during pattern matching
- propagate rewritten children when matching on parent nodes, allowing to
  replace dependent nodes in a single pass
- push context variables as keyword arguments to factory functions


ibis/common/patterns.py
```diff
@@ -585,15 +370,16 @@ class Replace(Slotted, Pattern):
...
+        super().__init__(pattern=pattern(matcher), builder=builder)
 
     def match(self, value, context):
         value = self.pattern.match(value, context)
         if value is NoMatch:
             return NoMatch
         # use the `_` reserved variable to record the value being replaced
         # in the context, so that it can be used in the replacer pattern
-        context[_] = value
+        context["_"] = value
```
kszucs, line 382: Chose to reserve `"_"` string key since `var("_")` now produces a `Deferred` object wrapping a `Variable("_")` where the wrapped variable is hashable but the outer deferred object is not, so it is not possible to pass the context as `{_: <value>}` anymore.

ibis/common/patterns.py
```diff
@@ -1867,6 +1620,10 @@ def pattern(obj: AnyType) -> Pattern:
         return _any
     elif isinstance(obj, Pattern):
         return obj
+    elif isinstance(obj, Deferred):
+        if not isinstance(obj.__builder__, Variable):
```
kszucs, line 1624: This validation should be moved to `Capture`

ibis/common/tests/test_deferred.py
```diff
@@ -0,0 +1,519 @@
...
+
+# @pytest.mark.parametrize(
+#     "f, sol",
+#     [
+#         (lambda t: _.x + t.a, "(_.x + <column[int64]>)"),
+#         (lambda t: _.x + t.a.sum(), "(_.x + <scalar[int64]>)"),
+#         (lambda t: ibis.date(_.x, 2, t.a), "date(_.x, 2, <column[int64]>)"),
+#     ],
+# )
+# def test_deferred_repr_with_exprs(f, sol):
```
kszucs, line 515: There is no specialized `repr` for value expressions now, need to figure out how could we add one.

ibis/expr/types/generic.py
```diff
@@ -541,7 +541,8 @@ def bind(table):
             if table := an.find_first_base_table(self.op()):
                 return bind(table)
             else:
-                return de.DeferredCall(bind, (_,))
+                # return de.DeferredCall(bind, (_,))
```
kszucs, line 544: Remove this line.

ibis/selectors.py
```diff
@@ -609,30 +606,6 @@ def if_all(selector: Selector, predicate: Deferred | Callable) -> IfAnyAll:
     return IfAnyAll(selector=selector, predicate=predicate, summarizer=operator.and_)
 
 
-class HashableSlice(Concrete, Coercible):
```
kszucs, line 612: Wasn't used anywhere.

ibis/tests/expr/test_window_frames.py
```diff
@@ -60,7 +60,7 @@ def test_window_builder_rows():
         w0.rows(5)
 
     w1 = w0.rows(5, 10)
```
kszucs, line 62: The methods now operate in-place on the window builder.

ibis/expr/builders.py
```diff
@@ -53,19 +52,14 @@ def when(self, case_expr: Any, result_expr: Any) -> Self:
         result_expr
             Value when the case predicate evaluates to true.
         """
-        return self.copy(
-            cases=self.cases + (case_expr,), results=self.results + (result_expr,)
-        )
+        self.cases += (case_expr,)
```
kszucs, line 55: The expression builders are now mutable objects inheriting from `Annotable` rather than from `Concrete`. This was necessary to support the unhashable `Deferred` objects.

ibis/common/deferred.py
```diff
@@ -0,0 +1,556 @@
...
+from abc import abstractmethod
+from typing import Any, Callable, TypeVar, overload
+
+from ibis.common.bases import FrozenSlotted as Slotted
+from ibis.common.bases import Hashable
+from ibis.common.collections import FrozenDict
+from ibis.util import is_iterable
+
+
+class Builder(Hashable):
```
kszucs, line 16: Hashable object describing the python operations used to construct the desired value. These don't support any sort of syntax sugar, they are rather similar to `ibis.expr.operations`.

ibis/common/deferred.py
```diff
@@ -0,0 +1,556 @@
...
+        -------
+        The constructed object.
+        """
+
+    @abstractmethod
+    def __eq__(self, other: Builder) -> bool:
+        ...
+
+
+class Deferred:
```
kszucs, line 49: The user facing wrapper on builder hierarchies, this is similar to `ibis.expr.types`. Since we need to overload `__eq__` the deferred objects should not be hashable (previously they had a hash implementation returning the id of the object to avoid collisions). 

On the other hand we need a hashable object structure for deferred objects to be used in the pattern system so I split the deferred base class into a different structure following the ibis IR abstraction.

ibis/expr/types/generic.py
kszucs: Removed.

ibis/common/patterns.py
kszucs: Moved.

ibis/common/deferred.py
kszucs, line 1: Perhaps call it `ibis.common.builders.py` instead?

ibis/common/bases.py
```diff
@@ -13,20 +13,43 @@
     from typing_extensions import Self
 
 
-class BaseMeta(type):
+class AbstractMeta(type):
```
kszucs, line 16: There wasn't much advantage of having a `Base` and an `Abstract(Base)` since the `__abstractmethod__` calculation should be cheap because it is happening on class definition.

jcrist:
Can you comment on the motivation here? An internal refactor could be done for many reasons:

- Reduce code (this PR increases LOC, so that doesn't seem accurate)
- Make a new feature or bugfix possible
- Improve performance?
- ...?

I'm curious what was lacking in the current implementation of deferred that this PR improves.

ibis/common/patterns.py
```diff
@@ -559,6 +340,10 @@ class Capture(Slotted, Pattern):
     pattern: Pattern
 
     def __init__(self, key, pat=_any):
+        if isinstance(key, Deferred):
```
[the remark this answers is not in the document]
kszucs: Updated.

kszucs:
Sorry, I should have written the description first. The main objective here is to avoid logic duplication for deferred and the right hand side of replacement rules/patterns. We started to gradually add similar syntax support for Builders defined in ibis.common.patterns.

The pattern builders are almost identical to the deferred objects except that they must support multiple placeholders (variables), they must be hashable and they must implement proper equality checks which are not supported by deferred expressions. 

Basically I moved ibis.expr.deferred to ibis.common.deferred but rewritten the internal abstraction to follow the builder logic existing in ibis.common.patterns.

ibis/expr/builders.py
kszucs: Switched backed to be immutable because `_.col.sum().over(frame)` should be supported where frame is a window builder and the entire object should be hashable.

ibis/common/patterns.py
```diff
@@ -585,15 +370,16 @@ class Replace(Slotted, Pattern):
     builder: Builder
 
     def __init__(self, matcher, replacer):
-        super().__init__(pattern=pattern(matcher), builder=builder(replacer))
+        builder = deferred(replacer).__builder__
```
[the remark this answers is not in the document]
kszucs: Updated.

ibis/expr/api.py
```diff
@@ -191,8 +191,8 @@
 True
 """
 
-
-_ = deferred = Deferred()
+# TODO(kszucs): consider to use `table` variable name instead
+_ = deferred = var("_")
```
jcrist, line 195: Since we construct a `var("_")` in `ibis.common.deferred` as well, perhaps just import it from there?

Also regarding the TODO, we already have `ibis.table`, so can't really name this `table`.

ibis/common/tests/test_deferred.py
```diff
@@ -0,0 +1,540 @@
...
+    p = v.copy
+    assert p.__builder__ == Attr(v, "copy")
+    assert p.__builder__.build({"v": [1, 2, 3]})() == [1, 2, 3]
+
+    p = v.copy()
+    assert p.__builder__ == Call(Attr(v, "copy"))
+    assert p.__builder__.build({"v": [1, 2, 3]}) == [1, 2, 3]
+
+
+class Table(dict):
```
jcrist, line 200: Why reimplement a table/column set of objects here rather than reusing our existing implementation? Is it just to avoid importing `ibis.expr` in tests for `ibis.common`? If so, I'd argue that's not worth the extra code in the tests file.

ibis/common/tests/test_patterns.py
```diff
@@ -24,13 +24,13 @@
 
 from ibis.common.annotations import ValidationError
 from ibis.common.collections import FrozenDict
+from ibis.common.deferred import Call, var
```
jcrist, line 27: IIUC this PR isn't merely a refactor, but something that adds support for deferred expressions in the rewrite side of patterns. If so, could we add some tests for this? Apologies if I missed them, I didn't see them below in this file.

jcrist (commented):
> We could say this PR add deferred like functionality to the patterns and tidies the implementation at the same time.

Got it - so IIUC this PR is adding a feature (deferred expressions to the patterns), and not just a refactor. Thanks for the explanation, knowing the context here is important for reviewing.

I'm a bit hesitant to couple the patterns and deferred expressions. While they _can_ share an implementation, conceptually they're different things. Currently `ibis/expr/deferred.py` is a pretty short self-contained file, making it easy to understand and make changes to the deferred implementation. Adding the pattern system as an unrelated downstream consumer of the same API may make it harder to modify this feature in the future. It's definitely an opinion thing, but IMO I'd prefer to have duplicate code here in the patterns rather than coupling patterns and deferred objects just to keep these systems separate. Not going to block on this though.

ibis/common/deferred.py
```diff
@@ -0,0 +1,574 @@
...
+        if isinstance(value, cls):
+            return value
+        elif isinstance(value, Deferred):
+            return value.__builder__
+        else:
+            raise CoercionError(f"Cannot coerce {value!r} to {cls.__name__!r}")
+
+
+class Deferred(Final):
+    __slots__ = ("__builder__", "__representation__")
```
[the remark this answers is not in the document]
kszucs: Makes sense, going to update.

ibis/common/deferred.py
```diff
@@ -0,0 +1,574 @@
...
+from typing import Any, Callable, TypeVar, overload
+
+from ibis.common.bases import Final, Hashable
+from ibis.common.bases import FrozenSlotted as Slotted
+from ibis.common.collections import FrozenDict
+from ibis.common.typing import Coercible, CoercionError
+from ibis.util import is_iterable
+
+
+class Builder(Coercible, Hashable):
```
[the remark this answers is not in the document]
kszucs: Actually I'm not satisfied with it either. Perhaps call the deferred (and hashable) operations as `Deferred` and have an `Expr` or something similar wrapper on top of them? 

1. `Builder` => `DeferredOp`, `Deferred` keep it as is
2. `Builder` => `Deferred`, `Deferred` => `DeferredExpr`? 
3. Suggestions?

ibis/common/deferred.py
```diff
@@ -0,0 +1,574 @@
...
+    """A deferred is a function that takes a context and returns a new object.
```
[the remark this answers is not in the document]
kszucs: Yes, these need to be updated.

ibis/common/deferred.py
```diff
@@ -0,0 +1,574 @@
...
+        super().__init__(name=name)
+
+    def __repr__(self):
+        return str(self.name)
+
+    def build(self, context):
+        return context[self.name]
+
+
+class Just(Slotted, Builder):
```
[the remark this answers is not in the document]
kszucs: I associate identifier from `Id`, so `Identity` would be better in that sense, but I still feel `Just` just a bit better :)

Another options could be `Same`, `Exactly`.

ibis/common/deferred.py
```diff
@@ -0,0 +1,574 @@
...
+        else:
+            return super().__create__(value)
+
+    def __init__(self, value):
+        if isinstance(value, list):
+            value = tuple(value)
+        elif isinstance(value, set):
+            value = frozenset(value)
+        elif isinstance(value, dict):
+            value = FrozenDict(value)
```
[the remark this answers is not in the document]
kszucs: Agree, I'm not satisfied with it either. We can have another hashable `Builder`/`Deferred` which reconstructs the unhashable value on `.build()` so this is easy to overcome.

ibis/common/deferred.py
```diff
@@ -0,0 +1,574 @@
...
+        elif args:
+            return f"{func}({args})"
+        elif kwargs:
+            return f"{func}({kwargs})"
+        else:
+            return f"{func}()"
+
+    def __call__(self, *args, **kwargs):
+        if self.args or self.kwargs:
+            raise TypeError("Further specification of Call object is not allowed")
```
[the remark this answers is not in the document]
kszucs: This is a leftover from the previous pattern implementation. There is an object called `Namespace` in patterns which is responsible to provide the `c.ClassName` shortcuts and that (wrongly) use this method for now. It needs to be removed.

ibis/common/deferred.py
```diff
@@ -0,0 +1,574 @@
...
+    def __init__(self, func, arg):
+        assert not isinstance(func, (Deferred, Builder))
+        super().__init__(func=func, arg=builder(arg))
+
+    def __repr__(self):
+        try:
+            symbol = _operator_symbols[self.func]
+            return f"{symbol}{self.arg!r}"
+        except KeyError:
+            return f"{self.func.__name__}(self.arg!r)"
```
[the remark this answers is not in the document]
kszucs: I was thinking of adding shortcuts to `Call` like the following:

```py
class Call(...):
    @classmethod
    def __create__(self, *args, **kwargs):
        if len(args) == 1 and not kwargs:
            return UnaryCall(*args)
        elif ...:  # e.g. check for binary calls
            ...
        else:
            return super().__create__(*args, **kwargs)
```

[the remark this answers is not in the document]
kszucs: Not used yet.

ibis/common/deferred.py
```diff
@@ -0,0 +1,574 @@
...
+    def __init__(self, func, left, right):
+        assert not isinstance(func, (Deferred, Builder))
+        super().__init__(func=func, left=builder(left), right=builder(right))
+
+    def __repr__(self):
+        try:
+            symbol = _operator_symbols[self.func]
+            return f"({self.left!r} {symbol} {self.right!r})"
+        except KeyError:
+            return f"{self.func.__name__}({self.left!r}, {self.right!r})"
```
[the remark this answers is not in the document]
kszucs: Same as for the `UnaryCall`.

ibis/common/deferred.py
```diff
@@ -0,0 +1,574 @@
...
+        # allow nesting deferred patterns in dicts
+        return Mapping(obj)
+    elif is_iterable(obj):
+        # allow nesting deferred patterns in tuples/lists
+        return Sequence(obj)
+    elif callable(obj):
+        return Factory(obj)
+    else:
+        # the object is used as a constant value
+        return Just(obj)
```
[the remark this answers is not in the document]
kszucs: Actually they are treated differently.  `Just(sequence)` won't try to create `builders` for each element of the sequence, it will "just" return the sequence as is (of course in an immutable form at the moment). 

On the other hand `builder(sequence)` will turn each element into a separate builder supporting nested builders.

ibis/expr/api.py
kszucs: What I mean under the todo note is:

```py
_ = deferred = var("table")
```

which should be used as 

```py
deferred.resolve(table=table)
```

or

```py
builder.build({"table": table})
```

which could be a more explicit regarding the nature of the `_` placeholder.

ibis/expr/builders.py
```diff
@@ -4,17 +4,17 @@
...
 
 import ibis
-import ibis.expr.datashape as ds
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 import ibis.expr.rules as rlz
 import ibis.expr.types as ir
 from ibis import util
 from ibis.common.annotations import annotated
+from ibis.common.deferred import Builder as DeferredBuilder
```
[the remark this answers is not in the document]
kszucs: Yeah, it is especially annoying.

ibis/common/tests/test_deferred.py
kszucs: Yes, that is the main reason. I prefer mocking to keep the testable logic minimal.

ibis/common/tests/test_patterns.py
kszucs: The tests exercising the `Replace` pattern use these deferred objects. This functionality has been covered before this PR just without the rich syntax sugar provided by the deferred API.

ibis/expr/builders.py
```diff
@@ -67,7 +67,7 @@ def else_(self, result_expr: Any) -> Self:
         """
         return self.copy(default=result_expr)
 
-    def end(self) -> ir.Value | Deferred:
+    def end(self) -> ir.Value:
```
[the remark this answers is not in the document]
kszucs: Correct. Hopefully we can remove this case builder soon enough.

ibis/common/deferred.py
kszucs: My intent here was to explicitly discourage the use of deferreds as dictionary keys. 

Imagine a use case like the following:

```py
# defined in module A
x = var("x")
mapping = {x: "a"}

# defined in module B
from moduleA import mapping

x = var("x")

mapping[x]  # raises KeyError
```

While it is logically correct, it is still surprising.






```

Currently they work uniformly because I restricted the `context` keys to be strings so `"_"` will be used across all deferred objects, but the previous implementation I had in patterns supported any kind of hashable keys for context. 

I was just more comfortable with

File: ibis/common/tests/test_patterns.py (hunk 15 above)
Comment (jcrist) [reply ibis/common/tests/test_patterns.py:27]: Sure. Since the "rich syntax sugar provided by the deferred API" is the new feature being added here though, I'd argue that there should be at least a test including a full example of this usage. Can you add at least a simple test showing how the pieces all fit together here?

File: ibis/expr/api.py (hunk 13 above)
Comment (jcrist) [reply ibis/expr/api.py:195]: The point of the nice repr's for deferred expressions is to show them the same as the code the user wrote to generate them. Defining `_ = var("table")` would be non-ideal in that case, since it would then repr as `table` instead of `_`.

File: ibis/common/deferred.py (hunk 19 above)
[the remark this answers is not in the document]
Comment (jcrist) [reply ibis/common/deferred.py:212]: If you feel strongly then keeping `Just` as is seems fine for now.

File: ibis/common/tests/test_deferred.py (hunk 14 above)
Comment (jcrist) [reply ibis/common/tests/test_deferred.py:200]: Can you add a comment above the mocked api saying so (and maybe use the word "Mock" in the class names)? I had to read a bit to understand that `Column` wasn't an ibis column but a local class definition.

File: ibis/common/deferred.py (hunk 17 above)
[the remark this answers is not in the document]
Comment (jcrist) [reply ibis/common/deferred.py:17]: I'd go with option 1.

File: ibis/common/deferred.py (hunk 9 above)
Comment (jcrist) [reply ibis/common/deferred.py]: That seems fair. I don't think we have any current APIs that support deferred values as dict keys anyway.

File: ibis/common/tests/test_deferred.py (hunk 14 above)
Comment (kszucs) [reply ibis/common/tests/test_deferred.py:200]: Sure, going to rename them.

File: ibis/common/deferred.py (hunk 16 above)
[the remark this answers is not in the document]
Comment (kszucs) [reply ibis/common/deferred.py:60]: Updated.

File: ibis/common/tests/test_deferred.py (hunk 14 above)
Comment (kszucs) [reply ibis/common/tests/test_deferred.py:200]: Renamed.

File: ibis/expr/builders.py (hunk 26 above)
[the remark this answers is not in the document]
Comment (kszucs) [reply ibis/expr/builders.py]: Restored.

File: ibis/common/deferred.py (hunk 17 above)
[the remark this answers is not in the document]
Comment (cpcloud) [reply ibis/common/deferred.py:18]: Option 1 seems good to me.

File: ibis/common/tests/test_patterns.py (hunk 15 above)
Comment (kszucs) [reply ibis/common/tests/test_patterns.py]: Added tests to `ibis/expr/operations/tests/test_rewrites.py`

File: ibis/common/bases.py (hunk 10 above)
Comment (kszucs) [reply ibis/common/bases.py:16]: Pulled out to a separate PR.

File: ibis/common/deferred.py (hunk 20 above)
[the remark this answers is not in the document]
Comment (kszucs) [reply ibis/common/deferred.py]: I introduced a `PseudoHashable` wrapper type which tries to hash the wrapped object and falls back to its identifier. If `Just()` receives an unhashable object it instantiates a `JustUnhashable` builder instead which uses the `PseudoHashable` wrapper to keep the builder itself hashable. This way we have a faster path for more common immutable objects and a slightly slower but still hashable one for unhashable inputs.

File: ibis/common/tests/test_deferred.py (hunk 3 above)
Comment (cpcloud) [reply ibis/common/tests/test_deferred.py:596]: @kszucs Is this resolved?

File: ibis/common/deferred.py (hunk 22 above)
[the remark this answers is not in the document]
Comment (kszucs) [reply ibis/common/deferred.py]: Updated.

File: ibis/common/deferred.py (hunk 23 above)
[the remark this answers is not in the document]
Comment (kszucs) [reply ibis/common/deferred.py]: Updated.

**Review by cpcloud — APPROVED** [review]
After looking through this PR, I think it makes sense to unify these implementations.

Deferred, while nominally distinct, doesn't appear to be materially different from pattern variables.

From what I see there are at least two benefits to sharing the implementation:

1. Replacements of more complex expressions are more readable.
2. The two implementations of variable resolution (one for pattern replacements, and one for binding leaves of deferred instances to an expression) are unified into a single concept of pattern replacement.

I think it's difficult to tell whether this refactor will make it more challenging to change the behavior of deferred. IMO I think this is low risk and we can always revisit this if it becomes a problem.

File: ibis/common/tests/test_deferred.py (hunk 3 above)
Comment (kszucs) [reply ibis/common/tests/test_deferred.py]: Yes.

File: ibis/common/deferred.py (hunk 17 above)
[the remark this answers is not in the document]
Comment (kszucs) [reply ibis/common/deferred.py:16]: `Deferred` is a better base class name, so I chose to rename `Builder` to `Deferred` and the wrapper deferred class to `DeferredProxy`.

File: ibis/expr/api.py (hunk 13 above)
Comment (kszucs) [reply ibis/expr/api.py:195]: Removed the comment.

Comment (kszucs) [reply ibis/expr/api.py:194]: Keeping as is.

File: ibis/common/deferred.py [hunk 27]
```diff
@@ -0,0 +1,608 @@
...
+        elif callable(obj):
+            return getattr(obj, "__name__", repr(obj))
+        else:
+            return repr(obj)
+
+    def resolve(self, context):
+        return self.value
+
+
+class JustUnhashable(FrozenSlotted, Deferred):
```
jcrist, line 265: Possible code simplification:

Could we unify this implementation with the above to only have a single `Just` class that may optionally wrap `value` in `PseudoHashable` if needed? I think the only difference would be the implementation of `.resolve` as:

```python
def resolve(self, context):
    return self.value.obj if isinstance(self.value, PseudoHashable) else self.value
```

Having one fewer wrapped class to worry about might be nice, up to you.

ibis/expr/api.py
```diff
@@ -191,8 +191,7 @@
 True
 """
 
-
-_ = deferred = Deferred()
+_ = deferred = var("_")
```
jcrist, line 194: Can we import this from `ibis.comon.deferred`, rather than calling `var("_")` in two places? Right now it's not clear to me why we create a `_` instance twice.

ibis/common/deferred.py
```diff
@@ -0,0 +1,608 @@
...
+    def __coerce__(cls, value):
+        if isinstance(value, cls):
+            return value
+        elif isinstance(value, DeferredProxy):
+            return value._deferred
+        else:
+            raise CoercionError(f"Cannot coerce {value!r} to {cls.__name__!r}")
+
+
+class DeferredProxy(Slotted, Immutable, Final):
```
jcrist, line 61: When we present `_` to users, we talk about deferred objects and deferred expressions. I personally think it'd be easier to explain and document if instances of those expressions were called `Deferred` rather than `DeferredProxy`. Use the simpler name for the user-facing thing.

Perhaps call this `Deferred` and the wrapped instances `DeferredOp` or `DeferredExpr`? :shrug:.

jcrist (commented):
A few comments/questions, but overall seems fine (provided tests pass).

ibis/expr/api.py
kszucs: Indeed, updating.

ibis/common/deferred.py
kszucs: Ideally it should be transparent to the users, at least we shouldn't expose neither of the types for public consumption, so how about deferring the decision? :)

I'm actually kinda satisfied with `DeferredProxy` since it does nothing but proxying to provide syntactic sugar.

ibis/common/deferred.py
kszucs: They are indeed almost identical, so your request totally makes sense. On the other hand I'd rather keep them separate for performance reasons. The `Just` deferreds are basically the "leaf" objects responsible to produce the requested values and they are frequently used. Perhaps it doesn't make a noticeable performance difference, but mutable values should be rarely used.

ibis/common/deferred.py
jcrist: > so how about deferring the decision?

I mean, if we're deferring the decision, then we should keep the class name the same as it was before this PR - `Deferred` :).

I wouldn't risk a naming bikeshed discussion if I didn't feel strongly about this. We need to call these things something in the documentation - currently we call them "deferred expressions" not "deferred proxies". These types will also show up in user's interactive sessions when introspecting objects, as well as in type annotations. Since the wrapped objects *are not user facing* then the naming of them feels less important to me. I don't see a reason to reserve the shorter/better `Deferred` name for an internal implementation detail forcing the user of `DeferredProxy` for the public type.

Further, there are cases in downstream APIs where we do need to access the type `Deferred` directly for introspection reasons (we do this in `ibisml` for example). Changing the class name here feels like a needless breaking change.

cpcloud: @kszucs What's the resolution here? Seems like we should keep `Deferred` for now to avoid breaking ibisml

kszucs: I renamed the `Builder => Resolver` and kept `Deferred` as it used to be, no breakage.

jcrist (approved):

PR #7279, opened 2023-10-02
Improve reference docs

NickCrews:
Just some small things that have been missing from the docs that have been bothering me for a while.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -966,6 +966,15 @@ def case() -> bl.SearchedCaseBuilder:
     Use the `.when` method on the resulting object followed by `.end` to create a
     complete case expression.
 
+    Returns
+    -------
+    SearchedCaseBuilder
+        A builder object to use for constructing a case expression.
+
+    See Also
+    --------
+    [`Value.case()`](./expression-generic.qmd#ibis.expr.types.generic.Value.case)
+
     Examples
     --------
     >>> import ibis
@@ -998,11 +1007,6 @@ def case() -> bl.SearchedCaseBuilder:
     │     3 │ *      │     7 │    21.0 │
     │     4 │ /      │     8 │     0.5 │
     └───────┴────────┴───────┴─────────┘
-
-    Returns
-    -------
-    SearchedCaseBuilder
-        A builder object to use for constructing a case expression.
     """
     return bl.SearchedCaseBuilder()
 
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -499,6 +499,10 @@ def desc(expr: ir.Column | str) -> ir.Value:
     expr
         The expression or column name to use for sorting
 
+    See Also
+    --------
+    [`Value.desc()`](./expression-generic.qmd#ibis.expr.types.generic.Value.desc)
+
     Examples
     --------
     >>> import ibis
@@ -532,6 +536,10 @@ def asc(expr: ir.Column | str) -> ir.Value:
     ----------
     expr
         The expression or column name to use for sorting
+
+    See Also
+    --------
+    [`Value.asc()`](./expression-generic.qmd#ibis.expr.types.generic.Value.asc)
 
     Examples
     --------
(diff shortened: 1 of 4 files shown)


ibis/expr/types/relations.py
```diff
@@ -104,6 +104,39 @@ def f(  # noqa: D417
...
 class Table(Expr, _FixedTextJupyterMixin):
+    """An immutable and lazy dataframe.
+
+    Analogous to a SQL table or a pandas DataFrame. A table expression contains
+    an [ordered set of named columns](./schemas.qmd#ibis.expr.schema.Schema),
+    each with a single known type. Unless explicitly ordered with an
+    [`.order_by()`](#ibis.expr.types.relations.Table.order_by),
+    the order of rows is undefined.
+
+    By "immutable", we mean that a table expression cannot be modified: every
```
cpcloud, line 115:
```suggestion
    Table immutability means that the data underlying an Ibis `Table` cannot be modified: every
```

ibis/expr/types/relations.py
```diff
@@ -104,6 +104,39 @@ def f(  # noqa: D417
...
+    method on a Table returns a new Table with those changes. By "lazy", we
+    mean that a Table expression does not contain any actual data.
```
cpcloud, line 117:
```suggestion
    method on a Table returns a new Table with those changes. Laziness
    means that an Ibis `Table` expression does not run your computation every time you construct a new, in contrast to pandas.
```

ibis/expr/types/relations.py
```diff
@@ -675,7 +710,7 @@ def _ensure_expr(self, expr):
 
     @property
     def columns(self) -> list[str]:
-        """The list of columns in this table.
+        """The `list[str]` of column names in this table.
```
cpcloud, line 713: Can we keep this docstring as is? The type annotation is already there, and the docstring is about the concept, not the specific way in which it is implemented.

NickCrews: Do you see how in the live docs on the website you can't see the type annotation?

ibis/expr/api.py
```diff
@@ -499,6 +499,10 @@ def desc(expr: ir.Column | str) -> ir.Value:
     expr
         The expression or column name to use for sorting
 
+    See Also
+    --------
+    [`Value.desc()`](#ibis.expr.types.generic.Value.desc)
```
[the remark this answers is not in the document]
NickCrews: In my locally built docs these seem to work because they are placed into the same .qmd file by quartodoc. Will they work differently on the published docs? I suppose making it absolute will make it easier to update the path in case we reorganize which .qmd file each item goes into, with this relative path I could see links silently breaking in the future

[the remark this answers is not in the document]
cpcloud: Hm, yeah good point. These *should* be caught be the link checker, but the full path is also easier to understand IMO

NickCrews:
Thanks for the review @cpcloud! I applied all your suggestions.

ibis/expr/types/generic.py
```diff
@@ -116,21 +184,26 @@ def cast(self, target_type: dt.DataType) -> Value:
 
         return op.to_expr()
 
-    def try_cast(self, target_type: dt.DataType) -> Value:
+    def try_cast(self, target_type: Any) -> Value:
```
NickCrews, line 187: IDK if you saw this type hint change @cpcloud. The old one was overly-restrictive. It might be nice if the dtype module exported some type alias for `TypeLike = Union[dt.DataType, str, np.dtype, ...]` so that here and in other places we could consume it.

ibis/expr/types/relations.py
NickCrews: filed at https://github.com/machow/quartodoc/issues/284

cpcloud (approved):
Nice!

NickCrews:
@cpcloud 
Per https://github.com/machow/quartodoc/issues/284, we could get type annotations on Table.columns by adding
```
Attributes
----------
columns
        a copy of the existing docstring
```
to `Table.__doc__`. What do you think, is this duplication worth it?

PR #7285, opened 2023-10-03
feat: implement window table valued functions

chloeh13q:
Implement window table valued functions & remove a public `translate()` method from Flink backend since that was a temporary solution for not having a full-fledged backend. Since we now have an actual Flink backend, `Backend.compile()` can be used to translate expressions instead. This also saves us from having to maintain two separate catalogs of translation rules.


ibis/expr/operations/tvf.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def __init__(self, table, **kwargs):
+        super().__init__(table=table, **kwargs)
+
+    @property
+    @abstractmethod
+    def function_type(self) -> str:
+        """Abstract class attribute.
+
+        This makes TableValuedFunction and WindowTVF abstract classes.
```
kszucs, line 38: Relation is already an abstract class requiring a `schema` attribute. Is the `function_type` attribute desired? Looks like the type uniquely determines the value of it.

ibis/expr/operations/tvf.py
```diff
@@ -0,0 +1,121 @@
...
+
+@public
+class TumbleTableFunction(WindowTVF):
+    """TUMBLE table-valued function."""
+
+    function_type = TableValuedFunctionTypes.TUMBLE
+    window_size: Scalar[dt.Interval]
+    offset: Optional[Scalar[dt.Interval]] = None
+
+    def __init__(self, table, time_col, window_size, offset=None, **kwargs):
```
kszucs, line 70: This init method can be omitted.

ibis/expr/operations/tvf.py
```diff
@@ -0,0 +1,121 @@
...
+@public
+class HopTableFunction(WindowTVF):
+    """HOP table-valued function."""
+
+    function_type = TableValuedFunctionTypes.HOP
+    window_size: Scalar[dt.Interval]
+    window_slide: Scalar[dt.Interval]
+    offset: Optional[Scalar[dt.Interval]] = None
+
+    def __init__(
```
kszucs, line 89: This init method can be omitted.

NickCrews:
This makes me consider namespaces for different kinds of functionality. I could see Tables getting a lot of temporal functionality like this, and it would be nice if those were organized in such a way they were
- close together, so you could look at all the temporal stuff together
- quarantined, so they don't pollute the namespace where more general-purpose functionality like .fillna() lives
Possibly a similar story for geospatial and other specialized domains.

IDK, maybe overkill. But might be nice to spend 15 minutes and look through pandas, geopandas, vaex, etc for examples of domain-specific stuff so we know what people are going to want in the future.

ibis/expr/operations/tvf.py
chloeh13q: `schema` is implemented on `WindowTVF`, so `TableValuedFunction` is an abstract class but `WindowTVF` isn't. The intention of having `function_type` is mostly just to make `WindowTVF` an abstract class, but I don't feel strongly about this implementation. We can certainly remove it.

jcrist (commented):
General comment on the API design here.

For reference for anyone not familiar with this aspect of Flink SQL (e.g. myself 30 minutes ago), these are all examples of [Windowing Table Valued Functions](https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/dev/table/sql/queries/window-tvf/). These are table-valued operations that return the input table with 3 new columns: `window_start`, `window_end`, and `window_time` (names are hardcoded), providing info on the window values for each row. Each function type is a different way of specifying a window.

I kind of agree with @NickCrews that these specific operations should be semantically grouped in some way. I also think we'll want to have an api that include the word "window" somewhere to better indicate that these are windowing operations.

I see a few ways of doing this:

## 1. Use a common prefix

The low-tech solution would be to a use a common prefix. Perhaps `window_tumble`/`window_hop`/`window_cumulate`, each with method signatures pretty much as you've written them.

```python
t.window_tumble("timestamp", ibis.interval(minutes=15))
```

## 2. A single method that takes an object specifying the window type

This would group all windowing methods into a single method that takes an object indicating the window type. This makes sense for all current windowing TVFs since they return tables with the same output schemas.

Perhaps something like:

```python
# Not attached to the name here
t.windowize(ibis.tumble("timestamp", ibis.interval(minutes=15)))

# or maybe the time_col is a common top-level column?
t.windowize("timestamp", ibis.tumble(ibis.interval(minutes=15)))
```

## 3. A window builder API, similar to `group_by`

This would split doing a windowing operation into a two-phase call, where the first method returns a `WindowedTable` value (or something) with additional methods for applying the windowing operation.

```python
# Not attached to the method name here
t.window_by("timestamp").tumble(ibis.interval(minutes=15))

# All current windowing TVFs take a `time_col` as an argument. If there's a chance
# that future APIs (like the sessioning windows called out as TODO in the flink
# docs), then we might not want to include the `time_col` as a top-level arg?
```

---

Overall I'm leaning towards option 3. It minimizes the impact on the `Table` namespace and groups the windowing APIs together in a way that should aid with discoverability and make documenting them easier. It also provides a clear location to add additional windowing APIs in the future. I'd also be fine with option 2, but would prefer to avoid option 1 if possible.

chloeh13q:
@jcrist I think that's reasonable. 

> # All current windowing TVFs take a `time_col` as an argument. If there's a chance
# that future APIs (like the sessioning windows called out as TODO in the flink
# docs), then we might not want to include the `time_col` as a top-level arg?

I think session windows would require `time_col` as an argument too.

Regarding the options above, option 3 looks fine to me. Option 2 is more similar to the way pyspark does it, e.g.
```python
windowedCountsDF = \
  eventsDF \
    .withWatermark("eventTime", "10 minutes") \
    .groupBy(“deviceId”, window("eventTime", "10 minutes", "5 minutes")) \
    .count()
```

Happy to move forward with option 3 if we can reach agreement here.

gforsyth:
I'm in favor of option 3 -- option 1 is ok.  
I would opt to avoid adding additional functions to the top-level namespace that only apply to a single backend (for now).  Plus it seems cleaner to work with the window builder API.

deepyaman:
+1 on option 3

jcrist:
Sounds like we're in agreement on option 3.

Aside - I think we should be able to support tumbling windows in a straightforward manner for a few of our existing backends (`duckdb` and `postgres` for sure, *probably* `clickhouse`, likely others in ways I haven't thought of yet). The other windowing types seem trickier to me, but may also be something we could support. So these don't have to be streaming-backend-only operations.

NickCrews:
Would option 3 support [all 4 of the functionalities supported in pandas](https://pandas.pydata.org/docs/user_guide/window.html)?

gforsyth:
> Would option 3 support [all 4 of the functionalities supported in pandas](https://pandas.pydata.org/docs/user_guide/window.html)?

Only 2 of those 4 have table-level support in pandas itself.  

The rolling and expanding windows seem like they map directly onto standard windowing bounds.  I've never come across weighted windows in a context other than signal processing, so I don't know that there's going to be support for those operations in backends that aren't pandas.

NickCrews:
Cool, that makes sense, then we can ignore the weighted stuff for now. Someone could implement it at a higher level wrapper for ibis if they really wanted.

chloeh13q:
I updated the user-facing API according to the discussion above, but kept the internal operator implementation mostly the same. Let me know what you think!

ibis/expr/types/tvf.py
```diff
@@ -0,0 +1,128 @@
...
+        ----------
+        window_size
+            Width of the tumbling windows.
+        offset
+            An optional parameter to specify the offset which window start should be shifted by.
+
+        Returns
+        -------
+        Table
+            Table expression.
```
jcrist, line 61: These docstrings could use some improvement (example, better description of what each method and parameter does), but this can be done in a follow-up if you want. The examples in particular would be easier to write once we add support for a batch backend like duckdb.

ibis/tests/expr/test_table.py
```diff
@@ -1881,3 +1881,43 @@ def test_invalid_distinct_empty_key():
     t = ibis.table(dict(a="int", b="string"), name="t")
     with pytest.raises(com.IbisInputError):
         t.distinct(on="c", keep="first")
+
+
+def test_tumble_tvf(schema, table):
```
jcrist, line 1886: `test_table` is long enough as is. I think since this functionality is new and more streaming specific, a new test file might be nice. `test_window_by.py` or something (depending on how we name the operations/types files).

ibis/expr/types/tvf.py
```diff
@@ -0,0 +1,128 @@
+from __future__ import annotations
```
jcrist, line 1: `tvf.py` isn't the best file name:

- It's an acronym that not everyone will be familiar with
- These functions are table valued functions, but they're also specific to time windowing.

Since we don't support any other table valued functions (and may not plan to, at least under that name), perhaps it'd make more sense to name/group this functionality based on what it does rather than the function signature (`f(table: Table) -> Table)`). Perhaps `window_by.py` or `temporal_windows.py`?

Same goes for `ibis/expr/operations/tvf.py`.

ibis/backends/flink/compiler/core.py
```diff
@@ -28,8 +30,16 @@ def _format_in_memory_table(self, op):
...
 
+    def _format_window_tvf(self, op) -> str:
+        return f"TABLE({op.function_type}({format_window_tvf_params(op, self)}))"
+
     def _format_table(self, op) -> str:
-        result = super()._format_table(op)
+        ctx = self.context
+        if isinstance(op, ops.WindowTVF):
+            formatted_table = self._format_window_tvf(op)
+            return str(formatted_table) + f" {ctx.get_ref(op)}"
```
jcrist, line 40:
```suggestion
            return f"{formatted_table} {ctx.get_ref(op)}"
```

jcrist (commented):
This is getting close! I'm happy with the API design, just some comments on the organization and internals. Excited to have this functionality land in ibis!

ibis/expr/operations/temporal_windows.py
```diff
@@ -0,0 +1,59 @@
...
+from public import public
+
+import ibis.expr.datatypes as dt
+from ibis.expr.operations.core import Column, Scalar  # noqa: TCH001
+from ibis.expr.operations.relations import Relation
+from ibis.expr.schema import Schema
+
+
+@public
+class WindowTVF(Relation):
```
jcrist, line 14: Perhaps `WindowingTVF` might be a better name?

And the subclassed ops would be: `TumbleWindowingTVF`, `HopWindowingTVF`, and `CumulateWindowingTVF`. This feels a bit better grammatically to me, and also ensures the tumble/hop/cumulate variants all have "window" in their name.

jcrist (approved):
One small naming nit/suggestion, otherwise this looks good to me! Nice work!

The test failure is a known flaky test (I merged in master which should resolve this).

ibis/expr/operations/temporal_windows.py
chloeh13q: Done!

chloeh13q:
@jim it looks like the same test is still failing, any idea?

jcrist:
Fixed the issue. Thanks y'all!

PR #7286, opened 2023-10-03
fix: datafusion: do_connect: properly handle config-is-actually-context

dlovell:
when `config` is an instance of `SessionContext`, it shouldn't subsequently be used as a config


cpcloud:
Any chance you can add a small test for this?

cpcloud (approved):
This LGTM, I can fix up the lints for you.

cpcloud:
Hm, it looks like I can't push your fork because GitHub doesn't let maintainers push to forked `master` branches, only non-`master` branches.

That's okay though. Can you run `black` and push up a `style: fix lints` commit (or similar)?

cpcloud:
Set to automerge!

PR #7297, opened 2023-10-04
docs: add deprecation to .nullifzero docstring

NickCrews:
Although users got a deprecationWarning
whenever they used these in code,
in the docs there is no
indication they are deprecated.

Not sure, possibly we could do some hack where we add this
function to the class after the class block, and then
it wouldn't even show up in the documentation?
but probably if a user sees the deprecation warning, they might
want to look up that method on the docs for more info,
so it might be nice if there were something there...

### DIFF
--- a/ibis/expr/types/numeric.py
+++ b/ibis/expr/types/numeric.py
@@ -8,10 +8,10 @@ from public import public
 
 import ibis
 import ibis.expr.operations as ops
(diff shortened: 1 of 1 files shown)


cpcloud (approved):
I'll follow this up with some automation of generating these docstrings.

PR #7301, opened 2023-10-06
feat(api): allow selectors in `order_by`

cpcloud:
This PR implements selector support in the `.order_by` method

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -1296,7 +1296,13 @@ class Table(Expr, _FixedTextJupyterMixin):
 
     def order_by(
         self,
-        by: str | ir.Column | Sequence[str] | Sequence[ir.Column] | None,
+        by: str
+        | ir.Column
+        | s.Selector
+        | Sequence[str]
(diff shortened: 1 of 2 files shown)


ibis/expr/types/relations.py
```diff
@@ -1421,19 +1427,21 @@ def order_by(
...
             if isinstance(item, tuple):
                 if len(item) != 2:
-                    raise ValueError(
-                        "Tuple must be of length 2, got {}".format(len(item))
-                    )
-                item = (bind_expr(self, item[0]), item[1])
-                used_tuple_syntax = True
+                    raise ValueError(f"Tuple must be of length 2, got {len(item):d}")
+                sort_keys.append(bind_expr(self, item[0]), item[1])
+            elif isinstance(item, s.Selector):
```
kszucs, line 1438: It would be nice to condense all this logic in `bind_expr()`, but that requires a larger overhaul.

cpcloud: Indeed, we should do that in a follow-up.

PR #7303, opened 2023-10-06
feat(exasol): add exasol backend

Nicoretti:
:wave: Hi Ibis Team,

I reached out to you a few months ago ([here](https://github.com/ibis-project/ibis/discussions/6448)), 
but I wasn't able to follow up on this in a more timely manner.

This PR contains an initial draft of adding an `Exasol` backend to `Ibis`.
I am aware that the PR is not very polished yet and also lacks a significant number of features.
That said, with the PR, I also wanted to establish a means of communication to get some feedback
and align with you while pushing the feature of an Exasol backend forward.

## Status Quo
* An Exasol backend for testing is set up 
  * Starting the database
  * Initializing data
  * ...
* A bare minimum SQLA-based backend is implemented
* A bare minimum of documentation is added
* Tests can be run
  * `just up exasol`
  * `just test exasol`
    * `715 failed, 430 passed, 38 skipped, 26312 deselected, 38 xfailed, 33 errors`

### Questions for/to the Ibis Project
* What features does the Ibis Project require for the MVP for a new backend?
* How do I update/work with the Nix shell when dependencies, etc. are added?
    * `nix-shell` failed after introducing new dependencies.
    * Is there a README or some other source that can help here (except for learning `Nix` :sweat_smile:)?

### Other Questions
* What features does the "Exasol Project Management" consider to be part of an MVP for Ibis (In clarification)?

## To-Do's
Edited:
```
   --- Removed ---
```
It would be great if you could give me some initial feedback on the PR and have a look at
my questions. I appreciate your effort and look forward to hearing from you.

best
Nico


docker/exasol/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM exasol/docker-db:latest-7.1
+RUN yum update -y --exclude=kernel* && \
+    yum install -y nc
```
cpcloud, line 3: Any chance we can avoid the build here? Updating dependencies in Docker can make failures hard to debug, so if possible I'd like to use the image directly instead of having a build step where it is isn't necessary.

docker-compose.yml
```diff
@@ -433,6 +433,24 @@ services:
     volumes:
       - oracle:/opt/oracle/data
 
+  exasol:
+    build: ./docker/exasol
+    image: ibis-exasol
+    privileged: true
```
cpcloud, line 439: Is this required? If so, can you add a short comment about why?

pyproject.toml
```diff
@@ -367,9 +373,10 @@ no_dot = true
 show_deps = true
 
 [tool.codespell]
-# local codespell matches `./docs`, pre-commit codespell matches `docs`
-skip = "*.lock,.direnv,.git,./docs/_freeze,docs/_freeze/**"
-ignore-regex = '\b(i[if]f|I[IF]F|AFE)\b'
+# notebooks are skipped because there's no straightforward way to ignore base64
+# encoded strings
+skip = "*.lock,.direnv,.git,docs/_freeze/**/html.json"
+ignore-regex = '\b(DOUB|i[if]f|I[IF]F|lamduh|AFE)\b'
```
cpcloud, line 379: These lines look a bit old. Do you need to rebase on top of `master`?

pyproject.toml
```diff
@@ -186,7 +184,13 @@ clickhouse = ["clickhouse-connect", "sqlalchemy"]
 dask = ["dask", "regex"]
 datafusion = ["datafusion"]
 druid = ["pydruid", "sqlalchemy"]
-duckdb = ["duckdb", "duckdb-engine", "sqlalchemy", "sqlalchemy-views"]
+duckdb = [
+  "duckdb",
+  "duckdb-engine",
+  "packaging",
+  "sqlalchemy",
+  "sqlalchemy-views",
```
cpcloud, line 192: These changes look like they are from a pre-7.0 commit. I think you need to rebase.

ibis/backends/exasol/tests/conftest.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+def pytest_configure(config):
+    global EXASOL
+    cli_arguments = config.option
+    kwargs = EXASOL_OPTION_GROUP.kwargs(os.environ, cli_arguments)
+    EXASOL = Exasol(**kwargs)
+
+
+def pytest_addoption(parser):
```
cpcloud, line 173: Can you give a short description of why this was needed? Not opposed to it, but we don't have special pytest options for any other backend. Just wondering if this is necessary to make testing exasol easier.

cpcloud (commented):
Thanks for the PR! Glad to hear back from you!

Gave this a once over. Overall, this looks pretty good :)

cpcloud:
To answer some of your questions:

> What features does the Ibis Project require for the MVP for a new backend?

We don't really have a documented bare minimum feature set for new backends.

If I had to say what it was, I think it'd be:

1. `con.table(name)` returns a table expression without failing
2. `con.table(name).execute()` returns a pandas DataFrame without failing

Basically, you should be able to run the equivalent of `SELECT * FROM t` and bring the results back into Python as a pandas DataFrame.

> How do I update/work with the Nix shell when dependencies, etc. are added?

We don't have specific documentation for this. I can add some.

I suspect you need to run `just lock`, and then `direnv reload` (if you're using `direnv`) or exit and re-enter the `nix-shell`.

cpcloud:
Regarding getting the tests to pass, the least painful strategy is to insert

```
pytestmark = pytest.mark.notimpl(["exasol"])
```

at the top of every `test_*.py` file (or insert `"exasol"` into the existing `pytestmark`'s list) and then start implement functionality.

When a test that is marked as `notimpl` passes, you'll see a pytest **failure** with a message like this:

```
[XPASS(strict)] ...
```

Indicating that you've successfully passed the test.

docker-compose.yml
cpcloud: Looked into this a bit.

In theory we should be able to use `cap_add` to specify the list of [capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) the container needs, but in practice the container appears to be using some unnecessary patterns that require `privileged: true` (which gives more privileges than can be specified in `cap_add`)

It looks like the container is trying to write to hugepages using the proc filesystem to verify that hugepages can be allocated. I don't think this is necessary, you should be able to allocate hugepages without needing to use the proc filesystem to check whether it's possible to do so.

The `IPC_LOCK` capability enables this:

image

This isn't a blocker, be we should work with whoever created this image to not require `privileged: true` to run it.

docker/exasol/Dockerfile
cpcloud: I was able to remove this and use the plain image. This was here to add the `nc` binary for a healthcheck. I altered the health check to run a simple SQL query instead, using `exaplus`.

Nicoretti:
Hi @cpcloud,

thanks for all the input and help, I'll pickup on it tomorrow.

best
Nico

ibis/backends/exasol/tests/conftest.py
Nicoretti: First up: after thinking about it, I don't think it is needed within the ibis project.

## Context / Background
* I am working on different `exasol` integrations/projects and therfore I have a need to be flexible regarding the configuration
   * From a single project perspective (ibis) with "a known set of settings" it should not be an issue to hard code this settings. (So I think I just took the wrong perspective on this one)
* The MYSQL and Postgres tests configurations support hard coded settings which can be overwritten via `env`
  * Because I already had some snippets to support `pytest`s configuration mechanism, I was wondering why not use it in the first place.

Nicoretti: I will replace it with the pattern in the `mysql` and `postgress` tests (hard coded + env).

docker-compose.yml
Nicoretti: **TL;DR:** I agree that one does not want the `privileged` flag to be set. Sadly, afik currently it is not possible to run the container without it.

## Details
To the best of my knowledge, there have already been discussions around this issue. However, I'm not sure if there has been any "final" outcome. I'll take your feedback as an opportunity to reignite the conversation about it. I will also take a closer look at how I can contribute to resolving this issue.

Nicoretti:
@cpcloud I assume the final PR will be `squash merged` and therefore did not follow your commit message conventions.
If that's not the case please let me know, then I'll adjust the commit messages.

cpcloud:
@Nicoretti We can clean up the commits once everything is green :) No need to worry about it for now.

ibis/backends/base/__init__.py
```diff
@@ -40,6 +40,8 @@
     "impala": "hive",
     "pyspark": "spark",
     "polars": "postgres",
+    # TODO: not sure if this the closest match
```
Nicoretti, line 43: From the DB Developers I got the feedback, that `Oracle` should be the closest match. That said in case of doubt we generally tend to aim for compliance with the SQL standard. Still we also have some compatibility with other things like e.g. date formats of `MySql` ...

docker-compose.yml
```diff
@@ -433,6 +433,23 @@ services:
...
+    ports:
+      - 8563:8563
+    healthcheck:
+      interval: 10s
+      retries: 9
+      timeout: 90s
+      test:
+        - CMD-SHELL
+        - /usr/opt/EXASuite-7/EXASolution-7.1.23/bin/Console/exaplus -c 127.0.0.1:8563 -u sys -p exasol -encryption OFF <<< 'SELECT 1'
```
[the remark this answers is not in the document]
Nicoretti: @cpcloud

ibis/backends/base/__init__.py
Nicoretti: @cpcloud

docker-compose.yml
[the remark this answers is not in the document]
cpcloud: Ah right. Let's pin. Perhaps in the mean time we can get the image developers to put an alias to `exaplus` somewhere that is version independent?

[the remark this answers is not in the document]
cpcloud: Alternatively, if we know there's only a single `exaplus` lying around we can put a glob here.

ibis/backends/base/__init__.py
cpcloud: Ok, let's leave it as oracle for now. Currently this is used in two major places:

1. printing generated sql
2. turning queries from other dialects into the backend dialect (exasol in this case)

docker-compose.yml
cpcloud: Happy to help in whatever way is needed. Not a blocker for this PR in any case. We can try to resolve this in parallel.

docker-compose.yml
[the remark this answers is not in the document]
cpcloud: We should pin the image version regardless. Renovate will take care of submitting version bump PRs.

docker-compose.yml
Nicoretti: @cpcloud I took some time and was talking to various people regarding the matter of the `--privileged` flag.

The TL;DR is:

* It is known
* It is rated as not ideal (should change)
* The huge pages issue is not the major reason why removing it
  would require a significant amount of effort
* It may be done in future, but there is no time table on it

docker-compose.yml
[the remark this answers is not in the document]
Nicoretti: I will pin it. Also tooling like exaplus may change, so taking care of it during an upgrade makes sense.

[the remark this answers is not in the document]
Nicoretti: **:spiral_notepad: side note**
From what I understood in my recent discussions, up-streaming "custom changes" to the docker image aren't supported by the the current workflow. Changes currently are expected to be done on top of the image (using the container as base image).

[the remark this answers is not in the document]
cpcloud: Ok, that seems reasonable.

The real thing I would like to see is `healthcheck.sh` or some equivalent (maybe a suggested command to run that is always available at a certain path in the container?) in the image's README.

I'm not particularly set on being able to access `exasol` at a particular location, we just need a health check of some sort.

pyproject.toml
Nicoretti: Thanks for pointing that out, I think I screwed this one up while resolving conflicts during the previous rebase.

pyproject.toml
Nicoretti: Thanks for pointing that out, I think I screwed this one up while resolving conflicts during the previous rebase.

Nicoretti:
Hi @cpcloud 

Just did another catchup with your `master` I was wondering if the following `workflow` for addressing merge conflicts in: `poetry.lock`, `requirements-dev.txt`, `poetry-overrides.nix` is correct:

1. Make sure `pyproject.toml` is correct (correctly merged)
2. Remove the conflicted `poetry.lock`
3. Run `poetry lock` or `poetry update`
4. Run `just relock`

cpcloud:
@Nicoretti That looks right!

The only thing I would change is that you should be able to merge steps 3 and 4 into

```
just lock
```

Nicoretti:
@cpcloud sry had to force push, because I had rebased the changes from the `ibis/master` into this branch and pushed it :sweat_smile:.

btw doing a "clean" lock, does not seem to work properly (even on `libis/master`). Poetry does not finish resolving (conflicts).

### Clean Lock
1. `rm poetry.lock`
2. `just lock`

A relock, with an existing `poetry.lock` is working fine though.

cpcloud:
@Nicoretti I am going to rebase this PR and fix up the merge conflicts!

docker-compose.yml
Nicoretti: Hi @cpcloud,

currently I try to upstream a health-check to the official `exasol docker image`. As of now this mostly looks like the code snippet(s) bellow. That said, from what I understood it is very likely that it only will be part of  `exasol docker images` for  versions `>= 8`.

So I was wondering if it would make sense, to reintroduce a "custom" docker image inhering from the exasol one. But this time I would just add the `healthcheck` bellow.  Building such a container should not have much overhead, compared to fetching the "official" one. Also by adapting this, we would have a somewhat "streamlined" `healthcheck` for the exasol DB. 

Just want to get your take on it, please let me now what your preferred solution is (current vs image with health script).

```python
#! /usr/bin/env python3
"""
This script implments a simple health check inteded for use as dockers health check.

For futher details regarding the docker health check,
see also https://docs.docker.com/engine/reference/builder/#healthcheck.
"""
import sys
import shutil
import argparse
from enum import IntEnum
from pathlib import Path
from subprocess import run
from functools import partial, wraps


_stdout = print
_stderr = partial(print, file=sys.stderr)


class Status(IntEnum):
    Healthy = 0
    Unhealthy = 1
    Reserved = 2


class BinaryNotFound(Exception):
    pass


def _find_binary(name: str, basepath: str | Path) -> Path:
    """
    Searches for a binary within the current $PATH and recursivly in the specified directory

    Search Order:
        1. current $PATH
        2. specified fs path

    Args:
        name: of the binary to search for.
        basepath: to search if the binary isn't found in the current $PATH.


    Return:
        Path to the first binary which matches the search criteria.


    Raises:
        BinaryNotFound: The specified binary couldn't be found.
        
    """
    basepath = Path(basepath)

    binary = shutil.which(name)
    if binary is not None and binary != '':
        return Path(binary)

    try:
        binary = next(Path(basepath).rglob(name))
    except StopIteration:
        raise BinaryNotFound(f"Could not find  binary [{name}].")

    return Path(binary)
    

def _health_check(username, password, host, port) -> Status:
    try:
        exaplus = _find_binary("exaplus", "/")
    except BinaryNotFound as ex:
        _stderr(f"{ex}")
        return Status.Unhealthy

    result = run([
        exaplus,
        "-c",
        f"{host}:{port}",
        "-u",
        f"{username}",
        "-p",
        f"{password}",
        "-encryption",
        "OFF",
        "-sql",
        "'SELECT 1;'"
    ], capture_output=True)
    status = Status.Healthy if result.returncode == 0 else Status.Unhealthy 

    return status


def main(argv=None):

    def _parser() -> ArgumentParser:
        parser = argparse.ArgumentParser(
                formatter_class=argparse.ArgumentDefaultsHelpFormatter
        )
        parser.add_argument(
            '--host', type=str, default='127.0.0.1',
            help='to connect to'
        )
        parser.add_argument(
            '--port', type=int, default=8563,
            help='to connect to'
        )
        parser.add_argument(
            '--username', type=str, default='sys',
            help='to authenticate against the db'
        )
        parser.add_argument(
            '--password', type=str, default='exasol',
            help='to authenticate against the db'
        )
        parser.add_argument(
            '--debug', action="store_true", default=False,
            help="disable the 'last resort' exception handler"
        )
        return parser

    def _main(username, password, host, port) -> Status:
        exaplus = _find_binary("exaplus", "/usr/opt")
        return _health_check(username, password, host, port)

    def _protect(f, *args, **kwargs) -> Status:
        
        @wraps(f)
        def wrapper(*args, **kwargs):
            try:
                return f(*args, **kwargs)
            except Exception as ex:
                _stdout(f"Error: {ex}")
                return Status.Unhealthy

        return wrapper

    parser = _parser()
    args = parser.parse_args(argv)

    entry_point = _main
    if not args.debug:
        entry_point = _protect(entry_point)

    return entry_point(
        args.username,
        args.password,
        args.host,
        args.port
    )


if __name__ == '__main__':
    sys.exit(main())
```

```Dockerfile
...
ADD healthcheck.py healthcheck.py
HEALTHCHECK --interval=10s --retries=10 --timeout=90s CMD python3 healthcheck.py
```

Nicoretti:
@cpcloud thx for the updateing

cpcloud:
I can help with the `poetry.lock` conflicts!

Nicoretti:
@cpcloud 
> I can help with the `poetry.lock` conflicts.
Thanks appreciate it, next time I'll give it another try ;).

1. To spare you some time
2. To enable myself (I think, I understood what went wrong the last time)


Side Note(s):
* Sorry for the slow progress, have been dealing with other things not in my direct control
* Next week I won't be able to work on the PR but I'll pick up on it the week after

best
Nico

ibis/backends/exasol/__init__.py
```diff
@@ -61,13 +61,31 @@ def do_connect(
...
+            warnings.filterwarnings("ignore", category=sa.exc.RemovedIn20Warning)
+            return super().list_tables(like=like, database=database)
+
+    def _get_sqla_table(
+        self, name: str, schema: str | None = None, autoload: bool = True, **kwargs: Any
+    ) -> sa.Table:
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=sa.exc.RemovedIn20Warning)
+            return super()._get_sqla_table(name, schema, autoload, **kwargs)
+
```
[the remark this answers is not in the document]
cpcloud: Hm, this should be a pass through I think. The reason to keep this is to allow the exasol dialect, which appears to be making some sqlalchemy 1.x-only calls to do so without ibis tests failing (we enable the failure to ensure ibis can work with both sqlalchemy 1.x and 2.x, since some dialects support both).

What happens if you `export SQLALCHEMY_WARN_20=1` and then run the relevant tests?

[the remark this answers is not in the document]
cpcloud: Long term the exasol dialect should probably be compatible with both major versions of sqlalchemy, many of the 1.x calls are calling `execute` with a `str`, which can be changed to `exec_driver_sql` to support 1.x and 2.x.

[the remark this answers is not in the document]
Nicoretti: My assumption is, it is due to the signature mismatch. Just wanted to get your feedback on it. Also was wondering if I missed something else. The info regarding SQLA 1.x vs 2.x was already quite valuable to me. :+1: 

`BaseAlchemyBackend`
```python
    def _get_sqla_table(
        self,
        name: str,
        *,
        namespace: ops.Namespace = ops.Namespace(),  # noqa: B008
        autoload: bool = True,
        **_: Any,
    ) -> sa.Table:

```

`ExasolBackend`
```python
    def _get_sqla_table(
        self,
        name: str,
        schema: str | None = None,
        autoload: bool = True,
        **kwargs: Any
    ) -> sa.Table:
```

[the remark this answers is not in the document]
Nicoretti: Alright
```python
    def _get_sqla_table(
            self, name: str,
            namespace: ops.Namespace = ops.Namespace(),  # noqa: B008
            autoload: bool = True, **kwargs: Any
    ) -> sa.Table:
        with warnings.catch_warnings():
            warnings.filterwarnings("ignore", category=sa.exc.RemovedIn20Warning)
            return super()._get_sqla_table(name=name, namespace=namespace, autoload=autoload, **kwargs)
```

did the trick ;).

Nicoretti:
Hi @cpcloud,

I am back on the PR. Sorry for the extra delay; I had to address a bug and some other issues in other projects I am maintaining.

Since my catch-up yesterday, I have experienced two issues. I'm just wondering if you may have any idea regarding them.

1. The `nix-shell` does not seem to work the way described [here](https://ibis-project.org/contribute/01_environment) anymore. This is probably due to [PR #7437](https://github.com/ibis-project/ibis/pull/7437). Is this intended?

2. Executing tests seems pretty flaky (not only in parallel mode):
   - `pytest -n 0 --dist no -m exasol`
     ```
     749 failed, 233 passed, 34 skipped, 27246 deselected, 240 xfailed, 24 errors in 350.39s (0:05:50)
     561 failed, 434 passed, 34 skipped, 27246 deselected, 242 xfailed, 9 errors in 442.03s (0:07:22)
     618 failed, 377 passed, 34 skipped, 27246 deselected, 242 xfailed, 9 errors in 465.57s (0:07:45)
     ```
   - `pytest -n 0 --dist loadgroup -m exasol`
      ```
      582 failed, 413 passed, 34 skipped, 27246 deselected, 242 xfailed, 9 errors in 469.88s (0:07:49)
      320 failed, 227 passed, 34 skipped, 27246 deselected, 153 xfailed, 546 errors in 323.05s (0:05:23)
      626 failed, 367 passed, 34 skipped, 27246 deselected, 242 xfailed, 11 errors in 435.29s (0:07:15)
      ```
   - `pytest -n auto -m exasol`
     ```
     512 failed, 483 passed, 34 skipped, 242 xfailed, 9 errors in 323.58s (0:05:23)
     518 failed, 478 passed, 34 skipped, 241 xfailed, 9 errors in 279.37s (0:04:39)
     443 failed, 415 passed, 34 skipped, 200 xfailed, 188 errors in 286.54s (0:04:46)
     ```
   - `just test exasol`
     ```
     533 failed, 462 passed, 34 skipped, 242 xfailed, 9 errors in 253.28s (0:04:13)
     518 failed, 478 passed, 34 skipped, 241 xfailed, 9 errors in 273.90s (0:04:33)
     545 failed, 448 passed, 34 skipped, 243 xfailed, 10 errors in 268.66s (0:04:28)
     ```

I was wondering if this rings a bell or sounds like a problem you've already encountered in the past with other backends.

Thanks for being so supportive all along.

Best,
Nico

Nicoretti:
Hi @gforsyth,

Thanks for the timely update and feedback, appreciate it. The random test execution is a very good point,  I didn't really had it on my radar yet.

Best,
Nico

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -125,6 +126,14 @@ class BaseAlchemyBackend(BaseSQLBackend):
...
     _temporary_prefix = "TEMPORARY"
 
+    def _from_url(self, url: str, **kwargs) -> BaseBackend:
+        """Construct an ibis backend from a SQLAlchemy-conforming URL."""
+        new_kwargs = kwargs.copy()
+        kwargs = {}
+        _, new_kwargs = self.inspector.dialect.create_connect_args(url)
+        new_kwargs = toolz.merge(kwargs, new_kwargs)
+        return self.connect(**new_kwargs)
+
```
Nicoretti, line 136: Hi @cpcloud and @gforsyth,

I stumbled upon this, while fixing issues with the `Exasol` backend.
I think this implementation should be more robust for sqla based backends, because SQLA dialects may adapt `url` specifics ([see here](https://docs.sqlalchemy.org/en/20/core/internals.html#sqlalchemy.engine.default.DefaultDialect.create_connect_args)).

For example, see [asyncpg](https://github.com/sqlalchemy/sqlalchemy/blob/4da59c6f534371d8c76cd728e8a459018e9112b6/lib/sqlalchemy/dialects/postgresql/asyncpg.py#L1106) and other dialects.

If you wan't, I can create an extra PR for just this change. If you think the change is not worth it, I'll drop the change in this PR too.

gforsyth: HI @Nicoretti !  We certainly welcome improvements to make implementations more robust, and calling out to `create_connect_args` seems like a good idea.
We also have some extra handling for reading from environment variables in https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/__init__.py#L37-L70 that we don't want to override completely.

(the fact that sqlalchemy-specific code is included in `base/sql` is a separate problem that we need to fix)

Nicoretti: Hi @gforsyth, 

Thanks for the feedback. I'll omit it from this pull request and will  propose a more comprehensive change in a separate pull request in the next few days.

ibis/backends/exasol/tests/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+    from typing import Any
+
+EXASOL_USER = os.environ.get("IBIS_TEST_EXASOL_USER", "sys")
+EXASOL_PASS = os.environ.get("IBIS_TEST_EXASOL_PASSWORD", "exasol")
+EXASOL_HOST = os.environ.get("IBIS_TEST_EXASOL_HOST", "localhost")
+EXASOL_PORT = int(os.environ.get("IBIS_TEST_EXASOL_PORT", 8563))
+IBIS_TEST_EXASOL_DB = os.environ.get("IBIS_TEST_EXASOL_DATABASE", "EXASOL")
+
+
+class TestConf(UnorderedComparator, ServiceBackendTest, RoundHalfToEven):
```
Nicoretti, line 22: Hi @cpcloud,

While digging into test issues with the Exasol backend, I stumbled upon a few things that got me curious. I was wondering if you may, could shed some light on them.

- **Stable Order in Default Expectations**: Why is the default expectation for a stable order present if sorting isn't explicitly requested (`ORDER BY`)? Is this because most backends tend to provide stable results?
- **Different Ways of Customizing Backend Tests**: What is the rationale behind having two different approaches to customizing backend tests (i.e., using inheritance and using class member-based settings)?

  Why not only use the class members approach? E.g.: 
  ```python
  TestConf(UnorderedComparator, ServiceBackendTest):
       ...
  ```
  VS
  ```python
   TestConf(ServiceBackendTest):
       comparator = UnorderedComparator
  ```
- **Limited Values in string_col of functional_alltypes**: Why does the `string_col` in `functional_alltypes` only contain values from `[0..9]`? This limitation doesn't seem beneficial for testing functions like `CAPITALIZE` or `LOWER`, and it also lacks a wide range of edge cases. Could this be expanded for more comprehensive testing?

Just trying to get a better understanding of the workspace because I feel like I've only grasped a small part of it so far.

Thanks a lot for all the help, I really appreciate it!

best
Nico

gforsyth: Hey @Nicoretti -- I can answer a few of these:

> * **Different Ways of Customizing Backend Tests**: What is the rationale behind having two different approaches to customizing backend tests (i.e., using inheritance and using class member-based settings)?

I don't believe there's a rationale here, they're artifacts of the long evolution of the project.  Further, many (most? all?) of the test class mixins can probably be merged into the base `BackendTest` class.  



> * **Limited Values in string_col of functional_alltypes**: Why does the `string_col` in `functional_alltypes` only contain values from `[0..9]`? This limitation doesn't seem beneficial for testing functions like `CAPITALIZE` or `LOWER`, and it also lacks a wide range of edge cases. Could this be expanded for more comprehensive testing?

It should _absolutely_ be expanded for more comprehensive testing.

cpcloud: To answer the first question:

> Stable Order in Default Expectations

Many backends do provide a stable ordering, but some don't and you can override the `assert_series_equal` and `assert_frame_equal` methods on `BackendTest` as you like. The BigQuery backend is an example of a backend that doesn't have a stable ordering without `ORDER BY` and overrides those methods appropriately.

docker-compose.yml
Nicoretti: I'll leave the `health-check` in this PR as is. Up-streaming it into the official container is still ongoing.
Still it is already clear that it only will be part of version(s) `>= 8.x.y`. Therefore, I'll address the change of the `health-check` when upgrading the DB-Backend version.

cpcloud:
@Nicoretti Is there any chance `sqlalchemy-exasol` can be made compatible with the latest sqlalchemy release that is <2 (1.4.50)?

Python 3.12 support is coming up and the sqlalchemy version that's supported by `sqlalchemy-exasol` is now constraining us to a version that doesn't provide binary wheels for python 3.12.

cpcloud:
Generally speaking it should be fine to set sqlalchemy constraints to allow patch versions, they tend not to break things in patch releases, so instead of the `<1.4.45` bound that's in `sqlalchemy-exasol` it would be fine to make that `<1.5` if you're not comfortable making it `<2`.

cpcloud:
I am running the test suite now and will help grind through the current failures. I'd like to avoid long running PRs if possible 😅

Nicoretti:
> Generally speaking it should be fine to set sqlalchemy constraints to allow patch version bumps, they tend not to break things in patch releases, so instead of the `<1.4.45` bound that's in `sqlalchemy-exasol` it would be fine to make that `<1.5` if you're not comfortable making it `<2`.

@cpcloud Generally, yes. I need to re-validate it against the test suite of that `sqlalchemy` version though. I'll take care of it this week. I hope that fits your schedule.

cpcloud:
@Nicoretti It looks like your approach so far is to go through every test file and implement functionality until tests pass.

I highly recommend a different approach that will get this merged upstream much faster:

Mark everything that is currently failing as `pytest.mark.notimpl` and create follow-up pull requests to add functionality.

This also has the side benefit of user-facing release notes that show when particular functionality was added.

cpcloud:
> > Generally speaking it should be fine to set sqlalchemy constraints to allow patch version bumps, they tend not to break things in patch releases, so instead of the `<1.4.45` bound that's in `sqlalchemy-exasol` it would be fine to make that `<1.5` if you're not comfortable making it `<2`.
> 
> @cpcloud Generally, yes. I need to re-validate it against the test suite of that `sqlalchemy` version though. I'll take care of it this week. I hope that fits your schedule.

Sounds good to me!

Nicoretti:
> I am running the test suite now and will help grind through the current failures. I'd like to avoid long running PRs if possible 😅

Thanks, no worries. I've also been in and out with the PR at times. So, don't worry, working through it from my side has been fine so far. Some of the recent test issues stemmed from execution order and implicit schema changes in the Exasol DB, which took me a bit to figure out.

I hope our interaction has been smooth for you too (don't want to unnecessarily bother you with irrelevant questions, etc.).

Best regards,
Nico

Nicoretti:
> @Nicoretti It looks like your approach so far is to go through every test file and implement functionality until tests pass.
> 
> I highly recommend a different approach that will get this merged upstream much faster:
> 
> Mark everything that is currently failing as `pytest.mark.notimpl` and create follow-up pull requests to add functionality.
> 
> This also has the side benefit of user-facing release notes that show when particular functionality was added.

100% Agreed. (wasn't  a 100% sure if that's good enough from ibis side).

cpcloud:
Yeah, we tend to prefer that approach since it can get in the hands of users more quickly and possibly help prioritize particular functionality (time series, strings, etc)

Nicoretti:
> Yeah, we tend to prefer that approach since it can get in the hands of users more quickly and possibly help prioritize particular functionality (time series, strings, etc)


> Yeah, we tend to prefer that approach since it can get in the hands of users more quickly and possibly help prioritize particular functionality (time series, strings, etc)


Great :+1:, to align an make sure we are on the same page, here is current plan of action:
(in that order)

* Finish my current task (not ibis, not sqla related)
* Detour to our sqlalchemy dialect to upgrade the `sqla` constrain
*  Finish up a few smaller non ibis related tasks
* Come back to ibis

cpcloud:
Solid plan, I like it!

cpcloud:
I am working through adding `xfail` markers to `test_temporal.py` right now, let's try avoid any duplicate work there :)

cpcloud:
Ok, `test_temporal.py` is done, I am now working on `test_aggregation.py`

Nicoretti:
@cpcloud I started upgrading (loosening the sqla constraint) for `sqlalchemy-exasol`. There are a few (hopefully minor) issues which need to be looked at. Once they are addressed and the [PR](https://github.com/exasol/sqlalchemy-exasol/pull/401) is merged, I'll create a release and `relock` this PR.

ibis/backends/tests/test_set_ops.py
```diff
@@ -115,6 +122,8 @@ def test_intersect(backend, alltypes, df, distinct):
         param(True, id="distinct"),
     ],
 )
+# Question: why doesn't it get green without strict=False? (smelly)
```
Nicoretti, line 125: @cpcloud do you have any insights on this? If not I can dig deeper on this later on (follow up PRs).

Nicoretti:
## Status Update

### TL;DR
I have updated and addressed most things in the PR and will wait for feedback.

---

### Detailed

As recently discussed with @cpcloud, I focused efforts to get the basic Exasol dialect PR done by disabling (`xfail`) missing features (which will be provided via follow-up PRs). In order to do this, I went through the tests and marked failing tests with `xfail`. For some (`12`), I needed to use `xfail(..., strict=False)` because they were a result of `parameterize` combinations. As far as I know, this could only be resolved by unrolling the `parameterize` combinations. Even though I personally feel/think that unrolling would be the preferable solution, I fear it will also increase the likelihood of tedious conflicts significantly, due to the fact that the entire parameter setup needs to be adjusted. So, in favor of a timely and "straightforward" merge, I opted for some `xfail(..., strict=False)`. If one or all of the usages do not match with your take on how this should be handled, please let me know.

### Other open/pending issues:

- The windows CI job is failing for Exasol:
  - The reason looks like to be that the DB image is Linux-based:
    - e.g. `Exasol window latest` `"Linux" cannot be used on this platform:...` (The Exasol DB only runs on Linux)
      - What would be a viable strategy here (from your point of view)?
- A few other jobs within the backends workflow also fail:
  - Do you maybe have an idea or hunch what it could be?
    - Is there something in my changes which affects those?
    - Is it likely that something during a rebase went wrong?
      (In that case, I could go and close diff the failing files)

I'll be happy to push the PR forward, but from my current point of view, I would wait for feedback before I continue
to also minimize the problem of a moving target.

Still, take your time; there is no need to rush. I just wanted to let you know what my current status regarding the PR is so we can align and avoid unnecessary work, downtimes, etc.

Hope that fits with you.

Best,
Nico

cpcloud:
Taking a look now! This looks pretty close.

cpcloud:
@Nicoretti I went through and had to make a few changes

- https://github.com/ibis-project/ibis/pull/7303/commits/029fa4098005553be69c41598646b80de98b963e: it looks like you might've deleted some CI setup unintentionally, so I revert those changes
- https://github.com/ibis-project/ibis/pull/7303/commits/78b9683e738530c7f9ac5a142b7c2f3490a23780: we're not testing any of our service-based backends on windows, last I checked none of the containers can work as is on windows.
- https://github.com/ibis-project/ibis/pull/7303/commits/2330e9a721025865771c9061cc075156c45e0f5c: the change to `gen_name` is bit annoying, but I understand it. I fixed an issue with oracle quoting of non-underscore-prefixed names

ibis/backends/exasol/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+    ) -> ir.Table:
+        raise NotImplementedError()
+
+    def read_csv(
+        self, path: str | Path, table_name: str | None = None, **kwargs: Any
+    ) -> ir.Table:
+        raise NotImplementedError()
+
+    def _metadata(self, query: str) -> Iterable[tuple[str, dt.DataType]]:
+        # table names can't start with underscore
```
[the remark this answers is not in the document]
Nicoretti: Mh good point, don't recall why it ended up here. I will remove it.

ibis/backends/exasol/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def read_parquet(
+        self, path: str | Path, table_name: str | None = None, **kwargs: Any
+    ) -> ir.Table:
+        raise NotImplementedError()
+
+    def read_csv(
+        self, path: str | Path, table_name: str | None = None, **kwargs: Any
+    ) -> ir.Table:
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
Nicoretti: Yes, was rather a marker for myself. Initially, it wasn't immediately clear to me that the test suite has nicely marked everything. I will remove it.

ibis/backends/exasol/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def list_tables(self, like=None, database=None):
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=sa.exc.RemovedIn20Warning)
+            return super().list_tables(like=like, database=database)
+
+    def _get_sqla_table(
+        self,
+        name: str,
+        namespace: ops.Namespace = ops.Namespace(),  # noqa: B008
```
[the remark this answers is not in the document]
Nicoretti: You are right, in case of doubt the `**kwargs` in the forward should take care of this.
I'll update the signature and the `super-class` call.

Nicoretti:
Hi @cpcloud,

thanks for the review, adjustments and help :heart: . Addressed or responded to the open points.

best
Nico

ibis/backends/exasol/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+        **kwargs: Any,
+    ) -> sa.Table:
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=sa.exc.RemovedIn20Warning)
+            return super()._get_sqla_table(
+                name=name, namespace=namespace, autoload=autoload, **kwargs
+            )
+
+    def _register_in_memory_table(self, _: ops.InMemoryTable) -> None:
+        raise NotImplementedError(self.name)
```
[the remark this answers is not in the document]
cpcloud: It looks like there's also https://github.com/exasol/pyexasol/blob/master/docs/REFERENCE.md#import_from_pandas which might be interesting to investigate as a follow up.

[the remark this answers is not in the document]
Nicoretti: @cpcloud good point. From what I understand, the [import_parallel](import_parallel()) method is more efficient for larger datasets when compared to [import_from_pandas](https://github.com/exasol/pyexasol/blob/542fa7b236be94242e3205e64665f6d70f2b22e8/pyexasol/callback.py#L69). Still, we can discuss which would be preferable.

ibis/backends/exasol/tests/conftest.py
```diff
@@ -0,0 +1,122 @@
...
+    supports_structs = False
+    supports_json = False
+    supports_map = False
+    reduction_tolerance = 1e-7
+    stateful = True
+    service_name = "exasol"
+    supports_tpch = False
+    # Question/NOTE: imports do not seem to work yet, if commented in tests will bes skipped
+    # deps = "sqlalchemy", "sqlalchemy-exasol"
+    deps = []
```
Nicoretti, line 44: @cpcloud should we address this in this PR?

Nicoretti: I see, my bad. I think I did not realize that it is the actual package, rather than the PyPI package name.

cpcloud:
Clouds are passing:

```
…/ibis on  exasol-backend is 📦 v7.1.0 via 🐍 v3.10.13 via ❄️  impure (ibis-3.10.13-env)
❯ pytest -m 'bigquery or snowflake' -n auto --dist loadgroup --snapshot-update -q
bringing up nodes...
.........x...x.............x....x...x.......xx.....x...................x.........x.x..........x..x.......x.x.x..x............x.......x..xx..x.x.xxxx.x.x.xxxxx.......xxx..x.x....x..xxx......x... [  6%]
x..x..x.......x.x....x......x.......x......x........x.....x..x.............x..xx..x..xx....x.....x.xxx.x....x..........x.......x....xx.x.xx...x....xx......xx.....x......x...x...x..xxx...x...... [ 12%]
..xx.xx.......x.x.........x........x.................s.......x....x.x....xx..x...x.xxxx........x......x.xx........x......................xx.............................x..x.....s...x........... [ 18%]
.........................x.................................x..x.x......x.xx..x.x..x..x....x...x....xx........x..xx.xx..x......x...x..x.....x.............x.......x.....xx.........xx......x....x. [ 25%]
.......x....x........xx...x..x......x.x...x....x...........x.x..x.x..xxx.....x....x...x.xx....x.x.x.x..........x...x..............x.............x..x..x.....x.x........xx.xx..........x..x....... [ 31%]
..x.x.x.....x...xx.xx...x......x..xx..x......................................................x..............x......x........x..xx.x.......x....x....x...x.......x.......x.........xx..x.....x.x.x [ 37%]
........................................x....................xx..........x..x...................x....................xx..ssssssssssssssssssssss..xx..xx.....x............x..x..x...x........x.... [ 44%]
.....xx......xx.....x...........xx...xx...xxx..........x..x...........xx...x.xx.x...xxxx.x.xxx..x.x...x.xx..........xxx...xxxxxxxxxxxxxxxxxx.xxx.xxx..x..xx...x...x..x...x..x...x..x........x.... [ 50%]
.....x..........xxx.x.x.............x.x....xx...............x...........xx.......x....x...x........x.......xx............x..x....x..........xx.........x..xx.x.s.x..............x....sx.......... [ 56%]
.x..........s..x........xx.....x....................x..s...............x.........................x..............xx...x...x.....................x..x..x....x..x.........x.....xx....x............. [ 63%]
.x........xxx....x.x.................xx..x.....x......x....x.....x..x.......x...x......x....xx....x.xx...............x......x...x.....xx.x.....xx.xx..x....x.........x.....x.xx...x.x....x.x..x.. [ 69%]
..x................x...x......xx....x..x....x............x...................................................................x................................x.x..x.x..........x..............x. [ 75%]
.........x...............................x........x...x.....................................................................................................x............x.......x............... [ 81%]
...............................................................x........................x.............................................................................x.....x.x.xx.......x.x..xx. [ 88%]
x................x......x......x................................................................................................................................................................. [ 94%]
........................................xs............................x................................x.................s............x...............................                            [100%]
2582 passed, 30 skipped, 449 xfailed in 459.64s (0:07:39)
```

ibis/backends/base/__init__.py
```diff
@@ -52,6 +52,8 @@
     # druid allows double quotes for identifiers, like postgres:
     # https://druid.apache.org/docs/latest/querying/sql#identifiers-and-literals
     "druid": "postgres",
+    # closest match see https://github.com/ibis-project/ibis/pull/7303#discussion_r1350223901
+    "exa.websocket": "oracle",
```
gforsyth, line 56: Here you state that `oracle` is the closest matching `sqlglot` dialect, but all the sqlglot calls in `exasol/__init__.py` are using the `postgres` dialect.  I think one of these needs to change -- should this be `postgres` here?  Or should the `dialect=` kwargs be `oracle`?

ibis/backends/tests/test_generic.py
```diff
@@ -489,7 +497,19 @@ def test_dropna_invalid(alltypes):
 
 @pytest.mark.parametrize("how", ["any", "all"])
 @pytest.mark.parametrize(
-    "subset", [None, [], "col_1", ["col_1", "col_2"], ["col_1", "col_3"]]
+    "subset",
+    [
+        None,
+        # The combination with any works but not the one with all
+        # Question: Is there a way to specifically suppress subset1-all?
```
gforsyth, line 504: Unfortunately not.  You'll need to unroll the nested parametrization to xfail the specific combination.

Nicoretti: Thanks, for the answer and also good catch (a leftover comment I've missed :sweat_smile:). I will remove it.

ibis/backends/base/__init__.py
Nicoretti: Unfortunately, I can't provide a link to my previous conversation with @cpcloud due to GitHub's limitation in displaying outdated comments during force rebasing.

To summarize our discussion: historically, @cpcloud generously assisted me by providing implementations using `postgress` for certain code segments. Later, we delved into identifying the closest SQL dialect supported by `sqlglot`. During this, I consulted our DB developers to determine the most similar dialect.

In short: Oracle appears to be the closest match, as per their insights. However, navigating dialects is complex, especially with Exasol, where support for specific syntax constructs varies across dialects. Hence, while there might not be an official closest match, the consensus from our DB developers leans toward Oracle.

Nicoretti:
@gforsyth 
> Are there any hints as to the cause of the flakiness?

One issues which caused failing tests in serial and parallel execution is fixed and noted in
the `create_schema` function (see snipet below).

```python
...
    # Exasol implicitly opens the created schema, therefore we need to restore
    # the previous context.
    action = (
        sa.text(f"OPEN SCHEMA {open_schema}")
        if open_schema
        else sa.text(f"CLOSE SCHEMA {name}")
    )
    con.exec_driver_sql(action)
```

As of today serially executed the tests should not be falky, but in regard to parallel execution there still seems to be
at least one or more issues to be identified and addressed.

Nicoretti:
Having rebased against, the latest master a backend test (`flink`) is failing during install.
I have seen similar things in the past, it potentionaly have been an `network` issues during the `pip install`.

Nicoretti:
Just rebased, `oracle` keeps failing, not entirely sure why this is (how it relates to my changes). 
Any ideas?

cpcloud:
Bringing in the upstream changes from 

https://github.com/ibis-project/ibis/compare/7f41bf29e38665c9a2cd323e4a6e09d144e22e2e..90aee1a05296526ee1a034327699e5bea0a3cb8b#diff-0090401604eb2e0aa1e71305358a1c922ea42f8119a64c91f9e5038b58e73984L233

broke the oracle tests.

I had to change the type inference related code for oracle in PR due to the change to `gen_name`. In the meantime, there was a change to this bit of code upstream to fix a [heisenbug](https://heisenbug.lol).

I'll fix it back up.

cpcloud:
It looks like you also reverted https://github.com/ibis-project/ibis/compare/90aee1a05296526ee1a034327699e5bea0a3cb8b..fc7c9a6e46a2bc769df583d0f044c22146daee9b#diff-70e8f790a7cd8392a8b812c47a95be84f3835051e37a873536cefb3cb726d360L97

which is an important change to ensure that we don't break BigQuery due to Exasol's inability to create a table whose name starts with `_`.

I'll add this back too.

Nicoretti:
> Bringing in the upstream changes from
> 
> https://github.com/ibis-project/ibis/compare/7f41bf29e38665c9a2cd323e4a6e09d144e22e2e..90aee1a05296526ee1a034327699e5bea0a3cb8b#diff-0090401604eb2e0aa1e71305358a1c922ea42f8119a64c91f9e5038b58e73984L233
> 
> broke the oracle tests.
> 
> I had to change the type inference related code for oracle in PR due to the change to `gen_name`. In the meantime, there was a change to this bit of code upstream to fix a [heisenbug](https://heisenbug.lol).
> 
> I'll fix it back up.

Thanks for your help! Heisen bugs ... had my fair share of those :sweat_smile:. 


> It looks like you also reverted https://github.com/ibis-project/ibis/compare/90aee1a05296526ee1a034327699e5bea0a3cb8b..fc7c9a6e46a2bc769df583d0f044c22146daee9b#diff-70e8f790a7cd8392a8b812c47a95be84f3835051e37a873536cefb3cb726d360L97
> 
> which is an important change to ensure that we don't break BigQuery due to Exasol's inability to create a table whose name starts with `_`.
> 
> I'll add this back too.

Thanks, my bad. I assumed that I have accidentally overwritten changes from the `upstream (master)` and therefore reverted some changes of mine to the latest `master`.

docs/backends/exasol.qmd
```diff
@@ -0,0 +1,96 @@
...
+```
+
+1. Adjust connection parameters as needed.
+
+## `conda`
+
+Install for Exasol:
+
+```{.bash}
+conda install -c conda-forge ibis-exasol
```
gforsyth, line 34: @Nicoretti -- I will try to remember to add this to the `conda-forge` recipe when we do our next release, but you should feel free to ping me to remind me.

Also, if you can add `sqlalchemy-exasol` to `conda-forge` that will greatly improve the turnaround time on getting `ibis-exasol` added.

gforsyth:
What's going on with the coverage warnings in `exasol/__init__.py`?  Are we failing to upload something?  Because there's no way the test suite is running and we aren't hitting `_get_sqla_table`

docs/backends/exasol.qmd
Nicoretti: > @Nicoretti -- I will try to remember to add this to the `conda-forge` recipe when we do our next release, but you should feel free to ping me to remind me.

Removing it from the docs would be a possibility too. I just used another backend documentation as template, wasn't aware about all implications. But if we want to keep it, I can over to create a ticket as reminder.

> Also, if you can add `sqlalchemy-exasol` to `conda-forge` that will greatly improve the turnaround time on getting `ibis-exasol` added.

Previously, `sqlalchemy-exasol` was available as a `conda-forge` package, but it was discontinued, possibly due to being outdated and underused. I need to determine if we should reintroduce it. Given that we now have a concrete use case with Ibis, it seems more likely that its reintroduction would be accepted.

Nicoretti:
> What's going on with the coverage warnings in `exasol/__init__.py`? Are we failing to upload something? Because there's no way the test suite is running and we aren't hitting `_get_sqla_table`

@gforsyth 
Good question, I just validated `_get_sqla_table` `(ibis/backends/exasol/__init__.py#L147-L148)` gets hit repeatedly.
Not sure on the sonar cloud part, I don't have to much experience with it yet.

gforsyth:
> @gforsyth
> Good question, I just validated `_get_sqla_table` `(ibis/backends/exasol/__init__.py#L147-L148)` gets hit repeatedly (when the suit is executed locally). Not sure on the sonar cloud part, I don't have to much experience with it yet.

Looks like @cpcloud fixed it!

docs/backends/exasol.qmd
gforsyth: I can help maintain the feedstock (what's one more on a big pile :grin:?) -- I think it's a good idea to have it available on conda-forge, if only so that `exasol` isn't the only backend that you can't install with conda/mamba.

gforsyth: recipe PR here: https://github.com/conda-forge/staged-recipes/pull/24777

gforsyth: the underscore threw me when I was looking before.  looks like the conda-forge package is available and up-to-date, so that part is taken care of.

cpcloud (approved):

cpcloud:
@gforsyth This LGTM! Anything else left to do?

gforsyth (approved):
This is great @Nicoretti!  Nothing blocking, just a few notes and we can tackle them in a follow-up, too.

ibis/backends/tests/test_client.py
```diff
@@ -251,6 +251,7 @@ def tmpcon(alchemy_con):
     ["mssql"],
     reason="mssql supports support temporary tables through naming conventions",
 )
+@mark.notimpl(["exasol"], reason="Exasol does not support temporary tables")
```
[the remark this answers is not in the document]
Nicoretti: This currently would be a `never` because it is not supported by Exasol and I don't know if there are any plans to do so in the near future.

ibis/backends/base/__init__.py
```diff
@@ -33,16 +33,16 @@
...
 
-
 _IBIS_TO_SQLGLOT_DIALECT = {
     "mssql": "tsql",
     "impala": "hive",
     "pyspark": "spark",
     "polars": "postgres",
     "datafusion": "postgres",
+    # closest match see https://github.com/ibis-project/ibis/pull/7303#discussion_r1350223901
+    "exa.websocket": "oracle",
```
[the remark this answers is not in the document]
Nicoretti: When I added it, the consistency with the entry in `_SQLALCHEMY_TO_SQLGLOT_DIALECT` somehow seemed to make sense. But thinking about it now ... you are right, good catch thx.

ibis/backends/tests/test_generic.py
```diff
@@ -489,7 +497,17 @@ def test_dropna_invalid(alltypes):
 
 @pytest.mark.parametrize("how", ["any", "all"])
 @pytest.mark.parametrize(
-    "subset", [None, [], "col_1", ["col_1", "col_2"], ["col_1", "col_3"]]
+    "subset",
+    [
+        None,
+        param(
+            [],
+            marks=pytest.mark.notimpl(["exasol"], raises=ExaQueryError, strict=False),
```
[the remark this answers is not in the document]
Nicoretti: From what I can tell `exasol-substet1-all` is reliably failing, due to the lack of support of `?` within the where clause:
```
pyexasol.exceptions.ExaQueryError:  
 (
     message     =>  where clause must be a boolean expression (Session: 1785240712777826304)
     dsn         =>  localhost:8563 
     user        =>  sys
     schema      =>  EXASOL
     session_id  =>  1785240712777826304
     code        =>  42000
     query       =>  
```                     
```sql
WITH t0 AS (SELECT t2."id"                                                                            AS "id",
                   t2."bool_col"                                                                      AS "bool_col",
                   t2."tinyint_col"                                                                   AS "tinyint_col",
                   t2."small_int"                                                                     AS "small_int",
                   t2."int_col"                                                                       AS "int_col",
                   t2."bigint_col"                                                                    AS "bigint_col",
                   t2."float_col"                                                                     AS "float_col",
                   t2."double_col"                                                                    AS "double_col",
                   t2."date_string_col"                                                               AS "date_string_col",
                   t2."string_col"                                                                    AS "string_col",
                   t2."timestamp_col"                                                                 AS "timestamp_col",
                   t2."year"                                                                          AS "year",
                   t2."month"                                                                         AS "month",
                   CASE WHEN (t2."int_col" = ?) THEN NULL ELSE t2."float_col" END                     AS col_1,
                   CASE WHEN (t2."int_col" = ?) THEN NULL ELSE t2."float_col" END                     AS col_2,
                   CASE WHEN (t2."int_col" = ? OR t2."int_col" = ?) THEN NULL ELSE t2."float_col" END AS col_3
            FROM functional_alltypes AS t2)
SELECT t1.col_1, t1.col_2, t1.col_3
FROM (SELECT t0.col_1 AS col_1, t0.col_2 AS col_2, t0.col_3 AS col_3
      FROM t0) AS t1
WHERE ? )
```
That said, I don't know if this only is a limitation of the web socket based driver/protocol, or rather a general one.
@gforsyth I can look more into the details If you want me too.

ibis/backends/tests/test_temporal.py
```diff
@@ -2447,6 +2505,7 @@ def test_date_scalar_from_iso(con):
     raises=sa.exc.DatabaseError,
     reason="ORA-22849 type CLOB is not supported",
 )
+@pytest.mark.notimpl(["exasol"], raises=AssertionError, strict=False)
```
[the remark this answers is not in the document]
Nicoretti: I think you are maybe on to something here. Just removed the marker locally and ran the test multiple times: 

* 10x test_temporal suit only 
  * Always Green
* 5x Full backend test suite
  * Always green

cpcloud:
Bombs away! Thanks for your work @Nicoretti, look forward to working with you more on this!

Nicoretti:
> Bombs away! Thanks for your work @Nicoretti, look forward to working with you more on this!

Looking forward to keep working with you too!

PR #7304, opened 2023-10-06
feat(ux): render url strings as links in rich table output

### LINKED ISSUE #7292: feat: clickable hyperlinks in rich repr

I have a column of URLs. It would be awesome if they were clickable in the printed table repr from rich.

not sure how to do this.

Call [`.protocol()`](https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.protocol) on all string columns, and if it is one of http,https,[something else?], then surround that string with the [rich [link] tags](https://rich.readthedocs.io/en/stable/markup.html#links)? Should this be done on a per-row basis, or if at least x% of rows match this condition then style all rows like this? Not sure what the performance implications of this would be, or how much we need to worry about it.

main

NA

### Code of Conduct

cpcloud:
Render strings that have at least a scheme and netloc as clickable links in the terminal. This PR does not avoid rendering strings as links if they are truncated. Closes #7292.

### DIFF
--- a/ibis/expr/types/pretty.py
+++ b/ibis/expr/types/pretty.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import datetime
 from functools import singledispatch
 from math import isfinite
+from urllib.parse import urlparse
 
 import rich
 from rich.align import Align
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
Seems fine to me -- did you mean to open this as a draft?

cpcloud:
Was going to add a test.

cpcloud:
Test added.

cpcloud:
This PR won't render links that are shorter than the configured maximum display length for strings. That's different than truncation happening to fit more columns into the display. This PR is not handling the second, and the first can they aren't rendered as links because the full string wouldn't show up anyway, so the link wouldn't be valid in general. It might happen to truncate at a valid link, but that wouldn't be the link in the data.

PR #7308, opened 2023-10-06
feat(ux): fix long links, add repr links in vscode

NickCrews:
Followup to https://github.com/ibis-project/ibis/pull/7304

Before, long links would take you to the truncated version, eg https://www.super.long.url.that.is.trucated.com
would both display and link to
https://www.super.long.url.tha...

Now it just renders as that, but links to the full url.

This also makes links render in vscode notebook environment. Before, links were rendering correctly in ipython console. But, they weren't
showing up as links in vscode notebook environment either. This commit adds vscode functionality.

### DIFF
--- a/ibis/expr/types/pretty.py
+++ b/ibis/expr/types/pretty.py
@@ -126,6 +126,7 @@ def _(dtype, values):
     for v in values:
         v = str(v)
         if v:
+            raw_v = v
             if len(v) > max_string:
                 v = v[: max_string - 1] + "…"
             v = v[:max_string]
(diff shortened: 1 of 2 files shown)


ibis/expr/types/pretty.py
```diff
@@ -139,12 +140,11 @@ def _(dtype, values):
                 v = "".join(
                     f"[dim]{repr(c)[1:-1]}[/]" if not c.isprintable() else c for c in v
                 )
-            elif len(v) <= max_string:
-                url = urlparse(v)
-                # check both scheme and netloc to avoid rendering e.g.,
-                # `https://` as link
-                if url.scheme and url.netloc:
-                    v = f"[link]{v}[/link]"
+            url = urlparse(raw_v)
```
NickCrews, line 143: Maybe I'm misunderstanding, but I didn't see the need for the previous `elif` check, so I removed it. please correct me if I missed something here.

cpcloud (approved):
Yep I think you're right! I wasn't aware of the link= syntax. Thanks for the fix!

PR #7312, opened 2023-10-08
fix(flink): correct the filtered count translation

deepyaman:
This will fix something like `ibis/backends/tests/test_aggregation.py::test_reduction_ops[flink-is_in-count_star]`.

Nit: I also took the liberty of changing `count` to `COUNT`, so that the translated SQL wouldn't be a mix of lowercased keywords and capitalized keywords. I don't know if there's ever been thought around standardizing this; maybe it only bothers me. :)

~This may not be the right long-term solution (`FILTER` syntax has some advantages), but at least this works.~ Update: I went ahead and just used `FILTER` syntax instead, I thought it would be a bigger deal.

### DIFF
--- a/ibis/backends/flink/registry.py
+++ b/ibis/backends/flink/registry.py
@@ -18,13 +18,12 @@ operation_registry = base_operation_registry.copy()
 
 
 def _count_star(translator: ExprTranslator, op: ops.Node) -> str:
-    # TODO(deepyaman): Use `FILTER` syntax; see note on `_filter` below.
     if (where := op.where) is not None:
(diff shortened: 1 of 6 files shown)


ibis/backends/flink/registry.py
```diff
@@ -18,7 +18,12 @@
 
 
 def _count_star(translator: ExprTranslator, op: ops.Node) -> str:
-    return "count(*)"
+    if (where := op.where) is not None:
+        condition = f" FILTER (WHERE {translator.translate(where)})"
```
cpcloud, line 22: Really glad to see Flink supports this syntax!

cpcloud (approved):
Sweet!

PR #7315, opened 2023-10-09
feat(common): make `Slotted` and `FrozenSlotted` pickleable

kszucs:
It also add unit tests for those two base classes.


cpcloud (changes requested):
This LGTM, but I believe this should definitely target `master`.

kszucs:
Updated the base branch.

kszucs (commented):
+1

PR #7316, opened 2023-10-09
feat(common): support `Self` annotations for `Annotable`

kszucs:
`Annotable` haven't been supporting `Self` annotations so far, mostly because is that the type being constructed is not available inside `AnnotableMeta.__new__()` yet. This can be overcome by deferring the type loading using a `LazyInstanceOf` pattern with a `ForwardRef`.

Possible improvement: `LazyInstanceOf` should have an alternative implementation for single type use, currently it is using `lazy_dispatched` which is tailer for a slightly different use case.


cpcloud:
This doesn't seem to be breaking any user-facing APIs. Can you target `master`?

kszucs:
@cpcloud updated

cpcloud (approved):

Issue #7280, opened 2023-10-02
feat: consolidate ibis.case(), Value.case(), Value.cases(), Value.substitute()


NickCrews:
### Is your feature request related to a problem?

All 4 functions are basically doing the same thing. It feels to me like we should be able to merge them all into once function with the same API. It is confusing to users (including me!) which they should use, and is extra burden keeping the docstrings etc all correct.

Also, a long-running grossness of .case() is that it returns a CaseBuilder object, which is undocumented. I don't think we want to add this CaseBuilder object to the public API, I think it would be totally doable and much simpler if we could keep it to a single function call.

### Describe the solution you'd like

I propose:

1. removing ibis.case() and Value.case()
2. keeping Value.cases() and Value.substitute(), and possible adding ibis.cases() (but I'm +0 on this). All of these have an API of

```python
def cases(value: Value, mapping: Mapping[Any, Any] | Iterable[tuple[Any, Any]], default: Any = None):
   ...
```
(not sure about `default` vs `else_`, and if `mapping` is the best name)

where the keys can be

- literals in which case they are compared by equality eg `"foo"` or `5`,
- Values, in which case they are compared by equality, eg `value.upper()`
- Deferreds, in which case they are bound to `value` and evaluated eg `_.upper() == _`
- None, ibis.null(), etc. this case currently isn't handled at all, this should be supported, even if we don't do the rest of these things.
- unary functions that take `value` and return one of the above eg `lambda val: val.upper() == "FOO"` or `lambda val: val.upper()`

The values can be
- literals
- deferred projections eg `_.lower()`
- unary functions that take `value` and return one of the above eg `lambda val: val.lower()`

I'm trying to think if there is some ambiguity if we accept both Values and BooleanValues similar to the ambiguity with selection vs filtering in `t[_.x.isnull()]`, but I can't think of it.

I want to support the usecase of chained comparing one column to another:

`(t.x == t.y).ifelse("same", (t.x > t.y).ifelse("greater", "less"))`

I think this would be done with

`t.x.case({lambda x: x == t.y: "same", lambda x: x > t.y: "greater"}, default="less")`

but this feels a little gross. Makes me think either a top level or a table base API would be better:

`t.cases({_.x == _.y: "same", _.x > _.y: "greater"}, default="less")`
`ibis.cases(t, {_.x == _.y: "same", _.x > _.y: "greater"}, default="less")`

I would want to make this

[...]

PR #7319, opened 2023-10-09
feat: remove .case(), move to .cases()

NickCrews:
Fixes https://github.com/ibis-project/ibis/issues/7280

This is still WIP for getting all tests to pass, but putting something concrete out here for feedback. I would love you initial thoughts here.

TODO:
- [x] core tests pass
- [ ] backend tests pass
- [ ] document

Questions:
- Don't remove .case(), just deprecate it?
- don't even hard deprecate it, merely soft-deprecate it by removing from examples and docs? It's not _that_ hard to transition to the new API, but it also wouldn't be that hard for us to stealthily continue supporting the old API.
- rename SearchedCase to just Case, since SimpleCase is now gone?
- make deferreds on non-Table values work: `t.values.cases((_ > 5), "big"), else_="small")`. This doesn't work currently either, so we're not moving backwards.
- The generated SQL isn't as idiomatic for some dialects (eg duckdb) when you pass in values. Eg t.values.cases((5, "five"))` generates
```sql
CASE
    WHEN values = 5 THEN "five"
```
instead of what it used to:
```sql
CASE values
   WHEN 5 THEN "five"
```
IDK if this has any performance penalty (the previous SQL looks ripe for a hash lookup, the new SQL does not, but IDK duckdb is probably smart enough to get around this)

### DIFF
--- a/docs/posts/ci-analysis/index.qmd
+++ b/docs/posts/ci-analysis/index.qmd
@@ -203,13 +203,11 @@ Let's also give them some names that'll look nice on our plots.
 stats = stats.mutate(
     raw_improvements=_.has_poetry.cast("int") + _.has_team.cast("int")
 ).mutate(
-    improvements=(
-        _.raw_improvements.case()
-        .when(0, "None")
-        .when(1, "Poetry")
-        .when(2, "Poetry + Team Plan")
-        .else_("NA")
-        .end()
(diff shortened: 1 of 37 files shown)


cpcloud:
@NickCrews Can you reopen this against `main`?

PR #7320, opened 2023-10-09
refactor(datafusion): create registry of time udfs to create them only once

mesejo:
This PR is based on this discussion: https://github.com/ibis-project/ibis/discussions/6967

The keys in the `udfs` dictionary are strings because we need to differentiate between signatures (i.e. input types) for the same function. The udf registry of datafusion is only based on the name of the functions.

### DIFF
--- a/ibis/backends/datafusion/compiler.py
+++ b/ibis/backends/datafusion/compiler.py
@@ -14,6 +14,7 @@ import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 from ibis import NA
+from ibis.backends.datafusion import registry
 from ibis.expr.operations.udf import InputType
 from ibis.formats.pyarrow import PyArrowType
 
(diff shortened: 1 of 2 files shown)


ibis/backends/datafusion/registry.py
```diff
@@ -0,0 +1,118 @@
...
+
+def extract_second(array: pa.Array) -> pa.Array:
+    return pc.cast(pc.second(array), pa.int32())
+
+
+def extract_millisecond(array: pa.Array) -> pa.Array:
+    return pc.cast(pc.millisecond(array), pa.int32())
+
+
+udfs = {
```
cpcloud, line 57:
```suggestion
UDFS = {
```

to give some indication that this variable is a module global.

ibis/backends/datafusion/compiler.py
```diff
@@ -937,34 +938,30 @@ def extract_hour(op, **kw):
...
-        input_types=[PyArrowType.from_ibis(op.arg.dtype)],
-        return_type=PyArrowType.from_ibis(op.dtype),
-        volatility="immutable",
-        name="extract_milliseconds_udf",
-    )
     arg = translate(op.arg, **kw)
-    return extract_milliseconds_udf(arg)
+
+    if op.arg.dtype.is_date():
+        return registry.UDFS["extract_millisecond_time"](arg)
```
cpcloud, line 944: It looks like we may not have test coverage for some of these. Would you mind adding some in a follow-up PR?

cpcloud:
Thanks for the PR!

PR #7322, opened 2023-10-10
feat(api): support deferred in reduction filters

cpcloud:
Add support for deferred objects in reduction filters.

This allows expressions like

```
In [3]: from ibis.interactive import *

In [4]: t = ex.penguins.fetch()

In [5]: t.bill_length_mm.mean(where=_.sex.lower() == "female")
Out[5]: 42.096969696969666
```


kszucs (approved):
LGTM

We have an emerging pattern here, would be nice to generalize this.

PR #7323, opened 2023-10-10
refactor(ir): remove `ibis.expr.streaming`

kszucs:
We shouldn't introduce a new module for a single object just yet. If there will be more streaming related construct we can factor them out to a separate module.

### DIFF
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -28,7 +28,7 @@ if TYPE_CHECKING:
     from pyflink.table import TableEnvironment
     from pyflink.table.table_result import TableResult
 
-    from ibis.expr.streaming import Watermark
+    from ibis.api import Watermark
 
 
 class Backend(BaseBackend, CanCreateDatabase):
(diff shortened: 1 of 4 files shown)


cpcloud:
cc @chloeh13q @deepyaman

jcrist (commented):
We know there will be new streaming-only apis. What harm is there in keeping `Watermark` in its own submodule for now? I'm not against removing it, but it also doesn't seem useful to remove it either.

kszucs:
No harm per se, but we used to have a similar TimeContext file containing implementation details of the pandas and dask backends slightly related to the ibis IR and it mostly provided noise and import cycles.

Watermark seems like a container to push implementation specific details down to the flink backend. 
It would be nice to see support for streaming operations in at least one additional backend before crafting a new streaming API. 

> We know there will be new streaming-only apis.
We can pull them out to appropiate files once we are there, but even then `ibis/expr/types/streaming.py` could be a mroe suitable location for that.

jcrist (approved):
Seems harmless.

PR #7325, opened 2023-10-11
refactor(clickhouse): use sqlglot for `create_table` implementation

cpcloud:
Move string munging in clickhouse create_table to sqlglot expressions.


gforsyth (approved):
looks good (pending test addition), one possible cleanup

PR #7327, opened 2023-10-11
feat(api): move analytic window functions to top-level

### LINKED ISSUE #5408: feat: add `ibis.dense_rank()`

This isn't a huge problem, I have something working, but more of a ux question. I have the following util function:

```python
from typing import Iterable

import ibis
from ibis.expr.types import (
    Column,
    IntegerColumn,
    Table,
)
import pandas as pd


def group_id(t: Table, keys: str | Column | Iterable[str | Column]) -> IntegerColumn:
    """Number each group from 0 to the "number of groups - 1" 
    
    This is equivalent to pandas.DataFrame.groupby(keys).ngroup().
    """
    # We need an arbitrary column to use for dense_rank
    col = t[t.columns[0]]
    return col.dense_rank().over(ibis.window(order_by=keys))
```
Example usage:
```python
df = pd.DataFrame(
    {
        "a": [5, 6, 6, 5],
        "b": [1, 2, 3, 1],
    }
)
t = ibis.memtable(df)

t.mutate(group_id=group_id(t, ["a", "b"]))
```
```
┏━━━━━━━┳━━━━━━━┳━━━━━━━━━━┓
┃ a     ┃ b     ┃ group_id ┃
┡━━━━━━━╇━━━━━━━╇━━━━━━━━━━┩
│ int64 │ int64 │ int64    │
├───────┼───────┼──────────┤
│     5 │     1 │        0 │
│     5 │     1 │        0 │
│     6 │     2 │        1 │
│     6 │     3 │        2 │
└───────┴───────┴──────────┘
```

I find it strange that I need to call `dense_rank()` on an arbitrary column of table. It feels dirty to just choose an arbitrary column. In raw SQL, DENSE_RANK() can stand all by itself, it doesn't have to associated with another expression. Similar to how ibis.row_number(). I would think that `ibis.dense_rank()` should work, and that the expression could bind to the table later, when I do the mutate(). Then I could remove the Table argument from my group_id() function.

Or perhaps I am missing something here. Or if you have a suggestion on how to do this task cleaner, I'd appreciate that too.

ibis.dense_rank() ? and similar for the other rank functions?

4.1.0

all

### Code of Conduct

cpcloud:
Moves all ranking-oriented analytic window functions to be top-level, and the existing implementations to use those top-level functions. Closes #5408.


kszucs (approved):

NickCrews:
Nice, I took 30 minutes trying to do this myself and then got lost in the weeds. Gives me vindication to see how much code had to get touched for this :) Thanks for picking this up!

PR #7328, opened 2023-10-11
feat(ux/duckdb): implement recursive collection type conversion for `.execute()`

cpcloud:
This PR implements recursive element conversion for arrays, structs, maps and json datatypes, for the pandas execution path. `ToJSONArray` and `ToJSONMap` are also implemented here for DuckDB.

### DIFF
--- a/ibis/formats/pandas.py
+++ b/ibis/formats/pandas.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 import warnings
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


ibis/formats/pandas.py
```diff
@@ -205,40 +206,78 @@ def convert_String(s, dtype, pandas_type):
...
+        return lambda row: {
+            key: convert_value(value) if value is not None else value
+            for key, value in dict(row).items()
+        }
+
+    @staticmethod
+    def convert_UUID_element(_):
+        from uuid import UUID
+
+        return lambda v: v if isinstance(v, UUID) else UUID(v)
```
jcrist, line 280: Do we not need to do a `None` check here?

jcrist (approved):
Overall LGTM

ibis/formats/pandas.py
cpcloud: No, this is handled by `na_action="ignore"` because there's no element recursion. Happy to move this back if you think that's more readable.

jcrist: Nah, if it works it's fine. Just wanted to make sure NULL values were handled for UUID columns.

PR #7330, opened 2023-10-11
feat(common): use patterns to filter out nodes during graph traversal

kszucs:


ibis/common/graph.py
```diff
@@ -88,19 +84,19 @@ def _recursive_get(obj: Any, dct: dict, filter: type) -> Any:
...
+    >>> _recursive_get(frozendict({1: 3}), dct)
     {1: 4}
-    >>> _recursive_get(frozendict({1: (1, 3)}), dct, filter=int)
+    >>> _recursive_get(frozendict({1: (1, 3)}), dct)
     {1: (2, 4)}
     """
-    if isinstance(obj, filter):
-        return dct[obj]
+    if isinstance(obj, Node):
+        return dct.get(obj, obj)
```
cpcloud, line 95: Not knowing much else about the intent here, I'd say this `get` should return `None` since this function is called `_recursive_get`, typically returning the `Node` doesn't match the input mapping's value type, making returning a `Node` very surprising when the lookup fails IMO.

ibis/common/graph.py
```diff
@@ -284,17 +264,19 @@ def __init__(self, mapping=(), /, **kwargs):
...
 
         Parameters
         ----------
-        root : Node
+        root
             Root node of the graph.
-        filter : Optional[type], default None
-            Type to filter out for the traversal, Node is filtered out by default.
+        filter
+            Pattern-like object to filter out nodes from the traversal. Essentially
```
cpcloud, line 277: Not a huge fan of `Essentially` here :)

Is it necessary? It seems like the sentence stands on its own with `Essentially`.

kszucs:
Converted to a draft PR because first, I would like to rebase #7245 on this PR to validate the use case.

cpcloud (commented):
LGTM.

Can you run the benchmarks? Looks like there might be some additional iteration happening now that a tuple is constructed from `_flatten_collections`.

ibis/common/graph.py
kszucs: Well, we need to keep the functionality as is, so it would better to rename the function instead. 

After this PR, when we stop the traversal at a `node`, its children are not going to be part of the graph, hence they won't be added to the results dictionary, but we still need to be able to recreate the `node` where we need the original children.

ibis/common/graph.py
kszucs: Nope, removing.

ibis/common/graph.py
kszucs: Naming suggestion? e.g. `_recursive_lookup`?

kszucs: Renamed.

cpcloud (approved):

PR #7332, opened 2023-10-11
fix(datatypes): correct unsigned integer bounds

gforsyth:
they were off by an order of magnitude.

added a silly test but it should catch any regressions that arise in future refactors.

### DIFF
--- a/ibis/expr/datatypes/tests/test_core.py
+++ b/ibis/expr/datatypes/tests/test_core.py
@@ -75,6 +75,23 @@ def test_dtype_from_classes(klass, expected):
     assert dt.dtype(klass) == expected
 
 
+@pytest.mark.parametrize(
+    ("uklass", "klass"),
+    [
+        (dt.UInt64, dt.Int64),
+        (dt.UInt32, dt.Int32),
+        (dt.UInt16, dt.Int16),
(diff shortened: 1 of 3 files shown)


ibis/expr/datatypes/tests/test_core.py
```diff
@@ -75,6 +75,23 @@ def test_dtype_from_classes(klass, expected):
     assert dt.dtype(klass) == expected
 
 
+@pytest.mark.parametrize(
+    ("uklass", "klass"),
+    [
+        (dt.UInt64, dt.Int64),
```
NickCrews, line 81: I would just test all 8 options literally to be totally clear and invite no possible problems:

(dt.UInt8, 0, 0xFF),
(dt.Int8, -32768, 32767),
etc

NickCrews:
Nice catch, thanks! One suggestion.

Is this going to break some user's code that used to work and now errors?

cpcloud:
> Nice catch, thanks! One suggestion.
> 
> Is this going to break some user's code that used to work and now errors?

Maybe, but isn't that the case for many bug fixes?

NickCrews:
right. 🤦‍♂️

cpcloud (approved):
Thank you!

PR #7335, opened 2023-10-12
feat(backends): add `ntile` support

cpcloud:
Add support for `ntile` across most of our SQL backends

### DIFF
--- a/ibis/backends/tests/test_window.py
+++ b/ibis/backends/tests/test_window.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from functools import partial
 from operator import methodcaller
 
 import numpy as np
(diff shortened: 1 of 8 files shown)


ibis/backends/tests/test_window.py
```diff
@@ -44,6 +44,55 @@
     HiveServer2Error = None
 
 
+# adapted from https://gist.github.com/xmnlab/2c1f93df1a6c6bde4e32c8579117e9cc
+def ntile(x, bucket: int):
```
jcrist, line 48: To not confuse this pandas implementation with `ibis.ntile` in the testing code below, would you mind renaming this to something like `pandas_ntile`?

jcrist (approved):
One small nit, otherwise LGTM.

ibis/backends/tests/test_window.py
cpcloud: Yep!

cpcloud: Done

cpcloud:
@xmnlab Just wanted to ping you about using your gist (the `pandas_ntile` implementation) in the test suite. Let me know if there's a license you want me to use, or if it's fine as is.

xmnlab:
@cpcloud , no specific license is needed for that. It is ok as it is. Thanks!!

PR #7336, opened 2023-10-12
refactor(clickhouse): use a pattern for one-to-zero index conversion of ranking window functions

cpcloud:
Use a pattern to convert output of ranked window functions from one-based indexing to zero-based indexing

### DIFF
--- a/ibis/backends/clickhouse/compiler/core.py
+++ b/ibis/backends/clickhouse/compiler/core.py
@@ -113,6 +113,13 @@ def translate(op: ops.TableNode, params: Mapping[ir.Value, Any]) -> sg.exp.Expre
     replace_notany_with_min_not = p.NotAny(x, where=y) >> c.Min(c.Not(x), where=y)
     replace_notall_with_max_not = p.NotAll(x, where=y) >> c.Max(c.Not(x), where=y)
 
+    # subtract one from ranking functions to convert from 1-indexed to 0-indexed
+    subtract_one_from_ranking_functions = p.WindowFunction(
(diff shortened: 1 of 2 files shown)


ibis/backends/clickhouse/compiler/core.py
```diff
@@ -113,13 +113,22 @@ def fn(node, _, **kwargs):
     replace_notany_with_min_not = p.NotAny(x, where=y) >> c.Min(c.Not(x), where=y)
     replace_notall_with_max_not = p.NotAll(x, where=y) >> c.Max(c.Not(x), where=y)
 
+    # subtract one from ranking functions to convert from 1-indexed to 0-indexed
+    subtract_one_from_ranking_functions = p.WindowFunction(
+        p.RankBase | p.NTile
```
kszucs, line 118: Can there be an intermediate `ops.Alias`? I assume no in usual circumstances.

cpcloud: Yes, an intermediate `Alias` can exist here.

This is a problem, but we should handle it at a different level.

Aliases should be allowed as input anywhere, and then removed everywhere except in three places:

- `ops.Selection.selections`
- `ops.Aggregation.metrics`
- `ops.Aggregation.by`

These are the only places where an alias is required.

Another way to think of it is that there are no `Value` operations where an `Alias` makes sense.

For example:

```python
t.x.name("y") + 1
```

would compile to

```sql
SELECT (x AS y) + 1
```

which isn't a valid operation.

PR #7338, opened 2023-10-12
chore(test): verbosely check [u]int dtype bounds

NickCrews:
Followup to https://github.com/ibis-project/ibis/pull/7332


NickCrews:
superseded by #7339

PR #7339, opened 2023-10-12
fix: uint upper bounds

NickCrews:
followup to #7332. A bug was sneaking through before :)

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -654,7 +654,7 @@ class UnsignedInteger(Integer):
     @property
     def bounds(self):
         exp = self.nbytes * 8
-        upper = 1 << exp
+        upper = (1 << exp) - 1
         return Bounds(lower=0, upper=upper)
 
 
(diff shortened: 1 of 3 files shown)


gforsyth (commented):
Thanks for the follow-up fix, @NickCrews !  Guess that'll show me to ignore explicit bounds.

I think the fix is causing the test failure in the implicit casting checks -- I would just remove or comment out the `(dt.uint64, dt.int64),` case for now and we can take care of that when we fix up the implicit vs. explicit casting stuff.

NickCrews:
Sorry if that snarky ":)" came out harsh! I didn't mean for that to be a dig at all.

gforsyth:
> Sorry if that snarky ":)" came out harsh! I didn't mean for that to be a dig at all.

not at all!

cpcloud:
This LGTM!

I removed the castable string changes, since they don't seem necessary to address the issue here. Those can be done in a follow-up but I also am -0.5 on the idea, `castable` isn't a public API, and we should avoid input transformation in internal APIs as much as we can, it makes the code harder to reason about.

cpcloud (approved):

PR #7342, opened 2023-10-12
fix: support self joins on memtables

jcrist:
I found this bug while refactoring a bit of this code to support `TABLESAMPLE` clauses. In short - the use of `ref_op` and `op` made it confusing which op to use, resulting in broken code for handling self joins on memtables. I renamed the variables to make it clear which `op` was the original operation and which was the `op` being compiled.


cpcloud (approved):

cpcloud:
Nice!

PR #7343, opened 2023-10-12
feat(oracle): add support for loading Oracle RAW and BLOB types

gforsyth:
Need to add a test for the `RAW` and `BLOB` loading and I don't like the fix for the column inference fallback

### DIFF
--- a/ibis/backends/oracle/__init__.py
+++ b/ibis/backends/oracle/__init__.py
@@ -9,6 +9,7 @@ import warnings
 from typing import TYPE_CHECKING, Any
 
 import oracledb
+import sqlglot as sg
 
 from ibis import util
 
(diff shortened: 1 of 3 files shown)


gforsyth:
Tests added and I'm much happier with this fix than the previous version.

ibis/backends/oracle/__init__.py
```diff
@@ -179,6 +180,18 @@ def _metadata(self, query: str) -> Iterable[tuple[str, dt.DataType]]:
 
         name = util.gen_name("oracle_metadata")
 
+        # If query is a table, adjust the query accordingly
+        sg_expr = sg.parse_one(query, dialect="oracle")
```
cpcloud, line 184: To make this maybe a bit less confusing, you can tell sqlglot how to parse the query:

```python
try:
    sg_expr = sg.parse_one(query, into=sg.exp.Table, dialect="oracle")
except ParseError:  # or whatever this error should be
    sg_expr = sg.parse_one(query, dialect="oracle")
```

It's a bit repetitive, but I think the intent is a bit more clear ("try to parse this query as a table first")

cpcloud: That said, this isn't a blocker!

cpcloud (approved):

ibis/backends/oracle/__init__.py
gforsyth: This is a good suggestion and requires less head-scratching.  Updated!

PR #7344, opened 2023-10-12
refactor(ir): remove `ops.Negatable`, `ops.NotAny`, `ops.NotAll`, `ops.UnresolvedNotExistsSubquery`

kszucs:
These were redundant operations, and now that we have better expression rewrite capabilities it is easier to replace operations nested under other value operations, e.g. `ops.Not`.

Supplemental changes:
- fixes a bug when we didn't wrap the analytical/reduction operation directly and we allowed any value operations inside a window function. This is now restricted to analytic/reduction ops
- managed to remove the window merge rewrite rule and the window propagation analysis function
- had to overcome that mssql not supports boolean expressions in selection positions, this fix actually resulted in 8 xpasses

### DIFF
--- a/ibis/backends/clickhouse/compiler/core.py
+++ b/ibis/backends/clickhouse/compiler/core.py
@@ -101,18 +101,6 @@ def translate(op: ops.TableNode, params: Mapping[ir.Value, Any]) -> sg.exp.Expre
         False, dtype="bool"
     )
 
-    # replace `NotExistsSubquery` with `Not(ExistsSubquery)`
-    #
-    # this allows to avoid having another rule to negate ExistsSubquery
-    replace_notexists_subquery_with_not_exists = p.NotExistsSubquery(...) >> c.Not(
-        c.ExistsSubquery(...)
-    )
-
-    # clickhouse-specific rewrite to turn notany/notall into equivalent
-    # already-defined operations
-    replace_notany_with_min_not = p.NotAny(x, where=y) >> c.Min(c.Not(x), where=y)
-    replace_notall_with_max_not = p.NotAll(x, where=y) >> c.Max(c.Not(x), where=y)
-
     # subtract one from ranking functions to convert from 1-indexed to 0-indexed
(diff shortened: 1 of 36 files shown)


cpcloud:
Really nice refactor. Great to see so much code deleted.

cpcloud (approved):
Assuming tests pass, this LGTM!

kszucs:
It will require some additional rewrites.

ibis/backends/clickhouse/compiler/core.py
```diff
@@ -124,9 +112,6 @@ def fn(node, _, **kwargs):
         replace_literals
         | replace_in_column_with_table_array_view
         | replace_empty_in_values_with_false
-        | replace_notexists_subquery_with_not_exists
-        | replace_notany_with_min_not
-        | replace_notall_with_max_not
```
cpcloud, line 129: Sweet, these were hacks anyway, even if they are elegant-looking :)

ibis/expr/types/generic.py
```diff
@@ -735,20 +735,23 @@ def over(
 
         def bind(table):
             frame = window.bind(table)
-            return ops.WindowFunction(self, frame).to_expr()
+            expr = an.windowize_function(self, frame)
+            if expr.equals(self):
+                raise com.IbisTypeError(
```
cpcloud, line 740: Do you have an example or test case for when this happens?

ibis/expr/types/generic.py
```diff
@@ -735,20 +735,23 @@ def over(
...
+        if isinstance(op, ops.WindowFunction):
             return op.func.to_expr().over(window)
         elif isinstance(window, bl.WindowBuilder):
             if table := an.find_first_base_table(self.op()):
                 return bind(table)
             else:
                 return Deferred(Call(bind, _))
         else:
-            return ops.WindowFunction(self, window).to_expr()
+            raise com.IbisTypeError("Unexpected window type: {window!r}")
```
cpcloud, line 754: Similarly wondering if you have an example or test case for when this happens.

Issue #7201, opened 2023-09-22
feat: Actual type hints (not representative) for UDFs


NickCrews:
### Is your feature request related to a problem?

Sorry that I'm making these issues on UDFs when I'd imagine you are about to release them :(

This is what I currently do:

```python
import pyarrow.compute as pc

@ibis.udf.scalar.pyarrow
def add_one(x: int) -> int:
    return pc.add(x, 1)
```

In the function, `x` isn't a python int, but is actually a pyarrow chunked array of type int. This both is confusing to me as the reader of this function, and it makes my editor sad.

### Describe the solution you'd like

Could we annotate in such a way that standard python tooling eg mypy is happy? Maybe using typing.Annotated

```python
from typing import Annotated

```python
import ibis
from pyarrow.lib import IntegerArray
from typing import Annotated
import pyarrow.compute as pc


@ibis.udf.scalar.pyarrow
def add_one(x: Annotated[IntegerArray, int]) -> Annotated[IntegerArray, int]:
    return pc.add(x, 1)

t = ibis.examples.penguins.fetch()
result = add_one(t.year)
```

### What version of ibis are you running?

main

### What backend(s) are you using, if any?

NA

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
UDF type hints are meant to be convenient for describing the element type of the inputs, not comprehensively compatible with mypy.

Is there some way we can have both?

I suppose we could add support for the `Annotated` versions of things for people who want better mypy compatibility.

I think this would ship after 7.0.

NickCrews:
I think accepting either `int` or `Annotated[pyarrow_type, int]` would just be extra code, but should work fine?

Thinking about this more, I might actually prefer if the type hints happened at the decoration stage, not the implementation stage:

```python
# don't love the __returns=...
# but I'm spitballing here
@ibis.udf.scalar.pyarrow(x=int, __returns=int)
def add_one(x: IntegerArray) -> IntegerArray:
    return pc.add(x, 1)

# switch to pandas implementation
# and the python type hints change (as they should)
# but the ibis hints don't change (as they shouldn't)

@ibis.udf.scalar.pandas(x=int, __returns=int)
def add_one(x: pd.Series) -> pd.Series:
    return x + 1

# can directly plug in 3rd party functions
# without needing to do a `def` block
from otherlib import process_series
ibis.udf.scalar.pandas(x=int, __returns=int)(process_series)

# or lambdas
[...]

PR #7357, opened 2023-10-13
feat: improve udf mypy compatibiity

jcrist:
- Removes unused `*args` argument to udf decorators
- Makes `ibis.udf.scalar.builtin`, `ibis.udf.scalar.python` and `ibis.udf.agg.builtin` compatible with mypy
- Makes `ibis.udf.scalar.pandas` and `ibis.udf.scalar.pyarrow` mypy compatible at _call_ time (so `myudf(1, 2)` will be valid). This does not resolve mypy compatibility of the UDF function definitions themselves. Supporting this will require an addition to the decorators. I'll address this in a follow-up PR.

Partially addresses #7201.

### DIFF
--- a/ibis/backends/postgres/__init__.py
+++ b/ibis/backends/postgres/__init__.py
@@ -206,7 +206,7 @@ WHERE p.proname = :name
         return op
 
     def _get_udf_source(self, udf_node: ops.ScalarUDF):
-        config = udf_node.__config__["kwargs"]
+        config = udf_node.__config__
         func = udf_node.__func__
         func_name = func.__name__
         return dict(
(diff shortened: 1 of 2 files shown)


jcrist:
Quick demo:

```python
import ibis


@ibis.udf.scalar.python
def myadd(a: int, b: int) -> int:
    return a + b


t = ibis.table({"x": "int", "y": "int"})


r1 = myadd(1, 2)
r2 = myadd(t.x, 2)
r3 = myadd(t.x, t.y)

reveal_type(r1)
reveal_type(r2)
reveal_type(r3)
```

```
$ mypy test.py 
test.py:16: note: Revealed type is "ibis.expr.types.generic.Value"
test.py:17: note: Revealed type is "ibis.expr.types.generic.Value"
test.py:18: note: Revealed type is "ibis.expr.types.generic.Value"
Success: no issues found in 1 source file
```

cpcloud (approved):
LGTM, excellent work!

PR #7359, opened 2023-10-13
refactor(sql): implement `Table.fillna`/`Table.dropna` using rewrite rules

jcrist:
Previously these were hacked in in a few different locations. Now all SQL backends share the same rewrite-rule based implementations, currently stored in a common location in `ibis.expr.rewrites`.

As a side effect, this also adds support for `Table.fillna` to the clickhouse backend.

### DIFF
--- a/ibis/backends/base/sql/compiler/query_builder.py
+++ b/ibis/backends/base/sql/compiler/query_builder.py
@@ -16,6 +16,7 @@ from ibis.backends.base.sql.compiler.translator import ExprTranslator, QueryCont
 from ibis.backends.base.sql.registry import quote_identifier
 from ibis.common.grounds import Comparable
 from ibis.config import options
+from ibis.expr.rewrites import rewrite_dropna, rewrite_fillna
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
@@ -517,6 +518,8 @@ class Compiler:
     support_values_syntax_in_select = True
     null_limit = None
 
+    rewrites = rewrite_fillna | rewrite_dropna
+
     @classmethod
     def make_context(cls, params=None):
         params = params or {}
@@ -535,6 +538,9 @@ class Compiler:
         # TODO(kszucs): consider to support a single type only
         if isinstance(node, ir.Expr):
             node = node.op()
+
+        if cls.rewrites:
+            node = node.replace(cls.rewrites)
 
         if context is None:
             context = cls.make_context()
--- a/ibis/expr/datatypes/value.py
+++ b/ibis/expr/datatypes/value.py
@@ -8,7 +8,7 @@ import ipaddress
 import json
 import uuid
 from collections.abc import Mapping, Sequence
-from typing import Any, NamedTuple
+from typing import Any
 
 import toolz
 from public import public
@@ -233,8 +233,9 @@ del infer.register
 
 
 @public
-class _WellKnownText(NamedTuple):
-    text: str
+class _WellKnownText:
+    def __init__(self, text: str):
+        self.text = text
 
     def __str__(self):
         return self.text
@@ -315,6 +316,8 @@ def normalize(typ, value):
                 return tuple(normalize(dt.linestring, item) for item in value)
             elif dtype.is_multipolygon():
                 return tuple(normalize(dt.polygon, item) for item in value)
+        elif isinstance(value, _WellKnownText):
+            return value
         return _WellKnownText(value.wkt)
     elif dtype.is_date():
         return normalize_datetime(value).date()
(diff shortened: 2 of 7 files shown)


ibis/backends/base/sql/compiler/query_builder.py
```diff
@@ -536,6 +539,9 @@ def to_ast(cls, node, context=None):
         if isinstance(node, ir.Expr):
             node = node.op()
 
+        if cls.rewrites:
+            node = node.replace(cls.rewrites)
```
jcrist, line 543: I looked into adding these generically in the `Backend` instead of in `Compiler`. However, here felt better since it was a much smaller code change. `to_ast` is called in many different places, trying to unify all those locations to also hit a backend's rewrite rules would be a larger refactor. IMO we can move rewrites into the `Backend` class later if needed.

ibis/expr/rewrites.py
```diff
@@ -0,0 +1,60 @@
...
+import ibis.expr.datatypes as dt
+import ibis.expr.operations as ops
+from ibis.common.patterns import pattern, replace
+from ibis.util import Namespace
+
+p = Namespace(pattern, module=ops)
+
+
+@replace(p.FillNa)
+def rewrite_fillna(_):
```
jcrist, line 16: I thought about using a more verbose name for these, but none felt right. `desugar_fillna` or `simplify_fillna` might be better, but :shrug:.

ibis/expr/datatypes/value.py
```diff
@@ -233,8 +233,9 @@ def infer_shapely_multipolygon(value) -> dt.MultiPolygon:
 
 
 @public
-class _WellKnownText(NamedTuple):
```
jcrist, line 236: When this was a `NamedTuple` the pattern matching traversal would recurse into it, resulting in issues reconstructing geospatial literal nodes.

I tried fixing a few places where it was checking for `isinstance(val, tuple)` to `type(val) is tuple` in the pattern handling code, but it endded up being easier to have this type not be a tuple type. I still think we might want to change those traversals to skip namedtuples (and maybe all subclasses of builtin containers).

kszucs: Could make sense, performance-wise too.

ibis/expr/rewrites.py
```diff
@@ -0,0 +1,60 @@
...
+    """Rewrite FillNa expressions to use more common operations."""
+    if isinstance(_.replacements, Mapping):
```
kszucs, line 18: Haven't taken a thorough look yet, just a quick look. One thing we may want to consider is to have more specific matchers in the rewrites to keep the replacement rules simpler and take their composition at the end.

In this scenario we clearly have a too generic `FillNa` operation (we should have a todo note about splitting it). Nonetheless we could match certain variants of the `FillNa` instances, according to the `FillNa.replacements` typehint we have two cases:

`UnionType[Value[dt.Numeric | dt.String], FrozenDict[str, Any]]`


```py
p.FillNa(replacements=p.Value)
```

```py
p.FillNa(replacements=Mapping) # or p.FillNa(replacements=FrozenDict)
```

we can also match on empty dictionary,  this should work since the dict value is turned into a more robuts dictionary matcher (though the empty case should be optimized to a `InstanceOf(Mapping)` pattern in a new `PatternMapping.__create__()`)

```py
p.FillNa(replacements={})  
```

So we *could* have 2 or three simpler replacement rules and then combine them with `|`. 

I'm not sure that should be the desired way to do it, after all we rewrite `FillNa` operations no matter what, not even sure whether would that result in more maintainable rules or not, but it worths discussing at least.

ibis/backends/base/sql/compiler/query_builder.py
cpcloud: I suspect this will be less of a refactor when more backends are sqlglot-based.

cpcloud: Agree that the compiler is the best place for these at the moment.

cpcloud (approved):

ibis/expr/rewrites.py
jcrist: > I'm not sure that should be the desired way to do it, after all we rewrite FillNa operations no matter what, not even sure whether would that result in more maintainable rules or not, but it worths discussing at least.

I agree it's worth discussing. Personally I like having all the code for this localized - it makes it clearer IMO to see how the specific operation (`fillna`) is simplified. Splitting it up into a few rewrite rules (or even splitting the op into 2 versions) would be a neutral to negative change in my mind. I'd expect the resulting code to be longer, and the loss of localization may make it harder to reason about.

Either way, this is a topic for a future PR.

ibis/expr/datatypes/value.py
jcrist: Should I open an issue about avoiding recursing into non-builtin types on the patterns/rewrites?

cpcloud:
The duckdb failure is known (working on a fix) and the impala one is unrelated

PR #7361, opened 2023-10-14
ci(trino): use upstream hive image to avoid building one from scratch

cpcloud:
Use an upstream hive image for testing trino instead of maintaining our own.


docker-compose.yml
```diff
@@ -188,25 +188,24 @@ services:
     volumes:
       - trino-postgres:/data
 
-  hive-metastore-mariadb:
-    image: mariadb:10.11.5
```
cpcloud, line 192: Switched this to postgres, as the starburst hive image uses some metastore storage type that isn't supported in mariadb.

docker-compose.yml
```diff
@@ -188,25 +188,24 @@ services:
...
     healthcheck:
       interval: 1s
       retries: 30
       test:
         - CMD
-        - mariadb-admin
-        - ping
+        - pg_isready
       timeout: 5s
+    command: -c port=23456
```
cpcloud, line 204: Have to use a different port than the default of 5432, which is bound by the `trino-postgres` container.

PR #7362, opened 2023-10-14
test(trino): clean up testing setup code

cpcloud:
This PR removes the extra `trino-postgres` container in favor of using the `hive.default` connector.

This avoids having to load the data with a separate postgres container (that required building a custom docker image).

Removing this container should improve service startup time and therefore Trino CI times.

Additionally, the Hive connector is very commonly used in Trino setups, so using that as the main testing connector should help work out any kinks with using that connector with ibis.

We're still using the memory connector for our one-off tests, due to some complexities with the hive connector:

1. `CREATE TABLE` queries don't seem to result in tables being visible in the catalog immediately. It seems to take anywhere from 0-100ms for a table to appear in the catalog after creating it. This also seems to be true for `INSERT`s.
2. I looked into using hive transactional tables to address point 1, but I found what appears to be [a bug in Trino related to this feature](https://github.com/trinodb/trino/issues/19407).
3. Running these tests (a lot of them are DDL or memtable queries) is much slower (adds about 10 minutes to CI). I assume that there's just a lot more machinery being creating a table in the hive connector than the memory connector, which would makes sense.


cpcloud:
Removing the postgres container doesn't help CI times much, but it does simplify testing code.

kszucs (approved):
LGTM

PR #7368, opened 2023-10-16
refactor(analysis): remove `_rewrite_filter()` in favor of using replacement patterns

kszucs:
`_rewrite_filter()` was mostly about the traversal itself, now it can be expressed with a simple pattern:

```py
node.replace(
    p.Reduction >> (lambda _: ops.TableArrayView(_.to_expr().as_table()), 
    filter=p.Value & ~p.WindowFunction & ~p.TableArrayView & ~p.ExistsSubquery
)
```

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -2281,11 +2281,7 @@ class Table(Expr, _FixedTextJupyterMixin):
         import ibis.expr.analysis as an
 
         resolved_predicates = _resolve_predicates(self, predicates)
-        predicates = [
-            an._rewrite_filter(pred.op() if isinstance(pred, Expr) else pred)
-            for pred in resolved_predicates
-        ]
-        return an.apply_filter(self.op(), predicates).to_expr()
+        return an.apply_filter(self.op(), resolved_predicates).to_expr()
 
     def nunique(self, where: ir.BooleanValue | None = None) -> ir.IntegerScalar:
         """Compute the number of unique rows in the table.
@@ -4158,12 +4154,13 @@ class CachedTable(Table):
         return current_backend._release_cached(self)
 
 
+# TODO(kszucs): used at a single place along with an.apply_filter(), should be
+# consolidated into a single function
 def _resolve_predicates(
     table: Table, predicates
 ) -> tuple[list[ir.BooleanValue], list[tuple[ir.BooleanValue, ir.Table]]]:
     import ibis.expr.types as ir
-    from ibis.expr.analysis import p, flatten_predicate
-    from ibis.common.deferred import _, Attr, Call
+    from ibis.expr.analysis import p, flatten_predicate, _
 
     # TODO(kszucs): clean this up, too much flattening and resolving happens here
     predicates = [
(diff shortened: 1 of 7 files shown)


ibis/tests/sql/snapshots/test_compiler/test_agg_filter/out.sql
```diff
@@ -10,6 +10,6 @@ t1 AS (
 SELECT t1.*
 FROM t1
 WHERE t1.`a` = (
-  SELECT max(t1.`a`) AS `blah`
+  SELECT max(t1.`a`) AS `Max(a)`
```
kszucs, line 13: The `Alias` is no longer pushed down into the subquery, but that should not be necessary.

ibis/expr/types/relations.py
```diff
@@ -4176,15 +4173,17 @@ def _resolve_predicates(
...
+    rules = (
+        # turn reductions into table array views so that they can be used as
+        # WHERE t1.`a` = (SELECT max(t1.`a`) AS `Max(a)`
+        p.Reduction >> (lambda _: ops.TableArrayView(_.to_expr().as_table()))
+        |
+        # resolve unresolved exists subqueries to IN subqueries
+        p.UnresolvedExistsSubquery >> (lambda _: _.resolve(table.op()))
+    )
+    # do not apply the rules below the following nodes
+    until = p.Value & ~p.WindowFunction & ~p.TableArrayView & ~p.ExistsSubquery
```
kszucs, line 4185: Here we either could use a common base class or trait to check against, or rethink how we treat the current `Reduction` nodes. Actually reductions are not *valid* on their own, we usually rewrite them to aggregations, subqueries or window functions. 

Perhaps we should treat them as API-level constructs? That way rewrites could get simpler, but we should still fit them somehow into the `Value` node hierarchy.

cpcloud: Why are reductions not valid on their own?

Isn't `t.x.sum()` a perfectly valid query?

ibis/tests/sql/snapshots/test_compiler/test_agg_filter/out.sql
cpcloud: Yep, the alias isn't needed here.

ibis/tests/sql/snapshots/test_compiler/test_subquery_where_location/decompiled.py
```diff
@@ -13,7 +13,7 @@
 param = ibis.param("timestamp")
 proj = alltypes.select(
     [alltypes.float_col, alltypes.timestamp_col, alltypes.int_col, alltypes.string_col]
-).filter(alltypes.timestamp_col < param)
+).filter(alltypes.timestamp_col < param.name("my_param"))
```
cpcloud, line 16: Is this change necessary?

cpcloud: Oh, I guess it's a generated code change, nevermind!

cpcloud (approved):
Sweet!

ibis/expr/types/relations.py
cpcloud: Ah, got it! Yeah, there's no valid SQL that corresponds to only the aggregation function call: you need a `SELECT`.

PR #7371, opened 2023-10-16
refactor(analysis): remove `substitute_parents()`

kszucs:
Apparently we don't require it anymore, I'd guess that the functionality of this function is superseded by sub_immediate_parents().

### DIFF
--- a/ibis/expr/analysis.py
+++ b/ibis/expr/analysis.py
@@ -144,33 +144,6 @@ def substitute(fn, node):
         return node
 
 
-def substitute_parents(node):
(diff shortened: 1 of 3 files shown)


ibis/expr/analysis.py
```diff
@@ -144,33 +144,6 @@ def substitute(fn, node):
         return node
 
 
-def substitute_parents(node):
```
kszucs, line 147: Apparently we don't require it anymore, I'd guess that the functionality of this function is superseded by `sub_immediate_parents()`.

cpcloud (approved):
Huzzah! 🎉

PR #7373, opened 2023-10-16
docs(trino): add note about specifying RBAC role with Starburst

gforsyth:


gforsyth:
Don't merge this -- going to poke at starburst a little bit first

gforsyth:
Ok, Starburst doesn't appear to require specifying a role in all situations, so I left a note in the docs to try it if there are issues listing tables.

Also added in some guidance in determining what to use for connecting to a starburst trino server.

.pre-commit-config.yaml
```diff
@@ -33,6 +33,7 @@ repos:
     rev: v2.2.6
     hooks:
       - id: codespell
+        entry: codespell --skip="*.js,*.html,*.css,*.svg"
```
cpcloud, line 36: Should we add this to the pyproject.toml config?

cpcloud (approved):

.pre-commit-config.yaml
gforsyth: yeah, I'll add that

cpcloud: In theory you should be able to only declare it in `pyproject.toml` and it'll get picked up by the pre-commit invocation

gforsyth: yep, that worked.

PR #7374, opened 2023-10-16
refactor(analysis): simplify and improve `pushdown_selection_filters()` 

kszucs:


ibis/expr/analysis.py
```diff
@@ -196,69 +196,30 @@ def apply_filter(op, predicates):
...
+    simplified = []
+    for pred in predicates:
+        if pred.match(
+            p.Reduction | p.Analytic | conflicting_projection, filter=p.Value
+        ):
+            return default
+
+        try:
+            simplified.append(pred.replace(pushdown_pattern))
+        except IntegrityError:
```
kszucs, line 219: I hate this, the `parent.table` is possibly a join, but occasionally the replacement works.

ibis/tests/sql/snapshots/test_sqlalchemy/test_where_correlated_subquery_with_join/out.sql
kszucs, line 1: I have not reviewed this change yet.

cpcloud: Plenty of the TPC-H queries are of this form, so execution of those tests should catch incorrect behavior here.

ibis/expr/operations/relations.py
```diff
@@ -531,6 +528,10 @@ def __init__(self, table, metrics, by, having, predicates, sort_keys):
             sort_keys=sort_keys,
         )
 
+    @attribute
+    def _projection(self):
+        return Projection(self.table, self.metrics + self.by)
```
cpcloud, line 533: Maybe this should be `by + metrics` since that's the order we return them in to the user?

```suggestion
        return Projection(self.table, self.by + self.metrics)
```

kszucs: Yes, though this will matter when we split the Selection and Aggregation objects.

cpcloud (approved):
SHIP ITTTTTT!

ibis/expr/analysis.py
```diff
@@ -391,6 +268,14 @@ def __init__(self, parent, proj_exprs):
     def get_result(self):
         roots = find_immediate_parent_tables(self.parent.op())
         first_root = roots[0]
+        parent_op = self.parent.op()
+
+        # reprojection of the same selections
```
kszucs, line 273: This is a new optimization actually, thanks to this we eliminate redundant projections.

kszucs:
Please wait with merging, I have additional code to remove :)

cpcloud:
Ready for merge?

PR #7375, opened 2023-10-16
ci(duckdb): bring back a faster version of isolated extension downloads

cpcloud:
Should fix the occasionally-failing windows duckdb tests


gforsyth (approved):
:shipit:

PR #7366, opened 2023-10-15
test(flink): turn on the common backend test suite

mfatihaktas:
# Summary

This is a follow-up to this [PR](https://github.com/ibis-project/ibis/pull/6920). The changes here pass/skip locally -- copied the logs below. Created this PR to test if they also complete successfully on the CI.

```bash
ibis on  test/flink/enable-common-tests [⇕$!?] via 🅒 ibis-dev took 23.6s
➜ pytest -m flink --showlocals --capture=no --color=yes 'ibis/backends/tests/'
========================================================================= test session starts ==========================================================================
platform darwin -- Python 3.10.12, pytest-7.4.0, pluggy-1.2.0
Using --randomly-seed=3842345966
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/mehmet/Desktop/ibis
configfile: pyproject.toml
plugins: snapshot-0.0.0, cov-4.1.0, clarity-1.0.1, mock-3.11.1, randomly-3.12.0, xdist-3.3.1, repeat-0.9.1, benchmark-4.0.0, httpserver-1.0.6, hypothesis-6.80.0
collected 22796 items / 21544 deselected / 1252 selected

ibis/backends/tests/test_generic.py x.......x....x...x..x...x......x...x......x.....x.....xx...................x..x.x...x.......x...x.......x.......x...x....xx.........x...x......x..x.xx...x
ibis/backends/tests/test_string.py ..............x.............x.....x..x...................x.....x.......x.....x.....
ibis/backends/tests/test_map.py xxxxxxxxxxxxxxxxxxxxxxxxxxx
ibis/backends/tests/test_struct.py xxxxxxxxxxxx
ibis/backends/tests/tpch/test_h02.py s
ibis/backends/tests/tpch/test_h07.py s
ibis/backends/tests/tpch/test_h15.py s
ibis/backends/tests/tpch/test_h22.py s
ibis/backends/tests/test_array.py x.xxx.x.xxxxxxxxxxxx.xxxxx.xxxxxxxxxx.xxxxxxxx.
ibis/backends/tests/tpch/test_h01.py s
ibis/backends/tests/test_api.py ............
ibis/backends/tests/tpch/test_h04.py s
ibis/backends/tests/test_set_ops.py ................
[...]

PR #7382, opened 2023-10-17
feat(flink): add remaining operators for Flink to pass/skip the common tests

mfatihaktas:
* This is a follow-up to this [previously merged PR](https://github.com/ibis-project/ibis/pull/7265).
* Contains the remaining changes in `ibis/ibis/backends/flink/registry.py` to pass/skip the common tests for Flink backend.
* Next step is the [PR](https://github.com/ibis-project/ibis/pull/7366) that contains changes in the common tests to make them pass/skip for Flink backend.

### DIFF
--- a/ibis/backends/flink/registry.py
+++ b/ibis/backends/flink/registry.py
@@ -40,9 +40,19 @@ def _extract_field(sql_attr: str) -> str:
 
 
 def _cast(translator: ExprTranslator, op: ops.generic.Cast) -> str:
-    if op.to.is_timestamp() and op.to.timezone:
-        arg_translated = translator.translate(op.arg)
-        return f"TO_TIMESTAMP(CONVERT_TZ(CAST({arg_translated} AS STRING), 'UTC+0', '{op.to.timezone}'))"
+    arg, to = op.arg, op.to
+    arg_translated = translator.translate(arg)
+    if to.is_timestamp():
+        if arg.dtype.is_numeric():
+            arg_translated = f"FROM_UNIXTIME({arg_translated})"
+
+        if to.timezone:
+            return f"TO_TIMESTAMP(CONVERT_TZ(CAST({arg_translated} AS STRING), 'UTC+0', '{to.timezone}'))"
+        else:
+            return f"TO_TIMESTAMP({arg_translated})"
+
+    elif to.is_json():
+        return arg_translated
 
     from ibis.backends.base.sql.registry.main import cast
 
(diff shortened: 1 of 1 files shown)


ibis/backends/flink/registry.py
```diff
@@ -40,9 +40,21 @@ def extract_field_formatter(translator: ExprTranslator, op: ops.Node) -> str:
...
+    if isinstance(to, Timestamp):
+        if arg.dtype.is_numeric():
+            arg_translated = f"FROM_UNIXTIME({arg_translated})"
+
+        if to.timezone:
+            return f"TO_TIMESTAMP(CONVERT_TZ(CAST({arg_translated} AS STRING), 'UTC+0', '{to.timezone}'))"
+        else:
+            return f"TO_TIMESTAMP({arg_translated})"
+
+    elif isinstance(to, JSON):
```
cpcloud, line 56: Don't check with `isinstance`, use the is_* methods. A method exists for every leaf type as well as every parent class.

```suggestion
    from ibis.expr.datatypes.core import JSON, Timestamp

    arg, to = op.arg, op.to
    arg_translated = translator.translate(arg)
    if to.is_timestamp():
        if arg.dtype.is_numeric():
            arg_translated = f"FROM_UNIXTIME({arg_translated})"

        if to.timezone:
            return f"TO_TIMESTAMP(CONVERT_TZ(CAST({arg_translated} AS STRING), 'UTC+0', '{to.timezone}'))"
        else:
            return f"TO_TIMESTAMP({arg_translated})"

    elif to.is_json():
```

ibis/backends/flink/registry.py
```diff
@@ -281,10 +251,37 @@ def _array_index(translator: ExprTranslator, op: ops.arrays.ArrayIndex):
...
+    key_array = translator.translate(op.keys)
+    value_array = translator.translate(op.values)
+
+    return f"MAP_FROM_ARRAYS({key_array}, {value_array})"
+
+
+def _map_get(translator: ExprTranslator, op: ops.maps.MapGet) -> str:
+    map_ = translator.translate(op.arg)
+    key = translator.translate(op.key)
+    return f"{map_} '[' {key} ']'"
```
cpcloud, line 280: This looks like the same issue we found when compiling array indexing. Can you remove the single quotes surrounding the brackets?

ibis/backends/flink/registry.py
```diff
@@ -281,10 +251,37 @@ def _array_index(translator: ExprTranslator, op: ops.arrays.ArrayIndex):
...
+
+
+def _day_of_week_index(translator: ExprTranslator, op: ops.Node) -> str:
+    arg = translator.translate(op.args[0])
```
cpcloud, line 284: Remember to use the name of the argument and not `op.args[index]`.

ibis/backends/flink/registry.py
```diff
@@ -319,6 +319,64 @@ def _string_to_timestamp(
     return f"TO_TIMESTAMP({arg}, {format_string})"
 
 
+def _time(translator: ExprTranslator, op: ops.temporal.Time) -> str:
+    from ibis.expr.datatypes.core import Timestamp
+
+    if isinstance(op.arg.dtype, Timestamp):
```
cpcloud, line 325: Use the `is_timestamp()` method:

```suggestion
    if op.arg.dtype.is_timestamp():
```

cpcloud (changes requested):

ibis/backends/flink/registry.py
mfatihaktas: Thanks, I replaced `isinstance()` with `is_*` in all my changes.

ibis/backends/flink/registry.py
mfatihaktas: Ah, sorry, I missed it again, even after reviewing my own code :) Thank you for catching it.

ibis/backends/flink/registry.py
mfatihaktas: Thanks for the reminder, fixed it.

ibis/backends/flink/registry.py
mfatihaktas: Thanks, fixed.

cpcloud (approved):

PR #7384, opened 2023-10-17
chore(bigquery): regenerate sql

cpcloud:
Regenerate some failing bigquery test snapshots. I think this is due to [these lines](https://github.com/ibis-project/ibis/pull/7377/files#diff-ff7982dc41d280b20e266548b77f73ce378fc7d859b88192e8e3f07e63af75aaR131-R132) from #7377. Previously we unconditionally produced an alias in the string compiler, in that PR we only produce one if the the retrieved reference does not match the alias


jcrist (approved):

jcrist:
Oop, sorry, forgot to check biquery on that PR (I haven't yet setup local testing on bigquery). FWIW the logic here now matches that done in the sqlalchemy backend and I _think_ should be correct. it's encouraging that this was the only breakage.

cpcloud:
No worries!

PR #7387, opened 2023-10-17
feat(api): support specifying `sig` in udf definitions

### LINKED ISSUE #7201: feat: Actual type hints (not representative) for UDFs

jcrist:
This adds support for a new `sig` parameter to all udf definition functions. If present, this will be used for determining the UDF signature instead of the function annotations. This can be useful for:

- Users who don't want to make use of annotations
- Programmatic usage, where generating a function with a signature may be tricky
- Pandas or pyarrow udfs where you want the decorated function to be annotated with pandas/pyarrow types, and need another way of expressing the types ibis should use
- Expressing more complicated input/output types than can easily be expressed via python annotations. `sig` accepts anything coercible to an ibis dtype, so struct/map/array types are more easily expressible

Fixes #7201.

### DIFF
--- a/ibis/expr/operations/udf.py
+++ b/ibis/expr/operations/udf.py
@@ -78,33 +78,47 @@ class _UDF(abc.ABC):
         input_type: InputType,
         name: str | None = None,
         schema: str | None = None,
+        signature: tuple[tuple, Any] | None = None,
         **kwargs,
     ) -> type[S]:
         """Construct a scalar user-defined function that is built-in to the backend."""
 
-        annotations = typing.get_type_hints(fn)
-        if (return_annotation := annotations.pop("return", None)) is None:
-            raise exc.MissingReturnAnnotationError(fn)
+        if signature is None:
+            annotations = typing.get_type_hints(fn)
+            if (return_annotation := annotations.pop("return", None)) is None:
+                raise exc.MissingReturnAnnotationError(fn)
+            fields = {
+                arg_name: Argument(
+                    pattern=rlz.ValueOf(annotations.get(arg_name)),
+                    default=param.default,
+                    typehint=annotations.get(arg_name, Any),
+                )
+                for arg_name, param in inspect.signature(fn).parameters.items()
+            }
+        else:
+            arg_types, return_annotation = signature
+            arg_names = list(inspect.signature(fn).parameters)
+            fields = {
+                arg_name: Argument(pattern=rlz.ValueOf(typ), typehint=typ)
+                for arg_name, typ in zip(arg_names, arg_types)
+            }
 
         func_name = name if name is not None else fn.__name__
 
-        fields = {
-            arg_name: Argument(
-                pattern=rlz.ValueOf(annotations.get(arg_name)), default=param.default
-            )
-            for arg_name, param in inspect.signature(fn).parameters.items()
-        } | {
-            "dtype": dt.dtype(return_annotation),
-            "__input_type__": input_type,
-            # must wrap `fn` in a `property` otherwise `fn` is assumed to be a
-            # method
-            "__func__": property(fget=lambda _, fn=fn: fn),
-            "__config__": FrozenDict(kwargs),
-            "__udf_namespace__": schema,
-            "__module__": fn.__module__,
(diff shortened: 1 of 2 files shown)


jcrist:
An example showing mypy being happy with a pandas udf:

```python
import ibis
import pandas as pd

t = ibis.table({"x": "int", "y": "int"})

@ibis.udf.scalar.pandas(sig=((int, int), int))
def foo(x: pd.Series, y: pd.Series) -> pd.Series:
    reveal_type(x)
    return 2 * (x + y)

reveal_type(foo)
reveal_type(foo(t.x, t.y))
```

And output:

```
test.py:8: note: Revealed type is "pandas.core.series.Series[Any]"
test.py:11: note: Revealed type is "def (*Any, **Any) -> ibis.expr.types.generic.Value"
test.py:12: note: Revealed type is "ibis.expr.types.generic.Value"
Success: no issues found in 1 source file
```

jcrist:
And an example returning a struct type:

```python
import ibis
import pandas as pd

t = ibis.table({"x": "int", "y": "int"})

@ibis.udf.scalar.builtin(
    sig=((int, int), "struct<quotient: int, remainder: int>")
)
def divmod(x, y):
    ...

print(divmod(t.x, t.y).type())
```

```
$ python test.py
struct<quotient: int64, remainder: int64>
```

ibis/expr/operations/udf.py
```diff
@@ -78,33 +78,47 @@ def _make_node(
         input_type: InputType,
         name: str | None = None,
         schema: str | None = None,
+        sig: tuple[tuple, Any] | None = None,
```
cpcloud, line 81: Not really a huge fan of the aesthetics of `sig`.

I don't like that the entire signature is specified as another keyword argument and I don't like all the punctuation.

Anyway, I'm an old man yelling at a cloud :)

Some alternatives that I like better:

```python
# input types must be a positional argument if passed, `returns` must be
# specified as a keyword argument.
@builtin((int, int), returns=str, **kwargs)
```

```python
# this may be controversial and/or require a bunch of invasive changes to
# implement :)
@builtin.signature(int, int).returns(str).options(**kwargs)
```

I think I could be convinced by `signature` if you don't like any of my alternatives, but at the very least `sig` feels a bit too short.

cpcloud (commented):
Implementation looks good, mostly just complaints about aesthetics 🙂

ibis/expr/operations/udf.py
jcrist: I don't mind the former, but it's tricky to implement in a sane way since the first positional argument may also accept a function. I really don't like the builder api in option 2 :).

Went with renaming to `signature` for now (these APIs are still experimental, we can always change it later).

While a few more keystrokes, this syntax has the perk of expanding nicely to support overloads as `signature=[((int, int), int), ((float, float), float)]` if we ever decide we want to support those. Having two arguments for `input_types` and `returns` would be harder to support this use case.

cpcloud (approved):

NickCrews:
Thank you both of you!

ibis/expr/operations/udf.py
NickCrews: I agree with everything here, but I think the current API is the best option I've seen.

PR #7390, opened 2023-10-18
chore(dev-deps): use unified lock file format for conda lock files

cpcloud:
This PR fixes our conda-lock setup and installation instructions to use the unified conda lock file format, which reduces the number of files in the repo and allows us to remove platform specific installation instructions. The remaining major issue is the incorrect resolution of the `seaborn-split` package.

One nice benefit of using the unified lock file format is that we can use pypi packages without having to upstream them to `conda-forge` first.


.pre-commit-config.yaml
```diff
@@ -56,7 +56,7 @@ repos:
     rev: v4.5.0
     hooks:
       - id: check-added-large-files
-        args: ["--maxkb=700"]
+        args: ["--maxkb=710"]
```
gforsyth, line 59: lol

gforsyth (approved):
nice!  those are some big ol' lock files you've got there.

PR #7393, opened 2023-10-18
fix(duckdb): ensure proper quoting when compiling cross database/schema tables

cpcloud:
This PR fixes cross-database/schema access for DuckDB.

### DIFF
--- a/ibis/backends/base/sql/alchemy/__init__.py
+++ b/ibis/backends/base/sql/alchemy/__init__.py
@@ -904,19 +904,27 @@ class AlchemyCrossSchemaBackend(BaseAlchemyBackend):
     currently active one.
     """
 
-    def _get_sqla_table(
-        self,
-        name: str,
-        schema: str | None = None,
-        database: str | None = None,
-        **_: Any,
-    ) -> sa.Table:
-        current_db = self.current_database
-        current_schema = self.current_schema
+    def _get_table_identifier(self, *, name, schema, database):
         if schema is None:
-            schema = current_schema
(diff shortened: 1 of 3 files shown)


gforsyth (approved):
looks good.  I'm debating if we should move `list_tables` into the `AlchemyCrossSchemaBackend` class, too, but that can wait.

ibis/backends/base/sql/alchemy/__init__.py
```diff
@@ -904,26 +904,44 @@ class AlchemyCrossSchemaBackend(BaseAlchemyBackend):
...
+            schema = sg.parse_one(schema, into=sg.exp.Identifier)
+        except sg.ParseError:
+            # not actually a table, but that's how sqlglot parses
+            # `CREATE SCHEMA` statements
+            parsed = sg.parse_one(schema, into=sg.exp.Table)
+
+            # user passed database="foo", schema="bar.baz", which is ambiguous
+            if database is not None:
+                raise com.IbisInputError(
+                    "Cannot specify both `database` and a dotted path in `schema`"
```
cpcloud, line 921: This code can't currently be hit because of the parent class implementation trying to use the deprecated `self.database` method.

cpcloud: Do you want me to remove this code, or are you referring to the `self.database` access?

gforsyth: oops, my bad.  I meant the `self.database` call but that was an aspirational "we should remove" -- not a blocker here

PR #7394, opened 2023-10-18
feat(duckdb): add `attach` and `detach` methods for adding and removing databases to the current duckdb session

cpcloud:
Add `attach` and `detach` methods mirroring the [DuckDB syntax](https://duckdb.org/docs/sql/statements/attach.html).


jcrist (approved):
Nice!

Issue #7329, opened 2023-10-11
feat: better consistency between list_tables and table


gforsyth:
The `list_tables` method in the `BaseBackend` takes a `database` kwarg to
specify which `database` to look in for tables.  Since `database` is a bit of an
overused term, this leads to confusing interactions.

For instance, in the SQLAlchemy backends, we pass the `database` kwarg as
`schema=database` inside the `inspector.get_*` methods.

However when creating a table, you would need to pass that same arg, say, `SYS`
as the `schema` arg:
```python
    con.list_tables(database="SYS")
    
    dual = con.table("DUAL", schema="SYS)
```

This is not great.  With, for example, the `oracle` backend:
```python
    con.list_tables(schema="SYS")  # errors, no schema kwarg allowed
    con.table("table", database="SYS")  # errors, ~SYS~ is a "schema" in this context
    
    con.list_tables(database="SYS")  # works
    con.table("table", schema="SYS") # fetches a table from ~SYS~
```

### 1.  Change the `database` kwarg to `schema`

```python
    def list_tables(self, like=None, database=None):
        tables = self.inspector.get_table_names(schema=database)
        views = self.inspector.get_view_names(schema=database)
        return self._filter_with_like(tables + views, like)
```
to:
```python
    def list_tables(self, like=None, schema=None):
        tables = self.inspector.get_table_names(schema=schema)
        views = self.inspector.get_view_names(schema=schema)
        return self._filter_with_like(tables + views, like)
```
Breaking change, and for backends like `sqlite`, `clickhouse`, `mysql`,
`impala`, `pyspark` might cause confusion because they don&rsquo;t have a `database
   -> schema` hierarchy, they only have `database`.

### 2.  Add a `schema` kwarg to `list_tables`
```python    
        def list_tables(self, like=None, database=None):
            tables = self.inspector.get_table_names(schema=database)
            views = self.inspector.get_view_names(schema=database)
            return self._filter_with_like(tables + views, like)
```    

to:

```python    
        def list_tables(self, like=None, schema=None, database=None):
            if database is not None:
                raise exc.UnsupportedOperationError(
                    "No support for database in list_tables for backend X"
                )
            tables = self.inspector.get_table_names(schema=schema)
            views = self.inspector.get_view_names(schema=schema)
            return self._filter_with_like(tables + views, like)
```
Still breaking for some backends. Also, except for `duckdb` and `flink`, the
extra kwarg serves no purpose, so the majority of backends will have a
[...]

PR #7395, opened 2023-10-18
Update `list_tables` to account for database->schema hierarchy where applicable

gforsyth:
Ok, need to add tests for this, but the changes are a bit scattershot and I thought we can use this as a sounding-board for whether we want to go down this path.

The general idea here is:

1. Backends that only have a concept of `database` remain unchanged (`clickhouse`, `druid`, `impala`, `mysql`, `pyspark`, `sqlite`)
2. Backends that allow you to _interact_ with tables in other database -> schema hierarchies now have a `schema` kwarg in `list_tables` to allow fine-grained specification (formalized in `bigquery`, added for `trino`, `mssql`, and `snowflake`, and already present in `duckdb`)
3. Backends that have this hierarchy but don't allow for cross-database operations have a kwarg change from `database` to `schema` -- this unifies the arguments of `list_tables` and `table` for these backends.

For `trino`, this is quite pleasing:

```python
[ins] In [1]: import ibis

[ins] In [2]: con = ibis.trino.connect(user="user", password="", host="localhost", port=8080, data
         ...: base="memory", schema="default")

[ins] In [3]: con.list_databases()
Out[3]: ['hive', 'jmx', 'memory', 'postgresql', 'system', 'tpcds', 'tpch']

[ins] In [4]: con.list_schemas(database="tpch")
Out[4]: 
['information_schema',
 'sf1',
 'sf100',
 'sf1000',
 'sf10000',
 'sf100000',
 'sf300',
 'sf3000',
 'sf30000',
 'tiny']

[ins] In [5]: con.list_tables(database="tpch", schema="sf1")
Out[5]: 
['customer',
 'lineitem',
 'nation',
 'orders',
 'part',
 'partsupp',
 'region',
 'supplier']
```

It also adds consistency for users who are operating on Oracle or Postgres and trying to connect to tables in another schema:

```python
[ins] In [1]: import ibis
co
[ins] In [2]: con = ibis.oracle.connect(user="ibis", password="ibis", host="localhost", port=1521,
         ...:  database="ibis_testing")

[ins] In [3]: con.list_tables()
Out[3]: 
['astronauts',
 'awards_players',
 'batting',
 'diamonds',
 'functional_alltypes',
 'win']

[ins] In [4]: con.list_tables(schema="SYS", like="EXU8OPT")
Out[4]: ['EXU8OPT']

[ins] In [5]: con.table("EXU8OPT", schema="SYS")
Out[5]: 
DatabaseTable: EXU8OPT
  PARAMETER string
  VALUE     decimal(38, 0)
```

Downsides are that there's a bunch of fragmentation in the individual `list_tables` methods.
Probably can encapsulate some of this in one of the intermediate classes like `AlchemyCanCreateSchema`

xref #7329

### DIFF
--- a/ibis/backends/bigquery/__init__.py
+++ b/ibis/backends/bigquery/__init__.py
@@ -19,6 +19,7 @@ import ibis
 import ibis.common.exceptions as com
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
+from ibis import util
 from ibis.backends.base import CanCreateSchema, CanListDatabases, Database
 from ibis.backends.base.sql import BaseSQLBackend
 from ibis.backends.bigquery.client import (
@@ -528,8 +529,45 @@ class Backend(BaseSQLBackend, CanCreateSchema, CanListDatabases):
         return self.list_schemas(like=like)
 
     def list_tables(
-        self, like: str | None = None, database: str | None = None
+        self,
+        like: str | None = None,
+        database: str | None = None,
+        schema: str | None = None,
     ) -> list[str]:
+        """List the tables in the database.
+
+        Parameters
+        ----------
+        like
+            A pattern to use for listing tables.
+        database
+            The database (project) to perform the list against.
+        schema
+            The schema (dataset) inside `database` to perform the list against.
+
+            ::: {.callout-warning}
+            ## `schema` refers to database hierarchy
+
+            The `schema` parameter does **not** refer to the column names and
+            types of `table`.
+            :::
+        """
+        if database is not None and schema is None:
+            util.warn_deprecated(
+                "database",
+                instead=(
+                    f"{self.name} cannot list tables only using `database` specifier. "
+                    "Include a `schema` argument."
+                ),
+                as_of="7.1",
+                removed_in="8.0",
+            )
+            database = sg.parse_one(database, into=sg.exp.Table).sql(dialect=self.name)
+        elif database is None and schema is not None:
+            database = sg.parse_one(schema, into=sg.exp.Table).sql(dialect=self.name)
+        else:
+            database = sg.table(schema, db=database).sql(dialect=self.name) or None
+
         project, dataset = self._parse_project_and_dataset(database)
         dataset_ref = bq.DatasetReference(project, dataset)
         result = [table.table_id for table in self.client.list_tables(dataset_ref)]
--- a/ibis/backends/mssql/__init__.py
+++ b/ibis/backends/mssql/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Literal
 
 import sqlalchemy as sa
+import sqlglot as sg
 import toolz
 
 from ibis.backends.base import CanCreateDatabase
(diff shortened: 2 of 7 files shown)


cpcloud:
5. motherduck does allow a `CREATE DATABASE` statement, making it literally and figuratively the odd duck 🦆

cpcloud:
I think sacrificing some consistency in the `list_*` APIs is worth the decrease in conceptual distance from ibis -> backend (fewer new concepts to map from ibis to the backend)

gforsyth:
Ok -- I'll rebase this against the changes in #7393 and then add some tests.  Then I can see about removing that deprecated `self.database` call so we can start grabbing tables more smoothly.

cpcloud:
I went ahead and removed that `self.database()` call in #7399. It wasn't strictly necessary to fix the related bug, but I didn't want to do another partial fix.

gforsyth:
> I went ahead and removed that `self.database()` call in #7399. It wasn't strictly necessary to fix the related bug, but I didn't want to do another partial fix.

Sounds good -- I'm taking a look at it now

ibis/backends/snowflake/__init__.py
```diff
@@ -472,6 +474,48 @@ def list_schemas(
...
+        """
+        query = "SHOW TABLES"
+
+        if database is not None and schema is None:
+            raise com.UnsupportedArgumentError(
+                f"{self.name} cannot list tables only using `database` specifier. "
+                "Include a `schema` argument."
+            )
+        database = (
+            sg.exp.Table(catalog=database, db=schema).sql(dialect=self.name) or None
```
cpcloud, line 509: Hm, I think we may need to use an API that handles the quoting here, like `sg.table(schema, db=database, quoted=True)`

gforsyth:
Rebased -- the changes here are not breaking, so can go into 7.1

I'm about to open a separate PR for the corresponding changes to `trino`, `snowflake`, and `bigquery` as those are breaking in the handling of keyword arguments to `list_tables`

ibis/backends/snowflake/__init__.py
gforsyth: hmm, yeah.  I'll need to also add a little extra handling for passing in a dotted path to `database` to support this being backwards compatible for the moment.  At least all of the extra code will go away in 8.x

cpcloud: Hm, I'm not sure we ever supported dotted paths with `database`. Might be okay to not handle that case.

gforsyth: ugh -- well, if we use `sg.table` then passing in a dotted path to `schema` will break

ibis/backends/mssql/__init__.py
```diff
@@ -84,6 +85,41 @@ def list_databases(self, like: str | None = None) -> list[str]:
...
         return self._scalar_query(sa.select(sa.func.schema_name()))
 
+    def list_tables(
+        self,
+        like: str | None = None,
+        database: str | None = None,
+        schema: str | None = None,
+    ) -> list[str]:
+        tablequery = sg.select("name").from_(
+            sg.exp.Table(this="tables", catalog=database, db="sys")
```
cpcloud, line 95: We should use `sg.table` here. `sg.exp.Table` doesn't automatically handle quoting:

```
In [4]: import sqlglot as sg

In [5]: sg.exp.Table(this="a b c", catalog="d e f").sql("duckdb")
Out[5]: 'd e f.a b c'

In [6]: sg.table("a b c", catalog="d e f").sql("duckdb")
Out[6]: '"d e f"."a b c"'
```

Generally wherever you don't need to handle parsing, prefer `sg.table` over `sg.exp.Table`

ibis/backends/trino/__init__.py
```diff
@@ -73,10 +74,49 @@ def current_schema(self) -> str:
...
+                ),
+                as_of="7.1",
+                removed_in="8.0",
+            )
+            database = sg.parse_one(database, into=sg.exp.Table).sql(dialect=self.name)
+        elif database is None and schema is not None:
+            database = sg.parse_one(schema, into=sg.exp.Table).sql(dialect=self.name)
+        else:
+            database = (
+                sg.table(schema, db=database, quoted=True).sql(dialect=self.name)
```
cpcloud, line 117: You shouldn't need to pass `quoted=True` here, Trino has sane quoting behavior.

ibis/backends/bigquery/__init__.py
```diff
@@ -528,8 +529,48 @@ def list_databases(self, like=None):
...
+                ),
+                as_of="7.1",
+                removed_in="8.0",
+            )
+            database = sg.parse_one(database, into=sg.exp.Table).sql(dialect=self.name)
+        elif database is None and schema is not None:
+            database = sg.parse_one(schema, into=sg.exp.Table).sql(dialect=self.name)
+        else:
+            database = (
+                sg.table(schema, db=database, quoted=True).sql(dialect=self.name)
```
cpcloud, line 570: `quoted=True` shouldn't be necessary here. `sg.table` will automatically determine whether the identifier needs to be quoted.

cpcloud (changes requested):
Couple of comments, but this is looking great.

gforsyth:
Thanks for the review @cpcloud -- fixed up the bits you flagged.

cpcloud (approved):
Awesome!

PR #7400, opened 2023-10-19
ci(docker): allow containers to start up a bit faster if possible

cpcloud:
Make the healthcheck policy easier to understand by removing timeouts and only using retry count plus healthcheck retry intervals. Also fixes an incorrectly specified dependency betweeen kudu and kudu-tserver: `kudu-tserver` should depend on `kudu`, not the other way around.


gforsyth (approved):
5 seconds faster on my machine :shipit:

PR #7404, opened 2023-10-19
chore: revert the apocalyse

cpcloud:
Revert the condapocalypse


.github/workflows/conda-lock.yml
```diff
@@ -12,7 +12,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - "3.9"
```
cpcloud, line 15: Removed this because we can't build the docs deps for 3.9 (because of limitations in plum [a dependency of `quartodoc`]).

cpcloud:
Merging this to see if I can get the `conda-lock.yml` action to work again.

PR #7407, opened 2023-10-20
feat(flink): implement windowed computations

chloeh13q:
Flink supports 4 types of computations based on windowing TVFs:
- Window aggregation
- Window top N
- Window join
- Window deduplication

They can be expressed via existing Ibis syntax, but it's worth considering a more friendly / alternative API for the user.

## Window aggregation
```python
expr = (
    t.window_by(time_col=t.i)
    .tumble(window_size=ibis.interval(minutes=15))
    .group_by(["window_start", "window_end", "g"])
    .aggregate(mean=_.d.mean())
)
```

## Window top N
```python
expr = t.window_by(time_col="i").tumble(
    window_size=ibis.interval(seconds=600),
)["a", "b", "c", "d", "g", "window_start", "window_end"]
expr = expr.mutate(
    rownum=ibis.row_number().over(
        group_by=["window_start", "window_end"], order_by=ibis.desc("g")
    )
)
expr = expr[expr.rownum <= 3]
```

Is there an OOTB shorthand for this? All I have found is the `topk()` function, which is defined on top of columns only.

## Window join
Will be covered in a separate PR.

## Window dedup
A special case of window top N in which the N is one and order by the processing time or event time.
```python
expr = t.window_by(time_col="i").tumble(
    window_size=ibis.interval(seconds=600),
)["a", "b", "c", "d", "g", "window_start", "window_end"]
expr = expr.mutate(
    rownum=ibis.row_number().over(
        group_by=["window_start", "window_end"], order_by=ibis.desc("g")
    )
)
expr = expr[expr.rownum <= 1]
```


deepyaman:
> They can be expressed via existing Ibis syntax, but it's worth considering a more friendly / alternative API for the user.
> 
> ## Window dedup
> A special case of window top N in which the N is one and order by the processing time or event time.
> 
> ```python
> expr = t.window_by(time_col="i").tumble(
>     window_size=ibis.interval(seconds=600),
> )["a", "b", "c", "d", "g", "window_start", "window_end"]
> expr = expr.mutate(
>     rownum=ibis.row_number().over(
>         group_by=["window_start", "window_end"], order_by=ibis.desc("g")
>     )
> )
> expr = expr[expr.rownum <= 1]
> ```

For example, for deduplication should be made to work using `distinct()`, I believe (see https://github.com/ibis-project/ibis/blob/f488f0ec63/ibis/backends/tests/test_generic.py#L1089 for example).

(The default translation using `distinct()` doesn't work, though.)

gforsyth (commented):
Hey @chloeh13q !  Sorry for the delay in getting to this.  Is this ready to go in?  I see the snapshot tests, have we confirmed that that's the expected output for these window computations?

chloeh13q:
@gforsyth Yep! The generated sql queries are correct for all the windowed computations (except for window joins, which I'll add in a separate PR). Do you have any thoughts on the API though? All the windowed computations here can already be expressed via existing Ibis syntax, but I'm wondering if we should introduce more friendly / alternative API for the user because it does delegate all the responsibilities to the user. I.e. the user would have to know how the Flink SQL syntax looks like for windowed computations in order to write them in Ibis. If you think the API looks good, then this is ready to go in!

gforsyth:
I think the windowed aggregation seems reasonable enough. The topk and dedup are both a little more verbose than I'd like. We can also use these as a starting point because they do, at least, work.

For dedup, I agree with @deepyaman something like `t.window_by(...).tumble(...).distinct()`

For topk maybe we can add that as a table operation that takes a column to order by and an optional `group_by`? I'll have to think about that a little more.

chloeh13q:
@gforsyth I've been experimenting with the example that uses `distinct()` as @deepyaman linked above. Underneath the hood, `distinct()` is implemented [using `arbitrary` and `group_by`](https://github.com/ibis-project/ibis/blob/master/ibis/expr/types/relations.py#L1183-L1188). The generated SQL looks like the following:
```
WITH t0 AS 
(SELECT t3."createTime" AS "createTime", t3."orderId" AS "orderId", t3."payAmount" AS "payAmount", t3."payPlatform" AS "payPlatform", t3."provinceId" AS "provinceId", CAST(:param_2 AS INTEGER) AS one 
FROM unbound_table_0 AS t3), 
t1 AS 
(SELECT t0."createTime" AS "createTime", t0."orderId" AS "orderId", t0."payAmount" AS "payAmount", t0."payPlatform" AS "payPlatform", t0."provinceId" AS "provinceId", t0.one AS one, row_number() OVER (ORDER BY t0."createTime" ASC) - :param_1 AS rownum 
FROM t0)
 SELECT t2."createTime", t2."orderId", t2."payAmount", t2."payPlatform", t2."provinceId", t2.one, t2.rownum 
FROM (SELECT t1."orderId" AS "orderId", first(t1."createTime") AS "createTime", first(t1."payAmount") AS "payAmount", first(t1."payPlatform") AS "payPlatform", first(t1."provinceId") AS "provinceId", first(t1.one) AS one, first(t1.rownum) AS rownum 
FROM t1 GROUP BY 1) AS t2
```
AFAIK, this syntax is not supported by Flink. It seems that there is only one way to do deduplication in Flink, and that is using `ROW_NUMBER()` with an `over` window, e.g.
```
SELECT *
  FROM (
    SELECT bidtime, price, item, supplier_id, window_start, window_end, 
      ROW_NUMBER() OVER (PARTITION BY window_start, window_end ORDER BY bidtime DESC) AS rownum
    FROM TABLE(
               TUMBLE(TABLE Bid, DESCRIPTOR(bidtime), INTERVAL '10' MINUTES))
  ) WHERE rownum <= 1;
```

One option would be to have `WindowedTable` return a class that inherits from `Table` and override the `distinct()` method there (since windowed TVFs only exist in Flink right now). 

Also cc @deepyaman @zhenzhongxu in case you guys have any thoughts on this.

zhenzhongxu:
> AFAIK, this syntax is not supported by Flink. 
This is an example where I believe behavior between streaming and batch would be different. When performing deduplication over a continuous stream, you would need to scope the logic over some sort of window. I think we'll need to consider how to override behaviors for different backends.

chloeh13q:
I created a [separate issue](https://github.com/ibis-project/ibis/issues/7556) for deduplication using `distinct()`. This current PR is ready for approval / merge.

gforsyth (commented):
The changes here look good -- unclear to me if the Flink failures in CI are related.  Does this need to be rebased on top of other Flink changes?

cpcloud (approved):
LGTM! Thanks!

PR #7409, opened 2023-10-20
feat(ux): show syntax-highlighted SQL if pygments is installed

cpcloud:
Adds syntax-highlighted output in the IPython console.

image

### DIFF
--- a/ibis/expr/sql.py
+++ b/ibis/expr/sql.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 import operator
 from functools import singledispatch
 from typing import IO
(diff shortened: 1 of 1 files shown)


cpcloud:
Slightly gnarlier example using BigQuery.

For this ibis code:

image

We get this SQL:

image

ibis/expr/sql.py
```diff
@@ -336,6 +336,23 @@ def __repr__(self) -> str:
...
 
+    def _repr_pretty_(self, p, cycle) -> str:
+        try:
+            from pygments import highlight
+            from pygments.formatters import TerminalFormatter
+            from pygments.lexers import SqlLexer
+        except ImportError:
+            output = str(self)
+        else:
+            output = highlight(
```
jcrist, line 347: Is there a chance that highlighting can error for any of the dialects we support?

Also, these lines appear uncovered, do we want to add a test for this?

jcrist (approved):
Nice! Just a question, otherwise LGTM

ibis/expr/sql.py
cpcloud: Possibly, but it's hard to say.

Picking an arbitrary example not common to most sql dialects (duckdb's struct literal syntax), highlighting seems to work:

image

I can catch `Exception` and warn in the case anything goes wrong (or maybe just avoid warning?)

cpcloud: Added a `contextlib.suppress` around the `highlight` call.

PR #7411, opened 2023-10-20
test(trino): use `tpch.tiny` for faster testing of tpch queries

cpcloud:
This is a little under 3x faster on my machine.

### DIFF
--- a/ibis/backends/tests/tpch/conftest.py
+++ b/ibis/backends/tests/tpch/conftest.py
@@ -59,7 +59,10 @@ def tpch_test(test: Callable[..., ir.Table]):
         result_expr = test(*args, **kwargs)
 
         result = result_expr.execute()
+        assert not result.empty
+
         expected = expected_expr.execute()
+        assert not expected.empty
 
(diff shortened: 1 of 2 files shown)


cpcloud:
Closing this out. I'm not sure why but the following query sometimes returns a single row and sometimes returns zero rows:

```sql
WITH t0 AS (
  SELECT
    suppkey,
    SUM(extendedprice * (1 - discount)) AS total_revenue
  FROM
    tpch.tiny.lineitem
  WHERE
    shipdate >= CAST('1996-01-01' AS date)
    AND shipdate < CAST('1996-04-01' AS date)
  GROUP BY
    1
)

SELECT * FROM t0
WHERE t0.total_revenue = (
  SELECT MAX(total_revenue) FROM t0
)
```

cpcloud:
The non-determinism can be addressed by using `DECIMAL` types. I opened [an issue](https://github.com/trinodb/trino/issues/19477) about this in the Trino repo.

cpcloud:
Opening a new PR for this.

Issue #6506, opened 2023-06-21
feat: Support `f(left: Table, right: Table)` functions as predicates to joins


NickCrews:
### Is your feature request related to a problem?

This is more of an issue of ergonomics than necessity.

I am making a framework based around ibis, where users need to be able to supply join conditions. For complex join conditions like `left.last_name == right.name_last` (eg with the need a reference to both the left and right table), a lambda is a nice choice for an API. Currently, I can easily wrap the current ibis API to make this work

```python
def join_on(x, y, predicate):
    return x.join(y, predicate(x, y))
```

but it would be cleaner if I could just pass this straight through.

### Describe the solution you'd like

the function must return one of the currently accepted types, eg

```python
def f(left: Table, right: Table) -> str | Sequence[str | tuple[str | ir.Column, str | ir.Column] | ir.BooleanColumn]:
    ...
```

### What version of ibis are you running?

main

### What backend(s) are you using, if any?

all

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

NickCrews:
Even more ergonomic would be if Ibis had special Deferred instances that could work for this, eg

```python
# I don't love "l" and "r", needs workshopping
from ibis import l, r

condition = l.user_id == r.user_id

t1 = ibis.read_parquet(p1)
t2 = ibis.read_parquet(p2)
t1.left_join(t2, condition)
```

NickCrews:
I think in that linked PR I discovered that this is already supported, just do `[(_. last_name, _.name_last)]`. I still would love it so that the outer list isn't needed. I think this is cleaner than a lambda, and I don't think a lambda is any more powerful than a tuple of deffereds, so I would say if we just clean up the outer list rough edge then this issue can be closed.

PR #7424, opened 2023-10-24
test: add more weird join predicate tests

NickCrews:
fixes https://github.com/ibis-project/ibis/issues/6506

As I was exploring https://github.com/ibis-project/ibis/issues/6506,
I deiscovered that there were all these join predicate
types that currenty work,
but only sort of "accidentally", and they
aren't actually tested for. This adds
tests for those.

Still TODO is to
- document and add examples for these
- determine if we can remove the need for the outer list for tuple keys. e.g. currently you need to to `[(_.col_in_left.upper(), _.col_in_right.upper())]` instead of merely `(_.col_in_left.upper(), _.col_in_right.upper())`, because the iterable of conditions gets promoted to a list, so ibis thinks the user really means two join keys: `[_.col_in_left.upper(), _.col_in_right.upper()]. I would love to remove the need for this outer list. Could we add a special case where if keys is a tuple of length 2, then assume this is a single join key, and don't `util.promote_list()` it?


cpcloud (approved):

cpcloud:
I think we can probably address point 2, but if we can hold off until 8.0 that would be ideal as there are some other refactors of joins happening right now. Is that okay?

cpcloud:
@NickCrews Would you mind changing the commit message to be a bit more specific than `weird`? Perhaps something like `test(api): flesh out join predicate tests`

NickCrews:
I will change the commit message and add to the docs.

Sounds good on the 8.0 target!

cpcloud (approved):

cpcloud:
@NickCrews I updated the commit message!

Please open the docs changes in a separate PR. Looking forward to them!

NickCrews:
Thanks! Im mostly out of office the next few weeks so I will be slow to respond. Looks great.

PR #7425, opened 2023-10-24
fix(deps): pin pyspark to a working version

cpcloud:
Pin PySpark to a known working upper bound


gforsyth:
Well, I suppose it's a good sign that there are a bunch of XPASS failures.

gforsyth:
lemme clean those up

Issue #7427, opened 2023-10-24
bug: broken geospatial `d_within` function


ncclementi:
### What happened?

The `d_within` function is supposed to accept `distance` (a float value), but it breaks with integers and floats. 

Integer behavior
```python
import ibis
t = ibis.table({"geo1": "geometry", "geo2": "geometry"}, name="t")
t.geo1.d_within(t.geo2, 3) # passing an int 
```

<details>
<summary>Traceback </summary>
<br>


```python-traceback
...
SignatureValidationError                  Traceback (most recent call last)                                                                  
Cell In[8], line 1                                                                                                                           
----> 1 t.geo1.d_within(t.geo2, distance=5)                                                                                                  
                                                                                                                                             
File ~/Documents/git/my_forks/ibis/ibis/expr/types/geospatial.py:198, in GeoSpatialValue.d_within(self, right, distance)                     
    179 def d_within(                                                                                                                        
    180     self,                                                                                                                            
    181     right: GeoSpatialValue,                                                                                                          
    182     distance: ir.FloatingValue,                                                                                                      
    183 ) -> ir.BooleanValue:                                                                                                                
    184     """Check if `self` is partially within `distance` from `right`.                                                                  
    185                                                                                                                                      
    186     Parameters                                                                                                                       
   (...)                                                                                                                                     
    196         Whether `self` is partially within `distance` from `right`.                                                                  
    197     """                                                                                                                              
--> 198     return ops.GeoDWithin(self, right, distance).to_expr()                                                                           
                                                                                                                                             
File ~/Documents/git/my_forks/ibis/ibis/common/bases.py:73, in AbstractMeta.__call__(cls, *args, **kwargs)                                   
     54 def __call__(cls, *args, **kwargs):                                                                                                  
     55     """Create a new instance of the class.                                                                                           
     56                                                                                                                                      
     57     The subclass may override the `__create__` classmethod to change the                                                             
   (...)                                                                                                                                     
     71     The newly created instance of the class. No extra initialization                                                                 
     72     """                                                                                                                                                                                                                                                                  [150/1167]
---> 73     return cls.__create__(*args, **kwargs)                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                           
File ~/Documents/git/my_forks/ibis/ibis/common/grounds.py:118, in Annotable.__create__(cls, *args, **kwargs)                                           
    115 @classmethod                                                                                                                                                                                                                                                                       
[...]

PR #7428, opened 2023-10-24
fix(repr): add dispatch for repr of GeoSpatialBinOps

gforsyth:
The `GeoUnaryOp`s are already handled by the rules for the parent
classes, but the `GeoSpatialBinOps` break because they often have an
additional keyword (`distance`) passed in that doesn't get handled by
the existing rule for binary operations.

xref #7427


cpcloud (approved):
A test could be nice, but not a blocker if you confirmed it locally.

gforsyth:
> A test could be nice, but not a blocker if you confirmed it locally.

Added some small tests and they turned up a small formatting issue with an extra trailing comma when geospatial bin ops don't have keyword arguments, so yay tests!

cpcloud (approved):

PR #7429, opened 2023-10-24
fix(ir): coerce integers passed to `Value[dt.Floating]` annotated values as `dt.float64`

kszucs:
Add special coercion paths for non-concrete datatypes like `dt.Floating` and `dt.Integer`.

Resolves https://github.com/ibis-project/ibis/issues/7427

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -499,11 +499,6 @@ class Bounds(NamedTuple):
 class Numeric(DataType):
     """Numeric types."""
 
-    @property
-    @abstractmethod
-    def largest(self) -> DataType:
-        """Return the largest type in this family."""
-
 
 @public
 class Integer(Primitive, Numeric):
@@ -631,11 +626,6 @@ class SignedInteger(Integer):
     """Signed integer values."""
 
     @property
-    def largest(self):
-        """Return the largest type of signed integer."""
-        return int64
-
-    @property
     def bounds(self):
         exp = self.nbytes * 8 - 1
         upper = (1 << exp) - 1
@@ -645,11 +635,6 @@ class SignedInteger(Integer):
 @public
 class UnsignedInteger(Integer):
     """Unsigned integer values."""
-
-    @property
-    def largest(self):
-        """Return the largest type of unsigned integer."""
-        return uint64
 
     @property
     def bounds(self):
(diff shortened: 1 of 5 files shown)


ibis/expr/operations/core.py
```diff
@@ -75,11 +75,17 @@ def __coerce__(
         if isinstance(value, Value):
             return value
 
-        try:
+        if T is dt.Integer:
```
kszucs, line 78: Special branches for numeric type families where we want to coerce to the smallest numeric types.

ibis/expr/datatypes/core.py
```diff
@@ -499,11 +499,6 @@ def __contains__(self, value: int) -> bool:
 class Numeric(DataType):
     """Numeric types."""
 
-    @property
-    @abstractmethod
-    def largest(self) -> DataType:
```
kszucs, line 504: Turns out largest is rarely used and it doesn't really make sense for decimals.

ibis/expr/datatypes/core.py
```diff
@@ -665,15 +650,10 @@ class Floating(Primitive, Numeric):
...
     column = "FloatingColumn"
 
-    @property
-    def largest(self):
-        """Return the largest type of floating point values."""
-        return float64
-
     @property
     @abstractmethod
     def nbytes(self) -> int:  # pragma: no cover
```
kszucs, line 655: It would be better to define `nbytes` on `dt.Numeric` but then we would need to have specific bitwidth decimals which could be worth doing based on the decimal precision, see https://clickhouse.com/docs/en/sql-reference/data-types/decimal

ibis/expr/operations/core.py
```diff
@@ -75,11 +75,17 @@ def __coerce__(
...
+            dtype = dt.infer(int(value))
+        elif T is dt.Floating:
+            dtype = dt.infer(float(value))
+        else:
```
kszucs, line 82: There are actually two branches handled by the `else`, when `T` is `None` in which case we could just use `dt.infer()` and when `T` is a `UnionType` e.g. `Value[dt.Numeric | dt.Boolean]`. The latter is used at just a couple of cases, so we could actually remove support for that.

ibis/tests/expr/test_decimal.py
```diff
@@ -68,7 +68,7 @@ def test_promote_decimal_type_max():
     [
         (None, None, (None, None)),
         (38, 2, (38, 2)),
-        (16, 2, (38, 2)),
+        (16, 2, (16, 2)),
```
kszucs, line 71: We no longer promote decimals to at least precision 38, though we can hardcode certain minimal precision per operation.

kszucs: Restored the original behaviour for `Sum`

cpcloud (approved):
Nice!

PR #7442, opened 2023-10-26
refactor(patterns): support more flexible sequence matching

kszucs:
- improve the implementation of `PatternSequence` and related supplemental patterns
- `TupleOf` pattern now aligns with `*Of` naming to match sequence of a single pattern rather than a tuple of patterns
- remove reduntand object creation logic for alternative but more performant implementations of certain patterns
- support the unpacking syntax to turn patterns into a `SomeOf` pattern
- short circuit `SequenceOf(Any())` to not traverse over the input sequence
- remove unused `PatternMapping` pattern


kszucs (commented):
+1

PR #7445, opened 2023-10-26
fix(datafusion): fix some temporal operations

mesejo:


cpcloud (approved):
Excellent! Hooray for UDF removal!

PR #7449, opened 2023-10-27
feat(duckdb): expose option to force reinstall extension

gforsyth:
This is apparently needed for (at least) Motherduck upgrades.


cpcloud (approved):
Seems legit!

PR #7453, opened 2023-10-27
feat(datafusion): add some array functions

mesejo:


ibis/backends/datafusion/compiler/values.py
```diff
@@ -710,3 +713,26 @@ def _if_else(op, *, bool_expr, true_expr, false_null_expr, **_):
...
+@translate_val.register(ops.ArrayConcat)
+def array_concat(op, *, arg, **_):
+    return F.array_concat(*arg)
+
+
+@translate_val.register(ops.ArrayPosition)
+def array_position(op, *, arg, other, **_):
+    return sg.exp.Coalesce(
+        this=F.array_position(arg, other),
+        expressions=sg.exp.Literal(this="0", is_string=False),
```
cpcloud, line 737: You should be able to use `sg.exp.convert(0)` here.

cpcloud: You might also be able to avoid the `Coalesce` construction and use `F.coalesce` instead, then you wouldn't need the convert call.

cpcloud (approved):
Couple of comments, nothing blocking.

ibis/backends/datafusion/compiler/values.py
cpcloud: Does `expressions` need to be a `list`?

mesejo: Thanks! I went with `F.coalesce(F.array_position(arg, other), 0)`

cpcloud (approved):

Issue #6399, opened 2023-06-08
Geospatial Support For DuckDB


wybert:
### Is your feature request related to a problem?

DuckDB has [spatial extension](https://duckdb.org/docs/extensions/spatial), I would like to see DuckDB spatial support in ibis

### Describe the solution you'd like

support all the functions like for PostGIS and HevayAIfunctions

### What version of ibis are you running?

5.1

### What backend(s) are you using, if any?

DuckDB

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
@wybert Welcome! 👋🏻

Thanks for the issue. This has definitely been in the back of our minds for a bit 🙂 

I think we were just waiting for someone to ask for it before doing the work!

I'm not sure this will make it into 6.0, and if not, then we should be able to get it in for the follow-up feature release (6.1).

NickCrews:
Is this [implemented](https://ibis-project.org/reference/expression-geospatial)? It's not on the [list of dtypes](https://ibis-project.org/reference/datatypes). I also don't understand how to construct these values (casting from strings, directly from binary files).

jcrist:
> Is this [implemented](https://ibis-project.org/reference/expression-geospatial)? It's not on the [list of dtypes](https://ibis-project.org/reference/datatypes).

Ibis has had geospatial support for a long time, this issue is for adding that support to duckdb. I'm not sure I fully understand your question, but the geospatial dtypes are in that list in the docs (e.g. `LineString`, `Point`, ...).

NickCrews:
> e.g. LineString, Point

Ah, I didn't see those. I thought there was a singular `GeoSpatialValue` type based on the API list, and I didn't see that in the dtype list. I get it now.

> this issue is for adding that support to duckdb

I get it now, TODO is just to add the translation stuff in `backends/duckdb/registry.py` (or similar). I thought those above dtypes and operations were added post-this-issue, and the docs and this issue just weren't updated. Sorry for the noise.

cpcloud:
Update on this: we're interested in adding support for geospatial, but we have a sizable refactor of the duckdb backend underway (#6996) that needs to land (to avoid relegating @gforsyth to rebase hell) before we can make progress on any new feature work for the duckdb backend.

cboettig:
:tada: are there updated  docs showing this new awesomeness in action?

cpcloud:
Not yet, but we have a blog post in PR!

PR #7454, opened 2023-10-27
feat(duckdb): Initial Geospatial support 

ncclementi:
Closes https://github.com/ibis-project/ibis/issues/6399

~~This PR add support for the functions that do not get wrapped `ST_ASEWKB` -> To be fixed in future PR~~
EDIT: We found a work around, that enables as to get implementation for the compatible fucntions in this table https://gist.github.com/ncclementi/fe53d75db2b9c049daefcbbbd5fe1ed4

TODO:
- [x] Clean up test suit: complete comparison against geopandas when possible and if not snapshot tests
- [x] Fix CI
- [x] Update d_within test (need to rebase against main)
- [x] Make sure geojson data is available on testing-data -> PR open https://github.com/ibis-project/testing-data/pull/5
- [x]  Use parametrize when possible on test suite and cleanup comments
- [x] add read_geo 
- [x] add `read_geo` test (xfail geometry column case)
- [x] modify test to read data from geojson on duckdb 
- [x] ~~Review and add duckdb support for existing ops supported by duckdb that are not part of postgres registry, like `ST_GeomFromText`~~ (out of scope, will do in a separate PR)


ncclementi:
Update: 
- After talking to Joris I was able to get two tests that were xfail to pass. There are small followup issue to report to duckdb related to the `unary_agg` and `area` function but they are not necessarily blocking this PR. 
- There are still some comments around that should be cleaned up but I think it'll make it easier to review if they are there at the moment. 
- The geojson data had the wrong `crs` . I fixed the crs in the data anyways to read correctly from geopandas, this PR needs to be merged https://github.com/ibis-project/testing-data/pull/7 . At the moment it's indifferent for duckdb but eventually it'll work see https://github.com/duckdb/duckdb_spatial/issues/156

ncclementi:
I'm having some problems with CI in this PR that I believe are unrelated to the PR itself. 
- https://github.com/ibis-project/ibis/actions/runs/6933267142/job/18859063928?pr=7454#step:10:29 related to scipy, I wonder if this is what's causing it https://github.com/ibis-project/ibis/commit/8bac145a574adbd148d847862a10fb4712168409 . 

- Also Scipy related https://github.com/ibis-project/ibis/actions/runs/6933267140/job/18859079625?pr=7454 

- I'm also having all the Nix jobs failing, it seems to be related to the size of zones.geojson, but this wasn't a problem before so I'm unsure what's happening here.

pyproject.toml
```diff
@@ -129,6 +129,8 @@ pytest-snapshot = ">=0.9.0,<1"
 pytest-xdist = ">=2.3.0,<4"
 requests = ">=2,<3"
 sqlalchemy = ">=1.4,<3"
+geopandas = ">=0.6.0,<1"
+shapely = "^2.0.2"
```
cpcloud, line 133: We already depend on these, so we don't need to list them twice.

cpcloud:
For the scipy issue, I suspect that if you:

1. remove the geopandas and shapely deps you added
2. rebase
3. run `poetry lock --no-update`

the issue will go away.

cpcloud:
This failure https://github.com/ibis-project/ibis/actions/runs/6933267141/job/18859238063?pr=7454 is caused by listing the shapely dependency twice

ncclementi:
Looks like if I remove `geopandas` then the nix jobs complain about not being able to import it see https://github.com/ibis-project/ibis/actions/runs/6935119825/job/18864639157?pr=7454

I still have the docs_pr job failing with the scipy issue https://github.com/ibis-project/ibis/actions/runs/6935119823/job/18864644110?pr=7454 as well as the generate poetry lockfile for sqlalchemy2. It looks like we might be installing a version of scipy that it doens't like it https://github.com/ibis-project/ibis/actions/runs/6935119837/job/18864639376?pr=7454#step:5:16296

cc: @cpcloud

cpcloud:
Ok, there were a couple of issues:

1. It looks like you relocked `poetry.lock` before I fixed the scipy issue, and the technically-correct-but-actually-broken lock file install scipy 1.6.1. I fixed that by running `git checkout upstream/master poetry.lock`.
2. The nix tests were failing for two reasons. The first reason is that the geospatial deps weren't installed. We could install them but since nix tests run in a network-free sandbox we won't be able to use the `spatial` duckdb extension which is the second reason nix tests were failing. There's no point in installing them only to not be able to run the tests so I made the geospatial tests use `pytest.importorskip` where appropriate. In general, optional features' dependencies (including in tests) like geospatial functionality should be gated behind `pytest.importorskip`.

ibis/backends/duckdb/__init__.py
```diff
@@ -494,6 +494,49 @@ def read_csv(
...
+        self,
+        source: str,
+        table_name: str | None = None,
+        **kwargs: Any,
+    ) -> ir.Table:
+        """Register a GEO file as a table in the current database.
+
+        Parameters
+        ----------
+        source
```
ncclementi, line 507: I just happen to try reading directly from a url, more specifically I tried and things work, I don't have a way from reading from s3 or the other cloud options, but given the nature of `st_read` my guess is this should work for any cloud format,
```python 
In [1]: import ibis
In [3]: t = con.read_geo("https://raw.githubusercontent.com/ibis-project/testing-data/master/geojson/zones.geojson")
In [4]: t.geom.area().to_pandas()
Out[5]: 
0      7.903953e+07
1      1.439095e+08
2      3.168508e+07
3      8.023733e+06
4      5.041488e+07
           ...     
258    3.974125e+07
259    4.263797e+07
260    3.468847e+06
261    6.957826e+06
262    6.636394e+06
Name: GeoArea(geom), Length: 263, dtype: float64
```

That being said there seems to be a regression that will be fixed  https://github.com/duckdb/duckdb_spatial/issues/182

ibis/backends/duckdb/tests/conftest.py
```diff
@@ -29,14 +48,27 @@ def preload(self):
...
                 """
             )
+        if geospatial_supported:
+            for table in TEST_TABLES_GEO:
+                yield (
+                    f"""
+                    INSTALL spatial;
+                    LOAD spatial;
+                    CREATE OR REPLACE TABLE {table} AS
+                    SELECT * FROM st_read('{geojson_dir / f'{table}.geojson'}')
```
gforsyth, line 69:
```suggestion
                    SELECT * FROM st_read('{geojson_dir / table}.geojson')
```
I _think_ this will work and avoids the nested f-string

ibis/backends/duckdb/tests/test_geospatial.py
```diff
@@ -0,0 +1,197 @@
...
+        param("geo_equals", {}, "geom_equals", {}, id="geo_eqs"),
+        param("covers", {}, "covers", {}, id="covers"),
+        param("covered_by", {}, "covered_by", {}, id="covered_by"),
+        param("crosses", {}, "crosses", {}, id="crosses"),
+        param("disjoint", {}, "disjoint", {}, id="disjoint"),
+        param("distance", {}, "distance", {}, id="distance"),
+        param("intersects", {}, "intersects", {}, id="intersects"),
+        param("overlaps", {}, "overlaps", {}, id="overlaps"),
+        param("touches", {}, "touches", {}, id="touches"),
+        param("within", {}, "within", {}, id="within"),
```
gforsyth, line 109: Do we need the future-proofing of the empty kwargs dict?  Or were there options that were being used that are gone now?

ibis/backends/duckdb/registry.py
```diff
@@ -28,6 +28,18 @@
...
+
+    class Geometry_WKB(Geometry):
+        as_binary = "ST_AsWKB"
+
+except ImportError:
+
+    class Geometry_WKB:
+        ...
+
+
```
gforsyth, line 42: I think we can import `Geometry_WKB` from `ibis.backends.duckdb.datatypes` and save the copy-paste here (that I did) -- and then one day when we hopefully rip this workaround out we only have to remove it in one spot.

gforsyth (commented):
This looks really nice @ncclementi !

I left a few comments, nothing major and nothing blocking.

There's one thing that I think we should probably handle -- currently the only place that the `spatial` extension gets loaded is if someone calls `read_geo`, but if they already have some geo tables in a DuckDB database, then we won't load the `spatial` extension and DuckDB will throw an error.

Ideally we wouldn't load `spatial` by default, but we will need a way to detect that there are geometry columns in a `ddb` file so we know when we need to.

I don't think that needs to block this PR, but it probably should block our next release until that gets sorted.

But really nice work here, and I think the cleanliness of the PR hides a lot of the blood, sweat and tears that went into this "good first issue."

gforsyth:
That oracle test failure is _definitely_ unrelated, I'll take a quick look at that.

ibis/backends/duckdb/tests/conftest.py
gforsyth: didn't even see that there -- nah, just leave it alone.  it works and it isn't user-facing

ibis/backends/duckdb/tests/test_geospatial.py
ncclementi: Good catch. I think we can remove them all, if in the future something gets added as a kwarg we can test for it and add it.

ncclementi:
@gforsyth Thank you for the review, I fix most of them, and left a question in one of the suggestions. 

About 

> There's one thing that I think we should probably handle -- currently the only place that the spatial extension gets loaded is if someone calls read_geo, but if they already have some geo tables in a DuckDB database, then we won't load the spatial extension and DuckDB will throw an error. 

>Ideally we wouldn't load spatial by default, but we will need a way to detect that there are geometry columns in a ddb file so we know when we need to.

> I don't think that needs to block this PR, but it probably should block our next release until that gets sorted.

That's a good point. How complicated would be to add such thing? I think depend on the answer we might want to add it here or leave it for a follow up.

gforsyth:
> How complicated would be to add such thing?

I'm not sure. I would start by connecting to an empty duckdb file, then reading in a geojson file and saving the contents to the duckdb database.

Then start a new session, connect to the duckdb database and start poking at it to see if there's any information other than "binary" in the table schema (there must be something, I'm thinking, or the geospatial extension wouldn't be able to tell which columns were "regular" binary and which ones were geo data)

ncclementi:
> I'm not sure. I would start by connecting to an empty duckdb file, then reading in a geojson file and saving the contents to the duckdb database. 
> Then start a new session, connect to the duckdb database and start poking at it to see if there's any information other than "binary" in the table schema (there must be something, I'm thinking, or the geospatial extension wouldn't be able to tell which columns were "regular" binary and which ones were geo data)

After our chat with the duckdb folks, looks like we can get the information of a `dtype = geometry` by doing `describe`. Then I believe we could check if `GEOMETRY` is on the dtypes and if so `load spatial`, now this will introduce an extra check for the every time we create a table, is that a good thing? Unless I'm missing something. 

@gforsyth would you be able to point on the right direction on where this would live, would it where we register a data source as a table? https://github.com/ibis-project/ibis/blob/d73c15610ecd772bcadabe769f4dc7605c4b99ba/ibis/backends/duckdb/__init__.py#L333

ncclementi:
@gforsyth the latest commit has the changes and corresponding tests for loading the spatial extension when a geo column exists. 

cc: @cpcloud for visibility.

ibis/backends/duckdb/registry.py
```diff
@@ -28,6 +28,14 @@
...
 from ibis.common.exceptions import UnsupportedOperationError
 
+try:
+    from ibis.backends.duckdb.datatypes import Geometry_WKB
+except ImportError:
+
+    class Geometry_WKB:
+        ...
+
+
```
gforsyth, line 38:
```suggestion
from ibis.backends.duckdb.datatypes import Geometry_WKB
```
we can rely on the import error handling in `datatypes.py` so only the import line is needed 

(it will also need to be  moved from where it is now that the try-except is gone or the linters will complain)

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -46,6 +46,68 @@ def test_read_parquet(data_dir):
...
+    gdf = t.head().to_pandas()
+    assert isinstance(gdf, gpd.GeoDataFrame)
+
+
+def test_read_geo(con, data_dir):
+    pytest.importorskip("geopandas")
+    t = con.read_geo(data_dir / "geojson" / "zones.geojson")
+    assert t.count().execute()
+
+
```
gforsyth, line 110:
```suggestion
```
It occurs to me that with the above `geopandas` test passing, there's nothing we need this test for, so I say we drop it.

ibis/backends/duckdb/__init__.py
```diff
@@ -1156,7 +1200,30 @@ def fetch_from_cursor(
...
+        a geospatial column is present in the dataframe.
+        """
+        import geopandas as gpd
+
+        geom_col = None
+        for name, dtype in schema.items():
+            if dtype.is_geospatial():
+                if not geom_col:
+                    geom_col = name
+                df[name] = gpd.GeoSeries.from_wkb(df[name])
```
gforsyth, line 1223: Quick explainer why we can't use the base implementation of this:

the base `to_geodataframe` uses shapely to take the output from postgis (or others), and converts it to a format compatible with a `geopandas.GeoArray`.  We already have the outputs from DuckDB spatial as a WKB, so we 
a) don't need to use shapely and 
b) can't use the base version because shapely gets angry if you try to `to_shape` on a WKB

gforsyth (changes requested):
Two small things to fix on my end and then I think this is good.

gforsyth:
For the `nix` failures for the `spatial` load test, you can add an `xfail` to the test like this:

```
from ibis.conftest import LINUX, SANDBOXED

@pytest.mark.xfail(
    LINUX and SANDBOXED,
    reason="nix on linux cannot download duckdb extensions or data due to sandboxing",
)
```

gforsyth:
And finally, this should get squashed down into fewer commits

ncclementi:
Looks like we have one failure on this job https://github.com/ibis-project/ibis/actions/runs/7039108148/job/19157653109?pr=7454 I guess it's because we are hacking around our way out of geoalchemy and sqlalchemy doesn't like it. 

Is there a way to skip this job for just that test? 

> And finally, this should get squashed down into fewer commits
Once I get everything green I'll get to squashing.

gforsyth:
> Is there a way to skip this job for just that test?

I think it's a legitimate failure -- in this case I think it's masking that we don't have `geoalchemy2` installed.

gforsyth:
I think for now we can add a `pytest.importorskip("geoalchemy2")` to that test

ncclementi:
> I think it's a legitimate failure -- in this case I think it's masking that we don't have geoalchemy2 installed.
 
But we did this on purpose, didn't we? I'm trying to see which one is the job to take a look but found it yet. 

> I think for now we can add a pytest.importorskip("geoalchemy2") to that test
👍

ibis/backends/duckdb/__init__.py
ncclementi: Just double checking, in case  I misunderstood.  Do we want to leave a comment in the code about this, or the comment in GH is just fine. ?

gforsyth: the GH comment is good, I don't think we need something inline

ibis/backends/duckdb/__init__.py
```diff
@@ -1255,3 +1322,16 @@ def _insert_dataframe(
...
+
+    def table(
+        self,
+        name: str,
+        database: str | None = None,
+        schema: str | None = None,
+    ) -> ir.Table:
+        expr = super().table(name=name, database=database, schema=schema)
+        # load geospatial only if geo columns
+        if any(typ.is_geospatial() for typ in expr.op().schema.types):
```
gforsyth, line 1334: If, at some point in the future, this starts to become a performance bottleneck, we can instead run this check in `SQL` and it will be faster (but not necessary to handle that now).

NickCrews:
Thank you for your work on this @ncclementi ! I am about to have to use this in an upcoming project!

PR #7467, opened 2023-10-31
feat(datafusion): add corr and covar

mesejo:
**Note:**

The `how` parameter seems to be ignored in the tests. Moreover, several backends do not define the parameter either. In particular, none of these backends implements it:
- [DuckDB](https://duckdb.org/docs/sql/aggregates.html#statistical-aggregates)
- [Postgres](https://www.postgresql.org/docs/9.4/functions-aggregate.html)
- [Polars](https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.corr.html) (mentions degrees of freedom but no `sample` or `population`) 
- [Pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.corr.html)

Does it make sense to have the `how` argument for `corr`?

### DIFF
--- a/ibis/backends/datafusion/compiler/values.py
+++ b/ibis/backends/datafusion/compiler/values.py
@@ -733,3 +733,34 @@ def array_concat(op, *, arg, **_):
 @translate_val.register(ops.ArrayPosition)
 def array_position(op, *, arg, other, **_):
     return F.coalesce(F.array_position(arg, other), 0)
+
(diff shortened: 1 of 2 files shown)


cpcloud (approved):

cpcloud:
@mesejo Can you open a separate issue for discussing the `how` argument of `corr`?

PR #7479, opened 2023-11-01
test(flink): deep dive on the tests marked for Flink in test_client.py

mfatihaktas:
[Previously](https://github.com/ibis-project/ibis/pull/6920), we made a pass over the tests under `ibis/ibis/backends/tests/` and marked them if they fail for the Flink backend. We are now making the second pass over these tests, and looking deeper into those marked previously. This PR contains the changes for the deep dive on `test_client.py`.

**Edit**: The scope of this PR got extended during the review process and after the merging with https://github.com/ibis-project/ibis/pull/7650. Now contains the following major changes:
- Adds support for creating `temporary view` for in-memory `obj` in `create_table()` and `create_view()`.
  - Raises error when the user-set `temp` argument does not match the supported behavior.
- Adds `rename_table()` for Flink backend.
- Adds unit tests that were previously missing for `create_table/view()` etc.
- Corrects some of the tests with `ibis.memtable()` by setting the backend using monkeypatch.

### DIFF
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -15,9 +15,12 @@ from ibis.backends.flink.compiler.core import FlinkCompiler
 from ibis.backends.flink.ddl import (
     CreateDatabase,
     CreateTableFromConnector,
+    CreateView,
     DropDatabase,
     DropTable,
+    DropView,
     InsertSelect,
+    RenameTable,
 )
 
 if TYPE_CHECKING:
@@ -116,18 +119,23 @@ class Backend(BaseBackend, CanCreateDatabase):
     def list_tables(
         self,
         like: str | None = None,
+        temp: bool = False,
         database: str | None = None,
         catalog: str | None = None,
     ) -> list[str]:
-        """Return the list of table names.
+        """Return the list of table/view names.
 
-        Return the list of table names in the specified database and catalog.
-        or the default one if no database/catalog is specified.
+        Return the list of table/view names in the `database` and `catalog`. If
+        `database`/`catalog` are not specified, their default values will be
+        used. Temporary tables can only be listed for the default database and
+        catalog, hence `database` and `catalog` are ignored if `temp` is True.
 
         Parameters
         ----------
         like : str, optional
             A pattern in Python's regex format.
+        temp : bool, optional
+            Whether to list temporary tables or permanent tables.
         database : str, optional
             The database to list tables of, if not the current one.
         catalog : str, optional
@@ -136,19 +144,29 @@ class Backend(BaseBackend, CanCreateDatabase):
         Returns
         -------
         list[str]
-            The list of the table names that match the pattern `like`.
+            The list of the table/view names that match the pattern `like`.
         """
-        tables = self._table_env._j_tenv.listTables(
-            catalog or self.current_catalog,
-            database or self.current_database,
-        )  # this is equivalent to the SQL query string `SHOW TABLES FROM|IN`,
+        catalog = catalog or self.current_catalog
+        database = database or self.current_database
+
+        # The following is equivalent to the SQL query string `SHOW TABLES FROM|IN`,
         # but executing the SQL string directly yields a `TableResult` object
+        if temp:
+            # Note (mehmet): TableEnvironment does not provide a function to list
+            # the temporary tables in a given catalog and database.
+            # Ref: https://nightlies.apache.org/flink/flink-docs-master/api/java/org/apache/flink/table/api/TableEnvironment.html
+            tables = self._table_env.list_temporary_tables()
+        else:
+            # Note (mehmet): `listTables` returns both tables and views.
+            # Ref: Docstring for pyflink/table/table_environment.py:list_tables()
+            tables = self._table_env._j_tenv.listTables(catalog, database)
+
         return self._filter_with_like(tables, like)
 
-    def _list_views(
+    def list_views(
         self,
         like: str | None = None,
-        temporary: bool = True,
+        temp: bool = False,
     ) -> list[str]:
         """Return the list of view names.
 
@@ -158,7 +176,7 @@ class Backend(BaseBackend, CanCreateDatabase):
         ----------
         like : str, optional
             A pattern in Python's regex format.
-        temporary : bool, optional
+        temp : bool, optional
             Whether to list temporary views or permanent views.
 
         Returns
@@ -166,17 +184,19 @@ class Backend(BaseBackend, CanCreateDatabase):
         list[str]
             The list of the view names that match the pattern `like`.
         """
-        if temporary:
+
+        if temp:
             views = self._table_env.list_temporary_views()
         else:
             views = self._table_env.list_views()
+
         return self._filter_with_like(views, like)
 
     def _fully_qualified_name(
         self,
         name: str,
-        database: str | None,
-        catalog: str | None,
+        database: str | None = None,
+        catalog: str | None = None,
     ) -> str:
         if is_fully_qualified(name):
             return name
@@ -279,6 +299,7 @@ class Backend(BaseBackend, CanCreateDatabase):
         table_expr = expr.as_table()
         sql = self.compile(table_expr, **kwargs)
         df = self._table_env.sql_query(sql).to_pandas()
+
         return expr.__pandas_result__(df)
 
     def create_table(
@@ -296,13 +317,15 @@ class Backend(BaseBackend, CanCreateDatabase):
     ) -> ir.Table:
         """Create a new table in Flink.
 
-        Note that: in Flink, tables can be either virtual (VIEWS) or regular
-        (TABLES).
-
+        In Flink, tables can be either virtual (VIEWS) or regular (TABLES).
         VIEWS can be created from an existing Table object, usually the result
         of a Table API or SQL query. TABLES describe external data, such as a
         file, database table, or message queue. In other words, TABLES refer
         explicitly to tables constructed directly from source/sink connectors.
+
+        When `obj` is in-memory (e.g., Dataframe), currently this function can
+        create only a TEMPORARY VIEW. If `obj` is in-memory and `temp` is False,
+        it will raise an error.
 
         Parameters
         ----------
@@ -341,44 +364,90 @@ class Backend(BaseBackend, CanCreateDatabase):
         import pyarrow_hotfix  # noqa: F401
 
         import ibis.expr.types as ir
-        from ibis.backends.flink.datatypes import FlinkRowSchema
 
         if obj is None and schema is None:
-            raise exc.IbisError("The schema or obj parameter is required")
+            raise exc.IbisError("`schema` or `obj` is required")
+        if isinstance(obj, (pd.DataFrame, pa.Table)) and not temp:
+            raise exc.IbisError(
+                "`temp` cannot be False when `obj` is in-memory. "
+                "Currently can create only TEMPORARY VIEW for in-memory data."
+            )
 
-        # in-memory data is created as views in `pyflink`
-        elif obj is not None:
-            if isinstance(obj, pa.Table):
-                obj = obj.to_pandas()
-            if isinstance(obj, pd.DataFrame):
-                table = self._table_env.from_pandas(
-                    obj, FlinkRowSchema.from_ibis(schema)
+        if overwrite:
+            if self.list_tables(like=name, temp=temp):
+                self.drop_table(
+                    name=name,
+                    catalog=catalog,
+                    database=database,
+                    temp=temp,
+                    force=True,
                 )
-            if isinstance(obj, ir.Table):
-                table = obj
-            return self.create_view(name, table, database=database, overwrite=overwrite)
 
-        # external data is created as tables in `pyflink`
-        else:
+        # In-memory data is created as views in `pyflink`
+        if obj is not None:
+            if isinstance(obj, pd.DataFrame):
+                dataframe = obj
+
+            elif isinstance(obj, pa.Table):
+                dataframe = obj.to_pandas()
+
+            elif isinstance(obj, ir.Table):
+                # Note (mehmet): If obj points to in-memory data, we create a view.
+                # Other cases are unsupported for now, e.g., obj is of UnboundTable.
+                # See TODO right below for more context on how we handle in-memory data.
+                op = obj.op()
+                if isinstance(op, ops.InMemoryTable):
+                    dataframe = op.data.to_frame()
+                else:
+                    raise exc.IbisError(
+                        "`obj` is of type ibis.expr.types.Table but it is not in-memory. "
+                        "Currently, only in-memory tables are supported. "
+                        "See ibis.memtable() for info on creating in-memory table."
+                    )
+            else:
+                raise exc.IbisError(f"Unsupported `obj` type: {type(obj)}")
+
+            # TODO (mehmet): Flink requires a source connector to create regular tables.
+            # In-memory data can only be created as a view (virtual table). So we decided
+            # to create views for in-memory data. Ideally, this function should only create
+            # tables. However, for that, we would need the notion of a "default" table,
+            # which may not be ideal. We plan to get back to this later.
+            # Ref: https://github.com/ibis-project/ibis/pull/7479#discussion_r1416237088
+            return self.create_view(
+                name=name,
+                obj=dataframe,
+                schema=schema,
+                database=database,
+                catalog=catalog,
+                temp=temp,
+                overwrite=overwrite,
+            )
+
+        # External data is created as tables in `pyflink`
+        else:  # obj is None, schema is not None
             if not tbl_properties:
                 raise exc.IbisError(
-                    "tbl_properties is required when creating table with schema"
+                    "`tbl_properties` is required when creating table with schema"
                 )
-            if overwrite:
-                self.drop_table(
-                    name=name, catalog=catalog, database=database, force=True
-                )
+            elif (
+                "connector" not in tbl_properties or tbl_properties["connector"] is None
+            ):
+                raise exc.IbisError("connector must be defined in `tbl_properties`")
 
+            # TODO (mehmet): Given that we rely on default catalog if one is not specified,
+            # is there any point to support temporary tables?
             statement = CreateTableFromConnector(
                 table_name=name,
                 schema=schema,
                 tbl_properties=tbl_properties,
                 watermark=watermark,
-                temp=temp,
+                temporary=temp,
                 database=database,
                 catalog=catalog,
             )
-            self._exec_sql(statement.compile())
+            sql = statement.compile()
+            self._exec_sql(sql)
+
             return self.table(name, database=database, catalog=catalog)
 
     def drop_table(
@@ -401,7 +470,7 @@ class Backend(BaseBackend, CanCreateDatabase):
         catalog
             Name of the catalog where the table exists, if not the default.
         temp
-            Whether a table is temporary or not.
+            Whether the table is temporary or not.
         force
             If `False`, an exception is raised if the table does not exist.
         """
@@ -410,20 +479,52 @@ class Backend(BaseBackend, CanCreateDatabase):
             database=database,
             catalog=catalog,
             must_exist=not force,
-            temp=temp,
+            temporary=temp,
         )
         self._exec_sql(statement.compile())
+
+    def rename_table(
+        self,
+        old_name: str,
+        new_name: str,
+        force: bool = True,
+    ) -> None:
+        """Rename an existing table.
+
+        Parameters
+        ----------
+        old_name
+            The old name of the table.
+        new_name
+            The new name of the table.
+        force
+            If `False`, an exception is raised if the table does not exist.
+        """
+        statement = RenameTable(
+            old_name=old_name,
+            new_name=new_name,
+            must_exist=not force,
+        )
+        sql = statement.compile()
+        self._exec_sql(sql)
 
     def create_view(
         self,
         name: str,
         obj: pd.DataFrame | ir.Table,
         *,
+        schema: sch.Schema | None = None,
         database: str | None = None,
         catalog: str | None = None,
+        force: bool = False,
+        temp: bool = False,
         overwrite: bool = False,
     ) -> ir.Table:
-        """Create a new view from an expression.
+        """Create a new view from a dataframe or table.
+
+        When `obj` is in-memory (e.g., Dataframe), currently this function can
+        create only a TEMPORARY VIEW. If `obj` is in-memory and `temp` is False,
+        it will raise an error.
 
         Parameters
         ----------
@@ -431,29 +532,72 @@ class Backend(BaseBackend, CanCreateDatabase):
             Name of the new view.
         obj
             An Ibis table expression that will be used to create the view.
+        schema
+            The schema for the new view.
         database
             Name of the database where the view will be created, if not
             provided the database's default is used.
         catalog
             Name of the catalog where the table exists, if not the default.
+        force
+            If `False`, an exception is raised if the table is already present.
+        temp
+            Whether the table is temporary or not.
         overwrite
-            Whether to clobber an existing view with the same name.
+            If `True`, remove the existing view, and create a new one.
 
         Returns
         -------
         Table
             The view that was created.
         """
-        if isinstance(obj, ir.Table):
-            # TODO(chloeh13q): implement CREATE VIEW for expressions
-            raise NotImplementedError
+        import pandas as pd
 
-        if overwrite and name in self._list_views():
-            self.drop_view(name=name, catalog=catalog, database=database, force=True)
+        from ibis.backends.flink.datatypes import FlinkRowSchema
 
-        qualified_name = self._fully_qualified_name(name, database, catalog)
-        self._table_env.create_temporary_view(qualified_name, obj)
-        return self.table(name, database=database, catalog=catalog)
+        if isinstance(obj, pd.DataFrame) and not temp:
+            raise exc.IbisError(
+                "`temp` cannot be False when `obj` is in-memory. "
+                "Currently supports creating only temporary view for in-memory data."
+            )
+
+        if overwrite and self.list_views(like=name, temp=temp):
+            self.drop_view(
+                name=name,
+                database=database,
+                catalog=catalog,
+                temp=temp,
+                force=True,
+            )
+
+        if isinstance(obj, pd.DataFrame):
+            qualified_name = self._fully_qualified_name(name, database, catalog)
+            if schema:
+                table = self._table_env.from_pandas(
+                    obj, FlinkRowSchema.from_ibis(schema)
+                )
+            else:
+                table = self._table_env.from_pandas(obj)
+            # Note (mehmet): We use `create_temporary_view` here instead of `register_table`
+            # as suggested in PyFlink source code due to the deprecation of `register_table`.
+            self._table_env.create_temporary_view(qualified_name, table)
+
+        elif isinstance(obj, ir.Table):
+            query_expression = self.compile(obj)
+            statement = CreateView(
+                name=name,
+                query_expression=query_expression,
+                database=database,
+                can_exist=force,
+                temporary=temp,
+            )
+            sql = statement.compile()
+            self._exec_sql(sql)
+
+        else:
+            raise exc.IbisError(f"Unsupported `obj` type: {type(obj)}")
+
+        return self.table(name=name, database=database, catalog=catalog)
 
     def drop_view(
         self,
@@ -461,6 +605,7 @@ class Backend(BaseBackend, CanCreateDatabase):
         *,
         database: str | None = None,
         catalog: str | None = None,
+        temp: bool = False,
         force: bool = False,
     ) -> None:
         """Drop a view.
@@ -473,14 +618,22 @@ class Backend(BaseBackend, CanCreateDatabase):
             Name of the database where the view exists, if not the default.
         catalog
             Name of the catalog where the view exists, if not the default.
+        temp
+            Whether the view is temporary or not.
         force
             If `False`, an exception is raised if the view does not exist.
         """
-        qualified_name = self._fully_qualified_name(name, database, catalog)
-        if not (self._table_env.drop_temporary_view(qualified_name) or force):
-            raise exc.IntegrityError(f"View {name} does not exist.")
-
         # TODO(deepyaman): Support (and differentiate) permanent views.
+
+        statement = DropView(
+            name=name,
+            database=database,
+            catalog=catalog,
+            must_exist=(not force),
+            temporary=temp,
+        )
+        sql = statement.compile()
+        self._exec_sql(sql)
 
     @classmethod
     @lru_cache
--- a/ibis/backends/flink/ddl.py
+++ b/ibis/backends/flink/ddl.py
@@ -6,9 +6,11 @@ import sqlglot as sg
 
 import ibis.expr.schema as sch
 from ibis.backends.base.sql.ddl import (
+    CreateTable,
     CreateTableWithSchema,
     DropObject,
     InsertSelect,
+    RenameTable,
     _CreateDDL,
     _format_properties,
     _is_quoted,
@@ -35,9 +37,12 @@ def _format_schema_element(name, t):
 
 def type_to_flink_sql_string(tval):
     if tval.is_timestamp():
-        return f"timestamp({tval.scale})"
+        return f"TIMESTAMP({tval.scale})"
     else:
-        return type_to_sql_string(tval)
+        sql_string = type_to_sql_string(tval)
+        if not tval.nullable:
+            sql_string += " NOT NULL"
+        return sql_string
 
 
 def _format_watermark_strategy(watermark: Watermark) -> str:
@@ -85,7 +90,7 @@ class CreateTableFromConnector(
         watermark: Watermark | None = None,
         database: str | None = None,
         catalog: str | None = None,
-        temp: bool = False,
+        temporary: bool = False,
         **kwargs,
     ):
         super().__init__(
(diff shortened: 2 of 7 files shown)


ibis/backends/flink/ddl.py
```diff
@@ -144,6 +149,43 @@ def _pieces(self):
...
+        )
+        self.name = name
+        self.query_expression = query_expression
+        self.catalog = catalog
+        self.temporary = temporary
+
+    @property
+    def _prefix(self):
+        if self.temporary:
+            return f"CREATE TEMPORARY VIEW"
```
deepyaman, line 175:
```suggestion
            return "CREATE TEMPORARY VIEW"
```

ibis/backends/flink/ddl.py
```diff
@@ -144,6 +149,43 @@ def _pieces(self):
...
+        else:
+            return f"CREATE VIEW"
```
deepyaman, line 177:
```suggestion
            return "CREATE VIEW"
```

deepyaman (commented):
Why is one of the tests failing, saying you need DuckDB? Haven't looked into it yet, but curious if you have.

ibis/backends/flink/ddl.py
mfatihaktas: Thanks, done.

ibis/backends/flink/ddl.py
mfatihaktas: Thanks, done.

mfatihaktas:
> Why is one of the tests failing, saying you need DuckDB? Haven't looked into it yet, but curious if you have.

I looked over [this error](https://github.com/ibis-project/ibis/actions/runs/6737030093/job/18313637452#step:13:190), but I could not reproduce it on my machine. I will look more.

ibis/backends/flink/__init__.py
```diff
@@ -331,13 +335,30 @@ def create_table(
...
-            if not tbl_properties:
-                raise exc.IbisError(
-                    "tbl_properties is required when creating table with schema"
+            if schema is not None and not schema.equals(obj.schema()):
+                raise exc.IbisTypeError(
+                    "Provided schema and Ibis table schema are incompatible. Please "
+                    "align the two schemas, or provide only one of the two arguments."
                 )
+
+            # TODO (mehmet): Is it OK to return a view here?
```
deepyaman, line 344:
```suggestion
```
I think so, based on the comment above:

```
        Note that: in Flink, tables can be either virtual (VIEWS) or regular
        (TABLES).
        VIEWS can be created from an existing Table object, usually the result
        of a Table API or SQL query. TABLES describe external data, such as a
        file, database table, or message queue. In other words, TABLES refer
        explicitly to tables constructed directly from source/sink connectors.
```

ibis/backends/flink/__init__.py
```diff
@@ -331,13 +335,30 @@ def create_table(
...
+                temporary=temp,
+                overwrite=overwrite,
+            )
+
+        if schema is not None:
+            tbl_properties = tbl_properties or {
+                "connector": "filesystem",
+                "path": "file:///tmp",
+                "format": "parquet",
+            }
```
deepyaman, line 361: We had discussed potentially making `tbl_properties` optional, right (not necessarily in this PR)? If that's the case, would you still have this block? I don't see other backends referencing `file:///tmp` or something in a hardcoded way.

ibis/backends/flink/__init__.py
```diff
@@ -384,12 +406,38 @@ def drop_table(
...
         self,
         name: str,
-        obj: ir.Table,
-        *,
+        obj: ir.Table | None = None,
+        query_expression: str | None = None,
         database: str | None = None,
+        catalog: str | None = None,
+        temporary: bool = False,
+        if_not_exists: bool = False,
```
deepyaman, line 440: Where are all these extra arguments coming from? My understanding is that something like `create_view` would provide more-or-less the same interface as other backends.

ibis/backends/tests/test_client.py
```diff
@@ -216,9 +216,10 @@ def test_sql(backend, con):
 }
 
 
-@mark.notimpl(["datafusion", "druid", "flink"])
+@mark.notimpl(["datafusion", "druid"])
 def test_create_table_from_schema(con, new_schema, temp_table):
     new_table = con.create_table(temp_table, schema=new_schema)
+
```
deepyaman, line 222:
```suggestion
```
Unnecessary diff

ibis/backends/tests/test_client.py
```diff
@@ -303,23 +303,29 @@ def test_rename_table(con, temp_table, temp_table_orig):
...
 
-@mark.notimpl(["datafusion", "druid", "flink", "polars"])
+@mark.notimpl(["datafusion", "druid", "polars"])
+@pytest.mark.broken(
+    ["flink"],
+    raises=ValueError,
+    reason=(
+        "table `FUNCTIONAL_ALLTYPES` does not exist"
+        "Note (mehmet): Not raised when only this test function is executed, "
+        "but can be reproduced by running all the test functions in this file."
```
deepyaman, line 325: This is problematic, because it won't pass CI. We don't currently have an issue when running something in a single file causes that issue.

ibis/backends/flink/__init__.py
```diff
@@ -435,11 +510,17 @@ def drop_view(
             If `False`, an exception is raised if the view does not exist.
         """
         qualified_name = self._fully_qualified_name(name, database, catalog)
-        if not (self._table_env.drop_temporary_view(qualified_name) or force):
```
deepyaman, line 438: Is there a reason you do everything with SQL, instead of some of the PyFlink functions? Not sure if something you and @chloeh13q discussed/based on how she's been implementing things, or if it's based on some investigation--just curious.

ibis/backends/flink/__init__.py
deepyaman: It would be helpful for me if you can point to a backend you may have referenced doing something like this off of; I'm just curious if other backends have different backend-specific `create_view` signatures, or just this one. I did see that the test code is unchanged, so that's good.

ibis/backends/flink/__init__.py
mfatihaktas: Good question. I have actually spent quite some time looking for a way to create a table without the `connector` option, but could not find any way. In my current understanding, tables have to be based on a "persistent" storage backend. There does not seem to be a "special connector" to data in memory. The closest to this is the [managed table](https://nightlies.apache.org/flink/flink-table-store-docs-release-0.1/docs/development/overview/#managed-table) as far as I could find.

ibis/backends/flink/__init__.py
mfatihaktas: > Where are all these extra arguments coming from?

I added the extra arguments according to the [doc](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sql/create/#create-view) for `create view`.

> It would be helpful for me if you can point to a backend you may have referenced doing something like this off of ...

I went over the definition of `create_view` for other backends, but could not spot any other with different signature.
I added the optional parameter `query_expression` per the [doc](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sql/create/#create-view) to make the `create_view` more flexible, but that parameter is currently not used. I can remove it to make it consistent with other backends. Please let me know.

ibis/backends/flink/__init__.py
mfatihaktas: I prefer SQL over PyFlink functions because I find the doc for Flink SQL more clear. In my experience, it is much easier to understand what we can do or cannot do with Flink while reading on the SQL doc relative to the PyFlink doc. I have not discussed this with @chloeh13q though.

ibis/backends/flink/__init__.py
```diff
@@ -331,13 +335,28 @@ def create_table(
             # INTO ... VALUES to keep things consistent
             self._table_env.create_temporary_view(qualified_name, table)
         if isinstance(obj, ir.Table):
-            # TODO(chloeh13q): implement CREATE TABLE for expressions
-            raise NotImplementedError
-        if schema is not None:
-            if not tbl_properties:
-                raise exc.IbisError(
-                    "tbl_properties is required when creating table with schema"
+            if schema is not None and not schema.equals(obj.schema()):
```
chloeh13q, line 338: I think the expected behavior is that "only one of `schema` or `obj` can be provided" (see the docstring)

ibis/backends/flink/__init__.py
chloeh13q: > I have actually spent quite some time looking for a way to create a table without the connector option, but could not find any way.

Do you mean in Flink? I think it is technically possible, as you mentioned, if the table is a managed one. I'm on the same page as Deepyaman; I'm also not sure that I understand this implementation / agree that this is a clean implementation here.

ibis/backends/flink/__init__.py
```diff
@@ -384,12 +404,38 @@ def drop_table(
...
+        must_exist: bool = True,
+    ) -> None:
+        """Rename an existing table.
+
+        Parameters
+        ----------
+        old_name
+            The old name of the table.
+        new_name
+            The new name of the table.
```
chloeh13q, line 420: Missing parameter description for `must_exist`

ibis/backends/flink/__init__.py
chloeh13q: Yeah I'd vote for removing `query_expression`, at least for the time being anyway. For the other arguments, they sound reasonable to me because I also had to add some arguments that are not in other backends, but I'd just be cautious that the naming of these parameters are consistent across methods (e.g. `if_not_exists`, `must_exist`, etc).

ibis/backends/flink/__init__.py
```diff
@@ -403,15 +449,42 @@ def create_view(
         database
             Name of the database where the view will be created, if not
             provided the database's default is used.
+        catalog
+            Name of the catalog where the table exists, if not the default.
+        if_not_exists
```
chloeh13q, line 454: Missing documentation for `temporary`

chloeh13q (commented):
Left some comments but the implementation looks mostly reasonable to me. I do see a lot of check warnings that the new lines are not covered by tests; I'd just make sure that everything is properly unit tested, especially where you're introducing different signature for a method in the Flink backend, each parameter should be tested properly to the extent possible (you can write custom backend tests in the `flink` folder, where I added tests for other DDL methods).

ibis/backends/flink/__init__.py
mfatihaktas: Thanks. Then we have two options:
1. Revert the change, so we have the following defined in `create_table()`:
```python
        if schema is not None:
            if not tbl_properties:
                raise exc.IbisError(
                    "tbl_properties is required when creating table with schema"
                )
```
which will lead to failures in > 20 tests.

2. Implement the change with [managed table](https://nightlies.apache.org/flink/flink-table-store-docs-release-0.1/docs/development/overview/#managed-table). However, enabling managed table will require additional steps/changes as described [here](https://nightlies.apache.org/flink/flink-table-store-docs-release-0.1/docs/development/overview/#setup-table-store).

@chloeh13q @deepyaman Which option makes more sense to you?

ibis/backends/flink/__init__.py
mfatihaktas: Thanks, removed `query_expression`.

> but I'd just be cautious that the naming of these parameters are consistent across methods (e.g. if_not_exists, must_exist, etc).

Good point. I spent some time contemplating on `if_not_exists` vs. `must_exist` vs. `force`. I decided to name these with "most explicit" names possible, but I guess here consistency across functions is more important than being explicit. This is why I now changed them to `force` in the related functions `create_view()`, `rename_table()`, `drop_table()`. Please let me know if you think otherwise.

ibis/backends/flink/__init__.py
mfatihaktas: Thanks, but there are several places in the `common tests` where both `schema` and `obj` are provided. We need to decide whether to 
(1) check if only one of them is provided and raise an error otherwise, 
(2) support providing both and check if the schemas match. 
Which one makes more sense to you? (The second increases test coverage.)

ibis/backends/flink/__init__.py
mfatihaktas: Thanks, added.

mfatihaktas:
> I do see a lot of check warnings that the new lines are not covered by tests; I'd just make sure that everything is properly unit tested, especially where you're introducing different signature for a method in the Flink backend, each parameter should be tested properly to the extent possible (you can write custom backend tests in the `flink` folder, where I added tests for other DDL methods).

@chloeh13q Thanks, great point. I will add these unit tests once I get the approval from you and @deepyaman about the function signatures :)

ibis/backends/flink/__init__.py
chloeh13q: I agree with @deepyaman .

ibis/backends/flink/__init__.py
chloeh13q: I checked the implementation in some of the other backends. Did you get this from the BigQuery backend? I guess it's fine if this is how another backend does it

mfatihaktas: Thanks, implemented `option 1`.

mfatihaktas:
> > Why is one of the tests failing, saying you need DuckDB? Haven't looked into it yet, but curious if you have.
> 
> I looked over [this error](https://github.com/ibis-project/ibis/actions/runs/6737030093/job/18313637452#step:13:190), but I could not reproduce it on my machine.
> 
> The issue is due to `options.default_backend` being `None` in `ibis/config.py: _default_backend()`:
> 
> ```python
> def _default_backend() -> Any:
>     if (backend := options.default_backend) is not None:
>         return backend
> 
>     try:
>         import duckdb as _  # noqa: F401
>     except ImportError:
>         raise com.IbisError(
>             """\
> You have used a function that relies on the default backend, but the default
> backend (DuckDB) is not installed.
> 
> You may specify an alternate backend to use, e.g.
> 
> ibis.set_backend("polars")
> 
> or to install the DuckDB backend, run:
> 
>     pip install 'ibis-framework[duckdb]'
> 
> or
> 
>     conda install -c conda-forge ibis-framework
> 
> For more information on available backends, visit https://ibis-project.org/install
> """
>         )
> 
>     import ibis
> 
>     options.default_backend = con = ibis.duckdb.connect(":memory:")
>     return con
> ```
> 
> At many places in `test_client.py`, `default_backend` is set with monkeypatch, e.g.,
> 
> ```python
> monkeypatch.setattr(ibis.options, "default_backend", ibis.pandas)
> ```
> 
> The issue might be due to test execution order.

@gforsyth Do you have any idea on what might be causing this issue? 
(The [full log](https://github.com/ibis-project/ibis/actions/runs/7066658959/job/19238925729?pr=7479) for reference.)

mfatihaktas:
> > I do see a lot of check warnings that the new lines are not covered by tests; I'd just make sure that everything is properly unit tested, especially where you're introducing different signature for a method in the Flink backend, each parameter should be tested properly to the extent possible (you can write custom backend tests in the `flink` folder, where I added tests for other DDL methods).
> 
> @chloeh13q Thanks, great point. I will add these unit tests once I get the approval from you and @deepyaman about the function signatures :)

@chloeh13q I added unit tests exercising the newly added lines in `test_ddl.py`, but the not-covered-by-tests warnings are still there. I am investigating why this happens, please let me know if you have any idea on why.

gforsyth:
> @gforsyth Do you have any idea on what might be causing this issue? (The [full log](https://github.com/ibis-project/ibis/actions/runs/7066658959/job/19238925729?pr=7479) for reference.)

If you create a `memtable`, by default, Ibis will use `DuckDB` to materialize it, so the first line of the test in question is 

```
t0 = ibis.memtable({"a": [1, 2, 3]})
```

And even though you then pass it to `ddl_con`, it is going to try to first register the `memtable` with the default backend.

You need to set `ibis.set_backend(ddl_con)` before you create the `memtable` (or you can do it with the `monekypatch` line as above).

It's very possible some of these tests are passing "incorrectly" because it's actually running on `DuckDB` (also possible that we wrote some of those incorrect tests)

gforsyth:
> You need to set `ibis.set_backend(ddl_con)` before you create the `memtable` (or you can do it with the `monekypatch` line as above).

sorry, definitely use the `monkeypatch` method so that this change doesn't move outside the scope of this test

ibis/backends/flink/__init__.py
```diff
@@ -145,6 +149,34 @@ def list_tables(
...
+        database : str, optional
+            The database to list tables of, if not the current one.
+        catalog : str, optional
+            The catalog to list tables of, if not the current one.
+
+        Returns
+        -------
+        list[str]
+            The list of the view names that match the pattern `like`.
+        """
```
gforsyth, line 176: This is a deviation from the existing API in Ibis.  

Our current convention is to include views in the output of `list_tables`.

ibis/backends/flink/__init__.py
```diff
@@ -145,6 +149,34 @@ def list_tables(
...
+        or the default one if no database/catalog is specified.
+
+        Parameters
+        ----------
+        like : str, optional
+            A pattern in Python's regex format.
+        database : str, optional
+            The database to list tables of, if not the current one.
+        catalog : str, optional
+            The catalog to list tables of, if not the current one.
```
gforsyth, line 170: As my other comment alludes to, we probably want to include the views in the output of `list_tables`, but if this stays in, the docstring should be fixed to refer to `views`

ibis/backends/flink/__init__.py
```diff
@@ -318,27 +350,50 @@ def create_table(
...
             # TODO(chloeh13q): alternatively, we can do CREATE TABLE and then INSERT
             # INTO ... VALUES to keep things consistent
             self._table_env.create_temporary_view(qualified_name, table)
+
         if isinstance(obj, ir.Table):
-            # TODO(chloeh13q): implement CREATE TABLE for expressions
-            raise NotImplementedError
+            # TODO (mehmet): Double check this with the reviewers!
+            if not self.list_tables(name):
+                return ibis.memtable(obj, schema=obj.schema())
```
gforsyth, line 376: There isn't support for creating a `memtable` out of an existing ibis expression, so i don't think you want this here.

ibis/backends/flink/__init__.py
```diff
@@ -318,27 +350,50 @@ def create_table(
...
+            else:
+                return self.create_view(
+                    name=name,
+                    obj=obj,
+                    database=database,
+                    catalog=catalog,
+                    temporary=temp,
+                    overwrite=overwrite,
+                )
```
gforsyth, line 385: This is confusing to me.  If a user defines an `UnboundTable` that they would like to create in the Flink backend, they're going to end up with a `view` instead -- I think a call to `create_table` should create a table.

mfatihaktas: Many thanks @gforsyth for all your comments. They urged me to go back and discuss the behavior of `create_table()` for Flink with @deepyaman (CC: @chloeh13q, @zhenzhongxu). Here is our understanding / conclusion:
* Flink needs a `source connector` to create a (regular) table.
    *  Ref: [Create Table](https://nightlies.apache.org/flink/flink-table-store-docs-master/docs/development/create-table/)
* In-memory data does not have a connector information. This is why, we can only create a `temporary view` in Flink with in-memory data. (`Views` are referred to as `virtual tables` in Flink.)
    * Ref: [Create Tables in the Catalog](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/common/#create-tables-in-the-catalog)
    > Tables can be either virtual (VIEWS) or regular (TABLES). VIEWS can be created from an existing Table object, usually the result of a Table API or SQL query. TABLES describe external data, such as a file, database table, or message queue.

Given these, we have two options in `create_table()` for Flink backend:
1. Require the connector information to be provided. This would imply that we do not support creating table from `in-memory` data (e.g., `pd.Dataframe`, `pa.Table`).
2. Create a view for in-memory data, as currently done.
3. As far as we understand it, `ir.Table` does not have connector information. If we go with option-1, then we won't support creating table from `ir.Table`. If we go with option-2 and `obj` points to an in-memory data (e.g., `obj = ibis.memtable([("a", 1.0)], columns=["a", "b"])`), we create a view. If `obj` is an `UnboundTable`, we do not support it.

Can you please share your opinion and feedback on these conclusions?

gforsyth: We'll probably have to revisit this after this PR and the work in #7640 are consolidated.

I think option 2 is fine for now.  I think there should be support for unbound tables, but that may require some notion of a "default" connector, which may not make sense -- that's still unclear to me.

mfatihaktas: Thanks @gforsyth. I implemented option 2 and consolidated my changes with those in https://github.com/ibis-project/ibis/pull/7640.

ibis/backends/flink/__init__.py
mfatihaktas: Thanks, included `views` in `list_tables()`.

ibis/backends/flink/__init__.py
mfatihaktas: Thanks, included views in `list_tables()`.

zhenzhongxu (commented):
Thanks for putting this together. I've reviewed mostly the part that consolidates with my other PR. Looks mostly good, had a few questions and nit picks.

ibis/backends/flink/__init__.py
```diff
@@ -326,7 +387,7 @@ def create_table(
             dictionary of key-value pairs (key1=val1, key2=val2, ...).
         watermark
             Watermark strategy for the table, only applicable on sources.
-        temp
+        temporary
```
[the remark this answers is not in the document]
mfatihaktas: Thanks, removed.

ibis/backends/flink/__init__.py
```diff
@@ -116,18 +119,65 @@ def drop_database(
     def list_tables(
         self,
         like: str | None = None,
+        temporary: bool = False,
```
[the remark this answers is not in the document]
mfatihaktas: Thanks, removed the argument `temporary`. Now `list_tables()` returns all tables and view, including the temporary ones.

ibis/backends/flink/__init__.py
```diff
@@ -116,18 +119,65 @@ def drop_database(
...
+            temporary=temporary,
+            database=database,
+            catalog=catalog,
+        )
+        view_name_list = self._list_views(
+            like=like,
+            temporary=temporary,
+        )
+
+        return table_name_list + view_name_list
```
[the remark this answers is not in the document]
mfatihaktas: Sorry, could not understand what you meant.
- `_list_tables()` returns only tables.
- `_list_views()` returns only views.
- `list_tables()` returns both tables and views.

So, `_list_tables()` does not return views, hence no duplicate would be returned by `list_tables()`. Please let me know if I am missing something.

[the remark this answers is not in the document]
zhenzhongxu: i was referring according to the documentation, that `_list_tables()` currently returns both tables and views.
specifically, this call `self._table_env._j_tenv.listTables()` is the culprit - see the docstring attached above.

ibis/backends/flink/__init__.py
```diff
@@ -344,27 +405,66 @@ def create_table(
...
+                    dataframe = op.data.to_frame()
+                else:
+                    raise exc.IbisError(
+                        f"Unsupported operation: {type(op)}. "
+                        "If `obj` is of `ir.Table`, the operation must be `InMemoryTable`."
+                    )
+            else:
+                raise exc.IbisError(f"Unsupported `obj` type: {type(obj)}")
+
+            if overwrite and self._list_views(like=name, temporary=True):
```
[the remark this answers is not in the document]
mfatihaktas: The reason to have different implementations for creating `view` in `create_table()` and `create_view()` is two-fold:
1. In `create_table()`, we create a `temporary view` for in-memory `obj` using PyFlink as
```py
            table = self._table_env.from_pandas(
                dataframe, FlinkRowSchema.from_ibis(schema)
            )
            self._table_env.create_temporary_view(qualified_name, table)
```
where `from_pandas()` implicitly creates an "intermediate" table source in Arrow and bases the PyFlink table on it. As PyFlink takes care of setting up the intermediate arrow source for us, it makes sense to use PyFlink API here.

2. In `create_view()`, I added `CreateView` to construct the SQL for creating view. Alternative would be using PyFlink API (as we did above for 1.), but I could not find a way to `CREATE VIEW AS SELECT` using PyFlink. I also think that the "convention" in `create_table()` with the SQL builder class `CreateTableFromConnector` is more flexible in general as we directly use Flink SQL without PyFlink as an intermediary. Also in my experience, Flink SQL is much better documented than PyFlink API.

Does this reasoning make sense to you?

ibis/backends/flink/__init__.py
```diff
@@ -410,20 +512,47 @@ def drop_table(
...
+        self._exec_sql(sql)
+
     def create_view(
         self,
         name: str,
         obj: pd.DataFrame | ir.Table,
         *,
         database: str | None = None,
         catalog: str | None = None,
+        temporary: bool = False,
```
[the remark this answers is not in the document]
mfatihaktas: Grepped `create_view`, but could not find any other backend with `temporary`. What is your opinion? Should we remove it from `create_view()` as well?

ibis/backends/flink/ddl.py
```diff
@@ -36,7 +38,10 @@ def type_to_flink_sql_string(tval):
     if tval.is_timestamp():
         return f"timestamp({tval.scale})"
     else:
-        return type_to_sql_string(tval)
+        sql_string = type_to_sql_string(tval)
+        if not tval.nullable:
+            sql_string += " not null"
```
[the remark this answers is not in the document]
mfatihaktas: Thanks, done.

ibis/backends/tests/test_client.py
```diff
@@ -868,23 +899,30 @@ def test_self_join_memory_table(backend, con):
...
-            ibis.memtable([("a", 1.0)], columns=["a", "b"]),
+            [("a", 1.0)],
+            lambda arg: ibis.memtable(arg, columns=["a", "b"]),
             id="python",
-            marks=pytest.mark.notimpl(["flink"], raises=NotImplementedError),
         ),
         param(
-            ibis.memtable(pd.DataFrame([("a", 1.0)], columns=["a", "b"])),
+            pd.DataFrame([("a", 1.0)]),
+            lambda arg: ibis.memtable(arg, columns=["a", "b"]),
```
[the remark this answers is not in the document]
mfatihaktas: Thanks for catching this, fixed it.

ibis/backends/flink/__init__.py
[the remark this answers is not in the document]
mfatihaktas: To create a memory table view with SQL, we would need to first create an arrow source from `obj` as done by PyFlink. We can surely do that by imitating what is being done by PyFlink, if we decide to do and prioritize it.

gforsyth (changes requested):
Hey @mfatihaktas -- sorry for the delay!  I had a few comments pending and I forgot to submit.

I've flagged a few things that I think need fixing.  Also, because this PR has changed so much (which is fine!), I think updating the PR description with what it is actually doing will be very helpful to our future selves.

ibis/backends/flink/__init__.py
```diff
@@ -400,7 +496,44 @@ def drop_table(
...
+            temporary=True,
+            force=force,
+        )
+        self._drop_table(
+            name=name,
+            database=database,
+            catalog=catalog,
+            temporary=False,
+            force=force,
+        )
```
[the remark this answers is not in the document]
gforsyth: If there are two to make sure the view gets dropped whether it is `temporary` or not, I would suggest updating `_drop_table` so you can pass in an argument to that effect, instead of duplicating the call with a single argument changed.

ibis/backends/flink/__init__.py
```diff
@@ -119,10 +122,10 @@ def list_tables(
         database: str | None = None,
         catalog: str | None = None,
     ) -> list[str]:
-        """Return the list of table names.
+        """Return the list of table/view names.
 
-        Return the list of table names in the specified database and catalog.
-        or the default one if no database/catalog is specified.
+        See the docstring in _list_tables() for the behavior depending on the
+        specification of `database` and `catalog`.
```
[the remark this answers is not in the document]
mfatihaktas: Thanks, modified the docstring as you suggested.

ibis/backends/flink/__init__.py
```diff
@@ -291,7 +351,6 @@ def create_table(
         catalog: str | None = None,
         tbl_properties: dict | None = None,
         watermark: Watermark | None = None,
-        temp: bool = False,
```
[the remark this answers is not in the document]
mfatihaktas: The suggestion to remove `temp` argument was brought up by @zhenzhongxu, which I agreed at the time.
Our rationale was that we can keep the interface simple initially by managing `temp` for the user, and expose it to user later if/when necessary. This train of thought came up after we decided to support creating only `temporary views` for in-memory `obj`. As we already "decide for the user" when `obj` is in-memory, we thought it is perhaps a good idea to eliminate `temp` from the interface.

However, given that `temp` is exposed to user in other backends, it makes sense to keep it for Flink as well. This however requires an additional step in `create_table()` to handle the case when `obj` is `Dataframe` but `temp` is `False`, which is not supported. Currently, I added a "disclaimer" in the docstring of `create_table()` (as well as `create_view()`) to inform the user that we override `temp` as `True` when `obj` is `Dataframe`. Also added a `TODO` there to check with the reviewers if this makes sense or not. Please take a look and let me know what you think.

ibis/backends/flink/__init__.py
```diff
@@ -341,44 +398,84 @@ def create_table(
...
+                # Other cases are unsupported for now, e.g., obj is of UnboundTable.
+                # See TODO right below for more context on how we handle in-memory data.
+                op = obj.op()
+                if isinstance(op, ops.InMemoryTable):
+                    dataframe = op.data.to_frame()
+                else:
+                    raise exc.IbisError(
+                        f"Unsupported operation: {type(op)}. "
+                        "If `obj` is of `ir.Table`, the operation must be `InMemoryTable`."
+                    )
```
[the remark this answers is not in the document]
mfatihaktas: Thanks, please take a look at the updated error message.

ibis/backends/flink/__init__.py
```diff
@@ -387,8 +484,7 @@ def drop_table(
         *,
         database: str | None = None,
         catalog: str | None = None,
-        temp: bool = False,
```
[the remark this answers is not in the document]
mfatihaktas: Addressed as described [above](https://github.com/ibis-project/ibis/pull/7479#discussion_r1424508857).

ibis/backends/flink/__init__.py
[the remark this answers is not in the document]
mfatihaktas: Removed while bringing `temp` back.

ibis/backends/flink/__init__.py
```diff
@@ -473,15 +675,64 @@ def drop_view(
...
+            temporary=True,
+            force=True,
+        )
+        self._drop_view(
+            name=name,
+            database=database,
+            catalog=catalog,
+            temporary=False,
+            force=True,
+        )
```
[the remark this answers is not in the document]
mfatihaktas: Removed while bringing `temp` back.

ibis/backends/flink/__init__.py
```diff
@@ -410,57 +543,126 @@ def drop_table(
...
-        return self.table(name, database=database, catalog=catalog)
+        return self.table(name=name, database=database, catalog=catalog)
 
     def drop_view(
         self,
         name: str,
         *,
         database: str | None = None,
         catalog: str | None = None,
+        temporary: bool = False,
```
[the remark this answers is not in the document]
mfatihaktas: Addressed while bringing `temp` back.

ibis/backends/flink/__init__.py
```diff
@@ -341,44 +365,86 @@ def create_table(
         import pyarrow_hotfix  # noqa: F401
 
         import ibis.expr.types as ir
-        from ibis.backends.flink.datatypes import FlinkRowSchema
 
         if obj is None and schema is None:
-            raise exc.IbisError("The schema or obj parameter is required")
+            raise exc.IbisError("`schema` or `obj` is required")
+        if isinstance(obj, pd.DataFrame) and temp is False:
```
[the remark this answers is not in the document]
mfatihaktas: Thanks, done.

ibis/backends/flink/__init__.py
```diff
@@ -341,44 +365,86 @@ def create_table(
...
+            temp = True
```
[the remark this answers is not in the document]
mfatihaktas: I agree that erroring is cleaner and more explicit than overriding, which is actually what I did at first :)
However with erroring, I encountered the following complication: `temp` is `False` by default for all backends and this makes sense to me.  In many places though, `create_table()` is called with in-memory `obj` and without `temp`, which triggers the error for `isinstance(obj, pd.DataFrame) and temp is False`. I could not find a clean way to modify the tests to make them fit this unusual behavior by Flink backend.

I switched to overriding with the following in mind: As discussed at some point, we might add support for creating tables (instead of creating views as we do now) which would enable supporting `temp=False` for in-memory `obj` as well. This would eliminate the need to override `temp`.

Given this background, do you have any other suggestion? Or should I just switch back to erroring (from overriding) and deal with the test failures by marking them for Flink?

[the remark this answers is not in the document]
gforsyth: I would error and mark the tests as `notimpl`, since those tests are for creating non-temp tables.  Then if we figure out a good pattern for persistent table/view creation from in-memory tables, we'll get a nice signal that it's working.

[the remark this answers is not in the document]
mfatihaktas: Thanks, switched to erroring.

gforsyth (approved):
Really nice work here, @mfatihaktas ! Thanks for bearing with me through the various rounds of changes and rebasing.

PR #7482, opened 2023-11-02
refactor(sql): align logic for filtered reductions

deepyaman:
* There were arbitrary (at least, to my untrained eye) discrepancies in how the `WHERE` clause was getting applied to the reduction arguments. In the base SQL implementation, it was only applied to the first argument, whereas it was applied to all arguments in the Alchemy implementation.
* I've adapted some of the logic from the Alchemy implementation, such as enabling casting values to `Int32` before certain aggregations (e.g. `SUM`). I've been very conservative, and only actually turned it on for the Flink backend.
* In general, I've tried to do some refactoring, but keep things as similar to the original code as possible. I'm open to a different approach, if it makes sense, but wanted to be conservative before aligning.

### DIFF
--- a/ibis/backends/base/sql/compiler/translator.py
+++ b/ibis/backends/base/sql/compiler/translator.py
@@ -187,6 +187,7 @@ class ExprTranslator:
     )
     _dialect_name = "hive"
     _quote_identifiers = None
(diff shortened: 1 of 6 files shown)


cpcloud (approved):
LGTM! Thanks!

PR #7483, opened 2023-11-02
fix(flink): re-implement `ops.ApproxCountDistinct`

deepyaman:
Note that this won't work in streaming mode: `APPROX_COUNT_DISTINCT aggregate function does not support yet for streaming.`

However, it works in batch mode, which is also how the test suite is run. (There are also other things that don't work in streaming mode, such as unbounded ordering, that are already implemented in the backend/passing batch tests, so this isn't any different.)


cpcloud (approved):

cpcloud:
Trino failures are stale, merging.

PR #7486, opened 2023-11-03
refactor: remove `show_sql` in favor of `print(to_sql)`

cpcloud:
Now that we syntax highlight the result of `to_sql` in IPython and Jupyter, we can decrease API surface a bit by removing `show_sql` in favor of `print(to_sql(...))` for folks that are not using IPython or Jupyter.

### DIFF
--- a/docs/how-to/timeseries/sessionize.qmd
+++ b/docs/how-to/timeseries/sessionize.qmd
@@ -74,6 +74,6 @@ sessionized = (
 )
 ```
 
-Calling `ibis.show_sql(sessionized)` displays the SQL query and can be used to confirm that this Ibis table expression does not rely on any join operations.
+Calling `print(ibis.to_sql(sessionized))` displays the SQL query and can be used to confirm that this Ibis table expression does not rely on any join operations.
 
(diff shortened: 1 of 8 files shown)


docs/how-to/timeseries/sessionize.qmd
```diff
@@ -74,6 +74,6 @@ sessionized = (
 )
 ```
 
-Calling `ibis.show_sql(sessionized)` displays the SQL query and can be used to confirm that this Ibis table expression does not rely on any join operations.
+Calling `print(ibis.to_sql(sessionized))` displays the SQL query and can be used to confirm that this Ibis table expression does not rely on any join operations.
```
tswast, line 77: Do we want to call out how to get syntax highlighting to show up in notebooks?

tswast (approved):

docs/how-to/timeseries/sessionize.qmd
cpcloud: Yep! It's just "remove the `print`" 😅

cpcloud: Did this in `docs/index.qmd`.

PR #7495, opened 2023-11-05
feat(datafusion): add isnull/isnan operations

mesejo:
This PR also fixes the offset, which is available but is broken for `count(*)` with parquet files.


cpcloud (approved):

cpcloud:
Excellent work! Thank you.

PR #7500, opened 2023-11-06
feat(datafusion): add StringLength, FindInSet, ArrayStringJoin

mesejo:
This PR also fixes negative indices for `substr`


gforsyth (approved):
Thanks, @mesejo ! This looks great!

PR #7501, opened 2023-11-06
test(flink): set up flink docker container for ci runner

jiupengsun:
We are experiencing slow Flink tests (~2 hours in ci testing). The main reason is PyFlink leverages [MiniCluster](https://github.com/apache/flink/blob/master/flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniCluster.java#L151) to execute Flink job. By default, it will start & stop a MiniCluster for [each](https://github.com/apache/flink/blob/master/flink-clients/src/main/java/org/apache/flink/client/deployment/executors/LocalExecutor.java#L84) job, which greatly slows down the execution speed if we have a lot tests.
There are two options to improve this:
1) Implement a new Executor which reuses MiniCluster for different jobs, and load this Executor in Flink to override the default one. However this approach is troublesome to implement.
2) connect an existed remote Flink cluster and run all tests there. This approach is much simpler and I'm using this approach in this PR. However PyFlink doesn't provide a direct API to connect with remote cluster, I have to directly call Flink Java API via Py4j.

In this PR, I check an environment value 'FLINK_REMOTE_CLUSTER' to decide whether we will run tests on local mini cluster or remote cluster, therefore it won't impact the existed ci pipeline. If we want to enable the Flink remote cluster for testing, there are some prerequisites:

1) Set up a Flink cluster

-  we can set up a local Flink cluster or using a remote one. But we need to include `flink-python-1.18.1.jar` in the Flink runtime path, because ibis depends on PyFlink to execute.
image

- well tuned TaskManager's memory to perform the tests. In my local environment, I set these key params in `flink-conf.yaml`:

```
taskmanager.memory.network.fraction: 0.4
taskmanager.memory.network.min: 512mb
taskmanager.memory.network.max: 2gb
```

2) Set up environment variables in test runner's environment.
```
export FLINK_REMOTE_CLUSTER=true
export FLINK_REMOTE_CLUSTER_ADDR=localhost
export FLINK_REMOTE_CLUSTER_PORT=8081
```

In this PR, I set up a Flink docker container for ci runner, so we can run all Flink tests by connecting to the remote Flink cluster, which is a Flink container running inside the same ci runner machine. This can accelerate the test running time so we don't need to run Flink tests in a separate [execution](https://github.com/ibis-project/ibis/blob/master/.github/workflows/ibis-backends-flink.yml).
Flink [tests](https://github.com/ibis-project/ibis/actions/runs/6777010810/job/18419937893?pr=7501) can be done around 10 mins with this change.

Since Flink tests depend on slightly different settings, so I create an individual workflow job (similar with PySpark) in ibis-backends.yaml.


kszucs:
Does it supercede https://github.com/ibis-project/ibis/pull/7452?

jiupengsun:
> @jiupengsun thanks for working on this! Apparently the actual execution tests are all skipped causing pytest to finish under 30 seconds. My guess is that `apache-flink` package is missing from the build, could you please check it?
Thanks @kszucs ! I also noticed that and working on the fix. And I will rebase this PR[7452](on https://github.com/ibis-project/ibis/pull/7452) once it's merged in.

kszucs:
@jiupengsun I'd prefer merging this PR instead of both of them so that we actually test the code changes in #7452.

ibis/backends/flink/tests/conftest.py
```diff
@@ -19,11 +19,44 @@ def connect(*, tmpdir, worker_id, **kw: Any):
...
+
+        from pyflink.java_gateway import get_gateway
+        from pyflink.table import (
+            EnvironmentSettings,
+            StreamTableEnvironment,
+            TableEnvironment,
+        )
+        from pyflink.table.table_environment import StreamExecutionEnvironment
+
+        if not bool(os.environ.get("FLINK_REMOTE_CLUSTER", False)):
```
cpcloud, line 32: Can we remove this check now?

jiupengsun: I feel this option is still useful that if you don't want to setup a flink docker locally, you can still run the unit tests. Otherwise we have to require developers to maintain a flink cluster in their box.

jiupengsun:
> @jiupengsun I'd prefer merging this PR instead of both of them so that we actually test the code changes in #7452.

sg, I've closed the other PR.

ibis/backends/flink/tests/conftest.py
cpcloud: I think it's okay to require developers to start the container.

We generally don't maintain multiple test deployment methods for any other backend, and I would like to avoid introducing that precedent.

.github/workflows/ibis-backends.yml
```diff
@@ -704,6 +704,70 @@ jobs:
         with:
           flags: backend,${{ matrix.backend.name }},${{ runner.os }},python-${{ steps.install_python.outputs.python-version }}
 
+  test_flink:
```
deepyaman, line 707: Is a separate job necessary? The process seems relatively similar to tests for other backends, wonder if it can't be part of the same matrix.

.github/workflows/ibis-backends.yml
```diff
@@ -704,6 +704,70 @@ jobs:
...
+        run: python -m pip install --upgrade pip 'poetry==1.6.1'
+
+      - name: install ibis
+        run: poetry install --without dev --without docs --extras flink
+
+      # FIXME(deepyaman): We reinstall pandas~=1.5 to get a version with
+      #   ArrowDtype, but this step will be removed once flink relaxes
+      #   its overly-restrictive dependencies (specifying pandas<1.4.0).
+      - name: install ibis-flink deps
+        run: poetry run pip install apache-flink grpcio-status pytest-split pandas~=1.5
```
deepyaman, line 747: This can be simplified now; see https://github.com/ibis-project/ibis/pull/7542

ibis/backends/flink/tests/conftest.py
```diff
@@ -42,6 +75,7 @@ class TestConfForStreaming(TestConf):
     def connect(*, tmpdir, worker_id, **kw: Any):
         """Flink backend is created in streaming mode here. To be used
         in the tests under ibis/ibis/backends/flink/tests/.
+        We only use mini cluster here for simplicity.
```
deepyaman, line 78: Is this true, even if the above code goes the remote cluster route?

docker-compose.yml
```diff
@@ -412,6 +412,38 @@ services:
...
+        jobmanager.rpc.address: flink-jobmanager
+        taskmanager.numberOfTaskSlots: 2
+        taskmanager.memory.process.size: 2048m
+        taskmanager.memory.network.fraction: 0.4
+        taskmanager.memory.network.min: 512mb
+        taskmanager.memory.network.max: 2gb
+    depends_on:
+      - flink-jobmanager
+    command: taskmanager
+    scale: 1
```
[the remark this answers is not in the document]
jiupengsun: I'm not sure if I understand your question. This `flink` entry is used to start a flink taskmanager container, which depends on jobmanager container. And this container is the one that runs all flink unit tests. Which part do you think is unnecessary?

.github/workflows/ibis-backends.yml
jiupengsun: yes it's necessary bc python 11 is not supported yet. So I separate flink tests just like pyspark tests.

.github/workflows/ibis-backends.yml
jiupengsun: lgtm, how about I rebase the code on yours after that PR is merged it?

ibis/backends/flink/tests/conftest.py
jiupengsun: Yes, `ibis/backends/flink/tests` are not running inside remote cluster since it's using `TestConfForStreaming`

docker-compose.yml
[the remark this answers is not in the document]
cpcloud: Sorry! I mean the `scale: 1` bit 😄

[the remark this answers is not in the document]
jiupengsun: oh I thought this specific property means only 1 taskmanager is spun up, which should be necessary?

.github/workflows/ibis-backends.yml
deepyaman: Sounds good, let me ask for a quick review

.github/workflows/ibis-backends.yml
deepyaman: You can just skip 3.11 in the matrix.

docker-compose.yml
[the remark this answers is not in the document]
deepyaman: I looked into it out of curiosity, found a couple things:

1. It seems `scale` is deprecated in favor of `replicas` (under `deploy` key).
2. `replicas` only takes effect if deployed in swarm mode.

So maybe it wouldn't make a difference? I was looking for the default `replicas`, but couldn't find it for now.

ibis/backends/flink/tests/conftest.py
jiupengsun: sure, I'll go that direction

.github/workflows/ibis-backends.yml
jiupengsun: beyond that reason, flink tests also required specific dependency (e.g. pandas 1.5) and environment variables. Therefore I think it's better to isolate this?

docker-compose.yml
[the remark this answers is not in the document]
jiupengsun: Thanks for the due diligence! I'll remove this property in that case.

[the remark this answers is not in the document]
cpcloud: You shouldn't need to deal with scaling or replicas at all, since this compose file is not going to be used for docker swarm or any kind of production deployment.

Plus, we're not using it anywhere else in the file :)

.github/workflows/ibis-backends.yml
gforsyth: #7542 is merged and this can be rebased now!

.github/workflows/ibis-backends.yml
jiupengsun: after rebase your PR, I think it's better to merge the two jobs as you mentioned.

cpcloud (approved):
Net negative lines of code! Love it!

🚢

cpcloud:
@jiupengsun Looks like there's a few failing tests: https://github.com/ibis-project/ibis/actions/runs/6916631103/job/18816909162?pr=7501

If you can fix those up this is good to merge!

jiupengsun:
> Details

hi @cpcloud Do you think this [failed cov test ](https://github.com/ibis-project/ibis/pull/7501/checks?check_run_id=18913294253)is important? or we can skip it?

cpcloud:
Thank you!

PR #7506, opened 2023-11-07
ci(mssql/apt): update before trying to install from apt

gforsyth:
Trying this out to see if it fixes the failure popping up on the mssql builds


gforsyth:
This seems to have fixed the CI issues, so I've gone ahead and added an `apt-get update` to all of our `install` calls (except where they are preceded in the same workflow by a standalone update step)

lostmygithubaccount (approved):
:shipit:

PR #7510, opened 2023-11-08
feat(polars): various Polars compiler enhancements

alexander-beedie:
A few improvements:

* Replace `map_elements` lambdas for Capitalize and Repeat with native expressions.
* Gracefully handles various upcoming deprecations (eg: `clip_min/max` → `clip`).
* Fixes start-only slice ("substr-start-only" test now passes).
* Streamlines usage of literal values (`_assert_literal` → `_literal_value`)
* Use explicit "scale" & "precision" param names with `pl.Decimal`, as we may make a breaking change to switch the order in the near-ish future (apparently we are the odd ones out at the moment ;)

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
 import calendar
-import functools
 import math
 import operator
 from collections.abc import Mapping
-from functools import partial
+from functools import partial, reduce, singledispatch
+from math import isnan
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


alexander-beedie:
Hmm; will take a look at the errors tomorrow... `pytest -m polars` seemed clean here 🤔

alexander-beedie:
Fixed; there's one unrelated Oracle test failing but the Polars ones are all clean ✌️

cpcloud (approved):

PR #7511, opened 2023-11-08
feat(duckdb): allow users to register fsspec filesystem with DuckDB

gforsyth:
DuckDB's python client introduced an integration with `fsspec` where you can use
`fsspec` to handle downloads from any supported `fsspec.AbstractFileSystem` and
feed that into various `DuckDB` `read_*` functions.

Upstream docs are here: https://duckdb.org/docs/guides/python/filesystems

The test is pulling from GCS, but it's also a 59 byte file -- if it ends up
being flaky we can always drop it.

Note for future self and other maintainers:
When we normalize filenames before passing them to DuckDB, we compare against a
known "allowlist" of various cloud blob prefixes (like `az`, `gcs`, `s3`, etc.)

If we _don't_ have the appropriate prefix in there, we mangle the URL quite badly.
For now, we can keep that list up-to-date, but we may want to revisit how we're handling that.

### DIFF
--- a/ibis/backends/duckdb/tests/test_register.py
+++ b/ibis/backends/duckdb/tests/test_register.py
@@ -350,3 +350,20 @@ def test_csv_with_slash_n_null(con, tmp_path):
     t = con.read_csv(data_path, nullstr="\\N")
     col = t.a.execute()
     assert pd.isna(col.iat[-1])
+
+
+@pytest.mark.xfail(
+    LINUX and SANDBOXED,
+    reason=("nix can't hit GCS because it is sandboxed."),
+)
+def test_register_filesystem_gcs(con):
+    import fsspec
+
+    gcs = fsspec.filesystem("gcs")
+
+    con.register_filesystem(gcs)
+    band_members = con.read_csv(
+        "gcs://ibis-examples/data/band_members.csv.gz", table_name="band_members"
+    )
+
+    assert band_members.count().to_pyarrow()
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -38,6 +38,7 @@ if TYPE_CHECKING:
 
     import pandas as pd
     import torch
+    from fsspec import AbstractFileSystem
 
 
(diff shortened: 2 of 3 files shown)


ibis/backends/duckdb/tests/test_register.py
```diff
@@ -350,3 +350,20 @@ def test_csv_with_slash_n_null(con, tmp_path):
...
+    LINUX and SANDBOXED,
+    reason=("nix can't hit GCS because it is sandboxed."),
+)
+def test_register_filesystem_gcs(con):
+    import fsspec
+
+    gcs = fsspec.filesystem("gcs")
+
+    con.register_filesystem(gcs)
+    band_members = con.read_csv(
```
cpcloud, line 365: Is there a way we can make the `con.register_filesystem` call automatic for the names in the allowed list?

gforsyth: Right now we could do that for `gcs`, but for `azure` or `s3` it would be ambiguous whether a user wanted to use fsspec or a native DuckDB extension.
I don't _think_ we want to have it auto register for only some filesystem-like things. 
We might be able to design a consistent set of fallback behaviors?

cpcloud: Nah, it's fine as is. Perhaps a follow up with documentation about what to do in various situations? Ideally we don't want to have different ways to do the very similar things.

cpcloud:
> If we don't have the appropriate prefix in there, we mangle the URL quite badly.

I believe this mangling is happening in DuckDB, right? We're not altering the URL in any way before compiling it into a SQL query IIRC.

gforsyth:
> > If we don't have the appropriate prefix in there, we mangle the URL quite badly.
> 
> I believe this mangling is happening in DuckDB, right? We're not altering the URL in any way before compiling it into a SQL query IIRC.

Nah, this is us trying to get the absolute path to a local file, which I don't think we need to do, so we should probably stop?

cpcloud (approved):

PR #7512, opened 2023-11-08
feat(duckdb): add dict-like for DuckDB settings

gforsyth:
Adds a dict-like passthrough object for setting options in the DuckDB backend.

The repr will show all values as stringified, but this is also how they'll be passed in.

This avoids users having to pass in options via calls to `raw_sql` or similar.

```python
[ins] In [1]: import ibis

[ins] In [2]: penguins = ibis.examples.penguins.fetch()

[ins] In [3]: con = ibis.get_backend(penguins)

[ins] In [4]: con.settings["null_order"] = "nulls_first"

[ins] In [5]: ibis.options.interactive = True

[ins] In [6]: penguins
Out[6]: 
┏━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━┓
┃ species ┃ island    ┃ bill_length_mm ┃ bill_depth_mm ┃ flipper_length_mm ┃ body_mass_g ┃ … ┃
┡━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━┩
│ string  │ string    │ float64        │ float64       │ int64             │ int64       │ … │
├─────────┼───────────┼────────────────┼───────────────┼───────────────────┼─────────────┼───┤
│ Adelie  │ Torgersen │           39.1 │          18.7 │               181 │        3750 │ … │
│ Adelie  │ Torgersen │           39.5 │          17.4 │               186 │        3800 │ … │
│ Adelie  │ Torgersen │           40.3 │          18.0 │               195 │        3250 │ … │
│ Adelie  │ Torgersen │           NULL │          NULL │              NULL │        NULL │ … │
│ Adelie  │ Torgersen │           36.7 │          19.3 │               193 │        3450 │ … │
│ Adelie  │ Torgersen │           39.3 │          20.6 │               190 │        3650 │ … │
│ Adelie  │ Torgersen │           38.9 │          17.8 │               181 │        3625 │ … │
│ Adelie  │ Torgersen │           39.2 │          19.6 │               195 │        4675 │ … │
│ Adelie  │ Torgersen │           34.1 │          18.1 │               193 │        3475 │ … │
│ Adelie  │ Torgersen │           42.0 │          20.2 │               190 │        4250 │ … │
│ …       │ …         │              … │             … │                 … │           … │ … │
└─────────┴───────────┴────────────────┴───────────────┴───────────────────┴─────────────┴───┘

[ins] In [7]: expr = penguins.order_by("bill_depth_mm").select("bill_depth_mm").head(n=2)

[ins] In [8]: expr
Out[8]: 
┏━━━━━━━━━━━━━━━┓
┃ bill_depth_mm ┃
┡━━━━━━━━━━━━━━━┩
│ float64       │
├───────────────┤
│          NULL │
│          NULL │
└───────────────┘

[ins] In [9]: con.settings["null_order"] = "nulls_last"

[ins] In [10]: expr
Out[10]: 
┏━━━━━━━━━━━━━━━┓
┃ bill_depth_mm ┃
┡━━━━━━━━━━━━━━━┩
│ float64       │
├───────────────┤
│          13.1 │
│          13.2 │
└───────────────┘
```

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -7,10 +7,7 @@ import contextlib
 import os
 import warnings
 from pathlib import Path
-from typing import (
-    TYPE_CHECKING,
-    Any,
-)
+from typing import TYPE_CHECKING, Any
 
 import duckdb
 import pyarrow as pa
@@ -69,10 +66,40 @@ _UDF_INPUT_TYPE_MAPPING = {
 }
 
 
+class _Settings:
+    def __init__(self, con):
+        self.con = con
+
+    def __getitem__(self, key):
+        try:
+            with self.con.begin() as con:
+                return con.exec_driver_sql(
+                    f"select value from duckdb_settings() where name = '{key}'"
+                ).one()
+        except sa.exc.NoResultFound:
+            raise KeyError(key)
+
+    def __setitem__(self, key, value):
+        with self.con.begin() as con:
+            con.exec_driver_sql(f"SET {key}='{value}'")
+
+    def __repr__(self):
+        with self.con.begin() as con:
+            kv = con.exec_driver_sql(
+                "select map(array_agg(name), array_agg(value)) from duckdb_settings()"
+            ).scalar()
+
+        return repr(dict(zip(kv["key"], kv["value"])))
+
+
 class Backend(AlchemyCrossSchemaBackend, CanCreateSchema):
     name = "duckdb"
     compiler = DuckDBSQLCompiler
     supports_create_or_replace = True
+
+    @property
+    def settings(self) -> _Settings:
+        return _Settings(self)
 
     @property
     def current_database(self) -> str:
--- a/ibis/backends/duckdb/tests/test_client.py
+++ b/ibis/backends/duckdb/tests/test_client.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import duckdb
+import pyarrow as pa
 import pytest
 import sqlalchemy as sa
 from pytest import param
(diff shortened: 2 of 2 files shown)


ibis/backends/duckdb/__init__.py
```diff
@@ -40,6 +38,38 @@
...
+class _SettingsPassThrough:
+    def __init__(self, con):
+        self.con = con
+
+    def __getitem__(self, key):
+        with self.con.begin() as con:
+            if value := con.exec_driver_sql(
+                f"select value from duckdb_settings() where name = '{key}'"
+            ).cursor.fetchone():
+                return value[0]
```
cpcloud, line 50:
```suggestion
            ).one():
                return value
```

ibis/backends/duckdb/__init__.py
```diff
@@ -40,6 +38,38 @@
     import torch
 
 
+class _SettingsPassThrough:
```
cpcloud, line 41: Can we call this `Settings` (or `_Settings`) and stick it near the base classes?

ibis/backends/duckdb/__init__.py
```diff
@@ -40,6 +38,38 @@
...
+                "select name, value from duckdb_settings()"
+            ).cursor.fetch_arrow_table()
+
+        rows = ["{"]
+        rows.extend(
+            f"'{key}': '{value}',"
+            for key, value in zip(table["name"].to_pylist(), table["value"].to_pylist())
+        )
+        rows.append("}")
+        return "\n".join(rows)
```
cpcloud, line 70: Perhaps we can simplify this a bit:

```suggestion
        with self.con.begin() as con:
            kv = con.exec_driver_sql(
                "select map(array_agg(name), array_agg(value)) from duckdb_settings()"
            ).scalar()

        return repr(dict(zip(kv["key"], kv["value"])))
```

ibis/backends/duckdb/__init__.py
```diff
@@ -74,6 +104,10 @@ class Backend(AlchemyCrossSchemaBackend, CanCreateSchema):
     compiler = DuckDBSQLCompiler
     supports_create_or_replace = True
 
+    @cached_property
```
cpcloud, line 107: Perhaps only `@property` is okay? We're only saving on construction, which shouldn't be expensive. If you still want to avoid it, a `_settings` instance attribute would be simpler than a cached property IMO.

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -132,3 +132,29 @@ def test_create_table_with_timestamp_scales(con, scale):
     schema = ibis.schema(dict(ts=dt.Timestamp(scale=scale)))
     t = con.create_table(gen_name("duckdb_timestamp_scale"), schema=schema, temp=True)
     assert t.schema() == schema
+
+
+def test_config_options(con):
+    def _get_null_value():
```
cpcloud, line 138: Testing this here seems unnecessary since it's testing that a particular DuckDB sql query works independent of ibis whereas the `con.settings` calls are testing an ibis API.

cpcloud (changes requested):
A few implementation comments! Overall design looks great!

ibis/backends/duckdb/tests/test_client.py
gforsyth: I was trying to show that the underlying setting is mutated, but I think I've got a better way to demonstrate that

cpcloud (approved):

cpcloud:
Thanks, this is great!

PR #7523, opened 2023-11-10
refactor(temporal): remove unnecesary `Temporal*` classes

### LINKED ISSUE #7504: docs: `TemporalValue.strftime` method not documented in reference
### Please describe the issue

I see that Ibis has support for a `strftime` method: https://github.com/ibis-project/ibis/blob/ca97dfba1dca428456a8c28b1c2e3b3692ec6547/ibis/expr/types/temporal.py#L26 but I don't see it documented here: https://ibis-project.org/reference/expression-temporal

### Code of Conduct

cpcloud:
This PR removes `Temporal*` classes, which only existed to avoid repeating the definition of `strftime`. With this PR, I repeat the definition and we get accurate docstrings for `strftime` for every type and some needless hierarchy is removed. Since this operatrion has not changed for quite a while, I think the repetition is worth the cost.

Closes #7504.

### DIFF
--- a/ibis/expr/types/temporal.py
+++ b/ibis/expr/types/temporal.py
@@ -21,37 +21,6 @@ if TYPE_CHECKING:
     import ibis.expr.types as ir
 
 
-@public
-class TemporalValue(Value):
-    def strftime(self, format_str: str) -> ir.StringValue:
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
:shipit:

Issue #7502, opened 2023-11-07
bug: DuckDB `IntervalValue.days` is broken


judahrand:
### What happened?

I would expect `IntervalValue.days` to return an integer which is the number of days in the interval.

### What version of ibis are you using?

`ca97dfba1dca428456a8c28b1c2e3b3692ec6547`

### What backend(s) are you using, if any?

DuckDB

### Relevant log output

```sh
import ibis
from ibis import _

conn = ibis.connect('duckdb://:memory:')

t = conn.sql("""
    SELECT
        CAST('2022-11-06 12:30:12' AS TIMESTAMP) AS ts1,
        CAST('2023-10-04 16:20:42' AS TIMESTAMP) AS ts2
    """,
).mutate(
    days=(_.ts2 - _.ts1).days,
).execute()


```
Traceback (most recent call last):
  File "/Users/<redacted>/.venv/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1910, in _execute_context
    self.dialect.do_execute(
  File "/Users/<redacted>/.venv/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
  File "/Users/<redacted>/.venv/lib/python3.11/site-packages/duckdb_engine/__init__.py", line 144, in execute
    self.__c.execute(statement, parameters)
duckdb.BinderException: Binder Error: No function matches the given name and argument types '*(DOUBLE, INTERVAL)'. You might need to add explicit type casts.
        Candidate functions:
        *(TINYINT, TINYINT) -> TINYINT
        *(SMALLINT, SMALLINT) -> SMALLINT
        *(INTEGER, INTEGER) -> INTEGER
        *(BIGINT, BIGINT) -> BIGINT
        *(HUGEINT, HUGEINT) -> HUGEINT
        *(FLOAT, FLOAT) -> FLOAT
        *(DOUBLE, DOUBLE) -> DOUBLE
        *(DECIMAL, DECIMAL) -> DECIMAL
        *(UTINYINT, UTINYINT) -> UTINYINT
        *(USMALLINT, USMALLINT) -> USMALLINT
        *(UINTEGER, UINTEGER) -> UINTEGER
        *(UBIGINT, UBIGINT) -> UBIGINT
        *(INTERVAL, BIGINT) -> INTERVAL
        *(BIGINT, INTERVAL) -> INTERVAL

[...]

PR #7526, opened 2023-11-10
docs(temporal): add `Literal` annotation to display possible units for `delta` method

cpcloud:
While tesing https://github.com/ibis-project/ibis/issues/7502 I saw that we don't have possible units listed for temporal delta operations. This PR fixes that.


gforsyth (approved):
:shipit:

PR #7532, opened 2023-11-11
ci: write benchmark data to parquet

cpcloud:
Converts JSON benchmark data to parquet before uploading to GCS.


gforsyth (approved):
:shipit:

Issue #7516, opened 2023-11-09
bug: parse_sql removes where clauses


AndrewRook:
### What happened?

Running the example from #4853 strips the where clause from the query. It's hard for me to say exactly why that's happening or whether it happens in all cases, but I did a couple of tweaks to that example and they all had this problem. Here is my test code:

```python
import ibis

sql = """
SELECT *, first_name as first FROM employee 
LEFT JOIN call ON employee.id = call.employee_id 
WHERE id < 5 
ORDER BY id DESC
"""

catalog = {
    "employee": {"first_name": "string", "last_name": "string", "id": "int64"},
    "call": {
        "start_time": "timestamp",
        "end_time": "timestamp",
        "employee_id": "int64",
        "call_outcome_id": "int64",
    },
    "call_outcome": {"outcome_text": "string", "id": "int64"},
}

expr = ibis.parse_sql(sql, catalog)

print(ibis.decompile(expr, format=True))
```
The output is missing the `.filter(...)` command, as shown in the log below. 

### What version of ibis are you using?

7.0.0

### Relevant log output

```sh
import ibis


employee = ibis.table(
    name="employee",
    schema={"first_name": "string", "last_name": "string", "id": "int64"},
)
call = ibis.table(
    name="call",
    schema={
        "start_time": "timestamp",
        "end_time": "timestamp",
        "employee_id": "int64",
        "call_outcome_id": "int64",
    },
)
leftjoin = employee.left_join(call, employee.id == call.employee_id)

result = leftjoin.select(
    [
        leftjoin.first_name,
        leftjoin.last_name,
        leftjoin.id,
        call.start_time,
        call.end_time,
        call.employee_id,
        call.call_outcome_id,
        leftjoin.first_name.name("first"),
    ]
).order_by(leftjoin.id.desc())
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

AndrewRook:
> Note that this is an experimental feature which we haven't been focusing so far, although the implementation is pretty generic making quick iterations possible.

Yep totally understand if you don't plan on prioritizing support for it. It is pretty cool though so I've been playing around with it. If I shouldn't file additional bug reports just let me know and I will refrain 😄.

cpcloud:
@AndrewRook Please keep filing bug reports! We're definitely interested to see how you're using it.

PR #7538, opened 2023-11-13
fix(ir): `ibis.parse_sql()` removes where clause

kszucs:
Resolves https://github.com/ibis-project/ibis/issues/7516

SQLGlot has a `"condition"` field in the `Join` expression object which hasn't been mapped to an ibis expression previously. It also provides a `projections` field which is omitted for now, we may need to consider another level of projection for that too.

### DIFF
--- a/ibis/expr/tests/snapshots/test_sql/test_parse_sql_join_with_filter/decompiled.py
+++ b/ibis/expr/tests/snapshots/test_sql/test_parse_sql_join_with_filter/decompiled.py
@@ -1,0 +1,32 @@
+import ibis
+
+
+call = ibis.table(
+    name="call",
+    schema={
+        "start_time": "timestamp",
+        "end_time": "timestamp",
+        "employee_id": "int64",
+        "call_outcome_id": "int64",
+        "call_attempts": "int64",
+    },
+)
+employee = ibis.table(
+    name="employee",
+    schema={"first_name": "string", "last_name": "string", "id": "int64"},
+)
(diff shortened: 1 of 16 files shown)


ibis/expr/tests/snapshots/test_sql/test_parse_sql_join_with_filter/decompiled.py
```diff
@@ -0,0 +1,32 @@
...
+        "employee_id": "int64",
+        "call_outcome_id": "int64",
+        "call_attempts": "int64",
+    },
+)
+employee = ibis.table(
+    name="employee",
+    schema={"first_name": "string", "last_name": "string", "id": "int64"},
+)
+proj = employee.left_join(call, employee.id == call.employee_id).filter(employee.id < 5)
```
kszucs, line 18: Here is the filter properly placed.

cpcloud (approved):
Can you apply this patch?

```patch
diff --git a/.gitattributes b/.gitattributes
index 8341a4b76..364bd5c28 100644
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,4 @@ requirements.txt linguist-generated=true
 docs/backends/app/requirements.txt linguist-generated=false
 docs/_freeze/**/html.json linguist-generated=true
 docs/**/*.excalidraw linguist-generated=true
+**/decompiled.py linguist-generated=true
```

cpcloud:
LGTM! I will add the generated file bits in a follow-up.

PR #7527, opened 2023-11-10
feat(bigquery): add support for temporary tables

cpcloud:
This PR adds support for temporary tables to BigQuery using [sessions](https://cloud.google.com/bigquery/docs/sessions-intro).

The first time any query is run, a session is created. This session has an id that
is then used for every subsequent query.

This session has an associated temporary dataset that is used when creating and
accessing temporary tables.

~One (un)fortunate side-effect is that I had to make some breaking changes to
`table` to ensure that cross-project and cross-dataset queries work. The
`database` argument to table now refers to a GCP project, and a new `schema`
argument refers to a BigQuery `dataset`.~ Changes are no longer breaking.

A few other nice features work with BigQuery now that temporary tables work, like reading example files as well as the `.cache()` method.

PR #7545, opened 2023-11-14
perf(bigquery): use more efficient representation for memtables

cpcloud:
This PR moves the BigQuery memtable implementation to be much more efficient than the current one by using temporary tables along with the `bq.Client.load_table_from_dataframe` method.

Depends on #7527.

### DIFF
--- a/ibis/backends/tests/test_array.py
+++ b/ibis/backends/tests/test_array.py
@@ -566,7 +566,18 @@ def test_array_map(backend, con, input, output):
 @pytest.mark.parametrize(
     ("input", "output"),
     [
-        param({"a": [[1, None, 2], [4]]}, {"a": [[2], [4]]}, id="nulls"),
+        param(
+            {"a": [[1, None, 2], [4]]},
+            {"a": [[2], [4]]},
+            id="nulls",
+            marks=[
+                pytest.mark.notyet(
+                    ["bigquery"],
+                    raises=BadRequest,
+                    reason="NULLs are not allowed as array elements",
+                )
+            ],
+        ),
         param({"a": [[1, 2], [4]]}, {"a": [[2], [4]]}, id="no_nulls"),
     ],
 )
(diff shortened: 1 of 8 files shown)


cpcloud:
BQ tests are passing:

```
cloud in 🌐 falcon in …/ibis on  bigquery-better-memtables is 📦 v7.0.0 via 🐍 v3.10.13 via ❄️  impure (ibis-3.10.13-env)
❯ pytest -m bigquery --snapshot-update -n auto --dist loadgroup -q
bringing up nodes...^[[B^[[B^[[B
ssssssssssssssssssssss.....s.............x......xx........x............x.s.......x.x......x....s........x.x.............x.x............x....x.....x.x...x..xxx.x...........xx.x..x.x...x.....x [ 12%]
...x.......x....x.x.x..................x..x....x...x......x.......xx...x.x.........x.....x.x...x.......xxxx.x....................x.x.......x....x.........xx...xx...x...x...x.....xx.........x [ 24%]
......x.xx......x....x..xxx..x...........x.............x.xx..x.........x......xx......xx.........xxx.x..x.x.......x....x.......x...x.xx......x..x..xx.xx.xxx....xx.xxxx...x.xxx..x.xx.x.x..x.x [ 36%]
....x..x.x...x...........s....xx..............x.x..x......x...x..x.....x...x....x..............xx.x....xx.....x...........x...............................................x................... [ 48%]
......................x....................x.x..........................................................x.x..........x.x........x......xx....x...xx.xxxxxxxxxxxxxxxxxxxx.xx..xx.x...xxxx....x. [ 60%]
....x....x..xx.x.xx....x.xx.x.x.....x.x.xx.x.................xx....x................x...x....x..x.x...xx.xx....x....................x.xx.x.x.x.x................x.x.x..x..xx.x..x............. [ 72%]
..............................................................................x................................................................................................................ [ 85%]
...................................................x.....x........x.....x.xx.x..x.x.xxx..x..................................x..xx...........x...................x............................. [ 97%]
..x...........s...........................                                                                                                                                                     [100%]
1283 passed, 27 skipped, 253 xfailed in 266.88s (0:04:26)
```

gforsyth (approved):
:ship:

ibis/backends/tests/test_array.py
```diff
@@ -711,10 +722,28 @@ def test_array_union(con):
...
+    "data",
+    [
+        param(
+            {"a": [[3, 2], [], []], "b": [[1, 3], [None], [5]], "c": range(3)},
+            id="nulls",
+            marks=[
+                pytest.mark.notyet(
+                    ["bigquery"],
+                    raises=BadRequest,
+                    reason="BigQuery doesn't support arrays with null elements",
```
cpcloud, line 735: Previously these tests passed because of BigQuery's quirky UNNEST behavior

1. We were creating data with `UNNEST(<ARRAY<...>>[...])`
2. The `ArrayIntersect` operation is implemented using `SELECT ... UNNEST(left) INTERSECT DISTINCT SELECT UNNEST(right)`

The difference comes from this quirk:

```sql
SELECT [1, NULL]; -- NOT allowed
SELECT * FROM UNNEST([1, NULL]); -- allowed
```

tswast:
When you say more efficient, do you have data you can share regarding the threshold of when this implementation becomes preferable? I'm aware that there's a 10 MB query size limit in BQ, so certainly sizes smaller than that. Is the thought that folks hit that size pretty quick, so why not just load the data always?

ibis/backends/tests/test_array.py
gforsyth: No, I don't think this warrants a breaking change commit.

cpcloud:
> When you say more efficient, do you have data you can share regarding the threshold of when this implementation becomes preferable? I'm aware that there's a 10 MB query size limit in BQ, so certainly sizes smaller than that. Is the thought that folks hit that size pretty quick, so why not just load the data always?

I didn't do any benchmarking, but yeah it's pretty easy to hit the limit. There's also the issue of the latency of generating a 10MB long string from iterating over the rows of a dataframe.

PR #7546, opened 2023-11-14
feat(bigquery): support reading parquet, json and csv files

cpcloud:
This PR adds support for reading Parquet, JSON and CSV files directly into BigQuery. Local paths, local globs, GCS paths, and GCS globs are all supported. Depends on #7545.

Loading local globs with and without `concurrent.futures`:

### Setup

```
cloud in 🌐 falcon in ~ via 🐍
❯ ls ~/data/trip-data
Permissions Size User  Date Modified Name
.rw-r--r--  151M cloud 26 Jun 11:34  yellow_tripdata_2016-01.parquet
.rw-r--r--  158M cloud 26 Jun 11:34  yellow_tripdata_2016-02.parquet
```

### WITH `concurrent.futures`

```

In [2]: con = ibis.bigquery.connect()

In [3]: ibis.set_backend(con)

In [4]: %time t = ibis.read_parquet("/home/cloud/data/trip-data/*.parquet")
CPU times: user 254 ms, sys: 191 ms, total: 445 ms
Wall time: 51 s
```

### WITHOUT `concurrent.futures`

```

In [2]: con = ibis.bigquery.connect()

In [3]: ibis.set_backend(con)

In [4]: %time t = ibis.read_parquet("/home/cloud/data/trip-data/*.parquet")
CPU times: user 213 ms, sys: 216 ms, total: 429 ms
Wall time: 1min 24s
```

So there's a nice latency decrease by running the load job for each file concurrently (this is using threads).

### DIFF
--- a/ibis/backends/bigquery/__init__.py
+++ b/ibis/backends/bigquery/__init__.py
@@ -2,7 +2,10 @@
 
 from __future__ import annotations
 
+import concurrent.futures
 import contextlib
+import glob
+import os
 import re
 import warnings
 from functools import partial
@@ -40,6 +43,7 @@ with contextlib.suppress(ImportError):
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Mapping
+    from pathlib import Path
 
     import pyarrow as pa
     from google.cloud.bigquery.table import RowIterator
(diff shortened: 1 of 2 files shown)


ibis/backends/bigquery/__init__.py
```diff
@@ -102,6 +123,162 @@ def __init__(self, *args, **kwargs) -> None:
             name, schema=self._session_dataset, database=self.billing_project
         ).op()
 
+    def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
+        self._make_session()
+
+        raw_name = op.name
+
+        project = self.billing_project
+        dataset = self._session_dataset
```
tswast, line 132: Any chance we could use the anonymous dataset, instead? Note, the anonymous dataset is also location-dependent, so we'll need to make sure we get that info when starting a BQ backend.

See: https://github.com/googleapis/python-bigquery-dataframes/pull/194 where BigQuery DataFrames moves away from BQ Session due to a 1 query at a time limitation.

cpcloud: This is the anonymous dataset if I understand correctly. I am grabbing the destination from the query that creates a session (in `_make_session()`).

cpcloud: Or are you suggesting that we create an anonymous dataset manually, and use that?

tswast: > This is the anonymous dataset if I understand correctly. I am grabbing the destination from the query that creates a session (in _make_session()).

Technically yes, the BQ Session implementation writes all of it's data to the anonymous dataset.

> Or are you suggesting that we create an anonymous dataset manually, and use that?

Kinda. I'm suggesting we don't use BQ Sessions since it's not actually necessary and has the 1 query at a time limitation, which has been a problem for us in BQ DataFrames. Instead, we can grab the anonymous dataset directly. See: https://github.com/googleapis/python-bigquery-dataframes/pull/158/files#diff-ca8b88da6ee41672b3d377eb32f9b63ab34a24ec18815c144d4667319ceff6f4

Any resources created there have an automatic time-to-live of 24 hours, but that can be adjusted on individual objects. (We are using 7 days for resources we create in BQ DataFrames for now, and I'm considering adding an ability to manually clean up or refresh the expiration).

cpcloud: Oh, I think I see now: I simply would avoid passing the session id, but use the anonymous dataset from the now-one-off session created from `SELECT 1`. I can give that a shot.

cpcloud: I put up https://github.com/ibis-project/ibis/pull/7548 to remove session use.

ibis/backends/bigquery/__init__.py
```diff
@@ -147,6 +151,136 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
+    ) -> ir.Table:
+        self._make_session()
+
+        if table_name is None:
+            table_name = util.gen_name(f"bq_read_{job_config.source_format}")
+
+        dataset = self._session_dataset.dataset_id
+        project = self._session_dataset.project
+
+        self.drop_table(table_name, schema=dataset, database=project, force=True)
```
tswast, line 169: FYI: This step isn't required if we set `job_config.write_disposition = "WRITE_TRUNCATE"`, but I see we need append-mode below for multiple files. Might be worth a comment explaining that.

tswast (approved):
Thanks!

ibis/backends/bigquery/__init__.py
cpcloud: Thanks for the review! Indeed, I took the `WRITE_TRUNCATE` approach first before tinkering with loading multiple files concurrently.

I'll add a comment.

cpcloud: Actually, I think since it's possible to cause data loss this way, I'm going to check whether the table exists first and error out.

cpcloud: Reconsidering my last statement, since this is all going in the anonymous dataset overwriting things is less risky, since it's hidden from most public-facing APIs/applications.

cpcloud:
BQ tests passing:

```
…/ibis on  bigquery-read-files is 📦 v7.0.0 via 🐍 v3.10.13 via ❄️  impure (ibis-3.10.13-env) took 3m49s
❯ pytest -m bigquery --snapshot-update -q -n auto --dist loadgroup
bringing up nodes...
ssssssssssssssssssssss.x.....x......x..x.x....x............xxx.............xx......xx..xx.xx....x.x.x........x.........xx..x....................x........................xx...x...x.......................... [ 13%]
............................................x..xx..........................x.x.....x..x..xx.x..xx.x..x...........x......x..x.....x......................x...xx.x.x.x.......x..x.x.x.x.........xx............. [ 26%]
x......x.......x...................x.x...x........x......x.x.x..xxxxxx.xx.xxxxxxxxxxxxxxx.x...xxxxxx.xxxx..xxxxxx........x..xx.x...x.x..x.x..x...xxx.x.......s.......xx.....x.........x.............s...x.... [ 39%]
.......................................................s......x................................................x.............................x...............xx...x.......x...x....x.........x.x.x..x........ [ 52%]
...x...x...x.....xx.xx.x......x.x....x...............x...x.....x...x........xx.xx.x....x.xxxx..x.x.x...x.x.x...x.s.....x.x...x...........x..x......x..xx..x...xxx.xx..xx.x.x..x...x.x.x........xxx.x......... [ 65%]
.....xxx.x....xx.............x.x....xx...x.x.x.xxxx.....x.....x...xxxx.x........x...xx...x.x.x............x.x...x....xx......................x...................................xx..xxx...x..........x..x... [ 78%]
............................................................................................................................................................................................................. [ 91%]
...........................................x......................................................x........s..........x.........                                                                              [100%]
1290 passed, 27 skipped, 246 xfailed in 369.33s (0:06:09)
```

PR #7547, opened 2023-11-14
fix(pandas): drop `__index_level_N__` cols before applying schema

gforsyth:
There's a pandas drop-in replacement from the Rapids team: https://rapids.ai/cudf-pandas/

This mostly works with Ibis, but our recent change to rendering the repr with `pyarrow` is failing because something in the cudf version adds an additional `index` column.
We don't know to expect this column, and asking `convert_table` to apply a schema with 5 columns to a table with 6 columns causes problems.

This is a hack, but maybe that's enough for now.

### DIFF
--- a/ibis/backends/pandas/__init__.py
+++ b/ibis/backends/pandas/__init__.py
@@ -298,6 +298,13 @@ class BasePandasBackend(BaseBackend):
         output = pa.Table.from_pandas(
             self.execute(table_expr, params=params, limit=limit, **kwargs)
         )
+
+        # cudf.pandas adds a column with the name `__index_level_0__` (and maybe
+        # other index level columns) but these aren't part of the known schema
+        # so we drop them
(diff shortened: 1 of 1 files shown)


ibis/backends/pandas/__init__.py
```diff
@@ -298,6 +298,10 @@ def to_pyarrow(
         output = pa.Table.from_pandas(
             self.execute(table_expr, params=params, limit=limit, **kwargs)
         )
+
+        output = output.drop_columns(
+            filter(lambda col: col.startswith("__index_level_"), output.column_names)
```
cpcloud, line 303: Can you add a short comment about the trailing `_` in `__index_level_`? I almost reviewed it as a mistake :)

cpcloud (approved):

cpcloud:
Thank you!

PR #7554, opened 2023-11-15
fix: ensure that executing an outer join with non-nullable columns works

### LINKED ISSUE #7487: bug: inconsistencies in results when exporting `to_pandas` and `to_pyarrow` when there are nulls in a not-nullable column
### What happened?

Hello,

I am experiencing some strange behavior when performing a join between two tables:

Depending on the method (right or inner, vs outer or left), I am getting an empty output when calling `to_pyarrow` on the result of the latter two methods, whereas I am getting the expected result when calling `to_pandas`. On the other hand, the first two methods give the same result in arrow or pandas.

One of the tables I'm joining is a slightly complex expression with self joins and aggregations.

```python
import ibis.deferred as ibd

conn.table("TABLE_2", schema="SCHEMA")
    .join(
        table_1_filtered, # result of various self-joins
        [
            ibd["date"] == table_1_filtered["date"],
            ibd["client"] == table_1_filtered["client"]
        ],
        how="outer"
    )
```

I get the same issue when calling `.outer_join` or `left_join`

Latest 8.x.x branch

mssql

### Relevant log output

I can provide additional info on tables if needed, though will need to anonymize the info to post it, so lmk if you think it would be relevant.

### Code of Conduct

cpcloud:
This PR fixes an issue where we were being overly broad about catching `ValueError`s coming from pyarrow. Additionally, join projection columns must always be made nullable, due to the possible presence of nulls in joins. Fixes #7487.


ibis/backends/tests/test_join.py
```diff
@@ -326,3 +326,23 @@ def test_join_with_trivial_predicate(awards_players, predicate, how, pandas_valu
...
+        param("outer", 2, id="outer"),
+    ],
+)
+def test_outer_join_nullability(con, how, nrows):
+    left = con.tables.win.select(x=lambda t: t.x.type().copy(nullable=False)).filter(
+        lambda t: t.x.isin((1, 2))
+    )
+    right = left.filter(lambda t: t.x == 1)
+    expr = left.join(right, "x", how=how)
+    assert all(typ.nullable for typ in expr.schema().types)
```
inigohidalgo, line 345: Should `"x"`'s nullability stay as is?

ibis/backends/base/__init__.py
```diff
@@ -327,23 +327,18 @@ def to_pyarrow(
...
-                table_expr, params=params, limit=limit, **kwargs
-            ) as reader:
-                table = (
-                    pa.Table.from_batches(reader)
-                    .rename_columns(table_expr.columns)
-                    .cast(arrow_schema)
-                )
-        except pa.lib.ArrowInvalid:
-            raise
-        except ValueError:
```
gforsyth, line 343: :headdesk:

ibis/backends/tests/test_join.py
gforsyth: No, I don't think so.  There shouldn't be any non-nullable outputs coming out of a join

gforsyth (approved):
Pretty sure that `ValueError` catch was mine.  Oof.

ibis/backends/tests/test_join.py
cpcloud: No, because it's possible for it to be null in the join projection **even if the input column is not null**.

Whether a column from a join project is null depends on both the join type, the predicate **and** the data, the latter of which we can't know statically.

```
D create or replace table t (x BIGINT NOT NULL);
D create or replace table s (x BIGINT NOT NULL);
D insert into t values (1), (2);
D insert into s values (1), (2), (3);
D select t.x as x, s.x as sx from t full join s on t.x=s.x;
┌───────┬───────┐
│   x   │  sx   │
│ int64 │ int64 │
├───────┼───────┤
│     1 │     1 │
│     2 │     2 │
│       │     3 │
└───────┴───────┘
```

cpcloud: > if you're joining on "x", and that x isn't nullable, could it become nullable? I'm not familiar with how SQL handles nulls in a join column.

It can. See my example 😅

inigohidalgo: Missed that as I was typing the comment hehe, thanks for the clarification :)

cpcloud: The nullability for all of our SQL engines except one (ClickHouse) is a constraint on input to an existing table: you can't insert any NULLs into a column with a `NOT NULL` constraint.

The nullability of a function call or expression in a `SELECT` statement is not constrained in any way. It might be used in the engine for optimization but that's way below ibis's operating level.

ibis/expr/operations/relations.py
```diff
@@ -311,7 +325,11 @@ class AnyInnerJoin(Join):
 
 @public
 class AnyLeftJoin(Join):
-    pass
+    @property
+    def schema(self) -> Schema:
+        return Schema(
+            {name: typ.copy(nullable=True) for name, typ in super().schema.items()}
+        )
```
cpcloud, line 332: Turns that all of these are necessary to address the case of non-overlapping columns 😅

PR #7558, opened 2023-11-16
test(flink): revise WIP marks in aggregation tests

deepyaman:
# Development notes

* `test_aggregation.py`
  * **notyet** Flink `collect` returns a multiset, which can't be converted (`Python vectorized UDF doesn't support logical type MULTISET<STRING> NOT NULL currently.`). However, [casting of multisets if currently limited](https://stackoverflow.com/a/69641983) (e.g. casting to an array fails).
  * **never** Flink doesn't support quantiles yet; even if and when it does, it would be [`APPROX_PERCENTILE`](https://issues.apache.org/jira/browse/FLINK-28736?jql=text%20~%20%22APPROX_PERCENTILE%22).
    * Same logic translates to `median`/`approx_median` support.
  * **notyet** Flink doesn't support corr/cov computation.
  * **notyet** `GROUP_CONCAT` could be implemented as `collect` followed by Flink `ARRAY_JOIN`, but `collect` itself isn't supported (see above).
  * **notimpl** Backend doesn't implement UDF support yet.

### DIFF
--- a/ibis/backends/flink/registry.py
+++ b/ibis/backends/flink/registry.py
@@ -448,12 +448,14 @@ operation_registry.update(
 )
 
 _invalid_operations = {
-    # ibis.expr.operations.strings
-    ops.Translate,
-    ops.FindInSet,
     # ibis.expr.operations.numeric
     ops.IsNan,
     ops.IsInf,
+    # ibis.expr.operations.reductions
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
:shipit:

Issue #7568, opened 2023-11-16
feat: disable always-failing CI runs in forks


NickCrews:
### Is your feature request related to a problem?

In my fork, some CI workflows always fail because only the upstream repo has the correct permissions, etc.

For example, it doesn't make sense for my fork to be uploading to GCS on a corn job: https://github.com/NickCrews/ibis/actions/runs/6885281429/workflow

nor this workflow:
https://github.com/NickCrews/ibis/actions/runs/6366788904/workflow

Can we add some `if` to this workflow so it only runs on the upstream repo?

I think this should work:

```
jobs:
  publish:
    if: github.repository_owner == 'ibis-project'
```

Thanks!

### Describe the solution you'd like

See above

### What version of ibis are you running?

main

### What backend(s) are you using, if any?

NA

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
Thanks for the issue! Happy to accept a PR!

PR #7569, opened 2023-11-16
chore(ci): skip bad CI runs on forks

NickCrews:
These CI workflows don't make sense to run on forks. As a fork owner,
I get emails every time these flows fail.
Let's save some junk emails and some
wasted electricity.

Double check me that I picked out the right workflows to skip, and that I didn't miss other ones.

fixes https://github.com/ibis-project/ibis/issues/7568

### DIFF
--- a/.github/workflows/conda-lock.yml
+++ b/.github/workflows/conda-lock.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   condalock:
+    if: github.repository_owner == 'ibis-project'
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


.github/workflows/conda-lock.yml
```diff
@@ -49,7 +50,7 @@ jobs:
           path: ci/conda-lock/*/${{ matrix.python-version }}.lock
 
   condalock_pr:
-    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
+    if: (github.repository_owner == 'ibis-project') && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')
```
NickCrews, line 53: double check me here, I think we won't know if this fails until the cron fires.

cpcloud: That's true, but there's nothing we can do about it AFAIK.

cpcloud (approved):
LGTM, thanks!

cpcloud:
I think you got all the cron-based workflows, which are the problematic ones.

PR #7570, opened 2023-11-16
feat(exporting): allow passing writer_opts to PyArrow ParquetWriter and CSVWriter

jayceslesar:
Small tweak to allow passing of writer specific args/kwargs. Also fixes a typo

### DIFF
--- a/ibis/backends/base/__init__.py
+++ b/ibis/backends/base/__init__.py
@@ -547,7 +547,7 @@ class _FileIOHandler:
         import pyarrow.parquet as pq
 
         with expr.to_pyarrow_batches(params=params) as batch_reader:
-            with pq.ParquetWriter(path, batch_reader.schema) as writer:
+            with pq.ParquetWriter(path, batch_reader.schema, **kwargs) as writer:
                 for batch in batch_reader:
                     writer.write_batch(batch)
 
@@ -582,7 +582,7 @@ class _FileIOHandler:
         import pyarrow.csv as pcsv
 
         with expr.to_pyarrow_batches(params=params) as batch_reader:
-            with pcsv.CSVWriter(path, batch_reader.schema) as writer:
+            with pcsv.CSVWriter(path, batch_reader.schema, **kwargs) as writer:
                 for batch in batch_reader:
                     writer.write_batch(batch)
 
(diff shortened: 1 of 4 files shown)


ibis/backends/base/__init__.py
```diff
@@ -546,8 +546,9 @@ def to_parquet(
         self._import_pyarrow()
         import pyarrow.parquet as pq
 
+        writer_opts = kwargs.get("writer_opts", {})
         with expr.to_pyarrow_batches(params=params) as batch_reader:
-            with pq.ParquetWriter(path, batch_reader.schema) as writer:
+            with pq.ParquetWriter(path, batch_reader.schema, **writer_opts) as writer:
```
cpcloud, line 551: How about instead passing `**kwargs`?

ibis/backends/base/__init__.py
```diff
@@ -581,8 +582,9 @@ def to_csv(
         self._import_pyarrow()
         import pyarrow.csv as pcsv
 
+        writer_opts = kwargs.get("writer_opts", {})
         with expr.to_pyarrow_batches(params=params) as batch_reader:
-            with pcsv.CSVWriter(path, batch_reader.schema) as writer:
+            with pcsv.CSVWriter(path, batch_reader.schema, **writer_opts) as writer:
```
cpcloud, line 587: Same as the Parquet version: How about instead passing `**kwargs`?

cpcloud (changes requested):
Thanks for PR!

Can you add a test, one for each of parquet and csv code paths? Putting them in `ibis/backends/tests/test_export.py` is fine.

ibis/backends/base/__init__.py
jayceslesar, line 551:
```suggestion
            with pq.ParquetWriter(path, batch_reader.schema, **kwargs) as writer:
```

ibis/backends/base/__init__.py
jayceslesar, line 587:
```suggestion
            with pcsv.CSVWriter(path, batch_reader.schema, **kwargs) as writer:
```

jayceslesar:
Working on adding tests but have had to exclude duckdb as its export options are not the same as pyarrow https://duckdb.org/docs/sql/statements/copy#csv-options

jayceslesar:
Heads up this is failing because the duckdb backend doesnt use the writers it just uses the duckdb copy kwargs, so I think we should exclude duckdb from this test unless we wanted it to dump normally, and then iterate the output in batches and stream to the parquet writer 🤷

cpcloud:
I opted for excluding duckdb from the tests you added.

Thank you for address my feedback!

cpcloud (approved):

PR #7574, opened 2023-11-17
feat(flink): implement nested schema support

chloeh13q:
Implement support for nested schema (`ROW`).

### DIFF
--- a/ibis/backends/flink/registry.py
+++ b/ibis/backends/flink/registry.py
@@ -4,7 +4,13 @@ from typing import TYPE_CHECKING
 
 import ibis.common.exceptions as com
 import ibis.expr.operations as ops
-from ibis.backends.base.sql.registry import aggregate, fixed_arity, helpers, unary
+from ibis.backends.base.sql.registry import (
+    aggregate,
+    fixed_arity,
+    helpers,
+    quote_identifier,
+    unary,
+)
 from ibis.backends.base.sql.registry import (
     operation_registry as base_operation_registry,
 )
@@ -15,6 +21,12 @@ if TYPE_CHECKING:
     from ibis.backends.base.sql.compiler import ExprTranslator
 
 operation_registry = base_operation_registry.copy()
+
+
+def type_to_sql_string(tval):
+    if tval.is_array():
+        return f"array<{helpers.type_to_sql_string(tval.value_type)}>"
+    return helpers.type_to_sql_string(tval)
 
 
 def _not(translator: ExprTranslator, op: ops.Node) -> str:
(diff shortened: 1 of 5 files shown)


ibis/backends/flink/registry.py
```diff
@@ -454,6 +467,8 @@ def _timestamp_from_ymdhms(
     # ibis.expr.operations.numeric
     ops.IsNan,
     ops.IsInf,
+    # ibis.expr.operations.reductions
+    ops.ArrayCollect,
```
deepyaman, line 471: Why do you need to add this? It doesn't exist in the base registry to begin with, so it has no real effect.

ibis/backends/flink/registry.py
```diff
@@ -61,10 +67,11 @@ def _cast(translator: ExprTranslator, op: ops.generic.Cast) -> str:
...
         return arg_translated
-
-    from ibis.backends.base.sql.registry.main import cast
-
-    return cast(translator=translator, op=op)
+    elif op.arg.dtype.is_temporal() and op.to.is_int64():
+        return f"1000000 * unix_timestamp({arg_translated})"
+    else:
+        sql_type = type_to_sql_string(op.to)
+        return f"CAST({arg_translated} AS {sql_type})"
```
deepyaman, line 74: What are these changes for? I'd need to double check (too many things change, so I could be misremembering), but I feel like I removed the very specific timestamp-to-int64 cast previously, because it was breaking some tests.

ibis/backends/flink/registry.py
```diff
@@ -382,6 +389,11 @@ def _timestamp_from_ymdhms(
     return f"CAST({concat_string} AS TIMESTAMP)"
 
 
+def _struct_field(translator, op):
+    arg = translator.translate(op.arg)
+    return f"{arg}.`{op.field}`"
```
deepyaman, line 394: Should this use `quote_identifier`?

ibis/backends/flink/registry.py
chloeh13q: This is exactly what `cast(translator=translator, op=op)` does. I just copied [the code from the function](https://github.com/ibis-project/ibis/blob/master/ibis/backends/base/sql/registry/main.py#L103C1-L110C54) instead of importing and using `cast`, because I need to override the implementation for `type_to_sql_string(op.to)`.

ibis/backends/flink/registry.py
chloeh13q: I took this code from BigQuery, but sure I can change it to `quote_identifier`.

ibis/backends/flink/registry.py
```diff
@@ -17,6 +23,12 @@
 operation_registry = base_operation_registry.copy()
 
 
+def type_to_sql_string(tval):
+    if tval.is_array():
+        return f"array<{helpers.type_to_sql_string(tval.value_type)}>"
+    return helpers.type_to_sql_string(tval)
```
cpcloud, line 29: Eventually we should use the SchemaMapper and TypeMapper classes for this kind of thing, but not a blocker for this PR.

cpcloud (approved):
LGTM!

@deepyaman Can you resolve your comment if it's addressed? I'll merge after that!

deepyaman:
> LGTM!
> 
> @deepyaman Can you resolve your comment if it's addressed? I'll merge after that!

I can't seem to resolve it, but I'm good!

chloeh13q:
@cpcloud Just rebased this PR. Should be ready for merge!

PR #7576, opened 2023-11-17
feat(api): add `ibis.range` function for generating sequences

### LINKED ISSUE #7169: feat: `ibis.range` (or similar) API for generating sequences

Not related to a problem.

Many of our backends support generating a range of numbers as an array. We should support that ibis.

Here are the mappings and a few details:

- BigQuery: [`GENERATE_ARRAY`](https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#generate_array)
- ClickHouse: [`numbers`](https://clickhouse.com/docs/en/sql-reference/table-functions/numbers)
- DataFusion: ?
- Druid: ?
- DuckDB: [`range`](https://duckdb.org/docs/sql/functions/nested.html#range-functions)
- Flink: ?
- Impala: ?
- Polars: ?
- Postgres: [`generate_series`](https://www.postgresql.org/docs/current/functions-srf.html) (returns rows, but can be wrapped in `array(select generate_series(...))`
- PySpark: [`sequence`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.sequence.html)
- Snowflake: [`ARRAY_GENERATE_RANGE`](https://docs.snowflake.com/en/sql-reference/functions/array_generate_range)
- Trino: `sequence`, list on https://trino.io/docs/current/functions/array.html#array-functions

`master`

N/A

### Code of Conduct

cpcloud:
Adds an `ibis.range` API for generating sequences. This cannot yet be used with `ibis.read_parquet` et al. That will be done in a follow-up.

Closes #7169.

Backend support:

- [x] BigQuery
- [x] ClickHouse
- [x] DuckDB
- [x] Polars
- [x] Postgres
- [x] PySpark
- [x] Snowflake
- [x] Trino

Follow-ups:

- date range
- timestamp range
- decimal range
- floating range

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import builtins
 import datetime
 import functools
 import numbers
@@ -143,6 +144,7 @@ __all__ = (
     "parse_sql",
     "pi",
     "random",
+    "range",
     "range_window",
     "read_csv",
     "read_delta",
@@ -471,7 +473,7 @@ def _memtable_from_dataframe(
         newcols = getattr(
             schema,
             "names",
-            (f"col{i:d}" for i in range(len(cols))),
+            (f"col{i:d}" for i in builtins.range(len(cols))),
         )
         df = df.rename(columns=dict(zip(cols, newcols)))
     op = ops.InMemoryTable(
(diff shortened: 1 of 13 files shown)


cpcloud:
cc @alexander-beedie There are some panics for certain somewhat pathological inputs to `pl.int_ranges`:

```python
pl.int_ranges(0, -7, eager=True)  # python returns an empty range here
pl.int_ranges(7, -7, eager=True)  # python returns an empty range here
pl.int_ranges(7, 0, eager=True)  # python returns an empty range here
pl.int_ranges(10, 0, -1, eager=True)  # any valid negative step seems to overflow
```

alexander-beedie:
> cc @alexander-beedie There are some panics for certain somewhat pathological inputs to `pl.int_ranges`:

[Fixed](https://github.com/pola-rs/polars/pull/12548); thanks for the heads-up  ;)

ibis/expr/api.py
```diff
@@ -1947,6 +1949,102 @@ def watermark(time_col: str, allowed_delay: ir.IntervalScalar) -> Watermark:
...
+    │       1 │
+    │       2 │
+    │       3 │
+    │       4 │
+    └─────────┘
+    """
+    raise NotImplementedError()
+
+
+@range.register(int)
```
kszucs, line 2033: Shouldn't this return with an array literal instead?

ibis/expr/operations/arrays.py
```diff
@@ -186,3 +186,18 @@ def dtype(self):
                 }
             )
         )
+
+
+class Range(Value):
```
kszucs, line 191: I assume this base class is for the follow-up range implementations.

cpcloud: Yep!

ibis/expr/operations/relations.py
```diff
@@ -495,7 +494,7 @@ def _projection(self):
 @public
 class DummyTable(Relation):
     # TODO(kszucs): verify that it has at least one element: Length(at_least=1)
-    values: VarTuple[Value[dt.Any, ds.Scalar]]
+    values: VarTuple[Value[dt.Any]]
```
kszucs, line 497: Why is this change required?

BTW the original could be spelled as `VarTuple[Scalar]`

cpcloud: Otherwise it's not possible to write `ibis.range(n).unnest()`. I need to be able to create a table from an unnested array without existing base relations, and `unnest(x)` is `ds.Columnar`.

ibis/expr/api.py
cpcloud: Why? It's perfectly fine to run `ibis.range(start, stop)` where `start` and `stop` are arbitrary integer expressions.

kszucs: Ohh, this is just overloading the first argument. I was thinking of a simplification where all the arguments passed to `ibis.range()` are integers where the result is equivalent with `ibis.literal(range(...))`.

ibis/expr/operations/relations.py
kszucs: Why not converting to table *before* calling unnest?

```py
In [9]: ibis.range(5).as_table()[0].unnest()
Out[9]:
r0 := DummyTable
  IntegerRange(0, 5, 1) array<int8>

IntegerRange(0, 5, 1): Unnest(r0.IntegerRange(0, 5, 1))
```

cpcloud: That UX is not really acceptable though 😂

cpcloud: It's perfectly valid to have column expressions that have no base relation, though it's not particularly common.

ibis/expr/api.py
cpcloud: Sure, but I will not be bending over backwards to make the type annotations more specific, especially since they are correct as is :)

kszucs (approved):

PR #7577, opened 2023-11-17
chore(deps): specify scipy dependency to allow locking

cpcloud:
This PR addresses an issue with the 1.10.x series of scipy, where we pick up scipy 1.6.1 due to wider python bounds than are supported by scipy. See https://github.com/scipy/scipy/issues/19513 for details, and an extensive discussion in https://github.com/numpy/numpy/issues/24810.


cpcloud:
SciPy is only used transitively, and only in our documentation dependencies, so this should have zero effect on end users as no required or backend-optional dependencies have changed.

cpcloud:
Merging this to alleviate relocking other dependencies!

PR #7579, opened 2023-11-17
docs(sql): emphasize the need to close a `raw_sql` cursor only when using `SELECT` statements

cpcloud:
Based on https://github.com/ibis-project/ibis/discussions/7578, add some related documentation.

### DIFF
--- a/ibis/backends/base/sql/__init__.py
+++ b/ibis/backends/base/sql/__init__.py
@@ -136,12 +136,16 @@ class BaseSQLBackend(BaseBackend):
         ::: {.callout-tip}
         ## Consider using [`.sql`](#ibis.backends.base.sql.BaseSQLBackend.sql) instead
 
-        If your query is a SELECT statement, you should use the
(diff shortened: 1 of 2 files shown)


ibis/backends/base/sql/__init__.py
```diff
@@ -136,12 +136,16 @@ def raw_sql(self, query: str):
...
 
-        If your query is a SELECT statement, you should use the
+        If your query is a `SELECT` statement you can use the
         [backend `.sql`](#ibis.backends.base.sql.BaseSQLBackend.sql) method to avoid
-        having to release the cursor returned from this method manually.
+        having to manually release the cursor returned from this method.
 
-        ::: {.callout-warning collapse="true"}
-        ## The returned cursor object must be **manually released** if you use `raw_sql`.
+        ::: {.callout-warning}
```
cpcloud, line 143: I removed the collapse so that the SELECT/DDL distinction is more visible.

gforsyth (approved):
docs!  this looks good.

PR #7591, opened 2023-11-18
Snowflake: Add Support for "read_csv" with https

IndexSeek:
Currently using the Snowflake backend with Ibis, you'll get an error if you specify a URL as a file path for read_csv. This works with other engines where the behavior is natively supported, but Snowflake requires those additional preprocessing steps. 

I added a portion to use a TempFile and load the content to the temporary stage. Complex file formats might cause issues with this, but I'll try to do some additional testing. 

We could likely add this to the other `read_...` methods, but I wanted to start with this for now. 

It's possible to make this work with S3, Azure Blob, GCS, etc., but that could get complex determining whether or not to use client storage options or integrations available natively in Snowflake. 

Here's some code you can use to test the new feature:

```python
from ibis.interactive import *
from pathlib import Path
from snowflake.ml.utils.connection_params import SnowflakeLoginOptions

con = ibis.snowflake.connect(**SnowflakeLoginOptions())

con.read_csv("https://storage.googleapis.com/ibis-tutorial-data/wowah_data/locations.csv")
```

### DIFF
--- a/ibis/backends/snowflake/__init__.py
+++ b/ibis/backends/snowflake/__init__.py
@@ -17,6 +17,7 @@ import textwrap
 import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
+from urllib.request import urlretrieve
 
 import pyarrow as pa
 import pyarrow_hotfix  # noqa: F401
@@ -777,10 +778,17 @@ $$"""
             )
             con.exec_driver_sql(create_infer_fmt)
 
-            # copy the local file to the stage
-            con.exec_driver_sql(
-                f"PUT 'file://{Path(path).absolute()}' @{stage} PARALLEL = {threads:d}"
-            )
+            if path.startswith("https://"):
+                with tempfile.NamedTemporaryFile() as tmp:
+                    urlretrieve(path, filename=tmp.name)
+                    tmp.flush()
+                    con.exec_driver_sql(
+                        f"PUT 'file://{tmp.name}' @{stage} PARALLEL = {threads:d} AUTO_COMPRESS = TRUE"
+                    )
+            else:
+                con.exec_driver_sql(
+                    f"PUT 'file://{Path(path).absolute()}' @{stage} PARALLEL = {threads:d} AUTO_COMPRESS = TRUE"
+                )
 
             # handle setting up the schema in python because snowflake is
             # broken for csv globs: it cannot parse the result of the following
--- a/ibis/backends/snowflake/tests/test_client.py
+++ b/ibis/backends/snowflake/tests/test_client.py
@@ -197,6 +197,22 @@ def test_read_csv_options(con, tmp_path):
     assert t.schema() == ibis.schema(dict(a="int64", b="int64"))
 
 
+def test_read_csv_https(con):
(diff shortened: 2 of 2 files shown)


cpcloud (commented):
LGTM, if a bit circuitous (internet to local to internet) :)

Is there any chance that this will ever be natively supported by Snowflake itelf?

Happy to help add a test and deal with the optional `requests` dependency.

ibis/backends/snowflake/__init__.py
```diff
@@ -752,10 +752,20 @@ def read_csv(
...
-            )
+            if path.startswith("https://"):
+                import requests
+
+                with tempfile.NamedTemporaryFile() as tmp:
+                    with requests.get(path, stream=True) as r:
+                        shutil.copyfileobj(r.raw, tmp)
+                    tmp.flush()
+                    con.exec_driver_sql(
+                        f"PUT 'file://{tmp.name}' @{stage} PARALLEL = {threads:d}"
```
cpcloud, line 763: Should we set `AUTO_COMPRESS = TRUE` to cut down on bandwidth?

IndexSeek: I think we can do that! It should speed things up a little bit.

IndexSeek:
So Snowflake has this in PrPr: https://docs.snowflake.com/en/sql-reference/sql/create-external-access-integration, but it seems a little excessive and often requires higher permission.I agree, it's strange going from Internet, Local, back to Internet, I wished it could be streamlined. 

I was thinking that we could simplify the code by using the Snowflake connector's `put_stream` function, but I wasn't sure if that would require a larger change.

Could we use urllib rather than requests to avoid that dep?

cpcloud:
> Could we use urllib rather than requests to avoid that dep?

I think so, and it looks like it'll be a bit cleaner:

```python
from url.request import urlretrieve

tmp_file, _ = urlretrieve("https://storage.googleapis.com/ibis-tutorial-data/wowah_data/locations.csv")

# PUT file://{tmp_file}
```

sfc-gh-twhite:
Would adding this functionality to the other Snowflake `read...` methods also be worthwhile? Perhaps breaking out this piece for reusability:

```python
if path.startswith("https://"):
    with tempfile.NamedTemporaryFile() as tmp:
        urlretrieve(path, filename=tmp.name)
        tmp.flush()
        con.exec_driver_sql(
            f"PUT 'file://{tmp.name}' @{stage} PARALLEL = {threads:d} AUTO_COMPRESS = TRUE"
        )
else:
    con.exec_driver_sql(
        f"PUT 'file://{Path(path).absolute()}' @{stage} PARALLEL = {threads:d} AUTO_COMPRESS = TRUE"
    )
```

I guess there's also the possibility of "http" being an accepted prefix.

cpcloud:
@sfc-gh-twhite Yeah, I think so!

Let's do that in a follow up though!

cpcloud (approved):
LGTM!

🚢

cpcloud:
The included test passes:   

```
ibis/backends/snowflake/tests/test_client.py::test_read_csv_https PASSED [100%]
```

PR #7593, opened 2023-11-19
feat: add `ArrayValue.flatten` method that removes one level of nesting from array values

### LINKED ISSUE #7575: feat(api): array flatten operation

This came up during @gforsyth's [recent blog post](https://ibis-project.org/posts/querying-pypi-metadata-compiled-languages/). Ideally the `flatten` UDF wouldn't be necessary.

Something like

```python
t.array_expr.flatten()
```

`master`

DuckDB in the blog post, but many (perhaps even all) of our array-supporting backends have this operation.

### Code of Conduct

cpcloud:
This PR implements `ArrayFlatten` which can be used to remove a single level of
nesting from a nested array value.

Closes #7575.

#### Backend support

- [ ] ~BigQuery~ (BigQuery doesn't support arrays of arrays)
- [x] ClickHouse
- [ ] ~DataFusion~ (Doesn't implement a flatten function)
- [x] DuckDB
- [x] Polars
- [ ] ~Postgres~ (Postgres doesn't preserve dimension information in user-facing way)
- [x] PySpark
- [x] Snowflake
- [x] Trino

### DIFF
--- a/ibis/expr/operations/arrays.py
+++ b/ibis/expr/operations/arrays.py
@@ -186,3 +186,19 @@ class ArrayZip(Value):
                 }
             )
         )
+
+
+@public
+class ArrayFlatten(Value):
+    """Flatten a nested array one level.
+
+    The input expression must have at least one level of nesting for flattening
+    to make sense.
+    """
+
+    arg: Value[dt.Array[dt.Array]]
+    shape = rlz.shape_like("arg")
+
+    @property
+    def dtype(self):
+        return self.arg.dtype.value_type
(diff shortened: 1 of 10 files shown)


ibis/expr/operations/arrays.py
```diff
@@ -186,3 +187,14 @@ def dtype(self):
                 }
             )
         )
+
+
+@public
+class ArrayFlatten(Value):
+    # the input must have at least one level of nested for flatten to make sense
+    arg: Annotated[Value[dt.Array], Attrs(dtype=Attrs(value_type=dt.Array))]
```
kszucs, line 195: Following should work actually:

```suggestion
    arg:Value[dt.Array[dt.Array]]
```

cpcloud: Nice, much better!

kszucs (approved):
Really nice addition!

PR #7597, opened 2023-11-20
refactor(bigquery): move `BigQueryType` to use sqlglot for type parsing and generation

### LINKED ISSUE #7531: feat: Add from_string method on BigQueryType to support integration with dbt-ibis

I've created https://github.com/binste/dbt-ibis so that dbt models can be written with Ibis. The integration with various Ibis backends is rather straightforward and the package already supports 8 backends. The only thing I usually need is to be able to convert a column data type as it appears in the database, for example `NUMERIC(38, 10)`, to an `ibis.expr.datatypes.core.DataType` (relevant code in `dbt-ibis` is [here](https://github.com/binste/dbt-ibis/blob/main/dbt_ibis/_dialects.py#L47) for reference). 

With the refactoring in Ibis 7, for most backends this got very easy as I can just use the dialect specific type mappers in `ibis.backends.base.sqlglot.datatypes` such as `DuckDBType` as they all have a `from_string` classmethod.

The BigQuery backend in Ibis does not use SQLGlot and the `ibis.backends.bigquery.datatypes.BigQueryType` class does not have a `from_string` method defined. Would it be possible to add this?

7.0.0 and I also checked master

_No response_

### Code of Conduct

cpcloud:
This PR moves the BigQuery type mapper to use sqlglot instead of the string munging we are currently using. Closes #7531.

### DIFF
--- a/ibis/backends/bigquery/datatypes.py
+++ b/ibis/backends/bigquery/datatypes.py
@@ -1,101 +1,124 @@
 from __future__ import annotations
 
 import google.cloud.bigquery as bq
-import sqlglot as sg
+import sqlglot.expressions as sge
 
 import ibis
 import ibis.expr.datatypes as dt
 import ibis.expr.schema as sch
-from ibis.formats import SchemaMapper, TypeMapper
-
-_from_bigquery_types = {
-    "INT64": dt.Int64,
-    "INTEGER": dt.Int64,
-    "FLOAT": dt.Float64,
-    "FLOAT64": dt.Float64,
-    "BOOL": dt.Boolean,
-    "BOOLEAN": dt.Boolean,
-    "STRING": dt.String,
-    "DATE": dt.Date,
-    "TIME": dt.Time,
-    "BYTES": dt.Binary,
-    "JSON": dt.JSON,
-}
+from ibis.backends.base.sqlglot.datatypes import SqlglotType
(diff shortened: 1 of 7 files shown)


kszucs:
It looks like a refactor rather than a new feature. Can you update the commit message?

ibis/backends/bigquery/datatypes.py
```diff
@@ -1,101 +1,124 @@
...
+    def _from_sqlglot_GEOGRAPHY(cls) -> dt.Decimal:
+        return dt.GeoSpatial(
+            geotype="geography", srid=4326, nullable=cls.default_nullable
+        )
+
+    @classmethod
+    def _from_sqlglot_TINYINT(cls) -> dt.Int64:
+        return dt.Int64(nullable=cls.default_nullable)
+
+    _from_sqlglot_UINT = (
```
kszucs, line 49: This is hard to read, I assume ruff formats it this way. 
Perhaps we should have a way to update the type mapping in a more convenient way.

kszucs (approved):
Besides the commit message it LGTM.

cpcloud:
It's both a feature (the addition of `from_string`, achieved by subclassing `SqlglotType`) and a refactor (the majority of the changes here).

Given the latter, I think it does make sense to use `refactor` here.

PR #7605, opened 2023-11-22
fix(flink): quote SQL identifiers indiscriminately

deepyaman:
Needed to update the list for Flink (ran into a couple of issues today, with `TIME` and `VALUE`).


pyproject.toml
```diff
@@ -376,7 +376,7 @@ show_deps = true
 [tool.codespell]
 # local codespell matches `./docs`, pre-commit codespell matches `docs`
 skip = "*.lock,.direnv,.git,./docs/_freeze,docs/_freeze/**,*.svg,*.css,*.html,*.js"
-ignore-regex = '\b(i[if]f|I[IF]F|AFE)\b'
+ignore-regex = '\b(i[if]f|I[IF]F|AFE|inout)\b'
```
deepyaman, line 379: Would be great if inline ignores were supported, but [they're not](https://github.com/codespell-project/codespell/issues/1212).

ibis/backends/tests/test_window.py
```diff
@@ -1215,7 +1215,7 @@ def test_first_last(backend):
     raises=com.OperationNotDefinedError,
     reason="not support by the backend",
 )
-@pytest.mark.notyet(["flink"], raises=Py4JJavaError, reason="not supported by Flink")
+@pytest.mark.broken(["flink"], raises=Py4JJavaError, reason="bug in Flink")
```
deepyaman, line 1218: There were two issues here:

1. `time` is a reserved keyword
2. Even then, there seems to be an issue: https://apache-flink.slack.com/archives/C03G7LJTS2G/p1700675938430139

The stack trace on the second says, `Caused by: org.apache.flink.util.FlinkRuntimeException: org.apache.flink.api.common.InvalidProgramException: Table program cannot be compiled. This is a bug. Please file an issue.`, but I'm still discussing with somebody on the Flink side; may update later with a Jira.

ibis/backends/flink/identifiers.py
```diff
@@ -0,0 +1,560 @@
+from __future__ import annotations
+
+# https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/dev/table/sql/overview/#reserved-keywords
```
deepyaman, line 3: I couldn't find anywhere in the code to pull this from programmatically. 🤷

ibis/backends/base/sql/registry/helpers.py
```diff
@@ -14,9 +14,11 @@ def format_call(translator, func, *args):
     return "{}({})".format(func, ", ".join(formatted_args))
 
 
-def quote_identifier(name, quotechar="`", force=False):
+def quote_identifier(
+    name, quotechar="`", force=False, base_identifiers=identifiers.base_identifiers
```
deepyaman, line 18: I was going to name it `reserved_keywords`, but I _think_ Flink's set is actually reserved + unreserved keywords. So I left it as `base_identifiers`.

ibis/backends/flink/identifiers.py
cpcloud: Instead of maintaining this list, why don't we quote every indentifier?

deepyaman: > Instead of maintaining this list, why don't we quote every identifier?

Done, as discussed.

cpcloud (approved):

cpcloud:
No need to squash!

PR #7607, opened 2023-11-23
docs(pypi-metadata-post): add Fortran pattern and fix regex

cpcloud:
Fixes an issue in the pypi blog post where Fortran was not being processed due to an incorrect RE2 regex (`{0-2}` vs `{0,2}`, the latter is the correct one).


cpcloud:
The plot isn't showing up due to an error, not sure why that is passing the docs build.

docs/posts/querying-pypi-metadata-compiled-languages/index.qmd
```diff
@@ -202,7 +206,7 @@ Now that the data are tidied, we can pass our expression directly to Altair and
 import altair as alt
 
 chart = (
-    alt.Chart(collapse_names)
+    alt.Chart(collapse_names.to_pandas())
```
cpcloud, line 209: There's an issue with pyarrow timezone discovery in the nix environment: the tz library that arrow-cpp uses hardcodes the location of time zone information to an FHS path and doesn't support an environment variable or any other external means of setting it.

This issue won't show up in CI because this document's execution is frozen.

gforsyth (approved):
:shipit:

PR #7610, opened 2023-11-24
fix(decompile): handle isin

AndrewRook:
Hi! 👋

In playing around with `parse_sql` I discovered that it can't handle `IN` clauses yet. I'm working on a fix for that but in doing so I realized that `decompile` _also_ can't seem to handle `isin`, which is a prerequisite for testing `parse_sql` based on how those unit tests are constructed. Without the patch, the unit test I added here errors out with 

```python
E         File "<string>", line 6
E           result = countries.filter(countries.continent.in_values((<ibis.expr.decompile.CallStatement object at 0x139e5db10>, <ibis.expr.decompile.CallStatement object at 0x139e5d330>)))
E                                                                    ^
E       SyntaxError: invalid syntax
```
I  **think** this is the right place to put in the fix. If it's not let me know and I can try again!


AndrewRook:
I'm not sure why the flink tests are failing, it doesn't seem to be related to my change but I can't really tell — running `just ci-check -m flink` locally passes, but judging by the Actions logs the ci is running more tests than I'm doing locally.

cpcloud:
Flink tests are unrelated here, but should be fixed in `master` now!

cpcloud (approved):

cpcloud:
@AndrewRook Many thanks for the PR!

PR #7623, opened 2023-11-27
ci(flink): don't raise compressed class space size

deepyaman:
Increasing compressed class space size is only necessary if running the test suite using mini-cluster, but the CI here is using a remote, Dockerized cluster.

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -353,12 +353,12 @@ jobs:
           IBIS_TEST_WEBHDFS_USER: hdfs
           IBIS_EXAMPLES_DATA: ${{ runner.temp }}/examples-${{ matrix.backend.name }}-${{ matrix.os }}-${{ steps.install_python.outputs.python-version }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ibis-backends.yml
```diff
@@ -353,20 +353,19 @@ jobs:
           IBIS_TEST_WEBHDFS_USER: hdfs
           IBIS_EXAMPLES_DATA: ${{ runner.temp }}/examples-${{ matrix.backend.name }}-${{ matrix.os }}-${{ steps.install_python.outputs.python-version }}
 
+      # FIXME(deepyaman): If some backend-specific test, in test_ddl.py,
+      #   executes before common tests, they will fail with:
+      #   org.apache.flink.table.api.ValidationException: Table `default_catalog`.`default_database`.`functional_alltypes` was not found.
+      #   Therefore, we quarantine backend-specific tests to avoid this.
```
deepyaman, line 359:
```suggestion
      #   Therefore, we run backend-specific tests second to avoid this.
```

PR #7624, opened 2023-11-27
docs(blog-post): add blog post comparing ibis to pandas and dask

cpcloud:
This PR adds a blog post that compares the performance of Ibis using the DuckDB
backend with Pandas and Dask.

The goal is to show that Ibis + DuckDB can be a faster, more memory-efficient
and more convenient alternative to Pandas and Dask.

I know performance can be contentious so I want to make sure folks from the
Pandas and Dask communities agree with the comparisons here.

I really want the comparison here to reflect accurate use of Pandas and Dask.

To that end I spent a bunch of time talking to @ncclementi and as well as
reading docs and tinkering with things myself before putting up this PR.

CC @mrocklin and @phofl for any Dask-related feedback or recommendations. If
there are ways to improve the Dask code here, I'm happy to do so.


docs/posts/pydata-performance/dask_impl.py
```diff
@@ -0,0 +1,40 @@
...
+import dask.config
+import dask.dataframe as pd
+from dask.distributed import Client
+
+if __name__ == "__main__":
+    dask.config.set({"dataframe.convert-string": True})
+
+    with Client(silence_logs=logging.ERROR):
+        df = pd.read_parquet(
+            "/data/pypi-parquet/*.parquet",
```
cpcloud, line 14: I'll make these paths a bit more friendly so that people run the workload from `$PWD`.

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+I started with this code:
+
+```{python}
+#| error: true
+import pandas as pd
+
+df = pd.read_parquet("/data/pypi-parquet/*.parquet")
+```
+
+Looks like pandas doesn't support globs. That's fine, we can use the builtin
```
gforsyth, line 205:
```suggestion
Looks like Pandas doesn't support globs. That's fine, we can use the builtin
```

docs/posts/pydata-performance/dask_impl.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import annotations
+
+import logging
+
+import dask.config
+import dask.dataframe as pd
```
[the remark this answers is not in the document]
cpcloud: Fair enough. Mostly did this to cut down on the diff but that's kind of swallowed by using the context manager.

phofl: The first thing I would recommend is to use dask_expr, which contains a re-implementation of dask.dataframe that has a query planning layer

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+potential issues with memory and give pandas the best possible shot.
+
+The [smallest file](#data-size) is 632 MiB on disk.
+
+```{python}
+import os
+
+smallest_file = min(glob.glob("/data/pypi-parquet/*.parquet"), key=os.path.getsize)
+
+%time df = pd.read_parquet(smallest_file)
```
phofl, line 228: please use ``dtype_backend="pyarrow"`` or set ``pd.options.infer_string = True``, the latter will be the default in pandas 3.0

Converting from arrow strings to objects takes up approx 80% of your runtime if you have string columns (see https://medium.com/towards-data-science/utilizing-pyarrow-to-improve-pandas-and-dask-workflows-2891d3d96d2b in dtype_backend in PyArrow-native I/O readers)

This will improve all your string operations by a huge factor as well

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+df = df.groupby(
+    [
+        df.uploaded_on.dt.floor("M"),
+        df.path.str.extract(r"\.([a-z0-9]+)$", 0),
+    ]
+)
+df
+```
+
+Here we hit the first API issue going back to an [old pandas
```
gforsyth, line 294:
```suggestion
Here we hit the first API issue going back to an [old Pandas
```

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+    .assign(
+        ext=lambda t: t.ext.str.replace(r"cxx|cpp|cc|c|hpp|h", "C/C++", regex=True)
+        .str.replace("^f.*$", "Fortran", regex=True)
+        .str.replace("rs", "Rust")
+        .str.replace("go", "Go")
+        .str.replace("asm", "Assembly")
+        .replace("", None)
+    )
+    .groupby(["month", "ext"])
+    .agg({"projects": lambda s: len(set(sum(s, [])))})
```
phofl, line 352: What is this line doing?

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+1. Later we'll see that Dask doesn't support lambda functions in `assign` and
+   instead we'll create an intermediate `DataFrame` for the `assign` call.
+
+Sort the values, add a new column and do the final aggregation:
+
+```{python}
+%%time
+df = (
+    df.reset_index()  # <1>
+    .sort_values("month", ascending=False)
```
phofl, line 342: Why would you do this before grouping? GroupBy sorts anyway

docs/posts/pydata-performance/index.qmd
phofl: Why are you using lambdas before you switch to more efficient things?

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+
+If I only have pandas at my disposal, I'm unsure of how I can avoid getting
+a bigger computer to run this query over the entire data set.
+
+### Rewriting the query to be fair
+
+At this point I wondered whether this was a fair query to compare Dask and
+Pandas against.
+
+After all, Pandas' (and therefore Dask's) use of object arrays and their
```
phofl, line 387: That's not correct anymore, Dask will use arrow backed strings by default for string columns since 2023.7.1

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+Let's try a tool that nominally scales to our problem: [Dask](https://dask.org).
+
+## Dask
+
+One really nice component of Dask is
+[`dask.dataframe`](https://docs.dask.org/en/stable/dataframe.html).
+
+Dask DataFrame implements a [good chunk of the pandas
+API](https://docs.dask.org/en/stable/dataframe.html#scope) and can be a drop-in
+replacement for pandas.
```
gforsyth, line 566:
```suggestion
replacement for Pandas.
```

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+from `import pandas as pd` to `import dask.dataframe as pd`.
+
+This worked and the workload completed. However, after talking to Dask
+expert and Ibis contributor [Naty Clementi](https://github.com/ncclementi) she
+suggested I try a few things:
+
+* Use [the distributed scheduler](https://distributed.dask.org/en/stable/).
+* Ensure that [`pyarrow` string arrays are
+  used](https://docs.dask.org/en/latest/configuration.html#dask) instead of
+  numpy object arrays.
```
gforsyth, line 580:
```suggestion
  NumPy object arrays.
```

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+Additionally, I'm using lambda functions instead of taking advantage of Pandas'
+fast built-in methods like `count`, `nunique` and others.
+
+Let's see if we can alter the original query to give Pandas and Dask a leg up.
+
+#### A story of two `GROUP BY`s
+
+It looks like we can remove the double `group_by` by moving the second `mutate`
+expression directly into the first `group_by` call. That's a nice feature of
+ibis: you don't have to first assign and then group like you do with pandas.
```
phofl, line 403: You don't have to do this with pandas either

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+    print(f"```python\n{f.read()}\n```")
+```
+
+Let's run the code:
+
+```{python}
+dask_results = %timeit -n1 -r1 -o %run dask_impl.py
+```
+
+That's a nice improvement over pandas: we finished the workload. Dask is still
```
gforsyth, line 601:
```suggestion
That's a nice improvement over Pandas: we finished the workload. Dask is still
```

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+good ways behind DuckDB out of the box!
```
gforsyth, line 602:
```suggestion
a good ways behind DuckDB out of the box!
```

gforsyth: or, alternatively, "Dask is still behind DuckDB out of the box"

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+duckdb_duration = duckdb_results.best
+speedup = int(round(pandas_expected_duration / duckdb_duration, 0))
+
+print(
+    f"I was able execute this computation about **{speedup}x faster** than you can "
+    "expect with pandas."
+)
+```
+
+In contrast, pandas ran out of memory **on a single file** without some hand
```
gforsyth, line 664:
```suggestion
In contrast, Pandas ran out of memory **on a single file** without some hand
```

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+holding, and while Dask didn't cause the Linux OOM-killer to start terrorizing
+my program it took about 4x longer than DuckDB, used a lot more memory and
+required configuration to get the best performance.
+
+### Pandas is untenable for this workload
+
+I can't load everything into memory and given that like Ibis + DuckDB runs this
+workload on my machine it doesn't seem worth the effort to make pandas scale to
```
gforsyth, line 672:
```suggestion
workload on my machine it doesn't seem worth the effort to make Pandas scale to
```

pyproject.toml
```diff
@@ -480,6 +480,8 @@ required-imports = ["from __future__ import annotations"]
 "docs/*.py" = ["INP001"]
 "*/decompiled.py" = ["ALL"]
 "ci/release/verify_release.py" = ["T201"] # CLI tool that prints stuff
+"pandas_impl.py" = ["T201"] # prints output using `print`
+"dask_impl.py" = ["T201"] # prints output using `print`
```
gforsyth, line 484: Should we set this for `docs/**/*.py`?

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+```
+
+In contrast, pandas ran out of memory **on a single file** without some hand
+holding, and while Dask didn't cause the Linux OOM-killer to start terrorizing
+my program it took about 4x longer than DuckDB, used a lot more memory and
+required configuration to get the best performance.
+
+### Pandas is untenable for this workload
+
+I can't load everything into memory and given that like Ibis + DuckDB runs this
```
phofl, line 671: This is a use case where I wouldn't recommend pandas anyway, that's not what it was designed for

gforsyth (commented):
Nice post!  

My current comments are all language nits and trying to maintain consistent casing of project names in prose (I opted for "Dask", "Pandas", "Ibis", which I believe are the preferred casings for each project).

This also made me want PR docs previews even more.

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+That's a nice improvement over pandas: we finished the workload. Dask is still
+good ways behind DuckDB out of the box!
+
+## Takeaways
+
+**Ibis + DuckDB is the only system tested that handles this workload well out of the box**
+
+Pandas couldn't handle the workload due to memory constraints, and Dask
+required some fairly advanced knowledge to get maximum performance.
+
```
phofl, line 610: I don't think that this is accurate. Arrow strings are on by default when you have a new Dask version and the distributed scheduler is in 10 minutes to Dask (https://docs.dask.org/en/stable/10-minutes-to-dask.html#scheduling)

docs/posts/pydata-performance/dask_impl.py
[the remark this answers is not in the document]
cpcloud: Fixed!

docs/posts/pydata-performance/dask_impl.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import annotations
+
+import logging
+
+import dask.config
+import dask.dataframe as dd
```
phofl, line 6: The first thing is that I would use dask-expr instead of dask.dataframe

docs/posts/pydata-performance/index.qmd
phofl: A comment worth adding is the difference between eager and lazy which explains this section

phofl (commented):
Some comments, 2 overarching topics:

- Do you create an inefficient query on purpose in the beginning? E.g. apply, sorting before grouping while groupby sorts anyway? If yes, then please ignore my comments to that effect but it would be nice if you would add a comment that this is very inefficient for the use-case that you are processing
- Dask configurations, but more on that in the comments

Which versions of pandas and Dask are you using? I might have missed that information if it's already in there

docs/posts/pydata-performance/dask_impl.py
phofl: Sorry this one got duplicated

pyproject.toml
cpcloud: Maybe, though it's certainly possible that spurious prints are in there and they'd be rendered on execution in a `qmd`.

Let's leave it for now and if we starting needing to add this to a bunch of files then we can't blanket ignore `T201` then.

docs/posts/pydata-performance/index.qmd
cpcloud: Because that replicates the original query.

docs/posts/pydata-performance/dask_impl.py
cpcloud: `dask_expr` is a bit too new AFAIK and not in common use, that's why I didn't use it.

docs/posts/pydata-performance/index.qmd
cpcloud: Did you read the rest of the post 😂

docs/posts/pydata-performance/dask_impl.py
phofl: FWIW I am expecting it to be the default in dask before too long

docs/posts/pydata-performance/index.qmd
cpcloud: The comment here is correct: Pandas and Dask's use of object arrays is inefficient, because object arrays are inefficient.

This comment is not about what the default is or should be or could be

docs/posts/pydata-performance/index.qmd
cpcloud: Happy to use an alternative, but I doubt whether I do it in assign matters that much.

docs/posts/pydata-performance/index.qmd
cpcloud: Regardless of whether it's appropriate, people will try it anyway. I think it's fair to make this statement.

docs/posts/pydata-performance/index.qmd
cpcloud: Happy to change it.

I think it's worth nothing that I am immediately presented with having to use `Client` and having to think about `fork` behavior making the out of the box experience a little less seamless than it perhaps could be.

docs/posts/pydata-performance/dask_impl.py
```diff
@@ -0,0 +1,40 @@
...
+from dask.distributed import Client
+
+if __name__ == "__main__":
+    dask.config.set({"dataframe.convert-string": True})
```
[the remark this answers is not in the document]
cpcloud: Happy to remove it, I am on the latest release as of last week (2023.11.0)

docs/posts/pydata-performance/index.qmd
phofl: Yes 100% agree with you on this (I was just contesting the advanced part, because I think that's not correct)

cpcloud: For example, the code I wrote originally didn't use the distributed scheduler and I couldn't port it without stick all the code under `if __name__ == "__main__":`

docs/posts/pydata-performance/index.qmd
phofl: Are we talking about strings or general nested data structure? That's not totally clear to me.

It is correct for general nested data structure, but Dask doesn't use object arrays anymore for strings except if you explicitly ask for it

cpcloud: Happy to add something like "pyarrow strings have been the default in dask since 2023.7.1" if that helps

docs/posts/pydata-performance/dask_impl.py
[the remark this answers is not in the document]
phofl: You can remove it then, the setting is on by default

docs/posts/pydata-performance/index.qmd
cpcloud: Fair point, I can remove this bit, it is in fact possible to do this without assign in pandas, though inconvenient due to the way `extract` behaves

phofl: I think you can get rid of that inconvenience if you set ``expand=False`` in ``extract``

cpcloud: 😂 Potato potato I guess!

docs/posts/pydata-performance/index.qmd
cpcloud: Sweet, giving this a try.

Thanks for the tip!

cpcloud: Indeed this is much faster for loading. I'll report back if there are any issues running the query, or push the results if all goes well!

docs/posts/pydata-performance/index.qmd
cpcloud: I mention later that this sorting here is kind of pointless. It's intentionally here though so I plan on leaving it.

docs/posts/pydata-performance/index.qmd
cpcloud: Same as https://github.com/ibis-project/ibis/pull/7624/files#r1406490006: this is intentional (and later I show that this is not really the best way to accomplish our goal anyway)

docs/posts/pydata-performance/index.qmd
phofl: > I mention later that this sorting here is kind of pointless. It's intentionally here though so I plan on leaving it.

Yeah I found that too, that's why I added 

> If yes, then please ignore my comments to that effect

to the top comment. It just confused me (as a reader) that the query was so inefficient. 


As I said, feel free to ignore if this is intentional

docs/posts/pydata-performance/index.qmd
cpcloud: This is actually referring to list arrays, so I will adjust the language accordingly and I won't mention string arrays since they're not particularly relevant for the section.

docs/posts/pydata-performance/index.qmd
cpcloud: Adjusted this bit, let me know what you think.

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+print(
+    f"I was able execute this computation about **{speedup}x faster** than you can "
+    "expect with pandas."
+)
+```
+
+In contrast, Pandas ran out of memory **on a single file** without some hand
+holding, and while Dask didn't cause the Linux OOM-killer to start terrorizing
+my program it took about 4x longer than DuckDB, used a lot more memory and
+required configuration to get the best performance.
```
mrocklin, line 645: What was the configuration you're referring to here?  The use of `Client()`?  If so, this feels analagous to DuckDB's creation of a connection object (although maybe that's unnecessary with Ibis)

docs/posts/pydata-performance/dask_impl.py
mrocklin: Also, if it reassures you, you can use dask-expr and still win.  DuckDB generally does much better than dask-expr on single-machines.  In your shoes I wouldn't worry too much about comparing against the newer system.

cpcloud: > Also, if it reassures you, you can use dask-expr and still win.

Not super concerned about winning :)

If it turned out that Dask beat DuckDB on memory usage and latency then it would a super interesting data point for us.

The post is mostly to showcase Ibis as a viable performant alternative when dealing with "medium" data.

Noted about dask-expr!

cpcloud: @mrocklin Where are the rest of the TPC-H queries in that video? I only see 1-7.

phofl: We copied the existing implementations from the Polars repository and they only had 1-7

We will add more at some point, but that's not a super high priority at the moment

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,696 @@
...
+```{python}
+dask_results = %timeit -n1 -r1 -o %run dask_impl.py
+```
+
+That's a nice improvement over Pandas: we finished the workload. Dask is still
+good ways behind DuckDB out of the box!
+
+## Takeaways
+
+**Ibis + DuckDB is the only system tested that handles this workload well out of the box**
```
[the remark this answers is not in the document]
cpcloud: Not sure, it's coming in part 2.

docs/posts/pydata-performance/index.qmd
cpcloud: Fair point about `Client` vs connection.

We support calling `ibis.read_parquet` without the need for an explicit connection.

Here are main my complaints:

1. The distributed scheduler isn't the default :)
2. I have to be aware of `fork` behavior when using it (sticking my code under a main check)
3. I have to explicitly tell it to shutdown workers (to get them to release memory). I understand a context manager is an option, but now I have another level of nesting.

mrocklin: Yeah, I think I agree with the confusion around processes-in-processes and `if __name__ == "__main__":`, that's certainly a valid concern.

I guess when I hear people talk about "required configuration" I assume they mean "I had to research and set a bunch of options in a config file" similar to the historical complaints with Spark or similar systems.  This feels different from that.  Maybe "used a lot more memory and required me to create a `Client()` object to get the best performance" gives a more faithful description?

Not a big deal though.  This is just splitting hairs.

mrocklin: Also, maybe worth noting, but all of the dask examples tend to start with

```python
from dask.distributed import Client
client = Client()
```

phofl:
You shouldn't need this workaround either, flooring timestamps to month frequency should work with the arrow backend, e.g.

```
df.uploaded_on.dt.floor("M")
```

cpcloud:
> You shouldn't need this workaround either, flooring timestamps to month frequency should work with the arrow backend, e.g.
> 
> ```
> df.uploaded_on.dt.floor("M")
> ```

Appreciate the tip, but it's not going to help in this case because of the offset issue (perhaps that's a bug?) and the missing `str.extract` implementation.

phofl:
In this case I'd recommend setting 

```
pd.options.future.infer_string = True
```

instead of using the arrow backend.
This will only infer strings as Arrow backed and will become the default in pandas 3 anyway, that's closer to what I expect the average user to experience and migration is a lot smother since every other dtype stays the same.

cpcloud:
Won't `pd.options.future.infer_string = True` still result in the issue I mentioned?

The bugs above show that I can't execute the operations I need on Arrow strings, due to either an offset-related bug (the `contains` calls) or missing functionality (`extract` call).

phofl:
No the missing string implementations are patched for this option to use a slower path that falls back to a map call basically, but you won't run into unsupported functionality

cpcloud:
> No the missing string implementations are patched for this option to use a slower path that falls back to a map call basically, but you won't run into unsupported functionality

The fallback may be implemented, but the string concatenation offset issue is still present when using

```python
pd.options.future.infer_string = True
```

I am not using `dtype_backend="pyarrow"`.

phofl:
Yikes ok, this is something we will address in 2.2 in a couple of weeks (we are using pa.string() instead of pa.large_string() at the moment, which is susceptible to these overflows as soon as you have a take operation), sorry I forgot that this was still the case

cpcloud:
Would it help if I created a bug report on the pandas issue tracker?

phofl:
No not really (I am in the process of changing this)

gforsyth (commented):
few small editing suggestions and some formatting hiccups that slipped through.

cpcloud:
I ended up having to set `split_row_group=True` to get Dask to reliably not kill the kernel due to an OOM error. Setting this dramatically improved the runtime (yay!) getting the result to about 2x slower than DuckDB, down from 4x.

Dask's memory use is still about 10x worse than DuckDB, and having to twiddle with so many flags and other APIs made this fairly challenging.

cpcloud:
Ah, thanks for reminding me!

I tried using dask-expr before setting `split_row_group=True` and it seemed to take longer to finish and use more memory. I replaced `dask.dataframe` with `dask_expr`.

Perhaps it's worth someone on the Dask team trying to run this workload.

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,691 @@
...
+I thought it would be interesting to see how other tools compare to this setup,
+so I decided I'd try to do the same workflow on the same machine using
+a few tools from across the ecosystem.
+
+I chose two incumbents--[pandas](https://pandas.pydata.org/) and
+[dask](https://www.dask.org/)--to see how they compare to Ibis + DuckDB on this
+workload. In part 2 of this series I will compare two newer engines--Polars and
+DataFusion--to Ibis + DuckDB.
+
+I've worked on both Pandas and Dask in the past but it's been such a long time
```
lostmygithubaccount, line 36: s/Pandas/pandas throughout?

docs/posts/pydata-performance/index.qmd
cpcloud: Just to close this out (I've noted elsewhere too) I wasn't able to use `dtype_backend="pyarrow"` due to:

1. The offset issue
2. Missing implementation of `extract` for pyarrow strings
3. The month snapping workaround not working with pyarrow backend series'

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,691 @@
...
+```
+
+1. We've since implemented [a `flatten` method](../../reference/expression-collections.qmd#ibis.expr.types.arrays.ArrayValue.flatten)
+   on array expressions so it's no longer necessary to define a UDF here. I'll
+   leave this code unchanged for this post. **This has no effect on the
+   performance of the query**. In both cases the generated code contains
+   a DuckDB-native call to [its `flatten`
+   function](https://duckdb.org/docs/sql/functions/nested.html).
+2. This is a small change from the original query that adds a final sort key to
+   make the results deterministic.
```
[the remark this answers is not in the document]
cpcloud: Yep!

They're in the wrong place. I'll check to make sure they're being rendered.

docs/posts/pydata-performance/index.qmd
cpcloud: I originally changed from pandas to Pandas from a previous review.

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,691 @@
...
+
+#### Conclusion
+
+If I only have Pandas at my disposal, I'm unsure of how I can avoid getting
+a bigger computer to run this query over the entire data set.
+
+### Rewriting the query to be fair
+
+At this point I wondered whether this was a fair query to compare Dask and
+Pandas against.
```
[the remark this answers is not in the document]
cpcloud: Yep, will remove.

docs/posts/pydata-performance/index.qmd
```diff
@@ -0,0 +1,691 @@
...
+
+```{python}
+#| echo: false
+#| output: asis
+show_file("./step3.py")
+with open("./step3.py") as f:
+    exec(f.read())
+```
+
+1. I added a second sort key (`project_count`) for deterministic output.
```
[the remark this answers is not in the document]
cpcloud: 🤔 ok, darn :(

I will add a note in a comment in the code then.

docs/posts/pydata-performance/index.qmd
gforsyth: sorry, that's my bad.  I guess pandas prefers lowercase in prose?  Not a huge deal either way -- I think the most important thing is for it to be consistent (which it is)

cpcloud:
Heads up, I plan to publish this post shortly.

I will likely do a stream on my YouTube channel in a few days highlighting various points in the post.

cpcloud:
FWIW, I did a number of runs on a cloud machine with slightly different specs and am now seeing similar runtime performance between dask and duckdb (🎉) but memory is still an issue. Thanks for all the help trying to optimize things!

docs/posts/pydata-performance/index.qmd
lostmygithubaccount: it is always lowercase -- I was just trying to find where that's documented but it's ridiculous difficult to search their site for (or isn't there anymore)

jcrist: Cody is correct here - I can't find the pandas style guide, but an old colleague of mine has a separate repo defining some common pydata style conventions that also agrees here: https://github.com/MrPowers/pydata-style-guide

cpcloud: Adjusted.

cpcloud: The only variation is that I capitalize Pandas at the start of a sentence. Pandas can't override the rules of English :)

cpcloud:
Merging on green!

lostmygithubaccount (approved):
thanks for all the work on this! great post

PR #7625, opened 2023-11-27
refactor(duckdb): use pyarrow for all memtable registration

### LINKED ISSUE #7621: bug: all-NULL columns are turned to int32 during rountrip to parquet
### What happened?

Perhaps this is specific to duckdb? or perhaps it is a bug within duckdb itself?

```python
import ibis

before = ibis.memtable({"a": [None, None, None]}, schema={"a": "string"})
before.to_parquet("tmp.parquet")
after = ibis.read_parquet("tmp.parquet")
```

before appears as
```
┏━━━━━━━━┓
┃ a      ┃
┡━━━━━━━━┩
│ string │
├────────┤
│ NULL   │
│ NULL   │
│ NULL   │
└────────┘
```
after appears as
```
┏━━━━━━━┓
┃ a     ┃
┡━━━━━━━┩
│ int32 │
├───────┤
│  NULL │
│  NULL │
│  NULL │
└───────┘
```
I want the dtype to be preserved as string.

7.1.0

duckdb

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
Fixes #7621 

Potential performance impact for large dataframes getting converted to `pyarrow` before getting piped into PyArrow, but also... do we care?

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -27,7 +27,6 @@ from ibis.backends.base.sql.alchemy import AlchemyCrossSchemaBackend
 from ibis.backends.base.sqlglot import C, F
 from ibis.backends.duckdb.compiler import DuckDBSQLCompiler
 from ibis.backends.duckdb.datatypes import DuckDBType
-from ibis.expr.operations.relations import PandasDataFrameProxy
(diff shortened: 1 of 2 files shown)


NickCrews:
Thanks for the quick fix!

PR #7639, opened 2023-11-29
chore(flink): use `pytestmark` over repeated marks

deepyaman:
Two options:

1. [Expect `to_sql` to fail, because we don't have a SQLGlot dialect for Flink, and don't feel like anything is "close enough."](https://github.com/ibis-project/ibis/pull/7639/commits/aef728321981585e79d38fbb15155a920b652b28)
2. [Use Hive (or some other?) dialect as a close-enough stand-in, in order to enable SQL rendering.](https://github.com/ibis-project/ibis/pull/7639/commits/4492ae5d0b6dcb0a5f45906e071ba2ecd386f4bb)

---

Double-checked--`to_sql` doesn't work without a SQLGlot dialect. ~I don't believe we've identified one that's "close enough".~

Wonder if https://github.com/apache/flink/blob/76f754833298beb3e7589d521e75325adf283bf3/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/delegation/PlannerContext.java#L270 might indicate that can start with something that's a basic ANSI dialect? However, using dialect `""` (the base SQLGlot dialect) results in SQL rendering errors, so I've tried Hive.

### DIFF
--- a/ibis/backends/tests/test_sql.py
+++ b/ibis/backends/tests/test_sql.py
@@ -13,7 +13,7 @@ from ibis.backends.conftest import _get_backends_to_test
 sa = pytest.importorskip("sqlalchemy")
 sg = pytest.importorskip("sqlglot")
 
-pytestmark = pytest.mark.notimpl(["druid"])
+pytestmark = pytest.mark.notimpl(["druid", "flink"])
 
 
 @pytest.mark.never(
@@ -26,7 +26,6 @@ pytestmark = pytest.mark.notimpl(["druid"])
     reason="Not clear how to extract SQL from the backend",
     raises=(exc.OperationNotDefinedError, NotImplementedError, ValueError),
 )
-@pytest.mark.notimpl(["flink"], "WIP")
 def test_table(backend):
     expr = backend.functional_alltypes.select(c=_.int_col + 1)
     buf = io.StringIO()
@@ -84,7 +83,6 @@ no_sql_extraction = pytest.mark.notimpl(
 )
 @not_sql
 @no_sql_extraction
-@pytest.mark.notimpl(["flink"], "WIP")
 def test_literal(backend, expr):
     assert ibis.to_sql(expr, dialect=backend.name())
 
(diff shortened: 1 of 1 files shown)


deepyaman:
My concern is around:

1. Hive dialect won't necessarily support everything we need
2. The "right" dialect is obviously not Hive, since Flink SQL supports a Hive and Default dialect, and we're not using the Hive dialect. So even if it compiles, I think it could end up wrong

I'd be inclined to wait until we/somebody creates a Flink SQLGlot dialect, in which case I can roll back to [aef7283](https://github.com/ibis-project/ibis/pull/7639/commits/aef728321981585e79d38fbb15155a920b652b28) after we come to agreement

ibis/backends/tests/snapshots/test_generic/test_many_subqueries/flink/out.sql
```diff
@@ -0,0 +1,36 @@
...
+    t0.`street`,
+    t0.`key`
+  FROM t0
+  INNER JOIN t1
+    ON t0.`key` = t1.`key`
+), t3 AS (
+  SELECT
+    t2.`street`,
+    (
+      ROW_NUMBER() OVER (ORDER BY t2.`street`) - 1
```
deepyaman, line 23:
```suggestion
      ROW_NUMBER() OVER (ORDER BY t5.`street` ASC) - 1
    ) AS `key`
  FROM data AS t5
), t1 AS (
  SELECT
    t0.`key`
  FROM t0
), t2 AS (
  SELECT
    t0.`street`,
    t0.`key`
  FROM t0
  INNER JOIN t1
    ON t0.`key` = t1.`key`
), t3 AS (
  SELECT
    t2.`street`,
    (
      ROW_NUMBER() OVER (ORDER BY t2.`street` ASC) - 1
```

cpcloud (approved):
Thanks!

cpcloud:
@deepyaman Is this ready to merge? It's still in draft, but it seems good to go!

deepyaman:
> @deepyaman Is this ready to merge? It's still in draft, but it seems good to go!

@cpcloud I was waiting to figure out which route to take, but based on https://github.com/ibis-project/ibis/pull/7639#issuecomment-1831216606 and talking with @zhenzhongxu, I've reverted https://github.com/ibis-project/ibis/pull/7639/commits/4492ae5d0b6dcb0a5f45906e071ba2ecd386f4bb, and it should be ready to go.

PR #7640, opened 2023-11-29
fix(flink): fix recreating table/view issue on flink backend

zhenzhongxu:
When force recreating a Flink table from an in memory df, it currently fails because a previous view is created.

Solution:
* For in memory df, we'll create a temporary view instead of an external table. The main reason is in-memory df would only work for local experimentation, and would be difficult to ship the df to a remote cluster for execution to justify for being a external table. 
* Refactor and clean up some of the existing logic to be more fluid around temp view creation/removal.

Tested in notebook and the fix works well. ~~May need pointers from @chloeh13q on whether we have existing test cases implemented?~~  Tests are added now.

image

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -319,7 +319,7 @@ jobs:
       - uses: syphar/restore-pip-download-cache@v1
         with:
           requirement_files: poetry.lock
-          custom_cache_key_element: ${{ steps.install_python.outputs.python-version }}
+          custom_cache_key_element: ${{ matrix.os }}-${{ steps.install_python.outputs.python-version }}
 
       - name: install poetry
         run: python -m pip install --upgrade pip 'poetry==1.7.0'
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -419,7 +419,7 @@ class Backend(BaseBackend, CanCreateDatabase):
     def create_view(
         self,
         name: str,
-        obj: ir.Table,
+        obj: pd.DataFrame | ir.Table | None = None,
         *,
         database: str | None = None,
         catalog: str | None = None,
@@ -448,6 +448,10 @@ class Backend(BaseBackend, CanCreateDatabase):
         """
         if obj is None:
             raise exc.IbisError("The obj parameter is required")
+
+        if isinstance(obj, ir.Table):
+            # TODO(chloeh13q): implement CREATE VIEW for expressions
+            raise NotImplementedError
 
         if overwrite and self.list_views(name):
             self.drop_view(name=name, catalog=catalog, database=database, force=True)
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -12,6 +12,7 @@ import ibis.expr.types as ir
 from ibis.backends.base import BaseBackend, CanCreateDatabase
 from ibis.backends.base.sql.ddl import fully_qualified_re, is_fully_qualified
 from ibis.backends.flink.compiler.core import FlinkCompiler
+from ibis.backends.flink.datatypes import FlinkRowSchema
 from ibis.backends.flink.ddl import (
     CreateDatabase,
     CreateTableFromConnector,
@@ -19,7 +20,6 @@ from ibis.backends.flink.ddl import (
     DropTable,
     InsertSelect,
 )
-from ibis.backends.flink.utils import ibis_schema_to_flink_schema
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -354,7 +354,7 @@ class Backend(BaseBackend, CanCreateDatabase):
                 obj = obj.to_pandas()
             if isinstance(obj, pd.DataFrame):
                 table = self._table_env.from_pandas(
-                    obj, ibis_schema_to_flink_schema(schema)
+                    obj, FlinkRowSchema.from_ibis(schema)
                 )
             if isinstance(obj, ir.Table):
                 table = obj
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -19,6 +19,7 @@ from ibis.backends.flink.ddl import (
     DropTable,
     InsertSelect,
 )
+from ibis.backends.flink.utils import ibis_schema_to_flink_schema
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -144,6 +145,34 @@ class Backend(BaseBackend, CanCreateDatabase):
         )  # this is equivalent to the SQL query string `SHOW TABLES FROM|IN`,
         # but executing the SQL string directly yields a `TableResult` object
         return self._filter_with_like(tables, like)
+
+    def list_views(
+        self,
+        like: str | None = None,
+        temporary: bool = True,
+    ) -> list[str]:
+        """Return the list of view names.
+
+        Return the list of view names in the specified database and catalog.
+        or the default one if no database/catalog is specified.
+
+        Parameters
+        ----------
+        like : str, optional
+            A pattern in Python's regex format.
+        temporary : bool, optional
+            Whether to list temporary views or permanent views.
+
+        Returns
+        -------
+        list[str]
+            The list of the view names that match the pattern `like`.
+        """
+        if temporary:
+            views = self._table_env.list_temporary_views()
+        else:
+            views = self._table_env.list_views()
+        return self._filter_with_like(views, like)
 
     def _fully_qualified_name(
         self,
@@ -319,26 +348,29 @@ class Backend(BaseBackend, CanCreateDatabase):
         if obj is None and schema is None:
             raise exc.IbisError("The schema or obj parameter is required")
 
-        if overwrite:
-            self.drop_table(name=name, catalog=catalog, database=database, force=True)
+        # in-memory data is created as views in `pyflink`
+        elif obj is not None:
+            if isinstance(obj, pa.Table):
+                obj = obj.to_pandas()
+            if isinstance(obj, pd.DataFrame):
+                table = self._table_env.from_pandas(
+                    obj, ibis_schema_to_flink_schema(schema)
+                )
+            if isinstance(obj, ir.Table):
+                table = obj
+            return self.create_view(name, table, database=database, overwrite=overwrite)
 
-        if isinstance(obj, pa.Table):
-            obj = obj.to_pandas()
-        if isinstance(obj, pd.DataFrame):
-            qualified_name = self._fully_qualified_name(name, database, catalog)
-            table = self._table_env.from_pandas(obj)
-            # in-memory data is created as views in `pyflink`
-            # TODO(chloeh13q): alternatively, we can do CREATE TABLE and then INSERT
-            # INTO ... VALUES to keep things consistent
-            self._table_env.create_temporary_view(qualified_name, table)
-        if isinstance(obj, ir.Table):
-            # TODO(chloeh13q): implement CREATE TABLE for expressions
-            raise NotImplementedError
-        if schema is not None:
+        # external data is created as tables in `pyflink`
+        else:
             if not tbl_properties:
                 raise exc.IbisError(
                     "tbl_properties is required when creating table with schema"
                 )
+            if overwrite:
+                self.drop_table(
+                    name=name, catalog=catalog, database=database, force=True
+                )
+
             statement = CreateTableFromConnector(
                 table_name=name,
                 schema=schema,
@@ -349,8 +381,7 @@ class Backend(BaseBackend, CanCreateDatabase):
                 catalog=catalog,
             )
             self._exec_sql(statement.compile())
-
-        return self.table(name, database=database)
+            return self.table(name, database=database)
 
     def drop_table(
         self,
@@ -391,6 +422,7 @@ class Backend(BaseBackend, CanCreateDatabase):
         obj: ir.Table,
         *,
         database: str | None = None,
+        catalog: str | None = None,
         overwrite: bool = False,
     ) -> ir.Table:
         """Create a new view from an expression.
@@ -404,6 +436,8 @@ class Backend(BaseBackend, CanCreateDatabase):
         database
             Name of the database where the view will be created, if not
             provided the database's default is used.
+        catalog
+            Name of the catalog where the table exists, if not the default.
         overwrite
             Whether to clobber an existing view with the same name.
 
@@ -412,7 +446,15 @@ class Backend(BaseBackend, CanCreateDatabase):
         Table
             The view that was created.
         """
-        raise NotImplementedError
+        if obj is None:
+            raise exc.IbisError("The obj parameter is required")
+
+        if overwrite and self.list_views(name):
+            self.drop_view(name=name, catalog=catalog, database=database, force=True)
+
+        qualified_name = self._fully_qualified_name(name, database, catalog)
+        self._table_env.create_temporary_view(qualified_name, obj)
+        return self.table(name, database=database)
 
     def drop_view(
         self,
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -12,7 +12,6 @@ import ibis.expr.types as ir
 from ibis.backends.base import BaseBackend, CanCreateDatabase
 from ibis.backends.base.sql.ddl import fully_qualified_re, is_fully_qualified
 from ibis.backends.flink.compiler.core import FlinkCompiler
-from ibis.backends.flink.datatypes import FlinkRowSchema
 from ibis.backends.flink.ddl import (
     CreateDatabase,
     CreateTableFromConnector,
@@ -146,15 +145,14 @@ class Backend(BaseBackend, CanCreateDatabase):
         # but executing the SQL string directly yields a `TableResult` object
         return self._filter_with_like(tables, like)
 
-    def list_views(
+    def _list_views(
         self,
         like: str | None = None,
         temporary: bool = True,
     ) -> list[str]:
         """Return the list of view names.
 
-        Return the list of view names in the specified database and catalog.
-        or the default one if no database/catalog is specified.
+        Return the list of view names.
 
         Parameters
         ----------
@@ -311,12 +309,11 @@ class Backend(BaseBackend, CanCreateDatabase):
         name
             Name of the new table.
         obj
-            An Ibis table expression or pandas table that will be used to
-            extract the schema and the data of the new table. If not provided,
-            `schema` must be given.
+            An Ibis table expression, pandas DataFrame, or PyArrow Table that will
+            be used to extract the schema and the data of the new table. An
+            optional `schema` can be used to override the schema.
         schema
-            The schema for the new table. Only one of `schema` or `obj` can be
-            provided.
+            The schema for the new table. Required if `obj` is not provided.
         database
             Name of the database where the table will be created, if not the
             default.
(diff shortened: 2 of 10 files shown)


ibis/backends/flink/__init__.py
```diff
@@ -412,7 +444,15 @@ def create_view(
...
         """
-        raise NotImplementedError
+        if obj is None:
+            raise exc.IbisError("The obj parameter is required")
+        
+        if overwrite and self.list_views(name):
+            self.drop_view(name=name, catalog=catalog, database=database, force=True)
+            
+        qualified_name = self._fully_qualified_name(name, database, catalog)
+        self._table_env.create_temporary_view(qualified_name, obj)
```
zhenzhongxu, line 454: @deepyaman FYI, since we are creating views from the session environment, for this reason, among other potential, the remote cluster execution using existing CLI might be challenging. (we'll also have to check if we have other implementation that directly leverages session environment instead of creating direct SQL statements)

zhenzhongxu:
@chloeh13q can you give some pointers on where should i put the tests in?

ibis/backends/flink/utils.py
```diff
@@ -326,3 +327,14 @@ def translate_literal(op: ops.Literal) -> str:
         return f"ARRAY{list(value)}"
 
     raise NotImplementedError(f"No translation rule for {dtype}")
+
+
+def ibis_schema_to_flink_schema(schema: sch.Schema) -> RowType:
```
deepyaman, line 332: Not a change request, but a question (maybe one of the maintainers knows the answer best):

What is the "right" way to do this? For example, [other backends implement `SchemaMapper`](https://github.com/ibis-project/ibis/blob/96243416771109fbaf383167470ac2481fe938ef/ibis/backends/bigquery/datatypes.py#L101); is that the current recommended way, or is it right to do something simple like this?

ibis/backends/flink/tests/test_ddl.py
```diff
@@ -17,6 +17,34 @@
 except ImportError:
     Py4JJavaError = None
 
+_awards_players_schema = sch.Schema(
```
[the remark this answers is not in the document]
zhenzhongxu: so the schema can be used in non-fixture situation (e.g., [here](https://github.com/ibis-project/ibis/pull/7640/files/caab87f1ab772095f05b6fadd541249ba0cb2052#diff-e7a6d6edeb616d3cc2b66f70924b3edf4b0bc51df2e2d653b9551030bac00cd2R181))

ibis/backends/flink/utils.py
gforsyth: subclassing `SchemaMapper` is the way to go.

zhenzhongxu: Thanks for the suggestion. Implemented both SchemaMapper and TypeMapper now.

ibis/backends/flink/datatypes.py
```diff
@@ -0,0 +1,73 @@
...
+    fl.DataTypes.DATE(): dt.Date,
+    fl.DataTypes.TIME(): dt.Time,
+    fl.DataTypes.TIMESTAMP(): dt.Timestamp,
+}
+
+
+class FlinkRowSchema(SchemaMapper):
+    @classmethod
+    def from_ibis(cls, schema: sch.Schema) -> list[fl.RowType]:
+        if schema is None:
```
chloeh13q, line 53: Would `None` be an acceptable parameter? If so, should the type hinting be `sch.Schema | None`?

ibis/backends/flink/datatypes.py
```diff
@@ -0,0 +1,73 @@
...
+            [
+                fl.DataTypes.FIELD(key, FlinkType.from_ibis(type(value)))
+                for key, value in schema.fields.items()
+            ]
+        )
+
+
+class FlinkType(TypeMapper):
+    @classmethod
+    def to_ibis(cls, typ: fl.DataType, nullable=True) -> dt.DataType:
```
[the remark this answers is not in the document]
zhenzhongxu: Nice catch, will fix.

.github/workflows/ibis-backends.yml
```diff
@@ -318,7 +318,7 @@ jobs:
       - uses: syphar/restore-pip-download-cache@v1
         with:
           requirement_files: poetry.lock
-          custom_cache_key_element: ${{ steps.install_python.outputs.python-version }}
+          custom_cache_key_element: ${{ matrix.os }}-${{ steps.install_python.outputs.python-version }}
```
deepyaman, line 321:
```suggestion
          custom_cache_key_element: ${{ steps.install_python.outputs.python-version }}
```
Is this necessary? Cache can't conflict across different OS executors, I think, [especially because they're cached to different directory base don OS](https://github.com/syphar/restore-pip-download-cache/blob/main/src/utils.ts#L22-L36).

ibis/backends/flink/datatypes.py
```diff
@@ -0,0 +1,97 @@
+from __future__ import annotations
+
+import pyflink.table.types as fl
```
deepyaman, line 3:
```suggestion
from pyflink.table.types import DataTypes
```
Nit: Not a fan of just making up aliases unnecessarily, when they're not really used idiomatically anywhere. If you really want to namespace it (not necessary here, I think), at least `from pyflink.table import types` makes more sense to me.

ibis/backends/flink/tests/test_ddl.py
```diff
@@ -28,36 +56,12 @@ def reset_con(con):
...
-            "float_col": dt.float32,
-            "double_col": dt.float64,
-            "date_string_col": dt.string,
-            "string_col": dt.string,
-            "timestamp_col": dt.timestamp(scale=3),
-            "year": dt.int32,
-            "month": dt.int32,
-        }
-    )
+    return _functiona_alltypes_schema
```
deepyaman, line 64: Nit: Why is it `_functiona` instead of `_functional`? Is it just a typo that is getting propagated?

ibis/backends/flink/tests/test_ddl.py
```diff
@@ -17,6 +17,34 @@
...
+        "int_col": dt.int32,
+        "bigint_col": dt.int64,
+        "float_col": dt.float32,
+        "double_col": dt.float64,
+        "date_string_col": dt.string,
+        "string_col": dt.string,
+        "timestamp_col": dt.timestamp(scale=3),
+        "year": dt.int32,
+        "month": dt.int32,
+    }
```
deepyaman, line 45: What's the point in moving these out?

deepyaman (commented):
Minor comments, for the most part. Looks great overall!

ibis/backends/flink/datatypes.py
zhenzhongxu: I think one of the benefits is to allow readers to easily distinguish between Ibis provided type primitives vs backend provided primitives to potentially avoid human mistakes
e.g., before code is clearer to me with the namespace, otherwise i have to guess what DataTypes.STRING() is.
```
if typ == fl.DataTypes.STRING():
       return dt.String(nullable=nullable)
```

Also [big query backend](https://github.com/ibis-project/ibis/blob/master/ibis/backends/bigquery/datatypes.py#L3) seems to use this convention as well. I don't have a strong opinion on this one. Can be useful to get a third opinion.

ibis/backends/flink/tests/test_ddl.py
zhenzhongxu: not sure, i did not change the previous spelling in this PR. @chloeh13q can you help confirming the intent?

ibis/backends/flink/tests/test_ddl.py
zhenzhongxu: so the schema can be used in non-fixture situation (e.g., [here](https://github.com/ibis-project/ibis/pull/7640/files/caab87f1ab772095f05b6fadd541249ba0cb2052#diff-e7a6d6edeb616d3cc2b66f70924b3edf4b0bc51df2e2d653b9551030bac00cd2R181)). Is there a better way to do this though?

ibis/backends/flink/tests/test_ddl.py
zhenzhongxu: Not sure, I didn't change the previous implementation. @chloeh13q can you help confirming the intent?

ibis/backends/flink/tests/test_ddl.py
gforsyth: there's not really a good solution for parametrizing fixtures -- there's a pytest plugin called `lazy_fixtures` which will let you pass the fixture as a variable to parametrize, but what you have here is a reasonable solution.

ibis/backends/flink/datatypes.py
zhenzhongxu: Fair enough, I'll change to the recommendation.

.github/workflows/ibis-backends.yml
deepyaman: > It means either the cached content was corrupted or there is an edge scenario this can break (e.g., what if you change from a specific os version to latest)

If I had to guess, it's more likely the latter, so just changing the cache key circumvented it. In which case, I think just changing it helped. I think it's worth trying to revert this, but happy to let somebody else comment first.

ibis/backends/flink/tests/test_ddl.py
deepyaman: Almost certainly just a typo, since I don't see this anywhere else. Doesn't matter to me, can leave it and I'll fix it in a quick PR later, or can update it here. 🤷

zhenzhongxu: let's fix in a separate pr

ibis/backends/flink/__init__.py
```diff
@@ -145,6 +145,34 @@ def list_tables(
         # but executing the SQL string directly yields a `TableResult` object
         return self._filter_with_like(tables, like)
 
+    def list_views(
+        self,
+        like: str | None = None,
+        temporary: bool = True,
+    ) -> list[str]:
+        """Return the list of view names.
```
gforsyth, line 153: the convention elsewhere in Ibis is to fold in views to the output from `list_tables` -- does this need to be a separate API for flink?

ibis/backends/flink/__init__.py
```diff
@@ -315,30 +343,34 @@ def create_table(
...
-        if overwrite:
-            self.drop_table(name=name, catalog=catalog, database=database, force=True)
+        # in-memory data is created as views in `pyflink`
+        elif obj is not None:
+            if isinstance(obj, pa.Table):
+                obj = obj.to_pandas()
+            if isinstance(obj, pd.DataFrame):
+                table = self._table_env.from_pandas(
+                    obj, FlinkRowSchema.from_ibis(schema)
+                )
```
gforsyth, line 358: The docstring says that only one of `obj` and `schema` can be provided, but it looks like here it's not only accepted but required.

You can use `PandasSchema.to_ibis` and pass in a list of `(name, dtype)` tuples to generate a compatible Ibis schema from the input.  (PyArrow also has a similar option if you want to generate the schema there for pyarrow inputs)

gforsyth: Ahh, I see the `None` handling in the `FlinkRowSchema` below now -- in that case, you can either make the change here or update the docstring to note that a schema can optionally be specified alongside a dataframe or pyarrow table

ibis/backends/flink/__init__.py
```diff
@@ -412,7 +446,19 @@ def create_view(
         Table
             The view that was created.
         """
-        raise NotImplementedError
+        if obj is None:
+            raise exc.IbisError("The obj parameter is required")
+
```
gforsyth, line 451: Do we want to make this a required positional argument instead and skip the `None` handling?

ibis/backends/flink/__init__.py
```diff
@@ -145,6 +145,34 @@ def list_tables(
...
+        Return the list of view names in the specified database and catalog.
+        or the default one if no database/catalog is specified.
+
+        Parameters
+        ----------
+        like : str, optional
+            A pattern in Python's regex format.
+        temporary : bool, optional
+            Whether to list temporary views or permanent views.
```
gforsyth, line 163: This is exposing a lot of Flink stuff to users and maybe that's exactly what is needed, but does a user often need to distinguish between temporary and permanent views?  Or do they want to see the views in the current session / catalog?

Does it make sense to return the union of both?

ibis/backends/flink/__init__.py
```diff
@@ -315,30 +343,34 @@ def create_table(
...
-            table = self._table_env.from_pandas(obj)
-            # in-memory data is created as views in `pyflink`
-            # TODO(chloeh13q): alternatively, we can do CREATE TABLE and then INSERT
-            # INTO ... VALUES to keep things consistent
-            self._table_env.create_temporary_view(qualified_name, table)
-        if isinstance(obj, ir.Table):
-            # TODO(chloeh13q): implement CREATE TABLE for expressions
-            raise NotImplementedError
-        if schema is not None:
+        # external data is created as tables in `pyflink`
```
gforsyth, line 363: Is another way of thinking of "external data" as data that requires a "connector"?  Asking for my own understanding.

ibis/backends/flink/__init__.py
```diff
@@ -349,8 +381,7 @@ def create_table(
                 catalog=catalog,
             )
             self._exec_sql(statement.compile())
-
-        return self.table(name, database=database)
+            return self.table(name, database=database)
```
gforsyth, line 384:
```suggestion
            return self.table(name, database=database, catalog=catalog)
```

?

ibis/backends/flink/__init__.py
```diff
@@ -412,7 +446,19 @@ def create_view(
...
         """
-        raise NotImplementedError
+        if obj is None:
+            raise exc.IbisError("The obj parameter is required")
+
+        if isinstance(obj, ir.Table):
+            # TODO(chloeh13q): implement CREATE VIEW for expressions
+            raise NotImplementedError
+
+        if overwrite and self.list_views(name):
```
gforsyth, line 456:
```suggestion
        if overwrite and name in self.list_views():
```
I find this a little more readable

ibis/backends/flink/__init__.py
```diff
@@ -412,7 +446,19 @@ def create_view(
...
+            self.drop_view(name=name, catalog=catalog, database=database, force=True)
+
+        qualified_name = self._fully_qualified_name(name, database, catalog)
+        self._table_env.create_temporary_view(qualified_name, obj)
+        return self.table(name, database=database)
```
gforsyth, line 461:
```suggestion
        return self.table(name, database=database, catalog=catalog)
```
?

ibis/backends/flink/tests/test_ddl.py
```diff
@@ -117,6 +121,145 @@ def test_create_table(con, awards_players_schema, temp_table, csv_source_configs
...
+)
+def test_recreate_in_mem_table(
+    con, employee_df, schema_props, temp_table, csv_source_configs
+):
+    # create table once
+    schema = schema_props[0]
+    if schema_props[1] is not None:
+        tbl_properties = csv_source_configs(schema_props[1])
+    else:
+        tbl_properties = None
```
gforsyth, line 191:
```suggestion
@pytest.mark.parametrize(
    "schema, table_name", [(None, None), (_awards_players_schema, "awards_players")]
)
def test_recreate_in_mem_table(
    con, employee_df, schema, table_name, temp_table, csv_source_configs
):
    # create table once
    if table_name is not None:
        tbl_properties = csv_source_configs(table_name)
    else:
        tbl_properties = None
```

only a suggestion -- I always have trouble remembering what the various elements of a container are supposed to represent without names

gforsyth (commented):
A few questions and a few suggestions, but overall this looks good to me.

ibis/backends/flink/datatypes.py
```diff
@@ -0,0 +1,97 @@
...
+        elif dtype.is_float32():
+            return DataTypes.FLOAT()
+        elif dtype.is_float64():
+            return DataTypes.DOUBLE()
+        elif dtype.is_date():
+            return DataTypes.DATE()
+        elif dtype.is_time():
+            return DataTypes.TIME()
+        elif dtype.is_timestamp():
+            return DataTypes.TIMESTAMP()
```
gforsyth, line 91: Is there any nullability information we need to pass on to Flink?

Also, definitely not a blocker, but if you have a dictionary mapping the ibis types to the flink types it can make it easier to quickly glance and see how the types are mapped between the two

ibis/backends/flink/__init__.py
zhenzhongxu: it's the correct behavior right now that list_table returns views. Good point though this method is only used internally so can be private. I'll make the change.

ibis/backends/flink/__init__.py
zhenzhongxu: updated the docstring.

ibis/backends/flink/__init__.py
zhenzhongxu: Good point. I've made this method internal/private now (also for an previous feedback), to avoid exposing the complexity to users.

ibis/backends/flink/__init__.py
zhenzhongxu: Yes, your understanding is correct. So "external tables" require specific connector configurations (captured by `tbl_properties`) for Flink to connect to as either sources or sinks. 

Food for thought: it may require a deeper discussion, but how does Ibis think about reusable connectors? I'd love to see a world where connectors can implemented once and reused cross different engines or even runtimes (e.g., leverage arrow or in memory data sharing protocol). Today, the data ecosystem is plagued with connectors, every single vendor reimplements the same connectors over and over again. This may not be within the scope of Ibis, but wanted to throw the idea out there.

ibis/backends/flink/datatypes.py
zhenzhongxu: good catch, fixed the nullable types

ibis/backends/flink/__init__.py
```diff
@@ -282,12 +310,11 @@ def create_table(
         name
             Name of the new table.
         obj
-            An Ibis table expression or pandas table that will be used to
-            extract the schema and the data of the new table. If not provided,
-            `schema` must be given.
+            An Ibis table expression, pandas or arrow dataframe that will
```
gforsyth, line 313: going to get nitpicky here :smiley: 


```suggestion
            An Ibis table expression, pandas DataFrame, or PyArrow Table that will
```

ibis/backends/flink/__init__.py
```diff
@@ -145,6 +145,34 @@ def list_tables(
...
 
+    def _list_views(
+        self,
+        like: str | None = None,
+        temporary: bool = True,
+    ) -> list[str]:
+        """Return the list of view names.
+
+        Return the list of view names in the specified database and catalog.
+        or the default one if no database/catalog is specified.
```
gforsyth, line 156: Non-blocker since this method is now private, but should it accept `database` and `catalog` kwargs?

ibis/backends/flink/__init__.py
gforsyth: We like reusable things and avoiding many-to-many style connectors -- I'm not sure if Ibis has a role to play here beyond adoption and cheerleading.

gforsyth (commented):
one nitpicky docstring comment, but then this is good to go!

ibis/backends/flink/__init__.py
zhenzhongxu: turns out pyflink API doesn't expose database & catalog filters, (the underlying JNI API does but it may not be necessary to expose that for now). I'll update the docstring.

gforsyth (approved):
This looks good to me @zhenzhongxu !  Can you squash down the "fix" commits ("address comments", "address review" etc.) into the appropriate parent commits?  Then this is good to go in!

PR #7645, opened 2023-11-30
chore(flink): pin flink backend pyflink dependency version

zhenzhongxu:
pin pyflink to 1.17.1 until 1.18 official release


zhenzhongxu:
turns out not needed.

PR #7646, opened 2023-11-30
chore(flink): current pyflink release only supports up to python 3.10

zhenzhongxu:


zhenzhongxu:
@deepyaman looks like [pyflink only supports up to python 3.10](https://nightlies.apache.org/flink/flink-docs-stable/docs/dev/python/installation/), but we are currently testing on 3.11? it also seems reverting back to 3.9 fails our test. Is this issue related to flink backend release blocker?

deepyaman:
> @deepyaman looks like [pyflink only supports up to python 3.10](https://nightlies.apache.org/flink/flink-docs-stable/docs/dev/python/installation/), but we are currently testing on 3.11? (i suspect this causes the [test failure here](https://github.com/ibis-project/ibis/actions/runs/7042660878/job/19167286581#step:17:150))
> 
> it also seems reverting back to 3.9 fails our test. Is this issue related to flink backend release blocker?

(Copying from internal discussion, since I saw this)

https://github.com/ibis-project/ibis/blob/master/.github/workflows/ibis-backends.yml#L171-L284 see the top of the block, line 171; 3.11 is part of excluded config.

We made the initial decision to just get the backend working on one version (previously did have a matrix including 3.10). My guess is that there shouldn't be anything that doesn't work on 3.10, though. For the most part, Ibis just tests 3.11 and 3.9 (except for PySpark, tested on 3.10).

PR #7663, opened 2023-12-05
fix(geospatial): pretty print data in interactive mode

cpcloud:
Pretty print geospatial columns

### DIFF
--- a/ibis/backends/duckdb/tests/test_register.py
+++ b/ibis/backends/duckdb/tests/test_register.py
@@ -92,12 +92,13 @@ def test_load_spatial_when_geo_column(tmpdir):
     assert "spatial" in loaded_ext.to_pandas().name.values
 
 
-@pytest.mark.xfail(raises=NotImplementedError)
 def test_read_geo_to_pyarrow(con, data_dir):
     pytest.importorskip("geopandas")
+    shapely = pytest.importorskip("shapely")
+
     t = con.read_geo(data_dir / "geojson" / "zones.geojson")
-    # can't convert geometry to arrow type yet
-    assert t.head().to_pyarrow()
+    raw_geometry = t.head().to_pyarrow()["geom"].to_pandas()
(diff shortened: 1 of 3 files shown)


cpcloud:
@ncclementi I think the test you had there was pretty good. I added a conversion to shapely objects to the test since that's what's happening in the pretty printing, so this PR should be good to go.

gforsyth (approved):
:shipit:

ncclementi:
Edit: Woops, it looks like this went in, 

I was looking into add a test to this. I was trying to replicate what we have here: https://github.com/ibis-project/ibis/blob/6bc9e1522d0e480a2546ab82c0b7fa1d55df7c3b/ibis/backends/tests/test_client.py#L975-L996

But in `ibis/backends/duckdb/tests/test_client.py` because in `bis/backends/duckdb/tests/conftest.py` we have `zones` with a geometry column making it easier to give this a try. 

I added this at the bottom 

```python
@pytest.mark.parametrize("interactive", [True, False])
def test_repr_geo(zones, interactive, monkeypatch):
    monkeypatch.setattr(ibis.options, "interactive", interactive)

    expr_geo = zones.select("geom")
    expr = zones.select("zone")
    
    sg = repr(expr_geo)

    # no control characters
    assert all(c.isprintable() or c in "\n\r\t" for c in s)
    if interactive:
        assert "/" in sg
    else:
         assert "/" not in sg
```
But it fails a when it checks for `assert "/" in sg` when interactive, although it's unclear why. Because if I put a `breakpoint()` and check then things look just fine.

cpcloud:
It's failing that assertion because `"/"` isn't a substring of the geometry repr.

The `"/"` is a test based on the known contents of `date_string_col` which look like `month/day/year`.

ncclementi:
> The "/" is a test based on the known contents of date_string_col which look like month/day/year.
🤦 I was looking at the wrong column, time to wrap up the day.

PR #7665, opened 2023-12-05
fix(datafusion): ensure that non-matching re_search calls return bool values when patterns do not match

cpcloud:
Fixes an issue where we assume that non matching patterns return an empty array, when in fact they return null


cpcloud:
Branch name isn't appropriate I realize, but I don't think it's worth it to create a new PR just to change the branch name.

kszucs (approved):

PR #7675, opened 2023-12-06
feat: improve error for memtable(has_dupe_columns)

NickCrews:
I was accidentally doing this and was getting an attribute
error when ibis tried to lookup the
dtype on the following lines.
This should point users in the right direction better.

### DIFF
--- a/ibis/formats/pandas.py
+++ b/ibis/formats/pandas.py
@@ -92,6 +92,12 @@ class PandasData(DataMapper):
                 ibis_dtype = schema[column_name]
             else:
                 pandas_column = df[column_name]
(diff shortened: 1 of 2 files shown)


ibis/formats/pandas.py
```diff
@@ -92,6 +92,12 @@ def infer_table(cls, df, schema=None):
                 ibis_dtype = schema[column_name]
             else:
                 pandas_column = df[column_name]
+                if isinstance(pandas_column, pd.DataFrame):
+                    raise ValueError(
+                        "Converting Pandas DataFrames with duplicate "
```
NickCrews, line 97: Open to suggestions for improving this error message more.

PR #7690, opened 2023-12-08
build: move `docker-compose.yml` to `compose.yaml`

deepyaman:
Reading https://docs.docker.com/compose/compose-file/03-compose-file/, apparently `compose.yaml` is preferred over `compose.yml` and `docker-compose.yml`. 

Also, `docker-compose` is only maintained as an alias of `docker compose` for compatibility in Docker Compose V2, and the `version` key in the Compose file is only maintained for backwards compatibility.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,4 +1,3 @@
-version: "3.4"
 services:
   clickhouse:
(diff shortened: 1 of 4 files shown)


cpcloud (approved):

cpcloud:
Thanks!

PR #7692, opened 2023-12-08
fix(memtable): ensure column names match provided data

gforsyth:
Previously, the `pandas` memtable constructor was performing two tasks
inconsistently, one was to subselect columns out of the provided
dataframe, the other was to rename those columns.
This led to some weird behavior where a mismatch in provided names could
lead to a dataframe consisting of NaNs.

Now, `columns` can only be provided to rename existing columns and there
is no subselection behavior.  If the length of the `columns` iterable
does not match the number of columns in the provided data, we error.

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -947,6 +947,39 @@ def test_memtable_construct(backend, con, monkeypatch):
     )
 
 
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
LGTM! Thanks for working through this.

PR #7693, opened 2023-12-08
ci(deps): use non-deprecated method of updating nix flakes

cpcloud:
Fixes the currently-failing `update-deps.yml` workflow.


cpcloud:
The only way to test this is to merge it and run the update-deps workflow. Merging.

PR #7700, opened 2023-12-09
test(ddl): use temporary tables where possible

cpcloud:
This avoids leaving tables around if anything goes wrong after create table. Unfortunately yield fixtures are not able to handle all failure modes, whereas if we let the database handle the cleanup where possible we do not need to worry about these other failure modes.


cpcloud:
Closing this for now, it's no longer worth the effort.

PR #7702, opened 2023-12-10
docs(build): allow building individual docs without rendering api docs first

cpcloud:
Before this change the API docs needed to be generated to render any single page. This skips including API docs if requested files do not exist.

### DIFF
--- a/docs/contribute/05_reference.qmd
+++ b/docs/contribute/05_reference.qmd
@@ -33,5 +33,17 @@ class TestConf(BackendTest):
         ...
 ```
 
-{{< include ../reference/BackendTest.qmd >}}
-{{< include ../reference/ServiceBackendTest.qmd >}}
+```{python}
+#| echo: false
+#| output: asis
+import os
+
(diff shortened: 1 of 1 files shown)


docs/contribute/05_reference.qmd
```diff
@@ -33,5 +33,17 @@ class TestConf(BackendTest):
...
+import os
+
+paths = [
+    "../reference/BackendTest.qmd",
+    "../reference/ServiceBackendTest.qmd",
+]
+
+for path in filter(os.path.exists, paths):
+    with open(path) as f:
+        print(f.read())
```
cpcloud, line 48: I tried using

```python
    print(f"{{{{< include {path} >}}}}")
```

but that shows this warning

```
WARNING (/nix/store/8879djd960kr6waqa41a4ypxd0sc34fk-quarto-1.4.339/share/filters/main.lua:13666) Shortcode 'include' not found
```

And doesn't render anything.

PR #7703, opened 2023-12-10
docs(blog-post): pydata performance part 2; polars and datafusion

cpcloud:
This PR adds a new blog post as a follow up to [part 1](https://ibis-project.org/posts/pydata-performance/).

The purpose of the post is to continue comparing the performance of the same
workload from part 1 against Polars and DataFusion.

TL; DR: Polars and DataFusion take about the same time to complete this
workload, somewhere between 8 to 10 minutes.

DuckDB completes this workload in about 40 to 60 seconds.

#### Polars

Polars ran out of memory out of the box, so I added the `streaming=True` flag
to `collect` calls in all cases (the Ibis implementation and the Polars native
API implementation).

I am aware that streaming execution is in an alpha state, but I'm not aware of
any other way to complete the workload without getting a machine with more
memory.

Memory use was still pretty high (mid 30GBs) and the workload segfaulted when
run on a machine with 64GB (different from the one I benchmarked on which has
about 94GB), but I was able to reliably complete the workload on my larger
cloud host.

Note that I compare the Ibis version with the Polars native API version to rule
out any large differences in performance caused by Ibis. Both methods have the
same performance characteristics.

I took a look at `perf top` while the Polars native workload was running and I saw
a few expected things and some surprising-to-me things, especially the `getenv` calls:

image

#### DataFusion

DataFusion never ran out of memory and had a memory profile similar to DuckDB:
single digit GBs peak memory.

However, it was still extremely slow compared to DuckDB, about 9-10 minutes to
run the whole workload.

Similarly to Polars I compared both the Ibis implementation and a hand-written
SQL version (built from the generated Ibis code). Both had the same performance

I also looked at `perf top` while the DataFusion workload was running and saw this:

image

### Next steps

I'd like to work with the community to see if I can do something to improve the
performance in either or both of Polars and DataFusion.

cc @ritchie46 @alamb

Would love y'all to take a look at what I'm doing and let me know how to get
better performance.

Note: all of the version and system information is in the preview links below, but here it is for ease of use:

image


cpcloud:
The Polars backend test failures are from the change to set `streaming=True` and are related to:

1. Checking the size of streamed arrow record batches. That's an ibis testing issue.
2. Calling `polars.scan_ndjson` with a glob which panics:

```
FAILED ibis/backends/tests/test_register.py::test_read_json_glob[polars] - pyo3_runtime.PanicException: not yet implemented
```

cpcloud:
Found a difference, but it's unlikely to explain the 10x delta: I am calling `drop_nulls` before the aggregation in the polars code. Fixing it now.

ibis/backends/datafusion/compiler/relations.py
```diff
@@ -139,13 +139,30 @@ def _limit(op: ops.Limit, *, table, n, offset, **_):
...
+        selections = by_names_quoted + metrics
+        cols = [
+            sg.column(name, quoted=True)
+            for name in op.table.schema.names
+            if name not in by_names
+        ]
+        table = sg.select(*cols, *by).from_(table).subquery()
+    else:
+        selections = (by + metrics) or (STAR,)
+
```
cpcloud, line 161: I will extract these changes into another PR with a proper test case.

cpcloud: Done in #7707.

lostmygithubaccount:
I see some interesting behavior w/ Polars on a mac -- running the native Polars code w/ `streaming=False`, time is relatively close to DuckDB (~72s for Polars  vs ~53s for DuckDB). DuckDB fully saturates all cores, Polars seems to use about 80-90% throughout the run (perhaps corresponding to the difference in runtime).

image

however, when I run with `streaming=True`, the run takes over 5-7 minutes. I'm on the latest PyPI release

gforsyth (approved):
This is great!  Three small comments, but I'd say this is ready to go.

docs/posts/pydata-performance-part2/index.qmd
```diff
@@ -0,0 +1,279 @@
...
+
+<details>
+
+<summary>DataFusion SQL</summary>
+
+```{python}
+#| echo: false
+#| output: asis
+show_file("./datafusion_native.sql", language="sql")
+```
```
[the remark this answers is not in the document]
cpcloud: I did that intentionally because the SQL is longish, but given that you looked at the SQL and still made this comment I'll see whether it can be set to uncollapsed by default :)

docs/posts/pydata-performance-part2/index.qmd
```diff
@@ -0,0 +1,279 @@
...
+date: 2023-12-11
+categories:
+  - blog
+  - case study
+  - ecosystem
+  - performance
+---
+
+**TL; DR**: Ibis supports both Polars and DataFusion. Both backends are have
+about the same runtime performance, and lag far behind DuckDB on this workload.
```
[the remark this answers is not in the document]
cpcloud: Fixed!

cpcloud:
Rendering now

docs/posts/pydata-performance-part2/index.qmd
```diff
@@ -0,0 +1,287 @@
...
+df.head()
+```
+
+::: {.panel-tabset}
+
+::: {.callout-note}
+## The full dataset results are not shown here for DataFusion and Polars
+These two engines took anywhere from 7 to 10 minutes to complete the workload,
+making iteration on the blog post impractical.
+:::
```
gforsyth, line 134: I can't find this on the rendered page -- it might be getting swallowed by the panel?

docs/posts/pydata-performance-part2/index.qmd
```diff
@@ -0,0 +1,287 @@
...
+#| output: asis
+show_file("./duckdb_ibis.py")
+```
+
+```{python}
+duckdb_ibis_results = %timeit -n1 -r1 -o %run duckdb_ibis.py
+df.head()
+```
+
+::: {.panel-tabset}
```
gforsyth, line 128:
```suggestion
### DataFusion and Polars

::: {.panel-tabset}
```

I think having the extra header here helps break up the page a bit between DuckDB and these results, and should also (hopefully) add an entry to the RHS menu

gforsyth (commented):
Two small things I noticed, neither blocking

docs/posts/pydata-performance-part2/index.qmd
cpcloud: Hm, that should be deleted because I am showing the full dataset results.

cpcloud: Fixed

docs/posts/pydata-performance-part2/index.qmd
cpcloud: Fixed!

docs/posts/pydata-performance-part2/index.qmd
```diff
@@ -0,0 +1,283 @@
...
+#| output: asis
+show_file("./duckdb_ibis.py")
+```
+
+```{python}
+duckdb_ibis_results = %timeit -n1 -r1 -o %run duckdb_ibis.py
+df.head()
+```
+
+### DataFusion and Polars
```
gforsyth, line 128: This is not showing up in the rendered html for some reason (which is very annoying)

cpcloud:
Alright, once the preview finishes and looks good I will merge this!

Thanks all!

PR #7687, opened 2023-12-07
docs(blog-post): add blogpost ibis duckdb geospatial

ncclementi:
- Closes #7661

PR #7706, opened 2023-12-11
chore(deps): bump poetry2nix

cpcloud:
Brings in the fix from https://github.com/nix-community/poetry2nix/pull/1462.


cpcloud:
Merging to assist with #7687

PR #7707, opened 2023-12-11
fix(datafusion): support computed group by when the aggregation is count distinct

cpcloud:
This PR works around an issue in DataFusion where computed grouping keys in a `GROUP BY` with at least one `COUNT(DISTINCT ...)` aggregation are not executable due to what appears to be a bug in DataFusion.

The workaround is to replace the computed grouping keys with a projection of the computed keys and the usual subsequent group by.


cpcloud:
Let's wait for #7708 to land so the backend test here can work across all backends.

gforsyth (approved):
:shipit:

PR #7708, opened 2023-12-11
fix(mssql): compute the length of strings correctly

cpcloud:
Fix string length computation in the MS SQL backend with a glorious hack. Thanks to @arkanovicz for the suggestion.


gforsyth (approved):
Glorious, indeed! Let this hack shine down upon us and guide us through the dark (untrimmed) spaces.

PR #7713, opened 2023-12-11
fix(oracle): ensure that `.sql` metadata results are in column-definition order

### LINKED ISSUE #7711: bug(oracle): truly bizarre CI failure with incorrect columns
### What happened?

https://github.com/ibis-project/ibis/actions/runs/7168330883/job/19516290928

This is reproducible but only under a very specific set of conditions:

1. You must run with `pytest -m oracle --randomly-seed=1601193982` (this suggests some kind of test order dependency, possibly created by some underlying state in the backend that isn't cleaned up)
2. The container must be freshly launched. If you run the test two times in a row the first run will fail and the second run will pass.

Looking at the PDB output below the proximate cause is that the column names are out of order.

64d29575f

Oracle

### Relevant log output

```sh
(Pdb) df
                nbr_missions name
0    Chang-Diaz, Franklin R.  7.0
1             Ross, Jerry L.  7.0
2          Malenchenko, Yuri  6.0
3          Foale, C. Michael  6.0
4      Brown, Curtis L., Jr.  6.0
..                       ...  ...
559      Dobrovolski, Georgi  1.0
560     Schmitt, Harrison H.  1.0
561        Pogue, William R.  1.0
562            Jahn, Sigmund  1.0
563              Atkov, Oleg  1.0

[564 rows x 2 columns]
```

### Code of Conduct

cpcloud:
Fixes #7711.

### DIFF
--- a/ibis/backends/oracle/__init__.py
+++ b/ibis/backends/oracle/__init__.py
@@ -237,14 +237,19 @@ class Backend(BaseAlchemyBackend):
             sa.column("data_precision"),
             sa.column("data_scale"),
             sa.column("nullable"),
+            sa.column("column_id"),
         )
-        metadata_query = sa.select(
-            t.c.column_name,
-            t.c.data_type,
-            t.c.data_precision,
-            t.c.data_scale,
-            case((t.c.nullable == "Y", True), else_=False).label("nullable"),
-        ).where(t.c.table_name == name)
+        metadata_query = (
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
Nice work tracking this down. I'm pretty sure this is effectively untestable so we'll just have to keep fingers crossed and revisit if we need to.

PR #7714, opened 2023-12-11
feat(geospatial): add support for GeoTransform on duckdb

### LINKED ISSUE #7716: feat(geospatial): Add support for ST_Transform for duckdb
At the moment [GeoTransform op](https://github.com/ibis-project/ibis/blob/26a3c1f3d75963d649ea1f84b662da77fcb98e7c/ibis/expr/operations/geospatial.py#L446-L451C24), only takes one integer input (srid) this is because the op was done for postgress which contains the crs data in the geometry.  But duckdb doesn't, and it's st_transform function requires 2 different string values (st_transform(geometry, 'EPSG:4326', 'ESRI:102718')) first string value is current crs, second the one to transform to.

ncclementi:
- Closes #7716
@gforsyth ~~I have this so far, it's definitely broken. Trying to figure out how to make it work, I think this is a problem with the types but not super clear, any pointers would be helpful.~~

~~I added a test to make it easier to try things.~~

### DIFF
--- a/ibis/backends/duckdb/tests/conftest.py
+++ b/ibis/backends/duckdb/tests/conftest.py
@@ -43,7 +43,7 @@ class TestConf(BackendTest):
     def preload(self):
         if not SANDBOXED:
             self.connection._load_extensions(
-                ["httpfs", "postgres_scanner", "sqlite_scanner"]
+                ["httpfs", "postgres_scanner", "sqlite_scanner", "spatial"]
             )
 
     @property
@@ -59,16 +59,23 @@ class TestConf(BackendTest):
                 SELECT * FROM read_parquet('{parquet_dir / f'{table}.parquet'}')
                 """
             )
-        if geospatial_supported:
+        if geospatial_supported and not SANDBOXED:
             for table in TEST_TABLES_GEO:
                 yield (
                     f"""
-                    INSTALL spatial;
-                    LOAD spatial;
                     CREATE OR REPLACE TABLE {table} AS
                     SELECT * FROM st_read('{geojson_dir / f'{table}.geojson'}')
                     """
                 )
+            yield (
+                """
+                CREATE or REPLACE TABLE geo (name VARCHAR, geom GEOMETRY);
+                INSERT INTO geo VALUES
+                    ('Point', ST_GeomFromText('POINT(-100 40)')),
+                    ('Linestring', ST_GeomFromText('LINESTRING(0 0, 1 1, 2 1, 2 2)')),
+                    ('Polygon', ST_GeomFromText('POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))'));
+                """
+            )
         yield from super().ddl_script
 
     @staticmethod
@@ -113,3 +120,13 @@ def lines_gdf(data_dir):
     gpd = pytest.importorskip("geopandas")
     lines_gdf = gpd.read_file(data_dir / "geojson" / "lines.geojson")
     return lines_gdf
+
+
+@pytest.fixture(scope="session")
+def geotable(con):
+    return con.table("geo")
+
+
+@pytest.fixture(scope="session")
+def gdf(geotable):
+    return geotable.execute()
--- a/ibis/backends/duckdb/tests/test_geospatial.py
+++ b/ibis/backends/duckdb/tests/test_geospatial.py
@@ -195,3 +195,18 @@ def test_geospatial_buffer(zones, zones_gdf):
     gp_buffer = zones_gdf.geometry.buffer(100.0, resolution=8)
 
     gtm.assert_geoseries_equal(buffer.to_pandas(), gp_buffer, check_crs=False)
+
+
+# using a smaller dataset for time purposes
+def test_geospatial_convert(geotable, gdf):
+    # geotable is fabricated but let's say the
+    # data is in CRS: EPSG:2263
+    # let's transform to EPSG:4326 (latitude-longitude projection)
+    geo_ll = geotable.geom.convert("EPSG:2263", "EPSG:4326")
+
+    gdf.crs = "EPSG:2263"
+    gdf_ll = gdf.geometry.to_crs(crs=4326)
+
+    gtm.assert_geoseries_equal(
+        geo_ll.to_pandas(), gdf_ll, check_less_precise=True, check_crs=False
+    )
--- a/ibis/expr/types/geospatial.py
+++ b/ibis/expr/types/geospatial.py
@@ -671,6 +671,25 @@ class GeoSpatialValue(NumericValue):
         """
         return ops.GeoTransform(self, srid).to_expr()
 
+    def convert(
+        self, source: ir.StringValue, target: ir.StringValue | ir.IntegerValue
+    ) -> GeoSpatialValue:
+        """Transform a geometry into a new SRID (CRS).
+
+        Parameters
+        ----------
+        source
+            CRS/SRID of input geometry
+        target
+            Target CRS/SRID
+
+        Returns
+        -------
+        GeoSpatialValue
+            Transformed geometry
+        """
+        return ops.GeoConvert(self, source, target).to_expr()
+
     def line_locate_point(self, right: PointValue) -> ir.FloatingValue:
         """Locate the distance a point falls along the length of a line.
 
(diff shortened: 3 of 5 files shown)


ncclementi:
I got something to work, but it returns all Infinity, I checked against duckdb and this has the same problem. 
Duckdb reproducer: 

```python
import duckdb
con = duckdb.connect()
con.load_extension("spatial")

zones_geo = con.sql("select geom from st_read('https://raw.githubusercontent.com/ibis-project/testing-data/master/geojson/zones.geojson')") 
con.sql("select st_transform(zones_geo.geom, 'EPSG:2263', 'EPSG:4326') from zones_geo")
```

```
unable to open database file
Cannot open https://cdn.proj.org/us_noaa_nyhpgn.tif: Network functionality not available
```
Not quite sure what this means, but it ends up returning all infinity (matching the result on ibis)

ibis/backends/duckdb/tests/test_geospatial.py
```diff
@@ -195,3 +195,24 @@ def test_geospatial_buffer(zones, zones_gdf):
...
+    zones_geom_ll = zones.geom.convert("EPSG:2263", "EPSG:4326")
+
+    zones_gdf_ll = zones_gdf.geometry.to_crs(crs=4326)
+    # breakpoint()
+
+    # this doesn't work for all the points
+    # gtm.assert_geoseries_equal(zones_geom_ll.to_pandas(), zones_gdf_ll,  check_less_precise=True, check_crs=False)
+
+    # this will pass but it's very big tolerance
+    assert all(
```
ncclementi, line 213: I don't like that to get the test to pass the tolerance even after using [`set_precision`](https://shapely.readthedocs.io/en/stable/reference/shapely.set_precision.html) is quite high (1e-2), The test is also a bit slow. 
I tried using the lines dataset, and I can get it to pass with `1e-3` but this si still not ideal. 
Alternatively we can do just a snapshot test. 

Any thoughts? 
cc: @gforsyth and @cpcloud

cpcloud: Can we test against a hand-rolled table that we put in `duckdb.sql`, and also open an issue on `duckdb_spatial` about this?

ibis/expr/types/geospatial.py
```diff
@@ -671,6 +671,24 @@ def transform(self, srid: ir.IntegerValue) -> GeoSpatialValue:
...
+    ) -> GeoSpatialValue:
+        """Transform a geometry into a new SRID (CRS).
+
+        Parameters
+        ----------
+        source_crs
+            CRS of input geometry
+        target_crs
+            Target CRS
+        Returns
```
cpcloud, line 685:
```suggestion
            Target CRS

        Returns
```

ncclementi:
TODO
- [ ] Test against a hand-rolled table that we put in duckdb.sql, (test is slow)
- [ ]  Open an issue on duckdb_spatial about differences with geopandas. 
- [ ] Cleanup `breakpoints()`.

ncclementi:
I created a table that only had 3 geometries, and run the convert test, and for these simple values the transformation doesn't have any discrepancies. I will still open the issue on `duckdb` about the case with more bigger/complex geometries. But I think we are good to go then. 

cc: @cpcloud

ci/schema/duckdb.sql
```diff
@@ -51,3 +51,11 @@ CREATE OR REPLACE TABLE map (idx BIGINT, kv MAP(STRING, BIGINT));
 INSERT INTO map VALUES
     (1, MAP(['a', 'b', 'c'], [1, 2, 3])),
     (2, MAP(['d', 'e', 'f'], [4, 5, 6]));
+
+INSTALL spatial;
+LOAD spatial;
```
ncclementi, line 56: I added this because the sqlalchemy duckdb compatibility job was failing. Is that the right way to go about it?

gforsyth: I don't think we want this here. We should be auto loading the extension. If removing this line makes a test fail then we should figure out where we're missing the extension check.

ncclementi: I remove the lines, and we get the SQLAlchemy 2 DuckDB job failing.
 
I think what's happening is that the data on the `duckdb.sql` get loaded here https://github.com/ibis-project/ibis/blob/4780fcb90f28af81dbd14c52b6228e8144412c6f/ibis/backends/duckdb/tests/conftest.py#L49-L72

which is relying on the `super().ddl_script` to create the table here https://github.com/ibis-project/ibis/blob/4780fcb90f28af81dbd14c52b6228e8144412c6f/ibis/backends/tests/base.py#L72-L80

and when it gets to the geospatial table it doesn't have the extension to do so. But it's not clear to me, where we should pass this info if it's not on the sql.

gforsyth: Hmm, but the super call should come after we've loaded so spatial in the duckdb conftest.

gforsyth: I'm on mobile and the test output won't load on my phone but I'll take a look when I'm back at a keyboard

gforsyth: I'm guessing the nix failures are due to sandboxing and those can be skipped / xfailed the same way we did for the other tests

ncclementi: But aren't we just loading the extension for the tables in TEST_TABLE_GEO? 

https://github.com/ibis-project/ibis/blob/4780fcb90f28af81dbd14c52b6228e8144412c6f/ibis/backends/duckdb/tests/conftest.py#L62-L71

gforsyth: They all get run in the same session so it shouldn't matter. But try adding spatial to the extension list in `preload` and see if that sorts it out

ncclementi: >I'm guessing the nix failures are due to sandboxing and those can be skipped / xfailed the same way we did for the other tests

Maybe I'm missing something but the other tests on `test_geospatial.py` we didn't have to do the skip due to sandbox. But I'm seeing a lot of failures now on the nix jobs, because it's trying to create the geo table and it's missing the extension. Let's look at this later in today when you are hands on keyboard.

gforsyth: Ok.  I'm not entirely sure how this was working before (with the existing `st_read` calls), but now we don't load any `geospatial` stuff if were in a nix build.

ibis/backends/duckdb/tests/conftest.py
```diff
@@ -59,16 +59,23 @@ def ddl_script(self) -> Iterator[str]:
...
             for table in TEST_TABLES_GEO:
                 yield (
                     f"""
-                    INSTALL spatial;
-                    LOAD spatial;
                     CREATE OR REPLACE TABLE {table} AS
                     SELECT * FROM st_read('{geojson_dir / f'{table}.geojson'}')
                     """
                 )
+            yield (
```
ncclementi, line 70: This makes sense, it's a slightly different than what @cpcloud  suggested but it works 🤷‍♀️

gforsyth (approved):
Hey @ncclementi !  The code here looks good.  Can you resolve the conflict with `master` and then squash this down into 1 commit?

ibis/expr/types/geospatial.py
```diff
@@ -671,6 +671,25 @@ def transform(self, srid: ir.IntegerValue) -> GeoSpatialValue:
         """
         return ops.GeoTransform(self, srid).to_expr()
 
+    def convert(
+        self, source_crs: ir.StringValue, target_crs: ir.StringValue | ir.IntegerValue
```
cpcloud, line 675: Can call these `source` and `target`? Since people seem to interchange CRS and SRID, I think it might be less confusing if there's no specific nomenclature attached to the argument names.

This would also make it easier to map to backends, which may or may not use the CRS or SRID terminology.

ncclementi: That makes sense. I'll get it in a few min

ncclementi:
@gforsyth I think this is ready. Is it possible that the exasol job failure is unrelated. ?

gforsyth:
> @gforsyth I think this is ready. Is it possible that the exasol job failure is unrelated. ?

1000% unrelated

gforsyth (approved):

gforsyth:
Thanks @ncclementi !

PR #7728, opened 2023-12-12
fix(pyspark,polars): add packaging extra

gforsyth:
Both `pyspark` and `polars` backends rely on `packaging` in non-test
code but we aren't installing it as an extra for those backends.


cpcloud (approved):
Thank you!

PR #7730, opened 2023-12-12
feat(base): add `to_pandas` method to BaseBackend

gforsyth:
It just wraps `execute` but I thought we should have it for consistency,
since it is present on expressions.


cpcloud (approved):
Sweet!

PR #7736, opened 2023-12-13
refactor(snowflake): use upstream map-from-arrays function instead of a custom UDF

cpcloud:
Remove a custom UDF in favor of equivalent upstream functionality


gforsyth (approved):
good ol' `OBJECT`

PR #7741, opened 2023-12-13
fix(duckdb): ensure that parameter names are unlikely to overlap with column names

### LINKED ISSUE #7676: bug: argument name for lambda in Array.map() can't conflict with existing Column names
Prrobably related to https://github.com/ibis-project/ibis/issues/7626

### What happened?

```python
import ibis

t = ibis.memtable({"x": [[1, 2, 3]]})
# t.x.map(lambda y: y + 1)  # works fine
t.x.map(lambda x: x + 1)
```

results in `BinderException: Binder Error: Invalid parameter name 't0.x': must be unqualified`

11e03fa7003ff05bfc243a7cd16b0f791b488737

duckdb

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #7676.

I created https://github.com/duckdb/duckdb/issues/9981 because I think this is a bug in DuckDB.

For now this is a workaround to address the issue.

### DIFF
--- a/ibis/expr/operations/core.py
+++ b/ibis/expr/operations/core.py
@@ -10,6 +10,7 @@ import ibis.expr.datashape as ds
 import ibis.expr.datatypes as dt
 import ibis.expr.rules as rlz
 from ibis import util
+from ibis.common.annotations import attribute
 from ibis.common.bases import Abstract
 from ibis.common.graph import Node as Traversable
 from ibis.common.grounds import Concrete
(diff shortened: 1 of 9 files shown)


ibis/expr/operations/core.py
```diff
@@ -183,5 +184,9 @@ class Argument(Value):
     shape: ds.DataShape
     dtype: dt.DataType
 
+    @attribute
+    def param(self) -> str:
```
kszucs, line 188: I am confusing it with `ops.ScalarParameter` based in the attribute name. By the way this seems to be an implementation detail of the sql backends (in particular duckdb), could you implement the "unique parameter name" generation in the compiler instead?

cpcloud: I'd rather not.

cpcloud: I have to do it for each backend individually then, and that's not worth the benefit.

kszucs: Okey then, keep it :)

kszucs (approved):

PR #7580, opened 2023-11-17
refactor(ir): rethink the relational operations 

kszucs:
Aims to:
- solve the chained joined issue
- strict but simple integrity checks when constructing relational nodes
- column dereferencing support at the API level so the user can reference fields from parent relations
- provide a separate step to simplify node trees
- proper subquery/foreign field handling

Couple of simplification rules currently added:
- `t[t] ~> t` aka. `SELECT * FROM t`
- subsequent filters are squashed into a single filter containing all the conditions
- subsequent projections are reduced to the last one using value inlining based on the previous projections

PR #7430, opened 2023-10-24
refactor(analysis): remove tha `Projector`

kszucs:

PR #7752, opened 2023-12-14
refactor(ir): split the relational operations

kszucs:
### Rationale & History

In the last couple of years we have been constantly refactoring the internals to make it easier to work with. 
Although we have made great progress, the current codebase is still hard to maintain and extend. One example of that complexity is the try to remove the `Projector` class in #7430. I had to realize that we are unable to improve the internals in smaller incremental steps, we need to make a big leap forward to make the codebase maintainable in the long run.

One of the hotspots of problems is the `analysis.py` module which tries to bridge the gap between the user-facing API and the internal representation. Part of its complexity is caused by loose integrity checks in the internal representation, allowing various ways to represent the same operation. This makes it hard to inspect, reason about and optimize the relational operations. In addition to that, it makes much harder to implement the backends since more branching is required to cover all the variations. 

We have always been aware of these problems, and actually we had several attempts to solve them the same way this PR does. However, we never managed to actually split the relational operations, we always hit roadblocks to maintain compatibility with the current test suite. Actually we were unable to even understand those issues because of the complexity of the codebase and number of indirections between the API, analysis functions and the internal representation. 

But(!) finally we managed to prototype a new IR in #7580 along with implementations for the majority of the backends, including `various SQL backends` and `pandas`. After successfully validating the viability of the new IR, we split the PR into smaller pieces which can be individually reviewed. This PR is the first step of that process, it introduces the new IR and the new API. The next steps will be to implement the remaining backends on top of the new IR.

### What does the PR do:

- Split the `ops.Selection` and `ops.Aggregration` nodes into proper relational algebra operations.
- Almost entirely remove `analysis.py` with the technical debt accumulated over the years.
- More flexible window frame binding: if an unbound analytical function is used with a window containing references to a relation then `.over()` is now able to bind the window frame to the relation.
- Introduce a new API-level technique to dereference columns to the target relation(s).
- Revamp the subquery handling to be more robust and to support more use cases with strict validation, now we have `ScalarSubquery`, `ExistsSubquery`, and `InSubquery` nodes which can only be used in the appropriate context.
- Use way stricter integrity checks for all the relational operations, most of the time enforcing that all the value inputs of the node must originate from the parent relation the node depends on.
- Introduce a new `JoinChain` operations to represent multiple joins in a single operation followed by a projection attached to the same relation. This enabled to solve several outstanding issues with the join handling (including the notorious chain join issue).
- Use straightforward rewrite rules collected in `rewrites.py` to reinterpret user input so that the new operations can be constructed, even with the strict integrity checks.
- Provide a set of simplification rules to reorder and squash the relational operations into a more compact form.
- Use mappings to represent projections, eliminating the need of internally storing `ops.Alias` nodes. In addition to that table nodes in projections are not allowed anymore, the columns are expanded to the same mapping making the semantics clear.
- Uniform handling of the various kinds of inputs for all the API methods using a generic `bind()` function.

#### Advantages:
- The operations are much simpler with clear semantics.
- The operations are easier to reason about and to optimize.
- The backends can easily lower the internal representation to a backend-specific form before compilation/execution, so the lowered form can be easily inspected, debugged, and optimized.
- The API is much closer to the users' mental model, thanks to the dereferencing technique.
- The backend implementation can be greatly simplified due to the simpler internal representation and strict integrity checks. As an example the pandas backend can be slimmed down by 4k lines of code while being more robust and easier to maintain.

#### Disadvantages:
- The backends must be rewritten to support the new internal representation.

### Technicalities

Used the following command to cherry-pick changes related to the `IR` from #7580 
```
git format-patch --subject-prefix=ir --stdout master..newrels -- ibis/common ibis/formats ibis/expr ibis/tests/expr ibis/*.py | git am -3 -k
```

### TODOs

These follow-ups must be addressed before doing a release:
- https://github.com/ibis-project/ibis/issues/7820
- https://github.com/ibis-project/ibis/issues/7819


ibis/tests/expr/test_table.py
```diff
@@ -1318,10 +1481,23 @@ def test_group_by_keys(table):
 def test_having(table):
     m = table.mutate(foo=table.f * 2, bar=table.e / 2)
 
-    expr = m.group_by("foo").having(lambda x: x.foo.sum() > 10).size()
-    expected = m.group_by("foo").having(m.foo.sum() > 10).size()
+    # TODO(kszucs): the test didn't contain the `.agg()` call before, the
```
kszucs, line 1484: this should be resolved now

ibis/expr/types/joins.py
```diff
@@ -0,0 +1,191 @@
...
+        for table in self._relations():
+            for k, v in table.values.items():
+                if isinstance(v, ops.Field) and v.rel not in tables:
+                    subs[v] = ops.Field(table, k)
+        values = {k: v.replace(subs, filter=ops.Value) for k, v in values.items()}
+
+        return self.finish(values)
+
+    # TODO(kszucs): figure out a solution to automatically wrap all the
+    # TableExpr methods including the docstrings and the signature
```
kszucs, line 153: Need to expose the full table expr API

ibis/expr/operations/relations.py
```diff
@@ -1,671 +1,466 @@
...
-
-    In SQL, you might have:
-
-    SELECT DISTINCT foo
-    FROM table
-
-    SELECT DISTINCT foo, bar
-    FROM table
-    """
+class View(PhysicalTable):
```
kszucs, line 417: Remove this.

ibis/expr/operations/relations.py
```diff
@@ -1,671 +1,466 @@
...
+class SQLQueryResult(Relation):
+    """A table sourced from the result set of a select query."""
 
-        # invariant due to Aggregation and AggregateSelection requiring a valid
-        # Selection
-        assert all(shares_some_roots(predicate, table) for predicate in predicates)
+    query: str
+    schema: Schema
+    source: Any
+    fields = FrozenDict()
```
kszucs, line 399: This must be `values` instead of `fields`, doesn't have a test case for it just in the SQL backends, perhaps we could move the entire op there?

ibis/expr/analysis.py
cpcloud, line 1: image

ibis/expr/types/joins.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # construct a new join link and add it to the join chain
+        link = ops.JoinLink(how, table=right, predicates=preds)
+        left = left.copy(rest=left.rest + (link,), values=values)
+
+        # return with a new JoinExpr wrapping the new join chain
+        return self.__class__(left)
+
+    def select(self, *args, **kwargs):
+        """Select expressions."""
```
cpcloud, line 135: Before we merge, let's copy the `Table.select` signature and docstring here and do the same for the rest of the APIs.

ibis/expr/types/relations.py
```diff
@@ -104,8 +99,76 @@ def f(  # noqa: D417
     return f
 
 
+# TODO(kszucs): cover it with tests
```
cpcloud, line 102: If we're covering it by other means, then I think it's fine to leave it as is unless you have specific cases you want to test that have no overlap with hitting this function through the user APIs.

ibis/expr/types/relations.py
```diff
@@ -104,8 +99,76 @@ def f(  # noqa: D417
...
+            )
+        if isinstance(node, ops.Alias):
+            result[node.name] = node.arg
+        else:
+            result[node.name] = node
+    return result
+
+
+def dereference_values(parents, values):
+    """Trace and replace fields from earlier relations in the hierarchy."""
```
cpcloud, line 156: This needs some elaboration, because it underlies a ton of the changes to generated SQL.

For example:

1. Why is the replace valid?
2. In what cases is it **not** valid?

ibis/expr/types/relations.py
```diff
@@ -104,8 +99,76 @@ def f(  # noqa: D417
...
+                    v = v.rel.values.get(v.name)
+            elif v.relations:
+                # do not dereference literal expressions
+                mapping[v] = ops.Field(parent, k)
+    return {k: v.replace(mapping, filter=ops.Value) for k, v in values.items()}
+
+
 @public
-class Table(Expr, _FixedTextJupyterMixin):
+class TableExpr(Expr, _FixedTextJupyterMixin):
```
cpcloud, line 171: Why are we changing the name back to `TableExpr`?

ibis/expr/types/relations.py
```diff
@@ -368,6 +431,12 @@ def __interactive_rich_console__(self, console, options):
         table = to_rich_table(self, width)
         return console.render(table, options=options)
 
+    def column(self, name: str | int) -> ir.Column:
```
cpcloud, line 434: Is this a new API? If so, it should go into a separate PR.

Try to avoid introducing any new public APIs in this PR, it makes it much harder to review.

Since you're using it below, make it private for now to speed up the review.

ibis/expr/types/relations.py
```diff
@@ -778,7 +819,7 @@ def schema(self) -> sch.Schema:
 
     def group_by(
         self,
-        by: str | ir.Value | Iterable[str] | Iterable[ir.Value] | None = None,
+        by: str | ir.Value | Iterable[str] | Iterable[ir.Value] = (),
```
cpcloud, line 822: Changing the default is fine, but getting rid of `None` support is a user facing change. What if someone is calling `group_by(by=None, metrics=...)`? That's a legitimate call. Why break it?

ibis/expr/types/relations.py
```diff
@@ -1003,33 +1049,53 @@ def aggregate(
...
+        groups = bind(self, by)
+        metrics = bind(self, (metrics, kwargs))
+        having = bind(self, having)
+
+        groups = unwrap_aliases(groups)
+        metrics = unwrap_aliases(metrics)
+        having = unwrap_aliases(having)
+
+        # TODO(kszucs): dereferecence CountStar and CountStarDistinct nodes
+        # called on the parent tables
```
cpcloud, line 1066: Still a TODO?

ibis/expr/types/relations.py
```diff
@@ -1003,33 +1049,53 @@ def aggregate(
...
+
+        groups = dereference_values(self.op(), groups)
+        metrics = dereference_values(self.op(), metrics)
+        having = dereference_values(self.op(), having)
+
+        # TODO(kszucs): must consider both the groups and the metrics for the
```
cpcloud, line 1072: What is the actual TODO here?

ibis/expr/types/relations.py
```diff
@@ -1797,21 +1837,13 @@ def mutate(
...
-        )
-        mutation_exprs = an.get_mutation_exprs(list(exprs), self)
-        return self.select(mutation_exprs)
+        # string and integer inputs are going to be coerced to literals instead
+        # of interpreted as column references like in select
+        node = self.op()
+        values = bind(self, (exprs, mutations), prefer_column=False)
+        values = unwrap_aliases(values)
+        values = {**node.fields, **values}
+        return self.select(**values)
```
cpcloud, line 1846:
```suggestion
        return self.select(**node.fields, **values)
```

Just chuck 'em all in there in one call!

ibis/expr/types/relations.py
```diff
@@ -2911,29 +2935,20 @@ def join(
...
-            "right": ops.RightJoin,
-            "left_semi": ops.LeftSemiJoin,
-            "semi": ops.LeftSemiJoin,
-            "anti": ops.LeftAntiJoin,
-            "cross": ops.CrossJoin,
-        }
-
-        klass = _join_classes[how.lower()]
-        expr = klass(left, right, predicates).to_expr()
+        # the first participant of the join shouldn't be a self reference
```
cpcloud, line 2940: Ok, but **why**?

ibis/expr/types/relations.py
```diff
@@ -2911,29 +2935,20 @@ def join(
...
+        left = left.op()
+        if isinstance(left, ops.SelfReference):
+            left = left.parent
 
-        # semi/anti join only give access to the left table's fields, so
-        # there's never overlap
-        if how in ("left_semi", "semi", "anti"):
-            return expr
+        # construct an empty join chain and wrap it with a JoinExpr
```
cpcloud, line 2945: This comment doesn't seem that helpful, it's describing the code below which isn't obfuscated in any way by indirection.

ibis/expr/types/relations.py
```diff
@@ -2976,14 +2991,20 @@ def asof_join(
...
-            left=left,
-            right=right,
-            predicates=predicates,
-            by=by,
-            tolerance=tolerance,
-        )
-        return ops.relations._dedup_join_columns(op.to_expr(), lname=lname, rname=rname)
+        if by:
+            # perform a regular join on the by columns first, then perform the
+            # asof join on the result
```
cpcloud, line 2996: This is also describing **what** rather than **why**.

ibis/tests/expr/test_table.py
```diff
@@ -242,14 +249,16 @@ def test_projection_no_expr(table, empty):
...
 
-def test_projection_invalid_nested_list(table):
-    errmsg = "must be coerceable to expressions"
-    with pytest.raises(com.IbisTypeError, match=errmsg):
-        table.select(["a", ["b"]])
-    with pytest.raises(com.IbisTypeError, match=errmsg):
-        table[["a", ["b"]]]
-    with pytest.raises(com.IbisTypeError, match=errmsg):
-        table["a", ["b"]]
+# FIXME(kszucs): currently bind() flattens the list of expressions, so arbitrary
```
cpcloud, line 252: Does this need to be fixed?

ibis/tests/expr/test_table.py
```diff
@@ -674,22 +719,23 @@ def test_aggregate_keywords(table):
...
-def test_filter_aggregate_pushdown_predicate(table):
-    # In the case where we want to add a predicate to an aggregate
-    # expression after the fact, rather than having to backpedal and add it
-    # before calling aggregate.
-    #
-    # TODO (design decision): This could happen automatically when adding a
-    # predicate originating from the same root table; if an expression is
-    # created from field references from the aggregated table then it
-    # becomes a filter predicate applied on top of a view
+# TODO(kszucs): revisit this test case, but this usage should not be allowed
```
cpcloud, line 722: Does this still need to be revisited?

ibis/tests/expr/test_table.py
```diff
@@ -698,31 +744,17 @@ def test_filter_on_literal_then_aggregate(table):
...
-
-
 def test_group_by_having_api(table):
     # #154, add a HAVING post-predicate in a composable way
     metric = table.f.sum().name("foo")
     postp = table.d.mean() > 1
 
-    expr = table.group_by("g").having(postp).aggregate(metric)
+    # TODO(kszucs): update table.aggregate() to automatically add the
+    # post-predicate to the list of metrics, also ensure that fields
```
cpcloud, line 753: Still a TODO?

ibis/tests/expr/test_table.py
```diff
@@ -1064,41 +1189,43 @@ def test_filter_join():
...
-    assert joined.columns == [
-        "foo",
-        "bar",
-        "value1",
-        "foo_right",
-        "bar_right",
-        "value2",
-    ]
+# TODO(kszucs): the inner join convenience to don't duplicate the equivalent
+# columns from the right table is not implemented yet
```
cpcloud, line 1193: Ah, this answers my question above!

cpcloud (changes requested):
Lots of great stuff here!

I think the PR description needs some prose :)

Here's some possible questions to answer in the description:

- How are you solving the chained join problem?
- What are the intended semantics of table references: i.e., what constitutes an integrity error?
- How does the subquery flattening and column pruning work?
- What still needs work and whether it can be done in a follow up
- Are there any expected or known breakages to user code

Ideally most of the core maintainers can read the description and use it as a guide for the PR.

ibis/expr/operations/reductions.py
```diff
@@ -199,15 +208,15 @@ def dtype(self):
...
+    quantile: Scalar[dt.Numeric]
 
     dtype = dt.float64
 
 
 @public
 class MultiQuantile(Filterable, Reduction):
     arg: Value
-    quantile: Value[dt.Array[dt.Float64]]
+    quantile: Scalar[dt.Array[dt.Float64]]
```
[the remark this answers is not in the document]
kszucs: Ooops, will do!

ibis/expr/operations/core.py
```diff
@@ -32,13 +32,18 @@ def op(self) -> Self:
...
-    @abstractmethod
+    # @abstractmethod
     def to_expr(self):
-        ...
+        raise NotImplementedError
 
     # Avoid custom repr for performance reasons
     __repr__ = object.__repr__
 
+    # TODO(kszucs): hidrate the __children__ traversable attribute
```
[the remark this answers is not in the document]
kszucs: Yes, I'm going to leave that note and later we can benchmark its effect.

ibis/expr/rewrites.py
```diff
@@ -73,9 +82,80 @@ def rewrite_sample(_):
...
+    preds = tuple(v.replace(rule) for v in _.predicates)
+
+    inner = ops.Filter(y.parent, preds)
+    rule = p.Field(y.parent, name) >> d.Field(inner, name)
+    projs = {k: v.replace(rule) for k, v in y.values.items()}
+
+    return ops.Project(inner, projs)
+
+
+# TODO(kszucs): @replace(p.Aggregate(y @ p.Project)) -> Aggregate()
```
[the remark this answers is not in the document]
kszucs: Not yet, removing.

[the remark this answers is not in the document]
kszucs: Removed.

ibis/expr/decompile.py
```diff
@@ -87,7 +84,7 @@ def translate(op, *args, **kwargs):
 
 
 @translate.register(ops.Value)
-@translate.register(ops.TableNode)
+# @translate.register(ops.TableNode)
```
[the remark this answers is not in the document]
kszucs: Removed.

ibis/expr/tests/snapshots/test_sql/test_parse_sql_table_alias/decompiled.py
```diff
@@ -1,9 +1,7 @@
 import ibis
 
 
-employee = ibis.table(
+result = ibis.table(
     name="employee",
     schema={"first_name": "string", "last_name": "string", "id": "int64"},
 )
-
-result = employee.select([employee.first_name, employee.last_name, employee.id])
```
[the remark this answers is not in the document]
kszucs: Yep, I love it as well.

ibis/expr/types/core.py
```diff
@@ -351,6 +352,25 @@ def compile(
             self, limit=limit, timecontext=timecontext, params=params
         )
 
+    def optimize(self, enable_reordering: bool = True) -> ir.Expr:
```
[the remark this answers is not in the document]
kszucs: Yes, I'm going to move this optimize method to `ibis.expr.rewrites.simplify()` or something like that.

ibis/expr/operations/reductions.py
```diff
@@ -319,7 +328,7 @@ class ApproxMedian(Filterable, Reduction):
 @public
 class GroupConcat(Filterable, Reduction):
     arg: Column
-    sep: Value[dt.String]
+    sep: Scalar[dt.String]
```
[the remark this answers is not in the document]
kszucs: Restoring.

ibis/expr/operations/sortkeys.py
```diff
@@ -28,6 +28,7 @@
 class SortKey(Value):
     """A sort operation."""
 
+    # TODO(kszucs): rename expr to arg or something else except expr
```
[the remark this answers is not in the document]
kszucs: I had no intention to update it in this PR, but I'm leaving the todo note.

ibis/expr/rewrites.py
```diff
@@ -1,81 +1,217 @@
...
+
+#     Errors as unsupported if a `seed` is specified.
+#     """
+
+#     if _.seed is not None:
+#         raise UnsupportedOperationError(
+#             "`Table.sample` with a random seed is unsupported"
+#         )
+
+#     return ops.Filter(_.parent, (ops.LessEqual(ops.RandomScalar(), _.fraction),))
```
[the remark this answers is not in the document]
kszucs: You tell me :) 
Are we going to need them in the backends?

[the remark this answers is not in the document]
cpcloud: Hm 🤔 I'm not sure they'll be needed in this exact form since a base class implementation of `visit_$OP` might suffice.

Let's leave 'em for now :)

ibis/expr/types/generic.py
```diff
@@ -756,21 +755,22 @@ def over(
...
                     "No reduction or analytic function found to construct a window expression"
                 )
             return expr
 
-        op = self.op()
         if isinstance(window, bl.WindowBuilder):
-            if table := an.find_first_base_table(self.op()):
-                return bind(table)
+            if node.relations:
+                table, *_ = node.relations
```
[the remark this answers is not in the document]
kszucs: Resolved.

ibis/expr/rewrites.py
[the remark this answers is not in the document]
kszucs: Yeah, these seem to be backend specific enough so they should either be removed or moved to the respective backends. I'm going to delete these, we can easily restore them.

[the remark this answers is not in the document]
kszucs: Removed them, we can restore if they are needed.

ibis/expr/types/joins.py
```diff
@@ -0,0 +1,191 @@
...
+            yield ops.Equals(left_value, right_value).to_expr()
+
+
+@public
+class JoinExpr(TableExpr):
+    def join(
+        self,
+        right,
+        predicates: Any,
+        # TODO(kszucs): add typehint about the possible join kinds
```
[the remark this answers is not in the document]
kszucs: Was easy enough. I introduced a `JoinKind` type and reused it 3 places.

ibis/expr/types/relations.py
kszucs: I prototyped everything in a single file at the beginning and I had name collisions. Restoring it.

kszucs: Restored.

ibis/expr/operations/relations.py
kszucs: Removed.

ibis/expr/operations/relations.py
kszucs: Updated.

ibis/expr/operations/temporal_windows.py
```diff
@@ -5,18 +5,26 @@
...
 
 import ibis.expr.datatypes as dt
+from ibis.common.annotations import attribute
 from ibis.expr.operations.core import Column, Scalar  # noqa: TCH001
 from ibis.expr.operations.relations import Relation
 from ibis.expr.schema import Schema
 
 
+# TODO(kszucs): these should be window frames specialized for timeseries operations
+# but they should inherit from ops.WindowFrame
```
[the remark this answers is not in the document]
kszucs: Removed.

ibis/tests/expr/test_table.py
kszucs: We don't eagerly rewrite anymore because that would break column dereferencing entirely. So I'm going to remove this test case. We can do the same using the rewrites BTW.

ibis/expr/types/generic.py
```diff
@@ -1183,21 +1187,24 @@ def as_table(self) -> ir.Table:
...
         True
         """
-        from ibis.expr.analysis import find_immediate_parent_tables
-
-        roots = find_immediate_parent_tables(self.op())
-        if len(roots) > 1:
+        parents = self.op().relations
+        values = {self.get_name(): self}
+
+        if len(parents) == 0:
```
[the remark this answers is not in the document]
cpcloud: This comment is not a blocker btw, so I'm going to resolve it.

ibis/expr/types/joins.py
```diff
@@ -0,0 +1,191 @@
...
+import ibis
+
+
+def disambiguate_fields(how, left_fields, right_fields, lname, rname):
+    if how in ("semi", "anti"):
+        # discard the right fields
+        return left_fields
+
+    lname = lname or "{name}"
+    rname = rname or "{name}"
```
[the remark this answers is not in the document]
cpcloud: Also not a blocker, resolving.

ibis/tests/expr/test_analysis.py
```diff
@@ -5,39 +5,10 @@
...
 import ibis.common.exceptions as com
 import ibis.expr.operations as ops
-from ibis.tests.util import assert_equal
 
 # Place to collect esoteric expression analysis bugs and tests
 
 
-# TODO(kszucs): not directly using an analysis function anymore, move to a
-# more appropriate test module
-def test_rewrite_join_projection_without_other_ops(con):
```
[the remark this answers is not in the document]
kszucs: I am not sure what it was testing but I restored the test case and updated according to the new join modeling.

ibis/expr/types/core.py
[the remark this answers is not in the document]
kszucs: Moved, and also removed the flag which was a remnant from the times where I tried to apply eager rewrites.

ibis/expr/types/relations.py
kszucs: Ohhh, this is from the pre-experimentation phase. Going to restore the `None` support but keep the new default. BTW `None` is a valid value, it gets coerced to a null literal.

cpcloud: >  BTW None is a valid value, it gets coerced to a null literal.

That's a breaking change I think, because previously it was being used as a sentinel to mean "no grouping keys"

ibis/expr/types/relations.py
cpcloud: What I mean here is just to add a slightly more informative comment about why this is the case.

ibis/expr/types/relations.py
cpcloud: I'll throw out a suggested replacement:

> `by` is an argument that comes from pandas, which for pandas was a convenient way to perform a standard join before the asof join, so we implement there here for consistency across backends.

ibis/expr/types/relations.py
kszucs: Restored, added a test too.

ibis/expr/types/relations.py
kszucs: Added.

ibis/tests/expr/test_table.py
kszucs: The API is less restrictive now, if we are okay with that then no, otherwise yes.

ibis/expr/types/joins.py
```diff
@@ -0,0 +1,191 @@
...
+            left.to_expr(),
+            right.to_expr(),
+            predicates,
+            deref_left=deref_left,
+            deref_right=deref_right,
+            deref_both=deref_both,
+        )
+        preds = flatten_predicates(list(preds))
+
+        # calculate the fields based in lname and rname, this should be a best
```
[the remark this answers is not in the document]
kszucs: Ooops, probably lost somewhere down the road. Updating.

ibis/tests/expr/test_table.py
cpcloud: Seems fine to me!

ibis/tests/expr/test_table.py
kszucs: I would prefer to defer it to a follow-up to unblock the backend PRs.

ibis/expr/types/joins.py
kszucs: I could use some suggestions here. Actually we should override `__getattribute__` to ensure that `self.finish()` gets called in all methods. Any suggestions for that?

gforsyth (commented):
small things.

I think all of our future selves would benefit from writing up the join-chain-deferencing ideas and putting them somewhere in the developer docs.
Maybe under a "Design" or "Internals" section?

Non-blocking, and a lot of what would go in that doc is already present in docstrings here, but I think it would be good to collect it all and then add some more detail.

ibis/expr/types/joins.py
kszucs: Done.

ibis/expr/types/generic.py
```diff
@@ -6,11 +6,14 @@
...
 import ibis
+from ibis.expr.rewrites import rewrite_window_input
 import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
-from ibis.common.deferred import Deferred
+from ibis.common.deferred import Deferred, _
+import ibis.expr.builders as bl
 import ibis.expr.operations as ops
 from ibis.common.grounds import Singleton
+from ibis.common.deferred import deferrable
```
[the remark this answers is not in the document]
kszucs: Isn't ruff supposed to reorder these?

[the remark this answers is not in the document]
kszucs: Updated.

ibis/expr/operations/relations.py
```diff
@@ -1,671 +1,457 @@
...
+    shape = ds.columnar
 
-    def __init__(self, schema, name, namespace) -> None:
-        if name is None:
-            name = genname()
-        super().__init__(schema=schema, name=name, namespace=namespace)
+    def __init__(self, rel, **kwargs):
+        if len(rel.schema) != 1:
+            raise IntegrityError(
+                f"Subquery must have exactly one column, got {len(self.rel.schema)}"
```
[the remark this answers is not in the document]
kszucs: Fixed and added test case.

ibis/expr/types/relations.py
```diff
@@ -2966,13 +2966,16 @@ def join(
         # must be wrapped in SelfReferences so that we can join the same table
         # with itself multiple times and to enable optimization passes later on
         left = left.op()
-        if isinstance(left, ops.SelfReference):
-            left = left.parent
+        if isinstance(left, ops.JoinChain):
+            # if the left side is already a join chain, we can reuse it
+            expr = left.to_expr()
```
cpcloud, line 2983: How does this occur coming from a user-constructed expression? How can the left side be a join chain and the next call ends up here?

ibis/tests/expr/test_selectors.py
```diff
@@ -383,7 +383,9 @@ def test_all(penguins):
         param(~s.r[3, 4, 5], sorted(set(range(8)) - {3, 4, 5}), id="neg_int_tuple"),
         param(s.r["island", "year"], ("island", "year"), id="string_tuple"),
         param(s.r[["island", "year"]], ("island", "year"), id="string_list"),
-        param(iter(["island", 4, "year"]), ("island", 4, "year"), id="mixed_iterable"),
+        # bind() doesn't work with iterables but only with sequences, we should
+        # rethink the supported input types
+        # param(iter(["island", 4, "year"]), ("island", 4, "year"), id="mixed_iterable"),
```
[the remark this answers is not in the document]
kszucs: Added support for it in bind()

ibis/expr/types/relations.py
```diff
@@ -155,10 +155,15 @@ def unwrap_aliases(values: Iterator[ir.Value]) -> Mapping[str, ir.Value]:
...
-    for parent in util.promote_list(parents):
+    parents = util.promote_list(parents)
+    for parent in parents:
         for k, v in parent.values.items():
             if isinstance(v, ops.Field):
-                while isinstance(v, ops.Field):
+                # track down the field in the hierarchy until no modification
+                # is made so only follow ops.Field nodes not arbitrary values;
+                # also stop tracking if the field belongs to a parent which
+                # we want to dereference to
```
cpcloud, line 167: What's an example of this using the user API?

ibis/expr/tests/test_newrels.py
```diff
@@ -1038,11 +1038,148 @@ def test_join_chain_gets_reused_and_continued_after_a_select():
     assert abc._finish().op() == expected
 
 
-def test_join_name_collisions():
+def test_self_join_extensive():
     a = ibis.table(name="a", schema={"a": "int64", "b": "string"})
 
     aa = a.join(a, [a.a == a.a])
+    aa_ = a.join(a, "a")
+    aa__ = a.join(a, [("a", "a")])
```
cpcloud, line 1054: Since there's a lot of the same letter floating around, it would really help readability IMO if you added a couple comments stating what each new `aaaaaa` variable is meant to test.

cpcloud (commented):
Lookin' good!

Can you make sure to capture the collision implementation in the PR description? It could even go in a follow up IMO.

ibis/expr/types/generic.py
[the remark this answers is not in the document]
gforsyth: It can, but `ruff format` doesn't do that, you need to do something like `ruff check . --select I001 --select I002 --fix`

ibis/tests/expr/test_table.py
kszucs: Created a follow-up issue for this https://github.com/ibis-project/ibis/issues/7819

ibis/expr/types/joins.py
kszucs: Fixed.

ibis/expr/tests/test_newrels.py
kszucs: These are just API variant of the same call. Anyway, I would rather defer the rewrite of tests to unblock the follow-up work.

cpcloud (approved):

cpcloud:
🎉 🎉 🚀 🤖

PR #7773, opened 2023-12-15
refactor(formats): move the `TableProxy` object to formats from the operations

kszucs:
The `TableProxy` classes were misplaced under `ibis.expr.operations.relations`, probably because the `formats` arrived later. This change moves the types to the respective `ibis.formats.pandas` and `ibis.formats.pyarrow` modules as well as use `ibis.util.PseudoHashable` to make them well, "pseudo" hashable.


cpcloud (approved):
Seems legit!

PR #7777, opened 2023-12-15
feat(flink): add `register()` and `read_***()` for Flink backend

mfatihaktas:
Adds the following functions in `ibis/backends/flink/__init__.py`:
- `register()`
- `read_file()`
- `read_parquet()`
- `read_csv()`
- `read_json()`

Addition of these functions clears several test functions in `test_param.py` and `test_register.py`.


cpcloud:
Let's avoid adding `register` for now. It's a bit crufty and from a time when we thought we might be able to unify all the `read_*` methods under a single `read(...)` function.

mfatihaktas:
> Let's avoid adding `register` for now. It's a bit crufty and from a time when we thought we might be able to unify all the `read_*` methods under a single `read(...)` function.

Thanks for the note, removed `register()`.

ibis/backends/flink/__init__.py
```diff
@@ -635,6 +638,111 @@ def drop_view(
...
+        table_name
+            An optional name to use for the created table. This defaults to
+            a sequentially generated name.
+
+        Returns
+        -------
+        ir.Table
+            The just-registered table
+        """
+        obj = self._get_dataframe_from_path(path)
```
jcrist, line 688: Does flink support natively loading directly from files? If so, I think we should use that.

If it doesn't, I'm not sure how I feel about automatically loading files using pandas and forwarding them that way. We do something similar for local in-memory backends like duckdb/pandas where file paths are unambiguously local. But for a potentially distributed system like flink, automatically using a local file reader may have unexpected behavior with path names, and also may be inefficient.

If we do decide to handle file reading manually using an in-memory reader here, we should use the ones provided by `pyarrow` directly, not pandas. These have builtin support for directory datasets, and better match the behavior of other backends.

ibis/backends/flink/__init__.py
```diff
@@ -635,6 +638,111 @@ def drop_view(
...
+            elif path_str.endswith("json"):
+                dataframe = pd.read_json(file_path, lines=True)
+            else:
+                raise ValueError(f"Unsupported file_path: {file_path}")
+
+            dataframe_list.append(dataframe)
+
+        return pd.concat(dataframe_list, ignore_index=True, sort=False)
+
+    def read_file(
```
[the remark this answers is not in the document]
mfatihaktas: Thanks, makes sense. Changed it to `_read_file`.

ibis/backends/flink/__init__.py
mfatihaktas: Thanks, I agree that reading the file is not ideal while creating the table. Flink actually supports creating tables with [filesystem connector](https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/table/filesystem/). This however requires specifying the schema, as `create_table()` also requires:
```python
def create_table(
        self,
        name: str,
        *,
        database: str | None = None,
        catalog: str | None = None,
        temp: bool = False,
        overwrite: bool = False,
    ) -> ir.Table:
        ...

                   raise exc.IbisError("`schema` or `obj` is required")
    ...
```

So I think we have two options:
1. Add a required argument `schema` for `read_***()`. This would deviate from the interface existing for other backends.
2. "Read" the schema from the file with `pyarrow` and feed it into `create_table()` with `filesystem` connector. This has the same issue you raised in accessing the files in a distributed system.

Which one do you think makes more sense? We could also implement both where the user can specify the schema, and if not specified we could construct the schema from the file and raise an error in case of a failure in accessing the file.

cpcloud:
@mfatihaktas Apologies again for the churn, can you reopen this PR against `main`?

mfatihaktas:
> @mfatihaktas Apologies again for the churn, can you reopen this PR against `main`?

Reopened: https://github.com/ibis-project/ibis/pull/7908

PR #7784, opened 2023-12-16
docs(perf): use an unordered list instead of an ordered one

cpcloud:
Use a bulleted list instead of numbered to help quarto render


cpcloud:
image

Looks good, merging!

PR #7786, opened 2023-12-16
fix(duckdb): allow None inputs to UDFs

cpcloud:
This PR fixes an issue in the duckdb backend where UDFs were using their input
*values* instead of their input types (from the UDF signature) to define a UDF with DuckDB's `create_function`.


gforsyth (approved):
:shipit:

PR #7788, opened 2023-12-16
docs: fix failing generic expression docstrings

### LINKED ISSUE #7787: docs: https://ibis-project.org/reference/expression-generic#examples-36
### Please describe the issue

There is an error:
```
IbisError: Expression contains unbound tables and therefore cannot be executed. Use ibis.<backend>.execute(expr) or assign a backend instance to `ibis.options.default_backend`.
```
I reproduce this. What about this one?

### Code of Conduct

cpcloud:
Fix `ibis.param`, `Value.name` and `ibis.table` APIs example docstrings, all of which were failing to render.

~I'm not entirely sure why these are failing given the `add_ibis` fixture's behavior, which should disable interactive mode before every doctest.~

Closes #7787.


cpcloud:
Ah, the `add_ibis` fixture isn't running when rendering docstrings using quarto!

Issue #7631, opened 2023-11-28
bug: regression with `percent_rank` compilation in 7.x (Deferred related?)


tswast:
### What happened?

`percent_rank` is working fine in 6.x, but there seems to have been a reversion in 7.x. Here's the working version in 6.1.0:

```python
import ibis
import google.cloud.bigquery

ibis.__version__

Out[2]: '6.1.0'

bq = ibis.bigquery.connect(project_id="swast-scratch")
bqclient = google.cloud.bigquery.Client(project="swast-scratch")

bqclient.load_table_from_json(
    [{"col": 1}, {"col": 2}, {"col": 3}, {"col": 4}, {"col": 5}],
    "swast-scratch.my_dataset.small_ints",
    job_config=google.cloud.bigquery.LoadJobConfig(write_disposition="WRITE_TRUNCATE"),
).result()

Out[8]: LoadJob<project=swast-scratch, location=US, id=09379d8f-d04c-4f8b-9acd-2d3b1655c914>

table = bq.table("swast-scratch.my_dataset.small_ints")
table["col"].percent_rank().over(ibis.window(group_by=table["col"].notnull()))


Out[14]: 
0    0.00
1    0.25
2    0.50
3    0.75
4    1.00
Name: PercentRank(col), dtype: float64

table["col"].percent_rank().over(ibis.window(group_by=table["col"].notnull())).compile()

Out[17]: 'SELECT percent_rank() OVER (PARTITION BY t0.`col` IS NOT NULL ORDER BY t0.`col` ASC) AS `PercentRank_col`\nFROM `swast-scratch.my_dataset.small_ints` t0
```

### What version of ibis are you using?

7.1.0

### What backend(s) are you using, if any?

BigQuery

### Relevant log output

```sh
In [1]: import ibis
   ...: import google.cloud.bigquery
   ...: 
   ...: ibis.__version__
   ...: 
Out[1]: '7.1.0'

In [2]: 
   ...: bq = ibis.bigquery.connect(project_id="swast-scratch")
   ...: bqclient = google.cloud.bigquery.Client(project="swast-scratch")
   ...: 
   ...: bqclient.load_table_from_json(
   ...:     [{"col": 1}, {"col": 2}, {"col": 3}, {"col": 4}, {"col": 5}],
   ...:     "swast-scratch.my_dataset.small_ints",
   ...:     job_config=google.cloud.bigquery.LoadJobConfig(write_disposition="WRITE_TRUNCATE"),
[...]

PR #7790, opened 2023-12-17
fix(ir): merge window frames for bound analytic window functions with a subsequent over call

kszucs:
Fixes https://github.com/ibis-project/ibis/issues/7631

https://github.com/ibis-project/ibis/pull/7327 removed the argument of the relevant analytic functions and changed the value methods to return with a `WindowFunction` with a frame ordered by the value. When `.over()` is called on a window function the window frames were not merged, while we had that functionality it was only used by the `table.group_by().select(reduction | analytic)` API calls (the `GroupedTable` helper object). 

This PR changes the `.over()` method to merge the window builder provided in the call with the window already existing in the expression if any. It also deduplicates the `order_by` and `group_by` expressions during merge now.


ibis/tests/expr/test_window_functions.py
```diff
@@ -48,10 +50,24 @@ def test_value_over_api(alltypes):
     w1 = ibis.window(rows=(0, 1), group_by=t.g, order_by=[t.f, t.h])
     w2 = ibis.window(range=(-1, 1), group_by=[t.g, t.a], order_by=[t.f])
 
-    expr = t.f.cumsum().over(rows=(0, 1), group_by=t.g, order_by=[t.f, t.h])
```
kszucs, line 51: I think we have been misinterpreting this scenario because we used the `cumulative_window()` frame constructed in the `cumsum()` call, but we also provided explicit rows window boundaries. I changed the behaviour to raise due to window frame boundary conflict.

cpcloud: Yeah, that makes sense to me.

ibis/tests/expr/test_window_functions.py
```diff
@@ -48,10 +50,24 @@ def test_value_over_api(alltypes):
...
-    expected = t.f.cumsum().over(w2)
+    expr = t.f.sum().over(range=(-1, 1), group_by=[t.g, t.a], order_by=[t.f])
+    expected = t.f.sum().over(w2)
     assert expr.equals(expected)
+
+
+def test_conflicting_window_boundaries(alltypes):
+    t = alltypes
+
+    with pytest.raises(ValueError):
```
cpcloud, line 65:
```suggestion
    with pytest.raises(ValueError, match="Unable to merge windows"):
```

So that we know the right `ValueError` is raised.

cpcloud (changes requested):
Small change request, but otherwise LGTM!

ibis/tests/expr/test_window_functions.py
kszucs: Shall we raise expression error instead?

cpcloud: Yeah, that would be easier to test and likely more clear that it's coming from Ibis

kszucs: Fixed.

ibis/tests/expr/test_window_functions.py
```diff
@@ -48,10 +51,24 @@ def test_value_over_api(alltypes):
...
+    t = alltypes
+
+    with pytest.raises(ExpressionError, match="Unable to merge windows"):
+        t.f.cumsum().over(rows=(0, 1))
+
+
+def test_rank_followed_by_over_call_merge_frames(alltypes):
+    t = alltypes
+    expr1 = t.f.percent_rank().over(ibis.window(group_by=t.f.notnull()))
+    expr2 = ibis.percent_rank().over(group_by=t.f.notnull(), order_by=t.f).resolve(t)
```
cpcloud, line 73: Doesn't calling `resolve` here mean the APIs are not equivlent?

kszucs: They are not equivalent. `t.f.percent_rank()` returns with an expression backed by an `ops.WindowFunction` whereas `ibis.percent_rank()` returns with a deferred since there is nothing to bind to.

cpcloud: I suspect we should do that, what do you think?

kszucs: In a follow-up?

cpcloud: SGTM.

cpcloud (approved):

cpcloud:
Might need to regen some code in the cloud backends, lemme see.

Issue #7775, opened 2023-12-15
bug: TypeError: unsupported operand type(s) for |: 'BooleanColumn' and 'NullScalar', regression in ibis 7.x


tswast:
### What happened?

Previously (ibis 6.x), we could apply &, |, and other binary operators with null literal values. 

```
import ibis
ibis.literal(True) & ibis.null()
# Out[1]: And(True, None): True & None
```

Now, this causes an error:

```
import ibis
ibis.literal(True) & ibis.null()
# TypeError: unsupported operand type(s) for &: 'BooleanScalar' and 'NullScalar'
```

### What version of ibis are you using?

7.1.0

### What backend(s) are you using, if any?

N/A

### Relevant log output

```sh
Python 3.10.9 (main, Dec 14 2022, 20:14:08) [GCC 12.2.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.15.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import ibis
   ...: ibis.literal(True) & ibis.null()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 2
      1 import ibis
----> 2 ibis.literal(True) & ibis.null()

TypeError: unsupported operand type(s) for &: 'BooleanScalar' and 'NullScalar'
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
@kszucs This was broken in c20ba7f, can you look into it?

cpcloud:
The issue is that `GenericInstanceOf` is too strict for the `Value` pattern:

```
In [1]: import ibis.expr.operations as ops

In [2]: ops.And(ops.Literal(None, dtype='null'), ops.Literal(True, dtype='boolean'))
...
SignatureValidationError: And(<ibis.expr.operations.generic.Literal object at 0x7f25727ee340>, <ibis.expr.operations.generic.Literal object at 0x7f257937e7f0>) has failed due to the following errors:
  `left`: <ibis.expr.operations.generic.Literal object at 0x7f25727ee340> is not coercible to a Value[Boolean, DataShape]

Expected signature: And(left: Value[Boolean, DataShape], right: Value[Boolean, DataShape])
```

[...]

PR #7791, opened 2023-12-17
fix(ir): implicitly convert `None` literals with `dt.Null` type to the requested type during value coercion

kszucs:
Resolves https://github.com/ibis-project/ibis/issues/7775

### DIFF
--- a/ibis/expr/operations/generic.py
+++ b/ibis/expr/operations/generic.py
@@ -195,6 +195,9 @@ class Literal(Scalar[T]):
         return repr(self.value)
 
 
+NULL = Literal(None, dt.null)
+
+
 @public
 class ScalarParameter(Scalar, Named):
     _counter = itertools.count()
@@ -313,3 +316,6 @@ class SearchedCase(Value):
     def dtype(self):
         exprs = [*self.results, self.default]
         return rlz.highest_precedence_dtype(exprs)
+
+
+public(NULL=NULL)
--- a/ibis/expr/tests/test_api.py
+++ b/ibis/expr/tests/test_api.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import operator
 from datetime import datetime
 
 import pandas as pd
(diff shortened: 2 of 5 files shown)


ibis/expr/tests/test_api.py
```diff
@@ -124,3 +125,16 @@ def test_duplicate_columns_in_memtable_not_allowed():
...
     with pytest.raises(IbisInputError, match="Duplicate column names"):
         ibis.memtable(df)
+
+
+def test_implicit_coercion_of_null_literal():
+    # GH #7775
+    expr1 = ibis.literal(True) & ibis.null()
+    expr2 = ibis.literal(True) & None
+
+    expected = ops.And(
```
cpcloud, line 135: How about a test for the operation from the issue title (Or), as well as Xor?

ibis/expr/operations/generic.py
```diff
@@ -195,6 +195,9 @@ def name(self):
         return repr(self.value)
 
 
+NULL = Literal(None, dt.null)
```
cpcloud, line 198: This is the definition of `ibis.NA`:

```
❯ rg 'def null\(' -A 10
ibis/expr/types/generic.py
1988:def null():
1989-    """Create a NULL/NA scalar."""
1990-    return literal(None)
```

Can we define `ibis.NA` as:

```
NA = ops.NULL  # or ops.generic.NULL?
```

to avoid duplicating the definition?

ibis/expr/tests/test_api.py
kszucs: Going to parametrize it

ibis/expr/operations/generic.py
kszucs: Yes, and we can also turn it into a singleton

kszucs: Didn't turn it into a singleton, because there isn't just a single NULL. We interpret `None` as `NULL` nonetheless the datatype.

ibis/expr/tests/test_api.py
kszucs: Parametrized with AND OR XOR

cpcloud (approved):
Thanks!

PR #7793, opened 2023-12-18
chore(dev-tools): use `ruff` to format code in docstrings

cpcloud:
After https://github.com/astral-sh/ruff/pull/8854 (released in ruff 0.1.8) we can now use ruff to format code in docstrings.


cpcloud:
Depends on #7792

PR #7434, opened 2023-10-25
[WIP] chore: move datafusion to sqlglot

mesejo:
This PR moves DataFusion to sqlglot. Kudos to @cpcloud for doing most of the heavy lifting.

PR #7796, opened 2023-12-18
refactor(sql): add a base `SQLGlot` backend for `DuckDB` and `ClickHouse`

kszucs:
Rewrite the DuckDB and ClickHouse backends using a new base `SQLGlotBackend` and associated compiler.


ibis/backends/base/sqlglot/compiler.py
```diff
@@ -0,0 +1,1300 @@
...
+            "inner": "inner",
+            "left": "outer",
+            "right": "outer",
+            "semi": "semi",
+            "anti": "anti",
+            "cross": "cross",
+            "outer": "outer",
+        }
+        if not predicates:
+            predicates += (True,)
```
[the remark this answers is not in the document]
kszucs: Perhaps. Is it optional now? I think mandatory predicates are preferable.

ibis/backends/tests/tpch/test_h01.py
```diff
@@ -30,7 +30,7 @@ def test_tpc_h01(lineitem):
         avg_qty=t.l_quantity.mean(),
         avg_price=t.l_extendedprice.mean(),
         avg_disc=t.l_discount.mean(),
-        count_order=t.count(),
+        count_order=lambda t: t.count(),
```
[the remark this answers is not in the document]
kszucs: Is this still failing? We can certainly fix this.

ibis/backends/base/sqlglot/compiler.py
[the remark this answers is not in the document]
cpcloud: Yeah, the API allows joins without a predicate, and defaults to `True` when that is the case.

I will see about handling the default at the API level and putting a `Length` bound on the operations.

ibis/backends/base/sqlglot/datatypes.py
```diff
@@ -277,11 +281,17 @@ def _from_ibis_Struct(cls, dtype: dt.Struct) -> sge.DataType:
 
     @classmethod
     def _from_ibis_Decimal(cls, dtype: dt.Decimal) -> sge.DataType:
+        if (precision := dtype.precision) is None:
+            precision = cls.default_decimal_precision
+
+        if (scale := dtype.scale) is None:
+            scale = cls.default_decimal_scale
```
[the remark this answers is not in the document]
cpcloud: Done in https://github.com/ibis-project/ibis/pull/7849

ibis/backends/base/sqlglot/__init__.py
```diff
@@ -1,97 +1,246 @@
...
+    name: ClassVar[str]
+
+    @classmethod
+    def has_operation(cls, operation: type[ops.Value]) -> bool:
+        # singledispatchmethod overrides `__get__` so we can't directly access
+        # the dispatcher
+        dispatcher = cls.compiler.visit_node.register.__self__.dispatcher
+        return dispatcher.dispatch(operation) is not dispatcher.dispatch(object)
+
+    def _transform(
```
kszucs, line 35: What does this method do/responsible for?

ibis/backends/base/sqlglot/__init__.py
```diff
@@ -1,97 +1,246 @@
...
+        if limit is not None:
+            table_expr = table_expr.limit(limit)
+
+        if params is None:
+            params = {}
+
+        sql = self.compiler.translate(table_expr.op(), params=params)
+        assert not isinstance(sql, sge.Subquery)
+
+        if isinstance(sql, sge.Table):
```
kszucs, line 84: When can this happen? When expr is a PhysicalTable?

ibis/backends/base/sqlglot/__init__.py
```diff
@@ -1,97 +1,246 @@
...
+            sql = sg.select(STAR).from_(sql)
+
+        assert not isinstance(sql, sge.Subquery)
+        return [self._transform(sql, table_expr)]
+
+    def compile(
+        self, expr: ir.Expr, limit: str | None = None, params=None, **kwargs: Any
+    ):
+        """Compile an Ibis expression to a ClickHouse SQL string."""
```
kszucs, line 93: The docstring seems to be off here.

ibis/backends/base/sqlglot/__init__.py
```diff
@@ -1,97 +1,246 @@
...
+        queries = self._to_sqlglot(expr, limit=limit, params=params, **kwargs)
```
kszucs, line 94: What is the advantage of having `._to_sqlglot()` separated from `.compile()`? Feels like having a `.compile()` should be sufficient.

kszucs: I mean we don't necessarily need to return with a string here, it could return with the sqlglot expression instead.

ibis/backends/base/sqlglot/__init__.py
```diff
@@ -1,97 +1,246 @@
...
+        """Execute an expression."""
+
+        self._run_pre_execute_hooks(expr)
+        table = expr.as_table()
+        sql = self.compile(table, limit=limit, **kwargs)
+
+        schema = table.schema()
+
+        with self._safe_raw_sql(sql) as cur:
+            result = self.fetch_from_cursor(cur, schema)
```
kszucs, line 228: `fetch_from_cursor` is not defined as an abstractmethod, it used to be defined in the base SQL  backend. BTW does this method need to be public othen than for backward compat?

ibis/backends/base/sqlglot/rewrites.py
```diff
@@ -127,5 +132,11 @@ def sqlize(node):
         | filter_to_select
         | sort_to_select
     )
+    if DEBUG:
+        print("--------- STEP 1 ---------")  # noqa: T201
```
kszucs, line 136: Please remove these print statements which I added :D

We should have a config option at some point to help us debug the rewrite phases.

ibis/backends/base/sqlglot/compiler.py
```diff
@@ -0,0 +1,1300 @@
...
+def add_one_to_nth_value_input(_, **__):
+    if isinstance(_.nth, ops.Literal):
+        nth = ops.Literal(_.nth.value + 1, dtype=_.nth.dtype)
+    else:
+        nth = ops.Add(_.nth, 1)
+    return _.copy(nth=nth)
+
+
+@public
+class SQLGlotCompiler(abc.ABC):
```
kszucs, line 141: Should this be `Concrete` instead? Something like:

cass SQLGlotCompiler(Concrete):

    @attribute
    def f(self):
        return FuncGen()

    @attribute
    def agg(self):
        return AggGen()
```

Comment (kszucs) [reply ibis/backends/base/sqlglot/compiler.py:141]: Just took a look at the base compiler and the clickhouse one and both seem to be stateless. Could we convert the compiler to be just a collection of classmethods instead?

Comment (cpcloud) [reply ibis/backends/base/sqlglot/compiler.py:141]: Perhaps, but I'd rather do it in a follow up because that is a lot of churn for the current code.

File: ibis/backends/base/sqlglot/__init__.py (hunk 4 above)
Comment (cpcloud) [reply ibis/backends/base/sqlglot/__init__.py:35]: It's implemented in the DuckDB backend to add the necessary `st_aswkb` calls to any geometry columns in the final select expression. This **must** be done as the last step, and cannot be done as an ibis rewrite, because we only want to rewrite the final projection that the user executes, not every projection with geometry columns.

File: ibis/backends/base/sqlglot/__init__.py (hunk 5 above)
Comment (cpcloud) [reply ibis/backends/base/sqlglot/__init__.py:84]: Yep.

File: ibis/backends/base/sqlglot/__init__.py (hunk 6 above)
Comment (cpcloud) [reply ibis/backends/base/sqlglot/__init__.py:93]: Thanks! Fixed.

File: ibis/backends/base/sqlglot/__init__.py (hunk 7 above)
Comment (cpcloud) [reply ibis/backends/base/sqlglot/__init__.py:94]: `_to_sqlglot` can return with a `list` of sqlglot expressions if there are DDL statements like `CREATE TEMPORARY FUNCTION`, which shows up in the BigQuery backend.

I can revert this change here though, and only add it when I port the bigquery backend.

File: ibis/backends/base/sqlglot/__init__.py (hunk 8 above)
Comment (cpcloud) [reply ibis/backends/base/sqlglot/__init__.py:228]: Made it private!

File: ibis/backends/base/sqlglot/rewrites.py (hunk 9 above)
Comment (cpcloud) [reply ibis/backends/base/sqlglot/rewrites.py:136]: Done!

File: ibis/backends/base/sqlglot/compiler.py (hunk 10 above)
Comment (kszucs) [reply ibis/backends/base/sqlglot/compiler.py:141]: SGTM

File: ibis/backends/tests/tpch/test_h01.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (cpcloud) [reply ibis/backends/tests/tpch/test_h01.py:33]: Yep, if I revert this change then the test fails with an integrity error.

File: ibis/backends/base/sqlglot/compiler.py (hunk 1 above)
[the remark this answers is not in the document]
Comment (cpcloud) [reply ibis/backends/base/sqlglot/compiler.py:928]: Handling this at the API level now.

File: ibis/backends/base/sqlglot/compiler.py [hunk 11]
```diff
@@ -0,0 +1,1300 @@
...
+        # call only to be used in the one place it would be needed: the
+        # ScalarParameter `translate_val` rule
+        params = {
+            # remove aliases from scalar parameters
+            param.op().replace(unwrap_scalar_parameter): value
+            for param, value in (params or {}).items()
+        }
+
+        op = op.replace(
+            replace_scalar_parameter(params)
```
kszucs, line 260: We should move this rule to rewrites as well, but I can do that in a follow-up.

ibis/backends/base/sqlglot/compiler.py
```diff
@@ -0,0 +1,1300 @@
...
+
+    @visit_node.register(ops.Negate)
+    def visit_Negate(self, op, *, arg):
+        return -paren(arg)
+
+    @visit_node.register(ops.Not)
+    def visit_Not(self, op, *, arg):
+        if isinstance(arg, sge.Filter):
+            return sge.Filter(
+                this=self._de_morgan_law(arg.this), expression=arg.expression
```
kszucs, line 431: Why is this transformation needed?

cpcloud: Removed!

ibis/backends/base/sqlglot/compiler.py
cpcloud: It's already there, no need to move it.

ibis/backends/base/sqlglot/__init__.py
cpcloud: Reverted this for duckdb and clickhouse. I'll remove the datafusion version of this method in #7434.

cpcloud: Going to address this one in a follow-up. Changing to return a sqlglot object from `compile` breaks a bunch of tests.

ibis/backends/clickhouse/tests/snapshots/test_select/test_array_join_in_subquery/out.sql
```diff
@@ -1,5 +1,5 @@
 SELECT
   t0.id IN (SELECT
     arrayJoin(t1.ids) AS ids
-  FROM way_view AS t1) AS "InColumn(id, ids)"
+  FROM way_view AS t1) AS ids
```
kszucs, line 4: Is this alias change desired?

cpcloud: Good question, I'll look into what's happening.

ibis/backends/clickhouse/tests/snapshots/test_select/test_join_self_reference/out.sql
```diff
@@ -1,5 +1,17 @@
 SELECT
-  t0.*
-FROM functional_alltypes AS t0
-INNER JOIN functional_alltypes AS t1
-  ON t0.id = t1.id
\ No newline at end of file
+  t1.id,
```
kszucs, line 2: Do we want to avoid explicitly listing the columns here? If yes, we can add a utility for that in a follow-up.

ibis/backends/clickhouse/tests/snapshots/test_select/test_named_from_filter_groupby/out1.sql
```diff
@@ -1,12 +1,17 @@
 SELECT
-  t0.key,
+  t1.key,
```
kszucs, line 2: Seems like we should have a SQL-like aggregate variant having the predicates encapsulated next to the aggregation groups and metrics, what do you think? I can add that in a follow-up if desired.

ibis/backends/clickhouse/tests/snapshots/test_select/test_array_join_in_subquery/out.sql
cpcloud: It looks like the operations' name changed to `ids` because of the `name` property introduced to the `Subquery` operation. I've changed that property to called `_name` since the base `name` implementation is used for naming expressions while the new `_name` is used for plucking out the single column used in an `InSubquery` operation.

```
```

ibis/backends/clickhouse/tests/snapshots/test_select/test_join_self_reference/out.sql
cpcloud: We definitely need the table name prefix here, but if there were a way we could generate `t1.*` that might be nice.

ibis/backends/clickhouse/tests/snapshots/test_select/test_array_join_in_subquery/out.sql
cpcloud: I actually just inlined the `_name` property since it's only used in `Subquery.value`.

ibis/backends/clickhouse/tests/snapshots/test_select/test_named_from_filter_groupby/out1.sql
cpcloud: Yeah, I think that could make sense. I wonder how difficult it would be to rolling that into `Select`. It might add a lot of complexity to do that. Either way, we can play around with the implementations.

kszucs: Created a follow-up https://github.com/ibis-project/ibis/issues/7851

ibis/backends/base/sqlglot/__init__.py
```diff
@@ -1,97 +1,246 @@
...
+            params = {}
+
+        sql = self.compiler.translate(table_expr.op(), params=params)
+        assert not isinstance(sql, sge.Subquery)
+
+        if isinstance(sql, sge.Table):
+            sql = sg.select(STAR).from_(sql)
+
+        assert not isinstance(sql, sge.Subquery)
+        return [self._transform(sql, table_expr)]
```
kszucs, line 88: Why is the list return type? Can there be more SQL queries here?

ibis/backends/base/sqlglot/__init__.py
kszucs: Could we rather override `_to_sqlglot()` in the duckdb compiler?

ibis/backends/clickhouse/tests/snapshots/test_select/test_join_self_reference/out.sql
kszucs: Created a follow-up: https://github.com/ibis-project/ibis/issues/7852

ibis/backends/duckdb/__init__.py
```diff
@@ -6,38 +6,43 @@
...
 import ibis.expr.types as ir
 from ibis import util
 from ibis.backends.base import CanCreateSchema
-from ibis.backends.base.sql.alchemy import AlchemyCrossSchemaBackend
-from ibis.backends.base.sql.alchemy.geospatial import geospatial_supported
-from ibis.backends.base.sqlglot import C, F
-from ibis.backends.duckdb.compiler import DuckDBSQLCompiler
-from ibis.backends.duckdb.datatypes import DuckDBType
+
+# from ibis.backends.base.sql.alchemy.geospatial import geospatial_supported
```
kszucs, line 30: This is a remnant of a rebase, should be removed.

ibis/backends/duckdb/tests/conftest.py
```diff
@@ -48,7 +49,7 @@ def preload(self):
 
     @property
     def ddl_script(self) -> Iterator[str]:
-        from ibis.backends.base.sql.alchemy.geospatial import geospatial_supported
+        # from ibis.backends.base.sql.alchemy.geospatial import geospatial_supported
```
kszucs, line 52: This should be removed.

cpcloud: Yep, done!

ibis/backends/duckdb/__init__.py
cpcloud: Thanks, done!

kszucs (commented):
+1

Great refactor!

PR #7797, opened 2023-12-18
refactor(pandas): port the pandas backend with an improved execution model

kszucs:
## Old Implementation

Since we need to reimplement/port all of the backends for #7752, I took an 
attempt at reimplementing the pandas backend using a new execution engine. 
Previously the pandas backend was implemented using a top-down execution model 
and each operation was executing using a multidispatched function. While it
served us well for a long time, it had a few drawbacks:
- it was often hard to understand what was going on due to the complex 
  preparation steps and various execution hooks
- the multidispatched functions were hard to debug, additionally they supported
  a wide variety of inputs making the implementation rather bulky
- due to the previous reaon, several inputs combinations were not supported, 
  e.g. value operations with multiple columnar inputs
- the `Scope` object was used to pass around the execution context which was
  created for each operation separately and the results were not reusable even 
  though the same operation was executed multiple times 

## New Implementation

The new execution model has changed in several ways:
- there is a rewrite layer before execution which lowers the input expression
  to a form closer to the pandas execution model, this makes it much easier to
  implement the operations and also makes the input "plan" inspectable
- the execution is now topologically sorted and executed in a bottom-up manner;
  the intermediate results are reused, making the execution more efficient while
  also aggressively cleaned up as soon as they are not needed anymore to reduce
  the memory usage
- the execute function is now single-dispatched making the implementation
  easier to locate and debug
- the inputs now broadcasted to columnar shape so that the same implementation
  can be used for multiple input shape combinations, this removes several 
  special cases from the implementation in exchange of a negligible performance
  overhead
- there are helper utilities making it easier to implement compute kernels for
  the various value operations: `rowwise`, `columnwise`, `elementwise`, 
  `serieswise`; if there are multiple implementations available for a given
  operation, the most efficient one is selected based on the input shapes

The new backend implementation has a higher feature coverage while the 
implementation is one third of the size of the previous one. 

BREAKING CHANGE: the `timecontext` feature is not supported anymore


ibis/backends/pandas/tests/test_core.py
```diff
@@ -50,52 +47,22 @@ def test_from_dataframe(dataframe, ibis_table, core_client):
...
-    right = left.view()
-    join = left.join(right, "plain_strings")[left.plain_int64]
-    result = join.execute()
-    assert result is not None
-    assert not result.empty
-    assert count[0] == 1
+    con = ibis.pandas.connect()
+    with pytest.raises(com.OperationNotDefinedError):
+        con.execute(expr)
+    # TODO(kszucs): it should raise an unbound expr error
```
cpcloud, line 63: Does this need to be addressed?

ibis/backends/pandas/executor/analytic.py
```diff
@@ -0,0 +1,106 @@
...
+    def agg(df, order_keys):
+        return pd.Series(np.arange(len(df)), index=df.index)
+
+    return agg
+
+
+@execute.register(ops.FirstValue)
+@execute.register(ops.LastValue)
+def execute_first_value(op, arg):
+    i = 0 if isinstance(op, ops.FirstValue) else -1
```
cpcloud, line 92:
```suggestion
    i = -1 * isinstance(op, ops.FirstValue)
```

if you're feeling cheeky :)

ibis/backends/pandas/executor/analytic.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def agg(df, order_keys):
+        return df[arg.name].iloc[i]
```
cpcloud, line 95:
```suggestion
        return df[arg.name].iat[i]
```

Supposedly `iat` is faster than `iloc` due to only handling scalar integer inputs (no arrays/sequences are allow)

ibis/backends/pandas/executor/core.py
```diff
@@ -0,0 +1,40 @@
...
+from ibis.common.exceptions import OperationNotDefinedError
+from ibis.formats.pandas import PandasData
+
+
+@singledispatch
+def execute(node, **kwargs):
+    raise OperationNotDefinedError(f"no rule for {type(node)}")
+
+
+def zuper(node, params):
```
cpcloud, line 16: Heh, this looks a lot like the SQL backend translation :)

ibis/backends/pandas/executor/generic.py
```diff
@@ -0,0 +1,319 @@
...
+from ibis.backends.pandas.executor.utils import asframe, columnwise
+from ibis.common.exceptions import OperationNotDefinedError
+
+
+@execute.register(ops.Literal)
+def execute_literal(op, value, dtype):
+    if dtype.is_interval():
+        value = pd.Timedelta(value, dtype.unit.short)
+    elif dtype.is_array():
+        value = np.array(value)
```
cpcloud, line 21: Shouldn't this be `list(value)`?

ibis/backends/pandas/executor/generic.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+def mapdecimal(func, s):
+    def wrapper(x):
+        try:
+            return func(x)
+        except decimal.InvalidOperation:
+            return decimal.Decimal("NaN")
+
+    return s.map(wrapper)
```
cpcloud, line 141:
```suggestion
    return s.map(wrapper, na_action="ignore"))
```

To ensure that NULL values are preserved and also to avoid wasting cycles on NULL handling.

ibis/backends/pandas/executor/nested.py
```diff
@@ -0,0 +1,181 @@
...
+
+import ibis.expr.operations as ops
+from ibis.backends.pandas.executor.core import execute
+from ibis.backends.pandas.executor.utils import elementwise, rowwise
+
+
+@execute.register(ops.ArrayLength)
+def execute_array_length(op, arg):
+    if isinstance(arg, pd.Series):
+        return arg.apply(len)
```
cpcloud, line 17: Why not `map`?

ibis/backends/pandas/executor/nested.py
```diff
@@ -0,0 +1,181 @@
...
+        return None
+    # list(...) to unpack iterable
+    return np.array(list(result), dtype="object")
+
+
+def safe_merge(left, right):
+    if left is None or left is pd.NA:
+        return None
+    elif right is None or right is pd.NA:
+        return None
```
cpcloud, line 114: Eventually we may want a decorator like

```python
@null_safe
def merge(left, right):
    return {**left, **right}


def null_safe(f):
    def wrapper(*args):
        if any(arg is None or arg is pd.NA):
            return None
        return f(*args)
    return wrapper
```

ibis/backends/pandas/executor/generic.py
```diff
@@ -0,0 +1,319 @@
...
+        return mapdecimal(lambda x: x.exp(), arg)
+    else:
+        return np.exp(arg)
+
+
+@execute.register(ops.Round)
+def execute_round(op, arg, digits):
+    if op.arg.dtype.is_decimal():
+        if digits is None:
+            return arg.map(round)
```
cpcloud, line 202: Anywhere you're using `Series.map` should probably be called with `na_action="ignore"`. It'll simplify the code.

ibis/backends/pandas/executor/nested.py
```diff
@@ -0,0 +1,181 @@
...
+        # try to deserialize the value -> return None if it's None
+        if (js := json.loads(value)) is None:
+            return None
+    except (json.JSONDecodeError, TypeError):
+        # if there's an error related to decoding or a type error return None
+        return None
+
+    try:
+        # try to extract the value as an array element or mapping key
+        return json.dumps(js[key])
```
cpcloud, line 170: Why `dumps` instead of leaving the value deserialized?

ibis/backends/pandas/executor/strings.py
```diff
@@ -0,0 +1,230 @@
...
+}
+
+_serieswise_functions = {
+    ops.Capitalize: lambda arg: arg.str.capitalize(),
+    ops.EndsWith: lambda arg, end: arg.str.endswith(end),
+    ops.Lowercase: lambda arg: arg.str.lower(),
+    ops.LPad: lambda arg, length, pad: arg.str.rjust(length, fillchar=pad),
+    ops.LStrip: lambda arg: arg.str.lstrip(),
+    # TODO(kszucs): while these are properly working, the pandas implementation
+    # doesn't seem to support POSIX patterns
```
cpcloud, line 97: This is why the pandas and dask backends depend on the third party `regex` module which does support posix classes.

ibis/backends/pandas/executor/windows.py
```diff
@@ -0,0 +1,173 @@
...
+                end_index = 0
+            elif end_index > length:
+                end_index = length
+
+        return (start_index, end_index)
+
+    def apply_analytic(self, func, **kwargs):
+        return self.parent.apply_analytic(func, **kwargs)
+
+    def apply_reduction(self, func, **kwargs):
```
cpcloud, line 77: There are interesting and more optimal ways to compute this, but we can leave that for a follow-up (or never!)

ibis/formats/pandas.py
```diff
@@ -115,8 +116,10 @@ def convert_table(cls, df, schema):
                 "schema column count does not match input data column count"
             )
 
-        for (name, series), dtype in zip(df.items(), schema.types):
-            df[name] = cls.convert_column(series, dtype)
+        columns = []
+        for (_, series), dtype in zip(df.items(), schema.types):
+            columns.append(cls.convert_column(series, dtype))
+        df = pd.concat(columns, axis=1)
```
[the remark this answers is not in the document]
kszucs: When the input df is a view (coming from a limit) it raises copy warnings because of mutating  a view.

.github/workflows/ibis-backends.yml
```diff
@@ -11,8 +11,8 @@ on:
       - ".envrc"
     branches:
       - master
-      - "*.x.x"
```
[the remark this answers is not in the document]
kszucs: Coming from a rebase where we both modified this file, restored.

ibis/backends/tests/test_interactive.py
```diff
@@ -33,6 +33,7 @@ def table(con):
     return con.table("functional_alltypes")
 
 
+@pytest.mark.notimpl(["pandas"])
```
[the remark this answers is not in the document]
kszucs: Correct, I think you added these tests in the SQL refactors.

ibis/common/dispatch.py
```diff
@@ -91,3 +92,52 @@ def call(arg, *args, **kwargs):
...
+                raise TypeError(f"Multiple attributes are defined with name {key}")
+
+        return type.__new__(cls, name, bases, namespace)
+
+    @classmethod
+    def __prepare__(cls, name, bases):
+        return _MultiDict()
+
+
+class Dispatched(metaclass=DispatchedMeta):
```
kszucs, line 142: Add docstrings

ibis/expr/operations/relations.py
```diff
@@ -112,8 +112,8 @@ def __init__(self, rel, **kwargs):
 
     @attribute
     def value(self):
-        name = self.rel.schema.names[0]
-        return self.rel.values[name]
+        (value,) = self.rel.values.values()
```
kszucs, line 115: Pull these changes out to another PR.

ibis/expr/types/joins.py
```diff
@@ -205,6 +219,79 @@ def join(
         # return with a new JoinExpr wrapping the new join chain
         return self.__class__(left, collisions=collisions)
 
+    def asof_join(
```
kszucs, line 222: Pull these changes out to another PR.

ibis/backends/pandas/executor/analytic.py
kszucs: It is more cryptic so rather not :)

ibis/backends/pandas/executor/analytic.py
kszucs: Updated.

ibis/backends/pandas/executor/core.py
kszucs: I moved this function to `Executor.execute()`

ibis/backends/pandas/executor/generic.py
kszucs: Updated everywhere.

ibis/backends/pandas/executor/nested.py
kszucs: Updated.

ibis/backends/pandas/executor/generic.py
kszucs: Updated.

ibis/backends/pandas/executor/nested.py
kszucs: Removed the dump call.

ibis/backends/pandas/executor/strings.py
kszucs: These could have been more performant implementations, but I removed them in favor of the already existing rowwise `regex` functions.

ibis/common/dispatch.py
kszucs: Pulled out to a separate PR which has been merged already.

ibis/expr/operations/relations.py
kszucs: Pulled out to a separate PR which has been merged already.

ibis/expr/types/joins.py
kszucs: Pulled out to #7859

ibis/backends/pandas/tests/test_join.py
```diff
@@ -281,35 +307,32 @@ def test_keyed_asof_join(
...
-        by="key",
-        tolerance=pd.Timedelta("2D"),
-    )
-    tm.assert_frame_equal(result[expected.columns], expected)
-    with pytest.raises(AssertionError):
-        tm.assert_series_equal(result["time"], result["time_right"])
-    with pytest.raises(AssertionError):
-        tm.assert_series_equal(result["key"], result["key_right"])
+# @merge_asof_minversion
+# def test_keyed_asof_join_with_tolerance(
```
kszucs, line 317: Tolerance support is not functioning properly

ibis/backends/pandas/tests/test_functions.py
```diff
@@ -238,8 +239,8 @@ def test_ifelse_returning_bool():
     true = ibis.literal(True)
     false = ibis.literal(False)
     expr = ibis.ifelse(one + one == two, true, false)
-    result = execute(expr.op())
-    assert result is True
+    result = ibis.pandas.connect().execute(expr)
+    assert result is True or result is np.True_
```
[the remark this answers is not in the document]
kszucs: We could normalize the output but that can lead to unexpected results for other backends. 
I suggest to solve this once we established an output specification for the pandas format.

ibis/expr/tests/test_newrels.py
```diff
@@ -1293,3 +1293,13 @@ def test_join_method_docstrings():
...
+
+
+def test_e():
+    t1 = ibis.table(name="t1", schema={"a": "int64", "b": "string"})
+    t2 = ibis.table(name="t2", schema={"a": "int64", "b": "string"})
+
+    joined = t1.left_join(t2, [t1.a == t2.a]).filter(t1.a < 5)
+
+    e = t1.left_join(joined, [t1.a == joined.a])
+    print(e.select(t1))
```
kszucs, line 1305: Add proper assertions and rename the test case.

ibis/backends/pandas/tests/test_join.py
kszucs: Fixed.

ibis/expr/tests/test_newrels.py
kszucs: Done.

ibis/backends/pandas/tests/test_core.py
kszucs: Fixed, also added support for execution unbound tables with the backend.

ibis/backends/pandas/tests/test_udf.py
```diff
@@ -364,26 +364,28 @@ def my_wm(v, w):
...
-    result = expr.execute().sort_values(["key", "a"])
-    expected = df.sort_values(["key", "a"]).assign(
-        rolled=lambda d: d.groupby("key")
-        .b.rolling(3, min_periods=1)
-        .apply(lambda x: x.mean(), raw=True)
-        .reset_index(level=0, drop=True)
-    )
-    tm.assert_frame_equal(result, expected)
+# TODO(kszucs): revisit this, duckdb produces the same result as the pandas
+# backend, but the expected result is different
```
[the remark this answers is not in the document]
kszucs: Created a follow-up https://github.com/ibis-project/ibis/issues/7893

ibis/backends/pandas/executor/generic.py
kszucs: It is working for now, we can revisit once the pandas out put format has clear guidelines about the result types.

cpcloud (approved):

PR #7798, opened 2023-12-18
WIP: see which filterwarnings are actually needed

NickCrews:
curious what tests fail in CI,
I'm prettu sure some of these lines can be deleted.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -276,72 +276,72 @@ norecursedirs = [
 filterwarnings = [
   # fail on any warnings that are not explicitly matched below
   "error",
-  # pyspark and impala leave sockets open
-  "ignore:Exception ignored in:",
-  # dask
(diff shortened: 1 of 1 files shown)


cpcloud:
Unfortunately most of our backends spew an oceanic level of warnings, of widely varying relevance for end users.

cpcloud:
If you realllllly want to go through the trouble of removing the "dead" warning ignores, I would recommend doing it locally per backend. GitHub Actions' UI isn't conducive to fast rendering of large numbers of test failures.

cpcloud:
It would also be fine to do it just for the backends you're able to easily run locally and leave the others as they are.

cpcloud:
Marking as draft so that folks don't yet pick it up for review!

cpcloud:
Let's revisit this once we finish porting the SQL backends over to sqlglot!

PR #7800, opened 2023-12-18
feat(flink): add primary key support

mfatihaktas:
Adds support for `primary keys` while creating tables in Flink. Adds a new argument `primary_keys` in `create_table()` following the design choices made for adding support for `watermark`.

### DIFF
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -312,6 +312,7 @@ class Backend(BaseBackend, CanCreateDatabase):
         catalog: str | None = None,
         tbl_properties: dict | None = None,
         watermark: Watermark | None = None,
+        primary_key: str | list[str] | None = None,
         temp: bool = False,
         overwrite: bool = False,
     ) -> ir.Table:
@@ -349,6 +350,11 @@ class Backend(BaseBackend, CanCreateDatabase):
             dictionary of key-value pairs (key1=val1, key2=val2, ...).
         watermark
             Watermark strategy for the table, only applicable on sources.
+        primary_key
+            A single column or a list of columns to be marked as primary. Raises
+            an error if the column(s) in `primary_key` is NOT a subset of the
+            columns in `schema`. Primary keys must be non-nullable in Flink and
+            the columns indicated as primary key will be designated as non-nullable.
         temp
             Whether a table is temporary or not.
         overwrite
@@ -441,6 +447,7 @@ class Backend(BaseBackend, CanCreateDatabase):
                 schema=schema,
                 tbl_properties=tbl_properties,
                 watermark=watermark,
+                primary_key=primary_key,
                 temporary=temp,
                 database=database,
                 catalog=catalog,
--- a/ibis/backends/flink/ddl.py
+++ b/ibis/backends/flink/ddl.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING
 
 import sqlglot as sg
 
+import ibis.common.exceptions as exc
 import ibis.expr.schema as sch
 from ibis.backends.base.sql.ddl import (
     CreateTable,
@@ -19,15 +20,19 @@ from ibis.backends.base.sql.ddl import (
 )
 from ibis.backends.base.sql.registry import quote_identifier
 from ibis.backends.flink.registry import type_to_sql_string
+from ibis.util import promote_list
 
 if TYPE_CHECKING:
+    from collections.abc import Sequence
+
     from ibis.api import Watermark
 
 
-def format_schema(schema):
+def format_schema(schema: sch.Schema):
     elements = [
         _format_schema_element(name, t) for name, t in zip(schema.names, schema.types)
     ]
+
     return "({})".format(",\n ".join(elements))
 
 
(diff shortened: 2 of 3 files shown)


ibis/backends/flink/ddl.py
```diff
@@ -54,15 +58,31 @@ def _format_watermark_strategy(watermark: Watermark) -> str:
 
 
 def format_schema_with_watermark(
-    schema: sch.Schema, watermark: Watermark | None = None
+    schema: sch.Schema,
+    watermark: Watermark | None = None,
+    primary_keys: set[str] | None = None,
```
deepyaman, line 63: I would say `primary_key`; there's really just one `primary_key`, but it can be compound.

Also, does it makes sense to also accept a `str`? I feel like at least `str | Iterable[str]` is more aligned with what other things that accept a key take (e.g. `group_by`).

ibis/backends/flink/ddl.py
```diff
@@ -54,15 +58,31 @@ def _format_watermark_strategy(watermark: Watermark) -> str:
...
+    if primary_keys is not None:
+        # Note (mehmet): Currently supports "NOT ENFORCED" only. For the reason
+        # of this choice, the following quote from Flink docs is self-explanatory:
+        # "SQL standard specifies that a constraint can either be ENFORCED or
+        # NOT ENFORCED. This controls if the constraint checks are performed on
+        # the incoming/outgoing data. Flink does not own the data therefore the
+        # only mode we want to support is the NOT ENFORCED mode. It is up to the
+        # user to ensure that the query enforces key integrity."
+        # Ref: https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/dev/table/sql/create/#primary-key
+        comma_separated_keys = ",".join(list(primary_keys))
```
deepyaman, line 83:
```suggestion
        comma_separated_keys = ",".join(primary_keys)
```

ibis/backends/flink/ddl.py
```diff
@@ -106,6 +127,15 @@ def __init__(
         self.catalog = catalog
         self.temporary = temporary
         self.watermark = watermark
+        self.primary_keys = primary_keys
+
+        # Check if `primary_keys` is a subset of the columns in `schema`.
+        if self.primary_keys and not self.primary_keys.issubset(set(schema.names)):
```
deepyaman, line 133:
```suggestion
        if self.primary_keys and not self.primary_keys <= set(schema.names):
```
Nit

ibis/backends/flink/tests/test_ddl.py
```diff
@@ -262,19 +262,38 @@ def test_force_recreate_in_mem_table(
         assert new_table.schema() == schema
 
 
-def test_create_source_table_with_watermark(
-    con, functional_alltypes_schema, temp_table, csv_source_configs
+@pytest.mark.parametrize("primary_keys", [None, {"id"}, {"id", "string_col"}])
+def test_create_source_table_with_watermark_and_primary_keys(
+    con, functional_alltypes_schema, temp_table, csv_source_configs, primary_keys
```
deepyaman, line 267: Is `functional_alltypes_schema` not needed here? What changed to specify it explicitly instead?

deepyaman (commented):
Unsolicited pass

ibis/backends/flink/ddl.py
mfatihaktas: Thanks. I changed the argument to `primary_key`, which is a `str` that contains a single column name or a comma-separated list of column names.

ibis/backends/flink/ddl.py
mfatihaktas: Thanks, done.

ibis/backends/flink/ddl.py
mfatihaktas: Thanks, done.

ibis/backends/flink/tests/test_ddl.py
mfatihaktas: > Is functional_alltypes_schema not needed here?

Yes, `functional_alltypes_schema` is not needed, removed it from the args.

> What changed to specify it explicitly instead?

Primary keys must be `non-nullable` in Flink:
> Flink will assume correctness of the primary key by assuming that the columns nullability is aligned with the columns in primary key. Connectors should ensure those are aligned.

For the following reason:
> Notes: In a CREATE TABLE statement, creating a primary key constraint will alter the columns nullability, that means, a column with primary key constraint is not nullable.
`new_table.schema()` returns nullable field types for primary-key columns, which leads to an assertion on this line:
```
assert new_table.schema() == schema
```
This is why I ended up defining a copy of `functional_alltypes_schema` with `"id": dt.int32(nullable=False)` and `"string_col": dt.string(nullable=False)`.

ibis/backends/flink/__init__.py
```diff
@@ -349,6 +350,12 @@ def create_table(
             dictionary of key-value pairs (key1=val1, key2=val2, ...).
         watermark
             Watermark strategy for the table, only applicable on sources.
+        primary_key
+            Column or comma-separated list of columns to be marked as PRIMARY.
```
jcrist, line 354: I *think* what we'd want here is a single column as a string, or multiple columns as a sequence of strings.

```python
con.create_table(..., primary_key="foo")
con.create_table(..., primary_key=["foo", "bar"])
```

Right now it looks like this always takes a string, which may happen to be comma delimited (e.g. `primary_key="foo,bar"`).

ibis/backends/flink/ddl.py
jcrist: I agree with @deepyaman here (see my comment above). This should be something like:

```suggestion
    primary_keys: str | Sequence[str] | None = None,
```

with consuming code below updated accordingly. You might find `ibis.util.promote_list` useful here.

jcrist (changes requested):
Besides the interface changes requested below, it looks like the added tests don't sufficiently cover the new code - would you mind expanding the test coverage a bit to hit more of the missed lines?

ibis/backends/flink/ddl.py
```diff
@@ -24,10 +25,13 @@
     from ibis.api import Watermark
 
 
-def format_schema(schema):
+def format_schema(
+    schema: sch.Schema,
```
chloeh13q, line 29: super small nit but we probably don't need this extra line

ibis/backends/flink/ddl.py
chloeh13q: list or set?

ibis/backends/flink/ddl.py
```diff
@@ -107,6 +128,15 @@ def __init__(
         self.temporary = temporary
         self.watermark = watermark
 
+        self.primary_keys = set(primary_key.split(",")) if primary_key else None
+        # Check if `primary_keys` is a subset of the columns in `schema`.
+        if self.primary_keys and not self.primary_keys <= set(schema.names):
+            raise exc.IbisError(
+                "`primary_keys` must be a subset of the columns in `schema`. \n"
```
chloeh13q, line 135: nit: `primary_keys` -> `primary_key` because that's the name of the argument you're using

mfatihaktas: Thanks, done.

ibis/backends/flink/__init__.py
mfatihaktas: Thanks, changed the function signature as you suggested.

mfatihaktas:
> Besides the interface changes requested below, it looks like the added tests don't sufficiently cover the new code - would you mind expanding the test coverage a bit to hit more of the missed lines?

@jcrist Thanks for the review. I expanded the test coverage. Will add more based on the test coverage report.

cpcloud:
@mfatihaktas Can you reopen this PR against `main`? Sorry about the churn!

mfatihaktas:
> @mfatihaktas Can you reopen this PR against `main`? Sorry about the churn!

Reopened: https://github.com/ibis-project/ibis/pull/7907

PR #7801, opened 2023-12-19
feat: add support for splitting a string using a regular expression

### LINKED ISSUE #7799: feat: String.re_split()

I have to split a string not on a constant char, but on a pattern.

An easy workaround I can now easily (🎉 yay thank you!) do for duckdb is

```python
@ibis.udf.scalar.builtin
def re_split(s: str, pattern: str) -> list[str]:
    ...
```

Or do [other various shenanigans](https://github.com/pola-rs/polars/issues/4819) but I figure that this could be common enough to make it builtin.

 If a backend doesn't implement this natively, then I think the expression could be rewritten as a less efficient  and hacky `s.re_replace(split_pattern, "a-very-unlikely-substring".split("a-very-unlikely-substring")`

bankends known to implement:
- duckdb
- postgres
- trino
- pandas
- dask

 known to not implement
- bigquery
- polars

NA

personally duckdb, but I understand ya gotta take the other backends in mind :)

### Code of Conduct

cpcloud:
Closes #7799.

### DIFF
--- a/ibis/backends/clickhouse/compiler/values.py
+++ b/ibis/backends/clickhouse/compiler/values.py
@@ -1050,3 +1050,8 @@ def _timestamp_range(op, *, start, stop, step, **_):
         func, F.range(0, F.timestampDiff(unit, start, stop), step_value)
     )
     return result
+
+
+@translate_val.register(ops.RegexSplit)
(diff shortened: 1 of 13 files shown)


gforsyth (approved):
Nice!

PR #7803, opened 2023-12-19
ci(flink): clean up docker image specifications

cpcloud:
We should be able to use a simpler image specification to make it easier to Mend Renovate to perform automatic upgrades to the image. We can also isolate the flink job manager a bit more by running the flink base image.


deepyaman (commented):
LGTM! :shipit:

PR #7810, opened 2023-12-20
feat(api): support median and quantile on more types

### LINKED ISSUE #7250: feat: DateValue.median() .quantile()

I need to do some summary stats on date columns. At least the very basic ones would be super useful!

Not exactly sure what backends have this builtin, and which we might have to do some hack of casting to an int, doing the op, and then casting back...

6.2.0

duckdb

### Code of Conduct

cpcloud:
Adds support for calling `median` and `quantile` on more types than numeric.

Tests are included for computing the median of dates + strings.

Closes #7250.

### DIFF
--- a/ibis/backends/duckdb/registry.py
+++ b/ibis/backends/duckdb/registry.py
@@ -489,8 +489,16 @@ operation_registry.update(
         ops.JSONGetItem: fixed_arity(_json_get_item, 2),
         ops.RowID: lambda *_: sa.literal_column("rowid"),
         ops.StringToTimestamp: _strptime,
-        ops.Quantile: reduction(sa.func.quantile_cont),
-        ops.MultiQuantile: reduction(sa.func.quantile_cont),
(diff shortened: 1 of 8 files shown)


cpcloud:
Ok, the commit linter is successfully not commenting on PRs with valid titles and descriptions.

I'll edit the PR title to be invalid to try the case where we want a comment to show up

cpcloud:
Last case to check is that the linter isn't spamming comments when additional edits don't fix the issue.

cpcloud:
Ok, looks like it's a bit spammy! Fixing it...

cpcloud:
Ok, I think the spam factory has been shutdown.

One more try

cpcloud:
I had one more edit, which was to actually fail the check if the message is bogus!

cpcloud:
Ok, looks like that's working. Lemme fix the title, and then we can try squash merging 😬

cpcloud:
Ok, squashing 😬

PR #7813, opened 2023-12-20
ci: move back to download artifact v4 until actions/download-artifact#249 is fixed

cpcloud:
Avoid annoying spurious failures in CI due to a broken v4 of the upload and download artifact actions.


gforsyth:
Ugh

```
 npm WARN exec The following package was not found and will be installed: commitlint@18.4.3
⧗   input: ci: move back to download artifact v4 until actions/download-artifact#249 is fixed
⚠   body must have leading blank line [body-leading-blank]
```

I think we can disable that warning

gforsyth:
> I think we can disable that warning

Although for now we can just add a blank line to the PR description body so we don't need to re-run CI

gforsyth (approved):
I think this is good to go and we can ignore the failing PR title lint

PR #7814, opened 2023-12-20
ci(dev-tools): stop warning on leading blank line in body

gforsyth:
Otherwise, we're stuck telling people to always start PR descriptions with a blank line which seems silly (and annoying).


cpcloud:
I think we need to change the `COMMIT_MSG: >` to `COMMIT_MSG: |` instead

cpcloud:
Fixed the `issue-number` problem in #7815.

PR #7817, opened 2023-12-20
fix(repr): default to pa.binary for all geospatial dtypes

gforsyth:
Does what it says on the tin. We will want to specialize this more once we have
proper support for `geoarrow` and friends, but for now, we should assume all
geospatial types should be rendered as binary.


cpcloud (approved):

ncclementi (approved):
Thanks @gforsyth LGTM!

PR #7818, opened 2023-12-20
fix(duckdb): allow table creation from expr with geospatial datatypes

gforsyth:
I thought, initially, that `WKB_BLOB` should be mapped on to something
like `dt.GeoSpatial`, but that leads to a situation where we try to call
`AsWKB` on an existing WKB column, and DuckDB does not like that.

Mapping it as `binary` works and the created columns are still correctly
detected as `geometry:geospatial`.

### DIFF
--- a/ibis/backends/duckdb/tests/test_geospatial.py
+++ b/ibis/backends/duckdb/tests/test_geospatial.py
@@ -219,7 +219,7 @@ def test_create_table_geospatial_types(geotable, con):
 
     # con = ibis.get_backend(geotable)
 
-    t = con.create_table(name, geotable)
+    t = con.create_table(name, geotable, temp=True)
 
     assert t.op().name in con.list_tables()
     assert any(map(methodcaller("is_geospatial"), t.schema().values()))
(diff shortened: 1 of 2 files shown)


ibis/backends/duckdb/tests/test_geospatial.py
```diff
@@ -210,3 +212,14 @@ def test_geospatial_convert(geotable, gdf):
...
         geo_ll.to_pandas(), gdf_ll, check_less_precise=True, check_crs=False
     )
+
+
+def test_create_table_geospatial_types(geotable, con):
+    name = ibis.util.gen_name("geotable")
+
+    # con = ibis.get_backend(geotable)
+
+    t = con.create_table(name, geotable)
```
cpcloud, line 222:
```suggestion
    t = con.create_table(name, geotable, temp=True)
```

Just so it gets dropped automatically when the test is finished.

cpcloud (approved):

cpcloud:
Oh right, we can click update now and let squash merging take care of the commits :)

PR #7831, opened 2023-12-21
refactor(ir): wrap `JoinChain.first` in `ops.SelfReference` similar to the rest of the join tables

kszucs:


ibis/expr/tests/test_newrels.py
```diff
@@ -1191,3 +1205,46 @@ def test_name_collisions_raise():
     abc = a.join(b, [a.a == b.a]).join(c, [a.a == c.a])
     with pytest.raises(IntegrityError):
         abc.filter(abc.a < 1)
+
+
+def test_self_view_join_followed_by_aggregate_correctly_dereference_fields():
```
kszucs, line 1210: This is the test case covering the issue, the other test changes were just required to conform with the new IR.

ibis/expr/tests/test_newrels.py
```diff
@@ -32,6 +35,19 @@
 )
 
 
+@contextlib.contextmanager
+def self_references(*tables):
+    old_counter = ops.SelfReference._uid_counter
+    # set a new counter with 1000 to avoid colliding with manually created
+    # self-references using t.view()
+    new_counter = itertools.count(1000)
```
cpcloud, line 43: Is 1000 significant in any way? Why not leave it out entirely?

ibis/expr/types/joins.py
```diff
@@ -95,7 +95,7 @@ def dereference_value(pred, deref_left, deref_right):
         return pred.replace(deref_both, filter=ops.Value)
 
 
-def prepare_predicates(left, right, predicates, deref_left, deref_right, deref_both):
+def prepare_predicates(left, right, predicates, deref_left, deref_right):
```
kszucs, line 98: This was an unused leftover.

ibis/expr/tests/test_newrels.py
kszucs: `t.view()` constructs a `SelfReference()` and this prevent colliding with it.

kszucs: If we start the counting again, then it is easy to accidentally generate identical references which makes the testing more error prone.

ibis/expr/tests/snapshots/test_format/test_memoize_filtered_tables_in_join/repr.txt
```diff
@@ -17,13 +17,15 @@ r2 := Filter[r1]
 r3 := Filter[r1]
   r1.kind == 'bar'
 
-r4 := SelfReference[r3]
+r4 := SelfReference[r2]
```
kszucs, line 20: These are rather ugly. We could just omit rendering self references and place a `!r2` or something prefixed like that in the `JoinChain` repr.

ibis/expr/tests/test_newrels.py
cpcloud: Ah, I misunderstood what the input argument to `count` does: it's the starting point of the counting. I thought it was the upper bound.

Carry on!

ibis/expr/tests/snapshots/test_format/test_memoize_filtered_tables_in_join/repr.txt
cpcloud: How about `&` if we're introducing new "syntax" in the repr?

ibis/expr/tests/test_newrels.py
cpcloud: Verified this by inspecting the IR repr, it looks good now!

```
(Pdb) pp expected_agg
r0 := UnboundTable: t
  a int64
  b int64
  f int64
  g string

r1 := Aggregate[r0]
  groups:
    g: r0.g
    a: r0.a
    b: r0.b
  metrics:
    total: Sum(r0.f)

r2 := SelfReference[r1]

r3 := SelfReference[r1]

r4 := JoinChain[r2]
  JoinLink[inner, r3]
    r2.a == r3.b
  values:
    g:           r2.g
    a:           r2.a
    b:           r2.b
    total:       r2.total
    g_right:     r3.g
    a_right:     r3.a
    b_right:     r3.b
    total_right: r3.total

Aggregate[r4]
  groups:
    g: r4.g
  metrics:
    metric: Max(r4.total - r4.total_right)
```

cpcloud (approved):

PR #7832, opened 2023-12-21
test(ir): cover constructing reductions in the core test suite

kszucs:


ibis/expr/types/generic.py
```diff
@@ -1355,15 +1355,16 @@ def __pandas_result__(self, df: pd.DataFrame) -> pd.Series:
         return PandasData.convert_column(df.loc[:, column], self.type())
 
     def _bind_reduction_filter(self, where):
+        rels = self.op().relations
```
kszucs, line 1358: We had an unnoticed bug here because the core test suite hasn't been testing this API. I added `test(ir)` scope for the commit, so we don't list a bugfix in the changelog since it is not a bug compared to the current release.

cpcloud: Oh, yeah 😅 I did fix this one locally!

cpcloud (approved):

Issue #7827, opened 2023-12-21
feat(impala): support dates


chris-park:
### What happened?

Unable to cast `timestamp` columns to `date` directly using the Impala backend. Here's a reproducible example:

```python

import ibis
ibis.options.interactive = True

df = ibis.memtable({
    'timestamp': ['2023-12-21 09:30:00']
})

to_date = lambda d: ibis.to_sql(df.timestamp.cast('date'), dialect=d)

# DuckDB 
to_date('duckdb')
# SELECT
#   CAST(t0.timestamp AS DATE) AS "Cast(timestamp, date)"
# FROM _ibis_pandas_memtable_y574b7whvjffrllsztk4jevkiu AS t0

# MySQL
to_date('mysql')
# SELECT
#   CAST(t0.timestamp AS DATE) AS `Cast(timestamp, date)`
# FROM (
#   SELECT
#     '2023-12-21 09:30:00' AS timestamp
# ) AS t0

# Impala
to_date('impala')
# KeyError: 'date'
# UnsupportedBackendType: date
#
# But both CAST(timestamp_column AS DATE) and TO_DATE(timestamp_column) is possible in Impala.
```

### What version of ibis are you using?

`8.0.0.dev321` on Windows 10

### What backend(s) are you using, if any?

Impala

### Relevant log output

```sh
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File .venv\lib\site-packages\ibis\backends\base\sql\registry\helpers.py:78, in type_to_sql_string(tval)
     77 try:
---> 78     return sql_type_names[name]
     79 except KeyError:

KeyError: 'date'

During handling of the above exception, another exception occurred:

UnsupportedBackendType                    Traceback (most recent call last)
Cell In[12], line 1
----> 1 to_date('impala')

Cell In[9], line 1
----> 1 to_date = lambda d: ibis.to_sql(df.timestamp.cast('date'), dialect=d)

File .venv\lib\site-packages\ibis\expr\sql.py:399, in to_sql(expr, dialect, **kwargs)
    396     else:
    397         read = write = getattr(backend, "_sqlglot_dialect", dialect)
--> 399 sql = backend._to_sql(expr, **kwargs)
    400 (pretty,) = sg.transpile(sql, read=read, write=write, pretty=True)
    401 return SQLString(pretty)

[...]

PR #7834, opened 2023-12-22
fix(impala): generate memtables using `UNION ALL` to work around sqlglot bug

cpcloud:
Work around [a bug in sqlglot](https://github.com/tobymao/sqlglot/issues/2717) where it cannot parse Impala `VALUES` syntax. Discovered by @chris-park in https://github.com/ibis-project/ibis/issues/7827#issuecomment-1867499217.


gforsyth (approved):
:shipit:

PR #7836, opened 2023-12-22
refactor(ir): introduce `JoinTable` operation unique to a `JoinChain` instead of using the globally unique `SelfReference`

kszucs:
This enables us to maintain join expression equality: `a.join(b).equals(a.join(b))`

So far we have been using `SelfReference` to make join tables unique, but it was globally unique which broke the equality check above. Therefore we need to restrict the uniqueness to the scope of the join chain. The simplest solution for that is to simply enumerate the join tables in the join chain, hence now all join participants must be `ops.JoinTable(rel, index)` instances. 

`ops.SelfReference` is still required to distinguish between two identical tables at the API level, but it is now decoupled from the join internal representation.

### DIFF
--- a/ibis/expr/format.py
+++ b/ibis/expr/format.py
@@ -162,7 +162,7 @@ def pretty(node):
 
     def mapper(op, _, **kwargs):
         result = fmt(op, **kwargs)
-        if isinstance(op, ops.Relation):
+        if isinstance(op, ops.Relation) and not isinstance(op, ops.JoinTable):
             tables[op] = result
             result = f"r{next(refcnt)}"
         return Rendered(result)
@@ -335,6 +335,11 @@ def _limit(op, parent, **kwargs):
 @fmt.register(ops.Distinct)
 def _self_reference(op, parent, **kwargs):
     return f"{op.__class__.__name__}[{parent}]"
+
+
+@fmt.register(ops.JoinTable)
+def _join_table(op, parent, index):
+    return parent
 
 
 @fmt.register(ops.Literal)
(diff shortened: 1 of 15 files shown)


ibis/expr/format.py
```diff
@@ -162,7 +162,7 @@ def pretty(node):
 
     def mapper(op, _, **kwargs):
         result = fmt(op, **kwargs)
-        if isinstance(op, ops.Relation):
+        if isinstance(op, ops.Relation) and not isinstance(op, ops.JoinTable):
```
kszucs, line 165: Since join tables can/should only occur in `JoinChains` it doesn't worth to show them in the repr.

cpcloud: Nice, this tightens up the repr quite a bit with long join chains.

ibis/expr/types/joins.py
```diff
@@ -46,34 +46,28 @@ def disambiguate_fields(how, left_fields, right_fields, lname, rname):
...
-            yield join.table
-
-
 def dereference_mapping_left(chain):
-    rels = dereference_targets(chain)
+    # construct the list of join table we wish to dereference fields to
+    rels = [chain.first]
+    for link in chain.rest:
+        if link.how not in ("semi", "anti"):
+            rels.append(link.table)
```
cpcloud, line 54:
```suggestion
    rels.extend(link.table for link in chain.rest if link.how not in ("semi", "anti"))
```

Just a stylistic preference, in no way a blocker.

cpcloud (approved):

PR #7839, opened 2023-12-23
refactor(ir): support join of joins without nesting the left side

kszucs:
@cpcloud this is going to fix several integrity errors in the duckdb refactor while producing a simpler join chain. 

A solution would have been to just remove the support of "continue the left chain"  in rather than just "nest the chains", but I managed to resolve the dereferencing issue by "peeling" all `JoinChain` fields in the input values. 

I still need to finish the PR.

### DIFF
--- a/ibis/expr/types/joins.py
+++ b/ibis/expr/types/joins.py
@@ -17,6 +17,8 @@ import functools
 from ibis.expr.types.relations import dereference_mapping
 import ibis
 
+from ibis.expr.rewrites import peel_join_field
+
 
 def disambiguate_fields(how, left_fields, right_fields, lname, rname):
     collisions = set()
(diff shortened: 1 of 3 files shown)


cpcloud (commented):
🚢

cpcloud (approved):

ibis/expr/types/joins.py
```diff
@@ -207,6 +209,9 @@ def select(self, *args, **kwargs):
         # if there are values referencing fields from the join chain constructed
         # so far, we need to replace them the fields from one of the join links
         subs = dereference_mapping_left(chain)
+        values = {
```
kszucs, line 212: I was experimenting with a more generic dereferencing method to allow inputs outside of the expression tree by searching the tree backwards until we find a field which can be dereferenced to the desired relation(s). It is promising, so we may be able to consolidate this.

PR #7841, opened 2023-12-23
fix(decompile): ensure that `SelfReference` is decompiled with a call to `.view()`

cpcloud:
Fixes decompilation of `SelfReference` and adds the ability to customize decompilation comparison.

### DIFF
--- a/ibis/backends/tests/sql/test_sql.py
+++ b/ibis/backends/tests/sql/test_sql.py
@@ -465,9 +465,9 @@ def test_gh_1045(test1, test2, test3, snapshot):
     t2 = test2
     t3 = test3
 
-    t3 = t3[[c for c in t3.columns if c != "id3"]].mutate(id3=t3.id3.cast("int64"))
+    t3 = t3.mutate(id3=t3.id3.cast("int64"))
 
-    t3 = t3[[c for c in t3.columns if c != "val2"]].mutate(t3_val2=t3.id3)
+    t3 = t3.mutate(t3_val2=t3.id3)
     t4 = t3.join(t2, t2.id2b == t3.id3)
 
     t1 = t1[[t1[c].name(f"t1_{c}") for c in t1.columns]]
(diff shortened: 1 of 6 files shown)


ibis/backends/tests/sql/test_sql.py
```diff
@@ -461,16 +461,20 @@ def test3():
 
 
 def test_gh_1045(test1, test2, test3, snapshot):
```
cpcloud, line 463: This test is changing, and thus the API is breaking here.

In the original issue the bug was an incorrect column pruning optimization.

The existing code doesn't work anymore, because the semantics of projections are clarified:

You can project out of a table using a parent table, only if any preceding projections still contain references used in the final projection.

That's a mouthful, so let's look at an example:

```python
t = ibis.table({"a": "int", "b": "string"})
```

This is allowed:

```python
t.mutate(x=t.a).select(t.b)  # t.mutate(x=t.a) still contains `t.b`
```

This is not allowed:

```python
t.drop("b").select(t.b)  # b no longer exists so you can't use it in a subsequent projection
```

I'll adjust the test to reflect this behavior.

cpcloud: **UPDATE:**

We *might* be able to relax this later.

ibis/tests/util.py
```diff
@@ -27,8 +29,30 @@ def assert_pickle_roundtrip(obj):
...
+):
+    """Assert that an ibis expression remains the same after decompilation.
+
+    Parameters
+    ----------
+    expr
+        The expression to decompile.
+    snapshot
+        A snapshot fixture.
+    eq
```
kszucs, line 49: Nice!

kszucs (approved):
Thanks @cpcloud!

ibis/backends/tests/sql/test_sql.py
kszucs: We need to accumulate these examples in the future docs since now we have clear semantics about the inputs we accept.

PR #7849, opened 2023-12-26
feat(datatypes): fix decimal support and add geospatial types to the sqlglot type mapper

cpcloud:
Fix support for decimal types and add support for basic geograph and geometry types to the sqlglot type mapper.


kszucs (approved):
Nice!

PR #7857, opened 2023-12-29
feat(common): add `Dispatched` base class for convenient visitor pattern implementation

kszucs:
Example use:

```py
class Visitor(Dispatched):

    def visit(self, x: int):
        assert isinstance(x, int)

    def visit(self, x: str):
        assert isinstance(x, str)


class ChildVisitor(Visitor):

    def visit(self, x):
        return super().visit(x)

    def visit(self, x: float):
        assert isinstance(x, float)
```

instead of 

```py
class Visitor(Dispatched):

    @functools.singledispatchmethod
    def visit(self, x):
        ...

    @visit.register(str)
    def visit_str(self, x):
        ...


class ChildVisitor(Visitor):

    @functools.singledispatchmethod
    def visit(self, x):
        return super().visit(x)

    @visit.register(float)
    def visit_float(self, x):
        ...
```

Used in practice to implement the new pandas executor in https://github.com/ibis-project/ibis/pull/7797

### DIFF
--- a/ibis/common/dispatch.py
+++ b/ibis/common/dispatch.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import abc
 import functools
+import inspect
 import re
 from collections import defaultdict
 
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
NICE.

PR #7859, opened 2023-12-29
refactor(api): revamp `asof` join predicates

kszucs:
Previously the `ASOF` join API was imprecise. 

The backends supporting `asof` joins require exactly one nearest match (inequality) predicate along with arbitrary number of ordinary join predicates, see [ClickHouse ASOF](https://clickhouse.com/docs/en/sql-reference/statements/select/join#asof-join-usage), [DuckDB ASOF](https://duckdb.org/docs/guides/sql_features/asof_join.html#asof-joins-with-the-using-keyword) and [Pandas ASOF](https://pandas.pydata.org/docs/reference/api/pandas.merge_asof.html). 

This change alters the API to `table.asof_join(left, right, on, predicates, ...)` where `on` is the nearest match predicate defaulting to `left[on] <= right[on]` if not an expression is given. I kept the `by` argument for compatibility reasons, but we should phase that out in favor of `predicates`.

Also ensure that all the join methods or `ir.Join` have the exact same docstrings as `ir.Table`.

BREAKING CHANGE: `on` paremater of `table.asof_join()` is now only accept a single predicate, use `predicates` to supply additional join predicates.

### DIFF
--- a/ibis/expr/types/joins.py
+++ b/ibis/expr/types/joins.py
@@ -1,22 +1,25 @@
+from __future__ import annotations
+
+import functools
+from public import public
+from typing import Any, Optional
+from collections.abc import Iterator, Mapping
+
+import ibis
+import ibis.expr.operations as ops
+
+from ibis import util
+from ibis.expr.types import Table, ValueExpr
+from ibis.common.deferred import Deferred
+from ibis.expr.analysis import flatten_predicates
+from ibis.common.exceptions import ExpressionError, IntegrityError
 from ibis.expr.types.relations import (
     bind,
     dereference_values,
+    dereference_mapping,
     unwrap_aliases,
 )
-from public import public
-import ibis.expr.operations as ops
-from ibis.expr.types import Table, ValueExpr
-from typing import Any, Optional
-from collections.abc import Iterator, Mapping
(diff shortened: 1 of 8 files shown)


ibis/expr/types/joins.py
```diff
@@ -205,8 +222,83 @@ def join(
...
 
-    def select(self, *args, **kwargs):
-        """Select expressions."""
+    @functools.wraps(Table.asof_join)
+    def asof_join(  # noqa: D102
+        self: Table,
+        right: Table,
+        on,
+        predicates=(),
+        by=(),
```
cpcloud, line 231: Is `by` used anywhere?

cpcloud: Ah, nevermind I see it.

cpcloud: Can we get rid of `by`? Having that alongside `predicates` and `on` seems a bit confusing.

cpcloud (approved):
LGTM! Excellent work!

ibis/expr/types/joins.py
kszucs: Created a follow up: https://github.com/ibis-project/ibis/issues/7869

PR #7862, opened 2023-12-29
feat(common): add a memory efficient `Node.map()` implementation

kszucs:
Alternative implementation of `map` to reduce memory usage. While `map` keeps all the results in memory until the end of the traversal, the new `map_clear()` method removes intermediate results as soon as they are not needed anymore.

Planned to be used in the pandas refactor #7797


kszucs:
Accidentally opened from a wrong branch, reopened in https://github.com/ibis-project/ibis/pull/7863

PR #7863, opened 2023-12-29
feat(common): add a memory efficient `Node.map()` implementation

kszucs:
Alternative implementation of `map` to reduce memory usage. While `map` keeps all the results in memory until the end of the traversal, the new `map_clear()` method removes intermediate results as soon as they are not needed anymore.

Planned to be used in the pandas refactor #7797

### DIFF
--- a/ibis/common/graph.py
+++ b/ibis/common/graph.py
@@ -245,14 +245,66 @@ class Node(Hashable):
         A mapping of nodes to their results.
         """
         results: dict[Node, Any] = {}
-        for node in Graph.from_bfs(self, filter=filter).toposort():
+
+        graph, _ = Graph.from_bfs(self, filter=filter).toposort()
+        for node in graph:
             # minor optimization to directly recurse into the children
             kwargs = {
                 k: _recursive_lookup(v, results)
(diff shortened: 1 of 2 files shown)


ibis/common/graph.py
```diff
@@ -245,15 +245,67 @@ def map(self, fn: Callable, filter: Optional[Finder] = None) -> dict[Node, Any]:
...
             kwargs = {
                 k: _recursive_lookup(v, results)
                 for k, v in zip(node.__argnames__, node.__args__)
             }
             results[node] = fn(node, results, **kwargs)
+
         return results
 
+    @experimental
+    def map_clear(
```
cpcloud, line 261: How about `map_prune` instead? `map_clear` makes me think everything gets deleted, but only no-longer-used results are cleared.

cpcloud (approved):
Small naming comment, otherwise LGTM!

ibis/common/graph.py
kszucs: Well, for me map_prune sounds like what you are describing. Can we merge it and maybe rename later, would like to try it out in the pandas PR :)

PR #7864, opened 2023-12-29
fix(common): intermediate result removal fails if there are duplicated dependencies

kszucs:


ibis/common/tests/test_graph.py
```diff
@@ -431,7 +431,9 @@ def test_node_find_topmost_dont_traverse_the_same_node_twice():
 
 
 def test_map_clear():
-    Z = MyNode(name="Z", children=[A])
+    Z = MyNode(name="Z", children=[A, A])
```
kszucs, line 434: `dependencies = (A, A)` in this case

cpcloud (approved):

Issue #6487, opened 2023-06-20
feat(duckdb): support `asof_join` in the DuckDB backend


cpcloud:
### Discussed in https://github.com/ibis-project/ibis/discussions/6482

<div type='discussions-op-text'>

<sup>Originally posted by **derekmahar** June 18, 2023</sup>
Does the [Ibis](https://ibis-project.org/) table [`asof_join`](https://ibis-project.org/api/expressions/tables/#ibis.expr.types.relations.Table.asof_join) function work with the [DuckDB](https://duckdb.org/) backend?  Does it use the SQL [`ASOF JOIN`](https://duckdb.org/docs/sql/query_syntax/from#as-of-joins) operator that DuckDB [0.8.0](https://github.com/duckdb/duckdb/releases/tag/v0.8.0) introduced?</div>

cpcloud:
@mesejo I think @gforsyth may already be working on this one!

gforsyth:
@kszucs added this as part of `the-epic-split`

PR #7881, opened 2024-01-02
fix(ir): `asof` join `tolerance` parameter handling generated incorrect expression

kszucs:
To correctly mimic pandas' tolerance argument we need to do the following steps:
1. Do the ASOF join with `on` and `by` parameters
2. Filter the joined result using the predicate creates be the tolerance: `left[on] <= right[on] + tolerance AND left[on] >= right[on] - toletance`
3. Join the filtered table to the `left` table again using a LEFT JOIN

This PR also adds support for ASOF joins in duckdb: fixes https://github.com/ibis-project/ibis/issues/6487

### DIFF
--- a/ibis/backends/base/sqlglot/compiler.py
+++ b/ibis/backends/base/sqlglot/compiler.py
@@ -868,14 +868,14 @@ class SQLGlotCompiler(abc.ABC):
             "anti": "left",
             "cross": None,
             "outer": "full",
-            "asof": "left",
+            "asof": "asof",
             "any_left": "left",
             "any_inner": None,
         }
         kinds = {
             "any_left": "any",
             "any_inner": "any",
-            "asof": "asof",
+            "asof": "left",
             "inner": "inner",
             "left": "outer",
             "right": "outer",
--- a/ibis/expr/tests/test_dereference.py
+++ b/ibis/expr/tests/test_dereference.py
@@ -17,9 +17,30 @@ def dereference_expect(expected):
     return {k.op(): v.op() for k, v in expected.items()}
 
 
+def test_dereference_project():
+    p = t.projection([t.int_col, t.double_col])
+
+    mapping = dereference_mapping([p.op()])
+    expected = dereference_expect(
+        {
+            p.int_col: p.int_col,
+            p.double_col: p.double_col,
+            t.int_col: p.int_col,
+            t.double_col: p.double_col,
+        }
+    )
+    assert mapping == expected
+
+
 def test_dereference_mapping_self_reference():
     v = t.view()
 
     mapping = dereference_mapping([v.op()])
-    expected = dereference_expect({})
+    expected = dereference_expect(
+        {
+            v.int_col: v.int_col,
+            v.double_col: v.double_col,
+            v.string_col: v.string_col,
+        }
+    )
     assert mapping == expected
--- a/ibis/expr/tests/test_dereference.py
+++ b/ibis/expr/tests/test_dereference.py
@@ -18,7 +18,7 @@ def dereference_expect(expected):
 
 
 def test_dereference_project():
-    p = t.projection([t.int_col, t.double_col])
+    p = t.select([t.int_col, t.double_col])
 
     mapping = dereference_mapping([p.op()])
     expected = dereference_expect(
--- a/ibis/expr/types/joins.py
+++ b/ibis/expr/types/joins.py
@@ -9,7 +9,7 @@ import ibis
 import ibis.expr.operations as ops
 
 from ibis import util
-from ibis.expr.types import Table, ValueExpr
+from ibis.expr.types import Table, Value
 from ibis.common.deferred import Deferred
 from ibis.expr.analysis import flatten_predicates
 from ibis.common.exceptions import ExpressionError, IntegrityError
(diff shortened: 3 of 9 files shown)


ibis/expr/types/joins.py
```diff
@@ -103,13 +103,14 @@ def prepare_predicates(
     for pred in util.promote_list(predicates):
         if pred is True or pred is False:
             yield ops.Literal(pred, dtype="bool")
-        elif isinstance(pred, ValueExpr):
-            node = pred.op()
-            yield dereference_value(node, deref_left, deref_right)
+        elif isinstance(pred, Value):
+            for node in flatten_predicates(pred.op()):
+                yield dereference_value(node, deref_left, deref_right)
```
kszucs, line 108: This change is not essential to the fix itself, but flattening must happen before trying to dereference the sides.

ibis/expr/types/joins.py
```diff
@@ -236,41 +236,58 @@ def asof_join(  # noqa: D102
...
         if tolerance is not None:
-            if not isinstance(on, str):
-                raise TypeError(
-                    "tolerance can only be specified when predicates is a string"
-                )
-            # construct a predicate with two sides from the two tables
-            predicates.append(self[on] <= right[on] + tolerance)
+            if isinstance(on, str):
+                # self is always a JoinChain so reference one of the join tables
+                left_on = self.op().values[on].to_expr()
```
kszucs, line 241: Dereferencing should handle this actually, so `right[on]` should be sufficient.

ibis/expr/types/joins.py
```diff
@@ -236,41 +236,58 @@ def asof_join(  # noqa: D102
...
+            else:
+                raise TypeError("`on` must be a string or a ValueExpr")
+
+            joined = self.asof_join(
+                right, on=on, predicates=predicates, lname=lname, rname=rname
+            )
+            filtered = joined.filter(
+                left_on <= right_on + tolerance, left_on >= right_on - tolerance
+            )
+            right_on = right_on.op().replace({right.op(): filtered.op()}).to_expr()
```
kszucs, line 259: Dereferencing fails to resolve it, manual replacement works. It would be nice to have the dereferencing mechanism support this case.

ibis/expr/types/relations.py
```diff
@@ -155,8 +155,14 @@ def unwrap_aliases(values: Iterator[ir.Value]) -> Mapping[str, ir.Value]:
 
 
 def dereference_mapping(parents):
-    mapping = {}
     parents = util.promote_list(parents)
+    mapping = {}
+
+    for parent in parents:
```
kszucs, line 161: Going to pull this change out since it actually fixes another bug with subqueries.

ibis/expr/operations/relations.py
```diff
@@ -205,6 +205,11 @@ def __init__(self, parent, identifier):
             identifier = next(self._uid_counter)
         super().__init__(parent=parent, identifier=identifier)
 
+    @attribute
+    def values(self):
```
kszucs, line 209: This fixes a bug in another test, pulling out.

kszucs: Done.

ibis/backends/base/sqlglot/compiler.py
```diff
@@ -868,14 +868,14 @@ def visit_JoinLink(self, op, *, how, table, predicates):
...
-            "asof": "left",
+            "asof": "asof",
             "any_left": "left",
             "any_inner": None,
         }
         kinds = {
             "any_left": "any",
             "any_inner": "any",
-            "asof": "asof",
+            "asof": "left",
```
cpcloud, line 878: This swap seems incorrect. Was it necessary for one of the backends?

ibis/expr/types/joins.py
```diff
@@ -236,41 +238,58 @@ def asof_join(  # noqa: D102
...
+                left_on = node.left.to_expr()
+                right_on = node.right.to_expr()
+            else:
+                raise TypeError("`on` must be a string or a ValueExpr")
+
+            joined = self.asof_join(
+                right, on=on, predicates=predicates, lname=lname, rname=rname
+            )
+            filtered = joined.filter(
+                left_on <= right_on + tolerance, left_on >= right_on - tolerance
```
cpcloud, line 259: Should this be `<=` and `>` instead of `<=` and `>=`?

ibis/expr/tests/test_dereference.py
```diff
@@ -17,9 +17,30 @@ def dereference_expect(expected):
     return {k.op(): v.op() for k, v in expected.items()}
 
 
+def test_dereference_project():
+    p = t.projection([t.int_col, t.double_col])
```
cpcloud, line 21: Let's avoid using this method. Can you `select` instead?

ibis/expr/types/joins.py
```diff
@@ -236,41 +238,58 @@ def asof_join(  # noqa: D102
...
+            )
+            right_on = right_on.op().replace({right.op(): filtered.op()}).to_expr()
+
+            result = self.left_join(
+                filtered, predicates=[left_on == right_on] + predicates
+            )
```
cpcloud, line 265: Can you write a comment about why the `left_join` is needed?

cpcloud (approved):
LGTM, just a request for comments about why the additional `left_join`.

ibis/backends/base/sqlglot/compiler.py
kszucs: We need `ASOF LEFT JOIN` and prior to the change it was generating `LEFT ASOF JOIN` which is not accepted by duckdb.

ibis/expr/types/joins.py
kszucs: Well, I am not sure about that. What happens if the join `on` predicate is `<=` vs `>=` then?

ibis/expr/tests/test_dereference.py
kszucs: Updated.

PR #7899, opened 2024-01-04
ci: remove `merge_group`

kszucs:


cpcloud (approved):

cpcloud:
While this change isn't related to the hang, we should remove it anyway.

PR #7871, opened 2023-12-31
refactor(trino): port to sqlglot

cpcloud:
Port the trino backend to sqlglot. I need to create some follow-up issues.

PR #7904, opened 2024-01-04
refactor(sqlglot): various sqlglot compiler and backend clean ups

cpcloud:
This PR pulls out some changes from #7871, including some removal of redundant translation rules and fixing some xpassing tests


ibis/backends/base/sqlglot/compiler.py
```diff
@@ -269,14 +269,24 @@ def visit_Alias(self, op, *, arg, name):
...
         if value is None:
             if dtype.nullable:
                 return NULL if dtype.is_null() else self.cast(NULL, dtype)
             raise com.UnsupportedOperationError(
                 f"Unsupported NULL for non-nullable type: {dtype!r}"
             )
-        elif dtype.is_integer():
+        else:
+            result = self.visit_NonNullLiteral(op, value=value, dtype=dtype)
+            if result is None:
```
kszucs, line 281: Is `None` used for error signaling here?

cpcloud: No, it's used to indicate to use the default literal rule

kszucs: The method naming is a bit hard to understand here.

cpcloud: - `visit_NonNullLiteral`: handles the non `None` case; avoids each backend having to handle `None`
- `visit_Literal`: handles `None` -> calls `visit_NonNullLiteral` -> calls `visit_DefaultLiteral` if the former returns `None`
- `visit_DefaultLiteral`: the base implementation of literal translation.

Most backends should only ever implement `visit_NonNullLiteral`

The point of this is to avoid enormous amounts of duplication of literal translation, most of which is the same across backends for the core types.

ibis/backends/clickhouse/compiler.py
```diff
@@ -41,7 +41,19 @@ def _aggregate(self, funcname: str, *args, where):
         has_filter = where is not None
         func = self.f[funcname + "If" * has_filter]
         args += (where,) * has_filter
-        return func(*args)
+
+        return func(*args, dialect=self.dialect)
+
+    @staticmethod
+    def _minimize_spec(start, end, spec):
```
kszucs, line 48: Shouldn't this be a rewrite instead?

cpcloud: Can you propose a rewrite rule? I'm not sure this can be done as a rewrite: it's altering the way specific ibis values are interpreted in sqlglot.

ibis/backends/base/sqlglot/compiler.py
cpcloud: Done.

cpcloud: Added docstrings for the relevant methods.

ibis/backends/clickhouse/compiler.py
kszucs: Ahh, this changes the sqlglot expression. Then can you rename the method to `_minimize_window_spec` and add a short comment about what it does?

kszucs (approved):

PR #7907, opened 2024-01-04
feat(flink): add primary key support

mfatihaktas:
Reviving https://github.com/ibis-project/ibis/pull/7800 which got closed after switching to `main`.


jcrist (approved):
LGTM! Thanks!

PR #7908, opened 2024-01-04
test(flink): deep dive on the tests marked for Flink in test_json.py

mfatihaktas:
Reviving https://github.com/ibis-project/ibis/pull/7777 which got closed after switching to `main`.

### DIFF
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -22,9 +22,11 @@ from ibis.backends.flink.ddl import (
     InsertSelect,
     RenameTable,
 )
+from ibis.util import gen_name
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
+    from pathlib import Path
 
     import pandas as pd
     import pyarrow as pa
@@ -119,9 +121,10 @@ class Backend(BaseBackend, CanCreateDatabase):
     def list_tables(
         self,
         like: str | None = None,
-        temp: bool = False,
+        *,
         database: str | None = None,
         catalog: str | None = None,
+        temp: bool = False,
     ) -> list[str]:
         """Return the list of table/view names.
 
@@ -198,7 +201,7 @@ class Backend(BaseBackend, CanCreateDatabase):
         database: str | None = None,
         catalog: str | None = None,
     ) -> str:
-        if is_fully_qualified(name):
+        if name and is_fully_qualified(name):
             return name
 
         return sg.table(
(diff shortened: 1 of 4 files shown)


ibis/backends/flink/__init__.py
```diff
@@ -635,6 +638,111 @@ def drop_view(
...
+            elif path_str.endswith("json"):
+                dataframe = pd.read_json(file_path, lines=True)
+            else:
+                raise ValueError(f"Unsupported file_path: {file_path}")
+
+            dataframe_list.append(dataframe)
+
+        return pd.concat(dataframe_list, ignore_index=True, sort=False)
+
+    def _read_file(
```
mfatihaktas, line 665: Bringing [this review item](https://github.com/ibis-project/ibis/pull/7777#discussion_r1440953218) here:

@jcrist wrote:
> Does flink support natively loading directly from files? If so, I think we should use that.

> If it doesn't, I'm not sure how I feel about automatically loading files using pandas and forwarding them that way. We do something similar for local in-memory backends like duckdb/pandas where file paths are unambiguously local. But for a potentially distributed system like flink, automatically using a local file reader may have unexpected behavior with path names, and also may be inefficient.

> If we do decide to handle file reading manually using an in-memory reader here, we should use the ones provided by pyarrow directly, not pandas. These have builtin support for directory datasets, and better match the behavior of other backends.

@mfatihaktas replied:

Thanks, I agree that reading the file is not ideal while creating the table. Flink actually supports creating tables with [filesystem connector](https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/table/filesystem/). This however requires specifying the schema, as `create_table()` also requires:
```python
def create_table(
        self,
        name: str,
        *,
        database: str | None = None,
        catalog: str | None = None,
        temp: bool = False,
        overwrite: bool = False,
    ) -> ir.Table:
        ...

                   raise exc.IbisError("`schema` or `obj` is required")
    ...
```

So I think we have two options:
1. Add a required argument `schema` for `read_***()`. This would deviate from the interface existing for other backends.
2. "Read" the schema from the file with `pyarrow` and feed it into `create_table()` with `filesystem` connector. This has the same issue you raised in accessing the files in a distributed system.

Which one do you think makes more sense? We could also implement both where the user can specify the schema, and if not specified we could construct the schema from the file and raise an error in case of a failure in accessing the file.

jcrist: Apologies for letting this linger.

I think we should do the following:

- Add a new `schema` argument to the flink implementations of `read_*()` that accepts a schema to use when creating the table (as you stated above). All flink `read_*()` methods would error if this argument wasn't passed.
- In a follow-up PR (or initially if you want), have `read_parquet` automatically infer the dataset schema using `pyarrow.datasets` if the schema isn't provided. We only do this for parquet instead of csv/json since for parquet this can be done semi-efficiently by reading the file metadata, while the other formats would require a full read of the data beforehand to infer the schema.

mfatihaktas: Thanks @jcrist.

> Add a new schema argument to the flink implementations of read_*() that accepts a schema to use when creating the table (as you stated above). All flink read_*() methods would error if this argument wasn't passed.

Made this change.

> In a follow-up PR (or initially if you want), have read_parquet automatically infer the dataset schema using pyarrow.datasets if the schema isn't provided. We only do this for parquet instead of csv/json since for parquet this can be done semi-efficiently by reading the file metadata, while the other formats would require a full read of the data beforehand to infer the schema.

Will create a task to follow up on this.

ibis/backends/flink/__init__.py
```diff
@@ -635,6 +638,126 @@ def drop_view(
...
+            The just-registered table
+        """
+        return self._read_file(
+            file_type="csv", path=path, schema=schema, table_name=table_name
+        )
+
+    def read_json(
+        self,
+        path: str | Path,
+        schema: sch.Schema,
```
jcrist, line 737: For all of these can you mark the `schema` parameter as

```suggestion
        schema: sch.Schema | None = None,
```

And then error nicely inside the `_read_file` method saying that the `schema` parameter is required for the `flink` backend? This keeps the method/function signatures the same across backends and provides a nicer error to the end user.

mfatihaktas: Thanks, I made the requested change.

ibis/backends/flink/__init__.py
```diff
@@ -635,6 +638,134 @@ def drop_view(
         sql = statement.compile()
         self._exec_sql(sql)
 
+    def _read_file(
+        self,
+        file_type: str,
+        path: str | Path,
+        schema: sch.Schema,
```
jcrist, line 645:
```suggestion
        schema: sch.Schema | None = None,
```

ibis/backends/flink/__init__.py
```diff
@@ -635,6 +638,134 @@ def drop_view(
...
+        ir.Table
+            The just-registered table
+
+        Raises
+        ------
+        IbisError
+            If `schema` is None.
+        """
+        if schema is None:
+            raise exc.IbisError("`schema` must be set while calling `_read_file()`")
```
jcrist, line 673:
```suggestion
            raise ValueError(f"`schema` must be explicitly provided when calling `read_{file_type}`")
```

jcrist (commented):
A couple small nits, but provided tests pass this LGTM.

ibis/backends/flink/__init__.py
mfatihaktas: Thanks for the improvements. Incorporated the suggested changes.

jcrist (approved):
Provided tests pass this LGTM

PR #7910, opened 2024-01-04
fix(datatypes): fix bad references in `to_numpy()`

deepyaman:
## Description of changes

Fix bad reference to `NumpyFormat`; use `NumpyType` as in the other `to_*` methods.

## Issues closed

Calling `.to_numpy()` on a `DataType` object resulted in:

```python
*** ImportError: cannot import name 'NumpyFormat' from 'ibis.formats.numpy' (/opt/miniconda3/envs/pandera-dev/lib/python3.11/site-packages/ibis/formats/numpy.py)
```

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -243,9 +243,9 @@ class DataType(Concrete, Coercible):
 
     def to_numpy(self):
         """Return the equivalent numpy datatype."""
-        from ibis.formats.numpy import NumpyFormat
+        from ibis.formats.numpy import NumpyType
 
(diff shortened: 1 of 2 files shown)


jcrist (approved):
Thanks! Can you add a test (I think in `ibis/expr/datatypes/tests/test_core.py`) that would hit this code path?

deepyaman:
> Thanks! Can you add a test (I think in `ibis/expr/datatypes/tests/test_core.py`) that would hit this code path?

Added a test. Not sure if there's a better way to get roundtripable types, but this should be sufficient (given that formats are already tested)?

jcrist (approved):
Thanks!

PR #7895, opened 2024-01-03
docs: add ibis.join() to docs

NickCrews:
Looks like this:

image

PR #7913, opened 2024-01-05
docs: add ibis.join() to docs

NickCrews:
redo of #7895


lostmygithubaccount:
redo of my comment there:

I think we need to update this doc too? https://ibis-project.org/tutorials/ibis-for-sql-users.html#joins

not necessarily in this PR

lostmygithubaccount (approved):

PR #7915, opened 2024-01-05
feat(ux): add Table and Column.preview()

### LINKED ISSUE #7172: feat: ux: inline way to adjust interactive config

I love the interactive defaults of ibis, how it executes the first 10 rows of a table and shows them with rich. Gets me by for 95% of the time.

But sometimes I want to look at say 30 rows. Or I'm looking at long strings or long nested types, and I don't want them cut off, so I want to be able to set `max_string` and `max_length`.

Something short and quick would be great!

An explicit method? Like `Table.preview(max_rows, max_columns, max_string, max_length, show_dtypes, ...)`. This would just be the existing method `__repr__` method, just exposed publicly.

implicitly? like `Table.limit(30)` just works and actually shows 30 rows? eg if the root Op was a LimitOp, then assume that is what the user wants, and use that instead of the config of 10. This has problems if the limit is like 10k or something and the user didn't actually want to preview the rows. So IDK, if less than 100 then use it, otherwise go back to the config value of 10? In the best case, this has the best UX, but also could lead to surprises. This also does not cover the other usecases (less needed for me, but would be nice) like max_string.

We could do either or both of these

## Existing pandas solution

`.limit(30).execute()`, but that gives me a pandas table which I don't like because
1. it doesn't show the dtype
2. it doesn't show all columns (as ibis does, which I love)
3. ugly line wrapping in some consoles

## Existing ibis config solution

set `ibis.options.repr.interactive.max_rows = 30`, and then unset afterwards.

main

NA

### Code of Conduct

NickCrews:
redo of #7408 targeting main instead of master

fixes #7172 

I wasn't sure how testing this should work so I didn't add any tests. I would really appreciate it if someone more familiar could pick this up and finish it off, I think I've gotten a lot of it out of the way.

Hopefully I've implemented this is such a way that it is testable.

Things I'm not sure about:

1. EDIT: I think this is fine to just keep them opaque as fmt_kwargs. adding **fmt_kwargs all over pretty.py. I didn't really want all the boilerplate that would be required if I was explicit with each kwarg. Possibly could pass around a container object such as options.Repr instead.
2. I encapsulated some logic into pretty.pretty_repr() because I thought that was going to be necessary for the functional change. It ended up not being needed, but I still think it is a good refactor to tie up some disparate ends. I can revert this though if you want.
3. EDIT: we decided this should just print. Should `.preview()` return a `rich.Table` for the caller/framework to render, or should it print to stdout directly? (ie the same difference as ibis.to_sql() vs ibis.show_sql())
I might argue for a kwarg like `def preview(..., file: io.StringIO | None = None)`, so the default is to just print to stdout.
This is because I imagine this method is mostly going to get used in notebooks or debuggers. For example, in the vscode debugger, if I want to see an ibis table, I have to `print()` it, otherwise I get the raw output with ANSI codes:
image
image

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -14,12 +14,14 @@ from ibis.common.deferred import Deferred, _, deferrable
 from ibis.common.grounds import Singleton
 from ibis.expr.rewrites import rewrite_window_input
 from ibis.expr.types.core import Expr, _binop, _FixedTextJupyterMixin
+from ibis.expr.types.pretty import to_rich
 from ibis.util import deprecated, warn_deprecated
 
 if TYPE_CHECKING:
     import pandas as pd
     import polars as pl
     import pyarrow as pa
+    import rich.table
 
     import ibis.expr.types as ir
     from ibis.formats.pyarrow import PyArrowData
@@ -1178,20 +1180,6 @@ class Value(Expr):
 
 @public
 class Scalar(Value):
-    def __interactive_rich_console__(self, console, options):
-        import rich.pretty
-
-        interactive = ibis.options.repr.interactive
-        return console.render(
-            rich.pretty.Pretty(
-                self.execute(),
-                max_length=interactive.max_length,
-                max_string=interactive.max_string,
-                max_depth=interactive.max_depth,
-            ),
-            options=options,
-        )
-
     def __pyarrow_result__(
         self, table: pa.Table, data_mapper: type[PyArrowData] | None = None
     ) -> pa.Scalar:
@@ -1307,10 +1295,60 @@ class Column(Value, _FixedTextJupyterMixin):
     def __array__(self, dtype=None):
         return self.execute().__array__(dtype)
 
-    def __interactive_rich_console__(self, console, options):
-        named = self.name(self.op().name)
-        projection = named.as_table()
-        return console.render(projection, options=options)
+    def preview(
+        self,
+        *,
+        max_rows: int | None = None,
+        max_length: int | None = None,
+        max_string: int | None = None,
+        max_depth: int | None = None,
+        console_width: int | float | None = None,
+    ) -> rich.table.Table:
+        """Print a subset as a single-column Rich Table.
+
+        This is an explicit version of what you get when you inspect
+        this object in interactive mode, except with this version you
+        can pass formatting options. The options are the same as those exposed
+        in `ibis.options.interactive`.
+
+        Parameters
+        ----------
+        max_rows
+            Maximum number of rows to display
+        max_length
+           Maximum length for pretty-printed arrays and maps.
+        max_string
+            Maximum length for pretty-printed strings.
+        max_depth
+            Maximum depth for nested data types.
+        console_width
+            Width of the console in characters. If not specified, the width
+            will be inferred from the console.
+
+        Examples
+        --------
+        >>> import ibis
+        >>> t = ibis.examples.penguins.fetch()
+        >>> t.island.preview(max_rows=3, max_string=5)  # doctest: +SKIP
+        ┏━━━━━━━━┓
+        ┃ island ┃
+        ┡━━━━━━━━┩
+        │ stri…  │
+        ├────────┤
+        │ Torg…  │
+        │ Torg…  │
+        │ Torg…  │
+        │ …      │
+        └────────┘
+        """
+        return to_rich(
+            self,
+            max_rows=max_rows,
+            max_length=max_length,
+            max_string=max_string,
+            max_depth=max_depth,
+            console_width=console_width,
+        )
 
     def __pyarrow_result__(
         self, table: pa.Table, data_mapper: type[PyArrowData] | None = None
--- a/ibis/expr/types/pretty.py
+++ b/ibis/expr/types/pretty.py
@@ -4,52 +4,63 @@ import datetime
 import json
 from functools import singledispatch
 from math import isfinite
+from typing import TYPE_CHECKING
 from urllib.parse import urlparse
 
 import rich
+import rich.table
 from rich.align import Align
-from rich.console import Console
+from rich.pretty import Pretty
 from rich.text import Text
 
 import ibis
 import ibis.expr.datatypes as dt
 
-# A console with all color/markup disabled, used for `__repr__`
-simple_console = Console(force_terminal=False)
+if TYPE_CHECKING:
+    from ibis.expr.types import Column, Expr, Scalar, Table
 
 
-def _format_nested(values):
-    interactive = ibis.options.repr.interactive
+def _format_nested(
+    values,
+    *,
+    max_length: int | None = None,
+    max_string: int | None = None,
+    max_depth: int | None = None,
+):
     return [
-        rich.pretty.Pretty(
+        Pretty(
             v,
-            max_length=interactive.max_length,
-            max_string=interactive.max_string,
-            max_depth=interactive.max_depth,
+            max_length=max_length or ibis.options.repr.interactive.max_length,
+            max_string=max_string or ibis.options.repr.interactive.max_string,
+            max_depth=max_depth or ibis.options.repr.interactive.max_depth,
         )
         for v in values
     ]
 
 
 @singledispatch
-def format_values(dtype, values):
-    return _format_nested(values)
+def format_values(dtype, values, **fmt_kwargs):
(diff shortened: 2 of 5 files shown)


ibis/expr/types/generic.py
```diff
@@ -1318,9 +1319,69 @@ def __array__(self, dtype=None):
...
+        This is an explicit version of what you get when you inspect
+        this object in interactive mode, except with this version you
+        can pass formatting options. The options are the same as those exposed
+        in `ibis.options.interactive`.
+
+        Parameters
+        ----------
+        max_rows
+            Maximum number of rows to display
+        show_types
```
cpcloud, line 1345: Can you remove this argument? I'm not sure there's a reason to hide type information.

cpcloud (changes requested):
LGTM, modulo removal of the `show_types` argument!

ibis/expr/types/generic.py
NickCrews: roger, I will do this. Still waiting to hear back on any other feedback (especially my question of returning vs printing the rich table) before I go in an change everything at the same time.

cpcloud:
@NickCrews Regarding returning versus printing, I vote for printing.

It seems quite inconvenient to have to write `print(t.preview())`.

Are there any places where automatically calling `print` won't work?

cpcloud:
Can you make `to_rich` private for now (as `_to_rich`)? I'd rather not expose something we don't know is useful beyond internals.

NickCrews:
yeah, easy enough to expose/figure it out later if you get a feature request, but hard to take back :)

ibis/expr/types/pretty.py
```diff
@@ -227,13 +275,36 @@ def format_dtype(dtype):
...
+) -> RichTable:
+    from ibis.expr.types import Column, Table
+
+    if max_rows is None:
+        max_rows = ibis.options.repr.interactive.max_rows
+    if max_columns is None:
+        max_columns = ibis.options.repr.interactive.max_columns
     if console_width is None:
         console_width = float("inf")
+    show_types = ibis.options.repr.interactive.show_types
```
NickCrews, line 296: should we just deprecate the entire `show_types` option from `ibis.options.repr.interactive`? and always make it true?

ibis/expr/types/generic.py
NickCrews: done.

NickCrews:
OK, not sure why the widths of the tables are changing in the doctests, I don't see what is different. But I can dig into that once the other questions I have are addressed.

ibis/expr/types/pretty.py
```diff
@@ -227,13 +275,36 @@ def format_dtype(dtype):
...
+
+    if isinstance(tablish, Column):
+        named = tablish.name(tablish.op().name)
```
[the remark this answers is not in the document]
cpcloud: Hm, I _think_ we may already be doing that in `the-epic-split`. We're close to merging that. Any chance we can revisit that when it's merged into `main`?

Happy to help with rebase of course :)

ibis/expr/types/pretty.py
cpcloud: Let's leave it as is.

ibis/expr/types/pretty.py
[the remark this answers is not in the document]
NickCrews: yeah it looks like this sorted itself out! nice!

NickCrews:
for myself: can repro the doctest failure with `pytest --doctest-modules ibis/expr/types/maps.py`

@cpcloud so now the only test failing are the doctests. I'm not sure exactly what is causing the change in behavior there. Want me/us to track that down and keep the behavior consistent? Or just update the docstrings to the new behavior? I sort of like the new behavior more (it leads to being more compact)

I also need to add tests for the actual .preview method, but let's get the implementation settled to avoid churn with me rewriting the test.

NickCrews:
OK the failing doctests are all for nested datatypes. There is some conditional logic inside `to_rich_table()`:

```python
        if show_types and not isinstance(dtype, (dt.Struct, dt.Map, dt.Array)):
            # Don't truncate non-nested dtypes
            min_width = max(min_width, len(dtype_str))
```

investigating this now

NickCrews:
no, that's not it. 

On `main`, `to_rich_table` is getting called with `width=153` (or the width of the rich console) found in __interactive_rich_console__.

on this branch, `to_rich_table` is getting called with `width=None`.

Looking into this more

NickCrews:
OK, there was an inconsistency between the `__interactive_rich_console__` implementations in Column vs Table. Now they are unified, along with Scalars

ibis/expr/types/pretty.py
```diff
@@ -219,21 +262,76 @@ def isnull(x):
     return out, min_width, max_width
 
 
-def format_dtype(dtype):
-    max_string = ibis.options.repr.interactive.max_string
+def format_dtype(dtype, max_string: int) -> Text:
```
NickCrews, line 265: this is a small change. Before, the passed in max_string was ignored, and we only used the one in options.repr.interactive.

So `t.island.preview(max_string=5)` now gives:

```
        ┏━━━━━━━━┓
        ┃ island ┃
        ┡━━━━━━━━┩
        │ stri…  │
        ├────────┤
        │ Torg…  │
        │ Torg…  │
        │ Torg…  │
        │ …      │
        └────────┘
```

without this, you get the full `string`

ibis/expr/types/pretty.py
```diff
@@ -219,21 +262,76 @@ def isnull(x):
...
     strtyp = str(dtype)
     if len(strtyp) > max_string:
         strtyp = strtyp[: max_string - 1] + "…"
     return Text.styled(strtyp, "dim")
 
 
-def to_rich_table(table, console_width=None):
-    if console_width is None:
-        console_width = float("inf")
+def to_rich(
```
NickCrews, line 272: could use singledispatch, but that felt like overkill for two branches

NickCrews:
@cpcloud OK now I got this all fixed up! Can you review this now? The implementation has changed decently since the last time you looked.

This is doing both a refactor and a new feature at the same time. If this is too much, I can go through and split this into two commits, first the refactor and then the feature.

ibis/expr/types/pretty.py
```diff
@@ -13,43 +14,72 @@
...
-simple_console = Console(force_terminal=False)
+if TYPE_CHECKING:
+    from rich.table import Table as RichTable
 
+    from ibis.expr.types import Column, Expr, Scalar, Table
 
-def _format_nested(values):
-    interactive = ibis.options.repr.interactive
+
+def pretty_repr(expr: ibis.Expr) -> str:
```
NickCrews, line 25: This function feels similar to `__interactive_rich_console__`, in a followup they might want to get unified.

NickCrews:
@cpcloud sorry to bug you, could this get a review? I can fix the rebase error once I get a +1 on the general approach

cpcloud:
@NickCrews I haven't forgotten :) We'll definitely get to it for 9.0, I can't promise anything this week though :)

cpcloud:
@NickCrews I just DM'd you on Zulip, was hoping I could pair with you on this to get it over the finish line.

NickCrews:
@cpcloud what do you want to do about this failing doctest? Basically, `.preview()` is returning a rich.Table. In a jupyter notebook, or ipython, this works and you get the full table output (I think because Table has a [_repr_mimebundle method](https://github.com/Textualize/rich/blob/0d7e598fe7ead1cca6b590c1de7f54ba19255d11/rich/jupyter.py#L41)). In the vanilla python REPL though, you just get `<rich.table.Table object at 0x12b961450>`. So either we can
- A: keep it as is. Sorry vanilla REPL people you get a bad experience, but everyone else is OK. Add a #ignore pragma to this doctest
- B: convert that Table to a str, so all environments get the full render. Sorry to anyone who wants to do anything interesting with the Table
- C: add a flag `format: Literal["str", "rich"] = "str"`. By default do A, but if someone really wants, they can get the rich Table.

cpcloud:
I think option A until we know more about how people are going to use it.

NickCrews:
OK @cpcloud, once you mark as resolved then this should be good to merge!

ibis/expr/types/pretty.py
cpcloud: Yeah, this seems fine as you have it.

cpcloud (approved):

Issue #7295, opened 2023-10-04
feat: remove Value.greatest() and Value.least()


NickCrews:
### Is your feature request related to a problem?

- when exposed as a method they are confusingly similar to `Column.min()` and `Column.max()`.
- these already are at `ibis.greatest()` and `ibis.least()`
- they make more sense as free, pure functions, than methods since all of the args have equal precedence. Compare this to Table.difference(), where the order of arguments matters, so it makes sense to have it as a method.

The only benefit that I can think offor how it is right now is that it allows method chaining. But I don't think that is worth it.

### Describe the solution you'd like

Either hard or soft deprecate them. I'd prefer hard but soft would be OK too. The existing users seem to have this figured out, I mostly want to remove footguns for new users, and it's not a large maintenance burden to continue supporting it since it can just steal the `ibis.greatest()` implementation, so just stopping advertising it could be OK.

### What version of ibis are you running?

main

### What backend(s) are you using, if any?

NA

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #7916, opened 2024-01-05
depr: deprecate Value.least() and Value.greatest()

NickCrews:
moves towards https://github.com/ibis-project/ibis/issues/7295.

This just removes the methods from the docs
(by removing the docstring)
and informs current users of the API as to the upgrade path
(using the `@deprecated` decorator)

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -8,8 +8,8 @@ from public import public
 import ibis
 import ibis.common.exceptions as com
 import ibis.expr.datatypes as dt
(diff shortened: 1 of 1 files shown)


NickCrews:
preview

cpcloud:
@jcrist @gforsyth @kszucs Thoughts on this one?

I personally am inclined to merge it.

Any objections?

gforsyth:
no objection from me

jcrist (approved):
Makes sense to me

cpcloud:
Thanks @NickCrews!

PR #7894, opened 2024-01-03
feat(flink): implement support for temporal join

mfatihaktas:
## Description of changes

Continuation of https://github.com/ibis-project/ibis/pull/7691.

PR #7921, opened 2024-01-05
feat(flink): implement support for temporal join

mfatihaktas:
## Description of changes

Continuation of https://github.com/ibis-project/ibis/pull/7691.

Also reopens https://github.com/ibis-project/ibis/pull/7894 that got previously closed after switching to `main`.

### DIFF
--- a/ibis/backends/base/sql/compiler/query_builder.py
+++ b/ibis/backends/base/sql/compiler/query_builder.py
@@ -98,7 +98,7 @@ class TableSetFormatter:
             rows = " UNION ALL ".join(f"(SELECT {raw_row})" for raw_row in raw_rows)
             return f"({rows})"
 
-    def _format_table(self, op):
+    def _format_table(self, op, include_alias: bool = True):
         # TODO: This could probably go in a class and be significantly nicer
         ctx = self.context
 
@@ -135,7 +135,7 @@ class TableSetFormatter:
             subquery = ctx.get_compiled_expr(op)
             result = f"(\n{util.indent(subquery, self.indent)}\n)"
 
-        if result != alias:
+        if include_alias and result != alias:
             result = f"{result} {alias}"
 
         if isinstance(orig_op, ops.Sample):
--- a/ibis/backends/flink/compiler/core.py
+++ b/ibis/backends/flink/compiler/core.py
@@ -3,10 +3,12 @@
 from __future__ import annotations
 
 import functools
+from io import StringIO
 
 import ibis.common.exceptions as com
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
+from ibis import util
 from ibis.backends.base.sql.compiler import (
     Compiler,
     Select,
@@ -18,6 +20,16 @@ from ibis.backends.flink.translator import FlinkExprTranslator
 
 
 class FlinkTableSetFormatter(TableSetFormatter):
+    _join_names = {
+        **TableSetFormatter._join_names,
+        ops.AsOfJoin: "JOIN",
+    }
+
+    def __init__(self, parent, node, indent=2):
+        super().__init__(parent, node, indent)
+
+        self.join_table_aliases = []
+
     def _quote_identifier(self, name):
         return quote_identifier(name, force=True)
 
@@ -42,13 +54,13 @@ class FlinkTableSetFormatter(TableSetFormatter):
             function_type = "CUMULATE"
         return f"TABLE({function_type}({format_windowing_tvf_params(op, self)}))"
 
-    def _format_table(self, op) -> str:
+    def _format_table(self, op, include_alias: bool = True) -> str:
         ctx = self.context
         if isinstance(op, ops.WindowingTVF):
             formatted_table = self._format_window_tvf(op)
             return f"{formatted_table} {ctx.get_ref(op)}"
         else:
-            result = super()._format_table(op)
+            result = super()._format_table(op=op, include_alias=include_alias)
 
         ref_op = op
         if isinstance(op, ops.SelfReference):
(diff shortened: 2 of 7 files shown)


ibis/backends/base/sql/compiler/query_builder.py
```diff
@@ -41,12 +41,19 @@ def __init__(self, parent, node, indent=2):
         self.indent = indent
 
         self.join_tables = []
```
mfatihaktas, line 43: Responding to https://github.com/ibis-project/ibis/pull/7894#discussion_r1441543842:

Good point. I assumed that the addition of `record_join_tables()` and `alias` arg for `_format_table()` would help the logic for all backends but I surely cannot make the call on that :)

@cpcloud What do you think about this change in `query_builder.py`?

ibis/expr/operations/relations.py
```diff
@@ -606,14 +606,27 @@ def _dedup_join_columns(expr: ir.Table, lname: str, rname: str):
...
+    # For inner joins composed exclusively of equality predicates, we can
+    # avoid renaming columns with colliding names if their values are
+    # guaranteed to be equal due to the predicate. Here we collect a set of
+    # colliding column names that are known to have equal values between
+    # the left and right tables in the join.
+    #
+    # TODO (mehmet): Included also asof joins, check with the reviewers about
+    # the implications!
+    # Edit: This change leads to assert failures in tests for other backends.
+    # Q. Why was the deduplication of columns implemented only for `InnerJoin`?
```
mfatihaktas, line 618: We need clarification on this question:
In `_dedup_join_columns()`, why was the deduplication of columns implemented only for `InnerJoin`?

cpcloud: The only case where the key columns can be deduplicated without loss of information is when:

1. The join is an inner join
2. The predicates are all equality predicates

If either of those things is false then deduplicating columns loses information since the join keys are not equivalent.

ibis/backends/base/sql/compiler/query_builder.py
```diff
@@ -41,12 +41,19 @@ def __init__(self, parent, node, indent=2):
...
+        self.join_table_aliases = []
         self.join_types = []
         self.join_predicates = []
 
     def _translate(self, expr):
         return self.parent._translate(expr)
 
+    def record_join_tables(self, op: ops.Relation):
```
mfatihaktas, line 51: @chloeh13q asked:

> would it be better to have key-value pairs as opposed to two separate lists?

The only place these two lists are used is here:

```python
        for jtype, table, alias, preds in zip(
                self.join_types, self.join_tables[1:],
                self.join_table_aliases[1:], self.join_predicates
        ):
```

We could certainly "encode" the implicit-index-based relation between these zipped lists in a more explicit data structure. But that would call for a slight refactoring of `TableSetFormatter`.

Perhaps, @cpcloud has opinions about this.

ibis/backends/base/sql/compiler/query_builder.py
```diff
@@ -98,7 +107,7 @@ def _format_in_memory_table(self, op):
             rows = " UNION ALL ".join(f"(SELECT {raw_row})" for raw_row in raw_rows)
             return f"({rows})"
 
-    def _format_table(self, op):
+    def _format_table(self, op, no_alias: bool = False):
```
mfatihaktas, line 110: @chloeh13q said:

> just a thought: no_alias = False is a double negative, which seems slightly confusing compared to just doing alias = True/False

Thanks, agreed. Switched to `include_alias`.

ibis/backends/flink/tests/test_ddl.py
```diff
@@ -489,3 +488,159 @@ def test_insert_simple_select(con, tempdir_sink_configs):
         temporary_file = next(iter(os.listdir(tempdir)))
         with open(os.path.join(tempdir, temporary_file)) as f:
             assert f.read() == '"fred flintstone",35\n"barney rubble",32\n'
+
+
+@pytest.mark.broken(
```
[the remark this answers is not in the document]
chloeh13q: I think you can create a new file, since it doesn't really fit into any of the existing ones

ibis/backends/flink/tests/test_ddl.py
```diff
@@ -489,3 +488,155 @@ def test_insert_simple_select(con, tempdir_sink_configs):
...
+
+    expr = table_left.asof_join(
+        table_right,
+        predicates=[
+            table_left["key"] == table_right["key"],
+            table_left["time"] >= table_right["time"],
+        ],
+    )
+    sql = con.compile(expr)
+    print(f"sql = \n{sql}")
```
chloeh13q, line 555: should this be an assert?

ibis/backends/flink/tests/test_ddl.py
```diff
@@ -489,3 +488,155 @@ def test_insert_simple_select(con, tempdir_sink_configs):
...
+
+    df = expr.to_pandas()
+    print(f"df = \n{df}")
```
chloeh13q, line 558: ditto

ibis/backends/flink/tests/test_ddl.py
chloeh13q, line 555: do we intend on keeping these print statements since this is a unit test?

ibis/backends/flink/compiler/core.py
```diff
@@ -60,6 +67,116 @@ def _format_table(self, op) -> str:
...
+        op = self.node
+
+        print(
+            "Before append: \n"
+            f"join_tables= {self.join_tables}"
+        )
+        if isinstance(op, ops.Join):
+            self._walk_join_tree(op)
+        else:
+            # TODO (mehmet): What does this do exactly?
```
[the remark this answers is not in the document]
mfatihaktas: Resolved this while working on the temporal-join demo.

ibis/backends/flink/tests/test_ddl.py
mfatihaktas: No, removed them.

ibis/backends/flink/tests/test_ddl.py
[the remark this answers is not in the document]
mfatihaktas: Ok, created `test_json.py` and moved the temporal join tests there.

ibis/backends/flink/tests/test_join.py
```diff
@@ -0,0 +1,165 @@
...
+
+    expected_df_to_str = (
+        "  id  bool_col  smallint_col  int_col           timestamp_col  id_right  bool_col_right  smallint_col_right  int_col_right     timestamp_col_right\n"
+        "6690      True             0        0 2010-11-01 00:00:00.000      6690            True                   0              0 2010-11-01 00:00:00.000\n"
+        "6691     False             1        1 2010-11-01 00:01:00.000      6691           False                   1              1 2010-11-01 00:01:00.000\n"
+        "6692      True             2        2 2010-11-01 00:02:00.100      6692            True                   2              2 2010-11-01 00:02:00.100\n"
+        "6693     False             3        3 2010-11-01 00:03:00.300      6693           False                   3              3 2010-11-01 00:03:00.300\n"
+        "6694      True             4        4 2010-11-01 00:04:00.600      6694            True                   4              4 2010-11-01 00:04:00.600"
+    )
+    assert join_df.to_string(index=False) == expected_df_to_str
```
chloeh13q, line 165: also wondering if it's prefered to use `tm.assert_frame_equal()` here instead? i haven't seen the comparison this way in ibis unit tests

mfatihaktas: Good point. I actually tried to validate with something like the following:
```python
    expected = pd.merge_asof(
        df, df,
        on="timestamp_col",
        by="id",
    )

    tm.assert_frame_equal(result, expected)
```
This however does not work as intended since the number of columns in `expected` will be less than that in `result` due to the lack of proper deduplication of columns for temporal join, [per this this thread](https://github.com/ibis-project/ibis/pull/7921#discussion_r1449132806).

cpcloud (changes requested):
I have some concerns about the use of `SYSTEM_TIME` as well as the amount of compiler code duplication happening here.

Wondering how important it is to release the Flink backend in 8.0.0 with this feature, or whether we can wait until the port to sqlglot is complete (9.0).

mfatihaktas:
> I have some concerns about the use of `SYSTEM_TIME` as well as the amount of compiler code duplication happening here.
> 
> Wondering how important it is to release the Flink backend in 8.0.0 with this feature, or whether we can wait until the port to sqlglot is complete (9.0).

Thanks @cpcloud. I discussed your comments and suggestions with the team. We agree that `as-of-join` does not fully include temporal join semantics and temporal join is also not required for the release. So, we decided to put a hold on this PR as you suggested, and created a backlog task to follow up on it. With that, I will close this PR for now.

PR #7942, opened 2024-01-08
fix(postgres): ensure that no timezone conversion takes place on timestamptz columns when selecting them out

### LINKED ISSUE #7872: Potential timezone bug in ibis-framework[postgres]
### What happened?

In upgrading from ibis-framework<7.0 to ibis-framework==7.2 many queries I had written suddenly started returning unexpected datetime values.  I believe this is due to a bug in the way timezones are being handled by the postgres ibis backend, but I can't be sure.

7.2.0

postgres

### Relevant log output

```sh
I was unable to reproduce my exact bug, but I do have code the replicates what could be a related error.

I have a complete docker set up to replicate the behavior here:
https://github.com/robdmc/ibis_debug

In re-reading my notebook I see a couple of comments that come across as grumbly.  I apologize in advance.  I created this at the end of a very long frustrating day of bug-hunting.  :)

I am deeply grateful for the work you all are doing with ibis.  I LOVE this project.  Thank you.
```

### Code of Conduct

cpcloud:
## Description of changes

This PR addresses an issue where we are unintentionally converting values from
their time zone to UTC by using the `AT TIME ZONE 'timezone'` syntax.

`AT TIME ZONE` is used to convert a `timestamp with time zone` to a `timestamp
without time zone` after adjusting for the timezone offset. This is not the
desired behavior.

The values we get back from the database are always in UTC, regardless of the
system timezone so this PR leaves them as is.

## Issues closed

Resolves #7872.

### DIFF
--- a/ibis/backends/postgres/registry.py
+++ b/ibis/backends/postgres/registry.py
@@ -309,11 +309,6 @@ def _table_column(t, op):
     sa_table = get_sqla_table(ctx, table)
     out_expr = get_col(sa_table, op)
 
-    if op.dtype.is_timestamp():
-        timezone = op.dtype.timezone
-        if timezone is not None:
-            out_expr = out_expr.op("AT TIME ZONE")(timezone).label(op.name)
-
     # If the column does not originate from the table set in the current SELECT
     # context, we should format as a subquery
     if t.permit_subquery and ctx.is_foreign_expr(table):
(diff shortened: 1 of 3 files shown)


cpcloud:
cc @robdmc for testing

gforsyth (approved):
Looks good to me.  We can leave this open until @robdmc has a chance to test, or merge it in now.

cpcloud:
Let's merge, and we can follow up with any necessary related fixes.

PR #7944, opened 2024-01-08
feat(duckdb-geospatial): add support for flipping coordinates

### LINKED ISSUE #7835: [discussion] Support `always_xy` for GeoConvert
After playing around a bit with duckdb `ST_Transform` function and our Ibis implementation `convert` I ran into couple of problems, that are mostly related to conventions, and choices of a what's the default.

After a discussion with @gforsyth we found out that in `duckdb`:

- Output is always in x, y format
- The assume input (with default value `always_xy=False`) is lat/long which is equivalent to y, x
- If the user is providing input data as lon/lat they need to pass `always_xy=True`

But it also depends on the projection parameters you pass in. I think the big source of confusion is that wgs84 (epsg 4326) is standardized to be in lat/lon (even though almost all gis systems ignore that and pretend its always lon/lat), while basically every other projection youd care about is in lon/lat. 

In `Geopandas` things are different (@jorisvandenbossche please correct us here if we are wrong) : 

- All the inputs are expected as lon/lat 
- Somehow things get handled for the case of `epsg 4326` 

I think this means that we should expose the `always_xy` variable to the user but maybe not?. That being said the name is terrible and we should change it, ideas are welcome.

ncclementi:
## Description of changes

- Extend api to support ST_FlipCoordinates
- Clarify convert (equivalent of st_transform) docs to mention it that coordinates are expected as XY (longitude/latitude)

## Issues closed

Closes #7835

### DIFF
--- a/ibis/expr/types/geospatial.py
+++ b/ibis/expr/types/geospatial.py
@@ -674,7 +674,8 @@ class GeoSpatialValue(NumericValue):
     def convert(
         self, source: ir.StringValue, target: ir.StringValue | ir.IntegerValue
     ) -> GeoSpatialValue:
-        """Transform a geometry into a new SRID (CRS).
+        """Transform a geometry into a new SRID (CRS). The coordinates are assumed
+        to always be XY (Longitude-Latitude).
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


ibis/expr/types/geospatial.py
```diff
@@ -687,6 +688,11 @@ def convert(
         -------
         GeoSpatialValue
             Transformed geometry
+
+        See Also
+        --------
+        [`flip_coordinates`](#ibis.expr.types.geospatial.GeoSpatialValue.flip_coordinates)
```
ncclementi, line 694: Is this the right way to reference a "see also"? I followed another example, I wasn't quite sure it works, have docs build broken locally at the moment, but didn't want to delayed pushing this.

gforsyth (approved):
:shipit: 

(sorry for the delay, I thought I merged this in hours ago)

PR #7948, opened 2024-01-10
test(datatypes): update test to test double quotes

deepyaman:
## Description of changes

Minor issue I found while looking at examples of string specification for datatypes.


cpcloud (dismissed):
Thanks!

cpcloud:
@deepyaman This should actually test both single and double quotes, instead of only testing one variant.

deepyaman:
> @deepyaman This should actually test both single and double quotes, instead of only testing one variant.

@cpcloud There's also an existing single quote test just above.  I can parameterize it? Thought about it,  but just made the most minimal fix initially.

cpcloud:
Ah, ok. Then this is fine as is!

cpcloud (approved):

PR #7952, opened 2024-01-10
docs(blog-post): replicate sapatial dev guru blog

ncclementi:
## Description of changes
Replicate Spatial Dev Guru's blog using Ibis

## Issues closed
Closes #7717

##TODO / Update on last query

The last query is not currently working, the code should be as follows:

```python
some_point = biketrip.limit(1)

biketrip.filter(_.pickup_point.within(
    some_point.select(_.pickup_point.buffer(500)).to_array())).execute()
```

The error occurs in the query in
```python-traceback

ibis/ibis/backends/base/sql/alchemy/query_builder.py", line 152, in _format_table

    result = alias if hasattr(alias, "name") else result.alias(alias)
```

For some reason both `result` and `alias` are `'t1'`. After chatting with @gforsyth we decided investigating the reason of this failure is not worth it as the `query_builder.py` code will no longer exist after the `sqlglot` port. The options are: 

1.Replicate the blog fully in the branch  `the-epic-split` (This branch should have the geospatial literals support) and things work (just tested it and compare to duckdb result), leave the blog on draft until the branch is merged. 

2. Release the blog without the last query.   

I personally think, that (1) is worth the wait, the last query would look like.  

```python
import shapely
p = shapely.geometry.Point((-8238257.278, 4969875.321))
pb = p.buffer(500)

biketrip.filter(biketrip.pickup_point.within(pb)).execute()
```

If people agrees, I'll update the blog to be written relying on `the-epic-split` and leave it on draft until that branch is merged. 

cc: @gforsyth and @jcrist

PR #7961, opened 2024-01-11
fix(sqlalchemy): properly handle aliases of extracted subqueries

jcrist:
This (potentially) fixes the bug found in #7952. With this fix the query at least runs successfully and I _think_ the fix makes sense. Pushing this up to see if it passes full CI for now. I haven't attempted to write a proper test for this yet.

### DIFF
--- a/ibis/backends/base/sql/alchemy/query_builder.py
+++ b/ibis/backends/base/sql/alchemy/query_builder.py
@@ -144,6 +144,11 @@ class _AlchemyTableSetFormatter(TableSetFormatter):
         elif ctx.is_extracted(op):
             if isinstance(orig_op, ops.SelfReference):
                 result = ctx.get_ref(op)
+            elif isinstance(alias, str):
+                result = sa.table(
+                    alias,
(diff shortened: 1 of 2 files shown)


ibis/backends/base/sql/alchemy/query_builder.py
```diff
@@ -144,6 +144,13 @@ def _format_table(self, op):
         elif ctx.is_extracted(op):
             if isinstance(orig_op, ops.SelfReference):
                 result = ctx.get_ref(op)
+            elif isinstance(alias, str):
+                result = sa.Table(
+                    alias,
+                    sa.MetaData(),
+                    *translator._schema_to_sqlalchemy_columns(orig_op.schema),
+                    quote=translator._quote_table_names,
+                )
```
cpcloud, line 153: Just a thought, but can we avoid the heavier-weight `sa.Table` and using `sa.table(alias, *translator...)`?

That said, take it with a grain of salt since this entire thing is going away soon.

cpcloud (approved):

cpcloud (approved):

PR #7972, opened 2024-01-12
refactor(sqlglot): make anonymous functions easier to use and remove `array_func` hack

cpcloud:
Clean up the use of anonymous functions and remove `array_func` hack in favor of `f.array`

### DIFF
--- a/ibis/backends/base/sqlglot/compiler.py
+++ b/ibis/backends/base/sqlglot/compiler.py
@@ -31,6 +31,8 @@ from ibis.expr.rewrites import (
 )
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     import ibis.expr.schema as sch
     import ibis.expr.types as ir
     from ibis.backends.base.sqlglot.datatypes import SqlglotType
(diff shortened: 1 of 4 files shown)


ibis/backends/base/sqlglot/compiler.py
```diff
@@ -59,32 +61,55 @@ def __getitem__(self, key: str) -> sge.Var:
...
+class AnonymousFuncGen:
+    __slots__ = ()
+
+    def __getattr__(self, name: str) -> Callable[..., sge.Anonymous]:
+        return lambda *args: sge.Anonymous(
+            this=name, expressions=list(map(sge.convert, args))
+        )
+
+    def __getitem__(self, key: str) -> Callable[..., sge.Anonymous]:
+        return getattr(self, key)
```
cpcloud, line 73: This isn't used internally anywhere, it's here for consistency with the other function generators.

Happy to remove it until we need it if that's desirable.

kszucs: I like consistency :)

kszucs (approved):

PR #7981, opened 2024-01-13
ci(postgres): avoid pgxn due to expired certificate

cpcloud:
Fixes currently-failing CI due to an expired `pgxn` SSL certificate. I removed the usage of pgxn entirely so we hopefully never have to deal with that issue again.


cpcloud:
Opened an issue upstream (https://github.com/pgxn/pgxn/issues/6) about this, but we're probably going to avoid pgxn for the foreseeable future.

cpcloud:
Merging on green to get CI back to passing.

PR #7950, opened 2024-01-10
refactor(pyspark): reimplement the backend using the new relational operations and sqlglot  

kszucs:
Summary
-------
Port the Spark backend to use the new relational operations as well as move to
Spark SQL from PySpark and take advantage of the new SQLGlot backend.

Test suite changes
------------------
```
OLD: 1223 passed, 33 skipped, 29316 deselected, 285 xfailed in 375.24s (0:06:15)
NEW: 1334 passed, 33 skipped, 30002 deselected, 155 xfailed in 333.94s (0:05:33)
```

Advantages
----------
- Spark SQL is faster than PySpark.
- Spark SQL has broader feature support than PySpark, for example it properly
  supports subqueries.
- We can reuse the base SQLGlot backend making the PySpark compiler as thin as
  500 lines of code.

I kept the previous datatype and schema mappers since some of the DDL operations
directly work on the PySpark session and Catalog objects, most of these would
be harder to express in Spark SQL.

PR #7989, opened 2024-01-15
feat(pyspark): add support for PySpark 3.5

kszucs:
depends on #7950

### DIFF
--- a/ibis/backends/conftest.py
+++ b/ibis/backends/conftest.py
@@ -4,6 +4,7 @@ import contextlib
 import importlib
 import importlib.metadata
 import itertools
+import operator
 from functools import cache
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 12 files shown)


ibis/backends/conftest.py
```diff
@@ -4,13 +4,15 @@
...
 import itertools
+import operator
 from functools import cache
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 import _pytest
 import numpy as np
 import pandas as pd
+import pkg_resources
```
cpcloud, line 15: Can you use `importlib.metadata` instead? `pkg_resources` is kind of awful.

pyproject.toml
```diff
@@ -79,7 +79,7 @@ pydata-google-auth = { version = ">=1.4.0,<2", optional = true }
 pydruid = { version = ">=0.6.5,<1", optional = true, extras = ["sqlalchemy"] }
 pymysql = { version = ">=1,<2", optional = true }
 pyodbc = { version = ">=4.0.39,<6", optional = true }
-pyspark = { version = ">=3,<3.4", optional = true } # pyspark is heavily broken by numpy >=1.24 and pandas >=2
+pyspark = { version = ">=3", optional = true }
```
cpcloud, line 82: Should we keep an upper bound on this at `<4`? PySpark does have a history of being incompatible with the latest pandas and numpy.

cpcloud (approved):
Sweet!

ibis/backends/conftest.py
kszucs: Sure, updating.

pyproject.toml
kszucs: Added.

cpcloud (approved):

PR #7992, opened 2024-01-16
refactor(pyspark): remove sqlalchemy dependency from the pyspark backend

cpcloud:
Remove the `sqlalchemy` dependency from the pyspark backend entirely.


jcrist (approved):

kszucs (approved):
Thanks!

PR #7996, opened 2024-01-16
fix(rewrite): avoid dereferencing all binary operations

cpcloud:
This PR fixes an issue where predicates containing non-comparison boolean operations raised integrity errors. @kszucs Can you help clarify why this works?

### DIFF
--- a/ibis/expr/tests/test_newrels.py
+++ b/ibis/expr/tests/test_newrels.py
@@ -1314,3 +1314,45 @@ def test_join_method_docstrings():
         join_method = getattr(joined, method)
         table_method = getattr(t1, method)
         assert join_method.__doc__ == table_method.__doc__
+
+
+def test_join_with_compound_predicate():
+    t1 = ibis.table(name="t", schema={"a": "string", "b": "string"})
+    t2 = t1.view()
+
+    joined = t1.join(
+        t2,
+        [
+            t1.a == t2.a,
+            (t1.a != t2.b) | (t1.b != t2.a),
+            (t1.a != t2.b) ^ (t1.b != t2.a),
+            (t1.a != t2.b) & (t1.b != t2.a),
+            (t1.a + t1.a != t2.b) & (t1.b + t1.b != t2.a),
+        ],
+    )
+    expr = joined[t1]
+    with join_tables(t1, t2) as (r1, r2):
+        expected = ops.JoinChain(
+            first=r1,
+            rest=[
+                ops.JoinLink(
+                    "inner",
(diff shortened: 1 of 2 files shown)


ibis/expr/types/joins.py
```diff
@@ -81,15 +81,15 @@ def dereference_sides(left, right, deref_left, deref_right):
...
+def dereference_comparison_op(pred, deref_left, deref_right):
     left, right = dereference_sides(pred.left, pred.right, deref_left, deref_right)
     return pred.copy(left=left, right=right)
 
 
 def dereference_value(pred, deref_left, deref_right):
     deref_both = {**deref_left, **deref_right}
-    if isinstance(pred, ops.Binary) and pred.left.relations == pred.right.relations:
-        return dereference_binop(pred, deref_left, deref_right)
+    if isinstance(pred, ops.Comparison) and pred.left.relations == pred.right.relations:
```
kszucs, line 91: The point of this branch is to separately try to dereference each side of the predicate. This basically allows us to write

```py
t1 = ...

t1.join(t1, t1.a == t1.a)
```

kszucs: But this indeed should be applied only to comparisons not e.g. `&` `|` ...

cpcloud: Thanks for the explanation!

kszucs: Otherwise it tries to attempt to dereference all the values in the predicate to either sides. The test case works because of `t.view()` creates two distinct tables so the references in the predicates are not ambiguous.

ibis/expr/tests/test_newrels.py
```diff
@@ -1314,3 +1314,45 @@ def test_join_method_docstrings():
         join_method = getattr(joined, method)
         table_method = getattr(t1, method)
         assert join_method.__doc__ == table_method.__doc__
+
+
+def test_join_with_compound_predicate():
+    t1 = ibis.table(name="t", schema={"a": "string", "b": "string"})
+    t2 = t1.view()
```
kszucs, line 1321: Without this it should fail, but that is okay since we are unable to figure out the sides for nested predicates.

kszucs (approved):
Nice! Thanks for the fix!

ibis/expr/tests/test_newrels.py
cpcloud: Sure, but the goal here was to have a test case for the specific problem. The test will fail as written without the change to the deref algorithm

PR #8005, opened 2024-01-17
refactor(dask): port the dask backend to the new execution model

kszucs:
Reimplementation of the dask backend on top of the new pandas executor. I had to adjust the pandas backend to support extending. This way the new dask implementation turned out to be pretty tidy. 

There are a couple of features which are not implemented using proper dask constructs, but rather have a fallback to local execution using pandas. The most notable are the window functions. The previous dask implementation supported just a couple of window cases, but this way we have full coverage at least.

Thanks to the new pandas base we have a wider feature coverage, see the removed xfails in the test suite.


ibis/backends/dask/helpers.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def asseries(value, like=None):
+    """Ensure that value is a pandas Series object, broadcast if necessary."""
+
+    if isinstance(value, dd.Series):
+        return value
+    elif isinstance(value, dd.core.Scalar):
+        # Create a Dask array from the scalar
+        array = da.full((1,), value, chunks=(1,))
```
kszucs, line 34: @jcrist @ncclementi do you know whether this is triggering a `.compute()`? The point of this function would be to *broadcast* the input value as a `dd.Series`, preferably without triggerint compute.

ncclementi: It shouldn't. Are you seeing anything suspicious that makes you think otherwise? 
I did a quick check just in case: 
```python
In [2]: import dask.array as da
In [3]: import dask.dataframe as dd

In [4]: value = 5 #value is a scalar right?

In [5]: array = da.full((1,), value, chunks=(1,))

In [6]: array
Out[6]: dask.array<full_like, shape=(1,), dtype=int64, chunksize=(1,), chunktype=numpy.ndarray>
```

kszucs: Well, value is a dask scalar, not a plain python literal.

ncclementi: It should still be alright, here is an example with a dask scalar.
```python
In [13]: import dask

In [14]: ddf = dask.datasets.timeseries()

In [15]: ddf.x.mean()
Out[15]: dd.Scalar<series-..., dtype=float64>

In [16]: val = ddf.x.mean()

In [17]: array = da.full((1,), val, chunks=(1,))

In [18]: array
Out[18]: dask.array<full_like, shape=(1,), dtype=float64, chunksize=(1,), chunktype=numpy.ndarray>

In [19]: isinstance(val, dd.core.Scalar)
Out[19]: True
```

jcrist: That does compute (it's going through the `__array__` protocol, which is likely a bug/misfeature).

I think what you want here is:


```suggestion
        array = da.from_delayed(value.to_delayed(), (1,), dtype=value.dtype)
```

This is a bit of a roundabout way to cast a scalar to a single-partition series, but given the existing API I'm not sure there's a better way.

ncclementi: Oh interesting, I didn't see that coming. I was assuming the what's on Output line 18, implied this was lazy, bu I guess not. My bad.

kszucs: Thanks! Already incorporated your suggestion.

ibis/formats/pandas.py
```diff
@@ -263,7 +263,7 @@ def convert_Decimal(cls, s, dtype, pandas_type):
             scale=dtype.scale,
             strict=False,
         )
-        return s.map(func, na_action="ignore")
+        return s.map(func)
```
cpcloud, line 266: Don't we want to leave NA value alone?

ibis/backends/pandas/executor.py
```diff
@@ -110,19 +88,71 @@ def visit(cls, op: ops.RandomScalar):
...
+        # decimal operations have special implementations
+        if op.dtype.is_decimal():
+            func = cls.kernels.elementwise_decimal[typ]
+            return cls.elementwise(func, **kwargs)
+
+        # prefer generic implementations if available
+        if func := cls.kernels.generic.get(typ):
+            return cls.generic(func, **kwargs)
+
+        first, *rest = operands.values()
```
cpcloud, line 118: Looks like `first` might not be used.

```suggestion
        _, *rest = operands.values()
```

cpcloud (approved):
LGTM! Thanks for slogging through it!

ibis/formats/pandas.py
kszucs: This is to convert floating nan values to `Decimal("Nan")`

cpcloud: Ah, okay! Carry on, then.

ibis/backends/pandas/executor.py
kszucs: Updated.

ibis/backends/dask/executor.py
```diff
@@ -0,0 +1,414 @@
...
+
+        return agg
+
+    @classmethod
+    def visit(
+        cls, op: ops.ReductionVectorizedUDF, func, func_args, input_type, return_type
+    ):
+        def agg(df):
+            # if df is a dask dataframe then we collect it to a pandas dataframe
+            # because the user-defined function expects a pandas dataframe
```
ncclementi, line 228: If the UDF expects a pandas dataframe, shouldn't we let the user do the compute on their own and then pass it through? It looks to me that we don't support udfs for dask dataframes, and we are triggering a compute so it'll be a pandas df. If the dataframe is big, then this will break things, and the user might not understand why.

ibis/backends/dask/helpers.py
```diff
@@ -0,0 +1,186 @@
...
+            return value
+        elif isinstance(value, dd.core.Scalar):
+            # Create a Dask array from the Dask scalar
+            try:
+                dtype = value.dtype
+            except AttributeError:
+                #     @property
+                #     def dtype(self):
+                # >       return self._meta.dtype
+                # E       AttributeError: 'Timestamp' object has no attribute 'dtype'
```
ncclementi, line 42: Is this left over debugging code, or is there on purpose to show what triggers the `AttributeError`?

ibis/backends/dask/tests/test_cast.py
```diff
@@ -53,9 +52,14 @@ def test_cast_string(t, df, from_, to, expected):
 @pytest.mark.parametrize(
     ("to", "expected"),
     [
-        ("string", "object"),
+        pytest.param(
+            "string",
```
ncclementi, line 56: I think this raise will also apply to `"string[pyarrow]"`.

ibis/backends/pandas/tests/test_cast.py
```diff
@@ -67,7 +68,13 @@ def test_cast_array(t, from_, to, expected):
 @pytest.mark.parametrize(
     ("to", "expected"),
     [
-        ("string", "object"),
+        pytest.param(
+            "string",
```
ncclementi, line 72: I think here and few of the cases below it'll also apply for `"string[pyarrow]"`

ncclementi (commented):
I left some comments/questions, but none of them are blockers.

cpcloud:
Thanks for the review @ncclementi!

@jcrist Can you review this PR? It's blocking #8161.

ibis/backends/dask/executor.py
```diff
@@ -0,0 +1,414 @@
...
+
+    @classmethod
+    def visit(cls, op: ops.DummyTable, values):
+        df, _ = cls.asframe(values)
+        return df
+
+    @classmethod
+    def visit(cls, op: PandasLimit, parent, n, offset):
+        n = n.compute().iat[0, 0]
+        offset = offset.compute().iat[0, 0]
```
jcrist, line 295: You should be able to avoid the extra compute here by doing a `map_partitions` call of a function that applies the `between` operation below. This will avoid excess compute if the values of `n`/`offset` are dependent on anything in `df`.

Something like (untested):

```python
# define this at the top-level for cheaper pickling
def limit_chunk(df, col, n, offset):
    n = n.iat[0, 0]
    offset = offset.iat[0, 0]
    if n is None:
        return df[df[col] >= offset]
    else:
        return df[df[col].between(offset, offset + n - 1)]

# then in the visit function, something like:
return df.map_partitions(limit, col=name, n=n, offset=offset, align_partitions=False, meta=df._meta).drop(columns=[name])
```

ibis/backends/dask/helpers.py
```diff
@@ -0,0 +1,186 @@
...
+            array = da.from_delayed(value.to_delayed(), (1,), dtype=dtype)
+            # Create a Dask series from the Dask array
+            return dd.from_array(array)
+        elif isinstance(value, pd.Series):
+            return dd.from_pandas(value, npartitions=1)
+        elif like is not None:
+            if isinstance(value, (tuple, list, dict)):
+                fn = lambda df: pd.Series([value] * len(df), index=df.index)
+            else:
+                fn = lambda df: pd.Series(value, index=df.index)
```
jcrist, line 53: When possible, it's better to use top-level functions instead of lambdas for cheaper pickling of tasks. This is probably negligible here, but it's good to keep in mind.

ibis/backends/dask/executor.py
jcrist: I agree that computing here like this would be unexpected and not something we should do.

IIUC these are the legacy udfs. IMO we should remove these as part of this release. They were only supported by the pandas and dask backends, and given the rewrite changed a bunch of the internals (and the only major user of these we know of made _heavy_ use of the internals) I suspect there's no user that would be able to upgrade to this release that was also using the legacy udfs. We should drop them for them IMO for the new udfs alone.

ibis/backends/dask/executor.py
```diff
@@ -0,0 +1,414 @@
...
+                mask = df[where.name]
+                lhs = df[left.name][mask].compute()
+                rhs = df[right.name][mask].compute()
+                return lhs.corr(rhs)
+
+        return agg
+
+    @classmethod
+    def visit(cls, op: ops.Covariance, left, right, where, how):
+        # TODO(kszucs): raise a warning about triggering compute()?
```
jcrist, line 204: dask has built in covariance and correlation methods, why are they not sufficient here/why do we need to compute?

IMO anything that needs to compute like this shouldn't warn, we should just not support it unless we can do it cleanly.

ibis/backends/dask/executor.py
```diff
@@ -0,0 +1,414 @@
...
+        result = df.map_partitions(mapper)
+        if op.dtype.is_struct():
+            result = result.apply(lambda row: row.to_dict(), axis=1)
+        return result
+
+    ############################# Reductions ##################################
+
+    @classmethod
+    def visit(cls, op: ops.ArgMin | ops.ArgMax, arg, key, where):
+        # TODO(kszucs): raise a warning about triggering compute()?
```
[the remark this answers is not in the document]
jcrist: I think we should just error as unsupported in any case where an explicit `.compute()` is needed. Dask has builtin ways of spelling these operations without compute, so if we can't use them now given the current architecture we should just not support them until we can rearchitect to call them appropriately.

jcrist (commented):
Just a few comments. Overall I'd be very happy to get this in and then iterate. This is _much_ cleaner than the current dask backend, thanks for all the work here!

I do think many of these places where we're explicitly calling `.compute()` as part of handling an operation are places where we should just drop support for that operation until we can spell it in a more efficient way (for many of these dask itself can do so efficiently, so the limit is more on our executor design than on dask). I'd rather have an `UnsupportedOperationError` for now than something that is made inefficient by something `ibis` is doing rather than the backend itself.

ibis/backends/dask/executor.py
```diff
@@ -0,0 +1,414 @@
...
+        if n is None:
+            df = df[df[name] >= offset]
+        else:
+            df = df[df[name].between(offset, offset + n - 1)]
+
+        return df.drop(columns=[name])
+
+    @classmethod
+    def visit(cls, op: PandasResetIndex, parent):
+        return add_globally_consecutive_column(parent)
```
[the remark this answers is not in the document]
kszucs: I need this to concatenate the window results, after joins and for offset. I agree that we should avoid where we can, going to create a follow-up issue.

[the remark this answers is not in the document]
kszucs: https://github.com/ibis-project/ibis/issues/8185

ibis/backends/dask/executor.py
kszucs: Created a follow-up https://github.com/ibis-project/ibis/issues/8186

ibis/backends/dask/executor.py
kszucs: I first tried it without calling compute, but with the masking an two different dask series I've got some error I can't recall now. Deferring to a follow-up.

kszucs: https://github.com/ibis-project/ibis/issues/8187

ibis/backends/dask/executor.py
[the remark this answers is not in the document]
kszucs: Follow-up issue https://github.com/ibis-project/ibis/issues/8187

cpcloud:
You can xfail `test_unnest_range` for now, it's a new test since the last rebase of `the-epic-split`.

.github/workflows/ibis-backends.yml
```diff
@@ -418,13 +418,13 @@ jobs:
...
-          #           title: Dask
-          #           deps:
-          #             - "dask[array,dataframe]@2022.9.1"
-          #             - "pandas@1.5.3"
-          #           extras:
-          #             - dask
+          - name: dask
+            title: Dask
+            deps:
+              - "dask[array,dataframe]@2022.9.1"
```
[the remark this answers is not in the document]
kszucs: I don't know, just kept the previous pin.

ibis/backends/dask/convert.py
```diff
@@ -0,0 +1,83 @@
...
+        elif pdt.is_numeric_dtype(s.dtype):
+            s = dd.to_datetime(s, unit="D")
+        else:
+            s = dd.to_datetime(s)
+
+        return s.dt.normalize()
+
+    @classmethod
+    def convert_String(cls, s, dtype, pandas_type):
+        # TODO(kszucs): should switch to the new pandas string type and convert
```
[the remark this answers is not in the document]
kszucs: Yeah, not even mentioning the consistent na handling :)

ibis/backends/dask/helpers.py
kszucs: The latter, going to leave it here for now so someone can fix it.

ibis/backends/dask/tests/test_cast.py
kszucs: I would assume so too.

ibis/backends/dask/helpers.py
kszucs: Created a follow-up: https://github.com/ibis-project/ibis/issues/8191

kszucs:
> I do think many of these places where we're explicitly calling .compute() as part of handling an operation are places where we should just drop support for that operation until we can spell it in a more efficient way (for many of these dask itself can do so efficiently, so the limit is more on our executor design than on dask). I'd rather have an UnsupportedOperationError for now than something that is made inefficient by something ibis is doing rather than the backend itself.

On one hand I agree, simply because it should be the ideal way to go. On the other hand at least these operations are usable, even though they are not as performant as they should be. Also some of the window functionality used to be supported, so raising would cause a regression. There is a third option is to raise a warning so that the user is aware about the performance problems and gradually improve them (if there is interest).

jcrist:
> On the other hand at least these operations are usable, even though they are not as performant as they should be. Also some of the window functionality used to be supported, so raising would cause a regression. 

I'd argue that the previous dask backend was both buggy and slow enough that no one is probably using it. A regression in functionality with an improvement in correctness of dask code (including not calling `.compute()` multiple times) feels worth it to me. I consider anywhere we're manually calling `compute` outside of returning results to the user a bug.

PR #8008, opened 2024-01-17
ci: migrate to setup-python dependency caching

cpcloud:
Apparently these actions are archived and functionality is rolled into setup-python. This PR attempts to validate that.

### DIFF
--- a/.github/workflows/check-generated-files.yml
+++ b/.github/workflows/check-generated-files.yml
@@ -26,20 +26,21 @@ jobs:
       - name: checkout
         uses: actions/checkout@v4
 
+      - name: install poetry
+        run: pipx install 'poetry==1.7.1'
+
       - name: install python
         uses: actions/setup-python@v5
         id: install_python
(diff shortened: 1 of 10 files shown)


cpcloud:
It looks like because we were caching dependencies in the sqlalchemy2 jobs, we were incorrectly passing some tests and failing to expose some incomplete skipping in duckdb geospatial testing. I've fixed those issues in this PR.

cpcloud:
Since we're moving away from sqlalchemy entirely (and thus sqlalchemy2 will no longer be a thing we have to deal with 🎉) I opted to avoid the complexity of generating a single lock file and downloading that file in dependent jobs. The sqlalchemy2 CI will be a bit slower until we land the port to sqlglot.

cpcloud:
Well, I guess it's not even that much slower to avoid generating the lockfile, even better!

gforsyth (approved):
:shipit:

PR #8014, opened 2024-01-18
fix(mutate/select): ensure that unsplatted dictionaries work in `mutate`and`select` APIs

### LINKED ISSUE #8013: bug: incorrect SQL generated on main (but not TES) for generating random numbers
### What happened?

I have this old code for generating a billion random numbers w/ an ugly SQL string:

```python
import ibis

con = ibis.connect("duckdb://data.ddb")

ROWS = 1_000_000_000

sql_str = ""
sql_str += "select\n"
for c in list(map(chr, range(ord("a"), ord("z") + 1))):
    sql_str += f"  random() as {c},\n"
sql_str += f"from generate_series(1, {ROWS})"

t = con.sql(sql_str)
con.create_table("billion", t, overwrite=True)
```

I tried to translate it to Ibis expressions as:

```python
t = (
    ibis.range(10)
    .unnest()
    .name("index")
    .as_table()
    .mutate({c: ibis.random() for c in list(map(chr, range(ord("a"), ord("z") + 1)))})
)
```

resulting in:

```html
┏━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┓
┃ index ┃ 'a'    ┃ 'b'    ┃ 'c'    ┃ 'd'    ┃ 'e'    ┃ 'f'    ┃ 'g'    ┃ 'h'    ┃ 'i'    ┃ 'j'    ┃ 'k'    ┃ 'l'    ┃ 'm'    ┃ 'n'    ┃ 'o'    ┃ 'p'    ┃ 'q'    ┃ 'r'    ┃ 's'    ┃ 't'    ┃ 'u'    ┃ 'v'    ┃ 'w'    ┃ 'x'    ┃ 'y'    ┃ 'z'    ┃
┡━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━┩
│ int8  │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │ string │
├───────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┼────────┤
│     0 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
│     1 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
│     2 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
│     3 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
│     4 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
│     5 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
│     6 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
│     7 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
│     8 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
│     9 │ a      │ b      │ c      │ d      │ e      │ f      │ g      │ h      │ i      │ j      │ k      │ l      │ m      │ n      │ o      │ p      │ q      │ r      │ s      │ t      │ u      │ v      │ w      │ x      │ y      │ z      │
└───────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┴────────┘
```

this is the SQL it's generating:

```sql
sql
SELECT
  t0.index,
  'a' AS "'a'",
  'b' AS "'b'",
  'c' AS "'c'",
  'd' AS "'d'",
  'e' AS "'e'",
  'f' AS "'f'",
  'g' AS "'g'",
  'h' AS "'h'",
  'i' AS "'i'",
  'j' AS "'j'",
  'k' AS "'k'",
  'l' AS "'l'",
  'm' AS "'m'",
  'n' AS "'n'",
  'o' AS "'o'",
  'p' AS "'p'",
  'q' AS "'q'",
  'r' AS "'r'",
  's' AS "'s'",
  't' AS "'t'",
  'u' AS "'u'",
  'v' AS "'v'",
  'w' AS "'w'",
  'x' AS "'x'",
  'y' AS "'y'",
  'z' AS "'z'"
FROM (
  SELECT
    UNNEST(RANGE(CAST(0 AS TINYINT), CAST(10 AS TINYINT), CAST(1 AS TINYINT))) AS index
) AS t0
```

confirmed by Phillip the same is working on `the-epic-split` branch. planning to show my old vs new code in #8004 

main

DuckDB

### Relevant log output

```sh
n/a
```

### Code of Conduct

cpcloud:
We were not handling unsplatted dicts in mutate/select, this PR fixes that. Fixes #8013.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -1800,17 +1800,24 @@ class Table(Expr, _FixedTextJupyterMixin):
         import ibis.expr.analysis as an
 
         exprs = [] if exprs is None else util.promote_list(exprs)
-        exprs = itertools.chain(
-            itertools.chain.from_iterable(
-                util.promote_list(_ensure_expr(self, expr)) for expr in exprs
-            ),
-            (
-                e.name(name)
-                for name, expr in mutations.items()
-                for e in util.promote_list(_ensure_expr(self, expr))
-            ),
+
+        new_exprs = []
+
+        for expr in exprs:
+            if isinstance(expr, Mapping):
+                new_exprs.extend(
+                    _ensure_expr(self, val).name(name) for name, val in expr.items()
+                )
+            else:
+                new_exprs.extend(util.promote_list(_ensure_expr(self, expr)))
+
+        new_exprs.extend(
+            e.name(name)
+            for name, expr in mutations.items()
+            for e in util.promote_list(_ensure_expr(self, expr))
         )
-        mutation_exprs = an.get_mutation_exprs(list(exprs), self)
+
+        mutation_exprs = an.get_mutation_exprs(new_exprs, self)
         return self.select(mutation_exprs)
 
     def select(
@@ -1993,31 +2000,34 @@ class Table(Expr, _FixedTextJupyterMixin):
         import ibis.expr.analysis as an
         from ibis.selectors import Selector
 
-        exprs = [
-            e
-            for expr in exprs
-            for e in (
-                expr.expand(self)
-                if isinstance(expr, Selector)
-                else map(self._ensure_expr, util.promote_list(expr))
-            )
-        ]
-        exprs.extend(
+        new_exprs = []
+
+        for expr in exprs:
+            if isinstance(expr, Selector):
+                new_exprs.extend(expr.expand(self))
+            elif isinstance(expr, Mapping):
+                new_exprs.extend(
+                    _ensure_expr(self, value).name(name) for name, value in expr.items()
+                )
+            else:
+                new_exprs.extend(map(self._ensure_expr, util.promote_list(expr)))
+
+        new_exprs.extend(
             self._ensure_expr(expr).name(name) for name, expr in named_exprs.items()
         )
 
-        if not exprs:
+        if not new_exprs:
             raise com.IbisTypeError(
                 "You must select at least one column for a valid projection"
             )
-        for ex in exprs:
+        for ex in new_exprs:
             if not isinstance(ex, Expr):
                 raise com.IbisTypeError(
                     "All arguments to `.select` must be coerceable to "
                     f"expressions - got {type(ex)!r}"
                 )
 
-        op = an.Projector(self, exprs).get_result()
+        op = an.Projector(self, new_exprs).get_result()
 
         return op.to_expr()
 
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -2007,7 +2007,7 @@ class Table(Expr, _FixedTextJupyterMixin):
                 new_exprs.extend(expr.expand(self))
             elif isinstance(expr, Mapping):
                 new_exprs.extend(
-                    _ensure_expr(self, value).name(name) for name, value in expr.items()
+                    self._ensure_expr(value).name(name) for name, value in expr.items()
                 )
             else:
                 new_exprs.extend(map(self._ensure_expr, util.promote_list(expr)))
--- a/ibis/backends/tests/test_array.py
+++ b/ibis/backends/tests/test_array.py
@@ -1066,7 +1066,7 @@ def test_repr_timestamp_array(con, monkeypatch):
     raises=com.OperationNotDefinedError,
 )
 def test_unnest_range(con):
-    expr = ibis.range(2).unnest().name("x").as_table().mutate({"y": ibis.literal(1.0)})
+    expr = ibis.range(2).unnest().name("x").as_table().mutate({"y": 1.0})
     result = con.execute(expr)
     expected = pd.DataFrame({"x": np.array([0, 1], dtype="int8"), "y": [1.0, 1.0]})
     tm.assert_frame_equal(result, expected)
--- a/ibis/backends/tests/test_array.py
+++ b/ibis/backends/tests/test_array.py
@@ -1059,3 +1059,14 @@ def test_repr_timestamp_array(con, monkeypatch):
     assert ibis.options.default_backend is con
     expr = ibis.array(pd.date_range("2010-01-01", "2010-01-03", freq="D").tolist())
     assert repr(expr)
+
+
+@pytest.mark.notyet(
+    ["dask", "datafusion", "flink", "pandas", "polars"],
+    raises=com.OperationNotDefinedError,
+)
+def test_unnest_range(con):
+    expr = ibis.range(2).unnest().name("x").as_table().mutate({"y": ibis.literal(1.0)})
+    result = con.execute(expr)
+    expected = pd.DataFrame({"x": np.array([0, 1], dtype="int8"), "y": [1.0, 1.0]})
+    tm.assert_frame_equal(result, expected)
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -1741,3 +1741,16 @@ def test_simple_memtable_construct(con):
     # because memtables have a unique name per table per process, so smoke test
     # it
     assert str(ibis.to_sql(expr, dialect=con.name)).startswith("SELECT")
+
+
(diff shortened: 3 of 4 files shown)


ibis/expr/types/relations.py
```diff
@@ -1993,31 +2000,34 @@ def select(
...
-        ]
-        exprs.extend(
+        new_exprs = []
+
+        for expr in exprs:
+            if isinstance(expr, Selector):
+                new_exprs.extend(expr.expand(self))
+            elif isinstance(expr, Mapping):
+                new_exprs.extend(
+                    _ensure_expr(self, value).name(name) for name, value in expr.items()
```
jcrist, line 2010: Note that `self._ensure_expr` and `_ensure_expr` differ in how they treat str/int indices. Previously we were always using `self._ensure_expr` in this method, now this uses a mix. We should probably standardize on one here (or add a comment if the discrepancy is intentional).

```suggestion
                    self._ensure_expr(value).name(name) for name, value in expr.items()
```

ibis/expr/types/relations.py
```diff
@@ -1993,31 +2000,34 @@ def select(
...
-                if isinstance(expr, Selector)
-                else map(self._ensure_expr, util.promote_list(expr))
-            )
-        ]
-        exprs.extend(
+        new_exprs = []
+
+        for expr in exprs:
+            if isinstance(expr, Selector):
+                new_exprs.extend(expr.expand(self))
```
jcrist, line 2007: `_ensure_expr` handles selector expansion, I *think* this case could be dropped and handled by the not-mapping branch below?

jcrist (commented):
Overall concept looks good to me, just a couple code nits/questions.

ibis/expr/types/relations.py
cpcloud: This won't work. `_ensure_expr(self, value)` handles the case of a literal, while `self._ensure_expr` does not.

This code is basically gone in `the-epic-split` FWIW.

cpcloud: We might be able to move entirely over to the function form, lemme check.

cpcloud: Using `_ensure_check(self, expr)` everywhere breaks a bunch of tests.

ibis/expr/types/relations.py
cpcloud: Unfortunately it won't work because `expr.expand(self)` returns a `list` and so `map(self._ensure_expr, ...)` returns an iterable of `list`s and that isn't supported by `Projector`.

ibis/expr/types/relations.py
jcrist: > This won't work. `_ensure_expr(self, value)` handles the case of a literal, while `self._ensure_expr` does not.

Right. I thought we didn't want to support auto-broadcasting literals in `t.select`, only in `t.mutate`, hence the difference in the implementations here. `t.select("fizz")` should error if there's no column `fizz` in `t`, not broadcast a new column with `"fizz"` in every row (where `mutate` would do that). Following that, I don't think we want to support literal values in a mapping passed to `t.select` either. IMO how we treat the values in the mapping should be the same as how we treat the values in `**kwargs` or `*args`.

cpcloud: Ah right.

What about the floating point case? Do we care? for `str`/`int` they are ambiguous with column references, but for other `float` that's not the case.

I guess if we make lack of literal support there consistent we should perhaps not allow floating point values there either?

cpcloud: There's other values we could (not) allow too, like `Decimal`s.

cpcloud: The main problem now is making a reasonable error message without checking for every possible thing under the sun. I suppose I could catch the attribute error when `.name` is called, but that seems less than ideal.

cpcloud: Ok, I reverted back to using `self._ensure_expr`-only in `select`.

Perhaps we can live with the `AttributeError` that's thrown when passing a Python literal value like `1.0` for now?

jcrist: > I guess if we make lack of literal support there consistent we should perhaps not allow floating point values there either?

We already don't, `t.select` currently won't coerce any literals to expressions. If you pass it as an arg (`t.select(1.5)`) you get a nice error, if you do it as a kwarg (`t.select(foo=1.5)`) you get an `AttributeError` for `'name'`. I guess for now just use `self._ensure_expr` and it doesn't make anything worse, and is at least consistent? As you said things are different already in TES branch.

cpcloud: SGTM, resolving.

jcrist (approved):

PR #8020, opened 2024-01-18
refactor(oracle): port to sqlglot

gforsyth:


ibis/backends/tests/test_client.py
```diff
@@ -1379,6 +1379,10 @@ def test_persist_expression_repeated_cache(alltypes):
     reason="mssql supports support temporary tables through naming conventions",
 )
 @mark.notimpl(["exasol"], reason="Exasol does not support temporary tables")
+@mark.broken(
+    ["oracle"],
+    reason="Among other reasons, oracles errors don't inherit from Exception",
+)
```
cpcloud, line 1385: Oh boy.

Was this working before (and xfailing)?

ibis/backends/tests/test_string.py
```diff
@@ -966,25 +965,22 @@ def test_levenshtein(con, right):
...
-    raises=sa.exc.DatabaseError,
-)
 @pytest.mark.parametrize(
     "expr",
     [
-        param(ibis.case().when(True, "%").end(), id="case"),
+        param(
+            ibis.case().when(True, "%").end(),
+            id="case",
+            marks=[],
```
cpcloud, line 974: Can this change be reverted since `marks` is empty?

ibis/backends/oracle/__init__.py
```diff
@@ -2,87 +2,43 @@
...
-# Get out your clippers, it's time to shave a yak.
-#
-# 1. oracledb is only supported in sqlalchemy 2.0
-# 2. Ergo, module hacking is required to avoid doing a silly amount of work
-#    to create multiple lockfiles or port snowflake away from sqlalchemy
-# 3. Also the version needs to be spoofed to be >= 7 or else the cx_Oracle
-#    dialect barfs
-oracledb.__version__ = oracledb.version = "7"
-
-sys.modules["cx_Oracle"] = oracledb
```
cpcloud, line 26: Yaaas

ibis/backends/oracle/__init__.py
```diff
@@ -211,6 +387,15 @@ def _metadata(self, query: str) -> Iterable[tuple[str, dt.DataType]]:
         if isinstance(sg_expr, sg.exp.Table):
             sg_expr = sg.select(STAR).from_(sg_expr)
 
+        def transformer(node):
+            if isinstance(node, sg.exp.Table):
+                return sg.table(node.name, quoted=True)
+            elif isinstance(node, sg.exp.Column):
+                return sg.column(col=node.name, quoted=True)
```
cpcloud, line 394: Oooh, I hadn't really thought about this but I suspect we should probably apply this transform in general to backends whose `quoted` value is `True`.

cpcloud: Anyway, follow-up fodder.

ibis/backends/oracle/__init__.py
```diff
@@ -269,37 +455,42 @@ def _metadata(self, query: str) -> Iterable[tuple[str, dt.DataType]]:
...
+                cursor, columns=schema.names, coerce_float=True
+            )
+        except Exception:
+            # clean up the cursor if we fail to create the DataFrame
+            #
+            # in the sqlite case failing to close the cursor results in
+            # artificially locked tables
+            cursor.close()
+            raise
+        df = OraclePandasData.convert_table(df, schema)
```
cpcloud, line 479: We have this method in many of the backends. Nothing to do here except note for a follow up to perhaps generalize a bit.

ibis/backends/oracle/compiler.py
```diff
@@ -0,0 +1,604 @@
...
+    #                                     self.f.concat(
+    #                                         self.f.lpad(
+    #                                             self.cast(minutes, dt.string), 2, "0"
+    #                                         ),
+    #                                         self.f.concat(
+    #                                             ":",
+    #                                             self.f.lpad(
+    #                                                 self.cast(seconds, dt.string),
+    #                                                 2,
+    #                                                 "0",
```
cpcloud, line 265: Monstrous.

ibis/backends/oracle/compiler.py
```diff
@@ -0,0 +1,604 @@
...
+            "us": "us",
+        }
+
+        if (unit := unit_mapping.get(unit.short)) in ("ms", "us", None):
+            raise com.UnsupportedOperationError(f"Unsupported truncate unit {unit}")
+
+        # The FuncGen for self.f.extract always resolves to sge.Extract
+        # and then complains about the missing expression argument
+        # so we construct it directly
+        return sge.Extract(this=unit, expression=arg)
```
cpcloud, line 453: I see your comment, but I think perhaps you may have specified the order incorrectly. This should be doable as:

```python
self.f.extract(self.v[unit], arg)
```

`self.v[name]` constructs `sge.Var(this=name)` which is the sqlglot-ism for the `YEAR` syntax in `EXTRACT(YEAR FROM my_column)`.

Also, is `extract` the right operation here? This should be truncating a date/time/timestamp, not returning one of its components.

cpcloud (changes requested):
Generally speaking looks great 👌🏻 

Just a few questions.

I will look into the impala/druid failures since they seem to be consistent (but are passing on `upstream/the-epic-split`)

ibis/backends/tests/test_client.py
gforsyth: hmm, I thought I removed that comment, must have reapplied it during a rebase.  I think I'm wrong about this.

ibis/backends/oracle/compiler.py
gforsyth: It felt bad writing it and it feels bad looking at it

ibis/backends/oracle/compiler.py
gforsyth: it is not, should be `trunc`, at least for the datetimes

ibis/backends/tests/test_client.py
gforsyth: this was working before because sqlalchemy intercepted the oracle error and included the name of the missing table/view.  `oracledb` raises an error that simply states that the table/view is unknown but doesn't include the name of the table the user tried to access.  for now, just marking as `notimpl` until we decide if want to wrap and reraise with more info.

ibis/backends/oracle/compiler.py
```diff
@@ -0,0 +1,617 @@
...
+
+def _datatype_sql(self: Oracle.Generator, expression: sge.DataType) -> str:
+    # Use this to handle correctly formatting timestamp precision
+    # e.g. TIMESTAMP (scale) WITH TIME ZONE vs. TIMESTAMP WITH TIME ZONE(scale)
+    if expression.is_type("timestamptz"):
+        for exp in expression.expressions:
+            if isinstance(exp, sge.DataTypeParam):
+                return f"TIMESTAMP ({self.sql(exp, 'this')}) WITH TIME ZONE"
+        return "TIMESTAMP WITH TIME ZONE"
+    return self.datatype_sql(expression)
```
gforsyth, line 51: I don't know if there's a better way to do this, but the regular mapping in `Oracle.Generator.TYPE_MAPPING` didn't expose the scale parameter

gforsyth:
Ok, I have this down to 3 failing tests locally.  I've been staring at it for long enough that I'm not going to get anywhere else today.  I'll pick it back up on Monday.

ibis/backends/oracle/compiler.py
cpcloud: At some point we should upstream the bug report to sqlglot

cpcloud (approved):
Assuming the tests eventually pass, this LGTM!

Epic!

cpcloud:
NICE

ibis/backends/oracle/compiler.py
gforsyth: https://github.com/tobymao/sqlglot/issues/2879

PR #8021, opened 2024-01-18
fix(docs): surround executable code blocks with interactive mode on/off

### LINKED ISSUE #8001: docs: interactive mode isn't persisted in some examples
### Please describe the issue

I'm looking at https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.pivot_longer. The first `>>>` block is rendered in interactive mode. The rest are not, but should be.

This is caused by https://github.com/ibis-project/ibis/commit/502b88cb0c86d33b1bd6fdd0b35a6d19f283268b. I couldn't find the PR or other info on why that commit was motivated.

Should we go through these cases and turn on interactive mode before every example? This seems pretty boilerplatey. Perhaps better would be to add a `quartodoc: +INTERACTIVE` pragma or something to each docstring, similar to `quartodoc: +SKIP`? Not sure what the best semantics there should be.

### Code of Conduct

cpcloud:
Fixes #8001. For each example were disabling interactive mode before rendering, which caused subsequent examples in the same docstring to use non-interactive rendering. The fix is to surround each code block with turning interactive mode on just before the block and off right after.

### DIFF
--- a/docs/_renderer.py
+++ b/docs/_renderer.py
@@ -28,8 +28,6 @@ class Renderer(qd.MdRenderer):
             lambda line: quartodoc_skip_doctest in line or skip_doctest in line
         )
 
-        has_executed_chunks = False
-
         for chunk in toolz.partitionby(chunker, lines):
             first, *rest = chunk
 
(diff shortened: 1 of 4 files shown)


docs/_renderer.py
```diff
@@ -67,22 +76,18 @@ def render(self, el: qd.ast.ExampleCode) -> str:
...
-        else:
-            return examples
+                if not skipped:
+                    result.append(
+                        dedent(
+                            """
+                            ```{python}
+                            #| echo: false
+
+                            import ibis
```
NickCrews, line 86: can drop the import?

NickCrews:
preview

cpcloud:
> preview

This no longer works, it's now restricted to people who have the ability to assign issue labels.

docs/_renderer.py
cpcloud: I suppose.

cpcloud: Done

cpcloud:
Still a couple non-interactive examples to fix.

cpcloud:
Ok, this is ready now.

kszucs (approved):

NickCrews:
Does this mean that any >>> ibis.options.interactive = True lines in the actual dpcstring are now redundant, because interactive mode is already on?

cpcloud:
> Does this mean that any >>> ibis.options.interactive = True lines in the actual ocstring are now redundant, because interactive mode is already on? Should we remove them?

No, because those blocks still run as doctests which is not related to quarto rendering.

NickCrews:
ahh, those two different execution modes are confusing to me. Sounds good makes sense!

Issue #8022, opened 2024-01-18
bug: `ibis.array([ibis.literal(1), ibis.literal(2)])` returns ArrayColumn, not ArrayScalar


NickCrews:
### What happened?

as it says. PR incoming.

### What version of ibis are you using?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

NickCrews:
Yay thank you @cpcloud and @kszucs !

kszucs:
Thanks @NickCrews for both the report and the help!

PR #8023, opened 2024-01-18
Tweak ibis.array() docstring, improve test, add xfail test

NickCrews:
adds an xfail test for https://github.com/ibis-project/ibis/issues/8022

I discovered that actually fixing the problem wasn't trivial, as I wasn't sure about the ibis internals and how to create an ArrayScalar object correctly. But still I think this is an improvement.

Not totally in love with how I patched ibis.options.repr.interactive.max_length, open to suggestions.

Also discovered that pandas and dask can't handle mixed scalars and columns, so I fixed that.

### DIFF
--- a/ibis/backends/dask/execution/arrays.py
+++ b/ibis/backends/dask/execution/arrays.py
@@ -6,6 +6,7 @@ from functools import partial
 import dask.dataframe as dd
 import dask.dataframe.groupby as ddgb
 import numpy as np
+import pandas as pd
 
 import ibis.expr.operations as ops
 from ibis.backends.dask.core import execute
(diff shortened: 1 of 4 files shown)


cpcloud:
I think the abstraction is a bit off for construction of array literals. We shouldn't be constructing an `ArrayColumn` simply because the inputs to `array` are expressions, but instead do something like we're doing with `map` and `struct`: use the **shape** of the inputs to determine the output shape of the literal value.

cpcloud:
@NickCrews If you don't mind I'm going to cherry pick the first two commits here and apply them to #8049!

NickCrews:
you're right, using .shape makes way more sense.

PR #8032, opened 2024-01-19
refactor(exasol): port to sqlglot

cpcloud:
This PR ports exasol to sqlglot instead of sqlalchemy.


pyproject.toml
```diff
@@ -77,6 +77,7 @@ polars = { version = ">=0.19.3,<1", optional = true }
 psycopg2 = { version = ">=2.8.4,<3", optional = true }
 pydata-google-auth = { version = ">=1.4.0,<2", optional = true }
 pydruid = { version = ">=0.6.5,<1", optional = true }
+pyexasol = { version = ">=0.25.2,<1", optional = true, extras = ["pandas"] }
```
cpcloud, line 80: Not 100% convinced we should bring in this extra. There's a pretty sizable hit from all the CSV conversion (despite being a thing that ostensibly helps performance).

ibis/backends/tests/test_window.py
```diff
@@ -1104,6 +1111,11 @@ def test_first_last(backend):
     reason="not support by the backend",
 )
 @pytest.mark.broken(["flink"], raises=Py4JJavaError, reason="bug in Flink")
+@pytest.mark.broken(
+    ["exasol"],
+    raises=ExaQueryError,
+    reason="database can't handle UTC timestamps in DataFrames",
```
cpcloud, line 1117: This might be fixable if we get rid of `import_from_pandas` and do the inserts ourselves.

pyproject.toml
cpcloud: I removed the hit from `export_to_pandas` (by doing the conversion ourselves).

I'm going to keep the extra for use with `import_from_pandas` which simplifies inserting into in-memory tables.

ibis/backends/tests/test_aggregation.py
```diff
@@ -1524,8 +1520,9 @@ def test_grouped_case(backend, con):
 
 
 @pytest.mark.notimpl(
-    ["datafusion", "mssql", "polars", "exasol"], raises=com.OperationNotDefinedError
+    ["datafusion", "mssql", "polars"], raises=com.OperationNotDefinedError
 )
+@pytest.mark.notimpl(["exasol"], raises=ExaQueryError)
```
kszucs, line 1525: Just a nit and unrelated to this pr is that the exception name `ExaQueryError` is not consistent with the rest of the error types having full backend name as the prefix, so I think it should be `ExasolQueryError`

ibis/backends/tests/test_dot_sql.py
```diff
@@ -45,17 +46,20 @@
...
 )
 def test_con_dot_sql(backend, con, schema):
-    alltypes = con.table("functional_alltypes")
+    alltypes = backend.functional_alltypes
     # pull out the quoted name
-    name = _NAMES.get(con.name, alltypes.op().name)
+    name = _NAMES.get(con.name, "functional_alltypes")
+    quoted = getattr(getattr(con, "compiler", None), "quoted", True)
+    dialect = _IBIS_TO_SQLGLOT_DIALECT.get(con.name, con.name)
+    cols = [
```
kszucs, line 54: Required due to backend specific quoting?

ibis/backends/tests/test_export.py
```diff
@@ -214,7 +213,9 @@ def test_table_to_parquet(tmp_path, backend, awards_players):
 
     df = pd.read_parquet(outparquet)
 
-    backend.assert_frame_equal(awards_players.to_pandas(), df)
+    backend.assert_frame_equal(
+        awards_players.to_pandas().fillna(pd.NA), df.fillna(pd.NA)
```
kszucs, line 217: Blaah.

ibis/backends/tests/test_dot_sql.py
cpcloud: Yes. This effectively makes identifiers case sensitive

ibis/backends/tests/test_export.py
cpcloud: Phase 2 after the port?? 😂

kszucs: Probably, we should clean up the output coercions.

kszucs (approved):
Nice!

PR #8040, opened 2024-01-19
feat(examples): add `zones` geojson example

cpcloud:
Adds a `zones` geospatial data example


ibis/examples/__init__.py
```diff
@@ -126,6 +126,33 @@ def __dir__() -> list[str]:
...
+        if table_name is None:
+            table_name = name
+
+        board = _get_board()
+
+        (path,) = board.pin_download(name)
+        return backend.read_geo(path)
+
+
+zones = Zones("zones", help="Taxi zones in New York City")
```
ncclementi, line 153: Ugh, I just noticed that this data is on [EPSG:2263](https://spatialreference.org/ref/epsg/2263/) which has units of feet. 
```suggestion
zones = Zones("zones", help="Taxi zones in New York City (EPSG:2263)")
```

ncclementi:
We should add a test that looks something like this: 
```python
def test_load_geo_example():
    pytest.importorskip("duckdb")
    
    con = ibis.duckdb.connect()
    
    t = getattr(ibis.examples, "zones").fetch(backend=con)
    assert t.columns == ['zone', 'LocationID', 'borough', 'x_cent', 'y_cent', 'geometry']
```

ncclementi (approved):
Thanks @cpcloud this LGTM!

PR #8041, opened 2024-01-19
feat(sql): extract common table expressions

kszucs:
Regenerated the snapshots for clickhouse and duckb. 

~~The aliasing has changed becuase they are generated in BFS order rather than the previous TOPO order, I can restore that.~~


kszucs:
@cpcloud could you please regenerate the snapshots for impala?

cpcloud (approved):
LGTM, great work!

PR #8046, opened 2024-01-20
fix(exasol): use slower but more reliable ddl and allow non-strict xfail on window function test

cpcloud:
- chore(exasol): avoid complex websocket callback for inserting memtables
- test(exasol): account for unordered results in window function tests


cpcloud:
Running the entire `exasol` test suite with `--count=10` I am unable to reproduce the previous failures, whereas without this PR they reproduce fairly quickly (though not always on the first run).

kszucs (approved):

PR #8048, opened 2024-01-20
fix(duckdb): ensure that casting to floating point values produces valid types in generated sql

cpcloud:
Fixes a bug when compiling expressions to DuckDB sql. Discussed [in Zulip](https://ibis-project.zulipchat.com/#narrow/stream/405265-tech-support/topic/Issue.20with.20casting.20and.20sql.20genertion.20in.20duckdb).

### DIFF
--- a/ibis/backends/duckdb/tests/snapshots/test_datatypes/test_cast_to_floating_point_type/float32/out.sql
+++ b/ibis/backends/duckdb/tests/snapshots/test_datatypes/test_cast_to_floating_point_type/float32/out.sql
@@ -1,0 +1,2 @@
+SELECT
+  CAST('1.0' AS REAL) AS "Cast('1.0', float32)"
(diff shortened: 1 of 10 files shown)


ibis/backends/duckdb/tests/snapshots/test_datatypes/test_cast_to_floating_point_type/float32/out.sql
```diff
@@ -0,0 +1,2 @@
+SELECT
+  CAST('1.0' AS REAL) AS "Cast('1.0', float32)"
```
cpcloud, line 2: This is produced from `sat.REAL` which compiles to `REAL` and is a valid DuckDB type. Making this return `FLOAT` isn't really worth the additional code since it'll be moot after the port to sqlglot.

kszucs (approved):

PR #8049, opened 2024-01-20
fix(datatypes): ensure that array construction supports literals and infers their shape from its inputs

### LINKED ISSUE #8022: bug: `ibis.array([ibis.literal(1), ibis.literal(2)])` returns ArrayColumn, not ArrayScalar

cpcloud:
We were previously returning `ArrayColumn` from `ibis.array` when any inputs were expressions regardless of their shape. This PR renames `ArrayColumn` -> `Array` and uses the input arguments shapes to determine the output array shape.

Fixes #8022.

### DIFF
--- a/ibis/backends/dask/execution/arrays.py
+++ b/ibis/backends/dask/execution/arrays.py
@@ -38,7 +38,11 @@ collect_list = dd.Aggregation(
 @execute_node.register(ops.Array, tuple)
 def execute_array_column(op, cols, **kwargs):
     vals = [execute(arg, **kwargs) for arg in cols]
+
     length = next((len(v) for v in vals if isinstance(v, dd.Series)), None)
+    if length is None:
+        return vals
+
     n_partitions = next((v.npartitions for v in vals if isinstance(v, dd.Series)), None)
 
     def ensure_series(v):
@@ -47,8 +51,6 @@ def execute_array_column(op, cols, **kwargs):
         else:
             return dd.from_pandas(pd.Series([v] * length), npartitions=n_partitions)
 
-    if length is None:
-        return vals
     # dd.concat() can only handle array-likes.
     # If we're given a scalar, we need to broadcast it as a Series.
     df = dd.concat([ensure_series(v) for v in vals], axis=1)
--- a/ibis/backends/dask/execution/arrays.py
+++ b/ibis/backends/dask/execution/arrays.py
@@ -38,10 +38,8 @@ collect_list = dd.Aggregation(
 @execute_node.register(ops.Array, tuple)
 def execute_array_column(op, cols, **kwargs):
     vals = [execute(arg, **kwargs) for arg in cols]
-    # At least one of the values will be a Series.
-    # Otherwise op would be an ArrayScalar, not an ArrayColumn.
-    length = next(len(v) for v in vals if isinstance(v, dd.Series))
-    n_partitions = next(v.npartitions for v in vals if isinstance(v, dd.Series))
+    length = next((len(v) for v in vals if isinstance(v, dd.Series)), None)
+    n_partitions = next((v.npartitions for v in vals if isinstance(v, dd.Series)), None)
 
     def ensure_series(v):
         if isinstance(v, dd.Series):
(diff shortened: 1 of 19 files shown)


ibis/backends/pandas/execution/arrays.py
```diff
@@ -17,10 +17,22 @@
...
+    vals = [execute(arg, **kwargs) for arg in cols]
+    length = next((len(v) for v in vals if isinstance(v, pd.Series)), None)
+
+    def ensure_series(v):
+        if isinstance(v, pd.Series):
+            return v
+        else:
+            return pd.Series(v, index=range(length))
+
+    if length is None:
```
NickCrews, line 31: nit: move this to directly beneath where we set `length =`, so that `ensure_series()` is only defined if it is going to get used.

ibis/backends/dask/execution/arrays.py
```diff
@@ -34,10 +35,23 @@
...
+    length = next((len(v) for v in vals if isinstance(v, dd.Series)), None)
+    n_partitions = next((v.npartitions for v in vals if isinstance(v, dd.Series)), None)
+
+    def ensure_series(v):
+        if isinstance(v, dd.Series):
+            return v
+        else:
+            return dd.from_pandas(pd.Series([v] * length), npartitions=n_partitions)
+
+    if length is None:
```
NickCrews, line 50: nit: move this to directly beneath where we set `n_partitions =`, so that `ensure_series()` is only defined if it is going to get used.

ibis/expr/types/arrays.py
```diff
@@ -1017,82 +1017,50 @@ def __getitem__(self, index: int | ir.IntegerValue | slice) -> ir.Column:
...
-
-    ```python
-    values = [1, 2, 3]
-    ibis.literal(values)
-    ```
-
     Parameters
     ----------
     values
         An iterable of Ibis expressions or a list of Python literals
```
NickCrews, line 1026: nit: `An iterable of Ibis expressions or Python literals` (the python literals do not need to be in a list)

ibis/backends/tests/test_array.py
```diff
@@ -42,29 +41,21 @@
...
-    "duckdb": "DOUBLE[]",
-    "postgres": "numeric[]",
-    "flink": "ARRAY<DECIMAL(2, 1) NOT NULL> NOT NULL",
-}
-
-
-def test_array_scalar(con, backend):
+@pytest.mark.notimpl(["flink"], raises=com.OperationNotDefinedError)
+def test_array_scalar(con):
     expr = ibis.array([1.0, 2.0, 3.0])
```
NickCrews, line 59: make one of these values an ibis.literal()? Perhaps pass in a tuple or some other Iterable besides a list (this one not nearly as important).

ibis/backends/dask/execution/arrays.py
```diff
@@ -34,10 +35,23 @@
...
-    df = dd.concat(cols, axis=1)
+    vals = [execute(arg, **kwargs) for arg in cols]
+    length = next((len(v) for v in vals if isinstance(v, dd.Series)), None)
+    n_partitions = next((v.npartitions for v in vals if isinstance(v, dd.Series)), None)
+
+    def ensure_series(v):
+        if isinstance(v, dd.Series):
+            return v
+        else:
+            return dd.from_pandas(pd.Series([v] * length), npartitions=n_partitions)
```
NickCrews, line 48: also worth double checking me here, I'm not familiar with dask, so I was mostly just trying to make this not crash, but I'm not sure if setting n_partitions this way makes sense/is the most performant.

NickCrews (changes requested):

ibis/expr/types/arrays.py
cpcloud: This already existed so I won't change it here.

ibis/backends/tests/test_array.py
cpcloud: I'm already doing that in the other test cases I added.

kszucs (approved):
Nice!

PR #8057, opened 2024-01-22
fix(sql): don't generate table aliases for `ops.JoinLink`

kszucs:


cpcloud:
I can regenerate the impala snapshots.

cpcloud (approved):

PR #8077, opened 2024-01-23
test(markers): test the test markers

gforsyth:
I had a thought about a potential edge case with `None` being the only
listed "Exception" in a `raises` argument.  This fixes the edge case and
also adds some simple tests to make sure our custom markers correctly
handle when an imported backend exception is set to `None` due to a
missing import.


jcrist (approved):
Makes sense to me!

PR #8089, opened 2024-01-25
chore(deps): remove unused lonboard dependency

cpcloud:
Removes an unused dependency on `lonboard`


cpcloud:
Ah, I am mistaken, we are using this but as a transitive dependency of leafmap!

PR #8091, opened 2024-01-25
docs(blog): update geospatial - no need to_array()

ncclementi:
## Description of changes

At the moment of writing this blog, there was a need to use[to_array()](https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.to_array) to generate a scalar subquery from a table with a single column (whose shape is scalar). Otherwise it wouldn't work. For example: 

```python
sts_near_broad = streets.filter(_.geom.d_within(broad_station.select(_.geom).to_array(), 10))
```
However, a user posted on Zulip, that this wasn't needed. See [Zulip: Geospatial-not clear why `.to_array()` is required](https://ibis-project.zulipchat.com/#narrow/stream/405265-tech-support/topic/Geospatial.20-.20not.20clear.20why.20.2Eto_array.28.29.20is.20required/near/417987170)

After checking, the user is right, there is no need now for this. 

This PR, updates the queries accordingly

### DIFF
--- a/docs/posts/ibis-duckdb-geospatial/index.qmd
+++ b/docs/posts/ibis-duckdb-geospatial/index.qmd
@@ -117,7 +117,7 @@ boroughs
 ```
 
 ```{python}
-boroughs.filter(boroughs.geom.intersects(broad_station.select(broad_station.geom).to_array()))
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
:shipit:

PR #8095, opened 2024-01-25
ci(docker): yes -> "yes" to avoid yaml parsing weirdness

jcrist:
The `minio` container was failing to start in my local setup since `yes` was being parsed as `True` due to yaml's ~features~ weirdness. I'm assuming since this wasn't already failing in CI that some other version of docker-compose has patched around this yaml issue, but it's easy enough to fix it for everyone here.


cpcloud (approved):

cpcloud:
Yes yes!

PR #8100, opened 2024-01-25
fix(api): forbid using `asc`/`desc` in selections

### LINKED ISSUE #8080: `table.mutate(c = _.a.desc())` differs across implementations
### What happened?

```python
In [2]: import ibis
   ...: from ibis import _
   ...:
   ...: con = ibis.pandas.connect()
   ...: con.create_table("table", ibis.memtable({"a": [1, 2, 3], "b": [4, 5, 6]}))
   ...: table = con.table("table")
   ...:
   ...: table.mutate(c = _.a.desc()).execute()
Out[2]:
   a  b  c
0  1  4  1
1  2  5  2
2  3  6  3

In [3]: import ibis
   ...: from ibis import _
   ...:
   ...: con = ibis.polars.connect()
   ...: con.create_table("table", ibis.memtable({"a": [1, 2, 3], "b": [4, 5, 6]}))
   ...: table = con.table("table")
   ...:
   ...: table.mutate(c = _.a.desc()).execute()
Out[3]:
   a  b  c
0  1  4  3
1  2  5  2
2  3  6  1
```

Looks like the results differ? (if I could hazard a guess - is pandas' index autoaligning something under the hood?)

7.2.0

pandas / polars

### Relevant log output

_No response_

### Code of Conduct

jcrist:
Fixes #8080.

### DIFF
--- a/ibis/expr/operations/relations.py
+++ b/ibis/expr/operations/relations.py
@@ -17,13 +17,14 @@ from ibis.common.grounds import Concrete
 from ibis.common.patterns import Between, InstanceOf
 from ibis.common.typing import Coercible, VarTuple
 from ibis.expr.operations.core import Alias, Column, Node, Scalar, Value
-from ibis.expr.operations.sortkeys import SortKey  # noqa: TCH001
+from ibis.expr.operations.sortkeys import SortKey
 from ibis.expr.schema import Schema
 from ibis.formats import TableProxy  # noqa: TCH001
 
 T = TypeVar("T")
 
 Unaliased = Annotated[T, ~InstanceOf(Alias)]
+NonSortKey = Annotated[T, ~InstanceOf(SortKey)]
 
 
 @public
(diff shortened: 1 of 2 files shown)


ibis/expr/operations/relations.py
```diff
@@ -171,7 +172,7 @@ def _check_integrity(values, allowed_parents):
 @public
 class Project(Relation):
     parent: Relation
-    values: FrozenDict[str, Unaliased[Value]]
+    values: FrozenDict[str, NonSortKey[Unaliased[Value]]]
```
jcrist, line 175: There may be other places we want to add this restriction, but I didn't want to bother with them in this PR for now.

kszucs:
I have been thinking about introducing a `Marker` base class for wrapping value operations with additional metadata, like `alias` or `ascending`, then we could use an `UnmarkedValue` uniformly across the annotations. Going to think about it.

kszucs (approved):

jcrist:
Since aliases are valid in many places but `asc()`/`desc()` are really only valid in `order_by()` I wonder if `asc()`/`desc()` should return non `Value` operations? This would make it easier to forbid e.g. `t.x.asc() + 1`.

kszucs:
Yes, that should be the correct approach.

PR #8107, opened 2024-01-26
feat(mssql): add hashbytes and test for binary output hash fns

### LINKED ISSUE #8082: feat: add support for HashBytes operation on multiple backends

I would need to use md5 function with several backends that I use on different environments (DuckDB, Pyspark, MSSQL), with Pyspark being the most essential one.

Add support for operation `ibis.expr.operations.generic.HashBytes` on the following backends : DuckDB, Pyspark, MSSQL

 7.2.0

DuckDB, Pyspark, MSSQL

### Code of Conduct

gforsyth:
## Description of changes

This adds support for `ops.HashBytes` to `mssql` and also adds a test for that functionality so it's easier to port when we merge in the epic split branch.

I've also added a new op, `HashHexDigest` which returns the hexdigest of various cryptographic hashing functions since I imagine this is what many users are _probably_ after.  This newer op (and corresponding `hexdigest` method) can also support many more backends, as most of them default to returning the string hex digest and not the raw binary output.

I tried to be very accurate in the `notimpl` and `notyet` portions of both tests and I think I've done that.

For now, only exposing DuckDB, Pyspark, and MSSQL so we don't add a huge extra burden for the epic split but also address the user request in #8082 

And I guess now we can commence debate over the method name?  :racehorse: 

## Issues closed

Resolves #8082

### DIFF
--- a/ibis/expr/operations/generic.py
+++ b/ibis/expr/operations/generic.py
@@ -273,6 +273,20 @@ class HashBytes(Value):
     shape = rlz.shape_like("arg")
 
 
+@public
+class HexDigest(Value):
+    arg: Value[dt.String | dt.Binary]
+    how: LiteralType[
+        "md5",
+        "sha1",
+        "sha256",
+        "sha512",
+    ]
+
+    dtype = dt.str
+    shape = rlz.shape_like("arg")
+
+
 # TODO(kszucs): we should merge the case operations by making the
 # cases, results and default optional arguments like they are in
 # api.py
--- a/ibis/expr/types/strings.py
+++ b/ibis/expr/types/strings.py
@@ -465,6 +465,40 @@ class StringValue(Value):
         """
         return ops.HashBytes(self, how).to_expr()
 
+    def hexdigest(
+        self,
+        how: Literal["md5", "sha1", "sha256", "sha512"] = "sha256",
+    ) -> ir.StringValue:
+        """Return the hash digest of the input as a hex encoded string.
+
+        Parameters
+        ----------
+        how
+            Hash algorithm to use
+
+        Returns
+        -------
+        StringValue
+            Hexadecimal representation of the hash as a string
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.memtable({"species": ["Adelie", "Chinstrap", "Gentoo"]})
+        >>> t.species.hexdigest()
+        ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
+        ┃ HexDigest(species)                                           ┃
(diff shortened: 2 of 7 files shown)


ibis/expr/types/strings.py
```diff
@@ -465,6 +465,24 @@ def hashbytes(
         """
         return ops.HashBytes(self, how).to_expr()
 
+    def hexdigest(
+        self,
+        how: Literal["md5", "sha1", "sha256", "sha512"] = "sha256",
+    ) -> ir.StringValue:
+        """Compute the binary hash value of the input.
```
jcrist, line 472: Maybe clearer, idk

```suggestion
        """Return the hash digest of the input as a hex encoded string.
```

ibis/expr/types/strings.py
```diff
@@ -465,6 +465,24 @@ def hashbytes(
...
+
+        Parameters
+        ----------
+        how
+            Hash algorithm to use
+
+        Returns
+        -------
+        StringValue
+            Hexadecimal representation of hash as a string
```
jcrist, line 482:
```suggestion
            Hexadecimal representation of the hash as a string
```

jcrist: Could also use an example in the docstring

jcrist (commented):
Overall LGTM. Just a few docstring requests.

> And I guess now we can commence debate over the method name? 🐎

I like the name `hexdigest` for the new operation, seems clear and matches the equivalent python stdlib name.

cpcloud:
Should we discuss deprecating whichever other method overlaps with the new `hexdigest`? I can't remember whether it's hash or hashbytes or something else.

ibis/expr/operations/generic.py
```diff
@@ -273,6 +273,20 @@ class HashBytes(Value):
     shape = rlz.shape_like("arg")
 
 
+@public
+class HashHexDigest(Value):
+    arg: Value[dt.String | dt.Binary]
```
cpcloud, line 278: Any chance we can call it `HexDigest`?

gforsyth: It is done.

gforsyth:
> Should we discuss deprecating whichever other method overlaps with the new hexdigest? I can't remember whether it's hash or hashbytes or something else.

So it would be hashbytes, although currently there is a difference in that hashbytes returns the binary hash and there are at least a few backends that support both of those options.

ibis/expr/operations/generic.py
kszucs: Can we reuse the `how` annotation for both `HashBytes` and `HashDigest`?

gforsyth: Hey @kszucs -- I would like to do that eventually but not here.  There's a fair bit of inconsistency in the output type of various hashing functions between the backends and I'd like to address that in `the-epic-split` instead of here.

kszucs: Sure, it is not that important.

cpcloud (approved):

PR #8118, opened 2024-01-27
build(docker): clean up minio setup

cpcloud:
Cleans up the minio container setup in preparation for the risingwave backend.


ibis/backends/trino/tests/conftest.py
```diff
@@ -37,6 +37,7 @@ class TestConf(ServiceBackendTest):
     # for numeric and decimal
 
     service_name = "minio"
+    data_volume = "/bitnami/minio/data"
```
cpcloud, line 40: This already exists by the time the container is running.

PR #8119, opened 2024-01-27
test(api): add test for alias generation

### LINKED ISSUE #7350: bug: `CompileError: Multiple, unrelated CTEs found`
### What happened?

EDIT: see below I found a much simpler repro

Ok, sorry I'm not sure what the exact cause of this is. Thanks for the help!

Basically, I
1. load some data
2. perform some cleaning/transformations to get `featured` table
3. self join `featured` on condition A to get `ja`.
4. self join `featured` on condition B to get `jb`.
5. ibis.union(ja, jb)

If I do NOT .cache `featured` before the joins, then I get the error during the union. If I DO .cache `featured` before the joins, then there is no error. Regardless of whether I .cache(), I can always .execute() `ja` and `jb` before the union and it does not error.

Exact repro steps :
1. run `git clone https://github.com/nickcrews/mismo.git` and CD into that repo
2. `git checkout 3105eaf7ef84ad5b629d7ce1177a05dc83469b68`
3. Make a venv and install frozen deps. Install PDM if you don't have it. Then run `pdm install`.
4. open [`/docs/examples/patent_deduplication.ipynb`](https://github.com/NickCrews/mismo/blob/3105eaf7ef84ad5b629d7ce1177a05dc83469b68/docs/examples/patent_deduplication.ipynb)
5. Run until you reach the cell containing `rules.block(featured, featured)`
6. If you comment out the .cache() call, you get the error. Try again with the .cache() and no error.

Let me know how else I can be helpful with debugging this.

7.0.0

duckdb

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Adds a test for #7350.

Closes #7350.

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -1219,7 +1219,7 @@ def execute_view(op, *, ctx: pl.SQLContext, **kw):
 
 @translate.register(ops.SelfReference)
 def execute_self_reference(op, **kw):
-    return translate(op.table, **kw)
+    return translate(op.parent, **kw)
 
 
 @translate.register(ops.JoinTable)
--- a/ibis/backends/tests/test_sql.py
+++ b/ibis/backends/tests/test_sql.py
@@ -195,3 +195,14 @@ def test_union_aliasing(backend_name, snapshot):
     result = top_ten.union(bottom_ten)
 
     snapshot.assert_match(str(ibis.to_sql(result, dialect=backend_name)), "out.sql")
+
+
+def test_union_generates_predictable_aliases(con):
+    t = ibis.memtable(
+        data=[{"island": "Torgerson", "body_mass_g": 3750, "sex": "male"}]
+    )
+    sub1 = t.inner_join(t.view(), "island").mutate(island_right=lambda t: t.island)
+    sub2 = t.inner_join(t.view(), "sex").mutate(sex_right=lambda t: t.sex)
+    expr = ibis.union(sub1, sub2)
+    df = con.execute(expr)
+    assert len(df) == 2
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -937,9 +937,6 @@ def test_agg_memory_table(con, monkeypatch):
     assert result == 3
 
 
-@pytest.mark.broken(
-    ["polars"], reason="join column renaming is currently incorrect on polars"
-)
 def test_self_join_memory_table(backend, con, monkeypatch):
     monkeypatch.setattr(ibis.options, "default_backend", con)
 


cpcloud:
Lemme see if I can make this an execution test to avoid adding 1500 new LoC.

ibis/backends/tests/test_sql.py
```diff
@@ -195,3 +195,14 @@ def test_union_aliasing(backend_name, snapshot):
     result = top_ten.union(bottom_ten)
 
     snapshot.assert_match(str(ibis.to_sql(result, dialect=backend_name)), "out.sql")
+
+
+def test_union_generates_predictable_aliases(con):
+    t = ibis.memtable(
+        data=[{"island": "Torgerson", "body_mass_g": 3750, "sex": "male"}]
+    )
+    sub1 = t.inner_join(t.view(), "island").mutate(island_right=lambda t: t.island)
```
kszucs, line 204: Actually you don't need to use `.view()` here anymore because of `dereference_sides()`

ibis/backends/polars/compiler.py
```diff
@@ -1219,7 +1219,7 @@ def execute_view(op, *, ctx: pl.SQLContext, **kw):
 
 @translate.register(ops.SelfReference)
 def execute_self_reference(op, **kw):
-    return translate(op.table, **kw)
+    return translate(op.parent, **kw)
```
kszucs, line 1222: Nice catch!

ibis/backends/tests/test_sql.py
```diff
@@ -195,3 +195,14 @@ def test_union_aliasing(backend_name, snapshot):
     result = top_ten.union(bottom_ten)
 
     snapshot.assert_match(str(ibis.to_sql(result, dialect=backend_name)), "out.sql")
+
+
+def test_union_generates_predictable_aliases(con):
```
kszucs, line 200: So this case was failing to run on `main`, right?

ibis/backends/tests/test_sql.py
cpcloud: Right, but that's the test case, so I'm going to leave it for now.

ibis/backends/tests/test_sql.py
cpcloud: Yep:

```
In [3]: import ibis
   ...:
   ...: from ibis import _
   ...:
   ...: t = ibis.examples.penguins.fetch()
   ...: # No error if you remove the .head() call
   ...: t = t.head(100)
   ...:
   ...: # The rename is simply needed so the schemas are the same so the union works
   ...: sub1 = t.inner_join(t.view(), "island").mutate(island_right=_.island)
   ...: sub2 = t.inner_join(t.view(), "sex").mutate(sex_right=_.sex)
   ...: u = ibis.union(sub1, sub2)
   ...: u.execute()
...
CompileError: Multiple, unrelated CTEs found with the same name: 't5'
```

kszucs (approved):
Thanks!

Issue #8058, opened 2024-01-22
bug: compilation failure in table expression mixing `ibis.row_number()`, `.isin()` and `.sample()`


ogrisel:
### What happened?

The following snippet that uses `table.sample` raises a low level SQLAlchemy exception.

```python
import ibis

t = ibis.memtable({"a": range(30)})
t = t.mutate(_id=ibis.row_number())
test = t.sample(fraction=0.25, seed=0)
t = t.mutate(is_test=t._id.isin(test._id))
train = t.filter(~t.is_test)
ibis.to_sql(train)
```

raises:

<details>

```python-traceback
Traceback (most recent call last):
  Cell In[2], line 9
    ibis.show_sql(train)
  File ~/miniforge3/envs/dev/lib/python3.11/site-packages/ibis/expr/sql.py:336 in show_sql
    print(to_sql(expr, dialect=dialect), file=file)
  File ~/miniforge3/envs/dev/lib/python3.11/site-packages/ibis/expr/sql.py:410 in to_sql
    sql = backend._to_sql(expr, **kwargs)
  File ~/miniforge3/envs/dev/lib/python3.11/site-packages/ibis/backends/base/sql/alchemy/__init__.py:187 in _to_sql
    sql = self.compile(expr, **kwargs).compile(
  File ~/miniforge3/envs/dev/lib/python3.11/site-packages/ibis/backends/base/sql/__init__.py:395 in compile
    return self.compiler.to_ast_ensure_limit(expr, limit, params=params).compile()
  File ~/miniforge3/envs/dev/lib/python3.11/site-packages/ibis/backends/base/sql/compiler/base.py:37 in compile
    compiled_queries = [q.compile() for q in self.queries]
  File ~/miniforge3/envs/dev/lib/python3.11/site-packages/ibis/backends/base/sql/compiler/base.py:37 in <listcomp>
    compiled_queries = [q.compile() for q in self.queries]
  File ~/miniforge3/envs/dev/lib/python3.11/site-packages/ibis/backends/base/sql/alchemy/query_builder.py:209 in compile
    frag = self._compile_table_set()
  File ~/miniforge3/envs/dev/lib/python3.11/site-packages/ibis/backends/base/sql/alchemy/query_builder.py:237 in _compile_table_set
    return self.table_set_formatter_class(self, self.table_set).get_result()
[...]

PR #8124, opened 2024-01-28
fix(common): don't match an `Object` pattern with more positional arguments defined than `__match_args__` has

kszucs:
Turns out one of the patterns were referencing non-existing fields and the `Object` pattern was still matching on it.

Fixes https://github.com/ibis-project/ibis/issues/8058.

### DIFF
--- a/ibis/common/patterns.py
+++ b/ibis/common/patterns.py
@@ -1264,17 +1264,26 @@ class Object(Slotted, Pattern):
             return InstanceOf(type)
         return super().__create__(type, *args, **kwargs)
 
-    def __init__(self, type, *args, **kwargs):
-        type = pattern(type)
+    def __init__(self, typ, *args, **kwargs):
+        if isinstance(typ, type) and len(typ.__match_args__) < len(args):
+            raise ValueError(
+                "The type to match has fewer `__match_args__` than the number "
+                "of positional arguments in the pattern"
+            )
+        typ = pattern(typ)
         args = tuple(map(pattern, args))
         kwargs = frozendict(toolz.valmap(pattern, kwargs))
-        super().__init__(type=type, args=args, kwargs=kwargs)
+        super().__init__(type=typ, args=args, kwargs=kwargs)
 
     def match(self, value, context):
         if self.type.match(value, context) is NoMatch:
             return NoMatch
 
-        patterns = {**dict(zip(value.__match_args__, self.args)), **self.kwargs}
+        # the pattern requirest more positional arguments than the object has
+        if len(value.__match_args__) < len(self.args):
+            return NoMatch
+        patterns = dict(zip(value.__match_args__, self.args))
+        patterns.update(self.kwargs)
 
         fields = {}
         changed = False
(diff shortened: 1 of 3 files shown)


ibis/expr/rewrites.py
```diff
@@ -211,7 +211,7 @@ def empty_in_values_right_side(_):
 
 @replace(
     p.WindowFunction(
-        p.PercentRank(y) | p.RankBase(y) | p.CumeDist(y) | p.NTile(y),
```
kszucs, line 214: None of `PercentRank`, `RankBase` or `CumeDist` has an argument.

ibis/backends/tests/sql/test_compiler.py
```diff
@@ -230,3 +231,15 @@ def test_union_order_by(snapshot):
     expr = t.order_by("b").union(t.order_by("b"))
     snapshot.assert_match(to_sql(expr), "out.sql")
     assert_decompile_roundtrip(expr, snapshot, eq=schemas_eq)
+
+
+def test_window_with_row_number_compiles(snapshot):
```
cpcloud, line 236: A couple things:

1. This doesn't seem like the correct place for this test since it's not actually varying the test over the backends.
2. `snapshot` isn't being used, did you mean to use it?

ibis/common/patterns.py
```diff
@@ -1264,17 +1264,26 @@ def __create__(cls, type, *args, **kwargs):
             return InstanceOf(type)
         return super().__create__(type, *args, **kwargs)
 
-    def __init__(self, type, *args, **kwargs):
-        type = pattern(type)
+    def __init__(self, typ, *args, **kwargs):
+        if isinstance(typ, type) and len(typ.__match_args__) < len(args):
+            raise ValueError(
+                "The type to match has less `__match_args__` than the number "
```
cpcloud, line 1270:
```suggestion
                "The type to match has fewer `__match_args__` than the number "
```

ibis/expr/rewrites.py
cpcloud: Good catch!

cpcloud (changes requested):
Just a couple small issues with the test case.

ibis/backends/tests/sql/test_compiler.py
kszucs: 1. Agree, though where should I test it then? Under `ibis/backends/base/sqlglot/tests`?
2. Yes, I intended to use the snapshot fixture, but forgot it eventually. W

ibis/common/patterns.py
kszucs: Heh :)

ibis/backends/tests/sql/test_compiler.py
cpcloud: `ibis/backends/base/sqlglot/tests` seems fine

kszucs: Moved.

ibis/common/patterns.py
kszucs: Updated.

kszucs:
@cpcloud addressed your comments

cpcloud:
I'm not sure why that job is failing, I thought we moved away from those caching actions.

cpcloud:
Oh, that's not yet in `the-epic-split` :(

cpcloud (approved):

PR #8127, opened 2024-01-28
feat(api): support the inner join convenience to not repeat fields known to be equal

kszucs:
- feat(api): support the inner join convenience to not repeat fields known to be equal
- test(sql): regenerate SQL snapshot for the inner join convenience feature


ibis/backends/tests/snapshots/test_generic/test_many_subqueries/bigquery/out.sql
```diff
@@ -1,36 +1,42 @@
...
   SELECT
-    t5.*,
-    (
-      row_number() OVER (ORDER BY t5.`street` ASC) - 1
-    ) AS `key`
-  FROM data AS t5
+    t5.street,
+    ROW_NUMBER() OVER (ORDER BY t5.street ASC) - 1 AS key
+  FROM (
+    SELECT
```
kszucs, line 6: I'm going to remove these redundant selects soon-ish.

kszucs:
@cpcloud could you please update the snapshots for the failing backends?

ibis/expr/types/joins.py
```diff
@@ -154,7 +249,15 @@ def __init__(self, arg, collisions=None):
...
         super().__init__(arg)
+        # the collisions and equalities are used to track the name collisions
+        # and the equality groups join fields based on equality predicates;
+        # these must be tracked in the join expression because the join chain
+        # operation doesn't hold any information about `lname` and `rname`
+        # parameters passed to the join methods and used to disambiguate field
+        # names; the collisions are used to raise an error if there are any
+        # name collisions after the join chain is finished
         object.__setattr__(self, "_collisions", collisions or set())
+        object.__setattr__(self, "_equalities", equalities or DisjointSet())
```
cpcloud, line 260: Just to confirm: there's an instance of `DisjoinSet` per instance of `Join`, and each instance of `Join` corresponds to a chain of joins. Is that correct?

cpcloud (approved):
LGTM, excellent work and novel use of a disjoint set. Thanks for doing this.

🚢 it

ibis/expr/types/joins.py
kszucs: Correct. The disjoint set gets copied as new links aded to the join chain.

PR #8128, opened 2024-01-29
docs(geospatial): add examples for duckdb supported  methods

### LINKED ISSUE #7959: docs: add example docstrings to geospatial methods
Now that the duckdb backend supports geospatial operations, we should be able to add example docstrings on any geo method that the duckdb backend supports.

Using the dataset (or datasets) added in #7958 may be useful here.

ncclementi:
## Description of changes
- Add docstring examples for duckdb geospatial functions

## Issues closed

Closes #7959

TODO/Problems/need help

- [x] Understand why CI is broken and fix it.
- [x] Make sure examples render properly in CI, I can't get the docs to build locally because of seg fault (xref https://github.com/duckdb/duckdb_spatial/issues/236)
- [x] Need to figure out quarto syntax to include in the geospatial expressions page `point` (which is in numeric types)
- [x] Decided what to do about  a few methdos that are specific to `LineStrings` and the zones example doesn't contain such type of geometry.

### DIFF
--- a/ibis/expr/types/geospatial.py
+++ b/ibis/expr/types/geospatial.py
@@ -20,6 +20,30 @@ class GeoSpatialValue(NumericValue):
         -------
         FloatingValue
             The area of `self`
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.examples.zones.fetch()
+        >>> t.geom.area()
+        ┏━━━━━━━━━━━━━━━┓
+        ┃ GeoArea(geom) ┃
+        ┡━━━━━━━━━━━━━━━┩
+        │ float64       │
+        ├───────────────┤
+        │  7.903953e+07 │
+        │  1.439095e+08 │
+        │  3.168508e+07 │
+        │  8.023733e+06 │
+        │  5.041488e+07 │
+        │  4.093479e+07 │
+        │  3.934104e+07 │
+        │  2.682802e+06 │
+        │  3.416422e+07 │
+        │  4.404143e+07 │
+        │             … │
+        └───────────────┘
         """
         return ops.GeoArea(self).to_expr()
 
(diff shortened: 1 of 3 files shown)


ibis/expr/types/geospatial.py
```diff
@@ -455,6 +973,10 @@ def start_point(self) -> PointValue:
         -------
         PointValue
             Start point
+
+        Example
+        -------
+        ?? this only applies to lines and no lines in t.geom
```
ncclementi, line 979: This method is specific too `LineStrings`, and we don't have a way to create one from the table to get an example

ibis/expr/types/geospatial.py
```diff
@@ -467,6 +989,10 @@ def end_point(self) -> PointValue:
         -------
         PointValue
             End point
+
+        Example
+        -------
+        ?? this only applies to lines and no lines in t.geom
```
ncclementi, line 995: This method is specific too `LineStrings`, and we don't have a way to create one from the table to get an example

ibis/expr/types/geospatial.py
```diff
@@ -368,7 +814,7 @@ def union(self, right: GeoSpatialValue) -> GeoSpatialValue:
         """Merge two geometries into a union geometry.
 
         Returns the pointwise union of the two geometries.
-        This corresponds to the non-aggregate version the PostGIS ST_Union.
+        This corresponds to the non-aggregate version.
```
cpcloud, line 817:
```suggestion
```

I think this might be confusing, it assumes knowledge of union aggregates.

ibis/expr/types/geospatial.py
cpcloud: A literal seems fine here, would that work?

ibis/expr/types/geospatial.py
cpcloud: Feel free to use as literal for anything that's specific to a certain type of geometry.

cpcloud (changes requested):
Thanks for churning through this!

Can you also apply `ruff format .`? The version of `ruff` listed in `pyproject.toml` should automatically format the code in docstrings.

ibis/expr/types/geospatial.py
```diff
@@ -75,6 +121,31 @@ def contains(self, right: GeoSpatialValue) -> ir.BooleanValue:
         -------
         BooleanValue
             Whether `self` contains `right`
+
+        Examples
+        --------
+        >>> from ibis.interactive import *
+        >>> import shapely
+        >>> t = ibis.examples.zones.fetch()
+        >>> p = shapely.Point(935996.821,191376.75) # centroid for zone 1
```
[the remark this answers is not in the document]
ncclementi: Checked, we don't seem to have that API.  But it looks like we can do:

```python
point = ibis.literal((1, 0), type="point").name("p")
point_geom = ibis.literal((1, 0), type="point:geometry").name("p")
```

That being said I'd rather leave the shapely api here, because in `the-epic-split` we can pass that directly.

ibis/expr/types/geospatial.py
```diff
@@ -379,6 +854,34 @@ def union(self, right: GeoSpatialValue) -> GeoSpatialValue:
...
         GeoSpatialValue
             Union of geometries
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> import shapely
+        >>> t = ibis.examples.zones.fetch()
+        Penn station zone centroid
```
cpcloud, line 864: You have to put a newline on either side of the prose, that's why my suggestion had them :)

Otherwise the lines are interpreted as output from the previous repl line.

ncclementi:
The failures I'm seeing are weird, because the expected and the result match. This is only happening for the operations that return geometries. The only difference I see is that the result seems to be truncated shorter than the expected value. But I'm not sure if that can be the problem.

ncclementi:
> It's a character-by-character check, so having the results displayed in fewer columns will break the doctests

@gforsyth ok, this makes sense, but I'm using whatever is the default but what gets produced on my end but that doesn't seem to match whatever CI is producing.

gforsyth:
> @gforsyth ok, this makes sense, but I'm using whatever is the default but what gets produced on my end but that doesn't seem to match whatever CI is producing.

yeah, it's frustrating.  The output gets resized depending on terminal width, so I would recommend narrowing your terminal (and then making note of how many columns you want the output to take up)

gforsyth:
> yeah, it's frustrating. The output gets resized depending on terminal width, so I would recommend narrowing your terminal (and then making note of how many columns you want the output to take up)

The github actions terminal is set to an 80 character width.  Might be possible to change that, but for now, I would just set your terminal to that for generating the doctest outputs

ncclementi:
Ok getting closer. 
The ones failing, I'm expecting them to fail (kind off) because of https://github.com/ibis-project/ibis/issues/8143 but, somehow that is working on CI, though the output is different because I was not able to create the expected output locally.

ncclementi:
I think this is ready for a final look, got to render the `NumericValue.point` in the geospatial expressions page too. I think it looks good.

cpcloud (approved):
SWEET 🚢 it!

ncclementi:
This is ready for final review. 

We'll solve/discuss the literals issue #8143 in a separate PR.

Issue #7345, opened 2023-10-12
bug(polars): joins requiring column renaming are currently broken


jcrist:
### What happened?

The polars backend currently gives an incorrect result for any joins where the left and right tables have conflicting column names.

```python
In [1]: import ibis

In [2]: ibis.options.interactive = True

In [3]: duckdb = ibis.duckdb.connect()

In [4]: polars = ibis.polars.connect()

In [5]: t1 = ibis.memtable({"x": [1, 2, 3], "y": [4, 5, 6], "z": ["a", "b", "c"]})

In [6]: t2 = ibis.memtable({"x": [3, 2, 1], "y": [7, 8, 9], "z": ["d", "e", "f"]})

In [7]: duckdb.execute(t1.join(t2, "x"))  # expected
Out[7]: 
   x  y  z  y_right z_right
0  1  4  a        9       f
1  2  5  b        8       e
2  3  6  c        7       d

In [8]: polars.execute(t1.join(t2, "x"))  # wrong result
Out[8]: 
   x  y  z  y_right z_right
0  3  6  c        6       c
1  2  5  b        5       b
2  1  4  a        4       a
```

Instead of including data from the right conflicting columns, the data is duplicated from the left conflicting columns. After inspecting a bit, the problem lies not with the join but with the follow-up selection.

The selections handled [here](https://github.com/ibis-project/ibis/blob/f2ae7cc02f84420a5b618180caf1e77267a0745e/ibis/backends/polars/compiler.py#L199) and [here](https://github.com/ibis-project/ibis/blob/f2ae7cc02f84420a5b618180caf1e77267a0745e/ibis/backends/polars/compiler.py#L174) are translating

```python
ops.Alias(ops.TableColumn(t2, "y"), "y_right")
```

as

```python
pl.col("y").alias("y_right")
```

since the latter lacks a table reference, the left table is used by polars to resolve `y` instead of the right table.

### What version of ibis are you using?

dev

### What backend(s) are you using, if any?

polars

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
@ritchie46 Is it possible to select columns from the right table in a join?

cpcloud:
To be totally clear, here's an example:

```
In [8]: left = pl.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6], "z": ["a", "b", "c"]}).lazy()

[...]

PR #8134, opened 2024-01-29
fix(polars): columns are picked from the correct side in case of conflicting names

kszucs:
It has been already fixed by [the rewrite converting the join chain to pairwise join operations](https://github.com/ibis-project/ibis/issues/7345
), so this only adds a test case.

Fixes https://github.com/ibis-project/ibis/issues/7345
Depends on https://github.com/ibis-project/ibis/pull/8127

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -988,9 +988,7 @@ def test_memtable_column_naming_mismatch(backend, con, monkeypatch, df, columns)
 
 
 @pytest.mark.notimpl(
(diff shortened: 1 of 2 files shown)


gforsyth:
If I remember correctly this also required upstream changes in polars -- do we want to bump the lower bound of supported polars versions to pick up the required change?  I'll track down that issue...

gforsyth:
oh, it's linked in #7345 -- but also, it doesn't look like it has been fixed upstream yet (although they plan to).

cpcloud:
I am rebasing this now.

PR #8149, opened 2024-01-30
refactor(sql): remove temporary table creation when using inline sql

cpcloud:
This PR fixes a long-standing annoyance with our `.sql` methods. Previously we pooped a bunch of temporary tables or views (depending on the backend), but after this PR the various `.sql` methods replace this hack with much more vanilla CTEs.

### DIFF
--- a/ibis/backends/base/sqlglot/__init__.py
+++ b/ibis/backends/base/sqlglot/__init__.py
@@ -151,6 +151,17 @@ class SQLGlotBackend(BaseBackend):
         """Return an ibis Schema from a backend-specific SQL string."""
         return sch.Schema.from_tuples(self._metadata(query))
 
+    def _get_sql_string_view_schema(self, child: ir.Table, query: str) -> sch.Schema:
+        # abort if somehow this was called with a non-view
+        assert isinstance(child.op(), ops.View), type(child.op())
+
+        obj = self._to_sqlglot(child)
+        dialect = self.compiler.dialect
+        parsed = sg.parse_one(query, read=dialect)
+        parsed.args["with"] = obj.args["with"]
+        sql = parsed.sql(dialect)
+        return self._get_schema_using_query(sql)
+
     def create_view(
         self,
         name: str,
@@ -194,27 +205,6 @@ class SQLGlotBackend(BaseBackend):
         )
         with self._safe_raw_sql(src):
             pass
-
-    def _get_temp_view_definition(self, name: str, definition: str) -> str:
-        return sge.Create(
-            this=sg.to_identifier(name, quoted=self.compiler.quoted),
-            kind="VIEW",
-            expression=definition,
-            replace=True,
-            properties=sge.Properties(expressions=[sge.TemporaryProperty()]),
-        )
-
-    def _create_temp_view(self, table_name, source):
-        if table_name not in self._temp_views and table_name in self.list_tables():
-            raise ValueError(
-                f"{table_name} already exists as a non-temporary table or view"
-            )
-
-        with self._safe_raw_sql(self._get_temp_view_definition(table_name, source)):
-            pass
-
-        self._temp_views.add(table_name)
-        self._register_temp_view_cleanup(table_name)
 
     def _register_temp_view_cleanup(self, name: str) -> None:
         """Register a clean up function for a temporary view.
--- a/ibis/backends/base/sqlglot/__init__.py
+++ b/ibis/backends/base/sqlglot/__init__.py
@@ -152,12 +152,15 @@ class SQLGlotBackend(BaseBackend):
         return sch.Schema.from_tuples(self._metadata(query))
 
     def _get_sql_string_view_schema(self, name, table, query) -> sch.Schema:
-        dialect = self.compiler.dialect
+        compiler = self.compiler
+        dialect = compiler.dialect
 
         cte = self._to_sqlglot(table)
         parsed = sg.parse_one(query, read=dialect)
         parsed.args["with"] = cte.args.pop("with", [])
-        parsed = parsed.with_(name, as_=cte, dialect=dialect)
+        parsed = parsed.with_(
+            sg.to_identifier(name, quoted=compiler.quoted), as_=cte, dialect=dialect
+        )
 
         sql = parsed.sql(dialect)
         return self._get_schema_using_query(sql)
--- a/ibis/backends/base/sqlglot/__init__.py
+++ b/ibis/backends/base/sqlglot/__init__.py
@@ -151,14 +151,14 @@ class SQLGlotBackend(BaseBackend):
         """Return an ibis Schema from a backend-specific SQL string."""
         return sch.Schema.from_tuples(self._metadata(query))
 
-    def _get_sql_string_view_schema(self, child: ir.Table, query: str) -> sch.Schema:
-        # abort if somehow this was called with a non-view
-        assert isinstance(child.op(), ops.View), type(child.op())
-
-        obj = self._to_sqlglot(child)
+    def _get_sql_string_view_schema(self, name, table, query) -> sch.Schema:
         dialect = self.compiler.dialect
+
+        cte = self._to_sqlglot(table)
         parsed = sg.parse_one(query, read=dialect)
-        parsed.args["with"] = obj.args["with"]
+        parsed.args["with"] = cte.args.pop("with", [])
+        parsed = parsed.with_(name, as_=cte, dialect=dialect)
+
         sql = parsed.sql(dialect)
         return self._get_schema_using_query(sql)
 
--- a/ibis/backends/base/sqlglot/compiler.py
+++ b/ibis/backends/base/sqlglot/compiler.py
@@ -261,31 +261,20 @@ class SQLGlotCompiler(abc.ABC):
         op, ctes = sqlize(op)
 
         aliases = {}
-        alias_counter = itertools.count()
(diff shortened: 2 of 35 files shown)


cpcloud:
@ssanderson I was finally able to clean this up ... years later 😂

ibis/expr/operations/relations.py
```diff
@@ -408,29 +408,40 @@ class SQLQueryResult(Relation):
...
     def schema(self):
-        return self.child.schema
+        import sqlglot as sg
+
+        # TODO(kszucs): avoid converting to expression
+        op = self.child
+        child = op.to_expr()
+        backend = child._find_backend()
+
+        obj = backend._to_sqlglot(child)
```
kszucs, line 439: Could you please move the sqlglot related code to `_get_schema_using_query()`? Either passing child as an optional argument to it, or if `_get_schema_using_query` is only used for this purpose then it could be even renamed to reflect this is for querying the sqlstringview schema.

ibis/expr/operations/relations.py
```diff
@@ -408,29 +408,40 @@ class SQLQueryResult(Relation):
 
 
 @public
-class SQLStringView(PhysicalTable):
-    """A view created from a SQL string."""
+class View(PhysicalTable):
```
kszucs, line 411: Could we rename this to something like `NamedRelation`? Possibly that should be a base class for `PhysicalTable(NamedRelation)`.

ibis/expr/operations/relations.py
```diff
@@ -408,29 +408,40 @@ class SQLQueryResult(Relation):
...
 
 @public
-class View(PhysicalTable):
-    """A view created from an expression."""
+class SQLStringView(Relation):
+    """A view created from a SQL string."""
 
-    child: Relation
-    name: str
+    child: View
```
kszucs, line 425: Then this could be `NamedRelation`

ibis/backends/base/sqlglot/rewrites.py
```diff
@@ -157,7 +157,9 @@ def extract_ctes(node):
 
     g = Graph.from_bfs(node, filter=(ops.Relation, ops.Subquery, ops.JoinLink))
     for node, dependents in g.invert().items():
-        if len(dependents) > 1 and isinstance(node, cte_types):
+        if isinstance(node, ops.View) or (
```
kszucs, line 160: Could we avoid having this special case here? IIC it is only to force turning a `View` into a CTE.

Without this the generated SQL would be a subselect, right?

ibis/backends/duckdb/__init__.py
```diff
@@ -1534,3 +1534,16 @@ def insert(
                 table_name,
                 obj if isinstance(obj, pd.DataFrame) else pd.DataFrame(obj),
             )
+
+    def _get_temp_view_definition(self, name: str, definition: str) -> str:
```
kszucs, line 1538: Does duckdb use these for DDL? At least you removed these methods for the other backends.

ibis/backends/base/sqlglot/rewrites.py
cpcloud: Without this change a cute won't be created, and View _must_ compile to a cte so it's visible by any subsequent query.

ibis/expr/operations/relations.py
cpcloud: I'd rather not do that in this PR, it's not directly related to the problem.

ibis/expr/operations/relations.py
cpcloud: Can be done in a follow up if someone is motivated to do so.

ibis/expr/operations/relations.py
```diff
@@ -408,29 +408,40 @@ class SQLQueryResult(Relation):
...
-        # TODO(kszucs): avoid converting to expression
-        backend = self.child.to_expr()._find_backend()
-        return backend._get_schema_using_query(self.query)
+        return self.child.schema
 
 
 @public
-class View(PhysicalTable):
-    """A view created from an expression."""
+class SQLStringView(Relation):
```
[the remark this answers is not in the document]
cpcloud: I don't want to rename anything in this PR

ibis/backends/duckdb/__init__.py
cpcloud: Yes DuckDB uses these for all the read_foo methods.

ibis/expr/operations/relations.py
cpcloud: Yep, I'll look into it.

ibis/backends/base/sqlglot/rewrites.py
cpcloud: Consider the case of joining two aliased ibis expressions. The aliases need to be visible at the top level for a subsequent join that uses them. An aliased subselect would hide all aliases except the immediate parent. AFAIK only CTEs allow defining multiple named relations with the same scope.

ibis/expr/types/relations.py
```diff
@@ -3263,7 +3255,13 @@ def sql(self, query: str, dialect: str | None = None) -> ir.Table:
         if dialect is not None:
             backend = self._find_backend()
             query = backend._transpile_sql(query, dialect=dialect)
-        op = ops.SQLStringView(child=self, name=next(_ALIASES), query=query)
+
+        child = (
+            self
+            if isinstance(self.op(), ops.View)
+            else self.alias(util.gen_name("sql_string_view"))
```
kszucs, line 3262: What happens when we generate an alias? How is query going to use that name?

ibis/backends/base/sqlglot/rewrites.py
kszucs: This code still confuses me, you add `View` here to the list of the relations which should be wrapped with a CTE, but you exclude the same type from the dict comprehension below.

cpcloud: I'm adding it to the list to be able to generate CTEs later, the ibis `CTE` object is only used in a call to `replace`, the dict of which filters out the `View` objects. There's are a few separate components of that process, that are _independent of this PR_:

1. The extraction of things that should become CTEs
2. The replacement of existing nodes with those CTEs
3. The construction of sqlglot CTEs based on step 1

I am reusing steps 1 and 3 for this PR, but operating on `View`s.

ibis/expr/types/relations.py
cpcloud: There's no requirement that a subsequent query use this name, it's merely to satisfy the constraint I introduced that `SQLStringView`'s `child` member type is `View` and not an arbitrary relation.

cpcloud: Maybe there's a better way to do that that avoids constructing a useless CTE.

ibis/expr/operations/relations.py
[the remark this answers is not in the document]
cpcloud: I don't feel particularly strongly about these names, but if you do I think you're probably the one to rename them :)

ibis/expr/operations/relations.py
cpcloud: Moved this functionality to the base sqlglot backend.

ibis/backends/base/sqlglot/compiler.py
```diff
@@ -266,18 +266,31 @@ def translate(self, op, *, params: Mapping[ir.Value, Any]) -> sge.Expression:
         def fn(node, _, **kwargs):
             result = self.visit_node(node, **kwargs)
 
-            if node is op:
+            # if it's not a relation then we don't need to do anything special
+            if not isinstance(node, ops.Relation):
```
kszucs, line 270: FWIW I prefer the previous branching because it is easier to spot unhandled or wrongly handled operation types.

ibis/backends/base/sqlglot/__init__.py
```diff
@@ -151,6 +151,17 @@ def _get_schema_using_query(self, query: str) -> sch.Schema:
         """Return an ibis Schema from a backend-specific SQL string."""
         return sch.Schema.from_tuples(self._metadata(query))
 
+    def _get_sql_string_view_schema(self, child: ir.Table, query: str) -> sch.Schema:
+        # abort if somehow this was called with a non-view
+        assert isinstance(child.op(), ops.View), type(child.op())
+
+        obj = self._to_sqlglot(child)
+        dialect = self.compiler.dialect
+        parsed = sg.parse_one(query, read=dialect)
```
kszucs, line 160: Could we avoid the parsing here? I like to think `.sql()` as the escape hatch if something is not supported by ibis or sqlglot :)

ibis/backends/base/sqlglot/compiler.py
cpcloud: There were only ever two cases: Relations and Values. I think I prefer the dedented version slighty because the more complex code is not buried.

ibis/backends/base/sqlglot/rewrites.py
cpcloud: The list needs to be assembled in topological order, we can't assemble the View CTEs separately from the others.

ibis/backends/base/sqlglot/__init__.py
cpcloud: Not really, there could be multiple `with`s that need to be added to the parsed query.

kszucs (approved):

PR #8151, opened 2024-01-30
fix(sqlglot): stop using removed singletons for true, false, null

gforsyth:
Fix for upstream removal in https://github.com/tobymao/sqlglot/pull/2884

Should we set a stricter upper bound on sqlglot before releasing 8.x?

### DIFF
--- a/ibis/backends/base/sqlglot/__init__.py
+++ b/ibis/backends/base/sqlglot/__init__.py
@@ -82,9 +82,9 @@ def interval(value, *, unit):
 
 C = ColGen()
 F = FuncGen()
-NULL = sg.exp.NULL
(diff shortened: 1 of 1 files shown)


jcrist (approved):
Presuming tests pass :+1:

tparrishbard:
Hi, this error is still throwing when using the duckdb specific framework. (pip install 'ibis-framework[duckdb]')

gforsyth:
Hi @tparrishbard!  This fix hasn't been released yet, although it should be available in the next day or two.  In the meantime, you can manually downgrade `sqlglot` to workaround the issue by running `pip install -U sqlglot<=20.8.0`

Issue #7909, opened 2024-01-04
meta: Port backends to new relational operators/sqlglot


jcrist:
We currently have a [long running branch](https://github.com/ibis-project/ibis/tree/the-epic-split) porting backends to use the new relational operators (and sql backends to use sqlglot) introduced by:
- [x] #7752

This is a meta issue to track and organize remaining backends that need to be ported before that branch can be merged with main. Feel free to edit this list to keep it up to date.

**SQL Backends**

- [x] bigquery (#7968)
- [x] clickhouse (#7796)
- [x] druid (#8006)
- [x] duckdb (#7796)
- [x] exasol (#8032)
- [x] flink (#8268)
- [x] impala (#7993)
- [x] mssql (#7936)
- [x] mysql (#7926)
- [x] oracle (#8020)
- [x] postgres (#7877)
- [x] snowflake (#7853)
- [x] sqlite (#8154)
- [x] trino (#7871)
- [x] risingwave (#8171)

**Other Backends**

- [x] dask (#8005)
- [x] datafusion (#7846)
- [x] pandas (#7797)
- [x] polars (#7868)
- [x] pyspark (#7950)

jcrist:
I'll start working on the `sqlite` backend now to familiarize myself with the new sqlglot code path. @cpcloud also just pinged on zulip saying he's working on the `mysql` backend. AFAIK all other backends are still available for work.

cpcloud:
For Impala, we should be able to use the `hive` sqlglot dialect and supplement by adding rules to its `TRANSFORMS` `dict` (see the DataFusion backend for a custom dialect subclass).

cpcloud:
I am starting on MS SQL now.

kszucs:
Working on PySpark. Since the tests are really slow it will take some time.

cpcloud:
I will start on BigQuery today or tomorrow, depending on how much progress I can make with mssql

gforsyth:
I'll start on Oracle

cpcloud:
I'm starting on the impala backend today.

cpcloud:
I'll start on Druid today.

lostmygithubaccount:
**Weekly Update**:
2024-01-18

[...]

PR #8074, opened 2024-01-23
refactor(sql): remove sqlalchemy from the codebase

cpcloud:
This PR removes sqlalchemy from the codebase (except sqlite, until that is ported to sqlglot). In draft until sqlite is ported

PR #8154, opened 2024-01-30
refactor(sqlite): port to SQLGlot

jcrist:
Part of #7909.


ibis/backends/conftest.py
```diff
@@ -570,16 +570,16 @@ def ddl_con(ddl_backend):
     return ddl_backend.connection
 
 
-@pytest.fixture(params=_get_backends_to_test(keep=("sqlite",)), scope="session")
+@pytest.fixture(params=_get_backends_to_test(keep=()), scope="session")
 def alchemy_backend(request, data_dir, tmp_path_factory, worker_id):
     """Set up the SQLAlchemy-based backends."""
-    return _setup_backend(request, data_dir, tmp_path_factory, worker_id)
+    pytest.skip("No SQLAlchemy backends remaining")
```
kszucs, line 576: Yeah!

cpcloud: I am deleting these entirely in #8074 🎉

cpcloud:
I think the CI is hitting some versioning issues. I believe GHA has a slightly older version of sqlite that may not support full/right outer joins and some functions (looks like `UNHEX` maybe?)

ibis/backends/sqlite/__init__.py
```diff
@@ -95,169 +76,518 @@ def do_connect(
...
+        else:
+            return out.copy(nullable=nullable)
+
+    def _inspect_schema(
+        self, cur: sqlite3.Cursor, table_name: str, database: str | None = None
+    ) -> Iterator[tuple[str, dt.DataType]]:
+        if database is None:
+            database = "main"
+
+        sql = f'SELECT name, type, "notnull" FROM "{database}".pragma_table_info("{table_name}")'
```
kszucs, line 185: It might be preferable to use sqlglot queries so ensure that the quoting/escaping is right.

jcrist:
Yeah, I just saw that. I'll take a look to see what sqlalchemy was doing here.

ibis/backends/sqlite/__init__.py
```diff
@@ -95,169 +76,518 @@ def do_connect(
...
+        with self.begin() as cur:
+            return sch.Schema.from_tuples(
+                self._inspect_schema(cur, table_name, database)
+            )
+
+    def _metadata(self, query: str) -> Iterator[tuple[str, dt.DataType]]:
+        with self.begin() as cur:
             # create a view that should only be visible in this transaction
-            con.exec_driver_sql(f"CREATE TEMPORARY VIEW {view} AS {query}")
+            view = f"__ibis_sqlite_metadata_{util.guid()}"
```
kszucs, line 240:
```suggestion
            view = util.gen_name("ibis_sqlite_metadata")
```

ibis/backends/sqlite/tests/conftest.py
```diff
@@ -69,43 +71,19 @@ def con(data_dir, tmp_path_factory, worker_id):
...
-        .compile(dialect=dialect, compile_kwargs={"literal_binds": True})
-    )
-
-
-@pytest.fixture(scope="session")
-def sqla_compile(dialect):
-    return lambda expr: str(
-        expr.compile(dialect=dialect, compile_kwargs={"literal_binds": True})
-    )
+    return lambda expr: Backend.compiler.translator_class(expr, context).get_result()
```
cpcloud, line 79: Should this be `ibis.to_sql(..., dialect="sqlite")`?

ibis/backends/sqlite/compiler.py
```diff
@@ -1,36 +1,515 @@
...
+    @visit_node.register(ops.DateDelta)
+    @visit_node.register(ops.TimestampDelta)
+    @visit_node.register(ops.TryCast)
+    def visit_Undefined(self, op, **kwargs):
+        return super().visit_Undefined(op, **kwargs)
+
+    @visit_node.register(ops.Cast)
+    def visit_Cast(self, op, *, arg, to) -> sge.Cast:
+        if to.is_timestamp():
+            if to.timezone not in (None, "UTC"):
```
kszucs, line 96: Unrelated, but this reminds me that we should store the timezone in a normalized form in `dt.Timestamp` rather than an arbitrary string.

ibis/backends/tests/test_aggregation.py
```diff
@@ -1347,11 +1346,14 @@ def test_topk_op(alltypes, df):
...
 )
-def test_topk_filter_op(alltypes, df, result_fn, expected_fn):
+def test_topk_filter_op(con, alltypes, df, result_fn, expected_fn):
     # TopK expression will order rows by "count" but each backend
     # can have different result for that.
     # Note: Maybe would be good if TopK could order by "count"
     # and the field used by TopK
+    if con.name == "sqlite":
+        # TODO: remove after CTE extraction is reimplemented
+        pytest.skip("topk -> semi-join performance has increased post SQLGlot refactor")
```
cpcloud, line 1356: Is this still an issue?

ibis/backends/sqlite/compiler.py
```diff
@@ -1,36 +1,515 @@
...
+        elif to.is_date():
+            return self.f.date(arg)
+        elif to.is_time():
+            return self.f.time(arg)
+        return super().visit_Cast(op, arg=arg, to=to)
+
+    @visit_node.register(ops.Limit)
+    def visit_Limit(self, op, *, parent, n, offset):
+        # SQLite doesn't support compiling an OFFSET without a LIMIT, but
+        # treats LIMIT -1 as no limit
```
kszucs, line 113: Tricky!

cpcloud (approved):
LGTM, assuming tests pass!

ibis/backends/tests/test_join.py
```diff
@@ -181,7 +181,10 @@ def test_mutate_then_join_no_column_overlap(batting, awards_players):
         param(lambda left, right: left.join(right, "year", how="semi"), id="how_semi"),
     ],
 )
-def test_semi_join_topk(batting, awards_players, func):
+def test_semi_join_topk(con, batting, awards_players, func):
+    if con.name == "sqlite":
+        # TODO: remove after CTE extraction is reimplemented
```
kszucs, line 186: CTE extraction has been merged, so I think this can be removed.

cpcloud: I just checked out the PR, and this is still very slow so perhaps CTE extraction isn't sufficient to address the issue or we're not extracting enough?

ibis/backends/tests/test_aggregation.py
cpcloud: Still slow, so yes it's still an issue.

jcrist:
> Yeah, I just saw that. I'll take a look to see what sqlalchemy was doing here.

Ok, the failing tests are of two kinds:

- Some scalar builtin functions don't exist in older sqlite versions. This is easy to work around.
- RIGHT and FULL OUTER joins don't exist in older sqlite versions. This is harder to work around. I _think_ this could be done either in the compiler when handling the JoinChain or via a rewrite rule (e.g. rewrite right joins as left joins with the tables flipped), but a clean implementation is evading my tired brain. I'd appreciate some thought here.

cpcloud:
Feel free to xfail the outer join cases for older sqlite, not worth it at this point to put in the work for that especially if "upgrade to a later version" addresses it.

ibis/backends/sqlite/__init__.py
jcrist: AFAICT there's no way in sqlglot to namespace a function call like this (like `{possibly_quoted_namespace}.{func_name}(...)`).

I've updated the code to use sqlglot to manually quote the identifiers as needed everywhere string templating was used.

ibis/backends/sqlite/tests/conftest.py
jcrist: IIRC I blindly copied that from a different backend. Looks fine, but we may want to cleanup the fixtures uniformly later.

ibis/backends/tests/test_aggregation.py
jcrist: As you noted below, yes, this still is slow even with CTE extraction. I'm not sure what the it is about how we're generating this query now that doesn't play well with sqlites optimizer.

cpcloud:
I'll take care of the lockfile mismatch and then merge this!

Thank you!

PR #8172, opened 2024-01-31
chore: quote everything

cpcloud:
I kept hitting small issues with quoting, so I decided to enable quoting every identifer all the time. Yeah, the SQL is uglier, but I really do not want to keep dealing with quoting issues.


ibis/backends/base/sqlglot/datatypes.py
```diff
@@ -253,6 +253,8 @@ def _from_sqlglot_INTERVAL(
             else:
                 unit = precision_or_span.this.this
             return dt.Interval(unit=unit, nullable=nullable)
+        elif isinstance(precision_or_span, sge.Var):
+            return dt.Interval(unit=precision_or_span.this, nullable=nullable)
```
cpcloud, line 257: This was caught by a hypothesis test failure.

kszucs: Do we have hypothesis tests for the sqlglot conversions?

kszucs (approved):

PR #8192, opened 2024-02-01
fix(flink): fix the xfailed outer join test in flink backend

chloeh13q:
## Description of changes

Fix the xfailed outer join test in flink backend.


cpcloud:
This will be fixed in my port of the Flink backend to sqlglot!

Issue #8165, opened 2024-01-31
feat: UX: improve Interval API


NickCrews:
### Is your feature request related to a problem?

I have this function

```python
def dob_to_age(dob: DateValue) -> FloatingValue:
    """Convert a date of birth to an age in years (float)."""
    return (ibis.now().date() - dob).days.cast(int) / 365.25
```

I had several annoying experiences implementing it

1. Dates and Times use a method for getting a component, eg `date.month()`. Interval uses a property. We should be consistent. I would vote for methods. Perhaps there are other places where there is this inconsistency we should consider when making this decision?
1. The docstring and [docs for Interval.days](https://ibis-project.org/reference/expression-temporal#ibis.expr.types.temporal.IntervalValue.years) don't say the return type (IntervalValue), so I had to experiment to figure that out. If we moved to methods, the docs would show the return type, so that would be better. But still I think the docstring should also say this.
1. The cast to int is needed because otherwise I get `unsupported operand type(s) for /: 'IntervalScalar' and 'float'`. I expect `.days()` to give me an `IntegerValue`, like `Date.month()` and friends.
1. It would be nice if this was a oneliner `(ibis.now().date() - dob).years()`. But I get a `Cannot convert interval value from unit D to unit Y` with this. Which I think makes sense, because semantics to time are not strict and in some interpretations there aren't 365.25 days in a year. Not sure there is anything to do about this, but just wanted to flag this as something I'm running into. Not sure what backends do, but probably we are at their mercy and should just do what they do.

### Describe the solution you'd like

I think the best usage would be `(ibis.now().date() - dob).days() / 365.25`

1. move to methods
1. return IntegerValue
1. add return type in docstring
1. in class docstring, add note about limitations of casting units.

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
Thanks for the issue.

I think for the specific case of age and for any case where you need an integer number of some unit, give the [`delta`](https://ibis-project.org/reference/expression-temporal#ibis.expr.types.temporal.TimestampValue.delta) method a try.

Ideally your function could be:

```python
def dob_to_age(dob: DateValue) -> FloatingValue:
    """Convert a date of birth to an age in years (float)."""
	return ibis.now().delta(dob, "days") / 365.25
```

The other issues are somewhat orthogonal here:

[...]

PR #8193, opened 2024-02-01
docs: include return type in Interval component properties

NickCrews:
Fixes https://github.com/ibis-project/ibis/issues/8165

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -433,10 +433,6 @@ quartodoc:
               members:
                 - to_unit
                 - negate
-                - __add__
-                - __sub__
-                - __mul__
-                - __floordiv__
                 - years
                 - quarters
                 - months
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -428,7 +428,26 @@ quartodoc:
                 - strftime
                 - time
                 - truncate
-            - IntervalValue
+            - name: IntervalValue
+              dynamic: true
+              members:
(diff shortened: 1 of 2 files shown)


cpcloud:
@NickCrews Thanks for the PR. Is there any chance that instead of this approach we can add these methods and properties individually to `_quarto.yml` (likely requiring `dynamic: true` for each of them)?

NickCrews:
[This is relevant](https://github.com/machow/quartodoc/issues/284). Now trying to poke around with the
```python
class A:
    attr_a: int
    """Docstring."""
```
format

NickCrews:
I think this is still a gradual improvement, even if it isn't ideal. I think it shouldn't be bad to reverse or improve this later if this functionality changes. At the least, it adds `__add__` documentation, which was missing.

docs/_quarto.yml
```diff
@@ -428,7 +428,26 @@ quartodoc:
...
-            - IntervalValue
+            - name: IntervalValue
+              dynamic: true
+              members:
+                - to_unit
+                - negate
+                - __add__
+                - __sub__
+                - __mul__
+                - __floordiv__
```
cpcloud, line 439:
```suggestion
```

We're probably going to get rid of these methods soon, they're extremely finicky and don't really work as advertised.

cpcloud (changes requested):

cpcloud (approved):
Thanks!

Issue #8170, opened 2024-01-31
docs: note how UDFs dont *need* arg types


NickCrews:
### Please describe the issue

Per https://ibis-project.zulipchat.com/#narrow/stream/405265-tech-support/topic/UDF.20.28Array.29.20-.3E.20Floating/near/419125714,

I am trying to get access to duckdb's [list_avg](https://duckdb.org/docs/archive/0.9.2/sql/functions/nested#list-aggregates) function, which has the signature (array<numeric>) -> numeric

I have tried these various methods:

```python
from ibis.expr import datatypes as dt

# SignatureValidationError: ValueOf(<class 'list'>) has failed due to the following errors:
#   `dtype`: <class 'list'> is not either None or coercible to a DataType
#
# Expected signature: ValueOf(dtype: Optional[DataType] = None)
# @ibis.udf.scalar.builtin
# def list_avg(arr: list) -> float:
#     ...


# TypeError: 'AnnotableMeta' object is not iterable
#
# @ibis.udf.scalar.builtin(signature=((dt.Array), dt.Floating))
# def list_avg(arr):
#     ...

# SignatureValidationError: Array(value_type=<class 'ibis.expr.datatypes.core.Numeric'>) has failed due to the following errors:
#   `value_type`: <class 'ibis.expr.datatypes.core.Numeric'> is not coercible to a DataType
# Expected signature: Array(value_type: DataType, nullable: bool = True)
#
# @ibis.udf.scalar.builtin(signature=((dt.Array(value_type=dt.Numeric)), dt.Floating))
# def list_avg(arr):
#     ...
```

Phillip pointed out that you can just leave out the arg type. Ibis only really needs to know the return type:

```python
@ibis.udf.scalar.builtin
def list_avg(arr) -> float:
    ...
```

We should document that this is the recommended solution if you run into this. Could really go over the top and have this suggestion in the error message. Should still recommend people include the arg type when possible, because that is better documentation?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

NickCrews:
Can we link to that how to guide from the reference docs? I was only looking at the reference and never found that guide. But that guide looks great!

NickCrews:
I can submit a PR

PR #8194, opened 2024-02-01
docs: improve UDF signature docs

NickCrews:
Also added callout to optional args signature for builtin UDFs. Didn't for the other UDF implementations.

Fixes https://github.com/ibis-project/ibis/issues/8170#issuecomment-1921802611


lostmygithubaccount:
the docs build error means you didn't check in (after rendering/previewing) the frozen execution outputs for the doc, so it's trying to run them

cpcloud:
It looks like we are running snowflake queries in this part of the docs so for now someone with credentials will need to run it. In the meantime we should ideally move the snowflake stuff to a different document or remove it.

NickCrews:
(not looking at the code as I say this) can we just switch the examples to duckdb or something the doesn't need creds?

NickCrews:
Does the rest of the changes look good though?

docs/how-to/extending/builtin.qmd
```diff
@@ -164,7 +161,9 @@ con.execute(cardinality([1, 2, 3]))
 con.execute(cardinality(["a", "b"]))
 ```
 
-When you bypass input types the errors you get back are backend dependent:
+When you skip giving input types, Ibis isn't able to catch your typing errors
```
cpcloud, line 164:
```suggestion
When you do not specify input types, Ibis isn't able to catch typing errors
```

NickCrews:
I removed/ported the snowflake stuff. we cab revert that if you want, but I think this is an improvement.

docs/how-to/extending/builtin.qmd
```diff
@@ -164,14 +116,16 @@ con.execute(cardinality([1, 2, 3]))
...
+early, and they are only caught during execution.
+The errors you get back are backend dependent:
 
 ```{python}
 #| error: true
 con.execute(cardinality("foo"))
 ```
 
-Here, Snowflake is informing us that the `ARRAY_SIZE` function does not accept
+Here, Duckdb is informing us that the `ARRAY_LENGTH` function does not accept
```
cpcloud, line 128:
```suggestion
Here, DuckDB is informing us that the `ARRAY_LENGTH` function does not accept
```

cpcloud:
@NickCrews Nah, let's keep it, this is strictly an improvement! Thank you!

cpcloud (approved):

cpcloud:
LGTM: https://pr-8194-b47c22b9f167a7f19b7670f18528cf8848007ad7--ibis-quarto.netlify.app/how-to/extending/builtin

Thanks again!

PR #8195, opened 2024-02-01
refactor(ir): remove the decimal precision promotion logic

### LINKED ISSUE #8189: bug(bigquery):  Numeric precision increase 1 unexpectedly after every sub/add operations
### What happened?

The numeric precision get automacially increase 1 per sub/add operations unexpectedly, though BQ only supports `precision=38` for `NUMERIC` and `precision=76` for `BIGNUMERIC`.

The code to repro the issue:

```
>>> import ibis
>>> bq = ibis.bigquery.connect(project_id="bigframes-dev")
>>> taxi_dt = bq.table("bigquery-public-data.new_york_taxi_trips.tlc_yellow_trips_2022")
>>> taxi_dt['trip_distance'].type()
Decimal(precision=38, scale=9, nullable=True)
>>> result = taxi_dt['trip_distance']+taxi_dt['trip_distance']
>>> result.type()
Decimal(precision=39, scale=9, nullable=True)
```
The expected `result.type()` would be `Decimal(precision=38, scale=9, nullable=True)`

7.1.0

BigQuery

### Relevant log output

_No response_

### Code of Conduct

chelsea-lin:
## Description of changes
This change is removing the decimal precision promotion logic from the unbounded ibis expr. The logic was introduced by #4330 so the decimal precision can matches the [DB2](https://www.ibm.com/docs/en/i/7.1?topic=operators-decimal-arithmetic-in-sql). However, this rule is not generic for all. For example, BQ only support `precision=38`. 

## Issues closed

* Resolves #8189

### DIFF
--- a/ibis/expr/rules.py
+++ b/ibis/expr/rules.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import operator
 from itertools import product, starmap
 from typing import Optional
 
@@ -108,39 +107,6 @@ def _promote_integral_binop(exprs, op):
     return dt.highest_precedence(dtypes)
 
 
-def _promote_decimal_binop(args, op):
-    if len(args) != 2:
-        return highest_precedence_dtype(args)
-
-    # TODO: Add support for setting the maximum precision and maximum scale
-    left = args[0].dtype
-    right = args[1].dtype
-
-    max_prec = 31 if left.precision <= 31 and right.precision <= 31 else 63
-    max_scale = 31
-
-    if op is operator.mul:
-        return dt.Decimal(
-            min(max_prec, left.precision + right.precision),
-            min(max_scale, left.scale + right.scale),
-        )
-    elif op is operator.add or op is operator.sub:
-        return dt.Decimal(
-            min(
-                max_prec,
-                max(
-                    left.precision - left.scale,
-                    right.precision - right.scale,
-                )
-                + max(left.scale, right.scale)
-                + 1,
-            ),
-            max(left.scale, right.scale),
-        )
-    else:
-        return highest_precedence_dtype(args)
-
-
 @public
 def numeric_like(name, op):
     @attribute
@@ -149,8 +115,6 @@ def numeric_like(name, op):
         dtypes = [arg.dtype for arg in args]
         if util.all_of(dtypes, dt.Integer):
             result = _promote_integral_binop(args, op)
-        elif util.all_of(dtypes, dt.Decimal):
-            result = _promote_decimal_binop(args, op)
         else:
             result = highest_precedence_dtype(args)
 
--- a/ibis/tests/expr/test_decimal.py
+++ b/ibis/tests/expr/test_decimal.py
@@ -36,15 +36,15 @@ def test_promote_decimal_type_mul(lineitem):
     col_1 = lineitem.l_extendedprice
     col_2 = lineitem.l_discount
     result = col_1 * col_2
-    assert result.type().precision == 24
-    assert result.type().scale == 4
+    assert result.type().precision == 12
+    assert result.type().scale == 2
 
 
 def test_promote_decimal_type_add(lineitem):
     col_1 = lineitem.l_extendedprice
     col_2 = lineitem.l_discount
     result = col_1 + col_2
-    assert result.type().precision == 13
+    assert result.type().precision == 12
     assert result.type().scale == 2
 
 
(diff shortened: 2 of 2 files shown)


ibis/expr/rules.py
```diff
@@ -149,8 +149,6 @@ def dtype(self):
         dtypes = [arg.dtype for arg in args]
         if util.all_of(dtypes, dt.Integer):
             result = _promote_integral_binop(args, op)
-        elif util.all_of(dtypes, dt.Decimal):
-            result = _promote_decimal_binop(args, op)
```
kszucs, line 153: Can you remove the now unused `_promote_decimal_binop`?

cpcloud:
@webmiche Thoughts on dropping the type evolution here? It seems to be causing an issue for BigQuery which has much simpler promotion rules than DB2.

kszucs:
@chelsea-lin I rebased your branch and retargeted the PR agains `main`.

webmiche:
> @webmiche Thoughts on dropping the type evolution here? It seems to be causing an issue for BigQuery which has much simpler promotion rules than DB2.

For context, I have not been working a lot with databases and ibis lately given that my project finished.

I think this touches the exact core discussion we had back in the initial PR: type inference rules are backend-specific whereas ibis tries to build data structures that are as generic as possible. And I guess what we have here is exactly that, a backend that does not follow the same type inference rules.

I think there is a range of possibilities of what could be done for this case. From the top of my head:

- inferring type rules according to the backend
- define ibis specific type rules and add clue code for backends
- make decimals "opaque" to some extent, i.e. make them not have a precision and scale at all (at least in some cases where this info is not directly encoded by the frontend)

I personally advise against inferring type inference rules from backend choice as this somewhat defeats the purpose of ibis in the first place. Also, the ibis specific type rules sound like just another specification on top of all the others, so that's no real solution.

I think the third choice is probably most reasonable. Honestly, I don't know exactly how the ibis data structures work currently, so I can't tell the ramifications of such a change. Maybe deleting the type inference/binop promotion-functionality already achieves this to some extent. Maybe, this change is only the first towards handling this in a principled way.

So, all in all, I am fine with dropping it. If I was still working on the project I implemented the type inference for originally, I would probably add some clue code to handle this in between the ibis data structures and my data structures. For this, I would definitely appreciate it if ibis communicates to me which decimals are defined by the frontend and which it is not sure about. The "opaque" decimal solution would achieve this, AFAIU.

Maybe @ingomueller-net has more thoughts on this.

cpcloud:
@webmiche Thanks for the response! I was thinking something similar re opaque decimals. We took this approach a long time ago with strings, and it has not proven to be an issue while also allowing us to mostly ignore the various `CHAR(N)`/`VARCHAR(N)` string types.

I think we'd have to implement this and try it out on a few backends to see what its blast radius would be. I suspect there might be some issues at the Ibis <-> pyarrow boundary.

ibis/expr/rules.py
chelsea-lin: Done!

chelsea-lin:
> @chelsea-lin I rebased your branch and retargeted the PR against `main`.
I also need the commit in `the-epic-split` branch. What's the process to do that?

kszucs:
We keep `the-spic-split` rebased on top of the `main` branch, so this change will reach both branches.

cpcloud:
Given this isn't break anything I'm inclined to merge it. @kszucs Thoughts?

kszucs (approved):
LGTM

kszucs:
> Given this isn't breaking anything I'm inclined to merge it. @kszucs Thoughts?

Agree.

ingomueller-net:
Thanks, everybody, for the work on Ibis and this issue!

What I have asked myself in the past due to this and similar issues is: what is the degree of portability that Ibis aims for? Since the precision promotion logic is different in some backends, Ibis programs may result in different results when run on different backends. Is that expected? Is there a place where these differences are tracked? When do we consider a backend as "correct" if we allow these kind of differences?

ingomueller-net:
To add to the issue of aggregations, note that the result of any non-associative function depends on the input order, including `first`, `last`, `array_agg` (not sure if it exists), `sum` on floats, but also `sum` on integer in most cases (since it may depend on the order whether or not an overflow occurs).

And then the result of most window functions obviously depends on the order.

Again, all of these are non-deterministic in many SQL systems, so you don't get a guarantee to obtain the same result from run to run *against the same system instance* there either. I, thus, think it'd be perfectly valid to declare all possible orderings as "correct." However, it has to be clear what *exactly* the semantics of each operation are, possibly referring to non-determinism for some operations.

ingomueller-net:
Oh, and another pointer that came up when I asked the same question [on substrait](https://github.com/substrait-io/substrait/issues/596): The semantics of `ORDER BY` aren't the same in every system, for example, for `NaN`s which could come before or after numbers.

PR #8202, opened 2024-02-02
chore(poetry): add strict upper bound to sqlglot

gforsyth:
SQLGlot has a non-standard view of semver, documented here:
https://github.com/tobymao/sqlglot#versioning

TLDR: it's BIG_BREAKING.BREAKING.NONBREAKING

so we should hard pin major.minor.

Worth noting that this may not be strict enough.  It's unclear to me
whether the contents of `sqlglot.expressions` are considered public.  I
will ask.

-------


gforsyth:
Confirmed via the sqlglot slack (not going to bother with a link because it's slack) that the contents of `sqlglot.expressions` are considered public unless objects are prefixed with an underscore.

jcrist (approved):

PR #8161, opened 2024-01-31
ci: re-enable pre-commit and other non-docs lints for `the-epic-split`

cpcloud:
Re-enable disabled tooling for the epic split.

Closes #8152.

PR #8215, opened 2024-02-04
fix(api): selection using a selector yielding multiple columns

kszucs:
Discovered and fixed by @cpcloud: https://github.com/ibis-project/ibis/pull/8161#discussion_r1477098975
This PR covers it with a test case.


cpcloud (approved):
Thank you :)

PR #8227, opened 2024-02-05
fix(api): restore and deprecate `ir.Table.to_array()`

kszucs:


.github/workflows/ibis-docs-main.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - "*.x.x"
+      - "the-epic-split"
```
cpcloud, line 8: Definitely do not want this here. It will publish docs from all of these branches. While you're at can you remove the `"*.x.x"`.

It's enough to build on PRs.

cpcloud (changes requested):

cpcloud:
Actually, I can push up the CI changes.

cpcloud (approved):

PR #8234, opened 2024-02-05
ci: allow codecov uploads to fail

cpcloud:
## Description of changes

Currently our CI fails when codecov fails to upload, which has been happening more frequently after
the upgrade to `codecov-action@v4`.

This PR allows the codecov step to fail to upload, since:

1. It's not a useful enough indicator of a failed build.
2. It's likely to pass on the next run, as it's often a transient issue.


gforsyth (approved):
:crossed_fingers: that codecov passes

Issue #5964, opened 2023-04-11
feat: DDL Materialized Views


dpprdan:
### Is your feature request related to a problem?

As mentioned in #5962 it would be nice to have DDL (`create`/`refresh`/`drop`) for materialized views.

### Describe the solution you'd like

Backends supported by ibis that support materialized views:

1. [PostgreSQL](https://www.postgresql.org/docs/current/rules-materializedviews.html) ([`create`](https://www.postgresql.org/docs/current/sql-creatematerializedview.html), [`refresh`](https://www.postgresql.org/docs/current/sql-refreshmaterializedview.html), [`drop`](https://www.postgresql.org/docs/current/sql-dropmaterializedview.html))

2. [BigQuery](https://cloud.google.com/bigquery/docs/materialized-views-intro) ([`create`](https://cloud.google.com/bigquery/docs/materialized-views-create), [`refresh`](https://cloud.google.com/bigquery/docs/materialized-views-manage#refresh), [`drop`](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#drop_materialized_view_statement))

3. [Snowflake](https://docs.snowflake.com/en/user-guide/views-materialized) ([`create`](https://docs.snowflake.com/en/sql-reference/sql/create-materialized-view), `refresh` 🤷🏻‍♂️, [`drop`](https://docs.snowflake.com/en/sql-reference/sql/drop-materialized-view))

4. [ClickHouse](https://clickhouse.com/docs/en/guides/developer/cascading-materialized-views) ([`create`](https://clickhouse.com/docs/en/sql-reference/statements/create/view#materialized-view), `refresh` 🤷🏻‍♂️, [`drop`](https://clickhouse.com/docs/en/sql-reference/statements/drop#drop-view))

5. Trino ([`create`](https://trino.io/docs/current/sql/create-materialized-view.html), [`refresh`](https://trino.io/docs/current/sql/refresh-materialized-view.html), [`drop`](https://trino.io/docs/current/sql/drop-materialized-view.html))

[...]

PR #8239, opened 2024-02-06
feat(risingwave): add streaming DDLs

KeXiangWang:
## Description of changes

Risingwave plays a role as a computing engine and storage system in Streaming ecosystems.
Usually, a streaming workload will include two/three systems like this:
`upstream data source --> Risingwave`
Or,
`upstream data source --> Risingwave --> downstream data sink`

This PR adds streaming support, mainly new relations related to streaming, for the Risingwave backend.

To be specific, Four types of new relations are introduced:
- `Source`, data sources that encompass a connector connected to an upstream data system like Kafka. A source works like an extraction or a placeholder for an upstream system, and although it has columns, it does not store any data itself.
- `Materialized View`, which is Risingwave's core concept for streaming. One can create an MV with a query on existing tables or sources. The data in MV is automatically updated in a real-time way. Users can access the data with a `select` statement just like accessing a table.
- `Table` with connector, works like a combination of a source and a normal table. The difference between `Table` with connector and `Source` is that, once the table is created, it will automatically start to consume data from upstream systems and update it into Risingwave.
- `Sink`, unlink a `Materialized View` which stores the result of a query in RW, users can choose to sink the result to a downstream system, e.g. Redis, and then read the data in the downstream system. Unlike an MV, User cannot access the data directly from a sink.

Some minor changes:
1. Pump the image version to a new nightly build, as the new image solves some issues found in the [previous PR](https://github.com/ibis-project/ibis/pull/7954).
2. Pump the `sqlalchemy-risingwave` version from `1.0.0` to `1.0.1`, some implementations are updated.
3. Mark one test, `test_semi_join`, as skipped for risingwave backend. The test sometimes stuck and it seems to be Risingwave's fault. I'll continue to investigate.

One issue introduced in this PR:
Risingwave backend's implementation is sqlalchemy-based. But sqlalchemy has no corresponding concept for sources and sinks. So, in order to work around this issue, I temporarily categorize a source as a view. So sqlalchemy can access a source's metadata (names and types of its columns) like a view. However, this causes a side effect, when users call `list_view`, they may see some unexpected views that are actually sources. This can be fixed by rewriting the implementation of the `list_view` func in the Risingwave Backend. I'll fix it later.


KeXiangWang:
It seems I have no access to invite reviewers. @gforsyth @cpcloud @chloeh13q @deepyaman Could you please help take a review or invite the appropriate ones?

jcrist:
Hi @KeXiangWang - thanks for opening this! 

We're unfortunately in the midst of a large internals refactor moving all backends to use `sqlglot` instead of `sqlalchemy`. We hope to merge https://github.com/ibis-project/ibis/tree/the-epic-split into `main` soon (maybe by end of week?) after which we'll be back to accepting PRs to `main`. Until that's done though we don't really want to merge additional large PRs as this would complicate getting the refactor in. We'll follow up after `the-epic-split` is merged to discuss how best to implement the work here using `sqlglot`, but until then things are blocked on us finishing up the internals refactor.

zhenzhongxu:
@chloeh13q @cpcloud: bump this up, is this something we can help wth getting the sqlglot refactoring?

KeXiangWang:
> It looks like the first order of business would be to rebase this PR on main.

OK, I'll try my best.

KeXiangWang:
Hi @cpcloud, I've rebased the PR to newest main and fix all the issue introduced by the `sqlglot` refactoring. Could you please help take a look.

Besides, I find some tests are unable to pass even with the main branch, so I leave them unchanged for now. Do you have any ideas? Is there any configuration I missed?

ibis/backends/risingwave/__init__.py
```diff
@@ -257,3 +311,364 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
+        schema: str | None = None,
+    ) -> None:
+        """Creating a sink.
+
+        Parameters
+        ----------
+        name
+            Sink name to Create.
+        sink_from
+            The table or materialized view name to sink from. Only one of `sink_from` or `ob` can be
```
gforsyth, line 568:
```suggestion
            The table or materialized view name to sink from. Only one of `sink_from` or `obj` can be
```

ibis/backends/risingwave/__init__.py
```diff
@@ -257,3 +311,364 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
+            ""
+            if encode_properties is None
+            else " " + self._format_properties(encode_properties)
+        )
+
+    def _format_properties(self, props):
+        tokens = []
+        for k, v in sorted(props.items()):
+            tokens.append(f"{k}='{v}'")
+        return "( {} ) ".format(", ".join(tokens))
```
gforsyth, line 674: You might as well combine these into a single method if the first is only doing a check for `None`

ibis/backends/risingwave/__init__.py
```diff
@@ -257,3 +311,364 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
+    def _encode_properties_to_str(self, encode_properties):
+        return (
+            ""
+            if encode_properties is None
+            else " " + self._format_properties(encode_properties)
+        )
+
+    def _format_properties(self, props):
+        tokens = []
+        for k, v in sorted(props.items()):
```
gforsyth, line 672: Do these need to be sorted?  If not, I think we should render them in the order they were provided.

gforsyth (commented):
Hey @KeXiangWang -- thanks for putting this together and sorry for the delay in reviewing!

I've taken first pass over the new DDL stuff.  I am not a streaming expert, so maybe @chloeh13q and @deepyaman can also chime in on some of the API design stuff.

Since some of these methods may become "standard" on other (future or current) streaming backends we want to be very deliberate in how we design them.  We especially want to make sure that we don't overindex on a single backend when coming up with names (this is very difficult).

ibis/backends/risingwave/__init__.py
```diff
@@ -257,3 +311,364 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
+            this=table,
+            kind="MATERIALIZED VIEW",
+            expression=self.compile(obj),
+        )
+        self._register_in_memory_tables(obj)
+
+        this = sg.table(name, catalog=database, quoted=self.compiler.quoted)
+        with self._safe_raw_sql(create_stmt) as cur:
+            if overwrite:
+                cur.execute(
```
mfatihaktas, line 399: Similar to other comments, can we not use `drop_materialized_view()` for this?

ibis/backends/risingwave/__init__.py
```diff
@@ -257,3 +311,364 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
+
+    def create_materialized_view(
+        self,
+        name: str,
+        obj: ir.Table,
+        *,
+        database: str | None = None,
+        schema: str | None = None,
+        overwrite: bool = False,
+    ) -> ir.Table:
```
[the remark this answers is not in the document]
chloeh13q: I think this information may be helpful in helping us decide on a suitable API: https://materialize.com/guides/materialized-views/#how-do-materialized-views-work-in-specific-databases

Personally I think it depends on whether people tend to think of materialized views as a special type of views, or completely separate from normal views. It sounds like several backends just treat materialized views as a special type of views, as opposed to "its own thing".

- Postgres: https://www.postgresql.org/docs/current/rules-materializedviews.html
- Materialize: https://materialize.com/guides/materialized-views/#materialized-view-example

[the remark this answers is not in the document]
chloeh13q: Also wondering: 

I saw that there is a ticket for implementing materialized views for a broader set of backends: https://github.com/ibis-project/ibis/issues/5964 - is this an API we'd want to include in the base class?

chloeh13q:
I have some open questions on a high level:

I understand the motivation behind introducing separate APIs for sources, materialized views, tables, and sinks for the RW backend. Sources, materialized views, tables, and sinks are distinct objects in RW. However, I wonder how generalizable this is across different streaming backends.

For the user who is already familiar with RW's API, this implementation feels natural. But for the user who is coming from another backend, it may cause some confusion.

For example, Flink also has the same concepts, but it doesn't use a different API for each. Source, tables, and sinks are all created with `CREATE TABLE`. A lot of these underlying differences are abstracted away from the user.

Of course, it's also okay that we just do this for the RW backend for now and refactor at a later point if there is a need to consolidate some of the APIs.

ibis/backends/risingwave/__init__.py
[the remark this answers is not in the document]
KeXiangWang: It's easy to make `materialized view` a special type if `view` in this PR. But I would say they should be different objects in databases. The underlying mechanism and performance are largely different. For streaming databases like Risingwave and Materialize, `Materialized view` is the core concept and hardest part to design, while `view` is much more trivial. So we would like to keep `materialized view` and `view` separate.

ibis/backends/risingwave/__init__.py
```diff
@@ -24,6 +25,23 @@
...
+    UPSERT = 4
+    DEBEZIUM_MONGO = 5
+
+
+class EncodeFormat(enum.Enum):
+    CSV = 0
+    JSON = 1
+    AVRO = 2
+    PROTOBUF = 3
+    BYTES = 4
```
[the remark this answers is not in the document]
KeXiangWang: Got it. Updated.

ibis/backends/risingwave/__init__.py
```diff
@@ -106,6 +124,11 @@ def create_table(
         obj: pd.DataFrame | pa.Table | ir.Table | None = None,
         *,
         schema: ibis.Schema | None = None,
+        # TODO(Kexiang): add `append only`
+        connector_props: dict | None = None,
```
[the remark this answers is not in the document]
KeXiangWang: updated it to tbl_properties

ibis/backends/risingwave/__init__.py
```diff
@@ -106,6 +124,11 @@ def create_table(
...
+        data_format: DataFormat | None = None,
+        encode_format: EncodeFormat | None = None,
+        encode_properties: dict | None = None,
```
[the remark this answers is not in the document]
KeXiangWang: Now, they are `tbl_properties ` and `encode_properties `.

ibis/backends/risingwave/__init__.py
```diff
@@ -122,6 +145,15 @@ def create_table(
...
             `obj` or `schema` must be specified
+        connector_props
+            The properties of the sink connector, providing the connector settings to push to the downstream data sink.
+            Refer https://docs.risingwave.com/docs/current/data-delivery/ for the required properties of different data sink.
+        data_format
+            The data format for the new source, e.g., DataFormat.PLAIN.
+        encode_format
+            The encode format for the new source, e.g., EncodeFormat.JSON.
+        encode_properties
+            The properties of encode format. Refer https://docs.risingwave.com/docs/current/sql-create-source/ for more details.
```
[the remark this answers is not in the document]
KeXiangWang: Tried to add more description..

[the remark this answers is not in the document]
KeXiangWang: Updated, now they are at the end of the kward-list.

ibis/backends/risingwave/__init__.py
```diff
@@ -257,3 +311,364 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
+        """
+        src = sge.Drop(
+            this=sg.table(
+                name, db=schema, catalog=database, quoted=self.compiler.quoted
+            ),
+            kind="MATERIALIZED VIEW",
+            exists=force,
+        )
+        with self._safe_raw_sql(src):
+            pass
```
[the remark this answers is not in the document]
KeXiangWang: Depends on how we consider `materializedv view` vs `view.` Personally, I'm inclined to keep `drop view` and `drop materialized view` as two independent functions if we have `create view` and `create materialized view` separate.

ibis/backends/risingwave/__init__.py
```diff
@@ -257,3 +311,364 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
+
+    def create_source(
+        self,
+        name: str,
+        schema: ibis.Schema,
+        connector_props: dict,
+        data_format: DataFormat,
+        encode_format: EncodeFormat,
+        *,
```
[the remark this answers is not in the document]
KeXiangWang: updated.

ibis/backends/risingwave/__init__.py
KeXiangWang: updated

ibis/backends/risingwave/__init__.py
KeXiangWang: `sorted` removed

ibis/backends/risingwave/__init__.py
KeXiangWang: 👀 Do you mean we should use `drop view`? Or you think we should use `drop_materialized_view()` here.

KeXiangWang:
> I wonder how generalizable this is across different streaming backends.
Good question. These concepts are common in streaming backends, although different systems may have different names for them. For example, as Flink doesn't have a normal persistent table in a traditional database, its tables are used to express a streaming job. While in RW, Materialize, and Timeplus, we use `materialized views`. For sources, Flink combines a streaming job and the dependent sources together as a Table, while RW and materialize use `sources` objects and Timeplus use `stream` objects.  
> Flink also has the same concepts, but it doesn't use a different API for each.
Flink positions itself as a 'compute engine,' which means it's designed not to store data but to process it. In Flink, a `table` is essentially a streaming job that interacts with sources or sinks. The outcomes of these jobs are not stored within Flink itself. In contrast, streaming databases such as RW, Materialize, and Timeplus, function as databases that store the results locally. While Flink tables are declared as `Tables`, you cannot directly query data from them. Instead, you must route the data to an external system to access the results. With Flink, you trade off a simpler API for the flexibility and simplicity of its architecture. On the other hand, streaming databases offer the ability to customize your streaming jobs by combining various objects to suit your needs, providing direct accessing or sinking to external systems at the same time.

ibis/backends/risingwave/__init__.py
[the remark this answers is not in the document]
zhenzhongxu: Do we also want to distinguish between incremental materialized view?

ibis/backends/risingwave/__init__.py
mfatihaktas: I meant using the appropriate function, ``drop_view()`` or `drop_materialized_view()`, instead of the `sqlglot` expression:
```
sge.Drop(kind="MATERIALIZED VIEW", this=this, exists=True).sql(
                        self.dialect
                    )
```
IMO, `sqlglot` expressions look complex and replacing them with calls to the appropriate functions would improve readability.

KeXiangWang: got it. updated

ibis/backends/risingwave/__init__.py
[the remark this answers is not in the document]
cpcloud: I don't think we should distinguish here without a clear explanation of what the difference between incremental and non-incremental streaming materialized views are.

It's also adding more scope than is needed for now.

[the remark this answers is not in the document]
cpcloud: I think we should keep this API separate for now, it's likely that materialized views will require some additional kwargs that don't apply to views. The word "view" here is really being stretched to its limit by this functionality and thinking about materialized views as another kind of view isn't really correct.

ibis/backends/risingwave/__init__.py
```diff
@@ -109,6 +158,11 @@ def create_table(
         database: str | None = None,
         temp: bool = False,
         overwrite: bool = False,
+        # TODO(Kexiang): add `append only`
+        tbl_properties: dict | None = None,
+        data_format: str | None = None,
+        encode_format: str | None = None,
+        encode_properties: dict | None = None,
```
cpcloud, line 165: Are all of these nouns the same ones used in the RisingWave documentation? If not, we should make them match and avoid introducing new nouns just for the Ibis API.

ibis/backends/risingwave/__init__.py
```diff
@@ -24,6 +24,55 @@
...
+    if format_str is None:
+        return
+    upper = format_str.upper()
+    if upper not in valid_data_formats:
+        raise ValueError(
+            f"Invalid data format: {format_str}. Allowed formats are {', '.join(valid_data_formats)}"
+        )
+
+
+def validate_encode_format(format_str: str):
```
cpcloud, line 48: Is all this validation necessary? What if a new data format or encoding is added? Can't we let this fall through to the driver to avoid kneecapping users who upgrade their risingwave installation?

ibis/backends/risingwave/__init__.py
```diff
@@ -257,3 +346,311 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
             with self.begin() as cur:
                 cur.execute(create_stmt_sql)
                 extras.execute_batch(cur, sql, data, 128)
+
+    def exec_driver_sql(self, query: str):
```
cpcloud, line 350: Why do you need this additional method? Try to avoid adding public APIs like this.

chloeh13q:
@KeXiangWang Yep that makes sense. Since Ibis works across engines & databases, I'd imagine that these are questions that Ibis will need to address. But I think it also depends on the background of the user, e.g., whether it's someone who's experimenting with streaming or someone who already. has some expertise in streaming and wants to try out the Ibis API. In any case, these may be questions that we can leave open and come back to, when there is user feedback backing up one way or another.

ibis/backends/risingwave/__init__.py
KeXiangWang: It would be more natural, if we replace `tbl_properties` with `connector_properties`, but it will differentiate from Flink's corresponding nouns.

KeXiangWang: Do you think it's appropreciate to rename it to `connector_properties`?

ibis/backends/risingwave/__init__.py
KeXiangWang: Make sense. removed.

ibis/backends/risingwave/__init__.py
KeXiangWang: Sorry, it should have been deleted. Removed just now.

ibis/backends/risingwave/__init__.py
gforsyth: If that matches the Risingwave terminology, then I think that is a good naming choice.

gforsyth:
@KeXiangWang -- we've just merged in #8655 which moves us away from using the word `schema` in any hierarchical sense.  We're standardizing on "database" as a collection of tables and "catalog" as a collection of database.
If this PR allows maintainers to push updates, I can take on porting over your work here to the new naming and push that up.

KeXiangWang:
> If this PR allows maintainers to push updates, I can take on porting over your work here to the new naming and push that up.

Thx @gforsyth . I really cannot fiind a button in this page to enable it, so I create a new PR [here](https://github.com/ibis-project/ibis/pull/8781). We can push the following updates to the new PR. Thanks!

gforsyth:
I've pushed up the fixes to #8781 -- @KeXiangWang, you can either cherry-pick that commit back to this PR, or we can close this PR out in favor of your newer one, whichever you prefer!

KeXiangWang:
Hi @gforsyth . Thanks for the effort. I've update the patch to this PR and I'll close https://github.com/ibis-project/ibis/pull/8781 when this PR is merged. Thanks!

KeXiangWang:
@cpcloud @gforsyth Any further comments? Can we merge this PR now?

ibis/backends/risingwave/__init__.py
```diff
@@ -268,3 +316,261 @@ def list_databases(
             databases = list(map(itemgetter(0), cur))
 
         return self._filter_with_like(databases, like)
+
+    def create_materialized_view(
```
cpcloud, line 320: Can you add `@experimental` here? This will indicate that the method may change between non major releases.

KeXiangWang: added.

ibis/backends/risingwave/__init__.py
```diff
@@ -131,22 +155,37 @@ def create_table(
...
         if obj is None and schema is None:
             raise ValueError("Either `obj` or `schema` must be specified")
 
-        if database is not None and database != self.current_database:
-            raise com.UnsupportedOperationError(
-                f"Creating tables in other databases is not supported by {self.name}"
+        if connector_properties is not None and (
+            encode_format is None or data_format is None
+        ):
+            raise com.RelationError(
```
gforsyth, line 179:
```suggestion
            raise com.UnsupportedArgumentError(
```

I think this is more appropriate here, but not a blocker

ibis/backends/risingwave/__init__.py
```diff
@@ -268,3 +317,262 @@ def list_databases(
...
+        name
+            Sink name to Create.
+        sink_from
+            The table or materialized view name to sink from. Only one of `sink_from` or `obj` can be
+            provided.
+        connector_properties
+            The properties of the sink connector, providing the connector settings to push to the downstream data sink.
+            Refer https://docs.risingwave.com/docs/current/data-delivery/ for the required properties of different data sink.
+        obj
+            An Ibis table expression or pandas table that will be used to extract the schema and the data of the new table. Only one of `sink_from` or `obj` can be provided.
```
gforsyth, line 520: this should probably be pandas DataFrame and the type hint above should also include `pd.DataFrame` -- however, is this supported / tested?

ibis/backends/risingwave/__init__.py
```diff
@@ -268,3 +317,262 @@ def list_databases(
...
+            The data format for the new source, e.g., "PLAIN". data_format and encode_format must be specified at the same time.
+        encode_format
+            The encode format for the new source, e.g., "JSON". data_format and encode_format must be specified at the same time.
+        encode_properties
+            The properties of encode format, providing information like schema registry url. Refer https://docs.risingwave.com/docs/current/sql-create-source/ for more details.
+        """
+        table = sg.table(name, db=database, quoted=self.compiler.quoted)
+
+        if encode_format is None != data_format is None:
+            raise com.RelationError(
```
gforsyth, line 533:
```suggestion
            raise com.UnsupportedArgumentError(
```

ibis/backends/risingwave/__init__.py
```diff
@@ -268,3 +317,262 @@ def list_databases(
...
+                "When creating sinks, both encode_format and data_format must be provided, or neither should be"
+            )
+
```
gforsyth, line 536: I think we should raise here if both `sink_from` and `obj` are provided -- currently this ignores `obj` silently if `sink_from` is provided.

ibis/backends/risingwave/__init__.py
```diff
@@ -268,3 +317,262 @@ def list_databases(
...
+        data_format
+            The data format for the new source, e.g., "PLAIN". data_format and encode_format must be specified at the same time.
+        encode_format
+            The encode format for the new source, e.g., "JSON". data_format and encode_format must be specified at the same time.
+        encode_properties
+            The properties of encode format, providing information like schema registry url. Refer https://docs.risingwave.com/docs/current/sql-create-source/ for more details.
+        """
+        table = sg.table(name, db=database, quoted=self.compiler.quoted)
+
+        if encode_format is None != data_format is None:
```
gforsyth, line 532: I think you want an `and` here -- as it's currently written I think this will error if both kwargs are `None`

```python
None is None != None is None
True != None is None
False is None
False
```

ibis/backends/risingwave/__init__.py
```diff
@@ -268,3 +317,262 @@ def list_databases(
...
+    @experimental
+    def create_materialized_view(
+        self,
+        name: str,
+        obj: ir.Table,
+        *,
+        database: str | None = None,
+        overwrite: bool = False,
+    ) -> ir.Table:
+        """Creating a materialized view. The created materialized view can be accessed like a normal table.
```
gforsyth, line 330:
```suggestion
        """Create a materialized view. Materialized views can be accessed like a normal table.
```

small language nit

gforsyth (changes requested):
Thanks for pushing on this @KeXiangWang !

I flagged a few style things and one bit of incorrect logic, I think.

ibis/backends/risingwave/__init__.py
KeXiangWang: Yes, you are right. `pandas table` should not be added. It should be a typo. Updated

ibis/backends/risingwave/__init__.py
KeXiangWang: Updated. But kind of confused: what's the differences between `RelationError `, `UnsupportedArgumentError` and `ValueError`?

ibis/backends/risingwave/__init__.py
KeXiangWang: updated.

ibis/backends/risingwave/__init__.py
KeXiangWang: udpated

ibis/backends/risingwave/__init__.py
KeXiangWang: updated

ibis/backends/risingwave/__init__.py
gforsyth: It's confusing!

`com.RelationError` should maybe go away, but currently it's only for internal errors related to expression construction -- no user should ever see a `RelationError`.

`ValueError` is (in my opinion) generally too vague to be useful, but it's tricky to balance adding a bunch of custom exceptions.

`UnsupportedArgumentError` seemed like a good fit to me, but there isn't a strict guide on when to use which exceptions in Ibis.  I think it's something we can improve for new contributors.

KeXiangWang: OKK, thanks for the explanation!
As there's no comments left, can we merge it now?

gforsyth (approved):
Thanks for all of your work on this @KeXiangWang !

PR #8249, opened 2024-02-06
chore: update flags for ruff 0.2.0

NickCrews:
I was getting "warning: The `--show-source` argument is deprecated. Use `--output-format=full` instead." in
https://github.com/ibis-project/ibis/actions/runs/7805418174/job/21289589595?pr=7915

This is the new API in [ruff 0.2.0](https://github.com/astral-sh/ruff/releases/tag/v0.2.0), and we now have that version
of ruff pinned.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
         types_or:
           - python
           - pyi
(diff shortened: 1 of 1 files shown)


cpcloud (approved):
Ah, great, thank you!

PR #8269, opened 2024-02-07
docs(api): expose relevant mixin members in temporal docs

### LINKED ISSUE #8179: docs: document Date.epoch_seconds() and others from mixin classes
### Please describe the issue

any thing defined within
- _DateComponentMixin
- _TimeComponentMixin

don't appear in the docs for their subclasses

### Code of Conduct

gforsyth:
## Description of changes

Explicitly listing out the members of `_DateComponentMixin` and
`_TimeComponentMixin` where relevant.

A bit verbose, but I also don't imagine these changing all that often.

Or maybe we can add support to `quartodoc` at some point to exclude
inheriting from certain parents?  dunno.

## Issues closed

Fixes #8179

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -367,9 +367,60 @@ quartodoc:
             name: Temporal expressions
             desc: Dates, times, timestamps and intervals.
           contents:
-            - TimestampValue
-            - DateValue
-            - TimeValue
+            - name: TimestampValue
+              members:
+                - add
+                - radd
+                - sub
+                - between
+                - bucket
+                - date
+                - day
+                - day_of_week
+                - day_of_year
+                - delta
+                - epoch_seconds
+                - hour
+                - microsecond
+                - millisecond
+                - minute
+                - month
+                - quarter
+                - second
+                - strftime
+                - time
+                - truncate
+                - week_of_year
+                - year
+            - name: DateValue
+              members:
+                - add
+                - radd
+                - sub
+                - day
+                - day_of_week
+                - day_of_year
+                - epoch_seconds
+                - month
+                - quarter
+                - strftime
+                - truncate
+                - week_of_year
+                - year
+            - name: TimeValue
+              members:
+                - add
+                - radd
+                - sub
+                - between
+                - hour
+                - microsecond
+                - millisecond
+                - minute
+                - second
+                - strftime
+                - time
+                - truncate
             - IntervalValue
             - DayOfWeek
             - name: now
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -160,6 +160,11 @@ docs-render:
 docs-preview:
     quarto preview docs
 
+# regen api and preview docs
+docs-api-preview:
```
jcrist, line 164: Nice

jcrist (approved):

NickCrews:
I think it is inevitable that something is going to get added to one of these classes and then not added to this list in the docs. But that might just be the price we have to pay. I can't think of a good way to add any guardrails to prevent that, and I'd rather have something go undocumented then there being duplicated code.

docs/_quarto.yml
```diff
@@ -367,9 +367,63 @@ quartodoc:
...
+                - day_of_year
+                - epoch_seconds
+                - month
+                - quarter
+                - strftime
+                - truncate
+                - week_of_year
+                - year
+            - name: TimeValue
+              include_inherited: true
```
cpcloud, line 413: Do we need this? If we're explicitly listing the inherited members it doesn't seem like we should.

ncclementi: I just tested this locally, and you are right. If we list them we don't need the `include_inherited: true`
```suggestion
```

docs/_quarto.yml
```diff
@@ -367,9 +367,63 @@ quartodoc:
             name: Temporal expressions
             desc: Dates, times, timestamps and intervals.
           contents:
-            - TimestampValue
-            - DateValue
-            - TimeValue
+            - name: TimestampValue
+              include_inherited: true
```
ncclementi, line 371:
```suggestion
```
See Phillip comment below, I tested this locally and we don't need this if we list the members

gforsyth:
> I think it is inevitable that something is going to get added to one of these classes and then not added to this list in the docs. But that might just be the price we have to pay. I can't think of a good way to add any guardrails to prevent that, and I'd rather have something go undocumented then there being duplicated code.

It's imperfect, but yeah.  The other way around, we end up only adding a new property to one of the child classes and not the other.

cpcloud:
> Hmm, not going to do it here, but I think we probably don't need to run the backend CI for changes to the `justfile`

We do use `just` to run most of our test suite in CI, so I'm not sure we can ignore changes to it as is.

If we wanted to inline those commands that's certainly possible, though the downside is potential drift between the justfile and commands run in CI.

gforsyth:
> We do use `just` to run most of our test suite in CI, so I'm not sure we can ignore changes to it as is.

Ahh, right.  Probably not worth it.

PR #8274, opened 2024-02-08
chore: update readme to include exasol and flink

ncclementi:
We were missing some backends from the readme,  this PR adds Exasol and Flink. I know we recently add RisingWave too but I didn't find a page on the docs to link to so leaving it out.


README.md
```diff
@@ -55,12 +55,14 @@ Ibis acts as a universal frontend to the following systems:
 
 - [Apache Arrow DataFusion](https://ibis-project.org/backends/datafusion/) (experimental)
 - [Apache Druid](https://ibis-project.org/backends/druid/) (experimental)
+- [Apache Flink](https://ibis-project.org/backends/flink)
 - [Apache Impala](https://ibis-project.org/backends/impala/)
 - [Apache PySpark](https://ibis-project.org/backends/pyspark/)
```
deepyaman, line 60: Nit: I do see search results, but I don't think I've ever heard somebody say "Apache PySpark" instead of "Apache Spark" or simply "PySpark".

deepyaman (approved):
Thanks!

Comment is a slightly-unrelated observation.

jcrist (approved):

README.md
jcrist: Yeah, I'd be inclined to say "Apache Spark" instead. Likewise we have a flink backend, not a "pyflink" backend.

deepyaman: I did at one point suggest standardizing on PySpark/PyFlink or Spark/Flink, but there was a reasonable response from somebody that, nobody uses PyFlink (so it's not meaningful), while PySpark is a thing that most Python users would be familiar with.

This specific instance, though, I think Apache PySpark just doesn't feel natural. 🤷

ncclementi: Well, I have a similar though when I saw that on our docs, we don't use the prefix "Apache" in the Backends tab. We can go for standardizing a name convention when it comes to backends, and remove the "Apache" prefix from the readme in all the applicable cases.

PR #8280, opened 2024-02-08
ci(clouds): allow running cloud backends with github label

gforsyth:
I had this thought the other day -- instead of relying on someone with cloud credentials, sometimes we might want to trigger the cloud runs on a PR manually.  So this does that following the same pattern as our docs preview.  Need to add the label `ci-run-cloud` and it should kick off bigquery and snowflake.

If we're game to try this, we can also make it more specific, and have labels for only running bigquery or snowflake, but before I start writing a GHA matrix generator I thought I'd see how folks feel about this.

### DIFF
--- a/.github/workflows/ibis-backends-cloud.yml
+++ b/.github/workflows/ibis-backends-cloud.yml
@@ -11,6 +11,9 @@ on:
       - ".envrc"
     branches:
       - main
+  pull_request_target:
+    types:
+      - labeled
 
 permissions:
   # this allows extractions/setup-just to list releases for `just` at a higher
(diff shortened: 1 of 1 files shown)


cpcloud (approved):
Seems as good a place as any to start.

+1

Thanks for doing this.

gforsyth:
I'll add the label here and confirm that it kicks off the jobs

cpcloud:
I think `pull_request_target` runs against the base branch, so you may need to add another checkout step like we have in `docs-preview.yml`:

```yaml
      - name: checkout
        if: github.event.label.name == 'ci-run-cloud'
        uses: actions/checkout@v4
        with:
          fetch-depth: 0
          ref: ${{ github.event.pull_request.head.sha }}
```

and then leave the current one but add a

```yaml
if: github.event.label.name != 'ci-run-cloud'
```

or similar.

But maybe not, let's see what happens.

cpcloud:
Feel free to merge and iterate

PR #8282, opened 2024-02-08
test PR for CI labels

gforsyth:


gforsyth:
Sweet.  We check out the correct commit now with the cloud-run label, and the label gets autoremoved by the bot after the run is triggered. Huzzah!

PR #8284, opened 2024-02-08
fix(trino): re-enable native TABLESAMPLE support

jcrist:
This was implemented natively on `main`, but got lost in the refactor.


cpcloud (approved):
My bad, thank you!

PR #8295, opened 2024-02-09
test(sql): remove xdist_group marker for `.sql` tests

cpcloud:
`.sql` tests can now run in separate processes because the implementation is no longer stateful


cpcloud:
There's some other dead code related to stateful dot-sql lying around, might as well clean that up here.

kszucs:
We keep throwing out code, love it!

Issue #8256, opened 2024-02-07
bug: UDFs for builtins taking 0 args


NickCrews:
### Is your feature request related to a problem?

I want to access duckdb's [uuid()](https://duckdb.org/docs/sql/functions/utility.html#utility-functions) function, but that errors:

```python
from ibis.expr import datatypes as dt


@ibis.udf.scalar.builtin
def uuid() -> dt.uuid: ...


uuid()
```

It doesn't have to do with uuid stuff, I think it is because the builtin takes 0 args:

```python
@ibis.udf.scalar.builtin
def version() -> str: ...


version()
```
also errors

<details>

```python-traceback
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[15], line 9
      5 def uuid() -> dt.uuid: ...
      8 # vf.mutate(person__id=uuid())
----> 9 uuid()

File ~/code/scg/atlas/.venv/lib/python3.11/site-packages/ibis/common/deferred.py:607, in deferrable.<locals>.wrapper.<locals>.inner(*args, **kwargs)
    605     builder = Call(func, *args, **kwargs)
    606     return Deferred(builder, repr=repr)
--> 607 return func(*args, **kwargs)

File ~/code/scg/atlas/.venv/lib/python3.11/site-packages/ibis/expr/operations/udf.py:133, in _UDF._make_wrapper.<locals>.construct(*args, **kwargs)
    131 @functools.wraps(fn)
    132 def construct(*args: Any, **kwargs: Any) -> ir.Value:
--> 133     return node(*args, **kwargs).to_expr()

File ~/code/scg/atlas/.venv/lib/python3.11/site-packages/ibis/common/bases.py:73, in AbstractMeta.__call__(cls, *args, **kwargs)
     54 def __call__(cls, *args, **kwargs):
     55     """Create a new instance of the class.
     56 
     57     The subclass may override the `__create__` classmethod to change the
   (...)
     71     The newly created instance of the class. No extra initialization
     72     """
---> 73     return cls.__create__(*args, **kwargs)

File ~/code/scg/atlas/.venv/lib/python3.11/site-packages/ibis/common/grounds.py:119, in Annotable.__create__(cls, *args, **kwargs)
    115 @classmethod
    116 def __create__(cls, *args: Any, **kwargs: Any) -> Self:
    117     # construct the instance by passing only validated keyword arguments
    118     kwargs = cls.__signature__.validate(cls, args, kwargs)
[...]

PR #8296, opened 2024-02-09
fix(duckdb): udfs builtins taking zero args

ncclementi:
## Description of changes

This PR fixes the error thrown when trying to create a `udf.scalar.builtin` for an function that takes no arguments, for example

```python
@ibis.udf.scalar.builtin
def version() -> str: ...

version()
```

Thank you @gforsyth for the help troubleshooting this, as it wasn't that straightforward to figure out the core issue.

For a bit more context, the error was happening because when we had a udf that takes no arguments, then the shape check happening here: 

```python
@public
class ScalarUDF(ops.Value):
    shape = rlz.shape_like("args")
```
was failing, because there are no arguments to grab the shape of.

This fix, now takes cares of that by  defaulting to a scalar shape when this happens. 

## Issues closed

Closes https://github.com/ibis-project/ibis/issues/8256

### DIFF
--- a/ibis/backends/duckdb/tests/test_udf.py
+++ b/ibis/backends/duckdb/tests/test_udf.py
@@ -57,6 +57,19 @@ def test_builtin_scalar(con, func):
     assert con.execute(expr) == expected
 
 
+def test_builtin_scalar_noargs(con):
+    @udf.scalar.builtin
+    def version() -> str:
+        ...
+
+    expr = version()
+
+    with con.begin() as c:
+        expected = c.exec_driver_sql("SELECT version()").scalar()
+
+    assert con.execute(expr) == expected
+
+
 @udf.agg.builtin
 def product(x, where: bool = True) -> float:
     ...
--- a/ibis/expr/operations/udf.py
+++ b/ibis/expr/operations/udf.py
@@ -10,11 +10,12 @@ from typing import TYPE_CHECKING, Any, Callable, Optional, TypeVar, overload
 from public import public
 
 import ibis.common.exceptions as exc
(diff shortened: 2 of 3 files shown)


ibis/backends/duckdb/tests/test_udf.py
```diff
@@ -57,6 +57,19 @@ def test_builtin_scalar(con, func):
     assert con.execute(expr) == expected
 
 
+def test_builtin_scalar_noargs(con):
+    @udf.scalar.builtin
+    def version() -> str:
+        ...
+
+    expr = version()
```
jcrist, line 65: The error before was an expression level error. Would you mind adding a test like this in `ibis/tests/expr/test_udf.py` to test just the expression handling in a backend-agnostic way? Something like (untested):

```python
def test_builtin_scalar_noargs():
    @udf.scalar.builtin
    def version() -> str:
        ...

    expr = version()
    assert expr.type().is_integer()
    assert expr.op().shape is ibis.expr.datashape.Scalar
```

jcrist (commented):
Fix looks good to me! Just a request to add an expr-only test.

ibis/backends/duckdb/tests/test_udf.py
ncclementi: done, but had to modify the last line to be
```python
 assert type(expr.op().shape) is ibis.expr.datashape.Scalar
 ```
 to get the test to pass though

jcrist (approved):
:shipit:

NickCrews:
Thank you!

Issue #8302, opened 2024-02-10
bug: different behavior of StringConcat with NULLs depending on backend


NickCrews:
### What happened?

```python
import ibis

NULL = ibis.literal(None, type=str)
assert ibis.duckdb.connect().execute(NULL + NULL) == ""
assert ibis.duckdb.connect().execute(NULL + "abc") == "abc"

assert ibis.polars.connect().execute(NULL + NULL) == None
assert ibis.polars.connect().execute(NULL + "abc") == None

assert ibis.sqlite.connect().execute(NULL + NULL) == None
assert ibis.sqlite.connect().execute(NULL + "abc") == None

# TypeError: unsupported operand type(s) for +: 'NoneType' and 'NoneType'
# assert ibis.pandas.connect().execute(NULL + NULL) == None
# assert ibis.pandas.connect().execute(NULL + "abc") == None

# in DuckDB, NULL propagation depends on the exact operator used.
# postgres has the same semantics as DuckDB.
# sqlite only has the || operator, which always propagates NULL.
import duckdb

conn = duckdb.connect()
assert conn.execute("SELECT NULL::TEXT || NULL::TEXT;").fetchone()[0] == None
assert conn.execute("SELECT NULL::TEXT || 'abc'::TEXT;").fetchone()[0] == None
assert conn.execute("SELECT CONCAT(NULL::TEXT, NULL::TEXT);").fetchone()[0] == ""
assert conn.execute("SELECT CONCAT(NULL::TEXT, 'abc'::TEXT);").fetchone()[0] == "abc"


# pandas always propagates NULL
import pandas as pd

a = pd.Series([None, "abc"], dtype="string")
b = pd.Series([None, None], dtype="string")
assert (a + b).tolist() == [pd.NA, pd.NA]

# polars always propagates NULL
import polars as pl

a = pl.Series([None, "abc"])
b = pl.Series([None, None])
assert (a + b).to_list() == [None, None]
```
I haven't tested other backends so IDK what they do.

This is currently the cause of the failing duckdb test case in https://github.com/ibis-project/ibis/pull/8270, where currently `ibis.literal(None, "string").capitalize()` results in "", not NULL.

We need to
1. fix duckdb
2. add pandas implementation
3. add some test cases
4. add docstring examples to show it off.

### What version of ibis are you using?

8.0.0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
Fixed by #8305.

PR #8303, opened 2024-02-10
fix: handle NULLs consistently in String.concat()

NickCrews:
Partially fixes https://github.com/ibis-project/ibis/issues/8302.
Still need to make pandas actually handle NULLs at all.

### DIFF
--- a/ibis/backends/base/sql/alchemy/registry.py
+++ b/ibis/backends/base/sql/alchemy/registry.py
@@ -492,6 +492,17 @@ def _substring(t, op):
         )
 
 
+def _string_concat(t, op):
+    # We can't use the builtin sa.func.concat because we want NULL + <anything>
+    # to be NULL. But with sa.func.concat, NULL + <anything> is <anything>.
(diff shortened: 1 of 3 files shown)


NickCrews:
Actually, is this going to just get made outdated by `the-epic-split`? Maybe we should wait for that to land before fixing this up. It would also be great if a core dev could finish this up: since I don't have the creds for all the backends, I can only test some backends by submitting them to CI. I think it should be obvious what needs to happen.

cpcloud:
This seems to be a postgres-specific behavior that duckdb inherited.

ClickHouse, Trino and BigQuery all propagate nulls.

cpcloud:
Agree that we should revisit this after `the-epic-split` is merged which will very likely be this upcoming week.

cpcloud:
I put up #8305 instead, against `the-epic-split`

cpcloud:
I'll cherry pick your tests and examples over to that PR, thanks!

Issue #7995, opened 2024-01-16
docs - improve search functionality


p-a-a-a-trick:
### Please describe the issue

I used the search tool on the website (ibis-project.org) to look for syntax/documentation on basic functionality and none of the top results were what I expected.

Attempting to find information on the sum method (or function) on any expression:
![image](https://github.com/ibis-project/ibis/assets/100629128/8ddf9685-cc1d-4fd0-8647-0f4fded4d948)

Attempting to find information on the aggregate method on a table expression:
![image](https://github.com/ibis-project/ibis/assets/100629128/826dbdb2-58d6-46ee-b0e9-366005b5e48f)

It would be helpful for new users if common operations were ranked higher on basic search results (looking up "sum" would lead you to the various sum functions/methods before the word "summary" or blog posts).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

lostmygithubaccount:
potentially useful: https://quarto.org/docs/websites/website-search.html

also once Ibis Birdbrain is more properly done could be an interesting experiment to try search with it

gforsyth:
@ogrisel posted these in #8086 and I wanted to not lose track of them:

So to conclude, here are some concrete recommendations:
    exact matches with Ibis public function / class names should probably always be ranked at the top;
    the result snippet should highlight the fact that it's an API exact match one or another (at least showing the function / class name in at the beginning of the snippet);
    the result snippet should ideally quote the beginning of the docstring rather than a random excerpt near the end of the docstring.

PR #8325, opened 2024-02-12
docs: improve searchibility with algolia

ncclementi:
This is a work in progress, not working yet we have an issue with the records size. 

I was trying to run the `.github/workflows/upload-algolia.py` locally and run into this kind of error

```python-traceback
algoliasearch.exceptions.RequestException: Record at the position 459 objectID=reference/expression
tables.html#methods is too big size=224921/10000 bytes. Please have a look at https://www.algolia.com
doc/guides/sending-and-managing-data/prepare-your-data/in-depth/index-and-records-size-and
usage-limitations/#record-size-limits
```
We are generating massive objects that go into the `search.json`
For example the object generated by this section https://ibis-project.org/reference/expression-tables#methods  is ~225KB (this one is entry 459) you can take a look at it here: https://ibis-project.org/search.json

Not even the paid planned would not allow this.
**For Build plans:**
- 10KB for any individual record
**For Standard, Premium and Grow plans:**
- 100 KB for any individual record
- 10 KB average record size across all records

I think a solution could be instead of using the `search.json` from quarto, try to see what kind of index does the algolia crawler generates. According to [this docs](https://support.algolia.com/hc/en-us/articles/12255446296721-Is-there-a-free-crawler-option) it sounds like we could [get it for free if we have a Netlify account](https://www.algolia.com/doc/tools/crawler/netlify-plugin/quick-start/), which I believe we have?

Sweet, there is already a crawler GHA https://github.com/algolia/algoliasearch-crawler-github-actions. 

But I'm not sure if that's the right way to go about this. Opening this to discussion. 

Closes: #7995

cc: @gforsyth


ncclementi:
I decided to start fiddling with the `search.json` and it's very tricky, because it's not only the records that have "Examples" on them, the ones that will generate big entries, we have other cases too.

We have ~68 cases where the objects are big, and the examples strip only takes care of ~15.
I think if we come up with very specific rules to take care of this, it won't be sustainable. 

We (with @cpcloud and @gforsyth ) tried for a bit connecting the algolia crawler via the netlify interface and we couldn't get the crawler kicking. 

The last thing we could try is handling everything via GHA, and see what happens. Try to follow this, 
https://github.com/algolia/algoliasearch-crawler-github-actions/blob/main/examples/netlify.yml and see what happens

PR #8330, opened 2024-02-13
docs: update the geospatial blogpost to use explicit scalar subqueries

kszucs:
I thought that I fixed it a while ago, but the errors are not surfacing from the ipython cells, so we need to check the rendered output.


cpcloud:
This looks fixed. Can you re-enable docs publishing?

image

kszucs:
> This looks fixed. Can you re-enable docs publishing?

I am going to do that in a follow-up.

cpcloud:
Roger that, SGTM.

PR #8332, opened 2024-02-13
refactor(deps): remove `multipledispatch` as a dependency

cpcloud:
This PR removes `multipledispatch` as a dependency of Ibis. After `the-epic-split` was merged our largest consumer of `multipledispatch` was removed, now there is this remaining bit of code to port over to use non-multipledispatch-style checks.


kszucs:
The approach looks good to me, going to have a more thorough look.

ibis/expr/datatypes/cast.py
```diff
@@ -46,166 +43,3 @@ def highest_precedence(dtypes: Iterator[dt.DataType]) -> dt.DataType:
         return functools.reduce(higher_precedence, collected)
     else:
         return dt.null
-
-
-@castable.register(dt.DataType, dt.DataType)
```
kszucs, line 51: Shall we have function for backward compatibility?

cpcloud: I don't think so, this isn't a public API.

ibis/expr/datatypes/core.py
```diff
@@ -513,6 +524,11 @@ class Integer(Primitive, Numeric):
     def nbytes(self) -> int:
         """Return the number of bytes used to store values of this type."""
 
+    def castable(self, other, value: int | None = None, **kwargs) -> bool:
```
kszucs, line 527: For better clarity could you use either `target` or `to` argument name instead of other? Otherwise I need to think twice which one is the source and which one is the target.

cpcloud: Happy to do that in an immediate follow up

cpcloud: Actually, I'll just push up another commit

ibis/expr/datatypes/cast.py
kszucs: How about removing `dt.cast()` as well in favor of the `cast()` method?

ibis/expr/datatypes/core.py
cpcloud: Done in the most recent commit bb1a2c5.

kszucs (approved):

PR #8335, opened 2024-02-13
refactor(dtype): move all the castable logic to a single function

kszucs:
This makes it easier to understand the implicit casting rules.

### DIFF
--- a/ibis/expr/datatypes/cast.py
+++ b/ibis/expr/datatypes/cast.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import functools
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from public import public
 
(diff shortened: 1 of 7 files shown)


ibis/expr/datatypes/tests/test_cast.py
```diff
@@ -1,8 +1,50 @@
 from __future__ import annotations
 
+import hypothesis as h
 import pytest
 
 import ibis.expr.datatypes as dt
+import ibis.tests.strategies as ibst
+
+
+@h.given(ibst.signed_integer_dtypes(), ibst.signed_integer_dtypes())
```
kszucs, line 10: I am planning to add more hypothesis test cases here.

ibis/expr/datatypes/cast.py
```diff
@@ -24,16 +24,120 @@ def cast(source: str | dt.DataType, target: str | dt.DataType, **kwargs) -> dt.D
     return target
 
 
+@public
```
kszucs, line 27: Should remove all of the public decorators in this module.

cpcloud (approved):

ibis/expr/datatypes/tests/test_cast.py
kszucs: Going to address this in a follow-up.

ibis/expr/datatypes/cast.py
kszucs: Going to address this in a follow-up.

PR #8338, opened 2024-02-13
refactor(backends): remove singledispatchmethod from the sql backends

### LINKED ISSUE #8283: chore(docs): verify the accuracy of the support matrix in `the-epic-split`
The matrix isn't as accurate with `the-epic-split`, which we should address to the extent possible before releasing 9.0.

cpcloud:
## Description of changes

This pull request does two things:

1. Removes use of `singledispatchmethod` in the SQL compilers
1. Fixes the support matrix accuracy for SQL backends

Follow-ups:

* ~Deal with filtering out RisingWave geospatial~ Handled here
* `__init_subclass__` for `SIMPLE_OPS` handled here
* Fix coverage accuracy for the non-SQL backends

## Issues closed

Fixes #8283.

Thanks to @jcrist for the `__init_subclass__` tip, that saved N backends duplication of filling in undefined operations.

### DIFF
--- a/docs/support_matrix.qmd
+++ b/docs/support_matrix.qmd
@@ -7,8 +7,6 @@ hide:
 
 ```{python}
 #| echo: false
-from pathlib import Path
-
 import pandas as pd
 
 import ibis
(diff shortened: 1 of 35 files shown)


kszucs (approved):
Nice!

PR #8339, opened 2024-02-13
fix(pyarrow): map date type to arrow date32 not date64

jcrist:
Previously we were mapping our `Date` type to `pa.date64`. Here I propose we map to `pa.date32` instead. This is what `pa.Table.from_pandas` does for date columns. Also support for `date32` seems to be more common across backends with native arrow support.


kszucs (approved):
LGTM

PR #8270, opened 2024-02-07
fix(strings): make `StringValue.capitalize()` consistent across backends

NickCrews:
Fixes #8271.

BREAKING CHANGE: Backends that previouslly used `initcap` (analogous to `str.title`) to implement `StringValue.capitalize()` will produce different results when the input string contains multiple words (a word's definition being backend-specific).

PR #8340, opened 2024-02-13
refactor: use sge.DPipe for StringConcat

NickCrews:
The default implementation in the SQLGlot compiler should use DPipe.

Exasol still needs it's own visit_StringConcat()
because it doesn't handle NULLs properly

This is in preparation for https://github.com/ibis-project/ibis/pull/8270, which will rely upon this.


cpcloud:
This won't work, because of the NULL handling of these other backends.

When I wrote those implementations there were all necessary to get things passing.

I tried DPipe first, but that doesn't exist/work on all backends, as is seen in the test suite :)

NickCrews:
uuuughhhhhhh

cpcloud:
Well, I think you could've left the DPipe in the base class, and left the others, but closing it is also fine!

PR #8341, opened 2024-02-13
feat(api): add disconnect method

### LINKED ISSUE #5940: feat: implement disconnect method for backend connections

Is there a method that an Ibis user can call to disconnect from a backend (without quitting/restarting the Python session)?

Implement a disconnect method, or better document it if it already exists.

5.0.0

Trino

### Code of Conduct

gforsyth:
## Description of changes

This adds a `disconnect` method to all backends.  Previously we didn't do
this since the actual connection was often wrapped in SQLAlchemy and our
ability to terminate the connection was unclear.

The DB-API spec states that there should be a `close` method, and that
any subsequent operations on a given `Connection` should raise an error
after `close` is called.

This _mostly_ works.

Trino, Clickhouse, Impala, and BigQuery do not conform to the DB-API in
this way.  They have the `close` method but don't raise when you make a
subsequent call.

Spark is not a DB-API, but the `spark.sql.session.stop` method does the right thing.

For the in-process backends I've chosen to raise, since I don't think
there's a clear meaning on what closing that connection would mean, but
happy to take any suggestions there.

## Issues closed

Resolves #5940

### DIFF
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -1460,3 +1460,23 @@ def test_list_databases_schemas(con_create_database_schema):
             con_create_database_schema.drop_schema(schema, database=database)
     finally:
         con_create_database_schema.drop_database(database)
+
+
+@pytest.mark.notyet(
+    ["pandas", "dask", "polars", "datafusion"],
+    reason="this is a no-op for in-memory backends",
+)
+@pytest.mark.notyet(
+    ["trino", "clickhouse", "impala", "bigquery"],
+    reason="Backend client does not conform to DB-API, subsequent op does not raise",
+)
+def test_close_connection(con):
+    new_con = getattr(ibis, con.name).connect(*con._con_args, **con._con_kwargs)
+
+    # Run any command that hits the backend
+    _ = new_con.list_tables()
+    new_con.disconnect()
+
+    # DB-API states that subsequent execution attempt should raise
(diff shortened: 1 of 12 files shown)


jcrist:
> For the in-process backends I've chosen to raise

Presuming by "in-process" you mean pandas/polars here and not `duckdb`/`sqlite`, I think the proper behavior here is a no-op rather than a raise. This lets users write more backend-agnostic code. We _might_ also clear the stored tables mapping that the pandas/polars backends have to release memory, but :shrug: if that's worth it.

ibis/backends/tests/test_client.py
```diff
@@ -1460,3 +1469,37 @@ def test_list_databases_schemas(con_create_database_schema):
...
+            MySQLInterfaceError,
+            OracleInterfaceError,
+            PsycoPg2InterfaceError,
+            Py4JJavaError,
+            PyDruidError,
+            PyODBCProgrammingError,
+            SnowflakeDatabaseError,
+            sqlite3ProgrammingError,
+        )
+    ):
```
gforsyth, line 1504: I know `pytest.raises(Exception)` is considered evil, but writing all of these out is running afoul of our (good) test environment isolation.  Hmmmmm

jcrist: Using `pytest.raises(Exception)` seems fine to me honestly. We only care that things don't work, not necessarily why.

PR #8343, opened 2024-02-13
feat: support iceberg read/write

### LINKED ISSUE #7712: feat: Iceberg table support

Support Iceberg tables in Ibis

Using: https://github.com/apache/iceberg-python

Main blocker is write support, tracked here: https://github.com/apache/iceberg-python/issues/23

`ibis.read_iceberg`

`table.to_iceberg`

n/a

local backends that would support this

### Code of Conduct

mfatihaktas:
## Description of changes

Explorative PR towards addressing https://github.com/ibis-project/ibis/issues/7712.

This draft PR 
- Aims to bring more clarity to the concerns raised in https://github.com/ibis-project/ibis/issues/7712#issuecomment-1850331739.
- Implements `ibis.read_iceberg()` by
  - Fetching the files that constitute the Iceberg table one by one.
  - Registering the table into the default backend through an `RecordBatchReader`.
- Implements `table.to_iceberg()` by
  - Writing into the Iceberg table through `RecordBatch`'s.

### `pyiceberg` dependency
Notice the following change in `environment.yml`
```
- pip:
     - git+https://github.com/apache/iceberg-python.git@main
```
This is because write support in `pyiceberg` is merged into `main` but not released yet:
- https://github.com/apache/iceberg-python/pull/41
- https://github.com/apache/iceberg-python/issues/23

Looks like the next release is on the way:
- [[DISCUSS] PyIceberg 0.6.0 release](https://lists.apache.org/thread/6hsqmdlv6q3f56syopfjfoprf9por6rx)

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -406,6 +406,64 @@ services:
     networks:
       - flink
 
+  iceberg-rest:
+    image: tabulario/iceberg-rest
+    container_name: iceberg-rest
+    networks:
+      iceberg:
+    ports:
+      - 8181:8181
+    environment:
+      - AWS_ACCESS_KEY_ID=admin
+      - AWS_SECRET_ACCESS_KEY=password
+      - AWS_REGION=us-east-1
+      - CATALOG_WAREHOUSE=s3://warehouse/
+      - CATALOG_IO__IMPL=org.apache.iceberg.aws.s3.S3FileIO
+      - CATALOG_S3_ENDPOINT=http://iceberg-minio:9000
+    depends_on:
+      - iceberg-minio
+      - iceberg-mc
+
+  iceberg-minio:
+    image: minio/minio
+    container_name: iceberg-minio
+    environment:
+      - MINIO_ROOT_USER=admin
+      - MINIO_ROOT_PASSWORD=password
+      - MINIO_DOMAIN=iceberg-minio
+    networks:
+      iceberg:
+        aliases:
+          - warehouse.iceberg-minio
+          - warehouse2.iceberg-minio
+    ports:
+      - 9001:9001
+      - 9000:9000
+    command: ["server", "/data", "--console-address", ":9001"]
+
+  iceberg-mc:
+    image: minio/mc
+    container_name: iceberg-mc
+    networks:
+      iceberg:
+    environment:
+      - AWS_ACCESS_KEY_ID=admin
+      - AWS_SECRET_ACCESS_KEY=password
+      - AWS_REGION=us-east-1
+    entrypoint: >
+      /bin/sh -c "
+      until (/usr/bin/mc config host add iceberg-minio http://iceberg-minio:9000 admin password) do echo '...waiting...' && sleep 1; done;
+      /usr/bin/mc rm -r --force iceberg-minio/warehouse;
+      /usr/bin/mc rm -r --force iceberg-minio/warehouse2;
+      /usr/bin/mc mb iceberg-minio/warehouse;
+      /usr/bin/mc mb iceberg-minio/warehouse2;
+      /usr/bin/mc policy set public iceberg-minio/warehouse;
+      /usr/bin/mc policy set public iceberg-minio/warehouse2;
+      tail -f /dev/null
+      "
+    depends_on:
+      - iceberg-minio
+
   kudu:
     cap_add:
       - SYS_TIME
@@ -572,6 +630,7 @@ services:
       - risingwave
 
 networks:
+  iceberg:
   impala:
     # docker defaults to naming networks "$PROJECT_$NETWORK" but the Java Hive
     # Metastore clients don't accept underscores in the thrift URIs and
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -47,8 +47,9 @@ if TYPE_CHECKING:
     import numpy as np
     import pandas as pd
     import pyarrow as pa
+    import pyiceberg
 
-    from ibis.common.typing import SupportsSchema
+    from ibis.common.typing import Iterator, SupportsSchema
 
 __all__ = (
     "aggregate",
@@ -150,6 +151,7 @@ __all__ = (
     "range_window",
     "read_csv",
     "read_delta",
+    "read_iceberg",
     "read_json",
     "read_parquet",
     "row_number",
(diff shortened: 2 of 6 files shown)


compose.yaml
```diff
@@ -406,6 +406,64 @@ services:
...
+      - AWS_SECRET_ACCESS_KEY=password
+      - AWS_REGION=us-east-1
+      - CATALOG_WAREHOUSE=s3://warehouse/
+      - CATALOG_IO__IMPL=org.apache.iceberg.aws.s3.S3FileIO
+      - CATALOG_S3_ENDPOINT=http://iceberg-minio:9000
+    depends_on:
+      - iceberg-minio
+      - iceberg-mc
+
+  iceberg-minio:
```
cpcloud, line 427: Can you try to reuse the existing `minio` image?

ibis/expr/api.py
```diff
@@ -2400,3 +2402,183 @@ def least(*args: Any) -> ir.Value:
...
+# TODO: Move
+# - _arrow_record_batch_iterator_from_iceberg_table()
+# - _arrow_batch_reader_from_iceberg_table()
+# somewhere else.
+def _arrow_record_batch_iterator_from_iceberg_table(
+    iceberg_table: pyiceberg.table.Table,
+    snapshot_id: str = None,
+    options: dict[str, str] = {},
+    limit: int = None,
+) -> Iterator[pa.RecordBatch]:
```
cpcloud, line 2416: I see this code was adapted from the pyiceberg client. I'm guessing that's because of one my original objections which is that pyiceberg wants to own everything from IO to compute.

I looked at the original pyiceberg client code before raising my objections in the original issue, and my objections still remain.

I'll reiterate them here.

* This is a lot of processing code for Ibis. We really try to relegate low level details like IO (especially for complex file formats) to the backend as much as possible.
* The implementation here is using a private `pyiceberg` API. Using a third-party private API needs a very strong justification IMO.
* I don't quite understand the value add here.

Iceberg's stated reason for existing is to support huge analytic tables. When I think of huge, I think of tables that definitely do not fit into RAM on any non-specialized machine.

The implementation here is reading every file into RAM. I understand that it's possible after filters are run and projections/columns are pruned the result may fit into memory for a given file.

What if 5 out of 10 fit, but the rest don't? I suppose that's addressed by this function being a generator and forcing batch-at-a-time processing. This isn't a panacea either: any information that an engine could use to its advantage during compute is now gone because record batches don't (and likely couldn't in general) hold all the useful metadata for a compute engine.

There are two ways that would be a better approach to getting Iceberg support into Ibis:

For backends that don't have a native Iceberg reader:

* `pyiceberg` should implement a way to convert iceberg tables into pyarrow datasets. This is what deltalake has done in `delta-rs`, and that's how we support it Ibis backends that don't have native support for reading it.

This still has the metadata limitation, but at least it wouldn't be reading everything into memory.

For backends that natively support Iceberg:

* Enable that functionality under a backend specific `read_iceberg` where it makes sense. This would presumably be similar to the way we support deltalake via `read_delta`

cpcloud (changes requested):
My concerns from the corresponding issue still remain.

compose.yaml
mfatihaktas: Yes definitely. Did not spend time on this as this PR is exploratory. Will reuse the existing `minio` image once we have a PR that is mergeable.

ibis/expr/api.py
mfatihaktas: > I see this code was adapted from the pyiceberg client. I'm guessing that's because of one my original objections which is that pyiceberg wants to own everything from IO to compute.

Yes. This was an attempt to eliminate the in-memory operations performed by `pyiceberg`.

> This is a lot of processing code for Ibis. We really try to relegate low level details like IO (especially for complex file formats) to the backend as much as possible.

I agree that this (or whatever the final logic) needs to go into `pyiceberg`.

> The implementation here is using a private pyiceberg API. Using a third-party private API needs a very strong justification IMO.

Yes. `_arrow_record_batch_iterator_from_iceberg_table()` is only meant to be used to concretely understand what it takes to use `pyiceberg` (in its current state) to read Iceberg tables in Ibis. It is not intended to be final or merged.

> I don't quite understand the value add here.

Ok, but I have a question to help expand this a bit :) 
Would not the value be the same as `read_delta()` added for Ibis backends that do not have Delta Lake support?

> The implementation here is reading every file into RAM. I understand that it's possible after filters are run and projections/columns are pruned the result may fit into memory for a given file.
> What if 5 out of 10 fit, but the rest don't? I suppose that's addressed by this function being a generator and forcing batch-at-a-time processing. 

Yes.

> This isn't a panacea either: any information that an engine could use to its advantage during compute is now gone because record batches don't (and likely couldn't in general) hold all the useful metadata for a compute engine.

I agree. As far as I understand it, the data in Iceberg is supposed to be queried via a compute/query engine. Iceberg's Java API exposes the `DataFile` class to scan the files in an Iceberg table and retrieve their metadata, e.g., column level stats such as lower/upper bounds for columns, null counts etc.
[https://tabular.io/blog/java-api-part-2/]

> There are two ways that would be a better approach to getting Iceberg support into Ibis:
> For backends that don't have a native Iceberg reader:
> pyiceberg should implement a way to convert iceberg tables into pyarrow datasets. This is what deltalake has done in delta-rs, and that's how we support it Ibis backends that don't have native support for reading it.
> This still has the metadata limitation, but at least it wouldn't be reading everything into memory.

I agree. Will create an issue on https://github.com/apache/iceberg-python to request `iceberg_table.to_pyarrow_dataset()`.

> For backends that natively support Iceberg:
> Enable that functionality under a backend specific read_iceberg where it makes sense. This would presumably be similar to the way we support deltalake via read_delta

Yes, I took `read_delta()` as the main design reference. I presumed that the goal with `read_iceberg()` would be that of `read_delta()`: enabling "non-ideal" read/write to Iceberg tables for all backends, then overriding it for "ideal" implementation for backends that support read/write to Iceberg tables.

mfatihaktas: Update on `iceberg_table.to_pyarrow_dataset()`. It seems `pyiceberg` community has been discussing this for some time: https://github.com/apache/iceberg/issues/7598. They seem to be actively working on it: https://github.com/apache/iceberg-python/issues/30. Will keep an eye on this issue.

Fokko: Hey everyone, thanks for raising the issues here. First some clarification:

> I see this code was adapted from the pyiceberg client. I'm guessing that's because of one my original objections which is that pyiceberg wants to own everything from IO to compute.

This is not the case. PyIceberg is used in many projects as a query planner, and the query engine should do the heavy lifting. PyIceberg will do all the optimizations, and prune away all the Parquet files that are not relevant for the query. It will return a set of tasks ([.plan_files()](https://github.com/apache/iceberg-python/blob/44948cd9bab78b509b6befc960b5f248b37f53fe/pyiceberg/table/__init__.py#L1356)). This is a test of tasks that each point to a Parquet file that needs to be read, potentially delete on read files, and including some additional information that relevant for the query.

Why PyIceberg is doing everything by itself currently is mostly because Arrow does not support [field-id](https://github.com/apache/parquet-format/blob/97ed3ba484d3b5a7b58678457eceb518b037ee04/src/main/thrift/parquet.thrift#L459) based projection on which Iceberg depends very heavily on.

An example, when you rename a field, `name -> username` then the name changes, but the field-id stays the same. When you do this rename, it is a metadata operation. Iceberg allows Parquet files with an older schema (you don't want to rewrite a petabyte of data). But when we do the query:

- When filtering `username = 'Fokko'` we don't want to push this the the Parquet reader by name, but by field-id.
- When reading the data into a batch, we want to say `field-id=1` should be read with the name `username` instead of `name` which it was written with.

To make sure that this is handled correctly, we have quite a lot of Arrow code in the PyIceberg codebase, that we don't want (and can be solved by adding field-id support to Arrow). You could say; let's do the name-mapping on the PyIceberg side, but then we do have to fetch the footer twice (once on the PyIceberg side, and once on the Arrow side).

> This is a lot of processing code for Ibis. We really try to relegate low level details like IO (especially for complex file formats) to the backend as much as possible.

After reading the actual data, I fully agree with you. To give some context:

In Iceberg we have the concept of [a FileIO](https://tabular.io/blog/iceberg-fileio-cloud-native-tables/) which is an abstraction for object stores. The `PyArrowFileIO` is one of them, and the most popular one. PyIceberg needs to fetch the Iceberg Avro manifest files to do the query planning and compile the list of tasks for the query engine to execute. Ideally, things like column projection etc should be pushed to the engine as mentioned above.

As you can see the `io` is a property of the table. This is populated by the `Catalog`. For the Iceberg native REST catalog, it will provide a credential for the IO to have read (or write) access to the table.

> Iceberg's stated reason for existing is to support huge analytic tables. When I think of huge, I think of tables that definitely do not fit into RAM on any non-specialized machine.

In the case of PyIceberg/Arrow, it is easier to efficiently prune away parquet files, and minimize the IO by only reading relevant data by skipping Parquet files altogether.

> What if 5 out of 10 fit, but the rest don't? I suppose that's addressed by this function being a generator and forcing batch-at-a-time processing. This isn't a panacea either: any information that an engine could use to its advantage during compute is now gone because record batches don't (and likely couldn't in general) hold all the useful metadata for a compute engine.

In this PR it expects a Parquet file to fit into memory; that seems to be reasonable to me.

> Update on iceberg_table.to_pyarrow_dataset(). It seems pyiceberg community has been discussing this for some time: apache/iceberg#7598. They seem to be actively working on it: apache/iceberg-python#30. Will keep an eye on this issue.

I'm personally not very optimistic about happening any time soon. Don't get me wrong; I would love to see this, but things like field-id projection, filtering, etc should be in before we can delegate that work to Arrow.

ibis/expr/api.py
```diff
@@ -2400,3 +2402,183 @@ def least(*args: Any) -> ir.Value:
...
+        options=options,
+        limit=limit,
+    )
+
+    scheme, netloc, _ = PyArrowFileIO.parse_location(iceberg_table.location())
+    if isinstance(iceberg_table.io, PyArrowFileIO):
+        fs = iceberg_table.io.fs_by_scheme(scheme, netloc)
+    else:
+        try:
+            from pyiceberg.io.fsspec import FsspecFileIO
```
Fokko, line 2435: It will only fall back to `Fsspec` if `PyArrow` is not installed, but I don't think that's the case with Ibis.

ibis/expr/api.py
cpcloud: @Fokko Heh, thanks for chiming in. I was just reading through your [related `polars` PR](https://github.com/pola-rs/polars/pull/10375).

cpcloud: Appreciate the response, and agree on all points with perhaps a bit of disagreement that expecting a single Parquet file in-memory reasonble.

However, given the constraints, I think it's fine for now.

I think if we're going to use pyiceberg, then we should avoid duplicating this code and just accept the limitation that everything lives in-memory in a PyArrow table for now and use the corresponding pyiceberg API (https://py.iceberg.apache.org/api/#apache-arrow).

We may be able to experiment with pushing down predicates directly from Ibis expressions into pyiceberg reads as well.

As soon as we have a user of this feature, we can start planning that work.

Fokko: > I think if we're going to use pyiceberg, then we should avoid duplicating this code and just accept the limitation that everything lives in-memory in a PyArrow table for now and use the corresponding pyiceberg API (https://py.iceberg.apache.org/api/#apache-arrow).

That API will materialize all the files into memory, and return a `pa.Table`, that's different than this PR where we yield per file.

> We may be able to experiment with pushing down predicates directly from Ibis expressions into pyiceberg reads as well.

That's an important one that I missed. The `iceberg_table.scan(..)` should pass in the `row_filter`, otherwise you won't have partition pruning, or skipping files based on the column statistics. We do have a converter from Iceberg to PyArrow: https://github.com/apache/iceberg-python/blob/44948cd9bab78b509b6befc960b5f248b37f53fe/pyiceberg/io/pyarrow.py#L545

cpcloud: > That API will materialize all the files into memory, and return a pa.Table, that's different than this PR where we yield per file.

I understand. I am saying that the amount of code isn't worth the reduced overhead without more feedback from people using the feature.

cpcloud: Regarding expressions, I think we'd go directly from Ibis expressions to Iceberg expressions. Is there any reason for us to do something else?

Fokko: > I understand. I am saying that the amount of code isn't worth the reduced overhead without more feedback from people using the feature.

We could also move it to the PyIceberg side, if you think it is valuable to iterate over the batches

> Regarding expressions, I think we'd go directly from Ibis expressions to Iceberg expressions. Is there any reason for us to do something else?

No, I think that would be ideal 👍

cpcloud:
Closing out as stale.

PR #8349, opened 2024-02-14
refactor(sql): automatically add simple ops implementations

cpcloud:
Follow up to #8338 to clean up the SIMPLE_OPS boilerplate.

### DIFF
--- a/ibis/backends/base/sqlglot/compiler.py
+++ b/ibis/backends/base/sqlglot/compiler.py
@@ -155,6 +155,20 @@ TRUE = sge.true()
 STAR = sge.Star()
 
 
+def parenthesize_inputs(f):
+    """Decorate a translation rule to parenthesize inputs."""
+
+    def wrapper(self, op, *, left, right):
+        return f(
+            self,
+            op,
+            left=self._add_parens(op.left, left),
+            right=self._add_parens(op.right, right),
+        )
+
+    return wrapper
+
+
 @public
 class SQLGlotCompiler(abc.ABC):
     __slots__ = "agg", "f", "v"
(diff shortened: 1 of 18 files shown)


cpcloud:
Looks like I'm on the road to breaking ... everything!

ibis/backends/base/sqlglot/compiler.py
```diff
@@ -192,22 +192,181 @@ class SQLGlotCompiler(abc.ABC):
     UNSUPPORTED_OPERATIONS: frozenset[type[ops.Node]] = frozenset()
     """Set of operations the backend doesn't support."""
 
+    SIMPLE_OPS = {
```
kszucs, line 195: Can you spell out `ops` for consistency with `UNSUPPORTED_OPERATIONS`?

ibis/backends/base/sqlglot/compiler.py
```diff
@@ -192,22 +192,181 @@ class SQLGlotCompiler(abc.ABC):
...
+        ops.IntervalMultiply: sge.Mul,
+        ops.IntervalSubtract: sge.Sub,
+    }
+
     def __init__(self) -> None:
         self.agg = AggGen(aggfunc=self._aggregate)
         self.f = FuncGen()
         self.v = VarGen()
 
     def __init_subclass__(cls, **kwargs):
```
kszucs, line 314: So the only reason we need to construct and attach methods here is because of `has_operation` looks for the method? Couldn't we move this logic into a handler which would compile all value nodes if those are not explicitly overwritten?

ibis/backends/base/sqlglot/compiler.py
cpcloud: I'd rather do that in a follow up

ibis/backends/base/sqlglot/compiler.py
cpcloud: Maybe, but I don't want to spend time making this improvement perfect here. If you have an idea for a further improvement, that can be done in a follow up.

cpcloud:
Doesn't seem like my CI fixes for label running helped 😞

ibis/backends/base/sqlglot/compiler.py
```diff
@@ -192,22 +206,168 @@ class SQLGlotCompiler(abc.ABC):
...
-            setattr(cls, f"visit_{leaf.__name__}", cls.visit_Undefined)
+            setattr(cls, methodname(op), cls.visit_Undefined)
 
-        super().__init_subclass__(**kwargs)
+        # override existing base class implementations
+        for op, target_name in cls.SIMPLE_OPS.items():
+            setattr(cls, methodname(op), make_impl(op, target_name))
+
+        # add simple ops that are not already implemented
+        for op, target_name in SQLGlotCompiler.SIMPLE_OPS.items():
```
kszucs, line 361: Why is this required?

ibis/backends/base/sqlglot/compiler.py
```diff
@@ -1195,158 +1356,102 @@ def visit_JoinTable(self, op, *, parent, index):
...
+    def visit_GreaterEqual(self, op, *, left, right):
+        return sge.GTE(this=left, expression=right)
+
+    @parenthesize_inputs
+    def visit_Greater(self, op, *, left, right):
+        return sge.GT(this=left, expression=right)
 
-    setattr(SQLGlotCompiler, f"visit_{_op.__name__}", _fmt)
+    @parenthesize_inputs
+    def visit_LessEqual(self, op, *, left, right):
```
kszucs, line 1402: Why are these spelled out instead of being generated like `SIMPLE_OPS`?  `BINARY_INFIX_OPS` could be a mapping.

cpcloud: Because implementations call the base class impl. I don't want to keep refactoring this. Is this a blocker?

ibis/backends/base/sqlglot/compiler.py
cpcloud: So that all child classes have these operations.

kszucs: Ohh, I thought these are already available on the class.

ibis/backends/base/sqlglot/compiler.py
kszucs: Nope, it is not a blocker.

kszucs (approved):

PR #8350, opened 2024-02-14
feat(api): add `.as_scalar()` method for turning expressions into scalar subqueries

kszucs:
Since there is no safe way to determine whether a table or a column contains a single cell we need to provide a way for the users to inform ibis that it should treat an expression as a scalar subquery. Previously it was the `.to_array()` method, but it only worked for table expressions. 

Note that when a scalar reduction is passed to certain API methods ibis is able to infer that the value should be turned into a scalar subquery and it implicitly does it. Using this method can be omitted in those cases.

### DIFF
--- a/ibis/expr/types/core.py
+++ b/ibis/expr/types/core.py
@@ -602,7 +602,15 @@ class Expr(Immutable, Coercible):
 
     def as_table(self) -> ir.Table:
         """Convert an expression to a table."""
-        raise NotImplementedError(type(self))
+        raise NotImplementedError(
+            f"{type(self)} expressions cannot be converted into tables"
+        )
+
+    def as_scalar(self) -> ir.Scalar:
+        """Convert an expression to a scalar."""
+        raise NotImplementedError(
+            f"{type(self)} expression cannot be converted into scalars"
+        )
 
 
 def _binop(op_class: type[ops.Binary], left: ir.Value, right: ir.Value) -> ir.Value:
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -1149,37 +1149,6 @@ class Value(Expr):
         """Sort an expression descending."""
         return ops.SortKey(self, ascending=False).to_expr()
 
-    def as_table(self) -> ir.Table:
-        """Promote the expression to a [Table](./expression-tables.qmd#ibis.expr.types.Table).
-
-        Returns
-        -------
-        Table
-            A table expression
-
-        Examples
-        --------
-        >>> import ibis
-        >>> t = ibis.table(dict(a="str"), name="t")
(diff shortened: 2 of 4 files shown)


ibis/expr/types/core.py
```diff
@@ -602,7 +602,11 @@ def unbind(self) -> ir.Table:
 
     def as_table(self) -> ir.Table:
         """Convert an expression to a table."""
-        raise NotImplementedError(type(self))
+        raise NotImplementedError("This expression cannot be converted to a table")
```
cpcloud, line 605: Perhaps a more specific error message about what `This` is would be more useful.

Something like `"{type(self)} expressions cannot be converted into tables"`

ibis/expr/types/core.py
```diff
@@ -602,7 +602,11 @@ def unbind(self) -> ir.Table:
...
+
+    def as_scalar(self) -> ir.Scalar:
+        """Convert an expression to a scalar."""
+        raise NotImplementedError("This expression cannot be converted to a scalar")
```
cpcloud, line 609: Similarly with `as_table`, showing the specific thing that cannot be converted into a scalar will help both users and developers understand where in their expression things are failing.

ibis/expr/types/core.py
kszucs: Updated.

ibis/expr/types/core.py
kszucs: Updated.

ibis/expr/types/generic.py
```diff
@@ -1234,6 +1203,23 @@ def __pandas_result__(self, df: pd.DataFrame) -> Any:
 
         return PandasData.convert_scalar(df, self.type())
 
+    def as_scalar(self):
+        """Inform ibis that the expression should be treated as a scalar.
+
+        If the expression is a literal, it will be returned as is. If it depends
+        on a table, it will be turned to a scalar subquery.
```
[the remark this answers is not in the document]
kszucs: Added.

ibis/expr/types/generic.py
```diff
@@ -1327,6 +1313,54 @@ def __pandas_result__(self, df: pd.DataFrame) -> pd.Series:
...
 
+    def as_scalar(self) -> Scalar:
+        """Inform ibis that the expression should be treated as a scalar.
+
+        Creates a scalar subquery from the column expression. Since ibis cannot
+        be sure that the column expression contains only one value, the column
+        expression is wrapped in a scalar subquery and treated as a scalar.
+
+        Note that the execution of the scalar subquery will fail if the column
+        expression contains more than one value.
```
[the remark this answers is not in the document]
kszucs: Added.

cpcloud (approved):

PR #8354, opened 2024-02-14
feat(bigquery): add the 'offset' parameter to unnest

chelsea-lin:

### DIFF
--- a/ibis/expr/operations/arrays.py
+++ b/ibis/expr/operations/arrays.py
@@ -103,6 +103,7 @@ class ArrayFilter(Value):
 @public
 class Unnest(Value):
     arg: Value[dt.Array]
+    offset: Optional[bool] = False
 
     shape = ds.columnar
 
--- a/ibis/expr/operations/arrays.py
+++ b/ibis/expr/operations/arrays.py
@@ -109,7 +109,10 @@ class Unnest(Value):
 
     @attribute
     def dtype(self):
-        return self.arg.dtype.value_type
+        if self.offset:
+            return dt.Struct(dict(offset=dt.uint64, value=self.arg.dtype.value_type))
+        else:
+            return self.arg.dtype.value_type
 
 
 @public
--- a/ibis/expr/operations/arrays.py
+++ b/ibis/expr/operations/arrays.py
@@ -103,7 +103,7 @@ class ArrayFilter(Value):
 @public
 class Unnest(Value):
     arg: Value[dt.Array]
-    offset: Optional[bool] = False
+    offset: bool
 
     shape = ds.columnar
 
--- a/ibis/expr/types/arrays.py
+++ b/ibis/expr/types/arrays.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import inspect
-from typing import TYPE_CHECKING, Callable, Optional
+from typing import TYPE_CHECKING, Callable
 
 from public import public
 
--- a/ibis/expr/types/arrays.py
+++ b/ibis/expr/types/arrays.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import inspect
-from typing import TYPE_CHECKING, Callable
+from typing import TYPE_CHECKING, Callable, Optional
 
 from public import public
 
@@ -266,8 +266,15 @@ class ArrayValue(Value):
 
     __mul__ = __rmul__ = repeat
 
-    def unnest(self) -> ir.Value:
+    def unnest(self,
+               offset: Optional[bool] = False
+        ) -> ir.Value:
         """Flatten an array into a column.
+
+        Parameters
+        ----------
+        offset
+            Whether to include the position index in the unnested array.
 
         ::: {.callout-note}
         ## Rows with empty arrays are dropped in the output.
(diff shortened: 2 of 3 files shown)


ibis/expr/operations/arrays.py
```diff
@@ -103,6 +103,7 @@ class ArrayFilter(Value):
 @public
 class Unnest(Value):
     arg: Value[dt.Array]
+    offset: Optional[bool] = False
```
cpcloud, line 106:
```suggestion
    offset: bool
```

`Optional[T]` is to indicate that a value of type `T` can be `None` or a instance of `T`, not whether the argument itself must be be specified.

It also doesn't need to be optional in the `Unnest` operation. That optionality should happen at the user API level.

ibis/expr/types/arrays.py
```diff
@@ -266,9 +266,16 @@ def repeat(self, n: int | ir.IntegerValue) -> ArrayValue:
 
     __mul__ = __rmul__ = repeat
 
-    def unnest(self) -> ir.Value:
+    def unnest(self,
+               offset: Optional[bool] = False
+        ) -> ir.Value:
```
cpcloud, line 271: How does a user access the offset value? Wouldn't the result be something more complex than an array?

Perhaps this operation should be a new internal Ibis operation, like `UnnestWithOffset`.

Alternatively, the `dtype` property of the existing `Unnest` operation needs to be adjusted to return something like `array<struct<offset: int64, value: T>>`?

In any event, there's some API design work here.

ibis/expr/types/arrays.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import inspect
-from typing import TYPE_CHECKING, Callable
+from typing import TYPE_CHECKING, Callable, Optional
```
cpcloud, line 4:
```suggestion
from typing import TYPE_CHECKING, Callable
```

ibis/expr/types/arrays.py
```diff
@@ -266,9 +266,16 @@ def repeat(self, n: int | ir.IntegerValue) -> ArrayValue:
 
     __mul__ = __rmul__ = repeat
 
-    def unnest(self) -> ir.Value:
+    def unnest(self,
+               offset: Optional[bool] = False
```
cpcloud, line 270:
```suggestion
               offset: bool = False
```

See my comment about `Optional` above.

cpcloud (changes requested):

ibis/expr/types/arrays.py
chelsea-lin: Thank you! I just noticed your comments here when I was creating #8356. Will follow up with your suggestion here!

chelsea-lin: If the `dtype` property returns `struct` type as you suggested, it requires the transformed SQL selects a `struct` column too. But [explode_to_unnest](https://github.com/tobymao/sqlglot/blob/6cb985ae1346c1a912ed6f81be30310ee1c91dfa/sqlglot/transforms.py#L168) in `sqlglot` returns two columns instead. One tricky solution is creating a BQ custom `TRANSFORM` to modify the sqlglot expression tree. Do you have another suggestions?

```
In [12]: ibis.to_sql(table.select(["scalar_column", table.z.unnest(offset=True)]).z)
Out[12]: 
SELECT
  `t1`.`z`
FROM (
  SELECT
    `t0`.`scalar_column`,
    IF(pos = pos_2, `z`, NULL) AS `z`,
    IF(pos = pos_2, pos_2, NULL) AS pos_2
  FROM `bigframes-dev`.`ibis_gbq_testing`.`array_types` AS `t0`
  CROSS JOIN UNNEST(GENERATE_ARRAY(0, GREATEST(ARRAY_LENGTH(`t0`.`z`)) - 1)) AS pos
  CROSS JOIN UNNEST(`t0`.`z`) AS `z` WITH OFFSET AS pos_2
  WHERE
    pos = pos_2
    OR (
      pos > (
        ARRAY_LENGTH(`t0`.`z`) - 1
      ) AND pos_2 = (
        ARRAY_LENGTH(`t0`.`z`) - 1
      )
    )
) AS `t1`

In [13]: ibis.to_sql(table.select(["scalar_column", table.z.unnest(offset=True)]).pos_2)
AttributeError                            Traceback (most recent call last)
Cell In[13], line 1
----> 1 ibis.to_sql(table.select(["scalar_column", table.z.unnest(offset=True)]).pos_2)

File ~/src/forkibis2/ibis/expr/types/relations.py:859, in Table.__getattr__(self, key)
    854     hint = common_typos[key]
    855     raise AttributeError(
    856         f"{type(self).__name__} object has no attribute {key!r}, did you mean {hint!r}"
    857     )
--> 859 raise AttributeError(f"'Table' object has no attribute {key!r}")

AttributeError: 'Table' object has no attribute 'pos_2'
```

chelsea-lin: Updated the change to build an sge expression to combine the offset and value as a STRUCT node. But it is blocked by https://github.com/tobymao/sqlglot/issues/2978
Any thoughts are welcome!

chelsea-lin: @cpcloud The `sqlglot` is not able to return a struct column combining `col` and `pos`, because `STRUCT(POSEXPODE(arr))` is not a `sparksql` grammar. Hence, I am thinking the ibis expression tree can construct two returned columns instead? If so, could you please give me some ideas how to achieve that?

cpcloud:
Closing this for now. I'm working on this locally and should hopefully have a PR up later this week or early next week.

PR #8358, opened 2024-02-14
docs: improve nested type docstrings

NickCrews:
The existing descriptions for these types at https://ibis-project.org/reference/expression-collections are pretty bad.

- Improve toplevel docstring for types so you can get a sense for what they type even is.
- ~~Add "corresponds to" list so people
can link this concept to concepts they are familiar to.~~
- fix explanations of how the datashape is determined in .map(), .array(), and .struct().
- some examples weren't really needed
- used "column" not "array" when describing datashape, since "array" has another meaning in ibis

### DIFF
--- a/ibis/expr/types/arrays.py
+++ b/ibis/expr/types/arrays.py
@@ -20,6 +20,25 @@ import ibis.common.exceptions as com
 
 @public
 class ArrayValue(Value):
+    """An Array is a nested type, a variable-length sequence of values of a single type.
+
+    Examples
+    --------
+    >>> import ibis
+    >>> ibis.options.interactive = True
+    >>> ibis.memtable({"a": [[1, None, 3], [4], [], None]})
+    ┏━━━━━━━━━━━━━━━━━━━━━━┓
+    ┃ a                    ┃
+    ┡━━━━━━━━━━━━━━━━━━━━━━┩
+    │ array<int64>         │
+    ├──────────────────────┤
+    │ [1, None, ... +1]    │
+    │ [4]                  │
+    │ []                   │
+    │ NULL                 │
+    └──────────────────────┘
+    """
+
     def length(self) -> ir.IntegerValue:
         """Compute the length of an array.
 
@@ -1065,10 +1084,13 @@ class ArrayColumn(Column, ArrayValue):
 def array(values: Iterable[V]) -> ArrayValue:
     """Create an array expression.
 
+    If any values are Columns, the result will be an ArrayColumn.
+    Otherwise, the result will be an ArrayScalar.
+
     Parameters
     ----------
     values
-        An iterable of Ibis expressions or a list of Python literals
+        An iterable of Ibis expressions or Python literals
 
     Returns
     -------
@@ -1076,7 +1098,7 @@ def array(values: Iterable[V]) -> ArrayValue:
 
     Examples
     --------
-    Create an array from scalar values
+    Create an array scalar from scalar values
 
     >>> import ibis
     >>> ibis.options.interactive = True
--- a/ibis/expr/types/arrays.py
+++ b/ibis/expr/types/arrays.py
@@ -20,7 +20,7 @@ import ibis.common.exceptions as com
 
 @public
 class ArrayValue(Value):
-    """An Array is a nested type, a variable-length sequence of values of a single type.
+    """An Array is a variable-length sequence of values of a single type.
 
     Examples
     --------
--- a/ibis/expr/types/arrays.py
+++ b/ibis/expr/types/arrays.py
@@ -1084,8 +1084,9 @@ class ArrayColumn(Column, ArrayValue):
 def array(values: Iterable[V]) -> ArrayValue:
     """Create an array expression.
 
-    If any values are Columns, the result will be an ArrayColumn.
-    Otherwise, the result will be an ArrayScalar.
+    If any values are [column expressions](../concepts/datatypes.qmd) the
+    result will be a column. Otherwise the result will be a
+    [scalar](../concepts/datatypes.qmd).
 
     Parameters
     ----------
--- a/ibis/expr/types/maps.py
+++ b/ibis/expr/types/maps.py
@@ -17,7 +17,14 @@ if TYPE_CHECKING:
 
 @public
 class MapValue(Value):
-    """A map literal or column expression.
+    """A dict-like collection with fixed-type keys and values.
+
+    This type is similar to a Python dictionary, except with the
+    restriction that all keys must have the same type, and all values
+    must have the same type.
+    For example, all keys are `string`s, and all values are `int64`s.
+    Each key must be unique within the map.
+    There can be 0, 1, or many key-value pairs in a map.
 
     Can be constructed with [`ibis.map()`](#ibis.expr.types.map).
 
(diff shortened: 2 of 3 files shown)


ibis/expr/types/structs.py
```diff
@@ -26,55 +26,36 @@ def struct(
...
-    │ {'a': 3, 'b': 'baz'}        │
-    └─────────────────────────────┘
+    >>> ibis.struct(dict(a=1, b="foo"))
+
+    Specify a specific type (note the 1 is now a float):
+    >>> ibis.struct(dict(a=1, b="foo"), type="struct<a: float, b: string>")
 
-    Create a struct array from columns and literals
+    Create a struct column from a column and a scalar literal
+    >>> t = ibis.memtable({"a": [1, 2, 3]})
```
cpcloud, line 58:
```suggestion
    Create a struct column from a column and a scalar literal
    
    >>> t = ibis.memtable({"a": [1, 2, 3]})
```

Line breaks are required between prose and code.

ibis/expr/types/structs.py
```diff
@@ -26,55 +26,36 @@ def struct(
...
-    │ struct<a: int64, b: string> │
-    ├─────────────────────────────┤
-    │ {'a': 1, 'b': 'foo'}        │
-    │ {'a': 2, 'b': 'bar'}        │
-    │ {'a': 3, 'b': 'baz'}        │
-    └─────────────────────────────┘
+    >>> ibis.struct(dict(a=1, b="foo"))
+
+    Specify a specific type (note the 1 is now a float):
+    >>> ibis.struct(dict(a=1, b="foo"), type="struct<a: float, b: string>")
```
cpcloud, line 55:
```suggestion
    Specify a specific type (note the 1 is now a float)

    >>> ibis.struct(dict(a=1, b="foo"), type="struct<a: float, b: string>")
```

ibis/expr/types/structs.py
```diff
@@ -99,12 +80,23 @@ def struct(
...
+    """A Struct is a nested type with ordered fields of fixed types.
+
+    Corresponds to
+    - Struct in DuckDB
+    - Struct in BigQuery
+    - Struct in Spark
+    - Struct in Impala
+    - Struct in Apache Arrow
+    - CompositeType in PostgreSQL
+    - etc
```
cpcloud, line 92: I'm not sure what the value of having these lists of backend types in Ibis's expression type docstrings is. I don't think the specific backend type is relevant for most users.

ibis/expr/types/maps.py
```diff
@@ -433,11 +446,10 @@ def map(
     keys: Iterable[Any] | Mapping[Any, Any] | ArrayColumn,
     values: Iterable[Any] | ArrayColumn | None = None,
 ) -> MapValue:
-    """Create a [map container object](https://docs.python.org/3/glossary.html#term-mapping).
+    """Create a MapValue.
```
cpcloud, line 449: Specifically not mentioning the Ibis type is intentional here. The specific expression type doesn't particularly matter for most usage.

Can we leave this and the other mentions of specific types out of the top-level docstrings?

cpcloud (changes requested):

ibis/expr/types/arrays.py
```diff
@@ -1058,18 +1084,21 @@ def __getitem__(self, index: int | ir.IntegerValue | slice) -> ir.Column:
 def array(values: Iterable[V]) -> ArrayValue:
     """Create an array expression.
 
+    If any values are Columns, the result will be an ArrayColumn.
+    Otherwise, the result will be an ArrayScalar.
```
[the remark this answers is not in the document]
NickCrews: I think that the 3 different datashapes of Scalar, Column, and Table are essential for a user to understand. A FloatingColumn has a .mean() function, but a FloatingScalar doesn't. When I .execute() something, am I going to get a python `int`, or pandas series of ints? I need to know which shape I am working with. I think I would have been a lot less confused with this when I first picked up ibis if this had been more forward.

The datashapes uses the same concept of broadcasting from SQL and from numpy, so for experienced users it shouldn't be new. For newbies, it might be confusing at first, but still I think it is essential to understand, and more of a disservice to them to hide it.

IDK this might just be an opinion deal though, so maybe we should get more user input here?

ibis/expr/types/maps.py
NickCrews: Similar to above, I was getting annoyed at these docstrings because they were hiding what they were returning. I need to know where to go look in the reference to see what operations I can actually do on the thing that is returned.

ibis/expr/types/structs.py
NickCrews: I was thinking that the verbose, precise definition of "a nested type with ordered fields of fixed types." is a little hard to grok. I wanted analogies, assuming that a user would be familiar with at least one of these things. It's not needed after the initial grokking, but helps with that. Open to removing these bits though if you still want to.

ibis/expr/types/maps.py
NickCrews: What was the reason for explicitly not mentioning the Ibis type? Trying to not expose an implementation detail?

ibis/expr/types/structs.py
cpcloud: I think this is probably too much information to throw at a user in a docstring.

I suspect something like a high level description of what a struct is suffices. Thoughts?

ibis/expr/types/maps.py
cpcloud: I think I was concerned about not having any explanatory references to what a `Value` is. If we had that, and we could link to it, then that would be ideal IMO.

ibis/expr/types/structs.py
NickCrews: Yeah I am +0 on this so lets do it.

ibis/expr/types/maps.py
NickCrews: My first thought is that Value could actually get decently ignored. I think if in the "datashapes" explanation doc we explained the concrete types of Scalar and Column, then we could just throw a sentence into that glossing over Values:

```
A bit of data in ibis is defined by two concepts at right angles to each other:

- A datatype (string, integer, floating, struct, etc)
- A datashape (scalar or columnar)

See this table:

|              | Scalar        | Column        |
|--------------|---------------|---------------|
| StringValue  | StringScalar  | StringColumn  |
| IntegerValue | IntegerScalar | IntegerColumn |
| StructValue  | StructScalar  | StructColumn  |

For example, and IntegerColumn object inherits from both the Column abstract base class, and the IntegerValue abstract base class.

These two characteristics determine what a bit of data can do:
```

What other concepts of Value are important that you would want to cover?

ibis/expr/types/arrays.py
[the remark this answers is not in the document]
NickCrews: I would love it if you wanted to take the lead on that.

ibis/expr/types/structs.py
NickCrews: I removed these "similar to" examples from the docstrings of array, struct, and map

ibis/expr/types/maps.py
NickCrews: See https://github.com/ibis-project/ibis/pull/8557

ibis/expr/types/arrays.py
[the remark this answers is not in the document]
NickCrews: See https://github.com/ibis-project/ibis/pull/8557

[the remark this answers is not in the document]
NickCrews: Should I add a link to that documentation page? If so, I'm not sure what the syntax is...

NickCrews:
@cpcloud now that the datashapes and datatypes guide is in, can you take a look at this again? Maybe @gforsyth also has thoughts here (this is your reward when you are helpful in other reviews 😉 ). Maybe this isn't perfect still, but I think it is a definite improvement.

ibis/expr/types/arrays.py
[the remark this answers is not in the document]
gforsyth: Should be the same as a regular markdown link, with an absolute path starting after `docs`

cpcloud:
Taking another look now. Thanks for bearing with me in the slightly longer review cycle 😅

ibis/expr/types/arrays.py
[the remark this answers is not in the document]
cpcloud: Done.

PR #8359, opened 2024-02-14
fix(flink): fix compilation of over aggregation query in flink backend

chloeh13q:
## Description of changes

This PR fixes the compilation of over aggregation query in Flink backend.

I identified the following issues:

1. The interval time value is not compiled with the appropriate precision. `ibis.interval(minutes=500)` compiles to `INTERVAL '500' MINUTE, which throws
```
org.apache.calcite.sql.validate.SqlValidatorException: Interval field value 500 exceeds precision of MINUTE(2) field
```
2. `range=(-ibis.interval(minutes=500), 0)` compiles to `[...] AND CAST(0 AS INTERVAL MINUTE)` following and Flink doesn't support rows/range following in over aggregation. This throws
```
org.apache.flink.table.api.TableException: OVER RANGE FOLLOWING windows are not supported yet.
```

I also removed the unused test snapshot directory that didn't get cleaned up.


ibis/backends/base/sqlglot/dialects.py
```diff
@@ -68,6 +69,45 @@ class Generator(Postgres.Generator):
         }
 
 
+def _calculate_precision(interval_value: int) -> int:
+    """Calculate interval precision.
+
+    FlinkSQL interval data types use leading precision and fractional-
```
zhenzhongxu, line 75: Is this file for the base sqlglot dialect? It looks like we are adding flink-specific logic into the base implementation? It may be good to get @cpcloud or @jcrist 's feedback on a better design pattern to use here.

zhenzhongxu: Cool, thanks for the note. Sorry got busy and couldn't review as often as I would like, I will have to rely more on the team :)

chloeh13q: No worries! ;)

ibis/backends/sql/dialects.py
```diff
@@ -68,6 +69,45 @@ class Generator(Postgres.Generator):
...
+    FlinkSQL interval data types use leading precision and fractional-
+    seconds precision. Because the leading precision defaults to 2, we need to
+    specify a different precision when the value exceeds 2 digits.
+
+    (see
+    https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/interval-literals)
+    """
+    # log10(interval_value) + 1 is equivalent to len(str(interval_value)), but is significantly
+    # faster and more memory-efficient
+    if interval_value == 0:
```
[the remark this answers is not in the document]
chloeh13q: `interval_value == 0` covers the case of `ibis.interval(0)`.
`interval_value < 0` should technically never get hit in this context because a negative interval is not allowed, but this is just to be proactive about error-catching and in case this function gets used elsewhere.

cpcloud (approved):

ibis/backends/sql/dialects.py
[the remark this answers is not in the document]
cpcloud: Not a blocker.

PR #8367, opened 2024-02-15
feat(duckdb): allow all-null columns in memtables

cpcloud:
Turns out we are artificially restricting the input of all-null columns in memtables. This works out of the box by simply deleting the check that raises an exception.


kszucs (approved):
Nice!

PR #8371, opened 2024-02-15
chore: add env for arm64 and xfail geospatial tests

ncclementi:
Add environment-arm64.yml to build dev env on mac m1 and m2. 

While doing this, a shapely/geos bug fix was exposed (see issue https://github.com/shapely/shapely/issues/1992). Including the corresponding fixes, when shapely is upgraded, the xfails should start xpass.

This PR: needs a follow up for documentation. To point out:
- that `apache-flink` install will downgrade some packages. 
- We are not including `go-shfmt` the pacakge is not available for amr64 and it's not on pip. I don't see this package on our current `requirement-dev.txt` so I'll ignore this. 
- `quarto >=1.4` is not available in `conda-forge` it'll have to be installed following https://quarto.org/docs/get-started/index.html


environment-arm64.yml
cpcloud, line 1: Can we put this and the other environment file into a `conda` directory?

ncclementi: sure thing, fixing now.

ncclementi:
I should take care of this one https://github.com/ibis-project/ibis/issues/8375 here too. Not sure what's the best way. Maybe pinning dask?

ncclementi:
With this environment, I tried 
- `just flink up` 
- `pytest ibis/backends/flink/tests`

and pretty much everything fails with 

```
E               RuntimeError: Java gateway process exited before sending its port number.
E               Stderr:
E               The operation couldn’t be completed. Unable to locate a Java Runtime.
E               Please visit http://www.java.com for information on installing Java.
```
for the record I do have installed `apache-flink`

Screenshot 2024-02-15 at 8 24 10 PM

cc: @chloeh13q @deepyaman any idea what's happening here?

chloeh13q:
@ncclementi I use sdkman to manage java because I have multiple versions of java running on my computer. Installing java 11 using sdkman is pretty straightforward, you just need to run
```sh
sdk install java 11.0.19-tem
```
and there is no extra step involved.

I don't remember having to symlink when I installed java11 via `brew install`; maybe try running the tests first to see whether it works?

ibis/backends/postgres/tests/test_geospatial.py
```diff
@@ -370,7 +376,7 @@ def test_geo_dataframe(geotable):
 @pytest.mark.parametrize(
     ("shape", "value"),
     [
-        # Geometry primitives (2D)
+        # # Geometry primitives (2D)
```
cpcloud, line 379:
```suggestion
        # Geometry primitives (2D)
```

cpcloud (approved):
Small nit, otherwise LGTM.

Thanks for powering through this 😬

cpcloud:
FWIW `go-shfmt` isn't a Python package so it won't show up on PyPI or in `requirements-dev.txt`.

PR #8373, opened 2024-02-16
test(duckdb): get things passing with duckdb 0.10.0

cpcloud:
Get tests passing with duckdb 0.10.0


cpcloud:
I'm going to unwind these cotangent changes and address the inconsistency in a follow-up PR.

cpcloud:
Well, not a huge fan of merging a dependency that results in a segfault.

I don't have a windows machine lying around to test this either.

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -110,7 +111,9 @@ def test_read_geo_from_url(con, monkeypatch):
     loaded_exts = []
     monkeypatch.setattr(con, "_load_extensions", lambda x, **_: loaded_exts.extend(x))
 
-    with pytest.raises((duckdb.IOException, duckdb.CatalogException)):
+    with pytest.raises(
+        (duckdb.IOException, duckdb.CatalogException, duckdb.NotImplementedException)
```
ncclementi, line 115: @cpcloud is this the reason you added `duckdb.NotImplementedException`? 

```python-traceback
E       duckdb.duckdb.NotImplementedException: Not implemented Error: HTTPFileSystem: DirectoryExists is not implemented!

../__init__.py:104: NotImplementedException
```

I ran into this while still in 0.9.2 and notice that it gets fixed if you force install `httpfs`. Did you saw these being raises in CI, or locally? I'd suspect we don't need that on CI, as I would imagine that extensions are being installed fresh in every ci run.

cpcloud: I saw these locally. We should really avoid as much divergence between CI and local as possible, so I'll probably install the httpfs extension and then report this as a bug upstream.

cpcloud:
Reported the failure here: https://github.com/duckdb/duckdb/issues/10787

cpcloud:
I'm going to rehash and force push a couple of times to see if the problem can be reproduced when running tests serially.

cpcloud:
Looks like something is causing a core dump on Linux now: https://github.com/ibis-project/ibis/actions/runs/7994707013/job/21833372662?pr=8373

😞

ncclementi:
> Looks like something is causing a core dump on Linux now: https://github.com/ibis-project/ibis/actions/runs/7994707013/job/21833372662?pr=8373 

Given that this stopped in geospatial, my guess is that this is the same than https://github.com/duckdb/duckdb_spatial/issues/236

If so, good news is Max and Tishj identified the problem comes from the spatial extension, and they can reproduce it. Bad news is this might not be fixed until the next release of the spatial extension (idk when would that be).

ncclementi:
So this is fixed https://github.com/duckdb/duckdb_spatial/issues/236  for v0.10.0 and the current nightly version of the spatial extension. Once that spatial extension gets released, we can recap here.

cpcloud:
Running tests serially seems to sidestep the windows segfault for now, and the geospatial issue doesn't seem to reproduce anymore.

I'll trigger a few more CI runs to see if they're still an issue and then we can merge if those don't yield anything useful.

cpcloud:
@ncclementi @gforsyth @jcrist Anyone opposed to merging this?

The windows segfaults don't seem to reproduce under serial testing (not great, but reported upstream) and the linux crashes also do not reproduce. I don't see a release of the geospatial extension, but I also have no idea whether we're picking up a fix or not in CI when we run `install geospatial;`

gforsyth (approved):

PR #8386, opened 2024-02-17
chore(dev-tools): remove unused dependencies in conda environment files

cpcloud:
Remove some unused dependencies in our conda env files.

### DIFF
--- a/conda/environment-arm64.yml
+++ b/conda/environment-arm64.yml
@@ -15,12 +15,10 @@ dependencies:
   - filelock >=3.7.0,<4
   - fsspec >=2022.1.0
   - gcsfs
-  - geoalchemy2 >=0.6.3
   - geopandas >=0.6
   - google-cloud-bigquery >=3,<4
   - google-cloud-bigquery-storage >=2,<3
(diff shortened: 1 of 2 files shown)


ncclementi:
Thanks @cpcloud . I'll give these ones a try in the m2 and I'll get back to you.

ncclementi (commented):
The arm64 environment works, only few requests: 
- Would you remove the same packages from the `environment-arm64-flink.yml` too? 
- Would you modify the name of the env to be just `ibis-dev` in both  environment-arm64.yml and `environment-arm64-flink.yml` ? (I forgot to clean this up on the previous PR) - unless we think the names are fine and it's safer to not override an existing dev environment.

cpcloud:
@ncclementi The flink environment doesn't have any of these dependencies.

ncclementi:
> The flink environment doesn't have any of these dependencies.

weird, I thought I started from a copy of the arm64 one.

Well it's good to go then. 🚀

ncclementi (approved):

cpcloud:
Thanks for the review!

PR #8389, opened 2024-02-18
fix(duckdb-sql): ignore importlib package errors when importing ibis.snowflake for transpilation

cpcloud:
Fixes an issue where we try to pluck package metadata for a non-existent package (`snowflake-connector-python`) when transpiling from one dialect to another; in this case duckdb sql to snowflake sql.


cpcloud:
This should get `main` passing again.

PR #8391, opened 2024-02-18
fix(pyspark): ensure to_delta works and is tested in ci

cpcloud:
Fixes an issue where we were not testing `to_delta` for pyspark in CI and breakage to its `to_delta` method after the-epic-split crept in due to the return type of `compile()` being a `str` now instead of pyspark DataFrame.

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -549,18 +549,14 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: pyspark-3.3
-            python-version: "3.9"
-            pyspark-version: 3.3
+          - python-version: "3.9"
+            pyspark-version: "3.3"
             deps:
-              - "pyspark@3.3"
               - "'pandas@<2'"
               - "'numpy@<1.24'"
-          - name: pyspark-3.5
-            python-version: "3.11"
-            pyspark-version: 3.5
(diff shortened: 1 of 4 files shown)


.github/workflows/ibis-backends.yml
```diff
@@ -563,6 +563,7 @@ jobs:
               - "pyspark@3.5"
               - "'pandas@>2'"
               - "'numpy@>1.24'"
+              - "delta-spark"
```
lostmygithubaccount, line 566: so wasn't testing because we didn't have this in CI, so it was import-or-skipping?

cpcloud: Yep!

ibis/backends/pyspark/__init__.py
```diff
@@ -172,30 +171,35 @@ def version(self):
 
     @property
     def current_database(self) -> str:
-        return self._catalog.currentDatabase()
+        [(db,)] = self._session.sql("SELECT CURRENT_DATABASE()").collect()
```
cpcloud, line 174: I had to move our catalog operations to use their SQL equivalents because using the pyspark APIs along with the delta catalog results in SQL parse errors. Very odd 🤷🏻

gforsyth (approved):

PR #8398, opened 2024-02-20
refactor(udfs): consolidate builtin udf compilation and failure modes for unimplemented udfs

cpcloud:
Removes dupcliated udf registration and default failure mode logic for sql backends.


gforsyth (approved):
:ship: it

PR #8400, opened 2024-02-20
fix(ir): fix window boundaries being forcefully casted

### LINKED ISSUE #8368: bug: when the start and end of window boundaries span across data types, ibis forces casting
### What happened?

When the start and end of window boundaries span across data types, ibis forces casting in the window boundaries: https://github.com/ibis-project/ibis/blob/17be43a1081c34582408d4c627191263ec948ada/ibis/expr/builders.py#L177-L182
and this is not executable in 5 backends (with the addition of Flink streaming, which I am working on).

Steps to reproduce this error:
https://github.com/ibis-project/ibis/blob/764c835598944c89cfca55802101d7effd1a6c78/ibis/backends/tests/test_window.py#L489-L520

Example:
```python
alltypes.mutate(val=alltypes.double_col.sum().over(ibis.window(preceding=1000, following=0, group_by=[t.string_col], order_by=[t.id]))
```
The window boundaries here span across integer data types, and the IR generated for this involves a casting. The generated SQL query looks something like
```sql
SELECT
  SUM(`t0`.`double_col`) OVER (ORDER BY `t0`.`id` ASC NULLS LAST ROWS BETWEEN 1000 preceding AND CAST(0 AS SMALLINT) following) AS `Sum(double_col)`
FROM `alltypes` AS `t0`
```
and the `CAST[...] following` syntax is not recognized on certain backends.

The same happens when using range intervals.

8.0

All backends

### Relevant log output

```sh
# MySQL
pymysql.err.ProgrammingError: (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'CAST(0 AS SMALLINT) following) AS `val`\nFROM (\n  SELECT\n    `t0`.`id`,\n    `t...' at line 15")
# RisingWave
sql parser error: Expected literal int, found: CAST at line:15, column:122
Near " BETWEEN 1000 preceding AND CAST"
# Impala
impala.error.HiveServer2Error: AnalysisException: For ROWS window, the value of a PRECEDING/FOLLOWING offset must be a constant positive integer: CAST(0 AS SMALLINT) FOLLOWING
# MS SQL
pyodbc.ProgrammingError: ('42000', "[42000] [FreeTDS][SQL Server]Incorrect syntax near 'CAST'. (102) (SQLExecDirectW)")
```

### Code of Conduct

chloeh13q:
## Description of changes

Fix issue #8368 

## Issues closed

Resolves #8368

### DIFF
--- a/ibis/expr/builders.py
+++ b/ibis/expr/builders.py
@@ -176,6 +176,8 @@ class WindowBuilder(Builder):
 
     def _maybe_cast_boundaries(self, start, end):
         if start and end:
+            if dt.same_base_type(start.dtype, end.dtype):
+                return start, end
             dtype = dt.higher_precedence(start.dtype, end.dtype)
             start = self._maybe_cast_boundary(start, dtype)
             end = self._maybe_cast_boundary(end, dtype)
--- a/ibis/expr/builders.py
+++ b/ibis/expr/builders.py
@@ -176,7 +176,7 @@ class WindowBuilder(Builder):
 
     def _maybe_cast_boundaries(self, start, end):
         if start and end:
-            if dt.same_base_type(start.dtype, end.dtype):
+            if start.dtype.is_same_base_type(end.dtype):
                 return start, end
             dtype = dt.higher_precedence(start.dtype, end.dtype)
             start = self._maybe_cast_boundary(start, dtype)
--- a/ibis/expr/builders.py
+++ b/ibis/expr/builders.py
@@ -168,19 +168,12 @@ class WindowBuilder(Builder):
         else:
             raise IbisInputError("Window frame can only depend on a single relation")
 
-    def _maybe_cast_boundary(self, boundary, dtype):
-        if boundary.dtype == dtype:
-            return boundary
-        value = ops.Cast(boundary.value, dtype)
-        return boundary.copy(value=value)
-
     def _maybe_cast_boundaries(self, start, end):
         if start and end:
-            if start.dtype.is_same_base_type(end.dtype):
-                return start, end
-            dtype = dt.higher_precedence(start.dtype, end.dtype)
-            start = self._maybe_cast_boundary(start, dtype)
-            end = self._maybe_cast_boundary(end, dtype)
+            if start.dtype.is_interval() and end.dtype.is_numeric():
+                return start, ops.Cast(end.value, start.dtype)
+            elif start.dtype.is_numeric() and end.dtype.is_interval():
+                return ops.Cast(start.value, end.dtype), end
         return start, end
 
     def _determine_how(self, start, end):
--- a/ibis/expr/datatypes/cast.py
+++ b/ibis/expr/datatypes/cast.py
@@ -145,26 +145,3 @@ def highest_precedence(dtypes: Iterator[dt.DataType]) -> dt.DataType:
         return functools.reduce(higher_precedence, collected)
     else:
         return dt.null
-
-
-@public
-def same_base_type(source: dt.DataType, target: dt.DataType):
-    if (
-        (source.is_boolean() and target.is_boolean())
-        or (source.is_integer() and target.is_integer())
-        or (source.is_floating() and target.is_floating())
-        or (source.is_decimal() and target.is_decimal())
-        or (source.is_string() and target.is_string())
-        or (source.is_uuid() and target.is_uuid())
-        or (source.is_date() and target.is_date())
-        or (source.is_timestamp() and target.is_timestamp())
-        or (source.is_interval() and target.is_interval())
-        or (source.is_time() and target.is_time())
-        or (source.is_json() and target.is_json())
-        or (source.is_array() and target.is_array())
-        or (source.is_map() and target.is_map())
-        or (source.is_struct() and target.is_struct())
-        or (source.is_geospatial() and target.is_geospatial())
-    ):
-        return True
-    return False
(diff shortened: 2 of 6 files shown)


ibis/expr/datatypes/cast.py
```diff
@@ -145,3 +145,26 @@ def highest_precedence(dtypes: Iterator[dt.DataType]) -> dt.DataType:
...
+        or (source.is_interval() and target.is_interval())
+        or (source.is_time() and target.is_time())
+        or (source.is_json() and target.is_json())
+        or (source.is_array() and target.is_array())
+        or (source.is_map() and target.is_map())
+        or (source.is_struct() and target.is_struct())
+        or (source.is_geospatial() and target.is_geospatial())
+    ):
+        return True
+    return False
```
cpcloud, line 170:
```suggestion
    return type(source) == type(target)
```

Would that work?

chloeh13q: I think in the case of 1000 and 0 one would return int16 and the other would return int8, so we'd have to check whether both inherit from the same immediate parent class, which is also an alternative but I'm wondering if this has more complexity involved

ibis/expr/builders.py
```diff
@@ -176,6 +176,8 @@ def _maybe_cast_boundary(self, boundary, dtype):
 
     def _maybe_cast_boundaries(self, start, end):
         if start and end:
+            if dt.same_base_type(start.dtype, end.dtype):
+                return start, end
             dtype = dt.higher_precedence(start.dtype, end.dtype)
             start = self._maybe_cast_boundary(start, dtype)
             end = self._maybe_cast_boundary(end, dtype)
```
cpcloud, line 183: Just curious, does removing these lines not address the problem?

ibis/expr/datatypes/cast.py
cpcloud: Ah right.

chloeh13q: The alternative would be to check the builtin type of the underlying value (instead of Ibis type), but again I think this might just be simpler 😅

ibis/expr/builders.py
chloeh13q: I tried doing that, and that breaks the windowing logic for at least some backends. For example, if the window is something like `-ibis.interval(seconds=20), 0` where one of the boundary is an interval and the other one is an int (most commonly 0), it ends up erroring out when executed on the backend (e.g. pandas). So essentially, instead of baking the casting logic into the IR we would move the casting logic out to backends where casting is necessary, which is doable but seems like a more complex solution

ibis/expr/datatypes/cast.py
cpcloud: The issue with this implementation is that it adds a place outside of the core datatypes module where we have to remember to update the code if we ever add a datatype. Given how infrequently we add a type it's unlikely we'll remember to update this code.

I think an abstract method on the `DataType` base class method `is_same_base_type(self, other)` that has the individual cases will help avoid creating data types that do not implement this functionality.

It'll be more code, but easier to maintain IMO.

chloeh13q:
@cpcloud Thanks for the suggestion! I implemented the changes.

ibis/expr/operations/window.py
```diff
@@ -71,7 +71,7 @@ class WindowFrame(Value):
     shape = ds.columnar
 
     def __init__(self, start, end, **kwargs):
-        if start and end and start.dtype != end.dtype:
```
kszucs, line 74: Could we use `start.dtype.castable(end.dtype)` here instead of introducing `is_same_base_type()`?

chloeh13q: The logic that we want to implement here is: 
1) when `start` and `end` have the same base type, don't cast even though it's castable (e.g., start=1000, end=0, start is int and end is int8). This is the case that was failing before, because we were using `dt.higher_precendence()` to determine the data type of higher precedence to cast to, and `dt.higher_precendence()` just uses the `castable()` method underneath the hood.
2) else (if `start` and `end` don't have the same base type), if `start` is castable to `end` or `end` is castable to start (e.g., start=ibis.interval(seconds=20), end=0), cast.
3) else (if `start` is not castable to `end` and `end` is not castable to `start`), throw exception.

So case 1 here was not addressed by checking whether the data type is castable.

ibis/expr/builders.py
```diff
@@ -176,6 +176,8 @@ def _maybe_cast_boundary(self, boundary, dtype):
 
     def _maybe_cast_boundaries(self, start, end):
         if start and end:
+            if start.dtype.is_same_base_type(end.dtype):
```
kszucs, line 179: Could you please cover this case in `ibis/expr/tests` so that `pytest -m core` can capture if there is a regression?

ibis/expr/builders.py
kszucs: I dislike introducing `is_same_base_type()` api just to overcome this issue. There are basically two kind of inputs are supported here, numeric and interval. Could we spell out all combinations instead, something like:

```py
def _maybe_cast_boundaries(self, start, end):
    if start.dtype.is_interval() and end.dtype.is_numeric():
        return start, ops.Cast(end, start.dtype)
    elif start.dtype.is_numeric() and end.dtype.is_interval():
        return ops.Cast(start, end.dtype), end
    else:
        return start, end
```

chloeh13q: Sure, I think that's fair.

> Also just check for `rlz.comparable(start, end)` during the `WindowFrame` validation.
I think this would yield the same result as above, when we were to remove the `_maybe_cast_boundaries()` logic entirely. At this point it should be required that window start and window end are both numeric or both intervals.

kszucs: Indeed, then we should validate for that.

kszucs (approved):
LGTM, thanks Chloe!

PR #8405, opened 2024-02-21
refactor(modules): remove unnecessary `base` submodule

cpcloud:
This PR removes an unnecessary level in the `ibis.backends` hierarchy and replaces `SQLGlotBackend` with `SQLBackend`.

PR #8406, opened 2024-02-21
refactor(backend-api): remove the `database()` API and implementation

cpcloud:
Remove the `BaseBackend.database()` API. This has been deprecated for a quite a while. Depends on #8405.


gforsyth (approved):
Approved, but we should wait until #8405 is in (just waiting on CI there)

PR #8408, opened 2024-02-21
c: try configuring algolia

cpcloud:
Playing with algolia configuration


cpcloud:
Kind of a bust, but opened https://github.com/algolia/algoliasearch-netlify/issues/858 for the plugin devs.

PR #8409, opened 2024-02-21
fix(mssql): restore any, all and cumulative versions

### LINKED ISSUE #8073: regression(mssql): investigate bringing back `any`/`all` and cumulative versions
### What happened?

These are now failing on `the-epic-split`

`the-epic-split`

mssql

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
## Description of changes

I took a look through what sqlalchemy was generating for these failing
test cases and then reimplemented it using sqlglot.

Any and All:
* Instead of `logical or` we do a `Max(IIF(condition, 1, 0))`
* Instead of `logical and` we do a `Min(IIF(condition, 1, 0))`

Cumulative versions:

The cumulative versions work, but required an additional check in the
`visit_Not` node. MSSQL doesn't support doing `IFF(MAX(IFF(...` -- the
argument to IFF has to be a boolean and the Window function isn't
considered a boolean for some reason, so the outer conditional needs to
be a case statement.- fix(mssql): restore any, all and cumulative versions
- fix(mssql): support aggregations with any/all

## Issues closed

* Resolves #8073

### DIFF
--- a/ibis/backends/mssql/compiler.py
+++ b/ibis/backends/mssql/compiler.py
@@ -63,8 +63,6 @@ class MSSQLCompiler(SQLGlotCompiler):
 
     UNSUPPORTED_OPERATIONS = frozenset(
         (
-            ops.Any,
-            ops.All,
             ops.ApproxMedian,
             ops.Arbitrary,
             ops.ArgMax,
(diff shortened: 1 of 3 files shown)


jcrist (approved):
Nice.

PR #8410, opened 2024-02-21
docs: use algolia for searchability

ncclementi:
Favor this PR instead of #8325 

Closes: https://github.com/ibis-project/ibis/issues/7995

Needs to happen: 
- [ ] Decide which Algolia account we will use, ideally a team one. 
- [ ] Create an index from the website called `prod_IBIS`
- [ ] Add Algolia API key to secrets (in their site, you can find it under Admin API key) 
- [ ] Test if it works.

To be able to know if this works, idk if the docs-preview label would be enough or if we will have to try this  from a branch, due to the secret.


.github/workflows/upload-algolia-index.yml
```diff
@@ -0,0 +1,29 @@
+# Uploads index to our Algolia Index
+name: Upload Index
+on:
+  workflow_dispatch:
+  repository_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
```
cpcloud, line 7: Should we run this in `ibis-docs-main.yml` instead?

.github/workflows/upload-algolia.py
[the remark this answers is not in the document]
ncclementi: This ended up being kind arbitrary, it was the only way I could get the index to upload. 
Technically the record (each single dict in the search.json) should be smaller than 10_000 bytes. Every other entry in the dictionary except `'text'` was quite small, so I started cutting `'text'` of and it wasn't until I cut it to 1000 that I was able to upload the index. Otherwise I kept getting this error that is unhelpful. 

```
    raise AlgoliaUnreachableHostException("Unreachable hosts")
algoliasearch.exceptions.AlgoliaUnreachableHostException: Unreachable hosts
```

I can explain that, for sure, but it is mainly because it works. Maybe there is another limitation that we are hitting that I'm unaware of.

.github/workflows/upload-algolia-index.yml
ncclementi: I guess we could, this doesn't take to long to run.  was keeping it separate because if this fails, it wouldn't messed with the docs build necessarily but, I don't have a strong opinion. I can move it, if you'd prefer so.

.github/workflows/upload-algolia.py
```diff
@@ -0,0 +1,62 @@
...
+import json
+import os
+from urllib.request import urlopen
+
+from algoliasearch.search_client import SearchClient
+
+api_key = os.getenv("ALGOLIA_API_KEY")
+app_id = os.getenv("ALGOLIA_APP_ID")
+index_name = "test_IBIS"  # os.getenv("ALGOLIA_INDEX")
+index_file = os.getenv("QUARTO_INDEX_PATH")  # search.json
```
cpcloud, line 12: can we use `os.environ` instead of the lower-level `os.getenv`?

.github/workflows/upload-algolia.py
```diff
@@ -0,0 +1,62 @@
...
+        with open(index_file, "wb") as f:
+            f.write(response.read())
+
+    client = SearchClient.create(app_id, api_key)
+    index = client.init_index(index_name)
+
+    with open(index_file) as f:
+        search = json.load(f)
+
+    # According to algolia docs, for the build plan each record (in our case this
```
cpcloud, line 40: Can you put a link to those docs here?

.github/workflows/upload-algolia.py
```diff
@@ -0,0 +1,62 @@
...
+    truncated_string = truncated_bytes.decode("utf-8", errors="ignore")
+
+    return truncated_string
+
+
+def main():
+    # Download the index generated by quarto from the ibis website
+    with urlopen("https://ibis-project.org/search.json") as response:
+        with open(index_file, "wb") as f:
+            f.write(response.read())
```
cpcloud, line 32: Do we actually need to write to disk here? How about calling `json.load(f)` directly instead of writing and then reading?

ncclementi:
I'm not sure I understand what's wrong with the PR title, that ci is complaining about. I check the ci job, but it's unclear to me what's happening.

.github/workflows/upload-algolia.py
cpcloud: Actually, how about `json.loads(response.read().decode("utf-8"))`?

.github/workflows/upload-algolia.py
```diff
@@ -0,0 +1,62 @@
...
+    # I opened an issue because the error is unhelpful, and unclear.
+    # https://github.com/algolia/algoliasearch-client-python/issues/565
+
+    # It wasn't until I cut the "text" field to max_size=1000 that I was able to
+    # get an index. My guess is we are hitting another limitation, but I was not
+    # able to find anything on the docs.
+
+    max_size = 1_000
+    for obj in search:
+        if len(obj["text"].encode("utf-8")) >= max_size:
```
cpcloud, line 53: Bit of a nit, but don't we want to only truncate if the string is strictly larger than max_size?

.github/workflows/upload-algolia-index.yml
cpcloud: I think we should move it so that we get CI failures if the script fails instead of it silently failing and us forgetting to look it at for a long time.

We can determine what to do depending on the nature of the failures if/when they happen.

cpcloud:
> I'm not sure I understand what's wrong with the PR title, that ci is complaining about. I check the ci job, but it's unclear to me what's happening.

There's a leading whitespace per https://github.com/ibis-project/ibis/actions/runs/7993226046/job/21828534190?pr=8410#step:5:38

image

cpcloud:
I think others have also experienced this issue. Is there some tool that's prepending and/or appending whitespace to commit headers?

ncclementi:
> I think others have also experienced this issue. Is there some tool that's prepending and/or appending whitespace to commit headers?

This is the first time that happens to me, and I use vim for commits, I did edit the title of the PR, I might have accidentally introduced a space, I have no clue 🤷‍♀️

.github/workflows/upload-algolia.py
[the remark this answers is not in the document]
cpcloud: Thanks, your comments below are 👌🏻

cpcloud (approved):

cpcloud (approved):

cpcloud:
Merging so we can iterate.

PR #8411, opened 2024-02-21
fix(mssql): restore unbounded window functions

### LINKED ISSUE #8072: regression(mssql): investigate bringing back unbounded window frame window functions
### What happened?

These are failing after `the-epic-split`

`the-epic-split`

mssql

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
## Description of changes

MSSQL requires an order-by clause in unbounded windows.  This is clearly
documented in
https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/database-engine-events-and-errors-10000-to-10999
(just kidding, it's a huge list of error codes you have to scroll through.  You
want Error 10756)

The behavior of the mssql sqlalchemy connector is to order by the field used in
the window function.  This seems silly, and potentially slightly
computationally wasteful, but not incorrect.

So, I've added a rewrite rule to the MSSQL compiler that copies this behavior
for all reduction functions that aren't UDFs, since those aren't supported
anyway (and would require a different rewrite rule, in any case).

## Issues closed

* Resolves #8072

### DIFF
--- a/ibis/backends/mssql/compiler.py
+++ b/ibis/backends/mssql/compiler.py
@@ -22,6 +22,8 @@ from ibis.backends.base.sqlglot.dialects import MSSQL
 from ibis.backends.base.sqlglot.rewrites import (
     exclude_unsupported_window_frame_from_ops,
     exclude_unsupported_window_frame_from_row_number,
+    p,
+    replace,
     rewrite_first_to_first_value,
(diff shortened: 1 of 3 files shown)


cpcloud (approved):
LGTM, thanks for digging around in the muck!

Issue #8247, opened 2024-02-06
feat(flink): support event-time temporal join


mfatihaktas:
### Is your feature request related to a problem?

Event-time temporal join in Flink enables joining a table against a **versioned** table. Tables in Flink are temporal/dynamic, i.e., row values can change over time, or rows can be added or deleted. A versioned table contains one or more versioned table snapshots. With event-time temporal join, a table can be enriched with values retrieved from a versioned table at a certain point in time.

More information is available on the [Flink doc](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sql/queries/joins/#event-time-temporal-join).

### Describe the solution you'd like

We previously implemented this support in this PR: https://github.com/ibis-project/ibis/pull/7921. We put it on hold during the reviews due to two reasons:
- We added temporal join support under `asof_join()`, and temporal join does not fit into the semantics of `asof-join` very well.
  - Decided to add a new interface function for temporal join during the PR review.
- Implementation of temporal join required some "compiler hacking". @cpcloud mentioned in his review of https://github.com/ibis-project/ibis/pull/7921 that such compiler hacking would not be needed after the sqlglot refactoring.

Roughly, the new API for event-time temporal join would look like:
```python
table_left = con.create_table(...)
table_right = con.create_table(...)

expr = table_left.temporal_join(
        table_right,
        predicates=[
            table_left["key"] == table_right["key"],
            table_left["time"] >= table_right["time"],
        ],
    )
```

### What version of ibis are you running?

7.2.0

### What backend(s) are you using, if any?

Flink

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ogrisel:
It might be important to distinguish between system time (table history managed automatically by the DBMS system) and business/validity time managed by the application/user.

System time might be what you get from time travel system as implemented in Apache Iceberg, Delta Lake and SQL 2011 temporal tables.

Bi-temporal modeling can be useful in many contexts, and Ibis might be required to know how system and valid time ranges are handled differently by the underlying backend.

Tri-temporal modeling (with an additional decision time) might be also be useful but adds an other layer of complexity and I am not sure Ibis has to do anything about that extra modeling layer. I might be 100% managed by application code.

- https://en.wikipedia.org/wiki/Temporal_database
- https://en.wikipedia.org/wiki/Bitemporal_modeling
- https://en.wikipedia.org/wiki/Decision_time

ogrisel:
[...]

PR #8412, opened 2024-02-21
feat(flink): temporal join support

mfatihaktas:
## Description of changes

Explorative PR towards addressing https://github.com/ibis-project/ibis/issues/8247.

This is my first time getting hands-on with `sqlglot` changes :) So, particularly seeking suggestions on
- Temporal join API
- The newly added op `VersionedDatabaseTable` and expression `TemporalJoin`.

**Note**: Added support only in the Flink backend for now. Did not spend time on how to error out when the user calls `tempora_join()` on a backend that does not support temporal join. Plan to address these once we reach an agreement on the API.

### DIFF
--- a/ibis/backends/flink/tests/test_ddl.py
+++ b/ibis/backends/flink/tests/test_ddl.py
@@ -609,7 +609,7 @@ def test_temporal_join(con, table_left_and_right_and_source_df):
     table_left, table_right, source_df = table_left_and_right_and_source_df
 
     table_right = table_right.at_time(table_left.timestamp_col)
-    join_expr = table_left.temporal_join(
+    join_expr = table_left.join(
         table_right,
         predicates=[
             table_left["id"] == table_right["id"],
@@ -630,7 +630,7 @@ def test_temporal_join(con, table_left_and_right_and_source_df):
   `t3`.`int_col` AS `int_col_right`,
   `t3`.`timestamp_col` AS `timestamp_col_right`
 FROM `table_left` AS `t1`
-JOIN `table_right` FOR SYSTEM_TIME AS OF `t1`.`timestamp_col` AS `t3`
+INNER JOIN `table_right` FOR SYSTEM_TIME AS OF `t1`.`timestamp_col` AS `t3`
   ON `t1`.`id` = `t3`.`id`"""
     assert sql == expected_sql
 
--- a/ibis/backends/flink/tests/test_ddl.py
+++ b/ibis/backends/flink/tests/test_ddl.py
@@ -591,7 +591,6 @@ def compare_temporal_joined_and_expected_dfs(
     joined_df = join_expr.to_pandas().sort_values("id")
     joined_df = joined_df.reindex(sorted(joined_df.columns), axis=1)
     joined_df = joined_df.drop("timestamp_col_right", axis=1)
-    # print(f"joined_df= \n{joined_df.to_string(index=False)}")
 
     expected_df = pd.merge_asof(
         source_df,
@@ -600,7 +599,6 @@ def compare_temporal_joined_and_expected_dfs(
         by="id",
     ).sort_values("id")
     expected_df = expected_df.reindex(sorted(expected_df.columns), axis=1)
-    # print(f"expected_df= \n{expected_df.to_string(index=False)}")
 
     assert (joined_df.values == expected_df.values).all()
 
@@ -617,7 +615,6 @@ def test_temporal_join(con, table_left_and_right_and_source_df):
     )
 
     sql = con.compile(join_expr)
-    print(f"sql= \n{sql}")
 
     expected_sql = """SELECT
   `t1`.`id`,
--- a/ibis/backends/flink/tests/test_ddl.py
+++ b/ibis/backends/flink/tests/test_ddl.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 import tempfile
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 import pandas as pd
 import pandas.testing as tm
@@ -15,6 +16,10 @@ import ibis.expr.datatypes as dt
 import ibis.expr.schema as sch
 from ibis.backends.conftest import TEST_TABLES
 from ibis.backends.tests.errors import Py4JJavaError
+
+
+if TYPE_CHECKING:
+    import ibis.expr.types as ir
 
 
 @pytest.fixture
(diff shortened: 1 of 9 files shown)


ibis/backends/flink/tests/test_ddl.py
```diff
@@ -534,3 +534,83 @@ def test_to_parquet(con, tmp_path, table_name):
...
+        primary_key="id",
+    )
+    con.insert(
+        "table_right",
+        obj=df,
+        schema=schema,
+    ).wait()
+
+    # Join `table_left` and `table_right`
+    expr = table_left.temporal_join(
```
kszucs, line 589: How about a slightly different API `left.temporal_join(right.at_time(right.timestamp))` ?

That way the versioned table operation could hold the `at_time` property which could eliminate the majority of custom join logic.

ibis/expr/operations/relations.py
```diff
@@ -330,6 +340,12 @@ class DatabaseTable(PhysicalTable):
     namespace: Namespace = Namespace()
 
 
+@public
+class VersionedDatabaseTable(DatabaseTable):
```
kszucs, line 344: So `at_time` could be placed here, and the `JoinKind` literal could have a new member to mark temporal joins.

ibis/backends/flink/tests/test_ddl.py
mfatihaktas: Thanks, I agree that this would be nicer. I added the API you suggested along side the initial one, just to keep the previous one as a reference until getting temporal join working.

ibis/expr/operations/relations.py
mfatihaktas: Thanks, made the suggested changed.

ibis/backends/sql/compiler.py
```diff
@@ -1134,6 +1135,105 @@ def visit_DatabaseTable(
...
+        #
+        # As much as I can see, there are two ways to deal with stale table references:
+        # - Keep track of the table evolution and use the same alias for different copies
+        # of the same table. Might require taking the `alias` dict out of `translate()`
+        # and exposing it for "global use".
+        #   - IMO, this would introduce a lot of tech-debt. May make sense only in the
+        #     presence of other use cases.
+        # - Implement a way to turn off aliasing for temporal joins.
+        #  - This also seems to require additional tech debt.
+
```
mfatihaktas, line 1223: @kszucs I documented an issue that I could not find a solution for as a long comment in `visit_VersionedDatabaseTable()`. To summarize, I am not able to get `at_time` compile with the right alias due to tables getting copied internally. By the time, SQL is generated for the `nodes`, `at_time` is bound to a "stale" reference of the corresponding table. This leads to generating an incorrect SQL for `at_time`, e.g., `t0.timestamp_col` while it is supposed to be `t2.timestamp_col`. This issue arises in both the initial API and the one you suggested.

I failed resolving/working around this in my initial attempts (though learned quite a bit about the compilation process in Ibis) but will continue looking into this. In the comment, I noted the two approaches that came to my mind but both are quite tedious and likely to be unnecessary. I feel that I am missing something obvious due to my lack of sufficient knowledge of the code base :) Do you have any suggestion for me?

mfatihaktas: Reading the source code more, I might be wrong about my judgement of the table copies "spoiling" the table references. The table reference disconnect for `at_time` seems to be originating from the `JoinTable`'s constructed off of `DatabaseTable` -- see the copied IR graph. Will update here once I get a better understanding.

image

kszucs: We need to wrap the join participant tables in `JoinTable` relations, so that the same table can be added to the join several times. This way we ensure that all tables in the join are distinct and the predicates reference distinct tables. 

This has the drawback that all value expressions referencing the wrapped table, must be replaced with the JoinTable, like `Field(r) => Field(JoinTable(r))`. We call this "dereferencing" and handled by `prepare_predicates()`.
This means that both `table_left` and `table_right` should only be connected to the respective `JoinTable` nodes, but there is an exception in the graph, the `timestamp_col` which indicates that it is not getting dereferenced to the join tables. Most certainly the problem is with the `at_time` here:

```py
link = ops.TemporalJoinLink(how="temporal", table=right, at_time=at_time, predicates=preds)
```

before passing to the join link `at_time` should be dereferenced:

```py
deref_left = dereference_mapping_left(chain)
at_time = at_time.replace(deref_left, filter=ops.Value)
```

Let me know whether that helps or not, then we can iterate with this PR. In the meantime I am trying to have a more detailed look at flink temporal joins.

mfatihaktas: Thanks @kszucs for the detailed explanation. Reading the source code a bit more, I was able to figure that the main reason is indeed the wrapping of participant tables in `JoinTable`. For the initial API, I was able to dereference `at_time` as you explained, but using a much uglier hack 😄:
```python
    def rebase_at_time_on_join_table(self, at_time: Column):
        """This is a "hack" only to demonstrate the challenge with generating
        the correct table alias for `at_time`. Shall be refactored out.

        Creates and returns a new `at_time` such that
        * New `at_time` is created from JoinTable `join_table`
        * `join_table` got created in `relations.py: temporal_join()` from DatabaseTable
        `database_table`.
        * The arg `at_time` got created from `database_table` by the user in the first
        place.

        """
        from ibis.common.graph import Graph

        at_time_source_table = at_time.op().find(finder=ops.DatabaseTable)[0]

        op = self.op()
        graph, dependents = Graph.from_bfs(op).toposort()
        at_time_join_table_op = dependents[at_time_source_table][0]
        at_time_join_table = Table(at_time_join_table_op)
        return at_time_join_table[f"{at_time.op().name}"]
```
I will replace this with your suggestion and share the update here.

For the second API (initiated by your review), I was not able to dereference `at_time` successfully -- my dereferencing "hack" messed up the dereferencing done for the predicates. I will try doing it with `dereference_mapping_left()` and `replace()`, and will share the update here.

mfatihaktas: @kszucs I was able to use `dereference_mapping_left()` and `replace()` to "rebase" the `at_time` field on the left `JoinTable`. I added `TODO` comments for things that I am not certain about.

kszucs: Thanks! Going to have another look hopefully tomorrow.

ibis/expr/types/joins.py
```diff
@@ -408,6 +409,68 @@ def asof_join(
...
+        at_time: Column = None,
+        lname: str = "",
+        rname: str = "{name}_right",
+    ):
+        if not isinstance(right.op(), ops.VersionedDatabaseTable):
+            raise IbisError(
+                "Right-table is not versioned. "
+                "Temporal join is defined only when the right-table is versioned."
+            )
+        elif at_time is None and right.at_time is None:
```
kszucs, line 427: Is this just a convenience to support defining `at_time` for both the table and the join operation?

What I am thinking is to restrict the API here so that the `ops.VersionedDatabaseTable`'s `at_time` attribute become mandatory and we remove the `at_time` argument here, something like:

```py
ordinary = con.table("my_ordinary_table")
versioned = con.table("my_versioned_table")

ordinary.temporal_join(
    versioned.at_time(ordinary.timestamp_col),
    [...additional predicates...]
)
```

That way we could remove the `ops.TemporalJoinLink` node as well, since it wouldn't hold an extra `at_time` property since that is always defined on the right table (`how` would carry the information about the join kind).

kszucs:
Is there another ibis backend which supports this kind of temporal join? I would like to avoid introducing global APIs highly tailored towards certain backends since it would defeat the purpose of ibis. Having it implemented for another backend could validate the API and IR design we have here.

kszucs:
Just had a closer look at the issue description which you have nicely collected the necessary information in, apparently the options are RisingWave and MySQL. How much work would it be to support risingwave here?

ibis/expr/types/joins.py
kszucs: Actually the nicest would probably be your example from #8247:

```py
expr = table_left.join(
    table_right.at_time(table_left.event_time_attribute),
    table_left.on_field == table_right.on_field
)
```

If the right table passed to `.join()` is a versioned table, then the join must be a temporal join.

ibis/expr/operations/relations.py
```diff
@@ -330,6 +340,13 @@ class DatabaseTable(PhysicalTable):
...
```
kszucs, line 344: Can other table-like objects be versioned? I'm thinking of views or generic `SELECT` queries here. If yes, it might be better to have a `TableSnapshot` relation instead referencing an arbitrary parent relation, like:

```py
class TableSnapshot(Relation):
    parent: Relation  # this would be an ops.DatabaseTable most of the cases
    at_time: Column[dt.Timestamp]
```

cpcloud: I believe we already discussed this, and the answer was no. Even if it turns out to be wrong, it should be addressed in a follow up PR.

ibis/expr/types/joins.py
mfatihaktas: Thanks. As you raised this earlier, I also like this API (API-2) more compared to the first one (API-1) requiring the `at_time` argument in `temporal_join()`. I have kept both API's so far in case one would require a more complex implementation than the other and that might influence our opinion about which one to pick. This did not turn out to be the case, so I will remove the option with `at_time` argument.

> If the right table passed to .join() is a versioned table, then the join must be a temporal join.

This is a good idea, will make this change.

There are, however, a few things that are currently not ideal with the implementation of API-2. I documented these items in `TODO`'s. Here is a summary of them:

- The current implementation requires `table_right.at_time(...)` to be on a separate line as
```
        table_right = table_right.at_time(table_left.timestamp_col)
        expr = table_left.temporal_join(
            table_right,
            predicates=[
                table_left["id"] == table_right["id"],
            ],
        )
```
If we do not over-write `table_right` as given above, i.e.,
```
        expr = table_left.temporal_join(
            table_right.at_time(table_left.timestamp_col),
            predicates=[
                table_left["id"] == table_right["id"],
            ],
        )
```
we will get
```
        E ibis.common.exceptions.IntegrityError: Cannot add <ibis.expr.operations.logical.Equals object at 0x17c374270> to projection, they belong to another relation
```
This is because `at_time()` returns a new copy of `table_right` and the predicates are defined with the old copy of `table_right`. 

- On removing `ops.TemporalJoinLink`:

> That way we could remove the ops.TemporalJoinLink node as well, since it wouldn't hold an extra at_time property since that is always defined on the right table (how would carry the information about the join kind).

I tried to implement this exact idea previously, and access `at_time` through `join_op -> table_right_op -> at_time_op` -- `join_sql()` in `dialects.py` is actually a deprecated-leftover from this effort and kept it for reference during the reviews. 
We first need to rebase `table_right.at_time` on `join_table_right` with
```
        # Rebase `at_time` on the JoinTable wrapping up the DatabaseTable
        # `at_time` field is based on.
        deref_left = dereference_mapping_left(left)
        at_time = at_time_op.replace(deref_left, filter=ops.Value)
```
Then to keep a reference to the rebased `at_time`, we need to make it an attribute of `right`. The only way that I could come up to do this was
``` 
right = right.op().copy(at_time=at_time)
```
However, this lead to the same error given above
```
        E ibis.common.exceptions.IntegrityError: Cannot add <ibis.expr.operations.logical.Equals object at 0x17c374270> to projection, they belong to another relation
```
This is again because now the (old) `right` table referenced in the predicate is different than the new `right` table. As far as I understand it, to get this working, we would need to have an additional step where we update the relevant `predicates` with the new `right` before calling `prepare_predicates()`.

@kszucs Do you have suggestions or ideas on addressing these two points?

mfatihaktas:
> Is there another ibis backend which supports this kind of temporal join? I would like to avoid introducing global APIs highly tailored towards certain backends since it would defeat the purpose of ibis. Having it implemented for another backend could validate the API and IR design we have here.

> Just had a closer look at the issue description which you have nicely collected the necessary information in, apparently the options are RisingWave and MySQL. How much work would it be to support risingwave here?

Adding support for `RisingWave` would have been great, but RisingWave seems to support only [`processing-time` temporal join](https://docs.risingwave.com/docs/current/query-syntax-join-clause/#process-time-temporal-joins). This PR adds support for only `event-time` temporal join. In `event-time` temporal join, rows are joined with their right-table versions specified by a given time-attribute (`at_time` in this PR). In `processing-time` temporal join, rows are always joined with their most-recent version in the right-table.

For `MySQL`, I could not find enough information on whether it supports temporal join or not. At the time, I could find only [this article](https://phpenterprisesystems.com/databases/212-how-to-work-joins-with-temporal-tables-in-mysql#:~:text=Working%20with%20joins%20in%20temporal,table%20based%20on%20specific%20conditions.) that discusses support for performing joins against `temporal tables`, which does not have the same semantics as Flink's temporal join. In my understanding, it is for joining against a particular version of the table that is specified by a given `timestamp` rather than a `time attribute`. So it is like first performing a time travel on the table(s) and then joining them.

ibis/expr/operations/relations.py
mfatihaktas: In my understanding, intermediate tables in Flink such as `SELECT` queries can be versioned and queried only with [Paimon](https://www.ververica.com/blog/apache-paimon-the-streaming-lakehouse) in place. Without `Paimon`, Flink supports [`versioned views`](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/concepts/versioned_tables/#versioned-table-views), against which temporal join can be performed.

ibis/expr/types/joins.py
mfatihaktas: I just "discovered" that Flink supports two different kinds of temporal join with `LEFT` and `INNER`. I did not read this in the docs, just found out by trying it out. With this, `how` in `TemporalJoinLink` would be set to either `left` or `right`. This might be another reason to keep `TemporalJoinLink` rather than "encoding" the join type in `JoinLink` as `temporal-left` or `temporal-inner`.

cpcloud:
Closing, as I don't think I can maintain this code at the moment. We can always revisit later!

PR #8414, opened 2024-02-21
fix(sql): support set operations wrapping subqueries

### LINKED ISSUE #8561: bug: `Parser Error: syntax error at or near "UNION"`
### What happened?

```python
import ibis

t = ibis.memtable({"x": [1, 2, 3, 4, 5, 6]})
u = ibis.union(
    t.filter(t.x <= 3).order_by(ibis.random()).limit(2),
    t.filter(t.x > 3).order_by(ibis.random()).limit(2),
)
u.execute()
```

result of `ibis.to_sql(u)`:

```sql
SELECT
  "t5"."x"
FROM (
  SELECT
    "t0"."x"
  FROM "ibis_pandas_memtable_sa56tjeudndxbhuxbbtgdj5oua" AS "t0"
  WHERE
    "t0"."x" <= CAST(3 AS TINYINT)
  ORDER BY
    RANDOM() ASC
  LIMIT 2
  UNION ALL
  SELECT
    "t0"."x"
  FROM "ibis_pandas_memtable_sa56tjeudndxbhuxbbtgdj5oua" AS "t0"
  WHERE
    "t0"."x" > CAST(3 AS TINYINT)
  ORDER BY
    RANDOM() ASC
  LIMIT 2
) AS "t5"
```

not exactly sure what the problem is here?

main

_No response_

### Relevant log output

_No response_

### Code of Conduct

jcrist:
Previously this would generate invalid SQL if the subqueries repeated components of a SELECT (e.g. included an `ORDER BY` clause).

Fixes #8561.


jcrist:
This is failing for two reasons:

- Some snapshot tests. I can update those if we think this fix is valid/useful.
- `mssql` doesn't support `ORDER BY` clauses (lacking a limit or offset) in subqueries. Can mark these tests as failing for `mssql`.

 A different but still kinda valid fix would be to strip `ORDER BY` clauses without limit/offset from subqueries in set operations, since the backend is free to ignore ordering of these subqueries. Mainly trying to avoid these queries failing at execution time due to generating invalid SQL.

cpcloud:
This seems like a useful fix to me. How did you come across it?

cpcloud (approved):
SQL is so annoying sometimes.

PR #8421, opened 2024-02-22
chore: resurrect `SupportsSchema` type; clean up unused ignored imports

cpcloud:
Bring back `SupportsSchema`

### DIFF
--- a/ibis/common/typing.py
+++ b/ibis/common/typing.py
@@ -4,24 +4,26 @@ import re
 import sys
 from abc import abstractmethod
 from itertools import zip_longest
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
No objections -- was there a particular issue that prompted this?

cpcloud:
Cleaning up some of the documentation build warnings discovered during Algolia inception.

cpcloud:
Just to put a finer point on it: we're using `SupportsSchema` as a type annotation for some APIs' input parameters and quartodoc is generating qmd files that refer to the previously non-existent type.

Issue #8250, opened 2024-02-06
meta: increase Flink's streaming backend coverage


chloeh13q:
### Is your feature request related to a problem?

While we have implemented support for the basic operators in the Flink backend, there are still some operators that are commonly seen in streaming workloads that we have not yet implemented support for. Some of these are pending the sqlglot refactoring work.

### Describe the solution you'd like

1. [DONE] ~~[P1] Alternative syntax for top k (follow-up to #7407)~~
2. [P1] Deduplication using `distinct()` (pending #7556)
3. [P0] Array expansion (pending sqlglot refactoring)
4. [P0] `ops.ArrayCollect` (pending UDF support)
5. [P2] Window joins with more complex syntax (`ANTI`/`SEMI`) (follow-up to #7966)
6. [P0] Temporal join (pending sqlglot refactoring) (follow-up to #7921)
7. [P2] Time travel query (pending sqlglot refactoring) (#8203)
8. [P2] [Pattern recognition](https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/dev/table/sql/queries/match_recognize/) (#8252)
9. [DONE] ~~[P0] `MAP` support (covered by `ibis/backends/tests/test_map.py`)~~
10. [P2] #8254 (testing)

### What version of ibis are you running?

8.0

### What backend(s) are you using, if any?

Flink

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

deepyaman:
Propose to include 9. `MAP` support in this issue; covered by `ibis/backends/tests/test_map.py` and useful to get UDFs with `map` working in https://github.com/ibis-project/ibis/pull/8142.

mfatihaktas:
Temporal join issue: https://github.com/ibis-project/ibis/issues/8247

chloeh13q:
Weekly update [2/29/24]

3. Array expansion: issue opened (#8457), feature implementation WIP
6. Temporal join WIP: issue opened (#8247), draft PR #8412 
10. Depends on #7712, draft PR #8343 to address #7712 

No update on remaining items.

chloeh13q:
Weekly update [3/6/24]

## P0

[3 - Array expansion] - PR in review #8511 

[4 - `ops.ArrayCollect`] - issue opened #8555, pausing implementation until Flink 1.20 release

[6 - Temporal join] - issue opened #8247, blocked by #8537 

[9 - MAP support] - DONE


## P1

[...]

PR #8425, opened 2024-02-22
feat(flink): add map support

mfatihaktas:
## Description of changes

Adds `MAP` support for Flink backend. Addresses item 9 in https://github.com/ibis-project/ibis/issues/8250.

### DIFF
--- a/ibis/backends/flink/compiler.py
+++ b/ibis/backends/flink/compiler.py
@@ -55,7 +55,6 @@ class FlinkCompiler(SQLGlotCompiler):
             ops.IsInf,
             ops.IsNan,
             ops.Levenshtein,
-            ops.MapMerge,
             ops.Median,
             ops.MultiQuantile,
             ops.NthValue,
@@ -81,7 +80,8 @@ class FlinkCompiler(SQLGlotCompiler):
         ops.ExtractDayOfYear: "dayofyear",
         ops.First: "first_value",
         ops.Last: "last_value",
-        ops.Map: "map_from_arrays",
+        ops.MapKeys: "map_keys",
+        ops.MapValues: "map_values",
         ops.Power: "power",
         ops.RandomScalar: "rand",
         ops.RegexSearch: "regexp",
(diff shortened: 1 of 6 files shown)


ibis/backends/flink/compiler.py
```diff
@@ -541,3 +541,33 @@ def visit_CountDistinct(self, op, *, arg, where):
...
+            keys = self.cast(keys, dt.Array(non_nullable_dtype))
+
+        return self.f.map_from_arrays(keys, values)
+
+    def visit_MapMerge(self, op: ops.MapMerge, *, left, right):
+        def get_non_nullable_keys(op, expr: sge.Expression) -> sge.Expression:
+            keys = self.f.map_keys(expr)
+            if op.dtype.key_type.nullable:
+                non_nullable_dtype = op.dtype.key_type.copy(nullable=False)
+                keys = self.cast(keys, dt.Array(non_nullable_dtype))
```
cpcloud, line 560: Is `array<int not null>` actually valid flink syntax?

cpcloud: I'm not following from the code why nullability matters here.

ibis/backends/sql/datatypes.py
```diff
@@ -1038,6 +1038,16 @@ class FlinkType(SqlglotType):
     default_decimal_precision = 38
     default_decimal_scale = 18
 
+    @classmethod
+    def from_ibis(cls, dtype: dt.DataType) -> sge.DataType:
+        """Convert a sqlglot type to an ibis type."""
+
+        typ = super().from_ibis(dtype)
+        if not dtype.nullable:
+            typ.set("kind", sge.NotNullColumnConstraint())
```
cpcloud, line 1047: `NotNullColumnConstraint()` isn't a valid type of value for `kind`.

Can you write down the thing you're trying to do and we can work towards a solution?

ibis/backends/flink/compiler.py
cpcloud: Apparently:

1. Yes that is valid flink syntax
2. Nullability matters because it is enforced by flink

ibis/backends/flink/compiler.py
```diff
@@ -541,3 +541,21 @@ def visit_CountDistinct(self, op, *, arg, where):
...
+        left_keys = self.f.map_keys(left)
+        left_values = self.f.map_values(left)
+
+        right_keys = self.f.map_keys(right)
+        right_values = self.f.map_values(right)
+
+        keys = self.f.array_concat(left_keys, right_keys)
+        values = self.f.array_concat(left_values, right_values)
+
+        return self.cast(self.f.map_from_arrays(keys, values), op.dtype)
```
cpcloud, line 561: Just noting here that I can't tell from [the docs for `map_from_arrays`](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/functions/systemfunctions/) what the behavior of duplicates keys is here.

cpcloud (approved):

ibis/backends/sql/datatypes.py
```diff
@@ -1041,3 +1043,17 @@ class FlinkType(SqlglotType):
     @classmethod
     def _from_ibis_Binary(cls, dtype: dt.Binary) -> sge.DataType:
         return sge.DataType(this=sge.DataType.Type.VARBINARY)
+
+    @classmethod
+    def _from_ibis_Map(cls, dtype: dt.Map) -> sge.DataType:
+        # key cannot be nullable in clickhouse
+        key_type = cls.from_ibis(dtype.key_type.copy(nullable=False))
+        value_type = cls.from_ibis(dtype.value_type)
+        return sge.DataType(
```
mfatihaktas, line 1052: Nice! I could not figure this :)

ibis/backends/flink/compiler.py
mfatihaktas: I also could not find information on the exact behavior of Flink for this. Running the tests, I "inferred" that Flink uses the last value for each key in the order they appear in the array.

cpcloud: Makes sense, thanks for looking into it!

PR #8427, opened 2024-02-22
refactor: rename SupportSchema -> SchemaLike, fix type definition

jcrist:
This is a follow-up to #8421.

- The use of a `TypeVar` here was incorrect - what you want is a `TypeAlias` instead. The previous version would cause type checkers to error.
- I moved the type definition to be adjacent to `Schema`. This makes the import paths cleaner since you don't need as many `if TYPE_CHECKING` blocks. Logically it also makes sense to me to put the union of schema-like things next to the `Schema` code.
- I renamed the type alias to `SchemaLike` which reads better IMO.


cpcloud (approved):
Thanks!

Issue #8356, opened 2024-02-14
feat(bigquery): Include `offset` parameter to unnest


chelsea-lin:
### Is your feature request related to a problem?

It's not possible to express that I want to save the original order, such as with the [WITH OFFSET AS clause for UNNEST in BigQuery](https://cloud.google.com/bigquery/docs/arrays#flattening_arrays). With the current version of [unnest](https://ibis-project.org/reference/expression-collections#ibis.expr.types.arrays.ArrayValue.unnest), it only returns a single column. I suppose it would be possible to add a with_offset parameter here, but then the return type might have to be a struct column or maybe a tuple of two columns.

This issue has the same request as #7781

### Describe the solution you'd like

I'd like to add an `offset` parameter to the `unnest` function. When `offset` is `True`, the compiler can returns a `sge.Posexplode()` rather than `sge.Explode()` node. Then, sqlglot can handle the SQL translation as expected. Here is the draft PR: https://github.com/ibis-project/ibis/pull/8354/files
With that change, the translated SQL looks good to me:
```
In [1]: import ibis
   ...: bq = ibis.bigquery.connect(project_id="bigframes-dev")
   ...: table = bq.table("bigframes-dev.ibis_gbq_testing.array_types")
   ...: ibis.to_sql(table.z.unnest(offset=True), dialect="bigquery")
Out[1]: 
SELECT
  IF(pos = pos_2, `z`, NULL) AS `z`,
  IF(pos = pos_2, pos_2, NULL) AS pos_2
FROM `bigframes-dev`.`ibis_gbq_testing`.`array_types` AS `t0`
CROSS JOIN UNNEST(GENERATE_ARRAY(0, GREATEST(ARRAY_LENGTH(`t0`.`z`)) - 1)) AS pos
CROSS JOIN UNNEST(`t0`.`z`) AS `z` WITH OFFSET AS pos_2
WHERE
  pos = pos_2
  OR (
    pos > (
      ARRAY_LENGTH(`t0`.`z`) - 1
    ) AND pos_2 = (
      ARRAY_LENGTH(`t0`.`z`) - 1
    )
  )
```

However, the ibis node type is not correct. The ibis assumes this node only returns a single value and ignore the `pos_2` here. Here is the repro code:
```
In [3]: type(table.z.unnest(offset=True))
Out[3]: ibis.expr.types.numeric.FloatingColumn

In [7]: table.select(["scalar_column", table.z.unnest(offset=True)])
Out[7]: 
[...]

PR #8429, opened 2024-02-22
feat(duckdb): add enumerate function to provide access to array element offsets

cpcloud:
Draft PR to demonstrate the API for `enumerate` discussed in #8356. Happy to expand the implement to other backends as needed.


ibis/backends/duckdb/compiler.py
```diff
@@ -400,3 +400,22 @@ def visit_HexDigest(self, op, *, arg, how):
 
     def visit_StringConcat(self, op, *, arg):
         return reduce(lambda x, y: sge.DPipe(this=x, expression=y), arg)
+
+    def visit_Enumerate(self, op, *, arg, start: int):
+        # adapted from visit_ArrayZip
+        i = sg.to_identifier("i")
```
cpcloud, line 406: This implementation isn't correct, it's not enumerating globally when it should be.

PR #8433, opened 2024-02-23
test(snowflake): xfail on test with window larger than 1000

cpcloud:
Fixes the failing upstream snowflake test caused by snowflake not supporting window frames sizes larger than 1000.


cpcloud:
cc @chloeh13q

chloeh13q (approved):
LGTM!

PR #8438, opened 2024-02-23
feat(api): add a `uuid` function for returning a new uuid

### LINKED ISSUE #8257: feat: expose a function for generating a new uuid
Some backends have functions for returning a newly generated UUID (e.g. postgres's [`gen_random_uuid`](https://www.postgresql.org/docs/current/functions-uuid.html) or duckdb's [`uuid`](https://duckdb.org/docs/sql/functions/utility.html)). Since Ibis includes support for `uuid` types, it might be nice to expose this as `ibis.uuid`.

*noted in a comment in https://github.com/ibis-project/ibis/issues/8256.*

jcrist:
Fixes #8257.

We implemented a bunch of these in a pairing session - there are still a few lingering backends that we should be able to support (but weren't added in this PR):

- `mssql`, `flink`, `pyspark`, `datafusion`, and `oracle` (through existing SQL functions)
- `pandas`, `dask`, maybe `polars` (through shelling out to a python udf)

### DIFF
--- a/ibis/backends/bigquery/compiler.py
+++ b/ibis/backends/bigquery/compiler.py
@@ -122,6 +122,7 @@ class BigQueryCompiler(SQLGlotCompiler):
         ops.Levenshtein: "edit_distance",
         ops.Modulus: "mod",
         ops.RandomScalar: "rand",
+        ops.RandomUUID: "generate_uuid",
         ops.RegexReplace: "regexp_replace",
         ops.RegexSearch: "regexp_contains",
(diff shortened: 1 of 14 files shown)


jcrist:
@jitingxu1 & @ncclementi - I pushed up what we did in the pairing session (plus a few ones after). I suggest we try to get this in and have y'all push up a follow-up adding the remaining backends.

cpcloud (approved):
LGTM!

PR #8445, opened 2024-02-25
perf(ir): avoid exponential growth on `name` attribute access

### LINKED ISSUE #8432: bug: regression: slow to compute name of deeply nested expressions
### What happened?

```python
import ibis

x = ibis.literal(1)
for i in range(100):
    x = x + 1
    # if I hardcode the name, this is instantaneous
    # x = x.name(f"step_{i}")
x
```

If I ctrl-c this after a few seconds, I see see eg

```
File ~/code/ibis/ibis/expr/operations/core.py:89, in Value.name(self)
     86 @property
     87 def name(self) -> str:
     88     names = (arg.name for arg in self.__args__ if hasattr(arg, "name"))
---> 89     return f"{self.__class__.__name__}({', '.join(names)})"

File ~/code/ibis/ibis/expr/operations/core.py:88, in <genexpr>(.0)
     86 @property
     87 def name(self) -> str:
---> 88     names = (arg.name for arg in self.__args__ if hasattr(arg, "name"))
     89     return f"{self.__class__.__name__}({', '.join(names)})"

    [... skipping similar frames: <genexpr> at line 88 (97 times), Value.name at line 89 (97 times)]

File ~/code/ibis/ibis/expr/operations/core.py:89, in Value.name(self)
     86 @property
     87 def name(self) -> str:
     88     names = (arg.name for arg in self.__args__ if hasattr(arg, "name"))
---> 89     return f"{self.__class__.__name__}({', '.join(names)})"

File ~/code/ibis/ibis/expr/operations/core.py:88, in <genexpr>(.0)
     86 @property
     87 def name(self) -> str:
---> 88     names = (arg.name for arg in self.__args__ if hasattr(arg, "name"))
     89     return f"{self.__class__.__name__}({', '.join(names)})"
```

main 09b6adaeecf1b6388866856c795022cfca4b2679

NA

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #8432. In short, the pattern we are using here for accessing `name` results in exponential growth of attribute accesses.

### DIFF
--- a/ibis/expr/operations/core.py
+++ b/ibis/expr/operations/core.py
@@ -85,7 +85,11 @@ class Value(Node, Coercible, DefaultTypeVars, Generic[T, S]):
     # TODO(kszucs): figure out how to represent not named arguments
     @property
     def name(self) -> str:
-        names = (arg.name for arg in self.__args__ if hasattr(arg, "name"))
+        names = (
+            name
+            for arg in self.__args__
+            if (name := getattr(arg, "name", None)) is not None
+        )
         return f"{self.__class__.__name__}({', '.join(names)})"
 
     @property
--- a/ibis/expr/tests/test_api.py
+++ b/ibis/expr/tests/test_api.py
@@ -147,3 +147,12 @@ def test_implicit_coercion_of_null_literal(op):
 
     assert expr1.op() == expected
     assert expr2.op() == expected
+
+
+def test_nested_name_property():
+    x = ibis.literal(1)
+    n = 100
+    for _ in range(n):  # noqa: F402
+        x = x + 1
+
+    assert x.op().name.count("Add") == n


kszucs (approved):

ibis/expr/operations/core.py
```diff
@@ -85,7 +85,11 @@ def __coerce__(
     # TODO(kszucs): figure out how to represent not named arguments
     @property
```
kszucs, line 86: I think we should turn this into `@attribute` in a follow-up since the `.name` property is accessed multiple times during both construction and compilation.

ibis/expr/tests/test_api.py
```diff
@@ -147,3 +147,12 @@ def test_implicit_coercion_of_null_literal(op):
...
     assert expr2.op() == expected
+
+
+def test_nested_name_property():
+    x = ibis.literal(1)
+    n = 100
+    for _ in range(n):  # noqa: F402
+        x = x + 1
+
+    assert x.op().name.count("Add") == n
```
NickCrews, line 158: @cpcloud do we want to add some way of checking for recursion explosion? Maybe use n=200 and then test this doesn't take longer than 3 seconds to compute?

NickCrews (approved):

ibis/expr/tests/test_api.py
cpcloud: It didn't finish in any reasonable amount of time, so just having it as a standard unit test seems fine.

NickCrews: ahh, I see, so you would just notice it during local dev, no automation needed. Sure that works.

PR #8448, opened 2024-02-26
fix(rewrites): change `TableColumn` -> `Field` in rewrites

gforsyth:
-------

## Description of changes

I don't know if just aren't using these rewrites or aren't testing them. 

Anyway, quick PR to remove the now renamed `ops.TableColumn` -- if it makes more sense to rip out the rewrites, I can do that instead.

### DIFF
--- a/ibis/expr/rewrites.py
+++ b/ibis/expr/rewrites.py
@@ -50,7 +50,7 @@ def rewrite_fillna(_):
 
     selections = []
     for name in _.parent.schema.names:
-        col = ops.TableColumn(_.parent, name)
+        col = ops.Field(_.parent, name)
         if (value := mapping.get(name)) is not None:
             col = ops.Alias(ops.Coalesce((col, value)), name)
         selections.append(col)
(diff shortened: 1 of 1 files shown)


kszucs:
Apparently we don't use `rewrite_dropna` and `rewrite_fillna` anymore. We should check whether there are backends not supporting `FillNa` and `DropNa` and if there are, we can keep these rewrites.

jcrist:
Yeah, these got dropped in the refactor. I _think_ we might want to bring them back though instead of implementing in the compiler (see https://github.com/ibis-project/ibis/blob/e4edc785a2e4324143c51c566458cf44af3be973/ibis/backends/sql/compiler.py#L1319-L1336) since doing this via a rewrite may allow fusing selects to generate nicer SQL. Doing this during compilation doesn't allow for that.

That said, happy to merge the fix as is (or delete the rewrites fully until/if we want them back).

gforsyth:
I can look at restoring the expr-level rewrites

gforsyth:
> SGTM, could you create a ticket to track this? I think we can merge it as is.

will do

kszucs (approved):

PR #8451, opened 2024-02-26
refactor(ir): remove `find_first_base_table()` analysis function

kszucs:
There were just a couple of places left where `find_first_base_table()` was still in use. Since the epic split refactor the preferred way to retrieve parent relation(s) is the `value.relations` property.

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -1808,32 +1808,20 @@ class Column(Value, _FixedTextJupyterMixin):
         Table
             A top-k expression
         """
+        from ibis.expr.types.relations import bind
 
-        from ibis.expr.analysis import find_first_base_table
+        try:
+            (table,) = self.op().relations
+        except ValueError:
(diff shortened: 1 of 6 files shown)


ibis/expr/types/generic.py
```diff
@@ -1808,32 +1808,20 @@ def topk(
...
         """
+        from ibis.expr.types.relations import bind
 
-        from ibis.expr.analysis import find_first_base_table
-
-        arg_table = find_first_base_table(self.op()).to_expr()
+        try:
+            (table,) = self.op().relations
+        except ValueError:
+            raise com.IbisTypeError("Unable to determine table for topk")
```
jcrist, line 1816: The other places where you re-raise a `IbisTypeError` you raise something like

```suggestion
            raise com.IbisTypeError("TopK must depend on exactly one table.")
```

jcrist (approved):
Overall LGTM

ibis/expr/types/generic.py
kszucs: Updated.

PR #8456, opened 2024-02-26
docs: fix typo in `order_by` docstring

peter-gy:
## Description of changes

Adjusted

> This means than shuffling a Table is super simple

to

>  This means that shuffling a Table is super simple


lostmygithubaccount (approved):
thanks!

PR #8458, opened 2024-02-26
refactor(ir): reduction to scalar subquery instead of a window function; remove frame.table; fix dereferencing of analytic function passed to .filter()

kszucs:


kszucs:
Closing in favor of smaller multiple smaller PRs.

PR #8465, opened 2024-02-27
refactor(api): remove now unsupported `max_lookback` window attribute

kszucs:
This feature was only supported by the pandas backend, but wasn\'t tested
properly and the support for it was entirely removed in the-epic-split refactor.

BREAKING CHANGE: `ibis.rows_with_max_lookback()` function and `ibis.window(max_lookback)` argument are removed


cpcloud (approved):

jcrist (approved):
:shipit:

Issue #8416, opened 2024-02-22
feat: make Deferred hashable


NickCrews:
### Is your feature request related to a problem?

I want to be able to use them in sets and as dict keys.

IDK if this will ruin some semantics of them, but if it is easy, then that would help me a lot. Thanks!

### Describe the solution you'd like

```python
hash(_.foo) == hash(_.foo)
hash(_.foo) != hash(_.foo.upper())
hash(_["bar"]) != hash(_.bar)
```

maybe its as simple as `def __hash__(self) -> str: return str(self)`???

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kszucs:
Closing due to the problems discussed in the linked PRs.

PR #8469, opened 2024-02-27
feat(common): make `Deferred` objects hashable

kszucs:
While implementing is straightforward, the semantics are not. The point of `Deferred` is to defer the evaluation of traditional python syntax such as the equality comparison `==`. Hence we need to implement `__eq__` to return another `Deferred` object which hasn't been allowed to be evaluated as a *truthy* value by raising from `__bool__`. 

When a hash collision occurs `__eq__` gets called and the result is evaluated a boolean. With the changes in this PR all deferred objects will evaluate as equal: `(_.a == _a) is True` as well as `bool(_.a != _.a)`. 
While this is tolerable for this esoteric use case, it violates the expected python behaviour so that if `a == b` then `hash(a) == hash(b)`.

Fixes [#8416](https://github.com/ibis-project/ibis/issues/8416)

### DIFF
--- a/ibis/common/tests/test_deferred.py
+++ b/ibis/common/tests/test_deferred.py
@@ -202,12 +202,6 @@ def test_deferred_supports_string_arguments():
     assert b.resolve({}) == "3.14"
 
 
-def test_deferred_object_are_not_hashable():
-    # since __eq__ is overloaded, Deferred objects are not hashable
-    with pytest.raises(TypeError, match="unhashable type"):
-        hash(_.a)
-
-
 def test_deferred_const():
     obj = const({"a": 1, "b": 2, "c": "gamma"})
 
@@ -501,14 +495,6 @@ def test_deferred_is_not_iterable(obj):
         next(obj)
 
 
-@pytest.mark.parametrize("obj", [_, _.a, _.a.b[0]])
-def test_deferred_is_not_truthy(obj):
-    with pytest.raises(
-        TypeError, match="The truth value of Deferred objects is not defined"
-    ):
-        bool(obj)
-
-
 def test_deferrable(table):
     @deferrable
     def f(a, b, c=3):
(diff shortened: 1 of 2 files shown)


ibis/common/tests/test_deferred.py
```diff
@@ -602,3 +583,25 @@ def test_deferred_namespace(table):
 def test_custom_deferred_repr(table):
     expr = _.x + table.a
     assert repr(expr) == "(_.x + <column[int]>)"
+
+
+def test_deferred_hashable():
+    assert bool(_)
+    assert bool(_.a)
+    assert bool(_.a == _.a)
```
cpcloud, line 591: Does this make the use of eq expressions as predicates confusing?

kszucs:
cc @NickCrews

ibis/common/tests/test_deferred.py
cpcloud: I think it will be confusing to support the eq expression as well as the default bool support. In some cases the == will result in an expression, but if used in any API calling bool, it will be converted to a bool. Can we continue to raise?

kszucs: > Can we continue to raise?

Nope, otherwise `dct[_.a]` is going to raise. I mainly put up the PR so we can discuss this in details :)

cpcloud: In that case I'm -0.5 on making deferred objects hashable.

Why should we take on the maintenance/documentation/explanatory burden here?

jcrist: I agree with Phillip that I'd be reluctant to lose the error on `__bool__`. One downside of not erroring on `__bool__` is we now have an object that compares equal to anything, which can lead to weird bugs. I once lost 2 hours because `np.dtype('f8') == None` is true (for reasons), I'd expect similarly tricky bugs to turn up if deferred objects effectively compare equal to anything.

I've pushed up an alternative solution in #8474. It may or may not solve the original ask, but I'd much prefer it to the solution here.

jcrist:
Given the conversation above I'm going to close this as we don't intend to merge it.

ibis/common/tests/test_deferred.py
kszucs: Cross-posting that the underlying `Resolver` object can be retrieved which is actually hashable, see https://github.com/ibis-project/ibis/pull/8474#issuecomment-1967389714 for details.

PR #8474, opened 2024-02-27
feat(api): make deferred objects hashable by their identity

### LINKED ISSUE #8416: feat: make Deferred hashable

jcrist:
Fixes #8416, an alternative to #8469.

In this case we hash `Deferred` objects as their identity. This means they can be used in sets or as dict keys, but only the exact instance will match in the hash lookup. Since `dict` and `set` check for object identity equality before dispatching to `__eq__` this means we can keep erroring nicely for `__bool__` (which I think we should) while still allowing `Deferred` objects to be hashable.

```python
In [1]: from ibis import _

In [2]: a = _.a

In [3]: d = {a: 1, _.b: 2}

In [4]: a in d
Out[4]: True

In [5]: _.a in d  # hash by identity only
Out[5]: False

In [6]: d[a]
Out[6]: 1
```

This implementation doesn't do weird things with the semantics of `__eq__`/`__hash__`/`__bool__` and should be innocuous to add. It's also what we _used to have_ before the refactor. The only downside is I'm not sure how useful it actually is. A few use cases this could help with:

- If we ever want to support a deferred API where the _key_ in a mapping may be deferred. Something like `ibis.map({_.x: 1})` for example.
- Deduplicating deferred expressions by identity (`list(set(exprs))`).
- Using a `dict` to associate metadata with a deferred expression `{deferred: some_metadata, other_deferred: other_metadata, ...}`.

### DIFF
--- a/ibis/common/deferred.py
+++ b/ibis/common/deferred.py
@@ -104,6 +104,9 @@ class Deferred(Slotted, Immutable, Final):
             f"The truth value of {self.__class__.__name__} objects is not defined"
         )
 
+    def __hash__(self):
+        return id(self)
+
     def __getitem__(self, name):
         return Deferred(Item(self, name))
 
--- a/ibis/common/tests/test_deferred.py
+++ b/ibis/common/tests/test_deferred.py
@@ -202,10 +202,21 @@ def test_deferred_supports_string_arguments():
     assert b.resolve({}) == "3.14"
 
 
-def test_deferred_object_are_not_hashable():
-    # since __eq__ is overloaded, Deferred objects are not hashable
-    with pytest.raises(TypeError, match="unhashable type"):
-        hash(_.a)
+def test_deferred_objects_are_hashable():
+    a = _.a
+    assert hash(a) == hash(a)
+
+    # Works in dicts
+    d = {a: 1, _.b: 2}
+    assert d[a] == 1
+    assert _.c not in d
+    assert _.a not in d  # hash by identity only
+
+    # Works in sets
+    s = {a, a, _.b}
+    assert len(s) == 2
+    assert a in s
+    assert _.a not in s  # hash by identity only
 
 
 def test_deferred_const():
@@ -539,11 +550,6 @@ def test_deferrable_repr():
         return x + 1
 
     assert repr(myfunc(_.a)) == "<test>"
-
-
-def test_deferred_set_raises():
-    with pytest.raises(TypeError, match="unhashable type"):
-        {_.a, _.b}  # noqa: B018
 
 
 @pytest.mark.parametrize(


jcrist:
cc @NickCrews - would this solve your need in #8416?

kszucs:
Also note that the underlying `Deferred._resolver` object is properly hashable and can be retrieved by `resolver(deferred_obj)` function, exposing that function could be another alternative.

NickCrews:
> cc @NickCrews - would this solve your need in https://github.com/ibis-project/ibis/issues/8416?

No it unfortunately wouldn't, with the use case I have the dict creation site is distant from the usage site, so the handle on the exact reference is lost in the mean time.

With this implementation I could see people putting `_.a` in a dict ("yay that worked, deferreds must be hashable!"), then in some different place trying to retrieve the value with a different `_.a`, then getting a KeyError, and being confused. I would rather have this users journey fail as soon as they try the insert and call __hash__().

Instead, the user the user can get this same behavior using `id()` directly. It is more explicit with its limitations.

Thanks for trying to solve this though, I really appreciate it. I don't really understand the `._resolver` implementation that @kszucs mentions, would love to understand the behavior of that implementation.

jcrist:
Given this doesn't solve your needs, I'll close this for now. Thanks for remaining positive even though we couldn't solve your problem in the way you hoped!

NickCrews:
thanks you two. It looks like `hash(resolver(_.a)) == hash(resolver(_.a))`, which means that I can do

```python
from ibis.common.deferred import resolver, _

d = {
    resolver(_.a): "foo",
    resolver(_.a + 1): "bar",
}
d[resolver(_.a)], d[resolver(_.a + 1)]
# ('foo', 'bar')
```

So personally this is workable for my use case, if a little unergonomic.

If we implemented `__hash__(self)` as `return hash(resolver(self))`, would that suffer from the same problems that @cpcloud mentioned in https://github.com/ibis-project/ibis/pull/8469#discussion_r1504183954?

kszucs:
#8469 uses the [same `__hash__` implementation](https://github.com/ibis-project/ibis/pull/8469/files#diff-b5738e427fbd7d473853ce6557a6e8887ddfbd16474aeb230a728464dfe91c80R205)

Issue #8442, opened 2024-02-25
bug: order_by() only works once


NickCrews:
### What happened?

```python
import ibis

t = ibis.memtable({"a": [2, 1, 3]})
ts = t.order_by(ibis.desc("a")).order_by("a")
ts.a.execute().tolist()
# [3, 2, 1]
```
I would expect the first order_by to be dropped. But the generated SQL makes me think the order_by rules are additive (which seems like nice UX so I can call order_by() in multiple steps), but we just don't properly deal with erasing clauses that should be overridden.

```python
ibis.to_sql(ts)
# SELECT
#   "t0"."a"
# FROM "ibis_pandas_memtable_hypojebj4jed7nes2bmsxletau" AS "t0"
# ORDER BY
#   "t0"."a" DESC,
#   "t0"."a" ASC
```

### What version of ibis are you using?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
Thanks!

This is a feature request, since this is intended behavior at the moment.

NickCrews:
Thanks for the fix! I'm worried this doesn't solve the problem in the general case though:

I'm now thinking of the case `t.order_by(ibis.desc(_.a)).order_by(_.a + 1)`. 

My desired behavior would be to order by `a + 1`, and for the `a` to be ignored. But with this new implementation, these two clauses are considered distinct, and so both are kept, and so we get back to `SELECT t.a FROM t ORDER BY t.a DESC, t.a + 1 ASC`, which is not what I desire.

I think this means that if we want to be able to support my desired behavior, then that would imply we need an algorithm to determine "which of the previously existing sortbys need to get dropped". This seems impossible in the general case.

I see two options (maybe you see more):

1. change semantics: chaining order_by()s is not additive. Every time you call it, you drop the old orderbys. Less featured, but less surprising.
2. keep semantics: chaining order_by()s is additive. If a user wants to completely override the sort order (what I want to do), then I need to construct a whole new relation, eg `t.order_by(ibis.desc(_.a)).view().order_by(_.a + 1)`. This actually gives [1,2,3], as I want. Probably should add a warning about this in the docs. More featured, but surprising in rare cases.

kszucs:
Nice catch! My guess is that we shouldn't fuse subsequent order operations into a single select since `ORDER BY a, a + 1` is also different than `FROM (... ORDER BY a) ORDER BY a + 1`.

PR #8475, opened 2024-02-27
fix(sql): overwrite the original sort key on successive order_by calls using the same key

kszucs:
After `the-epic-split` refactor we concatenated successive ordering keys (due to squashing selections) which can result in multiple order statements referring the same value expression. As turned out this can lead ambiguous behaviour based on the backend, so now we overwrite repeated order keys. 

For expression

```py
t.order_by(ibis.desc("a")).order_by("a")
```

Previously we compiled to:

```sql
SELECT t.a FROM t ORDER BY t.a DESC, t.a ASC
```

Now we compile to:

```sql
SELECT t.a FROM t ORDER BY t.a ASC
```

Fixes https://github.com/ibis-project/ibis/issues/8442


jcrist (approved):
Nice!

PR #8476, opened 2024-02-27
docs: update environment setup instructions

chloeh13q:
## Description of changes

Update environment setup instructions for `conda`.

### DIFF
--- a/docs/contribute/01_environment.qmd
+++ b/docs/contribute/01_environment.qmd
@@ -63,8 +63,8 @@ For a better development experience see the `conda` or `nix` setup instructions.
 ::: {.panel-tabset}
 
 ```{python}
-#| echo: false
-#| output: asis
+# | echo: false
+# | output: asis
 
 managers = {
     "conda": {
@@ -74,7 +74,7 @@ managers = {
     "mamba": {
         "name": "Mamba",
         "url": "https://github.com/mamba-org/mamba",
-    }
+    },
 }
 
 for manager, params in managers.items():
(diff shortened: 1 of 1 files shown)


docs/contribute/01_environment.qmd
```diff
@@ -101,16 +101,24 @@ for manager, params in managers.items():
...
     print("    cd ibis")
-    print(f"    {manager} env create --file environment.yml")
+    print(f"    {manager} env create --file conda/environment.yml")
     print("    ```")
     print()
+    if manager == "conda":
+        print()
+        print(
+            "    If you're using arm64 architecture (Mac M1/M2), use `conda/environment-arm64.yml` for setting up a dev environment for all the backends that are possible to install excluding Flink; use `conda/environment-arm64-flink.yml` for setting up a dev environment for all the backends that are possible to install including Flink. The reason to have two separate environments is because `apache-flink` forces `pyarrow` to downgrade to 11.0, which causes conflicts in other backends."
+        )
```
gforsyth, line 111: Can you wrap this string so it isn't on a single line?  
Either separate print statements or implicitly joined strings should work.

docs/contribute/01_environment.qmd
```diff
@@ -63,8 +63,8 @@ For a better development experience see the `conda` or `nix` setup instructions.
 ::: {.panel-tabset}
 
 ```{python}
-#| echo: false
-#| output: asis
+# | echo: false
+# | output: asis
```
gforsyth, line 67: I don't think there should be spaces between the `#` and the `|`

gforsyth (commented):
Yay more docs!  Just some formatting nits here.

gforsyth (approved):
:shipit:

PR #8478, opened 2024-02-27
feat(formats): add native support for polars `DataFrame` as an input and output type

### LINKED ISSUE #8453: `.to_polars()` method
It would be very convenient to convert a table or column directly to a Polars object using a simple `.to_polars()` method, just like we can currently do with `.to_pandas()`.

8.0.0

Polars

### Code of Conduct

jcrist:
This:

- Adds native support for converting ibis `DataType` and `Schema` objects to/from their `polars` equivalents.
- Adds a `to_polars` method on backends and expressions for executing a query and returning the equivalent native polars type.
- Adds native support for passing a `polars.DataFrame` object to `ibis.memtable`. In this case the data will be stored in the originating `polars.DataFrame` object (rather than transformed into pandas/pyarrow data), allowing for more efficient operation when mixing `polars.DataFrame` memtables with the `polars` backend.

While implementing this I found a few oddities in the `ibis.formats` api that I plan to cleanup in a follow-up refactor PR. For now I'd like the review here to focus on the polars implementation and test coverage rather than the internals.

Fixes #8453.


jcrist:
A quick demo:

```python

In [2]: t = ibis.examples.penguins.fetch()

In [3]: query = t.group_by("species").agg(_.bill_length_mm.mean())

In [4]: query
Out[4]: 
┏━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┓
┃ species   ┃ Mean(bill_length_mm) ┃
┡━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━┩
│ string    │ float64              │
├───────────┼──────────────────────┤
│ Adelie    │            38.791391 │
│ Chinstrap │            48.833824 │
│ Gentoo    │            47.504878 │
└───────────┴──────────────────────┘

In [5]: query.to_polars()  # compute and return a polars.DataFrame (note that this works with any backend, not just the polars backend)
Out[5]: 
shape: (3, 2)
┌───────────┬──────────────────────┐
│ species   ┆ Mean(bill_length_mm) │
│ ---       ┆ ---                  │
│ str       ┆ f64                  │
╞═══════════╪══════════════════════╡
│ Adelie    ┆ 38.791391            │
│ Chinstrap ┆ 48.833824            │
│ Gentoo    ┆ 47.504878            │
└───────────┴──────────────────────┘

In [6]: import polars as pl

In [7]: ratings = pl.DataFrame(  # some objective penguin rankings as an in-memory polars.DataFrame
   ...:     {
   ...:         "species": ["Adelie", "Chinstrap", "Gentoo"],
   ...:         "rating": ["good", "best", "better"],
   ...:     }
   ...: )

In [8]: query.join(ibis.memtable(ratings), "species")  # join directly on the in-memory data
Out[8]: 
┏━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┓
┃ species   ┃ Mean(bill_length_mm) ┃ rating ┃
┡━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━┩
│ string    │ float64              │ string │
├───────────┼──────────────────────┼────────┤
│ Adelie    │            38.791391 │ good   │
│ Chinstrap │            48.833824 │ best   │
│ Gentoo    │            47.504878 │ better │
└───────────┴──────────────────────┴────────┘
```

ibis/backends/__init__.py
```diff
@@ -223,6 +224,43 @@ def to_pyarrow(
...
+
+        Returns
+        -------
+        dataframe
+            A polars DataFrame holding the results of the executed expression.
+
+        """
+        import polars as pl
+
+        table = self.to_pyarrow(expr.as_table(), params=params, limit=limit, **kwargs)
```
kszucs, line 261: Is `expr.as_table()` required here?

jcrist: Ensuring the output is always a `pyarrow.Table` here makes massaging the results into the proper DataFrame/Series/Scalar values much more straightforward. It keeps that logic localized to the data mapper and not here.

ibis/expr/api.py
```diff
@@ -509,6 +511,31 @@ def _memtable_from_pyarrow_table(
     ).to_expr()
 
 
+@_memtable.register("polars.LazyFrame")
```
kszucs, line 514: It might be time (in a follow-up) to unify `ibis.table()` and `ibis.memtable()` for a better user experience, memtable become pretty stable and we could easily dispatch on known formats.

ibis/formats/polars.py
```diff
@@ -0,0 +1,176 @@
...
+    @classmethod
+    def from_ibis(cls, dtype: dt.DataType) -> pl.DataType:
+        """Convert an ibis type to a polars type."""
+        if dtype.is_decimal():
+            return pl.Decimal(
+                precision=dtype.precision,
+                scale=9 if dtype.scale is None else dtype.scale,
+            )
+        elif dtype.is_timestamp():
+            return pl.Datetime("ns", dtype.timezone)
```
kszucs, line 83: Since the ibis timestamps have a scale we can do more accurate conversions here using `dtype.unit.short`. I assume this was the previous conversion logic from the backend implementation. Since it requires additional testing it could you please create a follow-up ticket so we improve this later?

ibis/formats/tests/test_polars.py
```diff
@@ -0,0 +1,143 @@
...
+                ]
+            ),
+            id="struct",
+        ),
+    ],
+)
+def test_to_from_ibis_type(ibis_dtype, polars_type):
+    assert PolarsType.from_ibis(ibis_dtype) == polars_type
+    assert PolarsType.to_ibis(polars_type) == ibis_dtype
+    assert PolarsType.to_ibis(polars_type, nullable=False) == ibis_dtype(nullable=False)
```
kszucs, line 60: I would like to eventually phase out the `__call__` support for dtypes, so I would prefer `ibis_dtype.copy(nullable=False)` here.

kszucs (approved):
Nice! Easy review, thanks @jcrist!

ibis/expr/api.py
jcrist: I'm not sure if that will be possible (at least without some API breakage). For example - right now `ibis.table` can take a schema as the first positional argument, while `ibis.memtable` can take anything convertable to a `DataFrame`:

```python
In [1]: import ibis

In [2]: ibis.table([("x", "int"), ("y", "int")])  # a table with columns x & y
Out[2]: 
UnboundTable: unbound_table_0
  x int64
  y int64

In [3]: ibis.memtable([("x", "int"), ("y", "int")])  # a memtable with these rows
Out[3]: 
InMemoryTable
  data:
    PandasDataFrameProxy:
        col0 col1
      0    x  int
      1    y  int
```

Since the _types_ of the input are identical (but ibis interprets them differently in each context), I'm not sure if we could combine these functions in an intuitive way. Also personally I don't find the current separation between `table` and `memtable` to be a big API pain point.

PR #8481, opened 2024-02-27
fix(sql): don't fuse subsequent order by operations

kszucs:


kszucs:
Opening from another branch.

PR #8483, opened 2024-02-27
fix(sql): don't fuse subsequent order by operations

kszucs:
Previously we wrongly fused order by operations which is disabled now.


kszucs:
Closing in favor of https://github.com/ibis-project/ibis/pull/8618

PR #8486, opened 2024-02-28
fix(duckdb): parenthesize argument to StructField operation to support field access on CASE statements

### LINKED ISSUE #8485: bug(duckdb): struct field access on CASE generates SQL that duckdb cannot parse
### What happened?

OK, I'm really getting into the weeds here, this is a weird little case 😄 

```python
import ibis

s = ibis.struct({"a": 3})
x = ibis.case().when(s.a.isnull(), s).end()
y = x.a
print(ibis.to_sql(y))
# SELECT
#   CASE
#     WHEN (
#       {'a': CAST(3 AS TINYINT)}."a"
#     ) IS NULL
#     THEN {'a': CAST(3 AS TINYINT)}
#     ELSE CAST(NULL AS STRUCT(a TINYINT))
#   END."a" AS "a"
y.execute()  # Parser error due to the END."a"
```

main

duckdb

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #8485.

### DIFF
--- a/ibis/backends/tests/test_struct.py
+++ b/ibis/backends/tests/test_struct.py
@@ -133,3 +133,10 @@ def test_collect_into_struct(alltypes):
     val = result.val
     assert len(val.loc[result.group == "0"].iat[0]["key"]) == 730
     assert len(val.loc[result.group == "1"].iat[0]["key"]) == 730
+
+
+def test_field_access_after_case(con):
+    s = ibis.struct({"a": 3})
+    x = ibis.case().when(True, s).else_(ibis.struct({"a": 4})).end()
+    y = x.a
+    assert con.to_pandas(y) == 3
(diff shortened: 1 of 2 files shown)


ibis/backends/tests/test_struct.py
```diff
@@ -133,3 +134,13 @@ def test_collect_into_struct(alltypes):
     val = result.val
     assert len(val.loc[result.group == "0"].iat[0]["key"]) == 730
     assert len(val.loc[result.group == "1"].iat[0]["key"]) == 730
+
+
+@pytest.mark.notimpl(
+    ["postgres"], reason="struct literals not implemented", raises=PsycoPg2SyntaxError
```
gforsyth, line 140: ahh, probably need to add `risingwave` here, too

Issue #8450, opened 2024-02-26
feat: support empty URL for local backends in ibis.connect()


lostmygithubaccount:
### Is your feature request related to a problem?

often when testing against multiple backends, myself or a user will do something like:

```python

backend = "polars"

if backend == "polars":
    con = ibis.polars.connect()
...
```

Ibis supports a cleaner `ibis.connect()` that takes a backend-specific URI. those don't make sense for many of the local backends and are currently NotImplemented, but it would be nice for consistency and ease of use to support empty URIs for all of:

- pandas
- polars
- dask
- datafusion

and any others not supported now. should also support #8435 and #7710 once those are in (and any other future local backends)

### Describe the solution you'd like

```python

backend = "polars" # "duckdb" "pandas" "datafusion" "dask"

import ibis

con = ibis.connect(f"{backend}://")
```

adding anything to the URIs could warn or error 

### What version of ibis are you running?

main

### What backend(s) are you using, if any?

duckdb/pandas/datafusion/polars/dask/pyspark

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8490, opened 2024-02-28
docs: rework why Ibis article to explain what Ibis is and other updates

### LINKED ISSUE #8251: docs: what is Ibis?
### Please describe the issue

we should have a conceptual document clearly explaining what Ibis is

---

from [Will on Zulip](https://ibis-project.zulipchat.com/#narrow/stream/405263-general/topic/What.20is.20Ibis.3F/near/420136407):

> Where can you find a clear guide that states what Ibis is? It is very unclear to me what it can and can't do. The documentation tells you how to install it and run a few examples, but I see something like .filter(_.columnname == 3) and I have no idea what is supposed to be going on.

> In particular:

> 1. Can you use Ibis to create tables in e.g. a snowflake warehouse?
> 2. Does Ibis work for data workloads that are larger than memory?
> 3. How does it actually work? I picture it converting things into SQL, but I'm not clear at what point in evaluates things.
> 4. E.g., if I've built a table and I'm looking at the first 10 values, and then I add a new column, will it recalculate everything, or only the new column?

I answered:

> hi! we have a "Why Ibis" article here: https://ibis-project.org/why

> I had drafted a "What is Ibis" article at one point. Ibis is a Python dataframe library that decouples the API from the execution engine. Most Python dataframes (pandas, Polars, PySpark, Snowpark, etc.) tightly couple these -- resulting in slight differences in API and a lot of overhead in converting between them. Ibis instead uses an intermediary representation of its API to convert to backend-native code. For most backends, this is ends up being converted to SQL.

> Can you use Ibis to create tables in e.g. a snowflake warehouse?
Yes -- Ibis has a create_table method that, when used with the Snowflake backend, can be used to create tables in Snowflake (or other backends).

> Does Ibis work for data workloads that are larger than memory?
Yes -- Ibis will perform as well as the backend allows. For DuckDB and Polars, they generally support larger-than-memory operation.

> How does it actually work? I picture it converting things into SQL, but I'm not clear at what point in evaluates things.
Ibis expressions are compiled into some IR and then converted into backend-native code, usually SQL. There's some explanation here but it's a bit out of date: https://ibis-project.org/concepts/internals

> E.g., if I've built a table and I'm looking at the first 10 values, and then I add a new column, will it recalculate everything, or only the new column?
Generally only the new column, though I guess it could depend if you're using a view instead of a table. Ibis is lazily evaluated. You can use .cache() functions or turn on interactive mode

### Code of Conduct

### LINKED ISSUE #8488: docs: add RisingWave to the installs tab
### Please describe the issue

in `docs/_tabsets/instal.qmd`, add the RisingWave backend

I took a stab at this and could not get it to regenerate while rewriting another document, opening this for tracking (and if anyone has a tip -- simply adding a new line for it in the backends list of dictionaries didn't work w/ a `just docs-preview`)

### Code of Conduct

lostmygithubaccount:
## Description of changes

- add link to the Why VoDa blog post from who concept article
- makes some formatting edits in the who concept article per guidelines
- major rework of Why Ibis concept article, including what Ibis is
- add RisingWave to install tabset

## Issues closed

closes #8251
closes #8488


docs/why.qmd
```diff
@@ -2,24 +2,313 @@
...
+**Data engineers** can use Ibis to:
+
+- write and maintain complex ETL/ELT jobs
+- replace fragile SQL string pipelines with a robust Python API
+- replace PySpark with a more Pythonic API that supports Spark and many other
+ backends
+
+**Data analysts** can use Ibis to:
+
+- use Ibis interactive mode for rapid exploration
```
gforsyth, line 113: maybe we can reword this?  it currently reads as "... can use Ibis to use Ibis interactive mode... "

how about
```suggestion
- perform rapid exploratory data analysis using interactive mode
```

docs/why.qmd
```diff
@@ -2,24 +2,313 @@
...
+Often, data platforms evolve to support Python in some sequence like:
 
-## Portability
+1. Develop a fast query engine with a SQL frontend
+2. Gain popularity and need to support Python for data science and ML use cases
+3. Develop a bespoke pandas or PySpark-like dataframe library and ML
+ integrations
 
-You can reuse the same code across different backends.
+This third step is where Ibis comes in. Instead of spending X money and Y time
```
gforsyth, line 138: Were `X` and `Y` left in intentionally?  Maybe swap in for a generic measure?  "...instead of spending eyewatering amounts of money and time..."

docs/why.qmd
```diff
@@ -2,24 +2,313 @@
...
+Ibis takes inspiration from pandas and PySpark -- and R and SQL -- but is designed to be
+scalable from the start. If offers a neutral,
+[self-governed](https://github.com/ibis-project/governance) open source option
+for your data platform.
+:::
+
+### Ibis for data organizations
+
+Data organizations can use Ibis to standardize the interface for SQL and Python
+data practitioners. It also allows them to:
```
gforsyth, line 175: The `them` here is unclear -- is it referring to the data practitioners or the data organizations?

lostmygithubaccount:
also:

- want to update the connections to `ibis.connect("backend://")` for the simple example once #8450 is in (perhaps w/ a note on version needed)
- perhaps hone in on the why not pandas/PySpark API and also frame it as Ibis defines a bunch of operations, backends implement a subset, you can steer coverage for your backend; join Ibis as an open source project to help steer it as well; not playing catch up/trying to support new APIs as you would be w/ pandas/PySpark; might we worth mentioning Polars there too

docs/why.qmd
```diff
@@ -2,24 +2,313 @@
...
+        "PostgreSQL",
+        "PySpark",
+        "RisingWave",
+        "Snowflake",
+        "SQLite",
+        "Trino",
+    ],
+    "Expression-generating": ["Dask", "Polars"],
+    "Naïve execution": ["pandas"],
+}
```
gforsyth, line 234: For a script this long, maybe better served by pulling it out into a separate `.py` file and then importing and executing it in the qmd file?

As an aside, I wonder if we should add a category attribute (or something) to the backends that's one of "sql", "expression" or "naive".  Might be overkill, but we'll get tired of maintaining this list at some point.  (That can be for a followup, but I doubt this is the only place we'll find ourselves wanting to categorize the backends in)

docs/why.qmd
```diff
@@ -2,24 +2,313 @@
...
+fig.update_layout(
+    title_text="Ibis backend types", font_size=14, margin=dict(l=30, r=30, t=80, b=30)
+)
+
+fig.show()
+```
+
+As you can see, most backends generate SQL. Ibis uses
+[SQLGlot](https://github.com/tobymao/sqlglot) to transform Ibis expressions into
+SQL strings. You can also use the `.sql()` methods to mix in SQL strings,
```
gforsyth, line 310: Link `.sql()` to https://ibis-project.org/how-to/extending/sql#table.sql ?

docs/why.qmd
```diff
@@ -28,43 +317,59 @@ and robust framework for data manipulation in Python.
...
 
-- data catalog exploration
-- exploratory data analysis
-- transforming data
-- visualizing data
-- data science and machine learning
+- [scikit-learn](https://github.com/scikit-learn/scikit-learn)
+- [XGBoost](https://github.com/dmlc/xgboost)
+- [LightGBM](https://github.com/microsoft/lightgbm)
+- [PyTorch](https://github.com/pytorch/pytorch)
```
gforsyth, line 372: Broader documentation comment -- for everything like this that we advertise, I think we should aim to have short how-to docs on using X with Ibis.

gforsyth (commented):
This is really great, @lostmygithubaccount -- that tabset code sample at the top is especially clean.

Some small nits and some suggestions for (all of us for) the future, but I like the categorization and presentation here.

docs/why.qmd
lostmygithubaccount: it was but rewording for clarity to "spending a lot of time and money"

docs/why.qmd
lostmygithubaccount: how do I execute it after importing? just something like:

````
```{python}
from whatever_script_thing import whatever

whatever()
```
````

I guess that would do it

Comment (gforsyth) [reply docs/why.qmd]: yes -- you might have to wrap up the `fig`-producing code in a function and have it return `fig`, then you can do
```
from whatever import whatever

fig = whatever()
fig
```

Comment (lostmygithubaccount) [reply docs/why.qmd]: I just did `from backend_sankey import fig` and `fig.show()`

File: docs/why.qmd [hunk 7]
```diff
@@ -2,24 +2,224 @@
 title: "Why Ibis?"
 ---
 
+Ibis defines a Python dataframe API that supports any query engine -- the
+frontend for any backend data platform. This includes 20+ backends today, most
+SQL-based.  This allows Ibis to have excellent performance -- as good as the
+backend it is connected to -- with a consistent user experience.
+
```
amoeba, line 9: Would this page be a good place for the harder Ibis sell? Something like this:

What if you didn't have to learn yet another query language to try out a new database? What if moving from one analytical system to another didn't require you to rewrite all of your existing pipelines? What if you could just use the familiar, pandas-like DataFrame API you already use every day to do all of this and more?

Ibis solves these problem today for more than 20 popular database backends and the list will continue to grow. On the fly, Ibis translates your code to the query language of the backend you choose which means you get exactly the performance your database advertises without sacrificing the ability make another choice of backend in the future.

docs/why.qmd
```diff
@@ -2,24 +2,224 @@
...
+## What is Ibis?
+
 Ibis is the portable Python dataframe library.
```
amoeba, line 12:
```suggestion
Ibis is the portable Python dataframe library meaning it's write-once, run anywhere.
```

docs/why.qmd
```diff
@@ -2,24 +2,224 @@
...
 
-If you've had issues with scaling data transformation code in Python, need to
-work with data in multiple data platforms, find yourself translating between
-other Python dataframe APIs, or just want a great Python dataframe experience,
-Ibis is for you.
+We can demonstrate this with a simple example on a few local query engines:
```
amoeba, line 14:
```suggestion
We can demonstrate this by running the same code against different backends and showing we get the same result.
```

docs/concepts/who.qmd
```diff
@@ -12,16 +12,30 @@ most of the core development team employed there. As of writing, this includes
 five full-time developers, one technical product manager, and other staff who
```
zhenzhongxu, line 12: we have more than five now? :)

docs/why.qmd
```diff
@@ -2,24 +2,224 @@
 title: "Why Ibis?"
 ---
 
+Ibis defines a Python dataframe API that supports any query engine -- the
```
zhenzhongxu, line 5: nitpick: let's use a more precise word than "support" for people to resonate. maybe "connect," "execute," or "query."

docs/why.qmd
```diff
@@ -2,24 +2,224 @@
...
+frontend for any backend data platform. This includes 20+ backends today, most
```
zhenzhongxu, line 6: There are two main values that can be attractive to different audiences (1. people who care about Python df API and 2. people who care about connecting to various backends). If we articulate the two values independently that may help people to easily understand. 
(also wondering if there is a way we can capture reader statistics to understand who our audience is and what specific value they care more)?

zhenzhongxu, line 6: nit - more precise words:
frontend -> "language frontend" 
backend data platform -> data processing engine
remove "most SQL-based" - because it does not add value to this statement.
excellent performance -> flexibility or adaptability into a composable data architecture

docs/why.qmd
zhenzhongxu, line 12: My suggestion is to describe Ibis as a combination of two independent values and expand into concrete reasons why people should use Ibis over similar tools.
1. Lightweight Python dataframe for all your DS/analytics workload - we need examples to demonstrate why the API is more elegant, usable, performant than its competitors Pandas, Polars, PySpark, etc
2.Ability to for delayed execution on any modern compute engines - coupled with use case examples, maybe we should also get more (future) deep technical content to describe why Ibis is better.

docs/why.qmd
```diff
@@ -2,24 +2,224 @@
...
+Ibis is for data engineers, data analysts, and data scientists to use directly
+with their data platform(s) of choice. It also has benefits for [data
+platforms](#ibis-for-data-platforms), [data
+organizations](#ibis-for-data-organizations), and [data library
+developers](#ibis-for-data-library-developers).
+
+### Ibis for data practitioners
 
-## Portability
+You can use Ibis at any stage of your data workflow, no matter your role.
```
zhenzhongxu, line 102: The content here is high-level and probably appropriate for this blog post, but I believe we can be more convincing if we have production case studies with concrete problems solved with Ibis as a component. Maybe we can interview from existing users?

docs/why.qmd
```diff
@@ -2,24 +2,224 @@
...
+- work in a general-purpose, yet easy to learn, programming language without the
+ need for formatting SQL strings
+
+**Data scientists** can use Ibis to:
+
+- extract a sample of data for local iteration with a fast local backend
+- prototype with the same API that will be used in production
+- preprocess and feature engineer data before training a machine learning model
+
+### Ibis for data platforms
```
zhenzhongxu, line 125: i personally think data platforms/engines are integration partners, not users. For that reason, I almost feel like we should have a separate blog to focus on the whys for partners, to keep this one focused and avoid any potential confusion to users.

docs/why.qmd
```diff
@@ -2,24 +2,224 @@
...
 Most Python dataframes are tightly coupled to their execution engine. And many
 databases only support SQL, with no Python API. Ibis solves this problem by
 providing a common API for data manipulation in Python, and compiling that API
 into the backend's native language. This means you can learn a single API and
 use it across any supported backend (execution engine).
 
-![](./portability.png)
+Ibis supports three types of backend:
+
+1. SQL-generating backends
```
zhenzhongxu, line 207: these types are very technical, I feel integrating partners care more about these differences than the end users? I personally think if we break down the target audience more, then provide them with very specific targeted information, it'll make some difficult concepts easily understood by the public.

docs/why.qmd
```diff
@@ -28,43 +228,58 @@ and robust framework for data manipulation in Python.
...
+distributed backend like PySpark, BigQuery, or Trino by simply changing your
+connection string.
 
-### Local experience
+## Stream-batch unification
 
-Out of the box, Ibis offers a great local experience for working with many file
-formats.
+As of [Ibis 8.0](./posts/ibis-version-8.0.0-release/index.qmd), the first stream
+processing backends have been added. Since these systems tend to support SQL, we
```
zhenzhongxu, line 246: We can describe why unification is possible with a single Python dataframe API. The concept is stream-table duality. https://www.confluent.io/blog/kafka-streams-tables-part-1-event-streaming/

zhenzhongxu (commented):
Great write up, left a few comments.

docs/why.qmd
lostmygithubaccount: we want to avoid this in part because of bad connotations (Write once, debug everywhere): https://en.wikipedia.org/wiki/Write_once,_run_anywhere

docs/why.qmd
lostmygithubaccount: reworded slightly but want to keep this fairly simple for a first-time reader

docs/why.qmd
lostmygithubaccount: yes! let's do this in a follow up and combine w/ Z's comment below -- I think it should be a collapsed callout block where we go into a more detailed "sales pitch" on Ibis

lostmygithubaccount: https://github.com/ibis-project/ibis/issues/8514

docs/why.qmd
lostmygithubaccount: follow up for this: https://github.com/ibis-project/ibis/issues/8514

docs/why.qmd
lostmygithubaccount: follow up for this: https://github.com/ibis-project/ibis/issues/8514

docs/why.qmd
lostmygithubaccount: agreed! we do have a user testimonials page we should continue adding to

docs/why.qmd
lostmygithubaccount: I think we can let them see that for themselves :)

docs/concepts/who.qmd
lostmygithubaccount: updated!

docs/why.qmd
lostmygithubaccount: I'll leave this as a follow up blog -- "Ibis for data platforms" -- I think this still makes sense here

docs/why.qmd
lostmygithubaccount: I think end users still tend to wonder "how it works" and may call it overly ambitious as a product -- this + the graph/explanation below helps build confidence that it is primarily an API -> SQL translation and we can do that well across backends

cpcloud (approved):

docs/why.qmd
amoeba: Sounds great, thanks

docs/why.qmd
amoeba: nice catch, thanks

PR #8505, opened 2024-02-29
chore: add to filterwarnings a pandas 2.2.x deprecation warning

ncclementi:
I was running some tests `backends/flink/tests` locally, and I noticed this being raised by pandas 2.2.1

```
DeprecationWarning: Passing a BlockManager to DataFrame is deprecated and will raise in a future version. Use public APIs instead:
```

I'm not sure what we need to change on our end, but for now I think we need to filter it. For reference, it seems this comes from arrow. see https://github.com/pandas-dev/pandas/issues/56171


cpcloud:
2.2 breaks many many things. I would really avoid updating to that version if you can.

ncclementi:
Ok, then we should pin the envs yaml files. I'll close this one and get that in then.

PR #8506, opened 2024-02-29
chore: pin pandas <2.2 on conda envs

ncclementi:
Looks like pandas 2.2 breaks many things, so setting the dependency for <2.2 for now.


jcrist (approved):
Thanks, LGTM! Can you open a follow-up issue to track this so we can eventually resolve these?

ncclementi:
xref: https://github.com/ibis-project/ibis/issues/8508

PR #8511, opened 2024-03-01
feat(flink): implement support for array expansion

### LINKED ISSUE #8457: feat(flink): implement array expansion for Flink backend

[Array expansion](https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/dev/table/sql/queries/joins/#array-expansion) is not currently supported in the Flink backend, and this is a very important and commonly used functionality.

Support array expansion in the Flink backend

8.0

Flink

### Code of Conduct

chloeh13q:
## Description of changes

Implements support for array expansion.

## Issues closed

Resolves #8457


ibis/backends/sql/dialects.py
```diff
@@ -108,13 +109,86 @@ def _interval_with_precision(self, e):
     return f"INTERVAL {formatted_arg} {unit}"
 
 
+def _explode_to_unnest():
+    """Convert explode into unnest.
+
+    NOTE: Flink doesn't support UNNEST WITH ORDINALITY or UNNEST WITH OFFSET.
```
cpcloud, line 115: Is this the reason you copied the implementation over? If so, I don't think that's necessary. Ibis doesn't have an API for this yet, and even if we did, there's no reason to copy the implementation over when we can let the error bubble up from Flink.

chloeh13q: No it's not just this. Among other things that Flink doesn't support the one that is breaking is that Flink doesn't have `GenerateSeries`, so if we were to call the original function the whole thing errors out.

chloeh13q: Also, `transforms.explode_to_unnest()` takes in a parameter `index_offset` and it doesn't look like `index_offset` can be `None`, which means that the compiled query always includes `WITH OFFSET` and will never execute on the Flink backend

ibis/backends/tests/test_array.py
```diff
@@ -803,10 +803,13 @@ def test_array_intersect(con, data):
 )
 @pytest.mark.notimpl(["postgres"], raises=PsycoPg2SyntaxError)
 @pytest.mark.notimpl(["risingwave"], raises=PsycoPg2InternalError)
-@pytest.mark.notimpl(["datafusion", "flink"], raises=com.OperationNotDefinedError)
+@pytest.mark.notimpl(["datafusion"], raises=com.OperationNotDefinedError)
 @pytest.mark.broken(
     ["trino"], reason="inserting maps into structs doesn't work", raises=TrinoUserError
 )
+@pytest.mark.broken(
+    ["flink"], raises=Py4JJavaError, reason="error on memtable compilation"
```
cpcloud, line 811: It seems like we should avoid introducing a `broken` marker here by fixing the memtable compilation in this PR. Otherwise, we're potentially masking a bug in the unnest code generation.

chloeh13q: I tried to fix memtable compilation. Separate from the broken memtable compilation the three tests that I marked as broken also do not work because Flink doesn't have support for them (specifically, Flink doesn't support creating empty array literals and creating named structs with a single field; I found an open JIRA ticket for the former and submitted a JIRA ticket for the latter). I updated the failure reason for these three tests and will open another PR for memtable compilation.

chloeh13q:
Will rebase to main after we merge #8751 to resolve a couple of tests marked as broken.

chloeh13q:
@cpcloud Hey Philip - I merged #8751 that fixed the compilation of memtables with nested data. I then rebased and updated the broken tests. Would you mind taking another look at the PR? Thanks

cpcloud (approved):
This seems good to go for now, we'll have to keep an eye on the complexity of the impl here.

PR #8513, opened 2024-03-01
docs(how-to): add graphviz section

cpcloud:
Add graphviz `how-to`.

### DIFF
--- a/docs/how-to/visualization/graphs.qmd
+++ b/docs/how-to/visualization/graphs.qmd
@@ -1,0 +1,84 @@
+# GraphViz + Ibis
+
+Ibis supports visualizing an expression as a [directed
+graph](https://en.wikipedia.org/wiki/Directed_graph) using
+[GraphViz](https://graphviz.org/).
+
+To get started, make sure you've installed the necessary dependencies.
+
+```sh
+$ pip install 'ibis-framework[duckdb,examples,graphviz]'
+```
+
+::: {.callout-note collapse="true"}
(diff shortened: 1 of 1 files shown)


cpcloud:
I think this will be harder to find in Analytics. I don't think we can anticipate all reasons why someone might want to visualize an Ibis expression as a graph.

docs/how-to/visualization/graphs.qmd
```diff
@@ -0,0 +1,84 @@
...
+```{python}
+t = ibis.examples.penguins.fetch()
+expr = t.select(lowered=_.species.lower())
+to_graph(expr)
+```
+
+The way to read the graph is from top to bottom.
+
+* The top of the graph contains Ibis operations with no dependents.
+* The bottom of the graph contains Ibis operations with no dependencies.
```
lostmygithubaccount, line 45: this reads slightly off to me -- "top to bottom", then bullet points for top and bottom followed by the things in the middle -- in my mind would be top, then middle, then bottom, but not a big deal IMO

lostmygithubaccount (approved):
awesome! one minor comment

docs/how-to/visualization/graphs.qmd
cpcloud: Yeah, good point. I can rearrange.

cpcloud: Fixed!

Issue #8203, opened 2024-02-02
feat(api): Support time travel query


mfatihaktas:
### Is your feature request related to a problem?

Flink supports specifying a point in time and querying the corresponding data in a given table, aka `time travel`. As a use case, time travel would be useful when the results are desired/required to be reproducible. To support such use cases in Ibis/Flink, we would need to add time travel support for the Flink backend.

### Describe the solution you'd like

The API for time travel would look like the following
```python
table = con.create_table(
    ...
    watermark=ibis.watermark(...),
)

expr = table.at_time(<Timestamp expression>).select(...)
```

## Backends

Supporting time travel
* [DeltaLake](https://www.databricks.com/blog/2019/02/04/introducing-delta-time-travel-for-large-scale-data-lakes.html)
* Spark SQL
  * https://github.com/delta-io/delta/issues/128
  * https://github.com/delta-io/delta/pull/1288
  * https://issues.apache.org/jira/browse/SPARK-37219
  * https://iceberg.apache.org/docs/latest/spark-queries/#time-travel
* [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#for_system_time_as_of)
* [Flink](https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/dev/table/sql/queries/time-travel/)
* [MsSQL](https://learn.microsoft.com/en-us/sql/relational-databases/tables/querying-data-in-a-system-versioned-temporal-table?view=sql-server-ver16)
* [Trino](https://trino.io/docs/current/connector/iceberg.html?highlight=snapshot#using-snapshots)

## Notes
* Time travel in Flink is somewhat similar to [`asof` in pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.asof.html).
  * Flink time travel allows for accessing the snapshot of a table that corresponds to a given time expression.
  * Pandas `asof` allows for retrieving the last non-nan row within a dataframe that has a smaller value for a given <column, value> (e.g., the column might contain datetime strings).
[...]

PR #8517, opened 2024-03-01
feat(api): time travel query

mfatihaktas:
## Description of changes

Aims to address https://github.com/ibis-project/ibis/issues/8203.

This PR
- Adds `time_travel()` for `Table`.
- Adds `ibis.expr.types.temporal.TimeTravelTable` and `ops.TimeTravelDatabaseTable`.
- Adds `visit_TimeTravelDatabaseTable()` in the compiler for BigQuery and Flink, and `time_travel_table_sql()` in the SQL generator for Flink.
- Adds tests for time travel in BigQuery and Flink backends.

Notes:
- Could not add a functional test for BigQuery as I do not have access to a GCP env to run the functional tests.
- Will add more unit and functional tests.
  - Especially, for timestamp expressions and timestamps with time zones.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -139,6 +139,8 @@ services:
       AZURE_ABFS_OAUTH_SECRET: ""
       AZURE_ABFS_OAUTH_ENDPOINT: ""
       AZURE_WASB_ACCESS_KEY: ""
+    ports:
+      - 9083:9083
     depends_on:
       hive-metastore-db:
         condition: service_healthy
--- a/ibis/backends/flink/tests/test_time_travel.py
+++ b/ibis/backends/flink/tests/test_time_travel.py
@@ -1,0 +1,290 @@
+"""Tests in this module tests if
+(1) Ibis generates the correct SQL for time travel,
+(2) The generated SQL is executed by Flink without errors.
+They do NOT compare the time travel results against the expected results.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import pandas as pd
+import pytest
+
+import ibis
+import ibis.expr.datatypes as dt
+import ibis.expr.schema as sch
+import ibis.expr.types as ir
+from ibis.backends.flink.tests.utils import download_jar_for_package, get_catalogs
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+
+def create_temp_table(
+    table_name: str,
+    con,
+    data_dir: Path,
+    tempdir_sink_configs,
+    tmp_path_factory,
+):
+    # Subset of `functional_alltypes_schema`
+    schema = sch.Schema(
+        {
+            "id": dt.int32,
+            "bool_col": dt.bool,
+            "smallint_col": dt.int16,
+            "int_col": dt.int32,
+            "timestamp_col": dt.timestamp(scale=3),
+        }
+    )
+
+    df = pd.read_parquet(f"{data_dir}/parquet/functional_alltypes.parquet")
+    df = df[list(schema.names)]
+    df = df.head(20)
+
+    temp_path = tmp_path_factory.mktemp(table_name)
+    tbl_properties = tempdir_sink_configs(temp_path)
+
+    # Note: Paimon catalog supports 'warehouse'='file:...' only for temporary tables.
+    table = con.create_table(
+        table_name,
+        schema=schema,
+        tbl_properties=tbl_properties,
+        temp=True,
+    )
+    con.insert(
+        table_name,
+        obj=df,
+        schema=schema,
+    ).wait()
+
+    return table
+
+
+@pytest.fixture(scope="module")
+def temp_table(con, data_dir, tempdir_sink_configs, tmp_path_factory) -> ir.Table:
+    table_name = "test_table"
+
+    yield create_temp_table(
+        table_name=table_name,
+        con=con,
+        data_dir=data_dir,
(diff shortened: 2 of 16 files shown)


compose.yaml
```diff
@@ -139,6 +139,8 @@ services:
       AZURE_ABFS_OAUTH_SECRET: ""
       AZURE_ABFS_OAUTH_ENDPOINT: ""
       AZURE_WASB_ACCESS_KEY: ""
+    ports:
+      - 9083:9083
```
cpcloud, line 143: Why was this necessary? If you're looking to expose the hive metastore to the flink container, do that by adding this container to the flink network, not by exposing the port to the host.

ibis/backends/flink/tests/test_time_travel.py
```diff
@@ -0,0 +1,195 @@
...
+import ibis.expr.schema as sch
+from ibis.backends.conftest import TEST_TABLES
+from ibis.backends.flink.tests.utils import download_jar_for_package, get_catalogs
+from ibis.backends.tests.errors import Py4JJavaError
+
+
+@pytest.fixture
+def use_hive_catalog(con):
+    # Flink related
+    download_jar_for_package(
```
cpcloud, line 24: Can you do this in the flink dockerfile?

cpcloud (changes requested):

compose.yaml
mfatihaktas: Thanks. I actually first tried adding `networks: - flink` for `hive-metastore` but that led to
```
E   Caused by: java.net.ConnectException: Connection refused (Connection refused)
```

Digging deeper into this, I found that `TestConfForStreaming` (the `TestConf` used by the tests in `ibis/backends/flink/tests/`) creates a local Flink environment, which means the tests do not run on the dockerized environment. This is why we have to expose the meta-store to the local env.

I have put the env creation code in a new function `get_table_env(remote_env: bool, streaming_mode: bool)` to make it more transparent and configurable. Switched to remote env in `TestConfForStreaming` to see if we can run the tests without exposing the meta-store port, but that gave me 
```
Caused by: java.util.concurrent.ExecutionException: org.apache.flink.client.program.ProgramInvocationException: Job failed (JobID: 8065e44d3abe8da591f9c169cde351cd)
...
Caused by: java.io.IOException: Failed to create the parent directory: /private/var/folders/9g/7ncp4wjj22q25spxdy7t44080000gn/T/pytest-of-mehmet/pytest-520/test_table0
```
The error seems to be due to the Flink client being on a file system that is not shared with the Flink `TaskManager`. I guess this is why the env was left to be local in `TestConfForStreaming` at the time.

ibis/backends/flink/tests/test_time_travel.py
mfatihaktas: Adding JAR files in the container in this case won't help because the Flink specific tests do not run on the dockerized env, but on the local env as [described above](https://github.com/ibis-project/ibis/pull/8517#discussion_r1509780167).

ibis/expr/types/temporal.py
```diff
@@ -994,3 +994,23 @@ def full_name(self):
             The name of the day of the week
         """
         return ops.DayOfWeekName(self._expr).to_expr()
+
+
+@public
+class TimeTravelTable:
```
mfatihaktas, line 1000: Note: I think there is a way to eliminate the `__init__` and `__getattr__` from `TimeTravelTable`. Will look in to this tomorrow.

ibis/backends/flink/tests/test_time_travel.py
mfatihaktas: Downloading JAR's is failing on CI:
```
package_name = 'apache-flink'
jar_name = 'paimon-flink-1.18-0.8-20240301.002155-30'
jar_url = 'https://repository.apache.org/content/groups/snapshots/org/apache/paimon/paimon-flink-1.18/0.8-SNAPSHOT/paimon-flink-1.18-0.8-20240301.002155-30.jar'

    def download_jar_for_package(
        package_name: str,
        jar_name: str,
        jar_url: str,
    ):
        import os
        from importlib import metadata
    
        import requests
    
        # Find the path to package lib
        try:
            distribution = metadata.distribution(package_name)
            lib_path = distribution.locate_file("")
        except metadata.PackageNotFoundError:
            lib_path = None
    
        # Check if the JAR already exists
        jar_path = os.path.join(lib_path, "pyflink/lib", f"{jar_name}.jar")
        if os.path.exists(jar_path):
            return jar_path
    
        # Download the JAR
        response = requests.get(jar_url, stream=True)
        if response.status_code != 200:
>           raise SystemError(
                f"Failed to download the JAR file \n"
                f"\t jar_url= {jar_url} \n"
                f"\t response.status_code= {response.status_code}"
            )
E           SystemError: Failed to download the JAR file 
E           	 jar_url= https://repository.apache.org/content/groups/snapshots/org/apache/paimon/paimon-flink-1.18/0.8-SNAPSHOT/paimon-flink-1.18-0.8-20240301.002155-30.jar 
E           	 response.status_code= 404
```

The "easy" way to deal with this whole thing of downloading JAR's in Python would be skipping these tests by default and leaving them only for manual local testing.

cpcloud (changes requested):
Can we drop all the flink-related changes and just implement this for bigquery and snowflake (and perhaps one other backend)?

cpcloud:
Similar to #8412, I don't have time to maintain this code right now.

PR #8519, opened 2024-03-02
chore: add return `Table` typehint to `pivot_wider`

peter-gy:
## Description of changes

Adds `-> Table` to the signature of `pivot_wider` to ensure consistency with the other `Table` methods.

Each `Table` method with an indicated `Table` return in their docstring also has a corresponding `-> Table` in their signature, except `pivot_wider`, where no return typehint was present.


cpcloud (approved):
Thanks!

PR #8521, opened 2024-03-02
fix(graphviz): show proper field attributes of accessed relations and do not display join link property accesses

cpcloud:
This PR properly displays relation field names when accessed and avoids displaying references as `.None` when coming from a JoinLink object.


cpcloud:
Going to merge this.

PR #8510, opened 2024-02-29
feat: support custom node and edge attributes in `ibis.visualize`

peter-gy:
## Description of changes

Enables the customization of `graphviz` graphs produced by `ibis.visualize` by exposing `node_attr` and `edge_attr` as optional arguments. Changing node shapes, fonts and font sizes can be highly useful when dealing with complex expression graphs.

The default attributes (`DEFAULT_NODE_ATTRS`, `DEFAULT_EDGE_ATTRS`) are still considered, but the user-defined ones will take precedence.

PR #8527, opened 2024-03-03
feat(graphviz): node- and edge-specific custom attributes

peter-gy:
## Description of changes

Follow-up to #8510.

Adds support to optionally specifying custom Graphviz [node attributes](https://graphviz.org/docs/nodes/) and [edge attributes](https://graphviz.org/docs/edges/) based on arbitrary logic described by user-defined callbacks: `node_attr_callback` and `edge_attr_callback`.

With this, more complex visualization needs can be addressed, such as:

- I want to use a custom shape and fill color for all the `Field` nodes
- I want to have custom edge style and color between `Field` and `Comparison` nodes

Consider the example below.

```python
import ibis

left = ibis.table(dict(a="int64", b="string"), name="left")
right = ibis.table(dict(b="string", c="int64", d="string"), name="right")
expr = (
    left.inner_join(right, "b")
    .select(left.a, b=right.c, c=right.d)
    .mutate(
        arrays=ibis.array([1, 2, 3]),
    )
)

def get_node_attributes(node):
    if isinstance(node, ops.Literal):
        return {"shape": "hexagon", "style": "filled", "fillcolor": "yellow"}

def get_edge_attributes(u, v):
    if isinstance(u, ops.Field) and isinstance(v, ops.Comparison):
        return {"style": "dashed", "color": "red"}

    if isinstance(v, ops.Field):
        return {"arrowhead": "box", "color": "blue"}


expr.visualize(
    label_edges=True,
    node_attr={"color": "green", "fontname": "Roboto Mono"},
    node_attr_callback=get_node_attributes,
    edge_attr={"fontsize": "12", "fontname": "Comic Sans MS"},
    edge_attr_callback=get_edge_attributes,
)
```

custom-attrs


peter-gy:
An API design consideration we have here is whether we want to merge `node_attr` and `node_attr_callback` as well as `edge_attr` and `edge_attr_callback` to be of type `dict[str,str] | *AttrCallback | None` so that we don’t impose additional cognitive load on users with parameter variations.

What do you think?

cpcloud:
Pretty fancy graph there!

Keeping the separation between "default attributes for every node" and "a function that can be used to override specific nodes" seems like a good separation to me.

While I don't love the `_callback` naming, I'm not sure there's something that significantly better.

peter-gy:
> Pretty fancy graph there!
> 
> Keeping the separation between "default attributes for every node" and "a function that can be used to override specific nodes" seems like a good separation to me.
> 
> While I don't love the `_callback` naming, I'm not sure there's something that significantly better.

I agree with the need for the separation.

What about `_getter` as a suffix?

cpcloud:
`getter` SGTM!

cpcloud (approved):
Niiiiice!

PR #8529, opened 2024-03-03
fix(deps): support python 3.12

cpcloud:
Add 3.12 job and support in our nix builds and environment; fixes warnings about utcfromtimestamp from ibis code and ignores them in testing due to their use at the top level of dateutil.

### DIFF
--- a/.github/workflows/check-generated-files.yml
+++ b/.github/workflows/check-generated-files.yml
@@ -37,7 +37,7 @@ jobs:
         uses: actions/setup-python@v5
         id: install_python
         with:
-          python-version: "3.11"
+          python-version: "3.12"
           cache: pip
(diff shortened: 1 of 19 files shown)


cpcloud:
Looks like only DuckDB >=0.10 supports Python 3.12 :(

cpcloud:
I realize we have a 3-Python-version policy (I think we wrote that down somewhere), so I didn't add a 3.12 job anywhere that would significantly increase CI job count. For example, all backend workflows (except PySpark because in the past that has been annoying) did **not** get a 3.12 job.

gforsyth:
NEP 29 drop-date for Python 3.9 is April 5th, 2024, so either way we'll be back to 3 python versions fairly shortly.

gforsyth (approved):
:shipit:

PR #8531, opened 2024-03-03
feat(api): count nulls with topk

### LINKED ISSUE #8518: bug: topk() always says a column contains 0 NULLs
### What happened?

I would expect this to say that there is 1 NULL:
```python
import ibis

t = ibis.memtable({"x": [1, 1, 2, None]})
tk = t.x.topk(10)
tk
# ┏━━━━━━━━━┳━━━━━━━━━━┓
# ┃ x       ┃ Count(x) ┃
# ┡━━━━━━━━━╇━━━━━━━━━━┩
# │ float64 │ int64    │
# ├─────────┼──────────┤
# │     1.0 │        2 │
# │     2.0 │        1 │
# │    NULL │        0 │
# └─────────┴──────────┘
ibis.to_sql(tk)
# SELECT
#   "t1"."x",
#   "t1"."Count(x)"
# FROM (
#   SELECT
#     "t0"."x",
#     COUNT("t0"."x") AS "Count(x)"
#   FROM "ibis_pandas_memtable_46aexxpm3fdfhdexrstqhyuvyi" AS "t0"
#   GROUP BY
#     1
# ) AS "t1"
# ORDER BY
#   "t1"."Count(x)" DESC
# LIMIT 10
```

main

duckdb

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Default to count star instead of count(expr) in topk API. Closes #8518.

BREAKING CHANGE: `expr.topk(...)` now includes null counts. The row count of the `topk` call will not differ, but the number of nulls counted will no longer be zero. To drop the `null` row use the `dropna` method.

### DIFF
--- a/ci/schema/clickhouse.sql
+++ b/ci/schema/clickhouse.sql
@@ -94,3 +94,6 @@ INSERT INTO ibis_testing.win VALUES
     ('a', 2, 0),
     ('a', 3, 1),
     ('a', 4, 1);
+
+CREATE OR REPLACE TABLE ibis_testing.topk (x Nullable(Int64)) ENGINE = Memory;
+INSERT INTO ibis_testing.topk VALUES (1), (1), (NULL);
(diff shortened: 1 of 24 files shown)


NickCrews:
related, since this is breaking perhaps want to consider these both at the same time: https://github.com/ibis-project/ibis/issues/8540

cpcloud:
@NickCrews Making column names match should be done in a separate PR

kszucs (approved):
LGTM

PR #8537, opened 2024-03-04
feat(sql): support temporal joins [WIP]

kszucs:
Original PR: https://github.com/ibis-project/ibis/pull/8412


ibis/expr/tests/test_newrels.py
```diff
@@ -1567,3 +1568,31 @@ def test_subsequent_order_by_calls():
...
+    t1 = ibis.table(name="t1", schema={"a": "int64", "b": "string", "c": "timestamp"})
+    t2 = ibis.table(name="t2", schema={"d": "int64", "e": "string"})
+
+    joined = t1.join(t2.at_time(t1.c), [t1.a == t2.d])
+
+    right = t2.at_time(t1.c)
+    assert right.op() == ops.VersionedTable(t2, at_time=t1.c)
+
+    r1 = ops.JoinTable(t1, index=0).to_expr()
+    r2 = ops.JoinTable(ops.VersionedTable(t2, at_time=r1.c), index=1).to_expr()
```
kszucs, line 1583: An alternative implementation would be to have specialized join link (which might be considerable for asof joins as well), but we can defer that to a follow-up.

ibis/expr/operations/relations.py
```diff
@@ -161,6 +161,7 @@ def values(self):
     "any_inner",
     "any_left",
     "cross",
+    "temporal",
```
kszucs, line 164: We can encode left and inner variants using `temporal_inner` and `temporal_left`.

kszucs:
I won't have time to keep working on this in the near future, so closing it.

Issue #8507, opened 2024-02-29
bug(flink): Flink's docker setup does not work for arm64 (Apple M1/M2)


chloeh13q:
### What happened?

The `flink` docker container fails in the build process (i.e., when running `docker compose build flink`).

### What version of ibis are you using?

8.0

### What backend(s) are you using, if any?

Flink

### Relevant log output

```sh
92.06 Collecting pemja==0.3.0
92.16   Downloading pemja-0.3.0.tar.gz (48 kB)
92.46      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 48.5/48.5 KB 123.7 kB/s eta 0:00:00
92.65   Installing build dependencies: started
97.24   Installing build dependencies: finished with status 'done'
97.24   Getting requirements to build wheel: started
97.32   Getting requirements to build wheel: finished with status 'error'
97.32   error: subprocess-exited-with-error
97.32   
97.32   × Getting requirements to build wheel did not run successfully.
97.32   │ exit code: 255
97.32   ╰─> [1 lines of output]
97.32       Include folder should be at '/opt/java/openjdk/include' but doesn't exist. Please check you've installed the JDK properly.
97.32       [end of output]
97.32   
97.32   note: This error originates from a subprocess, and is likely not a problem with pip.
97.32 error: subprocess-exited-with-error
97.32 
97.32 × Getting requirements to build wheel did not run successfully.
97.32 │ exit code: 255
97.32 ╰─> See above for output.
97.32 
97.32 note: This error originates from a subprocess, and is likely not a problem with pip.
------
failed to solve: process "/bin/sh -c pip3 install apache-flink==1.18.1" did not complete successfully: exit code: 1
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chloeh13q:
I was able to resolve this by using the following Dockerfile
```
FROM flink:1.18.0

# install python3 and pip3
RUN apt-get update -y && \
apt-get install -y python3 python3-pip python3-dev openjdk-11-jdk-headless && rm -rf /var/lib/apt/lists/*
RUN ln -s /usr/bin/python3 /usr/bin/python

ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk-arm64

# install PyFlink
RUN pip3 install apache-flink==1.18.0
```
[...]

PR #8544, opened 2024-03-04
test(flink): enable turning on local test env

mfatihaktas:
## Description of changes

Aims to unblock people who cannot run the tests with Flink backend on the dockerized env as documented in https://github.com/ibis-project/ibis/issues/8507.

Allows running the tests with local env when env variable `IBIS_FLINK_LOCAL_ENV` is set to `true` or `1`.
E.g.
```
$ IBIS_FLINK_LOCAL_ENV=1 pytest -m flink ...

$ IBIS_FLINK_LOCAL_ENV="True" pytest -m flink ...

$ IBIS_FLINK_LOCAL_ENV=true pytest -m flink ...
```

### DIFF
--- a/ibis/backends/flink/tests/conftest.py
+++ b/ibis/backends/flink/tests/conftest.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import pandas as pd
 import pytest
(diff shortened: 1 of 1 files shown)


cpcloud:
I'd rather we wait until we can fix this without introducing external state like environment variables to make stuff work. Is there a particular reason we can't wait for the fix?

mfatihaktas:
> I'd rather we wait until we can fix this without introducing external state like environment variables to make stuff work. Is there a particular reason we can't wait for the fix?

Sure, I created this PR as the change is straightforward and quick. During today's triage meeting, we talked about people having issues time-to-time with setting up the Flink env on M1/M2. Even if we do not merge this PR, it can be used as a reference when needed by people who need the local Flink env to unblock themselves.

mfatihaktas:
@cpcloud Actually, even without the env variable, I think just the part with `get_table_env()` is worth to merge because it brings visibility into which env is local and which env is remote, and making it easy to change if/when needed.

kszucs:
I am in favor of this change. There are various environments where the docker setup may not be feasible or not working out of the box and with little code change  the developer can fall back to the local env.

ibis/backends/flink/tests/conftest.py
```diff
@@ -30,21 +42,43 @@ def connect(*, tmpdir, worker_id, **kw: Any):
...
+
+    @staticmethod
+    def connect(*, tmpdir, worker_id, **kw: Any):
+        """Flink backend is created in batch mode by default. This is to
+        comply with the assumption that the tests under ibis/ibis/backends/tests/
+        are for batch (storage or processing) backends.
+        """
+        import os
+
+        local_env = os.getenv("IBIS_FLINK_LOCAL_ENV", "False").lower() in ("true", "1")
```
cpcloud, line 79: When would we ever set this to `false`?

cpcloud:
Is this PR still necessary?

ibis/backends/flink/tests/conftest.py
mfatihaktas: I guess you are referring to the fact that `IBIS_FLINK_LOCAL_ENV` is binary, and we can simply interpret it being defined as the flag to run the tests locally. I chose to make it explicit with `true` and `false`, mostly due to personal preference over possibly having to `unset` env variables.

mfatihaktas:
> Is this PR still necessary?

It is not necessary to keep `IBIS_FLINK_LOCAL_ENV `. But I think there is value in keeping `get_table_env()` per my previous comment:

> @cpcloud Actually, even without the env variable, I think just the part with get_table_env() is worth to merge because it brings visibility into which env is local and which env is remote, and making it easy to change if/when needed.

ibis/backends/flink/tests/conftest.py
cpcloud: Sounds good, let's just delete the variable then :)

cpcloud (approved):
LGTM, modulo removal of the environment variable.

cpcloud:
@mfatihaktas Needs a rebase.

PR #8545, opened 2024-03-05
feat(datafusion, flink, mssql): add uuid operation

jitingxu1:
## Description of changes
Continued from [PR](https://github.com/ibis-project/ibis/pull/8438): 
add uuid operation for mssql, flink,  datafusion,


ibis/backends/oracle/compiler.py
```diff
@@ -227,6 +227,14 @@ def visit_RandomScalar(self, op):
         # Not using FuncGen here because of dotted function call
         return sg.func("dbms_random.value")
 
+    def visit_RandomUUID(self, op):
```
jitingxu1, line 230: @jcrist There is no builtin `uuid` in oracle, I used the `sys_guid` and reformat it into a uuid format, but not sure if it is ok. since it a not a real uuid, it fails the `version == 4` test.

ibis/backends/pyspark/compiler.py
```diff
@@ -70,6 +70,7 @@ class PySparkCompiler(SQLGlotCompiler):
         ops.Hash: "hash",
         ops.Log10: "log10",
         ops.LStrip: "ltrim",
+        ops.RandomUUID: "uuid",
```
jitingxu1, line 73: @jcrist  There was a bug in the previous version of Spark, it has been addressed in the latest 4.0 version.(spark jira [issue](https://issues.apache.org/jira/browse/SPARK-47104) and [pr](https://github.com/apache/spark/pull/45199)).

gforsyth: We don't currently support spark 4.0, so I don't think we should provide this mapping.

ibis/backends/oracle/compiler.py
gforsyth: If there's no `uuid`, then there's no `uuid` -- this can remain unimplemented for the Oracle backend unless we are very confident that the semantics match other implementations.

ibis/backends/pyspark/compiler.py
ncclementi: Could we open an issue documenting this so when we support spark 4.0 (not sure how realistic that is) then this could be a nice onboarding issue?

jitingxu1: I will remove pyspark uuid support for now and open an issue for future reference.

jitingxu1:
@gforsyth and @ncclementi I removed the oracle and pyspark, please take a review again. Thanks for your time.

gforsyth (commented):
Thanks @jitingxu1 !  The code here looks good.

Can you update the PR title to match what's implemented here?  Something like:

feat(datafusion, flink, mssql): add uuid operation

And then update the PR description to reflect the current code state?  That's what end up as the body of the commit on main.

gforsyth (approved):

Issue #8504, opened 2024-02-29
feat: support `Table.group_by(*keys)`


NickCrews:
### Is your feature request related to a problem?

same idea as https://github.com/ibis-project/ibis/pull/8472

### Describe the solution you'd like

see that issue

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8546, opened 2024-03-05
feat(api): support variadic arguments on `Table.group_by()`

NickCrews:
Fixes https://github.com/ibis-project/ibis/issues/8504

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -922,7 +922,7 @@ class Table(Expr, _FixedTextJupyterMixin):
 
     def group_by(
         self,
-        by: str | ir.Value | Iterable[str] | Iterable[ir.Value] | None = (),
+        *by: str | ir.Value | Iterable[str] | Iterable[ir.Value] | None,
         **key_exprs: str | ir.Value | Iterable[str] | Iterable[ir.Value],
     ) -> GroupedTable:
         """Create a grouped table expression.
@@ -978,9 +978,7 @@ class Table(Expr, _FixedTextJupyterMixin):
         """
         from ibis.expr.types.groupby import GroupedTable
 
-        if by is None:
-            by = ()
-
+        by = tuple(v for v in by if v is not None)
         groups = bind(self, (by, key_exprs))
         return GroupedTable(self, groups)
 
--- a/ibis/tests/expr/test_table.py
+++ b/ibis/tests/expr/test_table.py
@@ -799,6 +799,14 @@ def test_group_by_kwargs(table):
     assert_equal(expr, expected)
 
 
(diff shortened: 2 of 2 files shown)


ibis/expr/types/relations.py
```diff
@@ -978,9 +978,7 @@ def group_by(
         """
         from ibis.expr.types.groupby import GroupedTable
 
-        if by is None:
-            by = ()
-
+        by = tuple(v for v in by if v is not None)
```
cpcloud, line 981: Why wouldn't None be a valid grouping key?

NickCrews: There is an existing test in this same file:

```python
@pytest.mark.parametrize("group", [[], (), None])
def test_group_by_nothing(table, group):
    with pytest.raises(com.IbisInputError):
        table.group_by(group)
```

which is testing for the fairly help error message:

```
    def __init__(self, groupings, **kwargs):
        if not groupings:
>           raise com.IbisInputError("No group keys provided")
E           ibis.common.exceptions.IbisInputError: No group keys provided

ibis/expr/types/groupby.py:48: IbisInputError
```

without this little filter, this test fails with this diffferent, cryptic error:

```
<generator object bind at 0x13ad77de0>) has failed due to the following errors:
E             `groupings`: <generator object bind at 0x13ad77de0> is not a tuple of coercibles to a Value[+T, Columnar]
E           
E           Expected signature: GroupedTable(table: Relation, groupings: tuple[Value[+T, Columnar], ...], orderings: tuple[SortKey, ...] = (), havings: tuple[Value[Boolean, DataShape], ...] = ())
```

If I convert the result of `bind()` into a tuple, it becomes more clear the cause:

```
(None,)) has failed due to the following errors:
E             `groupings`: (None,) is not a tuple of coercibles to a Value[+T, Columnar]
E           
E           Expected signature: GroupedTable(table: Relation, groupings: tuple[Value[+T, Columnar], ...], orderings: tuple[SortKey, ...] = (), havings: tuple[Value[Boolean, DataShape], ...] = ())
```

So maybe we should change the implementation, but I think I want to keep the current public error behavior

NickCrews: this also maybe brings up a side issue that these other more internal errors could be tweaked to be more helpful, but that's separate

kszucs: I think `None` should be valid as well, and `bind()` will turn it to a null literal automatically.

cpcloud: In that case, someone can do a follow up to make the breaking change to `group_by(None)` behavior.

cpcloud (approved):
Sweet!

PR #8548, opened 2024-03-05
refactor(ir): make error messages more readable

kszucs:
Out error messages involving expressions (most of the cases) were hardly readable since the signature use the repr of the passed arguments. This PR changes the `expr.__repr__()` implementation to use the default python behaviour and keep the `expr.__str__()` implementation to return with the readable pretty print of the expression. Note that this doesn't affect the behaviour in ipython neither in interactive mode. 

I guess this change is going to be controversial, so I have time to add tests until we discuss it.

### DIFF
--- a/ibis/backends/tests/test_interactive.py
+++ b/ibis/backends/tests/test_interactive.py
@@ -35,7 +35,7 @@ def table(backend):
 
 @pytest.mark.notimpl(["dask", "pandas", "polars"])
 def test_interactive_execute_on_repr(table, queries, snapshot):
-    repr(table.bigint_col.sum())
+    str(table.bigint_col.sum())
     snapshot.assert_match(queries[0], "out.sql")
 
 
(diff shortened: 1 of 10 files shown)


ibis/backends/tests/test_interactive.py
```diff
@@ -74,26 +74,27 @@ def test_disable_query_limit(table, snapshot, queries):
...
+    str(table.string_col.topk(3))
 
 
-def test_histogram_repr_no_query_execute(table, queries):
-    tier = table.double_col.histogram(10).name("bucket")
-    expr = table.group_by(tier).size()
-    expr._repr()
+# FIXME(kszucs): there seems to be a bug with the histogram query
+# def test_histogram_repr_no_query_execute(table, queries):
+#     tier = table.double_col.histogram(10).name("bucket")
```
kszucs, line 89: This seemed to expose a bug where we render aggregations as group keys in the SQL query.

kszucs:
Closing this in favor a less invasive alternative.

PR #8541, opened 2024-03-04
ci(flink): support building the flink docker image on both arm64 and amd64 architectures

kszucs:

PR #8549, opened 2024-03-05
chore(docker): ensure that `just up <backend>` builds the image if necessary

kszucs:
Noticed in https://github.com/ibis-project/ibis/pull/8541 after changing the flink dockerfile that the image hasn't been rebuilt.

### DIFF
--- a/justfile
+++ b/justfile
@@ -98,7 +98,7 @@ download-data owner="ibis-project" repo="testing-data" rev="master":
 
 # start backends using docker compose; no arguments starts all backends
 up *backends:
-    docker compose up --wait {{ backends }}
+    docker compose up --build --wait {{ backends }}
 
 # stop and remove containers -> clean up dangling volumes -> start backends
 reup *backends:


justfile
```diff
@@ -98,7 +98,7 @@ download-data owner="ibis-project" repo="testing-data" rev="master":
 
 # start backends using docker compose; no arguments starts all backends
 up *backends:
-    docker compose up --wait {{ backends }}
+    docker compose up --build --wait {{ backends }}
```
cpcloud, line 101: I don't think this change is necessary. I never have this problem with any of the containers we build, they get built automatically if necessary.

cpcloud (changes requested):

justfile
cpcloud: Ah, nope. I'm wrong.

I don't really like this solution because it will rebuild images even if they don't need to be rebuilt, but I'm not sure what the alternative is.

cpcloud (approved):

justfile
kszucs: Docker layer cache will ensure that these are not actually rebuilt from scratch, it does nothing if the layer/image already exists.

cpcloud: I see a build of the image every time no matter what when running just up postgres.

gforsyth: I think it will still show the build no matter what, but the various steps should be prefixed with `CACHED`

gforsyth: e.g.
```
 => [postgres 1/2] FROM docker.io/postgis/postgis:15-3.3-alpine@sha256:4738bee  0.0s
 => CACHED [postgres 2/2] RUN apk add --no-cache postgresql15-plpython3         0.0s
 ```

cpcloud: Ah ok!

PR #8552, opened 2024-03-05
test: add xfail test for nested deferreds in Array.map()

NickCrews:
Example for #8553 

Can get this to fail with `pytest -m duckdb ibis/backends/tests/test_array.py::test_array_map_deferred`


cpcloud:
Superseded by #8602.

PR #8554, opened 2024-03-05
docs(concepts): remove outdated refs to SQLAlchemy

deepyaman:

### DIFF
--- a/gen_redirects.py
+++ b/gen_redirects.py
@@ -56,7 +56,6 @@ REDIRECTS = {
     "/api/backends/base/": "/reference/backends/base/",
     "/api/backends/pandas/": "/reference/backends/pandas/",
     "/api/backends/sql/": "/reference/backends/sql/",
-    "/api/backends/sqlalchemy/": "/reference/backends/sqlalchemy/",
     "/api/config/": "/reference/config/",
     "/api/datatypes/": "/reference/datatypes/",
     "/api/expressions/": "/reference/expressions/",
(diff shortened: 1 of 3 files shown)


gen_redirects.py
```diff
@@ -52,12 +52,6 @@
 # Untemplated redirects
 REDIRECTS = {
     "/api/": "/reference/",
-    "/api/backends/": "/reference/backends/",
```
gforsyth, line 55: I think we might want to leave this one in (so that now-very-old links redirect to the support matrix)

gen_redirects.py
```diff
@@ -52,12 +52,6 @@
...
-    "/api/backends/base/": "/reference/backends/base/",
-    "/api/backends/pandas/": "/reference/backends/pandas/",
```
gforsyth, line 57: any reason to remove the pandas redirect?

gen_redirects.py
```diff
@@ -52,12 +52,6 @@
...
-    "/api/backends/sql/": "/reference/backends/sql/",
-    "/api/backends/sqlalchemy/": "/reference/backends/sqlalchemy/",
-    "/api/config/": "/reference/config/",
```
gforsyth, line 60: this one perhaps we can re-target to `/how-to/configure/basics`?

gforsyth (commented):
Yay cleanup!  I have a few questions about some of the redirects we're removing, but nothing major.  Thanks for noticing that these were still hanging around!

gen_redirects.py
deepyaman: I think `"/reference/backends/"` isn't a valid target? https://ibis-project.org/reference/backends

Maybe it should just point to `"/backends/"` instead?

gen_redirects.py
deepyaman: Again, maybe it should be `"/backends/pandas/"` instead then.

gen_redirects.py
deepyaman: Alternatively, I could add an `"/reference/backends/": "/backends/"` redirect?

gforsyth: Maybe add the additional redirect, yeah.  I think that should be good.

deepyaman: OK, pretty much just done that.

cpcloud (approved):

PR #8557, opened 2024-03-05
docs: add concepts guide on Datatypes and Datashapes

NickCrews:
inspired by https://github.com/ibis-project/ibis/pull/8358

### DIFF
--- a/docs/concepts/datatypes_datashapes.qmd
+++ b/docs/concepts/datatypes_datashapes.qmd
@@ -1,0 +1,202 @@
+# Datatypes and Datashapes
+
+Every value in Ibis has two important properties: a type and shape.
+
+The type is probably familiar to you. It is something like
+
+- `Integer`
+- `Floating`
+- `String`
+- `Array`
+
+The shape is one of
+
+- `Scalar` (a single value)
+- `Column` (a series of values)
+
+## Datatype Flavors
+
+For some datatypes, there are further options that define them.
+For instance, `Integer` values can be signed or unsigned, and
+they have a precision. For example, "uint8", "int64", etc.
+These flavors don't affect their capabilities
+(eg both signed and unsigned ints have a `.abs()` method),
+but the flavor does impact how the underlying backend performs the computation.
+
+## Capabilities
+
+Depending on the combination of datatype and datashape, a value has
+different capabilities. For example:
+
+- All `String` values (both `StringScalars` and `StringColumns`) have the
+  method `.upper()` that transforms the string to uppercase.
+  `Floating` and `Array` values don't have this method, of course.
+- `IntegerColumn` and `FloatingColumn`  values have `.mean()`, `.max()`, etc methods
+  (which return `FloatingScalar` values),
+  because you can aggregate over them, since they are a collection of values.
+  On the other hand, `IntegerScalar` and `FloatingColumn` values do **not** have these
+  methods, because it doesn't make sense to take the mean or max of a single value.
+- If you call `.execute()` on these values, you get different results.
+  `Scalar` shapes result in plain python objects:
+  - `IntegerScalar`: plain python `int` object.
+  - `FloatingScalar`: plain python `float` object.
+  - `StringScalar`: plain python `str` object.
+  - `ArrayScalar`: plain python `list` object.
+  On the other hand, `Column` shapes result in `pandas.Series`:
+  - `IntegerColumn`: pd.Series of integers, with the same flavor.
+       For example, if the `IntegerColumn` was specifically "uint16",
+       then the pandas series will also be of type "uint16".
+  - `FloatingColumn`: pd.Series of floats with the same flavor.
+  - etc.
+
+## Broadcasting and Alignment
+
+There are rules for how different datashapes are combined. This is similar to
+how SQL and numpy handles merging datashapes, if you are familiar with them.
+
+```python
+import ibis
+
+ibis.options.interactive = True
+t1 = ibis.examples.penguins.fetch().head(100)
+t1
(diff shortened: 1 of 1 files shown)


docs/concepts/datatypes_datashapes.qmd
```diff
@@ -0,0 +1,202 @@
...
+Depending on the combination of datatype and datashape, a value has
+different capabilities. For example:
+
+- All `String` values (both `StringScalars` and `StringColumns`) have the
+  method `.upper()` that transforms the string to uppercase.
+  `Floating` and `Array` values don't have this method, of course.
+- `IntegerColumn` and `FloatingColumn`  values have `.mean()`, `.max()`, etc methods
+  (which return `FloatingScalar` values),
+  because you can aggregate over them, since they are a collection of values.
+  On the other hand, `IntegerScalar` and `FloatingColumn` values do **not** have these
```
gforsyth, line 37:
```suggestion
  On the other hand, `IntegerScalar` and `FloatingScalar` values do **not** have these
```

docs/concepts/datatypes_datashapes.qmd
```diff
@@ -0,0 +1,202 @@
...
+  methods, because it doesn't make sense to take the mean or max of a single value.
+- If you call `.execute()` on these values, you get different results.
```
gforsyth, line 39:
```suggestion
- If you call `.to_pandas()` on these values, you get different results.
```

or `to_pyarrow`?  but we're trying to get away from `execute`

docs/concepts/datatypes_datashapes.qmd
```diff
@@ -0,0 +1,202 @@
...
+
+## Broadcasting and Alignment
+
+There are rules for how different datashapes are combined. This is similar to
+how SQL and numpy handles merging datashapes, if you are familiar with them.
+
+```python
+import ibis
+
+ibis.options.interactive = True
```
gforsyth, line 60: Maybe make this whole section executable to handle rendering the output?

docs/concepts/datatypes_datashapes.qmd
```diff
@@ -0,0 +1,202 @@
...
+  - `IntegerColumn`: pd.Series of integers, with the same flavor.
+       For example, if the `IntegerColumn` was specifically "uint16",
+       then the pandas series will also be of type "uint16".
+  - `FloatingColumn`: pd.Series of floats with the same flavor.
+  - etc.
+
+## Broadcasting and Alignment
+
+There are rules for how different datashapes are combined. This is similar to
+how SQL and numpy handles merging datashapes, if you are familiar with them.
```
gforsyth, line 55:
```suggestion
how SQL and NumPy handles merging datashapes, if you are familiar with them.
```

I think it's always NumPy in prose

docs/concepts/datatypes_datashapes.qmd
```diff
@@ -0,0 +1,202 @@
...
+# │                    2049.0 │
+# │                         … │
+# └───────────────────────────┘
+```
+
+One requirement that might surprise you if you are coming from Numpy is
+Ibis's requirements on aligning `Columns`: In Numpy, if you have two arbitrary
+arrays, each of length 100, you can add them together, and it works because the
+elements are "lined up" based on position. Ibis is different. Because it is based
+around SQL, and SQL has no notion of inherint row ordering, you cannot "line up"
```
gforsyth, line 134:
```suggestion
around SQL, and SQL has no notion of inherent row ordering, you cannot "line up"
```

docs/concepts/datatypes_datashapes.qmd
gforsyth: ahh, I see below you are referencing pandas datashapes, so definitely `to_pandas`, then

docs/concepts/datatypes_datashapes.qmd
```diff
@@ -0,0 +1,202 @@
+# Datatypes and Datashapes
```
gforsyth, line 1:
```suggestion
---
title: Datatypes and Datashapes
---
```

for multi-word titles, this also handles setting the page name in a browser tab

gforsyth (commented):
Quick pass, but this is really nice @NickCrews -- thanks for putting it together!  I've flagged a few typos and some styling nits, nothing major.

I do think that most of the code examples can be made executable (I'm not sure about the one that raises the `RelationError`...)

lostmygithubaccount:
I would make the code blocks runnable; use a `try/except` to catch the error and print it out

cpcloud:
You can also set `error: true` for a given cell

NickCrews:
Thanks for all the thoughts! I agree with all of them.

Two lingering thoughts:

1. I totally ignored Tables. This could be read that Tables also have a datatype and datashape, which is wrong. But I thought it might just be confusing? if you have a good way of putting that in there, would be great
2. curious what you think about how I described "flavor". Is that the term we want to use? Are there other aspects of that that are important? I almost left it out, but I think it's not too distracting.

gforsyth:
> 1. I totally ignored Tables. This could be read that Tables also have a datatype and datashape, which is wrong. But I thought it might just be confusing? if you have a good way of putting that in there, would be great

I think this is ok to ignore -- if you wanted to be explicit and make sure it's covered, I think I would introduce it in a callout note and say something like (unpolished):

    A table is a collection of one or more columns. Because each column can have a different datatype and the number of columns 
is highly variable, a table itself does not have a datatype or a datashape.  



> 2\. curious what you think about how I described "flavor". Is that the term we want to use? Are there other aspects of that that are important? I almost left it out, but I think it's not too distracting.

I liked it and I think it's a good balance of a word that will both communicate the idea to readers less familiar with datatypes but still correct enough for more familiar readers.

gforsyth (approved):
I think this can go in, but I'll hold off on merging in case you have any other tweaks you'd like to make @NickCrews .  Thanks again for putting this together!

NickCrews:
Thanks for the reviews! This looks good to me, merge it!

cpcloud:
Gah, definitely meant to get in a review of this, but it was minor so I'll just put up a PR to change.

PR #8564, opened 2024-03-06
fix: `Node.map_clear` should return `Any`

peter-gy:
## Description of changes

Previously `Node.map_clear` had a `dict` return type annotation,  just as `Node.map`, even though the main rationale behind `map_clear` is that it returns only the result of the root node and discards all the intermediaries.


kszucs (approved):
Thanks @peter-gy!

PR #8568, opened 2024-03-06
refactor(sql): simplify `FirstValue`/`LastValue` usage

jcrist:
During recent refactors the `FirstValue`/`LastValue` have been pretty much removed from the codebase - at this point they only exist since some backends require spelling `first`/`last` as `first_value`/`last_value` when used in a window function context (other backends alias `first`/`last` to these for nicer SQL UX).

This PR:

- Moves `FirstValue`/`LastValue` to `ibis/backends/sql/rewrites.py` to treat them as SQL-specific IR.
- Refactors the `First -> FirstValue`/`Last -> LastValue` window function rewrites to always apply for all SQL backends, making our compilation more uniform here.

### DIFF
--- a/ibis/backends/sql/rewrites.py
+++ b/ibis/backends/sql/rewrites.py
@@ -63,6 +63,28 @@ class Select(ops.Relation):
 
 
 @public
+class FirstValue(ops.Analytic):
+    """Retrieve the first element."""
+
+    arg: ops.Column[dt.Any]
+
+    @attribute
+    def dtype(self):
+        return self.arg.dtype
+
+
+@public
+class LastValue(ops.Analytic):
+    """Retrieve the last element."""
+
+    arg: ops.Column[dt.Any]
+
+    @attribute
+    def dtype(self):
+        return self.arg.dtype
+
+
+@public
 class Window(ops.Value):
     """Window modelled after SQL's window statements."""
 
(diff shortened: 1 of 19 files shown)


ibis/backends/sql/rewrites.py
```diff
@@ -100,10 +122,23 @@ def sort_to_select(_, **kwargs):
...
 @replace(p.WindowFunction)
 def window_function_to_window(_, **kwargs):
-    """Convert a WindowFunction node to a Window node."""
+    """Convert a WindowFunction node to a Window node.
+
+    Also rewrites first -> first_value, last -> last_value.
+    """
+    func = _.func
+    if isinstance(func, (ops.First, ops.Last)):
+        if func.where is not None:
```
cpcloud, line 131: This doesn't seem conceptually invalid. Perhaps we should avoid failing here so that if/when this is implemented it's an `XPASS`.

jcrist: It's not clear how that would work - the issue here is that `first()`/`last()` in ibis can serve as both reductions and analytic functions, where in the windowing context they only work as analytic functions (no `where` clause available). Since this isn't a regression from 8.0 (or from the current behavior) I'm inclined to not figure out how to plumb `where` through here and get it to fail nicely until/if compilation works later.

kszucs (approved):

Issue #8143, opened 2024-01-29
bug: geospatial literals don't work out of the box with the default backend


ncclementi:
### What happened?

When trying to do an operation on a geospatial literal object, it fails due to missing "spatial" extension. Tried creating a connection, and loading the extension, but problem persists:

Minimal reproducible example
```python
In [1]: import ibis

In [2]: import shapely

In [3]: a = shapely.LineString([[0, 0], [1, 0], [1, 1]])

In [4]: al = ibis.literal(a, type="geometry")

In [5]: al
Out[5]: LINESTRING (0 0, 1 0, 1 1)

In [6]: al.length()
Out[6]: GeoLength(LINESTRING (0 0, 1 0, 1 1)): GeoLength(LINESTRING (0 0, 1 0, 1 1))

In [7]: from ibis.interactive import *

In [8]: al.length()
```

```python-traceback
ProgrammingError: (duckdb.duckdb.CatalogException) Catalog Error: Scalar Function with name "st_length" is not in the catalog, but it exists in the spatial extension.

Please try installing and loading the spatial extension:
INSTALL spatial;
LOAD spatial;


[SQL: SELECT ST_Length('LINESTRING (0 0, 1 0, 1 1)'::geometry) AS "GeoLength(LINESTRING (0 0, 1 0, 1 1))"]
(Background on this error at: https://sqlalche.me/e/14/f405)
```

```python
In [9]: con = ibis.duckdb.connect()

In [10]: con.load_extension("spatial")

In [11]: al.length()
```

```python-traceback
ProgrammingError: (duckdb.duckdb.CatalogException) Catalog Error: Scalar Function with name "st_length" is not in the catalog, but it exists in the spatial extension.

Please try installing and loading the spatial extension:
INSTALL spatial;
LOAD spatial;


[SQL: SELECT ST_Length('LINESTRING (0 0, 1 0, 1 1)'::geometry) AS "GeoLength(LINESTRING (0 0, 1 0, 1 1))"]
(Background on this error at: https://sqlalche.me/e/14/f405)
```

### What version of ibis are you using?

Ibis main branch

### What backend(s) are you using, if any?

duckdb

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ncclementi:
This came up while creating the examples for #8128, but I noticed that it wasn't failing in CI. The reason why in CI things worked is because some of the tests run before load the `spatial` extension. 

[...]

PR #8570, opened 2024-03-06
feat(geospatial): add support for duckdb operations on literals

ncclementi:
## Description of changes

This PR add support for duckdb geospatial operations directly on geospatial literals. 

I gave this a try, I'm not sure this is the best way to go, but happy to iterate over it if there is a better way around. 

## Issues closed

Closes: https://github.com/ibis-project/ibis/issues/8143

cc: @gforsyth (thanks for the pointer to `_run_pre_execute_hooks`)

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -1178,6 +1178,10 @@ class Backend(SQLBackend, CanCreateSchema, UrlFromPath):
                 )
             elif started is False:
                 self._record_batch_readers_consumed[t.name] = True
+
+        if expr.op().find((ops.GeoSpatialUnOp, ops.GeoSpatialBinOp)):
+            self.load_extension("spatial")
+
         super()._run_pre_execute_hooks(expr)
 
     def to_pyarrow_batches(
--- a/ibis/backends/duckdb/tests/test_geospatial.py
+++ b/ibis/backends/duckdb/tests/test_geospatial.py
@@ -306,3 +306,31 @@ def test_literal_geospatial_inferred(con, shp, expected, snapshot):
 def test_load_geo_example(con):
     t = ibis.examples.zones.fetch(backend=con)
     assert t.geom.type().is_geospatial()
+
+
+# For the next two tests we really want to ensure that
+# load_extenstion("spatial") hasn't been run yet. So we create a new connection
+# instead of using the  con fixture.
+
+# geospatial literal object,
+geo_line_lit = ibis.literal(
+    shapely.LineString([[0, 0], [1, 0], [1, 1]]), type="geometry"
+)
+
+
+@pytest.mark.parametrize("geo_line_lit", [geo_line_lit])
+def test_geo_unop_geo_literals(geo_line_lit):
+    # GeoSpatialUnOp operation on a geospatial literal
+    con = ibis.duckdb.connect()
+    expr = geo_line_lit.length()
+
+    assert con.execute(expr) == 2
+
+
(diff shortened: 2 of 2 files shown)


ibis/backends/duckdb/tests/test_geospatial.py
```diff
@@ -306,3 +306,13 @@ def test_literal_geospatial_inferred(con, shp, expected, snapshot):
 def test_load_geo_example(con):
     t = ibis.examples.zones.fetch(backend=con)
     assert t.geom.type().is_geospatial()
+
+
+def test_ops_geo_literals():
```
jcrist, line 311: Add a `con` here should fix your issue. Otherwise `.execute()` will instantiate a default backend.

```suggestion
def test_ops_geo_literals(con):
```

ibis/backends/duckdb/tests/test_geospatial.py
```diff
@@ -306,3 +306,13 @@ def test_literal_geospatial_inferred(con, shp, expected, snapshot):
...
+    # Do a geospatial operation on a geospatial literal object,
+
+    geo_line_lit = ibis.literal(
+        shapely.LineString([[0, 0], [1, 0], [1, 1]]), type="geometry"
+    )
+
+    assert geo_line_lit.length().execute() == 2
```
jcrist, line 318:
```suggestion
    assert con.execute(geo_line_lit.length()) == 2
```

ibis/backends/duckdb/__init__.py
```diff
@@ -1178,6 +1178,10 @@ def _run_pre_execute_hooks(self, expr: ir.Expr) -> None:
                 )
             elif started is False:
                 self._record_batch_readers_consumed[t.name] = True
+
+        if expr.op().__class__.__module__ == "ibis.expr.operations.geospatial":
```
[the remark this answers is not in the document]
ncclementi: Oh I see, this makes way more sense, I wasn't aware of how find worked but that makes a lot of sense. 
The length test triggers only `ops.GeoSpatialUnOp` I'll add another case that triggers `ops.GeoSpatialBinOp` for completion

ibis/backends/duckdb/tests/test_geospatial.py
```diff
@@ -306,3 +306,21 @@ def test_literal_geospatial_inferred(con, shp, expected, snapshot):
...
+
+
+@pytest.mark.parametrize(
+    ("op", "right", "expected"),
+    [
+        param("length", [], 2, id="GeoSpatialUnOp"),
+        param("distance", [shapely.Point(0, 0)], 0, id="GeoSpatialBinOp"),
+    ],
+)
+def test_ops_geo_literals(con, op, right, expected):
```
jcrist, line 318: It just occurred to me that we may want to use a new connection here since we have no guarantee that this geospatial test ran before other ones - for this test we really want to ensure that `load_extenstion("spatial")` hasn't been run for this connection yet. So in that case we don't want to use the `con` fixture here but instead create a new `con` in the test.

cpcloud (approved):
I think the framing of the unary op test case as as binary op with no arguments is a bit awkward, but non-blocking.

Implementation looks 👌🏻

ncclementi:
Just in case, Don't merge yet. 

I'll take care of Jim's comment about the `con` he is right, and I'll split the tests in two separate ones. I was trying to get cute with it, but it'll be more confusing for future devs.

ncclementi:
This one is ready to go/ ready for the last review.

cpcloud (approved):
LGTM, thanks!

PR #8572, opened 2024-03-06
fix(sql): replace CTEs within CTEs

kszucs:
While the CTE extraction worked properly, the CTEs haven't been replaced in the already extracted CTEs meaning that we didn't recursively replace.

### DIFF
--- a/ibis/backends/sql/rewrites.py
+++ b/ibis/backends/sql/rewrites.py
@@ -188,7 +188,7 @@ def extract_ctes(node):
     g = Graph.from_bfs(node, filter=(ops.Relation, ops.Subquery, ops.JoinLink))
     for node, dependents in g.invert().items():
         if isinstance(node, ops.View) or (
-            len(dependents) > 1 and isinstance(node, cte_types)
+            isinstance(node, cte_types) and len(dependents) > 1
         ):
             result.append(node)
 
@@ -236,9 +236,14 @@ def sqlize(
     simplified = sqlized.replace(merge_select_select)
 
     # extract common table expressions while wrapping them in a CTE node
-    ctes = extract_ctes(simplified)
-    subs = {cte: CTE(cte) for cte in ctes}
-    result = simplified.replace(subs)
+    ctes = set(extract_ctes(simplified))
+
+    def wrap(node, _, **kwargs):
+        new = node.__recreate__(kwargs)
+        return CTE(new) if node in ctes else new
+
+    result = simplified.replace(wrap)
+    ctes = reversed([cte.parent for cte in result.find(CTE)])
 
     return result, ctes
 
--- a/ibis/backends/sql/rewrites.py
+++ b/ibis/backends/sql/rewrites.py
@@ -188,7 +188,7 @@ def extract_ctes(node):
     g = Graph.from_bfs(node, filter=(ops.Relation, ops.Subquery, ops.JoinLink))
     for node, dependents in g.invert().items():
         if isinstance(node, ops.View) or (
-            isinstance(node, cte_types) and len(dependents) > 1
+            len(dependents) > 1 and isinstance(node, cte_types)
         ):
             result.append(node)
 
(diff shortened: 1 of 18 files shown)


cpcloud:
This looks like it might help with https://github.com/ibis-project/ibis/issues/8484.

@kszucs Are the remaining failures due to needing to regen the SQL snapshots for those backends?

ibis/backends/sql/rewrites.py
```diff
@@ -236,9 +236,14 @@ def sqlize(
...
-    subs = {cte: CTE(cte) for cte in ctes}
-    result = simplified.replace(subs)
+    ctes = set(extract_ctes(simplified))
+
+    def wrap(node, _, **kwargs):
+        new = node.__recreate__(kwargs)
+        return CTE(new) if node in ctes else new
+
+    result = simplified.replace(wrap)
+    ctes = reversed([cte.parent for cte in result.find(CTE)])
```
cpcloud, line 246: Is the `find` + `reversed` necessary to preserve topological order? If so, can you add a short comment about it here?

ibis/backends/sql/rewrites.py
```diff
@@ -188,7 +188,7 @@ def extract_ctes(node):
     g = Graph.from_bfs(node, filter=(ops.Relation, ops.Subquery, ops.JoinLink))
     for node, dependents in g.invert().items():
         if isinstance(node, ops.View) or (
-            len(dependents) > 1 and isinstance(node, cte_types)
+            isinstance(node, cte_types) and len(dependents) > 1
```
cpcloud, line 191: Why reverse the check order here?

cpcloud: Can you leave this as is? There's no obvious reason to churn this code right now.

ibis/backends/sql/rewrites.py
```diff
@@ -236,9 +236,14 @@ def sqlize(
     simplified = sqlized.replace(merge_select_select)
 
     # extract common table expressions while wrapping them in a CTE node
-    ctes = extract_ctes(simplified)
-    subs = {cte: CTE(cte) for cte in ctes}
-    result = simplified.replace(subs)
+    ctes = set(extract_ctes(simplified))
```
cpcloud, line 239:
```suggestion
    ctes = frozenset(extract_ctes(simplified))
```

Might as well make it immutable.

ibis/backends/sql/rewrites.py
kszucs: Sure, I was just discovering this issue while trying to extract mssql subqueries with sortby expressions and I had another instancecheck here which looked easier to understand. Going to restore.

ibis/backends/sql/rewrites.py
kszucs: Find traverse in DFS/BFS not topological. Replace do traverse in topological order so we could extract the rewritten CTEs in the closure, but my first naive attempt didn't work and I was lazy inspecting it. We can defer it to a follow-up.

kszucs:
> Are the remaining failures due to needing to regen the SQL snapshots for those backends?

Yes, thanks for updating those snapshots!

ibis/backends/sql/rewrites.py
kszucs: Going to update.

ibis/backends/sql/rewrites.py
```diff
@@ -236,9 +236,14 @@ def sqlize(
...
+
+    def wrap(node, _, **kwargs):
```
kszucs, line 241: Going to add this as comment, but the the problem with the previous solution was that `replace(dict)` matches on the original node and replaces the given value, but the ctes in the `subs` mapping are the original relation objects.

Instead we need to do the following: once we encounter a relation which is occurring more than once (`in ctes`) we need to wrap then in a `CTE` operation, which must propagate upwards in topological order, hence we are using a custom replacer which matches on the original (not the recreated) node.

ibis/backends/sql/rewrites.py
cpcloud: Ok, I will look through the snapshots again to make sure CTEs are occurring topological order.

ibis/backends/sql/rewrites.py
cpcloud:
```suggestion
            len(dependents) > 1 and isinstance(node, cte_types)
```

cpcloud (approved):

cpcloud:
Going to see whether this helps with #8484.

cpcloud:
Doesn't really help with #8484.

PR #8578, opened 2024-03-07
feat(exasol): add support for ExtractDayOfYear

Nicoretti:
## Description of changes
Add implementation for ExtractDayOfYear to Exasol backend

### DIFF
--- a/ibis/backends/exasol/compiler.py
+++ b/ibis/backends/exasol/compiler.py
@@ -58,7 +58,6 @@ class ExasolCompiler(SQLGlotCompiler):
             ops.DayOfWeekIndex,
             ops.DayOfWeekName,
             ops.ElementWiseVectorizedUDF,
-            ops.ExtractDayOfYear,
             ops.ExtractEpochSeconds,
             ops.ExtractQuarter,
             ops.ExtractWeekOfYear,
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
Sweet!

Nicoretti:
@cpcloud I've just rebased and simplified the implementation

ibis/backends/exasol/compiler.py
```diff
@@ -178,3 +177,6 @@ def visit_DateDelta(self, op, *, part, left, right):
         #       * part = hour -> hours_between
         #       * ...
         return self.f.days_between(left, right)
+
+    def visit_ExtractDayOfYear(self, op, *, arg):
+        return self.f.to_char(arg, "DDD")
```
cpcloud, line 182:
```suggestion
        return self.cast(self.f.to_char(arg, "DDD"), op.dtype)
```

So that the casting to int is done in the database and not in-memory by pandas/pyarrow.

cpcloud (changes requested):

ibis/backends/exasol/compiler.py
Nicoretti: Thanks @cpcloud. Initially, I had it like that. The tests have been passing without it 😅. I must admit, I'm still getting acquainted with the new architecture/structure. However, I really like it 👍.

cpcloud: Yep, they're passing because we cast results in python to ensure that the output dtypes match the ibis expression's types. In this case, it's much less efficient presumably, because casting from string to int in exasol is likely much faster than computing a bunch of strings only to throw them away immediately.

cpcloud (approved):

PR #8591, opened 2024-03-08
feat(exasol): add support for TypeOf

Nicoretti:
## Description of changes
Add implementation for TypeOf to Exasol backend

### DIFF
--- a/ibis/backends/exasol/compiler.py
+++ b/ibis/backends/exasol/compiler.py
@@ -84,7 +84,6 @@ class ExasolCompiler(SQLGlotCompiler):
             ops.TimestampDelta,
             ops.TimestampDiff,
             ops.TimestampSub,
-            ops.TypeOf,
             ops.Unnest,
             ops.Variance,
         )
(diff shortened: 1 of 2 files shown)


ibis/backends/exasol/compiler.py
```diff
@@ -187,3 +186,6 @@ def visit_DayOfWeekName(self, op, *, arg):
 
     def visit_ExtractQuarter(self, op, *, arg):
         return self.cast(self.f.to_char(arg, "Q"), op.dtype)
+
+    def visit_TypeOf(self, op, *, arg):
```
cpcloud, line 190: Simple mappings like this can be added to `SIMPLE_OPS`.

cpcloud (changes requested):

ibis/backends/exasol/compiler.py
Nicoretti: Nice, thanks.

cpcloud (approved):

Nicoretti:
@cpcloud, it appears that there are a few follow-up items here that I hadn't initially noticed. I'll close this for the time being and will recreate it once I've addressed those issues.

Side note: I plan to submit some more PRs next week.

PR #8592, opened 2024-03-08
perf(sql): prevent sqlglot from extensive deepcopying every time we create a sqlglot object

kszucs:
Still profiling and adding more `copy=False` options, apparently it greatly improves the performance.

According to profiling the recursive generation of sqlglot is still a bottleneck for big queries which cannot be fixed on the ibis side. There could be one option though to compile the fragments in a greedy fashion which are going to be cached by `Node.map()` and inject those as arbitrary strings to other sqlglot expressions.

The benchmark shows a pretty solid improvement:

```
------------------------------------------------------------------------------ benchmark 'test_compile[large-duckdb]': 2 tests -------------------------------------------------------------------------------
Name (time in ms)                                 Min                Max               Mean            StdDev             Median               IQR            Outliers       OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_compile[large-duckdb] (0399_182b6a5)     15.0069 (1.63)     43.1253 (1.38)     17.0702 (1.76)     5.8277 (2.57)     15.7781 (1.69)     0.3220 (1.28)         3;11   58.5818 (0.57)         59           1
test_compile[large-duckdb] (NOW)               9.2348 (1.0)      31.1592 (1.0)       9.6997 (1.0)      2.2679 (1.0)       9.3424 (1.0)      0.2524 (1.0)           1;7  103.0959 (1.0)          94           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Also measured the heap usage right after translation:

```py
BEFORE
Partition of a set of 3349536 objects. Total size = 418566016 bytes.
 Index  Count   %     Size   % Cumulative  % Kind (class / dict of class)
     0 994115  30 180331768  43 180331768  43 dict (no owner)
     1 435432  13 48768384  12 229100152  55 sqlglot.expressions.Literal
     2 435432  13 31351104   7 260451256  62 dict of sqlglot.expressions.Literal
     3  94643   3 14095526   3 274546782  66 str
     4 105890   3 11859680   3 286406462  68 sqlglot.expressions.Identifier
     5  23325   1  8770520   2 295176982  71 types.CodeType
     6 105890   3  7624080   2 302801062  72 dict of sqlglot.expressions.Identifier
     7  46724   1  7169992   2 309971054  74 list
     8  85244   3  6257008   1 316228062  76 tuple
     9  52938   2  5929056   1 322157118  77 sqlglot.expressions.Column

[...]

PR #8596, opened 2024-03-08
perf(sql): reduce the memory overhead of sqlglot compilation [WIP]

kszucs:
I experienced heavy memory usage during the reproduction of #8484, gigabytes of memory was used...

As turned out sqlglot expressions are aggressively deepcopied so we should minimize the amount of sqlglot expressions we hold references to. `Node.map()` keeps the output for each ibis expression in a mapping during translation which is not ideal in this scenario. 

Luckily #7863 added a memory efficient `.map_clear()` alternative which was primarily used in the pandas backend, but turns out to be useful here as well.

Depends on #8592


kszucs:
After the merge of https://github.com/ibis-project/ibis/commit/461293bb77945c37460e6927b7455f4f8a21a054 this isn't significantly improving the memory usage so closing it for now.

PR #8602, opened 2024-03-08
fix(api): improve error message raised on improper calls to array `map` or `filter`

### LINKED ISSUE #8553: bug: inner Deferred binds to wrong thing in `Table.select(_.x.map(_.b))`
### What happened?

See https://github.com/ibis-project/ibis/pull/8552

I think what is happening:
1. ibis encounters the _.x when evaluating the .select
2. it tries to bind it to the parent expression, the Table
3. then it recurses down into the deferred to further evaluate it
4. It encounters the _.b
5. It isn't smart enough to understand that this now refers to each array element. It just binds it to that outer Table
6. raises AttributeError

This seems like it might be a doozy to solve. Would need to make the idea of a context much more nuanced, with nested scopes??

main

_No response_

### Relevant log output

_No response_

### Code of Conduct

jcrist:
Fixes #8553, supersedes #8552.

We can't realistically catch the invalid use of deferred sin the original issue, but we can ensure that in the case the deferred resolves to something valid we raise a nice error message rather than whatever `inspect.signature` happens to raise.

### DIFF
--- a/ibis/expr/types/arrays.py
+++ b/ibis/expr/types/arrays.py
@@ -445,15 +445,19 @@ class ArrayValue(Value):
         """
         if isinstance(func, Deferred):
             name = "_"
-        else:
+            resolve = func.resolve
+        elif callable(func):
             name = next(iter(inspect.signature(func).parameters.keys()))
+            resolve = func
+        else:
+            raise TypeError(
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
LGTM!

PR #8616, opened 2024-03-11
perf(sql): don't compile pretty sql by default

kszucs:
We have been compiling pretty formatted SQL queries so far. While this was handy, pretty formatting doesn't give any value when directly passed to a backend, it mostly matters when presented to users.

Compiling a big query like the one provided in https://github.com/ibis-project/ibis/issues/8484 takes `3.83s` with `pretty=True` while it takes almost half the time `2.17s` with `pretty=False`.


cpcloud:
Does the use of formatted SQL have a meaningful impact on SQL generation performance?

cpcloud (approved):

Issue #8567, opened 2024-03-06
Deprecate legacy UDF functionality


jcrist:
Before the 9.0 release we should deprecate the older UDF/UDAF functionality that uses `ibis.legacy.udf`. For the `elementwise` case users can and should make use of the new `ibis.udf.*` apis. For `reduction` and `analytic` functions we don't have a method for handling these in the new system. That said, I suspect we have few/no users of this functionality in the wild and their backend-specific nature doesn't seem worth the maintenance burden.

jcrist:
Alternatively we could do a clean break and just delete it. I wouldn't be opposed, I suspect we have few users of this functionality anyway.

PR #8617, opened 2024-03-11
depr: deprecate and warn on `legacy` udf usage

ncclementi:
This PR adds warnings for deprecation of `ibis.legacy.udf` for `v9.0`. I'm not super sure about the `instead` message for `elementwise`, suggestions are welcomed. 

According to the issue, seems like we are not going to support `analytic` and `reduction` on the new system, hence the empty comment on `instead`

- Closes https://github.com/ibis-project/ibis/issues/8567

### DIFF
--- a/ibis/legacy/udf/vectorized.py
+++ b/ibis/legacy/udf/vectorized.py
@@ -19,6 +19,7 @@ from ibis.expr.operations import (
     ElementWiseVectorizedUDF,
     ReductionVectorizedUDF,
 )
+from ibis.util import deprecated
 
 if TYPE_CHECKING:
     import pandas as pd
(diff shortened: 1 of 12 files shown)


ncclementi:
Whoops I forgot to add `with pytest.warns(FutureWarning)` in the corresponding tests. Coming right up

 `

ibis/legacy/udf/vectorized.py
```diff
@@ -350,6 +353,7 @@ def elementwise(input_type, output_type):
     return _udf_decorator(ElementWiseVectorizedUDF, input_type, output_type)
 
 
+@deprecated(as_of="9.0", instead="")
```
cpcloud, line 356: Builtin reductions can still be accessed. Perhaps it is fine (and helpful?) to say `"use the `@ibis.udf.agg.builtin` decorator"` or something similar.

cpcloud (changes requested):

cpcloud (approved):
Thanks!

PR #8618, opened 2024-03-11
fix(sql): outer order by should take precedence over inner order by

kszucs:
Now we turn 

```
r0 := DatabaseTable: functional_alltypes
  id              int64
  bool_col        boolean
  tinyint_col     int8
  smallint_col    int16
  int_col         int32
  bigint_col      int64
  float_col       float32
  double_col      float64
  date_string_col string
  string_col      string
  timestamp_col   timestamp(6)
  year            int64
  month           int64

r1 := Sort[r0]
  asc r0.tinyint_col
  asc r0.bool_col

Sort[r1]
  asc r1.bool_col
  desc r1.tinyint_col + 1
```

into 

```
r0 := DatabaseTable: functional_alltypes
  id              int64
  bool_col        boolean
  tinyint_col     int8
  smallint_col    int16
  int_col         int32
  bigint_col      int64
  float_col       float32
  double_col      float64
  date_string_col string
  string_col      string
  timestamp_col   timestamp(6)
  year            int64
  month           int64

Select[r0]
  selections:
    id:              r0.id
    bool_col:        r0.bool_col
    tinyint_col:     r0.tinyint_col
    smallint_col:    r0.smallint_col
    int_col:         r0.int_col
    bigint_col:      r0.bigint_col
    float_col:       r0.float_col
    double_col:      r0.double_col
    date_string_col: r0.date_string_col
    string_col:      r0.string_col
    timestamp_col:   r0.timestamp_col
    year:            r0.year
    month:           r0.month
  sort_keys:
    asc r0.bool_col
    desc r0.tinyint_col + 1
    asc r0.tinyint_col
  schema:
    id              int64
    bool_col        boolean
    tinyint_col     int8
    smallint_col    int16
    int_col         int32
    bigint_col      int64
    float_col       float32
    double_col      float64
    date_string_col string
    string_col      string
    timestamp_col   timestamp(6)
    year            int64
    month           int64
```

Aims to resolve https://github.com/ibis-project/ibis/issues/8442#issuecomment-1967578311

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -519,9 +519,38 @@ def test_dropna_table(backend, alltypes, how, subset):
     backend.assert_frame_equal(result, expected)
 
 
-def test_select_sort_sort(alltypes):
-    query = alltypes[alltypes.year, alltypes.bool_col]
-    query = query.order_by(query.year).order_by(query.bool_col)
+def test_select_sort_sort(backend, alltypes, df):
+    t = alltypes
+
+    expr = t.order_by(t.year, t.id.desc()).order_by(t.bool_col)
+
+    result = expr.execute().reset_index(drop=True)
+    expected = df.sort_values(
+        ["bool_col", "year", "id"], ascending=[True, True, False]
+    ).reset_index(drop=True)
+
+    backend.assert_frame_equal(result, expected)
+
+
+def test_select_sort_sort_deferred(backend, alltypes, df):
+    t = alltypes
+
+    expr = t.order_by(t.tinyint_col, t.bool_col, t.id).order_by(
+        _.bool_col.asc(), (_.tinyint_col + 1).desc()
+    )
+    result = expr.execute().reset_index(drop=True)
+
+    df = df.assign(tinyint_col_plus=df.tinyint_col + 1)
+    expected = (
+        df.sort_values(
+            ["bool_col", "tinyint_col_plus", "tinyint_col", "id"],
+            ascending=[True, False, True, True],
+        )
+        .drop(columns=["tinyint_col_plus"])
+        .reset_index(drop=True)
+    )
+
+    backend.assert_frame_equal(result, expected)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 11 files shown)


NickCrews:
Will this correctly detect duplicates if the `__hash__`/identity of the order by clauses aren't the same? eg `t.order_by(t.a, t.c).order_by(t.b.asc(), _.a.desc())`? or more tricky `t.order_by(t.a, t.c).order_by(t.b.asc(), (t.a + 1).desc())`? I worry that trying to be clever and detect duplicates isn't going to always work, and instead we need to take the loss and just drop all previous order_bys, just using the most recent ones.

kszucs:
@NickCrews I added test cases for your examples. In the latter one 

```py
t.order_by(t.a, t.c).order_by(t.b.asc(), (t.a + 1).desc())
```

we would need to check that `ordering by t.a + 1` is semantically equivalent with `ordering by t.a`, while this is not impossible we would need more machinery which would produce the same result as executing `order by t.a + 1, t.a`.

cpcloud:
It seems like this is a slippery slope into semantic analysis hell.

Is merging sequential order bys really the right approach? What's the problem with taking the last `order_by`?

ibis/backends/tests/test_generic.py
```diff
@@ -1898,3 +1892,77 @@ def test_subsequent_overlapping_order_by(con, backend, alltypes, df):
...
+    ["druid"],
+    raises=PyDruidProgrammingError,
+    reason=(
+        "Query could not be planned. SQL query requires ordering a table by time column"
+    ),
+)
+@pytest.mark.broken(
+    ["dask"],
+    raises=AssertionError,
+    reason="dask doesn't support deterministic .sort_values()",
```
kszucs, line 1907: We can resolve this by rewriting the expression similarly to the SQL backends. Actually this should be done in a global simplification step implemented in `ibis/expr/rewrites.py`, I am planning to do that in a follow-up.

NickCrews:
Actually, if the order by expression uses a UDF, then I think it would be the halting problem to predict how that clause interacts with other clauses.

cpcloud (approved):
LGTM.

Just noting that we should eventually follow up with extracting order by keys from subqueries with limits, and merging those into the sort keys of the outer query.

PR #8626, opened 2024-03-12
fix(oracle): map bare `NUMBER` to `int64` and consolidate data type mapping code for shared inference

cpcloud:
Fixes a bug discovered during polars dot sql implementation where we were mapping empty precision and scale decimal types, but the oracle docs say that empty precision and scale together are mapped to number(38, 0).

The discovery during polars dot sql implementation is incidental, due to the test I changed, which is included here.


gforsyth (approved):
lovely -- thanks for wading around in all of that.

PR #8627, opened 2024-03-12
refactor(backends): remove `_metadata` method in favor of `_get_schema_using_query`

cpcloud:
Remove the unnecessary `_metadata` method in favor of `_get_schema_using_query`


kszucs (approved):
Much better, thanks for working on this!

PR #8630, opened 2024-03-12
feat(ir): support showing variable names used to create an expression in `repr()`

kszucs:
For the following code:

```py
import ibis

alltypes = ibis.table(...)

filtered = alltypes[alltypes.f > 0]
ordered = filtered.order_by("f")
projected = ordered[["a", "b", "f"]]

expr = projected.a + projected.b

ibis.options.repr.show_variables = True
repr(expr)
```

instead of showing 
```
r0 := UnboundTable: alltypes
  a int8
  b int16
  c int32
  d int64
  e float32
  f float64
  g string
  h boolean
  i timestamp
  j date
  k time

r1 := Filter[r0]
  r0.f > 0

r2 := Sort[r1]
  asc r1.f

r3 := Project[r2]
  a: r2.a
  b: r2.b
  f: r2.f

Add(a, b): r3.a + r3.b
```

now we can render the expression representation as 

```
alltypes := UnboundTable: alltypes
   a int8
   b int16
   c int32
   d int64
   e float32
   f float64
   g string
   h boolean
   i timestamp
   j date
   k time

 filtered := Filter[alltypes]
   alltypes.f > 0

 ordered := Sort[filtered]
   asc filtered.f

 projected := Project[ordered]
   a: ordered.a
   b: ordered.b
   f: ordered.f

 Add(a, b): projected.a + projected.b
```

This removes one level of indirection when we inspect an expression. Also not just tables but value nodes assigned to local variables are shown.

Note that it has its limitation based on where `repr()` is called, but `ibis.expr.format.pretty` now accepts a `scope` argument where the variable names can be passed.

### DIFF
--- a/ibis/backends/tests/test_interactive.py
+++ b/ibis/backends/tests/test_interactive.py
@@ -84,14 +84,6 @@ def test_interactive_non_compilable_repr_does_not_fail(table):
     repr(table.string_col.topk(3))
 
 
-def test_histogram_repr_no_query_execute(table, queries):
-    tier = table.double_col.histogram(10).name("bucket")
-    expr = table.group_by(tier).size()
-    expr._repr()
-
-    assert not queries
-
-
 def test_isin_rule_suppressed_exception_repr_not_fail(table):
     bool_clause = table["string_col"].notin(["1", "4", "7"])
     expr = table[bool_clause]["string_col"].value_counts()
--- a/ibis/expr/format.py
+++ b/ibis/expr/format.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
 import functools
+import inspect
 import itertools
 import textwrap
 import types
 from collections.abc import Mapping, Sequence
+from typing import Optional
 
 from public import public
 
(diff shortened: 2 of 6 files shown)


ibis/expr/types/core.py
```diff
@@ -76,35 +76,41 @@ def __coerce__(cls, value):
...
-                ]
-                return "\n".join(lines)
-        return capture.get().rstrip()
+        from ibis.expr.format import pretty
+
+        if opts.repr.show_variables:
+            import inspect
+
+            # get the variables available in the caller's frame
+            frame = inspect.currentframe().f_back
```
cpcloud, line 85: Good times.

ibis/backends/tests/test_interactive.py
```diff
@@ -84,12 +84,11 @@ def test_interactive_non_compilable_repr_does_not_fail(table):
...
-    tier = table.double_col.histogram(10).name("bucket")
-    expr = table.group_by(tier).size()
-    expr._repr()
-
-    assert not queries
+# def test_histogram_repr_no_query_execute(table, queries):
+#     tier = table.double_col.histogram(10).name("bucket")
+#     expr = table.group_by(tier).size()
+#     repr(expr)
+#     assert not queries
```
kszucs, line 91: This test case doesn't make much sense since the `queries` fixture sets the interactive repr on. 

I disabled the test case because the expression fails to execute, so we need a follow-up for this.

kszucs:
Requires a follow-up ticket to document this.

ibis/expr/types/core.py
kszucs: While this may not be supported in certain python implementations, I don't think that we need to be foolproof here since this functionality is entirely opt-in.

ibis/expr/tests/test_format.py
```diff
@@ -417,3 +417,24 @@ def shape(self):
...
+
+    filtered = alltypes[alltypes.f > 0]
+    ordered = filtered.order_by("f")
+    projected = ordered[["a", "b", "f"]]
+
+    add = projected.a + projected.b
+    sub = projected.a - projected.b
+    expr = add * sub
+
+    result = fmt(expr)
```
cpcloud, line 433: What is the repr when you have an expression chained together, without intermediate variables, when `show_variables` is `True`?

ibis/backends/tests/test_interactive.py
cpcloud: Well, the point was to have a test for the code path that never executes, but you've combined that all into `__repr__`.

If this was the only problematic test I'd say just delete it.

ibis/expr/types/core.py
cpcloud: Seems like this isn't sufficient for an IPython session:

```
In [1]: import ibis

In [2]: t = ibis.table({"a": "int", "b": "string"})

In [3]: tt = t.mutate(x=1000000000000000000)

In [4]: ttt = tt.mutate(y="s")

In [5]: tt
Out[5]:
r0 := UnboundTable: unbound_table_0
  a int64
  b string

Project[r0]
  a: r0.a
  b: r0.b
  x: 1000000000000000000

In [6]: ttt
Out[6]:
r0 := UnboundTable: unbound_table_0
  a int64
  b string

r1 := Project[r0]
  a: r0.a
  b: r0.b
  x: 1000000000000000000

Project[r1]
  a: r1.a
  b: r1.b
  x: r1.x
  y: 's'
```

cpcloud (approved):
Seems fine, but doesn't really work with IPython at the moment.

I'm a bit wary of testing this until we work out more of the kinks given how niche the feature is.

ibis/backends/tests/test_interactive.py
kszucs: Removed.

ibis/expr/format.py
```diff
@@ -145,43 +146,61 @@ def inline_args(fields, prefer_positional=False):
     return ", ".join(f"{k}={v}" for k, v in fields.items())
 
 
+def find_defining_frame(node):
+    for frame_info in inspect.stack()[::-1]:
+        for var in frame_info.frame.f_locals.values():
+            if isinstance(var, ir.Expr) and var.op() == node:
```
cpcloud, line 152: The equality check feels like it will cause a bunch of headaches when debugging very large expressions.

Just want to call that out as a potential pitfall here.

kszucs: It is an advanced opt-in configuration mostly going to be used by us. If it turns out to be stable we can build more advanced UX features assisting the users.

ibis/expr/tests/test_format.py
cpcloud: This works as it did previously it seems.

PR #8633, opened 2024-03-12
fix(sql): look for CTEs under value expressions as well

kszucs:
Common table expressions haven't been extracted for subqueries because we stopped traversing once reaching any value expression.

### DIFF
--- a/ibis/backends/sql/rewrites.py
+++ b/ibis/backends/sql/rewrites.py
@@ -17,7 +17,7 @@ from ibis.common.annotations import attribute
 from ibis.common.collections import FrozenDict  # noqa: TCH001
 from ibis.common.deferred import var
 from ibis.common.graph import Graph
-from ibis.common.patterns import Object, Pattern, _, replace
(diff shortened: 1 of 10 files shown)


cpcloud:
Would this also address https://github.com/ibis-project/ibis/issues/8484?

kszucs:
> Would this also address #8484?

I doubt because there are no subqueries, it could if we artificially introduce subqueries to not repeat tedious value expressions.

cpcloud:
@kszucs Is this ready for review?

kszucs:
I need to add test cases for the referenced issues.

kszucs:
> @kszucs Is this ready for review?

Yes. I will add test cases for the relevant `map` issues in a follow-up.

kszucs:
> > @kszucs Is this ready for review?
> 
> Yes. I will add test cases for the relevant `map` issues in a follow-up.

Turns out it won't resolve those since those subqueries don't occur more than once.

cpcloud (approved):

Issue #8639, opened 2024-03-13
docs: quick start at altair plot fails


raybellwaves:
### Please describe the issue

Firstly, thanks for this project!

I'm going through the quick start at https://ibis-project.org/ at got to the charting section. The altair plot fails as I get `NameError: name 'width' is not defined`

hard coding `width` and `height` will help users continue running through all sections.

![Screenshot 2024-03-12 at 8 52 55 PM](https://github.com/ibis-project/ibis/assets/17162724/fc9eacff-9b98-4e65-9c51-47408a8dc848)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

lostmygithubaccount:
they are defined in the code fold above, but agreed this could be confusing for someone running through it for the first time. thanks for opening!

<img width="834" alt="image" src="https://github.com/ibis-project/ibis/assets/54814569/4df3e5d3-e09e-4265-aa26-ece293bb0d18">

https://github.com/ibis-project/ibis/blob/main/docs/index.qmd#L150-L154

PR #8640, opened 2024-03-13
docs: add constants to width and height

raybellwaves:
Fixes https://github.com/ibis-project/ibis/issues/8639


lostmygithubaccount (changes requested):
would you mind also removing these lines to avoid any confusion? 

edit: whoops I messed up, the lines in the code fold above that set the height and width

raybellwaves:
> would you mind also removing these lines to avoid any confusion?
> 
> edit: whoops I messed up, the lines in the code fold above that set the height and width

Not sure if I follow.

cpcloud (approved):
LGTM, thanks!

Issue #8632, opened 2024-03-12
feat: add workarounds for duckdb MAP bugs


NickCrews:
### Is your feature request related to a problem?

see
https://github.com/duckdb/duckdb/issues/11115
and
https://github.com/duckdb/duckdb/issues/11116

This might be out of scope, and we should just wait for duckdb to fix these upstream, but we could add some workarounds for these bugs in ibis. I am running into these and am having to implement my own little wrappers, eg

```python
def _map_values(m: ir.MapValue) -> ir.ArrayValue:
    """workaround for https://github.com/duckdb/duckdb/issues/11116"""
    normal = m.values()
    null = ibis.literal(None, type=dt.Array(value_type=m.type().value_type))
    return m.isnull().ifelse(null, normal)
```

feel free to close this out if you don't want to take this on (I don't blame you if you don't 😉 )

### Describe the solution you'd like

I'm not sure where in the stack these fixes would need to go, at compile time, or just when creating expressions.

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
Thanks! Looks like you've gone down a map 🐰 🕳️ 

I think it's reasonable to add these workarounds for now.

NickCrews:
yeah seriously talk about a 🐰 🕳️ . I was convinced for hours it was bugs in my code wheeeeee

here is what I currently use. I have unit tests for functions that use these, but I don't have unit tests for these as individuals.

```python
def map_keys(m: ir.MapValue) -> ir.ArrayValue:
    """workaround for https://github.com/duckdb/duckdb/issues/11116"""
    normal = m.keys()
    null = ibis.literal(None, type=dt.Array(value_type=m.type().key_type))
    return m.isnull().ifelse(null, normal)


def map_values(m: ir.MapValue) -> ir.ArrayValue:
    """workaround for https://github.com/duckdb/duckdb/issues/11116"""
    normal = m.values()
    null = ibis.literal(None, type=dt.Array(value_type=m.type().value_type))
    return m.isnull().ifelse(null, normal)


def map_(keys: ir.ArrayValue, values: ir.ArrayValue) -> ir.MapValue:
    """workaround for https://github.com/duckdb/duckdb/issues/11115"""
    either_null = keys.isnull() | values.isnull()
    regular = ibis.map(keys, values)
    null = ibis.literal(
        None,
        type=dt.Map(
[...]

PR #8649, opened 2024-03-13
fix(duckdb): workaround for duckdb Map NULL bugs

NickCrews:
partially addresses https://github.com/ibis-project/ibis/issues/8632

### DIFF
--- a/ibis/backends/duckdb/tests/test_client.py
+++ b/ibis/backends/duckdb/tests/test_client.py
@@ -297,3 +297,11 @@ def test_list_tables_schema_warning_refactor(con):
 
     assert con.list_tables(database="shops") == icecream_table
     assert con.list_tables(database=("shops",)) == icecream_table
+
+
+@pytest.mark.xfail(
+    raises=(duckdb.InvalidInputException, duckdb.BinderException),
+    reason="https://github.com/ibis-project/ibis/issues/8632",
+)
+def test_map_null_workaround(con):
+    con.sql("SELECT MAP([0, 1, 2], NULL::INT[])").to_pyarrow()
--- a/ibis/backends/postgres/compiler.py
+++ b/ibis/backends/postgres/compiler.py
@@ -385,7 +385,10 @@ class PostgresCompiler(SQLGlotCompiler):
         )
 
     def visit_Map(self, op, *, keys, values):
-        return self.f.map(self.f.array(*keys), self.f.array(*values))
+        # map(["a", "b"], NULL) results in {"a": NULL, "b": NULL} in regular postgres,
+        # so we need to modify it to return NULL instead
+        regular = self.f.map(keys, values)
+        return self.if_(values.is_(NULL), NULL, regular)
 
     def visit_MapLength(self, op, *, arg):
         return self.f.cardinality(self.f.akeys(arg))
--- a/ibis/backends/tests/test_map.py
+++ b/ibis/backends/tests/test_map.py
@@ -10,7 +10,7 @@ from pytest import param
 import ibis
 import ibis.common.exceptions as exc
 import ibis.expr.datatypes as dt
-from ibis.backends.tests.errors import Py4JJavaError
+from ibis.backends.tests.errors import PsycoPg2InternalError, Py4JJavaError
 
 pytestmark = [
     pytest.mark.never(
@@ -37,6 +37,28 @@ mark_notimpl_risingwave_hstore = pytest.mark.notimpl(
     ["risingwave"],
     reason="function hstore(character varying[], character varying[]) does not exist",
 )
+
+
+@pytest.mark.notyet("clickhouse", reason="nested types can't be NULL")
+@pytest.mark.broken(["pandas", "dask"], reason="TypeError: iteration over a 0-d array")
+@pytest.mark.notimpl(
+    ["risingwave"],
+    raises=PsycoPg2InternalError,
+    reason="function hstore(character varying[], character varying[]) does not exist",
+)
+@pytest.mark.parametrize(
+    ("k", "v"),
+    [
+        param(["a", "b"], None, id="null_values"),
+        param(None, ["c", "d"], id="null_keys"),
+        param(None, None, id="null_both"),
+    ],
+)
+def test_map_nulls(con, k, v):
+    k = ibis.literal(k, type="array<string>")
+    v = ibis.literal(v, type="array<string>")
+    m = ibis.map(k, v)
+    assert con.execute(m) is None
 
 
 @pytest.mark.notimpl(["pandas", "dask"])
--- a/ibis/backends/duckdb/compiler.py
+++ b/ibis/backends/duckdb/compiler.py
@@ -199,6 +199,12 @@ class DuckDBCompiler(SQLGlotCompiler):
         any_arg_null = sg.or_(*(arr.is_(NULL) for arr in arg))
         return self.if_(any_arg_null, NULL, zipped_arrays)
 
+    def visit_Map(self, op, *, keys, values):
+        # workaround for https://github.com/ibis-project/ibis/issues/8632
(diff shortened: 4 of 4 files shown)


NickCrews:
Once that xfail test starts passing, then we could
- check for the duckdb version, and if we are running a broken version, then do the workaround, otherwise short-circuit to the simple path
- something else?

NickCrews:
errr, this is probably the wrong way around this, I should be fixing this at the duckdb level, not the global ibis level. Otherwise we miss MapValues that are created via other APIs, like ibis.literal()

ibis/backends/tests/test_map.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import duckdb
```
cpcloud, line 3: You cannot import this here. The tests in `ibis/backends/tests` must be runnable by arbitrary backends without having the dependencies for all backends installed.

For example, let's say I have the polars backend dependencies installed, but not the duckdb backend dependencies and I want to run the backend test suite using polars. Importing `duckdb` here (or the the dependencies for any non-polars backend) will immediately fail with an `ImportError`.

If you need to do something specific to a backend, put that code in `ibis/backend/${specific_backend}/tests`.

ibis/backends/tests/test_map.py
```diff
@@ -26,6 +28,34 @@
...
+)
+def test_map_nulls(con, kv):
+    k, v = kv
+    k = ibis.literal(k, type="array<string>")
+    v = ibis.literal(v, type="array<string>")
+    m = ibis.map(k, v)
+    assert con.execute(m) is None
+
+
+@pytest.mark.xfail(
```
cpcloud, line 49: Move this test to `ibis/backends/tests/test_client.py` or if there's another file that makes sense, then there.

ibis/expr/types/structs.py
```diff
@@ -75,6 +75,11 @@ def struct(
     """
     import ibis.expr.operations as ops
 
+    if isinstance(value, StructValue):
+        if type is not None:
+            return value.cast(type)
+        return value
```
cpcloud, line 81: This functionality seems unrelated to the original issue and needs tests. Can you add this feature in a follow up PR?

ibis/expr/types/arrays.py
```diff
@@ -1131,4 +1131,6 @@ def array(values: Iterable[V]) -> ArrayValue:
     │ [3, 47]              │
     └──────────────────────┘
     """
+    if isinstance(values, ir.ArrayValue):
+        return values
```
cpcloud, line 1135: This functionality seems unrelated to the original issue and needs tests. Can you add this feature in a follow up PR?

ibis/expr/types/maps.py
```diff
@@ -486,6 +484,8 @@ def map(
     │ {'b': 3}             │
     └──────────────────────┘
     """
+    if isinstance(keys, MapValue):
+        return keys
```
cpcloud, line 488: This functionality seems unrelated to the original issue and needs tests. Can you add this feature in a follow up PR?

cpcloud (changes requested):

ibis/expr/types/structs.py
NickCrews: Oops this was originally needed but then I adjusted the implementation and never reverted this. Will do. I think it's a good change though.

NickCrews: moved to #8666

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -276,3 +276,12 @@ def test_invalid_connect(tmp_path):
     url = f"duckdb://{tmp_path}?read_only=invalid_value"
     with pytest.raises(ValueError):
         ibis.connect(url)
+
+
+@pytest.mark.xfail(
+    raises=(duckdb.InvalidInputException, duckdb.BinderException),
```
cpcloud, line 282: Why are there two exceptions here?

ibis/backends/postgres/compiler.py
```diff
@@ -330,7 +330,10 @@ def visit_ToJSONArray(self, op, *, arg):
         )
 
     def visit_Map(self, op, *, keys, values):
-        return self.f.map(self.f.array(*keys), self.f.array(*values))
+        # map(["a", "b"], NULL) results in {"a": NULL, "b": NULL} in regular postgres,
```
cpcloud, line 333: Wow, that's ... weird.

ibis/backends/tests/test_map.py
```diff
@@ -26,6 +26,29 @@
...
+@pytest.mark.parametrize(
+    "kv",
+    [
+        param((["a", "b"], None), id="null_values"),
+        param((None, ["c", "d"]), id="null_keys"),
+        param((None, None), id="null_both"),
+    ],
+)
+def test_map_nulls(con, kv):
+    k, v = kv
```
cpcloud, line 45:
```suggestion
    ("k", "v"),
    [
        param(["a", "b"], None, id="null_values"),
        param(None, ["c", "d"], id="null_keys"),
        param(None, None, id="null_both"),
    ],
)
def test_map_nulls(con, k, v):
```

Let pytest do the work for you :)

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -276,3 +276,12 @@ def test_invalid_connect(tmp_path):
...
+    reason="https://github.com/ibis-project/ibis/issues/8632",
+)
+def test_map_null_workaround_xfail():
+    conn = duckdb.connect()
+    conn.sql("SELECT MAP([0, 1, 2], NULL::INT[]);").fetchone()
```
cpcloud, line 287:
```suggestion
def test_map_null_workaround_xfail(con):
    con.con.sql("SELECT MAP([0, 1, 2], NULL::INT[]);").fetchall()
```

ibis/backends/duckdb/tests/test_client.py
NickCrews: Depending on the python/duckdb version you get different errors, was running into both in CI. I wasn't able to actually figure out the exact pattern, I couldnt repro one of the errors locally.  You want a comment for this? Actually figure it out?

ibis/backends/postgres/compiler.py
NickCrews: I didn't look more into this, but I'm wondering if this allows you to use Maps as a HashSet, which actually is pretty cool. Maybe we even want to change ibis semantics to allow this?? Idk haven't thought how that would interact with other backends

ibis/backends/duckdb/tests/test_client.py
NickCrews: thanks, done.

ibis/backends/tests/test_map.py
NickCrews: thanks, done.

ibis/backends/duckdb/tests/test_client.py
NickCrews: OK I pushed with this xfail commented out so you can see [CI failures](https://github.com/ibis-project/ibis/actions/runs/8381568818?pr=8649):

- "duckdb + GeoSpatial *" tests raises BinderErrors
- the other duckdb tests raise InvalidInputExceptions

NickCrews: @cpcloud gentle nudge here :)

cpcloud: Sorry I missed this!

Totally fine to keep both exception types.

ibis/backends/postgres/compiler.py
cpcloud: interesting idea, but let's leave it for later :)

cpcloud (approved):

PR #8651, opened 2024-03-13
docs: added section on duckdb reading gcs files

pieter-factful:
## Description of changes

On Zulip I was supported in reading a jsonl file stored on GCS with duckdb. With more people likely te encounter the same issue, it was suggested to add it to the docs. It was suggested to add a new page to the 'how-to' section. The 'input/output' section seemed a good fit.


ncclementi:
@pieter-factful looks like you are missing some `pre-commits` that will take care of the linting if you run them . 

Here are some instructions: https://ibis-project.org/contribute/03_style

But in few words, in your conda environment you can do this, that will make sure to run them before you commit, and in most cases it will fix the files, then you can add them again.  

```sh
$ pip install pre-commit
$ pre-commit install 
```

ncclementi:
Whoops, @kszucs beat me for few seconds, I didn't see the comment.

docs/how-to/input-output/gcs_duckdb.qmd
```diff
@@ -0,0 +1,24 @@
...
+gcs = fsspec.filesystem("gcs")
+
+con.register_filesystem(gcs)
+
+t = con.read_json(<URL>)
+```
+
+Where `<URL>` is the url provided by GCS. Note that:
+- For private files use the URL with prefix `gs://` 
+- For public files you can use the URL with prefix `gcs://`
```
cpcloud, line 24:
```suggestion
Where `<URL>` is the url provided by GCS.
```

Both `gs://` and `gcs://` should work regardless of permissions. Is that not the case for you?

pieter-factful: I decided to leave out the note altogther. I didn't want to include the specific URL. And figured it unncessary to specifiy the prefix. 

As for the access, I checked and yes, looks like all 3 (gs/gcs/storage.googleapis.com all work. Odd, I ran into issues the other day. Maybe mixed up the files: I had both a public and a private version of the file.

pieter-factful:
@cpcloud @ncclementi @kszucs Looks like all checks passed - thanks for your help!

docs/how-to/input-output/gcs_duckdb.qmd
```diff
@@ -0,0 +1,23 @@
...
+con = ibis.duckdb.connect()
+
+import fsspec
+import gcsfs
+
+gcs = fsspec.filesystem("gcs")
+
+con.register_filesystem(gcs)
+
+#t = con.read_json(<URL)
```
cpcloud, line 19:
```suggestion
t = con.read_json("gs://...")
```

docs/how-to/input-output/gcs_duckdb.qmd
```diff
@@ -0,0 +1,23 @@
+---
+title: Loading Google Cloud Storage files with DuckDB
+---
+
+Loading data stored on Google Cloud Storage to DuckDB requires specifying `gcs` as filesystem.
+
+```{python}
```
cpcloud, line 7:
````suggestion
```python
````

When you wrap `python` in braces like this, quarto will execute the code block. Here we **don't** want to execute because you're not actually specifying a URL.

**Review by cpcloud — CHANGES_REQUESTED** [review]

File: docs/how-to/input-output/gcs_duckdb.qmd [hunk 4]
```diff
@@ -0,0 +1,23 @@
...
+
+import fsspec
+import gcsfs
+
+gcs = fsspec.filesystem("gcs")
+
+con.register_filesystem(gcs)
+
+t = con.read_json("gs://...")
+
```
cpcloud, line 20:
```suggestion
```

docs/how-to/input-output/gcs_duckdb.qmd
```diff
@@ -0,0 +1,23 @@
...
+```
+
+Where `<URL>` is the url provided by GCS.
```
cpcloud, line 23:
```suggestion
```

cpcloud (approved):

Issue #8650, opened 2024-03-13
bug: `Array.zip(NULL)` gives `[{f1: ..., f2:NULL}, {...}, ...]`, not NULL


NickCrews:
### What happened?

```python
import ibis
# the .map is workaround for https://github.com/ibis-project/ibis/issues/8641
a = ibis.literal([1, 2, 3], type="array<int64>").map(ibis._)
b = ibis.literal(None, type="array<int64>")
null_res = a.zip(b).execute()
assert null_res is None
```

### What version of ibis are you using?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8652, opened 2024-03-13
fix(api): return NULL when NULL is passed to `Array.zip`

NickCrews:
fixes https://github.com/ibis-project/ibis/issues/8650

### DIFF
--- a/ibis/backends/duckdb/compiler.py
+++ b/ibis/backends/duckdb/compiler.py
@@ -187,7 +187,7 @@ class DuckDBCompiler(SQLGlotCompiler):
             ]
         )
         func = sge.Lambda(this=body, expressions=[i])
-        regular = self.f.list_apply(
+        zipped_arrays = self.f.list_apply(
             self.f.range(
                 1,
                 # DuckDB Range excludes upper bound
@@ -196,8 +196,8 @@ class DuckDBCompiler(SQLGlotCompiler):
             func,
         )
         # if any of the input arrays in arg are NULL, the result is NULL
-        any_arg_null = reduce(lambda x, y: x | y.is_(NULL), arg, False)
-        return self.if_(any_arg_null, NULL, regular)
+        any_arg_null = sg.or_(*(arr.is_(NULL) for arr in arg))
+        return self.if_(any_arg_null, NULL, zipped_arrays)
 
     def visit_MapGet(self, op, *, arg, key, default):
         return self.f.ifnull(
--- a/ibis/backends/duckdb/compiler.py
+++ b/ibis/backends/duckdb/compiler.py
@@ -187,7 +187,7 @@ class DuckDBCompiler(SQLGlotCompiler):
             ]
         )
         func = sge.Lambda(this=body, expressions=[i])
-        return self.f.list_apply(
+        regular = self.f.list_apply(
             self.f.range(
                 1,
                 # DuckDB Range excludes upper bound
@@ -195,6 +195,9 @@ class DuckDBCompiler(SQLGlotCompiler):
             ),
             func,
         )
+        # if any of the input arrays in arg are NULL, the result is NULL
+        any_arg_null = reduce(lambda x, y: x | y.is_(NULL), arg, False)
+        return self.if_(any_arg_null, NULL, regular)
 
     def visit_MapGet(self, op, *, arg, key, default):
         return self.f.ifnull(
--- a/ibis/backends/tests/test_array.py
+++ b/ibis/backends/tests/test_array.py
@@ -850,6 +850,15 @@ def test_zip(backend):
     assert len(s[0][0]) == len(res.type().value_type)
     assert len(x[0]) == len(s[0])
 
+    # clickhouse nested types can't be null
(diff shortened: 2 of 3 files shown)


ibis/backends/duckdb/compiler.py
```diff
@@ -187,14 +187,17 @@ def visit_ArrayZip(self, op, *, arg):
...
+        regular = self.f.list_apply(
             self.f.range(
                 1,
                 # DuckDB Range excludes upper bound
                 self.f.greatest(*map(self.f.len, arg)) + 1,
             ),
             func,
         )
+        # if any of the input arrays in arg are NULL, the result is NULL
+        any_arg_null = reduce(lambda x, y: x | y.is_(NULL), arg, False)
```
NickCrews, line 199: this `reduce` was an uniformed acceptance from a github copilot suggestion (that worked, much wow), so please verify this is a good way to accomplish this.

ibis/backends/duckdb/compiler.py
```diff
@@ -187,14 +187,17 @@ def visit_ArrayZip(self, op, *, arg):
             ]
         )
         func = sge.Lambda(this=body, expressions=[i])
-        return self.f.list_apply(
+        regular = self.f.list_apply(
             self.f.range(
```
NickCrews, line 191: could instead put the null check up here, so that the input to `list_apply` is NULL. IDK if that is any more performant...

ibis/backends/tests/test_array.py
```diff
@@ -850,6 +850,15 @@ def test_zip(backend):
     assert len(s[0][0]) == len(res.type().value_type)
     assert len(x[0]) == len(s[0])
 
+    # clickhouse nested types can't be null
+    if backend.name() != "clickhouse":
```
NickCrews, line 854: was lazy about using pytest connection fixtures, etc. I can split this up better if you want, but this isn't that ugly to me.

ibis/expr/types/arrays.py
```diff
@@ -930,43 +930,41 @@ def zip(self, other: ArrayValue, *others: ArrayValue) -> ArrayValue:
         Array
             Array of structs where each struct field is an element of each input
             array.
+            The fields are named `f1`, `f2`, `f3`, etc.
```
cpcloud, line 933: nit: can you put this on the same line as the previous sentence?

ibis/backends/duckdb/compiler.py
cpcloud: > please verify this is a good way to accomplish this

lol at AI

```suggestion
        any_arg_null = reduce(lambda x, y: x | y.is_(NULL), arg, FALSE)
```

The `FALSE` may need to be imported

cpcloud (changes requested):

ibis/backends/duckdb/compiler.py
NickCrews: `sg.or_()` actually is what we want I think

NickCrews:
@cpcloud thanks for the review! all good suggestions, addressed them all. Check that you like the way I split out the new test, reusing a single decorator to avoid repeating myself.

ibis/backends/tests/test_array.py
```diff
@@ -818,8 +818,7 @@ def test_unnest_struct(con):
     tm.assert_series_equal(result, expected)
 
 
-@builtin_array
-@pytest.mark.notimpl(
+ARRAY_ZIP_NOTIMPL = pytest.mark.notimpl(
```
cpcloud, line 821:
```suggestion
array_zip_notimpl = pytest.mark.notimpl(
```

ibis/backends/tests/test_array.py
```diff
@@ -833,6 +832,10 @@ def test_unnest_struct(con):
     ],
     raises=com.OperationNotDefinedError,
 )
+
+
+@builtin_array
+@ARRAY_ZIP_NOTIMPL
```
cpcloud, line 838:
```suggestion
@array_zip_notimpl
```

ibis/backends/tests/test_array.py
```diff
@@ -851,6 +854,22 @@ def test_zip(backend):
     assert len(x[0]) == len(s[0])
 
 
+@builtin_array
+@ARRAY_ZIP_NOTIMPL
```
cpcloud, line 858:
```suggestion
@array_zip_notimpl
```

cpcloud (approved):
LGTM, just some suggestions that I will add via the GitHub UI.

cpcloud:
Interesting, looks like the backends that support zip (except clickhouse) all have this behavior by default.

cpcloud:
That said, the DuckDB implementation isn't calling a specific duckdb `zip` function, so it's kind of up to us to make it behave consistently.

ibis/backends/duckdb/compiler.py
cpcloud: I think it's fine as is.

cpcloud (approved):
Thanks!

Issue #8109, opened 2024-01-26
feat(duckdb): add support for reading tables from MySQL


jcrist:
DuckDB now has support for reading MySQL tables using the `mysql` extension. To match our existing `con.read_sqlite`/`con.read_postgres` methods we should add a new `con.read_mysql` method.

References:
- https://duckdb.org/2024/01/26/multi-database-support-in-duckdb.html
- https://duckdb.org/docs/extensions/mysql

PR #8656, opened 2024-03-14
feat(duckdb): add support for read_mysql

ncclementi:
- Closes https://github.com/ibis-project/ibis/issues/8109

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import ast
 import contextlib
 import os
+import urllib
 import warnings
 from operator import itemgetter
 from pathlib import Path
@@ -948,7 +949,6 @@ class Backend(SQLBackend, CanCreateSchema, UrlFromPath):
             )
             .sql(self.name, pretty=True)
         )
-
         out = self.con.execute(sql).fetch_arrow_table()
 
         return self._filter_with_like(out[col].to_pylist(), like)
@@ -987,6 +987,47 @@ class Backend(SQLBackend, CanCreateSchema, UrlFromPath):
         )
 
         return self.table(table_name)
+
+    def read_mysql(
+        self,
+        *,
+        uri: str,
+        catalog: str,
+        table_name: str | None = None,
+    ) -> ir.Table:
+        """Register a table from a MySQL instance into a DuckDB table.
+
+        Parameters
+        ----------
+        uri
+            A mysql URI of the form `mysql://user:password@host:port/database`
+        catalog
+            User-defined alias given to the MySQL database that is being attached
+            to DuckDB
+        table_name
+            The table to read
+
+        Returns
+        -------
+        ir.Table
+            The just-registered table.
+        """
+
+        parsed = urllib.parse.urlparse(uri)
+
+        if table_name is None:
+            raise ValueError("`table_name` is required when registering a mysql table")
+
+        self._load_extensions(["mysql"])
+
+        database = parsed.path.strip("/")
+
+        query_con = f"""ATTACH 'host={parsed.hostname} user={parsed.username} password={parsed.password} port={parsed.port} database={database}' AS {catalog} (TYPE mysql)"""
+
+        with self._safe_raw_sql(query_con):
+            pass
+
+        return self.table(table_name, schema=database, database=catalog)
 
     def read_sqlite(self, path: str | Path, table_name: str | None = None) -> ir.Table:
         """Register a table from a SQLite database into a DuckDB table.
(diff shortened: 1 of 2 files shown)


ibis/backends/duckdb/tests/test_register.py
```diff
@@ -184,6 +184,27 @@ def test_read_postgres(con, pgurl):  # pragma: no cover
...
+    s = ibis.schema(dict(x="float64", y="str"))
+
+    mysqlcon.create_table("duckdb_test", df, schema=s, overwrite=True)
+    yield mysqlcon.con
+    mysqlcon.drop_table("duckdb_test", force=True)
+
+
+def test_read_mysql(con, mysqlurl):  # pragma: no cover
+    table = con.read_mysql(
+        f"mysql://{mysqlurl.user.decode()}:{mysqlurl.password.decode()}@{mysqlurl.host}:{mysqlurl.port}/ibis_testing",
```
ncclementi, line 201: @cpcloud  here` mysqlurl.host = localhost` and it looks liek when this goes into the `query_con` that gets run with `_safe_raw_sql` in `read_mysql` in my mac it complains, while if I provide `127.0.0.1` I have no problems.

ncclementi:
Well looks like we triggered the `"localhost"` issue in CI too see failure [here](https://github.com/ibis-project/ibis/actions/runs/8286659965/job/22677178632?pr=8656#step:15:297)

```python-traceback
E               pymysql.err.OperationalError: (2003, "Can't connect to MySQL server on 'localhost' ([Errno 111] Connection refused)")

../../../.cache/pypoetry/virtualenvs/ibis-framework-2c2VoWo5-py3.9/lib/python3.9/site-packages/pymysql/connections.py:711: OperationalError
```

cpcloud:
I opened duckdb/duckdb_mysql#44 to track the localhost issue.

ibis/backends/duckdb/__init__.py
```diff
@@ -988,6 +988,47 @@ def read_postgres(
 
         return self.table(table_name)
 
+    ########
+    def read_mysql(
+        self,
+        uri: str,
+        catalog: str,
```
ncclementi, line 995: @gforsyth I don't know if this is the best choice of word. Talking to Phillip he mentioned that these aliases are registered as catalogs, hence the choice, but feel free to suggest something different.

ncclementi:
I understand why the DuckDB job fails, we are not bring the container up, we can do something similar to what we are doing in `read_postgres` by setting up an env variable and not run the test in CI. 

But I'm confused about why the nix one is failing, help appreciated.

ibis/backends/duckdb/__init__.py
cpcloud: `catalog` seems like right thing here given the direction we're going with naming those sorts of things.

ibis/backends/duckdb/__init__.py
```diff
@@ -988,6 +988,46 @@ def read_postgres(
 
         return self.table(table_name)
 
+    def read_mysql(
+        self,
+        uri: str,
```
cpcloud, line 993: Can we make all of these arguments required-keyword?

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -177,13 +177,48 @@ def pgurl():  # pragma: no cover
...
+    s = ibis.schema(dict(x="float64", y="str"))
+
+    mysqlcon.create_table("duckdb_test", df, schema=s, overwrite=True)
+    yield mysqlcon.con
+    mysqlcon.drop_table("duckdb_test", force=True)
+
+
+@pytest.mark.skipif(
+    os.environ.get("DUCKDB_MYSQL") is None, reason="avoiding CI shenanigans"
+)
```
cpcloud, line 204: I'll put up a PR to do this with markers. That way we can avoid mystery environment variables.

cpcloud (approved):
LGTM!

ibis/backends/duckdb/tests/test_register.py
cpcloud: Hm nevermind, this doesn't seem worth the effort right now.

ibis/backends/duckdb/__init__.py
ncclementi: Do you mean not, by not allowing table to be None, and trigger the Value Error later? 

Should I remove that from the postgres and sqlite cases too?

ncclementi:
So, this was fixed https://github.com/duckdb/duckdb_mysql/issues/44, but we can't use it unless we install the extension from the nightlies, so we'll have to wait a bit.

PR #8660, opened 2024-03-15
test(impala): ensure that topk table does not already exist

cpcloud:
Ensure that the topk table gets created anew every time the test suite is run for Impala.


gforsyth:
@cpcloud I believe the Flink issues are because apache-flink 1.19.0 was just released.  I guess we need an upper pin?

jcrist (approved):

PR #8662, opened 2024-03-15
chore(deps): pin flink < 1.19.0 for ci

gforsyth:


jcrist (approved):
Pinning makes sense for now, thanks for taking care of this.

PR #8664, opened 2024-03-15
feat(api): add `ibis.today()` for retrieving the current date

jcrist:
Responding to this user request: https://github.com/ibis-project/ibis/discussions/8213#discussioncomment-8798497

Also adds support for `ibis.now()` for `oracle` and `druid`. I assume these were marked as unsupported due to prior upstream issues in sqlglot? The compilation seems to work fine now.

PR #8665, opened 2024-03-15
fix(polars): ensure `t.select(col=scalar)` results in `len(t)` rows

jcrist:
Previously broadcasting a scalar as a column and then selecting only that column would result in only one row (rather than `len(t)` rows). This fixes that bug and adds a test.

Based on #8664, since this fixes a test in that PR too.


cpcloud (approved):
Thanks!

Issue #8289, opened 2024-02-08
feat: Support `type` arg to `ibis.array` and `ibis.map`


NickCrews:
### Is your feature request related to a problem?

This is already supported for `ibis.struct`.

I want to create an array of type `array<uint8>`. `ibis.array()` doesn't support a type param, so if the type can't be inferred from the values, I can't use it:

- If I want a NULL value, I have to use `ibis.literal(None, type="array<uint8>")`. 
- If I want length-0 value, I have to use `ibis.literal([], type="array<uint8>")`
- For longer-length arrays, I am able use `ibis.array` for the initial construction, but then I have to do a .cast() afterwards to ensure the type is correct: `ibis.array([1,2]).type()` shows `array<int8>`, so I have to do `ibis.array([1,2]).cast("array<uint8>")`

I haven't ever used map types, but I imagine it is a similar story.

It is annoying to use two different constructors depending on the length. There should be one way.

I could just use `ibis.literal`, but it's docstring says that constructing these complex types using it is going to be deprecated in a future release. If that is the case, then my workarounds have a shelf life. U

### Describe the solution you'd like

add a `type` param to these three functions. It should look just like it does for struct, I like that API.

Then, either:
- actually follow through with what the docstring says for literal and start the deprecation and removal timeline
- remove that note on future deprecation. This was added in [this commit](https://github.com/ibis-project/ibis/commit/edc955c43aec4ed72f6ba7c42e8d37d66b61809c#diff-ab1c80c5064c1f7db69704921149c6d9bb42f6dcb883aa4a3af0097a91eb5880), but I can't find a reasoning behind it.

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chloeh13q:
@NickCrews IIUC, your PR #8666 should also resolve this. If so, I will re-assign this issue to you.

NickCrews:
That would work. Also, that PR is turning into a bit of a doozy, if you wanted to help out there it would be appreciated!

chloeh13q:
@NickCrews Sure, I can take a look!

ncclementi:
@NickCrews I noticed there are multiple PRs open, draft or closed that were aiming to tackle this issue, but it's unclear what's the status. Would it be possible to leave a comment here on what are the steps needed to get this one across the line?

NickCrews:
[...]

PR #8666, opened 2024-03-15
feat: improve map(), struct(), array()

NickCrews:
fixes https://github.com/ibis-project/ibis/issues/8289

This does a lot of changes. It was hard for me to separate them out as I implemented them. But now that it's all hashed out, I can try to split this up into separate commits if you want. But that might be sorta hard in
some cases.

## Changes:

This is adding support for passing in None to all these constructors.
These use the new `ibis.null(<type>)` API to return `op.Literal(None, <type>)`s

Make these constructors idempotent: you can pass in existing Expressions into array(), etc.
The type argument for all of these now always has an effect, not just when passing in python literals. So basically it acts like a cast.

A big structural change is that now ops.Array has an optional
attribute "dtype", so if you pass in a 0-length sequence
of values the op still knows what dtype it is.

Several of the backends were always broken here, they just weren't getting caught. I marked them as broken, we can fix them in a followup.

You can test this locally with eg
`pytest -m <backend> -k factory ibis/backends/tests/test_array.py  ibis/backends/tests/test_map.py ibis/backends/tests/test_struct.py`

Also, fix executing Literal(None) on pandas and polars, 0-length arrays on polars

## Next steps

A possible nice follow-up would be to completely remove supporting structs, arrays, and maps inside ops.Literal.
Currently, when someone calls `ibis.array([1,2])`, that now results in ops.Array, but if they do `ibis.literal([1,2])`, it is represented as ops.Literal. It would be nice if inside ibis.literal() there was some logic like

```python
...
if isinstance(val, Mapping):
    return ibis.map(val)
if is_iterable(val):
    return ibis.array(val)
...
```

and then all the backends wouldn't have to even worry about supporting these nested types in `visit_Literal()`

### DIFF
--- a/ibis/backends/dask/helpers.py
+++ b/ibis/backends/dask/helpers.py
@@ -30,7 +30,7 @@ class DaskUtils(PandasUtils):
 
     @classmethod
     def asseries(cls, value, like=None):
-        """Ensure that value is a pandas Series object, broadcast if necessary."""
+        """Ensure that value is a dask Series object, broadcast if necessary."""
 
         if isinstance(value, dd.Series):
             return value
@@ -50,7 +50,7 @@ class DaskUtils(PandasUtils):
         elif isinstance(value, pd.Series):
             return dd.from_pandas(value, npartitions=1)
         elif like is not None:
-            if isinstance(value, (tuple, list, dict)):
+            if isinstance(value, (tuple, list, dict, np.ndarray)):
                 fn = lambda df: pd.Series([value] * len(df), index=df.index)
             else:
                 fn = lambda df: pd.Series(value, index=df.index)
--- a/ibis/backends/pandas/executor.py
+++ b/ibis/backends/pandas/executor.py
@@ -31,6 +31,7 @@ from ibis.backends.pandas.rewrites import (
 )
 from ibis.common.dispatch import Dispatched
 from ibis.common.exceptions import OperationNotDefinedError, UnboundExpressionError
+from ibis.formats.numpy import NumpyType
 from ibis.formats.pandas import PandasData, PandasType
 from ibis.util import any_of, gen_name
 
@@ -49,6 +50,8 @@ class PandasExecutor(Dispatched, PandasUtils):
 
     @classmethod
     def visit(cls, op: ops.Literal, value, dtype):
+        if value is None:
+            return None
         if dtype.is_interval():
             value = pd.Timedelta(value, dtype.unit.short)
         elif dtype.is_array():
@@ -220,8 +223,9 @@ class PandasExecutor(Dispatched, PandasUtils):
         return pd.Series(result, name=op.name)
 
     @classmethod
-    def visit(cls, op: ops.Array, exprs):
-        return cls.rowwise(lambda row: np.array(row, dtype=object), exprs)
+    def visit(cls, op: ops.Array, exprs, dtype):
+        np_type = NumpyType.from_ibis(dtype)
+        return cls.rowwise(lambda row: np.array(row, dtype=np_type), exprs)
 
     @classmethod
     def visit(cls, op: ops.StructColumn, names, values):
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -86,10 +86,14 @@ def _make_duration(value, dtype):
 def literal(op, **_):
     value = op.value
     dtype = op.dtype
+    if dtype.is_interval():
+        return _make_duration(value, dtype)
 
-    if dtype.is_array():
+    typ = PolarsType.from_ibis(dtype)
+    if value is None:
+        return pl.lit(None, dtype=typ)
+    elif dtype.is_array():
         value = pl.Series("", value)
-        typ = PolarsType.from_ibis(dtype)
         val = pl.lit(value, dtype=typ)
         return val.implode()
     elif dtype.is_struct():
@@ -98,14 +102,11 @@ def literal(op, **_):
             for k, v in value.items()
         ]
         return pl.struct(values)
-    elif dtype.is_interval():
-        return _make_duration(value, dtype)
     elif dtype.is_null():
         return pl.lit(value)
     elif dtype.is_binary():
         return pl.lit(value)
     else:
-        typ = PolarsType.from_ibis(dtype)
         return pl.lit(op.value, dtype=typ)
 
 
@@ -985,9 +986,12 @@ def array_concat(op, **kw):
 
 
 @translate.register(ops.Array)
-def array_column(op, **kw):
-    cols = [translate(col, **kw) for col in op.exprs]
-    return pl.concat_list(cols)
+def array_literal(op, **kw):
+    pdt = PolarsType.from_ibis(op.dtype)
+    if op.exprs:
+        return pl.concat_list([translate(col, **kw) for col in op.exprs]).cast(pdt)
+    else:
+        return pl.lit([], dtype=pdt)
 
 
 @translate.register(ops.ArrayCollect)
--- a/ibis/backends/sql/compiler.py
+++ b/ibis/backends/sql/compiler.py
@@ -1019,8 +1019,8 @@ class SQLGlotCompiler(abc.ABC):
             query = sg.select(STAR).from_(query)
         return needle.isin(query=query)
 
-    def visit_Array(self, op, *, exprs):
-        return self.f.array(*exprs)
+    def visit_Array(self, op, *, exprs, dtype):
+        return self.cast(self.f.array(*exprs), dtype)
 
     def visit_StructColumn(self, op, *, names, values):
         return sge.Struct.from_arg_list(
--- a/ibis/backends/tests/test_array.py
+++ b/ibis/backends/tests/test_array.py
@@ -32,6 +32,7 @@ from ibis.backends.tests.errors import (
     SnowflakeProgrammingError,
     TrinoUserError,
 )
+from ibis.common.annotations import ValidationError
 from ibis.common.collections import frozendict
 
 pytestmark = [
@@ -71,6 +72,75 @@ pytestmark = [
 # NB: We don't check whether results are numpy arrays or lists because this
 # varies across backends. At some point we should unify the result type to be
 # list.
+
+
+def test_array_factory(con):
+    a = ibis.array([1, 2, 3])
+    assert a.type() == dt.Array(value_type=dt.Int8)
+    assert con.execute(a) == [1, 2, 3]
+
+    a2 = ibis.array(a)
+    assert a.type() == dt.Array(value_type=dt.Int8)
+    assert con.execute(a2) == [1, 2, 3]
+
+
+def test_array_factory_typed(con):
+    typed = ibis.array([1, 2, 3], type="array<string>")
+    assert con.execute(typed) == ["1", "2", "3"]
+
+    typed2 = ibis.array(ibis.array([1, 2, 3]), type="array<string>")
+    assert con.execute(typed2) == ["1", "2", "3"]
+
+
+@pytest.mark.notimpl("flink", raises=Py4JJavaError)
+@pytest.mark.notimpl(["pandas", "dask"], raises=ValueError)
+def test_array_factory_empty(con):
+    with pytest.raises(ValidationError):
+        ibis.array([])
+
+    empty_typed = ibis.array([], type="array<string>")
+    assert empty_typed.type() == dt.Array(value_type=dt.string)
+    assert con.execute(empty_typed) == []
+
+
+@pytest.mark.notyet(
+    "clickhouse", raises=ClickHouseDatabaseError, reason="nested types can't be NULL"
+)
+@pytest.mark.notyet(
+    "flink", raises=Py4JJavaError, reason="Parameters must be of the same type"
+)
+def test_array_factory_null(con):
+    with pytest.raises(ValidationError):
+        ibis.array(None)
+    with pytest.raises(ValidationError):
+        ibis.array(None, type="int64")
+    none_typed = ibis.array(None, type="array<string>")
+    assert none_typed.type() == dt.Array(value_type=dt.string)
+    assert con.execute(none_typed) is None
+
+    nones = ibis.array([None, None], type="array<string>")
+    assert nones.type() == dt.Array(value_type=dt.string)
+    assert con.execute(nones) == [None, None]
+
+    # Execute a real value here, so the backends that don't support arrays
+    # actually xfail as we expect them to.
+    # Otherwise would have to @mark.xfail every test in this file besides this one.
+    assert con.execute(ibis.array([1, 2])) == [1, 2]
+
+
+@pytest.mark.broken(
+    ["datafusion", "flink", "polars"],
+    raises=AssertionError,
+    reason="[None, 1] executes to [np.nan, 1.0]",
+)
+def test_array_factory_null_mixed(con):
+    none_and_val = ibis.array([None, 1])
+    assert none_and_val.type() == dt.Array(value_type=dt.Int8)
+    assert con.execute(none_and_val) == [None, 1]
+
+    none_and_val_typed = ibis.array([None, 1], type="array<string>")
+    assert none_and_val_typed.type() == dt.Array(value_type=dt.String)
+    assert con.execute(none_and_val_typed) == [None, "1"]
 
 
 def test_array_column(backend, alltypes, df):
--- a/ibis/backends/tests/test_map.py
+++ b/ibis/backends/tests/test_map.py
@@ -11,6 +11,7 @@ import ibis
 import ibis.common.exceptions as exc
 import ibis.expr.datatypes as dt
 from ibis.backends.tests.errors import PsycoPg2InternalError, Py4JJavaError
+from ibis.common.annotations import ValidationError
 
 pytestmark = [
     pytest.mark.never(
@@ -37,6 +38,31 @@ mark_notimpl_risingwave_hstore = pytest.mark.notimpl(
     ["risingwave"],
     reason="function hstore(character varying[], character varying[]) does not exist",
 )
+
+
+@mark_notimpl_risingwave_hstore
+@pytest.mark.parametrize(
+    "values_factory",
+    [
+        lambda: ({"a": "b"},),
+        lambda: (["a"], ["b"]),
+        lambda: (ibis.map({"a": "b"}),),
+    ],
+)
+def test_map_factory_dict(con, values_factory):
+    vals = values_factory()
+    with pytest.raises(ValidationError):
+        ibis.map(*vals, type="array<string>")
+    assert con.execute(ibis.map(*vals)) == {"a": "b"}
+    assert con.execute(ibis.map(*vals, type="map<string, string>")) == {"a": "b"}
+
+
+def test_map_null(con):
+    with pytest.raises(ValidationError):
+        ibis.map(None)
+    with pytest.raises(ValidationError):
+        ibis.map(None, type="array<string>")
+    assert con.execute(ibis.map(None, type="map<string, string>")) is None
 
 
 @pytest.mark.notyet("clickhouse", reason="nested types can't be NULL")
@@ -503,6 +529,11 @@ values = pytest.mark.parametrize(
             marks=[
                 pytest.mark.notyet("clickhouse", reason="nested types can't be null"),
                 mark_notyet_postgres,
+                pytest.mark.notimpl(
+                    "flink",
+                    raises=Py4JJavaError,
+                    reason="Unexpected error in type inference logic of function 'COALESCE'",
+                ),
             ],
             id="struct",
         ),
--- a/ibis/backends/tests/test_struct.py
+++ b/ibis/backends/tests/test_struct.py
@@ -14,12 +14,14 @@ import ibis
 import ibis.expr.datatypes as dt
 from ibis import util
 from ibis.backends.tests.errors import (
+    ClickHouseDatabaseError,
     PolarsColumnNotFoundError,
     PsycoPg2InternalError,
     PsycoPg2SyntaxError,
     Py4JJavaError,
     PySparkAnalysisException,
 )
+from ibis.common.annotations import ValidationError
 from ibis.common.exceptions import IbisError
 
 pytestmark = [
@@ -27,6 +29,62 @@ pytestmark = [
     pytest.mark.notyet(["impala"]),
     pytest.mark.notimpl(["datafusion", "druid", "oracle", "exasol"]),
 ]
+
+mark_notimpl_postgres_literals = pytest.mark.notimpl(
+    "postgres", reason="struct literals not implemented", raises=PsycoPg2SyntaxError
+)
+
+
+@pytest.mark.notimpl("risingwave")
+@pytest.mark.broken("postgres", reason="JSON handling is buggy")
+@pytest.mark.notimpl(
+    "flink",
+    raises=Py4JJavaError,
+    reason="Unexpected error in type inference logic of function 'COALESCE'",
+)
+def test_struct_factory(con):
+    s = ibis.struct({"a": 1, "b": 2})
+    assert con.execute(s) == {"a": 1, "b": 2}
+
+    s2 = ibis.struct(s)
+    assert con.execute(s2) == {"a": 1, "b": 2}
+
+    typed = ibis.struct({"a": 1, "b": 2}, type="struct<a: string, b: string>")
+    assert con.execute(typed) == {"a": "1", "b": "2"}
+
+    typed2 = ibis.struct(s, type="struct<a: string, b: string>")
+    assert con.execute(typed2) == {"a": "1", "b": "2"}
+
+    items = ibis.struct([("a", 1), ("b", 2)])
+    assert con.execute(items) == {"a": 1, "b": 2}
+
+
+@pytest.mark.parametrize("type", ["struct<>", "struct<a: int64, b: int64>"])
+@pytest.mark.parametrize("val", [{}, []])
+def test_struct_factory_empty(val, type):
+    with pytest.raises(ValidationError):
+        ibis.struct(val, type=type)
+
+
+@pytest.mark.notimpl("risingwave")
+@mark_notimpl_postgres_literals
+@pytest.mark.notyet(
+    "clickhouse", raises=ClickHouseDatabaseError, reason="nested types can't be NULL"
+)
+@pytest.mark.broken(
+    "polars",
+    reason=r"pl.lit(None, type='struct<a: int64>') gives {'a': None}: https://github.com/pola-rs/polars/issues/3462",
+)
+def test_struct_factory_null(con):
+    with pytest.raises(ValidationError):
+        ibis.struct(None)
+    none_typed = ibis.struct(None, type="struct<a: float64, b: float>")
+    assert none_typed.type() == dt.Struct(fields={"a": dt.float64, "b": dt.float64})
+    assert con.execute(none_typed) is None
+    # Execute a real value here, so the backends that don't support structs
+    # actually xfail as we expect them to.
+    # Otherwise would have to @mark.xfail every test in this file besides this one.
+    assert con.execute(ibis.struct({"a": 1, "b": 2})) == {"a": 1, "b": 2}
 
 
 @pytest.mark.notimpl(["dask"])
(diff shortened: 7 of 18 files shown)


jcrist (commented):
Makes sense to me. Can you add a few tests in the appropriate `ibis/tests/expr/test_{dtype}.py` files?

Also curious - what was the use case that led to you wanting this change?

NickCrews:
in my first attempt at https://github.com/ibis-project/ibis/pull/8649, [I went about it the wrong way](https://github.com/ibis-project/ibis/commit/6925c908122174cd3e6690b50aa2777b6b6463d1#diff-d50c70a1d9c6f811ab5dba8574dbb12647c81ced87d539b19f5517ec037eca9eR495-R496), but I needed to ensure something was an Array, but the incoming thing might already have been an array.

ibis/expr/types/maps.py
```diff
@@ -435,8 +436,10 @@ def __getitem__(self, key: ir.Value) -> ir.Column:
 @public
 @deferrable
 def map(
-    keys: Iterable[Any] | Mapping[Any, Any] | ArrayColumn,
-    values: Iterable[Any] | ArrayColumn | None = None,
+    keys: Iterable[Any] | Mapping[Any, Any] | ir.ArrayValue | MapValue | None,
+    values: Iterable[Any] | ir.ArrayValue | None = None,
+    *,
+    type: str | dt.DataType | None = None,
```
NickCrews, line 442: For this factory, `type` is kwarg-only. Should the other factories also be made kwarg-only to match?

ibis/expr/types/arrays.py
```diff
@@ -1120,15 +1128,32 @@ def array(values: Iterable[V]) -> ArrayValue:
...
+    if values is None:
+        if type is None:
+            raise TypeError("type must be specified when values is None")
+        return ibis.literal(None, type=type)
+
+    if isinstance(values, ir.ArrayValue):
+        result = values
+    else:
+        values = tuple(values)
+        if len(values) == 0:
```
cpcloud, line 1151:
```suggestion
        if not values:
```

ibis/backends/tests/test_struct.py
```diff
@@ -23,6 +24,38 @@
 ]
 
 
+@pytest.mark.notimpl(["postgres"])
+# @pytest.mark.broken(["pandas", "dask"], reason="casting is broken")
```
cpcloud, line 28:
```suggestion
```

ibis/backends/tests/test_array.py
```diff
@@ -70,6 +70,40 @@
...
+    assert con.execute(typed2) == [1.0, 2.0, 3.0]
+
+
+@pytest.mark.notimpl("postgres", raises=PsycoPg2IndeterminateDatatype)
+def test_array_factory_empty(con):
+    with pytest.raises(TypeError):
+        ibis.array([])
+
+    empty_typed = ibis.array([], type="array<float64>")
+    assert str(empty_typed.type()) == "array<float64>"
```
cpcloud, line 90:
```suggestion
```

If you want to verify the type, use a datatype instance, not its string representation.

ibis/backends/tests/test_map.py
```diff
@@ -26,6 +30,48 @@
...
+    ["risingwave"],
+    raises=PsycoPg2InternalError,
+    reason="function hstore(character varying[], character varying[]) does not exist",
+)
+@pytest.mark.notyet("postgres", reason="only support maps of string -> string")
+def test_map_factory_empty(con):
+    with pytest.raises(TypeError):
+        ibis.map({})
+    empty_typed = ibis.map({}, type="map<string, string>")
+    assert str(empty_typed.type()) == "map<string, string>"
```
cpcloud, line 60:
```suggestion
```

Construct the type using a Python object, not a string.

ibis/backends/tests/test_map.py
```diff
@@ -26,6 +30,48 @@
...
+
+
+@pytest.mark.notyet(
+    "clickhouse", raises=ClickHouseDatabaseError, reason="nested types can't be NULL"
+)
+def test_map_factory_null(con):
+    with pytest.raises(TypeError):
+        ibis.map(None)
+    null_typed = ibis.map(None, type="map<string, string>")
+    assert str(null_typed.type()) == "map<string, string>"
```
cpcloud, line 71:
```suggestion
```

ibis/backends/tests/test_struct.py
```diff
@@ -23,6 +24,38 @@
...
+
+@pytest.mark.broken("polars", raises=AttributeError)
+@pytest.mark.notyet(
+    "clickhouse", raises=ClickHouseDatabaseError, reason="nested types can't be NULL"
+)
+def test_struct_factory_null(con):
+    with pytest.raises(TypeError):
+        ibis.struct(None)
+    none_typed = ibis.struct(None, type="struct<a: float64, b: float>")
+    assert str(none_typed.type()) == "struct<a: float64, b: float64>"
```
cpcloud, line 55:
```suggestion
```

ibis/expr/types/arrays.py
```diff
@@ -5,14 +5,16 @@
...
 import ibis.expr.operations as ops
+import ibis.expr.types as ir
 from ibis.common.deferred import Deferred, deferrable
 from ibis.expr.types.generic import Column, Scalar, Value
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
-    import ibis.expr.types as ir
+    from ibis.expr.types import dt
```
cpcloud, line 17:
```suggestion
    import ibis.expr.datatypes as dt
```

Definitely do not import internal module aliases from other modules.

ibis/expr/types/maps.py
```diff
@@ -4,15 +4,16 @@
...
+import ibis.expr.types as ir
 from ibis.common.deferred import deferrable
 from ibis.expr.types.generic import Column, Scalar, Value
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Mapping
 
-    import ibis.expr.types as ir
-    from ibis.expr.types.arrays import ArrayColumn
+    from ibis.expr import datatypes as dt
```
cpcloud, line 16: Please follow the same import style as the other Ibis modules:

```suggestion
    import ibis.expr.datatype as dt
```

ibis/expr/types/maps.py
cpcloud: Yes, make them consistent :)

ibis/expr/types/structs.py
```diff
@@ -62,26 +61,38 @@ def struct(
...
+        if type is None:
+            raise TypeError("Must specify type if value is None")
+        return literal(None, type=type)
+
+    if isinstance(value, StructValue):
+        result = value
     else:
-        return literal(collections.OrderedDict(fields), type=type)
+        fields = dict(value)
+        if any(isinstance(value, Value) for value in fields.values()):
```
cpcloud, line 86: We should really get rid of this branching here and not have this odd split between literal structs and "non-literal" structs. There's no real reason to differentiate AFAICT.

That can be done in a follow-up.

ibis/formats/pandas.py
```diff
@@ -363,6 +363,11 @@ def convert(values):
             if values is None:
                 return values
 
+            try:
+                values = iter(values)
+            except TypeError:
+                # iteration over 0-d array, happens when the element is NULL
+                return None
```
cpcloud, line 370: Did you add a test case for this code path?

ibis/backends/tests/test_array.py
```diff
@@ -70,6 +70,40 @@
...
 
+def test_array_factory(con):
+    a = ibis.array([1, 2, 3])
+    assert con.execute(a) == [1, 2, 3]
+    a2 = ibis.array(a)
+    assert con.execute(a2) == [1, 2, 3]
+    typed = ibis.array([1, 2, 3], type="array<float64>")
+    assert con.execute(typed) == [1.0, 2.0, 3.0]
+    typed2 = ibis.array(a, type="array<float64>")
+    assert con.execute(typed2) == [1.0, 2.0, 3.0]
```
cpcloud, line 81: Can you group each pair of lines with a newline between them? It's hard to read at the moment.

cpcloud (changes requested):

ibis/backends/pandas/convert.py
```diff
@@ -83,3 +83,11 @@ def convert_String(cls, s, dtype, pandas_type):
         # TODO(kszucs): should switch to the new pandas string type and convert
         # object columns using s.convert_dtypes() method
         return s.map(str, na_action="ignore").astype(object)
+
+    @classmethod
+    def convert_Struct(cls, s, dtype, pandas_type):
+        # dt.normalize returns a frozendict, which doesn't play well
+        # with the pandas executor
```
jcrist, line 90: It's just the formatter when pandas repr's a series containing a `FrozenDict`, the actual execution is fine. See #8687.

ibis/backends/pandas/convert.py
```diff
@@ -83,3 +83,11 @@ def convert_String(cls, s, dtype, pandas_type):
...
+        # https://github.com/ibis-project/ibis/issues/8687
+        func = lambda x: x if x is pd.NA else dict(dt.normalize(dtype, x))
```
jcrist, line 92: I _think_ you'll need to recurse here for nested structs.

ibis/backends/pandas/convert.py
NickCrews: It actually isn't, if you leave out this method the test fails. See the linked issue again, I added some clarification

ibis/backends/pandas/convert.py
NickCrews: dt.normalize will recurse, but that's not quite enough, that would lead to dict(frozendict(frozendict(..))). I assume we want dict all the way down. This seems hard to do.

better if we can address the linked bug, then just frozendict s all the way down and we don't need this workaround

ibis/backends/tests/test_struct.py
NickCrews: here is the test case that exercises the change in formats/pandas.py.

ibis/formats/pandas.py
NickCrews: Yes, test_struct_factory fails for pandas and dask without this change. That is why there was that stray forgotten commented out `xfail(["pandas", "dask"])` on that test.

ibis/expr/types/maps.py
NickCrews: oops, sorry this was blind acceptance of either vscode intellisense or copilor suggestsions.

ibis/expr/types/arrays.py
NickCrews: oops, must have been a blind vscode/copilot acceptance.

ibis/backends/tests/test_array.py
NickCrews: done.

ibis/backends/tests/test_array.py
NickCrews: done

ibis/backends/pandas/convert.py
NickCrews: This has been fixed in [#Frozen](https://github.com/ibis-project/ibis/pull/8693) so I removed this workaround

ibis/backends/pandas/convert.py
NickCrews: resolved in https://github.com/ibis-project/ibis/pull/8693, so this is no longer relevant.

ibis/formats/pandas.py
NickCrews: this is removed now

ibis/expr/types/structs.py
NickCrews: OK, I ended up doing this myself in this commit. The branching was causing me lots of problems too.

ibis/backends/tests/test_struct.py
NickCrews: no longer relevant.

NickCrews:
OK I just pushed an absolute doozy of a commit in https://github.com/ibis-project/ibis/pull/8666/commits/530fad0e29d37b52880b2767f527d3abc8790fdb. Curious how many backends that fails on. It required really overhauling the internal representation of ops.Array and ops.Struct to be able to distinguish between
- the overall value is NULL
- the contained values is NULL
Take a look at that commit description.

ncclementi:
Hi @NickCrews, checking in here. What's the status on this one? It looks like there are a lot of conflicts and red on CI 😅 .

NickCrews:
Yeeesssss this needs a bit of love :) Once my other pr with .cases lands I figured I would pick this up again.

ibis/backends/dask/helpers.py
```diff
@@ -47,7 +47,7 @@ def asseries(cls, value, like=None):
         elif isinstance(value, pd.Series):
             return dd.from_pandas(value, npartitions=1)
         elif like is not None:
-            if isinstance(value, (tuple, list, dict)):
+            if isinstance(value, (tuple, list, dict, np.ndarray)):
```
NickCrews, line 50: ops.Array executes to a np.ndarray (both before this change, and after). We never ran into this scenario in tests before, so this tweak here is adjusting an existing bug.

NickCrews:
@cpcloud @jcrist this is ready for review whenever you get the chance. This is a bit of a doozy that might want to get split up into multiple commits/PRs. But I think it would be good to get a check that the overall direction is right before I do that? The summary in the top of this PR is up to date for a summary. Maybe read that and then it would be probably best to look at the tests, and then the factory methods in types/arrays.py, types/maps.py, and types/structs.py

cpcloud:
@NickCrews 

> Also, fix a typing bug: map() can accept ArrayValues, not just ArrayColumns.

> Also, fix executing Literal(None) on pandas and polars, 0-length arrays on polars

> Also, fixing converting dtypes on clickhouse, Structs should be converted to nonnullable dtypes.

> Also, implement ops.StructColumn on pandas and dask

Please split these up into separate PRs.

Is 

> A big structural change is that now ops.Array has an optional
attribute "dtype", so if you pass in a 0-length sequence
of values the op still knows what dtype it is.

required for this PR?

Can't we handle this in separate PR? It seems like it's just to handle the empty array case.

NickCrews:
> Please split these up into separate PRs.

Yeah will do :)

> It seems like it's just to handle the empty array case.

It is. This is because currently empty arrays are represented by ops.Literals. This leads to needing to compile arrays inside visit_Literal. After this PR, ops.Array can handle empty arrays (ops.Struct and ops.Map already can I think), so we are then free to move towards removing all nested types from ops.Literal, and visit_Literal will only need to deal with non-nested values. What do you think, is this worth the churn?

cpcloud:
> visit_Literal will only need to deal with non-nested values. What do you think, is this worth the churn?

I think this is worth exploring for sure. I'm not 100% sure it will be a net improvement without implementing it though :) It sounds like a nice improvement if it works out though.

ibis/backends/pandas/executor.py
```diff
@@ -49,12 +49,14 @@ def visit(cls, op: ops.Node, **kwargs):
...
     @classmethod
     def visit(cls, op: ops.Literal, value, dtype):
+        if value is None:
+            return None
         if dtype.is_interval():
-            value = pd.Timedelta(value, dtype.unit.short)
-        elif dtype.is_array():
-            value = np.array(value)
-        elif dtype.is_date():
-            value = pd.Timestamp(value, tz="UTC").tz_localize(None)
```
cpcloud, line 57: Can we keep the previous style? The branches are intentionally written this way.

ibis/backends/polars/compiler.py
```diff
@@ -86,10 +86,14 @@ def _make_duration(value, dtype):
 def literal(op, **_):
     value = op.value
     dtype = op.dtype
+    if dtype.is_interval():
+        return _make_duration(value, dtype)
 
-    if dtype.is_array():
```
cpcloud, line 90: Can you avoid moving the code around here and **only** handle the `None` case? It's harder to review when these kinds of changes are mixed together.

ibis/backends/polars/compiler.py
```diff
@@ -980,9 +981,11 @@ def array_concat(op, **kw):
 
 
 @translate.register(ops.Array)
-def array_column(op, **kw):
-    cols = [translate(col, **kw) for col in op.exprs]
-    return pl.concat_list(cols)
+def array_literal(op, **kw):
+    if len(op.exprs) > 0:
```
cpcloud, line 985:
```suggestion
    if op.exprs:
```

ibis/backends/sql/compiler.py
```diff
@@ -970,8 +970,11 @@ def visit_InSubquery(self, op, *, rel, needle):
             query = sg.select(STAR).from_(query)
         return needle.isin(query=query)
 
-    def visit_Array(self, op, *, exprs):
-        return self.f.array(*exprs)
+    def visit_Array(self, op, *, exprs, dtype):
+        result = self.f.array(*exprs)
+        if len(exprs) == 0:
```
cpcloud, line 975:
```suggestion
        if not exprs:
```

ibis/backends/sql/datatypes.py
```diff
@@ -1007,8 +1007,10 @@ class ClickHouseType(SqlglotType):
     def from_ibis(cls, dtype: dt.DataType) -> sge.DataType:
         """Convert a sqlglot type to an ibis type."""
         typ = super().from_ibis(dtype)
-        if dtype.nullable and not (dtype.is_map() or dtype.is_array()):
-            # map cannot be nullable in clickhouse
+        # nested types cannot be nullable in clickhouse
+        if dtype.nullable and not (
+            dtype.is_map() or dtype.is_array() or dtype.is_struct()
+        ):
```
cpcloud, line 1013: Can you break this out into a separate PR?

ibis/backends/tests/test_array.py
```diff
@@ -72,6 +73,75 @@
...
+    assert a.type() == dt.Array(value_type=dt.Int8)
+    assert con.execute(a) == [1, 2, 3]
+
+    a2 = ibis.array(a)
+    assert a.type() == dt.Array(value_type=dt.Int8)
+    assert con.execute(a2) == [1, 2, 3]
+
+
+def test_array_factory_typed(con):
+    typed = ibis.array([1, 2, 3], type="array<string>")
```
cpcloud, line 87: I don't really like the implicit conversion here. Is it necessary?

ibis/expr/operations/structs.py
```diff
@@ -34,7 +34,9 @@ class StructColumn(Value):
 
     shape = rlz.shape_like("values")
 
-    def __init__(self, names, values):
+    def __init__(self, names: VarTuple[str], values: VarTuple[Value]):
+        if len(names) == 0:
+            raise ValidationError("StructColumn must have at least one field")
```
cpcloud, line 39: Please remove this change. We haven't decided whether to continue to allow empty structs. For now, they are allowed.

```suggestion
```

ibis/expr/operations/structs.py
```diff
@@ -43,6 +45,5 @@ def __init__(self, names, values):
         super().__init__(names=names, values=values)
 
     @attribute
-    def dtype(self) -> dt.DataType:
-        dtypes = (value.dtype for value in self.values)
-        return dt.Struct.from_tuples(zip(self.names, dtypes))
```
cpcloud, line 48: Can you avoid making style changes in this PR?

ibis/expr/types/arrays.py
```diff
@@ -1106,15 +1115,38 @@ def array(values: Iterable[V]) -> ArrayValue:
...
+            )
+
+    if values is None:
+        if type is None:
+            raise ValidationError("If values is None/NULL, dtype must be provided")
+        return ir.null(type)
+
+    values = tuple(values)
+    if len(values) > 0 and type is not None:
+        return ops.Array(values).to_expr().cast(type)
```
cpcloud, line 1150: Can we avoid doing operations like `cast` here? Doing this makes assumptions about how conversion should work in the backend and casting may or may not be the right approach. Instead, we should handle this in the backends.

It'll probably be extra code to do that.

ibis/expr/operations/arrays.py
```diff
@@ -7,23 +7,32 @@
...
+
+    def __init__(self, exprs, dtype: dt.Array | None = None):
+        if len(exprs) == 0:
+            if dtype is None:
+                raise ValidationError("If values is empty, dtype must be provided")
+            if not isinstance(dtype, dt.Array):
+                raise ValidationError(f"dtype must be an array, got {dtype}")
+        elif dtype is None:
+            dtype = dt.Array(rlz.highest_precedence_dtype(exprs))
+        super().__init__(exprs=exprs, dtype=dtype)
```
cpcloud, line 28: Why are we doing this validation here and not at the API entry point?

ibis/tests/expr/test_literal.py
```diff
@@ -109,42 +109,29 @@ def test_normalized_underlying_value(userinput, literal_type, expected_type):
...
     assert a.type() == dt.dtype(typestr)
 
 
 @pytest.mark.parametrize(
     "value",
     [
-        dict(field1="value1", field3=3.14),  # wrong field name
-        dict(field1="value1"),  # missing field
+        pytest.param(dict(field1="value1", field3=3.14), id="wrong_field"),
+        pytest.param(dict(field1="value1"), id="missing_field"),
```
cpcloud, line 119: import param from pytest to avoid some verbosity

ibis/expr/types/structs.py
```diff
@@ -62,26 +66,50 @@ def struct(
...
+
     fields = dict(value)
-    if any(isinstance(value, Value) for value in fields.values()):
-        names = tuple(fields.keys())
-        values = tuple(fields.values())
-        return ops.StructColumn(names=names, values=values).to_expr()
-    else:
-        return literal(collections.OrderedDict(fields), type=type)
+    names = tuple(fields.keys())
+    values = tuple(fields.values())
```
cpcloud, line 103: You shouldn't need to convert these to `tuple`

ibis/expr/operations/arrays.py
```diff
@@ -7,23 +7,32 @@
...
 class Array(Value):
     exprs: VarTuple[Value]
+    dtype: Optional[dt.Array] = None
+
+    def __init__(self, exprs, dtype: dt.Array | None = None):
+        if len(exprs) == 0:
+            if dtype is None:
+                raise ValidationError("If values is empty, dtype must be provided")
+            if not isinstance(dtype, dt.Array):
+                raise ValidationError(f"dtype must be an array, got {dtype}")
```
cpcloud, line 25: You don't need this at all, the declaration of `dtype: Optional[dt.Array]` creates a validation rule.

ibis/expr/operations/arrays.py
```diff
@@ -7,23 +7,32 @@
...
-from ibis.common.annotations import attribute
+from ibis.common.annotations import ValidationError, attribute
 from ibis.common.typing import VarTuple  # noqa: TCH001
 from ibis.expr.operations.core import Unary, Value
 
 
 @public
 class Array(Value):
     exprs: VarTuple[Value]
+    dtype: Optional[dt.Array] = None
```
cpcloud, line 18: I think I would prefer to split this operation into two to avoid this messy validation.

One op for inferred type and one for specified type. As is this logic seems unnecessarily complicated.

ibis/backends/tests/test_struct.py
```diff
@@ -79,6 +138,9 @@ def test_all_fields(struct, struct_df):
 
 @pytest.mark.notimpl(["postgres", "risingwave"])
 @pytest.mark.parametrize("field", ["a", "b", "c"])
+@pytest.mark.notyet(
+    ["flink"], reason="flink doesn't support creating struct columns from literals"
```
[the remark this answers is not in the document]
cpcloud: I think that's right.

ibis/backends/polars/compiler.py
NickCrews: Need to look at this again, but IIRC this was actually the minimum possible change, something was erroring with NULL duration handling. Will check again

ibis/backends/tests/test_array.py
NickCrews: The conversion is exactly the part of the API I was trying to test here. I think this conversion is something that this API should explicitly support. What do you think?

Maybe I add a param as

```python
@pytest.mark.param(raw, [[1,2,3], ["1", "2", "3"]])
def test_array_factory_typed(con, raw):
    typed = ibis.array(raw, type="array<string>")
    ...
```

ibis/expr/operations/structs.py
NickCrews: hmm, yeah IDK why I added this, thanks!

ibis/expr/types/arrays.py
NickCrews: Not sure I totally understand, like ops.Array should always have a dtype present, and the visit_Array() methods should always deal with it?

I wanted to support specifying a dtype in the ibis.array() API. To get there, I thought it would be better to split up the process into two stages: 1. create the value 2. cast to the right type. This leans better on the already existing casting implementations. But are you saying there are cases where this two-step process doesn't work, ie you lose some precision or something?

ibis/expr/operations/arrays.py
NickCrews: Currently, ibis.array() is the only entry point for creating an ops.Array. But, I wanted to future proof us: if in the future someone creates an ops.Array from some other place, I wanted them to be able to take advantage of this validation. What do you think? premature optimization?

ibis/expr/operations/arrays.py
NickCrews: You are right it has a validation rule, but the generic error message isn't that clear.

Also just to be clear (don't think it matters here) the validation from the signature happens during the `super().__init__()` call, so the validation will happen, but in the future, it hasn't happened yet.

ibis/expr/operations/arrays.py
NickCrews: If the complex validation stuff gets hoisted into ibis.array() (see below), do you still have this concern?

NickCrews:
I'm picking off some of the changes in individual PRs, I'll let you know when this is ready for another look here.

ibis/backends/sql/datatypes.py
NickCrews: done in https://github.com/ibis-project/ibis/pull/9305

ibis/expr/operations/structs.py
NickCrews: Actually, empty structs are currently disallowed. But working on fixing that in https://github.com/ibis-project/ibis/pull/9310

ibis/backends/polars/compiler.py
NickCrews: Yes, this reorg is needed. ibis's Duration.DAY and friends are not representable in polars' datatypes, so if you call `PolarsType.from_ibis(dtype)` on those, you get an error. So we need an early guard to look for the interval dtypes and return early.

ibis/backends/sql/compiler.py
```diff
@@ -1019,8 +1019,8 @@ def visit_InSubquery(self, op, *, rel, needle):
             query = sg.select(STAR).from_(query)
         return needle.isin(query=query)
 
-    def visit_Array(self, op, *, exprs):
-        return self.f.array(*exprs)
+    def visit_Array(self, op, *, exprs, dtype):
+        return self.cast(self.f.array(*exprs), dtype)
```
cpcloud, line 1023: This seems extraordinarily heavy-handed and overly broad.

Do we really need to cast the entire on constructing an array, for all SQL backends?

cpcloud:
This PR is still too big. Can you split the array, map and struct implementations into three separate PRs?

NickCrews:
Yeah splitting it probably is good.

- map is easy, no semantic changes going on there
- for struct, I want to add tests for empty structs. Therefore I think we need to decide on and merge my other PR regarding this behavior 
- for arrays, it's gonna be a little hairy too, because I am implementing empty arrays at the same time as adjusting the API. I don't really want to do them separately or it will just be churn. Unless you see a decent way to address those two things separately

cpcloud:
@NickCrews Just checking in here, can you split these up some more?

NickCrews:
OK, I just tried to split the struct and the map stuff into their own PRs, but both of those rely on the new `ibis.array(arg, typ)` API. So we need to get the array stuff in first. I'll split that out.

NickCrews:
split this into
- #9458 
- #9460 
- #9459

PR #8667, opened 2024-03-15
fix(pandas): map `date` type to `datetime64[s]`

jcrist:
Previously ibis's `date` type was mapped to `datetime64[D]`. This type isn't supported by pandas (the max is `datetime64[s]`), but the errors were hidden due to a `try/except` block causing things to fallback to `dtype('O')`. Pandas 2.2 changed the error raised here resulting in test failures.


cpcloud (approved):
Nice!

PR #8668, opened 2024-03-15
feat(mssql): use integrated auth when no user or password supplied

### LINKED ISSUE #8657: feat(mssql): Explicit Integrated Auth for MSSQL Connections

Microsoft SQL Server and Windows allows for the use of [Integrated Windows Authentication](https://learn.microsoft.com/en-us/aspnet/web-api/overview/security/integrated-windows-authentication) to connect to the database servers. Often this is also enforced as the only option, and only Windows credentials can be used, not username and password.

- Ibis does not have documentation on using Integrated Windows Authentication, however it is supported by PyODBC
- Ibis does not have explicit args in the MSSQL connection or a default behavior when username and password are 'None' to use Integrated Windows Authentication
- **con.table** does not seem to inherit the database from **ibis.mssql.connect**, and must use explicit args instead of just the table name with schema separated by a '.'

## Current Solution for Windows Integrated Authentication:

```python
import ibis
import pandas as pd

# Connect to the MS SQL Server database
conn = ibis.mssql.connect(
    host='my_server',
    user=None,
    password=None
    database='my_database',
    driver='SQL Server',          # ODBC Driver
    Trusted_Connection = 'yes'
)

# Query Table
tbl = con.table(name='my_table', schema='my_schema', database='my_database')
df = tbl.limit(20).to_pandas()

df.head()
```

1. Modify ibis.mssql.connect to assume 
2. Update Ibis [MSSQL backend documentation](https://ibis-project.org/backends/mssql) to include section for integrated / windows authentication.

## Proposed Solution for Windows Integrated Authentication

```python
import ibis
import pandas as pd

# If both user and password are set to None, assume Integrated Windows Authentication and set Trusted_Connection = 'yes'
con = ibis.mssql.connect(     
    host="my_server",
    database="my_database",
    driver='SQL Server'
)

# Query Table
tbl = con.table('my_schema.my_table')  # Use of MSSQL and a database being passed allows for single string to query table.
df = tbl.limit(20).to_pandas()

df.head()
```

8.0.0

MSSQL

### Code of Conduct

gforsyth:
The method for enabling Integrated Authentication with ODBC is to not
provide a `user` or `password`, and to set `Trusted_Connection =
'true'`.

If `Trusted_Connection` is already set in `kwargs`, do nothing, because
either the user set it to 'true' and it will work untouched, or they're
specifying 'false' for a reason and we should leave it alone.- feat(mssql): use integrated auth when no user or password supplied
- docs(mssql): add odbc tabset and integrated authentication section

Resolves #8657

### DIFF
--- a/docs/backends/mssql.qmd
+++ b/docs/backends/mssql.qmd
@@ -71,13 +71,31 @@ con = ibis.mssql.connect()  # <1>
 
 ### `ibis.mssql.connect`
 
+::: {.panel-tabset}
+
+## Windows
+
 ```python
 con = ibis.mssql.connect(
     user="username",
     password="password",
     host="hostname",
+    driver="SQL Server",
 )
 ```
+
+## OSX / Linux
+
+```python
+con = ibis.mssql.connect(
+    user="username",
+    password="password",
+    host="hostname",
+    driver="FreeTDS",  # if you are using unixODBC
+)
+```
+
+:::
 
 ::: {.callout-note}
 `ibis.mssql.connect` is a thin wrapper around [`ibis.backends.mssql.Backend.do_connect`](#ibis.backends.mssql.Backend.do_connect).
@@ -101,6 +119,20 @@ passing a properly-formatted MSSQL connection URL to `ibis.connect`:
 ```python
 con = ibis.connect(f"mssql://{user}:{password}@{host}:{port}")
 ```
+
+### Using Integrated Authentication
+
+Integrated Authentication allows users to log in via NTLM or Kerberos.
+To make use of integrated authentication, use `ibis.mssql.connect` as documented
+above, but do not specify values for `user` or `password`. The PyODBC driver
+will then default to Integrated Authentication.
+
+Some environments may require you to set additional keywords to connect.
+Additional information is available at the following pages:
+
+<https://learn.microsoft.com/en-us/aspnet/web-api/overview/security/integrated-windows-authentication>
+
+<https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/using-integrated-authentication>
 
 ```{python}
 #| echo: false
--- a/docs/backends/mssql.qmd
+++ b/docs/backends/mssql.qmd
@@ -130,9 +130,8 @@ will then default to Integrated Authentication.
 Some environments may require you to set additional keywords to connect.
 Additional information is available at the following pages:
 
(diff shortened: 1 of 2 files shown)


ibis/backends/mssql/__init__.py
```diff
@@ -76,6 +76,43 @@ def do_connect(
...
+            See https://learn.microsoft.com/en-us/sql/connect/odbc/windows/system-requirements-installation-and-driver-files
+        kwargs
+            Additional keyword arguments to pass to PyODBC.
+        """
+
+        # If no user/password given, assume Windows Integrated Authentication
+        # and set "Trusted_Connection" accordingly
+        # see: https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/using-integrated-authentication
+        if user is None and password is None and "Trusted_Connection" not in kwargs:
+            kwargs["Trusted_Connection"] = "yes"
```
jcrist, line 114: Could go either way on this, just commenting show a possibly cleaner spelling:

```suggestion
        if user is None and password is None:
            kwargs.setdefault("Trusted_Connection", "yes")
```

docs/backends/mssql.qmd
```diff
@@ -102,6 +120,20 @@ passing a properly-formatted MSSQL connection URL to `ibis.connect`:
...
+To make use of integrated authentication, use `ibis.mssql.connect` as documented
+above, but do not specify values for `user` or `password`. The PyODBC driver
+will then default to Integrated Authentication.
+
+Some environments may require you to set additional keywords to connect.
+Additional information is available at the following pages:
+
+<https://learn.microsoft.com/en-us/aspnet/web-api/overview/security/integrated-windows-authentication>
+
+<https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/using-integrated-authentication>
```
jcrist, line 135: Do we want to format these links in any way (bulleted list? add title instead of raw link)? Or do these render fine as is? (I've never seen the `<{link}>` formatting used in qmd before, not does it do something different than just `{link}`? from rendering it looks the same to me).

jcrist (approved):
Overall LGTM! Two small comments, feel free to ignore all.

cpcloud (approved):

docs/backends/mssql.qmd
cpcloud: Changed to bulleted list.

PR #8669, opened 2024-03-15
refactor(duckdb): use `.sql` instead of `.execute` in performance-senitive locations

### LINKED ISSUE #8631: perf: should we be using con.sql(), not con.execute()?

I was just reading through https://github.com/duckdb/duckdb/discussions/9710#discussioncomment-8399235, and in there Mytherin, one of the duckdb leads, mentions something about how python clients should prefer .sql() to .execute(). I don't quite understand the architecture of what he's saying, but I wanted to bring it to your attention. We use .execute(). Maybe it sounds like in duckdb>=0.10.0 there is no difference, but for our users who are on older duckdb versions, we should still switch to the performant way?

Sorry I'm not more useful here, I'm not sure what sort of benchmark to perform to test this out.

NA

main

_No response_

### Code of Conduct

jcrist:
In #6875 we switched to using `con.sql(sql)` instead of `con.execute(sql)` when retrieving an arrow table of results. The former is more efficient than the latter since the `.sql` api lets the query planner take advantage of the known output type, possibly resulting in a more efficient execution.

In the `the-epic-split` refactor this optimization was dropped (note - it seems less necessary since duckdb 0.9.0, although they say [a performance difference is still expected](https://github.com/duckdb/duckdb/discussions/8582#discussioncomment-8806576)). We add the optimization back here, and apply it to all locations where we return in-memory data.

We choose _not_ to use `duckdb_con.sql` instead of `duckdb_con.execute` in the public-facing `raw_sql` api, since:
- The returned object isn't strictly compatible with the `dbapi` API
- In the case of operations that don't return results (e.g. DDL), `con.sql` returns `None`

Fixes #8631.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -1234,19 +1234,16 @@ class Backend(SQLBackend, CanCreateSchema, UrlFromPath):
             :::
 
         """
-        cur = self._to_duckdb_relation(expr, params=params, limit=limit)
-        rb = cur.record_batch(batch_size=chunk_size)
+        self._run_pre_execute_hooks(expr)
+        table = expr.as_table()
+        sql = self.compile(table, limit=limit, params=params)
 
-        def batch_stream(con, cur, rb):
-            # A generator function to hold a reference to con and cur
-            # throughout the lifespan of the record batch reader. This is to
-            # work around previous bugs upstream bugs in duckdb where segfaults
-            # or corruption would occur if the cursor was collected before the
-            # record batch reader was exhausted.
-            yield from rb
+        def batch_producer(cur):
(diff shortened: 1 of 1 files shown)


jcrist:
Oh joy - segfaults on CI, but I can't reproduce them locally: https://github.com/ibis-project/ibis/actions/runs/8302276853/job/22724239870?pr=8669 :eyes: :eyes:.

cpcloud:
It's geospatial-related from duckdb/duckdb_spatial#236

https://github.com/ibis-project/ibis/actions/runs/8302276853/job/22724239870?pr=8669#step:15:125

cpcloud (approved):
LGTM, thanks!

Issue #8670, opened 2024-03-15
feat: Make `Struct`s a Mapping


NickCrews:
### Is your feature request related to a problem?

I want to treat Structs as a Mapping:

```python
def debug(**args):
    for k, v in args.items():
        print(f"{k}: {v}")


debug(**ibis.struct({"a": 1}))
# argument after ** must be a mapping, not StructScalar
```

### Describe the solution you'd like

PR incoming

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jcrist:
See #8671, we decided this wasn't a good fit.

PR #8671, opened 2024-03-15
feat: make Structs a Mapping

NickCrews:
This allows them to be used in more ergonomic ways.

Fixes https://github.com/ibis-project/ibis/issues/8670

### DIFF
--- a/ibis/expr/types/structs.py
+++ b/ibis/expr/types/structs.py
@@ -146,6 +146,15 @@ class StructValue(Value):
     def _ipython_key_completions_(self) -> list[str]:
         return sorted(self.type().names)
 
+    def keys(self):
+        return self.names
+
+    def values(self):
+        return [self[name] for name in self.names]
+
+    def items(self):
+        return [(name, self[name]) for name in self.names]
+
     def __getitem__(self, name: str) -> ir.Value:
         """Extract the `name` field from this struct.
 
--- a/ibis/tests/expr/test_struct.py
+++ b/ibis/tests/expr/test_struct.py
@@ -98,3 +98,29 @@ def test_nested_lift():
     )
     expr = t.a.b.lift()
     assert expr.schema() == ibis.schema({"x": "int", "y": "int"})
+
+
+def test_struct_is_mapping():
+    s = ibis.struct({"a": "foo", "b": "bar"})
+    assert tuple(s.keys()) == ("a", "b")
+    assert len(tuple(s.values())) == 2
+    assert len(tuple(s.items())) == 2
+
+    def f(**kwargs):
+        pass
+
+    f(**s)
+
+
+def test_struct_isnt_full_container():
+    # Disallow some things to avoid footguns:
+    # https://github.com/ibis-project/ibis/pull/8671#issuecomment-2004438438
+    s = ibis.struct({"a": "foo", "b": "bar"})
+    with pytest.raises(TypeError):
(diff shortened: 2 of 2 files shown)


ibis/expr/types/structs.py
```diff
@@ -85,7 +86,7 @@ def struct(
 
 
 @public
-class StructValue(Value):
+class StructValue(Value, IbisMapping):
```
cpcloud, line 89: I'd really like to avoid the subclass if we can.

Is it sufficient to implement the necessary methods?

cpcloud (changes requested):

NickCrews:
looking into this.

[This SO quesiton implies we need the inheritance](https://stackoverflow.com/questions/11890066/making-custom-containers-work-with-kwargs-how-does-python-expand-the-args). But I'm looking into the cpython source to confirm, [here is as far into the cpython as I've dug.](https://github.com/python/cpython/blob/7707b14489644073ab0153f5751c6ddbf3fc6f91/Objects/dictobject.c#L3685)

NickCrews:
Looks like `.keys(self)` and `__getitem__(self, key)` are the only two needed methods for kwarg unpacking.

I removed `__len__(self)` for now because
1. it is not totally needed, you can do `len(struct.keys())`
2. it is not symmetric with the behavior of many other Values, like Columns and Tables, where `__len__` raises an error. That can be a separate question.

For a similar reason, I removed `__iter__` as well. Don't want people to do `list(struct_column)` and get confused.

ibis/expr/types/structs.py
NickCrews: See my comment on the PR, and the updated version of the code.

ibis/expr/types/structs.py
```diff
@@ -146,6 +146,15 @@ def __dir__(self):
     def _ipython_key_completions_(self) -> list[str]:
         return sorted(self.type().names)
 
+    def keys(self):
+        return self.names
```
NickCrews, line 150: If this looks good I can add docstrings

jcrist:
Given all the footguns you've raised above, I'd vote to _not_ have structs be valid mappings at all (especially since as you've done here, they're not really valid `collections.abc.Mapping` anyway, they implement a weird subset just enough that they can be `**kwargs` into a function). Doing this here may make the user expect it would work for `MapValue` (which it can't), or that `ArrayValue` would be a valid sequence (which also isn't possible).

Another (weak) reason is to avoid polluting the structvalue namespace - right now struct fields can be accessed by attribute, provided they don't collide with any methods/attributes on the class itself. With this PR, any structs with fields named `keys`/`values`/`items` would no longer be able to access these fields as attributes.

NickCrews:
Those all seem like good reasons @jcrist, I might have to agree with you. Maybe we can revisit this in some other way to improve the UX without adding these foot guns. I'm glad we worked through this as an exercise though

ibis/expr/types/structs.py
cpcloud: Seems fine.

Eventually we may want to cache the `values` and `items` results upon first use. Since expressions are immutable there's no reason to do those computations more than once. That be done later if it ever becomes an issue.

jcrist:
> Those all seem like good reasons @jcrist, I might have to agree with you.

Sounds good. Lets close this out then for now.

PR #8682, opened 2024-03-17
fix(deps): bump sqlglot to pick up duckdb array fixes

cpcloud:
Closes #8641. The fix was made upstream in sqlglot.

This PR ballooned into some other fixes:

1. We're segfaulting a lot because of duckdb/duckdb_spatial#236. I removed geospatial tests from the duckdb 0.10 CI job and added a 0.9.2 job that should mitigate this until the next duckdb geospatial release.
2. The sqlglot version bump introduced some changed SQL, so I updated the snapshots for that
3. The sqlglot verison bump changed mssql limit output to use `fetch` instead of `top`, which breaks the case of `limit 0`. `fetch first 0 rows only` raises a DB error, while `top 0` is allowed. I elected to xfail this for now but perhaps we should raise an issue on the sqlglot side?
4. Because of the sqlglot bump, log function calls are now generated using `log(x, y)` (base first or not depends on the dialect) without explicitly constructing an `Anonymous` function using `self.f.anon`. This exposed a missing `log` implementation in the sqlite backend which I added.
5. I addressed the failing `test_zip_null` test that was failing for the cloud backends. BigQuery xfails with an AssertionError because it turns NULL arrays into empty arrays and the fix for snowflake was to check for any NULL input arrays and return NULL accordingly.

PR #8683, opened 2024-03-17
fix(snowflake): handle null inputs to array zip

cpcloud:
Fixes failing tests on `main` due to failing to handle zip in the cloud backends. Depends on #8682.


cpcloud:
Closing in favor of #8682.

PR #8685, opened 2024-03-18
ci(flink): support setting flink's version in the docker setup

kszucs:
The used flink version now can be set at a centralised place which is going to be used to build the appropriate flink image and then run it.

```console
# spin up the current default flink version 1.18.1
just up flink

# use another version
FLINK_VERSION=1.17.1 just up flink
```


jcrist (approved):
Nice

kszucs:
@cpcloud do you use `.env` locally? Asking because I'm adding it in this PR.

cpcloud:
I don't use it.

PR #8689, opened 2024-03-19
test(bigquery): isolate dataset to user and python version to avoid concurrency issues during testing

cpcloud:
This PR isolates bigquery test and CI run datasets by creating a unique dataset per user per python version (up to the minor version), so that each developer can run the tests separate from each other as well separate from GitHub actions CI runs.

I had to remove a snapshot test from `test_interactve.py` due to it compiling in the resolved table name, including the BigQuery dataset name, which would change with every user or python version that ran, and would thus fail spuriously.


cpcloud:
It's possible this mitigates some CI performance issues as well. I suspect that the 3.9 job and the 3.11 job were overwriting and recreating test tables many times causing some thrashing.

gforsyth:
Looks like Bigquery finished in ~36 minutes, which is... not great, but better than it's been?

PR #8693, opened 2024-03-19
refactor(common): make `FrozenDict` a subclass of `dict`

### LINKED ISSUE #8687: bug: execution of struct literals in the pandas and dask backends returns the wrong value
### What happened?

```python
import pandas as pd
from ibis.common.collections import frozendict
pd.Series([frozendict({"a": 5, "b": 6})])
# 0    (a, b)
# dtype: object
```

This is causing casting bugs in the pandas and dask backends:

```python
import ibis

con = ibis.pandas.connect()

s = ibis.struct({"a": 1, "b": 2})
s2 = s.cast("struct<a: float64, b: float64>")
con.execute(s2)
# {'a': 'a'}
```

found this in https://github.com/ibis-project/ibis/pull/8666

I think we want to make it so our builtin frozendict looks exactly like a dict to pandas. Then we get the right behavior here

main

_No response_

### Relevant log output

_No response_

### Code of Conduct

jcrist:
This was motivated to work around pandas not repr'ing `frozendict` elements properly (seen in #8687), but while poking at that I found:

- The existing code was unnecessarily nested (the actual dict was boxed in a `MappingProxyType` which was boxed in a `FrozenDict` - we can do better by just using storing the data in the `FrozenDict` itself).
- There was a bug in the hash implementation where the order mattered, meaning that `hash(frozendict(a=1, b=2)) != hash(frozendict(b=2, a=1))`. This has since been fixed.

Fixes #8687.

### DIFF
--- a/ibis/common/collections.py
+++ b/ibis/common/collections.py
@@ -281,7 +281,7 @@ class FrozenDict(dict, Mapping[K, V], Hashable):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        hashable = frozenset(dict(self).items())
+        hashable = frozenset(self.items())
         object.__setattr__(self, "__precomputed_hash__", hash(hashable))
 
     def __hash__(self) -> int:
--- a/ibis/common/collections.py
+++ b/ibis/common/collections.py
@@ -281,7 +281,7 @@ class FrozenDict(dict, Mapping[K, V], Hashable):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        hashable = tuple(sorted(dict(self).items()))
+        hashable = frozenset(dict(self).items())
         object.__setattr__(self, "__precomputed_hash__", hash(hashable))
 
     def __hash__(self) -> int:
@@ -294,7 +294,7 @@ class FrozenDict(dict, Mapping[K, V], Hashable):
         raise TypeError(f"Attribute {name!r} cannot be assigned to frozendict")
 
     def __reduce__(self) -> tuple:
-        return (FrozenDict, (dict(self),))
+        return (self.__class__, (dict(self),))
 
 
 class RewindableIterator(Iterator[V]):
--- a/ibis/common/collections.py
+++ b/ibis/common/collections.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import collections.abc
 from abc import abstractmethod
 from itertools import tee
-from types import MappingProxyType
 from typing import TYPE_CHECKING, Any, Generic, TypeVar
 
 from public import public
@@ -276,42 +275,26 @@ class MapSet(Mapping[K, V]):
 
 
 @public
-class FrozenDict(Mapping[K, V], Hashable):
-    """Immutable dictionary with a precomputed hash value."""
-
-    __slots__ = ("__view__", "__precomputed_hash__")
-    __view__: MappingProxyType
+class FrozenDict(dict, Mapping[K, V], Hashable):
+    __slots__ = ("__precomputed_hash__",)
     __precomputed_hash__: int
 
     def __init__(self, *args, **kwargs):
-        dictview = MappingProxyType(dict(*args, **kwargs))
-        dicthash = hash(tuple(dictview.items()))
-        object.__setattr__(self, "__view__", dictview)
-        object.__setattr__(self, "__precomputed_hash__", dicthash)
+        super().__init__(*args, **kwargs)
+        hashable = tuple(sorted(dict(self).items()))
+        object.__setattr__(self, "__precomputed_hash__", hash(hashable))
 
-    def __str__(self):
-        return str(self.__view__)
+    def __hash__(self) -> int:
+        return self.__precomputed_hash__
 
-    def __repr__(self):
-        return f"{self.__class__.__name__}({dict(self.__view__)!r})"
+    def __setitem__(self, key: K, value: V) -> None:
+        raise TypeError("'FrozenDict' object does not support item assignment")
 
     def __setattr__(self, name: str, _: Any) -> None:
         raise TypeError(f"Attribute {name!r} cannot be assigned to frozendict")
 
-    def __reduce__(self):
-        return self.__class__, (dict(self.__view__),)
-
-    def __iter__(self):
-        return iter(self.__view__)
-
-    def __len__(self):
-        return len(self.__view__)
-
-    def __getitem__(self, key):
-        return self.__view__[key]
-
-    def __hash__(self):
-        return self.__precomputed_hash__
+    def __reduce__(self) -> tuple:
+        return (FrozenDict, (dict(self),))
 
 
 class RewindableIterator(Iterator[V]):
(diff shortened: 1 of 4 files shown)


ibis/common/collections.py
```diff
@@ -276,42 +275,26 @@ def isdisjoint(self, other: collections.abc.Mapping) -> bool:
 
 
 @public
-class FrozenDict(Mapping[K, V], Hashable):
-    """Immutable dictionary with a precomputed hash value."""
-
-    __slots__ = ("__view__", "__precomputed_hash__")
-    __view__: MappingProxyType
+class FrozenDict(dict, Mapping[K, V], Hashable):
```
cpcloud, line 278: Hm, can we avoid the subclass?

If someone wrote

```python
def f(mapping: dict): ...
```

Then it would be acceptable to pass an instance of `FrozenDict`, which cannot be mutated, despite being a `dict`. That's extremely surprising to me, and I suspect it would be to others.

jcrist: Not for this to fix the related issues no. Both are caused by explicit `isinstance(x, dict)` checks when they probably want `Mapping` instead.

We _could_ hide the `dict` base class from `mypy` so mypy at least would error on that case, but for runtime checking this would still be a dict subclass. Not sure if that's worth it.

FWIW I think `frozendict` is unlikely to creep into user-facing code. We might still want to coerce them to actual `dict` instances on `.execute()` to avoid this. But for internal-use only this fix seems like a definite win, both in performance and in correctness.

NickCrews: > Both are caused by explicit isinstance(x, dict)

You are saying you found somewhere in pandas code that is incorrect? If you point this to me I can file a PR upstream.

jcrist: Not sure if it's the complete or right fix, but https://github.com/pandas-dev/pandas/blob/aa3e949e2a2b72588186cb1936edb535713aefa0/pandas/io/formats/printing.py#L223 is one part of the repr issue. The execution issue is (likely) in our code, but is also fixed by the cleanup to make frozendicts actual dicts so I didn't bother delving further.

ibis/common/collections.py
```diff
@@ -276,42 +275,26 @@ def isdisjoint(self, other: collections.abc.Mapping) -> bool:
...
+    __slots__ = ("__precomputed_hash__",)
     __precomputed_hash__: int
 
     def __init__(self, *args, **kwargs):
-        dictview = MappingProxyType(dict(*args, **kwargs))
-        dicthash = hash(tuple(dictview.items()))
-        object.__setattr__(self, "__view__", dictview)
-        object.__setattr__(self, "__precomputed_hash__", dicthash)
+        super().__init__(*args, **kwargs)
+        hashable = tuple(sorted(dict(self).items()))
```
kszucs, line 284: Could we use `frozenset(self.items())` here?

ibis/common/collections.py
```diff
@@ -276,42 +275,26 @@ def isdisjoint(self, other: collections.abc.Mapping) -> bool:
...
-    def __len__(self):
-        return len(self.__view__)
-
-    def __getitem__(self, key):
-        return self.__view__[key]
-
-    def __hash__(self):
-        return self.__precomputed_hash__
+    def __reduce__(self) -> tuple:
+        return (FrozenDict, (dict(self),))
```
kszucs, line 297:
```suggestion
        return (self.__class__, (dict(self),))
```

ibis/common/collections.py
kszucs: > both in performance and in correctness

Could you add a quick benchmark for this?

kszucs: I am in favor of this change if benchmarks show a clear win.

kszucs: Quickly ran the graph traversal benchmarks which indeed improved:

```
---------------------------------------------------------------------------- benchmark 'test_bfs': 2 tests -----------------------------------------------------------------------------
Name (time in ms)              Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_bfs (0680_5041894)     2.2367 (1.11)     2.9631 (1.14)     2.3231 (1.08)     0.1120 (1.06)     2.2752 (1.06)     0.0923 (1.0)         25;14  430.4501 (0.93)        171           1
test_bfs (NOW)              2.0154 (1.0)      2.6029 (1.0)      2.1532 (1.0)      0.1060 (1.0)      2.1370 (1.0)      0.1417 (1.54)         79;6  464.4347 (1.0)         239           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

---------------------------------------------------------------------------- benchmark 'test_dfs': 2 tests -----------------------------------------------------------------------------
Name (time in ms)              Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_dfs (0680_5041894)     2.2076 (1.11)     2.6657 (1.18)     2.2614 (1.09)     0.0681 (1.0)      2.2378 (1.10)     0.0683 (1.0)         22;13  442.2093 (0.92)        189           1
test_dfs (NOW)              1.9952 (1.0)      2.2626 (1.0)      2.0778 (1.0)      0.0777 (1.14)     2.0388 (1.0)      0.1393 (2.04)         13;0  481.2894 (1.0)          46           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------- benchmark 'test_replace_mapping': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                          Min                Max               Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_replace_mapping (0680_5041894)     9.3224 (1.12)     28.3315 (1.04)     10.2140 (1.14)     3.1599 (1.62)     9.6292 (1.12)     0.2436 (1.0)           1;3   97.9049 (0.87)         35           1
test_replace_mapping (NOW)              8.3443 (1.0)      27.1707 (1.0)       8.9253 (1.0)      1.9461 (1.0)      8.5792 (1.0)      0.4105 (1.69)          1;1  112.0406 (1.0)          92           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_replace_pattern': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                           Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_replace_pattern (0680_5041894)     12.4369 (1.08)     32.3992 (1.07)     13.5555 (1.11)     3.2492 (1.52)     12.7563 (1.08)     0.7327 (1.60)          2;2  73.7707 (0.90)         62           1
test_replace_pattern (NOW)              11.5567 (1.0)      30.1479 (1.0)      12.1997 (1.0)      2.1385 (1.0)      11.8208 (1.0)      0.4588 (1.0)           1;2  81.9690 (1.0)          74           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

ibis/common/collections.py
cpcloud:
```suggestion
        hashable = frozenset(dict(self).items())
```

I'll add this suggestion to a batch and commit it. We can always revert it before merge if it causes issues.

cpcloud (approved):

PR #8701, opened 2024-03-19
refactor(cleanup): remove deprecated methods and properties

gforsyth:
`op` as a method on another operation was deprecated in 4.0.
`output_dtype` and `output_shape` were deprecated in 7.0.

I think 2 major versions is a reasonable deprecation cycle.

BREAKING CHANGE: Deprecated methods and properties `op`, `output_dtype`, and `output_shape` are removed. `op` is no longer needed, and use `.dtype` and `.shape` respectively for the other two.


jcrist (approved):

kszucs:
I am feeling nostalgic :)

PR #8703, opened 2024-03-19
feat(common): support removing elements from a `DisjointSet`

kszucs:
Item removal has been missing from the `DisjointSet` implementation which I am going to use in an upcoming PR.

depends on https://github.com/ibis-project/ibis/pull/8702

### DIFF
--- a/ibis/common/egraph.py
+++ b/ibis/common/egraph.py
@@ -6,6 +6,7 @@ import math
 from collections.abc import Callable, Hashable, Iterable, Iterator, Mapping
 from typing import Any, TypeVar
 
+from ibis.common.bases import FrozenSlotted as Slotted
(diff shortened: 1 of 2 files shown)


ibis/common/egraph.py
```diff
@@ -94,6 +95,19 @@ def __getitem__(self, id) -> set[K]:
...
+    def __delitem__(self, id):
+        """Remove the given id from the disjoint set.
+
+        Parameters
+        ----------
+        id :
+            The id to remove from the disjoint set.
+
+        """
+        self[id].remove(id)
```
kszucs, line 107: This is slightly more complicated than this, need to update the parent pointers.

kszucs:
Closing in favor of a follow-up PR with more robust testing.

PR #8707, opened 2024-03-20
ci(duckdb): work around flaky windows extension downloading

cpcloud:
Work around flaky windows extension downloading in parallel test runs in CI (observed in the wild [here](https://github.com/ibis-project/ibis/actions/runs/8357537262/job/22876913288#step:14:109)).


gforsyth (approved):
seems like a reasonable workaround!

PR #8709, opened 2024-03-20
perf(common): reduce the average recusrion depth in `_flatten_collections`

kszucs:


ibis/common/graph.py
```diff
@@ -56,19 +56,18 @@ def _flatten_collections(node: Any) -> Iterator[N]:
...
     >>> assert list(_flatten_collections((c,))) == [c]
     >>> assert list(_flatten_collections([a, b, (c, a)])) == [a, b, c, a]
+    >>> assert list(_flatten_collections([{"b": b, "a": a}])) == [b, a]
 
     """
-    if isinstance(node, Node):
-        yield node
-    elif isinstance(node, (tuple, list)):
-        for item in node:
+    for item in node:
```
cpcloud, line 64: Seems like we should perhaps bite the bullet and convert this to an iterative algorithm and remove recursion entirely instead of investing effort in smaller improvements.

cpcloud: I can put up a follow up.

cpcloud (approved):

ibis/common/graph.py
cpcloud: Seems within noise level to me. Both approaches are an improvement, and I definitely prefer the iterative version for two reasons:

1. There's effectively no contest for deeply nested collections, iteration will win every time in a language that doesn't implement tail recursion
2. The iterative approach uses far less memory, because it's not storing entire stack frames for every nesting level.

cpcloud: I'll repost this comment in #8713

PR #8723, opened 2024-03-21
feat(exasol): add support for extract seconds

Nicoretti:
## Description of changes

Add support for extract seconds to exasol backend.


cpcloud (approved):
Nice find!

Issue #8715, opened 2024-03-20
pyspark operation: WindowFunction


mark-druffel:
### Is your feature request related to a problem?

N/A

### Describe the solution you'd like

My team is trying to use Ibis to refactor our data pipeline. We would need access to a handful of operations to fully convert to Ibis from vanilla pyspark including Window. 

### What version of ibis are you running?

8.0.0

### What backend(s) are you using, if any?

pyspark

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

deepyaman:
Hi @mark-druffel! Just wanted to check, do you have an example of what's not working with the PySpark backend (but may be working on something else)? As far as I can tell, the PySpark backend does have window support implemented, and I see passing tests.

mark-druffel:
Hey @deepyaman, thanks for the quick reply! I may have jumped the gun on opening this 😬  

I was just trying to plan my team's work and knew we had some pipeline nodes a couple sprints out that would rely on windowing. I looked at the [compatibility matrix](https://ibis-project.org/support_matrix) and based on that I thought it wouldn't work. We haven't actually written any code that failed... I should've gone through the work to create a reproducible example before opening this, I'm sorry for that.

jcrist:
Thanks for opening this @mark-druffel! This led us to find a bug in the operation support matrix (a few other supported ops were erroneously marked as unsupported). It's still not perfect, but should be more accurate now.

PR #8729, opened 2024-03-21
fix: fix some missing operations in the support matrix

jcrist:
Previously `has_operation` for the SQL backends was ignoring operations supported through rewrite rules, leading to some gaps in the support matrix. Related to question in #8715. The pandas/dask backends still aren't correct - I started down fixing these up, but got bogged down in the magic of the `Dispatched` class and decided to punt for now.

Also deletes some dead code in the base sql backend.

### DIFF
--- a/ibis/backends/sqlite/tests/test_client.py
+++ b/ibis/backends/sqlite/tests/test_client.py
@@ -8,6 +8,7 @@ import pytest
 from pytest import param
 
 import ibis
+import ibis.expr.operations as ops
 from ibis.conftest import not_windows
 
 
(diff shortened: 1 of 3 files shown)


ibis/backends/sqlite/tests/test_client.py
```diff
@@ -75,3 +76,14 @@ def test_connect(url, ext, tmp_path):
     con = ibis.connect(url(path))
     one = ibis.literal(1)
     assert con.execute(one) == 1
+
+
+def test_has_operation(con):
```
jcrist, line 81: Arbitrarily decided to test this in the sqlite backend. I think a backend-specific test makes the most sense (since we want to ensure a variety of sources of handling the ops are covered, this kind of needs to be backend-specific).

cpcloud: Seems fine to me.

ibis/backends/sqlite/tests/test_client.py
```diff
@@ -75,3 +76,14 @@ def test_connect(url, ext, tmp_path):
...
+    # Handled by hardcoded rewrite
+    assert con.has_operation(ops.Project)
+    # Handled by base class rewrite
+    assert con.has_operation(ops.Capitalize)
+    # Handled by compiler-specific rewrite
+    assert con.has_operation(ops.Sample)
+    # Handled by visit_* method
+    assert con.has_operation(ops.Cast)
```
cpcloud, line 89: We're probably going to start rewriting `Aggregation` to `Select` so we can merge aggs with subqueries where possible, is there a test we can add that will fail if/when we do that?

jcrist: Could just add `Aggregate` to the ops checked here. Will do.

Issue #8459, opened 2024-02-26
feat: Generate descriptive statistics for a ibis table


jitingxu1:
### Is your feature request related to a problem?

I was a panda user, I used pandas dataframe `describe()` function a lot to get a sense of the data. I found ibis have the `info()` function, but it does not return enough information. 

### Describe the solution you'd like

Option 1: 
Pandas dataframe describe -[describe()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.describe.html)

Analyzes both numeric and object series,
Numeric columns:
- count, mean, std, min, max
- user defined percentiles, like [0.1, 0.5, 0.9]

For object data (e.g. strings or timestamps), the result’s index will include count, unique, top, and freq. The top is the most common value. The freq is the most common value’s frequency. Timestamps also include the first and last items.

```
df = pd.DataFrame({'categorical': pd.Categorical(['d', 'e', 'f']),
                   'numeric': [1, 2, 3],
                   'object': ['a', 'b', 'c']
                   })
df.describe()
       numeric
count      3.0
mean       2.0
std        1.0
min        1.0
25%        1.5
50%        2.0
75%        2.5
max        3.0
```

### What version of ibis are you running?

8.0.0

### What backend(s) are you using, if any?

DuckDB

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
I would look at implementing this for a few backends.

We explicitly chose not to add the quantiles because most of our backends do not support a quantile reduction.

amoeba:
Sorry for being the "well, in R..." but, well, in R we have a neat package called [skimr](https://github.com/ropensci/skimr) that produces some nice output and does some things that I think are smart. Here's an example:

```
> tbl(con, "example_df") |> 
  skimr::skim()

── Data Summary ────────────────────────
                           Values                
Name                       tbl(con, "example_df")
Number of rows             1000                  
Number of columns          3                     
_______________________                          
Column type frequency:                           
  character                1                     
  numeric                  2                     
________________________                         
Group variables            None                  

── Variable type: character ──────────────────────────────────────────────────────────────────────────────────────────
  skim_variable n_missing complete_rate min max empty n_unique whitespace
1 letters               0             1   1   1     0       26          0

── Variable type: numeric ────────────────────────────────────────────────────────────────────────────────────────────
  skim_variable n_missing complete_rate    mean      sd p0   p25    p50    p75  p100 hist 
[...]

PR #8739, opened 2024-03-22
feat(api): add `describe` method to compute summary stats of table expressions

jitingxu1:
## Description of changes
Add summary stats for a table. Only calculate stats for numeric, string, and bool columns. 
- name
- type
- count (counts and nulls)

For numeric columns:
- min
- mean
- std
- percentile
- max

For category columns:
- mode
- unique

For Boolean columns:
- mean

My questions:

- Not sure if it is a good practice to use `execute()` here. I have another way to implement this, but have to handle the null columns. 
- Is unit test necessary for this, noticed we do not have tests for `info()`

## Issues closed
Resolves https://github.com/ibis-project/ibis/issues/8459


ibis/expr/types/relations.py
```diff
@@ -2859,6 +2859,98 @@ def info(self) -> Table:
             aggs.append(agg)
         return ibis.union(*aggs).order_by(ibis.asc("pos"))
 
+    def describe(
+        self, quantile: Sequence[ir.NumericValue | float] = [0.25, 0.5, 0.75]
```
cpcloud, line 2863: Do not use a mutable data structure like a `list` or a `dict` as default argument. If that argument gets mutated _anywhere_ the mutation is reflected in the caller.

Here's an example:

```
In [1]: def do_stuff(x: list[int]):
   ...:     x.append(42)

In [2]: my_list = [1, 2, 3]

In [3]: do_stuff(my_list)

In [4]: my_list
Out[4]: [1, 2, 3, 42]

In [5]: do_stuff(my_list)

In [6]: my_list
Out[6]: [1, 2, 3, 42, 42]
```

In this case you can use a `tuple`, since that's an immutable sequence.

```suggestion
        self, quantile: Sequence[ir.NumericValue | float] = (0.25, 0.5, 0.75)
```

ibis/expr/types/relations.py
```diff
@@ -2859,6 +2859,98 @@ def info(self) -> Table:
...
+                one["std"] = col.std().execute()
+                for q in quantile:
+                    pname = (
+                        f"p{int(q * 100)}" if q * 100 == int(q * 100) else f"p{q * 100}"
+                    )
+                    one[pname] = col.quantile(q).execute()
+                one["max"] = col.max().execute()
+            elif typ.is_string():
+                one["mode"] = col.mode().execute()
+                one["unique"] = col.nunique().execute()
```
cpcloud, line 2950: All of these execute calls seem like the wrong approach. This will incur a lot of overhead in comms between the database and the client.

Is there an Ibis expression you can construct (perhaps using `ibis.union`?) that would allow you to compute the results you're after using a single query?

cpcloud (changes requested):

ibis/expr/types/relations.py
jitingxu1: Thanks, sounds great.

jitingxu1:
@cpcloud soft ping for the second review after rewriting the implementation. Thanks a lot for your feedback and time.

ibis/expr/types/relations.py
```diff
@@ -2859,6 +2859,123 @@ def info(self) -> Table:
...
+                type=lit(str(typ)),
+                count=col.isnull().count(),
+                nulls=col.isnull().sum(),
+                unique=col.nunique(),
+                mode=col.mode() if typ.is_string() else lit(None).cast(str),
+                mean=col.mean() if typ.is_numeric() else lit(None).cast(float),
+                std=col.std() if typ.is_numeric() else lit(None).cast(float),
+                min=col.min().cast(float)
+                if typ.is_numeric()
+                else lit(None).cast(float),
```
cpcloud, line 2955: Can you break this out into a separate `if` statement?

ibis/expr/types/relations.py
```diff
@@ -2859,6 +2859,123 @@ def info(self) -> Table:
...
+                **{
+                    f"p{int(q * 100)}"
+                    if q * 100 == int(q * 100)
+                    else f"p{q * 100}": col.quantile(q).cast(float)
+                    if typ.is_numeric()
+                    else lit(None).cast(float)
```
cpcloud, line 2961: Can you break this out?

ibis/expr/types/relations.py
```diff
@@ -2859,6 +2859,123 @@ def info(self) -> Table:
...
+                    for q in quantile
+                },
+                max=col.max().cast(float)
+                if typ.is_numeric()
+                else lit(None).cast(float),
```
cpcloud, line 2966: Can you break this out?

cpcloud (changes requested):

cpcloud (changes requested):
Can you add a test for this somewhere? Ideally wherever we test the info method (I think this is in `test_generic.py`).

jitingxu1:
> Can you add a test for this somewhere? Ideally wherever we test the info method (I think this is in `test_generic.py`).

Changes made:

- added unit test.
- Stats calculation now exclusively targets numeric, string, and boolean columns, disregarding others.

TODOs:
Certain backends encounter issues due to inadequate support for functionalities like quantile, mode, or standdev.

- The following backends need to implement mode to facilitate the full functionality of describe:
clickhouse, pyspark, clickhouse, risingwave, impala, oracle
- Both mode and StandardDev are required for these backends: exasol, druid


Other backends require both mode and quantile: "datafusion", "impala", "trino", "mysql", "mssql", "trino", "flink"

cpcloud (approved):
LGTM. I'll rebase it and merge on green.

Issue #8710, opened 2024-03-20
feat(flink): support semi / anti window joins


mfatihaktas:
### Is your feature request related to a problem?

Creating this issue to keep track of the work planned in https://github.com/ibis-project/ibis/issues/8250 as

> [P2] Window joins with more complex syntax (ANTI/SEMI) (follow-up to https://github.com/ibis-project/ibis/pull/7966)

### Describe the solution you'd like

Semi- or anti-join support in `join()` as
```
left_window_tvf = left_table.window_by( ... ).<window_table_valued_function>( ... )
right_window_tvf = right_table.window_by( ... ).<window_table_valued_function>( ... )

expr = left_window_tvf.join(
        right_window_tvf,
        [<col1>, <col2>],
        how="semi"
)
```

### What version of ibis are you running?

8.0.0

### What backend(s) are you using, if any?

Flink

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8745, opened 2024-03-22
feat(flink): support semi/anti window join

mfatihaktas:
## Description of changes

Aims to close https://github.com/ibis-project/ibis/issues/8710.

Notes to reviewers:
* Adds `AntiWindowJoin` and `SemiWindowJoin` sqlglot expressions. Would there be a way to avoid these? I could not find an existing `sqlglot` expression that we can use.
* Adds two new entries in `JoinKind`: `anti_window` and `semi_window`. Added a `TODO` with a question on if we can detect a join to be window join by looking for `WindowingTVF` in the expression tress for `left` and `right`. However, that "auto-window-join-detect" might lead to unexpected behavior, hence chose to make window-joins explicit with new `JoinKind` entries.

**Edit**: Got feedback from @kszucs over Zulip, and refactored the previously added `ops` and `expressions` mentioned above. Now instead, `JoinChain` is re-written with `Filter` and `ExistsSubquery` by `FlinkCompiler`.


ibis/expr/rewrites.py
```diff
@@ -297,3 +297,80 @@ def simplify(node):
...
+            filter=AnyOf(p.JoinLink | p.Comparison | p.Field | p.JoinTable),
+        )
+
+        # Rebase predicates on `node` if the underlying table for
+        # a predicate lies within the tree starting at `node`.
+        # TODO (mehmet): Is there a cleaner way to do this?
+        # Perhaps using `.replace()`?
+        predicates = []
+        for pred in link.predicates:
+            relations_under_node = node.find(ops.Relation)
```
kszucs, line 352: I assume this can be moved out of the two loops.

ibis/expr/rewrites.py
```diff
@@ -297,3 +297,80 @@ def simplify(node):
...
+
+@replace(p.JoinChain)
+def rewrite_join_chain_for_semi_anti_join(_, y=None):
+    # TODO (mehmet): Added optional arg `y` to work around the
+    # error raised for `test_many_subqueries()`.
+    # Is there a cleaner way to address this?
+
+    links = _.find_topmost(ops.JoinLink)
+    semi_anti_links, new_links = [], []
+    for link in links:
```
kszucs, line 321: You can directly iterate over `for link in _.rest`

ibis/expr/rewrites.py
```diff
@@ -297,3 +297,80 @@ def simplify(node):
...
+            semi_anti_links.append(link)
+        else:
+            new_links.append(link)
+
+    # Nothing to do with the original join chain
+    if not semi_anti_links:
+        return _
+
+    if new_links:
+        node = _.copy(rest=tuple(new_links))
```
kszucs, line 332: Just following the logic: so here we create a new join chain without semi/anti links.

ibis/expr/rewrites.py
```diff
@@ -297,3 +297,80 @@ def simplify(node):
...
+    else:
+        node = _.first.parent
```
kszucs, line 334: Cool, otherwise we have just the first table from the join chain.

ibis/expr/rewrites.py
```diff
@@ -297,3 +297,80 @@ def simplify(node):
...
+
+    for link in semi_anti_links:
+        # Note: `filter` is necessary to avoid rebasing predicates
+        # in the `JoinChain`'s deeper down the tree. E.g., when
+        # a table is semi/anti-joined with an inner-join as
+        # table.join(inner_join, predicates=..., how="semi/anti")
+        link = link.replace(
```
kszucs, line 341: I assume this won't be required since the predicates are replaced below.

ibis/expr/rewrites.py
```diff
@@ -297,3 +297,80 @@ def simplify(node):
...
+        # a predicate lies within the tree starting at `node`.
+        # TODO (mehmet): Is there a cleaner way to do this?
+        # Perhaps using `.replace()`?
+        predicates = []
+        for pred in link.predicates:
+            relations_under_node = node.find(ops.Relation)
+            if pred.left.rel in relations_under_node:
+                pred = pred.copy(left=pred.left.copy(rel=node))
+            if pred.right.rel in relations_under_node:
+                pred = pred.copy(right=pred.right.copy(rel=node))
```
kszucs, line 356: Please confirm, the idea here is to replace the fields referencing the join tables under the predicates. This could be done in a way similar to the following:

```py
relation_mapping = {link.table: node for link in new_links}
new_predicate = pred.replace(relation_mapping)
```

ibis/expr/rewrites.py
```diff
@@ -297,3 +297,80 @@ def simplify(node):
...
+        filtered = ops.Filter(parent=link.table.parent, predicates=predicates)
+
+        subquery = ops.ExistsSubquery(filtered)
+        if link.how == "anti":
+            subquery = ops.Not(subquery)
+        node = ops.Filter(parent=node, predicates=[subquery])
+
+    # Project to the columns specified in the initial `JoinChain`.
+    # TODO (mehmet): Is there a more efficient way to apply projection
+    # on `node`?
```
kszucs, line 369: Well, we can directly create a `Project` node but maybe this is not required since semi and anti joins shouldn't change the output schema.

ibis/expr/rewrites.py
mfatihaktas: Thanks for catching this, done.

ibis/expr/rewrites.py
mfatihaktas: Thanks, done.

ibis/expr/rewrites.py
mfatihaktas: It is required, otherwise we get the following error down the path:
```
    def test_semi_join_sql(semi_join_expr_and_expected_df, snapshot):
        expr, _ = semi_join_expr_and_expected_df
>       sql = ibis.to_sql(expr, dialect="flink")

ibis/backends/flink/tests/test_join.py:315:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
ibis/expr/sql.py:385: in to_sql
    sg_expr = backend._to_sqlglot(expr.unbind(), **kwargs)
ibis/backends/flink/__init__.py:346: in _to_sqlglot
    return super()._to_sqlglot(expr, params=params)
ibis/backends/sql/__init__.py:160: in _to_sqlglot
    sql = self.compiler.translate(table_expr.op(), params=params)
ibis/backends/sql/compiler.py:525: in translate
    op, ctes = sqlize(
ibis/backends/sql/rewrites.py:284: in sqlize
    node = node.replace(reduce(operator.or_, rewrites))
ibis/common/graph.py:464: in replace
    results = self.map(replacer, filter=filter)
ibis/common/graph.py:265: in map
ibis/common/graph.py:188: in fn
    if (result := obj.match(recreated, ctx)) is NoMatch:
ibis/common/patterns.py:932: in match
    result = pattern.match(value, context)
ibis/common/patterns.py:932: in match
    result = pattern.match(value, context)
ibis/common/patterns.py:932: in match
    result = pattern.match(value, context)
ibis/common/patterns.py:932: in match
    result = pattern.match(value, context)
ibis/common/patterns.py:932: in match
    result = pattern.match(value, context)
ibis/common/patterns.py:391: in match
    return self.replacer.resolve(context)
ibis/common/deferred.py:326: in resolve
    return self.func(**context)
ibis/expr/rewrites.py:436: in rewrite_join_chain_for_semi_anti_join
    filtered = ops.Filter(parent=link.table.parent, predicates=predicates)
ibis/common/bases.py:72: in __call__
    return cls.__create__(*args, **kwargs)
ibis/common/grounds.py:120: in __create__
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <ibis.expr.operations.relations.Filter object at 0x15e8f1e40>
parent = <ibis.expr.operations.temporal_windows.TumbleWindowingTVF object at 0x15e8f1c60>
predicates = (<ibis.expr.operations.logical.Equals object at 0x15eb18ac0>,)

    def __init__(self, parent, predicates):
        from ibis.expr.rewrites import ReductionLike

        for pred in predicates:
            if pred.find(ReductionLike, filter=Value):
                raise IntegrityError(
                    f"Cannot add {pred!r} to filter, it is a reduction which "
                    "must be converted to a scalar subquery first"
                )
            if pred.relations and parent not in pred.relations:
>               raise IntegrityError(
                    f"Cannot add {pred!r} to filter, they belong to another relation"
                )
E               ibis.common.exceptions.IntegrityError: Cannot add <ibis.expr.operations.logical.Equals object at 0x15eb18ac0> to filter, they belong to \
another relation
```

ibis/expr/rewrites.py
mfatihaktas: > Please confirm, the idea here is to replace the fields referencing the join tables under the predicates. 

Yes, this step is meant to "rebase" the fields based on the join tables (`ReferenceTable`'s as newly named) under the predicates.

Issue #8186, opened 2024-02-01
perf(dask): avoid triggering compute for dynamic limit/offset


kszucs:
> You should be able to avoid the extra compute here by doing a `map_partitions` call of a function that applies the `between` operation below. This will avoid excess compute if the values of `n`/`offset` are dependent on anything in `df`.
>
> Something like (untested):

  ```python
  # define this at the top-level for cheaper pickling
  def limit_chunk(df, col, n, offset):
      n = n.iat[0, 0]
      offset = offset.iat[0, 0]
      if n is None:
          return df[df[col] >= offset]
      else:
          return df[df[col].between(offset, offset + n - 1)]
  
  # then in the visit function, something like:
  return df.map_partitions(limit, col=name, n=n, offset=offset, align_partitions=False, meta=df._meta).drop(columns=[name])
  ```

_Originally posted by @jcrist in https://github.com/ibis-project/ibis/pull/8005#discussion_r1473081193_

PR #8747, opened 2024-03-22
perf(dask): avoid triggering compute for dynamic limit/offset

mfatihaktas:
## Description of changes

Aims to close https://github.com/ibis-project/ibis/issues/8186.

### DIFF
--- a/ibis/backends/dask/executor.py
+++ b/ibis/backends/dask/executor.py
@@ -31,6 +31,23 @@ from ibis.util import gen_name
 # ruff: noqa: F811
 
 
+def limit_df(
+    df: dd.DataFrame,
+    col: str,
+    n: int | pd.DataFrame,
+    offset: int | pd.DataFrame,
+):
+    if isinstance(offset, pd.DataFrame):
+        offset = offset.iat[0, 0]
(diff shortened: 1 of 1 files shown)


cpcloud:
With the caveat that I don't know much about Dask, the code no longer looks like it obviously triggers compute.

@jcrist @ncclementi Can one of y'all review?

jcrist (commented):
This is better than what's there currently, but we could still do better in the long run:

- For the common case of literal limits and no offset we should be able to take a more efficient path of calling `ddf.head(n, npartitions=-1)`. We're currently hampered by the rewrites in the pandas backend always upcasting these scalars to tables with a single element (making the literal scalar nature opaque to the dask backend).
- Even for dynamic limits we should be able to rely on `ddf.head(n, npartitions=-1)`, but that also only works if the input is represented as scalar and not a table.
- Only the case where `offset` is non-0 requires something like what we're doing here (though again if `offset` is a scalar instead of a table there are more ergonomic ways of spelling this). IMO offset is a rare enough case that if we wanted to error at compile time for non-zero `offset` in the dask backend that would be fine.

Anyway, this should be fine to go in as is.

cpcloud (approved):

PR #8751, opened 2024-03-23
fix(flink): fix compilation of memtable with nested data

### LINKED ISSUE #8516: bug(flink): memtables do not compile/execute correctly on Flink backend
### What happened?

`ibis.memtable()` expressions do not compile/execute correctly on Flink backend.

Example:
1. ```
t = ibis.memtable({"arr": [[], ["a"], ["a", "b"]]})
con.execute(t)
``` throws
```
Caused by: org.apache.calcite.runtime.CalciteContextException: From line 4, column 9 to line 4, column 15: Require at least 1 argument
```

3. ```
jobs = ibis.memtable(
    {
        "steps": [
            [
                {"status": "success"},
                {"status": "success"},
                {"status": None},
                {"status": "failure"},
            ],
            [
                {"status": None},
                {"status": "success"},
            ],
        ]
    },
    schema=dict(steps="array<struct<status: string>>"),
)
con.execute(jobs)
```
throws
```
Caused by: org.apache.calcite.sql.parser.SqlParseException: Encountered "AS" at line 4, column 27.
```

8.0

Flink

### Relevant log output

_No response_

### Code of Conduct

chloeh13q:
## Description of changes

This PR aims to fix the compilation of memtables with nested data.

### What was broken

In particular, [Flink does not support the `STRUCT(1 AS `a`)` aliasing syntax to define named STRUCTs](https://issues.apache.org/jira/browse/FLINK-9161). In order to do so, we must use a workaround using `CAST`, e.g.,
```sql
SELECT CAST(('a', 1) as ROW<a STRING, b INT>);
```

However, Flink also does not allow you to directly construct ARRAYs of named STRUCTs using the `ARRAY[]` constructor. This is a bug that I have filed with the Flink community (JIRA ticket ref: https://issues.apache.org/jira/browse/FLINK-34898).

For the time being, we will need to use another `CAST` workaound that casts the entire nested array, e.g.,
```sql
SELECT cast(ARRAY[ROW(1)] as ARRAY<ROW<a INT>>);
```

### How to fix

To summarize,
- if it’s an array of named structs
`CAST(ARRAY[] AS ARRAY<ROW<>, ROW<>>)`
- if it’s named structs
`CAST(ROW() AS ROW<datatype of each field>)`
- if it’s unnamed structs
`ROW()`

I thought of two approaches to this:
1. Rewrite the operator mapping in the Flink backend (i.e., change the `visit_NonNullLiteral()` method)
2. Rewrite the translation rule in Flink's `Generator`

I found both implementations in different scenarios. 

## Issues closed

#8516


ibis/backends/flink/tests/test_memtable.py
```diff
@@ -0,0 +1,38 @@
...
+            {"value": "!array<!struct<a: !int, b: !int>>"},
+            [Row([Row([1, 2]), Row([2, 2])])],
+            id="named_struct_array",
+        ),
+    ],
+)
+def test_create_memtable(con, data, schema, expected):
+    t = ibis.memtable(data, schema=ibis.schema(schema))
+    # cannot use con.execute(t) directly because of some behavioral discrepancy between
+    # `TableEnvironment.execute_sql()` and `TableEnvironment.sql_query()`
```
chloeh13q, line 35: I raised an issue on Flink JIRA

gforsyth (commented):
Gnarly -- nice job tracking this down!
Can you also add a test that will start to XPASS when upstream supports using the `ARRAY` constructor with named structs?

ibis/backends/sql/dialects.py
```diff
@@ -135,6 +134,69 @@ class Generator(Hive.Generator):
...
+                    else:
+                        schema.append(f"{self.sql(e, 'alias')} {self.sql(e.type)}")
+                    values.append(self.sql(e, "this"))
+                else:
+                    values.append(self.sql(e))
+
+            size = len(expression.expressions)
+
+            if not size or len(schema) != size:
+                if unknown_type:
```
gforsyth, line 162: Is there a scenario where we have an `unknown_type` but `len(schema) == size`?

If not, I think the error can get moved up to where you are setting `unknown_type`

ibis/backends/tests/test_map.py
```diff
@@ -240,7 +240,11 @@ def test_literal_map_getitem_broadcast(backend, alltypes, df):
                 ),
                 pytest.mark.notyet(["pandas", "dask"]),
                 mark_notyet_postgres,
-                pytest.mark.notimpl("flink"),
+                pytest.mark.broken(
```
gforsyth, line 243:
```suggestion
                pytest.mark.notyet(
```

chloeh13q: JW: is there a heuristic for when to use the different pytest markers? I've been confused about when to use `pytest.mark.notyet` vs `pytest.mark.broken` vs `pytest.mark.notimpl`

gforsyth: The ideas are scattered around a bit, should probably stick them into the dev docs.

* `notimpl`: the backend can do a thing, we haven't mapped the op
* `notyet`: the backend cannot do a thing, but might in the future
* `never`: the backend will never support this / pass this test (common example here is a test running on sqlite that relies on strong typing)
* `broken`: this test broke and it's demonstrably unrelated to the PR I'm working on and fixing it shouldn't block this PR from going in (but we should fix it up pronto)

chloeh13q: Thanks! I can put these into the dev docs

chloeh13q:
@gforsyth Thanks for the review! I addressed all of the comments.

ibis/backends/sql/dialects.py
```diff
@@ -135,6 +134,64 @@ class Generator(Hive.Generator):
...
+                    for e in arg.expressions:
+                        if isinstance(e, sge.Alias):
+                            named_structs = True
+                # get rid of aliasing because we want to compile this as CAST instead
+                args = deepcopy(expression.expressions)
+                if named_structs:
+                    for arg in args:
+                        arg.set("expressions", [e.this for e in arg.expressions])
+
+                format_values = ", ".join([self.sql(arg) for arg in args])
```
gforsyth, line 185:
```suggestion
            first_arg = seq_get(expression.expressions, 0).copy()
            if isinstance(first_arg, sge.Struct):
                for arg in expression.expressions:
                    arg.set("expressions", [e.unalias() for e in arg.expressions])

                format_values = ", ".join(
                    [self.sql(arg) for arg in expression.expressions]
                )
```

there's an `unalias` method on sqlglot objects that I think lets us simplify this a fair bit (but I might not be covering some corner cases you've run across)

gforsyth (commented):
Hey, I was staring at this a bit more and I think I have a cleaner way to handle the unaliasing pass -- lmk what you think.

ibis/backends/sql/dialects.py
chloeh13q: Thanks for the pointer - I like this solution and it definitely makes the code much cleaner! I'm tempted to modify on a deepcopy of  `expression.expression` to avoid unexpected consequences with passing this around...

gforsyth (approved):
LGTM, thanks for bearing with my multi-stage review, @chloeh13q !

PR #8758, opened 2024-03-25
fix(deps): support pandas 2.2

cpcloud:
Various fixes to support pandas 2.2.

## Non-test-related changes:

- Exasol: set the timezone to UTC; without this change the database results don't match pandas' results
- Pandas + Dask: handle new timedelta units
- SQLite: use a different method for computing epoch seconds; the existing method didn't match pandas' results
- Data conversion: handle the fact that `ValueError` is now raised when trying to convert from naive -> timezone'd data


kszucs (approved):
LGTM

Issue #8771, opened 2024-03-25
bug: struct field names in DDLs should be quoted


NickCrews:
### What happened?

found in #8765.

```python
import ibis

url = "https://storage.googleapis.com/ibis-debugging/election2020_single_chunk.jsonl"
t = ibis.read_json(url)
t = t.cache()
```

gives `ParserException: Parser Error: syntax error at or near "order"`

### What version of ibis are you using?

main

### What backend(s) are you using, if any?

duckdb

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
@NickCrews Thanks for the issue! It would appear that the yaks require shaving 😂

NickCrews:
🐄 🪒

cpcloud:
Based only on the error message I suspect something isn't quoted where it should be, since "order" is part of the "order by" keyword sequence.

NickCrews:
You are correct. post #8772, I see the sql is

```sql
CREATE TEMPORARY TABLE "ibis_cache_2qenquif65d65pb2apheifvvfu" (
    "data" STRUCT(
        races STRUCT(
            race_id TEXT,
            race_slug TEXT,
            url TEXT,
            state_page_url TEXT,
            ap_polls_page TEXT,
            race_type TEXT,
            election_type TEXT,
            election_date DATE,
            runoff BOOLEAN,
            race_name TEXT,
            office TEXT,
            officeid TEXT,
            race_rating TEXT,
            seat TEXT,
            seat_name TEXT,
            state_id TEXT,
            state_slug TEXT,
            state_name TEXT,
            state_nyt_abbrev TEXT,
            state_shape TEXT,
            state_aspect_ratio DOUBLE,
            party_id TEXT,
            uncontested BOOLEAN,
            report BOOLEAN,
            result TEXT,
            result_source TEXT,
            gain BOOLEAN,
            lost_seat TEXT,
            votes BIGINT,
            electoral_votes BIGINT,
            absentee_votes BIGINT,
            absentee_counties BIGINT,
            absentee_count_progress TEXT,
            absentee_outstanding JSON,
            absentee_max_ballots JSON,
            provisional_outstanding JSON,
            provisional_count_progress JSON,
            poll_display TEXT,
            poll_countdown_display TEXT,
            poll_waiting_display TEXT,
            poll_time TEXT,
            poll_time_short TEXT,
            precincts_reporting BIGINT,
            precincts_total BIGINT,
[...]

PR #8772, opened 2024-03-25
feat(duckdb): show action,SQL on ParserException/BinderException

NickCrews:
This would make errors such as https://github.com/ibis-project/ibis/issues/8771
easier to debug


cpcloud:
I don't think we should do this. There's certainly security _and_ privacy concerns here, there's also the fact that printing out SQL generated by Ibis might make the traceback unreadably long.

NickCrews:
OK, that's fair. Could we do something else to make debugging easier though?
- include the SQL in the custom error, eg 
```python
raise IbisError(
                "Failed to parse SQL. This is a bug in Ibis. Please report it at "
                "https://github.com/ibis-project/ibis/issues.",
                 sql=sql
            ) from e
```
and this custom error type doesn't include the SQL in the repr
- Create an ibis logger and log this at the warning level
- add some debug mode where this is logged? (could be other uses for this)

PR #8779, opened 2024-03-26
refactor(ir): remove `ops.WindowFrame` and simplify the binding process

kszucs:
- **refactor(ir): remove WindowFrame.table property and make WindowFrame a plain Node**
- **refactor(pandas): window frame rewrites**
- **refactor(ir): merge window frame with window function**


jcrist (approved):
Seems fine, no strong thoughts either way.

PR #8781, opened 2024-03-26
feat(risingwave): add streaming DDLs, new

KeXiangWang:
## Description of changes

In https://github.com/ibis-project/ibis/pull/8239 (A PR from company fork), I cannot allow maintainer to edit the PR. So I create this PR from my personal fork. All the changes in https://github.com/ibis-project/ibis/pull/8239 are included here.


ibis/backends/tests/test_join.py
```diff
@@ -175,6 +175,7 @@ def test_mutate_then_join_no_column_overlap(batting, awards_players):
 @pytest.mark.notimpl(["druid"])
 @pytest.mark.notyet(["dask"], reason="dask doesn't support descending order by")
 @pytest.mark.notyet(["flink"], reason="Flink doesn't support semi joins")
+@pytest.mark.skip("risingwave")  # TODO(Kexiang): Risingwave's bug, investigating
```
gforsyth, line 178:
```suggestion
@pytest.mark.broken("risingwave")  # TODO(Kexiang): Risingwave's bug, investigating
```

whatever the reason, we want to get a signal when this test starts to pass, and with `skip`, that change will not be visible.

lostmygithubaccount:
I'm going to close this in favor of #8239 (please correct/reopen if I'm interpreting this wrong)

ibis/backends/tests/test_join.py
KeXiangWang: The problem is that this case does not always failed. It actually can pass sometimes. But I haven't got time to explore under what conditions it will fail/pass. I think `skip` is more appropriate for now.

PR #8782, opened 2024-03-26
fix(druid): import pydruid.db module explicitly

gforsyth:
I tried connecting to the druid container and got an
`AttributeError: module 'pydruid' has no attribute 'db'`


cpcloud (approved):
Seems legit!

Issue #8019, opened 2024-01-18
Should date.to_pandas() return datetime64? 


NickCrews:
### What happened?

This is what currently happens on main, per a change I instigated in https://github.com/ibis-project/ibis/issues/7299:
```python
import ibis

ibis.options.interactive = True

d = ibis.memtable({"date": ["2024-01-01", "2024-01-02"]}).cast({"date": "date"}).date
s = d.to_pandas()
print(type(s[0]))
# <class 'datetime.date'>
s
# 0    2024-01-01
# 1    2024-01-02
# Name: date, dtype: object
```

Perhaps this is what is intended, but this annoyingly breaking for me. I was making plots with altair as
`alt.Chart(table.to_pandas())` using tables with a `date` column. Before, the pandas DF had type datetime64. This is serializable to JSON for altair/vega, and worked. Now, the `datetime.date` objects are not serializable to JSON, and I get an error. I think the old behavior was better, since .to_pandas() implies to me that we are going to pandas-land, and so we should use as canonical-to-pandas-as-possible dtypes. 

It seems like you really thought about the semantics of this already per some comments in that issue and linked PR, but curious if there would be a problem with these semantics:

- DateScalar.to_pandas() -> pd.Timestamp
- DateColumn.to_pandas() -> pd.Series[datetime64]
- DateScalar.execute() -> pd.Timestamp (because by definition this uses pandas)
- DateColumn.execute() -> pd.Series[datetime64] (same)
- repr(Date) -> "YYYY-MM-DD" (even if under the hood this uses .to_pandas() and we have pd.Timestamps in intermediate steps)

### What version of ibis are you using?

main

### What backend(s) are you using, if any?

NA

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hottwaj:
Just came across this too and wanted to add a vote for getting it fixed :)

mfatihaktas:
@hottwaj Thanks for chiming in. The fix is implemented in this [PR](https://github.com/ibis-project/ibis/pull/8784). It is waiting for the merge per [this comment](https://github.com/ibis-project/ibis/pull/8784#issuecomment-2200017418).

PR #8784, opened 2024-03-26
fix: return `pd.Timestamp` or `pd.Series[datetime64]` for `date.to_pandas()` [WIP]

mfatihaktas:
## Description of changes

Aims to close https://github.com/ibis-project/ibis/issues/8019.


ibis/backends/dask/convert.py
```diff
@@ -74,7 +74,7 @@ def convert_Date(cls, s, dtype, pandas_type):
         else:
             s = dd.to_datetime(s)
 
-        return s.dt.normalize()
```
cpcloud, line 77: I think might want to keep this, but not 100% sure. In theory once you're in this function you shouldn't need to normalize ... in theory :)

ibis/backends/oracle/converter.py
```diff
@@ -12,7 +13,11 @@ def convert_Timestamp_element(cls, dtype):
 
     @classmethod
     def convert_Date_element(cls, dtype):
-        return datetime.date.fromisoformat
+        # TODO (mehmet): Not sure if we should modify this... Could
+        # not run the tests on this backend. If we do, should we also
+        # modify `convert_Timestamp_element()` and
+        # `convert_Time_element()` similarly?
+        return pd.Timestamp.fromisoformat
```
cpcloud, line 20: This is sort of up for grabs given that pandas doesn't have a standard way to represent an array of dates (the `_element` suffix implies [perhaps not in an obvious way] that this function is being called once per element of an array).

I think it's fine to also change this to using pandas timestamps.

cpcloud (commented):
Approach looks reasonable, the devil is in the details as can be seen in CI 😂

cpcloud (commented):
Approach looks reasonable, the devil is in the details as can be seen in CI 😂

ncclementi:
@mfatihaktas What's the status of this one? would you mind fixing the conflicts and triggering CI to see if things are green and request a review.

mfatihaktas:
> @mfatihaktas What's the status of this one? would you mind fixing the conflicts and triggering CI to see if things are green and request a review.

Thanks for checking on it. Regarding the status, I marked this PR as ready-for-review some time ago. I am busy with another project this week, but will try to resolve the conflicts before EOD on Friday.

cpcloud:
I think this PR is good to go, but it's slated for 10.0, so please **do not merge**.

cpcloud (approved):
DO NOT MERGE

gforsyth:
I made the PR title non-compliant with conventional commits, so we can avoid accidentally clicking the big green button.

PR #8786, opened 2024-03-26
perf(duckdb): avoid ifs with literal array slice indices

NickCrews:
This is related to https://github.com/ibis-project/ibis/issues/8770

Post this change, the SQL for `ibis.array([1,2,3])[1:3]`  is

```sql
SELECT
  LIST_SLICE([CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT)], 2, 3) AS "ArraySlice(Array(), 1, 3)"`
```

Before this change, you got the monstrosity, which would evaluate `[CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT)]` a whopping 6 times, instead of the needed once:

```sql
SELECT
  LIST_SLICE(
    [CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT)],
    LEAST(
      LENGTH([CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT)]),
      CASE
        WHEN CAST(1 AS TINYINT) >= 0
        THEN CAST(1 AS TINYINT)
        ELSE ARRAY_LENGTH([CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT)]) + GREATEST(
          CAST(1 AS TINYINT),
          -ARRAY_LENGTH([CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT)])
        )
      END
    ) + 1,
    CASE
      WHEN CAST(3 AS TINYINT) >= 0
      THEN CAST(3 AS TINYINT)
      ELSE ARRAY_LENGTH([CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT)]) + GREATEST(
        CAST(3 AS TINYINT),
        -ARRAY_LENGTH([CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT)])
      )
    END
  ) AS "ArraySlice(Array(), 1, 3)"
```

This makes me wonder if we should come up with some framework for other backends to be able to take advantage of this easily. Make some Operation `ArrayIndex(arr, idx)`? IDK exactly how best to do this.

PR #8788, opened 2024-03-27
perf: remove some branching when using Literals

NickCrews:
A different implementation of https://github.com/ibis-project/ibis/pull/8786


NickCrews:
This is maybe a bit gross, but unless https://github.com/tobymao/sqlglot/issues/3229 lands I think this is the cleanest we can do?

kszucs:
@NickCrews can we close it in favor of https://github.com/ibis-project/ibis/pull/8832 ?

NickCrews:
I still want to add the static analysis parts of this PR that reduce branching, but I think that should be implemented as a rewrite instead.

The substring bits of this PR are indeed no longer relevant.

So yeah let's close

PR #8797, opened 2024-03-27
test: add test for non-1 step in String slicing

NickCrews:
Before, ibis.literal(1) or any ibis expression was a valid step in slices, but it shouldn't have been, since this argument
is ignored.

### DIFF
--- a/ibis/expr/types/strings.py
+++ b/ibis/expr/types/strings.py
@@ -92,7 +92,8 @@ class StringValue(Value):
 
         if isinstance(key, slice):
             start, stop, step = key.start, key.stop, key.step
-            if step is not None and not isinstance(step, ir.Expr) and step != 1:
+
+            if isinstance(step, ir.Expr) or (step is not None and step != 1):
                 raise ValueError("Step can only be 1")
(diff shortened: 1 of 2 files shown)


ibis/expr/types/strings.py
```diff
@@ -93,7 +93,7 @@ def __getitem__(self, key: slice | int | ir.IntegerScalar) -> StringValue:
         if isinstance(key, slice):
             start, stop, step = key.start, key.stop, key.step
 
-            if step is not None and not isinstance(step, ir.Expr) and step != 1:
+            if step is not None and step != 1:
```
cpcloud, line 96: Is this raising a `ValueError` when `step` is an Ibis expression because of the override of `__bool__` on `Expr`?

cpcloud (approved):
Assuming the `step` actually works the way I'm asking, this is good to go.

ibis/expr/types/strings.py
NickCrews: Oh, oops this is raising a `ValueError: The truth value of an Ibis expression is not defined` and the test is actually catching that. I want it to actually catch this custom error. fixing now.

PR #8798, opened 2024-03-27
docs(blog): portable dataflows with Ibis and Hamilton

zilto:
## Description of changes

Ported [a blog post from the Hamilton Substack](https://blog.dagworks.io/p/portable-dataflows-with-ibis-and) to the Quarto format for cross-posting on the Ibis blog. This was discussed with @lostmygithubaccount on Zulip.

First time using Quarto.  I managed to successfully build `.qmd` files and preview the post. Everything seemed fine. However, I see other blogs have files under `_freeze/posts/`, but I couldn't find how to generate them.


lostmygithubaccount:
thanks! I think this will fail the docs PR check, as `hamilton` will not be installed in the environment. you'd need to render the document yourself with quarto (via `quarto preview` or `quarto render`) and check in the changed `_freeze` files. this will have quarto use the saved execution output for the blog instead of trying to run it

cpcloud:
@lostmygithubaccount It's passing because the code blocks aren't executed.

lostmygithubaccount:
ah missed that -- will review soon! @zilto you may want to have the code blocks actually execute (and then my comment above would apply)

other offhand comment before a proper review would be wrapping the lines to ~90 characters -- see https://ibis-project.org/contribute/03_style#text-formatting

zilto:
> @lostmygithubaccount It's passing because the code blocks aren't executed.

aaah, that makes a lot sense. I was scratching my head at why no `_freeze/posts/hamilton-ibis` directory was generated.

Will make the changes

zilto:
I updated the `.qmd` file to use executable code blocks and generate the `_freeze/` file. 

Now, my code examples depend on a parquet dataset that is 23KB. I didn't commit it since `.parquet` files are git ignored by default. Should I include it to make the code runnable?

lostmygithubaccount:
@zilto apologies for the slow response here! 

> Now, my code examples depend on a parquet dataset that is 23KB. I didn't commit it since .parquet files are git ignored by default. Should I include it to make the code runnable?

I think this is small enough that just adding it is fine. alternatives would be: 

1. host this somewhere publicly that can be read from
2. switch to using a built-in `ibis.examples` dataset

neither really seem worth it in this case, and 2) would mean changing a lot of the code

gforsyth:
@zilto -- if you add the parquet file here (or you can send it over Zulip to me -- at least I think that's possible) I can stick it in one of our buckets and give you a public http link to pull from.
I agree checking it in is not a huge deal, but it's also easy to stick in the bucket and have a reasonably permanent link (in the medium-term, anyway)

lostmygithubaccount:
it is trying to run the Python code (and not finding the hamilton import), suggesting the execution freeze doesn't exactly match after you made some changes. can you try re-executing and ensuring it's up to date?

zilto:
Ok, odd bug... I just updated the files

cpcloud (approved):

PR #8801, opened 2024-03-27
feat(api): add `catalog` and `database` kwargs to `ibis.table`

### LINKED ISSUE #8253: feat(table): add namespace kwarg to ibis.table
A user asked about how to specify a table namespace when creating an unbound table.

The `UnboundTable` op does accept a namespace argument, but we don't expose that option in `ibis/expr/api.py::table`.

We will run afoul of the dual-definition of `schema` here, but if the user has to first construct the namespace object, at least that will be slightly separated.  

Still not a good solution, really, to the schema/schema issue, but I don't know what is, short of breaking all ibis code.

gforsyth:
This allows specifying 1 or 2 levels of table positional hierarchy in a
call to `ibis.table`. I've updated the reprs for `UnboundTable` and
`DatabaseTable` to show that hierarchy when it is specified.

I've also swapped the attribute order for `ops.Namespace` as a
convenience to allow unpacking the argnames in the correct hierarchical
order.

Resolves #8253

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -313,8 +313,14 @@ _table_names = (f"unbound_table_{i:d}" for i in itertools.count())
 def table(
     schema: SchemaLike | None = None,
     name: str | None = None,
+    catalog: str | None = None,
+    database: str | None = None,
 ) -> ir.Table:
     """Create a table literal or an abstract table without data.
+
+    Ibis uses the word database to refer to a collection of tables, and the word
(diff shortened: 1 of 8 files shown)


cpcloud:
Looks like rogue parenthesis decided it didn't need to exist!

cpcloud (approved):

cpcloud:
👌🏻 Nice UX improvement, thanks!

PR #8814, opened 2024-03-28
refactor(formats): remove unnecessary schema argument from schema inference

kszucs:
`sch.infer()` and `DataMapper.infer_table()` implementations had an unnecessary schema argument which wan't actually used.

### DIFF
--- a/ibis/formats/__init__.py
+++ b/ibis/formats/__init__.py
@@ -168,15 +168,13 @@ class DataMapper(Generic[S, C, T]):
         raise NotImplementedError
 
     @classmethod
-    def convert_table(cls, obj: T, schema: Schema) -> T:
+    def convert_table(cls, obj: T) -> T:
         """Convert a format-specific table to the given ibis schema.
 
         Parameters
         ----------
         obj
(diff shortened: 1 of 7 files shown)


cpcloud (approved):

ibis/formats/__init__.py
```diff
@@ -168,15 +168,13 @@ def convert_column(cls, obj: C, dtype: DataType) -> C:
         raise NotImplementedError
 
     @classmethod
-    def convert_table(cls, obj: T, schema: Schema) -> T:
+    def convert_table(cls, obj: T) -> T:
```
gforsyth, line 171: This now seems like a nonsensical function.  Convert it how?  and to what?  

If every child class that uses this defines a schema argument, shouldn't we keep it here?

cpcloud: Good catch, seems like this should stay.

gforsyth: I can put up a PR to restore it

gforsyth: https://github.com/ibis-project/ibis/pull/8821

Issue #8773, opened 2024-03-25
refactor(ir): loosen the ops.JoinChain integrity checks


kszucs:
### What happened?

Currently we require the join tables to have a monotonically increasing index whereas we only should check that all join tables are unique. This will enable us way simpler join chain rewrites without the need of reindexing the `JoinTable` nodes.

### What version of ibis are you using?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8817, opened 2024-03-28
refactor(ir): loosen the join integrity checks

kszucs:
Enables us easier join chain rewrites since we don't need to maintain a monotonically increasing index for the join tables.

Resolves https://github.com/ibis-project/ibis/issues/8773

### DIFF
--- a/ibis/expr/tests/test_newrels.py
+++ b/ibis/expr/tests/test_newrels.py
@@ -531,7 +531,6 @@ def test_project_before_and_after_filter():
     )
 
 
-# TODO(kszucs): add test for failing integrity checks
 def test_join():
     t1 = ibis.table(name="t1", schema={"a": "int64", "b": "string"})
     t2 = ibis.table(name="t2", schema={"c": "int64", "d": "string"})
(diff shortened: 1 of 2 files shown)


ibis/expr/tests/test_newrels.py
```diff
@@ -561,6 +560,26 @@ def test_join():
...
+        r1, [ops.JoinLink("inner", r2, [r1.a == r2.a])], values=values
+    )
+    assert isinstance(chain, JoinChain)
+
+    # not unique tables
+    r1 = ops.JoinTable(t1, 10).to_expr()
+    r2 = ops.JoinTable(t1, 10).to_expr()
+    values = {"a": r1.a, "b": r1.b, "c": r2.a, "d": r2.b}
+    with pytest.raises(IntegrityError):
+        ops.JoinChain(r1, [ops.JoinLink("inner", r2, [r1.a == r2.a])], values=values)
```
cpcloud, line 580: Can you construct only the exact thing that raising here? It's hard to tell what's raising just by reading the code.

cpcloud (changes requested):

ibis/expr/tests/test_newrels.py
kszucs: Updating.

kszucs: Reduced the example as much as I can.

ibis/expr/tests/test_newrels.py
```diff
@@ -561,6 +560,26 @@ def test_join():
...
+    chain = ops.JoinChain(r1, [ops.JoinLink("inner", r2, [True])], values={})
+    assert isinstance(chain, JoinChain)
+
+    # not unique tables
+    r1 = ops.JoinTable(t1, 10)
+    r2 = ops.JoinTable(t1, 10)
+    assert r1 == r2
+    assert hash(r1) == hash(r2)
+    with pytest.raises(IntegrityError):
+        ops.JoinChain(r1, [ops.JoinLink("inner", r2, [True])], values={})
```
cpcloud, line 580: Can't you move the join link construction out of the with block?

cpcloud (approved):

PR #8819, opened 2024-03-28
depr(api): deprecate `how` parameter to `arbitrary`

### LINKED ISSUE #8810: feat: `arbitrary`/`any_value` aggregate
Ibis columns have an existing `arbitrary` method on them for returning an "arbitrary" value in the column. Currently the signature is:

```
def arbitrary(self, how: Literal["first", "last", "heavy"] = "first"):
```

where the `how` defaults to returning the first value. Many backends have an `ANY_VALUE`/`ANY` aggregate that is supposed to return an arbitrary value (in many backends this just returns the first value, but in [bigquery](https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#any_value) and [snowflake](https://docs.snowflake.com/en/sql-reference/functions/any_value) this specifically returns an arbitrary value (distinctly not guaranteed to be first)).

A few questions/problems with the current `arbitrary` implementation:

- How do `col.arbitrary(how="first")` and `col.first()` differ (same for `last`)? They have different backing ops, but for most backends the SQL generated seems to be the same. Why have both spellings?
- The bigquery backend currently compiles `how="first"` to use `ANY_VALUE`, which AFAICT from the bigquery docs isn't always guaranteed to be the first value.
- For use cases where you do want an arbitrary value (and want to give the planner full freedom to optimize however it sees fit), would we want to add a `how="any"` to `arbitrary` (and possibly/probably make it the default `how`)? For backends that don't have an `ANY_VALUE` spelling we could still compile this to `first` as before.

jcrist:
This deprecates the `how` parameter to `arbitrary`.

The operation intent (supporting something like bigquery's `ANY_VALUE` or clickhouse's `ANY`) means that for any of these `how` parameters the output is nondeterministic and backend specific. Having `how="first"` may lull users into thinking this returns the "first" value in the column, when in reality that's not guaranteed (clickhouse, for example, says this is the "first encountered value of a column", but this is nondeterministic).

Since we had no guarantees about what the output for any of these `how` parameters was, I've opted to remove them fully from the implementation and instead warn if `how` is passed in (and then ignore `how`).

This also cleans up the implementations per backend to better reflect the intent behind `arbitrary`, and improved the test to better test that implementations are compatible.

Fixes #8810.


ibis/backends/snowflake/compiler.py
```diff
@@ -164,6 +164,9 @@ def visit_Literal(self, op, *, value, dtype):
             return sge.HexString(this=value.hex())
         return super().visit_Literal(op, value=value, dtype=dtype)
 
+    def visit_Arbitrary(self, op, *, arg, where):
+        return self.f.get(self.agg.array_agg(arg, where=where), 0)
```
jcrist, line 168: Unfortunately snowflake's `ANY_VALUE` doesn't have the same standard null-skipping behavior that other backends' implementations do. Since snowflake is the one out of compliance I decided that enforcing returning a non-null value (if possible) was more important than relying on the simpler `ANY_VALUE` builtin.

cpcloud: Hm, would `min` work instead of building a potentially large array?

jcrist: This is what we do for compiling `First` in snowflake, just mimicked it here.  I was thinking about implementing this as `min` for all other backends in a followup since that's till an arbitrary row if you think that'd make more sense.

cpcloud: Oh, this is pre-existing. Nevermind!

cpcloud (approved):
LGTM, thanks!

gforsyth (commented):
Language nit (guess who? Yes, it's Gil) -- I don't think this is a deprecation if the existing behavior no longer works.  I'm fine with just removing it and leaving the warning, but I think we should mark it as a breaking change, since behavior will change if someone was previously passing in `first`, no?

gforsyth:
Also, I just read the description more carefully.  Disregard plz.

PR #8820, opened 2024-03-28
refactor(sql): disable merging subsequent select nodes to avoid excessive inlining

kszucs:
For less usual use cases like https://github.com/ibis-project/ibis/issues/8484 merging subsequent projections can result in excessive inlining.

cc @NickCrews @cpcloud

### DIFF
--- a/ibis/backends/sql/rewrites.py
+++ b/ibis/backends/sql/rewrites.py
@@ -160,13 +160,13 @@ def merge_select_select(_, **kwargs):
     `predicates`, `sort_keys` and keeping the outer `selections`. All selections
     from the inner Select are inlined into the outer Select.
     """
-    # don't merge if either the outer or the inner select has window functions
-    for v in _.selections.values():
-        if v.find(Window, filter=ops.Value):
-            return _
+    # only merge subsequent projections if the inner projection is a simple
+    # containing only field references, this avoids extensive inlining causing
+    # the query to grow exponentially for complex expressions
     for v in _.parent.selections.values():
-        if v.find((Window, ops.Unnest), filter=ops.Value):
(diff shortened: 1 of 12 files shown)


NickCrews:
I would love to see the diff in SQL that this generates.

I am thinking about how to tune this for performance. I imagine some queries would be faster to inline, some faster to have a bunch of CTEs. Am I understanding this correclty that now, every `.select()` and `mutate()` in user code will result in exactly one new CTE? I'm not sure the best way to solve this "tuning" problem, but maybe leaving it up to users to choose the style they want:

```python
t = t.select(a=f(t.a))
t = t.select(a=g(t.a))
t = t.select(a=h(t.a))
```

results in a bunch of CTEs vs

```python
t = t.select(a=h(g(f(t.a)))))
```

results in a big inlined nesting.

I cant think of a better way to expose this tuning to users? Can you?

kszucs:
Well, this is a blocker so must be fixed for 9.0.

I assume you mean subselects under CTEs (because we consider a relation as a CTE if it occurs more than once). The generated queries are less nice now of course. I have a couple of ideas but first I need to try them out.

cpcloud:
We need to start without the tuning, and go back to letting the database optimize until we have evidence to the contrary.

There's no reason to optimize queries or give people footguns until we have some concrete evidence that non-inlined queries perform worse than the inlined ones.

Right now, we only have evidence that inlining makes things much worse for actual applications.

Pretty strong -1 on any "big" optimizations, let alone giving people to tune the level of optimization.

cpcloud:
Seems like this is ready to go barring the failures, which appear to be due to snapshot mismatches. I'l regenerate those now.

cpcloud:
Closing in favor of #8825

PR #8822, opened 2024-03-28
fix(clickhouse): use backwards compatible string search function

cpcloud:
Makes our implementation of string find and substring operations backwards compatible with clickhouse 23.


cpcloud:
I cherry-picked this onto #8809 locally and the related tests pass.

kszucs (approved):

PR #8825, opened 2024-03-29
fix(sql): avoid excessive inlining during `Select` merge

kszucs:
Another take on https://github.com/ibis-project/ibis/pull/8820 by introducing a complexity metric and only merge `Select` nodes if the outcome is less complex.

### DIFF
--- a/ibis/backends/sql/rewrites.py
+++ b/ibis/backends/sql/rewrites.py
@@ -119,6 +119,30 @@ def first_to_firstvalue(_, **kwargs):
     return _.copy(func=klass(_.func.arg))
 
 
+def complexity(node):
+    """Assign a complexity score to a node.
+
+    Subsequent projections can be merged into a single projection by replacing
+    the fields referenced in the outer projection with the computed expressions
+    from the inner projection. This inlining can result in very complex value
+    expressions depending on the projections. In order to prevent excessive
+    inlining, we assign a complexity score to each node.
+
+    The complexity score assigns 1 to each value expression and adds up in the
+    tree hierarchy unless there is a Field node where we don't add up the
+    complexity of the referenced relation. This way we treat fields kind of like
+    reusable variables considering them less complex than they were inlined.
+    """
+
+    def accum(node, *args):
+        if isinstance(node, ops.Field):
+            return 1
+        else:
+            return 1 + sum(args)
+
+    return node.map_nodes(accum)[node]
+
+
 @replace(Object(Select, Object(Select)))
 def merge_select_select(_, **kwargs):
     """Merge subsequent Select relations into one.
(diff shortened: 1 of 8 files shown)


ibis/backends/sql/rewrites.py
```diff
@@ -161,14 +161,13 @@ def merge_select_select(_, **kwargs):
...
-        if v.find((Window, ops.Unnest), filter=ops.Value):
-            return _
-    for v in _.predicates:
-        if v.find((ops.ExistsSubquery, ops.InSubquery), filter=ops.Value):
+
+    # TODO(kszucs): need a better api to find_under() a certain node without
+    # considering to filter the starting node out
+    for child in _.__children__ + _.parent.__children__:
+        if child.find(
+            (Window, ops.ExistsSubquery, ops.InSubquery, ops.Unnest), filter=ops.Value
```
kszucs, line 169: Really we should include scalar subqueries here as well since those can be correlated as well. The subqueries could be removed from here altogether if we don't stop traversing the predicates after reaching a non-value.

ibis/expr/operations/core.py
```diff
@@ -33,6 +33,11 @@ def equals(self, other) -> bool:
     # def __children__(self):
     #     return super().__children__
 
+    @attribute
+    def mass(self):
+        """Set of relations the value node depends on."""
+        return 1 + sum(n.mass for n in self.__children__)
```
cpcloud, line 39: Since you're introducing a heuristic here instead of disabling the optimization altogether, it needs some rationale in the docstring.

- How did you arrive at this particular calculation?
- Where does it breakdown?
- How does it address the problem that we observed with inlining?

cpcloud (changes requested):

cpcloud:
@NickCrews If you want to take this for a spin that would be great.

After this PR, the number of lines of SQL for 8.0.0 and in this PR are at the same order of magnitude, which IMO is enough of an improvement to merge it.

Here's a file with each version's generated SQL from the notebook provided in #8484:

[8.0.0-nb.txt](https://github.com/ibis-project/ibis/files/14836802/8.0.0-nb.txt)
[8825-nb.txt](https://github.com/ibis-project/ibis/files/14836803/8825-nb.txt)

(they're `.txt` because GH doesn't allow uploading `.sql` files for some reason)

kszucs:
@cpcloud I added docstrings to the complexity calculation.

cpcloud:
@kszucs Should this close #8484?

cpcloud (approved):

NickCrews:
Thank you so much for this! I just tested with the notebook in #8484 
- 8.0.0: fastest
- this PR: ~10% slower than 8.0.0
- main: 30+ times slower than 8.0.0

So this is definitely an improvement.

The theory of the implementation sounds right, even if the details go over my head a bit.

cpcloud:
Ideally over time we can get back the 10%, but probably not before the 9.0 release.

I think this is the last major blocker for 9.0.

PR #8826, opened 2024-03-29
test(bigquery): fix struct field quoting in type translation tests

cpcloud:
Fixes the two failing tests due to now quoting all struct field names.


cpcloud:
Merging to get CI passing again.

Issue #8712, opened 2024-03-20
chore: make args required-keywords in  make `read_postgres`, `read_sqlite` to be consistent with read_mysql


ncclementi:
Recently, in https://github.com/ibis-project/ibis/pull/8656 we discussed making the arguments of `read_mysql` required-keywords, and agreed that as a followup we will make this consistent in `read_postgres`and `read_sqlite`. 

Action Item: 
- Make args in `read_postgres`and `read_sqlite` required-keyword

iangow:
Not clear that this "consistency" is helpful, as it introduces inconsistencies elsewhere. If you search [here](https://ibis-project.org/backends/duckdb#ibis.backends.duckdb.Backend), you will see many references to `schema = `. So sometimes for PostgreSQL users, it will be `schema = ` and other times `database =`. (One can have multiple databases, each with multiple schemas in PostgreSQL.)

gforsyth:
Hey @iangow -- hopefully on that page any reference to `schema=` will be exclusively related to an Ibis schema mapping column names to dtypes.

If there are any remaining references to schema as a hierarchical location mapping, those should definitely be removed.

iangow:
> Hey @iangow -- hopefully on that page any reference to `schema=` will be exclusively related to an Ibis schema mapping column names to dtypes.
> 
> If there are any remaining references to schema as a hierarchical location mapping, those should definitely be removed.

It seems there are several places where this is *not* the case. Sometimes this argument is listed as `[deprecated]`, but not always (e.g., when there's no documentation). For `read_postgres()` with the DuckDB backend, there was (it seems) no deprecation … the argument just disappeared.

Not sure that every PostgreSQL user (`read_postgres()`) will know that "database" is being used in the MySQL sense without more documentation.

iangow:
This note appears at times:

```
::: {.callout-note} ## Ibis does not use the word schema to refer to database hierarchy. A collection of tables is referred to as a database. A collection of database is referred to as a catalog. These terms are mapped onto the corresponding features in each backend (where available), regardless of whether the backend itself uses the same terminology. :::
```

It doesn't render properly in most cases. It looks good under `list_databases()`, but not under `insert()` or `list_tables()` or `truncate_table()`. Perhaps this note should also go in `read_postgres()`.

gforsyth:
If an API is marked experimental we may have just removed it outright -- I'll look at the docs rendering issues and also add a note to `read_postgres`

PR #8829, opened 2024-03-29
chore: make arguments required-keyword for read_postgres and read_sqlite

ncclementi:
In https://github.com/ibis-project/ibis/pull/8656 we made the args for `read_mysql`  required-keyword. 

This PR, makes the args for `read_postgres` and `read_sqlite` to be  required-keyword to be consistent. 
It also fixes a broken test that runs only locally, `test_read_postgres` (thanks @gforsyth for the patch suggestion). 

- Closes https://github.com/ibis-project/ibis/issues/8712

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -987,7 +987,7 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
         return self._filter_with_like(out[col].to_pylist(), like)
 
     def read_postgres(
-        self, uri: str, table_name: str | None = None, schema: str = "public"
+        self, uri: str, *, table_name: str | None = None, database: str = "public"
     ) -> ir.Table:
         """Register a table from a postgres instance into a DuckDB table.
 
@@ -997,8 +997,8 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
             A postgres URI of the form `postgres://user:password@host:port`
         table_name
             The table to read
-        schema
-            PostgreSQL schema where `table_name` resides
+        database
+            PostgreSQL database (schema) where `table_name` resides
 
         Returns
         -------
@@ -1015,7 +1015,7 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
         self._create_temp_view(
             table_name,
             sg.select(STAR).from_(
-                self.compiler.f.postgres_scan_pushdown(uri, schema, table_name)
+                self.compiler.f.postgres_scan_pushdown(uri, database, table_name)
             ),
         )
 
@@ -1023,8 +1023,8 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
 
     def read_mysql(
         self,
-        *,
         uri: str,
+        *,
         catalog: str,
         table_name: str | None = None,
     ) -> ir.Table:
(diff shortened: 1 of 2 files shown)


ibis/backends/duckdb/__init__.py
```diff
@@ -987,7 +987,7 @@ def list_tables(
         return self._filter_with_like(out[col].to_pylist(), like)
 
     def read_postgres(
-        self, uri: str, table_name: str | None = None, schema: str = "public"
+        self, *, uri: str, table_name: str | None = None, schema: str = "public"
```
cpcloud, line 990: Let's keep `uri` non-keyword.

```suggestion
        self, uri: str, *, table_name: str | None = None, schema: str = "public"
```

ibis/backends/duckdb/__init__.py
```diff
@@ -1062,7 +1062,9 @@ def read_mysql(
 
         return self.table(table_name, database=(database, catalog))
 
-    def read_sqlite(self, path: str | Path, table_name: str | None = None) -> ir.Table:
+    def read_sqlite(
+        self, *, path: str | Path, table_name: str | None = None
```
cpcloud, line 1066:
```suggestion
        self, path: str | Path, *, table_name: str | None = None
```

cpcloud (dismissed):

ibis/backends/duckdb/__init__.py
ncclementi: Just to be sure, because in `read_mysql`, we made `uri` required, do we want to revert that? I think we should be consistent across the board.

ibis/backends/duckdb/__init__.py
ncclementi: same comment than above, we made `uri` required in `read_mysql`

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -234,9 +238,6 @@ def test_read_sqlite(con, tmp_path):
     ft = con.read_sqlite(path, table_name="t")
     assert ft.count().execute()
 
-    with pytest.raises(ValueError):
-        con.read_sqlite(path)
-
```
gforsyth, line 239: Does this no longer raise?

ncclementi: I should have put a note, this is already tested in a separate test, I thought it was redundant see

https://github.com/ibis-project/ibis/blob/9cb4c8d17f9d7960c216b324a0153b9c785ea09f/ibis/backends/duckdb/tests/test_register.py#L241-L249

gforsyth: Ahh, gotcha.  thanks!

ibis/backends/duckdb/__init__.py
```diff
@@ -987,7 +987,7 @@ def list_tables(
         return self._filter_with_like(out[col].to_pylist(), like)
 
     def read_postgres(
-        self, uri: str, table_name: str | None = None, schema: str = "public"
+        self, uri: str, *, table_name: str | None = None, schema: str = "public"
```
gforsyth, line 990:
```suggestion
        self, uri: str, *, table_name: str | None = None, database: str = "public"
```

sorry, just caught this.  we want to avoid re-introducing this language

gforsyth (changes requested):
One change, then I think this is good.

I'm fine with the docstring of `read_postgres` saying something like 

```python
database
    PostgreSQL database (schema) where `table_name` resides
```

but I'd like to avoid having `schema` as the argument name in Ibis

gforsyth (approved):
:shipit:

Issue #8796, opened 2024-03-27
perf: natively support `String.slice(start, end)` as well as `String.substr(start, length)`


NickCrews:
### Is your feature request related to a problem?

Currently, `ibis.literal("abc")[1:]` gets compiled to `SUBSTRING('abc', 2, LENGTH('abc') - CAST(1 AS TINYINT))` on duckdb. However,  duckdb has a builtin [`array_slice(string, start, end)`](https://duckdb.org/docs/sql/functions/char#array_slicelist-begin-end) function, that would 1. be simpler SQL and would avoid duplicating compiling the subexpression `'abc'` 2. probably be faster.

### Describe the solution you'd like

I think we need to have a data model in ibis that can represent both kinds of specification, (start, length) and (start, end). I could see this being either

1. Add a `StringSlice` Op in addition to the existing `Substring` op
2. Modify the existing `Substring` op so that it can either take a `length` or an `end`.

I would lean towards 1, I bet that would lead to less ifelse branching throughout the code.

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kszucs:
I think a `StringSlice` operation would be handy which we can rewrite to `Substring` for backends not natively supporting negative indices.

PR #8832, opened 2024-03-30
feat(ir): add `StringSlice`

NickCrews:
Fixes https://github.com/ibis-project/ibis/issues/8796

### DIFF
--- a/ibis/backends/tests/test_string.py
+++ b/ibis/backends/tests/test_string.py
@@ -563,107 +563,6 @@ def uses_java_re(t):
             id="rstrip",
         ),
         param(
-            lambda t: t.date_string_col.substr(2, 3),
-            lambda t: t.date_string_col.str[2:5],
-            id="substr",
-        ),
-        param(
-            lambda t: t.date_string_col.substr(2),
-            lambda t: t.date_string_col.str[2:],
-            id="substr-start-only",
-            marks=[
-                pytest.mark.broken(
-                    ["mssql"],
-                    reason="substr requires 3 arguments",
-                    raises=PyODBCProgrammingError,
-                ),
-            ],
-        ),
-        param(
-            lambda t: t.date_string_col.left(2),
-            lambda t: t.date_string_col.str[:2],
-            id="left",
-        ),
-        param(
-            lambda t: t.date_string_col.right(2),
-            lambda t: t.date_string_col.str[-2:],
-            id="right",
-        ),
-        param(
-            lambda t: t.date_string_col[1:3],
-            lambda t: t.date_string_col.str[1:3],
-            id="slice",
-        ),
-        param(
-            lambda t: t.date_string_col[-2],
-            lambda t: t.date_string_col.str[-2],
-            id="negative-index",
-        ),
-        param(
-            lambda t: t.date_string_col[t.date_string_col.length() - 1 :],
-            lambda t: t.date_string_col.str[-1:],
-            id="expr_slice_begin",
-            # TODO: substring #2553
-            marks=[
-                pytest.mark.notimpl(
-                    ["polars"],
-                    raises=com.UnsupportedArgumentError,
-                    reason=(
-                        "Polars does not support columnar argument Subtract(StringLength(date_string_col), 1)"
-                    ),
-                ),
-                pytest.mark.xfail_version(datafusion=["datafusion==35"]),
-            ],
-        ),
-        param(
-            lambda t: t.date_string_col[: t.date_string_col.length()],
-            lambda t: t.date_string_col,
-            id="expr_slice_end",
-            # TODO: substring #2553
-            marks=[
-                pytest.mark.notimpl(
-                    ["polars"],
-                    raises=com.UnsupportedArgumentError,
-                    reason=(
-                        "Polars does not support columnar argument Subtract(StringLength(date_string_col), 1)"
-                    ),
-                ),
-            ],
-        ),
-        param(
-            lambda t: t.date_string_col[:],
-            lambda t: t.date_string_col,
-            id="expr_empty_slice",
-            # TODO: substring #2553
-            marks=[
-                pytest.mark.notimpl(
-                    ["polars"],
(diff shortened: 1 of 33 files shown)


ibis/expr/operations/strings.py
```diff
@@ -63,6 +63,16 @@ class Substring(Value):
     shape = rlz.shape_like("arg")
 
 
+@public
+class StringSlice(Value):
+    arg: Value[dt.String]
+    start: Optional[Value[dt.Integer]] = None
+    end: Optional[Value[dt.Integer]] = None
```
NickCrews, line 70: a better name would be (start, stop), but python uses (start, end) in slices, so I decided to be consistent with that.

ibis/expr/types/strings.py
```diff
@@ -58,25 +58,25 @@ def __getitem__(self, key: slice | int | ir.IntegerScalar) -> StringValue:
...
-        ┃ Substring(food, 0, 3) ┃
-        ┡━━━━━━━━━━━━━━━━━━━━━━━┩
-        │ string                │
-        ├───────────────────────┤
-        │ bre                   │
-        │ che                   │
-        │ ric                   │
-        └───────────────────────┘
+        ┏━━━━━━━━━━━━━━━━━━━━━━┓
+        ┃ StringSlice(food, 3) ┃
```
NickCrews, line 62: this repr isn't great: both [:3] and [3:] result in `StringSlice(food, 3)`. But that is separate problem.

ibis/tests/expr/test_string.py
```diff
@@ -96,16 +95,6 @@ def test_contains(table):
         "foo" in table.g  # noqa: B015
 
 
-@pytest.mark.parametrize(
-    ("left_slice", "right_start", "right_stop"),
-    [(slice(None, 3), 0, 3), (slice(2, 6), 2, 4)],
-)
-def test_getitem_slice(table, left_slice, right_start, right_stop):
```
NickCrews, line 103: this test is checking for an implementation detail, it probably should never have existed.

ibis/backends/tests/test_string.py
```diff
@@ -663,60 +676,13 @@ def uses_java_re(t):
...
-        param(
-            lambda t: t.string_col.replace("1", "42"),
-            lambda t: t.string_col.str.replace("1", "42"),
-            id="replace",
-        ),
     ],
 )
-def test_string(backend, alltypes, df, result_func, expected_func):
-    expr = result_func(alltypes).name("tmp")
+def test_substring(backend, alltypes, df, result_func, expected_func):
```
NickCrews, line 681: the diff for these test cases is really messy, sorry. What I did:
- move some test cases to separate test_substring test
- remove from xfails/brokens since I fixed/implemented things
- removed no test cases
- added the `positive-index` test case

cpcloud: Thanks for the explanation, I was going to ask why you deleted test cases 😂

ibis/tests/expr/test_string.py
cpcloud: Agreed.

cpcloud (approved):
LGTM, question about uncovered lines when raising errors at the API level.

ibis/expr/types/strings.py
```diff
@@ -92,29 +92,20 @@ def __getitem__(self, key: slice | int | ir.IntegerScalar) -> StringValue:
...
-                if stop is not None and stop < 0:
-                    raise ValueError(
-                        "Negative slicing not yet supported, got stop value "
-                        f"of {stop:d}"
-                    )
-                if stop is None:
-                    stop = self.length()
-
-            return self.substr(start, stop - start)
+            if start is not None and not isinstance(start, ir.Expr) and start < 0:
```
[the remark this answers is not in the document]
NickCrews: there is not currently a test case. Agreed should add one. In this PR or a followup?

ibis/expr/rewrites.py
```diff
@@ -82,6 +82,26 @@ def rewrite_dropna(_):
...
+@replace(p.StringSlice)
+def rewrite_stringslice(_, **kwargs):
+    """Rewrite StringSlice in terms of Substring."""
+    if _.end is None:
+        return ops.Substring(_.arg, start=_.start)
+    if _.start is None:
+        return ops.Substring(_.arg, start=0, length=_.end)
+    if (
+        isinstance(_.start, ops.Literal)
+        and isinstance(_.start.value, int)
```
kszucs, line 94: The rewrite rule is good as is, just noting that this could be spelled as a pattern as well: `p.StringSlice(p.Literal(int), p.Literal(int)`.

ibis/expr/types/strings.py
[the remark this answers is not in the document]
cpcloud: Can be done in a follow-up.

kszucs:
Thanks @NickCrews!

ibis/expr/rewrites.py
NickCrews: As in there would be a different rule for each of the patterns? (Literal, None), (None, literal), (literal, literal)? Maybe a full example of how this could be spelled? If there are different rules then would I need to import three different rewrites in all the compilers?

I'd love to learn for next time.

kszucs: I think this is ideal in this current scenario, just wanted to highlight you that nested patterns are also supported. 

There are various ways to spell this, here is a more complicated one showing what can be done with the rewrite system:

```py
from ibis.expr.rewrites import p, d

p.StringSlice(end=None) >> d.Substring(_.arg, start=_.start) |
p.StringSlice(start=None) >> d.Substring(_.arg, start=0, length=_.end) |
p.StringSlice(p.Literal(int), p.Literal(int)) >> d.Substring(_.arg, start=_.start, length=_.end.value - _.start.value) |
p.StringSlice >> d.Substring(_.arg, start=_.start, length=d.Subtract(_.end, _.start)
```

PR #8834, opened 2024-03-30
chore(deps): downgrade nixpkgs

cpcloud:
Downgrade nixpkgs to avoid picking up newer versions of xz. This is a developer-tools-only issue. We can revisit the 312 support in the nix job once nixos-unstable-small finishes building (it has the xz revert patch applied).


cpcloud:
Merging.

PR #8860, opened 2024-04-02
feat(pandas, polars, dask, datafusion): enable create local backends with empty url

### LINKED ISSUE #8450: feat: support empty URL for local backends in ibis.connect()

jitingxu1:
## Description of changes
support empty URL for local backends in ibis.connect()

## Issues closed
* Resolves #8450

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -8,6 +8,7 @@ import keyword
 import re
 import sys
 import urllib.parse
+import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, ClassVar
 from urllib.parse import parse_qs, urlparse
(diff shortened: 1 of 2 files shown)


ibis/backends/__init__.py
```diff
@@ -1458,5 +1459,26 @@ class NoUrl:
...
+        -------
+        BaseBackend
+            A backend instance
+
+        """
+        # Check if additional components are provided in the URL
+        if len(url.split("://")) > 1:
+            warnings.warn(
+                f"Additional components in the URL are ignored. url = '{url}'"
+            )
```
gforsyth, line 1483: I would remove the warning entirely, but if you'd like to keep it, I think you'll want to either use `urlparse` or check that the last element of the `split` call is not an empty string.

jitingxu1: Thanks for your feedback. I will remove the warning.

gforsyth (approved):

PR #8861, opened 2024-04-02
feat(common): add `Node.find_below()` methods to exclude the root node from filtering

kszucs:
Currently we are unable to locate certain values under a relation while applying a value filter because the root node gets filtered out, like:

```py
relation.find(pattern, filter=ops.Value)
```

Used in practice at https://github.com/ibis-project/ibis/pull/8825

```py
for child in _.__children__ + _.parent.__children__:
    if child.find((Window, ops.ExistsSubquery, ops.InSubquery, ops.Unnest), filter=ops.Value):
        ...
```
   
Which can be simplified to `_.find_below() or _.parent.find_below()` after this feature.

I am not that satisfied with the name, but marked as experimental so we can change it anytime. 
This feature will also help with pretty printing the dereference mapping without traversing the dictionary keys in `_flatten_collections()`.

### DIFF
--- a/ibis/common/graph.py
+++ b/ibis/common/graph.py
@@ -11,7 +11,7 @@ from ibis.common.bases import Hashable
 from ibis.common.collections import frozendict
 from ibis.common.patterns import NoMatch, Pattern
 from ibis.common.typing import _ClassInfo
-from ibis.util import experimental
+from ibis.util import experimental, promote_list
 
 if TYPE_CHECKING:
     from typing_extensions import Self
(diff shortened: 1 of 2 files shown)


ibis/common/graph.py
```diff
@@ -340,9 +340,39 @@ def find(
...
 
         """
-        nodes = Graph.from_bfs(self, filter=filter, context=context).nodes()
+        graph = Graph.from_bfs(self, filter=filter, context=context)
         finder = _coerce_finder(finder, context)
-        return [node for node in nodes if finder(node)]
+        return [node for node in graph.nodes() if finder(node)]
+
+    @experimental
+    def find_below(
```
cpcloud, line 348: Are you planning on using this somewhere?

cpcloud (approved):
Seems fine, just wondering what the plans are for `find_below`.

ibis/common/graph.py
kszucs: @cpcloud added description

Issue #8735, opened 2024-03-21
chore: deprecate register functionality 


ncclementi:
> I think we should remove the register function entirely -- it predates most of the other more specific methods and it's a mess (and is completely untested).

For context see https://github.com/ibis-project/ibis/issues/8728

@gforsyth  please add any context that I might have missed here

Issue #6593, opened 2023-07-06
feat: Standardize `read_in_memory` across backends


jcrist:
### Is your feature request related to a problem?

Originally we had `con.register` to handle data ingestion of anything a backend could support. This method works fine and is defined for many backends, but is a bit "magical" in that it determines the input format based on the type and value of the input. Later on we split part of this functionality out into standard `read_csv`/`read_parquet`/`read_*` methods, which `register` dispatches to.

The duckdb backend also has a `read_in_memory` method for reading from an in-memory object. Unlike files (where we dispatch based on things like the file extension), in memory data always has an explicit python type, so having a method like `read_in_memory` to handle any in-memory data sources seems like a nice interface. It would be good to standardize this interface across all the backends that could support it. For the most part this would be extracting out the existing functionality from the `.register` methods and moving it to a new `read_in_memory` method.

### Describe the solution you'd like

For any backend that could support it to define a method like:

```python
def read_in_memory(self, source: pd.DataFrame | pa.Table | pa.RecordBatchReader, table_name: str | None = None) -> ir.Table:
    ...
```

### What version of ibis are you running?

dev

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chloeh13q:
I might be missing some context/motivation, but I was wondering: would this not function effectively the same as "create_table()` with an in-memory `obj`?

gforsyth:
We discussed this today briefly and I think @chloeh13q is right -- this ends up being (mostly) a specialized version of passing an in-memory object to `create_table`.

Currently, if the `obj` isn't an Ibis expression, we convert it to a `memtable` and then create the table from that.

I think that's a reasonable default/fallback behavior for most backends and allows for users to insert pyarrow tables, polars dataframes and pandas dataframes into all backends.  This is also currently implemented.

For DuckDB, Datafusion, and Polars, there are extra options available.

For Datafusion and Polars, there are native methods to read in pyarrow tables, pyarrow recordbatchreaders, polars dataframes, pandas dataframes, etc. Datafusion creates all of these objects as "Tables", Polars does not distinguish between tables and views, so they are also created as what Ibis refers to as "tables".  

Adding extra handling for the fast-path for querying these objects will allow us to deprecate `register` without also introducing a new `read_in_memory` to these backends.  We can also deprecate some of the existing format-specific methods in those backends (e.g. polars `read_pandas`).

[...]

PR #8863, opened 2024-04-02
refactor: deprecate register api

ncclementi:
Looks like I messed something up in the rebase, I'll fix it soon. 
 
- Closes https://github.com/ibis-project/ibis/issues/8735

### DIFF
--- a/ibis/backends/datafusion/tests/test_register.py
+++ b/ibis/backends/datafusion/tests/test_register.py
@@ -25,24 +25,28 @@ def test_read_parquet(conn, data_dir):
 
 def test_register_table(conn):
     tab = pa.table({"x": [1, 2, 3]})
-    conn.register(tab, "my_table")
+    with pytest.warns(FutureWarning, match="v9.0"):
+        conn.register(tab, "my_table")
     assert conn.table("my_table").x.sum().execute() == 6
 
 
 def test_register_pandas(conn):
     df = pd.DataFrame({"x": [1, 2, 3]})
-    conn.register(df, "my_table")
-    assert conn.table("my_table").x.sum().execute() == 6
+    with pytest.warns(FutureWarning, match="v9.0"):
+        conn.register(df, "my_table")
+        assert conn.table("my_table").x.sum().execute() == 6
 
 
 def test_register_batches(conn):
     batch = pa.record_batch([pa.array([1, 2, 3])], names=["x"])
-    conn.register(batch, "my_table")
-    assert conn.table("my_table").x.sum().execute() == 6
+    with pytest.warns(FutureWarning, match="v9.0"):
+        conn.register(batch, "my_table")
+        assert conn.table("my_table").x.sum().execute() == 6
 
 
 def test_register_dataset(conn):
     tab = pa.table({"x": [1, 2, 3]})
     dataset = ds.InMemoryDataset(tab)
-    conn.register(dataset, "my_table")
-    assert conn.table("my_table").x.sum().execute() == 6
+    with pytest.warns(FutureWarning, match="v9.0"):
+        conn.register(dataset, "my_table")
+        assert conn.table("my_table").x.sum().execute() == 6
--- a/ibis/backends/datafusion/tests/test_register.py
+++ b/ibis/backends/datafusion/tests/test_register.py
@@ -25,21 +25,21 @@ def test_read_parquet(conn, data_dir):
 
 def test_register_table(conn):
     tab = pa.table({"x": [1, 2, 3]})
-    with pytest.warns(FutureWarning, match="v9.0"):
+    with pytest.warns(FutureWarning, match="v9.1"):
         conn.register(tab, "my_table")
     assert conn.table("my_table").x.sum().execute() == 6
 
 
 def test_register_pandas(conn):
     df = pd.DataFrame({"x": [1, 2, 3]})
-    with pytest.warns(FutureWarning, match="v9.0"):
+    with pytest.warns(FutureWarning, match="v9.1"):
         conn.register(df, "my_table")
         assert conn.table("my_table").x.sum().execute() == 6
 
 
 def test_register_batches(conn):
     batch = pa.record_batch([pa.array([1, 2, 3])], names=["x"])
-    with pytest.warns(FutureWarning, match="v9.0"):
+    with pytest.warns(FutureWarning, match="v9.1"):
         conn.register(batch, "my_table")
         assert conn.table("my_table").x.sum().execute() == 6
 
@@ -47,6 +47,6 @@ def test_register_batches(conn):
 def test_register_dataset(conn):
     tab = pa.table({"x": [1, 2, 3]})
     dataset = ds.InMemoryDataset(tab)
-    with pytest.warns(FutureWarning, match="v9.0"):
+    with pytest.warns(FutureWarning, match="v9.1"):
         conn.register(dataset, "my_table")
         assert conn.table("my_table").x.sum().execute() == 6
--- a/ibis/backends/duckdb/tests/test_client.py
+++ b/ibis/backends/duckdb/tests/test_client.py
@@ -268,7 +268,9 @@ def test_connect_duckdb(url, tmp_path):
 )
 def test_connect_local_file(out_method, extension, test_employee_data_1, tmp_path):
     getattr(test_employee_data_1, out_method)(tmp_path / f"out.{extension}")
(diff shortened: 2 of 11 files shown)


ibis/backends/datafusion/tests/test_register.py
```diff
@@ -25,24 +25,28 @@ def test_read_parquet(conn, data_dir):
 
 def test_register_table(conn):
     tab = pa.table({"x": [1, 2, 3]})
-    conn.register(tab, "my_table")
-    assert conn.table("my_table").x.sum().execute() == 6
+    with pytest.warns(FutureWarning, match="v9.0"):
+        conn.register(tab, "my_table")
+        assert conn.table("my_table").x.sum().execute() == 6
```
gforsyth, line 30:
```suggestion
    with pytest.warns(FutureWarning, match="v9.0"):
        conn.register(tab, "my_table")
    assert conn.table("my_table").x.sum().execute() == 6
```

We should only have the line (or lines) that warns in the context manager block

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -268,9 +268,11 @@ def test_connect_duckdb(url, tmp_path):
...
 def test_connect_local_file(out_method, extension, test_employee_data_1, tmp_path):
     getattr(test_employee_data_1, out_method)(tmp_path / f"out.{extension}")
-    con = ibis.connect(tmp_path / f"out.{extension}")
-    t = next(iter(con.tables.values()))
-    assert not t.head().execute().empty
+    with pytest.warns(FutureWarning, match="v9.0"):
+        # ibis.connect uses con.register
+        con = ibis.connect(tmp_path / f"out.{extension}")
+        t = next(iter(con.tables.values()))
+        assert not t.head().execute().empty
```
gforsyth, line 275:
```suggestion
        con = ibis.connect(tmp_path / f"out.{extension}")
    t = next(iter(con.tables.values()))
    assert not t.head().execute().empty
```

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -256,12 +256,13 @@ def test_read_sqlite_no_table_name(con, tmp_path):
...
+        path = tmp_path / "test.db"
 
-    sqlite_con = sqlite3.connect(str(path))
-    sqlite_con.execute("CREATE TABLE t AS SELECT 1 a UNION SELECT 2 UNION SELECT 3")
-    ft = con.register(f"sqlite://{path}", "t")
-    assert ft.count().execute()
+        sqlite_con = sqlite3.connect(str(path))
+        sqlite_con.execute("CREATE TABLE t AS SELECT 1 a UNION SELECT 2 UNION SELECT 3")
+        ft = con.register(f"sqlite://{path}", "t")
+        assert ft.count().execute()
```
gforsyth, line 265: `warns` should only be around the `con.register` call

ibis/backends/tests/test_register.py
```diff
@@ -99,7 +99,8 @@ def gzip_csv(data_dir, tmp_path):
 )
 def test_register_csv(con, data_dir, fname, in_table_name, out_table_name):
     with pushd(data_dir / "csv"):
-        table = con.register(fname, table_name=in_table_name)
+        with pytest.warns(FutureWarning, match="v9.0"):
+            table = con.register(fname, table_name=in_table_name)
 
     assert any(out_table_name in t for t in con.list_tables())
     if con.name != "datafusion":
```
gforsyth, line 106: I think this test should be rewritten to test `read_csv` (and we can probably rename the test file, too)

ibis/backends/tests/test_register.py
```diff
@@ -154,7 +156,8 @@ def test_register_with_dotted_name(con, data_dir, tmp_path):
     f.parent.mkdir()
     data = data_dir.joinpath("csv", "diamonds.csv").read_bytes()
     f.write_bytes(data)
-    table = con.register(str(f.absolute()))
+    with pytest.warns(FutureWarning, match="v9.0"):
+        table = con.register(str(f.absolute()))
```
gforsyth, line 160: same, this is now a `read_csv` test

ibis/backends/tests/test_register.py
```diff
@@ -216,9 +219,10 @@ def test_register_parquet(
     pq.write_table(table, tmp_path / fname.name)
 
     with pushd(tmp_path):
-        table = con.register(f"parquet://{fname.name}", table_name=in_table_name)
+        with pytest.warns(FutureWarning, match="v9.0"):
+            table = con.register(f"parquet://{fname.name}", table_name=in_table_name)
```
gforsyth, line 223: this should now be a read_parquet test

ibis/backends/tests/test_register.py
```diff
@@ -255,10 +259,14 @@ def test_register_iterator_parquet(
...
-            table_name=None,
-        )
+        with pytest.warns(FutureWarning, match="v9.0"):
+            table = con.register(
+                [
+                    "parquet://functional_alltypes.parquet",
+                    "functional_alltypes.parquet",
+                ],
+                table_name=None,
+            )
```
gforsyth, line 269: this can be testing reading an iterable of parquet files with `read_parquet`

ibis/backends/tests/test_register.py
```diff
@@ -287,10 +295,12 @@ def test_register_pandas(con):
     pd = pytest.importorskip("pandas")
     df = pd.DataFrame({"x": [1, 2, 3], "y": ["a", "b", "c"]})
 
-    t = con.register(df)
+    with pytest.warns(FutureWarning, match="v9.0"):
+        t = con.register(df)
```
gforsyth, line 299: The in-memory tests we can probably leave alone for now, but we should add a comment that these should be updated  to work with the upcoming `read_in_memory` and we should also add a polars dataframe to the mix

ibis/backends/tests/test_register.py
```diff
@@ -351,8 +362,9 @@ def test_csv_reregister_schema(con, tmp_path):
             ]
         )
 
-    # For a full file scan, expect correct schema based on final row
-    foo_table = con.register(foo, table_name="same")
+    with pytest.warns(FutureWarning, match="v9.0"):
+        # For a full file scan, expect correct schema based on final row
+        foo_table = con.register(foo, table_name="same")
```
gforsyth, line 367: this can be `read_csv`

gforsyth (changes requested):
Sorry for the delay @ncclementi !

The deprecation implementation is good -- I think some of the tests can be updated now and some should probably wait until we have a standardized `read_in_memory`

ibis/backends/tests/test_register.py
ncclementi: @gforsyth I noticed that ion the same file we do have already `test_read_csv` and `test_read_csv_glob` similarly for parquet. Do we still want to convert these tests then? what are we testing that is different in this case?

ncclementi:
I modify the tests in question to use `read_csv` or `read_parquet`  I wasn't sure what the name of the test should be or if we need them given that there are separate `read_csv` and `read_parquet`, though the register tests seem to be more complete.  Happy to re-visit.

ncclementi:
**Update regarding adapting `test_register_csv` and similar to use `con.read_csv`**

Turns out that doing this would require to modify the tests quite a bit given that now this will trigger some XPASS failures, and the tests will become way more convoluted than what they are at the moment. 

After chatting with @cpcloud, the suggestion was to leave the tests as they are now, and discuss a decision for when we remove the `register` api completely.

gforsyth:
Ok, I'm good with the changes here, but I think we need a consensus from a few folks on whether to merge now or not.
If we want to add this in Ibis 9.0, I think we also need to handle standardizing `read_in_memory` xref #6593 before we release, because it seems irresponsible to deprecate functionality without offering a replacement.

cpcloud:
@gforsyth Brings up a good point. Let's hold off on merging to 9.0 for now to avoid delaying its release. We can merge this soon after 9.0 is cut (not immediately, just to wait for the bug reports to come in and cut a 9.0.1 or 9.1 bug fix release).

ncclementi:
I just rebased, I think when CI finishes we can get this one in since 9.0 is out

ncclementi:
@gforsyth should we get this one in, or are there still concerns merging this deprecation without having an alternative `read_in_memory` api?

ncclementi:
Ready, waiting for https://github.com/ibis-project/ibis/pull/9251 to be merged.

cpcloud (approved):

Issue #8382, opened 2024-02-16
feat: Ability to Distinguish between Table and Views


sarsasid:
### Is your feature request related to a problem?

I see that we have an option to find Tables using 

`con.list_tables()` method, but in many DBs Views are also considered as tables, as in INFORMATION_SCHEMA most DB's list Views as well

If there is a way to add a param to backend to remove Views that would be great

### Describe the solution you'd like

Add a default param `table_types=[]` to `con.list_tables()` Function

> Suggesting as table_types, as some DB will have different other types as well and will be flexible for users to specify table type

In all DB's logic add a condition, `f table_types;`

- SQL Server, add filter table_type IN ( ','.join(table_type) )
- Postgres, add filter table_type IN ( ','.join(table_type) )
- BigQuery, add filter table_type IN ( ','.join(table_type) )
etc

### What version of ibis are you running?

8.0.0

### What backend(s) are you using, if any?

SQL Server

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gforsyth:
This is an exceptionally hairy problem for something that seems so simple.

## Current API

The goal of the existing API is to treat everything tabular as an abstract table, so tables, views, materialized views, etc, are all "tables" in this context.  So you `list_tables()`, you get all the abstract tables in your particular `catalog.database` namespace, so tables, views, materialized views, temp tables, etc.

## If we separate into `list_views` and `list_tables`

First, we break a bunch of stuff, but more importantly, I think this is bad for usability.

Consider the following if we have separate `list_views` and `list_tables`:

```python
>>> import ibis

>>> con = ibis.duckdb.connect()
>>> con.read_parquet("some_file.parquet")
>>> con.list_tables()
[]
```

For a new user, especially, I think this would be suboptimal.  We're now requiring them to understand implementation details specific to the backend they are using.  You need to know that DuckDB creates a view so that it can do lazy-loading and pushdowns into the parquet file.  Depending on the backend, it might also get loaded in as a temporary table, so now you have to know which engine you are on so you know which command to run to check whether the parquet file has been loaded in as expected.

## Propoal: `ddl` accessor

Conceptually, the distinction between and `view` and a `table` is entirely at the DDL-level, so let's treat it that way.

We leave `list_tables` alone, it still returns a list of tabular things.

[...]

PR #8864, opened 2024-04-02
feat(ddl): distinguish views from tables while listing them

mfatihaktas:
## Description of changes

First step towards addressing https://github.com/ibis-project/ibis/issues/8382.

This
- Adds `no_views` arg to `list_tables()` for clickhouse, Flink and pyspark backends.

Asking for early feedback. If the change is reasonable, will proceed to make the same change for the remaining backends.

**Edit**: After the discussion on this PR:
* Made `list_tables()` and `list_views()` explicit.
  * For some of the backends, I could not find a direct way to distinguish views from the tables. Added `TODO` comments to decide for those during the reviews.
* Added `list()` to list tables and views together.

### DIFF
--- a/ibis/backends/clickhouse/__init__.py
+++ b/ibis/backends/clickhouse/__init__.py
@@ -197,19 +197,13 @@ class Backend(SQLBackend, CanCreateDatabase):
             databases = []
         return self._filter_with_like(databases, like)
 
-    def list_tables(
-        self, like: str | None = None, database: str | None = None
+    def _list_tables_or_views(
+        self,
+        views: bool,
+        like: str | None = None,
+        database: str | None = None,
     ) -> list[str]:
-        """List the tables in the database.
-
-        Parameters
-        ----------
-        like
-            A pattern to use for listing tables.
-        database
-            Database to list tables from. Default behavior is to show tables in
-            the current database.
-        """
+        """Helper function for `list_tables/views()`."""
 
         query = sg.select(C.name).from_(sg.table("tables", db="system"))
 
@@ -218,7 +212,8 @@ class Backend(SQLBackend, CanCreateDatabase):
         else:
             database = sge.convert(database)
 
-        query = query.where(C.database.eq(database).or_(C.is_temporary))
+        view_cond = C.engine.eq("View") if views else C.engine.neq("View")
+        query = query.where(C.database.eq(database).or_(C.is_temporary), view_cond)
 
         with self._safe_raw_sql(query) as result:
             results = result.result_columns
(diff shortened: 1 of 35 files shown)


ibis/backends/clickhouse/__init__.py
```diff
@@ -197,17 +197,23 @@ def list_databases(self, like: str | None = None) -> list[str]:
         return self._filter_with_like(databases, like)
 
     def list_tables(
-        self, like: str | None = None, database: str | None = None
+        self,
+        like: str | None = None,
+        database: str | None = None,
+        no_views: bool = False,
```
gforsyth, line 203: generally in this case I prefer the following pattern:
```suggestion
        include_views: bool = True,
```

mfatihaktas: Thanks, makes sense, I will switch to `include_views`.

mfatihaktas:
> Instead, how about we have a list_foo method for each kind of object, and introduce a new method, list_rels (or list_relations) listing the union of all tabular objects. 
> We could also not implement the latter until someone asks for it.

Thanks @cpcloud , I also think that exposing `list_foo()` methods that do exactly what their names suggest would be cleaner and easier to maintain. Just to note, this will be a breaking change as `list_tables()` list views together with the tables.

I will proceed with this approach then.

mfatihaktas:
> The one that immediately comes to mind is
> user does a read_csv with DuckDB, even passes in a table_name kwarg
> user calls list_tables and there is nothing there because read_csv creates a view

As another example to this, when `create_table()` is called with an in-memory object, it creates a view rather than a table. Flink requires a source connector for tables and we chose to create a view instead of the alternative where we would first write the obj into a temporary file and then create a table pointing to the file.

cpcloud:
> > The one that immediately comes to mind is
> > user does a read_csv with DuckDB, even passes in a table_name kwarg
> > user calls list_tables and there is nothing there because read_csv creates a view
> 
> As another example to this, when `create_table()` is called with an in-memory object, it creates a view rather than a table. Flink requires a source connector for tables and we chose to create a view instead of the alternative where we would first write the obj into a temporary file and then create a table pointing to the file.

This isn't true in general, just for the Flink backend. All other backends will create physical persistent storage when `create_table` is called.

cpcloud:
> I'm not opposed to explicit `list_` methods but I think we need to consider the user with less SQL experience and think about how we want to handle certain scenarios.
> 
> The one that immediately comes to mind is
> 
> * user does a `read_csv` with DuckDB, even passes in a `table_name` kwarg
> * user calls `list_tables` and there is nothing there because `read_csv` creates a view

IMO this isn't a huge loss, but would require some documentation in `list_tables` and in any of the `read_foo` methods that create views.

Another option would be to materialize the table, but I don't know if we want to do that. Repeated use of large CSVs will be faster, but the initial function call will scale with the CSV file rather than the number of sample lines like it does now.

mfatihaktas:
As a side question: 

`BasePandasBackend`, which is used for `pandas` and `dask` backends,  has `create_view()` defined as
```python
def create_view(
        self,
        name: str,
        obj: ir.Table,
        *,
        database: str | None = None,
        overwrite: bool = False,
    ) -> ir.Table:
        return self.create_table(
            name, obj=obj, temp=None, database=database, overwrite=overwrite
        )
```
Given that the notion of `virtual table (view)` is non-existent for these backends, should we remove `create_view()` to make the API explicit as part of this effort/issue?

mfatihaktas:
Thank you all for the discussion. The first pass is ready for review now:
- Made `list_tables()` and `list_views()` explicit.
  - For some of the backends, I could not find a direct way to distinguish views from the tables. Added `TODO` comments to decide for those during the reviews.
- Added `list()` for listing both tables and views.

ncclementi:
@mfatihaktas What's the status of this one, is this blocked or does it need a review? 
If you need a review would you mind solving the conflicts first and then we can ping someone

mfatihaktas:
> @mfatihaktas What's the status of this one, is this blocked or does it need a review? If you need a review would you mind solving the conflicts first and then we can ping someone

Thanks for checking on this. Regarding the status, I marked this PR as ready-for-review some time ago. I am busy with another project this week, but will try to resolve the conflicts before EOD on Friday.

jcrist:
Closing in favor of the approach described here: https://github.com/ibis-project/ibis/issues/8382#issuecomment-2237214114

PR #8875, opened 2024-04-03
docs(release-notes): add v to start of anchor tags (#8624)

### LINKED ISSUE #8624: docs: Make links to sections in release notes stable across releases
### Please describe the issue

I noticed while grabbing a hyperlink to the latest Ibis release from https://ibis-project.org/release_notes that the fragment identifier for the most recent release is ambiguous: https://ibis-project.org/release_notes#section. I then noticed that all the anchor links are similar. Quarto (or pandoc?) appears to not like when headings start with numbers and falls back to some sort of other behavior. The result is that, on every new Ibis release https://ibis-project.org/release_notes#section will jump the user to the latest release and it won't be possible to have a stable link to a particular release section.

While Ibis' users may not expect these links to be stable, I think it's really nice to just have stable anchor hyperlinks.

I'd be happy to send in a PR for a change if this is of interest. I think just changing the headers is the way to go, rather than messing with any custom hacks with Pandoc or similar. Perhaps just prepending "v" or "Ibis " to each heading? While I'm in there I could take a look around the rest of the site to find any other issues.

### Code of Conduct

lee-coates:
## Description of changes
adds a v to the start of each version's anchor tag to allow the table of contents to have non-ambiguous fragment identifiers in the links, as the relevant issue suggests.

## Issues closed

* Resolves #8624


lee-coates:
will resubmit on dev branch

cpcloud:
I'm going to try to do this with a quarto filter so that we don't have to remember to update this on every release. PR incoming.

PR #8876, opened 2024-04-03
fix: disallow empty struct

### LINKED ISSUE #8460: bug: error when handling empty / undefined struct
### What happened?

We have a user on Zulip who was trying to make a memtable out of something like the following (representative of an upstream data source):

```python
ibis.memtable(
[
    {
        'year': '2024',
        'period': 'M01',
        'periodName': 'January',
        'latest': 'true',
        'value': '3.7',
        'footnotes': {}
    }
]
)
```

`footnotes` is getting parsed as an empty / undefined struct and is causing an error when we attempt to retrieve the value of the field.  

```python
File ~/github.com/ibis-project/ibis/ibis/expr/types/core.py:424, in Expr.to_pyarrow(self, params, limit, **kwargs)
    396 @experimental
    397 def to_pyarrow(
    398     self,
   (...)
    402     **kwargs: Any,
    403 ) -> pa.Table:
    404     """Execute expression and return results in as a pyarrow table.
    405 
    406     This method is eager and will execute the associated expression
   (...)
    422         A pyarrow table holding the results of the executed expression.
    423     """
--> 424     return self._find_backend(use_default=True).to_pyarrow(
    425         self, params=params, limit=limit, **kwargs
    426     )

File ~/github.com/ibis-project/ibis/ibis/backends/duckdb/__init__.py:1243, in Backend.to_pyarrow(self, expr, params, limit, **_)
   1240 sql = self.compile(table, limit=limit, params=params)
   1242 with self._safe_raw_sql(sql) as cur:
-> 1243     table = cur.fetch_arrow_table()
   1245 return expr.__pyarrow_result__(table)

InternalException: INTERNAL Error: Attempted to access index 0 within vector of size 0
```

Could we try to adjust the schema in this case to something that won't barf when it's empty?

main

DuckDB

### Relevant log output

_No response_

### Code of Conduct

chloeh13q:
## Description of changes

Disallow empty struct in type inference. Pyarrow allows the construction of empty structs but duckdb doesn't.

I think what the user in #8460 was attempting to create was an empty value in a struct column that has defined fields. You can accomplish this with the following:

```python
>>> import ibis
>>>
>>>
>>> t = ibis.memtable(
...             [
...                 {
...                     "year": "2024",
...                     "period": "M01",
...                     "periodName": "January",
...                     "latest": "true",
...                     "value": "3.7",
...                     "footnotes": None,
...                 }
...             ],
...             schema={"year": "string", "period": "string", "periodName": "string", "latest": "string", "value": "string", "footnotes": "struct<key: string>"}
...         )
>>> t
┏━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━┓
┃ year   ┃ period ┃ periodName ┃ latest ┃ value  ┃ footnotes ┃
┡━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━┩
│ string │ string │ string     │ string │ string │ struct<k… │
├────────┼────────┼────────────┼────────┼────────┼───────────┤
│ 2024   │ M01    │ January    │ true   │ 3.7    │ NULL      │
└────────┴────────┴────────────┴────────┴────────┴───────────┘
```

I do want to note here that ibis interprets `{}` as a struct whose keys point to empty values rather than a struct that is empty:

```python
>>> t = ibis.memtable(
...             [
...                 {
...                     "year": "2024",
...                     "period": "M01",
...                     "periodName": "January",
...                     "latest": "true",
...                     "value": "3.7",
...                     "footnotes": {},
...                 }
...             ],
...             schema={"year": "string", "period": "string", "periodName": "string", "latest": "string", "value": "string", "footnotes": "struct<key: string>"}
...         )
>>> t
┏━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┓
┃ year   ┃ period ┃ periodName ┃ latest ┃ value  ┃ footnotes            ┃
┡━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━┩
│ string │ string │ string     │ string │ string │ struct<key: string>  │
├────────┼────────┼────────────┼────────┼────────┼──────────────────────┤
│ 2024   │ M01    │ January    │ true   │ 3.7    │ {'key': None}        │
└────────┴────────┴────────────┴────────┴────────┴──────────────────────┘
```

## Issues closed

#8460

### DIFF
--- a/ibis/backends/dask/__init__.py
+++ b/ibis/backends/dask/__init__.py
@@ -170,7 +170,9 @@ class Backend(BasePandasBackend, NoUrl):
             schema = self.schemas[table_name]
         except KeyError:
             df = self.dictionary[table_name]
-            self.schemas[table_name] = schema = PandasData.infer_table(df.head(1))
+            if isinstance(df, dd.DataFrame):
+                df = df.compute()
+            self.schemas[table_name] = schema = PandasData.infer_table(df)
 
         return schema
 
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -828,6 +828,11 @@ class Struct(Parametric, MapSet):
     scalar = "StructScalar"
     column = "StructColumn"
 
+    def __init__(self, fields, **kwargs):
+        if fields is None or len(fields) == 0:
+            raise ValueError("Empty struct type is not supported")
+        super().__init__(fields=fields, **kwargs)
+
     @classmethod
     def from_tuples(
         cls, pairs: Iterable[tuple[str, str | DataType]], nullable: bool = True
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -830,7 +830,7 @@ class Struct(Parametric, MapSet):
 
     def __init__(self, fields, **kwargs):
         if fields is None or len(fields) == 0:
-            raise ValueError("Empty struct type is not supported")
+            raise TypeError("Empty struct type is not supported")
         super().__init__(fields=fields, **kwargs)
 
     @classmethod
--- a/ibis/expr/datatypes/value.py
+++ b/ibis/expr/datatypes/value.py
@@ -42,8 +42,6 @@ def infer(value: Any) -> dt.DataType:
 @infer.register(collections.OrderedDict)
 def infer_struct(value: Mapping[str, Any]) -> dt.Struct:
     """Infer the [`Struct`](./datatypes.qmd#ibis.expr.datatypes.Struct) type of `value`."""
-    if not value:
-        raise TypeError("Empty struct type not supported")
     fields = {name: infer(val) for name, val in value.items()}
     return dt.Struct(fields)
 
--- a/ibis/formats/pyarrow.py
+++ b/ibis/formats/pyarrow.py
@@ -90,8 +90,6 @@ class PyArrowType(TypeMapper):
             value_dtype = cls.to_ibis(typ.value_type, typ.value_field.nullable)
             return dt.Array(value_dtype, nullable=nullable)
         elif pa.types.is_struct(typ):
-            if typ.num_fields == 0:
-                raise ValueError("Empty struct type is not supported")
             field_dtypes = {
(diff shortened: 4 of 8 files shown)


NickCrews:
It seems to me like `{}` does not properly follow the schema of "struct<key: string>", and so `ibis.literal({}, type="struct<key: string>")` should just fail. The user's input is malformed, and they will need to deal with it external to ibis. Or, perhaps they could use a `map<string, string>` type instead, which properly represents a datatype with a variable number of keys.

IDK, maybe we could make ibis lenient, but it feels like a slippery slope. If we allow this, then it feels like we should make it so that `ibis.struct({"a": "foo"}).cast("struct<a: string, b:int>")` to succeed, and result in `{"a": "foo", "b": None}` (ie if you cast something to a struct, anytime a field isn't present, fill it in with NULL). This feels OK to me, but I don't think backends have good builtin support for this, we would have to add workarounds all over the place.

ibis/backends/dask/__init__.py
```diff
@@ -170,7 +170,9 @@ def get_schema(self, table_name, *, database=None):
             schema = self.schemas[table_name]
         except KeyError:
             df = self.dictionary[table_name]
-            self.schemas[table_name] = schema = PandasData.infer_table(df.head(1))
+            if isinstance(df, dd.DataFrame):
+                df = df.compute()
+            self.schemas[table_name] = schema = PandasData.infer_table(df)
```
chloeh13q, line 175: The original code was breaking a bunch of tests because in one of the datasets we're using there is a struct column that looks like `[{}, None, {1: "a", 2: "b"}]`. We need all of the data in that column in order to infer the struct type there...

PR #8878, opened 2024-04-03
refactor(api): make input value coercion of `mutate()` identical to `select()`

kszucs:
String and integer literals passed to select() are interpreted as columns whereas mutate() interpreted them as literals.

BREAKING CHANGE: strings passed to table.mutate() are now interpreted as column references instead of literals, use `ibis.literal(string)` to pass the string as a literal


kszucs:
I am actually not a big fan of `t.select(42)` trying to locate a column. Treating strings as column references is fine but integers are rather weird.

cpcloud (approved):
Makes sense. Still a good chunk of failures though.

cpcloud:
+1 to removing the string-and-int-as-column references in mutate.

We have `rename` for the use cases this covers.

kszucs:
Now both are consistent. My preference is to remove both, but I am afraid that this would be too big of a breaking change. Though we actually break it now, and provide a config option to fall back to the previous behavior.

jcrist:
> +1 to removing the string-and-int-as-column references in mutate and select (I can't remember which one has it and which one doesn't)

IMO we'll still want to support strings as column references. `t.select("x", "y", "z")` is such a common pattern, and feels intuitive to me (while `t.select(0, 1, 2)` doesn't).

kszucs:
Ok, I think we have an agreement on removing the `int` case. Going to put up a follow-up for that.

cpcloud:
We need to document this as part of this PR or the next one, because it seems pretty non trivial to articulate what the rules are especially around strings.

What exactly is the behavior of each of `str` and `int` **without** calling `literal` in each of `select` and `mutate`?

kszucs:
Created a follow-up issue for the documentation https://github.com/ibis-project/ibis/issues/8879

PR #8883, opened 2024-04-03
test(benchmarks): fix constructing large benchmark expression

kszucs:
Fixing build error: https://github.com/ibis-project/ibis/actions/runs/8541350879/job/23400454706


cpcloud:
We should try to figure out why this isn't failing in PRs, but we can do that in a follow-up.

kszucs (commented):
+1

PR #8884, opened 2024-04-03
refactor(api): treat integer inputs as literals instead of column references

kszucs:
Removing the int-as-column references in mutate and select. Further discussed in #8878.

BREAKING CHANGE: Integer inputs to `select` and `mutate` are now always interpreted as literals. Columns can still be accessed by their integer index using square-bracket syntax.


cpcloud (approved):
LGTM! Thank you!

PR #8901, opened 2024-04-05
feat: support UUIDs to pyarrow on more backends

### LINKED ISSUE #8902: bug: `postgres.to_pyarrow(ibis.uuid())` errors
### What happened?

See the xfailing test in https://github.com/ibis-project/ibis/pull/8901

```
ibis/backends/__init__.py:221: in to_pyarrow
    table = pa.Table.from_batches(reader, schema=arrow_schema)
pyarrow/table.pxi:4104: in pyarrow.lib.Table.from_batches
    ???
pyarrow/ipc.pxi:666: in pyarrow.lib.RecordBatchReader.__next__
    ???
pyarrow/ipc.pxi:700: in pyarrow.lib.RecordBatchReader.read_next_batch
    ???
pyarrow/error.pxi:88: in pyarrow.lib.check_status
    ???
ibis/backends/sql/__init__.py:375: in <genexpr>
    pa.array(map(tuple, batch), type=array_type)
pyarrow/array.pxi:343: in pyarrow.lib.array
    ???
pyarrow/array.pxi:42: in pyarrow.lib._sequence_to_array
    ???
pyarrow/error.pxi:154: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pyarrow.lib.ArrowTypeError: Expected bytes, got a 'UUID' object

pyarrow/error.pxi:91: ArrowTypeError
```

in ibis/formats/pyarrow.py, we say that ibis.UUID should map to pa.string. This makes sense, there is no builting UUID in pyarrow.

but then in SQLBackend.to_pyarrow_batches(), we do
```python
schema = expr.as_table().schema()
        array_type = schema.as_struct().to_pyarrow()
        arrays = (
            pa.array(map(tuple, batch), type=array_type)
            for batch in self._cursor_batches(
                expr, params=params, limit=limit, chunk_size=chunk_size
            )
        )
```

here, `batch` is tuples of the python standard lib `uuid.UUID` from the psyopg2 cursor. Then, the `pa.array(<uuid.UUIDs> type=pa.string)` call fails.

I see a few possible paths here:
- do the conversion on the backend side, before fetching results. Then the psyopg2 cursor returns actual strings. We already have the ibis->pyarrow type mappings, we could then roundtrip the other way first, eg `expr.cast(type.to_arrow().to_ibis())` Probably the easiest, but might lead to the wrong types sometimes if some info is lost during this roundtrip?
- do the conversion after fetching from the cursor. maybe hardcode the few datatypes that are problems?
- submit a patch to upstream pyarrow to accept uuid.UUID types

main

postgres

### Relevant log output

_No response_

### Code of Conduct

NickCrews:
partially fixes #8902.

Implements UUID execution to pyarrow on some backends, and adds notimpl tests for the rest.

### DIFF
--- a/ibis/backends/impala/__init__.py
+++ b/ibis/backends/impala/__init__.py
@@ -1144,7 +1144,6 @@ class Backend(SQLBackend):
         limit: int | str | None = None,
         **kwargs: Any,
     ) -> pa.Table:
-        import pyarrow as pa
         import pyarrow_hotfix  # noqa: F401
 
         from ibis.formats.pyarrow import PyArrowData
@@ -1152,12 +1151,9 @@ class Backend(SQLBackend):
         self._run_pre_execute_hooks(expr)
 
         table_expr = expr.as_table()
-        output = pa.Table.from_pandas(
-            self.execute(table_expr, params=params, limit=limit, **kwargs),
-            preserve_index=False,
-        )
-        table = PyArrowData.convert_table(output, table_expr.schema())
-        return expr.__pyarrow_result__(table)
+        df = self.execute(table_expr, params=params, limit=limit, **kwargs)
+        pa_table = PyArrowData.convert_table(df, table_expr.schema())
+        return expr.__pyarrow_result__(pa_table)
 
     def to_pyarrow_batches(
         self,
--- a/ibis/formats/pyarrow.py
+++ b/ibis/formats/pyarrow.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
 
+import pandas as pd
 import pyarrow as pa
 import pyarrow_hotfix  # noqa: F401
 
@@ -243,7 +244,14 @@ class PyArrowData(DataMapper):
             return scalar
 
     @classmethod
-    def convert_column(cls, column: pa.Array, dtype: dt.DataType) -> pa.Array:
+    def convert_column(
+        cls, column: pa.Array | pd.Series, dtype: dt.DataType
+    ) -> pa.Array:
+        if isinstance(column, pd.Series):
+            if dtype.is_uuid():
+                # pyarrow doesn't support UUIDs, so we need to convert them to strings
+                column = column.astype(str)
+            column = pa.Array.from_pandas(column)
         desired_type = PyArrowType.from_ibis(dtype)
         if column.type != desired_type:
             return column.cast(desired_type)
@@ -251,7 +259,15 @@ class PyArrowData(DataMapper):
             return column
 
     @classmethod
-    def convert_table(cls, table: pa.Table, schema: Schema) -> pa.Table:
+    def convert_table(cls, table: pa.Table | pd.DataFrame, schema: Schema) -> pa.Table:
+        if isinstance(table, pd.DataFrame):
+            table = pa.Table.from_arrays(
+                [
+                    cls.convert_column(table[col], dtype)
+                    for col, dtype in schema.items()
+                ],
+                names=schema.names,
+            )
         desired_schema = PyArrowSchema.from_ibis(schema)
         pa_schema = table.schema
 
--- a/ibis/backends/flink/__init__.py
+++ b/ibis/backends/flink/__init__.py
@@ -24,6 +24,7 @@ from ibis.backends.flink.ddl import (
 from ibis.backends.sql import SQLBackend
 from ibis.backends.tests.errors import Py4JJavaError
 from ibis.expr.operations.udf import InputType
+from ibis.formats.pyarrow import PyArrowData
 from ibis.util import gen_name
 
 if TYPE_CHECKING:
@@ -940,7 +941,6 @@ class Backend(SQLBackend, CanCreateDatabase, NoUrl):
         limit: int | str | None = None,
         **kwargs: Any,
     ):
-        import pyarrow as pa
         import pyarrow_hotfix  # noqa: F401
 
         ibis_table = expr.as_table()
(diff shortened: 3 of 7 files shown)


ibis/formats/pyarrow.py
```diff
@@ -243,15 +244,30 @@ def convert_scalar(cls, scalar: pa.Scalar, dtype: dt.DataType) -> pa.Scalar:
             return scalar
 
     @classmethod
-    def convert_column(cls, column: pa.Array, dtype: dt.DataType) -> pa.Array:
+    def convert_column(
+        cls, column: pa.Array | pd.Series, dtype: dt.DataType
+    ) -> pa.Array:
+        if isinstance(column, pd.Series):
+            if dtype.is_uuid():
```
NickCrews, line 251: I don't love this special casing here. we could do doing something like 
`target_pd_type = PandasType.from_ibis(PyarrowType.to_ibis(PyarrowType.from_ibis(dtype)))`
which would eval to

```
PandasType.from_ibis(PyarrowType.to_ibis(PyarrowType.from_ibis(dt.UUID)))
PandasType.from_ibis(PyarrowType.to_ibis(pa.string))
PandasType.from_ibis(dt.String)
pd.StringType
```
but IDK that is a bit of a brainf$#$%k as well.

IDK

NickCrews: sounds good lets keep it as is.

ibis/backends/impala/__init__.py
```diff
@@ -1144,20 +1144,16 @@ def to_pyarrow(
...
         self._run_pre_execute_hooks(expr)
 
         table_expr = expr.as_table()
-        output = pa.Table.from_pandas(
-            self.execute(table_expr, params=params, limit=limit, **kwargs),
-            preserve_index=False,
-        )
-        table = PyArrowData.convert_table(output, table_expr.schema())
-        return expr.__pyarrow_result__(table)
+        df = self.execute(table_expr, params=params, limit=limit, **kwargs)
```
cpcloud, line 1154: Nice, once less conversion layer!

ibis/formats/pyarrow.py
```diff
@@ -243,15 +244,30 @@ def convert_scalar(cls, scalar: pa.Scalar, dtype: dt.DataType) -> pa.Scalar:
...
         desired_type = PyArrowType.from_ibis(dtype)
         if column.type != desired_type:
             return column.cast(desired_type)
         else:
             return column
 
     @classmethod
-    def convert_table(cls, table: pa.Table, schema: Schema) -> pa.Table:
+    def convert_table(cls, table: pa.Table | pd.DataFrame, schema: Schema) -> pa.Table:
+        if isinstance(table, pd.DataFrame):
```
cpcloud, line 263: We definitely do not want to mix pandas and pyarrow here, that's the primary reason these are implemented as separate classes.

Can you think of another way to solve this problem that doesn't require conflating the two conversion paths?

cpcloud (changes requested):
We need to avoid mixing pyarrow and pandas conversion paths.

ibis/formats/pyarrow.py
NickCrews: Hmmm makes sense, will take a look.

NickCrews:
OK, I think this brings up a larger philosophical question: Do we want to totally separate the pandas and pyarrow codepaths, or can they rely on each other?

Currently, to get pyarrow results from a backend:
1. for [some backends](https://github.com/ibis-project/ibis/blob/84974fee9f28073fc296a763b9e4a3da611d0315/ibis/backends/sql/__init__.py#L335-L379) we go straight from the DB cursor object to pyarrow arrays, never needing pandas.
2. In the backends that I touch in this PR, we go through the path of db_cursor -> pandas -> pyarrow.

I think the coupling between pandas and pyarrow for this conversion isn't inherently bad (we don't need to implement the db -> pyarrow path!), but I agree that it should be isolated, so we are very clear where we are mixing these two ecosystems, so that for the backends that don't need it, you can just have pyarrow installed, you don't need pandas.

So I see two options:
1. keep this db_cursor -> pandas -> pyarrow path, but just sequester it into some 3rd module that is external to both ibis/formats/pandas.py and ibis/formats/pyarrow.py
2. in these backends that don't have it yet, implement the db_cursor -> pyarrow conversion directly.

I think I would lean towards 2. I want to remove reliance on pandas as much as possible. Possibly this implementation won't be that hard for these other backends.

ibis/formats/pyarrow.py
NickCrews: @cpcloud responded in a top-level comment.

cpcloud:
I think we'd to eventually be able to offer Ibis without requiring `pyarrow` or `pandas`, or least without requiring `pandas`. Many systems are starting to have arrow-native endpoints that don't involve pandas, so `db -> pyarrow` is actually better for those cases.

There's also the potential of using something that doesn't depend on either of those for the core (like printing tables), so I think we'd like to keep things as isolated from one another as possible.

Even more is the fact that sending anything through pandas is likely to result in some kind of type or value alteration that doesn't happen with pyarrow. Especially with NULLs, pandas is likely to do something completely different and incompatible with what pyarrow would do.

NickCrews:
Ok, when I get back to this I'll try the db -> arrow method!

cpcloud:
Is this PR still viable?

NickCrews:
viable, I just stopped needing it personally so the urgency of it dropped a lot compared the 5 million other PRs I have open haha. Feel free to close if you want, and re-open once someone actually finds time to work on it.

PR #8906, opened 2024-04-06
feat(postgres): use port in connection string

NickCrews:
Before this was just ignored, so if you tryied to connect to
anything besides port 5432, it would
just use the default of 5432


cpcloud (approved):
Thanks!

Issue #8581, opened 2024-03-07
feat(ir): more flexible dereferencing support for the right hand side of joins


kszucs:
### Is your feature request related to a problem?

The right hand side of joins is not dereferenced, for temporal joins we need more flexible dereferencing. Blocks #8412 #8537 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8916, opened 2024-04-09
feat(ir): more flexible dereferencing for join right-hand side

kszucs:
The join dereferencing logic required explicit references to the exact right hand side table which disallowed more convenient syntaxes like `t1.join(t2.at_time(t1.c), [t1.a == t2.d])`. 

Resolves https://github.com/ibis-project/ibis/issues/8581

### DIFF
--- a/ibis/backends/impala/tests/snapshots/test_sql/test_join_aliasing/out.sql
+++ b/ibis/backends/impala/tests/snapshots/test_sql/test_join_aliasing/out.sql
@@ -62,7 +62,7 @@ INNER JOIN (
       GROUP BY
         1
     ) AS `t10`
-      ON `t8`.`d` = `t10`.`d`
+      ON `t10`.`d` = `t10`.`d`
   ) AS `t11`
   WHERE
     `t11`.`row_count` < (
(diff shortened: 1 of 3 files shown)


kszucs:
There is a more complicated test case in the impala test suite, inspecting it.

kszucs:
The multi-table dereference mapping calculation must be slightly smarter by tracking the distance from the target relations and picking the closest one (unless there are multiple ones at the same distance in which case an ambiguity error should be raised).

cpcloud:
@kszucs The two failures are pretty easy to fix. Otherwise, is this PR ready to go in?

gforsyth:
CI is green if @kszucs is happy with this.

ibis/backends/impala/tests/snapshots/test_sql/test_join_aliasing/out.sql
```diff
@@ -62,7 +62,7 @@ INNER JOIN (
       GROUP BY
         1
     ) AS `t10`
-      ON `t8`.`d` = `t10`.`d`
+      ON `t10`.`d` = `t10`.`d`
```
cpcloud, line 65: This doesn't look correct to me. Should be `t8` as it was before.

cpcloud (changes requested):

kszucs:
No, I am not happy with this yet.

cpcloud:
Should we close this in favor of #8992?

kszucs:
Yes

Issue #7819, opened 2023-12-20
ux: revisit input value binding which currently supports arbitrary nesting of inputs in sequences and mappings.


kszucs:
### What happened?

Follow-up of https://github.com/ibis-project/ibis/pull/7752

### What version of ibis are you using?

`the-epic-split` branch

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #8304, opened 2024-02-10
feat: expose `Table._ensure_expr()`


NickCrews:
### Is your feature request related to a problem?

I am writing a framework for record linkage on top of Ibis. As part of that, I have an API that takes as arguments
1. A Table
2. *something* that references a column within that table

So the basic examples for point 2 are
1. a string eg "my_col"
2. a deferred eg `_.my_col.upper()[:3]`
3. a lambda eg `lambda table: table.my_col.upper().cast(int)`

It would be nice if there was a universal API on Tables that allowed me convert all of these to a Column. I can't use `__getitem__`, because that can return a Table:

```python
import ibis

t = ibis.memtable({"island": [1, 2, 3, 4, 5]})

print(type(t["island"]))  # Column
print(type(t[_.island]))  # Table
print(type(t[lambda t: t.island]))  # Table
```

[Currently I am using Table._ensure_expr](https://github.com/NickCrews/mismo/blob/40cf632d6f67e539e84b0c46aaec5a772801f864/mismo/block/_array.py), but that feels icky since it is private.

### Describe the solution you'd like

maybe `Table.column(Any) -> Column`?

We should think about if the new method would shadow the name of a column in the Table, but I hope that people aren't nameing their columns "column"...

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
Thanks for the feature request!

Like many other things, the implementation of the binding process is changing quite a bit in `the-epic-split`.

`Table._ensure_expr` is removed in that branch, and replaced with a function called `bind` that universally handles inputs, and converts them into an iterable of expressions.

`bind` is more generic than handling just a single column though. It not only handles strings, deferreds and lambdas, but also selectors, mappings and iterables of all those things.

That `bind` API looks like this:

```python
exprs = bind(table, "island")
exprs = bind(table, _.island)
exprs = bind(table, lambda t: t.island)
exprs = bind(table, s.matches("island")
exprs = bind(table, table.island)
exprs = bind(table, ["island"])
exprs = bind(table, {"eye-land": "island"})
```

@kszucs Thoughts on making this API public after we merge `the-epic-split`?

kszucs:
I think we can expose bind, though not sure whether this should be exposed as a function or a method. Preference?

NickCrews:
[...]

PR #8917, opened 2024-04-09
refactor(api): restrict arbitrary input nesting

kszucs:
Previously we allowed arbitrary nesting of input expressions for various
API methods like `table.join`, `table.group_by`, etc. While this was
backward compatible with `8.0.0` it also exposes additional ambiguity
which can be confusing for users and difficult to reason about.

This change restricts the nesting of input expressions to a
"single level" of nesting with the exception of the first positional
argument which can be a list of expressions, but only if there are no
more positional arguments. Examples:

```python
t.select([t.foo, t.bar])  # OK
t.select(t.foo, t.bar)    # OK
t.select([t.foo], t.bar)  # Error
t.select(t.foo, name=t.bar)  # OK
t.select([t.foo], name=t.bar)  # OK
```

Aims to resolve https://github.com/ibis-project/ibis/issues/7819 and https://github.com/ibis-project/ibis/issues/8304, depends on https://github.com/ibis-project/ibis/pull/8884

### DIFF
--- a/ibis/expr/builders.py
+++ b/ibis/expr/builders.py
@@ -226,17 +226,15 @@ class WindowBuilder(Builder):
         return self.copy(orderings=self.orderings + util.promote_tuple(expr))
 
     def bind(self, table):
-        from ibis.expr.types.relations import bind
-
         if table is None:
             if self._table is None:
                 raise IbisInputError("Cannot bind window frame without a table")
             else:
                 table = self._table.to_expr()
 
-        grouping = bind(table, self.groupings)
-        orderings = bind(table, self.orderings)
-        return self.copy(groupings=grouping, orderings=orderings)
+        return self.copy(
+            groupings=table.bind(self.groupings), orderings=table.bind(self.orderings)
+        )
 
 
 class LegacyWindowBuilder(WindowBuilder):
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -110,7 +110,9 @@ def bind(table: Table, value) -> Iterator[ir.Value]:
     elif isinstance(value, Selector):
         yield from value.expand(table)
     elif callable(value):
-        yield value(table)
+        # rebind, otherwise the callable is required to return an expression
+        # which would preclude support for expressions like lambda _: 2
+        yield from bind(table, value(table))
     else:
         yield literal(value)
 
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -229,12 +229,12 @@ class Table(Expr, _FixedTextJupyterMixin):
         # bind keyword arguments where each entry can produce only one value
         # which is then named with the given key
         for key, arg in kwargs.items():
-            try:
-                (value,) = bind(self, arg)
-            except ValueError:
+            bindings = tuple(bind(self, arg))
+            if len(bindings) != 1:
                 raise com.IbisInputError(
                     "Keyword arguments cannot produce more than one value"
                 )
+            (value,) = bindings
             values.append(value.name(key))
 
         # dereference the values to `self`
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -91,18 +91,11 @@ def _regular_join_method(
     return f
 
 
-# TODO(kszucs): should use (table, *args, **kwargs) instead to avoid interpreting
-# nested inputs
-def bind(table: Table, value: Any, int_as_column=False) -> Iterator[ir.Value]:
+def bind(table: Table, value) -> Iterator[ir.Value]:
     """Bind a value to a table expression."""
     if isinstance(value, str):
         # TODO(kszucs): perhaps use getattr(table, value) instead for nicer error msg
         yield ops.Field(table, value).to_expr()
-    elif isinstance(value, bool):
-        yield literal(value)
-    elif int_as_column and isinstance(value, int):
-        name = table.columns[value]
-        yield ops.Field(table, name).to_expr()
     elif isinstance(value, ops.Value):
         yield value.to_expr()
     elif isinstance(value, Value):
@@ -116,13 +109,6 @@ def bind(table: Table, value: Any, int_as_column=False) -> Iterator[ir.Value]:
         yield value.resolve({"_": table})
     elif isinstance(value, Selector):
         yield from value.expand(table)
-    elif isinstance(value, Mapping):
-        for k, v in value.items():
-            for val in bind(table, v, int_as_column=int_as_column):
-                yield val.name(k)
-    elif util.is_iterable(value):
-        for v in value:
-            yield from bind(table, v, int_as_column=int_as_column)
     elif callable(value):
         yield value(table)
     else:
@@ -135,7 +121,7 @@ def unwrap_aliases(values: Iterator[ir.Value]) -> Mapping[str, ir.Value]:
     for value in values:
         node = value.op()
         if node.name in result:
-            raise com.IntegrityError(
+            raise com.IbisInputError(
                 f"Duplicate column name {node.name!r} in result set"
             )
         if isinstance(node, ops.Alias):
@@ -143,29 +129,6 @@ def unwrap_aliases(values: Iterator[ir.Value]) -> Mapping[str, ir.Value]:
         else:
             result[node.name] = node
     return result
-
-
-def dereference_values(
-    parents: Iterable[ops.Parents], values: Mapping[str, ops.Value]
-) -> Mapping[str, ops.Value]:
-    """Trace and replace fields from earlier relations in the hierarchy.
-
-    For more details see :class:`ibis.expr.rewrites.DerefMap`.
-
-    Parameters
-    ----------
-    parents
-        The relations we want the values to point to.
-    values
-        The values to dereference.
-
-    Returns
-    -------
-    The same mapping as `values` but with all the dereferenceable fields
-    replaced with the fields from the parents.
-    """
-    dm = DerefMap.from_targets(parents)
-    return {k: dm.dereference(v) for k, v in values.items()}
 
 
 @public
@@ -246,6 +209,40 @@ class Table(Expr, _FixedTextJupyterMixin):
             return where
 
         return where.resolve(self)
+
+    def bind(self, *args, **kwargs):
+        # allow the first argument to be either a dictionary or a list of values
+        if len(args) == 1:
+            if isinstance(args[0], dict):
+                kwargs = {**args[0], **kwargs}
+                args = ()
+            else:
+                args = util.promote_list(args[0])
+
+        # bind positional arguments
+        values = []
+        for arg in args:
+            values.extend(bind(self, arg))
+
+        # bind keyword arguments where each entry can produce only one value
+        # which is then named with the given key
+        for key, arg in kwargs.items():
+            try:
+                (value,) = bind(self, arg)
+            except ValueError:
+                raise com.IbisInputError(
+                    "Keyword arguments cannot produce more than one value"
+                )
+            values.append(value.name(key))
+
+        # dereference the values to `self`
+        dm = DerefMap.from_targets(self.op())
+        result = []
+        for original in values:
+            value = dm.dereference(original.op()).to_expr()
+            value = value.name(original.get_name())
+            result.append(value)
+        return tuple(result)
 
     def as_scalar(self) -> ir.ScalarExpr:
         """Inform ibis that the table expression should be treated as a scalar.
@@ -769,7 +766,12 @@ class Table(Expr, _FixedTextJupyterMixin):
             limit, offset = util.slice_to_limit_offset(what, self.count())
             return self.limit(limit, offset=offset)
 
-        values = tuple(bind(self, what, int_as_column=True))
+        args = [
+            self.columns[arg] if isinstance(arg, int) else arg
+            for arg in util.promote_list(what)
+        ]
+        values = self.bind(args)
+
         if isinstance(what, (str, int)):
             assert len(values) == 1
             return values[0]
@@ -954,7 +956,7 @@ class Table(Expr, _FixedTextJupyterMixin):
         from ibis.expr.types.groupby import GroupedTable
 
         by = tuple(v for v in by if v is not None)
-        groups = bind(self, (by, key_exprs))
+        groups = self.bind(*by, **key_exprs)
         return GroupedTable(self, groups)
 
     # TODO(kszucs): shouldn't this be ibis.rowid() instead not bound to a specific table?
@@ -1133,15 +1135,12 @@ class Table(Expr, _FixedTextJupyterMixin):
 
         node = self.op()
 
-        groups = bind(self, by)
-        metrics = bind(self, (metrics, kwargs))
-        having = tuple(bind(self, having))
+        groups = self.bind(by)
+        metrics = self.bind(metrics, **kwargs)
+        having = self.bind(having)
 
         groups = unwrap_aliases(groups)
         metrics = unwrap_aliases(metrics)
-
-        groups = dereference_values(node, groups)
-        metrics = dereference_values(node, metrics)
 
         # the user doesn't need to specify the metrics used in the having clause
         # explicitly, we implicitly add them to the metrics list by looking for
@@ -1160,7 +1159,7 @@ class Table(Expr, _FixedTextJupyterMixin):
 
         if having:
             # apply the having clause
-            agg = agg.filter(*having)
+            agg = agg.filter(having)
             # remove any metrics that were only used in the having clause
             if metrics != original_metrics:
                 agg = agg.select(*groups.keys(), *original_metrics.keys())
@@ -1319,7 +1318,7 @@ class Table(Expr, _FixedTextJupyterMixin):
             having = lambda t: t.count() == 1
             method = "first"
         elif keep in ("first", "last"):
-            having = None
+            having = ()
             method = keep
         else:
             raise com.IbisError(
@@ -1327,11 +1326,7 @@ class Table(Expr, _FixedTextJupyterMixin):
             )
 
         aggs = {col.get_name(): getattr(col, method)() for col in (~on).expand(self)}
-
-        gb = self.group_by(on)
-        if having is not None:
-            gb = gb.having(having)
-        res = gb.agg(**aggs)
+        res = self.aggregate(aggs, by=on, having=having)
 
         assert len(res.columns) == len(self.columns)
         if res.columns != self.columns:
@@ -1672,9 +1667,8 @@ class Table(Expr, _FixedTextJupyterMixin):
         │     2 │ B      │     6 │
         └───────┴────────┴───────┘
         """
-        keys = bind(self, by)
+        keys = self.bind(*by)
         keys = unwrap_aliases(keys)
-        keys = dereference_values(self.op(), keys)
         if not keys:
             raise com.IbisError("At least one sort key must be provided")
 
@@ -1921,7 +1915,7 @@ class Table(Expr, _FixedTextJupyterMixin):
         # string and integer inputs are going to be coerced to literals instead
         # of interpreted as column references like in select
         node = self.op()
-        values = bind(self, (exprs, mutations))
+        values = self.bind(*exprs, **mutations)
         values = unwrap_aliases(values)
         # allow overriding of fields, hence the mutation behavior
         values = {**node.fields, **values}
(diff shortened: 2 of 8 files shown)


ibis/expr/types/relations.py
```diff
@@ -293,6 +281,19 @@ def _bind_reduction_filter(self, where):
 
         return where.resolve(self)
 
+    def bind(self, *args, **kwargs):
```
kszucs, line 284: Needs docstrings and must be covered with tests, but first we should have a consensus on the API behaviour.

kszucs:
There are some ambiguities around the first arguments which can be a list of expressions:

```py
table.select(["a", ["b"]])

# constructs

r0 := UnboundTable: table
  a int8
  b int16
  c int32
  d int64
  e float32
  f float64
  g string
  h boolean
  i timestamp
  j date
  k time

Project[r0]
  a:      r0.a
  ('b',): ['b']
```

`a` is interepreted as a `column` whereas `["b"]` interpreted as an `array literal`.

kszucs:
Apparently we also have a backend test case using a mapping as a single positional argument.

cpcloud:
Why can't we keep

```python
t.select([t.foo], t.bar)
```
?

It doesn't seem meaningfully different from allowing

```python
t.select([t.foo], bar=t.bar)
```

kszucs:
If we allow:

```py
t.select([t.foo], t.bar)
```

then should we allow

```py
t.select([t.foo], [t.bar])
```

as well?

cpcloud:
I think so. It's really things like `t.select([[...]], [[...]])` that should be disallowed.

kszucs:
My preference would be to either provide args as variadics or as a single list-like argument. Even then we have confusion when to treat list-like inputs as literals vs value list.

jcrist:
We talked today during triage and decided that at minimum we want to remove the support for arbitrarily nested expressions (as Phillip noted above).

I actually agree with the initial proposal at the top (only accept lists of column refs if select is a 1-arg function, otherwise treat it as variadic). Coincidentally this matches how polars `select` works, so there's some ecosystem-consistency there. The trick with doing this as a breaking change though is that previously working code may instead silently be treated as an array literal.

With this PR, what happens here? Does this error, or does it silently coerce these to array literals?

```python
t.select(["a", "b"], ["c", "d"])  # is this two array literals or an error?

t.select("a", "b", ["c", "d"])   # is ["c", "d"] an array literal, or does this error?
```

Either way, I think we _definitely_ want to forbid the case that phillip brought up before 9.0. I'm not opposed to the full changes here either, I think they make things more consistent, but we can always do that later IMO in 10.0.

kszucs:
```py
t.select(["a", "b"], ["c", "d"])  # is this two array literals or an error?
```
yes, two array literals, not erroring

```py
t.select("a", "b", ["c", "d"])   # is ["c", "d"] an array literal, or does this error?
```
yes, an array literal, not erroring

kszucs:
I would prefer deciding what should be the ideal long term behaviour then extend that to be backward compatible and either raise a deprecation warning or provide a config option to keep the previous behaviour.

cpcloud:
Let's keep the full changeset here, and make sure to have both a `BREAKING CHANGE` bit as well as documentation on the behavior in the relevant APIs (`select` and `mutate`).

kszucs:
Another thing I'm not sure about is whether we should do the dereferencing here or not? Like should `t.bind(totally_different_table.col)` raise?

cpcloud:
Seems like it _might_ keep `bind` more flexible and easier to maintain if we dereference elsewhere.

kszucs:
Actually it could help code consolidation. I can try it after #8916 gets merged.

ibis/tests/expr/test_table.py
```diff
@@ -2082,3 +2082,67 @@ def test_unbind_with_namespace():
 
     assert s.op() == expected.op()
     assert s.equals(expected)
+
+
```
NickCrews, line 2086: Would love to see test cases for every ifelse statement in the implementation:
- bools (raw, no ibis.literal(bool))
- ints (raw)
- lambda
- a Table

NickCrews (changes requested):

ibis/tests/expr/test_table.py
NickCrews: Also a test for deferreds, eg _.bind("a")

cpcloud:
We should probably kill `destructure()` now that proper alternatives are available.

cpcloud:
I'm going to rip out `destructure` in a separate PR.

ibis/expr/types/relations.py
```diff
@@ -95,18 +95,11 @@ def f(  # noqa: D417
     return f
 
 
-# TODO(kszucs): should use (table, *args, **kwargs) instead to avoid interpreting
-# nested inputs
-def bind(table: Table, value: Any, int_as_column=False) -> Iterator[ir.Value]:
+def bind(table: Table, value) -> Iterator[ir.Value]:
```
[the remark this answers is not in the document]
NickCrews: that workflow makese sense, thanks! Indeed these requests aren't hard blockers. I think my request for test cases is a blocker for me though, see my comment down there.

ibis/tests/expr/test_table.py
kszucs: Table was already there, added tests for the rest.
`bind(_, "column_name")` case isn't supposed to work since `bind()` expects an actual table expression not a deferred as the first argument. Adding a test case for `bind(t, _.column_name)`.

kszucs:
> I'm going to rip out destructure in a separate PR. 

Updated the test case to expect an input error rather than an integrity error. We should update `unwrap_aliases()` to raise ibis input error instead of integrity error too.

ibis/tests/expr/test_table.py
```diff
@@ -267,10 +267,10 @@ def test_projection_array_expr(table):
...
-def test_projection_no_expr(table, empty):
-    with pytest.raises(com.IbisTypeError, match="must select at least one"):
-        table.select(empty)
+# @pytest.mark.parametrize("empty", [list(), dict()])
+# def test_projection_no_expr(table, empty):
+#     with pytest.raises(com.IbisTypeError, match="must select at least one"):
+#         table.select(empty)
 
 
 # FIXME(kszucs): currently bind() flattens the list of expressions, so arbitrary
```
kszucs, line 276: Enable this test case.

ibis/tests/expr/test_table.py
```diff
@@ -267,10 +267,10 @@ def test_projection_array_expr(table):
     assert_equal(result, expected)
 
 
-@pytest.mark.parametrize("empty", [list(), dict()])
-def test_projection_no_expr(table, empty):
-    with pytest.raises(com.IbisTypeError, match="must select at least one"):
-        table.select(empty)
+# @pytest.mark.parametrize("empty", [list(), dict()])
+# def test_projection_no_expr(table, empty):
```
kszucs, line 271: Remove this test case.

ibis/tests/expr/test_table.py
kszucs: Done.

kszucs:
One confusing case:

```py
In [4]: table = ibis.table({"a": "int", "b": "string"})

In [5]: table.select(["a", ["b"]])
Out[5]:
r0 := UnboundTable: unbound_table_1
  a int64
  b string

Project[r0]
  a:      r0.a
  ('b',): ['b']
```

ibis/expr/types/relations.py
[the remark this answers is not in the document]
kszucs: `t.bind()` also dereference the value to `t` now. `bind()` simply coerces the input to a value expression and remains a private function.

ibis/expr/types/relations.py
kszucs: Tests are there but the docstring is missing.

ibis/expr/builders.py
```diff
@@ -226,17 +226,17 @@ def order_by(self, expr) -> Self:
         return self.copy(orderings=self.orderings + util.promote_tuple(expr))
 
     def bind(self, table):
-        from ibis.expr.types.relations import bind
+        # from ibis.expr.types.relations import bind
```
cpcloud, line 229:
```suggestion
```

cpcloud (approved):

cpcloud:
@NickCrews Can you review this again? This is a remaining blocker for the 9.0 release which we'd like to get out early next week.

ibis/expr/types/relations.py
```diff
@@ -769,7 +766,12 @@ def __getitem__(self, what):
             limit, offset = util.slice_to_limit_offset(what, self.count())
             return self.limit(limit, offset=offset)
 
-        values = tuple(bind(self, what, int_as_column=True))
+        args = [
+            self.columns[arg] if isinstance(arg, int) else arg
+            for arg in util.promote_list(what)
+        ]
+        values = self.bind(args)
```
NickCrews, line 773: Am I understanding this right that `.__getitem__` accepts ints, but `.select` doesn't? This is a separate PR, but I would love to rip the bandaid off and just disallow ints everywhere.

ibis/tests/expr/test_table.py
```diff
@@ -2077,3 +2077,87 @@ def test_unbind_with_namespace():
...
+    expected = (t.a, t.b)
+    assert eq(exprs, expected)
+
+    # single selector arg
+    exprs = t.bind(s.all())
+    expected = (t.a, t.b)
+    assert eq(exprs, expected)
+
+    # single tuple arg
+    exprs = t.bind([1, "a"])
```
NickCrews, line 2119: based on the comment I think this was supposed to be a tuple, but its a list. But maybe go way weird and pass in an iterable to really stress it: `(x for x in [1, "a"])`

ibis/tests/expr/test_table.py
```diff
@@ -2077,3 +2077,87 @@ def test_unbind_with_namespace():
...
+
+    t = ibis.table({"a": "int", "b": "string"}, name="t")
+
+    # boolean literals
+    exprs = t.bind(True, False)
+    expected = (ibis.literal(True), ibis.literal(False))
+    assert eq(exprs, expected)
+
+    # int literals
+    exprs = t.bind(1, 2)
```
NickCrews, line 2094: can we get a test for when the resulting columns would have duplicate names? eg `t.bind(1, 1)`

ibis/tests/expr/test_table.py
```diff
@@ -2077,3 +2077,87 @@ def test_unbind_with_namespace():
...
     assert s.op() == expected.op()
     assert s.equals(expected)
+
+
+def test_table_bind():
+    def eq(left, right):
+        return all(a.equals(b) for a, b in zip(left, right))
+
+    t = ibis.table({"a": "int", "b": "string"}, name="t")
+
```
NickCrews, line 2087: would love a case for no args eg `t.bind()`. I would expect this to result in an empty tuple. Erroring would be OK if there is some problem with the empty tuple, but that would be my second choice.

ibis/tests/expr/test_table.py
```diff
@@ -2077,3 +2077,87 @@ def test_unbind_with_namespace():
...
+    # boolean literals
+    exprs = t.bind(True, False)
+    expected = (ibis.literal(True), ibis.literal(False))
+    assert eq(exprs, expected)
+
```
NickCrews, line 2092: Would love 2 cases, one for for a completely unrelated table `ibis.table({"z": "int"}).bind(t)` and another for a related one (I can only think of a join result but maybe there's some other instance of this) `t.right_join(ibis.table({"a": "int"}), "a").bind(t.b)`. I'm not that attached to what the behavior is, but I think we should write it down so we can test for future breakages.

ibis/expr/types/relations.py
cpcloud: I think it's still useful to be able to do this. I use it during ad-hoc interactive analysis for referring to columns that have generated names.

ibis/tests/expr/test_table.py
cpcloud: Don't we already have a test for the latter?

ibis/tests/expr/test_table.py
cpcloud: This doesn't seem like a blocker to me.

ibis/tests/expr/test_table.py
```diff
@@ -2077,3 +2077,87 @@ def test_unbind_with_namespace():
...
+    # int literals
+    exprs = t.bind(1, 2)
+    expected = (ibis.literal(1), ibis.literal(2))
+    assert eq(exprs, expected)
+
+    # lambda input
+    exprs = t.bind(lambda t: t.a, lambda t: t.b)
```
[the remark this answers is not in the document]
cpcloud: This doesn't seem like a blocker.

ibis/expr/types/relations.py
```diff
@@ -247,6 +210,40 @@ def _bind_reduction_filter(self, where):
 
         return where.resolve(self)
 
+    def bind(self, *args, **kwargs):
+        # allow the first argument to be either a dictionary or a list of values
+        if len(args) == 1:
```
NickCrews, line 215: I'm a -.5 on this special casing. This causes `t.bind([1,"a"])`  to be `(ibis.literal(1), t.a)` (behavior A). I would prefer for it to be `ibis.array([1, "a"])` (behavior B). I looked back through this PR to try to get the context, but I might not have gotten all of it. As I understand it, switching A to B

- Would be the preferred behavior by everyone else if we were starting from a clean slate with no history.
- Might silently change the behavior of existing user code (we could add an easily-silencable warning for the 9.0 and maybe 10.0 versions?). This is the main downside.
- Would this cause any other paper cuts or make impossible some other task I'm not thinking of?

ibis/tests/expr/test_table.py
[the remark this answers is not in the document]
cpcloud: Ok, this isn't actually supported. I'll push up a fix.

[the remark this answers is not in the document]
cpcloud: Done.

ibis/expr/types/relations.py
```diff
@@ -247,6 +210,40 @@ def _bind_reduction_filter(self, where):
...
+        values = []
+        for arg in args:
+            values.extend(bind(self, arg))
+
+        # bind keyword arguments where each entry can produce only one value
+        # which is then named with the given key
+        for key, arg in kwargs.items():
+            try:
+                (value,) = bind(self, arg)
+            except ValueError:
```
NickCrews, line 234: This would gobble up unrelated errors: `t.bind(lambda t: int("foo"))`. Let's be a little slower here:
```
bindings = bind(self, arg)
if len(bindings) != 1:
    raise com.IbisInputError(
                    "Keyword arguments cannot produce more than one value"
                )
values.append(bindings[0].name(key))
```

ibis/tests/expr/test_table.py
cpcloud: Added a test with a generator.

ibis/tests/expr/test_table.py
NickCrews: do you mean the `t.bind(t)` case? I was hoping for something a bit more complex. Or did I miss something?

ibis/tests/expr/test_table.py
NickCrews: I see what you're saying, but I think I still want it. Perhaps in the future we switch to a different implementation, would still want the test.

ibis/expr/types/relations.py
NickCrews: yeah I see what you're saying. I was trying to get these two methods to be totally equivalent, but I think I am convinced that the utility is worth the inconsistency.

NickCrews:
I just took a long look.

Most things look great, a few nits, but the one substantial thing is the `t.bind([1,"a"])` behavior. I have an inline comment above. Can someone fill me in on the context there a little more? I don't think this is a blocker for me.

ibis/tests/expr/test_table.py
cpcloud: Let's separate blocking comments from non blocking comments.

Adding a test for this should in no way block this PR from landing.

I disagree that we should add this test. It adds no material value to testing the behavior of `bind`. Adding a test that exercises duplicate column detection is just wasted CPU cycles and maintainer effort.

cpcloud:
@NickCrews Thanks for doing a deep dive on tests. There were indeed a few cases that weren't tested and also the `lambda _: 2` case wasn't yet implemented. Appreciate the feedback!

ibis/expr/types/relations.py
cpcloud: > Might silently change the behavior of existing user code (we could add an easily-silencable warning for the 9.0 and maybe 10.0 versions?).

The benefit of behavior B doesn't seem worth the pain this would cause given history.

I don't think changing to behavior B makes anything impossible, but it's a breaking API change that seems unnecessary.

ibis/expr/types/relations.py
cpcloud: Yeah, we should avoid catching the error here. I'll fix it up.

cpcloud: Done and test added.

ibis/expr/types/relations.py
kszucs: > Would be the preferred behavior by everyone else if we were starting from a clean slate with no history.

It would be my preference, but yes, that would be too big of a breakage. If we wish to change it we can also do that for later releases.

ibis/expr/types/relations.py
kszucs: Nice catch! Thanks for the patch too!

ibis/tests/expr/test_table.py
kszucs: So bind's responsibility to coerce the input arguments to a series of value expressions, nothing more.

Other methods like `.select() / .aggregate() / ...` do ensure that there are no duplicate names by calling `unwrap_aliases()`, so the check does happen but at a different level.

NickCrews:
Cool this looks great then!

ibis/tests/expr/test_table.py
NickCrews: I think it is a blocker for me, at the very least to simply raise a NotImplemented error so we keep the door open for ourselves later

ibis/expr/types/relations.py
NickCrews: sounds good let's keep it as is for now. Thank you!

ibis/tests/expr/test_table.py
NickCrews: oooh, actually there is also `t.bind(unrelated_table.column)` which currently returns `unrelated_table.column`, which I don't think quite makes sense (I would expect an error unless we can walk up the join chain to find it). So actually all this seems like undefined behavior. I think if we just don't publicize this API or add it to the type hints then I'm happy. Would love to make this more flushed out later but it is a bigger task than we should block on.

ibis/tests/expr/test_table.py
NickCrews: OK, I think this is a misunderstanding caused by me not explaining what I thought the expected behavior should be. I have always used .bind in the larger context of `t.mutate(t.bind(x))`, so I have never wanted duplicate column names, so I originally thought that `t.bind(1, 1)` should error. But

> So bind's responsibility to coerce the input arguments to a series of value expressions, nothing more.

makes sense. I think you're right, dupe results should be allowed, and treated the same as any other result, and thus don't need a test. I was operating under the assumption they should be treated differently, and thus should have a test.

ibis/tests/expr/test_table.py
NickCrews: Or I guess per the conversation below, maybe again this isn't the responsibility of .bind to do this check

ibis/tests/expr/test_table.py
NickCrews: I would maybe tweak the phrasing of the responsibility to "coerce the arguments to a tuple of value expressions (if they aren't already)." to emphasize that we don't do anything more clever/furhter inspection if the arg is already a value expression. This makes it more clear that `t.bind(other_table.column)` just results in `other_table.column`. I think this would be a good candidate to go in the summary line of the docstring.

cpcloud:
@NickCrews Can you approve the PR?

NickCrews (approved):

NickCrews:
Sorry I was on the mobile app where the approval status isn't even shown so i didn't know there was anything to do

PR #8919, opened 2024-04-09
test(snowflake): fix snapshots

cpcloud:
Fixes snapshots.

I also disabled dependency caching except in cases where we're installing all extras: benchmarking and doctests. It doesn't seem like we're getting much from it in other cases except [bizarre failures](https://github.com/ibis-project/ibis/actions/runs/8632266124/job/23662562629?pr=8758).

PR #8923, opened 2024-04-10
feat(api): allow *arg syntax with GroupedTable methods

NickCrews:
this brings it in line with the normal Table methods

### DIFF
--- a/ibis/expr/types/groupby.py
+++ b/ibis/expr/types/groupby.py
@@ -63,7 +63,7 @@ class GroupedTable(Concrete):
         else:
             return GroupedArray(field, self)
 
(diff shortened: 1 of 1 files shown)


NickCrews:
It looks to me like the dask failures are flakes unrelated to this PR? I couldn't repro locally with `just ci-check -m dask --numprocesses auto --dist=loadgroup` (Im on macOS, not ubuntu like the failures, so maybe that's an issue), and looking at the stacktrace it looks like the error is entirely within dask code, it never even touches ibis code.

cpcloud:
I think you were hitting the 3.11.9 issue in upstream dask. That was addressed yesterday in #8919 .

cpcloud (approved):

Issue #8926, opened 2024-04-10
feat: add `.pipe()` to ibis


koaning:
### Is your feature request related to a problem?

I'm working on an ibis livestream today and while there's a bunch of stuff to like, I noticed that if I want to apply functions to tables that I gotta do that via something like:

```python
def aggregate(tbl):
    return tbl.group_by(tbl.developer).aggregate([tbl.count().name("c")])

aggregate(t)
```

This works, but I worry about the elegance in the long run. In this case the function is doing a simple aggregation but what if I have a function that adds sessions and another that removes bots before I aggregate? The call might look something like this.

```python
aggregate(remove_bots(add_session(t)))
```

It may even get more nesty if I have arguments to these functions as well.

```python
aggregate(remove_bots(add_session(t, cutoff=5), limit=10)))
```

What I like to do instead when I write polars or pandas is to do something like this:

```python
(
  df
    .pipe(add_session, cutoff=5)
    .pipe(remove_bots, limit=10)
    .pipe(aggregate)
)
```

I was able to get something similar working in ibis via:

```python
from ibis.expr.types.relations import Table

def pipe(tbl, func, *args, **kwargs):
    return func(tbl, *args, **kwargs)

Table.pipe = pipe
```

This is a hacky solution by all means, but at a quick glance ... it seems to pull off the trick!

```python
t.pipe(aggregate)
```

### What is the motivation behind your request?

This approach should allow folks to split their concerns a bit more. I should mention that I am somewhat opinionated on this. I've given a [somewhat popular Python talk](https://www.youtube.com/watch?v=yXGCKqo5cEY) about this style of writing pipelines and I also [introduced the `.pipe()` method in polars](https://github.com/pola-rs/polars/pull/82).

It feels like something that would fit this library too and if folks don't mind ... I'll gladly have a proper stab at it. 

### Describe the solution you'd like

Something like this should work. 

```python
from ibis.expr.types.relations import Table

def pipe(tbl, func, *args, **kwargs):
    return func(tbl, *args, **kwargs)

Table.pipe = pipe
```

I'd love to hear about any edge cases, if any. I'm totally unaware of the details in this library so I'll gladly hear it if I'm overlooking something. The typing probably needs to be considered, but to me it feels like this feature may be relatively easy to add.

### What version of ibis are you running?

8.0.0

### What backend(s) are you using, if any?

I tried this on DuckDB and Pandas.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8927, opened 2024-04-10
Adding `pipe` for method chaining

koaning:
A first stab at https://github.com/ibis-project/ibis/issues/8926. This is totally to help get the ball rolling if folks prefer we discuss this in the issue first that's totally cool and I can create a new PR after.

## Description of changes

Adds a `.pipe()` method to allow for method chaining. 

## Issues closed

Resolves https://github.com/ibis-project/ibis/issues/8926

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -981,6 +981,9 @@ class Table(Expr, _FixedTextJupyterMixin):
         by = tuple(v for v in by if v is not None)
         groups = bind(self, (by, key_exprs))
         return GroupedTable(self, groups)
+    
+    def pipe(self, func, *args, **kwargs):
+        return func(self, *args, **kwargs)
 
     # TODO(kszucs): shouldn't this be ibis.rowid() instead not bound to a specific table?
     def rowid(self) -> ir.IntegerValue:


koaning:
I can do that, sure, but just to check ... do we agree that we want this feature in? I'd prefer to get a confirmation in before going down the formal path.

lostmygithubaccount:
can't speak for Phillip, but it seems like a reasonable addition to me -- probably not something I'd personally use but I think we'd welcome the contribution!

koaning:
That's enough for me to do a bit more work :)

One thing, a bunch of tests seem to fail that aren't related to my PR. Shall I ignore those for now?

lostmygithubaccount:
they might be fixed by now? you will want to run the pre-commit to lint/foramt (looks like just removing trailing whitespace), I can update the PR title for the conventional commits one

koaning:
Mhm ... ok that's unexpected it seems that there's already a `test_pipe.py` file that tests a `.pipe()` method. This is referring to a method that's attached to the `Expr` class.

I'm now wondering if I am adding a new feature of if I am going to fix an old one. Let me dive in a bit ...

koaning:
Ahhh wait, it's currently implemented on `table()` but not on a backend. Will do that now.

koaning:
Right. I think the issue here was that my notebook had a typo that I mistook for a feature that was missing. It seems `pipe` is already implemented. No need for this PR. Will also close the issue.

lostmygithubaccount:
I was not aware of that! sorry for the spinning on this 🙃 thanks for working on it regardless!

PR #8928, opened 2024-04-10
chore(dev-tools): remove unnecessary `all` extra

cpcloud:
Removes the `all` extra from `pyproject.toml`. The `--all-extras` flag now works across the board in `poetry` so this manually maintained extra is no longer necessary.


gforsyth (approved):
:shipit:

PR #8941, opened 2024-04-11
docs(release-notes): fix up version header

### LINKED ISSUE #8624: docs: Make links to sections in release notes stable across releases

cpcloud:
Add permalinks to release notes versions.

Fixes #8624.


cpcloud:
Not a huge fan of having to write Lua to do this, but c'est la vie.

amoeba:
Hey @cpcloud, following https://github.com/quarto-dev/quarto/issues/413, this seems to work for me locally, in case that would work here:

```patch
diff --git a/docs/release_notes.md b/docs/release_notes.md
index f8def1134..e28af30ea 100644
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -1,7 +1,7 @@
 Release notes
 ---

-## [8.0.0](https://github.com/ibis-project/ibis/compare/7.2.0...8.0.0) (2024-02-05)
+## [8.0.0](https://github.com/ibis-project/ibis/compare/7.2.0...8.0.0) (2024-02-05) {#8.0.0}


 ### ⚠ BREAKING CHANGES
```

cpcloud:
It might, but the automation around release notes makes it probably roughly the same amount of work as what we currently have

cderv:
@amoeba yes this  is indeed (bad?) Pandoc default for auto identifier for section that does not have explicit and don't start with letter. 
Not that `# v8.0.0` would correctly get you `id="v8.0.0"` on the HTML  - any header starting with a letter first really. 

but probably not what you want here. I'm sharing just in case changing headers would be worth making to prevent special processing.

amoeba:
Right, thanks for pointing that out @cderv. Pandoc is probably saving users from some awful edge cases elsewhere else.

Issue #8888, opened 2024-04-04
test: verify that existing `pyspark` backend functionality works in streaming mode


mfatihaktas:
### Is your feature request related to a problem?

As described in the meta-issue https://github.com/ibis-project/ibis/issues/8868, our initial understanding is that Spark SQL has the same syntax for both batch and streaming. This is why, anything at the intersection of these two should just work.

### Describe the solution you'd like

We need to verify our understanding by testing the existing `pyspark` backend functionality in streaming mode. This includes adding a streaming test env and streaming-specific tests under `ibis/backends/pyspark/tests/`.

### What version of ibis are you running?

8.0.0

### What backend(s) are you using, if any?

pyspark

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8945, opened 2024-04-11
test(pyspark): validate pyspark-specific tests for streaming

mfatihaktas:
## Description of changes

Aims to address https://github.com/ibis-project/ibis/issues/8888.

### DIFF
--- a/ibis/backends/pyspark/__init__.py
+++ b/ibis/backends/pyspark/__init__.py
@@ -303,8 +303,60 @@ class Backend(SQLBackend, CanCreateDatabase):
         df = self._session.createDataFrame(data=op.data.to_frame(), schema=schema)
         df.createOrReplaceTempView(op.name)
 
+    def _execute_stream(self, spark_df: DataFrame) -> pd.DataFrame:
+        # TODO (mehmet): Spark provides memory sink to write the
+        # query result as an in-memory table:
+        # "Memory sink (for debugging) - The output is stored in
+        # memory as an in-memory table. Both, Append and Complete
+        # output modes, are supported. This should be used for
+        # debugging purposes on low data volumes as the entire
+        # output is collected and stored in the driver's
+        # memory. Hence, use it with caution."
+        # Ref: https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html#output-sinks
+        #
+        # E.g.,
+        # cursor.query.writeStream.format("memory").queryName("table_name").start()
+        #
+        # This in-memory table might conflict with those defined
+        # by the user. Hence, choosing to use the file-sink
+        # instead. Though, I don't think this is ideal either.
+
+        import tempfile
+
+        import pandas as pd
+
+        tmpdir = tempfile.mkdtemp()
+        dir_ = f"{tmpdir}/pyspark_fetch_from_cursor"
+        (
+            spark_df.writeStream.format("parquet")
+            .option("checkpointLocation", dir_)
+            .option("path", dir_)
+            .trigger(availableNow=True)
+            .start()
+            .awaitTermination()
+        )
+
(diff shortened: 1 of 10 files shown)


ibis/backends/pyspark/__init__.py
```diff
@@ -303,8 +303,60 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
...
+        files = glob.glob(f"{dir_}/*.parquet")
+        df_list = [pd.read_parquet(f) for f in files]
+        df = pd.concat(df_list, ignore_index=True)
+
+        return df
+
     def _fetch_from_cursor(self, cursor, schema):
-        df = cursor.query.toPandas()  # blocks until finished
+        if cursor.query.isStreaming:
+            df = self._execute_stream(cursor.query)
```
chloeh13q, line 356: I don't think we actually support converting the results back into pandas df in the Flink backend right now... Also these outputs can get arbitrarily large because it runs on continuous data. I get that the intention is the maintain the same interface/UX across streaming and batch jobs, but I wonder if this makes sense

ibis/backends/pyspark/tests/conftest.py
```diff
@@ -22,22 +25,41 @@ def set_pyspark_database(con, database):
 class TestConf(BackendTest):
     deps = ("pyspark",)
 
-    def _load_data(self, **_: Any) -> None:
+    def _load_data_helper(self, for_streaming: bool = False):
```
chloeh13q, line 28: just `streaming`?

ibis/backends/pyspark/tests/conftest.py
```diff
@@ -22,22 +25,41 @@ def set_pyspark_database(con, database):
...
         sort_cols = {"functional_alltypes": "id"}
 
+        if for_streaming:
+            # "By default, Structured Streaming from file based sources
+            # requires you to specify the schema, rather than rely on
+            # Spark to infer it automatically. ... .  For ad-hoc use
+            # cases, you can re-enable schema inference by setting
+            # spark.sql.streaming.schemaInference to true."
+            # Ref: https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html#schema-inference-and-partition-of-streaming-dataframesdatasets
+            s.sql("set spark.sql.streaming.schemaInference=true")
```
chloeh13q, line 46: You can pass this as a config to the session builder https://github.com/ibis-project/ibis/blob/b8894d01fde9cba7dc06b2182b7ee5f42f6903d8/ibis/backends/pyspark/tests/conftest.py#L169-L191. I think that helps group everything together instead of having this single line of SQL here

ibis/backends/pyspark/tests/conftest.py
```diff
@@ -177,48 +204,139 @@ def connect(*, tmpdir, worker_id, **kw):
...
+    # `TestConf` class for streaming, but only need to create
+    # streaming counterparts for the test tables. Still added this
+    # class to keep the testing uniform with Flink. This class is used
+    # in `con_streaming()` fixture (streaming counterpart of `con()`)
+    # to create a new spark session and load the `***_streaming`
+    # tables for testing.  However, either `con()` or
+    # `con_streaming()` can be used to execute any batch/streaming
+    # job. This is why, we set `autouse=True` for `con_streaming()`
+    # to create the streaming tables, and then rely solely on `con()`
+    # to operate on those tables in the tests.
```
chloeh13q, line 220: It makes sense to try to keep the testing uniform, but I also wonder if it's cleaner/easier to just reuse the same test con, given this difference in behavior... It feels a little unncessary. Maybe others have some thoughts.

cpcloud:
@mfatihaktas @chloeh13q Should we close this out for now?

PR #8946, opened 2024-04-12
test: check for ordering with Table.distinct(keep=...)

NickCrews:
Before, we didn't actually verify that the resulting row from each group
was the first or last one. Now we order
beforehand, do the distinct, and then check that the ordering was preserved throughout.

I started poking because I think ibis is giving me wrong results here. Still need to dig in more, but the initial failing tests seem to corroborate that...

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -1144,12 +1144,7 @@ def test_pivot_wider(backend):
     reason="function last(double precision) does not exist, do you mean left or least",
 )
 def test_distinct_on_keep(backend, on, keep):
-    from ibis import _
-
-    t = backend.diamonds.mutate(one=ibis.literal(1)).mutate(
-        idx=ibis.row_number().over(order_by=_.one, rows=(None, 0))
-    )
-
+    t = backend.diamonds.mutate(idx=ibis.row_number())
     expr = t.distinct(on=on, keep=keep).order_by(ibis.asc("idx"))
     result = expr.execute()
     df = t.execute()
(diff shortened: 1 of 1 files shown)


NickCrews:
These CI failures make it look to me like this functionality isn't working at all, because using `first()` like this doesn't actually consider prior orderings at all. At least in duckdb, I think we should be doing something like [SELECT first(amount ORDER BY date ASC) FROM sales;](https://duckdb.org/docs/sql/aggregates) if we want any sort of ordering semantics.

A few possible paths I see:
- A: switch to args to `keep: Literal["arbitrary", "none"] = "arbitrary"`. If users want control over which row is selected, they have to do the agg themselves manually.
- B: Keep the API the same, actually fix the bug. IDK how hard this would be. Probably add a `order_by` to the `ops.First` operation, and then inside the .distinct function would have to transfer the order_by clauses of the parent relation over to this ops.First. Sounds perilous.
- C: switch to args to `keep: Literal["arbitrary", "none"] | <callback> = "arbitrary"`, and add a `order_by` to the `ops.First`. Then users can something like

```python

def get_agg(table: ir.Table, col: str):
    if col == "date":
        return t[col].min()
    if col == "amount":
        return t[col].max()
    return col.first(order_by=t.date)

table.distinct(on=..., keep=get_agg)
```

IDK that seems complicated. Probably should just leave that up to users to do manually.

cpcloud:
The fix for the issue here will have to go in a follow-up. @jcrist has a PR up to address this more thoroughly, but IIRC he hit some snags.

PR #8950, opened 2024-04-12
docs(build): cache notebooks during render

cpcloud:
This PR enables notebook caching via jupyter-cache. Locally I observed between 2-3x faster rendering. Happy to not merge this if people do not think it is enough of an improvement.


cpcloud:
Oh, I think we just updated the same thing in the same place.

cpcloud:
@lostmygithubaccount if you approve, we can merge

lostmygithubaccount (approved):
thanks!

gforsyth (approved):
>2x speedup locally

SHIPIT

PR #8951, opened 2024-04-12
docs(altair): remove to_pandas from altair example

gforsyth:
Noticed in @koaning's livestream that we never updated the Altair example to remove the call to `to_pandas`

Here's the local render:

### DIFF
--- a/docs/how-to/visualization/altair.qmd
+++ b/docs/how-to/visualization/altair.qmd
@@ -23,7 +23,7 @@ in Ibis tables or expressions:
 import altair as alt
 
 chart = (
-    alt.Chart(t.group_by("species").agg(count=ibis._.count()).to_pandas())
+    alt.Chart(t.group_by("species").agg(count=ibis._.count()))
     .mark_bar()
     .encode(
         x="species",


cpcloud (approved):
Nice

koaning:
d0h! feel free to leave a comment on Youtube to notify folks about this.

gforsyth:
Not at all, thanks for bringing it to our attention!

koaning:
Since I have you here, any comments on my sklearn predictions observation? Is there an easy way to add a column with predictions from a scikit-learn model in ibis? My initial impression was "no" but I'll gladly be told otherwise.

gforsyth:
Yes, a few!

For the `range` example (which is obviously _not_ your end goal with sklearn predictions) you can use `ibis.row_number()`

For the larger problem of what we call positional joins, it's tricky because `pandas` and other in-memory formats have an implicit ordering they can take advantage of.

So something slightly gross you can do is:

```python
t.mutate(id=ibis.row_number()).join(ibis.memtable({"id": ..., "predictions": ...}, "id")
```

where you would've had to make sure you assign the row numbers before hand so you have the join key to relink it afterwards.

You should also check out our https://github.com/ibis-project/ibisml where we are thinking about these issues.

And if you're game, come join our Zulip!  We're usually around on there and it works well for semi-synchronous discussions!

https://ibis-project.zulipchat.com/join/o34akeq2pcxoplvgn4ag7wtk/

koaning:
Ah! Hadn't thought of it that way, but this memtable approach makes sense. It's a small bummer that it requires a  (expensive?) join operation but I can see how this is stable across backends.

Final question, I'm currently hosting and planning a bunch of podcast episodes for this new podcast that we're doing over on the probabl side. If you happen to know someone from the ibis project that would make for a solid guest on the podcast ... do let me know!

cpcloud:
👋🏻 Hi @koaning I'd be happy to go on the podcast! Thanks for reaching out. Perhaps let's discuss on Zulip in a DM?

Issue #8479, opened 2024-02-27
refactor(polars): more accurate type mapping for timestamps


jcrist:
> Since the ibis timestamps have a scale we can do more accurate conversions here using `dtype.unit.short`. I assume this was the previous conversion logic from the backend implementation. Since it requires additional testing it could you please create a follow-up ticket so we improve this later?

_Originally posted by @kszucs in https://github.com/ibis-project/ibis/pull/8478#discussion_r1504846304_

PR #8954, opened 2024-04-12
feat(polars): add more accurate type mapping for timestamps

ncclementi:
- Closes https://github.com/ibis-project/ibis/issues/8479

polars datetime types have a fixed time unit (us, ns, or ms). Ibis timestamp types do too (but we also have seocnds as an option). In the case of polars we're currently mapping all timestamps to `ns`, this PR makes things a bit more specific and uses proper units. 

I'm still not sure how to handle the case `Ibis seconds` to maybe `ms in polars` ? Need a bit help on how the conversion of units happens, more specifically where is the value part handled? we would need to do `ibis_seconds*1000` = `polars_ms` at the moment I'm raising, similarly to what we do in durations, but I need to tackle this differently, see all the current failures in CI.

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -612,6 +612,9 @@ class Timestamp(Temporal, Parametric):
         elif unit == TimestampUnit.NANOSECOND:
             scale = 9
         else:
+            # TODO: remove raise path as it's never triggered
+            # Timestamp op has a restriction that only the literal
+            # ints from 0 through 9 can be passed as scale
             raise ValueError(f"Invalid unit {unit}")
         return cls(scale=scale, timezone=timezone, nullable=nullable)
 
@@ -627,6 +630,9 @@ class Timestamp(Temporal, Parametric):
         elif 7 <= self.scale <= 9:
             return TimestampUnit.NANOSECOND
         else:
+            # TODO: remove raise path as it's never triggered
+            # TimestampUnit, which is a (child of) Enum
+            # so it'll raise in the Enum class constructor instead
             raise ValueError(f"Invalid scale {self.scale}")
 
     @property
--- a/ibis/formats/tests/test_polars.py
+++ b/ibis/formats/tests/test_polars.py
@@ -31,9 +31,19 @@ from ibis.formats.polars import PolarsData, PolarsSchema, PolarsType  # noqa: E4
         param(dt.uint64, pl.UInt64, id="uint64"),
         param(dt.float32, pl.Float32, id="float32"),
         param(dt.float64, pl.Float64, id="float64"),
-        param(dt.timestamp, pl.Datetime("ns", time_zone=None), id="timestamp"),
         param(
-            dt.Timestamp("UTC"), pl.Datetime("ns", time_zone="UTC"), id="timestamp_tz"
+            dt.Timestamp(scale=9), pl.Datetime("ns", time_zone=None), id="timestamp_ns"
+        ),
+        param(
+            dt.Timestamp(scale=6), pl.Datetime("us", time_zone=None), id="timestamp_us"
+        ),
+        param(
+            dt.Timestamp(scale=3), pl.Datetime("ms", time_zone=None), id="timestamp_ms"
+        ),
+        param(
+            dt.Timestamp("UTC", scale=9),
+            pl.Datetime("ns", time_zone="UTC"),
+            id="timestamp_tz",
         ),
         param(dt.Interval(unit="ms"), pl.Duration("ms"), id="interval_ms"),
         param(dt.Interval(unit="us"), pl.Duration("us"), id="interval_us"),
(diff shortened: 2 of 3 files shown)


ibis/expr/datatypes/tests/test_core.py
```diff
@@ -518,6 +518,11 @@ def test_timestamp_unit():
         assert dt.Timestamp(scale=scale).unit == TimestampUnit.NANOSECOND
 
 
+def test_timestamp_unit_raise():
```
ncclementi, line 521: I added this because it seemed like this https://github.com/ibis-project/ibis/blob/93552812ee9e8e0e3397bc226cc20c381fcd173b/ibis/expr/datatypes/core.py#L604-L606 wasn't covered. 

But turns out this test is causing a raise in a different place, which is confusing.

ibis/expr/datatypes/tests/test_core.py
```diff
@@ -518,6 +518,11 @@ def test_timestamp_unit():
...
+    with pytest.raises(ValueError, match="Invalid scale"):
```
ncclementi, line 522: Same than comment above

ibis/expr/datatypes/tests/test_core.py
```diff
@@ -531,6 +536,11 @@ def test_timestamp_from_unit():
     )
 
 
+def test_timestamp_from_unit_raise():
```
ncclementi, line 539: Similar comment than above, but in from_unit code. 

https://github.com/ibis-project/ibis/blob/93552812ee9e8e0e3397bc226cc20c381fcd173b/ibis/expr/datatypes/core.py#L604-L605

ibis/expr/datatypes/tests/test_core.py
gforsyth: The `Timestamp` op has a restriction that only the literal ints from 0 through 10 can be passed as `scale` -- so while it isn't technically unit-tested, it's not possible to set a value outside of that range without raising a `SignatureValidationError`

ncclementi: That makes sense, but then why having a code path to "raise" that would never been triggered.

ibis/expr/datatypes/tests/test_core.py
gforsyth: this one is because in `ibis/expr/datatypes/core.py:595`, the `from_unit` class method, the `unit` argument gets passed to `TimestampUnit`, which is a (child of) `Enum` and so the error raises from the Enum class constructor because it disallows values it doesn't know about.

In this case, one of:

```python
@public
class TimestampUnit(TemporalUnit):
    SECOND = "s"
    MILLISECOND = "ms"
    MICROSECOND = "us"
    NANOSECOND = "ns"
```

so it will always raise a `ValueError` and it's never going to hit the `else`

ibis/expr/datatypes/tests/test_core.py
gforsyth: i would wager that the `raise` predates the enforcement of values at the op level

ibis/expr/datatypes/tests/test_core.py
ncclementi: This path also seem to be triggered. Something like `dt.Timestamp.from_unit("foo")`  that I thought would trigger the `ValueError(f"Invalid unit {unit}")` but it raises  because of the `enum` bounding. (I think)?

ibis/formats/tests/test_polars.py
```diff
@@ -60,6 +70,25 @@ def test_to_from_ibis_type(ibis_dtype, polars_type):
...
+        param(dt.Timestamp(), pl.Datetime("ns", time_zone=None), id="timestamp_scn"),
+        param(
+            dt.Timestamp("UTC", scale=0),
+            pl.Datetime("ns", time_zone="UTC"),
+            id="timestamp_tz_sc0",
+        ),
+    ],
+)
+def test_from_ibis_type_seconds(ibis_dtype, polars_type):
+    # we accept seconds as an ibis type, default to ns in polars
```
ncclementi, line 88: @jcrist in the original issue, you suggested maybe defaulting to `"ms"` for when we go from `"s"` in Ibis to polars. I though it made sense  keeping the behavior we had before for that particular case which is `"ns"`. Any objections?

ibis/expr/datatypes/core.py
```diff
@@ -602,6 +602,9 @@ def from_unit(cls, unit, timezone=None, nullable=True):
         elif unit == TimestampUnit.NANOSECOND:
             scale = 9
         else:
+            # TODO: remove raise path as it's never triggered
+            # Timestamp op has a restriction that only the literal
+            # ints from 0 through 10 can be passed as scale
```
cpcloud, line 607: I think it should be 0 to 9, hopefully we're not allowing subnanoseconds 😬

ncclementi: Whoops, nice catch.

ncclementi:
After looking more into this PR, I'm not sure if we are correctly casting the values when mapping the timestamps, base on this code: 

https://github.com/ibis-project/ibis/blob/b48f451de77bd9af832cea10d7ddf8c44097c45a/ibis/backends/polars/compiler.py#L157-L171

That being said the casting "potential issue" is separate from the mapping of the types, so I proposed opening a separate issue to investigate that, and for now get this PR in.

ncclementi:
I did a little more digging and on main (as well as this branch). The casting is broken, see https://github.com/ibis-project/ibis/issues/9091 

I'm not sure what are the next steps here, regarding this PR.

ibis/formats/tests/test_polars.py
cpcloud: It still seems a bit odd to me that asking for seconds I would get back milliseconds. I say we raise in the seconds case as an `UnsupportedOperationError` the same any other missing functionality.

ncclementi: This will break the current behavior where we ask for seconds and return `ns`. Just making sure we are ok with breaking that.

cpcloud:
@ncclementi Anything blocking this PR? I know it's been awhile, but does it make sense to merge this in for 9.2?

ncclementi:
Technically no, as the typing gets fixed in this PR, although the actual casting is kind of broken see https://github.com/ibis-project/ibis/issues/9091. 

Let me rebase, and kick CI again. I'd be ok merging this and we can look into the casting of the values in a followup.

cpcloud (approved):
LGTM, thanks!

PR #8956, opened 2024-04-13
ci(deps): remove dep caching and pipx usage in favor of plain pip

cpcloud:
Remove dependency caching and pipx usage to avoid CI hiccups from mismatching pipx python versions and target test python versions.

### DIFF
--- a/.github/workflows/check-generated-files.yml
+++ b/.github/workflows/check-generated-files.yml
@@ -30,9 +30,6 @@ jobs:
       - name: checkout
         uses: actions/checkout@v4
 
-      - name: install poetry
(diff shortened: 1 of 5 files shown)


cpcloud:
The use of pip install inside of poetry (which was previously installed with a different version of Python) was causing wheels associated with a different version of Python to be installed than the one that was under test. See https://github.com/ibis-project/ibis/actions/runs/8668105648/job/23772495886?pr=8787 for an example.

Note the attempted install of a Python 3.10 `torch` wheel, while the Python version being tested is 3.11.

cpcloud:
Merging, to get the other deps PR (#8787) over the finish line.

PR #8958, opened 2024-04-13
feat(api): add API for unwrapping JSON values into backend-native values

cpcloud:
Adds `str`, `int`, `float`, and `bool` properties to `JSONValue` as well as an `unwrap_as` method for easier programmatic usage and more fine-grained casting.

Unless someone really hates the static property names, I'd prefer to keep them as they are. Open to alternative names for `unwrap_as` though.

In theory this can all be done with casting, but if you look at what's being done in the various backends it's typically a lot more involved than that. Trino in particular requires queries over JSON to be `VARCHAR` inputs, when then have to be cast **back** to its `JSON` type to be able to cast _that_ to the desired output type.

Complicating the cast branching _just_ for the `JSON -> not JSON` case seemed like the wrong tradeoff.

I went with these names to match the `map` and `array` APIs, and to match the short type names we have for the specific types (`str`, `int`, `float`, and `bool`), which exist to match the equivalent Python types.

### DIFF
--- a/ibis/expr/types/json.py
+++ b/ibis/expr/types/json.py
@@ -6,6 +6,8 @@ from typing import TYPE_CHECKING
 
 from public import public
 
+import ibis.common.exceptions as exc
+import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 from ibis.expr.types import Column, Scalar, Value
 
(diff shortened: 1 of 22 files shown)


cpcloud:
@dixdev I've added your example into the docstring for the `JSONValue.string` API.

Would you mind taking this PR a spin?

dixdotdev:
Yep happy to try this out tomorrow. I've reviewed the code from mobile and it seems like a pretty elegant solution, but proof is in the execution!

dixdotdev:
So I've taken a spin through it, and while it does work as intended I think that as a replacement for ``.cast(type)`` it deviates a bit syntactically, particularly when plucking out datatypes from iterators or objects.

I think ideally the solution should allow for a similar syntactic expression so your original naming with ``.unwrap(type)``, or ``.unwrap_as(type)`` might make more sense when considering other functional uses.

ibis/expr/types/json.py
```diff
@@ -98,6 +100,132 @@ def __getitem__(
...
+        ┃ unwrapped_int64 ┃ unwrapped_int32 ┃ original             ┃
+        ┡━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━┩
+        │ int64           │ int32           │ json                 │
+        ├─────────────────┼─────────────────┼──────────────────────┤
+        │               1 │               1 │ 1                    │
+        │            NULL │            NULL │ None                 │
+        │            NULL │            NULL │ NULL                 │
+        │               2 │               2 │ 2                    │
+        └─────────────────┴─────────────────┴──────────────────────┘
+        >>> t.select(unwrapped=t.jfloat.unwrap_as("float32"), original=t.jfloat)
```
dixdotdev, line 172: Shouldn't this be``"float64"``? I don't see an unwrap method for ``"float32"``, but the return below is also ``"float32"`` which seems interesting?

ibis/expr/types/json.py
```diff
@@ -98,6 +100,132 @@ def __getitem__(
...
+        ┃ jstring              ┃ jbool                ┃ jint                 ┃ … ┃
+        ┡━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━╇━━━┩
+        │ json                 │ json                 │ json                 │ … │
+        ├──────────────────────┼──────────────────────┼──────────────────────┼───┤
+        │ 'a'                  │ True                 │ 1                    │ … │
+        │ ''                   │ False                │ None                 │ … │
+        │ NULL                 │ None                 │ NULL                 │ … │
+        │ None                 │ NULL                 │ 2                    │ … │
+        └──────────────────────┴──────────────────────┴──────────────────────┴───┘
+        >>> t.select(unwrapped=t.jstring.unwrap_as(str), original=t.jstring)
```
dixdotdev, line 135: Should ``.unwrap_as(str)`` in this actually be ``.unwrap_as("string")``, or are both valid as it's a dtype?

dixdotdev (commented):
I've tried it out for my implementations and it seems to work as expected with the ``.unwrap_as()`` as we discussed- just a couple of queries in the docstrings for clarity.

ibis/expr/types/json.py
cpcloud: Both are valid. The `dt.dtype` call will normalize inputs.

ibis/expr/types/json.py
cpcloud: You can give as specific a type as you want and Ibis will try to cast the result to that type.

The `float32` here is meant to show that you can use a more specific floating point type as well as the more-typical `float`/`float64` (`float` is an alias for `float64`).

I'll add some prose to the docstrings.

gforsyth:
new unwrap tests are passing on clouds:

```sh
🐚 pytest -m "bigquery or snowflake" ibis/backends/tests/test_json.py -k "unwrap"
================================ test session starts ================================
platform linux -- Python 3.12.2, pytest-8.1.1, pluggy-1.4.0
Using --randomly-seed=2966770701
rootdir: /home/gil/github.com/ibis-project/ibis
configfile: pyproject.toml
plugins: xdist-3.5.0, hypothesis-6.100.1, snapshot-0.9.0, randomly-3.15.0, mock-3.14.0, benchmark-4.0.0, cov-5.0.0, repeat-0.9.3, clarity-1.0.1, pytest_httpserver-1.0.10
collected 240 items / 224 deselected / 16 selected                                  

ibis/backends/tests/test_json.py ................                             [100%]

=================== 16 passed, 224 deselected in 80.76s (0:01:20) ===================
```

gforsyth (approved):
This looks good to me!  I suppose this will show Risingwave as supporting these ops in the support matrix, but we can add explicit `NotImplemented`s in a followup.

PR #8965, opened 2024-04-15
ci(snowflake): install python with specific version constraint marker

cpcloud:
Fixes the snowpark install in CI.


cpcloud:
Merging to try to get CI green again.

Issue #8921, opened 2024-04-09
bug: inlining expressions leads to wrong results for non-pure functions


NickCrews:
### What happened?

See this:

```python
import ibis
import random

ibis.options.interactive = True


@ibis.udf.scalar.python(side_effects=True)
def my_random(x: float) -> float:
    return random.random()


t = ibis.memtable({"x": [1, 2, 3, 4, 5]})
# this happens with both the builtin random, and UDFs:
# t = t.mutate(y=ibis.random())
t = t.mutate(y=my_random(t.x))
t = t.mutate(size=(t.y > 0.5).ifelse("big", "small"))
t
# ┏━━━━━━━┳━━━━━━━━━━┳━━━━━━━━┓
# ┃ x     ┃ y        ┃ size   ┃
# ┡━━━━━━━╇━━━━━━━━━━╇━━━━━━━━┩
# │ int64 │ float64  │ string │
# ├───────┼──────────┼────────┤
# │     1 │ 0.208385 │ big    │
# │     2 │ 0.566183 │ big    │
# │     3 │ 0.836448 │ small  │
# │     4 │ 0.530235 │ small  │
# │     5 │ 0.526687 │ small  │
# └───────┴──────────┴────────┘
```

Notice that the `size` column is sometimes incorrect. This is because size is getting calculated from a totally uncorrelated `RANDOM()` in the generated SQL. `ibis.to_sql(t)` shows:

```sql
SELECT
  "t0"."x",
  RANDOM() AS "y",
  CASE WHEN RANDOM() > CAST(0.5 AS DOUBLE) THEN 'big' ELSE 'small' END AS "size"
FROM "ibis_pandas_memtable_m5jukuqv2rcrhkpg2vsdu6k5fe" AS "t0"
```

For this to work as expected, there can only be a single `RANDOM()` in the generated SQL. I think that the agressive inlining that we do to merge selects/mutates together is over-aggressive. After a short talk with @cpcloud, this inlining is only correct when the entire expression is pure. Either we need to somehow know when an expression is pure (most are, except for (some) UDFs, random(), maybe now(), and maybe others I'm not thinking of?), or just be conservative and only do this inlining when the subsequent selects/mutates are just reselecting/renaming columns with no other changes.

### What version of ibis are you using?

main

### What backend(s) are you using, if any?

I'm not sure if this is present on all backends, but I think so.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kszucs:
[...]

PR #8967, opened 2024-04-15
fix(ir): make impure ibis.random() and ibis.uuid() functions return unique node instances

kszucs:
Resolves https://github.com/ibis-project/ibis/issues/8921


ibis/backends/bigquery/compiler.py
```diff
@@ -698,3 +696,6 @@ def visit_CountDistinct(self, op, *, arg, where):
         if where is not None:
             arg = self.if_(where, arg, NULL)
         return self.f.count(sge.Distinct(expressions=[arg]))
+
+    def visit_RandomUUID(self, op, **kwargs):
```
cpcloud, line 700: Why do you need to split these out and handle `kwargs`? The existing `SIMPLE_OPS` loop should handle the case of unused arguments.

cpcloud:
I think this is a bug fix, since it has to do with output correctness.

ibis/backends/bigquery/compiler.py
kszucs: It actually renders the arguments.

ibis/expr/operations/generic.py
```diff
@@ -183,6 +183,17 @@ class Constant(Scalar, Singleton):
...
+    uid: Optional[int] = None
+
+    def __init__(self, uid, **kwargs):
+        if uid is None:
+            uid = next(self._counter)
+        super().__init__(uid=uid, **kwargs)
+
+
 @public
 class TimestampNow(Constant):
```
cpcloud, line 198: This also needs to subclass from `Impure` along with `DateNow`

cpcloud (changes requested):

kszucs:
> There's also `now` and `today` and probably others.

I am planning to address that in a follow-up.

cpcloud:
Any reason why those can't be done in this PR?

jcrist:
Are `now` and `today` impure in the same sense? Implemented using `CURRENT_TIMESTAMP` these will return the same value for every instance of the expression when the query executes (even though they may return a different value for two subsequent executions). In contrast `random()` and `uuid()` will generally return a new value for each call in an expression (some backends like clickhouse handle this a bit differently).

cpcloud:
Ah, good point. I hadn't considered that `select now() = now()` whereas `select random() <> random()`.

Do we know whether that's true for all of our backends?

cpcloud:
Anyway, that's sufficiently different that I think it can be done in a follow up.

jcrist:
> Do we know whether that's true for all of our backends?

I wouldn't be surprised if it's broken for pandas/dask/polars, but semantically I think it _should_ be true for all backends.

ibis/backends/bigquery/compiler.py
cpcloud: Ah, right. Extra kwargs get rendered in the `SIMPLE_OPS` loop.

Can you change the `**kwargs` to be `**_` to indicate that they are not used?

kszucs:
We still have a problem which is definitely solvable but would expand the scope of this PR heavily:


```py
# semantically this should produce two different columns
t.select(x=ibis.random(), y=ibis.random())`
```

```py
# semantically this should produce identical columns
rand = ibis.random()
t.select(x=rand, y=rand)
```

Currently we compile both as `rand() AS x, rand() AS y` which behaves differently on the backends. After this PR the two random values are equal from ibis' perspective, but ideally we should ensure that they produce the same columns on all backends. 
The solution could be to inject an additional subselect:

```sql
SELECT *, randcol AS x, randcol AS y FROM (SELECT *, rand() as randcol FROM ...)  # where randcol name would be generated
```

cpcloud:
Can that be done in a follow up?

kszucs:
> Can that be done in a follow up?

Definitely! That is slightly harder and we need to work around backend specifics.

Yes, please regenerate the snapshots for the failing backends!

cpcloud:
Clouds are passing:

```
…/ibis on  impure is 📦 v8.0.0 via 🐍 v3.11.8 via ❄️   impure (ibis-3.11.8-env)
❯ pytest -m 'bigquery or snowflake' -n auto --dist loadgroup --snapshot-update
==================================================================================== test session starts ====================================================================================
platform linux -- Python 3.11.8, pytest-8.1.1, pluggy-1.4.0
Using --randomly-seed=168840051
rootdir: /home/cloud/src/ibis
configfile: pyproject.toml
plugins: xdist-3.5.0, pytest_httpserver-1.0.10, cov-5.0.0, benchmark-4.0.0, randomly-3.15.0, hypothesis-6.100.1, repeat-0.9.3, snapshot-0.9.0, mock-3.14.0, clarity-1.0.1
16 workers [3509 items]
.....................................x......................x....................x.x.....x.....x............x.....x..........xx..................x...xxx............................x [  5%]
.x.........x..x.........................xx.xx.x.xxx...x.xxx.xxx..x....x.....x.....x...x........x.xx.x..x...........x.x......x.......x.x...x..x..x.xx..................x.............. [ 10%]
.x..x...xx....x.x...x....x.............x...................x....x.xx..x........x.x.............x......x..........x...............x.............x......x.............x.x..x........... [ 15%]
....xx.....x........x.xx...x......x..........x........x...........x..x........x...............x.....x...........................xx.......s....x.x..x........................x.x.....x [ 20%]
..x..xx...x..xx.x.xx..x.x..xxx...xx.xxxxx.x.x...xxxxxx....xxxx..xxxx.xxx.x...xxxx..x..x.xx..x...x.x..x..x..xx..xx..x..........x.......x...x......................................x... [ 25%]
.x......x......x...x.....x...xxx.xx.x...xx.x..x.x....x..x..x...........x.........xx.x.......x.....xx............xx..x........x...x......x....x......xxx...x....x.xx...............x.. [ 30%]
.x.......x.........x.............................................s.....s..s......................s...x..s..s...s....x...........................x.......x.........s.................. [ 36%]
................x.........x..x..........x................x..x.......s..................x.........x...................sssssssssssssssssss.sss...x...........x.......x................. [ 41%]
.....................x.....xxx..........x......x....x......x.................x..........x.............................................................x.............................. [ 46%]
........x........x.....x........................................x.x...x.x.xx.x.x..x...xx...x.x...xxx......xxx........x..x..x......x..x...x..............x....x.xx.................... [ 51%]
.........x.....x...........x...x...x.............x...x.x..xx...x........xx.x..x....x.x......x..x.x..x.......x...............x.....x.x..........x.x.....xx..............x............. [ 56%]
.............x...x..................x..x......x...............................x.......x....x.....x.....x.....................x.x...x.......x....x...........s...................xxxx. [ 61%]
.xxx.xx.....xxxx.x.xx.xxxxx....x.xxxx.xxxxxxxxxxxxxxxx.xxxxxxxxxxxx.xxxxxxxxxxxxx.x.xxxxxxxxx.xxxxxxxxxxxxxx.xxxxx....xxxx..x.xx..x.x...x.x.x..x..............x...........x..x.x..... [ 67%]
........x.....x....x.........x.x..xx.xxxx.x.x.x...x..xx..x..x.xx.........x..x....x......x.x..x........x.....x.............x.x.xx.....x...................xx.........x.........x...... [ 72%]
..x.......xxx..........x......xx..............x................s.......................s.................sx..x..sx..x.x.s.sxx.sxxx.x.xx....x.x....x..x.xxx...x....xx..x..x........xx.. [ 77%]
....x..x.s....x........s...s..........s..x.......x...x...........x........x.xxx..........xx.....x........x....x.....x.....x..x....x........x......x.x...x...x........x.x..x..xx...... [ 82%]
......x............x..x...xx..........x......x.......s............x................................................................................x................................. [ 87%]
.....................................................................x..x.xx......................................................................................................... [ 92%]
...........................................................................................................................................................x......................... [ 98%]
..................................................x..................                                                                                                                 [100%]
================================================================= 2921 passed, 45 skipped, 543 xfailed in 356.95s (0:05:56) =================================================================
```

cpcloud (approved):

PR #8971, opened 2024-04-15
test(snowflake-snowpark): testing running ibis code in a snowpark stored procedure

### LINKED ISSUE #5877: feat(snowflake): run ibis code inside of stored procedure

cpcloud:
Add a test demonstrating running Ibis inside of a Snowpark stored procedure.

Closes #5877.

### DIFF
--- a/ibis/backends/snowflake/__init__.py
+++ b/ibis/backends/snowflake/__init__.py
@@ -222,8 +222,8 @@ $$ {defn["source"]} $$"""
             create_object_udfs=create_object_udfs,
         )
 
-    def _setup_session(self, *, con, session_parameters, create_object_udfs: bool):
-        self.con = con
+    def _setup_session(self, *, session_parameters, create_object_udfs: bool):
+        con = self.con
 
         # enable multiple SQL statements by default
         session_parameters.setdefault("MULTI_STATEMENT_COUNT", 0)
@@ -318,12 +318,15 @@ $$ {defn["source"]} $$"""
         │ ansonca01 │    16 │
         └───────────┴───────┘
         """
+        import snowflake.connector
+
         backend = cls(_from_snowpark=True)
-        backend._setup_session(
-            con=session._conn._conn,
-            session_parameters={},
-            create_object_udfs=create_object_udfs,
-        )
+        backend.con = session._conn._conn
+        with contextlib.suppress(snowflake.connector.errors.ProgrammingError):
+            # stored procs on snowflake don't allow session mutation it seems
+            backend._setup_session(
+                session_parameters={}, create_object_udfs=create_object_udfs
+            )
         return backend
 
     def reconnect(self) -> None:
--- a/ibis/backends/snowflake/tests/test_udf.py
+++ b/ibis/backends/snowflake/tests/test_udf.py
@@ -1,11 +1,17 @@
 from __future__ import annotations
 
+import os
+import shutil
+import tempfile
+
 import pandas.testing as tm
 import pytest
 from pytest import param
 
+import ibis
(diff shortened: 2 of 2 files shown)


ibis/backends/snowflake/__init__.py
```diff
@@ -317,12 +317,15 @@ def from_snowpark(cls, session, *, create_object_udfs: bool = True) -> Backend:
...
+        import snowflake.connector
+
         backend = cls(_from_snowpark=True)
-        backend._setup_session(
-            con=session._conn._conn,
-            session_parameters={},
-            create_object_udfs=create_object_udfs,
-        )
+        backend.con = session._conn._conn
+        with contextlib.suppress(snowflake.connector.errors.ProgrammingError):
```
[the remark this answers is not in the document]
cpcloud: Looks like this is related to caller's rights versus owner's rights (https://community.snowflake.com/s/question/0D5VI000007waK90AI/alter-session-not-works-inside-snowpark-sessionsql).

We have to handle both cases since it's not within Ibis's control, so this code has to stick around for now.

IndexSeek:
I was looking through the owner's versus caller's rights, I think you are right about having to leave that portion for now.

ibis/backends/snowflake/tests/test_udf.py
```diff
@@ -150,3 +154,108 @@ def cases(value, mapping):
...
+    # so we modify the utime of the _copied_ code since snowflake has
+    # some annoying checks for zipping files that do not allow files
+    # older than 1980 ¯\_(ツ)_/¯
+    os.utime(pkgpath, None)
+    for root, dirs, files in os.walk(pkgpath):
+        for path in dirs + files:
+            os.utime(os.path.join(root, path), None)
+
+
+def add_packages(d, session):
```
IndexSeek, line 170: This is a super clever solution, much easier than zipping and staging from my previous attempts. Very nice!

ibis/backends/snowflake/tests/test_udf.py
```diff
@@ -150,3 +154,108 @@ def cases(value, mapping):
...
+        packages=[
+            "snowflake-snowpark-python",
+            "toolz",
+            "atpublic",
+            "bidict",
+            "pyarrow",
+            "pandas",
+            "numpy",
+        ],
+        replace=True,
```
IndexSeek, line 256: I was able to get this test passing registering the stored procedure with owner's rights. I did recieve an internal error when trying to register as caller's rights, but that allowed for the session parameters to be adjusted as I suspected but caused an errored out the stored procedure itself.  

Could you try adding the `execute_as="caller",` arg on the stored procedure registration to confirm that this error is not caused by an account setting on my side?

cpcloud: I get this error when I pass `execute_as="caller"`:

```
E       snowflake.connector.errors.ProgrammingError: 000603 (XX000): SQL execution internal error:
E       Processing aborted due to error 300010:3474929812; incident 3000484.
```

kszucs (approved):

PR #8972, opened 2024-04-15
fix(snowflake): initialize the parent class on construction

cpcloud:
Forgot to include this change in the previous related PR. That will show me to not run the cloud tests in CI.


cpcloud:
The Flink backend test is a separate issue. I'll put up a PR for that next.

PR #8985, opened 2024-04-17
fix(duckdb): workaround remaining null map issues

### LINKED ISSUE #8632: feat: add workarounds for duckdb MAP bugs

cpcloud:
Closes #8632.

Note that of the backends that support nullable maps, DuckDB is the only one whose `NULL` handling semantics differ.

Other backends that support nullable maps and are implemented in Ibis and tested in CI:

- Postgres (only maps of string -> string via the postgres `hstore` extension)
- PySpark
- Snowflake
- Trino

BREAKING CHANGE: Calling the `get` or `contains` method on `NULL` map values now returns `NULL`. Use `coalesce(map.get(...), default)` or `coalesce(map.contains(), False)` to get the previous behavior.

### DIFF
--- a/ibis/backends/duckdb/compiler.py
+++ b/ibis/backends/duckdb/compiler.py
@@ -43,10 +43,7 @@ class DuckDBCompiler(SQLGlotCompiler):
         ops.Hash: "hash",
         ops.IntegerRange: "range",
         ops.TimestampRange: "range",
-        ops.MapKeys: "map_keys",
         ops.MapLength: "cardinality",
-        ops.MapMerge: "map_concat",
-        ops.MapValues: "map_values",
         ops.Mode: "mode",
         ops.TimeFromHMS: "make_time",
         ops.TypeOf: "typeof",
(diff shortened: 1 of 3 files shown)


cpcloud:
Seems like the chocolatey package server is having a moment.

cpcloud:
xfailed the single failing flink test

PR #8992, opened 2024-04-18
feat(ir): more flexible dereferencing support for join right hand side

kszucs:
Enables to use fields from parent tables of the join right hand side instead of enforcing to use the same exact table:

```py
t1 = ibis.table(name="t1", schema={"a": "int64", "b": "string"})
t2 = ibis.table(name="t2", schema={"c": "int64", "d": "string"})

t3 = t2.mutate(e=t2.c + 1)
joined = t1.join(t3, [t1.a == t2.c])  # here we use t2.c instead of t3.c
```

Identify ambiguous cases and raise an error, like the following case:

```py
t.join(t, [t.a == t.a])
```

depends on:
- #9043 
- #9041 

fixes https://github.com/ibis-project/ibis/issues/8581


ibis/expr/format.py
```diff
@@ -151,7 +152,8 @@ def get_defining_frame(expr):
     """Locate the outermost frame where `expr` is defined."""
     for frame_info in inspect.stack()[::-1]:
         for var in frame_info.frame.f_locals.values():
-            if isinstance(var, ir.Expr) and expr.equals(var):
+            # if isinstance(var, ir.Expr) and expr.equals(var):
```
kszucs, line 155: Going to apply this change in a different PR.

ibis/expr/rewrites.py
```diff
@@ -23,6 +28,55 @@
...
+            value = value.rel.values.get(value.name)
+            distance += 1
+        if value is not None:
+            yield value, distance
+
+    def dereference(self, value):
+        ambigs = value.find(lambda x: x in self.ambigs, filter=ops.Value)
+        if ambigs:
+            raise IbisInputError(
+                f"Ambiguous field reference {ambigs!r} in expression {value!r}"
```
kszucs, line 75: The repr must be improved.

ibis/common/graph.py
```diff
@@ -66,8 +66,9 @@ def _flatten_collections(node: Any) -> Iterator[N]:
             yield item
         elif isinstance(item, (tuple, list)):
             yield from _flatten_collections(item)
-        elif isinstance(item, (dict, frozendict)):
-            yield from _flatten_collections(item.values())
+        elif isinstance(item, dict):
+            items = itertools.chain.from_iterable(item.items())
```
kszucs, line 70: Going to pull this out to a separate PR.

ibis/expr/types/joins.py
```diff
@@ -112,47 +112,18 @@ def disambiguate_fields(
...
-def dereference_value(pred, deref_left, deref_right):
-    deref_both = {**deref_left, **deref_right}
-    if isinstance(pred, ops.Comparison) and pred.left.relations == pred.right.relations:
-        left, right = dereference_sides(pred.left, pred.right, deref_left, deref_right)
-        return pred.copy(left=left, right=right)
-    else:
-        return pred.replace(deref_both, filter=ops.Value)
+def dereference_value(pred, deref_left, deref_right, deref_both):
+    return deref_both.dereference(pred)
+    # if isinstance(pred, ops.Comparison) and pred.left.relations == pred.right.relations:
```
kszucs, line 117: This code bit is responsible for allowing `a.join(a, [a.col == a.col])`

kszucs: Removing it once I have confirmation about that we are disallowing it.

cpcloud: This seems very ambiguous to me and _somewhat_ unlikely. +1 for removing it

cc @jcrist @gforsyth for opinions

ibis/common/graph.py
kszucs: https://github.com/ibis-project/ibis/pull/9041

ibis/expr/types/joins.py
gforsyth: I am +1 on removing it

ibis/expr/format.py
kszucs: https://github.com/ibis-project/ibis/pull/9043

ibis/expr/types/joins.py
jcrist: Also +1. Rip it out.

ibis/expr/rewrites.py
kszucs: Deferring it to a follow-up.

ibis/backends/impala/tests/snapshots/test_sql/test_nested_join_base/out.sql
```diff
@@ -7,7 +7,7 @@ WITH `t1` AS (
     1
 )
 SELECT
-  `t5`.`uuid`,
+  `t3`.`uuid`,
```
cpcloud, line 10: This looks incorrect to me. Check out the expression:


```python
    t = ibis.table(dict(uuid="string", ts="timestamp"), name="t")
    counts = t.group_by("uuid").size()
    max_counts = counts.group_by("uuid").aggregate(max_count=lambda x: x[1].max())
    result = max_counts.left_join(counts, "uuid").select(counts)
```

The `uuid` column in the final `.select` call is coming from the **right** side of the join, so it should remain `t5.uuid` I think.

cpcloud: The right side `Field` cannot dereference to the left side field because this is a left outer join, so the join key values for a given row may be different.

kszucs: The following is the repr (after setting `ibis.options.repr.show_variables = True`):

```
t := UnboundTable: t
  uuid string
  ts   timestamp

counts := Aggregate[t]
  groups:
    uuid: t.uuid
  metrics:
    CountStar(t): CountStar(t)

max_counts := Aggregate[counts]
  groups:
    uuid: counts.uuid
  metrics:
    max_count: Max(counts['CountStar(t)'])

JoinChain[max_counts]
  JoinLink[left, counts]
    max_counts.uuid == counts.uuid
  values:
    uuid:         counts.uuid
    CountStar(t): counts['CountStar(t)']
```

Isn't this the expected expr?

cpcloud: You're right! I misread. Resolved.

ibis/expr/rewrites.py
```diff
@@ -23,6 +28,55 @@
 name = var("name")
 
 
+class DerefMap(Concrete, Traversable):
```
cpcloud, line 31: Can you add a docstring? This is a pretty important part of the library, and it would help folks understand how it fits into everything if there were some docs.

ibis/expr/tests/test_newrels.py
```diff
@@ -1233,9 +1299,11 @@ def test_self_join_with_intermediate_selection():
...
         assert join.op() == expected
 
-    aa = a.join(a, [a.a == a.a])["a", "b_right"]
-    aaa = aa.join(a, [aa.a == a.a])
-    with join_tables(a, a, a) as (r1, r2, r3):
+    a_ = a.view()
+    a__ = a.view()
+    aa = a.join(a_, [a.a == a_.a])["a", "b_right"]
+    aaa = aa.join(a__, [aa.a == a__.a])
+    with join_tables(aaa) as (r1, r2, r3):
```
cpcloud, line 1306: Hm, let's come up with some more unique variable names here 😂

ibis/expr/tests/test_newrels.py
```diff
@@ -1173,10 +1205,28 @@ def test_join_chain_gets_reused_and_continued_after_a_select():
 def test_self_join_extensive():
     a = ibis.table(name="a", schema={"a": "int64", "b": "string"})
 
-    aa = a.join(a, [a.a == a.a])
+    with pytest.raises(IbisInputError, match="Ambiguous field reference"):
```
cpcloud, line 1208: Nice, glad we can detect and raise for this case!

cpcloud (approved):
LGTM mostly, just a few small requests.

Would really like to see a docstring for `DerefMap` but I won't block on that.

ibis/expr/tests/test_newrels.py
kszucs: I can do a1 and a2 :)

ibis/common/typing.py
```diff
@@ -259,3 +260,21 @@ class Coercible(Abstract):
...
     def __coerce__(cls, value: Any, **kwargs: Any) -> Self: ...
+
+
+def get_defining_frame(obj):
+    """Locate the outermost frame where `obj` is defined."""
+    for frame_info in inspect.stack()[::-1]:
+        for var in frame_info.frame.f_locals.values():
+            if obj is var:
+                return frame_info.frame
+    raise ValueError(f"No defining frame found for {obj}")
```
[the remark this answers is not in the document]
kszucs: I don't think it is important for now since it is an opt-in feature.

ibis/expr/tests/test_newrels.py
cpcloud: SGTM, thanks!

PR #9008, opened 2024-04-18
feat(api): expose common types in the top-level `ibis` namespace

### LINKED ISSUE #8717: feat: expose common public types in top-level namespace
When writing code consuming `ibis` objects as part of its api, I often want to write functions that take in ibis tables/columns/scalars/values. From a UX perspective it would be nice to export a few common types in the top-level namespace so I can refer to these types in both annotations and docs as `ibis.Table` instead of `ibis.expr.types.Table`.

For my use case `ibis.Table` and `ibis.Column` alone would be sufficient. For completeness exporting `ibis.Scalar`, `ibis.Value`, and maybe `ibis.Deferred` may also be nice.

jcrist:
Adds `Table`, `Column`, `Scalar`, `Value`, and `Deferred` to the top-level `ibis` namespace. These are mostly useful for downstream users consuming the ibis api (either type-annotating their code, or using `isinstance` checks).

Fixes #8717.


ibis/expr/api.py
```diff
@@ -51,6 +54,15 @@
     from ibis.expr.schema import SchemaLike
 
 __all__ = (
```
jcrist, line 56: I ran `:sort` on this list in vim, which is why there's a bunch of extra movement here.

Turns out we were already exporting `Expr` and `Schema` types (but no others), but these were buried in the `__all__` list before :).

cpcloud (approved):

PR #9009, opened 2024-04-18
docs(tutorial): deploy a jupyterlite repl on the website

cpcloud:
Extremely experimental. Let's see how a PR deployment goes.


docs/index.qmd
```diff
@@ -40,6 +41,40 @@ about:
...
+
+```{python}
+#| echo: false
+#| output: asis
+from urllib.parse import urlencode
+
+lines = """
+%pip install numpy pandas tzdata
+import pyodide_js
+await pyodide_js.loadPackage("https://storage.googleapis.com/ibis-wasm-wheels/pyarrow-16.0.0.dev2661%2Bg9bddb87fd-cp311-cp311-emscripten_3_1_46_wasm32.whl")
```
cpcloud, line 56: This can ideally be removed when there's a pyodide distribution that ships with pyarrow.

docs/ibis-wasm-wheels_cors_config.json
cpcloud, line 1: This doesn't necessarily need to be here, I just wanted to write down the settings that are applied to the `gs://ibis-wasm-wheels` bucket for posterity.

docs/index.qmd
```diff
@@ -40,6 +41,40 @@ about:
...
+%clear
+%pip install 'ibis-framework[duckdb]'
+from ibis.interactive import *
+penguins = ibis.read_csv("penguins.csv")
+penguins.head()
+"""
+
+params = [
+    ("toolbar", "1"),
+    ("theme", "JupyterLab Night"),
```
cpcloud, line 67: It'd be nice to have the theme switch with the quarto toggle. I'm not sure how to do that. I think the procedure would be something like:

1. find the button element
2. attach an onclick listener to that button that sets the theme

docs/index.qmd
```diff
@@ -40,6 +41,40 @@ about:
 ::: {#about}
 :::
 
+
+## Try it out!
```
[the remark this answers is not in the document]
lostmygithubaccount: yeah roughly ~10% of traffic is on iOS, so I don't think we can have the homepage crash there. perhaps under the "Getting started" tab, as like "(Experimental) Try Ibis in the browser" and a warning at the top that iOS is likely to crash

docs/index.qmd
lostmygithubaccount: this is also an issue for general plotting, there was some prior art there but it looked semi-complicated when I checked

docs/index.qmd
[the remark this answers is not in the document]
lostmygithubaccount: and we can work this better into getting started content/on the home page going forward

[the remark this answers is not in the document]
cpcloud: Moved to the beginning of the `getting_started` tutorial.

Issue #9011, opened 2024-04-18
bug: devcontainer is missing GDAL


anentropic:
### What happened?

I checked out the repo and opened it in VS Code. It found a `.devcontainer` config and prompted me to open in a container... I thought this might be easier than messing around with conda etc

But it doesn't work. The `updateContent.sh` script fails trying to pip install dependencies with an error:

```
Failed to get options via gdal-config: [Errno 2] No such file or directory: 'gdal-config'
      CRITICAL:root:A GDAL API version must be specified. Provide a path to gdal-config using a GDAL_CONFIG environment variable or use a GDAL_VERSION environment variable.
```
Presumably gdal is not installed in the `mcr.microsoft.com/vscode/devcontainers/python:3.11` base image

By adding `"remoteUser": "vscode"` to the devcontainer.json I was able to `sudo apt-get install libgdal-dev`

This got further but hit another problem:
```
Defaulting to user installation because normal site-packages is not writeable
Obtaining file:///workspaces/ibis
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing editable metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [36 lines of output]
      Traceback (most recent call last):
        File "/usr/local/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/usr/local/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/usr/local/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 181, in prepare_metadata_for_build_editable
          return hook(metadata_directory, config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-qu9btf27/overlay/lib/python3.11/site-packages/poetry/core/masonry/api.py", line 42, in prepare_metadata_for_build_wheel
          poetry = Factory().create_poetry(Path(".").resolve(), with_groups=False)
[...]

PR #9012, opened 2024-04-18
chore(container): add gdal to codespace

gforsyth:
xref #9011 

dunno if it's this simple or not, but it might be

### DIFF
--- a/.devcontainer/updateContent.sh
+++ b/.devcontainer/updateContent.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+apt update
+apt install libgdal-dev
+
 # install ibis
 python3 -m pip install ipython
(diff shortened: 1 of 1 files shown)


.devcontainer/updateContent.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+apt update
+apt install libgdal-dev
```
cpcloud, line 4: This prevents mac users from using this codespace, barring some magic I am unaware of that would make this work.

gforsyth: I don't understand codespaces, I thought it was running in a container.

I guess we can use `micromamba` to build a cross-platform thing, but I'll leave that alone for now

cpcloud: I think it is and I am wrong. Feel free to reopen if you want.

PR #9014, opened 2024-04-18
test: add test for impure function correlation behavior

NickCrews:
Related to https://github.com/ibis-project/ibis/issues/8921,
trying to write down exactly what the expected behavior is.

I figure we can use this PR to hash out exaclty what we want the semantics to be, and then the other discussions might be easier because our goal is written down precisely somewhere. Please let me know if you agree or disagree with this behavior, or if there are other tests we should add.

Need to fix the UDF test case in a followup. Also wasn't sure where to put these tests, I put them in their own little file but if you point me elsewhere Iwill move them.

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -1057,6 +1057,7 @@ def test_between(backend, alltypes, df):
     backend.assert_series_equal(result, expected)
 
 
+@pytest.mark.notyet(["flink"], reason="timestamp subtraction doesn't work")
 def test_interactive(alltypes, monkeypatch):
     monkeypatch.setattr(ibis.options, "interactive", True)
 
--- a/ibis/backends/tests/test_impure.py
+++ b/ibis/backends/tests/test_impure.py
@@ -11,6 +11,9 @@ from ibis.backends.tests.errors import Py4JJavaError
 
 tm = pytest.importorskip("pandas.testing")
 
+# Concurrent execution of CREATE OR REPLACE FUNCTION in postgres fails
+# This ensures that all tests in this module run in the same process as
+# long as --dist=loadgroup is passed, which it is.
 pytestmark = pytest.mark.xdist_group("impure")
 
 no_randoms = [
--- a/ibis/backends/tests/test_impure.py
+++ b/ibis/backends/tests/test_impure.py
@@ -2,27 +2,20 @@ from __future__ import annotations
 
 import sys
 
-import pandas.testing as tm
 import pytest
 
 import ibis
 import ibis.common.exceptions as com
 from ibis import _
-from ibis.backends.tests.errors import (
-    PsycoPg2InternalError,
-    Py4JJavaError,
-    PyDruidProgrammingError,
-)
+from ibis.backends.tests.errors import Py4JJavaError
+
+tm = pytest.importorskip("pandas.testing")
+
+pytestmark = pytest.mark.xdist_group("impure")
 
 no_randoms = [
     pytest.mark.notimpl(
-        ["dask", "pandas", "polars"], raises=com.OperationNotDefinedError
-    ),
-    pytest.mark.notimpl("druid", raises=PyDruidProgrammingError),
-    pytest.mark.notyet(
-        "risingwave",
-        raises=PsycoPg2InternalError,
-        reason="function random() does not exist",
+        ["polars", "druid", "risingwave"], raises=com.OperationNotDefinedError
     ),
 ]
 
@@ -32,19 +25,16 @@ no_udfs = [
         [
             "bigquery",
             "clickhouse",
-            "dask",
             "druid",
             "exasol",
             "impala",
             "mssql",
             "mysql",
             "oracle",
-            "pandas",
             "trino",
             "risingwave",
         ]
     ),
-    pytest.mark.notimpl("pyspark", reason="only supports pandas UDFs"),
     pytest.mark.notyet(
         "flink",
         condition=sys.version_info >= (3, 11),
@@ -55,16 +45,7 @@ no_udfs = [
 
 no_uuids = [
     pytest.mark.notimpl(
-        [
-            "druid",
-            "exasol",
-            "oracle",
-            "polars",
-            "pyspark",
-            "risingwave",
-            "pandas",
-            "dask",
-        ],
+        ["druid", "exasol", "oracle", "polars", "pyspark", "risingwave"],
         raises=com.OperationNotDefinedError,
     ),
     pytest.mark.notyet("mssql", reason="Unrelated bug: Incorrect syntax near '('"),
@@ -82,11 +63,7 @@ def my_random(x: float) -> float:
 mark_impures = pytest.mark.parametrize(
     "impure",
     [
-        pytest.param(
-            lambda _: ibis.random(),
-            marks=no_randoms,
-            id="random",
-        ),
+        pytest.param(lambda _: ibis.random(), marks=no_randoms, id="random"),
         pytest.param(
             lambda _: ibis.uuid().cast(str).contains("a").ifelse(1, 0),
             marks=[
@@ -107,6 +84,7 @@ mark_impures = pytest.mark.parametrize(
 )
 
 
+# You can work around this by .cache()ing the table.
 @pytest.mark.notyet("sqlite", reason="instances are uncorrelated")
 @mark_impures
 def test_impure_correlated(alltypes, impure):
@@ -120,14 +98,12 @@ def test_impure_correlated(alltypes, impure):
     #   t AS (SELECT random() AS common)
     # SELECT common as x, common as y FROM t
     # Then both x and y should have the same value.
-    df = (
-        alltypes.select(common=impure(alltypes))
-        .select(x=_.common, y=_.common)
-        .execute()
-    )
+    expr = alltypes.select(common=impure(alltypes)).select(x=_.common, y=_.common)
+    df = expr.execute()
     tm.assert_series_equal(df.x, df.y, check_names=False)
 
 
+# You can work around this by .cache()ing the table.
 @pytest.mark.notyet("sqlite", reason="instances are uncorrelated")
 @mark_impures
 def test_chained_selections(alltypes, impure):
@@ -153,9 +129,7 @@ impure_params_uncorrelated = pytest.mark.parametrize(
             lambda _: ibis.random(),
             marks=[
                 *no_randoms,
-                pytest.mark.notyet(
-                    ["impala", "trino"], reason="instances are correlated"
-                ),
+                pytest.mark.notyet(["impala"], reason="instances are correlated"),
             ],
             id="random",
         ),
@@ -164,9 +138,7 @@ impure_params_uncorrelated = pytest.mark.parametrize(
             lambda _: ibis.uuid().cast(str).contains("a").ifelse(1, 0),
             marks=[
                 *no_uuids,
-                pytest.mark.notyet(
-                    ["mysql", "trino"], reason="instances are correlated"
-                ),
+                pytest.mark.notyet(["mysql"], reason="instances are correlated"),
             ],
             id="uuid",
         ),
@@ -174,7 +146,8 @@ impure_params_uncorrelated = pytest.mark.parametrize(
             lambda table: my_random(table.float_col),
             marks=[
                 *no_udfs,
-                pytest.mark.notyet("duckdb", reason="instances are correlated"),
+                # no "impure" argument for pyspark yet
+                pytest.mark.notimpl("pyspark"),
             ],
             id="udf",
         ),
@@ -182,6 +155,7 @@ impure_params_uncorrelated = pytest.mark.parametrize(
 )
 
 
+# You can work around this by doing .select().cache().select()
 @pytest.mark.notyet(["clickhouse"], reason="instances are correlated")
 @impure_params_uncorrelated
 def test_impure_uncorrelated_different_id(alltypes, impure):
@@ -191,15 +165,57 @@ def test_impure_uncorrelated_different_id(alltypes, impure):
     # eg if you look at the following SQL:
     # select random() as x, random() as y
     # Then x and y should be uncorrelated.
-    df = alltypes.select(x=impure(alltypes), y=impure(alltypes)).execute()
+    expr = alltypes.select(x=impure(alltypes), y=impure(alltypes))
+    df = expr.execute()
     assert (df.x != df.y).any()
 
 
+# You can work around this by doing .select().cache().select()
 @pytest.mark.notyet(["clickhouse"], reason="instances are correlated")
 @impure_params_uncorrelated
 def test_impure_uncorrelated_same_id(alltypes, impure):
     # Similar to test_impure_uncorrelated_different_id, but the two expressions
     # have the same ID. Still, they should be uncorrelated.
     common = impure(alltypes)
-    df = alltypes.select(x=common, y=common).execute()
+    expr = alltypes.select(x=common, y=common)
+    df = expr.execute()
     assert (df.x != df.y).any()
+
+
+@pytest.mark.notyet(
+    [
+        "duckdb",
+        "clickhouse",
+        "datafusion",
+        "mysql",
+        "impala",
+        "mssql",
+        "trino",
+        "flink",
+        "bigquery",
+    ],
+    raises=AssertionError,
+    reason="instances are not correlated but ideally they would be",
+)
+@pytest.mark.notyet(
+    ["sqlite"],
+    raises=AssertionError,
+    reason="instances are *sometimes* correlated but ideally they would always be",
+    strict=False,
+)
+@pytest.mark.notimpl(
+    ["polars", "risingwave", "druid", "exasol", "oracle", "pyspark"],
+    raises=com.OperationNotDefinedError,
+)
+def test_self_join_with_generated_keys(con):
+    # Even with CTEs in the generated SQL, the backends still
+    # materialize a new value every time it is referenced.
+    # This isn't ideal behavior, but there is nothing we can do about it
+    # on the ibis side. The best you can do is to .cache() the table
+    # right after you assign the uuid().
+    # https://github.com/ibis-project/ibis/pull/9014#issuecomment-2399449665
+    left = ibis.memtable({"idx": list(range(5))}).mutate(key=ibis.uuid())
+    right = left.filter(left.idx < 3)
+    expr = left.join(right, "key")
+    result = con.execute(expr.count())
+    assert result == 3
--- a/ibis/backends/tests/test_impure.py
+++ b/ibis/backends/tests/test_impure.py
@@ -99,9 +99,7 @@ mark_impures = pytest.mark.parametrize(
             lambda table: my_random(table.float_col),
             marks=[
                 *no_udfs,
-                pytest.mark.notyet(
-                    ["flink", "postgres"], reason="instances are uncorrelated"
-                ),
+                pytest.mark.notyet(["flink"], reason="instances are uncorrelated"),
             ],
             id="udf",
         ),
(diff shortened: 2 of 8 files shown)


NickCrews:
oooh, these CI failures are revealing differences in the backend behaviors... wheeee into the 🐰 🕳️ we go!

ibis/backends/tests/test_impure.py
```diff
@@ -0,0 +1,67 @@
...
+
+import ibis
+from ibis import _
+
+_counter = 0
+
+
+@ibis.udf.scalar.python(side_effects=True)
+def get_id() -> int:
+    global _counter  # noqa: PLW0603
```
cpcloud, line 14: I think it's sufficient to invoke `random.random()` here. The `global` counter is unnecessarily complicated.

cpcloud: The other consideration is that unless the database provides a flag for informing it that the UDF is impure, there's nothing Ibis can do to prevent incorrect assumptions based on use of global state.

You should really avoid this kind of assumption/architecture in general.

Still, we can make sure that UDFs are effectively assumed to behave as impure functions by default.

NickCrews: hmm right that is way better

NickCrews:
I can think of two reasons a user might care about the semantics here, and I hope we can support both of their needs:
1. impureness/correlatedness. If a function is impure, they may want it to be executed once, or many times, depending on if they want them to be correlated. See [this example](https://github.com/ibis-project/ibis/issues/8921#issuecomment-2064963393)
2. performance. If some computation is slow, they only want it to happen a single time.

So I think this means that we as ibis authors can't assume what the goals of the user is, and how many times they want an expression executed. Therefore, we shouldn't do any clever rewrites or mergings of selects. I think we need to keep a more 1:1 correspondence between what the user writes and the SQL we produce: Every time the user does a .select(), .mutate(), etc, (except for simple column renamings, and maybe a few other cases) that leads to exactly one more `with select .... as ...` in the generated SQL.

IDK, what do you think of this train of reasoning? I think I'm fairly convinced that those two use cases are the requirements for success, but perhaps there is a different/better way of accomplishing that goal

kszucs:
@NickCrews please rebase to test with https://github.com/ibis-project/ibis/pull/9023 change included

cpcloud:
@NickCrews My only objection would be that

> performance. If some computation is slow, they only want it to happen a single time.

Is not something we can enforce even if we never merge any select statements. This kind of guarantee is at the level of the query engine.

NickCrews:
@cpcloud yup you are right with guarantees,  "suggesting" to the backend is the best we can do.

What do you think in general of my proposal of "one CTE per .select"? I'm not sure if you're skeptical of the whole thing or just the performance claims... Thanks!

NickCrews:
I'm trying to decide what is higher priority:

 An implementation that is faster 90% of the time, but does clever things and therefore isn't able to be tuned by the user that 10% of the time they need it

Vs

An implementation that is a bit slower in the majority of cases, but is always fine tunable to get the perf you need in the edge cases.

NickCrews:
@cpcloud @kszucs I think this is ready for review whenever you get the chance! I think this is the groundwork for defining the current state, and after we get this in then we can start talking about what we think ideal behavior should be, and how to get there.

NickCrews:
Anything I can do here to help move this forward?

cpcloud:
@NickCrews Can you write a docstring for each of these tests explaining what's being tested in each? I know that's atypical, but this is a very hairy problem with lots of specifics that are important to understand, and words like "impure" and "correlation" need to be precisely defined so that everyone is on the same page about exactly what is being tested.

NickCrews:
I think that is a good idea. Will do.

NickCrews:
oops, I just wrote the notes as comments, not docstrings, I assume that is still OK? I think this is ready to review now, thanks!

NickCrews:
@cpcloud this is ready for a review whenever you get the chance! (assuming CI is still green after the rebase I just did)

NickCrews:
The nix CI failure looks like a flake, but IDK what is going on with postgres, any idea? just mark as notimpl?

cpcloud:
I'm not sure what you mean by flake here. The new test you added for impure function behavior is breaking in some strange way, which is something we have discussed in other issues.

I'm pretty much ready to declare bankruptcy for now on any guarantees for any impure functions runs as UDFs. People are pretty much on their own right now if they're expecting something resembling reasonable behavior from stateful UDFs.

ibis/backends/tests/test_impure.py
```diff
@@ -0,0 +1,207 @@
...
+    # An "impure" expression is random(), uuid(), or some other non-deterministic UDF.
+    # If we evaluate it for two different rows in the same relation,
+    # we might get different results. This is expected.
+    # But, as soon as we .select() it into a new relation, then that "locks in" the
+    # value, and any further references to it will be the same.
+    # eg if you look at the following SQL:
+    # WITH
+    #   t AS (SELECT random() AS common)
+    # SELECT common as x, common as y FROM t
+    # Then both x and y should have the same value.
```
cpcloud, line 124: I don't think this is actually an expectation that Ibis can enforce.

Trino, for example evaluates CTEs once **per reference**, leading to annoying behavior like this:

```
trino:default> with t as (select 1 as i, random() as x) select * from t union select * from t;
 i |         x
---+--------------------
 1 | 0.8021965004733393
 1 | 0.3296580934824612
(2 rows)

Query 20240724_211040_00063_56j8j, FINISHED, 1 node
Splits: 98 total, 98 done (100.00%)
0.22 [0 rows, 0B] [0 rows/s, 0B/s]
```

[This issue has been open for years](https://github.com/trinodb/trino/issues/10), it's issue 10 on an issue tracker with nearly 23k total issues + PRs combined, for a project that is used in production at multiple large companies, so I don't have high hopes that it will be addressed any time soon.

No amount of CTE-ization that we do is going to change the way Trino works.

ibis/backends/tests/test_impure.py
```diff
@@ -0,0 +1,207 @@
...
+                *no_udfs,
+                pytest.mark.broken("duckdb", reason="instances are correlated"),
+            ],
+            id="udf",
+        ),
+    ],
+)
+
+
+@pytest.mark.broken(["clickhouse"], reason="instances are correlated")
```
cpcloud, line 187: Did you look into why this is?

Two calls to `rand` generate the same number:

```
falcon :) select rand32(), rand32();

SELECT
    rand32(),
    rand32()

Query id: edfb1ecd-8070-48d9-8fc7-1b49b6f46fd9

   ┌───rand32()─┬───rand32()─┐
1. │ 3907068806 │ 3907068806 │
   └────────────┴────────────┘

1 row in set. Elapsed: 0.001 sec.
```

Do we now need to start seeding these functions with a random number from Python to pass to clickhouse?

What if someone is share a clickhouse connection across threads, does Ibis now need to manage random number generate thread local state?

These are the kinds of problems that we have to consider before writing ourselves into a corner with test cases.

ibis/backends/tests/test_impure.py
NickCrews: Hmm, darn, I agree with your analysis of the situation. I agree it sounds like we can never guarantee this for trino. 

But that is trinos fault for not following the SQL standard, and I don't think we should throw out "better" semantics because they're not "perfect". Ibis exposes plenty of APIs that are inconsistent between backends, but those are much less obvious about what the proper behavior should be. If we feel OK about exposing inconsistent behavior in those circumstances, when it is not obvious what they correct behavior is, then I think it should be an easy decision to expose this inconsistent behavior.

Any user of trino who runs into this bug is out in the cold no matter what we decide here, they're gonna get weird results. But I as a user of duckdb or the other backends could see benefits if we make it so each select/mutate is a new relation.

If we don't enforce the semantics of "one select=one relation", then what is the alternative here? Say I wanted to end up with the above SQL, so that I could call random() once and then reference it multiple times. Is ibis making no guarantees, ie the only way would be for me to cache the intermediate result ie `x.mutate(common=ibis.random()).cache().select(x="common", y="common")`?

NickCrews: wait, am I reading these test markers right that trino is actually passing this test? If I try

```python
import ibis
e = ibis.table({"x": int}).select(common=ibis.random()).select(x="common", y="common")
ibis.to_sql(e, "trino")
```

I get 

```sql
SELECT
  "t1"."common" AS "x",
  "t1"."common" AS "y"
FROM (
  SELECT
    RAND() AS "common"
  FROM "unbound_table_1" AS "t0"
) AS "t1"
```

which looks to me like it should be susceptible to the bug you link above?

ibis/backends/tests/test_impure.py
NickCrews: Hmm, that is interesting. I don't think the SQL spec is clear here on what is correct behavior, and I suppose even if I didn't know SQL and I looked at `select rand(), rand();` I wouldn't be surprised if you told me they could eval to the same value.

What about relaxing this test to checking that `t.select(x=ibis.random()).select(y=ibis.random())` yields `x != y`?

I think this would leave us with ibis semantics of
- if in the same select statement, two pure functions WILL give the same results (duh)
- if in different select statements, two pure functions WILL give the same results (duh)
- if in the same select statement, two impure functions may or may not give the same results. Ibis makes no guarantees. IDK, someone might still be able to footgun themselves, but at least right now I can't think of a way of protecting them...
- if in different select statements, two impure functions WILL give different results.
- if any value (pure or impure) is defined in one select statement, then any references to it in subsequent .select()s WILL all give the same value

NickCrews:
I would still love to enforce the `test_impure_correlated` test and the `test_chained_selections` test`, so we can enforce that `table.select(common=ibis.random()).select(x="common", y="common")` works as expected, on as many backends as possible. This is a real pain point for me, currently I have to .cache() the tables in intermediate stages to get this guarantee, and I'm scared I'm going to forget to do it sometime. It appears that this will never work on [Trino](https://github.com/trinodb/trino/issues/10) or [sqlite](https://stackoverflow.com/questions/64328853/sqlite-random-function-in-cte), since they don't follow the SQL spec. Is the main issue here that we want ibis to give the same behavior, no matter the backend you use? That seems unachievable, but are there any other concerns here that I could try to mitigate?

The `test_impure_uncorrelated` test I am happy to drop. I'm not sure that is part of the SQL standard, and anyway it is not nearly so obvious what `select RANDOM() as r1, RANDOM() as r2` is supposed to give.
- If users want guaranteed correlation, they can do a multi-step CTE as above: `t.select(r1=ibis.random()).mutate(r2="r1")`
- If users want guaranteed uncorrelation, they can do a multi-step CTE as `t.select(r1=ibis.random()).mutate(r2=ibis.random())` (or perhaps I'm mistaken, and even this won't guarantee uncorrelation on some backends??)

cpcloud:
I think we can keep the same merging strategy we have now, which is based on a simple notion of complexity comparison, and treat impure (random, uuid) and potentially impure (UDFs) functions as maximally complex which will prevent projections with those expressions from being merged.

cpcloud:
@NickCrews I think I've managed to come up with a decent compromise, which is to treat `ops.Impure` instances as having maximum complexity, and thus prevent merging selects with those operations while retaining the ability to merge pure-function-only selects.

Doing this exposed a few bugs too, which I've fixed here.

ibis/backends/tests/test_impure.py
cpcloud: It might pass, but it's not **guaranteed** to.

ibis/backends/tests/test_impure.py
```diff
@@ -0,0 +1,203 @@
...
+from ibis import _
+from ibis.backends.tests.errors import (
+    PsycoPg2InternalError,
+    Py4JJavaError,
+    PyDruidProgrammingError,
+)
+
+tm = pytest.importorskip("pandas.testing")
+
+pytestmark = pytest.mark.xdist_group("impure")
```
NickCrews, line 18: why is this needed?

ibis/expr/decompile.py
```diff
@@ -84,6 +83,11 @@ def translate(op, *args, **kwargs):
     raise NotImplementedError(op)
 
 
+@translate.register(ops.TimestampNow)
```
NickCrews, line 86: is this needed for this PR or just an unrelated fixup? (looks like the right change regardless)

ibis/backends/tests/test_impure.py
```diff
@@ -0,0 +1,203 @@
...
+                ),
+            ],
+            id="uuid",
+        ),
+        pytest.param(
+            lambda table: my_random(table.float_col),
+            marks=[
+                *no_udfs,
+                # no "impure" argument for pyspark yet
+                pytest.mark.notimpl("pyspark"),
```
NickCrews, line 173: notyet instead of notimpl, since it's a problem on the backend side?

NickCrews:
Nice job finding those bugs and fixing them! Looks great. I was surprised when I had to make duckdb as xfail, I thought these semantics they would get right.

Reminder to myself: I want to also add other possibly impure functions to `test_impure_correlated`, such as `Array.collect()` and `.arbitrary()`.

I think that compromise using complexity works for me. If it were me, I think I would opt for not merging, (more complex SQL, but simpler implementation and guaranteed as-correct-as-possible), but I'll live with it. For example, if we add more impure functions in the future, it would be very possible to accidentally not inherit from `ops.Impure`, and we would have a bug.

cpcloud:
Collect and arbitrary are not impure. Whatever non determinism there is you have to address in some other way.

cpcloud:
...and the current way to do that is with the order_by keyword argument that landed in 9.3.

ibis/backends/tests/test_impure.py
cpcloud: It's not, we just don't pass it through, unless pyspark doesn't have the ability to set this property.

ibis/backends/tests/test_impure.py
cpcloud: Concurrent execution of `CREATE OR REPLACE FUNCTION` in postgres doesn't seem to work. This ensures that all tests in this module run in the same process as long as `--dist=loadgroup` is passed, which it is.

ibis/expr/decompile.py
cpcloud: It's needed for this PR, but I'm not entirely sure why. I'll poke around a bit.

ibis/expr/operations/udf.py
```diff
@@ -51,7 +51,7 @@ class InputType(enum.Enum):
 
 
 @public
-class ScalarUDF(ops.Value):
+class ScalarUDF(ops.Impure):
```
cpcloud, line 54: I'm going to split out builtin UDFs and actually-user-defined ones, since builtins aren't in general impure.

cpcloud: Will do in a follow-up though.

ibis/backends/tests/test_impure.py
NickCrews: ah, I see, thanks. Looks like [it is implemented in spark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.udf.html?highlight=udf#pyspark.sql.functions.udf), we just need to expose it.

ibis/backends/tests/test_impure.py
NickCrews: Makes sense, that seems like something we shouldn't worry about further. Maybe add a comment here? Also fine to not, if someone removes it they will find out from failing tests.

ibis/backends/tests/test_impure.py
cpcloud: Let's do this in a follow-up. Lucky for us, each engine appears to use a slightly different way of specifying whether a UDF is deterministic 🙄

cpcloud:
TODO: add a test for self join with unique keys on the original table.

cpcloud (approved):

cpcloud:
It seems like this behavior is intentional across our backends, and that writing a query as a CTE **in no way** guarantees how it will be evaluated.

NickCrews:
Darn it.

Looks like [DuckDB is aware of this](https://github.com/duckdb/duckdb/issues/12875), and they want to fix it, but not currently feasible for them to fix.

NickCrews:
ughh, what do we want to do? In order to get the self-join behavior that the user wants, I think the backend would have to materialize a temporary table. But I really don't think we want to do that automatically for them. So the best we could do is to detect this happening, and then error? That sounds quite hairy.

cpcloud:
I think the best we can do is to recommend calling `.cache()` in these cases. That's the only thing that will guarantee the desired semantics.

NickCrews:
Do we do any sort of check to ensure people don't footgun themselves, or do we only just let them get weird results and then have to start searching the issues/docs? I can't think of where this should go in the docs, maybe we want a "common pitfalls" or "troubleshooting" page? I'm not sure if there are other footguns like this that we should call out that would also fit in there

cpcloud:
> Do we do any sort of check to ensure people don't footgun themselves, or do we only just let them get weird results and then have to start searching the issues/docs? I can't think of where this should go in the docs, maybe we want a "common pitfalls" or "troubleshooting" page? I'm not sure if there are other footguns like this that we should call out that would also fit in there

We don't have a check for this and I don't think it's worth adding because of the complexity of such a check.

> I'm not sure if there are other footguns like this that we should call out that would also fit in there.

There are almost certainly are, but let's try to stay focused on **this** issue before moving on to other things.

NickCrews:
OK, sounds good to me to just get flink passing and then get this PR merged and call it a day, at least the behavior is written down somewhere. How does this sound as a plan?

I added a few comments to the tests so someone sees what our conclusions were.

ibis/backends/tests/test_impure.py
cpcloud: I'll add a comment, because I already forgot when revisiting this in review 😅 !

cpcloud:
Still not entirely sure why the Flink backend is failing.

cpcloud:
Turns out this never worked, but was passing because:

1. The rich output formatter selects a subset of columns before calling `to_pyarrow` on the table expression to avoid computing a bunch of stuff that wouldn't ever show up in the terminal due to lack of space imposed by user settings or the width of the terminal.
2. This PR caused the problem to show up because previously, column inlining optimization allowed the `delta_col` column in the test to be inlined in the top-level select, and then the subset was computed, which did **not** include `delta_col`, so that part of the SQL query never ran and thus never triggered the exception.

This PR disables column inlining for anything subclassing `ops.Impure` due to the nature of impure operations--impure roughly equates to "stateful"--of which the `now()` function is one.

This means that in this test case the inner query will remain, and thus likely (though not guaranteed (!) because the engine might still eliminate the inner query) still evaluate the inner query. In the flink case, this is what happens.

For now, I'll xfail the flink backend on this test.

Going forward, we should figure out a way to allow elimination of subexpressions **before** inlining, as removal of unused impure expressions should allow more opportunities for inlining.

ibis/backends/tests/test_generic.py
```diff
@@ -1057,6 +1057,7 @@ def test_between(backend, alltypes, df):
     backend.assert_series_equal(result, expected)
 
 
+@pytest.mark.notyet(["flink"], reason="timestamp subtraction doesn't work")
```
NickCrews, line 1060: Instead of xfailing, can you instead add `.drop("delta_col")` so that we still get the majority of this test on flink? It would make it so that the rest of the backends won't be tested for TimeDelta dtypes, but that should be tested more thoroughly elsewhere. More context in [cpcloud's comment here](https://github.com/ibis-project/ibis/pull/9014#issuecomment-2607166321).

NickCrews: JK, I see we explicitly ADD the delta_col here in the test. xfailing sounds better.

ibis/backends/tests/test_impure.py
NickCrews: I added a comment

NickCrews:
> Turns out this never worked, but was passing because:

wow, that is very interesting. Nice job figuring that one out!

PR #9015, opened 2024-04-18
perf(dask): avoid unnecessary calls to `compute` for `cov`/`corr`/`argmin`/`argmax`

### LINKED ISSUE #8187: perf(dask): avoid calling compute for `ops.Covariance`, `ops.ArgMin` and `ops.ArgMax`
> dask has built in covariance and correlation methods, why are they not sufficient here/why do we need to compute?
>
> IMO anything that needs to compute like this shouldn't warn, we should just not support it unless we can do it cleanly.

_Originally posted by @jcrist in https://github.com/ibis-project/ibis/pull/8005#discussion_r1473095470_

jcrist:
Avoids calling `compute` when executing `cov`/`corr`/`argmin`/`argmax`.

This necessarily drops support for `cov`/`corr` with `how="pop"`, since dask doesn't include native support (yet).

Fixes #8187.

### DIFF
--- a/ibis/backends/dask/executor.py
+++ b/ibis/backends/dask/executor.py
@@ -27,7 +27,7 @@ from ibis.backends.pandas.rewrites import (
     PandasWindowFunction,
     plan,
 )
-from ibis.common.exceptions import UnboundExpressionError
+from ibis.common.exceptions import UnboundExpressionError, UnsupportedOperationError
 from ibis.formats.pandas import PandasData, PandasType
 from ibis.util import gen_name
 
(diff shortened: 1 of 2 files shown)


cpcloud (approved):

cpcloud:
Hm, looks like some failures made it through 🤔

jcrist:
Weird. I clicked "enable auto-merge", wouldn't have expected that to merge if CI wasn't completely green.

cpcloud:
Yep, looking into it.

cpcloud:
For some reason the `backends` "summary" job was skipped. I think it might be because our `ibis-backends-skip-helper.yml` workflow's job names don't exactly match our `ibis-backends.yml` workflow's job names.

This is a hairy part of GHA that is both hard to reason about and test 😬

PR #9021, opened 2024-04-19
feat: make Table.bind() a public API

NickCrews:
I want access to this as a library author in [mismo](https://github.com/NickCrews/mismo/blob/a5874fab5c20505055e77545a83a60fe323eccd2/mismo/_util.py#L25)


kszucs:
I also does this in https://github.com/ibis-project/ibis/pull/8917

cpcloud:
Closing this out in favor of #8917. I think we have consensus there on the primary issue so we should merge it ASAP.

Issue #8908, opened 2024-04-06
feat: Add to_date and to_time to StringValue


saschahofmann:
### Describe the solution you'd like

I think it would make sense to have date and time parsing functions similar to the already existing `to_timestamp` function.

A workaround right now could be to use `to_timestamp` and extract the time or date part but e.g. BigQuery supports extra parsing functions `PARSE_TIME` and `PARSE_DATE`

### What version of ibis are you running?

8.0.0

### What backend(s) are you using, if any?

BigQuery, Postgres

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

 
- [x] to_date #9030 
- [ ] to_time

deepyaman:
@saschahofmann Your request makes sense! I'll look to implement this, with `to_date` and `to_time` falling back to `to_timestamp` followed by extracting the appropriate component for backends that don't support it directly.

saschahofmann:
Cool. Also to happy to look into it but I would need some pointers on where these things are implemented.

deepyaman:
> Cool. Also to happy to look into it but I would need some pointers on where these things are implemented.

Amazing!

So, `to_timestamp` is defined here: https://github.com/ibis-project/ibis/blob/356e4593f21bf662f5e55f307694a616c712eeec/ibis/expr/types/strings.py#L1275-L1302

You would need to add functions for `to_date` and `to_time` to the API. I'd highly recommend just starting with one; probably `to_date` will run into less issues (e.g. with timezones).

You'll also need to add a corresponding operator: https://github.com/ibis-project/ibis/blob/356e4593f21bf662f5e55f307694a616c712eeec/ibis/expr/operations/temporal.py#L74

Finally, you should add tests, like https://github.com/ibis-project/ibis/blob/main/ibis/backends/tests/test_temporal.py#L1431 

Once you've done this, you implement translations. For example:

* Base SQLGlot translation: https://github.com/ibis-project/ibis/blob/356e4593f21bf662f5e55f307694a616c712eeec/ibis/backends/sql/compiler.py#L281
* BigQuery override, if necessary: https://github.com/ibis-project/ibis/blob/356e4593f21bf662f5e55f307694a616c712eeec/ibis/backends/bigquery/compiler.py#L247
* A bunch of other overrides, as necessary: https://github.com/search?q=repo%3Aibis-project%2Fibis%20stringtotimestamp&type=code

[...]

PR #9030, opened 2024-04-20
feat: add to_date function to StringValue

saschahofmann:
## Description of changes

* Adds `to_date` to string types that accepts a format string and parses a string to Date type.

* Uses a cast to timestamp and extract time as a fallback like this
```python
def visit_StringToDate(self, op, *, arg, format_str):
        return self.f.date(self.f.str_to_time(arg, format_str))
```
* Implements native functions for bigquery, clickhouse, MySQL, oracle, postgres, and snowflake

## Issues closed
Implements half of #8908


saschahofmann:
@gforsyth I am struggling a little with running the tests. I couldn't find much on setting up the database. I can see that in conftest.py some env variable are read so I provided those for my local postgres database via the .env file (on a side note: I found it weird that the .env is committed? )

Now I seem to be missing an extension
```
psycopg2.errors.UndefinedFile: could not open extension control file "/opt/homebrew/share/postgresql@14/extension/plpython3u.control": No such file or directory
``` 

and it doesn't seem trivial to install it. Is there any documentation on setting up the env for running the tests?

gforsyth:
Hey @saschahofmann -- rather than using your own instance, I'd recommend installing `just` and running `just up postgres`.
That will spin up a postgres container with all the extensions needed and at the port that our pytest setup expects by default.

gforsyth:
There are also some instructions here: https://ibis-project.org/contribute/02_workflow#running-the-test-suite

If you're using a mac M2 it's more involved because of the architecture

saschahofmann:
Perfect. No idea why i didnt find that section. Will experiment around and hopefully have something more complete soon.

saschahofmann:
I know implemented the default fallback in sql/compiler.py L804 like this
```python
def visit_StringToDate(self, op, *, arg, format_str):
        return self.f.date(self.f.str_to_time(arg, format_str))
```
Is there a way to do it using ibis (something like `to_timestamp().date()`. Also I was thinking whether SQLGlot might have implemented it for its dialect already but I couldn't find it.

saschahofmann:
For postgres, the format for to_timestamp seems to be translated to the postgres native format but its not obvious to me where this happens? I would need to do something similar for to_date

gforsyth:
@saschahofmann -- `sqlglot` handles the timestamp formatting here: https://github.com/tobymao/sqlglot/blob/d3472c664fdfb7c9cfa9a54c6b0491b605cf4913/sqlglot/dialects/postgres.py#L186-L193

saschahofmann:
Alright. I found that SQLGlot has actually implemented StrToDate so I am just using that. What do I need to do make the doc test pass?

ibis/expr/types/strings.py
```diff
@@ -1302,6 +1302,35 @@ def to_timestamp(self, format_str: str) -> ir.TimestampValue:
...
+        >>> ibis.options.interactive = True
+        >>> t = ibis.memtable({"ts": ["20170206"]})
+        >>> t.ts.to_date("%Y%m%d")
+        ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
+        ┃ StringToDate(ts, '%Y%m%d')      ┃
+        ┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
+        │ date                            │
+        ├─────────────────────────────────┤
+        │ 2017-02-06                      │
+        └─────────────────────────────────┘
```
gforsyth, line 1330:
```suggestion
        ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
        ┃ StringToDate(ts, '%Y%m%d') ┃
        ┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
        │ date                       │
        ├────────────────────────────┤
        │ 2017-02-06                 │
        └────────────────────────────┘
```

I think it's just an annoying terminal width issue -- this _should_ fix it

saschahofmann:
I think its ready for review !

gforsyth (commented):
This looks really nice, @saschahofmann !

I think this patch will get snowflake passing:

```
index 75bfd7ce6..c0915d059 100644
--- a/ibis/backends/snowflake/compiler.py
+++ b/ibis/backends/snowflake/compiler.py
@@ -78,6 +78,7 @@ class SnowflakeCompiler(SQLGlotCompiler):
         ops.Hash: "hash",
         ops.Median: "median",
         ops.Mode: "mode",
+        ops.StringToDate: "to_date",
         ops.StringToTimestamp: "to_timestamp_tz",
         ops.TimeFromHMS: "time_from_parts",
         ops.TimestampFromYMDHMS: "timestamp_from_parts",
```

saschahofmann:
Huh, I think I actually had this but removed it because I thought it was in SQLGlot. Will put it back

saschahofmann:
I assume that the failing test is unrelated? At least its not clear to me where this would be coming from?

gforsyth (approved):
Thanks for putting this together @saschahofmann!  The snowpark error is definitely unrelated and _might_ be flaky, so I'm going to merge this and see what happens on CI.

PR #9043, opened 2024-04-23
feat(ir): support pretty printing arbitrary traversable objects

kszucs:
This enables pretty formatting the dereference mappings and also the replacements mappings we use in rewrites for better inspection. Also moved the irrelevant logic out from `format.py`.

### DIFF
--- a/ibis/common/typing.py
+++ b/ibis/common/typing.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import inspect
 import re
 import sys
 from abc import abstractmethod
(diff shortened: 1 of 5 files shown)


ibis/common/typing.py
```diff
@@ -259,3 +260,21 @@ class Coercible(Abstract):
     @classmethod
     @abstractmethod
     def __coerce__(cls, value: Any, **kwargs: Any) -> Self: ...
+
+
+def get_defining_frame(obj):
```
kszucs, line 265: Perhaps not the typing module is the best place for this, but we don't have an inspect module here yet. We can reorganize later since these are designed for public use.

PR #9046, opened 2024-04-24
chore(snowflake): remove XPASSing xfail on udf test

gforsyth:
This showed up in #9030 and I wasn't sure if it was flaky, but it looks like Snowpark can now run Ibis as `caller`, too!


cpcloud (approved):
Nice!

Issue #8805, opened 2024-03-27
feat(api): API for converting binary columns to geometry columns 


cpcloud:
### Is your feature request related to a problem?

https://ibis-project.zulipchat.com/#narrow/stream/405265-tech-support/topic/spatial.20column.20woes

This works using `.sql` but that's only a workaround, we should have an API for this.

### Describe the solution you'd like

A way to either:

- `t.geometry.cast("geometry")`
- `t.geometry.to_geometry()` (API TBD)

Or both. Perhaps the latter can be used when casting.

### What version of ibis are you running?

`main`

### What backend(s) are you using, if any?

DuckDB

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cboettig:
just a note that we can currently hack around this only by using raw_sql:  

```python
import ibis
from ibis import _
from shapely.geometry import box
bounds = box(-2493045.0, 176655.0, 2343105.0, 3310995.0)

parquet = "https://data.source.coop/cboettig/pad-us-3/pad-us3-combined.parquet"
con = ibis.duckdb.connect()
con.load_extension("spatial")

# little ick:
con.raw_sql(f"CREATE VIEW pad AS SELECT *, st_geomfromwkb(geometry) as geom from read_parquet('{parquet}')")
pad = con.table("pad")
```

but now this is fast:

```python
%%time
# testing
(pad.
    filter(_.geom.within(bounds)).
    group_by([_.Mang_Type]).
    aggregate(n = _.count()).
    to_pandas()
)
```


Wall time: 14.8 s


Compare this to opening identical data as flatgeobuf in duckdb:

```
%%time
fgb = "https://data.source.coop/cboettig/pad-us-3/pad-us3-combined.fgb"

(con.read_geo(fgb).
    filter(_.geom.within(bounds)).
    group_by([_.Mang_Type]).
    aggregate(n = _.count()).
    to_pandas()
)

```

Wall time: 4min 10s

jcrist:
> `t.geometry.to_geometry()` (API TBD)

Small comment - I think we should reserve `to_*` methods for things like `to_parquet`/`to_csv`/... that create output artifacts and run eagerly.

cpcloud:
Ah good point. I wonder if `as_geo()` would be preferable? Or perhaps `as_geometry()`?

jcrist:
[...]

PR #9062, opened 2024-04-26
feat(geo-duckdb): support casting binary to geometry

ncclementi:
Add support to cast wkb to geometry type in duckdb backend

Closes https://github.com/ibis-project/ibis/issues/8805

This PR: needs a data file, CI will fail until https://github.com/ibis-project/testing-data/pull/8 gets merged

### DIFF
--- a/ibis/backends/duckdb/tests/conftest.py
+++ b/ibis/backends/duckdb/tests/conftest.py
@@ -33,6 +33,15 @@ TEST_TABLES_GEO = {
     ),
 }
 
+TEST_TABLE_GEO_PARQUET = {
+    "geo_wkb": ibis.schema(
+        {
+            "name": "string",
+            "geom": "binary",
+        }
+    ),
+}
+
 
 class TestConf(BackendTest):
     supports_map = True
@@ -65,6 +74,14 @@ class TestConf(BackendTest):
                     CREATE OR REPLACE TABLE {table} AS
                     SELECT * FROM st_read('{geojson_dir / f'{table}.geojson'}')
                     """
+                )
+            for table in TEST_TABLE_GEO_PARQUET:
+                # the ops on this table will need the spatial extension
+                yield (
+                    f"""
+                CREATE OR REPLACE TABLE {table} AS
+                SELECT * FROM read_parquet('{parquet_dir / f'{table}.parquet'}')
+                """
                 )
             yield (
                 """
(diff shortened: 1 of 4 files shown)


ibis/backends/conftest.py
```diff
@@ -144,6 +144,12 @@ def is_older_than(module_name, given_version):
             "total_eva_hrs": "float64",
         }
     ),
+    "geo_wkb": ibis.schema(
+        {
+            "name": "string",
+            "geom": "binary",
+        }
+    ),
```
gforsyth, line 152: entries here are also expected to have a `csv` file in testing-data -- might be better off sticking this in the duckdb `conftest.py` for now.

ibis/backends/duckdb/tests/test_geospatial.py
```diff
@@ -340,3 +340,10 @@ def test_geo_binop_geo_literals(ext_dir, geo_line_lit):
     expr = geo_line_lit.distance(shapely.Point(0, 0))
 
     assert con.execute(expr) == 0
+
+
+def test_cast_wkb_to_geo(con):
+    t = con.table("geo_wkb")
+    geo_expr = t.geometry.cast("geometry").centroid()
+    con.execute(geo_expr)  # check that the we can do a geo op
```
cpcloud, line 348: This test should verify that the output column is actually a geometry/geography dtype, otherwise it's not actually testing the effect of the cast.

ibis/backends/duckdb/tests/conftest.py
```diff
@@ -66,6 +75,14 @@ def ddl_script(self) -> Iterator[str]:
                     SELECT * FROM st_read('{geojson_dir / f'{table}.geojson'}')
                     """
                 )
+            for table in TEST_TABLE_GEO_PARQUET:
+                # the ops on this table will need the spatial extension
+                yield (
+                    f"""
+                CREATE OR REPLACE TABLE {table} AS
+                SELECT * FROM read_parquet('{parquet_dir / f'{table}.parquet'}')
+                """
```
cpcloud, line 84:
```suggestion
                    CREATE OR REPLACE TABLE {table} AS
                    SELECT * FROM read_parquet('{parquet_dir / f'{table}.parquet'}')
                    """
```

cpcloud (changes requested):

ibis/backends/duckdb/tests/test_geospatial.py
ncclementi: The output column will be converted to pandas, I was not able to figure out how to check that was a geometry, but I thought that if I can perform a geospatial operation then that confirm it's a geometry column, otherwise I wouldn't be able too.

cpcloud: I think an assertion that the column is a geopandas series is enough. The result isn't an Ibis expression, so e.g., checking `result.type().is_geospatial()` won't work. Something like `assert isinstance(result, gpd.GeoSeries)` should suffice.

cpcloud (approved):

PR #9064, opened 2024-04-26
fix(sql): don't fuse subsequent filter operations

kszucs:
"fixes" https://github.com/ibis-project/ibis/issues/9058

Issue #9058, opened 2024-04-25
bug: merging selections combines filters in incorrect way


NickCrews:
### What happened?

I do a filter-mutate-filter, and the order needs to stay in that order to be correct, but ibis appears to jumble the order of these steps:

```python
import ibis
from ibis import _

ibis.options.interactive = True

t = ibis.read_parquet(
    "https://github.com/NickCrews/election-data/releases/download/v2024-04-05_0/cleaned.parquet"
)
t = t.cache()
t = t.filter(_.votes != "*").mutate(votes=_.votes.cast("int64")).filter(_.votes > 0)
# occasionally, but not always, I get
# ConversionException: Conversion Error: Could not convert string '*' to INT64
t.votes.mean()
```

This gives the sql

```sql
SELECT
  AVG("t1"."votes") AS "Mean(votes)"
FROM (
  SELECT
    "t0"."year",
    "t0"."date",
    "t0"."state_po",
    "t0"."county_name",
    "t0"."county_fips",
    "t0"."jurisdiction_name",
    "t0"."jurisdiction_fips",
    "t0"."district",
    "t0"."office",
    "t0"."magnitude",
    "t0"."special",
    "t0"."stage",
    "t0"."precinct",
    "t0"."writein",
    "t0"."candidate",
    "t0"."party_detailed",
    "t0"."mode",
    CAST("t0"."votes" AS BIGINT) AS "votes",
    "t0"."readme_check",
    "t0"."county_fips2",
    "t0"."jurisdiction_fips2"
  FROM "ibis_cache_hys22fcdbnbmpdoasprivdb3pm" AS "t0"
  WHERE
    "t0"."votes" <> '*' AND CAST("t0"."votes" AS BIGINT) > CAST(0 AS TINYINT)
) AS "t1"
```

I think we need to keep these as two separate relations. I think this is somewhat related to https://github.com/ibis-project/ibis/pull/9014#issuecomment-2065547342

### What version of ibis are you using?

main

### What backend(s) are you using, if any?

duckdb

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gforsyth:
Thanks for the report @NickCrews -- we'll probably disable merging filters across selects to avoid this happening.  In the interim (and we won't release 9.0 until we sort this out), you can probably use `try_cast` to handle this without the failure mode.

kszucs:
[...]

PR #9065, opened 2024-04-26
feat(sql): add option to enable/disable select merging

kszucs:
Add a global option to enable or disable merging select statements where possible. This can serve as an escape route if the optimization fails. It is currently enabled by default, we may change to make it opt-in. 

xref #9064 and https://github.com/ibis-project/ibis/issues/9058

### DIFF
--- a/ibis/backends/sql/compiler.py
+++ b/ibis/backends/sql/compiler.py
@@ -27,6 +27,7 @@ from ibis.backends.sql.rewrites import (
     rewrite_capitalize,
     sqlize,
 )
+from ibis.config import options
 from ibis.expr.operations.udf import InputType
 from ibis.expr.rewrites import replace_bucket
 
@@ -453,7 +454,12 @@ class SQLGlotCompiler(abc.ABC):
         # substitute parameters immediately to avoid having to define a
         # ScalarParameter translation rule
         params = self._prepare_params(params)
-        op, ctes = sqlize(op, params=params, rewrites=self.rewrites)
+        op, ctes = sqlize(
+            op,
+            params=params,
+            rewrites=self.rewrites,
+            fuse_selects=options.sql.fuse_selects,
+        )
 
         aliases = {}
         counter = itertools.count()
--- a/ibis/backends/sql/rewrites.py
+++ b/ibis/backends/sql/rewrites.py
@@ -205,6 +205,7 @@ def sqlize(
     node: ops.Node,
     params: Mapping[ops.ScalarParameter, Any],
     rewrites: Sequence[Pattern] = (),
+    fuse_selects: bool = True,
 ) -> tuple[ops.Node, list[ops.Node]]:
     """Lower the ibis expression graph to a SQL-like relational algebra.
 
@@ -216,6 +217,8 @@ def sqlize(
         A mapping of scalar parameters to their values.
     rewrites
         Supplementary rewrites to apply to the expression graph.
+    fuse_selects
+        Whether to merge subsequent Select nodes into one where possible.
 
     Returns
     -------
@@ -240,7 +243,10 @@ def sqlize(
     )
 
     # squash subsequent Select nodes into one
-    simplified = sqlized.replace(merge_select_select)
+    if fuse_selects:
+        simplified = sqlized.replace(merge_select_select)
+    else:
+        simplified = sqlized
 
     # extract common table expressions while wrapping them in a CTE node
     ctes = frozenset(extract_ctes(simplified))
--- a/ibis/config.py
+++ b/ibis/config.py
@@ -47,6 +47,9 @@ class SQL(Config):
 
     Attributes
     ----------
+    fuse_selects : bool
+        Whether to fuse consecutive select queries into a single query where
+        possible.
     default_limit : int | None
         Number of rows to be retrieved for a table expression without an
         explicit limit. [](`None`) means no limit.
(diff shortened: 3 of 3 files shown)


cpcloud:
I'm not sure why we should keep select merging around at all. Does it have any non-aesthetic effect or benefit?

kszucs:
> I'm not sure why we should keep select merging around at all. Does it have any non-aesthetic effect or benefit?

It reduces the query size.

cpcloud:
> It reduces the query size. Also the user facing compiled query is much more readable with fusing enabled.

We don't have any evidence that either of these things is worth the bugs we've already had to fix with merging selects, and are likely to continue having to fix if we keep it around.

cpcloud:
Why not wait for issues to come in for the others? If we never have to add these optimizations, then that's code we don't have to maintain/support/test, etc.

cpcloud:
I feel pretty strongly about this given all the time and effort we have spent to avoid these bugs from the previous implementation.

It seems like if we keep merging around we're back to having to whack-a-mole incorrect optimizations yet again.

Personally I don't want to spend time on that anymore.

I think we're in a really great place with our core abstractions otherwise, and the work you've done on it has really moved the whole project forward.

cpcloud:
I'm not saying we should never explore this problem, only that we should avoid enabling optimizations--right before then next release--that we have seen are at high risk for correctness issues when we can simply disable them.

kszucs:
This PR is about providing an option to enable/disable the select merges. I would prefer to discuss the default value in the relevant issue.

kszucs:
Closing in favor of https://github.com/ibis-project/ibis/pull/9065

cpcloud:
I took the weekend to think about some of the issues here and did a bit of research on the SQL standard specification of predicate short-circuiting.

* According to [this SO Q&A](https://stackoverflow.com/a/909770) the SQL standard does **not** specify the evaluation order of predicates. That means that an engine is free to reorder (or not) predicates in any way AFAICT, including if it would change the semantics of the query.

For example, an engine can evaluate a fallible expression **first**, even if a subsequent expression with short circuiting would avoid the fallible expression evaluation failing.

In practice, DuckDB merges a fallible cast with an infallible operation **in the same way** producing the **exact same plan** regardless of how the SQL is specified:

```
D explain select * from (select * from t where cast(x as int) != 0) where x <> '*';

│┌───────────────────────────┐│
││       Physical Plan       ││
│└───────────────────────────┘│
┌───────────────────────────┐
│           FILTER          │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│ ((x != '*') AND (CAST(x AS│
│       INTEGER) != 0))     │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│           EC: 1           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         SEQ_SCAN          │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             t             │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             x             │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│           EC: 5           │
└───────────────────────────┘
D explain select * from (select * from t where x <> '*') where cast(x as int) != 0;

│┌───────────────────────────┐│
││       Physical Plan       ││
│└───────────────────────────┘│
┌───────────────────────────┐
│           FILTER          │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│ ((x != '*') AND (CAST(x AS│
│       INTEGER) != 0))     │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│           EC: 1           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         SEQ_SCAN          │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             t             │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             x             │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│           EC: 5           │
└───────────────────────────┘
```

This means that even if Ibis decided to generate subqueries that map one-to-one to SQL selects it provides absolutely zero guarantee that generating the SQL in that way produces specific behavior.

IMO this means that Ibis should be free to merge predicates, assuming our implementation of that process is correct.

Now, I'm not trying to gloss over the fact that we have had a number of issues with select merging, and so I think we should provide a way to disable it if it can potentially help unblock users.

I think the opt-in behavior would be kind of pointless, since people will never opt-in without lots of documentation and if we're going to go that route we should disable it.

On the other hand, if we think this is long-term a thing we want enabled, then we definitely need feedback on it from users. The only way to get that feedback AFAICT is to enable this feature.

That does mean we have to accept that the current implementation may not be 100% finished and will elicit bug reports.

The opt-out behavior IMO seems like an acceptable balance here: people can disable select merging on an as-needed basis, and we can continuously evaluate whether keeping it around is causing overhead that isn't worth it.

[Disabling select merging](https://github.com/ibis-project/ibis/pull/9070) it turns out is also not without issue: [SQLite fails due to a query we generate being too large for the parser stack](https://github.com/ibis-project/ibis/actions/runs/8875942618/job/24366459836?pr=9070#step:14:98), and [MS SQL loses the ability to specify order by in many cases as well as `LIKE` functionality in certain positions](https://github.com/ibis-project/ibis/actions/runs/8875942618/job/24366465798?pr=9070).

+1 on merging this PR and cutting a release after #8917 is merged.

Issue #9038, opened 2024-04-22
feat: add `catalog` & `database` support to `create_table`


mark-druffel:
## TLDR
I'm wondering if it's intended that the `database` argument in `create_table` works different than the one in `drop_table`? `create_table` only accepts a `str` and `drop_table` accepts a `tuple`. 

If I set the catalog and database via pyspark, `create_table` works as excepted, but I can't figure out a way to do so in my `create_table`, I had to do it through the pyspark session directly:
```
from pyspark.sql import SparkSession
import ibis
spark = SparkSession.builder.getOrCreate()
ispark = ibis.pyspark.connect(session = spark)
ispark._session.catalog.setCurrentCatalog("comms_media_dev")
ispark._session.catalog.setCurrentDatabase("dart_extensions")
ispark.create_table(name = "raw_camp_info", obj = df, overwrite = True, format="delta")
```
I can drop a table without accessing the pyspark session:
```
ispark.drop_table(name = "raw_camp_info", database=tuple(["comms_media_dev", "dart_extensions"]))
```

## Additional Details
To drop my table I can just specify the catalog and database in my call:
```
from pyspark.sql import SparkSession
import ibis
spark = SparkSession.builder.getOrCreate()
ispark = ibis.pyspark.connect(session = spark)
ispark.drop_table(name = "raw_camp_info", database=tuple(["comms_media_dev", "dart_extensions"]))
```

Trying the same approach with `create_table` if fails:
```
ispark.create_table(name = "raw_camp_info", obj = df, overwrite = True, format="delta", database=tuple(["comms_media_dev", "dart_extensions"]))
```
> py4j.Py4JException: Method setCurrentDatabase([class java.util.ArrayList]) does not exist

I also tried with dot separator:
```
ispark.create_table(name = "raw_camp_info", obj = df, overwrite = True, format="delta", database="comms_media_dev.dart_extensions")
```
> AnalysisException: too many name parts for schema name: comms_media_dev.dart_extensions

I then tried to set the catalog and provide the database name and got a permissions error. Looking through the error, it looks like create_table didn't pass the `database` argument because the database was set to default (i.e. `comms_media_dev.default`:
```
ispark._session.catalog.setCurrentCatalog("comms_media_dev")
ispark.create_table(name = "raw_camp_info", obj = df, overwrite = True, format="delta", database="dart_extensions")
```

[...]

PR #9067, opened 2024-04-26
fix(pyspark): unwind catalog/database settings in same order they were set

gforsyth:
We set the catalog using a context manager, and we set the database also
using a context manager. There is a weird edge case:

set catalog to comms_media_dev (succeeds)
set database to dart_extensions within comms_media_dev (succeeds)

then we write the table, great! Now we try to change catalog and database back in reverse order and...

set database to default (the previous value that we saved) within comms_media_dev (fails, you do not have permission to access that)

set catalog back to spark_catalog (or previous value, but we never get here because of the previous error.

So what we need to do is instead:

set catalog
set database
write table
set catalog back
set database back

It would be really great if spark would allow for setting both of these values at the same time, but that is apparently not a thing.

xref #9038

### DIFF
--- a/ibis/backends/pyspark/tests/test_client.py
+++ b/ibis/backends/pyspark/tests/test_client.py
@@ -1,8 +1,11 @@
 from __future__ import annotations
 
+import pytest
+
 import ibis
 
 
+@pytest.mark.xfail_version(pyspark=["pyspark<3.4"], reason="no catalog support")
 def test_catalog_db_args(con, monkeypatch):
     monkeypatch.setattr(ibis.options, "default_backend", con)
     t = ibis.memtable({"epoch": [1712848119, 1712848121, 1712848155]})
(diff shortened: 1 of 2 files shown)


gforsyth:
I forgot to include `catalog` in the arguments to `self.table` so we were looking for a table in a different DB that we _hadn't_ just created.

ibis/backends/pyspark/tests/test_client.py
```diff
@@ -20,3 +23,19 @@ def test_catalog_db_args(con, monkeypatch):
...
+def test_create_table_no_catalog(con, monkeypatch):
+    monkeypatch.setattr(ibis.options, "default_backend", con)
+    t = ibis.memtable({"epoch": [1712848119, 1712848121, 1712848155]})
+
+    # create a table in specified catalog and db
+    con.create_table("t2", database=("default"), obj=t, overwrite=True)
+
+    assert "t2" not in con.list_tables()
+    assert "t2" in con.list_tables(database="default")
+    assert "t2" in con.list_tables(database=("default"))
```
cpcloud, line 37: Did you mean to write something else here? `"default"` and `("default")` are equivalent Python expressions.

gforsyth: oh, that was just unthinking editing -- I'll rip that out

cpcloud (approved):

markdruffel-8451:
@gforsyth Just a heads up this is now working for my use case (multiple catalogs). Thanks so much for all your help and definitely feel free to hit me up if you ever need help testing pyspark backend 👍

PR #9075, opened 2024-04-29
feat(sql): use `SELECT *` for complete reprojections

kszucs:
Reduce the verbosity of the generated SQL queries by projecting `*` instead of all of the parent fields.

### DIFF
--- a/ibis/backends/tests/snapshots/test_sql/test_cte_refs_in_topo_order/bigquery/out.sql
+++ b/ibis/backends/tests/snapshots/test_sql/test_cte_refs_in_topo_order/bigquery/out.sql
@@ -1,6 +1,6 @@
 WITH `t1` AS (
   SELECT
-    `t0`.`key`
+    *
   FROM `leaf` AS `t0`
   WHERE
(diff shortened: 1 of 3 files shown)


gforsyth:
> @cpcloud could you please update the snapshots for the failing backends?

I've got it

ibis/backends/sql/rewrites.py
```diff
@@ -52,6 +52,9 @@ class Select(ops.Relation):
     predicates: VarTuple[ops.Value[dt.Boolean]] = ()
     sort_keys: VarTuple[ops.SortKey] = ()
 
+    def is_star_selection(self):
+        return tuple(self.values.items()) == tuple(self.parent.fields.items())
```
kszucs, line 56: Use `tuple()` to also check for projection order.

cpcloud: Worth caching this?

kszucs: Nope, it is only used during compilation which happens once.

PR #9077, opened 2024-04-29
docs: update sankey diagram

lostmygithubaccount:
## Description of changes

this started as trying to get dynamic light/dark mode for plotting (with the sankey diagram, but any) working with Quarto. I still couldn't figure it out. there's definitely a path that's like save both versions of the image and use JavaScript to dynamically load the right one. another option, if we could get the HTML for the page in Python, would be to use that

so anyway, this actually just updates the sankey diagram to put it in a funciton to choose light/dark mode, minor code improvements, and improving the colors

one suggestion from Jim was simplifying the cateogires and I agree, so now it's just SQL/DataFrame


lostmygithubaccount:
don't like the PNG, let me play with that

lostmygithubaccount:
I like the colors and change to 2 categories, but it's way harder to read. going back to draft mode

lostmygithubaccount:
it's better now IMO

docs/backends_sankey.py
```diff
@@ -2,8 +2,28 @@
...
+
+    new_hex_code = f"#{new_r:02x}{new_g:02x}{new_b:02x}"
+
+    return new_hex_code
+
+
+category_colors = {
+    "Ibis API": "#7C65A0",
+    "SQL code": "#6A9BC9",
+    "DataFrame code": "#D58273",
```
jcrist, line 22: Can't comment directly on the image, but the new image colors look a lot better. +1.

Also slight preference for dropping the "code" suffix and just going with "SQL" and "DataFrame" or maybe "SQL" and "DataFrame API" idk.

jcrist (approved):
This looks great, nice work.

Issue #9069, opened 2024-04-28
bug: pandas class SingleBlockManager not defined when declaring pandas scalar udf input type as pd.Series


ted0928:
### What happened?

I'm not sure if this problem is due to a bug or a usage issue.
Here's my code

``` python
import ibis
import pandas as pd
from pyspark.sql import SparkSession
ibis.options.interactive = True
session = SparkSession.builder.getOrCreate()
connect = ibis.pyspark.connect(session)
source = connect.create_view('source', ibis.memtable(dict(id=[1,2,3], name=['a', 'b', 'c'])))

@ibis.udf.scalar.pandas
def add_one(x:int)->int:
    return x + 1

df = source.mutate(id2=add_one(source.id))
```
this works.  But actually, x is a pd.Series type at runtime. 

When i declare x as pd.Series，raise Exception when getting the type hint.
``` python
@ibis.udf.scalar.pandas
def add_one(x:pd.Series)->int:
    return x + 1
```
```python
Traceback (most recent call last):
  File "/Users/ning.ln/Java/ibis/spark_test.py", line 8, in <module>
    @ibis.udf.scalar.pandas
     ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ning.ln/Java/ibis/ibis/expr/operations/udf.py", line 384, in pandas
    return _wrap(
           ^^^^^^
  File "/Users/ning.ln/Java/ibis/ibis/expr/operations/udf.py", line 83, in _wrap
    return wrap(fn) if fn is not None else wrap
           ^^^^^^^^
  File "/Users/ning.ln/Java/ibis/ibis/expr/operations/udf.py", line 80, in wrap
    deferrable(wrapper(input_type, fn, **kwargs)), fn
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ning.ln/Java/ibis/ibis/expr/operations/udf.py", line 154, in _make_wrapper
    node = cls._make_node(fn, input_type, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ning.ln/Java/ibis/ibis/expr/operations/udf.py", line 115, in _make_node
    fields = {
             ^
  File "/Users/ning.ln/Java/ibis/ibis/expr/operations/udf.py", line 117, in <dictcomp>
    pattern=rlz.ValueOf(annotations.get(arg_name)),
[...]

PR #9079, opened 2024-04-29
docs(udfs): improve udf docstring examples

gforsyth:
UDF arguments need to be Ibis expressions.  In the existing docstring examples
for the UDF decorators, it can be unclear that the integers being passed to
`add_one` are implicitly converted to Ibis literals.

To make it clearer, I've added memtables to all the examples, and used pandas-
and pyarrow-specific functions to the examples.


gforsyth:
xref #9069

Issue #9076, opened 2024-04-29
bug(geo-duckdb): ibis.read_parquet() binary cast to geometry 


ncclementi:
### What happened?

```python
In [1]: import ibis
   ...: from ibis import _
   ...: ibis.options.interactive = True
   ...: 
   ...: parquet = "https://data.source.coop/cboettig/pad-us-3/pad-us3-combined.parquet"
   ...: table = ibis.read_parquet(parquet)

In [2]: t = table.mutate(geo_col=_.geometry.cast("geometry"))
In [3]: t.geo_col
```

```python-traceback
CatalogException: Catalog Error: Scalar Function with name "st_geomfromwkb" is not in the catalog, but it exists in the spatial extension.

Please try installing and loading the spatial extension:
INSTALL spatial;
LOAD spatial;

```

### What version of ibis are you using?

main

### What backend(s) are you using, if any?

duckdb

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ncclementi:
@cpcloud  you mentioned the `to_sqlglot` method in the compiler would be a good place to fit in the logic to load the spatial extension, but we don't have that method in the duckdb compiler. In the base compiler under `backends/sql/compiler.py` I only see this 

https://github.com/ibis-project/ibis/blob/1926eb40ae4a8aadb4899645745145247636056c/ibis/backends/sql/compiler.py#L644-L648

I think Ideally we should be able to load the spatial extension when the casting is `to` a geometry type. 

The amgic happens here, but here there is no `con` to be able to load the extension. 

https://github.com/ibis-project/ibis/blob/1926eb40ae4a8aadb4899645745145247636056c/ibis/backends/duckdb/compiler.py#L340-L341

cpcloud:
We definitely can't and shouldn't load extensions in the compiler, that would mean you could never use the compiler without having already connected to DuckDB.

This has to happen in the backend, just before execution.

You can do this in a `_to_sqlglot` override in the duckdb `Backend` class:

```python
def _to_sqlglot(self, ...):
    if any geospatial types:
        self.load_extensions(["geospatial"])  # don't remember the exact API
    
    return super()._to_sqlglot(...)
```

PR #9080, opened 2024-04-29
fix(duckdb): load extension when executing geospatial expressions

ncclementi:
- Closes https://github.com/ibis-project/ibis/issues/9076

### DIFF
--- a/ibis/backends/duckdb/tests/test_geospatial.py
+++ b/ibis/backends/duckdb/tests/test_geospatial.py
@@ -310,7 +310,7 @@ def test_load_geo_example(con):
 
 
 # For the next two tests we really want to ensure that
-# load_extenstion("spatial") hasn't been run yet, so we create a new connection
+# load_extension("spatial") hasn't been run yet, so we create a new connection
 # instead of using the con fixture.
 
 
@@ -319,7 +319,7 @@ def geo_line_lit():
     return ibis.literal(shapely.LineString([[0, 0], [1, 0], [1, 1]]), type="geometry")
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture
 def ext_dir(tmp_path_factory):
     # this directory is necessary because of Windows extension downloads race
     # condition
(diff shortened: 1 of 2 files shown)


ibis/backends/duckdb/tests/test_geospatial.py
```diff
@@ -347,3 +347,14 @@ def test_cast_wkb_to_geo(con):
     geo_expr = t.geometry.cast("geometry")
     assert geo_expr.type().is_geospatial()
     assert isinstance(con.execute(geo_expr), gpd.GeoSeries)
+
+
+def test_load_spatial_casting(ext_dir):
```
cpcloud, line 352: If you're going to use a custom extension directory you need to produce a new one specific to this test call.

As is, `ext_dir` is a session-level fixture, which means that if another test runs before this one that happens to install the geospatial extension then you're not testing the thing that needs to be tested (or its success is implicitly depending on the fact that another test ran before it, and that'll lead to a flaky test), which is that geospatial operations load the extension.

cpcloud (changes requested):

ibis/backends/duckdb/tests/test_geospatial.py
ncclementi: I want to be able have a new connection that doesn't have the extension loaded. I was following what we did here 
https://github.com/ibis-project/ibis/blob/fead40be6a25f24672348fdefa8b35d0e7469f80/ibis/backends/duckdb/tests/test_geospatial.py#L322-L334

Which sounds like it's also being missed used here too? I can scope it to "function", though I'm not sure that's the path we want.

cpcloud: > Which sounds like it's also being missed used here too?

Probably 😅 

If you change the fixture decorator from

```python
```

to 

```python
@pytest.fixture
```

that should be sufficient (no `scope="function"` necessary).

cpcloud (approved):

PR #9089, opened 2024-04-30
fix(arrays): array.remove() erroneously removes NULLs

NickCrews:
Still a WIP, figuring out what should be desired.

`ibis.array([1, None, 2]).remove(2)` currently results in `[1]` on many backends. I would expect the NULLs to remain, eg `[1, None]`

### DIFF
--- a/ibis/backends/pandas/kernels.py
+++ b/ibis/backends/pandas/kernels.py
@@ -144,6 +144,12 @@ def array_position_rowwise(row):
         return -1
 
 
+def array_remove_rowwise(row):
+    if row["arg"] is None:
+        return None
+    return [x for x in row["arg"] if x != row["other"]]
+
+
 def array_slice_rowwise(row):
     arg, start, stop = row["arg"], row["start"], row["stop"]
     if isnull(start) and isnull(stop):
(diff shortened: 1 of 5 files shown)


cpcloud (changes requested):
What is the fix here?

What was the original behavior? What's the new behavior? Why is it desired?

Blank PR descriptions make it really difficult to prioritize reviews on behavioral changes, please include them in those types of PRs!

NickCrews:
Sorry, this is very much a WIP, should have marked it as such. Added a very basic description to the OP.

cpcloud:
I think the postgres choice of using `IS NOT DISTINCT FROM` semantics in remove is inconsistent with `NULL` input behavior in nearly every other function and API.

If an argument is `NULL`, that means it's unknown and for consistency the only sane thing to do would be to return `NULL` in such cases.

Now, it's important to have convenience APIs, for example for removing all NULLs in an array.

I propose that we make the semantics of Ibis's `Array.remove` match those of Trino (which returns `NULL` when `NULL` is the element to search for), and add a new API `compress` (?) that strips all nulls from an array.

NickCrews:
@cpcloud just to be clear, currently `ibis.array([1, None, 2]).remove(2)` results in `[1]`. In this PR I just want to fix it so this results in `[1, None]`. Were you thinking I was trying to make `Array.remove(None)` strip NULLs? I am happy with not supporting that, and expecting users to do `Array.filter(_.notnull())`

cpcloud:
Ah, yep, I thought that's what you had in mind. Glad to see we're in agreement here!

NickCrews:
Yup that's my bad for a vague PR title and no description (I edited it to be more clear after I saw your confusion, you're not being gaslit here, it WAS confusing)

cpcloud (approved):

NickCrews:
Thanks @cpcloud for finishing this up!

PR #9092, opened 2024-04-30
fix(deps): remove pydruid sqlalchemy dependency

cpcloud:
Bump pydruid to get rid of its transitive sqlalchemy dependency


requirements-dev.txt
```diff
@@ -273,7 +273,7 @@ snowflake-connector-python==3.9.0 ; python_version >= "3.9" and python_version <
 sortedcontainers==2.4.0 ; python_version >= "3.9" and python_version < "4.0"
 soupsieve==2.5 ; python_version >= "3.10" and python_version < "3.13"
 sphobjinv==2.3.1 ; python_version >= "3.10" and python_version < "3.13"
-sqlalchemy==2.0.29 ; python_version >= "3.9" and python_version < "4.0"
+sqlalchemy==2.0.29 ; python_version >= "3.10" and python_version < "3.13"
```
cpcloud, line 276: We still depend on sqlalchemy **for development dependencies only** by way of `jupyter-cache`.

gforsyth (approved):
sweet.

semver question

If we bump our lower bound of `pydruid` by a patch version, does that count as breaking?  I say no.

cpcloud:
Turns out to be spelled out explicitly in semver: https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api.

At most a feature release if there are no public API changes.

PR #9094, opened 2024-05-01
fix: replace NaNs with None in some backends when loading from pandas dataframe

### LINKED ISSUE #8792: bug: several example datasets broken on postgres
### What happened?

```python
[ins] In [1]: import ibis

[ins] In [2]: con = ibis.postgres.connect(
         ...:     user="postgres",
         ...:     password="postgres",
         ...:     host="0.0.0.0",
         ...:     port=5432,
         ...:     database="ibis_testing",
         ...: )

[ins] In [3]: ibis.set_backend(con)

[ins] In [4]: ibis.examples.starwars.fetch()
NumericValueOutOfRange                    Traceback (most recent call last)
Cell In[4], line 1
----> 1 ibis.examples.starwars.fetch()

File ~/github.com/ibis-project/ibis/ibis/examples/__init__.py:85, in Example.fetch(self, table_name, backend)
     81 # TODO: It should be possible to avoid this memtable call, once all
     82 # backends support passing a `pyarrow.Table` to `create_table`
     83 # directly.
     84 obj = ibis.memtable(table)
---> 85 return backend.create_table(table_name, obj, temp=True, overwrite=True)

File ~/github.com/ibis-project/ibis/ibis/backends/postgres/__init__.py:692, in Backend.create_table(self, name, obj, schema, database, temp, overwrite)
    689     else:
    690         table = obj
--> 692     self._run_pre_execute_hooks(table)
    694     query = self._to_sqlglot(table)
    695 else:

File ~/github.com/ibis-project/ibis/ibis/backends/__init__.py:1030, in BaseBackend._run_pre_execute_hooks(self, expr)
   1028 self._define_udf_translation_rules(expr)
   1029 self._register_udfs(expr)
-> 1030 self._register_in_memory_tables(expr)

File ~/github.com/ibis-project/ibis/ibis/backends/sql/__init__.py:257, in SQLBackend._register_in_memory_tables(self, expr)
    255 def _register_in_memory_tables(self, expr: ir.Expr) -> None:
    256     for memtable in expr.op().find(ops.InMemoryTable):
--> 257         self._register_in_memory_table(memtable)

File ~/github.com/ibis-project/ibis/ibis/backends/postgres/__init__.py:155, in Backend._register_in_memory_table(self, op)
    153 with self.begin() as cur:
    154     cur.execute(create_stmt_sql)
--> 155     execute_batch(cur, sql, data, 128)

File /nix/store/xglf45k8zzyfsk3a18c0f8j3kv77gkh7-python3-3.12.2-env/lib/python3.12/site-packages/psycopg2/extras.py:1216, in execute_batch(cur, sql, argslist, page_size)
   1214 for page in _paginate(argslist, page_size=page_size):
   1215     sqls = [cur.mogrify(sql, args) for args in page]
-> 1216     cur.execute(b";".join(sqls))

NumericValueOutOfRange: bigint out of range
```

`penguins` is also broken.  For `starwars`, at least, the issue is that the `height` column is defined as `bigint` when it should _definitely_ be a float.

main @ b3e27eb9d

postgres

### Relevant log output

_No response_

### Code of Conduct

chloeh13q:
## Description of changes

Examples were broken on MySQL backend and PySpark backend when there are null values in a numeric column. 

Druid, PySpark, RW don't support examples.

- Exasol - did not test
- Flink - broken
- Impala - did not test
- MSSQL - broken #9095 
- MySQL - fixed
- Oracle - did not test
- PostgreSQL - fixed
- Snowflake - did not test

## Issues closed

#8792

### DIFF
--- a/ibis/backends/postgres/__init__.py
+++ b/ibis/backends/postgres/__init__.py
@@ -12,8 +12,10 @@ from typing import TYPE_CHECKING, Any, Callable
 from urllib.parse import parse_qs, urlparse
 
 import numpy as np
+import pandas as pd
 import sqlglot as sg
 import sqlglot.expressions as sge
+from pandas.api.types import is_float_dtype
 
 import ibis
 import ibis.common.exceptions as com
@@ -146,7 +148,14 @@ class Backend(SQLBackend, CanListCatalog, CanCreateDatabase, CanCreateSchema):
             columns = schema.keys()
             df = op.data.to_frame()
             # nan gets compiled into 'NaN'::float which throws errors in non-float columns
-            df = df.replace(np.nan, None)
+            # In order to hold NaN values, pandas automatically converts integer columns
+            # to float columns if there are NaN values in them. Therefore, we need to convert
+            # them to their original dtypes (that support pd.NA) to figure out which columns
+            # are actually non-float, then fill the NaN values in those columns with None.
+            convert_df = df.convert_dtypes()
+            for col in convert_df.columns:
+                if not is_float_dtype(convert_df[col]):
+                    df[col].replace(np.nan, None, inplace=True)
 
             data = df.itertuples(index=False)
             cols = ", ".join(
--- a/ibis/backends/postgres/__init__.py
+++ b/ibis/backends/postgres/__init__.py
@@ -155,7 +155,7 @@ class Backend(SQLBackend, CanListCatalog, CanCreateDatabase, CanCreateSchema):
             convert_df = df.convert_dtypes()
             for col in convert_df.columns:
                 if not is_float_dtype(convert_df[col]):
-                    df[col].replace(np.nan, None, inplace=True)
+                    df[col] = df[col].replace(np.nan, None)
 
             data = df.itertuples(index=False)
             cols = ", ".join(
--- a/ibis/backends/postgres/__init__.py
+++ b/ibis/backends/postgres/__init__.py
@@ -11,6 +11,7 @@ from operator import itemgetter
 from typing import TYPE_CHECKING, Any, Callable
 from urllib.parse import parse_qs, urlparse
 
+import numpy as np
 import sqlglot as sg
 import sqlglot.expressions as sge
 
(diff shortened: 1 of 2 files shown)


ibis/backends/postgres/__init__.py
```diff
@@ -144,6 +145,9 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
 
             columns = schema.keys()
             df = op.data.to_frame()
+            # nan gets compiled into 'NaN'::float which throws errors in non-float columns
+            df = df.replace(np.nan, None)
+
```
gforsyth, line 150: Can we limit this replacement to columns that aren't already floating point?  I think replacing a pandas NaN with `NULL` makes sense for non-float columns, but I'm wary of doing it across the board

gforsyth (commented):
Thanks for putting this in @chloeh13q !  I'm on board with trying this out -- I think for `mysql`, where `nan` isn't allowed at all, this is a good solution.
For postgres, we should be a little more specific in the conversion.

I also think this could work for `mssql` and a few other backends that currently can't load a large chunk of the examples.

chloeh13q:
>For postgres, we should be a little more specific in the conversion.
Sounds good.

>I also think this could work for mssql and a few other backends that currently can't load a large chunk of the examples.
I think mssql examples are broken for a different reason, at least on my local; I filed a ticket #9095 yesterday and I'm getting `Incorrect syntax` on all of the examples. Do you have other backends in mind? Or you just meant this in a generic sense

chloeh13q:
I'm having trouble spinning up testing some of the backends so I'm not able to test whether the examples work. But I can confirm that the examples are now fixed in MySQL and postgres with this PR!

gforsyth:
xref #9110

gforsyth (approved):
LGTM

cpcloud:
Do we have some tests for this? How do we know this won't regress?

gforsyth:
We should enable _some_ of the examples tests for postgres and mysql (I don't think we need to run all of them)

PR #9096, opened 2024-05-01
feat(api): move from .case() to .cases()

NickCrews:
Redo of #7914 (with substantive changes) and  #9039 (merely switching the base repo to the correct one, my fork)

I will try to keep the summary below up to date with what this PR does:

## Breaking change to `Value.cases()`

Before the API was

```python
def cases(
        self,
        case_result_pairs: Iterable[tuple[ir.BooleanValue, Value]],
        default: Value | None = None,
    ) -> Value:
```

Now it is

```python
def cases(self, *branches: tuple[Value, Value], else_=None) -> Value:
```

In other words, it is now variadic, and the else_ is kwarg. The `else_` naming convention is now consistent with `Value.substitute()`

This will be a hard breaking change for users when they upgrade to 10.0.

Alternatives considered (I am currently going for the simplest but harshest path):
- Some gnarly compact code that detects and translates old-style use of the API, and emits a warning. This was deemed too complex.
- Some slightly less gnarly code that merely detects the old-style API, and errors.

## Soft-deprecates `ibis.case()` and `Value.case()`

Existing users will be able to continue to use these APIs with no problems. There will be no warnings or anything else. However, we will remove these functions from the docs, and make the docstring advise to use the new API.
I want there to be ONE endorsed way of doing these conditionals, so I think only advertising one API is very important to me.

Alternatives considered:
- Hard-deprecation: Add `@util.deprecated(instead="use ibis.cases() instead", as_of="10.0")` to these functions. 
- Hard-removal: remove them entirely in 10.0
- Still possible to do: hide .case() even more from users and tab completion, eg behind `__getattr__()`

See [this comment thread](https://github.com/ibis-project/ibis/pull/9096#discussion_r1661007465) for more discussion.


NickCrews:
Any idea as to why the [datafusion, exasol, and risingwave column tests](https://github.com/ibis-project/ibis/actions/runs/8912639606/job/24476499173?pr=9096) are failing? I still have trouble getting those backends running on my M1 so I can't debug locally very well.

NickCrews:
@cpcloud gentle nudge for a review here :)

NickCrews:
@cpcloud anything I can do to help move this forward/easier to review?

ncclementi:
@NickCrews would you mind fixing the conflicts and CI, then I'll nudge the team to get you a review soon.

NickCrews:
Thanks for the poke. I have a lot of other [PRs in flight](https://github.com/ibis-project/ibis/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+author%3ANickCrews) at the moment with ibis, many of which I think are a lot closer to the finish line than this, I'd love to get those merged first before I add another ball to juggle.

NickCrews:
One thing that should get merged first is #9334, where I fix a datashape bug that is exposed in this PR. Once that is in, then this PR will be simpler.

ibis/backends/pandas/executor.py
```diff
@@ -167,6 +167,8 @@ def visit(cls, op: ops.IsNan, arg):
     def visit(
         cls, op: ops.SearchedCase | ops.SimpleCase, cases, results, default, base=None
     ):
+        if not cases:
+            return default
```
cpcloud, line 171: This is a weird edge case. We should enforce through our pattern system that `cases` is never empty.

ibis/backends/sql/compiler.py
```diff
@@ -954,6 +954,8 @@ def visit_VarianceStandardDevCovariance(self, op, *, how, where, **kw):
     )
 
     def visit_SimpleCase(self, op, *, base=None, cases, results, default):
+        if not cases:
+            return default
```
cpcloud, line 958: `cases` should never be empty.

For example, why would someone ever write this:

```python
t.x.cases(else_="foo")
```

ibis/backends/tests/test_generic.py
```diff
@@ -2027,6 +2024,123 @@ def test_sample_with_seed(backend):
...
+    reason="the int64 RBI column is .to_pandas()ed to an object column, which is incomparable to ints",
+    raises=AssertionError,
+)
+def test_value_cases_column(batting):
+    df = batting.to_pandas()
+    expr = batting.RBI.cases(
+        (5, "five"), (4, "four"), (3, "three"), else_="could be good?"
+    )
+    result = expr.execute()
+    expected = pd.Series(
```
cpcloud, line 2110: Do you need this constructor here? You're passing a sequence to `Counter`, and that should be agnostic to the sequence type.

ibis/backends/tests/test_generic.py
```diff
@@ -2027,6 +2024,123 @@ def test_sample_with_seed(backend):
...
+def test_ibis_cases_column(batting):
+    t = batting
+    df = batting.to_pandas()
+    expr = ibis.cases(
+        (t.RBI < 5, "really bad team"),
+        (t.teamID == "PH1", "ph1 team"),
+        else_=t.teamID,
+    )
+    result = expr.execute()
+    expected = pd.Series(
```
cpcloud, line 2134: Don't think you need the additional constructor here.

ibis/backends/tests/test_generic.py
```diff
@@ -2040,6 +2154,37 @@ def test_substitute(backend):
...
+def test_value_cases_null(con):
+    """CASE x WHEN NULL never gets hit"""
+    e = ibis.literal(5).nullif(5).cases((None, "oops"), else_="expected")
+    assert con.execute(e) == "expected"
+
+
+@pytest.mark.broken("pyspark", reason="raises a ResourceWarning that we can't catch")
+def test_case(con):
+    # just to make sure that the deprecated .case() method still works
+    with pytest.warns(FutureWarning, match=".cases"):
```
cpcloud, line 2175: Each of these calls needs to be in a separate **single** `with pytest.warns` block.

Otherwise you're only asserting that one of the many lines in the block warns.

ibis/expr/api.py
```diff
@@ -1119,56 +1120,76 @@ def interval(
...
+    └────────────────┘
     """
-    return bl.SearchedCaseBuilder()
+    for b in branches:
+        try:
+            condition, result = b
+        except (TypeError, ValueError) as e:
+            raise ValueError(
+                "Each branch must be a tuple of (condition, result)"
+            ) from e
```
cpcloud, line 1190: We really should avoid these kinds of up front checks unless there's a really good reason to do so. Especially ones that catch incredibly generic except types like `ValueError` and `TypeError`.

In this case we should be able to leverage the pattern system to describe the input shape. Can you try doing that?

ibis/expr/operations/generic.py
```diff
@@ -275,11 +275,24 @@ class SimpleCase(Value):
...
-    shape = rlz.shape_like("base")
-
-    def __init__(self, cases, results, **kwargs):
+    def __init__(self, base, cases, results, default):
         assert len(cases) == len(results)
-        super().__init__(cases=cases, results=results, **kwargs)
+
+        for case in cases:
+            if not rlz.comparable(base, case):
+                raise TypeError(
```
cpcloud, line 283: Can you leave this as is? I don't see why these changes are necessary.

ibis/expr/operations/generic.py
```diff
@@ -295,14 +308,14 @@ class SearchedCase(Value):
 
     def __init__(self, cases, results, default):
         assert len(cases) == len(results)
```
cpcloud, line 310: Why are we allowing empty `cases` and `results`?

ibis/expr/operations/generic.py
```diff
@@ -275,11 +275,24 @@ class SimpleCase(Value):
...
+        for case in cases:
+            if not rlz.comparable(base, case):
+                raise TypeError(
+                    f"Base expression {rlz._arg_type_error_format(base)} and "
+                    f"case {rlz._arg_type_error_format(case)} are not comparable"
+                )
+
+        if default.dtype.is_null() and results:
+            default = Cast(default, rlz.highest_precedence_dtype(results))
+        super().__init__(base=base, cases=cases, results=results, default=default)
```
cpcloud, line 290: These changes are too aggressive for this PR IMO. Can you leave the original behavior? We should really really avoid throwing in additional operations in op construction, it makes debugging very difficult.

I question whether this is actually necessary. We should try to leave this out.

ibis/expr/types/generic.py
```diff
@@ -717,19 +719,16 @@ def substitute(
...
-            expr = ibis.case()
-            try:
-                null_replacement = value.pop(None)
-            except KeyError:
-                pass
-            else:
-                expr = expr.when(self.isnull(), null_replacement)
-            for k, v in value.items():
-                expr = expr.when(self == k, v)
+            branches = sorted(value.items())
```
cpcloud, line 722: Why are you sorting here?

ibis/expr/types/generic.py
```diff
@@ -865,99 +864,80 @@ def notnull(self) -> ir.BooleanValue:
...
-        │ M                                   │
-        │ NULL                                │
-        │ NULL                                │
-        │ NULL                                │
-        │ NULL                                │
-        └─────────────────────────────────────┘
-        """
-        import ibis.expr.builders as bl
+    @staticmethod
+    def _norm_cases_args(*args, **kwargs):
```
cpcloud, line 873: I am very opposed to this function. Instead of massaging this to be something close to maintainable, why not just wait until 10.0? Doesn't seem worth the trouble to me.

ibis/expr/types/generic.py
```diff
@@ -968,48 +948,64 @@ def cases(
...
-            builder = builder.when(case, result)
-        return builder.else_(default).end()
+        branches, else_ = self._norm_cases_args(*args, **kwargs)
+
+        branches2 = []
+        for b in branches:
+            b = tuple(b)
+            try:
+                test, result = b
+            except (TypeError, ValueError) as e:
```
cpcloud, line 1000: Duplicating this input munging code is another reason we should try very hard to avoid it and use the pattern system.

ibis/expr/types/numeric.py
```diff
@@ -1143,13 +1142,7 @@ def label(self, labels: Iterable[str], nulls: str | None = None) -> ir.StringVal
...
         └───────┴─────────┘
         """
-        return (
-            functools.reduce(
-                lambda stmt, inputs: stmt.when(*inputs), enumerate(labels), self.case()
-            )
-            .else_(nulls)
-            .end()
-        )
+        return self.cases(*enumerate(labels), else_=nulls)
```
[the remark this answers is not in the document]
cpcloud: Seems reasonable.

cpcloud (changes requested):
Thanks for pushing through!

I still have many concerns about the way things are being done in this PR. There also seem to be a few scope increases (e.g., the cast insertion) that should really be avoided.

ibis/backends/sql/compiler.py
NickCrews: This feels similar to whether or not we should allow empty Table schemas and empty structs.

No one would ever write this literally, but I personally ran into this with some code like `branches = [(case, result) for case, result in branches if not is_bad(result)]`

Supporting empty branches is rarely necessary and able to be solvable externally, but semantically it is well-defined, consistent, and unsurprising: the whole thing evals to `else_`.

I lean towards supporting it, but since it can be solved externally we can limit the scope.

Also, probably I can adjust this to catch the empty case at the `ibis.cases()` level and just return `else_`, and then avoid even creating an ops.SimpleCase

ibis/expr/api.py
NickCrews: I was thinking about if you supply two iterables of different lengths to zip (and unexpected things happen), but now I realize that could never happen here, so I think I can just nuke this careful code you're right.

I want to add a test for `ibis.cases((True, 4), "oops")` to ensure we get some error early, but as long as we do I don't think it needs to be a bespoke error.

ibis/expr/operations/generic.py
NickCrews: consider `ibis.literal("foo").cases((1, "small"), (4, "big"))`. Currently, this only errors at runtime. This check makes this error at construction time.

I can put this in a separate PR, but I think this is a good improvement I'd like to see, what do you think?

ibis/expr/operations/generic.py
NickCrews: Lets keep the disussion of this to the other thread in here

ibis/expr/types/generic.py
NickCrews: that is a great question lol. I'll remove it and see if any tests fail

ibis/expr/types/generic.py
NickCrews: Yeah I would love to drop this. I was just trying to provide an upgrade path. Ideally:
- a user upgrades to 9.2. They now get a deprecation warning, but their code continues to work. There is another API available they can switch to without the deprecation warning.
- They upgrade to 10.0. The old API disappears

Without this compat gnarliness, then
- a user upgrades to 9.2 They see no change
- they upgrade to 10.0. Their code breaks, and won't work until they switch over to the new API.

As a user I definitely prefer the first experience, but if we provide nice actionable error messages then the second experience is definitely bearable. Should we go for the 2nd experience?

Similarly, should we keep the old CaseBuilder stuff around, or should we also hard-delete it all at once and provide an actionable error message?

ibis/expr/operations/generic.py
NickCrews: I was trying to make this match the `__init__` of SearchedCase, but I think your right, I think this can get removed from both!

ibis/backends/tests/test_generic.py
```diff
@@ -2281,6 +2393,34 @@ def test_substitute(backend):
...
+        assert con.execute(ibis.case().when(True, "yes").end()) == "yes"
```
cpcloud, line 2407: Do you actually need to call `execute` here? Doesn't the warning happening in the API call? If so, please remove any API calls that are not necessary for triggering the warning.

ibis/expr/operations/generic.py
cpcloud: Any changes that affect validation and/or type checking or construction of operations need to go in separately from API changes. If you want to make this change please put it in a separate PR.

ibis/backends/sql/compiler.py
cpcloud: Please drop this change so we can merge this PR.

When you introduce a change that adds new behavior like this it's going to make the review take much longer.

The fewer behavioral changes and decisions that are made in a PR, the faster we can merge it. It's much easier to debate something like this when it's not buried inside of another giant changeset.

cpcloud (changes requested):
Still too much going on here. See review comments.

ibis/backends/tests/test_generic.py
NickCrews: you are right, thank you!

NickCrews: oh but before I change this, can you comment at https://github.com/ibis-project/ibis/pull/9096/files/e0fa647ebed6e2fffbd549a68834f2dc8727330c#r1676148277? If we are going to just delete the builder API, then we don't need this entire test

ibis/expr/operations/generic.py
NickCrews: PR ready for review at https://github.com/ibis-project/ibis/pull/9559, thanks!

ibis/expr/types/numeric.py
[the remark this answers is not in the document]
NickCrews: Made a tracking issue at https://github.com/ibis-project/ibis/issues/9612

NickCrews:
Once https://github.com/ibis-project/ibis/pull/9559 lands then I think the scope of this will be significantly reduced and this should be much easier, thank you for your reviews!

ibis/backends/pandas/executor.py
NickCrews: now that we officially disallowed empty cases in https://github.com/ibis-project/ibis/pull/9559, this is no longer needed.

ibis/expr/operations/generic.py
NickCrews: we decided in https://github.com/ibis-project/ibis/pull/9559 to disallow empty cases

ibis/backends/tests/test_generic.py
NickCrews: Actually, this test both 1. tests that FutureWarning happens but also 2. checks that the old API still returns the correct result. So if we keep this test at all (depending on our decision in https://github.com/ibis-project/ibis/pull/9096#discussion_r1676148277), then this .execute() should remain. Resolving this convo, but re-open if you disagree.

ibis/expr/operations/generic.py
NickCrews: fixed in #9559

ibis/expr/types/generic.py
NickCrews: removed it.

ibis/expr/types/generic.py
NickCrews: Also it's good to separate out the possible breaking changes. This is what I propose (which I think you agree with IIUC):
- ibis.case(): we add the deprecation in 10.0, add a test that the old API raises a deprecation but still gives the right result. This isn't too hard to continue support. We fully remove in 11.0
- Value.case(): Same as above
- Value.cases(): remove this `_norm_cases_args()` gnarliness. This will just be hard breaking change for users when they upgrade 9.x to 10.0. No warning or upgrade path, sorry folks.

NickCrews:
OK, this is much simpler now, there are no functionality changes, only API changes.

The one question in the air is what to do with the old APIS. Can you respond in [the thread](https://github.com/ibis-project/ibis/pull/9096#discussion_r1676148277) and then I can adjust things?

jcrist:
Going through our PR backlog - @cpcloud, can you please give this another review when you have a moment?

ibis/expr/types/generic.py
cpcloud: Yes, that is correct.

I will look into making these changes in this PR.

cpcloud:
Thinking about this PR some more, I'm not sure why we're going to all this trouble.

Do we **really** need to remove the existing `case` API? What problems is it causing or what additional maintenance does it add?

Right now `cases` is implemented in terms of the more primitive `case`, which is much less complex than what is being proposed here.

I think I am now -1 on this PR altogether. It seems like it might be fine to break `cases` and make it variadic for 10, but removing `case` seems wholly unecessary.

ibis/tests/expr/test_case.py
```diff
@@ -48,106 +48,71 @@ def test_ifelse_function_deferred(table):
...
+    expr4 = ibis.cases(
+        (table.a == 5, table.f),
+        (table.b == 128, table.b * 2),
+        (table.c == 1000, table.e),
+        else_=_.d,
+    ).resolve(table)
+
+    assert (
+        repr(deferred)
+        == "cases(((_.a == 5), <column[float64]>), ((_.b == 128), <column[int32]>), ((_.c == 1000), <column[float32]>), else_=<column[int64]>)"
```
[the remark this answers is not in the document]
NickCrews: I agree, I removed it in the commit I just pushed.

ibis/expr/types/generic.py
NickCrews: In the commit I just pushed, I think we end up with the behavior above:
- removed the `_norm_cases_args()` gnarliness
- kept the deprecation warnings for ibis.case() and Value.case(). I also am fine with removing the deprecation, letting past users keep using them with no problem, but just not surfacing them in the docs or otherwise advertising them.

NickCrews:
I'm still motivated to get this merged. I am trying to keep the original post at the top of this PR in sync with the changes, to provide a good summary of the changes, why we are doing them, and alternatives considered. I think I have mitigated some of your concerns above. Please read that and see what you think? Thanks!

NickCrews:
@cpcloud gentle ping, whenever you get the chance :)

cpcloud (approved):

NickCrews:
@cpcloud looks like the docs CI is failing just for permissions issues, so I think this is good to merge!

cpcloud:
Please do not merge anything that has CI failures.

This job in particular is testing whether the documentation can be built, and that definitely is a hard requirement fro merging a PR.

ibis/expr/types/generic.py
```diff
@@ -967,48 +897,54 @@ def cases(
...
+        │     8 │ /      │     4 │     2.0 │
+        │     9 │ bogus  │     5 │    NULL │
+        │    10 │ NULL   │     6 │    NULL │
+        └───────┴────────┴───────┴─────────┘
         """
-        builder = self.case()
-        for case, result in case_result_pairs:
-            builder = builder.when(case, result)
-        return builder.else_(default).end()
+        if not branches:
```
cpcloud, line 942: If this is a requirement, then enforce it in the signature. I'll push up a change.

cpcloud:
I put up #10290 to address the docs build failure.

docs/posts/ci-analysis/index.qmd
```diff
@@ -203,14 +203,12 @@ Let's also give them some names that'll look nice on our plots.
...
 ).mutate(
-    improvements=(
-        _.raw_improvements.case()
-        .when(0, "None")
-        .when(1, "Poetry")
-        .when(2, "Poetry + Team Plan")
-        .else_("NA")
-        .end()
-    ),
+    improvements=_.raw_improvements.cases(
```
cpcloud, line 206: Most of the files in this directory have pre-computed outputs, and if you change them, they'll rerun.

In this case, this causes a failure, because it's trying to authenticate to BigQuery (I'm guessing browser-based auth is first in the credential chain, and thus the message about a missing browser).

Even if the job **did** have permission to render this quarto document, it would **still fail**, because we verify that for documents that are "frozen" that they aren't being re-rendered unintentionally.

When stuff like this fails, it's not something that should really ever just be merged, hence the `required` property of the docs build check.

NickCrews: OK, so the lesson for me is, if I change something in docs/posts/, I should re-run it locally before pushing. for this bigquery notebook, I won't be able to do this since I don't have creds, but for the rest I should be good to go

ibis/expr/types/generic.py
NickCrews: Nice!

NickCrews:
@cpcloud I added the `breaking change` label to this PR, since it does hard-break Value.cases(), as described in the OP at the top.

cpcloud:
Ah oops. Well, we'll have to add it manually after the release as it's already merged.

PR #9097, opened 2024-05-01
feat(api): support `type` arg to ibis.null()

NickCrews:
This is mostly a UX nicety. Instead of needing to do `ibis.null().cast(int)`, you can do it as `ibis.null(int)`.

pandas tests were breaking due to not being resilient to None scalars. So I fixed a bunch of affected Ops, not just the ones that I needed to change.

### DIFF
--- a/ibis/backends/pandas/kernels.py
+++ b/ibis/backends/pandas/kernels.py
@@ -348,7 +348,7 @@ _generic = {
 }
 
 
-def none_proof(func):
+def none_safe(func):
     def wrapper(*args, **kwargs):
         if any(map(isnull, args)):
             return None
@@ -358,7 +358,7 @@ def none_proof(func):
 
 
 generic = {
-    **{k: none_proof(v) for k, v in _generic.items()},
+    **{k: none_safe(v) for k, v in _generic.items()},
     ops.IsNull: pd.isnull,
     ops.NotNull: pd.notnull,
     ops.IsInf: np.isinf,
--- a/ibis/backends/pandas/kernels.py
+++ b/ibis/backends/pandas/kernels.py
@@ -282,7 +282,8 @@ reductions = {
     ops.ArrayCollect: lambda x: x.tolist(),
 }
 
-generic = {
+
+_generic = {
     ops.Abs: abs,
     ops.Acos: np.arccos,
     ops.Add: operator.add,
@@ -315,8 +316,6 @@ generic = {
     ops.IntervalFloorDivide: operator.floordiv,
     ops.IntervalMultiply: operator.mul,
     ops.IntervalSubtract: operator.sub,
-    ops.IsInf: np.isinf,
-    ops.IsNull: pd.isnull,
     ops.Less: operator.lt,
     ops.LessEqual: operator.le,
     ops.Ln: np.log,
@@ -327,7 +326,6 @@ generic = {
     ops.Negate: lambda x: not x if isinstance(x, (bool, np.bool_)) else -x,
     ops.Not: lambda x: not x if isinstance(x, (bool, np.bool_)) else ~x,
     ops.NotEquals: operator.ne,
-    ops.NotNull: pd.notnull,
     ops.Or: operator.or_,
     ops.Power: operator.pow,
     ops.Radians: np.radians,
@@ -348,6 +346,24 @@ generic = {
     ops.StringJoin: lambda xs, sep: reduce(lambda x, y: x + sep + y, xs),
     ops.Log: lambda x, base: np.log(x) if base is None else np.log(x) / np.log(base),
 }
+
+
+def none_proof(func):
+    def wrapper(*args, **kwargs):
+        if any(map(isnull, args)):
+            return None
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+generic = {
+    **{k: none_proof(v) for k, v in _generic.items()},
+    ops.IsNull: pd.isnull,
+    ops.NotNull: pd.notnull,
+    ops.IsInf: np.isinf,
+}
+
 
 columnwise = {
     ops.Clip: lambda df: df["arg"].clip(lower=df["lower"], upper=df["upper"]),
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -59,10 +59,7 @@ def test_null_literal(con, backend):
         backend_name = backend.name()
         assert con.execute(expr.typeof()) == NULL_BACKEND_TYPES[backend_name]
 
-    with pytest.raises(AttributeError):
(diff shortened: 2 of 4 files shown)


ibis/backends/tests/test_generic.py
```diff
@@ -53,13 +53,32 @@
...
+    reason="https://github.com/ibis-project/ibis/issues/9109",
+    raises=AssertionError,
+)
+def test_null_literal_typed(con, backend):
+    expr = ibis.null(bool)
+    assert pd.isna(con.execute(expr))
+    assert pd.isna(con.execute(expr.negate()))
+    assert pd.isna(con.execute(expr.cast(str).upper()))
+    with pytest.raises(AttributeError):
+        expr.upper()
```
cpcloud, line 80: I don't think this is necessary.

ibis/backends/tests/test_generic.py
```diff
@@ -53,13 +53,32 @@
...
+    assert pd.isna(con.execute(expr))
 
     with contextlib.suppress(com.OperationNotDefinedError):
         backend_name = backend.name()
         assert con.execute(expr.typeof()) == NULL_BACKEND_TYPES[backend_name]
 
+    with pytest.raises(AttributeError):
+        expr.upper()
+    with pytest.raises(AttributeError):
+        expr.cast(str).max()
```
cpcloud, line 65: These tests that are testing the lack of a particular method aren't that useful and add visual noise to the test suite. Testing that `upper` method doesn't exist on a non-string type is testing that a method named upper was not defined. We just don't define the method and are there many such non-existent methods across the codebase. I don't think it's a good use of code to test that we don't define this or that method.

Since those methods are entirely determined by the expression's dtype--by design--why not just check that the value of `expr.type()` is something in particular?

`max` on `str` is also likely to break, I suspect we'll probably add that at some point in the near future.

ibis/backends/pandas/kernels.py
```diff
@@ -349,6 +347,24 @@ def arbitrary(arg):
     ops.Log: lambda x, base: np.log(x) if base is None else np.log(x) / np.log(base),
 }
 
+
+def none_proof(func):
```
cpcloud, line 351:
```suggestion
def none_safe(func):
```

I think safe is a much more common word to use for this functionality.

cpcloud (changes requested):

ibis/backends/tests/test_generic.py
NickCrews: > why not just check that the value of expr.type() is something in particular?

much better idea. Changed to that.

ibis/expr/types/generic.py
```diff
@@ -2223,9 +2223,23 @@ class NullColumn(Column, NullValue):
 
 
 @public
-def null():
-    """Create a NULL/NA scalar."""
-    return ops.NULL.to_expr()
+def null(type: dt.DataType | str | None = None) -> Value:
+    """Create a NULL/NA scalar.
```
cpcloud, line 2227:
```suggestion
    """Create a NULL scalar.
```

Let's keep terminology consistent where possible.

ibis/expr/types/generic.py
```diff
@@ -2223,9 +2223,23 @@ class NullColumn(Column, NullValue):
...
+
+    By default, the type will be NULLTYPE. This is castable and comparable to any type,
```
cpcloud, line 2229:
```suggestion
    `NULL`s with an unspecified type are castable and comparable to values of any other type,
```

`NULLTYPE` isn't the name of anything in the Ibis codebase. I think it's probably going to be confusing that way.

ibis/expr/types/generic.py
```diff
@@ -2223,9 +2223,23 @@ class NullColumn(Column, NullValue):
...
+    but lacks datatype-specific methods:
+
+    >>> import ibis
+    >>> ibis.null().upper().execute() is None
```
cpcloud, line 2233:
```suggestion
    >>> ibis.null().upper()
```

The `execute` here is unnecessary.

ibis/expr/types/generic.py
```diff
@@ -2223,9 +2223,23 @@ class NullColumn(Column, NullValue):
...
+    Traceback (most recent call last):
+        ...
+    AttributeError: 'NullScalar' object has no attribute 'upper'
+    >>> ibis.null(str).upper().execute() is None
+    True
```
cpcloud, line 2238:
```suggestion
    >>> ibis.null(str).upper()
    None
```

ibis/expr/api.py
```diff
@@ -200,11 +200,25 @@
...
 >>> import ibis
->>> my_null = ibis.NA
->>> my_null.isnull()
+>>> assert ibis.NA.execute() is None
+>>> ibis.NA.isnull().execute()
+True
+
+datatype-specific methods aren't available on `NA`:
+
+>>> ibis.NA.upper().execute() is None  # quartodoc: +EXPECTED_FAILURE
```
cpcloud, line 214:
```suggestion
>>> ibis.NA.upper()  # quartodoc: +EXPECTED_FAILURE
```

ibis/expr/api.py
```diff
@@ -200,11 +200,25 @@
...
+Traceback (most recent call last):
+  ...
+AttributeError: 'NullScalar' object has no attribute 'upper'
+
+Instead, use the typed `ibis.null`:
+
+>>> ibis.null(str).upper().execute() is None
 True
```
cpcloud, line 222:
```suggestion
>>> ibis.null(str).upper()
None
```

cpcloud (changes requested):

ibis/expr/types/generic.py
NickCrews: This needs to stay the same.

Without the .execute(), this just shows the compiled repr: `Uppercase(None): Uppercase(None)`

and if you repr the None directly, it doesn't show up in the output:

```
>>> ibis.null(str).upper().execute()
<nothing here>
```

Therefore I added the `is None` check so you actually see something.

ibis/expr/api.py
NickCrews: I think this is best as-is. See other comment.

ibis/expr/types/generic.py
cpcloud: Since you're calling `execute` anyway, just set `ibis.options.interactive = True` first and then the value will repr as `None`

ibis/expr/api.py
cpcloud: I'll submit a PR to address these.

cpcloud:
@NickCrews If you're going to enable automerge you should use the squash option to avoid introducing a bunch of `chore: apply review suggestions` commits.

PR #9101, opened 2024-05-02
build(ruff): don't use keep-runtime-typing anymore

deepyaman:
## Description of changes

* We don't support Python 3.8 anymore
* We don't use pydantic, FastAPI, etc.
* All of the changes this would have guarded against were made in 25946f9865 anyway

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -440,9 +440,6 @@ unfixable = [
   "F841",   # unused variables
 ]
 
-[tool.ruff.lint.pyupgrade]
-keep-runtime-typing = true
-
 [tool.ruff.lint.isort]
 required-imports = ["from __future__ import annotations"]
 


kszucs:
We actually rely on runtime typing and if we not use the right annotation we immediately break 3.9 support which we actually test against.

deepyaman:
> We actually rely on runtime typing and if we not use the right annotation we immediately break 3.9 support which we actually test against.

Like what, for example? All this does is automatically perform PEP 585 (`List[int]` -> `list[int]`) and PEP 604 (`Union[str, int]` -> `str | int`) rewrites; see https://docs.astral.sh/ruff/settings/#lintpyupgrade.

cpcloud (approved):

cpcloud:
Thanks!

kszucs:
> We actually rely on runtime typing and if we not use the right annotation we immediately break 3.9 support which we actually test against.

Never mind, I was thinking of the following error from 3.8:

```
Python 3.8.19 | packaged by conda-forge | (default, Mar 20 2024, 12:49:57)
[Clang 16.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> list[int]
TypeError: 'type' object is not subscriptable
```

PR #9114, opened 2024-05-03
fix(api): treat `col == None` or `col == ibis.NA` as `col.isnull()`

jcrist:
Previously we coerced `co == None` to `IdenticalTo(col, None)`, but wouldn't do so for `ibis.NA` (resulting in that compiling as `Equals(col, None)` (which would result in incorrect SQL). This PR fixes this eager branch to handle any null literal, and compiles it the same as `isnull`/`notnull` (usually `col IS NULL`) rather than the more verbose `col IS NOT DISTINCT FROM NULL` that using `IdenticalTo` would result in.

Other options for this fix would be to:
- Always return `Equals` and use a rewrite rule to rewrite these cases. I opted against this since we'd have to have the same detection logic later, and this rewrite rule would be applied for every backend. Honestly having `Equals.__new__`/`NotEquals.__new__` optionally return `IsNull`/`NotNull` would be cleaner IMO, but that's harder to do with our operation infrastructure.
- Always return `Equals`, and branch when compiling to generate the proper SQL there. This is a bit harder to do, and also would need to be implemented for dataframe backends so `col == None` would still use the proper `isnull` call.

### DIFF
--- a/ibis/tests/expr/test_value_exprs.py
+++ b/ibis/tests/expr/test_value_exprs.py
@@ -352,16 +352,14 @@ def test_notnull(table):
 
 @pytest.mark.parametrize(
     "value",
-    [
-        param(lambda: None, id="none"),
-        param(lambda: ibis.NA, id="NA"),
-        param(lambda: ibis.literal(None, type="int32"), id="typed-null"),
-    ],
+    [None, ibis.NA, ibis.literal(None, type="int32")],
+    ids=["none", "NA", "typed-null"],
 )
 def test_null_eq_and_ne(table, value):
-    other = value()
-    assert (table.a == other).equals(table.a.isnull())
-    assert (table.a != other).equals(table.a.notnull())
+    assert (table.a == value).equals(table.a.isnull())
+    assert (value == table.a).equals(table.a.isnull())
+    assert (table.a != value).equals(table.a.notnull())
+    assert (value != table.a).equals(table.a.notnull())
 
 
 @pytest.mark.parametrize("column", ["e", "f"], ids=["float32", "double"])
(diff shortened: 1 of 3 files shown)


cpcloud:
How would someone compute a `NULL` preserving equality with this change? It seems like it would no longer be possible to do that.

For the semantics you've defined here, you can already do this with `a.identical_to(NA)`.

cpcloud:
Oh, it looks like we're doing that already, nevermind!

ibis/tests/expr/test_value_exprs.py
```diff
@@ -350,6 +350,20 @@ def test_notnull(table):
...
+    [
+        param(lambda: None, id="none"),
+        param(lambda: ibis.NA, id="NA"),
+        param(lambda: ibis.literal(None, type="int32"), id="typed-null"),
+    ],
+)
+def test_null_eq_and_ne(table, value):
+    other = value()
+    assert (table.a == other).equals(table.a.isnull())
+    assert (table.a != other).equals(table.a.notnull())
```
cpcloud, line 364:
```suggestion
        param(None, id="none"),
        param(ibis.NA, id="NA"),
        param(ibis.literal(None, type="int32"), id="typed-null"),
    ],
)
def test_null_eq_and_ne(table, value):
    assert (table.a == value).equals(table.a.isnull())
    assert (table.a != value).equals(table.a.notnull())
```

Doesn't seem like the `lambda`s are necessary here.

cpcloud (approved):

cpcloud (changes requested):
Seems like the reversed operand case doesn't work. Looking into it.

cpcloud (approved):

PR #9115, opened 2024-05-03
feat(ux): improve error message on unequal schemas during set ops

NickCrews:
I got tired of `ibis.union()` failing but only saying "the schemas are different"

The relevant tests are in ibis/tests/expr/test_set_operations.py, looks like the bases are already covered pretty well, don't think I need to add any more

### DIFF
--- a/ibis/common/collections.py
+++ b/ibis/common/collections.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any, Generic, TypeVar
 from public import public
 
 from ibis.common.bases import Abstract, Hashable
+from ibis.common.exceptions import ConflictingValuesError
 
 if TYPE_CHECKING:
     from typing_extensions import Self
@@ -202,12 +203,13 @@ class MapSet(Mapping[K, V]):
         # A key-value pair is conflicting if the key is the same but the value is
         # different.
         common_keys = self.keys() & other.keys()
-        for key in common_keys:
-            left, right = self[key], other[key]
-            if left != right:
-                raise ValueError(
-                    f"Conflicting values for key `{key}`: {left} != {right}"
-                )
+        conflicts = {
+            (key, self[key], other[key])
+            for key in common_keys
+            if self[key] != other[key]
+        }
+        if conflicts:
+            raise ConflictingValuesError(conflicts)
         return common_keys
 
     def __ge__(self, other: collections.abc.Mapping) -> bool:
--- a/ibis/expr/operations/relations.py
+++ b/ibis/expr/operations/relations.py
@@ -10,7 +10,11 @@ from public import public
 import ibis.expr.datashape as ds
 import ibis.expr.datatypes as dt
 from ibis.common.annotations import attribute
-from ibis.common.collections import FrozenDict, FrozenOrderedDict
+from ibis.common.collections import (
+    ConflictingValuesError,
+    FrozenDict,
+    FrozenOrderedDict,
+)
 from ibis.common.exceptions import IbisTypeError, IntegrityError, RelationError
 from ibis.common.grounds import Concrete
 from ibis.common.patterns import Between, InstanceOf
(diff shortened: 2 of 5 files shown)


ibis/expr/operations/relations.py
```diff
@@ -288,10 +288,24 @@ class Set(Relation):
...
-        elif left.schema.names != right.schema.names:
+        err_msg = "Table schemas must be equal for set operations."
+        try:
+            missing_from_left = right.schema - left.schema
+            missing_from_right = left.schema - right.schema
+        except ValueError as e:
+            if "Conflicting values" in str(e):
+                raise RelationError(err_msg) from e
+            else:
+                raise
```
cpcloud, line 299: We should really avoid this sort of exception hacking unless it's required.

In no way are the messages coming out of exceptions stable, including messages coming from library internals.

If this `ValueError` is something we control, then the appropriate thing to do here is to raise a more specific exception in the schema subtract operation and then catch that specific type here.

Raising `ValueError` is about as useful as other mostly-useless idioms like raising `Exception` or `RuntimeError`, which means catching `ValueError` is very heavy handed (and requires checking for specificity using string hacking).

ibis/expr/schema.py
```diff
@@ -67,7 +67,7 @@ def _name_locs(self) -> dict[str, int]:
         return {v: i for i, v in enumerate(self.names)}
 
     def equals(self, other: Schema) -> bool:
-        """Return whether `other` is equal to `self`.
+        """Return whether `other` is equal to `self`. Order matters.
```
cpcloud, line 70:
```suggestion
        """Return whether `other` is equal to `self`.
        
        The order of fields in the schema is taken into account when computing equality.
```

cpcloud (changes requested):

ibis/expr/schema.py
NickCrews: lol yeah I was totally being lazy here :)

ibis/expr/operations/relations.py
NickCrews: OK, check out the new way I do it.

ibis/common/collections.py
```diff
@@ -150,6 +150,16 @@ def __eq__(self, other):
         return dict(self.items()) == dict(other.items())
 
 
+class ConflictingValuesError(ValueError):
```
cpcloud, line 153:
```suggestion
class ConflictingValuesError(Exception):
```

I understand this breaks code that looks like `except ValueError:` but our exception types are not (yet) part of the public API.

cpcloud: Actually nevermind, let's leave it as a subclass for now.

ibis/expr/operations/relations.py
NickCrews: well, actually now I'm thinking about this more. If we make a new error, then if a user wants to do `schema1 | schema2`, then they will need to import a special ibis exception so they can catch it. I actually am tempted to [keep the error as some builtin exception](https://youtu.be/o9pEzgHorH0?si=3bk4RFVyR-9x9ZjP&t=617) and then expect us/users to match via the string.

cpcloud: I am very against reusing standard library exceptions unless it unambiguously makes sense.

`ValueError` IMO almost never makes sense to reuse. Generally speaking exceptions that map to a huge variety of errors are terrible candidates for reuse, because any downstream code that uses a `ValueError` exception handler has now opted into a potentially massive set of error conditions. It's really not that different from catching `Exception`.

That video isn't about matching builtins AFAICT, it's about not naming exceptions redundantly.

I am definitely on board with avoiding the `SomethingWentWrongException` in favor of `SomethingWentWrong`.

NickCrews: I linked to the very start of where he starts talking about exceptions, [later](https://youtu.be/o9pEzgHorH0?si=hV8VqoOpinRG8mPv&t=746) he actually gets to his thesis of "stop writing custom exceptions". He talks about that more scattered through the whole talk, you sorta have to listen to the whole thing to get everything (not expecting you to)

Yeah that is the downside to using generic exceptions. I think there are pros and cons of both, which I prefer probably depends on the circumstance. One factor is how likely is the user going to realize something is fallible and try to catch it. If they always wrap the `schema1 | schema2` call in a try/catch, then the bubbling up error isn't really a problem. But if someone forgets to do that, then then genericness is a problem. I might expect with no evidence, that in this case, the vast majority of people would not wrap these set ops in a try/catch (they are probably gonna be data scientists) and thus that makes me more likely to agree with you.

I guess the thing I want to avoid is ibis exporting 20 different errors that someone has to keep track of. I would love it if we could use some other existing ibis Exception class here, or do something to simplify the situation

cpcloud: The other typical approach is to have a generic class of error and then some kind of enum-like object embedded in it that allows someone to differentiate.

This is effectively OOP-ization of the `errno` approach from C programming. `OSError` in the stdlib is good example of this approach.

As a user, I don't really care how many exception types there are. I care about getting specific, actionable feedback about what went wrong. If there are 20 ways in which things can go wrong, I want to know about that. Sure, those could all be just `IbisError`. From an interactive-mode-mostly-end-user perspective, I don't the exception count matters.

As a library author I care about the number of exception types only if they aren't stable. Even in an unstable world where exceptions are changing often, I would much rather have to deal with 20 exception types than have 20 different `if "string" in str(e)`-style checks littered all over my code, liable to break in hard-to-debug ways just because someone decided to fix a typo or capitalize a letter in their error message.

I'm not saying we need a new exception type for every possible error. We should determine whether it's appropriate on a case by case basis and we shouldn't set some arbitrary limit like "no more than 20 custom exception types".

cpcloud:
I'll rebase your PR to pick up the fix for the failing docs build.

ibis/common/collections.py
cpcloud, line 153: Can you put this in our `exceptions` module? Sorry, Jack Diedrich might not approve 😂

cpcloud (approved):

PR #9121, opened 2024-05-05
docs: minor edit to homepage

lostmygithubaccount:
## Description of changes

reading back the homepage, I think using `table_name` in `read_parquet` is confusing. instead, use a `.alias` call later


lostmygithubaccount:
never mind, then I don't like the SQL printed out for the `grouped` table

PR #9123, opened 2024-05-06
docs: use interactive mode instead of execute for typed null docstrings

cpcloud:
Use `ibis.options.interactive = True` in typed-null docstrings instead of calling `.execute()`.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -205,8 +205,8 @@ This is an untyped NULL. If you want a typed NULL, use eg `ibis.null(str)`.
 Examples
 --------
 >>> import ibis
->>> assert ibis.NA.execute() is None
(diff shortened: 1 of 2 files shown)


ibis/expr/api.py
```diff
@@ -218,7 +218,8 @@
 
 Instead, use the typed `ibis.null`:
 
->>> ibis.null(str).upper().execute() is None
+>>> ibis.options.interactive = True
```
jcrist, line 221: Do we need to set this again down here, or is setting it above sufficient?

cpcloud: Oh, good point, I think it should be sufficient to set it once!

jcrist (approved):

PR #9130, opened 2024-05-06
test(examples): add penguins example test

gforsyth:
The palmer penguins data has null valued integers, which when used to
create a `memtable`, leads to pandas casting a column to `float` because
it reads in the nulls as NaN.

We added a fix for this for some backends in #9094.


cpcloud (approved):
Thank you!

Issue #8584, opened 2024-03-07
feat(parquet): figure out convention for multi-file parquet writing


gforsyth:
Hmmm, I remember us talking about this before (sorry, this issue had slipped my mind when we talked about this last week).

Currently all our `to_csv`/`to_parquet` writers produce a single file (rather than a directory of files). There was an open question as to what the expected behavior was for backends where a single-file output is tricky/inefficient/impossible (backends like spark or dask).

Enumerating all the options I can think of:

### 1. `to_csv`/`to_parquet` always outputs a single file

We could fallback to pyarrow or error for backends where this is tricky. This is the current behavior.

### 2. `to_csv`/`to_parquet` always output a directory

Backends that only write a single file would convert to writing a directory with a single file. This would be a breaking change, but should work with any backend's native csv/parquet writer. This would make a common case of writing a small csv of results trickier though.

### 3. `to_csv`/`to_parquet` write the backend-native style

The output style (single file or directory) depends on the backend. This is easy to implement, but would mean different backends would result in different behaviors.

### 4. `to_csv`/`to_parquet` have an option to write a directory

These methods would write to a single file by default, but would have an option to instead write to a directory. :shrug: whether this is a `directory=True` option, or inferred somehow from the input path. This would mean that users using backends that can only efficiently write to directories may need to opt into this support, but would make things explicit about the output type.

```python
t.to_csv("test.csv")  # would error for the spark backend? Or fallback to pyarrow?
t.to_csv("test/", directory=True)  # output a directory can use pyspark native behavior
```

### 5. New methods for writing directories

`to_csv`/`to_parquet` keep their existing behavior, and we add new `to_csv_dir`/`to_parqet_dir` (or much better named) methods for writing directories. Same caveats/questions as 4.

---

Right now I'm leaning towards 4. An extra flag seems fine to me, and I like it better than adding another top-level-method per file format just to support partitioned writing. cc'ing @gforsyth for a 2nd opinion though, since he may remember what conclusions we came to last time this came up.

_Originally posted by @jcrist in https://github.com/ibis-project/ibis/issues/6615#issuecomment-1629759554_

Additionally, I've documented (although this is now at least partly out of date) how a few systems handle various parquet partitioning schemes: https://gist.github.com/gforsyth/8dd4ca981b2beed6ef4db80f5e8afbfd

Opening this issue so we have something to track that isn't a comment in a closed PR.

Probably need to write out a taxonomy of what parquet functionality is supported by each backend natively

[...]

PR #9131, opened 2024-05-06
feat(pyspark): provide a mode option to manage both batch and streaming connections

### LINKED ISSUE #8984: feat(spark): can we support connectors better?

No

In streaming engines, we need to connect to sources and sinks. Flink provides a relatively straightforward API so that all the connector configurations can be passed in as table properties, and everything can be written in SQL (including the definition of watermarks, connector configurations, etc). Because of that, we're able to have a relatively clean API in Ibis to define connectors in the Flink backend.

On the other hand, this is not the case in Spark. There is a lot of code you need to write in a Spark streaming job, including manually parsing the Kafka messages, from binary to string, then to your expected schema. E.g.,
```python
## to set up the source
from pyspark.sql.functions import col, from_json
from pyspark.sql.types import StructField, StructType, TimestampType, LongType, DoubleType, IntegerType

schema = StructType(
    [
        StructField('createTime', TimestampType(), True), 
        StructField('orderId', LongType(), True), 
        StructField('payAmount', DoubleType(), True), 
        StructField('payPlatform', IntegerType(), True), 
        StructField('provinceId', IntegerType(), True),
    ])

streaming_df = session.readStream\
    .format("kafka")\
    .option("kafka.bootstrap.servers", "localhost:9092")\
    .option("subscribe", "payment_msg")\
    .option("startingOffsets","earliest")\
    .load()\
    .select(from_json(col("value").cast("string"), schema).alias("parsed_value"))\
    .select("parsed_value.*")\
    .withWatermark("createTime", "10 seconds")

streaming_df.createOrReplaceTempView("streaming_df")

## this is where ibis code would go

## to set up the sink
result_df = (window_agg
.writeStream
.outputMode("append")
.format("kafka")
.option("checkpointLocation", "checkpoint")
.option("kafka.bootstrap.servers", "localhost:9092")
.option("topic", "sink")
.start())
```

Obviously, this is connector-specific, but Spark also only supports a handful of connectors ootb (for any additional ones that are not supported, the user can implement their own logic).

Spark does not support defining watermarks in SQL so the watermark definition has to be written in Spark DataFrame API. The motivation is to make the UX better so that the user 1) doesn't need to write a bunch of boilerplate code and 2) doesn't need to know where the pyspark code stops and the ibis code begins.

Support connectors better in Spark. Still need some more design around how we're going to do this exactly, and to what extent we can do this, but, e.g., a starting point could be to wrap all the options in a kwarg.

main

Spark

### Code of Conduct

chloeh13q:
## Description of changes

Provide a mode option to manage both batch and streaming connections

## Issues closed

#8984


chloeh13q:
Along the lines of #8584, it's worth pointing out the fact that pyspark in _streaming_ mode requires all read and write paths to be directories rather than files. I think Impala has something similar

chloeh13q:
xref: https://github.com/ibis-project/testing-data/pull/9

jcrist (commented):
I think a `mode="streaming"`/`mode="batch"` kwarg to `connect` makes sense. However, is it possible to implement this while keeping everything within a single `Backend` class?

- Our current magical `ibis.<backend>.connect` method isn't really setup to handle things the way you're doing here (as you've found by having to reimplement it).
- It looks like a lot of the methods in the streaming backend are pretty similar to the batch implementations - uniting them and branching when necessary feels doable. If necessary, we can always create a small abstraction layer internal to the `Backend` class if needed.
- For the new `read_kafka`/`to_kafka` methods, raising `NotImplementedError` (or something) when in batch mode seems fine.

ibis/backends/pyspark/__init__.py
```diff
@@ -849,10 +910,266 @@ def to_pyarrow_batches(
...
         pa = self._import_pyarrow()
         pa_table = self.to_pyarrow(
             expr.as_table(), params=params, limit=limit, **kwargs
         )
         return pa.ipc.RecordBatchReader.from_batches(
             pa_table.schema, pa_table.to_batches(max_chunksize=chunk_size)
         )
+
+    @util.experimental
+    def read_kafka(
```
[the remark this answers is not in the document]
chloeh13q: Sounds good. I will split these into 3 separate PRs.

ibis/backends/pyspark/tests/test_streaming/conftest.py
```diff
@@ -0,0 +1,32 @@
...
+
+def write_to_memory(self, expr, table_name):
+    if self.mode == "batch":
+        raise NotImplementedError
+    df = self._session.sql(expr.compile())
+    df.writeStream.format("memory").queryName(table_name).start()
+
+
+@pytest.fixture(scope="session")
+def con(data_dir, tmp_path_factory, worker_id):
```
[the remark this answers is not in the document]
chloeh13q: I think I moved the `con` fixture into a subdirectory because I was having trouble loading the data correctly... But I'm moving the new methods into a separate PR so we can address it there.

ibis/backends/pyspark/__init__.py
```diff
@@ -662,8 +686,13 @@ def read_parquet(
             The just-registered table
 
         """
-        source = util.normalize_filename(source)
-        spark_df = self._session.read.parquet(source, **kwargs)
+        if self.mode == "streaming":
+            raise NotImplementedError(
+                "Pyspark in streaming mode does not support direction registration of parquet files. "
+                "Please use `read_parquet_directory` instead."
```
chloeh13q, line 692: We don't technically have `read_parquet_directory` implemented yet but I figured that it may be okay because it's going in in a separate PR? But I can delete this if needed.

ibis/backends/pyspark/tests/test_import_export.py
```diff
@@ -0,0 +1,19 @@
...
+    [
+        methodcaller("read_delta", path="test.delta"),
+        methodcaller("read_csv", source_list="test.csv"),
+        methodcaller("read_parquet", path="test.parquet"),
+        methodcaller("read_json", source_list="test.json"),
+    ],
+)
+@pytest.mark.notyet(["pyspark"], raises=NotImplementedError)
+def test_streaming_import_not_implemented(con_streaming, method):
+    method(con_streaming)
```
jcrist, line 19: The `notyet`/`notimpl`/... marks shouldn't be used in backend-specific tests, they're more for indicating levels of support for the cross-backend test suite. Instead you should use `pytest.raises` to check that the error is raised appropriately.

```suggestion
def test_streaming_import_not_implemented(con_streaming, method):
    with pytest.raises(NotImplementedError):
        method(con_streaming)
```

jcrist (commented):
One small fixup, otherwise this looks good to me!

ibis/backends/pyspark/tests/test_import_export.py
chloeh13q: oh uh okay... [gil had told me to use the `notyet` marker in a flink test](https://github.com/ibis-project/ibis/pull/8751#discussion_r1538137244) - should that have been `pytest.raises`?

jcrist: In that case you moved an `xfail` to a `notyet` mark for testing an error raised *due to an upstream issue*. If/when flink supports that SQL that test will start passing - using a `notyet` marker there (and the reason indicated) lets us know that at some point in the future we may expect that test to start passing. I personally would have been fine with an `xfail` there, but :shrug:.

In contrast, here we're testing that an error we raise in our own code is actually raised. `pytest.raises` is the proper mechanism for testing that, since it explicitly states what's being tested.

chloeh13q: I see, that makes sense. Thanks for explaining! Just fixed.

jcrist (approved):
:shipit:

PR #9133, opened 2024-05-06
docs(descheming): move callout note outside of parameter description

gforsyth:
Callouts inside of a parameter description aren't being rendered in the
docs, so I've moved this callout into the main body of the docstring so
it renders correctly.

I've also added the same callout to the `read_postgres` method on the
DuckDB backend, as postgres users may try to use `schema` in the
hierarchical sense.

xref #8712


jcrist (approved):
LGTM, failure is (I think) unrelated.

Issue #9135, opened 2024-05-06
bug: hash() in some backends returns a UINT64, but ops.Hash has dtype int64


NickCrews:
### What happened?

```python
import ibis
ibis.options.interactive = True

t = ibis.memtable({"x": ["BATMAN"]})
t = t.mutate(y=t.x.hash())
t.cache()
```

The duckdb hash of "BATMAN" happens to be 18165473391510152529, which is a uint64 too big to fit into a int64. When we call .cache(), the backend first does a DDL to create a table, but it makes the dtype of the column int64. Then when it tries to insert, it fails of course with `Conversion Error: Type UINT64 with value 18165473391510152529 can't be cast because the value is out of range for the destination type INT64`.

postgres's hash for instance returns an int, not a uint. I assume other backends do different things.

I'm not sure what to do here: should we keep the ops.Hash always having a int64 type, and then fix duckdb's `visit_Hash()` to do some casting?

EDIT: yes, we get the right thing, and it should be pretty performant to do something like `select HASH('BATMAN')::BIT::int64;`

### What version of ibis are you using?

main

### What backend(s) are you using, if any?

duckdb

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jcrist:
A quick low-effort survey of our implementations:

```
duckdb: uint64
clickhouse: uint64
snowflake: int64
mssql: int (doesn't say if signed or not)
bigquery: int64
spark: int64
impala: int64
oracle: uint32
```

Given this, I think coaxing to an `int64` seems fine, but we'll also probably need to do this for clickhouse too.

PR #9136, opened 2024-05-06
fix: convert the uint64's from some backends' hash() to the desired int64

NickCrews:
fixes https://github.com/ibis-project/ibis/issues/9135

I don't love the logic that does the conversion. It seems more complicated than it needs to be, and I don't like how I had to implement it once for each backend, but maybe there is no universal way to do this. If you can think of a cleaner way to do this or sidestep the need entirely that would be awesome

### DIFF
--- a/ibis/backends/clickhouse/compiler.py
+++ b/ibis/backends/clickhouse/compiler.py
@@ -235,7 +235,7 @@ class ClickHouseCompiler(SQLGlotCompiler):
         return self.f.intDivOrZero(arg, self.f.abs(arg))
 
     def visit_Hash(self, op, *, arg):
-        return self.f.sipHash64(arg)
+        return self.f.reinterpretAsInt64(self.f.sipHash64(arg))
 
     def visit_HashBytes(self, op, *, arg, how):
         supported_algorithms = {
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -1206,7 +1206,8 @@ def execute_union(op, **kw):
 
 @translate.register(ops.Hash)
 def execute_hash(op, **kw):
-    return translate(op.arg, **kw).hash()
+    # polars' hash() returns a uint64, but we want to return an int64
+    return translate(op.arg, **kw).hash().reinterpret(signed=True)
 
 
 def _arg_min_max(op, func, **kw):
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -1439,11 +1439,14 @@ def test_distinct_on_keep_is_none(backend, on):
         "trino",  # checksum returns varbinary
     ]
 )
-def test_hash_consistent(backend, alltypes):
-    h1 = alltypes.string_col.hash().execute(limit=10)
-    h2 = alltypes.string_col.hash().execute(limit=10)
+def test_hash(backend, alltypes):
+    # check that multiple executions return the same result
+    h1 = alltypes.string_col.hash().execute(limit=20)
+    h2 = alltypes.string_col.hash().execute(limit=20)
     backend.assert_series_equal(h1, h2)
-    assert h1.dtype in ("i8", "uint64")  # polars likes returning uint64 for this
+    # check that the result is a signed 64-bit integer, no nulls
+    assert h1.dtype == "i8"
+    assert h1.notnull().all()
 
 
 @pytest.mark.notimpl(["trino", "oracle", "exasol", "snowflake"])
--- a/ibis/backends/clickhouse/tests/snapshots/test_functions/test_hash/out.sql
+++ b/ibis/backends/clickhouse/tests/snapshots/test_functions/test_hash/out.sql
@@ -1,3 +1,3 @@
 SELECT
-  sipHash64("t0"."string_col") AS "Hash(string_col)"
+  reinterpretAsInt64(sipHash64("t0"."string_col")) AS "Hash(string_col)"
 FROM "functional_alltypes" AS "t0"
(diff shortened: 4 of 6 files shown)


ibis/backends/polars/compiler.py
```diff
@@ -1203,9 +1203,19 @@ def execute_union(op, **kw):
     return result
 
 
+def bit_cast_uint64_to_int64(uint: pl.Series) -> pl.Series:
```
NickCrews, line 1206: It feels like there should be a much simpler way to do this or just sidestep the problem completely...

ibis/backends/polars/compiler.py
```diff
@@ -1203,9 +1203,19 @@ def execute_union(op, **kw):
...
+    happy_path = uint.cast(pl.Int64, strict=False)
+    in_range = uint - (1 << 63)
+    casted = in_range.cast(pl.Int64, strict=False)
+    corrected = casted - ((1 << 63) - 1)
+    INT64_MAX = 2**63 - 1
+    return pl.when(uint <= INT64_MAX).then(happy_path).otherwise(corrected)
```
jcrist, line 1212: This doesn't handle wraparound conversion the way you probably intended (yours changes the total ordering at the split point) - ideally converting a `uint64` to an `int64` is just a cast and an addition of `INT64_MIN`. 

Try this instead:

```suggestion
    sub_then_cast = (uint - 2 ** 63).cast(pl.Int64, strict=False)
    cast_then_sub = uint.cast(pl.Int64, strict=False) + (-2 ** 63)
    return sub_then_cast.fill_null(cast_then_sub)
```

Comparison:

```
┌──────────────────────┬──────────────────────┬──────────────────────┐
│ x                    ┆ yours                ┆ mine                 │
│ ---                  ┆ ---                  ┆ ---                  │
│ u64                  ┆ i64                  ┆ i64                  │
╞══════════════════════╪══════════════════════╪══════════════════════╡
│ 18446744073709551615 ┆ 0                    ┆ 9223372036854775807  │
│ 18446744073709551614 ┆ -1                   ┆ 9223372036854775806  │
│ 18446744073709551613 ┆ -2                   ┆ 9223372036854775805  │
│ 9223372036854775809  ┆ -9223372036854775806 ┆ 1                    │
│ 9223372036854775808  ┆ -9223372036854775807 ┆ 0                    │
│ 9223372036854775807  ┆ 9223372036854775807  ┆ -1                   │
│ 2                    ┆ 2                    ┆ -9223372036854775806 │
│ 1                    ┆ 1                    ┆ -9223372036854775807 │
│ 0                    ┆ 0                    ┆ -9223372036854775808 │
└──────────────────────┴──────────────────────┴──────────────────────┘
```

ibis/backends/clickhouse/compiler.py
```diff
@@ -242,8 +242,20 @@ def visit_Sign(self, op, *, arg):
...
+        casted = self.f.accurateCastOrNull(
+            in_range, self.type_mapper.to_string(dt.Int64())
+        )
+        corrected = casted - ((1 << 63) - 1)
+        INT64_MAX = 2**63 - 1
+        return self.if_(uint <= INT64_MAX, happy_path, corrected)
+
     def visit_Hash(self, op, *, arg):
-        return self.f.sipHash64(arg)
+        return self.bit_cast_uint64_to_int64(self.f.sipHash64(arg))
```
jcrist, line 258: Clickhouse has a builtin function for doing reinterpret casts, no need for mucking around here:

```suggestion
        return self.f.reinterpretAsInt64(self.f.sipHash64(arg))
```

ibis/backends/tests/test_generic.py
```diff
@@ -1442,7 +1442,24 @@ def test_hash_consistent(backend, alltypes):
...
+        "mysql",  # CHECKSUM TABLE but not column
+        "trino",  # checksum returns varbinary
+    ]
+)
+def test_hash_is_int64(alltypes):
+    # some backend's hash() return a uint64, not a int64.
+    # Check that these backends successfully cast the result to int64
+    # so it is consistent with the other backends.
+    res = alltypes.string_col.hash().cast(str).cast("int64").execute(limit=10)
+    assert res.dtype == "i8"
```
jcrist, line 1462: I'd just combine this test with the above, since it's effectively one more line in the existing test (and all the marks are the same).

```suggestion
    # Check that dtype is consistently i8
    assert h1.dtype == "i8"
```

ibis/backends/clickhouse/compiler.py
NickCrews: Ah, hallelujah! Thanks!

ibis/backends/polars/compiler.py
NickCrews: Since this is taking a hash as input, the ordering of the input isn't meaningful, so I wasn't trying to maintain ordering. Only trying to maintain the entropy. I should probably rename the function. What do you think, should we care about order? I think we should still consider performance though, so we should break ties using that

jcrist: I don't think we should care about order, just noting that the implementation I posted here does things the same way a reinterpret cast would, while yours does something different. As long as it's a 1:1 mapping between uint64 -> int64 then I don't think we care what it is as long as it's efficient enough.

jcrist: Actually, turns out polars has a builtin for this too - can just do `col.reinterpret(signed=True)`. Much cleaner.

NickCrews: excellent, thank you!

gforsyth:
Hey @NickCrews -- I think the exasol failure is fixed on `main` if you want to rebase this

NickCrews:
hey @gforsyth can you give this a review when you get the chance? Thanks!

jcrist (approved):
Provided tests pass, this lgtm! Thanks for working on this.

NickCrews:
@jcrist you all do the rebase and merge, right? Not squash?

jcrist:
Depends on the commits and the desired changelog entries (multiple user facing changes would need multiple commits to note). Since this is only 1 commit and change, either works tbh.

Issue #8679, opened 2024-03-16
.create_table method does not work with BigQuery if the input table is an ibis table


aemikarafi:
### What happened?

I expect is to just create a new table in BigQuery

### What version of ibis are you using?

8.0.0

### What backend(s) are you using, if any?

BigQuery

### Relevant log output

```sh
#### Environment details

  - OS type and version:
  - Python version: `Python 3.9.18`
  - pip version: `pip 24.0`
  - `google-cloud-bigquery` version: `google-cloud-bigquery==3.19.0`
  - `ibis-framework` version: `ibis-framework==8.0.0`

#### Steps to reproduce
1) Create an ibis table
2) Save the ibis table to BigQuery

#### Code example


from ibis.interactive import *
con = ibis.connect("bigquery://smart-dumb")
con.set_database("data_2023")
new = con.tables.DataCreateTable_Customer_20240125_083529
con.create_table("baba", new)


#### Stack trace

...
/~/.local/lib/python3.9/site-packages/google/cloud/bigquery/job/query.py:1589) if self._query_results is None or not self._query_results.complete:
...
BadRequest: 400 Syntax error: Missing whitespace between literal and alias at [1:26]; reason: invalidQuery, location: query, message: Syntax error: Missing whitespace between literal and alias at [1:26]
```
This happens because the resulting query looks like

```
CREATE TABLE smart-dumb.data_2023.baba AS WITH t0 AS (
...
)
```
there is no back ticks on the name.
I am surprised why that is, because the SELECT * FROM  from source inside has backticks, such as `smart-dumb`.data_2023.DataCreateTable_Customer_20240125_083529

A quick workaround would help.

Thanks!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gforsyth:
Hey @aemikarafi !

I can't reproduce this on `main` and our dataset and project_id also have the same mix of hyphens and underscores:

```python
[ins] In [1]: import ibis

[ins] In [2]: con = ibis.connect("bigquery://ibis-gbq")

[ins] In [3]: con.set_database("ibis_gbq_testing")

[ins] In [4]: astronauts = con.tables.astronauts

[ins] In [5]: con.create_table("astro2", astronauts)
Out[5]: 
DatabaseTable: astro2
  id                       int64
  number                   int64
  nationwide_number        int64
  name                     string
  original_name            string
  sex                      string
  year_of_birth            int64
  nationality              string
  military_civilian        string
  selection                string
[...]

PR #9141, opened 2024-05-07
fix(bigquery): quote all parts of table names

gforsyth:
The BigQuery identifier quoting semantics are bonkers
https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#identifiers

my-table is OK, but not mydataset.my-table

mytable-287 is OK, but not mytable-287a

Just quote everything.

xref #8679

### DIFF
--- a/ibis/backends/bigquery/__init__.py
+++ b/ibis/backends/bigquery/__init__.py
@@ -125,6 +125,27 @@ def _remove_null_ordering_from_unsupported_window(
     return node
 
 
+def _force_quote_table(table: sge.Table) -> sge.Table:
+    """Force quote all the parts of a bigquery path.
+
+    The BigQuery identifier quoting semantics are bonkers
+    https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#identifiers
+
+    my-table is OK, but not mydataset.my-table
+
+    mytable-287 is OK, but not mytable-287a
+
+    Just quote everything.
+    """
+    for key in ("this", "db", "catalog"):
+        if (val := table.args[key]) is not None:
+            if isinstance(val, sg.exp.Identifier) and not val.quoted:
+                val.args["quoted"] = True
+            else:
+                table.args[key] = sg.to_identifier(val, quoted=True)
+    return table
+
+
 class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema):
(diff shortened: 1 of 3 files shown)


ibis/backends/bigquery/__init__.py
```diff
@@ -125,6 +125,27 @@ def _remove_null_ordering_from_unsupported_window(
     return node
 
 
+def _force_quote_table(table: sge.Table) -> sge.Table:
+    """Force quote all the parts of a bigquery path.
+
+    The BigQuery identifier quoting semantics are bonkers
+    https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#identifiers
+
+    my-table is OK, but not mydataset.my-table
```
cpcloud, line 134: What would the `my-table` unqualified with a dataset even mean?

ibis/backends/bigquery/tests/unit/test_client.py
```diff
@@ -30,3 +31,23 @@ def test_parse_project_and_dataset_raises_error():
...
+        "a-7b0a.dev_test_dataset.test_ibis5",
+    ],
+)
+def test_force_quoting(bq_path_str):
+    table = sg.parse_one(bq_path_str, into=sg.exp.Table, read="bigquery")
+    table = _force_quote_table(table)
+
+    for key in ("this", "db", "catalog"):
+        if (val := table.args[key]) is not None:
+            assert val.quoted
```
cpcloud, line 53: can we instead test the expected output as part of the test parameterization?

ibis/backends/bigquery/__init__.py
gforsyth: I don't know -- why would Google include that example in their documentation?

ibis/backends/bigquery/tests/unit/test_client.py
gforsyth: yep

gforsyth:
running bigquery tests locally -- I suspect this may update a few snapshots

ibis/backends/bigquery/__init__.py
cpcloud: LOL

cpcloud (approved):

PR #9143, opened 2024-05-07
feat(pyarrow): support Arrow PyCapsule interface on `ibis.Table` objects

### LINKED ISSUE #9140: feat: Implement Arrow PyCapsule Interface

Currently Ibis integrates with Arrow via the `to_pyarrow` method. The downside of this is that library consumers have to:

1. Be aware of ibis
2. Look for specific ibis data types
3. Know that they can call this `to_pyarrow` method, which tends to be named differently in different libraries. E.g. DuckDB calls it `.arrow()` and Polars calls it `.to_arrow()`.

The [Arrow PyCapsule Interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) is a new standard for exchanging Arrow data in Python. Among other benefits, this defines a single method name (`__arrow_c_stream__`) that is public and standardized

This means that other libraries don't have to build specific connectors to Polars, DuckDB, Ibis, pyarrow, etc, but rather can implement support for _any_ input object with an `__arrow_c_stream__` method. For my particular use case, a geospatial visualization library I develop, [Lonboard](https://github.com/developmentseed/lonboard), is Arrow based and looks for this method.

Implement an `__arrow_c_stream__` method wherever there's currently a `to_pyarrow` method. This could be as simple as

```py
    def __arrow_c_stream__(self, requested_schema):
        return self.to_pyarrow().__arrow_c_stream__(requested_schema)
```

where it uses the fact that the pyarrow `Table` class implements this as of v14 or so (not 100% sure which version it was added in)

I haven't run ibis yet but `__arrow_c_stream__` is not found in a code search of the repo.

_No response_

### Code of Conduct

jcrist:
Fixes #9140.

### DIFF
--- a/ibis/backends/tests/test_export.py
+++ b/ibis/backends/tests/test_export.py
@@ -4,6 +4,7 @@ import pandas as pd
 import pyarrow as pa
 import pyarrow.csv as pcsv
 import pytest
+from packaging.version import parse as vparse
 from pytest import param
 
 import ibis
@@ -42,6 +43,20 @@ no_limit = [
 ]
 
 limit_no_limit = limit + no_limit
+
+
+@pytest.mark.skipif(
+    vparse(pa.__version__) < vparse("14"), reason="pyarrow >= 14 required"
+)
+def test_table___arrow_c_stream__(awards_players):
+    res = pa.table(awards_players)
+    sol = awards_players.to_pyarrow()
+    assert res.equals(sol)
+
+    # With explicit schema
+    schema = awards_players.schema().to_pyarrow()
+    res = pa.table(awards_players, schema=schema)
+    assert res.equals(sol)
 
 
 @pytest.mark.parametrize("limit", limit_no_limit)
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -188,6 +188,9 @@ class Table(Expr, _FixedTextJupyterMixin):
 
         return IbisDataFrame(self, nan_as_null=nan_as_null, allow_copy=allow_copy)
 
+    def __arrow_c_stream__(self, requested_schema: object | None = None) -> object:
+        return self.to_pyarrow().__arrow_c_stream__(requested_schema)
+
     def __pyarrow_result__(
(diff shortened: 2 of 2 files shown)


ibis/backends/tests/test_export.py
```diff
@@ -44,6 +44,19 @@
 limit_no_limit = limit + no_limit
 
 
+def test_table___arrow_c_stream__(awards_players):
```
cpcloud, line 47: Not enough underscores here.

cpcloud (approved):
Nothing blocking, can address the pyarrow skipping in a follow up

ibis/backends/tests/test_export.py
```diff
@@ -44,6 +44,19 @@
...
+        pytest.skip("Arrow PyCapsule interface not available in this pyarrow version")
+    # Smoketest for __arrow_c_stream__ interface
+    # Here we just plumb through to the underlying pyarrow implementation
+    obj = awards_players.__arrow_c_stream__()
+    assert obj is not None
+
+    schema = awards_players.schema().to_pyarrow().__arrow_c_schema__()
+    obj2 = awards_players.__arrow_c_stream__(schema)
+    assert obj2 is not None
+
```
jorisvandenbossche, line 58: You could add a check that `pa.table(awards_players)` runs (or check explicitly that it returns the same as `to_pyarrow`, like `pa.table(awards_players).equals(awards_players.to_pyarrow())`)

jcrist: Ah, that's a much better test, thanks!

ibis/backends/tests/test_export.py
```diff
@@ -44,6 +44,19 @@
 limit_no_limit = limit + no_limit
 
 
+def test_table___arrow_c_stream__(awards_players):
+    if not hasattr(pa.Table, "__arrow_c_stream__"):
```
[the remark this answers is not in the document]
jorisvandenbossche: It was added in version 14.0.0, if you want to check explicitly

PR #9149, opened 2024-05-07
fix(bigquery): quote qualified memtable names

gforsyth:
xref #8679

This should hopefully fix the last failure mode noted by @kajen3 -- we weren't quoting the fully qualified path to memtables when we create them.


cpcloud (approved):

cpcloud:
Thanks for follow up here!

PR #9155, opened 2024-05-08
fix(duckdb): use existing table repr for settings view

cpcloud:
Fixes the repr of the duckdb backend settings property to be the table containing all settings


gforsyth (approved):

gforsyth:
Nice!

Issue #9112, opened 2024-05-03
bug(bigquery): different name of column in sql code and table.columns


jitingxu1:
### What happened?

Here is the code could be used to reproduce the error:
```python

import ibis
ibis.options.interactive = True

con = ibis.bigquery.connect(project_id="xxx", dataset_id="xxx")

t = ibis.memtable({
    "x": ["a", "b"],
    "y": ["0", "1"]
})
con.create_table(
    "t", t.to_pyarrow(), overwrite=True
)

t = con.table("t")
t = t.mutate(
    x1=ibis.literal("c"),
    y=_.y.cast(dt.int32),
)

x = t.drop("y")
y = t.y.cast(dt.int64)

```
when I run 
```python
y.as_table().columns
```
I got ```['Cast(y, int64)']```

But the name in sql code is different in `ibis.to_sql(y)`- name in sql code is `Cast_y_ int64`
```
SELECT
  CAST(CAST(`t0`.`y` AS INT64) AS INT64) AS `Cast_y_ int64`
FROM `voltrondata-demo`.`nycflights13`.`t` AS `t0`
```

This will produce errors in ibisml, becuase ibisml use the table's column name to extract target - https://github.com/ibis-project/ibis-ml/blob/main/ibisml/core.py#L135
```python
import ibisml as ml

step = ml.TargetEncode(["x"])
table, targets, index = ml.core.normalize_table(x, y)
print(targets)
metadata = ml.core.Metadata(targets=targets)
step.fit_table(table,metadata)
step.transform_table(x)
```

Errror
```
Invalid field name "Cast(y, int64)_f82688".
```

***************Updated the error message**********************
```
400 POST https://bigquery.googleapis.com/upload/bigquery/v2/projects/voltrondata-demo/jobs?uploadType=multipart: Invalid field name "Cast(y, int64)_ef9a46". Fields must contain the allowed characters, and be at most 300 characters long. For allowed characters, please refer to https://cloud.google.com/bigquery/docs/schemas#column_names
```

Not very sure, we should solve this in ibis or ibisml. 

### What version of ibis are you using?

9

### What backend(s) are you using, if any?

bigquery

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jitingxu1:
Rewrite the problem description with an easy example

jitingxu1:
***************Updated the error message**********************
```
[...]

PR #9160, opened 2024-05-08
fix(bigquery): strip whitespace from bigquery field names

gforsyth:
Removes extra whitespace from the generated field name.

xref #9112 but doesn't strictly solve the issue


gforsyth:
running snapshot updates now...

gforsyth:
snapshots updated

cpcloud (approved):

Issue #8540, opened 2024-03-04
feat: make .topk() and .value_counts() output same column names


NickCrews:
### Is your feature request related to a problem?

Ideally, these two methods would return the same data, except that .topk() is .order_by()ed and .limit()ed.

But currently the output column names are different (`a_count` vs `Count(a)`)

```python
import ibis

ibis.options.interactive = True

t = ibis.memtable({"a": [1, 1, 2, None]})
t.a.value_counts()
# ┏━━━━━━━━━┳━━━━━━━━━┓
# ┃ a       ┃ a_count ┃
# ┡━━━━━━━━━╇━━━━━━━━━┩
# │ float64 │ int64   │
# ├─────────┼─────────┤
# │     2.0 │       1 │
# │    NULL │       1 │
# │     1.0 │       2 │
# └─────────┴─────────┘
t.a.topk(10)
# ┏━━━━━━━━━┳━━━━━━━━━━┓
# ┃ a       ┃ Count(a) ┃
# ┡━━━━━━━━━╇━━━━━━━━━━┩
# │ float64 │ int64    │
# ├─────────┼──────────┤
# │     1.0 │        2 │
# │     2.0 │        1 │
# │    NULL │        0 │
# └─────────┴──────────┘
```

### Describe the solution you'd like

I think adding a param `name`, with the same semantics as `lname` and `rname` in join, with a default of `{name}_count`, would be nice. Or, just making this always be the behavior, with no option, would also be ok. I vote for `{name}_count` over `Count({name})`, because it is easier to construct the first one programatically, eg

```python
cols = ["a", "b", "c"]
counts = [t[col].value_counts().filter(_[col + "_count"] > 0)]
```

Also this is more consistent with our default behavior of adding suffixes in .join()

Anything else we could do to make these two methods have the same API would be great, but I think it doesn't make sense to unify them any more.

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

NickCrews:
This is now complete!

PR #9163, opened 2024-05-09
fix: make topk() use {name}_count format, like .value_counts()

NickCrews:
Fixes (partially) https://github.com/ibis-project/ibis/issues/8540

this is breaking, but I think much better.


cpcloud:
We are going to do a point release soon, and since this is a breaking change it won't go into that release. This will need to go into the next breaking release which will be a while given we have a number of other things to do.

NickCrews:
That sounds fine to me. Just keep this open and merge it in once were ready?

cpcloud:
Closing this out. As of 9.5 you can pass your own name to the `name` parameter of `value_counts` and `topk`.

PR #9177, opened 2024-05-13
refactor(snowflake): replace array repeat udf with builtin transform function

cpcloud:
Turns out we can do this without a UDF. Thanks to @krzysztof-kwitt for the idea!

PR #9182, opened 2024-05-13
test(arrays): add array repeat test on column expression

cpcloud:
Follow up to #9177 that adds a test for repeating a column array.


cpcloud:
Apologies for the churn :grim

gforsyth (approved):
Well, at least we noticed it quickly!

Confimed that the new test is currently failing on `main` and passes here.

:shipit:

PR #9204, opened 2024-05-16
Fix ibis connect duckdb when pass config

SothanaV:
## Description of changes

- change pass config to duckdb if use ibis duckdb as a backend


cpcloud:
This PR does not have a description that matches what it does not describes why it's necessary. The TOML change also doesn't have any stated motivation.

cpcloud:
Closing this for now. We can't accept this PR as is:

1. The description doesn't match the changes.
2. Loading the httpfs extension is not something we should do at at startup.
3. Executing each individual configuration `SET` as a separate query is also nothing something we should do at startup.

PR #9211, opened 2024-05-19
feat(api): add `Table.value_counts` for easy group by count on multiple fields

cpcloud:
Adds a `value_counts` method to tables for convenience.

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -2228,11 +2228,9 @@ def test_null_isin_null_is_null(con):
     assert pd.isna(con.to_pandas(expr).iat[0])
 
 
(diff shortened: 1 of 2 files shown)


gforsyth (commented):
Sweet!  Looks like `dask` won't support the multiple order keys?  Or at least the version we're pinned on.  So I think those can be `notyet`ted -- and the `pyspark` failure is `int32` vs `int64` shenanigans which we can probably just relax in the `assert` check.

gforsyth:
Ok... so dask fails on 3.11 in the backend suite, but not on 3.11 in the nix build

gforsyth:
or rather, it `XPASS`es on 3.11, but not on 3.10 or 3.12.  

cool cool cool

PR #9213, opened 2024-05-20
feat: drop Python 3.9 and test on Python 3.10/3.12

deepyaman:
## Description of changes

According to NEP-29, we could have dropped Python 3.9 last month.

* Test on 3.10/3.12 in CI, except for a few exceptions (e.g. Flink).
* Apply `pyupgrade` rules to remove code for Python<3.10.
* Remove workaround for DuckDB geospatial tests.

## Issues closed

N/A


ci/make_geography_db.py
```diff
@@ -62,7 +62,10 @@ def make_geography_db(
             cols = ibis_schema.names
             path = Path(d, f"{table_name}.jsonl")
             path.write_text(
-                "\n".join(json.dumps(dict(zip(cols, row))) for row in data[table_name])
+                "\n".join(
+                    json.dumps(dict(zip(cols, row, strict=False)))
```
deepyaman, line 66: Kind of annoying that this happened, but the reason is that https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/ is available when targeting 3.10.

Guess these could mostly be `strict=True`, or could just ignore B905 for stylistic reasons.

jcrist: I'd vote to just disable that rule.

ibis/backends/bigquery/__init__.py
```diff
@@ -1031,7 +1031,7 @@ def create_table(
             import pyarrow as pa
             import pyarrow_hotfix  # noqa: F401
 
-            if isinstance(obj, (pd.DataFrame, pa.Table)):
+            if isinstance(obj, pd.DataFrame | pa.Table):
```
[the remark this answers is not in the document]
deepyaman: Yeah, for sure. Turns out it's a separate rule: https://docs.astral.sh/ruff/rules/non-pep604-isinstance/

And the docs even say it's slower. I'm surprised they'd enable that by default...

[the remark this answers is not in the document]
deepyaman: Done

ci/make_geography_db.py
deepyaman: and done!

jcrist (approved):
Provided tests pass, this looks good to me! We should get a review by one other person before merging (@gforsyth?) just to ensure I didn't miss anything around the CI setup.

deepyaman:
> Provided tests pass, this looks good to me! We should get a review by one other person before merging (@gforsyth?) just to ensure I didn't miss anything around the CI setup.

FYI tests have passed; the nix 3.9 test just won't complete, because it's removed.

cpcloud:
I'll remove the required nix 3.9 test

cpcloud:
@jcrist @deepyaman I think historically we made this a breaking change. Should we considering changing that convention?

.github/workflows/ibis-backends.yml
```diff
@@ -243,12 +237,32 @@ jobs:
               extras:
                 - dask
           - os: ubuntu-latest
-            python-version: "3.9"
+            python-version: "3.11"
```
ncclementi, line 240: shouldn't this be the 3.10, if it was 3.9 before?

jcrist:
> I think historically we made this a breaking change. Should we considering changing that convention?

I think that dropping a Python version in a minor release seems fine (it's also pretty common in the ecosystem). Since we've updated the minimum version requirement, even pip's solver won't allow them to install the new release with an older version of python. The "breakage" here isn't something a user should be able to encounter.

.github/workflows/ibis-backends.yml
deepyaman: The diff is a bit weird. It was a Dask entry before, this one is a Flink entry. The reason we test on 3.11 for Flink still is that PyFlink doesn't support 3.12.

cpcloud (approved):
LGTM.

Can we make this a feature or fix PR?

deepyaman:
> LGTM.
> 
> Can we make this a feature or fix PR?

Done! Also rebased.

Issue #9170, opened 2024-05-10
feat(api): Add `order_by` parameter to the `collect()` method


chelsea-lin:
### Is your feature request related to a problem?

The `collect()` method in Ibis is a convenient way to aggregate data into arrays when working with BigQuery backends. It translates to the `ARRAY_AGG` function in BigQuery SQL. However, there's currently no built-in way to specify an `ORDER BY` clause within the `ARRAY_AGG` aggregation. This means that the order of elements in the resulting array is not guaranteed, which can be problematic when the order of elements matters for the downstream analysis.

### Describe the solution you'd like

I propose adding an `order_by` parameter to the `collect()` method. This parameter would accept either:
- A string or list of strings representing column names to order by.
- An Ibis expression indicating the ordering.

This would allow users to explicitly control the ordering of elements within the aggregated array, making the results more predictable and useful.

### What version of ibis are you running?

9.0.0

### What backend(s) are you using, if any?

BigQuery

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jcrist:
Thanks for opening this! I agree that supporting ordering in `array_agg` would be useful (we're also adding an `order_by` kwarg to `Table.distinct` for similar reasons).

I'm not sure how well an `order_by` kwarg would work though in the existing `collect()` method though - given that `collect` is on a `Column` instead of a `Table`, having the ordering refer to columns on the parent table may feel weird.

```python
t = ibis.table({"a": "int", "b": "int, "c": "int"})

t.a.collect(order_by=("b", "c"))  # here `b` and `c` refer to columns on the parent table of `a`
t.a.collect(order_by=(t.b, t.c))   # explicit column references also work
```

Typing all that out doesn't feel _that_ odd, but AFAIK we don't have any other columnar operations that support an ordering specification, so we don't have an existing pattern to mirror here.

cpcloud:
`group_concat` is another operation where this would make sense, along with the `first` and `last` aggregate functions. I think we should add them on a case by case basis for now and do an audit of the aggs where ordering has an effect on the output.

jcrist:
A quick survey of backends and current ops shows those are the main 4 operations. I've taken a quick look through implementing this (needed a fun friday break) and like most things ibis it's turned up some internal cleanups we'll want to do first but I _think_ this should be only a medium-sized lift to implement.

PR #9222, opened 2024-05-21
refactor(sql): extract aggregate handling out into common utility class

jcrist:
This is an initial refactor moving towards #9170.

Previously every backend implemented their own `_aggregate` function - many of them copy-pasted (with slight variations) of each other.

To add a new `order_by` kwarg to `_aggregate` would require editing all 16 copies of this function, which would be _annoying_. This PR is an attempt to centralize their implementations into a common `Gen` class. The class takes a config flag to handle the common cases, the uncommon cases are then handled by backend-specific subclasses.

This also extracts the `.agg` attribute out to be a class variable (err, descriptor) rather than an instance variable.

An alternative implementation would be adding boolean flags directly to the `SQLGlotCompiler` class, but IIRC we intentionally moved away from those in the SQLAlchemy -> SQLGlot refactor. Could go either way here.

### DIFF
--- a/ibis/backends/bigquery/compiler.py
+++ b/ibis/backends/bigquery/compiler.py
@@ -122,14 +122,6 @@ class BigQueryCompiler(SQLGlotCompiler):
         ops.TimestampNow: "current_timestamp",
     }
 
-    def _aggregate(self, funcname: str, *args, where):
-        func = self.f[funcname]
-
-        if where is not None:
-            args = tuple(self.if_(where, arg, NULL) for arg in args)
-
-        return func(*args, dialect=self.dialect)
-
     @staticmethod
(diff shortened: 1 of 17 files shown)


jcrist:
> An alternative implementation would be adding boolean flags directly to the `SQLGlotCompiler` class

This would be simpler/less magical and would be my preference, went with the current extra level of abstraction since I thought I remembered someone else being against flags on the compiler.

gforsyth (approved):
This looks good to me -- I also have vague memories of someone having issues with adding flags on the compiler class, but I can't remember any specifics.

I suppose we can always revisit and simplify this implementation if we decide that we want to make the change, but this is a clear win over our existing scattered implementations.

PR #9232, opened 2024-05-22
feat(polars): accept list of CSVs to read_csv

### LINKED ISSUE #9230: Polars backend can read only 1 csv

Polars can read multiple files at the same time but in ibis, it can read only 1 file at the time which is not practical
https://ibis-project.org/backends/polars#ibis.backends.polars.Backend.read_csv

_No response_

I would like ibis when using polars backend to be able to do this

ibis.read_csv(['file1,csv','file2.csv'])

10.0.0.dev71

Polars

### Code of Conduct

gforsyth:
Polars now accepts a list of CSVs to `scan_csv` so we can expose this to
end-users.

A few caveats:
- it doesn't accept lists of globs, or lists of compressed CSVs, so we
  flatten the list of paths if it only has one element in case it is a
  glob or csv.gz

Fixes #9230

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -38,13 +38,6 @@ if TYPE_CHECKING:
     from fsspec import AbstractFileSystem
 
 
-def normalize_filenames(source_list):
-    # Promote to list
-    source_list = util.promote_list(source_list)
-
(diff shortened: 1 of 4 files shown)


jcrist (approved):
Nice!

PR #9236, opened 2024-05-22
test(snowflake): use env-vars only for ibis.snowflake.connect test

gforsyth:
The previous version of this test isn't passing on CI for reasons I can't quite understand, so I'm trying this instead.  We'll see?

### DIFF
--- a/ibis/backends/snowflake/tests/test_client.py
+++ b/ibis/backends/snowflake/tests/test_client.py
@@ -372,25 +372,27 @@ def test_timestamp_memtable(con):
     tm.assert_frame_equal(result, df)
 
 
-def test_connect_without_snowflake_url(con):
+def test_connect_without_snowflake_url():
     # We're testing here that the non-URL connection works.
     # Specifically that a `database` location passed in as "catalog/database"
     # will be parsed correctly
-    user = con.con.user
-    account = con.con.account
-    catalog = con.con.database
-    database = con.con.schema
-    warehouse = con.con.warehouse
+    user = os.getenv("SNOWFLAKE_USER")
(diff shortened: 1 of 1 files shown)


ibis/backends/snowflake/tests/test_client.py
```diff
@@ -372,25 +372,27 @@ def test_timestamp_memtable(con):
...
-    database = con.con.schema
-    warehouse = con.con.warehouse
+    user = os.getenv("SNOWFLAKE_USER")
+    account = os.getenv("SNOWFLAKE_ACCOUNT")
+    catalog = os.getenv("SNOWFLAKE_DATABASE")
+    database = os.getenv("SNOWFLAKE_SCHEMA")
+    warehouse = os.getenv("SNOWFLAKE_WAREHOUSE")
 
     if (password := os.getenv("SNOWFLAKE_PASSWORD")) is None:
         pytest.mark.skip(reason="No snowflake password set, nothing to do here")
```
jcrist, line 386: When inside a test, you'll want to do `pytest.skip` instead of `pytest.mark.skip`. The former raises and skips the test, the latter won't do anything (and later on you'll get an error).

jcrist (approved):
Small fixup from previous code, otherwise LGTM. :crossed_fingers: this passes.

ibis/backends/snowflake/tests/test_client.py
gforsyth: ahh, nice catch, thanks.

gforsyth:
I inadvertently clobbered the bigquery runs here by attempting to test #9238, but my plan was just to confirm that snowflake passes, and that worked, so I'm going to merge this.  :crossed_fingers:

PR #9238, opened 2024-05-22
chore: allow overlapping docs-preview and cloud-ci runs via label

gforsyth:
Or at least, that's the hope here.  Up to now, applying one of these
labels while the other is in place has led to a clobbering of any CI
jobs triggered by adding a label.
My hope is that the use of `contains` will prevent that from happening
since it's about whether that label is _present_, not only if the most
recent labeling _event_ matches the target name.


gforsyth:
Nope.

PR #9241, opened 2024-05-23
feat(duckdb): allow to use named in-memory db

JulienZolli:
## Description of changes

Allows to use a named in-memory database with the DuckDB backend.

## Issues closed

* Resolves #9240

PR #9242, opened 2024-05-23
test(duckdb): test multiple connections to named in-memory db

gforsyth:
Also test that named DB catalogs aren't bleeding over into unnamed memory dbs

This is a test to ensure we don't accidentally remove the functionality added in
#9241 for connecting to named in-memory databases in DuckDB.


cpcloud (approved):

cpcloud:
Thanks!

PR #9251, opened 2024-05-24
feat(all): enable passing in-memory data to create_table

### LINKED ISSUE #6593: feat: Standardize `read_in_memory` across backends

gforsyth:
This PR adds/codifies support for passing in-memory data to `create_table`.

The default behavior for most backends is to first create a `memtable` with
whatever `obj` is passed to `create_table`, then we create a table based on that
`memtable` -- because of this, semantics around `temp` tables and
`catalog.database` locations are handled correctly. 

After the new table (that the user has provided a name for) is created, we
drop the intermediate `memtable` so we don't add two tables for every in-memory
object passed to `create_table`.

Currently most backends fail when passed `RecordBatchReaders`, or a single
`RecordBatch`, or a `pyarrow.Dataset` -- if we add support for these to
`memtable`, all of those backends would start working, so I've marked those
xfails as `notimpl` for now.

A few backends _don't_ work this way:

`polars` reads in the table directly using their fast-path local-memory reading stuff.

`datafusion` uses a fast-path read, then creates a table from the table that is
created by the fast-path -- this is because the `datafusion` dataframe API has
no way to specify things like `overwrite`, or table location, but the CTAS from
already present tables is very quick (and _possibly_ zero-copy?) so no issue
there.

`duckdb` has a refactored `read_in_memory` (which we should deprecate), but it
isn't entirely hooked up inside of `create_table` yet, so some paths may go via
`memtable` creation, but `memtable` creation on DuckDB is especially fast, so
I'm all for fixing this up eventually.

`pyspark` works with the intermediate `memtable` -- there are possibly
fast-paths available, but they aren't currently implemented.

`pandas` and `dask` have a custom `_convert_object` path

TODO:
* ~[ ] Flink~  Flink can't create tables from in-memory data?
* [x] Impala
* [x] BigQuery 
* [x] Remove `read_in_memory` from datafusion and polars

Resolves #6593 
xref #8863

Signed-off-by: Gil Forsyth <gil@forsyth.dev>
- refactor(duckdb): add polars df as option, move test to backend suite
- feat(polars): enable passing in-memory data to create_table
- feat(datafusion): enable passing in-memory data to create_table
- feat(datafusion): use info_schema for list_tables
- feat(duckdb): enable passing in-memory data to create_table
- feat(postgres): allow passing in-memory data to create_table
- feat(trino): allow passing in-memory date to create_table
- feat(mysql): allow passing in-memory data to create_table
- feat(mssql): allow passing in-memory data to create_table
- feat(exasol): allow passing in-memory data to create_table
- feat(risingwave): allow passing in-memory data to create_table
- feat(sqlite): allow passing in-memory data to create_table
- feat(clickhouse): enable passing in-memory data to create_table
- feat(oracle): enable passing in-memory data to create_table
- feat(snowflake): allow passing in-memory data to create_table
- feat(pyspark): enable passing in-memory data to create_table
- feat(pandas,dask): allow passing in-memory data to create_table


ibis/backends/datafusion/__init__.py
```diff
@@ -294,6 +301,44 @@ def get_schema(
...
+            An optional name to use for the created table. This defaults to
+            a sequentially generated name.
+        kwargs
+            Keyword arguments to forward to a memory-format-specific reader function.
+
+        Returns
+        -------
+        ir.Table
+            The just-registered table
+
```
cpcloud, line 334:
```suggestion
```

ibis/backends/datafusion/__init__.py
```diff
@@ -670,3 +723,92 @@ def truncate_table(
...
+    """Workaround inability to overwrite tables in dataframe API.
+
+    Datafusion has helper methods for loading in-memory data, but these methods
+    don't allow overwriting tables.
+    The SQL interface allows creating tables from existing tables, so we register
+    the data as a table using the dataframe API, then run a
+
+    CREATE [OR REPLACE] TABLE table_name AS SELECT * FROM in_memory_thing
+
+    and that allows us to toggle the overwrite flag
```
cpcloud, line 739:
```suggestion
    and that allows us to toggle the overwrite flag.
```

cpcloud (approved):
Nitiest of nits. LGTM overall!

gforsyth:
Question: would folks rather I add `polars` as an extra to many CI jobs to test polars inputs to `create_table` or put an `importorskip` around it?

cpcloud:
@gforsyth Can we add polars to one or two backends instead of all of them?

gforsyth:
> @gforsyth Can we add polars to one or two backends instead of all of them?

Yep -- added it explicitly to MySQL, MSSQL, and Oracle.  And we already have it available on the DuckDB jobs, and the Postgres torch job (and obviously on the polars jobs) -- seems like reasonably good coverage?

cpcloud (approved):

PR #9253, opened 2024-05-25
ci(snowflake): ensure that snowpark tests run in their own schema

cpcloud:
Ensure that snowpark tests and non-snowpark tests run in different schemas.


cpcloud:
No need to wait for bigquery here, and the schema changes won't take effect until the workflow runs against `main`, merging.

PR #9263, opened 2024-05-28
feat(polars): implement `create_view`/`drop_view`/`drop_table`

jcrist:
For the `polars` backend:

- Implements `create_view`/`drop_view` as aliases for the equivalent table versions (as we do for pandas and dask).
- Implements `drop_table`


gforsyth (approved):
:shipit:

PR #9264, opened 2024-05-28
fix(ddl): use column names, not position, for insertion order

### LINKED ISSUE #9249: bug: .insert() uses column position, not names, to align
### What happened?

When I insert new data, I expect the columns to get aligned by name. But we seem to be naive, and not ensure the src and dst schemas are in the same order, and thus we get the default SQL behavior of data being aligned by position:

```python
import ibis

db = ibis.duckdb.connect()
db.create_table("t", [{"a": 1, "b": 2}])
db.insert("t", [{"b": 22, "a": 11}])
db.table("t")
# ┏━━━━━━━┳━━━━━━━┓
# ┃ a     ┃ b     ┃
# ┡━━━━━━━╇━━━━━━━┩
# │ int64 │ int64 │
# ├───────┼───────┤
# │     1 │     2 │
# │    22 │    11 │
# └───────┴───────┘
```

main

duckdb, but I bet this is present in other backends too.

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
Depends on #9251 -- better handling for various memtable creation and insertion from there.

Resolves #9249 

- fix(impala): use column names for insertion order, not position
- fix(sqlite): use column names, not position, for insertion order
- fix(SQLBackend): use column names, not position, for insertion order
- fix(clickhouse): use column names, not position, for insertion order
- test(all): add test case for insert-by-name


gforsyth:
Hmm, I'm going to have to rethink this -- moving this to draft for the moment

gforsyth:
Ok, this is ready for a look, but I'll be OOO for the next while, so feel free to fix it up or nuke it as necessary.

cpcloud (approved):
This seems good!

We should probably start to do some consolidation of backend DDL code, even if we don't get to the point of having something like a set of internal DDL op classes.

ibis/backends/clickhouse/__init__.py
```diff
@@ -423,7 +423,25 @@ def insert(
...
-        query = sge.insert(self.compile(obj), into=name, dialect=self.name)
+        # Compare the columns between the target table and the object to be inserted
+        # If they don't match, assume auto-generated column names and use positional
+        # ordering.
+        columns = (
+            obj.columns
+            if not set(parent_cols := self.get_schema(name).names).difference(
+                obj.columns
+            )
+            else parent_cols
```
cpcloud, line 434: I'll rename these to `target_*` since `parent` is a bit confusing in this context: there's generally tree relationship between the input table and and the target table.

cpcloud: Done.

cpcloud:
Consolidate the implementation under a new private backend on `SQLBackend` (`_build_insert_query`)

cpcloud (approved):

Issue #9247, opened 2024-05-24
feat: better distinguish ibis scalars from evaluated results in interactive repr


jcrist:
Currently in interactive mode, we evaluate `Table`, `Column` and `Scalar` expressions and display their results. For `Table` and `Column` values the repr is easily distinguishable from other python types. For scalar values it's not - they display the same as their in-memory counterparts. When working interactively (especially in a fast backend like `duckdb`) this can sometimes be confusing, as it makes it easy to forget that some object is an ibis scalar expression rather than its evaluated result.

```python
In [1]: import ibis

In [2]: t = ibis.memtable({"x": [1, 2, None], "y": [3, 4, 5]})

In [3]: ibis.options.interactive = True

In [4]: t  # reprs in a unique enough way to know it's not a builtin type
Out[4]: 
┏━━━━━━━━━┳━━━━━━━┓
┃ x       ┃ y     ┃
┡━━━━━━━━━╇━━━━━━━┩
│ float64 │ int64 │
├─────────┼───────┤
│     1.0 │     3 │
│     2.0 │     4 │
│    NULL │     5 │
└─────────┴───────┘

In [5]: t.x  # same here
Out[5]: 
┏━━━━━━━━━┓
┃ x       ┃
┡━━━━━━━━━┩
│ float64 │
├─────────┤
│     1.0 │
│     2.0 │
│    NULL │
└─────────┘

In [6]: t.x.sum()  # reprs the same as a float. Is this a float, or an ibis expression? Hard to say from just looking at the repr.
Out[6]: 3.0

In [7]: ibis.literal(None)  # same here - is this a None or an ibis object?
Out[7]: None
```

This issue is partly to blame for confusion around #8833 - in this case the subexpression reprs as `None`, which led them to believe it would be treated the same as `None` if passed to a method to construct a larger expression. Better visually distinguishing these types would help here IMO.

I think it would be good to amend our interactive repr for scalars to better distinguish them still as ibis objects. I'm not sure what styling elements would be best to use here - perhaps just box it, the same as we do for columns or tables?

```python
In [8]: t.x.sum()

Out[8]:
┌─────┐ 
│ 3.0 │
└─────┘
```

Issue #8833, opened 2024-03-30
bug: many Operations can't handle NULL optional arguments


NickCrews:
### What happened?

consider the implementation of

```python
@public
class Substring(Value):
    arg: Value[dt.String]
    start: Value[dt.Integer]
    length: Optional[Value[dt.Integer]] = None
```

This allows you to leave out the length argument, meaning "until the end of the string". But this is only represented on the python side. If the `length` is evaluated at runtime to be NULL, then this errors on some backends, like postgres, or gives the wrong result on duckdb.

For example, `ibis.literal("abcde").substr(2, ibis.literal(1).nullif(1))` results in NULL in duckdb, but I would expect it to be "cde".

During compilation, we are naive and do the NULL checking only on the python side:

```python
def visit_Substring(self, op, *, arg, start, length):
        start += 1
        arg_length = self.f.length(arg)

        if length is None:
            return self.if_(
                start >= 1,
                self.f.substring(arg, start),
                self.f.substring(arg, start + arg_length),
            )
        return self.if_(
            start >= 1,
            self.f.substring(arg, start, length),
            self.f.substring(arg, start + arg_length, length),
        )
```

I discovered this when adding more tests to #8832 .

I think what we should do is make Substring more like

```python
@public
class Substring(Value):
    arg: Value[dt.String]
    start: Value[dt.Integer]
    length:Value[dt.Integer] = ibis.null()
```

and then use sql CASE statements if we can't determine the nullness statically.

but I wasn't able to get this to work. Does this seem like the right direction? Any tips on what to do here?

### What version of ibis are you using?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
None indicates that the argument wasn't provided, NULL is something else. We should avoid conflating these two uses.

cpcloud:
The failures notwithstanding, the duckdb behavior looks correct to me in this case: null inputs produce null outputs, NULL doesn't mean "no argument".

The representation of "argument not provided" is unrelated to whether an input is NULL.

NickCrews:
hmmm, I think you are right. We are victims of the SQL standard here. I don't really like this footgun:
- `ibis.literal("abcde").substr(2, None)` gives "cde", but
- `ibis.literal("abcde").substr(2, ibis.null())` gives `NULL`
[...]

PR #9265, opened 2024-05-28
chore: add box to scalar expr in interactive mode

ncclementi:
This is an attempt to close https://github.com/ibis-project/ibis/issues/9247 and possible the discussion around https://github.com/ibis-project/ibis/issues/8833

I'm using rich Panel and the `box.SQUARE` style to wrap scalar values into a box. There are multiple styles (see https://rich.readthedocs.io/en/stable/appendix/box.html), I chose the one that looks the most similar to what we have now. 

TODO:
- Do we need a tests for this? I wasn't quite sure how to add a test for it. 
- Modify doctests accordingly if we like this change

Closes https://github.com/ibis-project/ibis/issues/9247

Here is how it renders in ipython. 

```python
In [1]: import ibis

In [2]: ibis.options.interactive = True

In [3]: ibis.literal(1).nullif(1)
Out[3]: 
┌──────┐
│ None │
└──────┘

In [4]: ibis.literal(1).nullif(0)
Out[4]: 
┌───┐
│ 1 │
└───┘

In [5]: ibis.literal("abcdefghi").substr(3)
Out[5]: 
┌──────────┐
│ 'defghi' │
└──────────┘

In [6]: print(ibis.literal("abcdefghi").substr(3))
┌──────────┐
│ 'defghi' │
└──────────┘

In [7]: print(ibis.literal(True))
┌──────┐
│ True │
└──────┘

In [8]: import datetime

In [9]: date = datetime.datetime(2015, 1, 1, 12, 34, 56)

In [10]: print(ibis.literal(date))
┌──────────────────────────────────┐
│ Timestamp('2015-01-01 12:34:56') │
└──────────────────────────────────┘


```

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -207,7 +207,9 @@ Examples
 >>> import ibis
 >>> ibis.options.interactive = True
 >>> ibis.NA.isnull()
-True
+┌──────┐
+│ True │
+└──────┘
 
 datatype-specific methods aren't available on `NA`:
 
@@ -219,7 +221,9 @@ AttributeError: 'NullScalar' object has no attribute 'upper'
 Instead, use the typed `ibis.null`:
 
 >>> ibis.null(str).upper().isnull()
-True
+┌──────┐
+│ True │
+└──────┘
 """
 
 deferred = _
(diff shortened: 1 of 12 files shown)


ncclementi:
After chatting in triage we are ok with the change, I will go ahead and modify the doctests to match the new repr.

cpcloud (approved):

PR #9266, opened 2024-05-28
feat(pyspark): support reading from and writing to Kafka

chloeh13q:
## Description of changes

Introduce `read_kafka` and `to_kafka` methods to support reading from and writing to Kafka in streaming mode.


ibis/backends/pyspark/__init__.py
```diff
@@ -929,3 +940,110 @@ def to_pyarrow_batches(
...
+                "Reading from Kafka in batch mode is not supported"
+            )
+        spark_df = self._session.readStream.format("kafka")
+        if options is not None:
+            for k, v in options.items():
+                spark_df = spark_df.option(k, v)
+        spark_df = spark_df.load()
+
+        # parse the values of the Kafka messages using the provided schema
+        if auto_parse is True:
```
cpcloud, line 988: This should probably be

```suggestion
        if auto_parse:
```

unless you want a call like `auto_parse=1` to not take this branch.

ibis/backends/pyspark/__init__.py
```diff
@@ -929,3 +940,110 @@ def to_pyarrow_batches(
...
+            The just-registered table
+        """
+        if self.mode == "batch":
+            raise NotImplementedError(
+                "Reading from Kafka in batch mode is not supported"
+            )
+        spark_df = self._session.readStream.format("kafka")
+        if options is not None:
+            for k, v in options.items():
+                spark_df = spark_df.option(k, v)
```
cpcloud, line 984: There's a slightly less nested pattern for this:

```suggestion
        for k, v in (options or {}).items():
            spark_df = spark_df.option(k, v)
```

If `options` is `None` it will be an empty `dict` and thus the body of the loop will never be executed, otherwise, it's not empty and the body will be executed.

ibis/backends/pyspark/__init__.py
```diff
@@ -929,3 +940,110 @@ def to_pyarrow_batches(
...
+        watermark
+            Watermark strategy for the table.
+        auto_parse
+            Whether to parse Kafka messages automatically. If `False`, the source is read
+            as binary keys and values. If `True`, the key is discarded and the value is
+            parsed using the provided schema.
+        schema
+            Schema of the value of the Kafka messages.
+        options
+            Additional arguments passed to PySpark as .option("key", "value").
```
cpcloud, line 969: From the code it seems like these are Kafka options, by way of PySpark, but they're not options for PySpark. Is that true? If so, this description should be a more clear that it's KafKa optionality being affected here. Alternatively, `Additional arguments passe to Kafka via PySpark` would also be fine.

ibis/backends/pyspark/__init__.py
```diff
@@ -929,3 +940,110 @@ def to_pyarrow_batches(
...
+    @util.experimental
+    def read_kafka(
+        self,
+        table_name: str | None = None,
+        watermark: Watermark | None = None,
+        auto_parse: bool = False,
+        schema: sch.Schema | None = None,
+        options: Mapping[str, str] | None = None,
+    ) -> ir.Table:
+        """Register a Kafka topic as a table in the current database.
```
cpcloud, line 953:
```suggestion
        """Register a Kafka topic as a table.
```

The fact that the table is in the current database doesn't seem particularly informative here.

ibis/backends/pyspark/__init__.py
```diff
@@ -11,7 +11,14 @@
...
 from pyspark.sql import SparkSession
-from pyspark.sql.functions import PandasUDFType, pandas_udf
+from pyspark.sql.functions import (
+    PandasUDFType,
+    col,
+    from_json,
+    pandas_udf,
+    struct,
+    to_json,
+)
```
cpcloud, line 21: The pattern we use for this module is `import pyspark.sql.functions as F` and then refer to the module's objects as `F` wherever they are used. Having a `col` thing floating around IMO makes this code hard to understand given how many things something named `col` could refer to.

ibis/backends/pyspark/__init__.py
```diff
@@ -89,6 +99,10 @@ def nullify_type_mismatched_value(raw):
     return unwrap
 
 
+def _format_interval_as_string(interval):
```
cpcloud, line 102:
```suggestion
def _interval_to_string(interval):
```

ibis/backends/pyspark/__init__.py
```diff
@@ -929,3 +940,110 @@ def to_pyarrow_batches(
...
+            raise NotImplementedError("Writing to Kafka in batch mode is not supported")
+        df = self._session.sql(expr.compile())
+        if auto_format is True:
+            df = df.select(
+                to_json(struct([col(c).alias(c) for c in df.columns])).alias("value")
+            )
+        sq = df.writeStream.format("kafka")
+        if options is not None:
+            for k, v in options.items():
+                sq = sq.option(k, v)
```
cpcloud, line 1047: Same as above:

```suggestion
        for k, v in (options or {}).items():
            sq = sq.option(k, v)
```

cpcloud (changes requested):
A few changes are needed here.

ibis/backends/pyspark/__init__.py
chloeh13q: What do you mean by "Kafka options" but not "options for PySpark"? On the URL that I linked in the line below https://spark.apache.org/docs/latest/structured-streaming-kafka-integration.html it lists out all the configurations that Pyspark are required vs supported when connecting to Kafka sink/source from Pyspark, and it doesn't explicitly mention that they're passed to Kafka directly. They're not PySpark session configurations, but they are configurations that PySpark manages

cpcloud: Ah, I think I am thrown off by the fact that they are clearly only related to Kafka, but being specified in the PySpark API.

What you have here is fine as is then!

cpcloud (approved):
LGTM!

Curious to see how this gets used!

PR #9272, opened 2024-05-30
feat(pyspark): implement new experimental read/write directory methods

### LINKED ISSUE #8984: feat(spark): can we support connectors better?

chloeh13q:
## Description of changes

Implement new experimental read/write directory methods in Pyspark backend to support streaming read/write.

## Issues closed

#8984


chloeh13q:
xref: https://github.com/ibis-project/testing-data/pull/9

ibis/backends/pyspark/__init__.py
```diff
@@ -929,3 +927,186 @@ def to_pyarrow_batches(
         return pa.ipc.RecordBatchReader.from_batches(
             pa_table.schema, pa_table.to_batches(max_chunksize=chunk_size)
         )
+
+    @util.experimental
+    def read_csv_directory(
```
cpcloud, line 932: I thought we had agreed on `read_*_dir`.

chloeh13q: Oh in the notes it says we agreed on having separate methods, for example, "`to_csv_dir`/`to_parqet_dir` (or much better named)" so I named them `to_*_directory`. But I'm happy to change it to `to_*_dir` if that's better, I don't feel strongly about this.

cpcloud: Let's go with `dir` for now.

ibis/backends/pyspark/tests/conftest.py
```diff
@@ -360,6 +325,7 @@ def con(data_dir, tmp_path_factory, worker_id):
 @pytest.fixture(scope="session")
 def con_streaming(data_dir, tmp_path_factory, worker_id):
     backend_test = TestConfForStreaming.load_data(data_dir, tmp_path_factory, worker_id)
+    backend_test._load_data()
```
cpcloud, line 328: Not 100% sure why this is needed, let me investigate a bit what's going on.

chloeh13q: thanks!

chloeh13q: I figured out what the problem is! The pyspark backend loads statefully and the streaming conf ended up reusing the same temp directory from the batch conf, which already exists when the batch conf tried to load data, so the streaming conf skips data loading. Because `tmpdir` is passed as a fixture, I added a line of code that changes the directory naming for the streaming conf.

ibis/backends/pyspark/tests/test_import_export.py
```diff
@@ -1,9 +1,42 @@
...
+    if self.mode == "batch":
+        raise NotImplementedError
+    df = self._session.sql(expr.compile())
+    df.writeStream.format("memory").queryName(table_name).start()
+
+
+@pytest.fixture(autouse=True, scope="function")
+def stop_active_jobs(con_streaming):
+    yield
+    for sq in con_streaming._session.streams.active:
```
cpcloud, line 31: Any reason not to put this in the `con_streaming` fixture?

`yield` fixtures will only run the code after the `yield` on teardown.

chloeh13q: (IIRC) it's because we need this to run on a function scope (after every test) rather than session scope

cpcloud (approved):
Let's merge this! Since we're using the `experimental` we are free to break this across non-major versions based on user feedback/bugs/etc.

PR #9285, opened 2024-06-01
build: pin semrel and conventionalcommits deps in release scripts

cpcloud:


cpcloud:
Merging, the release notes script is passing now.

PR #9288, opened 2024-06-01
ci: run one nix job on arm64 runner

cpcloud:
Testing out the new GitHub-hosted arm64 runners with a single nix job.


cpcloud:
Merging to get minimal arm64 coverage for now. We can always revisit if this becomes a burden in any way.

PR #9290, opened 2024-06-02
fix(polars,mysql): avoid execution of query in `_get_schema_using_query`

### LINKED ISSUE #9162: bug: `.sql()` sometimes execute the query in order to get the schema
### What happened?

The `.sql()` method calls `_get_schema_using_query()` underneath the hood, which uses the query to get the schema if a schema is not passed as an argument. The implementation of `_get_schema_using_query()` differs across backends, but for the most part, if the backend provides a way to analyze the query, that's what we use. If the backend doesn't, a lot of times we create a new view/table, execute the query, and then drop the view/table so that the method has no side effect. In some backends, however, we're just executing the query. Examples: Flink, PySpark.

main

_No response_

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Closes #9162.

### DIFF
--- a/ibis/backends/mysql/__init__.py
+++ b/ibis/backends/mysql/__init__.py
@@ -192,7 +192,9 @@ class Backend(SQLBackend, CanCreateDatabase):
         table = util.gen_name(f"{self.name}_metadata")
 
         with self.begin() as cur:
-            cur.execute(f"CREATE TEMPORARY TABLE {table} AS {query}")
+            cur.execute(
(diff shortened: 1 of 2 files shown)


ncclementi (approved):
LGTM. 

If possible, is there a way we can add test to make sure that if and when we add a new backend, we don't trigger execution?

cpcloud:
Hm, good question. I don't think so, since we often use arbitrary queries to access metadata and I don't think there's a way to easily determine whether such a query has just accessed metadata.

PR #9299, opened 2024-06-03
feat(datafusion): use pyarrow for type conversion

cpcloud:
## Description of changes

This PR moves datafusion type system conversion to use PyArrow, which is much
more robust than the SQL hacking we've been doing.

~I've added two new methods to support this, required to workaround datafusion inconsistencies in output types
but also potentially useful for ClickHouse~:

- ~`DataType._as_nullable`: returns a nullable version of the type, recursively converting for nested types.~
- ~`DataType._as_non_nullable`: similar to `_as_nullable` but going the other direction, making everything **non**-nullable.~

~Two analogous methods have been added to `Schema` as well.~

Unit tests are included, including a hypothesis test to cover as many datatypes as possible.

This now is specific to datafusion to avoid an unnecessary maintenance burden for functionality that might never be used.


cpcloud:
Going to make this only apply to datafusion instead of a broader set of functionality. No reason to do it unless needed/asked for and it's currently only needed in the DataFusion case.

jcrist (approved):

Issue #7276, opened 2023-10-02
refactor: deprecate `NA`/`fillna`/`dropna` in favor of name that uses `null` instead of `na`


jcrist:
Ibis has a few APIs that make use of the name "na".

- `Table.dropna` for dropping rows containing null values
- `Table.fillna` for replacing null values with specific values
- `ibis.NA` as a null-value expression singleton

Nowhere else in `ibis` do we make use of the term "NA". Also, since some backends (e.g. `pandas`) conflate `NaN` and `NULL`, some users may expect `Table.fillna` to also replace `NaN` values in all backends (it does not). IMO we should deprecate these APIs in favor of ones that use the word "null" instead.

A few options:

**Table.dropna**

- `Table.dropnull`?
- `Table.drop_null`?
- `Table.drop_nulls`? (this is the name `polars` uses)

My preference would be for `Table.drop_null`.

**Table.fillna**

- `Table.fillnull`?
- `Table.fill_null`? (this is the name `polars` uses)
- `Table.fill_nulls`?
- `Table.ifnull`? (since we have `Column.nullif`)

My preference would be for `Table.fill_null`.

**ibis.NA**

This one I'm tempted to remove entirely.

- None should be valid as an input everywhere `ibis.NA` is
- There's always `ibis.null()` for the rare cases where you need to construct a null value expression
- If we still want a constant null, we could name it `ibis.NULL` instead?

gforsyth:
I am definitely in favor of removing `ibis.NA` and moving away from the `na` name in the `fill_` and `drop_` method.
I'm good with `drop_null` and `fill_null` -- I am not on board with having one of them be plural and the other singular.

NickCrews:
related: #7147

I agree with everything you have here. I would lean towards fillnull and dropnull, there is plenty of existing precedent within ibis for names without underscores, and it is easier to type and readable enough. But if you went with underscores I wouldn't be furious :)

cpcloud:
+1 on consistency with plurals, no strong opinions on the separating underscore. I do find `drop_null` ever so slightly easier to read, but I'd also be okay with `dropnull`. Whoever implements this can have the final say :)

kszucs:
I'm also in favor of the deprecations and have a slight preference towards underscore-less names.

ncclementi:
Folks, looking to get started with this. Couple of question on how to go about a deprecation cycle. 

[...]

PR #9300, opened 2024-06-03
refactor: deprecate fillna and dropna api

ncclementi:
WIP 
- Closes [#7276](https://github.com/ibis-project/ibis/issues/7276) 

## TODO

- [ ] Remove ibis.NA
- [ ] Understand the purpose of this,
https://github.com/ibis-project/ibis/blob/1c7c6e3785900cabbf25abbea5ed8505c8391b60/ibis/expr/format.py#L267-L271


ibis/expr/types/generic.py
```diff
@@ -401,6 +401,27 @@ def fillna(self, fill_value: Scalar) -> Value:
...
+        Returns
+        -------
+        Value
+            `self` filled with `fill_value` where it is `NULL`
+        """
+        return ops.Coalesce((self, fill_value)).to_expr()
+
+    @deprecated(as_of="10.0", instead="use fillnull instead")
+    def fillna(self, fill_value: Scalar) -> Value:
+        """Replace any null values with the indicated fill value.
```
jcrist, line 413: IMO we should delete the whole docstring and replace it with

```
"""Deprecated - use `fillnull` instead."""
```

ibis/expr/types/generic.py
```diff
@@ -401,6 +401,27 @@ def fillna(self, fill_value: Scalar) -> Value:
...
+        Parameters
+        ----------
+        fill_value
+            Value with which to replace `NA` values in `self`
+
+        See Also
+        --------
+        [`Value.coalesce()`](./expression-generic.qmd#ibis.expr.types.generic.Value.coalesce)
+        [`ibis.coalesce()`](./expression-generic.qmd#ibis.coalesce)
+
```
jcrist, line 424: Should also update the implementation below to just call `self.fillnull` instead.

ibis/expr/types/relations.py
```diff
@@ -2719,6 +2724,87 @@ def fillna(
...
+            (`'all'`).
+
+        Returns
+        -------
+        Table
+            Table expression
+        """
+        if subset is not None:
+            subset = self.bind(subset)
+        return ops.DropNull(self, how, subset).to_expr()
```
jcrist, line 2786: Should update the implementation to just call `self.dropnull` instead.

ibis/expr/types/relations.py
```diff
@@ -2756,7 +2842,7 @@ def fillna(
                         f"value of type {val_type} - pass in an explicit mapping "
                         f"of fill values to `fillna` instead."
                     )
-        return ops.FillNa(self, replacements).to_expr()
+        return ops.FillNull(self, replacements).to_expr()
```
jcrist, line 2845: Should update the implementation to just call `self.fillnull` instead.

ibis/expr/types/relations.py
```diff
@@ -2719,6 +2724,87 @@ def fillna(
...
+        if subset is not None:
+            subset = self.bind(subset)
+        return ops.DropNull(self, how, subset).to_expr()
+
+    @deprecated(as_of="10.0", instead="use fillnull instead")
+    def fillna(
+        self,
+        replacements: ir.Scalar | Mapping[str, ir.Scalar],
+    ) -> Table:
+        """Fill null values in a table expression.
```
jcrist, line 2793: IMO we should delete the whole docstring and replace it with

```
"""Deprecated - use `fillnull` instead"""
```

ibis/expr/types/relations.py
```diff
@@ -2719,6 +2724,87 @@ def fillna(
...
+                    )
+        return ops.FillNull(self, replacements).to_expr()
+
+    @deprecated(as_of="10.0", instead="use dropnull instead")
+    def dropna(
+        self,
+        subset: Sequence[str] | str | None = None,
+        how: Literal["any", "all"] = "any",
+    ) -> Table:
+        """Remove rows with null values from the table.
```
jcrist, line 2767: IMO we should delete the whole docstring and replace it with

```
"""Deprecated - use `dropnull` instead."""
```

jcrist (commented):
Just a few small comments, otherwise looks good to me.

I suspect we're not going to release a 9.1 release, but if we are we might want to hold off on merging this until after that since this would add new warnings (cc @cpcloud for a yay/nay on if we're doing a 9.1 release).

ncclementi:
> I suspect we're not going to release a 9.1 release 

We already got some deprecation warnings out with a deprecated since 9.1 for the register api so maybe we were planning on releasing that one? If we do happy to wait.

jcrist:
Oh, in that case then maybe we're fine with adding warnings on a point release? No strong thoughts, just want to make sure we're consistent here.

ncclementi:
I guess depending on when are we releasing 9.1 and if we are releasing, we should then change the warnings in this PR to say deprecated since `9.1`.

cpcloud:
Looking at the API here, I think `fillnull` should have an underscore (I'm not sure if we have words not separated like this in any other API), and I'm leaning towards `fill_nulls`, to indicate that more than one null can be filled.

Thoughts? @jcrist @ncclementi

jcrist:
> Looking at the API here, I think fillnull should have an underscore (I'm not sure if we have words not separated like this in any other API), and I'm leaning towards fill_nulls, to indicate that more than one null can be filled.

We already discussed this in #7276. I agree that I prefer the underscored versions for API consistency (and legibility). I also agree with @gforsyth that the plurality needs to be consistent between `fill_null` & `drop_null`. I have a slight preference for the non-plural versions.

ncclementi:
Looks like we have more votes for and `_` name, I will make this change. I personally don't have a preference for the plural vs singular versions, since Jim has a preference I'll go with the singular one. 

@cpcloud Since we plan to have this for release `9.1` I'll modify the warnings to be consistent with this.

docs/tutorials/ibis-for-pandas-users.qmd
```diff
@@ -532,17 +532,17 @@ the column name for the value to apply to.
...
 
 ```{python}
-no_null_peng = penguins.fillna(dict(bill_depth_mm=0, bill_length_mm=0))
+no_null_peng = penguins.fill_null(dict(bill_depth_mm=0, bill_length_mm=0))
 ```
 
 ### Replacing `NULL`s
 
-Both pandas and Ibis have `fillna` methods which allow you to specify a replacement value
+The Ibis equivalent of pandas `fillna` is `filnull`, this method allows you to specify a replacement value
```
gforsyth, line 540:
```suggestion
The Ibis equivalent of pandas `fillna` is `fill_null`, this method allows you to specify a replacement value
```

ibis/expr/types/relations.py
```diff
@@ -2756,7 +2808,7 @@ def fillna(
                         f"value of type {val_type} - pass in an explicit mapping "
                         f"of fill values to `fillna` instead."
                     )
-        return ops.FillNa(self, replacements).to_expr()
+        return self.fill_null(replacements)
```
jcrist, line 2811: You can delete the full implementation above and just keep this line - no need to duplicate any of the `fill_null` function body in the old deprecated method.

ibis/expr/types/relations.py
```diff
@@ -2740,7 +2740,59 @@ def fillna(
...
+    @deprecated(as_of="9.1", instead="use drop_null instead")
+    def dropna(
+        self,
+        subset: Sequence[str] | str | None = None,
+        how: Literal["any", "all"] = "any",
+    ) -> Table:
+        """Deprecated - use `drop_null` instead."""
+
+        if subset is not None:
+            subset = self.bind(subset)
```
jcrist, line 2770: No need for these lines here, just rely on `drop_null`.

```suggestion
```

gforsyth (commented):
Nice!  Just one typo mangled fix in the pandas tutorial

jcrist (approved):
:shipit:

PR #9305, opened 2024-06-04
fix(clickhouse): more explicitly disallow null structs

NickCrews:
I remove the docstring because it is obvious from the type signature,
and the docstring was wrong. The superclass has the correct docstring though,
just inherit that.

Part of https://github.com/ibis-project/ibis/pull/8666


cpcloud:
FYI docstrings are not inherited, but I still don't think we need it here :)

cpcloud (approved):

cpcloud:
Crushing it!

NickCrews:
ahh, thanks for the point re docstrings. VSCode is doing the same thing as ipython and shows the parent class's docstring when I hover over the method, but I see now that is being misleading as to what is actually happening :)

PR #9310, opened 2024-06-04
test: disallow empty structs

NickCrews:
working towards https://github.com/ibis-project/ibis/issues/8289

I'm not sure how useful empty structs are, since it seems like
only bigquery, dask, and pandas actually support them.
But still, if you stay in ibis-land, perhaps it is useful.
ie doing datatype kung-fu or only using them for intermediate calculations.
Not that hard for us to support it, so why not.

I'm not sure of the history of the specific disallowment
that I am removing from the type inference.

Relevant context:

- https://github.com/ibis-project/ibis/pull/8876
- https://github.com/ibis-project/ibis/issues?q=empty+struct

### DIFF
--- a/ibis/backends/dask/tests/conftest.py
+++ b/ibis/backends/dask/tests/conftest.py
@@ -151,9 +151,9 @@ def pandas_df():
             "array_of_float64": [[1.0, 2.0], [3.0], []],
             "array_of_int64": [[1, 2], [], [3]],
             "array_of_strings": [["a", "b"], [], ["c"]],
-            "map_of_strings_integers": [{"a": 1, "b": 2}, None, {}],
-            "map_of_integers_strings": [{}, None, {1: "a", 2: "b"}],
-            "map_of_complex_values": [None, {"a": [1, 2, 3], "b": []}, {}],
+            "map_of_strings_integers": [{"a": 1, "b": 2}, {}, None],
+            "map_of_integers_strings": [{1: "a", 2: "b"}, {}, None],
+            "map_of_complex_values": [{"a": [1, 2, 3], "b": []}, {}, None],
         }
     )
 
--- a/ibis/backends/dask/tests/test_maps.py
+++ b/ibis/backends/dask/tests/test_maps.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import numpy as np
 import pandas as pd
 import pytest
 
@@ -13,7 +12,9 @@ from dask.dataframe.utils import tm  # noqa: E402
 def test_map_length_expr(t):
     expr = t.map_of_integers_strings.length()
     result = expr.execute()
-    expected = pd.Series([0, None, 2], name="MapLength(map_of_integers_strings)")
+    expected = pd.Series(
+        [2, 0, None], dtype="object", name="MapLength(map_of_integers_strings)"
+    )
     tm.assert_series_equal(result, expected, check_index=False)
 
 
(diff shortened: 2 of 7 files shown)


ibis/backends/tests/test_struct.py
```diff
@@ -29,6 +31,14 @@
 ]
 
 
+@pytest.mark.notimpl("duckdb", raises=DuckDBParserException)
+@pytest.mark.notimpl("polars", raises=PolarsComputeError)
+def test_struct_factory_empty(con):
+    s = ibis.struct({})
+    result = con.execute(s)
+    assert result == {}
```
cpcloud, line 39: I don't think this is supported in many of our backends. How is this spelled in DuckDB for example?

NickCrews: moving to a top-level thread in this PR.

NickCrews:
> I don't think this is supported in many of our backends. How is this spelled in DuckDB for example?

Indeed, it is only supported on bigquery, dask, and pandas currently. Possibly could be implemented in other backends with some effort. But still, if you stay in ibis-land, perhaps it is useful. ie doing datatype kung-fu or only using them for intermediate calculations. Not that hard for us to support it, so why not.

Based off your [comment in this other PR](https://github.com/ibis-project/ibis/pull/8666#discussion_r1623224344), I was thinking that your desired behavior was to allow empty structs, so I was working towards that.

But fine switching these tests to test for disallowment instead, if that is the desired behavior. but I just want to nail down the semantics here before I go back to #8666

NickCrews:
@jcrist @kszucs @gforsyth curious what yall think about the costs/benefits of allowing/disallowing empty structs

cpcloud:
> Not that hard for us to support it, so why not.

Well, famous last words :)

> I was thinking that your desired behavior was to allow empty structs

I don't think I've seen a strong justification for their support in the form of a concrete use case, so I'd be in favor of dropping support for empty struct.

If we go down that route, we should also drop support for empty schemas for consistency, which also do not have any documented concrete user-facing use case.

cpcloud:
Someone really needs to provide a concrete use case for empty structs, given that effectively *one* backend supports it.

"PyArrow supports it" is not a strong enough reason to do something that isn't well supported across our backends at all.

NickCrews:
hmm, our pyarrow type roundtrip fuzzing tests are failing. Can I get some help with hypothesis to limit it to only give us fields of length 1+?

cpcloud:
@kszucs Can you comment on why empty structs are useful, independent of their existence in PyArrow? I still don't follow what their utility is from the perspective of an end user.

NickCrews:
Just to be clear, I swapped this PR to now be checking for disallowment

cpcloud:
Yep, jut wanting to keep discussion all in one place!

kszucs:
> @kszucs Can you comment on why empty structs are useful, independent of their existence in PyArrow? I still don't follow what their utility is from the perspective of an end user.

Struct values might be the outcome of other operations like converting a JSON typed column to a strictly typed struct typed column which would be fallible if empty struct typed are not supported. User defined functions may produce empty struct types as well.

cpcloud:
> Struct values might be the outcome of other operations like converting a JSON typed column to a strictly typed struct typed column which would be fallible if empty struct typed are not supported. User defined functions may produce empty struct types as well.

None of these operations can actually be executed on anything other than BigQuery, so being able to spell them in the type system is entirely moot.

jcrist:
What's the status of this PR? Does it supersede the work in #8876?

cpcloud:
> What's the status of this PR? Does it supersede the work in #8876?

I don't think so. This PR contains tests and a bit of implementation while 8876 contains more implementation work.

cpcloud:
That said, I'm going to close both of them out in an effort to set expectations on the amount of maintenance I can take on right now.

PR #9318, opened 2024-06-06
fix: drop nulls in `.collect()` aggregation

### LINKED ISSUE #9313: bug: `.collect()` handling of null values is backend-specific
The null handling behavior of `.collect()` is currently backend specific.

The `collect` method is an aggregation that returns an array of all items in the group. The difference in behavior here is what happens with `NULL` elements:

### Null behavior is configurable
- `bigquery`

### Nulls always excluded
- `clickhouse`
- `pyspark`
- `snowflake`

### Nulls included, but may be excluded with an extra filter
- `datafusion`
- `duckdb`
- `postgres`
- `risingwave`
- `trino`

Currently we just do whatever the backend does (except for `snowflake`, where there's a hacky hacky hack to keep nulls).

Since some backends have no way to include nulls (and most other aggregates exclude them), I'd vote to change the behavior in ibis to exclude nulls from `collect()` to unify this behavior across backends.

jcrist:
Fixes #9313, see that issue for the original context.

This PR standardizes the `null` handling behavior of our `.collect()` method to always exclude `NULL` values from the aggregated array. Previously the behavior was backend dependent.

This is _technically_ a breaking change for some backends. I currently haven't flagged it in the commit, but if we deem it worth flagging we probably also want to hold off merging if we plan on doing a 9.1 release (since a breaking change would need to go in 10.0). I'm inclined to treat this more as a behavioral bug, which is why I didn't mark it as breaking here.

I also haven't tested this on the cloud backends, :crossed_fingers: that it passes in CI.

### DIFF
--- a/ibis/backends/datafusion/compiler.py
+++ b/ibis/backends/datafusion/compiler.py
@@ -14,6 +14,7 @@ import ibis.expr.operations as ops
 from ibis.backends.sql.compiler import FALSE, NULL, STAR, AggGen, SQLGlotCompiler
 from ibis.backends.sql.datatypes import DataFusionType
 from ibis.backends.sql.dialects import DataFusion
+from ibis.backends.sql.rewrites import exclude_nulls_from_array_collect
 from ibis.common.temporal import IntervalUnit, TimestampUnit
 from ibis.expr.operations.udf import InputType
 from ibis.formats.pyarrow import PyArrowType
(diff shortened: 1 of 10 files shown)


jcrist:
> I also haven't tested this on the cloud backends, 🤞 that it passes in CI.

It is unclear to me why [this test failed](https://github.com/ibis-project/ibis/actions/runs/9394718191/job/25872873129?pr=9318), but other than that things look to be all green!

cpcloud (approved):

cpcloud:
I think considering the inconsistency as a bug is fine, we can always do another bug fix release if it's considered too onerous a change.

NickCrews:
just got bitten by this change. If we don't wait for a major release, this deserves a very prominent note in the release notes. I'd still say release it ASAP though.

Issue #9311, opened 2024-06-04
refactor: remove ibis.NA 


ncclementi:
Opening an issue to discuss a separate the refactor of `ibis.NA` because doing it as part of https://github.com/ibis-project/ibis/pull/9300 will get messy. 

To discuss:
Does removing `ibis.NA`imply:

1. Replace `ibis.NA` for `ibis.null()`
2. Get rid of untyped nulls all together, as in not use `ibis.null()` but just None

jcrist:
Another option is to rename it to `ibis.NULL`, but I'd vote for option 1.

I'm not really sure what option 2 means - in _many_ cases APIs that accept `None` should implicitly coerce it to a literal `NULL`, but not in all cases. There will always cases where calling `ibis.null()` (or `ibis.literal(None)`, which is the same thing) will be necessary or more ergonomic.

cpcloud:
I don't think can we remove `ibis.null()`, it accepts a dtype and can create a typed null which is necessary and/or useful in some cases.

I would vote for the following:

1. Remove NA
2. Don't add NULL
3. Use `None` wherever _untyped_ NULLs previously worked.

`ibis.null(...)` remains unchanged.

cpcloud:
As far implementation goes, we can basically search and replace `ibis.NA` with `ibis.null()`. If folks want an untyped NULL, that's `None` (after these changes).

ncclementi:
For implementation purposes, do we want to deprecate `ibis.NA`or remove it as a breaking change. If it's a deprecation, how do we deprecate `ibis.NA = null()` the `@deprecated` seems to only work for methods/functions.

PR #9323, opened 2024-06-06
refactor: remove ibis NA

ncclementi:
- Closes https://github.com/ibis-project/ibis/issues/9311

This PR removes `ibis.NA`, replacing it appearances for `ibis.null()` 

Merge after https://github.com/ibis-project/ibis/pull/9300 - I'll rebase based on main once merged


ncclementi:
I accidentally change the html from a doc freeze, that had a `ibis.NA` what's the best way to fix that? I've never generated a frozen page for docs.

jcrist:
Copying a question from the issue into the PR:

> For implementation purposes, do we want to deprecate ibis.NAor remove it as a breaking change. If it's a deprecation, how do we deprecate ibis.NA = null() the @deprecated seems to only work for methods/functions.

I think most users never touch `ibis.NA`, so I wouldn't be *opposed* to doing a clean break. In that case though I'd think we'd want to wait until 10.0 since that would definitely be a breaking change. Given `fillna`/`dropna` are cleanly deprecated in the other PR, I have a slight preference for adding a warning here as well so they all can be released at the same time.

Two ways I can see to do a cleaner deprecation:

- Expose `ibis.NA` via a module `__getattr__` on the `ibis` namespace, and warn when that attribute is accessed. This would warn every time `ibis.NA` or `from ibis import NA` is executed.
- Create a new `NAScalar` type (like the existing `NullScalar`) that warns in the `.op()` method before returning the existing op representation. This would warn every time `ibis.NA` is used to build up larger expressions.

I think the former is a better deprecation method, but either would work well enough.

ibis/__init__.py
```diff
@@ -125,3 +128,18 @@ def connect(*args, **kwargs):
         setattr(proxy, name, getattr(backend, name))
 
     return proxy
+
+
+def __getattr__(name: str) -> Any:
```
ncclementi, line 133: Since we already had a `__getattr__` I went this route to keep the typing for the `load_backend` portion. Once we deprecate NA completely we can remove the workaround

ncclementi:
I'll rebase here once we merged https://github.com/ibis-project/ibis/pull/9300

ibis/__init__.py
```diff
@@ -125,3 +128,18 @@ def connect(*args, **kwargs):
...
+    if name == "NA":
+        warnings.warn(
+            "Accessing 'ibis.NA' is deprecated as of v9.1 and will be removed in a future version. "
+            "Use 'ibis.null()' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        import ibis
+
+        return ibis.null()
```
jcrist, line 143: I _think_ this should work instead, since `null` is already imported above:

```suggestion
        return null()
```

ibis/__init__.py
```diff
@@ -125,3 +128,18 @@ def connect(*args, **kwargs):
         setattr(proxy, name, getattr(backend, name))
 
     return proxy
+
+
+def __getattr__(name: str) -> Any:
+    if name == "NA":
+        warnings.warn(
+            "Accessing 'ibis.NA' is deprecated as of v9.1 and will be removed in a future version. "
+            "Use 'ibis.null()' instead.",
```
jcrist, line 137:
```suggestion
            "The 'ibis.NA' constant is deprecated as of v9.1 and will be removed in a future "
            "version. Use 'ibis.null()' instead.",
```

ncclementi:
The amount of conflicts got a bit nasty, it'll be faster to open a new PR and make it cleaner. I'll put this up in a few min.

ibis/__init__.py
ncclementi: Has it? I don't see an import of null.

PR #9330, opened 2024-06-07
docs: add Theseus in sankey diagrams

lostmygithubaccount:
## Description of changes

adds Voltron Data's Theseus engine as a SQL backend in the sankey diagrams


cpcloud:
This seem a bit out of place given that every other backend in the Sankey diagram has:

1. Its own page under the `Backends` tab with links, information etc.
2. A column in the operations matrix
3. Exists as a backend in the main repo.

I'm not sure what value we're adding by putting this in here given how little information only putting this in the diagram adds.

lostmygithubaccount:
that's fair -- I'd also prefer it's open sourced before putting it there, but didn't see much harm in adding it. I'll close this out for now

Issue #8672, opened 2024-03-15
feat: Make Table a Mapping


NickCrews:
### Is your feature request related to a problem?

Similar to #8671 

Specifically, I want to be able to to `s = ibis.struct(my_table)`

### Describe the solution you'd like

This isn't going to work unless we allow `__len__()` in Table. But that opens up users to misusing len(t), when they want to use .count(). Probably not worth it, that would be a really annoying and would get a LOT of people.

So perhaps we just forget this? IDK is there something else we could do clever?

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #9332, opened 2024-06-07
feat: make Tables a Mapping

NickCrews:
Fixes https://github.com/ibis-project/ibis/issues/8672

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -516,6 +516,24 @@ class Table(Expr, _FixedTextJupyterMixin):
             console_width=console_width,
         )
 
+    def keys(self) -> list[str]:
+        """Return the column names of the table.
+
+        Returns
+        -------
+        list
+            Column names
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.examples.penguins.fetch()
+        >>> t.keys()[:3]
+        ['species', 'island', 'bill_length_mm']
+        """
+        return self.columns
+
     def __getitem__(self, what):
         """Select items from a table expression.
 
--- a/ibis/tests/expr/test_table.py
+++ b/ibis/tests/expr/test_table.py
@@ -123,6 +123,19 @@ def test_table_tab_completion():
     assert items.issuperset(table.columns)
 
 
+def test_keys(table):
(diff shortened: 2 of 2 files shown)


ibis/expr/types/relations.py
```diff
@@ -516,6 +516,24 @@ def preview(
...
+
+        Returns
+        -------
+        list
+            Column names
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
```
cpcloud, line 530:
```suggestion
```

This isn't necessary, you're not displaying any Ibis expressions.

cpcloud (changes requested):

ibis/tests/expr/test_table.py
```diff
@@ -123,6 +123,19 @@ def test_table_tab_completion():
     assert items.issuperset(table.columns)
 
 
+def test_keys(table):
+    keys = table.keys()
+    assert keys == table.columns
+
+
```
cpcloud, line 130:
```suggestion
```

The test below seems sufficient. Alternatively you can put this assertion there.

cpcloud (changes requested):

cpcloud:
Are you planning on making `Table` a proper `Mapping` subclass, or are you just interested in the kwarg unpacking behavior?

jcrist:
If you're only interested in kwarg unpacking in `ibis.struct`, could we instead just make `ibis.struct(table)` work (adding a special case for a `Table` arg) rather than relying implicitly on kwarg unpacking? This will be both more efficient (since we won't need to do the work to validate all columns share a parent relation) and will let us avoid the addition of a `.keys()` method on a table (which IMO feels pretty weird).

FWIW my thoughts [here](https://github.com/ibis-project/ibis/pull/8671#issuecomment-2007967971) on making structs a mapping apply equally to the case here.

NickCrews:
I talk about most of these concerns in https://github.com/ibis-project/ibis/issues/8672.

> only interested in kwarg unpacking in ibis.struct

Not just that, I want to take advantage of generic ** unpacking to do `{"col1": "default_that_might_be_overwritten", **table, "col2": "always_this_val"}`.

NickCrews:
closing for now as stale.

PR #9333, opened 2024-06-07
chore(refactor): handle 0-length args in highest_precedence_shape()

NickCrews:
moving towards https://github.com/ibis-project/ibis/pull/8666

### DIFF
--- a/ibis/expr/rules.py
+++ b/ibis/expr/rules.py
@@ -5,6 +5,7 @@ from typing import Optional
 
 from public import public
 
+import ibis.expr.datashape as ds
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 from ibis import util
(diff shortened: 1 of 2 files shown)


ibis/expr/rules.py
```diff
@@ -16,6 +17,9 @@
 
 @public
 def highest_precedence_shape(nodes):
+    nodes = tuple(nodes)
+    if len(nodes) == 0:
```
cpcloud, line 21:
```suggestion
    if not nodes:
```

cpcloud (changes requested):

cpcloud:
What's the rationale for the *general* case of no arguments being `scalar` versus the empty array case which seems clear that it should be.

This seems like it should raise an exception in the general case and we should specialize the semantics in cases where it makes sense.

NickCrews:
hmm I think you are right, we want to be explicit and have users be resonsible for dealing with this edge case. Imagine there is some case we haven't found that causes a 0-length arg. With this change, that would now give scalar, even if that might be silently wrong. I think we would rather have that undiscovered situation do a hard error, then get a bug report about it.

slightly related as I explored this more: #9334

PR #9334, opened 2024-06-07
fix(datatypes): infer the correct shape for many existing ops

NickCrews:
Found this when looking into https://github.com/ibis-project/ibis/pull/9333

Usually when using eg duckdb's REGEXP_REPLACE() function, the pattern and replacement are always scalar constants. But actually it can accept columnar for all three of it's arguments.

These sort of assumptions appeared to be all over the place in the code base. I found these fixs by grepping for "shape_like" and manually looking at all the instances.

I didn't add any test cases, that felt like a monumental thing to do I wasn't willing to put the time in for. But IDK, I'm not sure I would even want to put in all the tests, it would be SO much boilerplate.

To deal with Nodes sometimes having optional args, I modified shape_like() to be more flexible.
IDK, that was sort of a lazy approach, we could be much more verbose and have each op do this filtering individually, but I didn't that was worth it.

### DIFF
--- a/ibis/expr/rules.py
+++ b/ibis/expr/rules.py
@@ -77,6 +77,7 @@ def shape_like(name):
     def shape(self):
         args = getattr(self, name)
         args = args if util.is_iterable(args) else [args]
+        args = [a for a in args if a is not None]
         return highest_precedence_shape(args)
 
     return shape
--- a/ibis/expr/operations/arrays.py
+++ b/ibis/expr/operations/arrays.py
@@ -48,7 +48,7 @@ class ArraySlice(Value):
     stop: Optional[Value[dt.Integer]] = None
 
     dtype = rlz.dtype_like("arg")
-    shape = rlz.shape_like("arg")
+    shape = rlz.shape_like("args")
 
 
 @public
@@ -71,13 +71,11 @@ class ArrayConcat(Value):
 
     arg: VarTuple[Value[dt.Array]]
 
+    shape = rlz.shape_like("arg")
+
     @attribute
     def dtype(self):
         return dt.Array(dt.highest_precedence(arg.dtype.value_type for arg in self.arg))
-
-    @attribute
-    def shape(self):
-        return rlz.highest_precedence_shape(self.arg)
 
 
 @public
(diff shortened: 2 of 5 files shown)


ibis/expr/rules.py
```diff
@@ -77,6 +77,7 @@ def shape_like(name):
     def shape(self):
         args = getattr(self, name)
         args = args if util.is_iterable(args) else [args]
+        args = [a for a in args if a is not None]
```
cpcloud, line 80: Can we handle this in `highest_precedence_shape`?

cpcloud (changes requested):
Thanks for this, I think it's a good idea!

Just a question about where we should handle the `None` check.

ibis/expr/rules.py
NickCrews: I was taking your cue from https://github.com/ibis-project/ibis/pull/9333#issuecomment-2155523018 and trying to only push this logic as far down as it needs to go. For example, Operations with optional arguments can still use `shape_like()` and this deals with that case, so those ops don't need to add in their own filtering logic. I thought that was common enough between Operations that it was worth factoring to here. But, other places in ibis besides Values might want to use `highest_precedence_shape`, and I think they should have to deal with `None`s in their own way. If one of those users doesn't deal with Nones, we want that to raise an error, not silently give the wrong answer.

This logic was also why I didn't add `args = [a for a in args if a is not None]` to `dtype_like()` above: It isn't needed yet, and if a user of dtyp_like ever starts accidentally passing in Nones, I want there to be an error so that we can notice and then decide what to do.

I can push this down to `highest_precedence_shape` though if you still think that's a good idea.

NickCrews: Ironically, now that you are making me think of it, I actually now want to push the `args = args if util.is_iterable(args) else [args]` logic down into `highest_precedence_shape`, because I think that would NEVER be a footgun. What do you think, should I do that (maybe in a separate PR)?

NickCrews: oh, also blocking for #9096

@cpcloud I am fine with pushing logic down a level further, just let me know what you think

cpcloud: This is fine as is for now! Thanks!

cpcloud (approved):
LGTM, I'll rebase and merge when CI passes.

NickCrews:
@cpcloud hopefully that was OK for me to hit the merge button, the CI passed! Thank you!

PR #9342, opened 2024-06-10
chore(api): remove defunct add_operation from backends

### LINKED ISSUE #9158: bug: `add_operation` attempts to call into removed `translator_class`
### What happened?

Previously (under Ibis 7.2.0), we had defined a `SAFE_CAST` operation for our usage of Ibis with bigquery with the following:

```python
import ibis
import ibis.expr.rules as rlz
from ibis.backends.bigquery.datatypes import BigQueryType
from ibis.expr.operations.core import Value as ValueOp
from ibis.expr.types import Value as ValueType


class SafeCast(ValueOp):
    """Safe cast value or column to a specific data type."""

    arg: ValueOp
    to: dt.DataType

    shape = rlz.shape_like("arg")

    @property
    def name(self):  # noqa
        return f"SAFE_CAST({self.arg.name} as {self.to})"

    @property
    def dtype(self):  # noqa
        return self.to


def safe_cast(col, dtype):
    """Safely cast a column to a specific data type."""
    return SafeCast(col, dtype).to_expr()


@ibis.bigquery.add_operation(SafeCast)
def _safe_cast_gbq(translator, expr):
    arg, target_dtype = expr.args
    compiled_arg = translator.translate(arg)
    sql_type = BigQueryType.from_ibis(target_dtype)
    return f"SAFE_CAST({compiled_arg} as {sql_type})"


ValueType.safe_cast = safe_cast
```

When attempting to update to Ibis 9.0.0, this broke with the error:
`AttributeError: 'BigQueryCompiler' object has no attribute 'translator_class'`

Not sure if this is a bug or an intended change in the way a user can add an operation to Ibis

7.2.0 originally, 9.0.0 for the error

BigQuery

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
This functionality was never ported to work with the new compiler
infrastructure in 9.0.

Not marking as a breaking change because the breakage is already done.

Resolves #9158

### DIFF
--- a/ibis/__init__.py
+++ b/ibis/__init__.py
@@ -95,7 +95,6 @@ def __getattr__(name: str) -> BaseBackend:
     # - connect
     # - compile
     # - has_operation
-    # - add_operation
     # - _from_url
     # - _to_sqlglot
     #
@@ -116,7 +115,6 @@ def __getattr__(name: str) -> BaseBackend:
     proxy.connect = connect
     proxy.compile = backend.compile
     proxy.has_operation = backend.has_operation
-    proxy.add_operation = backend.add_operation
     proxy.name = name
     proxy._from_url = backend._from_url
     proxy._to_sqlglot = backend._to_sqlglot
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -20,7 +20,7 @@ from ibis import util
 from ibis.common.caching import RefCountedCache
 
 if TYPE_CHECKING:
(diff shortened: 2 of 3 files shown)


jcrist (approved):

LennartKloppenburg:
Hello, 

I have no idea when this will be read and by whom, but I was wondering what is now the way to add custom methods to different backends after cleaning up the `add_operation` method

I.e. I want to add `JUSTIFY_INTERVAL` to the BigQuery backend but I can't find a guide on how to do this.
Thanks!

PR #9344, opened 2024-06-10
refactor: remove ibisNA

### LINKED ISSUE #9311: refactor: remove ibis.NA

ncclementi:
- Closes #9311 

This PR removes ibis.NA, replacing it appearances for ibis.null()


jcrist (approved):
One small nit, otherwise LGTM!

PR #9348, opened 2024-06-11
feat(postgres): provide translation for `hash` ops

deepyaman:
## Description of changes

Postgres supports hashing on a per-type basis, with "extended" versions providing 64-bit integer output. To list these, run `\df hash*extended`:

   Schema   |           Name           | Result data type |   Argument data types    | Type 
----------- | ------------------------ | ---------------- | ------------------------ | -----
 pg_catalog | hash_aclitem_extended    | bigint           | aclitem, bigint          | func
 pg_catalog | hash_array_extended      | bigint           | anyarray, bigint         | func
 pg_catalog | hash_multirange_extended | bigint           | anymultirange, bigint    | func
 pg_catalog | hash_numeric_extended    | bigint           | numeric, bigint          | func
 pg_catalog | hash_range_extended      | bigint           | anyrange, bigint         | func
 pg_catalog | hash_record_extended     | bigint           | record, bigint           | func
 pg_catalog | hashbpcharextended       | bigint           | character, bigint        | func
 pg_catalog | hashcharextended         | bigint           | "char", bigint           | func
 pg_catalog | hashenumextended         | bigint           | anyenum, bigint          | func
 pg_catalog | hashfloat4extended       | bigint           | real, bigint             | func
 pg_catalog | hashfloat8extended       | bigint           | double precision, bigint | func
 pg_catalog | hashinetextended         | bigint           | inet, bigint             | func
 pg_catalog | hashint2extended         | bigint           | smallint, bigint         | func
 pg_catalog | hashint4extended         | bigint           | integer, bigint          | func
 pg_catalog | hashint8extended         | bigint           | bigint, bigint           | func
 pg_catalog | hashmacaddr8extended     | bigint           | macaddr8, bigint         | func
 pg_catalog | hashmacaddrextended      | bigint           | macaddr, bigint          | func
 pg_catalog | hashnameextended         | bigint           | name, bigint             | func
 pg_catalog | hashoidextended          | bigint           | oid, bigint              | func
 pg_catalog | hashoidvectorextended    | bigint           | oidvector, bigint        | func
 pg_catalog | hashtextextended         | bigint           | text, bigint             | func
 pg_catalog | hashtidextended          | bigint           | tid, bigint              | func
 pg_catalog | hashvarlenaextended      | bigint           | internal, bigint         | func

This PR implements as many of them as possible, from the set of available Ibis datatypes in ibis/backends/sql/datatypes.py. No attempt to cast is made (e.g. hashing booleans is unsupported, rather than casting to an integer and then using the available hash method).

### DIFF
--- a/ibis/backends/postgres/compiler.py
+++ b/ibis/backends/postgres/compiler.py
@@ -588,3 +588,26 @@ class PostgresCompiler(SQLGlotCompiler):
         return self.cast(arg, op.to)
 
     visit_TryCast = visit_Cast
+
+    def visit_Hash(self, op, *, arg):
+        arg_dtype = op.arg.dtype
+
+        if arg_dtype.is_int16():
+            return self.f.hashint2extended(arg, 0)
+        elif arg_dtype.is_int32():
+            return self.f.hashint4extended(arg, 0)
+        elif arg_dtype.is_int64():
+            return self.f.hashint8extended(arg, 0)
+        elif arg_dtype.is_float32():
+            return self.f.hashfloat4extended(arg, 0)
+        elif arg_dtype.is_float64():
+            return self.f.hashfloat8extended(arg, 0)
+        elif arg_dtype.is_string():
+            return self.f.hashtextextended(arg, 0)
+        elif arg_dtype.is_macaddr():
+            return self.f.hashmacaddr8extended(arg, 0)
+
+        raise com.UnsupportedOperationError(
+            f"Hash({arg_dtype!r}) operation is not supported in the "
+            f"{self.dialect} backend"
+        )
--- a/ibis/backends/postgres/compiler.py
+++ b/ibis/backends/postgres/compiler.py
@@ -592,16 +592,8 @@ class PostgresCompiler(SQLGlotCompiler):
     def visit_Hash(self, op, *, arg):
         arg_dtype = op.arg.dtype
 
-        if arg_dtype.is_int16():
-            return self.f.hashint2extended(arg, 0)
-        elif arg_dtype.is_int32():
-            return self.f.hashint4extended(arg, 0)
-        elif arg_dtype.is_int64():
-            return self.f.hashint8extended(arg, 0)
-        elif arg_dtype.is_float32():
-            return self.f.hashfloat4extended(arg, 0)
-        elif arg_dtype.is_float64():
-            return self.f.hashfloat8extended(arg, 0)
+        if arg_dtype.is_numeric():
+            return self.f.hash_numeric_extended(arg, 0)
         elif arg_dtype.is_string():
             return self.f.hashtextextended(arg, 0)
         elif arg_dtype.is_macaddr():
--- a/ibis/backends/postgres/compiler.py
+++ b/ibis/backends/postgres/compiler.py
@@ -592,8 +592,16 @@ class PostgresCompiler(SQLGlotCompiler):
     def visit_Hash(self, op, *, arg):
         arg_dtype = op.arg.dtype
 
-        if arg_dtype.is_numeric():
-            return self.f.hash_numeric_extended(arg, 0)
+        if arg_dtype.is_int16():
+            return self.f.hashint2extended(arg, 0)
+        elif arg_dtype.is_int32():
+            return self.f.hashint4extended(arg, 0)
+        elif arg_dtype.is_int64():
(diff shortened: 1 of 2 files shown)


ibis/backends/tests/test_generic.py
```diff
@@ -1501,10 +1501,13 @@ def test_distinct_on_keep_is_none(backend, on):
         "trino",  # checksum returns varbinary
     ]
 )
-def test_hash(backend, alltypes):
+@pytest.mark.parametrize(
+    "dtype", ["smallint", "int", "bigint", "float", "double", "string"]
```
deepyaman, line 1505: I'm just not testing `macaddr` here, since `alltypes` doesn't have that column. Should I just add a `macaddr` column?

On top of that, I also think something is off with `macaddr` to begin with, though, since `PostgresType.from_ibis(dt.macaddr)` returns `DataType(this=Type.VARCHAR)`, and `test_macaddr_literal` in ibis/backends/tests/test_network.py expects a text type to be returned; I think the change should be bigger, but wanted to get an opinion first. :)

ibis/backends/postgres/compiler.py
```diff
@@ -588,3 +588,26 @@ def visit_Cast(self, op, *, arg, to):
         return self.cast(arg, op.to)
 
     visit_TryCast = visit_Cast
+
+    def visit_Hash(self, op, *, arg):
+        arg_dtype = op.arg.dtype
+
+        if arg_dtype.is_int16():
+            return self.f.hashint2extended(arg, 0)
```
deepyaman, line 596: Postgres `hash*extended` functions require a seed to be passed. I've just passed 0, since Ibis `hash` doesn't take a seed argument.

ibis/backends/postgres/compiler.py
```diff
@@ -588,3 +588,26 @@ def visit_Cast(self, op, *, arg, to):
...
+        elif arg_dtype.is_int32():
+            return self.f.hashint4extended(arg, 0)
+        elif arg_dtype.is_int64():
+            return self.f.hashint8extended(arg, 0)
+        elif arg_dtype.is_float32():
+            return self.f.hashfloat4extended(arg, 0)
+        elif arg_dtype.is_float64():
+            return self.f.hashfloat8extended(arg, 0)
```
gforsyth, line 604: Would all of these and also decimal types be covered by `hashnumericextended`?

deepyaman: Oh, true...

```suggestion
        if arg_dtype.is_numeric():
            return self.f.hash_numeric_extended(arg, 0)
```

ibis/backends/tests/test_generic.py
gforsyth: adding a column to `alltypes` will be a headache -- I think we could add a standalone macaddr hash test, but that can also happen in a followup

gforsyth (approved):

ibis/backends/tests/test_generic.py
deepyaman: Oh, sorry, what I had in mind was that can just add the column to the `alltypes` table inside the test, but a standalone test would also work.

In any case, yes, happy to do it in a follow-up, just because I'm not sure about some of the other behaviors mentioned for `macaddr` (and also because I think that's a much less important use case). 😅

ibis/backends/postgres/compiler.py
cpcloud: Why are the `extended` versions required?

cpcloud: Oh right, to get the 64 bit versions.

ibis/backends/tests/test_generic.py
```diff
@@ -1504,10 +1504,13 @@ def test_distinct_on_keep_is_none(backend, on):
...
+)
+def test_hash(backend, alltypes, dtype):
     # check that multiple executions return the same result
-    h1 = alltypes.string_col.hash().execute(limit=20)
-    h2 = alltypes.string_col.hash().execute(limit=20)
+    h1 = getattr(alltypes, f"{dtype}_col").hash().execute(limit=20)
+    h2 = getattr(alltypes, f"{dtype}_col").hash().execute(limit=20)
```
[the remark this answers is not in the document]
deepyaman: True.

PR #9350, opened 2024-06-11
chore(deps): upgrade lonboard and remove leafmap dep

ncclementi:
I had couple of conflicts, so I didn't want to push directly to the original PR. https://github.com/ibis-project/ibis/pull/9137 but I think this should be ok. 

This PR includes changes to fix the plotting in geospatial blog.


cpcloud (approved):
Niiiiice. Lots of transitive dependencies removed❗

PR #9352, opened 2024-06-11
ci: replace apt-get with homebrew

cpcloud:
Simplify the CI system deps installation a bit by using homebrew instead of apt-get.


cpcloud:
This is waaaaaaay slower (> 120s) than using `apt-get` (~20s), closing.

PR #9358, opened 2024-06-11
fix: exclude null values from `first` and `last` aggregations

jcrist:
Similar to #9318. The handling of `NULL` values in `first`/`last` previously was backend specific. We now standardize all backends to exclude `NULL` values from these aggregations.

### DIFF
--- a/ibis/backends/dask/kernels.py
+++ b/ibis/backends/dask/kernels.py
@@ -16,6 +16,14 @@ rowwise = {
     ops.DateAdd: lambda row: row["left"] + row["right"],
 }
 
+
+def maybe_pandas_reduction(func):
+    def inner(df):
+        return df.reduction(func) if isinstance(df, dd.Series) else func(df)
+
+    return inner
+
+
 reductions = {
     **pandas_kernels.reductions,
     ops.Mode: lambda x: x.mode().loc[0],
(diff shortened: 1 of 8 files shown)


cpcloud (approved):
LGTM!

jcrist:
Gah, this has surfaced a `datafusion` bug: https://github.com/apache/datafusion/issues/10878.

A few options:

- `xfail` the bad queries for now. They pass fine on earlier versions of `datafusion`, and should pass again in the future
- Drop `FILTER` clause usage in datafusion (maybe only for the faulty version?) in favor of using `ifelse` and `NULL` exclusion. This would ensure `datafusion` code works on all `datafusion` releases, with some slight added code complexity for us.

gforsyth:
> `xfail` the bad queries for now. They pass fine on earlier versions of `datafusion`, and should pass again in the future

I vote for marking this as a `notyet`

cpcloud:
> Drop FILTER clause usage in datafusion (maybe only for the faulty version?) in favor of using ifelse and NULL exclusion. This would ensure datafusion code works on all datafusion releases, with some slight added code complexity for us.

If you've already got this working, then great! I remember looking into it awhile ago and it not being particularly straightforward to get working.

cpcloud:
Also

> xfail the bad queries for now. They pass fine on earlier versions of datafusion, and should pass again in the future

SGTM.

PR #9361, opened 2024-06-12
feat(duckdb): use `delta_scan` instead of reading pyarrow datasets

cpcloud:
Use the new `delta` extension and `delta_scan` to read deltalake files with DuckDB. Unfortunately there is a bug with reading timestamps **without** timezones, where the extension interprets those as having a timezone. I will file a bug report.


cpcloud:
Closing this out until the extension gets native support on Windows. Rather not end up doing the native-vs-pyarrow branching that we do elsewhere.

PR #9362, opened 2024-06-12
ci(clickhouse): use a different health check

cpcloud:
For some reason the existing clickhouse health check is not working on GHA, despite it working locally for me. Using the specific ipv4 address works in both case. Probably some DNS or other networking detail has changed and is leaking through from the host configuration into the container.


cpcloud:
Merging on green to get CI passing again.

cpcloud:
Both ClickHouse jobs are green, merging!

PR #9366, opened 2024-06-12
docs(search): append scraped API records to algolia index in CI

gforsyth:
There's a larger issue here, which is that the quarto `search.json`
doesn't seem to include a bunch of items which we generate using
`quartodoc`, which makes the docs very unhelpful for someone trying to
search (especially) method names.

This is definitely a hack, but I've tried uploading these records
manually and it does make a noticeable improvement.

And yes, I am scraping through QMD files to grab the anchors, and
descriptions, and method names, and yes that's gross, but computers are gross.

I'm planning to spend a bit more time to try to better understand how we
can better augment the algolia index so our search is more useful, but
this is both a start, and a proof-of-concept that we can append to our
existing index.

### DIFF
--- a/.github/workflows/ibis-docs-main.yml
+++ b/.github/workflows/ibis-docs-main.yml
@@ -69,9 +69,17 @@ jobs:
         run: |
           python -m pip install --upgrade algoliasearch
 
-      - name: Create and Upload Index
+      - name: Create and Upload Base Index
         run: |
           python .github/workflows/upload-algolia.py
+        env:
+          ALGOLIA_WRITE_API_KEY: ${{ secrets.ALGOLIA_WRITE_API_KEY }}
+          ALGOLIA_APP_ID: HS77W8GWM1
+          ALGOLIA_INDEX: prod_ibis
+
+      - name: Create and Upload API Records to index
+        run: |
+          python .github/workflows/upload-algolia-api.py
         env:
           ALGOLIA_WRITE_API_KEY: ${{ secrets.ALGOLIA_WRITE_API_KEY }}
           ALGOLIA_APP_ID: HS77W8GWM1
--- a/.github/workflows/upload-algolia-api.py
+++ b/.github/workflows/upload-algolia-api.py
@@ -1,0 +1,93 @@
+from __future__ import annotations  # noqa: INP001
+
+import os
+import re
+from functools import partial
+
+from algoliasearch.search_client import SearchClient
+
+api_key = os.environ["ALGOLIA_WRITE_API_KEY"]
+app_id = os.environ["ALGOLIA_APP_ID"]
+index_name = os.environ["ALGOLIA_INDEX"]
+
+
+# These are QMD files generated with help from Quartodoc.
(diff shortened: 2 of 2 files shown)


gforsyth:
I ran this script locally and generated a `records.json` file which I then uploaded via the algolia dashboard, so you can check that method names now show up prominently in the search bar (so long as another PR doesn't get merged and reset the index...)

.github/workflows/upload-algolia-api.py
```diff
@@ -0,0 +1,72 @@
...
+    "docs/reference/expression-collections.qmd",
+    "docs/reference/expression-generic.qmd",
+    "docs/reference/expression-geospatial.qmd",
+    "docs/reference/expression-numeric.qmd",
+    "docs/reference/expression-strings.qmd",
+    "docs/reference/expression-tables.qmd",
+    "docs/reference/expression-temporal.qmd",
+]
+
+HORRID_REGEX = re.compile(r"\|\s*\[(\w+)\]\((#[\w.]+)\)\s*\|\s*(.*?)\s*\|")
```
jcrist, line 24: :scream:

.github/workflows/ibis-docs-main.yml
```diff
@@ -69,10 +69,18 @@ jobs:
...
           ALGOLIA_APP_ID: HS77W8GWM1
           ALGOLIA_INDEX: prod_ibis
+
+      - name: Create and Upload API Records to index
+        run: |
+          python .github/workflows/upload-algolia-api.py
+        env:
+          ALGOLIA_WRITE_API_KEY: ${{ secrets.ALGOLIA_WRITE_API_KEY }}
+          ALGOLIA_APP_ID: HS77W8GWM1
+          ALGOLIA_INDEX: prod_ibis
```
jcrist, line 86: Is there a way to avoid duplicating these envs across steps? Like a top-level `env` mapping? Not a big deal.

.github/workflows/upload-algolia-api.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import annotations  # noqa: INP001
```
jcrist, line 1: This honestly isn't that horrid of a script. A few more comments/docstrings to explain the method to the method madness would help though. "This script generates records for algolia to search for all methods/functions because ...., the records look like ...."

jcrist (approved):
All code is hacks and this is an effective one. +1 from me.

cpcloud (approved):
I agree with Jim, I've seen and done much worse 😂

.github/workflows/ibis-docs-main.yml
gforsyth: probably, I'll try to consolidate in a follow-up when i add a few more tweaks to the algolia index creation.

.github/workflows/upload-algolia-api.py
gforsyth: Tried to document a bit more.  Time to :ship:

PR #9372, opened 2024-06-13
ci: make sure args is never empty

cpcloud:
Apparently empty arrays do not expand in the way I expected.


cpcloud:
Passing on the macos-14 nix build, merging.

PR #9377, opened 2024-06-13
chore(docs): actually run the algolia index augmenter

gforsyth:
Don't worry, I'm a professional.


cpcloud (approved):
Running it is indeed a good idea!

Issue #9378, opened 2024-06-13
bug(oracle): Ibis assumes that a boolean type exists in the version of oracle being used 


cpcloud:
### What happened?

I can't use Oracle <23c with Ibis because of its use of boolean columns, which were only added in very recent versions of Oracle.

### What version of ibis are you using?

`main`

### What backend(s) are you using, if any?

Oracle

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
Closing this out, I don't think we're going to spend a ton of time supporting Oracle-without-bools.

Issue #9179, opened 2024-05-13
bug: Oracle Table alias


Baazion:
### What happened?

When Selecting on Oracle query gets parsed with AS for Table-Alias, which is not supported in Oracle

### What version of ibis are you using?

9.0.0

### What backend(s) are you using, if any?

Oracle

### Relevant log output

```sh

services = conn.table("SOME_TABLE").select("sst_nr")
expr = services.count()
print(ibis.to_sql(expr))

SELECT
  COUNT(*) AS "CountStar()"
FROM (
  SELECT
    "t0"."sst_nr"
  FROM "SOME_TABLE" AS "t0"
) AS "t1"

oracledb.exceptions.DatabaseError: ORA-00923: FROM keyword not found where expected

Works when manually executing in Oracle:
SELECT
  COUNT(*) AS "CountStar()"
FROM (
  SELECT
    "t0"."SST_NR"
  FROM some_table "t0"
) "t1"
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gforsyth:
Hi @Baazion -- thanks for reporting this!  Can you tell us which version of Oracle you are running against?

We only have a Docker image running 23.4 and table aliasing works as expected there -- I'm not sure which version this was added in?

Baazion:
`SELECT * FROM v$version;`

> `Oracle Database 19c Standard Edition 2 Release 19.0.0.0.0 - Production`

gforsyth:
Ok, well, 19c is a LTS release and they promise support for another 2 years, so we can try to remove keyword that's causing the syntax error.  Thanks, @Baazion !

cpcloud:
This may need to be reported upstream to SQLGlot, and hopefully in the mean time there's a SQLGlot compiler flag we can use to set the aliasing style.

gforsyth:
Hmm, looks like `sqlglot` does the right thing -- I wonder where we're adding the `AS`:

```python
[nav] In [3]: expr = sg.parse_one(
         ...:     '''SELECT
         ...:   COUNT(*) AS "CountStar()"
         ...: FROM (
         ...:   SELECT
         ...:     "t0"."sst_nr"
         ...:   FROM "SOME_TABLE" AS "t0"
         ...: ) AS "t1"''',
         ...:     read="oracle",
         ...: )

[ins] In [4]: expr.sql(dialect="oracle")
[...]

PR #9379, opened 2024-06-13
fix(oracle): use case statements to grab table metadata instead of bools

gforsyth:
xref: #9179 #9378 

Definitely more fixes needed, I just didn't want to lose track of this branch.

I was able to get some sense of the failures by switching the Oracle docker container:

```diff
       - druid
 
   oracle:
-    image: gvenzl/oracle-free:23.4-slim
+    image: gvenzl/oracle-xe:18.4.0-slim
     environment:
       ORACLE_PASSWORD: ibis
       ORACLE_DATABASE: IBIS_TESTING
```

I couldn't find a version 19 container.
Also note that this requires a bunch of annoying changes to the conftest setup because Oracle 18 doesn't support `DROP TABLE IF EXISTS`

### DIFF
--- a/ibis/backends/oracle/__init__.py
+++ b/ibis/backends/oracle/__init__.py
@@ -624,3 +624,18 @@ class Backend(SQLBackend, CanListDatabase, CanListSchema):
 
     def _clean_up_cached_table(self, op):
         self._clean_up_tmp_table(op.name)
+
+    def _to_sqlglot(
+        self, expr: ir.Expr, *, limit: str | None = None, params=None, **_: Any
+    ):
+        """Compile an Ibis expression to a sqlglot object."""
+        table_expr = expr.as_table()
+        conversions = {
+            name: ibis.ifelse(table_expr[name], 1, 0)
+            for name, typ in table_expr.schema().items()
+            if typ.is_boolean()
+        }
+
+        if conversions:
+            table_expr = table_expr.mutate(**conversions)
+        return super()._to_sqlglot(table_expr, limit=limit, params=params)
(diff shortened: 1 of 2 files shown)


cpcloud:
Started playing around with this some more and the outlook isn't great.

I was able to get the test setup working on 18c, but that's the tip of the iceberg.

Here are some things we'd have to do to get this to work:

1. The conversion of boolean -> `CASE`-with-int needs to happen in subqueries, not just the final outer query, otherwise the database errors with this beautiful error message

> oracledb.exceptions.DatabaseError: ORA-00923: FROM keyword not found where expected

getting this right has proved to be elusive for MS SQL which in 2024 **still** does not have boolean type 😮‍💨 

2. The bool conversion also has to handle the null case (which isn't a huge deal)

which has effectively nothing to do with what the problem is

2. `DROP` has to execute some PL/SQL, but only in the case that `force=True`, yay.

My vote is to scrap this PR and only support Oracle starting at version `v${THE_ONE_WITH_PROPER_BOOLS}`.

gforsyth:
> My vote is to scrap this PR and only support Oracle starting at version `v${THE_ONE_WITH_PROPER_BOOLS}`.

That is definitely OK with me.

gitgud5000:
I'm using an Oracle v19 db and this fix solved the error I was getting when using `con.table()`:
> oracledb.exceptions.DatabaseError: ORA-00923: FROM keyword not found where expected

Will this fix be available in the next release?

gforsyth:
> Will this fix be available in the next release or will you only officially support >=23c?

Hey @gitgud5000 -- I think we currently have to limit ourselves to only supporting >=23c.  While some of the fixes in here get you further along using Ibis with Oracle 19 (as you have discovered), there are many remaining pitfalls, especially around query correctness, that need to be addressed due to the lack of proper bool support (among other things).

Issue #9125, opened 2024-05-06
refactor: standardize use of keyword/positional-only arguments


jcrist:
Python 3.8+ supports both keyword-only and positional-only arguments. We might want to survey our existing APIs and try to consistently make use of positional and keyword-only arguments in cases where they make sense.

They main way this would show up as an improvement is making it easier for us to refactor internals without making things a (potentially) breaking change for users. Say we had an existing function:

```python
def my_excellent_function(arg1, arg2=None, arg3=None):
   ...
```

With this, users can call this in a few different formats:

```python
my_excellent_function(1)  # positional with defaults
my_excellent_function(1, 2, 3)  # all as positionals
my_excellent_function(arg1=1, arg2=2, arg3=3)  # all as keywords
my_excellent_function(1, 2, arg3=3)  # mix of positional and keyword
```

Allowing these different calling conventions makes amending this API trickier. We have to worry about adding new keyword arguments only at the end, and never renaming any of the positional arguments (since the user may have called them by name). Moving to a stricter positional-only/keyword-only subset would mean that:

- We can change the names of positional arguments later without worrying that users may have passed them by keyword
- We can reorder keyword arguments without worrying that users may have have passed them positionally

ncclementi:
@gforsyth is the implementation of this one a requirement for 10.0?

gforsyth:
At the very least, we should, I think, make all the breaking changes to positional argument names so that they are consistent across backends.

cpcloud:
@gforsyth It's not 100% clear to me whether the goal here was to make expression APIs more strict (from the original issue description it sounds like it could be), or whether based your statement we should tackle backend APIs for this as well (or perhaps first?).

I went ahead did the expression API version, which will probably be more contentious than backend APIs:

I was thinking that in the interest of getting a release out this week, we should avoid doing the backend APIs, and save those for 11.0.0 (which I don't think we need to wait to long to do).

Issue #8994, opened 2024-04-18
[meta] Ibis backend public API stability


lostmygithubaccount:
**Acceptance criteria**: This quarter, we will have a precise plan and initial work toward a stable public backend interface.

The purpose of this work is to set a stable foundation for new backends to be added, particularly be external contributors/organizations. We should work toward allowing external backends to be fully contributed, maintained, and owned in external repositories.

ncclementi:
After re-triaging/scoping some issues these issues belong to this epic, leaving them here for reference:

## P0
- [ ] https://github.com/ibis-project/ibis/issues/6593  

## P1 
- [ ] https://github.com/ibis-project/ibis/issues/9000 (needs to wait until 9.1) 
- [ ] https://github.com/ibis-project/ibis/issues/8735  (PR open https://github.com/ibis-project/ibis/pull/8863 blocked on #6593)
- [ ] https://github.com/ibis-project/ibis/issues/8382 (PR open https://github.com/ibis-project/ibis/pull/8864 needs review)
- [ ] https://github.com/ibis-project/ibis/issues/8584 
- [ ] https://github.com/ibis-project/ibis/issues/8898 

## P2 
- [ ] https://github.com/ibis-project/ibis/issues/8133 
- [ ] https://github.com/ibis-project/ibis/issues/8727 
- [ ] https://github.com/ibis-project/ibis/issues/8866

gforsyth:
I've opened a PR with a POC for lightly enforcing signature compatibility between backends in #9383

ncclementi:
Closing in favor of https://github.com/ibis-project/ibis/issues/9638

PR #9383, opened 2024-06-13
refactor(all): unify method signatures across backends

gforsyth:
We want to ensure that, for a given backend, that the argument names, plus usage of positional, positional-only, keyword, and keyword-only arguments match, so that there is API consistency when moving between backends.

I've grabbed a few small parts of some of the utilities in Scott Sanderson's
`python-interface` project (https://github.com/ssanderson/python-interface).

While the upstream is no longer maintained, the goal of that project
aligns quite well with some of the issues we face with maintaining
consistent interfaces across backends.

Whether we keep the `pytest` approach for testing the signatures or not, this has proven reasonably effective at finding spots where our argument names aren't matching.  (I think I'm in favor of keeping it, since it flagged up some inconsistencies in `to_parquet_dir` that were added after I opened this PR initially)

I've tried to keep the various refactor commits reasonably atomic so that this isn't an absolute nightmare to review.

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -388,7 +388,7 @@ class _FileIOHandler:
         )
 
     def read_json(
-        self, path: str | Path, table_name: str | None = None, **kwargs: Any
+        self, path: str | Path, /, *, table_name: str | None = None, **kwargs: Any
     ) -> ir.Table:
         """Register a JSON file as a table in the current backend.
 
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -493,7 +493,8 @@ class _FileIOHandler:
         expr
             The ibis expression to execute and persist to parquet.
         directory
-            The data source. A string or Path to the directory where the parquet file will be written.
+            A string or Path to the directory where the parquet file will be
+            written.
         params
             Mapping of scalar parameter expressions to value.
         **kwargs
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -337,7 +337,7 @@ class _FileIOHandler:
         }
 
     def read_parquet(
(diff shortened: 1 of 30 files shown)


ibis/backends/tests/test_signatures.py
```diff
@@ -0,0 +1,26 @@
...
+
+@pytest.mark.parametrize("base_cls, method", params)
+def test_signatures(base_cls, method, backend_cls):
+    if not hasattr(backend_cls, method):
+        pytest.skip(f"Method {method} not present in {backend_cls}, skipping...")
+
+    base_sig = inspect.signature(getattr(base_cls, method))
+    backend_sig = inspect.signature(getattr(backend_cls, method))
+
+    assert compatible(base_sig, backend_sig, check_annotations=False)
```
gforsyth, line 26: I imagine we could break up the base classes into groups where we think type annotations should match vs. those where we expect some variation.

gforsyth, line 26: I have these backwards here, and might make all of these (required) keyword arguments?

Usage is `compatible(implementation, interface_definition, ...)`

gforsyth:
xref #8994

ncclementi:
xref: https://github.com/ibis-project/ibis/issues/9125

ibis/backends/__init__.py
```diff
@@ -477,7 +478,7 @@ def to_parquet(
     def to_parquet_dir(
         self,
         expr: ir.Table,
-        directory: str | Path,
+        path: str | Path,
```
ncclementi, line 481: The use of `directory` was an ask in the `to_parquet_dir` PR, see https://github.com/ibis-project/ibis/pull/9781#discussion_r1709931796 . DO we want to go backwards on that?

gforsyth:
Closing in favor of #10008

Issue #7252, opened 2023-09-29
feat(api): standardize the position of null values in sorted columns and provide an api for specifying null order


cpcloud:
### Discussed in https://github.com/ibis-project/ibis/discussions/7246

<div type='discussions-op-text'>

<sup>Originally posted by **ianmcook** September 28, 2023</sup>
Currently Ibis does not standardize the position of nulls in ordered output. If a user calls `t.order_by(_.x)` or `t.order_by(_.x.desc())` and the column `x` has a mix of null and non-null values, then whether the null values appear at the top (first) or the bottom (last) of the results depends entirely on the default behavior of the Ibis backend you are using. Also, Ibis currently does not expose an option to allow users to control whether nulls sort first or last. 

The behavior of the backend engines is not consistent, but almost all of them do expose an option to control whether nulls sort first or last:

| backend     | default sort position of nulls           | user can specify nulls first or last |
| ----------- | ---------------------------------------- | -------------------------------------|
| bigquery    | first for ascending, last for descending | yes                                  |
| clickhouse  | last                                     | yes                                  |
| dask        | last                                     | yes                                  |
| datafusion  | last for ascending, first for descending | yes                                  |
| druid       | ?                                        | ?                                    |
| duckdb      | last ([since 0.8](https://github.com/duckdb/duckdb/pull/7174)) | yes                                  |
| impala      | last for ascending, first for descending | yes                                  | 
| mssql       | first for ascending, last for descending | no                                   |
| mysql       | first for ascending, last for descending | no                                   |
| oracle      | last for ascending, first for descending | yes                                  |
| pandas      | last                                     | yes                                  |
| polars      | first                                    | yes                                  |
| postgres    | last for ascending, first for descending | yes                                  |
| pyspark     | first for ascending, last for descending | yes                                  |
| snowflake   | last for ascending, first for descending | yes                                  |
| sqlite      | first for ascending, last for descending | yes                                  |
| trino       | last                                     | yes                                  |

For the backends that do not directly expose the option to control nulls first or nulls last in SQL (mssql, mysql), there is a simple workaround to control this: do `ORDER BY x IS NOT NULL, x` to get nulls first and do `ORDER BY x IS NULL, x` to get nulls last.

Note that this does not consider NaNs in float columns, which are another story and should probably be discussed separately.

**Should Ibis standardize null sort order behavior across all the backends?**</div>

gforsyth:
Also from #7246:

[...]

PR #9385, opened 2024-06-14
feat(api): add `nulls_first=False` argument to `order_by`

ncclementi:
- Closes https://github.com/ibis-project/ibis/issues/7252

This is work in progress

## TODO

- [x] Sort out (pun intended) how to deal with backends that don't expose option for ordering nulls. 
        - [x]   mssql
        - [x]   mysql 
EDIT: They get handled by sqlglot and follow the expected behavior

- [x] Add test in `test_api.py` to check that this `t.order_by(ibis.desc("b", nulls_first=True))` is equal than `t.order_by(t["b"].desc(nulls_first=True))`
- [x] fix window tests errors (Thanks @gforsyth for your the "diving into deep into the weeds" session)
 
- [x] Decide how and deprecate https://github.com/ibis-project/ibis/blob/2dac5e48d80e4251d3d682307ff90e00e52455f5/ibis/expr/operations/sortkeys.py#L38-L51

Options are: 
- deprecate passing a tuple completely
- support 2-tuples and 3-tuples (no deprecation)
- don't support 3-tuples, leave 2-tuples for legacy reasons. 
 
NOTE: I can't find any tests using this, not clear to me what do we want to do here. 
EDIT: see https://github.com/ibis-project/ibis/issues/9413

- [x] Figure out why the doctests and plenty others are failing,I think it's because if we don't say `desc()` or `asc()` is taking the `nulls_first` default value set in `SortKeys` which is `True` (I'm not sure yet, need to investigate) **EDIT:** doctest and other tests using order_by that didn't trigger `desc` or `asc` rely on defaul `nulls_first` in `SortKeys`. Before this value wasn't provided and SQLGlot handel this by triggering nulls_first=False. So to be consistent we set it to False in the class default. 

~~I know why they are failing, it's unclear what path we want to take here. Need to discuss~~

https://github.com/ncclementi/ibis/blob/78f26dada233a3da3a5e145fc32d590e058f1a4e/ibis/expr/operations/sortkeys.py#L28-L37

- [x] Discuss/fix: Now we have a lot of snapshot tests failing, because of the fix in the `window_merge_frames` rewrite, where we are now passing `nulls_first` see 
https://github.com/ibis-project/ibis/blob/a17ff2ec4eae0629fbfc00294b51c2089be8faf1/ibis/expr/rewrites.py#L310-L314

- [x] Add tests that orders by more than one column, has nulls, and with different value of nulls_first.


cpcloud:
Fixing up the snapshots now.

ibis/backends/pandas/executor.py
```diff
@@ -593,11 +593,18 @@ def visit(cls, op: ops.Sort, parent, keys):
         # 2. sort the dataframe using those columns
         # 3. drop the sort key columns
         ascending = [key.ascending for key in op.keys]
+        na_pos_dict = {True: "first", False: "last"}
```
ncclementi, line 596: I'm not sure this is the best way to pass `na_position`. Same hack than in the dask executor.

ibis/backends/polars/compiler.py
```diff
@@ -227,10 +227,14 @@ def sort(op, **kw):
 
     by = list(newcols.keys())
     descending = [key.descending for key in op.keys]
+    # NOT SURE IF THIS IS THE RIGHT THING TO DO
+    nulls_first = all(key.nulls_first for key in op.keys)
```
ncclementi, line 231: Not sure if this is the best way to go about it.

ibis/backends/polars/compiler.py
```diff
@@ -227,10 +227,14 @@ def sort(op, **kw):
...
     try:
-        lf = lf.sort(by, descending=descending, nulls_last=True)
+        lf = lf.sort(by, descending=descending, nulls_last=not nulls_first)
     except TypeError:  # pragma: no cover
-        lf = lf.sort(by, reverse=descending, nulls_last=True)  # pragma: no cover
+        lf = lf.sort(
```
ncclementi, line 235: While testing this, I noticed that the `reverse` keyword doesn't work here. I'm not sure what are we trying to do, this is doesn't seem to be tested so I'm not sure what behavior we want.

ibis/backends/dask/executor.py
```diff
@@ -364,10 +364,17 @@ def visit(cls, op: ops.Sort, parent, keys):
         # 2. sort the dataframe using those columns
         # 3. drop the sort key columns
         ascending = [key.ascending for key in op.keys]
+        na_pos_dict = {True: "first", False: "last"}
+        na_position = na_pos_dict[all(key.nulls_first for key in op.keys)]
```
ncclementi, line 368: pandas doesn't support a list that contains different options for null orders. It only accepts `na_position`= "last" or "first".

What do we  want to do here? we could check that if there is more than one to break.

ibis/backends/sql/dialects.py
```diff
@@ -336,6 +338,7 @@ def _create_sql(self, expression: sge.Create) -> str:
     return self.create_sql(expression)
 
 
+Oracle.NULL_ORDERING = "nulls_are_large"
```
ncclementi, line 341: I opened the issue in sqlglot but I'm not sure if I explained it right. Feel free to chime in
https://github.com/tobymao/sqlglot/issues/3684

ibis/backends/sql/dialects.py
```diff
@@ -336,6 +338,7 @@ def _create_sql(self, expression: sge.Create) -> str:
     return self.create_sql(expression)
```
ncclementi, line 340:
```suggestion
Hack around https://github.com/tobymao/sqlglot/issues/3684
```

ibis/backends/sql/dialects.py
```diff
@@ -335,7 +337,8 @@ def _create_sql(self, expression: sge.Create) -> str:
             )
     return self.create_sql(expression)
 
-
+Hack around https://github.com/tobymao/sqlglot/issues/3684
```
ncclementi, line 340:
```suggestion
# Hack around https://github.com/tobymao/sqlglot/issues/3684
```

ibis/expr/operations/sortkeys.py
```diff
@@ -40,15 +41,15 @@ def __coerce__(cls, value, T=None, S=None):
         if isinstance(value, tuple):
```
ncclementi, line 41: see: https://github.com/ibis-project/ibis/issues/9413

ibis/backends/dask/executor.py
```diff
@@ -364,10 +364,23 @@ def visit(cls, op: ops.Sort, parent, keys):
...
         # 3. drop the sort key columns
         ascending = [key.ascending for key in op.keys]
+        nulls_first = [key.nulls_first for key in op.keys]
+
+        if all(nulls_first):
+            na_position = "first"
+        elif not any(nulls_first):
+            na_position = "last"
+        else:
+            raise ValueError("dask does not support different columns ordering")
```
ncclementi, line 374: We currently don't have any tests checking for ordering more than one column with different values of nulls_first. My guess is that we should add this kind of test in test_generic.

ncclementi:
Need to add tests that check for order by more than one column, has nulls, and with different value of `nulls_first`.

ibis/backends/dask/executor.py
cpcloud: Yep, that sounds right!

cpcloud, line 374:
```suggestion
            raise ValueError("dask does not support specifying null ordering per column")
```

ibis/backends/pandas/executor.py
```diff
@@ -593,11 +593,24 @@ def visit(cls, op: ops.Sort, parent, keys):
...
         # 3. drop the sort key columns
         ascending = [key.ascending for key in op.keys]
+        nulls_first = [key.nulls_first for key in op.keys]
+
+        if all(nulls_first):
+            na_position = "first"
+        elif not any(nulls_first):
+            na_position = "last"
+        else:
+            raise ValueError("pandas does not support different columns ordering")
```
cpcloud, line 603:
```suggestion
            raise ValueError("pandas does not support specifying null ordering for individual columns")
```

cpcloud (approved):
Great work!

We can add the tests for multiple `nulls_first` specifications here or in a follow-up, up to you.

ncclementi:
> We can add the tests for multiple nulls_first specifications here or in a follow-up, up to you.

I'll added it here for consistency, working on it now. I'll ping you when done.

ibis/backends/tests/test_generic.py
```diff
@@ -599,6 +600,126 @@ def test_order_by_random(alltypes):
...
+    if con.name == "pandas" and (nf1 != nf2):
+        with pytest.raises(
+            ValueError, match="pandas does not support different columns ordering"
+        ):
+            result = con.execute(expr).reset_index(drop=True)
+    else:
+        result = con.execute(expr).reset_index(drop=True)
+        expected = pd.DataFrame(expected)
+
+        tm.assert_frame_equal(result.replace({np.nan: None}), expected)
```
ncclementi, line 720: If I do this locally `.replace({np.nan: None})` I get this

```python
>           tm.assert_frame_equal(result.replace({np.nan: None}), expected)
E           AssertionError: Attributes of DataFrame.iloc[:, 0] (column name="col1") are different
E           
E           Attribute "dtype" are different
E           [left]:  object
E           [right]: float64
```

only in this test. I don't see it in the other ones.

cpcloud: It's the `expected` that needs the `astype`, not the `result`. I'll push up a fix.

ncclementi: Yes, what's not clear to me why in the other tests we needed it. Locally for me, they all passed without it.

cpcloud:
After losing a few brain cells trying to understand how to set collation and sorting in MS SQL, we gave up and decided to xfail the multi key sort column null ordering tests for backends that for some reason known only to the original developers that strings should compare ignoring case **by default**.

And it should be outrageously difficult to change that behavior.

PR #9398, opened 2024-06-17
refactor(pandas-converter): output Python types for scalar outputs

cpcloud:
Sometimes we output Python types for scalar queries and other times we output NumPy types. This PR changes the output to always be Python types in the case of scalar queries. Changing the output type of `.execute` is a breaking change.


cpcloud:
Alternatively, we leave this as is.

cpcloud:
Meh, doesn't seem worth it right now. I'll adjust the numpy 2.0 upgrade to use a different repr in the doctests.

PR #9400, opened 2024-06-17
chore(deps): bump lower bound of pytest to 8.2.0

gforsyth:
We make use of the `exc_type` kwarg in `pytest.importorskip` which was
added in 8.2.0.


cpcloud (approved):
Thanks!

PR #9404, opened 2024-06-18
fix(sec): remove most instances of possible sql injection

cpcloud:
This PR adds ruff's `flake8-bandit` checks and fixes the lints or explicitly
ignores them (where it's low risk or unavoidable).

Most of the work here was in building SQL using sqlglot instead of constructing
queries using format strings.

That's not 100% foolproof for cases like `.sql()` methods, but we're at least
covering the bases in cases where we have more control over input.

### DIFF
--- a/ibis/backends/sql/__init__.py
+++ b/ibis/backends/sql/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import abc
+from functools import partial
 from typing import TYPE_CHECKING, Any, ClassVar
 
 import sqlglot as sg
@@ -427,14 +428,14 @@ class SQLBackend(BaseBackend, _DatabaseSchemaHandler):
 
         self._run_pre_execute_hooks(obj)
 
-        query = self._build_insert_query(
+        query = self._build_insert_from_table(
             target=table_name, source=obj, db=db, catalog=catalog
         )
 
         with self._safe_raw_sql(query):
             pass
 
-    def _build_insert_query(
+    def _build_insert_from_table(
         self, *, target: str, source, db: str | None = None, catalog: str | None = None
     ):
         compiler = self.compiler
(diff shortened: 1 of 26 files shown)


ibis/backends/sql/__init__.py
```diff
@@ -463,6 +464,30 @@ def _build_insert_query(
         )
         return query
 
+    def _build_insert_template(
+        self, *, table, schema: sch.Schema, columns: bool = False, fmt: str = "?"
+    ) -> str:
```
jcrist, line 469: Since all backends use this, would you mind adding a small docstring, mainly clarifying what the args are?

Also, I _think_ `table` is always passed as `sg.table(name, quoted=quoted)` at every call site - it _might_ be cleaner to only pass the `name` instead to avoid this duplication (don't we set `quoted=True` everywhere anyway?). No strong thoughts.

ibis/util.py
```diff
@@ -678,3 +678,28 @@ def __ne__(self, other):
...
+    ----------
+    n
+        The total number of elements.
+    chunk_size
+        The size of each chunk.
+
+    Returns
+    -------
+    int
+        THE start and end indices of each chunk.
```
jcrist, line 696:
```suggestion
    tuple[int, int]
        The start and end indices of each chunk.
```

(unless you really wanted that emphasis)

jcrist (approved):
Just a cursory review, honestly mostly relying on tests here to have caught any regressions.

ibis/util.py
cpcloud: 😂

ibis/backends/sql/__init__.py
cpcloud: Happy to add a docstring.

sqlite uses a table with a specific catalog, so i'll just leave that bit as is.

cpcloud: Eh, I can add a `catalog` argument that'll default to `None` in most cases.

PR #9410, opened 2024-06-20
fix(trino): allow passing the `auth` keyword

### LINKED ISSUE #9409: Problem of authentication with JWTAuthentication and Trino
While trying to execute this code:
```Python
import google.auth
import ibis
from google.auth.transport.requests import Request
from trino.auth import JWTAuthentication
from trino.dbapi import connect

credentials, project_id = google.auth.default()
credentials.refresh(Request())

con = ibis.trino.connect(
    user="me",
    auth=JWTAuthentication(credentials.token),
    http_scheme="https",
    host="https://ur.l",
    port="8000",
    database="my_catalog",
    schema="my_schema",
)
con.list_databases()
```

I am getting the following error:
TypeError: trino.dbapi.connect() got multiple values for keyword argument 'auth'

The same code is working without any issue in order to connect directly to trinodb.

The issue is coming from the following line, duplicating the auth parameters that is coming both from the **kwargs and from the auth=password,:
https://github.com/ibis-project/ibis/blob/bdc1b3fbb8d44f9246c957b58372d3df84247468/ibis/backends/trino/__init__.py#L303

In order to fix this problem, the following code is working:
```Python
con = ibis.trino.connect(
    user="me",
    password=JWTAuthentication(credentials.token),
    http_scheme="https",
    host="https://ur.l",
    port="8000",
    database="my_catalog",
    schema="my_schema",
)
```

I would then suggest to:
- either add logic to not duplicate the auth parameters if already in the **kwargs
- I don't know if the issue is the same with https://ibis-project.org/backends/trino#authenticating-with-sso, but if it is, updating the documentation to an actually working code, or adding a new parts dedicated to JWTAuthentication.

Thanks! :smile:

cpcloud:
In Ibis 10, I will change the connect API here to just forward kwargs, but for now this change is backward compatible and can be shipped in a non-major release.

Closes #9409.

### DIFF
--- a/ibis/backends/trino/__init__.py
+++ b/ibis/backends/trino/__init__.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import contextlib
+import warnings
 from functools import cached_property
 from operator import itemgetter
 from typing import TYPE_CHECKING, Any
@@ -43,11 +44,12 @@ class Backend(SQLBackend, CanListCatalog, CanCreateDatabase, CanCreateSchema):
         catalog, db = url.path.strip("/").split("/")
         self.do_connect(
             user=url.username or None,
-            password=url.password or None,
+            auth=url.password or None,
             host=url.hostname or None,
             port=url.port or None,
             database=catalog,
             schema=db,
+            **kwargs,
         )
         return self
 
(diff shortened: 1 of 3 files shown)


cpcloud:
Actually, I'm going to deprecated `password`.

PR #9411, opened 2024-06-20
test(common): replace error message snapshot with match

cpcloud:
Fixes an issue with one of our tests that assumes a specific error message structure. This structure changed in Python 3.12.4, causing the changed test to fail on `main`.


ncclementi (approved):
LGTM!, thanks for fixing this one.

Issue #9216, opened 2024-05-21
bug: `create_table(temp=True)` timing out due to slow table existence check


asiunov:
### What happened?

`create_table` for BigQuery backend may become inappropriately inefficient.

`bigquery.Backend.create_table(name, df, temp=True)` calls `_register_in_memory_tables`, which calls `_register_in_memory_table`, which checks if the table exists using:
```python
if raw_name not in self.list_tables(database=(project, dataset)):
```

The problem for BigQuery is caused by the fact that BQ uses hidden datasets to store temp data (see `_make_session` function). These datasets are reused between sessions and may collect more and more tables within table expiration window. I've got 10s of thousands of temp tables. The listing for such large dataset takes minutes (hours, etc) and even causes request throttling from google side.

I have to use this monkey-patching to bypass the issue:
```python
import google.api_core.exceptions as google_exceptions
import google.cloud.bigquery as bq_api
import ibis.expr.operations as ibis_ops
import sqlglot as sg
from ibis.backends.bigquery import Backend as BigQueryBackend
from ibis.backends.bigquery.datatypes import BigQuerySchema


def table_exists(backend: BigQueryBackend, table_name: str, database: tuple[str, str] | str | None = None) -> bool:
  try:
    _ = backend.table(table_name, database=database)
    return True
  except google_exceptions.NotFound as e:
    if "Not found: Table" in e.message:
      return False
    else:
      raise


def _register_in_memory_table(self: BigQueryBackend, op: ibis_ops.InMemoryTable) -> None:
  raw_name = op.name

  project = self._session_dataset.project
  dataset = self._session_dataset.dataset_id

  if not table_exists(self, raw_name, database=(project, dataset)):
    table_id = sg.table(raw_name, db=dataset, catalog=project, quoted=False).sql(dialect=self.name)

    bq_schema = BigQuerySchema.from_ibis(op.schema)
    load_job = self.client.load_table_from_dataframe(
      op.data.to_frame(),
      table_id,
      job_config=bq_api.LoadJobConfig(
        # fail if the table already exists and contains data
        write_disposition=bq_api.WriteDisposition.WRITE_EMPTY,
        schema=bq_schema,
      ),
    )
    load_job.result()


BigQueryBackend._register_in_memory_table = _register_in_memory_table
```

### What version of ibis are you using?

9.0.0.

### What backend(s) are you using, if any?

BigQuery

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

[...]

Issue #9215, opened 2024-05-21
feat: a method for table existence check


asiunov:
### Is your feature request related to a problem?

If I understand correctly, the best way to check table existence is by running `table_name in list_tables(like=table_name)`, but this may be inefficient if the database contains a lot of tables, and may be incorrect if `table_name` contains regex chars.

By the way, probably another feature request is to push the filter to server: `list_tables` does filtering on client side (at least for BQ, Snowflake, DuckDB), which is inefficient if there are a lot of tables.

### Describe the solution you'd like

A new method `Backend.table_exists`.

Also, maybe change `Backend.list_tables(like)` to server-side filtering.

### What version of ibis are you running?

9.0.0

### What backend(s) are you using, if any?

DuckDB, Snowflake, BigQuery

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jcrist:
> the best way to check table existence is by running table_name in list_tables(like=table_name)

Depending on what you intend to do if it does exist, it _might_ be more efficient to do `con.table(table_name)` and branch if that operation fails. In the case it exists though, creating the table will do some work to fetch the schema and construct a `ibis.Table` object.

```python
try:
    t = con.table(table_name)
except Exception:
    # table doesn't exist
    ...
else:
    # table does exist
    ...
```

That said, a builtin method on `con` just checking for existence could definitely be made more efficient.

Can you comment a little more on your use case here? In what cases are you checking for the existence of tables?

asiunov:
Thank you for your reply @jcrist

The problem with this approach is that `Exception` is too broad; what if some sort of connection failure happens? Table existence is not checked in this case. And there is no any specific exception for missing table error: all implementations of `backend.table(...)` raise their own exception types. So, as a workaround, I had to figure out every exception and catch specific ones depending on backend.

About the use case: I develop a service that manages tables, I do not work with hardcoded table names, they are supplied by users. User provides a table name and the service needs to check whether the table exists.

gforsyth:
> So, as a workaround, I had to figure out every exception and catch specific ones depending on backend.

Welcome to Ibis! :grinning:

jcrist:
> I develop a service that manages tables, I do not work with hardcoded table names, they are supplied by users. User provides a table name and the service needs to check whether the table exists.

What actions do you do if the table exists already? What do you do if it doesn't exist?

[...]

PR #9419, opened 2024-06-21
perf(duckdb): speed up memtable registration

cpcloud:
Speeds up memtable registration by a factor of some of multiple of N, where N is the number of tables in the database. In other words, the time to register a memtable is no longer dependent on the number of tables in the database.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -1556,8 +1556,12 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
             self._register_in_memory_table(memtable)
 
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
-        # only register if we haven't already done so
-        if (name := op.name) not in self.list_tables():
+        name = op.name
+        try:
+            # this handles tables _and_ views
+            self.con.table(name)
+        except (duckdb.CatalogException, duckdb.InvalidInputException):
+            # only register if we haven't already done so
             self.con.register(name, op.data.to_pyarrow(op.schema))
 
     def _register_udfs(self, expr: ir.Expr) -> None:
--- a/ibis/tests/benchmarks/test_benchmarks.py
+++ b/ibis/tests/benchmarks/test_benchmarks.py
@@ -4,6 +4,7 @@ import copy
 import functools
 import inspect
 import itertools
+import math
 import os
 import string
 from operator import attrgetter, itemgetter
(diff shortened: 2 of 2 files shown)


ibis/backends/duckdb/__init__.py
```diff
@@ -1556,8 +1556,12 @@ def _register_in_memory_tables(self, expr: ir.Expr) -> None:
             self._register_in_memory_table(memtable)
 
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
-        # only register if we haven't already done so
-        if (name := op.name) not in self.list_tables():
+        name = op.name
+        try:
+            # this handles tables _and_ views
+            self.con.table(name)
+        except (duckdb.CatalogException, duckdb.InvalidInputException):
```
cpcloud, line 1563: The `InvalidInputException` is here for cases where `name` matches a variable in scope that DuckDB will try to magically turn into a table. I don't like it, but that's life 😮‍💨

ibis/backends/duckdb/__init__.py
```diff
@@ -1556,8 +1556,12 @@ def _register_in_memory_tables(self, expr: ir.Expr) -> None:
             self._register_in_memory_table(memtable)
 
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
-        # only register if we haven't already done so
-        if (name := op.name) not in self.list_tables():
+        name = op.name
+        try:
+            # this handles tables _and_ views
+            self.con.table(name)
```
jcrist, line 1562: This does some extra unnecessary work that it might be nice to avoid (fine to do in a follow-up PR).
- Queries the schema of the table
- Coerces the schema into an ibis schema
- Creates a table object

In the future it might be nice to replace this with a `con.exists`/`con.table_exists`/`con.has_table` API (as proposed in #9215).

Alternatively/additionally we might add a connection-local cache for handling memtables on the `con` object itself. Since memtables have unique ids for names, a simple `dict`/`set` on the `con` should be sufficient for avoiding re-registering tables, and shouldn't run the risk of getting out of sync.

jcrist (approved):
Fine by me to merge for now. We should really apply this kind of change universally across all our backends. Someone raised a similar issue in #9216 also caused by this, but on the bigquery backend instead.

ibis/backends/duckdb/__init__.py
cpcloud: Note this isn't calling `self.table`, it's `self.con.table`

jcrist: Ah, missed that. Thanks.

PR #9423, opened 2024-06-21
feat(api): add `TableUnnest` operation to support cross-join unnest semantics as well as `offset`

### LINKED ISSUE #7781: feat(bigquery): `ArrayValue.as_table(offset_name: str | None)` as a table-valued function

It's difficult for me to do some things that I'm used to doing in BigQuery, such as `UNNEST([...])` on a bunch of literal values or a generated array, similar to how `memtable` worked in Ibis 6.x.

Also, the current implementation of `unnest()` doesn't work on deeply nested arrays of structs of arrays in the current implementation. An alternative API closer to how BQ semantics work for unnest would be really useful, IMO.

As an alternative to the current `unnest()` semantics, which transforms an `ArrayColumn` into `XColumn`, I'd like to see `ArrayValue.as_table()` which works similarly to UNNEST in BigQuery, where it's primarily used in a `FROM` clause, often in a [correlated join](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#correlated_join) but not always.

I suspect the "treat this unnested array as a table with/without offsets" could be doable in other engines besides BigQuery, especially if correlated joins are ignored for now.

7.1.0

BigQuery

### Code of Conduct

Edit: Updated to `as_table` to reflect similar function in `StructValue`.

cpcloud:
Would love to get feedback from folks on the API especially @tswast. The new test is likely to fail on many backends. Looking for a bit of feedback before beefing up the test suite.

Closes #7781.

### DIFF
--- a/ibis/backends/bigquery/compiler.py
+++ b/ibis/backends/bigquery/compiler.py
@@ -690,3 +690,47 @@ class BigQueryCompiler(SQLGlotCompiler):
         table = sg.to_identifier(parent.alias_or_name, quoted=quoted)
         column = sge.Column(this=star, table=table)
         return sg.select(column).from_(parent)
+
+    def visit_TableUnnest(
+        self, op, *, parent, column, offset: str | None, keep_empty: bool
+    ):
+        quoted = self.quoted
+
+        column_alias = sg.to_identifier(
+            util.gen_name("table_unnest_column"), quoted=quoted
+        )
+
+        selcols = []
+
+        table = sg.to_identifier(parent.alias_or_name, quoted=quoted)
+
+        opname = op.column.name
+        overlaps_with_parent = opname in op.parent.schema
+        computed_column = column_alias.as_(opname, quoted=quoted)
+
+        # replace the existing column if the unnested column hasn't been
+        # renamed
+        #
+        # e.g., table.unnest("x")
+        if overlaps_with_parent:
+            selcols.append(
+                sge.Column(this=sge.Star(replace=[computed_column]), table=table)
+            )
+        else:
+            selcols.append(sge.Column(this=STAR, table=table))
+            selcols.append(computed_column)
+
+        if offset is not None:
+            offset = sg.to_identifier(offset, quoted=quoted)
+            selcols.append(offset)
+
+        unnest = sge.Unnest(
+            expressions=[column],
+            alias=sge.TableAlias(columns=[column_alias]),
+            offset=offset,
+        )
+        return (
+            sg.select(*selcols)
+            .from_(parent)
+            .join(unnest, join_type="CROSS" if not keep_empty else "LEFT")
+        )
--- a/ibis/backends/tests/test_array.py
+++ b/ibis/backends/tests/test_array.py
@@ -1447,6 +1447,9 @@ def test_table_unnest_with_keep_empty(con):
     ["datafusion", "pandas", "polars", "dask", "flink", "pyspark"],
     raises=com.OperationNotDefinedError,
 )
+@pytest.mark.notyet(
+    ["risingwave"], raises=PsycoPg2InternalError, reason="not supported in risingwave"
+)
 def test_table_unnest_column_expr(backend):
     t = backend.array_types
     expr = t.unnest(t.y.map(lambda v: v.cast("str") + "'s").name("plural"))
--- a/ibis/backends/tests/test_array.py
+++ b/ibis/backends/tests/test_array.py
@@ -1391,7 +1391,7 @@ def test_zip_unnest_lift(con):
 
 
 @pytest.mark.notimpl(
-    ["datafusion", "pandas", "polars", "dask", "flink", "pyspark"],
+    ["datafusion", "pandas", "polars", "dask", "flink"],
     raises=com.OperationNotDefinedError,
 )
 @pytest.mark.parametrize(
@@ -1407,7 +1407,7 @@ def test_table_unnest(backend, colspec):
 
 
 @pytest.mark.notimpl(
-    ["datafusion", "pandas", "polars", "dask", "flink", "pyspark"],
+    ["datafusion", "pandas", "polars", "dask", "flink"],
     raises=com.OperationNotDefinedError,
 )
 def test_table_unnest_with_offset(backend):
@@ -1423,17 +1423,17 @@ def test_table_unnest_with_offset(backend):
     idx = iter(df.idx.values)
     expected = (
         df.assign(**{col: df[col].map(lambda v: v[next(idx)])})
-        .sort_values(["idx", "y"])
+        .sort_values(["idx", col])
         .reset_index(drop=True)[["idx", col]]
     )
 
-    expr = t.unnest(col, offset="idx")[["idx", col]].order_by("idx", "y")
+    expr = t.unnest(col, offset="idx")[["idx", col]].order_by("idx", col)
     result = expr.execute()
     tm.assert_frame_equal(result, expected)
 
 
 @pytest.mark.notimpl(
-    ["datafusion", "pandas", "polars", "dask", "flink", "pyspark"],
+    ["datafusion", "pandas", "polars", "dask", "flink"],
     raises=com.OperationNotDefinedError,
 )
 def test_table_unnest_with_keep_empty(con):
@@ -1444,7 +1444,7 @@ def test_table_unnest_with_keep_empty(con):
 
 
 @pytest.mark.notimpl(
-    ["datafusion", "pandas", "polars", "dask", "flink", "pyspark"],
+    ["datafusion", "pandas", "polars", "dask", "flink"],
     raises=com.OperationNotDefinedError,
 )
 @pytest.mark.notyet(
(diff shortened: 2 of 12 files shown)


ibis/backends/sql/compiler.py
```diff
@@ -1512,6 +1512,33 @@ def visit_Unsupported(self, op, **_):
...
+
+        if offset is not None:
+            offset = sg.to_identifier(offset, quoted=quoted)
+            columns.append(offset)
+
+        expr = sg.select(*columns)
+        alias = sge.TableAlias(columns=[column_alias])
+        return expr.from_(parent).join(
+            sge.Unnest(expressions=[column], alias=alias, offset=offset),
+            how="CROSS",
```
tswast, line 1539: Any chance we could add a `how` parameter? It'd be useful to have `LEFT` join support so BigQuery DataFrames can better emulate pandas's explode method.

ibis/expr/types/relations.py
```diff
@@ -4677,6 +4677,13 @@ def value_counts(self) -> ir.Table:
             lambda t: t.count().name("_".join(columns) + "_count")
         )
 
+    def unnest(self, column, offset: str | None = None) -> Table:
```
tswast, line 4680: +1 This seems pretty natural to me. :-)

ibis/backends/sql/compiler.py
tswast: Or a boolean parameter `preserve_empty_arrays` or something? If true, then it does a left join.

ibis/expr/types/relations.py
tswast: Would we want to call it explode to mimic pandas? That said, I suppose this makes sense with regards to internal consistency with the other "unnest".

cpcloud: One thing that I think might be a bit confusing is that sqlglot calls unnest-in-select-position `explode` and unnest-in-join-position `unnest` so if called the unnest here explode that would be the reverse convention of sqlglot.

Personally I think keeping the name the same, seems like it will be better for conceptual clarity, since the operations are doing the same thing at their core.

ibis/backends/sql/compiler.py
cpcloud: added a `keep_empty` parameter, defaulting to `False`

cpcloud:
Ok, I've started going down the rabbit hole of supporting this on as many of our backends that support arrays as possible.

ibis/backends/bigquery/compiler.py
```diff
@@ -690,3 +690,44 @@ def visit_DropColumns(self, op, *, parent, columns_to_drop):
         table = sg.to_identifier(parent.alias_or_name, quoted=quoted)
         column = sge.Column(this=star, table=table)
         return sg.select(column).from_(parent)
+
+    def visit_TableUnnest(
+        self, op, *, parent, column, offset: str | None, keep_empty: bool
+    ):
+        quoted = self.quoted
```
tswast, line 697: For my own education, what does `quoted` mean in this context?

ibis/backends/bigquery/compiler.py
```diff
@@ -690,3 +690,44 @@ def visit_DropColumns(self, op, *, parent, columns_to_drop):
...
+        column_alias = sg.to_identifier(
+            util.gen_name("table_unnest_column"), quoted=quoted
+        )
+
+        selcols = []
+
+        table = sg.to_identifier(parent.alias_or_name, quoted=quoted)
+
+        opname = op.column.name
+        if opname in op.parent.schema:
```
tswast, line 708: I don't fully understand the purpose of these different branches here or in the other implementations. Could you add a comment?

Is it related to avoiding name collisions between the original column being unnested and the unnested version of it?

ibis/backends/tests/test_array.py
```diff
@@ -1384,7 +1384,75 @@ def test_zip_unnest_lift(con):
...
+@pytest.mark.notimpl(
+    ["datafusion", "pandas", "polars", "dask", "flink"],
+    raises=com.OperationNotDefinedError,
+)
+@pytest.mark.parametrize(
+    "colspec",
+    ["y", lambda t: t.y, ibis._.y],
+    ids=["string", "lambda", "deferred"],
+)
+def test_table_unnest(backend, colspec):
```
tswast, line 1402: I'm curious if `table.unnest` works with doubly nested (`array<struct<array<>>`) fields? If so, could you add a test for that case?

ibis/backends/bigquery/compiler.py
cpcloud: It's a flag to indicate whether to quote everything.

It's a bit weird, because it's currently set to `True` for every backend, but that was from a time when I thought we might be able to not quote everything.

It turned out to be much easier to just quote everything all the time.

ibis/backends/bigquery/compiler.py
cpcloud: > Is it related to avoiding name collisions between the original column being unnested and the unnested version of it?

That's exactly it. I can add some commentary.

ibis/backends/tests/test_array.py
cpcloud: Added a test case in [`971b005` (#9423)](https://github.com/ibis-project/ibis/pull/9423/commits/971b0052c745a1610d5e8bb29240298024031675). Let me know whether it's sufficient.

ibis/backends/bigquery/compiler.py
cpcloud: Done

gforsyth (approved):
6 SQL backends, 6 slightly different (or just plain gnarly) SQL implementations.

Yay standards!

Anyway, this looks good to me -- really nice docstring for `Table.unnest`

cpcloud:
Clouds are good:

```
…/ibis on  bigquery-table-unnest is 📦 v9.1.0 via 🐍 v3.10.14 via ❄️   impure (ibis-3.10.14-env) took 30s
bringing up nodes...
.....................x.x.....................x...........x...x.x........xx.....x....x.................x.x..xx............x..s....x.x................................x................. [  4%]
.........x.......x...........................x......................x..x.xxx....x..xxxx..xx.......x...x...x......x....................s...................s..........................s [  9%]
.....................x.....x....x...x.....xx.........................x.....................x..x.......x.........x....x.................x..x..............x.....................x...... [ 14%]
........x.......................x....x......xx..x.x.....x.....x.x........x......x..x..........x....xx....xx.x...x....x..........x.x........x...................x......x.....xx.x.....x [ 19%]
.........x...xx.xxxxxx.x.x.......x.x..xxx.x....x.x........x..............x......x....................x.x..xx....xx..x.......................................x...........xx............ [ 24%]
....x............................x...x.....x......................xx.......x......................x........xx.x.....x..............x..........x........x...x........x............x.... [ 29%]
..x.............x..........x.............sssssssssssssssssssss.............x............x...xs..............x........x..x.............x....................x.......................... [ 34%]
......................................x........................................x.......................x.........................x.....x......s.x..............x.xxx..x...xxx...xxx.xx [ 39%]
xxx..xx.xx.x.xx..xxxxx.x.xx..xxx........x..xx...x..x...xxx.xxxxxxx..xxx....x....x..x.xxx....x...xxx...xx.x........x...xxx..xx.x...x.xx.x...x..xx.xx....xx..x.xx.xxx....xxx......x..... [ 44%]
.xx........................x.......x...........................x................x............................x.........xx.....x...........x..x..................x.x.........xxxx.xxx.x [ 49%]
xx...xx...x......x.x..x...........s.....s...............................................s.........x.x..x........x.............xx....x........x.......x.............................x.. [ 54%]
........x...........x.........................x....................................................x.........................................x................xx..x..x........x....x.. [ 59%]
.x.....x...x.........x...x.......x.xx.x.............x..x...x.x.........x.x.........x.x....x......xxxxxxxx..x.x.x.x...x.xx.x..x.....xxx..x...x.................xx...x......x.......x.xx [ 64%]
..x..x.............x..........x.......x..x.......x....xx............x.x..........x.....x..x.x..x..xx.......x.........x..............x..x.........................................x.x.. [ 69%]
....x.....xx.........x...x.....x..x...x..xx..........................x............x..x....xx....xx...x......................x..x..x......................x...........x......s.....x... [ 73%]
......................xxx......x........xx.............x........x...xx..x....xxx....x.x..x..............x...x...x.xx..........x......x.x.....x.....x...x...x....s..x..x............... [ 78%]
.........s....x...x...x.x..s.s.x.........xxxxxxx..xx..x..xxxxxxxxxxx.xxxxxxxxxxxxxxx.xxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.sxxxxsx.xx.xx.... [ 83%]
...........................x.......................x.........................................................................s.....................................x.x................ [ 88%]
.................................s.................................................................................................................................................... [ 93%]
..........................................................................................................x........................................................................... [ 98%]
..................................................                                                                                                                                     [100%]
3099 passed, 39 skipped, 552 xfailed in 835.57s (0:13:55)
```

PR #9434, opened 2024-06-24
fix(mssql): fix temporary table creation and implement `cache`

### LINKED ISSUE #9431: bug: mssql backend create_table with temp=True get incorrect syntax exception
### What happened?

for simple temp table with mssql backend like this:
```
c = ibis.mssql.connect(...)
df = pd.DataFrame.from_dict({'name': ['john', 'joe']})
c.create_table('team', df, temp=True)
```

looks like for mssql dialect, the correct way to create temp table is `create table #table_name`, but ibis is still generate it as `create temporary table table_name`.

9.0.0

mssl

### Relevant log output

```sh
with self._safe_raw_sql(create_stmt + " WITH (HEAP)") as cur:
  File "/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "=/lib/python3.11/site-packages/ibis/backends/mssql/__init__.py", line 268, in _safe_raw_sql
    cur.execute(query, *args, **kwargs)
pyodbc.ProgrammingError: ('42000', "[42000] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Parse error at line: 1, column: 86: Incorrect syntax near '+'. (103010) (SQLExecDirectW)")
```

### Code of Conduct

gforsyth:
MSSQL has two kinds of temporary tables.

1. Local temporary tables, created in `tempdb.dbo`, prefixed with a
single `#` and suffixed with ~~112 underscores and a number.
2. Global temporary tables, created in `tempdb.dbo`, and prefixed with a
double `#`.

The suffixing on local temporary tables is to avoid name collisions
between separate user sessions trying to create the same temporary table
name.
Otherwise, the two types of temporary tables are functionally
equivalent. Users need specific access to look at temporary tables that
they didn't create, and all temp tables are cleaned up after the session
closes and all stored procedures referencing those tables have run.

Because of the slightly wacky semantics of local temporary tables, this
PR makes all temp tables created by the MSSQL backend "global". This
makes the names much more predictable.

A user creating a temporary table with the name "hithere", will find
that there is no table called "hithere" in the output of
`con.list_tables()`. Instead, they will find a table called "##hithere"
if they run `con.list_tables(database=("tempdb", "dbo"))`.

The returned table reference from the `create_table` call works the same
as any other table reference, and you can create persistent tables from
temporary tables.  The only major hiccough is the perpending of the `##`
on the table name, but that is the behavior of MSSQL.

I've also added in support for passing in explicit `catalog` and
`database` to `create_table`, which was missing (I believe this was an
oversight, since MSSQL definitely supports this).

Resolves #9431

### DIFF
--- a/ibis/backends/mssql/__init__.py
+++ b/ibis/backends/mssql/__init__.py
@@ -133,6 +133,12 @@ class Backend(SQLBackend, CanCreateCatalog, CanCreateDatabase, CanCreateSchema, 
     def get_schema(
         self, name: str, *, catalog: str | None = None, database: str | None = None
     ) -> sch.Schema:
+        # TODO: this is brittle and should be improved. We want to be able to
+        # identify if a given table is a temp table and update the search
+        # location accordingly.
+        if name.startswith("ibis_cache_"):
+            catalog, database = ("tempdb", "dbo")
+            name = "##" + name
         conditions = [sg.column("table_name").eq(sge.convert(name))]
 
         if database is not None:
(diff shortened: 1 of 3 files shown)


ibis/backends/mssql/__init__.py
```diff
@@ -481,20 +484,56 @@ def create_table(
...
+            The table that was created.
+
+        """
         if obj is None and schema is None:
             raise ValueError("Either `obj` or `schema` must be specified")
 
-        if database is not None and database != self.current_database:
-            raise com.UnsupportedOperationError(
-                "Creating tables in other databases is not supported by Postgres"
+        if temp and overwrite:
```
cpcloud, line 524: I guess this means `DROP TABLE ##anything` just isn't supported in MS SQL?

ibis/backends/mssql/__init__.py
```diff
@@ -133,6 +133,9 @@ def do_connect(
     def get_schema(
         self, name: str, *, catalog: str | None = None, database: str | None = None
     ) -> sch.Schema:
+        if name.startswith("ibis_cache_"):
```
cpcloud, line 136: Probably should have a `TODO` here to have a slightly more robust solution to the problem of cached table lookups. I can definitely see non-internal uses of a table with this prefix.

cpcloud (approved):

PR #9451, opened 2024-06-27
feat(api): allow grouping by scalar values

cpcloud:
Allow grouping by scalar values. Needed for TPC-DS. In particular, query 27 uses this functionality.

Also a somewhat common pattern for building unions of aggregations from disparate tables.

PR #9452, opened 2024-06-27
fix(druid): handle typed nulls where possible

cpcloud:
Fixes a bug where we were not casting nulls to a specific type where we should.

Discovered during #9451.


gforsyth (approved):
This feels like something that could use a proper test to avoid regression, but looks good overall!

cpcloud:
Gonna rebase, which should pick up another test. It's not specific for this case but the code path does hit it.

PR #9453, opened 2024-06-27
refactor(tests): add tpc ds setup and rearrange tpc setup

cpcloud:
This PR adds support for testing TPC-DS queries for correctness against the
handwritten DuckDB SQL in basically the same way that we do now for TPC-H.

The methodology is shared across H and DS query sets.


ibis/backends/tests/tpc/conftest.py
```diff
@@ -0,0 +1,122 @@
...
+
+            result_expr = test(*args, **kwargs)
+
+            ibis_sql = ibis.to_sql(result_expr, dialect=backend_name)
+
+            assert result_expr._find_backend(use_default=False) is backend.connection
+            result = backend.connection.execute(result_expr)
+            assert not result.empty
+
+            expected = expected_expr.execute()
```
gforsyth, line 90:
```suggestion
            result = backend.connection.to_pandas(result_expr)
            assert not result.empty

            expected = expected_expr.to_pandas()
```

micronit

gforsyth (approved):
:shipit:

PR #9458, opened 2024-06-27
feat(api): improve ibis.array()

NickCrews:
Picking out the array stuff from https://github.com/ibis-project/ibis/pull/8666

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -87,25 +87,27 @@ def literal(op, **_):
     value = op.value
     dtype = op.dtype
 
-    if dtype.is_array():
-        value = pl.Series("", value)
-        typ = PolarsType.from_ibis(dtype)
-        val = pl.lit(value, dtype=typ)
-        return val.implode()
+    # There are some interval types that _make_duration() can handle,
+    # but PolarsType.from_ibis can't, so we need to handle them here.
+    if dtype.is_interval():
+        return _make_duration(value, dtype)
+
+    typ = PolarsType.from_ibis(dtype)
+    if value is None:
+        return pl.lit(None, dtype=typ)
+    elif dtype.is_array():
+        return pl.lit(pl.Series("", value).implode(), dtype=typ)
     elif dtype.is_struct():
         values = [
             pl.lit(v, dtype=PolarsType.from_ibis(dtype[k])).alias(k)
             for k, v in value.items()
         ]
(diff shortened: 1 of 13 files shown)


ibis/backends/polars/compiler.py
```diff
@@ -87,25 +87,25 @@ def literal(op, **_):
     value = op.value
     dtype = op.dtype
 
-    if dtype.is_array():
-        value = pl.Series("", value)
-        typ = PolarsType.from_ibis(dtype)
-        val = pl.lit(value, dtype=typ)
-        return val.implode()
+    if dtype.is_interval():
+        return _make_duration(value, dtype)
```
NickCrews, line 91: have to move this up here because there are some interval dtypes that _make_duration() can handle, but `PolarsType.from_ibis(dtype)` errors on.

NickCrews:
I'm thinking about this more and how I don't like the current conditional casting logic, and how only sometimes is the dtype in ops.Array needed. Currently, when constructing an op.Array, all the arguments are taken as they are, and the casting only happens during compile time. So you can have an `op.Array(exprs=(1,2), dtype="array<string>")`. This is leading to some conditional logic at compile time, deciding when a cast is needed or not. What could be simpler is if we ensure that all the exprs inside an op.Array are already the correct type during/before construction. so we would never find a `op.Array(exprs=(1,2), dtype="array<string>")`, instead we would have found the value_type of string, and pre-cast all the arguments, so we would have `op.Array(exprs=(cast(1, string),cast(2, string)), dtype="array<string>")`. Then in the compile step, we would need to look at the dtype iff the length was 0. This then logically leads to @cpcloud 's suggestion of having two different ops, one for 0-length arrays with a dtype, and another for arrays with vals. Also, by pushing this member-casting to construction time, we would catch more misuses, such as `ibis.array([1,2], type="array<array<string>>"`.

So the logic would happen earlier, something like

```python
def array(vals, *, type):
    vals = promote_tuple(vals)
    type = ibis.dtype(type)
    if not type.is_array():
         raise
    if not vals:
         if not type:
               raise
         return ops.EmptyArray(type)
    val_type = type.value_type
    casted = [value(e).cast(val_type) for e in exprs]
    return ops.Array(casted)

def value(v):
     if isinstance(v, (ibis.Expr, ibis.Deferred)):
         return v
     return ibis.literal(v)
````

NickCrews:
closing in favor of #9473

PR #9459, opened 2024-06-27
feat(api): improve map()

NickCrews:
works towards https://github.com/ibis-project/ibis/issues/8289

depends on the new array API of #9458 

One this is adding support for passing in None.
These use the new `ibis.null(<type>)` API to return `op.Literal(None, <type>)`s

Also, now map() is idempotent: you can
pass in existing Expressions into map().
The type argument for all of these now always has an effect, not just when passing in python literals. So basically it acts like a cast if the argument is already an expression.

You can test this locally with eg
`pytest -m <backend> -k factory   ibis/backends/tests/test_map.py`

### DIFF
--- a/ibis/backends/tests/test_map.py
+++ b/ibis/backends/tests/test_map.py
@@ -11,6 +11,7 @@ import ibis
 import ibis.common.exceptions as exc
 import ibis.expr.datatypes as dt
 from ibis.backends.tests.errors import PsycoPg2InternalError, Py4JJavaError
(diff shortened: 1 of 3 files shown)


cpcloud:
I would like to close out the Map, Array and Struct API improvement PRs and start over with **just one** of them, with a very specific use case that doesn't work, and try to address that incrementally. These PRs are all interdependent and stuffed full of hugely-consequential API changes.

NickCrews:
sounds good, let's start with array for now.

PR #9461, opened 2024-06-27
test: add some basic cast tests

NickCrews:
discovered that casting arrays didn't work in druid when implementing https://github.com/ibis-project/ibis/pull/9458.
So I thought I was write down what I found.


cpcloud (approved):
Thanks!

Would you mind opening issues for these failures when you get a chance? No rush!

cpcloud:
Ok, after seeing this fail on BigQuery and Snowflake I took a closer look at the test case and I think there are some problems that need to be corrected.

Reverting.

PR #9466, opened 2024-06-28
fix(uri-parsing): handle password with bracket in connection url

grieve54706:
## Description of changes
If the password is with a bracket in the connection url, `urllib.parse.urlparse(url)` will raise the error `ValueError: Invalid IPv6 URL` in `backends.connect()`. But if the caller quotes the password to avoid parsing failing, ibis does not unquote back the password.

### Reproducible example
```python
import ibis
import pytest

invalid_password = "invalid_password["
with pytest.raises(ValueError) as e:
    ibis.connect(f"postgres://username:{invalid_password}@localhost:5432/dbname")
assert str(e.value) == "Invalid IPv6 URL"
```

### DIFF
--- a/ibis/backends/clickhouse/__init__.py
+++ b/ibis/backends/clickhouse/__init__.py
@@ -6,7 +6,7 @@ import glob
 from contextlib import closing
 from functools import partial
 from typing import TYPE_CHECKING, Any, Literal
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, unquote_plus, urlparse
 
 import clickhouse_connect as cc
 import pyarrow as pa
@@ -82,7 +82,7 @@ class Backend(SQLBackend, CanCreateDatabase):
 
         connect_args = {
             "user": url.username,
-            "password": url.password or "",
+            "password": unquote_plus(url.password) or "",
             "host": url.hostname,
             "database": database or "",
         }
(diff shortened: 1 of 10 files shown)


ibis/backends/snowflake/__init__.py
```diff
@@ -17,7 +17,7 @@
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 from urllib.parse import parse_qs, urlparse
-from urllib.request import urlretrieve
+from urllib.request import unquote_plus, urlretrieve
```
gforsyth, line 20: I think most of the CI failures stem from this import, which should be from `urllib.parse`

ibis/backends/clickhouse/__init__.py
```diff
@@ -82,7 +82,7 @@ def _from_url(self, url: str, **kwargs) -> BaseBackend:
 
         connect_args = {
             "user": url.username,
-            "password": url.password or "",
+            "password": unquote_plus(url.password) or "",
```
gforsyth, line 85: the `password` attribute can be `None` depending on the connection string, so there needs to be a `None` check before calling `unquote_plus`

gforsyth (commented):
Thanks for putting this in @grieve54706 !

I marked a few spots that are causing the CI failures

grieve54706:
@gforsyth Thanks for your suggestion. You saved my life.

cpcloud (approved):
LGTM, I'll do a few cleanups of the excepting string matching in #9482

cpcloud:
@grieve54706 @gforsyth Is there a corresponding issue for this PR? NBD if not, just wondering!

gforsyth:
I don't think there was

grieve54706:
Nope. Thank you for your time.

PR #9473, opened 2024-06-29
WIP: try ops.EmptyArray

NickCrews:
WIP, testing a different approach to https://github.com/ibis-project/ibis/pull/9458


NickCrews:
@cpcloud OK this is finally looking pretty good and is ready for a review!

ibis/expr/types/arrays.py
```diff
@@ -1108,15 +1116,49 @@ def array(values: Iterable[V]) -> ArrayValue:
...
+                f"If no type passed, values must be an array, got {values.type()}"
+            )
+
+    if values is None:
+        if type is None:
+            raise ValidationError("If values is None/NULL, type must be provided")
+        return ir.null(type)
+
+    values = tuple(values)
+    if len(values) == 0:
```
cpcloud, line 1150: Please use standard Python idioms for this.

```suggestion
    if not values:
```

ibis/expr/types/arrays.py
```diff
@@ -1108,15 +1116,49 @@ def array(values: Iterable[V]) -> ArrayValue:
...
+            return values
+        else:
+            raise ValidationError(
+                f"If no type passed, values must be an array, got {values.type()}"
+            )
+
+    if values is None:
+        if type is None:
+            raise ValidationError("If values is None/NULL, type must be provided")
+        return ir.null(type)
```
cpcloud, line 1147: I don't like this. It's unnecessary to have to support another way to construct a null value.

If someone wants a `NULL` with an array type, they should use the `null` function.

ibis/expr/types/arrays.py
```diff
@@ -1108,15 +1116,49 @@ def array(values: Iterable[V]) -> ArrayValue:
...
-    │ [3, 47]              │
+    │ [1.0, 47.0]          │
+    │ [2.0, 47.0]          │
+    │ [3.0, 47.0]          │
     └──────────────────────┘
     """
-    return ops.Array(tuple(values)).to_expr()
+    type = dt.dtype(type) if type is not None else None
+    if type is not None and not isinstance(type, dt.Array):
+        raise ValidationError(f"type must be an array, got {type}")
```
cpcloud, line 1132: Can we just let this fall through and bubble up as it would? This function is already getting stuffed with a bunch of branching, and this doesn't seem to add much value.

ibis/expr/types/arrays.py
```diff
@@ -1108,15 +1116,49 @@ def array(values: Iterable[V]) -> ArrayValue:
...
+        value_type = type.value_type if type is not None else None
+        values = [_value(v, value_type) for v in values]
+        return ops.Array(values).to_expr()
+
+
+def _value(x, type) -> ir.Value:
+    if isinstance(x, (ir.Value, Deferred)):
+        return x.cast(type) if type is not None else x
+    else:
+        return ibis.literal(x, type=type)
```
cpcloud, line 1164: What is the purpose of `literal` here?

cpcloud (changes requested):
There's still too much unnecessary munging here in my opinion.

ibis/expr/types/arrays.py
NickCrews: If you do ibis.array([1, 2], type="array<float>"), we somehow need to get those ints to floats. My original implementation stored them in the array as ints, and convert to float during compilation. But that required storing the eventual type in ops.Array. so here we pre convert all the elements to floats, and now ops.Array can just use rlz.highest_precedence_dtype(). In order to do that pre casting, we need this literal.

ibis/expr/types/arrays.py
NickCrews: I'm guarding against the footgun of someone passing in "float" instead of "array<float>" (I literally accidentally did this as I adjusted the tests this last round), which would silently give the wrong result for some inputs. I agree not needed, I can remove, just a nicety

ibis/expr/types/arrays.py
NickCrews: I have a thought here but I better do it at my computer

ibis/expr/types/arrays.py
NickCrews: Oops sorry will do.

ibis/expr/types/arrays.py
NickCrews: See these tests in mismo:
- [test_jaccard](https://github.com/NickCrews/mismo/blob/e536163d779f4b0a5b2c6dd13a0d3c4587c6250d/mismo/sets/tests/test_compare.py#L10-L27)
- [test_array_min](https://github.com/NickCrews/mismo/blob/e536163d779f4b0a5b2c6dd13a0d3c4587c6250d/mismo/arrays/tests/test_array.py#L13-L30)
- [test_array_mean_median](https://github.com/NickCrews/mismo/blob/e536163d779f4b0a5b2c6dd13a0d3c4587c6250d/mismo/arrays/tests/test_array.py#L53-L80)

I need to construct an `array<int>` with a variety of values:
- NULL
- []
- [1,2,3]

There are three different ibis APIs for doing this:
- `ibis.null(type="array<int>")`: this only works for the NULL case. I would need some if/else branching in my test setup to use this for the NULL case, but would need something else for the other cases.
- `ibis.literal(val, type="array<int>")`. This is currently what I do, and it works for all 3 inputs here, no if/else branching required. But the docstring for literal() says that it is going to get deprecated soon for constructing complex types, and I should use array()/map()/struct() instead.
- `ibis.array()`: If we keep the functionality the way I propose here, then I can do `ibis.array(val, type="array<int>")` and it works for all three input cases. If we don't allow passing in None, then I would need to do some if/else branching to use ibis.null() in that case.

I think it can be summarized as SOMEONE is gonna need to do this `if values is None:` branching, and I would rather have it be down here in one place in library code, than in many places in user code.

NickCrews:
I'm gonna be out of office for this week. So I hope when I get back I see that y'all will have figured it all out and everything is merged and happy 😁

cpcloud:
Part of the reason that pandas API is so large and complex to use and maintain (in my opinion) is that it favors flexibility over consistency. I think we should err on the side of consistency as a project, and provide flexibility where there's not a ton of additional complexity.

That means that sometimes the complexity of handling multiple input types gets pushed to the user and I would argue that that is the best place to handle non-trivial input normalization since the application has much better information about how it wants those inputs to look.

I am pretty against opening up all the `ibis.*` constructor APIs to accept other Ibis expressions as it adds a bunch of requirements and decision making around what's valid and how we validate the input _beyond_ what we're already doing.

For nested types and tables, things get _really_ complex as now you have to decide whether to accept **mixed** Ibis-expression-Python-value objects, and are we going to validate all of that?

ibis/expr/types/arrays.py
```diff
@@ -1108,15 +1116,49 @@ def array(values: Iterable[V]) -> ArrayValue:
...
+    │ [2.0, 47.0]          │
+    │ [3.0, 47.0]          │
     └──────────────────────┘
     """
-    return ops.Array(tuple(values)).to_expr()
+    type = dt.dtype(type) if type is not None else None
+    if type is not None and not isinstance(type, dt.Array):
+        raise ValidationError(f"type must be an array, got {type}")
+
+    if isinstance(values, ir.Value):
```
cpcloud, line 1134: I don't think we're going to accept inputs like this. Can you get rid of this code path?

ibis/expr/types/arrays.py
cpcloud: I'm still not sure why this is required. Is there some API that fails to execute if you don't do these casts?

What if the elements themselves are arrays? Doesn't this then need to be recursive in the casting/call to `_value`?

cpcloud (changes requested):
Can we please leave all the additional complexity and decisions out here beyond addressing the empty array case and allow the user to specify a type?

ibis/expr/types/arrays.py
NickCrews: Shoot, github escaped the `<>` I posted, I meant `ibis.array([1, 2], type="array<float>")` not `ibis.array([1, 2], type="array")`

Now does the concern make sense? Two ways of storing this:
1. store the inputs as is, and keep track of the dtype separately `op.Array(values=[1,2], value_type="float")`
2. cast the inputs early, then we don't need to store the dtype: `op.Array(values=[op.Literal(1, "float"), op.Literal(2, "float")])`

I am trying to go for option 2.

> I'm still not sure why this is required

Specificaly, if I remove the .literal() here, then we pass a plain python `1` into the op.Array constructor, and then using pattern matching it tries to coerce that to an ibis expression, which yields a `op.Literal(1, int)`. But we wanted it to be a float, so we have to do the casting up front here.

> What if the elements themselves are arrays

you mean what if x were a python list of ibis values? ie someone did `array_of_arrays = ibis.array([[ibis.literal(1)], [ibis.literal(2)]])`? Yes, this would fail, good catch. not sure what to do about it yet....

NickCrews:
> Can we please leave all the additional complexity and decisions out here beyond addressing the empty array case and allow the user to specify a type?

Yeah I think this is probably a good idea. I'll trim it down as I can.

To confirm, I am assuming that `array_of_arrays = ibis.array([[ibis.literal(1)], [ibis.literal(2)]])` is a valid input. You also want to be able to support this, right?

cpcloud:
> To confirm, I am assuming that `array_of_arrays = ibis.array([[ibis.literal(1)], [ibis.literal(2)]])` is a valid input. You also want to be able to support this, right?

No, I'm saying we shouldn't support this. It opens the door to mixing ibis expressions and python types, and we'd have to unify both and it's a lot of additional complexity.

NickCrews:
I currently have lots of code that looks like `t.mutate(emails=ibis.array([t.email0, t.email1, t.email2])`. Am I missing something here, how can I accomplish this goal?

cpcloud:
@NickCrews I am still having a lot of trouble following everything here. I suspect that the PR is still trying to do too much.

Can you provide a minimum list of cases showing

1. the current behavior
2. the desired behavior


and avoid speculating on specific solutions? This problem space (complex type casting and inference) is **really** hairy, and it's very very difficult to review related PRs if they are trying to address a bunch of things at once.

NickCrews:
Yes, I agree it is really hairy, and I think that would help. Maybe I start with a PR that just adds or changes tests to precisely show what the current behavior is

ncclementi:
@NickCrews what's the status of this one? Is the plan to close it and start a new separate PR with a more targeted scope?

PR #9477, opened 2024-06-29
feat(caching): tie lifetime of cached tables to python refs

coroa:
## Description of changes

Makes use of `weakref.finalizer`s to release the cached table automatically when the cached table (or more precisely its `.op()`) is garbage collected.

```python
In [1]: import pandas as pd

In [2]: import ibis

In [3]: con = ibis.duckdb.connect()

In [4]: df = pd.DataFrame([[0, 1], [2, 3]], columns=["foo", "bar"])

In [5]: noncached = con.create_table("tab", df).mutate(foo=2 * tab.foo)

In [6]: cached = noncached.cache()

In [7]: con.tables
Out[7]: 
Tables
------
- ibis_cache_wvrjoptclzhudglqapz4zusnlu
- tab

In [8]: del cached

In [9]: con.tables
Out[9]: 
Tables
------
- tab
```

If I understand the nodal expression tree structure correctly, then `cached.op()` is preserved in all dependent expressions. This PR adds a `weakref.finalize(cached.op(), ...)` ([docs](https://docs.python.org/3/library/weakref.html#weakref.finalize)) onto `cached.op()` which releases the physical cached copy from the databases.

The idempotency of `.cache()` is also preserved by returning the same table reference.


coroa:
The `bidi` dependency (for a bidirectional dictionary) is not necessary anymore with the RefCountedCache implementation in this PR, but quite obviously my [attempt](https://github.com/ibis-project/ibis/pull/9477/commits/980f37b99e166f70d8486fe241bf4a901ab5d7e9) at removing it was incomplete.

conda/environment.yml
```diff
@@ -3,9 +3,8 @@ channels:
   - conda-forge
 dependencies:
   # runtime dependencies
-  - apache-flink
+  # - apache-flink
```
coroa: Ups, my custom environment fixes bled into this commit, inadvertently. Good thing, i had to revert it.

ibis/expr/types/relations.py
```diff
@@ -4672,13 +4650,19 @@ def value_counts(self) -> ir.Table:
...
-        self.release()
+        self.release(stacklevel=3)
 
     def __enter__(self):
         return self
 
-    def release(self):
+    def release(self, stacklevel=2):
         """Release the underlying expression from the cache."""
+        warnings.warn(
```
cpcloud, line 4660: Should we remove these in version 10? If so, let's add a message here about that.

Alternatively, I think `util.deprecated` might be possible to use here, which would give a slightly more standardized (standard from the perspective of Ibis anyway) error message.

ibis/common/caching.py
```diff
@@ -47,56 +49,50 @@ def __init__(
...
-
-        self.refs[inv_key] -= 1
-        assert self.refs[inv_key] >= 0, f"refcount is negative: {self.refs[inv_key]:d}"
-
-        if not self.refs[inv_key]:
-            del self.cache[inv_key], self.refs[inv_key]
-            self.finalize(key)
+        cached = self.lookup(name)
+        finalizer = finalize(cached, self._release, key)
+        finalizer.atexit = False
```
cpcloud, line 75: Does this mean that cached tables are not guaranteed to be cleaned up?

ibis/common/caching.py
```diff
@@ -29,8 +28,11 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+CacheEntry = namedtuple("CacheEntry", ["name", "ref", "finalizer"])
+
+
 class RefCountedCache:
-    """A cache with reference-counted keys.
+    """A cache with (implicitly) reference-counted values.
```
cpcloud, line 35:
```suggestion
    """A cache with implicitly reference-counted values.
```

Seems important enough to not parenthesize!

ibis/common/caching.py
```diff
@@ -47,56 +49,50 @@ def __init__(
...
-        self.refs: Counter = Counter()
         self.populate = populate
         self.lookup = lookup
         self.finalize = finalize
-        # Somehow mypy needs a type hint here
-        self.names: defaultdict = defaultdict(generate_name)
+        self.generate_name = generate_name
         self.key = key or (lambda x: x)
 
+        self.cache: dict[Any, CacheEntry] = dict()
```
cpcloud, line 58: Not to relitigate this whole PR but would a `WeakValueDictionary` also work here instead of explicit finalizers?

Not sure if `WeakValueDictionary` lets you control what happens on weakref finalization, but if it doesn't that I suppose that would be a non-starter.

ibis/common/caching.py
```diff
@@ -47,56 +49,50 @@ def __init__(
...
-            return default
+        if (res := self.cache.get(key)) is not None:
+            return res.ref()
+        return default
 
     def __getitem__(self, key):
-        result = self.cache[key]
-        self.refs[key] += 1
-        return result
+        return self.cache[key].ref()
```
cpcloud, line 66: Should this raise a `KeyError` if it returns `None`?

ibis/common/caching.py
```diff
@@ -47,56 +49,50 @@ def __init__(
...
+        finalizer = finalize(cached, self._release, key)
+        finalizer.atexit = False
+
+        self.cache[key] = CacheEntry(name, ref(cached), finalizer)
+
+        return cached
+
+    def release(self, name: str) -> None:
+        # Could be sped up with an inverse dictionary,
+        # but explicit release is discouraged, anyway
```
cpcloud, line 83: This seems fine for now. Worst case scenario is enormous amounts of cached tables, but I suspect this would scale well into the tens of thousands. We can always revisit if this becomes a bottleneck.

ibis/expr/types/relations.py
```diff
@@ -1881,8 +1882,8 @@ def difference(self, table: Table, *rest: Table, distinct: bool = True) -> Table
         └───────┘
         """
         node = ops.Difference(self, table, distinct=distinct)
-        for table in rest:
-            node = ops.Difference(node, table, distinct=distinct)
+        for other in rest:
+            node = ops.Difference(node, other, distinct=distinct)
```
cpcloud, line 1886: Can we leave these as they were just to reduce review noise?

cpcloud (commented):
Thanks for digging around in the guts here :)

A couple of questions and requests, but I generally like where this PR is going!

ibis/common/caching.py
cpcloud: Seems like the finalizer behavior of the values in a `WeakValueDictionary` can't easily be customized.

ibis/common/caching.py
```diff
@@ -47,56 +49,50 @@ def __init__(
...
-            self.finalize(key)
+        cached = self.lookup(name)
+        finalizer = finalize(cached, self._release, key)
+        finalizer.atexit = False
+
+        self.cache[key] = CacheEntry(name, ref(cached), finalizer)
+
+        return cached
+
+    def release(self, name: str) -> None:
```
cpcloud, line 81: Should we just merge the implementation of `_release` into this method?

cpcloud: Oh, I see you're using it elsewhere, nevermind!

ibis/expr/types/relations.py
```diff
@@ -4672,13 +4650,19 @@ def value_counts(self) -> ir.Table:
...
 
     def __enter__(self):
         return self
 
-    def release(self):
+    def release(self, stacklevel=2):
         """Release the underlying expression from the cache."""
+        warnings.warn(
+            "Explicitly releasing cached tables is deprecated "
+            "(they are released when the table relation leaves scope)",
```
cpcloud, line 4662:
```suggestion
            "Explicitly releasing cached tables is deprecated. "
            "Tables are released when the table expression leaves scope.",
```

ibis/common/caching.py
```diff
@@ -47,56 +49,50 @@ def __init__(
...
     ) -> None:
-        self.cache = bidict()
-        # Somehow mypy needs a type hint here
-        self.refs: Counter = Counter()
         self.populate = populate
         self.lookup = lookup
         self.finalize = finalize
-        # Somehow mypy needs a type hint here
-        self.names: defaultdict = defaultdict(generate_name)
+        self.generate_name = generate_name
```
cpcloud, line 55: Seems like there's only a single call site for this, can we inline the `generate_name` in `store` and remove the attribute altogether?

cpcloud: Can also be done in a follow-up!

cpcloud: Actually, nevermind, I'll submit a PR to clean some stuff up here that's independent of this PR!

ibis/expr/types/relations.py
jcrist: I actually would prefer we keep these ways to explicitly release cached tables. Complex code can make understanding if the cached tables has any references left tricky, and can also result in code littered in dels (which is a code smell imo). There should be a way to make both patterns still work after this change.

cpcloud: Yeah, that makes sense to me. Keeping these seems like a good idea.

coroa: > There should be a way to make both patterns still work after this change.

Both patterns do work with the current implementation and can continue to work (the linear search in `release` should maybe be helped by an inverse dictionary).

ibis/common/caching.py
coroa: `atexit` means when called by the interpreter shutdown ([docs](https://docs.python.org/3/library/atexit.html)), ie. in a situation where the sessions are being closed (or have been, actually). I assumed that, since we create temporary tables, which afaik are only valid for a current session, it was redundant to explicitly request their removal.

ibis/expr/types/relations.py
coroa: pre-commit's ruff insisted. Happy to remove.

ibis/common/caching.py
coroa: It needs to loose a race for that to happen (since the reference invalidation triggers the removal of the cache entry) and `__getitem__` is unused, but you are right!

ibis/common/caching.py
coroa: Do we want to remove the `DeprecationWarning` and maybe update the docs and release notes to make the new behaviour better known, then?

ibis/expr/types/relations.py
coroa: Do we want to remove the `DeprecationWarning` and maybe update the docs and release notes to make the new behaviour better known, then?

ibis/common/caching.py
cpcloud: It's not necessarily redundant, because unfortunately not all backends have support for temporary tables.

cpcloud:
I will clean up the `ruff` lints in #9489!

ibis/expr/types/relations.py
cpcloud: Yeah, that sounds like a good idea!

ibis/common/caching.py
cpcloud: I *think* we want keep the default value of `atexit` (`True`), and allow the cleanup to run on interpreter shutdown.

coroa: > I think we want keep the default value of atexit (True), and allow the cleanup to run on interpreter shutdown.

Ok, then let's test that.

cpcloud: > depends on whether the database connection is still there

True, and that's a big if since it renders the handler useless if the connection isn't there.

I'd say let's leave the default behavior for now.

coroa:
> I will clean up the `ruff` lints in #9489!

Once that gets merged, i'll rebase ontop

cpcloud:
@coroa #9489 is merged!

ibis/expr/types/relations.py
coroa: Ok, reverted the changes around the `DeprecationWarning` and added some additional detail to the docstring.

cpcloud (approved):
LGTM!

coroa:
Ok, I rebased ontop of `main` and removed the DeprecationWarning and updated the `.cache()` doc-string with another sentence on the explicit and implicit cache eviction.

Only things left on my list of todos are:
- [ ] Release notes
- [ ] Check and update user guid

cpcloud:
Looks like there's maybe a bit of `DeprecationWarning` still hanging around?

cpcloud:
A release note will be generated from the commit message, so no need to do anything there.

coroa:
> Looks like there's maybe a bit of `DeprecationWarning` still hanging around?

Ah yes, i forgot about all the tests i had to change :).

coroa:
Couldn't find any user-guide explanations of the cache'ing mechanism. So i guess the doc-string is the best we have atm. And that is up-to-date.

So, from my side this is good to go.

ibis/common/caching.py
coroa: I actually had one originally and look into the implementation whether subclassing one could easily replace what i am doing here. I did not see a good way to do it more easily.

cpcloud (approved):
LGTM!

@jcrist Did you wanna take a look here again?

ibis/common/caching.py
```diff
@@ -47,56 +49,53 @@ def __init__(
...
+                return
+
+        raise IbisError(
+            "Key has already been released. Did you call "
+            "`.release()` twice on the same expression?"
+        )
+
+    def _release(self, key) -> None:
+        entry = self.cache.pop(key)
+        self.finalize(entry.name)
```
jcrist, line 99: This call may error for a few reasons:
- Table doesn't exist (I think we should ignore these always, as noted elsewhere).
- Spurious error
- Connection is in weird state since python is finalizing

Since _most_ backends will cleanup temp tables automatically (and errors on exit can give a bad impression of a tool), I wonder if want to do something like:

```python
try:
    self.finalize(entry.name)
except Exception:
    # silence all errors if system is shutting down
    if not sys.is_finalizing():
        raise
```

This way we can still catch bugs in our code since most release calls won't silence everything, but in the case the system is shutting down and e.g. the network is in a weird state we don't see a mountain of failed-to-release-table errors on exit.

jcrist (commented):
Overall LGTM! This is something we'd talked about doing internally before, nice to see it handled in such a nice way be a new contributor!

coroa (commented):
Thanks for the nice words

coroa (commented):
Thanks for the nice words

ibis/backends/sql/__init__.py
```diff
@@ -275,8 +275,8 @@ def drop_view(
     def _load_into_cache(self, name, expr):
         self.create_table(name, expr, schema=expr.schema(), temp=True)
 
-    def _clean_up_cached_table(self, op):
-        self.drop_table(op.name)
+    def _clean_up_cached_table(self, name):
+        self.drop_table(name)
```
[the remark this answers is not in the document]
coroa: Done

ibis/backends/duckdb/__init__.py
```diff
@@ -281,8 +281,8 @@ def create_table(
     def _load_into_cache(self, name, expr):
         self.create_table(name, expr, schema=expr.schema(), temp=True)
 
-    def _clean_up_cached_table(self, op):
-        self.drop_table(op.name)
+    def _clean_up_cached_table(self, name):
+        self.drop_table(name)
```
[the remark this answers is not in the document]
coroa: Done, and works

ibis/common/caching.py
```diff
@@ -47,56 +49,53 @@ def __init__(
...
+        # but explicit release is discouraged, anyway
+        for key, entry in self.cache.items():
+            if entry.name == name:
+                self._release(key)
+                return
+
+        raise IbisError(
+            "Key has already been released. Did you call "
+            "`.release()` twice on the same expression?"
+        )
```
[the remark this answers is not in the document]
coroa: Makes sense

ibis/common/caching.py
coroa: Ok, I added that.

coroa:
Unsure what tripped the mssql backend. Unfortunately, have not been able to run those tests locally.

cpcloud:
One transient error and one XPASS, which is a win!

cpcloud:
I'll fix this up and merge!

cpcloud:
Clouds are passing:

```
…/ibis on  self-releasing-cached-tables is 📦 v9.1.0 via 🐍 v3.10.14 via ❄️   impure (ibis-3.10.14-env)
bringing up nodes...
.x.x...........................x....s................x..........s...........................x.................xx..........x..x............x..........................x...x........................... [  5%]
............x....................x....................s..................s................................................s......................................................s..x.x.......x...... [ 10%]
x.......x.x..x.........x....x..x.xx....x..........x....................x....................................................x...............................................sssssssssssssssssssss.... [ 16%]
xxxxxxxxxx.x.xx.x.x....x.x.....................................x.............x........x......xx.x.........................x.......x....x.x...........x...x.......................x................... [ 21%]
...x....x..........................x.x....................x........x.......x.x.......x..x......................x..........................xx...............x......x.x..x...x...x......x.............. [ 26%]
.....x.............................x....xx.......x......x..xxxx.xx.x.x.xxx..xx.x...xxxx...x..xx.xxx..x.x..xx...xx.x.x...x......xxx.xxxx.xxx.x...xx.xx.xxxxxx.xx...x.xx.......x.........x.....x....... [ 32%]
xx..........x............................x...x..x..................x....x.....x............x.....x................x..x.....xxx..x.x...xxx.xxxx.xx......xx.x...x..x.....x..............x.............. [ 37%]
..x.x.x.x.............x..x.....x..x.x.......x.x............................x....x..x.x..x...x..x.....x...x..............x...xxxxx....x..................x.........x.......xx..x.x.................... [ 42%]
.....s.............x..............s.x...x....x..............sx..xx.s.x...x..x.......s......x.x......x..........x...................x...........x...........................x..................xx..... [ 48%]
.x.........x.....................x...........................s..........................s....................s..........x................................x..............x.x...x..x...........x...x... [ 53%]
..............xxx.....xx......................x.xx..x......x.x..x...x.x......x............x.x..x.......x..................x.......xx..xx..xx...xxxxxxxxxx...xxx...xxxx.xxxx..x...x.x..x..x.xxx.xx..x. [ 58%]
xxxxx..x.x.x...x..................x..x...x...x....x........x........x.....x.x....x.............x.x.x...xx.....x....x...x.......x...xxxxx................x................x..xx....xx................. [ 64%]
....x...x....x................x.............x.x.....x........x.x.x....................x..........x..xx.....x.......x.x...........x..xx.......x..x................x.....x......xxx...x.x.xxx..xxxxxxxx [ 69%]
xxxx.xxxxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx..x......................x.........................x.....................x...x........ [ 74%]
....................................................................x.xxxxxx.xx.x....xx.xx....x..x.......x....................x..............x........x.x.xx...x.x...x.....x....x...xx..............x [ 80%]
..xx....x...xx..x..x.....x.....................x..x.................x.x.....x.......x...........x........x....x..x.x.......x.......x................................................................. [ 85%]
..........................................................................................................................................................................s.......................... [ 90%]
.........x.............................................................................................................................................................................x.x........... [ 96%]
......................................................s...ss......................................x...........................................                                                        [100%]
3097 passed, 39 skipped, 552 xfailed in 840.67s (0:14:00)
```

cpcloud:
@coroa Thanks for pushing this through, great work!

coroa:
Thanks for the quick reviews and comments!

Issue #9474, opened 2024-06-29
bug: duplicate column names in filters leads to error


NickCrews:
### What happened?

```python
import ibis

t = ibis.examples.penguins.fetch()
t.filter(
    t.species == "Adelie",
    t.species == "Adelie",
)
# IbisInputError: Duplicate column name "Equals(species, 'Adelie')" in result set
```
Since neither of these columns are actually in the result set, I would expect this to work.

OK, I admit this is a bit of a contrived example, but there are instances where different expressions have the same name, which is how I actually ran into this bug:

```python
t.filter(
    t.island.fill_null("") == "Adelie",
    t.species.fill_null("") == "Adelie",
)
# IbisInputError: Duplicate column name "Equals(Coalesce(), 'Adelie')" in result set
```

perhaps that is just a bug with coalesce though? We should make them have different names?

I think the "pure" solution would be to do some renaming of these temporary columns on the fly inside .filter(), but that might be really hairy, so an easier mitigation would be just to stamp out instances where different expressions map to the same name.

### What version of ibis are you using?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
> I think the "pure" solution would be to do some renaming of these temporary columns on the fly inside .filter()

I don't think so. I think we might want to instead relax whatever constraint is enforcing this, but **only** inside `filter`. Best to look into what's happening before we speculate on solutions though :)

cpcloud:
I think the expression naming collision in this case is also an issue, as the same problem shows up in a `select` invocation.

PR #9480, opened 2024-06-30
fix(ir): flatten iterables when constructing name, to fix incorrect duplication of names

cpcloud:
Addresses an issue found while fixing #9474 where we didn't use variadic tuple
arguments when constructing the names of ops.


ibis/backends/bigquery/tests/unit/snapshots/test_compiler/test_to_timestamp_timezone/out.sql
```diff
@@ -1,3 +1,3 @@
 SELECT
-  parse_timestamp('%F %Z', CONCAT(`t0`.`date_string_col`, ' America/New_York'), 'UTC') AS `StringToTimestamp_StringConcat_'%F %Z'`
+  parse_timestamp('%F %Z', CONCAT(`t0`.`date_string_col`, ' America/New_York'), 'UTC') AS `StringToTimestamp_StringConcat_date_string_col_' America_New_York'_'%F %Z'`
```
gforsyth, line 2: I think this is just ugly, not an issue, since bigquery quotes with backticks, yeah?

gforsyth (approved):

ibis/backends/bigquery/tests/unit/snapshots/test_compiler/test_to_timestamp_timezone/out.sql
cpcloud: Yeah, it's gross, but it should be fine. I'll add a `system` test to verify that the column name is allowed.

cpcloud: Done, and passing:

```
cloud in 🌐 falcon in …/ibis on  fix-duplicate-projections is 📦 v9.1.0 via 🐍 v3.10.14 via ❄️   impure (ibis-3.10.14-env)
❯ pytest ibis/backends/bigquery/tests/system/test_client.py::test_complex_column_name -q
.                                                                                                                                                                                      [100%]
1 passed in 4.88s
```

cpcloud:
Just a note here: for very complex unnamed expressions, `name` can be enormous. The solution for now is to not let ibis generate names, and to give things explicit names. Most of the time our APIs make that easy or required, so this case should be relatively rare I think.

PR #9481, opened 2024-06-30
revert: test: add some basic cast tests

cpcloud:
Reverts ibis-project/ibis#9461


cpcloud:
@NickCrews Can you open another PR with these changes?

I'll explain what I think needs to change in that one.

PR #9482, opened 2024-06-30
fix(urls): standardize pass-through of parsed query parameters

### LINKED ISSUE #9456: Bigquery: ibis.connect doesn't pass through parameters parsed from connection string
With version 9.0.0 of ibis, passing:

`ibis.connect(bigquery://my_project/my_dataset?location=us-central1)`

Does not pass the parameters to `backend._from_url`. I would expect `location=us-central1` to be passed to the underlying bigquery connection.

I'm not sure if this is intentional, because the `ibis.connect` function goes to the trouble of parsing the parameters from the connection string but then uses `orig_kwargs` for 

https://github.com/ibis-project/ibis/blob/33ec7547209cc7775f85bd46a30f971333ea35bb/ibis/backends/__init__.py#L1394

Alternatively, it could be solved in a backend specific way for bigquery by parsing out the configuration here:

https://github.com/ibis-project/ibis/blob/33ec7547209cc7775f85bd46a30f971333ea35bb/ibis/backends/bigquery/__init__.py#L335

My usecase (and why I don't particularly want to specify location directly) is that I want users to be able to pass any connection string they want through to Ibis from configuration and use this as a means to configure the connection appropriately, subject to me filtering supported backends for my application. I haven't tried parameters with any other of the backends.

This is similar to the sqlalchemy connection string logic mentioned in https://github.com/ibis-project/ibis/issues/2532#issuecomment-794573375. 

Any advice on how the general ibis connection strings handle parameters like this for a consistent user experience? I'd be happy to submit a PR for the latter option, it should be quite simple.

cpcloud:
This PR cleans up our URL handling a bit.

1. We were doing a bunch of unnecessary url parse -> url unparse, only to
   immediately again parse the URL as the first call in every `_from_url`
   implementation.
1. Each backend was separately handling converting query parameters to single
   scalars where possible, due to the unparsing
1. Query parameters were sometimes passed into `kwargs` and sometimes parsed
   again to then be used as kwargs later

So, I did the following:

1. Pass the `ParseResult` into the `_from_url` implementation.
1. Convert query parameters into single values where possible _before_ calling `backend._from_url`
1. Pass all query parameters as `**kwargs` into the `_from_url` call.

Fixes #9456.

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -9,7 +9,6 @@ import re
 import urllib.parse
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, ClassVar
-from urllib.parse import parse_qs, urlparse
 
 import ibis
 import ibis.common.exceptions as exc
@@ -21,6 +20,7 @@ from ibis.common.caching import RefCountedCache
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Mapping, MutableMapping
+    from urllib.parse import ParseResult
 
     import pandas as pd
     import polars as pl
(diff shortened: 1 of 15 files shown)


gforsyth:
Do we want to get #9466 in first?

cpcloud:
> Do we want to get #9466 in first?

Yep.

Is that waiting on anything?

cpcloud:
Merged #9466 and rebased this PR on `main`. Hopefully just waiting on CI at this point!

gforsyth (approved):
noice! :shipit:

PR #9483, opened 2024-06-30
test(redo): add some basic cast tests

NickCrews:
redo of #9461

### DIFF
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -28,6 +28,7 @@ from ibis.backends.tests.errors import (
     OracleDatabaseError,
     PolarsSchemaError,
     PsycoPg2InternalError,
+    PsycoPg2SyntaxError,
     Py4JJavaError,
     PyDruidProgrammingError,
     PyODBCDataError,
@@ -1756,6 +1757,55 @@ def test_hexdigest(backend, alltypes):
     h2 = df["string_col"].apply(hash_256).rename("HexDigest(string_col)")
 
     backend.assert_series_equal(h1, h2)
+
+
+@pytest.mark.parametrize(
+    ("from_type", "to_type", "from_val", "expected"),
+    [
+        param("int", "float", 0, 0.0, id="int_to_float"),
+        param("float", "int", 0.0, 0, id="float_to_int"),
+        param("string", "int", "0", 0, id="string_to_int"),
+        param("string", "float", "0", 0.0, id="string_to_float"),
+        param(
+            "array<int>",
+            "array<string>",
+            [0, 1, 2],
+            ["0", "1", "2"],
+            marks=[
+                pytest.mark.notimpl(["pandas"], reason="casts to ['0']"),
+                pytest.mark.notimpl(["druid"], raises=PyDruidProgrammingError),
+                pytest.mark.notimpl(["oracle"], raises=OracleDatabaseError),
+                pytest.mark.never(
+                    ["exasol", "impala", "mssql", "mysql", "sqlite"],
+                    reason="backend doesn't support arrays",
+                ),
+            ],
+            id="array",
+        ),
+        param(
+            "struct<a: int, b: string>",
+            "struct<a: string, b: int>",
+            {"a": 0, "b": "1"},
+            {"a": "0", "b": 1},
+            marks=[
+                pytest.mark.notimpl(["flink"], raises=Py4JJavaError),
+                pytest.mark.notimpl(["druid"], raises=PyDruidProgrammingError),
+                pytest.mark.notimpl(["oracle"], raises=OracleDatabaseError),
+                pytest.mark.notimpl(["postgres"], raises=PsycoPg2SyntaxError),
+                pytest.mark.notimpl(["risingwave"], raises=PsycoPg2InternalError),
+                pytest.mark.never(
+                    ["datafusion", "exasol", "impala", "mssql", "mysql", "sqlite"],
+                    reason="backend doesn't support structs",
+                ),
+            ],
+            id="struct",
+        ),
+    ],
+)
+def test_cast(con, from_type, to_type, from_val, expected):
+    expr = ibis.literal(from_val, type=from_type).cast(to_type)
+    result = con.execute(expr)
+    assert result == expected
 
 
 @pytest.mark.notimpl(["pandas", "dask", "oracle", "sqlite"])


ibis/backends/tests/test_generic.py
```diff
@@ -1758,6 +1759,74 @@ def hash_256(col):
...
+    [
+        param("int", "float", 0, 0.0),
+        param("float", "int", 0.0, 0),
+        param("string", "int", "0", 0),
+        param("string", "float", "0", 0.0),
+        param(
+            "timestamp",
+            "string",
+            datetime.datetime(2023, 1, 1),
+            "2023-01-01 00:00:00",
```
cpcloud, line 1773: Why is this the expected value?

Why wouldn't we include the subsecond values in the string output, for example?

I don't think this particular test is fleshed out enough to be included here, it's stipulating a specific behavior for something that needs more discussion before writing a test.

You've marked a bunch of these tests as broken, and that's not really true here. Broken should really be reserved for things that are bugs in Ibis that are ~impossible to workaround, behaviors in the backend that are ~impossible to workaround, or bugs in the backend that are ~impossible to workaround.

Here's it's just the old problem of "things behave differently".

I think these test cases should be removed, because there's no consensus on what the behavior should be other than "return a string".

ibis/backends/tests/test_generic.py
```diff
@@ -1758,6 +1759,74 @@ def hash_256(col):
...
+        ),
+        param(
+            "struct<a: int, b: string>",
+            "struct<a: string, b: int>",
+            {"a": 0, "b": "1"},
+            {"a": "0", "b": 1},
+            marks=[
+                pytest.mark.never(
+                    ["datafusion", "exasol", "impala", "mssql", "mysql", "sqlite"],
+                    reason="backend doesn't support arrays",
```
cpcloud, line 1812:
```suggestion
                    reason="backend doesn't support structs",
```

ibis/backends/tests/test_generic.py
```diff
@@ -1758,6 +1759,74 @@ def hash_256(col):
...
+                ),
+                pytest.mark.notimpl(["flink"], raises=Py4JJavaError),
+                pytest.mark.notimpl(["druid"], raises=PyDruidProgrammingError),
+                pytest.mark.notimpl(["oracle"], raises=OracleDatabaseError),
+                pytest.mark.notimpl(["postgres"], raises=PsycoPg2SyntaxError),
+                pytest.mark.notimpl(["risingwave"], raises=PsycoPg2InternalError),
+            ],
+        ),
+    ],
+    ids=str,
```
cpcloud, line 1822: Please give these more meaningful names. Right now the test ids look like line noise 😂

ibis/backends/tests/test_generic.py
```diff
@@ -1758,6 +1759,74 @@ def hash_256(col):
     backend.assert_series_equal(h1, h2)
 
 
+@pytest.mark.parametrize(
+    ("from_type", "to_type", "from_val", "expected"),
+    [
+        param("int", "float", 0, 0.0),
+        param("float", "int", 0.0, 0),
+        param("string", "int", "0", 0),
+        param("string", "float", "0", 0.0),
```
cpcloud, line 1768: Are you sure we aren't testing these anywhere?

cpcloud (dismissed):

ibis/backends/tests/test_generic.py
NickCrews: Yeah good thought. If you grep for `def test.*_cast`, you can find that there is definitely some duplication in specific backend tests. But it doesn't look like there is anything in `ibis/backends/tests/`, except for a few specific ones in `test_temporal.py`. There are a few backend-less tests in `ibis/expr/datatypes/tests/test_cast.py`. They look to be like 7 years old!

In an ideal world I would go for
1. As much as possible, we pull tests from the specific backends to be shared by all backends
2. When there is duplication, I would remove the backend-less tests, I would rather fully test for the full executed value instead of merely "does our expression system say the result is the correct type"

what do you think?

ibis/backends/tests/test_generic.py
NickCrews: yeah this is a good point, thanks. So should we just open an issue to track that, and delete these cases? I'm fine with that.

re test labels: what should I use as the marker for "backends behave differently, it's possible for us to paper over the top of that in ibis, but we just haven't gotten to it yet"? `mark.notimpl`? If so, I can update https://ibis-project.org/contribute/02_workflow#setting-up-non-trivial-backends to match, as I interpreted that to mean that I should use `mark.broken` so I think it is a bit misleading.

cpcloud: Yeah, we should probably revisit those docs a bit.

Also possible we all have slightly different interpretations of the markers so maybe we should clarify that too :)

I think these would be ... notyet?

Perhaps it's time to consider a new marker for this scenario? I'm sure @gforsyth has some thoughts here.

gforsyth: `pytest.mark.welcome_to_the_jungle`?

I agree that `broken` isn't right for this. I think I added that marker initially so that contributors who opened PRs that exposed previously broken functionality weren't on the hook for fixing things way out of scope.

For these tests to go in, I think we want to first agree on what the "correct" (or just our expected) precision and format is for a timestamp.

Then, for those backends that support a custom format for timestamp outputs, those would be marked as `notimpl`, since it would be on us to add in handling for massaging the output into the desired format.

For those backends that don't offer a way to massage the output, those would be `notyet`.


The goal here, always, is to note which things require only effort from Ibis contributors (`notimpl`) vs. those which require upstream changes (`notyet`) vs. those which should be abandoned as folly (`never`).

ibis/backends/tests/test_generic.py
gforsyth: > As much as possible, we pull tests from the specific backends to be shared by all backends

I'm on board with moving tests from the backend-specific locations to the shared locations, especially because there IS test duplication there.



> When there is duplication, I would remove the backend-less tests, I would rather fully test for the full executed value instead of merely "does our expression system say the result is the correct type"

This not so sure of.  It might be a bit duplicative when things are working, or when things are failing at the backend, but we'd want the signal that we screwed up some bit of our internals vs. wacky backend behavior.

ibis/backends/tests/test_generic.py
NickCrews: tracking issue for timestamp -> string casting: #9534 

tracking issue for test mark adjustments: https://github.com/ibis-project/ibis/issues/9535

ibis/backends/tests/test_generic.py
NickCrews: Sounds good.

Does this PR look acceptable as it is, and can we do this test consolidation in a separate PR?

NickCrews: @gforsyth any thoughts here? would love to get this merged. Thank you!

gforsyth (approved):

PR #9484, opened 2024-06-30
fix(polars): add workaround to compile Array<Array> correctly

NickCrews:
Workaround for https://github.com/pola-rs/polars/issues/17294

pl.concat_list(Iterable[T]) results in pl.List[T], EXCEPT when T is a
pl.List, in which case pl.concat_list(Iterable[pl.List[T]]) results in pl.List[T].
If polars ever supports a more consistent array constructor,
we should switch to that.

Found this when working on https://github.com/ibis-project/ibis/pull/9473


gforsyth (approved):
Nice!  Thanks @NickCrews !

PR #9492, opened 2024-07-01
chore: patch scipy 1.14 for new usage of xcrun

cpcloud:
Fixes nixos mac builds for newer scipy versions that run xcrun.


cpcloud:
Fixing this upstream in poetry2nix in https://github.com/nix-community/poetry2nix/pull/1721.

PR #9508, opened 2024-07-03
test(clickhouse): add TPC-H and TPC-DS queries

cpcloud:
Adds TPC-H and TPC-DS testing to clickhouse.

Depends on #9525.

### DIFF
--- a/ibis/backends/clickhouse/tests/conftest.py
+++ b/ibis/backends/clickhouse/tests/conftest.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
+import concurrent.futures
 import contextlib
 import os
+import subprocess
 from typing import TYPE_CHECKING, Any
 
 import pytest
+import sqlglot as sg
 
 import ibis
 import ibis.expr.types as ir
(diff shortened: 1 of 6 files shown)


ibis/backends/clickhouse/compiler.py
```diff
@@ -705,3 +708,31 @@ def visit_TableUnnest(
             )
 
         return select
+
+    def _cleanup_names(
+        self, exprs: Mapping[str, sge.Expression]
+    ) -> Iterator[sge.Expression]:
+        """Compose `_gen_valid_name` and `_dedup_name` to clean up names in projections.
+
+        ClickHouse has a bug where this fails to find the final `"o"."a"` column:
```
jcrist, line 717: Is there a link for this bug? If so, might want to add it to the docstring here as a reference.

jcrist (approved):
Overall LGTM!

ibis/backends/clickhouse/compiler.py
cpcloud: I'll rebase, which should pick up the place where I worked around and linked this bug (#9525)

cpcloud: Yep, this should be in #9525.

PR #9510, opened 2024-07-03
feat(parse_sql): add support for sge.Cast and sge.DataType

gforsyth:
This makes parsing TPC-H query 1 break in a different place, so progress?


cpcloud:
Looks like we don't have a way to attach un-table-prefixed columns to their source table. This would likely be needed for most real-world scenarios as humans tend not to prefix columns with the table they come from when writing SQL.

cpcloud:
Should we close this out in favor of a deeper exploration of query 1 through the decompiler?

gforsyth:
Yep!

PR #9516, opened 2024-07-04
feat(polars): support version 1.0 and later

cpcloud:
Support Polars 1.0 and set lower bound to 1.0. There are many behavioral changes in Polars 1.0, such that it makes it difficult to justify supporting so many older versions. I say we cut losses and jump to 1.0 as the lower bound.


cpcloud:
Core dump reported upstream in https://github.com/pola-rs/polars/issues/17425.

chloeh13q:
p.s. probably want to update the conda env files too - they're not upper bounded right now and I was running into problems when I updated my environment, but since this is going in soon (I think?) maybe we can just directly bump it to 1.+ there too instead of having an intermediary change?

cpcloud:
@chloeh13q Yep, thanks!

gforsyth (approved):
:shipit:

PR #9518, opened 2024-07-04
refactor(viz): avoid repeatedly rendering redundant schemas in graphviz output

cpcloud:
Avoids rendering schemas for `Simple` relations in graphviz output, which are the relations that do not alter their schema in any way.


gforsyth:
A non-trivial query:

Before:
tmp65zr3_jm

After:
tmphnxtrvnj


This seems a bit cleaner to me?  Thoughts?

gforsyth (approved):

PR #9519, opened 2024-07-04
fix(datatypes): proper handling of srid in geospatial datatypes

### LINKED ISSUE #9465: bug: Loading geometry from postgres with SRID gives error
### What happened?

Very similar to #8807. When trying to load a table with a geometry that has an SRID set, I get the following error:

`TypeError: SqlglotType._from_sqlglot_GEOMETRY() takes from 1 to 2 positional arguments but 3 were given`

Here is a reproducible example:

```sql
drop table if exists v_test;
CREATE TABLE v_test (
    id int8 NULL,
    lonlat geometry(Point, 4326) NULL
);

INSERT INTO v_test (id, lonlat)
VALUES 
    (2, st_setsrid(st_makepoint(5.7, 53.2),4326));
```

```python
import ibis

pg_con = ibis.connect(
    "postgresql://blablabla",
)

pg_con.sql("select * from v_test").execute()
```

If it helps, loading this table with geopandas does not give errors.

ibis-framework[postgres,geospatial]==9.1.0

(Also tried --pre)

Postgres

### Relevant log output

_No response_

### Code of Conduct

ncclementi:
- Closes #9465

This is still broken, some how the srid value is not being passed properly to `_from_sqlglot_GEOMETRY`, I think it's because I need to change something in `from_ibis` too, but not sure how to handle passing the param.

I'm constructing well, the ibis_type in the test, I see 
```
pp ibis_type
GeoSpatial(geotype='geometry', srid=1234, nullable=True)
```

But when this is passed to `SqlglotType.from_ibis(ibis_type)` I'm missing somthing that I don't quite pin point yet. 

Any comments are welcome


cpcloud:
This isn't hitting the codepath because the round trip assertion first produces a SQLGlot type, but without the SRID, which is incorrect, and then converts that back to an Ibis type that doesn't have the SRID (because it wasn't produced).

cpcloud:
You have to also adjust the `_from_ibis_GeoSpatial` method to include the SRID in the SQLGlot output.

ncclementi:
I added the srid in the  `_from_ibis_GeoSpatial`  and that seems to be generating the right thing, but for the roundtrip, I'm not sure how to modify the `to_ibis` pass the srid properly to the method. As is, with the test example I'm working, I get
```
-> restored_dtype = SqlglotType.to_ibis(sqlglot_result)
(Pdb) pp sqlglot_result
DataType(
  this=Type.GEOMETRY,
  expressions=[
    DataTypeParam(
      this=Literal(this=1234, is_string=False))])
(Pdb) pp SqlglotType.to_ibis(sqlglot_result)
*** KeyError: '1234'
```

Clearly I'm missing something here, but not sure what the syntax is to be ale to pass the srid, 

```python
 if method := getattr(cls, f"_from_sqlglot_{typecode.name}", None):
            dtype = method(*typ.expressions)
 ```

**Comment by cpcloud** [comment]
Pushed up a few changes to get this working for both geometry and geography, was a bit gnarlier than I thought. The issue was that all the Point/MultiPoint etc types need to be handled separately so that they can be passed as a `DataTypeParam` argument when coming **from** Ibis, and then also that they are transformed to the right bits when going **to** Ibis.

File: ibis/backends/sql/tests/test_datatypes.py [hunk 1]
```diff
@@ -63,10 +64,16 @@ def test_roundtripable_types(ibis_type):
...
+    assert sqlglot_result.args["this"] == getattr(
+        sge.DataType.Type, ibis_type.geotype.upper()
     )
+    assert sqlglot_result.args["expressions"] == [
+        sge.Var(this=ibis_type.__class__.__name__.upper())
+    ]
+    assert SqlglotType.to_ibis(sqlglot_result) == ibis_type
+
+
+# def test_geotype_with_srid
```
ncclementi, line 76: whoops
```suggestion
```

ibis/tests/strategies.py
```diff
@@ -178,22 +178,28 @@ def struct_dtypes(
...
 def specific_geometry_dtypes(nullable=_nullable):
+    geotype = st.one_of(st.just("geography"), st.just("geometry"))
     return st.one_of(
-        st.builds(dt.Point, nullable=nullable),
-        st.builds(dt.LineString, nullable=nullable),
-        st.builds(dt.Polygon, nullable=nullable),
-        st.builds(dt.MultiPoint, nullable=nullable),
-        st.builds(dt.MultiLineString, nullable=nullable),
-        st.builds(dt.MultiPolygon, nullable=nullable),
+        st.builds(dt.Point, geotype=geotype, nullable=nullable),
```
ncclementi, line 197: do we need to add a cases with `srid` not `None` here? or the test with the general Geometry and geography dtype is enough?

cpcloud: Just added some, nice catch!

ibis/tests/strategies.py
```diff
@@ -178,34 +178,40 @@ def struct_dtypes(
...
 
 
-def geometry_dtypes(nullable=_nullable, srid=None):
+def geometry_dtypes(nullable=_nullable):
     return st.builds(
         dt.GeoSpatial,
         geotype=st.just("geometry"),
         nullable=nullable,
-        srid=st.just(srid),
+        srid=st.one_of(st.just(None), st.just(4321)),
```
ncclementi, line 186: Is this to build the two separate strategies? If so, pretty neat : )

ibis/tests/strategies.py
ncclementi: as I posted, you fixed this, nice!

ibis/tests/strategies.py
cpcloud: Yep!

cpcloud:
Looks like I broke a few things. Have an idea how to fix it, but checking out for the day, happy 4th!

cpcloud:
Gonna add the test case from the original issue here as well!

cpcloud (approved):
@ncclementi This LGTM if you want to give it another look!

ibis/backends/sql/datatypes.py
```diff
@@ -27,8 +27,6 @@
     typecode.ENUM16: dt.String,
     typecode.FLOAT: dt.Float32,
     typecode.FIXEDSTRING: dt.String,
-    typecode.GEOMETRY: partial(dt.GeoSpatial, geotype="geometry"),
```
ncclementi, line 30: Small question, we don't need these ones any more because we will always git this path
`method := getattr(cls, f"_from_sqlglot_{typecode.name}", None)` in `to_ibis()` now?

ncclementi (commented):
LGTM, thanks for taking this across the line, looks like it got a bit gnarly.

I left a comment with one question, but it's mostly for my understanding.

ibis/backends/sql/datatypes.py
cpcloud: No, they were never used prior to this PR as well.

cpcloud:
Thanks @ncclementi! Merging.

Issue #9486, opened 2024-06-30
feat: add `how="positional"` to `Table.join()`


NickCrews:
This came up in https://github.com/ibis-project/ibis/issues/9324.

Originally the proposal was for a ibis.hconcat() function that does this for us. But I think we can and should keep everything related to "merging two different relations" in a single place. IDK, are there other methods of merging/joining tables together?

duckdb has this built in. Pandas and dask also would be trivial. Other backends we can add a `ibis.row_number()` column to both relations, then joining on that.

IDK how to handle if the tables are different lengths:
- I would say, ideally, we should require them to be the same length.
- IDK where/how we should enforce that. I don't think it's possible during expression construction, I think we would have to do it at execution time. But then would we have to execute the join result and inspect it, before continue executing it? Not great. So maybe we can't enforce the same-length restriction.

cpcloud:
Thanks for the issue!

I think it makes sense to expose positional, but going out of our to implement it for backends that don't explicitly support it is probably out of scope.

PR #9533, opened 2024-07-06
feat(api): add positional joins

NickCrews:
closes https://github.com/ibis-project/ibis/issues/9486. 

Add positional join to duckdb, pandas, and dask backends.

Amazingly simple to implement, a big cheers to the previous developers that made this code so clean and easy to edit!

By luck or magic, the pandas and dask backends seem to just work with no modification!

Should this be documented somewhere? I hesitate to clutter up the examples with this simple example. Really, I think the better thing that ibis really needs is a how-to guide on joining, where someone *actually* goes deep into the weeds and explains how things work.

### DIFF
--- a/ibis/backends/tests/test_join.py
+++ b/ibis/backends/tests/test_join.py
@@ -385,3 +385,33 @@ def test_join_conflicting_columns(backend, con):
         }
     )
     backend.assert_frame_equal(result, expected)
+
+
+@pytest.mark.never(
+    [
+        "bigquery",
+        "clickhouse",
+        "datafusion",
+        "druid",
+        "exasol",
+        "flink",
+        "impala",
+        "mssql",
+        "mysql",
+        "oracle",
+        "polars",
+        "postgres",
+        "pyspark",
+        "risingwave",
+        "snowflake",
+        "sqlite",
+    ],
+    reason="Users can implement this with ibis.row_number(): https://github.com/ibis-project/ibis/issues/9486",
+)
+def test_positional_join(backend, con):
+    t1 = ibis.memtable({"x": [1, 2, 3]})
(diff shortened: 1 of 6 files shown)


cpcloud:
> By luck or magic, the pandas and dask backends seem to just work with no modification!

Hm, CI seems to think otherwise 😅

cpcloud:
@jcrist @gforsyth Opinions on opening up the join API to `how="positional"`?

ibis/backends/tests/test_join.py
```diff
@@ -385,3 +385,33 @@ def test_join_conflicting_columns(backend, con):
...
+        "clickhouse",
+        "datafusion",
+        "druid",
+        "exasol",
+        "flink",
+        "impala",
+        "mssql",
+        "mysql",
+        "oracle",
+        "polars",
```
gforsyth, line 402: It doesn't have to be a part of this PR, but `polars` almost certainly supports positional joins via `concat`, so this should probably be marked as a `notimpl`

gforsyth:
They're a thing in at least DuckDB and polars, so it seems ok to enable them to me.  I would push back against a top-level `positional_join` method, but I think adding it to `how` is alright.

ibis/backends/tests/test_join.py
cpcloud: Easy enough, I'll just implement it.

jcrist:
> They're a thing in at least DuckDB and polars, so it seems ok to enable them to me. I would push back against a top-level `positional_join` method, but I think adding it to `how` is alright.

Fully agree.

cpcloud (approved):

NickCrews:
Thank you all!

PR #9549, opened 2024-07-10
feat(geospatial): use geoarrow extension types when returning geometry columns as pyarrow

paleolimbot:
## Description of changes

Currently, GeoSpatial types are returned to pyarrow as `pa.binary()`, which drops the SRID and geometry/geography-ness of the column. Returning these columns marked as a GeoSpatial type allows other tools (e.g., lonboard for visualization) to pick this up.

I haven't kicked the tires on this at all yet and I almost certainly haven't thought through the implications here!


kylebarron:
This looks awesome! Along with https://github.com/ibis-project/ibis/pull/9143, with this PR you should be able to pass an `ibis.Table` directly into `lonboard.viz`, because the `viz` function will see that an `__arrow_c_stream__` dunder exists, then be able to interpret the data directly.

ibis/formats/pyarrow.py
```diff
@@ -161,7 +193,23 @@ def from_ibis(cls, dtype: dt.DataType) -> pa.DataType:
...
-            return pa.binary()
+            # Resolve CRS
+            if dtype.srid is None:
+                crs = None
+            elif dtype.srid == 4326:
+                crs = geoarrow_types.OGC_CRS84
+            else:
+                # Warn for dropped CRS? Or geoarrow.types would need a lookup table
+                # for srid -> PROJJSON
+                crs = None
```
kylebarron, line 204: Is pyproj a dependency? Is it safe to assume that an srid refers to an EPSG code?

ncclementi: Looks like it comes along, when we install `pip install 'ibis-framework[duckdb, geospatial]'`. 

```sh
Installing collected packages: pytz, tzdata, typing-extensions, toolz, sqlglot, six, pygments, 
pyarrow-hotfix, parsy,  packaging, numpy, mdurl, duckdb, click, certifi, bidict, attrs, atpublic,
shapely, python-dateutil, pyproj, pyarrow, markdown-it-py, cligj, click-plugins, rich, pandas,
 fiona, ibis-framework, geopandas
```

paleolimbot: Nice!

> Is it safe to assume that an srid refers to an EPSG code?

It depends where it comes from...a user typing it probably meant EPSG code, but otherwise it would be backend-specific (e.g., queryable using https://postgis.net/docs/using_postgis_dbmanagement.html#spatial_ref_sys_table ). It is usually an EPSG code in most practical use-cases (but could be a user-defined ID).

I'll workshop this a little and see what we can do!

cpcloud:
@paleolimbot Let us know when you're ready for a review!

cpcloud (approved):

pyproject.toml
```diff
@@ -173,7 +174,7 @@ visualization = ["graphviz"]
 decompiler = ["black"]
 deltalake = ["deltalake"]
 examples = ["pins", "fsspec"]
-geospatial = ["geopandas", "shapely"]
+geospatial = ["geoarrow-types", "geopandas", "shapely"]
```
cpcloud, line 177: Since we're importing pyproj directly, we need to add it as an optional dependency for the geospatial extra, I can take care of that.

cpcloud: Done!

cpcloud (approved):
Excellent!

pyproject.toml
```diff
@@ -64,6 +64,8 @@ db-dtypes = { version = ">=0.3,<2", optional = true }
 deltalake = { version = ">=0.9.0,<1", optional = true }
 duckdb = { version = ">=0.8.1,<2", optional = true }
 geopandas = { version = ">=0.6,<2", optional = true }
+geoarrow-types = { version = ">=0.2,<1", optional = true }
+pyproj = { version = ">=3.3.0,<4", optional = true }
```
ncclementi, line 68: Just to be sure, we are making `pyproj` an explicit dependency. I know it comes when installing geopandas, which is already a dependency. I guess explicit better than implicit in this case.

cpcloud: Yeah, it's generally a bad idea to assume that a package will forever and always include a dependency that you require. If there's a top-level import for a package, it needs to be included as an explicitly dependency.

PR #9554, opened 2024-07-11
feat(duckdb): speedup to_pandas/execute

lostmygithubaccount:
## Description of changes

use `duckdb_table.df()` directly instead of going through Apache Arrow to pandas for significant speedup of `table.execute()`/`table.to_pandas()`

## Issues closed

zulip topic: https://ibis-project.zulipchat.com/#narrow/stream/405265-tech-support/topic/ibis.2Educkdb.20over.20300X.20slower.20than.20raw.20duckdb


lostmygithubaccount:
not sure why so many tests are failing

robdmc:
I wanted to mention here that adopting this PR may alter the way timezones are handled in the returned dataframe.  See the zulip topic linked above.

cpcloud:
This is the right approach to the issue here, which is that apparently passing `timestamp_as_object=True` to arrow array's `to_pandas` method is much slower than passing `timestamp_as_object=False`.

cpcloud:
Closing in favor of #9556.

PR #9555, opened 2024-07-11
docs: add a note on docs preview

chloeh13q:
## Description of changes

add a note on docs preview

### DIFF
--- a/docs/contribute/01_environment.qmd
+++ b/docs/contribute/01_environment.qmd
@@ -221,6 +221,15 @@ just docs-preview
 
 to build and serve the documentation.
 
+::: {.callout-tip}
+If you're getting reference errors (e.g., `reference/operations.qmd` not found), you may need to run
+```bash
+just docs-api-preview
+```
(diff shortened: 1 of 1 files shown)


chloeh13q:
@lostmygithubaccount There's `just docs-api-preview` and `just docs-preview` and `just docs-api-preview` actually does already do both (which I think is what you meant by updating the justfile instead so that `just docs-preview` does it for API docs too). I'm guessing that we kept both because `just docs-preview` is slightly cheaper and faster... But we could also just not have `just docs-preview`...

ncclementi:
I run into this couple of times, I think it's useful to have it on the docs, and I agree that if you are not touching API docs, and you want to render the docs, I wouldn't want to have to build the API docs all the time. 

 But I wonder if there is a way of adding it as a Warning right before the just docs-preview is run, 

I'm not sure what's the right syntax. 

But in here, running something like `echo "Warning: make sure to run the API docs `just docs-api-preview` or you will get a reference error"`

PR #9556, opened 2024-07-12
perf(duckdb): speedup timestamp conversion by avoiding conversion to object

cpcloud:
Redo of #9554. The slow bit here was object conversion of timestamps, which seemed to only be hit by this one test.

### DIFF
--- a/ibis/formats/pandas.py
+++ b/ibis/formats/pandas.py
@@ -200,8 +200,10 @@ class PandasData(DataMapper):
 
     @classmethod
     def convert_Timestamp(cls, s, dtype, pandas_type):
-        if isinstance(dtype, pd.DatetimeTZDtype):
-            return s.dt.tz_convert(dtype.timezone)
+        if isinstance(pandas_type, pd.DatetimeTZDtype) and isinstance(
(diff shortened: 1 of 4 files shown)


cpcloud:
Still about 5x left that we're losing, but that can be done in a follow up.

cpcloud:
Ok found a few of the remaining Xs in a bug 🐞

ibis/formats/pandas.py
```diff
@@ -200,8 +200,10 @@ def convert_Boolean(cls, s, dtype, pandas_type):
 
     @classmethod
     def convert_Timestamp(cls, s, dtype, pandas_type):
-        if isinstance(dtype, pd.DatetimeTZDtype):
```
cpcloud, line 203: This check was always `False`, because `dtype` is always an Ibis type.

gforsyth (approved):
This looks good to me, and I think the shift from `ns` to `us` precision is a non-issue here, since DuckDB doesn't support `ns` timestamps in the first-place, so it's mostly a very small change in the representation of the results, but the values aren't fundamentally changing.

PR #9559, opened 2024-07-12
fix(case): fix dshape, error on noncomparable and empty cases

NickCrews:
This is pinning down the expected behavior for cases before tackling the case() to cases() switch in https://github.com/ibis-project/ibis/pull/9096 so that PR can be simpler

### DIFF
--- a/ibis/tests/expr/test_case.py
+++ b/ibis/tests/expr/test_case.py
@@ -215,15 +215,11 @@ def test_case_mixed_type():
     assert result["label"].type().equals(dt.string)
 
 
-def test_err_on_nonbool(table):
+def test_err_on_nonbool_expr(table):
     with pytest.raises(SignatureValidationError):
         ibis.case().when(table.a, "bar").else_("baz").end()
-
-
-@pytest.mark.xfail(reason="Literal('foo', type=bool), should error, but doesn't")
-def test_err_on_nonbool2():
     with pytest.raises(SignatureValidationError):
-        ibis.case().when("foo", "bar").else_("baz").end()
+        ibis.case().when(ibis.literal(1), "bar").else_("baz").end()
 
 
 def test_err_on_noncomparable(table):
--- a/ibis/tests/expr/test_case.py
+++ b/ibis/tests/expr/test_case.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
+import pytest
+
 import ibis
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
 from ibis import _
+from ibis.common.annotations import SignatureValidationError
 from ibis.tests.util import assert_equal, assert_pickle_roundtrip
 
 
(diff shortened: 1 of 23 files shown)


ibis/tests/expr/test_case.py
```diff
@@ -162,8 +200,6 @@ def test_multiple_case_null_else(table):
     op = expr.op()
     assert isinstance(expr, ir.StringColumn)
     assert isinstance(op.default.to_expr(), ir.Value)
-    assert isinstance(op.default, ops.Cast)
```
NickCrews, line 165: this is an implementation detail we shouldn't test

cpcloud:
Ok, I think this is pretty close. I'm going to take it over and hopefully merge it.

ibis/tests/expr/test_case.py
```diff
@@ -177,3 +213,31 @@ def test_case_mixed_type():
...
+
+def test_err_on_nonbool(table):
+    with pytest.raises(SignatureValidationError):
+        ibis.case().when(table.a, "bar").else_("baz").end()
+
+
+@pytest.mark.xfail(reason="Literal('foo', type=bool), should error, but doesn't")
+def test_err_on_nonbool2():
+    with pytest.raises(SignatureValidationError):
+        ibis.case().when("foo", "bar").else_("baz").end()
```
cpcloud, line 226: I'm going to try to fix this issue in this PR.

cpcloud: This is slightly weird, because we allow conversion from `int` to `bool` during `literal` input normalization, which does match the Python behavior of e.g., `bool(1)`. This would be a breaking change, and we have a test (`test_ifelse_function_literals`) for this so I'm going to delete the test in this PR, as this behavior is intentional, there's even a comment in the test, specifically about this behavior!

cpcloud: FTR this is the kind of hairy problem that we should not just start writing a bunch of tests for without exploring the issue more. It's a significant behavior change.

cpcloud (approved):

ibis/tests/expr/test_case.py
NickCrews: I think this would be my ideal behavior:
- ibis.case().when("foo") -> error
- ibis.case().when("True") -> error
- ibis.case().when(1) -> error
- ibis.case().when(True) -> works
- ibis.case().when(ibis.literal("True", type=bool)) -> works

in general, I think we should follow the guidelines of

- If handed an ibis Expr, it must be .is_boolean()
- If handed anything else, we pass it to ibis.literal(x) (without explicitly saying `type=bool`!) and the result had better be .is_boolean(), else error.

Is this the behavior you would want? if so, we should adjust the other test to match

cpcloud: I think that behavior has more cases to understand and handle than the current behavior, which is fairly easy to explain:

1. Not an ibis expression -> call the `bool` builtin on the input
2. Ibis expression -> check that it's `.type()` is boolean.

Having a third case of "call `bool` on the input only if the input type is specified is **a lot** of work to implement, including a big literal refactoring to differentiate between a type that was passed and one that wasn't with very little benefit IMO.

Also, just to make sure we're on the same page:

```
In [1]: bool("True") == bool("False")
Out[1]: True
```

Ibis tries to follow Python behavior where it makes sense.

NickCrews: > 1. Not an ibis expression -> call the bool builtin on the input
> 2. Ibis expression -> check that it's .type() is boolean.

Nice, yes this is better than what I proposed. But it still does have a bit of a footgun, I don't love how `ibis.case().when("False", ...)` does opposite of what you might expect. Could we adjust the algorithm to

1. Not an ibis expression -> `assert isinstance(arg, bool)`, then do `arg = ibis.literal(arg)`
2. Ibis expression -> check that it's .type() is boolean.

It's a little more restrictive (users might need to do `case.where(bool(x))` where before they just had `case.where(x)`) and slightly breaking, but I think this might also catch some real bugs in people's code.

> Also, just to make sure we're on the same page:

Thanks, I was NOT on the same page, I mistakenly thought `bool("False") == False` and `bool("foo")` errored.

cpcloud: I don't have much to back up this claim, but we haven't gotten any bug reports about this yet, but we can certainly have a discussion when they come up. I'd rather not incur a bunch of development overhead for something that might happen.

NickCrews: sounds good!

PR #9560, opened 2024-07-12
chore(test): consolidate ifelse() and case() tests

NickCrews:
Prepping for https://github.com/ibis-project/ibis/pull/9096

- moves several .ifelse() and .substitute() tests from ibis/backens/test/test_generic.py into a new test_conditionals.py
- moves some .case() tests from the pandas and dask backends into there, so the other backends are now tested as well, and removed duplication.
- adds a test case that `ibis.literal(5).nullif(5).case().when(None, "oops").else_("expected")` results in "expected". Not saying this is actually the best behavior,
but it is currently what we claim to be expected. Perhaps we want to revisit.

### DIFF
--- a/ibis/backends/dask/tests/test_operations.py
+++ b/ibis/backends/dask/tests/test_operations.py
@@ -773,64 +773,6 @@ def test_quantile_group_by(batting, batting_pandas_df):
     tm.assert_series_equal(result, expected, check_index=False)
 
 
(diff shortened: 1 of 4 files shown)


cpcloud (approved):
LGTM!

cpcloud:
`None` being treated as `False` is consistent with how `NULL`s behave in other expressions where this comes up like joins.

PR #9552, opened 2024-07-11
refactor(deps): make pyarrow optional for non-backend installs

cpcloud:
Make pyarrow an optional dependency for installs that do not require a backend.

PR #9564, opened 2024-07-13
optional numpy and pandas

cpcloud:
Attempt to make pandas and numpy optional dependencies for installs that do not require a backend. Depends on #9552.


cpcloud:
cc @jonmmease

After this PR, you should be able to get an Ibis installation with none of the heavier dependencies, resulting in an install size of less than 30 MB.

jonmmease:
Very cool! What features will work with the minimal dependencies with this PR?

Will it be possible to perform sql generation for dialects (like BigQuery) without installing the execution dependencies?

cpcloud:
> Will it be possible to perform sql generation for dialects (like BigQuery) without installing > the execution dependencies?

Not yet :)

But that's the next step, just need to move the compilers out of the individual backends.

gforsyth (approved):
I'm not opposed to this change, and I think it's useful to have a tiny core package that people can install for non-execution purposes.

ibis/backends/bigquery/client.py
```diff
@@ -70,6 +69,8 @@ def bq_param_array(dtype: dt.Array, value, name):
 @bigquery_param.register
 def bq_param_timestamp(_: dt.Timestamp, value, name):
     # TODO(phillipc): Not sure if this is the correct way to do this.
+    import pandas as pd
```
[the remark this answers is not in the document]
gforsyth: oh, oops -- didn't remember that I left this comment 3 weeks ago.  I stand by it, but it's already been changed.

gforsyth:
I'm going to leave this open in case @jcrist wants to take a pass

ibis/formats/pandas.py
```diff
@@ -21,23 +16,19 @@
...
-        "Install pandas >= 1.5.0 for interop with pandas and arrow dtype support"
-    )
-
 geospatial_supported = _find_spec("geopandas") is not None
 
 
 class PandasType(NumpyType):
     @classmethod
     def to_ibis(cls, typ, nullable=True):
+        import pandas as pd
```
kszucs, line 29: Importing pandas globally in this file would be reasonable so that `ibis.formats.pandas` actually depends on pandas. Same applies to `ibis.formats.numpy`.

cpcloud: Yep, thanks for the poke! I was able to move these imports up without breaking anything and still allowing no pyarrow/pandas/numpy installation.

ibis/formats/numpy.py
```diff
@@ -1,37 +1,51 @@
...
-_to_numpy_types = {v: k for k, v in _from_numpy_types.items()}
-
-
-class NumpyType(TypeMapper[np.dtype]):
+if TYPE_CHECKING:
+    import numpy as np
+
+
+@functools.cache
+def _from_numpy_types():
```
cpcloud, line 17: Gonna move these back to a pure `dict` impl.

cpcloud:
@jcrist Could I get your review on this?

ibis/backends/sqlite/__init__.py
```diff
@@ -19,7 +19,6 @@
 from ibis.backends import UrlFromPath
 from ibis.backends.sql import SQLBackend
 from ibis.backends.sql.compilers.base import C, F
-from ibis.backends.sqlite.converter import SQLitePandasData
```
jcrist, line 22: Are changes like this still needed? IIUC

- Backends will still require pandas/pyarrow/numpy to work properly
- `ibis.to_sql` and expressions should work without these deps, but that only requires importing the compilers, not the backend itself

If possible, I'd like to avoid imports from within methods if they can remain top-level.

jcrist (approved):

ibis/backends/sqlite/__init__.py
cpcloud: I think it worked. Let's see what CI says.

PR #9575, opened 2024-07-13
ci(trino): pin to 449 to avoid truncate table memory bug

cpcloud:
Pin trino for now until the truncate table statement for in-memory connector can be fixed. See https://github.com/trinodb/trino/pull/22337#issuecomment-2227135370 for a reproducer.


gforsyth (approved):
0 + 3 == 6
:shipit:

PR #9576, opened 2024-07-13
feat(dtypes): fall back to `dt.unknown` for unknown types

gforsyth:
We've had a number of issues where we haven't exposed a mapping to a
particular (usually Postgres) dtype.

We'll still have requests to provide useful mappings, depending on the
upstream dtype, but this should prevent users being unable to even open
a table because there's an unknown / user-defined dtype present in the
table.

To test, I've removed the special handling for the `vector` and
`ts.vector` types that we were explicitly mapping to `dt.unknown`.

Trying this one again...


cpcloud (approved):
Thanks!

PR #9590, opened 2024-07-15
refactor(compilers): move compilers out of the backend dependency path

cpcloud:
Moves the compilers out of the path of backend-specific dependencies, to allow using them with a base installation of Ibis. cc @jonmmease


gforsyth:
Ahh, looks like we just need to update this import:

```python
from ibis.backends.sql.compiler import ALL_OPERATIONS
```

in the support matrix generation script

ibis/backends/sql/compilers/datafusion.py
```diff
@@ -136,6 +135,8 @@ def visit_Cast(self, op, *, arg, to):
         if to.is_timestamp():
             return self._to_timestamp(arg, to)
         if to.is_decimal():
+            from ibis.formats.pyarrow import PyArrowType
+
             return self.f.arrow_cast(arg, f"{PyArrowType.from_ibis(to)}".capitalize())
```
cpcloud, line 140: This isn't 100% free of dependencies here, because we're using the pyarrow type system for casts, because datafusion doesn't have any other way to represent types than their pyarrow string representation. So, unfortunately, if you're converting from non-datafusion dialects to datafusion then you'll need pyarrow installed.

gforsyth: I think this is a big improvement and this will only trip up users if they are going from non-datafusion dialects to datafusion AND if they have decimal columns.  :shipit:

cpcloud: Yep, it's a somewhat specific use case I think.

PR #9592, opened 2024-07-15
docs: add `deferred`, `range`, `e`, and `pi` to API reference

tswast:
## Description of changes

Some top-level APIs were missing from the API reference. This adds them.

### DIFF
--- a/docs/presentations-listing.json
+++ b/docs/presentations-listing.json
@@ -1,8 +1,0 @@
-{
-  "listing": "/presentations.html",
-  "items": [
-    "/presentations/pycon2024/maintainers.html",
-    "/presentations/linkedin-meetup-2024-04-24.html",
-    "/presentations/overview.html"
-  ]
-}
--- a/docs/presentations-listing.json
+++ b/docs/presentations-listing.json
@@ -1,0 +1,8 @@
+{
+  "listing": "/presentations.html",
+  "items": [
+    "/presentations/pycon2024/maintainers.html",
+    "/presentations/linkedin-meetup-2024-04-24.html",
+    "/presentations/overview.html"
+  ]
+}
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -2094,8 +2094,8 @@ def range(start, stop, step) -> ir.ArrayValue:
     Integer ranges are supported, as well as timestamp ranges.
 
     ::: {.callout-note}
-    `start` is inclucive and `stop` is exclusive, just like Python's builtin
-    [`range`](range).
+    `start` is inclusive and `stop` is exclusive, just like Python's builtin
+    [`range`](:py:class:`range`).
 
     When `step` equals 0, however, this function will return an empty array.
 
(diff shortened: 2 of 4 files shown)


docs/presentations-listing.json
cpcloud, line 1: Is this file needed/related to the PR's motivation?

tswast: No, `just docs-preview` seems to have created that. I can delete.

tswast:
Re: `file:///home/runner/work/ibis/ibis/docs/_output/reference/range | Failed: Cannot find file`. I'll try some fixes. Adding a more specific package name seemed to help for the others.

cpcloud:
That means that there's a file referenced somewhere that doesn't exist. That usually means that wherever you're referencing `range`, the way you've referenced it is incorrect.

cpcloud:
Ah, looks like you're not referencing `range` anywhere, so that may not be the issue.

gforsyth:
Looks like the docstring for `ibis.range` has an invalid link -- let me poke it

gforsyth:
I was wrong about interlinks, we do have them, so it may just be how we're spelling it.  Poking locally, still

tswast:
Thanks @gforsyth I see you're updating the PR.

Looks like https://docs.python.org/3/library/functions.html#func-range would be the right link, but yeah if interlinks work it would be good to reference the built-in that way.

gforsyth:
ok, that took way too long to figure out, but it's rendering correctly for me now

ibis/expr/api.py
```diff
@@ -2094,8 +2094,8 @@ def range(start, stop, step) -> ir.ArrayValue:
     Integer ranges are supported, as well as timestamp ranges.
 
     ::: {.callout-note}
-    `start` is inclucive and `stop` is exclusive, just like Python's builtin
-    [`range`](range).
+    `start` is inclusive and `stop` is exclusive, just like Python's builtin
+    [`range`](:py:class:`range`).
```
cpcloud, line 2098: The original is correct here. You can't use Sphinx syntax in quartodoc.

cpcloud, line 2098:
```suggestion
    [](`range`).
```

I believe this should fix the reference error.

cpcloud: Ah, didn't see your comment in the top-level PR.

I think we should probably avoid the Sphinx syntax since it kind of gives the impression that people should write Sphinx.

Admittedly, it's probably unlikely that someone will come across this and then additionally conclude that Sphinx is in use, but if there's a way we can do it that matches everything else like it (referencing a builtin object) then I think we should.

gforsyth: I swear I tried the suggested code before and it didn't render, but I went back to confirm and it works.

Interlinks on quartodoc do support the sphinx syntax, just in case we ever need it for specifying an ambiguous link, but I'll switch this one to use the simpler syntax.

cpcloud: > Interlinks on quartodoc do support the sphinx syntax

😮

cpcloud (approved):

PR #9593, opened 2024-07-15
chore(justfile): add check to justfile docs preview commands

gforsyth:
the `quartodoc` output lives in `docs/reference`, so if that folder
doesn't exist, or if it is empty, we can run the `docs-apigen` justfile
target to make sure the build succeeds.

There will definitely still be failure modes in here, but this should
cover the most common ones.

xref #9555

### DIFF
--- a/justfile
+++ b/justfile
@@ -171,10 +171,26 @@ docs-apigen *args:
 
 # build documentation
 docs-render:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    # Check if the folder "reference" exists and has contents
+    if [ ! -d "docs/reference" ] || [ -z "$(ls -A docs/reference)" ]; then
+        just docs-apigen
+    fi
+
     quarto render docs
 
 # preview docs
 docs-preview:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    # Check if the folder "reference" exists and has contents
+    if [ ! -d "docs/reference" ] || [ -z "$(ls -A docs/reference)" ]; then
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -171,10 +171,26 @@ docs-apigen *args:
 
 # build documentation
 docs-render:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    # Check if the folder "reference" exists and has contents
+    if [ ! -d "docs/reference" ] || [ -z "$(ls -A docs/reference)" ]; then
+        `just docs-apigen`
```
cpcloud, line 179:
```suggestion
        just docs-apigen
```

justfile
```diff
@@ -171,10 +171,26 @@ docs-apigen *args:
...
     quarto render docs
 
 # preview docs
 docs-preview:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    # Check if the folder "reference" exists and has contents
+    if [ ! -d "docs/reference" ] || [ -z "$(ls -A docs/reference)" ]; then
+        `just docs-apigen`
```
cpcloud, line 191:
```suggestion
        just docs-apigen
```

cpcloud (changes requested):
I think we just want to run the script, not run it and then run its output as a command.

Other than that, looks good!

justfile
gforsyth: lol, dunno what I had in mind here...

cpcloud (approved):

PR #9596, opened 2024-07-15
docs(arbitrary): add example for `arbitrary` method docstring

cpcloud:
Add an example to the `arbitrary` method docstring.


jcrist (approved):
Since these are arb

PR #9603, opened 2024-07-16
feat(backends): support creation from a DB-API con

### LINKED ISSUE #8877: feat: Reuse the existing DB connection while creating backend

Currently all db operations are done thru SQLAlchemy/SQLGlot, thus only DBAPI compliant interfaces are exposed. It is required to reuse an existing db connection, e.g., duckdb.connec("mydb.db"), so that all previously registered python function (via duckdb.create_function) to work, also, it would be easier to guarantee the in-memory duckdb to be shared among ibis and non-ibis codes

Change the current connect method to accept an existing connection

9.0.0.dev568

DuckDB, Polars, MySQL

### Code of Conduct

deepyaman:
## Description of changes

Enable backend creation from an existing DB-API connection

## Issues closed

Resolves #8877

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -767,6 +767,7 @@ class BaseBackend(abc.ABC, _FileIOHandler):
     def __init__(self, *args, **kwargs):
         self._con_args: tuple[Any] = args
         self._con_kwargs: dict[str, Any] = kwargs
+        self._can_reconnect: bool = True
         # expression cache
         self._query_cache = RefCountedCache(
             populate=self._load_into_cache,
@@ -856,7 +857,10 @@ class BaseBackend(abc.ABC, _FileIOHandler):
     # TODO(kszucs): should call self.connect(*self._con_args, **self._con_kwargs)
     def reconnect(self) -> None:
         """Reconnect to the database already configured with connect."""
-        self.do_connect(*self._con_args, **self._con_kwargs)
+        if self._can_reconnect:
+            self.do_connect(*self._con_args, **self._con_kwargs)
+        else:
+            raise exc.IbisError("Cannot reconnect to unconfigured {self.name} backend")
 
     def do_connect(self, *args, **kwargs) -> None:
         """Connect to database specified by `args` and `kwargs`."""
--- a/ibis/backends/bigquery/__init__.py
+++ b/ibis/backends/bigquery/__init__.py
@@ -478,6 +478,33 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema):
 
         self.partition_column = partition_column
 
+    @classmethod
+    def from_connection(
+        cls,
+        client: bq.Client,
+        partition_column: str | None = "PARTITIONTIME",
+        storage_client: bqstorage.BigQueryReadClient | None = None,
+    ) -> Backend:
+        """Create a BigQuery `Backend` from an existing ``Client``.
+
+        Parameters
+        ----------
+        client
+            A ``Client`` from the ``google.cloud.bigquery`` package.
+        partition_column
+            Identifier to use instead of default ``_PARTITIONTIME`` partition
+            column. Defaults to ``'PARTITIONTIME'``.
+        storage_client
+            A ``BigQueryReadClient`` from the
+            ``google.cloud.bigquery_storage_v1`` package. If not set, one is
+            created using the ``project_id`` and ``credentials``.
+        """
+        return ibis.bigquery.Backend.connect(
+            client=client,
+            partition_column=partition_column,
+            storage_client=storage_client,
+        )
+
     def disconnect(self) -> None:
         self.client.close()
 
--- a/ibis/backends/bigquery/__init__.py
+++ b/ibis/backends/bigquery/__init__.py
@@ -478,6 +478,7 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema):
 
         self.partition_column = partition_column
 
+    @util.experimental
     @classmethod
     def from_connection(
         cls,
--- a/ibis/backends/bigquery/__init__.py
+++ b/ibis/backends/bigquery/__init__.py
@@ -485,25 +485,28 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema):
         client: bq.Client,
         partition_column: str | None = "PARTITIONTIME",
         storage_client: bqstorage.BigQueryReadClient | None = None,
+        dataset_id: str = "",
     ) -> Backend:
         """Create a BigQuery `Backend` from an existing ``Client``.
 
         Parameters
         ----------
         client
-            A ``Client`` from the ``google.cloud.bigquery`` package.
+            A `Client` from the `google.cloud.bigquery` package.
         partition_column
-            Identifier to use instead of default ``_PARTITIONTIME`` partition
-            column. Defaults to ``'PARTITIONTIME'``.
+            Identifier to use instead of default `_PARTITIONTIME` partition
+            column. Defaults to `'PARTITIONTIME'`.
         storage_client
-            A ``BigQueryReadClient`` from the
-            ``google.cloud.bigquery_storage_v1`` package. If not set, one is
-            created using the ``project_id`` and ``credentials``.
+            A `BigQueryReadClient` from the `google.cloud.bigquery_storage_v1`
+            package.
+        dataset_id
+            A dataset id that lives inside of the project attached to `client`.
         """
-        return ibis.bigquery.Backend.connect(
+        return ibis.bigquery.connect(
             client=client,
             partition_column=partition_column,
             storage_client=storage_client,
+            dataset_id=dataset_id,
         )
 
     def disconnect(self) -> None:
--- a/ibis/backends/clickhouse/__init__.py
+++ b/ibis/backends/clickhouse/__init__.py
@@ -163,6 +163,7 @@ class Backend(SQLBackend, CanCreateDatabase):
             **kwargs,
         )
 
+    @util.experimental
     @classmethod
     def from_connection(cls, con: cc.driver.Client) -> Backend:
         """Create an Ibis client from an existing ClickHouse Connect Client instance.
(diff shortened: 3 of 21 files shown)


ibis/backends/sql/__init__.py
```diff
@@ -547,6 +547,21 @@ def truncate_table(
         with self._safe_raw_sql(f"TRUNCATE TABLE {ident}"):
             pass
 
+    @classmethod
+    def from_dbapi_connection(cls, con: Any, **kwargs: Any) -> BaseBackend:
```
cpcloud, line 551: How about `from_connection` or `from_conn` (slight preference for the latter) so that backends that don't implement the DBAPI, like PySpark, don't have a method with a name that doesn't make sense for their implementation.

ibis/backends/tests/test_client.py
```diff
@@ -1764,3 +1764,8 @@ def test_insert_using_col_name_not_position(con, first_row, second_row, monkeypa
     # Ideally we'd use a temp table for this test, but several backends don't
     # support them and it's nice to know that data are being inserted correctly.
     con.drop_table(table_name)
+
+
+def test_from_dbapi_connection(con):
```
cpcloud, line 1769: Ideally we can run the entire test suite using a backend constructed this way, which will give us more confidence that the API works.

That can be done in a follow up though.

ibis/backends/__init__.py
```diff
@@ -767,6 +767,7 @@ class BaseBackend(abc.ABC, _FileIOHandler):
     def __init__(self, *args, **kwargs):
         self._con_args: tuple[Any] = args
         self._con_kwargs: dict[str, Any] = kwargs
+        self.can_reconnect: bool = True
```
jcrist, line 770: Would you mind making this private? I think we should be more intentional with what's a public attribute on a backend connection, and this feels more like an implementation detail to me. Can always expose it later if needed, but it's harder to take something away.

ibis/backends/sql/__init__.py
jcrist: I have a slight preference for the former, since I anticipate forgetting if it's `from_conn` or `from_con` (especially since `con = ...` is the common naming).

jcrist (commented):
It'd be nice to make this available as `ibis.<backend>.from_connection` as well. The code that does this is gross, but lives [here](https://github.com/ibis-project/ibis/blob/946f761d2e0351352b344afb0780c4c6ba7ac322/ibis/__init__.py#L118-L127). Could add it with a `hasattr` check here, or with adding `from_connection` to `_top_level_methods` on base SQL backend (and the snowflake backend, since it appends). No strong thoughts.

ibis/backends/sql/__init__.py
cpcloud: Fair enough, let's go with the former then!

ibis/backends/tests/test_client.py
deepyaman: How do you envision doing this without running the whole test suite again? Or it's fine to run it again?

ibis/backends/clickhouse/__init__.py
```diff
@@ -163,6 +163,20 @@ def do_connect(
             **kwargs,
         )
 
+    @classmethod
+    def from_dbapi_connection(cls, con: cc.driver.Client) -> Backend:
```
deepyaman, line 167: The latest request from this functionality comes from @sh-rp and the dltHub team: https://github.com/ibis-project/ibis/issues/8877#issuecomment-2180759817

However, for a case like ClickHouse, I'm not 100% sure we can bridge the gap, because dltHub uses `clickhouse-driver` (community package, longer history, potentially better at scale?) and Ibis uses `clickhouse-connect` (official ClickHouse package, better HTTP support). From https://github.com/ClickHouse/clickhouse-connect/issues/91, it looks like there was a lot of consideration around unification, but, in the end, doesn't seem likely.

cpcloud: Scale seems like a non-issue (e.g., https://github.com/ClickHouse/clickhouse-connect/issues/91#issuecomment-1423343917).

We switched to `clickhouse-connect` due to a bunch of data type issues and choices that didn't align well with Ibis (ClickHouse enums were completely broken due to the refusal to play nice with Python's built in `Enum` types), and at the time the maintainer was unresponsive and hadn't made a release in many months.

We're definitely not prepared to support multiple ways to connect to ClickHouse right now and it's unlikely we will do so in the future.

ibis/backends/tests/test_client.py
cpcloud: You can (somewhat arbitrarily) pick a backend and modify its `conftest.py` to use this way of initialization.

ibis/backends/__init__.py
deepyaman: Done.

ibis/backends/sql/__init__.py
deepyaman: Done.

ibis/backends/clickhouse/__init__.py
deepyaman: Makes sense! I just wanted to note this, and appreciate the explanation of why Ibis moved to `clickhouse-connect`.

ibis/backends/tests/test_client.py
deepyaman: But then we wouldn't have the other way of initialization, right? Or, because there's `ddl_con` and `con` fixtures both, it would be OK?

Will look to explore this in a followup then; will just get the rest of the backends added and tests passing here first.

deepyaman:
Oh, I forgot about BigQuery and Snowflake...

deepyaman:
> It'd be nice to make this available as `ibis.<backend>.from_connection` as well. The code that does this is gross, but lives [here](https://github.com/ibis-project/ibis/blob/946f761d2e0351352b344afb0780c4c6ba7ac322/ibis/__init__.py#L118-L127). Could add it with a `hasattr` check here, or with adding `from_connection` to `_top_level_methods` on base SQL backend (and the snowflake backend, since it appends). No strong thoughts.

Done!

ibis/backends/duckdb/__init__.py
```diff
@@ -473,6 +473,30 @@ def do_connect(
 
         self.con = duckdb.connect(str(database), config=config, read_only=read_only)
 
+        self._post_connect(extensions)
+
+    @classmethod
+    def from_connection(
+        cls,
+        con: duckdb.DuckDBPyConnection,
+        extensions: Sequence[str] | None = None,
```
cpcloud, line 482: Is there any reason for us to accept partial arguments like this? What's the issue with every `from_connection` implementation accepting **only** the connection/context/whatever object?

cpcloud (changes requested):
A couple questions! Looking really good!

ibis/backends/duckdb/__init__.py
deepyaman: My goal was to enable as much of the options you may have when connecting using `connect` as possible. The additional arguments are always optional, though, and should be as though you just passed the `con` object.

PySpark is the one case where I think you have to at least offer `mode`, as you can't infer that in any way from the `SparkSession`.

cpcloud: Hm, okay.

Can we remove all optional arguments that aren't strictly required?

cpcloud: Longer term we should really try to avoid having arguments like `mode` that we use that can't be determined from the backend itself.

cpcloud: Actually, let's leave this as is, but slap `@experimental` on all the `from_connection` implementations.

deepyaman: I was looking at it in the meantime. For posterity:

* I think I could remove additional arguments from DuckDB and Exasol
* PySpark and BigQuery have some things that can't be gotten from the connection, as mentioned
* I think SQLite `_type_map` also can't be gotten from an existing connection
* The existing `from_snowpark` supports `create_object_udfs` argument already, so that's kept

> Actually, let's leave this as is, but slap `@experimental` on all the `from_connection` implementations.
Sounds good!

deepyaman: Done slapping `@experimental` everywhere.

cpcloud (approved):
Excellent work!

cpcloud:
Gonna run BQ and Snowflake locally real quick...

cpcloud:
Found a small in BQ, fixing it now.

cpcloud:
Tested the clouds locally, but running here for completeness. Will merge on green and then cut 9.2!

ibis/backends/bigquery/__init__.py
```diff
@@ -485,26 +485,31 @@ def from_connection(
...
+            A dataset id that lives inside of the project attached to `client`.
         """
-        return ibis.bigquery.Backend.connect(
+        con = ibis.bigquery.connect(
             client=client,
             partition_column=partition_column,
             storage_client=storage_client,
+            dataset_id=dataset_id,
         )
+        con._can_reconnect = False
```
deepyaman, line 511: @cpcloud Is this correct? This would mean that you also can't reconnect from a regular `ibis.bigquery.connect` call.

ibis/backends/bigquery/__init__.py
```diff
@@ -485,26 +485,31 @@ def from_connection(
...
         partition_column
-            Identifier to use instead of default ``_PARTITIONTIME`` partition
-            column. Defaults to ``'PARTITIONTIME'``.
+            Identifier to use instead of default `_PARTITIONTIME` partition
+            column. Defaults to `'PARTITIONTIME'`.
         storage_client
-            A ``BigQueryReadClient`` from the
-            ``google.cloud.bigquery_storage_v1`` package. If not set, one is
-            created using the ``project_id`` and ``credentials``.
+            A `BigQueryReadClient` from the `google.cloud.bigquery_storage_v1`
```
deepyaman, line 500: Nit: I had these with double '`' to be consistent with the `do_connect` docstring; maybe I can update `do_connect` in a follow-up to be consistent (don't want to re-trigger CI now, just for a docs change).

cpcloud: Yeah, that makes sense!

ibis/backends/bigquery/__init__.py
cpcloud: Oh, good point!

cpcloud: Removed!

PR #9606, opened 2024-07-16
fix(pyspark): set lower bound of pyspark to 3.3.3 to avoid maintenance burden of pytest collection hook

cpcloud:
Increase the lower bound of pyspark to avoid a big hack in test collection. Also reduces hacking around numpy imports when making those optional in #9564.


gforsyth (approved):
:shipit:

PR #9607, opened 2024-07-16
refactor(polars): remove numpy usage entirely

cpcloud:
Remove numpy usage from the polars backend entirely, it is no longer needed to work around missing functionality.


jcrist (approved):
Nice

PR #9608, opened 2024-07-16
docs(algolia): add methods from backend pages to algolia index

### LINKED ISSUE #9600: docs: missing to_parquet() docs
### Please describe the issue

xref zulip: https://ibis-project.zulipchat.com/#narrow/stream/405265-tech-support/topic/specify.20partition.20columns.20in.20to_parquet.28.29.3F/near/451643876 

We have docstrings but I wasn't able to find this rendered, is this intentional because it's experimental or they are somehow missing? 

 https://github.com/ibis-project/ibis/blob/f50cbfc204712a6465684c3597b26473b683b463/ibis/expr/types/core.py#L568-L604

### Code of Conduct

gforsyth:
## Description of changes

The backend-specific methods are added to the backend pages
programmatically using `quarto`.  That seems to lead to them not showing
up in our Algolia search, because they aren't being scraped.

It's a little gross, but I've added in a section to our api docs
rendering scripts to also dump the corresponding records for each method
into a backend-specific json file during `quarto` rendering.

Then those backend-specific records files are loaded and added to the
Algolia index in `upload-algolia-api.py`

## Issues closed

Resolves #9600

### DIFF
--- a/.github/workflows/upload-algolia-api.py
+++ b/.github/workflows/upload-algolia-api.py
@@ -1,5 +1,7 @@
 from __future__ import annotations  # noqa: INP001
 
+import glob
+import json
 import os
 import re
 from functools import partial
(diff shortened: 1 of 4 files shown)


cpcloud (approved):
Seems fine to me! I guess we should be careful about moving `Backend` classes around now to avoid breaking the index, but that seems low risk given that they haven't moved for a very long time.

.github/workflows/upload-algolia-api.py
```diff
@@ -94,6 +96,19 @@ def main():
...
+    # There is a hook in docs/backends/_templates/api.qmd that calls
+    # `dump_methods_to_json_for_algolia` that serializes all the backend methods
+    # to a backend-specific json file in docs/backends/
+    # (Not Pandas and Impala because those backend pages don't use the template)
+    #
+    # Here, we load those records and upload them to the Algolia index
+    for record_json in glob.glob("docs/backends/*.json"):
+        with open(record_json) as f:
+            records = json.load(f)
+        index.save_objects(records)
```
cpcloud, line 110: How expensive are these calls? Is it worth trying to make a single call?

cpcloud: If there's any effort required, I'd say let's leave it until we notice a massive slowdown caused by this (which may never actually occur!)

gforsyth: nah, pretty easy to smoosh them together

PR #9617, opened 2024-07-17
test(ci): ensure that we do not skip all of `test_client.py` if polars is not installed

cpcloud:
Avoid skipping all of `test_client.py` when polars is not installed.


cpcloud:
Confirmed that `test_client.py` is now running for Exasol: https://github.com/ibis-project/ibis/actions/runs/9973508235/job/27558993516?pr=9617#step:16:76

ibis/backends/tests/test_expr_caching.py
```diff
@@ -0,0 +1,146 @@
...
+def test_persist_expression_multiple_refs(backend, con, alltypes):
+    non_cached_table = alltypes.mutate(
+        test_column=ibis.literal("calculation"), other_column=ibis.literal("big calc 2")
+    )
+    op = non_cached_table.op()
+    cached_table = non_cached_table.cache()
+
+    cached_df = cached_table.to_pandas()
+    backend.assert_frame_equal(non_cached_table.to_pandas(), cached_df)
+    assert sys.getrefcount(op) == 2
```
cpcloud, line 78: cc @jcrist for all things reference count related 😬 

This assertion passes, but if you apply this diff (inlining the `cached_table.to_pandas()` call, then refcount assertion fails. Here's the diff:

```patch
diff --git a/ibis/backends/tests/test_expr_caching.py b/ibis/backends/tests/test_expr_caching.py
index 9551f4133..f38df40c4 100644
--- a/ibis/backends/tests/test_expr_caching.py
+++ b/ibis/backends/tests/test_expr_caching.py
@@ -73,8 +73,7 @@ def test_persist_expression_multiple_refs(backend, con, alltypes):
     op = non_cached_table.op()
     cached_table = non_cached_table.cache()
 
-    cached_df = cached_table.to_pandas()
-    backend.assert_frame_equal(non_cached_table.to_pandas(), cached_df)
+    backend.assert_frame_equal(non_cached_table.to_pandas(), cached_table.to_pandas())
     assert sys.getrefcount(op) == 2
 
     name = cached_table.op().name
```

Digging around a bit with `gc.get_referrers`, it looks like a stack frame from inside pyspark's `pandas/conversion.py` is holding a reference to `cached_table` in the inlined case but not when the DataFrame is lifted out into a variable.

Would appreciate any thoughts you have here. I'll probably keep the original code and xfail pyspark for now if there's no obvious way forward here.

cpcloud: I'm pushing some other commits, and I'll avoid force pushing for now. You'll want to poke around at https://github.com/ibis-project/ibis/pull/9617/commits/6d0971b73314df22972df194b4168ae817dfc652.

cpcloud: Going to force push to pick up the flink fix.

gforsyth (approved):
Looks good to me.

PR #9622, opened 2024-07-17
feat(flink): create views from more mem data types

deepyaman:
## Description of changes

Related to #9617, this marks a few tests in `ibis/backends/tests/test_client.py` as `notimpl`, because they would mean updating a view (think that's generally not reasonable?).

At the same time, there's no reason we shouldn't support at least constructing the memtable from a broader set of types. This goes through the `ibis.memtable` route, since it already supports a variety of types.


cpcloud (approved):
Ship it!

PR #9624, opened 2024-07-17
docs(clickhouse): put the clickhouse tutorial into sidebar

chloeh13q:
## Description of changes

Not sure how this happened but (probably my fault) I realized that the ClickHouse tutorial was not being indexed in the sidebar. You can access it at https://ibis-project.org/tutorials/data-platforms/clickhouse currently but it does not show up under navigation. I think it's supposed to be under "Cloud Data Platforms"


cpcloud (approved):
Thanks!

PR #9625, opened 2024-07-17
test: add test case for UDF null handling

NickCrews:
I discovered this slightly inconsistent null handling behavior with python UDFs (maybe only in duckdb, waiting to see what CI says):

When passed NULLs as scalars, [duckdb does what it says it should](https://duckdb.org/docs/api/python/function.html#null-handling), and never actually passes them into the function.
But when passed columns, the UDF seems to receive any NULLs as a None.

I'm not sure if this is a bug on the duckdb side (I think it is), or on the ibis side.

Either way, it might be nice to officially define what the expected behavior is. Not sure what other backends do here. I would think not passing in NULLs might be better:
- it's gonna be more performant if we can handle it on the backend side
- it removes a whole class of bugs that a user might cause
- If ONE of our backends ALWAYS filters out null, then it is going to be impossible to pass in NULLs to udfs there.

### DIFF
--- a/ibis/backends/tests/test_udf.py
+++ b/ibis/backends/tests/test_udf.py
@@ -4,6 +4,7 @@ import sys
 
 from pytest import mark, param
 
+import ibis
 import ibis.common.exceptions as com
 from ibis import _, udf
 from ibis.backends.tests.errors import Py4JJavaError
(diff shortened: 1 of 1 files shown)


cpcloud:
Agree that defining NULL handling semantics for UDFs is something we should do.

This definitely seems like a bug on the DuckDB side, it's reproducible without Ibis in the mix.

Instead of adding a PR with only a test, I think it makes sense to open a different PR that implements a NULL handling with the backends that we have that support UDFs.

NickCrews:
Here is the duckdb bug: https://github.com/duckdb/duckdb/issues/13138

PR #9634, opened 2024-07-18
feat(api): add programmatic `epoch` API for computing unix epoch in seconds, milliseconds, and microseconds

cpcloud:
Adds an `epoch` API for getting UNIX epochs in different units. Currently supports seconds, milliseconds and microseconds. I did only a few of the backends so that there could be some smaller follow ups for people getting their feet wet in Ibis development by adding a translation rule for one of the unimplemented backends.

### DIFF
--- a/ibis/backends/sql/compilers/bigquery.py
+++ b/ibis/backends/sql/compilers/bigquery.py
@@ -113,6 +113,9 @@ class BigQueryCompiler(SQLGlotCompiler):
         ops.TimeFromHMS: "time_from_parts",
         ops.TimestampNow: "current_timestamp",
         ops.ExtractHost: "net.host",
+        ops.ExtractEpochMilliseconds: "unix_millis",
+        ops.ExtractEpochMicroseconds: "unix_micros",
+        ops.ExtractEpochSeconds: "unix_seconds",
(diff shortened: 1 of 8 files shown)


cpcloud:
Definitely not ready to merged for 9.2.

cpcloud:
If a backend doesn't have an exact way to compute integer units with more precision than seconds, then we should simply not implement that operation. Otherwise if we do something like multiply the unix timestamp by a scale factor and then cast, because of floating point values we can potentially get incorrect results.

E.g.,

```
D select epoch('2243-01-01 01:00:00.000001'::timestamptz);
┌───────────────────────────────────────────────────────────────────────┐
│ epoch(CAST('2243-01-01 01:00:00.000001' AS TIMESTAMP WITH TIME ZONE)) │
│                                double                                 │
├───────────────────────────────────────────────────────────────────────┤
│                                                     8615052000.000002 │
└───────────────────────────────────────────────────────────────────────┘
```

(DuckDB has the proper `epoch_ms` and `epoch_us`, so that's not a concern for that backend specifically, this is just an illustrative example).

cpcloud:
Closing for now.

PR #9649, opened 2024-07-21
feat(api): support selectors in window function `order_by` and `group_by`

### LINKED ISSUE #9637: feat(api): support selectors in window function `group_by` and `order_by` arguments

Yes, the problem is that I can't use selectors with `over(group_by=..., order_by=...)`.

I'm trying to use selectors with window function specifications.

I'd like to use selectors with window function specifications.

`main`

DuckDB

### Code of Conduct

cpcloud:
Support selectors in window function `group_by` and `order_by` arguments. Closes #9637.

### DIFF
--- a/ibis/common/selectors.py
+++ b/ibis/common/selectors.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+import abc
+from typing import TYPE_CHECKING
+
+from ibis.common.grounds import Concrete
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    import ibis.expr.types as ir
+
+
+class Selector(Concrete):
+    """A column selector."""
+
+    @abc.abstractmethod
+    def expand(self, table: ir.Table) -> Sequence[ir.Value]:
+        """Expand `table` into value expressions that match the selector.
+
+        Parameters
+        ----------
+        table
+            An ibis table expression
+
+        Returns
+        -------
+        Sequence[Value]
+            A sequence of value expressions that match the selector
+
+        """
--- a/ibis/expr/builders.py
+++ b/ibis/expr/builders.py
@@ -13,6 +13,7 @@ from ibis.common.annotations import annotated, attribute
 from ibis.common.deferred import Deferred, Resolver, deferrable
 from ibis.common.exceptions import IbisInputError
 from ibis.common.grounds import Concrete
+from ibis.common.selectors import Selector  # noqa: TCH001
 from ibis.common.typing import VarTuple  # noqa: TCH001
 
 if TYPE_CHECKING:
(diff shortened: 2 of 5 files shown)


ibis/common/selectors.py
```diff
@@ -0,0 +1,49 @@
...
+
+import abc
+from typing import TYPE_CHECKING
+
+from ibis.common.grounds import Concrete
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    import ibis.expr.types as ir
```
kszucs, line 11: I would prefer not to move the base selector class to `ibis.common` since it is tied to `ir.Table`. 
I think `Selector`s could be viewed as another kind of deferred producing a sequence of outputs and deferreds are decoupled from the `ir`.

kszucs: If I recall correctly even `bind()` would properly handle that, so refactoring selectors as `Resolver`s could simplify the codebase, but that should be done in a separate PR.

ibis/expr/builders.py
```diff
@@ -145,8 +146,8 @@ class WindowBuilder(Builder):
     how: Literal["rows", "range"] = "rows"
     start: Optional[RangeWindowBoundary] = None
     end: Optional[RangeWindowBoundary] = None
-    groupings: VarTuple[Union[str, Resolver, ops.Value]] = ()
-    orderings: VarTuple[Union[str, Resolver, ops.SortKey]] = ()
+    groupings: Selector | VarTuple[Union[str, Resolver, Selector, ops.Value]] = ()
```
kszucs, line 149: Is the first `Selector |` actually needed?

ibis/common/selectors.py
cpcloud: I moved the class to avoid an import cycle and having it anywhere else didn't really fit.

If you have another suggestion that isn't to refactor the whole class :) I'm happy to hear it!

kszucs: Ohh, I wasn't aware of the import cycle.

ibis/expr/builders.py
cpcloud: Hm, maybe not!

cpcloud: Nope, not needed!

ibis/expr/builders.py
```diff
@@ -145,8 +146,8 @@ class WindowBuilder(Builder):
     how: Literal["rows", "range"] = "rows"
     start: Optional[RangeWindowBoundary] = None
     end: Optional[RangeWindowBoundary] = None
-    groupings: VarTuple[Union[str, Resolver, ops.Value]] = ()
-    orderings: VarTuple[Union[str, Resolver, ops.SortKey]] = ()
+    groupings: VarTuple[Union[str, Resolver, Selector, ops.Value]] = ()
+    orderings: VarTuple[Union[str, Resolver, Selector, ops.SortKey]] = ()
```
cpcloud, line 150: Since there's a bit of noise from dealing with the import cycle from using `Selector` at the top level of `builders.py`, I'm calling out that this is the change that enables use of a `Selector` in the `group_by` and `order_by` window APIs.

We're already calling `Table.bind` on these two fields in the appropriate place, we simply needed to allow them in the signature of this builder to allow them to get passed through the API.

jcrist (approved):

PR #9652, opened 2024-07-22
fix(api): add support for using deferreds in the `argmin`/`argmax` `key` argument

cpcloud:
Adds support for using deferreds in the `argmin` and `argmax` `key` argument.


gforsyth (approved):
:sunglasses:

PR #9657, opened 2024-07-22
feat(api): support deferred or string column names in `cov`/`corr` me…

jcrist:
Supports things like `t.a.corr("b")`/`t.a.corr(_.b)`.


cpcloud (approved):
Nice!

PR #9658, opened 2024-07-22
fix(mssql): avoid calling `.commit()` unless a DDL operation is being performed

### LINKED ISSUE #9654: test(mssql): flaky test around dataframe interchange protocol
[This CI run](https://github.com/ibis-project/ibis/actions/runs/10042271184/job/27752348001) fails every once in a while in the same way. Not sure what's happening but it would be nice if we could track down the flakiness at some point.

cpcloud:
This PR attempts to address flaky behavior of MS SQL observed in CI.

It seems like calling `commit()`/`rollback()` with a `SELECT` statement, AKA autocommit behavior
can cause this issue.

The origin is not 100% clear to me, but there are a number of places where the
same problem (function sequence error) with the same solution show up
(disabling autocommit or not calling commit with a `SELECT` statement):

- https://stackoverflow.com/questions/25769043/function-sequence-error-in-pyodbc
- https://github.com/ansible-collections/community.general/issues/1137
- https://www.reddit.com/r/learnpython/comments/x8568m/function_sequence_error_in_sqlalchemy/
- https://github.com/explorerhq/django-sql-explorer/issues/423

Here I avoid calling `cur.commit()` unless a DDL statement is being executed.

The remaining case is `raw_sql()`, which I opted to avoid calling `commit` in
since that would have this problem when calling it with a `SELECT` statement.

Users of `raw_sql` are therefore responsible for handling commit/rollback when
invoking it for DDL statements.

Closes #9654.

### DIFF
--- a/ibis/backends/mssql/__init__.py
+++ b/ibis/backends/mssql/__init__.py
@@ -252,6 +252,11 @@ class Backend(SQLBackend, CanCreateCatalog, CanCreateDatabase, CanCreateSchema, 
 
     @contextlib.contextmanager
     def begin(self):
+        with contextlib.closing(self.con.cursor()) as cur:
+            yield cur
+
+    @contextlib.contextmanager
+    def _ddl_begin(self):
         con = self.con
         cur = con.cursor()
         try:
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
thanks for tracking this down!

:shipit:

PR #9659, opened 2024-07-22
feat(api): accept more input types in `ibis.range`

### LINKED ISSUE #9557: bug(timestamps): `ibis.range` doesn't accept `ibis.timestamp` objects (ibis expressions)
### What happened?

```

In [2]: expr = ibis.range(ibis.now() - ibis.interval(days=1), ibis.now(), ibis.interval(seconds=1))
```

Fails with 

```
TypeError: Unable to normalize <class 'ibis.expr.types.temporal.TimestampScalar'> to timestamp
```

Looks like we expect timestamp-like inputs to `range` to be Python literals, but we should allow arbitrary timestamp-typed expressions.

`main`

DuckDB

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Closes #9557.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -2237,7 +2237,9 @@ def _timestamp_range(
     step: datetime.timedelta | ir.IntervalValue,
 ) -> ir.ArrayValue:
     return ops.TimestampRange(
-        start=normalize_datetime(start), stop=normalize_datetime(stop), step=step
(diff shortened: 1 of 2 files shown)


jcrist (approved):

cpcloud:
Ok, looks like I, uh, didn't add very many API level tests the first time around. I'll add some that aren't just doctests 😬

PR #9661, opened 2024-07-22
refactor(joins): require explicit abstract table as RHS of joins

### LINKED ISSUE #9571: feat: add support for polars, arrow RBR as RHS of a join
We currently support transparent `memtable` creation from pandas DataFrames and pyarrow Tables if they are provided as the RHS of a join expression.
We should extend that to include other data inputs that are supported by `memtable`, namely polars DataFrames and arrow RecordBatchReaders

gforsyth:
## Description of changes

We have (had) limited support for passing in in-memory objects as the
RHS of a join, where we would create a memtable for the user and then
use that.  For backends where memtable creation is expensive, or for
queries where there may be multiple calls to the same in-memory data, it
is better to be explicit and first register the in-memory data with the
backend using either `memtable` or `create_table`.

BREAKING CHANGE: Passing a `pyarrow.Table` or a `pandas.DataFrame` as
the right-hand-side of a join is no longer supported.

To join against in-memory data, you can pass the in-memory object to
`ibis.memtable` or `con.create_table` and use the resulting table object
instead.

## Issues closed

* Resolves #9571


gforsyth:
Converting this to draft to avoid merging early

cpcloud (approved):

PR #9663, opened 2024-07-22
feat(pyarrow): support arrow PyCapsule interface in more places

### LINKED ISSUE #9660: feat: Implement Arrow PyCapsule Interface for data import, schema

This is a follow up to a previous feature request (https://github.com/ibis-project/ibis/issues/9140). That added _export_ via the PyCapsule interface but I think there's also further gains from being able to _import_ via the PyCapsule interface as well. 

Additionally, the ibis schema is presumably Arrow or arrow-compatible under the hood? It would be great to implement the PyCapsule Interface on the schema object as well.

Easier interoperability between Ibis and other Python Arrow-based libraries.

I'm not sure how ibis works under the hood and whether there's an easy API for Arrow import that's already in Python? I see there's a `read_parquet` method but is that pushed down to relevant implementations?

Latest

_No response_

### Code of Conduct

jcrist:
~Two changes here:~

- ~Support `__arrow_c_schema__` in `ibis.Schema` objects~
- Support objects implementing `__arrow_c_stream__` as inputs to `ibis.memtable`

Fixes #9660.

### DIFF
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -927,27 +927,6 @@ def test_self_join_memory_table(backend, con, monkeypatch):
         param(
             lambda: pa.table({"a": ["a"], "b": [1]}).to_batches()[0],
             "df_arrow_single_batch",
-            marks=[
-                pytest.mark.notimpl(
-                    [
-                        "bigquery",
-                        "clickhouse",
-                        "duckdb",
-                        "exasol",
-                        "impala",
-                        "mssql",
-                        "mysql",
-                        "oracle",
-                        "postgres",
-                        "pyspark",
-                        "risingwave",
-                        "snowflake",
-                        "sqlite",
-                        "trino",
-                        "databricks",
-                    ]
-                )
-            ],
             id="pyarrow_single_batch",
         ),
         param(
--- a/ibis/backends/tests/test_generic.py
+++ b/ibis/backends/tests/test_generic.py
@@ -1222,6 +1222,24 @@ def test_memtable_construct_from_pyarrow(backend, con, monkeypatch):
     )
 
 
+def test_memtable_construct_from_pyarrow_c_stream(backend, con):
+    pa = pytest.importorskip("pyarrow")
+
+    class Opaque:
+        def __init__(self, table):
+            self._table = table
+
+        def __arrow_c_stream__(self, *args, **kwargs):
+            return self._table.__arrow_c_stream__(*args, **kwargs)
+
+    table = pa.table({"a": list("abc"), "b": [1, 2, 3]})
+
+    t = ibis.memtable(Opaque(table))
+
+    res = con.to_pyarrow(t.order_by("a"))
+    assert res.equals(table)
+
+
 @pytest.mark.parametrize("lazy", [False, True])
 def test_memtable_construct_from_polars(backend, con, lazy):
     pl = pytest.importorskip("polars")
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -412,42 +412,55 @@ def memtable(
 
 @lazy_singledispatch
 def _memtable(
-    data: pd.DataFrame | Any,
+    data: Any,
     *,
     columns: Iterable[str] | None = None,
     schema: SchemaLike | None = None,
     name: str | None = None,
 ) -> Table:
-    import pandas as pd
+    if hasattr(data, "__arrow_c_stream__"):
+        # Support objects exposing arrow's PyCapsule interface
+        import pyarrow as pa
 
+        data = pa.table(data)
+    else:
+        import pandas as pd
+
+        data = pd.DataFrame(data, columns=columns)
+    return _memtable(data, columns=columns, schema=schema, name=name)
+
+
+@_memtable.register("pandas.DataFrame")
+def _memtable_from_pandas_dataframe(
+    data: pd.DataFrame,
+    *,
+    columns: Iterable[str] | None = None,
+    schema: SchemaLike | None = None,
+    name: str | None = None,
+) -> Table:
     from ibis.formats.pandas import PandasDataFrameProxy
 
-    if not isinstance(data, pd.DataFrame):
-        df = pd.DataFrame(data, columns=columns)
-    else:
-        df = data
-
-    if df.columns.inferred_type != "string":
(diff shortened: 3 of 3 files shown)


jcrist:
I'm going to split the schema one out into a new PR, I'm not sure if the `__arrow_c_stream__` input support makes sense.

jcrist:
I'm not sure if supporting `__arrow_c_stream__`-implementing objects in `ibis.memtable` makes sense. It certainly is easy to do. But these objects don't necessarily already hold data in memory - there's a reason that "stream" instead of "table" is used here. In particular `RecordBatchReaders` and other streaming interfaces also implement this method. To support this in a `memtable`, we'd necessarily have to load all data into memory, which may be a performance footgun.

Indeed, this change originally ran into test failures because supporting this implicitly added support for `con.create_table("new", record_batch_reader)` to multiple new backends. We _could_ fix this by explicitly erroring for `RecordBatchReader` types (as done here) asking the user to write the code needed to explicitly load everything into memory themselves.

I'm tempted to do the same here for all objects exposing `__arrow_c_stream__`. This would mean that the only code change here would be an error message for these objects telling users how to fix their code, but I'd rather the user fully understand the performance impact than accidentally load GBs of data into RAM.

Thoughts @kylebarron?

jcrist:
> I'm tempted to do the same here for all objects exposing `__arrow_c_stream__`.

Thinking more on this, I think this is the correct approach. I really don't want to be encouraging users into doing inefficient things, and implicitly slurping up data in anything exposing a `__arrow_c_stream__` method seems risky. Barring comments from others I plan to add a nice user-facing error that tells users how to explicitly load data when an object like this is passed to `ibis.memtable`, but not doing it for the user since they may not have understood the implications. Will hold off for a bit before doing this for comment from others (@kylebarron :))

kylebarron:
Managing when to materialize an Arrow stream is indeed tricky. I'd say it's best to be a conscious decision by the end user, where they can choose among APIs that do or don't materialize the stream. E.g. this is why both `pyarrow.table` and `pyarrow.RecordBatchReader.from_stream` exist for the same C stream input, so that the user can decide whether to materialize it or keep it lazy.

In terms of ibis, is it possible for something like a `StreamTable` to exist? For ibis to hold the stream under the hood as a representation of future data, and then to pass that stream off to whatever backend when it executes the query?

As long as it's documented that an `ibis.memtable` is always fully materialized in memory, I'd argue that it's still beneficial to accept an `__arrow_c_stream__`, because it's the standard way for a user to transfer some Arrow table that is already in memory.

kylebarron:
> Indeed, this change originally ran into test failures because supporting this implicitly added support for `con.create_table("new", record_batch_reader)` to multiple new backends. We _could_ fix this by explicitly erroring for `RecordBatchReader` types (as done here) asking the user to write the code needed to explicitly load everything into memory themselves.

I think this is fair. As I understand it: users can explicitly pass data with `__arrow_c_stream__` to `ibis.memtable`, so it's explicit that data will be fully loaded in memory. If you did have some sort of "streaming table" that you could pass to backends, I think that would be fine to use implicitly within `con.create_table`.

jcrist:
> In terms of ibis, is it possible for something like a `StreamTable` to exist? For ibis to hold the stream under the hood as a representation of future data, and then to pass that stream off to whatever backend when it executes the query?

Not really, no. Ibis expects that data in an `ibis.memtable` is reusable in multiple queries. Since a `RecordBatchReader` (or other streams) cannot be traversed multiple times, using them in this way isn't viable. Some backends do support registering `RecordBatchReader`s as sources - in these cases you can query _once_ on them, but that's expected. Likewise we have some streaming backends like `flink` where a source table is a stream, but that's backend specific.

TL;DR: `ibis.memtable` objects need to hold data in memory as part of the op graph, but there are other mechanisms in some backends for wrapping streaming sources.

---

Re: your other comments, I'm having a hard time understanding if you're agreeing or disagreeing with my proposal above :). To clarify, there are two options here:

1. Implicitly coerce any stream (whether in memory or not) into a fully concrete in-memory source when passed to `ibis.memtable`. This has performance impacts for things that aren't already in memory, and may not be the "right" thing to do in all cases.  Right now we explicitly don't do this for `RecordBatchReader`s, but if we were to support `__arrow_c_stream__`-implementing-types in `ibis.memtable`, then we'd want to support those too.

2. Add a nice user-facing error for if a user calls `ibis.memtable` on a `RecordBatchReader`/`__arrow_c_stream__`-object that tells the user how to explicitly load the data they're asking for. This requires a bit more effort from the user, but avoids letting the user accidentally do the wrong thing. I've done that here for one type already in this PR: https://github.com/ibis-project/ibis/blob/28aff8acec4fec7af6d12a4dfd6590569be93a64/ibis/expr/api.py#L558-L562.

Right now I'm leaning towards option 2, but can see both sides. Note that if we do decide that option 1 is what we want, this would require a larger lift since currently `create_table`/`insert` implicitly use `ibis.memtable` in some cases where we'd want a more efficient path for streaming sources. Option 2 is really easy for us to do, but that's because it's just adding some nicer error messages and no actual support :).

kylebarron:
Sorry for not being clear; I agree that it's better not to implicitly materialize the stream in memory and prefer option 2. I really like that error message.

kylebarron:
If it's well-documented that `ibis.memtable` is a fully in-memory object (as the name suggests), then I think option 1 is appropriate. If a user passes an object into the `ibis.memtable` constructor, then it should follow that the source will be fully materialized. 

You had previously referenced
```py
con.create_table("new", record_batch_reader)
```

Does that implicitly use an `ibis.memtable` object? I wouldn't infer from the name `con.create_table` that `record_batch_reader` would be materialized; this is where maybe Arrow C Stream objects shouldn't be automatically used, instead erroring and suggesting the user to call
```py
con.create_table("new", ibis.memtable(record_batch_reader))
```

> Add a nice user-facing error for if a user calls ibis.memtable on a RecordBatchReader/__arrow_c_stream__-object

It _happens_ that `pyarrow.RecordBatchReader` has a `read_all` method, but that's a library-specific API, not something inherent to all stream holders. And if the producer isn't pyarrow-based, then the user may not want to pass it through a pyarrow object (maybe this is moot if ibis itself will always depend on pyarrow).

cpcloud (approved):

PR #9677, opened 2024-07-23
fix(drop): ignore order for `DropColumns` equality

deepyaman:
## Description of changes

Since the ordering doesn't matter for `DropColumns`, a `VarTuple[str]` is probably the wrong data structure for the `columns_to_drop` attribute. This is a bug in the implementation of `DropColumns`—we should either change the type (a `frozenset[str]` would make more sense), or override the comparison so that `.equals` treats those as equal.


cpcloud (approved):
LGTM, the Ms SQL failures seem unrelated

PR #9683, opened 2024-07-24
chore(deps): remove `packaging` package where it is not used

cpcloud:
We aren't using the `packaging` package in the oracle or polars backends anymore, so let's remove it.


ncclementi (approved):
LGTM!

Issue #9535, opened 2024-07-07
feat(test): improve pytest mark names and/or docs


NickCrews:
### Is your feature request related to a problem?

inspired from https://github.com/ibis-project/ibis/pull/9483#discussion_r1661157508

the current test mark names aren't great.

### What is the motivation behind your request?

Not obvious what they mean, and aren't even that strictly defined/used. This makes it difficult
- when writing a test to know which mark to use 
- when I see a mark how to interpret it

### Describe the solution you'd like

I can see several different axes by which we could measure a failing test:
- fixableness: is this easy or hard to fix? how much effort should we put in?
- cause: is this an upstream bug? a bug in ibis?
- commision vs ommission: is this an actual wrong behavior, or have we just not gotten around to implementing it yet?
- severity: sometimes tests fail due to some platform/OS/unrelated bug, or we can't get the right version of some package installed in CI, but for an end user the functionality works just fine.
- another axes you can think of?

In an ideal world we could encode all of these axes. But of course we can't have 2^3=8 (or more) different marks for each possible combo. So my interpretation is we currently use marks that describe 95% of test failures:

- pytest.mark.never: We have given up on this, we "expect" this to be broken, you probably don't want to sink time in here. eg anything to do with arrays or structs in sqlite, they will probably never be added to sqlite. eg not fixable, the cause is upstream. commission/ommision isn't really relevant.
- pytest.mark.notimpl: We label ibis as the cause, and error by ommision. It is fixable, but could be easy or hard
- pytest.mark.notyet: upstream is the cauase. could be commission or ommission. It is fixable, but I at least interpret this as difficult.
- pytest.mark.broken: this is fixable, but IDK how hard. this is ibis's fault, by commission.

Are there other corners of our axes space that we don't have a good name for?

are some axes more important than others and we want to emphasize them more? I think fixableness is one of the most important ones. WHY something is broken is usually an educated guess anyways, if you actually dig into it the cause and commission/ommission you might find something different, but as a reader of a test case I don't actually care too much, I just want to know if I should try to fix this or if I can expect someone else to ever fix it.

I think our current marks actually work pretty well. Maybe just rename "notyet" to "upstream"? Then all the other marks mean it is ibis's fault.

[...]

PR #9688, opened 2024-07-24
test: remove `broken` marker, better define `notimpl` and `notyet`

NickCrews:
inspired by https://github.com/ibis-project/ibis/issues/9535

The broken mark had the meaning of "this broke for some reason, but it wasn't the fault of this PR, so lets ignore it for now". I find this only meaningful in the context of that individual PR though: as soon as the PR is merged, the usefulness drops to near 0. I come across this mark as I'm reading code and I have no idea what action I should take. notimpl, notyet, and never are much more useful, they signal where the fault lies and what to do about it.

I went through and converted all the instances of broken to one of these other 3 marks. It was fairly fast and definitely error prone. I could see the argument for converting these all to plain xfails when we aren't sure. I'm not sure how bad it is for these marks to be incorrect.

I tweaked the prose that describes these 3 marks. I would love your editing of them.


ibis/backends/tests/test_temporal.py
```diff
@@ -41,17 +41,19 @@
 )
 from ibis.common.annotations import ValidationError
 
+mark_druid_timestamp_is_string = pytest.mark.notimpl(
```
NickCrews, line 44: I also factored out this mark that was used about 50 times in this file. I could keep this out of this PR too.

NickCrews: nvm this was made irrelevant by #9692

gforsyth (approved):
Thanks for putting this in @NickCrews -- I agree the utility of `broken` vanishes after a PR is merged.

At a glance, the marks you've chosen look reasonable - some of them may be wrong, but it's both subjective and imperfect, so I think we can just deal with it.

I think we should try this for a bit -- we may end up restoring something similar.  To your point about "plain xfails", that may be the path we want to take (with whatever extra stuff we'll need to specify which backends to xfail for a given test).

Issue #9468, opened 2024-06-28
feat: Consistent tablenotfound exception across backends


ssabdb:
### Is your feature request related to a problem?

Different backends expose tables not being found in different ways.

For example,
```
connection = ibis.connect("pandas://")
connection.table("missing-table")
KeyError: 'missing-table'
```

```
connection = ibis.connect("bigquery://....")
connection.table("missing-table")
google.api_core.exceptions.NotFound: '...'
```

### What is the motivation behind your request?

What I'd love to be able to do is to catch a consistent error in the handling code which is agnostic of the underlying backend, e.g.

```
def my_function(connection):
    try:
        connection.table("missing-table")
    except IbisTableNotFound:
        <handle cleanly>
```

### Describe the solution you'd like

.

### What version of ibis are you running?

9.0.0

### What backend(s) are you using, if any?

DuckDB, Bigquery

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
+1, definitely a good idea!

relmaazouz:
This was added on the 9.1.0 version:
![Screenshot 2024-07-20 at 02 32 45](https://github.com/user-attachments/assets/2bc47a0e-9539-4863-a522-7dc0fe787ff6)

ssabdb:
It raises IbisError on postgres, but:
  * IbisError seems like a generic exception, not an exception which is explicit about TableNotFound. A table not existing is sufficiently generic and common across backends that (in my opinion) it should be part of the API.
  * This definitely isn't the case with the bigquery backend even in 10.x, when released. I don't think that's an error with Bigquery, I think that's inconsistency across backends on how to catch missing tables.
 
![image](https://github.com/user-attachments/assets/d93a8082-9517-4a59-a59d-ef2c6a199bfb)

relmaazouz:
It should explicit the nested error message.

ncclementi:
Going to get started on this one. I think having a `TableNotFound` being raised across backends when there is no table, is the most reasonable thing here. 

@relmaazouz I'm not sure I follow your last comment.

ssabdb:
Great to see moving forwards!

A few thoughts:
- Backwards compatibility: I've already got code which handles the different exceptions across the limited numbers of backends I support and I'm sure there will be other code out there too. This will break that. I don't think there's any way of preventing that from happening but I don't know what this project's compatibility policy is? 
[...]

PR #9695, opened 2024-07-25
refactor(table-api): unify exception type for all backends to `TableNotFound` when a table does not exist

ncclementi:
Closes https://github.com/ibis-project/ibis/issues/9468

TODO: 
- [x] Decide wether we want to raise the original backend dependent exception when existing 
- [x] pandas and dask, unclear what exactly are we doing when raising KeyError
- [x] datafusion raise exception, not sure where is being caught
- [x] exasol seems to be letting thee exception to the backend, need to catch it before. 
- [x] bigquery and snowflake, do separate test
- [x] understand flink and how to handle the error
- [x] Trino test somethign weird with the quotes/name of table I think I should move the `!r` to the instance of the exception. 
- [x] Druid seems to be letting the exception to the backend, need to catch it before. 
- [x] impala seems to be letting thee exception to the backend, need to catch it before. 
- [x] fix postgres test_create_and_drop_table
- [x] fix mysql exception catch

BREAKING CHANGE: Missing tables now all raise a single Ibis exception. Library code that catches backend-specific exceptions to handle missing tables will need to be adjusted to use `TableNotFound` instead.


ibis/backends/mysql/__init__.py
```diff
@@ -227,8 +226,11 @@ def get_schema(
         ).sql(self.dialect)
 
         with self.begin() as cur:
-            cur.execute(sge.Describe(this=table).sql(self.dialect))
-            result = cur.fetchall()
+            try:
+                cur.execute(sge.Describe(this=table).sql(self.dialect))
+                result = cur.fetchall()
+            except Exception:  # noqa: BLE001
```
ncclementi, line 232: I don't like this, we could modify TableNotFound to also include the message of the original exception raised. Unclear weather that's the path or we are happy with this being raised in Ibis land.

cpcloud: This seems like the wrong approach here. Instead of catching the exception, we should check if `result` is empty and if it is, raise `TableNotFound`. Otherwise we're potentially hiding a slew of bugs as `TableNotFound`. Anything from incorrect sqlglot code generation to transient errors from the server.

ibis/backends/polars/__init__.py
```diff
@@ -72,7 +72,11 @@ def list_tables(self, like=None, database=None):
         return self._filter_with_like(list(self._tables.keys()), like)
 
     def table(self, name: str) -> ir.Table:
-        schema = sch.infer(self._tables[name])
+        try:
+            schema = sch.infer(self._tables[name])
+        except KeyError:
+            raise com.TableNotFound(name)
```
cpcloud, line 78:
```suggestion
        try:
            table = self._tables[name]
        except KeyError:
            raise com.TableNotFound(name)
            
        schema = sch.infer(table)
```

Otherwise, we're potentially catching bugs from `KeyError`s raised in `sch.infer`, which we definitely cannot guarantee won't happen.

Generally speaking, making a single call in a `try:` block is preferred, especially if the exception type you're catching is extremely common or generic.

ibis/backends/postgres/tests/test_client.py
```diff
@@ -137,7 +137,7 @@ def test_create_and_drop_table(con, temp_table, params):
 
     con.drop_table(temp_table, **params)
 
-    with pytest.raises(com.IbisError):
+    with pytest.raises(com.TableNotFound, match="Table not found"):
```
cpcloud, line 140:
```suggestion
    with pytest.raises(com.TableNotFound, match=temp_table):
```

Probably want to check that the table name is mentioned, not only that the message says "Table not found".

ibis/backends/trino/tests/test_client.py
```diff
@@ -164,7 +164,9 @@ def test_table_access_database_schema(con):
     t = con.table("region", database=("tpch", "sf1"))
     assert t.count().execute()
 
-    with pytest.raises(exc.IbisError, match='Table not found: tpch."tpch.sf1".region'):
+    with pytest.raises(
+        exc.TableNotFound, match='Table not found: tpch."tpch.sf1".region'
```
cpcloud, line 168:
```suggestion
        exc.TableNotFound, match=r'Table not found: tpch\."tpch\.sf1"\.region'
```

Tiny nit, since `match` is a regex, let's make sure the dots are actual dots :)

(I realize this was pre-existing)

ibis/common/exceptions.py
```diff
@@ -21,6 +21,13 @@
     from collections.abc import Callable
 
 
+class TableNotFound(Exception):
+    def __init__(self, table_name, message="Table not found"):
+        self.table_name = table_name
```
cpcloud, line 26: Are we using this attribute somewhere outside the constructor here?

cpcloud (changes requested):
Couple changes needed, thanks for doing this!

ibis/backends/mysql/__init__.py
ncclementi: I tried that first, but it fails before in here, so I don't get to do the `fetchall()`:
```
-> cur.execute(sge.Describe(this=table).sql(self.dialect))
(Pdb) pp cur.execute(sge.Describe(this=table).sql(self.dialect))
*** pymysql.err.ProgrammingError: (1146, "Table 'ibis_testing.foo' doesn't exist")
```

ibis/backends/polars/__init__.py
ncclementi: I see your point but, there is a problem with this as `self._tables[name]` doesn't raise the `KeyError`. it's 
`self._tables[name]` is `<LazyFrame at 0x12C92B250>` it only raises once we try to infer the schema, so the workflow suggested doesn't work. Not sure what's the best way to do this here

ibis/common/exceptions.py
ncclementi: I initially thought I might but it looks like I'm not.  
```suggestion
```

ibis/backends/druid/__init__.py
```diff
@@ -130,11 +133,17 @@ def get_schema(
...
-            .sql(self.dialect)
-        )
+        try:
+            schema = self._get_schema_using_query(
+                sg.select(STAR)
+                .from_(sg.table(table_name, db=database, catalog=catalog))
+                .sql(self.dialect)
+            )
+        except ProgrammingError as e:
+            if re.search(r"\bINVALID_INPUT\b", str(e)):
```
ncclementi, line 143: I wasn't able to get anything better than this matching, if anyone has any ideas, suggestions welcome

ibis/backends/flink/__init__.py
```diff
@@ -303,7 +304,16 @@ def get_schema(
         qualified_name = sg.table(table_name, db=catalog, catalog=database).sql(
             self.name
         )
-        table = self._table_env.from_path(qualified_name)
+        try:
+            table = self._table_env.from_path(qualified_name)
+        except Py4JJavaError as e:
+            # This seems to msg specific but not sure what a good work around is
+            if re.search(
+                rf"Table `{re.escape(table_name)}` was not found",
```
ncclementi, line 312: I don't like this, but I'm not sure if there is a better way to get the error matching here. Suggestions welcomed.

ibis/backends/pyspark/__init__.py
```diff
@@ -514,7 +515,14 @@ def get_schema(
         table_loc = self._to_sqlglot_table((catalog, database))
         catalog, db = self._to_catalog_db_tuple(table_loc)
         with self._active_catalog_database(catalog, db):
-            df = self._session.table(table_name)
+            try:
+                df = self._session.table(table_name)
+            except AnalysisException as e:
+                # we can alternative check for e.getErrorClass == 'TABLE_OR_VIEW_NOT_FOUND'
+                if e.getSqlState() == "42P01":
+                    # spark sqlerror state for undefined_table see https://spark.apache.org/docs/3.4.1/sql-error-conditions-sqlstates.html
```
ncclementi, line 523: I think there are less chances that the code name, but we can check for `e.getErrorClass == 'TABLE_OR_VIEW_NOT_FOUND'` or both, suggestions welcome.

ncclementi:
I think this is good to get a first path of review. I left some comments on the cases where the exception caught is not great/pretty and we are doing some for of string matching. If anyone has better ideas for those, suggestions are more than welcome. 

Notes/help needed: 
- Exasol and Impala don't run on mac arm64 which makes it harder to test/fix those. 
- How do we trigger bigquery and snowflake to run?

ibis/backends/tests/test_client.py
```diff
@@ -1629,3 +1629,8 @@ def test_from_connection(con, top_level):
     new_con = backend.from_connection(getattr(con, CON_ATTR.get(con.name, "con")))
     result = int(new_con.execute(ibis.literal(1, type="int")))
     assert result == 1
+
+
+def test_table_not_found(con):
+    with pytest.raises(com.TableNotFound, match="Table not found"):
+        con.table("foo")
```
cpcloud, line 1636:
```suggestion
        con.table(util.gen_name("table_not_found"))
```

Let's use a name that no human would reasonably author by accident. You might need `from ibis import util` for this as well.

ibis/backends/pyspark/__init__.py
cpcloud: Checking the error class seems much more maintainable.

cpcloud:
The PySpark error import needs an `ImportError` guard to handle the module path change from 3.3 to 3.5. 

There are some examples around the codebase for this that were recently added.

ibis/backends/druid/__init__.py
```diff
@@ -130,11 +133,17 @@ def get_schema(
...
-            sg.select(STAR)
-            .from_(sg.table(table_name, db=database, catalog=catalog))
-            .sql(self.dialect)
-        )
+        try:
+            schema = self._get_schema_using_query(
+                sg.select(STAR)
+                .from_(sg.table(table_name, db=database, catalog=catalog))
+                .sql(self.dialect)
+            )
```
cpcloud, line 141:
```suggestion
        query = (
            sg.select(STAR)
            .from_(sg.table(table_name, db=database, catalog=catalog))
            .sql(self.dialect)
        )
        try:
            schema = self._get_schema_using_query(query)
```

So that it's clear what can raise here.

cpcloud (changes requested):
Bit code movement needed, but other than that looking good!

cpcloud:
I added implementations for Impala and Exasol.

cpcloud:
Looking into Snowflake. There seems to be plenty of structured information about errors, but interpreting what it means is another story.

ibis/backends/flink/__init__.py
ncclementi: @cpcloud do we think this is alright for flink? It's a bit specific but I didn't find an alternative, If so I'll remove the comment here

cpcloud:
Added a Snowflake implementation. It's ... something.

ncclementi:
I added this to the 10.0 milestone as it will have breaking changes for those using Ibis as a library.

ibis/backends/flink/__init__.py
cpcloud: It seems like there's some poking around the java object that could be done.

cpcloud:
Ok, BigQuery implementation is up.

ibis/backends/pyspark/__init__.py
```diff
@@ -514,7 +521,12 @@ def get_schema(
         table_loc = self._to_sqlglot_table((catalog, database))
         catalog, db = self._to_catalog_db_tuple(table_loc)
         with self._active_catalog_database(catalog, db):
-            df = self._session.table(table_name)
+            try:
+                df = self._session.table(table_name)
+            except AnalysisException as e:
+                if e.getErrorClass == "TABLE_OR_VIEW_NOT_FOUND":
+                    raise com.TableNotFound(table_name)
```
cpcloud, line 528:
```suggestion
                    raise com.TableNotFound(table_name) from e
```

ibis/backends/snowflake/__init__.py
```diff
@@ -646,11 +646,30 @@ def get_schema(
...
+            # standard
+            #
+            # Nowhere does this exist in Snowflake's documentation but this
+            # exists in MariaDB's docs and matches the SQLSTATE error code
+            #
+            # https://mariadb.com/kb/en/sqlstate/
+            # https://mariadb.com/kb/en/mariadb-error-code-reference/
+            # and the least helpful version: https://docs.snowflake.com/en/developer-guide/snowflake-scripting/exceptions#handling-an-exception
+            if e.sqlstate == "42S02":
+                raise com.TableNotFound(table.sql(self.dialect))
```
cpcloud, line 671:
```suggestion
                raise com.TableNotFound(table.sql(self.dialect)) from e
```

cpcloud (changes requested):
Just a couple more missing `from e`s

ibis/backends/flink/__init__.py
ncclementi: I tried multiple thinks, I wasn't able to land in anything better than this, I tried inspecting `e.java_exception` to get the `getSQLState` but I wasn't able. Maybe I'm missing something.

ibis/backends/pyspark/__init__.py
```diff
@@ -514,7 +521,12 @@ def get_schema(
         table_loc = self._to_sqlglot_table((catalog, database))
         catalog, db = self._to_catalog_db_tuple(table_loc)
         with self._active_catalog_database(catalog, db):
-            df = self._session.table(table_name)
+            try:
+                df = self._session.table(table_name)
+            except AnalysisException as e:
+                if e.getErrorClass() == "TABLE_OR_VIEW_NOT_FOUND":
+                    raise com.TableNotFound(table_name) from e
```
cpcloud, line 528:
```suggestion
                if e.getErrorClass() == "TABLE_OR_VIEW_NOT_FOUND":
                    raise com.TableNotFound(table_name) from e
                raise
```

You still need to raise for any other `AnalysisException` that isn't handled by the error class check.

cpcloud (changes requested):

ibis/backends/pyspark/__init__.py
cpcloud: Also this check doesn't seem to be working for PySpark 3.3.3, I'm looking into it.

cpcloud: Fixed in the latest commit.

ibis/backends/druid/__init__.py
```diff
@@ -130,11 +133,18 @@ def get_schema(
...
+        query = (
             sg.select(STAR)
             .from_(sg.table(table_name, db=database, catalog=catalog))
             .sql(self.dialect)
         )
+        try:
+            schema = self._get_schema_using_query(query)
+        except PyDruidProgrammingError as e:
+            if re.search(r"\bINVALID_INPUT\b", str(e)):
+                raise com.TableNotFound(table_name) from e
```
cpcloud, line 145:
```suggestion
                raise com.TableNotFound(table_name) from e
            raise
```

Remember that you have to continue to `raise`. There are other reasons besides the missing table that these errors may occur.

ibis/backends/flink/__init__.py
```diff
@@ -303,7 +304,16 @@ def get_schema(
...
+                str(e.java_exception.toString()),
+            ):
+                raise exc.TableNotFound(table_name) from e
```
cpcloud, line 315:
```suggestion
                raise exc.TableNotFound(table_name) from e
            raise
```

ibis/backends/flink/__init__.py
cpcloud: I went ahead and changed this to catch the error, and then check for table existence instead of matching.

`INVALID_INPUT` is way too general of a string to match on in any case.

cpcloud:
I think this is good to go.

What an absolute hellscape this problem is.

Thanks for powering through @ncclementi!

ncclementi:
Thanks @cpcloud, great job you too! Thanks for the pairing and troubleshooting. When we are ready for 10.0 we can rebase and deal with conflicts if we have to.

ncclementi:
Hey y'all I tried to rebase this locally and did not go well 😞 so I'd need some help getting this one up to mark to be able to consolidate changes.

ibis/backends/mysql/__init__.py
```diff
@@ -224,8 +224,15 @@ def get_schema(
         ).sql(self.dialect)
 
         with self.begin() as cur:
-            cur.execute(sge.Describe(this=table).sql(self.dialect))
-            result = cur.fetchall()
+            query = sge.Describe(this=table).sql(self.dialect)
```
cpcloud, line 227: Let's avoid executing code under an open cursor if that code doesn't require the cursor to be open.

ibis/backends/mysql/__init__.py
```diff
@@ -474,7 +481,7 @@ def _in_memory_table_exists(self, name: str) -> bool:
                 cur.fetchall()
         except pymysql.err.ProgrammingError as e:
             err_code, _ = e.args
-            if err_code == ER.NO_SUCH_TABLE:
```
cpcloud, line 477: Any reason to change the import here?

cpcloud (approved):
Just a couple small things, but otherwise this LGTM.

ibis/backends/mysql/__init__.py
ncclementi: This show up as a conflict when merging main because it was initially without the `ER.` on my end but this was used in other PR with that was merged added into main with `ER.`, happy to unify it with `ER.`

cpcloud (approved):

PR #9698, opened 2024-07-26
perf(schema): speed up failure case of duplicate keys when constructing schemas

cpcloud:
Adds a modest 2x performance improvement when trying to construct schemas with duplicate keys.


cpcloud:
Merging!

PR #9700, opened 2024-07-26
feat(mssql): support case-sensitive collations

### LINKED ISSUE #9494: feat: Support for Microsoft Fabric Warehouse

NO.

xref #9454 

I would like to be able to interact with a typical Microsoft Fabric Warehouse, which has quite similar interfaces as any other Datawarehousing technologies that `ibis` already supports.
It would be nice if we could continue using the `ibis.mssql` backend for it.

9.1.0

DuckDB & SQLServer

### Code of Conduct

gforsyth:
This should enable support for users working with Microsoft Fabric Warehouse

## Description of changes

This updates a few internal table and view calls to use the
case-sensitive names which will allows us to (hopefully) support both
case-sensitive and case-insensitive collations.

The correct casing for the tables and views we use often (and the
corresponding columns):

Info schema tables:
- INFORMATION_SCHEMA.COLUMNS
- INFORMATION_SCHEMA.SCHEMATA
- INFORMATION_SCHEMA.TABLES
Temp table location: tempdb.dbo
Catalogs: sys.databases
Databases: sys.schemas

I also had to swap out a stored procedure that _really_ didn't like
being used with the case-sensitive collation, but I think the
replacement is nicer anyway (we can push down some column selection to
the engine).

## Issues closed

* Resolves #9494

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -73,7 +73,7 @@ services:
       retries: 20
       test:
         - CMD-SHELL
-        - $(find / -name sqlcmd -type f -executable) -C -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "SELECT 1"
+        - $(find /opt -name sqlcmd -type f -executable) -C -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "SELECT 1"
     ports:
       - 1433:1433
     volumes:
--- a/compose.yaml
+++ b/compose.yaml
@@ -73,7 +73,7 @@ services:
       retries: 20
       test:
         - CMD-SHELL
-        - sleep 10 && $(find / -name sqlcmd -type f -executable) -C -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "IF DB_ID('ibis_testing') is NULL BEGIN CREATE DATABASE [ibis_testing] END"
+        - $(find / -name sqlcmd -type f -executable) -C -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "IF DB_ID('ibis_testing') is NULL BEGIN CREATE DATABASE [ibis_testing] END"
     ports:
       - 1433:1433
     volumes:
--- a/compose.yaml
+++ b/compose.yaml
@@ -63,6 +63,11 @@ services:
     environment:
       MSSQL_SA_PASSWORD: 1bis_Testing!
       ACCEPT_EULA: "Y"
+      # The default collation in MSSQL isS QL_Latin1_General_CP1_CI_AS
+      # where the CI stands for Case Insensitive.
+      # We use a case-sensitive collation for testing so that we don't
+      # break users with case-sensitive collations.
+      MSSQL_COLLATION: "Latin1_General_100_BIN2_UTF8"
     healthcheck:
       interval: 1s
       retries: 20
--- a/compose.yaml
+++ b/compose.yaml
@@ -73,7 +73,7 @@ services:
       retries: 20
       test:
         - CMD-SHELL
-        - $(find / -name sqlcmd -type f -executable) -C -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "IF DB_ID('ibis_testing') is NULL BEGIN CREATE DATABASE [ibis_testing] END"
+        - $(find / -name sqlcmd -type f -executable) -C -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "SELECT 1"
     ports:
       - 1433:1433
     volumes:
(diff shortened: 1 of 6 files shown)


cpcloud:
Brain is hurting at the prospect of debugging why the 3.10 job passed but not the 3.12 one 🧠

cpcloud:
Well, locally 3.12.4 (the same version as in CI) is all green for me with this PR 🤔

cpcloud:
Deleting my sqlserver image, because ms is known to alter the image under that tag without notice.

cpcloud:
Re-running CI, to see if that was a fluke run. Not if a successful run is more concerning, but we'll have some additional information!

gforsyth:
I'm also having some weird behavior locally.  Initially things were green on my local, now they're not.  I might add in a few checks to ensure that the collation is being respected

gforsyth:
> Not sure if a successful run is more concerning, but we'll have some additional information!

well... it succeeded this time...

gforsyth:
I'm marking this as a draft until I can figure out how to make our testing setup behave deterministically w.r.t. collation

gforsyth:
Wow I hate everything.

We use a docker health-check in the `compose.yaml` file to create the `ibis_testing` catalog so that it is there for us to insert data into when we start `pytest`.

mssql doesn't set collation until it's done restoring/creating various backups, so if the health check fires before the collation is set, then it gets set to the default collation, and you can't change that collation without removing the volume.

cpcloud:
Thanks for tracking that down. @ncclementi and I observed this behavior but struggled to get to the bottom of it in a reasonable amount of time.

What is the MS recommended sequence of steps for creating a database with a specific collation such that it isn't defeated by some other MS SQL process?

gforsyth:
> What is the MS recommended sequence of steps for creating a database with a specific collation such that it isn't defeated by some other MS SQL process?

I really don't know.  Right now my only plan is to delay the health check...

gforsyth:
This is a semi-brittle fix to the collation race condition, but it seems to work consistently on my machine and CI?

compose.yaml
```diff
@@ -63,12 +63,17 @@ services:
...
+      # We use a case-sensitive collation for testing so that we don't
+      # break users with case-sensitive collations.
+      MSSQL_COLLATION: Latin1_General_100_BIN2_UTF8
     healthcheck:
       interval: 1s
       retries: 20
       test:
         - CMD-SHELL
-        - $(find / -name sqlcmd -type f -executable) -C -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "IF DB_ID('ibis_testing') IS NULL BEGIN CREATE DATABASE [ibis_testing] END"
+        - sleep 10 && $(find / -name sqlcmd -type f -executable) -C -S localhost -U sa -P "$$MSSQL_SA_PASSWORD" -Q "IF DB_ID('ibis_testing') is NULL BEGIN CREATE DATABASE [ibis_testing] END"
```
cpcloud, line 76: Hm, I feel like there's gotta be a saner way than this 😬 

That said, if we were going to do it this way a comment would be helpful to explain the left turn into wondering caused by encountering a `sleep 10` 😂

cpcloud:
I have an idea: let's try to create database as part of the `conftest.py` setup instead of the health check hack (guilty as charged on that front).

cpcloud:
Ok, I think I might have something working!

ibis/backends/mssql/tests/conftest.py
```diff
@@ -35,13 +35,23 @@ class TestConf(ServiceBackendTest):
...
         return self.data_dir.joinpath("csv").glob("*.csv")
 
+    def postload(self, **kw: Any):
+        self.connection = self.connect(database=IBIS_TEST_MSSQL_DB, **kw)
+
+    def _load_data(self, *, database: str = IBIS_TEST_MSSQL_DB, **_):
+        with self.connection._safe_raw_sql(
+            "IF DB_ID('ibis_testing') is NULL BEGIN CREATE DATABASE [ibis_testing] END"
+        ):
+            pass
```
gforsyth, line 45: Nice!  I was getting cryptic user errors when I tried this, but I must've made a syntax error somewhere

cpcloud: Probably was the other thing I had to change, `database=None` doesn't do what you might expect.

cpcloud:
Locally all passing with

```
pytest -m mssql -n auto --dist loadgroup --snapshot-update
```

ibis/backends/mssql/tests/test_client.py
```diff
@@ -35,11 +36,23 @@
     ("DATETIME", dt.Timestamp(scale=3)),
     # Characters strings
     ("CHAR", dt.string),
-    ("TEXT", dt.string),
+    param(
+        "TEXT",
```
cpcloud, line 40: Good to know that the `TEXT` type doesn't support modern text.

cpcloud:
This seems like a bug fix or a feature given that a thing that was previously not working is now working, also 🥳 for that!

cpcloud (approved):
LGTM, ship it!

PR #9701, opened 2024-07-26
feat(api): add `length` parameter to the `String` datatype

### LINKED ISSUE #9382: feat: allow specify the length of string column

We are using Snowflake backend, and trying to load parquet file into the snowflake, the parquet has columns with fixed length string type, but after read_parquet called, all column became VARCHAR(16777216),
tried other backend, like duckdb/mysql, all seems to upscale the string type to max string length instead of maintaining the length from source.
ideally this should be presered.

Our data governance require some strict data type checking, and preserving the source string length will be greatly valuable. 

I would like to extend ibis schema to allow specify max length of a string column, and also make read_parquet to honor the fixed length physical type.

9.0.0

MySQL, MSSQL, Snowflake

### Code of Conduct

cpcloud:
Adds an optional `length` parameter to the `String` datatype.

Closes #9382.

### DIFF
--- a/ibis/backends/clickhouse/tests/test_datatypes.py
+++ b/ibis/backends/clickhouse/tests/test_datatypes.py
@@ -125,7 +125,7 @@ def test_array_discovery_clickhouse(con):
         ),
         param(
             "Array(FixedString(32))",
-            dt.Array(dt.String(nullable=False), nullable=False),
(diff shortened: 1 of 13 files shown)


cpcloud:
I am unlikely to have time to wrap this up, as implementing this well requires handling cases like union of `string` columns with `string(N)` columns which would require deeper changes to the type system.

Issue #7073, opened 2023-09-07
feat: Array.sum(), .min(), .max(), mean(), etc.


NickCrews:
### Is your feature request related to a problem?

Also, I could see .mean(), .min(), etc.

[Here are the supported agg functions for duckdb](https://duckdb.org/docs/archive/0.8.1/sql/functions/nested#list-aggregates). Not sure about other backends.

Related to https://github.com/ibis-project/ibis/issues/6975, except in this case I can't do the workaround that I mention there. Perhaps that issue is a special case of this.

### Describe the solution you'd like

Probably top-level functions on ArrayValue, but that lends itself to only making sense for the few core agg functions that all backends support. Once you get to weird agg functions like `kurtosis` that duckdb supports, it becomes harder to justify adding a `ArrayValue.kurtosis()` just to support duckdb. Perhaps a "sparse" API similar to duckdb's of `ArrayValue.agg("sum")` makes more sense in that it allows access to rare aggregation functions that only are in some backends.

### What version of ibis are you running?

main

### What backend(s) are you using, if any?

duckdb

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
I think these operations can be implemented for a variety of backends:

- BigQuery (operations on arrays can be done in a subquery)
- ClickHouse (there are a bunch of `arrayFoo` operations that we would map to, e.g., [`arraySum`](https://clickhouse.com/docs/en/sql-reference/functions/array-functions#arraysum))
- DuckDB (already discussed in the issue title)
- Trino (using the [`reduce`](https://trino.io/docs/current/functions/array.html#reduce) function where necessary)
- PySpark (has some of the requested functions [built in](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/functions.html#collection-functions))
- Polars (there are a bunch under the [`list`](https://pola-rs.github.io/polars/py-polars/html/reference/expressions/list.html) namespace)

NickCrews:
Nice! What do you think about the dense (array.max()) vs sparse (array.agg("max")) API? Should we come up with a matrix of func vs backend to see which are commonly supported? Could also have dense API for the most common aggs, but also include sparse API to allow filling in the gaps

cpcloud:
I think having both seems reasonable. `.agg` is also a bit more friendly for programmatic use.

NickCrews:
proposal for list of dense ops:

- .any()
- .all()
- .min()
- .max()
- .sum()
[...]

PR #9704, opened 2024-07-27
feat(array): implement min, max, any, all, sum, mean

NickCrews:
fixes https://github.com/ibis-project/ibis/issues/7073


ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,219 @@ def flatten(self) -> ir.ArrayValue:
...
+        │ [None, False]        │ False   │
+        │ [None, True]         │ True    │
+        │ [None]               │ NULL    │
+        │ []                   │ NULL    │
+        │ NULL                 │ NULL    │
+        └──────────────────────┴─────────┘
+        """
+        return ops.ArrayAll(self).to_expr()
+
+    def array_min(self) -> ir.NumericValue:
```
cpcloud, line 1151:
```suggestion
    def min(self) -> ir.NumericValue:
```

We should name these consistently, and I don't think that this could ever mean anything other than a reduction within each array.

ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,219 @@ def flatten(self) -> ir.ArrayValue:
...
+        │ [1, 2, ... +1]       │     1 │
+        │ [None, 6]            │     6 │
+        │ [None]               │  NULL │
+        │ []                   │  NULL │
+        │ NULL                 │  NULL │
+        └──────────────────────┴───────┘
+        """
+        return ops.ArrayMin(self).to_expr()
+
+    def array_max(self) -> ir.NumericValue:
```
cpcloud, line 1182:
```suggestion
    def max(self) -> ir.NumericValue:
```

ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,219 @@ def flatten(self) -> ir.ArrayValue:
...
+        │ [1, 2, ... +1]       │     3 │
+        │ [None, 6]            │     6 │
+        │ [None]               │  NULL │
+        │ []                   │  NULL │
+        │ NULL                 │  NULL │
+        └──────────────────────┴───────┘
+        """
+        return ops.ArrayMax(self).to_expr()
+
+    def array_sum(self) -> ir.NumericValue:
```
cpcloud, line 1213:
```suggestion
    def sum(self) -> ir.NumericValue:
```

ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,219 @@ def flatten(self) -> ir.ArrayValue:
...
+        │ [1, 2, ... +1]       │     6 │
+        │ [None, 6]            │     6 │
+        │ [None]               │  NULL │
+        │ []                   │  NULL │
+        │ NULL                 │  NULL │
+        └──────────────────────┴───────┘
+        """
+        return ops.ArraySum(self).to_expr()
+
+    def array_mean(self) -> ir.FloatingValue:
```
cpcloud, line 1244:
```suggestion
    def mean(self) -> ir.FloatingValue:
```

ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,220 @@ def flatten(self) -> ir.ArrayValue:
...
+        return ops.ArrayAny(self).to_expr()
+
+    def all(self) -> ir.BooleanValue:
+        """Return whether all elements in the array are true (ignoring nulls).
+
+        Returns
+        -------
+        BooleanValue
+            Whether all elements in the array are true (ignoring nulls).
+            Will be NULL if no elements are non-NULL.
```
cpcloud, line 1113:
```suggestion
            Will be NULL if all elements are NULL.
```

Seems a bit easier to grok IMO.

ibis/expr/types/arrays.py
cpcloud: Oh, gosh, I guess some of these are actually defined in some of our backends.

DuckDB seems to define `min` as elementwise min...

Ugh, I really do not like the naming scheme `array_*`.

cpcloud: I think pluralizing the name of the singular reduction operation makes a lot of sense here and aligns well with the semantics of a collection.

- `.mins()`
- `.maxs()`
- `.means()`
- `.sums()`

I feel like this makes sense given the result shape of these operations is `Columnar`.

This also would never show up on non `Array` columns (Ibis has always intentionally worked like that), so there's much less chance of confusion.

ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,220 @@ def flatten(self) -> ir.ArrayValue:
...
+        -------
+        Value
+            Mean of the values in the array
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.memtable({"arr": [[1, 2, 3], [None, 6], [None], [], None]})
+        >>> t.mutate(x=t.arr.array_mean())
```
cpcloud, line 1261:
```suggestion
        >>> t.mutate(x=t.arr.means())
```

ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,220 @@ def flatten(self) -> ir.ArrayValue:
...
+        -------
+        Value
+            Sum of the values in the array
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.memtable({"arr": [[1, 2, 3], [None, 6], [None], [], None]})
+        >>> t.mutate(x=t.arr.array_sum())
```
cpcloud, line 1230:
```suggestion
        >>> t.mutate(x=t.arr.sums())
```

ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,220 @@ def flatten(self) -> ir.ArrayValue:
...
+        -------
+        Value
+            Maximum value in the array
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.memtable({"arr": [[1, 2, 3], [None, 6], [None], [], None]})
+        >>> t.mutate(x=t.arr.array_max())
```
cpcloud, line 1199:
```suggestion
        >>> t.mutate(x=t.arr.maxs())
```

ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,220 @@ def flatten(self) -> ir.ArrayValue:
...
+        -------
+        Value
+            Minimum value in the array
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.memtable({"arr": [[1, 2, 3], [None, 6], [None], [], None]})
+        >>> t.mutate(x=t.arr.array_min())
```
cpcloud, line 1168:
```suggestion
        >>> t.mutate(x=t.arr.mins())
```

ibis/expr/types/arrays.py
cpcloud: Also these aren't necessarily going to be `Numeric`, since strings have a well-defined ordering.

ibis/expr/types/arrays.py
NickCrews: That was purposefull so empty arrays are included, though I agree that ALSO is subtle. Maybe "all are NULL or empty" and just be explicit?

ibis/backends/sql/compilers/pyspark.py
```diff
@@ -589,3 +594,43 @@ def _format_window_interval(self, expression):
         this = expression.this.this  # avoid quoting the interval as a string literal
 
         return f"{this}{unit}"
+
+    def visit_ArraySumAgg(self, op, *, arg, output):
```
NickCrews, line 598: OMG hahah wow pyspark is so needy here. Duckdb team needs to use this for their talking point of "ergonomic SQL"

ibis/expr/types/arrays.py
NickCrews: It is a little confusing to use this plural-sounding thing on scalars, eg ibis.literal([1, 2,3]).mins()?
I agree array_min is overly verbose, but I think I'd prefer that to confusing? Drawing a blank for alternatives...

NickCrews (commented):
Thanks for implementing the other back ends! I can't get trini running on my dev machine so debugging the failing tests is gonna be hard, if you are able to fix that it would be awesome!

cpcloud:
> Thanks for implementing the other back ends! I can't get trini running on my dev machine so debugging the failing tests is gonna be hard, if you are able to fix that it would be awesome!

If you view a scalar as a one row, one column table, does that clarify why it makes sense as an operation?

NickCrews:
That wasn't my mental model for a scalar, but with that model then yes this does make sense. This makes me want to write/edit something on the concepts section of the docs that explains this. Since I've been using ibis quite a lot but still never grokked that, it makes me think this isn't advertised enough

Ok I think mins() is probably the best we can do! I'll push this change.

ibis/expr/types/arrays.py
NickCrews: I went with `all are NULL or empty`

NickCrews:
How do I get codespell to ignore the looks-like-a-misspelling of `alls`?

NickCrews:
I need to fix the see also links, the urls are wrong

cpcloud:
Regarding Polars, there's some incorrect behavior of any and all, so I'll leave those out for now. I've created a ticket upstream (https://github.com/pola-rs/polars/issues/17917).

cpcloud:
Actually, we can workaround the polars issue (which isn't a bug, it turns out).

cpcloud:
Ok, I was able to get Trino passing, as well as Snowflake and BigQuery.

To ensure that BigQuery works I parameterized the `data` argument to include two cases: one for nulls and one for non-nulls. These tests also run for every backend, so we can have some confidence that the no-nulls cases are behaving as expected.

cpcloud:
To be crystal clear here, the current implementation ensures that the semantics of the array aggregations match the behavior of the corresponding columnar aggregations. IMO this the right choice to ensure that reasoning about the two cases' aggregation requires only one conceptual understanding for both "plural" aggregates as well as columnar ones.

ibis/backends/sql/compilers/pyspark.py
cpcloud: Yeah, this one was a doozy! TBF Trino also requires this kind of special treatment, because using the fully general `reduce` (Trino) or `aggregate` (Spark) requires handling null behavior.

cpcloud:
Also got postgres and risingwave working, using the strategy as bigquery, which is this:

```sql
SELECT (SELECT AGG(el) FROM UNNEST(array_column))
FROM t
```

ibis/backends/sql/compilers/postgres.py
```diff
@@ -670,3 +670,33 @@ def visit_TableUnnest(
                 join_type="CROSS" if not keep_empty else "LEFT",
             )
         )
+
+    def _unnest(self, expression, *, as_):
+        alias = sge.TableAlias(columns=[sg.to_identifier(as_)])
+        return sge.Unnest(expressions=[expression], alias=alias)
+
+    def visit_ArrayReduction(self, op, *, arg, reduction):
```
NickCrews, line 678: Same comment as others

ibis/backends/sql/compilers/pyspark.py
```diff
@@ -589,3 +594,42 @@ def _format_window_interval(self, expression):
...
```
NickCrews, line 598: Similar as others, rename to eg _array_agg?

ibis/backends/tests/test_array.py
```diff
@@ -1478,3 +1471,104 @@ def test_table_unnest_array_of_struct_of_array(con):
...
+    x = [y for y in x if not pd.isna(y)]
+    if not x:
+        return None
+    return agg(x)
+
+
+@pytest.mark.parametrize(
+    ("agg", "baseline_func"),
+    [
+        (ir.ArrayValue.sums, partial(_agg_with_nulls, sum)),
```
NickCrews, line 1493: Nit: I personally find partial mix more confusing than just writing out the lambda so I almost always do that. Not a blocker for sure, but if you ever aren't sure which to use, there is a little weight that could tip the scales for you

ibis/backends/tests/test_array.py
```diff
@@ -1478,3 +1471,104 @@ def test_table_unnest_array_of_struct_of_array(con):
...
+        (ir.ArrayValue.means, partial(_agg_with_nulls, statistics.mean)),
+    ],
+    ids=["sums", "mins", "maxs", "means"],
+)
+@notimpl_aggs
+@pytest.mark.parametrize(
+    "data",
+    [
+        param(
+            [[1, 2, 3], [None, 6], [5], [None], [], None],
```
NickCrews, line 1505: Should these be reduced to just [None,6] and [None]? the other cases are covered by the non-null param

ibis/backends/tests/test_array.py
```diff
@@ -1478,3 +1471,104 @@ def test_table_unnest_array_of_struct_of_array(con):
...
+            [
+                [True, False],
+                [True, None],
+                [False, None],
+                [True],
+                [False],
+                [None],
+                [],
+                None,
+            ],
```
NickCrews, line 1551: Similar as above

ibis/expr/operations/arrays.py
```diff
@@ -266,3 +266,60 @@ class TimestampRange(Range):
     start: Value[dt.Timestamp]
     stop: Value[dt.Timestamp]
     step: Value[dt.Interval]
+
+
+class ArrayAgg(Value):
```
NickCrews, line 271: Should we rename to _ArrayAgg? I dont think this should be exposed outside this file

ibis/expr/operations/arrays.py
```diff
@@ -266,3 +266,60 @@ class TimestampRange(Range):
...
+
+@public
+class ArrayMax(ArrayAgg):
+    """Compute the maximum value of an array."""
+
+
+# in duckdb summing an array of ints leads to an int, but for other backends
+# it might lead to a float??
+@public
+class ArraySum(ArrayAgg):
```
NickCrews, line 293: Should this also work on arrays of bools? Could punt to a followup if someone actually requests it

ibis/expr/types/arrays.py
```diff
@@ -1059,6 +1059,243 @@ def flatten(self) -> ir.ArrayValue:
...
+        │ [None, False]        │ False   │
+        │ [None, True]         │ True    │
+        │ [None]               │ NULL    │
+        │ []                   │ NULL    │
+        │ NULL                 │ NULL    │
+        └──────────────────────┴─────────┘
+        """
+        return ops.ArrayAny(self).to_expr()
+
+    def alls(self) -> ir.BooleanValue:  # codespell:ignore alls
```
NickCrews, line 1112: Since we added to the code spell config file we should remove these inline noqas

NickCrews (commented):
A few small bits but definitely not blocking! Thanks so much @cpcloud !

ibis/backends/tests/test_array.py
cpcloud:
```suggestion
            [[None, 6], [None]],
```

ibis/expr/operations/arrays.py
cpcloud: I think this is fine to leave as is. `ibis.expr.operations` isn't public.

ibis/expr/operations/arrays.py
cpcloud: I think we can wait until someone requests it.

ibis/backends/tests/test_array.py
cpcloud:
```suggestion
            [[True, None], [False, None], [None]],
```

cpcloud (approved):

PR #9712, opened 2024-07-29
fix(bigquery): repr geospatial values in interactive mode

cpcloud:
Fixes a bug in the BigQuery backend when looking at geospatial types in interactive mode.

### DIFF
--- a/ibis/backends/bigquery/__init__.py
+++ b/ibis/backends/bigquery/__init__.py
@@ -683,12 +683,36 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema):
         self._define_udf_translation_rules(expr)
         sql = super()._to_sqlglot(expr, limit=limit, params=params, **kwargs)
 
+        table_expr = expr.as_table()
+        geocols = frozenset(
+            name for name, typ in table_expr.schema().items() if typ.is_geospatial()
+        )
+
         query = sql.transform(
             _qualify_memtable,
             dataset=getattr(self._session_dataset, "dataset_id", None),
             project=getattr(self._session_dataset, "project", None),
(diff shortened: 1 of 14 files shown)


ibis/backends/bigquery/__init__.py
```diff
@@ -683,12 +683,31 @@ def _to_sqlglot(
...
         ).transform(_remove_null_ordering_from_unsupported_window)
-        return query
+
+        if not geocols:
+            return query
+
+        quoted = self.compiler.quoted
+        return sg.select(
+            sge.Star(
+                replace=[
```
ncclementi, line 703: I'm not sure I quite follow what is happening here. Are we converting text into binary for geo columns, or what is it exactly that we are returning instead of `query` here? Can you explain it in a comment, or in the main comment in the PR what the problem was and how this fixes it?

I can't run the bigquery backend on my laptop so a bit more context would be helpful

ncclementi (approved):
Left a comment asking for a bit more context but outside that it looks good.

ibis/backends/bigquery/__init__.py
cpcloud: I'll add a comment, but this is the same thing we had to do in duckdb to get it to properly display geometries in interactive mode, though here it's for a slightly different reason.

cpcloud: Added a comment in the latest commit.

cpcloud:
Oops, rebased, the comment is [here](https://github.com/ibis-project/ibis/pull/9712/commits/976ebfc86c68eaf5dfbbd159e2b227de7203498c#diff-70e8f790a7cd8392a8b812c47a95be84f3835051e37a873536cefb3cb726d360R700-R704).

PR #9713, opened 2024-07-29
refactor(duckdb): use replace to generate less sql

cpcloud:
Refactors the duckdb output to geospatial to use `* REPLACE` syntax to generate less code when replacing a small number of geospatial columns (usually one) among an otherwise large set of columns. Additionally, we now only need to loop over expression columns once.


gforsyth (approved):
nice!

PR #9727, opened 2024-07-30
docs: fix typo in `drop` method docstring

cpcloud:
Fixes a typo in the `drop` method failure mode example.


cpcloud:
The amount of compute we're triggering here to render a docstring typo fix feels ... excessive.

Issue #9446, opened 2024-06-26
Improve backends test suite coverage


jcrist:
Ibis has a common test suite found in `ibis/backends/tests/` for ensuring methods have identical behavior across backends. Most of these tests make use of one (or more) tables preloaded into the backends (and accessed via fixtures). A common one is `functional_alltypes`, but there are others.

The `functional_alltypes` dataset is very old and isn’t sufficiently interesting to catch all bugs. For example, the `string_col` column is currently just the numbers 0-9 as strings (so all ascii, 1-char length strings). This has resulted in some bugs in functionality making it through to releases, only to be discovered later by users (or us when trying to make changes later). As a recent example, the `NULL` handling behavior of `collect/first/last` wasn’t consistent across backends, but wasn’t caught since our test data didn’t contain a rich enough mix of null/non-null data.

We've been responding to bugs where not all backends have matching behavior for some functions. I propose we spend some time improving our test suite and doing some proactive bug hunting.

Improvements _might_ include:

- Adding one (or several) more interesting datasets that would better cover features tested
- Updating the backend tests to use these datasets (and fixing/filing issues for any bugs discovered)
- Evaluating the existing test functions for each method to ensure edge cases have coverage
- Making more use of `memtable` inputs for some functions to give more targeted inputs for testing edge cases
- Fuzzing using `hypothesis`

Intentionally not being prescriptive here, open to any methods that might incrementally improve our testing process.

PR #9728, opened 2024-07-30
chore: use Hypothesis to generate `float_col` data

deepyaman:
Testing changes related to generation of a better `functional_alltypes` (see https://github.com/ibis-project/ibis/issues/9446)

## Description of changes

Had a chat with @gforsyth earlier today about https://github.com/ibis-project/ibis/issues/9446 (see also https://voltrondata.slack.com/archives/C06HYLURYF8/p1722277286473539). Based on those conversations, taking the approach of trying to make the test data more interesting VERY incrementally, starting with something relatively "easy" (float).

A major goal is to not have to xfail a bunch of tests that were previously passing. To that end, I ended up removing all nulls and NaNs for now, since they cause tons of failures across the board (I think it's also largely because of how the tests are written using pandas). The results on https://github.com/ibis-project/ibis/pull/9728 are now somewhat reasonable; for 70-80ish percent of the backends, only 2-3 tests fail, and those can probably be fixed by setting a tolerance.

For MS SQL, Oracle, maybe SQLite, I think the issue may at least partially be the load scripts (haven't verified). However, for things like RisingWave and Exasol, quite possibly also DataFusion, there may be some deeper issues.

So, keeping the goal of not having to xfail a bunch of working stuff, wanted to get some opinions on some options:

1. Can be more constrained in the randomly-generated data (next, I can try no infinity; will probably fix 90% of the issues). This is still much more interesting than the current data (see attached image).
2. Can replace problematic values (infinities) on load for the largely failing backends. This is a bit weird in that there's a sort of test coverage being reflected in the data used; on the flip side, it's not like we are going to go add infinity support to databases.

I don't have a strong preference here. Not sure how important Inf handling really is (think null values and even NaN should be a lot higher priority), so maybe #1 is a good enough route for now?

**Update**: Removing infinities didn't actually help all that much...

### DIFF
--- a/justfile
+++ b/justfile
@@ -102,7 +102,7 @@ ci-doctest *args:
     just _doctest "poetry run" {{ args }}
 
 # download testing data
-download-data owner="ibis-project" repo="testing-data" rev="master":
+download-data owner="deepyaman" repo="testing-data" rev="chore/generate-float-data":
     #!/usr/bin/env bash
     set -euo pipefail
 


deepyaman:
@gforsyth @jcrist Going to close this, based on the discussion to try defining test data in test files (e.g. as `memtable`s) instead.

PR #9729, opened 2024-07-30
feat(api): support `order_by` in order-sensitive aggregates (`collect`/`group_concat`/`first`/`last`)

### LINKED ISSUE #9170: feat(api): Add `order_by` parameter to the `collect()` method

jcrist:
This adds a new `order_by` keyword argument to order-sensitive aggregates (`collect`/`group_concat`/`first`/`last`) to specify the ordering the aggregation should use. By default no ordering is specified, and the result is backend dependent (following the current behavior).

```python
In [1]: import ibis

In [2]: t = ibis.memtable({"x": [1, 3, 2, 4], "y": [10, 9, 8, 7]}, name="test")

In [3]: t.x.collect().execute()  # no ordering specified, backend-dependent result
Out[3]: [1, 3, 2, 4]

In [4]: t.x.collect(order_by="y").execute()  # use the specified ordering, or error if unsupported
Out[4]: [4, 2, 3, 1]

In [5]: ibis.to_sql(t.x.collect(order_by="y"))
Out[5]: 
SELECT
  ARRAY_AGG("t0"."x" ORDER BY "t0"."y" ASC) FILTER(WHERE
    "t0"."x" IS NOT NULL) AS "ArrayCollect(x, (y,))"
FROM "test" AS "t0"
```

This is generally implemented using either an aggregate-internal `ORDER BY` clause, or via `WITHIN GROUP (...)` (backend dependent). The SQL required here can vary significantly between backends unfortunately :/.

Fixes #9170.

### DIFF
--- a/ibis/backends/sql/compilers/base.py
+++ b/ibis/backends/sql/compilers/base.py
@@ -82,18 +82,13 @@ class AggGen:
 
         __getitem__ = __getattr__
 
-    __slots__ = ("supports_filter", "supports_order_by", "requires_within_group")
+    __slots__ = ("supports_filter", "supports_order_by")
 
     def __init__(
-        self,
-        *,
-        supports_filter: bool = False,
-        supports_order_by: bool = False,
-        requires_within_group: bool = False,
+        self, *, supports_filter: bool = False, supports_order_by: bool = False
     ):
         self.supports_filter = supports_filter
         self.supports_order_by = supports_order_by
-        self.requires_within_group = requires_within_group
 
     def __get__(self, instance, owner=None):
         if instance is None:
@@ -137,13 +132,8 @@ class AggGen:
             args = tuple(compiler.if_(where, arg, NULL) for arg in args)
 
         if order_by and self.supports_order_by:
-            if self.requires_within_group:
-                out = sge.WithinGroup(
-                    this=func(*args), expression=sge.Order(expressions=order_by)
-                )
-            else:
-                *rest, last = args
-                out = func(*rest, sge.Order(this=last, expressions=order_by))
+            *rest, last = args
+            out = func(*rest, sge.Order(this=last, expressions=order_by))
         else:
             out = func(*args)
 
--- a/ibis/backends/sql/compilers/base.py
+++ b/ibis/backends/sql/compilers/base.py
@@ -128,7 +128,7 @@ class AggGen:
         if order_by and not self.supports_order_by:
             raise com.UnsupportedOperationError(
                 "ordering of order-sensitive aggregations via `order_by` is "
-                "not supported for this backend"
+                f"not supported for the {compiler.dialect} backend"
             )
 
         if where is not None and not self.supports_filter:
--- a/ibis/backends/sql/compilers/base.py
+++ b/ibis/backends/sql/compilers/base.py
@@ -132,10 +132,7 @@ class AggGen:
             )
 
         if where is not None and not self.supports_filter:
-            args = tuple(
-                arg if isinstance(arg, sge.Literal) else compiler.if_(where, arg, NULL)
-                for arg in args
-            )
+            args = tuple(compiler.if_(where, arg, NULL) for arg in args)
 
         if order_by and self.supports_order_by:
             *rest, last = args
--- a/ibis/backends/sql/compilers/base.py
+++ b/ibis/backends/sql/compilers/base.py
@@ -82,16 +82,18 @@ class AggGen:
 
         __getitem__ = __getattr__
 
-    __slots__ = ("supports_filter", "supports_order_by")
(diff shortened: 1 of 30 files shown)


jcrist:
I don't have cloud testing setup locally, so I'm :crossed_fingers: that I did everything correctly for bigquery/snowflake.

ibis/backends/sql/compilers/base.py
```diff
@@ -109,21 +119,34 @@ def aggregate(
...
+                "not supported for this backend"
+            )
 
-        if self.supports_filter:
-            return sge.Filter(
-                this=func(*args),
-                expression=sge.Where(this=where),
+        if where is not None and not self.supports_filter:
+            args = tuple(
+                arg if isinstance(arg, sge.Literal) else compiler.if_(where, arg, NULL)
```
cpcloud, line 136: I don't think ignoring `where` for all literals is correct.

There are other aggregate functions with more than one argument, e.g., `cov_samp(x, 2.4) filter (where y > 1)`. In that example `2.4` should be filtered out of the aggregate computation via a `NULL` and it wouldn't be in this implementation.

`group_concat` should be special cased and shouldn't go through the default `self.agg` mechanism, which I think was why it was that way in a few backends before.

cpcloud (changes requested):
Thanks!

One concern about ignoring filters with literal inputs.

cpcloud:
There's a couple things that can be pulled out here, I'll do a couple separate PRs.

1. RisingWave seems to support UNNEST now
2. Oracle `group_concat` was just implemented incorrectly, but is pretty easy to get working.

ibis/backends/sql/compilers/base.py
```diff
@@ -79,10 +82,18 @@ def __getattr__(self, name: str) -> Callable:
...
-    __slots__ = ("supports_filter",)
+    __slots__ = ("supports_filter", "supports_order_by", "requires_within_group")
 
-    def __init__(self, *, supports_filter: bool = False):
+    def __init__(
+        self,
+        *,
+        supports_filter: bool = False,
+        supports_order_by: bool = False,
+        requires_within_group: bool = False,
```
jcrist, line 92: I originally implemented things like this, but only snowflake would use this syntax for anything except group_concat (and that one had to be specialized in many backends anyway). Felt cleaner to split it out to a snowflake-specific subclass than add an extra kwarg here. Either works though 🤷.

ibis/backends/sql/compilers/base.py
jcrist: I'm not sure I follow this - if an aggregation takes multiple args and we only apply `where` to the non-literals (or even non-scalars), wouldn't null handling still work the same? Or even more concretely - do these not yield the same result?

```
cov_samp(iff(y > 1, x, NULL), 2.4)

cov_samp(iff(y > 1, x, NULL), iff(y > 1, 2.4, NULL)
```

ibis/backends/sql/compilers/base.py
cpcloud: Hm, yeah, that might be better, let me fiddle with it a bit more. I think I've got everything passing except risingwave.

ibis/backends/sql/compilers/base.py
cpcloud: Assuming backends handle nulls the same in two argument functions, your argument works for the specific case of `cov_samp`, that is,

```sql
cov_sample(iff(y > 1, x, NULL), 2.4)
```

will evaluate to NULL, the same as

```sql
cov_sample(iff(y > 1, x, NULL), iff(y > 1, 2.4, NULL))
```

However, this doesn't generalize to arbitrary aggregates, because the actual input values change depending on the value of `y > 1`, so my argument is that we can't do this in general, because it assumes that all aggregates will always handle nulls the same way.

cpcloud:
I think there's probably an argument to be made for add `order_by` to `arbitrary` but that can be done in a follow up.

cpcloud (approved):

cpcloud:
Sweet, thanks @jcrist! Merging.

PR #9730, opened 2024-07-30
docs(algolia): add custom attributes to backend and core methods

gforsyth:
There are several (mostly linked) improvements here:

- docs(algolia): add custom attributes to backend and core methods

Adds 2 custom attributes, one which notes whether the method is part of
the core API (not on a backend), the other which holds the name of the
backend.

We then have an algolia ranking setup where we use the value of
`core` (descending) to break ties, so `Table.drop` will show up above `RandomBackend.drop`

- fix(quartodoc): search up through `BaseBackend` for docstrings

The above fix exposed a bunch of empty docstrings for many backend methods inheriting from a grandparent class object.

We were looking at the `resolved_bases` to find docstrings that were
defined in a parent class, but this search only goes up a single level
of inheritance.  For all of our sql-based backends, this means we only
go up to `SQLBackend`, not `BaseBackend`.

Manually adding `BaseBackend` to the resolution list fills out a LOT of
missing docstrings on the backend docs pages.

- fix(quartodoc): don't render deprecated schema methods

Having done that, we were now successfully rendering schema methods that are deprecated and we don't want to advertise, so I turned that off.

- fix(quartodoc): don't duplicate entries in resolved_bases

It was possible for these resolved_bases to get duplicated, so I stopped that.

- feat(algolia): configure custom algolia ranking

A lot of algolia config happens in their web UI, which I don't like because it makes it hard to know what we've changed.
This automates the custom ranking that makes use of the new attribute `core` added in the first commit.

I also moved all of the algolia scripts to a directory to keep them together since that collection keeps growing.


docs/backends/_utils.py
```diff
@@ -36,7 +36,11 @@ def find_member_with_docstring(member):
         return member
 
     cls = member.parent
-    for base in cls.resolved_bases:
+    resolved_bases = cls.resolved_bases
+    # If we're a SQLBackend (likely) then also search through to `BaseBackend``
+    if (sqlbackend := resolved_bases[0]).name == "SQLBackend":
```
cpcloud: Is there a more generic fix for this, like altering the lookup code to traverse the class hierarchy?

cpcloud (approved):
LGTM! Just a non-blocking question.

PR #9731, opened 2024-07-31
test(risingwave): mark `test_union_aliasing` as notimpl because arbitrary aggregate function is not supported

cpcloud:
Apparently risingwave has implement support for `UNNEST`.


cpcloud:
Actually, the only thing worth changing is the `reason` in the test. unnest has been working for awhile.

PR #9739, opened 2024-07-31
fix(algolia): add parent class docstring to algolia index

gforsyth:
## Description of changes

Follow-up to #9730.

I fixed adding the parent docstrings in the rendered docs, but those
weren't showing up in the search summary on Algolia.  They do now.


cpcloud (changes requested):
Nice.

cpcloud (approved):

PR #9741, opened 2024-07-31
test(arrays): remove unnecessary series construction in `test_array_remove`

NickCrews:
followup to https://github.com/ibis-project/ibis/pull/9089, I think this is a bit more intuitive


cpcloud:
Reflecting on this a bit, I think the handwritten expected values are more maintainable and easy to understand.

I'll clean up the test body bit though. For example, I don't think we need to construct a series for the expected values.

cpcloud (approved):

Issue #9448, opened 2024-06-26
Support `read_parquet`/`read_csv` for most backends


jcrist:
Currently we only support `read_parquet` for backends that have native support (like `duckdb`). In contrast, we support `to_parquet` for all backends, falling back to a common `pyarrow` implementation if a backend doesn't natively support it.

To provide more uniform feature coverage, we could write an equivalent common `pyarrow` (or other) implementation of our IO input methods (`read_parquet`/`read_csv`/...) that backends like `postgres` could fall back on.

jcrist:
If `duckdb` is installed, there's also the option of using `duckdb` for faster support for `parquet`/`csv` table loading. I think that _could_ be done in a 2nd pass, see #8110, but a pass using `pyarrow` first to cover all backends makes sense to me as a good first step.

jitingxu1:
HI @jcrist I could take this one. Will do it one by one. 

Plan to add `read_parquet` and `read_csv`. How about `read_json` and `read_delta`?

csubhodeep:
Any updates on this one?

PR #9744, opened 2024-08-01
feat: support read_parquet for backend with no native support

jitingxu1:
## Description of changes

Support `read_parquet `for backends that do not have native support (like duckdb). This implementation leverages the PyArrow [read_table](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.read_table.html) function.

If a backend does not have its own version, it will fall back on this pyarrow implementation. 

## Issues closed

This addresses part of issue #9448. Additional tasks related to this issue will be completed and submitted individually.


cpcloud:
@jitingxu1 This PR has a lot of failures. Can you take a look so we can decide how to move forward?

ibis/backends/tests/test_register.py
```diff
@@ -425,6 +425,9 @@ def test_register_garbage(con, monkeypatch):
 def test_read_parquet(con, tmp_path, data_dir, fname, in_table_name):
     pq = pytest.importorskip("pyarrow.parquet")
 
+    if con.name in ["oracle", "exasol"]:
+        pytest.skip("Skip Exasol and Oracle because of the global pytestmark")
```
jitingxu1, line 429: Hi @cpcloud There is a global pytestmark for `oracle` and `exasol`, not sure how to override it in this specific test.  I tried to skip it, if it is the one we want, I could skip it in the `test_read_parquet_glob` too.

Since we are adding `read_parquet()` for some other backends, looks like we need to remove some backend from the `notyet` mark for this test `test_register_garbage`

ibis/backends/tests/test_register.py
```diff
@@ -476,7 +465,11 @@ def test_read_parquet_glob(con, tmp_path, ft_data):
     for fname in fnames:
         pq.write_table(ft_data, tmp_path / fname)
 
-    table = con.read_parquet(tmp_path / f"*.{ext}")
+    if con.name == "clickhouse":
+        # clickhouse does not support read directory
+        table = con.read_parquet(tmp_path / f"*.{ext}")
```
cpcloud, line 470: This doesn't seem like the right approach. You're changing what's being tested. Why can't you leave this code unchanged here?

ibis/backends/tests/test_register.py
cpcloud: Generally when this happens you have to mark every other test in the file with the same marker, and then remove the backends you want to pass from the individual test marker.

cpcloud (changes requested):

ibis/backends/tests/test_register.py
jitingxu1: the pyarrow `read_table` cannot read things like `tmp_path/*.parquet` and clickhouse cannot read the `directory`

gforsyth: The test is whether the backend can read a glob of parquet files, the answer to that seems to be "no", so it should be marked as `notyet`

jitingxu1:
rewrite it to support urls as the input:
- regular url: `https`, 'ftp' and so on
- fsspec compatible url: `s3`, `gcp`
- local files
    - support: single file, directory, glob patterns

@cpcloud for review again, Thanks

ibis/backends/__init__.py
```diff
@@ -1199,6 +1202,61 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+        return self.table(table_name)
+
+    def _get_pyarrow_table_from_path(self, path: str | Path, **kwargs) -> pa.Table:
+        pq = util.import_object("pyarrow.parquet")
+
+        path = str(path)
+        # handle url
+        if util.is_url(path):
+            headers = kwargs.pop("headers", {})
+            req_info = urllib.request.Request(path, headers=headers)  # noqa: S310
```
cpcloud, line 1242: A bunch of annotations are showing up here, suggesting there's no test coverage for all this additional code.

Is all this extra code necessary if we're not going to confirm that it works?

cpcloud (changes requested):

jitingxu1:
increased the test coverage. @cpcloud

ibis/backends/__init__.py
cpcloud, line 1242: Do we really need to use `urllib` here? Why doesn't fsspec work?

ibis/backends/tests/test_register.py
```diff
@@ -448,6 +455,61 @@ def test_read_parquet(con, tmp_path, data_dir, fname, in_table_name):
...
+# test reading a Parquet file from a URL request for backends using pyarrow
+# that do not have their own implementation
+@pytest.mark.parametrize(
+    ("url", "in_table_name"),
+    [
+        ("http://example.com/functional_alltypes.parquet", "http_table"),
+        ("sftp://example.com/path/to/functional_alltypes.parquet", "sftp_table"),
+    ],
+)
+@pytest.mark.notimpl(
```
cpcloud, line 467: Some of these seem like `never` cases, specifically the ones that already support reading parquet natively.

ibis/backends/__init__.py
```diff
@@ -1236,6 +1239,123 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+        ... )
+
+        """
+
+        table = self._get_pyarrow_table_from_path(path, **kwargs)
+        table_name = table_name or util.gen_name("read_parquet")
+        self.create_table(table_name, table)
+        return self.table(table_name)
+
+    def _get_pyarrow_table_from_path(self, path: str | Path, **kwargs) -> pa.Table:
```
cpcloud, line 1333: Why can't the implementation of this just be:

```python
return pq.read_table(path)
```

Did you try that already?

jitingxu1: I tried that in my first commit, it cannot handle all the cases:

such as glob pattern and Parquet files hosted on some uri: i.e HTTPS SFTP 

Pyarrow implements natively the following filesystem subclasses:

[Local FS](https://arrow.apache.org/docs/python/filesystems.html#filesystem-localfs) ([LocalFileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.LocalFileSystem.html#pyarrow.fs.LocalFileSystem))

[S3](https://arrow.apache.org/docs/python/filesystems.html#filesystem-s3) ([S3FileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html#pyarrow.fs.S3FileSystem))

[Google Cloud Storage File System](https://arrow.apache.org/docs/python/filesystems.html#filesystem-gcs) ([GcsFileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.GcsFileSystem.html#pyarrow.fs.GcsFileSystem))

[Hadoop Distributed File System (HDFS)](https://arrow.apache.org/docs/python/filesystems.html#filesystem-hdfs) ([HadoopFileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html#pyarrow.fs.HadoopFileSystem))

jitingxu1: @cpcloud does this make sense to you?

ibis/backends/__init__.py
```diff
@@ -1307,6 +1310,161 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+        # Handle local file paths or patterns
+        paths = glob.glob(path)
+        if not paths:
+            raise ValueError(f"No files found matching pattern: {path!r}")
+        elif len(paths) == 1:
+            paths = paths[0]
+
+        return pq.read_table(paths, **kwargs)
+
+    def _cached(self, expr: ir.Table):
```
cpcloud, line 1430: Why was the addition of the caching APIs necessary?

jitingxu1: ah, it was from previous merging, just removed it.

jitingxu1:
HI @cpcloud ,

I got several timeout error in the CI for this PR, is there something we need to fix in another PR
```
FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_74[trino]@tpcds - Failed: Timeout >90.0s
FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_72[trino]@tpcds - Failed: Timeout >90.0s
FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_94[trino]@tpcds - Failed: Timeout >90.0s
FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_92[trino]@tpcds - Failed: Timeout >90.0s
FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_07[trino]@tpcds - Failed: Timeout >90.0s
```

ibis/backends/tests/test_register.py
jitingxu1: changed it to never

ibis/backends/__init__.py
jitingxu1: changed to fssepc

ibis/backends/tests/test_register.py
```diff
@@ -96,6 +94,7 @@ def gzip_csv(data_dir, tmp_path):
         "trino",
     ]
 )
+@pytest.mark.notimpl(["druid", "exasol", "oracle"])
```
ncclementi, line 97: I'm confuse about this, it looked like before we had support for this, why are we loosing this now?

ibis/backends/tests/test_register.py
```diff
@@ -107,7 +106,7 @@ def test_register_csv(con, data_dir, fname, in_table_name, out_table_name):
 
 
 # TODO: rewrite or delete test when register api is removed
-@pytest.mark.notimpl(["datafusion"])
+@pytest.mark.notimpl(["datafusion", "druid", "exasol", "oracle"])
```
ncclementi, line 109: Some comment than above, why are we loosing this?

ibis/backends/tests/test_register.py
jitingxu1: they were marked as not implemented at the begin of the file before: https://github.com/ibis-project/ibis/blob/main/ibis/backends/tests/test_register.py#L21

now I moved it to each test, because it no longer applies to all tests. 

> I'm confuse about this, it looked like before we had support for this, why are we loosing this now?

ibis/backends/tests/test_register.py
jitingxu1: same as above

jitingxu1:
Hi @cpcloud 

In this PR, I have the trino/impala test on `read_parquet`, it reads about 7300 rows from `functional_alltypes.parquet`(seems like insertion after 10k rows will have a performance issue), I suspect it impacts the trino database performance, I have the following timeout error in other tests, I suspect it is caused by read large parquet file in `test_read_parquet`, does this make sense?

Should I skip the Trino and Impala in this test too? Or do you have better way to handle this?
 
> I got several timeout error in the CI for this PR, is there something we need to fix in another PR
> 
> ```
> FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_74[trino]@tpcds - Failed: Timeout >90.0s
> FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_72[trino]@tpcds - Failed: Timeout >90.0s
> FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_94[trino]@tpcds - Failed: Timeout >90.0s
> FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_92[trino]@tpcds - Failed: Timeout >90.0s
> FAILED ibis/backends/tests/tpc/ds/test_queries.py::test_07[trino]@tpcds - Failed: Timeout >90.0s
> ```
> 
> is it related to trino setup in the CI?

gforsyth:
Ok, that's one pass for the nix mac osx job.  I'm going to cycle it a few times to make sure.

ibis/backends/__init__.py
```diff
@@ -1269,6 +1271,128 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+            An optional name to use for the created table. This defaults to
+            a sequentially generated name.
+        **kwargs
+            Additional keyword arguments passed to the pyarrow loading function.
+            See https://arrow.apache.org/docs/python/generated/pyarrow.parquet.read_table.html
+            for more information.
+
+            When reading data from cloud storage (such as Amazon S3 or Google Cloud Storage),
+            credentials can be provided via the `filesystem` argument by creating an appropriate
+            filesystem object (e.g., `pyarrow.fs.S3FileSystem`).
```
gforsyth, line 1299: This is true, and additionally, `pq.read_table` also supports the standard aws auth patterns (environment variables, or aws SSO credentials, or instance credentials)

ibis/backends/__init__.py
```diff
@@ -1269,6 +1271,128 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+
+        # Handle local file paths or patterns
+        paths = glob.glob(path)
+        if not paths:
+            raise ValueError(f"No files found matching pattern: {path!r}")
+        elif len(paths) == 1:
+            paths = paths[0]
+
+        return pq.read_table(paths, **kwargs)
+
```
gforsyth, line 1395: I think we should reconsider handling all of these cases -- this sort of branching logic means that when a user reports an error, we'll have any number of possible culprits to consider, and it makes it harder to debug for everyone.

I think (and I could be wrong) that nearly all of these cases are covered by `pq.read_table` by itself, and that's much easier to document and debug.

`read_table` also has support for being passed an `fsspec` object, so if someone needs to read from a hypertext url, they can use `fsspec` as a shim for that.  (This is something we can add a note about in the docstring).

ibis/backends/__init__.py
```diff
@@ -1269,6 +1271,128 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+        self.create_table(table_name, table)
+        return self.table(table_name)
+
+    def _get_pyarrow_table_from_path(self, path: str | Path, **kwargs) -> pa.Table:
+        import pyarrow.parquet as pq
+
+        path = str(path)
+        # handle url
+        if util.is_url(path):
+            import fsspec
```
gforsyth, line 1376: `fsspec` is not a dependency of Ibis (it's in our test suite) so this would need extra import handling if we leave it in (but see my other comments)

gforsyth (commented):
Most of my comments are inline -- I think we should start off offering this with a simplified implementation.  It's nice to offer users options, but I think we should be measured in how much we enable directly.

gforsyth:
> Ok, that's one pass for the nix mac osx job. I'm going to cycle it a few times to make sure.

Ok, skipping the mocked URL test on DuckDB seems to have resolved the nested transaction failures on the nix osx CI job

jitingxu1:
> > Ok, that's one pass for the nix mac osx job. I'm going to cycle it a few times to make sure.
> 
> Ok, skipping the mocked URL test on DuckDB seems to have resolved the nested transaction failures on the nix osx CI job

Thank you so much.

ibis/backends/__init__.py
```diff
@@ -1269,6 +1271,100 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
             f"{cls.name} backend has not implemented `has_operation` API"
         )
 
+    @util.experimental
+    def read_parquet(
+        self, path: str | Path | BytesIO, table_name: str | None = None, **kwargs: Any
+    ) -> ir.Table:
```
jitingxu1, line 1277: Instead of BytesIO, I could pass the fsspec object, It could be HTTPFile if we pass an HTTP url. Not sure what is the best way to handle the type of `path`

@gforsyth any suggestion?

gforsyth: I think `fsspec` is a good option.

cpcloud:
This PR is stale and had a number of unaddressed issues.

PR #9746, opened 2024-08-01
test(tpcds): bump scale factor to 0.45

cpcloud:
Up the scale factor of the TPC-DS data. This found a quite a number of bugs in the existing queries.

### DIFF
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -950,7 +950,8 @@ def test_12(web_sales, item, date_dim):
         .mutate(
             revenueratio=_.itemrevenue
             * 100.000
-            / _.itemrevenue.sum().over(group_by=_.i_class)
+            # snowflake divide by zero is an error, all others return a value (null, nan, or inf)
+            / _.itemrevenue.sum().over(group_by=_.i_class).nullif(0.0)
         )
         .order_by(_.i_category, _.i_class, _.i_item_id, _.i_item_desc, "revenueratio")
         .limit(100)
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -2033,7 +2033,7 @@ def test_41(item):
             _.i_manufact_id.between(738, 738 + 40),
             lambda i1: item.filter(
                 lambda s: (
-                    (i1.i_manufact_id == s.i_manufact_id)
+                    (i1.i_manufact == s.i_manufact)
                     & (
                         (
                             (s.i_category == "Women")
@@ -2062,7 +2062,7 @@ def test_41(item):
                     )
                 )
                 | (
-                    (i1.i_manufact_id == s.i_manufact_id)
+                    (i1.i_manufact == s.i_manufact)
                     & (
                         (
                             (s.i_category == "Women")
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -2473,16 +2473,16 @@ def test_49(
 
     return (
         in_web.mutate(
-            return_rank=rank().over(order_by=_.return_ratio),
-            currency_rank=rank().over(order_by=_.currency_ratio),
+            return_rank=rank().over(order_by=_.return_ratio) + 1,
+            currency_rank=rank().over(order_by=_.currency_ratio) + 1,
         )
         .filter((_.return_rank <= 10) | (_.currency_rank <= 10))
         .mutate(channel=lit("web"))
         .relocate("channel", before="item")
         .union(
             in_cat.mutate(
-                return_rank=rank().over(order_by=_.return_ratio),
-                currency_rank=rank().over(order_by=_.currency_ratio),
+                return_rank=rank().over(order_by=_.return_ratio) + 1,
+                currency_rank=rank().over(order_by=_.currency_ratio) + 1,
             )
             .filter((_.return_rank <= 10) | (_.currency_rank <= 10))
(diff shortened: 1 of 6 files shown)


gforsyth:
Seems like both clickhouse and trino have ordering issues with query 49?

```
E             Index | Obtained | Expected      
E             2     | 1601     | 3483 ± 3.5e-03
E             3     | 3483     | 1601 ± 1.6e-03
E             4     | 3281     | 6259 ± 6.3e-03
E             5     | 6259     | 3281 ± 3.3e-03
```

cpcloud:
Yep, there's something funky with 49's `rank` call. Not yet sure whether the issue is related to Ibis or the engine.

To make life easy 😒, 3 out of 5 backends do not fail this query.

Passing:

- DuckDB
- DataFusion
- Snowflake

Failing:

- Trino
- ClickHouse

cpcloud:
I think the `ORDER BY` in the query is not unique, so the two values that don't match are simply swapped:

```
E             4     | 3281     | 6259 ± 6.3e-03
E             5     | 6259     | 3281 ± 3.3e-03
```

cpcloud:
Good thing query 49 isn't huge and complicated!

cpcloud:
Wat, now duckdb is failing 😂

cpcloud:
... and then passing ...

so yeah it's some kind of non-determinism.

gforsyth:
> so yeah it's some kind of non-determinism.

yeah, and one of the trino jobs is passing, but not the other.

ibis/backends/tests/tpc/ds/test_queries.py
```diff
@@ -2498,14 +2509,14 @@ def test_49(
...
         )
         .order_by(
-            _.channel.asc(nulls_first=True),
-            _.return_rank.asc(nulls_first=True),
-            _.currency_rank.asc(nulls_first=True),
-            _.item.asc(nulls_first=True),
+            _[0].asc(nulls_first=True),
+            _[3].asc(nulls_first=True),
+            _[4].asc(nulls_first=True),
+            _[2].asc(nulls_first=True),
```
cpcloud, line 2519: This is causing the non-determinism :)

It should be `_[1]` since we're 0-indexed and SQL column numbers are 1-indexed.

🥲

cpcloud: ClickHouse still fails but for a different reason. It seems to compute `rank() over()` differently than everything because of course it does!

cpcloud:
Ok, it turns out that our default window behavior doesn't match the default SQL behavior, but only for ClickHouse 😒

cpcloud:
We should definitely square the circle of the meaning of default window frames in Ibis as part of API stability. We might have to break some stuff, but it'll be for the best.

ibis/backends/tests/tpc/ds/test_queries.py
```diff
@@ -942,7 +957,8 @@ def test_12(web_sales, item, date_dim):
         .mutate(
             revenueratio=_.itemrevenue
             * 100.000
-            / _.itemrevenue.sum().over(group_by=_.i_class)
+            # snowflake divide by zero is an error, all others return a value (null, nan, or inf)
```
cpcloud, line 960: Do we want to unify this behavior?

I think there's only one option given the support matrix for nan and infinity looks like Swiss cheese, and that's to return `NULL` on divide by zero.

Errors are out, because they'd require looking at the value (or using some kind of database error function, which is also not well-supported).

cpcloud: Maybe an API that's explicit about this would be better, like `t.x.safe_div(t.y)` naming TBD…

cpcloud: Hm, I think this might be a window `range` issue as well. Looking…

cpcloud: Nope, divide by zero still, even when being explicit about the `range=(None, 0)` bits.

cpcloud: Hilarious, sqlglot is inserting a `NULLIF(..., 0)` during the transpile:

```
(Pdb) sg.parse_one('select x / y', read='duckdb').sql('snowflake')
'SELECT x / NULLIF(y, 0)'
```

ibis/backends/tests/tpc/ds/test_queries.py
gforsyth: wow (all of) that's frustrating

ibis/backends/tests/tpc/ds/test_queries.py
cpcloud: Hm, wonder why we aren't picking that up in the snowflake compiler. Looking into it.

gforsyth: I think returning NULL for a divide-by-zero is reasonable?

cpcloud: Ah, because sqlglot has knowledge of what the origin dialect was, so it knows that to emulate the same behavior it has to make the `Div` operator as `safe=True` 🔍

cpcloud:
Ok to summarize:

- Rollup is still needed
- No bug fixes to Ibis itself are required (yet!), still, 63 minus the 7 or so that needed rollup is pretty good.
- Allowing empty results is still highly suspicious 🤨, we probably want to try to get of those entirely.
- We want to make division safe for all backends
- We need to do make some (possibly breaking) changes and ensure that window frame semantics match across backends where possible.

cpcloud:
@gforsyth This is ready for a review, things are 🟢

cpcloud:
Lemme run clouds too

PR #9748, opened 2024-08-01
fix(sql): only return tables in `current_database`

### LINKED ISSUE #9686: bug(postgres): cross-schema column mixing
### What happened?

On `con.table("accounts").to_pandas()` I receive an error `ValueError: 10 columns passed, passed data had 5 columns`. I believe ibis is picking up columns from an `accounts` table in a different schema (`raw`).

I expected the default `con.table("accounts")` to only pick columns from the default `accounts` table (`public` schema/database).

I can avoid this by specifying `database`, but figured I'd post this ticket in case this is unintended behavior.

```sql
CREATE TABLE accounts (
    account_id INTEGER PRIMARY KEY NOT NULL UNIQUE
    ,account_name TEXT NOT NULL
    ,account_timezone TEXT
    ,account_notes TEXT
    ,account_active BOOLEAN NOT NULL
);

CREATE TABLE IF NOT EXISTS raw.accounts (
    id TEXT
    ,name TEXT
    ,timezone TEXT
    ,notes TEXT
    ,active TEXT
);
```

```python
sc.con.table("accounts")
Out[88]: 
DatabaseTable: accounts
  account_id       !int32
  id               string
  account_name     !string
  name             string
  account_timezone string
  timezone         string
  notes            string
  account_notes    string
  account_active   !boolean
  active           string
```

Also, sorry for the raw_sql. (:

9.1

Postgres

### Relevant log output

```sh
sc.con.table("accounts")
Out[93]: 
DatabaseTable: accounts
  account_id       !int32
  id               string
  account_name     !string
  name             string
  account_timezone string
  timezone         string
  notes            string
  account_notes    string
  account_active   !boolean
  active           string

sc.con.raw_sql("DROP TABLE accounts")
Out[94]: <cursor object at 0x71512acb55d0; closed: 0>

sc.con.table("accounts")
Out[95]: 
DatabaseTable: accounts
  id       string
  name     string
  timezone string
  notes    string
  active   string

sc.con.table("accounts", schema="public")
/home/patrick/anaconda3/envs/.../lib/python3.12/site-packages/ibis/backends/sql/__init__.py:36: FutureWarning: `schema` is deprecated as of v9.0, removed in v10.0; Use the `database` kwarg with one of the following patterns:
database="database"
database=("catalog", "database")
database="catalog.database"
  util.warn_deprecated(

  Cell In[96], line 1
    sc.con.table("accounts", schema="public")

  File ~/anaconda3/envs/.../lib/python3.12/site-packages/ibis/backends/sql/__init__.py:137 in table
    table_schema = self.get_schema(name, catalog=catalog, database=database)

  File ~/anaconda3/envs/.../lib/python3.12/site-packages/ibis/backends/postgres/__init__.py:600 in get_schema
    raise com.IbisError(f"Table not found: {name!r}")

IbisError: Table not found: 'accounts'


sc.con.raw_sql("""
CREATE TABLE accounts (
    account_id INTEGER PRIMARY KEY NOT NULL UNIQUE
    ,account_name TEXT NOT NULL
    ,account_timezone TEXT
    ,account_notes TEXT
    ,account_active BOOLEAN NOT NULL
);
""")
Out[97]: <cursor object at 0x71512acb4e50; closed: 0>

sc.con.table("accounts", database="public")
Out[99]: 
DatabaseTable: public.accounts
  account_id       !int32
  account_name     !string
  account_timezone string
  account_notes    string
  account_active   !boolean

sc.con.table("accounts")
Out[100]: 
DatabaseTable: accounts
  account_id       !int32
  id               string
  account_name     !string
  name             string
  account_timezone string
  timezone         string
  notes            string
  account_notes    string
  account_active   !boolean
  active           string
```

### Code of Conduct

gforsyth:
## Description of changes

This started as a fix of #9686 but it turned up a few issues in other backends.

The short version is: when we call `get_schema` without specifying a
`catalog.database` prefix, we get tables that exist in many (sometimes all) the
databases on a given backend (that match the name passed to `get_schema`).

Postgres has things like `search_path` to help with this, but the tables on `search_path` are slightly different than what's returned by `\dt` (just to make this extra fun).

What's implemented here (and what I'm proposing as a general standard) is:

When you ask for a table named `foo`, we will look through the
`current_database` and wherever temp tables are stored for a match and return
the schema of that table.

My idea here is that anything that shows up in the output of `list_tables` (or
`con.tables`) should be accessible via `con.table` without additional arguments.

There's at least one edge-case which I discovered while writing this
description, which is if the temp table has the same name as a table in the
`current_database`, the table returned will be the interleaved schema of the two
tables, which is terrible.

I'll fix that, too.

## Issues closed

* Fixes #9686

## Todo List of things to fix (not all in this PR):

* [ ] decide on whether `temp` tables take precedence over concrete tables or vice-versa
* [x] add spark, datafusion and clickhouse exceptions to test
* [x] add cross-database table creation to mysql
* [x] fix mssql table collection
* [x] fix trino table collection
* [x] separate out risingwave `get_schema` because it doesn't have postgres-specific function
* [x] confirm that exasol can't create tables in other databases

### DIFF
--- a/docker/mysql/startup.sql
+++ b/docker/mysql/startup.sql
@@ -1,5 +1,4 @@
 CREATE USER 'ibis'@'localhost' IDENTIFIED BY 'ibis';
 CREATE SCHEMA IF NOT EXISTS test_schema;
-GRANT CREATE, DROP ON *.* TO 'ibis'@'%';
-GRANT CREATE,SELECT,DROP ON `test_schema`.* TO 'ibis'@'%';
+GRANT CREATE,SELECT,DROP ON *.* TO 'ibis'@'%';
 FLUSH PRIVILEGES;
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -500,7 +500,7 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
         query = (
             sg.select(f.anon.unnest(f.list_append(C.aliases, C.extension_name)))
             .from_(f.duckdb_extensions())
-            .where(sg.and_(C.installed, C.loaded))
+            .where(C.installed, C.loaded)
         )
         with self._safe_raw_sql(query) as cur:
             installed = map(itemgetter(0), cur.fetchall())
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -341,11 +341,23 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
         """
         conditions = [sg.column("table_name").eq(sge.convert(table_name))]
 
-        if catalog is not None:
-            conditions.append(sg.column("table_catalog").eq(sge.convert(catalog)))
+        # If no catalog.database is specified, assume the current ones
+        db = database or self.current_database
+        cat = catalog or self.current_catalog
 
-        if database is not None:
-            conditions.append(sg.column("table_schema").eq(sge.convert(database)))
+        # If a catalog isn't specified, then include temp tables in the
+        # returned values
+        # temp tables live in the `temp` catalog
+        # We assume that we should return temp tables that match the database
+        # specified
+        if catalog is None:
+            conditions.append(
+                sg.column("table_catalog").isin(sge.convert(cat), sge.convert("temp"))
+            )
+        else:
+            conditions.append(sg.column("table_catalog").eq(sge.convert(cat)))
+
+        conditions.append(sg.column("table_schema").eq(sge.convert(db)))
 
         query = (
             sg.select(
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -337,53 +337,29 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
         -------
         sch.Schema
             Ibis schema
-
         """
-        conditions = [sg.column("table_name").eq(sge.convert(table_name))]
-
-        # If no catalog.database is specified, assume the current ones
-        db = database or self.current_database
-        cat = catalog or self.current_catalog
-
-        # If a catalog isn't specified, then include temp tables in the
-        # returned values
-        # temp tables live in the `temp` catalog
-        # We assume that we should return temp tables that match the database
-        # specified
-        if catalog is None:
-            conditions.append(
-                sg.column("table_catalog").isin(sge.convert(cat), sge.convert("temp"))
+        query = sge.Describe(
+            this=sg.table(
+                table_name, db=database, catalog=catalog, quoted=self.compiler.quoted
             )
-        else:
-            conditions.append(sg.column("table_catalog").eq(sge.convert(cat)))
+        ).sql(self.dialect)
 
-        conditions.append(sg.column("table_schema").eq(sge.convert(db)))
-
-        query = (
-            sg.select(
-                "column_name",
-                "data_type",
-                sg.column("is_nullable").eq(sge.convert("YES")).as_("nullable"),
-            )
-            .from_(sg.table("columns", db="information_schema"))
-            .where(sg.and_(*conditions))
-            .order_by("ordinal_position")
-        )
-
-        with self._safe_raw_sql(query) as cur:
-            meta = cur.fetch_arrow_table()
-
-        if not meta:
+        try:
+            result = self.con.sql(query)
+        except duckdb.CatalogException:
             raise exc.IbisError(f"Table not found: {table_name!r}")
+        else:
+            meta = result.fetch_arrow_table()
 
         names = meta["column_name"].to_pylist()
-        types = meta["data_type"].to_pylist()
-        nullables = meta["nullable"].to_pylist()
+        types = meta["column_type"].to_pylist()
+        nullables = meta["null"].to_pylist()
 
+        type_mapper = self.compiler.type_mapper
         return sch.Schema(
             {
-                name: self.compiler.type_mapper.from_string(typ, nullable=nullable)
-                for name, typ, nullable in zip(names, types, nullables)
+                name: type_mapper.from_string(typ, nullable=null == "YES")
+                for name, typ, null in zip(names, types, nullables)
             }
         )
 
--- a/ibis/backends/risingwave/__init__.py
+++ b/ibis/backends/risingwave/__init__.py
@@ -586,3 +586,9 @@ class Backend(PostgresBackend):
         )
         with self._safe_raw_sql(src):
             pass
+
+    @property
+    def _session_temp_db(self) -> str | None:
+        # Return `None`, because RisingWave does not implement temp tables like
+        # Postgres
+        return None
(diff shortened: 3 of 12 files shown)


ibis/backends/tests/test_client.py
```diff
@@ -1629,3 +1629,34 @@ def test_from_connection(con, top_level):
     new_con = backend.from_connection(getattr(con, CON_ATTR.get(con.name, "con")))
     result = int(new_con.execute(ibis.literal(1, type="int")))
     assert result == 1
+
+
+def test_no_accidental_cross_database_table_load(con_create_database):
```
cpcloud, line 1634: It's amazing that you were able to write a cross backend test for this.

I have a test case that I will push up for DuckDB that involves `attach`, to test the robustness of the implementation across catalogs that also contain an overlapping table name.

cpcloud:
It looks like `temp` takes precedence if the names collide, at least in DuckDB:

```
D create table t (x int);
D create temp table t (y int);
D table t;
┌────────┐
│   y    │
│ int32  │
├────────┤
│ 0 rows │
└────────┘
```

ibis/backends/duckdb/__init__.py
```diff
@@ -341,11 +341,23 @@ def get_schema(
...
+        # returned values
+        # temp tables live in the `temp` catalog
+        # We assume that we should return temp tables that match the database
+        # specified
+        if catalog is None:
+            conditions.append(
+                sg.column("table_catalog").isin(sge.convert(cat), sge.convert("temp"))
+            )
+        else:
+            conditions.append(sg.column("table_catalog").eq(sge.convert(cat)))
```
cpcloud, line 358: Given this part of the PR description:

> When you ask for a table named foo, we will look through the current_database and wherever temp tables are stored for a match and return the schema of that table.

Shouldn't the implementation be this?

```suggestion
        conditions.append(
            sg.column("table_catalog").isin(sge.convert(cat), sge.convert("temp"))
        )
```

That is, always return temp tables?

Maybe I'm misreading and the fact that catalog is specified changes the desired output.

cpcloud:
Pushed up the additional test. It passed without changes 🥳

gforsyth:
> It looks like `temp` takes precedence if the names collide, at least in DuckDB:

Ok, I guess we need to decide on what convention we want to enforce?

ibis/backends/duckdb/__init__.py
gforsyth: This is going to have to be rewritten anyway, to only return tables from `temp` (or `main`) when there's a name collision

gforsyth:
lol, I'm going to have to relax the cross-backend exception check a fair bit until we get Naty's `TableNotFound` PR in, but I think this should go in before 10.0

gforsyth:
ok, `datafusion` and `clickhouse` are failing b/c the exception doesn't match.

MSSQL is failing the way postgres and duckdb did before the fix.

gforsyth:
A hilarious error message from MySQL:
```
        if database is not None and database != self.current_database:
>           raise com.UnsupportedOperationError(
                "Creating tables in other databases is not supported by Postgres"
            )
E           ibis.common.exceptions.UnsupportedOperationError: Creating tables in other databases is not supported by Postgres
```

Need to fix that, too...

gforsyth:
Ok, this turned into a bit of a nightmare.  I'll pick it back up tomorrow or Monday

ibis/backends/duckdb/__init__.py
cpcloud: Cleaned up this up to use `DESCRIBE catalog.database.table`, which will give the semantics of temp tables -> non temp tables.

docker/mysql/startup.sql
```diff
@@ -1,5 +1,4 @@
 CREATE USER 'ibis'@'localhost' IDENTIFIED BY 'ibis';
 CREATE SCHEMA IF NOT EXISTS test_schema;
-GRANT CREATE, DROP ON *.* TO 'ibis'@'%';
-GRANT CREATE,SELECT,DROP ON `test_schema`.* TO 'ibis'@'%';
+GRANT CREATE,SELECT,DROP ON *.* TO 'ibis'@'%';
```
cpcloud, line 3: Necessary to allow the ibis user to operate on newly created databases.

ibis/backends/duckdb/__init__.py
```diff
@@ -337,41 +337,29 @@ def get_schema(
...
-
-        if database is not None:
-            conditions.append(sg.column("table_schema").eq(sge.convert(database)))
-
-        query = (
-            sg.select(
-                "column_name",
-                "data_type",
-                sg.column("is_nullable").eq(sge.convert("YES")).as_("nullable"),
+        query = sge.Describe(
```
cpcloud, line 341: `DESCRIBE` in DuckDB used to be different than using `information_schema`, but only for CSV files, and that case has been fixed in a version that we no longer support.

ibis/backends/risingwave/__init__.py
```diff
@@ -586,3 +586,9 @@ def drop_sink(
         )
         with self._safe_raw_sql(src):
             pass
+
+    @property
+    def _session_temp_db(self) -> str | None:
```
cpcloud, line 591: This was actually enough to make tests pass, since otherwise `get_schema` works just fine.

ibis/backends/trino/__init__.py
```diff
@@ -139,20 +139,18 @@ def get_schema(
...
         query = (
             sg.select(
-                "column_name",
-                "data_type",
-                sg.column("is_nullable").eq(sge.convert("YES")).as_("nullable"),
+                C.column_name,
+                C.data_type,
+                C.is_nullable.eq(sge.convert("YES")).as_("nullable"),
             )
             .from_(sg.table("columns", db="information_schema", catalog=catalog))
```
cpcloud, line 148: I wish we could use `DESCRIBE` here, but it doesn't contain a nullability column 😞

cpcloud:
Exasol does indeed allow creating tables in other databases, but there was a bug in our code (we weren't quoting identifiers).

cpcloud:
The only remaining task is to decide how we want to scope temp tables versus non-temp tables in backends that support them. We can defer that to a follow-up IMO.

cpcloud:
BigQuery and Snowflake are failing because the test assumes the backend can differentiate between int32 and int64, which isn't true for both BigQuery and Snowflake. All integers are at least int64 for both, Snowflake is actually int128 (by way of using `decimal(38, 0)`).

cpcloud:
I adjusted the test to use `int64` and `string` because of the issue described in the previous comment.

gforsyth:
I raised https://github.com/apache/datafusion-python/issues/796 in re: a better exception for missing tables, at least

ibis/backends/tests/test_client.py
```diff
@@ -1629,3 +1624,60 @@ def test_from_connection(con, top_level):
...
+                GoogleNotFound,
+            ),
+        ),
+        # datafusion really needs to get their exception story in order
+        #
+        # we only want to allow base Exception when we're testing datafusion
+        # otherwise any exceptions, including those that are unrelated to the
+        # problem under test will be considered correctly raising
+        *((Exception,) * (con.name == "datafusion")),
+    )
```
gforsyth, line 1672: gnarly

gforsyth (commented):
Awesome -- thanks for carrying this one over the line.

Oh, lol, I'm still the PR author and cannot approve

ibis/backends/tests/test_client.py
cpcloud: I can make this a bit less nested if that would help

cpcloud: Unraveled it a bit

gforsyth (commented):
Looks good to me!

cpcloud (approved):

PR #9753, opened 2024-08-02
feat(pyspark): add support for pyarrow and python UDFs

### LINKED ISSUE #9074: feat: support pyarrow UDFs for pyspark backend

Pyspark now supports [Arrow UDFs](https://spark.apache.org/docs/latest/api/python/user_guide/sql/arrow_pandas.html#arrow-python-udfs) that facilitate efficient row-by-row executions using Arrow as a backend e.g.

```python
import pandas as pd
from pyspark.sql.functions import udf

@udf(returnType="int",useArrow=True)
def add_one(x:int) -> int:
    return x + 1

#Create column using pyarrow-udf
df = pd.DataFrame({"a":[1,2,3]})
dfs = spark.createDataFrame(df)
dfs.withColumn("b", add_one("a")).show()
```

However, the equivalent function using `ibis` raises a NotImplementedError, because only Pandas-based vectorized UDFs are supported

```python
import ibis
from ibis import _

@ibis.udf.scalar.pyarrow
def add_one_pyarrow(x:int) -> int:
    return x + 1

@ibis.udf.scalar.pandas
def add_one_pandas(x:int) -> int:
    return x + 1

con = ibis.pyspark.connect(spark)
con.create_table("df", df, format="delta", overwrite=True)

table = con.table("df")
table.mutate(b=add_one_pandas(_.a)).execute()
table.mutate(b=add_one_pyarrow(_.a)).execute() #raises NotImpletmentedError
```

Pandas-based UDFs are not supported for the DuckDB backend, but Arrow-based ones are. For my use case, I would like to ensure parity between using either backend as much as possible, so being able to use Arrow-based UDFs on a pyspark table would be very useful

I'd like a solution that would allow me to use an Arrow-based UDF on a pyspark table

9.0.0.dev686

Pyspark

### Code of Conduct

jstammers:
## Description of changes

This PR modifies the pyspark Backend so that it can support UDFs implemented in pure python or pyarrow by making use of the `pyspark.sql.udf` wrapper.

To successfully run the unit tests for a pyarrow UDF, I needed to ensure that the spark worker executed the correct python interpreter by setting
```bash
export PYSPARK_PYTHON=$(which python)
```
within my nix shell. This might require some further configuration of the nix environment in order to pass during the CI tests.

## Issues closed

* Resolves #9074

### DIFF
--- a/ibis/backends/pyspark/tests/test_udf.py
+++ b/ibis/backends/pyspark/tests/test_udf.py
@@ -45,14 +45,14 @@ def test_python_udf(t, df):
     tm.assert_frame_equal(result, expected)
 
 
-@pytest.mark.skipif(PYSPARK_LT_35, reason="pyarrow UDFs require PySpark 3.5+")
+@pytest.mark.xfail(PYSPARK_LT_35, reason="pyarrow UDFs require PySpark 3.5+")
 def test_pyarrow_udf(t, df):
     result = t.mutate(repeated=pyarrow_repeat(t.str_col, 2)).execute()
     expected = df.assign(repeated=df.str_col * 2)
     tm.assert_frame_equal(result, expected)
 
 
-@pytest.mark.skipif(not PYSPARK_LT_35, reason="pyarrow UDFs require PySpark 3.5+")
+@pytest.mark.xfail(not PYSPARK_LT_35, reason="pyarrow UDFs require PySpark 3.5+")
 def test_illegal_udf_type(t):
     @ibis.udf.scalar.pyarrow
     def my_add_one(x) -> str:
--- a/ibis/backends/pyspark/tests/test_udf.py
+++ b/ibis/backends/pyspark/tests/test_udf.py
@@ -4,6 +4,8 @@ import pandas.testing as tm
 import pytest
 
 import ibis
+from ibis.expr.tests.snapshots.test_sql.test_parse_sql_aggregation_with_multiple_joins.decompiled import \
+    result
 
 pytest.importorskip("pyspark")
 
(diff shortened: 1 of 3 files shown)


cpcloud:
```bash
export PYSPARK_PYTHON=$(which python)
```

🤔 interesting...

We don't run the full backend test suite through nix (only the backends that can be run locally with the exception of PySpark). You might have figured out the way we can now do that. I'll give it a try 🥳

cpcloud:
It looks like you might need to xfail the UDF tests for the earlier version of PySpark that we test against (3.3.3).

cpcloud:
Sweet, setting `PYSPARK_PYTHON` does allow pyspark to work in the nix environment!

I'll put up a separate PR to add that to our `shellHook` so everyone using nix can pick it up.

jstammers:
Thanks for that spot @cpcloud - I'd mistakenly assumed that it was implemented in a previous version. The pyspark docs weren't exactly clear on that. I've pushed some changes that should hopefully check that the pyarrow UDFs fail for pyspark < 3.5.

ibis/backends/pyspark/tests/test_udf.py
```diff
@@ -22,12 +23,36 @@ def df(con):
...
     tm.assert_frame_equal(result, expected)
 
 
+def test_python_udf(t, df):
+    result = t.mutate(repeated=py_repeat(t.str_col, 2)).execute()
+    expected = df.assign(repeated=df.str_col * 2)
+    tm.assert_frame_equal(result, expected)
+
+
+@pytest.mark.skipif(PYSPARK_LT_35, reason="pyarrow UDFs require PySpark 3.5+")
```
cpcloud, line 48:
```suggestion
@pytest.mark.xfail(PYSPARK_LT_35, reason="pyarrow UDFs require PySpark 3.5+")
```

ibis/backends/pyspark/tests/test_udf.py
```diff
@@ -22,12 +23,36 @@ def df(con):
...
+def test_pyarrow_udf(t, df):
+    result = t.mutate(repeated=pyarrow_repeat(t.str_col, 2)).execute()
+    expected = df.assign(repeated=df.str_col * 2)
+    tm.assert_frame_equal(result, expected)
+
+
+@pytest.mark.skipif(not PYSPARK_LT_35, reason="pyarrow UDFs require PySpark 3.5+")
```
cpcloud, line 55:
```suggestion
@pytest.mark.xfail(not PYSPARK_LT_35, reason="pyarrow UDFs require PySpark 3.5+")
```

ibis/backends/tests/test_udf.py
```diff
@@ -174,10 +184,11 @@ def add_one_pyarrow(s: int) -> int:  # s is series, int is the element type
             add_one_pyarrow,
             marks=[
                 mark.notyet(
-                    ["snowflake", "sqlite", "pyspark", "flink"],
+                    ["snowflake", "sqlite", "flink"],
                     raises=NotImplementedError,
                     reason="backend doesn't support pyarrow UDFs",
-                )
+                ),
+                pyspark_version_mismatch,
```
cpcloud, line 191: We actually have as special marker for this situation, there are plenty of other backends where this kind of problem comes up :)

```suggestion
                mark.xfail_version(pyspark=["pyspark<3.5"]),
```

ibis/backends/tests/test_udf.py
```diff
@@ -6,6 +6,12 @@
 
 import ibis.common.exceptions as com
 from ibis import _, udf
+
+try:
+    from ibis.backends.pyspark import PYSPARK_LT_35
+except ImportError:
+    # Workaround when pyspark is not installed
+    PYSPARK_LT_35 = True
```
cpcloud, line 14:
```suggestion
```

Should be able to git rid of these if you use the `xfail_version` marker as suggested below.

ibis/backends/tests/test_udf.py
```diff
@@ -31,10 +37,16 @@
     reason="Docker image has Python 3.10, results in `cloudpickle` version mismatch",
 )
 
+pyspark_version_mismatch = mark.notimpl(
+    ["pyspark"],
+    condition=PYSPARK_LT_35,
+    raises=NotImplementedError,
+    reason="pyarrow UDFs require PySpark 3.5+",
+)
+
```
cpcloud, line 46:
```suggestion
```

Should be able to git rid of these if you use the `xfail_version` marker as suggested below.

cpcloud (approved):
Looks good! I will commit my suggested changes and the push up any fixes needed!

PR #9755, opened 2024-08-02
fix(compiler-internals): define unsupported operations after simple operations

cpcloud:
This avoids the case of a base class `SIMPLE_OPS` dict overriding the `UNSUPPORTED_OPS` list.

### DIFF
--- a/ibis/backends/sql/compilers/risingwave.py
+++ b/ibis/backends/sql/compilers/risingwave.py
@@ -32,11 +32,6 @@ class RisingWaveCompiler(PostgresCompiler):
         ),
     )
 
-    SIMPLE_OPS = {
-        ops.First: "first_value",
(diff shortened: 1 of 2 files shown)


ibis/backends/sql/compilers/risingwave.py
```diff
@@ -49,6 +44,12 @@ def visit_Correlation(self, op, *, left, right, how, where):
             op, left=left, right=right, how=how, where=where
         )
 
+    def visit_First(self, op, *, arg, where):
+        return self.agg.first_value(arg, where=where)
```
cpcloud, line 48: This is implemented this way as a test case for the definition order, but it turns out to be needed in #9729.

cpcloud:
Merging this, so we can wrap up #9729

PR #9766, opened 2024-08-04
refactor(sql): make compilers usable with a base install

cpcloud:
This PR makes our compilers usable without having any backend dependencies installed.

The main goal is to make `ibis.to_sql` avoid hitting any backend dependencies.

The approach here is to move all sqlglot-ifcation, including

- transforms related to output, like geospatial (e.g., duckdb) an boolean workarounds (e.g., mssql)
- UDF compilation

entirely into the compilers.

One benefit of this is that now the output of `ibis.to_sql` and the output of `con.compile` will actually match. Previously, they would only match when the expression was attached to a backend.


ibis/backends/bigquery/tests/system/test_client.py
```diff
@@ -226,17 +224,6 @@ def test_exists_table_different_project(con):
     assert "foobar" not in con.list_tables(database=dataset)
 
 
-def test_multiple_project_queries(con, snapshot):
```
cpcloud, line 229: This test was pretty useless given the execution test immediately below it.

ibis/backends/sql/compilers/bigquery/__init__.py
```diff
@@ -21,9 +23,81 @@
...
+
+    BigQuery has only partial support for NULL FIRST/LAST in RANGE windows so
+    we remove it from any window frame clause that doesn't support it.
+
+    Here's the support matrix:
+
+    ✅ sum(x) over (order by y desc nulls last)
+    🚫 sum(x) over (order by y asc nulls last)
+    ✅ sum(x) over (order by y asc nulls first)
+    🚫 sum(x) over (order by y desc nulls first)
```
jcrist, line 64: Unicode in comments is not something I would have thought of, but is kinda nice.

jcrist (approved):

PR #9776, opened 2024-08-06
feat(duckdb): add support for passing a subset of column types to `read_csv`

cpcloud:
Add support for the `types` argument to `read_csv` in the DuckDB backend.


gforsyth (approved):
Nice.

PR #9778, opened 2024-08-06
feat(api): support `include_null` arg in `collect`/`first`/`last` aggregations

### LINKED ISSUE #9703: consider: option to include NULLs in Array.collect()?
### What happened?

In https://github.com/ibis-project/ibis/issues/9313 we unified behavior across backends and made it so Array.collect() excluded NULLs

This behavior change broke [this util function of mine](My function that relies on this property is [here](https://github.com/NickCrews/mismo/blob/fd2b260c187da3197839074812b1768d9dbd9fc8/mismo/arrays/_array.py#L35-L76).

This was due to my reliance on the previously-true-on-duckdb property that `unnest()` and `collect()` were inverses (minus the fact that `unnest([]) -> no row` and `unnest(NULL) -> no row`, so in order for `x.unnest().collect()` to get back to x I needed a little post-process massaging). I hadn't realized that my implementation there wasn't fully portable to other backends, this issue is informative, thank you!

Do you see another way to accomplish what I'm trying to do there that can sidestep this behavior change? Or could we consider adding a `exclude_nulls=True` option?

Regardless, would you like a PR that documents the intended behavior in .collect()s docstring? It was difficult for me to find this issue, I thought this behavior might have just been accidental.

main

duckdb, though I ideally want my code as linked above to be portable to pyspark, and pyspark always exludes NULLs from list_agg(), so an entierly new implementation might be the only way to really solve this...

### Relevant log output

_No response_

### Code of Conduct

jcrist:
This adds an `ignore_null` keyword argument to `collect`/`first`/`last` aggregations, defaulting to `True`. If set to `False`, these aggregations will include null values when processing (e.g. `collect` will return an array with all values, not just the non-null ones). Note that `ignore_null=False` isn't supported by all backends, while `ignore_null=True` (the default) is.

I still have a follow-up TODO to propagate this option to the window-function versions of `first`/`last`, but I'd like to do that in a separate PR.

Fixes #9703.


ibis/backends/sql/compilers/base.py
```diff
@@ -331,15 +330,13 @@ class SQLGlotCompiler(abc.ABC):
...
         ops.Exp: "exp",
-        ops.First: "first",
         FirstValue: "first_value",
         ops.GroupConcat: "group_concat",
         ops.IfElse: "if",
         ops.IsInf: "isinf",
         ops.IsNan: "isnan",
         ops.JSONGetItem: "json_extract",
         ops.LPad: "lpad",
-        ops.Last: "last",
```
jcrist, line 342: Every backend that supports these ops overrides the builtin definitions. As such, it made more sense to me to drop default support, letting the backends that don't support these ops avoid adding them to the `UNSUPPORTED_OPS` set.

ibis/backends/sql/compilers/clickhouse.py
```diff
@@ -604,6 +601,27 @@ def visit_ArrayUnion(self, op, *, left, right):
     def visit_ArrayZip(self, op: ops.ArrayZip, *, arg, **_: Any) -> str:
         return self.f.arrayZip(*arg)
 
+    def visit_ArrayCollect(self, op, *, arg, where, order_by, ignore_null):
+        if not ignore_null:
+            raise com.UnsupportedOperationError(
+                "`ignore_null=False` is not supported by the clickhouse backend"
```
jcrist, line 607: The duplication of this code everywhere is a bit annoying. This _could_ be done in a generic rewrite rule (that doesn't rewrite but just errors if the `ignore_null` is False), but honestly that'd only save a few lines of code (there's only a few backends which do support these aggregations but don't support this option) and increase coupling - duplicating seemed cleaner IMO.

jcrist:
Ok, we're down to one test failure, but it's a frustrating one.

To support the `where` arg in aggregations for backends that lack a `filter` syntax, we typically compile these to something like `SOME_AGG(IFF(cond, column, NULL))` - since aggregations ignore nulls (by default), replacing the skipped values with `NULL` results in equivalent expression. However, if you set `ignore_null=False`, then this behavior is no longer correct. To work around this, I think we'll need to raise for backends like this if `where` is present and `ignore_null=False` (currently far only bigquery and pyspark have this issue).

jcrist:
Ok, this is ready for review. Re: keyword naming, I went with `ignore_null` since that's what people seemed to like in #9703. Happy to change to `ignore_nulls` (plural) instead if people think that reads cleaner. No strong thoughts.

ncclementi:
> Happy to change to ignore_nulls (plural) instead if people think that reads cleaner. No strong thoughts.
In https://github.com/ibis-project/ibis/pull/9300 we also went for the singular version `fill_null` and `drop_null` so we might want to keep it all singular.

cpcloud:
Is there any chance we can rename this to `keep_null` to avoid the double negative?

I personally find `ignore_null=False` not immediately obvious, because I have to think about what "don't ignore" means, which IMO is much easier to understand if we replace "don't ignore" with "keep".

Another argument would be that `keep_null` is more obvious for people who actually want to alter the default value.

Regarding pluralization, maintaining internal consistency within Ibis seems preferable, and if we go with `keep_null` then we probably won't need to field questions about "why did we follow the SQL naming convention but not include the **s**?"

jcrist:
I agree about the double negative.

My only strong stance here is that we use the same keyword name everywhere, even if the default is `True` for some ops and `False` for other ops. If a method has configurable null handling, we should have the same keyword arg name for predictability. I also have a _slight_ preference for a plural kwarg name (polars, spark, and every user issue asking for this has used the `ignore_nulls` name). I recognize that we use singular names in our methods, but I don't think having plural names for keyword args is inconsistent.

How about `include_nulls` or `include_null` instead of `keep_null`?  cc @gforsyth as well for naming thoughts. I'd like to only rename things in this PR once :).

cpcloud:
My only somewhat strong opinion here is to avoid the double negative. Slight preference for pluralization matching the method name, but I certainly won't block on that.

gforsyth:
I'm fine with both `include_` and `keep_` as prefixes.  I also have a preference for plurals in kwargs -- I am still trying to figure out why, but I do.  We need a linguist.

jcrist:
I've renamed this to `include_null` - I've waffled a bit on the pluralization, kept it singular for consistency.

ibis/backends/sql/compilers/pyspark.py
```diff
@@ -240,11 +240,27 @@ def visit_FirstValue(self, op, *, arg):
     def visit_LastValue(self, op, *, arg):
         return sge.IgnoreNulls(this=self.f.last(arg))
 
-    def visit_First(self, op, *, arg, where, order_by):
-        return sge.IgnoreNulls(this=self.agg.first(arg, where=where, order_by=order_by))
+    def visit_First(self, op, *, arg, where, order_by, include_null):
+        if where is not None and include_null:
+            raise com.UnsupportedOperationError(
+                "Combining `include_null=True` and `where` is not supported "
+                "by pyspark"
```
cpcloud, line 247: Lotta boilerplate with the error handling, but we can clean that up later.

jcrist: Agreed, but I'd rather copy-paste stuff around than develop something that increases coupling. It's only a couple methods where this arg even makes sense.

cpcloud (approved):
:shipit:

ibis/backends/sql/compilers/pyspark.py
cpcloud: Yep, agreed.

PR #9779, opened 2024-08-06
feat(decompile): make the decompiler run on TPCH query 1

gforsyth:
Context: the sql -> ibis expression parser/creator runs off of a `sqlglot.Plan` object

The `sqlglot` plan optimizer replaces a few items with aliases.  In other places, it doesn't use
fully-qualified column names.  Both of these prevent us from reliably converting into an Ibis expression.

### Dereferencing aggregation operands

The sqlglot planner will pull out computed operands into a separate
section and alias them e.g.
```
Context:
    Aggregations:
    - SUM("_a_0") AS "sum_disc_price"
    Operands:
    - "lineitem"."l_extendedprice" * (1 - "lineitem"."l_discount") AS _a_0
```

For the purposes of decompiling, we want these to be inline, so here we
replace those new aliases with the parsed sqlglot expression

### Dereferencing projections

The sqlglot planner will (sometimes) alias projections to the aggregate
that precedes it.

```
- Sort: lineitem (132849388268768)
  Context:
    Key:
      - "l_returnflag"
      - "l_linestatus"
  Projections:
    - lineitem._g0 AS "l_returnflag"
    - lineitem._g1 AS "l_linestatus"
    <snip>
  Dependencies:
  - Aggregate: lineitem (132849388268864)
    Context:
      Aggregations:
        <snip>
      Group:
        - "lineitem"."l_returnflag"  <-- this is _g0
        - "lineitem"."l_linestatus"  <-- this is _g1
        <snip>
```

These aliases are stored in a dictionary in the aggregate `groups`, so if
those are pulled out beforehand then we can use them to replace the
aliases in the projections.

### Dereferencing sort keys

The sqlglot planner doesn't fully qualify sort keys

```
- Sort: lineitem (132849388268768)
  Context:
    Key:
      - "l_returnflag"
      - "l_linestatus"
```

~For now we do a naive thing here and prepend the name of the sort~
~operation itself, which (maybe?) is the name of the parent table.~
~*This is definitely the most brittle part of the existing decompiler*~
 
I've mucked around with this a little bit, and while it's a _touch_ hacky, using the deferred operator with the not-fully-referenced sort-keys works well.


ibis/expr/tests/test_sql.py
```diff
@@ -149,3 +149,107 @@ def test_parse_sql_in_clause(snapshot):
...
+    l_returnflag,
+    l_linestatus
+ORDER BY
+    l_returnflag,
+    l_linestatus;
+    """
+
+    expr = ibis.parse_sql(sql, tpch_catalog)
+    code = ibis.decompile(expr, format=True)
+    snapshot.assert_match(code, "ibis_tpch1.py.txt")
```
cpcloud, line 224: We can probably call these `out.py.txt` to at least avoid having to adjust the path for every test. The name of the test is already a directory, so there's not much to be gained by duplicating the test name in the final output file.

ibis/expr/tests/test_sql.py
```diff
@@ -149,3 +149,107 @@ def test_parse_sql_in_clause(snapshot):
...
+            o_orderdate,
+            o_shippriority
+        ORDER BY
+            revenue DESC,
+            o_orderdate
+        LIMIT 10"""
+
+    expr = ibis.parse_sql(sql, tpch_catalog)
+    code = ibis.decompile(expr, format=True)
+    snapshot.assert_match(code, "ibis_tpch3.py.txt")
```
cpcloud, line 255: If we call these `out.tpch.py` then we can tell ruff to globally ignore them in `pyproject.toml` with a glob **and** have automatic syntax highlighting.

ibis/expr/sql.py
```diff
@@ -276,6 +374,24 @@ def convert_in(in_, catalog):
...
+
+
+@convert.register(sge.DataType)
+def datatype(datatype, catalog):
+    return SqlglotType().to_ibis(datatype)
+
+
+@convert.register(sge.Count)
+def count(count, catalog):
+    return ibis._.count()
```
cpcloud, line 392: Since `Count` can operate on columns, and a count on a column has a different meaning than on a table (the column version ignores nulls like other aggregates), this can't just be underscore-dot-count.

This is what is causing some of the test failures in CI.

cpcloud (changes requested):

gforsyth:
Ok, this is ready for a look.  I think we can improve the testing beyond just DuckDB, but for now this is a definite improvement over the current decompiler and I think there is more we can do to improve it.

If we're good with this, I can look into adding more of the TPCH tests

ibis/backends/duckdb/tests/test_decompile_tpch.py
```diff
@@ -0,0 +1,101 @@
...
+    tpch_query_file = SQL_QUERY_PATH / f"{tpch_query:02d}.sql"
+    with open(tpch_query_file) as f:
+        sql = f.read()
+
+    expr = ibis.parse_sql(sql, tpch_catalog)
+    code = ibis.decompile(expr, format=True)
+    snapshot.assert_match(code, "out_tpch.py")
+
+    # Import just-created snapshot
+    SNAPSHOT_MODULE = f"ibis.backends.duckdb.tests.snapshots.test_decompile_tpch.test_parse_sql_tpch.tpch{tpch_query:02d}.out_tpch"
```
cpcloud, line 86: yikes 😬

ibis/backends/tests/sql/snapshots/test_select_sql/test_bool_bool/decompiled.py
```diff
@@ -5,4 +5,4 @@
     name="airlines", schema={"dest": "string", "origin": "string", "arrdelay": "int32"}
 )
 
-result = airlines.filter((airlines.dest.cast("int64") == 0) == True)
+result = airlines.filter((((airlines.dest.cast("int64") == 0)) == True))
```
cpcloud, line 8: We're a lisp now!

cpcloud (approved):
Do we have this function marked as `@̶̜̍ë̷͔́x̵̺́p̶̠̈́e̶̠͐r̵̹͐i̴̜̎m̶̩̎e̸̛̹ň̶̺ṯ̷̃å̴͉l̷̼͘` 😅

ibis/backends/duckdb/tests/test_decompile_tpch.py
gforsyth: I plan on making this better in a follow up, I promise.

gforsyth:
> Do we have this function marked as `@̶̜̍ë̷͔́x̵̺́p̶̠̈́e̶̠͐r̵̹͐i̴̜̎m̶̩̎e̸̛̹ň̶̺ṯ̷̃å̴͉l̷̼͘` 😅

I am happy to report that we do

PR #9780, opened 2024-08-06
docs: fix spelling in pivot_longer explanation

IndexSeek:
Fixing typo in `pivot_longer` explanation. 

I have also submitted an upstream PR to catch this in the future, although it's unlikely we'll re-encounter it. 

https://github.com/codespell-project/codespell/pull/3507


cpcloud (approved):
Thanks!

Issue #8898, opened 2024-04-05
bug(pyspark): export functions don't take param or limit


deepyaman:
### What happened?

Expect to be able to pass param or limit, as with other backends.

### What version of ibis are you using?

8

### What backend(s) are you using, if any?

PySpark

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

alperentahta:
Could you please give more information, I can take this task

deepyaman:
> Could you please give more information, I can take this task

If you look at the [`to_delta()` implementation](https://github.com/ibis-project/ibis/blob/1c7c6e3/ibis/backends/pyspark/__init__.py#L862), you'll notice that the signature doesn't take `params`, but the [base implementation](https://github.com/ibis-project/ibis/blob/1c7c6e3785900cabbf25abbea5ed8505c8391b60/ibis/backends/__init__.py#L518) does. `params` could be passed to the `compile()` call in `to_delta()`.

[This issue occurred in more places when I originally filed it](https://github.com/deepyaman/ibis/blob/feat/flink/native-parquet-and-csv-export/ibis/backends/pyspark/__init__.py#L707-L789); may just be a very minimal fix now, if anything?

PR #9781, opened 2024-08-06
feat(output-formats): add support for to_parquet_dir

ncclementi:
Initial implementation to add support to write parquet files to a directory as a single file or writing each batch to a different file. 

Partially resolves https://github.com/ibis-project/ibis/issues/8584#issue-2174862538 (it doesn't cover writing csv to dirs)

TODOs: 

- [x] Implement DuckDB case (EDIT) -> looks like duckdb only supports multiple files writing when `partitioned_by` which is already supported via `to_parquet`, unless I'm missing something. 
- [x] remove? `limit= ` from pyspark, not sure why is there. Looks like this was a request for `to_delta` (see https://github.com/ibis-project/ibis/issues/8898), and it was replicated for` to_parquet_dir`
- [ ] Understand why in `ibis/expr/types/core.py`  the method `to_parquet` in the `class Expr` doesn't pass through the `params`?

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -474,6 +474,58 @@ class _FileIOHandler:
                     writer.write_batch(batch)
 
     @util.experimental
+    def to_parquet_dir(
+        self,
+        expr: ir.Table,
+        path: str | Path,
+        filename_prefix: str = "out",
+        write_batches: bool = False,
+        *,
+        params: Mapping[ir.Scalar, Any] | None = None,
+        **kwargs: Any,
+    ) -> None:
+        """Write the results of executing the given expression to a parquet file in a directory.
+
+        This method is eager and will execute the associated expression
+        immediately.
+
+        Parameters
+        ----------
+        expr
+            The ibis expression to execute and persist to parquet.
+        path
+            The data source. A string or Path to the directory where the parquet file will be written.
+        filename_prefix
+            The prefix name of the parquet file. What is before `.parquet`.
+        write_batches
+            If True writes each batch into a different file.
+        params
+            Mapping of scalar parameter expressions to value.
+        **kwargs
+            Additional keyword arguments passed to pyarrow.parquet.ParquetWriter
+
+        https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html
+
+        """
+
+        dir_path = Path(path)
+        dir_path.mkdir(parents=True, exist_ok=True)
+
+        dir_path_file = dir_path / filename_prefix
+
+        if write_batches:
+            self._import_pyarrow()
+            import pyarrow.parquet as pq
+
+            with expr.to_pyarrow_batches(params=params) as batch_reader:
+                for i, batch in enumerate(batch_reader):
+                    file_path = f"{dir_path_file}_{i}.parquet"
+                    with pq.ParquetWriter(file_path, batch.schema, **kwargs) as writer:
+                        writer.write_batch(batch)
+        else:
+            self.to_parquet(expr, f"{dir_path_file}.parquet", params=params, **kwargs)
+
+    @util.experimental
     def to_csv(
         self,
         expr: ir.Table,
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -477,7 +477,7 @@ class _FileIOHandler:
     def to_parquet_dir(
         self,
         expr: ir.Table,
-        path: str | Path,
+        directory: str | Path,
         *,
         params: Mapping[ir.Scalar, Any] | None = None,
         **kwargs: Any,
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -478,8 +478,6 @@ class _FileIOHandler:
         self,
         expr: ir.Table,
         path: str | Path,
-        filename_prefix: str = "out",
-        write_batches: bool = False,
         *,
         params: Mapping[ir.Scalar, Any] | None = None,
         **kwargs: Any,
@@ -495,10 +493,6 @@ class _FileIOHandler:
             The ibis expression to execute and persist to parquet.
         path
             The data source. A string or Path to the directory where the parquet file will be written.
-        filename_prefix
-            The prefix name of the parquet file. What is before `.parquet`.
-        write_batches
-            If True writes each batch into a different file.
         params
             Mapping of scalar parameter expressions to value.
         **kwargs
@@ -507,23 +501,17 @@ class _FileIOHandler:
         https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html
 
         """
+        self._import_pyarrow()
+        import pyarrow.parquet as pq
 
         dir_path = Path(path)
         dir_path.mkdir(parents=True, exist_ok=True)
 
-        dir_path_file = dir_path / filename_prefix
-
-        if write_batches:
-            self._import_pyarrow()
-            import pyarrow.parquet as pq
-
-            with expr.to_pyarrow_batches(params=params) as batch_reader:
-                for i, batch in enumerate(batch_reader):
-                    file_path = f"{dir_path_file}_{i}.parquet"
-                    with pq.ParquetWriter(file_path, batch.schema, **kwargs) as writer:
-                        writer.write_batch(batch)
-        else:
-            self.to_parquet(expr, f"{dir_path_file}.parquet", params=params, **kwargs)
+        with expr.to_pyarrow_batches(params=params) as batch_reader:
+            for i, batch in enumerate(batch_reader):
+                file_path = f"{dir_path}/out_{i}.parquet"
+                with pq.ParquetWriter(file_path, batch.schema, **kwargs) as writer:
+                    writer.write_batch(batch)
 
     @util.experimental
     def to_csv(
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -504,12 +504,10 @@ class _FileIOHandler:
         self._import_pyarrow()
         import pyarrow.dataset as ds
 
-        dir_path = Path(path)
-
         # by default write_dataset creates the directory
         with expr.to_pyarrow_batches(params=params) as batch_reader:
             ds.write_dataset(
-                batch_reader, base_dir=dir_path, format="parquet", **kwargs
+                batch_reader, base_dir=directory, format="parquet", **kwargs
             )
 
     @util.experimental
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -507,11 +507,9 @@ class _FileIOHandler:
         dir_path = Path(path)
         dir_path.mkdir(parents=True, exist_ok=True)
 
+        import pyarrow.dataset as ds
         with expr.to_pyarrow_batches(params=params) as batch_reader:
-            for i, batch in enumerate(batch_reader):
-                file_path = f"{dir_path}/out_{i}.parquet"
-                with pq.ParquetWriter(file_path, batch.schema, **kwargs) as writer:
-                    writer.write_batch(batch)
+            ds.write_dataset(batch_reader, base_dir=dir_path, format="parquet", **kwargs)
 
     @util.experimental
     def to_csv(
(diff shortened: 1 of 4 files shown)


ibis/backends/pyspark/__init__.py
```diff
@@ -1283,10 +1283,12 @@ def _to_filesystem_output(
...
         df = self._session.sql(self.compile(expr, params=params, limit=limit))
         if self.mode == "batch":
-            df = df.write.format(format)
+            df = df.write.format(format).mode(
+                "overwrite"
+            )  # df = df.write.format(format)
             for k, v in (options or {}).items():
                 df = df.option(k, v)
-            df.save(path)
+            df.save(os.fspath(path))  # df.save(path)
```
ncclementi, line 1291: @chloeh13q this is the change you suggested in #9790 but with only that change I get. I added mode overwrite but I'm not sure if that's what we want, plus the test fails with an assertion error. I think I might be missing something on how to_parquet works for pyspark.  
```
E               pyspark.errors.exceptions.captured.AnalysisException: [PATH_ALREADY_EXISTS] Path file:/private/
var/folders/ky/_zw0jkyd7hl3f9hfy6c994zw0000gn/T/pytest-of-naty/pytest-27/test_table_to_parquet_dir_pysp0
already exists. Set mode as "overwrite" to overwrite the existing path.
```
cc: @cpcloud since you asked about the test

ibis/backends/tests/test_export.py
```diff
@@ -215,6 +215,20 @@ def test_table_to_parquet(tmp_path, backend, awards_players):
     )
 
 
+def test_table_to_parquet_dir(tmp_path, backend, awards_players):
+    outparquet_dir = tmp_path
```
cpcloud, line 219: This fails because `tmp_path` is a directory. Check out the usage of it throughout the codebase, and also the [pytest docs on `tmp_path`](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-fixture) to get an idea for how to use it correctly.

ibis/backends/__init__.py
```diff
@@ -473,6 +473,46 @@ def to_parquet(
...
+        """
+        self._import_pyarrow()
+        import pyarrow.parquet as pq
+
+        dir_path = Path(path)
+        dir_path.mkdir(parents=True, exist_ok=True)
+
+        with expr.to_pyarrow_batches(params=params) as batch_reader:
+            for i, batch in enumerate(batch_reader):
+                file_path = f"{dir_path}/out_{i}.parquet"
```
cpcloud, line 512: Naming files this way will lead to annoying consequences like sorting file names in lexicographic order instead of the sequential order that they are written.

Since we don't know how many batches there are, you'll have to write them out and then rename them using a format string padded with zeroes that reflect the order of magnitude of the number of files written.

ibis/backends/__init__.py
```diff
@@ -473,6 +473,46 @@ def to_parquet(
...
+                with pq.ParquetWriter(file_path, batch.schema, **kwargs) as writer:
+                    writer.write_batch(batch)
```
cpcloud, line 514: This whole thing can be reduced quite a bit by using pyarrow datasets.

You should be able to also remove to `mkdir` call above, since `write_dataset` appears to already do that.

```suggestion
        import pyarrow.dataset as ds
        with expr.to_pyarrow_batches(params=params) as batch_reader:
            ds.write_dataset(batch_reader, base_dir=dir_path, format="parquet", **kwargs)
```

cpcloud (changes requested):

ibis/backends/__init__.py
cpcloud: If we use pyarrow then users can deal with this annoyance themselves using the `basename_template` parameter.

ibis/backends/pyspark/__init__.py
cpcloud: See my comment below, this is a bug in your test code, and we shouldn't do anything like overwrite by default.

ncclementi:
So I'm forcing the test to write more than one file, and I have assertion errors because of sorting problems. But it looks like I can't provide `force_sort` directly to assert_frames and for the base backend level this is set to False. 

I can do the sort by columns  in the test directly, unless there is a better way to approach this.

ibis/backends/tests/test_export.py
```diff
@@ -215,6 +215,22 @@ def test_table_to_parquet(tmp_path, backend, awards_players):
...
 
 
+def test_table_to_parquet_dir(tmp_path, backend, awards_players):
+    outparquet_dir = tmp_path / "out"
+    # max_ force to write more than one parquet file
+    awards_players.to_parquet_dir(
+        outparquet_dir, max_rows_per_file=3000, max_rows_per_group=3000
+    )
+
+    parquet_files = list(outparquet_dir.glob("*.parquet"))
```
cpcloud, line 230: Globs are not guaranteed to return in any particular order, so you need to sort them.

```suggestion
    parquet_files = sorted(outparquet_dir.glob("*.parquet"), key=lambda path: int(path.with_suffix("").name.split("-")[1])
```

ibis/expr/types/core.py
```diff
@@ -604,6 +604,37 @@ def to_parquet(
...
+        immediately.
+
+        Parameters
+        ----------
+        path
+            The data source. A string or Path to the directory where the parquet file will be written.
+        filename_prefix
+            The prefix name of the parquet file. What is before `.parquet`.
+        write_batches
+            If True writes each batch into a different file.
```
cpcloud, line 627:
```suggestion
```

These aren't arguments in the signature, so they shouldn't be documented, or if they are supported all all backends, then we should add them to the signature with the appropriate defaults and document them.

ibis/backends/pyspark/__init__.py
```diff
@@ -1286,7 +1286,7 @@ def _to_filesystem_output(
             df = df.write.format(format)
             for k, v in (options or {}).items():
                 df = df.option(k, v)
-            df.save(path)
+            df.save(os.fspath(path))
```
cpcloud, line 1289: Is this code actually being tested anywhere specific, or does it get hit as part of another code path?

ibis/backends/__init__.py
```diff
@@ -473,6 +473,45 @@ def to_parquet(
                 for batch in batch_reader:
                     writer.write_batch(batch)
 
+    @util.experimental
+    def to_parquet_dir(
+        self,
+        expr: ir.Table,
+        path: str | Path,
```
cpcloud, line 480:
```suggestion
        directory: str | Path,
```

Seems less ambiguous (even though the method name ends with `_dir`).

ibis/backends/__init__.py
```diff
@@ -473,6 +473,45 @@ def to_parquet(
...
+        """Write the results of executing the given expression to a parquet file in a directory.
+
+        This method is eager and will execute the associated expression
+        immediately.
+
+        Parameters
+        ----------
+        expr
+            The ibis expression to execute and persist to parquet.
+        path
```
cpcloud, line 494:
```suggestion
        directory
```

cpcloud (changes requested):

ibis/expr/types/core.py
ncclementi: whoops good catch, I'll clean it up right away,

ibis/backends/pyspark/__init__.py
ncclementi: This wasn't tested before, but I believe it gets hit in the test I'm running, because before this the test failed in the read parquet part. Maybe @chloeh13q has more context

ibis/backends/__init__.py
ncclementi: @cpcloud if we do this change which I think it makes sense, do we want to be consistent with the pyspark implementation and changed the name of it in there too?

cpcloud: Yep, but both of those are fine to do in a follow-up.

ibis/backends/pyspark/__init__.py
cpcloud: Ah, it's that pyspark doesn't support `Path` objects 🙄

cpcloud (approved):
LGTM, thanks for adding this!

cpcloud:
Let's run the clouds to make sure this works there too!

ncclementi:
It looks like the cloud ones were skipped, not sure why.

cpcloud:
Oh, probably because you can't have two PRs running them simultaneously, it's annoying.

cpcloud:
I will run them locally.

ibis/expr/types/core.py
```diff
@@ -604,6 +604,33 @@ def to_parquet(
...
+        Parameters
+        ----------
+        directory
+            The data source. A string or Path to the directory where the parquet file will be written.
+        params
+            Mapping of scalar parameter expressions to value.
+        **kwargs
+            Additional keyword arguments passed to pyarrow.parquet.ParquetWriter
+
+        https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html
```
gforsyth, line 629:
```suggestion
            Additional keyword arguments passed to pyarrow.dataset.write_dataset
            
        https://arrow.apache.org/docs/python/generated/pyarrow.dataset.write_dataset.html
```

cpcloud:
Clouds look good

```
cloud in 🌐 falcon in …/ibis on  to_pq_dir is 📦 v9.2.0 via 🐍 v3.10.14 via ❄️   impure (ibis-3.10.14-env)
bringing up nodes...
......x.....x.......................x......x.................xxx..xxxx.x.x.xxxx.........x...........x......x..x..................x..............xx.x.x..x....xx...x.x......x..x....x.xx.x.x.x.........x.x..x.x....x..x.xxx.xx.x.xx...x.xxx.x.x..xx.x...xx...xxxx.x..x...x.....x.x.x.....xx.xx.xxxxx.x.xxxxxxxxx...x..x...x.xx....x.x...x...x.....sx........................xx...x.... [  9%]
....x........x......x........................................x.........x.......x.x........x..........x..xx..x....................x..............x.................................x...................................x..x.....x.........x..............x.....x..x........x...x.......x.............x....x.x.....................................s..........x........................ [ 19%]
.....s....s................................x...............sx....x................x.......................x..............................x..x.............................x...............s.ssssssssssssssssssss.........................................................x....x...xx..........x...xx............x...xx.........x.......x...x....sssssssssssssssssssssssssssssssssssss [ 28%]
ssssssssssssssssssssxsssssxsx...x.......x...x.......x..x.x.x...x......x....x......x...........x........x...........x..x.........x..x...xxxxxxxxxx.x...........xx.............x............x.......x....xx...x..x.....x....................x...x..x........................................x......x............xx...x....x...........xx......x...............x........................ [ 38%]
...x.x..............................................................................................x...........................x.....x...x...................x......................x............x........xx.......x.....s........x............xx.....xx.x..xxxxxx.x.x...x......xxxxx.x...xxxxxx.xxxxxxxxx.xxx..........x....x.x...xxx...............x.......x................x..... [ 47%]
....x....xxx..xxxxxxxxxx.xxxxx.xx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.xxxxxxxxxxxxxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.x...............x..........................................x.........................................................................................x..................x.......................x..........x..x....x.x..x.........x.....x..x [ 57%]
............x...x...x.....x....x..............x..........x...x.............xx.................xx.x..x.x......x....s........................s............x................s.........................x...................x..................................x..x.............s....................x..x...x.....................x........x.....x.......................................x [ 66%]
...................................x.x..x.......x......................x..............x......xx......x......x.....x....x......x....x.....x.............x....x...x....x........x....x..x.xx........xx.x.....xxxxxxxxxxxx..................xx.x.x.xx.....x...x.........................x........x...x....xx......x................x......xx....x........x.....x..x.x......x............ [ 76%]
..x....xx........x.x...............x........x.x.....x....x.....x..x..x........x.....x..xx...x..x.x..x..............x.......x.x..........xx............x...x.xx..x...xx............x..x..............xx.xx...x......x.x.xxx..xx...x...sxx..x.x....x..x..x....x.x..............sx..s..x....x....s.......x...x....s.........s................................x....................x..x.. [ 86%]
.........................................................................................................................................................................................x...........................................................s..............................................s................................................................................ [ 95%]
...........................................................................................................................................................................                                                                                                                                                                                                           [100%]
3238 passed, 102 skipped, 561 xfailed in 703.77s (0:11:43)
```

PR #9787, opened 2024-08-07
feat(ux): allow window functions in predicates and compile to `QUALIFY` where possible

### LINKED ISSUE #9775: feat: Support `QUALIFY` syntax to filter on analytic expression

_No response_

I'm a developer of [BigQuery Dataframes ](https://github.com/googleapis/python-bigquery-dataframes). I would like to programatically build SQL queries using ibis. We seek to build concise SQL to preserve readability and stay under 1mb text size limit. One way to achieve this is to allow directly filtering on analytic expressions using the QUALIFY syntax

I'd like to express a qualify expression as preferably `table.filter(t["col"].mean().over(window) > 0.4)`. Right now this produces something like `
IntegrityError: Cannot add <ibis.expr.operations.logical.Greater object at )=0x************> to filter, it is a reduction which must be converted to a scalar subquery first
`. In bakends supporting qualify, this would produce `QUALIFY` sql syntax, while other backends may convert to subquery.

8.0.0

BigQuery

### Code of Conduct

cpcloud:
Closes #9775.


cpcloud:
Ideally we can get this in for 9.3, but if it turns out to be a huge pain then we can of course wait.

jcrist (approved):
Nice

ibis/backends/tests/test_generic.py
```diff
@@ -344,28 +344,16 @@ def test_filter(backend, alltypes, sorted_df, predicate_fn, expected_fn):
...
+@pytest.mark.notimpl(["polars", "druid", "exasol", "pandas", "dask"])
+@pytest.mark.notyet(
+    ["oracle"],
+    raises=OracleDatabaseError,
+    reason="sqlglot `eliminate_qualify` transform produces underscores in aliases, which is not allowed by oracle",
+)
+@pytest.mark.never(
+    ["mssql"],
+    raises=PyODBCProgrammingError,
+    reason="sqlglot transform produces an order by in a subquery, which is not allowed by mssql",
```
cpcloud, line 356: I think this is actually an issue with the test, not mssql or sqlglot. Looking into it.

cpcloud: Yep, this is now passing for MS SQL, and I cleaned up the markers to better showcase what's not passing and why.

PR #9790, opened 2024-08-07
fix(pyspark): fix a bug in spark's to_parquet_dir method

chloeh13q:
## Description of changes

Fix a bug in spark's to_parquet_dir method. Spark's write method only accepts strings as paths, and the current implementation will throw something like
```python
AttributeError: 'PosixPath' object has no attribute '_get_object_id'
```

I was already doing this in streaming mode, just forgot to do this in batch mode.

xref: #9781


cpcloud:
Do we have a test for this?

chloeh13q:
@cpcloud @ncclementi is adding one in #9781, so I didn't add one.

cpcloud:
I think this was fixed in @ncclementi's PR, closing!

PR #9792, opened 2024-08-08
chore: consolidate error handling in __rich_console__

NickCrews:
This is the first step of some larger refactoring of the repr'ing code that I want to do, to hide a bunch more implementation details into pretty.py so that there is a better defined, smaller API.
I want to clean some things up first though.

We can pin down more precisely where the TranslationError happens: when we call
` to_rich(self, console_width=console_width)` and actually
try to execute the expression.
So let's consolidate the two try-catches down to just one, now there is less to think about. Now, `self._interactive_repr()` is really just `_capture_rich_renderable()`,
and so move this out into a pure function
where it is more obvious what it's responsibility is.

### DIFF
--- a/ibis/expr/types/core.py
+++ b/ibis/expr/types/core.py
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
     import polars as pl
     import pyarrow as pa
     import torch
-    from rich.console import Console
+    from rich.console import Console, RenderableType
 
     import ibis.expr.types as ir
     from ibis.backends import BaseBackend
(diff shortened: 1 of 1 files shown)


cpcloud (approved):
Seems like a reasonable change!

I know @jcrist has thoughts on the current rich table rendering.

@NickCrews Are you planning on changing the repr in any significant way?

We're more likely to build our our table repr at this point given the difficulty we've had with rich, than to invest a bunch of time in improvements.

NickCrews:
Thanks for the quick response!

I want to
- make .preview() be the correct (eg wide and scrollable) width in jupyter environments. Currently it gives back a rich.Table, with all of the elements rendered as wide as they need. But then this rich.Table is rendered with the default rich Console, not with our custom implementation that modifies ConsoleOptions. The result is that even though all the elements are wide, the entire table is squished inside rich's default [`JUPYTER_DEFAULT_COLUMNS=115`](https://github.com/Textualize/rich/blob/e9f75c9912ed25b9777bc0257853370951220b17/rich/console.py#L77), which means that a lot of the columns are actually only a few columns wide and totally ileglible.
- move all the repr stuff into pretty.py, getting it out of core.py. Just makes things easier to follow

Other changes, like supporting a more HTMLy output format for jupyter environments, eg with pagination, I am very down for, but I wasn't thinking of doing this.

PR #9794, opened 2024-08-08
feat(snowflake): implement interval arithmetic

cpcloud:
Closes #9783.

PR #9799, opened 2024-08-08
feat(timestamps): add support for timestamp/date +/- intervals for additional backends

cpcloud:
Based on #9794, I decided to investigate whether we can do a bit better on date/timestamp +/- interval-style operations.

Many backends support adding and subtracting intervals from dates and timestamps, but do **not** support intervals as a first-class type, and we can allow that functionality to work.

This PR adds support for those operations to mssql and oracle and consolidates some existing implementation details around the `IntervalFromInteger` operation.

### DIFF
--- a/ibis/backends/sql/compilers/mssql.py
+++ b/ibis/backends/sql/compilers/mssql.py
@@ -86,14 +86,12 @@ class MSSQLCompiler(SQLGlotCompiler):
         ops.BitXor,
         ops.Covariance,
         ops.CountDistinctStar,
-        ops.DateAdd,
         ops.DateDiff,
-        ops.DateSub,
         ops.EndsWith,
         ops.IntervalAdd,
-        ops.IntervalFromInteger,
-        ops.IntervalMultiply,
         ops.IntervalSubtract,
+        ops.IntervalMultiply,
+        ops.IntervalFloorDivide,
         ops.IsInf,
         ops.IsNan,
         ops.LPad,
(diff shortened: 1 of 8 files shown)


ibis/backends/tests/test_temporal.py
```diff
@@ -631,32 +641,21 @@ def convert_to_offset(offset, displacement_type=displacement_type):
...
 @pytest.mark.notimpl(["exasol"], raises=com.OperationNotDefinedError)
 def test_integer_to_interval_date(backend, con, alltypes, df, unit):
     interval = alltypes.int_col.to_interval(unit=unit)
-    array = alltypes.date_string_col.split("/")
-    month, day, year = array[0], array[1], array[2]
-    date_col = ibis.literal("-").join(["20" + year, month, day]).cast("date")
+    month = alltypes.date_string_col[:2]
+    day = alltypes.date_string_col[3:5]
+    year = alltypes.date_string_col[6:8]
+    date_col = ("20" + year + "-" + month + "-" + day).cast("date")
```
cpcloud, line 658: More backends implement the primitives here than the previous code, which used arrays.

ibis/backends/sql/compilers/mssql.py
```diff
@@ -404,6 +400,8 @@ def visit_Cast(self, op, *, arg, to):
             return arg
         elif from_.is_integer() and to.is_timestamp():
             return self.f.dateadd(self.v.s, arg, "1970-01-01 00:00:00")
+        elif from_.is_integer() and to.is_interval():
+            return sge.Interval(this=arg, unit=self.v[to.unit.singular])
```
cpcloud, line 404: Since this is such a common implementation I'm going to try and encode the logic in the base implementation of `visit_Cast`

gforsyth (approved):
Looks good -- not sure if you wanted to handle the casting work in this PR or separately

cpcloud:
I'll do it in a follow-up now that you mention it.

cpcloud:
Lemme run the clouds

PR #9802, opened 2024-08-08
ci(snowflake): acquiesce to the dependency gods and unpin cloudpickle

cpcloud:
Unpinning cloudpickle because the specific version does not matter for the test case we are using it in.


gforsyth (approved):
:scream:

PR #9808, opened 2024-08-09
feat(flink): support `ArrayValue.collect()`

### LINKED ISSUE #9801: feat: ArrayCollect aggregation for Flink backend
### What happened?

Attempting to aggregate rows to an array using collect() is resulting in an operation not supported error for ArrayCollect.

Flink supports this functionality through the [ARRAY_AGG function](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/functions/systemfunctions/) while the Flink dialect inherits from Hive which uses COLLECT_LIST.

I was attempting to implement this myself until the recent changes supporting order_by in some aggregates dropped and it derailed my changes. Would appreciate some help here as it seems like some additional plumbing is needed for this to work in flink.

9.3.0

flink, duckdb

### Relevant log output

_No response_

### Code of Conduct

ArtnerC:
## Description of changes

Flink inherits from the Hive backend which uses COLLECT_LIST where flink uses ARRAY_AGG. Rename to use ARRAY_AGG and implement Flink specific aggregation options.

## Issues closed

* Resolves #9801


cpcloud:
Looks like this is only available in Flink 1.20, per [this JIRA ticket](https://issues.apache.org/jira/browse/FLINK-21949).

cpcloud:
@chloeh13q @deepyaman Can y'all give this a review?

jcrist (approved):
LGTM. Curious: why did you add the array sort changes in this PR?

ibis/backends/sql/compilers/flink.py
```diff
@@ -575,5 +575,21 @@ def visit_MapMerge(self, op: ops.MapMerge, *, left, right):
...
+    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null):
+        if order_by:
+            raise com.UnsupportedOperationError(
+                "ordering of order-sensitive aggregations via `order_by` is "
+                "not supported for this backend"
+            )
+        # the only way to get filtering *and* respecting nulls is to use
+        # `FILTER` syntax, but it's broken in various ways for other aggregates
+        out = self.f.array_agg(arg)
+        if not include_null:
```
jcrist, line 587: Note that Flink does support `IGNORE NULLS` for this operation, but it's broken if the argument is anything more complicated than a direct column reference. Using manual filtering like you're doing here is the best we can do for now.

cpcloud:
> LGTM. Curious: why did you add the array sort changes in this PR?

So I didn't have to muck around with converting everything to frozensets and comparing. Flink doesn't return arrays in the order they were unnested in when you collect them.

PR #9824, opened 2024-08-13
depr(api): deprecate `StructValue.destructure` API

deepyaman:
## Description of changes

Deprecate `StructValue.destructure` in favor of `StructValue.lift` or `Table.unpack`.


ibis/backends/tests/test_vectorized_udf.py
deepyaman, line 1: Note that this file seems to only be running on pandas, Dask, and PySpark backends for some reason...

cpcloud: Because those are historically the only backends that supported the vectorized UDF API.

cpcloud (approved):

PR #9827, opened 2024-08-13
fix(repr): format scalar values with same logic as columnar values

### LINKED ISSUE #9820: bug: repr for count now includes np.int64 or whatever type
### What happened?

on a fresh install:

`pip install 'ibis-framework[duckdb,examples]'`

repr now seems to include numpy types in the output. for instance:

```python
[ins] In [1]: import ibis

[ins] In [2]: ibis.options.interactive = True

[nav] In [3]: t = ibis.examples.penguins.fetch()

[ins] In [4]: t.count()
Out[4]:
┌───────────────┐
│ np.int64(344) │
└───────────────┘
```

main

duckdb

### Relevant log output

_No response_

### Code of Conduct

jcrist:
Previously our interactive repr just called `str(value)` on scalars before passing off to `rich`. We now apply the same logic we do for columnar values to scalars as well. This provides a more uniform interactive repr experience, where scalars and columns have identical formatting rules.

Fixes #9820

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -1123,9 +1123,9 @@ class Value(Expr):
         >>> one = ibis.literal(1)
         >>> two = ibis.literal(2)
         >>> two.identical_to(one + one)
-        ┌──────────┐
-        │ np.True_ │
-        └──────────┘
+        ┌──────┐
+        │ True │
+        └──────┘
         """
         try:
             return ops.IdenticalTo(self, other).to_expr()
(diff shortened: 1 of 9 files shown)


ibis/expr/types/generic.py
```diff
@@ -1175,19 +1175,19 @@ def group_concat(
...
+        ┌────────────────────────┐
+        │ 39.1: 39.5: 40.3: 36.7 │
+        └────────────────────────┘
 
         >>> t.bill_length_mm.group_concat(sep=": ", where=t.bill_depth_mm > 18)
-        ┌──────────────┐
-        │ '39.1: 36.7' │
-        └──────────────┘
+        ┌────────────┐
+        │ 39.1: 36.7 │
```
cpcloud, line 1189: Do the string `"90"` and the number `90` repr the same now?

jcrist: Yes, but that's the same as they do for columns as well.

jcrist: I suppose we could quote strings for scalars only?

cpcloud: I guess the main difference is that with the column you get the type information.

-0 on losing the quotes, but not a blocker.

cpcloud (approved):

ibis/expr/types/generic.py
cpcloud: Eh, let's just leave it and see how it feels to use it.

PR #9830, opened 2024-08-13
fix(pyspark): suppress errors for unsupported Databricks serverless compute properties

### LINKED ISSUE #9275: Config Value spark.sql.mapKeyDedupPolicy not Supported by Databricks SQL Warehouse
Getting the error `spark.sql.mapKeyDedupPolicy` is not supported by Databricks SQL Warehouses when using ibis pyspark with a Databricks SQL Warehouse Cluster.

See: https://community.databricks.com/t5/data-engineering/spark-settings-in-sql-warehouse/td-p/7959

Set in do_connnect:
https://github.com/ibis-project/ibis/blame/e425ad57899f8ebbea29b57bb53cedb40ebd7193/ibis/backends/pyspark/__init__.py#L180
```python
self._session.conf.set("spark.sql.mapKeyDedupPolicy", "LAST_WIN")
```

Workaround could be as simple as:
```python
try:
    spark.conf.set("spark.sql.mapKeyDedupPolicy", "LAST_WIN")
except Exception as e:
    if "not available" in str(e):
        print("Likely running in a SQL Warehouse")
    else:
        raise e  # Re-raise other exceptions

```

but I'm not sure what other approaches there might be.

ArtnerC:
## Description of changes

Suppress errors when setting spark properties on Databricks serverless compute that are not supported.

## Issues closed

* Resolves #9275


gforsyth:
Hey @ArtnerC, thanks for putting this in!

The Risingwave failure is spurious, github actions just had a bad morning.  But I think the `ModuleNotFound` error in the `pyspark` job is legit -- and you can run `pre-commit run --all-files` to handle the lint error.

cpcloud:
Will clean up the imports here and merge!

cpcloud (approved):
Thanks for powering through!

ArtnerC:
I think this one is going to do it. Sorry for the spam!

cpcloud:
Don't worry about the patch failure, the pyspark coverage hasn't been uploaded yet.

PR #9831, opened 2024-08-13
docs(snowflake): fix the syntax for passing schema

deepyaman:


docs/backends/snowflake.qmd
```diff
@@ -119,7 +120,8 @@ APIs. All values of `authenticator` are supported.
 con = ibis.snowflake.connect(
     user="user",
     account="safpqpq-sq55555",
-    database="my_database/my_schema",
+    database="my_database",
```
deepyaman, line 123:
```suggestion
    database="my_catalog/my_database",
```
Does it make sense to have something like this?

docs/backends/snowflake.qmd
```diff
@@ -79,7 +79,8 @@ con = ibis.snowflake.connect(
     user="user",
     password="password",
     account="safpqpq-sq55555",
-    database="IBIS_TESTING/IBIS_TESTING",
+    database="IBIS_TESTING",
+    schema="IBIS_TESTING",
```
deepyaman, line 83:
```suggestion
    database="my_database",
    schema="my_schema",
```

or 

```suggestion
    database="my_catalog/my_database",
    schema="my_schema",
```

may be more helpful?

docs/backends/snowflake.qmd
cpcloud: Hm, probably not. The idea is to do as little customization here so we can just dump kwargs to the underlying connector.

deepyaman: Sorry, what I meant is, based on https://github.com/ibis-project/ibis/blob/9.3.0/ibis/backends/snowflake/tests/test_client.py#L411-L412, it seems like you can pass in the form `database="catalog/database"`. However, this isn't as clear from the examples, so I was wondering if it makes sense to document it as such.

cpcloud: Sorry, I should clarify that I believe we removed that parsing functionality in 9.0. I don't think the comment is relevant anymore.

cpcloud (approved):

PR #9832, opened 2024-08-13
refactor: DDL accessor [WIP] 

ncclementi:
Very rough start, so far I created the DDL accessor and I'm trying to get to run the some methods as they are (no changes yet to functionality) from the accessor like `con.ddl.list_`, this  works. But I'm having problems triggering the deprecation when doing `con.list_`. 

I thought if I deprecate the abstract methods that would do it, but it doesn't because it looks immediately for the backend specific implementation. Not sure what's the best solution here design wise. 

** How do we deprecate in this case? **
If I move the deprecation to the Backend implementation of let's say `list_tables` as an example, it will trigger both when doing `con.list_tables()` and `con.ddl.list_tables()` 

cc: @gforsyth for when you are back


ncclementi:
@gforsyth I did a first path at

- Implementing the `list_*` for duckdb
- Safeguarding the `DDLAccessor` with a higher level of `NotImplementedError`.
- Modifying `con.tables` to use `list_*` making sure that it uses the one that are implemented for the specific backend. 
- Added api tests for all the `list_*` and one for modified the previous `con.list_tables` test (now named `test_list_all_tables_and_views`)

Right now we have a lot of tests that rely on asserts that use `con.list_tables` I haven't modified any of them yet since in some cases it should be `con.tables` and others might be `con.ddl.list_tables`. If we are good with this idea/design, I'll move forward with cleaning up tests and start working on the other backends.

ibis/backends/__init__.py
```diff
@@ -47,6 +47,31 @@ class TablesAccessor(collections.abc.Mapping):
...
+            "list_views",
+            "list_temp_views",
+        ]
+
+        self._tables = []
+        for method_name in list_methods:
+            self._tables.extend(self._execute_if_exists(method_name))
+
+        # TODO: maybe there is a better way of removing duplicates, for now
+        self._tables = list(set(self._tables))
```
gforsyth, line 63: I think there are two things to consider here, one that is actionable now, one that is a TODO for later on:

1. (now) If `_tables` is only populated at backend initialization, then `con.tables` won't show any tables/views that a user creates using, e.g. `read_parquet` -- I think this needs to be computed on demand.
2. (later) If we show the _type_ of a table object in the output of `con.tables`, we might not want to deduplicate the results, but rather show that there is say a table called `foo` AND a view called `foo`

ibis/backends/duckdb/__init__.py
```diff
@@ -973,86 +973,133 @@ def read_delta(
...
+        col = "table_name"
+        sql = (
+            sg.select(col)
+            .from_(sg.table("tables", db="information_schema"))
+            .where(
+                C.table_catalog.eq(sge.convert(catalog)),
+                C.table_schema.eq(sge.convert(database)),
+                C.table_type.eq("BASE TABLE"),
+            )
+            .sql(self.name, pretty=True)
```
gforsyth, line 997: I think you can drop the `pretty` bit, it adds some small amount of overhead and it isn't necessary

gforsyth, line 997: Now that you have this pattern established, I think it might make sense to pull this sqlglot query into a helper function -- it's the same query for each of the `list_` methods you are adding except the predicates to the `where` call, so maybe each `list_` function just constructs a list of those predicates and everything else is a shared implementation.

ibis/backends/__init__.py
ncclementi: >  1. (now) If _tables is only populated at backend initialization, then con.tables won't show any tables/views that a user creates using, e.g. read_parquet -- I think this needs to be computed on demand.

But isn't the computation being taken care by the fact that we have this method in the Backend class? 

```python
    @functools.cached_property
    def tables(self):
    ...
```

As a matter of fact I just tried this with a file I have locally, and it shows up: 

```python
In [1]: import ibis

In [2]: con = ibis.get_backend()

In [3]: t = con.read_parquet("docs/posts/ibis-lonboard-geo/overture_maps_data/download_dc/dc-infra.geoparquet", "my_geo_table")

In [4]: con.tables
Out[4]: 
Tables
------
- my_geo_table
```

For 2) I wonder if we want to do this now, or rethink the repr later on a followup

ibis/backends/__init__.py
```diff
@@ -62,29 +87,70 @@ def __getattr__(self, name) -> ir.Table:
...
     def _ipython_key_completions_(self) -> list[str]:
-        return self._backend.list_tables()
+        return self._tables
+
+
+class DDLAccessor:
+    """ddl accessor list views."""
+
+    def __init__(self, backend: BaseBackend):
+        self._backend = backend
```
cpcloud, line 118: I know we have this elsewhere, but I think the circular reference here adds a lot of complexity to the lifetime semantics of a backend instance, especially due to the strong reference held by the `tables` `cached_property`.

Can we make this a weak reference? This ensures that a backend instance cannot stick around solely because there's a DDL accessor hanging around retaining a reference to the backend.

ncclementi: I've only seen examples of us using weak references in some of our test. just to make sure, are you suggesting doing this? 

```python
    def __init__(self, backend: BaseBackend):
        self._backend = weakref.ref(backend)
```

cpcloud: Thinking about this some more, I think we probably need to keep the reference cycle.

For example,

the following code should work:

```python
con = ibis.backend.connect(...)
tables = con.tables
del con
list(tables)
```

I think this would fail if `con` is a weakref.

Anyway, let's keep it as is until I can understand whether this would fail and why.

ncclementi: Actually, having a weak referent it doesn't even let access the tables. It fails when I'm trying to do access teh methods to list things



```python
In [1]: import ibis

In [2]: con = ibis.get_backend()

In [3]: penguins = ibis.examples.penguins.fetch()

In [4]: con.tables
```

```python-traceback
File ~/Documents/git/my_forks/ibis/ibis/backends/__init__.py:122, in DDLAccessor._raise_if_not_implemented(self, method_name)
    121 def _raise_if_not_implemented(self, method_name: str):
--> 122     method = getattr(self._backend, method_name)
    123     if not callable(method):
    124         raise NotImplementedError(
    125             f"The method {method_name} is not implemented for the {self._backend.name} backend"
    126         )

AttributeError: 'weakref.ReferenceType' object has no attribute '_list_tables'
```

cpcloud: That's because you're using `weakref.ref` incorrectly, deferencing requires invoking the object like a function:

```python
obj = my_object()
obj_ref = weakref.ref(obj)
underlying = obj_ref()
# `underlying` is either `None` or the referrent. 
```

What you want is `weakref.proxy(self)` which behaves like the object it holds.

I opened a PR to clean up existing circular references in https://github.com/ibis-project/ibis/pull/9895, you can see the

ibis/backends/__init__.py
gforsyth: The 2nd point can definitely happen in a follow-up, I just think we should leave a `TODO` there.

Re: the first point, no, I think this is not correct, because tables that are created _after_ the initial instantiation (triggered by asking for `con.tables`) won't be picked up, because the `self._tables` attribute in the `Accessor` won't get re-initialized.

e.g.:

```python
[ins] In [1]: import ibis

[ins] In [2]: con = ibis.duckdb.connect()

[ins] In [3]: con.create_table("table1", {"col": [["abc"], ["abcd"]]})
Out[3]: 
DatabaseTable: memory.main.table1
  col array<string>

[ins] In [4]: con.tables
Out[4]: 
Tables
------
- table1

[nav] In [5]: con.create_table("table2", {"col": [["abc"], ["abcd"]]})
Out[5]: 
DatabaseTable: memory.main.table2
  col array<string>

[ins] In [6]: con.tables
Out[6]: 
Tables
------
- table1
```

ncclementi: >  The 2nd point can definitely happen in a follow-up, I just think we should leave a TODO there.

This makes sense, I'll do that. 

> Re: the first point, no, I think this is not correct, because tables that are created after the initial instantiation (triggered by asking for con.tables) won't be picked up, because the self._tables attribute in the Accessor won't get re-initialized.

Oh, I see thank for the clarification.  One solution is  make the tables method  a `@property` instead of `@functools.cached_property`. Is that what you meant by "I think this needs to be computed on demand"

gforsyth: That will suffer from the same issue.  The `self._tables` attribute is being populated in the `__init__` of the class -- once the Accessor is initialized, then `_tables` is frozen and doesn't get updated.

gforsyth: Oh, maybe I misunderstood what you were saying -- if `_tables` is a `property` in the `TablesAccessor` class then that will work

ibis/backends/__init__.py
ncclementi: I rebased (and will do if there are more changes) against your PR, to keep track of this plus my changes to be consistent.

ibis/backends/__init__.py
ncclementi: I was referring to the `tables` method, and by doing that your example now shows both tables. That being said idk if that's by happy accident. 

#this code no is adapted to reflect Phillips PR
```python
    @property #I meant this
    def tables(self):
        """An accessor for tables in the database.
        ...
        """
        return TablesAccessor(weakref.proxy(self))
```

```python
In [1]: import ibis

In [3]: con.create_table("table1", {"col": [["abc"], ["abcd"]]})
Out[3]: 
DatabaseTable: memory.main.table1
  col array<string>

In [4]: con.tables
Out[4]: 
Tables
------
- table1

In [5]: con.create_table("table2", {"col": [["abc"], ["abcd"]]})
Out[5]: 
DatabaseTable: memory.main.table2
  col array<string>

In [6]: con.tables
Out[6]: 
Tables
------
- table1
- table2
```

gforsyth: Ok, that works but it's creating a new `TablesAccessor` object each time.  I don't know if we care about that overhead, but I was thinking that the thing that should be a `property` was `_tables`, and then the above `tables` can remain a `cached_property`

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -292,15 +293,17 @@ def test_list_tables_schema_warning_refactor(con):
         "diamonds",
         "functional_alltypes",
         "win",
-    }.issubset(con.list_tables())
+    }.issubset(con.tables)
 
     icecream_table = ["ice_cream"]
 
     with pytest.warns(FutureWarning):
+        # I'M HAVING AN ISSUE HERE THAT THE TEST ITSELF RUNS BUT WHEN I RUN
```
ncclementi, line 301: I'm not sure what's happening here, I think it might be related to the caching. when I run this test isolated it passes, but when I run the whole `test_client.py` the LHS ends up with more tables that I'm not sure where they are coming from.

ncclementi:
I'm trying to get the DuckDB backend green so I can move on to the other backend's implementation but I'm having a few issues that I idk where they are coming from. I think if we can fix those we are on a safe point to move onto the other backends.

ncclementi:
Discussion for postgres: 

So, we are currently  on main doing something weird with `list_tables()` in postgres that doesn't match necessarily any behavior in psotgres sql. 

I created a table, a temp table, a view a temp view and a foreign table in psql. 
- `\dt`  only lists table and temp table
- `\dv` only lists view and temp view
- `\det` only list foreign table
- `\dvtE` list them all

on Ibis `con.list_tables()` 
- lists the view, table, and foreign table (no temp table or temp view) 

On the PR I propose, 
-  having the same than in duckdb case `list_tables`, `list_temp_tables`, `list_views`, `list_temp_views` 
-  add `list_foreign_tables` 
- `con.tables`: list them all 

@gforsyth @cpcloud does this sound reasonable?

gforsyth:
> does this sound reasonable?

yeah, I think that sounds good!

PR #9840, opened 2024-08-14
refactor(impala): remove impala table

### LINKED ISSUE #9365: refactor: remove `ImpalaTable` subclass
We used to have a few `ibis.Table` subclasses adding backend-specific functionality to a `Table` object. These have all been removed *except* for `ImpalaTable`. We should remove this to standardize things across backends.

jcrist:
This removes the `ImpalaTable` subclass. Relevant methods that were on that class have been moved to the `Backend` class instead.

We used to have backend-specific subclasses of `Table` that added some backend-specific methods. Most of those were removed in prior releases, `ImpalaTable` was the last remaining one.

As this is a breaking change, we should hold off on merging this until we're ready to start prepping for a 10.0 release. Depends on #9839.

Fixes #9365.


contang0:
What impact will this have on Impala backend users, if any?

gforsyth:
> What impact will this have on Impala backend users, if any?

The only impact on Impala users is the removal of certain methods on the (now removed) `ImpalaTable` class.  So while previously with Impala you could do something like `my_table.insert(some_rows)`, now you have to do `con.insert(my_table, some_rows)`, which is consistent with all the other backends.

gforsyth (approved):
:shipit:

PR #9843, opened 2024-08-14
refactor(api): use as_* for type coercion methods

### LINKED ISSUE #9788: refactor(api): rename `to_*` methods for type coercion to something else
Currently ibis has several `to_*` methods - some compute and return a result (`to_pandas`, `to_polars`, `to_parquet`, ...), while others are for type coercion and continue to build up an expression  (`to_timestamp`, `to_interval`, ...).

I think we should rename the latter methods (with a deprecation period), keeping the `to_*` prefix for things that compute and return a result.

The current list is:

```
IntegerValue.to_timestamp(self, unit)
IntegerValue.to_interval(self, unit)

StringValue.to_timestamp(self, format_str)
StringValue.to_date(self, format_str)

IntervalValue.to_unit(self, target_unit)
```

A few options:

- Use `as_*` prefixes instead. This is what we did for the geospatial API (e.g. `as_ewkb`, `as_binary`, ...)
- Use a more specific verb like `parse_*` for the string ones?
- ?

ncclementi:
Closes #9788

### DIFF
--- a/ibis/backends/tests/test_temporal.py
+++ b/ibis/backends/tests/test_temporal.py
@@ -592,7 +592,7 @@ def test_date_truncate(backend, alltypes, df, unit):
 def test_integer_to_interval_timestamp(
     backend, con, alltypes, df, unit, displacement_type
 ):
-    interval = alltypes.int_col.to_interval(unit=unit)
+    interval = alltypes.int_col.as_interval(unit=unit)
     expr = (alltypes.timestamp_col + interval).name("tmp")
 
     def convert_to_offset(offset, displacement_type=displacement_type):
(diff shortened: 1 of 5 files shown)


ibis/backends/tests/test_temporal.py
```diff
@@ -1260,12 +1267,18 @@ def test_integer_to_timestamp(backend, con, unit):
 @pytest.mark.notimpl(["exasol"], raises=com.OperationNotDefinedError)
 def test_string_to_timestamp(alltypes, fmt):
     table = alltypes
-    result = table.mutate(date=table.date_string_col.to_timestamp(fmt)).execute()
+    result_as = table.mutate(date=table.date_string_col.as_timestamp(fmt)).execute()
+
+    with pytest.warns(FutureWarning, match="v10.0"):
```
ncclementi, line 1272: I had this here, to check that we were still doing the right thing with `to_*` but I think this is causing the pyspark ones to fail. 

From Phillip ins a similar situation: "a few of the driver libraries are really bad about closing sockets (or more generally cleaning up the things that create sockets)"

Maybe we don't test this...

jcrist (approved):
LGTM! Are we fine releasing this before 10.0 (it's just a deprecation, not a breaking change), or do we want to hold off merging until 10.0? cc @cpcloud for thoughts.

cpcloud:
Seems fine to release before 10, to give some time for deprecation.

PR #9844, opened 2024-08-15
feat(sql): support inserts with default constraints

IndexSeek:
## Description of changes

This change intends to support scenarios where a user needs to insert into a table where a table contains DEFAULT values on columns, meaning the incoming object may have fewer columns than the target table schema. 

It's common to insert into tables with DEFAULT constraints; in many cases, these are sequences, and it is a bit tricky to provide the sequence "nextval" approach today. 

This topic was initially brought up in [Zulip](https://ibis-project.zulipchat.com/#narrow/stream/405263-general/topic/insert.20with.20default.20columns.20on.20target.20table).

Here is an example of this in practice:

```py
import ibis
import pandas as pd

con = ibis.duckdb.connect()
con.raw_sql("CREATE TABLE example (a INTEGER DEFAULT 1, b VARCHAR NOT NULL);")

df = pd.DataFrame(
    {
        "a": [1, 2, 3],
        "b": ["foo", "bar", "baz"],
    }
)
```

### Backend error
If the user excludes a required column (e.g., NOT NULL without a DEFAULT) from the container used for insert, the backend will error out with its respective error. 

```py
con.insert("example", df[["a"]])
```

### Successful insert with excluded columns
```py
con.insert("example", df[["b"]])
con.table("example").to_pandas()
```

a | b
-- | --
1 | foo
1 | bar
1 | baz

I'm still working through the tests for this, but wanted to put this out here in case anyone wanted to take a look before I can return to it.

### DIFF
--- a/ibis/backends/sql/__init__.py
+++ b/ibis/backends/sql/__init__.py
@@ -439,7 +439,6 @@ class SQLBackend(BaseBackend, _DatabaseSchemaHandler):
             columns=[sg.to_identifier(col, quoted=quoted) for col in columns],
             dialect=compiler.dialect,
         )
-
         return query
 
     def _build_insert_template(
--- a/ibis/backends/sql/__init__.py
+++ b/ibis/backends/sql/__init__.py
@@ -425,10 +425,13 @@ class SQLBackend(BaseBackend, _DatabaseSchemaHandler):
         # Compare the columns between the target table and the object to be inserted
         # If source is a subset of target, use source columns for insert list
         # Otherwise, assume auto-generated column names and use positional ordering.
-        source_cols = source.columns
-        target_cols = self.get_schema(target).names
+        target_cols = self.get_schema(target).keys()
 
-        columns = source_cols if set(source_cols).issubset(target_cols) else target_cols
+        columns = (
+            source_cols
+            if (source_cols := source.schema().keys()) <= target_cols
+            else target_cols
+        )
 
         query = sge.insert(
             expression=self.compile(source),
--- a/ibis/backends/sql/__init__.py
+++ b/ibis/backends/sql/__init__.py
@@ -423,25 +423,20 @@ class SQLBackend(BaseBackend, _DatabaseSchemaHandler):
         compiler = self.compiler
         quoted = compiler.quoted
         # Compare the columns between the target table and the object to be inserted
-        # If they don't match, assume auto-generated column names and use positional
-        # ordering.
+        # If source is a subset of target, use source columns for insert list
+        # Otherwise, assume auto-generated column names and use positional ordering.
         source_cols = source.columns
-        columns = (
-            target_cols
-            if set(target_cols := self.get_schema(target).names).difference(source_cols)
-            else source_cols
-        )
+        target_cols = self.get_schema(target).names
+
+        columns = source_cols if set(source_cols).issubset(target_cols) else target_cols
 
         query = sge.insert(
             expression=self.compile(source),
             into=sg.table(target, db=db, catalog=catalog, quoted=quoted),
-            columns=[
-                sg.to_identifier(col, quoted=quoted)
-                for col in columns
-                if col in source_cols
-            ],
+            columns=[sg.to_identifier(col, quoted=quoted) for col in columns],
             dialect=compiler.dialect,
         )
+
         return query
 
     def _build_insert_template(
--- a/ibis/backends/sql/__init__.py
+++ b/ibis/backends/sql/__init__.py
@@ -427,17 +427,19 @@ class SQLBackend(BaseBackend, _DatabaseSchemaHandler):
         # ordering.
         source_cols = source.columns
         columns = (
-            source_cols
-            if not set(target_cols := self.get_schema(target).names).difference(
-                source_cols
-            )
-            else target_cols
+            target_cols
+            if set(target_cols := self.get_schema(target).names).difference(source_cols)
+            else source_cols
         )
 
         query = sge.insert(
             expression=self.compile(source),
             into=sg.table(target, db=db, catalog=catalog, quoted=quoted),
-            columns=[sg.to_identifier(col, quoted=quoted) for col in columns],
+            columns=[
+                sg.to_identifier(col, quoted=quoted)
+                for col in columns
+                if col in source_cols
+            ],
             dialect=compiler.dialect,
         )
         return query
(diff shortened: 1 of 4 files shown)


IndexSeek:
I have made more progress on the tests. I got the Oracle backend passing, but it required a hacky solution to ensure the identifier was being quoted.

```py
import ibis.backends.sql.compilers as sc

quoted = getattr(sc, con.dialect.__name__.lower()).compiler.quoted
```

It works, but I wonder if there is a better way. 

I will try to look into the remaining backends that are still failing, which are: Clickhouse, Trino, Druid, Exasol.

ibis/backends/sql/__init__.py
```diff
@@ -427,17 +427,19 @@ def _build_insert_from_table(
...
         )
 
         query = sge.insert(
             expression=self.compile(source),
             into=sg.table(target, db=db, catalog=catalog, quoted=quoted),
-            columns=[sg.to_identifier(col, quoted=quoted) for col in columns],
+            columns=[
+                sg.to_identifier(col, quoted=quoted)
+                for col in columns
+                if col in source_cols
```
cpcloud, line 441: Let's make a variable for `source.schema` and then use `col in variable` to do this lookup.

ibis/backends/tests/test_client.py
```diff
@@ -1750,3 +1753,53 @@ def test_cross_database_join(con_create_database, monkeypatch):
...
+@pytest.mark.notyet(
+    ["druid"], raises=NotImplementedError, reason="doesn't support create_table"
+)
+@pytest.mark.notimpl(["clickhouse"], reason="create table isn't implemented")
+@pytest.mark.notyet(["pandas", "dask", "polars"], reason="Doesn't support insert")
+@pytest.mark.notimpl(
+    ["flink"], reason="Temp tables are implemented as views, which don't support insert"
+)
+def test_insert_into_table_missing_columns(con, monkeypatch):
+    monkeypatch.setattr(ibis.options, "default_backend", con)
```
cpcloud, line 1767: Let's simplify the test by using `sg.parse_one(sql_string, read="whatever dialect you want")`.

IndexSeek: Thank you for this guidance! That made things much tidier.

ibis/backends/sql/__init__.py
IndexSeek: I believe the change I have made will reduce the time complexity to avoid needing to check again against the source_cols list. I have rewritten things a bit here, using a variable named `columns` to refer to the column list to be used for the insert SQL expression.

IndexSeek:
I swapped things around here to use `issubset` - my idea was that if the source contains fewer columns than the target, but the column names are the same, that's fine, as we can use the source's columns in the insert list. 

`issubset` will also evaluate to `True` if the columns are the same in both source and target, regardless of order. In this scenario, we want to use the order of source, so that the SQL expression will be written properly. 

So if my source schema is:
```py
ibis.Schema {
  b  string
  a  int64
}
```

But my target schema is:
```py
ibis.Schema {
  a  int64
  b  string
}
```

This is okay, because the underlying SQL expression will be rewritten to do:
```sql
INSERT INTO target (b, a)
SELECT * 
FROM source;
```

IndexSeek:
I have marked this one ready for review, but I have a couple of questions. 

1. The Oracle backend is still failing, and I wasn't sure if there was a clean way to address that with a SQLGlot method to ensure quoting the identifiers in the Create expression from `parse_one` or if I should adjust the `ct_sql` variable to enclose the object identifiers in double-quotes. My concern is that the Snowflake backend may encounter a similar outcome when cloud runs are triggered. 
2. Druid is marked "notyet," but fails because `raw_sql` is not supported. Should I include that in the list with exasol?

ibis/backends/tests/test_client.py
```diff
@@ -1750,3 +1751,27 @@ def test_cross_database_join(con_create_database, monkeypatch):
...
+)
+@pytest.mark.notimpl(["clickhouse"], reason="create table isn't implemented")
+@pytest.mark.notyet(["pandas", "dask", "polars"], reason="Doesn't support insert")
+@pytest.mark.notyet(["exasol"], reason="Backend does not support raw_sql")
+@pytest.mark.notimpl(
+    ["impala", "pyspark", "trino"], reason="Default constraints are not supported"
+)
+def test_insert_into_table_missing_columns(con):
+    table_name = gen_name("table")
+    ct_sql = f"CREATE TABLE {table_name} (a INT DEFAULT 1, b VARCHAR(10));"
```
cpcloud, line 1767: Definitely need to aggressively quote every identifier here.

cpcloud (changes requested):

ibis/backends/tests/test_client.py
IndexSeek: Good call, thank you! After quoting, I got past that portion of the error, but ran into a different error from VARCHAR(10) getting translated to a CLOB type, which didn't play nicely with my assertion. I just adjusted to test with two int columns this time around. All seems to be working on the Oracle side now.

cpcloud:
I'll add an xfail marker to the Druid backend, since it doesn't support `CREATE TABLE`.

cpcloud:
Oh, looks like you did that already. Let me see what's failing then.

cpcloud:
@IndexSeek I improved the implementation bit: we can use our `Schema.keys()` method (whose output behaves like a `set`) instead of constructing a `set` and calling `issubset` with a sequence (which itself will construct a `set` if the input argument isn't already a `set`).

cpcloud (approved):

IndexSeek:
> @IndexSeek I improved the implementation bit: we can use our `Schema.keys()` method (whose output behaves like a `set`) instead of constructing a `set` and calling `issubset` with a sequence (which itself will construct a `set` if the input argument isn't already a `set`).
> 
> This avoids constructing two throwaway `set`s, the effects of which can show up in wide-table use cases.

This is a very nice improvement. Thank you for the assistance here and the explanation, @cpcloud!

cpcloud:
Running the cloud backend test suite, then if that's all green this is good to merge!

ibis/backends/tests/test_client.py
```diff
@@ -1765,7 +1765,11 @@ def test_cross_database_join(con_create_database, monkeypatch):
     ["impala", "pyspark", "trino"], reason="Default constraints are not supported"
 )
 def test_insert_into_table_missing_columns(con, temp_table):
-    ct_sql = f'CREATE TABLE "{temp_table}" ("a" INT DEFAULT 1, "b" INT)'
+    ident = sg.table(
+        temp_table, db=con.current_database, catalog=con.current_catalog, quoted=True
+    )
+    raw_ident = ident.sql("duckdb")
```
IndexSeek, line 1771: This is cool, I was wondering if there was a clean/easy way to do this with SQLGlot! Hopefully this gets the cloud tests all passing now. 

Thanks again for your help with this one!

PR #9848, opened 2024-08-15
fix(mssql): remove sort key to keep order

### LINKED ISSUE #9846: bug: mssql pandas data can't be converted by correct type
### What happened?

If I send a query statement to mssql, the result data cannot be converted by paired type.

Because ibis changes the `_get_schema_using_query` of mssql at `9.3.0`
```python
# For some reason when using "Latin1_General_100_BIN2_UTF8"
# the stored procedure `sp_describe_first_result_set` starts throwing errors about DLL loading.
# This "dynamic management function" uses the same algorithm and allows
# us to pre-filter the columns we want back.
# The syntax is:
# `sys.dm_exec_describe_first_result_set(@tsql, @params, @include_browse_information)`
query = f"""SELECT name,
                is_nullable AS nullable,
                system_type_name,
                precision,
                scale
            FROM
                sys.dm_exec_describe_first_result_set({tsql}, NULL, 0)"""
with self._safe_raw_sql(query) as cur:
    rows = cur.fetchall()
```
the rows will get
```
[('orderkey', True, 'int', 10, 0), ('custkey', True, 'int', 10, 0), ('orderstatus', True, 'varchar(max)', 0, 0), ('totalprice', True, 'varchar(max)', 0, 0), ('orderdate', True, 'date', 10, 0) .....]
```
different before
```
[(False, 1, 'orderkey', True, 56, 'int', 4, 10, 0, None, None, None, None, None, None, None, None, None, None, False, False, False, None, None, None, None, None, False, None, True, False, False, None, None, None, 38, 4, None, None), (False, 2, 'custkey', True, 56, 'int', 4, 10, 0, None, None, None, None, None, None, None, None, None, None, False, False, False, None, None, None, None, None, False, None, True, False, False, None, None, None, 38, 4, None, None) ......]
```
But we still use `sorted(rows, key=itemgetter(1))` to iterate rows. It will make the order wrong.
```python
for (
    name,
    nullable,
    system_type_name,
    precision,
    scale,
) in sorted(rows, key=itemgetter(1)):
```

9.3.0

mssql

### Relevant log output

```sh
../app/model/connector.py:19: in query
    return self.connection.sql(sql).limit(limit).to_pandas()
../.venv/lib/python3.11/site-packages/ibis/expr/types/relations.py:3568: in to_pandas
    return self.execute(**kwargs)
../.venv/lib/python3.11/site-packages/ibis/expr/types/core.py:404: in execute
../.venv/lib/python3.11/site-packages/ibis/backends/sql/__init__.py:289: in execute
    result = self._fetch_from_cursor(cur, schema)
../.venv/lib/python3.11/site-packages/ibis/backends/sql/__init__.py:106: in _fetch_from_cursor
    df = PandasData.convert_table(df, schema)
../.venv/lib/python3.11/site-packages/ibis/formats/pandas.py:119: in convert_table
    columns.append(cls.convert_column(series, dtype))
../.venv/lib/python3.11/site-packages/ibis/formats/pandas.py:151: in convert_column
    result = convert_method(obj, dtype, pandas_type)
../.venv/lib/python3.11/site-packages/ibis/formats/pandas.py:234: in convert_Date
    return s.map(try_date, na_action="ignore")
../.venv/lib/python3.11/site-packages/pandas/core/series.py:4700: in map
    new_values = self._map_values(arg, na_action=na_action)
../.venv/lib/python3.11/site-packages/pandas/core/base.py:921: in _map_values
    return algorithms.map_array(arr, mapper, na_action=na_action, convert=convert)
../.venv/lib/python3.11/site-packages/pandas/core/algorithms.py:1745: in map_array
    return lib.map_infer_mask(
lib.pyx:2895: in pandas._libs.lib.map_infer_mask
    ???
lib.pyx:2932: in pandas._libs.lib._map_infer_mask
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

v = '1_370'

    def try_date(v):
        if isinstance(v, datetime.datetime):
            return v.date()
        elif isinstance(v, str):
            if v.endswith("Z"):
                return datetime.datetime.fromisoformat(v[:-1]).date()
>           return datetime.date.fromisoformat(v)
E           ValueError: Invalid isoformat string: '1_370'

../.venv/lib/python3.11/site-packages/ibis/formats/pandas.py:230: ValueError
```

### Code of Conduct

grieve54706:
## Description of changes
Iterate rows directly without sorting by key.

## Issues closed
Resolves #9846

### DIFF
--- a/ibis/backends/mssql/__init__.py
+++ b/ibis/backends/mssql/__init__.py
@@ -244,24 +244,21 @@ class Backend(SQLBackend, CanCreateCatalog, CanCreateDatabase, CanCreateSchema, 
         # us to pre-filter the columns we want back.
         # The syntax is:
         # `sys.dm_exec_describe_first_result_set(@tsql, @params, @include_browse_information)`
-        query = f"""SELECT name,
-                        is_nullable AS nullable,
-                        system_type_name,
-                        precision,
-                        scale
-                    FROM
-                        sys.dm_exec_describe_first_result_set({tsql}, NULL, 0)"""
+        query = f"""
+        SELECT
+          name,
+          is_nullable,
+          system_type_name,
+          precision,
+          scale
+        FROM sys.dm_exec_describe_first_result_set({tsql}, NULL, 0)
+        ORDER BY column_ordinal
+        """
         with self._safe_raw_sql(query) as cur:
             rows = cur.fetchall()
 
         schema = {}
-        for (
-            name,
-            nullable,
-            system_type_name,
-            precision,
-            scale,
-        ) in sorted(rows, key=itemgetter(1)):
+        for name, nullable, system_type_name, precision, scale in rows:
             newtyp = self.compiler.type_mapper.from_string(
                 system_type_name, nullable=nullable
             )
--- a/ibis/backends/mssql/tests/test_client.py
+++ b/ibis/backends/mssql/tests/test_client.py
@@ -81,6 +81,29 @@ def test_get_schema(con, server_type, expected_type, temp_table):
     assert con.sql(f"SELECT * FROM [{temp_table}]").schema() == expected_schema
 
 
+def test_schema_type_order(con, temp_table):
+    columns = []
+    pairs = {}
+
+    for i, db_type in enumerate(DB_TYPES):
+        if isinstance(db_type, tuple) and len(db_type) == 2:
+            server_type, expected_type = db_type
+            column_name = f"col_{i}"
+            columns.append(f"{column_name} {server_type}")
+            pairs[column_name] = expected_type
+
+    stmt = f"CREATE TABLE [{temp_table}] ({', '.join(columns)})"
+
+    with con.begin() as c:
+        c.execute(stmt)
+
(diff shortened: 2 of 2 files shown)


cpcloud:
This isn't the correct way to fix the problem (which is definitely a problem!).

The code here was changed, but the `sorted` key was not updated to reflect the change.

We still need to sort by the column's id (an integer indicating its position).

Thank you for adding a test. I will push up the correct changes.

cpcloud (approved):

gforsyth (approved):

grieve54706:
> This isn't the correct way to fix the problem (which is definitely a problem!).
> 
> The code here was changed, but the `sorted` key was not updated to reflect the change.
> 
> We still need to sort by the column's id (an integer indicating its position).
> 
> Thank you for adding a test. I will push up the correct changes.

Hi @cpcloud, thank you for your close.

PR #9855, opened 2024-08-16
test(duckdb): add test for supporting fixed-size array types

### LINKED ISSUE #7963: feat(duckdb): handle fixed-length `ARRAY` type

Not related to a problem

Support for [DuckDB's fixed-length `ARRAY`](https://duckdb.org/docs/sql/data_types/array) type.

I don't think we need to add a new type here for now. We can infer these as `dt.Array(T)` where `T` is the element type of the `ARRAY`

`main`

DuckDB

### Code of Conduct

jcrist:
These are already supported automatically through sqlglot, just adding a test here. Fixes #7963.

One open question: currently these are cast to variable-length lists in `to_pyarrow` (matching the existing behavior of `dt.Array` types), but it could be argued that we should pass them through as fixed-length lists instead. This would require a bit of work, but nothing too tricky.


jcrist:
Failure is unrelated.

cpcloud (approved):

PR #9856, opened 2024-08-16
feat(api): add `DateValue.epoch` api for computing days since epoch

cpcloud:
## Description of changes

Adds `DateValue.epoch` method and underlying `ops.UnixDate`.

This was originally meant to help with window function queries with range
bounds where the backend doesn't support intervals in preceding/following, but
I think it has general utility and is much more clear IMO than casting a
date value to an integer.

### DIFF
--- a/ibis/backends/sql/dialects.py
+++ b/ibis/backends/sql/dialects.py
@@ -4,7 +4,6 @@ import contextlib
 import math
 from copy import deepcopy
 
-import sqlglot as sg
 import sqlglot.expressions as sge
 import sqlglot.generator as sgn
 from sqlglot import transforms
@@ -28,32 +27,6 @@ ClickHouse.Generator.TRANSFORMS |= {
     sge.ArraySort: rename_func("arraySort"),
     sge.LogicalAnd: rename_func("min"),
     sge.LogicalOr: rename_func("max"),
-    sge.DateFromParts: lambda self, e: sg.func(
-        "toDate",
-        sg.func(
-            "concat",
-            sg.func(
-                "leftPad",
-                sg.func("toString", e.args["year"]),
-                sge.convert(4),
-                sge.convert("0"),
-            ),
-            sge.convert("-"),
-            sg.func(
-                "leftPad",
-                sg.func("toString", e.args["month"]),
-                sge.convert(2),
-                sge.convert("0"),
-            ),
-            sge.convert("-"),
-            sg.func(
-                "leftPad",
-                sg.func("toString", e.args["day"]),
-                sge.convert(2),
-                sge.convert("0"),
-            ),
-        ),
-    ).sql(self.dialect),
 }
 
 
@@ -69,7 +42,6 @@ class DataFusion(Postgres):
             sge.Array: rename_func("make_array"),
             sge.ArrayContains: rename_func("array_has"),
             sge.ArraySize: rename_func("array_length"),
-            sge.DateFromParts: rename_func("make_date"),
         }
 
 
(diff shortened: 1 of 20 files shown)


gforsyth:
Propose to codify the word `uxfail` as you've implicitly defined it above.

jcrist:
Is this operation semantically the same as `date_col.delta(ibis.date(1970, 1, 1), "d")`? If so, is there a reason why that's insufficient? We might even add an `ibis.epoch()` function that's effectively an alias for `ibis.date(1970, 1, 1)` in this context. I find the `delta` spelling a bit easier to understand just from looking at it.

Alternatively, could we spell this as `epoch_days()` (to match the existing `epoch_seconds`)? I find the `epoch()` method to be a bit confusing, since it's not clear to me from the name alone what it's referencing (`days_since_epoch` would be even clearer, but :shrug:).

cpcloud:
> Is this operation semantically the same as `date_col.delta(ibis.date(1970, 1, 1), "d")`? If so, is there a reason why that's insufficient? We might even add an `ibis.epoch()` function that's effectively an alias for `ibis.date(1970, 1, 1)` in this context. I find the `delta` spelling a bit easier to understand just from looking at it.

It's not insufficient, just more verbose than I'd like.

> Alternatively, could we spell this as `epoch_days()` (to match the existing `epoch_seconds`)? I find the `epoch()` method to be a bit confusing, since it's not clear to me from the name alone what it's referencing (`days_since_epoch` would be even clearer, but 🤷).

`epoch_days` seems reasonable and keeps consistency with `epoch_seconds`.

My original plan was to have `epoch` on `DateValue` and `TimestampValue` (with the latter having a `unit` parameter) but I think three separate methods `epoch_secs`/`epoch_millis`/`epoch_micros` is probably better (with an eventual `epoch_seconds` deprecation).

ibis/expr/types/temporal.py
```diff
@@ -474,6 +474,43 @@ def delta(
...
+        >>> d = date(2020, 1, 1)
+        >>> ibis.date(d)
+        ┌────────────┐
+        │ 2020-01-01 │
+        └────────────┘
+        >>> ibis.date(d).epoch_days()
+        ┌───────┐
+        │ 18262 │
+        └───────┘
+        >>> t = ibis.memtable({"date_col": [d]})
```
jcrist, line 494: Slightly simpler example, no need to involve `datetime.date` in here at all.

```suggestion
        >>> import ibis
        >>> ibis.options.interactive = True
        >>> date = ibis.date(2020, 1, 1)
        >>> date
        ┌────────────┐
        │ 2020-01-01 │
        └────────────┘
        >>> date.epoch_days()
        ┌───────┐
        │ 18262 │
        └───────┘
        >>> t = date.name("date_col").as_table()
```

ibis/backends/sql/dialects.py
```diff
@@ -27,6 +28,32 @@
     sge.ArraySort: rename_func("arraySort"),
     sge.LogicalAnd: rename_func("min"),
     sge.LogicalOr: rename_func("max"),
+    sge.DateFromParts: lambda self, e: sg.func(
```
jcrist, line 31: I actually find reading these in the sqlglot dialect harder to understand than the previous version in the compiler. These are pretty complicated implementations (length-wise), doing more in the dialect means we have additional places to check for complex logic when trying to understand how something is compiled (compiler, rewrite rules, our `dialects.py` file, sqlglot's dialects, sqlglot's rewrite rules). If possible, I'd find it easier to understand compilation if we try to keep most of our implemented logic in the compiler itself rather than the dialects here.

Just a -0.25 though, if you feel strongly about doing it this way then that's fine.

ibis/expr/types/temporal.py
```diff
@@ -474,6 +474,43 @@ def delta(
...
+        >>> t.mutate(epoch=t.date_col.epoch_days())
+        ┏━━━━━━━━━━━━┳━━━━━━━┓
+        ┃ date_col   ┃ epoch ┃
+        ┡━━━━━━━━━━━━╇━━━━━━━┩
+        │ date       │ int64 │
+        ├────────────┼───────┤
+        │ 2020-01-01 │ 18262 │
+        └────────────┴───────┘
+        """
+        return ops.EpochDays(self).to_expr()
```
jcrist, line 512: Do we need a new op here? Is there a benefit to not defining this in terms of the existing `delta` method + a literal for the epoch? Doing it that way would avoid the need to define this op for every backend, instead relying on the more general `delta`.

jcrist (commented):
Overall happy with the user-facing API, just a few comments on the implementation.

ibis/backends/sql/dialects.py
cpcloud: Unfortunately without these (or something like it), it's a non-starter to implement this without using a new operation. Happy to try the no-new-op route, but to do that we need sqlglot to correctly compile `DateFromParts` correctly for all the backends it can, which requires porting the previous code to this.

cpcloud: In general I agree with your point, but this happened to be an annoying case where it seemed better to centralize this logic in sqlglot rather than repeating in the compiler.

ibis/expr/types/temporal.py
cpcloud: Oh, I see. I misinterpreted `delta` as not-the-method-but-the-concept, in which case, yeah maybe we can avoid the sqlglot munging and also a new op.

jcrist (approved):

jcrist:
LGTM! Leaving the merge to you - I assume you want to squash some/all of these commits.

PR #9860, opened 2024-08-16
feat(pyspark, exasol): implement `quantile`

jcrist:
Also removes the base implementation for `Quantile`/`MultiQuantile` - it was only used by `postgres`. Moving it to the postgres compiler lets backends lacking support avoid adding it to `UNSUPPORTED_OPS`.


cpcloud (approved):
Nice!

PR #9863, opened 2024-08-17
docs(repl): remove hack to support ibis-in-the-browser

cpcloud:
Aspirational PR to remove hacks that support ibis in the browser. Based on the dask-removal PR, which makes using python 3.12 a lot easier. Python 3.12 is required for later versions of Pyodide.


cpcloud:
This depends on a release of pyodide, which hopefully will happen in the next couple of weeks.

PR #9870, opened 2024-08-19
feat(exasol): implement a bunch of missing aggs

jcrist:
In a previous PR I noticed that the exasol backend had a bunch of xfails with `strict=False` that were masking deeper weirdness in the exasol backend. This removes the `strict=False`, fixes some tests, and expands coverage of aggregations in exasol.


gforsyth (commented):
Looks good, I think just need to restore the `ExaQueryError` import to fix up the tests.

gforsyth (approved):

jcrist:
Gah, rebasing turned up some more un-xfailing tests. I _think_ I've fixed everything now.

PR #9880, opened 2024-08-20
docs(blog): ibis + clickhouse + shiny for better pypi stats

lostmygithubaccount:
## Description of changes

Ibis + ClickHouse + Shiny for Python = better PyPI stats.

Adds a blog post showing an end-to-end Python analytics + BI stack on public data.


ncclementi:
Can we replace this Pie chart for a bar plot? It's very hard for the eye to distinguish the different of the small percentages, especially the smaller ones. (Also happy to provide more context on why pie-charts are not a good way of communicating data)

Also the colors seem to be repeated, pip and conda look the same. Maybe plot just the top 5? 

Screenshot 2024-08-20 at 6 05 52 PM

lostmygithubaccount:
will do -- it's actually not in the shiny app

TODOs:

- [ ] update pie chart to something else or remove
- [ ] record and publish YouTube video for dashboard section
- [ ] thumbnail

I probably also want to get the part 2 of this ready before publishing (it's 80% done)

docs/posts/better-pypi-stats/index.qmd
```diff
@@ -0,0 +1,358 @@
...
+Let's start by looking at the most downloaded packages:
+
+```{python}
+overall_t = con.table("pypi_downloads")
+
+top_k = 10_000
+overall_t = (
+    overall_t.order_by(ibis.desc("count"))
+    .limit(top_k)
+    .mutate(rank=1 + ibis.row_number().over(order_by=ibis.desc("count")))
```
cpcloud, line 80: Is adding one here necessary? It seems like you're just after the ordering, which is there without the addition of one.

cpcloud (approved):

PR #9881, opened 2024-08-20
feat(api): approximate quantiles

### LINKED ISSUE #9541: feat(bigquery): general `approx_quantile` similar to `approx_median` but for arbitrary quantile

There isn't a great way to use BigQuery's APPROX_QUANTILES feature to get values besides the [approximate median](https://ibis-project.org/reference/expression-generic.html#ibis.expr.types.generic.Column.approx_median) in Ibis.

To mimic pandas [`describe`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.describe.html), BigQuery DataFrames uses APPROX_QUANTILES to get an approximate 25th percentile, median, and 75th percentile.

Note: the percentiles are configurable in pandas, but unfortunately BigQuery SQL's # of bins approach makes it difficult to support arbitrary percentiles.

From BigQuery DataFrames, perspective, it'd be great if there was an API to get evenly-spaced approximate quantiles, but perhaps a approx_quantile function that takes an integer from 0 - 100 would be more flexible? Or 0 to 1 to mimic pandas, but with the note that some backends like bigquery only support precision up to a certain point (maybe nearest 0.05 or nearest 0.01)?

8.x, but working on 9.x upgrade

BigQuery

### Code of Conduct

jcrist:
This adds a new `approx_quantile` method for computing approximate quantiles. Like the corresponding exact version, it can take:
- A single quantile (`col.approx_quantile(0.5)`), returning a single scalar result
- An array of quantiles (`col.approx_quantile([0.25, 0.75])`), returning an array of results

Unlike `col.quantile`, this is only implemented for numeric types (for now). Most backends support only numeric values for the approximate version, as the algorithms don't easily adapt to non-numeric types. If needed, we _could_ move this to be on a generic type, but most implementations would error for other dtypes.

Fixes #9541.

### DIFF
--- a/ibis/backends/sql/compilers/snowflake.py
+++ b/ibis/backends/sql/compilers/snowflake.py
@@ -608,6 +608,12 @@ $$""",
         quantile = self.f.percentile_cont(quantile)
         return sge.WithinGroup(this=quantile, expression=order_by)
 
+    def visit_ApproxQuantile(self, op, *, arg, quantile, where):
+        if where is not None:
+            arg = self.if_(where, arg, NULL)
+
+        return self.f.approx_percentile(arg, quantile)
+
     def visit_CountStar(self, op, *, arg, where):
         if where is None:
             return super().visit_CountStar(op, arg=arg, where=where)
--- a/ibis/expr/operations/reductions.py
+++ b/ibis/expr/operations/reductions.py
@@ -211,6 +211,11 @@ class Median(QuantileBase):
 
 
 @public
+class ApproxMedian(Median):
+    """Compute the approximate median of a column."""
+
+
+@public
 class Quantile(QuantileBase):
     """Compute the quantile of a column."""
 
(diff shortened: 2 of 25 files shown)


ibis/backends/sql/compilers/snowflake.py
```diff
@@ -608,6 +608,12 @@ def visit_Quantile(self, op, *, arg, quantile, where):
         quantile = self.f.percentile_cont(quantile)
         return sge.WithinGroup(this=quantile, expression=order_by)
 
+    def visit_ApproxQuantile(self, op, *, arg, quantile, where):
+        if where is not None:
+            arg = self.if_(where, arg, NULL)
+
+        return self.f.approx_percentile(arg, quantile)
```
cpcloud, line 615: I guess we can't chuck this into `SIMPLE_OPS`?

ibis/expr/operations/reductions.py
```diff
@@ -210,13 +210,25 @@ class Median(QuantileBase):
     """Compute the median of a column."""
 
 
+@public
+class ApproxMedian(Median):
```
cpcloud, line 214: Any reason to subclass `Median` here instead of `QuantileBase`?

cpcloud (approved):
LGTM, a couple of questions!

ibis/backends/sql/compilers/snowflake.py
jcrist: No - similar to `group_concat` some backends get grumpy if `quantile` isn't a literal.

ibis/expr/operations/reductions.py
jcrist: Small refactor to ensure all approximate methods have consistent typing with their non-approximate versions.

PR #9885, opened 2024-08-21
feat(api): support `SchemaLike` in `Backend.create_table()`

NickCrews:
This requires a test and implementation for other backends, but I wanted to get some feedback before I went for it.

Currently, you have to pass a real ibis.Schema to create_table(). Now it supports dicts and iterables of (name, type) pairs.
This is now consistent with `ibis.table()`.

Example usage:

```python
import ibis
db= ibis.duckdb.connect()
db.create_table(
    "t",
    {"x": []},
    schema={"x": "string"},
)
```

currently, this errors. With this change, it works.

### DIFF
--- a/ibis/backends/bigquery/__init__.py
+++ b/ibis/backends/bigquery/__init__.py
@@ -923,7 +923,7 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema):
         | pl.LazyFrame
         | None = None,
         *,
-        schema: ibis.Schema | None = None,
+        schema: sch.SchemaLike | None = None,
         database: str | None = None,
         temp: bool = False,
         overwrite: bool = False,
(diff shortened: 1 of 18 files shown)


cpcloud (commented):
Seems like a good idea to me!

NickCrews:
Ideally in the docs, I would want the type annotations to boil down to the base types, and show that schema can be a Mapping, an ibis Schema, an interable of str,str pairs, etc. But, if we use the nice shorthand `SchemaLike` typealias, does it show up literally as that in the type hints? I triggered a docs preview to see. If it only shows the alias, I think I would vote for being verbose and writing out the whole thing, I want the documentation to be clear. Others have thoughts here?

NickCrews:
I just threw up the test and implemented it for a few backends, let's see which actually fail and need an implementation change

NickCrews:
hmm, darn, yeah the docs render as the opaque "SchemaLike". Any opinions here on what this should look like?

The [quartodoc documentation](https://machow.github.io/quartodoc/get-started/overview.html) is intimidating for me, I'm not sure if there is some option in there of `simplify_type_aliases=True` (I doubt it exists). If that indeed doesn't exist, then should I just be verbose and write out the full type?
image

cpcloud:
Let's really try to avoid getting caught up in the docs issue so that the PR can be merged quickly. The rendering of types in docs is separate from this PR.

cpcloud (approved):

PR #9886, opened 2024-08-21
docs(blog): walking talking cube and os lms for data

lostmygithubaccount:
## Description of changes

Adds two blogs: "Taking a random cube for a walk and making it talk" and "Using open source language models for data".

The first aims to be short, fun, and explore randomized synthetic data generation with Ibis. It leads into the seconds, generating better fake data with language models. This is used as an opportunity to draw a parallel on separating the interface from the backend (i.e. the language model tooling space is...complicated...right now). After some testing, I realized that `logit_bias` doesn't work via Ollama. Some re-writing and generally more work is needed on the second blog, though I'd like to keep them in one PR and releasing on the same day.


docs/posts/walking-talking-cube/index.qmd
```diff
@@ -0,0 +1,286 @@
...
+    s.across(
+        s.c("a", "b", "c"),  # <1>
+        ibis._.sum().over(window),  # <2>
+    ),
+).order_by("timestamp")
+walked
+```
+
+1. Alternatively, you can use `s.of_type(float)` to select all float columns.
+2. Use the `ibis._` selector to reference a deferred column expression.
```
cpcloud, line 132:
```suggestion
2. Use the `ibis._` object to reference a deferred column expression.
```

`_` isn't a selector

PR #9895, opened 2024-08-22
refactor(internals): remove `cached_property` on the `tables` accessor to avoid a reference cycle

cpcloud:
## Description of changes

This PR removes circular references to backends `TablesAccessor`.

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -41,10 +41,9 @@ class TablesAccessor(collections.abc.Mapping):
     >>> con = ibis.sqlite.connect("example.db")
     >>> people = con.tables["people"]  # access via index
     >>> people = con.tables.people  # access via attribute
-
     """
 
-    def __init__(self, backend: BaseBackend):
+    def __init__(self, backend: BaseBackend) -> None:
         self._backend = backend
 
     def __getitem__(self, name) -> ir.Table:
@@ -1005,7 +1004,7 @@ class BaseBackend(abc.ABC, _FileIOHandler):
 
         """
 
-    @functools.cached_property
+    @property
     def tables(self):
         """An accessor for tables in the database.
 


ibis/backends/duckdb/tests/test_client.py
```diff
@@ -403,3 +404,23 @@ def test_read_csv_with_types(tmp_path, input, all_varchar):
...
+    before = len(gc.get_referrers(con))
+    tables = con.tables
+    after = len(gc.get_referrers(con))
+
+    assert after == before
+
+    # valid call, and there are no tables in the database
+    assert not list(tables)
+
+    del con
```
gforsyth, line 422: Is there a reason this in in the `duckdb` tests instead of the general backend tests?

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -403,3 +404,23 @@ def test_read_csv_with_types(tmp_path, input, all_varchar):
...
+
+    # no longer valid because the backend has been manually decref'd
+    with pytest.raises(ReferenceError):
+        list(tables)
```
gforsyth, line 426: I think the answer is no, but can we set the message that gets raises by the `ReferenceError`? Something like "The connection associated with this table set has been closed."

ibis/backends/impala/__init__.py
```diff
@@ -205,7 +204,7 @@ def from_connection(cls, con: hs2.HiveServer2Connection) -> Backend:
     def _post_connect(self) -> None:
         self.options = {}
 
-    @cached_property
+    @property
```
ncclementi, line 207: for learning purposes, what's the reason of moving away from a cached property. Isn't it `version` something that we would want to avoid executing multiple times?

cpcloud: Actually, I changed these based on misunderstanding of `cached_property`. It turns out that it uses regular instance attributes. I thought it was using `lru_cache` and when that's used as a decorator on a method, it leaks memory due to the closed-over cache holding an instance of the class until the interpreter shuts down.

So, the reason is "I didn't understand how `cached_property` worked", but now that I do, I'll revert these.

ibis/backends/duckdb/tests/test_client.py
cpcloud: Yes, because the test requires no other references to a backend in order to pass, and when `con` is used as a fixture there's more than one reference out.

gforsyth: ahh, gotcha

ibis/backends/duckdb/tests/test_client.py
cpcloud: The error is somewhat reasonable IMO. I'd rather avoid raising another exception if possible.

Here's the error:

```
>>> import ibis
>>> con = ibis.duckdb.connect()
>>> tables = con.tables
>>> tables
Tables
------
>>> del con
>>> tables
  File "/home/cloud/src/github.com/ibis/ibis/backends/__init__.py", line 88, in __repr__
    tables = self._backend.list_tables()
ReferenceError: weakly-referenced object no longer exists
```

cpcloud: I agree that it's probably a bit too low-level to be useful, but introducing a new exception to handle this seems a bit overkill. I'm open to it though if you feel strongly about it.

ibis/backends/impala/__init__.py
cpcloud: Reverted that change.

ibis/backends/duckdb/tests/test_client.py
cpcloud: One thing that's not great about introducing a new exception is that I'd have to introduce a new property to catch all `ReferenceError`s and re-raise our custom exception, and I'd have to do that everywhere a `weakref.proxy` is used.

gforsyth: I don't feel strongly enough about it to merit wrapping the exception.  My hope was that there was a way of tweaking the text in the `ReferenceError` but I don't believe that's possible without catching and re-raising

gforsyth: Also, I think this path is likely to be relatively uncommon

ibis/backends/__init__.py
```diff
@@ -805,12 +812,7 @@ def __init__(self, *args, **kwargs):
         self._con_args: tuple[Any] = args
         self._con_kwargs: dict[str, Any] = kwargs
         self._can_reconnect: bool = True
-        # expression cache
-        self._query_cache = RefCountedCache(
-            populate=self._load_into_cache,
-            lookup=lambda name: self.table(name).op(),
-            finalize=self._clean_up_cached_table,
-        )
+        self._query_cache = RefCountedCache(weakref.proxy(self))
```
cpcloud, line 815: `RefCountedCache` should probably be renamed to `Cache` or just anything that doesn't spell out its implementation, which isn't even correct anymore.

cpcloud:
Actually I'm going to split the cache changes out into a separate PR, which I don't think is a breaking change, and leave this PR just for the `.tables` accessor changes which are a breaking change.

cpcloud:
This is failing because of the references held by the cache, so this PR now requires #9903 to pass.

ibis/backends/__init__.py
```diff
@@ -1018,7 +1025,7 @@ def tables(self):
         >>> people = con.tables.people  # access via attribute
 
         """
-        return TablesAccessor(self)
+        return TablesAccessor(weakref.proxy(self))
```
jcrist, line 1028: FWIW removing the `cached_property` here also seems sane (and simpler). Creating a `TablesAccessor` is super cheap, no reason to hold onto a reference to it on the backend. If that'd work, I'd prefer that to using weakrefs here at all.

cpcloud: Yep, agree. Let's leave out the `weakref`s until there's a more concrete problem they solve.

ibis/backends/duckdb/tests/test_client.py
cpcloud: Removed weakrefs altogether.

jcrist (approved):

PR #9902, opened 2024-08-22
feat(selectors): allow naming in the `across` selector and support non-expression methods

### LINKED ISSUE #9900: bug: methods like `get_name()` and `type()` do not work with underscore API
### What happened?

When I call various methods of Ibis column expressions using the underscore API, it raises various errors. For example:

```py
import ibis
from ibis import _
import ibis.selectors as s
import pandas as pd

df = pd.DataFrame(data={'col1': [1, 2],
                        'col2': [3, 4],
                        'col3_secret': [5, 6],
                        'col4': [7, 8]})
t = ibis.memtable(df)

t.mutate(
  s.across(
    s.all(),
    ibis.null(type=_.type()) if _.get_name().endswith('_secret') else _
  )
)
```

The errors seem to vary depending on which specific method is called.

I can use a lambda to work around this:

```py
t.mutate(
  s.across(
    s.all(),
    lambda x: ibis.null(type=x.type()) if x.get_name().endswith('_secret') else x
  )
)
```

It'd be great if this worked with the underscore API. Thanks!

9.1.0

_No response_

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Test method calls in across selector; allow naming to work in across calls.

Closes #9900.

### DIFF
--- a/ibis/common/tests/test_deferred.py
+++ b/ibis/common/tests/test_deferred.py
@@ -6,6 +6,7 @@ import pickle
 import pytest
 from pytest import param
 
+import ibis
 from ibis.common.bases import Slotted
 from ibis.common.collections import FrozenDict
 from ibis.common.deferred import (
@@ -296,6 +297,9 @@ class ColumnMock(ValueMock):
 
     def __deferred_repr__(self):
         return f"<column[{self.dtype}]>"
+
+    def type(self):
+        return self.dtype
 
 
 class UnaryMock(ValueMock):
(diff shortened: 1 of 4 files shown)


cpcloud:
@ianmcook I think this gets you most or all of the way to your use case.

`_.type()` and `_.get_name()` already work (e.g., if you did a redundant set of casts for example), but `null` wasn't deferrable (now is in this PR).

gforsyth (approved):
:shipit:

PR #9903, opened 2024-08-22
refactor(internals): simplify the query cache

cpcloud:
Remove unnecessary generalization and parameterization of inputs to the query cache.

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -16,7 +16,7 @@ import ibis.config
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
 from ibis import util
-from ibis.common.caching import RefCountedCache
+from ibis.common.caching import QueryCache
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Mapping, MutableMapping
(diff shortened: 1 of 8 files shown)


ibis/backends/__init__.py
```diff
@@ -805,12 +806,7 @@ def __init__(self, *args, **kwargs):
         self._con_args: tuple[Any] = args
         self._con_kwargs: dict[str, Any] = kwargs
         self._can_reconnect: bool = True
-        # expression cache
-        self._query_cache = RefCountedCache(
-            populate=self._load_into_cache,
-            lookup=lambda name: self.table(name).op(),
-            finalize=self._clean_up_cached_table,
-        )
+        self._query_cache = QueryCache(weakref.proxy(self))
```
cpcloud, line 809: Actually, the `weakref` here is pointless, because `ops.DatabaseTable`s produced by the `Table.cache()` method hold a strong reference to the backend.

While I don't like that tables can outlive the backend, I'm not it's reasonable to artificially disable that without a strong motivation.

The cleanups are still beneficial here, so I'll keep this PR open and cleanup the title and description and remove the `weakref`.

ibis/backends/polars/__init__.py
```diff
@@ -407,6 +403,7 @@ def create_view(
     def drop_table(self, name: str, *, force: bool = False) -> None:
         if name in self._tables:
             del self._tables[name]
+            self._context.unregister(name)
```
cpcloud, line 406: This is a bug, so I'll break it out into a separate change to ensure it has test coverage.

cpcloud:
Closing in favor of #9938.

PR #9904, opened 2024-08-22
feat(api): support deferred objects in `literal`

cpcloud:
Allow passing deferred inputs to `literal`.

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -2427,6 +2427,7 @@ def null(type: dt.DataType | str | None = None) -> Value:
 
 
 @public
+@deferrable
 def literal(value: Any, type: dt.DataType | str | None = None) -> Scalar:
     """Create a scalar expression from a Python value.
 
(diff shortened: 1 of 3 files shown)


ncclementi (commented):
Is there a place in the docs, where we could add an example of this? I'm trying to think of how would one would use it and why.  

We have this https://ibis-project.org/how-to/analytics/chain_expressions but I'm not sure it is the right place

ncclementi (approved):

ibis/expr/types/generic.py
```diff
@@ -2480,6 +2481,33 @@ def literal(value: Any, type: dt.DataType | str | None = None) -> Scalar:
...
+
+    Literals can also be used in a deferred context.
+
+    Here's an example of constructing a table of a column's type repeated for
+    every row:
+
+    >>> from ibis import _, selectors as s
+    >>> ibis.options.interactive = True
+    >>> t = ibis.examples.penguins.fetch()
+    >>> t.select(s.across(s.all(), ibis.literal(_.type(), type=str).name(_.get_name())))
```
jcrist, line 2493: This is kind of (IMO) an odd use case, and it takes up a bunch of vertical docstring space (all rows have the same value). Any way you could limit the output (`.limit(1)` perhaps) a bit just to improve the ratio of `good-use-of-literal:odd-use-of-literal` in this doc :)?

jcrist (approved):
Seems harmless. The reason we didn't support this before is that `ibis.literal` doesn't take an expression for any input type. The use case of using `deferred` in the `selectors` API to get non-expression values never occurred to me. Not a common use case, but easy enough to support.

ibis/expr/types/generic.py
cpcloud: Yeah, agree.

cpcloud: Done.

PR #9908, opened 2024-08-23
feat: Support read_csv for backends with no native support

jitingxu1:
## Description of changes

For backends that lack native read_csv support, `pyarrow.csv.read_csv()` will be used.

- Read a single file
- Read all files in a directory, something like this:  `./directory/*`
- Read files matching a glob pattern
- Support cloud storage systems like S3 and GCP

## Issues closed

Partially solve #9448

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -3,13 +3,13 @@ from __future__ import annotations
 import abc
 import collections.abc
 import functools
+import glob
 import importlib.metadata
 import keyword
 import re
 import urllib.parse
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, ClassVar
-import glob
 
 import ibis
 import ibis.common.exceptions as exc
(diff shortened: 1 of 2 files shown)


ibis/backends/tests/test_register.py
```diff
@@ -390,10 +394,12 @@ def test_csv_reregister_schema(con, tmp_path):
         "clickhouse",
         "dask",
         "datafusion",
-        "flink",
+        "druid",
+        "exasol" "flink",
```
cpcloud, line 398: You're missing a comma after `"exasol"`.

Python string literals are automatically concatenated.

```suggestion
        "exasol",
        "flink",
```

ibis/backends/tests/test_register.py
```diff
@@ -577,14 +582,22 @@ def num_diamonds(data_dir):
...
+        if con.name in (
+            "pyspark",
+            "sqlite",
+            "mysql",
+            "postgres",
+            "risingwave",
+            "impala",
+            "mssql",
+            "trino",
+        ):
```
cpcloud, line 599: This is a bit weird: for most of the backends in this list the notion of "current working directory" doesn't make sense.

cpcloud (changes requested):

ibis/backends/tests/test_register.py
jitingxu1: Yes, I am not sure what happened, I will double check it.

jitingxu1: It is `pyarrow` having trouble respecting CWD when trying to read the CSV file. I have to provide pyarrow with the full path to the csv file. 

does this make sense to you @cpcloud 

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
ibis/backends/__init__.py:1312: in read_csv
    file_system, path = fs.FileSystem.from_uri(path)
pyarrow/_fs.pyx:477: in pyarrow._fs.FileSystem.from_uri
    ???
pyarrow/error.pxi:155: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   pyarrow.lib.ArrowInvalid: URI has empty scheme: 'diamonds.csv'
```

jitingxu1:
skip the Trino and impala in tests


@cpcloud request to review

ibis/backends/__init__.py
```diff
@@ -1269,6 +1270,108 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+            elif key == "memory_pool":
+                memory_pool = value
+            else:
+                raise ValueError(f"Invalid args: {key!r}")
+
+        read_options = pcsv.ReadOptions(**read_options_args)
+        parse_options = pcsv.ParseOptions(**parse_options_args)
+        convert_options = pcsv.ConvertOptions(**convert_options_args)
+        if memory_pool:
+            memory_pool = pa.default_memory_pool()
```
gforsyth, line 1346: If someone has specified a memory pool, shouldn't we use that memory pool?

ibis/backends/__init__.py
```diff
@@ -1269,6 +1270,108 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+            if hasattr(pcsv.ReadOptions, key):
+                read_options_args[key] = value
+            elif hasattr(pcsv.ParseOptions, key):
+                parse_options_args[key] = value
+            elif hasattr(pcsv.ConvertOptions, key):
+                convert_options_args[key] = value
+            elif key == "memory_pool":
+                memory_pool = value
+            else:
+                raise ValueError(f"Invalid args: {key!r}")
```
gforsyth, line 1340: There should be a link in the docstring, at least, to indicate where users can look for what valid options are.

And if a user passes in their own dictionary of `ReadOptions`, should we accept that?

gforsyth: We might also want to add examples in the docstrings (and tests) of common CSV arguments, like `skip_rows` and `column_names`

ibis/backends/__init__.py
jitingxu1: good catch, Correct it.

ibis/backends/__init__.py
jitingxu1: > And if a user passes in their own dictionary of `ReadOptions`, should we accept that?

Great. Thanks for your comments. I could add support for doing this. 

I have some followups, taking the readOptions as an example, it could be
1. dictionary of ReadOptions , i.e. 
```python
read_options_dict = {'skip_rows': 1, 'column_names': ['animals', 'n_legs', 'entry']
con.read_csv(path/to/csv, read_options=read_options_dict)
```
2. ReadOptions Object,  i.e.
```python
read_options = csv.ReadOptions(autogenerate_column_names=True, skip_rows=1)
con.read_csv(path/to/csv, read_options=read_options)
```
3. attributes of ReadOptions:
```python
con.read_csv(path/to/csv, skip_rows=1, autogenerate_column_names=True)
```
Should we support all these cases?
Do we need to consider if there is a conflicts between 2 and 3, for example:
```python
read_options = csv.ReadOptions(autogenerate_column_names=True, skip_rows=1)
# skip_rows is defined twice
con.read_csv(path/to/csv, read_options=read_options, skip_rows=2)
```

Thanks again for your input.

gforsyth: I like to think about how I would describe the available options in a single sentence.  Once that sentence gets too complicated, we're doing too much.

I think we could support options 2 & 3, and collisions are handled by overriding the contents of e.g. `ReadOptions` with the explicit kwarg.  So in your example above, we would skip the first 2 rows.

gforsyth: Argh, you know, I think I just changed my mind.  If anyone needs hyper-specific setup for reading a csv with pyarrow, they can just read in the CSV with `pyarrow` themselves and then pass in the resulting table.

So ignore the above.

gforsyth: I _do_ think that having to look at 3 different pages in the pyarrow docs for all the options is _very_ annoying, so we can explicitly document common options in the `kwargs` section of the docstring and then point to the various upstream pages for users who want a full list.

ibis/backends/tests/test_register.py
```diff
@@ -594,3 +614,73 @@ def test_read_csv(con, data_dir, in_table_name, num_diamonds):
...
+    if con.name in (
+        "duckdb",
+        "polars",
+        "bigquery",
+        "clickhouse",
+        "datafusion",
+        "snowflake",
+        "pyspark",
+    ):
+        pytest.skip(f"{con.name} implements its own `read_parquet`")
```
jitingxu1, line 657: these backends have their own implementation, some of these options still could pass this test, so I skip these backends.

ibis/backends/tests/test_register.py
```diff
@@ -594,3 +614,73 @@ def test_read_csv(con, data_dir, in_table_name, num_diamonds):
...
+        "duckdb",
+        "polars",
+        "bigquery",
+        "clickhouse",
+        "datafusion",
+        "snowflake",
+        "pyspark",
+    ],
+    reason="backend implements its own read_csv",
+)
```
gforsyth, line 640:
```suggestion
```
You can remove this since you are skipping them inside the test body

ibis/backends/__init__.py
```diff
@@ -1269,6 +1270,117 @@ def has_operation(cls, operation: type[ops.Value]) -> bool:
...
+                    read_options=read_options,
+                    parse_options=parse_options,
+                    convert_options=convert_options,
+                    memory_pool=memory_pool,
+                )
+                pyarrow_tables.append(pyarrow_table)
+
+        pyarrow_table = pa.concat_tables(pyarrow_tables)
+        table_name = table_name or util.gen_name("read_csv")
+        self.create_table(table_name, pyarrow_table)
```
gforsyth, line 1381: Hm, I think this should probably be a temp table or a `memtable`, because none of our other `read_*` functions create a persistent object

gforsyth: `memtable` is probably a good option

cpcloud:
This PR seems overly broad for a first pass.

We probably shouldn't handle anything but local files to start and add support for other types of reads when someone asks for it.

Closing as stale.

PR #9914, opened 2024-08-25
feat(oracle): implement `mode` aggregation

cpcloud:
Add support for the `mode` aggregation to Oracle


jcrist (approved):

ibis/backends/tests/test_generic.py
```diff
@@ -895,31 +883,17 @@ def test_table_info_large(con):
...
-                    ],
+                    ["clickhouse", "exasol", "impala", "pyspark", "risingwave"],
                     raises=com.OperationNotDefinedError,
                     reason="mode is not supported",
                 ),
                 pytest.mark.notimpl(
                     ["oracle"],
-                    raises=com.OperationNotDefinedError,
-                    reason="Mode is not supported and ORA-02000: missing AS keyword",
+                    raises=OracleDatabaseError,
```
cpcloud, line 895: There's still an issue with using mode in `describe`, but at least now users can call `mode` outside of that context. I'm not entirely sure why mode isn't working in `describe`.

PR #9912, opened 2024-08-23
fix(pivot-wider): handle the case of empty `id_cols`

cpcloud:
Fixes an issue with handling simpler cases of `Table.pivot_wider` where you want to take two columns and use one as the header and one as the first and only row.

PR #9917, opened 2024-08-25
refactor(selectors): remove janky `Predicate` class and unify `Selector`s under a single interface

cpcloud:
## Description of changes

This is a refactor of selectors internals.

The disparity between the `Predicate` class and the `Selector` has always
bothered me, because not every selector can be easily made into a predicate.

And, for the ones that can, there are some often-used ones that are really
awkward to express as predicates, like the `c` selector, which expressed as
a predicate led to performance problems and IMO unnecessarily complex solutions
to that performance problem.

I thought to myself, "What are selectors doing fundamentally?"

And the answer I came up with is that they are producing a sequence of column
names that can be used to subset columns in a table.

Set operations ended up being a much more natural way to model the problem, while
allowing subclasses of `Selector` to customize operations for whatever reason
e.g., avoiding any overhead if there's a fast path (selecting no columns) or if
overriding is necessary for the functionality (e.g., `And`).

So, I turned `Predicate` into `Where`, and rewrote a bunch of the selectors
that were previously based on `where` into their own specialized classes, and
boiled everything down to expanding the names of a selector based on an input
table, and rewrote the top-level `expand` method in terms of name expansion.

Previously we were also incurring the overhead of table column selection in
every case, and for a good chunk of selectors, we can avoid that overhead by
computing directly on the table's columns.

### DIFF
--- a/ibis/common/selectors.py
+++ b/ibis/common/selectors.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import abc
+from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
 from ibis.common.grounds import Concrete
+from ibis.common.typing import VarTuple  # noqa: TCH001
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
@@ -14,7 +16,6 @@ if TYPE_CHECKING:
 class Selector(Concrete):
     """A column selector."""
 
-    @abc.abstractmethod
     def expand(self, table: ir.Table) -> Sequence[ir.Value]:
         """Expand `table` into value expressions that match the selector.
 
@@ -29,3 +30,73 @@ class Selector(Concrete):
             A sequence of value expressions that match the selector
 
         """
+        names = self.expand_names(table)
+        return list(map(table.__getitem__, filter(names.__contains__, table.columns)))
+
+    @abc.abstractmethod
+    def expand_names(self, table: ir.Table) -> frozenset[str]:
+        """Compute the set of column names that match the selector."""
+
+    def __and__(self, other: Selector) -> Selector:
+        """Compute the logical conjunction of two `Selector`s.
+
+        Parameters
+        ----------
+        other
+            Another selector
+        """
+        return And(self, other)
+
+    def __or__(self, other: Selector) -> Selector:
+        """Compute the logical disjunction of two `Selector`s.
+
+        Parameters
+        ----------
+        other
+            Another selector
+        """
+        return Or(self, other)
+
+    def __invert__(self) -> Selector:
+        """Compute the logical negation of a `Selector`."""
+        return Not(self)
+
+
+class Or(Selector):
+    left: Selector
+    right: Selector
+
+    def expand_names(self, table: ir.Table) -> frozenset[str]:
+        return self.left.expand_names(table) | self.right.expand_names(table)
+
+
+class And(Selector):
+    left: Selector
+    right: Selector
+
+    def expand_names(self, table: ir.Table) -> frozenset[str]:
+        return self.left.expand_names(table) & self.right.expand_names(table)
+
+
+class Any(Selector):
+    selectors: VarTuple[Selector]
+
+    def expand_names(self, table: ir.Table) -> frozenset[str]:
+        names = (selector.expand_names(table) for selector in self.selectors)
+        return frozenset.union(*names)
+
+
+class All(Selector):
+    selectors: VarTuple[Selector]
+
+    def expand_names(self, table: ir.Table) -> frozenset[str]:
+        names = (selector.expand_names(table) for selector in self.selectors)
+        return frozenset.intersection(*names)
+
+
+class Not(Selector):
+    selector: Selector
+
+    def expand_names(self, table: ir.Table) -> frozenset[str]:
+        names = self.selector.expand_names(table)
+        return frozenset(col for col in table.columns if col not in names)
--- a/ibis/common/selectors.py
+++ b/ibis/common/selectors.py
@@ -13,9 +13,8 @@ if TYPE_CHECKING:
     import ibis.expr.types as ir
 
 
-class Selector(Concrete):
-    """A column selector."""
-
+class Expandable(Concrete):
+    @abc.abstractmethod
     def expand(self, table: ir.Table) -> Sequence[ir.Value]:
         """Expand `table` into value expressions that match the selector.
 
@@ -30,12 +29,18 @@ class Selector(Concrete):
             A sequence of value expressions that match the selector
 
         """
-        names = self.expand_names(table)
-        return list(map(table.__getitem__, filter(names.__contains__, table.columns)))
 
     @abc.abstractmethod
     def expand_names(self, table: ir.Table) -> frozenset[str]:
         """Compute the set of column names that match the selector."""
+
+
+class Selector(Expandable):
+    """A column selector."""
+
+    def expand(self, table: ir.Table) -> Sequence[ir.Value]:
+        names = self.expand_names(table)
+        return list(map(table.__getitem__, filter(names.__contains__, table.columns)))
 
     def __and__(self, other: Selector) -> Selector:
         """Compute the logical conjunction of two `Selector`s.
--- a/ibis/common/selectors.py
+++ b/ibis/common/selectors.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import abc
-from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
 from ibis.common.grounds import Concrete
--- a/ibis/common/selectors.py
+++ b/ibis/common/selectors.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import abc
 from typing import TYPE_CHECKING
 
+from ibis.common.bases import Abstract
 from ibis.common.grounds import Concrete
 from ibis.common.typing import VarTuple  # noqa: TCH001
 
@@ -12,7 +13,9 @@ if TYPE_CHECKING:
     import ibis.expr.types as ir
 
 
-class Expandable(Concrete):
+class Expandable(Abstract):
+    __slots__ = ()
+
     @abc.abstractmethod
     def expand(self, table: ir.Table) -> Sequence[ir.Value]:
         """Expand `table` into value expressions that match the selector.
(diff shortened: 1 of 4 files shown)


cpcloud:
Since we benchmark selectors for the wide table case, I ran them, and there's a consistent 10% improvement.

Nothing to write home about, but I'll take the win.

```
-------------------------------------------------------------------------------- benchmark 'test_selectors[10000]': 2 tests --------------------------------------------------------------------------------
Name (time in ms)                             Min                 Max                Mean            StdDev              Median                IQR            Outliers     OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_selectors[10000] (0001_949fbea)     919.4184 (1.10)     937.7005 (1.11)     928.4652 (1.10)     8.3591 (2.51)     927.5156 (1.10)     15.7539 (2.96)          2;0  1.0770 (0.91)          5           1
test_selectors[10000] (NOW)              837.6641 (1.0)      845.8418 (1.0)      841.4304 (1.0)      3.3308 (1.0)      840.3802 (1.0)       5.3216 (1.0)           2;0  1.1885 (1.0)           5           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------- benchmark 'test_selectors[1000]': 2 tests -------------------------------------------------------------------------------
Name (time in ms)                           Min                 Max               Mean             StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_selectors[1000] (0001_949fbea)     86.8199 (1.13)      87.4022 (1.0)      87.0923 (1.05)      0.1968 (1.0)      87.0289 (1.12)     0.3323 (1.0)           3;0  11.4821 (0.95)          9           1
test_selectors[1000] (NOW)              76.9421 (1.0)      110.7883 (1.27)     82.6303 (1.0)      12.3097 (62.55)    77.4166 (1.0)      1.0610 (3.19)          2;2  12.1021 (1.0)          13           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
========================================================================================= 2 passed in 9.53s ==========================================================================================
```

ibis/selectors.py
```diff
@@ -159,8 +128,17 @@ def numeric() -> Predicate:
     return of_type(dt.Numeric)
 
 
+class OfType(Selector):
+    predicate: Callable[[dt.DataType], bool]
+
+    def expand_names(self, table: ir.Table) -> frozenset[str]:
+        return frozenset(
+            name for name, typ in table.schema().items() if self.predicate(typ)
```
cpcloud, line 136: This case should be a bit faster too, since we're not going through `Table.__getitem__` anymore, we're just looking at the schema, since we only care about types.

ibis/common/selectors.py
```diff
@@ -29,3 +28,79 @@ def expand(self, table: ir.Table) -> Sequence[ir.Value]:
             A sequence of value expressions that match the selector
 
         """
+
+    @abc.abstractmethod
+    def expand_names(self, table: ir.Table) -> frozenset[str]:
```
jcrist, line 33: Using a `frozenset` here feels a bit weird to me. I understand that sets are now technically ordered unique collections, but I guess after years of python usage where that wasn't true relying on the ordering properties of a `frozenset` (as opposed to a `tuple`/`list`) in an interface feels wrong. Like, if I got back a `frozenset` from an API, I wouldn't intuitively think that the ordering of that set was meaningful or mattered. Would you be open to changing this to a `tuple[str]` or `list[str]` instead? The uniqueness properties could still be enforced by the implementations without requiring it in the return type.

ibis/common/selectors.py
```diff
@@ -29,3 +28,79 @@ def expand(self, table: ir.Table) -> Sequence[ir.Value]:
...
+        """Compute the set of column names that match the selector."""
+
+
+class Selector(Expandable):
```
jcrist, line 37: This now looks a lot more like the selectors implementation I wrote for ibis-ml (https://github.com/ibis-project/ibis-ml/blob/main/ibis_ml/select.py) (there's _mild_ reasons why that project needed it's own implementation). Not sure if there's anything worth cribbing from there, but figured I'd raise it.

ibis/common/selectors.py
```diff
@@ -29,3 +28,79 @@ def expand(self, table: ir.Table) -> Sequence[ir.Value]:
...
+    """A column selector."""
+
+    def expand(self, table: ir.Table) -> Sequence[ir.Value]:
+        names = self.expand_names(table)
+        return list(map(table.__getitem__, filter(names.__contains__, table.columns)))
```
jcrist, line 42: General comment - IMO `map`/`filter` are (usually) an antipattern, the list comprehension version of this is both shorter and reads a lot cleaner to me.

```suggestion
        return [table[n] for n in table.columns if n in names]
```

ibis/common/selectors.py
cpcloud, line 33: Gonna see if I can get rid of this abc method on `Expandable`.

ibis/common/selectors.py
```diff
@@ -29,3 +28,79 @@ def expand(self, table: ir.Table) -> Sequence[ir.Value]:
...
+
+        Parameters
+        ----------
+        other
+            Another selector
+        """
+        return And(self, other)
+
+    def __or__(self, other: Selector) -> Selector:
+        """Compute the logical disjunction of two `Selector`s.
```
jcrist, line 55: "logical disjunction" is not a phrase I'd expect most users to be familiar with. If we view selectors as generating a set of column names, "union" (here) and "intersection" (for `__and__`) may be a bit clearer.

ibis/common/selectors.py
jcrist: Thinking further, the generation here of names uses a lot of the `frozenset` apis for ease, perhaps using it as a return type is fine. It's not user facing really anyway. :shrug:.

ibis/selectors.py
```diff
@@ -586,66 +606,113 @@ def if_all(selector: Selector, predicate: Deferred | Callable) -> IfAnyAll:
...
+        return []
+
+    def expand_names(self, table: ir.Table) -> frozenset[str]:
+        return frozenset()
+
+
+@public
+def none() -> Selector:
+    """Return no columns."""
+    return NoColumns()
```
jcrist, line 715: When does this selector come up? Why might a user want to use it?

jcrist (commented):
Overall this looks like a nice improvement!

ibis/common/selectors.py
cpcloud: The `frozenset` here is also used for fast membership testing in a number of places, it's not only about uniqueness.

ibis/common/selectors.py
cpcloud: I don't consider `map` and `filter` an anti-pattern, I've never really understood any of the arguments for avoiding them.

cpcloud: Here, avoiding the overhead of iterating in Python (I realize this might not be huge in most cases) seemed like an acceptable trade-off given I personally didn't consider one to be horribly worse.

Happy to change it though.

jcrist: In this case at least using a comprehension removes the need for writing out two magic method names, which is an improvement. IMO `map`/`filter`/`reduce` do have their place, but in many cases I still prefer the comprehension. Up to you, not going to argue more here.

ibis/common/selectors.py
cpcloud: This was pre-existing. It can be changed in a follow-up. I was originally going for precision, and conceptually clarity.

ibis/selectors.py
cpcloud: In `pivot_wider`, check out #9912.

ibis/common/selectors.py
cpcloud: Fair enough, the magic method names are a little annoying 😅

ibis/selectors.py
jcrist: Still not seeing why it's necessary. In that PR you're using it in a test, but wouldn't `id_cols=[]` work just as well? It's harmless, so happy to add it, but I still see when a user would reach for this.

cpcloud: It's conceptually simpler, for both end users *and* developers to convert `id_cols` to a selector.

What selector would you convert `id_cols=[]` into?

If you're thinking that it'd be special-cased, well, then I guess I disagree with handling it that way, if we have an existing unifying concept that address it and we don't have to remember "oh yeah, and empty sequences are handled in this special way".

cpcloud: I'm not actually handling the case of an empty sequence in `_to_selector`, so thanks for bringing this up!

ibis/common/selectors.py
cpcloud: `__sub__` looks interesting, I'm guessing that's supported in tidyr/dplyr?

ibis/selectors.py
cpcloud: I think there's happy medium here: users who want to use `id_cols=[]` totally can, and the implementation of that as a selector is `none()`.

jcrist: > What selector would you convert id_cols=[] into?

Since everywhere we accept both selectors and lists of columns/expressions they have identical meaning, I'd expect `id_cols=[]` to be the same as `ibis.selectors.c()`. It looks like `_to_selector()` currently fails on `_to_selector([])` (which seems like a bug).

> Developers can work with a single kind of thing, and end users can reason about their input as a selector, or something that is convertible to one.

Right, I'd agree with that. But for the end user `[]` seems like a more natural way to say "no columns". Since it's coercible to a selector, any selector API should combine it just fine IMO and it feels more natural than typing `s.none()`.

cpcloud: Thanks also for bringing up the `c()`-with-no-arguments-case, I'll add a test for that too.

cpcloud: Ok, done!

ibis/common/selectors.py
jcrist: Maybe (can't remember)? Selecting "things that match one thing, excluding things that match another thing" was something I was doing a lot of, so the short-hand was also nice.

cpcloud: Will do in a follow up. I also want to move everything into a single file, it's really annoying having two files for the implementation here.

jcrist (approved):

PR #9918, opened 2024-08-25
fix(oracle): implement current_catalog and current_database correctly

cpcloud:
Fixes some hacks around `current_catalog` and `current_database` for the Oracle backend

### DIFF
--- a/ibis/backends/oracle/__init__.py
+++ b/ibis/backends/oracle/__init__.py
@@ -192,8 +192,16 @@ class Backend(SQLBackend, CanListDatabase, CanListSchema):
         return self
 
     @property
-    def current_database(self) -> str:
+    def current_catalog(self) -> str:
         with self._safe_raw_sql(sg.select(STAR).from_("global_name")) as cur:
+            [(catalog,)] = cur.fetchall()
+        return catalog
+
+    @property
+    def current_database(self) -> str:
+        # databases correspond to users, other than that there's
+        # no notion of a database inside a catalog for oracle
+        with self._safe_raw_sql(sg.select("user").from_("dual")) as cur:
             [(database,)] = cur.fetchall()
         return database
 
--- a/ibis/backends/tests/test_client.py
+++ b/ibis/backends/tests/test_client.py
@@ -1753,24 +1753,13 @@ def test_cross_database_join(con_create_database, monkeypatch):
     ["impala", "pyspark", "trino"], reason="Default constraints are not supported"
 )
 def test_insert_into_table_missing_columns(con, temp_table):
-    try:
-        db = getattr(con, "current_database", None)
-    except NotImplementedError:
-        db = None
+    db = getattr(con, "current_database", None)
 
-    # UGH
-    if con.name == "oracle":
-        db = None
-
-    try:
-        catalog = getattr(con, "current_catalog", None)
-    except NotImplementedError:
-        catalog = None
-
-    raw_ident = ".".join(
-        sg.to_identifier(i, quoted=True).sql("duckdb")
-        for i in filter(None, (catalog, db, temp_table))
-    )
+    raw_ident = sg.table(
+        temp_table,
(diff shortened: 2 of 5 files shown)


ibis/backends/oracle/__init__.py
```diff
@@ -192,8 +192,16 @@ def _from_url(self, url: ParseResult, **kwargs):
...
-    def current_database(self) -> str:
+    def current_catalog(self) -> str:
         with self._safe_raw_sql(sg.select(STAR).from_("global_name")) as cur:
+            [(catalog,)] = cur.fetchall()
+        return catalog
+
+    @property
+    def current_database(self) -> str:
+        # databases correspond to users, other than that there's
+        # no notion of a database inside a catalog for oracle
```
gforsyth, line 203: So given that there's no notion of a database inside a catalog, should we call this `database`?

If a thing is a collection of tables, e.g. `IBIS_TESTING`, then that should be called `database`.

cpcloud: well, you can definitely specify tables like this:

```sql
CREATE TABLE $MY_USER.$MY_TABLE
```

so it seems like it fits the bill in practice, if not nominally.

cpcloud: `IBIS_TESTING` has no single noun attached to it AFAICT.

The documentation for `global_name` uses the word "database" to describe it, while on `self.con` the return value of `self.con.service_name` is `"IBIS_TESTING"`.

gforsyth: I had happily forgotten about `service_name`.

Hmm, so we can _also_ specify things this way:
```python
[ins] In [4]: con.table("ALL_DOMAINS", database="SYS")
Out[4]: 
DatabaseTable: SYS.ALL_DOMAINS
  OWNER            !string
  NAME             !string
  COLS             !int64
  DISCRIMINANTCOLS int64
  BUILTIN          boolean
  DATA_DISPLAY     string
  DISPLAY_LENGTH   int64
  DATA_ORDER       string
  ORDER_LENGTH     int64
  SELECTOR         string
  SELECTOR_LENGTH  int64
  TYPE             string
```

e.g. `SYS.ALL_DOMAINS` is a valid "database" in the Ibis terminology.

Oh wait, and I see that it shows up in `list_databases`, because we haven't changed that.

Ok, if the major change here is that we store the "service_name" (or whatever) in `current_catalog`, and there is no `list_catalogs` implemented (because it wouldn't make sense), and other collections of tables show up in `list_databases`, then I'm good with this.

ibis/backends/tests/test_client.py
```diff
@@ -1753,24 +1753,13 @@ def test_cross_database_join(con_create_database, monkeypatch):
...
 
-    raw_ident = ".".join(
-        sg.to_identifier(i, quoted=True).sql("duckdb")
-        for i in filter(None, (catalog, db, temp_table))
-    )
+    raw_ident = sg.table(
+        temp_table,
+        db=db if db is None else sg.to_identifier(db, quoted=True),
+        quoted=True,
+    ).sql("duckdb")
```
gforsyth, line 1762: this is much less gross

gforsyth (approved):

ibis/backends/oracle/__init__.py
cpcloud: `SYS` [is actually a *user*](https://docs.oracle.com/en/database/oracle/oracle-database/19/admin/getting-started-with-database-administration.html#GUID-79AB6187-1522-4EB7-8FAD-E4322262AC65), so I think it fits what I've done here quite well.

cpcloud: 1. `list_catalogs` doesn't make sense AFAICT
2. `list_databases` *does* makes sense, and a "database" corresponds to a user (I believe this user-as-database pattern shows up in other systems as well).

PR #9921, opened 2024-08-26
perf(dot-columns): speed up heavily used `.columns`

cpcloud:
## Description of changes

This PR changes the `Table.columns` property to return the underlying
`tuple` of names from the operation's `schema` attribute.

Right now, every `.columns` access incurs the overhead of creating a `list`,
which remained for backwards compatibility.

I was refactoring selectors in #9917, and noticed some performance differences
that boiled down to accessing this attribute many times in the case of
selectors operating on very wide tables.

The other approach I considered was t continue returning a `list`, and speed up
the access using `functools.cached_property`.

While this doesn't break the API, it seemed incredibly wasteful to store two
copies of every table's columns, especially in these slower cases where
tables have a ton of columns.

BREAKING CHANGE: `Table.columns` is now a `tuple` instead of a `list`. Your usage of `.columns` may require updating if you're depending specifically on `.columns` being a `list`.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -674,8 +674,9 @@ class Table(Expr, _FixedTextJupyterMixin):
             limit, offset = util.slice_to_limit_offset(what, self.count())
             return self.limit(limit, offset=offset)
 
+        columns = self.columns
         args = [
-            self.columns[arg] if isinstance(arg, int) else arg
+            columns[arg] if isinstance(arg, int) else arg
             for arg in util.promote_list(what)
         ]
         if util.all_of(args, str):
@@ -765,8 +766,8 @@ class Table(Expr, _FixedTextJupyterMixin):
         return self.columns
 
     @property
-    def columns(self) -> list[str]:
-        """The list of column names in this table.
+    def columns(self) -> tuple[str, ...]:
+        """Return a [](`tuple`) of column names in this table.
 
         Examples
         --------
@@ -774,16 +775,16 @@ class Table(Expr, _FixedTextJupyterMixin):
         >>> ibis.options.interactive = True
         >>> t = ibis.examples.penguins.fetch()
         >>> t.columns
-        ['species',
+        ('species',
          'island',
          'bill_length_mm',
          'bill_depth_mm',
          'flipper_length_mm',
          'body_mass_g',
          'sex',
-         'year']
+         'year')
         """
-        return list(self.schema().names)
+        return self._arg.schema.names
 
     def schema(self) -> sch.Schema:
         """Return the [Schema](./schemas.qmd#ibis.expr.schema.Schema) for this table.
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -413,7 +413,7 @@ class Table(Expr, _FixedTextJupyterMixin):
         Columns not present in the input schema will be passed through unchanged
 
         >>> t.columns
-        ['species', 'island', 'bill_length_mm', 'bill_depth_mm', 'flipper_length_mm', 'body_mass_g', 'sex', 'year']
+        ('species', 'island', 'bill_length_mm', 'bill_depth_mm', 'flipper_length_mm', 'body_mass_g', 'sex', 'year')
         >>> expr = t.cast({"body_mass_g": "float64", "bill_length_mm": "int"})
         >>> expr.select(*cols).head()
         ┏━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┓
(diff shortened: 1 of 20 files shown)


cpcloud:
The difference in performance is pretty stark:

```
---------------------------------------------------------------------------------- benchmark 'test_dot_columns[1000000]': 2 tests ---------------------------------------------------------------------------------
Name (time in ms)                                Min                Max               Mean            StdDev             Median               IQR            Outliers             OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_dot_columns[1000000] (0001_d552afa)     10.3734 (>1000.0)  14.1146 (>1000.0)  10.6059 (>1000.0)  0.4976 (>1000.0)  10.4641 (>1000.0)  0.1202 (>1000.0)      5;11         94.2872 (0.00)         90           1
test_dot_columns[1000000] (NOW)               0.0003 (1.0)       0.0090 (1.0)       0.0003 (1.0)      0.0001 (1.0)       0.0003 (1.0)      0.0000 (1.0)      946;2493  2,904,766.3008 (1.0)      185529           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------------- benchmark 'test_dot_columns[100000]': 2 tests ------------------------------------------------------------------------------------
Name (time in us)                                Min                 Max                Mean             StdDev              Median               IQR             Outliers  OPS (Kops/s)            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_dot_columns[100000] (0001_d552afa)     197.4140 (853.20)   278.9990 (174.87)   207.2664 (849.73)   13.9447 (>1000.0)  201.0715 (829.57)   7.7540 (>1000.0)    186;200        4.8247 (0.00)       1642           1
test_dot_columns[100000] (NOW)                0.2314 (1.0)        1.5954 (1.0)        0.2439 (1.0)       0.0137 (1.0)        0.2424 (1.0)      0.0057 (1.0)      2999;3160    4,099.6842 (1.0)      193051          21
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

--------------------------------------------------------------------------------- benchmark 'test_dot_columns[10000]': 2 tests ---------------------------------------------------------------------------------
Name (time in us)                              Min                Max               Mean            StdDev             Median               IQR             Outliers  OPS (Kops/s)            Rounds  Iterations
test_dot_columns[10000] (0001_d552afa)     19.5370 (61.05)    48.8320 (6.99)     21.2852 (61.88)    1.5586 (25.24)    21.0490 (61.73)    0.4710 (469.09)   2090;3084       46.9810 (0.02)      35120           1
test_dot_columns[10000] (NOW)               0.3200 (1.0)       6.9830 (1.0)       0.3440 (1.0)      0.0617 (1.0)       0.3410 (1.0)      0.0010 (1.0)      601;33246    2,907.2263 (1.0)      106519           1

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
========================================================================================= 3 passed in 9.38s ==========================================================================================
```

cpcloud:
Sure, we're in microsecond territory, but this property is used quite a bit, and hit pretty heavily in selector processing.

cpcloud:
Another approach is to replace the use of `.columns` in internals, (perhaps with a private `_columns` property), which is annoying but also not a breaking change.

The downside there is that with `.columns` and `._columns`, developers can easily choose the wrong one and get bad performance.

ibis/expr/types/relations.py
```diff
@@ -880,25 +880,25 @@ def _ipython_key_completions_(self) -> list[str]:
         return self.columns
 
     @property
-    def columns(self) -> list[str]:
-        """The list of column names in this table.
+    def columns(self) -> tuple[str, ...]:
```
jcrist, line 883: I guess this is _technically_ a breaking change, though I'd be surprised if this meaningfully affected downstream usage. :shrug: if wrapping it in a `list` (should be negligible in perf cost for realistic ibis usage) would be better to avoid all the ripple-effect changes. Would certainly make this a shorter PR.

jcrist: My mistake - I thought the cost here was coming from abstractions in going through `self.schema().names`, not the list conversion itself. Nevermind.

jcrist (approved):

cpcloud:
The 3 cloud tests that were failing are now passing.

```
ibis/backends/snowflake/tests/test_client.py::test_insert PASSED         [ 33%]
ibis/backends/bigquery/tests/system/test_client.py::test_parted_column[date] PASSED [ 66%]
ibis/backends/bigquery/tests/system/test_client.py::test_parted_column[timestamp] PASSED [100%]

========================================================================================= 3 passed in 39.24s =========================================================================================
```

PR #9922, opened 2024-08-26
fix(polars): use `drop_table` when cleaning the cache and remove duplicated `_remove_table` method

cpcloud:
Fixes an issue in the polars backend where tables were still visible in SQL queries after being dropped. Ripped out of #9903.

### DIFF
--- a/ibis/backends/tests/errors.py
+++ b/ibis/backends/tests/errors.py
@@ -65,10 +65,11 @@ try:
     from polars.exceptions import InvalidOperationError as PolarsInvalidOperationError
     from polars.exceptions import PanicException as PolarsPanicException
     from polars.exceptions import SchemaError as PolarsSchemaError
+    from polars.exceptions import SQLInterfaceError as PolarsSQLInterfaceError
 except ImportError:
     PolarsComputeError = PolarsPanicException = PolarsInvalidOperationError = (
         PolarsSchemaError
-    ) = PolarsColumnNotFoundError = None
+    ) = PolarsColumnNotFoundError = PolarsSQLInterfaceError = None
 
 try:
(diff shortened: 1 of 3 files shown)


ibis/backends/tests/errors.py
```diff
@@ -65,10 +65,11 @@
     from polars.exceptions import InvalidOperationError as PolarsInvalidOperationError
     from polars.exceptions import PanicException as PolarsPanicException
     from polars.exceptions import SchemaError as PolarsSchemaError
+    from polars.exceptions import SQLInterfaceError as PolarsSQLInterfaceError
 except ImportError:
     PolarsComputeError = PolarsPanicException = PolarsInvalidOperationError = (
         PolarsSchemaError
-    ) = PolarsColumnNotFoundError = None
+    ) = PolarsColumnNotFoundError = PolarsSQLInterfaceError = None
```
jcrist, line 72: This is some super surprising formatting.

jcrist (approved):
:shipit:

ibis/backends/tests/errors.py
cpcloud: Yeah, chained assignment with a bunch of aliased long-named classes is 🤮

cpcloud: OTOH, it's not like this code sees anything other than rote additions and deletions, so optimizing it for aesthetics feels a bit pointless.

jcrist: Yeah, this wasn't a "do something about this comment", more of a "huh, `ruff` formatted that odd" comment.

cpcloud: For sure! I'm just having the usual Monday-morning strong opinion :)

PR #9923, opened 2024-08-26
feat(sql): fuse `distinct` with other select nodes when possible

### LINKED ISSUE #9905: feat: simplify generated select distinct expressions

Originally discussed [here](https://ibis-project.zulipchat.com/#narrow/stream/405265-tech-support/topic/distinct.28.29.20generates.20unnecessary.20subquery/near/463475939).

When compiling expressions that use `.distinct()` to SQL,  it seems like `ibis` always generates a subquery selection.

E.g., consider the following ibis program.

```python
import ibis

# mock tables
table_a = ibis.table(
    [
        ('id', 'int64'),
        ('event_id', 'int64'),
        ('event_code', 'string'),
        ('event_date', 'date'),
        ('event_timestamp', 'timestamp'),
        ('event_type', 'int64'),
    ],
    name='table_a'
)

table_b = ibis.table(
    [
        ('id', 'int64'),
        ('location_id', 'int64'),
        ('external_id', 'string'),
    ],
    name='table_b'
)

# Create the expression
expr  = (
    table_a.left_join(table_b, table_a.id == table_b.id)
    .filter(table_a.event_type == 12345)
    .select(
        [
            table_a.id,
            table_b.location_id,
            table_b.external_id,
            table_a.event_id,
            table_a.event_code,
            table_a.event_date,
            table_a.event_timestamp,
        ]
    )
    .distinct()
)

ibis.to_sql(expr)
```

This will produce something like:
```sql
SELECT DISTINCT
  *
FROM (
  SELECT
    "t4"."id",
    "t4"."location_id",
    "t4"."external_id",
    "t4"."event_id",
    "t4"."event_code",
    "t4"."event_date",
    "t4"."event_timestamp"
  FROM (
    SELECT
      "t2"."id",
      "t2"."event_id",
      "t2"."event_code",
      "t2"."event_date",
      "t2"."event_timestamp",
      "t2"."event_type",
      "t3"."id" AS "id_right",
      "t3"."location_id",
      "t3"."external_id"
    FROM "table_a" AS "t2"
    LEFT OUTER JOIN "table_b" AS "t3"
      ON "t2"."id" = "t3"."id"
  ) AS "t4"
  WHERE
    "t4"."event_type" = 12345
) AS "t5"
```

but ideally we'd get something like:
```sql
SELECT DISTINCT
  "t4"."id",
  "t4"."location_id",
  "t4"."external_id",
  "t4"."event_id",
  "t4"."event_code",
  "t4"."event_date",
  "t4"."event_timestamp"
FROM (
  SELECT
    "t2"."id",
    "t2"."event_id",
    "t2"."event_code",
    "t2"."event_date",
    "t2"."event_timestamp",
    "t2"."event_type",
    "t3"."id" AS "id_right",
    "t3"."location_id",
    "t3"."external_id"
  FROM "table_a" AS "t2"
  LEFT OUTER JOIN "table_b" AS "t3"
    ON "t2"."id" = "t3"."id"
) AS "t4"
WHERE
  "t4"."event_type" = 12345
AS "t5"
```

Analysts say the SQL generated by ibis is sometimes difficult to read, which can make it difficult to debug ibis logic and, importantly, also makes it difficult to communicate about what ibis is doing to audiences that primarily understand SQL.

This example with `distinct()` is one that came up frequently in my discussions. So, the goal of this feature is primarily to improve the legibility of SQL generated by ibis to humans.  It's also possible that it could also improve performance for some backends, though I imagine most would optimize this away when generating their internal execution plan.

per @cpcloud [here](https://ibis-project.zulipchat.com/#narrow/stream/405265-tech-support/topic/distinct.28.29.20generates.20unnecessary.20subquery/near/463498002):
> This would probably have to be an optimization in our expression rewriting system.

Alternatively, if simplifying distinct expressions generically is too complex, allowing a `.select()` expression to take a `distinct=True` argument and generating directly a `SELECT DISTINCT` would also be an acceptable solution.

9.3.0

this problem is backend-independent

### Code of Conduct

jcrist:
This generates more concise SQL when chaining relational operations with `.distinct()`.

Fixes #9905.

### DIFF
--- a/ibis/backends/sql/rewrites.py
+++ b/ibis/backends/sql/rewrites.py
@@ -53,6 +53,7 @@ class Select(ops.Relation):
     predicates: VarTuple[ops.Value[dt.Boolean]] = ()
     qualified: VarTuple[ops.Value[dt.Boolean]] = ()
     sort_keys: VarTuple[ops.SortKey] = ()
+    distinct: bool = False
 
     def is_star_selection(self):
         return tuple(self.values.items()) == tuple(self.parent.fields.items())
@@ -126,6 +127,12 @@ def filter_to_select(_, **kwargs):
 def sort_to_select(_, **kwargs):
     """Convert a Sort node to a Select node."""
     return Select(_.parent, selections=_.values, sort_keys=_.keys)
+
+
+@replace(p.Distinct)
+def distinct_to_select(_, **kwargs):
+    """Convert a Distinct node to a Select node."""
+    return Select(_.parent, selections=_.values, distinct=True)
 
 
 @replace(p.DropColumns)
@@ -244,6 +251,48 @@ def merge_select_select(_, **kwargs):
     if _.parent.find_below(blocking, filter=ops.Value):
         return _
 
+    if _.parent.distinct:
+        # The inner query is distinct.
+        #
+        # If the outer query is distinct, it's only safe to merge if it's a simple subselection:
+        # - Fusing in the presence of non-deterministic calls in the select would lead to
+        #   incorrect results
+        # - Fusing in the presence of expensive calls in the select would lead to potential
+        #   performance pitfalls
+        if _.distinct and not all(
+            isinstance(v, ops.Field) for v in _.selections.values()
+        ):
+            return _
+
+        # If the outer query isn't distinct, it's only safe to merge if the outer is a SELECT *:
+        # - If new columns are added, they might be non-distinct, changing the distinctness
+        # - If previous columns are removed, that would also change the distinctness
+        if not _.distinct and not _.is_star_selection():
+            return _
+
+        distinct = True
+    elif _.distinct:
+        # The outer query is distinct and the inner isn't. It's only safe to merge if either
+        # - The inner query isn't ordered
+        # - The outer query is a SELECT *
+        #
+        # Otherwise we run the risk that the outer query drops columns needed for the ordering of
+        # the inner query - many backends don't allow select distinc queries to order by columns
+        # that aren't present in their selection, like
+        #
+        #   SELECT DISTINCT a, b FROM t ORDER BY c  --- some backends will explode at this
+        #
+        # An alternate solution would be to drop the inner ORDER BY clause, since the backend will
+        # ignore it anyway since it's a subquery. That feels potentially risky though, better
+        # to generate the SQL as written.
+        if _.parent.sort_keys and not _.is_star_selection():
+            return _
+
+        distinct = True
+    else:
+        # Neither query is distinct, safe to merge
+        distinct = False
+
     subs = {ops.Field(_.parent, k): v for k, v in _.parent.values.items()}
     selections = {k: v.replace(subs, filter=ops.Value) for k, v in _.selections.items()}
 
@@ -266,6 +315,7 @@ def merge_select_select(_, **kwargs):
         predicates=unique_predicates,
         qualified=unique_qualified,
         sort_keys=unique_sort_keys,
+        distinct=distinct,
     )
     return result if complexity(result) <= complexity(_) else _
 
@@ -289,6 +339,7 @@ def sqlize(
     node: ops.Node,
     params: Mapping[ops.ScalarParameter, Any],
     rewrites: Sequence[Pattern] = (),
+    post_rewrites: Sequence[Pattern] = (),
     fuse_selects: bool = True,
 ) -> tuple[ops.Node, list[ops.Node]]:
     """Lower the ibis expression graph to a SQL-like relational algebra.
@@ -300,7 +351,9 @@ def sqlize(
     params
         A mapping of scalar parameters to their values.
     rewrites
-        Supplementary rewrites to apply to the expression graph.
+        Supplementary rewrites to apply before SQL-specific transforms.
+    post_rewrites
+        Supplementary rewrites to apply after SQL-specific transforms.
     fuse_selects
         Whether to merge subsequent Select nodes into one where possible.
 
@@ -322,6 +375,7 @@ def sqlize(
         | project_to_select
         | filter_to_select
         | sort_to_select
+        | distinct_to_select
         | fill_null_to_select
         | drop_null_to_select
         | drop_columns_to_select
@@ -334,6 +388,9 @@ def sqlize(
         simplified = sqlized.replace(merge_select_select)
     else:
         simplified = sqlized
+
+    if post_rewrites:
+        simplified = simplified.replace(reduce(operator.or_, post_rewrites))
 
     # extract common table expressions while wrapping them in a CTE node
     ctes = extract_ctes(simplified)
@@ -349,6 +406,46 @@ def sqlize(
 
 
 # supplemental rewrites selectively used on a per-backend basis
+
+
+@replace(Select)
+def split_select_distinct_with_order_by(_):
+    """Split a `SELECT DISTINCT ... ORDER BY` query when needed.
+
+    Some databases (postgres, pyspark, ...) have issues with two types of
+    ordered select distinct statements:
+
+    ```
+    --- ORDER BY with an expression instead of a name in the select list
+    SELECT DISTINCT a, b FROM t ORDER BY a + 1
+
+    --- ORDER BY using a qualified column name, rather than the alias in the select list
+    SELECT DISTINCT a, b as x FROM t ORDER BY b  --- or t.b
+    ```
+
+    We solve both these cases by splitting everything except the `ORDER BY`
+    into a subquery.
+
+    ```
+    SELECT DISTINCT a, b FROM t WHERE a > 10 ORDER BY a + 1
+    --- is rewritten as ->
+    SELECT * FROM (SELECT DISTINCT a, b FROM t WHERE a > 10) ORDER BY a + 1
+    ```
+    """
+    # risingwave and pyspark also don't allow qualified names as sort keys, like
+    #   SELECT DISTINCT t.a FROM t ORDER BY t.a
+    # To avoid having specific rewrite rules for these backends to use only
+    # local names, we always split SELECT DISTINCT from ORDER BY here. Otherwise we
+    # could also avoid splitting if all sort keys appear in the select list.
+    if _.distinct and _.sort_keys:
+        inner = _.copy(sort_keys=())
+        subs = {v: ops.Field(inner, k) for k, v in inner.values.items()}
+        sort_keys = tuple(s.replace(subs, filter=ops.Value) for s in _.sort_keys)
+        selections = {
+            k: v.replace(subs, filter=ops.Value) for k, v in _.selections.items()
+        }
+        return Select(inner, selections=selections, sort_keys=sort_keys)
+    return _
 
 
 @replace(p.WindowFunction(func=p.NTile(y), order_by=()))
--- a/ibis/backends/tests/sql/test_select_sql.py
+++ b/ibis/backends/tests/sql/test_select_sql.py
@@ -254,6 +254,67 @@ def test_projection_filter_fuse(projection_fuse_filter, snapshot):
     snapshot.assert_match(to_sql(expr3), "out.sql")
 
 
+@pytest.mark.parametrize(
+    "transform",
+    [
+        # Fused
+        param(
+            lambda t: t.distinct(),
+            id="distinct",
+        ),
+        param(
+            lambda t: t.select("a", "b").distinct(),
+            id="select-distinct",
+        ),
+        param(
+            lambda t: t.distinct().select("a", "b").distinct(),
+            id="distinct-select-distinct",
+        ),
+        param(
+            lambda t: t.distinct().filter(_.a > 10),
+            id="distinct-filter",
+        ),
+        param(
+            lambda t: t.distinct().filter(_.a > 10).order_by("a"),
+            id="distinct-filter-order_by",
+        ),
+        param(
+            lambda t: t.order_by("a").distinct(),
+            id="order_by-distinct",
+        ),
+        param(
+            lambda t: t.select("a", d=(_.b % 2)).distinct(),
+            id="non-trivial-select-distinct",
+        ),
+        # Not Fused
+        param(
+            lambda t: t.distinct().select("a", "b"),
+            id="distinct-select",
+        ),
+        param(
+            lambda t: t.distinct().select("a", d=(_.b % 2)),
+            id="distinct-non-trivial-select",
+        ),
+        param(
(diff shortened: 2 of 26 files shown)


ibis/backends/sql/rewrites.py
```diff
@@ -244,6 +255,22 @@ def merge_select_select(_, **kwargs):
     if _.parent.find_below(blocking, filter=ops.Value):
         return _
 
+    if not _.distinct and _.parent.distinct:
+        # SELECT ... FROM (SELECT DISTINCT ...)
+        if all(isinstance(v, ops.Field) for v in _.selections.values()):
```
cpcloud, line 260: This seems like it might be expensive in cases with lots of outer selection columns, but I don't have a great solution.

ibis/backends/sql/rewrites.py
```diff
@@ -244,6 +255,22 @@ def merge_select_select(_, **kwargs):
...
+            # Can fuse a trivial subselection after a distinct, like
+            #   SELECT a, b FROM (SELECT DISTINCT ...) -> SELECT DISTINCT a, b, ...
+            distinct = True
+        else:
+            # Other cases aren't safe to fuse as they might introduce non-distinct
+            # columns where applying a distinct would change the result.
+            return _
```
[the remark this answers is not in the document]
cpcloud: Seems totally fine. The composition of `distinct` seems tricky.

ibis/backends/sql/rewrites.py
cpcloud: It seems like detecting a reprojection is potentially very useful and we do have `is_star_selection()` but it's not going to be cheaper than what you've done here.

ibis/backends/sql/rewrites.py
```diff
@@ -244,6 +255,22 @@ def merge_select_select(_, **kwargs):
...
+    else:
+        # SELECT DISTINCT ... FROM (SELECT DISTINCT ...) -> SELECT DISTINCT ...
```
cpcloud, line 269: What about this?

```sql
SELECT DISTINCT RANDOM_FUNCTION(a)
FROM (
  SELECT DISTINCT a -- assume this subquery returns one row
  FROM t
)
```

I don't think that's equivalent to

```sql
SELECT DISTINCT RANDOM_FUNCTION(a)
FROM t
```

ibis/backends/sql/rewrites.py
```diff
@@ -244,6 +255,22 @@ def merge_select_select(_, **kwargs):
...
+        # SELECT DISTINCT ... FROM (SELECT ...) -> SELECT DISTINCT ...
```
cpcloud, line 270: I think this is only valid if the `distinct`-ed columns are guaranteed to be a subset-of-or-equal-to the parent columns. Pretty sure we don't have a way to construct anything else, but just wanted to call it out here.

ibis/backends/sql/rewrites.py
```diff
@@ -244,6 +255,22 @@ def merge_select_select(_, **kwargs):
...
+        # SELECT ... FROM (SELECT ...) -> SELECT ...
```
cpcloud, line 271: Might be just me, but I think here you're saying both `distinct`s are `False` and so the merging behavior is unchanged. Is that right?

jcrist: Yes. Just enumerating the 4 possible combinations of outer-query/inner-query `distinct` value.

ibis/backends/sql/rewrites.py
jcrist: Hmmm, you're correct. We do balk already (above) if either the inner or outer query do contain impure functions like `random`. If you're ok with the logic remaining the same, I'll add a test for this case to ensure it doesn't change.

ibis/backends/sql/rewrites.py
jcrist: Are you saying this is only valid if the selection in the outer column is trivial (e.g. only column names)? If so, I don't see why that limitation is required, the following queries should be identical:

```sql
SELECT DISTINCT a, c % 2 AS d FROM (SELECT a, b + 1 AS c, FROM test);
SELECT DISTINCT a, (b + 1) % 2 AS d FROM test;
```

If I'm understanding correctly that you disagree, do you have a failing example?

ibis/backends/sql/rewrites.py
jcrist: Hmmm, given the above example though, I wonder if we don't want to rewrite it anyway. Generalizing this example to

```sql
SELECT DISTINCT SOME_FUNCTION(a)
FROM (
  SELECT DISTINCT a -- assume a has duplicates so this distinct reduces the set by a noticeable factor
  FROM t
)
```

We'd currently rewrite this to

```sql
SELECT DISTINCT SOME_FUNCTION(a) FROM t
```

If `SOME_FUNCTION` is deterministic, rewriting the former as the latter is correct.

However, if `SOME_FUNCTION` is expensive, doing the distinct in a subquery may be a major performance boost over the fused version (and rewriting the latter as the former isn't something I'd expect a backend's optimizer to do?).

Perhaps we never want to fuse chained distincts unless it's a trivial subselection, since fusing would prevent the user from doing such an optimization?

cpcloud: > Perhaps we never want to fuse chained distincts unless it's a trivial subselection, since fusing would prevent the user from doing such an optimization?

This seems like a good first pass approach to me.

ibis/backends/sql/rewrites.py
cpcloud: Yeah, it looks like the notation is saying SELECT of SELECT becomes a single SELECT.

How about this?

```suggestion
        # SELECT ... FROM (SELECT ...) -> unchanged
```

ibis/backends/sql/rewrites.py
cpcloud: I'm asking whether we even have a way to construct `distinct` operations that are not always effectively `DISTINCT *`.

ibis/backends/sql/rewrites.py
jcrist: I just redid the comments completely, hopefully that's clearer.

ibis/backends/sql/rewrites.py
jcrist: Ah, yeah. There's no way to write those directly, but as select nodes are joined in `merge_select_select` they can come up. I believe I've covered all the possibilities with tests now.

jcrist:
Hmmm, this has turned up some SQL analysis bugs (err, I think they're bugs) in spark & risingwave. In both cases the presence of a column alias inside a `SELECT DISTINCT` is causing the original column name to not be in scope in the `ORDER BY` part of the query (it is still in scope in other parts like `WHERE`).

```sql
--- Order by the alias works
SELECT DISTINCT x AS z FROM test ORDER BY z
--- Order by the original name fails, saying no column named `z` found
SELECT DISTINCT x AS z FROM test ORDER BY x
--- Order by the original name works if you remove the DISTINCT
SELECT x AS z FROM test ORDER BY x
```

- Is this a bug in our SQL generated that most backends just happen to support, or is this a bug in pyspark/risingwave?
- Is using the alias instead of the original name everywhere a valid fix?

jcrist:
Ok, I have a fix for this (I think), and now know more dumb things about SQL dialects than I did before.

ibis/backends/sql/rewrites.py
```diff
@@ -351,6 +408,46 @@ def wrap(node, _, **kwargs):
 # supplemental rewrites selectively used on a per-backend basis
 
 
+@replace(Select)
+def split_select_distinct_with_order_by(_):
```
jcrist, line 412: There are dragons here.

ibis/backends/sql/rewrites.py
```diff
@@ -244,6 +251,48 @@ def merge_select_select(_, **kwargs):
     if _.parent.find_below(blocking, filter=ops.Value):
         return _
 
+    if _.parent.distinct:
```
jcrist, line 254: There are also dragons here.

ibis/backends/tests/test_generic.py
```diff
@@ -1436,6 +1437,53 @@ def test_pivot_wider(backend):
...
+
+@pytest.mark.parametrize(
+    "ops",
+    [
+        param(ops, id="-".join(ops))
+        for ops in permutations(("select", "distinct", "filter", "order_by"))
+        if ops.index("select") < ops.index("distinct")
+    ],
+)
+def test_select_distinct_filter_order_by_commute(backend, alltypes, df, ops):
```
jcrist, line 1466: When fusing is enabled (which it is by default), some of our operations are commutative (or at least should be).

- `select`/`filter`/`order_by` always commute (as well as their variants like `drop`).
- `distinct` also commutes with these (sans `select` in most cases), provided certain restrictions are met. A little more nuanced.

This test checks that behavior across our backends so users can rely on it a little more. This turned up bugs.

ibis/backends/tests/test_generic.py
```diff
@@ -1436,6 +1437,53 @@ def test_pivot_wider(backend):
...
+    sol = (
+        df.drop_duplicates()
+        .rename(columns={"x": "y", "y": "x"})
+        .sort_values(["x", "y"])
+    )
+    res = con.to_pandas(expr)
+    backend.assert_frame_equal(res, sol)
+
+
+def test_select_distinct_order_by_expr(backend, alltypes, df):
```
jcrist, line 1452: This test and the one above it check that we work around any restrictions in handling ordered select distinct statements that might be imposed by the backend.

ibis/backends/tests/sql/test_select_sql.py
```diff
@@ -254,6 +254,67 @@ def test_projection_filter_fuse(projection_fuse_filter, snapshot):
...
+            lambda t: t.order_by("a").drop("a").distinct(),
+            id="order_by-drop-distinct",
+        ),
+        param(
+            lambda t: t.order_by("a").distinct().drop("a"),
+            id="order_by-distinct-drop",
+        ),
+    ],
+)
+def test_fuse_distinct(snapshot, transform):
```
cpcloud, line 312: Excellent job on coverage here.

cpcloud:
Snowflake is passing, but BigQuery is raising this exception:

```
E               google.api_core.exceptions.BadRequest: 400 ORDER BY clause expression references t0.id which is not visible after SELECT DISTINCT at [11:7]; reason: invalidQuery, location: query, message: ORDER BY clause expression references t0.id which is not visible after SELECT DISTINCT at [11:7]
E
E               Location: US
E               Job ID: d7ab4e0d-b30f-4afc-8ed7-286b71a895a5
```

@jcrist I'm guessing BigQuery needs the new rewrite rule applied?

jcrist:
Huh, thought I'd run the cloud tests in this PR already, will fix up.

jcrist:
Cloud tests are passing, should be good-to-go.

ibis/backends/sql/rewrites.py
```diff
@@ -244,6 +251,48 @@ def merge_select_select(_, **kwargs):
...
+        # The inner query is distinct.
+        #
+        # If the outer query is distinct, it's only safe to merge if it's a simple subselection:
+        # - Fusing in the presence of non-deterministic calls in the select would lead to
+        #   incorrect results
+        # - Fusing in the presence of expensive calls in the select would lead to potential
+        #   performance pitfalls
+        if _.distinct and not all(
+            isinstance(v, ops.Field) for v in _.selections.values()
```
cpcloud, line 263: Does this cover the alias-in-outer-project case?

```sql
SELECT a, b AS c
FROM (
  SELECT DISTINCT
    a, b
  FROM t
)
```

would become

```sql
SELECT DISTINCT
  a, b AS c
FROM t
```

If not, fine to either handle later or perhaps never if it doesn't come up.

ibis/backends/sql/rewrites.py
```diff
@@ -244,6 +251,48 @@ def merge_select_select(_, **kwargs):
...
+        # Otherwise we run the risk that the outer query drops columns needed for the ordering of
+        # the inner query - many backends don't allow select distinc queries to order by columns
+        # that aren't present in their selection, like
+        #
+        #   SELECT DISTINCT a, b FROM t ORDER BY c  --- some backends will explode at this
+        #
+        # An alternate solution would be to drop the inner ORDER BY clause, since the backend will
+        # ignore it anyway since it's a subquery. That feels potentially risky though, better
+        # to generate the SQL as written.
+        if _.parent.sort_keys and not _.is_star_selection():
```
cpcloud, line 288: We should probably turn `is_star_selection` into a cached attribute to avoid looping through all columns so many times.

I can do that in a follow up.

ibis/backends/sql/rewrites.py
cpcloud: Missed opportunity to call this `supplant_merged_with_ordered_distinct` 😂

ibis/backends/tests/test_generic.py
cpcloud: Do we need some issues, or are these fixed in this PR?

cpcloud (approved):
LGTM, thanks for chugging through the muck!

ibis/backends/sql/rewrites.py
jcrist: This branch is for both being distinct, so:

```
SELECT DISTINCT a, b as c
FROM (SELECT DISTINCT a, b FROM t)
```

In this case, yes, the aliases are properly handled.

The branch on line 270 handles the `SELECT ... FROM (SELECT DISTINCT ...)` case. Right now it only works with `SELECT *` cases, but we might be able to make it work with outer queries that rename columns but otherwise select all of them. Right now I don't think that's worth it.

ibis/backends/tests/test_generic.py
jcrist: See https://github.com/ibis-project/ibis/issues/10005

ibis/backends/sql/rewrites.py
jcrist: Good idea, I'll push something quick up now.

jcrist: Actually, nvm. Currently we can't cache things with a `cached_property` (since we use `__slots__`). And using `attribute means this is precomputed on `__init__`, which we don't want to do since it won't always be used. Any `Select` node that survives through the IR -> rewrite -> compile chain will only have this field accessed at most twice (in two different spots). I think leaving it as a method for now is best, we can cleanup `attribute` handling later if needed.

Issue #6195, opened 2023-05-11
feat: make `Table.cache()` a no-op for tables that are already concrete in a backend


jcrist:
Currently `Table.cache()` will result in a new copy of the data being stored in the backend, even if the data is already a "concrete" table in the backend.

Ideally if a table is already concrete (backed by a physical table, not a view, in the corresponding backend) then `table.cache()` would be a no-op. This would better enable writing generic functions that make use of `.cache` without unnecessarily duplicating data.

```python
t = con.table("my_table")  # a concrete table

_ = t.cache()  # a no-op

_ = t.mutate(foo=t.bar + 1).cache()  # not a no-op, since `t` isn't a physical table

t = con.table("some-view")  # a view

_ = t.cache()  # not a no-op, since `t` is a view
```

cpcloud:
@jcrist I can't remember, are we not already doing this?

jcrist:
We are not. `t.cache(); t.cache()` won't cache a table twice, but it's not a no-op if a table is already a physical table.

jcrist:
This is easy to do if we assume that `PhysicalTable` ops don't need to be cached (we might also expand this to simple column subselection on these tables like `t.select("a", "b", "c")` which should be cheap), while all other expressions get cached.

However, this assumption isn't true, since some backends (e.g. duckdb) return views for `read_csv`/`read_parquet`, which won't be as cheap to access for repeat queries as a temporary table produced by `.cache()`. Since these views are also mapped to `DatabaseTable`, it's not easy to determine whether a `PhysicalTable` is really "physical" or not. In #9931 I note that I (and some user's code like @lostmygithubaccount I've seen) have sometimes used `read_csv(...).cache()` to create a temp table in duckdb since by default `read_csv(...)` just creates a view. It'd be nice to keep that workflow working while reducing overhead for cases where it's not needed.

A few options:
- Add a new IR node for views (possibly a subclass of `DatabaseTable`) and use those instead for those apis. This would make determining whether an op is worth caching easier.
- Query inside `_load_into_cache` more info about a backing table to determine if it's cheap or not, and if it is take a fast path to avoid the temp-table creation.

Neither seems terribly onerous (mostly just plumbing), but the latter is much less invasive.

jcrist:
I'm going to try implementing the latter (probably for a small subset of backends to start) and see how it looks. I think this should be doable without being too invasive.

PR #9938, opened 2024-08-27
refactor: simplify cache, add support for `mssql`

jcrist:
This:

- Simplifies the implementation of `.cache()`. Abstracting things out to a general `RefCountedCache` led to some hard-to-follow code paths for no added benefit (we weren't using that class anywhere else, and there were no isolated tests for the functionality). Moving to a mixin in the `BaseBackend` made the methods required a lot easier to follow.
- Given the simpler code path, I was able to find a bug in the `mssql` implementation that was preventing support. `.cache()` now fully works with `mssql`.

Net decrease in LOC, while adding functionality and tests.

This change was motivated by #6195, which I'll add as a follow-up.

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -839,7 +839,7 @@ class CacheHandler:
                     raise
 
     def _create_cached_table(self, name: str, expr: ir.Table) -> ir.Table:
-        return self.create_table(name, expr, temp=True)
+        return self.create_table(name, expr, schema=expr.schema(), temp=True)
 
     def _drop_cached_table(self, name: str) -> None:
         self.drop_table(name, force=True)
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -6,9 +6,11 @@ import functools
 import importlib.metadata
 import keyword
 import re
+import sys
 import urllib.parse
+import weakref
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar, NamedTuple
 
 import ibis
 import ibis.common.exceptions as exc
@@ -16,7 +18,6 @@ import ibis.config
 import ibis.expr.operations as ops
 import ibis.expr.types as ir
 from ibis import util
-from ibis.common.caching import RefCountedCache
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Mapping, MutableMapping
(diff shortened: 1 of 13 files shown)


ibis/backends/tests/test_expr_caching.py
```diff
@@ -11,31 +11,24 @@
...
-    reason="mssql supports support temporary tables through naming conventions",
-)
 @mark.notimpl(["exasol"], reason="Exasol does not support temporary tables")
 @pytest.mark.never(
     ["risingwave"],
     raises=com.UnsupportedOperationError,
     reason="Feature is not yet implemented: CREATE TEMPORARY TABLE",
 )
-def test_persist_expression(backend, alltypes):
+def test_persist_expression_foo(backend, alltypes):
```
cpcloud, line 20: Can we remove the `foo` here? 🤣

cpcloud, line 20:
```suggestion
def test_persist_expression(backend, alltypes):
```

ibis/backends/__init__.py
```diff
@@ -777,7 +778,76 @@ def drop_schema(
...
+        """Release a cached table given its name.
+
+        This is a no-op if the cached table is already released.
+
+        Parameters
+        ----------
+        name
+            The name of the cached table.
+        """
+        if (entry := self._cache_name_to_entry.pop(name, None)) is not None:
```
cpcloud, line 833: Isn't `name` equivalent to `entry.name`? If so, can't we get rid of storing `name` as a field on `_CacheEntry`?

cpcloud (commented):
Nice!

Couple of bits of questions/feedback on the implementation.

ibis/backends/__init__.py
```diff
@@ -777,7 +778,76 @@ def drop_schema(
...
 
-class BaseBackend(abc.ABC, _FileIOHandler):
+class _CacheEntry(NamedTuple):
+    name: str
+    orig_op: ops.Relation
+    cached_op_ref: weakref.ref[ops.Relation]
+    finalizer: weakref.finalize
+
+
+class _CacheHandler:
```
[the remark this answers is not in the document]
jcrist: Sure, was just following the convention of `_FileIOHandler`.

ibis/backends/__init__.py
jcrist: Nice catch, fixed.

jcrist:
Hmmm, I thought pre-commit hooks still ran when rebasing? Not sure how that formatting issue slipped through. Anyway, this should be good-to-go.

cpcloud (approved):

PR #9939, opened 2024-08-27
test(tpcds): add test for tpcds query 97

jitingxu1:
## Description of changes

code looks ok, did not understand why failed tolerance, not sure if we should change the tolerance for individual test .

### DIFF
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -3325,6 +3325,39 @@ def test_96(store_sales, household_demographics, time_dim, store):
 
 
 @tpc_test("ds")
+def test_97(store_sales, date_dim, catalog_sales):
+    ssci = (
+        store_sales.join(date_dim, [("ss_sold_date_sk", "d_date_sk")])
+        .filter(_.d_month_seq.between(1200, 1200 + 11))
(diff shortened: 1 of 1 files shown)


ibis/backends/tests/tpc/ds/test_queries.py
```diff
@@ -3227,3 +3227,36 @@ def test_82(item, inventory, date_dim, store_sales):
...
     )
+
+
+@tpc_test("ds")
+def test_97(store_sales, date_dim, catalog_sales):
+    ssci = (
+        store_sales.join(date_dim, [("ss_sold_date_sk", "d_date_sk")])
+        .filter(_.d_month_seq.between(1200, 1200 + 11))
+        .group_by(_.ss_customer_sk, _.ss_item_sk)
+        .select(_.ss_customer_sk.name("customer_sk"), _.ss_item_sk.name("item_sk"))
```
cpcloud, line 3238: This is incorrect. In SQL a `GROUP BY` with no aggregates is equivalent to `DISTINCT`, so you have to use the `.distinct()` method here.

I'll fix it up!

cpcloud:
In general, if your Ibis code doesn't match the results exactly, it's probably the Ibis code that is wrong, not the expected result. Try to look at the code you wrote first before making adjustments to tolerance.

Adjusting tolerance for summation of integers is almost certainly something we should never do :)

PR #9940, opened 2024-08-27
feat(api): support `order_by` in `arbitrary` aggregation

cpcloud:
Adds support for passing `order_by` to `arbitrary`

### DIFF
--- a/ibis/backends/sql/compilers/datafusion.py
+++ b/ibis/backends/sql/compilers/datafusion.py
@@ -135,11 +135,11 @@ class DataFusionCompiler(SQLGlotCompiler):
             return self.f.arrow_cast(arg, f"{PyArrowType.from_ibis(to)}".capitalize())
         return self.cast(arg, to)
 
-    def visit_Arbitrary(self, op, *, arg, where):
+    def visit_Arbitrary(self, op, *, arg, where, order_by):
         cond = ~arg.is_(NULL)
         if where is not None:
             cond &= where
-        return self.agg.first_value(arg, where=cond)
+        return self.agg.first_value(arg, where=cond, order_by=order_by)
 
     def visit_Variance(self, op, *, arg, how, where):
         if how == "sample":
--- a/ibis/backends/sql/compilers/pyspark.py
+++ b/ibis/backends/sql/compilers/pyspark.py
@@ -262,11 +262,9 @@ class PySparkCompiler(SQLGlotCompiler):
             out = sge.IgnoreNulls(this=out)
         return out
 
-    def visit_Arbitrary(self, op, *, arg, where):
+    def visit_Arbitrary(self, op, *, arg, where, order_by):
         # For Spark>=3.4 we could use any_value here
-        if where is not None:
-            arg = self.if_(where, arg, NULL)
-        return sge.IgnoreNulls(this=self.f.first(arg))
+        return sge.IgnoreNulls(this=self.agg.first(arg, where=where, order_by=order_by))
 
     def visit_Median(self, op, *, arg, where):
         return self.agg.percentile(arg, 0.5, where=where)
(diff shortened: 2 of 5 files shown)


gforsyth:
I think we can just xfail `pandas` and `dask` here, since deprecated backends don't get new features.

jcrist:
What's the motivation here? `arbitrary` selects a truly arbitrary value from a column - ordering shouldn't affect this operation at all. I don't see why adding an `order_by` to this agg makes sense, even with ordering the backend is free to pick any value from the column.

cpcloud:
> even with ordering the backend is free to pick any value from the column.

Not sure what you mean, some dbs support ordering in this function:

```
D select any_value(x order by x) from (select 'a' x union select 'b' union select 'c');
┌─────────────────────────┐
│ any_value(x ORDER BY x) │
│         varchar         │
├─────────────────────────┤
│ a                       │
└─────────────────────────┘
D select any_value(x order by x desc) from (select 'a' x union select 'b' union select 'c');
┌──────────────────────────────┐
│ any_value(x ORDER BY x DESC) │
│           varchar            │
├──────────────────────────────┤
│ c                            │
└──────────────────────────────┘
D select arbitrary(x order by x desc) from (select 'a' x union select 'b' union select 'c');
┌──────────────────────────────┐
│ arbitrary(x ORDER BY x DESC) │
│           varchar            │
├──────────────────────────────┤
│ c                            │
└──────────────────────────────┘
D select arbitrary(x order by x asc) from (select 'a' x union select 'b' union select 'c');
│ arbitrary(x ORDER BY x ASC) │
│           varchar           │
├─────────────────────────────┤
│ a                           │
```

In fact, in Trino this is the only efficient way to do a `first`/`last` reduction with an ordering:

```
❯ trino --catalog memory --schema default
trino:default> select arbitrary(x order by x asc) from (select 'a' x union select 'b' union select 'c');
 _col0
-------
 a
(1 row)

Query 20240827_153520_00029_4jb9m, FINISHED, 1 node
Splits: 84 total, 84 done (100.00%)
0.24 [0 rows, 0B] [0 rows/s, 0B/s]

trino:default> select arbitrary(x order by x desc) from (select 'a' x union select 'b' union select 'c');
 _col0
-------
 a
(1 row)

Query 20240827_153524_00034_4jb9m, FINISHED, 1 node
Splits: 84 total, 84 done (100.00%)
0.16 [0 rows, 0B] [0 rows/s, 0B/s]
```

cpcloud:
Just because there's an ordering key doesn't mean there's no arbitrariness. If the ordering key isn't unique then there's still going to be a possibly arbitrary row chosen.

jcrist:
> Not sure what you mean, some dbs support ordering in this function:

In duckdb `arbitrary` is an alias for `first`, where an ordering does make sense. Likewise they implement `any_value` like `first`, but ignoring nulls. Not all backends define `any_value` with that semantics though. Duckdb also allows `order by` in most aggs, even if the ordering would be meaningless.

> In fact, in Trino this is the only efficient way to do a first/last reduction with an ordering:

In trino `any_value`/`arbitrary` aren't defined to mean "first", they mean "an arbitrary non-null value": https://trino.io/docs/current/functions/aggregate.html#any_value. We define `first`/`last` for trino using `array_agg` because of this.

---

What I'm saying is that in ibis the semantics of `arbitrary` are "any value in the column". It's more like bigquery's [any_value](https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#any_value). Given this definition, it's unclear what effect adding an ordering would have. If you order a column, but still pick "any arbitrary value", then what is the ordering doing? I recognize that for many backends `arbitrary` is aliased to `first`, but that's an implementation detail for those backends, the semantics are still "any arbitrary value".

PR #9933, opened 2024-08-26
test(tpcds): add queries 72-74

gforsyth:
## Description of changes

Trino syntax error on the join chains, and datafusion doesn't have date delta

PR #9941, opened 2024-08-27
fix(trino): remove hack that silently breaks join transpilation

cpcloud:
Fixes an issue discovered in #9933 where my failing to remove a workaround caused a bug in transpilation joins.


jcrist (approved):
Should we add a test for a join case that failed before?

cpcloud:
Oh, yeah, good idea!

PR #9944, opened 2024-08-27
test(cleanup): remove dead `xfail_version` clutter

cpcloud:
Remove some test decorator clutter that is no longer required.


cpcloud:
Trino timeout here, and the fix is already merged upstream.

gforsyth (approved):

PR #9945, opened 2024-08-27
feat(api): support `rollup`, `cube` and `grouping_sets` APIs

cpcloud:
## Description of changes

WIP PR implementing support for `rollup`, `cube` and `grouping sets`.

For background on what these are used for and how they work, see
https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-GROUPING-SETS.

In discussing the implementation, I'm going to refer to groupings sets only,
but everything applies to rollups and cubes as well (rollups and cubes are
shorthand for a longer form `GROUPING SET`).

These concepts represent a new kind of "thing" for Ibis. They are neither
column nor scalar expressions that can be evaluated, nor are they tables
really.

The approach I took here was to add a non-`Value` operation for each construct.
These are desugared into `tuple`s of `tuple`s of `Value` expressions (because
multiple grouping sets specifications are allowed) wherever grouping sets are
supported.

One tricky thing is that I had to partition the key specifically requested in
group by, as distinct from those unique value expressions in grouping sets, to
support constructs like this:

```sql
SELECT a, b
FROM t
GROUP BY ROLLUP (a, b)
```

which is not equivalent to

```sql
SELECT a, b
FROM t
GROUP BY a, b, ROLLUP (a, b)
```

Right now, I'm looking for feedback on the approach before fleshing out the
test suite.


gforsyth:
This seems like a reasonable approach to me.  
Small feedback that can always be addressed later:

I think it would be nice to allow passing in bare strings, values, or deferreds to `grouping_sets`, along with tuples of them, e.g.

```python
ibis.grouping_sets(("a", "b"), "a", "b")
```

this would better mimic the SQL syntax (god, i can't believe I'm suggesting that...) and also avoids the awkward spelling of

```python
ibis.grouping_sets(("a", "b"), ("a",), ("b",))
```

And on the naming front, I thing we should take a pointer from DuckDB and rename `grouping` to `grouping_id` -- I think `grouping` is going to cause confusion.

but overall seems solid!

cpcloud:
All good points, definitely changing `grouping` -> `grouping_id`. I'm half tempted to call it `group_id`, those extra three characters are killing me.

cpcloud:
For a full feature and correct implementation, grouping sets will depend on the earliest version of sqlglot contains https://github.com/tobymao/sqlglot/pull/3985.

cpcloud:
This would break users who having columns named `rollup`, `cube`, or `grouping_sets` but that seems okay?

jcrist:
> What do people think about making these keyword arguments to `group_by`?
>
> This would break users who having columns named `rollup`, `cube`, or `grouping_sets` but that seems okay?

I _slightly_ dislike that. Not because it might break existing users (seems unlikely), but because it's harder to visually distinguish keyword-arguments serving as aliases in `group_by` from those specific kwargs.

```python
t.group_by(one=t.a, cube=(t.b,), three=t.c)  # intentionally terrible formatting
```

Another potential downside is it doesn't let users as easily use multiple of these constructs in the same query. Following this line from the docs:

> If multiple grouping items are specified in a single GROUP BY clause, then the final list of grouping sets is the cross product of the individual items.

With separate objects, I'd to be able to pass in multiple of the same kind of object, while kwargs would force the user to handle the composition of them themselves (confession - I'm new to this whole concept, so apologies if this is a dumb example):

```python
t.group_by(ibis.cube("a", "b"), ibis.cube("c", "d"))
```

Individual objects also may be easier for programmatic usage - a system would only need a single list of things to group by, rather than splitting out the various grouping set kind of things.

None of these are blockers of course, and all can be worked around. Just noting a few potential downsides.

cpcloud:
Hm, all good points. I think I"ll stick with the objects for now.

ibis/backends/tests/tpc/queries/duckdb/ds/05.sql
```diff
@@ -81,8 +81,8 @@ WITH ssr AS
      AND d_date BETWEEN cast('2000-08-23' AS date) AND cast('2000-09-06' AS date)
      AND wsr_web_site_sk = web_site_sk
    GROUP BY web_site_id)
-SELECT channel ,
-       id ,
+SELECT nullif(channel, '') AS channel ,
+       nullif(id, '') AS id ,
```
cpcloud, line 85: This change, and the same change in `80.sql` are because of ClickHouse's obsession with never using null values. Apparently this includes the grouping keys for `rollup` and friends.

ibis/expr/types/groupby.py
```diff
@@ -38,10 +40,19 @@ class GroupedTable(Concrete):
...
+    # groupings is allowed to be empty when there are some form of grouping
+    # sets provided
+    #
+    # groupings are *strictly* the things the user has explicitly requested to
+    # group by that are not part of a grouping set
+    groupings: VarTuple[ops.Value]
     orderings: VarTuple[ops.SortKey] = ()
     havings: VarTuple[ops.Value[dt.Boolean]] = ()
 
+    grouping_sets: VarTuple[VarTuple[VarTuple[ir.Value]]] = ()
```
cpcloud, line 52: This makes me sad, and I'll add a comment.

ibis/expr/types/groupby.py
```diff
@@ -291,3 +319,60 @@ def __init__(self, arr, parent):
...
+    return Cube(dims)
+
+
+@public
+def grouping_sets(*dims):
+    return GroupingSets(tuple(map(tuple, map(ibis.util.promote_list, dims))))
+
+
+@deferrable
+def group_id(first, *rest):
```
cpcloud, line 377: All these new functions need docstrings.

ibis/expr/types/relations.py
```diff
@@ -990,9 +993,35 @@ def group_by(
...
+
+            groups.extend(self.bind(**key_exprs))
+            return groups, grouping_sets, rollups, cubes
+
         by = tuple(v for v in by if v is not None)
-        groups = self.bind(*by, **key_exprs)
-        return GroupedTable(self, groups)
+        groups, grouping_sets, rollups, cubes = partition_groups(*by, **key_exprs)
+        if not (groups or grouping_sets or rollups or cubes):
+            raise com.IbisInputError("No grouping keys provided")
```
cpcloud, line 1017: Previously we were using an `Annotated` type hint to enforce that `groups` was non-empty, but we can't do that anymore because it **can** be empty if the user provides at least one grouping set, rollup, or cube and we don't have a way to spell that kind of relationship in the type system (nor do I think we need to build that right now).

ibis/expr/types/relations.py
```diff
@@ -1040,6 +1069,9 @@ def aggregate(
         metrics: Sequence[ir.Scalar] | None = (),
         by: Sequence[ir.Value] | None = (),
         having: Sequence[ir.BooleanValue] | None = (),
+        grouping_sets=(),
+        rollups=(),
+        cubes=(),
```
cpcloud, line 1074: I don't like exposing these arguments in the `aggregate`. Here's why:

1. I have to bind again, similar to what I am doing in the `group_by` method.
1. Putting the grouping set arguments alongside `by` means there's potential for confusion and mixing: e.g., someone can passing `ibis.cube` into `by`, and/or they can pass it into `cube`. If we ultimately go this route then we'll have to handle that.

I suppose neither of these things is the actual worst, but I'm open to suggestions on how to avoid exposing these additional arguments.

ibis/expr/types/relations.py
```diff
@@ -1128,8 +1166,29 @@ def aggregate(
                 else:
                     metrics[metric.name] = metric
 
-        # construct the aggregate node
-        agg = ops.Aggregate(node, groups, metrics).to_expr()
+        keys = frozendict(
```
cpcloud, line 1169: This bit is far too optimistic and very proof-of-concept. More unit tests are required to throw some chaos at this code.

ibis/tests/expr/test_aggregation.py
```diff
@@ -73,3 +73,28 @@ def test_aggregation_where(table, func):
...
+    expr = t.group_by(gs).agg(n=_.count())
+    result = ibis.to_sql(expr, dialect="duckdb")
+    assert len(result)
+
+
+def test_cube(t):
+    gs = ibis.cube("a", "b")
+    expr = t.group_by(gs).agg(n=_.count())
+    result = ibis.to_sql(expr, dialect="duckdb")
+    assert len(result)
```
cpcloud, line 100: These tests are woefully naive and optimistic, need to make them grumpier.

cpcloud:
Keeping this as draft until I can at least address my own feedback here.

ibis/expr/types/relations.py
gforsyth: It doesn't get away from having to bind again (and actually adds a bit more work), but what if we just have all grouping expressions passed in to `by`?

cpcloud: I'll poke at it a bit. It might be easier if there were bound and unbound versions of the rollup/cube/grouping sets objects. Bound would be used internally only, while the user facing APIs are all unbound. The current ones are effectively Unbound.

gforsyth:
I wish I hadn't learned this while working through reviewing this, but I guess it's valid to pass a `ROLLUP` or a `CUBE` as an element of a `GROUPING SET`?  I don't think that works right now, and maybe we leave that for a follow-up, but I thought we should note it down for ourselves.

cpcloud:
Which of course means you can nest grouping sets arbitrarily.

```
D create table t (a int, b int);
D select 1 from t group by grouping sets (rollup (a, b), a, grouping sets (cube (a, b)));
┌───────┐
│   1   │
│ int32 │
├───────┤
│     1 │
│     1 │
└───────┘
```

cpcloud:
This is a *special* level of hell.

cpcloud:
Another "fun" thing with these constructs is that some backends implement rollup/cube/grouping sets on an empty table such that the "grand total" group (usually spelled `()`) return zero rows, and some return one row.

For example:

```sql
create table t (c int);
select * from t group by rollup (c);
```

cpcloud:
I think I'm going to cut this from the 10.0 milestone since it's not a breaking change and we need to get a release out.

PR #9947, opened 2024-08-27
test(tpcds): add query 75

gforsyth:
This is wrong on a small handful of rows and I can't spot the mistake,
but pushing it up before I forget about it.

### DIFF
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -3530,7 +3530,7 @@ def test_75(
     )
 
     all_sales = (
-        _catalog_sales.union(_store_sales, _web_sales)
+        _catalog_sales.union(_store_sales, _web_sales, distinct=True)
         .group_by(
             _.d_year, _.i_brand_id, _.i_class_id, _.i_category_id, _.i_manufact_id
         )
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -3441,6 +3441,10 @@ def test_74(customer, store_sales, date_dim, web_sales):
     return expr
 
 
+@pytest.mark.notyet(
+    ["datafusion"],
+    reason="Invalid argument error: RowConverter column schema mismatch, expected Int32 got Int64",
+)
 @tpc_test("ds")
 def test_75(
     catalog_sales,
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -3483,7 +3483,8 @@ def test_75(
                 _.i_category_id,
                 _.i_manufact_id,
                 sales_cnt=sale_quant - rtn_quant.coalesce(0),
-                sales_amt=sale_price - rtn_amt.coalesce(0.0),
+                sales_amt=sale_price
+                - rtn_amt.coalesce(ibis.literal(0.0, type="decimal(7,2)")),
             )
         )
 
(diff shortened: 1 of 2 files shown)


ibis/backends/tests/tpc/ds/test_queries.py
```diff
@@ -3441,6 +3441,140 @@ def test_74(customer, store_sales, date_dim, web_sales):
...
+        _.ws_ext_sales_price,
+        web_returns,
+        web_returns.wr_order_number,
+        web_returns.wr_item_sk,
+        web_returns.wr_return_quantity,
+        web_returns.wr_return_amt,
+    )
+
+    all_sales = (
+        _catalog_sales.union(_store_sales, _web_sales)
```
cpcloud, line 3533: You can't spot the difference because it's incredibly non-obvious and only after about 30 minutes of tinkering did I realize what the problem is: `a UNION b` only yields distinct rows, while `a UNION ALL b` yields every row from `a` and `b`.

In Ibis, `union` defaults to `distinct=False` which produces `UNION ALL`.

Generally speaking `UNION` is *incredibly* rare in practice, partially because it tends to be outrageously expensive to keep a giant hash table of all unique rows hanging around. I'm not 100% sure why this is the default, but I'm guessing it's because it's a rather large performance foot gun.

```suggestion
        _catalog_sales.union(_store_sales, _web_sales, distinct=True)
```

gforsyth: Wow.  Thank you.  I stared at that SQL for a loooooong time

cpcloud (approved):

gforsyth:
blarg, I'll take a look at the datafusion error.  looks like some type promotion stuff

gforsyth:
Ok, I fixed Clickhouse by adding an explicit cast to `decimal(7,2)` that Clickhouse was upset about for both the transpiled sql and the Ibis query (it doesn't like comparing `decimal(7,2) < float64)`) -- Datafusion is failing to execute the `expected_expr` with `Invalid argument error: RowConverter column schema mismatch, expected Int32 got Int64`, despite my manually adding in a cast to `int64` for all the `int32` columns.  

That can use some more investigation, but I'm just failing it here.

ibis/backends/tests/tpc/queries/duckdb/ds/75.sql
```diff
@@ -13,7 +13,7 @@ WITH all_sales AS
              i_category_id ,
              i_manufact_id ,
              cs_quantity - COALESCE(cr_return_quantity,0) AS sales_cnt ,
-             cs_ext_sales_price - COALESCE(cr_return_amount,0.0) AS sales_amt
+             cs_ext_sales_price - COALESCE(cr_return_amount,CAST(0.0 AS DECIMAL(7,2))) AS sales_amt
```
cpcloud, line 16: This seems like it might be a red herring, since the literal value of the operation passes:

```
SELECT coalesce(CAST(0., 'Decimal(17, 2)'), 0., CAST(0., 'Decimal(7, 2)'))

Query id: 907baa21-6411-4134-a53f-a068ac26ca46

   ┌─coalesce(CAST(0., 'Decimal(17, 2)'), 0., CAST(0., 'Decimal(7, 2)'))─┐
1. │                                                                   0 │
   └─────────────────────────────────────────────────────────────────────┘
```

cpcloud: We're probably inserting casts in Ibis where we shouldn't be.

cpcloud: Lemme poke around and see if I can remove those.

cpcloud: Ok, it seems like ClickHouse is handling the literal case

😮‍💨

cpcloud: This is fine as is, since it's ClickHouse's problem

gforsyth: yeah, not very satisfying, I know

PR #9951, opened 2024-08-28
test(tpc): add test to verify and list all queries are written

cpcloud:
Adds a test for tpch and tpcds to verify that all queries are written, and to list all the ones that are not, when run with `--runxfail`


gforsyth (approved):
:shipit:

PR #9954, opened 2024-08-28
test(tpcds): add queries 76-78

gforsyth:
## Description of changes

76 and 78 are passing, 77 requires rollup, but I tried to write it so it would pass when `rollup` exists, but no promises.


cpcloud (approved):
Nice!

PR #9955, opened 2024-08-28
test(tpcds): add query 64

gforsyth:
So this should probably not get merged yet?  Or we can merge it and mark the
DuckDB failure and investigate further.

My notes so far:

There is some _weird_ stuff happening _only_ inside Pytest when running
this query on DuckDB.  This doesn't happen with any other tpc-ds
queries, and the query in question seems to run without issue on the
other backends that can run the tpc-ds queries.

If I generate sf=1 tpc-ds data in the DuckDB CLI, then run the included
sql file (which is the SQL generated by Ibis for this query), it runs in
about half a second.

If I `%load ibis_tpcds_64_local.py` and then run the `expr` there (which
is a copy-paste of the test code), it takes about half a second.

If I remove the `pytest.timeout` from the `tpc_test` decorator, this
same query, even at sf=.45 (which is empty at that size), takes
_minutes_ to run.

I also thought it might be the difference between using `memory` as the catalog instead of `tpcds`, but no, if I create `tpcds.ddb`, then create the tables in there, it still runs very quickly.

### DIFF
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -3209,14 +3209,6 @@ def test_63(item, store_sales, date_dim, store):
 
 
 @tpc_test("ds", result_is_empty=True)
-@pytest.mark.notyet(
-    ["duckdb"],
-    reason=(
-        "This query times out on DuckDB _only_ when run via pytest. "
-        "No, we have no idea why this happens."
-    ),
-    strict=False,
-)
 def test_64(
     catalog_sales,
     catalog_returns,
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import calendar as cal
 from operator import ge, itemgetter, lt
-from pathlib import Path
 
 import pytest
 
@@ -5184,7 +5183,6 @@ def test_99(catalog_sales, warehouse, ship_mode, call_center, date_dim):
     )
 
 
-@pytest.mark.xfail(raises=AssertionError, reason="not all queries are implemented yet")
 def test_all_queries_are_written():
     variables = globals()
     numbers = range(1, 100)
@@ -5195,12 +5193,5 @@ def test_all_queries_are_written():
         if f"test_{query_number:02d}" in variables:
             query_numbers.remove(query_number)
 
-    file_size = (
-        lambda qn: Path(__file__)
-        .parents[1]
-        .joinpath("queries", "duckdb", "ds", f"{qn:d}.sql")
-        .stat()
-        .st_size
-    )
-    remaining_queries = sorted(query_numbers, key=file_size)
+    remaining_queries = sorted(query_numbers)
     assert remaining_queries == []
--- a/ibis/backends/tests/tpc/ds/test_queries.py
+++ b/ibis/backends/tests/tpc/ds/test_queries.py
@@ -3215,6 +3215,7 @@ def test_63(item, store_sales, date_dim, store):
         "This query times out on DuckDB _only_ when run via pytest. "
         "No, we have no idea why this happens."
     ),
+    strict=False,
 )
 def test_64(
     catalog_sales,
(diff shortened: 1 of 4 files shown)


cpcloud:
This query is absolutely monstrous.

ibis/backends/tests/tpc/ds/test_queries.py
```diff
@@ -3213,6 +3213,169 @@ def test_63(item, store_sales, date_dim, store):
...
+        .join(d3, _.c_first_shipto_date_sk == d3.d_date_sk)
+        .join(store, _.ss_store_sk == store.s_store_sk)
+        .join(cd1, _.ss_cdemo_sk == cd1.cd_demo_sk)
+        .join(
+            cd2,
+            [
+                _.c_current_cdemo_sk == cd2.cd_demo_sk,
+                _.cd_marital_status != cd2.cd_marital_status,
+            ],
+        )
```
cpcloud, line 3290: For some reason, commenting out this join allows the query to finish.

One thing that might be in the mix here that unfortunately still isn't possible to disable is the magic variable discovery done by the DuckDB client.

I can't see any variables that would overlap here, especially in the method that's invoking `self.con.sql` (`_to_duckdb_relation`), but it does seem like that's the only thing feature-wise in the various backends' Python clients that differentiates DuckDB.

At the very least we need the ability to disable this behavior.

gforsyth: I'm just deeply confused why it only happens inside `pytest`

cpcloud: Yes, that doesn't make much sense.

gforsyth:
ok, well, it passes on the other backends and the timeout will handle `duckdb` for now.

I'm also wondering, for the sake of correctness checks, if we should shunt the `tpcds` data creation into CI at `sf=1`

gforsyth:
oh FFS, of course it doesn't timeout on _some_ of the DuckDB jobs

gforsyth:
This query is cursed

cpcloud:
The ClickHouse timeout also does not surprise me, but it also seems to trigger what looks like a resource leakage: the remaining TPC-DS queries cannot run, because the client from query 64 appears to still be hanging around.

cpcloud:
Two totally weird facts:

1. `.tmp` grows quite large while waiting for the query to finish
1. Loading TPC-DS data using `CREATE TABLE` instead of `CREATE VIEW` allows the query to finish on DuckDB quickly.

cpcloud:
Here's a stripped down version of the query that has all the same annoying and mysterious characteristics we're observing:

```python
   @tpc_test("ds", result_is_empty=True)
   def test_64(store_sales, customer_demographics):
       cd1 = customer_demographics
       cd2 = customer_demographics.view()

       expr = (
           store_sales.join(cd1, _.ss_cdemo_sk == cd1.cd_demo_sk)
           .join(
               cd2[["cd_marital_status"]], cd1.cd_marital_status != cd2.cd_marital_status
           )
           .select(cd1.cd_marital_status)
           .limit(1)
       )

       return expr
```

cpcloud:
Found a reproducer without `pytest`:

```python
import duckdb


def test_thing():
   con = duckdb.connect()
   sql = """
SELECT
 t3.cd_marital_status
FROM read_parquet('ci/ibis-testing-data/tpcds/sf=0.45/parquet/store_sales.parquet') AS t2
JOIN read_parquet('ci/ibis-testing-data/tpcds/sf=0.45/parquet/customer_demographics.parquet') AS t3
 ON ss_cdemo_sk = cd_demo_sk
JOIN read_parquet('ci/ibis-testing-data/tpcds/sf=0.45/parquet/customer_demographics.parquet') AS t4
 ON t3.cd_marital_status <> t4.cd_marital_status
LIMIT 1
"""
   result = con.sql(sql)
   result = result.arrow()
   assert len(result) == 1


if __name__ == "__main__":
   test_thing()
```

cpcloud:
Reported upstream here https://github.com/duckdb/duckdb/issues/13657

cpcloud:
Since this is fixed upstream already and the DuckDB release is out soon, I'll suggest that for now we alter our `duckdb/conftest.py` to create tables instead of views.

Not entirely sure what to do about ClickHouse yet.

cpcloud (approved):
Thank you for chugging through this abomination.

cpcloud:
Snowflake passing query 64:

```
❯ pytest -m snowflake ibis/backends/tests/tpc/ds/test_queries.py -k 64 -q
.                                                                                                                                                                                              [100%]
1 passed, 1980 deselected in 18.66s
```

PR #9964, opened 2024-08-30
docs: fix doctests for postgres backend

gmcrocetti:
## Description of changes

Fixes doctests of the postgres backend. I couldn't help but notice that the `run doctests` [step](https://github.com/ibis-project/ibis/actions/runs/10617401640/job/29430553173) of CI is not running for this file and others. I can be overlooking something or is that expected ?

demo

## Issues closed

There's no open issue for this. Let me know case this doesn't make any sense.


cpcloud (approved):

cpcloud:
Thanks!

We don't run doctests in the backend docs, because that would require spinning up a backend.

cpcloud:
I suppose we could do that in the backend test suite proper, 🤔 let me see what's possible there.

PR #9974, opened 2024-08-30
ci: run arm64 job on slightly beefier runner

cpcloud:


cpcloud:
Merging to beef up the runner and get back to reliable green.

Issue #9931, opened 2024-08-26
feat(io): support way to ensure `read_*` apis create physical tables in the backend


jcrist:
Currently `read_*` methods (`read_parquet`/`read_csv`/...) may create a table or view, backend dependent. To the user executing code both should work, but the performance characteristics of each may differ depending on where the data actually lives. Likewise, a view may result in different results over time if the source files are updated by something else, while reading into the backend (and persisting it there) would avoid this issue.

To work around this, I've written code like:

```python
t = ibis.read_parquet(...).cache()  # force doing the read now
```

For backends that create a view this is equivalent (though it leaves a lingering view around). For other backends though this will duplicate a physical table (until #6195 is implemented), doubling the data stored.

It would be nice to add a keyword argument to the `read_*` methods that allows forcing things to be a physical table.

For uniformity it *might* be nice to also have a way to force things to be a view, but:
- not all backends support views, while all have a table concept
- I personally don't have a use case for requiring that this generates a view instead of a table.

Really the distinction we're looking for is semantics of "is the data loaded into the backend, or is it a view on where it sits at rest", not necessarily the DDL terms the backend uses for Table/View. The difference in polars between `read_*` vs `scan_*`, for example.

Possible spellings:

```python
# True means table, False means view (or something else), None is backend default?
# or True means table, False means backend default (maybe table, maybe something else?)
t = con.read_parquet(..., as_table=True)  # or `table=True`, `use_table=True`, `force_table=True`?

# Force a table if False, otherwise :shrug:.
# kinda dislike this one, but it is more precise
t = con.read_parquet(..., maybe_view=False)  # or `allow_view=False`?

t = con.read_parquet(..., type="table")  # or `type="view"`. I hate this one
```

cpcloud:
I think having separate methods makes sense for this, while keyword arguments clutter the API.

How about `load_*` equivalents that unconditionally persist data and mirrorthe `read_*` APIs?

cpcloud:
I think best distinction we can do between any approach here is:

1. Things persist after the session ends
2. Things do not persist after the session ends

The former almost certainly requires physical storage, while the latter can be more less or whatever's well-suited to backend.

If people need something specific, they have to call `create_{view,table}`

jcrist:
> I think having separate methods makes sense for this, while keyword arguments clutter the API.

[...]

PR #9976, opened 2024-08-30
feat(api): avoid caching physical tables

### LINKED ISSUE #6195: feat: make `Table.cache()` a no-op for tables that are already concrete in a backend

jcrist:
Fixes #6195.

This makes `table_expr.cache()` a no-op for physical tables and other expressions not worth caching (defined here as simple column subsets of physical tables).

```python
t = con.table("my_table")

t2 = t.cache()  # this is a no-op since t points to an existing table
assert t2.op() is t.op()

subset = t.select("a", "b")
subset2 = subset.cache()  # this is a no-op, since subset is a simple selection of a concrete table
assert subset2.op() is subset.op()

expr = t.filter(t.a > 10)
expr2 = expr.cache()  # this isn't a no-op, since expr does some actual work
assert expr2.op() is not expr.op()
```

In the original issue (#6195) I wanted to only avoid caching things that were backed by an existing physical table (don't cache tables, do cache views, do cache expressions, ...). In practice writing `_is_this_a_real_database_table` for every backend would be annoying (maybe less annoying once the `.ddl` accessor lands). I'm also not sure if avoiding caching views is that bad.

We do make exceptions for two backends though - `duckdb` and `pyspark`. Both of these backends have lazy-loading for file inputs (meaning creating the `ibis.Table` doesn't load the data yet), and loading the data can sometimes be expensive. In both cases the output of `read_parquet` et. al. is backed by a `TEMP VIEW`. For these backends alone we also will cache temp views to work around this. This means that the pattern of:

```python
t = con.read_parquet("test.parquet")  # this doesn't load data yet
t2 = t.cache()  # this does load data
```

will work for both these backends still. Other backends where `read_parquet` reads immediately (not as a view) will now have `.cache()` as a no-op (where before it would duplicate the data).

---

All that said, I'm not happy with how squishy the definition of "do we cache this expression" is. I think the above is pragmatic and easy to do, but it's also hard to explain since there's so much "it depends" going on. Mostly pushing this up for discussion for now (I haven't added tests yet).

A few options that would make this simpler/more uniform/easier to explain:

- Always cache anything that isn't backed by a true database TABLE (or a simple column subset of one). This would require more (simple) functions across our backends to check what kind of thing was backing an `ops.DatabaseTable`, but the uniformity there may be worth it.
- Always skip caching anything backed by a `PhysicalTable`. This makes the choice much easier, since it requires no backend introspection. For `duckdb`/`pyspark`/others add an optional kwarg to control whether the data is loaded immediately or lazily (#9931). I know I closed that one in favor of #6195, but now I'm waffling. This moves the decision about how to load data to the caller rather than after the call, which feels better to me (more localized control). Could be a backend-specific kwarg (so not there for all our backends), or something more general. Possible names: `cache=True`, `view=False`, `lazy=False`, idk.
- Maybe scrap all of this and leave the current behavior. I opened the original issue with some ML use cases in mind where I wanted to make use of `.cache()` uniformly to ensure I was working on a physical table for efficiency, but now I'm not sure if the complications here are worth it.

Currently leaning towards option 2, but could hear arguments for any of them.

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -805,17 +805,22 @@ class CacheHandler:
             Cached table
         """
         entry = self._cache_op_to_entry.get(table.op())
-        if entry is None or (cached_op := entry.cached_op_ref()) is None:
-            cached_op = self._create_cached_table(util.gen_name("cached"), table).op()
-            entry = CacheEntry(
-                table.op(),
-                weakref.ref(cached_op),
-                weakref.finalize(
-                    cached_op, self._finalize_cached_table, cached_op.name
-                ),
-            )
-            self._cache_op_to_entry[table.op()] = entry
-            self._cache_name_to_entry[cached_op.name] = entry
+        if entry is not None and (cached_op := entry.cached_op_ref()) is not None:
+            # Entry was already cached
+            return ir.CachedTable(cached_op)
+
+        if not self._should_cache_table_expr(table):
+            # Expression isn't worth caching, no-op
+            return ir.CachedTable(table.op())
+
+        cached_op = self._create_cached_table(util.gen_name("cached"), table).op()
+        entry = CacheEntry(
+            table.op(),
+            weakref.ref(cached_op),
+            weakref.finalize(cached_op, self._finalize_cached_table, cached_op.name),
+        )
+        self._cache_op_to_entry[table.op()] = entry
+        self._cache_name_to_entry[cached_op.name] = entry
         return ir.CachedTable(cached_op)
 
     def _finalize_cached_table(self, name: str) -> None:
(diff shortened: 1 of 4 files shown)


jcrist:
cc @cpcloud and @gforsyth for thoughts on if any of this is sane.

ibis/backends/__init__.py
```diff
@@ -838,6 +843,31 @@ def _finalize_cached_table(self, name: str) -> None:
...
 
+    def _should_cache_table_expr(self, expr: ir.Table) -> bool:
+        """Checks if a given table expression is worth caching."""
+        op = expr.op()
+
+        # Don't cache if an expression is a column subselection of a physical table.
+        while isinstance(op, (ops.Project, ops.DropColumns)):
+            if isinstance(op, ops.Project) and not all(
+                isinstance(v, ops.Field) for v in op.values.values()
+            ):
```
cpcloud, line 854: Most of this code is unrelated to the backend and is about some property of an expression.

Can we define this as an attribute on `ops.Relation` subclasses? Seems like it'd be much less squishy that way.

ibis/backends/duckdb/__init__.py
```diff
@@ -1589,6 +1589,24 @@ def _get_schema_using_query(self, query: str) -> sch.Schema:
             }
         )
 
+    def _should_cache_physical_table(self, op: ops.PhysicalTable) -> bool:
+        if isinstance(op, (ops.DatabaseTable, ops.UnboundTable)):
```
cpcloud, line 1593: This feels like a property we should encode on the operation itself instead of leaving it up to the backend.

Is that not possible for some reason?

ibis/backends/__init__.py
jcrist: Meaning the decision is consistent and only based on the operation type (no query in the backend needed)? That's option 2 listed above.

PR #9997, opened 2024-09-02
fix(polars): use `flatten` API for `ArrayFlatten` implementation to avoid large string upcast

### LINKED ISSUE #9995: bug: flatten does not work with polars
### What happened?

I'm taking the example from the [docs](https://ibis-project.org/reference/expression-collections#examples-6) for the `flatten` operation. I'm just adding the line to set the backend to `polars`

```python
import ibis
import ibis.selectors as s
from ibis import _

# Use polars
ibis.set_backend("polars")

schema = {
     "empty": "array<array<int>>",
     "happy": "array<array<string>>",
     "nulls_only": "array<array<struct<a: array<string>>>>",
     "mixed_nulls": "array<array<string>>",
 }
data = {
     "empty": [[], [], []],
     "happy": [[["abc"]], [["bcd"]], [["def"]]],
     "nulls_only": [None, None, None],
     "mixed_nulls": [[], None, [None]],
 }
import pyarrow as pa
t = ibis.memtable(
     pa.Table.from_pydict(
         data,
         schema=ibis.schema(schema).to_pyarrow(),
     )
)

t.happy.flatten()
```

I expected to see:
```
┏━━━━━━━━━━━━━━━━━━━━━┓
┃ ArrayFlatten(happy) ┃
┡━━━━━━━━━━━━━━━━━━━━━┩
│ array<string>       │
├─────────────────────┤
│ ['abc']             │
│ ['bcd']             │
│ ['def']             │
└─────────────────────┘
```

But instead I see the following error:
`ArrowNotImplementedError: Unsupported cast from large_list<item: large_string> to utf8 using function cast_string`

After some investigation, I think that the issue comes from the [polars translator](https://github.com/ibis-project/ibis/blob/54888968db818971c6db7e758f20c91a70f37abb/ibis/backends/polars/compiler.py#L996) for the operation ArrayFlatten:
```python
@translate.register(ops.ArrayFlatten)
def array_flatten(op, **kw):
    return pl.concat_list(translate(op.arg, **kw))
```

Changing the implementation the following seems to solve the issue and produces the expected result
```python
@translate.register(ops.ArrayFlatten)
def array_flatten(op, **kw):
    return translate(op.arg, **kw).flatten()
```

9.3.0

polars 1.5.0

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Closes #9995.

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -1008,7 +1008,14 @@ def array_collect(op, in_group_by=False, **kw):
 
 @translate.register(ops.ArrayFlatten)
 def array_flatten(op, **kw):
-    return pl.concat_list(translate(op.arg, **kw))
+    result = translate(op.arg, **kw)
+    return (
+        pl.when(result.is_null())
+        .then(None)
+        .when(result.list.len() == 0)
+        .then([])
+        .otherwise(result.flatten())
+    )
 
 
 _date_methods = {
--- a/ibis/backends/polars/tests/test_client.py
+++ b/ibis/backends/polars/tests/test_client.py
@@ -2,8 +2,12 @@ from __future__ import annotations
 
 import pytest
 
+import ibis
 from ibis.backends.tests.errors import PolarsSQLInterfaceError
 from ibis.util import gen_name
+
+pd = pytest.importorskip("pandas")
+tm = pytest.importorskip("pandas.testing")
 
 
 def test_cannot_run_sql_after_drop(con):
@@ -22,3 +26,14 @@ def test_cannot_run_sql_after_drop(con):
     con.drop_table(name)
     with pytest.raises(PolarsSQLInterfaceError):
         con.sql(sql)
+
+
+def test_array_flatten(con):
+    data = {"id": range(3), "happy": [[["abc"]], [["bcd"]], [["def"]]]}
+    t = ibis.memtable(data)
+    expr = t.select("id", flat=t.happy.flatten()).order_by("id")
+    result = con.to_pyarrow(expr)
+    expected = pd.DataFrame(
+        {"id": data["id"], "flat": [row[0] for row in data["happy"]]}
+    )
+    tm.assert_frame_equal(result.to_pandas(), expected)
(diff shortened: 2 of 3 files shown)


ibis/backends/polars/compiler.py
```diff
@@ -1014,7 +1014,14 @@ def array_collect(op, in_group_by=False, **kw):
...
 @translate.register(ops.ArrayFlatten)
 def array_flatten(op, **kw):
-    return pl.concat_list(translate(op.arg, **kw))
+    result = translate(op.arg, **kw)
+    return (
+        pl.when(result.is_null())
+        .then(None)
+        .when(result.list.len() == 0)
+        .then([])
+        .otherwise(result.flatten())
```
cpcloud, line 1023: The shenanigans here are because of https://github.com/pola-rs/polars/issues/18522.

ibis/backends/tests/test_array.py
```diff
@@ -1061,9 +1056,9 @@ def test_array_flatten(backend, flatten_data, column, expected):
     data = flatten_data[column]
     t = ibis.memtable({column: data["data"]}, schema={column: data["type"]})
     expr = t[column].flatten()
-    result = backend.connection.execute(expr)
+    result = backend.connection.to_pyarrow(expr)
```
jcrist, line 1059: Why this change here? Does `con.to_pandas` not work, but `con.to_pyarrow().to_pandas()` does work? If so, we should fix that.

jcrist (approved):

ibis/backends/tests/test_array.py
cpcloud: It works, but `to_pyarrow` reproduces the issue due to polars having different typing behavior for `concat_list` versus `flatten`.

ibis/backends/polars/tests/test_client.py
```diff
@@ -22,3 +25,14 @@ def test_cannot_run_sql_after_drop(con):
     con.drop_table(name)
     with pytest.raises(PolarsSQLInterfaceError):
         con.sql(sql)
+
+
+def test_array_flatten(con):
```
cpcloud, line 30: Moved this here in order to avoid dealing with Snowflake's PyArrow array output, which differs from other backends because of the untyped nature of its arrays.

PR #10001, opened 2024-09-03
test(risingwave): bump to latest risingwave release and enable now-passing tests

cpcloud:
I noticed we had a very old version of RisingWave, and bumped it to the latest release, which fixed a number of bugs and added support for `RANGE` window function frames, which caused a number of tests to start XPASS-ing.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -547,7 +547,7 @@ services:
       - impala
 
   risingwave:
-    image: ghcr.io/risingwavelabs/risingwave:nightly-20240204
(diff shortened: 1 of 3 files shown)


jcrist (approved):
LGTM

jcrist:
Side question - why does the bot upgrade some images but not others? I noticed risingwave had an old image the other day (debugging the distinct fusion PR), but thought that may have been intentional since the bot usually bumps them.

cpcloud:
I think it might be because their image is stored on ghcr.

PR #10003, opened 2024-09-03
fix(backends): pass kwargs to _from_url() in every case

nextchamp-saqib:
The original `kwargs` are not passed to the `backend._from_url()` if `scheme != "file"`


cpcloud (changes requested):
This PR needs a test!

cpcloud:
I will add a test here so that we can get this in for 9.4.1, if it is in fact a bug.

cpcloud (approved):

PR #10004, opened 2024-09-03
docs(datatypes): install rich traceback handler for prettier exception tracebacks

cpcloud:
Cleans up the exception printing for the datatypes.qmd file showing invalid mixing of unrelated columns.

### DIFF
--- a/docs/concepts/datatypes.qmd
+++ b/docs/concepts/datatypes.qmd
@@ -101,6 +101,21 @@ t2
 ```
 
 ```{python}
+#| echo: false
+import IPython
+import rich
+import rich.traceback as rtb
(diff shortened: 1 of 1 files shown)


cpcloud:
This is kind of weird and special casey, so it's fine if this is rejected in favor of leaving it as is.

It just didn't seem worth the effort to try and roll in rich exception printing everywhere.

cpcloud:
Here's what it looks like with this PR

image

gforsyth:
> This is kind of weird and special casey

I think explainer docs are the right place to do weird and special-casey things in the interest of making the information more accessible

gforsyth (approved):

Issue #9987, opened 2024-09-01
bug: horizontal sum 30 times slower via Ibis


MarcoGorelli:
### What happened?

I ran this:
```python
In [1]: import polars as pl; import ibis; import numpy as np

In [2]: rng = np.random.default_rng(1)

In [3]: arr = rng.integers(-100, 100, size=(100, 100))

In [4]: df = pl.DataFrame(arr)

In [5]: t = ibis.memtable(df)

In [6]: results = %timeit -o df.select(pl.sum_horizontal('*'))
2.69 ms ± 72.4 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [7]: results.best
Out[7]: 0.0025612837100015893

In [8]: from functools import reduce

In [9]: from operator import add

In [10]: results = %timeit -o t.select(SUM=reduce(add, map(t.__getitem__, t.columns))).to_polars()
60.3 ms ± 2.81 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [11]: results.best
Out[11]: 0.05732215999978507
```

With 255 columns (as in the financial data I was trying this out with) I get
```
Binder Error: Maximum recursion depth exceeded (Maximum: 128) while binding
```
as reported here https://ibis-project.zulipchat.com/#narrow/stream/405263-general/topic/.E2.9C.94.20sum.28axis.3D1.29

Running duckdb natively I could just do
```

In [19]: result = %timeit -o duckdb.sql(f'select {"+".join(df.columns)} as sum from df')
1.8 ms ± 69.5 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [20]: result.best
Out[20]: 0.0016851913000000423
```
which is also fast (and doesn't crash on 255 columns)

Please introduce an expression in Ibis which can efficiently make use of Polars' `sum_horizontal`

### What version of ibis are you using?

9.3.0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

MarcoGorelli:
> Or just for the above to be more efficient compilation-wise?

this is fine, anything that's (almost) as good as `sum_horizontal` / duckdb's native `'+'` string concatenation works, thanks 🙏

cpcloud:
The 128 limit is not an Ibis issue, it's a limit in DuckDB's expression parsing regardless of how Ibis parenthesizes expressions.

It's reproducible without Ibis. I'm not sure what you're testing.

[...]

PR #10007, opened 2024-09-03
perf(sql): avoid extraneous operator parenthesis

jcrist:
This PR is motivated by #9987, and is one part of decreasing the compilation overhead here.

This PR:
- Adds a benchmark for the problem. On main the SQL generation tests fail due to excessive recursion within SQLGlot
- Modifies our binary operator handling code to reduce some boilerplate
- Modifies the parenthesizing behavior of associative operators to avoid unnecessary parenthesis. Doing so unlocks an optimization within SQLGlot to generate SQL without recursion (excessive parenthesis defeat this optimization).
- Adds a snapshot test for operators to check the behavior

There are still a few cases where we excessively apply parenthesis (mostly in backends where a function is used instead of a binary operator), but this seemed good enough for now and is a nice improvement overall. After this PR the benchmarks now run successfully. I have some follow-ups to further improve performance in other places, but figured I'd split the PRs out by topic.

### DIFF
--- a/ibis/backends/sql/compilers/base.py
+++ b/ibis/backends/sql/compilers/base.py
@@ -239,20 +239,6 @@ TRUE = sge.true()
 STAR = sge.Star()
 
 
-def parenthesize_inputs(f):
-    """Decorate a translation rule to parenthesize inputs."""
-
-    def wrapper(self, op, *, left, right):
-        return f(
-            self,
-            op,
-            left=self._add_parens(op.left, left),
-            right=self._add_parens(op.right, right),
-        )
-
-    return wrapper
-
-
 @public
 class SQLGlotCompiler(abc.ABC):
     __slots__ = "f", "v"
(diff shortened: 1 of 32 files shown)


cpcloud:
Nice!

What does the performance versus DuckDB raw SQL look like?

jcrist:
```python
from functools import reduce
from operator import add
from timeit import Timer

import pyarrow as pa
import ibis

con = ibis.duckdb.connect()

N = 100
data = pa.Table.from_pydict({f"x{i}": [1, 2, 3] for i in range(N)})
t = con.create_table("test", data)
expr = t.select(sum=reduce(add, map(t.__getitem__, t.columns)))
sql = f"SELECT {'+'.join(t.columns)} AS sum FROM test"

for name, func in [
    ("duckdb", lambda: con.raw_sql(sql).arrow()),
    ("ibis-duckdb", lambda: expr.to_pyarrow()),
]:
    n, t = Timer(func).autorange()
    print(f"{name}: {1000 * t / n:.1f} ms")
```

On this PR:

```
$ python bench.py
duckdb: 2.6 ms
ibis-duckdb: 25.9 ms
```

With another upcoming PR:

```
$  python bench.py
duckdb: 2.5 ms
ibis-duckdb: 7.2 ms
```

ibis/backends/sql/compilers/base.py
```diff
@@ -469,6 +460,19 @@ def impl(self, _, *, _name: str = target_name, **kw):
         for op, target_name in cls.SIMPLE_OPS.items():
             setattr(cls, methodname(op), make_impl(op, target_name))
 
+        # Define binary op methods, only if BINARY_INFIX_OPS is set on the
+        # compiler class.
+        if binops := cls.__dict__.get("BINARY_INFIX_OPS", {}):
```
cpcloud, line 465: Is this checking specifically for subclass overrides of this attribute?

ibis/backends/sql/compilers/base.py
```diff
@@ -393,45 +379,50 @@ class SQLGlotCompiler(abc.ABC):
...
-        # Binary operations
-        ops.Add,
-        ops.Subtract,
-        ops.Multiply,
-        ops.Divide,
-        ops.Modulus,
-        ops.Power,
+    BINARY_INFIX_OPS = {
+        # Numeric
+        ops.Add: (sge.Add, True),
```
cpcloud, line 384: The bool here seems like it should be a property of the operation instead of a flag.

It's hard to tell what the boolean is referring to unless you know where it's being used.

What do you think about sticking a class attribute on the ops for this property?

ibis/backends/sql/compilers/base.py
```diff
@@ -469,6 +460,19 @@ def impl(self, _, *, _name: str = target_name, **kw):
         for op, target_name in cls.SIMPLE_OPS.items():
             setattr(cls, methodname(op), make_impl(op, target_name))
 
+        # Define binary op methods, only if BINARY_INFIX_OPS is set on the
+        # compiler class.
+        if binops := cls.__dict__.get("BINARY_INFIX_OPS", {}):
+
+            def make_binop(sge_cls, associative):
```
cpcloud, line 467: Seems like most or all of the `associative` arguments can go away if you make it an op attribute.

cpcloud (approved):
Not blocking, but I do think the `.associative` approach would be a bit cleaner.

ibis/backends/sql/compilers/base.py
jcrist: Yeah. If we did `cls.BINARY_INFIX_OPS` it would define the methods in every subclass (since no subclasses currently define `BINARY_INFIX_OPS`), clobbering any explicit `visit_*` methods the subclass defined. We'd need to do this for `SIMPLE_OPS` etc... too, except that every compiler subclass currently sets that attribute themselves with at least one override.

PR #10008, opened 2024-09-03
test(sigcheck): check function signature parity across backends

gforsyth:
Opening this in favor of #9383 -- that PR also included all of the breaking changes to unify function signatures and it was too much at once.  This PR adds only the signature checking mechanism, plus the requisite xfails to lay out which inconsistencies are currently in Ibis.

## Motivation

We want to ensure that, for a given backend, that the argument names, plus usage of positional, positional-only, keyword, and keyword-only arguments match, so that there is API consistency when moving between backends.

I've grabbed a few small parts of some of the utilities in Scott Sanderson's
`python-interface` project (https://github.com/ssanderson/python-interface).

While the upstream is no longer maintained, the goal of that project
aligns quite well with some of the issues we face with maintaining
consistent interfaces across backends.

Note that while the upstream project focused on _runtime_ enforcement of these signatures matching, here it is only run in the test suite.

## Rough procedure

Any method that doesn't match can be skipped entirely (this is useful for things like `do_connect`, which cannot reasonably be assumed to match) or individually (by specifying a `pytest.param` and marking the failing backends).

Then we scrape across the common parent classes and add any methods that are NOT currently specified in the pre-existing xfailed ones.  It's a bit of a nuisance, but it's done, and ideally the manual listing of the inconsistent methods goes away as we unify things.

I've opted for not checking that type annotations match, because that seems... unreasonable.

This would satisfy #9125 once all of the xfail markers are removed, e.g., it checks that all keyword and positional arguments are standardized.

### DIFF
--- a/ibis/backends/tests/test_signatures.py
+++ b/ibis/backends/tests/test_signatures.py
@@ -1,0 +1,27 @@
+from __future__ import annotations
+
+import inspect
+
+import pytest
+
+from ibis.backends import _FileIOHandler
+from ibis.backends.tests.signature.typecheck import compatible
+
+params = []
+
+for module in [_FileIOHandler]:
+    methods = list(filter(lambda x: not x.startswith("_"), dir(module)))
+    for method in methods:
+        params.append((_FileIOHandler, method))
+
+
+@pytest.mark.parametrize("base_cls, method", params)
+def test_signatures(base_cls, method, backend_cls):
+    if not hasattr(backend_cls, method):
+        pytest.skip(f"Method {method} not present in {backend_cls}, skipping...")
+
+    base_sig = inspect.signature(getattr(base_cls, method))
+    backend_sig = inspect.signature(getattr(backend_cls, method))
+
+    # Usage is compatible(implementation_signature, defined_interface_signature, ...)
+    assert compatible(backend_sig, base_sig, check_annotations=False)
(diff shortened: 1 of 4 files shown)


ibis/backends/tests/test_signatures.py
```diff
@@ -0,0 +1,195 @@
...
+                "pyspark",
+            ]
+        ),
+    ),
+    "to_parquet_dir": pytest.param(
+        BaseBackend,
+        "to_parquet_dir",
+        marks=pytest.mark.notyet(["pyspark"]),
+    ),
+    "truncate_table": pytest.param(
```
cpcloud, line 164: So, basically the goal is to, over time, remove as many elements of this dict as possible?

cpcloud (approved):

ibis/backends/tests/test_signatures.py
gforsyth: Ideally, we'd remove all of them.  There's a more generic "skip for all backends" list for things like `do_connect`, but I think we should be able to remove all of the entries on this list if we standardize positional keyword names, and enforce keyword-only arguments

gforsyth: Also, once this is merged in, it will also prevent us from adding new methods with differing argument names (assuming those methods are defined in the base classes, which they should be)

PR #10011, opened 2024-09-04
perf(ir): Reduce recreated nodes in replace

### LINKED ISSUE #9987: bug: horizontal sum 30 times slower via Ibis

jcrist:
This is on top of #10007.

During analysis/compilation ibis makes a lot of calls to `Node.replace`. This traverses the op graph attempting, matching against every node and replacing those that match according to a rewrite rule/mapping/function.

In common operation _most_ nodes don't match a specific pattern. Previously even nodes who didn't match a replacer and whose children didn't match were still recreated via the `__recreate__` method. This method invokes the full validation machinery to create a new instance, and can be _expensive_.

To work around this, we now track whether a node's children have been replaced. If the haven't (and the node doesn't match a specific rewrite rule), we use the original node unchanged rather than recreating it. This can have substantial performance implications, especially for large op graphs.

This is part of the work to resolve #9987.


jcrist:
Relevant pytest benchmark, comparing against #10007. No use comparing against `main` since the benchmark errors out due to recursion errors.

```
------------------------------------------------------------------------------ benchmark 'test_large_add[compile-128]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                  Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_large_add[compile-128] (0001_6ea09d7)     56.9698 (3.04)     72.6744 (1.89)     59.2685 (3.02)     4.7686 (1.33)     57.6405 (3.04)     0.4090 (1.58)          2;3  16.8724 (0.33)         18           1
test_large_add[compile-128] (NOW)              18.7390 (1.0)      38.5406 (1.0)      19.6444 (1.0)      3.5739 (1.0)      18.9855 (1.0)      0.2588 (1.0)           1;2  50.9050 (1.0)          30           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

-------------------------------------------------------------------------------- benchmark 'test_large_add[compile-256]': 2 tests --------------------------------------------------------------------------------
Name (time in ms)                                   Min                 Max                Mean            StdDev              Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_large_add[compile-256] (0001_6ea09d7)     113.6950 (3.05)     129.3550 (2.41)     117.6677 (2.93)     6.6371 (1.19)     114.1953 (3.03)     6.9677 (11.61)         2;0   8.4985 (0.34)          8           1
test_large_add[compile-256] (NOW)               37.2581 (1.0)       53.7490 (1.0)       40.1524 (1.0)      5.5758 (1.0)       37.7044 (1.0)      0.6001 (1.0)           4;6  24.9051 (1.0)          27           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------- benchmark 'test_large_add[construct-128]': 2 tests ----------------------------------------------------------------------------
Name (time in ms)                                   Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_large_add[construct-128] (0001_6ea09d7)     5.8337 (1.04)     6.1676 (1.05)     5.8834 (1.05)     0.0451 (1.51)     5.8721 (1.04)     0.0246 (1.0)         14;17  169.9693 (0.96)        169           1
test_large_add[construct-128] (NOW)              5.5883 (1.0)      5.8541 (1.0)      5.6295 (1.0)      0.0299 (1.0)      5.6252 (1.0)      0.0247 (1.00)         13;6  177.6364 (1.0)         169           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_large_add[construct-256]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                    Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
test_large_add[construct-256] (0001_6ea09d7)     11.9588 (1.05)     27.9307 (2.37)     12.2340 (1.06)     1.7236 (25.72)    12.0447 (1.05)     0.0526 (1.0)           1;4  81.7396 (0.94)         85           1
test_large_add[construct-256] (NOW)              11.4349 (1.0)      11.7728 (1.0)      11.5243 (1.0)      0.0670 (1.0)      11.5066 (1.0)      0.0664 (1.26)         19;7  86.7731 (1.0)          87           1
```

jcrist:
Comparing against native duckdb for this operation:

```python
from functools import reduce
from operator import add
from timeit import Timer

import pyarrow as pa

import ibis

con = ibis.duckdb.connect()

N = 100
data = pa.Table.from_pydict({f"x{i}": [1, 2, 3] for i in range(N)})
t = con.create_table("test", data)
expr = t.select(sum=reduce(add, map(t.__getitem__, t.columns)))
sql = f"SELECT {'+'.join(t.columns)} AS sum FROM test"

for name, func in [
    ("duckdb", lambda: con.raw_sql(sql).arrow()),
    ("ibis-duckdb", lambda: expr.to_pyarrow()),
]:
    n, t = Timer(func).autorange()
    print(f"{name}: {1000 * t / n:.1f} ms")
```

```
$ python bench.py
duckdb: 2.6 ms
ibis-duckdb: 7.3 ms
```

Note that the overhead (~5 ms) is fixed for the query size, not the data size. For larger data the overhead would be much smaller. Also note that `duckdb` is _not_ great at this query as written (adding a large number of columns together is not a core use case), but there are faster ways to write this in duckdb.

ibis/backends/sql/rewrites.py
```diff
@@ -385,24 +385,23 @@ def sqlize(
...
-    result = simplified.replace(wrap)
-    ctes = [cte.parent for cte in result.find(CTE, ordered=True)]
+        def apply_ctes(node, kwargs):
+            new = node.__recreate__(kwargs) if kwargs else node
+            return CTE(new) if node in ctes else new
 
-    return result, ctes
+        result = result.replace(apply_ctes)
+        return result, [cte.parent for cte in result.find(CTE, ordered=True)]
+    return result, []
```
cpcloud, line 404: Could we switch this around and exit early on `not ctes`? Then you can dedent the code for the `else` condition.

ibis/backends/sql/rewrites.py
```diff
@@ -385,24 +385,23 @@ def sqlize(
...
+        result = result.replace(reduce(operator.or_, post_rewrites))
 
     # extract common table expressions while wrapping them in a CTE node
-    ctes = extract_ctes(simplified)
+    ctes = extract_ctes(result)
 
-    def wrap(node, _, **kwargs):
-        new = node.__recreate__(kwargs)
-        return CTE(new) if node in ctes else new
+    if ctes:
```
cpcloud, line 396: Does this save any time? Just curious. No need to dig into it either if you don't already know.

cpcloud (approved):
Excellent work!

ibis/backends/sql/rewrites.py
jcrist: Yes. If there are no CTEs then we avoid an entire op graph traversal, along with the logic overhead within `replace`.

ibis/backends/sql/rewrites.py
jcrist: :shrug:, perhaps mildly better, but not worth spending CI cycles on. May move it around in a follow up.

Issue #8135, opened 2024-01-29
refactor(ir): make `ops.SortKey` a non-value operation


kszucs:
Since aliases are valid in many places but `asc()`/`desc()` are really only valid in `order_by()` I wonder if `asc()`/`desc()` should return non `Value` operations? This would make it easier to forbid e.g. `t.x.asc() + 1`.

_Originally posted by @jcrist in https://github.com/ibis-project/ibis/issues/8100#issuecomment-1912256153_

PR #10016, opened 2024-09-04
fix(sql): properly parenthesize binary ops containing named expressions

### LINKED ISSUE #9975: bug: incorrect parenthesization when mixing binary operators named expressions
### What happened?

The `parenthesize_inputs` function is designed to improve SQL generation by adding parentheses for certain operators. It functions correctly in scenarios like this:
```
t = bq.table("bigframes-dev.ibis_gbq_testing.functional_alltypes")
f1 = t.int_col + t.int_col
ibis.to_sql(f1 * t.int_col, dialect="bigquery")
``` 
This produces the expected SQL output:
```
SELECT
  (
    `t0`.`int_col` + `t0`.`int_col`
  ) * `t0`.`int_col` AS `Multiply_Add_int_col_int_col_int_col`
FROM `bigframes-dev`.`ibis_gbq_testing`.`functional_alltypes` AS `t0`
```

However, the functionality breaks down when column aliases are introduced:
```
bq = ibis.bigquery.connect(project_id="bigframes-dev")
t = bq.table("bigframes-dev.ibis_gbq_testing.functional_alltypes")
f1 = t.int_col + t.int_col
ibis.to_sql(f1.name("col1") * t.int_col.name("col2"), dialect="bigquery")
```
This generates the following incorrect SQL:
```
SELECT
  `t0`.`int_col` + `t0`.`int_col` * `t0`.`int_col` AS `Multiply_col1_col2`
FROM `bigframes-dev`.`ibis_gbq_testing`.`functional_alltypes` AS `t0`
```

9.2.0

BigQuery

### Relevant log output

_No response_

### Code of Conduct

jcrist:
Fixes #9975.

In the long run the best fix is probably something like #8135 (except for `Alias`) so we could automatically coerce aliases to their core values within `Value.__coerce__` (except for the very few ops where we care about preserving them). For now though this is a pragmatic fix that resolves the issue.

### DIFF
--- a/ibis/backends/impala/__init__.py
+++ b/ibis/backends/impala/__init__.py
@@ -266,7 +266,7 @@ class Backend(SQLBackend):
     def _fetch_from_cursor(self, cursor, schema):
         from ibis.formats.pandas import PandasData
 
-        results = fetchall(cursor)
+        results = fetchall(cursor, schema.names)
         return PandasData.convert_table(results, schema)
 
     @contextlib.contextmanager
@@ -1260,9 +1260,10 @@ class Backend(SQLBackend):
                     cur.execute(insert_stmt, row)
 
 
-def fetchall(cur):
+def fetchall(cur, names=None):
     batches = cur.fetchcolumnar()
-    names = list(map(operator.itemgetter(0), cur.description))
+    if names is None:
+        names = list(map(operator.itemgetter(0), cur.description))
     df = _column_batches_to_dataframe(names, batches)
     return df
 
--- a/ibis/backends/polars/__init__.py
+++ b/ibis/backends/polars/__init__.py
@@ -466,12 +466,19 @@ class Backend(BaseBackend, NoUrl):
         streaming: bool = False,
         **kwargs: Any,
     ) -> pl.DataFrame:
-        lf = self.compile(expr, params=params, **kwargs)
+        table_expr = expr.as_table()
+        lf = self.compile(table_expr, params=params, **kwargs)
         if limit == "default":
             limit = ibis.options.sql.default_limit
         if limit is not None:
             lf = lf.limit(limit)
-        return lf.collect(streaming=streaming)
+        df = lf.collect(streaming=streaming)
+        # XXX: Polars sometimes returns data with the incorrect column names.
+        # For now we catch this case and rename them here if needed.
+        expected_cols = tuple(table_expr.columns)
+        if tuple(df.columns) != expected_cols:
+            df = df.rename(dict(zip(df.columns, expected_cols)))
+        return df
 
     def execute(
         self,
--- a/ibis/backends/tests/sql/snapshots/test_compiler/test_subquery_where_location/decompiled.py
+++ b/ibis/backends/tests/sql/snapshots/test_compiler/test_subquery_where_location/decompiled.py
@@ -11,7 +11,7 @@ alltypes = ibis.table(
     },
 )
 param = ibis.param("timestamp")
-f = alltypes.filter((alltypes.timestamp_col < param.name("my_param")))
+f = alltypes.filter((alltypes.timestamp_col < param))
 agg = f.aggregate([f.float_col.sum().name("foo")], by=[f.string_col])
 
 result = agg.foo.count()
--- a/ibis/formats/pandas.py
+++ b/ibis/formats/pandas.py
@@ -109,14 +109,11 @@ class PandasData(DataMapper):
 
     @classmethod
     def convert_table(cls, df, schema):
-        if len(schema) != len(df.columns):
-            raise ValueError(
-                "schema column count does not match input data column count"
-            )
+        if schema.names != tuple(df.columns):
+            raise ValueError("schema names don't match input data columns")
 
         columns = {
-            name: cls.convert_column(series, dtype)
-            for (name, dtype), (_, series) in zip(schema.items(), df.items())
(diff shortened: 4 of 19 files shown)


ibis/backends/tests/sql/snapshots/test_compiler/test_subquery_where_location/decompiled.py
```diff
@@ -11,7 +11,7 @@
     },
 )
 param = ibis.param("timestamp")
-f = alltypes.filter((alltypes.timestamp_col < param.name("my_param")))
+f = alltypes.filter((alltypes.timestamp_col < param))
```
jcrist, line 14: I modified this test to avoid the name since it didn't effectively matter. Roundtripping an aliased parameter used in a context where aliases are ignored doesn't seem necessary.

ibis/expr/operations/core.py
```diff
@@ -172,6 +172,14 @@ class Binary(Value):
     left: Value
     right: Value
 
+    def __init__(self, left, right, **kwargs):
+        # Dealias left and right, passing through any additional fields
+        if isinstance(left, Alias):
+            left = left.arg
+        if isinstance(right, Alias):
+            right = right.arg
```
jcrist, line 180: I'm aware of `unwrap_alias`, but didn't feel like reorging the code to make use of it here. The alias check is only 2 lines of code anyway.

cpcloud: What about other operations? This seems like it only solves the problem for binary operations.

cpcloud: Oh, because it's not going into the op.

Well, that's definitely something we should fix in `unwrap_aliases`, not here.

cpcloud: I think if you replace `unwrap_alias` with this, then it solves the issue without needing to hack:

```python
from ibis.expr.rewrites import p, x

def unwrap_alias(node: ops.Value) -> ops.Value:
    """Unwrap an alias node."""
    return node.replace(p.Alias(x) >> x, filter=ops.Value)
```

ibis/tests/expr/test_value_exprs.py
```diff
@@ -731,6 +731,12 @@ def test_string_mul(table, left, right):
     assert isinstance(expr.op(), ops.Repeat)
 
 
+def test_binop_strips_aliases(table):
```
cpcloud, line 734: This test isn't necessary with the rewrite rule approach.

ibis/expr/operations/core.py
jcrist: Good idea! I think applying it in `sqlize` makes more sense than at the `expr/types` level. Pushed something up.

jcrist:
Hmmm, this _almost_ works. The failure point is the `TableUnnest` operation which calculates its schema names dynamically from the `column` attribute (a `Value[dt.Array]`). When we strip aliases, the node is recreated and a new schema is generated, which causes the rewritten expr to have a different schema (see the failing doctest). I _think_ solving this would be as simple as adding a `column_name` field in `TableUnnest`, rather than relying on `op.column.name`. If that seems viable (cc @cpcloud) I'm happy to finish this up.

cpcloud:
Yeah that seems very viable.

ibis/backends/polars/__init__.py
```diff
@@ -466,12 +466,19 @@ def _to_dataframe(
...
+        table_expr = expr.as_table()
+        lf = self.compile(table_expr, params=params, **kwargs)
         if limit == "default":
             limit = ibis.options.sql.default_limit
         if limit is not None:
             lf = lf.limit(limit)
-        return lf.collect(streaming=streaming)
+        df = lf.collect(streaming=streaming)
+        # XXX: Polars sometimes returns data with the incorrect column names.
+        # For now we catch this case and rename them here if needed.
```
jcrist, line 477: See https://github.com/pola-rs/polars/issues/18581 for one case of this.

ibis/formats/pandas.py
```diff
@@ -109,14 +109,11 @@ def infer_table(cls, df):
 
     @classmethod
     def convert_table(cls, df, schema):
-        if len(schema) != len(df.columns):
-            raise ValueError(
-                "schema column count does not match input data column count"
-            )
+        if schema.names != tuple(df.columns):
+            raise ValueError("schema names don't match input data columns")
```
jcrist, line 113: Previously we'd implicitly rename in `PandasData.convert_table`, but not in the other formats. This would lead to name mismatch bugs that would only show up in `to_pyarrow`/`to_polars` (hence the `Table.unnest()` bug noticed above that failed only in a doctest (goes through `to_pyarrow()`)). I've instead opted to remove this in `to_pandas` in favor of fixes upstream in the specific backends.

ibis/backends/impala/__init__.py
```diff
@@ -266,7 +266,7 @@ def raw_sql(self, query: str):
     def _fetch_from_cursor(self, cursor, schema):
         from ibis.formats.pandas import PandasData
 
-        results = fetchall(cursor)
+        results = fetchall(cursor, schema.names)
```
jcrist, line 269: The cursor names aren't case sensitive, which can lead to a name mismatch. Just use the schema names when we have a schema.

jcrist (commented):
Provided tests pass, I think this is ready.

ibis/formats/pandas.py
cpcloud: Seems like the right move.

cpcloud (approved):
NICE

PR #10020, opened 2024-09-04
test(strings): improve selected string tests

gforsyth:
BigQuery:

-   `rstrip` and `lstrip` are broken because of a SQLGlot bug, already fixed upstream.

Snowflake:

-   `rstrip` and `lstrip` are broken because of a SQLGlot bug, already fixed upstream.

MSSQL:

-   `contains` might work, but requires &ldquo;fulltext index&rdquo; and I can&rsquo;t be bothered
-   `rpad` not defined
-   `lpad` not defined
-   `find_in_set` not defined

Oracle:

-   Treats `len(🐍) == 2`, should be able to use `LENGTHC` instead
-   `strip` maps to `trim`, but `TRIM` doesn&rsquo;t accept custom characters, unlike `RTRIM` and `LTRIM`

Flink:

-   Treats `len(🐍) == 2`
-   `strip` maps to `trim`, but `TRIM` doesn&rsquo;t accept custom characters, unlike `RTRIM` and `LTRIM`
    There is a function in the docs called `BTRIM` but maybe it&rsquo;s only in dev?  But that should work?

Impala (woof):

-   Treats `len(🐍) == 4`, `len(Éé) == 4`
-   No `TRIM` for anything but whitespace.  This appears to be a limitation of our Impala compiler.
-   Accented characters do not get upper-cased or lower-cased, just left as-is.

Risingwave

-   Accented characters do not get upper-cased or lower-cased, just left as-is.

MySQL

-   Treats `len(🐍) == 4`, should use `CHAR_LENGTH`

PySpark

-   `TRIM` only trims spaces

SQLite

-   Accented characters do not get upper-cased or lower-cased, just left as-is.

Clickhouse

-   Lengths of accented characters and emoji are of byte width, not character length
-   This also impacts padding
    BUT:
    There are UTF8 aware versionf for all of these

<a id="org8b87fa0"></a>

# What should the string APIs do?

## `find_in_set`

We state in the docs for `find_in_set` that if a value contains a comma, we
return `-1` (the same behavior as if there is no match).  Currently no backend
does this, the 5 backends that support the operation all return the index of the
comma-containing field (`duckdb`, `datafusion`, `mysql`, `postgres`,
`risingwave`)

We can just update the docstring?

<a id="org1899fed"></a>

## Padding

The SQL convention with padding is that if the original string is longer than
the pad length, the string is trimmed to that length.

The Python convention is to leave strings >= pad<sub>length</sub> alone.

Which would we like to adopt?

I think the SQL convention here is <span class="underline">terrible</span> and we should follow Python.

<a id="org641a8d5"></a>

## UTF-8

Some backends will support this, some won&rsquo;t, but, we should be consistent.

I think that `length`, `lpad`, and `rpad` should all be UTF-8 variants.

Propose adding `len_bytes` as an additional string method for the byte length of a string.

### DIFF
--- a/ibis/backends/impala/__init__.py
+++ b/ibis/backends/impala/__init__.py
@@ -513,7 +513,7 @@ class Backend(SQLBackend):
         if schema is not None:
             schema = ibis.schema(schema)
 
-        if temp is not None:
+        if temp:
             raise NotImplementedError(
                 "Impala backend does not yet support temporary tables"
             )
--- a/ibis/backends/impala/tests/snapshots/test_string_builtins/test_string_builtins/lstrip/out.sql
+++ b/ibis/backends/impala/tests/snapshots/test_string_builtins/test_string_builtins/lstrip/out.sql
@@ -1,3 +1,3 @@
 SELECT
-  LTRIM(`t0`.`string_col`) AS `LStrip(string_col)`
+  LTRIM(`t0`.`string_col`, ' \t\n\r\v\f') AS `LStrip(string_col)`
 FROM `functional_alltypes` AS `t0`
--- a/ibis/backends/impala/tests/snapshots/test_string_builtins/test_string_builtins/rstrip/out.sql
+++ b/ibis/backends/impala/tests/snapshots/test_string_builtins/test_string_builtins/rstrip/out.sql
@@ -1,3 +1,3 @@
 SELECT
-  RTRIM(`t0`.`string_col`) AS `RStrip(string_col)`
+  RTRIM(`t0`.`string_col`, ' \t\n\r\v\f') AS `RStrip(string_col)`
 FROM `functional_alltypes` AS `t0`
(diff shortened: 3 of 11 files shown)


gforsyth:
Ok, I've fixed `strip`, `lstrip` and `rstrip` for all backends except:

* `pyspark`: because they don't have upstream rtrim and ltrim functions that take characters to trim
* `snowflake`: upstream fixed sqlglot bug, just waiting on it to release
* `bigquery`: upstream fixed sqlglot bug, just waiting on it to release

I'm going to open issues to track the other proposed work in this PR description and handle them separately, since some of those might be breaking changes, while everything here is strictly a bugfix.

cpcloud (approved):
🚢 it!

cpcloud:
Doing a rebase merge to pick up all the individual fixes.

PR #10031, opened 2024-09-05
docs(accursed): add cursed knowledge page

gforsyth:
We have a lot of catching up to do, but we can at least note down these
horrors as we reencounter them.


cpcloud (approved):
Strictly cursed!

PR #10037, opened 2024-09-06
fix(duckdb): support version 1.1.0

cpcloud:
Fixes related to DuckDB 1.1. I am not sure why partitioning results in a different column ordering now. This is not ready for merge until DuckDB cuts an official release. Other than the partitioning question this is ready to go. Tests will fail until a release, because of the version number comparison being wonky due to the dev suffix.


ibis/backends/tests/test_aggregation.py
```diff
@@ -466,6 +466,11 @@ def mean_and_std(v):
             lambda t, where: t.string_col.approx_nunique(where=where),
             lambda t, where: t.string_col[where].nunique(),
             id="approx_nunique",
+            marks=pytest.mark.xfail_version(
+                duckdb=["duckdb>=1.1"],
+                raises=AssertionError,
+                reason="not exact, even at this tiny scale",
+            ),
```
jcrist, line 473: Don't need to handle it in this PR, but we should probably split any approx function tests out to test approximate results rather than just xfailing.

jcrist (approved):

ibis/backends/tests/test_aggregation.py
cpcloud: Agree.

PR #10038, opened 2024-09-06
perf(duckdb): make default backend initialization a bit cheaper

cpcloud:
## Description of changes

This PR makes the default backend slightly faster to initialize by using the
`duckdb` module's `default_connection` attribute.

Not a massive savings, but local testing shows around a 10% speedup without a ton
of additional code on our side.

### DIFF
--- a/ibis/config.py
+++ b/ibis/config.py
@@ -178,7 +178,7 @@ def _default_backend() -> Any:
         return backend
 
     try:
-        import duckdb as _  # noqa: F401
+        import duckdb
     except ImportError:
         raise com.IbisError(
             """\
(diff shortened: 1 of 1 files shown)


ibis/config.py
```diff
@@ -203,7 +203,9 @@ def _default_backend() -> Any:
 
     import ibis
 
-    options.default_backend = con = ibis.duckdb.connect(":memory:")
+    options.default_backend = con = ibis.duckdb.from_connection(
+        duckdb.default_connection
```
jcrist, line 207: What apis in duckdb result in their default connection being used when using duckdb alone? Is there any risk in us sharing this connection with other things a user might be using?

cpcloud: Yeah, perhaps.

I'll close the PR, since the gain isn't worth having to think about who owns the connection.

PR #10042, opened 2024-09-06
fix(duckdb): free memtables based on operation lifetime

### LINKED ISSUE #10041: ibis.memtable leaks memory with default (duckdb) backend
### What happened?

Creating ephemeral ibis.memtables uses memory that is (seemingly) never freed. Reproducible example: 
```
import ibis
import os
print(os.getpid())

row = {
    'a': 123,
    'b': 'asdf',
    'c': None
}

rows = [row] * 1000000

for i in range(1, 10000):
    ibis.memtable(rows).count().execute()
```
Above script ran >20 minutes on a 32GB Win11 machine before crashing with 
```
numpy.core._exceptions._ArrayMemoryError: Unable to allocate 7.63 MiB for an array with shape (1000000,) and data type uint64
```
On my machine, task manager reported process using ~20GB for a long time before crashing, so I instead recommend tracking with perfmon (Performance Monitor), and monitoring Private Bytes from Process V2 object. This metric monotonically increased until 50+GB before crashing, but I understand this can differ between systems. 

This contrived example mimics a workflow where data chunks are read, modified using Ibis API, and written out. This memory leakage imposes an artificial limit to such pipelines.

Any workarounds would be very appreciated 🙏 

9.0.0

Only tested with default duckdb 0.10.3

### Relevant log output

```sh
Win11, Python 3.12.4, ibis 9.0.0, duckdb 0.10.3
```

### Code of Conduct

cpcloud:
Closes #10041.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -7,6 +7,7 @@ import contextlib
 import os
 import urllib
 import warnings
+import weakref
 from operator import itemgetter
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
@@ -1604,6 +1605,12 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
         except (duckdb.CatalogException, duckdb.InvalidInputException):
             # only register if we haven't already done so
             self.con.register(name, op.data.to_pyarrow(op.schema))
+
+            # if we don't aggressively unregister tables duckdb will keep a
+            # reference to every memtable ever registered, even if there's no
+            # way for a user to access the operation anymore, resulting in a
+            # memory leak
+            weakref.finalize(op, self.con.unregister, name)
 
     def _register_udfs(self, expr: ir.Expr) -> None:
         con = self.con
--- a/ibis/backends/duckdb/tests/test_client.py
+++ b/ibis/backends/duckdb/tests/test_client.py
@@ -417,3 +417,12 @@ lat,lon,geom
     path.write_bytes(data)
     t = con.read_csv(path, all_varchar=all_varchar, **input)
     assert t.schema()["geom"].is_geospatial()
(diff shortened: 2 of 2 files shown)


cpcloud:
@liudvikasakelis Would you mind giving this a try?

ibis/backends/duckdb/__init__.py
```diff
@@ -1598,6 +1599,12 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
             # only register if we haven't already done so
             self.con.register(name, op.data.to_pyarrow(op.schema))
 
+            # if we don't aggressively unregister tables duckdb will keep a
+            # reference to every memtable ever registered, even if there's no
+            # way for a user to access the operation anymore, resulting in a
+            # memory leak
+            weakref.finalize(op, self.con.unregister, name)
```
cpcloud, line 1606: Ultimately we probably need to do similar things for other local backends, and perhaps even non-local ones to ensure that we don't balloon storage.

jcrist: The non-local ones at least shouldn't fill local memory, but they would result in an excess of unreachable temp tables. Something to fixup for later (I'll open an issue).

jcrist (approved):

ibis/backends/duckdb/__init__.py
ncclementi: Do you want to open an issue, if you haven't already, to track this for other local backends?

ncclementi (approved):
LGTM

ibis/backends/duckdb/__init__.py
jcrist: Already done: https://github.com/ibis-project/ibis/issues/10044

cpcloud:
What hell hath I wrought?

Issue #10044, opened 2024-09-06
bug: ibis should release resources when an `ibis.memtable` op is GC'd


jcrist:
`ibis.memtable` is useful for constructing small tables out of in-memory data in a backend agnostic way. When `ibis` executes an expression containing a `memtable` node, a new "table" is created in the backend with the backing data (only if it wasn't created yet), and then used like any other table when executing the query. How the "table" is created is backend dependent:

- For most SQL backends (`postgres`, `bigquery`, ...) registering a memtable is equivalent to `con.create_table(unique_name, data, temp=True)`. These are temporary tables that are automatically cleaned up when the session is closed.
- Some local (e.g. `duckdb`) have more native ways of registering in memory data that doesn't require a copy. These are also cleaned up when the session ends.

Ibis currently will avoid duplicate registrations of the same memtable in the same backend, but we don't do anything to automatically cleanup the memtables if they go out of scope, instead we rely on them being cleaned up when the connection is closed.

In #10041 (and #10042) it was noted that this behavior is nonideal in the local case, since these backends generally keep around references to the in-memory data, even if the user has no way of accessing those tables anymore. The situation would be the same in the remote SQL backend case too, except in that case the local memory usage would stay the same and there'd just be an excessive number of temp tables created in the backend. The local backend case is definitely a bigger issue, but we might want to automatically de-register memtables on GC in all cases.

Sketch of the intended memtable behavior:

```python
# Create a new memtable. Note that this doesn't do anything 
# in any backends yet, just creates a new expr node
t = ibis.memtable(...)

# Execute a memtable-backed expression on `con`
# This registers a new memtable in `con`
con.execute(t.select("a"))

# Execute another expression on `con` backed by the same memtable.
# Since this one was already registered, the backend just reuses the
# existing table ref.
con.execute(t.select("b"))

# Execute an expression on `con2` backed by the memtable.
# Since this is a new backend, a new memtable is registered
con2.execute(t.select("b"))

# Close `con2`, releasing resources (including those used in it by `t`)
con2.disconnect()

# Drop all references to `t`, `con` can now cleanup any resources used by `t`
del t
```

PR #10053, opened 2024-09-07
refactor(backends): split memtable existence check out

cpcloud:
First of ~two~ a few PRs to implement #10044. This one factors out memtable existence checks to allow a generic implementation of register-only-if-not-exists, as well as to allow improvement over time of this check as we discover faster existence check implementations.

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -1109,11 +1109,15 @@ class BaseBackend(abc.ABC, _FileIOHandler, CacheHandler):
         if self.supports_python_udfs:
             raise NotImplementedError(self.name)
 
+    def _in_memory_table_exists(self, name: str) -> bool:
+        return name in self.list_tables()
+
     def _register_in_memory_tables(self, expr: ir.Expr) -> None:
         for memtable in expr.op().find(ops.InMemoryTable):
-            self._register_in_memory_table(memtable)
+            if not self._in_memory_table_exists(memtable.name):
+                self._register_in_memory_table(memtable)
 
-    def _register_in_memory_table(self, op: ops.InMemoryTable):
+    def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         if self.supports_in_memory_tables:
             raise NotImplementedError(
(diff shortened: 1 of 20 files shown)


ibis/backends/druid/__init__.py
```diff
@@ -33,7 +33,6 @@ class Backend(SQLBackend):
     name = "druid"
     compiler = sc.druid.compiler
     supports_create_or_replace = False
-    supports_in_memory_tables = True
```
cpcloud, line 36: The default value of this flag is `True`, hence the removal in a few places.

ibis/backends/exasol/__init__.py
```diff
@@ -253,50 +253,47 @@ def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
                 f"got null typed columns: {null_columns}"
             )
 
-        # only register if we haven't already done so
-        if (name := op.name) not in self.list_tables():
```
cpcloud, line 257: The only bits I changed here were to remove this conditional, and then dedent what was previously indented inside the branch.

cpcloud:
I'll split out the perf improvements to a separate PR.

ibis/backends/__init__.py
```diff
@@ -1109,11 +1109,15 @@ def _register_udfs(self, expr: ir.Expr) -> None:
         if self.supports_python_udfs:
             raise NotImplementedError(self.name)
 
+    def _table_exists(self, name: str) -> bool:
```
cpcloud, line 1112: Perhaps this should be call `_memtable_exists` to discourage expand the scope to include `database`/`catalog` lookups.

jcrist: You're too fast for me - I had this exact comment written in a pending review :). For consistency with the other methods it should probably be `_in_memory_table_exists` though (or rename the other ones :shrug:).

jcrist (approved):
LGTM (besides the method name).

ibis/backends/__init__.py
cpcloud: Renamed to `_in_memory_table_exists`.

PR #10055, opened 2024-09-07
refactor(backends): clean up resources produced by `memtable`

### LINKED ISSUE #10044: bug: ibis should release resources when an `ibis.memtable` op is GC'd

cpcloud:
## Description of changes

PR on top of #10053 to clean up temporary tables produced by `ibis.memtable`.

The approach is a generalization of the one taken in #10042.

The main caveat here, which I don't _think_ is a blocker is that the following
code will start to fail:

```python
table = ibis.memtable({"a": [1, 2, 3]}, name="foo")
register_memtable(table)

table = None  # very likely to cause `table` op to be GC'd, and thus delete the underlying resource
run_query_against("foo")  # fails because `foo` has been dropped
```

## Issues closed

Closes #10044.

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import abc
 import collections.abc
+import contextlib
 import functools
 import importlib.metadata
 import keyword
@@ -1116,7 +1117,9 @@ class BaseBackend(abc.ABC, _FileIOHandler, CacheHandler):
         for memtable in expr.op().find(ops.InMemoryTable):
             if not self._in_memory_table_exists(memtable.name):
                 self._register_in_memory_table(memtable)
-                self._register_memtable_finalizer(memtable)
+                weakref.finalize(
+                    memtable, self._finalize_in_memory_table, memtable.name
+                )
 
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         if self.supports_in_memory_tables:
@@ -1124,10 +1127,15 @@ class BaseBackend(abc.ABC, _FileIOHandler, CacheHandler):
                 f"{self.name} must implement `_register_in_memory_table` to support in-memory tables"
             )
 
-    def _register_memtable_finalizer(self, op: ops.InMemoryTable) -> None:
+    def _finalize_in_memory_table(self, name: str) -> None:
+        """Wrap `_finalize_memtable` to suppress exceptions."""
+        with contextlib.suppress(Exception):
+            self._finalize_memtable(name)
+
+    def _finalize_memtable(self, name: str) -> None:
         if self.supports_in_memory_tables:
             raise NotImplementedError(
-                f"{self.name} must implement `_register_memtable_finalizer` to support in-memory tables"
+                f"{self.name} must implement `_finalize_memtable` to support in-memory tables"
             )
 
     def _run_pre_execute_hooks(self, expr: ir.Expr) -> None:
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -1116,11 +1116,18 @@ class BaseBackend(abc.ABC, _FileIOHandler, CacheHandler):
         for memtable in expr.op().find(ops.InMemoryTable):
             if not self._in_memory_table_exists(memtable.name):
                 self._register_in_memory_table(memtable)
+                self._register_memtable_finalizer(memtable)
 
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         if self.supports_in_memory_tables:
             raise NotImplementedError(
                 f"{self.name} must implement `_register_in_memory_table` to support in-memory tables"
+            )
+
+    def _register_memtable_finalizer(self, op: ops.InMemoryTable) -> None:
+        if self.supports_in_memory_tables:
+            raise NotImplementedError(
+                f"{self.name} must implement `_register_memtable_finalizer` to support in-memory tables"
             )
 
     def _run_pre_execute_hooks(self, expr: ir.Expr) -> None:
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -7,7 +7,6 @@ import contextlib
 import os
 import urllib
 import warnings
-import weakref
 from operator import itemgetter
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
@@ -1614,7 +1613,7 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         self.con.register(op.name, op.data.to_pyarrow(op.schema))
 
-    def _register_memtable_finalizer(self, op: ops.InMemoryTable):
+    def _finalize_memtable(self, name: str) -> None:
         # if we don't aggressively unregister tables duckdb will keep a
         # reference to every memtable ever registered, even if there's no
         # way for a user to access the operation anymore, resulting in a
@@ -1622,7 +1621,7 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
         #
         # we can't use drop_table, because self.con.register creates a view, so
         # use the corresponding unregister method
-        weakref.finalize(op, self.con.unregister, op.name)
+        self.con.unregister(name)
 
     def _register_udfs(self, expr: ir.Expr) -> None:
         con = self.con
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -160,12 +160,9 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
             properties.append(sge.TemporaryProperty())
             catalog = "temp"
 
-        temp_memtable_view = None
-
         if obj is not None:
             if not isinstance(obj, ir.Expr):
                 table = ibis.memtable(obj)
-                temp_memtable_view = table.op().name
             else:
                 table = obj
 
@@ -233,9 +230,6 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
                             actions=[sge.RenameTable(this=final_table)],
                         ).sql(self.name)
                     )
-
-        if temp_memtable_view is not None:
-            self.con.unregister(temp_memtable_view)
 
         return self.table(name, database=(catalog, database))
 
@@ -1620,10 +1614,14 @@ class Backend(SQLBackend, CanCreateDatabase, CanCreateSchema, UrlFromPath):
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         self.con.register(op.name, op.data.to_pyarrow(op.schema))
 
+    def _register_memtable_finalizer(self, op: ops.InMemoryTable):
         # if we don't aggressively unregister tables duckdb will keep a
         # reference to every memtable ever registered, even if there's no
         # way for a user to access the operation anymore, resulting in a
         # memory leak
+        #
+        # we can't use drop_table, because self.con.register creates a view, so
+        # use the corresponding unregister method
         weakref.finalize(op, self.con.unregister, op.name)
 
     def _register_udfs(self, expr: ir.Expr) -> None:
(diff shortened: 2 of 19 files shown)


ibis/backends/exasol/__init__.py
```diff
@@ -382,31 +384,27 @@ def create_table(
         else:
             temp_name = name
 
-        table = sg.table(temp_name, catalog=database, quoted=quoted)
-        target = sge.Schema(this=table, expressions=column_defs)
+        table_expr = sg.table(temp_name, catalog=database, quoted=quoted)
```
cpcloud, line 387: I changed these variable names because `table = <not the memtable>` causes it to get GC'd. I gave another example of this in the PR description.

cpcloud:
Still need to add at least one cross-backend test to verify the behavior.

cpcloud:
I'll split out the faster memtable checks into a separate PR once #10053 is merged.

cpcloud:
Going to split this out into in-memory versus not-in-memory backends since the former are more important to clean up more aggressively.

cpcloud:
After hours of trying to understand what is happening with MySQL I was able to get a reproducer:

I can only reproduce this with a Python 3.12 environment.

```
pytest --randomly-seed=159558151 -m mysql -x --pdb ibis/backends/tests/test_generic.py -vv
```

Still not sure why executing SQL in a finalizer causes the socket on the connection to get set to `None`, so I'm going to disable memtable cleanup for MySQL for now. It's fairly low risk in that memtables are created with `TEMPORARY`, so they'll get clean up when the session ends.

We won't be able to prevent build up of disk space when constructing large memtables in a loop, but we won't keep all the storage around when the session ends.

cpcloud:
I know I said I would break out the in-memory backends into their own PR, but that would require no-op-ing the non-in-memory backends and then changing that code again. That doesn't seem worth the trouble, given the hopefully manageable small number of changes here.

ibis/backends/exasol/__init__.py
```diff
@@ -278,12 +278,14 @@ def process_item(item: Any):
...
             if not df.empty:
                 self.con.ext.insert_multi(name, rows)
-        atexit.register(self._clean_up_tmp_table, ident)
 
-    def _clean_up_tmp_table(self, ident: sge.Identifier) -> None:
-        with self._safe_raw_sql(
-            sge.Drop(kind="TABLE", this=ident, exists=True, cascade=True)
-        ):
+    def _register_memtable_finalizer(self, op: ops.InMemoryTable):
+        weakref.finalize(op, self._clean_up_tmp_table, op.name)
```
jcrist, line 283: Side nit: `exasol` sets `supports_in_memory_tables = True` at the top, but since it redefines all the memtable methods should probably just delete that flag.

ibis/backends/__init__.py
```diff
@@ -1116,13 +1116,20 @@ def _register_in_memory_tables(self, expr: ir.Expr) -> None:
...
                 self._register_in_memory_table(memtable)
+                self._register_memtable_finalizer(memtable)
 
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         if self.supports_in_memory_tables:
             raise NotImplementedError(
                 f"{self.name} must implement `_register_in_memory_table` to support in-memory tables"
             )
 
+    def _register_memtable_finalizer(self, op: ops.InMemoryTable) -> None:
```
jcrist, line 1127: Having individual backends define their own method to create a finalizer feels a bit off to me. Could you instead define a method with a standard signature as a finalizer, like:

```python
def _finalize_memtable(self, name: str) -> None:
    ...
```

Then call `weakref.finalize` in `_register_in_memory_tables` uniformly for all backends? Should be effectively the same, and would keep all the weakref code in one spot.

jcrist (commented):
A could questions/suggestions around error handling, but otherwise LGTM.

ibis/backends/__init__.py
cpcloud: I think it's important the overridden methods be separate as opposed to together with the registration. This makes it easier to identify which backends need custom finalizers.

ibis/backends/duckdb/__init__.py
```diff
@@ -1620,10 +1614,14 @@ def _in_memory_table_exists(self, name: str) -> bool:
...
 
+    def _register_memtable_finalizer(self, op: ops.InMemoryTable):
         # if we don't aggressively unregister tables duckdb will keep a
         # reference to every memtable ever registered, even if there's no
         # way for a user to access the operation anymore, resulting in a
         # memory leak
+        #
+        # we can't use drop_table, because self.con.register creates a view, so
+        # use the corresponding unregister method
         weakref.finalize(op, self.con.unregister, op.name)
```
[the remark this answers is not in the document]
cpcloud: Definitely prefer option 2. I'd like to avoid adding any state.

ibis/backends/tests/test_client.py
```diff
@@ -1767,3 +1767,71 @@ def test_insert_into_table_missing_columns(con, temp_table):
...
+    ["clickhouse", "flink"],
+    raises=AssertionError,
+    reason="memtables are assembled every time",
+)
+@pytest.mark.notyet(
+    ["mysql"],
+    raises=AssertionError,
+    reason="can't execute SQL inside of a finalizer without breaking everything",
+)
+def test_memtable_cleanup_by_overwriting_variable(con):
```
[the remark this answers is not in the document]
cpcloud: Yep, I don't think it's hitting anything special.

ibis/backends/tests/test_client.py
```diff
@@ -1767,3 +1767,71 @@ def test_insert_into_table_missing_columns(con, temp_table):
...
+    t = ibis.memtable({"a": [1, 2, 3], "b": list("def")}, name=name)
+
+    # the table isn't registered until we actually execute, and since we
+    # haven't yet executed anything, the table shouldn't be there
+    assert name not in con.list_tables()
+
+    # execute, which means the table is registered and should be visible in
+    # con.list_tables()
+    con.execute(t.select("a"))
+    assert name in con.list_tables()
```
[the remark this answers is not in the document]
cpcloud: Temporary tables are guaranteed to show up there. In most cases a memtable is a temporary table.

I agree though that your suggested assertion is more robust to interface changes.

ibis/backends/__init__.py
cpcloud: Oh, I think I misunderstood.

cpcloud: Yeah, I think I can do what you're suggesting.

ibis/backends/duckdb/__init__.py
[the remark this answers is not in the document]

ibis/backends/__init__.py
cpcloud: Done.

jcrist (approved):
Presuming tests pass, LGTM.

cpcloud:
BigQuery is good:

```
❯ pytest -m bigquery -n auto --dist loadgroup --snapshot-update -q
bringing up nodes...
...........sssssssssssssssssssss.......x.....x.........x............x.............x....x........x.x..xx.....s..x..x..x...x......x...x.s...ssssssssssssssssssssssssssssssssssssssssssssssssssss [  8%]
sssssssssssssssssssssssssssssssssssssssssssssss.........s......x...xx.x.xx........x.....s........................x........x......x...........s................................................ [ 17%]
.x......x.....................x......x.........................x........................................................xx.....x....x.X............x.............xx..x.......x.....xx....x...x [ 26%]
.....xx.......x.x..x.....X.x...x.x....x.....x....x.x...........x......x.xx.x.xXxx...x.x.xxXxx....Xx....x...x........x.........x...x......x..................................x......x.........x [ 35%]
..........x.x...x..xxxxx.xx.x..xx.x.x.........x...x..x.........x...x..........x....x.......x..............x......x....x.............x....x.x............x.............x............x.......... [ 44%]
....xx.........x....x...x.x..x.............xx...x.xx.x.x..xx...xxxx..xxxx...x.....xxxx.xx.....x......xxx.....x.x.........x....x..xx..x....x........x.x...x..............x.......x.x....xx.x... [ 53%]
........x.x..x............x........xx..xx...x..x.....x......x.x.x...x....x......xx...x.......x...x....x.x...x..x..............x..x.......................x........x...x...x.....x.........x... [ 62%]
...x..x......x.......x........x..........x..x..............x.x...x..........x.....x....x..x................xx....xx.....x.....................x.................s.........s................... [ 71%]
.....................................................x.s.xxxx.xxx.x.xx.xxx.x..xx...xxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.xxxxxxxxxx.xxxxxxxxxxxxxxxxxx.xxxxxx................... [ 79%]
.............................x.............................x.....................x.......x........x...............x.x...x.....................x............................................... [ 88%]
..............x..x...................................................xxxxxxxx.xx...........................x...x.......................x..............................................x....... [ 97%]
.................................................                                                                                                                                              [100%]
1663 passed, 128 skipped, 343 xfailed, 5 xpassed in 344.90s (0:05:44)
```

cpcloud:
Snowflake is good:

```
❯ pytest -m snowflake -n 8 --dist loadgroup --snapshot-update -q && pytest -m bigquery -n auto --dist loadgroup --snapshot-update -q
bringing up nodes...
.......x.....................s..................x..............x....x.............................x..................x...................x...x........x...x..................x................ [  9%]
............x.....x.......x........x..x.x...............x....................x....................................xx....................x...x..................x...x.........................x [ 19%]
.x.........x.....x.......x..x.........x..................x............................x..........x........x....x.....x......................x..........x.................x............x....... [ 29%]
..x.........x.....x..xx....x...x.......x..............xx..s.................x...............xxx.xx...............x......x..x.......x.x.x.......x......................s........x.............. [ 39%]
.......................s.......s........x.....xx.........xx...x..............x............xx.x......x..................x....xx......x............x.................x....x............x..x..... [ 49%]
x.x.....xx.x......x...xx..xx...x..x....................x..................x...............................................................x..............x.................................xx. [ 59%]
.........................x..xx..xxxx...x.x...x.xx.x.........xx..xx.....x.xxxx.x.xxxxxxxxx.x..xxxxx.x....xxxxx...x...xx.xx..xxx....xxxx..........xxx...x.....x.xx..x......x.................... [ 69%]
..........x..........x.s...x...x.........x..x..xxxxxxxxxxxx.x....................................................x........................x.......x........................................... [ 79%]
.....................x..x...........x........x.....x..x......................x...x........x.............xx...x...x.........x...............x.x.........x..........x....x...xxxxxxxx.x...x....x [ 89%]
.x.xx..........s...s...s.s.................................................................................................................................................................... [ 99%]
..........
```

jcrist:
Looks like the new test might be a bit flaky on bigquery: https://github.com/ibis-project/ibis/actions/runs/10798011642/job/29950561913

cpcloud:
Blah, okay

PR #10060, opened 2024-09-09
feat(mssql): add lpad and rpad ops

IndexSeek:
## Description of changes

I noticed that MSSQL was the only backend missing lpad and rpad, so I wanted to add support for these. 

The LPAD operation was a bit tricky. I'm wondering if there might be a more straightforward way, but after several attempts, this seemed to handle all of the cases well.

### DIFF
--- a/ibis/backends/sql/compilers/mssql.py
+++ b/ibis/backends/sql/compilers/mssql.py
@@ -525,16 +525,22 @@ class MSSQLCompiler(SQLGlotCompiler):
         return arg.like(self.f.concat("%", end))
 
     def visit_LPad(self, op, *, arg, length, pad):
-        return self.f.left(
-            self.f.right(
+        return sge.Case(
+            ifs=[self.if_(length <= self.f.length(arg), arg)],
+            default=self.f.left(
                 self.f.concat(self.f.replicate(pad, length - self.f.length(arg)), arg),
-                length + self.f.length(arg),
+                length,
             ),
-            length,
         )
 
     def visit_RPad(self, op, *, arg, length, pad):
-        return self.f.left(self.f.concat(arg, self.f.replicate(pad, length)), length)
+        return sge.Case(
+            ifs=[self.if_(length <= self.f.length(arg), arg)],
+            default=self.f.left(
+                self.f.concat(arg, self.f.replicate(pad, length - self.f.length(arg))),
+                length,
+            ),
+        )
 
 
 compiler = MSSQLCompiler()
--- a/ibis/backends/sql/compilers/mssql.py
+++ b/ibis/backends/sql/compilers/mssql.py
@@ -95,7 +95,6 @@ class MSSQLCompiler(SQLGlotCompiler):
         ops.IntervalFloorDivide,
         ops.IsInf,
         ops.IsNan,
-        ops.LPad,
         ops.Levenshtein,
         ops.Map,
         ops.Median,
@@ -106,7 +105,6 @@ class MSSQLCompiler(SQLGlotCompiler):
         ops.RegexSearch,
         ops.RegexSplit,
         ops.RowID,
-        ops.RPad,
         ops.StringSplit,
         ops.StringToDate,
         ops.StringToTimestamp,
@@ -525,6 +523,18 @@ class MSSQLCompiler(SQLGlotCompiler):
 
     def visit_EndsWith(self, op, *, arg, end):
         return arg.like(self.f.concat("%", end))
+
+    def visit_LPad(self, op, *, arg, length, pad=" "):
+        return self.f.left(
+            self.f.right(
+                self.f.concat(self.f.replicate(pad, length - self.f.length(arg)), arg),
+                length + self.f.length(arg),
+            ),
+            length,
+        )
+
+    def visit_RPad(self, op, *, arg, length, pad=" "):
+        return self.f.left(self.f.concat(arg, self.f.replicate(pad, length)), length)
 
 
 compiler = MSSQLCompiler()
--- a/ibis/backends/sql/compilers/mssql.py
+++ b/ibis/backends/sql/compilers/mssql.py
@@ -524,7 +524,7 @@ class MSSQLCompiler(SQLGlotCompiler):
     def visit_EndsWith(self, op, *, arg, end):
         return arg.like(self.f.concat("%", end))
 
-    def visit_LPad(self, op, *, arg, length, pad=" "):
+    def visit_LPad(self, op, *, arg, length, pad):
         return self.f.left(
             self.f.right(
                 self.f.concat(self.f.replicate(pad, length - self.f.length(arg)), arg),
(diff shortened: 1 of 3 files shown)


IndexSeek:
Here was my initial implementation before it became what it is now:
```
    def visit_LPad(self, op, *, arg, length, pad=" "):
        return self.f.right(self.f.concat(self.f.replicate(pad, length), arg), length)
```

But it was giving me this result:
| string_col | index_col | LPad(string_col, 4, '-') |
|------------|------------|--------------------------|
| AbC\t      | 0          | AbC\t                    |
| \n123\n    | 1          | \n                       |
| abc, 123   | 2          | 123                      |
| 123        | 3          | -123                     |
| aBc        | 4          | -aBc                     |
| 🐍         | 5          | ---🐍                    |
| ÉéÈèêç     | 6          | Èèêç                     |

Where I was expecting this result:
| string_col | index_col | LPad(string_col, 4, '-') |
|------------|------------|--------------------------|
| AbC\t      | 0          | AbC\t                    |
| \n123\n    | 1          | \n123                    |
| abc, 123   | 2          | abc,                     |
| 123        | 3          | -123                     |
| aBc        | 4          | -aBc                     |
| 🐍         | 5          | ---🐍                    |
| ÉéÈèêç     | 6          | ÉéÈè                     |

ibis/backends/sql/compilers/mssql.py
```diff
@@ -526,5 +524,17 @@ def visit_StartsWith(self, op, *, arg, start):
     def visit_EndsWith(self, op, *, arg, end):
         return arg.like(self.f.concat("%", end))
 
+    def visit_LPad(self, op, *, arg, length, pad=" "):
```
cpcloud, line 527:
```suggestion
    def visit_LPad(self, op, *, arg, length, pad):
```

ibis/backends/sql/compilers/mssql.py
```diff
@@ -526,5 +524,17 @@ def visit_StartsWith(self, op, *, arg, start):
...
+        return self.f.left(
+            self.f.right(
+                self.f.concat(self.f.replicate(pad, length - self.f.length(arg)), arg),
+                length + self.f.length(arg),
+            ),
+            length,
+        )
+
+    def visit_RPad(self, op, *, arg, length, pad=" "):
```
cpcloud, line 536:
```suggestion
    def visit_RPad(self, op, *, arg, length, pad):
```

cpcloud (changes requested):
Thanks for working through this!

cpcloud:
@IndexSeek The pad functions are not consistent in their truncation behavior right now, so this PR should handle that at least for MS SQL.

The desired behavior is the Python behavior, which is to never truncate if the padded string would be shorter than existing string.

For example:

```pycon
>>> s = "abc, 123"
>>> s.ljust(4, '-')
'abc, 123'
```

cpcloud:
Added @gforsyth for review since he's been doing a bunch of work trying to vanquish (or least contain) the dark underbelly of accursed database strings.

IndexSeek:
> The desired behavior is the Python behavior, which is to never truncate if the padded string would be shorter than existing string.
> 
> For example:
> 
> ```
> >>> s = "abc, 123"
> >>> s.ljust(4, '-')
> 'abc, 123'
> ```

That makes total sense! I have an idea (that involves a case expression) of how we can make this work, and I'll try to tackle it this evening. Thanks for the initial review @cpcloud and the quick fix on the default value!

> Added @gforsyth for review since he's been doing a bunch of work trying to vanquish (or least contain) the dark underbelly of accursed database strings.

Thanks for looping @gforsyth in! Containing the strings is no easy feat! When I looked through the tests, I thought the following section would make a great candidate for addition to the new "Cursed Knowledge" page:

https://github.com/ibis-project/ibis/blob/3533827f3ab4fa63eb1915b0e8d250c3fe4dab90/ibis/backends/tests/test_string.py#L1143-L1152

cpcloud:
Oh, yeah, that page is just getting started 😂

gforsyth:
> Oh, yeah, that page is just getting started 😂

It will grow without bound

gforsyth:
I think I'll add a few more (technically redundant) test tables for padding and length (don't let this block you here @IndexSeek) so we can more explicitly map out "levels" of string support into things like:
1. Python style padding for "simple" strings (ascii-like)
2. Python style padding for basic unicode (accented letters, non-english alphabets)
3. Python style padding for graphemes, treating emoji as single characters

IndexSeek:
I have adjusted the functions now to support Python-style padding. I was able to reuse much of the previous logic with a slight change, essentially:
```
if the length in the l/rpad argument is greater than or equal to the length of the arg being padded:
    return the arg being padded
else:
    return the original expression
```

Since this behavior differs from the other backends, I added "mssql" to the notyet section for `reason="Python style padding, e.g. doesn't trim strings to pad-length"`. I hope this will work in the meantime!

gforsyth (approved):
This looks good to me, @IndexSeek !  As noted, I have some new string data that's a little more simple-ASCII for validating the simpler `lpad` and `rpad` cases and your implementation here matches Python's padding behavior, modulo issues with the unclear definition of what "length" is.

ibis/backends/sql/compilers/mssql.py
```diff
@@ -526,5 +524,23 @@ def visit_StartsWith(self, op, *, arg, start):
...
+        )
+
+    def visit_RPad(self, op, *, arg, length, pad):
+        return sge.Case(
+            ifs=[self.if_(length <= self.f.length(arg), arg)],
+            default=self.f.left(
+                self.f.concat(arg, self.f.replicate(pad, length - self.f.length(arg))),
+                length,
+            ),
+        )
```
cpcloud, line 543: These can both be simplified to `return self.if_(...)` since there's only a single truth case and a default, but that can be done in a follow-up.

cpcloud (approved):

IndexSeek:
> This looks good to me, @IndexSeek ! As noted, I have some new string data that's a little more simple-ASCII for validating the simpler `lpad` and `rpad` cases and your implementation here matches Python's padding behavior, modulo issues with the unclear definition of what "length" is.

Thank you for the review! "length" is definitely tricky here. 

SQL Server/T-SQL has the [DATALENGTH](https://learn.microsoft.com/en-us/sql/t-sql/functions/datalength-transact-sql) function, returning the number of bytes, where as it's [LEN](https://learn.microsoft.com/en-us/sql/t-sql/functions/len-transact-sql) function returns the number of characters. But it seems that some backends treat `LEN` like `DATALENGTH`. ¯\\__(ツ)_\_/¯

I am excited to review the new string data - it will be great to cover more cases!

ibis/backends/sql/compilers/mssql.py
IndexSeek: Thank you for guiding this simplification and approving it in #10085!

PR #10067, opened 2024-09-09
perf(backends): speed up most memtable existence checks

cpcloud:
Adds some slightly faster checks for memtable existence. At some point we may want to expand this to non-memtables, but for now these are only implemented for memtables.

### DIFF
--- a/ibis/backends/pyspark/__init__.py
+++ b/ibis/backends/pyspark/__init__.py
@@ -411,10 +411,17 @@ class Backend(SQLBackend, CanListCatalog, CanCreateDatabase):
             self._session.udf.register(f"unwrap_json_{typ.__name__}", unwrap_json(typ))
         self._session.udf.register("unwrap_json_float", unwrap_json_float)
 
+    def _in_memory_table_exists(self, name: str) -> bool:
+        sql = f"SHOW TABLES IN {self.current_database} LIKE '{name}'"
+        return bool(self._session.sql(sql).count())
+
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         schema = PySparkSchema.from_ibis(op.schema)
         df = self._session.createDataFrame(data=op.data.to_frame(), schema=schema)
         df.createTempView(op.name)
+
+    def _finalize_memtable(self, name: str) -> None:
+        self._session.catalog.dropTempView(name)
 
     @contextlib.contextmanager
     def _safe_raw_sql(self, query: str) -> Any:
--- a/ibis/backends/sqlite/__init__.py
+++ b/ibis/backends/sqlite/__init__.py
@@ -338,6 +338,18 @@ class Backend(SQLBackend, UrlFromPath):
 
         return sge.Create(kind="TABLE", this=target)
 
+    def _in_memory_table_exists(self, name: str) -> bool:
+        ident = sg.to_identifier(name, quoted=self.compiler.quoted)
+        query = sg.select(sge.convert(1)).from_(ident).limit(0).sql(self.dialect)
+        try:
+            with self.begin() as cur:
+                cur.execute(query)
+                cur.fetchall()
+        except sqlite3.OperationalError:
+            return False
+        else:
+            return True
+
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         table = sg.table(op.name, quoted=self.compiler.quoted, catalog="temp")
         create_stmt = self._generate_create_table(table, op.schema).sql(self.name)
--- a/ibis/backends/trino/__init__.py
+++ b/ibis/backends/trino/__init__.py
@@ -552,6 +552,21 @@ class Backend(SQLBackend, CanListCatalog, CanCreateDatabase, CanCreateSchema):
         df = TrinoPandasData.convert_table(df, schema)
         return df
 
+    def _in_memory_table_exists(self, name: str) -> bool:
+        ident = sg.to_identifier(name, quoted=self.compiler.quoted)
+        sql = sg.select(sge.convert(1)).from_(ident).limit(0).sql(self.dialect)
+
+        try:
+            with self.begin() as cur:
+                cur.execute(sql)
+                cur.fetchall()
+        except trino.exceptions.TrinoUserError as e:
+            if e.error_name == "TABLE_NOT_FOUND":
+                return False
+            raise
+        else:
+            return True
+
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
         schema = op.schema
         if null_columns := [col for col, dtype in schema.items() if dtype.is_null()]:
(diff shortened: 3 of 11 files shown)


ibis/backends/sqlite/__init__.py
```diff
@@ -338,6 +338,17 @@ def _generate_create_table(self, table: sge.Table, schema: sch.Schema):
 
         return sge.Create(kind="TABLE", this=target)
 
+    def _in_memory_table_exists(self, name: str) -> bool:
+        ident = sg.to_identifier(name, quoted=self.compiler.quoted)
+        query = sg.select(STAR).from_(ident).limit(0).sql(self.dialect)
```
cpcloud, line 343: I wish this could be a generic implementation, but the exception handling is different for every backend, and it didn't seem worth the trouble to make *that* an overridable method.

jcrist (approved):
Just a couple questions. Once #9695 lands we could also change the default implementation to

```python
try:
    self.table(name)
    return True
except TableNotFound:
    return False
```

ibis/backends/trino/__init__.py
```diff
@@ -557,6 +557,21 @@ def _fetch_from_cursor(self, cursor, schema: sch.Schema) -> pd.DataFrame:
         df = TrinoPandasData.convert_table(df, schema)
         return df
 
+    def _in_memory_table_exists(self, name: str) -> bool:
+        sql = sge.Describe(
+            this=sg.to_identifier(name, quoted=self.compiler.quoted)
+        ).sql(self.dialect)
```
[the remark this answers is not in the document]
cpcloud: `DESCRIBE` seems less likely to execute a query, but that's not based on anything empirical, so perhaps generalizing to `LIMIT 0` is fine, and then we can revisit on an as needed basis.

cpcloud:
Going to leave off postgres for now.

cpcloud:
Ok, I managed to audit all the backends except for Druid and Flink. Those can be tackled in a follow-up.

ibis/backends/sqlite/__init__.py
cpcloud: Looked into most of the remaining backends.

ibis/backends/pyspark/__init__.py
```diff
@@ -411,11 +411,18 @@ def _register_udfs(self, expr: ir.Expr) -> None:
             self._session.udf.register(f"unwrap_json_{typ.__name__}", unwrap_json(typ))
         self._session.udf.register("unwrap_json_float", unwrap_json_float)
 
+    def _in_memory_table_exists(self, name: str) -> bool:
+        sql = f"SHOW TABLES IN {self.current_database} LIKE '{name}'"
```
cpcloud, line 415: Somehow `self._session.catalog.tableExists` doesn't give the same answer as the code I have here, even when passing a fully quoted `default.$memtable` (the second `dbName` parameter is deprecated and the deprecation warning suggests passing the fully qualified name.)

ibis/backends/trino/__init__.py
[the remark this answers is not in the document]
cpcloud: Got rid of describe in favor of

```sql
SELECT 1
FROM t
LIMIT 0
```

which should be fast and return no data in the case of an existing table `t`.

cpcloud:
Clouds are passing:

```
…/ibis on  faster-memtable-existence-checks is 📦 v9.4.0 via 🐍 v3.10.14 via ❄️   impure (ibis-3.10.14-env)
❯ pytest -m 'snowflake or bigquery' -n 8 --dist loadgroup --snapshot-update -q
bringing up nodes...
x.....................................s....x.....x..s.........................x..........x........................x........xx....x..........................x..xx..................xx.............xx.........x [  5%]
.....................x....x...x.....x..............x........................x..............x...x..........................x.....xx.xx.........................x...........x.......x.....x................x.... [ 10%]
...........xx.....x.x........xx...x....x..x...x....x.x.........xx.xx.xxxxxxxxxx.x.x..x..x...xx..xxxx...x.x.xx..xxx..xx.x.....xxx.x.....xxx.xx.xxx.x..xxxx.x...x....x.x.xxx.xx.xx.....xx.x...xx.xx..xxx..xxx... [ 15%]
x.x.x.x.x..xx.x......................x.xx.........x.xx.x.s.ssssssssssssssssssss..xx.......x........................................x..................x..........x.x..............................x........... [ 20%]
x.................xx...................x..x....x......x......x...x.x.....x...........x.........x..s..x........x......x.....................x.................x......x............x.......x.x.x.xx....xxx.xs..x [ 25%]
..xx...x................................x........s...x.......ssssssssssssssssssssssssssssssssssssssssssssssssssss.ssssssssssssssssssssssssssss.sssssssssssssssssss........x.............s..............x...... [ 30%]
.....x.....................................x........x........................x............................................................x..x................x..........x..xx..xx........x...x............... [ 35%]
....................................................................x...........x..xx...x....x.x...........................................x................................x.......................x......... [ 40%]
.x....................x.....................x..x.....................................xxx...x........................................x...x...x.....x.x............x.x.........x.......xx.............x......... [ 45%]
....x......s........x............x............x...x.x...x.................x.............x.................x.......x......x...x.....x.x.........x.xxx....................x.x.........x.......................x. [ 50%]
.x.x...x..........x............x........................X......x........x.xxx.........x........x....x...x.......x.x.x.x...x......xxx.X........x..x.x..x.x.....xx.x....xxX..X....x........X....x...xxxxxx.xx.xx [ 56%]
x.xxxx.xx.xxxx.xxx.xxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx..x.x.xx.....................x............x..x...x......................x.x... [ 61%]
........x......x.x.x.....x...x..............................x...x.x..................x......x...............xx..x...................xx.x.........x.x.s..s...x.......x.........xxxXx.xs...s........x.x......... [ 66%]
..x...............x....x.x.xxx...xx.xx.x..x...x...x...x.x..........................s....ss.......x................................x..............x....x..xx.........x.......x.....x............x.xx.x.....x.xx [ 71%]
...x..xx...x....xxx..x.....x............x..xx.x.....x.x.........xx.x..x..............x.........x...............x........................................x..................................................... [ 76%]
.......x.....x...........................................xx....x.............xx................................................x..x..........x....................x.x..............x.......x..x.....xx.x..xxxx [ 81%]
xxxxx.....xx.xx.x.xx....x.xxxxx..x..x..xx...............x....x....xx.....x.............................x..................................................................................x................... [ 86%]
....x...................................................................................x..xx.....................................x........................................................................... [ 91%]
.............................................................................................................................................s........x..x.................................................... [ 96%]
....................................................................s...s.......x........................................s.....                                                                                [100%]
3330 passed, 138 skipped, 567 xfailed, 6 xpassed in 922.55s (0:15:22)
```

PR #10071, opened 2024-09-09
docs: add docs for versions 7.x and 8.x

gforsyth:
forgot to do this for the last two major versions -- here are all the rendered docs for all 7.x and 8.x releases


cpcloud:
1MM+ lines, I dig it.

PR #10073, opened 2024-09-09
depr(api): deprecate filtering/expression projection in `Table.__getitem__`

### LINKED ISSUE #10045: depr: deprecate `select` & `filter` behavior of `Table.__getitem__`
Currently the `Table.__getitem__` method does a lot of things:

1. Passing in a column name (or integer index) selects out the column (returning a Column).
2. Passing in multiple column names (or indices) does either a Select or Filter, column type dependent (returning a Table).
3. Passing in a slice does a limit with offset (returning a Table)
4. Passing in one or more boolean column expressions filters by those columns (returning a Table).
5. Passing in one or more other column expressions selects those columns (returning a Table).

Note that these cannot be mixed (there's no way to select columns and filter/slice in one call).

The last two can make for some confusing to read code, and potential footguns:

```python
# What does this do, given a, b, & c are all `ibis.Column` types?
t[a, b, c]
# - If they're _all_ boolean columns, this is a filter
# - If _any_ of them aren't boolean columns, this is a select

# What if they're all strings?
t["a", "b"]
# - If _all_ columns referenced are boolean columns, this is a filter
# - If _any_ of them aren't boolean columns, this is a select

# What if you only had one selection?
t["a"]  # returns `Column` "a"
t[t.a]  # Maybe filters if `a` is a boolean column, otherwise returns a table selecting `a`
t[["a"]]  # Maybe filters if `a` is a boolean column, otherwise returns a table selecting `a`



# - For `select`, these all are equivalent and result in a Table:
e1 = t.select("a")
e2 = t.select(t.a)
e3 = t.select(_.a)
assert e1.equals(e2) and e1.equals(e3)

# For `t[...]` they're not equivalent:
e4 = t["a"]  # This is a Column
e5 = t[t.a]  # This is a Table
e6 = t[_.a]  # This is a Table
assert e5.equals(e6)
assert e4.equals(t.a)
```

I'd advocate that behaviors 4 and 5 should be deprecated (along with the implicit filtering in 2) in favor of `filter` and `select` respectively. This would mean that `__getitem__` would be for selecting one or more columns from a table, or slicing a selection of rows, and nothing else. Notably none of these take in expressions as inputs (err, except for the `slice`). Instead we'd support:

```python
# 1. Selecting a single column out by name or index, returning a `Column`
t["a"]
t[1]

# 2. Selecting multiple columns out by passing a sequence to `__getitem__`, returning a `Table`
t[["a", "b"]]

# Note that we'd drop the implicit `tuple` version, the following is deprecated in favor of the former
t["a", "b"]

# 3. Slicing a table for a limit w/ offset, returning a `Table`
t[10:20]
```

See https://github.com/ibis-project/ibis/discussions/7259 for prior discussion.

jcrist:
This PR deprecates filtering and expression projection in `Table.__getitem__`. See the relevant issue (#10045) for reasoning.

Currently the warning says we'll remove in 10.0 - if we think that's too short a deprecation period I'm happy to bump to 11.0.

The magic syntax is _old_ and was used in quite a few places in the test suite and docs. I _think_ I got most of them, but won't be surprised if this PR fails CI initially due to a few lingering references.

Fixes #10045.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import itertools
 import operator
 import re
+import warnings
 from collections import deque
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from keyword import iskeyword
@@ -559,16 +560,16 @@ class Table(Expr, _FixedTextJupyterMixin):
             console_width=console_width,
         )
 
-    def __getitem__(self, what):
-        """Select items from a table expression.
-
-        This method implements square bracket syntax for table expressions,
-        including various forms of projection and filtering.
+    def __getitem__(self, what: str | int | slice | Sequence[str | int]):
+        """Select one or more columns or rows from a table expression.
 
         Parameters
         ----------
         what
-            Selection object. This can be a variety of types including strings, ints, lists.
+            What to select. Options are:
+            - A `str` column name or `int` column index to select a single column.
+            - A sequence of column names or indices to select multiple columns.
+            - A slice to select a subset of rows.
 
         Returns
         -------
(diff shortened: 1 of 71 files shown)


ibis/expr/types/relations.py
```diff
@@ -748,83 +641,56 @@ def __getitem__(self, what):
...
+        if util.all_of(args, str):
+            return self.select(args)
+
+        # Once this deprecation is removed, we'll want to error here instead.
+        warnings.warn(
+            "Selecting/filtering arbitrary expressions in `Table.__getitem__` is "
+            "deprecated and will be removed in version 10.0. Please use "
+            "`Table.select` or `Table.filter` instead.",
+            FutureWarning,
+        )
```
jcrist, line 690: The only relevant code is here, everything else is swapping `table[...]` for either `select` or `filter` in the tests or docs.

PR #10077, opened 2024-09-10
chore(mysql): port to MySQLdb instead of pymysql

cpcloud:
Replaces pymysql with mysqlclient, mostly out of frustration with bizarre GC behavior discovered during #10055. I think this is probably a breaking change due to some changes in how types are inferred for JSON, INET and UUID types.

BREAKING CHANGE: Ibis now uses the `MySQLdb` driver. You may need to install MySQL client libraries to **build** the extension.


ibis/backends/mysql/__init__.py
```diff
@@ -150,22 +144,20 @@ def do_connect(
             month : int32
 
         """
-        self.con = pymysql.connect(
+        self.con = MySQLdb.connect(
             user=user,
-            host=host,
+            host="127.0.0.1" if host == "localhost" else host,
```
cpcloud, line 149: localhost will try to use a socket, which is different from the behavior of pymysql, so I made localhost usage backwards compatible. There's a `unix_socket` argument if anyone really needs it.

gforsyth (approved):

PR #10079, opened 2024-09-10
depr(selectors): deprecate `c` and `r` selectors in favor of `cols` and `index`

### LINKED ISSUE #9924: refactor: a couple selectors rough edges
Noting a few (small) rough edges I have with our current selectors API. Definitely opinion based, and also small papercuts.

- We have a few selectors that have single-character names. To anyone not familiar with ibis's API, `s.c` is going to appear pretty opaque. I'd propose we rename these as:
  - `c -> cols`
  - `r -> range` (or `inds`/`indices` since it's for selecting by index/indices, see below)

  We could keep the single-character versions around as aliases if people want, but I'd prefer to have the canonical names be something more readable.

- Some of the names (`all`, `range` if we rename above) collide with builtins. This has required a hack in our doctest machinery to avoid clobbering `all` with `ibis.selectors.all`. R calls this one `everything` (as does `ibis-ml`), perhaps we should as well?

jcrist:
This PR renames the single character selectors to have more readable/meaningful names:

- `c` -> `cols`
- `r` -> `index`

The old names are deprecated and may be removed in a future release.

It also adds example docstrings for the selectors touched and a few other minor cleanups.

Fixes #9924.


cpcloud (approved):

ibis/selectors.py
```diff
@@ -639,15 +672,48 @@ def expand_names(self, table: ir.Table) -> frozenset[str]:
...
+class Indexable(Singleton):
     def __getitem__(self, key: str | int | slice | Iterable[int | str]):
         if isinstance(key, slice):
             key = Slice(key.start, key.stop, key.step)
-        return ColumnSlice(key)
+        return ColumnIndex(key)
+
+
+index = Indexable()
+"""Select columns by index.
```
jcrist, line 683: This is annoying.

`index` is an "attribute" object in `griffe`. As such, it expects a docstring below the object definition (as done here). This works, the docstring is properly rendered. However, during runtime inspection the user won't see this docstring since it's not actually attached to anything.

Moving the docstring to class level on `Indexable` means an ipython/jupyter user can do `ibis.selectors.index?` and see the docstring. `quartodoc`/`griffe` won't properly find the docstring though unless you set `dynamic: true` (maybe a bug in `quartodoc`/`griffe`?). If you set `dynamic: true` though, the `Examples` aren't rendered properly as markdown (this is probably a bug in `quartodoc`). Manually wrapping in the docstring works, but is a pain.

For now I've opted to do the easy thing and just get our docs rendering.

cpcloud: cc @machow for visibility

PR #10080, opened 2024-09-10
depr(api): deprecate `bool_val.negate()`/`-bool_val` in favor of `~bool_val`

### LINKED ISSUE #10074: depr: deprecate `-bool_val` in favor of `~bool_val`
While slogging through test updates for #10073 I was very surprised to learn that in ibis `-bool_val` does the same thing as the existing `~bool_val` (and is used for boolean negation a few places in our tests and docs). The existence of this isn't hurting anything exactly, but I found it a bit surprising (`-True` is not `False`, so why should `-bool_val` invert things?). Proposing we deprecate it in favor of `__invert__`/`~bool_val`.

jcrist:
Fixes #10074.

Just tossing this up for discussion since it was easy.


ibis/expr/types/logical.py
```diff
@@ -227,34 +228,16 @@ def __invert__(self) -> BooleanValue:
         │ NULL     │
         └──────────┘
         """
-        return self.negate()
+        return ops.Not(self).to_expr()
 
     def negate(self) -> BooleanValue:
```
jcrist, line 233: I'm not as against `bool_val.negate()` as I am against `-bool_val`. That said, since `BooleanValue` subclasses from `NumericValue`, having `.negate()` mean something other than `-(bool_val.cast("int64"))` would be a bit surprising IMO. Opted to deprecate both, in the future I think we should have this raise, same as `bool_val * -1` would.

cpcloud: Yeah, I guess there's a broader question here of whether we want to remove all of our Python-bool like behavior.

For example, we allow summing a column of bools, which is most backends is not allowed, and we end up casting.

If we did that, it _would_ go against our tendency to simulate Python builtin-type behavior where possible.

cpcloud: (In this case I agree with removing the logical negation meaning of `-` FWIW)

jcrist: > I guess there's a broader question here of whether we want to remove all of our Python-bool like behavior, effectively all the bool-is-a-subclass-of-int stuff.

I'm not against that. Currently it looks like we support numeric reductions on bools, but not treating them as integers in scalar operations. `bool_val * -1` currently errors:

```python
In [1]: import ibis                                                                                                   
                                                                                                                      
In [2]: t = ibis.table({"x": "bool"})                                                                                 
                                                                                                                      
In [3]: t.x * -1                                                                                                      
IbisTypeError                             Traceback (most recent call last)
...
IbisTypeError: Cannot compute precedence for `boolean` and `int8` types
```

Agree that's a larger question though.

cpcloud (approved):

PR #10083, opened 2024-09-10
feat(api): add `name` argument to `topk`/`value_counts`

### LINKED ISSUE #9948: feat: add `name` param to `.value_counts()` and `topk()`

With simple usages such as `t.age.value_counts()`, you get a predictable result column name of `age_count`. However, with more complex expressions, such as `t.age.fill_null(0).value_counts()`, it is hard to predict what the column name is. It is important to know the column name so you can do eg `t.age.value_counts().age_count.min()`.

This is also relevant for .topk(), I think they should have the same API.

Related issues/PRs
- https://github.com/ibis-project/ibis/pull/5418
- https://github.com/ibis-project/ibis/issues/8540
- https://github.com/ibis-project/ibis/pull/9163

Change the Column API to `def value_counts(*, name: str | Callable[[str], str] | None = None)`, with the same semantics as [Table.rename()](https://ibis-project.org/reference/expression-tables.html#ibis.expr.types.relations.Table.rename), except here we don't support Mappings, because we are only renaming one thing. Examples:

- "my_count"
- "{name}_n"
- lambda name: name.title() + " Counts"

Change the Table API to `def value_counts(*, name: str | Callable[[list[str]], str] | None = None)`, with SIMILAR semantics as above, except there are a few thorns since now our transformation is of the shape `Iterable[str] -> str`:

- "my_count"
- (I can't think of an intuitive way to support template strings)
- lambda names: str(names) + " counts"

Also would be fine limiting support here to only support string literals, and requiring users to do their lamba externally, that seems like it would always be possible.

The topk API would just have the existing `n` param left as is: `def topk(n: int = 10, *, name: ...)`

I can roll this change into https://github.com/ibis-project/ibis/pull/9163 if we want, or do it separately/after 10.0, since this would be non-breaking.

main

NA

### Code of Conduct

jcrist:
This adds a new `name` argument to `topk` & `value_counts`. Fixes #9948.

I think adding a `name` arg to these helpers makes sense. For other builtin aggregate methods you can chain a `.name("myname")` call on the end, but in these cases the return type is a `Table`. Adding an option to specify the name of the new column is nice for ux.

I opted to go for a simple `name: str | None` version, rather than supporting format strings/callables like `rename`. If a user wants to programmatically generate the name then they can do so themselves outside of this method. Keep it simple.

One open question is whether this argument should be keyword only. I went with positional or keyword for now, no strong thoughts.

Also added a docstring example for `topk` while I was at it.

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -2002,17 +2002,52 @@ class Column(Value, _FixedTextJupyterMixin):
     def topk(self, k: int, by: ir.Value | None = None) -> ir.Table:
         """Return a "top k" expression.
 
+        Computes a Table containing the top `k` values by a certain metric
+        (defaults to count).
+
         Parameters
         ----------
         k
-            Return this number of rows
+            The number of rows to return.
         by
-            An expression. Defaults to `count`.
+            The metric to compute "top" by. Defaults to `count`.
 
         Returns
         -------
         Table
-            A top-k expression
+            The top `k` values.
+
+        Examples
+        --------
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> t = ibis.examples.diamonds.fetch()
+
+        Compute the top 3 diamond colors by frequency:
+
+        >>> t.color.topk(3)
+        ┏━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┓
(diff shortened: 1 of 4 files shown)


NickCrews:
This looks perfect, thanks @jcrist !

PR #10087, opened 2024-09-11
test(polars): replace register with create_table

IndexSeek:
## Description of changes

Tackling a couple TODOs to replace the `con.register` and `con._add_table` methods for the Polars tests.


jcrist (approved):
Thanks!

PR #9970, opened 2024-08-30
ci(backends): run backend doctests in CI

cpcloud:
Run backend doctests in CI during the usual backend test suite.

PR #10088, opened 2024-09-11
chore: backend doctests

gmcrocetti:


gmcrocetti:
> Thanks for these changes! I've applied them to #9970!

o/. I was about to finish it first thing in the morning but glad you took it over.

PR #10090, opened 2024-09-11
docs(dropdowns): make dropdowns scrollable and easier to see in navigation bar

ramlakhanmadheshiya:
## Description of changes

Dropdowns and tools in navigation bar were not appearing properly, I have added some custom `CSS` to fix it! 

| Left long dropdowns | Right tools dropdown |
|--------|--------|
| image | image | 
| After this PR | After this PR |
| *added a scroll bar* image | image |


cpcloud:
This seems like something that should be fixed upstream in Quarto, or perhaps even one feature and one bugfix.

- Feature: scrolling dropdowns
- Fix: avoid rendering out of the left or right viewport

ramlakhanmadheshiya:
We could use the part of CSS I added until Quarto fixes it! I will open an issue for this in Quarto!

cpcloud (approved):
Let's give this a shot! We can always revert if people complain or something gets weird with CSS (as it usually does), but I suspect this is a net improvement.

Much appreciated @ramlakhanmadheshiya!

PR #10094, opened 2024-09-11
fix(examples): use memtables for examples to allow backends that do not support temporary tables to support examples

cpcloud:
Fixes an import bug that can show up when calling register on the datafusion backend without having imported `pyarrow.dataset` before calling `register`, as well as moving to memtables for examples to bypass `temp=True` shenanigans.

The import bug wasn't caught due to importing `pyarrow.dataset` at the top of the test module.

### DIFF
--- a/ibis/examples/__init__.py
+++ b/ibis/examples/__init__.py
@@ -78,11 +78,7 @@ class Example(Concrete):
                     if pyarrow.types.is_null(field.type):
                         table = table.set_column(i, field.name, table[i].cast("string"))
 
-            # TODO: It should be possible to avoid this memtable call, once all
-            # backends support passing a `pyarrow.Table` to `create_table`
-            # directly.
-            obj = ibis.memtable(table)
-            return backend.create_table(table_name, obj, temp=True, overwrite=True)
+            return ibis.memtable(table, name=table_name)
 
 
 _FETCH_DOCSTRING_TEMPLATE = """\
--- a/ibis/examples/__init__.py
+++ b/ibis/examples/__init__.py
@@ -80,7 +80,7 @@ class Example(Concrete):
 
             obj = ibis.memtable(table, name=table_name)
             backend._register_in_memory_tables(obj)
-            return backend.table(table_name)
+            return backend.table(table_name).cache()
 
 
 _FETCH_DOCSTRING_TEMPLATE = """\
--- a/ibis/examples/__init__.py
+++ b/ibis/examples/__init__.py
@@ -78,7 +78,9 @@ class Example(Concrete):
                     if pyarrow.types.is_null(field.type):
                         table = table.set_column(i, field.name, table[i].cast("string"))
 
-            return ibis.memtable(table, name=table_name)
+            obj = ibis.memtable(table, name=table_name)
+            backend._register_in_memory_tables(obj)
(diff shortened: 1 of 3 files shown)


ibis/examples/__init__.py
```diff
@@ -78,11 +78,7 @@ def fetch(
                     if pyarrow.types.is_null(field.type):
                         table = table.set_column(i, field.name, table[i].cast("string"))
 
-            # TODO: It should be possible to avoid this memtable call, once all
-            # backends support passing a `pyarrow.Table` to `create_table`
-            # directly.
-            obj = ibis.memtable(table)
-            return backend.create_table(table_name, obj, temp=True, overwrite=True)
+            return ibis.memtable(table, name=table_name)
```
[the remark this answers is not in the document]
cpcloud: Hm, true. 🤔

ibis/examples/__init__.py
```diff
@@ -78,11 +78,9 @@ def fetch(
...
                         table = table.set_column(i, field.name, table[i].cast("string"))
 
-            # TODO: It should be possible to avoid this memtable call, once all
-            # backends support passing a `pyarrow.Table` to `create_table`
-            # directly.
-            obj = ibis.memtable(table)
-            return backend.create_table(table_name, obj, temp=True, overwrite=True)
+            obj = ibis.memtable(table, name=table_name)
+            backend._register_in_memory_tables(obj)
+            return backend.table(table_name)
```
jcrist, line 83: Won't this drop `obj` (the memtable) at the end of the function, causing it to be collected and the table removed?

cpcloud: ARG, yes

cpcloud: What if I immediately call `.cache()`?

cpcloud: I wonder if we should just allow `temp=True` datafusion. Seems like a lot of effort just to disable a behavior that wasn't so bad.

cpcloud: `memtable` seems like currently the only sane way to do that.

cpcloud: I'll push up the `cache` "solution" and we can discuss further.

cpcloud:
An alternative solution might be to invert the dependency here and allow backends to control how they produce an example table.

cpcloud:
Yeah, that makes sense.

cpcloud:
Closing this for now.

ibis/examples/__init__.py
ncclementi: > That might have weird consequences for the ddl work @ncclementi is doing? That said, I do think that an engine like datafusion (where all tables are kinda temporary anyway) ignoring temp completely does make sense in isolation.

I know this is closed, but for completion of the discussion this wouldn't affect the work itself. My only concern would be if someone creates a "temporary" and then tries to do a `con.ddl.list_temp_tables()` it will get a not implemented error, because the `create_table` swallows the `TEMPORARY` and such table shows a `BASE_TABLE` in the information_schema, meaning that it will and is listed when `con.ddl.list_tables()` is invoked. 

That being said, there is an upstream issue to raise when `TEMPORARY` is used in the SQL which will cause the same problem down the road.

PR #10098, opened 2024-09-11
ci: test examples

cpcloud:
We were not testing examples in CI because we did not install the `examples` extra. Now we are.

### DIFF
--- a/ibis/backends/datafusion/__init__.py
+++ b/ibis/backends/datafusion/__init__.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any
 
 import datafusion as df
 import pyarrow as pa
+import pyarrow.dataset as ds
 import pyarrow_hotfix  # noqa: F401
 import sqlglot as sg
 import sqlglot.expressions as sge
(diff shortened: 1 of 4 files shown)


ibis/backends/datafusion/__init__.py
```diff
@@ -372,7 +373,7 @@ def _register(
             self.con.deregister_table(table_name)
             self.con.register_record_batches(table_name, [[source]])
             return self.table(table_name)
-        elif isinstance(source, pa.dataset.Dataset):
+        elif isinstance(source, ds.Dataset):
```
ncclementi, line 376: This seems to be unrelated to including examples in the tests suit, just making sure you are ok with that.

ibis/backends/datafusion/tests/test_register.py
```diff
@@ -45,6 +44,8 @@ def test_register_batches(conn):
 
 
 def test_register_dataset(conn):
+    import pyarrow.dataset as ds
```
ncclementi, line 47: same that the comment above. If we are good with these nits getting in here, all good.

ncclementi (approved):
LGTM, just two small comments to check weather those changes belong here or not.

ibis/backends/datafusion/__init__.py
cpcloud: These are definitely related. They aren't hit unless the examples are run.

ibis/backends/datafusion/tests/test_register.py
cpcloud: Definitely related, for the same reason.

PR #10106, opened 2024-09-12
docs: avoid needing to render API docs for any preview/render invocation

cpcloud:
Because we include a generated file, this forces developers to render API docs, or delete the file that references the generated file. This PR fixes that by deleting the file using the include, and referencing the included file directly in `_quarto.yml`


cpcloud:
Pages and dropdown check out:

image
image

gforsyth (approved):
SHIP IT  (and thank you for figuring this out)

PR #10108, opened 2024-09-13
refactor(backends): replace register usage with read

IndexSeek:
## Description of changes

Replacing a few additional existing "register" methods from the 9.1 deprecation.

### DIFF
--- a/ibis/backends/__init__.py
+++ b/ibis/backends/__init__.py
@@ -1425,10 +1425,15 @@ def connect(resource: Path | str, **kwargs: Any) -> BaseBackend:
             return ibis.duckdb.connect(path, **kwargs)
         elif path.endswith((".sqlite", ".db")):
             return ibis.sqlite.connect(path, **kwargs)
-        elif path.endswith((".parquet", ".csv", ".csv.gz")):
-            # Load parquet/csv/csv.gz files with duckdb by default
+        elif path.endswith((".csv", ".csv.gz")):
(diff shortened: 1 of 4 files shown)


IndexSeek:
I found a few more instances of `con.register` floating around in the DuckDB backend in the following methods, but I wasn't sure of the best way to address these. 

https://github.com/ibis-project/ibis/blob/34c465cde2429147beb018fb5fb5b9a0308c6c2b/ibis/backends/duckdb/__init__.py#L900
https://github.com/ibis-project/ibis/blob/34c465cde2429147beb018fb5fb5b9a0308c6c2b/ibis/backends/duckdb/__init__.py#L983
https://github.com/ibis-project/ibis/blob/34c465cde2429147beb018fb5fb5b9a0308c6c2b/ibis/backends/duckdb/__init__.py#L1619

cpcloud:
We relly should deprecate that usage of `connect('path/to/file')`.

IndexSeek:
> Thanks!
> 
> Re the other uses of `register`, those are calls to the duckdb client `register` method, not the `register` method in ibis - they're fine to leave around :).

You're welcome! Thank you for the review and explanation as well. We definitely want to keep those around! 

> We really should deprecate that usage of `connect('path/to/file')`.

I agree. Admittedly, I didn't know it was possible to use `ibis.connect` in that way. It is pretty neat, but I can't anticipate it being used very often as I suspect it requires something like `list_tables` to get the table's name created from the file after the fact.

PR #10117, opened 2024-09-13
docs(code_of_conduct): update committee members and reporting email

gforsyth:
## Description of changes

The instructions for reporting CoC violations are out-of-date, as are the
members of the committee to review any violations.

I've updated the page so the email goes to a private Zulip channel with only the
committee members in it, and the names of those members.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -23,15 +23,11 @@ Ibis is governed by the
 ## Reporting and Enforcement Violations
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported by contacting the Ibis Code of Conduct committee at
-ibis-conduct@googlegroups.com. You can also report
+(privately) reported by contacting the [Ibis Code of Conduct
+committee](mailto:ibis-conduct.5e76f3596b52d3bb7ca89a83e6cdbb54.show-sender@streams.zulipchat.com).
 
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -23,15 +23,16 @@ Ibis is governed by the
 ## Reporting and Enforcement Violations
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported by contacting the Ibis Code of Conduct committee at
-ibis-conduct@googlegroups.com. You can also report
+(privately) reported by contacting the [Ibis Code of Conduct
+committee](mailto:ibis-conduct.5e76f3596b52d3bb7ca89a83e6cdbb54.show-sender@streams.zulipchat.com).
 
 The committee currently consists of:
```
lostmygithubaccount, line 29: should this list be removed in favor of a link to https://github.com/ibis-project/governance/blob/main/governance.md#steering-committee-members so it doesn't need to be maintained in both places?

gforsyth: Yeah, let's do that for now.  There's a future where those aren't the same set, but we aren't in it yet.

gforsyth: Done

cpcloud (approved):

jcrist (approved):

PR #10119, opened 2024-09-13
refactor(bigquery): remove unnecessary and misspelled bigquery string find implementation

tswast:
Typo `visti` -> `visit`


gforsyth (commented):
wait, what?  hmm, I have to take a look at our tests

gforsyth:
Thanks for flagging (and fixing) this @tswast !  I'm going to add `find` to the updated string tests and see what shakes out.

cpcloud:
I think this means:

1. We're not testing all the error modes
2. The base implementation works fine (since that's what'll be called) and we can probably remove the BigQuery-specific implementation?

cpcloud:
I think we can just delete the bigquery implementation since it's actually more limited than the base implementation and the base implementation has been successfully passing for many many CI runs now.

cpcloud (approved):

PR #10121, opened 2024-09-13
feat(api): add `distinct` option to `collect`

jcrist:
This adds a new `distinct` option to `collect` (defaulting to `False`). When True, only distinct elements are collected.

Depending on the backend calling `x.collect(distinct=True)` may be more efficient than `x.collect().unique()`, and has the added benefit of not disrupting array ordering for backends where `unique` may not respect ordering.

### DIFF
--- a/ibis/backends/sql/compilers/bigquery/__init__.py
+++ b/ibis/backends/sql/compilers/bigquery/__init__.py
@@ -479,16 +479,24 @@ class BigQueryCompiler(SQLGlotCompiler):
             return self.f.parse_timestamp(format_str, arg, timezone)
         return self.f.parse_datetime(format_str, arg)
 
-    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null):
-        if where is not None and include_null:
-            raise com.UnsupportedOperationError(
-                "Combining `include_null=True` and `where` is not supported "
-                "by bigquery"
-            )
-        out = self.agg.array_agg(arg, where=where, order_by=order_by)
+    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null, distinct):
+        if where is not None:
+            if include_null:
+                raise com.UnsupportedOperationError(
+                    "Combining `include_null=True` and `where` is not supported by bigquery"
+                )
+            if distinct:
+                raise com.UnsupportedOperationError(
+                    "Combining `distinct=True` and `where` is not supported by bigquery"
+                )
+            arg = compiler.if_(where, arg, NULL)
+        if distinct:
+            arg = sge.Distinct(expressions=[arg])
+        if order_by:
+            arg = sge.Order(this=arg, expressions=order_by)
         if not include_null:
-            out = sge.IgnoreNulls(this=out)
-        return out
+            arg = sge.IgnoreNulls(this=arg)
+        return self.f.array_agg(arg)
 
     def _neg_idx_to_pos(self, arg, idx):
         return self.if_(idx < 0, self.f.array_length(arg) + idx, idx)
--- a/ibis/backends/sql/compilers/datafusion.py
+++ b/ibis/backends/sql/compilers/datafusion.py
@@ -327,7 +327,11 @@ class DataFusionCompiler(SQLGlotCompiler):
     def visit_ArrayPosition(self, op, *, arg, other):
         return self.f.coalesce(self.f.array_position(arg, other), 0)
 
-    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null):
+    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null, distinct):
+        if distinct:
+            raise com.UnsupportedOperationError(
+                "`collect` with `distinct=True` is not supported"
+            )
         if not include_null:
             cond = arg.is_(sg.not_(NULL, copy=False))
             where = cond if where is None else sge.And(this=cond, expression=where)
--- a/ibis/backends/sql/compilers/flink.py
+++ b/ibis/backends/sql/compilers/flink.py
@@ -572,20 +572,24 @@ class FlinkCompiler(SQLGlotCompiler):
     def visit_StructColumn(self, op, *, names, values):
         return self.cast(sge.Struct(expressions=list(values)), op.dtype)
 
-    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null):
+    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null, distinct):
         if order_by:
             raise com.UnsupportedOperationError(
                 "ordering of order-sensitive aggregations via `order_by` is "
                 "not supported for this backend"
             )
-        # the only way to get filtering *and* respecting nulls is to use
-        # `FILTER` syntax, but it's broken in various ways for other aggregates
-        out = self.f.array_agg(arg)
         if not include_null:
             cond = arg.is_(sg.not_(NULL, copy=False))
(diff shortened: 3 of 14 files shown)


ibis/backends/tests/test_aggregation.py
```diff
@@ -1301,67 +1302,75 @@ def test_group_concat_ordered(alltypes, df, filtered):
...
-        alltypes.filter(_.bigint_col == 10)
-        .id.cast("str")
-        .collect(where=ibis_cond, order_by=_.id.desc())
-        .execute()
-    )
-    expected = list(
-        df.id[(df.bigint_col == 10) & pd_cond].sort_values(ascending=False).astype(str)
-    )
-    assert result == expected
+def gen_test_collect_marks(distinct, filtered, ordered, include_null):
```
jcrist, line 1305: The other option was to add a bunch of `strict=False` checks everywhere, making the test less strict. In the long run I might want to add a new pytest shorthand for handling parametrizing a test with a cross-product of parameters with markers for certain parameter combos, but for now breaking the mark generation into a utility function didn't seem too bad.

ibis/backends/sql/compilers/flink.py
```diff
@@ -572,20 +572,24 @@ def visit_MapMerge(self, op: ops.MapMerge, *, left, right):
...
             cond = arg.is_(sg.not_(NULL, copy=False))
             where = cond if where is None else sge.And(this=cond, expression=where)
+        out = self.f.array_agg(arg)
         if where is not None:
             out = sge.Filter(this=out, expression=sge.Where(this=where))
+        if distinct:
+            # TODO: Flink supposedly supports `ARRAY_AGG(DISTINCT ...)`, but it
+            # doesn't work with filtering (either `include_null=False` or
+            # additional filtering). Their `array_distinct` function does maintain
+            # ordering though, so we can use it here.
```
jcrist, line 591: Flink does some really weird stuff here. One parameter combo somehow results in filtering out all but _two_ null values for unknown reasons.

ibis/expr/operations/reductions.py
```diff
@@ -376,6 +376,15 @@ class ArrayCollect(Filterable, Reduction):
     arg: Column
     order_by: VarTuple[SortKey] = ()
     include_null: bool = False
+    distinct: bool = False
+
+    def __init__(self, arg, order_by, distinct, **kwargs):
+        if distinct and order_by and [arg] != [key.expr for key in order_by]:
+            raise ValidationError(
+                "`collect` with `order_by` and `distinct=True` and may only "
+                "order by the collected column"
```
jcrist, line 385: No backend supports this, so I opted to error in our IR rather than at runtime.

jcrist:
Cloud tests are passing fine, should be ready for review.

ibis/expr/tests/test_reductions.py
```diff
@@ -161,3 +162,22 @@ def test_ordered_aggregations_no_order(method):
...
     assert q1.equals(q2)
     assert q1.equals(q3)
+
+
+def test_collect_distinct():
+    t = ibis.table({"a": "string", "b": "int", "c": "int"}, name="t")
+    # Fine
+    t.a.collect(distinct=True)
+    t.a.collect(distinct=True, order_by=t.a.desc())
+    (t.a + 1).collect(distinct=True, order_by=(t.a + 1).desc())
```
cpcloud, line 172: Are these necessary if we're testing them in an execution test?

ibis/backends/sql/compilers/snowflake.py
```diff
@@ -452,25 +452,36 @@ def visit_TimestampFromUNIX(self, op, *, arg, unit):
...
 
-    def _array_collect(self, *, arg, where, order_by, include_null):
+    def _array_collect(self, *, arg, where, order_by, include_null, distinct=False):
         if include_null:
             raise com.UnsupportedOperationError(
                 "`include_null=True` is not supported by the snowflake backend"
             )
+        if where is not None and distinct:
+            raise com.UnsupportedOperationError(
+                "Combining `distinct=True` and `where` is not supported by snowflake"
```
cpcloud, line 462:
```suggestion
                "Combining `distinct=True` and `where` is not supported by Snowflake"
```

Only because it's a proper noun and not a generic one :)

Alternatively, we can leave this for a follow up and audit all the backends for proper noun capitalization.

ibis/backends/sql/compilers/bigquery/__init__.py
```diff
@@ -479,16 +479,24 @@ def visit_StringToTimestamp(self, op, *, arg, format_str):
...
-            raise com.UnsupportedOperationError(
-                "Combining `include_null=True` and `where` is not supported "
-                "by bigquery"
-            )
-        out = self.agg.array_agg(arg, where=where, order_by=order_by)
+    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null, distinct):
+        if where is not None:
+            if include_null:
+                raise com.UnsupportedOperationError(
+                    "Combining `include_null=True` and `where` is not supported by bigquery"
```
cpcloud, line 486:
```suggestion
                    "Combining `include_null=True` and `where` is not supported by BigQuery"
```

ibis/backends/sql/compilers/bigquery/__init__.py
```diff
@@ -479,16 +479,24 @@ def visit_StringToTimestamp(self, op, *, arg, format_str):
...
+                )
+            if distinct:
+                raise com.UnsupportedOperationError(
+                    "Combining `distinct=True` and `where` is not supported by bigquery"
```
cpcloud, line 490:
```suggestion
                    "Combining `distinct=True` and `where` is not supported by BigQuery"
```

ibis/backends/sql/compilers/datafusion.py
```diff
@@ -327,7 +327,11 @@ def visit_ArrayRepeat(self, op, *, arg, times):
     def visit_ArrayPosition(self, op, *, arg, other):
         return self.f.coalesce(self.f.array_position(arg, other), 0)
 
-    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null):
+    def visit_ArrayCollect(self, op, *, arg, where, order_by, include_null, distinct):
+        if distinct:
+            raise com.UnsupportedOperationError(
+                "`collect` with `distinct=True` is not supported"
```
cpcloud, line 333:
```suggestion
                "`collect` with `distinct=True` is not supported by DataFusion"
```

ibis/expr/operations/reductions.py
cpcloud: Seems fine, but generally we should avoid overriding `__init__` in most cases. For more complicated validations like this one, overriding `__init__` is often the only way to reasonably spell the conditon.

cpcloud (approved):
Nothing blocking, thanks for adding this!

ibis/expr/operations/reductions.py
jcrist: Yeah, that's why I did it here.

ibis/expr/tests/test_reductions.py
jcrist: This test is for checking the validation check on `order_by`, I want to have decent coverage here in a single test. We do hit one of these cases in the execution test, but that's more for testing execution is correct, not the expression validation.

ibis/backends/sql/compilers/snowflake.py
jcrist: I know there's more of these in the codebase, I'll leave this as a follow up.

PR #10122, opened 2024-09-13
fix(joins): allow chaining positional and cross joins

gforsyth:
A user on Zulip
https://ibis-project.zulipchat.com/#narrow/stream/405265-tech-support/topic/Creating.20a.20new.20column.20from.20a.20numpy.20array.3F/near/470013765
reported an error about a missing `predicate` argument when trying to
chain positional joins.

This allows chaining the join types that don't have
predicates (`positional` and `cross`).


jcrist (approved):
Nice fix.

PR #10132, opened 2024-09-15
chore(deps): remove the `pandas` extra

cpcloud:
Removes the `pandas` extra, which I forgot to remove when removing the pandas backend.


cpcloud:
Going to use this PR to clean up useless mentions of the pandas backend in comments and other places as well.

PR #10149, opened 2024-09-17
refactor(api): remove schema

ncclementi:
BREAKING CHANGE: Removed hierarchical usage of schema. 

Ibis uses the following naming conventions:
- schema: a mapping of column names to datatypes
- database: a collection of tables
- catalog: a collection of databases

### DIFF
--- a/ibis/backends/duckdb/tests/test_client.py
+++ b/ibis/backends/duckdb/tests/test_client.py
@@ -82,6 +82,11 @@ def test_cross_db(tmpdir):
     assert t1_from_con2.schema() == t2.schema()
     assert t1_from_con2.execute().equals(t2.execute())
 
+    foo_t1_from_con2 = con2.table("t1", database="test1.foo")
+
+    assert foo_t1_from_con2.schema() == t2.schema()
+    assert foo_t1_from_con2.execute().equals(t2.execute())
+
 
 def test_attach_detach(tmpdir):
     path1 = str(tmpdir.join("test1.ddb"))
--- a/ibis/backends/duckdb/tests/test_client.py
+++ b/ibis/backends/duckdb/tests/test_client.py
@@ -78,7 +78,7 @@ def test_cross_db(tmpdir):
 
     con2.attach(path1, name="test1", read_only=True)
 
-    t1_from_con2 = con2.table("t1", database="test1")
+    t1_from_con2 = con2.table("t1", database="test1.main")
     assert t1_from_con2.schema() == t2.schema()
     assert t1_from_con2.execute().equals(t2.execute())
 
--- a/ibis/backends/duckdb/tests/test_client.py
+++ b/ibis/backends/duckdb/tests/test_client.py
@@ -78,15 +78,9 @@ def test_cross_db(tmpdir):
 
     con2.attach(path1, name="test1", read_only=True)
 
-    with pytest.warns(FutureWarning):
-        t1_from_con2 = con2.table("t1", schema="main", database="test1")
+    t1_from_con2 = con2.table("t1", database="test1")
     assert t1_from_con2.schema() == t2.schema()
     assert t1_from_con2.execute().equals(t2.execute())
-
-    with pytest.warns(FutureWarning):
-        foo_t1_from_con2 = con2.table("t1", schema="foo", database="test1")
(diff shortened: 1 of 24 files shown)


ncclementi:
I'm not sure I understand how the `test_signature.py` works, because I clearly broke all of them, but is failing due to an XPASS, any help on that front would be appreciated.

cpcloud:
If something is xpassing, generally the approach is to delete the marker that's causing the problem.

gforsyth:
the signature check is to enforce (eventually) that all the backends call all positional arguments the same thing, etc (that's why currently, almost all of them are xfailed)

When you removed the schema kwarg, it made some of the backend-specific table methods match the method definition from BaseBackend.  
also, keyword argument names also have to match, unless those arguments are keyword-only arguments

removing entries from those xfails is a good thing, though

ibis/backends/oracle/tests/test_client.py
```diff
@@ -67,14 +67,11 @@ def stats_one_way_anova(x, y, value: str) -> int:
...
-def test_list_tables_schema_warning_refactor(con):
+def test_list_tables(con):
     assert con.list_tables()
 
     with pytest.raises(exc.IbisInputError):
-        con.list_tables(database="not none", schema="not none")
-
-    with pytest.warns(FutureWarning):
-        assert con.list_tables(schema="SYS", like="EXU8OPT") == ["EXU8OPT"]
+        con.list_tables(database="not none")
```
ncclementi, line 74: This seems to be broken, but I can't run the Oracle test suite locally, I'll need some help debugging this one.

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -78,16 +78,10 @@ def test_cross_db(tmpdir):
 
     con2.attach(path1, name="test1", read_only=True)
 
-    with pytest.warns(FutureWarning):
-        t1_from_con2 = con2.table("t1", schema="main", database="test1")
+    t1_from_con2 = con2.table("t1", database="test1")
```
gforsyth, line 81:
```suggestion
    t1_from_con2 = con2.table("t1", database="test1.main")
```

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -78,16 +78,10 @@ def test_cross_db(tmpdir):
...
     assert t1_from_con2.schema() == t2.schema()
     assert t1_from_con2.execute().equals(t2.execute())
 
-    with pytest.warns(FutureWarning):
-        foo_t1_from_con2 = con2.table("t1", schema="foo", database="test1")
```
gforsyth, line 87: This shouldn't be removed, but refactored to use either a dotted path or a tuple of path elements, like `"test1.foo"`

gforsyth (commented):
Looks good overall,  i think we lost one bit of testing that we shouldn't have.

ibis/backends/trino/tests/test_client.py
```diff
@@ -185,14 +185,6 @@ def test_list_tables_schema_warning_refactor(con):
...
     assert con.list_tables()
 
-    # Error if user mixes tuple inputs and string inputs for database and schema
-    with pytest.raises(FutureWarning):
-        with pytest.raises(exc.IbisInputError):
-            con.list_tables(database=("tuple", "ohstuff"), schema="str")
-
-    with pytest.warns(FutureWarning):
-        assert con.list_tables(database="tpch", schema="sf1") == tpch_tables
-
```
ncclementi, line 195: @gforsyth Is this a similar case than the one pointed out above, about missing tests. Do we need to adapt this ones to pass `database="tpch.sf1"` I can't run the trino one locally to try them out

ibis/backends/duckdb/tests/test_client.py
ncclementi: fixed, I added it back using a dotted path.

ibis/backends/trino/tests/test_client.py
gforsyth: Yeah, no issues here because it's no longer possible to "overspecify" a path

gforsyth (approved):
Good riddance!

gforsyth:
I've got to check to make sure the commit message is going to be formatted correctly, then we can merge

PR #10151, opened 2024-09-17
feat(polars): allow user to specify "engine" kwarg

deepyaman:
## Description of changes

Modeled after support for "streaming" kwarg, but I have a couple questions:

1. I don't see any tests for "streaming", so I didn't add any for "engine"; is this fine?
2. I didn't see any attempt to support older Polars versions that didn't accept "streaming", so I didn't try to do that for "engine", either. However, Polars is technically stable now; should we not force people to get latest?

### DIFF
--- a/ibis/backends/polars/tests/test_client.py
+++ b/ibis/backends/polars/tests/test_client.py
@@ -60,3 +60,23 @@ def test_memtable_polars_types(con):
     res = con.to_polars((t.x + t.y + t.z).name("test"))
     sol = (df["x"] + df["y"] + df["z"]).rename("test")
     pl.testing.assert_series_equal(res, sol)
+
+
+@pytest.mark.parametrize("to_method", ["to_pyarrow", "to_polars"])
+def test_streaming(con, mocker, to_method):
+    t = con.table("functional_alltypes")
+    to_method_spy = mocker.spy(con, to_method)
+    mocked_collect = mocker.patch("polars.LazyFrame.collect")
+    getattr(con, to_method)(t, streaming=True)
+    to_method_spy.assert_called_once_with(t, streaming=True)
+    mocked_collect.assert_called_once_with(streaming=True, engine="cpu")
+
+
+@pytest.mark.parametrize("to_method", ["to_pyarrow", "to_polars"])
+def test_engine(con, mocker, to_method):
+    t = con.table("functional_alltypes")
+    to_method_spy = mocker.spy(con, to_method)
+    mocked_collect = mocker.patch("polars.LazyFrame.collect")
+    getattr(con, to_method)(t, engine="gpu")
+    to_method_spy.assert_called_once_with(t, engine="gpu")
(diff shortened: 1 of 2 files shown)


cpcloud:
We don't have tests for streaming, but we have a blog post that uses it, so it has had coverage.

It would be good to verify this behavior does **something** in the test suite. It can be a smoke test.

jcrist:
Maybe to avoid requiring a GPU for support we could do a mocktest just to ensure we're plumbing the `engine` kwarg through properly?

cpcloud:
Yeah, using `mocker.spy` would be sufficient IMO.

deepyaman:
Yep, sounds good, I was planning to mock it (and maybe actually go ahead and do the same for making sure `streaming` gets passed, too).

ibis/backends/polars/tests/test_client.py
```diff
@@ -60,3 +60,23 @@ def test_memtable_polars_types(con):
...
+
+
+@pytest.mark.parametrize("to_method", ["to_pyarrow", "to_polars"])
+def test_streaming(con, mocker, to_method):
+    t = con.table("functional_alltypes")
+    to_method_spy = mocker.spy(con, to_method)
+    mocked_collect = mocker.patch("polars.LazyFrame.collect")
+    getattr(con, to_method)(t, streaming=True)
+    to_method_spy.assert_called_once_with(t, streaming=True)
+    mocked_collect.assert_called_once_with(streaming=True, engine="cpu")
```
jcrist, line 72: No need to `spy` on the `to_{method}` here when you're directly calling the method in this test. The `patch` on `collect` is sufficient.

```suggestion
    mocked_collect = mocker.patch("polars.LazyFrame.collect")
    getattr(con, to_method)(t, streaming=True)
    mocked_collect.assert_called_once_with(streaming=True, engine="cpu")
```

ibis/backends/polars/tests/test_client.py
```diff
@@ -60,3 +60,23 @@ def test_memtable_polars_types(con):
...
+
+
+@pytest.mark.parametrize("to_method", ["to_pyarrow", "to_polars"])
+def test_engine(con, mocker, to_method):
+    t = con.table("functional_alltypes")
+    to_method_spy = mocker.spy(con, to_method)
+    mocked_collect = mocker.patch("polars.LazyFrame.collect")
+    getattr(con, to_method)(t, engine="gpu")
+    to_method_spy.assert_called_once_with(t, engine="gpu")
+    mocked_collect.assert_called_once_with(streaming=False, engine="gpu")
```
jcrist, line 82: Same here.

```suggestion
    mocked_collect = mocker.patch("polars.LazyFrame.collect")
    getattr(con, to_method)(t, engine="gpu")
    mocked_collect.assert_called_once_with(streaming=False, engine="gpu")
```

cpcloud (approved):

PR #10156, opened 2024-09-18
ci(pyspark): name the output path when downloading jar

### LINKED ISSUE #10148: ci(pyspark): local catalog not found, ci job failing somewhat reliably
https://github.com/ibis-project/ibis/actions/runs/10902339563/job/30254037712?pr=10133

I can't reproduce this locally, but it does seem to reproduce in our CI.

cpcloud:
If we don't do this, sometimes the jar ends up as `remotecontent` instead of a basename like `iceberg-spark-runtime-3.5_2.12-1.6.1.jar`.

Fixes #10148.

### DIFF
--- a/.github/workflows/ibis-backends.yml
+++ b/.github/workflows/ibis-backends.yml
@@ -650,7 +650,7 @@ jobs:
         run: docker compose logs
 
   test_pyspark:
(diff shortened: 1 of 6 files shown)


cpcloud:
Ah, this is happening because the way exceptions are structured changed 🤦🏻

gforsyth:
Ugh, thanks for slogging through this.

PR #10162, opened 2024-09-18
docs(datafusion): add imdb live demo reference to end of presentation

gforsyth:
I forgot to add this reference and I wanted to make sure people had access to (more or less) what I typed during the live demo.

### DIFF
--- a/docs/presentations/datafusion-meetup-nyc-2024/talk.qmd
+++ b/docs/presentations/datafusion-meetup-nyc-2024/talk.qmd
@@ -369,6 +369,24 @@ Nope. You should use Ibis _with_ `X`.
 
 ::: {.panel-tabset}
 
+### IMDB Top 10
+
+```python
+import ibis
+
+con = ibis.datafusion.connect()
+ibis.set_backend(con)
+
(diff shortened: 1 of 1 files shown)


docs/presentations/datafusion-meetup-nyc-2024/talk.qmd
```diff
@@ -369,6 +369,24 @@ Nope. You should use Ibis _with_ `X`.
 
 ::: {.panel-tabset}
 
+### IMDB Top 10
+
+```python
```
cpcloud, line 374:
```suggestion
```{python}
```

Not sure if we want to run this in the build, so we know it's always copypastable.

gforsyth: I'm leaning towards no.

While I strongly expect the API usage above to remain unchanged, if it _did_ change, I wouldn't want to edit an older presentation to reflect the newer API.

cpcloud: Good point.

cpcloud (approved):

PR #10163, opened 2024-09-18
docs(synonyms): add synonym list to redirect searches with no results

### LINKED ISSUE #10084: docs: how to get unique values of a column of any type
### Please describe the issue

It looks like the way of doing this is 
```python
penguins[["species"]].distinct()
```
based on the answers of this older issue https://github.com/ibis-project/ibis/issues/6029 but it's still not documented or at least I couldn't find that's how you do it. 

Besides it's normal for someone to search for `unique` and this only apply to arrays on ibis. 

### Code of Conduct

gforsyth:
We get a list of "missed searched" from Algolia and we can add synonyms
that will point to "better" search terms.  Some times we can't know the
right context for the missed searches, but these two seemed obvious and
the list can grow with time.

Fixes #10084.

### DIFF
--- a/.github/workflows/algolia/configure-algolia-api.py
+++ b/.github/workflows/algolia/configure-algolia-api.py
@@ -9,6 +9,15 @@ app_id = os.environ["ALGOLIA_APP_ID"]
 index_name = os.environ["ALGOLIA_INDEX"]
 
 
+ONE_WAY_SYNONYMS = {
+    # A list of search terms that have (historically) not returned results
+    # that we can map to existing search terms that we know are good
+    "md5": ["hashbytes"],
+    "fetchdf": ["to_pandas", "to_polars", "to_pyarrow"],
+    "unique": ["distinct"],
+}
+
+
 def main():
     client = SearchClient.create(app_id, api_key)
     index = client.init_index(index_name)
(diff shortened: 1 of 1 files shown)


.github/workflows/algolia/configure-algolia-api.py
```diff
@@ -9,6 +9,14 @@
 index_name = os.environ["ALGOLIA_INDEX"]
 
 
+ONE_WAY_SYNONYMS = {
+    # A list of search terms that have (historically) not returned results
+    # that we can map to existing search terms that we know are good
+    "md5": ["hashbytes"],
+    "fetchdf": ["to_pandas", "to_polars", "to_pyarrow"],
```
jcrist, line 16: Is this a place we would want to handle better routing of "unique" -> "distinct"? Or is that a docstring change in the `distinct` doc to increase the search relevance there?

cpcloud (approved):
Nice, `ONE_WAY_SYNONYMS` seems like it could get quite large 👀

.github/workflows/algolia/configure-algolia-api.py
gforsyth: This is a place to add that, so if someone searches `unique` the top hit would be `distinct`.  I'll add that.

gforsyth:
> Nice, `ONE_WAY_SYNONYMS` seems like it could get quite large 👀

BIGGER EVERY DAY

.github/workflows/algolia/configure-algolia-api.py
gforsyth: oh, actually, let me test this because I want to see if this just adds `distinct` or if it eliminates the valid `unique` on Arrays

gforsyth: Awwwww yeaaahhhh

PR #10168, opened 2024-09-19
fix(polars): use elementwise flatten to flatten nested arrays

### LINKED ISSUE #10135: bug: polars flatten does not work as expected
### What happened?

Following up on https://github.com/ibis-project/ibis/issues/9995

Thanks @gforsyth and @cpcloud for the quick fix of the above issue. However I don't think it totally fixed the issue.

I am trying to get an `array<array<int64>>` to become an `array<int64>`

Using Polars:

```python
import ibis
ibis.set_backend("polars")
t = ibis.memtable([
    {
        "arr": [[1, 5, 7], [3,4]]
    },
])
t.arr.flatten()
```

```
┏━━━━━━━━━━━━━━━━━━━┓
┃ ArrayFlatten(arr) ┃
┡━━━━━━━━━━━━━━━━━━━┩
│ array<int64>      │
├───────────────────┤
│ [1, 5, ... +1]    │
│ [3, 4]            │
└───────────────────┘
```

Using pandas:

```python
ibis.set_backend("pandas")
t = ibis.memtable([
    {
        "arr": [[1, 5, 7], [3,4]]
    },
])
t.arr.flatten()
```

```
┏━━━━━━━━━━━━━━━━━━━┓
┃ ArrayFlatten(arr) ┃
┡━━━━━━━━━━━━━━━━━━━┩
│ array<int64>      │
├───────────────────┤
│ [1, 5, ... +3]    │
└───────────────────┘
```
Notice how using `pandas` I get a single line back with a single "flattened" array, while using `polars`, I get 2 rows back. Hence I think something is still missing somewhere, unless there is another way to achieve this using ibis or I misunderstand the flatten function. Anyways, I believe the result should be the same by using pandas or polars.

I expect the result to be `[1, 5, 7, 3, 4]` in both cases.

9.5.0

polars 1.5.0

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #10135.

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -1021,7 +1021,11 @@ def array_flatten(op, **kw):
         .then(None)
         .when(result.list.len() == 0)
         .then([])
-        .otherwise(result.flatten())
+        # polars doesn't have an efficient API (yet?) for removing one level of
+        # nesting from an array so we use elementwise evaluation
+        #
+        # https://github.com/ibis-project/ibis/issues/10135
+        .otherwise(result.list.eval(pl.element().flatten()))
     )
 
 
--- a/ibis/backends/tests/errors.py
+++ b/ibis/backends/tests/errors.py
@@ -103,6 +103,7 @@ except ImportError:
     TrinoUserError = None
 
 try:
+    from psycopg2.errors import ArraySubscriptError as PsycoPg2ArraySubscriptError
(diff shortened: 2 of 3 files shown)


cpcloud:
I still want to see if I can workaround the null/empty issue here, so moving to draft for now.

cpcloud:
I think this is the best we can do for now.

PR #10169, opened 2024-09-19
docs(build): run pyodide with python 3.12 in the browser repl

cpcloud:
We are building docs with python 3.12 now. This PR bumps the pyodide setup to use 3.12 as well, along with duckdb 1.1.0.


cpcloud:
Ok, well, pyodide is not letting me ship my own duckdb wheel, which is pretty annoying so we'll just have to wait until the next pyodide release. I'll start a PR over there to get duckdb 1.1.0 up.

cpcloud:
Oh, looks like that's already been done. So, just waiting on them to release.

PR #10180, opened 2024-09-20
chore(docs): clean up current pyodide build

cpcloud:
While we cannot yet upgrade to using python3.12, we can clean up the current jupyterlite build a bit


cpcloud:
Behaving as expected:

image

cpcloud:
Looking good, merging.

image

PR #10184, opened 2024-09-20
docs(build): fetch all commits to enable proper dynamic versioning in the docs build

cpcloud:


cpcloud:
Otherwise we end up with this:

image

ncclementi (approved):

gforsyth:
Ahhh. I wonder if that's why the algolia index sometimes gets messed up

PR #10186, opened 2024-09-20
fix(api): use `to_pyarrow()` instead of `execute()` when pretty printing scalars

### LINKED ISSUE #10185: chore: scalar printing in interactive mode uses `execute()` but other types use `to_pyarrow()`
Everything should be going through pyarrow, unless there's a specific reason we are doing this.

cpcloud:
Closes #10185.

### DIFF
--- a/ibis/backends/tests/test_interactive.py
+++ b/ibis/backends/tests/test_interactive.py
@@ -107,6 +107,10 @@ def test_no_recursion_error(con, monkeypatch):
         repr(expr)
 
 
+@pytest.mark.notimpl(
+    ["impala", "flink", "pyspark"],
+    reason="backend calls `execute` as part of pyarrow conversion",
+)
 def test_scalar_uses_pyarrow(con, table, monkeypatch, mocker):
     monkeypatch.setattr(ibis.options, "interactive", True)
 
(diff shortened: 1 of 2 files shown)


cpcloud:
Right, we have some backends that actually go through pandas -> pyarrow.

ibis/backends/tests/test_interactive.py
```diff
@@ -105,3 +105,18 @@ def test_no_recursion_error(con, monkeypatch):
...
+def test_scalar_uses_pyarrow(con, table, monkeypatch, mocker):
+    monkeypatch.setattr(ibis.options, "interactive", True)
+
+    execute_spy = mocker.spy(con, "execute")
+    to_pyarrow_spy = mocker.spy(con, "to_pyarrow")
+
+    repr(table.limit(1).string_col)
+
+    # execute doesn't get called
+    execute_spy.assert_not_called()
```
jcrist, line 119: > Right, we have some backends that actually go through pandas to get to pyarrow.

If things pass otherwise, the fix seems fine to me without this test at all. No need to assert we're not calling `execute` when reading the formatting code it's clear we go through `to_pyarrow`.

cpcloud: The main point of this test was to make it explicit what happens and in this case make us aware of any backends that have coupling between the pandas and pyarrow format converters. I think I'll leave the test, with appropriate xfails, as a form of documentation.

jcrist (approved):

PR #10205, opened 2024-09-24
feat(api): add support for passing an optional index parameter to array map and filter

### LINKED ISSUE #10201: feat: allow the index as an argument to list functions such as filter and map

Not a problem so much as an enhancement.

Backends like pyspark and duckdb allow you to use the index of the element as an argument in the function calls to array functions like map and filter. (e.g.: for [duckdb](https://duckdb.org/docs/sql/functions/lambda.html#indexes-as-parameters), [pyspark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.transform.html)). 

Having the index of the element is helpful when e.g.: you want to filter a list to take every 3rd element.

Allow an optional second argument to functions like [map](https://ibis-project.org/reference/expression-collections#ibis.expr.types.arrays.ArrayValue.map) and [filter](https://ibis-project.org/reference/expression-collections#ibis.expr.types.arrays.ArrayValue.filter) which would reflect the index of the element.

I can make workarounds for this, by creating an additional column containing the indexes, then running an array_zip between that and my original list to create a struct but it's not as clean

9.2.0

DuckDB, PySpark

### Code of Conduct

cpcloud:
## Description of changes

This PR adds support for passing an additional zero-based index argument to
the callable of array map and filter.

Some of this is a hellscape.

- Postgres, one-based, so we apply a new rewrite rule to all uses of the index argument in the function body 
- DuckDB, one-based, so it gets the rule application
- BigQuery: zero-based so pretty reasonable modification here
- Snowflake: hellscape, because there's no native way to do this, so we have a snowflake-specific rewrite rule that transforms zipped arrays into struct field access.
- ClickHouse: probably the most reasonable implementation, generate an index array and call `arrayMap((x, i) -> do_stuff, arg, index)`.
- Trino: not too bad, just some shenanigans in `filter` to filter values after calling `zip_with(arg, range, return_null_if_predicate_is_false)`.

## Issues closed

Closes #10201.


ibis/expr/types/arrays.py
```diff
@@ -467,26 +507,28 @@ def map(self, func: Deferred | Callable[[ir.Value], ir.Value]) -> ir.ArrayValue:
...
-            raise TypeError(
-                f"`func` must be a Deferred or Callable, got `{type(func).__name__}`"
-            )
 
-        parameter = ops.Argument(
-            name=name, shape=self.op().shape, dtype=self.type().value_type
-        )
-        body = resolve(parameter.to_expr())
-        return ops.ArrayMap(self, param=parameter.param, body=body).to_expr()
+        You can optionally include a second index argument in map function
```
jcrist, line 511: Typo, guessing you meant something like

```suggestion
        You can optionally include a second index argument in the mapped function
```

ibis/expr/types/arrays.py
```diff
@@ -383,7 +385,45 @@ def join(self, sep: str | ir.StringValue) -> ir.StringValue:
...
+
+        body = resolve(**kwargs)
+        return parameter.param, index, body
+
+    def map(
+        self,
+        func: Deferred
+        | Callable[[ir.Value], ir.Value]
+        | Callable[[ir.Value, ir.Value], ir.Value],
+    ) -> ir.ArrayValue:
```
jcrist, line 426: Can't comment on the direct line - could you add a bit about the possible signatures of `func` below in the `Parameters` section? May also want to include that the index is 0-based.

Same goes for `filter` below.

ibis/backends/sql/rewrites.py
```diff
@@ -603,3 +603,19 @@ def lower_sample(_, **kwargs):
             "`Table.sample` with a random seed is unsupported"
         )
     return ops.Filter(_.parent, (ops.LessEqual(ops.RandomScalar(), _.fraction),))
+
+
+@replace(p.ArrayMap | p.ArrayFilter)
+def subtract_one_from_index_argument(_, **kwargs):
```
jcrist, line 609: For readability when looking at a rewrite list on a specific compiler, would you mind adding something to the rule name that better indicates this is for `ArrayMap`/`ArrayFilter` ops? Perhaps

```suggestion
def subtract_one_from_array_map_filter_index(_, **kwargs):
```

ibis/backends/tests/test_array.py
```diff
@@ -474,6 +474,59 @@ def test_array_map(con, input, output, func):
...
+    ("input", "output"),
+    [
+        param(
+            {"a": [[1, None, 2], [4]]},
+            {"a": [[2, None, 5], [5]]},
+            marks=[
+                pytest.mark.notyet(
+                    ["bigquery"],
+                    raises=GoogleBadRequest,
+                    reason="BigQuery doesn't support arrays with null elements",
```
jcrist, line 499: In other tests we transform the result further to avoid retrieving arrays containing null values directly from the query (opting instead for maybe a count of all values or some other transformed aggregate). Not sure if that makes sense here or if it's worth it, just noting.

jcrist (approved):
Just a couple small nits, overall this LGTM!

ibis/backends/tests/test_array.py
cpcloud: The issue is in the construction of the memtable. BigQuery simply cannot hold `NULL`s elements in arrays, it's a fundamental limitation of their system.

ibis/expr/types/arrays.py
cpcloud: Done!

cpcloud:
Clouds are passing:

```
…/ibis on  array-map-and-filter-with-index is 📦 v9.5.0 via 🐍 v3.12.5 via ❄️   impure (shell)
❯ pytest -m snowflake -n 8 --dist loadgroup --snapshot-update -qx && pytest -m bigquery -n auto --dist loadgroup --snapshot-update -q
bringing up nodes...
...............x............x........................x..........................................................................................x...........x...........x.x................................... [ 10%]
.............x.................................s....................................s......s............................................x....x......xx.......x.x..............x..x...x.......x.....x........xx [ 20%]
..xx....x......x...x..x....x.x.x.x...x......x.....x....x....x.........x.......x.x..x....x...x.........x.x......x...............s....................xx.............x................................x........x [ 31%]
...........x.....x...............................xx........x....s...........x.x...x..........x...xx...........................x..................x...x.................................................x...... [ 41%]
..................x....................................................................................................x................xx....x............xx........x.x.x.................................... [ 51%]
.......x.x.x........x..........x....xx...xxxx..............x..x........x....x...x....x..x...x......x.x....xx.........x..x...xx.....................x...x..............x...........................x....x..x... [ 62%]
.....................x.x.........x...........x..............x.................................xx..........................x...............................................x.x........x......x......xx....x..x. [ 72%]
...........x....x..x.....xx....x..............................................x.x................x............x.......x...sxxxx....xxx.x.x.xx.xx.x.xxxxxxxxxxx.x......x..xx.xxx..xx.x...x...x..xx.xx.xxxx..xxx [ 82%]
x.xxx..x..xxx..x.x.x..x.x..x.x.x...x....xx.x.xxx..x.xxxx...x......x..x..x.xxxxxx.x.s.x.xx.x..x................................................................................................................ [ 93%]
.......................................................s.....s..........................s................x.................................                                                                    [100%]
1755 passed, 10 skipped, 228 xfailed in 224.32s (0:03:44)
bringing up nodes...
..........................s..........................sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss..x...xx...x.s..ssX.....x..sssssssssssssssssssss...x.. [  9%]
.....x.........xx.......x.............x.........x...x..x..............x.......x.........x......x.x...x...x......x...........x.......................xx.....x.....xx....x.......................x...x...x...... [ 18%]
.....x...x........................................................x.........................x....................x............s.........x.........................x........................................x.. [ 27%]
.........................................x...................xx......x...x..x...x.......xxx.x...x.........x....x........xx.....xx..X..x...x...x.........xx......Xxxx......xXx....x.x...X...........xx.x..x...x [ 37%]
s...x.x...x............................s............................................x..s.....x.......x..........x.xxxx.x.x.....xx.x..xxxx...x.....x..xxx...x.xxx.x..x...x.x................x..x....x........x. [ 46%]
........x.xxx........x............x........xx.....x....x............x.........x...x..........x.............................x...........x....x..............xx..x...x................xxxxxxxxxxx...xxxx.x....xx [ 55%]
xxxxxxxx....x.........x......x..x.........x...x...x.....x..........................x.x.x...........x..................x.....x....................x.............x............x.......x..x..x.x...........xx.... [ 64%]
x..x...x.........x.......x.............x....x....x.......x...x...xx..x.xx...x.x............x.......x.........xx..x..........x......xx...x..x.x..x...x...x.......x.....x...x.x.......x..xx........xxxx......... [ 74%]
.x..x..xxxxx..xxxxx...xx.xxxxxxxxxxxxxxxxxxxxx.xxxx.xxxxxx.xx.xxxxxxxxxxxxxxxxxxxxx.xxxxxxxxxxxxxxxxxxxxxxx.xxxxxxxxx........x.......xx....................x..........x............x........x................. [ 83%]
...............................x........x....x..x...x.......x.x.xx.xxxxx..xxx.x.....x..............x...........................................................................x.............................. [ 92%]
....................................................................s.s.....x.......s......x...................s..........x.......................................                                             [100%]
1730 passed, 132 skipped, 355 xfailed, 5 xpassed in 324.38s (0:05:24)
```

PR #10206, opened 2024-09-24
feat(memtable): add pyarrow.dataset support

### LINKED ISSUE #10082: feat(duckdb): restore or add new method for consuming `pyarrow.dataset`
In #9666 we deprecated the `read_in_memory` method on the DuckDB backend in favor of creating a `memtable` or passing an object directly to `create_table`.

This does leave out `pyarrow.dataset.Dataset`s, which we previously supported via `read_in_memory`.  They have a `to_table()` method, but materializing them may not be ideal, especially since they can point to (lists of) remote resources and DuckDB can (I believe) push filters down into the dataset reads to do column pruning and filtering.

We could add a memtable wrapper for them, although the deferred aspect is a bit different from other memtable-able objects.  
Whatever we decide on might also apply to pyarrow record-batch-readers.

gforsyth:
## Description of changes

This adds support for using `memtable` with `pyarrow.dataset.Dataset`
objects. Mostly this looks the same as the `PyArrowTableProxy`.

Some deviations:
- An additional `to_pyarrow_lazy` method for grabbing the underlying
object without materializing it, for backends that can perform
pushdowns (currently only enabled with DuckDB).
- Datasets are hashable objects, so we override the PseudoHashable
`hash` implementation (this seemed like less trouble than special-casing
this one memtable everywhere else we look for `TableProxy` objects)
- I've added a `cached_property` so we can avoid repeatedly
materializing the data -- this does _not_ apply to the `lazy` view.

## Issues closed

Resolves #10082


gforsyth:
I'm going to xfail polars for now -- I think we can strip out all of the custom in-memory reading stuff and that will make this work, but that seems like a separate PR

ibis/formats/pyarrow.py
```diff
@@ -341,3 +346,46 @@ def to_polars(self, schema: Schema) -> pl.DataFrame:
...
+
+    def to_frame(self) -> pd.DataFrame:
+        """Convert this input to a pandas DataFrame."""
+        return self._cache.to_pandas()
+
+    def to_pyarrow(self, schema: Schema) -> pa.Table:
+        """Convert this input to a PyArrow Table."""
+        return self._cache
+
+    def to_pyarrow_lazy(self, schema: Schema) -> ds.Dataset:
```
jcrist, line 377: Could we call this `to_pyarrow_dataset` instead? `to_pyarrow_lazy` feels less clear to me.

ibis/backends/tests/test_client.py
```diff
@@ -929,25 +929,7 @@ def test_self_join_memory_table(backend, con, monkeypatch):
...
-                        "mysql",
-                        "oracle",
-                        "polars",
-                        "postgres",
-                        "pyspark",
-                        "risingwave",
-                        "snowflake",
-                        "sqlite",
-                        "trino",
-                    ]
```
jcrist, line 949: I'm a bit concerned about this. Duckdb will push the queries down into the dataset, avoiding loading the full dataset in memory at any point. All other backends will go the `to_frame()`/`to_pyarrow()` route though, which will fully load the data into memory before passing it to create a temp table in the backend. For small data that's fine, but if a user has a Dataset instead of a Table, it's probably larger data. I'm a bit wary of automatically loading this into memory all at once.

A different option would be to error when registering `Dataset` backed `memtable` objects for backends that lack an efficient way to handle them, telling the user to instead call `ibis.memtable(dataset.to_table())` themselves (we could raise this in `to_frame()`/`to_pyarrow()` on the `DatasetProxy`). This makes the potential perf footgun known.

ibis/formats/pyarrow.py
```diff
@@ -341,3 +346,46 @@ def to_polars(self, schema: Schema) -> pl.DataFrame:
...
+
+    def __len__(self):
+        return self.obj.count_rows()
+
+    # pyarrow datasets are hashable, so we override the hash from TableProxy
+    def __hash__(self):
+        return hash(self.obj)
+
+    @cached_property
+    def _cache(self):
```
jcrist, line 366: If we keep around the `to_frame`/`to_pyarrow` codepaths, I'm a bit wary of caching here. For non-local backends the data will be used to create a temp table (meaning we don't need it in local memory anymore). I'd opt to remove the caching for now.

ibis/formats/pyarrow.py
gforsyth: yep

ibis/backends/tests/test_client.py
gforsyth: Yeah, I think this makes sense.

gforsyth: Should we add a `FootgunError` to `common.exceptions`?

jcrist: Nah, I'd just use one of the common ones for this. `UnsupportedOperationError` seems fine, as does just a simple `TypeError`.

ibis/formats/pyarrow.py
```diff
@@ -349,8 +349,15 @@ def to_polars(self, schema: Schema) -> pl.DataFrame:
...
-    __slots__ = ("obj", "__dict__")
+    __slots__ = ("obj",)
     obj: V
+    """You are trying to use a PyArrow Dataset with a backend
+            that will require materializing the entire dataset in local
+            memory.
+
+            If you would like to materialize this dataset, please construct the
+            memtable directly by running `ibis.memtable(my_dataset.to_table())`
+            """
```
jcrist, line 360: odd copy-paste error?

ibis/backends/duckdb/__init__.py
```diff
@@ -1609,7 +1609,10 @@ def _in_memory_table_exists(self, name: str) -> bool:
             return True
 
     def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
-        self.con.register(op.name, op.data.to_pyarrow(op.schema))
+        if hasattr(op.data, "to_pyarrow_dataset"):
+            self.con.register(op.name, op.data.to_pyarrow_dataset(op.schema))
+        else:
+            self.con.register(op.name, op.data.to_pyarrow(op.schema))
```
cpcloud, line 1615: Eventually we may want something like `to_native`, that returns the underlying object.

In the meantime, we can just try to call this and catch the attribute error.

```suggestion
        try:
            self.con.register(op.name, op.data.to_pyarrow_dataset(op.schema))
        except AttributeError:
            self.con.register(op.name, op.data.to_pyarrow(op.schema))
```

ibis/formats/pyarrow.py
```diff
@@ -341,3 +346,58 @@ def to_polars(self, schema: Schema) -> pl.DataFrame:
...
+
+class PyArrowDatasetProxy(TableProxy):
+    __slots__ = ("obj",)
+    obj: V
+
+    def __init__(self, obj: V):
+        self.obj = obj
+
+    def __len__(self):
+        return self.obj.count_rows()
```
cpcloud, line 359: Do we need this method?

```suggestion
```

cpcloud (approved):

PR #10207, opened 2024-09-24
feat: several `Table.sample` improvements

jcrist:
This PR:
- Implements `sample` for `polars`
- Moves the errors for `risingwave`/`druid` to be compile time rather than execution time
- Centralizes the implementation of compiling tablesamples, fixing a TODO in the compiler
- Compiles `Table.sample` to the backend's native `TABLESAMPLE` clause *when possible* for `bigquery`, `postgres`, `mssql`, `oracle`, and `impala`. This is _much_ more efficient when using `method="block"` as it lets the backend avoid reading in whole blocks of data, dramatically reducing IO. This was the main goal of this PR.
- Adds a SQL snapshot test to ensure the native implementation is being used when possible, and expands our execution tests to ensure execution works for all cases.


lostmygithubaccount:
how for Polars? my understanding is lazyframes still don't support `.sample` per https://github.com/pola-rs/polars/issues/3933 (apologies if I'm missing this)

ibis/backends/polars/compiler.py
```diff
@@ -819,6 +819,24 @@ def distinct(op, **kw):
...
+    table = translate(op.parent, **kw)
+    # Disable predicate pushdown since `t.filter(...).sample(...)` could have
+    # different statistical or performance characteristics than
+    # `t.sample(...).filter(...)`. Same for slice pushdown.
+    return table.map_batches(
+        lambda df: df.sample(fraction=op.fraction),
+        predicate_pushdown=False,
+        slice_pushdown=False,
+        streamable=True,
+    )
```
jcrist, line 837: @lostmygithubaccount - see here. I'm subscribed to that issue for when `LazyFrame` gets native support, but until then doing it via `map_batches` seems fine. No way to handle a `seed`, but things work fine otherwise.

cpcloud (approved):

PR #10210, opened 2024-09-24
docs(algolia): feed the index a little snack

gforsyth:
## Description of changes

Noticed someone searched for `strptime` and got no results (there is no `strptime` method).  Seemed like a good candidate for a synonym.

I also noticed some methods missing from the index and added those in.


cpcloud (approved):
Yum yum!

PR #10214, opened 2024-09-24
fix(datatypes): give a more useful error message when trying to cast structs with different fields

NickCrews:
Before, this would error with a KeyError


cpcloud:
Thanks!

Is there a user-facing API implication here?

NickCrews:
> Is there a user-facing API implication here?

The included test case is the simplest public API I could think of, but I actually discovered it when merging two incompatible `array<struct>`s

```python
x = ibis.struct({"x": 5})
y = ibis.struct({"y": 7})
ibis.array([x]).concat(ibis.array([y]))
```

cpcloud (approved):
LGTM!

PR #10218, opened 2024-09-25
fix(snowflake): only compile `sample` to `TABLESAMPLE` on physical tables

jcrist:
In #10207 an xpassing test for snowflake was introduced. Instead of fixing the test marker, this expands our support for `sample` in snowflake by only compiling it to `TABLESAMPLE` on physical tables. Snowflake's `TABLESAMPLE` does work on subqueries, but only if `method="row"` and `seed=None`. At the physical plan this should be identical to filter with `random`; I don't anticipate a perf difference here (and it's a lot easier to avoid adding a snowflake-specific rule to handle this case).

### DIFF
--- a/ibis/backends/sql/compilers/snowflake.py
+++ b/ibis/backends/sql/compilers/snowflake.py
@@ -60,7 +60,12 @@ class SnowflakeCompiler(SQLGlotCompiler):
     LOWERED_OPS = {
         ops.Log2: lower_log2,
         ops.Log10: lower_log10,
-        ops.Sample: lower_sample(),
+        # Snowflake's TABLESAMPLE _can_ work on subqueries, but only by row and without
+        # a seed. This is effectively the same as `t.filter(random() <= fraction)`, and
+        # using TABLESAMPLE here would almost certainly have no benefit over the filter
+        # version in the optimized physical plan. To avoid a special case just for
+        # snowflake, we only use TABLESAMPLE on physical tables.
+        ops.Sample: lower_sample(physical_tables_only=True),
     }
 
     UNSUPPORTED_OPS = (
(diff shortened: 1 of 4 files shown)


ibis/backends/tests/test_generic.py
```diff
@@ -2125,22 +2125,7 @@ def test_dynamic_table_slice_with_computed_offset(backend):
...
-    "method",
-    [
-        "row",
-        param(
-            "block",
-            marks=[
-                pytest.mark.notimpl(
-                    ["snowflake"],
-                    raises=SnowflakeProgrammingError,
-                    reason="SAMPLE clause on views only supports row wise sampling without seed.",
```
jcrist, line 2138: This error message was a bit confusing to me. I initially thought it was saying "SAMPLE on subqueries with no seed only works row wise" (implying that if a seed was set, then block wise would work), but actualy it means "SAMPLE on subqueries only works row wise with no seed"

ibis/backends/sql/compilers/snowflake.py
```diff
@@ -60,7 +60,12 @@ class SnowflakeCompiler(SQLGlotCompiler):
     LOWERED_OPS = {
         ops.Log2: lower_log2,
         ops.Log10: lower_log10,
-        ops.Sample: lower_sample(),
+        # Snowflake's TABLESAMPLE _can_ work on subqueries, but only by row and without
+        # a seed. This is effectively the same as `t.filter(random() <= fraction)`, and
+        # using TABLESAMPLE here would almost certainly have no benefit over the filter
+        # version in the optimized physical plan. To avoid a special case just for
```
cpcloud, line 66: Just curious, what is the special case here?

cpcloud (approved):

PR #10242, opened 2024-09-27
docs(blog): minor edits to last blog

lostmygithubaccount:
## Description of changes

moves an import per preference

the main thing this does is re-compute the `color` column on the random cube and the `faked` table. since I switched laptops, I was using an older version of the data that suffered from an old histogram issue I reported + the faked data wasn't matching the code


lostmygithubaccount:
I'm confused (git skill issues) I'll fix and reopen this later

PR #10259, opened 2024-10-01
docs: update README backends

IndexSeek:
## Description of changes

Removing pandas from the list of backends in the README and adjusting the phrasing around the count. 

I wasn't sure whether to say "15+ backends" or "nearly 20," so I opted for the latter, which was closer to the actual number, 18.


cpcloud (approved):
Thanks!

PR #10278, opened 2024-10-06
feat(api): add `StringValue.as_time` for parsing strings into times

### LINKED ISSUE #8908: feat: Add to_date and to_time to StringValue

saschahofmann:
## Description of changes

* Adding `as_time` method that converts string to TimeValue based on a formatted. By default converts to timestamps and casts to time
* Implements `parse_time` for BigQuery and `to_date` for Snowflake
* [ ] Is it necessary to implement the pyspark_format tests as well, I had the impression that its mostly about dates?

## Issues closed

* Closes #8908

## Note

I find it kinda tricky to setup the tests/dev environment e.g. the docker postgres database always collides with my own local postgres so I have to shut it down in order to run the tests. There are also a lot of errors in the VSCode test discovery, I think its because of the two environments with and without Flink on Mac.

For pyspark I am getting this

```
pyspark.errors.exceptions.base.PySparkRuntimeError: [JAVA_GATEWAY_EXITED] Java gateway process exited before sending its port number.
```

For duckdb
```
duckdb.duckdb.IOException: IO Error: No files found that match the pattern "ibis/ci/ibis-testing-data/parquet/geo_wkb.parquet"
```

### DIFF
--- a/ibis/backends/tests/test_temporal.py
+++ b/ibis/backends/tests/test_temporal.py
@@ -1179,7 +1179,7 @@ def test_integer_to_timestamp(backend, con, unit):
     raises=com.OperationNotDefinedError,
 )
 @pytest.mark.notimpl(["exasol"], raises=com.OperationNotDefinedError)
-def test_string_to_timestamp(alltypes, fmt):
+def test_string_as_timestamp(alltypes, fmt):
     table = alltypes
     result = table.mutate(date=table.date_string_col.as_timestamp(fmt)).execute()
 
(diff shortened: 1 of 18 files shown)


saschahofmann:
@deepyaman I finally got around to it and didn't find too many databases with their own implementation so it was quite straightforward

ibis/backends/tests/test_temporal.py
```diff
@@ -1260,6 +1260,31 @@ def test_string_to_date(alltypes, fmt):
         assert val.strftime("%m/%d/%y") == result["date_string_col"][i]
 
 
+
+@pytest.mark.parametrize(
```
cpcloud, line 1264: There's only one parameter here, so this is overkill.

cpcloud (approved):

PR #10287, opened 2024-10-09
perf(clickhouse): avoid unnecessary construction of a pyarrow table when inserting a pandas dataframe

cpcloud:


kszucs (approved):

ibis/backends/clickhouse/__init__.py
```diff
@@ -438,9 +438,7 @@ def insert(
         if isinstance(obj, pa.Table):
             return self.con.insert_arrow(name, obj, settings=settings, **kwargs)
```
cpcloud, line 439: Looks like we don't hit this code path in a test anywhere. I'll add a test in another PR.

cpcloud: #10288

PR #10302, opened 2024-10-11
fix(mssql): exclude window frame from rank

### LINKED ISSUE #10291: bug: MSSQL `RANK OVER` does not support row clause
### What happened?

It looks like if `rows` are not specified when calling `.over()`, by default the compiler will add "UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING" to the rows clause. This causes issue with T-SQL since the rows clause [cannot be specified when using its ranking functions. ](https://learn.microsoft.com/en-us/sql/t-sql/functions/rank-transact-sql?view=sql-server-ver16#arguments)

When executing the query, this causes a SQL error saying "The function "rank" may not have a window frame"

Compiled:
```
SELECT 
[t0].[col1], 
[t0].[col2], 
[t0].[col3], 
RANK() OVER (PARTITION BY [t0].[PfId] ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) - 1 AS [rank_] 
FROM table [t0]
```

The query would work after removing `ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING`

10.0.0dev97

MSSQL

### Relevant log output

_No response_

### Code of Conduct

IndexSeek:
## Description of changes

Resolves an error where the RANK window function added a `ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING` window frame, which is unsupported by MSSQL. 

> The <rows or range clause/> of the OVER clause cannot be specified for the RANK function. For more information, see [OVER Clause (Transact-SQL)](https://learn.microsoft.com/en-us/sql/t-sql/queries/select-over-clause-transact-sql?view=sql-server-ver16).

## Issues closed

Resolves #10291

### DIFF
--- a/ibis/backends/mssql/tests/snapshots/test_window/test_rank_no_window_frame/out.sql
+++ b/ibis/backends/mssql/tests/snapshots/test_window/test_rank_no_window_frame/out.sql
@@ -1,0 +1,5 @@
+SELECT
+  [t0].[color],
+  [t0].[price],
+  RANK() OVER (PARTITION BY [t0].[color] ORDER BY CASE WHEN [t0].[price] IS NULL THEN 1 ELSE 0 END, [t0].[price] ASC) - 1 AS [MinRank()]
+FROM [diamonds_sample] AS [t0]
--- a/ibis/backends/mssql/tests/test_window.py
+++ b/ibis/backends/mssql/tests/test_window.py
@@ -1,0 +1,11 @@
+from __future__ import annotations
+
+import ibis
+
+
+def test_rank_no_window_frame(con, snapshot):
+    t = ibis.table(schema=dict(color=str, price=int), name="diamonds_sample")
+    expr = t.mutate(ibis.rank().over(group_by="color", order_by="price"))
+    sql = ibis.to_sql(expr, dialect="mssql")
+
+    snapshot.assert_match(sql, "out.sql")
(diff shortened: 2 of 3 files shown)


gforsyth (commented):
Nice! Thanks @IndexSeek ! 

I imagine you tested this out locally, but maybe we should add a snapshot test to confirm that we're generating the correct SQL?

IndexSeek:
Thank you @gforsyth, happy to help! 

I was thinking a test for this scenario would be worthwhile, but wasn't sure how best to go about it. 

Would we want to set something up a test to check that the text `ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING` should not appear in the compiled SQL for these backends that do not support the window frame?

gforsyth:
Yeah, these can be tricky to test. 

You can look at https://github.com/ibis-project/ibis/blob/f817a7ad6ed5ba02afa5e1208896ba8e03439858/ibis/backends/duckdb/tests/test_datatypes.py#L107
For an example of a snapshot test. I think if you have an expression that will generate a rank query, then we can check in the snapshot and that will confirm for us that the window bounds are excluded on mssql

IndexSeek:
Thank you for pointing me to an example. I gave this a try; hopefully, it covers this scenario. I'm happy to make any changes as needed!

gforsyth (approved):
Nice! Thanks @IndexSeek !

PR #10318, opened 2024-10-15
test(datafusion): ensure that tpc_66 is not strict for datafusion

cpcloud:
Make TPC-DS query 66 not flaky for DataFusion


gforsyth (approved):
:shipit:

PR #10322, opened 2024-10-16
feat(polars): add ArrayRemove operation

IndexSeek:
## Description of changes

Adds ArrayRemove support for the Polars backend via [polars.Series.list.set_difference](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.list.set_difference.html#).

I didn't `translate` `op.other` this time around, I was having issues when doing that (it wasn't returning anything for some reason). Happy to make any changes if it is necessary to do the `translate` or otherwise adjust.

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -1521,3 +1521,10 @@ def visit_ArrayContains(op, **kw):
     arg = translate(op.arg, **kw)
     value = translate(op.other, **kw)
     return arg.list.contains(value)
+
+
+@translate.register(ops.ArrayRemove)
+def visit_ArrayRemove(op, **kw):
+    arg = translate(op.arg, **kw)
+    value = _literal_value(op.other)
+    return arg.list.set_difference(pl.lit([value]))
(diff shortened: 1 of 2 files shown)


kszucs (approved):

ibis/backends/polars/compiler.py
```diff
@@ -1521,3 +1521,10 @@ def visit_ArrayContains(op, **kw):
     arg = translate(op.arg, **kw)
     value = translate(op.other, **kw)
     return arg.list.contains(value)
+
+
+@translate.register(ops.ArrayRemove)
+def visit_ArrayRemove(op, **kw):
+    arg = translate(op.arg, **kw)
+    value = _literal_value(op.other)
```
kszucs, line 1529: Based on the documentation it could be a non-literal as well https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.list.set_difference.html 

I can see that you tried that as well but had some problems, could you create a follow-up ticket for it?

IndexSeek: Thanks for the review! I have created https://github.com/ibis-project/ibis/issues/10326 so that this can be reviewed/discussed in more detail.

PR #10333, opened 2024-10-18
fix(duckdb): use simple GEOMETRY type for all geospatial data (#10324)

### LINKED ISSUE #10324: bug: Unable to cache table after creating geometry column
### What happened?

I'm trying to create a table with a `geometry` column using given lat/long coodinates.

I can successfully create this column, but when calling `.cache()`, I get a `ParserError`.

Here's an example that replicates it

```python
 from ibis.interactive import *
 t = ibis.examples.zones.fetch()
 t = t.mutate(geom=_.x_cent.point(_.y_cent)).cache()
```

I don't see this behaviour when I do the following
```
 t = ibis.examples.zones.fetch()
 t = t.cache() #geom already exists in the zones table
```

The only material difference in the SQL in each case is the use of `ST_POINT` when creating the geometry 

```python
SELECT
  *
  REPLACE (ST_ASWKB("geom") AS "geom")
FROM (
  SELECT
    "t0"."OBJECTID",
    "t0"."Shape_Leng",
    "t0"."Shape_Area",
    "t0"."zone",
    "t0"."LocationID",
    "t0"."borough",
    "t0"."x_cent",
    "t0"."y_cent",
    ST_POINT("t0"."x_cent", "t0"."y_cent") AS "geom"
  FROM "ibis_read_geo_xqclf77uwbhqzmcrvtctt27c34" AS "t0"
)
```

```python
SELECT
  *
  REPLACE (ST_ASWKB("geom") AS "geom")
FROM (
  SELECT
    *
  FROM "ibis_read_geo_qwgreoxngfbmjdveyqna5ldkfy"
)
```

9.5.0

DuckDB

### Relevant log output

```sh
ParserException: Parser Error: Expected a constant as type modifier
File <command-2809322511957941>, line 2
      1 t = ibis.examples.zones.fetch()
----> 2 t = t.mutate(geom=_.x_cent.point(_.y_cent)).cache()
```

### Code of Conduct

anjakefala:
Ensure compatibility with DuckDB's implementation of geospatial types.

I would love to add tests. Which is the appropriate test file?

- [ ] added tests 

## Description of changes

In https://github.com/ibis-project/ibis/issues/10324 the temporary table generated the incorrect SQL definition of `GEOMETRY(POINT)` instead of the simple `GEOMETRY`. This ensures that the simple `GEOMETRY` will be used for all geospatial types.

## Issues closed

* Resolves #10324

### DIFF
--- a/ibis/backends/duckdb/tests/test_geospatial.py
+++ b/ibis/backends/duckdb/tests/test_geospatial.py
@@ -564,9 +564,13 @@ def test_to_geo_geom_only(con, driver, tmp_path):
     assert dread.count().execute() == 2
 
 
-def test_cache_geometry(con):
+def test_cache_geometry(con, monkeypatch):
     # ibis issue #10324
-    data = ibis.memtable({"x": [1, 3], "y": [2, 4]})
-    data = data.mutate(geom=data.x.point(data.y)).cache()
+
+    # monkeypatching is necessary to ensure the correct backend is used for
+    # caching
+    monkeypatch.setattr(ibis.options, "default_backend", con)
+    data = ibis.memtable({"x": [1], "y": [2]})
+    data = data.select(geom=data.x.point(data.y)).cache()
     result = data.execute()
-    assert result["geom"].iloc[0] == shapely.Point(1, 2)
+    assert result.at[0, "geom"] == shapely.Point(1, 2)
--- a/ibis/backends/duckdb/tests/test_geospatial.py
+++ b/ibis/backends/duckdb/tests/test_geospatial.py
@@ -562,3 +562,11 @@ def test_to_geo_geom_only(con, driver, tmp_path):
     dread = con.read_geo(out)
 
     assert dread.count().execute() == 2
+
+
+def test_cache_geometry(con):
+    # ibis issue #10324
+    data = ibis.memtable({"x": [1, 3], "y": [2, 4]})
+    data = data.mutate(geom=data.x.point(data.y)).cache()
+    result = data.execute()
+    assert result["geom"].iloc[0] == shapely.Point(1, 2)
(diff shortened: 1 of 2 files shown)


gforsyth:
Thanks @anjakefala ! 

Since this change is specific to the DuckDB backend, tests should go in `ibis/backends/DuckDB/tests/`

For this particular fix, either `test_datatypes.py` or `test_geospatial.py` work -- whichever feels like a better fit to you!

cpcloud:
@anjakefala Ping on this! Still interested in adding a test?

anjakefala:
@cpcloud Yes, sorry! I'm travelling soon, but was planning to add tests today!

anjakefala:
It seems that the example zone already contained a `geom` column. Reproducing the scenario still unfortunately passed on `main`:

```
 565                                                                                                                                                                                  
 566 def test_cache_geometry(con):                                                                                                                                                    
 567     # ibis issue #10324                                                                                                                                                          
 568     data = ibis.memtable({"x": [1, 3], "y": [2, 4]})                                                                                                                             
 569     data = data.mutate(geom=data.x.point(data.y))                                                                                                                                
 570     data = data.mutate(geom=data.x.point(data.y)).cache()                                                                                                                        
 571     result = data.execute()                                                                                                                                                      
 572     assert result['geom'].iloc[0] == shapely.Point(1, 2)  
```

cpcloud:
Your test fails for me on `main` with the same error reported by the user. Are you sure you were running your test against `main`?

cpcloud (approved):

anjakefala:
@cpcloud HUH. Okay, then. Carry on.

cpcloud:
Thanks!

PR #10342, opened 2024-10-21
ci(google-auth): use workload identity federation instead of json

cpcloud:


tswast (commented):
Love it! Still working on getting approvals to enable this in our ibis-gbq test project.

.github/workflows/ci-data.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
       - uses: google-github-actions/auth@v2
         with:
-          credentials_json: ${{ secrets.GCP_CREDENTIALS }}
+          workload_identity_provider: "${{ secrets.WIF_PROVIDER_NAME }}"
```
tswast, line 45: Per https://github.com/google-github-actions/auth/blob/main/README.md#workload-identity-federation-through-a-service-account do we want to add the currently used service account name here as a `service_account` key?

The alternative is that we use the workload identity pool as the principal in our IAM permissions, so we'd need to grant it permission to do all the same things the current service account can do.

.github/workflows/ibis-backends-cloud.yml
```diff
@@ -125,7 +127,7 @@ jobs:
 
       - uses: google-github-actions/auth@v2
         with:
-          credentials_json: ${{ secrets.GCP_CREDENTIALS }}
+          workload_identity_provider: "${{ secrets.WIF_PROVIDER_NAME }}"
```
tswast, line 130: Same here, include the `service_account`?

.github/workflows/ci-data.yml
tswast: Let's wait a second on adding the service account. I think I can mail a change internally to grant the pool access to BigQuery resources needed to run the tests. Will try that first and see if additional approvals would be needed.

tswast: Looks like I can grant these permissions directly, copying the same project-level permissions as the existing service account(s). Ignore this thread.

.github/workflows/ibis-backends-cloud.yml
tswast: Disregard.

cpcloud:
@tswast Should we just give this a go and merge it?

tswast:
Sorry, I misspoke at PyData NYC. Looks like I'm still waiting for approval on one more internal change (689521213 for my Goolger's reference) before this can work.

cpcloud:
@tswast Any updates here?

tswast:
Thanks for the reminder. I'm still waiting on a thumbs up from our security team. I think I went through the right process, but it's been stuck on the last LGTM for a couple months now. 😭 I'll ping them again today. Maybe I can get another reviewer assigned?

.github/workflows/ci-data.yml
tswast, line 45: Per https://github.com/google-github-actions/auth/blob/main/README.md#preferred-direct-workload-identity-federation we probably need a `project_id` here, too. With key files, it can pull that from the JSON, but I don't think the default project ID is available in most other authentication methods.

```suggestion
          project_id: 'ibis-gbq'
          workload_identity_provider: "${{ secrets.WIF_PROVIDER_NAME }}"
```

Also, I got approval and created the provider

```
gcloud iam workload-identity-pools providers describe "ibis" \
  --project="${PROJECT_ID}" \
  --location="global" \
  --workload-identity-pool="github" \
  --format="value(name)"
```

Output:

```
projects/546535678771/locations/global/workloadIdentityPools/github/providers/ibis
```

FYI for making the "secret". Alternatively, I think it'd be fine to hardcode this.

tswast (approved):
Approving, but consider resolving the comments on `.github/workflows/ci-data.yml` before merging.

.github/workflows/ci-data.yml
cpcloud: Hardcoding it seems a bit more understandable, so I'll do that.

cpcloud: Actually an environment variable maybe is slightly better only because we repeat this in 3 places?

cpcloud:
@tswast Thanks for pushing this through, merging on green!

cpcloud:
@tswast It looks like two tests are failing on `main` (which is still a win since everything else is passing 🥳) see here: https://github.com/ibis-project/ibis/actions/runs/12507938156/job/34896929010

Can you help debug the permissions here?

cpcloud:
Looks like step 5 [here](https://github.com/google-github-actions/auth/blob/main/README.md#direct-wif) might be required (giving permissions to specific cloud resources).

tswast:
Thanks for the heads up. Taking a look now. Step 5 does have a concerning warning, which I wonder if applies in this case?

>  However, not all Google Cloud resources support principalSet identities

For example, the failure to read `gs://ibis-testing-libraries/lodash.min.js` from BigQuery is concerning, as I have granted the GitHub provider access to that bucket in two different ways. I'm not seeing any notes about `principalSet` in the [BigQuery + GCS](https://cloud.google.com/bigquery/docs/external-data-cloud-storage) or [BigQuery + JS libraries](https://cloud.google.com/bigquery/docs/user-defined-functions#including-javascript-libraries) docs. I'll do some experimentation and see if I can narrow down what's going on.

cpcloud:
Thanks. If it's helpful, writing data to a bucket (the benchmark data upload) is working, so accessing a bucket directly is not a problem.

Just to clarify, from my POV I see two things that might be an issue:

1. Reading a bucket from within BigQuery (perhaps specific to JS libraries)
2. Accessing another project's data, even if it's publicly accessible, like the one we're testing against.

cpcloud:
If you can't find anything in the next few business days, we can xfail those tests until we understand whether there's an issue.

tswast:
I tried out GitHub Actions on another repo (https://github.com/tswast/code-snippets/pull/3) to see if I could reproduce the BigQuery + JS library failure, but I was able to call `bqutil.fn.xml_to_json` in https://github.com/tswast/code-snippets/actions/runs/12521512469/job/34928670087

I suspect it might be a permission issue rather than a BigQuery bug with such credentials. I'll try to take a look at the ibis setup a bit more next week.

PR #10345, opened 2024-10-21
docs: correct link generated in operations matrix

### LINKED ISSUE #10270: docs: operation support matrix does not have working links
### Please describe the issue

The [operation support matrix](https://ibis-project.org/backends/support/matrix) has links to each operation, but they don’t work.

Currently they go to something of the form `https://ibis-project.org/backends/support/operations.html#ibis.expr.operations.analytic.Analytic` but it appears that has moved to `https://ibis-project.org/reference/operations#ibis.expr.operations.analytic.Analytic`.

### Code of Conduct

Jlobblet:
## Description of changes

The [operation support matrix](https://ibis-project.org/backends/support/matrix) has links to each operation, but they don’t work. This fixes that by updating the link to the current address of those operations.

## Issues closed

- Resolves #10270

### DIFF
--- a/docs/backends/support/matrix.qmd
+++ b/docs/backends/support/matrix.qmd
@@ -88,7 +88,7 @@ def make_support_matrix():
 
     def make_link(parts):
         module, op = parts[-2:]
-        return f'<a href="./operations.html#ibis.expr.operations.{module}.{op}">{op}</a>'
+        return f'<a href="/reference/operations#ibis.expr.operations.{module}.{op}">{op}</a>'
 
     support_matrix = (
         pd.DataFrame(support)


cpcloud:
@Jlobblet Thanks!

I've kicked off a docs-preview job for this PR, so in about 10-15 minutes we can check that your fix works!

Jlobblet:
Looks like there are still some oddities around things present in the matrix but not the reference (e.g., DenseRank) but that is beyond the scope of this PR

Jlobblet:
Other things I have spotted so far that don't match up:

- core category
- some relations
- rewrites category
- udf category
- vectorized category

cpcloud:
Yep, I think we are simply missing all those in the API docs. I'm not sure if we can generate them automatically and keep the same layout.

At the moment we're simply listing all the operations we want to show up in the docs in `_quarto.yml`, but ideally quarto docs would discover this for us and generate the output.

cpcloud:
In any event, I'm going to merge this PR. Thanks again!

cpcloud (approved):

PR #10351, opened 2024-10-22
fix(api): ensure consistent typing in `round` output type

### LINKED ISSUE #10262: bug: chained casts are incorrectly merged
### What happened?

```python
import ibis

ibis.memtable({"x": [1.2, 3.4]}).x.round().cast(int).cast(str).execute().to_list()
# gives
# ['1.0', '3.0']
# I expect
# ['1', '3']

# not a problem with literals?
ibis.literal(1.2).round().cast(int).cast(str).execute()
# '1'
```

main

duckdb

### Relevant log output

_No response_

### Code of Conduct

cpcloud:
Fixes #10262.

BREAKING CHANGE: `None` is no longer allowed as the `digits` argument to `round`. Use `0` instead. Default behavior is unchanged.

### DIFF
--- a/ibis/expr/operations/numeric.py
+++ b/ibis/expr/operations/numeric.py
@@ -148,19 +148,31 @@ class Round(Value):
     """Round a value."""
 
     arg: StrictNumeric
-    # TODO(kszucs): the default should be 0 instead of being None
-    digits: Optional[Integer] = None
+    digits: Integer
 
     shape = rlz.shape_like("arg")
 
     @property
     def dtype(self):
-        if self.arg.dtype.is_decimal():
-            return self.arg.dtype
-        elif self.digits is None:
-            return dt.int64
-        else:
-            return dt.double
+        digits = self.digits
+        arg_dtype = self.arg.dtype
+
+        raw_digits = getattr(digits, "value", None)
+
+        # decimals with literal-typed digits return decimals
+        if arg_dtype.is_decimal() and raw_digits is not None:
+            return arg_dtype.copy(scale=raw_digits)
+
+        nullable = arg_dtype.nullable
+
+        # if digits are unspecified that means round to an integer
+        if raw_digits is not None and raw_digits == 0:
+            return dt.int64.copy(nullable=nullable)
+
+        # otherwise one of the following is true:
+        # 1. digits are specified as a more complex expression
+        # 2. self.arg is a double column
+        return dt.double.copy(nullable=nullable)
 
 
 @public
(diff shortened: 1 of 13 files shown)


ibis/expr/operations/numeric.py
```diff
@@ -148,19 +148,31 @@ class Round(Value):
     """Round a value."""
 
     arg: StrictNumeric
-    # TODO(kszucs): the default should be 0 instead of being None
-    digits: Optional[Integer] = None
+    digits: Integer
```
NickCrews, line 151: Should we restrict this to a plain python int? eg known at expression construction time? I think we do. Duckdb only supports constant precisions, eg `SELECT round(1.234, x) FROM (SELECT 2 AS x) AS t;` errors. IDK about other backends, but that seems likely as well. In the `if raw_digits == 0` code below, we are making an assumption that it is not an ibis expression.

ibis/expr/operations/numeric.py
```diff
@@ -148,19 +148,31 @@ class Round(Value):
...
+            return arg_dtype.copy(scale=raw_digits)
+
+        nullable = arg_dtype.nullable
+
+        # if digits are unspecified that means round to an integer
+        if raw_digits is not None and raw_digits == 0:
+            return dt.int64.copy(nullable=nullable)
+
+        # otherwise one of the following is true:
+        # 1. digits are specified as a more complex expression
```
NickCrews, line 173: see above, I wonder if we should only support digits being a python int

ibis/tests/expr/test_sql_builtins.py
```diff
@@ -148,7 +148,7 @@ def test_sign(functional_alltypes, lineitem):
 def test_round(functional_alltypes, lineitem):
     result = functional_alltypes.double_col.round()
     assert isinstance(result, ir.IntegerColumn)
-    assert result.op().args[1] is None
+    assert result.op().args[1] == ibis.literal(0).op()
```
NickCrews, line 151: this seems like it is testing the implementation and I would just remove it, but also fine with keeping it.

ibis/expr/operations/numeric.py
cpcloud: Yeah, supporting only literal `int` seems like the right move for now. I was trying to preserve backwards compatibility.

FYI, there are many possible return types here regardless of supporting only `int`.

ibis/expr/operations/numeric.py
cpcloud: I'm going to make this change in a subsequent PR.

ibis/tests/expr/test_sql_builtins.py
cpcloud: Going to keep it for now.

PR #10353, opened 2024-10-23
fix(compiler): respect order of ops in FloorDivide

NickCrews:
Consider `ibis.literal(6) // (1 / ibis.literal(2))`.

Before, this compiled to `6 / 1 / 2 -> 3`, when we really want `6 / (1 / 2) -> 12`.

This fixes that.

Found in https://github.com/NickCrews/mismo/issues/74

### DIFF
--- a/ibis/backends/sql/compilers/base.py
+++ b/ibis/backends/sql/compilers/base.py
@@ -847,7 +847,7 @@ class SQLGlotCompiler(abc.ABC):
         return arg
 
     def visit_FloorDivide(self, op, *, left, right):
-        return self.cast(self.f.floor(left / right), op.dtype)
+        return self.cast(self.f.floor(sge.paren(left) / sge.paren(right)), op.dtype)
 
     def visit_Ceil(self, op, *, arg):
         return self.cast(self.f.ceil(arg), op.dtype)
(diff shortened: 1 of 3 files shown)


cpcloud (approved):
Thanks!

NickCrews:
Uhhhh it is VERY annoying to get these very opaque flink errors and then not be able to run flink locally to debug.

cpcloud:
Ultimately the error isn't opaque in this case:

https://github.com/ibis-project/ibis/actions/runs/11470981543/job/31921152707?pr=10353#step:16:256

image

It is possible to run the Flink backend, but it's a bit of a slog to get it working due to their Python API requiring an ancient version of arrow.

NickCrews:
Ah, thanks, I didn't see that buried in the rest of the stacktrace. I will look harder next time.

I wonder if this is pointing to that `1/2` in flink performs integer division and results in 0. I made my test case be immune to this, and added a tiny separate test case to check if my hunch is true. If so, then I can open a separate PR that adds that test, marked as xfail on flink.

NickCrews:
int division test filed at https://github.com/ibis-project/ibis/pull/10359#issue-comment-box

PR #10358, opened 2024-10-23
feat(duckdb): forward UDF configuration dict as kwargs during registration

NickCrews:
The backend-specific kwargs were not forwarded during udf registration. Now they are.

I discovered this when I hit an error just like the test shows.

I also did a little refactoring beforehand, I can remove that commit if you want.

I think a lot of the other backends also suffer from this bug, jusdging by how I don't see their registration hooks using `udf_node.__config__` at all. But this is a tricky thing to test, since the args are backend-specific.

If you can think of a better way to test this I'm open to it, the way I did here seems sort of brittle, dependent on implementation details of duckdb?

### DIFF
--- a/ibis/backends/duckdb/tests/test_udf.py
+++ b/ibis/backends/duckdb/tests/test_udf.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import duckdb
 import pytest
 from pytest import param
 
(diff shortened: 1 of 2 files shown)


ibis/backends/duckdb/tests/test_udf.py
```diff
@@ -103,3 +104,24 @@ def dont_intercept_null(x: int) -> int:
...
+    def nullify_two(x: int) -> int:
+        return None if x == 2 else x
+
+    no_kwargs = udf.scalar.python()(nullify_two)
+    with_kwargs = udf.scalar.python(null_handling="special")(nullify_two)
+
+    # If we return go Non-NULL -> Non-NULL, then passing null_handling="special"
+    # will not change the result
+    assert con.execute(no_kwargs(ibis.literal(1))) == 1
+    assert con.execute(with_kwargs(ibis.literal(1))) == 1
```
NickCrews, line 121: I can just remove this if you want, this is just to give background on what passing `null_handling="special"` does, but that's not really important.

NickCrews:
An [unrelated spark test failed](https://github.com/ibis-project/ibis/actions/runs/11488423051/job/31975132256?pr=10358) on the first run, when I [reran it](https://github.com/ibis-project/ibis/actions/runs/11488423051?pr=10358) it passed. So I think that is just an unrelated flake?

cpcloud (approved):

ibis/backends/duckdb/tests/test_udf.py
cpcloud: It's fine to keep it.

PR #10359, opened 2024-10-23
test: add test that `/` performs true division, not floor division

NickCrews:
I can file an issue to track the actual flink fix for this if you want.


cpcloud (changes requested):
Seems like a test-only PR here isn't going to do much beyond add an epsilon amount of time to overall test runs.

I'll poke at fixing this.

cpcloud (approved):

cpcloud (approved):

NickCrews:
Thanks @cpcloud! Nice job tag teaming this

PR #10361, opened 2024-10-24
feat(api): add `name` kwarg to `Table.value_counts()`

NickCrews:
This now matches Column.value_counts()

The doctest can act as a test I figure.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -4625,7 +4625,8 @@ class Table(Expr, _FixedTextJupyterMixin):
         └───────┴───────────┴───────────────────┘
         """
         columns = self.columns
-        name = name or "_".join(columns) + "_count"
+        if name is None:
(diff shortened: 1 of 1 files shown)


ibis/expr/types/relations.py
```diff
@@ -4619,9 +4625,8 @@ def value_counts(self) -> ir.Table:
         └───────┴───────────┴───────────────────┘
         """
         columns = self.columns
-        return self.group_by(columns).agg(
-            lambda t: t.count().name("_".join(columns) + "_count")
-        )
+        name = name or "_".join(columns) + "_count"
```
cpcloud, line 4628:
```suggestion
        if name is None:
        	name = "_".join(columns) + "_count"
```

I always find this self-assignment pattern awkward to read.

cpcloud (changes requested):

cpcloud (approved):

PR #10363, opened 2024-10-24
fix(trino): remove invalid ranking function window frames

cpcloud:
Fixes the failing trino tests.


cpcloud:
Both Trino jobs passed, merging!

PR #10367, opened 2024-10-24
docs(examples): add temporal `truncate` examples

IndexSeek:
## Description of changes

Adds doc examples for the TimestampValue and DateValue truncate methods.


IndexSeek:
Is it worthwhile to break this up into additional examples or maybe not include so many columns?

I think demonstrating the behavior here is helpful, especially for truncating down to the week, where the year can change due to the start of the week.

cpcloud:
> Is it worthwhile to break this up into additional examples or maybe not include so many columns?
> 
> I think demonstrating the behavior here is helpful, especially for truncating down to the week, where the year can change due to the start of the week.

Fine to do this in a follow-up. Thanks for churning all these examples out!

cpcloud (approved):

IndexSeek:
> Fine to do this in a follow-up. Thanks for churning all these examples out!

Awesome, thank you for the review and for the new commit. That gives me a good example for adding more examples here. I will keep on keeping on!

PR #10381, opened 2024-10-28
feat(deps): support pyarrow 18

cpcloud:
Bump pyarrow upper bound to 18.


kszucs (approved):
The nix build is failing for some reason, otherwise it looks good to me.

cpcloud:
The reason is because the MacOS SDK is older than what pyarrow has published (and arrow-cpp in nixpkgs is at 18.0.0 yet (but I opened a PR for that https://github.com/NixOS/nixpkgs/pull/351868)).

It would be great if pyarrow published 11_0 arm64 macos wheels.

cpcloud:
@kszucs Should I open an issue about the additional arm64 macos wheels?

cpcloud:
We can also disable the aarch64-darwin build until https://github.com/NixOS/nixpkgs/pull/346043 is merged into `nixpkgs-unstable-small`.

noklam:
I see this is included in 10.0.0 roadmap, any idea when will this be released? We are currently blocked by this because we need PyArrow 18 (we have tons of dependencies like spark etc, so it will only work with PyArrow 18 for Python 3.13).

We can wait a bit if it's going to be release soon, otherwise I may conditionally drop ibis support for Python 3.13

https://github.com/kedro-org/kedro-plugins/pull/938

PR #10383, opened 2024-10-28
docs(examples): add temporal interval examples

IndexSeek:
## Description of changes

Adds an example showing what the output of `ibis. interval` looks like, adding and subtracting ten days from a timestamp and gets a bit more granular with a defined interval. 

I feel like it's particularly important to have an example for `interval` because I often find myself trying to figure out how to express the equivalent of the SQL `DATEADD` operation.


cpcloud (approved):
Excellent, thank you!

PR #10393, opened 2024-10-29
feat(dt): allow trailing comma in struct string shorthands

NickCrews:
Now, `dt.dtype("struct<a: int32, b: string,>")` is allowed


ibis/expr/datatypes/tests/test_parse.py
```diff
@@ -159,6 +159,16 @@ def test_struct_with_string_types():
...
+def test_struct_trailing_comma_single():
+    result = dt.dtype("struct<a: int32,>")
+    assert result == dt.Struct.from_tuples([("a", dt.int32)])
+
+
+def test_struct_trailing_comma_multi():
+    result = dt.dtype("struct<a: int32, b: string  , >")
+    assert result == dt.Struct.from_tuples([("a", dt.int32), ("b", dt.string)])
+
+
```
cpcloud, line 171: This is missing a test case of `struct<,>`, which current does **not** error. I would argue shouldn't parse even though we're currently allowing empty structs.

cpcloud: I will add this test case and fix the PR.

PR #10397, opened 2024-10-30
fix(sql): include dialect when creating sqlglot functions

NickCrews:
See https://github.com/tobymao/sqlglot/issues/4312


ibis/backends/sql/compilers/base.py
```diff
@@ -1230,7 +1234,7 @@ def __sql_name__(self, op: ops.ScalarUDF | ops.AggUDF) -> str:
         )
 
     def visit_ScalarUDF(self, op, **kw):
-        return self.f[self.__sql_name__(op)](*kw.values())
+        return self.f[self.__sql_name__(op)](*kw.values(), dialect=self.dialect)
```
cpcloud, line 1237: This isn't the correct place to add the `dialect` argument. What about all the other many places where we're invoking a function from `self.f`? This doesn't just apply to UDFs, right?

cpcloud (changes requested):

ibis/backends/sql/compilers/base.py
NickCrews: Lol, we had the same thought. If you look at my previous attempts on this PR (before the final force push I did), then you will see I started doing exactly what you are doing now. But then I decided it was a whole can of worms, and I was being consistent with what already existed, so then I reduced my scope to get this merged. Can we do the refactor in a followup, and just get this fix in?

cpcloud: It's pretty close, so I'll just wrap it up here.

cpcloud (approved):

ibis/backends/sql/compilers/base.py
NickCrews: Nice job, this looks way better!

PR #10400, opened 2024-10-30
refactor(compiler): fix `Compiler.dialect` type hint, remove unneeded `duckdb` compilation rule for `RegexReplace`

NickCrews:


ibis/backends/sql/compilers/duckdb.py
```diff
@@ -523,9 +523,6 @@ def visit_TimestampNow(self, op):
         """DuckDB current timestamp defaults to timestamp + tz."""
         return self.cast(super().visit_TimestampNow(op), dt.timestamp)
 
-    def visit_RegexReplace(self, op, *, arg, pattern, replacement):
-        return self.f.regexp_replace(arg, pattern, replacement, "g")
```
cpcloud, line 527: I don't think you can remove this. Did you see that the last argument is `"g"`?

cpcloud: Ahhh, I didn't see that we're passing the `"g"` in the base class rule! Carry on!

PR #10401, opened 2024-10-31
docs(examples): add remaining non-TimeValue temporal examples

IndexSeek:
## Description of changes

Add examples for the remaining non-TimeValue-only temporal operations. This time, I opted not to use memtable examples and used the scalar options to demonstrate functionality. 

There's a lot here, so I broke them into separate commits if we wanted to pluck any out that seemed self-explanatory. Admittedly, many of these did feel self-explanatory, but I didn't know if it was still better to have examples for all of them for completeness. I am okay with squashing all of these commits if we want to proceed with these examples if that is better. 

I also used `ibis.date` and `ibis.timestamp` rather than `datetime.date` or `datetime.datetime`. It might be worth standardizing all these examples to use one or the other, but I figured we could discuss that or pick it up in a different PR.

### DIFF
--- a/ibis/expr/types/temporal.py
+++ b/ibis/expr/types/temporal.py
@@ -27,7 +27,16 @@ class _DateComponentMixin:
     """Temporal expressions that have a date component."""
 
     def epoch_seconds(self) -> ir.IntegerValue:
-        """Extract UNIX epoch in seconds."""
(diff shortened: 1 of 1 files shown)


cpcloud (approved):
This is great, thank you!

PR #10402, opened 2024-10-31
docs(api): indicate that `re_search` need only match part of a string

NickCrews:
I am never sure if the pattern needs to match the ENTIRE string, or only part of it. This makes that clear.


cpcloud (approved):
Thanks!

PR #10410, opened 2024-11-01
perf(duckdb): reduce branching factor for generated ArrayDistinct code

NickCrews:
If I do something like `some_really_complex_array_expression.unique()`, then the SQL for `some_really_complex_array_expression` was duplicated 4 times. Now it is only duplicated 3 times.

This is important for perf, because duckdb sometimes appears to not be smart, and actually does a computation for each time a subexpression appears.
See https://github.com/duckdb/duckdb/discussions/14649.
So this can reduce the computation time to 3/4 of what it was.

I want to go through our other compilation steps and do similar optimizations whenever possible.

### DIFF
--- a/ibis/backends/sql/compilers/duckdb.py
+++ b/ibis/backends/sql/compilers/duckdb.py
@@ -143,15 +143,17 @@ class DuckDBCompiler(SQLGlotCompiler):
         )
 
     def visit_ArrayDistinct(self, op, *, arg):
+        return self._array_distinct(arg)
+
+    def _array_distinct(self, arg: sge.Expression) -> sge.Expression:
+        x = sg.to_identifier("x")
+        is_null = sge.Lambda(this=x.is_(NULL), expressions=[x])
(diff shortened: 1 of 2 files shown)


cpcloud (approved):

ibis/backends/sql/compilers/duckdb.py
```diff
@@ -143,15 +143,17 @@ def visit_StructColumn(self, op, *, names, values):
...
+    def _array_distinct(self, arg: sge.Expression) -> sge.Expression:
+        x = sg.to_identifier("x")
+        is_null = sge.Lambda(this=x.is_(NULL), expressions=[x])
+        contains_null = self.f.list_bool_or(self.f.list_apply(arg, is_null))
         return self.if_(
             arg.is_(NULL),
             NULL,
             self.f.list_distinct(arg)
-            + self.if_(
-                self.f.list_count(arg) < self.f.len(arg),
```
cpcloud, line 151: To be totally clear, the new code removes the call to `self.f.len(arg)` which is responsible for getting rid of one of the duplications.

PR #10411, opened 2024-11-01
perf(duckdb): use builtin negative array slicing in newer duckdbs

NickCrews:
Similar rationale to https://github.com/ibis-project/ibis/pull/10410

### DIFF
--- a/ibis/backends/sql/compilers/duckdb.py
+++ b/ibis/backends/sql/compilers/duckdb.py
@@ -182,16 +182,22 @@ class DuckDBCompiler(SQLGlotCompiler):
         arg_length = self.f.len(arg)
 
         if start is None:
-            start = 0
+            start = 1
+        elif isinstance(op.start, ops.Literal):
+            start = op.start.value
+            start += start >= 0
         else:
-            start = self.f.least(arg_length, self._neg_idx_to_pos(arg, start))
+            start = self.if_(start < 0, start, start + 1)
 
         if stop is None:
             stop = arg_length
+        elif isinstance(op.stop, ops.Literal):
+            stop = int(op.stop.value)
+            stop -= stop < 0
         else:
-            stop = self._neg_idx_to_pos(arg, stop)
+            stop = self.if_(stop < 0, stop - 1, stop)
 
-        return self.f.list_slice(arg, start + 1, stop)
+        return self.f.list_slice(arg, start, stop)
 
     def visit_ArrayMap(self, op, *, arg, body, param, index):
         expressions = [param]
--- a/poetry.lock
+++ b/poetry.lock
@@ -8320,4 +8320,4 @@ visualization = ["graphviz"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.10"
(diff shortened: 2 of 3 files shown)


ibis/backends/sql/compilers/duckdb.py
```diff
@@ -177,16 +177,19 @@ def visit_ArrayRepeat(self, op, *, arg, times):
         return self.f.flatten(self.f.list_apply(self.f.range(times), func))
 
     def visit_ArraySlice(self, op, *, arg, start, stop):
+        import duckdb
```
cpcloud, line 180: The compilers **must** be usable without any dependencies installed, so you can't do this.

ibis/backends/sql/compilers/duckdb.py
```diff
@@ -177,16 +177,19 @@ def visit_ArrayRepeat(self, op, *, arg, times):
...
             start = 0
-        else:
-            start = self.f.least(arg_length, self._neg_idx_to_pos(arg, start))
-
         if stop is None:
             stop = arg_length
-        else:
+
+        # TODO: once we only support duckdb >= 0.9.0, we can remove this
+        if duckdb.__version__ < "0.9.0":
```
cpcloud, line 190: We should just bump to 0.9.0 as a lower bound for the next release, then we don't have to futz with with this.

I'll update your PR to make these changes.

cpcloud (changes requested):

cpcloud (approved):

cpcloud:
Well, clearly negative slicing doesn't work as expected 😅

cpcloud:
This should replace things like `-3 + -1` with the literal `-4`.

ibis/backends/sql/compilers/duckdb.py
NickCrews: Ah, didn't know this. Thanks!

ibis/backends/sql/compilers/duckdb.py
```diff
@@ -182,16 +182,22 @@ def visit_ArraySlice(self, op, *, arg, start, stop):
...
 
         if start is None:
-            start = 0
+            start = 1
+        elif isinstance(op.start, ops.Literal):
+            start = op.start.value
+            start += start >= 0
         else:
-            start = self.f.least(arg_length, self._neg_idx_to_pos(arg, start))
+            start = self.if_(start < 0, start, start + 1)
```
NickCrews, line 190: Duckdb why do you have to use such inconsistent indexing semantics 😢

NickCrews:
@cpcloud did you bump the min version to 0.10.0 because I didn't read duckdbs changelog/got blame correctly, and 0.9.0 wasn't adequate?

cpcloud:
> @cpcloud did you bump the min version to 0.10.0 because I didn't read duckdbs changelog/got blame correctly, and 0.9.0 wasn't adequate?

0.10 is the first release that officially supports the backwards compatible 1.0 storage format, and it didn't seem like bumping to just 0.9 was worth it given that bumping to either is a breaking change from the previous lower bound, so I bumped to 0.10.

NickCrews:
Sounds reasonable! If someone squawks that they need 0.9.0 support we can revisit

PR #10412, opened 2024-11-01
feat(api): add FieldNotFoundError

NickCrews:
I have been getting sick of typing some_table_or_struct.field_that_doesnt_exist_or_has_a_small_typo and then getting a useless error message. This PR makes that UX much better.

Still need to add tests, but I wanted to get this up here for some initial thoughts before I invested more time. Is this something we want to pursue?


ibis/expr/types/relations.py
```diff
@@ -240,21 +240,39 @@ def _fast_bind(self, *args, **kwargs):
                 args = ()
             else:
                 args = util.promote_list(args[0])
-        # bind positional arguments
+
         values = []
+        errs = []
+        # bind positional arguments
         for arg in args:
```
NickCrews, line 247: This is somewhat orthogonal to the FieldNotFoundError stuff. But if I have multiple bogus columns, eg `table.select("bogus", "also_bogus")` I only see the first error.

ibis/expr/types/relations.py
```diff
@@ -240,21 +240,39 @@ def _fast_bind(self, *args, **kwargs):
...
+                errs.append(e)
+                continue
             if len(bindings) != 1:
                 raise com.IbisInputError(
                     "Keyword arguments cannot produce more than one value"
                 )
             (value,) = bindings
             values.append(value.name(key))
+        if errs:
+            raise com.IbisError(
```
NickCrews, line 272: not sure what the type of this should be.

ibis/expr/types/relations.py
```diff
@@ -739,8 +757,9 @@ def __getattr__(self, key: str) -> ir.Column:
         """
         try:
             return ops.Field(self, key).to_expr()
-        except com.IbisTypeError:
-            pass
+        except com.FieldNotFoundError as e:
```
NickCrews, line 760: There is a slight difference in ux here that I'd love to unify
- if I do `Table.totally_bogus`, then I get `AttributeError: 'Table' object has no attribute 'bogus'`
- if I do `Table["totally_bogus"]`,  then I get `FieldNotFoundError: 'bogus' not found in Table object. Possible options: {'x'}`

I'm not sure which is better. If we say `Possible options:...` then that only includes the field names, and misses all the Table methods. But, all methods are 1. in the docs and 2. should have tab completion in many cases, so I bet typos are a lot more likely on column typos than method typos. So I think the FieldNotFoundError with the suggestion might be better.

NickCrews: Also a difference between Tables and Structs. I'd love for them to have the same behavior too.

ibis/expr/types/structs.py
```diff
@@ -205,7 +205,7 @@ def __getitem__(self, name: str) -> ir.Value:
         KeyError: 'foo_bar'
         """
         if name not in self.names:
-            raise KeyError(name)
+            raise FieldNotFoundError(self, name, self.names)
```
NickCrews, line 208: I think it's OK to be breaking here and not raise a KeyError anymore?

NickCrews: KeyError seems semantically a little wrong. I think it should be for collections with a dynamic set of keys. But structs have a static set of keys, so FieldNotFoundError, as a subclass of AttributeError, seems better to me.

NickCrews:
@cpcloud does this look like an idea worth considering? Any behavior requirements that you think I would need to make an implementation satisfy?

NickCrews:
@cpcloud should I invest more time in this or do you think this whole idea is not the right direction?

NickCrews:
That sounds like good behavior to me to return an AttributeError for `.dot` accesses. I think that makes sense to "refuse the temptation to guess" and raise an AttributeError and not a FieldNotFoundError, because that is better to be caught programmatically. But, what about being a **little** bit fancy/helpful, and returning a `TableAttributeError`, which is a subclass of `AttributeError`, which has the helpful "did you mean this column?" error message? Then this is the best of both worlds?

There is one bit of nuance on attribute access though. If I do `t.bogus`, I agree that should just be the above behavior, it is impossible to guess if the user meant a method or a column, so just return an AttributeError (or subclass of it as I suggest above). However, if I do `t.select(_.bogus)`, then this is unambiguous that they wanted a Column, and so returning a FieldNotFoundError seems better.

In pseudocode:

```python
# Raised on t.bogus
# Includes helpful suggestion "did you mean <best match among all columns AND attributes/methods>"
class TableAttributeError(AttributeError, IbisError): ...

# Raised on my_struct.bogus
# Includes helpful suggestion "did you mean <best match among all fields AND attributes/methods>"
# Maybe should get consolidated with TableAttributeError
class StructAttributeError(AttributeError, IbisError): ...

# Raised on table_or_struct["bogus", "bogus2"] and table_or_struct.select(_.bogus, _.bogus2) and my_struct["bogus"]
# Includes helpful suggestion "did you mean <best match among ONLY columns/fields>"
# Note that this contains ALL not found fields.
# I think that would be simpler to just have one error, and not one for single FieldNotFound.
# Should this inherit from AttributeError and/or KeyError?
# I think AttributeError since 1. these are statically known and 2. then a user can `catch AttributeError`
# and have all of these errors caught.
class FieldsNotFoundError(AttributeError, IbisError): ...
 ```

What do you think of this spec?

**Comment by NickCrews** [comment]
> and do nothing fancy
> we should really refuse the temptation to guess

So I understand your goals, is your reasoning here so that 1. our maintenance is easier or 2. so we don't accidentally do the wrong thing for the user?

**Comment by NickCrews** [comment]
Sounds good, thanks, I'll try to update this implementation and let's see what we think about it!

File: ibis/expr/types/relations.py (hunk 3 above)
Comment (NickCrews) [reply ibis/expr/types/relations.py]: I think we decided that for now, the `Table.totally_bogus` SHOULD give a plain AttributeError for now. We can maybe extend it later.

File: ibis/expr/types/relations.py (hunk 2 above)
Comment (NickCrews) [reply ibis/expr/types/relations.py]: no longer relevant, this now raises a FieldsNotFoundError.

File: ibis/expr/types/relations.py (hunk 1 above)
Comment (NickCrews) [reply ibis/expr/types/relations.py]: This is implemented as FieldsNotFoundError

**Comment by NickCrews** [comment]
Not sure what is causing [the postgres failure](https://github.com/ibis-project/ibis/actions/runs/13428575878/job/37520654230#step:12:1) but I think it's unrelated.

File: ibis/common/exceptions.py [hunk 5]
```diff
@@ -45,6 +48,37 @@ class RelationError(ExpressionError):
     """RelationError."""
 
 
+class FieldsNotFoundError(AttributeError, IbisError):
+    """When you try to access `table_or_struct.select("foo", "bar")` or `table_or_struct["foo"]`."""
+
+    def __init__(
+        self,
+        container: object,
+        names: str | Iterable[str],
```
NickCrews, line 57: Not totally sold on the names of these arguments/attributes.

ibis/expr/types/structs.py
```diff
@@ -200,13 +200,13 @@ def __getitem__(self, name: str) -> ir.Value:
...
+        >>> t.s["foo_bar"]  # doctest: +ELLIPSIS
         Traceback (most recent call last):
-            ...
-        KeyError: 'foo_bar'
+        ...
+        ibis.common.exceptions.FieldsNotFoundError: 'foo_bar' not found in StructColumn object. Possible options: ('a', 'b')
         """
         if name not in self.names:
-            raise KeyError(name)
+            raise FieldsNotFoundError(self, name, self.names)
```
NickCrews, line 209: We didn't have a test for` Struct.__getitem` before, but if we had, then we would see that this change is breaking. Before `Struct["foo_bar"]` raised a KeyError, now it raises a FieldsNotFoundError. I think something that derives from AttributeError is more semantically correct: we know what fields are present at compile time, whereas I think of KeyError I think of for when you have a runtime-missing key. IDK if this is standard.

ibis/tests/expr/test_table.py
```diff
@@ -546,13 +568,13 @@ def test_order_by_scalar(table, key, expected):
...
     [
-        ("bogus", com.IbisTypeError),
-        (("bogus", False), com.IbisTypeError),
-        (ibis.desc("bogus"), com.IbisTypeError),
-        (_.bogus, AttributeError),
-        (_.bogus.desc(), AttributeError),
+        "bogus",
+        ("bogus", False),
+        ibis.desc("bogus"),
+        _.bogus,
```
NickCrews, line 576: FieldsNotFoundError is a subclass of AttributeError, so this is only breaking for the test cases that looked for IbisTypeError.

NickCrews:
As you can see in the updated tests, this change is potentially breaking for users who are trying to catch IbisTypeError. But, it does unify these two errors so they only need to catch a FieldsNotFoundError, so that is a plus. I think this breakage is OK. We could make FieldsNotFoundError subclass from BOTH AttributeError and IbisTypeError? IDK, I could see how table.bogus is a sort of TypeError.

ibis/common/exceptions.py
```diff
@@ -45,6 +48,37 @@ class RelationError(ExpressionError):
...
+        existing_options: Iterable[str],
+    ) -> None:
+        self.container = container
```
cpcloud, line 60:
```suggestion
```

This field seems unused. If so, you can remove it as an instance attribute.

ibis/tests/expr/test_struct.py
```diff
@@ -41,7 +42,30 @@ def test_struct_getattr():
...
+    assert expr["a"].get_name() == "a"
+
+    with pytest.raises(TypeError, match="Deferred"):
+        expr[_.A]
+
+    with pytest.raises(com.FieldsNotFoundError, match="A") as excinfo:
+        expr["A"]
+    assert isinstance(excinfo.value, AttributeError)
+
+    assert excinfo.value.container is expr
```
cpcloud, line 60: You're using `container` in tests here, but that doesn't seem like a good reason to store an expression as an attribute on an exception.

ibis/common/exceptions.py
```diff
@@ -45,6 +48,37 @@ class RelationError(ExpressionError):
...
+                low2orig[low]
+                for low in difflib.get_close_matches(name.lower(), low2orig.keys())
+            )
+            if len(typos) == 1:
+                msg = f"'{name}' not found in {container.__class__.__name__} object. Did you mean '{next(iter(typos))}'?"
+            elif len(typos) > 1:
+                msg = f"'{name}' not found in {container.__class__.__name__} object. Did you mean one of {typos}?"
+            else:
+                msg = f"'{name}' not found in {container.__class__.__name__} object. Possible options: {self.existing_options}"
+            self.suggestions[name] = typos
```
cpcloud, line 77: This also seems like it's only being used in tests, so I'd remove it.

cpcloud (changes requested):

ibis/common/exceptions.py
```diff
@@ -45,6 +48,37 @@ class RelationError(ExpressionError):
     """RelationError."""
 
 
+class FieldsNotFoundError(AttributeError, IbisError):
```
cpcloud, line 51: Why did you decide to subclass `AttributeError` here?

ibis/expr/types/relations.py
```diff
@@ -682,13 +698,18 @@ def __getitem__(self, what: str | int | slice | Sequence[str | int]):
...
+            if isinstance(what, str):
+                return ops.Field(self.op(), what).to_expr()
+            elif isinstance(what, int):
+                return ops.Field(self.op(), self.columns[what]).to_expr()
+            elif isinstance(what, slice):
+                limit, offset = util.slice_to_limit_offset(what, self.count())
+                return self.limit(limit, offset=offset)
+        except com.FieldsNotFoundError:
+            # the raised FieldsNotFoundError contains the Op, but we want
+            # to only expose the Expr.
```
cpcloud, line 711: Let's avoid storing unused state on exceptions. This can have unintended consequences for object lifetimes due to exceptions keeping stack frames around for tracebacks.

Please remove any unused fields from your new exception types.

NickCrews: Oh wow, I'd never thought about how that might effect lifetimes, that's really interesting. Sounds good, I am removing them. I originally had them because in an earlier version I did actually need to keep track of them, but now I don't need that, so I'll stop storing them.

ibis/tests/expr/test_table.py
NickCrews: EDIT: now that I made FieldsNotFound NOT a subclass of Attribute error, this is now breaking for all of these test cases.

ibis/common/exceptions.py
NickCrews: This was because it originally was a `FieldNotFoundError` (singular), so it made more sense in the sense of "you tried to lookup some statically known thing on this object that doesn't exist". But now that it stores multiple missing fields, the heirarchy doesn't make sense, so I removed AttributeError as a base class, now it is just inherits from IbisError.

ibis/expr/types/structs.py
NickCrews: EDIT: now FieldsNotFoundError derives ONLY from IbisError, and neither KeyError nor AttributeError.

NickCrews:
@cpcloud this is ready for another review whenever you get the chance. Thank you!

NickCrews:
@cpcloud is there anything I can do to reduce the effort to review this? want to chat about this change live in eg zulip?

NickCrews:
@cpcloud any way we can get this in as a breaking change for 11.0?

cpcloud:
Yes, we can definitely get this in for 11.

PR #10420, opened 2024-11-02
docs(examples): add type example

IndexSeek:
## Description of changes

Adds a few examples given different data types and the the output of `column.type`. I wanted to include the timestamp here since it shows the output class of that type with more than nullable. 

Here's a doc preview:

image

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -88,7 +88,30 @@ class Value(Expr):
 
     # TODO(kszucs): should rename to dtype
     def type(self) -> dt.DataType:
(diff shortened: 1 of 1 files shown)


IndexSeek:
I noticed a TODO on type to redefine this as `dtype` rather than `type`. @kszucs, I'm happy to take on this effort in a separate PR to deprecate `type` in favor of `dtype` if you would like. If we go that route, is there potential for confusion with `ibis.dtype()`?

cpcloud (approved):
Thanks, keep 'em coming!

PR #10421, opened 2024-11-02
docs(examples): add remaining selectors examples

IndexSeek:
## Description of changes

This adds examples for the remaining selectors except `s.none()`. I need to brainstorm for a good example of that one, unless anyone has ideas off the bat.

Selectors may still be my favorite Ibis feature, so I am happy to provide direct examples for using them. 😎

### DIFF
--- a/ibis/selectors.py
+++ b/ibis/selectors.py
@@ -274,6 +274,15 @@ def endswith(suffixes: str | tuple[str, ...]) -> Selector:
     suffixes
         Suffixes to compare column names against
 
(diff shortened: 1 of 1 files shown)


cpcloud:
Thanks! The `s.none()` selector has a test case that could be useful, but it will probably not be possible to showcase its utility outside the context of a complex tabular operation like `pivot_wider`/`pivot_longer`

cpcloud (approved):

PR #10424, opened 2024-11-03
chore(dev-deps): move to `uv` for development

cpcloud:
## Description of changes

This PR moves our dependency management over to [`uv`](https://github.com/astral-sh/uv).

~Wanted to get a head start on CI and see what shakes out.~

### TODOs:

- [x] ~[`uv2nix` issue](https://github.com/adisbladis/uv2nix/pull/46) that needs to be fixed.~ This is fixed
- [x] preserve existing dev versioning scheme (~right now it's `major.post$N.dev0`~).
- [x] ~nix package builds (that's what runs tests in the `nix.yml` GHA workflows) aren't working at all right now~ not working the way I'd like, but tests are being run similarly to how they were under poetry2nix.
- [x] release scripts need to be ported, but I think I can mostly keep them the same


cpcloud:
cc @grieve54706 if you want to give this PR a try

cpcloud:
cc @gforsyth and @kszucs for thoughts here

grieve54706:
Thanks for pinging me.
I'm unsure how to test this PR. Should I also adjust my `pyproject.toml` for `uv` and then install Ibis?

kszucs:
I am in favor of the change. Do we need to update the guides/docs as well?

kszucs (approved):
LGTM, note that I haven't been looking at the nix changes.

cpcloud:
> I am in favor of the change. Do we need to update the guides/docs as well?

Yep, lemme do that.

cpcloud:
@kszucs Guides are now updated.

cpcloud:
Ok, I've somewhat improved the nix builds. I'm still not sure if the way I did it is 100% kosher, but now the nix builds have the same behavior as before: they're testing ibis as part of a package build.

nix/ibis.nix
```diff
@@ -1,61 +1,46 @@
...
-
-  pythonImportsCheck = [ "ibis" ] ++ map (backend: "ibis.backends.${backend}") (backends ++ [ "datafusion" ]);
+  # Ibis-framework was already built as a part of the env, this is just running
+  # tests. Symlink the built test env for convenience.
+  #
+  # Note: Testing could technically be done as a part of the virtualenv
+  # constructor derivation.
+  installPhase = ''
+    runHook preInstall
+    cp -r ${pythonEnv} $out
```
adisbladis, line 43: Symlinking this should work with `dontFixup = true`.

cpcloud: Ah, right, thank you!

cpcloud: Fixed!

cpcloud:
Merging, so that I can work out any kinks in the cloud CI.

PR #10425, opened 2024-11-03
docs(examples): add selector example for none

IndexSeek:
## Description of changes

Adds an example of using the `s.none()` selector based on discussion in https://github.com/ibis-project/ibis/pull/10421. This is the last column selector needing an example on the [docs page](https://ibis-project.org/reference/selectors).


gforsyth (approved):
Thanks @IndexSeek !

Issue #6974, opened 2023-08-30
feat: Better tempfile deletion


NickCrews:
### Is your feature request related to a problem?

I have slowly been running out of disk space on my laptop over the last few months. I just investigated, and it was due to 32Gb of temp files in `/Users/nickcrews/.cache/ibis-duckdb/`. This must have been due to repeated process crashing and the temp files never getting cleaned up.

### Describe the solution you'd like

Ideally, these files get cleaned up automatically.

relevant code is at https://github.com/ibis-project/ibis/blob/d4ea3bf05a77a291b209c288d60dcc806cae7f75/ibis/backends/duckdb/__init__.py#L205-L210

Why not make the temp directory be under `/tmp`, which if I understand correctly is wiped at every boot? Really, I think these temp files should get cleaned up at the end of each process, right, since duckdb is in-process? I think XDG_CACHE_HOME is meant for between-run caches, like `pip` might use.

### What version of ibis are you running?

main

### What backend(s) are you using, if any?

duckdb, but perhaps other backends also suffer?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

lostmygithubaccount:
I did one time have 1.4 TB in there (silly cross join I let run for hours)

cpcloud:
> Why not make the temp directory be under `/tmp`, which if I understand correctly is wiped at every boot?

It's complicated 😄

`/tmp` is only wiped at boot if your system uses tmpfs for `/tmp` which isn't always the case. You can use whatever file system you want to back `/tmp`. We can't really assume it's anything about the lifetime of the bytes in `/tmp`.

The primary reason to not use `/tmp`-backend-by-tmpfs is because it is [backed by virtual memory](https://www.kernel.org/doc/html/latest/filesystems/tmpfs.html#tmpfs), so you're not actually getting spill-to-disk behavior by using it.

Finally, the system temporary directory can pretty much be anything, it doesn't have to be `/tmp`.

These are the main reasons we chose to use a location that is very likely to be actual hard disk.

As to the issue of disk space filling up, I suspect it's a DuckDB issue: ibis simply sets the location of the temporary directory, but has no hand in its cleanup.

Do you have an example where the data are spilled to disk and not clean up on a successful exit of the process?

cpcloud:
Regarding unsuccessful exits of the Python process: that's yet more complicated and depends on what caused the shutdown of the program. If it's a signal then that depends on what signal it is 😄.

[...]

PR #10433, opened 2024-11-04
refactor(duckdb): pass `temp_directory` through as-is to `duckdb.connect`

NickCrews:
Let's make the pass-through from ibis to duckdb more transparent, so you get closer to the same behavior as when using `import duckdb; duckdb.connect()`.

I considered also removing the nicety of creating the temp_dir for users, but decided to keep it.
Now, we only create parent. duckdb will make dir itself only if it needs. This means that for low-memory loads,
a dir might never be needed. This better matches duckdb's behavior.
Also, adjust our test to really check for the end desired behavior, not an implementation detail
of if WE create the dir.

Related to https://github.com/ibis-project/ibis/issues/6974

BREAKING CHANGE: Special handling of the `temp_directory` argument passed to Ibis is removed in favor of passing the argument through directly to `duckdb.connect`. Interior nodes of directory trees must be created, e.g., using `Path.mkdir(exists_ok=True, parents=True)`, `mkdir -p`, etc.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import ast
 import contextlib
-import os
 import urllib
 import warnings
 from operator import itemgetter
@@ -362,7 +361,6 @@ class Backend(SQLBackend, CanCreateDatabase, UrlFromPath):
         self,
         database: str | Path = ":memory:",
         read_only: bool = False,
-        temp_directory: str | Path | None = None,
         extensions: Sequence[str] | None = None,
         **config: Any,
     ) -> None:
@@ -374,9 +372,6 @@ class Backend(SQLBackend, CanCreateDatabase, UrlFromPath):
             Path to a duckdb database.
         read_only
             Whether the database is read-only.
-        temp_directory
-            Directory to use for spilling to disk. Only set by default for
-            in-memory connections.
         extensions
             A list of duckdb extensions to install/load upon connection.
         config
@@ -394,19 +389,7 @@ class Backend(SQLBackend, CanCreateDatabase, UrlFromPath):
             ("md:", "motherduck:", ":memory:")
         ):
             database = Path(database).absolute()
-
-        if temp_directory is None:
-            temp_directory = (
-                Path(os.environ.get("XDG_CACHE_HOME", Path.home() / ".cache"))
-                / "ibis-duckdb"
-                / str(os.getpid())
-            )
-        else:
-            Path(temp_directory).mkdir(parents=True, exist_ok=True)
-            config["temp_directory"] = str(temp_directory)
-
         self.con = duckdb.connect(str(database), config=config, read_only=read_only)
-
         self._post_connect(extensions)
 
     @util.experimental
--- a/ibis/backends/duckdb/tests/test_register.py
+++ b/ibis/backends/duckdb/tests/test_register.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import os
 import sqlite3
-import tempfile
 from pathlib import Path
 
 import duckdb
@@ -135,32 +134,6 @@ def test_read_json(con, data_dir, tmp_path):
     nrows = pqt.count().execute()
     assert nrows
     assert nrows == jst.count().execute()
-
-
-def test_temp_directory(tmp_path):
-    query = "SELECT current_setting('temp_directory')"
-
-    # 1. in-memory + no temp_directory specified
-    con = ibis.duckdb.connect()
-
-    value = con.raw_sql(query).fetchone()[0]
-    assert value  # we don't care what the specific value is
-
-    temp_directory = Path(tempfile.gettempdir()) / "duckdb"
-
-    # 2. in-memory + temp_directory specified
-    con = ibis.duckdb.connect(temp_directory=temp_directory)
-    value = con.raw_sql(query).fetchone()[0]
-    assert value == str(temp_directory)
-
-    # 3. on-disk + no temp_directory specified
-    # untested, duckdb sets the temp_directory to something implementation
-    # defined
-
-    # 4. on-disk + temp_directory specified
-    con = ibis.duckdb.connect(tmp_path / "test2.ddb", temp_directory=temp_directory)
-    value = con.raw_sql(query).fetchone()[0]
-    assert value == str(temp_directory)
 
 
 @pytest.fixture(scope="session")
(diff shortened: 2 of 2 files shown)


ibis/backends/duckdb/__init__.py
```diff
@@ -394,19 +389,10 @@ def do_connect(
...
-                Path(os.environ.get("XDG_CACHE_HOME", Path.home() / ".cache"))
-                / "ibis-duckdb"
-                / str(os.getpid())
-            )
-        else:
-            Path(temp_directory).mkdir(parents=True, exist_ok=True)
-            config["temp_directory"] = str(temp_directory)
-
+        if temp_directory := config.get("temp_directory"):
+            # Only create parent. duckdb will make dir itself only if it needs.
```
cpcloud, line 393: Is this behavior documented somewhere? We were doing this because at the time, DuckDB did **not** create a temp directory for in-memory connections unless the user explicitly specified one.

cpcloud (changes requested):

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -383,8 +383,10 @@ def test_memtable_with_nullable_pyarrow_not_string(con):
 
 def test_set_temp_dir(tmp_path):
     path = tmp_path / "foo" / "bar"
-    ibis.duckdb.connect(temp_directory=path)
-    assert path.exists()
+    con = ibis.duckdb.connect(temp_directory=path, max_memory="1MB")
+    con.raw_sql("CREATE TABLE big AS SELECT range(100_000) as i")
```
cpcloud, line 387: Ah, I see.

Thanks for adding this test, I agree this is much better than the previous one.

ibis/backends/duckdb/__init__.py
```diff
@@ -394,19 +389,10 @@ def do_connect(
             ("md:", "motherduck:", ":memory:")
         ):
             database = Path(database).absolute()
-
-        if temp_directory is None:
```
cpcloud, line 398: I don't think this is a good idea to get rid of.

You're effectively making it so that in-memory connections no longer work on larger-than-memory data unless `temp_directory` is passed. That seems strictly worse than being inconsistent with `duckdb.connect`.

cpcloud (changes requested):

ibis/backends/duckdb/__init__.py
cpcloud: In other words, your new test should pass without needing to provide the `temp_directory` argument.

NickCrews: According to https://duckdb.org/docs/configuration/overview.html#global-configuration-options, it says the default temp_directory is "⟨database_name⟩.tmp or .tmp (in in-memory mode)", which makes me think that either way, duckdb is gonna make the temp dir

NickCrews: oh, just saw max_temp_directory_size. yeah I will add a test that checks that if you
- do not pass temp dir
- set the memory limit low and do some computation
- then we expect `./.tmp/` to be created and have some content

cpcloud: Oh, nice, it looks like it's now behaving as desired :)

In that case, can we remove the option entirely?

NickCrews: Is that the only other behavior that we want to test for? Or is there another test case I should add that we want to pass?

cpcloud: Actually we can't, because duckdb won't create intermediate directories

cpcloud: Personally, I'd much rather have duckdb pooping files in the XDG-standard place (usually `~/.cache/$APPLICATION`), rather than in the current directory.

cpcloud: If we're aiming for maximum transparency then we should literally just pass the behavior through and verify that out-of-core processing works.

NickCrews: hmm, lol just a matter of taste I think, I would much rather temp files be in the current dir than somewhere else.

I think I would love to just wash our hands of it and pass through the kwargs transparently and not get bogged down. I just pushed some tests, but they are failing for some weird `duckdb.duckdb.TransactionException: TransactionContext Error: Failed to commit: failed to pin block of size 256.0 KiB (824.0 KiB/976.5 KiB used)` error that I don't want to have to work around.

NickCrews: Like I'm not sure it should be obvious what to do. I could see arguments either way that we should 1. NOT do any magic creating dirs for people or 2. we definitely should create the needed intermediate dirs. Plenty of software does both, I think either would be sane.

cpcloud: I think we should remove any special duckdb temp_directory handling and use consolidate on the upstream behavior.

cpcloud: Effectively that should mean: remove all variable references to `temp_directory` and just let the argument pass through to the `duckdb.connect` call as part of `**kwargs`.

NickCrews: Sounds good to me. Keep it simple. I'm gonna be out of internet for about a week, I can make those changes when I get back.

cpcloud: I'll fix it up and wrap up the PR, thanks for getting it started!

ibis/backends/duckdb/tests/test_register.py
```diff
@@ -381,12 +354,45 @@ def test_memtable_with_nullable_pyarrow_not_string(con):
...
+    con.raw_sql("SET memory_limit = '1MB'")
+    con.raw_sql("CREATE TABLE big AS SELECT range(1_000_000) as i")
+
+
+def test_temp_dir_unset_inmemory():
+    con = ibis.duckdb.connect()
+    p = _query_temp_dir_setting(con)
+    p.unlink(missing_ok=True)
+    _trigger_oom(con)
+    assert list(p.iterdir())
```
cpcloud, line 372: These tests are incredibly flaky, since the behavior of setting the memory limit depends **greatly** on what else is running on the machine.

The best we can do here is to verify that `temp_directory` was set, but that's kind of pointless.

We should 100% avoid adding a bunch of tests that are verifying upstream default behavior.

cpcloud: I'm going to remove most of these tests. I'll keep that ones that explicitly set the value of `temp_directory` to make sure that actually is being set on the pass-through. Everything else will be removed.

cpcloud (changes requested):

cpcloud (approved):

PR #10437, opened 2024-11-05
fix(mssql): escape special characters in passwords

grieve54706:
## Description of changes
I found the error 
```
'IM002', '[IM002] [unixODBC][Driver Manager]Data source name not found and no default driver specified (0) (SQLDriverConnect)'
```
Because the password of mssql includes special characters like `{R;3G1/8Al2AniRye` that start with `{` or include `;`.
It should be covered by `{` and `}` and replace `}` with`}}`.

Reference:
https://github.com/mkleehammer/pyodbc/wiki/Connecting-to-databases
https://stackoverflow.com/questions/78531086/pyodbc-connection-string-correctly-escaping-password-with-special-characters/78532507#78532507

### DIFF
--- a/ibis/backends/mssql/__init__.py
+++ b/ibis/backends/mssql/__init__.py
@@ -176,12 +176,16 @@ class Backend(SQLBackend, CanCreateCatalog, CanCreateDatabase):
         self.con = pyodbc.connect(
             user=user,
             server=f"{host},{port}",
-            password=password,
+            password=self._escape_special_characters(password),
             driver=driver,
             **kwargs,
         )
 
         self._post_connect()
+
+    @staticmethod
+    def _escape_special_characters(value: str) -> str:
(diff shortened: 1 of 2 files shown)


gforsyth:
Hey @grieve54706 -- thanks for putting this in.  What would happen if a user passed in a properly escaped password directly?  We should make sure that this works in that case, too.

As for testing, I would recommend a test just of the escaping function, ensuring that unescaped strings are properly escaped, and properly escaped strings are left untouched

grieve54706:
Thanks @gforsyth for your reply.
We don't need to escape `{`. We can find this message 
```
occurrences of `{` inside a quoted value can be as-is, since the end of a quoted value is indicated by an unescaped `}`.
```
in [wiki](https://github.com/mkleehammer/pyodbc/wiki/Connecting-to-databases) of pyodbc

PR #10440, opened 2024-11-05
fix(duckdb): handle null propagation correctly in array concatenation

cpcloud:
Fixes a bug in NULL handling of our DuckDB implementation of `ops.ArrayConcat`. Previously NULL concatted with non-null returned the non-null value. This PR changes the behavior to return NULL if any of the lists being concatenated are null.

### DIFF
--- a/ibis/backends/sql/compilers/duckdb.py
+++ b/ibis/backends/sql/compilers/duckdb.py
@@ -349,8 +349,12 @@ class DuckDBCompiler(SQLGlotCompiler):
         )
 
     def visit_ArrayConcat(self, op, *, arg):
-        # TODO(cpcloud): map ArrayConcat to this in sqlglot instead of here
-        return reduce(self.f.list_concat, arg)
+        return reduce(
+            lambda x, y: self.if_(
+                x.is_(NULL).or_(y.is_(NULL)), NULL, self.f.list_concat(x, y)
+            ),
+            arg,
+        )
 
     def visit_IntervalFromInteger(self, op, *, arg, unit):
         if unit.short == "ns":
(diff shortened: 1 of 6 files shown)


cpcloud:
This is probably going to live forever because of https://github.com/duckdb/duckdb/issues/11781 and the desire for postgres compatibility.

cc @NickCrews for a heads up here.

cpcloud:
Ok, it looks like we should be able to have our cake and eat it too by using the `||` in the duckdb case.

cpcloud:
In the meantime, I'll use Mark's [suggested macro](https://github.com/duckdb/duckdb/issues/14692#issuecomment-2457146174).

cpcloud:
I tested this against DuckDB 1.1.3 and it behaves as expected/desired.

cpcloud:
Unfortunately the macro is a no-go because we need its functionality to work for read-only databases and macros cannot be created when `read_only=True`.

NickCrews:
Lol I just found the same thing when trying to optimize our compilation of ArrayConcat (similar to my other small perf PRs), and filed https://github.com/duckdb/duckdb/issues/14663, basically the same bug as you.

I'll read this PR more when I get to my computer, but thanks for the work here!

ibis/backends/sql/compilers/duckdb.py
```diff
@@ -349,8 +349,12 @@ def visit_UnwrapJSONBoolean(self, op, *, arg):
         )
 
     def visit_ArrayConcat(self, op, *, arg):
-        # TODO(cpcloud): map ArrayConcat to this in sqlglot instead of here
-        return reduce(self.f.list_concat, arg)
+        return reduce(
```
NickCrews, line 352: @cpcloud FYI in [new duckb releases](https://github.com/duckdb/duckdb/discussions/14685) we should be able to do list_concat(a, b, c, d, ...), but we probably want to wait a bit before switching to that so people can still use older duckdb versions.

PR #10447, opened 2024-11-06
chore: add todos on switching to newer duckdb list_concat() syntax

NickCrews:
I want these changes to eventually make it in

### DIFF
--- a/ibis/backends/sql/compilers/duckdb.py
+++ b/ibis/backends/sql/compilers/duckdb.py
@@ -349,6 +349,10 @@ class DuckDBCompiler(SQLGlotCompiler):
         )
 
     def visit_ArrayConcat(self, op, *, arg):
+        # TODO: duckdb 1.2 was the first release to fix NULL handling.
+        # Once that is our minimum supported duckdb version, we can
+        # use `x || y || z ...` and drop the ifs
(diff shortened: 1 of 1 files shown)


cpcloud:
We can't do this because `list_concat` will always follow the postgres `array_cat` behavior, which is **NOT** to propagate `NULL`s.

ibis/backends/sql/compilers/duckdb.py
```diff
@@ -349,7 +349,14 @@ def visit_UnwrapJSONBoolean(self, op, *, arg):
         )
 
     def visit_ArrayConcat(self, op, *, arg):
+        # TODO: duckdb 1.1.3 was the first release to support list_concat(a,b,c,d,...)
```
cpcloud, line 352: We can't do this because list_concat will always follow the postgres array_cat behavior, which is NOT to propagate NULLs.

cpcloud (changes requested):

ibis/backends/sql/compilers/duckdb.py
NickCrews: Oops, yes that's correct. I will fix these up

cpcloud (approved):

PR #10449, opened 2024-11-07
docs(examples): add usage examples for `get_backend` and `get_name`

IndexSeek:
## Description of changes

Adds examples by creating a table using an in-memory duckdb connection for [`get_backend`](https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.get_backend) and [`get_name`](https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.get_name). 

I did `backend.name` to get "duckdb" in the example for `get_backend` rather than `backend` to avoid getting something like "<ibis.backends.duckdb.Backend at 0x7fffd919b470>".  I'm not sure if both that and the `type` usage are necessary; maybe we could use one or the other if we don't want to use both.

### DIFF
--- a/ibis/expr/types/core.py
+++ b/ibis/expr/types/core.py
@@ -383,6 +383,16 @@ class Expr(Immutable, Coercible):
         BaseBackend
             The Ibis backend.
 
+        Examples
+        --------
+        >>> import ibis
+        >>> con = ibis.duckdb.connect()
+        >>> t = con.create_table("t", {"id": [1, 2, 3]})
+        >>> backend = t.get_backend()
+        >>> backend.name
+        'duckdb'
+        >>> type(backend)
+        <class 'ibis.backends.duckdb.Backend'>
         """
         return self._find_backend(use_default=True)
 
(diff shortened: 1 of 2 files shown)


cpcloud:
You can use `  # doctest: +ELLIPSIS` (note the two spaces preceding the `#`) when dealing with values that can vary between invocations of Python:

```pycon
>>> class Foo: pass  # doctest: +ELLIPSIS
>>> Foo()
<__main__.Foo at ...>
```

cpcloud (approved):

IndexSeek:
> You can use ` # doctest: +ELLIPSIS` (note the two spaces preceding the `#`) when dealing with values that can vary between invocations of Python:
> 
> ```
> >>> class Foo: pass  # doctest: +ELLIPSIS
> >>> Foo()
> <__main__.Foo at ...>
> ```

This is super useful; thank you for sharing that trick! I will keep this in mind for future scenarios where it might be cleaner to show the instance of the class itself.

PR #10454, opened 2024-11-08
docs(examples): add unbind example

IndexSeek:
## Description of changes

Adds an example demonstrating that given a table in both DuckDB and Polars, we can take the table from DuckDB, unbind it, and then execute it using the Polars backend.


cpcloud (approved):
Thanks!

PR #10457, opened 2024-11-08
docs(examples): add `having` example for GroupedTable

IndexSeek:
## Description of changes

This adds a [`having`](https://ibis-project.org/reference/expression-tables#ibis.expr.types.groupby.GroupedTable.having) example for usage on a `GroupedTable`. I followed it up with an `aggregation` to show the removal of the "a" rows. 

It could be useful also to show the `having` method without aggregation results in a `GroupedTable,` but I wasn't 100% sure if that would be worth showing.


gforsyth (approved):
Thanks @IndexSeek !

PR #10466, opened 2024-11-10
fix(mysql): handle database names that must be quoted in `list_tables`

cpcloud:
- **chore: remove some uses of sg.exp for sge**
- **test(mysql): use a dash in the database name**
- **fix(mysql): handle database names that must be quoted in `list_tables`**


cpcloud:
I can alternatively just create a one-off database for testing the `list_tables` call.

cpcloud:
The only downside there is that we wouldn't be verifying the whole test suite against a funkier name.

gforsyth:
Nah, I think this is good.  Same idea as testing against a case-sensitive collation in mssql -- make everything work in the harder scenario

gforsyth (approved):
:shipit:

PR #10473, opened 2024-11-12
chore(deps): implement Nix testing as separate derivation from env

adisbladis:
## Description of changes

I recently wrote documentation on [how to do testing with uv2nix](https://adisbladis.github.io/uv2nix/usage/testing.html).

This PR brings the testing setup of ibis more in line with these docs.

Changes done:
- Move tests to passthru.tests

- Expose tests as flake checks

- Build tests as a separate pipeline step Separates building the environment from tests.

- Reduce the size of the `ibis*` non-devenv attributes. Testing dependencies are no longer added to the main attributes.


cpcloud (approved):
Wonderful, this is great! Thank you.

PR #10480, opened 2024-11-12
test(duckdb): add test for chunk_size param in to_pyarrow_batches

anjakefala:
## Description of changes

Added test to finish off https://github.com/ibis-project/ibis/pull/10445

### DIFF
--- a/ibis/backends/duckdb/tests/test_client.py
+++ b/ibis/backends/duckdb/tests/test_client.py
@@ -427,19 +427,19 @@ def test_pyarrow_batches_chunk_size(con):  # 10443
 
     t = ibis.memtable(
         {
-            "id": np.arange(500000),
-            "name": np.random.choice(["Alice", "Bob", "Carol", "Dave"], size=500000),
-            "age": np.random.randint(20, 70, size=500000),
+            "id": np.arange(10_000),
+            "name": np.random.choice(["Alice", "Bob", "Carol", "Dave"], size=10_000),
+            "age": np.random.randint(20, 70, size=10_000),
         }
     )
-    batches = t.to_pyarrow_batches(chunk_size=100_000)
-    assert len(next(batches)) == 100_000
-    assert len(next(batches)) == 100_000
+    batches = con.to_pyarrow_batches(t, chunk_size=4096)
+    assert len(next(batches)) == 4096
+    assert len(next(batches)) == 4096
 
-    batches = t.to_pyarrow_batches(chunk_size=800)
+    batches = con.to_pyarrow_batches(t, chunk_size=800)
     assert len(next(batches)) == 800
     assert len(next(batches)) == 800
 
-    batches = t.to_pyarrow_batches(chunk_size=-1)
+    batches = con.to_pyarrow_batches(t, chunk_size=-1)
     with pytest.raises(TypeError):
         next(batches)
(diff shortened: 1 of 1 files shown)


ibis/backends/duckdb/tests/test_client.py
```diff
@@ -420,3 +420,26 @@ def test_memtable_doesnt_leak(con, monkeypatch):
...
+    import numpy as np
+
+    t = ibis.memtable(
+        {
+            "id": np.arange(500000),
+            "name": np.random.choice(["Alice", "Bob", "Carol", "Dave"], size=500000),
+            "age": np.random.randint(20, 70, size=500000),
+        }
+    )
+    batches = t.to_pyarrow_batches(chunk_size=100_000)
```
cpcloud, line 435:
```suggestion
    batches = con.to_pyarrow_batches(t, chunk_size=100_000)
```

You need to use `con` here, otherwise you'll implicitly set the default backend, which is what's currently causing CI to fail.

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -420,3 +420,26 @@ def test_memtable_doesnt_leak(con, monkeypatch):
...
+    assert len(next(batches)) == 100_000
+    assert len(next(batches)) == 100_000
+
+    batches = t.to_pyarrow_batches(chunk_size=800)
```
cpcloud, line 439:
```suggestion
    batches = con.to_pyarrow_batches(t, chunk_size=800)
```

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -420,3 +420,26 @@ def test_memtable_doesnt_leak(con, monkeypatch):
...
+    assert len(next(batches)) == 800
+    assert len(next(batches)) == 800
+
+    batches = t.to_pyarrow_batches(chunk_size=-1)
```
cpcloud, line 443:
```suggestion
    batches = con.to_pyarrow_batches(t, chunk_size=-1)
```

ibis/backends/duckdb/tests/test_client.py
```diff
@@ -420,3 +420,26 @@ def test_memtable_doesnt_leak(con, monkeypatch):
...
     assert name not in con.list_tables()
     assert len(df) == 3
+
+
+def test_pyarrow_batches_chunk_size(con):  # 10443
+    import numpy as np
+
+    t = ibis.memtable(
+        {
+            "id": np.arange(500000),
```
cpcloud, line 430: Can we make these smaller? I think the smallest chunk size duckdb can produce is 2048 (by default, and configurable only at compile time IIRC), so any multiple of that greater than 1 should be fine to test chunk size.

cpcloud (changes requested):

ibis/backends/duckdb/tests/test_client.py
anjakefala: Interesting, a chunk size of 800 worked. But we can indeed make these smaller!

ibis/backends/duckdb/tests/test_client.py
anjakefala: I'll absolutely change it! When does using the table implicitly set the default backend?

cpcloud (approved):

ibis/backends/duckdb/tests/test_client.py
cpcloud: Whenever you don't explicitly use a backend instance like `con`. _Something_ has to do the compute, and if you haven't specified that thing explicitly by constructing a backend instance and then using it, or by calling `ibis.set_backend(<some backend instance>)` then Ibis is going to pick the default backend (DuckDB).

PR #10485, opened 2024-11-13
feat(geospatial): accept geopandas GDFs in `memtable`

### LINKED ISSUE #10453: bug: create_table() fails on geopandas
### What happened?

It would be great to be able to go from geopandas to ibis without serializing to disk first.  `create_table()` is  I think the usual way to do this, but it fails with this unexpected error here:  

```python
import ibis
from ibis import _
con = ibis.duckdb.connect(extensions=["spatial"])

gdf = (
    con
    .read_geo("/vsicurl/https://dsl.richmond.edu/panorama/redlining/static/mappinginequality.gpkg")
    .filter(_.city == "New Haven")
    .execute()
   )


con.create_table("redlines", gdf)
```

Throws:

```
ArrowTypeError: ('Input object was not a NumPy array', 'Conversion failed for column geom with type geometry')

```

9.5.0

duckdb

### Relevant log output

_No response_

### Code of Conduct

gforsyth:
## Description of changes

Passing a `geopandas.GeoDataFrame` directly to `memtable` will yield a bunch of
errors because our various `to_arrow` and other proxy methods don't know how to
handle the bundled geo data.

As an intermediate bit of functionality until things like `geoarrow` are more
widely supported, this PR adds an extra dispatch to the `memtable` constructor
to cast geometry columns to WKB, which allows most backends to ingest the
resulting data with the geo columns as binary, which can then be cast to
`geometry` columns in backends that support that.

## Issues closed

Resolves #10453

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -47,6 +47,7 @@ if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
     from pathlib import Path
 
+    import geopandas as gpd
     import pandas as pd
     import polars as pl
     import pyarrow as pa
@@ -550,6 +551,23 @@ def _memtable_from_polars_dataframe(
         schema=sch.infer(data) if schema is None else schema,
         data=PolarsDataFrameProxy(data),
     ).to_expr()
+
+
+@_memtable.register("geopandas.geodataframe.GeoDataFrame")
+def _memtable_from_geopandas_geodataframe(
+    data: gpd.GeoDataFrame,
+    *,
+    name: str | None = None,
+    schema: SchemaLike | None = None,
+    columns: Iterable[str] | None = None,
+):
+    # The Pandas data proxy and the `to_arrow` method on it can't handle
+    # geopandas geometry columns. But if we first make the geometry columns WKB,
+    # then the geo column gets treated (correctly) as just a binary blob, and
+    # DuckDB can cast it to a proper geometry column after import.
+    wkb_df = data.to_wkb()
+
+    return _memtable(wkb_df, name=name, schema=schema, columns=columns)
 
 
 def _deferred_method_call(expr, method_name, **kwargs):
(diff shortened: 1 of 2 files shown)


ibis/expr/api.py
```diff
@@ -552,6 +552,23 @@ def _memtable_from_polars_dataframe(
     ).to_expr()
 
 
+@_memtable.register("geopandas.geodataframe.GeoDataFrame")
+def _memtable_from_geopandas_geodataframe(
+    data: pl.DataFrame,
```
cpcloud, line 557: Should this `GeoDataFrame`?

gforsyth: oh, yup

ibis/expr/api.py
```diff
@@ -552,6 +552,23 @@ def _memtable_from_polars_dataframe(
...
+    *,
+    name: str | None = None,
+    schema: SchemaLike | None = None,
+    columns: Iterable[str] | None = None,
+):
+    # The Pandas data proxy and the `to_arrow` method on it can't handle
+    # geopandas geometry columns. But if we first make the geometry columns WKB,
+    # then the geo column gets treated (correctly) as just a binary blob, and
+    # DuckDB can cast it to a proper geometry column after import.
+    wkb_df = data.to_wkb()
```
cpcloud, line 567: Does DuckDB correctly infer the binary column as a geospatial type?

gforsyth: it does not, it infers it as `binary`, but you can then cast to `geometry` and that works

cpcloud: I see. Is it worth modifying the schema here to pass in columns that we know are geometry? Also fine to not do that and do it in a follow up, or never if it doesn't come up again.

gforsyth: I might try it in a follow-up, but this at least lets users create the table in the first place

gforsyth: Ahh, right, I did try this before.  I think if we say "this is a geometry column" when it's WKB, DuckDB errors (somewhat annoyingly).  You need the explicit type cast for it to work.

cpcloud (approved):

cpcloud:
Thanks @gforsyth!

PR #10490, opened 2024-11-14
fix(mssql): use nvarchar to avoid non ascii be question mark

grieve54706:
## Description of changes
In MSSQL, if the column name is non-ASCII, it will be a question mark when getting schema by `sys.dm_exec_describe_first_result_set`

```python
con = ibis.mssql.connect(...)
expr = con.sql("SELECT 1 AS [калона]")
schema = expr.schema()
names = schema.names
assert len(names) == 1
assert names[0] == "калона"
```
```
E    AssertionError: assert '??????' == 'калона'     
E    - калона
E    + ??????
```

We should make the sql string be a `nvarchar` for Unicode.

### DIFF
--- a/ibis/backends/mssql/__init__.py
+++ b/ibis/backends/mssql/__init__.py
@@ -328,7 +328,7 @@ class Backend(SQLBackend, CanCreateCatalog, CanCreateDatabase):
           scale,
           error_number,
           error_message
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
Excellent, thank you!

PR #10495, opened 2024-11-14
docs(examples): align Table.get_backend with ibis.get_backend examples

IndexSeek:
## Description of changes

This makes the example more concise and similar to the `ibis.get_backend()` example. That `# doctest: +ELLIPSIS` usage comes in handy. 

Adds the "See Also" to link to the other `ibis.get_backend` method.


cpcloud (approved):
Thanks!

PR #10496, opened 2024-11-15
fix(docs): add null examples header and expected failure

IndexSeek:
## Description of changes

Adds the section header for the examples for [null](https://ibis-project.org/reference/expression-generic#ibis.null). 

I added the expected failure annotation, but the `# quartodoc: +EXPECTED_FAILURE` comment still shows up in the render. 

image

I wasn't sure if there was something special I needed to do for it not to show up, like how this is here but the # doesn't show up in the [pivot_longer examples](https://ibis-project.org/reference/expression-tables.html#examples-28). 

https://github.com/ibis-project/ibis/blob/4d9901897fbcb4a480bafda3f98d85a9a32f9ed6/ibis/expr/types/relations.py#L3807-L3811

image

### DIFF
--- a/ibis/expr/types/generic.py
+++ b/ibis/expr/types/generic.py
@@ -2723,9 +2723,11 @@ def null(type: dt.DataType | str | None = None) -> Value:
     `NULL`s with an unspecified type are castable and comparable to values,
     but lack datatype-specific methods:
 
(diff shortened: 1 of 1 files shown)


cpcloud (approved):
Nice catch!

PR #10503, opened 2024-11-16
fix(deps): remove the upper bounds for `pyarrow`

### LINKED ISSUE #10492: feat: please consider not setting upper bounds on dependencies

Ibis currently sets upper bounds on dependencies. E.g. the latest release pins `pyarrow<18`

This causes issues in CI. For example [here](https://github.com/vega/altair/pull/3591) it was blocking upgrading to Python 3.13

I'd suggest not setting upper bounds on dependencies, as suggested here https://iscinumpy.dev/post/bound-version-constraints/

> Capping dependencies has long term negative effects, especially for libraries, and should never be taken lightly. A library is not installed in isolation; it has to live with other libraries in a shared environment. Only add a cap if a dependency is known to be incompatible or there is a high (>75%) chance of it being incompatible in its next release. Do not cap by default - capping dependencies makes your software incompatible with other libraries that also have strict lower limits on dependencies, and limits future fixes. 

9.5.0

Polars

### Code of Conduct

deepyaman:
## Issues closed

* Resolves #10492


gforsyth:
I'm in favor of this.  I think we should remove more than just the `pyarrow` upper pin, but I'll also let @cpcloud weigh in here.

cpcloud:
Seems like it's probably fine to remove upper bounds on some of these, pyarrow is probably the most important one, and we do the others in follow-ups as folks have time.

cpcloud:
@deepyaman Can you update the PR and resolve the conflicts?

cpcloud (approved):

cpcloud:
Made it a `fix` PR so that it'll show up in the release notes.

PR #10504, opened 2024-11-16
docs(contribute): include instructions to use Pixi

deepyaman:
## Description of changes

Share how I set up my development environment using Pixi.


docs/contribute/01_environment.qmd
```diff
@@ -58,7 +62,10 @@ for manager, params in managers.items():
     print(f"1. Install `gh`")
     print()
     print("    ```sh")
-    print(f"    {manager} install -c conda-forge gh")
+    if manager == "conda" or manager == "mamba":
+        print(f"    {manager} install -c conda-forge gh")
+    elif manager == "pixi":
+        print(f"    {manager} global install gh")
```
deepyaman, line 68: I didn't do this because I don't use `gh`, but wanted to keep things consistent...

docs/contribute/01_environment.qmd
```diff
@@ -67,37 +74,48 @@ for manager, params in managers.items():
...
+        "backends that are possible to install excluding Flink; use "
+        "`conda/environment-arm64-flink.yml` for setting up a dev environment for all "
+        "the backends that are possible to install including Flink. The reason to have "
+        "two separate environments is because `apache-flink` forces `pyarrow` to "
+        "downgrade to 11.0, which causes conflicts in other backends."
+    )
+    print()
+    if manager == "pixi":
+        print("    ::: {.callout-note}")
+        print("    This command will initialize a `pixi.toml` file and also modify ")
```
deepyaman, line 99: Alternatively, we could provide Pixi configuration in `pyproject.toml` (or `pixi.toml`), but this would be more involved to keep in sync; this approach worked well enough for me so far, with minimal effort.

docs/contribute/01_environment.qmd
```diff
@@ -67,37 +74,46 @@ for manager, params in managers.items():
...
+        print(f"    {manager} shell")
     print("    ```")
     print()
     print(
         f"1. Install your local copy of `ibis` into the {manager.capitalize()} environment"
     )
     print()
     print("    ```sh")
-    print("    pip install -e .")
+    print("    uv pip install -e .")
```
gforsyth, line 116: Where'd `uv` come from?  Should this just be `pip` still?

deepyaman: 1. `pip` isn't installed by default in the environment created this way (I could, but...)
2. Since #10424, `uv` is included in `conda/environment.yml` anyway, so why not use `uv` over `pip`? Seems there was a decision to generally move to `uv`, and it's what's being used in CI.

gforsyth: Ah, I didn't realize that `uv` had also been added to the yml files. All good, then!

gforsyth (approved):

gforsyth:
Dunno about that nix job CI is waiting on but I'm going to merge this

PR #10507, opened 2024-11-18
docs(examples): add `asof_join` weather example

IndexSeek:
## Description of changes

Sets up a scenario for an example of using [`asof_join`](https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.asof_join) using weather data, including sensor data collection and various weather events.

### DIFF
--- a/ibis/expr/types/relations.py
+++ b/ibis/expr/types/relations.py
@@ -3215,6 +3215,88 @@ class Table(Expr, _FixedTextJupyterMixin):
         -------
         Table
             Table expression
+
+        Examples
+        --------
+        >>> from datetime import datetime, timedelta
+        >>> import ibis
+        >>> ibis.options.interactive = True
+        >>> sensors = ibis.memtable(
+        ...     {
+        ...         "site": ["a", "b", "a", "b", "a"],
+        ...         "humidity": [0.3, 0.4, 0.5, 0.6, 0.7],
+        ...         "event_time": [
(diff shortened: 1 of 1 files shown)


ibis/expr/types/relations.py
```diff
@@ -3215,6 +3215,88 @@ def asof_join(
...
+        └────────┴────────────────┴─────────────────────────┘
+
+        We can find the closest event to each sensor reading with a 1 second tolerance.
+        Using the "site" column as a join predicate ensures we only match events that
+        occurred at or near the same site as the sensor reading.
+
+        >>> tolerance = timedelta(seconds=1)
+        >>> sensors.asof_join(events, on="event_time", predicates="site", tolerance=tolerance).drop(
+        ...     "event_time_right"
+        ... )
```
gforsyth, line 3288: Maybe we can order here by `event_time`?  Or by `humidity`, I guess -- but we'll want a consistent ordering so the doctests don't fail randomly due to row order

gforsyth (commented):
Thanks for putting this together @IndexSeek !  One small nit to avoid pain in the future.

ibis/expr/types/relations.py
IndexSeek: Thank you for the reminder, good catch! This ordering not only helps ensure the doctests pass but I think it makes the `asof_join` result a bit clearer to the user seeing the gaps where they are.

gforsyth (approved):
:shipit:

Issue #10231, opened 2024-09-26
feat: show `df.count(), len(df.schema())` when displaying `interactive = True` tables


kdheepak:
### Is your feature request related to a problem?

When interactively exploring data, it's always an additional step to get the "shape" of the data. This is what the display looks like right now:

<img width="975" alt="image" src="https://github.com/user-attachments/assets/a0407168-257e-4ae6-aa26-44ba4d3343c2">

### What is the motivation behind your request?

I find myself typing the following often:

```python
print(df.count(), len(df.schema()))
```

### Describe the solution you'd like

I'd like for the default display to print the shape of the dataframe (along with the name of the table too would be nice). Polars prints the shape of the data as the first line and it is very convenient:

<img width="521" alt="image" src="https://github.com/user-attachments/assets/b7d489e5-15bb-40c6-be8b-3e5261c6aec7">

### What version of ibis are you running?

version 9.5.0

<img width="390" alt="image" src="https://github.com/user-attachments/assets/c7457a71-d752-4b4a-9d0c-6ed46589b7c3">

### What backend(s) are you using, if any?

DuckDB

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kdheepak:
In interactive mode though I’m not really concerned about performance, right?

cpcloud:
Why wouldn't performance be relevant in interactive mode?

kdheepak:
I figure because in interactive mode I'm using exploring the data and I'm printing the dataframe often. After exploring and building the sequence of expressions, I move the code to a package into a function or method in a class and only then I really care about performance. 

In interactive mode, I'm usually cleaning data or validating data and seeing the number of rows can speed up the process for me. For example, when working with timeseries data, seeing that there are `8760` rows tells me I have hourly data for 365 days of the year. Seeing `8784` would tell me it is a leap year or `8736` would tell me a day is missing. Not having it printed out by default means I have to explicitly look for these issues.

It also gives me a sense for how long a computation might take or how long it might take to write the data to disk. I'm not sure if this directly applies here, but when I've used `pandas` in the past if I have a few thousand rows, I can use `df.apply` with a python lambda and get away a hacky data cleaning workflow. If I have a few million rows, then I know have to look for a more performant alternative. 

I also am assuming that people working with CSV files don't have too many rows that calculating rows becomes an issue. You definitely have a much better sense of how the performance varies across different backends of ibis, so I'd obviously defer to your judgement.

---

[...]

PR #1000, opened 2017-05-30
DOC/BUG: Replace cloudera/ibis with pandas-dev/ibis

cpcloud:

PR #10518, opened 2024-11-21
feat(repr): add `show_count` option to interactive repr

NickCrews:
Resolves [#1000](https://github.com/ibis-project/ibis/issues/10231)

Doesn't have tests yet. I want to get feedback on the semantics before I do that. Please let me know what you think of this. Thanks!

Currently looks like

```python
import ibis

t = ibis.memtable({"foo": range(1_000), "bar": range(1_000), "baz": range(1_000)})
t.preview(show_count=True)
#   3 cols by 1_000 rows   
# ┏━━━━━━━┳━━━━━━━┳━━━━━━━┓
# ┃ foo   ┃ bar   ┃ baz   ┃
# ┡━━━━━━━╇━━━━━━━╇━━━━━━━┩
# │ int64 │ int64 │ int64 │
# ├───────┼───────┼───────┤
# │     0 │     0 │     0 │
# │     1 │     1 │     1 │
# │     2 │     2 │     2 │
# │     3 │     3 │     3 │
# │     4 │     4 │     4 │
# │     5 │     5 │     5 │
# │     6 │     6 │     6 │
# │     7 │     7 │     7 │
# │     8 │     8 │     8 │
# │     9 │     9 │     9 │
# │     … │     … │     … │
# └───────┴───────┴───────┘
```

I made the config default be `show_count=True`. I can change this with pushback, but this is what I would start with. I would love to brainstorm a few benchmark test cases to run to see the perf difference. Some ideas:
  1. `ibis.duckdb.connect("mydb.db").table("my_table")` (I expect ~0 difference)
  2. `ibis.duckdb.connect().read_parquet("t.pq")` (I expect~0 difference)
  3. `ibis.duckdb.connect().read_csv("million_rows.csv")` (I expect small difference)
  4. `ibis.duckdb.connect().read_csv("billion_rows.csv")` (I expect large difference)
  5. `ibis.duckdb.connect().read_csv("million_rows.csv").some_expensive computation()` (I expect a difference, size depends on semantics of the function)

I considered adding a `.repr_options` attribute to expressions as described in https://github.com/ibis-project/ibis/issues/10231#issuecomment-2462961711, but I decided that was too complicated.

I considered showing the table name in the repr, eg with `table.get_name()`, but that is a separate question.

Currently, the option has the semantics of `show_count: bool`, and we ALWAYS show the column count. I considered other encodings such as `show_shape: Literal["rows", "cols", "both", None]`, but I thought that was overkill.


lboller-pwbm:
This looks fantastic!

I tried it out in a Jupyter kernel with a fairly wide table, which appears as a horizontally scrollable output. The dimensions appear but are center-justified, so the user has to scroll horizontally quite a bit to see them. I would suggest making it left-justified.

ibis/expr/types/pretty.py
```diff
@@ -433,7 +413,16 @@ def _to_rich_table(
...
+    if options.show_count:
+        # use underscore to be friendly to i18n and python REPL
+        nrows = f"{table.count().execute():_}"
+    else:
+        nrows = "…"
+    if isinstance(tablish, ir.Table):
+        dims = f"{orig_ncols:_} cols by {nrows} rows"
+    else:
+        dims = f"{nrows} rows"
+    rich_table = rich.table.Table(title=dims, padding=(0, 1, 0, 1))
```
kdheepak, line 425:
```suggestion
    rich_table = rich.table.Table(title=dims, padding=(0, 1, 0, 1), title_justify="left")
```

ibis/config.py
```diff
@@ -76,6 +79,7 @@ class Interactive(Config):
     max_string: int = 80
     max_depth: int = 1
     show_types: bool = True
+    show_count: bool = True
```
gforsyth, line 82: I think this is a nice feature to have.  I do not think it should default to true.

Interactive mode is supposed to be interactive, and something that can be computationally expensive isn't something we want to turn on by default.

The main failure case I can think of here is reading from CSVs in a blob store.  With `show_count=False`, this can be done with a range request and only download the first N rows (ish), except if there's an order-by.  `show_count=True` would mean any operation would trigger a download of the entire file.

And yes, that is one more reason not to use CSVs, but :shrug:

NickCrews: OK, I am OK with defaulting to False. Let's just wait for Philip to chime in to verify, but I assume he is going to have the same opinion as you.

I really don't think it is worth the complexity/cleverness, but throwing it out there that we could make each relation have its own config, and that could be auto-determined based on if it's a CSV/parquet/physical table, local/remote, etc.

gforsyth: > I assume he is going to have the same opinion as you.

We are slowly merging into the same person.  The transformation is nearly complete.

> I really don't think it is worth the complexity/cleverness, but throwing it out there that we could make each relation have its own config, and that could be auto-determined based on if it's a CSV/parquet/physical table, local/remote, etc.

I mean, it's definitely a reliably cheap(er) operation on anything that isn't a CSV, but I agree this seems like too much complexity.  I think a better move is your suggestion of persistent default configuration

ibis/expr/types/pretty.py
```diff
@@ -433,7 +413,18 @@ def _to_rich_table(
...
                 break
 
-    rich_table = rich.table.Table(padding=(0, 1, 0, 1))
+    if options.show_count:
+        # use underscore to be friendly to i18n and python REPL
+        nrows = f"{table.count().execute():_}"
+    else:
+        nrows = "…"
+    if isinstance(tablish, ir.Table):
+        dims = f"{orig_ncols:_} cols by {nrows} rows"
```
cpcloud, line 422: Can we make this follow the long-standing convention of "rows by columns"?

ibis/config.py
cpcloud: Yes, let's leave the default to `False`, otherwise it's a huge performance footgun/bug magnet.

ibis/expr/types/_rich.py
```diff
@@ -433,7 +413,18 @@ def _to_rich_table(
...
                 break
 
-    rich_table = rich.table.Table(padding=(0, 1, 0, 1))
+    if options.show_count:
+        # use underscore to be friendly to i18n and python REPL
+        nrows = f"{table.count().execute():_}"
+    else:
+        nrows = "…"
+    if isinstance(tablish, ir.Table):
+        dims = f"{orig_ncols:_} cols by {nrows} rows"
```
cpcloud: Actually, let's keep it the way you have it. If we get some people complaining about it we can always change it (or do nothing since AFAICT it's not a huge problem either way).

cpcloud:
@NickCrews I'll try and fix this up and get it into 10.6

cpcloud:
This didn't make it into 10.6, but it will be in the next release.

ibis/config.py
NickCrews: OK in the new version I changed to default of False

NickCrews:
In [`92922e3` (#10518)](https://github.com/ibis-project/ibis/pull/10518/commits/92922e3c37c8c4d23674c505d48f0510738bd6c9) I added an optional feature I can remove if you don't like it: the first time you call to_rich_table(), if the show_count is False, we append a help message of " (set \`ibis.options.repr.interactive.show_count=True\` to show)" to the end of the "6 cols by ... rows"

NickCrews:
@cpcloud if I get a +1 to this behavior and implementation, I can do the grunt work of transitioning all the docstrings.

If you know of a way to automate that migration for those few hundred failures, I'm all ears lol

PR #10520, opened 2024-11-21
build(ruff): remove unnecessary target-version use

deepyaman:
## Description of changes

`target-version` is duplicative/unnecessarily-overriding `requires-python`; see https://docs.astral.sh/ruff/settings/#target-version.


gforsyth (approved):
Nice catch!

PR #10522, opened 2024-11-22
feat(ux): include basename of path in generated table names in read_*()

NickCrews:
I often read in many files, and then when I do `con.tables`, it is not obvious which is which.
This should make it a little more user-friendly.

I don't think think we need to be perfect here, I was just aiming for slightly better. For example, if you pass a collection of paths, then we arbitrarily pick the first one to be representative. Same thing with globs, the generated name isn't always the prettiest. If the user wants more precise semantics, they can name the table themselves.

I also simplified the polars.read_parquet() logic while I was in there, it was needed in order to make this change.


ibis/util.py
```diff
@@ -522,6 +522,21 @@ def gen_name(namespace: str) -> str:
...
+
+    Examples
+    --------
+    >>> gen_name_from_path("path/to/myfile.csv", "csv")  # doctest: +ELLIPSIS
+    'ibis_read_csv_myfile_...'
+    """
+    root, _ext = os.path.splitext(path)
+    basename = os.path.basename(root)
+    basename = re.sub(r"[^a-zA-Z0-9_]", "_", basename).strip("_")
+    prefix = f"read_{file_type}_{basename}"
```
gforsyth, line 536: Since you already have the file extension, do you want to just use that for the `file_type` part of the generated name?

ibis/util.py
```diff
@@ -522,6 +522,21 @@ def gen_name(namespace: str) -> str:
...
+    """Create a user-friendly unique identifier from a file path.
+
+    Examples
+    --------
+    >>> gen_name_from_path("path/to/myfile.csv", "csv")  # doctest: +ELLIPSIS
+    'ibis_read_csv_myfile_...'
+    """
+    root, _ext = os.path.splitext(path)
+    basename = os.path.basename(root)
+    basename = re.sub(r"[^a-zA-Z0-9_]", "_", basename).strip("_")
```
gforsyth, line 535:
```suggestion
    basename = re.sub(r"\W", "_", basename).strip("_")
```

the compilation cost for your regex is pretty low anyway, but this one is prebuilt

NickCrews: I think I would slightly prefer to keep this as is, this code path is gonna get hit at most ~10k?? times in a session, so I think we don't need to worry about perf, and at least personally, I have to go look up what \W is, but the current spelling requires no lookup. If you feel strongly I will change it though <3

image

ibis/util.py
NickCrews: Yeah, I think you are right, let's just keep it simple. I actually have another idea on how to make this more user-friendly. Will push changes soon.

ibis/util.py
gforsyth: I dunno, those 4 nanoseconds look enticing!

Yeah, we can keep this as is 😅

ibis/util.py
NickCrews: well, my idea was to use the full path. But MySQL has a limit of 64 chars for table names, so pretty much any name of any reasonabkle length is too long. So I kept it to just use the basename as it currently does.

ibis/util.py
```diff
@@ -522,6 +522,28 @@ def gen_name(namespace: str) -> str:
...
+    >>> gen_name_from_path("s3://path/to/myfile.csv")  # doctest: +ELLIPSIS
+    'ibis_read_s3__path__to__myfile__csv...'
+    >>> gen_name_from_path("s3://long_long_long_path/to/myfile.csv")  # doctest: +ELLIPSIS
+    'ibis_read_s3__myfile__csv...'
+    """
+    basename = os.path.basename(path)
+    basename = re.sub(r"[^a-zA-Z0-9_]", "_", basename)
+    # MySQL has a limit of 64 characters for table names.
+    # Let's not give users runtime errors because of this.
+    basename = basename[-25:]
```
NickCrews, line 541: do we want to push this logic down into `gen_name()` so we never create a too-long name?

NickCrews:
Failing tests are revealing that our old tests were buggy: we currently verify that register() worked by doing a `assert "my_table" in con.list_tables()`. But, some test cases reuse the same name ("my_table"), and never clean up after themselves. So even if the register() doesn't work as intended, the test will still pass because that table is still left around by the last test case.

I think this means we should be more careful about cleaning up after ourselves. I started adding a helper class/context manager for this, but it was getting too complex, and wasn't related to this PR, so I think it should happen in a followup PR. So it's somewhere, here is an idea on what that helper class could look like:

```python
class BackendJanitor:
    """Cleans up tables created during a test.

    Examples
    --------
    >>> con = ibis.sqlite.connect(":memory:")
    >>> con.create_table("foo", schema=dict(x="int64"))
    >>> with BackendJanitor(con) as janitor:
    ...     con.create_table("bar", schema=dict(y="int64"))
    >>> con.list_tables()
    ['foo']
    >>> janitor.tables_final
    {'bar', 'foo'}
    """

    def __init__(self, con: ibis.BaseBackend):
        self.con = con
        self.tables_initial = set(con.list_tables())

    @property
    def tables_now(self) -> set[str]:
        return set(self.con.list_tables())

    @property
    def tables_final(self) -> set[str]:
        """The set of tables that existed at the point of the last call to `clean_up`."""
        if not hasattr(self, "_tables_final"):
            raise ValueError("need to call `clean_up` first")
        return self._tables_final

    def clean_up(self):
        """Remove all tables created since initialization."""
        self._tables_final = self.tables_now
        for table_name in self.tables_final - self.tables_initial:
            self.con.drop_table(table_name)

    def __enter__(self):
        return self

    def __exit__(self, *args):
        self.clean_up()
```

NickCrews:
ehh, man, I really don't love how I implemented the ensured-cleanup of the created tables. I will do that properly in another PR before/after this if you want.

ibis/backends/tests/test_register.py
```diff
@@ -39,6 +39,19 @@ def pushd(new_dir):
         os.chdir(previous_dir)
 
 
+def drop(table: ibis.Table):
+    backend = table._find_backend()
+    name = table.get_name()
+    try:
+        backend.drop_table(name)
```
NickCrews, line 46: This is one example use case that may inform the API we design in https://github.com/ibis-project/ibis/issues/8382

ibis/util.py
```diff
@@ -522,6 +522,26 @@ def gen_name(namespace: str) -> str:
     return f"ibis_{namespace}_{uid}"
 
 
+def gen_name_from_path(path: str | Path) -> str:
```
cpcloud, line 525: I think the idea of baking **any** part of the path into an identifier is a very bad idea.

Lots of software has weird limitations around paths, that we should not bring into Ibis, like the thing you've already discovered about MySQL.

In particular, paths can be arbitrary byte sequences, they're not required to be UTF-8, Latin-1, anything in between, or anything in particular except a sequence of bytes.

Is this implementation prepared to handle that? If not, then we shouldn't do it, because it's adding a new set of unnecessary assumptions. If you need to keep track of what tables map to what files, DuckDB has an argument for that, or if you're not using DuckDB then it's better done upstream in the system that does the reading.

cpcloud (changes requested):
I really do not like this idea.

ibis/util.py
NickCrews: I can write some code to handle the "random sequence of bytes". I think it would just replace those with the empty string, or an underscore, or something. I don't think the behavior here matters that much, as long as it doesn't error, because this should be so rare. I think this means that the implementation should be able to be fairly simple and maintainable.

If I come up with something not terrible would you consider it @cpcloud? I strongly still like the main idea of this PR.

cpcloud: I really don't like the idea of replace sequences of bytes with anything.

What are we gaining by doing any of this, that makes it a worthwhile maintenance tradeoff? In other words, why should we bake in new assumptions about paths?

Right now, I don't see an improvement from all the effort here that justifies the maintenance burden.

PR #10537, opened 2024-11-26
docs(examples): add hashbytes example

IndexSeek:
## Description of changes

Adds non-auto computed documentation example for [`hashbytes`](https://ibis-project.org/reference/expression-strings.html#ibis.expr.types.strings.StringValue.hashbytes).

Edit: removed previous addition of MD2 and MD4 hashing algorithms and reverted that commit. 

~~Also adds support for MD2 and MD4 hashing algorithms for Microsoft SQL Server's [HASHBYTES](https://learn.microsoft.com/en-us/sql/t-sql/functions/hashbytes-transact-sql) function.~~

~~Technically, quoting from the docs:~~

~~> Beginning with SQL Server 2016 (13.x), all algorithms other than SHA2_256, and SHA2_512 are deprecated.~~

~~Not sure if it's worth including this portion of the change or not.~~

### DIFF
--- a/ibis/backends/sql/compilers/mssql.py
+++ b/ibis/backends/sql/compilers/mssql.py
@@ -447,7 +447,7 @@ class MSSQLCompiler(SQLGlotCompiler):
         return self.if_(arg, 1, 0).eq(0)
 
     def visit_HashBytes(self, op, *, arg, how):
-        if how in ("md2", "md4", "md5", "sha1"):
+        if how in ("md5", "sha1"):
             return self.f.hashbytes(how, arg)
         elif how == "sha256":
             return self.f.hashbytes("sha2_256", arg)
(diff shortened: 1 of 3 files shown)


gforsyth (commented):
Thanks @IndexSeek !  

My inclination is to leave `md2` and `md4` without explicit support unless a user requests that functionality.  I don't want our inclusion of those options to read as tacit support for using them.

IndexSeek:
> Thanks @IndexSeek !
> 
> My inclination is to leave `md2` and `md4` without explicit support unless a user requests that functionality. I don't want our inclusion of those options to read as tacit support for using them.

You're welcome! That makes sense to me; I reverted that commit, so now it's just the hash bytes example.

gforsyth (approved):

PR #10544, opened 2024-11-29
feat(deps): install project and dev and test dependencies in devcontainer

### LINKED ISSUE #10532: Pre-install dev and core test dependencies in devcontainer container

I was hoping to have dev dependencies ready to use, and to run core tests without needing to install both sets of dependencies.

Having this context ready to use through VS Code, with any OS supporting devcontainers, seems like something that most  contributors using `devcontainers` would want, instead of each contributor needing to install these dependencies into the container.

I want to be able to run uv commands to interact with the environment, for example updating dependency groups as needed while supporting a given backend.

I would like to have a `uv .venv` that is activated so that when a terminal becomes available after container start-up, it is possible to run the commands

`uv pip list` and see all of the installed dependencies,

and

`pytest -m core` to run the core tests.

9.5.0

_No response_

### Code of Conduct

nickpowersys:
## Description of changes
* Install `uv` in development container
* Use `uv sync` to install project and its dependencies, dev and tests dependency groups, and selected optional dependencies/extras
* Switch the container user to the `vscode` user
* Give the `vscode` user ownership and permissions to the `uv` virtual environment and other directories needed for running `uv run pytest -m core`
* Enables use of Git within container, so commits can be pushed to/pulled from remote

## Issues closed
* Resolves #10532


cpcloud (approved):
Seems legit, thanks!

nickpowersys:
I am working on docs, and hope to finalize them this week.

cpcloud (approved):

PR #10545, opened 2024-11-29
refactor(register): remove deprecated register method

### LINKED ISSUE #10175: refactor(api): remove register API
In this PR https://github.com/ibis-project/ibis/pull/8863 we deprecated the register API, we should remove it.

gforsyth:
# refactor(register): remove deprecated register method

BREAKING CHANGE: The deprecated `register` method has been removed.
Please use the file-specific `read_*` methods instead. For in-memory
objects, pass them to `ibis.memtable` or `create_table`.

## Description of changes

The `register` method is removed everywhere. Test files have been renamed to
`test_io.py` instead of `test_register.py`, or they have been removed entirely
because they are duplicates of tests in `test_client.py` or elsewhere.

## Issues closed

#10175


gforsyth:
Hmm, so the `datafusion` backend has an option which is unique to that backend, where we allow a user to pass in a dictionary of `{table_name: file_to_load}` -- at connection creation, we hand each of these files off to `register` to load all of the files as tables.

I recognize the convenience here, but it's also against our general push towards requiring explicit table creation.  

Do we rip that out, too?  Or leave in `_register` as a private method in the datafusion backend to handle this?

IndexSeek:
Thank you for prepping this!

> Do we rip that out, too? Or leave in `_register` as a private method in the datafusion backend to handle this?

That's tricky. I can't imagine it's used too widely, but I vote to leave it there if anyone uses that functionality with the DataFusion backend today. 

Could we mark it deprecated and for the `instead` guide the user in creating these tables explicitly?

gforsyth:
> That's tricky. I can't imagine it's used too widely, but I vote to leave it there if anyone uses that functionality with the DataFusion backend today.
> 
> Could we mark it deprecated and for the `instead` guide the user in creating these tables explicitly?

Thanks, @IndexSeek !  This seemed like a good suggestion, so I've remove the public `register` method, but left the private one in along with deprecating its usage in loading files at connection time.

cpcloud (approved):
🚢 it

Issue #10535, opened 2024-11-26
docs: cases() documentation doesn't work for me at all


lexual:
### Please describe the issue

Can't get the conditional cases() examples to work (as opposed matching on a value), am I missing something?:

```
$ python -c 'import ibis; print(ibis.__version__); ibis.cases'
9.5.0
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/var/home/lexfed/code/lex_budget/_venv/lib64/python3.12/site-packages/ibis/__init__.py", line 142, in __getattr__
    return load_backend(name)
           ^^^^^^^^^^^^^^^^^^
  File "/var/home/lexfed/code/lex_budget/_venv/lib64/python3.12/site-packages/ibis/__init__.py", line 66, in load_backend
    raise AttributeError(msg)
AttributeError: module 'ibis' has no attribute 'cases'. . Did you mean: 'case'?
```

I tried other examples that specify I should be using `else_` kwarg, and it complains about it not existing, and it looks like it should be `default` ?

This is another verbatim from the docs:

```
$ python
Python 3.12.6 (main, Sep  9 2024, 00:00:00) [GCC 14.2.1 20240801 (Red Hat 14.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ibis
>>> ibis.options.interactive = True
>>> v = ibis.memtable({"values": [1, 2, 1, 2, 3, 2, 4]}).values
>>> v.to_polars()
shape: (7,)
Series: 'values' [i64]
[
        1
        2
        1
        2
        3
        2
        4
]
>>> ibis.cases((v == 1, "a"), (v > 2, "b"), else_="unk").name("cases")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/var/home/lexfed/code/lex_budget/_venv/lib64/python3.12/site-packages/ibis/__init__.py", line 142, in __getattr__
    return load_backend(name)
           ^^^^^^^^^^^^^^^^^^
  File "/var/home/lexfed/code/lex_budget/_venv/lib64/python3.12/site-packages/ibis/__init__.py", line 66, in load_backend
    raise AttributeError(msg)
AttributeError: module 'ibis' has no attribute 'cases'. . Did you mean: 'case'?
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

lexual:
Thanks @IndexSeek 

Appreciate the apology, but not a big deal for me.
[...]

Issue #5347, opened 2023-01-27
docs: remove docs versioning


cpcloud:
It's not clear if versioning is adding a ton of value right now and does create some annoying problems. Let's remove the versioning!

PR #10551, opened 2024-12-03
docs: add callout notes for some APIs changes and additions

NickCrews:
fixes https://github.com/ibis-project/ibis/issues/10535

Other things I considered:

## using some other mechanism, like sphinx versionadded directives

like https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded

but I don't think our doc system uses sphinx. So I think we are limited to using quarto's constructs?

## using a different callout type

[There is warning, tip, etc](https://quarto.org/docs/authoring/callouts.html#callout-types). I thought note made the most sense.

## Adding a `@versionadded` decorator

Similar to the `@experimental` and `@deprecated` decorators in util.py. But I thought adding them inline like this wasn't that hard.

## Placing the callout in a different place in the docstring

eg further up, where it might be more obvious. [But the official python docs put them down at the bottom](https://docs.python.org/3/library/enum.html#enum.Enum.__format__), so I chose to follow that convention. I think that future-proofs us if there are multiple callouts in a row.

## Some sort of autocheck

In future PRs, how can we be sure that docstrings are updated like this? I couldn't think of a good way to do it. Maybe if there is a `breaking-change` tag on the issue? but that doesn't work for `Added in version` changes. I think the best we can do is to just have all maintainers be aware of this. If we like this approach then we can tag in all the maintainers into this issue?

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -392,7 +392,6 @@ class Backend(SQLBackend, CanCreateDatabase, UrlFromPath):
         >>> import ibis
         >>> ibis.duckdb.connect(threads=4, memory_limit="1GB")  # doctest: +ELLIPSIS
         <ibis.backends.duckdb.Backend object at 0x...>
-
         """
         if not isinstance(database, Path) and not database.startswith(
             ("md:", "motherduck:", ":memory:")
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -365,6 +365,15 @@ class Backend(SQLBackend, CanCreateDatabase, UrlFromPath):
     ) -> None:
         """Create an Ibis client connected to a DuckDB database.
 
+        ::: {.callout-note title="Changed in version 10.0.0"}
+        Before, we had special handling if the user passed the `temp_directory`
+        parameter, setting a custom default, and creating intermediate
+        directories if necessary. Now, we do nothing, and just pass the value
+        directly to DuckDB. You may need to add
+        `Path(your_temp_dir).mkdir(exists_ok=True, parents=True)`
+        to your code to maintain the old behavior.
+        :::
+
         Parameters
         ----------
         database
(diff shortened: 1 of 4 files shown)


NickCrews:
Hmm, I think actually it's a bit hard to see down at the bottom. Maybe it would be better if the notes were right before the Parameters section so it's easier to see, it doesn't take up much room. If there were every any detailed notes, the call-out would start collapsed?

NickCrews:
@cpcloud @gforsyth this is ready for review whenever you have a chance! I think I am in favor of moving the call-out to just before the Parameters section, but wanted your confirmation before I made that effort.

NickCrews:
@IndexSeek you had chimed in on the related issue, do you have any thoughts here?

IndexSeek:
I agree. It might make more sense to move it above the parameters section. That's a good place for it.

We don't have any Quarto callout blocks in the other docstrings. I have seen the Sphinx directives used in docstrings before; I was curious to know if this was common. 

I think this is a valuable change, but I want to ensure we're not undoing https://github.com/ibis-project/ibis/issues/5347 as you commented on the other issue.

gforsyth:
I'll try to take a closer look this week. Broadly on board with adding this information and I think above parameters is a fine place for it.

I don't think it's counter to #5347 -- having version related information is fine, we just didn't want the headache of publishing versioned docs.

ibis/backends/duckdb/__init__.py
```diff
@@ -384,6 +393,7 @@ def do_connect(
         >>> import ibis
         >>> ibis.duckdb.connect(threads=4, memory_limit="1GB")  # doctest: +ELLIPSIS
         <ibis.backends.duckdb.Backend object at 0x...>
+
```
cpcloud, line 396: This seems unnecessary.

```suggestion
```

cpcloud (changes requested):
Other than the spurious additional newline, LGTM!

cpcloud (approved):

PR #10559, opened 2024-12-06
fix(docs): render func args as args, not list of args

### LINKED ISSUE #10558: docs(quartodoc): `render_methods` is showing arguments as a list
I noticed this while working on #10557 but I think it's been reported on Zulip, as well.

Something in our `render_methods` (or `render_method`) helpers is rendering function signatures with all of the args as list elements, so instead of

```python
# desired output
connect(self, *args, **kwargs)
```

we're instead getting

```python
connect(['self', '*args', '**kwargs'])
```

gforsyth:
## Description of changes

Fixes the way we render docstrings in the docs so that we show the argument
names, and keyword defaults, as they should be entered, and not as a list of
strings.

## Issues closed

Fixes #10558

### DIFF
--- a/docs/backends/_utils.py
+++ b/docs/backends/_utils.py
@@ -63,7 +63,7 @@ def render_method(*, member, renderer: MdRenderer) -> Iterator[str]:
     yield f"{header} {name} {{ #{member.path} }}"
     yield "\n"
     if params is not None:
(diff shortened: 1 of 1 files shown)


IndexSeek (approved):
Nice! Thank you for tackling this one.

I also remember seeing the report on Zulip.

PR #10582, opened 2024-12-16
feat(python): support python 3.13

cpcloud:
Add support Python 3.13


.github/workflows/nix.yml
```diff
@@ -38,11 +38,14 @@ jobs:
...
           - "3.11"
           - "3.12"
+          - "3.13"
         include:
           - os: ubuntu-arm64-24.04 # aarch64-linux
             python-version: "3.12"
           - os: macos-14 # aarch64-darwin
             python-version: "3.12"
+          - os: macos-14 # aarch64-darwin
+            python-version: "3.13"
```
gforsyth, line 48: Does this need a corresponding entry in the skip helper?

gforsyth (approved):
nice!

.github/workflows/nix.yml
cpcloud: Done.

nix/overlay.nix
```diff
@@ -53,6 +53,11 @@ let
         # Use base package set from pyproject.nix builders
         (pkgs.callPackage pyproject-nix.build.packages {
           inherit python;
+          stdenv = pkgs.stdenv.override {
+            targetPlatform = pkgs.stdenv.targetPlatform // {
+              darwinSdkVersion = "12.0";
```
cpcloud, line 58: This isn't strictly required for python 3.13 support, but it sure does make life easier by allowing more wheels to be used on macOS. A bunch of the more-complex-to-build wheels like scipy and scikit-learn are only available using a 12.0 darwin SDK build.

cpcloud:
The remaining single failure was fixed in https://github.com/kylebarron/arro3/issues/276 and once the 0.4.5 wheels are uploaded (after [this CI run](https://github.com/kylebarron/arro3/actions/runs/12356587820) succeeds), then I'll update `uv.lock` using `just update` and we should be good to merge this PR.

cpcloud:
Ok, bumped those dumps by way of `just update`.

cpcloud:
I'll do some follow up to see whether we can run any of the clouds with 3.13.

PR #10584, opened 2024-12-16
test(bigquery): fix array concat test

cpcloud:
Fixes the upstream failing test cases for bigquery.

### DIFF
--- a/ibis/backends/tests/test_array.py
+++ b/ibis/backends/tests/test_array.py
@@ -192,7 +192,13 @@ def test_array_concat_scalar(con, op):
     reason="doesn't support nullable arrays",
 )
 @pytest.mark.notyet(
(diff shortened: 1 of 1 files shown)


cpcloud:
Tests are now correctly xfailing:

```
ibis/backends/tests/test_array.py::test_array_concat_with_null[bigquery-right] XFAIL (bigquery treats null arrays as empty and a ValueError is raised from `pd.isna` when the value is not a
truthy value)                                                            [ 50%]
ibis/backends/tests/test_array.py::test_array_concat_with_null[bigquery-left] XFAIL (bigquery treats null arrays as empty and a ValueError is raised from `pd.isna` when the value is not a
truthy value)                                                            [100%]

================================================================================= 36 deselected, 2 xfailed in 50.23s =================================================================================
```

PR #10595, opened 2024-12-18
refactor(api): remove `IntegerValue.label`

### LINKED ISSUE #9612: feat: remove `IntegerValue.label()`
Putting this up for tracking it, not immediately actionable.

Once #9096 lands then I think we should remove the `IntegerValue.label(labels,  nulls)` method. Users can easily switch to `val.cases(*enumerate(labels), else_=nulls)`, and we can have this pointer in the deprecation notice.

_Originally posted by @NickCrews in https://github.com/ibis-project/ibis/pull/9096#discussion_r1586541965_

cpcloud:
Closes #9612.

BREAKING CHANGE: `IntegerValue.label` is redundant with the `IntegerValue.cases` method, use that instead. Replace `expr.label(labels)` with `expr.cases(*enumerate(labels))`

### DIFF
--- a/ibis/backends/impala/tests/test_bucket_histogram.py
+++ b/ibis/backends/impala/tests/test_bucket_histogram.py
@@ -81,8 +81,8 @@ def test_bucket_assign_labels(table, snapshot):
     bucket = table.f.bucket(buckets, include_under=True)
 
     size = table.group_by(bucket.name("tier")).size()
-    labelled = size.tier.label(
-        ["Under 0", "0 to 10", "10 to 25", "25 to 50"], nulls="error"
+    labelled = size.tier.cases(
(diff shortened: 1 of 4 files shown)


gforsyth (approved):
Looks good to me -- maybe we can add an example of the new usage to the breaking message?

NickCrews:
+1 to including a guide/example on how to migrate in the commit message, which can then get included in the changelog.

I would also not categorize this as a refactor(API) commit if there is a tag for breaking(API) or something. Or we should add one. refactor as a tag is definitely misleading.

Besides those two things looks great! Thanks.

NickCrews (approved):

cpcloud:
The category of the change and whether it's breaking are two mostly independent properties.

PR #10617, opened 2024-12-26
fix(api): ensure memtable schema and columns match

deepyaman:
## Description of changes

Before this change, if you pass a `SchemaLike` that isn't actually a `Schema`, it won't automatically rename the auto-generated pandas column names. For example:

```python
import pandas as pd

df = pd.DataFrame([[1, 2], [3, 4]])
schema = {"a": "int64", "b": "int64"}
t = ibis.memtable(df, schema=schema)
t.a.execute()  # Raises a KeyError
```

This PR goes ahead and constructs the `Schema`, if passed—work that was going to be done anyway. That way, `getattr(schema, "names")` finds the passed names, and the rename is performed.

### DIFF
--- a/ibis/expr/tests/test_api.py
+++ b/ibis/expr/tests/test_api.py
@@ -143,6 +143,15 @@ def test_duplicate_columns_in_memtable_not_allowed():
         ibis.memtable(df)
 
 
+def test_memtable_column_names_match_schema():
+    pd = pytest.importorskip("pandas")
+
+    df = pd.DataFrame([[1, 2], [3, 4]])
+    schema = {"a": "int64", "b": "int64"}
+    t = ibis.memtable(df, schema=schema)
+    t.a.execute()  # Raises a KeyError if the column name does not match
+
+
 @pytest.mark.parametrize(
     "op",
     [
--- a/ibis/expr/tests/test_api.py
+++ b/ibis/expr/tests/test_api.py
@@ -149,7 +149,7 @@ def test_memtable_column_names_match_schema():
     df = pd.DataFrame([[1, 2], [3, 4]])
     schema = {"a": "int64", "b": "int64"}
     t = ibis.memtable(df, schema=schema)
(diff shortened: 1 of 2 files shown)


cpcloud (changes requested):
You need to monkey patch the default backend setting

ibis/expr/tests/test_api.py
```diff
@@ -143,6 +143,15 @@ def test_duplicate_columns_in_memtable_not_allowed():
         ibis.memtable(df)
 
 
+def test_memtable_column_names_match_schema():
+    pd = pytest.importorskip("pandas")
+
+    df = pd.DataFrame([[1, 2], [3, 4]])
+    schema = {"a": "int64", "b": "int64"}
+    t = ibis.memtable(df, schema=schema)
+    t.a.execute()  # Raises a KeyError if the column name does not match
```
cpcloud, line 152: Execution tests should not be under this directory

cpcloud (changes requested):

ibis/expr/tests/test_api.py
deepyaman: How about I don't execute, and instead I check that the columns of the dataframe underlying the memtable have been renamed/match--would that work?

deepyaman:
> You need to monkey patch the default backend setting

Where?

cpcloud:
> > You need to monkey patch the default backend setting
> 
> Where?
> 
> (Also, why? Did I cause the nix failure; doesnt seem related to my change, even though its related to the default backend?)

The default backend attribute is global state, and tests run in a random order. This means that **any** of our suites could have failed in the way, you just got lucky 😄

ibis/expr/tests/test_api.py
cpcloud: That seems better, and then you don't need to muck around with using `monkeypatch` to set the default backend.

deepyaman: Done!

cpcloud (approved):

PR #10620, opened 2024-12-27
chore: relax upper bounds on required dependencies (except sqlglot)

### LINKED ISSUE #10492: feat: please consider not setting upper bounds on dependencies

MarcoGorelli:
## Description of changes

similar to #10503

closes #10492 even more

I've kept the upper bound on sqlglot, but i'd appreciate it if the other pins could be relaxed please

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,13 +26,13 @@ classifiers = [
   "Topic :: Software Development :: User Interfaces",
 ]
 dependencies = [
-  "atpublic>=2.3,<6",
-  "parsy>=2,<3",
-  "python-dateutil>=2.8.2,<3",
+  "atpublic>=2.3",
+  "parsy>=2",
+  "python-dateutil>=2.8.2",
   "pytz>=2022.7",
   "sqlglot>=23.4,<26.1",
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
Thanks!

We should keep the pandas and numpy bounds since they both break quite hard on their major releases.

For others, please follow up with additional PRs if you're up for it!

PR #10623, opened 2024-12-27
feat(polars): add `Intersection` and `Difference` ops

IndexSeek:
## Description of changes

Adds support for [`Intersection`](https://ibis-project.org/reference/operations.html#ibis.expr.operations.relations.Intersection) and [`Difference`](https://ibis-project.org/reference/operations.html#ibis.expr.operations.relations.Difference) operations for the Polars backend. The approach uses pl.SQLContext to register the dataframes and query using [INTERSECT](https://docs.pola.rs/api/python/stable/reference/sql/set_operations.html#intersect) and [EXCEPT](https://docs.pola.rs/api/python/stable/reference/sql/set_operations.html#except), respectively. 

I saw that a few other functions used a `ctx` argument which might be using `pl.SQLContext` in a similar manner, I wasn't sure if that argument could be used here to avoid some of the boilerplate. Here are some of those functions I am referring to:

https://github.com/ibis-project/ibis/blob/28bafd14153cb7bb5bcc44063975f9346d55b76a/ibis/backends/polars/compiler.py#L69-L72

https://github.com/ibis-project/ibis/blob/28bafd14153cb7bb5bcc44063975f9346d55b76a/ibis/backends/polars/compiler.py#L1287-L1297

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -1256,6 +1256,56 @@ def execute_union(op, **kw):
     return result
 
 
+@translate.register(ops.Intersection)
+def execute_intersection(op, *, ctx, **kw):
+    left = gen_name("polars_intersect_left")
+    right = gen_name("polars_intersect_right")
+
+    ctx.register_many(
+        frames={
+            left: translate(op.left, ctx=ctx, **kw),
+            right: translate(op.right, ctx=ctx, **kw),
+        }
+    )
+
+    sql = (
+        sg.select(STAR)
+        .from_(sg.to_identifier(left, quoted=True))
+        .intersect(sg.select(STAR).from_(sg.to_identifier(right, quoted=True)))
+    )
+
+    result = ctx.execute(sql.sql(Polars), eager=False)
+
+    if op.distinct is True:
+        return result.unique()
+    return result
+
+
+@translate.register(ops.Difference)
+def execute_difference(op, *, ctx, **kw):
+    left = gen_name("polars_diff_left")
+    right = gen_name("polars_diff_right")
+
+    ctx.register_many(
+        frames={
+            left: translate(op.left, ctx=ctx, **kw),
+            right: translate(op.right, ctx=ctx, **kw),
+        }
+    )
+
+    sql = (
+        sg.select(STAR)
+        .from_(sg.to_identifier(left, quoted=True))
+        .except_(sg.select(STAR).from_(sg.to_identifier(right, quoted=True)))
+    )
+
+    result = ctx.execute(sql.sql(Polars), eager=False)
+
+    if op.distinct is True:
+        return result.unique()
+    return result
+
+
 @translate.register(ops.Hash)
 def execute_hash(op, **kw):
     # polars' hash() returns a uint64, but we want to return an int64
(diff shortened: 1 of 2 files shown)


cpcloud:
@IndexSeek Thanks for the PR as always 😄 

I think it's reasonable to reuse the `ctx` argument in your implementations here rather than constructing a new `SQLContext`.

IndexSeek:
> @IndexSeek Thanks for the PR as always 😄
> 
> I think it's reasonable to reuse the `ctx` argument in your implementations here rather than constructing a new `SQLContext`.

You're welcome! That seems like a good plan. I gave this a try using the `ctx` from kwargs. I was experimenting with a way to reuse the existing tables as part of the context, but for some reason when I was doing `op.left.name` it would work standalone but when running tests I would get things like:

AttributeError: 'Sort' object has no attribute 'name'

This was working for me when using a memtable in an IPython shell, but not on the tests. 

Do you know if there is a more efficient way than needing to manually register the tables as "left" and "right"? 

Here's what I was trying where it seems to work using `op.left.name`, etc.:

```py
@translate.register(ops.Difference)
def execute_difference(op, **kw):
    ctx = kw.get("ctx")
    sql = (
        sg.select(STAR)
        .from_(sg.to_identifier(op.left.name, quoted=True))
        .except_(sg.select(STAR).from_(sg.to_identifier(op.right.name, quoted=True)))
    )
    result = ctx.execute(sql.sql(Polars), eager=False)
    if op.distinct is True:
        return result.unique()
    return result
```

```pycon
   ...: 
   ...: ibis.set_backend("polars")
   ...: 
   ...: t1 = ibis.memtable({"number_col": [1, 2, 3]})
   ...: t2 = ibis.memtable({"number_col": [2, 3, 4]})
   ...: t1.difference(t2)
Out[1]: 
┏━━━━━━━━━━━━┓
┃ number_col ┃
┡━━━━━━━━━━━━┩
│ int64      │
├────────────┤
│          1 │
└────────────┘
```

ibis/backends/polars/compiler.py
```diff
@@ -1256,6 +1256,46 @@ def execute_union(op, **kw):
     return result
 
 
+@translate.register(ops.Intersection)
+def execute_intersection(op, **kw):
+    ctx = kw.get("ctx")
+    ctx.register_many(
+        frames={
+            "left": translate(op.left, **kw),
+            "right": translate(op.right, **kw),
```
cpcloud, line 1265:
```suggestion
def execute_intersection(op, *, ctx, **kw):
    ctx.register_many(
        frames={
            "left": translate(op.left, ctx=ctx **kw),
            "right": translate(op.right, ctx=ctx, **kw),
```

ibis/backends/polars/compiler.py
```diff
@@ -1256,6 +1256,46 @@ def execute_union(op, **kw):
...
+    return result
+
+
+@translate.register(ops.Difference)
+def execute_difference(op, **kw):
+    ctx = kw.get("ctx")
+    ctx.register_many(
+        frames={
+            "left": translate(op.left, **kw),
+            "right": translate(op.right, **kw),
```
cpcloud, line 1285:
```suggestion
def execute_difference(op, *, ctx, **kw):
    ctx.register_many(
        frames={
            "left": translate(op.left, ctx=ctx, **kw),
            "right": translate(op.right, ctx=ctx, **kw),
```

cpcloud (changes requested):

cpcloud (approved):

cpcloud:
@IndexSeek Thank you!

Issue #10243, opened 2024-09-27
docs: mention and clarify meaning of ! in schemas


hottwaj:
### Please describe the issue

Current docs online about tables, schemas and data types do not mention that schema types may include an "!" prefix, e.g. a schema might look like this

```
{'not null float': '!float64',
 'nullable float': 'float64',}
```

see e.g. https://ibis-project.org/reference/schemas

Probably should be a separate ticket, but I'd also suggest that the use of ! to indicate "not null" could do with a re-think because I think its ambiguous depending on your background e.g. my expectation was that this was the correct interpretation

```
{'required float': 'float64',  # but ibis actually uses !float64 for this
 'not required float': '!float64',}  # ibis uses float64 for this
```

Thanks!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
Thanks, yep, that should be documented. Any interesting in making a PR to add some docs?

ahaan1984:
Hi! Can I try working on this issue?

cpcloud:
> Hi! Can I try working on this issue?

Yes, of course!

ahaan1984:
Hi, I'm sorry for lack of communication from my end, I couldn't work on the issue as I was sick, I'll get a PR created in a few days.

daniel-bale:
not sure should I open a new issue

ibis.dtype does not accept string value with nullable=False. It is confusing if one not knowing "!" prefix exists.
Doc should mention this or output more helpful error message.

Also, I cannot find string name of data types in docs. It would be nice to have a table to list all and also mention the "!" prefix.

gforsyth:
Hi @daniel-bale -- all the datatype names are listed on this page: https://ibis-project.org/reference/datatypes

And we should support passing `nullable` as a kwarg when using string-parsing for dtypes.  I can put up a PR for that.

PR #10632, opened 2024-12-30
feat(dtypes): allow passing `nullable` kwarg to string parsed dtypes

gforsyth:
## Description of changes

Noted in #10243, passing `nullable` to `ibis.dtype` when using a string name for
the dtype leads to an unhelpful error.  Instead, we should just support this.

This adds support for 'primitive' dtype specifications -- there's no defined
behavior for how to specify nested nullable types with a single keyword
argument, but this should cover most cases.

## Issues closed

xref #10243 but doesn't close it

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -73,8 +73,8 @@ def dtype(value: Any, nullable: bool = True) -> DataType:
 
 
 @dtype.register(str)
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
🚢

cpcloud:
Seems more like a new feature, but no big deal if you want to keep it as a bugfix.

PR #10634, opened 2024-12-30
feat(polars): add bitwise operations (`bit_and`, `bit_or`, `bit_xor`)

IndexSeek:
## Description of changes

Add support for [`bit_and`], [`bit_or`], and [`bit_xor`] for the Polars backend. This uses [`bitwise_and`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.bitwise_and.html#), [`bitwise_or`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.bitwise_or.html#), and [`bitwise_xor`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.bitwise_xor.html#) Polars expression methods. 

The `is_in` tests are failing still:
FAILED ibis/backends/tests/test_aggregation.py::test_reduction_ops[polars-is_in-bit_xor] - AssertionError: 
FAILED ibis/backends/tests/test_aggregation.py::test_reduction_ops[polars-is_in-bit_and] - AssertionError: 
FAILED ibis/backends/tests/test_aggregation.py::test_reduction_ops[polars-is_in-bit_or] - AssertionError: 

I've tried to repro to see why they are failing, but I may be overlooking something. Here is how I repro'd:

```pycon
In [2]: import numpy as np
In [3]: import pandas as pd
In [4]: ibis.set_backend("polars")
In [5]: alltypes = ibis.read_parquet("ci/ibis-testing-data/parquet/functional_alltypes.parquet")
In [6]: alltypes = alltypes.filter(_.id < 1550, _.string_col.isin(["1", "7"]))
In [7]: alltypes.agg(tmp=_.bigint_col.bit_or())
Out[7]: 
┏━━━━━━━┓
┃ tmp   ┃
┡━━━━━━━┩
│ int64 │
├───────┤
│    78 │
└───────┘
In [8]: np.bitwise_or.reduce(alltypes.execute().bigint_col).squeeze()
Out[8]: np.int64(78)
```

Here is what the test is reporting:
```
E           AssertionError: 
E           Not equal to tolerance rtol=0.001, atol=0
E           
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference among violations: 48
E           Max relative difference among violations: 0.61538462
E            ACTUAL: array(126)
E            DESIRED: array(78)

ibis/backends/tests/test_aggregation.py:568: AssertionError
============================================== short test summary info ==============================================
FAILED ibis/backends/tests/test_aggregation.py::test_reduction_ops[polars-is_in-bit_or] - AssertionError: 
```

I'm wondering maybe if rows aren't being properly filtered or something.

### DIFF
--- a/ibis/backends/polars/compiler.py
+++ b/ibis/backends/polars/compiler.py
@@ -717,6 +717,9 @@ _reductions = {
     ops.Median: "median",
     ops.Min: "min",
     ops.Sum: "sum",
+    ops.BitOr: "bitwise_or",
+    ops.BitAnd: "bitwise_and",
+    ops.BitXor: "bitwise_xor",
 }
 
 
@@ -1614,21 +1617,3 @@ def visit_StringFind(op, **kw):
     end = translate(op.end, **kw) if op.end is not None else None
     expr = arg.str.slice(start, end).str.find(_literal_value(op.substr), literal=True)
     return pl.when(expr.is_null()).then(-1).otherwise(expr + start)
-
-
-@translate.register(ops.BitAnd)
-def visit_BitAnd(op, **kw):
-    arg = translate(op.arg, **kw)
-    return arg.bitwise_and()
-
-
-@translate.register(ops.BitOr)
-def visit_BitOr(op, **kw):
-    arg = translate(op.arg, **kw)
-    return arg.bitwise_or()
-
-
-@translate.register(ops.BitXor)
-def visit_BitXor(op, **kw):
-    arg = translate(op.arg, **kw)
-    return arg.bitwise_xor()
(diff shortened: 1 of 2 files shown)


cpcloud:
@IndexSeek Thanks!

The issue was that you didn't thread the `where` argument through the rule.

I was able to solve the problem by reusing the `_reductions` mapping, which we use to construct reduction dispatch rules that have this bit of code which handles the filtering:

```python
       if op.where is not None:
           arg = arg.filter(translate(op.where, **kw))
```

cpcloud (approved):

IndexSeek:
> @IndexSeek Thanks!
> 
> The issue was that you didn't thread the `where` argument through the rule.
> 
> I was able to solve the problem by reusing the `_reductions` mapping, which we use to construct reduction dispatch rules that have this bit of code which handles the filtering:
> 
> ```python
>        if op.where is not None:
>            arg = arg.filter(translate(op.where, **kw))
> ```

You're welcome! I suspected it had something to do with the filter, thanks for sharing that snippet. That will probably be useful in future ops. 

Adding it to the `_reductions` dictionary is a much cleaner approach. ✨

PR #10635, opened 2024-12-31
feat(polars): add `TimeFromHMS` op

IndexSeek:
## Description of changes

Adds [`ops.TimeFromHMS`](https://ibis-project.org/reference/operations#ibis.expr.operations.temporal.TimeFromHMS) support for the Polars backend using [`pl.time`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.time.html).

I tried adding this to the `_date_methods` and even `_binops` dictionaries, but it
wasn't taking, so I split out a new function.


cpcloud (approved):
Nice!

cpcloud:
The MS SQL failure looks transient. I'll kick it off again once the other jobs finish.

PR #10659, opened 2025-01-08
feat(postgres): use `psycopg` rather than `psycopg2`

judahrand:
## Description of changes

The newer `psycopg` package generally has better performance and is more maintained than the older `psycopg2` package.

### DIFF
--- a/ibis/backends/risingwave/__init__.py
+++ b/ibis/backends/risingwave/__init__.py
@@ -7,11 +7,9 @@ from operator import itemgetter
 from typing import TYPE_CHECKING, Any
 from urllib.parse import unquote_plus
 
-import psycopg2
 import sqlglot as sg
 import sqlglot.expressions as sge
 from pandas.api.types import is_float_dtype
-from psycopg2 import extras
 
 import ibis
 import ibis.backends.sql.compilers as sc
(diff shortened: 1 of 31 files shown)


judahrand:
I was planning to continue this PR when I got the chance. I realized after starting that Rising Wave needed to move too because of the inheritance structure but that one is more difficult to iterate on in local development.

gforsyth:
> I was planning to continue this PR when I got the chance. I realized after starting that Rising Wave needed to move too because of the inheritance structure but that one is more difficult to iterate on in local development.

@judahrand -- what kind of machine are you on?  There's a `risingwave` container in our testing setup, so you should be able to `just up risingwave` (but this may be trickier if you're on an M1 or M2 mac)

cpcloud:
If the main use case here is for the postgres backend, then I would suggest we decouple the backends first and then do the upgrade. I can put up a PR for that.

As is, even after fixing the top level risingwave integration there are some new, reproducible segfaults.

ibis/backends/risingwave/__init__.py
```diff
@@ -106,48 +105,6 @@ def _from_url(self, url: ParseResult, **kwargs):
 
         return self.connect(**kwargs)
 
-    def _register_in_memory_table(self, op: ops.InMemoryTable) -> None:
```
cpcloud, line 109: This was a duplicated definition not caught by ruff, and that I introduced in #10669.

cpcloud (approved):

PR #10666, opened 2025-01-10
fix(typing): update common typing to skip non-existent attributes

### LINKED ISSUE #10665: bug: Hamilton pipeline loading ibis extension produces AttributeError: __provide__ when running on Windows
### What happened?

We have a Hamilton pipeline project that makes use of the ibis-framework extension for data handling.  We noticed on our local windows development sandbox that it produces an error when trying to get attributes for a specific object in the common typing module. The attribute that it fails on is `__provides__`. This is on a closed system so I am un able to copy and paste as easily as I hoped. I will try and get the stack track extracted from the image and added to the issue.

image

95dec78c-50a1-47db-a2f4-b96bae02886c

9.5.0

duckdb, oracle, pandas, polars, sqlite

### Relevant log output

_No response_

### Code of Conduct

millsks:
Update common typing to skip attributes when attempting to load properties for various data type objects and do not exist.

Resolves #10665

### DIFF
--- a/ibis/tests/test_typing.py
+++ b/ibis/tests/test_typing.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import Any
 
 from ibis.common.typing import get_type_hints
--- a/ibis/tests/test_typing.py
+++ b/ibis/tests/test_typing.py
@@ -1,4 +1,3 @@
-import pytest
 from typing import Any
 
 from ibis.common.typing import get_type_hints
--- a/ibis/tests/test_typing.py
+++ b/ibis/tests/test_typing.py
@@ -1,0 +1,17 @@
+import pytest
+from typing import Any
+
+from ibis.common.typing import get_type_hints
+
+class TestClass:
+    a: int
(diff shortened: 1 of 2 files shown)


ibis/tests/test_typing.py
```diff
@@ -0,0 +1,22 @@
...
+        return "test"
+
+    @property
+    def c(self) -> Any:
+        raise AttributeError("Property 'c' raises AttributeError")
+
+
+def test_get_type_hints_with_attribute_error():
+    hints = get_type_hints(TestClass, include_properties=True)
+    assert hints == {"a": int, "b": str}
```
cpcloud, line 22: This is failing because the attribute error isn't raised until accessing an instance property, but the function that extracts these properties operates on classes, not instances.

cpcloud (changes requested):

cpcloud:
I don't think we have enough information about the original issue to warrant a solution yet. Let's discuss on the original issue if you're still interested in fixing this.

PR #10669, opened 2025-01-15
refactor(risingwave): decouple the risingwave backend from the postgres backend

cpcloud:
Decouple the risingwave backend from the postgres backend to allow use of `psycopg` (not `psycopg2`) with the postgres backend in #10659.


gforsyth (approved):
:shipit:

Issue #10413, opened 2024-11-01
feat: add method to_json()  (at least for duckdb backend)


cboettig:
### Describe the solution you'd like

Duckdb supports https://duckdb.org/docs/guides/file_formats/json_export.html, including support for JSON line format or ARRAY format.  

native export methods are particularly handy since they can stream data too big for RAM, and can easily stream to remote sources like S3 buckets.  It would be great to see this included among the other .to_* methods supported by duckdb.  

I could take a go at a PR perhaps if you gave me a few pointers?  but might be faster to roll this without me.  

### What version of ibis are you running?

9.5.0

### What backend(s) are you using, if any?

DuckDB

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
This seems entirely reasonable. Not sure what other backends support writing JSON, but it's valuable enough for DuckDB and generically useful enough to have it available as an API IMO.

cpcloud:
Since this is a feature addition, I'm going to remove it from the 10.0 milestone in the interest of trying to get a release out in January 2025.

cboettig:
sounds good.  Just noting the my current work-around to this (because somehow I always forget this)

```python
expr = con.read_parquet(...).filter(...)
query = ibis.to_sql(expr)
con.raw_sql(f"COPY ({query}) TO 's3://{bucket}/{path}/example.json' (FORMAT JSON, ARRAY true);")
```

PR #10681, opened 2025-01-17
feat(duckdb): add to_json() to Table and duckdb backend

NickCrews:
Fixes https://github.com/ibis-project/ibis/issues/10413

I didn't bother adding this to any of the other backends. There is no handy pyarrow.to_json() function that can save the day here. I think we would need to to backend-specific implementations.

Open todos that could come later, but I don't think should be blocking for this PR:
- Consider taking some of the [duckdb-specific config options](https://duckdb.org/docs/sql/statements/copy.html#json-options), and hoisting them to be "official" options across all backends. I could see a `format=Literal["array", "jsonl"] = "array"` being the most likely choice, but not needed until we survey the other backends.
- For dates/datetimes/decimals, where there is no corresponding JSON type, we need to decide what to do.
- Once we decide, implement that.
- Maybe add more tests to ensure that backends produce the same JSON strings? eg maybe the implementation of the test currently, of just reading back using pandas, is too lenient, and multiple on-disk values would lead to the same result. Maybe we actually want to look at the on-disk representation.

### DIFF
--- a/ibis/backends/duckdb/__init__.py
+++ b/ibis/backends/duckdb/__init__.py
@@ -8,7 +8,7 @@ import urllib
 import warnings
 from operator import itemgetter
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 import duckdb
 import pyarrow as pa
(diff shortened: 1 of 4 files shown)


cpcloud:
> I could see a `format=Literal["array", "jsonl"] = "array"` being the most likely choice, but not needed until we survey the other backends.

I think I'd rather have `to_json` and `to_jsonl` methods. Then there's never an ambiguity about what the word "JSON" means when you're reading the code.

> Maybe add more tests to ensure that backends produce the same JSON strings?

I think we should avoid this and instead test that round-tripping produces semantically equivalent pyarrow tables or dataframes.

For example, two JSONL files with the same rows in a different order are equivalent, but comparing their on-disk representation would indicate they were different.

NickCrews:
> I think I'd rather have to_json and to_jsonl methods

Sounds good to me. I'll just keep this as `to_json` for now and NOT expose any format options. we can add `to_jsonl` later if requested.

> I think we should avoid this and instead test that round-tripping produces semantically equivalent pyarrow tables or dataframes.

Want me to switch from pandas (as I do here) to pyarrow? Or is pandas fine?

NickCrews:
I added `trino` as an xfail. Now tests should be passing.

I also actually implemented all the duckdb-specific options of compression, dateformat, and timestampformat. I didn't add any tests though since it felt like overkill.

I explicitly did NOT add the `array` option, which means users cannot get jsonl through this backdoor. I want that to actually get implemented with the `to_jsonl` solution when the time comes.

NickCrews:
failing CI looks like an unrelated flake, it didn't fail in [this previous CI run](https://github.com/ibis-project/ibis/actions/runs/12823049965?pr=10681)

cpcloud:
> Want me to switch from pandas (as I do here) to pyarrow?

DataFrames are fine if that's what you started with. No need to redo.

ibis/backends/duckdb/__init__.py
```diff
@@ -1566,6 +1566,45 @@ def to_geo(
...
+        dateformat
+            Date format string.
+        timestampformat
+            Timestamp format string.
+        """
+        opts = f", COMPRESSION '{compression.upper()}'"
+        if dateformat:
+            opts += f", DATEFORMAT '{dateformat}'"
+        if timestampformat:
+            opts += f", TIMESTAMPFORMAT '{timestampformat}'"
```
cpcloud, line 1603: Eventually this should be done with sqlglot as much as possible, but this is fine for now.

cpcloud (approved):
Thanks!

ibis/backends/duckdb/__init__.py
NickCrews: Ah I see, thanks!

Issue #9185, opened 2024-05-13
feat: Table.to_records()


NickCrews:
In many places in my code I do something like `table.to_pandas().to_records()`. It could be nice to have a method that goes straight there. IDK, not a huge deal, the existing solution is really not that bad. But just wanted to get an issue up for discussion.

Based on @mw3i in https://github.com/ibis-project/ibis/issues/5391#issuecomment-2107806744_:

Sorry for the late reply. Since ibis doesn't have upsert, our team uses the [dataset library](https://dataset.readthedocs.io/en/latest/) for inserts/updates/upserts. We do all our normal code in ibis, and then updates/upserts are done with:

```python
import dataset

params = f"{dialect}://{user}:{password}@{url}:{port}/{database}" # <-- sqlalchemy database uri string
with dataset.connect(params) as dbx:
    dbx[tablename].upsert_many(
        list_of_dictionaries, # <-- each dict is a db entry
        column_to_use_as_identifier,
    )
```

NickCrews:
I think this implementation would work:

```
    def as_dicts(self, chunk_size: int = 1_000_000) -> Iterable[dict[str, Any]:
        """Iterate through each row as a `dict` of column_name -> value."""
        for batch in self.to_pyarrow_batches(chunk_size=chunk_size):
            yield from batch.to_pylist()
```

1. it uses pyarrow, which I think we want to prioritize more than pandas
2. it allows streaming, not materializing the whole thing at once. Users can

PR #10697, opened 2025-01-21
feat: add Expr.to_dicts()

NickCrews:
fixes https://github.com/ibis-project/ibis/issues/9185

I'm not sure if we want to sneak this in for 10.0, the implementation doesn't look that hard to review, but the increased API is really what to watch out for here I think.


ibis/backends/__init__.py
```diff
@@ -586,6 +586,44 @@ def to_delta(
         with expr.to_pyarrow_batches(params=params) as batch_reader:
             write_deltalake(path, batch_reader, **kwargs)
 
+    @util.experimental
+    def to_dicts(
```
kszucs, line 590: I don't have a strong preference, but maybe call it `to_records()`?

cpcloud:
One of the issues with methods like these (to_list, to_dicts, to_records) is that their API tends to add support for different shapes by stuffing in a bunch of non-mutually exclusive arguments.

Can we a pick a name for this that is both concise **and** mostly-unambiguously indicates the shape of the dictionary/list-of-dictionaries?

cpcloud:
Here's my current thoughts:

- `to_dicts`: implies a sequence because of the pluralization, pigeonholes the output type in perpetuity to literal python dictionaries
- `to_records`: implies a mapping or something like named tuple (a thing with fields), leaves the individual record type open.

NickCrews:
I went through the same train of thought as Phillip and that is why I chose to_dicts after considering to_records. Phillip are you saying you are happy with to_dicts or do you see a problem with it?

The two basic shapes I see are Iterable[mapping] (this PR) and Iterable[Iterable] such as an Iterable of tuples or named tuples. Is there another basic shape we might want to support?

I DEFINITELY want different methods for the different shapes. Eg avoid pandas chaos where their to_records() gives you totally different things depending on what you pass to orient. But I'm not sure if I think we need a single method for every concrete type. Eg do we want to_tuples() and to_namedtuples(), or to_rows(record_type=tuple | namedtuple)?
I think I am more of a fan of splitting them into separate methods. One reason is that namedtuples blue the line between being an Iterable and being a mapping.

cpcloud:
> Phillip are you saying you are happy with to_dicts or do you see a problem with it?

The only problem I can see with `to_dicts()` is that if we ever want to return something other than a list of `dict`s then we have to implement a new method.

`to_records` slightly hedges against that possibility by using a more abstract noun.

On the other hand that may never ben an issue.

NickCrews:
I could see how in the future we might want to return a `Iterable[tuple]`. However, I would NEVER want us to have one function that returns either `Iterable[Mapping]` or `Iterable[tuple]`, they should be two different functions. So what should they be called? I worry that `to_records` could be understood as returning EITHER of these, so I think we should be more explicit, and avoid that name. I think my favorites at this point are the explicit `to_dicts()` and `to_tuples()`, and make them typed as returning the concrete type. If we add options for eg other Mapping-likes, then we `@typing.overload` them with those other concrete options.

We COULD call it `.to_mappings()`, but that just feels not idiomatic, I think that would be more confusing than helpful.

cpcloud:
> If we add options for eg other Mapping-likes, then we @typing.overload them with those other concrete options.

I'm not sure that overloading on return type makes sense. How would a type system be able to determine  which overload was meant?

NickCrews:
Sorry, I think I over-explained that, I'm just thinking of normal overloading

```python
# or could use the type dict instead of the string literal "dict"
@typing.overload
def to_dicts(self, *, cls: Literal["dict"]) -> Iterable[dict[str, Any]]: ...

@typing.overload
def to_dicts(self, *, cls: Literal["mydict"]) -> Iterable[MyDict[str, Any]]: ...

def to_dicts(self, *, cls = "dict"): ...
```

PR #10700, opened 2025-01-21
fix(bigquery): let bigquery backend respect window frame set by users

### LINKED ISSUE #10699: bug: cumsum function does not work as expected in Bigquery backend due to ibis default window frame logic
### What happened?

I would like to calculate the cumulative sum of a target column by rows by using ibis. Here is a sample code snippet used:
```
import ibis
import ibis.backends.bigquery as bigquery

con = ibis.backends.bigquery.connect(project_id="pers-decision-engine-dev", dataset_id="pde_food_sow_20250121_dktlu3ov")
tbl = con.table("pers-decision-engine-dev.pde_food_sow_20250121_dktlu3ov.ranking_stg_chosen_exploit_offers_grow")
tbl_muted = tbl.mutate(running_cost=tbl.cost.cumsum(order_by="alpha"))
print(ibis.to_sql(tbl_muted))
```
The generated SQL is
```
SELECT
  `t0`.`cust_id`,
  `t0`.`campaign_id`,
  `t0`.`cost`,
  `t0`.`model_predicted_cost`,
  `t0`.`calibration_factor`,
  `t0`.`rewards_estimate`,
  `t0`.`static_rand`,
  `t0`.`roi`,
  `t0`.`alpha`,
  SUM(`t0`.`cost`) OVER (ORDER BY `t0`.`alpha` ASC) AS `running_cost` -- problematic window function to work out cumulative sum by rows
FROM `pers-decision-engine-dev`.`pde_food_sow_20250121_dktlu3ov`.`ranking_stg_chosen_exploit_offers_grow` AS `t0`
```
According to [the paper searched in google here](https://medium.com/google-cloud/bigquery-window-functions-advanced-techniques-for-data-professionals-e0a6ca3f9502#:~:text=If%20the%20ORDER%20BY%20expression,UNBOUNDED%20PRECEDING%20AND%20CURRENT%20ROW%20.) the Bigquery logic with default window frame is :

Image

I did a experiment to prove this Bigquery behavior with different window frames by the calculation of the cumulative sum by rows:

Image

From above result, It can tell that only the `rows_window_frame_running_rewards ` with explict rows window frame giving out the correct running sum result.

Therefore, an explicit rows window frame must be respected by ibis instead of dropping the window spec and using the default window frame when the range is `BETWEEN UNBOUND PRECEDING AND CURRENT ROW`. 

The Bigquery default window frame behavior shouldn't be used by ibis to implement the `cumusum` function and this logic further impacts the `cumulative_window`, `rows_window` and et al.

As limited testing, the problem exists:
- 9.2.0
- 9.5.0

big query

### Relevant log output

```sh
(venv) ➜  ibis-debug pip list | grep ibis
ibis-framework                9.5.0
(venv) ➜  ibis-debug python -V
Python 3.10.16
(venv) ➜  ibis-debug
```

### Code of Conduct

maxshine:
## Description of changes

Removed the bigquery specific logic dropping window frame of `BETWEEN UNBOUND PRECEDING AND CURRENT ROW` so it will keep the window frame as set by users.

## Issues closed

* Resolves #10699

### DIFF
--- a/ibis/backends/sql/compilers/bigquery/__init__.py
+++ b/ibis/backends/sql/compilers/bigquery/__init__.py
@@ -324,13 +324,6 @@ class BigQueryCompiler(SQLGlotCompiler):
 
     @staticmethod
     def _minimize_spec(start, end, spec):
-        if (
-            start is None
-            and isinstance(getattr(end, "value", None), ops.Literal)
-            and end.value.value == 0
-            and end.following
-        ):
-            return None
         return spec
 
     def visit_BoundingBox(self, op, *, arg):


maxshine:
Hi @cpcloud 

As I observed from CI workflow, Bigquery backend test suites were successful
image

image



Would you please kindly point me to the failed cases and I am happy to contribute to the corresponding tests?

Thanks

Issue #10703, opened 2025-01-22
feat: support Deferreds as predicates to joins


NickCrews:
### Is your feature request related to a problem?

I have these two tables, which I want to join regardless of capitalization. I want to be able to do this:

```python
import ibis
t1 = ibis.memtable({"x": ["a", "b", "c"]})
t2 = ibis.memtable({"x": ["A", "B", "Z"]})
j = t1.join(t2, ibis._.x.upper())
# expect {"x": ["a", "b"], "x_right": ["A", "B"]}
```

Currently I have to actually pass `t1.x.upper() == t2.x.upper()`.
These new semantics are basically the same as the current semantics of passing `"x"` as the preds: For the given pred, assume it is referencing a column in the table, and get that Column.

### What is the motivation behind your request?

just a little ux improvement

### Describe the solution you'd like

modify the join code to be resolve Deferreds as I describe.

### What version of ibis are you running?

main

### What backend(s) are you using, if any?

should work for all of them

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

NickCrews:
After @kszucs's [comment](https://github.com/ibis-project/ibis/pull/10704#discussion_r1924852603), I think this needs more general discussion of the spec before I dive into an implementation.

After thinking about it and experimenting with current behavior, I have a more generalized proposal for how to adjust join conditions.

I didn't describe it in the original post, but a large motivation of my suggestions here is that they allow for reusing predicates between different joins, since they never reference a table directly:

```python
preds = [
    "a",
    _.x.upper(),
    lambda t: t.y.abs(),
    lambda l, r: l.foo == r.bar,
    (_.baz, _.qux),
]
a.join(b, preds)
c.join(d, preds)
```

Ok, for the below discussion, consider this setup:

```python
import ibis
from ibis import _

lt = ibis.memtable({"f": [1, 2, 3]})
rt = ibis.memtable({"f": [2, 3, 4]})
```

# "Single" Values

Currently, if given something that is "singular", it should be resolved against each table individually, and then combined with ==.

## `lt.join(rt, "f")`
Currently works as expected. My proposals below are just generalizations of this.

## `lt.join(rt, lambda t: t.f)`

Currently, works as expected, the same as `lt.join(rt, "f")`

After, it works the same.

## `lt.join(rt, _.f)`

Currently, it fails (resolves against `lt`, giving `lt.f`, which isn't a BooleanColumn)

[...]

PR #10704, opened 2025-01-22
feat(api): support passing Deferreds as predicates to joins

NickCrews:
resolves https://github.com/ibis-project/ibis/issues/10703


ibis/expr/types/joins.py
```diff
@@ -168,7 +168,7 @@ def prepare_predicates(
 
     left, right = chain.to_expr(), right.to_expr()
     for pred in util.promote_list(predicates):
-        if isinstance(pred, (Value, Deferred, bool)):
+        if isinstance(pred, (Value, bool)):
```
NickCrews, line 171: I'm a little terrified that simply deleting this didn't break any existing tests. Do you know what this is for? If it's on purpose, we should add a test.

kszucs: It was used for the following case: `_.left_field == t.right_field` 

The problem is that a predicate possibly references two tables but the deferred is a single object, we cannot easily (and always correctly) decompose it into two sides. Hence we require the deferred predicates to use the left table as the placeholder.

kszucs: IIRC there was a failing test case at the time of the refactoring. If that's not the case then we should add one to clarify this use case.

NickCrews: I wrote a more general overview of this behavior [here](https://github.com/ibis-project/ibis/issues/10703#issuecomment-2613725392), let's move discussion there, since it is more about the spec, and not the implementation.

PR #10716, opened 2025-01-24
docs: clarify `distinct` in difference(), add examples

NickCrews:
I recently had to use this, and the semantics were not clear to me.
I had to do this experimentation myself, so I thought I would write them down.

### DIFF
--- a/ibis/expr/api.py
+++ b/ibis/expr/api.py
@@ -2180,7 +2180,7 @@ def difference(table: ir.Table, *rest: ir.Table, distinct: bool = True) -> ir.Ta
     >>> t1 = ibis.memtable({"a": [7, 8, 8, 9, 9, 9]})
     >>> t2 = ibis.memtable({"a": [8, 9]})
 
-    With distinct=True, if a row ever appears in any of `*rest`,
+    With `distinct=True`, if a row ever appears in any of `*rest`,
     it will not appear in the result.
     So here, all appearances of 8 and 9 are removed:
 
@@ -2193,7 +2193,7 @@ def difference(table: ir.Table, *rest: ir.Table, distinct: bool = True) -> ir.Ta
     │     7 │
     └───────┘
 
-    With `distinct=False`, the algorithm is more of a multiset/bag difference.
+    With `distinct=False`, the algorithm is a [multiset](https://en.wikipedia.org/wiki/Multiset) difference.
     This means, that since 8 and 9 each appear once in `t2`,
     the result will be the input with a single instance of each removed:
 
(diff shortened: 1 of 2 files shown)


ibis/expr/api.py
```diff
@@ -2173,35 +2177,50 @@ def difference(table: ir.Table, *rest: ir.Table, distinct: bool = True) -> ir.Ta
     --------
     >>> import ibis
     >>> ibis.options.interactive = True
-    >>> t1 = ibis.memtable({"a": [1, 2]})
-    >>> t1
+    >>> t1 = ibis.memtable({"a": [7, 8, 8, 9, 9, 9]})
+    >>> t2 = ibis.memtable({"a": [8, 9]})
+
+    With distinct=True, if a row ever appears in any of `*rest`,
```
cpcloud, line 2183:
```suggestion
    With `distinct=True`, if a row ever appears in any of `*rest`,
```

ibis/expr/api.py
```diff
@@ -2173,35 +2177,50 @@ def difference(table: ir.Table, *rest: ir.Table, distinct: bool = True) -> ir.Ta
...
     │ int64 │
     ├───────┤
-    │     1 │
-    │     2 │
+    │     7 │
     └───────┘
-    >>> t2 = ibis.memtable({"a": [2, 3]})
-    >>> t2
+
+    With `distinct=False`, the algorithm is more of a multiset/bag difference.
```
cpcloud, line 2196:
```suggestion
    With `distinct=False`, the algorithm is a [multiset](https://en.wikipedia.org/wiki/Multiset) difference.
```

Let's add a link to the wikipedia page for multiset in case the terminology or concept is unfamiliar to users.

ibis/expr/api.py
```diff
@@ -2173,35 +2177,50 @@ def difference(table: ir.Table, *rest: ir.Table, distinct: bool = True) -> ir.Ta
...
-    │     3 │
+    │     7 │
+    │     8 │
+    │     9 │
+    │     9 │
     └───────┘
-    >>> ibis.difference(t1, t2)
+
+    With multiple tables in `*rest`, we apply the operation consecutively.
+    Here, we we remove two 8s and two 9s:
```
cpcloud, line 2213:
```suggestion
    Here, we remove two eights and two nines:
```

cpcloud (changes requested):

cpcloud (approved):

NickCrews:
Thanks @cpcloud, those edits look great!

Issue #10391, opened 2024-10-29
feat: better repr for DataType


NickCrews:
### Is your feature request related to a problem?

This is a just a small papercut. If I have a complex dtype, the repr is hard to read, all in one line:

```python
import ibis

dt = ibis.dtype(
    "array<struct<line1: string, line2: string, city: string, state: string, zipcode: string>>"
)
dt
# Array(value_type='Struct([('line1', String(nullable=True)), ('line2', String(nullable=True)), ('city', String(nullable=True)), ('state', String(nullable=True)), ('zipcode', String(nullable=True))], nullable=False), nullable=False)
```

I think the most readable form would be a pretty-printed version of the string DSL (option A):

```
dtype("""!array<
    !struct<
        line1: string,
        line2: string,
        city: string,
        state: string,
        zipcode: string
    >
>""")
```

which is also able to be copy-pasted back into a REPL to recreate the original dtype.

Could also just do the string (option B):

```
!array<
    !struct<
        line1: string,
        line2: string,
        city: string,
        state: string,
        zipcode: string
    >
>
```

We should be clever, and only do newlines/pretty if the type is over some metric of complexity, perhaps just string length?

eg I would want `array<string>` and not

```
array<
    string
>
```

I am basing these ideas off of two needs: human readbility, and being able to repro the dtype in the REPL from the output. Are there other considerations I'm not thinking of?

### Describe the solution you'd like

see above

### What version of ibis are you running?

main

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cpcloud:
> We should be clever, and only do newlines/pretty if the type is over some metric of complexity, perhaps just string length?

I'd encourage you to look into [the staggering complexity](https://journal.stuffwithstuff.com/2015/09/08/the-hardest-program-ive-ever-written/) of pretty printing a string before taking any action on this.

We should **definitely** avoid anything that moves us in the direction of having to format the type string.

NickCrews:
Ooof, ok thanks for that, I didn't expect it would be that tricky, but I'm sold.

What about going to the more succinct "array<>" DSL, but that's the only change, and keeping everything on one line as it currently is?

cpcloud:
I'm not sure why we need to change `__repr__` here. It was an intentional choice **not** to make `__repr__` the same as `__str__` for data types.

`__repr__` is really for development printing and `__str__` is for prett(ier) printing.

[...]

PR #10726, opened 2025-01-26
feat: return DSL string for DataDatatype.__repr__()

NickCrews:
fixes https://github.com/ibis-project/ibis/issues/10391

I can add more tests or otherwise modify them if you think wise, but probably I think what we have here is sufficient.

One sanity check could be a roundtrip check?
```python
for dsl in ["int8", "uint8" ,...]:
    dt = ibis.dtype(dsl)
    assert dsl == str(dt)
    assert dsl == repr(dt)
```


cpcloud:
See my comment on #10391. I don't think we should do this.

NickCrews:
I think I agree, closing.

PR #10729, opened 2025-01-26
feat(datatypes): add support for fixed length arrays

### LINKED ISSUE #10719: feat: Improve support for fixed length array in DuckDB

I was trying to rewrite a DuckDB query to Ibis, and ran into this error:
```
BinderException: Binder Error: No function matches the given name and argument types 'array_cosine_distance(DOUBLE[], DOUBLE[])'. You might need to add explicit type casts.
```

Here is the function and the attempt at rewriting it:

```python
import duckdb
import ibis
from ibis import _
from sentence_transformers import SentenceTransformer
from sentence_transformers.models import StaticEmbedding

static_embedding = StaticEmbedding.from_model2vec("minishlab/potion-base-8M")
model = SentenceTransformer(modules=[static_embedding])


def similarity_search_duckdb(
    query: str,
    k: int = 5,
    dataset_name: str = "ai-blueprint/fineweb-bbc-news-embeddings",
    embedding_column: str = "embeddings",
):
    query_vector = model.encode(query)
    embedding_dim = model.get_sentence_embedding_dimension()

    sql = f"""
        SELECT 
            *,
            array_cosine_distance(
                {embedding_column}::float[{embedding_dim}], 
                {query_vector.tolist()}::float[{embedding_dim}]
            ) as distance
        FROM 'hf://datasets/{dataset_name}/**/*.parquet'
        ORDER BY distance
        LIMIT {k}
    """
    return ibis.memtable(duckdb.sql(sql).to_arrow_table())


t1 = similarity_search_duckdb("What is the future of AI?")
print(t1)


@ibis.udf.scalar.builtin
def array_cosine_distance(x, y) -> float:
    """Compute cosine similarity between two vectors."""


def similarity_search_ibis(
    query: str = "What is the future of AI?",
    k: int = 5,
    dataset_name: str = "ai-blueprint/fineweb-bbc-news-embeddings",
    embedding_column: str = "embeddings",
):
    # Use same model as used for indexing
    query_vector = model.encode(query)
    embedding_dim = model.get_sentence_embedding_dimension()

    return (
        ibis.read_parquet(f"hf://datasets/{dataset_name}/**/*.parquet")
        .mutate(
            distance=array_cosine_distance(
                _[embedding_column].cast("array<float64>"),
                ibis.array(query_vector).cast("array<float64>"),
            )
        )
        .order_by(_.distance.desc())
        .limit(k)
        .drop(embedding_column)
    )

t2 = similarity_search_ibis("What is the future of AI?")
print(t2)
```

I found a related issue: https://github.com/ibis-project/ibis/issues/7963

This ensures DuckDB tables with fixed-length arrays can be used in Ibis, but from what I can tell Ibis treats this as a variable-length array. I haven't found a way to create a fixed-length array in a DuckDB table using Ibis. The array_cosine_distance function only supports fixed-length arrays.

I'm trying to use Ibis wherever I would previously have used SQL or Pandas, both as a learning exercise and because I'm hoping Ibis can become my default data manipulation library.

I'd like some way to create fixed-length array columns in DuckDB using Ibis.

9.5.0

duckdb 1.1.3

### Code of Conduct

cpcloud:
Add `length` field to `Array` datatypes to support fixed-length arrays.

Closes #10719.

### DIFF
--- a/ibis/expr/datatypes/core.py
+++ b/ibis/expr/datatypes/core.py
@@ -8,6 +8,7 @@ from abc import abstractmethod
 from collections.abc import Iterable, Iterator, Mapping, Sequence
 from numbers import Integral, Real
 from typing import (
+    Annotated,
     Any,
     Generic,
     Literal,
@@ -27,7 +28,7 @@ from ibis.common.annotations import attribute
 from ibis.common.collections import FrozenOrderedDict, MapSet
 from ibis.common.dispatch import lazy_singledispatch
 from ibis.common.grounds import Concrete, Singleton
-from ibis.common.patterns import Coercible, CoercionError
+from ibis.common.patterns import Between, Coercible, CoercionError
 from ibis.common.temporal import IntervalUnit, TimestampUnit
 
 
@@ -50,14 +51,14 @@ def dtype(value: Any, nullable: bool = True) -> DataType:
     >>> ibis.dtype("int32")
     Int32(nullable=True)
     >>> ibis.dtype("array<float>")
-    Array(value_type=Float64(nullable=True), nullable=True)
+    Array(value_type=Float64(nullable=True), length=None, nullable=True)
 
     DataType objects may also be created from Python types:
 
     >>> ibis.dtype(int)
     Int64(nullable=True)
     >>> ibis.dtype(list[float])
-    Array(value_type=Float64(nullable=True), nullable=True)
+    Array(value_type=Float64(nullable=True), length=None, nullable=True)
 
     Or other type systems, like numpy/pandas/pyarrow types:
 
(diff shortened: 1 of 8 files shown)


ibis/expr/datatypes/core.py
```diff
@@ -309,6 +310,10 @@ def is_enum(self) -> bool:
         """Return True if an instance of an Enum type."""
         return isinstance(self, Enum)
 
+    def is_fixed_length_array(self) -> bool:
```
kszucs, line 313: Can you add a test case for it?

kszucs:
Could you please create a follow-up ticket to adjust the hypothesis strategy [here](https://github.com/ibis-project/ibis/blob/main/ibis/tests/strategies.py#L153)?

kszucs (approved):
Just a nit, otherwise it LGTM.

cpcloud:
> Could you please create a follow-up ticket to adjust the hypothesis strategy [here](https://github.com/ibis-project/ibis/blob/main/ibis/tests/strategies.py?rgh-link-date=2025-01-28T09%3A25%3A41Z#L153)?

No need, I adjusted it here and there were no issues!

PR #10748, opened 2025-01-29
fix(postgres): fix insertion of `NaT`/`None` into timestamp columns

### LINKED ISSUE #10742: bug: [POSTGRES] Temp table creation breaks if NULL in date column
### What happened?

When trying to create a temp table in the Postgres backend, the creation fails due to the pandas NaT value replacing actual None values. This leads to an invalid SQL insert statement. I assume that this is also the case for the other backends but have not verified it.

```python
connection = ibis.postgres.connect(<db_args>)
test_data = pl.DataFrame({"testcol": [datetime.now(), datetime.now(), None]})
connection.create_table(name="tempTable123", obj=test_data, temp=True)
```

Got a dirty fix working by transforming the column to str and then replacing `'NaT'` with actual None values while trying some things.
Inside `ibis/backends/postgres/__init__.py` added `from pandas.api.types import is_datetime64_any_dtype` import and the following lines inside `_register_in_memory_table`
```python
...
for col in convert_df.columns:
...
    if is_datetime64_any_dtype(convert_df[col]):
        df[col] = df[col].astype(str).apply(lambda x: None if x == "NaT" else x)
```

9.5.0

Postgres

### Relevant log output

```sh
psycopg2.errors.InvalidDatetimeFormat: invalid input syntax for type timestamp: "NaT"
LINE 1: ...e_7jtpt3un5nhz7ebyga2syhligu" ("testcol") VALUES ('NaT'::tim...
```

### Code of Conduct

cpcloud:
Fixes #10742.

### DIFF
--- a/ibis/backends/postgres/__init__.py
+++ b/ibis/backends/postgres/__init__.py
@@ -8,6 +8,7 @@ from operator import itemgetter
 from typing import TYPE_CHECKING, Any
 from urllib.parse import unquote_plus
 
+import psycopg
 import sqlglot as sg
 import sqlglot.expressions as sge
 from pandas.api.types import is_float_dtype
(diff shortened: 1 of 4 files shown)


cpcloud:
Merging on green to get this into 10.0!

PR #10753, opened 2025-01-30
feat: make Table.join() use the same logic as Table.rename()

NickCrews:
WIP, I need to edit this description to share my motivation


cpcloud:
> change the default values for lname param of the .join() methods, from "" to "{name}". This by itself doesn't actually have any behavior change, it just is a lot more symmetrical with the other argument. This default is a remnant of when we used the lsuffix and rsuffix API.

This seems like a behavior change when chaining joins with overlapping column names.

It's a bit no tests are failing, as this comes up in the TPC-DS query set quite a few times.

Can you add a small test that fails before this change and then passes with your change so that we have a record of this behavior being run at least somewhere?

Generally speaking, reusing renaming doesn't seem like a huge win to me, because while there may be commonality between the `.rename` API and `join` renaming, I think sharing logic between these two makes understanding each one in isolation more difficult. `rename` is an explicit request from the user, while join renaming is mostly hidden and automatic.

NickCrews:
> This seems like a behavior change when chaining joins with overlapping column names.

Sorry, can you be more verbose with an example?
 
Did you see the commit description for
https://github.com/ibis-project/ibis/commit/28848ff803b18b99103b4d32e034e274185e75e6? I don't think changing the default will be a behavior change, because there is some existing code that makes "" and "{name}" behave the same. Whether or not we remove that code, and make the two behave differently, is a separate question.

PR #10778, opened 2025-02-03
test(join): add test that colliding column names will get merged into one

NickCrews:
I'm adding this PR mostly as a demo of what I think might be undesirable behavior. Should this instead error, because the two input columns are getting mapped to the same output column?

### DIFF
--- a/ibis/tests/expr/test_table.py
+++ b/ibis/tests/expr/test_table.py
@@ -1786,6 +1786,9 @@ def test_join_lname_rname(how):
     expr = method(right, rname="right_{name}", lname="left_{name}")
     assert expr.columns == ("left_id", "first_name", "right_id", "last_name")
 
+    expr = method(right, rname="{name}_z", lname="{name}_z")
+    assert expr.columns == ("id_z", "first_name", "last_name")
+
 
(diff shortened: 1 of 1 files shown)


cpcloud:
Yep, this isn't entirely correct.

The name is only allowed to be merged into a single output column in the case of inner joins (with equality predicates only, which is implicit in this test).

NickCrews:
In general, I think it is very fragile and bad to need to explicitly wrap every single public method of `Table` with the `finished()` decorator in the `Join` class. It also is some spooky-action-at-a-distance behavior here, where things will only blow up at some later function call, not close to where the source of the problem is. This is different from the rest of ibis, which usually tries to error as early/close to the original cause/function call as possible. I would much prefer it if we forced the user, at the time they call `.join()` to tell us how to resolve these collisions. But this is almost definitely a breaking API change, and not in scope for this PR. But just writing it down somewhere as I'm thinking of it.

cpcloud (approved):

kszucs:
We could automatically add the methods/properties after the class definition to make it more future-proof to the `ir.Table` changes.

Issue #10749, opened 2025-01-29
bug: Schema not found the second time a spark table is accessed


christophediprima:
### What happened?

I am trying to insert into two different iceberg tables using spark connect. But as soon as I access a table using `get_schema`,`table` or `insert` I won't be able to do it a second time.

My code looks like this:

```python
import ibis
from   pyspark.sql import SparkSession

session = SparkSession.builder.getOrCreate()
con = ibis.pyspark.connect(session)

new_dataframe = Dataframe creation here

con.insert(
  "my_table",
  new_dataframe,
  database="my_catalog.my_database",
)

new_dataframe_2 = Dataframe creation here

con.insert(
  "my_other_table",
  new_dataframe_2,
  database="my_catalog.my_database",
)
```

First insert will work but not the second one. If first insert is commented, second one will work. Same thing will happen regardless of the `get_schema`,`table` or `insert` method.

### What version of ibis are you using?

ibis-framework[pyspark]==10.0.0.dev490

### What backend(s) are you using, if any?

PySpark

### Relevant log output

```sh
pyspark.errors.exceptions.connect.AnalysisException: [SCHEMA_NOT_FOUND] The schema `my_catalog`. cannot be found. Verify the spelling and correctness of the schema and catalog.
If you did not qualify the name with a catalog, verify the current_schema() output, or qualify the name with the correct catalog.
To tolerate the error on drop use DROP SCHEMA IF EXISTS.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

cpcloud:
Thanks for the issue!

Can you show a fully reproducible example that I can copy-paste into a file or a Python REPL?

The 

> Dataframe creation here

bits seem critical to reproducing the problem for example.

Toy data is completely fine.

gforsyth:
There does seem to be a bug here, which is that `force=True` on `con.create_database` isn't generating the expected `IF NOT EXISTS` clause

gforsyth:
Found the issue, I'll put up a PR

christophediprima:
Thanks for your replies. I used .sql as a workaround so I can provide a working full example to reproduce the initial bug I was reporting.

Did you find the solution for my initial bug report or `force=True` one?

gforsyth:
`force=True` -- I can't reproduce your initial bug if the `database` and `table` have been created

gforsyth:
(Although that's with a local spark session, not using Iceberg)

PR #10782, opened 2025-02-03
fix(pyspark): generate `IF NOT EXISTS` if `force=True`

gforsyth:
## Description of changes

We were using the wrong keyword in `sqlglot` for this so it was being
swallowed silently.

## Issues closed

xref #10749

### DIFF
--- a/ibis/backends/pyspark/__init__.py
+++ b/ibis/backends/pyspark/__init__.py
@@ -507,7 +507,7 @@ class Backend(SQLBackend, CanListCatalog, CanCreateDatabase):
 
         sql = sge.Create(
             kind="DATABASE",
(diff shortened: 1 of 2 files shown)


cpcloud (approved):

cpcloud:
Looks like we're making this same mistake in MySQL as well. I'll PR that here in a sec.

cpcloud:
@gforsyth Does this fix the issue you xref'd? Or are you just noting that they are related? If it does fix it, let's add the `Fixes` bit so that it gets closed when this PR gets merged.

gforsyth:
No, there's something else going on in that issue but it's why I came across this bug

PR #10784, opened 2025-02-04
fix(interactive-repr): ensure that null scalars can be repr'd interactively

### LINKED ISSUE #10780: bug(interactive-repr): interactive repr of null date scalar raises an exception
### What happened?

```
In [10]: from ibis.interactive import *

In [11]: s = ibis.null("date")

In [12]: s
```

raises an exception:

```
│ …/ibis/expr/types/pretty.py:139 in _                                                             │
│                                                                                                  │
│   136 @format_values.register(dt.Date)                                                           │
│   137 def _(dtype, values, **fmt_kwargs):                                                        │
│   138 │   dates = [v.date() if isinstance(v, datetime.datetime) else v for v in values]          │
│ ❱ 139 │   return [Text.styled(d.isoformat(), "magenta") for d in dates]                          │
│   140                                                                                            │
│   141                                                                                            │
│   142 @format_values.register(dt.Time)                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'NoneType' object has no attribute 'isoformat'
```

`main` @ 12a45cfeaa25129b58632f1ea2dccbf468a84fbb

_No response_

### Relevant log output

```sh

```

### Code of Conduct

cpcloud:
Fixes #10780.

### DIFF
--- a/ibis/backends/duckdb/tests/test_datatypes.py
+++ b/ibis/backends/duckdb/tests/test_datatypes.py
@@ -1,11 +1,16 @@
 from __future__ import annotations
 
+import string
+
+import hypothesis as h
+import hypothesis.strategies as st
 import numpy as np
 import pytest
 from pytest import param
 
 import ibis
 import ibis.expr.datatypes as dt
+import ibis.tests.strategies as its
 from ibis.backends.sql.datatypes import DuckDBType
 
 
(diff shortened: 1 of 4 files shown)


cpcloud:
This is going to fail with lots of crashing until the next version of DuckDB is released. DuckDB 1.1.3 had some issues with converting bigger fixed length arrays to arrow.

cpcloud:
I'll reduce the test case here to limit the upper bound of fixed length arrays. After all, the thing being fixed and test here is not arrow conversion, but `NULL` printing.

PR #10795, opened 2025-02-05
fix(dev-env): fix aarch64 nix builds

cpcloud:
Fixes failing nix builds


cpcloud:
Seems like we're hitting https://github.com/astral-sh/uv/issues/6105, so I may just merge this in and rerun it on `main` until the nix cache is populated, which will at least get us to passing until the next nixpkgs/uv2nix upgrade and/or the issue itself is addressed.

It may be less of a problem in CI than locally, since the CI machine should at least be running on hardware that ultimately matches the environment build, whereas locally I am building these envs on an x86_64 host (not with cross compilation, but with qemu).

PR #10797, opened 2025-02-06
chore(deps): remove upper bounds on rest of deps except for numpy and pandas

NickCrews:
Followup of https://github.com/ibis-project/ibis/pull/10620
Also fixes https://github.com/ibis-project/ibis/issues/10492

I wanted to use duckdb 1.2 with my production version of ibis 9.5.0, but I can't because ibis set an upper bound on it. This particular issue was addressed in https://github.com/ibis-project/ibis/commit/db0e0c2fc3dc47a8b76b8ddc21460def5cc33df9, but this general class of problem is going to happen again in the future with some other python package. In general, unless we KNOW a version of a peer dep is incompatible, we should not add a restriction

Per [Phillip's comment](https://github.com/ibis-project/ibis/pull/10620#pullrequestreview-2523952591), I kept the upper bound on numpy and pandas, but that's it.


cpcloud:
What's the issue with releasing more often, or cutting minor releases when a dependency upper bound gets automatically updated?

NickCrews:
I have a project that is using ibis 9.5.0, and if there are breaking changes if I upgrade to ibis 10.0, I don't want to spend the effort right now to migrate. If there is a minor version update of ibis on pypi that I can go to that has no breaking changes, then this would work for me. But that's not how we develop ibis, right? we just put everything new on `main`, and if we wanted to cut a minor version update, then we would have to fork a new branch and only cherry pick over the non-breaking changes? And we don't currently do that, right?

What is the problem that you are trying to avoid by keeping the upper bound? regardless of what we do, won't renovate-bot go through every week and do a `uv lock --upgrade` to update the uv.lock file, and then CI will tell us if a new version of a library is incompatible, and THEN we add in the upper bound? Is this more work than cherry picking commits to make frequent non-breaking minor releases?

cpcloud:
The main issue with the `uv lock --upgrade` approach is that it bumps every dependency wholesale, rather than creating separate PRs for the immediate dependencies, which is what happens when there's a bound.

The problem I'm trying to avoid with keep the upper bound is broken installations.

cpcloud:
That said, I think removing the bounds is probably the way to go. Can you document the policy of "your installation may break if a new version of one of Ibis's dependencies introduces a breaking change that affects Ibis" somewhere.

NickCrews:
I updated what will be shown at https://ibis-project.org/concepts/versioning, I think that is actually a fairly good place to put that info.

I could also make it so there are inline notes in the pyproject.toml so that we don't accidentally add upper bounds again in the future? Or is there some other better/automatable way to enforce this policy?

cpcloud:
Dev deps should absolutely remain as is. Plotly 6.0.0 is broken and for dev deps keeping upper bounds doesn't affect them in the same way as library dependencies.

NickCrews:
Cool, that sounds fine to me. Thank you!

PR #10806, opened 2025-02-06
chore: remove trailing whitespace from release notes

cpcloud:
`semantic-release` seems to have added an extra trailing space when generating release notes. HOW DARE IT!


cpcloud:
Good thing, we'll kick off a small army's worth of compute to validate everything here!

